PR #224, opened 2022-12-21
cmd: reverse txid bytes before printing them

positiveblue:
Ensure that the tx id bytes are properly reversed before printing them so we can use the output directly in chain explorers.

This is something that we need to keep in mind because every time that we add a tx id using `[]bytes` won't match the `string` version used for outpoints (which are [reversed before getting encoded](https://github.com/btcsuite/btcd/blob/718d268a8cf55af9b329bac9e171009352fbc10a/chaincfg/chainhash/hash.go#L71)).

Fixes [223](https://github.com/lightninglabs/taro/issues/223)


positiveblue:
Closing in favour of #230

Issue #91, opened 2022-08-24
bip-taro+assets+vm: remove group key duplication, add explicit verification 


Roasbeef:
The original intent of the group key is to allow other assets to be nested under similar assets in the SMT itself. To ensure that this key doesn't collide with other assets [we _tweak_ the "plain" group key with some information unique to the asset and it's location in the chain](https://github.com/lightninglabs/taro/blob/16d56c1b6fbeee544bedda236d69cf33125dd760/asset/asset.go#L89). 

However today in the code, we _encode_ this tweaked key in the final TLV leaf. This means that verifiers aren't able to verify that the key was the result of a valid tweak. Instead, we should be encoding the _plain_ group key (called just [RawKey](https://github.com/lightninglabs/taro/blob/16d56c1b6fbeee544bedda236d69cf33125dd760/asset/asset.go#L270-L272) today). Then the verifiers should take that key, apply the tweak, and validate the final signature against this key. 

I can think of two ways we can do this:
   1. Encode the raw key above, then have all verifiers re-compute the tweaked key to use that properly everywhere 
   2. Have the genesis asset place the _plain_ key in the witness itself 

I personally like the second option as it has less downstream impact. One other bonus is that we'll be able to get rid of duplication across transfers: today we always encode the group key sig, but don't really need to. Instead, this only needs to be verified at the genesis asset (once again the sig can be passed in as witness). 

With the above changes, for a new asset that has a group key, the witness will be: `<plain_fam_pubkey> <sig>`. Which kinda looks like a `p2wkh` output. If the group key is present on the genesis asset (the tweaked key field), then verifiers should: verify the tweaked derivation, verify the msg sig against that key.

dstadulis:
we need to make sure all parties have the keys they need with the tweaks to do proper validation

dstadulis:
Issue goal: ensure family key cannot be collided with explicit verificaiton.

dstadulis:
Capturing some design jam thoughts of how to Asset Group / Key family issuance authortization is handled:

>Rotating the Asset Group / family_key key itself would mean the top level key of all the commitments themselves change:

>Assess if tarod should treat the family key itself as another taproot output key. Then this way, a taro transaction author can do w/e multi-sig threshold, etc construction, in the script path.

This logic should be added to the PR fixing this issue. 

>"the genesis witness should satisfy the tweaked family key, which means either script path or keyspend (the way it is rn)"

halseth:
[...]

PR #225, opened 2022-12-21
Raw key witness encoding

halseth:
See proposed BIP change: https://github.com/Roasbeef/bips/pull/33

To allow verification of the group key, we encode the `raw internal key` as well as the signature over the `asset_id` in the witness of the minting transaction. This lets the verifier derive the group key and verify the signature at genesis.

Rem questions:
- [ ] To allow deriving the group key for a second tranche (having a different `asset_id` than was used to create the group key, we must somehow reference the original `asset_id`?
- [ ] As mentioned in #91 we could change the digest that is signed. Not determined yet what is the best approach here.


halseth:
> Awesome to see the new commits @halseth. Any github comments to contextualize the new commits? We're targeting this PR for v0.2 and I'm wondering what else your PR needs for completion.

@dstadulis I'm working on finalizing the BIP in parallel with the implementation here. I think we have settled on something that works now and I'll try to implement that next week.

Any specific cut-off date for v0.2?

dstadulis:
> Any specific cut-off date for v0.2?

2023-03-10

Roasbeef:
I think this is still relevant and just needs a rebase? 

The other thing we need to do here is ensure that we can handle the version upgrade, as w'ell now have another version with slightly updated verification logic.

dstadulis:
> think this is still relevant and just needs a rebase?

@halseth can you update?

halseth:
@dstadulis @Roasbeef We should agree on the BIP changes first, I rebased https://github.com/Roasbeef/bips/pull/35 that one is ready to merge

dstadulis:
> should agree on the BIP changes first, I rebased [Roasbeef/bips#35](https://github.com/Roasbeef/bips/pull/35) that one is ready to merge

now that's merged, is it possible to finalize the code so it's ready for review?

Issue #223, opened 2022-12-20
rpc: anchor_txid bytes aren't properly reversed on RPC interface 


habibitcoin:
Using latest version of main, note the output of `tarocli assets list` incorrectly flips the bits

```
taro % tarocli assets list
{
    "assets": [
        {
            "version": 0,
            "asset_genesis": {
                "genesis_point": "9ec6ca5637ce2a32e2f528ea64c8decae3a79f22f9ff840e5ef18ae1e5f846a3:1",
                "name": "habibtaro_usd",
                "meta": "68616269627461726f2e636f6d",
                "asset_id": "f130525230298d0edba78266d3a5b41afee6ae81504199c9ee9a96d12ab130ba",
                "output_index": 0,
                "genesis_bootstrap_info": "a346f8e5e18af15e0e84fff9229fa7e3cadec864ea28f5e2322ace3756cac69e000000010d68616269627461726f5f7573640d68616269627461726f2e636f6d0000000000",
                "version": 0
            },
            "asset_type": "NORMAL",
            "amount": "21000000",
            "lock_time": 0,
            "relative_lock_time": 0,
            "script_version": 0,
            "script_key": "02947a51f3b2de3a8c9079330c5edb8ac2b4f9b57cb9d74ccfbb1fc5c0b70d994e",
            "asset_group": {
                "raw_group_key": "036d35daee7a356b0c2207f699bfd69c3491083825e72991d0d3d233dbbdd967ac",
                "tweaked_group_key": "023e3fbb157c2db3b601ede8153b4a8785a791d792e4ea35ddc8d7799d8546fc0c",
                "asset_id_sig": "da2b37a032d7ab3ceca35955023fcaacb558882646d1180df36832547606f3b08c5f1420a3254b257f67d5e50b49d12387a097c26eca36303fa9b4d22e3ac057"
            },
            "chain_anchor": {
[...]

PR #230, opened 2023-01-05
Txid reversal fix

jharveyb:
Alternate fix for #223 by changing the txid types in the protos to strings. Should cover all user-facing display of TXIDs outside of outpoints, which are already displayed correctly.


positiveblue (approved):
One of the advantages to be in alpha is that we can actually modify old fields in the protos.

This approach is much better @jharveyb :100:

PR #235, opened 2023-01-17
Add RPC endpoint for subscribing to ChainPorter event notifications

ffranr:


tarofreighter/chain_porter.go
```diff
@@ -81,6 +81,14 @@ type ChainPorter struct {
 
 	exportReqs chan *AssetParcel
 
+	// subscribers is a map of components that want to be notified on new
+	// events, keyed by their subscription ID.
+	subscribers map[uint64]*chanutils.EventReceiver[Event]
```
Roasbeef, line 86: Can alternatively use/copy/replicate this map type: https://github.com/lightningnetwork/lnd/blob/48c8c1bf48ed9be7a014803ea4c3b4624407dcc3/lntemp/node/sync_map.go#L5

Then this would be:
```go
subscribers SyncMap[uint64, *chanutils.EventReceiver[Event]]
```

Roasbeef: Also we could alternatively lift out this new state/methods into a new struct, then _embed_ that struct within this one.

tarorpc/taro.proto
```diff
@@ -628,3 +635,21 @@ message SendAssetResponse {
...
+        ExecuteSendStateEvent execute_send_state_event = 1;
+    }
+}
+
+message ExecuteSendStateEvent {
+    // Execute timestamp.
+    int64 timestamp = 1;
+
+    // The send state that is about to be executed.
+    string send_state = 2;
```
Roasbeef, line 654: Seems like this should instead be another enum? Otherwise not sure what the actual strucutre would be. 

Or was this meant to be `SendAssetEvent` here?

tarorpc/taro.proto
```diff
@@ -628,3 +635,21 @@ message SendAssetResponse {
 
     int64 total_fee_sats = 5;
 }
+
+message SubscribeSendAssetEventNtfnsRequest {
+}
```
Roasbeef, line 640: Writing down prior comment here re the request of a ID here to permit delivering backlog notifications.

tarorpc/taro.proto
```diff
@@ -99,6 +99,13 @@ service Taro {
     proof file information the receiver needs to fully receive the asset.
     */
     rpc SendAsset (SendAssetRequest) returns (SendAssetResponse);
+
+    /*
+    SubscribeSendAssetEventNtfns registers a subscription to the event
+    notification stream which relates to the asset sending process.
+    */
+    rpc SubscribeSendAssetEventNtfns (SubscribeSendAssetEventNtfnsRequest)
```
Roasbeef, line 107: Zooming out a bit to think about more generically useful calls for future pplication developers, what if we generalized this to a sort of proof distribution listening event? AFAICT, we care about this rn, as we want to be able to write itests against instances where we receive, update, or send proofs. Perhaps we can morph this then into a call centered around notifying applications when we receive or sync new proofs? 

This would then generalize to the case of base universes. Then using this, an app can show a sort of loading bar when a user is starting new proof bootstrap

rpcserver.go
```diff
@@ -1103,3 +1108,72 @@ func (r *rpcServer) SendAsset(ctx context.Context,
...
+						"stream send event: %w", err)
+					errorsChan <- err
+					return
+				}
+			}
+		}
+	}()
+
+	// Handle errors from the goroutine.
+	if err := <-errorsChan; err != nil {
```
Roasbeef, line 1155: This should also select on the main quit statement.

rpcserver.go
```diff
@@ -1103,3 +1108,72 @@ func (r *rpcServer) SendAsset(ctx context.Context,
...
+				}
+
+				err = ntfnStream.Send(rpcEvent)
+				if err != nil {
+					err := fmt.Errorf("failed to RPC "+
+						"stream send event: %w", err)
+					errorsChan <- err
+					return
+				}
+			}
```
Roasbeef, line 1150: Same here re select on the main quit statement. We can probably also just use the context from the `in` here too.

tarofreighter/chain_porter.go
```diff
@@ -1154,6 +1162,21 @@ func (p *ChainPorter) publishSubscriberEvent(event Event) {
...
 var _ chanutils.EventPublisher[Event, bool] = (*ChainPorter)(nil)
 
+// ExecuteSendStateEvent is an event which is sent to the ChainPorter's event
+// subscribers before a state is executed.
+type ExecuteSendStateEvent struct {
+	// timeStamp is the time the event was created.
+	timeStamp time.Time
+
+	// sendState is the state that is about to be executed.
+	SendState SendState
```
Roasbeef, line 1172: Perhaps the main interface should also allow access to this field?

tarofreighter/chain_porter.go
```diff
@@ -577,6 +577,14 @@ func adjustFundedPsbt(pkt *tarogarden.FundedPsbt, anchorInputValue int64) {
 // stateStep attempts to step through the state machine to complete a Taro
 // transfer.
 func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
+	// Notify subscribers that the state machine is about to execute a
+	// state.
+	stateEvent := ExecuteSendStateEvent{
```
Roasbeef, line 582: Ahh ok I see the usage here now: we can emit events from the main state machine to relevant RPC callers.

itest/send_test.go
```diff
@@ -13,6 +16,50 @@ import (
...
+			switch eventTyped := event.Event.(type) {
+			case *tarorpc.SendAssetEvent_ExecuteSendStateEvent:
+				executeSendStateEvent := eventTyped.ExecuteSendStateEvent
+				foundSendSates[executeSendStateEvent.SendState] = struct{}{}
+			}
+		}
+
+		// Ensure that we have found the broadcast state.
+		_, ok := foundSendSates["SendStateBroadcast"]
+		require.True(t.t, ok)
```
Roasbeef, line 60: IIUC failing the test in another goroutine at time doesn't actually work?

itest/send_test.go
```diff
@@ -13,6 +16,50 @@ import (
...
+			// Check for the transaction broadcast state.
+			switch eventTyped := event.Event.(type) {
+			case *tarorpc.SendAssetEvent_ExecuteSendStateEvent:
+				executeSendStateEvent := eventTyped.ExecuteSendStateEvent
+				foundSendSates[executeSendStateEvent.SendState] = struct{}{}
+			}
+		}
+
+		// Ensure that we have found the broadcast state.
+		_, ok := foundSendSates["SendStateBroadcast"]
```
Roasbeef, line 59: See comment about the enum here, then we can just check this directly and not force callers to remember what the string of `SendStateBroadcast` is.

Roasbeef (changes requested):

tarofreighter/chain_porter.go
ffranr: I've gone down the embedded structure route.

tarorpc/taro.proto
ffranr: `ExecuteSendStateEvent` is one type of event that can be streamed. That event type is published before the state machine executes a state.

The `SendAssetEvent` message wrapps all streamed event types.

tarorpc/taro.proto
ffranr: I agree that it makes sense to ad notifications for receiving assets also. You mention proof distribution events, i think it makes sense to include other asset sending/receiving events also (like waiting for tx confirmation). Or should that sort of notification stream be restricted to proofs?

tarofreighter/chain_porter.go
ffranr: The doc is wrong here. Should be `SendState` rather than `sendState`. The field should be accessible on the struct as is. Should I wrap it in a method? I'm not sure what's best practice here.

itest/send_test.go
ffranr: Adding `require.True(t.t, false)` to the goroutine fails the test with:

```
    send_test.go:62: 
        	Error Trace:	/home/user/dev/taro/itest/send_test.go:62
        	            				/home/user/dev/taro/itest/asm_amd64.s:1594
        	Error:      	Should be true
        	Test:       	TestTaroDaemon/basic_send
    --- FAIL: TestTaroDaemon/basic_send (4.24s)
```

but is there some doubt as to whether this test will fail reliably?

rpcserver.go
ffranr: I've added
```
case <-r.quit:
    return
```
to the goroutine's switch statement in the latest commit. Is that the correct change?

rpcserver.go
ffranr: I've added
```
case <-r.quit:
    return
```
to the goroutine's switch statement in the latest commit. Is that the correct change?

rpcserver.go
ffranr: I've change the error handling statement to be:
```
	select {
	// Handle errors from the goroutine.
	case err := <-errorsChan:
		if err != nil {
			return fmt.Errorf("error forwarding send asset event "+
				"notifications to RPC stream send: %w", err)
		}
	case <-r.quit:
		return nil
	}
```

Should the `case <-r.quit` raise an error? I think I've seen that in other places in the code. But I'm not sure if it makes sense to do that here. Does it?

tarofreighter/chain_porter.go
```diff
@@ -81,6 +93,8 @@ type ChainPorter struct {
 
 	exportReqs chan *AssetParcel
 
+	notifications
```
guggero, line 96: If this struct is embedded anyway, I don't really see any reason for it. Even makes it slightly harder to read as compared to declaring `subscribers` and `subscriberMtx` here directly.
Also, if we declare it here directly, we can actually initialize the `subscribers` map in `NewChainPorter()` directly.

tarofreighter/chain_porter.go
```diff
@@ -1089,6 +1112,54 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
...
+
+// RemoveSubscriber removes a subscriber from the set of subscribers that will
+// be notified of any new events that are broadcast.
+func (p *ChainPorter) RemoveSubscriber(
+	subscriber *chanutils.EventReceiver[Event]) error {
+
+	p.subscriberMtx.Lock()
+	defer p.subscriberMtx.Unlock()
+
+	_, ok := p.subscribers[subscriber.ID()]
```
guggero, line 1148: This will panic if `RemoveSubscriber` is called before a subscriber was added. One more reason to initialize the map in `NewChainPorter()`.

tarofreighter/chain_porter.go
```diff
@@ -1089,6 +1112,54 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
...
+
+	subscriber.Stop()
+	delete(p.subscribers, subscriber.ID())
+
+	return nil
+}
+
+// publishSubscriberEvent publishes an event to all subscribers.
+func (p *ChainPorter) publishSubscriberEvent(event Event) {
+	for _, sub := range p.subscribers {
```
guggero, line 1162: Need to hold the lock over the subscribers here as well to avoid it being changed while we send out updates.

tarofreighter/interface.go
```diff
@@ -240,4 +241,11 @@ type Porter interface {
 	// Stop signals that the asset minter should attempt a graceful
 	// shutdown.
 	Stop() error
+
+	chanutils.EventPublisher[Event, bool]
```
guggero, line 245: nit: missing Godoc.

tarorpc/taro.proto
guggero: Yes, maybe it would be worth adding both the `deliverExisting` boolean and `deliverFrom` timestamps from the `RegisterSubscriber` interface here so we could actually implement it in the porter.

rpcserver.go
guggero: In case the return/output stream is closed from the client side we need to catch `case <-ntfnStream.Context().Done():` as well. I think that's what Laolu was referring to.

rpcserver.go
```diff
@@ -1103,3 +1108,80 @@ func (r *rpcServer) SendAsset(ctx context.Context,
...
+	)
+	err := r.cfg.ChainPorter.RegisterSubscriber(eventSubscriber, false, false)
+	if err != nil {
+		return err
+	}
+
+	// This goroutine will read from the ChainPorter event subscription and
+	// forward to the RPC stream.
+	errorsChan := make(chan error, 1)
+	go func() {
```
guggero, line 1129: I don't think we need to do this in a goroutine and then block by waiting on the `errorsChan`. If you take a look at `SubscribeChannelEvents` in `lnd`'s `rpcserver.go` you see that we just have the top level `for` that blocks until the stream is aborted. Then we can also directly return any errors and don't need another channel.

rpcserver.go
```diff
@@ -1103,3 +1108,80 @@ func (r *rpcServer) SendAsset(ctx context.Context,
...
 	}, nil
 }
+
+// SubscribeSendAssetEventNtfns registers a subscription to the event
+// notification stream which relates to the asset sending process.
+func (r *rpcServer) SubscribeSendAssetEventNtfns(
+	in *tarorpc.SubscribeSendAssetEventNtfnsRequest,
+	ntfnStream tarorpc.Taro_SubscribeSendAssetEventNtfnsServer) error {
+
+	eventSubscriber := chanutils.NewEventReceiver[tarofreighter.Event](
```
guggero, line 1118: Should `defer eventSubscriber.Stop()` here to avoid the queues running indefinitely.

rpcserver.go
```diff
@@ -1103,3 +1108,80 @@ func (r *rpcServer) SendAsset(ctx context.Context,
...
+				Timestamp: event.Timestamp().Unix(),
+				SendState: event.SendState.String(),
+			},
+		}
+		return &tarorpc.SendAssetEvent{
+			Event: &eventRpc,
+		}, nil
+	}
+
+	return nil, fmt.Errorf("unknown event type")
```
guggero, line 1186: nit: use `default` case and add the unknown type (`"%T", event`) to the log statement?

rpcserver.go
```diff
@@ -1103,3 +1108,80 @@ func (r *rpcServer) SendAsset(ctx context.Context,
...
+	}
+
+	return nil
+}
+
+// marshallSendAssetEvent maps a ChainPorter event to its RPC counterpart.
+func marshallSendAssetEvent(eventInterface tarofreighter.Event) (*tarorpc.SendAssetEvent, error) {
+	switch event := eventInterface.(type) {
+	case *tarofreighter.ExecuteSendStateEvent:
+		eventRpc := tarorpc.SendAssetEvent_ExecuteSendStateEvent{
```
guggero, line 1175: nit/personal preference: use `&tarorpc.Send...` instead of using `&eventRpc` later. Probably makes no difference at all in the code that's being generated, we just usually write it that way in most of our code.

rpcserver.go
```diff
@@ -1103,3 +1108,80 @@ func (r *rpcServer) SendAsset(ctx context.Context,
...
+			ExecuteSendStateEvent: &tarorpc.ExecuteSendStateEvent{
+				Timestamp: event.Timestamp().Unix(),
+				SendState: event.SendState.String(),
```
guggero, line 1178: What Laolu probably meant in a previous comment about an enum is to not use the `String()` on `SendState` here but instead translate it into a gRPC enum with the same elements. That makes it less fragile when it comes to renaming the states. But it also means more work to marshal/unmarshal between the different enums. So IMO not blocking for the moment (and also I'm going to change a lot of those states in my upcoming PR :sweat_smile: ).

tarofreighter/chain_porter.go
```diff
@@ -583,6 +583,14 @@ func adjustFundedPsbt(pkt *tarogarden.FundedPsbt, anchorInputValue int64) {
 // stateStep attempts to step through the state machine to complete a Taro
 // transfer.
 func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
+	// Notify subscribers that the state machine is about to execute a
+	// state.
+	stateEvent := ExecuteSendStateEvent{
```
guggero, line 588: Same here re using the pointer (`&ExecuteSendStateEvent{`) on struct initialization instead of when using it. Just to be consistent throughout the code base.

tarofreighter/chain_porter.go
```diff
@@ -1160,6 +1168,21 @@ func (p *ChainPorter) publishSubscriberEvent(event Event) {
 // chanutils.EventPublisher interface.
 var _ chanutils.EventPublisher[Event, bool] = (*ChainPorter)(nil)
 
+// ExecuteSendStateEvent is an event which is sent to the ChainPorter's event
+// subscribers before a state is executed.
+type ExecuteSendStateEvent struct {
+	// timestamp is the time the event was created.
+	timestamp time.Time
```
guggero, line 1175: Since this is non-exported, should we add a `NewExecuteSendStateEvent` function that sets the timestamp to now and one `NewExecuteSendStateEventAt` that allows you to set a custom timestamp?

itest/send_test.go
```diff
@@ -13,6 +17,50 @@ import (
 // testBasicSend tests that we can properly send assets back and forth between
 // nodes.
 func testBasicSend(t *harnessTest) {
+	var (
+		ctxb = context.Background()
+		wg   = sync.WaitGroup{}
```
guggero, line 22: nit: can just be `wg sync.WaitGroup` for auto-initialization.

itest/send_test.go
```diff
@@ -13,6 +17,50 @@ import (
...
+			// Check for the transaction broadcast state.
+			switch eventTyped := event.Event.(type) {
+			case *tarorpc.SendAssetEvent_ExecuteSendStateEvent:
+				executeSendStateEvent := eventTyped.ExecuteSendStateEvent
+				foundSendSates[executeSendStateEvent.SendState] = struct{}{}
+			}
+		}
+
+		// Ensure that we have found the broadcast state.
+		_, ok := foundSendSates[tarofreighter.SendStateBroadcast.String()]
```
guggero, line 60: nit: could use `require.Contains()` for a nicer error message on failure.

itest/send_test.go
guggero: I think it's okay to fail within a goroutine as long as you make sure it actually runs. So I think we need to add a `wg.Wait()` somewhere.

guggero (commented):
Nice work. Looks pretty close, only a couple of code style related comments.

rpcserver.go
ffranr: Ah good point. Thank you both.

tarofreighter/chain_porter.go
ffranr: I don't mind either way. Went for embedded following this comment: https://github.com/lightninglabs/taro/pull/235#discussion_r1080714456

I will switch back to directly adding the fields so that I can initialise the `subscribers` map in `NewChainPorter()`.

rpcserver.go
ffranr: You're right of course! Thanks for pointing that out.

itest/send_test.go
ffranr: In the latest commits, this has been modified to setting and checking a bool flag. The flag is true if the target state has been found.

rpcserver.go
ffranr: Ok, I'll leave as is for now.

tarofreighter/chain_porter.go
ffranr: I'll add `NewExecuteSendStateEvent`.

I don't have a use case for `NewExecuteSendStateEventAt` at the moment. I'll leave it out for that reason, unless you think I should add it anyway for convention or something like that, please let me know if that's the case.

itest/send_test.go
ffranr: `wg.Wait()` added :+1:

tarorpc/taro.proto
ffranr: I didn't mean to add backlog support in this PR. I figured I would add those fields in a future PR. Would that work for you guys?

rpcserver.go
ffranr: The `errorsChan` and related setup has been removed in the latest commits.

tarofreighter/chain_porter.go
```diff
@@ -1089,6 +1112,80 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
...
+	// SendState is the state that is about to be executed.
+	SendState SendState
+}
+
+// Timestamp returns the timestamp of the event.
+func (e *ExecuteSendStateEvent) Timestamp() time.Time {
+	return e.timestamp
+}
+
+func NewExecuteSendStateEvent(state SendState) *ExecuteSendStateEvent {
```
guggero, line 1182: nit: Godoc.

rpcserver.go
```diff
@@ -1103,3 +1108,75 @@ func (r *rpcServer) SendAsset(ctx context.Context,
...
+
+		// Handle the case where the RPC server is shutting down.
+		case <-r.quit:
+			return nil
+		}
+	}
+}
+
+// marshallSendAssetEvent maps a ChainPorter event to its RPC counterpart.
+func marshallSendAssetEvent(eventInterface tarofreighter.Event) (*tarorpc.SendAssetEvent, error) {
```
guggero, line 1166: nit: line length.

guggero (approved):
LGTM :tada:

itest/send_test.go
```diff
@@ -13,6 +17,61 @@ import (
...
+				break
+			}
+
+			// If err is not EOF, then we expect it to be nil.
+			require.NoError(t.t, err)
+
+			// Check for the transaction broadcast state.
+			switch eventTyped := event.Event.(type) {
+			case *tarorpc.SendAssetEvent_ExecuteSendStateEvent:
+				executeSendStateEvent := eventTyped.ExecuteSendStateEvent
```
jharveyb, line 65: Optional but could add logging of the timestamp + state here, something like:

```
timestamp := time.UnixMicro(executeSendStateEvent.Timestamp)
t.Logf("timestamp: %v", timestamp.Format(time.RFC3339Nano))
t.Logf("state: %v", executeSendStateEvent.SendState)
```

rpcserver.go
```diff
@@ -1103,3 +1108,75 @@ func (r *rpcServer) SendAsset(ctx context.Context,
...
+	switch event := eventInterface.(type) {
+	case *tarofreighter.ExecuteSendStateEvent:
+		eventRpc := &tarorpc.SendAssetEvent_ExecuteSendStateEvent{
+			ExecuteSendStateEvent: &tarorpc.ExecuteSendStateEvent{
+				Timestamp: event.Timestamp().Unix(),
```
jharveyb, line 1171: IMO this should be `event.Timestamp().UnixMicro()` - right now this steps drops us to second-level precision, so all the timestamps show as the same time. See my other comment on how to log the timestamps at full precision.

tarofreighter/chain_porter.go
```diff
@@ -560,6 +578,11 @@ func adjustFundedPsbt(pkt *tarogarden.FundedPsbt, anchorInputValue int64) {
 // stateStep attempts to step through the state machine to complete a Taro
 // transfer.
 func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
+	// Notify subscribers that the state machine is about to execute a
+	// state.
+	stateEvent := NewExecuteSendStateEvent(currentPkg.SendState)
```
jharveyb, line 583: 🔍  💯

tarorpc/taro.proto
jharveyb: Sounds like in the future we'd use this nesting to more easily extend to other event types from other subsystems?

Was a bit confused re: the extra nesting.

tarorpc/taro.yaml
```diff
@@ -58,5 +58,9 @@ http:
       post: "/v1/taro/send"
       body: "*"
 
+    - selector: tarorpc.Taro.SubscribeSendAssetEventNtfns
```
jharveyb, line 61: Not clear to me what happens when you ping this endpoint - do you get multiple events, or some handle to then stream events from, or something else?

Not blocking just curious.

jharveyb (commented):
Cool PR! Neat have this added without much change to the send state machine.

One nit re: timestamp conversion, otherwise LGTM.

tarorpc/taro.proto
dstadulis: Was starting to write a document to enumerate the entities within the daemon and the associated states but then I had a thought.  Might we want to, at least eventually, use this event-information conduit to ferry external events back into the daemon?  E.g. the daemon being notified about the state of a channel or HTLC from whatever the payment channel daemon (lnd) is doing.

tarorpc/taro.yaml
guggero: We use `lnd`'s WebSocket proxy that allows you to turn streaming RPCs into WebSockets: https://github.com/lightninglabs/taro/blob/main/server.go#L337

See https://github.com/lightningnetwork/lnd/blob/master/docs/rest/websockets.md for example code on how to use such endpoints.

tarorpc/taro.proto
ffranr: > we'd use this nesting to more easily extend to other event types

Yes, that's what I was going for.

rpcserver.go
ffranr: Should I format as follows:
```
func marshallSendAssetEvent(eventInterface tarofreighter.Event) (
	*tarorpc.SendAssetEvent, error) {

	switch event := eventInterface.(type) {
```	
I'm trying to avoid the dangling `(`.

rpcserver.go
ffranr: Good point, thanks! I'll switch to using `UnixMicro`.

rpcserver.go
guggero: In this case I'd probably break before the argument and then again before the `error` (if necessary).

ffranr: Good to know, thanks.

tarorpc/taro.yaml
ffranr: > See https://github.com/lightningnetwork/lnd/blob/master/docs/rest/websockets.md for example code on how to use such endpoints.

Nice docs, thanks for the link.

tarorpc/taro.proto
ffranr: Currently, in some instances, external events do reach the taro daemon via notification channels. Here's an example where we receive a notification once an on-chain transaction has confirmed:
https://github.com/ffranr/taro/blob/d73b1f85c2ffc7f3b5e08afb2f40452e84788be4/tarofreighter/chain_porter.go#L269-L269

> Might we want to, at least eventually, use this event-information conduit to ferry external events back into the daemon?

I would suggest that the taro daemon should receive notification events from lnd in a similar fashion to how on-chain related events are received.

ffranr: Currently, in some instances, external events do reach the taro daemon via notification channels. Here's an example where we receive a notification once an on-chain transaction has confirmed:
https://github.com/ffranr/taro/blob/d73b1f85c2ffc7f3b5e08afb2f40452e84788be4/tarofreighter/chain_porter.go#L269-L269

> Might we want to, at least eventually, use this event-information conduit to ferry external events back into the daemon?

I would suggest that the taro daemon should receive notification events from lnd in a similar fashion to how on-chain related events are received.

itest/send_test.go
```diff
@@ -107,4 +134,65 @@ func testBasicSend(t *harnessTest) {
...
+	err = eventNtfns.CloseSend()
+	require.NoError(t.t, err)
+
+	wg.Wait()
+}
+
+// assertRecvSendSateExecEvent asserts that the given send state execution event
+// notification was received. This function will block until the event is
+// received or the event stream is closed.
+func assertRecvSendSateExecEvent(
```
jharveyb, line 148: Indeed easier to read + change later on and reuse in other tests.

itest/send_test.go
```diff
@@ -107,4 +134,65 @@ func testBasicSend(t *harnessTest) {
...
+
+			break
+		}
+
+		// If err is not EOF, then we expect it to be nil.
+		require.NoError(t.t, err)
+
+		// Check for the transaction broadcast state.
+		switch eventTyped := event.Event.(type) {
+		case *tarorpc.SendAssetEvent_ExecuteSendStateEvent:
```
jharveyb, line 179: Seems like we may want a default case here that logs the unknown type? Or maybe fails the test.

jharveyb (approved):
Looks good 🚀

itest/send_test.go
ffranr: Code at this point will certainly encounter events of types other than `ExecuteSendStateEvent`. I would like it to ignore those types. The goal is to identify a specific type rather than make an assertion on all event types that this code might encounter.

Roasbeef (approved):
LGTM 🕸️

PR #243, opened 2023-01-25
psbt script spend [1/3]: preparation

guggero:
This PR contains a bunch of small fixes or preparations towards the big refactor for supporting PSBT based script spends.


rpcserver.go
```diff
@@ -1192,3 +1298,51 @@ func marshallSendAssetEvent(
 		return nil, fmt.Errorf("unknown event type: %T", eventInterface)
 	}
 }
+
+// unmarshalScriptKey parses the RPC script key into the native counterpart.
```
jharveyb, line 1302: Double checking; for these keys, we want to accept a pubkey AND an optional key descriptor. For the descriptor, the tweak can be nil since the default is also nil.

For the descriptor, we want to accept a pubkey AND an optional key locator.

Is there any key family sanity checking we should have here?

Also, if the above is correct, maybe `unmarshalScriptKey` should parse the pubkey before the key descriptor so the priority is more clear.

jharveyb (approved):
Awesome PR! Great to have progress on accepting external keys.

Is there a particular reason we reject setting only one of the two address keys?

asset/asset.go
```diff
@@ -403,6 +403,19 @@ type GroupKey struct {
 	Sig schnorr.Signature
 }
 
+// NewGroupKey is a helper function that creates a new group key wrapper from a
+// group public key. If the public key is nil then a nil group key struct is
+// returned.
+func NewGroupKey(groupPubKey *btcec.PublicKey) *GroupKey {
```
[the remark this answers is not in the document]
guggero: Hmm, looks like this isn't used in any of the three parts... Probably from an intermediate refactor. Dropped the commit.

rpcserver.go
guggero: Good questions, I guess the code could use a few explanatory comments (which I added).

On the script key level, we allow the descriptor to be nil, which would then indicate "this key is not derived by the wallet that's backing the Taro daemon". So a fully externally derived key.

The tweak being `nil` indicates the usage of BIP 86 as opposed to having a script tree that's being committed to.

On the key descriptor level, we inherit the rules from `lnd`'s APIs: A public key is generally enough to find the corresponding private key (e.g. when signing). But when available the locator information should still be provided, but it's not strictly mandatory.

rpcserver.go
```diff
@@ -1251,3 +1298,62 @@ func marshallSendAssetEvent(
...
+func unmarshalScriptKey(rpcKey *tarorpc.ScriptKey) (asset.ScriptKey, error) {
+	var (
+		scriptKey asset.ScriptKey
+		err       error
+	)
+	scriptKey.PubKey, err = schnorr.ParsePubKey(rpcKey.PubKey)
```
ffranr, line 1308: nit: Consider adding a comment above this line, something like: "Attempting to parse public key as a schnorr public key."

rpcserver.go
```diff
@@ -1251,3 +1298,62 @@ func marshallSendAssetEvent(
...
+	if err != nil {
+		return scriptKey, err
+	}
+
+	// The key descriptor is optional for script keys that are completely
+	// independent of the backing wallet.
+	if rpcKey.KeyDesc != nil {
+		keyDesc, err := unmarshalKeyDescriptor(rpcKey.KeyDesc)
+		if err != nil {
+			return scriptKey, err
```
ffranr, line 1318: nit: `return nil, err` (along with function signature changes, etc.) would be clearer here, IMO.

rpcserver.go
```diff
@@ -1251,3 +1298,62 @@ func marshallSendAssetEvent(
...
+
+	var (
+		desc keychain.KeyDescriptor
+		err  error
+	)
+
+	// The public key of a key descriptor is mandatory. It is enough to
+	// locate the corresponding private key in the backing wallet. But to
+	// speed things up (and for additional context), the locator should
+	// still be provided if available.
```
ffranr, line 1345: :+1:

ffranr (approved):
I appreciate the comments and the nice commit structure! Nice work.

rpcserver.go
```diff
@@ -355,59 +356,117 @@ func (r *rpcServer) fetchRpcAssets(ctx context.Context) (
...
+				AssetId:     prevID.ID[:],
+				ScriptKey:   prevID.ScriptKey[:],
 			}
+
+			var rpcSplitCommitment *tarorpc.SplitCommitment
+			if witness.SplitCommitment != nil {
+				rootAsset, err := marshalAsset(
+					&witness.SplitCommitment.RootAsset,
+					true,
+				)
```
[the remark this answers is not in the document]
guggero: You mean to cause a stack overflow from infinite recursion? I don't think so, since this data is only coming from the DB so far.

tarorpc/taro.proto
```diff
@@ -2,7 +2,7 @@ syntax = "proto3";
 
 package tarorpc;
 
-option go_package = "github.com/lightninglabs/tarorpc";
+option go_package = "github.com/lightninglabs/taro/tarorpc";
```
Roasbeef, line 5: 👍

Roasbeef (commented):
posthumous LGTM 🐙

Issue #24, opened 2022-04-19
bip-taro: implement provable burning and verification


Roasbeef:
One important feature is the ability to _burn_ assets, and also prove to third parties that those assets have been burnt. 

This addition is defined currently in this PR to my BIP repo: https://github.com/Roasbeef/bips/pull/21. 

At a high level, we'll use a series of non-inclusion proofs, both within the tapscript level and also across the other outputs in the transaction. 

This is related to https://github.com/lightninglabs/taro/issues/20 as the set of non-inclusion proofs outlined there will be used to ensure the asset commitment doesn't persist. 

# Steps To Completion
- [ ] Implement a routine that burns an asset by removing the commitment from the asset tree.
- [ ] Implement a routine related to the above that generates a non-inclusion proof on the taproot-asset level proving the asset is no longer committed to. 
- [ ] Use the above, along side #20 to implement the final component which is to prove that no other commitments are valid in the transaction. Assuming widespread BIP 86 usage is a safe assumption.
- [ ] Update taproot-asset BIPs to reflect protocol requirements for provable burning.

jharveyb:
After some offline discussion with @guggero, I think we've narrowed down the options here. Generally, we're aware of two approaches:

- Burn via proof-of-omission. Instead if re-anchoring an asset into the next anchor output, or spending it, omit the asset from the tap tree. This allows for a proof-of-exclusion to serve as a proof-of-burn.

- - Drawback: May collide with future protocol changes to allow for asset 'teleporting' via ANYPREVOUT-like mechanics, that could replace asset reanchoring with new signatures. I think we could mitigate that possible collision by requiring a more strict asset state before omission, like using a specific sighash.


- Burn via a provaly unspendable script key, followed by omission. The proof for the next spend of the asset includes a field in the associated proof that provides the data needed to verify that the current script key is unspendable / unsatisfiable. The asset is then never moved again, and therefore omitted from future spends of the anchor output.

- - This option ***cannot*** use a shared NUMS key, since that would cause burnt assets to collide in the tap tree if transfers are batched in a pocket universe, or other transfer batching scheme (the tree requires tap asset script keys to be unique).

[...]

PR #244, opened 2023-01-25
psbt script spend [2/3]: the big refactor

guggero:
Depends on https://github.com/lightninglabs/taro/pull/243.

Fixes https://github.com/lightninglabs/taro/issues/122.

This PR contains all the refactor commits required to support PSBT based script spends.


taroscript/send_test.go
```diff
@@ -301,201 +312,100 @@ func createGenesisProof(t *testing.T, state *spendData) {
...
+		Amount:                  addr.Amount,
+		ScriptKey:               asset.NewScriptKey(&addr.ScriptKey),
+		AnchorOutputIndex:       receiverExternalIdx,
+		AnchorOutputInternalKey: &addr.InternalKey,
+	}}
+
+	vPacket := &taropsbt.VPacket{
+		Input:       input,
+		Outputs:     outputs,
+		ChainParams: &address.RegressionNetTaro,
```
jharveyb, line 340: The test addresses are all for mainnet, so we may want to add a network check to `taropsbt.FromAddress()` and elsewhere.

taroscript/send.go
```diff
@@ -290,154 +164,181 @@ func IsValidInput(input *commitment.TaroCommitment,
...
-	// Update the input of the virtual TX, generate a witness,
-	// and attach it to the copy of the new Asset.
-	validatedAsset := updatedDelta.NewAsset.Copy()
-	prevWitnessCount := len(updatedDelta.NewAsset.PrevWitnesses)
-
+	// For each input asset leaf, we need to produce a witness. Update the
+	// input of the virtual TX, generate a witness, and attach it to the
+	// copy of the new Asset.
+	//
+	// TODO(guggero): I think this is wrong... We shouldn't look at
```
jharveyb, line 316: ++ This needs a closer look.

taroscript/send.go
```diff
@@ -446,202 +347,301 @@ func CompleteAssetSpend(internalKey btcec.PublicKey, delta SpendDelta,
...
 
 	// Update each split asset to store the root asset with the witness
 	// attached, so the receiver can verify inclusion of the root asset.
-	for key := range updatedDelta.SplitCommitment.SplitAssets {
-		updatedDelta.SplitCommitment.SplitAssets[key].Asset.
-			PrevWitnesses[0].SplitCommitment.
-			RootAsset = *validatedAsset.Copy()
-	}
+	for idx := range outputs {
+		if outputs[idx].IsChange || outputs[idx].Interactive {
```
jharveyb, line 379: `splitAssets` should also include the split root / change output.

Related, I think `Interactive` and `!IsSplit` are mutually exclusive? At least while we only have one input.

taroscript/send_test.go
```diff
@@ -301,201 +312,100 @@ func createGenesisProof(t *testing.T, state *spendData) {
...
+	require.Equal(t, raw.Amount, signed.Amount)
+	require.Equal(t, raw.LockTime, signed.LockTime)
+	require.Equal(t, raw.RelativeLockTime, signed.RelativeLockTime)
+	require.Equal(t, len(raw.PrevWitnesses), len(signed.PrevWitnesses))
+
+	// The signed asset should have a single signature in the witness stack.
+	require.NotNil(t, signed.PrevWitnesses[0].TxWitness)
+	require.Len(t, signed.PrevWitnesses[0].TxWitness, 1)
+	require.Len(t, signed.PrevWitnesses[0].TxWitness[0], 64)
 	if split {
```
jharveyb, line 402: Can also check for unspendability here.

taroscript/send_test.go
```diff
@@ -877,59 +772,181 @@ func TestPrepareAssetCompleteSpend(t *testing.T) {
...
-// TestCompleteAssetSpend tests edge cases around signing a witness for
-// an asset transfer and validating that transfer with the Taro VM.
-func TestCompleteAssetSpend(t *testing.T) {
+		unvalidatedAsset := pkt.Outputs[0].Asset.Copy()
+		err = taroscript.SignVirtualTransaction(
+			pkt.Input, pkt.Outputs, state.signer, state.validator,
+		)
+		require.NoError(t, err)
+
+		checkSignedAsset(
```
jharveyb, line 835: Should be `true` since checking split root?

taroscript/send_test.go
```diff
@@ -877,59 +772,181 @@ func TestPrepareAssetCompleteSpend(t *testing.T) {
...
+		err := taroscript.PrepareOutputAssets(pkt.Input, pkt.Outputs)
+		require.NoError(t, err)
+
+		unvalidatedAsset := pkt.Outputs[0].Asset.Copy()
+		err = taroscript.SignVirtualTransaction(
+			pkt.Input, pkt.Outputs, state.signer, state.validator,
+		)
+		require.NoError(t, err)
+
+		checkSignedAsset(
```
jharveyb, line 936: Same note here re: should be `true` for check.

taroscript/send.go
```diff
@@ -446,202 +347,301 @@ func CompleteAssetSpend(internalKey btcec.PublicKey, delta SpendDelta,
...
-	}
-	receiverCommitment, ok := commitments[receiverStateKey]
-	if !ok {
-		return ErrMissingTaroCommitment
-	}
+		// The commitment must be defined at this point.
+		//
+		// TODO(guggero): Merge multiple Taro level commitments that use
+		// the same external output index.
+		if outputCommitment == nil {
```
jharveyb, line 614: Nit: Can be moved earlier to right after we define `outputCommitment`.

tarofreighter/parcel.go
```diff
@@ -481,88 +215,86 @@ func (s *sendPackage) createProofs() (spendProofs, error) {
 
 	// If we require a split, then we'll need to prove exclusion for both
 	// parties.
-	if isSplit {
-		senderParams.NewAsset = &s.SendDelta.NewAsset
+	if s.VirtualPacket.Input.IsSplit {
```
jharveyb, line 218: These two cases are mostly the same now, could dedupe a few lines.

taroscript/send.go
```diff
@@ -446,202 +347,301 @@ func CompleteAssetSpend(internalKey btcec.PublicKey, delta SpendDelta,
...
 	}
 
-	commitments[receiverStateKey] = *receiverTaroCommitment
+	for i := range outputs {
+		out := outputs[i]
+		psbtOut := &spendPkt.Outputs[out.AnchorOutputIndex]
+		psbtOut.TaprootInternalKey = schnorr.SerializePubKey(
+			out.AnchorOutputInternalKey,
+		)
+		if out.AnchorOutputBip32Derivation != nil {
```
jharveyb, line 563: IIUC these can be nil from non-local keys given in an address, is there another case where that would be true?

jharveyb (commented):
Looking great so far! Want to better understand some bits but feels close.

taroscript/send_test.go
jharveyb: Looks like these checks are partially added in the follow-up PR, but they wouldn't apply in the tests since we create `VPackets` directly.

I think this could be replaced with `FromAddress` for the positive test cases.

taroscript/send.go
jharveyb: Confirmed that the VM expects the split asset representing the root to be present in this list:

https://github.com/lightninglabs/taro/blob/82daf05d4fd50d14e47010365c14a1416e806350/vm/vm.go#L152

And we pass the full list in tests:

https://github.com/lightninglabs/taro/blob/82daf05d4fd50d14e47010365c14a1416e806350/vm/vm_test.go#L266

IIUC the root split asset is the only place storing the merkle proof of inclusion for the root asset, so we should still be verifying it.

taropsbt/interface.go
```diff
@@ -0,0 +1,170 @@
...
+	Anchor Anchor
+
+	// TapscriptSibling is the tapscript sibling of this asset. This will
+	// usually be blank.
+	TapscriptSibling []byte
+
+	// asset is the full instance of the asset being spent. It is not
+	// exported because the assets script key must be encoded in the PSBT
+	// input struct for the signing to work correctly.
+	asset *asset.Asset
```
Roasbeef, line 105: May have been lost in some of the comments in the original mega PR but: don't we also need an SMT reveal here along with the tapscript reveal? This then permits verification that the asset actually lives within the anchor above. Or is the assumption that it'll be fetched out of band with a Universe API?

Roasbeef: w/ the above, we can also potentially break out portions of the `proof.Proof` struct to reside here as well.

taropsbt/interface.go
```diff
@@ -109,6 +114,51 @@ func (i *VInput) Asset() *asset.Asset {
 	return i.asset
 }
 
+// serializeScriptKey serializes the input asset's script key as the PSBT
+// derivation information on the virtual input.
+func (i *VInput) serializeScriptKey(key asset.ScriptKey, coinType uint32) {
+	if key.TweakedScriptKey == nil {
```
Roasbeef, line 120: ?

taropsbt/interface.go
```diff
@@ -0,0 +1,312 @@
...
+			uint32(keyDesc.Family) +
+				uint32(hdkeychain.HardenedKeyStart),
+			0,
+			keyDesc.Index,
+		},
+	}
+
+	return bip32Derivation, &psbt.TaprootBip32Derivation{
+		XOnlyPubKey:          bip32Derivation.PubKey[1:],
+		MasterKeyFingerprint: bip32Derivation.MasterKeyFingerprint,
```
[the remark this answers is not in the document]
Roasbeef: IIRC it's used by some hardware wallets to do additional verification.

taroscript/send.go
```diff
@@ -290,154 +164,181 @@ func IsValidInput(input *commitment.TaroCommitment,
...
-	// verify that the desired split is possible. We can later regenerate a
-	// split with the final output indexes.
-	if updatedDelta.Locators == nil {
-		updatedDelta.Locators = CreateDummyLocators(
-			[][32]byte{senderStateKey, receiverStateKey},
-		)
+	// This should be caught way earlier but just to make sure that we never
+	// overflow when converting the input amount to int64 we check this
+	// again.
+	inputAsset := input.Asset()
```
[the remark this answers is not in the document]
Roasbeef: Alternatively, we can modify the API of `VInput` s.t you must pass in the asset via some method, where it then gets copied.

taroscript/send.go
```diff
@@ -0,0 +1,164 @@
...
+	if inputAsset.Amount > math.MaxInt64 {
+		return fmt.Errorf("amount int64 overflow")
+	}
+
+	// A collectible cannot be split into individual pieces. So there can
+	// only be a tombstone and a recipient output.
+	if inputAsset.Type == asset.Collectible && len(outputs) > 2 {
```
Roasbeef, line 185: I think we only need to enforce this for the case where `Interactive` is false?

taroscript/send.go
```diff
@@ -290,154 +164,181 @@ func IsValidInput(input *commitment.TaroCommitment,
...
+
+	// If we have an interactive full value send, we don't need a tomb stone
+	// at all.
+	inputIDCopy := input.PrevID
+	if interactiveFullValueSend(input, outputs) {
+		// We'll now create a new copy of the old asset, swapping out
+		// the script key. We blank out the tweaked key information as
+		// this is now an external asset.
+		outputs[1].Asset = inputAsset.Copy()
+		outputs[1].Asset.ScriptKey = outputs[1].ScriptKey
```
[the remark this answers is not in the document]
Roasbeef: IIUC, that's done as this is the top level `ScriptKey` struct that contains the raw key and tweaked key (tho in this case only the tweaked key matters).

taroscript/send.go
```diff
@@ -0,0 +1,164 @@
...
+
+	return nil
+}
+
+// interactiveFullValueSend returns true if the given outputs spend the input
+// fully and interactively.
+func interactiveFullValueSend(input *taropsbt.VInput,
+	outputs []*taropsbt.VOutput) bool {
+
+	return len(outputs) == 2 &&
```
Roasbeef, line 644: Shouldn't this be `outputs == 1`? So it would be the same as a consolidation transaction.

taroscript/send.go
Roasbeef: Copy pasting my comment from the other PR here: 

Yeah so we need to sign the new asset (which also anchors the split), along with the other leaves that occupied that same prev anchor output.

Here we're instead, re-signing all the old inputs for a merge (iiuc) which isn't correct.

In addition to just the Vinput, we also need the TaroCommitment for that, as for now we need to re-sign all the other inputs. FWIW, we only support spending a single input for now, so things are restricted to just what was in that sole tree.

---

Ah ok this was lifted from the old code. It's actually partially untested, since we only support a single input. This would be for the "merge" case, where we need to sign more than one taro input.

Once we lift the merge restriction (ends up needing a diff proof file creation), then this is equivalent to signing multiple Bitcoin inputs when you need to consolidate change to make a spend.

taroscript/send.go
```diff
@@ -153,6 +154,107 @@ func PrepareOutputAssets(input *taropsbt.VInput,
...
+// full asset in case of an interactive full amount send) by creating a
+// signature over the asset transfer, verifying the transfer with the Taro VM,
+// and attaching that signature to the new Asset.
+func SignVirtualTransaction(input *taropsbt.VInput, outputs []*taropsbt.VOutput,
+	signer Signer, validator TxValidator) error {
+
+	prevAssets := commitment.InputSet{
+		input.PrevID: input.Asset(),
+	}
+	newAsset := outputs[1].Asset
```
Roasbeef, line 301: Seems to assume a strict ordering re the outputs here?

taroscript/send.go
```diff
@@ -354,6 +359,132 @@ func CreateOutputCommitments(inputCommitment *commitment.TaroCommitment,
...
+
+	return spendPkt, nil
+}
+
+// UpdateTaprootOutputKeys updates a PSBT with outputs embedding TaroCommitments
+// involved in an asset send. The sender must attach the Bitcoin input holding
+// the corresponding Taro input asset to this PSBT before finalizing the TX.
+// Locators MUST be checked beforehand.
+func UpdateTaprootOutputKeys(btcPacket *psbt.Packet,
+	outputs []*taropsbt.VOutput,
```
Roasbeef, line 585: Passes in a Bitcoin PSBT but still operates with taro virtual outputs?

taroscript/send.go
```diff
@@ -354,6 +359,132 @@ func CreateOutputCommitments(inputCommitment *commitment.TaroCommitment,
...
+		outputCommitment := outputCommitments[idx]
+
+		// The external output index cannot be out of bounds of the
+		// actual TX outputs. This should be checked earlier and is just
+		// a final safeguard here.
+		if vOut.AnchorOutputIndex >= uint32(len(btcPacket.Outputs)) {
+			return ErrInvalidOutputIndexes
+		}
+
+		btcOut := btcPacket.Outputs[vOut.AnchorOutputIndex]
```
Roasbeef, line 602: What about instances wherein you have multiple virtual outputs in a single Bitcoin output? Example would be: sending two collectibles to someone in the same transaction.

taroscript/send.go
```diff
@@ -354,6 +359,132 @@ func CreateOutputCommitments(inputCommitment *commitment.TaroCommitment,
...
+		}
+
+		// Create the scripts corresponding to the receiver's
+		// TaroCommitment.
+		//
+		// NOTE: We currently default to the Taro commitment having no
+		// sibling in the Tapscript tree. Any sibling would need to be
+		// checked to verify that it is not also a Taro commitment.
+		script, err := PayToAddrScript(
+			*internalKey, nil, *outputCommitment,
```
Roasbeef, line 626: I think we also already pass the sibling along side the outputs? So then can just pass that here, knowing it'll be `nil` in most places we care about.

tarofreighter/wallet.go
```diff
@@ -0,0 +1,241 @@
...
+	// something has gone wrong with the DB.
+	internalKey := assetInput.InternalKey
+	if internalKey.Family != taroscript.TaroKeyFamily {
+		return nil, nil, fmt.Errorf("invalid internal key family "+
+			"for selected input: %v %v", internalKey.Family,
+			internalKey.Index)
+	}
+
+	inBip32Derivation, inTrBip32Derivation :=
+		taropsbt.Bip32DerivationFromKeyDesc(
```
Roasbeef, line 165: Weird line folding...but doesn't seem easily avoidable 🤷🏾‍♂️

tarofreighter/wallet.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	// At this point, we have a valid "coin" to spend in the commitment, so
+	// we'll add the relevant information to the virtual TX's input.
+	//
+	// TODO(roasbeef): still need to add family key to PrevID.
+	pkt := &taropsbt.VPacket{
+		Outputs: make([]*taropsbt.VOutput, 2),
+	}
+	pkt.Input = &taropsbt.VInput{
```
Roasbeef, line 179: 😎

tarofreighter/wallet.go
```diff
@@ -0,0 +1,241 @@
...
+			InternalKey:       internalKey.PubKey,
+			MerkleRoot:        anchorMerkleRoot,
+			Bip32Derivation:   inBip32Derivation,
+			TrBip32Derivation: inTrBip32Derivation,
+		},
+		PInput: psbt.PInput{
+			SighashType: txscript.SigHashDefault,
+		},
+	}
+	pkt.SetInputAsset(assetInput.Asset)
```
Roasbeef, line 199: Re prior comment, can then make this always copy internally.

tarofreighter/wallet.go
```diff
@@ -0,0 +1,241 @@
...
+		Amount:            receiverAddr.Amount,
+		IsChange:          false,
+		Interactive:       false,
+		AnchorOutputIndex: 1,
+		ScriptKey:         asset.NewScriptKey(&receiverAddr.ScriptKey),
+	}
+
+	// If we are sending the full value of the input asset, or sending a
+	// collectible, we will need to create a split with un-spendable change.
+	pkt.Outputs[0] = &taropsbt.VOutput{
```
Roasbeef: So then assumption is that this isn't used for the interactive flow?

Roasbeef (changes requested):

taroscript/send.go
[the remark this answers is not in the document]
guggero: Yeah, on the output level we only really need the full output key. But if we do have the derivation information (because the key is local to our node), we want to serialize it in the PSBT. But it's optional, really.

taroscript/send_test.go
guggero: Ah good point about the mismatch. I fixed it in the tests. Though I don't know what value to check against in `FromAddress()`? Also, `FromAddress()` does slightly less than this does, so it would be another partial refactor which I'd like to avoid at this point.

taroscript/send.go
guggero: Hmm, I'm not sure about this. We don't have that split asset anymore here, it's discarded in `PrepareOutputAsset`, as we only need the root asset (which is a copy of the split asset at the root locator, with the tx witness and split commitment root set). It's true that the VM will validate it if supplied. But it doesn't require it, and all the proofs seem to work without it (in the old code we only ever used that split asset in this verify call after signing, it wasn't used anywhere else afterwards).
So because I think it's easier to understand what happens, and because the VM doesn't require it, it's easier to just drop it?

If there was crucial information missing, then the proofs wouldn't validate, right?

taroscript/send_test.go
guggero: Yep, makes sense, added.

taroscript/send_test.go
guggero: Yes, correct, fixed.

taroscript/send.go
guggero: Yes, for non-local keys, mostly. Or for malformed/incomplete packets, since this could come from an RPC. So we'd rather be cautious and fail with a "missing derivation information" error later on.

taropsbt/interface.go
guggero: Not sure what the question refers to. But for non-local keys, we might not have the information about how the script key was created (meaning we only know the Taproot output key, not the internal key and tweak).

taroscript/send.go
guggero: Yes, this basically maps the virtual outputs onto BTC level outputs and creates the BTC level pkScripts. We just need to know what BTC level output index (=`AnchorOutputIndex`) each virtual output maps to.

guggero: This will later on be used to merge multiple virtual transactions (e.g. for multiple asset sends within the same anchor TX) outputs into the correct BTC level outputs.

taroscript/send.go
guggero: Yes, this will be done at a later stage (meaning, another PR). See TODO about merging multiple Taro level commitments into a single one.

taroscript/send.go
guggero: No, we only pass around the taproot sibling for the _input_ we spend. So this would need to be defined on the BTC transaction level. I'll take a look at that when I do the PR that allows external signing of the BTC PSBT stuff (which will also be the PR that makes sure everything works with those Taproot siblings).

taropsbt/interface.go
guggero: Good point, I did miss that part in the previous PR. Since the proofs shouldn't be too large (as compared to the full input commitment), I'll try adding it. And if that will simplify the proof creation, even better!

tarofreighter/wallet.go
guggero: Yes. I'm not yet sure whether we should just introduce a convention where the _second_ (index 1) output is **always** the recipient and the _first_ (index 0) virtual output is always the change output, even if there is not going to be any change or tombstone in case of an interactive send. Or whether we should always try to dynamically find out which output is for the recipient. I think the latter will be a bit more complicated to follow in the send code, while the former makes it more difficult to use the API because you need to know this convention.

taroscript/send.go
guggero: I added a TODO in the hopefully right place.

taroscript/send.go
guggero: Yes, updated.

taroscript/send.go
guggero: I fixed the whole logic around the number of outputs. If it's a full value interactive send, there now can be just one output.

taroscript/send.go
guggero: Yes, was related to the "convention" I was initially going with. But now the logic only looks at the `IsChange` and `IsInteractive` flags and doesn't assume a specific location for things. In this case the first output will always be the one with the signature, since either it's the change output (or tombstone) in case of a split, or it's the new asset in case of an interactive full value send.

tarofreighter/wallet.go
guggero: Okay, I went with the second option, we now always determine things dynamically, there's no convention other than: **If** there is a change output, it must be at position 0.

taroscript/send.go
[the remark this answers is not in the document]
jharveyb: Ok, so we set `outputs[0].ScriptKey` earlier when creating the VPacket, and are now moving it to the receiver asset? And that key likely has no tweak information.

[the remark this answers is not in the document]
guggero: Yes, correct. Because at the point where you create the virtual packet, there is no asset to set the script key on yet. That's why it's a separate field to store it until we've created the output asset.

taroscript/send.go
jharveyb: Yeah, I think the lack of failures is a related issue where, from the VM, we can't know that we verified all leafs of of the split tree, only that there is non-inflation.

AFAICT we don't use the proof-of-inclusion for the root asset later, and running `validateSplit()` with the root locator doesn't add any validation for the root besides that inclusion proof.

If we are dropping the root locator then `validateSplit()` needs a change to enforce non-zero amounts for splits, but that can be a separate PR.

taroscript/send_test.go
jharveyb: These changes work, I meant that both `FromAddress()` and the test logic to make packets should copy the chainparams from the address.

taroscript/send.go
jharveyb: @Roasbeef does this sound correct?

tarofreighter/wallet.go
```diff
@@ -0,0 +1,644 @@
...
+	receiverAddr address.Taro) (*taropsbt.VPacket,
+	*commitment.TaroCommitment, error) {
+
+	// We start by creating a new virtual transaction that will be used to
+	// hold the asset transfer. Because sending to an address is always a
+	// non-interactive process, we can use this function that always creates
+	// a change output.
+	vPkt := taropsbt.FromAddress(&receiverAddr)
+
+	desc := &taroscript.RecipientDescriptor{
```
jharveyb, line 132: Is there a reason we don't populate the script key from the address as well?

tarofreighter/wallet.go
```diff
@@ -0,0 +1,644 @@
...
+
+// FundPacket funds a virtual transaction, selecting assets to spend in order to
+// pay the given recipient. The selected input is then added to the given
+// virtual transaction.
+func (f *AssetWallet) FundPacket(ctx context.Context,
+	desc *taroscript.RecipientDescriptor,
+	vPkt *taropsbt.VPacket) (*commitment.TaroCommitment, error) {
+
+	// The input and address networks must match.
+	if !address.IsForNet(vPkt.ChainParams.TaroHRP, f.cfg.ChainParams) {
```
jharveyb, line 153: 💯

tarofreighter/parcel.go
```diff
@@ -571,55 +301,54 @@ func (s *sendPackage) createProofs() (spendProofs, error) {
...
 }
 
 // deliverResponse delivers a response for the parcel back to the receiver over
 // the specified response channel.
 func (s *sendPackage) deliverResponse(respChan chan<- *PendingParcel) {
-	oldRoot := s.InputAsset.Commitment.TapscriptRoot(nil)
+	oldRoot := s.InputCommitment.TapscriptRoot(nil)
 
+	senderAsset := s.VirtualPacket.Outputs[0].Asset
+	receiverAsset := s.VirtualPacket.Outputs[1].Asset
```
jharveyb, line 344: Should check if our send was interactive here, otherwise this panics? Unless deferring that for follow-up PR.

taroscript/send_test.go
```diff
@@ -2095,119 +1805,101 @@ func TestAddressValidInput(t *testing.T) {
 	}
 }
 
+func addrToDesc(addr address.Taro) *taroscript.RecipientDescriptor {
+	return &taroscript.RecipientDescriptor{
```
jharveyb, line 1809: Also dropping script key here.

taroscript/send_test.go
```diff
@@ -301,201 +312,129 @@ func createGenesisProof(t *testing.T, state *spendData) {
...
+
+	if fullValueInteractive {
+		outputs = []*taropsbt.VOutput{{
+			Interactive: true,
+			Amount:      addr.Amount,
+			ScriptKey: asset.NewScriptKey(
+				&addr.ScriptKey,
+			),
+			AnchorOutputIndex:       receiverExternalIdx,
+			AnchorOutputInternalKey: &state.spenderPubKey,
```
jharveyb, line 346: Shouldn't this key be `addr.InternalKey` / receiver-controlled in the interactive case?

jharveyb (commented):
Some further nits that may be resolved in other PRs, otherwise looks ready 💯

taropsbt/interface.go
Roasbeef: Yeah this is just the inclusion proof that the input asset actually exists (not full proof history). With that, I don't think we need to worry about any length limits for the PSBT, assuming file/direct/encoded transfer.

taropsbt/interface.go
```diff
@@ -0,0 +1,213 @@
...
+	Anchor Anchor
+
+	// asset is the full instance of the asset being spent. It is not
+	// exported because the assets script key must be encoded in the PSBT
+	// input struct for the signing to work correctly.
+	asset *asset.Asset
+
+	// assetProof is the proof that the asset being spent was committed to
+	// in the anchor transaction above.
+	assetProof *commitment.Proof
```
Roasbeef, line 146: Is this useful if none of the header or tapscript information is also included? 

My inclusion proof earlier, meant the actual full input proof: https://github.com/lightninglabs/taro/blob/main/proof/proof.go#L30-L69

Note that this "just" the full inclusion proof for the previous input, and not total asset history. A few of these fields are already present here like the transaction, and merkle proof, etc. 

Perhaps we useful middle ground would be a `InclusionProof() *proof.Proof` method that assembles this struct for callers given the information in this PSBT?

taroscript/send.go
```diff
@@ -0,0 +1,232 @@
...
+
+		if !vOut.Interactive {
+			return fmt.Errorf("single output must be interactive")
+		}
+
+		if vOut.Amount != inputAsset.Amount {
+			return ErrInvalidSplitAmounts
+		}
+
+		if vOut.ScriptKey.PubKey.IsEqual(asset.NUMSPubKey) {
```
Roasbeef, line 243: I think this is ok for now, but we might lift restriction later once we implement provable burns (#24).

taroscript/send.go
```diff
@@ -0,0 +1,232 @@
...
+			return commitment.ErrInvalidScriptKey
+		}
+
+	// A two output transaction must have the change at index 0 if it is a
+	// non-interactive send.
+	case len(outputs) == 2:
+		if outputs[1].IsChange {
```
Roasbeef, line 250: This restrictions seem artificial, given that we have the `IsChange` field now. Alternatively, we just need something like:
```
haveChangeOutput := chanutils.Any(outputs, func(o *VOutput) bool{
    return o.IsChange
})
```

Also in terms of future proofing two outputs can be used for interactive/sync sends: in cases where the sender actually requires a"normal" change output. 

Maybe we ran to rename the `IsChange` field to be something like `IsSplitChange`? Then it signals that this is a non-interactive send, which iiuc, is a sort of special path that only the porter will actually use.

taroscript/send.go
Roasbeef: Including the root there was meant to allow for the root + all splits to be validated in a single go. 

> Hmm, I'm not sure about this. We don't have that split asset anymore here, it's discarded in PrepareOutputAsset, as we only need the root asset (which is a copy of the split asset at the root locator, with the tx witness and split commitment root set).

Sounds like y'all are referring to the same thing here? The root asset here is the same as "change" in this parlance, which is the asset that anchors all the created external splits.

tarofreighter/wallet.go
Roasbeef: I think because this is described in the `vPSBT` package?

guggero: Because we only use the descriptor to fund and then verify the inputs to send to the address. So the recipient's script key isn't needed. I removed it from the struct and renamed it to `FundingDescriptor` to make it a bit more clear.

tarofreighter/parcel.go
guggero: Yes, that'll be done in the follow-up PR.

taroscript/send_test.go
guggero: Yep, that's on purpose. I hope with the new name it's a bit more clear why we don't need the script key.

taropsbt/interface.go
guggero: I wasn't sure if we wanted to have all the duplicated data (like `PrevOut`, `AnchorTx` and `Asset`) with the full proof. And it also creates an issue with a circular package dependency if we use `*proof.Proof`. But I see how it can be useful, so I opted to use a byte slice instead of the structured `*proof.Proof`.

guggero: I'll add a commit to verify the proof before signing as well as using the input proof when creating the receiver proofs in part 3 tomorrow.

taroscript/send.go
guggero: The `IsChange` is not just meant for split change, also normal change. I did incorporate your suggestion with the `Any` and also removed some of the rules to make this even more generic.

taroscript/send.go
guggero: We talked about it again offline and I think we are in sync now. I added a commit (7170c4a8) to solidify this updated assumption in the VM. I hope you agree with the change as well, @Roasbeef.

guggero:
I think I addressed all comments in this PR, ready for final review.

taroscript/send.go
jharveyb: From offline discussion:

Modify the vInput format to include an optional field for the merkle proof that the root asset is included in the split commitment tree. Also update encode/decode to include this field, and update `SignVirtualTransaction` to pass this proof to the VM in the existing form.

guggero:
Latest update: After chatting with @ffranr a bit and explaining the code, I decided to rename the `IsChange` field to `IsSplitRoot`. So things should be more clear and less subjective now.

I also added the new field `SplitAsset` that is only set when `IsSplitRoot` is `true`, that houses the split asset before it was turned into the root asset (as discussed, for validation only).

ffranr (commented):
Great documentation, nice work!

chanutils/func.go
```diff
@@ -70,3 +72,35 @@ func All[T any](xs []T, pred func(T) bool) bool {
 
 	return true
 }
+
+// Any returns true if the passed predicate returns true for any item in the
+// slice.
+func Any[T any](xs []T, pred func(T) bool) bool {
+	for _, x := range xs {
+		x := x
```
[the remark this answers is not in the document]
guggero: Yes, in this specific case it wasn't necessary, because calling a method will put the pointer on the stack, which causes the Golang loop variable "bug" not to apply. But in general, it's good practice to do this anyway (or switch to the index based access which I did).

taropsbt/interface.go
```diff
@@ -0,0 +1,287 @@
...
+	// Interactive, when set to true, indicates that the receiver of the
+	// output is aware of the asset transfer and can therefore receive a
+	// full value send directly and without a tombstone in the change
+	// output.
+	Interactive bool
+
+	// AnchorOutputIndex indicates in which output index of the BTC
+	// transaction this asset output should be committed to. Multiple asset
+	// outputs can be committed to within the same BTC transaction output.
+	AnchorOutputIndex uint32
```
ffranr, line 293: We discussed `AnchorOutputIndex` yesterday I'm just adding a note here to remind myself:

Should the virtual output be concerned with the anchor output index? Perhaps the anchor output index should be decided later in the process whilst taking other fields in `VOutput` into account. The number of recipients needs to be taken into account also when assigning anchor output indexes.

taropsbt/interface.go
```diff
@@ -0,0 +1,275 @@
...
+		if p.Outputs[i].Asset == nil {
+			return false, fmt.Errorf("recipient asset %d is nil", i)
+		}
+
+		// If there is a split commitment witness present, we know there
+		// is a split going on.
+		if p.Outputs[i].Asset.HasSplitCommitmentWitness() {
+			return true, nil
+		}
+	}
```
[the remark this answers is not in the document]
guggero: > The doc mentions inspecting input assets but the function doesn't do that. Should it?

No, this only looks at the outputs. And the doc only mentions when this is expected to be the case, not that it actually needs to look at the inputs (which wouldn't tell you this anyway, since the outputs determine whether it's a split or not).

> Should there be an equivalent for Outputs?

I don't think there is a need for it, no. The reason it was added for the input asset was to make sure we always create a copy. But because an output asset can only be created by making a copy of the input (or creating a split commitment, which internally copies the assets as well), that restriction is not necessary. The reason why an asset can be nil in the output is because when we call the `FundVirtualTransaction`, we don't yet know what the asset is supposed to look like, we only say what amount we want to send where.

> Also why not for i, output := range p.Outputs?

Good point, switched.

taropsbt/interface.go
```diff
@@ -0,0 +1,275 @@
...
+func (p *VPacket) FirstInteractiveOutput() (*VOutput, error) {
+	result, err := chanutils.First(p.Outputs, VOutIsInteractive)
+	if err != nil {
+		return nil, fmt.Errorf("no interactive output found")
+	}
+
+	return result, nil
+}
+
+// Anchor is a struct that contains all the information about an anchor output.
```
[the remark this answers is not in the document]
guggero: Hmm, the word "anchor" (as opposed to "virtual") should already indicate that it's on the bitcoin level. Or do you think that is not sufficient?

taropsbt/interface.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// proof is the proof blob that proves the asset being spent was
+	// committed to in the anchor transaction above. This cannot be of type
+	// proof.Proof directly because that would cause a circular dependency.
+	proof []byte
+}
+
+// Asset returns the input's asset that's being spent.
+func (i *VInput) Asset() *asset.Asset {
+	return i.asset
```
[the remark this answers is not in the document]
guggero: A copy is created when setting the field. So we don't need to copy again when accessing it (as it should only be read in most cases anyway).

taropsbt/interface.go
guggero: Correct, the virtual transaction doesn't need to know the _exact_ anchor output index, it just needs to know how many different ones there should be (e.g. how many recipients).

jharveyb (approved):
Looks good 🚀

Roasbeef (approved):
LGTM 🐙

PR #245, opened 2023-01-25
psbt script spend [3/3]: allow signing for asset script path spends through PSBTs

guggero:
~Depends on https://github.com/lightninglabs/taro/pull/243.~
Depends on https://github.com/lightninglabs/taro/pull/244.
~Depends on https://github.com/btcsuite/btcd/pull/1942.~

Fixes https://github.com/lightninglabs/taro/issues/75.

This PR contains the PSBT serialization logic for transporting asset related meta data in a BIP-174 based PSBT packet as well as all the required RPCs to allow creating and sending to a Taro address that locks assets to a script spend path.


jharveyb (commented):
Looking good! Surprisingly small changes to get script working. Only major suggested change is the FetchCommitment usage.

Will revisit to double-check the encode and Taproot logic.

taropsbt/decode.go
```diff
@@ -0,0 +1,377 @@
...
+			"%w", err)
+	}
+	chainParams, err := address.Net(string(hrp))
+	if err != nil {
+		return nil, fmt.Errorf("error parsing Taro chain params HRP: "+
+			"%w", err)
+	}
+
+	// Currently only one input is supported.
+	if len(packet.Inputs) != 1 {
```
Roasbeef, line 63: Perhaps move this to a higher level validation layer (within the `AssetWallet` itself) so we don't need to go back and modify this later once we add additional inputs?

taropsbt/decode.go
```diff
@@ -0,0 +1,377 @@
...
+		prevID            *asset.PrevID
+		anchorValue       uint64
+		anchorSigHashType uint64
+	)
+
+	mapping := []struct {
+		key     []byte
+		decoder func([]byte) error
+	}{{
+		key:     PsbtKeyTypeInputTaroPrevID,
```
Roasbeef, line 113: TLV within PSBT 👽

Def a fan of, as the OG PSBT serialization format is a bit wonky.

tarodb/assets_store.go
```diff
@@ -660,6 +660,34 @@ func constraintsToDbFilter(query *AssetQueryFilters) QueryAssetFilters {
...
+		AnchorPoint:   anchorPointBytes,
+		MinAmt: sql.NullInt64{
+			Int64: int64(amt),
+			Valid: true,
+		},
+	}
+
+	if groupKey != nil {
+		key := groupKey.GroupPubKey
+		filter.KeyGroupFilter = key.SerializeCompressed()
```
Roasbeef, line 685: isn't the asset ID enough to be able to uniquely pick out an asset?

tarodb/assets_store.go
```diff
@@ -1226,6 +1254,30 @@ func queryChainAssets(ctx context.Context, q ActiveAssetsStore,
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to create filter: %w", err)
+	}
+
+	commitments, err := a.queryCommitments(ctx, filter)
+	if err != nil {
+		return nil, fmt.Errorf("unable to query commitments: %w", err)
+	}
+
+	if len(commitments) != 1 {
```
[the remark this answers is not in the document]
Roasbeef: Yeh see other comments re letting more than 1 go thru here, and then instead adding the temporary restriction at a higher level (within the wallet wallet itself).

tarodb/assets_store.go
```diff
@@ -1226,6 +1254,30 @@ func queryChainAssets(ctx context.Context, q ActiveAssetsStore,
 	return matchingAssets, nil
 }
 
+// FetchCommitment returns a specific commitment identified by the given asset
+// parameters. If no commitment is found, ErrNoCommitment is returned.
+func (a *AssetStore) FetchCommitment(ctx context.Context, id asset.ID,
```
Roasbeef, line 1259: Any reason this was added instead of using `SelectCommitment` which is more generic? Alternatively if a simpler API is needed elsewhere, then this can just call directly into that.

taroscript/send.go
```diff
@@ -313,28 +312,24 @@ func SignVirtualTransaction(input *taropsbt.VInput, outputs []*taropsbt.VOutput,
...
-
-		newWitness, err := SignTaprootKeySpend(
-			*input.Asset().ScriptKey.RawKey.PubKey, virtualTxCopy,
-			prevAsset, 0, txscript.SigHashDefault, signer,
-		)
-		if err != nil {
-			return err
-		}
+	// TODO(guggero): Update this once we want to support more than one
+	// asset being spent in a single transaction.
```
Roasbeef, line 316: See other comments re leaving the existing logic as it should work once we allow multiple inputs. We can also do simpler exercises of this by doing things like send two collectibles via an interactive flow in itests.

taroscript/tx.go
```diff
@@ -327,35 +329,116 @@ func InputScriptSpendSigHash(virtualTx *wire.MsgTx, input *asset.Asset,
...
+	witness := wire.TxWitness{sig.Serialize()}
+	if vIn.SighashType != txscript.SigHashDefault {
+		witness[0] = append(witness[0], byte(vIn.SighashType))
+	}
+
+	// If this was a script spend, we also have to add the script itself and
+	// the control block to the witness, otherwise the verifier will reject
+	// the generated witness.
+	if spendDesc.SignMethod == input.TaprootScriptSpendSignMethod {
+		witness = append(witness, spendDesc.WitnessScript)
```
Roasbeef, line 439: 😈

tarofreighter/chain_porter.go
```diff
@@ -187,6 +187,25 @@ func (p *ChainPorter) RequestShipment(req *AssetParcel) (*PendingParcel, error)
 	}
 }
 
+// RequestSignedDelivery requests a shipment of a virtual transaction that has
+// already been signed. This directly injects the deliverable into the state
+// machine at the state where it is anchored on chain.
+func (p *ChainPorter) RequestSignedDelivery(vPacket *taropsbt.VPacket,
```
Roasbeef, line 193: Very cool that we're able to enable some external interaction with such a small added shim.

tarofreighter/chain_porter.go
```diff
@@ -229,28 +248,45 @@ func (p *ChainPorter) taroPorter() {
...
+					req.Dest.ScriptKey.SerializeCompressed())
+
+				// Initialize a package with the destination
+				// address.
+				sendPkg = &sendPackage{
+					ReceiverAddr: req.Dest,
+				}
+
+			case req.pkg != nil:
+				sendPkg = req.pkg
```
Roasbeef, line 268: Add log that this is the PSBT flow?

tarofreighter/parcel.go
```diff
@@ -79,6 +79,10 @@ type AssetParcel struct {
 	// Dest is the address that should be used to satisfy the transfer.
 	Dest *address.Taro
 
+	// pkg is the actual package that is being shipped. This is not set in
+	// case we request delivery to a destination address.
+	pkg *sendPackage
```
Roasbeef, line 84: Perhaps use a distinct struct? Need to double check if `Dest` is still sent in this case, as otherwise might be some panics elsewhere since the field isn't set. 

We could also modify the actual channel use here to use a type param based on an interface that exposes the data needed via methods.

rpcserver.go
```diff
@@ -1040,6 +1064,133 @@ func (r *rpcServer) AddrReceives(ctx context.Context,
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching input commitment: %w",
+			err)
+	}
+
+	rpcsLog.Debugf("Selected commitment for anchor point %v, requesting "+
+		"delivery", inputCommitment.AnchorPoint)
+
+	resp, err := r.cfg.ChainPorter.RequestSignedDelivery(
```
Roasbeef, line 1184: So then this also assumes that the wallet as is will fully handle all the PSBT components? Could imagine another world where there's another external round to sign the Bitcoin portion (as that has a multi-sig/musig2 layer or w/e).

tarorpc/assetwalletrpc/assetwallet.proto
```diff
@@ -0,0 +1,140 @@
...
+    rpc AnchorVirtualPsbts (AnchorVirtualPsbtsRequest)
+        returns (tarorpc.SendAssetResponse);
+}
+
+message FundVirtualPsbtRequest {
+    oneof template {
+        /*
+        Use an existing PSBT packet as the template for the funded PSBT.
+
+        TODO(guggero): Actually implement this. We can't use the "reserved"
```
Roasbeef, line 40: Missing link here is parsing? Can def push that out to other PRs since this series itself is already somewhat hefty.

itest/psbt_test.go
```diff
@@ -0,0 +1,180 @@
...
+		t.t, t, t.lndHarness.BackendCfg, t.lndHarness.Bob,
+		t.universeServer,
+	)
+	defer func() {
+		require.NoError(t.t, secondTarod.stop(true))
+	}()
+
+	var (
+		alice = t.tarod
+		// aliceLnd = t.lndHarness.Alice
```
Roasbeef, line 47: Stray comment?

itest/psbt_test.go
```diff
@@ -0,0 +1,180 @@
...
+		ctxb, &walletrpc.KeyReq{
+			KeyFamily: int32(taroscript.TaroKeyFamily),
+		},
+	)
+	require.NoError(t.t, err)
+	bobScriptInternalKey, err := btcec.ParsePubKey(
+		bobScriptKeyDesc.RawKeyBytes,
+	)
+	require.NoError(t.t, err)
+	bobInternalKeyDesc, err := bobLnd.WalletKitClient.DeriveNextKey(
```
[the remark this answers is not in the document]
Roasbeef: As in re-expose the lnd API used to create keys in the first place?

itest/psbt_test.go
```diff
@@ -0,0 +1,180 @@
...
+	wrpc "github.com/lightninglabs/taro/tarorpc/assetwalletrpc"
+	"github.com/lightninglabs/taro/taroscript"
+	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/lnrpc/walletrpc"
+	"github.com/stretchr/testify/require"
+)
+
+// testPsbtSend tests that we can properly send assets back and forth between
+// nodes with the use of PSBTs.
+func testPsbtSend(t *harnessTest) {
```
Roasbeef, line 23: 😎

Roasbeef (changes requested):
Excellent work! Gotta love when a few well crafted abstractions nicely fall into place. 

See comments in line, some carried over from the PR before it was split up. 

Few things popped out:
  * Ensuring that all the decoding + parsing is generic up until we restrict things to a single input deeper within the pipeline. 
  * Adding a similar coin selection mutex like we have in `lnd` over everything, which ensures that we don't have weird double spending issues. Alternatively, we can pipe everything through the porter and have it do single threaded coin selection + locking accounting. 
  * Update the test to also have a checksig in addition to the pre-image reveal so the vPSBT signing can be exercised there.

tarodb/assets_store.go
```diff
@@ -660,6 +660,34 @@ func constraintsToDbFilter(query *AssetQueryFilters) QueryAssetFilters {
...
+	if err != nil {
+		return QueryAssetFilters{}, fmt.Errorf("unable to encode "+
+			"outpoint: %w", err)
+	}
+
+	filter := QueryAssetFilters{
+		AssetIDFilter: id[:],
+		AnchorPoint:   anchorPointBytes,
+		MinAmt: sql.NullInt64{
+			Int64: int64(amt),
```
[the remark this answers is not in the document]
guggero: I removed the amount here and replaced it by the script key, as that should select the correct input asset with full precision (as compared to the amount).

tarodb/assets_store.go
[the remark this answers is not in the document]
guggero: You're right, with the previous query it could've selected two coins of the same asset if they had the same value (or both a value greater than what we need). I replaced the amount with the script key in the query, which should only ever give one result (since the script key needs to be unique per anchor TX, I believe).

@Roasbeef at this level we are looking to get one specific asset input, independent of how many we allow on the virtual TX level.

tarodb/assets_store.go
guggero: No, in case of a normal asset where I have multiple coins, we actually need the script key, which I added now. We don't necessarily need the group key for our specific query, but because this function creates a generic filter, I'm leaving that in for now.

tarodb/assets_store.go
guggero: The `SelectCommitment` didn't allow us to specify the asset we want precisely enough. There was always the chance that we'd get the wrong one if we had multiple asset UTXOs of the same value. I fixed that by adding the `FetchCommitment` method which now also takes the script key to make an asset UTXO identifiable.

tarofreighter/parcel.go
guggero: Yeah, it wasn't really clear what fields would be accessed in what case. I refactored the parcels to make everything more generic and clear.

rpcserver.go
guggero: Yes, definitely. The BTC level PSBT components will come in a later PR and give you full control over how the anchor TX is created, signed and published.

tarorpc/assetwalletrpc/assetwallet.proto
guggero: Not just parsing, but a whole bunch of additional validations (e.g. only one input, first output set as change output, script keys set properly, coin selected properly and so on). So yeah, going to leave it for now.

dstadulis:
full value send test needs to be written.

taropsbt/interface.go
```diff
@@ -0,0 +1,404 @@
...
+	"github.com/lightninglabs/taro/address"
+	"github.com/lightninglabs/taro/asset"
+	"github.com/lightninglabs/taro/commitment"
+	"github.com/lightningnetwork/lnd/keychain"
+)
+
+// We define a set of Taro specific global, input and output PSBT key types here
+// that correspond to the custom types defined in the VPacket below. We start at
+// 0x70 because that is sufficiently high to not conflict with any of the keys
+// specified in BIP174. Also, 7 is leet speak for "t" as in Taro.
```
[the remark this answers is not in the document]
guggero: Added two more sentences to the comment.

tarofreighter/chain_porter.go
```diff
@@ -229,28 +248,45 @@ func (p *ChainPorter) taroPorter() {
...
-				req.Dest.ScriptKey.SerializeCompressed())
-
-			// Initialize a package with the destination address.
-			sendPkg := sendPackage{
-				ReceiverAddr: req.Dest,
+			var sendPkg *sendPackage
+
+			// The request either has a destination address we want
+			// to send to, or a send package is already initialized.
+			switch {
```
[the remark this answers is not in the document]
guggero: I refactored the parcels into two distinct types to make this more clear.

tarorpc/assetwalletrpc/assetwallet.proto
guggero: We'll start to support this to enable the interactive send case. But I decided to split that into a separate PR since more changes are necessary than initially expected.

guggero:
Thanks a lot for all the feedback, @Roasbeef and @jharveyb!

I addressed all of the comments and rebased on the latest version of part 2.

 
>     * Ensuring that all the decoding + parsing is generic up until we restrict things to a single input deeper within the pipeline.

Good idea, done!

 
>     * Adding a similar coin selection mutex like we have in `lnd` over everything, which ensures that we don't have weird double spending issues. Alternatively, we can pipe everything through the porter and have it do single threaded coin selection + locking accounting.

I agree. Though I think a mutex only adds any value if we can also mark a coin as being used/locked in the DB. I created an issue (https://github.com/lightninglabs/taro/issues/261) to track that.

 
>     * Update the test to also have a checksig in addition to the pre-image reveal so the vPSBT signing can be exercised there.

Yes, done.

I'm also going to split off the interactive send changes into a separate PR (I guess it will be [4/3] :sweat_smile: ), since the changes were a bit more involved than expected.

itest/psbt_test.go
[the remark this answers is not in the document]
guggero: Yeah, I think it could be useful to have this call be threaded through the daemon so we don't need to know what key family to use in `lnd` when talking to `tarod`. Will add that to the PR that adds the BTC level PSBT stuff (for funding and signing the anchor transactions externally).

tarofreighter/parcel.go
```diff
@@ -147,19 +94,138 @@ func (s SendState) String() string {
...
+	}
+}
+
+// pkg returns the send package that should be delivered.
+func (p *PreSignedParcel) pkg() *sendPackage {
+	log.Infof("New signed delivery request with %d outputs",
+		len(p.vPkt.Outputs))
+
+	// Initialize a package with the destination address.
+	return &sendPackage{
```
jharveyb, line 202: nit: inaccurate comment.

jharveyb (commented):
Looks good, only nits left IMO + changes in preceding PR.

itest/psbt_test.go
```diff
@@ -0,0 +1,336 @@
...
+	err = fundedPacket.Serialize(&b)
+	require.NoError(t.t, err)
+
+	signedResp, err := bob.SignVirtualPsbt(
+		ctxb, &wrpc.SignVirtualPsbtRequest{
+			FundedPsbt: b.Bytes(),
+		},
+	)
+	require.NoError(t.t, err)
+	require.Contains(t.t, signedResp.SignedInputs, uint32(0))
```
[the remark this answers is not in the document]
guggero: Hmm, but since `SignVirtualPsbt` explicitly mentions virtual transactions in its name, I think it should be clear what the indexes refer to? Added a comment anyway.

tarogarden/mock.go
```diff
@@ -36,7 +36,7 @@ func RandSeedlings(t testing.TB, numSeedlings int) map[string]*Seedling {
 			AssetType:      asset.Type(rand.Int31n(2)),
 			AssetName:      assetName,
 			Metadata:       test.RandBytes(32),
-			Amount:         uint64(rand.Int63()),
+			Amount:         uint64(rand.Int31()),
```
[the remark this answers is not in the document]
guggero: Cool, fixed it there as well.

taropsbt/decode.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+// decode decodes the given PInput into the current VInput.
+func (i *VInput) decode(pIn psbt.PInput) error {
+	i.PInput = pIn
+
+	var (
+		prevID            *asset.PrevID
+		anchorValue       uint64
+		anchorSigHashType uint64
```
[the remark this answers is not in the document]
guggero: Nice catch, fixed and test case updated!

taropsbt/encode.go
```diff
@@ -0,0 +1,354 @@
...
+	if err := p.Serialize(&b); err != nil {
+		return "", err
+	}
+
+	return base64.StdEncoding.EncodeToString(b.Bytes()), nil
+}
+
+// encode encodes the current VInput struct into a PInput and a wire.TxIn.
+func (i *VInput) encode() (psbt.PInput, *wire.TxIn, error) {
+	txIn := &wire.TxIn{}
```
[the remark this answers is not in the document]
guggero: Good point, I removed it from the `encode()` method.

taroscript/send.go
```diff
@@ -290,358 +176,564 @@ func IsValidInput(input *commitment.TaroCommitment,
...
+
+		// We need at least the script key set on each output.
+		if scriptPubKey == nil {
+			return fmt.Errorf("output %d has no script key", idx)
+		}
+
+		// Any non change output shouldn't send to an un-spendable
+		// script, burning assets is done in a different way.
+		if !outputs[idx].IsChange &&
+			scriptPubKey.IsEqual(asset.NUMSPubKey) {
```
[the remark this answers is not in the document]
guggero: Yes, makes sense. Did it in part 2 though.

guggero:
Addressed all comments and added the input proof validation as well.

tarofreighter/wallet.go
```diff
@@ -0,0 +1,717 @@
...
+	error) {
+
+	// Now we'll use the signer to sign all the inputs for the new taro
+	// leaves. The witness data for each input will be assigned for us.
+	signedInputs := make([]uint32, len(vPkt.Inputs))
+	for idx := range vPkt.Inputs {
+		// Before we sign the transaction, we want to make sure the
+		// inclusion proof is valid and the asset is actually committed
+		// in the anchor transaction.
+		err := verifyInclusionProof(vPkt.Inputs[idx])
```
jharveyb, line 334: Probably want a TODO here to add a call to `ChainBridge` later on? To verify the anchor TX from the proof is in the chain.

tarofreighter/wallet.go
```diff
@@ -0,0 +1,717 @@
...
+	anchorKey, err := proof.ExtractTaprootKeyFromScript(
+		vIn.Anchor.PkScript,
+	)
+	if err != nil {
+		return fmt.Errorf("unable to parse anchor pk script taproot "+
+			"key: %w", err)
+	}
+
+	inclusionProof := assetProof.InclusionProof
+	proofKey, _, err := inclusionProof.DeriveByAssetInclusion(
```
jharveyb, line 372: nit: have to check this err.

jharveyb (approved):
Looking good! Just have to remove `proof{Encoder,Decoder}`.

tarofreighter/wallet.go
guggero: Great point! That also reminded me that we didn't check the anchor TX in the input vs. the one in the proof, which I added now. So only the "is the block in the chain" logic remains, for which I added a TODO.

taroscript/tx.go
```diff
@@ -327,35 +329,123 @@ func InputScriptSpendSigHash(virtualTx *wire.MsgTx, input *asset.Asset,
...
+		}
+		leafHash := leaf.TapHash()
+		if !bytes.Equal(leafHash[:], derivation.LeafHashes[0]) {
+			return nil, fmt.Errorf("specified leaf hash in " +
+				"taproot BIP32 derivation but corresponding " +
+				"taproot leaf script was not found")
+		}
+
+		spendDesc.SignMethod = input.TaprootScriptSpendSignMethod
+		spendDesc.TapTweak = vIn.TaprootMerkleRoot
```
Roasbeef, line 421: Why is the tap tweak needed for a script spend from the tree? This is only needed for keyspend paths.

taroscript/tx.go
```diff
@@ -327,35 +329,123 @@ func InputScriptSpendSigHash(virtualTx *wire.MsgTx, input *asset.Asset,
...
+	case len(vIn.TaprootMerkleRoot) == sha256.Size &&
+		len(derivation.LeafHashes) == 0:
+
+		spendDesc.SignMethod = input.TaprootKeySpendSignMethod
+		spendDesc.TapTweak = vIn.TaprootMerkleRoot
+
+	// One leaf hash and a merkle root means we're signing a specific
+	// script.
+	case len(vIn.TaprootMerkleRoot) == sha256.Size &&
+		len(derivation.LeafHashes) == 1:
```
Roasbeef, line 397: Does this restrict to only a control block of size 1? So you only have a single sibling.

Roasbeef (approved):
LGTM 🌳

Just have one clarifying question (about the leaf hashes in the PSBT), and another re a field that doesn't need to be set for a script path spend (the tap tweak).

taroscript/tx.go
guggero: Good point, removed.

taroscript/tx.go
guggero: No, we just only support signing for a single leaf at a time. The PSBT spec allows you to specify multiple leaf hashes and that would indicate to the signer that it should create a signature for each of them. But since we always know which path we're going to use, we haven't implemented this in `lnd` either. I updated the comment to make it more clear.

PR #249, opened 2023-02-01
multi: remove existing meta field in asset TLV to replace w/ meta hash optionally revealed in genesis minting proof

Roasbeef:
Related to https://github.com/lightninglabs/taro/issues/62. 

## Problem Statement 

In this PR, we fix an issue with the meta field as defined today:
  * The field in practice may be very large (hundreds of KBs), and is serialized along with each asset TLV in a proof chain. 
  * This means that the proof size is a function of the size of the meta field itself, meaning more data to lug around for clients.

### Meta -> Meta Hash in TLV 

Instead, we make the meta field itself a _hash commitment_. Only the _meta hash_ field exists in every asset TLV. This means the asset genesis (which rn is serialized along with addresses) nearly constant sized other than the tag (which we can also instead make into a commitment). The meta hash is then the hash of the TLV serialization of the meta reveal (see below). 

To start, the meta has a single type: opaque blob. This type is itself a TLV, so we can add more types in the future, and also other types contingent on the type itself. Eg: we can add a JSON blob type, or a MIME type that then relies on the existence of some other string to fully bind the structure of the meta bytes. 

### MetaReveal in Proof File Blob

We then add a new optional field to the proof format that allows the minter (or anyone that knows of the pre-image) to reveal it within the _first_ proof state transition. As this is only exist in the first state, which is to be bootstrapped by the users from a Base Universe, the proof sizes are no longer dependent on the asset meta itself (constant value at the start). We then add a rule that this can only exist for assets that have a genesis witness (minted assets). 

### New `assets_meta` table in the database 

Along the way we modify the DB to add a new `assets_meta` table and reference that directly. This allows for larger meta blobs, as we no longer need to read the entire thing if we want to look at the genesis details for assets. Two query mechanisms based on the asset ID and the asset hash have also been added. 

### RPC + CLI Changes

On the RPC layer, we now expose the meta field to callers, and display the hash in most other locations. We also add an API that lets callers fetch the meta based on the hash or asset ID. 

On the CLI, we propagate all the above updates, then also add a new option to read the meta from a file on disk, as it may be too large to pass as a command line string. 

## Follow up Work

Update the spec accordingly.


Roasbeef:
Spec PR here: https://github.com/Roasbeef/bips/pull/34

tarodb/sqlc/migrations/000002_assets.up.sql
```diff
@@ -31,6 +31,21 @@ CREATE TABLE IF NOT EXISTS genesis_points (
     anchor_tx_id INTEGER REFERENCES chain_txns(txn_id)
 );
 
+-- assets_meta is a table that holds all the metadata information for genesis
+-- assets that we either created, or bootstrapped from the relevant Base
+-- Universe.
+CREATE TABLE IF NOT EXISTS assets_meta (
+    meta_id INTEGER PRIMARY KEY,
+
+    meta_data_hash BLOB CHECK(length(meta_data_hash) = 32) UNIQUE,
```
guggero, line 40: nit: move the `UNIQUE` keyword to before the `CHECK` keyword to match the other checks we already have in place?

tarodb/sqlc/migrations/000002_assets.up.sql
```diff
@@ -31,6 +31,21 @@ CREATE TABLE IF NOT EXISTS genesis_points (
...
+
+    -- TODO(roasbeef): also have other opque blob here for future fields?
+    meta_data_blob BLOB,
+
+    meta_data_type SMALLINT
+);
+CREATE INDEX IF NOT EXISTS meta_hash_index on assets_meta(meta_data_hash);
```
guggero, line 47: Thanks to the `UNIQUE` keyword I think we don't need an explicit index: https://stackoverflow.com/questions/3379292/is-an-index-needed-for-a-primary-key-in-sqlite

tarodb/sqlc/migrations/000002_assets.up.sql
```diff
@@ -42,7 +57,7 @@ CREATE TABLE IF NOT EXISTS genesis_assets (
 
     asset_tag TEXT UNIQUE NOT NULL,
 
-    meta_data BLOB,
+    meta_data_id INTEGER REFERENCES assets_meta(meta_id),
```
guggero, line 60: Is the meta field optional? I think it's not on the Golang side. So the question would be, would we actually insert `NULL` here if the `MetaHash` array is a zero array? I'd say we probably should just insert the zero-hash as a meta record and make this field `NOT NULL`.

tarodb/asset_minting.go
```diff
@@ -437,11 +472,14 @@ func fetchAssetSprouts(ctx context.Context, q PendingAssetStore,
 		assetGenesis := asset.Genesis{
 			FirstPrevOut: genesisPrevOut,
 			Tag:          sprout.AssetTag,
-			Metadata:     sprout.MetaData,
 			OutputIndex:  uint32(sprout.GenesisOutputIndex),
 			Type:         asset.Type(sprout.AssetType),
 		}
 
+		if len(sprout.MetaHash) != 0 {
+			copy(assetGenesis.MetaHash[:], sprout.MetaHash[:])
```
guggero, line 480: nit: can remove `[:]` from `sprout.MetaHash`.

tarodb/asset_minting.go
```diff
@@ -487,6 +525,38 @@ func fetchAssetSprouts(ctx context.Context, q PendingAssetStore,
 	return taroCommitment, nil
 }
 
+// fetchAssetMetas attempts to fetch the asset meta reveal for each of the
+// passed assets.
+func fetchAssetMetas(ctx context.Context, db PendingAssetStore,
+	assets []*asset.Asset) (
+	map[asset.SerializedKey]*proof.MetaReveal, error,
+) {
```
guggero, line 533: nit: weird wrapping.

tarodb/asset_minting_test.go
```diff
@@ -85,7 +87,8 @@ func TestCommitMintingBatchSeedlings(t *testing.T) {
 	assetStore, _, _ := newAssetStore(t)
 
 	ctx := context.Background()
-	const numSeedlings = 5
+	//const numSeedlings = 5
```
guggero, line 90: nit: commented out code.

tarodb/assets_common.go
```diff
@@ -308,29 +312,81 @@ type FetchGenesisStore interface {
...
+			Valid: true,
+		}
+
+	// Otherwise, we'll just be inserting only the meta hash. At a later
+	// time, the reveal/blob can also be inserted.
+	case assetGen != nil:
+		metaHash = assetGen.MetaHash
+
+	// In the default case, there's no actual meta reveal, so we'll just
+	// use the meta hash of all zeroes.
```
guggero, line 377: Ah, I guess that answers my previous question about the zero hash. So we can make the `meta_data_id` non-null in the DB, which removes one instance of `NullInt32`.

tarorpc/taro.proto
```diff
@@ -123,6 +129,32 @@ enum AssetType {
     */
     COLLECTIBLE = 1;
 }
+
+enum AssetMetaType {
+    /*
+    Opaque is used for asset meta blobs that have no true structure and instead
+    should be interested as opaque blobs.
+    */
+    OPAQUE = 0;
```
guggero, line 138: nit: all enum values are in the same namespace that doesn't include the enum name itself. So we couldn't have another value called `OPAQUE` in a different enum. We should prefix the value, e.g. `META_TYPE_OPAQUE` to avoid future collisions.

tarorpc/taro.proto
```diff
@@ -732,3 +764,13 @@ message ExecuteSendStateEvent {
...
     string send_state = 2;
 }
+
+message FetchAssetMetaRequest {
+    oneof asset {
+        // The asset ID of the asset to fetch the meta for.
+        bytes asset_id = 1;
+
+        // The 32-byte meta hash of the asset meta.
+        bytes meta_hash = 2;
```
guggero, line 774: Same question here re: uniqueness of the hash.

tarodb/sqlc/queries/assets.sql
```diff
@@ -572,3 +572,33 @@ RETURNING script_key_id;
...
+    DO UPDATE SET meta_data_blob = EXCLUDED.meta_data_blob, 
+        meta_data_type = EXCLUDED.meta_data_type
+RETURNING meta_id;
+
+-- name: FetchAssetMeta :one
+SELECT meta_data_hash, meta_data_blob, meta_data_type
+FROM assets_meta
+WHERE meta_id = $1;
+
+-- name: FetchAssetMetaByHash :one
```
guggero, line 594: Is the hash really unique? If I don't specify one or use the zero length string, can't there be collisions in the hashes?

guggero (commented):
Very nice! I like the idea of adding a type to the meta data to allow for more standardized structure!
Main question is about uniqueness/optionality of the meta hash itself.

tarodb/asset_minting.go
```diff
@@ -130,7 +135,7 @@ type PendingAssetStore interface {
 	// FetchSeedlingsForBatch is used to fetch all the seedlings by the key
 	// of the batch they're included in.
 	FetchSeedlingsForBatch(ctx context.Context,
-		rawKey []byte) ([]AssetSeedling, error)
+		rawKey []byte) ([]sqlc.FetchSeedlingsForBatchRow, error)
```
jharveyb, line 138: I think we can avoid these intermediate structs by modifying the seedling model to align with the queries and vice versa - not strictly required though.

tarodb/asset_minting.go
```diff
@@ -487,6 +525,38 @@ func fetchAssetSprouts(ctx context.Context, q PendingAssetStore,
...
+
+	assetMetas := make(map[asset.SerializedKey]*proof.MetaReveal)
+	for _, assetT := range assets {
+		assetID := assetT.ID()
+
+		assetMeta, err := db.FetchAssetMetaForAsset(ctx, assetID[:])
+		switch {
+		// If the asset doesn't have a meta data reveal, then we can
+		// skip it.
+		case errors.Is(err, sql.ErrNoRows):
```
jharveyb, line 543: 👍🏽

tarodb/assets_store.go
```diff
@@ -1841,6 +1861,70 @@ func (a *AssetStore) QueryParcels(ctx context.Context,
 	return deltas, nil
 }
 
+// ErrAssetNotFound is returned when an asset meta is not found in the
+// database.
+var ErrAssetMetaFound = fmt.Errorf("asset meta not found")
```
jharveyb, line 1866: nit: Should be `ErrAssetMetaNotFound` ?

tarodb/sqlc/migrations/000002_assets.up.sql
jharveyb, line 40: I don't think this matches the behavior on the Golang side.

IIUC, we allowed empty metadata before - `[]byte{}` gets stored as `NULL`, which was allowed before.

Now I think an unset metadata hash would be either `[]byte{}` or all 0's. But I don't think we can use the zero hash across multiple assets because of the `UNIQUE` constraint?

I think maybe we'd want to map both []byte{} and the zero hash to NULL.

tarodb/sqlc/migrations/000002_assets.up.sql
jharveyb: I think to match the previous behavior, either this field can be NULL or the `meta_data_hash` can be NULL.

tarodb/sqlc/models.go
```diff
@@ -122,6 +122,13 @@ type AssetWitness struct {
 	SplitCommitmentProof []byte
 }
 
+type AssetsMetum struct {
```
jharveyb, line 125: Typo?

tarorpc/taro.proto
```diff
@@ -123,6 +129,32 @@ enum AssetType {
     */
     COLLECTIBLE = 1;
 }
+
+enum AssetMetaType {
+    /*
+    Opaque is used for asset meta blobs that have no true structure and instead
+    should be interested as opaque blobs.
```
jharveyb, line 136: typo, interested -> interpreted

jharveyb (commented):
Looking good so far! Happy to see how narrow these changes are.

Some notes on more input checking and DB specifics.

asset/asset.go
```diff
@@ -730,8 +730,7 @@ func (a *Asset) Copy() *Asset {
 	assetCopy := *a
 
 	// Perform a deep copy of all pointer data types.
-	assetCopy.Genesis.Metadata = make([]byte, len(a.Genesis.Metadata))
-	copy(assetCopy.Genesis.Metadata, a.Genesis.Metadata)
+	copy(assetCopy.Genesis.MetaHash[:], a.Genesis.MetaHash[:])
```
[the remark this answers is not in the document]
Roasbeef: Yep, removed!

proof/meta.go
```diff
@@ -0,0 +1,83 @@
...
+	var (
+		metaHash [asset.MetaHashLen]byte
+		b        bytes.Buffer
+	)
+
+	_ = m.Encode(&b)
+
+	h := sha256.New()
+	h.Write(b.Bytes())
+	h.Sum(metaHash[0:0])
```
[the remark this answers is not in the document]
Roasbeef: Sure, I actually had to fix a bug here since the `metaHash` originally didn't reslice. Simpler to just have it use `sha256.Sum256`.

proof/proof.go
```diff
@@ -63,6 +63,16 @@ type Proof struct {
 	// asset of the split.
 	SplitRootProof *TaprootProof
 
+	// MetaReveal is the set of bytes that were revealed to prove the
+	// derivation of the meta data hash contained in the genesis asset.
+	//
+	// TODO(roasbeef): use even/odd framing here?
```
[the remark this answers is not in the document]
Roasbeef: Yeah, basically if this should be even or odd or not. We'd need to go back and actually make this uniform if so. I think ultimately it makes sense to _force_ a reveal if the field is present, this way all the other merkle tree structures based on this would be canonical.

proof/mint.go
```diff
@@ -76,18 +76,53 @@ func encodeAsProofFile(proof *Proof) (Blob, error) {
 	return b.Bytes(), nil
 }
 
+// MintingLobOption allows the caller to modify how the final set of minting
```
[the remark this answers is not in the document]
Roasbeef: Fixed! Surprised our mega linter doesn't catch stuff like this tbh...

tarodb/sqlc/migrations/000002_assets.up.sql
Roasbeef: Yes, it's optional so a NULL can be inserted here. This would mean the asset has no meta data associated with it, and the `metaHash` is all zeroes. 

We wouldn't explicitly insert anything, since this uses the `NullInt` type on the Go side. The all zeroes meta has no meaning, so I don't think we need to insert it. Queries that determine if an asset has a meta or not, would check for NULL there, or do a join, which would return empty rows.

tarodb/asset_minting.go
Roasbeef: Yeah the main edge case in our style guide....

tarodb/assets_common.go
Roasbeef: So with the current logic, we'll end up inserting all zeroes for the hash+type, then return that primary key for `UpsertAssetMeta`. Those that have no meta then end up pointing towards that.

tarorpc/taro.proto
Roasbeef: Hmm, I thought (at least in Go), we'd be seeing `AssetMetaType_OPAQUE`?

tarodb/sqlc/queries/assets.sql
Roasbeef: If you don't specify one, then you have the all zero meta hash. That leads to nothing (empty byte slice). 

If you use a zero length string, than that's still a meta, which leads to a different meta hash. We now require that the reveal be mandatory for genesis proofs.

cmd/tarocli/assets.go
```diff
@@ -94,10 +107,39 @@ func mintAsset(ctx *cli.Context) error {
...
+		assetMeta = &tarorpc.AssetMeta{
+			Data: []byte(ctx.String(assetMetaBytesName)),
+			Type: tarorpc.AssetMetaType(ctx.Int(assetMetaTypeName)),
+		}
+
+	case ctx.String(assetMetaFilePathName) != "":
+		metaPath := tarocfg.CleanAndExpandPath(
+			ctx.String(assetMetaFilePathName),
+		)
+		metaFileBytes, err := ioutil.ReadFile(metaPath)
```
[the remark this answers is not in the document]
Roasbeef: Yeah I think we can use an `io.LimitReader`, but the main question is: how much should we limit them to? Looks like gRPC has a limit of 2 GB in message size, so maybe then 1 GB or so for this?

cmd/tarocli/assets.go
```diff
@@ -56,8 +60,17 @@ var mintAssetCommand = cli.Command{
 			Usage: "the target supply of the minted asset",
 		},
 		cli.StringFlag{
-			Name:  assetMetaName,
-			Usage: "the metadata associated with the asset",
+			Name:  assetMetaBytesName,
```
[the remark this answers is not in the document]
Roasbeef: Yeah we want to accept the meta then make sure the serialization matches, since that's what the verifier will be checking.

proof/mint.go
```diff
@@ -76,18 +76,53 @@ func encodeAsProofFile(proof *Proof) (Blob, error) {
...
+	}
+}
+
+// WithAssetMetaReveals is a MintingBlobOption that allows the caller to attach
+// meta reveal information to the initial minting blob created.
+func WithAssetMetaReveals(metaReveals map[asset.SerializedKey]*MetaReveal,
+) MintingBlobOption {
+
+	return func(o *mintingBlobOpts) {
+		o.metaReveals = metaReveals
```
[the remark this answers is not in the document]
Roasbeef: As is yes. So you mean we should try to _merge_ the maps? 

FWIW, at the one call site, the caller prepares things for the entire minting batch.

rpcserver.go
```diff
@@ -1359,3 +1364,47 @@ func unmarshalKeyDescriptor(
...
+// either by the asset ID for that asset, or a meta hash.
+func (r *rpcServer) FetchAssetMeta(ctx context.Context,
+	in *tarorpc.FetchAssetMetaRequest) (*tarorpc.AssetMeta, error) {
+
+	var (
+		assetMeta *proof.MetaReveal
+		err       error
+	)
+	switch {
+	case in.GetAssetId() != nil:
```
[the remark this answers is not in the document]
Roasbeef: Added!

tarodb/asset_minting.go
Roasbeef: Usually we type def these, but I left it in for now. Happy to change though.

tarodb/sqlc/migrations/000002_assets.up.sql
Roasbeef: It would be a `[32]byte`, I pushed a modified version of one of the commits to test this logic explicitly.

tarodb/sqlc/models.go
Roasbeef: Nah it does this name mangling thing to try to pluralize the table name: person goes to persons, etc. In this case, idk what it did lol (this is generated code).

tarorpc/taro.proto
guggero: Yes, and that's why it's annoying in Go, since it becomes `AssetMetaType_META_TYPE_OPAQUE`. But on the gRPC level you get a startup panic if you try to load a package that has multiple enum value names that collide.

Roasbeef: Ah makes sense, will add the suffix so things won't break in other langs.

tarodb/sqlc/migrations/000002_assets.up.sql
Roasbeef: Ok so with the way things are rn: we always insert the "all zero" meta on disk. As a result, anything that doesn't have a meta, will end up pointing to that (dedup).

tarodb/sqlc/queries/assets.sql
Roasbeef: Re ppl messing with the hash, anyone can do something like: take a pdf that was time stamped, flip a single bit, then get a diff hash. For the length prefix, we use TLV so things are always canonical. So the same artifact will always produce the same hash, unless someone has messed with it.

cmd/tarocli/assets.go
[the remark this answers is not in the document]
jharveyb: Maybe 25 MB, common email attachment size limit? And later allow massive preimages with some flag on startup.

Not strongly attached to 25.

tarodb/sqlc/migrations/000002_assets.up.sql
jharveyb: In that case can't this be `NOT NULL` now?

proof/encoding.go
```diff
@@ -447,3 +447,46 @@ func BoolDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
...
+		}
+		var reveal MetaReveal
+		err := reveal.Decode(bytes.NewReader(revealBytes))
+		if err != nil {
+			return err
+		}
+		*typ = &reveal
+		return nil
+	}
+	return tlv.NewTypeForEncodingErr(val, "MetaReveal")
```
guggero, line 472: nit: `*MetaReveal` to match above?

tarodb/sqlc/migrations/000002_assets.up.sql
```diff
@@ -247,7 +262,7 @@ CREATE TABLE IF NOT EXISTS asset_seedlings (
 
     asset_supply BIGINT NOT NULL,
 
-    asset_meta BLOB,
+    asset_meta_id INTEGER REFERENCES assets_meta(meta_id),
```
guggero, line 264: Same here re `NOT NULL` as discussed above?

tarodb/asset_minting.go
```diff
@@ -535,6 +573,36 @@ func fetchAssetSprouts(ctx context.Context, q PendingAssetStore,
 	return taroCommitment, nil
 }
 
+// fetchAssetMetas attempts to fetch the asset meta reveal for each of the
+// passed assets.
+func fetchAssetMetas(ctx context.Context, db PendingAssetStore,
+	assets []*asset.Asset) (map[asset.SerializedKey]*proof.MetaReveal, error) { // nolint:lll
```
guggero, line 579: nit: break before `error` return type instead of adding nolint directive?

tarogarden/planter.go
```diff
@@ -194,6 +194,15 @@ func (c *ChainPlanter) Start() error {
 			log.Infof("Launching ChainCaretaker(%x)",
 				batch.BatchKey.PubKey.SerializeCompressed())
 
+			// For batches before the actual assets have been
+			// committed, we'll need to populate this field
+			// manually.
+			if batch.AssetMetas == nil {
+				batch.AssetMetas = make(
+					map[asset.SerializedKey]*proof.MetaReveal,
```
guggero, line 202: We could create a type alias for this to make handling a bit easier.

guggero (approved):
LGTM :tada:

tarodb/asset_minting.go
Roasbeef: Looks weird visually imo, but done.

guggero: Could now also use the new `AssetMetas` type to shorten the function definition.

PR #263, opened 2023-02-10
psbt script spend [4/3]: interactive full value send

guggero:
Work in progress, only last 5 commits are new.


guggero:
@dstadulis, @Roasbeef looking for an ACK to invest time into refactoring the DB to be more generic in terms of transfers/parcels (to support both interactive full value sends as well as potential multi-receiver sends). This PoC shows that interactive full value sends are possible but need adjustments to the storage logic.

dstadulis:
>@dstadulis, @Roasbeef looking for an ACK to invest time into refactoring the DB to be more generic in terms of transfers/parcels

Recapping another convo, we will go ahead with prioritizing improving the DB schema to accommodate transfers with mutliple senders and recipients e.g. more generalized design

PR #267, opened 2023-02-14
send: re-anchor non-input assets

ffranr:
Associated issue: https://github.com/lightninglabs/taro/issues/241


taropsbt/interface.go
```diff
@@ -78,6 +78,11 @@ type VPacket struct {
 	// and Interactive being set to true.
 	Outputs []*VOutput
 
+	// StaleAssets is a list of all stale assets. A stale asset is a
+	// non-input assets which require re-signing and re-anchoring as a
+	// result of the spend.
+	StaleAssets []*asset.Asset
```
Roasbeef, line 84: Don't think we want this to live here, as these are sort of "implicit spends". The other party doesn't need to know about these.

tarofreighter/wallet.go
```diff
@@ -313,6 +318,57 @@ func (f *AssetWallet) SignVirtualPacket(vPkt *taropsbt.VPacket) ([]uint32,
...
+				// also the sender. We therefore set interactive
+				// to true to indicate that the receiver is
+				// aware of the transfer.
+				Interactive: true,
+
+				AnchorOutputIndex: 0,
+				ScriptKey:         a.ScriptKey,
+				Asset:             passiveAssets[idx].Copy(),
+			}},
+		}
```
ffranr, line 366: I think after this I still need to do:
```
vPackets[idx].SetInputAsset(0, a, nil)
```
right?

But instead of `nil` I suppose I need to derive and pass through the inclusion proof?

ffranr: No need for the proof in the end. Which makes sense because this vpacket isn't sent to anyone. I figured the proof might have been necessary for validation. But that wasn't the case either. The VM was complaining about something else.

taropsbt/interface.go
ffranr: Removed :+1:

PR #268, opened 2023-02-16
server: stop returning an error that is always nil

positiveblue:
The `NewServer` function signature is `func(cfg *Config) (*Server, error)` but the error is always `nil`. Even in other projects like `loop`, `pool` or `lnd` we did not find a case were an error needed to be returned.

We can delete it from the signature from now and add it in the future (or in a wrapped function) if needed.

### DIFF
--- a/tarocfg/server.go
+++ b/tarocfg/server.go
@@ -142,7 +142,7 @@ func CreateServerFromConfig(cfg *Config, cfgLogger btclog.Logger,
 		}
 	}
 
-	server, err := taro.NewServer(&taro.Config{
+	server := taro.NewServer(&taro.Config{
 		DebugLevel:  cfg.DebugLevel,
 		ChainParams: cfg.ActiveNetParams,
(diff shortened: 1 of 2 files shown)


tarocfg/server.go
```diff
@@ -208,9 +208,6 @@ func CreateServerFromConfig(cfg *Config, cfgLogger btclog.Logger,
 			TaroAddrBook: tarodbAddrBook,
 		},
 	})
-	if err != nil {
-		return nil, fmt.Errorf("unable to start server: %v", err)
-	}
 
 	return server, nil
```
positiveblue, line 212: we could return directly where we define server, but given that it is a long function call (config is defined in-place) I left it here.

guggero (approved):

jharveyb (approved):
Nice catch 👍🏽

guggero:
The merge queue doesn't work for some reason. I can't merge, @Roasbeef can you please push the button when you have a minute?

Roasbeef:
Trying the merge queue again now that this is in: https://github.com/lightninglabs/taro/pull/269

Issue #272, opened 2023-02-21
Create `list batches` funcitonality to enumerate asset-update batches


dstadulis:
This tracking issue will be fulfilled by a small PR to create `list batches` funcitonality to enumerate asset-update batches within Taro.  Purpose of the small PR is to get the feature into `main` quickly for it to be built on for other feature branches (avoiding rebase churn)

jharveyb:
Fixed with https://github.com/lightninglabs/taproot-assets/releases/tag/v0.2.0

Issue #266, opened 2023-02-14
Currencies getting lost when minting multiple currencies with skip_batch in one block


snow884:
When I try to minting multiple currencies in one block currencies I submitted for minting get "lost" and never appear
```
for i in {1..20}; do tarocli assets mint --type=normal --name=NewCurrencyFromLoop$i --meta="test minting number $i" --skip_batch --supply=10000$i; done
```

This returns:
```
{
    "batch_key": "036099a9139325a058325b1754967139ef50b21052cc3aef7a9dc95914981c4765"
}
{
    "batch_key": "02a51a403ce9b5dfa70d9e1ade216a52d48eea551f4aa194699cd8f80eb6db0fd1"
}
{
    "batch_key": "0300658b33d38d598c14f33305e2b905c1a5b45718bba010a9fada9f482a303cb2"
}
{
    "batch_key": "02b139344bccb3e63ab1178560adb35fd31672f5ef555f40ec25d97de067dd282b"
}
...
```

Wait for a 5 blocks as that should be long enough even on testnet.

```
$ tarocli assets balance | grep NewCurrencyFromLoop
```

only gives
```
"name": "NewCurrencyFromLoop1"
```

So all of the 20 currencies that were minted got lost.

I know that there is the options to enable batches but still it happens to be a lot that I have to submit multiple mint commands in one block.

dstadulis:
## Summary
hey @snow884 if I understand your intention you'd like to mint assets `NewCurrencyFromLoop1..20` in a single block.

Your issue may be described as:
Calling `--skip_batch` after a minting call, but before a minting txn is included in a block, the daemon doesn't rewrite a bitcoin transaction with the subsequent minting events.

## Fixes
### Quick
The short term amelioration would be: remove `--skip_batch` from your loop.  Then all the minting transactions will be grouped together in a minting batch which ought to included `NewCurrencyFromLoop1..20` 

### Long Term
Longer term fix could be to either
1. Throw a warning when calling `mint` between a `mint` call but before a block which includes the first mint is reeived
2. Modify the minting function so that it rewrites the minting transaction if a subsequent `mint --skip_batch` is called to include the subsequent minting
3. Modify the minting function so that it uses a different UTXO to mint subsequent calls (I need to check on where the delta would be on this work)

snow884:
Thank you for quick reply @dstadulis  . 

I will try minting the currencies without `--skip_batch`

Roasbeef:
[...]

Issue #252, opened 2023-02-02
minting: expose means to force a batch without adding an asset


jharveyb:
Related to #251.

Right now we have all the needed code to force the batch ticker, but no way to trigger that without also adding an asset to a batch. A user may decide to force a batch after adding all desired assets, which is currently not supported.

Issue #251, opened 2023-02-02
minting: expose method to cancel a pending batch


jharveyb:
Related to #233.

Right now minting batches are finalized based on a timer, or the user forcing a batch. If a mistake is made specifying the new assets, the user can't easily cancel that request (even a `tarod` restart may not work). The planter (and RPC and CLI) should expose a means to cancel a pending batch.

Once the timer has fired and the caretaker is in the process of finalizing the batch, cancellation becomes more difficult - it may be useful to show the 'time until next tick' so users can cancel early if needed.

jharveyb:
Fixed with https://github.com/lightninglabs/taproot-assets/releases/tag/v0.2.0

PR #274, opened 2023-02-21
minting: Add Finalize, Cancel, and ListBatches methods

jharveyb:
Addresses #251, #252, and #272. Should address user reports like #266. Part of #233.

In this PR we make batched asset minting more flexible by exposing means to finalize a batch without modifying the batch, cancel a pending batch, and display the contents of the pending batch.

This also allows users to mint batches without waiting for the minting timer to tick. Separating modifying a batch and 'minting commands' like finalizing or cancelling a batch also allows for a simpler user flow when minting a batch manually.

TODOs:

- [ ] Add ListBatches RPC
- [ ] Add itest case with cancellation and batch listing
- [ ] Enforce unique seedling name constraint in the planter instead of in the DB


tarodb/asset_minting.go
```diff
@@ -600,7 +600,8 @@ func (a *AssetMintingStore) FetchNonFinalBatches(
 			// sprouts (full defined assets, but not yet mined).
 			switch batches[i].BatchState {
 			case tarogarden.BatchStatePending,
-				tarogarden.BatchStateFrozen:
+				tarogarden.BatchStateFrozen,
+				tarogarden.BatchStateCancelled:
```
guggero, line 604: nit: add newline after multi-line `case`.

tarogarden/planter.go
```diff
@@ -260,6 +263,27 @@ func freezeMintingBatch(ctx context.Context, pLog MintingStore,
...
+func cancelMintingBatch(ctx context.Context, pLog MintingStore,
+	batch *MintingBatch) (*btcec.PublicKey, error) {
+
+	batchKey := batch.BatchKey.PubKey
+	log.Infof("Cancelling MintingBatch(key=%x, num_assets=%v",
+		batchKey.SerializeCompressed(), len(batch.Seedlings))
+
+	// In order to cancel a batch, we need to update the state of the batch
+	// to BatchStateCancelled.
+	err := pLog.UpdateBatchState(
```
guggero, line 277: nit: looks like this could fit on a single line.

tarogarden/planter.go
```diff
@@ -260,6 +263,27 @@ func freezeMintingBatch(ctx context.Context, pLog MintingStore,
 	)
 }
 
+// cancelMintingBatch cancels a target minting batch which means that the batch
+// will not be passed to a caretaker.
+func cancelMintingBatch(ctx context.Context, pLog MintingStore,
```
guggero, line 268: nit: s/pLog/store/?

rpcserver.go
```diff
@@ -302,24 +302,50 @@ func (r *rpcServer) DebugLevel(ctx context.Context,
...
+	if req.Asset == nil && !req.FinalizeBatch {
+		return nil, fmt.Errorf("asset is required")
+	}
+
+	// Finalizing a batch implies not adding another asset to the batch.
+	if req.FinalizeBatch && req.Asset != nil {
+		return nil, fmt.Errorf("cannot finalize batch and add asset")
+	}
+
+	if req.FinalizeBatch {
```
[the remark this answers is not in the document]
jharveyb: Yeah, I was thinking that these 'batch commands' should really be separate on the RPC and CLI - just realized that we can add actual subcommands to the 'mint' subcommand instead of using flags.

So I can then have separate functions in `rpcserver` and add new endpoints + minting subcommands.

PR #278, opened 2023-05-16
commitment: add asset type field to asset commitment structure

ffranr:
All assets included within a given asset commitment must have the same asset type. This change helps to reinforce that idea.


commitment/asset.go
```diff
@@ -98,11 +101,23 @@ func parseCommon(assets ...*asset.Asset) (*AssetCommitment, error) {
...
 	}
 
+	var assetType *asset.Type
 	maxVersion := asset.Version(0)
 	assetGenesis := assets[0].Genesis.ID()
 	assetGroupKey := assets[0].GroupKey
 	assetsMap := make(CommittedAssets, len(assets))
 	for _, asset := range assets {
+		// Set the asset type if we are handling the first asset.
+		if assetType == nil {
```
guggero, line 111: Looking at this rule, wouldn't it make more sense to check the `TapCommitmentKey` instead? Since only assets with the same top level key can be in the same asset commitment (the type is indirectly part of the asset ID of that key or if grouped then the minting process should make sure only assets of the same type can be grouped together).

guggero (commented):
Nice quality of life improvement. But I actually think that we should check on the `TapCommitmentKey` instead, which includes the asset ID (or group key) which should both already be covered by type checks (or maybe worth taking a quick look at whether we prevent different types of assets being minted into the same group).

commitment/asset.go
ffranr: Good point. I've added a check for consistent `TapCommitmentKey`.

I've left in the asset type check. I figured I need to pickup the asset type from one asset anyway, so I might as well check it against the other assets. We have an error type for asset type mismatch and I figured the check would be more specific and easier to debug than matching `TapCommitmentKey`. Let me know what you think!

guggero (approved):

dstadulis:
quickly funnel assets into their proper commitment space

jharveyb (approved):
LGTM, was originally thinking that we would want to encode the type as well but I now think that's not needed. 🚀 

Second commitment message needs to use a different acronym though,

Issue #282, opened 2023-05-17
rpc: use hex encoding for all REST RPC outputs 


Roasbeef:
Right now for the REST calls, we'll return base64 for all the values, which isn't super easy to handle for devs. 

We have a [fork](https://github.com/lightninglabs/protobuf-go-hex-display) of the protobuf lib that allows us to do this, here's an example in lnd: https://github.com/lightningnetwork/lnd/blob/2fd40d16f1ec7a93d17870d27dfc3d488d34a7c9/lnrpc/rpc_utils.go#L40-L47

PR #286, opened 2023-05-19
Use hex encoding for raw bytes in REST API

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/282.


taprpc/taprpc_utils.go
```diff
@@ -0,0 +1,42 @@
...
+		UseProtoNames:   true,
+		UseHexForBytes:  true,
+	}
+
+	// RESTJsonUnmarshalOpts is a struct that holds the default unmarshal
+	// options for un-marshaling REST JSON into protobuf messages. This
+	// should be used when interacting with the REST proxy only.
+	RESTJsonUnmarshalOpts = &protojson.UnmarshalOptions{
+		AllowPartial:   false,
+		UseHexForBytes: true,
```
positiveblue, line 40: q: I guess it's to late to change this in [LND](https://github.com/lightningnetwork/lnd/blob/ac3e7be9a063bdb559855f3b5af15e0323878450/lnrpc/rpc_utils.go#L44) too because other tools are using already base64 encoding?

positiveblue (approved):

go.mod
```diff
@@ -203,3 +202,7 @@ require (
 	modernc.org/token v1.0.1 // indirect
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
+
+// We want to format raw bytes as hex instead of base64. The forked version
+// allows us to specify that as an option.
+replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-display v1.30.0-hex-display
```
Roasbeef, line 208: Why not just import it directly?

taprpc/taprpc_utils.go
guggero: I thought about adding an option to give the user the choice. But the problem is that yes, most of the apps using the REST interface now rely on the raw bytes being base64 encoded. So it shouldn't be the user's choice but rather the developer's. This gives me an idea though: Maybe there is a way to specify the desired encoding as a HTTP header field? For example: `Accept: application/json+hex` instead of just `Accept: application/json`...

go.mod
guggero: Because we need to keep it as the original package name, if I import it directly I get this error:
```
        module declares its path as: google.golang.org/protobuf
                but was required as: github.com/lightninglabs/protobuf-go-hex-display
```

And renaming the package in the fork would require a way bigger diff than just https://github.com/protocolbuffers/protobuf-go/commit/9c1f7c5b64c15ea4ffc186c309df4bf73680f1f2 as we'd need to refactor a bunch of usages of the `internal` package as well. So rebasing the hex render feature on top of a new version in the future would become a bigger task, potentially causing us to lag behind on the versions of this package.

Roasbeef (approved):
LGTM 🔮

Roasbeef (approved):
LGTM 🔮

PR #316, opened 2023-05-23
tapcfg: update default testnet universe port

guggero:
Updates the port in the default testnet universe federation server address. The server will support both ports for a while until this change has been released with a proper version for a while.


Roasbeef (approved):
LGTM 🔧

PR #318, opened 2023-05-23
BIP: re-add submodule for BIPs, remove local copy from docs folder

guggero:
Now that we have an updated branch for the Taproot Assets BIP (https://github.com/Roasbeef/bips/tree/bip-tap), we can re-add that as a git submodule and remove the local copies in the `docs/` folder.


dstadulis (approved):
LGTM
Are there any modifications we should make to fetch the BIP info?

>You must run two commands: git submodule init to initialize your local configuration file, and git submodule update to fetch all the data from that project and check out the appropriate commit listed in your superproject:

https://git-scm.com/book/en/v2/Git-Tools-Submodules#_cloning_submodules

guggero:
>  Are there any modifications we should make to fetch the BIP info?

I think just the normal `git submodule init` should be enough.

jharveyb (approved):
LGTM 💯 

Verified with `diff $MAIN/docs/ $BRANCH/bips/`

I had to run `git submodule update` instead of init, not sure why. I think we want to update the README to use `git clone --recurse-submodules` on L52, otherwise this dir will be empty.

PR #319, opened 2023-05-23
mint: fix REST encoding of mint RPC

guggero:
This is a follow-up to #286 which allowed developers to specify any raw
bytes as a hex string in the REST API (instead of the library default of
base64).
The only area where that change didn't apply is the URI part of REST
calls. For those we need to use a string field to support hex encoding,
the same way we already did for the universe REST API.


jharveyb (approved):
LGTM - IIUC we already display these fields in hex so I think copy & paste should continue working.

taprpc/universerpc/universe.yaml
```diff
@@ -8,35 +8,30 @@ http:
 
     - selector: universerpc.Universe.QueryAssetRoots
       get: "/v1/taproot-assets/universe/roots/asset-id/{id.asset_id_str}"
-
-    - selector: universerpc.Universe.QueryAssetRoots
-      get: "/v1/taproot-assets/universe/roots/group-key/{id.group_key_str}"
+      additional_bindings:
```
Roasbeef, line 11: 👍

Roasbeef (approved):
LGTM 🧊

PR #321, opened 2023-05-24
Migrate deprecated "io/ioutil" calls to "os" in project from Go 1.16/1.17 to 1.19

ben2077:
This commit upgrades the project's Go version to 1.19 and addresses the deprecated "io/ioutil" calls that were present in Go versions 1.16 and 1.17. The deprecated calls have been migrated to their respective counterparts in the "os" package.


cmd/tapcli/assets.go
```diff
@@ -3,12 +3,11 @@ package main
 import (
 	"encoding/hex"
 	"fmt"
-	"io/ioutil"
-
 	"github.com/lightninglabs/taproot-assets/tapcfg"
 	"github.com/lightninglabs/taproot-assets/taprpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
 	"github.com/urfave/cli"
+	"os"
```
ffranr, line 10: I think `"os"` should be grouped with the entries before the `"github.com/..."` entries. Which is the pattern we use in the rest of the codebase. However, the linter would pass either way.

```
import (
	"encoding/hex"
	"fmt"
        "os"

	"github.com/lightninglabs/taproot-assets/tapcfg"
	"github.com/lightninglabs/taproot-assets/taprpc"
	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
	"github.com/urfave/cli"
)
```

cmd/tapcli/assets.go
```diff
@@ -3,12 +3,11 @@ package main
...
 )
```
ffranr: I think `"os"` should be grouped with the entries before the `"github.com/..."` entries. Which is the pattern we use in the rest of the codebase. However, the linter would pass either way.

```
import (
	"encoding/hex"
	"fmt"
        "os"

	"github.com/lightninglabs/taproot-assets/tapcfg"
	"github.com/lightninglabs/taproot-assets/taprpc"
	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
	"github.com/urfave/cli"
)
```

ffranr (changes requested):
Looks good! :partying_face: Thank you for your contribution.

The commit message should probably be:
```
multi: replace deprecated uses of "io/ioutil" with "os"

Some uses of package "io/ioutil" are deprecated since Go version 1.19.
In this commit, the deprecated calls have been migrated to their
respective counterparts in the "os" package.
```

which limits the line length of the commit message to 72 columns. It also includes a short subject line.

cmd/tapcli/assets.go
ffranr: Saw that you've updated :+1: . Commits need to be squashed into one.

ffranr (approved):
Fantastic! Nice work.

guggero (approved):

PR #322, opened 2023-05-24
Separate usages of "math/rand" and "crypto/rand" in single Go file using package alias

ben2077:
This commit refactors a single Go file to differentiate between the usage of "math/rand" and "crypto/rand". A package alias, "crand", is introduced exclusively for importing "crypto/rand". This separation is employed due to the superior implementation of the "read" function in "crypto/rand".

### DIFF
--- a/internal/test/helpers.go
+++ b/internal/test/helpers.go
@@ -1,6 +1,7 @@
 package test
 
 import (
+	crand "crypto/rand"
 	"math/rand"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


jharveyb:
Bit confused about the intent here, all the `rand.Read()` calls are in tests, and usually seeded with a timestamp. I think the timestamp seeding is useful to potentially replicate test runs if needed, thouugh I don't think we've used that so far.

FWIW, the only use of `crypto/rand` I see in the repo is here: https://github.com/lightninglabs/taproot-assets/blob/4f22232d8a972e6b866b4fe5822d8598fb297c4e/tapdb/macaroons.go#L145

Which could be replaced with `rand.Read(rootKey)`.

Roasbeef:
CI run approved.

jharveyb:
Won't use `crypto/rand` in test helpers now since we've added deterministic test vector generation, so closing this.

PR #325, opened 2023-05-25
itest: add optional 10k asset test

jharveyb:
Adds support for optional itests, starting with a stress test of the minter with a batch of 10,000 assets, each with 4kB of metadata.

Fixes https://github.com/lightninglabs/taproot-assets/issues/361.


Makefile
```diff
@@ -172,9 +172,21 @@ itest-trace: build-itest itest-only-trace
...
 	rm -rf itest/regtest; date
-	$(GOTEST) ./itest -v -tags="$(ITEST_TAGS)" $(TEST_FLAGS) $(ITEST_FLAGS) -btcdexec=./btcd-itest -logdir=regtest
+	$(GOTEST) ./itest -v -tags="$(ITEST_TAGS)" $(TEST_FLAGS) $(ITEST_FLAGS) -test.short -btcdexec=./btcd-itest -logdir=regtest
 
 itest-only-trace: aperture-dir
+	@$(call print, "Running integration tests with ${backend} backend.")
+	rm -rf itest/regtest; date
+	$(GOTEST) ./itest -v -tags="$(ITEST_TAGS)" $(TEST_FLAGS) $(ITEST_FLAGS) -test.short -loglevel=trace -btcdexec=./btcd-itest -logdir=regtest
+
+optionalitest: build-itest optional-itest-only
```
guggero, line 182: nit: `optional-itest`?

guggero (commented):
Looks good! I like the idea with the two test lists, probably a bit easier to understand (and also to implement) than the build tag variant.

jharveyb:
> Looks good! I like the idea with the two test lists, probably a bit easier to understand (and also to implement) than the build tag variant.

Another motivating reason was that trying to use mutually exclusive build tags to swap out the test list caused some IDE issues with gopls, which could cause issues later on.

itest/assertions.go
```diff
@@ -92,55 +94,90 @@ func assetScriptKeyIsLocalCheck(isLocal bool) assetCheck {
 	}
 }
 
+// sortAssetsByName converts an unordered list of assets to a map of lists of
+// assets, where all assets in a list have the same name.
+func sortAssetsByName(assets []*taprpc.Asset) map[string][]*taprpc.Asset {
+	assetLists := make(map[string][]*taprpc.Asset)
+	for _, a := range assets {
+		if _, ok := assetLists[a.AssetGenesis.Name]; !ok {
+			assetLists[a.AssetGenesis.Name] = []*taprpc.Asset{}
```
guggero, line 104: I don't think we need to pre-initialize this, since `append()` will work even if the first argument is `nil`.

itest/assertions.go
```diff
@@ -92,55 +94,90 @@ func assetScriptKeyIsLocalCheck(isLocal bool) assetCheck {
...
 // additional checks on that asset.
-func assertAssetState(t *harnessTest, tapd *tapdHarness, name string,
-	metaHash []byte, assetChecks ...assetCheck) *taprpc.Asset {
+func assertAssetState(t *harnessTest, assets map[string][]*taprpc.Asset,
+	name string, metaHash []byte, assetChecks ...assetCheck) *taprpc.Asset {
 
-	t.t.Helper()
+	var a *taprpc.Asset
+	matchingAssets, ok := assets[name]
+	require.True(
```
guggero, line 123: Could use `require.Contains()` to avoid needing to add additional error context.

itest/assertions.go
```diff
@@ -92,55 +94,90 @@ func assetScriptKeyIsLocalCheck(isLocal bool) assetCheck {
...
+// waitForBatchState polls until the planter has reached the desired state with
+// the current batch.
+func waitForBatchState(t *harnessTest, ctx context.Context, tapd *tapdHarness,
+	timeout time.Duration, targetState mintrpc.BatchState) bool {
 
-				break
-			}
-		}
+	breakTimeout := time.After(timeout)
+	ticker := time.NewTicker(50 * time.Millisecond)
```
guggero, line 153: Should we extract all of this into a copy of `wait.NoError` and `wait.Predicate` analogues that allow us to specify the poll interval instead? In the `fn` package for example?

itest/assertions.go
```diff
@@ -92,55 +94,90 @@ func assetScriptKeyIsLocalCheck(isLocal bool) assetCheck {
 	}
 }
 
+// sortAssetsByName converts an unordered list of assets to a map of lists of
+// assets, where all assets in a list have the same name.
+func sortAssetsByName(assets []*taprpc.Asset) map[string][]*taprpc.Asset {
```
guggero, line 100: This doesn't really sort but rather group?

itest/multi_asset_group_test.go
```diff
@@ -201,6 +177,28 @@ func testMintMultiAssetGroups(t *harnessTest) {
 	)
 }
 
+// Verify that the correct asset was used as the group anchor by re-deriving the
```
guggero, line 180: nit: Godoc name mismatch.

itest/mint_batch_stress_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	correctOp := fn.All(uniKeys.AssetKeys, func(key *unirpc.AssetKey) bool {
+		return key.GetOpStr() == mintOutpoint
+	})
+	require.True(t.t, correctOp)
+
+	// If we create a second tapd instance and sync the universe state,
+	// the synced tree should match the source tree.
+	bob := setupTapdHarness(
+		t.t, t, t.lndHarness.Bob, nil,
```
guggero, line 168: Can't we replace the code that adds the federation server by just adding
```
		func(params *tapdHarnessParams) {
			params.startupSyncNode = t.tapd
			params.startupSyncNumAssets = len(rpcAssets)
		},
```
here?

itest/assertions.go
jharveyb: Yes, the intent was to separate waiting for the minter to reach a certain state, and asserting properties over the batch.

The slowdown before was caused by two features of the previous code:

- Using `wait.NoError` to actually wait for the planter state, meant we also used a `wait.NoError` call in every call of `assertAssetState`.
- `assertAssetState` called `ListAssets()`, so we would also call that RPC for each entry in the batch.

So, for asserting properties over larger batches (even just 100 assets), this caused the whole test to run much slower.

The intent now is to wait for the desired batch state, call `ListAssets()` once, and then run the assertions over the results without any waiting logic or extra RPC calls.

`ListAssets()` actually still slows down as batch size increases, but that's sort of a separate issue. `waitForBatchState` is very similar to `waitForNTXsInMempool`, I agree that this pattern could be moved into `fn`.

itest/assets_test.go
```diff
@@ -154,12 +157,17 @@ func mintAssetsConfirmBatch(t *harnessTest, tapd *tapdHarness,
 	require.NoError(t.t, err)
 	require.NotEmpty(t.t, batchResp.BatchKey)
 
+	mempoolWaitTimeout := defaultWaitTimeout
```
guggero, line 160: This should probably also be set on the context timeout above. But the real thing I ran into when running this locally is this: https://github.com/lightninglabs/taproot-assets/blob/0a58cdbf873f4634f05aa6f40fa75f888f0f5195/tapgarden/caretaker.go#L805
This will cancel the context after 1 minute. So we should probably make that `WithCtxQuitNoTimeout`...

guggero:
I pushed up a couple of fixes and confirmed the 1k test succeeds locally on my machine both with SQLite and Postgres. The 10k test still takes too long to finish within the 60m test timeout, so I think we first need to get some of the optimizations in.
But IMO all the fixes in this PR are enough to justify getting it in as it currently is.

So there are multiple fixes in here that address https://github.com/lightninglabs/taproot-assets/issues/361.

tapdb/sqlc/queries/universe.sql
```diff
@@ -87,11 +87,16 @@ SELECT * FROM universe_servers;
 
 -- name: InsertNewSyncEvent :exec
 WITH root_asset_id AS (
-    SELECT leaves.universe_root_id AS id
+    SELECT DISTINCT leaves.universe_root_id AS id
     FROM universe_leaves leaves
     JOIN genesis_info_view gen
         ON leaves.asset_genesis_id = gen.gen_asset_id
-    WHERE gen.asset_id = @asset_id
+    RIGHT OUTER JOIN key_group_info_view groups
```
Roasbeef, line 94: So I think we can simplify this, and just target the group key directly. 

Retracing a bit of history here, the query first looked like this: https://github.com/lightninglabs/taproot-assets/blame/c037025ebb22c54087736589bd31fb570a04b738/tapdb/sqlc/queries/universe.sql#L101-L105

The issue with that is that the `asset_id` field there for assets with a group key is basically whichever asset was inserted first. As a result, if you tried to log the sync of one of those leaves, then things fail as maybe that was the wrong leaf. 

We then changed it to target the leaf instead (since we know the asset id of the leaf), then use that to get the pointer ID back to the root. 

However if we have the group key, then we can just target that diretly. So we can have a query that looks similar to the old one: 
```sql
    SELECT id
    FROM universe_roots
    WHERE group_key = @group_key
```

So with this, we don't need to modify that other view, we can just combine the queries and pick whichever doesn't return `NULL`
```sql
SELECT COALESCE(
    (SELECT leaves.universe_root_id AS id
    FROM universe_leaves leaves
    JOIN genesis_info_view gen
        ON leaves.asset_genesis_id = gen.gen_asset_id
    WHERE gen.asset_id = @asset_id),
    (SELECT id
    FROM universe_roots
    WHERE group_key = @group_key)
) AS id;
```

jharveyb:
LGTM but not approving b/c original author.

tapgarden/caretaker.go
```diff
@@ -533,7 +533,7 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
 		//
 		// TODO(roasbeef): need to invalidate asset creation if on
 		// restart leases are gone
-		ctx, cancel := b.WithCtxQuit()
+		ctx, cancel := b.WithCtxQuitNoTimeout()
```
ffranr, line 536: Perhaps the timeout should still exist but be proportional to the amount of work to complete. Future work for another PR perhaps.

tapdb/sqlc/queries/universe.sql
```diff
@@ -103,31 +103,53 @@ WHERE server_host = @target_server;
 SELECT * FROM universe_servers;
 
 -- name: InsertNewSyncEvent :exec
-WITH root_asset_id AS (
+WITH group_key_root_id AS (
+    SELECT id
+    FROM universe_roots
+    WHERE group_key = @group_key_x_only
```
ffranr, line 109: In the interest of clarity: rather than `x_only` should we use `pub` here to designate this as the public key? Or does the public key require something more by convention than what's on the x-axis?

ffranr (approved):
Nice work guys!

universe/base.go
```diff
@@ -241,24 +241,26 @@ func (a *MintingArchive) FetchIssuanceProof(ctx context.Context, id Identifier,
...
+					"key=%v): %v", id.StringForLog(),
 					spew.Sdump(key), err)
 			}
 		}()
 	}()
 
-	return withBaseUni(a, id, func(baseUni BaseBackend) ([]*IssuanceProof, error) {
-		return baseUni.FetchIssuanceProof(ctx, key)
-	})
+	return withBaseUni(
```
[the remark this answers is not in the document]
guggero: Line was too long. I guess it should go to its own commit, but since we only touch this file once in this PR I thought it was fine to sneak some formatting changes in.

universe/interface.go
```diff
@@ -50,6 +50,19 @@ func (i *Identifier) String() string {
...
+	groupKey := "<nil>"
+	if i.GroupKey != nil {
+		groupKey = hex.EncodeToString(
+			schnorr.SerializePubKey(i.GroupKey),
+		)
+	}
+
+	return fmt.Sprintf("%v (asset_id=%x,group_key=%v)",
+		i.String(), i.AssetID[:], groupKey)
+}
```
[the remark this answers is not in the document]
guggero: I like JSON as well, but I don't want to start mixing formats, as we normally use `spew.Sdump()` in the code base. Here that isn't sufficient because we have defined a `String()` method that doesn't show all information (but `spew.Sdump()` always uses that method if it exists).

tapdb/sqlc/queries/universe.sql
guggero: We always have public keys in the DB, never private keys. The difference to other public keys is that we always have 33-byte compressed public keys (that include the parity byte). But here we need to ignore that parity byte (see https://github.com/lightninglabs/taproot-assets/pull/328#issuecomment-1568794371 for some additional context) and only use the remaining 32 bytes, as mentioned here: https://github.com/lightninglabs/taproot-assets/blob/6317b3046b590cac04cd060c3c29ee947d40a07b/tapdb/sqlc/migrations/000007_universe.up.sql#L15

tapdb/universe.go
```diff
@@ -292,6 +295,13 @@ func (b *BaseUniverseTree) RegisterIssuance(ctx context.Context,
 		return nil, err
 	}
 
+	// Up to this point many writers can perform all required read
+	// operations to prepare and validate the keys. But since we're now
+	// actually starting to write, we want to limit this to a single writer
+	// at a time.
+	b.registrationMtx.Lock()
```
Roasbeef, line 302: I think we should revert this change. We want to make sure that our db concurrency control can handle situations like this. For sqlite, there's only a single writer. For postgres, the new serialization mode should detect conflicts, then abort so it can retry safely.

Roasbeef: Gotcha, perhaps we need to increase either the back off timeout, or the number of attempts (I just kinda chose 10 randomly). Wanting to try to tune these params, as otherwise we may end up hitting this in other areas that have more involved db transactions.

Issue #302, opened 2023-05-09
bip: test vector creation 


Roasbeef:
One final thing we want to do before we request the final BIP number sis request test vectors for all the relevant sub-BIPs. 

I think one way we can automate a lot of this is leverage our existing TDD tests: they're all already induvidual test vectors, so we can add a new `-json` flag to output them in a formatted JSON output (written to a file). 

This is a master tracking issue to keep track of all the test vectors we'd want to add: 

## Test Vector Checklist

- [ ] asset TLV encoding 

- [ ] addr encoding

- [ ] proof file encoding

- [ ] VPSBT encoding 
- [ ] positive + negative tests for VM validation 
- [ ] positive + negative tests for proof validation 
- [ ] MS-SMT tree tests
   * creation from canned leaf (hash check to match)
   * overflow detection 
   * proof compression + decompression 
   * inclusion proof validation 
   * exclusion proof validation

guggero:
Adding some thoughts on this before starting work.

## Format

Should we use inline test vectors as text (JSON) in the mediawiki documents or link to one or more JSON files in the repo?
=> Actual downloadable JSON files have a lot of advantages, can be turned into unit tests directly.

Proposed general JSON format (taking the [MuSig2/BIP-0327 vectors](https://github.com/bitcoin/bips/tree/master/bip-0327/vectors) as a template/inspiration):

```
{
    "shared_data_items": [
      "item 0",
      "item 1",
      "item 2",
      "item 3",
     ],
    "valid_test_cases": [
        {
            "shared_data_items_indices": [0, 1],
            "expected": "abc",
            "comment": "Foo bar baz",
        },
        {
            "shared_data_items_indices": [2, 3],
            "expected": "def",
            "comment": "Bar"
        }
    ],
    "error_test_cases": [
        {
            "shared_data_items_indices": [0, 4],
            "error": "invalid_pubkey",
            "comment": "Public key invalid",
            "tapd_err": "invalid public key: unsupported format"
        },
```

## Content

Where we don't already have unit tests where we can add code to export the test vectors as JSON files, we should make sure we cover at least the following states:
 - Minimal/empty set (e.g. minimal required values for an asset or vPSBT or empty MS-SMT)
 - Maximal/full set (e.g. all values set on an asset or vPSBT)
 - Examples for invalid values
 - Positive/valid examples for different types (e.g. normal vs. collectible assets, grouped vs non-grouped assets, interactive vs. non-interactive vPSBTs)

[...]

PR #326, opened 2023-05-26
multi: add test vectors for asset TLV encoding

guggero:
Opening as draft so we can share helper functions and concepts.


dstadulis:
action: go to ready for review

asset/testdata/asset_tlv_encoding.json
```diff
@@ -0,0 +1,177 @@
...
...00000000000000984935e54049d8e034222d45ca161aac3d6a2c8f555b596afe99c1f386fd345300000000000000008ead572e77bf1522e570e306096945faa70c27cf7fc92e4e412872865ebcd20d000000000000000042af08a53490158553d5138cfe380acfb49d6e1b4d3b223f1c9962bf9e0767260000000000000000a7d596226ff12cc379a1c733f57f8adc54e6f58f572b1da849122381651b02fa0000000000000000cb710c2bc64511ee93f9c6c09c530c9eead588ffe3a3bf81dfcaead04abb9627000000000000000028c5d45761cf38ed4fb2d4820bc57b9e591a04e9623b1f46da49f3ffa9dbf506000000000000000038f542347e7b8ee5d194890efc220bdf3f49d967c8a3e69d7c113aa0a95d290d0000000000000000b024dbaa5aa401f43071f273fdc98d13381628fe6b5cefd482ca8d64275cf29b00000000000000008f945cfd43b0996b000314bc186816cae2e6698c11ca3749b2d7a00718fb8c3f000000000000000017a3d2b53dfb08a1ab386e6c6f846c78d1aee8c4521e558c01dafecfcda757f0000000000000000013fefdf542341f1f92be3d67058cc4e7eab1303c48ee7fb8873c9dd56c30eedc0000000000000000dd8023b710a813409dd35d7ffab8c71a7625eb15b5bdb981862a5d270b1622a30000000000000000a9ed7261d36f1df934db5cfc81e1ea6e7c7117203dc4b88ccf336e6385153ec100000000000000005a61e238f07e3a8114e39670c1e5ff430913d5793028258cf8a49282efee441100000000000000000000000000000000000000000000000001000000000000000000000000000000",
```
Roasbeef, line 23: Proof seems kiiinda large here? Would expect smaller with just a simple asset and a compressed inclusion proof with a single leaf.

address/address_test.go
```diff
@@ -167,6 +169,8 @@ func TestNewAddress(t *testing.T) {
 	}
 
 	for _, testCase := range testCases {
+		testCase := testCase
```
Roasbeef, line 173: and it strikes yet again.....

address/address_test.go
```diff
@@ -308,3 +309,71 @@ func TestAddressEncoding(t *testing.T) {
 		}
 	}
 }
+
+// TestBIPTestVectors tests that the BIP test vectors are passing.
+func TestBIPTestVectors(t *testing.T) {
+	t.Parallel()
+
+	testVectors := &TestVectors{}
+	test.ParseTestVectors(
```
Roasbeef, line 318: Nice helper func 👍

address/address_test.go
```diff
@@ -308,3 +309,71 @@ func TestAddressEncoding(t *testing.T) {
...
+		t, "testdata/address_tlv_encoding.json", &testVectors,
```
Roasbeef, line 319: I think I'm missing how/where the test vectors were _generated_ in the first place. Eg: would expect an optional flag on an existing (?) test to output the vectors in a specified location.

Roasbeef, line 319: Also perhaps move the path/name of the vectors to a constant?

commitment/mock.go
```diff
@@ -64,3 +65,63 @@ func ParseTapscriptSibling(t testing.TB, ts string) *TapscriptPreimage {
...
+			Version: uint8(p.AssetProof.Version),
+			AssetID: hex.EncodeToString(p.AssetProof.AssetID[:]),
+		}
+	}
+
+	return tp
+}
+
+type TestProof struct {
+	AssetProof        *TestAssetProof        `json:"asset_proof"`
```
Roasbeef, line 90: Maybe prefix with Smt? IIUC this is the smt inclusion proof for na asset while the bottom is the entire state transition proof?

Roasbeef (commented):
Nice work with all the hlper functions! Should make it super easy to add new tests in the future. 

The main thing I think I'm missing in this PR is: how are the vectors generated in the first place? I see the helper funcs, and also code to verify that we [ass the test vectors, but don't see the test vector generation code.

dstadulis:
There are some randomly generated vectors -- would using a seed from a rand source could help.

## TODO / Missing:
[] Considering using a flag.
[] Possibly adding coverage VM 
There aren't edge cases address
[] Need more vectors
Default structure is in

We can later add more vectors / edge cases.

asset/asset.go
```diff
@@ -66,11 +66,6 @@ var (
 	NUMSCompressedKey = ToSerialized(NUMSPubKey)
 	NUMSScriptKey     = ScriptKey{
 		PubKey: NUMSPubKey,
-		TweakedScriptKey: &TweakedScriptKey{
```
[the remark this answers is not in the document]
guggero: Hmm, good question. Maybe... But I don't think it should matter as we're never going to use the internal key as we always check whether it's the NUMS key before attempting to use it in an `lnd` RPC.

address/address_test.go
guggero: Good point! I went ahead and added a build flag (can be activated using `make unit gen-test-vectors=true`) to generate the files and also invested some time to make the content deterministic. So it should be as easy as adding new test cases in Golang to generate extended test vectors!

commitment/mock.go
guggero: I tried to keep the names of the original structs:
https://github.com/lightninglabs/taproot-assets/blob/0a58cdbf873f4634f05aa6f40fa75f888f0f5195/commitment/proof.go#L54

Though I agree it's not super clear now. IMO it was way clearer what the proofs meant when they were called `AssetProof` and `TaroProof`, as that directly corresponded to the two levels of trees.

Happy to rename either of them, but I would suggest to do it in both the test struct and the "production code" struct?

guggero:
Taking this out of draft status as we now have deterministically generated test vectors for assets, addresses, proofs and vPSBTs. This does not fully address https://github.com/lightninglabs/taproot-assets/issues/302 yet, but is a big step towards that.

guggero:
I added more test vectors in the latest push, now also including VM positive/negative tests and MS-SMT tests, so this PR should now cover everything described in https://github.com/lightninglabs/taproot-assets/issues/302.
We should still add more edge case vectors later, but this PR is already large enough, so that should happen in a follow-up PR.

Roasbeef (approved):
:lgtm: 🎉

Great work giving us a solid base to build on and move towards initial BIP number acquisition. 

> Reviewed 27 of 32 files at r1, 14 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/326#-:-N_Bb7V94kStGgLiS86_:b2kzbps)* status: all files reviewed, 8 unresolved discussions (waiting on @guggero, @halseth, and @jharveyb)

___
*[`mssmt/encoding_test.go` line 148 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/326#-N_BaU_H7rse8PGh6Hyl:-N_BaU_H7rse8PGh6Hym:b-8fx2u7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/67fb444d9d4c76f0e9b6ed8c7b9da77022b71d0e/mssmt/encoding_test.go#L148)):*
> ```Go
> 	// Write test vectors to file. This is a no-op if the "gen_test_vectors"
> 	// build tag is not set.
> 	test.WriteTestVectors(t, proofsTestVectorName, testVectors)
> ```

:+1:

asset/asset_test.go
```diff
@@ -12,22 +12,34 @@ import (
...
+			"821525f66a4a85ea8b71e482a74f382d2ce5ebeee8fdb2172f47" +
+			"7df4900d310536c0",
 	)
 	sig, _ = schnorr.ParseSignature(sigBytes)
+
+	generatedTestVectorName = "asset_tlv_encoding_generated.json"
+
+	allTestVectorFiles = []string{
+		generatedTestVectorName,
+		"asset_tlv_encoding_error_cases.json",
```
jharveyb, line 41: Are the `error_cases` tests populated manually, or am I missing where the logic is for generating those?

address/address_test.go
```diff
@@ -217,7 +233,37 @@ func TestAddressEncoding(t *testing.T) {
...
+			f: func() (*Tap, string, error) {
+				return randEncodedAddress(
+					t, &MainNetTap, false, false,
+					asset.Normal,
+				)
+			},
+			err: nil,
+		},
+		{
+			name: "simnet group collectible",
```
jharveyb, line 266: nit: comment should be signet, not simnet

vm/vm_test.go
```diff
@@ -602,43 +613,156 @@ func TestVM(t *testing.T) {
...
 			}
-			if len(splitSet) == 0 {
-				err := verify(nil)
-				require.Equal(t, testCase.err, err)
-				return
+			if testCase.err == nil {
+				validVectors.ValidTestCases = append(
+					validVectors.ValidTestCases, tv,
+				)
+			} else {
```
jharveyb, line 640: 💯 

Really cool to have autogen for both positive and negative test vectors!

jharveyb (approved):
LGTM! This is an awesome PR, very excited to see these vectors used by others.

AFAICT only the VM negative test vectors are autogenerated right now, so I'm a bit confused as to why that can't be done for the other packages as well. Non-blocking.

asset/asset_test.go
guggero: Yeah, so basically if a test vector file name doesn't have a corresponding package-level variable for it (and it only appears in the `allTestVectorFiles` slice) then it means it's currently not autogenerated but contains some manual negative tests.

PR #328, opened 2023-05-26
universe: RPC fixes

jharveyb:
Display 33-byte script keys instead of 32 bytes. We should audit the current structs are parsing to make it internally consistent, so that copy & paste works on the CLI.

### DIFF
--- a/cmd/tapcli/universe.go
+++ b/cmd/tapcli/universe.go
@@ -3,12 +3,8 @@ package main
 import (
 	"bytes"
 	"encoding/hex"
-	"errors"
 	"fmt"
-	"strconv"
-	"strings"
 
-	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	tap "github.com/lightninglabs/taproot-assets"
 	"github.com/lightninglabs/taproot-assets/proof"
 	"github.com/lightninglabs/taproot-assets/taprpc/universerpc"
@@ -280,41 +276,18 @@ var universeProofQueryCommand = cli.Command{
 	Action: universeProofQuery,
 }
 
-func parseUniOutpoint(ctx *cli.Context) (*universerpc.Outpoint, error) {
-	// Parse a bitcoin outpoint in the form txid:index into a
-	// wire.OutPoint struct.
-	parts := strings.Split(ctx.String(outpointName), ":")
-	if len(parts) != 2 {
-		return nil, errors.New("outpoint should be of " +
-			"the form txid:index")
-	}
-	txidStr := parts[0]
-	if hex.DecodedLen(len(txidStr)) != chainhash.HashSize {
-		return nil, fmt.Errorf("invalid hex-encoded "+
-			"txid %v", txidStr)
-	}
-
-	outputIndex, err := strconv.ParseInt(parts[1], 10, 32)
-	if err != nil {
-		return nil, fmt.Errorf("invalid output "+
-			"index: %v", err)
-	}
-
-	return &universerpc.Outpoint{
-		HashStr: txidStr,
-		Index:   int32(outputIndex),
-	}, nil
-}
-
 func parseAssetKey(ctx *cli.Context) (*universerpc.AssetKey, error) {
-	outpoint, err := parseUniOutpoint(ctx)
+	outpoint, err := tap.UnmarshalOutpoint(ctx.String(outpointName))
 	if err != nil {
 		return nil, err
 	}
 
 	return &universerpc.AssetKey{
 		Outpoint: &universerpc.AssetKey_Op{
-			Op: outpoint,
+			Op: &unirpc.Outpoint{
+				HashStr: outpoint.Hash.String(),
+				Index:   int32(outpoint.Index),
+			},
 		},
 		ScriptKey: &universerpc.AssetKey_ScriptKeyStr{
 			ScriptKeyStr: ctx.String(scriptKeyName),
--- a/itest/universe_test.go
+++ b/itest/universe_test.go
@@ -7,7 +7,9 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	tap "github.com/lightninglabs/taproot-assets"
 	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightninglabs/taproot-assets/taprpc"
 	unirpc "github.com/lightninglabs/taproot-assets/taprpc/universerpc"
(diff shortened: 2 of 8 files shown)


rpcserver.go
```diff
@@ -2322,7 +2322,7 @@ func unmarshalLeafKey(key *unirpc.AssetKey) (universe.BaseKey, error) {
 
 	switch {
 	case key.GetScriptKeyBytes() != nil:
-		pubKey, err := schnorr.ParsePubKey(
+		pubKey, err := btcec.ParsePubKey(
```
Roasbeef, line 2325: Doesn't match this: https://github.com/lightninglabs/taproot-assets/blob/dc5782b0bd0eb8b97b68fc9c50274e1fc8fa8db8/rpcserver.go#L2206-L2217

universe/auto_syncer.go
```diff
@@ -109,7 +110,13 @@ func (f *FederationEnvoy) Start() error {
 		})
 		err := f.AddServer(serverAddrs...)
 		if err != nil {
-			log.Warnf("unable to add universe servers: %v", err)
+			// On restart, we'll get an error for universes already
```
Roasbeef, line 113: I think we want to actually check against the expected error here. Instead of re-wrapping to add that text context, we want to just ignore it all together and move on (no need to log as it'll be printed each time).

Roasbeef (changes requested):

universe/auto_syncer.go
jharveyb: Agree, though I hit a dependency loop when trying to check against an error defined in tapdb/sqlerrors. Will look at defining it elsewhere.

jharveyb: Agree, though I hit a dependency loop when trying to check against an error defined in tapdb/sqlerrors. Will look at defining it elsewhere.

Roasbeef: Yeah I think we should define it in the `universe` package, then the error is returned like in some other similar areas.

jharveyb:
From offline discussion: A better approach would be accepting both 33 and 32 byte script keys when possible. In the context of universes, we only store script keys as 32 bytes, but a user may provide the same key in both formats as other CLI commands will show asset script keys in 33 byte form.

Will update this PR to add handling for both script key sizes.

universe/auto_syncer.go
```diff
@@ -108,7 +108,10 @@ func (f *FederationEnvoy) Start() error {
 			return NewServerAddrFromStr(a)
 		})
 		err := f.AddServer(serverAddrs...)
-		if err != nil {
+		// On restart, we'll get an error for universe servers already
+		// inserted in our DB, since we can't store duplicates.
+		// We can safely ignore that error.
+		if err != ErrDuplicateUniverse {
```
Roasbeef, line 114: Style nit: can also use `errors.Is` here.

Roasbeef: Another way to handle this would also just be an noop `UPSERT` into the db. I think either works though.

cmd/tapcli/universe.go
```diff
@@ -280,41 +276,18 @@ var universeProofQueryCommand = cli.Command{
...
-
-	return &universerpc.Outpoint{
-		HashStr: txidStr,
-		Index:   int32(outputIndex),
-	}, nil
-}
-
 func parseAssetKey(ctx *cli.Context) (*universerpc.AssetKey, error) {
-	outpoint, err := parseUniOutpoint(ctx)
+	outpoint, err := tap.UnmarshalOutpoint(ctx.String(outpointName))
```
Roasbeef, line 280: 👍

rpcserver.go
```diff
@@ -2356,9 +2398,7 @@ func unmarshalLeafKey(key *unirpc.AssetKey) (universe.BaseKey, error) {
 			return baseKey, err
 		}
 
-		pubKey, err := schnorr.ParsePubKey(
-			scriptKeyBytes,
-		)
+		pubKey, err := parseUserScriptKey(scriptKeyBytes)
```
Roasbeef, line 2401: Hmm, in some cases we actually still want the 32-byte key here, though I think we may just lean towards that serializtaion late rdown the call path. Will double check those areas.

itest/universe_test.go
```diff
@@ -263,6 +266,43 @@ func testUniverseFederation(t *harnessTest) {
...
+		firstAsset[0].ChainAnchor.AnchorOutpoint,
+	)
+	require.NoError(t.t, err)
+	require.Len(t.t, firstScriptKey, btcec.PubKeyBytesLenCompressed*2)
+
+	firstAssetProofQuery := unirpc.UniverseKey{
+		Id: &unirpc.ID{
+			Id: &unirpc.ID_AssetId{
+				AssetId: firstAssetID,
+			}},
```
Roasbeef, line 283: Weird double close bracket here re the line folding.

itest/universe_test.go
```diff
@@ -263,6 +266,43 @@ func testUniverseFederation(t *harnessTest) {
 	// should also be able to query for stats specifically for the asset.
 	assertUniverseStats(t.t, bob, 1, 0, 1)
 
+	// We should also be able to fetch the new asset from Bob's Universe,
+	// and query for that asset with the compressed script key.
+	firstAssetID := firstAsset[0].AssetGenesis.AssetId
+	firstScriptKey := hex.EncodeToString(firstAsset[0].ScriptKey)
```
Roasbeef, line 272: Maybe move this to the other universe test? Given doesn't have to specifically do with any federation sync operations.

Roasbeef (changes requested):

rpcserver.go
jharveyb: I think this doesn't break currently because we default to using script keys tweaked BIP 86 style? But could break with a user-supplied script key.

rpcserver.go
```diff
@@ -2330,6 +2346,34 @@ func (r *rpcServer) AssetLeaves(ctx context.Context,
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid output index: %v", err)
+	}
+
+	return &wire.OutPoint{
+		Hash:  *txid,
+		Index: uint32(outputIndex),
+	}, nil
+}
+
```
ffranr, line 2376: I think this can be replaced by `NewOutPointFromString` from `btcd.wire`.

rpcserver.go
```diff
@@ -2446,6 +2467,7 @@ func (r *rpcServer) marshalIssuanceProof(ctx context.Context,
 	if err != nil {
 		return nil, err
 	}
+
```
ffranr, line 2470: I would consider moving the initialisation of `uniRoot` down to where its fields are populated
```
uniRoot.AssetName = assetLeaf.Asset.AssetGenesis.Name
uniRoot.Id = req.Id
```

ffranr (changes requested):
Nice work! :cake:

rpcserver.go
jharveyb: Good catch! I'm wondering if we care that this would move the `btcd` depedency from a tagged version to master / a specific commit?

jharveyb:
We have the same problem with group keys, since they are shown as 33-byte on the CLI in the outpoint of `assets list`, but the universe ID is constructed from the 32-byte version of the key. IIUC, those should be safe to just truncate since they will always have an even parity byte.

jharveyb:
Discussed offline with @guggero to clear up some confusion:

- Tweaked keys are stored internally in the compressed form, along with the corresponding internal key and any non-empty tweak information, so that can produce signatures by calling out to the LND signer via RPC.
- We don't display this internal key or key tweak for asset script keys, but we do display some of this information for the asset group key.
 - Tweaked asset script keys, and the asset group key, will always end up with even parity due to schnorr serialization before storage; this does not affect our ability to sign with these keys.
 
 Given that these keys will always have even parity, I think for this PR, we can just check the user-provided key for the even parity byte and then remove it / truncate the key before passing it along further.

rpcserver.go
jharveyb: Tracked in #362.

jharveyb:
Switched to key truncation + tested as working locally with an odd-parity group key.

guggero (approved):
Nice, LGTM :tada:

itest/universe_test.go
```diff
@@ -122,6 +124,44 @@ func testUniverseSync(t *harnessTest) {
...
+
+	// We should also be able to fetch an asset from Bob's Universe, and
+	// query for that asset with the compressed script key.
+	firstAssetID := rpcSimpleAssets[0].AssetGenesis.AssetId
+	firstScriptKey := hex.EncodeToString(rpcSimpleAssets[0].ScriptKey)
+	firstOutpoint, err := tap.UnmarshalOutpoint(
+		rpcSimpleAssets[0].ChainAnchor.AnchorOutpoint,
+	)
+	require.NoError(t.t, err)
+	require.Len(t.t, firstScriptKey, btcec.PubKeyBytesLenCompressed*2)
```
ffranr, line 136: Why is that true?

ffranr (approved):
Nice work :+1:

itest/universe_test.go
jharveyb: Which part, fetching the asset or the key length?

We synced the minted assets to Bob's universe earlier, and we always display the key in compressed form, so as a hex-encoded string it should always be 66 characters.

ffranr: Length makes sense now, thanks :+1:

Roasbeef (approved):
LGTM 🐲

Issue #333, opened 2023-05-31
tapd/universe: fix potential concurrency issues related to concurrent SMT tree insertion 


Roasbeef:
Related to https://github.com/lightninglabs/taproot-assets/issues/320 and https://github.com/lightninglabs/taproot-assets/issues/324

Since going testnet production, we've seen some issues with the default Universe server related to concurrent operations causing read/write hazards/anomalies. 

We've seen instances on the prod server for certain assets where all the leaves are in place, but the universe _root_ is actually incorrect. Here's an example:
  * Leaves: https://testnet.universe.lightning.finance/v1/taproot-assets/universe/leaves/asset-id/83f3564d453ca736368fb71a6989ae8bc547d5dbac5a9d3e4f983fe43040dc29
  * Root: https://testnet.universe.lightning.finance/v1/taproot-assets/universe/roots/asset-id/83f3564d453ca736368fb71a6989ae8bc547d5dbac5a9d3e4f983fe43040dc29

In this example we have ~100 leaves, but the root commits to 11, as if there were only 11 collectibles. This then results in an invalid proof being sent out, as the root doesn't match the leaf when inserted into the tree. We _do_ always read the root from disk in order to generate an inclusion proof, but if the proof is invalid, then this will be as well. 

We have a few options here:
  * Create a system for fine grained tree locking to ensure concurrent safe MS-SMT access. 
  * Use PG advisory/table locking to lock all relevant tables for these operations. 
  * Add internal consistnecy checks to ensure the server never serves incorrect data. 
  * Update the default postgres concurrency settings. 

I think the last one may be the best options, we we've seen this elsewhere when we want to do things like log the set of universe events. AFAICT, the default is actually read uncommitted, which can introduce a lot of issues: https://github.com/jackc/pgx/blob/c7733fe52efc445f13fe9dd3e1b829587400dfb2/stdlib/sql.go#L376-L389

Roasbeef:
Related issue in pgx: https://github.com/jackc/pgx/issues/1040

Roasbeef:
Another option is to just slap a read/write mutex on it for now as well.

dstadulis:
Subissue idea

itest to assert sql and postgres methods match

Roasbeef:
Was fixed by #334

PR #334, opened 2023-05-31
tapdb: use the serializable isolation by default for postgres

Roasbeef:
In this commit, we start to use the serializable snapshot isolation by default for postgres. For sqlite, this is a noop as only a single writer is allowed at any time. With this mode, we should guard against all types of anomalies, at the cost of extra locking and potential perf hits.

One thing we'll need to test out is if the DB driver will automatically do retries or not if a transaction cannot be serialized.

This is an attempt at fixing #333 at a global level.

### DIFF
--- a/tapdb/interfaces.go
+++ b/tapdb/interfaces.go
@@ -3,6 +3,8 @@ package tapdb
 import (
 	"context"
 	"database/sql"
+	"errors"
+	prand "math/rand"
 	"time"
 
 	"github.com/lightninglabs/taproot-assets/tapdb/sqlc"
@@ -12,6 +14,17 @@ var (
 	// DefaultStoreTimeout is the default timeout used for any interaction
 	// with the storage/database.
 	DefaultStoreTimeout = time.Second * 10
+)
+
+const (
+	// DefaultNumTxRetries is the default number of times we'll retry a
+	// transaction if it fails with an error that permits transaction
+	// repetition.
+	DefaultNumTxRetries = 10
+
+	// DefaultRetryDelay is the default delay between retries. This will be
+	// used to generate a random delay between 0 and this value.
+	DefaultRetryDelay = time.Millisecond * 50
 )
 
 // TxOptions represents a set of options one can use to control what type of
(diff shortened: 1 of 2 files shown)


Roasbeef:
itest failure show we do indeed have an issue here:
> 2023-05-31 22:26:22.717 [WRN] UNIV: unable to log sync event: unknown postgres error: ERROR: null value in column "universe_root_id" violates not-null constraint (SQLSTATE 23502)

Roasbeef:
Alternative path for the last commit: https://github.com/powerman/pqx/blob/a6a4bb664f620589441e7b28d20464134af550c2/serialize.go#L16

Roasbeef:
> itest failure show we do indeed have an issue here:

Fixed with the latest commit: we now retry (forever). We should log, and also consider an upper limit for retry as well.

tapdb/interfaces.go
```diff
@@ -113,7 +115,20 @@ func (t *TransactionExecutor[Q]) ExecTx(ctx context.Context,
...
+
+		// At this point, we know the DB wasn't able to properly
+		// serialize the error, so we'll re-execute everything to try
+		// once again.
+		var serializationErr *ErrSerializationError
+		if errors.As(dbErr, serializationErr) {
+			// Roll back the transaction, then pop back up to try
+			// once again.
+			//
+			// TODO(roasbeef): needs retry limit
```
guggero, line 128: Should we fix this TODO to avoid endless re-tries?

tapdb/interfaces.go
```diff
@@ -100,6 +101,7 @@ func NewTransactionExecutor[Querier any](db BatchedQuerier,
 func (t *TransactionExecutor[Q]) ExecTx(ctx context.Context,
 	txOptions TxOptions, txBody func(Q) error) error {
 
+txStart:
```
guggero, line 104: This will cause a new `defer` with a rollback to be caused for each retry. And IIUC then `Rollback()` will return an error if it is called twice. So I think we need to roll this out into an actual loop, otherwise we might get weird side effects or errors.

guggero (changes requested):
Good catch! This should indeed make things safer.

Though with the current implementation I think we might get into weird situations with the combinations of `goto` and `defer()` (see inline comments).

tapdb/interfaces.go
Roasbeef: Yeah I think so, open question: what's a good limit for retries? My strategy was going to be deploy the modified version in staging/testnet, then check the logs to see how often we retry. We have @jharveyb's mega mint script, so that can be used to calibrate.

Roasbeef:
Tacked on a proper retry loop PTAL. 

Also has logging now so we can see how many times things need to be retried in practice.

Roasbeef:
Re calling rollback multiple times, looks like any second call will be a noop here due to the CAS: https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/database/sql/sql.go;l=2279-2284

guggero (approved):
Very nice, LGTM :tada:

guggero:
> Re calling rollback multiple times, looks like any second call will be a noop here due to the CAS: https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/database/sql/sql.go;l=2279-2284

Ah, yes. I overlooked that we just ignore the `ErrTxDone` in the `defer` call, so it's a no-op indeed.

Issue #320, opened 2023-05-24
[sync with universe] unable to sync universe: proof for key  is invalid


Umiiii:
Hi, I am trying to mint my own taproot assets and I followed the steps:
- Launched bitcoin testnet with `bitcoind`, where the `bitcoin.conf` looks like:
```
txindex=1
rpcuser=umi
rpcpassword=whatever
rpcallowip=127.0.0.1
rpcconnect=127.0.0.1
datadir=/bitcoin/data
server=1
daemon=1
zmqpubrawblock=tcp://127.0.0.1:28332
zmqpubrawtx=tcp://127.0.0.1:28333
```

- Launched `lnd`, with `lnd.conf`:
```
externalip=whatever
bitcoin.active=1
bitcoin.testnet=1
bitcoin.node=bitcoind
btcd.rpchost=localhost
btcd.rpcuser=umi
btcd.rpcpass=whatever
```

- Launched `tapd`, with 
`tapd --network=testnet --debuglevel=debug --lnd.host=localhost:10009 --lnd.macaroonpath=~/.lnd/data/chain/bitcoin/testnet/admin.macaroon --lnd.tlspath=~/.lnd/tls.cert`

- Unlock lnd, create wallet address, and receive some testnet bitcoin:
```
lncli unlock
lncli --network=testnet newaddress p2tr
lncli --network=testnet walletbalance
```

after that, mint asset:
```
root@umibitcoin ~ # tapcli assets mint --type normal --name fantasycoin --supply 100 --meta_bytes "fantastic money"
{
    "batch_key": "0395abdd75f14cdb7ce6b70bd822146a24e9734450c9585fd3759eca76c001d508"
}
root@umibitcoin ~ # tapcli assets mint finalize
{
    "batch_key": "0395abdd75f14cdb7ce6b70bd822146a24e9734450c9585fd3759eca76c001d508"
}

```

```
root@umibitcoin ~ # tapcli assets list
{
    "assets": [
        {
            "version": 0,
            "asset_genesis": {
                "genesis_point": "6b47b7f3b5b4f016795cb9189fc1c03e8383aa811cb30b9d545232714dae0796:1",
                "name": "umicoin",
                "meta_hash": "4ff1b02c93487f8762f49d09877d2c8edcb19104ecff0bc0119b4f22b85b3e4c",
                "asset_id": "a466936816b579a3e94ddcfb6acebe6561783533fc726cbc0bb9c31ec7a046de",
                "output_index": 0,
[...]

PR #335, opened 2023-05-31
WIP: Universe root deletion RPC

jharveyb:
Interim fix for https://github.com/lightninglabs/taproot-assets/issues/320, which should be reworked once we land other DB changes.

Add a command to delete specific universe trees, which is useful for avoiding universe sync failures. Ideally this would be done more automatically with cascading deletes in the DB.

Successfully tested as fixing the root cause of #320, my `tapd` no longer shows an incorrect root for that asset group. Sync with the default universe still fails.


tapdb/universe.go
```diff
@@ -64,6 +64,15 @@ type BaseUniverseStore interface {
 	QueryUniverseLeaves(ctx context.Context,
 		arg UniverseLeafQuery) ([]UniverseLeaf, error)
 
+	// DeleteUniverseLeaves is used to delete leaves that reside in a
+	// universe tree.
+	DeleteUniverseLeaves(ctx context.Context, namespace string) error
```
Roasbeef, line 69: See prior comment, I think we can ideally lean on the `ON DELETE CASCADE` behavior.

tapdb/mssmt.go
```diff
@@ -57,6 +57,9 @@ type TreeStore interface {
 	// leaf) from the store.
 	DeleteNode(ctx context.Context, n DelNode) (int64, error)
 
+	// DelteRoot deletes a root node from the store.
+	DeleteRoot(ctx context.Context, namespace string) (int64, error)
```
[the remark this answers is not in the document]
Roasbeef: Scratch that, looks like we'll need to do something a bit more involved to also have it work with sqlite: https://simonwillison.net/2020/Sep/23/sqlite-advanced-alter-table/

tapdb/sqlc/queries/mssmt.sql
```diff
@@ -51,6 +51,9 @@ WITH subtree_cte (
 -- name: DeleteNode :execrows
 DELETE FROM mssmt_nodes WHERE hash_key = $1 AND namespace = $2; 
 
+-- name: DeleteRoot :execrows
+DELETE FROM mssmt_roots WHERE namespace = $1;
```
Roasbeef, line 55: Still getting through the diff, but perhaps we should add another variant of `DeleteNode` that also deletes just based on namespace. Or we can modify things to ignore the namespace with a `sqlc.narg` if it isn't set.

tapdb/universe.go
```diff
@@ -64,6 +64,15 @@ type BaseUniverseStore interface {
...
+
+	// DeleteUniverseRoot is used to delete the root of a universe tree.
+	DeleteUniverseRoot(ctx context.Context, namespace string) error
+
+	DeleteUniverseEvents(ctx context.Context, namespace string) error
```
Roasbeef, line 74: Style nit: missing `godoc` comment.

tapdb/universe.go
```diff
@@ -617,4 +617,83 @@ func (b *BaseUniverseTree) MintingLeaves(ctx context.Context,
 	return leaves, nil
 }
 
+// DeleteUniverse deletes the entire universe tree.
+func (b *BaseUniverseTree) DeleteUniverse(ctx context.Context) (string, error) {
+	var writeTx BaseUniverseStoreOptions
+
+	dbErr := b.db.ExecTx(ctx, &writeTx, func(db BaseUniverseStore) error {
+		// Fetch all base keys for the current universe.
+		universeKeys, err := db.FetchUniverseKeys(ctx, b.smtNamespace)
```
Roasbeef, line 626: Hmm, if we decide to go with deletion directly on the table, then we can speed things up quite a bit. As is, since we call `Delete` with the key below, it'll actually do all the normal SMT operations (bubble up the new sum, etc, etc) that isn't actually needed.

Roasbeef:
Missing an entry for REST in the `yaml` file also.

tapdb/universe.go
Roasbeef: We only want to just wipe all the relate smt nodes from the db. Using the `Delete` call as below will actually recompute the root each time, which causes extra reads/insertions. Instead we can implement the deletion purely here in the db (delete all leaves w/ namespace, then root), side stepping the main external interface.

tapdb/sqlc/queries/mssmt.sql
jharveyb: Made a separate call with a more distinct name.

Roasbeef:
Diff looking better, I think this is just missing unit tests + an itest (to replicate the scenario we execute live on the instance).

taprpc/universerpc/universe.proto
```diff
@@ -19,6 +19,8 @@ service Universe {
     */
     rpc QueryAssetRoots (AssetRootQuery) returns (QueryRootResponse);
 
+    rpc DeleteAssetRoot (AssetRootQuery) returns (QueryRootResponse);
```
guggero, line 22: We should create and use a new message for the response, otherwise things might get awkward in the future (and we might need to care about backward compatibility then).

Also, needs a documentation comment.

mssmt/store.go
```diff
@@ -244,6 +252,41 @@ func (d *DefaultStore) DeleteCompactedLeaf(key NodeHash) error {
...
+	d.cntDeletes++
+
+	return nil
+}
+
+// DeleteAllNodes deletes all nodes in the MS-SMT.
+func (d *DefaultStore) DeleteAllNodes() error {
+	// Delete leaves, then compacted leaves, then branches.
+	leafKeys := maps.Keys(d.leaves)
+	for ind := range leafKeys {
```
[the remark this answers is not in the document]
jharveyb: Not missing anything, just defaulted to index-based loops.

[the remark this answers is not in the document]
jharveyb: Turns out clearing of a map can be done with `maps.Clear()`, and will end up in the stdlib for golang 1.21.

jharveyb:
Added unit tests + itests, no large changes otherwise.

Needed one change here to handle trees with missing leaves:

https://github.com/lightninglabs/taproot-assets/blob/7cabe130f9e184ec0bab209177df255de5a3142f/mssmt/store.go#L301

mssmt/store.go
```diff
@@ -244,29 +252,66 @@ func (d *DefaultStore) DeleteCompactedLeaf(key NodeHash) error {
...
+
+	return nil
+}
+
 // GetChildren returns the left and right child of the node keyed by the given
 // NodeHash.
 func (d *DefaultStore) GetChildren(height int, key NodeHash) (
 	Node, Node, error) {
 
+	EmptyNodeHash := EmptyTree[height].NodeHash()
```
guggero, line 286: This should be a lower case var, otherwise it looks like a package level exported constant/var.

mssmt/store.go
```diff
@@ -244,29 +252,66 @@ func (d *DefaultStore) DeleteCompactedLeaf(key NodeHash) error {
...
 	getNode := func(height uint, key NodeHash) Node {
-		if key == EmptyTree[height].NodeHash() {
+		if key == EmptyNodeHash {
```
guggero, line 288: This changes the behavior of this whole function if I understand correctly. Because we're no longer comparing with the correct empty hash. We call `getNode` with `height` but then further down with `height+1`. So without any additional context, this change looks wrong to me.

cmd/tapcli/universe.go
```diff
@@ -139,6 +140,46 @@ func universeRoots(ctx *cli.Context) error {
...
+	client, cleanUp := getUniverseClient(ctx)
+	defer cleanUp()
+
+	universeID, err := parseUniverseID(ctx, true)
+	if err != nil {
+		return err
+	}
+
+	rootReq := &universerpc.DeleteRootQuery{
+		Id: universeID,
```
guggero, line 172: Missing the group key here.

guggero (commented):
Found one potential issue, the rest looks pretty good!

mssmt/store.go
jharveyb: Ah yes, didn't realize emptyHash would be compared twice - so I think moving it back into getNode() would match the old behavior?

cmd/tapcli/universe.go
jharveyb: Handled inside `parseUniverseID` ?

guggero: Ah right, sorry.

tapdb/universe_test.go
```diff
@@ -354,6 +371,29 @@ func TestUniverseTreeIsolation(t *testing.T) {
 		}
 		return false
 	}))
+
+	// We should be able to delete one Universe with no effect on the other.
```
ffranr, line 375: :+1:

ffranr (approved):
Great improvements :+1: 

One or two nits, and one doc fix.

tapdb/mssmt_test.go
```diff
@@ -232,6 +237,24 @@ func TestTreeDeletion(t *testing.T) {
...
+					// not deleted explicitly above. Rebuild
+					// the tree and delete all nodes again.
+					// We should not be able to fetch the
+					// children of the test branch.
+					err = tx.DeleteAllNodes()
+					require.NoError(t, err)
+					insertAllNodes(t, tx)
+
+					err = tx.DeleteAllNodes()
+					require.NoError(t, err)
```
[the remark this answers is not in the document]
jharveyb: Yes it is, this PR doesn't change any of the existing delete functions. The new functions delete the whole tree, or the root, vs. a specific branch

tapdb/universe.go
```diff
@@ -618,4 +618,56 @@ func (b *BaseUniverseTree) MintingLeaves(ctx context.Context,
 	return leaves, nil
 }
 
+// DeleteUniverse deletes the entire universe tree.
+func (b *BaseUniverseTree) DeleteUniverse(ctx context.Context) (string, error) {
```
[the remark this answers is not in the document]
jharveyb: I'd rather leave it as `DeleteUniverse` to keep naming distinct if we decide to later add support for deleting specific keys and leaves.

jharveyb:
CI failure in postgres itests, but the logs show all itests passing, in ~520s; not sure what the issue may be there?

guggero:
Yeah, saw that too, probably just a weird timing issue in the itest framework.

PR #336, opened 2023-06-01
tapdb: properly set max connections for postgres

Roasbeef:
The default is unlimited, means that with the default postgres limit of 100, you'll run into errors.

### DIFF
--- a/tapdb/postgres.go
+++ b/tapdb/postgres.go
@@ -33,7 +33,7 @@ type PostgresConfig struct {
 	User               string `long:"user" description:"Database user."`
 	Password           string `long:"password" description:"Database user's password."`
 	DBName             string `long:"dbname" description:"Database name to use."`
-	MaxOpenConnections int32  `long:"maxconnections" description:"Max open connections to keep alive to the database server."`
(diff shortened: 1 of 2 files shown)


tapdb/postgres.go
```diff
@@ -72,6 +72,10 @@ func NewPostgresStore(cfg *PostgresConfig) (*PostgresStore, error) {
 		return nil, err
 	}
 
+	rawDb.SetMaxOpenConns(maxConns)
```
guggero, line 75: Shouldn't this be `cfg.MaxOpenConnections` with a fall back to this constant if the config value is 0?

Roasbeef: Ah yeah, need to respect the config, will add that. I'm seeing this actively on the testnet universe server, and also could be the culprit of some of the issues we're seeing.

jharveyb:
Stray migrations / tables?

Roasbeef:
@jharveyb fixed!

guggero (approved):

jharveyb:
Tested with #334, and confirmed these change fix the proof upload issues we observed before.

jharveyb (approved):

dstadulis:
Nice catch

PR #342, opened 2023-06-06
address+tapdb: rename field: Tap.Version -> Tap.AssetVersion

ffranr:
This commit clarifies that the version information currently stored in the tap address relates to the asset version and not to the address version.


jharveyb (approved):
LGTM 🚀

Roasbeef (approved):
LGTM ⏰

PR #345, opened 2023-06-07
universe: move sync specific structs from interface.go -> syncer.go

ffranr:
This commit is a simple refactor for clarity.


guggero:
I don't think we should do this. The `interface.go` file is a pattern that we use in `lnd` a lot. It contains the basic types the package as a whole provides, including interfaces of the main implementations. Kind of similar to how the `doc.go` file contains package-level documentation (a pattern recommended by Golang that we don't really follow though).
So IMO this refactor actually removes clarity because it breaks an existing pattern. And the question of "where do you put the interface definition if you have two implementations?" is also answered with the existing pattern.

ffranr:
@guggero Thank for taking a look. I get the pattern now.

PR #346, opened 2023-06-08
Preliminary refactor in preparation for adding multiverse support

ffranr:
Preliminary refactor PR made in an attempt to split up the work a bit.

### DIFF
--- a/tapdb/universe.go
+++ b/tapdb/universe.go
@@ -311,6 +311,46 @@ func (b *BaseUniverseTree) RegisterIssuance(ctx context.Context,
 	key universe.BaseKey, leaf *universe.MintingLeaf,
 	metaReveal *proof.MetaReveal) (*universe.IssuanceProof, error) {
 
+	var (
+		writeTx BaseUniverseStoreOptions
+
+		err           error
+		issuanceProof *universe.IssuanceProof
+	)
+
+	// Limit to a single writer at a time.
+	b.registrationMtx.Lock()
+	defer b.registrationMtx.Unlock()
+
+	dbErr := b.db.ExecTx(ctx, &writeTx, func(dbTx BaseUniverseStore) error {
+		issuanceProof, _, err = universeRegisterIssuance(
+			ctx, dbTx, b.id, key, leaf, metaReveal,
+		)
+		return err
+	})
+	if dbErr != nil {
+		return nil, dbErr
+	}
+
+	return issuanceProof, nil
+}
+
+// universeRegisterIssuance inserts a new minting leaf within the universe
+// tree, stored at the base key.
+//
+// This function returns the newly registered issuance proof and the new
+// universe root.
+//
+// NOTE: This function accepts a db transaction, as it's used when making
+// broader DB updates.
+func universeRegisterIssuance(ctx context.Context, dbTx BaseUniverseStore,
+	id universe.Identifier, key universe.BaseKey,
+	leaf *universe.MintingLeaf,
+	metaReveal *proof.MetaReveal) (*universe.IssuanceProof, mssmt.Node,
+	error) {
+
+	namespace := idToNameSpace(id)
+
 	// With the tree store created, we'll now obtain byte representation of
 	// the minting key, as that'll be the key in the SMT itself.
 	smtKey := key.UniverseKey()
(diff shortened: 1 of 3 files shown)


tapdb/universe_forest.go
```diff
@@ -68,54 +68,60 @@ func NewBaseUniverseForest(db BatchedUniverseForest) *BaseUniverseForest {
...
+		uniRoots []universe.BaseRoot
+		readTx   = NewBaseUniverseForestReadTx()
+	)
 
-	dbErr := b.db.ExecTx(ctx, &readTx, func(db BaseUniverseForestStore) error {
-		dbRoots, err := db.UniverseRoots(ctx)
-		if err != nil {
-			return err
-		}
+	dbErr := b.db.ExecTx(
```
Roasbeef, line 79: Not sure if this is a fix as introduces another layer of indentation?

config.go
```diff
@@ -95,7 +95,7 @@ type Config struct {
 
 	ChainPorter tapfreighter.Porter
 
-	BaseUniverse *universe.MintingArchive
```
Roasbeef, line 98: FWIW I think base universe still fits: it's only issuance proofs rn.

Roasbeef: Also for all other callers: they only care that they're able to register the issuance proofs, the exact naming of the caller doesn't mean much as long as the interface is upheld.

universe/interface.go
```diff
@@ -360,11 +360,12 @@ type Syncer interface {
 // DiffEngine is a Universe diff engine that can be used to compare the state
 // of two universes and find the set of assets that are different between them.
 type DiffEngine interface {
-	BaseForest
```
Roasbeef, line 363: Was mainly done to dedup on method names. Which other methods would be added here?

tapdb/universe_forest.go
ffranr: `dbErr := b.db.ExecTx(...` was over the line length limit. But I've reverted to save on indentation :+1:

universe/interface.go
ffranr: I've added two more commits to this PR. They add a `RegisterIssuance` method to the base forest. When called this new method will update the base universe and the universe forest DB tables using a single DB transaction.

I don't think `DiffEngine` should include that sort of write methods.

config.go
ffranr: I noticed that `MintingArchiveConfig` had a `UniverseForest` field. I was planning on using the `MintingArchive` to update the universe forest (and base universe simultaneously). So I thought it would make sense to generalise the field name `BaseUniverse` to reflect that the forest might be updated also.

Should I not use the minting archive to update the universe forest?

Roasbeef: >  I was planning on using the MintingArchive to update the universe forest (and base universe simultaneously). So I thought it would make sense to generalize the field name BaseUniverse to reflect that the forest might be updated also.


So the way I imagine it, the minting archive itself doesn't really change much. Instead, the interface that gives it `RegisterIssuance` is what'll be swapped out to instead point to the on-disk multi-verse, that handles all the normal issuance stuff, but then also updates the multiverse tree.

tapdb/universe.go
```diff
@@ -264,6 +264,37 @@ func (b *BaseUniverseTree) RegisterIssuance(ctx context.Context,
...
 	metaReveal *proof.MetaReveal) (*universe.IssuanceProof, error) {
 
+	var (
+		writeTx BaseUniverseStoreOptions
+
+		err           error
+		issuanceProof *universe.IssuanceProof
+	)
+	dbErr := b.db.ExecTx(ctx, &writeTx, func(dbTx BaseUniverseStore) error {
+		issuanceProof, err = universeRegisterIssuance(
```
Roasbeef, line 274: 👍 

this is exactly what I had in mind re db tx pas thru to support re-use elsewhere.

Roasbeef: Still getting through the diff, but we'll also want to do this for the fetch call as well, since we'll want to fetch the other inclusion proof into the universe tree after that.

tapdb/universe_forest.go
```diff
@@ -125,3 +128,36 @@ func (b *BaseUniverseForest) RootNodes(
...
+		issuanceProof *universe.IssuanceProof
+	)
+
+	dbErr := b.db.ExecTx(
+		ctx, &writeTx, func(dbTx BaseUniverseForestStore) error {
+			issuanceProof, err = universeRegisterIssuance(
+				ctx, dbTx, id, key, leaf, metaReveal,
+			)
+
+			// TODO(ffranr): update universe forest
```
Roasbeef, line 152: Might be useful to have it return stuff like the new universe root for that asset (the above func) so this can then be inserted into the multiverse tree.

universe/interface.go
```diff
@@ -194,6 +194,10 @@ type BaseForest interface {
 	// of assets tracked in the base Universe.
 	RootNodes(ctx context.Context) ([]BaseRoot, error)
 
+	RegisterIssuance(ctx context.Context, id Identifier, key BaseKey,
```
Roasbeef, line 197: Doesn't need to be done now, but eventually we'll also want to distinguish from the a multi-verse of only issuance, from one of all transfers for an asset, from one of all transfers for any asset. 

Also notice the other `BaseUniverse` interface doesn't actually take an ID as it's bound to that asset id or group key. As this is a base multi-verse/forst, I think we'd also want to carry that over. You'd then have something similar to the current base uni struct here that is able to fetch the multiverse needed for the action.

config.go
ffranr: I've reverted this change.

tapdb/universe_forest.go
ffranr: It was included in the returned issuance proof, but I think it's a good idea to return it separately also. :+1:

ffranr:
I don't intend on adding more to this PR (except when responding to reviews). I think it's good to go as a refactoring PR. I'll make other PRs for the rest of the work.

tapdb/universe_forest.go
```diff
@@ -40,7 +40,8 @@ func NewBaseUniverseForestReadTx() BaseUniverseForestOptions {
 }
 
 // BasedUniverseForest is a wrapper around the base universe forest that allows
-// us perform batch queries with all the relevant query interfaces.
+// us perform batch transactional databse queries with all the relevant query
```
guggero, line 43: nit: allows us _to_ perform.

tapdb/universe.go
```diff
@@ -425,9 +425,41 @@ func universeRegisterIssuance(ctx context.Context, dbTx BaseUniverseStore,
...
-	// of a specific issuance, or all of the issuances for that minting
-	// outpoint.
+	var (
+		readTx = NewBaseUniverseReadTx()
+
+		err    error
+		proofs []*universe.IssuanceProof
+	)
+
+	err = b.db.ExecTx(ctx, &readTx, func(dbTx BaseUniverseStore) error {
```
guggero, line 435: Should we stick with the `dbErr` naming here? I guess otherwise it's a bit confusing as the `err` will be overwritten. So I'd move the `var err error` inside the closure, then just have `dbErr` on the outside.

guggero (approved):

tapdb/universe.go
```diff
@@ -311,6 +311,46 @@ func (b *BaseUniverseTree) RegisterIssuance(ctx context.Context,
...
+
+// universeRegisterIssuance inserts a new minting leaf within the universe
+// tree, stored at the base key.
+//
+// This function returns the newly registered issuance proof and the new
+// universe root.
+//
+// NOTE: This function accepts a db transaction, as it's used when making
+// broader DB updates.
+func universeRegisterIssuance(ctx context.Context, dbTx BaseUniverseStore,
```
jharveyb, line 346: This is a neat pattern!

jharveyb (approved):
LGTM! Looking forward to the follow-up PRs.

Feels like commits 1-3 and 4-5 can be squashed together, but non-blocking.

Issue #313, opened 2023-05-22
tarogarden: universe stats DB issues after batch confirmation


jharveyb:
Seems like a possible race in updating the local base universe after a minting batch is confirmed.

Log line:

```
2023-05-17 09:55:14.793 [WRN] UNIV: unable to log new proof event: unknown sqlite error: constraint failed: NOT NULL constraint failed: univ
erse_events.universe_root_id (1299)
```

Source:

https://github.com/lightninglabs/taproot-assets/blob/1a3f86c09ac044ff3b9e27cf90a314eecb2e0625/universe/base.go#L220

Source of SQL error:

https://github.com/lightninglabs/taproot-assets/blob/1a3f86c09ac044ff3b9e27cf90a314eecb2e0625/tapdb/sqlc/queries/universe.sql#L101

Callsite in minter:

https://github.com/lightninglabs/taproot-assets/blob/1a3f86c09ac044ff3b9e27cf90a314eecb2e0625/tapgarden/caretaker.go#L946

AFAICT we create the universe root inside `RegisterIssuance` here:

https://github.com/lightninglabs/taproot-assets/blob/1a3f86c09ac044ff3b9e27cf90a314eecb2e0625/tapdb/universe.go#L309

But since the call to log this new proof event is in a separate goroutine, it may execute before the matching universe root has been created. This error line appeared consistently so I don't think it was an intermittent race or error condition.

PR #348, opened 2023-06-08
tapdb/universe: fix universe event logging for grouped assets

Roasbeef:
In this commit, we fix the universe event logging for grouped assets. As is, when we go to insert assets with a group, then the "first" inserted asset is what's known as the `asset_id` in the `universe_roots` table. When we go to log a proof insert/sync event, and the even only has an asset ID, we may end up with a failed query, as only the "first" asset ID would return the row we want.

In order to fix this, rather than trying to match the asset ID with a root, we instead find the _leaf_ for that asset ID, then use that to get the primary key of the universe root via it's foreign key ref.

Fixes https://github.com/lightninglabs/taproot-assets/issues/313

### DIFF
--- a/tapdb/sqlc/queries/universe.sql
+++ b/tapdb/sqlc/queries/universe.sql
@@ -87,9 +87,11 @@ SELECT * FROM universe_servers;
 
 -- name: InsertNewSyncEvent :exec
 WITH root_asset_id AS (
-    SELECT id
-    FROM universe_roots
-    WHERE asset_id = @asset_id
+    SELECT leaves.universe_root_id AS id
+    FROM universe_leaves leaves
+    JOIN genesis_info_view gen
+        ON leaves.asset_genesis_id = gen.gen_asset_id
+    WHERE gen.asset_id = @asset_id
 )
 INSERT INTO universe_events (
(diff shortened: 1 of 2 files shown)


jharveyb:
These errors are no longer present when registering asset issuance after minting, but I still see them for universe sync events.

Can reproduce with `make itest icase=universe_sync`, source:

https://github.com/lightninglabs/taproot-assets/blob/c037025ebb22c54087736589bd31fb570a04b738/universe/syncer.go#L189

guggero (approved):

jharveyb (approved):
LGTM 🚀

dstadulis:
> These errors are no longer present when registering asset issuance after minting, but I still see them for universe sync events.
> 
> Can reproduce with `make itest icase=universe_sync`, source:
> 
> https://github.com/lightninglabs/taproot-assets/blob/c037025ebb22c54087736589bd31fb570a04b738/universe/syncer.go#L189

JHB didn't block PR but will need to continue https://github.com/lightninglabs/taproot-assets/issues/354

ffranr:
This has been deployed to testnet and staging.

PR #351, opened 2023-06-09
itest: basic send test re-uses the same recv address for multiple sends

ffranr:
This aim of this change is just to make it clear that an address can be re-used for multiple send events.


jharveyb:
The address here isn't being reused; the asset ID and amount are the same but the address should have a fresh script key on each loop iteration.

jharveyb:
Could rename this test to something like `testBasicSendUnidirectional`, since we can't run it individually as is. 

`make itest icase=basic_send` will run both this and `testBasicSendPassiveAsset` right now.

Roasbeef (approved):
LGTM 🌪️

Issue #332, opened 2023-05-31
transaction rejected: output already spent


bilthon:
I'm running `tapd` v0.2.0-alpha on polar and get this error in one instance:

```
2023-05-31 16:05:31.003 [INF] RPCS: Starting RPC Server
2023-05-31 16:05:31.002 [INF] FRTR: Attempting to resume delivery for anchor_txid=22cf1da4f9ace79187227298c490a0b497ed21a2cabebfcedac40b8b5a61b28d
2023-05-31 16:05:31.002 [INF] FRTR: Attempting to resume delivery for anchor_txid=b990385ec0db9375f5fce1ae33052ef414aa1a5429b9becf68d2301a52f49011
2023-05-31 16:05:31.004 [INF] RPCS: RPC server listening on 0.0.0.0:10029
2023-05-31 16:05:31.005 [INF] GRDN: Resuming 0 pending inbound asset events
2023-05-31 16:05:31.006 [INF] GRDN: Loading wallet transactions starting at block height 0
2023-05-31 16:05:31.005 [INF] FRTR: ChainPorter executing state: SendStateBroadcast
2023-05-31 16:05:31.005 [INF] FRTR: ChainPorter executing state: SendStateBroadcast
2023-05-31 16:05:31.006 [INF] CONF: Starting HTTPS REST proxy listener at 0.0.0.0:8089
2023-05-31 16:05:31.008 [INF] RPCS: gRPC proxy started at 0.0.0.0:8089
2023-05-31 16:05:31.009 [INF] SRVR: Taproot Asset Daemon fully active!
2023-05-31 16:05:31.008 [INF] FRTR: Broadcasting new transfer tx, txid=22cf1da4f9ace79187227298c490a0b497ed21a2cabebfcedac40b8b5a61b28d
2023-05-31 16:05:31.008 [INF] GRDN: Checking 5 wallet transactions for inbound assets, this might take a while
2023-05-31 16:05:31.011 [DBG] GRDN: Inspecting tx b990385ec0db9375f5fce1ae33052ef414aa1a5429b9becf68d2301a52f49011 for Taproot Asset address outputs
2023-05-31 16:05:31.009 [INF] FRTR: Broadcasting new transfer tx, txid=b990385ec0db9375f5fce1ae33052ef414aa1a5429b9becf68d2301a52f49011
[...]

PR #355, opened 2023-06-12
itest: asset send procedure reserves UTXO

ffranr:
This PR adds an itest which ensures that the asset send procedure reserves the spending UTXO.

I originally created this test to try to mess with UTXOs while the send procedure was pending.

Created as an attempt to solve: https://github.com/lightninglabs/taproot-assets/issues/332


ffranr:
This test doesn't currently work in conjunction with the other existing tests. A side effect from the other tests is affecting the test added in this PR.

Issue #315, opened 2023-05-22
tarogarden: cache block header verifier results during batch confirmation


jharveyb:
Related to #314.

Generating proofs for a large batch is liable to time out due to delay from fetching blocks via RPC in the `BatchStateConfirmed` state.

Specifically, we create a `GenHeaderVerifier` that will be called at least twice, during proof generation and import. In this part of the state machine, we are repeatedly fetching the same block, which we already fetched and stored before generating these proofs.

We could replace the 'real' headerVerifier with a mock verifier that just checks for block header equality. If we wanted to use header verification for some purpose like reorg detection, it likely shouldn't be done for each proof.

jharveyb:
This is more tractable now that we have re-org handling. P sure what we want here is to use the caching verifier in `BatchStateBroadcast`, but NOT use it in `planter.updateMintingProofs()`, which is what actually handles re-orgs.

We also already have a template for caching verifier callbacks than use an LRU cache. To properly test this we should update the mint batch stress test to generate large blocks.

jharveyb:
This is more tractable now that we have re-org handling. P sure what we want here is to use the caching verifier in `BatchStateBroadcast`, but NOT use it in `planter.updateMintingProofs()`, which is what actually handles re-orgs.

We also already have a template for caching verifier callbacks than use an LRU cache. To properly test this we should update the mint batch stress test to generate large blocks.

jharveyb:
Related, we can now use https://lightning.engineering/api-docs/api/lnd/neutrino-kit/get-block-header in the non-caching header verifier so speed that up as well.

guggero:
That API is only available to Neutrino backends. We should add a generic one to the chain kit RPC service.

jharveyb:
Can implement this once lnd 0.17.1 is taped out. We'll need a version check somewhere to create the correct header verifier to match the lnd version.

jharveyb:
Specifically, we replace this `GetBlock` call IIUC?

https://github.com/lightninglabs/taproot-assets/blob/3e6579cbacb839acdc6bcc73c622bab7f91ac881/chain_bridge.go#L125

dstadulis:
@jharveyb has a draft branch of this which works.

Will post it after https://github.com/lightninglabs/lndclient/pull/161 is merged

dstadulis:
will need to Tag release for lnd-client after lnd release arrives

Issue #314, opened 2023-05-22
tarogarden: low timeouts and perf issues when finalizing a batch of 10k assets in one asset group


jharveyb:
The planter, caretaker, and custodian all share the same default timeout of 30 seconds.

https://github.com/lightninglabs/taproot-assets/blob/1a3f86c09ac044ff3b9e27cf90a314eecb2e0625/tapgarden/caretaker.go#L51

In the caretaker, the timeout is used to make new contexts for each state machine step; so each is expected to complete within that 30 second deadline.

In `BatchStateFrzoen`, we promote all seedlings to sprouts and write them to disk. This step timed out with a minting batch of 10,000 assets, all in one asset group, with no script keys specified, and a 4.1k metadata preimage for each asset.

I suspect the most significant bottlenecks here are group key signing / tweaking, script key generation, and metadata hashing. From observing the machine usage during minting, the main bottleneck seemed to be key derivation.

I think we could derive script keys and apply BIP-0086 tweaks earlier in this state, and in parallel with something similar to an `ErrGroup`, since the derivation order does not need to match the seedling's place in the batch, and we once we enter this state we know exactly how many script keys are needed.

PR #357, opened 2023-06-13
tapgarden: minter perf fixes

jharveyb:
Addresses issues #314 and #315.

Implements a header verifier that caches the correct block hash instead making multiple RPC calls for each minting proof generated, and also moves some slower methods in the sprout creation loop into separate goroutines.

It would be better to use a `threadPool` there instead of starting & stopping the goroutines manually. I think at this point the sprout creation loop is current bottlenecked on key tweaking for new group keys, but not entirely sure.

### DIFF
--- a/tapgarden/caretaker.go
+++ b/tapgarden/caretaker.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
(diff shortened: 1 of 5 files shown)


Roasbeef:
I think these optimizations are still relevant? cc @positiveblue

jharveyb:
Sort of, I think the caching header verifier is but that isn't evident when using a network with empty blocks like the itest conditions. After further testing I don't think the script key parallel calls are really relevant.

I think the (more complex) bottleneck is the proof insertion & DB operations that happen after TX confirmation, but there are pending PRs for batch insertion now. Would be good to get data on perf before & after those changes.i

jharveyb:
Mostly replaced by `fn.ParSlice` and existing work on batching requests.

PR #358, opened 2023-06-13
tapfreighter: resume pending parcels using the main porter goroutine

ffranr:
In this commit, we ensure that pending parcels are resumed using the same goroutine that processes the normal send parcel processing flow.

The goal of this change is to ensure that the locking mechanism enforced by a singular parcel processing goroutine is not compromised by the execution of a secondary parcel goroutine.

### DIFF
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -116,31 +116,34 @@ func (p *ChainPorter) Start() error {
 	p.startOnce.Do(func() {
 		log.Infof("Starting ChainPorter")
 
-		// Before we re-launch the main goroutine, we'll make sure to
-		// restart any other incomplete sends that may or may not have
-		// had the transaction broadcaster.
+		// Start the main chain porter goroutine.
+		p.Wg.Add(1)
+		go p.assetsPorter()
+
+		// Identify any pending parcels that need to be resumed and add
+		// them to the exportReqs channel so they can be processed by
+		// the main porter goroutine.
 		ctx, cancel := p.WithCtxQuit()
 		defer cancel()
-		pendingParcels, err := p.cfg.ExportLog.PendingParcels(ctx)
+		outboundParcels, err := p.cfg.ExportLog.PendingParcels(ctx)
 		if err != nil {
 			startErr = err
 			return
 		}
 
-		log.Infof("Resuming delivery of %v pending asset parcels",
-			len(pendingParcels))
+		// We resume delivery using the normal parcel delivery mechanism
+		// by converting the outbound parcels into pending parcels.
+		for idx := range outboundParcels {
+			outboundParcel := outboundParcels[idx]
+			log.Infof("Attempting to resume delivery for "+
+				"anchor_txid=%v",
+				outboundParcel.AnchorTx.TxHash().String())
 
-		// Now that we have the set of pending sends, we'll make a new
-		// goroutine that'll drive the state machine till the broadcast
-		// point (which we might be repeating), and final terminal
-		// state.
-		for _, parcel := range pendingParcels {
-			p.Wg.Add(1)
-			go p.resumePendingParcel(parcel)
+			// At this point the asset porter should be running.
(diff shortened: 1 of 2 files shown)


tapfreighter/chain_porter.go
```diff
@@ -116,31 +116,39 @@ func (p *ChainPorter) Start() error {
...
-		// Now that we have the set of pending sends, we'll make a new
-		// goroutine that'll drive the state machine till the broadcast
-		// point (which we might be repeating), and final terminal
-		// state.
+		// At this point the asset porter should be running, so it will
+		// pick up the pending parcels and attempt to deliver them.
 		for _, parcel := range pendingParcels {
-			p.Wg.Add(1)
-			go p.resumePendingParcel(parcel)
+			p.exportReqs <- parcel
```
guggero, line 150: Why not just:
```
		for idx := range outboundParcels {
			p.exportReqs <- NewPendingParcel(outboundParcels[idx])
		}
```

Then we don't need the loop above. And fix a potential loop variable issue.

guggero: Also, should we add the log message back here?
```
	log.Infof("Attempting to resume delivery for anchor_txid=%v",
		pkg.AnchorTx.TxHash().String())
```

guggero (commented):
Very nice fix!

Does that mean you were successful in reproducing the error described in https://github.com/lightninglabs/taproot-assets/issues/332? If yes, should we link the issue to the PR so it's closed by it?

jharveyb:
Started testing this manually with scenario 2 from here: https://github.com/lightninglabs/taproot-assets/issues/332#issuecomment-1588932991

I think we can actually fix this another way. If we ignore the error `output already spent` from the `PublsihTransaction` call in `SendStateBroadcast` here:

https://github.com/lightninglabs/taproot-assets/blob/f46f3f10d9032e87a8cab2d2a328e8aa6334b242/tapgarden/caretaker.go#L716

Then, for already confirmed transactions, we should recognize the confirmation in the next state and continue as normal instead of shutting down the caretaker. As-is I think we would still fail if the send TX already confirmed.

guggero:
> If we ignore the error output already spent from the PublsihTransaction call in SendStateBroadcast here

I don't think we want to ignore the error in the current configuration. The `output already spent` error happens because we don't have any asset coin locking implemented yet. So we avoid running into problems by only having a single goroutine active for issuing transfers. So a user is forced to wait between sends until the first one confirms.
If a user instead restarts the daemon before the first transfer is confirmed, the restart continues the transfer in a separate goroutine, which makes it possible for another one being issued through the RPC in the meantime. And because we don't have asset coin locking, it will select the same on-chain input for the second transfer and it will never go through.
This PR fixes that issue by using the same goroutine for resuming parcels that is also used for handling new parcels from RPC requests.

jharveyb:
That makes sense; but I think with these changes the resume of the parcel will still fail if the send TX was already confirmed, and we won't reach the later states of the send logic?

I think maybe we would want to check for send TX confirmation and jump past re-broadcasting if confirmed.

I'll read more closely + try and repro again.

ffranr:
> we would want to check for send TX confirmation and jump past re-broadcasting if confirmed

That sounds right to me @jharveyb  !

jharveyb:
> > we would want to check for send TX confirmation and jump past re-broadcasting if confirmed
> 
> That sounds right to me @jharveyb !

We already have the appropriate logic in `waitForTransferTxConf`, but I'm not sure if that will pick up a confirmation in a previous block; maybe we can request a scan forward from the height hint to the present block (do we care if the TX is not already confirmed, but already in the mempool? I think not), and then move to broadcast if that fails instead of waiting for a confirmation that would not arrive.

ffranr:
@jharveyb @guggero I think we should still merge this PR regardless. I think it simplifies the handling of pending packages. I don't see why we should start a goroutine for each pending package, feels like premature optimisation.

guggero (approved):

jharveyb (approved):

Issue #350, opened 2023-06-09
[universes] add validation to prevent universe from adding federation connection with itself / loopback


dstadulis:
Currently, it's possible for federations to erroneously/needlessly add their own address as a federation member.  This behavior is errant and can cause unnecessary resource consumption.

the federation addition code should be modified to validate that the universe endpoint / resource being added is not itself.

PR #364, opened 2023-06-15
universe: add server info RPC, check we're not connecting to ourselves

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/350.

### DIFF
--- a/itest/universe_test.go
+++ b/itest/universe_test.go
@@ -314,17 +314,39 @@ func testUniverseFederation(t *harnessTest) {
 		require.NoError(t.t, bob.stop(true))
 	}()
 
-	ctx := context.Background()
+	ctxb := context.Background()
+	ctxt, cancel := context.WithTimeout(ctxb, defaultWaitTimeout)
+	defer cancel()
 
 	// Now that Bob is active, we'll make a set of assets with the main node.
 	firstAsset := mintAssetsConfirmBatch(t, t.tapd, simpleAssets[:1])
 	require.Len(t.t, firstAsset, 1)
 
+	// Make sure we can't add ourselves to the universe.
+	_, err := t.tapd.AddFederationServer(
+		ctxt, &unirpc.AddFederationServerRequest{
+			Servers: []*unirpc.UniverseFederationServer{{
+				Host: t.tapd.rpcHost(),
+			}},
+		},
+	)
+	require.ErrorContains(t.t, err, "cannot add ourselves")
+
+	// Make sure we can't add an invalid server to the universe.
+	_, err = t.tapd.AddFederationServer(
+		ctxt, &unirpc.AddFederationServerRequest{
+			Servers: []*unirpc.UniverseFederationServer{{
+				Host: "foobar this is not even a valid address",
+			}},
+		},
+	)
+	require.ErrorContains(t.t, err, "no such host")
+
 	// We'll now add the main node, as a member of Bob's Universe
 	// federation. We expect that their state is synchronized shortly after
 	// the call returns.
-	_, err := bob.AddFederationServer(
-		ctx, &unirpc.AddFederationServerRequest{
+	_, err = bob.AddFederationServer(
+		ctxt, &unirpc.AddFederationServerRequest{
 			Servers: []*unirpc.UniverseFederationServer{
 				{
 					Host: t.tapd.rpcHost(),
@@ -337,7 +359,7 @@ func testUniverseFederation(t *harnessTest) {
 	// If we fetch the set of federation nodes, then the main node should
 	// be shown as being a part of that set.
 	fedNodes, err := bob.ListFederationServers(
-		ctx, &unirpc.ListFederationServersRequest{},
+		ctxt, &unirpc.ListFederationServersRequest{},
 	)
 	require.NoError(t.t, err)
 	require.Equal(t.t, 1, len(fedNodes.Servers))
@@ -352,6 +374,11 @@ func testUniverseFederation(t *harnessTest) {
 	// should also be able to query for stats specifically for the asset.
 	assertUniverseStats(t.t, bob, 1, 0, 1)
 
+	// Test the content of the universe info call.
+	info, err := bob.Info(ctxt, &unirpc.InfoRequest{})
+	require.NoError(t.t, err)
+	require.EqualValues(t.t, 1, info.NumAssets)
+
 	// We'll now make a new asset with Bob, and ensure that the state is
 	// properly pushed to the main node which is a part of the federation.
 	newAsset := mintAssetsConfirmBatch(t, bob, simpleAssets[1:])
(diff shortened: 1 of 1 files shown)


rpcserver.go
```diff
@@ -2705,6 +2733,33 @@ func (r *rpcServer) AddFederationServer(ctx context.Context,
 
 	serversToAdd := fn.Map(in.Servers, unmarshalUniverseServer)
 
+	// Before we add the server as a federation member, we check that we can
+	// actually connect to it and that it isn't ourselves.
+	for idx := range serversToAdd {
```
jharveyb, line 2738: I think this loop + info check prevents fixes the issue when trying to add ourself via RPC, but not via config file / at startup. I think this should be moved deeper into the universe code, into `AddServer`:

https://github.com/lightninglabs/taproot-assets/blob/e74932ba9182a773f68252fe6a0f4b34144443c9/universe/auto_syncer.go#L364

So that it's also executed at start.

jharveyb (commented):
LGTM, pending the possible change to execute at startup.

perms/perms.go
```diff
@@ -185,5 +185,6 @@ var (
 		"/universerpc.Universe/AssetLeaves":     {},
 		"/universerpc.Universe/QueryProof":      {},
 		"/universerpc.Universe/InsertProof":     {},
+		"/universerpc.Universe/Info":            {},
```
Roasbeef, line 188: Hmm, do we actually want people to be able to hit this?

universe_rpc_registrar.go
```diff
@@ -128,4 +111,34 @@ func (r *RpcUniverseRegistar) RegisterIssuance(ctx context.Context,
 
 // A compile time interface to ensure that RpcUniverseRegistrar implements the
 // universe.Registrar interface.
-var _ universe.Registrar = (*RpcUniverseRegistar)(nil)
+var _ universe.Registrar = (*RpcUniverseRegistrar)(nil)
+
+// ConnectUniverse connects to a remote Universe server using the provided
+// server address.
+func ConnectUniverse(
```
Roasbeef, line 118: 👍

rpcserver.go
```diff
@@ -2733,6 +2733,33 @@ func (r *rpcServer) AddFederationServer(ctx context.Context,
...
+		}
+
+		// We don't allow adding ourselves as a federation member.
+		info, err := conn.Info(ctx, &unirpc.InfoRequest{})
+		if err != nil {
+			return nil, fmt.Errorf("error getting info from "+
+				"server %v: %w", server.HostStr(), err)
+		}
+
+		if info.RuntimeId == r.runtimeID {
```
Roasbeef, line 2757: Hmm, yeah seems simple enough, wondering if there're other alternatives we should consider 🤔

Roasbeef: Also fwiw, this isn't the call path that would lead to us adding ourselves, instead it's here: https://github.com/lightninglabs/taproot-assets/blob/7ab8a5760d671d7446d31871b60ba699618abbad/tapcfg/server.go#L180-L201


https://github.com/lightninglabs/taproot-assets/blob/main/universe/auto_syncer.go#L105-L118

This'll actually end up skipping this RPC path.

Roasbeef (changes requested):

rpcserver.go
guggero: Hmm, I somehow managed to completely forget that we had that other code path from the config side. Which was the whole point of https://github.com/lightninglabs/taproot-assets/issues/350 now that I think about it... Anyway, changed things a bit now to include that other code path as well.

perms/perms.go
guggero: If we want this simple check to work, then yes, we need users to be able to connect directly without a macaroon and query this info. That's why I chose not to include any sensitive information. Maybe we even need to remove the two counts we have as those issue DB queries. If that is a concern and we just want to use this for the ID, maybe we could rename the RPC to just `ServerID`?

rpcserver.go
guggero: > wondering if there're other alternatives we should consider

I thought about it a bit and I think it's not trivial to find out if we are connecting to ourselves just from the URL/domain name. Because what we have configured as our RPC listener port might be completely different from the domain name/port we are given to connect to (due to load balancers and other things inbetween). So I thought this was the simplest approach.
But I'm open to other approaches.

> Also fwiw, this isn't the call path that would lead to us adding ourselves, instead it's here: 

Yeah, I somehow forgot about that path before. Fixed now.

jharveyb (approved):
LGTM! No strong feeling about if we should leave the DB calls in.

rpcserver.go
Roasbeef: > But I'm open to other approaches.


Hmm, good point about the ports, etc....not sure there's a good option other than this. Only other thing can think of is extending to also support using `brontide` over gRPC, then we'd have the public key we can check against. 

Also if we want a bit more assurance that we won't have to worry about collisions here, then we can sample a `[32]byte` with the CSPRNG. I think we can run with this for now, then circle back to see if was have false positives in the wild (a bit of extra logging can help us detect that). 

In terms of gaming, worst thing I can think of is someone sort of bans a server from connecting to them as they mimic their runtime ID, but then they can already to that by banning the IP itself.

perms/perms.go
Roasbeef: > Maybe we even need to remove the two counts we have as those issue DB queries. 

In theory users can already obtain this information if the universe server is open, albeit with even more DB queries.

rpcserver.go
```diff
@@ -2581,6 +2581,27 @@ func (r *rpcServer) InsertProof(ctx context.Context,
...
+
+	uniServers, err := r.cfg.FederationDB.UniverseServers(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return &unirpc.InfoResponse{
+		RuntimeId:      r.runtimeID,
+		NumAssets:      universeStats.NumTotalAssets,
+		NumSyncServers: uint32(len(uniServers)),
```
Roasbeef, line 2601: I think let's remove this bit? Given this is information that would be shown which isn't leakd anywhere else in the set of calls.

Roasbeef (changes requested):

rpcserver.go
guggero: Okay, makes sense, removed.

rpcserver.go
guggero: Wouldn't we also need to know the public key of the server when using `brontide`? Or is there a variant of the Noise protocol where you learn the other side's long-term keys without also doing SPAKE?

I don't worry about collision too much. If there was a collision between my local daemon and the universe, I would just restart my daemon and the collision would be gone.
And I agree about banning, I don't think we introduce any new assumptions regarding that with this approach.

Roasbeef (approved):
LGTM 🥬

Roasbeef:
Needs a rebase!

PR #365, opened 2023-06-15
itest: resume pending package send

ffranr:
This PR adds an itest which ensure that we resume the send procedure which was interrupted by a tapd restart.

### DIFF
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -127,6 +127,98 @@ func testBasicSend(t *harnessTest) {
 	wg.Wait()
 }
 
+// testResumePendingPackageSend tests that we can properly resume a pending
+// package send after a restart.
+func testResumePendingPackageSend(t *harnessTest) {
+	ctxb := context.Background()
+
+	sendTapd := t.tapd
+
+	// Setup a receiver node.
+	recvLnd := t.lndHarness.Bob
+	recvTapd := setupTapdHarness(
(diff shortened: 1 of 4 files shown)


itest/send_test.go
```diff
@@ -127,6 +127,98 @@ func testBasicSend(t *harnessTest) {
...
+		},
+	)
+	require.NoError(t.t, err)
+	assertAddrCreated(t.t, recvTapd, rpcAssets[0], recvAddr)
+
+	// We will now start two asset send events in sequence. We will stop and
+	// restart the sending node during each send. During one sending event
+	// we will mine whilst the sending node is stopped. During the other
+	// sending event we will only mine once the sending node is restarted.
+	for i := range []int{0, 1} {
```
guggero, line 176: This works as intended, even though `i` will contain the _index_, not the slice values. Could also just use the c-like syntax: `for i := 0; i < 2; i++ {`.

guggero (approved):
Nice test, LGTM :tada:

jharveyb (approved):
Nice to have this in (and that restart works!) 🚀

itest/send_test.go
ffranr: Will fix in a future PR, thanks.

ffranr: Fixed in this PR: https://github.com/lightninglabs/taproot-assets/pull/371

PR #366, opened 2023-06-16
Fix universe sync test

guggero:
Found while trying to reproduce a different error.
Looks like we didn't really compare the correct universe roots in the integration tests.

### DIFF
--- a/itest/assertions.go
+++ b/itest/assertions.go
@@ -728,7 +728,7 @@ func assertUniverseStateEqual(t *testing.T, a, b *tapdHarness) {
 	rootsA, err := a.AssetRoots(ctxb, &unirpc.AssetRootRequest{})
 	require.NoError(t, err)
 
-	rootsB, err := a.AssetRoots(ctxb, &unirpc.AssetRootRequest{})
+	rootsB, err := b.AssetRoots(ctxb, &unirpc.AssetRootRequest{})
 	require.NoError(t, err)
 
 	assertUniverseRootsEqual(t, rootsA, rootsB)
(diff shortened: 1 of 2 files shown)


jharveyb (approved):
Good catch! 💯

itest/assertions.go
```diff
@@ -722,13 +723,29 @@ func assertUniverseRootsEqual(t *testing.T, a, b *unirpc.AssetRootResponse) {
...
+}
+
 func assertUniverseStateEqual(t *testing.T, a, b *tapdHarness) {
 	ctxb := context.Background()
 
 	rootsA, err := a.AssetRoots(ctxb, &unirpc.AssetRootRequest{})
 	require.NoError(t, err)
 
-	rootsB, err := a.AssetRoots(ctxb, &unirpc.AssetRootRequest{})
+	rootsB, err := b.AssetRoots(ctxb, &unirpc.AssetRootRequest{})
```
Roasbeef, line 748: 🙈

itest/assertions.go
```diff
@@ -722,13 +723,29 @@ func assertUniverseRootsEqual(t *testing.T, a, b *unirpc.AssetRootResponse) {
 	}
 }
 
+func succeedEventually(t *testing.T, f func(*testing.T),
+	timeout, retryInterval time.Duration) {
+
+	t.Helper()
+
+	require.Eventually(t, func() bool {
```
Roasbeef, line 731: Ahh, this super useful! 

Otherwise you need to modify things to return an error for the usual `wait` instances.

PR #369, opened 2023-06-17
tapcfg: ensure log rotator created after config validation

Roasbeef:
In this commit, we fix a bug that would cause the logs to always go to the default log location even if a different base dir was set. Before this commit, we'd create the log rotators before the call to `ValidateConfig`, which is where we modify the log dir location to be relative to the final base dir.

The fix here is simple: create the log rotator _after_ the config has been validated.

### DIFF
--- a/tapcfg/config.go
+++ b/tapcfg/config.go
@@ -414,17 +414,6 @@ func LoadConfig(interceptor signal.Interceptor) (*Config, btclog.Logger, error) 
 		return nil, nil, err
 	}
 
-	// Initialize logging at the default logging level.
-	tap.SetupLoggers(cfg.LogWriter, interceptor)
-	err = cfg.LogWriter.InitLogRotator(
-		filepath.Join(cfg.LogDir, defaultLogFilename),
-		cfg.MaxLogFileSize, cfg.MaxLogFiles,
-	)
-	if err != nil {
-		_, _ = fmt.Fprintln(os.Stderr, err.Error())
-		return nil, nil, err
-	}
-
 	cfgLogger := cfg.LogWriter.GenSubLogger("CONF", nil)
 
 	// Make sure everything we just loaded makes sense.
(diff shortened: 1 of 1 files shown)


tapcfg/config.go
```diff
@@ -439,6 +428,17 @@ func LoadConfig(interceptor signal.Interceptor) (*Config, btclog.Logger, error)
 		return nil, nil, err
 	}
 
+	// Initialize logging at the default logging level.
```
guggero, line 431: I think the log call on line 424 and 427 will now go to `stdout` only, since we haven't configured the logger yet? Maybe we should explicitly print those to `stderr` instead?

guggero (commented):
NIce fix!

ffranr (approved):
I think Oli's suggestion [here](https://github.com/lightninglabs/taproot-assets/pull/369/files#r1233633484) makes sense to me.

Another idea that makes sense to me is to complete the log setup as soon as possible after start (so before any non-log config validation).
I would consider refactoring by adding a `ValidateLogConfig` function and calling that before validating any other config fields. That way even config validation log messages should behave as the user has specified.

Either way!

Roasbeef:
> Another idea that makes sense to me is to complete the log setup as soon as possible after start (so before any non-log config validation).


I went this route to minimize changes, and not break the `litd` integration we have (uses these functions, but adds in its own logger).

tapcfg/config.go
Roasbeef: Yeah it'll just go to `stdout`, tested this locally. Anything logged in `ValidateConfig` will also go to `stderr` as well. Basically what happened is that during the refactor for `tapd` stuff, we made some assumptions about the way things are set up. I think we can do another refactor here (as robyn hinted below) to fix things, but then we'd also need to change the `litd` creates the daemons.

Issue #368, opened 2023-06-16
Asset meta data is overwritten incorrectly/not passed around correctly when performing receives


habibitcoin:
Using latest version on main

I've now tried to receive/pass around the metadata for a few different assets, and it seems that receiving an asset causes the meta_data to be incorrectly populated

Step 1: I fetch the meta data for an asset I am interested in receiving. Presumably I was able to fetch this because it's stored in someone's universe

```
habibitcoin@648bbcf308054721fb175b6d:~$ tapcli --rpcserver=193.149.176.138:10029 assets meta --asset_id fead959d3257a9b2f5aea591eac16ade40a9837e45c7b64685b2f313f837e4d8
{
[...]

PR #370, opened 2023-06-19
Asset metadata overwrite fix

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/368.

As @habibitcoin already correctly identified, one of our SQL queries is incorrect and overwrites valid meta information with `NULL` when importing a proof.


ffranr (approved):
Good to know about `COALESCE` :+1:

habibitcoin:
Woot! Confirmed this works locally as well. Thanks!

jharveyb (approved):
LGTM! Now thinking if we need to update other queries to use COALESCE as well.

PR #377, opened 2023-06-22
itest: add missing build tag

jharveyb:
Avoid building itests when building for a release.

### DIFF
--- a/itest/addrs_test.go
+++ b/itest/addrs_test.go
@@ -1,3 +1,5 @@
+//go:build itest
+
 package itest
 
 import (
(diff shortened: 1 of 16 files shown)


guggero:
I don't think we should add build tags to each file, as that makes it a bit harder to use (for example, the itest files aren't compiled anymore when running the unit tests, which is sometimes handy to detect compilation errors early).
Because the release binary shouldn't have any reference to the `itest` package at all, it will not be built by default.
So something is wrong and the offender seems to be `proof/aperture.go`. That needs to be moved to the `itest` package, then we resolve the original issue.

guggero:
Replaced by https://github.com/lightninglabs/taproot-assets/pull/378.

PR #375, opened 2023-06-22
multi: fix aperture config + bump golang version

jharveyb:
In this commit, we fix a test harness config value that overflowed. We also bump the golang version used in all docker images to 1.20.5.

PR #378, opened 2023-06-23
Fix multiple small issues with release, prepare for 0.2.2-alpha

guggero:
Replaces https://github.com/lightninglabs/taproot-assets/pull/377.

Adds some Golang version bumps that were missed in https://github.com/lightninglabs/taproot-assets/pull/375 and updates the comments on what files such a bump needs to update.

Also makes sure that we catch release compilation issues early by running the release build in CI.

Since there was already a tag pushed for `v0.2.1-alpha`, we need to bump the version to `v0.2.2-alpha`.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -22,6 +22,12 @@ env:
   GOPATH: /home/runner/work/go
   GO111MODULE: on
 
+  # If you change this value, please change it in the following files as well:
+  # /Dockerfile
+  # /dev.Dockerfile
+  # /make/builder.Dockerfile
+  # /taprpc/Dockerfile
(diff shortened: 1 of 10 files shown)


.github/workflows/main.yaml
```diff
@@ -77,6 +77,22 @@ jobs:
       - name: compile code
         run: go install -v ./...
 
+  ########################
+  # cross compilation
+  ########################
+  cross-compile:
```
ffranr, line 89: `release-build` instead of `cross-compile`?

ffranr (approved):

jharveyb (approved):
LGTM, verified that all golang versions are bumped.

PR #380, opened 2023-06-26
script: pin GOAMD64 for release builds

jharveyb:
Addresses some build reproducibility issues discovered during the release process for v0.2.2. Builders may have `GOAMD64` set locally, which would lead them to build different binaries for 64-bit machines.

This PR also includes some fixes for warnings flagged by shellcheck.

This fix was tested by changing the `GOAMD64" env. var. locally and re-building.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -111,6 +111,7 @@ function check_tag_correct() {
 
     # If the tapd reported version contains a suffix, remove it, so we can match
     # the tag properly.
+    # shellcheck disable=SC2001
     tapd_version=$(echo "$tapd_version" | sed -e 's/-\(alpha\|beta\)\(\.rc[0-9]\+\)\?//g')
 
     # Match git tag with tapd version.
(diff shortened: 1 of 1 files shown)


guggero (approved):

scripts/release.sh
```diff
@@ -111,6 +111,7 @@ function check_tag_correct() {
 
     # If the tapd reported version contains a suffix, remove it, so we can match
     # the tag properly.
+    # shellcheck disable=SC2001
```
ffranr, line 114: Should this new line be included in this PR?

ffranr (approved):

ffranr:
I guess we don't have a bash linter, but changes like
`rm -rf $maindir` -> `rm -rf "$maindir"` should have been cought by it.

scripts/release.sh
jharveyb: Yes, shellcheck (the linter I used to find these issues) flags this sed expression but I don't think we want to replace it with something bash-specific / less portable.

PR #381, opened 2023-06-27
asset: Add Asset versioning support - V1

jharveyb:
Fixes https://github.com/lightninglabs/taproot-assets/issues/392.


jharveyb:
This PR is now part 1 of 3ish for actually bumping the asset version.

This includes the new fields in the `Asset` struct, without being populated, new encoders/decoders, the new TLV numbers + record funcs, and rejection of unknown asset versions before they are used in an MS-SMT.

One piece missing is rejection of unknown assets for RPC inputs.

asset/asset.go
```diff
@@ -22,12 +23,20 @@ import (
 // compressed, 33-byte form.
 type SerializedKey [33]byte
 
+// SchnorrKey is a type for representing an x-only public key, serialized in the
+// 32-byte form.
+type SchnorrKey [32]byte
```
Roasbeef, line 29: 👍

asset/asset.go
```diff
@@ -22,12 +23,20 @@ import (
...
+
 // SchnorrSerialized returns the Schnorr serialized, x-only 32-byte
 // representation of the serialized key.
 func (s SerializedKey) SchnorrSerialized() []byte {
 	return s[1:]
 }
 
+func (s SchnorrKey) SchnorrSerialized() []byte {
```
Roasbeef, line 42: Should we start to make this into a slim interface? Or are the type defs above sufficient?

asset/asset.go
```diff
@@ -36,6 +45,15 @@ func (s SerializedKey) CopyBytes() []byte {
 	return c
 }
 
+// CopyBytes returns a copy of the underlying array as a byte slice.
+func (s SchnorrKey) CopyBytes() []byte {
+	return fn.CopyBytes(s[:])
+}
+
+func (s SchnorrKey) IsZero() bool {
```
Roasbeef, line 59: Missing `godoc` comment.

asset/asset.go
```diff
@@ -53,6 +79,8 @@ var (
 	// asset split leaves.
 	ZeroPrevID PrevID
 
+	ZeroGroupKey [32]byte
```
Roasbeef, line 88: Missing `godoc` comment. Also looks like we can use `SchnorrKey` here.

asset/asset.go
```diff
@@ -658,6 +686,14 @@ type Asset struct {
 	// its unique ID within the Taproot Asset protocol.
 	Genesis
 
+	// Type uniquely identifies the type of Taproot asset.
+	ExplicitType Type
```
Roasbeef, line 703: Why's this needed? 

I would imagine that we can actually keep the `Genesis` struct here as is. The difference would be that upon decode, you would then "supplement" this information from the DB or elsewhere. In most cases, you still want to know this information, but you might obtain it in an extra step. The universe changes made a similar gap where we already know the genesis before decoding the asset, so we supplement that as needed.

asset/asset.go
```diff
@@ -692,6 +728,11 @@ type Asset struct {
 	// together across distinct asset IDs, allowing further issuance of the
 	// asset to be made possible.
 	GroupKey *GroupKey
+
+	// XOnlyGroupKey is the tweaked public key that is used to associate
+	// assets together across distinct asset IDs, allowing further issuance
+	// of the asset to be made possible.
+	XOnlyGroupKey SchnorrKey
```
Roasbeef, line 748: Similar here as well: for a logical asset we want to know this information, but it can be obtained via another source.

asset/records.go
```diff
@@ -31,6 +26,15 @@ const (
...
 	LeafGroupKey            LeafTlvType = 10
+
+	// Types for future asset format.
+	// TODO: LeafID and LeafGenesis have conflicting types. The BIP
+	// specifies LeafID, but this would require wallets to store the
+	// preimage to the asset ID elsewhere, when it could be included in the
+	// TLV itself as done with LeafGenesis.
+	LeafAssetID       LeafTlvType = 2
+	LeafAssetType     LeafTlvType = 4
+	LeafXOnlyGroupKey LeafTlvType = 17
```
Roasbeef, line 37: 17? 

Why also why would ID and type need to be serialized? You only need the asset ID. You already have the genesis after initial boostrap, so from that you can verify that the asset ID here is valid.

asset/records.go
```diff
@@ -49,8 +53,11 @@ func NewLeafVersionRecord(version *Version) tlv.Record {
 	)
 }
 
-func NewLeafIDRecord(id *[sha256.Size]byte) tlv.Record {
-	return tlv.MakePrimitiveRecord(LeafID, id)
+func NewLeafIDRecord(id *ID) tlv.Record {
+	const recordSize = sha256.Size
+	return tlv.MakeStaticRecord(LeafAssetID, id, recordSize, IDEncoder,
```
Roasbeef, line 58: Reason this changed was just to be able to use this new concrete type?

asset/records.go
```diff
@@ -143,6 +150,13 @@ func NewLeafScriptKeyRecord(scriptKey **btcec.PublicKey) tlv.Record {
 	)
 }
 
+func NewLeafXOnlyGroupKeyRecord(groupKey **GroupKey) tlv.Record {
```
Roasbeef, line 153: See above, not sure we'll need this.

fn/memory.go
```diff
@@ -36,3 +36,10 @@ func CopySlice[T any](slice []T) []T {
 	copy(newSlice, slice)
 	return newSlice
 }
+
+// CopyBytes returns a deep copy of the given byte slice.
+func CopyBytes(b []byte) []byte {
```
Roasbeef, line 41: Isn't htis the same as `CopySlice[[]byte]`?

Roasbeef: https://go.dev/play/p/_D5C9txX2TI

tapdb/sqlc/migrations/000002_assets.up.sql
```diff
@@ -172,6 +172,10 @@ CREATE TABLE IF NOT EXISTS assets (
     
     genesis_id INTEGER NOT NULL REFERENCES genesis_assets(gen_asset_id),
 
+    asset_computed_id BLOB UNIQUE,
```
Roasbeef, line 175: Same here, I don't think we need any of these fields. We'll already know the genesis information before we even are sent an asset. That information will always be there, as it must be known in order to properly verify an asset. 

In other words, we shouldn't really need to change this on disk structure. What we're changing is the wire level serialization: where these fields are encoded vs not. Ideally all the higher level logic stays the exact same (or needs minimal changes at least).

asset/asset.go
```diff
@@ -735,6 +740,12 @@ type Asset struct {
 	XOnlyGroupKey SchnorrKey
 }
 
+// IsUnknownVersion returns true if an asset has a version that is not
+// recognized by this implementation of tap.
+func (a *Asset) IsUnknownVersion() bool {
+	return a.Version != V0
```
Roasbeef, line 754: Morph into a switch statement now?

proof/proof.go
```diff
@@ -99,6 +99,16 @@ type Proof struct {
 	// NUMS key, to prove that the creator of the proof is able to produce
 	// a valid signature to spend the asset.
 	ChallengeWitness wire.TxWitness
+
+	// GenesisReveal is an optional Genesis struct, that must be provided
+	// for issuance proofs. This allows for derivation of the asset ID.
+	GenesisReveal *asset.Genesis
```
Roasbeef, line 105: 👍

proof/proof.go
```diff
@@ -99,6 +99,16 @@ type Proof struct {
...
+
+	// GroupKeyReveal is the set of bytes that represent the internal key
+	// and Tapscript root used to derive the group key for the asset group.
+	// The field is defined as internal_key || tapscript_root, and must be
+	// provided for issuance proofs of grouped assets.
+	GroupKeyReveal *asset.GroupKeyReveal
```
Roasbeef, line 111: Add additional context here (and also the `GroupKeyReveal` struct) on the full derivation? So the first tweak with the asset ID info, then the extra layer with the actual tapscript root.

Roasbeef (changes requested):

asset/asset.go
```diff
@@ -36,6 +45,15 @@ func (s SerializedKey) CopyBytes() []byte {
 	return c
 }
 
+// CopyBytes returns a copy of the underlying array as a byte slice.
+func (s SchnorrKey) CopyBytes() []byte {
+	return fn.CopyBytes(s[:])
+}
+
+func (s SchnorrKey) IsZero() bool {
+	return bytes.Equal(s[:], ZeroGroupKey[:])
```
guggero, line 60: Since they're both of the same type and an array, can just use `s == ZeroGroupKey` (which I imagine is faster as well).

asset/asset.go
```diff
@@ -658,6 +686,14 @@ type Asset struct {
 	// its unique ID within the Taproot Asset protocol.
 	Genesis
 
+	// Type uniquely identifies the type of Taproot asset.
+	ExplicitType Type
+
+	// ID is an asset's unique identifier within the Taproot Asset protocol.
```
guggero, line 705: nit: Godoc comment name doesn't match struct member name (here and for `ExplicitType`).

guggero (commented):
Did a quick first pass to grok the changes. From Laolu's comments it sounds like this might change quite a bit soon, so will do a second pass after those changes.

asset/asset.go
jharveyb: Not sure what you mean wrt. slim interface - I think using the `any` type would cause trouble elsewhere. Maybe we can do something with a custom type constraint though?

asset/asset.go
jharveyb: AH true, I forgot about that part of offline discussion.

I think the group key field would still change though, since the sig is being moved into the witness.

tapdb/sqlc/migrations/000002_assets.up.sql
jharveyb: I think we still need changes for the group key side though, to now store the tapscript root?

asset/asset.go
jharveyb: So maybe we actually want to modify the `AssetGroup` struct to include the tapscript root for the group key? For it to maintain the same definition of group key + info needed to derive it.

fn/memory.go
jharveyb: Yup, I misinterpreted the 'shallow copy' warning on `CopySlice`.

tapdb/sqlc/migrations/000002_assets.up.sql
jharveyb: Specifically a new field in the `asset_groups` table to store tapscript roots.

asset/asset.go
```diff
@@ -28,6 +28,11 @@ type SerializedKey [33]byte
 // 32-byte form.
 type SchnorrKey [32]byte
 
+// GroupKeyReveal is a type for representing the data used to derive a group key
+// for an asset. Specifically, internal_key || tapscript_root.
+// TODO(jhb): Implement constructor for this field that includes length checks
+type GroupKeyReveal [GroupKeyRevealLen]byte
```
halseth, line 34: Would it make more sense to make this a struct 
```go
type GroupKeyReveal {
    Key [32]byte
    Root [32]byte
}
```
that we could deserialize directly into?

asset/records.go
jharveyb: I think we do want it? If we're leaving in the current logic for group key struct encode/decode in this PR.

Changing it now to pull only the key from `GroupKey` though instead of using a new field in the `Asset` struct.

proof/proof.go
jharveyb: Added as godoc for the actual `GroupKeyReveal` struct.

jharveyb:
Removed the DB changes for assets, and made `GroupKeyReveal` a struct vs. flat byte array. Updated TLV numbers and reverted the changes to the asset struct.

IIUC the only DB change needed is a new field to store the actual group key tapscript root? @Roasbeef

asset/asset.go
Roasbeef: Mean moreso making `SchnorrSerialized` into an interface and using that elsewhere. Then either key can be passed in depending on context. I think we want to have this be explicit though, so we can ensure we're using the correct serialization when necessary.

asset/asset.go
Roasbeef: Yeah true, the key changes as now it's _just_ a key and not also the sig that comes along side it.

asset/asset.go
Roasbeef: Wouldn't that just live in the initial genesis proof? At that point you've already verified the derivation. 

I do agree that we need to store the group key derivation explicitly, so this way we can refer back to it (verify the inclusion proof for the tapscript leaf, etc) when it comes to further minting.

Roasbeef: > So maybe we actually want to modify the AssetGroup struct to include the tapscript root for the group key?

Or in something like `GroupDerivationInfo` or w/e which is read when we need to go verify subsequent minting events?

asset/records.go
Roasbeef: > I think we do want to change the encoding to only the x-only key though? Just not change the top-level struct.

x-only for the group key? Rationale being that it's now an actual taproot output key so it should itself be 32-bytes I take it?

tapdb/sqlc/migrations/000002_assets.up.sql
Roasbeef: > Specifically a new field in the asset_groups table to store tapscript roots.

Yep agreed this is necessary. I think the other question this leads into is: shoudl we break out the group key derivation info into another sub-table? Today we have the ability to crate and spend form script keys that have script paths, this logic is done primary outside the daemon though. Given that minting right now is something fully encapsulated in the daemon, do we also need to store the leaves as well for the group key tapscript tree, or would we rely again on the caller to handle that?

Roasbeef:
> IIUC the only DB change needed is a new field to store the actual group key tapscript root? @Roasbeef

Yep, I think so. Also see this comment on if we want to also store the script leaf information in the DB as well, if we assume the caller would do that: https://github.com/lightninglabs/taproot-assets/pull/381#discussion_r1279988255. If the latter, then we may need to expose some more utility RPCs to allow the caller to create a valid minting witness for their chosen part (so the give us the control block, etc).

asset/records.go
jharveyb: Yeah, right now we actually don't follow the BIP since we do serde for the group key as compressed 33-byte vs. Schnorr 32-byte.

https://github.com/lightninglabs/taproot-assets/blob/429b3ea2762ba5ef8a351d53454c672c2a54017b/asset/encoding.go#L557

Compared to the script key,  this key will always be 32-byte b/c it can only be derived with a TapTweak, so we could just do serde with 32 bytes instead of 33.

A separate question is whether we should store it in the DB as 32 or 33 bytes - it's 33 right now, IIRC that was for consistency with all the other keys we store?

jharveyb:
Following offline discussion, the current plan is:

 - Keep the group key as 33-byte, since `TapTweak` won't guarantee even parity for the final tweaked key. This also simplifies key handling in the DB overall.

(This requires a BIP update).

 - Adapt the asset DB tables to move away from storing the minting witness in the `asset_group_sigs` table, and instead store in the `asset_witness` table or a new table like `group_witnesses`. We should end up with no duplication of the minting witnesses, and queries to fetch these witnesses shouldn't be a large change.

asset/asset.go
```diff
@@ -459,6 +463,21 @@ type GroupKey struct {
 	Sig schnorr.Signature
 }
 
+// GroupKeyReveal is a type for representing the data used to derive the tweaked
+// key used to identify an asset group. The final tweaked key is the result of:
+// TapTweak(internalTweakedKey, tapscriptRoot)
+type GroupKeyReveal struct {
+	// InternalTweakedKey is the tweaked public key that is a result of:
+	// groupInternalKey + sha256(groupInternalKey || assetID) * G.
+	InternalTweakedKey SchnorrKey
```
halseth, line 472: This should be the non-tweaked key? (called `asset_raw_group_key`) in the BIP. Otherwise one cannot it has been tweaked with the `asset_id`

halseth (changes requested):

asset/asset.go
jharveyb: Ah true, I think I was mixing this up with `GroupKey.GroupPubKey`, which would now be the doubly-tweaked key.

tapdb/sqlc/migrations/000002_assets.up.sql
jharveyb: Discussed offline, resolved to just store the root for now.

jharveyb:
Ok, this has been updated to meet the discussed changes:

 - Tweaked group key is back to 33 bytes.
 - Better godoc comments explaining the tweaking process.
 - New DB table to store witnesses for asset group reissuance, that stores a tapscript root and an arbitrary witness stack.
 - Updates to other DB logic to maintain current functionality.
 
 AFAICT, those DB updates went well (unit tests on SQLite passing), but there is some mistake in the change to the asset group info view. The new `UpsertAssetGroupWitness` query should also be better tested.
 
 The error caused by the view update:
 
 ```bash
 CREATE VIEW key_group_info_view AS
SELECT
sig_id, gen_asset_id, tapscript_root, witness_stack, tweaked_group_key,
        raw_key, key_index, key_family
FROM asset_group_sigs sigs
JOIN asset_group_witnesses witnesses
        ON sigs.witness_id = witnesses.witness_id
JOIN asset_groups groups
        ON sigs.group_key_id = groups.group_id
JOIN internal_keys keys
        ON keys.key_id = groups.internal_key_id
WHERE sigs.gen_asset_id IN (SELECT gen_asset_id FROM genesis_info_view);
(details: ERROR: relation "asset_group_witnesses" does not exist (SQLSTATE 42P01))
Test:       	TestSelectCommitment/asset_with_key_group
 ```

asset/asset_test.go
```diff
@@ -198,18 +198,20 @@ func TestAssetEncoding(t *testing.T) {
...
-		Genesis: Genesis{
-			FirstPrevOut: wire.OutPoint{
-				Hash:  hashBytes1,
-				Index: 1,
-			},
-			Tag:         "asset",
-			MetaHash:    [MetaHashLen]byte{1, 2, 3},
-			OutputIndex: 1,
-			Type:        1,
+	splitGen := Genesis{
```
guggero, line 201: nit (commit message): what does "serde" mean? Sounds like a vim command :)

asset/records.go
```diff
@@ -49,8 +51,11 @@ func NewLeafVersionRecord(version *Version) tlv.Record {
 	)
 }
 
-func NewLeafIDRecord(id *[sha256.Size]byte) tlv.Record {
-	return tlv.MakePrimitiveRecord(LeafID, id)
+func NewLeafIDRecord(id *ID) tlv.Record {
+	const recordSize = sha256.Size
+	return tlv.MakeStaticRecord(LeafAssetID, id, recordSize, IDEncoder,
```
guggero, line 56: nit: formatting of multi-line function call.

proof/encoding.go
```diff
@@ -336,3 +338,43 @@ func MetaTypeDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
 	}
 	return tlv.NewTypeForEncodingErr(val, "MetaType")
 }
+
+func GroupKeyRevealEncoder(w io.Writer, val any, buf *[8]byte) error {
```
guggero, line 342: A unit test for these encoders/decoders would be nice.

tapdb/sqlc/migrations/000002_assets.up.sql
```diff
@@ -107,14 +107,24 @@ CREATE TABLE IF NOT EXISTS asset_groups (
...
     -- TODO(roasbeef): not needed since already in assets row?
     gen_asset_id INTEGER NOT NULL REFERENCES genesis_assets(gen_asset_id) UNIQUE,
 
     group_key_id INTEGER NOT NULL REFERENCES asset_groups(group_id)
 );
 
+CREATE TABLE IF NOT EXISTS asset_group_witnesses (
+    witness_id INTEGER PRIMARY KEY,
+
+    tapscript_root BLOB,
```
guggero, line 121: Do we hope that we can re-use some of these witnesses? Otherwise the alternative would be to just add both of these fields to the `asset_group_sigs` table and reduce the number of tables needed.

tapdb/sqlc/queries/assets.sql
```diff
@@ -155,13 +155,21 @@ RETURNING group_id;
...
-    genesis_sig, gen_asset_id, group_key_id
+    witness_id, gen_asset_id, group_key_id
 ) VALUES (
     $1, $2, $3
 ) ON CONFLICT (gen_asset_id)
     DO UPDATE SET gen_asset_id = EXCLUDED.gen_asset_id
 RETURNING sig_id;
 
+-- name: UpsertAssetGroupWitness :one
+INSERT INTO asset_group_witnesses (
```
guggero, line 166: This isn't really an upsert, as there is nothing to conflict on here. So it will always insert a new record...

asset/asset.go
```diff
@@ -488,6 +488,85 @@ func (g *GroupKey) IsEqualGroup(otherGroupKey *GroupKey) bool {
...
+// of a script witness.
+func IsGroupSig(witness wire.TxWitness) *schnorr.Signature {
+	var potentialSig []byte
+	if len(witness) == 1 {
+		potentialSig = witness[0]
+	}
+
+	// We can safely drop the signature parsing error, since a witness that
+	// isn't a signature may still be a valid script witness.
+	if len(potentialSig) == schnorr.SignatureSize {
```
guggero, line 502: nit: invert condition and continue the happy path non-indented?

asset/asset.go
```diff
@@ -488,6 +488,85 @@ func (g *GroupKey) IsEqualGroup(otherGroupKey *GroupKey) bool {
...
+	err := TxWitnessEncoder(&b, &witness, &buf)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"unable to serialize group witness: %w", err,
+		)
+	}
+
+	witnessStack = make([]byte, b.Len())
+	copy(witnessStack, b.Bytes())
+	return witnessStack, nil
```
guggero, line 549: I think `b.Bytes()` of a `bytes.Buffer` is safe to use without an additional copy.

asset/asset.go
```diff
@@ -488,6 +488,85 @@ func (g *GroupKey) IsEqualGroup(otherGroupKey *GroupKey) bool {
...
+
+	var (
+		buf          [8]byte
+		b            bytes.Buffer
+		witnessStack []byte
+	)
+
+	err := TxWitnessEncoder(&b, &witness, &buf)
+	if err != nil {
+		return nil, fmt.Errorf(
```
guggero, line 542: nit: formatting of `Errorf()` call can use the compact style. Here and below.

asset/asset.go
```diff
@@ -488,6 +488,85 @@ func (g *GroupKey) IsEqualGroup(otherGroupKey *GroupKey) bool {
...
+	}
+
+	// We can safely drop the signature parsing error, since a witness that
+	// isn't a signature may still be a valid script witness.
+	if len(potentialSig) == schnorr.SignatureSize {
+		groupSig, _ := schnorr.ParseSignature(potentialSig)
+		return groupSig
+	}
+
+	return nil
```
guggero, line 507: Maybe have a second boolean parameter instead of returning nil for an `Is...()` method?

guggero (commented):
Looks pretty good, nice work!

tapdb/sqlc/migrations/000002_assets.up.sql
jharveyb: I don't think so - I think actually the `tapscript_root` should go into `asset_groups` since it is static per group key, and then `witness_stack` could indeed be a field in `asset_group_sigs`.

One consideration is that the witnesses aren't really bounded size right now, and I suppose if they are in a separate table then we have more options in not fetching them for some queries? Not sure if that's a relevant concern.

asset/asset_test.go
guggero: aaaah, okay, got it. how about "ser/de-ser"? Still pretty short but IMO a bit easier to decipher.

halseth: I also thought it was some fancy unix command or something 😂

asset/asset.go
```diff
@@ -427,6 +431,23 @@ type GroupKey struct {
 	Sig schnorr.Signature
 }
 
+// GroupKeyReveal is a type for representing the data used to derive the tweaked
+// key used to identify an asset group. The final tweaked key is the result of:
+// TapTweak(internalTweakedKey, tapscriptRoot)
+type GroupKeyReveal struct {
+	// RawGroupKey is the public key that is tweaked twice to derive the
+	// final tweaked group key. The final tweaked key is the result of:
+	// InternalKey =  RawGroupKey * sha256(InternalKey || assetID) * G.
```
halseth, line 440: typo:  `InternalKey =  RawGroupKey * sha256(RawGroupKey || assetID) * G`

asset/asset.go
```diff
@@ -427,6 +431,23 @@ type GroupKey struct {
 	Sig schnorr.Signature
 }
 
+// GroupKeyReveal is a type for representing the data used to derive the tweaked
+// key used to identify an asset group. The final tweaked key is the result of:
+// TapTweak(internalTweakedKey, tapscriptRoot)
```
halseth, line 436: nit: inconsistent with naming below

proof/proof.go
```diff
@@ -99,6 +99,18 @@ type Proof struct {
 	// NUMS key, to prove that the creator of the proof is able to produce
 	// a valid signature to spend the asset.
 	ChallengeWitness wire.TxWitness
+
+	// GenesisReveal is the Genesis information for an asset, that must be
+	// provided for issuance proofs, and must be empty for non-issuance
+	// proofs. This allows for derivation of the asset ID and rederivation
+	// of an asset group key.
```
halseth, line 106: nit: maybe mention that not all assets will have an asset group key.

proof/file.go
```diff
@@ -190,11 +194,35 @@ func (f *File) Decode(r io.Reader) error {
 	return nil
 }
 
+// IsUnknownVersion returns true if a proof has a version that is not
+// recognized by this implementation of tap.
+func (f *File) IsUnknownVersion() bool {
```
halseth, line 199: same nit as earlier

tapdb/sqlc/migrations/000002_assets.up.sql
halseth: Agree that it would make sense to have the static `tapscript_root` with the asset group itself. No strong opinion on where the witness stack goes.

asset/asset.go
```diff
@@ -488,6 +488,85 @@ func (g *GroupKey) IsEqualGroup(otherGroupKey *GroupKey) bool {
...
+// IsGroupSig return true if the passed witness is a single Schnorr signature,
+// which is used to perform an asset group reissuance with a signature instead
+// of a script witness.
+func IsGroupSig(witness wire.TxWitness) *schnorr.Signature {
+	var potentialSig []byte
+	if len(witness) == 1 {
+		potentialSig = witness[0]
+	}
+
+	// We can safely drop the signature parsing error, since a witness that
```
halseth, line 500: This comment seems inconsistent with the method godoc?

halseth (changes requested):
Looks very good 😄

asset/asset.go
```diff
@@ -694,6 +699,17 @@ type Asset struct {
 	GroupKey *GroupKey
 }
 
+// IsUnknownVersion returns true if an asset has a version that is not
```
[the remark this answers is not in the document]
jharveyb: I think `IsUnknown` works better, since we use that as the condition to fail early where this function is called.

Otherwise we end up with something like `if !asset.IsKnownVersion() {` and inverting the result anyways.

asset/asset.go
jharveyb: Aligned names to those used in the godoc for the `GroupKey` struct.

jharveyb:
Ok, I renamed `asset_group_sigs.genesis_sig` to `witness_stack` and added a `tapscript_root` field to `asset_groups`. The length for `tapscript_root` is not enforced at the DB level, but `GroupKeyReveal.TapscriptRoot` is `[32]byte` and not `[]byte` so I think that's fine for now.

IMO we can defer other DB field renaming once we actually add support for non-sig witnesses in another PR.

I added test coverage for `UpsertAssetGroupKey` and the helpers that serialize and parse witnesses. I also updated the proof encode/decode to fix some pointer issues.

Right now we only insert an empty tapscript root, since we only support single signatures as witnesses. That functionality should be deferred to a follow-up PR.

proof/file.go
```diff
@@ -203,8 +231,8 @@ func (f *File) NumProofs() int {
 // ProofAt returns the proof at the given index. If the file is empty, this
 // returns nil.
 func (f *File) ProofAt(index uint32) (*Proof, error) {
-	if f.IsEmpty() {
-		return nil, ErrNoProofAvailable
+	if err := f.IsValid(); err != nil {
```
Roasbeef, line 234: I think we also want a version in the proof itself? https://github.com/lightninglabs/taproot-assets/blob/main/proof/proof.go#L43

So one top level is a version that describes how to parse and the serialization, etc. The other is a version for each proof that maybe tells how to validate, or which fields are expected to be there, etc.

tapdb/sqlc/migrations/000002_assets.up.sql
```diff
@@ -107,7 +109,9 @@ CREATE TABLE IF NOT EXISTS asset_groups (
 CREATE TABLE IF NOT EXISTS asset_group_sigs (
     sig_id INTEGER PRIMARY KEY,
 
-    genesis_sig BLOB NOT NULL, 
+    -- The witness stack can contain either a single Schnorr signature for key
+    -- spends of the tweaked group key, or a more complex script witness.
+    witness_stack BLOB NOT NULL,
```
Roasbeef, line 114: I think let's go ahead and rename now? Don't see much of a benefit of doing it later. Particularly since we'll merge this into a side branch that doesn't need to be fully working until we go to merge things back into master.

tapdb/sqlc/migrations/000002_assets.up.sql
```diff
@@ -92,6 +92,8 @@ CREATE TABLE IF NOT EXISTS asset_groups (
 
     tweaked_group_key BLOB UNIQUE NOT NULL CHECK(length(tweaked_group_key) = 33), 
 
+    tapscript_root BLOB,
```
Roasbeef, line 95: 👍

asset/asset.go
```diff
@@ -487,6 +487,79 @@ func (g *GroupKey) IsEqualGroup(otherGroupKey *GroupKey) bool {
 	return g.GroupPubKey.IsEqual(&otherGroupKey.GroupPubKey)
 }
 
+// IsGroupSig attempts to parse the given witness as a single Schnorr signature,
+// and returns the signature and true if it is valid.
+func IsGroupSig(witness wire.TxWitness) (*schnorr.Signature, bool) {
+	if len(witness) != 1 {
```
Roasbeef, line 493: These functions also need to mind the annex if we're going for 1:1 equivalence. There's some functions in `btcd/txscript` that should be useful for that.

tapdb/assets_common.go
```diff
@@ -259,8 +258,11 @@ func upsertGroupKey(ctx context.Context, groupKey *asset.GroupKey,
 		return nullID, fmt.Errorf("unable to insert internal key: %w",
 			err)
 	}
+	// Insert the group key with an empty tapscript root since we only
+	// support signatures as asset group key witnesses for now.
 	groupID, err := q.UpsertAssetGroupKey(ctx, AssetGroupKey{
 		TweakedGroupKey: tweakedKeyBytes,
+		TapscriptRoot:   []byte{},
```
Roasbeef, line 265: Perhaps add this onto the struct now so harder to forget to insert it down the line? May save some debugging saga.

tapdb/assets_common.go
```diff
@@ -494,7 +502,8 @@ func parseGroupKeyInfo(tweakedKey, rawKey, genesisSig []byte,
 		PubKey: untweakedKey,
 	}
 
-	groupSig, err := schnorr.ParseSignature(genesisSig)
+	// TODO(jhb): make generic for non-sig witnesses
+	groupSig, err := asset.ParseGroupSig(witness)
```
Roasbeef, line 506: We'd then modify `GroupKey` to include both the reveal information and raw witness, no? Then that way given that you can verify the minting if needed. May depend on the actual call site though.

Roasbeef (changes requested):
Super close. I think worth it to just do the rename now. We'll also want a version in the proof object itself. We should also handle the annex cases as rn we're still going for a 1:1 mapping.

asset/records.go
```diff
@@ -31,6 +26,13 @@ const (
 	LeafScriptVersion       LeafTlvType = 8
 	LeafScriptKey           LeafTlvType = 9
 	LeafGroupKey            LeafTlvType = 10
+
+	// Types for future asset format.
+	// TODO: LeafID and LeafGenesis have conflicting types. The BIP
```
halseth, line 31: this TODO is no longer relevant?

asset/asset.go
```diff
@@ -427,6 +430,23 @@ type GroupKey struct {
 	Sig schnorr.Signature
 }
 
+// GroupKeyReveal is a type for representing the data used to derive the tweaked
+// key used to identify an asset group. The final tweaked key is the result of:
+// TapTweak(groupInternalKey, tapscriptRoot)
+type GroupKeyReveal struct {
+	// RawGroupKey is the public key that is tweaked twice to derive the
+	// final tweaked group key. The final tweaked key is the result of:
+	// groupInternalKey =  RawKey * sha256(RawKey || assetID) * G.
```
halseth, line 439: BIP was updated to have the assetID be first in the sha256: https://github.com/Roasbeef/bips/pull/33#issuecomment-1669247706

tapdb/assets_common.go
halseth: When building on top of this PR I found at least the reveal information to be useful to be contained in the `GroupKey` struct.

asset/asset.go
guggero: Or see https://github.com/lightninglabs/pool/blob/master/poolscript/script.go#L466

tapdb/asset_minting.go
```diff
@@ -572,7 +572,8 @@ func fetchAssetSprouts(ctx context.Context, q PendingAssetStore,
 			if err != nil {
 				return nil, err
 			}
-			groupSig, err := schnorr.ParseSignature(sprout.GenesisSig)
+			// TODO(jhb): make generic for non-sig witnesses
+			groupSig, err := asset.ParseGroupSig(sprout.WitnessStack)
```
guggero, line 576: Maybe we should update the `asset.GroupKey` to contain the parsed `TxWitness` instead of the sig? Then we can move this TODO to where we validate the sig?

guggero (commented):
Getting very close indeed :tada:

tapdb/sqlc/migrations/000002_assets.up.sql
jharveyb: Yup fair enough.

asset/asset.go
jharveyb: Totally forgot about the annex, thanks for the catch - So IIUC for the current sig-only functionality, we should now allow any annex as long the first element is a sig?

tapdb/assets_common.go
jharveyb: True, I guess it should sort of track the group key view. Will make them optional though, not sure if we have a call site where these would not be populated yet like importing the tweaked group key before having the witness or so.

jharveyb:
Ok, the PR has been updated to try and replace the group sig field with a witness, and add the witness and Tapscript root to the GroupKey struct. The problem as-is is that the nil-ness behavior for the sig as `*schnorr.Signature` and a similar use with `wire.TxWitness` / `[][]byte` are different.

Specifically, both fields are `BLOB NOT NULL` in the DB. While the sig is just converted to `[]byte` with Serialize(), we use the TLV encode/decode for witnesses here, which IIUC means that an uninitialized witness and an initialized but empty witness will serialize differently.

Separately from the encoding of a nil vs. empty witness, currently the parsing of a signature from a witness is currently possibly too strict, since an empty witness may not always be an error case. For example, in some of the Universe subsystem, we may load or store a GroupKey with the sig unset, simply because it isn't relevant at that point and/or was validated earlier.

One option for adding more consistency for storing the witness, is to actually have the field as just `BLOB`, and then check for both a nil or empty witness when parsing a stored witness instead of failing in case of a nil witness.

tapdb/sqlc/migrations/000002_assets.up.sql
guggero: We assumed that with the rename you meant also renaming the table from `asset_group_sigs` to `asset_group_witnesses`? That's addressed in the latest push.

proof/file.go
guggero: Added that as a last commit.

guggero:
I took this PR over while @jharveyb is absent. Spoke to @Roasbeef and the intention is to merge this to the `v0-3-0-staging` branch while not 100% working. For example the unit and integration tests will only work again once we actually add the group key signature as the asset witness in the minting proof.

But I think the PR is in a state where it is mostly complete for a first part and we can work towards fixing everything on this staging branch.

guggero (approved):

proof/proof_test.go
```diff
@@ -505,6 +507,93 @@ func TestGenesisProofVerification(t *testing.T) {
...
+				ownershipProofHexFileName, []byte(expectedStr),
+				0644,
+			)
+			require.NoError(t, err)
+
+		// Any proof that starts with the comment "valid regtest proof
+		// file index X", we add to the proof-file.hex file.
+		case strings.Contains(comment, "valid regtest proof file index"):
+			fileProofs = append(fileProofs, *proof)
+		}
```
halseth, line 613: add default case for unexpected comment?

asset/records.go
```diff
@@ -15,12 +14,7 @@ import (
 type LeafTlvType = tlv.Type
 
 const (
-	LeafVersion LeafTlvType = 0
-	// TODO: LeafID and LeafGenesis have conflicting types. The BIP
-	// specifies LeafID, but this would require wallets to store the
-	// preimage to the asset ID elsewhere, when it could be included in the
-	// TLV itself as done with LeafGenesis.
-	LeafID                  LeafTlvType = 1
+	LeafVersion             LeafTlvType = 0
```
halseth, line 17: `genesis` is replaced with pure `assetID` in the BIP, but we can make that change in a follow-up PR.

asset/asset.go
```diff
@@ -422,6 +425,23 @@ type GroupKey struct {
 	Sig schnorr.Signature
 }
 
+// GroupKeyReveal is a type for representing the data used to derive the tweaked
+// key used to identify an asset group. The final tweaked key is the result of:
+// TapTweak(groupInternalKey, tapscriptRoot)
+type GroupKeyReveal struct {
+	// RawGroupKey is the public key that is tweaked twice to derive the
```
halseth, line 447: nit: godoc name mismatch

tapdb/assets_common.go
```diff
@@ -275,10 +276,10 @@ func upsertGroupKey(ctx context.Context, groupKey *asset.GroupKey,
 	// together otherwise disparate asset IDs).
 	//
 	// TODO(roasbeef): sig here doesn't actually matter?
-	groupSigID, err := q.UpsertAssetGroupSig(ctx, AssetGroupSig{
-		GenesisSig: groupKey.Sig.Serialize(),
-		GenAssetID: genAssetID,
-		GroupKeyID: groupID,
+	groupSigID, err := q.UpsertAssetGroupWitness(ctx, AssetGroupWitness{
```
halseth, line 288: nit: variable name still `sig`

asset/asset.go
```diff
@@ -426,8 +427,17 @@ type GroupKey struct {
 	//   groupInternalKey + sha256(groupInternalKey || genesisOutPoint) * G
 	GroupPubKey btcec.PublicKey
 
-	// Sig is a signature over an asset's ID by `Key`.
-	Sig schnorr.Signature
+	// TapscriptRoot is the root of the Tapscript tree that commits to all
+	// script spend conditions for the group key. Instead of spending an
+	// asset, these scripts are used to define witnesses more complex than
+	// a Schnorr signature for reissuing assets.
+	TapscriptRoot [sha256.Size]byte
```
halseth, line 434: Important to note that this can be empty (key-spend only output key). Should we make this nillable, or is it good enough to have it be all zero?

halseth (approved):
Mostly just nits, will be good to have this merged in order to build on top 👍

proof/proof_test.go
guggero: There are more than those three test cases in the file, we just export some of them into a different format to test. So we already ignore a bunch of them. Will make it more apparent by adding an empty default case with a comment.

asset/asset.go
guggero: Good point. Turned it into a `[]byte` with a comment. Can do some additional validation later on.

proof/records.go
```diff
@@ -286,3 +295,36 @@ func MetaRevealTypeRecord(metaType *MetaType) tlv.Record {
...
+	return tlv.MakeDynamicRecord(
+		GenesisRevealType, genesis, recordSize, GenesisRevealEncoder,
+		GenesisRevealDecoder,
+	)
+}
+
+func GroupKeyRevealRecord(reveal **asset.GroupKeyReveal) tlv.Record {
+	recordSize := func() uint64 {
+		if reveal == nil || *reveal == nil {
+			panic("nil reveal")
```
halseth, line 319: should we really panic here?

guggero:
Going to merge this into the `v0-3-0-staging` branch now so we can continue building on top. Can always fix remaining issues later on that branch.

PR #385, opened 2023-07-05
Remove lnd references in user-visible issue copy

dstadulis:
When clicking "New Issue", the copy in the label referenced lnd which is confusing given expectations are to create a taproot-assets issue


ffranr (approved):
Not sure why I glazed over that in the previous PR!

dstadulis:
> Not sure why I glazed over that in the previous PR!

wasn't clear (to me at least) this copy would have been shown to users

Roasbeef (approved):
LGTM 🧿

Issue #387, opened 2023-07-06
[enhancement] Add block height / `GenesisHeight` to asset proofs


dstadulis:
### Background

To produce information related to assets known to universes, services will want to query by date.  Adding this information will enable reduced resource consumption and discrete information to be selected / queried.

See #367 and comment https://github.com/lightninglabs/taproot-assets/issues/367#issuecomment-1611914226 for more details

The branch https://github.com/lightninglabs/taproot-assets/tree/asset-stats-improvements has added `GenesisHeight` to universe structs but this block height needs to be stored in the asset proof file to be incorporated.

i.e.
https://github.com/lightninglabs/taproot-assets/blob/cafe59a004068639b3898c9691a4c07ac1784645/tapdb/sqlc/universe.sql.go#L446-L454

### Deliverable

1 Modify the asset proof information/file to store height
2 (IIUC these checks would be necessary) add validation logic to reject proofs with mismatching block heights

Issue #367, opened 2023-06-16
Additional improvements for the stats RPCs


jamaljsr:
There are currently two universe RPC methods ([QueryAssetStats](https://lightning.engineering/api-docs/api/taproot-assets/universe/query-asset-stats) and [UniverseStats](https://lightning.engineering/api-docs/api/taproot-assets/universe/universe-stats)) which return stats about the assets the node is aware of. As I am working on exposing this information in Terminal, I've discovered some improvements that could be made to improve the dev/user experience.

1. These two endpoints currently require a macaroon to access them. We want to be able to query the LL universe server without requiring the user to authenticate. There are two options that I can think of to solve this, but I'm open to other suggestions.
    1. add a config flag which allows anyone to access these specific RPC methods without needing to supply a macaroon
    2. bake a custom macaroon with limited access to these RPCs only and hard-code that into the Terminal app
2. `QueryAssetStats` can currently sort by asset id, name, and type. We also would like to be able to sort by `total_syncs`, `total_proofs`, `genesis_height`
3. In `QueryAssetStats`, also include the fields `genesis_point` and `group_key` for each asset.
4. In `QueryAssetStats`, each asset has a `genesis_height` field. Would it be possible to also include the timestamp of this block so that the web app doesn't need to make N more API requests to convert the block height to a timestamp? We'd like to display the date/time of when the asset was created.
5. We'd like to display a chart showing the number of total assets over time in daily increments. Given a start and end date, the RPC would return an array of objects containing the fields: `timestamp`, `total_assets`, `total_syncs`, `total_proofs`. There would be one entry for each day between the supplied dates.

Roasbeef:
For #5, here's the queries we had for sqlite + postgres, we need to add a slim abstraction so we an define an interfaces and have an implementation for sqlite vs postges:

sqlite
```
SELECT
    strftime('%Y-%m-%d %H:%M:00', event_time, '-5 minutes') AS interval_start,
    SUM(CASE WHEN event_type = 'SYNC' THEN 1 ELSE 0 END) AS total_sync_events,
    SUM(CASE WHEN event_type = 'NEW_PROOF' THEN 1 ELSE 0 END) AS total_new_proof_events
FROM
    universe_events
WHERE
    event_type IN ('SYNC', 'NEW_PROOF') AND -- Specify the event types you're interested in
    event_time >= @start_time AND event_time <= @end_time -- Filter events based on the time window
GROUP BY
[...]

PR #389, opened 2023-07-10
Add block height to asset mint proof

ffranr:
Addresses this issue: https://github.com/lightninglabs/taproot-assets/issues/387


proof/records.go
```diff
@@ -55,6 +56,12 @@ func BlockHeaderRecord(header *wire.BlockHeader) tlv.Record {
 	)
 }
 
+func BlockHeightRecord(height *uint32) tlv.Record {
```
guggero, line 59: This commit needs to be earlier, otherwise the previous one doesn't compile.

guggero: Or just squash this with the previous one, as there are some changes to the same line in here anyway.

taprpc/taprootassets.proto
```diff
@@ -199,6 +199,9 @@ message AnchorInfo {
     anchor output.
     */
     bytes tapscript_sibling = 7;
+
+    // The height of the block which contains the anchor transaction.
+    uint32 block_height = 8;
```
guggero, line 204: We usually generate the protos in the same commit as we add/update fields.

proof/proof.go
```diff
@@ -152,6 +151,7 @@ func (p *Proof) DecodeRecords() []tlv.Record {
 		MetaRevealRecord(&p.MetaReveal),
 		AdditionalInputsRecord(&p.AdditionalInputs),
 		ChallengeWitnessRecord(&p.ChallengeWitness),
+		BlockHeightRecord(&p.BlockHeight),
```
guggero, line 154: Should we add this field to one of the unit tests? I would assume that would've caught the "TLV stream must be canonical" error that caused you to move this field to the bottom.

guggero (commented):
Looks good, just needs a small addition to a unit test.

taprpc/taprootassets.proto
jharveyb: Also applies to the sqlc generate step as well IIRC.

proof/proof.go
ffranr: In the latest commits, I've added the block height field into a proof encode/decode unit test.

jharveyb:
Do we also want to return the confirmed block height when querying assets outside of the `genesis_info_view`?

One way would be to modify `QueryAssets` to also return the block height:

https://github.com/lightninglabs/taproot-assets/blob/22932e01e366334b9efb016a6c39c3edfe909ccb/tapdb/sqlc/queries/assets.sql#L372

And then update `dbAssetsToChainAssets` to populate the corresponding field in returned assets:

https://github.com/lightninglabs/taproot-assets/blob/22932e01e366334b9efb016a6c39c3edfe909ccb/tapdb/assets_store.go#L706

I don't think this actually blocks #367, but for testing it would be useful to have a query that accesses the `block_height` field added to `genesis_info_view`.

ffranr:
> Do we also want to return the confirmed block height when querying assets outside of the genesis_info_view?

That sounds like a good idea @jharveyb . I'd like to merge this PR as is and make asset related changes in a different PR.
I think I'll make a different PR for the block height verification also.

guggero (approved):
LGTM :tada:

jharveyb (approved):
LGTM 👍🏽

PR #393, opened 2023-07-10
QueryAssetStats and UniverseStats no auth client access

ffranr:
This PR adds the `QueryAssetStats` and `UniverseStats` endpoints to the macaroon whitelist so that they can be accessed without authentication.


jharveyb (approved):
LGTM 👍🏽

PR #396, opened 2023-07-12
Improve universe stats

### LINKED ISSUE #367: Additional improvements for the stats RPCs

guggero:
~Depends on #389.~

Fixes #367.

This PR addresses items 2, 3, 4 and 5 of #367.

@jamaljsr can you please check if this is what you need?


rpcserver.go
```diff
@@ -3107,10 +3137,93 @@ func (r *rpcServer) QueryAssetStats(ctx context.Context,
...
+	if err == nil {
+		return int64(cacheTS)
+	}
+
+	hash, err := r.cfg.Lnd.ChainKit.GetBlockHash(ctx, int64(height))
+	if err != nil {
+		return 0
+	}
+
+	block, err := r.cfg.Lnd.ChainKit.GetBlock(ctx, hash)
```
jharveyb, line 3175: Non-blocking: I saw that this wasn't exposed via `ChainKit` but it seems like here (and elsewhere), we would want to use something like `GetBlockHeader()` and not fetch a whole block?

Or is that not exposed because some chain backends don't support that?

tapdb/universe_stats_test.go
```diff
@@ -127,6 +131,10 @@ func TestUniverseStatsEvents(t *testing.T) {
 	// asset above. We'll mark these each first as a new proof.
 	for i := 0; i < numAssets; i++ {
 		sh.logProofEventByIndex(i)
+
+		// Increment the clock by a full day to ensure that the event
+		// is grouped into its own day.
+		testClock.SetTime(testClock.Now().Add(24 * time.Hour))
```
jharveyb, line 137: Running this locally, I had a weird issue where the first iteration of this loop would modify both `testClock` and `yesterday`, causing the test to fail (reproducibly). Not sure why this is happening or why it isn't observed in CI.

jharveyb (approved):
Looks good! Cool to have a migration in + DB-specific logic support.

rpcserver.go
guggero: We do have a `GetBlockHeader()` method for every chain backend. The reason we can't use it here is because there currently is no RPC in `lnd` that exposes the call.
I agree that we'll want to add this as an optimization. I'll see if I can get a PR in for 0.17.

tapdb/universe_stats_test.go
guggero: Very weird, I couldn't reproduce that. What Golang version were you experiencing this with? Also, were you running the test directly, using a debugger or through `make unit`?

tapdb/universe.go
```diff
@@ -300,8 +300,6 @@ func upsertAssetGen(ctx context.Context, db UpsertAssetStore,
 		return 0, fmt.Errorf("unable to anchor genesis tx: %w", err)
 	}
 
-	// TODO(roasbeef): need to mark that this is a imported gen?
```
ffranr: I read this TODO as: now that there is no difference between an imported and local genesis record, should we mark the imported genesis as having been imported?

And in that sense I think the TODO should stay.

go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	github.com/lightninglabs/neutrino/cache v1.1.1
 	github.com/lightningnetwork/lnd v0.16.0-beta.rc1
 	github.com/lightningnetwork/lnd/cert v1.2.1
+	github.com/lightningnetwork/lnd/clock v1.1.0
```
ffranr, line 29: I'm not sure why we're adding `lnd/clock`. What is the goal of this commit?

tapdb/sqlc/queries/universe.sql
```diff
@@ -116,15 +116,15 @@ WITH group_key_root_id AS (
     LIMIT 1
 )
 INSERT INTO universe_events (
-    event_type, universe_root_id, event_time
+    event_type, universe_root_id, event_time, event_timestamp
```
ffranr, line 119: What's the difference between `event_time` and `event_timestamp`?

tapdb/universe_stats_test.go
jharveyb: Have Go `1.20.5` here, ran into this with the VSCode test runner in both normal and debug modes.

But not when running `make unit` directly - I think we can ignore it.

tapdb/universe_stats.go
```diff
@@ -53,6 +68,16 @@ type UniverseStatsStore interface {
...
+
+	// QueryAssetStatsPerDaySqlite returns the stats for a given asset
+	// grouped by day in a SQLite specific format.
+	QueryAssetStatsPerDaySqlite(ctx context.Context,
+		q AssetStatsPerDayQuery) ([]AssetStatsPerDay, error)
+
+	// QueryAssetStatsPerDayPostgres returns the stats for a given asset
+	// grouped by day in a Postgres specific format.
+	QueryAssetStatsPerDayPostgres(ctx context.Context,
+		q AssetStatsPerDayQueryPg) ([]AssetStatsPerDayPg, error)
```
ffranr, line 80: Why do we have backend specific query functions here? I thought that wasn't necessary with a unix timestamp.

tapdb/universe_stats.go
```diff
@@ -112,9 +137,10 @@ func (u *UniverseStats) LogSyncEvent(ctx context.Context,
 		}
 
 		return db.InsertNewSyncEvent(ctx, NewSyncEvent{
-			EventTime:     u.clock.Now(),
-			AssetID:       uniID.AssetID[:],
-			GroupKeyXOnly: groupKeyXOnly,
+			EventTime:      u.clock.Now(),
+			EventTimestamp: u.clock.Now().UTC().Unix(),
```
ffranr, line 141: `EventTime` seems redundant now, is it?

tapdb/universe_stats.go
```diff
@@ -53,6 +68,16 @@ type UniverseStatsStore interface {
 	// universe/
 	QueryUniverseAssetStats(ctx context.Context,
 		arg UniverseStatsQuery) ([]UniverseStatsResp, error)
+
+	// QueryAssetStatsPerDaySqlite returns the stats for a given asset
+	// grouped by day in a SQLite specific format.
+	QueryAssetStatsPerDaySqlite(ctx context.Context,
+		q AssetStatsPerDayQuery) ([]AssetStatsPerDay, error)
```
ffranr, line 75: Perhaps `AssetStatsPerDay` should be renamed to be sqlite specific to match the pattern for `AssetStatsPerDayPg`.

ffranr (commented):
There's a lot of work here, good job Oli on getting it together!

I would question whether we need to aggregate the events data at all. Or at least whether it makes sense to aggregate by day. That seems like the sort of thing we could just ask a client to do for the time being.

If we are to aggregate, I wonder if we could just keep the aggregation window as an argument.

I'm guessing the terminal team requested this level of aggregation.

rpcserver.go
```diff
@@ -3107,10 +3137,48 @@ func (r *rpcServer) QueryAssetStats(ctx context.Context,
...
+
+// getBlockTimestamp returns the timestamp of the block at the given height.
+func (r *rpcServer) getBlockTimestamp(ctx context.Context,
+	height uint32) int64 {
+
+	// Shortcut any lookup in case we don't have a valid height in the first
+	// place.
+	if height == 0 {
+		return 0
+	}
```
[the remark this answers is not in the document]
guggero: This is to remain backward compatible, as we'll be running this patch in our testnet instance that will have proofs without a timestamp. So I think erroring out here is a bad idea, as a single invalid proof would break the whole stats RPC.

go.mod
guggero: To be able to test sync/proof events that are inserted into the DB with the current timestamp. If we used `time.Now()`, those would only be a few milliseconds apart. But we want to simulate events from multiple days, so we use a clock that we can overwrite. Added this info to the commit message.

tapdb/sqlc/queries/universe.sql
guggero: Just the format. One is the Golang string representation of a `time.Time` while the other is a Unix timestamp in seconds. There's a comment in `tapdb/sql/migrations/000008_universe_events_timestamp.up.sql` that explains why we have the duplicate field (mainly to avoid this being a breaking DB change).

tapdb/universe_stats.go
guggero: See comment above about non-breaking. We can clean this up for 0.3 when we're going to break things anyway.

tapdb/universe_stats.go
guggero: The available time calculation functions are different between the backends (see the two different queries).
We know that we'll run into such differences here and in `lnd` in the future anyway, so this kind of serves as a proof of concept for backend specific queries.

tapdb/universe_stats.go
guggero: The two query and response structs are identical, but `sqlc` creates unique ones anyway. And since in Golang we can't just convert one into the other easily (AFAIK), we need to use both. I chose one of them as the return type for the exported method (the API surface of the `tapdb` package), so it would be confusing if that had "Sqlite" in its name.
The alternative would be to have a third alias type, but that seemed overly excessive.

tapdb/universe.go
jharveyb: TBF we mark script keys as local vs. imported, so we don't lose the ability to distinguish asset locality.

tapdb/sqlc/queries/universe.sql
ffranr: Shouldn't hold back this PR 👍

ffranr (approved):
🥳

Issue #390, opened 2023-07-10
[enhancement]: Add universe-server configuration option to enable client access without authentication


dstadulis:
### Background
---
#367 requests to serve Universe stats with minimum prerequisites. E.g. Enable a universe server to respond to queries without requiring the client / user authenticate to the universe server:

>add a config flag which allows anyone to access [([QueryAssetStats](https://lightning.engineering/api-docs/api/taproot-assets/universe/query-asset-stats) and [UniverseStats](https://lightning.engineering/api-docs/api/taproot-assets/universe/universe-stats))] RPC methods without needing to supply a macaroon

As a `universe operator` I would like to `respond to requests without requiring authentication from querying clients` in order to `make access to my service easily available without the need to identification / user friction`

As a `universe operator` I would like to `respond to requests without requiring authentication from querying clients` in order to `not require Terminal users to be logged in to query Universe stats`

### Deliverables
---
1. Add universe config flag to reduce the server's required client authentication for ([QueryAssetStats](https://lightning.engineering/api-docs/api/taproot-assets/universe/query-asset-stats) and [UniverseStats](https://lightning.engineering/api-docs/api/taproot-assets/universe/universe-stats)) gRPC calls
Relevant additions to list of RPCs will be made here:
https://github.com/lightninglabs/taproot-assets/blob/7a5d5cae25f980c0ead87c0578584eafee689624/rpcperms/interceptor.go#L98-L105
which was delivered in commit https://github.com/lightninglabs/taproot-assets/commit/1660aa7ce97e34fe4ade8c1d932e2521386a1da4

1. Consider a holistic design / more comprehensive set, of the configuration flags if other, unspecified-in-this-issue, calls or services might also benefit from not needing authentication.

### Alternative solutions
---
> bake a custom macaroon with limited access to these RPCs only and hard-code that into the Terminal app

Advantages:
- Universe server could deny simple/naive resource exhaustion / DoS attempts

Disadvantages:
- Changes necessary to Terminal
- maintaining the hard-coded macaroons
- less private asset-retrieval requests

ffranr:
I feel like the right solution here is to add a `macaroon-whitelist` command line argument.
And then the user could just add:
```
tapd ... --macaroon-whitelist=QueryAssetStats --macaroon-whitelist=UniverseStats
```

Do we have precedence for this in LND? @guggero ?

guggero:
[...]

PR #398, opened 2023-07-13
Add CLI flag to skip authentication for stats related RPC endpoints

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/390

I've tested that the CLI flag reaches the correct config, but I haven't tested further.


cmd/tapcli/main.go
```diff
@@ -310,6 +310,11 @@ func main() {
 			Name:  "macaroonip",
 			Usage: "If set, lock macaroon to specific IP address.",
 		},
+		cli.BoolFlag{
```
guggero, line 313: This is not needed, as this is a server-side config only.

guggero (approved):
Nice! Just one nit, otherwise LGTM :tada:

PR #400, opened 2023-07-13
Add proof block height verification

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/387

### DIFF
--- a/chain_bridge.go
+++ b/chain_bridge.go
@@ -81,6 +81,39 @@ func (l *LndRpcChainBridge) GetBlockHash(ctx context.Context,
 	return blockHash, nil
 }
 
+// VerifyBlock returns an error if a block (with given header and height) is not
+// present on-chain. It also checks to ensure that block height corresponds to
+// the given block header.
+func (l *LndRpcChainBridge) VerifyBlock(ctx context.Context,
+	header wire.BlockHeader, height uint32) error {
+
+	// TODO(ffranr): Once we've released 0.3.0, every proof should have an
+	// assigned height. At that point, we should return an error for proofs
+	// with unset (zero) block heights.
+	if height == 0 {
+		_, err := l.GetBlock(ctx, header.BlockHash())
+		return err
+	}
+
+	// Ensure that the block hash matches the hash of the block
+	// found at the given height.
+	hash, err := l.GetBlockHash(ctx, int64(height))
+	if err != nil {
+		return err
+	}
+
+	expectedHash := header.BlockHash()
+	if hash != expectedHash {
+		return fmt.Errorf("block hash and block height "+
+			"mismatch; (height: %x, hashAtHeight: %s, "+
+			"expectedHash: %s)", height, hash, expectedHash)
+	}
+
+	// Ensure that the block header corresponds to a block on-chain.
+	_, err = l.GetBlock(ctx, header.BlockHash())
+	return err
+}
+
 // CurrentHeight return the current height of the main chain.
 func (l *LndRpcChainBridge) CurrentHeight(ctx context.Context) (uint32, error) {
 	info, err := l.lnd.Client.GetInfo(ctx)
--- a/chain_bridge.go
+++ b/chain_bridge.go
@@ -67,6 +67,20 @@ func (l *LndRpcChainBridge) GetBlock(ctx context.Context,
 	return block, nil
 }
 
+// GetBlockHash returns the hash of the block in the best blockchain at the
+// given height.
+func (l *LndRpcChainBridge) GetBlockHash(ctx context.Context,
+	blockHeight int64) (chainhash.Hash, error) {
+
+	blockHash, err := l.lnd.ChainKit.GetBlockHash(ctx, blockHeight)
+	if err != nil {
+		return chainhash.Hash{}, fmt.Errorf("unable to retrieve "+
+			"block hash: %w", err)
+	}
+
+	return blockHash, nil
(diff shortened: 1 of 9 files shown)


chain_bridge.go
```diff
@@ -81,6 +81,38 @@ func (l *LndRpcChainBridge) GetBlockHash(ctx context.Context,
 	return blockHash, nil
 }
 
+// VerifyBlock returns an error if a block (with given header and height) is not
+// present on-chain. It also checks to ensure that block height corresponds to
+// the given block header.
+func (l *LndRpcChainBridge) VerifyBlock(ctx context.Context,
+	header wire.BlockHeader, height uint32) error {
+
+	_, err := l.GetBlock(
```
guggero, line 90: I think we should do this last, only once we know we're expecting the block to be in the current chain (since this is the most "expensive" call in terms of resources).

chain_bridge.go
```diff
@@ -81,6 +81,38 @@ func (l *LndRpcChainBridge) GetBlockHash(ctx context.Context,
 	return blockHash, nil
 }
 
+// VerifyBlock returns an error if a block (with given header and height) is not
+// present on-chain. It also checks to ensure that block height corresponds to
+// the given block header.
+func (l *LndRpcChainBridge) VerifyBlock(ctx context.Context,
```
guggero, line 87: We should error out early if the block height is 0.

itest/assertions.go
```diff
@@ -265,14 +265,42 @@ func verifyProofBlob(t *testing.T, tapd *tapdHarness,
...
-	headerVerifier := func(blockHeader wire.BlockHeader) error {
-		hash := blockHeader.BlockHash()
+	headerVerifier := func(header wire.BlockHeader, height uint32) error {
+		hash := header.BlockHash()
 		req := &chainrpc.GetBlockRequest{
 			BlockHash: hash.CloneBytes(),
 		}
 		_, err := tapd.cfg.LndNode.RPC.ChainKit.GetBlock(ctxb, req)
-		return err
+		if err != nil {
```
guggero, line 274: Same here, probably makes sense to only do this after we've confirmed the block should be in the current chain.

tapgarden/caretaker.go
```diff
@@ -1025,11 +1025,11 @@ func GetTxFee(pkt *psbt.Packet) (int64, error) {
...
 func GenHeaderVerifier(ctx context.Context,
-	chainBridge ChainBridge) func(header wire.BlockHeader) error {
+	chainBridge ChainBridge) func(wire.BlockHeader, uint32) error {
 
-	return func(blockHeader wire.BlockHeader) error {
-		_, err := chainBridge.GetBlock(
-			ctx, blockHeader.BlockHash(),
+	return func(header wire.BlockHeader, height uint32) error {
+		err := chainBridge.VerifyBlock(
+			ctx, header, height,
```
guggero, line 1032: nit: could now fit on a single line with the method call?

proof/proof_test.go
```diff
@@ -315,16 +316,20 @@ func genRandomGenesisWithProof(t testing.TB, assetType asset.Type,
 		[]*btcutil.Tx{btcutil.NewTx(genesisTx)}, false,
 	)
 	merkleRoot := merkleTree[len(merkleTree)-1]
+	// We'll use the genesis hash of the mainnet chain as the parent block.
```
guggero, line 319: nit: add a newline before a comment (here and below).

guggero (commented):
Looks good! Just one comment about the order of RPC calls, should be a small thing to fix.

jharveyb (commented):
In agreement with previous comments - seems like we should have a negative test in the itests? Or as a unit test, not sure how easy it is to reorg under itest.

chain_bridge.go
```diff
@@ -81,6 +81,38 @@ func (l *LndRpcChainBridge) GetBlockHash(ctx context.Context,
...
+	header wire.BlockHeader, height uint32) error {
+
+	// TODO(ffranr): Once we've released 0.3.0, every proof should have an
+	// assigned height. At that point, we can remove this check.
```
guggero, line 91: nit: "At that point we can return an actual error here".

proof/proof_test.go
```diff
@@ -511,6 +521,18 @@ func TestProofBlockHeaderVerification(t *testing.T) {
...
+	proof.BlockHeader.Nonce = originalBlockHeader.Nonce
+
+	// Modify proof block height, then check that the verification function
+	// propagates the correct error.
+	proof.BlockHeight += 1
+	_, actualErr = proof.Verify(
+		context.Background(), nil, headerVerifier,
+	)
+	actualErrInner = errors.Unwrap(actualErr)
+	require.Equal(t, actualErrInner, errHeaderVerifier)
```
guggero, line 535: nit: can just use `require.ErrorIs()` (I think that's what it's called).

guggero (approved):
Just two more nits, otherwise LGTM :tada:

ffranr:
> In agreement with previous comments - seems like we should have a negative test in the itests? Or as a unit test, not sure how easy it is to reorg under itest.

@jharveyb I've updated this PR with an additional negative unit test orientated around block height. Oli mentioned that the re-org work is planned for 0.3.0. Perhaps after that work is complete we can revisit itests for proof block fields?

jharveyb (approved):
Forgot that the reorg work will happen later and separately - LGTM!

PR #416, opened 2023-07-25
Multiverse fetch/register issuance

ffranr:
This PR contains new functionality for registering an asset issuance in a multiverse MS-SMT.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416)


tapdb/universe_forest.go
```diff
@@ -125,3 +130,69 @@ func (b *BaseUniverseForest) RootNodes(
 
 	return uniRoots, nil
 }
+
+// RegisterIssuance inserts a new minting leaf within the universe tree, stored
+// at the base key.
+func (b *BaseUniverseForest) RegisterIssuance(ctx context.Context,
```
Roasbeef, line 136: Can we also add fetching here? So then we'll be able to add that second level of proof and also verification.

universe/interface.go
```diff
@@ -41,14 +41,20 @@ type Identifier struct {
 	GroupKey *btcec.PublicKey
 }
 
-// String returns a string representation of the ID.
-func (i *Identifier) String() string {
+// Bytes returns a bytes representation of the ID.
+func (i *Identifier) Bytes() [32]byte {
```
Roasbeef, line 45: If we add this, then it should also be used in the db when we got to map the identifier into a set of bytes.

tapdb/universe_forest.go
```diff
@@ -125,3 +130,69 @@ func (b *BaseUniverseForest) RootNodes(
...
+			leafNode := mssmt.NewLeafNode(
+				universeRootHash[:], assetGroupSum,
+			)
+
+			// Use asset ID (or asset group hash) as the upper tree
+			// leaf node key. This is the same as the asset specific
+			// universe ID.
+			leafNodeKey := id.Bytes()
+
+			_, err = multiverseTree.Insert(
```
Roasbeef, line 209: The issuance proof returned here should now also include the inclusion path into the multi-verse root.

universe/interface.go
```diff
@@ -222,6 +222,12 @@ type BaseForest interface {
 	// of assets tracked in the base Universe.
 	RootNodes(ctx context.Context) ([]BaseRoot, error)
 
+	// RegisterIssuance inserts a new minting leaf within the universe tree,
+	// stored at the base key.
+	RegisterIssuance(ctx context.Context, id Identifier, key BaseKey,
```
Roasbeef, line 227: I think we want to modify the naming a bit here, or just flatten it a bit. At this level of abstraction there isn't a distinction between a multiverse that contains all assets, one that contains all proofs for a given asset, and one that contains only issuance proofs for one or all assets. This is referring to the `BaseForest` interface name here, that was meant to signify the issuance-only multi-verse tree.

tapdb/universe_forest.go
```diff
@@ -131,6 +131,33 @@ func (b *BaseUniverseForest) RootNodes(
...
+	universeKey universe.BaseKey) ([]*universe.IssuanceProof, error) {
+
+	var (
+		readTx = NewBaseUniverseReadTx()
+		proofs []*universe.IssuanceProof
+	)
+
+	dbErr := b.db.ExecTx(ctx, &readTx, func(dbTx BaseUniverseForestStore) error {
+		var err error
+		proofs, err = universeFetchIssuanceProof(
```
Roasbeef, line 149: See other comment, this also needs to have a multiverse root inclusion proof along side it.

Roasbeef (changes requested):

universe/interface.go
ffranr: I don't think we do that for the identifier. I can see instances where we make use of the identifier's fields. But otherwise I think we convert the identifier into a `namespace` which is the same as `Identifier.String()`. So I've added a commit to make that clear.

universe/interface.go
ffranr: I've renamed `BaseForest` to `Forest`.

universe/interface.go
Roasbeef: Just to clarify, I was referring to this: https://github.com/lightninglabs/taproot-assets/blob/main/tapdb/universe.go#L138-L146

Roasbeef (changes requested):
> Reviewed 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416)* status: 2 of 5 files reviewed, 9 unresolved discussions (waiting on @ffranr)

___
*[`tapdb/universe.go` line 526 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiTzPG1tmIu42UT62-:-NaiTzPG1tmIu42UT620:b-6qdm25) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/tapdb/universe.go#L526)):*
> ```Go
> 
> 	if len(universeLeaves) == 0 {
> 		return nil, nil
> ```

Why `nil` error here now?
___
*[`tapdb/universe_forest.go` line 181 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiU6a481P3VFsYsFNj:-NaiU6a481P3VFsYsFNk:bx8rqgj) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/tapdb/universe_forest.go#L181)):*
> ```Go
> 		}
> 
> 		for i := range proofs {
> ```

:+1: 

I think we can just name as Multiverse root here? Forest was sort of some tie over naming, unless we like that and want to elevate it more. 
___
*[`universe/base.go` line 143 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiUU2M8JHY8snqyHkF:-NaiUU2M8JHY8snqyHkG:b-3r21j0) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/base.go#L143)):*
> ```Go
> 	// universe forest. If so, then we'll return the existing issuance
> 	// proof.
> 	issuanceProofs, err := a.cfg.UniverseForest.FetchIssuanceProof(
> ```

Not so sure we want to do away with the base universe distinction here. I think there's a greater question of: will what's in `base.go` allow you to get proofs of any arbitrary state transition of an asset, or just the minting proofs? 
___
*[`universe/base.go` line 150 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiUkgK9wR9Yr_4rMHf:-NaiUkgK9wR9Yr_4rMHg:bbaw0jp) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/base.go#L150)):*
> ```Go
> 	}
> 
> 	if len(issuanceProofs) > 0 {
> ```

Ah so opted to return nil for this? IMO it's better to make that error condition explicit.

Roasbeef (changes requested):
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416)* status: 2 of 5 files reviewed, 11 unresolved discussions (waiting on @ffranr)

___
*[`universe/interface.go` line 152 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiV61rCj3Im2nyDh1B:-NaiV61rCj3Im2nyDh1C:b-rgaswd) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/interface.go#L152)):*
> ```Go
> 	// ForestRoot is the root of the universe forest that the asset is
> 	// located within.
> 	ForestRoot mssmt.Node
> ```

See comment here re Multiverse nomenclature, above we inherit the existing Universe nomenclature. 
___
*[`universe/interface.go` line 242 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiV3ryBVbVqAil5fiJ:-NaiV3ryBVbVqAil5fiK:b3ksi) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/interface.go#L242)):*
> ```Go
> 	// the minting outpoint will be returned. If neither are specified, then
> 	// proofs for all the inserted leaves will be returned.
> 	FetchIssuanceProof(ctx context.Context, id Identifier,
> ```

g

Roasbeef (commented):
> Reviewed 1 of 5 files at r1, 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416)* status: all files reviewed, 11 unresolved discussions (waiting on @ffranr)

ffranr (commented):
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416)* status: all files reviewed, 11 unresolved discussions (waiting on @Roasbeef)

*[a discussion](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#gh-1273945343:-Nalaeod696MAuyDoIaD:b-896fix) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

Can we also add fetching here? So then we'll be able to add that second level of proof and also verification. 
</blockquote></details>

Done.

___
*[a discussion](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#gh-1278167643:-Nalagt7C6iOOswjg86s:b-896fix) (no related file):*

Just to clarify, I was referring to this: https://github.com/lightninglabs/taproot-assets/blob/main/tapdb/universe.go#L138-L146
</blockquote></details>

Done.

___
*[a discussion](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#gh-1278167933:-NalaiPl5YWvYXknELuV:b-896fix) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

The issuance proof returned here should now also include the inclusion path into the multi-verse root. 
</blockquote></details>

Done.

___
*[a discussion](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#gh-1278168546:-NalaqOy7NpblfMS-eFn:b-896fix) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

See other comment, this also needs to have a multiverse root inclusion proof along side it. 
</blockquote></details>

Done.

___
___
*[`tapdb/universe.go` line 526 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiTzPG1tmIu42UT62-:-NaloW2CCtLCIH8YfMS0:b-cqova6) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/tapdb/universe.go#L526)):*

Why `nil` error here now?
</blockquote></details>

I made this change so that I could call `a.cfg.UniverseForest.FetchIssuanceProof` in `MintingArchive.RegisterIssuance` without needing to error handle if no proofs were found. I could just check that the returned proof list was empty.

Why did `universeFetchIssuanceProof` return an error if no proofs were found?
___
*[`tapdb/universe_forest.go` line 181 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiU6a481P3VFsYsFNj:-NalWZsQ5RH8iPQxWjYS:b-qqlmfu) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/tapdb/universe_forest.go#L181)):*

:+1: 

I think we can just name as Multiverse root here? Forest was sort of some tie over naming, unless we like that and want to elevate it more. 
</blockquote></details>

I've gone with multiverse. But there are more names to cleanup in code untouched by this PR.
___
*[`universe/base.go` line 143 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiUU2M8JHY8snqyHkF:-NalqRvh-7oc0Uml-6dz:b-xwd290) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/base.go#L143)):*
> Not so sure we want to do away with the base universe distinction here.

The problem I had was that `baseUni.FetchIssuanceProof(ctx, key)` didn't populate the new multiverse fields, so I would need to populate them at this point. Where as with `a.cfg.UniverseForest.FetchIssuanceProof` those new multiverse fields (in `IssuanceProof`) would have been populated.
___
*[`universe/base.go` line 150 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiUkgK9wR9Yr_4rMHf:-NalfcCl-w4pWwmMCpGP:bt0ebk0) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/base.go#L150)):*
> Ah so opted to return nil for this?

I don't think I understand your comment. Before the changes in this PR, we were returning
```
issuanceProof := proofs[0]
return issuanceProof, nil
```

here if any issuance proofs were found. We were already returning a `nil` error.

> IMO it's better to make that error condition explicit.

Do you mean explicitly checking for the no proofs found error at this point?
___
*[`universe/interface.go` line 152 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiV61rCj3Im2nyDh1B:-NalZfiP5fS_QnPV1K3k:b-896fix) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/interface.go#L152)):*

See comment here re Multiverse nomenclature, above we inherit the existing Universe nomenclature. 
</blockquote></details>

Done.
___
*[`universe/interface.go` line 242 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiV3ryBVbVqAil5fiJ:-Nalqs-U0KUgSD6Jyft-:bhn6rcj) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/interface.go#L242)):*

g
</blockquote></details>

I've updated the docs for these methods.

ffranr:
@Roasbeef I've added the RPC changes to this PR. They are pretty minimal. `QueryProof` and `InsertProof` should now both return the new multiverse fields. This is because `r.cfg.BaseUniverse` used in those functions actually queries the multiverse.

Roasbeef (changes requested):
> Reviewed 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416)* status: all files reviewed, 6 unresolved discussions (waiting on @ffranr and @guggero)

*[a discussion](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#gh-1278168447:-NaxEfM68wN8beQaD6lv:b-endxwa) (no related file):*
<details><summary><i>Previously, ffranr (Robyn Ffrancon) wrote…</i></summary><blockquote>

I've renamed `BaseForest` to `Forest`.
</blockquote></details>

Ah to clarify, I din't mean we needed to change the name just now. But the over all comment was that "forest" isn't used anywhere in the BIPs, and was just a place holder name. We'll need to make a decision re the actual semantics, eg: base means just issuance proofs as is. As you start to add this functionality, it starts to move away from that. 

___
___
*[`rpcserver.go` line 2717 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaxFd-87-4Lj3NTtSPZ:-NaxFd-87-4Lj3NTtSP_:b-boj80q) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/443b5ce25240e39f520d1094844a9eabfeaa3355/rpcserver.go#L2717)):*
> ```Go
> 		UniverseInclusionProof:   uniProof,
> 		AssetLeaf:                assetLeaf,
> 		MultiverseRoot:           multiverseRoot,
> ```

We should update the itests to also verify this information is populated and valid (can verify the inclusion proofs). 
___
*[`tapdb/universe.go` line 526 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiTzPG1tmIu42UT62-:-NaxFCip2_kZBs46V0u3:b-j2eext) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/tapdb/universe.go#L526)):*
> Why did `universeFetchIssuanceProof` return an error if no proofs were found?

So we can return a better error the the caller (we do this elsewhere like eg coin selection when no coins can satisfy the request). Otherwise, if you return `nil, nil` in that case, the caller needs to take care to not trigger a nil pointer panic. IMO it's better to make the error explicit: if you try to fetch a leaf that doesn't exist, the caller should be notified as such. 

How does this manifest to the RPC layer? Will we return `nil, nil` rather than an error that the leaf doesn't exist? 
___
*[`tapdb/universe_forest.go` line 181 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiU6a481P3VFsYsFNj:-NaxFlLT2Alt-0HORUSa:by4ot84) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/tapdb/universe_forest.go#L181)):*
<details><summary><i>Previously, ffranr (Robyn Ffrancon) wrote…</i></summary><blockquote>

I've gone with multiverse. But there are more names to cleanup in code untouched by this PR.
</blockquote></details>

SGTM re multi-verse. 
___
*[`universe/base.go` line 143 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiUU2M8JHY8snqyHkF:-NaxFxZ82PQBZ_OvtgEE:b-8my10y) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/base.go#L143)):*
<details><summary><i>Previously, ffranr (Robyn Ffrancon) wrote…</i></summary><blockquote>

> Not so sure we want to do away with the base universe distinction here.

The problem I had was that `baseUni.FetchIssuanceProof(ctx, key)` didn't populate the new multiverse fields, so I would need to populate them at this point. Where as with `a.cfg.UniverseForest.FetchIssuanceProof` those new multiverse fields (in `IssuanceProof`) would have been populated.
</blockquote></details>

I think that could be addressed with another layer of indirection. In other words, everything can now call into the multi-verse, which then calls into the base universe for the lower level operations. Eg, we'll also want to ensure normal fetching returns the multi-verse proofs, as well as the response value of this method. 
___
*[`universe/base.go` line 150 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiUkgK9wR9Yr_4rMHf:-NaxGUjACJQVKENaKdYN:b-7rdhls) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/base.go#L150)):*
<details><summary><i>Previously, ffranr (Robyn Ffrancon) wrote…</i></summary><blockquote>

> Ah so opted to return nil for this?

I don't think I understand your comment. Before the changes in this PR, we were returning
```
issuanceProof := proofs[0]
return issuanceProof, nil
```

here if any issuance proofs were found. We were already returning a `nil` error.

> IMO it's better to make that error condition explicit.

Do you mean explicitly checking for the no proofs found error at this point?
</blockquote></details>

I think I mean more the flip of the control flow: before if we got an error, we'd assume that meant no proof then we'd go to insert for the first time. If I were to re-write this section slightly, I'd check for the concrete error we want, then return the error as is if it was unexpected. 

IMO returning `nil` when nothing is found makes it harder on the caller to make sure they don't trigger a panic. Though I do agree that prior fragment would assume that only one proof was ever retuned, when in reality if you only partially specify the `MintingLeaf` struct, it's possible to get more than one proof.

universe/interface.go
```diff
@@ -41,14 +41,19 @@ type Identifier struct {
...
 	}
 
-	return hex.EncodeToString(i.AssetID[:])
+	return i.AssetID
+}
+
+// String returns a string representation of the ID.
+func (i *Identifier) String() string {
+	idBytes := i.Bytes()
+	return hex.EncodeToString(idBytes[:])
```
guggero, line 56: nit: could use `fn.ByteSlice(i.Bytes())` directly.

tapdb/universe_forest.go
```diff
@@ -125,3 +130,87 @@ func (b *BaseUniverseForest) RootNodes(
...
+	leaf *universe.MintingLeaf,
+	metaReveal *proof.MetaReveal) (*universe.IssuanceProof, error) {
+
+	var (
+		writeTx       BaseUniverseForestOptions
+		issuanceProof *universe.IssuanceProof
+	)
+
+	dbErr := b.db.ExecTx(
+		ctx, &writeTx, func(dbTx BaseUniverseForestStore) error {
```
guggero, line 208: style nit: could extract this closure into a named one defined before the call to `b.db.ExecTx()` which would save us one level of indentation for the whole block.

tapdb/universe.go
```diff
@@ -523,7 +523,7 @@ func universeFetchIssuanceProof(ctx context.Context,
 	}
 
 	if len(universeLeaves) == 0 {
-		return nil, ErrNoUniverseProofFound
+		return nil, nil
```
guggero, line 526: Shouldn't we do this on another level? Sounds to me like we'd want to know if the DB doesn't have a certain proof. But maybe we don't want to return an error on the RPC interface?

tapdb/universe_forest.go
```diff
@@ -7,17 +7,22 @@ import (
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/mssmt"
+	"github.com/lightninglabs/taproot-assets/proof"
 	"github.com/lightninglabs/taproot-assets/tapdb/sqlc"
 	"github.com/lightninglabs/taproot-assets/universe"
 )
 
+const mssmtNamespace = "multiverse"
```
guggero, line 15: Since all of those trees are MS-SMT trees, how about `multiverseNS` instead?

taprpc/universerpc/universe.proto
```diff
@@ -267,6 +267,14 @@ message AssetProofResponse {
 
     // The asset leaf itself, which includes the asset and the issuance proof.
     AssetLeaf asset_leaf = 4;
+
+    // MultiverseRoot is the root of the multiverse tree that includes this
+    // asset leaf.
+    MerkleSumNode multiverse_root = 5;
```
guggero, line 273: Should we also test the values of those fields in an itest?

universe/interface.go
ffranr: Do you want me to write the function like this?
```
// String returns a string representation of the ID.
func (i *Identifier) String() string {
	idBytes := fn.ByteSlice(i.Bytes())
	return hex.EncodeToString(idBytes)
}
```

Or do you mean like this?
```
// String returns a string representation of the ID.
func (i *Identifier) String() string {
	return hex.EncodeToString(fn.ByteSlice(i.Bytes()))
}
```

tapdb/universe.go
ffranr: I've reverted my change such that `FetchIssuanceProof` returns `ErrNoUniverseProofFound` like it did before. I'll check the RPC side.

ffranr:
> How does the multiverse root and proof look like for a universe that only cares about a single asset?

@guggero the upper multiverse tree would consist of a root node and a single leaf. The lower (inner) tree would consist of a root node and one leaf node for each issuance.

> I guess it would still be there but never updated after the first issuance?

I think that's right where an issuance only multiverse is concerned.

ffranr (commented):
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416)* status: all files reviewed, 10 unresolved discussions (waiting on @guggero and @Roasbeef)

*[a discussion](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#gh-1278168447:-NbG9G4dDaTDvCej49Gu:bogyqcf) (no related file):*

Ah to clarify, I din't mean we needed to change the name just now. But the over all comment was that "forest" isn't used anywhere in the BIPs, and was just a place holder name. We'll need to make a decision re the actual semantics, eg: base means just issuance proofs as is. As you start to add this functionality, it starts to move away from that. 
</blockquote></details>

Ah I get it now. I've changed `BaseForest` to `BaseMultiverse` for now. But I'll make another PR to rename everything consistently.

___
*[a discussion](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#gh-1284578762:-NbG9ZsB4ajw-7UOL62V:b-896fix) (no related file):*
<details><summary><i>Previously, guggero (Oliver Gugger) wrote…</i></summary><blockquote>

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

style nit: could extract this closure into a named one defined before the call to `b.db.ExecTx()` which would save us one level of indentation for the whole block.
</blockquote></details>

Done.

___
*[a discussion](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#gh-1284585863:-NbG9iZRAH1IsImIgeru:b-896fix) (no related file):*
<details><summary><i>Previously, guggero (Oliver Gugger) wrote…</i></summary><blockquote>

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

Shouldn't we do this on another level? Sounds to me like we'd want to know if the DB doesn't have a certain proof. But maybe we don't want to return an error on the RPC interface?
</blockquote></details>

Done.

___
*[a discussion](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#gh-1284589780:-NbG9o4JDzLP2TtPzNn0:b-896fix) (no related file):*
<details><summary><i>Previously, guggero (Oliver Gugger) wrote…</i></summary><blockquote>

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

Since all of those trees are MS-SMT trees, how about `multiverseNS` instead?
</blockquote></details>

Done.

___
___
*[`rpcserver.go` line 2717 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaxFd-87-4Lj3NTtSPZ:-NbFVlLXAf2mlvMmnw7P:b-896fix) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/443b5ce25240e39f520d1094844a9eabfeaa3355/rpcserver.go#L2717)):*

We should update the itests to also verify this information is populated and valid (can verify the inclusion proofs). 
</blockquote></details>

Done.
___
*[`tapdb/universe.go` line 526 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiTzPG1tmIu42UT62-:-NbFfbX6BEXdhCXskynX:bavivd) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/tapdb/universe.go#L526)):*

> Why did `universeFetchIssuanceProof` return an error if no proofs were found?

So we can return a better error the the caller (we do this elsewhere like eg coin selection when no coins can satisfy the request). Otherwise, if you return `nil, nil` in that case, the caller needs to take care to not trigger a nil pointer panic. IMO it's better to make the error explicit: if you try to fetch a leaf that doesn't exist, the caller should be notified as such. 

How does this manifest to the RPC layer? Will we return `nil, nil` rather than an error that the leaf doesn't exist? 
</blockquote></details>

I've reverted this change. `a.cfg.UniverseForest.FetchIssuanceProof` now returns an error if a proof is not found. The error is also returned by the RPC endpoint.
___
*[`universe/base.go` line 150 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiUkgK9wR9Yr_4rMHf:-NbG4q2s8gVDx-TljQ-x:b-uefc6d) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/base.go#L150)):*

I think I mean more the flip of the control flow: before if we got an error, we'd assume that meant no proof then we'd go to insert for the first time. If I were to re-write this section slightly, I'd check for the concrete error we want, then return the error as is if it was unexpected. 

IMO returning `nil` when nothing is found makes it harder on the caller to make sure they don't trigger a panic. Though I do agree that prior fragment would assume that only one proof was ever retuned, when in reality if you only partially specify the `MintingLeaf` struct, it's possible to get more than one proof. 
</blockquote></details>

I've modified this part so that it checks for the concrete error `ErrNoUniverseProofFound`.

ffranr (commented):
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416)* status: 0 of 12 files reviewed, 10 unresolved discussions (waiting on @guggero and @Roasbeef)

___
*[`universe/base.go` line 143 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NaiUU2M8JHY8snqyHkF:-NbGMvYq09AcegIrQRU7:ba0kol7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a776f938ad0a75b045763daed595910f724f0ebf/universe/base.go#L143)):*
> will what's in `base.go` allow you to get proofs of any arbitrary state transition of an asset

The `MintingArchive` that's in `base.go` currently only deals with issuance. I think it would make sense to replace it with a persistent implementation that can get proofs of any arbitrary state transition.

I think it would be simpler conceptually to have a one-to-one map between MS-SMT tree types (universe/multiverse) and their corresponding persistent implementations.

Roasbeef:
> One thing I don't yet fully understand (should probably read the BIP again) is: How does the multiverse root and proof look like for a universe that only cares about a single asset?
> I guess it would still be there but never updated after the first issuance?

If a Universe only cares about a single asset, then the multi-verse proof would just be the most minimal MS-SMT inclusion proof: so just the hash, then the series of compressed nodes. In theory this can be left out all together. It matters more once the issuance events are also timestamped in the chain. 

For an asset w/o a group key, then yeah things are immutable after that first issuance. You just use this to then reference back to the root genesis output for all verification. In that case though as you note at, the multi-verse root isn't all that useful. It would be useful however when wanting to create a sort of tracking set of two distinct, but related assets. I thin there's a large design space here w.r.t what the user configuration for managing these trees and cohorts looks like in the end.

Roasbeef (changes requested):
> Reviewed 11 of 12 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416)* status: 11 of 12 files reviewed, 7 unresolved discussions (waiting on @ffranr and @guggero)

___
*[`itest/universe_test.go` line 187 at r7](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NbHTVBE4adN6DCFndlC:-NbHTVBE4adN6DCFndlD:b-t0d7yx) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/9edb933730999f80168621a2a4551ccfcd36dce7/itest/universe_test.go#L187)):*
> ```Go
> 	// Convert asset ID bytes into a fixed size array.
> 	var assetIdFixedSize [32]byte
> 	if len(firstAssetID) <= len(assetIdFixedSize) {
> ```

Why's this length check needed at all vs just the copy? It can also use `fn.ToArray`: https://github.com/lightninglabs/taproot-assets/blob/a4b3c0693da97faf76c92c58f20b2ee675b0f105/fn/memory.go#L23-L30
___
*[`universe/interface.go` line 231 at r7](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NbHTzuE4QloOkuIBiiF:-NbHTzuE4QloOkuIBiiG:bxxqg9e) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/9edb933730999f80168621a2a4551ccfcd36dce7/universe/interface.go#L231)):*
> ```Go
> // particular base universe, while this is used to obtain aggregate information
> // about the universes.
> type BaseMultiverse interface {
> ```

I think maybe instead we can keep the name here, and instead enforce the super/subset behavior with a either a new interface method, or a method that returns the type/description of the set of sorted assets? Non-blocking though, as something we'll need to stamp everywhere else in the codebase as we progress with these changes.

Roasbeef (approved):
:lgtm:

LGTM 🗿

Fat fingered the block thing 

> Reviewed 1 of 12 files at r7.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-:-NbHUCmQ7Xo-gqJgcc8G:b-gu8i3c)* status: all files reviewed, 7 unresolved discussions (waiting on @ffranr and @guggero)

ffranr (commented):
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/416)* status: 11 of 12 files reviewed, 7 unresolved discussions (waiting on @guggero and @Roasbeef)

___
*[`itest/universe_test.go` line 187 at r7](https://reviewable.io/reviews/lightninglabs/taproot-assets/416#-NbHTVBE4adN6DCFndlC:-NbKXxAXDev95jluRxFk:b-896fix) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/9edb933730999f80168621a2a4551ccfcd36dce7/itest/universe_test.go#L187)):*

Why's this length check needed at all vs just the copy? It can also use `fn.ToArray`: https://github.com/lightninglabs/taproot-assets/blob/a4b3c0693da97faf76c92c58f20b2ee675b0f105/fn/memory.go#L23-L30
</blockquote></details>

Done.

universe/base.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/davecgh/go-spew/spew"
+	"github.com/juju/errors"
```
guggero, line 11: This is the wrong `errors`! Should be from the standard package.

tapdb/universe.go
```diff
@@ -523,7 +517,7 @@ func universeFetchIssuanceProof(ctx context.Context,
 	}
 
 	if len(universeLeaves) == 0 {
-		return nil, nil
+		return nil, universe.ErrNoUniverseProofFound
```
guggero, line 520: Squash this commit with the previous one?

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/jackc/pgconn v1.12.0
 	github.com/jackc/pgerrcode v0.0.0-20201024163028-a0d42d470451
 	github.com/jessevdk/go-flags v1.4.0
+	github.com/juju/errors v0.0.0-20220331221717-b38fca44723b
```
guggero, line 22: @Roasbeef I'm not sure if Reviewable is incorrect or I'm using it wrong: But it was hiding this change from me ("there are no changes in these files or you don't need to review them"), which is quite weird... Going to stay with GitHub for now I think.

universe/base.go
```diff
@@ -137,15 +138,22 @@ func (a *MintingArchive) RegisterIssuance(ctx context.Context, id Identifier,
...
-		issuanceProof := proofs[0]
-		return issuanceProof, nil
+	issuanceProofs, err := a.cfg.UniverseForest.FetchIssuanceProof(
+		ctx, id, key,
+	)
+	switch {
+	case err == nil && len(issuanceProofs) > 0:
+		// No error returned and at least one proof found, so we'll
+		// return the first one.
+		return issuanceProofs[0], nil
```
guggero, line 151: nit: add empty line after non-trivial `case` block.

itest/universe_test.go
```diff
@@ -211,6 +245,14 @@ func testUniverseSync(t *harnessTest) {
 	)
 }
 
+// UnmarshalMerkleSumNode un-marshals a protobuf MerkleSumNode.
```
guggero, line 248: nit: godoc name/case mismatch.

guggero (commented):
Veeeerry close, just a small issue with a Golang import.

guggero (approved):
LGTM :tada: 

Just needs a `go mod tidy` to revert the change to the `go.mod`, then we're good to go!

Issue #317, opened 2023-05-23
Ensure daemon is robust to chain re-orgs


ffranr:
_unfinished_

guggero:
I did a quick investigation in how the `lnd` RPCs behave on a re-org.
Adding some additional debug statements and RPC calls to the `open_channel_reorg_test` itest:

```
=== RUN   TestLightningNetworkDaemon/tranche00/64-of-132/btcd/open_channel_reorg_test
    lnd_open_channel_test.go:90: Funding txid: b5e0bb6db4657d9ede47c4fb6550ad240ccbd66a2b865f53936a01e0bbf7d813
    lnd_open_channel_test.go:113: Got transaction event, tx_hash=b5e0bb6db4657d9ede47c4fb6550ad240ccbd66a2b865f53936a01e0bbf7d813, block_hash=5d98ef3a15a6b92a455e899461789927b3cb1a35a435324eacbc6401b8bdee9d, confirmations=1
    lnd_open_channel_test.go:143: Got confirmation event for tx_hash=b5e0bb6db4657d9ede47c4fb6550ad240ccbd66a2b865f53936a01e0bbf7d813, block_hash=5d98ef3a15a6b92a455e899461789927b3cb1a35a435324eacbc6401b8bdee9d

(actual reorg happens)

    lnd_open_channel_test.go:145: Got reorg event for tx_hash=b5e0bb6db4657d9ede47c4fb6550ad240ccbd66a2b865f53936a01e0bbf7d813
    lnd_open_channel_test.go:113: Got transaction event, tx_hash=b5e0bb6db4657d9ede47c4fb6550ad240ccbd66a2b865f53936a01e0bbf7d813, block_hash=3e903b6f24bedf8329941375f9510024b5ecf8391fcf1367159e4c42e45b9163, confirmations=1
    lnd_open_channel_test.go:143: Got confirmation event for tx_hash=b5e0bb6db4657d9ede47c4fb6550ad240ccbd66a2b865f53936a01e0bbf7d813, block_hash=3e903b6f24bedf8329941375f9510024b5ecf8391fcf1367159e4c42e45b9163

```

 - `transaction event` is from `lnrpc.SubscribeTransactions()` (doesn't need any input params)
 - `confirmation event` or `reorg event` is from `chainrpc.RegisterConfirmationsNtfn()` (needs TXID and output pkScript as input params)

ffranr:
[...]

PR #419, opened 2023-07-26
re-org safety: watch asset transactions and re-create proofs when necessary

guggero:
~Depends on https://github.com/lightninglabs/taproot-assets/pull/424, the first 5 commits aren't part of this PR anymore.~

Fixes https://github.com/lightninglabs/taproot-assets/issues/317.

The PR is fully ready for review!

TODOs:
 - ~Add re-org watcher to custodian to watch received assets~
 - ~Add re-org watcher to freighter to watch sent assets~
 - ~Allow minting proofs to be updated in the universe/federation after re-org~
 - ~Integration tests for above mentioned use cases~
 - ~Unit test re-org watcher~

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/419)


ffranr (commented):
I haven't finished reviewing, looks good so far. I think you could roll the first two commits into a PR and get those merged as is. I'll try out the new review tool with this PR.

guggero:
> I haven't finished reviewing, looks good so far. I think you could roll the first two commits into a PR and get those merged as is. I'll try out the new review tool with this PR.

Good idea, done: https://github.com/lightninglabs/taproot-assets/pull/424

tapgarden/interface.go
```diff
@@ -89,11 +89,11 @@ const (
 	// created.
 	BatchStateFinalized BatchState = 5
 
-	// BatchStateCancelled denotes that a batch has been cancelled, and
-	// will not be passed to a caretaker.
+	// BatchStateSeedlingCancelled denotes that a batch has been cancelled,
```
Roasbeef, line 92: 👍

tapcfg/config.go
```diff
@@ -100,6 +100,10 @@ const (
 	// defaultUniverseSyncInterval is the default interval that we'll use
 	// to sync Universe state with the federation.
 	defaultUniverseSyncInterval = time.Minute * 10
+
+	// defaultReOrgSafeDepth is the default number of confirmations we'll
+	// wait for before considering a transaction safely buried in the chain.
+	defaultReOrgSafeDepth = 6
```
Roasbeef, line 106: Ahh 6, the magic Bitcoin number! 😆

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,424 @@
...
+	updateCb ProofUpdateCb
+}
+
+// anchorTxNotification is a struct that holds all proof watch subscriptions for
+// a single anchor transaction. It is possible to have different callbacks for
+// different proofs within the same anchor transaction.
+type anchorTxNotification struct {
+	proofsRegistrations []*proofRegistration
+
+	cancel context.CancelFunc
```
Roasbeef, line 28: context FTW!

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,424 @@
...
+}
+
+// numProofs returns the total number of proofs that are being watched for this
+// anchor transaction.
+func (a *anchorTxNotification) numProofs() int {
+	return fn.Reduce(
```
Roasbeef, line 34: 😎

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,424 @@
...
+		func(agg int, reg *proofRegistration) int {
+			return agg + len(reg.proofs)
+		},
+	)
+}
+
+// firstProof returns the first proof that is being watched for this anchor
+// transaction.
+func (a *anchorTxNotification) firstProof() *proof.Proof {
+	for _, reg := range a.proofsRegistrations {
```
Roasbeef, line 45: Why can't this just be `a.proofsRegistrations[0]`? Or rather even just check the length first.

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,424 @@
...
+	if err != nil {
+		w.reportErr(fmt.Errorf("unable to register for block "+
+			"epoch notifications: %w", err))
+		return
+	}
+
+	for {
+		select {
+		case newProofs := <-w.incomingProofs:
+			txHash := newProofs.proofs[0].AnchorTx.TxHash()
```
Roasbeef, line 156: Perhaps we can do a quick sanity check here that all the proofs have the same anchor tx hash?

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,424 @@
...
+func (w *ReOrgWatcher) reportErr(err error) {
+	select {
+	case w.cfg.ErrChan <- err:
+	case <-w.Quit:
+	}
+}
+
+// waitForConf waits for the anchor transaction of the given proofs to reach a
+// safe confirmation depth.
+func (w *ReOrgWatcher) waitForConf(ctx context.Context,
```
Roasbeef, line 289: Move this above the main loop so it's defined before it's used?

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,424 @@
...
+	go func() {
+		defer confEvent.Cancel()
+		defer w.Wg.Done()
+		defer cancel()
+
+		for {
+			select {
+			// If the transaction was confirmed normally, we don't
+			// need to do anything and can stop watching it.
+			case conf, ok := <-confEvent.Confirmed:
```
Roasbeef, line 340: Don't we always expect to his this case? Since it's already confirmed by the time we do the initial registration request.

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,424 @@
...
+				log.Infof("Anchor TX %v was re-organized out "+
+					"of the chain, will update proof "+
+					"with next confirmation", txHash)
+
+				// We continue to watch the transaction until
+				// it reaches a safe confirmation depth. We
+				// basically only use this signal to log the
+				// message above. We expect another confirmation
+				// to come in once the transaction is included
+				// in a new block in the re-organized chain.
```
Roasbeef, line 362: Would think we also need to send a signal back to the main goroutine that a re-org occurred.

proof/archive.go
```diff
@@ -449,3 +449,55 @@ func (m *MultiArchiver) RemoveSubscriber(
...
+		ScriptKey: *scriptPubKey,
+	}
+
+	blob, err := archive.FetchProof(ctx, locator)
+	if err != nil {
+		return fmt.Errorf("unable to fetch current proof: %w", err)
+	}
+
+	f := &File{}
+	if err := f.Decode(bytes.NewReader(blob)); err != nil {
```
Roasbeef, line 472: In the future, it would be ideal to be able to do this _without_ reading the entire file in. IIUC, this is made a bit more difficult since we store the number of proofs as a prefix. However, I think we could do something like:
  * read the number of proofs
  * append the new proof and update the incremental hash 
  * seek into the file (depends on what type of `io.Reader` we actually have) to update the count of the number of proofs

tapgarden/planter.go
```diff
@@ -801,6 +805,30 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
...
+// (the batch is finalized after a single confirmation).
+func (c *ChainPlanter) updateMintingProofs(proofs []*proof.Proof) error {
+	ctx, cancel := c.WithCtxQuitNoTimeout()
+	defer cancel()
+
+	headerVerifier := GenHeaderVerifier(ctx, c.cfg.ChainBridge)
+	for idx := range proofs {
+		p := proofs[idx]
+
+		err := proof.ReplaceLastProofInBlob(
```
Roasbeef, line 823: I think there's an edge case here if after we've done the minting, the asset is them immediately transferred after. We'd then have the updated proof, but override the final transfer.

tapgarden/custodian.go
```diff
@@ -590,6 +603,27 @@ func (c *Custodian) setReceiveCompleted(event *address.Event,
...
+// updateReceivedProofs is called by the re-org watcher when it detects a re-org
+// and has updated the received proofs.
+func (c *Custodian) updateReceivedProofs(proofs []*proof.Proof) error {
+	// Let's not be interrupted by a shutdown.
+	ctxt, cancel := c.CtxBlocking()
+	defer cancel()
+
+	headerVerifier := GenHeaderVerifier(ctxt, c.cfg.ChainBridge)
+	for idx := range proofs {
+		err := proof.ReplaceLastProofInBlob(
```
Roasbeef, line 615: I think the case also applies here: recv, transfer, then re-org happens.

tapdb/universe.go
```diff
@@ -323,9 +324,15 @@ func (b *BaseUniverseTree) RegisterIssuance(ctx context.Context,
...
 	dbErr := b.db.ExecTx(ctx, &writeTx, func(dbTx BaseUniverseStore) error {
-		issuanceProof, _, err = universeRegisterIssuance(
-			ctx, dbTx, b.id, key, leaf, metaReveal,
-		)
+		if replace {
+			issuanceProof, _, err = universeReplaceIssuance(
+				ctx, dbTx, b.id, key, leaf,
+			)
+		} else {
+			issuanceProof, _, err = universeRegisterIssuance(
```
Roasbeef, line 332: Alternatively, Universe servers could have a rule where they only accept assets that have been sufficiently buried. This wouldn't help with the multi-verse case though, when you want to send a proof to someone as soon as it's been confirmed.

universe/interface.go
```diff
@@ -169,9 +169,9 @@ type BaseBackend interface {
 	// tree, stored at the base key. The metaReveal type is purely
 	// optional, and should be specified if the genesis proof committed to
 	// a non-zero meta hash.
-	RegisterIssuance(ctx context.Context,
-		key BaseKey, leaf *MintingLeaf,
-		metaReveal *proof.MetaReveal) (*IssuanceProof, error)
+	RegisterIssuance(ctx context.Context, key BaseKey, leaf *MintingLeaf,
+		metaReveal *proof.MetaReveal, replace bool) (*IssuanceProof,
```
Roasbeef, line 173: Is the bool actually needed here? Seems like we can just assume that if we get a new valid proof, but with an updated block hash (actually in the chain), then we can go ahead and accept it.

Roasbeef (changes requested):

tapgarden/re-org_watcher.go
ffranr: After that check, I think it might make sense to pass in `txHash` into `waitForConf` rather than select the first proof once more.

jharveyb: The check could go in `WatchProofs()` to be as early as possible.

Related, AFAICT we only access that first proof / any proofs to look up the block height, block hash, or TXID for the group of proofs. We could store those 3 vars directly as members of `proofRegistration` instead of fetching them from a proof repeatedly.

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,424 @@
...
+	for idxR := range proofNtfn.proofsRegistrations {
+		r := proofNtfn.proofsRegistrations[idxR]
+		for idxP := range r.proofs {
+			p := r.proofs[idxP]
+
+			// We can now update the proof with the new
+			// block.
+			p.BlockHeight = conf.BlockHeight
+			p.BlockHeader = conf.Block.Header
+			merkleProof, err := proof.NewTxMerkleProof(
```
jharveyb, line 405: If all the proofs in a registration share an anchor TX, I think we can make this merkle proof once per-registration and then just copy it into each actual asset proof?

tapgarden/caretaker.go
```diff
@@ -958,12 +958,23 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
...
+			mintingProofBlobs,
 		)
 		if err != nil {
 			return 0, fmt.Errorf("unable to confirm batch: %w", err)
 		}
 
+		// Now that we've confirmed the batch, we'll hand over the
+		// proofs to the re-org watcher.
+		newProofs := make([]*proof.Proof, 0, len(mintingProofs))
+		for key := range mintingProofs {
```
jharveyb, line 971: nit: could use maps.Values.

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,424 @@
...
+					"proofs: %w", err))
+				return
+			}
+
+		case newBlock := <-newBlockChan:
+			log.Infof("New block at height %d", newBlock)
+			w.bestHeight = uint32(newBlock)
+
+			for txid := range w.pendingProofs {
+				proofNtfn := w.pendingProofs[txid]
```
ffranr, line 233: nit: I would move this line down so that it's just before `proofNtfn.cancel()`. We only need `proofNtfn` within the `if` clause (if we're cancelling).

tapgarden/custodian.go
jharveyb: As in we transfer while the recv is 0-conf, or something like a 1-block reorg that unconfirms the original recv anchor? Or the worse case of a 2+-block reorg unconfirming both anchors.

Wondering if we'd also need to track this for the input UTXOs for minting, since those could also be 1-conf and we'd have the same situation.

tapgarden/re-org_watcher.go
guggero: Yes, we should always initially get a confirmation event here. What we're actually interested in are _subsequent_ events in case of a re-org. But I updated some of the log messages so they also make sense for the first confirmation (which we already expect to happen at this point, so shouldn't be `log.Warnf()` for the `Anchor TX ... was already confirmed in block`).

tapgarden/re-org_watcher.go
guggero: We can't really do anything when the TX is re-organized out of the chain. That basically meant it went back into the mempool. So we just log the info here and wait for it to be confirmed in a block again.
Or do you suggest we mark the proof as invalid here as it's possible it will never make it into the chain again?

proof/archive.go
guggero: Yes, definitely. I thought we already had an issue, but couldn't find it. Will create a new one.

jharveyb (commented):
Looks good! The callback in each subsystem pattern is really neat.

I don't think I saw rebroadcast logic for the actual anchor TX after a reorg, is that handled by the backing LND or did I just miss it?

tapgarden/planter.go
guggero: Hmm, you're right. We do check that we're trying to update the correct proof (script key check in `ReplaceLastProofInBlob`). But I guess we should instead try to find which proof we need to replace (it might not be the last one in the file) and replace that instead of always assuming it's the last in the file.

tapdb/universe.go
guggero: Hmm, we could add that restriction exclusively for issuance proofs. But then if you mint and want to send immediately, you would be blocked by that restriction I guess.

universe/interface.go
guggero: Hmm, yes, if we change `InsertUniverseLeaf` into an upsert, this is possible. Makes more sense, updated.

tapgarden/re-org_watcher.go
guggero: No, we also need `proofNtfn` to call `firstRegistration()` to get the block height.

tapgarden/re-org_watcher.go
ffranr: It seems to me that informing the main goroutine might come in handy when it comes to signalling to a user (via UI) that the previously confirmed send event is now unconfirmed again.

tapgarden/planter.go
ffranr: > an edge case here if after we've done the minting, the asset is them immediately transferred after

I wander if we would encounter a similar problem when sending an asset twice in close succession.

tapgarden/re-org_watcher.go
guggero: Re-orgs happen very rarely AFAIK. And I would assume that if they happen the new chain would very likely contain very similar transactions. So the chance of a transaction being un-confirmed for a long time is probably quite low. So I'm not sure we actually want to build out any user-facing thing here. And if we do, I'd suggest doing it in a follow-up PR.

ffranr (commented):
This is great. I couldn't see any further issues that haven't already been commented on.

The only dynamics that concern me are those that relate to multiple actions (mint, send, recv) occurring in close succession. But I think that this PR in its current state is still a big improvement and should probably be merged regardless.

tapgarden/re-org_watcher.go
Roasbeef: > Or do you suggest we mark the proof as invalid here as it's possible it will never make it into the chain again?

I think I was just sort of confused re the control flow here (message passing ordering). Will take another pass over things with fresh eyes.

proof/archive.go
Roasbeef: SGTM, I think once we kick off the load tests (cc @positiveblue) this'll start to pop up in a few places.

tapgarden/planter.go
Roasbeef: Yeah I guess it also depends on behavior re nested zero conf transactions as well. Even if confirmed, eg: minting confirmed in block `A`, then transfer in block `B`, then re-org back to block `Z`. In this case we need to update both of them, but also keep the ordering: `A'`, then `B'` and not the other way around.

Roasbeef: Assuming the re-org watcher has _both_ requests (so watch for `A` and then `B`), if it groups things by sort of root prev ID (the descendent chain), then it should be able to enforce the ordering of updates.

guggero:
I've updated the proof watching logic and also added a new itest to verify the new behavior.

Unfortunately the way we store proofs is a bit weird, so this requires a bit of an explanation.
Assume we have a chain of proofs (A mints, sends full amount to itself, then sends some to B, receiving change) of an asset with ID `xyz`:
`ScriptKey_A_1 -> ScriptKey_A_2 -> ScriptKey_A_3/ScriptKey_B_3` where `ScriptKey_x` stands for a script key belonging to node `x` and `_y` stands for the block where that proof was confirmed in.

If node A minted (`ScriptKey_A_1`) the asset and transferred it to itself (`ScriptKey_A_2`), then sent part of it it out to node B (`ScriptKey_B_3` and `ScriptKey_A_3` for the change), then the two nodes have the following proofs on disk (and in the DB):
```
node A:
  proofs/xyz/ScriptKey_A_1.assetproof (num_proofs=1)
  proofs/xyz/ScriptKey_A_2.assetproof (num_proofs=2)
  proofs/xyz/ScriptKey_A_3.assetproof (num_proofs=3)

node B:
  proofs/xyz/ScriptKey_B_3.assetproof (num_proofs=3)
```

Now further assume that both blocks 2 and 3 are not yet fully confirmed (<6 confs) and a re-org happens that removes both anchor transactions out of the chain, then we need to update all proofs for block 2 and 3.

So for block 2 we would need to update the following files:
```
node A:
  proofs/xyz/ScriptKey_A_2.assetproof (replace the last proof)
  proofs/xyz/ScriptKey_A_3.assetproof (replace the second-to-last proof)

node B:
  proofs/xyz/ScriptKey_B_3.assetproof (replace the second-to-last proof)
```

And for block 3 we would need to update the following files:
```
node A:
  proofs/xyz/ScriptKey_A_3.assetproof (replace last proof)

node B:
  proofs/xyz/ScriptKey_B_3.assetproof (replace last proof)
```

Now the issue is that in case both anchor transactions for state 2 and 3 confirm in the same block now, we don't know in what order the confirmation notifications come in. So however we order things, the proof files where we need to update more than one proof will be temporarily invalid until we have updated _both_ proofs.

Currently we do this by just using a dummy header verifier (see `TODO` in `ReplaceProofInBlob`).

I think this is fine for now as eventually the proofs will be valid again once all confirmation notifications were acted upon. I don't think it's worth attempting to do a full dependency tracking and resolution mechanism to address this (it would require the proof watcher to be refactored quite a bit).

Curious to hear your thoughts though.

jharveyb (changes requested):
New changes makes sense! I think this approach works for now, and realistically reorgs should be rare enough that this works out. I'm a bit concerned about reading all proofs on startup, but AFAICT that will come from the DB and not files on disk?

> Reviewed 4 of 30 files at r4, 36 of 36 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/419#-:-NbHh2mh3YFsS5nNNZlR:byq0boq)* status: all files reviewed, 20 unresolved discussions (waiting on @ffranr, @guggero, @positiveblue, and @Roasbeef)

___
*[`itest/re-org_test.go` line 326 at r5](https://reviewable.io/reviews/lightninglabs/taproot-assets/419#-NbHgudRBd4fbQ1lxSV6:-NbHgudRBd4fbQ1lxSV7:bjmtdb4) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/5ab7cc1c540100a8f44abf66b327dc5eab2493d3/itest/re-org_test.go#L326)):*
> ```Go
> 
> 	// Cleanup by mining the minting tx again.
> 	newBlock := t.lndHarness.MineBlocksAndAssertNumTxes(1, 2)[0]
> ```

Would be cool to also test the wallet re-broadcast in this situation, but not needed in this PR.
___
*[`tapgarden/re-org_watcher.go` line 407 at r5](https://reviewable.io/reviews/lightninglabs/taproot-assets/419#-NbHBLqX0LFAosd0pAR6:-NbHBLqX0LFAosd0pAR7:b-q7kyts) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/5ab7cc1c540100a8f44abf66b327dc5eab2493d3/tapgarden/re-org_watcher.go#L407)):*
> ```Go
> 		case newBlock := <-newBlockChan:
> 			log.Infof("New block at height %d", newBlock)
> 			w.bestHeight.Store(newBlock)
> ```

Do we need to check that these blocks are actually connected? Otherwise I think if we have `confs = w.cfg.SafeDepth-1`, and we observe a reorg that's 1 block longer than the current best chain, we could stop watching some anchor TXs prematurely.

Or maybe this is handled inside the chain bridge itself?
___
*[`tapgarden/re-org_watcher.go` line 499 at r5](https://reviewable.io/reviews/lightninglabs/taproot-assets/419#-NbHDM441RE3UBz3GjYq:-NbHDM441RE3UBz3GjYr:b4c9lyr) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/5ab7cc1c540100a8f44abf66b327dc5eab2493d3/tapgarden/re-org_watcher.go#L499)):*
> ```Go
> 		// watcher expects only proofs to be grouped that are committed
> 		// in the same anchor transaction.
> 		if w.ShouldWatch(p) {
> ```

:thumbsup:

jharveyb (approved):

ffranr:
@guggero Are the universe stats updated if a issuence is re-orged?

Also, do we restrict universe syncing to proofs which are at adequate depth?

(Even if these issues exist I don't think they should block this PR.)

ffranr (approved):
I'm finding it difficult to feel confident that we've addressed every edge case that we might care about when it comes to re-orgs. There's a lot to think about. But I think that this is a fantastic step in the right direction, so I think we should merge this.

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/419#-:-NbKOnGY6PF_GtOZ6OCS:b-9ensp9)* status: all files reviewed, 20 unresolved discussions (waiting on @guggero, @jharveyb, @positiveblue, and @Roasbeef)

guggero:
@jharveyb 

> I'm a bit concerned about reading all proofs on startup, but AFAICT that will come from the DB and not files on disk?

The proofs are queried from both the file system and the DB and the first one that has a certain proof returns it. It is a bit unfortunate that we have to do this. Maybe if this becomes a bottleneck we could also store the proof's block height in the DB so we can filter more efficiently?

> Would be cool to also test the wallet re-broadcast in this situation, but not needed in this PR.

I checked again, and because we're using the `btcwallet`'s TX publish mechanism, each transaction we create in `tapd` is automatically re-broadcast by the wallet (now on every new incoming block I think). Since that's part of `lnd`'s responsibility, I don't think we need to add an extra test in this repo.

> Do we need to check that these blocks are actually connected?

That's also part of `lnd`'s validation logic. All re-orgs are validated by the chain backend first before `lnd` sends out notifications. So again, I don't think we need to check this in `tapd`.

@ffranr 

> Are the universe stats updated if a issuence is re-orged?

Yes, once the issuer detects a re-org, it will update the universe(s) it initially sent the issuance proof to. And the universe will do an `upsertAssetGen()` which includes the block height of the new proof.

> Also, do we restrict universe syncing to proofs which are at adequate depth?

No. But what I added is logic in the proof watcher to address this:
Let's say an asset is minted by Alice in block 100, Bob syncs his daemon with Alice's universe, a re-org happens. Alice will update the universe and let it know the proof changed. But Bob doesn't know that, and doesn't really care. BUT, Bob now receives an asset from Alice with the updated issuance proof, which is valid. And the next time Bob syncs with the Universe, he notices the roots have changed and re-syncs the proof as well.
So that shouldn't be an issue, really (can always manually re-sync to the universe if necessary).

jharveyb (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/419)* status: all files reviewed, 20 unresolved discussions (waiting on @ffranr, @guggero, @positiveblue, and @Roasbeef)





@ guggero thanks for double-checking on the `btcwallet` and `lnd` behavior in this situation!

Re: proof querying, yes tagging proofs with block height sounds helpful, since the current locator scheme provides no sense of time / ordering.

Also, for the archiver proof reading, would the file backend terminate early if the DB returns first, or complete with the result dropped? Wondering if slow proof reads would slow down DB reads for the same proof or set of proofs (can we short-circuit proof reads, essentially). No changes needed in this PR.

guggero:
> Also, for the archiver proof reading, would the file backend terminate early if the DB returns first, or complete with the result dropped? Wondering if slow proof reads would slow down DB reads for the same proof or set of proofs (can we short-circuit proof reads, essentially). No changes needed in this PR.

Reading a proof goes through each backend sequentially and the first that returns a non-error is used and the rest is then skipped. Since we add the DB first, then the file archive, the DB will always be asked first. If we feel like that might slow things down, we could perhaps switch the order of the backends. But that might have other impacts, so I'd do it in a separate PR.

guggero:
The itest ran into:
```
test_harness.go:142: Error outside of test: *errors.errorString Error running server: received critical error from subsystem: received confirmation for unknown anchor TX 06d6e9abd34ef40a31c7b660e6b6112b44a3b1332a9b62ad4f40f0ff0d272a2b
```

Will take a look.

guggero:
> The itest ran into:
> 
> ```
> test_harness.go:142: Error outside of test: *errors.errorString Error running server: received critical error from subsystem: received confirmation for unknown anchor TX 06d6e9abd34ef40a31c7b660e6b6112b44a3b1332a9b62ad4f40f0ff0d272a2b
> ```
> 
> Will take a look.

Okay, looks like this was a flake, wasn't able to reproduce locally. But I removed the error and turned it into a log statement since this might happen in some edge cases around restart and when a block and the conf notification come in at the same time (if we process the block update first and the conf notification afterwards).

PR should be good to go now!

proof/file.go
```diff
@@ -222,6 +222,23 @@ func (f *File) ProofAt(index uint32) (*Proof, error) {
 	return proof, nil
 }
 
+// LocateProof calls the given predicate for each proof in the file and returns
+// the first proof (and the index) where the predicate returns true. If no proof
+// is found, this returns ErrNoProofAvailable.
+func (f *File) LocateProof(cb func(*Proof) bool) (*Proof, uint32, error) {
+	for i := 0; i < f.NumProofs(); i++ {
```
Roasbeef, line 229: Perhaps this should scan _backwards_ instead? Since most of the time we're only interested in the last few proofs when a re-org happens. Could add a bool here, or expose via another method.

proof/file.go
```diff
@@ -282,28 +299,44 @@ func (f *File) AppendProof(proof Proof) error {
...
 	}
 
-	f.proofs[len(f.proofs)-1] = &hashedProof{
+	f.proofs[index] = &hashedProof{
 		proofBytes: proofBytes,
 		hash:       hashProof(proofBytes, prevHash),
 	}
 
+	// We now need to re-hash all proofs after this one.
+	for i := index + 1; i < uint32(len(f.proofs)); i++ {
```
Roasbeef, line 334: 👍

server.go
```diff
@@ -142,6 +144,20 @@ func (s *Server) initialize(interceptorChain *rpcperms.InterceptorChain) error {
 		return fmt.Errorf("unable to start asset custodian: %v", err)
 	}
 
+	// We need to fetch all unspent assets from the database so the re-org
+	// watcher can properly detect re-orgs for not yet fully buried assets.
+	ctx := context.Background()
+	chainAssets, err := s.cfg.AssetStore.FetchAllAssets(ctx, false, nil)
```
Roasbeef, line 150: I think we only need to read all the assets not sufficiently buried? So we can  use a depth filter here to check what the height of their current confirmed managed UTXO is. This could help us consderiably cut down on start up time as the amount of assets the daemon holds scales up.

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,565 @@
...
+
+			f := &proof.File{}
+			err = f.Decode(bytes.NewReader(blob))
+			if err != nil {
+				startErr = fmt.Errorf("unable to decode "+
+					"proof: %w", err)
+				return
+			}
+
+			err = w.MaybeWatch(f, w.DefaultUpdateCallback())
```
Roasbeef, line 153: Re the earlier suggestion, if we only pass the items that actually are in "danger" of a re-org, then we wouldn't need this check. We'd also be able to avoid reading all the proof data above here as well (for all the open assets that is).

tapgarden/re-org_watcher.go
```diff
@@ -0,0 +1,565 @@
...
+				}
+
+			// If the transaction was re-organized out, we need to
+			// update the proof and store it in the proof archive.
+			case <-reOrgChan:
+				log.Infof("Anchor TX %v was re-organized out "+
+					"of the chain, will update proof "+
+					"with next confirmation", txHash)
+
+				// We continue to watch the transaction until
```
Roasbeef, line 244: Ahh ok I think that's the behavior I missed before: that we'll get another confirmation notification in the event of a re-org. So given that, why do we even need to reorg channel in the first place?

tapgarden/re-org_watcher.go
guggero: I think even with the optimized query where we only fetch the assets not yet buried, we'll want to inspect the full proof file as there might be ancestor transactions that haven't fully confirmed yet.

tapgarden/re-org_watcher.go
guggero: For now, just for logging, which was helpful for debugging the itest. In the future, we could react to it.

Roasbeef:
Needs a rebase after the utxo locking PR was merged!

Issue #421, opened 2023-07-27
[bug]: ImportProofs failed because of proof path error on Windows


btcwer:
### Background

I ran tapd on my Windows host. It looked like a new asset minted successfully using tapcli, but actually not. When using tapcli to transfer the asset, an error occurred complaining that no proof found.
Then I dug into tapd's log, finding that there was an error at the end of asset minting as below:

[ERR] GRDN: unable to advance state machine: unable to insert proofs: unable to store proof: open C:\Users\user1\AppData\Local\Tapd\data\testnet\proofs\af41b433e950c5c35436458749047deb025796b43f827391eb9b3d6c192c7000\02dfa973b2a20304caccd48352f35f3d374bf74b0d28cd90daa70780d191a8d8e4.assetproof: The system cannot find the path specified.  

After investigation, I think the bug should be here in the ImportProofs function:
os.MkdirAll(path.Dir(proofPath), 0750)

The path.Dir() should be switched to filepath.Dir(), as path.Dir can't handle Windows back slash properly.

I can't rebuild tapd project with above fix due to my env, but I have verified the fix in a test program.

### My environment

bitcoind(Testnet), lnd, tapd running on the same Windows host

tapd release: tapd v0.2.3-alpha (taproot-assets-windows-amd64-v0.2.3.zip)

guggero:
Thanks for the report and the investigation. Using `filepath` makes a lot of sense, I created a fix for it.

PR #422, opened 2023-07-27
multi: use filepath instead of path to support Windows

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/421.


ffranr (approved):
:+1: 

I suppose we could also just change to
```
import (
    path "path/filepath"
)
```

Roasbeef (approved):
LGTM 🔥

PR #424, opened 2023-07-27
re-org safety: preparatory commits

guggero:
Depends on https://github.com/lightningnetwork/lnd/pull/7859.

Spun out of https://github.com/lightninglabs/taproot-assets/pull/419 in order to reduce the final diff.
All commits are mostly related to re-org safety but might appear out of context without also taking a look at #419.


proof/archive.go
```diff
@@ -214,6 +219,12 @@ func (f *FileArchiver) ImportProofs(_ context.Context,
 			return err
 		}
 
+		// Can't replace a file that doesn't exist yet.
+		if replace && !lnrpc.FileExists(proofPath) {
```
Roasbeef, line 223: 👍

tapdb/assets_store.go
```diff
@@ -1329,22 +1329,67 @@ func (a *AssetStore) importAssetFromProof(ctx context.Context,
...
 	proofs ...*proof.AnnotatedProof) error {
 
 	var writeTxOpts AssetStoreTxOptions
 	err := a.db.ExecTx(ctx, &writeTxOpts, func(q ActiveAssetsStore) error {
 		for _, p := range proofs {
-			err := a.importAssetFromProof(ctx, q, p)
-			if err != nil {
-				return fmt.Errorf("unable to import asset: %w",
-					err)
+			if replace {
```
Roasbeef, line 1381: Would be good to have a basic test for this in the db.

Roasbeef (changes requested):

proof/mint.go
```diff
@@ -149,25 +153,20 @@ func NewMintingBlobs(params *MintParams,
 	}
 
 	ctx := context.Background()
-	blobs := make(AssetBlobs, len(proofs))
+
+	// Before we return the proofs, we'll verify that we actually generated
+	// valid ones.
```
ffranr, line 158: how about: `// Verify generated proofs.`

rpcserver.go
```diff
@@ -1099,11 +1099,13 @@ func (r *rpcServer) VerifyProof(ctx context.Context,
...
-	_, err = proofFile.Verify(
-		ctx, headerVerifier,
-	)
+	_, err = proofFile.Verify(ctx, headerVerifier)
 	valid := err == nil
 
+	if !valid {
+		rpcsLog.Errorf("Proof verification failed with err: %v", err)
+	}
+
```
ffranr, line 1108: This pattern is strange. I think we should get rid of the variable `valid` and this `if` clause can just inspect `err`.

When populating the `Valid` field in the `return` statement, we can just set it to `err == nil`.

server.go
```diff
@@ -297,15 +297,18 @@ func (s *Server) RunUntilShutdown(mainErrChan <-chan error) error {
...
 	case err := <-mainErrChan:
 		if err == nil {
+			srvrLog.Debug("Main err chan closed")
 			return nil
 		}
 
 		// We'll report the error to the main daemon, but only if this
 		// isn't a context cancel.
 		if fn.IsCanceled(err) {
+			srvrLog.Debugf("Got context canceled error: %v", err)
```
ffranr, line 311: :+1:

ffranr (approved):
Nice changes!

jharveyb (approved):
Looks good! No further comment besides adding test coverage for the new `replace` option.

tapdb/assets_store.go
guggero: Makes sense, added.

guggero:
PR is ready now, addressed all the review comments. Though we should wait for https://github.com/lightningnetwork/lnd/pull/7859 to be merged (updated that as well with a reproducing unit test).

Roasbeef:
`lnd` PR has been merged!

guggero:
Done, ready for final review.

Roasbeef (approved):
LGTM ☔️

PR #429, opened 2023-08-08
tappsbt: fix incorrect test vectors

guggero:
The address already contains the internal key as a compressed (33-byte) public key. But the test vectors were incorrectly using the x-only (32-byte) representation of the internal key, which lead to an incorrect stripping of the parity byte in the wrong place, which this PR now fixes.

### DIFF
--- a/tappsbt/address.go
+++ b/tappsbt/address.go
@@ -2,7 +2,6 @@ package tappsbt
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


jharveyb (approved):
LGTM, great catch!

Roasbeef (approved):
LGTM 🌪️

Issue #261, opened 2023-02-09
asset wallet: introduce coin locking and coin selection mutex


guggero:
The asset wallet introduced in https://github.com/lightninglabs/taro/pull/244 does not allow a coin to be marked as being used (=locking), which needs to be done at the database layer.
Once we can mark asset coins as being used/locked, we also need to make sure that access to coin selection and the locking APIs are covered by a single coin selection mutex similar to how it's done in `lnd`.

Roasbeef:
May also want to have a bit that says if an output has been fully spent or not. Either that, or we need to delete it.

dstadulis:
related issue #374

PR #431, opened 2023-08-08
wallet: add asset coin locking

guggero:
~Depends on https://github.com/lightninglabs/taproot-assets/pull/429.~

Fixes https://github.com/lightninglabs/taproot-assets/issues/261.

This PR introduces two new "coin locks":
 - One on the wallet level (similar to `btcwallet`'s `WithCoinSelectLock`) that allows multiple DB calls to be made while holding an exclusive lock (e.g. do coin selection then "lease" a specific coin to spend, those cannot be done in a single DB transaction the way the code is structured)
 - One on the `managed_utxo` level, which is called a "lease" (to reflect the terminology used in `lnd`: An application can "lease" (=lock/reserve) a bitcoin level UTXO to be spent in a transaction. So all assets committed to that BTC output are not available for other RPC users as long as the UTXO is being leased.

TODOs (might decide to put those into a second PR):
 - ~Allow releasing a lease (unlock coin) through RPC~
 - ~Allow filtering assets by `leased=true/false`~


tapdb/assets_store.go
```diff
@@ -202,6 +205,14 @@ type ActiveAssetsStore interface {
 	// serialized outpoint.
 	DeleteManagedUTXO(ctx context.Context, outpoint []byte) error
 
+	// LeaseManagedUTXO leases a managed UTXO identified by the passed
+	// serialized outpoint.
+	LeaseManagedUTXO(ctx context.Context, arg LeaseManagedUTXO) error
```
Roasbeef, line 212: Same type and method name here? Maybe for the type name: `InsertLeasedUTXO`?

tapdb/assets_store.go
```diff
@@ -202,6 +205,14 @@ type ActiveAssetsStore interface {
...
+
+	// ReleaseManagedUTXO releases a managed UTXO identified by the passed
```
Roasbeef, line 214: A biit of am ambiguity here re: release vs re-lease. Maybe we should move to unlock/lock?

tapdb/assets_store.go
```diff
@@ -358,6 +369,19 @@ type ChainAsset struct {
 	// AnchorTapscriptSibling hash is equal to the Taproot root hash of the
 	// anchor output.
 	AnchorTapscriptSibling []byte
+
+	// AnchorLeaseOwner is the identity of the application that currently
```
Roasbeef, line 386: Not sure this needs to be part of the `ChainAsset` struct, but still getting through the diff.

Roasbeef: Also perhaps we can constrain this a bit further, I think it's meant to be a `[32]byte`?

tapdb/assets_store.go
```diff
@@ -718,6 +742,17 @@ func dbAssetsToChainAssets(dbAssets []ConfirmedAsset,
 			AnchorMerkleRoot:       sprout.AnchorMerkleRoot,
 			AnchorTapscriptSibling: sprout.AnchorTapscriptSibling,
 		}
+
+		// We only set the lease info if the lease is actually still
+		// valid and hasn't expired.
+		owner := sprout.AnchorLeaseOwner
+		expiry := sprout.AnchorLeaseExpiry
+		if len(owner) > 0 && expiry.Valid &&
```
Roasbeef, line 763: How/when are leases to be cleaned up?

tapdb/assets_store.go
```diff
@@ -1425,10 +1469,14 @@ func queryChainAssets(ctx context.Context, q ActiveAssetsStore,
...
-// parameters. If no commitment is found, ErrNoCommitment is returned.
+// parameters. If no commitment is found, ErrNoCommitment is returned. With
+// mustBeLeased the caller decides whether the asset output should've been
+// leased before or not. If mustBeLeased is false, then the state of the lease
+// is not checked.
 func (a *AssetStore) FetchCommitment(ctx context.Context, id asset.ID,
 	anchorPoint wire.OutPoint, groupKey *asset.GroupKey,
-	scriptKey *asset.ScriptKey) (*tapfreighter.AnchoredCommitment, error) {
+	scriptKey *asset.ScriptKey,
+	mustBeLeased bool) (*tapfreighter.AnchoredCommitment, error) {
```
Roasbeef, line 1492: So this is a sort of sanity check we want to implement in the constraints of a DB transaction? What about a similar coarse coin selection mutex to what we have in `lnd`? 

Looking at it from another angle: wouldn't one already fail a coin selection attempt if the outputs needed are already leased? Since first we'd try to perform coin selection with either `ListEligibleCoins` or `SelectForAmount`?

tapdb/assets_store.go
```diff
@@ -1468,12 +1523,65 @@ func (a *AssetStore) ListEligibleCoins(
...
 	assetFilter.Spent = sqlBool(false)
+	assetFilter.Leased = sqlBool(false)
 
 	return a.queryCommitments(ctx, assetFilter)
 }
 
+// LeaseCoin leases/locks/reserves a coin for the given lease owner until the
+// given expiry. This is used to prevent multiple concurrent coin selection
+// attempts from selecting the same coin.
+func (a *AssetStore) LeaseCoin(ctx context.Context, utxoOutpoint wire.OutPoint,
```
Roasbeef, line 1559: Hmm, I wonder if we want leasing at a lower level of abstractions: the full prev ID schema. I guess this works tho, since you can only ever spend that UTXO once. 

In the future, more granular leasing might be needed once we start to optimistically combine outgoing sends in the freighter.

tapdb/sqlc/migrations/000002_assets.up.sql
```diff
@@ -143,7 +143,19 @@ CREATE TABLE IF NOT EXISTS managed_utxos (
...
 
-    txn_id INTEGER NOT NULL REFERENCES chain_txns(txn_id)
+    txn_id INTEGER NOT NULL REFERENCES chain_txns(txn_id),
+    
+    -- The identity of the application that currently has a lease on this UTXO.
+    -- If NULL, then the UTXO is not currently leased. A lease means that the
+    -- UTXO is being reserved/locked to be spent in an upcoming transaction and
+    -- that it should not be available for coin selection through any of the
+    -- wallet RPCs.
+    lease_owner BLOB,
```
Roasbeef, line 153: Could alternatively be a distinct table, then the unlock/lock process is an insert/delete vs modification.

tapdb/assets_store.go
```diff
@@ -320,6 +322,17 @@ type AssetStore struct {
...
+
+	// coinLock is a read/write mutex that is used to ensure that only one
+	// goroutine is attempting to call any coin selection related methods at
+	// any time. This is necessary as some of the calls to the store (e.g.
+	// ListEligibleCoins -> LeaseCoin) are called after each other and
+	// cannot be placed within the same database transaction. So calls to
+	// those methods must hold this coin lock. We use this in conjunction
+	// with an atomic  boolean, so we can check if the lock is held without
+	// having to acquire the mutex.
+	coinLock     sync.Mutex
```
Roasbeef, line 334: I think this should live one level higher in the porter. So here: https://github.com/lightninglabs/taproot-assets/blob/edf59c62d5e4acd407d2675414e5d86e78eaf47b/tapfreighter/wallet.go#L143. 

In the past we've had some hard to track down bugs as a result of inserting user-space DB level mutexes. This way we also make minimal assumptions re concurrency control beyond what's provided by the DB.

tapdb/assets_store.go
```diff
@@ -320,6 +322,17 @@ type AssetStore struct {
...
+	coinLockHeld atomic.Bool
```
Roasbeef, line 335: I don't think we'd need this atomic either. 

Similar to `lnd`, we'd also want to ensure that any relevant RPCs also respect the mutex. If they go to the wallet as suggested above, then that should be handled. Alternatively, they can grab the mutex itself, similar to the closure pattern we have in lnd.

tapdb/assets_store.go
```diff
@@ -2640,6 +2677,20 @@ func (a *AssetStore) FetchAssetMetaByHash(ctx context.Context,
 	return assetMeta, nil
 }
 
+// WithCoinLock will acquire the coin lock and execute the passed function.
+// Callers to any coin related functions should use this to ensure that the
+// coin lock is held across the entire set of operations.
+func (a *AssetStore) WithCoinLock(fn func() error) error {
```
Roasbeef, line 2683: See comments above about moving this. I think sort of unclear with it being so low level exactly how the caller should interact with it.

tapfreighter/interface.go
```diff
@@ -87,8 +87,16 @@ type CoinLister interface {
...
+	//
+	// NOTE: The coin lock must be held when calling this function.
 	ListEligibleCoins(context.Context,
 		CommitmentConstraints) ([]*AnchoredCommitment, error)
+
+	// WithCoinLock will acquire the coin lock and execute the passed
+	// function. Callers to any coin related functions should use this to
+	// ensure that the coin lock is held across the entire set of
+	// operations.
+	WithCoinLock(fn func() error) error
```
Roasbeef, line 114: I think this placement makes sense, but then I think the implementation can just reside within `CoinSelect ` (the concrete implementation).

tapfreighter/wallet.go
```diff
@@ -385,23 +394,34 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
...
-	)
-	if err != nil {
-		return nil, fmt.Errorf("unable to complete coin selection: %w",
-			err)
-	}
 
-	log.Infof("Identified %v eligible asset inputs for send of %d to %x",
-		len(eligibleCommitments), fundDesc.Amount, fundDesc.ID[:])
+	var selectedCommitments []*AnchoredCommitment
+	err := f.cfg.CoinSelector.WithCoinLock(func() error {
```
Roasbeef, line 442: 👍

tapfreighter/wallet.go
```diff
@@ -385,23 +394,34 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
...
-		ctx, constraints,
-	)
-	if err != nil {
-		return nil, fmt.Errorf("unable to complete coin selection: %w",
-			err)
-	}
 
-	log.Infof("Identified %v eligible asset inputs for send of %d to %x",
-		len(eligibleCommitments), fundDesc.Amount, fundDesc.ID[:])
+	var selectedCommitments []*AnchoredCommitment
```
Roasbeef, line 441: Perhaps we also want to have a wrapper type here that makes it easier to release all the coin in a single go?

tapfreighter/wallet.go
```diff
@@ -385,23 +394,34 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
...
+				"%w", err)
+		}
 
-	selectedCommitments, err := f.cfg.CoinSelector.SelectForAmount(
-		fundDesc.Amount, eligibleCommitments, PreferMaxAmount,
-	)
+		log.Infof("Identified %v eligible asset inputs for send of %d "+
+			"to %x", len(eligible), fundDesc.Amount, fundDesc.ID[:])
+
+		selectedCommitments, err = f.cfg.CoinSelector.SelectForAmount(
```
Roasbeef, line 454: Looks like leasing the coins is missing here? Otherwise two callers can come in series, then obtain the same set of coins, and attempt to lease them both.

tapdb/sqlc/queries/assets.sql
```diff
@@ -676,6 +685,16 @@ WHERE (
 DELETE FROM managed_utxos
 WHERE outpoint = $1;
 
+-- name: LeaseManagedUTXO :exec
+UPDATE managed_utxos
+SET lease_owner = @lease_owner, lease_expiry = @lease_expiry
```
Roasbeef, line 690: See above about making this an insert. That way, we can have the unique key be the outpoint itself. Then if two callers try to lease the same UTXO, they get an error. Otherwise, with the current API the same coin can be leased twice.

tapdb/assets_store_test.go
```diff
@@ -677,16 +683,21 @@ func TestSelectCommitment(t *testing.T) {
...
+			assetGen.genAssets(t, store, tc.assets)
 
 			// With the assets inserted, we'll now attempt to query
 			// for the set of matching assets based on the
 			// constraints.
-			selectedAssets, err := assetsStore.ListEligibleCoins(
-				ctx, tc.constraints,
-			)
+			var selectedAssets []*tapfreighter.AnchoredCommitment
+			err := store.WithCoinLock(func() error {
```
Roasbeef, line 694: Re the prior comments, we then wouldn't have to leak the details of the coin selection lock into the database level. Instead, we can have a higher level unit test that just gives a mocked implementation of `CoinLister`.

tapfreighter/chain_porter.go
```diff
@@ -202,14 +202,10 @@ func (p *ChainPorter) assetsPorter() {
...
-			if err != nil {
-				log.Warnf("Unable to advance state machine: %v",
-					err)
-				req.kit().errChan <- err
-				continue
-			}
+			// possible in its own goroutine. The status will be
+			// reported through the different channels of the send
+			// package.
+			go p.advanceState(sendPkg, req.kit())
```
Roasbeef, line 208: 👍

tapfreighter/wallet.go
```diff
@@ -152,6 +174,27 @@ func (s *CoinSelect) ListEligibleCoins(ctx context.Context,
...
 
+// LeaseCoin leases/locks/reserves a coin for the given lease owner
+// until the given expiry. This is used to prevent multiple concurrent
+// coin selection attempts from selecting the same coin.
+//
+// NOTE: The coin lock must be held when calling this function.
+func (s *CoinSelect) LeaseCoin(ctx context.Context, utxoOutpoint wire.OutPoint,
+	leaseOwner []byte, expiry time.Time) error {
+
+	return s.coinLister.LeaseCoin(ctx, utxoOutpoint, leaseOwner, expiry)
```
Roasbeef, line 185: Not related to this line of code, but wanting to jot this down: I think we should make sure that once an outgoing package has been written to disk, with the transfer transaction broadcast (but not yet confirmed), that UTXO shouldn't be available for leasing. Otherwise, if confirmations take a while, then we end up double spending ourselves. 

With the way things are in `btcwallet`, once a mined transaction is inserted, then the lease is also removed: https://github.com/btcsuite/btcwallet/blob/07be54bc22cf4dd2dd16a9788d2fc88952c438ae/wtxmgr/tx.go#L462-L471

tapfreighter/wallet.go
```diff
@@ -416,6 +459,21 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
 				"%w", err)
 		}
 
+		// We now need to lock/lease/reserve those selected coins so
+		// that they can't be used by other processes.
+		expiry := time.Now().Add(defaultCoinLeaseDuration)
```
Roasbeef, line 464: 👍

Roasbeef (changes requested):
Great work! Changes look to be headed in the right direction at a high level. I think my major review comment relates to exactly _where_ the coin selection lock should live it. Right now it lives at the bottom of the abstraction layer in the db, which also means we now have a mutex in the way of db code. In the past (eg lnd), we've run into issues with user space mutexes on top of what the db is already doing. I think we should try to lift the mutex up to the highest level, understanding that some calls may just want a snapshot of the db, and don't necessarily need the same level of contextual locking as needed when trying to perform a transfer. 

Also as is, it looks like an output can be locked twice. If the lease are a new table, and then we use an outpoint as the primary key, we can reject such behavior as the unique conflict then fails. Finally, I didn't see where the leases are expired, seems to be sort of implicit rn?

tapdb/assets_store.go
guggero: Right, makes sense. I went with `UpdateUTXOLease` and `DeleteUTXOLease` since we'll also want to be able to extend the lease in the future.

tapdb/assets_store.go
guggero: I called it `DeleteUTXOLease` here but will try to make it clear that this unlocks a UTXO on the higher levels. Wanted to re-use as much of the terminology we use in `lnd` (even if it's not optimal) to avoid confusion.

tapdb/assets_store.go
guggero: I think it's useful to eventually show this in the asset list RPC. That way we don't need an explicit `ListAssetLeases` RPC, we can just add the `leased` boolean as a filter on the `ListAssets` request params.

tapdb/assets_store.go
guggero: This isn't currently done. Not sure what the best approach would be. In `btcwallet` I think we check the leases on certain actions. We could do the same here as well, e.g. whenever we list the coins or insert a new lease.
Or we could do it with a timed job. Going to see what makes most sense while addressing the other comments.

tapdb/assets_store.go
guggero: The `FetchCommitment` is needed when we go to anchor a signed vPSBT. So at that point the caller MUST have leased the outpoint already.
Flow is:
 1. Fund vPSBT -> Calls `ListEligibleCoins`, then `SelectForAmount`, then adds lease for the "selected" coin. PSBT is returned to caller (or freighter state machine is advanced, but either way, this step is done)
 2. Sign vPSBT -> No DB interaction required
 3. AnchorVirtualPsbts -> Calls `FetchCommitment` to load the inputs required to create the new tree. At this point we select the coin being spent and we need to make sure it was previously leased or is currently still leased.

The other place where `FetchCommitment` is called is for creating an asset ownership proof. We also need to select the full commitment for that, but there we don't care if the asset is currently leased or not.

tapdb/assets_store.go
guggero: Makes sense. So we would either batch sends together in a timed manner. Or use RBF to replace a previously published anchoring TX? Either way, I think we'll need some more design around that and then probably go with a new column on the asset table itself.

tapdb/sqlc/migrations/000002_assets.up.sql
guggero: I did look into that but found that it makes the coin selection SELECT statement quite a bit more complicated. Because then we'd need to do a `LEFT JOIN` on that leases table and also have WHERE statements that evaluates those columns based on the `leased` parameter.

tapfreighter/wallet.go
guggero: Yes, that's added in a later commit. Added a TODO here for now to help with the review flow.

tapdb/sqlc/queries/assets.sql
guggero: See comment above. I think it makes things unnecessarily hard.
And a coin cannot be leased twice because of the wallet-level coin lock:
 1. Caller A comes in, does: `WalletLock( ListEligibleCoins => returns coins X, Y, Z => SelectForAmount => select coin X => LeaseCoin(X))`
 2. Caller B comes in while call A is still ongoing. Once it acquires the lock, it does: `WalletLock( ListEligibleCoins => returns coins Y, Z => SelectForAmount => select coin Y => LeaseCoin(Y))`

tapdb/sqlc/queries/assets.sql
```diff
@@ -398,7 +400,14 @@ JOIN internal_keys
 JOIN managed_utxos utxos
     ON assets.anchor_utxo_id = utxos.utxo_id AND
       (utxos.outpoint = sqlc.narg('anchor_point') OR
-       sqlc.narg('anchor_point') IS NULL)
+       sqlc.narg('anchor_point') IS NULL) AND
+       CASE
+           WHEN sqlc.narg('leased') = true THEN
+               (utxos.lease_owner IS NOT NULL AND utxos.lease_expiry > @now)
+           WHEN sqlc.narg('leased') = false THEN
+               (utxos.lease_owner IS NULL OR utxos.lease_expiry <= @now)
```
ffranr, line 408: Perhaps we should add `OR utxos.lease_expiry IS NULL` to this `WHEN` clause.

guggero: Good catch, added.

guggero:
Latest push addresses most comments. Though I think I should address the following TODOs before a full re-review:
 - Add more unit tests around the leases and the `CoinSelect` struct
 - Periodically (or upon certain requests) remove old leases
 - Allow filtering assets by lease status on the RPC, assert leases in itest

ffranr (commented):
Nice work! Looks good to me. I don't have too much to add beyond Laolu's comments.

I think we should probably add an integration test to ensure that the send process fails if all candidate UTXOs are locked.

guggero:
> I think we should probably add an integration test to ensure that the send process fails if all candidate UTXOs are locked.

We do that here: https://github.com/lightninglabs/taproot-assets/blob/814ea4b374179fda9a119629e0602b333c749563/itest/send_test.go#L766

tapdb/sqlc/migrations/000002_assets.up.sql
Roasbeef: Hmm not sure I follow, for lease aware coin selection, don't we just need to first get the set of leased asset IDs (the internal primary key one), then assert that the returned set isn't found amongst that table? So imagining it as a CTE. 

In either case will check out the latest posted version here.

tapdb/sqlc/queries/assets.sql
Roasbeef: Ah yeh that's indeed the case with the full API path as exists today. 

I think I was just viewing it more from the PoV of the lowest level data structure here: the table itself. If we used the `outpoint` (or `outpoint || assetID` or w/e) as a primary key, then we'd be able to entirely rely on the DB to preserve the invariant that something can only be leased once: another insert would fail. As is, it would be possible for to callers (strictly on the db level just to zoom in) to each lease the same UTXO, with the second replacing/mutating the `lease_owner` field. 

As you mention above it still works as is, but if we wanted to try and normalize things further, I think we'd put the leases in a distinct table. Though up until now, things were normalized mostly on a best effort basis, so we don't necessarily have a strict policy here. Either way, just wanted to point this out as we recently-ish learned that the `sqlite` migration support isn't as fully featured as `postgres` so it'll be more difficult to change later.

tapdb/assets_store.go
```diff
@@ -726,7 +764,12 @@ func dbAssetsToChainAssets(dbAssets []ConfirmedAsset,
 // constraintsToDbFilter maps application level constraints to the set of
 // filters we use in the SQL queries.
 func constraintsToDbFilter(query *AssetQueryFilters) QueryAssetFilters {
-	var assetFilter QueryAssetFilters
+	assetFilter := QueryAssetFilters{
+		Now: sql.NullTime{
+			Time:  time.Now().UTC(),
```
Roasbeef, line 769: Not blocking at all here, but I think we'll want to eventually start to use `clock.Clock` (from lnd) here so we can eventually mock out the time for unit tests.

tapfreighter/chain_porter.go
```diff
@@ -24,10 +24,6 @@ import (
 
 // ChainPorterConfig is the main config for the chain porter.
 type ChainPorterConfig struct {
-	// CoinSelector is the interface used to select input coins (assets)
-	// for the transfer.
-	CoinSelector CoinSelector
```
Roasbeef, line 29: 👍

tapfreighter/wallet.go
```diff
@@ -383,25 +412,15 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
...
-	eligibleCommitments, err := f.cfg.CoinSelector.ListEligibleCoins(
-		ctx, constraints,
-	)
-	if err != nil {
-		return nil, fmt.Errorf("unable to complete coin selection: %w",
-			err)
+		MinAmt:   fundDesc.Amount,
 	}
-
-	log.Infof("Identified %v eligible asset inputs for send of %d to %x",
```
Roasbeef, line 396: Nice simplification!

Roasbeef (commented):
The change set is looking pretty good! Will pull it down and mess around with some of the new functionality on the CLI to see how intuitive/surprising it is

guggero:
Added all unit tests that I wanted to add, ready for another round of review.

tapdb/assets_store.go
```diff
@@ -722,15 +749,33 @@ func dbAssetsToChainAssets(dbAssets []ConfirmedAsset,
 			AnchorMerkleRoot:       sprout.AnchorMerkleRoot,
 			AnchorTapscriptSibling: sprout.AnchorTapscriptSibling,
 		}
+
+		// We only set the lease info if the lease is actually still
+		// valid and hasn't expired.
+		owner := sprout.AnchorLeaseOwner
+		expiry := sprout.AnchorLeaseExpiry
+		if len(owner) > 0 && expiry.Valid &&
+			expiry.Time.UTC().After(a.clock.Now().UTC()) {
```
ffranr, line 758: nit: the if condition is getting complicated here.

I would consider evaluating a bool variable before the if clause:
```
var (
    hasOwner = len(owner) > 0
    isExpired = expiry.Valid && expiry.Time.UTC().After(a.clock.Now().UTC())
)

if hasOwner && isExpired {
```

Or maybe not, it's a style.

tapdb/assets_store.go
```diff
@@ -1819,9 +1819,13 @@ func (a *AssetStore) queryCommitments(ctx context.Context,
...
 // the set of changes to disk (the pending inputs and outputs) but doesn't mark
-// the batched spend as being finalized.
+// the batched spend as being finalized. The final lease owner and expiry are
+// the lease parameters that are set on the input UTXOs, since we assume the
+// parcel will be broadcast after this call. So we'll want to lock the input
+// UTXOs for forever, which means the expiry should be far in the future.
 func (a *AssetStore) LogPendingParcel(ctx context.Context,
-	spend *tapfreighter.OutboundParcel) error {
+	spend *tapfreighter.OutboundParcel, finalLeaseOwner [32]byte,
+	finalLeaseExpiry time.Time) error {
```
ffranr, line 1828: nit: I don't understand why we need the `final` prefix here in `finalLeaseOwner` and `finalLeaseExpiry`. Consider removing?

tapdb/assets_store_test.go
```diff
@@ -594,6 +614,246 @@ func (a *assetGenerator) bindKeyGroup(i int, op wire.OutPoint) *btcec.PublicKey
...
+
+	// At first, none of the assets should be leased.
+	selectedAssets, err := assetsStore.FetchAllAssets(
+		ctx, false, false, nil,
+	)
+	require.NoError(t, err)
+	require.Len(t, selectedAssets, numAssets)
+
+	// Now we lease the first asset for 1 hour. This will cause the second
+	// one also to be leased, since it's on the same anchor transaction.
```
ffranr, line 787: :+1:

ffranr (approved):
I think I've managed to review the changes since my last review. Everything looks good to me. Nice addition of the clock to different db stores. And a test which ensures that locking an outpoint affects each asset anchored at that outpoint :+1:

Roasbeef (approved):
LGTM 🦆

PR #432, opened 2023-08-08
address: add proof courier TLV field to Tap address

ffranr:
Add mandatory TLV field. Ensure encoding/decoding unit tests pass.

### DIFF
--- a/address/address.go
+++ b/address/address.go
@@ -78,6 +78,12 @@ const (
 	TapScriptVersion uint8 = 0
 )
 
+// ProofCourierAddr is the address of the proof courier that will be used to
+// distribute related proofs for this address.
+//
+// NOTE: See the proof package for more details.
+type ProofCourierAddr []byte
+
 // Tap represents a Taproot Asset address. Taproot Asset addresses specify an
 // asset, pubkey, and amount.
 type Tap struct {
@@ -118,6 +124,10 @@ type Tap struct {
 	// assetGen is the receiving asset's genesis metadata which directly
 	// maps to its unique ID within the Taproot Asset protocol.
 	assetGen asset.Genesis
+
+	// ProofCourierAddr is the address of the proof courier that will be
+	// used to distribute related proofs for this address.
+	ProofCourierAddr ProofCourierAddr
 }
 
 // New creates an address for receiving a Taproot asset.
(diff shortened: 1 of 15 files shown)


ffranr:
Closes issue https://github.com/lightninglabs/taproot-assets/issues/434

address/address.go
```diff
@@ -326,6 +330,7 @@ func (a *Tap) DecodeRecords() []tlv.Record {
 		newAddressInternalKeyRecord(&a.InternalKey),
 		newAddressTapscriptSiblingRecord(&a.TapscriptSibling),
 		newAddressAmountRecord(&a.Amount),
+		newProofCourierAddrRecord(&a.ProofCourierAddr),
```
guggero, line 333: Needs to be added to the encode records as well. And to the unit test (which would've found this).

address/records.go
```diff
@@ -32,6 +32,9 @@ const (
 
 	// addrAmountType is the TLV type of the amount of the asset.
 	addrAmountType addressTLVType = 8
+
+	// addrProofCourierType is the TLV type of the proof courier address.
+	addrProofCourierAddrType addressTLVType = 8
```
guggero, line 37: This is colliding with the `addrAmountType`. Must be 10 to be mandatory.

address/records.go
```diff
@@ -91,3 +94,16 @@ func newAddressAmountRecord(amount *uint64) tlv.Record {
 		asset.VarIntEncoder, asset.VarIntDecoder,
 	)
 }
+
+func newProofCourierAddrRecord(addr *string) tlv.Record {
+	addrBytes := []byte(*addr)
```
guggero, line 99: I don't think this will work for decoding, as the result isn't written back into the `addr` pointer. Maybe we should instead make the addr just a `[]byte`?

guggero (changes requested):
Approach looks okay, though I don't think the code works just yet.

address/address.go
```diff
@@ -118,6 +118,10 @@ type Tap struct {
 	// assetGen is the receiving asset's genesis metadata which directly
 	// maps to its unique ID within the Taproot Asset protocol.
 	assetGen asset.Genesis
+
+	// ProofCourierAddr is the address of the proof courier that will be used to distribute
+	// related proofs for this address.
+	ProofCourierAddr []byte
```
Roasbeef, line 124: I think we'd want this to be further abstracted. What's `[]byte` even mean here? Would expect a `string`, but then you'd abstract away from the actual underlying mechanism.

Roasbeef (changes requested):

address/address.go
ffranr: I can add a more descriptive type with an enum for supported protocols. And `[]byte` serialization/deserialization.

ffranr: Oli said the following in keybase:

> I think a valid answer to the comment linked above is just: to avoid circular package dependencies the address just knows it's a string that contains an URI like scheme. how that looks and what it means is defined in the proof package (with the courier)

I agree with him and I've encountered the circular package dependency problem.

ffranr:
I'm closing this in favour of https://github.com/lightninglabs/taproot-assets/pull/450 where the branch has been pushed to `lightninglabs/taproot-assets`. I can then open another PR based on this new branch.

Issue #428, opened 2023-08-07
[bug]: `oneof` in REST query parameters doesn't work


inviti8:
### Background

Querying a taproot metadata doesn't appear to work as described, or I am missing something.  I have successfully minted a test asset using REST, however cannot construct a working url for the meta data query of my minted asset.  

### Your environment

 LTND: Version: 0.16.4-beta commit=v0.16.4-beta, build=production, logging=default, debuglevel=info
TAPD: Latest

### Steps to reproduce

```
REST_HOST = 'localhost:8089'
MACAROON_PATH = 'TAPROOT-ASSETS_DIR/regtest/taproot-assets.macaroon'
TLS_PATH = 'TAPROOT-ASSETS_DIR/tls.cert'

asset_id = ASSET ID GOES HERE'
meta_hash = 'META HASH GOES HERE'


url = f'https://{REST_HOST}/v1/taproot-assets/assets/meta?asset_id={asset_id}'

macaroon = codecs.encode(open(MACAROON_PATH, 'rb').read(), 'hex')
headers = {'Grpc-Metadata-macaroon': macaroon}
r = requests.get(url, headers=headers, verify=TLS_PATH)
print(r.json())
```

### Expected behavior

Should return the asset meta.

### Actual behavior

Instead returns: 
`{'code': 2, 'message': 'asset ID must be 32 bytes', 'details': []}`

Thanks much for any insight into solving this.

guggero:
Can you give us an example of a the asset ID you're using?

inviti8:
Here is the output I get from Listing the minted assets via REST.  Seems like the asset_id is 64 bytes:

` "asset_id": "1b793395880c56c83b605c2af7292c95992b522c3bcdc5fd8c00b6c9b89aa95c"`

Here's the full return from list assets:

```
{
  "assets": [
    {
      "version": 0,
      "asset_genesis": {
        "genesis_point": "f60d0015a68fc01a1c6e098777e9523d49726e975cc5307be46262f519236bd6:0",
        "name": "MyCollectible",
        "meta_hash": "4fd760a5fbfcd78a904e00e50eb0ee03c1d2f9a3b245c88f1f34de93ca8fafbb",
        "asset_id": "1b793395880c56c83b605c2af7292c95992b522c3bcdc5fd8c00b6c9b89aa95c",
        "output_index": 0,
        "version": 0
      },
      "asset_type": "COLLECTIBLE",
      "amount": "1",
[...]

PR #445, opened 2023-08-09
Fix REST parameter issue in GET call

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/428.


jharveyb (approved):
LGTM! I imagine we may have some other fields that need this treatment, could make this more generic later on.

Roasbeef (approved):
LGTM 🐊

Issue #374, opened 2023-06-22
tapdb: remove write lock for RegisterIssuance method 


Roasbeef:
Continued from this comment thread here: https://github.com/lightninglabs/taproot-assets/pull/325/commits/b190d32087a3d579fbfca4cd06548d59d0b46694#r1238780637

Right now we have a write lock in that area, as with high concurrency (for sqlite at least), we end up hitting the retry timeout very quickly. This works and is an ok stop gap, but doesn't fundamental address the issue, as with further concurrency elsewhere (eg: making a buncha addrs). So IMO, we want to address the actual issue here to ensure we can handle a high concurrent write load. 

I think we have a few options here: 
  * Use a minimum back off, and also make it exponential: https://github.com/lightninglabs/taproot-assets/blob/main/tapdb/interfaces.go#L101-L106
     * Right now it can end up waiting for effectively 0 ms (so no time at all). Instead we want a min setting, and for it to back off exponentially up until a period of time. 
  * Tune the timeout, and also the number of retries. Right now we use a max retry setting of 10 times, which means it can wait _worst case_ for 500 ms. I think in the local setting this can get hit quickly too as we'd have hundreds of goroutines trying to insert here. 
  * Properly batch the proof insertion. In the planter, we add the proofs one by one, but instead we can make a single transaction and insert it all at once. For the RPC interface, we can also allow many proofs to be sent over the wire.

PR #449, opened 2023-08-11
tapdb: allow batch inserting proofs when syncing universe, remove write lock for RegisterIssuance

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/374.

Opening as draft, want to address the following TODOs before full review:
 - ~Code cleanup~
 - Unit tests for batch inserts
 - ~Allow federation envoy to also support batch inserting, will speed up local insertion~


tapdb/interfaces.go
```diff
@@ -91,23 +99,52 @@ type BatchedQuerier interface {
...
+	// For each subsequent delay, we double the initial delay. This still
+	// gives us a somewhat random delay, but it still increases with each
+	// attempt. If we double something n times, that's the same as
+	// multiplying the value with 2^n. We limit the power to 32 to avoid
+	// overflows.
+	factor := time.Duration(math.Pow(2, math.Min(float64(attempt), 32)))
+	actualDelay := initialDelay * factor
+
+	// Cap the delay at the maximum configured value.
+	if actualDelay > t.maxRetryDelay {
```
Roasbeef, line 143: 👍

tapdb/interfaces.go
```diff
@@ -91,23 +99,52 @@ type BatchedQuerier interface {
...
-func (t *txExecutorOptions) randRetryDelay() time.Duration {
-	return time.Duration(prand.Int63n(int64(t.retryDelay))) //nolint:gosec
+// randIncreasingRetryDelay returns a random retry delay between -50% and +50%
+// of the configured delay that is doubled for each attempt and capped at a max
+// value.
+func (t *txExecutorOptions) randIncreasingRetryDelay(
+	attempt int) time.Duration {
+
+	halfDelay := t.initialRetryDelay / 2
+	randDelay := prand.Int63n(int64(t.initialRetryDelay)) //nolint:gosec
```
Roasbeef, line 124: We'll also want to lift this version back to lnd @positiveblue, see that other comment about unifying the sqlite code as well.

tapdb/interfaces.go
```diff
@@ -91,23 +99,52 @@ type BatchedQuerier interface {
...
 }
 
-// randRetryDelay returns a random retry delay between 0 and the configured max
-// delay.
-func (t *txExecutorOptions) randRetryDelay() time.Duration {
-	return time.Duration(prand.Int63n(int64(t.retryDelay))) //nolint:gosec
+// randIncreasingRetryDelay returns a random retry delay between -50% and +50%
+// of the configured delay that is doubled for each attempt and capped at a max
+// value.
+func (t *txExecutorOptions) randIncreasingRetryDelay(
```
Roasbeef, line 120: IMO fine to keep the existing name, a bit shorter, but this works too.

tapdb/universe.go
```diff
@@ -244,23 +244,17 @@ func upsertAssetGen(ctx context.Context, db UpsertAssetStore,
 		}
 	}
 
-	genProof := &proof.Proof{}
-	err = genProof.Decode(bytes.NewReader(genesisProof))
```
Roasbeef, line 248: This is a breaking change, the reason why the raw leaf is used is to decouple it from the file format which includes the prefix, length information, etc. As designed, one can iterate backwards through the universe leaves for a given asset lineage, then assemble all those into proof files as is to verify. With this change, they'd need to strip out all the version/prefix information before they were able to do that. 


Otherwise, let's say we update the disk format in the future (we add an MMR commitment or something like that), would we also go to attempt to update all the serialized state transitions as well? 

I get the optimization motivation here though, perhaps we can memoize the encoding? So you call `Encode` and it returns an existing value instead of doing it over and over again. 

In case we do decide to go this direction, we'd also need to update the BIPs.

universe/base.go
```diff
@@ -165,7 +165,46 @@ func (a *MintingArchive) RegisterIssuance(ctx context.Context, id Identifier,
 	// it as a file first as that's what the expected wants.
 	//
 	// TODO(roasbeef): add option to skip proof verification?
-	assetSnapshot, err := leaf.GenesisProof.Verify(
+	assetSnapshot, err := a.verifyIssuanceProof(ctx, &IssuanceItem{
+		ID:   id,
+		Key:  key,
+		Leaf: leaf,
```
Roasbeef, line 172: Meta reveal never actually set in the struct?

universe/base.go
```diff
@@ -249,6 +250,49 @@ func (a *MintingArchive) verifyIssuanceProof(ctx context.Context,
...
+// the proofs within the batch have already been checked that they don't yet
+// exist in the local database.
+func (a *MintingArchive) RegisterNewIssuanceBatch(ctx context.Context,
+	items []*IssuanceItem) error {
+
+	log.Infof("Verifying %d new proofs for insertion into Universe",
+		len(items))
+
+	for idx := range items {
+		assetSnapshot, err := a.verifyIssuanceProof(ctx, items[idx])
```
Roasbeef, line 264: Can verify them in parallel (using a similar `Collect` pattern) here.

universe/interface.go
```diff
@@ -283,6 +288,19 @@ type IssuanceItem struct {
 	MetaReveal *proof.MetaReveal
 }
 
+// BatchRegistrar is an interface that allows a caller to register a batch of
+// issuance items within a base universe.
+type BatchRegistrar interface {
```
Roasbeef, line 293: I think we can also use this in the caretaker, so at `BatchStateConfirmed` when we go to insert all the items into the local tree.

universe/syncer.go
```diff
@@ -195,29 +221,43 @@ func (s *SimpleSyncer) syncRoot(ctx context.Context, remoteRoot BaseRoot,
...
-			// TODO(roasbeef): this is actually giving a lagging
-			// proof for each of them
-			_, err = s.cfg.LocalRegistrar.RegisterIssuance(
-				ctx, uniID, key, leafProof.Leaf,
-			)
-			if err != nil {
-				return fmt.Errorf("unable to register "+
-					"issuance proof: %w", err)
+			mutex.Lock()
+			defer mutex.Unlock()
```
Roasbeef, line 231: I think if we kept the channel approach, then the caller below (that would normally to `Collect`) on the leaves can handle chunking+submitting the batch?

tapgarden/caretaker.go
```diff
@@ -934,12 +965,29 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
...
+				if len(issuanceBatch) < batchSize &&
+					idx+1 != len(committedAssets) {
+
+					continue
+				}
+
+				// We have a full batch or are at the latest
+				// asset, so we'll submit the batch to the
+				// universe.
+				err = submitBatch(idx+1, len(committedAssets))
```
Roasbeef, line 988: 👍

tapgarden/caretaker.go
```diff
@@ -934,12 +965,29 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
...
 				if err != nil {
-					return 0, fmt.Errorf("unable to "+
-						"register issuance: %v", err)
+					return 0, err
```
Roasbeef, line 990: I think this also needs to handle indepmotence? May work already, but the scenario is if we have 1k assets to insert, only do 500 of them (the chunk size) then crash. This may return an error that we'll need to be able to gracefully handle. I think it'll just return teh same proof it already has though.

Roasbeef (changes requested):

tapdb/universe.go
Roasbeef: Nvm, chatted offline, this is just about moving from a blob to the struct.

universe/base.go
guggero: Yeah, I guess in this case it doesn't make sense to use the `IssuanceItem` struct, since the meta reveal is only known _after_ verifying the proof. Reverted to individual parameters.

tapdb/interfaces.go
guggero: Yeah, unnecessarily long, reverted.

universe/base.go
guggero: Yeah, seems simple enough. Though on my machine this only took around 105ms for 100 proofs with the old code anyway. But this seems to have brought it down to around 13ms per 100 proofs :exploding_head:

universe/interface.go
guggero: Definitely. We do that in a later commit.

universe/syncer.go
guggero: Yes, that makes sense. Changed the code here and in the caretaker to use a channel approach.

tapgarden/caretaker.go
guggero: Hmm, good question... What do we currently do if the `BatchStateConfirmed` state fails? I assume we retry on next startup and would try to re-insert the proofs, is that what you mean?
I think since we only have upserts in the `universeRegisterIssuance` function, it should already work?

guggero:
I was able to complete a full asset mint with 10k assets with Postgres on my local machine today :100: 

Here are some statistics of the run:
```
Total time elapsed:			5,516s (1h 32m)
Creating 10k seedlings:			21s
Confirming batch:			84s (1m 24s)
Inserting 10k proof (local universe):	2,031s

<start second node>
Creating leaf diff:			300ms
Fetching 10k leaves:			2s
Inserting 10k leaves (syncer):		3,337s (55m)
```

guggero:
Still want to add some additional unit tests to the database around batched insertion, but the PR should otherwise be ready for re-review.

tapgarden/caretaker.go
Roasbeef: Yeah, I think we'll just retry on startup, then the `upserts` and up with us doing nothing. Then we also have the logic that checks to see if an item already exists in the tree first as well. So I think we're fine after all.

tapdb/multiverse.go
```diff
@@ -273,3 +273,70 @@ func (b *BaseMultiverse) RegisterIssuance(ctx context.Context,
...
 }
+
+// RegisterBatchIssuance inserts a new minting leaf batch within the multiverse
+// tree and the universe tree that corresponds to the given base key(s).
+func (b *BaseMultiverse) RegisterBatchIssuance(ctx context.Context,
+	items []*universe.IssuanceItem) error {
+
+	insertProof := func(item *universe.IssuanceItem,
+		dbTx BaseMultiverseStore) error {
+
```
Roasbeef, line 284: I think re that question about idempotence, _this_ is where (or we'd add the initial check to skip if already present?

universe/syncer.go
```diff
@@ -195,27 +195,57 @@ func (s *SimpleSyncer) syncRoot(ctx context.Context, remoteRoot BaseRoot,
...
+		return err
+	}
+	close(fetchedLeaves)
 
-			// TODO(roasbeef): this is actually giving a lagging
-			// proof for each of them
-			_, err = s.cfg.LocalRegistrar.RegisterIssuance(
-				ctx, uniID, key, leafProof.Leaf,
+	var (
+		batchSize     = 200
```
Roasbeef, line 239: How as this arrived at? Some basic trial and iteration, or a number that looked good enough?

tapgarden/caretaker.go
```diff
@@ -888,65 +898,105 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
...
-				_, err = b.cfg.Universe.RegisterIssuance(
-					ctx, uniID, baseKey, mintingLeaf,
-				)
-				if err != nil {
-					return 0, fmt.Errorf("unable to "+
-						"register issuance: %v", err)
-				}
+			// local base universe. We skip this step if there is
+			// no universe configured.
+			if b.cfg.Universe == nil {
```
Roasbeef, line 903: 👍  nice to short circuit and avoid the extra indentation

tapgarden/caretaker.go
```diff
@@ -888,65 +898,105 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
...
+				ID:   uniID,
+				Key:  baseKey,
+				Leaf: mintingLeaf,
+			}
+		}
+		close(universeItems)
+
+		// Batch insert the new proofs into the local universe, if we
+		// have one configured.
+		if b.cfg.Universe != nil {
```
Roasbeef, line 963: If we don't have a universe configured, then we also don't need `universeItems` either. Though I think this is really only the case for unit tests (avoiding a nil pointer painc).

universe/syncer.go
```diff
@@ -52,6 +61,23 @@ func NewSimpleSyncer(cfg SimpleSyncCfg) *SimpleSyncer {
 func (s *SimpleSyncer) executeSync(ctx context.Context, diffEngine DiffEngine,
 	syncType SyncType, idsToSync []Identifier) ([]AssetSyncDiff, error) {
 
+	// Prevent the syncer from running twice. We also don't want to block on
+	// acquiring the mutex, so we use a secondary flag that is guarded by a
+	// mutex instead.
+	s.isSyncingMtx.Lock()
+	if s.isSyncing {
```
Roasbeef, line 68: When would this be called concurrently? Just users trying to sync multiple times on the CLI/RPC?

universe/interface.go
```diff
@@ -265,6 +265,23 @@ type Registrar interface {
...
+	ID Identifier
+
+	// Key is the base key that the leaf is or will be stored at.
+	Key BaseKey
+
+	// Leaf is the minting leaf that was created.
+	Leaf *MintingLeaf
+
+	// MetaReveal is the meta reveal that was created.
+	MetaReveal *proof.MetaReveal
```
Roasbeef, line 287: Still here in latest diff.

universe/syncer.go
guggero: I should've mentioned this somewhere, yes. I did a couple of trial runs with 50, 100, 200 and 500 on my local machine with Postgres and this seemed to be the sweet spot. Probably heavily dependent on the DB, disk speed, latency and so on. But it also seemed like a reasonable number to avoid spamming the logs with too many batch update messages.

universe/syncer.go
guggero: This is to avoid the timed universe sync to run while we're still syncing from an RPC sync request. I noticed this while running the hour-long 10k mint where we spin up a secondary node and sync for almost 40 minutes, then the timed sync actually caused the whole test to fail because it interfered with the already running sync.

universe/interface.go
guggero: It is being used [here](https://github.com/lightninglabs/taproot-assets/blob/082a08b6a04ffc58bc46f94aec817648f884031a/universe/base.go#L276). It just wasn't set for the call to `verifyIssuanceProof` where I unrolled the parameters again.

universe/syncer.go
jharveyb: May be useful to have this as part of `SimpleSyncCfg` instead? Ditto for the other place we have a `batchSize` variable.

tapgarden/caretaker.go
```diff
@@ -888,65 +898,105 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
...
+				GenesisProof: mintingProof,
+				Amt:          newAsset.Amount,
+			}
+			universeItems <- &universe.IssuanceItem{
+				ID:   uniID,
+				Key:  baseKey,
+				Leaf: mintingLeaf,
+			}
+		}
+		close(universeItems)
```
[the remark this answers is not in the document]
jharveyb: Agree, i think `CollectBatch()` needs no change for that to work. Related, not streaming items means duplicating all proofs for the batch in `universeItems` before writing, which could be a fair bit of memory.

Could drop in a TODO but leave for a future PR.

universe/syncer.go
jharveyb: Feels like this construct could be replaced with an `atomic.Bool`?

make/testing_flags.mk
```diff
@@ -35,6 +35,11 @@ ifneq ($(nodelete),)
 ITEST_FLAGS += -nodelete
 endif
 
+# Run the optional tests.
+ifneq ($(optional),)
```
jharveyb, line 39: 👍🏽

universe/base.go
```diff
@@ -223,34 +246,64 @@ func (a *MintingArchive) RegisterIssuance(ctx context.Context, id Identifier,
...
-	issuanceProof, err := a.cfg.Multiverse.RegisterIssuance(
-		ctx, id, key, leaf, assetSnapshot.MetaReveal,
+	return assetSnapshot, nil
+}
+
+// RegisterNewIssuanceBatch inserts a batch of new minting leaves within the
+// target universe tree (based on the ID), stored at the base key(s). We assume
+// the proofs within the batch have already been checked that they don't yet
+// exist in the local database.
+func (a *MintingArchive) RegisterNewIssuanceBatch(ctx context.Context,
```
jharveyb, line 261: In `RegisterIssuance` we attempted to fetch each proof before insertion, and that's dropped in this batched version. IIUC the actual registration is idempotent, so the main consequence of not having that fetching here is re-verifying those proofs that may already be inserted?

jharveyb (commented):
Didn't yet bench locally but this was a great read! Cool to see `fn` used to great effect.

Confirmed no regression with SQLite at batch size 1k. Left most of my comments on existing threads.

Roasbeef:
Needs a rebase!

universe/syncer.go
guggero: Ah yes, I wasn't aware of the `CompareAndSwap` method, which does exactly what we want.

universe/base.go
guggero: I would say the main difference is that `RegisterIssuance` is also the method that's being called by the RPC when a proof is sent to an universe. And we probably do want to have the fetch first to make sure we don't start an "expensive" write operation just to end up in an upsert that doesn't really change any data. That would be a DoS vector.
But since we _know_ all the proofs in a batch of leaves we are syncing from another universe, we kind of already know those should all be new. And `RegisterNewIssuanceBatch` isn't currently hooked up to an RPC so it's not at risk of being a DoS vector.

jharveyb (approved):
LGTM! 🚀

universe/syncer.go
```diff
@@ -165,9 +169,25 @@ func (s *SimpleSyncer) syncRoot(ctx context.Context, remoteRoot BaseRoot,
...
+	// local registrar as they're fetched.
+	var (
+		fetchedLeaves = make(chan *IssuanceItem, len(keysToFetch))
+		newLeafProofs []*MintingLeaf
+		batchSyncEG   errgroup.Group
+	)
+
+	// We use an error group to simply the error handling of a goroutine.
+	batchSyncEG.Go(func() error {
+		newLeafProofs, err = s.batchStreamNewItems(
```
Roasbeef, line 199: Very cool pattern! I think we might need just one more light layer of abstraction to make it a bit more readable, but I dig the structured concurrency here.

Roasbeef (approved):
LGTM 🐙

PR #450, opened 2023-08-15
address: add and populate proof courier addr field in Tap address

ffranr:
* Add mandatory TLV field.
* Ensure encoding/decoding unit tests pass.
* Add structure to populate field from tapcfg config.

This PR replaces https://github.com/lightninglabs/taproot-assets/pull/432 
I've moved the branch to `lightninglabs/taproot-assets` from `ffranr/taproot-assets`.


ffranr:
@guggero any idea why the test vector check is breaking? How does the test work? Is it trying to reproduce the test vectors exactly?

address/book.go
```diff
@@ -138,6 +138,10 @@ type BookConfig struct {
 	// Chain points to the chain the address.Book is active on.
 	Chain ChainParams
 
+	// ProofCourierAddr is the address of the proof courier service to use
+	// for proof transfer.
+	ProofCourierAddr ProofCourierAddr
```
guggero, line 143: Should we call this `FallbackProofCourierAddr` since we would only use this for addresses that don't have one set?

address/book.go
```diff
@@ -327,15 +332,31 @@ func (b *Book) NextScriptKey(ctx context.Context,
...
 
-	return b.cfg.Store.QueryAddrs(ctx, params)
+	addrs, err := b.cfg.Store.QueryAddrs(ctx, params)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add the proof courier address to all Tap addresses.
+	for i := range addrs {
+		addrs[i].Tap.ProofCourierAddr = b.cfg.ProofCourierAddr
```
guggero, line 342: I think we should check if the courier address is empty before overwriting it? Otherwise you would never see a custom address if you set one explicitly.

guggero: Same in all calls below in this file.

address/encoding.go
```diff
@@ -40,3 +40,26 @@ func compressedPubKeyDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error
 		val, "*btcec.PublicKey", l, btcec.PubKeyBytesLenCompressed,
 	)
 }
+
+func proofCourierAddrEncoder(w io.Writer, val any, buf *[8]byte) error {
+	if t, ok := val.(*ProofCourierAddr); ok {
+		addrBytes := []byte(*t)
+		return tlv.EVarBytes(w, &addrBytes, buf)
+	}
+	return tlv.NewTypeForEncodingErr(val, "*address.ProofCourierAddr")
```
guggero, line 49: nit: can remove the `address.` part in the name to make it a bit shorter (that's just for eventual error messages).

tapdb/addrs_test.go
```diff
@@ -168,6 +171,11 @@ func TestAddressInsertion(t *testing.T) {
 	dbAddrs, err := addrBook.QueryAddrs(ctx, address.QueryParams{})
 	require.NoError(t, err)
 
+	// Set proof courier addresses on dbAddrs.
+	for i := range dbAddrs {
```
guggero, line 175: So if we explicitly store a proof courier address, this shouldn't be necessary. So nowhere in this file I would expect the `ProofCourierAddr` member of an address struct being overwritten. Instead I think we need to update the `addrs` and add the address there as a new column.

address/book.go
```diff
@@ -443,3 +473,7 @@ func (b *Book) RemoveSubscriber(
 
 	return nil
 }
+
+func (b *Book) ProofCourierAddr() ProofCourierAddr {
```
guggero, line 477: nit: missing Godoc. And do we really need this?

address/book.go
guggero: Ah, we're missing the persistence to the DB in this PR, that's why this was added. Actually I think the diff would be smaller by adding the new DB field.

tapcfg/config.go
```diff
@@ -254,7 +258,7 @@ type Config struct {
 	BatchMintingInterval time.Duration `long:"batch-minting-interval" description:"A duration (1m, 2h, etc) that governs how frequently pending assets are gather into a batch to be minted."`
 
 	// The following options are used to configure the proof courier.
-	ProofCourierMode string                    `long:"proofcouriermode" choice:"hashmail" description:"Type of proof courier to use."`
+	ProofCourierAddr *proof.CourierAddr        `group:"proofcourier" long:"proofcourier" description:"Proof courier service address."`
```
guggero, line 261: I think this needs a `group:"proofcourier"` tag. And to make it look similar to the other grouped configs we have, maybe remove the `long:""` and `description:""` tags?

guggero (commented):
Very nice! Looks great, almost ready.
I'm not sure if the intention was to add the database persistence in a separate PR. I'm totally fine with that and won't block the PR on it, but I think the actual diff might turn out even smaller if we add the persistence (then we don't need to manually set the courier in several places). Up to you!

address/book.go
ffranr: I think `ProofCourierAddr` is the primary source for a courier addr when generating new addresses (using `Book.NewAddress`). In that sense it will not be a fallback.

address/book.go
ffranr: This is only used in a unit test where we need a new random tap address to have the same proof courier address as that found in the address book. Here:
https://github.com/ffranr/taproot-assets/blob/add-proof-courier-to-addr/tapgarden/custodian_test.go#L168-L171

tapcfg/config.go
ffranr: > I think this [doesn't] need a...

I think you're saying I should remove `group:"proofcourier"`?

And the same for `HashMailCourier` just below it?

tapdb/addrs.go
```diff
@@ -311,7 +311,7 @@ func (t *TapAddressBook) QueryAddrs(ctx context.Context,
 		dbAddrs, err := db.FetchAddrs(ctx, AddrQuery{
 			CreatedAfter:  params.CreatedAfter.UTC(),
 			CreatedBefore: params.CreatedBefore.UTC(),
-			NumOffset:     int32(params.Offset),
+			NumOffset:     params.Offset,
```
Roasbeef, line 314: 👍

address/address.go
```diff
@@ -78,6 +78,12 @@ const (
 	TapScriptVersion uint8 = 0
 )
 
+// ProofCourierAddr is the address of the proof courier that will be used to
+// distribute related proofs for this address.
+//
+// NOTE: See the proof package for more details.
+type ProofCourierAddr []byte
```
Roasbeef, line 85: If we're going with the URI approach, then should this actually be a `string` for the purposes of internal representation? 

We may also want to consider the tradeoffs of the custom URI approach vs either the type for each courier type appraoch or the type prefix approach. 

Nice thing is for the URI approach, we _should_ be able to re-use this method from the stdlib: https://pkg.go.dev/net/url#ParseRequestURI. One other thing to consider is how we track the set of URIs we know about, and/or mark those as required vs not.

address/book.go
Roasbeef: Also perhaps we may want to add several of such instances in the future? Given that maybe we start to add the nostr option or w/e, but not all clients support that. We should think through what that update process looks like and how clients/wallet would handle such a flow.

address/book.go
Roasbeef: Yeah I think the addrs should have this on disk. Then this way we avoid "downgrades" due to someone modifying the default commit.

address/mock.go
```diff
@@ -14,8 +14,15 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// RandProofCourierAddr returns a proof courier address with fields populated
+// with valid but random values.
+func RandProofCourierAddr() ProofCourierAddr {
+	return ProofCourierAddr("hashmail://rand.hashmail.proof.courier:443")
```
Roasbeef, line 20: I think we'd want to make the addr string into a richer type so it can handle encoding/decoding these strings itself. See above where I linked to a stdlib function that should prove to be useful.

address/testdata/address_tlv_encoding_generated.json
```diff
@@ -23,9 +24,10 @@
         "script_key": "02a0afeb165f0ec36880b68e0baabd9ad9c62fd1a69aa998bc30e9a346202e078f",
         "internal_key": "02a0afeb165f0ec36880b68e0baabd9ad9c62fd1a69aa998bc30e9a346202e078f",
         "tapscript_sibling": "",
-        "amount": 3510942875414458836
+        "amount": 3510942875414458836,
+        "proof_courier_addr": "hashmail://rand.hashmail.proof.courier:443"
       },
-      "expected": "tapsb1qqqsqq3q3t94z4pxzsja6cflmgvaacq22854c6fpm790gzzmc358q6u5dlkqgggz5zh7k9jlpmpk3q9k3c9640v6m8rzl5dxn25e30psax35vgpwq78svggz5zh7k9jlpmpk3q9k3c9640v6m8rzl5dxn25e30psax35vgpwq78ssz0lxzu4luvrc3cagnze50n",
+      "expected": "tapsb1qqqsqq3q3t94z4pxzsja6cflmgvaacq22854c6fpm790gzzmc358q6u5dlkqgggz5zh7k9jlpmpk3q9k3c9640v6m8rzl5dxn25e30psax35vgpwq78svggz5zh7k9jlpmpk3q9k3c9640v6m8rzl5dxn25e30psax35vgpwq78ssz0lxzu4luvrc3cagz32dpshx6rdv95kcw309aexzmny9e5xzumgd4skjmpwwpex7mmx9e3k7atjd9jhyw35xses89l6tn",
```
Roasbeef, line 30: May want to look into some sort of compression here for the addr eventually....

I think could be another arg for a more compact representation, though the human readability (at least when decoded is nice too).

Roasbeef (changes requested):

proof/courier.go
```diff
@@ -30,9 +32,48 @@ const (
...
+	default:
+		return "disabled_courier"
+	}
+}
+
+// CourierAddr is the address of the proof courier service that will be used to
+// transfer proofs related to the asset send process.
+type CourierAddr struct {
+	// Protocol is the protocol of the proof courier.
+	Protocol CourierType `long:"protocol" choice:"hashmail" description:"Proof courier service protocol."`
```
Roasbeef, line 55: This is meant to be a global value/setting? I would think instead we have the initial config, then the user can override that whenever they make an address.

proof/courier.go
```diff
@@ -30,9 +32,48 @@ const (
...
+	Port uint16 `long:"port" description:"Proof courier service port."`
+
+	// Hostname is the hostname of the proof courier service.
+	Hostname string `long:"hostname" description:"Proof courier service hostname."`
+}
+
+// URI returns the Uniform Resource Identifier string representation of the
+// proof courier address.
+func (a CourierAddr) URI() string {
+	return fmt.Sprintf(
```
Roasbeef, line 67: See the URI funcs linked.

Roasbeef (changes requested):

ffranr:
The two things I have left to do here (as far as I can tell):

1. Proof courier addr set as CLI arg tap address generation.
2. Proof courier addr uses [`URL`](https://pkg.go.dev/net/url) type (from `url` package) instead of `[]byte` in `address` package.

tapcfg/config.go
guggero: Sorry, I meant it needs a `namespace:""` tag, same as the hash mail courier below (or `Lnd`).

address/book.go
guggero: You can just use a random address in that code. Again, this should go away once we persist the proof courier address. Then the address book doesn't need a source of the default courier as that will all be handled on the RPC level (either the user specifies a specific courier URI to use for an address they create or we fall back to the default one configured in the daemon).

proof/courier.go
ffranr: This effectively allows us to set a default courier address when starting the daemon. Then, if a courier address isn't specified when generating a tap address via RPC, we can fall back to this default courier address.

address/address.go
ffranr: I've changed this to a `url.URL` type.

address/book.go
ffranr: Added tracking issue for design task: https://github.com/lightninglabs/taproot-assets/issues/455

rpcserver.go
```diff
@@ -987,6 +988,23 @@ func (r *rpcServer) NewAddr(ctx context.Context,
 
 	var err error
 
+	// Parse the proof courier address if one was provided, otherwise use
+	// the default specified in the config.
+	var proofCourierAddr *url.URL
+
+	if r.cfg.DefaultProofCourierAddr != nil {
```
guggero, line 995: When would this be `nil`? I think we should be able to rely on this being set, as we have the fallback parsing in the `server.go` file. So we can just do: `proofCourierAddr := r.cfg.DefaultProofCourierAddr`.

taprpc/taprootassets.proto
```diff
@@ -566,6 +566,9 @@ message Addr {
     transfer assets described in this address.
     */
     bytes taproot_output_key = 9;
+
+    // A proof courier address for use in proof transfer.
```
guggero, line 570: nit: "The proof courier address..."?

guggero (approved):
LGTM :tada:

address/address.go
```diff
@@ -118,14 +119,18 @@ type Tap struct {
 	// assetGen is the receiving asset's genesis metadata which directly
 	// maps to its unique ID within the Taproot Asset protocol.
 	assetGen asset.Genesis
+
+	// ProofCourierAddr is the address of the proof courier that will be
+	// used to distribute related proofs for this address.
+	ProofCourierAddr *url.URL
```
Roasbeef, line 125: Any reason to use a pointer over actual value here?

address/address.go
```diff
@@ -118,14 +119,18 @@ type Tap struct {
...
 }
 
 // New creates an address for receiving a Taproot asset.
 func New(genesis asset.Genesis, groupKey *btcec.PublicKey,
 	groupSig *schnorr.Signature, scriptKey btcec.PublicKey,
 	internalKey btcec.PublicKey, amt uint64,
 	tapscriptSibling *commitment.TapscriptPreimage,
-	net *ChainParams) (*Tap, error) {
+	net *ChainParams, proofCourierAddr *url.URL) (*Tap, error) {
```
Roasbeef, line 133: Non-blocking, but starting to be a lot of values, can maybe split out into a new sub-struct for the params? I think the genesis+group key stuff can use `GroupKeyInfo`, but maybe to wait until we get the v1 asset stuff in.

proof/courier.go
```diff
@@ -21,18 +22,58 @@ import (
...
+func ParseCourierAddr(addr string) (*url.URL, error) {
+	// Parse URI.
+	urlAddr, err := url.ParseRequestURI(addr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid proof courier URI address: %w",
+			err)
+	}
+
+	// Validate port number.
+	if urlAddr.Port() == "" {
```
Roasbeef, line 62: Port may not always be relevant. Instead I think `CourierType` could expose a method used for validating the courier based on its type. May want to spin it into an interface at that point as well. 

Non blocking for now though, but we'll want to remove this eventually to support more flexible courier types.

tapcfg/server.go
```diff
@@ -156,11 +156,36 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
...
+		fallbackHashmailCourierAddr,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse fallback proof "+
+			"courier address: %v", err)
+	}
+
+	// If default proof courier address is set, use it as the default.
+	if cfg.DefaultProofCourierAddr != "" {
+		proofCourierAddr, err = proof.ParseCourierAddr(
```
Roasbeef, line 174: I think would also make sense to eventually enable this to be configurable when one goes to create an address. Similarly, I think it would eventually be useful to support multiple couriers encoded in a given addr. 

I think non-blocking now as we don't need it, but can imagine an itest in the future where two are specified, one is down, but the other can be used for the final transfer.

rpcserver.go
```diff
@@ -987,6 +987,19 @@ func (r *rpcServer) NewAddr(ctx context.Context,
 
 	var err error
 
+	// Parse the proof courier address if one was provided, otherwise use
+	// the default specified in the config.
+	proofCourierAddr := r.cfg.DefaultProofCourierAddr
+	if in.ProofCourierAddr != "" {
```
Roasbeef, line 993: 👍

address/mock.go
```diff
@@ -14,8 +15,20 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// RandProofCourierAddr returns a proof courier address with fields populated
+// with valid but random values.
+func RandProofCourierAddr(t testing.TB) *url.URL {
```
jharveyb, line 20: non-blocking nit: returns a static address? Could be useful to have a rationale or TODO here if we want it to be random later on.

taprpc/taprootassets.proto
```diff
@@ -624,6 +627,12 @@ message NewAddrRequest {
     commitment of the asset.
     */
     bytes tapscript_sibling = 5;
+
+    /*
+    An optional proof courier address for use in proof transfer. If unspecified,
+    the daemon configured default address will be used
```
jharveyb, line 633: nit: missing period.

jharveyb (approved):
LGTM 💯

Roasbeef (approved):
LGTM 🌙

tapcfg/server.go
ffranr: > enable [`DefaultProofCourierAddr`] to be configurable when one goes to create an address

Perhaps via a new RPC argument flag for setting the given address as the new default proof courier address. I'll open an issue for that.

proof/courier.go
ffranr: Good point. Tracking here: https://github.com/lightninglabs/taproot-assets/issues/462

address/address.go
ffranr: No. I will fix this in a future PR. I don't think this field should be a pointer because it is a mandatory field.

address/address.go
ffranr: I'll add a TODO in `New` function doc.

ffranr:
Follow up PR to address the last two reviews: https://github.com/lightninglabs/taproot-assets/pull/463

PR #452, opened 2023-08-16
monitoring: add prometheus metrics collector

### LINKED ISSUE #298: monitoring: add basic prometheus export info
We should add optional an prometheus scraper. Relevant stuff we want to track includes asset balances, universe stat info, etc, etc. 

If we don't want to add this directly to the main daemon, then we can make a small optional sub-command to epxort the information. So basically taromon, similar to `lndmon`.

positiveblue:
Add monitoring package. Currently, the only registered metrics are for the grpc server interface.

We did not follow the same route as lnd, where there is a external process collecting and exposing most of the metrics (`lndmon`). It will be tapd itself the one that will register and expose the metrics.

How to test this PR?
- Make sure the binary is build with the `monitoring` tag (`dev` includes it by default)
- Make sure `prometheus.active` is set to true in the config/cli tapd params
- Hit the prometheus server port (`8989` by default)

Fixes #298


monitoring/prometheus.go
```diff
@@ -0,0 +1,125 @@
...
+// launches the HTTP server that Prometheus will hit to scrape our metrics.
+func (p *PrometheusExporter) Start() error {
+	// If we're not active, then there's nothing more to do.
+	if !p.config.Active {
+		return nil
+	}
+
+	// Before we register the rpc servers, we'll ensure that the collector
+	// will export latency metrics for the histogram.
+	grpc_prometheus.EnableHandlingTimeHistogram()
```
Roasbeef, line 45: I think let's also put this behind a flag similar to `lnd`? It can end up adding a lot of metrics, which can end up bloating prom a bit.

monitoring/prometheus.go
```diff
@@ -0,0 +1,125 @@
+package monitoring
+
+import (
+	"net/http"
+	"sync"
+
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
```
Roasbeef, line 7: So this repo is now ready only: https://github.com/grpc-ecosystem/go-grpc-prometheus#deprecated-go-grpc-interceptors-for-prometheus-monitoring

We should update import paths to keep with the maintained version.

Roasbeef (changes requested):

monitoring/prometheus.go
positiveblue: right, even with the counters we have quite a bit of series. With the new library we can disable them if need it in the future too.

monitoring/prometheus.go
positiveblue: Updated to `github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus` :+1:

make/testing_flags.mk
```diff
@@ -1,5 +1,6 @@
-DEV_TAGS = dev
+DEV_TAGS = dev $(MONITORING_TAGS)
 RPC_TAGS = autopilotrpc chainrpc invoicesrpc peersrpc routerrpc signrpc verrpc walletrpc watchtowerrpc wtclientrpc
+MONITORING_TAGS = monitoring
```
guggero, line 3: nit: move this to the top to make the order a bit more intuitive?

monitoring/prometheus.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Finally, we'll launch the HTTP server that Prometheus will use to
+	// scape our metrics.
+	go func() {
+		http.Handle(
+			"/metrics",
+			promhttp.HandlerFor(reg, promhttp.HandlerOpts{}),
+		)
+
+		// TODO(positiveblue): add server timeouts?
```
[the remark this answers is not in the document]
guggero: Actually, while testing this, I noticed that the gRPC stats weren't exported. We need to use this code to make everything work as expected:

```
// Start registers all relevant metrics with the Prometheus library, then
// launches the HTTP server that Prometheus will hit to scrape our metrics.
func (p *PrometheusExporter) Start() error {
	// If we're not active, then there's nothing more to do.
	if !p.config.Active {
		return nil
	}

	// Make sure that the server metrics has been created.
	if serverMetrics == nil {
		return fmt.Errorf("server metrics not set")
	}

	reg := prometheus.NewRegistry()
	reg.MustRegister(collectors.NewProcessCollector(
		collectors.ProcessCollectorOpts{},
	))
	reg.MustRegister(collectors.NewGoCollector())
	reg.MustRegister(serverMetrics)

	// Make ensure that all metrics exist when collecting and querying.
	serverMetrics.InitializeMetrics(p.config.RPCServer)

	// Next, we'll attempt to register all our metrics. If we fail to
	// register ANY metric, then we'll fail all together.
	if err := p.registerMetrics(); err != nil {
		return err
	}

	// Finally, we'll launch the HTTP server that Prometheus will use to
	// scape our metrics.
	go func() {
		promMux := http.NewServeMux()
		promMux.Handle("/metrics", promhttp.HandlerFor(
			reg, promhttp.HandlerOpts{
				EnableOpenMetrics:   true,
				MaxRequestsInFlight: 1,
			}),
		)

		log.Infof("Prometheus listening on %v", p.config.ListenAddr)

		pprofServer := &http.Server{
			Addr:              p.config.ListenAddr,
			Handler:           promMux,
			ReadHeaderTimeout: 5 * time.Second,
		}

		// Start the prometheus server.
		err := pprofServer.ListenAndServe()
		if err != nil && !errors.Is(err, http.ErrServerClosed) {
			log.Errorf("Serving prometheus got err: %v", err)
		}
	}()

	return nil
}
```

guggero (commented):
Very nice!
Tested everything and found a few small issues, once those are fixed this is good to go!

guggero (approved):
tACK, LGTM :tada:

Roasbeef (approved):
LGTM 🦜

Roasbeef:
I think the remaining follow up will be to start to add the relevant gauges and counters. I listed some candidates in the OP issue, but there're also other relevant base level items to export.

Issue #433, opened 2023-08-08
tapd: load test binary place holder


Roasbeef:
This issue is a place holder to track some of the external progress re refactorings towards load tests (productionizing the 100k minting test as an example).

Roasbeef:
The binary is done.

PR #457, opened 2023-08-17
itest: abstract the mint batch stress test 

positiveblue:
In the first commit we make most of the checks used in the itest able to work with tapd/go testing primitives (like rpc clients and `*testing.T` structs) instead of depending on `testHarness`. That will allow us use those methods in tests that do not involve tapd harness but external tapd processes.

In the second commit we inject all the rpc clients and chain functionalities that it needs to operate so it does not depend on our itest suite and can be executed against real instances.

Part of #433

### DIFF
--- a/itest/mint_batch_stress_test.go
+++ b/itest/mint_batch_stress_test.go
@@ -83,7 +83,7 @@ func mintBatchStressTest(t *harnessTest, batchSize int,
 	}
 
 	// Use the first asset of the batch as the asset group anchor.
-	collectibleAnchorReq := copyRequest(&collectibleRequestTemplate)
+	collectibleAnchorReq := CopyRequest(&collectibleRequestTemplate)
 	incrementMintAsset(collectibleAnchorReq.Asset, 0)
 	collectibleAnchorReq.EnableEmission = true
 	batchReqs[0] = collectibleAnchorReq
@@ -91,7 +91,7 @@ func mintBatchStressTest(t *harnessTest, batchSize int,
 	// Generate the rest of the batch, with each asset referencing the group
 	// anchor we created above.
 	for i := 1; i < batchSize; i++ {
-		groupedAsset := copyRequest(&collectibleRequestTemplate)
+		groupedAsset := CopyRequest(&collectibleRequestTemplate)
 		incrementMintAsset(groupedAsset.Asset, i)
 		groupedAsset.Asset.GroupAnchor = collectibleAnchorReq.Asset.Name
 		batchReqs[i] = groupedAsset
@@ -108,7 +108,7 @@ func mintBatchStressTest(t *harnessTest, batchSize int,
 
 	// We can re-derive the group key to verify that the correct asset was
 	// used as the group anchor.
-	collectibleAnchor := verifyGroupAnchor(
+	collectibleAnchor := VerifyGroupAnchor(
 		t.t, mintBatch, collectibleAnchorReq.Asset.Name,
 	)
 	collectGroupKey := collectibleAnchor.AssetGroup.TweakedGroupKey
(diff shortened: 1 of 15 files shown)


itest/mint_batch_stress_test.go
```diff
@@ -30,29 +31,109 @@ var (
...
-	mintBatchStressTest(t, 100, defaultWaitTimeout)
+	timeout := defaultWaitTimeout
+
+	// If we create a second tapd instance and sync the universe state,
+	// the synced tree should match the source tree.
+	bob := setupTapdHarness(
+		t.t, t, t.lndHarness.Bob, nil,
+	)
+	defer func() {
+		require.NoError(t.t, bob.stop(!*noDelete))
```
guggero, line 42: Maybe we should make the `noDelete` package level flag part of the `harnessTest` so we don't need to provide any special build/test flags when running the code externally?

itest/mint_batch_stress_test.go
```diff
@@ -30,29 +31,109 @@ var (
...
+	}()
+
+	mintBatches := func(reqs []*mintrpc.MintAssetRequest) []*taprpc.Asset {
+		return mintAssetsConfirmBatch(
+			t, t.tapd, reqs, withMintingTimeout(timeout),
+		)
+	}
+
+	getImageMetadataBytes := func() []byte {
+		return GetImageMetadataBytes(t.t, ImageMetadataFileName)
```
guggero, line 80: Will we need to inject some special metadata files later on? Or could we just read the bytes here and then pass them into `mintBatchStressTest` directly?

jharveyb: I agree the pattern looks weird now with the simple test cases, but could be useful for injecting different metadata later on for a larger matrix of batch sizes + metadata sizes or types, or randomized metadata.

itest/re-org_test.go
```diff
@@ -85,14 +85,15 @@ func testReOrgMint(t *harnessTest) {
...
-	waitForProofUpdate(t.t, t.tapd, assetList[0], newBlockHeight)
+	WaitForProofUpdate(t.t, t.tapd, assetList[0], newBlockHeight)
 
 	// We now try to validate the issuance proof of the two assets we
 	// minted again. The re-org watcher should have updated the proofs and
 	// pushed them to the proof store. They should be valid now.
 	for idx := range assetList {
 		a := assetList[idx]
-		assertAssetProofs(t.t, t.tapd, a)
+		chainClient := t.tapd.cfg.LndNode.RPC.ChainKit
```
jharveyb, line 95: non-blocking nit: Here and in other loops where we use `chainClient`, we reassign the variable inside the loop instead of before the loop. That seems unnecessary?

Other loops at lines 224, 350, etc.

itest/mint_batch_stress_test.go
```diff
@@ -30,29 +31,109 @@ var (
...
 	imageMetadataBytes, err := hex.DecodeString(
 		strings.Trim(string(imageMetadataHex), "\n"),
 	)
-	require.NoError(t.t, err)
+	require.NoError(t, err)
+
+	return imageMetadataBytes
+}
 
+func mintBatchStressTest(
```
jharveyb, line 131: Is there some plan to make this public later on, and then call it from the new test binary? Or does it need further adjustments.

Alternatively, I'm misunderstanding which parts of the existing itests are going to get re-used, and this would get 'reimplemented' in the standalone binary.

jharveyb (approved):
Looks good! Just nits, excited to see these tests running outside of itest.

itest/re-org_test.go
positiveblue: good catch, your version makes the code more readable too, so added :+1:

itest/mint_batch_stress_test.go
positiveblue: This test is going to be moved to the `perf_test` packages. I think no need to keep in the itest package but not execute it?

jharveyb: makes sense - if it's in `perf_test` I think I'd want the test to still have a Makefile target somehow, or be able to run it locally as I can with the optional tests currently.

guggero (approved):

Issue #441, opened 2023-08-08
proof+tapgarden: populate the `GenesisReveal` and `GroupKeyReveal` fields in minting proofs


jharveyb:
Depends on resolving #440 with PR #381.

To support the changes specified in https://github.com/Roasbeef/bips/pull/33 (more complex group key derivation + reveal genesis and group key information in the minting proof), we need to populate asset issuance and reissuance proofs with `GenesisReveal` and `GroupKeyReveal` data. This reveal data should be populated in the caretaker, stored on disk, and sent to any connected universes (just like any other proof field).

A key way to test this would be to receive the `GenesisReveal` and `GroupKeyReveal` data for an asset from a universe sync with a `tapd` instance that is not the asset minter.

Pessimistically marking this as size Medium - I think it could be Small, and AFAICT right now the code changes should not be particularly complex, but I think it will cross a few packages with non-trivial changes, including the universe logic.

PR #448, opened 2023-08-11
Proof: genesis reveal + and groupkey reveal validation

halseth:
This PR builds on #381, validating the `GroupKeyReveal` and `GenesisReveal` at proof verification time.

Note that this PR doesn't validate whether the group key witness is valid, as this is a more significant change that will be in a follow-up PR.

Partially fixes #394 

TODO:
- [x] Requires #381 
- [ ] Unit tests

PR #458, opened 2023-08-18
proof: Populate genesis and meta reveals in minting proofs

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/441.

Depends on https://github.com/lightninglabs/taproot-assets/pull/448.

Work in progress, things don't fully work yet.


jharveyb:
Ok, this has updated to also populate the new reveal fields in proofs decoded via RPC. This allows us to verify that these new fields are encoded and decoded correctly via the `mint assets` itest.

The protos for proofs will change again in a later PR to include the group witness, and actual verification of these reveal fields will be implemented in #448 .

guggero:
Closing, so @jharveyb can re-open.

PR #459, opened 2023-08-21
Chain porter proof courier service initialised using configurable address

### LINKED ISSUE #446: Asset send/receive procedure makes use of Tap address hashmail proof courier addr
Issue https://github.com/lightninglabs/taproot-assets/issues/434 adds a new TLV field to the Tap address which will hold an address for the target proof courier service.

This issue is satisfied by implementing two related features:
1. populating the new field at the point that the receiver node generates a Tap address (prerequisite to sending an asset)
2. the sending node reads the new field and transfers the proof to the target proof courier.

This issue is only concerned with **hashmail** proof courier services (which we already support). The address added in https://github.com/lightninglabs/taproot-assets/issues/434 should specify the proof courier protocol. For now, we should return an error if it is not a hashmail address. (We will extend with RPC support later.)

ffranr:
This PR includes changes which allow us to persist a proof courier address for use in resuming a pending package's send process.

Fixes #446

### DIFF
--- a/proof/courier.go
+++ b/proof/courier.go
@@ -39,41 +39,6 @@ const (
 	ApertureCourier = "hashmail"
 )
 
-// NewCourierType returns the CourierType that corresponds to the given string.
-func NewCourierType(scheme string) (CourierType, error) {
-	switch scheme {
-	case ApertureCourier:
-		return ApertureCourier, nil
-	}
-
-	return DisabledCourier, fmt.Errorf("unknown courier address "+
-		"protocol: %v", scheme)
-}
-
-func ParseCourierAddr(addr string) (*url.URL, error) {
-	// Parse URI.
-	urlAddr, err := url.ParseRequestURI(addr)
-	if err != nil {
-		return nil, fmt.Errorf("invalid proof courier URI address: %w",
-			err)
-	}
-
-	// Validate port number.
-	if urlAddr.Port() == "" {
-		return nil, fmt.Errorf("proof courier URI address port "+
-			"unspecified: %w", err)
-	}
-
-	// Validate protocol supported.
-	_, err = NewCourierType(urlAddr.Scheme)
-	if err != nil {
-		return nil, fmt.Errorf("invalid proof courier protocol: %w",
-			err)
-	}
-
-	return urlAddr, nil
-}
-
 // CourierHarness interface is an integration testing harness for a proof
 // courier service.
 type CourierHarness interface {
@@ -90,18 +55,147 @@ type CourierHarness interface {
 // receiver can use this to fetch a proof from the sender.
 //
 // TODO(roasbeef): FileSystemCourier, RpcCourier
-type Courier[Addr any] interface {
+type Courier interface {
 	// DeliverProof attempts to delivery a proof to the receiver, using the
 	// information in the Addr type.
-	DeliverProof(context.Context, Addr, *AnnotatedProof) error
+	DeliverProof(context.Context, *AnnotatedProof) error
 
 	// ReceiveProof attempts to obtain a proof as identified by the passed
 	// locator from the source encapsulated within the specified address.
-	ReceiveProof(context.Context, Addr, Locator) (*AnnotatedProof, error)
+	ReceiveProof(context.Context, Locator) (*AnnotatedProof, error)
 
 	// SetSubscribers sets the set of subscribers that will be notified
 	// of proof courier related events.
 	SetSubscribers(map[uint64]*fn.EventReceiver[fn.Event])
+}
+
+// CourierAddr is a fully validated courier address (including protocol specific
+// validation).
+type CourierAddr interface {
+	// Url returns the url.URL representation of the courier address.
+	Url() *url.URL
+
+	// NewCourier generates a new courier service handle.
+	NewCourier(ctx context.Context, cfg *CourierCfg,
+		recipient Recipient) (Courier, error)
+}
+
+// ParseCourierAddrString parses a proof courier address string and returns a
+// protocol specific courier address instance.
+func ParseCourierAddrString(addr string) (CourierAddr, error) {
+	// Parse URI.
+	urlAddr, err := url.ParseRequestURI(addr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid proof courier URI address: %w",
+			err)
+	}
+
+	return ParseCourierAddrUrl(*urlAddr)
+}
+
+// ParseCourierAddrUrl parses a proof courier address url.URL and returns a
+// protocol specific courier address instance.
+func ParseCourierAddrUrl(addr url.URL) (CourierAddr, error) {
+	// Create new courier addr based on URL scheme.
+	switch addr.Scheme {
+	case ApertureCourier:
+		return NewHashMailCourierAddr(addr)
+	}
+
+	return nil, fmt.Errorf("unknown courier address protocol: %v",
+		addr.Scheme)
+}
+
+// HashMailCourierAddr is a hashmail protocol specific implementation of the
+// CourierAddr interface.
+type HashMailCourierAddr struct {
+	addr url.URL
+}
+
+// Url returns the url.URL representation of the hashmail courier address.
+func (h *HashMailCourierAddr) Url() *url.URL {
+	return &h.addr
+}
+
+// NewCourier generates a new courier service handle.
+func (h *HashMailCourierAddr) NewCourier(_ context.Context, cfg *CourierCfg,
+	recipient Recipient) (Courier, error) {
+
+	hashMailCfg := HashMailCourierCfg{
+		TlsCertPath:        cfg.TlsCertPath,
+		ReceiverAckTimeout: cfg.ReceiverAckTimeout,
+		BackoffCfg:         cfg.BackoffCfg,
+	}
+
+	hashMailBox, err := NewHashMailBox(&h.addr, hashMailCfg.TlsCertPath)
+	if err != nil {
+		return nil, fmt.Errorf("unable to make mailbox: %v",
+			err)
+	}
+
+	subscribers := make(
+		map[uint64]*fn.EventReceiver[fn.Event],
+	)
+	return &HashMailCourier{
+		cfg:         &hashMailCfg,
+		recipient:   recipient,
+		mailbox:     hashMailBox,
+		deliveryLog: cfg.DeliveryLog,
+		subscribers: subscribers,
+	}, nil
+}
+
+// NewHashMailCourierAddr generates a new hashmail courier address from a given
+// URL. This function also performs hashmail protocol specific address
+// validation.
+func NewHashMailCourierAddr(addr url.URL) (*HashMailCourierAddr, error) {
+	if addr.Scheme != ApertureCourier {
+		return nil, fmt.Errorf("expected hashmail courier protocol: %v",
+			addr.Scheme)
+	}
+
+	// We expect the port number to be specified for a hashmail service.
+	if addr.Port() == "" {
+		return nil, fmt.Errorf("hashmail proof courier URI address " +
+			"port unspecified")
+	}
+
+	return &HashMailCourierAddr{
+		addr,
+	}, nil
+}
+
+// NewCourier instantiates a new courier service handle given a service URL
+// address.
+func NewCourier(ctx context.Context, addr url.URL, cfg *CourierCfg,
+	recipient Recipient) (Courier, error) {
+
+	courierAddr, err := ParseCourierAddrUrl(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	return courierAddr.NewCourier(ctx, cfg, recipient)
+}
+
+// CourierCfg contains general config parameters applicable to all proof
+// couriers.
+type CourierCfg struct {
+	// TlsCertPath is an optional TLS certificate file path. If unset, then
+	// the system's TLS trust store is used.
+	TlsCertPath string
+
+	// ReceiverAckTimeout is the maximum time we'll wait for the receiver to
+	// acknowledge the proof.
+	ReceiverAckTimeout time.Duration
+
+	// BackoffCfg configures the behaviour of the proof delivery
+	// functionality.
+	BackoffCfg *BackoffCfg
+
+	// DeliveryLog is the log that the courier will use to record the
+	// attempted delivery of proofs to the receiver.
+	DeliveryLog DeliveryLog
 }
 
 // ProofMailbox represents an abstract store-and-forward mailbox that can be
@@ -403,10 +497,14 @@ type BackoffCfg struct {
 	MaxBackoff time.Duration `long:"maxbackoff" description:"The maximum backoff time to wait before retrying to deliver the proof to the receiver."`
 }
 
-// HashMailCourier is an implementation of the Courier interfaces that
+// HashMailCourier is a hashmail proof courier service handle. It implements the
+// Courier interface.
 type HashMailCourier struct {
 	// cfg contains the courier's configuration parameters.
 	cfg *HashMailCourierCfg
+
+	// recipient describes the recipient of the proof.
+	recipient Recipient
 
 	mailbox ProofMailbox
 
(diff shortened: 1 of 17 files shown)


ffranr:
Having spoken with @guggero , I'm going to modify this PR as follows:

The goal is for the proof courier address to survive the chain porter send package log/resume cycle so that it is available when constructing the proof courier service handle.

To accomplish that, instead of adding the proof courier address as a new field in the asset transfer table, I will add a new addr field to vPSBT outputs and corresponding db table.

guggero:
Just to clarify: my intention was to store the full, serialized tap-address in the vPSBT and transfer_output_table, not just the proof courier address. that way we get ourselves some future proofing if we ever add another attribute to a tap-address, then don't need yet another field in vPSBT/transfer-output.
I think that outweighs the downside of making the vPSBT slightly bigger with the somewhat duplicate info. Or do you strongly disagree on this, @Roasbeef?

Roasbeef:
> To accomplish that, instead of adding the proof courier address as a new field in the asset transfer table,

Or we just also store the full address as well which may be useful for record keeping purposes? 

> serialized tap-address in the vPSBT and transfer_output_table, not just the proof courier address. that way we get ourselves some future proofing if we ever add another attribute to a tap-address, then don't need yet another field in vPSBT/transfer-output.

Why's it need to be the vPBST vs just a new column in the row?

ffranr:
Following a call with @Roasbeef , this is the latest direction for this PR:

Where the chain porter is concerned, we no longer add the `Tap` address to `VPacket.VOutput`. Instead, we now carry the addresses in `sendPackage` using a map from output index to `Tap` address. Then, we carry the proof courier address only in transfer output and it's SQL table.

This PR also includes changes to the custodian, which now instantiates a proof courier service using the address that we lookup in our address book.

This PR also changes the `proof.Courier` interface: `proof.Courier[Addr any]` -> `proof.Courier`. However, the `recipient` arg still remains in the `DeliverProof` and `ReceiveProof` methods. My idea there is that we might, in the future, want to connect to a service once and then reuse that service for multiple different `recipient` lookups. So I figured I wouldn't make a courier service handle specific to a `recipient`.

I think the following test breaks between 00:00 and 01:00 https://github.com/lightninglabs/taproot-assets/blob/fad3d845f68ec7fa09eb0c05f5f9c7455358928e/tapdb/universe_stats_test.go#L201C42-L201C42

This PR doesn't include a new `RpcCourier`, I'll add that in another PR. I would prefer not to extend this PR if possible.

This PR should now also close https://github.com/lightninglabs/taproot-assets/issues/462

tapfreighter/chain_porter.go
```diff
@@ -768,16 +768,18 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
...
+				ctx, addrParcel.destAddrs...,
+			)
 		if err != nil {
 			return nil, fmt.Errorf("unable to fund address send: "+
 				"%w", err)
 		}
 
 		currentPkg.VirtualPacket = fundSendRes.VPacket
 		currentPkg.InputCommitments = fundSendRes.InputCommitments
+		currentPkg.OutputIdxToAddr = outputIdxToAddr
```
guggero, line 809: This will work for a send where only the chain porter is involved throughout the process (e.g. from an `AddressParcel`). But it will fail once we do actual PSBT funding (meaning there is no in-memory state in the chain porter between the `SendStateVirtualCommitmentSelect` and the `SendStateVirtualSign` steps, because all state is purely kept in the vPSBT by the caller between the different RPC calls that trigger these steps).

So unless we want to refactor the whole chain porter state machine to become stateful across RPC calls (and with that also across restarts), the only "place" that this information can be stored between calls is the vPSBT itself.

The reason why no integration test currently fails is because we still manually export and import the proof in those tests instead of using a courier (we should probably change that soon as well).

tapfreighter/chain_porter.go
```diff
@@ -55,9 +55,7 @@ type ChainPorterConfig struct {
 	// TODO(roasbeef): replace with proof.Courier in the future/
 	AssetProofs proof.Archiver
 
-	// ProofCourier is used to optionally deliver the final proof to the
-	// user using an asynchronous transport mechanism.
-	ProofCourier proof.Courier
+	ProofCourierCfg *proof.CourierCfg
```
guggero, line 58: nit: missing Godoc.

guggero (commented):
Nice progress on this! Everything around the proof courier interfaces and delivery looks great. Though I don't think we can get away without storing the address in the vPSBT as well (see inline comment) for the full vPSBT interactive cases.

tapfreighter/chain_porter.go
guggero: Discussed this offline and it sounds like we're okay with this for now, since as soon as things are somewhat interactive, the proof transfer is expected to happen outside of the porter/courier as well.

tapfreighter/parcel.go
```diff
@@ -208,7 +208,7 @@ type PreSignedParcel struct {
 	inputCommitment *commitment.TapCommitment
 }
 
-// A compile-time assertion to ensure AddressParcel implements the parcel
+// A compile-time assertion to ensure PreSignedParcel implements the parcel
```
Roasbeef, line 211: 👍

proof/courier.go
```diff
@@ -90,20 +55,123 @@ type CourierHarness interface {
 // receiver can use this to fetch a proof from the sender.
 //
 // TODO(roasbeef): FileSystemCourier, RpcCourier
-type Courier[Addr any] interface {
+type Courier interface {
 	// DeliverProof attempts to delivery a proof to the receiver, using the
 	// information in the Addr type.
-	DeliverProof(context.Context, Addr, *AnnotatedProof) error
+	DeliverProof(context.Context, Recipient, *AnnotatedProof) error
```
Roasbeef, line 61: I don't think we want to hard code the `Recipient` struct here. Instead we can lift that information up into the `Courier` implementation itself. In other words, when the instance is created, it already knows "who" it's sending to, and then later you bind the "what". 

For hash mail "who" is the hash mail addr, and the what is the proof. 

For multi-verses, the "who" is the server+outpoint info, and the what is the proof itself. 

Thinking about it a bit more, for the multi-verse instance, we already have the outpoint, as it's contained in the last state transition in the prof.

proof/courier.go
```diff
@@ -90,20 +55,123 @@ type CourierHarness interface {
...
 }
 
+// CourierAddr is a fully validated courier address (including protocol specific
+// validation).
+type CourierAddr interface {
+	// Url returns the url.URL representation of the courier address.
+	Url() *url.URL
+
+	// NewCourier generates a new courier service handle.
+	NewCourier(ctx context.Context, cfg *CourierCfg) (Courier, error)
```
Roasbeef, line 80: Similarly here, will the config actually be universal? What extra information does an addr need, that isn't in the addr itself to create a courier instance?

proof/courier.go
```diff
@@ -90,20 +55,123 @@ type CourierHarness interface {
...
+// Url returns the url.URL representation of the hashmail courier address.
+func (h *HashMailCourierAddr) Url() *url.URL {
+	return &h.addr
+}
+
+// NewCourier generates a new courier service handle.
+func (h *HashMailCourierAddr) NewCourier(_ context.Context,
+	cfg *CourierCfg) (Courier, error) {
+
+	hashMailBox, err := NewHashMailBox(&h.addr, cfg.HashMailCfg.TlsCertPath)
```
Roasbeef, line 124: Ah so in this case, it tacks on the TLS cert path. 

In practice, how would a user even obtain that in the first place? Imagine the hashmail addr is in the addr, where does it get the cert in the first place? IMO we'd just rely on the normal CA chain for that, and in the case of a `localhost` addr, it can bypass normal cert verification. 

Eliminating this would make the `NewCourier` method more generic, and encodes a minimal set of assumptions.

proof/courier.go
```diff
@@ -90,20 +55,123 @@ type CourierHarness interface {
...
+		subscribers: subscribers,
+	}, nil
+}
+
+// NewHashMailCourierAddr generates a new hashmail courier address from a given
+// URL. This function also performs hashmail protocol specific address
+// validation.
+func NewHashMailCourierAddr(addr url.URL) (*HashMailCourierAddr, error) {
+	// We expect the port number to be specified for a hashmail service.
+	if addr.Port() == "" {
```
Roasbeef, line 146: I think it should also verify aspects such as the `Scheme` (should be hashmail in this case), etc.

proof/courier.go
```diff
@@ -90,20 +55,123 @@ type CourierHarness interface {
...
+		mailbox:     hashMailBox,
+		deliveryLog: cfg.DeliveryLog,
+		subscribers: subscribers,
+	}, nil
+}
+
+// NewHashMailCourierAddr generates a new hashmail courier address from a given
+// URL. This function also performs hashmail protocol specific address
+// validation.
+func NewHashMailCourierAddr(addr url.URL) (*HashMailCourierAddr, error) {
```
Roasbeef, line 144: Any reason to not take a `string` here directly, then have it parse the URL itself?

proof/courier.go
```diff
@@ -90,20 +55,123 @@ type CourierHarness interface {
...
+
+	courierAddr, err := ParseCourierAddrUrl(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	return courierAddr.NewCourier(ctx, cfg)
+}
+
+type CourierCfg struct {
```
Roasbeef, line 169: Missing `godoc` comment.

proof/courier.go
```diff
@@ -90,20 +55,123 @@ type CourierHarness interface {
...
+	}
+
+	return &HashMailCourierAddr{
+		addr,
+	}, nil
+}
+
+// NewCourier instantiates a new courier service handle given a service URL
+// address.
+func NewCourier(ctx context.Context, addr url.URL, cfg *CourierCfg) (Courier,
```
Roasbeef, line 158: 👍

tapcfg/server.go
```diff
@@ -180,12 +180,12 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
...
 
-	var hashMailCourier proof.Courier[proof.Recipient]
+	var hashMailCourier proof.Courier
 	if cfg.HashMailCourier != nil &&
-		proofCourierAddr.Scheme == proof.ApertureCourier {
+		proofCourierAddr.Type() == proof.ApertureCourier {
 
 		hashMailBox, err := proof.NewHashMailBox(
-			proofCourierAddr, cfg.HashMailCourier.TlsCertPath,
+			proofCourierAddr.Url(), cfg.HashMailCourier.TlsCertPath,
```
Roasbeef: It's worth nothing that this'll actually dial and try to establish the TCP connection right here and there: https://github.com/lightninglabs/taproot-assets/blob/ef0f3f1cf9dd7e2dea5131e2b275eef6f1e8bd84/proof/courier.go#L183-L187

I think we instead want to defer this out further, as the hashmail may not even be used at all (in favor of other couriers). Or is this primarily for the itests?

tapgarden/custodian.go
```diff
@@ -337,7 +337,13 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 			return err
 		}
 
-		if c.cfg.ProofCourier == nil || addr == nil {
+		// TODO(ffranr): This proof courier disabled check should be
+		//  removed. It was implemented because some integration test do
+		//  not setup and use a proof courier.
+		proofCourierDisabled := c.cfg.ProofCourierCfg == nil ||
```
Roasbeef, line 343: The variable name `proofCourierDisabled` isn't 100% accurate, as before this was also a fall through if the output didn't have an address of ours.

tapgarden/custodian.go
```diff
@@ -351,16 +357,30 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
...
 
+			assetID := addr.AssetID
+
 			log.Debugf("Waiting to receive proof for script key %x",
 				addr.ScriptKey.SerializeCompressed())
 
-			assetID := addr.AssetID
+			// Initiate proof courier service handle from the proof
+			// courier address found in the Tap address.
+			courier, err := proof.NewCourier(
```
Roasbeef, line 367: 👍

proof/courier.go
Roasbeef: So then I think we can simplify slightly by just having it accept the delivery log?

tapcfg/server.go
```diff
@@ -180,32 +180,14 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 		}
 	}
 
-	var hashMailCourier proof.Courier
-	if cfg.HashMailCourier != nil &&
-		proofCourierAddr.Type() == proof.ApertureCourier {
-
-		hashMailBox, err := proof.NewHashMailBox(
-			proofCourierAddr.Url(), cfg.HashMailCourier.TlsCertPath,
```
Roasbeef: Ok cool, I see it was removed in this diff.

tapfreighter/chain_porter.go
```diff
@@ -607,14 +607,43 @@ func (p *ChainPorter) transferReceiverProof(pkg *sendPackage) error {
...
+		)
+		if err != nil {
+			return fmt.Errorf("unable to initiate proof courier "+
+				"service handle: %w", err)
+		}
+
+		// Update courier events subscribers before attempting to
+		// deliver proof.
+		p.subscriberMtx.Lock()
+		courier.SetSubscribers(p.subscribers)
```
Roasbeef, line 631: Why wasn't this needed before?

Roasbeef (changes requested):

proof/courier.go
ffranr: Extra info are things such as:

* TLS certificate
* Backoff procedure parameters
* timeout

proof/courier.go
ffranr: I agree on the TLS certificate. I think we still need a config for the backoff procedure parameters.

proof/courier.go
ffranr: I understand, will move recipient into the new handle creator.

proof/courier.go
ffranr: Ok, no problem. I can move verification of all properties into protocol specific new address functions.

proof/courier.go
ffranr: I was doing some verification before this point so that we could get to a `url.URL` type as soon as possible. Rather than a string.

tappsbt/address.go
```diff
@@ -46,6 +51,9 @@ func FromAddresses(receiverAddrs []*address.Tap,
 		ScriptKey:         asset.NUMSScriptKey,
 	})
 
+	// Map from output index to address.
+	outputIdxToAddr := make(OutputIdxToAddr)
```
guggero, line 55: nit: we could pre-allocate this with the number of receiver addresses.

proof/courier.go
guggero: I think TLS certificates we should just ignore with `tlsConfig = &tls.Config{InsecureSkipVerify: true}`  (and assume that most universes will have a valid, public certificate CA, and if not, we shouldn't fail on that). Since the proofs aren't super sensitive, we don't need to care about transport security that much.

guggero (approved):
LGTM pending @Roasbeef's comments.

tapfreighter/chain_porter.go
ffranr: Courier subscribers were set before in `ChainPorter.RegisterSubscriber`. That setup made sense when there was a single courier instance tide to a variable in the `ChainPorter` struct. But now that we're creating and destroying a courier handle within `ChainPorter.transferReceiverProof` we don't have a chance to hook into `ChainPorter.RegisterSubscriber`.

tapgarden/custodian.go
ffranr: :+1: changed to `skipProofCourier`.

tapcfg/server.go
ffranr: I think, as you noted in another comment, this section of code was removed in a subsequent commit.

proof/courier.go
ffranr: I've added a commit to skip TLS certificate verification (following your comments and this one https://github.com/lightninglabs/taproot-assets/pull/459#discussion_r1310123692 )

Roasbeef (approved):
LGTM 🏄🏾

PR #463, opened 2023-08-22
Cleanup proof courier addr

ffranr:
Major change in this PR is that the proof courier addr field in the Tap address structure is no longer a pointer. I think this change makes sense because the field is mandatory and should never be `nil`.

Created this PR to address these review comments: https://github.com/lightninglabs/taproot-assets/pull/450#pullrequestreview-1588024716 and https://github.com/lightninglabs/taproot-assets/pull/450#pullrequestreview-1588020864

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -989,9 +989,9 @@ func (r *rpcServer) NewAddr(ctx context.Context,
 
 	// Parse the proof courier address if one was provided, otherwise use
 	// the default specified in the config.
-	proofCourierAddr := r.cfg.DefaultProofCourierAddr
+	courierAddr := r.cfg.DefaultProofCourierAddr
 	if in.ProofCourierAddr != "" {
-		proofCourierAddr, err = proof.ParseCourierAddr(
+		courierAddr, err = proof.ParseCourierAddr(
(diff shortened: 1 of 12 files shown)


jharveyb (approved):
LGTM! Happy to have TODOs added and a follow-up PR that can be reviewed quickly.

rpcserver.go
```diff
@@ -1000,6 +1000,14 @@ func (r *rpcServer) NewAddr(ctx context.Context,
 		}
 	}
 
+	// Check that the proof courier address is set. This should never
+	// happen, but we check anyway to avoid panics (possibly caused by
+	// future erroneous config changes).
+	if courierAddr == nil {
```
positiveblue, line 1006: I think this is already checked when creating the config `genServerConfig` but I guess it does not hurt to double check :+1: 

Also, couldn't we just eliminate the pointer from  `cfg.DefaultProofCourierAddr` itself?

ffranr: `cfg.DefaultProofCourierAddr` is optional because we also have a fallback address if a custom default isn't set.

Let me know if that's sub-optimal or if there's a better way.

PR #471, opened 2023-08-28
GitHub: use maintained version of docker layer cache action

guggero:
Looks like the old `satackey/action-docker-layer-caching` action has finally stopped working: https://github.com/lightninglabs/taproot-assets/actions/runs/5999726192/job/16270427354

We update to a forked version that is maintained and also allow the test to continue if the caching fails.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -42,7 +42,8 @@ jobs:
         uses: actions/checkout@v3
 
       - name: docker image cache
-        uses: satackey/action-docker-layer-caching@v0.0.11
+        uses: jpribyl/action-docker-layer-caching@v0.1.1
+        continue-on-error: true
 
       - name: Generate sql models
         run: make sqlc-check


.github/workflows/main.yaml
```diff
@@ -42,7 +42,8 @@ jobs:
         uses: actions/checkout@v3
 
       - name: docker image cache
-        uses: satackey/action-docker-layer-caching@v0.0.11
+        uses: jpribyl/action-docker-layer-caching@v0.1.1
+        continue-on-error: true
```
dstadulis, line 46: After some reading about this workflow, I saw this comment.  Adding a link in case it's relevant for reviewer to assess soundness

https://stackoverflow.com/questions/62045967/is-there-a-way-to-continue-on-error-while-still-getting-correct-feedback#:~:text=If%20a%20job,in%20the%20other_workflow.

dstadulis (commented):
FWIW, `jpribyl/action-docker-layer-caching` has some sort of CI failure itself.  Might be inconsequential: https://github.com/jpribyl/action-docker-layer-caching/actions?query=workflow%3A%22Readme+Test%22

https://github.com/jpribyl/action-docker-layer-caching/actions/runs/5757014264/job/15607468381#step:10:9

guggero:
We just use this caching step to make sure we don't hit the Docker Hub pull limit. We don't actually care that much if it fails from time to time, as long as it does its job most of the time.

PR #472, opened 2023-08-28
Populate genesis meta reveals

jharveyb:
Replacement PR for #458 to switch the author.

Tracking issue: https://github.com/lightninglabs/taproot-assets/issues/441

This PR includes some RPC changes not specified in that issue, but which were needed to get checks working in itest. The protos for proofs will change again in a later PR to include the group witness, and actual verification of these reveal fields will be implemented in a future PR.


halseth (approved):
Great, this is getting there!

itest/assets_test.go
```diff
@@ -199,6 +199,7 @@ func mintAssetUnconfirmed(t *harnessTest, tapd *tapdHarness,
 	unconfirmedAssets := groupAssetsByName(listRespUnconfirmed.Assets)
 	for _, assetRequest := range assetRequests {
 		metaHash := (&proof.MetaReveal{
+			Type: proof.MetaOpaque,
```
Roasbeef, line 202: IIUC, this was the existing default value, so what does this change add?

taprpc/taprootassets.proto
```diff
@@ -236,6 +236,24 @@ message AssetGroup {
     specified by the above key.
     */
     bytes asset_witness = 3;
+
+    // TODO(jhb): update to include tapscript_root
```
Roasbeef, line 240: Why's this need the root, wouldn't that instead be in `GroupKeyReveal` only? The witness would include the information needed to arrive at the root along with the tapscript leaf being revealed.

itest/assets_test.go
guggero: No, the `proof.MetaOpaque` is value `1`. But on the RPC it's `0`. So we didn't have it translated properly between native and RPC types.

guggero (approved):
LGTM :tada:

taprpc/taprootassets.proto
jharveyb: True, will remove in follow-up.

Issue #467, opened 2023-08-24
Add test to ensure that a transfer proof can be inserted using RPC endpoint `InsertProof` and retrieved using RPC endpoint `QueryProof`.


dstadulis:

guggero:
Is this related to universes? Because we do this in basically all integration tests so far.

Issue #439, opened 2023-08-08
Generalise proof query/insert RPC endpoints to support transfer proofs


ffranr:
Generalise proof query and insert RPC endpoints such that they support transfer proofs. They currently only support issuance proofs.

The RPC endpoints are called `QueryProof` and `InsertProof`.

This change will include storing transfer proofs within the multiverse.

Deliverables:
- [ ] Delete code reference in this comment https://github.com/lightninglabs/taproot-assets/issues/439#issuecomment-1672304820
- [x] #467

Roasbeef:
So right now the RPC interface doesn't make a distinction between them. The only area we gate things is here: https://github.com/lightninglabs/taproot-assets/blob/70774640f868988f8c5aa6eb7c977c6e9a105bf1/universe/base.go#L199-L205

ffranr:
I've renamed this issue and changed the description but the actual work won't deviate from what we've planed for this sprint.

ffranr:
Addressed here: https://github.com/lightninglabs/taproot-assets/pull/473

Issue #344, opened 2023-06-07
Implement multiverse transfer proof distribution via RPC courier


ffranr:
A tapd instance should be able to serve as a multiverse transfer proof distribution service. A receiver during the asset send procedure should be able to query the service for the transfer proof associated with the send event.

This work depends on issue https://github.com/lightninglabs/taproot-assets/issues/434

dstadulis:
Need:
Draft PR to add RPC endpoints to see if that would satisfy OKR

ffranr:
@Roasbeef For the time being, should we ignore authentication constraints for the new proof courier RPC endpoints?

Roasbeef:
@ffranr why would there be any authentication constraints? Why do new proof courier RPC endpoints even need to exist?

ffranr:
> why would there be any authentication constraints?

Because a third party service serving as a courier (proof temporary store/retrieve) for free can be abused.

> Why do new proof courier RPC endpoints even need to exist?

To replace our use of hashmail with RPC endpoints that can tie in multiverse construction/verification.

Roasbeef:
> Because a third party service serving as a courier (proof temporary store/retrieve) for free can be abused.

What's the difference between a proof courier as you envision it compared to the universe RPCs after https://github.com/lightninglabs/taproot-assets/pull/416? A universe server allows insertion of a valid state transition proof (rn just issuance, but that just needs a conditional removed), does that differ from what a courier needs to do? Does the Universe interface have any built in DoS protection or cost to submit a valid proof instance? 

> To replace our use of hashmail with RPC endpoints that can tie in multiverse construction/verification.

Does the daemon actually implement end `hashmail` RPC endpoints itself? How does this relate to #432? In that PR, what should be the default behavior if there's no Universe server specified as a TLV field?

ffranr:
Here are some ideas/perspectives I discussed with Oli today:

> What's the difference between a proof courier as you envision it compared to the universe RPCs after https://github.com/lightninglabs/taproot-assets/pull/416? A universe server allows insertion of a valid state transition proof (rn just issuance, but that just needs a conditional removed), does that differ from what a courier needs to do?

A node might want to keep its courier service and multiverse/universe trees separate. For example, the node might only want to track issuance proofs over the long term. If a courier service updates a universe tree with each transfer proof then sync partners would be prompted to sync. The universe root would be modified and the sync node would need to sync even if they were only interested in new issuances.

[...]

PR #473, opened 2023-08-29
Add multiverse RPC proof courier

ffranr:
This PR adds a new multiverse RPC proof service handle for use in the chain porter.

This PR should close these issues when merged:
* https://github.com/lightninglabs/taproot-assets/issues/439
* https://github.com/lightninglabs/taproot-assets/issues/467
* https://github.com/lightninglabs/taproot-assets/issues/344

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/473)

### DIFF
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -822,6 +822,37 @@ func (a *Asset) IsBurn() bool {
 	return IsBurnKey(a.ScriptKey.PubKey, a.PrevWitnesses[0])
 }
 
+// PrimaryPrevID returns the primary prev ID of an asset. This is the prev ID of
+// the first witness, unless the first witness is a split-commitment witness,
+// in which case it is the prev ID of the first witness of the root asset.
+// The first witness effectively corresponds to the asset's direct lineage.
+func (a *Asset) PrimaryPrevID() (*PrevID, error) {
+	if len(a.PrevWitnesses) == 0 {
+		return nil, fmt.Errorf("asset missing previous witnesses")
+	}
+
+	// The primary prev ID is stored on the root asset if this asset is a
+	// split output. We determine whether this asset is a split output by
+	// inspecting the first previous witness.
+	primaryWitness := a.PrevWitnesses[0]
+	isSplitOutput := IsSplitCommitWitness(primaryWitness)
+
+	// If this is a split output, then we need to look up the first PrevID
+	// in the split root asset.
+	if isSplitOutput {
+		rootAsset := primaryWitness.SplitCommitment.RootAsset
+		if len(rootAsset.PrevWitnesses) == 0 {
+			return nil, fmt.Errorf("asset missing previous " +
+				"witnesses")
+		}
+		return rootAsset.PrevWitnesses[0].PrevID, nil
+	}
+
+	// This asset is not a split output, so we can just return the PrevID
+	// found in the first witness.
+	return primaryWitness.PrevID, nil
+}
+
 // Copy returns a deep copy of an Asset.
 func (a *Asset) Copy() *Asset {
 	assetCopy := *a
--- a/itest/tapd_harness.go
+++ b/itest/tapd_harness.go
@@ -177,6 +177,13 @@ func newTapdHarness(t *testing.T, ht *harnessTest, cfg tapdConfig,
 			ReceiverAckTimeout: receiverAckTimeout,
 			BackoffCfg:         backoffCfg,
 		}
+
+	case *UniverseRPCHarness:
+		finalCfg.DefaultProofCourierAddr = fmt.Sprintf(
+			"%s://%s", proof.UniverseRpcCourierType,
+			typedProofCourier.ListenAddr,
+		)
+
 	default:
 		finalCfg.DefaultProofCourierAddr = ""
 		finalCfg.HashMailCourier = nil
--- a/itest/universerpc_harness.go
+++ b/itest/universerpc_harness.go
@@ -1,0 +1,51 @@
+package itest
+
+import (
+	"testing"
+
+	"github.com/lightninglabs/taproot-assets/proof"
+	"github.com/lightningnetwork/lnd/lntest/node"
+	"github.com/stretchr/testify/require"
+)
+
+// UniverseRPCHarness is an integration testing harness for the universe tap
+// service.
+type UniverseRPCHarness struct {
+	// service is the instance of the universe tap service.
+	service *tapdHarness
+
+	// ListenAddr is the address that the service is listening on.
+	ListenAddr string
+}
+
+// NewUniverseRPCHarness creates a new test harness for a universe tap service.
+func NewUniverseRPCHarness(t *testing.T, ht *harnessTest,
+	lndHarness *node.HarnessNode) *UniverseRPCHarness {
+
+	service, err := newTapdHarness(
+		t, ht, tapdConfig{
+			NetParams: harnessNetParams,
+			LndNode:   lndHarness,
+		}, nil, nil, nil,
+	)
+	require.NoError(t, err)
+
+	return &UniverseRPCHarness{
+		service:    service,
+		ListenAddr: service.rpcHost(),
+	}
+}
+
+// Start starts the service.
+func (h *UniverseRPCHarness) Start(_ chan error) error {
+	return h.service.start(false)
+}
+
+// Stop stops the service.
+func (h *UniverseRPCHarness) Stop() error {
+	return h.service.stop(true)
+}
+
+// Ensure that NewUniverseRPCHarness implements the proof.CourierHarness
+// interface.
+var _ proof.CourierHarness = (*UniverseRPCHarness)(nil)
--- a/proof/courier.go
+++ b/proof/courier.go
@@ -54,8 +54,6 @@ type CourierHarness interface {
 // part of the non-interactive send flow. A sender can use this given the
 // abstracted Addr/source type to send a proof to the receiver. Conversely, a
 // receiver can use this to fetch a proof from the sender.
-//
-// TODO(roasbeef): FileSystemCourier, RpcCourier
 type Courier interface {
 	// DeliverProof attempts to delivery a proof to the receiver, using the
 	// information in the Addr type.
--- a/proof/courier.go
+++ b/proof/courier.go
@@ -14,6 +14,8 @@ import (
 	"github.com/lightninglabs/lightning-node-connect/hashmailrpc"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/taprpc"
+	unirpc "github.com/lightninglabs/taproot-assets/taprpc/universerpc"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
@@ -32,6 +34,10 @@ const (
 	// HashmailCourierType is a courier that uses the hashmail protocol to
 	// deliver proofs.
 	HashmailCourierType = "hashmail"
+
+	// UniverseRpcCourierType is a courier that uses the daemon universe RPC
+	// endpoints to deliver proofs.
+	UniverseRpcCourierType = "universerpc"
 )
 
 // CourierHarness interface is an integration testing harness for a proof
@@ -95,6 +101,8 @@ func ParseCourierAddrUrl(addr url.URL) (CourierAddr, error) {
 	switch addr.Scheme {
 	case HashmailCourierType:
 		return NewHashMailCourierAddr(addr)
+	case UniverseRpcCourierType:
+		return NewUniverseRpcCourierAddr(addr)
 	}
 
 	return nil, fmt.Errorf("unknown courier address protocol "+
@@ -155,6 +163,70 @@ func NewHashMailCourierAddr(addr url.URL) (*HashMailCourierAddr, error) {
 	}
 
 	return &HashMailCourierAddr{
+		addr,
+	}, nil
+}
+
+// UniverseRpcCourierAddr is a universe RPC protocol specific implementation of
+// the CourierAddr interface.
+type UniverseRpcCourierAddr struct {
+	addr url.URL
+}
+
+// Url returns the url.URL representation of the courier address.
+func (h *UniverseRpcCourierAddr) Url() *url.URL {
+	return &h.addr
+}
+
+// NewCourier generates a new courier service handle.
+func (h *UniverseRpcCourierAddr) NewCourier(_ context.Context,
+	cfg *CourierCfg, recipient Recipient) (Courier, error) {
+
+	// Ensure that the courier address is a universe RPC address.
+	if h.addr.Scheme != UniverseRpcCourierType {
+		return nil, fmt.Errorf("unsupported courier protocol: %v",
+			h.addr.Scheme)
+	}
+
+	// Connect to the universe RPC server.
+	dialOpts, err := serverDialOpts()
+	if err != nil {
+		return nil, err
+	}
+
+	serverAddr := fmt.Sprintf(
+		"%s:%s", h.addr.Hostname(), h.addr.Port(),
+	)
+	conn, err := grpc.Dial(serverAddr, dialOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	client := unirpc.NewUniverseClient(conn)
+
+	// Instantiate the events subscribers map.
+	subscribers := make(
+		map[uint64]*fn.EventReceiver[fn.Event],
+	)
+
+	return &UniverseRpcCourier{
+		recipient:   recipient,
+		client:      client,
+		deliveryLog: cfg.DeliveryLog,
+		subscribers: subscribers,
+	}, nil
+}
+
+// NewUniverseRpcCourierAddr generates a new universe RPC courier address from a
+// given URL. This function also performs protocol specific address validation.
+func NewUniverseRpcCourierAddr(addr url.URL) (*UniverseRpcCourierAddr, error) {
+	// We expect the port number to be specified.
+	if addr.Port() == "" {
+		return nil, fmt.Errorf("proof courier URI address port " +
+			"unspecified")
+	}
+
+	return &UniverseRpcCourierAddr{
 		addr,
 	}, nil
 }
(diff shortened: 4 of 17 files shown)


proof/courier.go
```diff
@@ -790,7 +915,157 @@ func (h *HashMailCourier) SetSubscribers(
...
+	assetLeaf := unirpc.AssetLeaf{
+		Asset:         rpcAsset,
+		IssuanceProof: rawProof,
+	}
+
+	// Submit proof to courier.
+	_, err = c.client.InsertProof(ctx, &unirpc.AssetProof{
+		Key:       &universeKey,
+		AssetLeaf: &assetLeaf,
+	})
```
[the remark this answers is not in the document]
guggero: Hmm, I'm not 100% sure how we need to do this. On the receiver side we expect the full proof file (containing the whole chain of proofs). But the receiver will only try to fetch that full proof file given the last proof's script key (at least that's how it works now with the hashmail courier).

So I think we might need to change the universe to also allow us to store and fetch whole files, not just single proofs. Not sure how that would look in the tree though.

Or, I guess the alternative would be to do it exactly as you propose, send each individual proof to the universe (that being a no-op for proofs that already exist), but then when _fetching_ a proof in the universe RPC courier we would need to assemble the whole proof chain/file again by iteratively querying each proof in the chain.

Not sure which path you had in mind, @Roasbeef.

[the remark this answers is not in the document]
ffranr: We could also reconstruct a proof file on the server side and offer that up via some RPC (perhaps the same RPC endpoints we currently have but pluralised?).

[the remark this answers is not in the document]
Roasbeef: > So I think we might need to change the universe to also allow us to store and fetch whole files, not just single proofs. Not sure how that would look in the tree though.

The way things are set up rn, you only store the key for _that_ state transition. We'll have a bit of a boostrap phase to start with, but you can imagine that in the future, just like a block explorer, you'll be able to easily look up w/e you need in the universe. Eventually, there'll be a large set of common prefixes (initial issuance, then all the UTXOs from that) that'll be widely cached/known. 

If we store the full proof file, then we'd give up the deduplication we get. Consider that if there was only a single UTXO that created an asset, then as you go on, all assets have a larger shared prefix. By only storing the state transitions in isolation, we can avoid needing to store that prefix over and over again. 

Going back to the top, assuming that we're starting from scratch, with a fresh universe server just spun up: yes the sender will submit each state transition they have individually. On the flip side, after the receiver gets that last state transition, then then crawl backwards via the pointers in the tree to assemble the proof back to genesis. You can think of it kind of like the popularized, but never implemented, "backwards header sync" of Bitcoin. You get the last header, then use the universe APIs to walk backwards.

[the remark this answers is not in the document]
Roasbeef: > We could also reconstruct a proof file on the server side and offer that up via some RPC (perhaps the same RPC endpoints we currently have but pluralised?).

What's the "server side" in this example?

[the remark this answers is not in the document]
guggero: The universe server. So instead of the receiver needing to make multiple RPC requests to fetch the whole chain, there could be a new RPC that assembles the full chain on the universe server and sends it back as a full file to the client. But maybe that can be done as an optimization later on.

ffranr:
The integration test changes reveal that the proof insertion into the new universe RPC courier is failing.

itest/tapd_harness.go
```diff
@@ -166,6 +166,35 @@ func newTapdHarness(ht *harnessTest, cfg tapdConfig,
...
+		if proofReceiverAckTimeout != nil {
+			receiverAckTimeout = *proofReceiverAckTimeout
+		}
+
+		finalCfg.DefaultProofCourierAddr = fmt.Sprintf(
+			"%s://%s", proof.UniverseRpcCourierType,
+			typedProofCourier.ListenAddr,
+		)
+
+		finalCfg.HashMailCourier = &proof.HashMailCourierCfg{
```
guggero, line 193: Can't we simply set this to `nil`? As it shouldn't be used in this case?

itest/universerpc_harness.go
```diff
@@ -0,0 +1,52 @@
...
+type UniverseRPCHarness struct {
+	// service is the instance of the universe tap service.
+	service *tapdHarness
+
+	// ListenAddr is the address that the service is listening on.
+	ListenAddr string
+}
+
+// NewUniverseRPCHarness creates a new test harness for a universe tap service.
+func NewUniverseRPCHarness(t *testing.T, ht *harnessTest,
```
guggero, line 21: It's customary to return a pointer to a struct. That way in `itest/test_harness.go` you can simply do:

```
		proofCourier = NewUniverseRPCHarness(t, ht, lndHarness.Bob)
```

proof/courier.go
```diff
@@ -854,6 +926,195 @@ func (h *HashMailCourier) SetSubscribers(
...
+	// Decode annotated proof into proof file.
+	proofFile := &File{}
+	err := proofFile.Decode(bytes.NewReader(annotatedProof.Blob))
+	if err != nil {
+		return err
+	}
+
+	// Iterate over each proof in the proof file and submit to the courier
+	// service.
+	for i := range proofFile.proofs {
```
guggero, line 965: I think we should use the proof file's helper methods here, like `proofFile.NumProofs()` and `proofFile.ProofAt(idx)`. Then we don't need to do some of the decoding ourselves.

proof/courier.go
```diff
@@ -854,6 +926,195 @@ func (h *HashMailCourier) SetSubscribers(
...
+		if err := transitionProof.Decode(
+			bytes.NewReader(proofBlob),
+		); err != nil {
+			return nil, err
+		}
+
+		revProofs = append(revProofs, transitionProof)
+
+		// Break if no previous witnesses
+		prevWitnesses := transitionProof.Asset.PrevWitnesses
```
guggero, line 1059: I think we should instead query for `HasGenesisWitness()`. Any situation other than that should be an error if the witness is empty.

proof/courier.go
```diff
@@ -854,6 +926,195 @@ func (h *HashMailCourier) SetSubscribers(
...
+				"proof file: %w", err)
+		}
+	}
+
+	// Encode the full proof file.
+	var buf bytes.Buffer
+	if err := proofFile.Encode(&buf); err != nil {
+		return nil, fmt.Errorf("error encoding proof file: %w", err)
+	}
+	proofFileBlob := buf.Bytes()
```
guggero, line 1095: I think we now need to inform the subscribers about a new proof having arrived? Maybe that's why the integration tests aren't working yet, as the custodian isn't informed about new proofs?

guggero (commented):
Looks very good! I dig how small the diff turned out to be in the end :tada:

dstadulis:
This PR will need to rebase to incorporate #483 (which may be subsumed into a PR @guggero opens)

ffranr:
Waiting for rebase on https://github.com/lightninglabs/taproot-assets/pull/484

ffranr:
I think this is failing during proof verification. The proof courier's VM is trying to verify that the asset associated with the proof was correctly split from its input asset. But I don't think that the input asset snapshot is available in the proof.

In a separate part of the codebase we perform proof verification on a complete proof file. And in this way we have the previous asset snapshot at any given transition proof. See https://github.com/ffranr/taproot-assets/blob/main/proof/verifier.go#L433-L450

When inserting a proof into the universe RPC courier we insert every proof in turn starting at the issuance proof. So we could reform the proof file before verification. But this strategy would entail reforming the proof file every time a transition proof is inserted into the courier.

@guggero any thoughts on this?

taprpc/universerpc/marshal.go
```diff
@@ -0,0 +1,45 @@
...
+		Index:   int32(outPoint.Index),
+	}
+}
+
+// MarshalAssetKey returns an RPC ready AssetKey.
+func MarshalAssetKey(outPoint wire.OutPoint,
+	scriptKeyPubKey *btcec.PublicKey) AssetKey {
+
+	scriptKeyBytes := scriptKeyPubKey.SerializeCompressed()
+	rpcOutPoint := MarshalOutpoint(outPoint)
```
guggero, line 25: If we return a pointer from these helper functions directly, we can use them inline and don't need to create another variable. Makes things more compact and IMO a bit easier to read. And in gRPC land, things are _always_ pointers of structs, so might as well return them as that in specific marshal functions.

proof/courier.go
```diff
@@ -854,6 +926,199 @@ func (h *HashMailCourier) SetSubscribers(
...
+
+	return &AnnotatedProof{
+		Locator: originLocator,
+		Blob:    proofFileBlob,
+	}, nil
+}
+
+// SetSubscribers sets the subscribers for the courier. This method is
+// thread-safe.
+func (c *UniverseRpcCourier) SetSubscribers(
```
guggero, line 1109: We're currently not using anything subscriber related. Can we just remove this, the map and the mutex? Or is it part of an interface?

universe/base.go
```diff
@@ -5,6 +5,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/lightninglabs/taproot-assets/asset"
```
guggero, line 9: nit: need to run `make fmt`.

universe/base.go
```diff
@@ -184,11 +187,21 @@ func (a *MintingArchive) RegisterIssuance(ctx context.Context, id Identifier,
...
 	//
 	// TODO(roasbeef): add option to skip proof verification?
-	assetSnapshot, err := newProof.Verify(ctx, nil, a.cfg.HeaderVerifier)
+
+	// Before we can validate a non-issuance proof we need to fetch the
+	// previous asset snapshot (which is the proof verification result for
+	// the previous/parent proof in the proof file).
+	prevAssetSnapshot, err := a.getPrevAssetSnapshot(ctx, id, newProof)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch previous asset "+
```
guggero, line 198: nit: use `%w` when wrapping an error.

universe/base.go
```diff
@@ -245,6 +258,72 @@ func (a *MintingArchive) RegisterIssuance(ctx context.Context, id Identifier,
...
+	if newProof.Asset.HasGenesisWitness() {
+		return nil, nil
+	}
+
+	// Query for proof associated with the previous asset.
+	prevID := newProof.Asset.PrevWitnesses[0].PrevID
+	prevOutPoint := prevID.OutPoint
+
+	var allZeros [32]byte
+	if reflect.DeepEqual(prevOutPoint.Hash.CloneBytes(), allZeros[:]) {
```
guggero, line 275: Can just compare `if prevOutPoint.Hash == allZeroes {`. Though I think this shouldn't be necessary, as that should be covered in the `HasGenesisWitness()`. Let's discuss offline.

guggero: Okay, I know why this is the case. If this is a split commitment, you instead need to use the root asset's `PrevID`. See this commit from the burn PR: https://github.com/lightninglabs/taproot-assets/blob/e1af71979112702614f8d1e72aae914cab9cd270/asset/witness.go#L19

guggero (commented):
Took another look, looking really good. The issue with the empty `PrevID` is the split commitment as suspected. Should be very easy to fix.

universe/base.go
guggero: Maybe we could generalize that part into a `func GetFirstPrevOut(witness TxWitness)` function?

taprpc/universerpc/marshal.go
ffranr: I'm happy to go down that route because it seems to be a common pattern in our code. But I want to say that there's something clearly wrong somewhere (with Go or otherwise) when we choose pointers for the sake of convenience and not for their function. Normally in most languages I would avoid pointers as much as possible.

universe/base.go
ffranr: I'll definitely spin this bit of logic into a function. I think it will help document (in a single spot) what's going on.

proof/courier.go
ffranr: It's part of the proof courier interface. I think the hashmail courier's backoff procedure uses this to publish events.

proof/courier.go
```diff
@@ -854,6 +926,203 @@ func (h *HashMailCourier) SetSubscribers(
...
+
+// ReceiveProof attempts to obtain a proof file from the courier service. The
+// final proof in the target proof file is identified by the given locator.
+func (c *UniverseRpcCourier) ReceiveProof(ctx context.Context,
+	originLocator Locator) (*AnnotatedProof, error) {
+
+	// Sleep for a bit to give the sender sufficient time to transfer the
+	// proof to the proof courier service.
+	//
+	// TODO(ffranr): Replace this with the backoff procedure as found in the
```
guggero, line 1023: TODO for this PR or follow-up?

rpcserver.go
```diff
@@ -2734,17 +2734,28 @@ func (r *rpcServer) QueryProof(ctx context.Context,
 		return nil, err
 	}
 
+	rpcsLog.Debugf("[QueryProof]: fetching proof at "+
```
guggero, line 2737: nit: `%x` with `.String()`, will double hex encode.

universe/base.go
```diff
@@ -245,6 +257,60 @@ func (a *MintingArchive) RegisterIssuance(ctx context.Context, id Identifier,
 	return issuanceProof, nil
 }
 
+func (a *MintingArchive) getPrevAssetSnapshot(ctx context.Context,
```
guggero, line 260: nit: missing Godoc.

universe/base.go
```diff
@@ -245,6 +257,60 @@ func (a *MintingArchive) RegisterIssuance(ctx context.Context, id Identifier,
...
+		return nil, nil
+	}
+
+	// Query for proof associated with the previous asset.
+	prevID := newProof.Asset.PrimaryPrevID()
+
+	// Parse script key for previous asset.
+	prevScriptKeyBytes := prevID.ScriptKey
+	prevScriptKeyPubKey, err := btcec.ParsePubKey(
+		prevScriptKeyBytes[:],
```
guggero, line 275: nit: can just use `prevID.ScriptKey[:]` here.

universe/base.go
```diff
@@ -245,6 +257,60 @@ func (a *MintingArchive) RegisterIssuance(ctx context.Context, id Identifier,
...
+	uniID Identifier, newProof proof.Proof) (*proof.AssetSnapshot, error) {
+
+	// If this is a genesis proof, then there is no previous asset (and
+	// therefore no previous asset snapshot).
+	if newProof.Asset.HasGenesisWitness() {
+		return nil, nil
+	}
+
+	// Query for proof associated with the previous asset.
+	prevID := newProof.Asset.PrimaryPrevID()
```
guggero, line 270: Since this _can_ be nil in case there is no witness (shouldn't happen, but better be safe than sorry), we should check and return an error (invalid proof), otherwise this can trigger a panic if I send an invalid proof to the universe.

guggero: Or maybe we shouldn't just return `nil` but have it return an error instead. Better be ultra cautious in any validation code.

guggero (commented):
Very nice, great to see this working!
Just one blocking comment, the rest are nits.

asset/asset.go
```diff
@@ -808,6 +808,42 @@ func (a *Asset) IsUnSpendable() bool {
...
+// witness.
+func IsSplitCommitWitness(witness Witness) bool {
+	return witness.PrevID != nil && len(witness.TxWitness) == 0 &&
+		witness.SplitCommitment != nil
+}
+
+// PrimaryPrevID returns the primary prev ID of an asset. This is the prev ID of
+// the first witness, unless the first witness is a split-commitment witness,
+// in which case it is the prev ID of the first witness of the root asset.
+func (a *Asset) PrimaryPrevID() *PrevID {
```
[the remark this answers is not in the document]
ffranr: I was hoping this conversation would come up.

I used "primary" because I think that it is special. The sort of questions I want to ask are: why does the first witness get special treatment? Is it different from the others? Why do we have a method just for the first witness?

If it is different from the others then should we recognise its difference based on whether its first or something else? Calling it "primary" gives us that flexibility.

proof/courier.go
ffranr: Follow up PR. It works as is now.

asset/asset.go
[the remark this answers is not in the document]
guggero: The first PrevOut is basically the "direct lineage" of a proof, the continuation of a straight line (e.g. a flat proof file with linked proofs). Any other inputs must be specified in the `AdditionalInputs` of a proof, so they require a different treatment.
"Primary" to me indicates a different role than the other entries (as in the primary instance of a DB cluster is the only one that is in write mode). But the first PrevID doesn't have a different role per se, it references something slightly different than the other ones. But let's not spend too much time on this discussion, I'm okay with either name.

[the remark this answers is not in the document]
ffranr: Ok, thanks for your thoughts. I'll add the "direct lineage" idea into the doc.

ffranr:
Rebased. Which was a bit non-trivial.

universe/base.go
```diff
@@ -298,6 +311,58 @@ func (a *MintingArchive) RegisterNewIssuanceBatch(ctx context.Context,
...
+	uniID Identifier, newProof proof.Proof) (*proof.AssetSnapshot, error) {
+
+	// If this is a genesis proof, then there is no previous asset (and
+	// therefore no previous asset snapshot).
+	if newProof.Asset.HasGenesisWitness() {
+		return nil, nil
+	}
+
+	// Query for proof associated with the previous asset.
+	prevID, err := newProof.Asset.PrimaryPrevID()
```
guggero, line 326: We're still missing a `nil` check for the `prevID`. Normally we'd validate a proof first so this wouldn't be necessary. But because we're accessing fields from it before we've validated it, someone could craft a malicious proof and get the universe to panic when we access `prevID.ScriptKey` below.

guggero (commented):
Still one missing check, other than that this looks good!

universe/base.go
```diff
@@ -306,6 +311,58 @@ func (a *MintingArchive) RegisterNewIssuanceBatch(ctx context.Context,
...
 }
 
+// getPrevAssetSnapshot returns the previous asset snapshot for the passed
+// proof. If the proof is a genesis proof, then nil is returned.
+func (a *MintingArchive) getPrevAssetSnapshot(ctx context.Context,
+	uniID Identifier, newProof proof.Proof) (*proof.AssetSnapshot, error) {
+
+	// If this is a genesis proof, then there is no previous asset (and
+	// therefore no previous asset snapshot).
+	if newProof.Asset.HasGenesisWitness() {
```
guggero, line 321: Ah, and I guess we also need to make sure `newProof.Asset` isn't nil here.

guggero: Ah yes, my bad. Sorry.

universe/base.go
ffranr: I've added a `prevID == nil` check below this function call.

guggero (approved):
LGTM :tada:

ffranr:
I'll create another PR to rename methods such as `RegisterIssuance` and interfaces such as `BaseMultiverse` which now support transfer proofs in addition to issuance proofs.

proof/courier.go
```diff
@@ -159,6 +167,70 @@ func NewHashMailCourierAddr(addr url.URL) (*HashMailCourierAddr, error) {
...
+	conn, err := grpc.Dial(serverAddr, dialOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	client := unirpc.NewUniverseClient(conn)
+
+	// Instantiate the events subscribers map.
+	subscribers := make(
+		map[uint64]*fn.EventReceiver[fn.Event],
```
Roasbeef, line 207: Thought there was some other higher level wrapper here?

proof/courier.go
```diff
@@ -849,6 +921,206 @@ func (h *HashMailCourier) SetSubscribers(
...
+	// recipient describes the recipient of the proof.
+	recipient Recipient
+
+	// client is the RPC client that the courier will use to interact with
+	// the universe RPC server.
+	client unirpc.UniverseClient
+
+	// deliveryLog is the log that the courier will use to record the
+	// attempted delivery of proofs to the receiver.
+	deliveryLog DeliveryLog
```
Roasbeef, line 934: Would assume this now plays a pretty minimal role. Only thing can think of is if the server is down for w/e reason (unreachable).

proof/courier.go
```diff
@@ -849,6 +921,206 @@ func (h *HashMailCourier) SetSubscribers(
...
+			outPoint, proofAsset.ScriptKey.PubKey,
+		)
+		universeID := unirpc.MarshalUniverseID(proofAsset.ID())
+		universeKey := unirpc.UniverseKey{
+			Id:      universeID,
+			LeafKey: assetKey,
+		}
+
+		// Submit proof to courier.
+		_, err = c.client.InsertProof(ctx, &unirpc.AssetProof{
```
Roasbeef, line 995: We'll want to double check what the return is here if the proof already exists. To avoid bailing out early when not needed.

proof/courier.go
```diff
@@ -849,6 +921,206 @@ func (h *HashMailCourier) SetSubscribers(
...
+	// Decode annotated proof into proof file.
+	proofFile := &File{}
+	err := proofFile.Decode(bytes.NewReader(annotatedProof.Blob))
+	if err != nil {
+		return err
+	}
+
+	// Iterate over each proof in the proof file and submit to the courier
+	// service.
+	for i := 0; i < proofFile.NumProofs(); i++ {
```
Roasbeef, line 958: Related to the below, I think we can optimize things a bit here (need to check about the validation site though) by inserting things _backwards_. Overtime, one can expect that the prefix is known nearly universally, so you just add that last transition, then check that N-1 is there (or not), bailing out if it is.

proof/courier.go
```diff
@@ -849,6 +921,206 @@ func (h *HashMailCourier) SetSubscribers(
...
+// final proof in the target proof file is identified by the given locator.
+func (c *UniverseRpcCourier) ReceiveProof(ctx context.Context,
+	originLocator Locator) (*AnnotatedProof, error) {
+
+	// Sleep for a bit to give the sender sufficient time to transfer the
+	// proof to the proof courier service.
+	//
+	// TODO(ffranr): Replace this with the backoff procedure as found in the
+	// hashmail courier.
+	time.Sleep(5 * time.Second)
```
Roasbeef, line 1018: Why's this needed at all? Seems like an itest quirk? Why can't we rely on the normal back off stuff?

proof/courier.go
```diff
@@ -849,6 +921,206 @@ func (h *HashMailCourier) SetSubscribers(
...
+		// Decode transition proof from query response.
+		proofBlob := resp.AssetLeaf.IssuanceProof
+		var transitionProof Proof
+		if err := transitionProof.Decode(
+			bytes.NewReader(proofBlob),
+		); err != nil {
+			return nil, err
+		}
+
+		revProofs = append(revProofs, transitionProof)
```
Roasbeef, line 1057: Re this and above, what about the additional inputs? So the case where you have an asset, but it was created by merging 3 change UTXOs.

proof/courier.go
```diff
@@ -95,6 +101,8 @@ func ParseCourierAddrUrl(addr url.URL) (CourierAddr, error) {
 	switch addr.Scheme {
 	case HashmailCourierType:
 		return NewHashMailCourierAddr(addr)
+	case UniverseRpcCourierType:
+		return NewUniverseRpcCourierAddr(addr)
```
Roasbeef, line 103: Gotta love it when a good abstraction comes together 😎

universe/base.go
```diff
@@ -236,14 +236,6 @@ func (a *MintingArchive) verifyIssuanceProof(ctx context.Context, id Identifier,
 		return nil, fmt.Errorf("asset id mismatch: expected %v, got %v",
 			id.AssetID, newAsset.ID())
 
-	// The outpoint of the final resting place of the asset should match
-	// the leaf key
-	//
-	// TODO(roasbeef): this restrict to issuance
-	case assetSnapshot.OutPoint != key.MintingOutpoint:
```
Roasbeef, line 243: +1 for removing here 

I think this is also where we'd add the logic to ensure we only get issuance proofs or not (depending on the future config).

universe/base.go
```diff
@@ -207,10 +219,11 @@ func (a *MintingArchive) RegisterIssuance(ctx context.Context, id Identifier,
 // verifyIssuanceProof verifies the passed minting leaf is a valid issuance
 // proof, returning the asset snapshot if so.
 func (a *MintingArchive) verifyIssuanceProof(ctx context.Context, id Identifier,
-	key BaseKey, leaf *MintingLeaf) (*proof.AssetSnapshot, error) {
+	key BaseKey, leaf *MintingLeaf,
+	prevAssetSnapshot *proof.AssetSnapshot) (*proof.AssetSnapshot, error) {
 
 	assetSnapshot, err := leaf.GenesisProof.Verify(
-		ctx, nil, a.cfg.HeaderVerifier,
+		ctx, prevAssetSnapshot, a.cfg.HeaderVerifier,
```
Roasbeef, line 226: I guess we can also rename `GenesisProof` above?

universe/base.go
```diff
@@ -298,6 +311,62 @@ func (a *MintingArchive) RegisterNewIssuanceBatch(ctx context.Context,
...
+	}
+
+	prevProof := prevProofs[0].Leaf.GenesisProof
+
+	// Construct minimal asset snapshot for previous asset.
+	// This is a minimal the proof verification result for the
+	// previous (input) asset. We know that it was already verified
+	// as it was present in the multiverse/universe archive.
+	return &proof.AssetSnapshot{
+		Asset:    &prevProof.Asset,
```
Roasbeef, line 365: What about the other set of assets in the case of inputs merged to create this new asset?

Roasbeef (commented):
Great work! Gotta love it when a design plan comes together neatly 😎

Main thing that popped out is to ensure that we're handling cases of merged inputs for an asset properly.

Roasbeef:
Also I think we can use the `HasSplitCommitmentWitness` method in most places here?

proof/proof.go
```diff
@@ -131,7 +131,9 @@ type Watcher interface {
 // Proof encodes all of the data necessary to prove a valid state transition for
 // an asset has occurred within an on-chain transaction.
 type Proof struct {
-	// PrevOut is the previous on-chain outpoint of the asset.
+	// PrevOut is the previous on-chain outpoint of the asset. This outpoint
+	// is that of the first on-chain input. Outpoints which correspond to
+	// the other inputs can be found in AdditionalInputs.
```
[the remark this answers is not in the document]
ffranr: The existence of `PrevOut` as a field was/is confusing to me. Because in the context of a multi input spend (merge) we seem to make an arbitrary choice of "the asset" that we're spending, is that correct?

> this is the outpoint where the asset currently resides (or previously).

In the context of a multi input spend, what do you mean by "the asset" here?

It's this confusion/uncertainty that motivated me to add more to the comment to make sure I understand what's going on. I think by "the asset" we mean the first input, and that's what I've tried to capture in the comment.

asset/asset.go
```diff
@@ -808,6 +808,44 @@ func (a *Asset) IsUnSpendable() bool {
 		a.Amount == 0
 }
 
+// IsSplitCommitWitness returns true if the witness is a split-commitment
+// witness.
+func IsSplitCommitWitness(witness Witness) bool {
```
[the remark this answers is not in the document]
ffranr: Yes I agree, and the same goes for the new universe marshalling function. I'll see what I can do about making greater use of these functions after this PR is merged.

proof/courier.go
ffranr: At the moment each courier type handles its own list of subscribers. I think it does make sense to add a higher level wrapper but we haven't done that yet and it's not in this PR.

universe/base.go
ffranr: Yes, there's also a `BaseKey` and a few other things to rename. I've got a separate PR on the go for that.

proof/courier.go
ffranr: By mistake I actually didn't use this field, so I'm going to modify this PR to make use of it now.

proof/courier.go
ffranr: The custodian waits for a relevant on-chain output and if found attempts to retrieve a proof file from the proof courier. But the sending node might still be in the middle of submitting a proof file one proof at a time to the proof courier service. Which means that the entire proof file (and the latest proof prefix) isn't available yet. So the receiving node waits before attempting to retrieve the proof file.

We need some sort of pause before trying to retrieve a proof file because (if all parties are involved are online) we need a time window for the sender to submit the proof and the courier to validate and serve. Otherwise the receiver's first attempt at retrieval will likely always fail. The backoff procedure does not include that delay but would overcome this issue by failing once and retrying.

We should have a constant for the `time.Sleep(5 * time.Second)` delay, I'll add that. I'm also going to move the sleep into the custodian.

I'll see what I can do about using the backoff procedure within the universe RPC courier, but it probably would be best to do that in a different PR.

proof/courier.go
ffranr: The `testBasicSendUnidirectional` itest covers this because it performs two different send events for the same asset (single genesis), it spends the change. I've also manually tested that `InsertProof` can correctly handle a proof that it's seen before.

proof/courier.go
ffranr: Each individual (transition) `Proof` contains a field called ` AdditionalInputs []File`. The entire proof file for each additional inputs is included in that field. At least, that's my current understanding after speaking with Oli.

We noticed that this is a memory inefficiency within the universe tree (on the courier server side). Each leaf of the universe tree contains a proof and each one of those proofs may contain entire proof files for additional inputs. Those proof files may already exist in the universe as other leaves.

However, from a tapd peer sending/receiving perspective, this setup makes things simpler. I'll add an issue to talk about the potential optimisation here. It will help me get a clear picture of what's going on.

ffranr: Issue here: https://github.com/lightninglabs/taproot-assets/issues/503

ffranr:
> Also I think we can use the HasSplitCommitmentWitness method in most places here?

Where are you referring to please @Roasbeef ?

universe/base.go
ffranr: I think this relies on the additional inputs field, see https://github.com/ffranr/taproot-assets/blob/main/proof/verifier.go#L206-L213

And my answer here: https://github.com/lightninglabs/taproot-assets/pull/473#discussion_r1324892755

Roasbeef (approved):
LGTM 🐊

> Reviewed 6 of 10 files at r7, 7 of 7 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/473#-:-NeihETU3pnactsusuTg:b-lilaks)* status: 13 of 17 files reviewed, 34 unresolved discussions (waiting on @ffranr and @guggero)

___
*[`tapgarden/custodian.go` line 391 at r8](https://reviewable.io/reviews/lightninglabs/taproot-assets/473#-Neih-Ku3PERv53MDsTC:-Neih-Ku3PERv53MDsTD:bf4cmt5) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/147af79b670d14ba7b77ba74dcfcdb02b949aaba/tapgarden/custodian.go#L391)):*
> ```Go
> 			// Sleep to give the sender an opportunity to transfer
> 			// the proof to the proof courier service.
> 			time.Sleep(defaultProofRetrievalDelay)
> ```

I think let's endeavor to remove this in a follow up PR. IMO the backoff should just handle this, and it can speed up our itests with it gone as well.

PR #474, opened 2023-08-29
build: update to Go 1.21.0

Roasbeef:


guggero:
The RPC and linter containers also need to switch to `-alpine` (or `-bullseye` where we do some `apt install`).

Roasbeef:
Pushed up a new version that's moved to bookworm.

guggero:
The RPC compilation docker file needs a change too, as we have hard-coded versions in there (to make sure the compiled/formatted output is deterministic). See https://github.com/lightningnetwork/lnd/pull/7944#discussion_r1312706437 for more info.

Dockerfile
```diff
@@ -4,7 +4,7 @@
 # /make/builder.Dockerfile
 # /taprpc/Dockerfile
 # /tools/Dockerfile
-FROM golang:1.20.5-alpine as builder
+FROM golang:1.21.0-bookworm as builder
```
guggero, line 7: We can't just switch from `-alpine` (completely different Linux distro with a different package manager) over to Debian. And the Alpine ones are more stable in terms of us just being able to bump the version numbers, compared to Debian. So I'd rather do it the other way, switch from Debian to Alpine (but that requires deeper changes as well, so maybe let's defer that).

dev.Dockerfile
```diff
@@ -15,12 +15,12 @@ ENV CGO_ENABLED=0
 RUN make install
 
 # FINAL IMAGE
-FROM alpine as final
+FROM bookworm as final
```
guggero, line 18: Same here, we want to keep the slim alpine base image, as the Debian is much larger.

taprpc/Dockerfile
```diff
@@ -4,7 +4,7 @@
 # /dev.Dockerfile
 # /make/builder.Dockerfile
 # /tools/Dockerfile
-FROM golang:1.20.5-buster
+FROM golang:1.21.0-bookworm
```
guggero, line 7: Need to bump the protobuf compiler and formatter below too, see comment above.

guggero (changes requested):
Almost there!
Linter seems to have been a flake, passes now.

guggero (approved):
Nice, LGTM :tada:

Dockerfile
jharveyb: This switch of distro is still in here?

guggero: Oh yes, I missed this one in my review.

PR #477, opened 2023-08-30
Provable asset burning

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/24.

~Depends on https://github.com/lightninglabs/taproot-assets/pull/482.~

Still work in progress, main things missing are:
 - ~More unit tests~
 - ~CLI command `tapcli assets burn`~
 - Integration tests with 4 test cases:
    - ~Burn all coins in a commitment (expect error because of edge case)~
    - ~Burn all assets of one ID while there are other (passive) assets (=> expect root asset for burn)~
    - ~Burn some assets, get some change back (=> expect split asset for burn)~
    - ~Burn assets from multiple inputs~


rpcserver.go
```diff
@@ -1892,6 +1904,155 @@ func (r *rpcServer) SendAsset(_ context.Context,
...
+	}
+
+	fundedVPkt, err := r.cfg.AssetWallet.FundPacket(ctx, desc, vPkt)
+	if err != nil {
+		return nil, fmt.Errorf("error funding address send: %w", err)
+	}
+
+	// If there is only a single output, it means that we are attempting to
+	// burn all the assets that were selected and that there are no passive
+	// assets either. This would mean we'd create a BTC level output with
```
Roasbeef, line 1980: Wouldn't this be the same as the tombstone scenario? With the only difference being this new burn NUMs key is used instead of the existing burn/nums key?

Roasbeef: Or it's even just a send as "normal", with no need for an anchor at all, as it looks sort of like UTXO consolidation.

tapfreighter/parcel.go
```diff
@@ -203,9 +203,9 @@ type PreSignedParcel struct {
 	// vPkt is the virtual transaction that should be delivered.
 	vPkt *tappsbt.VPacket
 
-	// inputCommitment is the commitment for the input that is being spent
-	// in the virtual transaction.
-	inputCommitment *commitment.TapCommitment
+	// inputCommitments are the commitments for the input that are being
+	// spent in the virtual transaction.
+	inputCommitments tappsbt.InputCommitments
```
Roasbeef, line 208: Ah so we didn't actually support multi-input here before?

asset/asset.go
```diff
@@ -808,6 +808,45 @@ func (a *Asset) IsUnSpendable() bool {
...
+	if a.ScriptKey.PubKey == nil {
+		return false
+	}
+
+	var prevID PrevID
+
+	// If this is a split output, then we need to look up the first PrevID
+	// in the split root asset.
+	if a.HasSplitCommitmentWitness() {
+		if len(a.PrevWitnesses) == 0 ||
```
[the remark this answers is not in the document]
guggero: Ah you're right, those checks are already in `HasSplitCommitmentWitness()`. Was just being cautious, but I guess that makes it a bit too verbose.

asset/asset.go
```diff
@@ -808,6 +808,45 @@ func (a *Asset) IsUnSpendable() bool {
...
+			a.PrevWitnesses[0].SplitCommitment == nil {
+
+			return false
+		}
+
+		rootAsset := a.PrevWitnesses[0].SplitCommitment.RootAsset
+		if len(rootAsset.PrevWitnesses) == 0 ||
```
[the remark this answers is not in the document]
guggero: This is just to avoid a panic if we ever get an incorrectly constructed proof. I guess we would always validate a proof first, but I'd say better safe than sorry?

rpcserver.go
guggero: Ah, true. But I guess that way we'd end up sending 1k sats to a BTC output that can't be spent anymore (at least not with the current RPCs). I can remove that restriction if we want and create an issue that we'll eventually need a "sweep BTC from outputs with empty commitments/tombstones/burns".

asset/burn_test.go
```diff
@@ -0,0 +1,125 @@
...
+func TestDeriveBurnKey(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name        string
+		prevID      PrevID
+		expectedKey string
+	}{
+		{
+			name:   "empty prev ID",
```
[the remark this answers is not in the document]
guggero: I just added that for completeness' sake (and to have a second test vector).
Not sure if this could ever be used in real life, as it would mean that the mint itself would go into a burn key directly. So mint and burn in the very same state transition. Maybe it could be useful to reveal possession of a group key or something?
Can remove the test vector or indicate it is not really a valid value.

tapfreighter/parcel.go
guggero: Yes, for the "externally signed" (vPSBT) flow we didn't fully implement the multi input everywhere, only for the address send. So there might be more commits needed to support some of the integration test cases I have in mind.

asset/burn.go
```diff
@@ -0,0 +1,42 @@
...
+
+// DeriveBurnKey derives a provably un-spendable but unique key by tweaking the
+// public NUMS key with a tap tweak:
+//
+//	burnTweak = h_tapTweak(NUMSKey || outPoint || assetID || scriptKey)
+//	burnKey = NUMSKey + burnTweak*G
+//
+// The firstPrevID must be the PrevID from the first input that is being spent
+// by the virtual transaction that contains the burn.
+func DeriveBurnKey(firstPrevID PrevID) *btcec.PublicKey {
```
jharveyb, line 20: Compared to the latest sketch, this doesn't consider if an asset is grouped or not. IIUC that's alright as outpoint + ID + scriptkey will be globally unique anyways, and we don't need this burn key tweak to map directly to an asset's location in the Tap tree.

guggero:
Addressed all comments and fixed all TODOs. This is ready for full review.

Roasbeef:
Pushed up a rebased version.

asset/burn.go
Roasbeef: Hmm true, but yeah don't think it matters since the asset ID is unique.

tapfreighter/wallet.go
```diff
@@ -511,6 +521,141 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
...
+		Inputs: []*tappsbt.VInput{{
+			PrevID: asset.PrevID{
+				ID: fundDesc.ID,
+			},
+		}},
+		Outputs: []*tappsbt.VOutput{{
+			Amount:            0,
+			Type:              tappsbt.TypeSplitRoot,
+			AnchorOutputIndex: 0,
+			ScriptKey:         asset.NUMSScriptKey,
```
Roasbeef, line 600: Thought tombstones weren't needed for this case? cc @jharveyb

tapfreighter/wallet.go
```diff
@@ -511,6 +521,141 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// We want to avoid a BTC output being created that just sits there
+	// without an actual commitment in it. So if we are not getting any
+	// change or passive assets in this output, we'll not want to go through
+	// with it.
+	changeOut := fundedPkt.VPacket.Outputs[0]
```
Roasbeef, line 630: If the above is unchanged though, then doesn't this lead to us just sending those other funds to the NUMs key?

tapdb/assets_store.go
```diff
@@ -2445,22 +2445,44 @@ func (a *AssetStore) ConfirmParcelDelivery(ctx context.Context,
...
+					"key: %w", err)
+			}
+
+			isNumsKey := scriptPubKey.IsEqual(asset.NUMSPubKey)
 			isTombstone := isNumsKey &&
 				out.Amount == 0 &&
 				out.OutputType != int16(
 					tappsbt.TypePassiveAssetsOnly,
 				)
+			isBurn := !isNumsKey && len(witnessData) > 0 &&
```
Roasbeef, line 2475: Confused here: thought burning was _just_ sending to the output, with no special witness needed?

tapscript/send.go
```diff
@@ -443,6 +440,17 @@ func PrepareOutputAssets(ctx context.Context, vPkt *tappsbt.VPacket) error {
 				return ErrInvalidCollectibleSplit
 			}
 
+			// There should only be a tombstone output in an
+			// interactive flow if we need to transport passive
+			// assets. Otherwise, for an interactive send we don't
+			// need a tombstone output and this wouldn't be a two
+			// output collectible send.
+			if !rootOut.Type.CanCarryPassive() &&
```
Roasbeef, line 448: Needs unit test

tapdb/assets_store.go
Roasbeef: Nvm needed for if the asset was a split.

tapfreighter/wallet.go
jharveyb: Yes, but I think from `fundPacketWithInputs` and the logic right after this, that key will never stay as NUMS if the amount is non-zero; this either becomes a normal split, or the tombstone is removed.

tapfreighter/wallet.go
jharveyb: IIUC the tombstone output is added here so that `fundPacketWithInputs` handles change calculation on building a valid change output if needed. The branch later on handles the full-value burn situation, so we should never end up with an actual tombstone. I think the comment could be a bit more verbose in nothing that the final result will not be a tombstone.

jharveyb: IIUC the tombstone output is added here so that `fundPacketWithInputs` handles change calculation on building a valid change output if needed. The branch later on handles the full-value burn situation, so we should never end up with an actual tombstone. I think the comment could be a bit more verbose in nothing that the final result will not be a tombstone.

rpcserver.go
```diff
@@ -1966,6 +1972,98 @@ func (r *rpcServer) SendAsset(_ context.Context,
...
+		copy(assetID[:], assetIDBytes)
+
+	default:
+		return nil, fmt.Errorf("asset ID must be specified")
+	}
+
+	if in.AmountToBurn == 0 {
+		return nil, fmt.Errorf("amount to burn must be specified")
+	}
+	if in.ConfirmationText != AssetBurnConfirmationText {
```
jharveyb, line 2004: 💯

itest/burn_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// Preparation: We derive a couple of keys, so we can spread out our
+	// assets over several outputs, which we are going to use for the first
+	// couple of test cases.
+	scriptKey1, anchorInternalKeyDesc1 := deriveKeys(t.t, t.tapd)
+	scriptKey2, anchorInternalKeyDesc2 := deriveKeys(t.t, t.tapd)
+	scriptKey3, anchorInternalKeyDesc3 := deriveKeys(t.t, t.tapd)
+	scriptKey4, _ := deriveKeys(t.t, t.tapd)
+
+	// We create the following outputs:
```
jharveyb, line 48: Very helpful for following the flow & cases here!

jharveyb (approved):
Looks good, cool to see how the freighter changes are relatively contained!

Gave my nits offline or in existing threads.

tapfreighter/wallet.go
guggero: Correct, they aren't. This is just to give the wallet an asset output it can attach any passive assets to. We remove it again later if there are no passive assets. Added a more explicit comment to it.

tapfreighter/wallet.go
guggero: Correct. We might want to eventually refactor the whole wallet code a bit to either make some of the assumptions less implicit or by just documenting them better.

Roasbeef (approved):
LGTM 🦜

Issue #287, opened 2023-03-07
Fix unique key constraint in assets table


guggero:
Currently the `UNIQUE` constraint in the `assets` is broken, since adding the primary key of the table to the `UNIQUE` constraint basically makes it useless: https://github.com/lightninglabs/taro-private/blob/main/tarodb/sqlc/migrations/000002_assets.up.sql#L177

AFAIK the constraint was changed to allow multiple assets with the NUMS script key.
Instead we could use NULL as the `script_key_id`, interpret that as being the NUMS point and turn the unique constraint into a `CHECK` function that checks different uniqueness levels, depending on whether the `script_key_id` is null or not.

guggero:
Duplicate of https://github.com/lightninglabs/taproot-assets/issues/691.

PR #481, opened 2023-08-31
Fix regeneration of static proof files

guggero:
This is a replacement PR for a commit that is already on the `v0-3-0-staging` branch (https://github.com/lightninglabs/taproot-assets/commit/60aabd05c85a24bc317c78798d177cb761f26c29) but turned out to not fully work.

This is a new approach where we have a new `make` goal (`make gen-itest-test-vectors`) that updates those regtest test files and test vectors with content from an actual integration test run. This previously was done manually whenever something in the proof file encoding format changed.

I'll rebase the `v0-3-0-staging` branch once this is merged.


guggero:
@jharveyb this should allow us to update all proof test files and test vectors very easily once the new genesis/meta reveal and witness changes are done.

jharveyb (approved):
LGTM! Having the Postgres DB persisted is a nice add-on.

itest/send_test.go
```diff
@@ -250,6 +256,13 @@ func testBasicSendPassiveAsset(t *harnessTest) {
 	secondAsset := rpcAssets[1]
 	genInfo2 := secondAsset.AssetGenesis
 
+	testVectors := &proof.TestVectors{}
+	addProofTestVectorFromFile(
```
Roasbeef, line 260: Rationale for this is that we're able to have more complete coverage as these vectors are generated e2e vs with canned data? Trying to understand what the gap here was with the existing unit test based vectors.

Roasbeef (approved):
LGTM 🚛

Left some comment sin line to better understand the gaps here.

itest/send_test.go
```diff
@@ -327,10 +314,9 @@ func testBasicSendPassiveAsset(t *harnessTest) {
 	// Assert that the outbound transfer was confirmed.
 	expectedAmtAfterSend = assets[1].Asset.Amount - numUnitsSend
 	confirmAndAssertOutboundTransfer(
-		t, t.tapd, sendResp, genInfo.AssetId,
+		t, t.tapd, sendResp, genInfo2.AssetId,
 		[]uint64{expectedAmtAfterSend, numUnitsSend}, 1, 2,
 	)
-	_ = sendProof(t, t.tapd, recvTapd, recvAddr.ScriptKey, genInfo)
```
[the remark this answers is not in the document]
guggero: Yes, covered here: https://github.com/lightninglabs/taproot-assets/issues/287.

itest/send_test.go
guggero: The issue is that I created some of the test vectors (for example `proof/testdata/proof_tlv_encoding_other.json`) by manually taking some proofs from an integration test run to add some "real" proofs that actually fully validate and aren't just unit test random/dummy data.
I usually updated those files manually again when something in the proof format changed. But that became a bit tedious and wasn't really documented anywhere. So now this process is automated.

Issue #414, opened 2023-07-24
[bug]: Initial universe sync fails due to empty 


jharveyb:
### Background

When starting a new tapd server, default universe sync fails due to a default empty universe server being present in the universe federation list.

### Your environment

tapd 0.2.3 + PR 413, lnd 0.16.2-beta, bitcoind 0.25.1, testnet

### Steps to reproduce

Initialize a new tapd server on testnet. Start an initial universe sync with:

`tapcli u sync --universe_host testnet.universe.lightning.finance:10029`

### Expected behavior

Server syncs all leaves without an error.

### Actual behavior

From the server logs at the default logging level:

```bash
2023-07-24 17:24:20.549 [INF] UNIV: Adding new Universe server to Federation, addrs=([]universe.ServerAddr) (len=2 cap=2) {
 (universe.ServerAddr) {
  ID: (uint32) 0,
  addrStr: (string) "",
  addr: (net.Addr) <nil>
 },
 (universe.ServerAddr) {
  ID: (uint32) 0,
  addrStr: (string) (len=40) "testnet.universe.lightning.finance:10029",
  addr: (net.Addr) <nil>
 }
}

2023-07-24 17:24:20.550 [WRN] UNIV: unable to add universe servers: <nil>
2023-07-24 17:24:20.550 [INF] RPCS: Starting RPC Server
2023-07-24 17:24:20.550 [INF] UNIV: Syncing Universe state with server=(universe.ServerAddr) {
 ID: (uint32) 0,
 addrStr: (string) (len=40) "testnet.universe.lightning.finance:10029",
 addr: (net.Addr) <nil>
}
```

PR #482, opened 2023-09-04
Small fixes in freighter and wallet

guggero:
A couple of commits extracted out of https://github.com/lightninglabs/taproot-assets/pull/477 that don't necessarily have to do with burning itself.

Fixes https://github.com/lightninglabs/taproot-assets/issues/414.

### DIFF
--- a/tapfreighter/wallet.go
+++ b/tapfreighter/wallet.go
@@ -508,6 +508,14 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
 		return nil, err
 	}
 
+	return f.fundPacketWithInputs(ctx, fundDesc, vPkt, selectedCommitments)
+}
+
+// fundPacketWithInputs funds a virtual transaction with the given inputs.
+func (f *AssetWallet) fundPacketWithInputs(ctx context.Context,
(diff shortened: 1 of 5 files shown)


tapfreighter/wallet.go
```diff
@@ -515,6 +515,12 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
 
 	totalInputAmt := uint64(0)
 	for _, anchorAsset := range selectedCommitments {
+		// We only use the sum of all assets of the same ID to avoid
+		// counting passive assets as well.
+		if anchorAsset.Asset.ID() != fundDesc.ID {
+			continue
+		}
+
```
ffranr, line 531: I don't understand this change. Couldn't passive asset also have the same asset ID at this point?

ffranr (approved):
:+1: Looks good!

Just one question.

tapfreighter/wallet.go
guggero: Not necessarily. If you look at `removeActiveCommitments`, we declare all assets as "passive" that don't have the same `TapCommitmentKey()`.
But I guess we should use the same comparison here as well, not just the ID. Will update.

jharveyb (approved):
LGTM! Nice catch re: input value.

PR #483, opened 2023-09-04
Remove sendProof where possible from itests

ffranr:
`sendProof` is a function which transports proof by querying from the sender node and inserting into the receiver node (all via RPC). This function should be avoided and the proof courier system should be used whenever possible.


guggero:
Replaced by https://github.com/lightninglabs/taproot-assets/pull/484.

PR #484, opened 2023-09-04
Fix proof courier in integration test

guggero:
Replaces https://github.com/lightninglabs/taproot-assets/pull/483.

We noticed that the hashmail proof courier didn't work properly in the integration tests (while working on the new RPC based courier).
This PR fixes two issues that lead to the hashmail proof courier not working correctly in the itest:
 - One test sends to the same TAP address multiple times, but the custodian only ever looked at one and then bailed out (because we didn't delete completed events, kudos to @ffranr for finding this).
 - The hashmail courier looks at previous send attempts and then goes into a backoff wait if the same proof (or at least what looks like the same proof because it has the same script key and asset ID because it's two different transfers to the same address). That backoff time was simply longer than we waited for the proof to arrive.

### DIFF
--- a/tapgarden/custodian.go
+++ b/tapgarden/custodian.go
@@ -337,11 +337,16 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 			return err
 		}
 
+		// We are not interested in the outpoint if we don't know of a
+		// pre-stored address associated with it.
+		if addr == nil {
+			continue
+		}
+
 		// TODO(ffranr): This proof courier disabled check should be
 		//  removed. It was implemented because some integration test do
 		//  not setup and use a proof courier.
-		skipProofCourier := c.cfg.ProofCourierCfg == nil || addr == nil
-		if skipProofCourier {
+		if c.cfg.ProofCourierCfg == nil {
 			continue
 		}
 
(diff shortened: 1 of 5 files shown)


tapgarden/custodian.go
```diff
@@ -580,12 +585,20 @@ func (c *Custodian) mapProofToEvent(p proof.Blob) error {
 
 	// Check if any of our in-flight events match the last proof's state.
 	for _, event := range c.events {
-		if AddrMatchesAsset(event.Addr, &lastProof.Asset) {
+		if AddrMatchesAsset(event.Addr, &lastProof.Asset) &&
+			event.Outpoint == lastProof.OutPoint() {
+
```
ffranr, line 590: :key:

tapgarden/custodian.go
```diff
@@ -580,12 +585,20 @@ func (c *Custodian) mapProofToEvent(p proof.Blob) error {
...
 			// state of the address event to mark it as completed
 			// successfully.
-			return c.setReceiveCompleted(event, lastProof, file)
+			err = c.setReceiveCompleted(event, lastProof, file)
+			if err != nil {
+				return fmt.Errorf("error updating event: %w",
+					err)
+			}
+
+			delete(c.events, event.Outpoint)
```
ffranr, line 601: We used to `return` here after finding a single matching event. I think it's better as you have it now.

ffranr (approved):
Nice! :100:

jharveyb (approved):
Looks good! 🚀

PR #487, opened 2023-09-05
itest: add perf testing binary

### LINKED ISSUE #469: Implement Performance Testing Binary for `tapd` Instances
Develop a binary for performance testing which can execute load tests against "real" `tapd` instances.

The tests need to be able to run in our infrastructure (regtest) and locally. For that, the perf tests need to be decoupled of the environment using dependency injection (tap rpc clients,  chain interactions,...)

 The binary is responsible for:
   - Setup Phase: Load generic configuration from a file, allowing command-line flag overrides. Configuration includes test selection, create rpc clients for the main tapd instances (Alice and Bob), Prometheus configuration, etc.
   
   - Test Execution: 
         - set up phase for the specifc test (create more clients, load file data, etc..)
         - record metrics for the current execution
         - execute test logic
         
   - Shutdown: push test metrics to  prometheus push gateway  (some of them may be needed to be pulled from the tapd instances itself)

The binary will be compiled using `go test -c`, and custom configuration for specific tests will be loaded from a separate path.

Acceptance Criteria:

- [x] Ability to load and override configurations. 
- [x] Execution of tests on tapd interfaces. <= Using `mintBatchStressTest` as show case
- [ ] Collection and pushing of metrics. <= can be added in a follow up PR
    
Part of #433

positiveblue:
Add the scaffolding of the perf testing binary. The binary gets a list of tests and executes them (sequentially) against external tapd/lnd/btcd instances.

First commit is just making more itest helpers public so they can be used in the perf_testing package.

I was not able to run the tests against a bitcoind instance because the `Mine` rpc endpoint was disabled. btcd may fix that

How to test it:
- Compile the binary: `go test -c -o perf_test` (no makefile target yet)
- Start btcd, Alice (lnd/tapd) and Bob (lnd/tapd)
- Create`perf-test.conf` using the sample cfg 
- `./perf_test`

Fix #469


itest/performance/mint_batch_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/itest"
+	"github.com/lightninglabs/taproot-assets/taprpc"
+	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	unirpc "github.com/lightninglabs/taproot-assets/taprpc/universerpc"
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed testdata/8k-metadata.hex
+var imageMetadataHex []byte
```
guggero, line 25: Very nice, wasn't aware that you could directly load into bytes, thought you always needed to use `embed.Fs`.
That reminds me, would be nice to have a `make` goal to compile this test binary so we don't have to do that manually.

itest/performance/perf_test.go
```diff
@@ -0,0 +1,49 @@
...
+// execTestCase is the method in charge of executing a single test case.
+func execTestCase(t *testing.T, ctx context.Context, testName string,
+	cfg *Config) {
+
+	testTimeout := time.Duration(cfg.TestTimeout) * time.Minute
+	ctxt, cancel := context.WithTimeout(ctx, testTimeout)
+	defer cancel()
+
+	switch testName {
+	case "mint_batch_stress":
```
guggero, line 43: We could use the same mechanism as we already do in the itest where we define all test cases as a list and run them as sub tests if it is enabled by the config. But just an idea, non-blocking.

guggero (approved):
Looks great, can't wait to see it in action!

itest/performance/perf_test.go
positiveblue: this is a good point, I will update it to follow the same pattern before merging

itest/utils.go
```diff
@@ -0,0 +1,237 @@
...
+
+	copied := make([]*mintrpc.MintAssetRequest, len(reqs))
+	for idx := range reqs {
+		copied[idx] = CopyRequest(reqs[idx])
+	}
+	return copied
+}
+
+// ParseOutPoint
+func ParseOutPoint(s string) (*wire.OutPoint, error) {
```
Roasbeef, line 44: Style nit: incomplete `godoc`.

itest/performance/config.go
```diff
@@ -0,0 +1,136 @@
...
+	User     string `long:"user" description:"bitcoind/btcd user name"`
+	Password string `long:"password" description:"bitcoind/btcd password"`
+	TLSPath  string `long:"tlspath" description:"Path to btcd's TLS certificate, if TLS is enabled"`
+}
+
+// Config holds the main configuration for the performance testing binary.
+type Config struct {
+	// TestCases is a comma separated list of test cases that will be
+	// executed.
+	TestCases string `long:"test-case" description:"the test case that will be executed"`
```
Roasbeef, line 42: I think this can be a `[]byte`, then it'll pick it up as a slice. So if we want cases `A`, `B`, and `C`, the binary can use the args: `--testcase=A --testcase=B --testcase=B`. This way, any upstream templating (in helm or w/e) doesn't need to handle the CSV format.

itest/performance/config.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Bob is the configuration for the secondary user in the network.
+	Bob *User `group:"bob" namespace:"bob" description:"bob related configuration"`
+
+	// Bitcoin is the configuration for the bitcoin backend.
+	Bitcoin *BitcoinConfig `group:"bitcoin" namespace:"bitcoin" long:"bitcoin" description:"bitcoin client configuration"`
+
+	// BatchSize is the number of assets to mint in a single batch. This is only
+	// relevant for some test cases.
+	BatchSize int `long:"batch-size" description:"the number of assets to mint in a single batch"`
```
Roasbeef, line 55: I think this works for now, but in the future, we might want to modify things to extract the `TestCases` struct into its own sub-config. Imagine a test that wants to send an asset back n forth 1k times, and with a group key and w/o, etc -- we'll want additional dimensions in the arg to specify.

itest/performance/config.go
```diff
@@ -0,0 +1,136 @@
+package perf_test
```
Roasbeef, line 1: Name suggestion for the binary: `loadtest`?

itest/performance/mint_batch_test.go
```diff
@@ -0,0 +1,193 @@
...
+		cfg.Alice.Tapd.Port)
+
+	timeout := time.Duration(cfg.TestTimeout) * time.Minute
+	mintBatchStressTest(
+		t, ctx, bitcoinClient, alice, bob, aliceHost, cfg.BatchSize,
+		imageMetadataBytes, timeout,
+	)
+}
+
+func mintBatchStressTest(t *testing.T, ctx context.Context,
```
Roasbeef, line 56: Looks like this mostly copies the existing tests vs being able to re-use the same test scenario func?

itest/performance/perf_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestPerformance executes the configured performance tests.
+func TestPerformance(t *testing.T) {
+	cfg, err := LoadConfig()
+	require.NoError(t, err, "unable to load main config")
+
+	//
```
Roasbeef, line 19: Stray starting comment.

itest/performance/config.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Bitcoin is the configuration for the bitcoin backend.
+	Bitcoin *BitcoinConfig `group:"bitcoin" namespace:"bitcoin" long:"bitcoin" description:"bitcoin client configuration"`
+
+	// BatchSize is the number of assets to mint in a single batch. This is only
+	// relevant for some test cases.
+	BatchSize int `long:"batch-size" description:"the number of assets to mint in a single batch"`
+
+	// TestSuiteTimeout is the timeout for the entire test suite in minutes.
+	TestSuiteTimeout int `long:"test-suite-timeout" description:"the timeout for the entire test suite in minutes"`
```
Roasbeef, line 58: You can make both this and the one below a `time.Duration`. Then users can specify `5m`, `1s`, etc.

itest/performance/testdata/8k-metadata.hex
```diff
@@ -0,0 +1 @@
...3e45f01710bce053047f01c10b3e45f01710bce053047f01c10b3e45f01710bce0535607ff96815ef29f34ce3a9fd2f8421fc640d30315bce08b19687aa082177c31034d0f54f0822f66a0e9810a5ef0c50c343d50c10bbe9881a6072a78c11733d0f440052ff862069a1ea8e0055fcc40d30315bce08b19687aa082177c31034d0f54f0822f66a0e9810a5ef0c50c343d50c10bbe9881a607fa86e06ff91191713fa34f0f54f0822f66f4e9810a5ef0c58c3e3d50c10bbe98d1a7072a78c11733faf440052ff862469f1ea8e0055fcce8d30315bce08b197d7aa082177c31a34f0f54f0822f66f4e9810a5ef0c58c3e3d50c10bbe98d1a7072a78c11733faf440052ff862469f1ee8689ce7fffa3bbef1e1f39225a5077afc099ecfb3a4f4408f3fc1f37996941ee8f127783ecf92d2033dfe04cfe759527aa0c79fe0f93c4b4a0ff4f8133c9f6749e9811e7f82e7f32c293dd0e34ff07c9e25a5077afc099ecfb3a4f4408f3fc1f37996941ee8f127783ecf92d2033dfe04cfe759527aa0c79fe0f93c4b4a0ff4f87b47f0eff869181f2941d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029669de9810a9e62d6991ea8e029f63f3df5c34a25705adb0000000049454e44ae426082
```
Roasbeef, line 1: I think we already have this saved somewhere as well?

Roasbeef (changes requested):

itest/performance/config.go
guggero: You mean `[]string`, right? Otherwise it'll still be a single string.

Roasbeef:
Needs a rebase!

itest/performance/config.go
Roasbeef: Err yeah, `[]string`.

Roasbeef (approved):
LGTM 🧩

Issue #402, opened 2023-07-13
proof: add a set of magic bytes to the serialized proof file format


Roasbeef:
In order to easily detect if a blob is a proof file (has the version header, etc), or just a single state transition, we should add a set of magic bytes, similar to what `sqlite` does: https://www.sqlite.org/fileformat.html#magic_header_string

All serialized proof files then _must_ start with this value, and it can make distinguishing between raw files and a state transition easier. Related to #401 

[Edit] Summarized task list:

- [ ] Update binary encoding of proof files and proof transitions to include the prefix
- [ ] Update all test vectors and unit test data
- [ ] Update the `DecodeProof` RPC to detect the format and decode it accordingly
- [ ] Go through all RPC methods and fields that accept or return proofs and make it very clear if it's a file or an individual proof

guggero:
In addition to adding a prefix (I'd vote for `0x54 0x41 0x50 0x50` => `TAPP` for a single proof and `0x54 0x41 0x50 0x46` => `TAPF` for a proof file), we should go through the RPC interface and add a comment (and possibly rename fields where necessary) to make it more clear whether a field contains (or expects) a file or a single proof.

Summarized task list:

- [ ] Update binary encoding of proof files and proof transitions to include the prefix
- [ ] Update all test vectors and unit test data
- [ ] Update the `DecodeProof` RPC to detect the format and decode it accordingly
- [ ] Go through all RPC methods and fields that accept or return proofs and make it very clear if it's a file or an individual proof

PR #488, opened 2023-09-06
proof: add magic bytes to individual proofs and proof files

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/402.


rpcserver.go
```diff
@@ -1152,11 +1152,18 @@ func (r *rpcServer) VerifyProof(ctx context.Context,
 	}
 	valid := err == nil
 
-	decodedProof, err := r.marshalProofFile(ctx, proofFile, 0, false, false)
+	p, err := proofFile.ProofAt(uint32(proofFile.NumProofs() - 1))
```
ffranr, line 1155: We could use `proofFile.LastProof()` here.

rpcserver.go
```diff
@@ -1168,57 +1175,87 @@ func (r *rpcServer) VerifyProof(ctx context.Context,
 func (r *rpcServer) DecodeProof(ctx context.Context,
 	in *taprpc.DecodeProofRequest) (*taprpc.DecodeProofResponse, error) {
 
-	if len(in.RawProof) == 0 {
+	if len(in.RawProof) < 4 {
```
ffranr, line 1178: I think `proof.PrefixMagicBytesLength` would be clearer here, rather than `4`. But also see my other comment on the idea of not exporting the magic bytes concept out of the `proof` package.

taprpc/assetwalletrpc/assetwallet.swagger.json
```diff
@@ -668,7 +668,8 @@
         },
         "new_proof_blob": {
           "type": "string",
-          "format": "byte"
+          "format": "byte",
+          "description": "The new individual transition proof (not a full proof file) that proves\nthe new inclusion of the asset within the new AnchorTx."
```
ffranr, line 672: nit: `the new inclusion of the asset` -> `the inclusion of the asset`

taprpc/taprootassets.proto
```diff
@@ -504,6 +504,8 @@ message TransferOutput {
 
     uint64 amount = 4;
 
+    // The new individual transition proof (not a full proof file) that proves
+    // the new inclusion of the asset within the new AnchorTx.
```
ffranr, line 508: nit: `the new inclusion` -> `the inclusion`

taprpc/taprootassets.proto
```diff
@@ -684,6 +686,8 @@ message DecodeAddrRequest {
 }
 
 message ProofFile {
+    // The raw proof files encoded as bytes. Must be a file and not just an
+    // individual mint/transfer proof.
     bytes raw_proof = 1;
```
ffranr, line 691: It would probably be a good idea to rename this field to `proof_file_bytes` at some point. Not sure if this is the right PR for that.

taprpc/taprootassets.proto
```diff
@@ -693,7 +697,8 @@ message DecodedProof {
     // The index depth of the decoded proof, with 0 being the latest proof.
     uint32 proof_at_depth = 1;
 
-    // The total number of proofs contained in the raw proof.
+    // The total number of proofs contained in the raw proof (this will always
+    // be 1 if a single mint/transition proof was given as the raw_proof).
     uint32 number_of_proofs = 2;
```
ffranr, line 702: > this will always be 1 if a single mint/transition proof was given as the raw_proof

In the comment above you say that `raw_proof` must be a file. In which case, it can not consist of 1 transition proof.

I think you should make this change in the highlighted comment:
`a single mint/transition proof was given as the raw_proof` -> `the serialised proof file given in raw_proof contained a single mint (issuance) proof`

ffranr (commented):
Great to have this change, thank you!

rpcserver.go
```diff
@@ -1168,57 +1175,87 @@ func (r *rpcServer) VerifyProof(ctx context.Context,
...
-	withMetaReveal bool) (*taprpc.DecodedProof, error) {
-
-	decodedProof, err := proofFile.ProofAt(depth)
-	if err != nil {
-		return nil, err
-	}
-
+// marshalProof turns a transition proof into an RPC DecodedProof.
+func (r *rpcServer) marshalProof(ctx context.Context, p *proof.Proof,
+	withPrevWitnesses, withMetaReveal bool) (*taprpc.DecodedProof, error) {
```
ffranr, line 1249: nit: lint error, needs new line after function signature.

rpcserver.go
guggero: Totally, just forgot we had that. Updated.

rpcserver.go
guggero: Makes sense, changed.

rpcserver.go
```diff
@@ -1168,57 +1175,87 @@ func (r *rpcServer) VerifyProof(ctx context.Context,
...
-	if err := proofFile.Decode(bytes.NewReader(in.RawProof)); err != nil {
-		return nil, fmt.Errorf("unable to decode proof file: %w", err)
-	}
+	var (
+		prefix      = in.RawProof[:proof.PrefixMagicBytesLength]
+		proofReader = bytes.NewReader(in.RawProof)
+		rpcProof    *taprpc.DecodedProof
+	)
+	switch {
+	case bytes.Equal(prefix, proof.PrefixMagicBytes[:]):
```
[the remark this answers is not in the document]
guggero: You're absolutely right, makes way more sense. Changed.

taprpc/taprootassets.proto
guggero: Yeah, I think this _is_ the PR for doing that.

taprpc/taprootassets.proto
guggero: In this case we actually explicitly allow either a single proof or a proof file (which _could_ just contain a single proof, but would look differently when encoded as a file) to be given. But yeah, maybe my phrasing wasn't optimal. How about the new version?

ffranr:
Postgres itest failed in CI, I've re-run.

taprpc/taprootassets.proto
ffranr: Got it, looks good.

ffranr (approved):
Nice!

guggero:
> Postgres itest failed in CI, I've re-run.
> 
> EDIT: and it's failed again with:
> 
> ```
> 2023-09-07 13:08:59.719 [INF] TADB: Connecting to Postgres fixture: ***localhost:32828/test?sslmode=disable
> ```

Yeah, the postgres itest failure is really weird. It looks like the process is just killed after some time. Sounds like an issue with the GitHub environment. Maybe we're using too much memory? I'll try to investigate. Here's something I found: https://github.com/Lightning-Universe/lightning-bolts/issues/409

rpcserver.go
```diff
@@ -1156,11 +1157,18 @@ func (r *rpcServer) VerifyProof(ctx context.Context,
 	}
 	valid := err == nil
 
-	decodedProof, err := r.marshalProofFile(ctx, proofFile, 0, false, false)
+	p, err := proofFile.LastProof()
```
jharveyb, line 1160: Is this a bug fix? IIUC this is now handling the most recent proof instead of the genesis proof (which according to the godoc is what we want).

jharveyb (approved):
Good to have this feature and make the RPC & CLI more consistent! + unblock builders

rpcserver.go
guggero: Yes, I guess so. The parameter was called `depth` in `marshalProofFile`, even though it used that as the index. So before we would return the very first proof as the decoded proof. But the RPC comment mentions: `// The decoded last proof in the file if the proof file was valid.`, so I think it's correct now.

PR #489, opened 2023-09-06
rpcserver: validate rpc requests

positiveblue:
Make sure that we check that we check that req fields are not null before accessing them.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -294,6 +294,10 @@ func (r *rpcServer) GetInfo(context.Context,
 func (r *rpcServer) MintAsset(ctx context.Context,
 	req *mintrpc.MintAssetRequest) (*mintrpc.MintAssetResponse, error) {
 
+	if req.Asset == nil {
+		return nil, fmt.Errorf("asset cannot be nil")
+	}
+
 	// An asset name is mandatory, and cannot be the empty string.
 	if len(req.Asset.Name) == 0 {
 		return nil, fmt.Errorf("asset name cannot be empty")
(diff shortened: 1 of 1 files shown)


rpcserver.go
```diff
@@ -294,6 +294,14 @@ func (r *rpcServer) GetInfo(context.Context,
 func (r *rpcServer) MintAsset(ctx context.Context,
 	req *mintrpc.MintAssetRequest) (*mintrpc.MintAssetResponse, error) {
 
+	if req == nil {
```
guggero, line 297: I think the gRPC library actually makes sure this cannot happen. At least we don't have this check anywhere else. This would only be a problem if we call the methods directly. Or was there ever an instance where this could be triggered from RPC/REST?

rpcserver.go
```diff
@@ -294,6 +294,14 @@ func (r *rpcServer) GetInfo(context.Context,
...
+		return nil, fmt.Errorf("request cannot be nil")
+	}
+
+	if req.Asset == nil {
```
guggero, line 301: :+1:

guggero (commented):
Nice fix!

rpcserver.go
positiveblue: you are right, chat gpt fooled me. I check for the subfields and I saw that there was some reflection tricks when you use the methods to get the fields instead of accessing the fields directly. But if you send a nil req, the connection fails

`rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp :0: connect: connection refused"`

Roasbeef (approved):
LGTM 🧩

Issue #466, opened 2023-08-23
asset+tapgarden+vm: Update group key signing to produce a witness over a minting virtual TX


jharveyb:
Sub-issue of #394, depends on PRs #448 and #458 or similar.

As part of the asset and proof spec changes, asset membership in a group is now authorized by a generic witness stack and not a signature over the asset `Genesis` data. This witness must be validated over a TX, which will now be the virtual TX that represents an asset minting, as specified here https://github.com/Roasbeef/bips/pull/33 .

This requires new logic to construct these minting virtual TXs, adjustments of VM (and possibly other) logic to allow an asset witness on minting if and only if the asset has a group, and further extensions to the VM to actually validate this new witness if the asset is being minted and has a group.

I *think* these changes should be limited mostly to the VM but I'm not sure. Fixing other subsystems may be deferred to a future PR.

**Requirements**

- [ ] New logic to construct virtual TXs for asset minting
- [ ] Modify group key signing logic to sign over the minting virtual TX
- [ ] Modify minting to include the group key witness in the `Asset` PrevWitness field
- [ ] New logic to verify that a virtual TX is for asset minting
- [ ] Modify VM asset minting exception to verify the witness for asset group mints
- [ ] New VM unit tests for asset group witnesses that use the script path

Pessimistically sizing this as Medium due to uncertainty around the exact format of the virtual TX and the VM changes needed to get this working.

dstadulis:
The virtual tx construction is somewhat unclear in the BIP -- and may not be entirely what we want to do.  JHB to provide more context

jharveyb:
Linking to BIP comment about the exact structure of the virtual TX and some underspecification there:

https://github.com/Roasbeef/bips/pull/33#issuecomment-1690271321

PR #490, opened 2023-09-06
Group key witness support

jharveyb:
Implements changes outlined in https://github.com/lightninglabs/taproot-assets/issues/466.

Known TODOs:

- [ ] VM execution in `DeriveGroupKey` to verify witness correctness
- [ ] Copy group witness to `asset.PrevWitnesses` where need so the witness gets serialized
- [ ] update `proof.Verify` to use new genesis asset definition
- [ ] Resolve dependency cycle with `tapscript` package
- [ ] New unit test cases for features introduced in past PRs

Approach:

- Modify the `virtualTx` methods to add new branches for creating a virtual TX for a genesis grouped asset.

- Modify the VM to prevent an early exit for genesis grouped assets, and add early assertions for such assets.

- Update `DeriveGroupKey` to sign over a virtualTx representing minting of a genesis grouped asset.

- Update `DeriveGroupKey` to accept external scripts and support different signing methods.

Open questions:

- Can we reuse the existing `VirtualTxSigner` for the `GenesisSigner`?

- Is there other data needed for a group key script path authorization here besides a single `TapLeaf`? In `CreateTaprootSignature` I see validation against other fields but not use of those other fields for signing.

- How can we resolve the `tapscript` dependency cycle? I see two issues, one being the placement of `RawKeyGenesisSigner` code, which AFAICT is only used in tests. The other is that, the `asset` package can't call methods from the `tapscript` package without causing a cycle, but with the current setup `DeriveGroupKey` must build a virtual TX at some point. I think one option may be to move some of the group signing logic into the `tapscript` package.

### DIFF
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -12,9 +12,13 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcutil/psbt"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightninglabs/taproot-assets/commitment"
 	"github.com/lightninglabs/taproot-assets/mssmt"
+	"github.com/lightninglabs/taproot-assets/tapscript"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/keychain"
 	"github.com/lightningnetwork/lnd/tlv"
@@ -469,10 +473,21 @@ func (g *GroupKeyReveal) GroupPubKey(assetID ID) (*btcec.PublicKey, error) {
 		return nil, err
 	}
 
-	internalKey := input.TweakPubKeyWithTweak(rawKey, assetID[:])
+	return GroupPubKey(rawKey, assetID[:], g.TapscriptRoot[:])
+}
+
+func GroupPubKey(rawKey *btcec.PublicKey, singleTweak, tapTweak []byte) (
+	*btcec.PublicKey, error) {
+
+	if len(singleTweak) != sha256.Size || len(tapTweak) != sha256.Size {
+		return nil, fmt.Errorf("tweaks must be %d bytes", sha256.Size)
+	}
+
+	internalKey := input.TweakPubKeyWithTweak(rawKey, singleTweak)
 	tweakedGroupKey := txscript.ComputeTaprootOutputKey(
-		internalKey, g.TapscriptRoot[:],
+		internalKey, tapTweak,
 	)
+
 	return tweakedGroupKey, nil
 }
 
@@ -727,14 +742,11 @@ func NewScriptKeyBip86(rawKey keychain.KeyDescriptor) ScriptKey {
 // GenesisSigner is used to sign the assetID using the group key public key
 // for a given asset.
 type GenesisSigner interface {
-	// SignGenesis tweaks the public key identified by the passed key
-	// descriptor with the the first passed Genesis description, and signs
-	// the second passed Genesis description with the tweaked public key.
-	// For minting the first asset in a group, only one Genesis object is
-	// needed, since we tweak with and sign over the same Genesis object.
-	// The final tweaked public key and the signature are returned.
-	SignGenesis(keychain.KeyDescriptor, Genesis,
-		*Genesis) (*btcec.PublicKey, *schnorr.Signature, error)
+	// SignVirtualTx generates a signature according to the passed signing
+	// descriptor and TX.
+	SignVirtualTx(signDesc *lndclient.SignDescriptor, tx *wire.MsgTx,
+		prevOut *wire.TxOut) (*btcec.PublicKey, *schnorr.Signature,
+		error)
 }
 
 // RawKeyGenesisSigner implements the GenesisSigner interface using a raw
@@ -757,42 +769,27 @@ func NewRawKeyGenesisSigner(priv *btcec.PrivateKey) *RawKeyGenesisSigner {
 // For minting the first asset in a group, only one Genesis object is
 // needed, since we tweak with and sign over the same Genesis object.
 // The final tweaked public key and the signature are returned.
-func (r *RawKeyGenesisSigner) SignGenesis(keyDesc keychain.KeyDescriptor,
-	initialGen Genesis, currentGen *Genesis) (*btcec.PublicKey,
+func (r *RawKeyGenesisSigner) SignVirtualTx(signDesc *lndclient.SignDescriptor,
+	virtualTx *wire.MsgTx, prevOut *wire.TxOut) (*btcec.PublicKey,
 	*schnorr.Signature, error) {
 
-	if !keyDesc.PubKey.IsEqual(r.privKey.PubKey()) {
+	signerPubKey := r.privKey.PubKey()
+
+	if !signDesc.KeyDesc.PubKey.IsEqual(signerPubKey) {
 		return nil, nil, fmt.Errorf("cannot sign with key")
 	}
 
-	// TODO(jhb): Update to two-phase tweak
-	tweakedPrivKey := txscript.TweakTaprootPrivKey(
-		*r.privKey, initialGen.GroupKeyTweak(),
+	sig, err := SignVirtualTx(r.privKey, signDesc, virtualTx, prevOut)
+
+	// Derive the final tweaked public key also used for signing.
+	tweakedGroupKey, err := GroupPubKey(
+		signerPubKey, signDesc.SingleTweak, signDesc.TapTweak,
 	)
-
-	// If the current genesis is not set, we are minting the first asset in
-	// the group. This means that we use the same Genesis object for both
-	// the key tweak and to create the asset ID we sign. If the current
-	// genesis is set, the asset type of the new asset must match the type
-	// of the first asset in the group.
-	id := initialGen.ID()
-	if currentGen != nil {
-		if initialGen.Type != currentGen.Type {
-			return nil, nil, fmt.Errorf("asset group type mismatch")
-		}
-
-		id = currentGen.ID()
-	}
-
-	// TODO(roasbeef): this actually needs to sign the digest of the asset
-	// itself
-	idHash := sha256.Sum256(id[:])
-	sig, err := schnorr.Sign(tweakedPrivKey, idHash[:])
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("cannot tweak group key: %w", err)
 	}
 
-	return tweakedPrivKey.PubKey(), sig, nil
+	return tweakedGroupKey, sig, nil
 }
 
 // A compile-time assertion to ensure RawKeyGenesisSigner meets the
@@ -801,20 +798,105 @@ var _ GenesisSigner = (*RawKeyGenesisSigner)(nil)
 
 // DeriveGroupKey derives an asset's group key based on an internal public
 // key descriptor, the original group asset genesis, and the asset's genesis.
-func DeriveGroupKey(genSigner GenesisSigner, rawKey keychain.KeyDescriptor,
-	initialGen Genesis, currentGen *Genesis) (*GroupKey, error) {
+func DeriveGroupKey(genSigner GenesisSigner, groupKey *GroupKey,
+	scriptLeaf *psbt.TaprootTapLeafScript, initialGen Genesis,
+	newAsset *Asset) (*GroupKey, error) {
 
-	groupPubKey, sig, err := genSigner.SignGenesis(
-		rawKey, initialGen, currentGen,
+	// First, perform the final checks on the asset being authorized for
+	// group membership.
+	if newAsset == nil || newAsset.IsUnknownVersion() {
+		return nil, fmt.Errorf("genesis asset is missing or unknown " +
+			"version")
+	}
+
+	if newAsset.Genesis != initialGen && initialGen.Type != newAsset.Type {
+		return nil, fmt.Errorf("asset group type mismatch")
+	}
+
+	if scriptLeaf != nil && groupKey.TapscriptRoot == [32]byte{} {
+		return nil, fmt.Errorf("cannot script spend group key with " +
+			"empty tapscript root")
+	}
+
+	// Populate the sign descriptor with static and default fields.
+	genesisTweak := initialGen.ID()
+	signDesc := &lndclient.SignDescriptor{
+		KeyDesc:       groupKey.RawKey,
+		SingleTweak:   genesisTweak[:],
+		DoubleTweak:   nil,
+		TapTweak:      []byte{},
+		WitnessScript: []byte{},
+		HashType:      txscript.SigHashDefault,
+		InputIndex:    0,
+	}
+	if groupKey.TapscriptRoot != [32]byte{} {
+		signDesc.TapTweak = groupKey.TapscriptRoot[:]
+	}
+
+	tweakedGroupKey, err := GroupPubKey(
+		signDesc.KeyDesc.PubKey, signDesc.SingleTweak,
+		signDesc.TapTweak,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("cannot tweak group key: %w", err)
+	}
+
+	// Reimplement tapscript.InputAssetPrevOut to avoid a dependency cycle.
+	prevOutScript, err := txscript.NewScriptBuilder().
+		AddOp(txscript.OP_1).
+		AddData(schnorr.SerializePubKey(tweakedGroupKey)).
+		Script()
+	prevOut := &wire.TxOut{
+		Value:    int64(newAsset.Amount),
+		PkScript: prevOutScript,
+	}
+	prevAssets := commitment.InputSet{
+		ZeroPrevID: {newAsset},
+	}
+
+	// Now, create the virtual transaction that represents this asset
+	// minting.
+	// TODO(jhb): resolve import cycles with tapscript package
+	virtualTx, _, err := tapscript.VirtualTx(newAsset, prevAssets)
+	if err != nil {
+		return nil, err
+	}
+	populatedVirtualTx := tapscript.VirtualTxWithInput(
+		virtualTx, newAsset, 0, nil,
+	)
+
+	// The last field to set for signing is the signing method needed, which
+	// can be inferred from the group key tapscript root and provided
+	// script leaf.
+	switch {
+	case bytes.Equal(signDesc.TapTweak, []byte{}):
+		signDesc.SignMethod = input.TaprootKeySpendBIP0086SignMethod
+	case scriptLeaf != nil:
+		signDesc.SignMethod = input.TaprootScriptSpendSignMethod
+		// TODO(jhb): extra validation here?
+		signDesc.WitnessScript = scriptLeaf.Script
+	case scriptLeaf == nil:
+		signDesc.SignMethod = input.TaprootKeySpendSignMethod
+	}
+
+	_, sig, err := genSigner.SignVirtualTx(
+		signDesc, populatedVirtualTx, prevOut,
 	)
 	if err != nil {
 		return nil, err
 	}
 
+	witness := wire.TxWitness{sig.Serialize()}
+	if signDesc.SignMethod == input.TaprootScriptSpendSignMethod {
+		witness = append(witness, signDesc.WitnessScript)
+		witness = append(witness, scriptLeaf.ControlBlock)
+	}
+
 	return &GroupKey{
-		RawKey:      rawKey,
-		GroupPubKey: *groupPubKey,
-		Witness:     wire.TxWitness{sig.Serialize()},
+		RawKey:        groupKey.RawKey,
+		GroupPubKey:   *tweakedGroupKey,
+		TapscriptRoot: groupKey.TapscriptRoot,
+		Witness:       witness,
 	}, nil
 }
 
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -244,6 +244,7 @@ type PrevID struct {
 }
 
 // Hash returns the SHA-256 hash of all items encapsulated by PrevID.
+// TODO(jhb): Is this safe to call on empty PrevIDs?
 func (id PrevID) Hash() [sha256.Size]byte {
 	h := sha256.New()
 	_ = wire.WriteOutPoint(h, 0, 0, &id.OutPoint)
@@ -959,6 +960,27 @@ func (a *Asset) HasGenesisWitness() bool {
 
 	witness := a.PrevWitnesses[0]
 	if witness.PrevID == nil || len(witness.TxWitness) > 0 ||
+		witness.SplitCommitment != nil {
+
+		return false
+	}
+
+	return *witness.PrevID == ZeroPrevID
+}
+
+// HasGenesisWitnessForGroup determines whether an asset has a valid witness
+// for a genesis asset in an asset group. This asset must have a non-empty group
+// key, and a single prevWitness with a zero PrevID, empty split commitment
+// proof, and non-empty witness.
+func (a *Asset) HasGenesisWitnessForGroup() bool {
+	if a.GroupKey == nil || len(a.PrevWitnesses) != 1 {
+		return false
+	}
+
+	// The single PrevWitness must have a ZeroPrevID, non-empty witness, and
+	// nil split commitment.
+	witness := a.PrevWitnesses[0]
+	if witness.PrevID == nil || len(witness.TxWitness) == 0 ||
 		witness.SplitCommitment != nil {
 
 		return false
--- a/asset/mock.go
+++ b/asset/mock.go
@@ -3,10 +3,17 @@ package asset
 import (
 	"bytes"
 	"encoding/hex"
+	"fmt"
 	"testing"
 
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/taproot-assets/internal/test"
 	"github.com/lightninglabs/taproot-assets/mssmt"
+	"github.com/lightningnetwork/lnd/input"
 	"github.com/stretchr/testify/require"
 )
 
@@ -53,6 +60,110 @@ func RandGroupKeyWithSigner(t testing.TB, genesis Genesis) (*GroupKey, []byte) {
 	require.NoError(t, err)
 
 	return groupKey, privateKey.Serialize()
+}
+
+// SignOutputRaw creates a signature for a single input.
+// Taken from lnd/lnwallet/btcwallet/signer:L344, SignOutputRaw
+func SignOutputRaw(priv *btcec.PrivateKey, tx *wire.MsgTx,
+	signDesc *input.SignDescriptor) (*schnorr.Signature, error) {
+
+	witnessScript := signDesc.WitnessScript
+
+	privKey := priv
+	var maybeTweakPrivKey *btcec.PrivateKey
+
+	switch {
+	case signDesc.SingleTweak != nil:
+		maybeTweakPrivKey = input.TweakPrivKey(
+			privKey, signDesc.SingleTweak,
+		)
+
+	case signDesc.DoubleTweak != nil:
+		maybeTweakPrivKey = input.DeriveRevocationPrivKey(
+			privKey, signDesc.DoubleTweak,
+		)
+
+	default:
+		maybeTweakPrivKey = privKey
+	}
+
+	privKey = maybeTweakPrivKey
+
+	// In case of a taproot output any signature is always a Schnorr
+	// signature, based on the new tapscript sighash algorithm.
+	if !txscript.IsPayToTaproot(signDesc.Output.PkScript) {
+		return nil, fmt.Errorf("mock signer: output script not taproot")
+	}
+
+	sigHashes := txscript.NewTxSigHashes(
+		tx, signDesc.PrevOutputFetcher,
+	)
+
+	// Are we spending a script path or the key path? The API is slightly
+	// different, so we need to account for that to get the raw signature.
+	var (
+		rawSig []byte
+		err    error
+	)
+	switch signDesc.SignMethod {
+	case input.TaprootKeySpendBIP0086SignMethod,
+		input.TaprootKeySpendSignMethod:
+
+		// This function tweaks the private key using the tap root key
+		// supplied as the tweak.
+		rawSig, err = txscript.RawTxInTaprootSignature(
+			tx, sigHashes, signDesc.InputIndex,
+			signDesc.Output.Value, signDesc.Output.PkScript,
+			signDesc.TapTweak, signDesc.HashType, privKey,
+		)
+
+	case input.TaprootScriptSpendSignMethod:
+		leaf := txscript.TapLeaf{
+			LeafVersion: txscript.BaseLeafVersion,
+			Script:      witnessScript,
+		}
+		rawSig, err = txscript.RawTxInTapscriptSignature(
+			tx, sigHashes, signDesc.InputIndex,
+			signDesc.Output.Value, signDesc.Output.PkScript,
+			leaf, signDesc.HashType, privKey,
+		)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return schnorr.ParseSignature(rawSig)
+}
+
+func SignVirtualTx(priv *btcec.PrivateKey, signDesc *lndclient.SignDescriptor,
+	tx *wire.MsgTx, prevOut *wire.TxOut) (*schnorr.Signature, error) {
+
+	prevOutFetcher := txscript.NewCannedPrevOutputFetcher(
+		prevOut.PkScript, prevOut.Value,
+	)
+
+	sigHashes := txscript.NewTxSigHashes(tx, prevOutFetcher)
+
+	fullSignDesc := input.SignDescriptor{
+		KeyDesc:           signDesc.KeyDesc,
+		SingleTweak:       signDesc.SingleTweak,
+		DoubleTweak:       signDesc.DoubleTweak,
+		TapTweak:          signDesc.TapTweak,
+		WitnessScript:     signDesc.WitnessScript,
+		SignMethod:        signDesc.SignMethod,
+		Output:            signDesc.Output,
+		HashType:          signDesc.HashType,
+		SigHashes:         sigHashes,
+		PrevOutputFetcher: prevOutFetcher,
+		InputIndex:        signDesc.InputIndex,
+	}
+
+	sig, err := SignOutputRaw(priv, tx, &fullSignDesc)
+	if err != nil {
+		return nil, err
+	}
+
+	return sig, nil
 }
 
 // RandScriptKey creates a random script key for testing.
--- a/tapscript/tx.go
+++ b/tapscript/tx.go
@@ -71,65 +71,99 @@ func virtualTxInPrevOut(root mssmt.Node) *wire.OutPoint {
 	)
 }
 
+// virtualGenesisTxIn computes the single input of a Taproot Asset virtual
+// transaction where the asset is both a genesis asset and part of an asset
+// group. The input MS-SMT commits to the genesis asset with witnesses removed.
+func virtualGenesisTxIn(newAsset *asset.Asset) (*wire.TxIn, mssmt.Tree, error) {
+	inputTree := mssmt.NewCompactedTree(mssmt.NewDefaultStore())
+
+	// TODO(bhandras): thread the context through.
+	ctx := context.TODO()
+
+	// For genesis assets with a group key, we must clear the witness field
+	// before building the input MS-SMT. We use this copy of the genesis
+	// asset as the input asset for the virtual transaction.
+	copyWithoutWitness := newAsset.Copy()
+	for i := range copyWithoutWitness.PrevWitnesses {
+		copyWithoutWitness.PrevWitnesses[i].TxWitness = nil
+	}
+
+	key := asset.ZeroPrevID.Hash()
+	leaf, err := copyWithoutWitness.Leaf()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	_, err = inputTree.Insert(ctx, key, leaf)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	treeRoot, err := inputTree.Root(context.Background())
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO(roasbeef): document empty hash usage here
+	prevOut := virtualTxInPrevOut(treeRoot)
+
+	return wire.NewTxIn(prevOut, nil, nil), inputTree, nil
+}
+
 // virtualTxIn computes the single input of a Taproot Asset virtual transaction.
 // The input prevout's hash is the root of a MS-SMT committing to all inputs of
 // a state transition.
 func virtualTxIn(newAsset *asset.Asset, prevAssets commitment.InputSet) (
 	*wire.TxIn, mssmt.Tree, error) {
 
-	// Genesis assets shouldn't have any inputs committed, so they'll have
-	// an empty input tree.
-	isGenesisAsset := newAsset.HasGenesisWitness()
 	inputTree := mssmt.NewCompactedTree(mssmt.NewDefaultStore())
-	if !isGenesisAsset {
-		// For each input we'll locate the asset UTXO beign spent, then
-		// insert that into a new SMT, with the key being the hash of
-		// the prevID pointer, and the value being the leaf itself.
-		inputsConsumed := make(
-			map[asset.PrevID]struct{}, len(prevAssets),
-		)
+	// For each input we'll locate the asset UTXO beign spent, then
+	// insert that into a new SMT, with the key being the hash of
+	// the prevID pointer, and the value being the leaf itself.
+	inputsConsumed := make(
+		map[asset.PrevID]struct{}, len(prevAssets),
+	)
 
-		// TODO(bhandras): thread the context through.
-		ctx := context.TODO()
+	// TODO(bhandras): thread the context through.
+	ctx := context.TODO()
 
-		for _, input := range newAsset.PrevWitnesses {
-			// At this point, each input MUST have a prev ID.
-			if input.PrevID == nil {
-				return nil, nil, ErrNoInputs
-			}
-
-			// The set of prev assets are similar to the prev
-			// output fetcher used in taproot.
-			prevAsset, ok := prevAssets[*input.PrevID]
-			if !ok {
-				return nil, nil, ErrNoInputs
-			}
-
-			// Now we'll insert this prev asset leaf into the tree.
-			// The generated leaf includes the amount of the asset,
-			// so the sum of this tree will be the total amount
-			// being spent.
-			key := input.PrevID.Hash()
-			leaf, err := prevAsset.Leaf()
-			if err != nil {
-				return nil, nil, err
-			}
-			_, err = inputTree.Insert(ctx, key, leaf)
-			if err != nil {
-				return nil, nil, err
-			}
-
-			inputsConsumed[*input.PrevID] = struct{}{}
+	for _, input := range newAsset.PrevWitnesses {
+		// At this point, each input MUST have a prev ID.
+		if input.PrevID == nil {
+			return nil, nil, ErrNoInputs
 		}
 
-		// In this context, the set of referenced inputs should match
-		// the set of previous assets. This ensures no duplicate inputs
-		// are being spent.
-		//
-		// TODO(roasbeef): make further explicit?
-		if len(inputsConsumed) != len(prevAssets) {
-			return nil, nil, ErrInputMismatch
+		// The set of prev assets are similar to the prev
+		// output fetcher used in taproot.
+		prevAsset, ok := prevAssets[*input.PrevID]
+		if !ok {
+			return nil, nil, ErrNoInputs
 		}
+
+		// Now we'll insert this prev asset leaf into the tree.
+		// The generated leaf includes the amount of the asset,
+		// so the sum of this tree will be the total amount
+		// being spent.
+		key := input.PrevID.Hash()
+		leaf, err := prevAsset.Leaf()
+		if err != nil {
+			return nil, nil, err
+		}
+		_, err = inputTree.Insert(ctx, key, leaf)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		inputsConsumed[*input.PrevID] = struct{}{}
+	}
+
+	// In this context, the set of referenced inputs should match
+	// the set of previous assets. This ensures no duplicate inputs
+	// are being spent.
+	//
+	// TODO(roasbeef): make further explicit?
+	if len(inputsConsumed) != len(prevAssets) {
+		return nil, nil, ErrInputMismatch
 	}
 
 	treeRoot, err := inputTree.Root(context.Background())
@@ -220,8 +254,19 @@ func virtualTxOut(asset *asset.Asset) (*wire.TxOut, error) {
 func VirtualTx(newAsset *asset.Asset, prevAssets commitment.InputSet) (
 	*wire.MsgTx, mssmt.Tree, error) {
 
+	var (
+		txIn      *wire.TxIn
+		inputTree mssmt.Tree
+		err       error
+	)
+
 	// We'll start by mapping all inputs into a MS-SMT.
-	txIn, inputTree, err := virtualTxIn(newAsset, prevAssets)
+	switch newAsset.HasGenesisWitnessForGroup() {
+	case true:
+		txIn, inputTree, err = virtualGenesisTxIn(newAsset)
+	case false:
+		txIn, inputTree, err = virtualTxIn(newAsset, prevAssets)
+	}
 	if err != nil {
 		return nil, nil, err
 	}
@@ -263,8 +308,19 @@ func InputAssetPrevOut(prevAsset asset.Asset) (*wire.TxOut, error) {
 	var pkScript []byte
 	switch prevAsset.ScriptVersion {
 	case asset.ScriptV0:
-		var err error
-		pkScript, err = PayToTaprootScript(prevAsset.ScriptKey.PubKey)
+		var (
+			err           error
+			validationKey = prevAsset.ScriptKey.PubKey
+		)
+
+		// If the input asset is a genesis asset that is part of an
+		// asset group, we need to validate the group witness against
+		// the tweaked group key and not the genesis asset script key.
+		if prevAsset.HasGenesisWitnessForGroup() {
+			validationKey = &prevAsset.GroupKey.GroupPubKey
+		}
+
+		pkScript, err = PayToTaprootScript(validationKey)
 		if err != nil {
 			return nil, err
 		}
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -191,15 +191,24 @@ func (vm *Engine) validateWitnessV0(virtualTx *wire.MsgTx, inputIdx uint32,
 		return ErrInvalidScriptVersion
 	}
 
-	// An input MUST have a prev out and also a valid witness.
-	if witness.PrevID == nil || len(witness.TxWitness) == 0 {
+	// An input must have a valid witness.
+	if len(witness.TxWitness) == 0 {
 		return newErrKind(ErrInvalidTransferWitness)
 	}
 
-	// The parameters of the new and old asset much match exactly.
-	err := matchesAssetParams(vm.newAsset, prevAsset, witness)
-	if err != nil {
-		return err
+	// Genesis grouped assets will have a nil PrevID and match the prevAsset
+	// since it is a copy of the original asset.
+	if !vm.newAsset.HasGenesisWitnessForGroup() {
+		// An input MUST have a prev out and also a valid witness.
+		if witness.PrevID == nil {
+			return newErrKind(ErrInvalidTransferWitness)
+		}
+
+		// The parameters of the new and old asset much match exactly.
+		err := matchesAssetParams(vm.newAsset, prevAsset, witness)
+		if err != nil {
+			return err
+		}
 	}
 
 	// Update the virtual transaction input with details for the specific
@@ -288,6 +297,21 @@ func (vm *Engine) Execute() error {
 			return newErrKind(ErrInvalidGenesisStateTransition)
 		}
 		return nil
+	}
+
+	// Genesis assets in an asset group have a witness that must be
+	// verified to prove group membership.
+	// TODO(jhb): Set group witness in PrevWitnesses
+	if vm.newAsset.HasGenesisWitnessForGroup() {
+		if len(vm.splitAssets) > 0 || len(vm.prevAssets) > 0 {
+			return newErrKind(ErrInvalidGenesisStateTransition)
+		}
+
+		// For genesis assets in an asset group, set the previous asset
+		// as the genesis asset.
+		vm.prevAssets = commitment.InputSet{
+			asset.ZeroPrevID: vm.newAsset,
+		}
 	}
 
 	// If we have an asset split, then we need to validate the state
--- a/tapgarden/mock.go
+++ b/tapgarden/mock.go
@@ -452,14 +452,18 @@ func NewMockGenSigner(keyRing *MockKeyRing) *MockGenSigner {
 	}
 }
 
-func (m *MockGenSigner) SignGenesis(desc keychain.KeyDescriptor,
-	initialGen asset.Genesis, currentGen *asset.Genesis) (*btcec.PublicKey,
+func (m *MockGenSigner) SignVirtualTx(signDesc *lndclient.SignDescriptor,
+	virtualTx *wire.MsgTx, prevOut *wire.TxOut) (*btcec.PublicKey,
 	*schnorr.Signature, error) {
 
-	priv := m.KeyRing.Keys[desc.KeyLocator]
+	priv := m.KeyRing.Keys[signDesc.KeyDesc.KeyLocator]
 	signer := asset.NewRawKeyGenesisSigner(priv)
-	return signer.SignGenesis(desc, initialGen, currentGen)
+	return signer.SignVirtualTx(signDesc, virtualTx, prevOut)
 }
+
+// A compile-time assertion to ensure MockGenSigner meets the GenesisSigner
+// interface.
+var _ asset.GenesisSigner = (*MockGenSigner)(nil)
 
 type MockProofArchive struct {
 }
--- a/tapscript/mock.go
+++ b/tapscript/mock.go
@@ -1,14 +1,11 @@
 package tapscript
 
 import (
-	"fmt"
-
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
-	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/lndclient"
-	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightninglabs/taproot-assets/asset"
 )
 
 type MockSigner struct {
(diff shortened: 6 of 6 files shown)


jharveyb:
I don't see a neat solution to the dependency cycle issues, so just moving some of the virtual TX creation logic for these special group genesis assets into `asset` and accepting some duplication with `tapscript` to keep this moving forward.

asset/asset.go
```diff
@@ -967,6 +968,27 @@ func (a *Asset) HasGenesisWitness() bool {
 	return *witness.PrevID == ZeroPrevID
 }
 
+// HasGenesisWitnessForGroup determines whether an asset has a valid witness
```
halseth, line 1053: remove use of word "valid" perhaps? Since we are not checking the signature

tapscript/tx.go
```diff
@@ -71,65 +71,99 @@ func virtualTxInPrevOut(root mssmt.Node) *wire.OutPoint {
...
+
+	// For genesis assets with a group key, we must clear the witness field
+	// before building the input MS-SMT. We use this copy of the genesis
+	// asset as the input asset for the virtual transaction.
+	copyWithoutWitness := newAsset.Copy()
+	for i := range copyWithoutWitness.PrevWitnesses {
+		copyWithoutWitness.PrevWitnesses[i].TxWitness = nil
+	}
+
+	key := asset.ZeroPrevID.Hash()
```
halseth, line 91: Yeah, this looks correct I think! During VM execution the prev witness will be zero and point back to this zero-id.

tapscript/tx.go
```diff
@@ -220,8 +254,19 @@ func virtualTxOut(asset *asset.Asset) (*wire.TxOut, error) {
...
 
+	var (
+		txIn      *wire.TxIn
+		inputTree mssmt.Tree
+		err       error
+	)
+
 	// We'll start by mapping all inputs into a MS-SMT.
-	txIn, inputTree, err := virtualTxIn(newAsset, prevAssets)
+	switch newAsset.HasGenesisWitnessForGroup() {
```
halseth, line 264: hehe, just `if/else`? 😛

tapscript/tx.go
```diff
@@ -220,8 +254,19 @@ func virtualTxOut(asset *asset.Asset) (*wire.TxOut, error) {
...
+	case true:
+		txIn, inputTree, err = virtualGenesisTxIn(newAsset)
```
halseth, line 266: Very nice, diff was not as large as I expected 👍

tapscript/tx.go
```diff
@@ -263,8 +308,19 @@ func InputAssetPrevOut(prevAsset asset.Asset) (*wire.TxOut, error) {
...
-		pkScript, err = PayToTaprootScript(prevAsset.ScriptKey.PubKey)
+		var (
+			err           error
+			validationKey = prevAsset.ScriptKey.PubKey
+		)
+
+		// If the input asset is a genesis asset that is part of an
+		// asset group, we need to validate the group witness against
+		// the tweaked group key and not the genesis asset script key.
+		if prevAsset.HasGenesisWitnessForGroup() {
```
halseth, line 319: My idea here was to set the `prevAsset.ScriptKey = groupKey` when creating the `prevAsset` copy. Is that feasible?

halseth (commented):
Def looks to be on the right track! 😀

tapscript/tx.go
jharveyb: I thought we don't do else 😅  Can change.

tapscript/tx.go
jharveyb: Covered here:

https://github.com/Roasbeef/bips/pull/33#discussion_r1259002226

I think this still needs to get reflected in the BIP text. AFAICT, the `pkScript` here is what we actually pass to the internal `txscript.Engine` to validate the witness against, so it's the one we need to now be the group key. The asset used to build the virtual TX should have the tweaked group key set, but not swapped for the script key.

jharveyb:
> What do you mean by this? The script path auth is just the existing witness field right? Or do you mean the group key derivation reveal? The group key witness (for minting) should be more or less the same as the flow for signing virtual txns normally. IIUC, we don't need any _new_ fields in the `VInput`, instead the existing fields are populated as normal. The main difference now is that we swap out the public key (not the input asset) for the group key asset. 

From the RPC level, what data types to I need to provide as arguments when minting a grouped asset into a group, where I want to use a script path and not a simple signature?

And then, how does that get passed in to the final signing descriptor (I think this part is clearer to me now).

Roasbeef:
> From the RPC level, what data types to I need to provide as arguments when minting a grouped asset into a group, where I want to use a script path and not a simple signature?

Ah, I think for that, we'd just make sure we're able to handle such spends internally via unit tests, then defer the expansion of the RPC interface to satisfy more complex group minting conditions. So to start, we can just allow them to specify their own group key and we use that instead of the auto generated one. The caller would then need to handle storing all the tapcript leaves, just like the itests we have to do non keyspend spends for normal assets. 

With that said, if we wanted to do some design sketches, I'd say: 
> what data types to I need to provide as arguments when minting a grouped asset into a group, where I want to use a script path and not a simple signature?

You'd need to specify:
  * The group key to use.
  * The full witness field for that group key. 

We can't necessarily generate the witness ourselves, as we may not even control the key used. Instead, the caller would use the `vPSBT` APIs to craft the minting transaction. To facilitate this, we'd need to expose calls that can be used to lock UTXOs (they want to know which outpoint will be used), or we'd add that as an arg to the minting process. 

> And then, how does that get passed in to the final signing descriptor (I think this part is clearer to me now).

We don't do signing in this context, as we might not be able to. The minting witness is just the witness of a "normal" spend, but it can spend the "all zero" outpoint, as it's a proof that the group key _can_ be satisfied. You can think of it like the PSBT based channel funding for lnd: we don't know what inputs the caller is using, but we can provide them with the information needed to make the funding transaction (the funding outpoint).

vm/vm.go
```diff
@@ -290,6 +299,21 @@ func (vm *Engine) Execute() error {
...
+	// TODO(jhb): Set group witness in PrevWitnesses
+	if vm.newAsset.HasGenesisWitnessForGroup() {
+		if len(vm.splitAssets) > 0 || len(vm.prevAssets) > 0 {
+			return newErrKind(ErrInvalidGenesisStateTransition)
+		}
+
+		// For genesis assets in an asset group, set the previous asset
+		// as the genesis asset.
+		vm.prevAssets = commitment.InputSet{
+			asset.ZeroPrevID: vm.newAsset,
```
Roasbeef, line 313: 👍

asset/asset.go
```diff
@@ -469,10 +473,21 @@ func (g *GroupKeyReveal) GroupPubKey(assetID ID) (*btcec.PublicKey, error) {
 		return nil, err
 	}
 
-	internalKey := input.TweakPubKeyWithTweak(rawKey, assetID[:])
+	return GroupPubKey(rawKey, assetID[:], g.TapscriptRoot[:])
+}
+
+func GroupPubKey(rawKey *btcec.PublicKey, singleTweak, tapTweak []byte) (
```
Roasbeef, line 479: Missing `godoc`` comment.

asset/asset.go
```diff
@@ -727,14 +742,11 @@ func NewScriptKeyBip86(rawKey keychain.KeyDescriptor) ScriptKey {
...
-	// descriptor with the the first passed Genesis description, and signs
-	// the second passed Genesis description with the tweaked public key.
-	// For minting the first asset in a group, only one Genesis object is
-	// needed, since we tweak with and sign over the same Genesis object.
-	// The final tweaked public key and the signature are returned.
-	SignGenesis(keychain.KeyDescriptor, Genesis,
-		*Genesis) (*btcec.PublicKey, *schnorr.Signature, error)
+	// SignVirtualTx generates a signature according to the passed signing
+	// descriptor and TX.
+	SignVirtualTx(signDesc *lndclient.SignDescriptor, tx *wire.MsgTx,
```
Roasbeef, line 747: For our purposes, I think we can assume it's just keyspends for now, so we can actually keep the existing `KeyDescriptor`? Still getting thru the diff though.

Roasbeef: Alternatively, we can rely on this method to produce the virtual tx itself. So it just needs the `Genesis` for the "root" asset, then the asset being created (`asset.Asset`), then it returns the witness as is. Also I don't think the first return value is valid any longer with the new scheme.

asset/asset.go
```diff
@@ -757,42 +769,27 @@ func NewRawKeyGenesisSigner(priv *btcec.PrivateKey) *RawKeyGenesisSigner {
...
-	}
+	sig, err := SignVirtualTx(r.privKey, signDesc, virtualTx, prevOut)
 
-	// TODO(roasbeef): this actually needs to sign the digest of the asset
-	// itself
-	idHash := sha256.Sum256(id[:])
-	sig, err := schnorr.Sign(tweakedPrivKey, idHash[:])
+	// Derive the final tweaked public key also used for signing.
+	tweakedGroupKey, err := GroupPubKey(
+		signerPubKey, signDesc.SingleTweak, signDesc.TapTweak,
```
Roasbeef, line 786: See above, I think we can keep the sign desc as is, then it just derives the tweaks from: the asset passed in, and the genesis to use for signing purposes. We can also tweak that slightly to be `GroupKey` or w/e (assuming it includes the key and the genesis information).

Roasbeef: This way the interface is a bit more opinionated.

asset/asset.go
```diff
@@ -757,42 +769,27 @@ func NewRawKeyGenesisSigner(priv *btcec.PrivateKey) *RawKeyGenesisSigner {
...
-	// of the first asset in the group.
-	id := initialGen.ID()
-	if currentGen != nil {
-		if initialGen.Type != currentGen.Type {
-			return nil, nil, fmt.Errorf("asset group type mismatch")
-		}
-
-		id = currentGen.ID()
-	}
+	sig, err := SignVirtualTx(r.privKey, signDesc, virtualTx, prevOut)
```
Roasbeef, line 782: Error is ignored here.

asset/asset.go
```diff
@@ -801,20 +798,105 @@ var _ GenesisSigner = (*RawKeyGenesisSigner)(nil)
...
+	}
+
 	return &GroupKey{
-		RawKey:      rawKey,
-		GroupPubKey: *groupPubKey,
-		Witness:     wire.TxWitness{sig.Serialize()},
+		RawKey:        groupKey.RawKey,
+		GroupPubKey:   *tweakedGroupKey,
+		TapscriptRoot: groupKey.TapscriptRoot,
+		Witness:       witness,
```
Roasbeef, line 899: Why do we need to attach a witness here? The witness would reside in the asset being created (the very first one), then we just need to care about the key and tapscript root. Then this can be re-used everywhere, with the group minting identified based on the asset itself (prev id zero, etc). 

Or seems this is both the initial derivation _and_ also the first signing? 

For group keys in assets, we don't actually need to pass that witness along each time for serialization, so I assume this is just for book keeping or convenience purpose?

asset/asset.go
```diff
@@ -801,20 +798,105 @@ var _ GenesisSigner = (*RawKeyGenesisSigner)(nil)
 
 // DeriveGroupKey derives an asset's group key based on an internal public
 // key descriptor, the original group asset genesis, and the asset's genesis.
-func DeriveGroupKey(genSigner GenesisSigner, rawKey keychain.KeyDescriptor,
-	initialGen Genesis, currentGen *Genesis) (*GroupKey, error) {
+func DeriveGroupKey(genSigner GenesisSigner, groupKey *GroupKey,
+	scriptLeaf *psbt.TaprootTapLeafScript, initialGen Genesis,
```
Roasbeef, line 802: I think we can use the `AssetGroup` struct in lieu of the group key, then the initial genesis?

Roasbeef, line 802: Also why `scriptLeaf` vs the actual script root? We may not know what the leaves are (for now caller will handle the external witness signing, etc).

asset/asset.go
```diff
@@ -801,20 +798,105 @@ var _ GenesisSigner = (*RawKeyGenesisSigner)(nil)
...
+	if newAsset == nil || newAsset.IsUnknownVersion() {
+		return nil, fmt.Errorf("genesis asset is missing or unknown " +
+			"version")
+	}
+
+	if newAsset.Genesis != initialGen && initialGen.Type != newAsset.Type {
+		return nil, fmt.Errorf("asset group type mismatch")
+	}
+
+	if scriptLeaf != nil && groupKey.TapscriptRoot == [32]byte{} {
```
Roasbeef, line 816: At this phase, I don't think we can really assert much about the script leaf. 

For now we can handle just the keyspend path. Anything more advanced than that (musig2, etc, etc) will require the caller to drive the API to do the fancy stuff.

asset/asset.go
```diff
@@ -801,20 +798,105 @@ var _ GenesisSigner = (*RawKeyGenesisSigner)(nil)
...
+	}
+
+	// Populate the sign descriptor with static and default fields.
+	genesisTweak := initialGen.ID()
+	signDesc := &lndclient.SignDescriptor{
+		KeyDesc:       groupKey.RawKey,
+		SingleTweak:   genesisTweak[:],
+		DoubleTweak:   nil,
+		TapTweak:      []byte{},
+		WitnessScript: []byte{},
```
Roasbeef, line 828: If the script leaf was needed (don't think it is), then the actual script would be placed here.

asset/asset.go
```diff
@@ -801,20 +798,105 @@ var _ GenesisSigner = (*RawKeyGenesisSigner)(nil)
...
+		HashType:      txscript.SigHashDefault,
+		InputIndex:    0,
+	}
+	if groupKey.TapscriptRoot != [32]byte{} {
+		signDesc.TapTweak = groupKey.TapscriptRoot[:]
```
Roasbeef, line 833: So this is only actually needed when doing a key path spend.

asset/asset.go
```diff
@@ -801,20 +798,105 @@ var _ GenesisSigner = (*RawKeyGenesisSigner)(nil)
...
+
+	tweakedGroupKey, err := GroupPubKey(
+		signDesc.KeyDesc.PubKey, signDesc.SingleTweak,
+		signDesc.TapTweak,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("cannot tweak group key: %w", err)
+	}
+
+	// Reimplement tapscript.InputAssetPrevOut to avoid a dependency cycle.
```
Roasbeef, line 844: Leaving comment here so we make sure it's addressed. Alternatively, we can just move this function out of the `asset` package and into `tapscript`?

Roasbeef (changes requested):

tapscript/tx.go
halseth: See my comment linked on the BIP:

> The modification only affects the referenced previous leaf, that is used to construct the previous outpoint for the virtual tx. The output leaf (still committing to asset_script_key) remains the same.

asset/asset.go
jharveyb: In the minter, we set the group key info of the asset based on the output of `DeriveGroupKey`. I think ideally we'd modify `asset.New()` to copy the witness from here passed `GroupKey` into the genesis asset witness for the new asset.

asset/asset.go
jharveyb: Yeah the `GroupKey` was intended to carry an optional tapscript root, but if that's no longer needed then `AssetGroup` does make more sense.

asset/asset.go
jharveyb: So sounds like for now we're only going to support signing for group keys with empty tapscript roots?

asset/asset.go
```diff
@@ -967,6 +968,27 @@ func (a *Asset) HasGenesisWitness() bool {
 	return *witness.PrevID == ZeroPrevID
 }
 
+// HasGenesisWitnessForGroup determines whether an asset has a valid witness
+// for a genesis asset in an asset group. This asset must have a non-empty group
+// key, and a single prevWitness with a zero PrevID, empty split commitment
+// proof, and non-empty witness.
+func (a *Asset) HasGenesisWitnessForGroup() bool {
+	if a.GroupKey == nil || len(a.PrevWitnesses) != 1 {
```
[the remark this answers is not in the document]
jharveyb: I think this still needs some change, we actually have two new cases:

-An asset with the existing genesis asset witness, that also has the group key set (needs a witness generated).
-An grouped genesis asset with a witness (empty PrevID, nil SplitCommitment, non-nil TxWitness, non-nil GroupKey), should still be considered a genesis asset

tapscript/tx.go
```diff
@@ -71,65 +71,99 @@ func virtualTxInPrevOut(root mssmt.Node) *wire.OutPoint {
...
 // The input prevout's hash is the root of a MS-SMT committing to all inputs of
 // a state transition.
 func virtualTxIn(newAsset *asset.Asset, prevAssets commitment.InputSet) (
 	*wire.TxIn, mssmt.Tree, error) {
 
-	// Genesis assets shouldn't have any inputs committed, so they'll have
-	// an empty input tree.
-	isGenesisAsset := newAsset.HasGenesisWitness()
 	inputTree := mssmt.NewCompactedTree(mssmt.NewDefaultStore())
-	if !isGenesisAsset {
```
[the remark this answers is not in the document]
jharveyb: Yes, this point was never reached with a genesis asset before because the VM would have exited early on any genesis assets before this line.

Now, we have `virtualGenesisTxIn` and a genesis asset should still never appear at this point, so the branch is unneeded.

jharveyb:
Read through the feedback, my current plan is:

- Move the virtual TX construction logic behind another interface, `asset.GenesisTxBuilder`, to resolve dependency issues.
- Simplify `DeriveGroupKey` to only handle the key spend path with an empty tapscript root.
- New methods on `Asset` so its clearer when a grouped genesis asset needs a witness or already has a witness.
- Update `asset.new()` or add logic to the minter s.t. the group key witness is copied into the asset during minting.

asset/asset.go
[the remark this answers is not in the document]
jharveyb: This has been addressed with two new methods, `NeedsGenesisWitnessForGroup` and `HasGenesisWitnessForGroup`.

The `Needs` variant covers an asset with a genesis asset witness, and a group key set. This is used as a sanity check within the `GenesisTxBuilder` implementation, and to branch on in the virtual TX construction logic.

The `Has` variant is used in the VM and the proof verification logic, since at those points any genesis grouped asset should have its group witness attached, and now it only needs to be validated. At those points an asset with an empty group witness would be considered invalid.

asset/asset.go
jharveyb: Resolved offline.

asset/asset.go
jharveyb: The actual group witness copy into the genesis asset witness is here in `asset.New()`:

https://github.com/lightninglabs/taproot-assets/blob/843584a12a7bcefd85003f584c24fe48e8705713/asset/asset.go#L951

jharveyb:
Latest version is on this branch with the described changes. The commits were re-done and should be relatively readable.

Moving on to improving test coverage.

jharveyb:
Another source of dependency issues:

https://github.com/lightninglabs/taproot-assets/blob/843584a12a7bcefd85003f584c24fe48e8705713/commitment/asset.go#L151

Verifying the group witness on asset commitment insertion.

Roasbeef:
> Verifying the group witness on asset commitment insertion.

So I think we can actually move this out entirely. Looks like the only time we call `VerifySignature` is `Upsert` and also when we make a new asset commitment. Now that we have the group key reveal stuff in the proof file verification, we can just move that there. So now we only verify the signature when we create assets, and also when we verify proofs for assets.

asset/mock.go
```diff
@@ -61,6 +65,160 @@ func RandGroupKeyWithSigner(t testing.TB, genesis Genesis) (*GroupKey, []byte) {
...
+			PkScript: pkScript,
+		}, nil
+	default:
+		return nil, ErrUnknownVersion
+	}
+}
+
+// Forked from tapscript/tx/virtualTxIn to remove checks for non-nil PrevIDs on
+// inputs and full consumption of inputs.
+func VirtualGenesisTxIn(newAsset *Asset) (*wire.TxIn, mssmt.Tree, error) {
```
Roasbeef, line 101: I think we can move this out of `mock.go`, so residing in like `asset/group_genesis.go` or something like that. After we get things all working we maybe just want to move it purely within `txscript`, then we keep the `asset` package concerned with mainly serialization and initial sanity checks.

tapscript/mint.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/taproot-assets/asset"
+)
+
+// MockGroupTxBuilder is an implementation of the asset.GenesisTxBuilder
+// interface that constructs virtual transactions for grouped asset genesis.
+type BaseGroupTxBuilder struct{}
```
Roasbeef, line 12: godoc name mismatch?

asset/asset.go
```diff
@@ -756,47 +743,61 @@ func NewRawKeyGenesisSigner(priv *btcec.PrivateKey) *RawKeyGenesisSigner {
...
+}
+
+// A compile-time assertion to ensure RawKeyGenesisSigner meets the
+// GenesisSigner interface.
+var _ GenesisSigner = (*RawKeyGenesisSigner)(nil)
+
+type RawGroupTxBuilder struct{}
 
-		id = currentGen.ID()
+func (m *RawGroupTxBuilder) BuildGenesisTx(newAsset *Asset) (*wire.MsgTx,
```
Roasbeef, line 786: Does this version need to live in the `asset` package? Can we get away with just the version in `tapscript`?

tapscript/mint.go
```diff
@@ -0,0 +1,50 @@
...
+
+func BuildGenesisTx(newAsset *asset.Asset) (*wire.MsgTx,
```
Roasbeef, line 14: Missing a `godoc` template.

virtual_tx_builder.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/taproot-assets/asset"
+	"github.com/lightninglabs/taproot-assets/tapscript"
+)
+
+// GroupTxBuilder is an implementation of the asset.GenesisTxBuilder interface
+// that constructs virtual transactions for grouped asset genesis.
+type GroupTxBuilder struct{}
+
+func (b *GroupTxBuilder) BuildGenesisTx(newAsset *asset.Asset) (*wire.MsgTx,
```
Roasbeef, line 13: Looks like the 3rd instance of this in the diff? Wondering if we can consolidate somewhat. Might need to modify the structure of the mock packages and/or tests.

vm/vm.go
```diff
@@ -191,15 +191,24 @@ func (vm *Engine) validateWitnessV0(virtualTx *wire.MsgTx, inputIdx uint32,
...
+	if len(witness.TxWitness) == 0 {
 		return newErrKind(ErrInvalidTransferWitness)
 	}
 
-	// The parameters of the new and old asset much match exactly.
-	err := matchesAssetParams(vm.newAsset, prevAsset, witness)
-	if err != nil {
-		return err
+	// Genesis grouped assets will have a nil PrevID and match the prevAsset
+	// since it is a copy of the original asset.
```
Roasbeef, line 200: Maybe rewrite comment like: 
> For assets w/o a group witness, we want to check the prev asset to make sure it's spending like-for-like.

asset/asset.go
```diff
@@ -468,10 +469,25 @@ func (g *GroupKeyReveal) GroupPubKey(assetID ID) (*btcec.PublicKey, error) {
 		return nil, err
 	}
 
-	internalKey := input.TweakPubKeyWithTweak(rawKey, assetID[:])
+	if g.TapscriptRoot != nil {
```
Roasbeef, line 472: Can we just collapse this? So just calling it with nil directly.

asset/asset.go
```diff
@@ -468,10 +469,25 @@ func (g *GroupKeyReveal) GroupPubKey(assetID ID) (*btcec.PublicKey, error) {
...
+		return GroupPubKey(rawKey, assetID[:], g.TapscriptRoot[:])
+	}
+
+	return GroupPubKey(rawKey, assetID[:], nil)
```
Roasbeef, line 476: `[]byte{}`?

asset/asset.go
```diff
@@ -468,10 +469,25 @@ func (g *GroupKeyReveal) GroupPubKey(assetID ID) (*btcec.PublicKey, error) {
...
+}
+
+func GroupPubKey(rawKey *btcec.PublicKey, singleTweak, tapTweak []byte) (
+	*btcec.PublicKey, error) {
+
+	if len(singleTweak) != sha256.Size || len(tapTweak) != sha256.Size {
```
Roasbeef, line 482: `tapTweak` can be `nil`, if there's no tapscript root.

asset/asset.go
```diff
@@ -801,19 +817,58 @@ var _ GenesisTxBuilder = (*RawGroupTxBuilder)(nil)
...
+	}
+
+	if initialGen.Type != newAsset.Type {
+		return nil, fmt.Errorf("asset group type mismatch")
+	}
+
+	// Compute the tweaked group key and set it in the asset before
+	// creating the virtual minting transaction.
+	genesisTweak := initialGen.ID()
+	tweakedGroupKey, err := GroupPubKey(rawKey.PubKey, genesisTweak[:], nil)
```
Roasbeef, line 837: Eventually should also have the script root here. Could have it be a functional option or something like that for now.

asset/asset.go
```diff
@@ -801,19 +817,58 @@ var _ GenesisTxBuilder = (*RawGroupTxBuilder)(nil)
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot build virtual tx: %w", err)
+	}
+
+	// Build the static signing descriptor needed to sign the virtual
+	// minting transaction. This is restricted to group keys with an empty
+	// tapscript root and key path spends.
+	signDesc := &lndclient.SignDescriptor{
+		KeyDesc:     rawKey,
+		SingleTweak: genesisTweak[:],
```
Roasbeef, line 859: Don't have the double tweak rn as the tapscript layer isn't there?

asset/asset.go
```diff
@@ -939,18 +939,26 @@ func New(genesis Genesis, amount, locktime, relativeLocktime uint64,
...
+	// Valid genesis asset witness.
+	genesisWitness := Witness{
+		PrevID:          &PrevID{},
+		TxWitness:       nil,
+		SplitCommitment: nil,
+	}
+
+	// Genesis assets with an asset group must have the group witness stored
+	// in the genesis asset witness, if present.
+	if groupKey != nil && groupKey.Witness != nil {
```
Roasbeef, line 951: Still think we don't necessarily need the witness field within the `GroupKey` itself. I can see how it's convenient here though since we're already accepting in the group key.

Roasbeef (changes requested):

asset/mock.go
jharveyb: Moved in latest version out to `asset/asset` near the other group logic - I agree that we probably want to rearrange these implementations overall.

jharveyb:
Addressed a few comments, updated this main branch and the WIP branch with unfinished test fixes here:

https://github.com/lightninglabs/taproot-assets/tree/group-key-witness-support-wip

There was an issue with not using the correct branching condition in `VirtualTx` for reaching `VirtualGenesisTxIn` - since `VirtualTx` is called both when building the Virtual TX for signing, and when verifying in the VM, I think we'd actually want to branch on the genesis grouped asset both before and after the group witness has been attached.

However, VM validation for normal state transitions is now failing, so I think the branch condition as I've written it is too broad - I think this could be fixed by also considering `prevAssets` at that point. So something like:

If both the new asset and the prev asset are genesis grouped assets, then it must be a genesis grouped asset mint, branch to VirtualGenesisTxIn.

We still have the sanity check very early in the VM that I think should stay as `HasGenesisWitnessForGroup`.

The WIP branch also has some fixes for other tests but those haven't been exercised yet since they usually depend on the VM working anyways.

asset/asset.go
jharveyb: Moved out of scope for this PR.

asset/asset.go
jharveyb: Not really, since that would cause dependency cycles for many tests that use `RandGroupKey()`. Similar to how we had the raw key genesis signer and the real `gen_signer`.

asset/asset.go
guggero: Also, should we use `len(groupKey.Witness) != 0` to avoid this being true for empty witnesses?

tapgarden/caretaker.go
```diff
@@ -536,6 +536,15 @@ func (b *BatchCaretaker) seedlingsToAssetSprouts(ctx context.Context,
 				err)
 		}
 
+		// Verify the group witness if present.
+		if sproutGroupKey != nil {
+			err := b.cfg.TxValidator.Execute(newAsset, nil, nil)
+			if err != nil {
+				return nil, fmt.Errorf("unable to verify"+
```
guggero, line 543: nit: missing space after "verify".

commitment/commitment_test.go
```diff
@@ -342,28 +332,41 @@ func TestMintTapCommitment(t *testing.T) {
...
+			// Test case concerns amount sanity checking, skip
+			// group key creation.
+			var err error
+			amt := details.Amount
+			invalidNormalAmt := details.Type != asset.Collectible &&
+				((amt != nil && *amt == uint64(0)) || amt == nil)
+			invalidCollctibleAmt := amt != nil && *amt != uint64(1) &&
+				details.Type == asset.Collectible
+
+			switch invalidNormalAmt || invalidCollctibleAmt {
```
guggero, line 345: nit: same here re true/false switch statements.

commitment/commitment_test.go
```diff
@@ -342,28 +332,41 @@ func TestMintTapCommitment(t *testing.T) {
...
-			protoAsset, err := asset.New(
-				testCase.g, amt, details.LockTime,
-				details.RelativeLockTime,
-				asset.NewScriptKeyBip86(details.ScriptKey), nil,
-			)
+			// Test case concerns amount sanity checking, skip
+			// group key creation.
+			var err error
+			amt := details.Amount
+			invalidNormalAmt := details.Type != asset.Collectible &&
```
guggero, line 340: We could create two package-level variables `zero` and `one` to make it easier to use with the collectible amounts.

vm/vm.go
```diff
@@ -191,24 +191,37 @@ func (vm *Engine) validateWitnessV0(virtualTx *wire.MsgTx, inputIdx uint32,
...
+		prevOutFetcher, err = asset.GenesisPrevOutFetcher(*prevAsset)
+
+	case false:
+		// An input MUST have a prev out and also a valid witness.
+		if witness.PrevID == nil {
+			return newErrKind(ErrInvalidTransferWitness)
+		}
+
+		// The parameters of the new and old asset much match exactly.
+		err := matchesAssetParams(vm.newAsset, prevAsset, witness)
```
guggero, line 218: This overwrites the outer `err`, making the comparison below incorrect (this is what the linter is complaining about).

guggero (commented):
Great work, very nice to see this getting close.

asset/mock.go
```diff
@@ -61,6 +63,87 @@ func RandGroupKeyWithSigner(t testing.TB, genesis Genesis) (*GroupKey, []byte) {
 
 	return groupKey, privateKey.Serialize()
 }
+
+// Copied from tapscript/tx.go.
+func virtualGenesisTxWithInput(virtualTx *wire.MsgTx, input *Asset,
```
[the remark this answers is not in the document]
jharveyb: As in move them from tapscript and get rid of the redunant mock ones? I'm in favor.

tapscript/mint.go
```diff
@@ -0,0 +1,50 @@
...
+	*wire.TxOut, error) {
+
+	// First, we check that the passed asset is a genesis grouped asset
+	// that has no group witness.
+	if !newAsset.NeedsGenesisWitnessForGroup() {
+		return nil, nil, fmt.Errorf("asset is not a genesis grouped" +
```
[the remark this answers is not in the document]
jharveyb: Mm if we move the required TX utilities to asse to think we could have one implementation there and then the top-level wrapper, and get rid of the wrapper in the tapscript package.

commitment/commitment_test.go
```diff
@@ -329,8 +342,28 @@ func TestMintTapCommitment(t *testing.T) {
...
+			// amount of 1 here, and pass the original amount to
+			// Min() below.
+			var amt uint64
+			if details.Amount == nil {
+				amt = 1
+			} else {
+				amt = *details.Amount
+			}
+
+			// TOO(jhb): fix nil amount issues, use this err
```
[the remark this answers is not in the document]
jharveyb: I consider it resolved by the logic below or skipping the group key logic but forgot to remove the TODO itself.

jharveyb:
Latest push is further fixes to address some unit test fixes + fix group insertion when syncing from a Universe (h/t guggero).

Most itests pass, but reissuance does not work - adding a `GroupVerifier` interface now to be used when verifying proofs, to check if there is a verified group that matches the group key in the proof.

jharveyb:
Adding a `groupVerifier` has worked for fixing the reissuance test.

From offline discussion, I realized that the GroupKeyReveal should be omitted for genesis proofs of reissued assets, and instead the asset's group key should only be checked against stored group keys to ensure that the daemon has previously verified the group anchor (along with group witness verification).

However, this introduces new ordering constraints in many places where we handle grouped assets (storing to disk, verifying a proof, or populating a universe tree). The anchor asset for a group MUST be inserted before attempting to verify or store any other member of that asset group.

Right now many DB operations are unordered over e.x. the set of all assets in a minting batch, which I'm updating now.

jharveyb:
All itests are now passing locally & with the current state of this PR. Latest changes were to sort assets such that group anchors are inserted on disk before any possible group reissuances, since group key verification now relies on the group anchor being present on disk.

Pending changes are addressing reviewer feedback and changing the `GroupKey.TapscriptRoot` member from `[32]byte` to `[]byte`. As an array the nil-ness of the field is harder to test, since it is set to [32]byte{0x0} when not explicitly initialized.

asset/asset.go
jharveyb: Pretty sure this is addressed.

dstadulis:
TODOs:

- [ ] Adding feedback from previous review
- [ ] Array to slice changes

asset/asset.go
jharveyb: GroupKey was dropped so i think the `DeriveGroupKey` args are ok as-is.

tapcfg/server.go
```diff
@@ -278,13 +278,13 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
...
-				KeyRing:     keyRing,
-				GenSigner: tap.NewLndRpcGenSigner(
-					lndServices,
-				),
+				Wallet:                walletAnchor,
+				ChainBridge:           chainBridge,
+				Log:                   assetMintingStore,
+				KeyRing:               keyRing,
+				GenSigner:             virtualTxSigner,
+				GenTxBuilder:          &tap.GroupTxBuilder{},
```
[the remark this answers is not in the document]
jharveyb: Would be useful but I don't want to add that for this PR.

tapscript/mint.go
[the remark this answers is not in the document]
jharveyb: Turns out unifying these is the same as moving all the TX-related logic into `asset` - which I don't think we should do in this PR.

asset/tx.go
```diff
@@ -109,3 +112,79 @@ func InputGenesisAssetPrevOut(prevAsset Asset) (*wire.TxOut, error) {
...
+		return nil, nil, err
+	}
+
+	// Now, create the virtual transaction that represents this asset
+	// minting.
+	virtualTx, err := virtualGenesisTx(newAsset)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot tweak group key: %w", err)
+	}
+	populatedVirtualTx := virtualGenesisTxWithInput(
```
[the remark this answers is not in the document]
jharveyb: Yes, though resolved now by moving some TX utilities to `asset`.

asset/asset.go
```diff
@@ -469,11 +469,30 @@ func (g *GroupKeyReveal) GroupPubKey(assetID ID) (*btcec.PublicKey, error) {
...
+
+	if len(singleTweak) != sha256.Size {
+		return nil, fmt.Errorf("genesis tweak must be %d bytes",
+			sha256.Size)
+	}
+
+	// A nil Tapscript root will be represented as an empty byte slice after
+	// encoding and decoding, so we need to check for both forms.
+	internalKey := input.TweakPubKeyWithTweak(rawKey, singleTweak)
+	if tapTweak == nil || bytes.Equal(tapTweak, []byte{}) {
```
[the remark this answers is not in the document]
Roasbeef: Yeah both mean the same thing now that we're just using `[]byte` here.

asset/asset.go
```diff
@@ -726,19 +745,63 @@ func NewScriptKeyBip86(rawKey keychain.KeyDescriptor) ScriptKey {
 
 // DeriveGroupKey derives an asset's group key based on an internal public
 // key descriptor, the original group asset genesis, and the asset's genesis.
-func DeriveGroupKey(genSigner GenesisSigner, rawKey keychain.KeyDescriptor,
-	initialGen Genesis, currentGen *Genesis) (*GroupKey, error) {
+func DeriveGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
+	rawKey keychain.KeyDescriptor, initialGen Genesis,
+	newAsset *Asset) (*GroupKey, error) {
```
Roasbeef, line 740: I think we only need the `newAsset` argument as the genesis is now always the same (the one used to derive the key in the first place), and the asset include a `Genesis` instance.

asset/mock.go
```diff
@@ -266,15 +261,29 @@ func RandID(t testing.TB) ID {
 	return a
 }
 
+// AssetNoErr creates an asset and fails the test if asset creation fails.
```
Roasbeef, line 264: `NewAssetNoErr`?

commitment/commitment_test.go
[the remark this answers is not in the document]
Roasbeef: Can be removed now?

commitment/commitment_test.go
```diff
@@ -329,8 +342,28 @@ func TestMintTapCommitment(t *testing.T) {
 	for _, testCase := range testCases {
 		success := t.Run(testCase.name, func(t *testing.T) {
 			details := testCase.f()
-			groupKey := asset.RandGroupKey(t, testCase.g)
-			_, _, err := Mint(testCase.g, groupKey, details)
+			// Test cases with a nil amount should fail in the call
+			// to Mint() below, not on asset creation. Use the safe
+			// amount of 1 here, and pass the original amount to
+			// Min() below.
+			var amt uint64
```
Roasbeef: Why wasn't this the case before?

commitment/commitment_test.go
```diff
@@ -352,15 +385,26 @@ func TestMintAndDeriveTapCommitment(t *testing.T) {
 
 	const assetType = asset.Normal
 	const numAssets = 5
+	var genesis1protoasset *asset.Asset
```
Roasbeef, line 391: Style not: not fully camelCase, also can nest under the same `var` statement.

Roasbeef: Also doesn't look like this one actually needs to be pre-declared?

commitment/commitment_test.go
```diff
@@ -422,7 +466,12 @@ func TestMintAndDeriveTapCommitment(t *testing.T) {
 	// proofs). We'll reuse the same asset details, except we'll mint them
 	// with a distinct genesis and group key.
 	genesis2 := asset.RandGenesis(t, assetType)
-	groupKey2 := asset.RandGroupKey(t, genesis2)
+	genesis2protoasset := asset.AssetNoErr(
```
Roasbeef, line 472: Same comment here re camel case.

commitment/commitment_test.go
```diff
@@ -527,9 +576,7 @@ func TestSplitCommitment(t *testing.T) {
 		{
 			name: "single input split commitment",
 			f: func() (*asset.Asset, *SplitLocator, []*SplitLocator) {
-				input := randAsset(
```
Roasbeef, line 530: Hmm, perhaps we still want to retain the randomization here w.r.t the actual asset details? IIUC the new dep here is that for everything to validate, we need to create thing in the following order: genesis, group key, asset.

commitment/commitment_test.go
```diff
@@ -438,8 +487,16 @@ func TestSplitCommitment(t *testing.T) {
 	outPoint := wire.OutPoint{}
 	genesisNormal := asset.RandGenesis(t, asset.Normal)
 	genesisCollectible := asset.RandGenesis(t, asset.Collectible)
-	groupKeyNormal := asset.RandGroupKey(t, genesisNormal)
-	groupKeyCollectible := asset.RandGroupKey(t, genesisCollectible)
+	normalProtoAsset := randAsset(t, genesisNormal, nil)
+	collectibleProtoAsset := randAsset(t, genesisCollectible, nil)
+	groupKeyNormal := asset.RandGroupKey(t, genesisNormal, normalProtoAsset)
+	groupKeyCollectible := asset.RandGroupKey(
+		t, genesisCollectible, collectibleProtoAsset,
```
Roasbeef, line 497: Why's this need both the genesis and the actual asset? The asset includes the genesis.

Roasbeef: Ah nvm, with the way that `DeriveGroupKey` func works, it always wants to sign, and during reissuance, you're signing the new asset.

tapdb/universe.go
```diff
@@ -232,9 +233,29 @@ func upsertAssetGen(ctx context.Context, db UpsertAssetStore,
 
 	// Finally, if there's a group key associated with the asset, then
 	// we'll insert that now as well.
-	if groupKey != nil {
-		_, err := upsertGroupKey(
-			ctx, groupKey, db, genPointID, genAssetID,
+	if groupKey != nil && genesisProof.GroupKeyReveal == nil {
+		fmt.Printf("nil group reveal!")
```
Roasbeef: Stray print statement. Should be an error?

tapdb/universe_test.go
```diff
@@ -116,10 +116,21 @@ func randMintingLeaf(t *testing.T, assetGen asset.Genesis,
 		GenesisProof: randProof(t),
 		Amt:          uint64(rand.Int31()),
 	}
+	// The asset within the genesis proof is random; reset the asset genesis
```
Roasbeef, line 119: Style nit: missing newline above.

tapdb/universe_test.go
```diff
@@ -460,6 +472,20 @@ func TestUniverseLeafQuery(t *testing.T) {
 		targetKey.MintingOutpoint = rootMintingPoint
 
 		leaf := randMintingLeaf(t, assetGen, id.GroupKey)
+		if id.GroupKey != nil {
+			// All assets are sharing the same genesis and group
+			// key, so they must also shae the same group witness.
```
Roasbeef, line 477: shae -> share

Roasbeef: Is this for the case of a multi-mint, so same asset just with diff UTXOs? If not, then totally possilbe (if they're reissuance), that these values are the same, but they have a unique witness.

asset/asset.go
Roasbeef: Should use length here, otherwise this should get rejected later as it has a witness, but may not be prev out zero or nl,e tc.

tapgarden/caretaker.go
Roasbeef: Log will be mangled without the space.

proof/mock.go
```diff
@@ -49,6 +50,29 @@ func MockHeaderVerifier(header wire.BlockHeader, height uint32) error {
...
+
+// MockGroupAnchorVerifier is a mock verifier which approves of all group anchor
+// geneses.
+//
+// Group anchor verification usually involves accurately computing a group key,
+// and many unit tests are not focused on group key functionality but still use
+// functions that require a group anchor verifier. This function is used in
+// those cases.
+func MockGroupAnchorVerifier(gen *asset.Genesis,
+	groupKey *asset.GroupKey) error {
```
Roasbeef, line 72: I think this would also need the tapscript root to verify properly? Still getting through commit.

proof/verifier.go
```diff
@@ -356,6 +356,17 @@ func (p *Proof) verifyGroupKeyReveal() error {
 // block header is invalid (usually: not present on chain).
 type HeaderVerifier func(blockHeader wire.BlockHeader, blockHeight uint32) error
 
+// GroupVerifier is a callback function which returns an error if the given
+// group key has not been imported by the tapd daemon. This can occur if the
+// issuance proof for the group anchor has not been imported or synced.
+type GroupVerifier func(groupKey *btcec.PublicKey) error
```
Roasbeef, line 373: Do we just actually want a predicate here, or is the error useful down the line for test assertions?

proof/verifier.go
```diff
@@ -356,6 +356,17 @@ func (p *Proof) verifyGroupKeyReveal() error {
...
+
+// GroupAnchorVerifier is a callback function which returns an error if the
+// given genesis is not the asset genesis of the group anchor. This would be
+// true for an reissuance into an existing group.
+type GroupAnchorVerifier func(gen *asset.Genesis,
```
Roasbeef, line 378: Having trouble parsing the use of this from the function signature and also comment name. IIUC, you also need the tapscript root to verify the mapping, but then this seems like just the normal genesis reveal verification?

tapgarden/caretaker.go
```diff
@@ -1188,3 +1189,104 @@ func GenHeaderVerifier(ctx context.Context,
...
+	mintingStore MintingStore) func(*btcec.PublicKey) error {
+
+	return func(groupKey *btcec.PublicKey) error {
+		if groupKey == nil {
+			return fmt.Errorf("cannot verify empty group key")
+		}
+
+		// This query will err if no stored group has a matching
+		// tweaked group key.
+		_, err := mintingStore.FetchGroupByGroupKey(
```
Roasbeef, line 1259: Can add a cache here in the future to speed things up.

tapgarden/caretaker.go
```diff
@@ -1188,3 +1189,104 @@ func GenHeaderVerifier(ctx context.Context,
...
+}
+
+// GenGroupAnchorVerifier generates a caching group anchor verification callback
+// function given a DB handle.
+func GenGroupAnchorVerifier(ctx context.Context,
+	mintingStore MintingStore) func(*asset.Genesis,
+	*asset.GroupKey) error {
+
+	// Cache anchors for groups that were previously fetched.
+	groupAnchors := make(map[asset.SerializedKey]*asset.Genesis)
```
Roasbeef, line 1279: `cache.LRU`? https://pkg.go.dev/github.com/lightninglabs/neutrino/cache@v1.1.2/lru#NewCache

Can also use above.

tapgarden/caretaker.go
```diff
@@ -1188,3 +1189,104 @@ func GenHeaderVerifier(ctx context.Context,
...
+			)
+			if err != nil {
+				return err
+			}
+
+			groupAnchors[assetGroupKey] = storedGroup.Genesis
+			groupAnchor = storedGroup.Genesis
+		}
+
+		if gen.ID() != groupAnchor.ID() {
```
Roasbeef, line 1297: Ah ok, so this verifies that the given asset is actually the root/anchor asset for the group key. In other words, we can use this to verify if an asset is the first asset to be issued with a group key.

tapgarden/caretaker.go
```diff
@@ -1188,3 +1189,104 @@ func GenHeaderVerifier(ctx context.Context,
...
+
+	return func(gen *asset.Genesis, groupKey *asset.GroupKey) error {
+		assetGroupKey := asset.ToSerialized(&groupKey.GroupPubKey)
+		groupAnchor, ok := groupAnchors[assetGroupKey]
+
+		if !ok {
+			// TODO(jhb): add tapscript root support
+			singleTweak := gen.ID()
+			tweakedGroupKey, err := asset.GroupPubKey(
+				groupKey.RawKey.PubKey, singleTweak[:], nil,
```
Roasbeef, line 1324: Assumes no tapscript root is present.

tapgarden/caretaker.go
```diff
@@ -1188,3 +1189,104 @@ func GenHeaderVerifier(ctx context.Context,
...
+
+// GenRawGroupAnchorVerifier generates a group anchor verification callback
+// function. This anchor verifier recomputes the tweaked group key with the
+// passed genesis and compares that key to the given group key. This verifier
+// is only used in the caretaker, before any asset groups are stored in the DB.
+func GenRawGroupAnchorVerifier(ctx context.Context) func(*asset.Genesis,
+	*asset.GroupKey) error {
+
+	// Cache group anchors we already verified.
+	groupAnchors := make(map[asset.SerializedKey]*asset.Genesis)
```
Roasbeef, line 1314: Can also use LRU cache linked above here with minimal overhead.

proof/mint.go
```diff
@@ -259,21 +263,28 @@ func committedProofs(baseProof *Proof, taprootAssetRoot *commitment.TapCommitmen
...
-
-			assetProof.GroupKeyReveal = &asset.GroupKeyReveal{
-				RawKey: asset.ToSerialized(
-					groupKey.RawKey.PubKey,
-				),
-				TapscriptRoot: tapscriptRoot,
+			err := anchorVerifier(
+				&newAsset.Genesis, groupKey,
+			)
+			if err == nil {
```
Roasbeef, line 274: I think adjusting the naming of `anchorVerifier` and also the callback type could make this easier to understand. 

IIUC, we're just checking here that this asset is the actual anchor asset for the group.

proof/verifier.go
```diff
@@ -453,28 +471,41 @@ func (p *Proof) Verify(ctx context.Context, prev *AssetSnapshot,
...
 		if err := p.verifyGroupKeyReveal(); err != nil {
 			return nil, err
 		}
 	}
 
-	// 7. Either a set of asset inputs with valid witnesses is included that
+	// 7. Verify group key for asset transfers. Any asset with a group key
+	// must carry a group key that has already been imported and verified.
+	if !isGenesisAsset && hasGroupKey {
+		if err := p.verfyGenesisGroupKey(groupVerifier); err != nil {
```
Roasbeef, line 498: 👍

tapdb/asset_minting.go
```diff
@@ -939,14 +939,27 @@ func (a *AssetMintingStore) AddSproutsToBatch(ctx context.Context,
 	// assets committed to within the root commitment specified.
 	assets := assetRoot.CommittedAssets()
 
+	// Before we store any assets from the batch, we need to sort the assets
+	// so that we insert group anchors before reissunces. This is required
+	// to store the asset genesis as the group anchor. All future group
+	// anchor verification depends on inserting group anchors before
+	// reissuances here. We use the raw group anchor verifier since there
+	// is not yet any stored asset group to reference in the verifier.
+	anchorVerifier := tapgarden.GenRawGroupAnchorVerifier(ctx)
```
Roasbeef, line 948: Hmm couldn't we just rely on the upsert behavior here for the genesis?

address/address.go
```diff
@@ -110,12 +109,6 @@ type Tap struct {
 	// asset to be made possible.
 	GroupKey *btcec.PublicKey
 
-	// groupWitness is a stack of witness elements that authorizes the
-	// membership of an asset in a partiular asset group. The witness can
-	// be a single signature or a script from the tapscript tree committed
-	// to with the TapscriptRoot, and follows the witness rules in BIP-341.
-	groupWitness wire.TxWitness
```
Roasbeef, line 117: Hmm, not sure why we ever had this....

Roasbeef (changes requested):

asset/tx.go
```diff
@@ -109,3 +112,79 @@ func InputGenesisAssetPrevOut(prevAsset Asset) (*wire.TxOut, error) {
...
+	}
+
+	return sig, nil
+}
+
+// A compile-time assertion to ensure RawKeyGenesisSigner meets the
+// GenesisSigner interface.
+var _ GenesisSigner = (*RawKeyGenesisSigner)(nil)
+
+type RawGroupTxBuilder struct{}
```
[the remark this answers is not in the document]
jharveyb: Needed for tests where we can't import from `tapscript`.

asset/asset.go
jharveyb: Not sure what you mean, `initialGen == newAsset.Genesis` only for the group creation. We also use the full group anchor genesis for an asset type check here instead of just passing the single tweak.

commitment/commitment_test.go
jharveyb: Should be resolved - we just want invalid amounts to fail because of a call to `Mint()` and not another function used for test setup.

commitment/commitment_test.go
jharveyb: Yup I think this is a leftover from debugging before we removed witness validation from commitment tree construction.

tapdb/universe_test.go
jharveyb: It shouldn't be possible in practice because geneses can never be the same, even in the same batch (asset tags must be unique per-batch).

I think the closest real-world case would be a multi-asset group in the same batch, but for this test we randomly choose to use a group key for the universe ID, so I decided to match that randomization vs. forcing a group key.

tapgarden/caretaker.go
jharveyb: Fixed size seems like a problem? We create these callbacks on startup and the size isn't bounded.

tapdb/asset_minting.go
jharveyb: No -  the problem is that from the P.O.V. of the DB queries, the group anchor is defined as the grouped asset with a witness that was stored first.

This wasn't a problem before as only one group member, the group anchor, had a witness.

Now that all group members have a witness, their insertion must be ordered and the anchor must be inserted first. Otherwise the wrong genesis is returned when querying for the group info later on.

address/address.go
jharveyb: Me neither, but I'm guessing it was related to the check in asset commitments.

proof/verifier.go
jharveyb: Error was useful while working on tests - I think also we want to pass any DB error that could occur inside the verifier vs. only handling the expected error of a missing group.

proof/verifier.go
jharveyb: Similar to the group key reveal verification, but in that context we're checking across different parts of the proof, and here we're checking against the DB or re-deriving the key. Made the comment more verbose.

tapgarden/caretaker.go
jharveyb: Seems like something that could be added to `fn` with an Entry API similar to Rust HashMaps if we end up using this pattern a lot.

jharveyb:
Updated to address reviewer feedback + BIP vectors are updated to add the new proof reveal fields + `GroupKey.TapscriptRoot` is now `[]byte`, which simplified handling in a few places.

tapdb/universe.go
```diff
@@ -233,27 +233,36 @@ func upsertAssetGen(ctx context.Context, db UpsertAssetStore,
...
-		reveal := genesisProof.GroupKeyReveal
-		rawKey, err := reveal.RawKey.ToPubKey()
-		if err != nil {
-			return 0, err
-		}
-
+	if groupKey != nil {
+		// Every group-related issuance must be accompanied by a group
+		// witness.
+		groupWitness := genesisProof.Asset.PrevWitnesses[0].TxWitness
```
guggero, line 239: Should we Squash the changes in this file with 42f097ac to make the review flow somewhat easier? Since this isn't really related to storing the group anchors early, just storing them correctly?

tapdb/universe.go
```diff
@@ -233,27 +233,36 @@ func upsertAssetGen(ctx context.Context, db UpsertAssetStore,
...
+			reveal := genesisProof.GroupKeyReveal
+			rawKey, err := reveal.RawKey.ToPubKey()
+			if err != nil {
+				return 0, err
+			}
+
+			fullGroupKey.RawKey = keychain.KeyDescriptor{
+				PubKey: rawKey,
+			}
+			if len(reveal.TapscriptRoot) != 0 {
```
guggero, line 258: I don't think we need a check or a copy here. Can just do `fullGroupKey.TapscriptRoot = reveal.TapscriptRoot` or not?

tapgarden/caretaker.go
```diff
@@ -462,10 +473,23 @@ func (b *BatchCaretaker) seedlingsToAssetSprouts(ctx context.Context,
...
 		}
 
+		// If a group witness needs to be produced, then we will need a
+		// partially filled asset as part of the signing process.
+		if groupInfo != nil || seedling.EnableEmission {
+			protoAsset, err = asset.New(
+				assetGen, amount, 0, 0, tweakedScriptKey, nil,
+			)
+			if err != nil {
+				return nil, fmt.Errorf("unable to create"+
```
guggero, line 484: nit: space missing in error message before line break.

tapgarden/caretaker.go
guggero: Still missing in latest diff.

tapgarden/caretaker.go
```diff
@@ -1188,3 +1189,113 @@ func GenHeaderVerifier(ctx context.Context,
...
+
+	// Cache known group keys that were previously fetched.
+	assetGroups := fn.NewSet[asset.SerializedKey]()
+
+	return func(groupKey *btcec.PublicKey) error {
+		if groupKey == nil {
+			return fmt.Errorf("cannot verify empty group key")
+		}
+
+		assetGroupKey := asset.ToSerialized(groupKey)
```
guggero, line 1260: Since we don't know if this is accessed by goroutines in parallel, do we need to add a mutex to guard the cache as well?

proof/append_test.go
```diff
@@ -23,7 +23,7 @@ func genTaprootKeySpend(t testing.TB, privKey btcec.PrivateKey,
 
 	t.Helper()
 
-	virtualTxCopy := tapscript.VirtualTxWithInput(
+	virtualTxCopy := asset.VirtualTxWithInput(
```
guggero, line 26: nit: looks like this should be in an earlier commit to allow all commits to compile on their own (can use `git rebase -i v0-3-0-staging --exec "make unit pkg=... case=_none_"` to test that, though it looks like the `v0-3-0staging` branch has a compilation issue already, so no priority, can do that before merging it into `main`)

proof/mint.go
```diff
@@ -259,21 +263,22 @@ func committedProofs(baseProof *Proof, taprootAssetRoot *commitment.TapCommitmen
...
```
guggero, line 274: So what if the `err` _is not_ `nil`? Should we add a return and then do the below non-indented?

tapgarden/caretaker.go
```diff
@@ -1164,6 +1185,39 @@ func SortSeedlings(seedlings []*Seedling) []string {
...
+		case fullAsset.GroupKey != nil:
+			err := anchorVerifier(
+				&fullAsset.Genesis,
+				fullAsset.GroupKey,
+			)
+
+			switch {
+			case err == nil:
+				anchorAssets = append(anchorAssets, fullAsset)
+			default:
```
guggero, line 1208: Should we instead return any errors that occur here? Otherwise they might be swallowed and only surface at a later place where they might be harder to pinpoint.

tapdb/universe.go
```diff
@@ -252,7 +252,7 @@ func upsertAssetGen(ctx context.Context, db UpsertAssetStore,
 			Witness: genesisProof.Asset.PrevWitnesses[0].TxWitness,
 		}
 		if len(reveal.TapscriptRoot) != 0 {
-			copy(fullGroupKey.TapscriptRoot[:], reveal.TapscriptRoot)
+			copy(fullGroupKey.TapscriptRoot, reveal.TapscriptRoot)
```
guggero: I don't think we need the `id` and the `copy` anymore, can just directly assign.

tapgarden/caretaker.go
```diff
@@ -1164,6 +1185,39 @@ func SortSeedlings(seedlings []*Seedling) []string {
 	return allSeedlings
 }
 
+// SortAssets sorts the batch assets such that assets that are group anchors are
+// partitioned from all other assets.
+func SortAssets(fullAssets []*asset.Asset,
+	anchorVerifier proof.GroupAnchorVerifier) ([]*asset.Asset,
+	[]*asset.Asset) {
+
+	var anchorAssets, nonAnchorAssets []*asset.Asset
```
guggero, line 1194: Since we expect the non-anchor witnesses to be the vast majority, should we go ahead and pre-initialize the slice to the size of `fullAssest`? Since if there are many thousands of assets, the time spent on growing the slice might not be insignificant.

universe/base.go
```diff
@@ -264,26 +268,70 @@ func (a *MintingArchive) RegisterNewIssuanceBatch(ctx context.Context,
...
-				ctx, i.ID, i.Key, i.Leaf,
-			)
-			if err != nil {
-				return err
-			}
+	// Issuances that also create an asset group, group anchors, must be
+	// verified and stored before any issuances that may be reissuances into
+	// the same asset group. This is required for proper verification of
+	// reissuances, which may be in this batch.
+	var anchorItems, nonAnchorItems []*IssuanceItem
```
guggero, line 275: Same here re pre-allocating the slice that we expect the majority of items to be in.

guggero (approved):
Awesome to see this all coming together, great work :100: 

Mostly nits and optimizations, so giving it the green checkmark :tada: 

Not sure if the coverage test failure is a flake, re-triggering it to see what's going on.

asset/asset.go
```diff
@@ -428,7 +428,8 @@ type GroupKey struct {
 	// GroupPubKey is the tweaked public key that is used to associate assets
 	// together across distinct asset IDs, allowing further issuance of the
 	// asset to be made possible. The tweaked public key is the result of:
-	//   groupInternalKey + sha256(groupInternalKey || genesisOutPoint) * G
+	// internalKey = rawKey + singleTweak * G
```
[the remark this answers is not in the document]
jharveyb: Didn't know about that feature before, nice.

proof/mint.go
jharveyb: We don't want to return nor skip the rest of this logic, since we still need to store the proof.

tapgarden/caretaker.go
jharveyb: Added concrete errors so we can now pass along anything else.

tapgarden/caretaker.go
jharveyb: Good point, in the global config I made a separate once for each subsystem, but I forgot that some of these paralellize internally.

tapscript/tx.go
jharveyb: Resolved in BIP repo.

halseth (approved):
This looks great now 👌 💯

jharveyb:
Rebased and addressed most recent feedback.

Roasbeef (approved):
LGTM 🍌

PR #491, opened 2023-09-07
Fix docker build

guggero:
https://github.com/lightninglabs/taproot-assets/pull/474 had an accidental base image change that broke the docker build. To avoid this in the future, we also add a CI step to check the release and dev docker build.


jharveyb (approved):
LGTM - are the built images stored / persisted anywhere? Seems like we'd only want that for tagged releases.

guggero:
> LGTM - are the built images stored / persisted anywhere? Seems like we'd only want that for tagged releases.

No, they are discarded. This is just to make sure things are working and we don't only run into that when the release image build is triggered (which does exactly the same but then also pushes them to a repo).

guggero:
Tacked on a commit to fix the Postgres itest getting killed by GitHub because we spawned too many Postgres docker containers (the sub tests didn't clean them up properly).

The tests themselves are still a bit flakey, but will try to address in a different PR.

itest/integration_test.go
```diff
@@ -56,34 +56,31 @@ func TestTaprootAssetsDaemon(t *testing.T) {
 			testCase.name)
 
 		success := t.Run(testCase.name, func(t1 *testing.T) {
-			// Create a subtest harness for each test case.
-			subTestLnd := lndHarness.Subtest(t1)
-
```
ffranr, line 61: But ideally, don't we want a new LND node for each itest? This is encase LND might hold some state across tests and erroneously cause a test to pass (via false positive).

.github/workflows/main.yaml
```diff
@@ -236,7 +233,7 @@ jobs:
   ########################
   integration-test-postgres:
     name: run itests postgres
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
ffranr, line 236: We do we need to change this?

And, if it needs to be updated, shouldn't we change it to a fixed version so that it doesn't potentially stop working at some point in the future?

itest/test_list_on_test.go
```diff
@@ -23,6 +23,18 @@ var testCases = []*testCase{
...
+		test: testReOrgMint,
+	},
+	{
+		name: "re-org send",
+		test: testReOrgSend,
+	},
+	{
+		name: "re-org mint and send",
+		test: testReOrgMintAndSend,
+	},
```
ffranr, line 37: We should probably add a comment to explain that they should be run first because otherwise the postgres itests are flaky.

itest/integration_test.go
guggero: At the moment the only thing we require of `lnd` is it having some funds and it tracking keys and addresses. So there is no state that can conflict between the tests. And as long as that is the case, I think we should take the quite significant speedup we get from not restarting the nodes.

.github/workflows/main.yaml
guggero: We've been running with this in our CI for a very long time now (in lnd) and it never lead to any issues. Downgrading was just a previous attempt to fix this problem (since apparently older images were less prone to kill on out of memory).

itest/test_list_on_test.go
guggero: Good point, added a comment.

ffranr (approved):
Nice bug hunting! :bug:

Issue #403, opened 2023-07-14
[bug]: assets with group key don't have asset id displayed after universe bootstrap


kulisekf:
When trying to create address for asset using group_key tapcli return error

### Environment
Polar v2.0.0; Docker v20.10.21; Compose v2.13.0 (Win 10) - I also wanted to try it on Ubuntu 20.04 with manual installation of lnd and taproot-assets on testnet, but I didn't want to burden Universe unnecessarily

network setup: 1 Bitcoin Core v25.0; 2 LND v0.16.2-beta; 2 Taproot Assets v0.2.0-alpha

### Steps to reproduce
first i minted the asset:
`tapcli assets mint --type normal --name TestGroup --supply 1000 --meta_bytes "test" --enable_emission true`
`tapcli assets mint finalize`
Quick Mine some block
`tapcli assets mint --type normal --name TestGroup --supply 100 --group_key {tweaked_group_key from previous mint asset}`
`tapcli assets mint finalize`
Quick Mine some block

subsequently on the second Taproot Assets node:
sync universe - I did this using the polar GUI instead of in the terminal (I expect the same result)
using the    `tapcli universe roots`    command I listed assets - there is only one, mined in the previous step
I then tried to create an address using group_key since this asset contains it instead of asset_id
`tapcli addrs new --asset_id edef115aec9971700741c989ac1f781d4649b093b711bbe908018e12b02d8541 --amt 500`

### Actual behavior
[tapcli] unable to make addr: rpc error: code = Unknown desc = unable to make new addr: unable to make address for unknown asset edef115aec9971700741c989ac1f781d4649b093b711bbe908018e12b02d8541: asset group is unknown

Maybe this issue is related to https://github.com/lightninglabs/taproot-assets/issues/299 not sure

ffranr:
@kulisekf thanks for starting this issue! I'll take a look at this issue in the next few days.

My initial intuition tells me that it's not related to https://github.com/lightninglabs/taproot-assets/issues/299 , but whatever the cause I think there's certainly room to improve the error message!

kulisekf:
Now I'm looking, the group_key value obtained from the universe roots does not match any value from the assets_group section from assets list, while I would expect it to be the same as the tweaked_group_key (this reasoning of mine is based on an experiment in which I created on the node where I mined the assets the address where I put the asset_id of one of those group mints and in the information about the address that is written when creating the address to the console there is a group_key that has the same value as tweaked_group_key)

[...]

PR #492, opened 2023-09-07
CLI: Improve minting user experience

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/403.

This PR fixes a number of small issues around minting assets in the command line:
 - Return the full batch (and not just the batch key) when minting an asset into a batch (makes it easier to group assets, no need to do another call to find out group key)
 - Return full batch when finalizing a batch
 - Wait for batch to be broadcast (or error out) when finalizing a batch, returning the proper status or an error to the user instead of just logging
 - Allow batch finalization to be re-tried if it fails due to a temporary error (e.g. no confirmed coins available)
 - Fix an issue with how the batch state was asserted in the integration test
 - Add the asset IDs and amounts to the response of the `tapcli universe roots` CLI command (`AssetRoots` RPC).


itest/assets_test.go
```diff
@@ -168,21 +168,29 @@ func mintAssetUnconfirmed(t *harnessTest, tapd *tapdHarness,
 	defer cancel()
 
 	// Mint all the assets in the same batch.
-	for _, assetRequest := range assetRequests {
+	for idx, assetRequest := range assetRequests {
 		assetResp, err := tapd.MintAsset(ctxt, assetRequest)
 		require.NoError(t.t, err)
-		require.NotEmpty(t.t, assetResp.BatchKey)
+		require.NotEmpty(t.t, assetResp.Batch)
```
jharveyb, line 174: nit: duplicate line

rpcserver.go
```diff
@@ -379,8 +379,24 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 				update.Error)
 		}
 
+		batches, err := r.cfg.AssetMinter.ListBatches(update.BatchKey)
+		if err != nil {
+			return nil, fmt.Errorf("unable to list batches: %w",
+				err)
+		}
+
+		rpcBatches, err := fn.MapErr(batches, marshalMintingBatch)
```
jharveyb, line 388: Would be useful to have a `--short` option that omits the `Assets` array, for large batches (or some other form of shorter batch response?).

tapgarden/caretaker.go
```diff
@@ -60,6 +60,17 @@ type BatchCaretakerConfig struct {
 
 	GardenKit
 
+	// BroadcastCompleteChan is used to signal back to the caller that the
+	// batch has been broadcast and is now waiting for confirmation. Either
+	// this channel _or_ BroadcastErrChan is sent on, never both.
+	BroadcastCompleteChan chan struct{}
```
jharveyb, line 66: 💯

tapgarden/interface.go
```diff
@@ -44,7 +44,7 @@ type Planter interface {
 
 	// FinalizeBatch signals that the asset minter should finalize
 	// the current batch, if one exists.
-	FinalizeBatch() (*btcec.PublicKey, error)
+	FinalizeBatch() (*MintingBatch, error)
```
jharveyb, line 47: This breaks a unit test helper here: https://github.com/lightninglabs/taproot-assets/blob/c8bed246d2f3db1fd9c12cc85d0dcada18b65d59/tapgarden/planter_test.go#L235

tapgarden/planter.go
```diff
@@ -595,12 +581,33 @@ func (c *ChainPlanter) gardener() {
 				log.Infof("Finalizing batch %x",
 					batchKey.SerializeCompressed())
 
-				// A force tick must be sent in a separate
-				// goroutine to prevent deadlock.
-				go func() {
-					c.cfg.BatchTicker.Force <- time.Time{}
-				}()
-				req.Resolve(batchKey)
+				caretaker, err := c.finalizeBatch()
```
jharveyb, line 584: Not sure I understand why this logic is added here and not in the existing case in the surrounding `case` statement here: https://github.com/lightninglabs/taproot-assets/blob/c8bed246d2f3db1fd9c12cc85d0dcada18b65d59/tapgarden/planter.go#L470

One issue seems to be returning state from that case, but as-is finalizing the batch via the batch ticker will also not handle possible broadcast errors on the new `BroadcastErrChan`.

tapgarden/planter.go
```diff
@@ -468,38 +470,19 @@ func (c *ChainPlanter) gardener() {
 	for {
 		select {
 		case <-c.cfg.BatchTicker.Ticks():
-			// No pending batch, so we can just continue back to
-			// the top of the loop.
+			// There is no pending batch, so we can just abort.
```
jharveyb, line 473: Also, this interface for finalizing a batch is now async while the `FinalizeBatch` call is now sync?

jharveyb (commented):
Looks good overall, just some comments re: keeping caretaker interfaces consistent.

Also, the new `GroupedAssets` field could be checked in itest assertions like `AssertUniverseRootEqual` and `AssertUniverseRootEquality`.

tapgarden/interface.go
guggero: Yes, I found another issue with the tests, which is now fixed.

rpcserver.go
guggero: Good idea, done.

tapgarden/planter.go
guggero: Finalizing a batch was never sync in the first place. We just fired off the caretaker and never waited for any response. So errors would only ever be logged in the log and not returned to the caller.
But yes, the `c.cfg.BatchTicker.Ticks()` is still async, as there is no "caller" to return a status to. And since this path currently only is taken on daemon startup, this isn't really used anymore (maybe we should also remove that path and completely remove the ticker after all? since IMO batch creation should be explicit).

tapgarden/planter.go
guggero: Because the `c.cfg.BatchTicker.Ticks()` path is only called on startup, there is no "caller" to return errors to. So I don't think we need to explicitly wait for an error to be returned, as it will only be logged anyway. See my comment below about potentially removing the ticker path completely? Because IMO a batch shouldn't suddenly be completed just because the daemon is restarted.

tapgarden/planter.go
jharveyb: AFAICT the ticker is only directly accessed in the unit tests for `tickMintingBatch`, and indeed does not need to be exposed any more. IIUC there is also no consequence from moving the `finalizeBatch` call into the inner `switch{}` statement? If so i'm ok with removing the ticker and that case to handle `Ticks()` entirely.

Didn't think about restart, but I agree that properly saving batch state is the priority for crash safety. Can turn off resume-on-restart here:
https://github.com/lightninglabs/taproot-assets/blob/c5bdbfe25a6c5485ee15be0d91da3af80e3b7493/tapgarden/planter.go#L288

tapgarden/planter_test.go
```diff
@@ -220,19 +220,87 @@ func (t *mintingTestHarness) assertNoPendingBatch() {
...
+		return fmt.Errorf("no new batches created")
+	}, defaultTimeout)
+	require.NoError(t, err)
+
+	require.Len(t, newBatches, 1)
+
+	return newBatches[0]
+}
+
+func (t *mintingTestHarness) assertNoBatchCreated(
```
jharveyb, line 294: New but unused?

jharveyb (approved):
LGTM! Needs a rebase and removing that test func. to pass linting.

I think @Roasbeef could take this over?

cmd/tapcli/assets.go
```diff
@@ -187,15 +199,26 @@ var finalizeBatchCommand = cli.Command{
...
 }
 
 func finalizeBatch(ctx *cli.Context) error {
 	ctxc := getContext()
 	client, cleanUp := getMintClient(ctx)
 	defer cleanUp()
 
-	resp, err := client.FinalizeBatch(ctxc, &mintrpc.FinalizeBatchRequest{})
+	resp, err := client.FinalizeBatch(ctxc, &mintrpc.FinalizeBatchRequest{
+		ShortResponse: ctx.Bool(shortResponseName),
```
Roasbeef, line 220: Alternatively, we can just do this on the CLI side? Still getting thru the diff tho.

rpcserver.go
```diff
@@ -379,29 +379,57 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 				update.Error)
 		}
 
+		batches, err := r.cfg.AssetMinter.ListBatches(update.BatchKey)
```
Roasbeef, line 382: We can get this via the `update` channel above instead. That way, we don't need to scan through the results once we allow more than one pending batch.

rpcserver.go
```diff
@@ -2102,10 +2135,25 @@ func marshallSendAssetEvent(
...
+	}
+
+	rpcBatch := &mintrpc.MintingBatch{
+		BatchKey: batch.BatchKey.PubKey.SerializeCompressed(),
+		State:    rpcBatchState,
+	}
 
-	rpcAssets := make([]*mintrpc.MintAsset, 0, len(batch.Seedlings))
+	// If we don't need to include the seedlings, we can return here.
+	if skipSeedlings {
```
Roasbeef, line 2152: Ah ok, this makes more sense here now.

tapgarden/planter.go
Roasbeef: Tick was added initially to support optimistic aggregation. Now we have the explicit batch creation, so can safely be removed if we don't see any value at all in it.

tapgarden/planter.go
```diff
@@ -595,12 +576,22 @@ func (c *ChainPlanter) gardener() {
 				log.Infof("Finalizing batch %x",
 					batchKey.SerializeCompressed())
 
-				// A force tick must be sent in a separate
-				// goroutine to prevent deadlock.
-				go func() {
-					c.cfg.BatchTicker.Force <- time.Time{}
```
Roasbeef, line 601: Yeah if we're getting rid of this send, then the ticker can def be removed.

tapgarden/caretaker.go
Roasbeef, line 66: I think these can instead reside on the caller level? Eventually we'll support multiple pending batches (diff namespaces) at which point these won't necessarily be global.

Roasbeef: Ah I see it's per caretaker, so np here imo.

Roasbeef (approved):
LGTM 🦠

Issue #455, opened 2023-08-17
Sender node encourages user to update node when encountering unsupported proof courier address protocol


ffranr:
As a first step in sending an asset, the receiver node generates a tap address. The address contains a proof courier service address (which is used in transferring proofs related to the send procedure).

The proof courier address includes a schema/protocol specifier. We must assume that a particular version of the tap daemon can only support a subset of all possible proof courier address protocols. (Later tap daemons could support protocols that earlier daemons do not.)

We need to design a means of selecting/negotiating a viable proof courier protocol based on the capabilities of the receiver and sender nodes.

Roasbeef:
> We need to design a means of selecting/negotiating a viable proof courier protocol based on the capabilities of the receiver and sender nodes.



I don't think the sender+receiver would have an interactive session to communicate that. This is why the potential defaults (and also the eventual sync feature of cross multiverse instances) are important.

guggero:
I agree that we probably don't need to handle that. If the receiver is on an old version, then the sender should be fine. If the sender is on an old version and doesn't understand the address protocol, then I think an error message along the lines of "unknown proof courier protocol <proto> in address, please update your tapd" should be fine.

ffranr:
> If the sender is on an old version and doesn't understand the address protocol, then I think an error message along the lines of "unknown proof courier protocol in address, please update your tapd" should be fine.

I'll update this issue's title to reflect that this is the work to be done.

Roasbeef:
Fixed by https://github.com/lightninglabs/taproot-assets/pull/497

PR #497, opened 2023-09-11
tapfreighter: validate proof courier address before commencing send

ffranr:
This PR closes issue https://github.com/lightninglabs/taproot-assets/issues/455 .

### DIFF
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -172,6 +172,15 @@ func (p *ChainPorter) Stop() error {
 // RequestShipment is the main external entry point to the porter. This request
 // a new transfer take place.
 func (p *ChainPorter) RequestShipment(req Parcel) (*OutboundParcel, error) {
+	// Perform validation on the parcel before we continue. This is a good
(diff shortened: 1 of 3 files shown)


tapfreighter/chain_porter.go
```diff
@@ -1030,6 +1038,24 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
 	}
 }
 
+// validateAddrParcel validates the given address parcel.
+func (p *ChainPorter) validateAddrParcel(addrParcel AddressParcel) error {
+	for idx := range addrParcel.destAddrs {
```
positiveblue, line 1043: the check for empty `destAddrs` is done before getting here right?

ffranr:
@positiveblue Thanks for the review!

I think I've found a better methodology for this change. I've updated the PR, apologies for the late change.

positiveblue (approved):
I like the new approach more, we could incorporate something like this in our rpc flows too :100:

Roasbeef (approved):
LGTM 💣

PR #499, opened 2023-09-11
multi: add CLI flag to enable public access to uni proof courier RPCs

ffranr:
This PR removes the universe proof courier RPC endpoints (`QueryProof` and `InsertProof`) from the default macaroon whitelist and adds a CLI flag to enable them. Which is the same method used for the universe stats RPC endpoints.

This PR also removes a config param called `AcceptRemoteUniverseProofs` now that we shift access control to macaroons.


Roasbeef:
I think this might already be covered by this flag? https://github.com/lightninglabs/taproot-assets/blob/main/config.go#L84 

Or is this where we wan to do things like: you can send issuance proofs, but not normal transfer proofs?

ffranr:
@Roasbeef I now notice that the universe RPCs are whitelisted by default. I missed that before.

```
	// For now, these are the Universe related read/write methods. We permit
	// InsertProof as a valid proof requires an on-chain transaction, so we
	// gain a layer of DoS defense.
	defaultMacaroonWhitelist = map[string]struct{}{
		"/universerpc.Universe/AssetRoots":      {},
		"/universerpc.Universe/QueryAssetRoots": {},
		"/universerpc.Universe/AssetLeafKeys":   {},
		"/universerpc.Universe/AssetLeaves":     {},
		"/universerpc.Universe/QueryProof":      {},
		"/universerpc.Universe/InsertProof":     {},
		"/universerpc.Universe/Info":            {},
	}
```

This PR was supposed to be about whitelisting them via a CLI flag. I think that would be better. Smaller attack surface for most nodes.

Thanks for pointing out the `AcceptRemoteUniverseProofs` config field. I hadn't spotted that. I think we should remove that flag and use the macaroons permissions whitelist for access control.

I'll update the PR to what I think we should do. Please let me know if you have any thoughts on this.

dstadulis:
For security context.  Comment from guggero:
the RPC port is not accessible by the network by default: https://github.com/lightninglabs/taproot-assets/blob/575eaf75acc0f5121c786779f72d47c2fae640e7/tapcfg/config.go#L701

guggero (approved):
LGTM :tada:

guggero:
Looks like the linter also found a now unused variable, then CI should be green :heavy_check_mark:

Roasbeef (approved):
LGTM 🎽

Issue #500, opened 2023-09-12
Add versioning to Tap address


ffranr:
The Tap address structure can be found here: https://github.com/ffranr/taproot-assets/blob/676af1888321f3f6dceab9b0a4e53d1628b211b6/address/address.go#L83C1-L83C1

In this issue we will add a `Version` field to the Tap structure. We will also add code to validate the new version field. And also code for decoding/encoding the new field.

The purpose of this change is to ensure that we can modify the Tap address in the future and handle new address versions soundly.

Example commit for a similar change: https://github.com/lightninglabs/taproot-assets/commit/d1ade18a1e333389621bd63f335bb27048bd7122

Issue #461, opened 2023-08-22
Fallback: Support encoding multiple proof courier addresses in single Tap address


ffranr:
It is currently possible to encode a single proof courier address in a Tap address (implemented here: https://github.com/lightninglabs/taproot-assets/pull/450). This issue is about extending that capability such that we can include a secondary fallback proof courier address within a Tap address.

This enhancement might come in handy if we find that proof courier services are frequently flaky.

dstadulis:
Alternatively if #498 is implemented, this feature wouldn't be necessary.

dstadulis:
> Alternatively if #498 is implemented, this feature wouldn't be necessary.

@ffranr can you confirm if #461 can be closed, as it seems #498 is completed

ffranr:
This issue isn't adding much at all given what we've merged.

PR #501, opened 2023-09-12
Add version field to Tap address

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/500

### DIFF
--- a/address/records.go
+++ b/address/records.go
@@ -13,8 +13,8 @@ import (
 type addressTLVType = tlv.Type
 
 const (
-	// addrVersionType is the TLV type of the addr version.
-	addrVersionType addressTLVType = 0
+	// addrAssetVersionType is the TLV type of the address's asset version.
+	addrAssetVersionType addressTLVType = 0
 
 	// addrAssetIDType is the TLV type of the asset ID.
 	addrAssetIDType addressTLVType = 2
@@ -41,7 +41,7 @@ const (
 
 func newAddressVersionRecord(version *asset.Version) tlv.Record {
 	return tlv.MakeStaticRecord(
-		addrVersionType, version, 1, asset.VersionEncoder,
+		addrAssetVersionType, version, 1, asset.VersionEncoder,
 		asset.VersionDecoder,
 	)
 }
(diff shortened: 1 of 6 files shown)


address/records.go
```diff
@@ -41,7 +41,7 @@ const (
 
 func newAddressVersionRecord(version *asset.Version) tlv.Record {
 	return tlv.MakeStaticRecord(
-		addrVersionType, version, 1, asset.VersionEncoder,
+		addrAssetVersionType, version, 1, asset.VersionEncoder,
```
Roasbeef, line 47: Open question: is there a case where we'd want to bump the version of the asset (to be created), but keep the addr version the same? I think you've been hinting at some of these questions re the over proliferation of versions: do we actually need them all, and for the places we already have them, do we have a set plan for usage?

Roasbeef: Related question is if we have just the addr version: what is the primary use of such a version? Let's say we were to do some of the compression stuff, but having the internal key and the script key be the same. Do we need a new version there? Or is the implicit knowledge that if only one is specified, it applies for both fields enough?

Roasbeef (approved):
LGTM 🎟️

Clean diff, changes look good. Just left a question re discussion of how this version would be used in practice.

address/records.go
ffranr: > implicit knowledge that if only one is specified, it applies for both fields enough?

I think that implicit knowledge makes documentation complicated. A version field is simpler.

I can think of certain addr specific updates that we might want to implement in the future that shouldn't involve a change in asset version. Where the simplest path would involve bumping a Tap address version. Adding a secondary proof courier address (after Tap address compression) is one possible feature that we've discussed briefly in the past: https://github.com/lightninglabs/taproot-assets/issues/461 .

We might also want to include a new expiration field within a Tap address for example. So that a receiver has the option of issuing a new refreshed address at a later date. This gives the receiver the option to stop using a proof courier that they've previously encoded into an address. Which might come in handy if a receiver has publicly shared a repeat use address (where sending nodes use a local saved copy of the address), but the address specified proof courier service is suddenly refusing to cooperate.

jharveyb (approved):
LGTM! Wondering if we'd want to reorder this to the beginning of the TLV stream but that'd be for a separate PR anyways.

address/records.go
jharveyb: Agree, and right now we also don't have a flexible field in the Address like the Metadata field we have for Assets, so the room to add in other fields in the future is limited without some versioning.

PR #509, opened 2023-09-18
multi: add address version to database

guggero:
With the recent PR https://github.com/lightninglabs/taproot-assets/pull/501 we changed the asset version within an address into the address' version itself. We still need the asset version in the address as well though, so this commit restores it, including the database storage.

EDIT: This re-orders all taproot address TLV field values to get them nicely incrementing. I'll follow up with a PR to the BIP.

### DIFF
--- a/address/address.go
+++ b/address/address.go
@@ -23,23 +23,17 @@ var (
 	// ErrUnsupportedHRP is an error returned when we attempt to encode a
 	// Taproot Asset address with an HRP for a network without Taproot Asset
 	// support.
-	ErrUnsupportedHRP = errors.New(
-		"address: unsupported HRP value",
-	)
+	ErrUnsupportedHRP = errors.New("address: unsupported HRP value")
 
 	// ErrMismatchedHRP is an error returned when we attempt to decode a
 	// Taproot Asset address with an HRP that does not match the expected
 	// network.
-	ErrMismatchedHRP = errors.New(
-		"address: network mismatch",
-	)
+	ErrMismatchedHRP = errors.New("address: network mismatch")
 
 	// ErrInvalidBech32m is an error returned when we attempt to decode a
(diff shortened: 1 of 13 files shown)


ffranr:
@guggero I don't think the asset version field was removed in https://github.com/lightninglabs/taproot-assets/pull/501/files#diff-ccfff66eef18bce3fec0f292a24a84b7163361f1af5090d00b6139fdf8115479R102

ffranr (commented):
@guggero I don't think the asset version field was removed in https://github.com/lightninglabs/taproot-assets/pull/501/files#diff-ccfff66eef18bce3fec0f292a24a84b7163361f1af5090d00b6139fdf8115479R102

We just added a new field for the Tap version. The AssetVersion field was not removed or replaced.

I can see that this PR bumps all of the Tap TLV values. I think we should note that in the commit.

guggero:
> @guggero I don't think the asset version field was removed in https://github.com/lightninglabs/taproot-assets/pull/501/files#diff-ccfff66eef18bce3fec0f292a24a84b7163361f1af5090d00b6139fdf8115479R102
> 
> We just added a new field for the Tap version. The AssetVersion field was not removed or replaced.
> 
> I can see that this PR bumps all of the Tap TLV values. I think we should note that in the commit.

Ah yes, you're right, sorry. But I guess we didn't end up storing the address version in the DB.

Updated the PR description to mention the TLV values.

ffranr:
> we didn't end up storing the address version in the DB

Yes, true!

ffranr:
@guggero I think you still need to update the commit message.

guggero:
> @guggero I think you still need to update the commit message.

Right, fixed.

ffranr (approved):
Looks good to me! :+1:

Roasbeef (approved):
LGTM 🦚

Issue #508, opened 2023-09-18
Failed to find coin for address re-use in regtest


ffranr:
This issue tracks a bug reported here: https://github.com/jamaljsr/polar/issues/775#issuecomment-1714919858

Some aspects that I've found interesting whilst investigating so far:
* This bug was encountered in Polar.
* The relevant LND wallet in Polar manages a single UTXO where as in our integration tests each wallet manages ~100 UTXOs.
* This issue/bug may only be relevant in a regtest environment where mining can happen significantly faster than on mainnet.

Developers who have encountered this issue are requesting a new feature which will allow them to customise the coin lock time in a regtest environment. Or to observe minting rather than a set lock time.

A similar issue was reported here: https://lightningcommunity.slack.com/archives/C03B3556HQ8/p1694700896132179

ffranr:
Attempt at reproducing bug: https://github.com/lightninglabs/taproot-assets/pull/515

ffranr:
I've spoken again with @jamaljsr . He mentioned that there was a delay message between send events. And that the coin locking error message was produced whilst attempting to kick off send during that delay.

I'm pretty certain that the delay was caused by the proof courier's backoff procedure erroneously resuming (with a delay) from the stale state of a previous send. This is what I've observed via debugging.

Not too long ago we merged this commit into `main`: https://github.com/lightninglabs/taproot-assets/commit/328e9094caf401e6e395da8a1f5acd2e1c2b70e9 It stops the itests from complaining when the courier's backoff procedure resumes from a bad state.

I've created this PR to fix the courier's backoff procedure initial state: https://github.com/lightninglabs/taproot-assets/pull/529

I think that the coin locking error message is valid and doesn't need to be changed. If the coin was locked in a send event (which was ongoing because of the erroneous courier delay) then it was correct to raise the error message it did.

PR #515, opened 2023-09-20
bug hunting: attempt to lock coins for Tap addr until 10min time out

ffranr:
This PR shouldn't be merged.

This PR attempts to reproduce bug: https://github.com/lightninglabs/taproot-assets/issues/508

It does so by creating an itest that does the following using the same Tap address throughout:
1. Kick off an asset send event. Do not mine.
2. Kick off another send event. This should through up a coin lock error and force the address to get stuck behind a timeout.
3. Mine 6 blocks.
4. Kick off another send event which should have succeeded (but does not) were it not for step 2.

In it's current state, this PR does not capture the bug. I thought that it used to, but having cleaned up the code, it doesn't anymore.

### DIFF
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -129,6 +129,83 @@ func testBasicSendUnidirectional(t *harnessTest) {
 	wg.Wait()
 }
 
+// testBasicSend tests that we can properly send assets back and forth between
+// nodes.
+func testB(t *harnessTest) {
+	ctxb := context.Background()
+	const numUnits = 10
+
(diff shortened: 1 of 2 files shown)


Roasbeef:
> In it's current state, this PR does not capture the bug. I thought that it used to, but having cleaned up the code, it doesn't anymore.

I thought it was more like:
  * send to asset using addr `A`
  * mine a block 
  * send again using addr `A`
  * it fails oddly

ffranr:
Closing in favour of https://github.com/lightninglabs/taproot-assets/pull/529

Issue #506, opened 2023-09-14
Change how asset name is displayed


Crypt-iQ:
Right now, you can mint an asset with a utf-8 name that messes up the terminal output when printed:
```
2023-09-14 09:48:44.535 [INF] GRDN: Adding AssetSeedling(name="<=i7R"KOTNEpO'>qf֝}l#
                                                                                   %Ŵ2
.z%4u~&:_oWl{(!R'i6F7eo                                                               A^,akG}(Nkm ފ/'i#,yf
                       b]߸<_<!/L*13ZBpݩZU%6ʛPY
$       6s6iè't! g:é:                         2nRgC£3/éG;rKXMz"OG.vhJx¨ (SHnxGsb3nZTrLN"wvJQⓩGt *?ܘdq-t    ҤҮP%BqzsNCéDĞCkc6qH*vC:ЖXeçk,;ʹç+PùE    D
,lP3k6yr2S'6ɏ!¨Olp<`ѕy ,§

                         :0iY£Y2vv    tIyT>)S
                                               <KhujU41ǜZèh§*Eg5p5:?Ơ;nQBgP=!rn6ɇ&Ḥ龅8qùrs0Lẉ&§W
                                                                                                J6vecSQ feeJ    0˫&m(2b    >*^ZW&
@$n݁Młv§lӿl:kuSjT+
75ADPYӕ£F N*WujOjV¨pmݮ$ͪc'¨)鿳
cÞr%ThWS&/Hd*ՁPi)Dt0    2-Gr;7˾màӾ
                                  mbN٥N*;4ù5UȽ
                                              7ъܓ+%dùPr_UԱq&o_¨k£כ.BxN 2yU
ԜMdù
    e)gע-.)f
```

It also breaks other lines further down:
```
2023-09-14 09:54:11.758 [ERR] RPCS: [/└␋┼├⎼⎻␌.M␋┼├/M␋┼├A⎽⎽␊├]: ▒⎽⎽␊├ ␌▒┼┼⎺├ ␉␊ ┼␋┌
```

One way to fix this is by requiring ascii names instead.

Roasbeef:
Two things here: 
  1. How we display, if it'll bork stuff we shouldn't display it. 
  2. How we limit the size, can pick something sane, don't need some mega name.

PR #518, opened 2023-09-20
Sanitise asset name

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/506
Closes https://github.com/lightninglabs/taproot-assets/issues/523

### DIFF
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -8,6 +8,9 @@ import (
 	"fmt"
 	"io"
 	"reflect"
+	"strings"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
@@ -16,6 +19,13 @@ import (
 	"github.com/lightninglabs/taproot-assets/mssmt"
 	"github.com/lightningnetwork/lnd/keychain"
 	"github.com/lightningnetwork/lnd/tlv"
+)
+
+const (
+	// MaxAssetNameLength is the maximum byte length of an asset's name.
+	// This byte length is equivalent to character count for single-byte
+	// UTF-8 characters.
+	MaxAssetNameLength = 64
 )
 
 // SerializedKey is a type for representing a public key, serialized in the
@@ -1051,5 +1061,31 @@ func ValidateAssetName(name string) error {
 	if len(name) == 0 {
 		return fmt.Errorf("asset name cannot be empty")
 	}
+
+	// Ensure the asset name is not too long.
+	if len(name) > MaxAssetNameLength {
+		return fmt.Errorf("asset name cannot exceed %d bytes",
+			MaxAssetNameLength)
+	}
+
+	// Ensure the asset name is a valid UTF-8 string.
+	if !utf8.ValidString(name) {
+		return fmt.Errorf("asset name is not a valid UTF-8 string")
+	}
+
+	// Ensure each character is printable.
+	for _, char := range name {
+		if !unicode.IsPrint(char) {
+			hexValue := fmt.Sprintf("\\x%X", char)
+			return fmt.Errorf("asset name cannot contain "+
+				"unprintable character: %s", hexValue)
+		}
+	}
+
+	// Ensure the asset name does not contain only spaces.
+	if len(strings.TrimSpace(name)) == 0 {
+		return fmt.Errorf("asset name cannot contain only spaces")
+	}
+
 	return nil
 }
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -100,7 +100,7 @@ type Genesis struct {
 
 	// Tag is a human-readable identifier for the asset. This does not need
 	// to be unique, but asset issuers should attempt for it to be unique if
-	// possible.
+	// possible. Users usually recognise this field as the asset's name.
 	//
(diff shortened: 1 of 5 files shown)


asset/asset.go
```diff
@@ -18,6 +18,13 @@ import (
 	"github.com/lightningnetwork/lnd/tlv"
 )
 
+const (
+	// MaxAssetNameLength is the maximum byte length of an asset's name.
+	// This byte length is equivalent to character count for single-byte
+	// UTF-8 characters
```
guggero, line 24: nit: missing full stop at end of sentence.

rpcserver.go
```diff
@@ -302,9 +302,9 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 		return nil, fmt.Errorf("asset cannot be nil")
 	}
 
-	// An asset name is mandatory, and cannot be the empty string.
-	if len(req.Asset.Name) == 0 {
-		return nil, fmt.Errorf("asset name cannot be empty")
+	err := asset.ValidateAssetName(req.Asset.Name)
```
guggero, line 305: Should we also invoke this function in `Seedling.validateFields()`?

guggero (approved):
Very nice, LGTM :tada:

jharveyb:
I think we could make progress here without clamping to ASCII - something like `strings.ToValidUTF8()` + `unicode.IsPrint()` would filter out control characters IIRC without filtering out other languages.

https://pkg.go.dev/strings#ToValidUTF8
https://pkg.go.dev/unicode#IsPrint

ffranr:
@jharveyb , good idea.

I've made use of `unicode.IsPrint()`. I think that should take account of valid UTF8 (so I don't think we need `strings.ToValidUTF8()`). Let me know what you think!

jharveyb:
> @jharveyb , good idea.
> 
> I've made use of `unicode.IsPrint()`. I think that should take account of valid UTF8 (so I don't think we need `strings.ToValidUTF8()`). Let me know what you think!

I think having invalid UTF-8 in the string when iterating over it would cause an issue and possibly incorrect decoding. I think we could use `IsValidUTF8()` to detect if the entire string is valid before we actually handle each rune (and exit early if it's invalid).

Re: control characters, not sure if we should escape them or reject them. They are not `IsPrint()` IIUC.

asset/asset.go
```diff
@@ -1045,3 +1053,64 @@ func (a *Asset) Leaf() (*mssmt.LeafNode, error) {
...
+	}
+	return nil
+}
+
+// escapeWhiteSpace escapes the given character if it is a white space
+// character.
+func escapeWhiteSpace(char int32) string {
+	switch char {
+	case ' ':
+		return "\\s"
```
guggero, line 1074: I think white spaces are probably okay in the name?

asset/asset.go
```diff
@@ -1045,3 +1053,64 @@ func (a *Asset) Leaf() (*mssmt.LeafNode, error) {
...
+	if len(name) > MaxAssetNameLength {
+		return fmt.Errorf("asset name cannot exceed %d bytes",
+			MaxAssetNameLength)
+	}
+	return nil
+}
+
+// escapeWhiteSpace escapes the given character if it is a white space
+// character.
+func escapeWhiteSpace(char int32) string {
```
guggero, line 1071: Should we use the internal `rune` type to make this a bit more explicit?

guggero (approved):
LGTM :tada:

asset/asset.go
ffranr: Good point!

jharveyb:
Looks good! I think `escapeWhitespace()` is no longer needed as those chars are already rejected by the `IsPrint()` check.

Would be good to have some test cases for this; can pull different string types from https://go.dev/blog/strings

rpcserver.go
Roasbeef, line 305: Do we also need to mirror this when validating?

asset/asset.go
```diff
@@ -1051,5 +1061,31 @@ func ValidateAssetName(name string) error {
 	if len(name) == 0 {
 		return fmt.Errorf("asset name cannot be empty")
 	}
+
+	// Ensure the asset name is not too long.
+	if len(name) > MaxAssetNameLength {
```
Roasbeef, line 1066: Confirmed these checks still allow emojis ;)

asset/asset.go
```diff
@@ -18,6 +21,13 @@ import (
 	"github.com/lightningnetwork/lnd/tlv"
 )
 
+const (
+	// MaxAssetNameLength is the maximum byte length of an asset's name.
+	// This byte length is equivalent to character count for single-byte
+	// UTF-8 characters.
+	MaxAssetNameLength = 64
```
Roasbeef, line 28: If we want to enforce this as a protocol level check, then we also need to apply this during proof verification.

Roasbeef (changes requested):

asset/asset.go
ffranr: Added a commit to address this.

ffranr:
I've added an extra commit which will validate the asset's name during proof verification.
Let me know what you guys think @jharveyb @Roasbeef

guggero (approved):
Looks great, LGTM :tada:

jharveyb (approved):
Looks good! 🚀

Issue #464, opened 2023-08-23
asset/BIP: create new asset serialization version to elide the inclusion of the witness field in the serialization


Roasbeef:
While writing the TAP channels bLIP, it occurred me that in many instances, one wants to be able to fully construct a top-level taproot output (with a TAP commitment) _without_ first signing the TAP layer itself. One example is sending the existing `commit_sig` message: as is, we'd need an entire state machine round trip to first bind the TAP layer, before being able to sign on the Bitcoin layer. For the most part, this mirrors the current vPSBT flow of: sign TAP, then anchor and sign Bitcoin. However, any protocol that needs a nested set of transactions would run into this dependency issue. This is the exact same problem that was solved on Bitcoin with the inclusion of segwit. In other words, we should add an option to support segwit-like functionality on the TAP layer. 

At a high level, this just means excluding or blanking out the witness field before serialization to be included in the MS-SMT. In all other instances (eg: proof file format), the witness would be serialized as normal. 

One open question here is: what are the downsides? On the Bitcoin layer, the segwit commitment serves to bind the "correct" witness hash commitment to a block. Otherwise, one would get a Bitcoin block that looked invalid, but was actually just someone swapping out the witness values for something else. 

## Steps To Completion 

- [ ] Create a new asset TLV serialization version. 

- [ ] Update the `Leaf` method of the asset struct to use the witnessless serialization: https://github.com/lightninglabs/taproot-assets/blob/0c668cd64424288b19a9bda9f6affcc8fd085237/asset/asset.go#L995C1-L1002
     - [ ] As we do this, we should make sure to unify this new behavior with the other spots that we ened up blanking out a witness before including in the tree. 

- [ ] Add RPC calls that allow minting, sending, and re-anchoring into the new asset type.

PR #520, opened 2023-09-21
asset: introduce new asset version v1 for segwit-like encoding

Roasbeef:
In this commit, we add a new asset version to introduce segwit-like asset encoding where we don't encode the witness vector of an asset's inputs. This applies on the asset version level, so all v1 asset won't have that vector encoded. Note that this is _only_ for the MS-SMT tree, any other TLV encoding such as the asset proof or state transitions remain the same.

Fixes https://github.com/lightninglabs/taproot-assets/issues/464

## TODO 
- [x] Update RPC calls to allow minting for the new asset type, and creating addrs with the new asset type. 
- [x] Update unit tests to also test for the new asset type
- [x] Add version field to vPSBT output
- [x] Add ability to create addrs for v1 outputs
- [x] Add ability to mint v2 assets
- [x] Update `TransferOutputs` (?) on disk to support v0 -> v1 mapping
- [x] Create itests with a nested set of transactions to set new asset type correctness. 
- [x] Update test vectors


guggero:
Ah, so we wouldn't completely _move_ the witness to the proof file? We'd still encode it within the asset in the proof file, but wouldn't commit to the witness in the tree to allow for more flexible constructions? Makes sense, I guess and allows for a very compact diff.

Roasbeef:
> Ah, so we wouldn't completely move the witness to the proof file?

Yep, going for a minimal approach here. It would still always be encoded, but when we go to map it to a leaf in the tree, we don't include the witness information.

Universe trees (which sort of stamp the final transition/issuance proof) will still commit to the witness data, as they include the proof file state transition as a leaf. So that'll end up being the sort of canonical version. By using the tapscript VM, we gain all the extra protections for witness malleability as well. We might want to double check that we're using all the strictest flags there.

Roasbeef:
Updated with relevant commits threading the new state through the entire codebase. Removed from draft as well. 

Two main things still remaining:
  1. Update the `TransferOutput` to also update the asset version. This'll allow us to test things like updating passive assets from v0 to v1, etc. 
  2. Add itests for minting/sending/receiving with the new asset version.

asset/encoding.go
```diff
@@ -390,16 +390,48 @@ func TxWitnessDecoder(r io.Reader, val any, buf *[8]byte, _ uint64) error {
...
 	if t, ok := val.(*[]Witness); ok {
 		if err := tlv.WriteVarInt(w, uint64(len(*t)), buf); err != nil {
 			return err
 		}
 		for _, assetWitness := range *t {
 			var streamBuf bytes.Buffer
-			if err := assetWitness.Encode(&streamBuf); err != nil {
-				return err
+			switch {
+			case len(encodeType) == 1 &&
```
guggero, line 412: Are we using a variadic function just to not break any existing calls? Since there are only two call sites, shouldn't we make this more explicit instead?

commitment/commitment_test.go
```diff
@@ -1177,3 +1184,52 @@ func TestIsTaprootAssetCommitmentScript(t *testing.T) {
...
+	// Next, we'll use the assets to create two root tap commitments.
+	commitmentWitness, err := FromAssets(asset1)
+	require.NoError(t, err)
+
+	commitmentNoWitness, err := FromAssets(assetNoWitness)
+	require.NoError(t, err)
+
+	// The two commitment should be identical as this asset version leaves
+	// out the tx witness field in the leaf encoding.
+	require.Equal(
```
guggero, line 1231: Should we also assert that the same asset with `V0` has a _different_ commitment? Probably kind of pointless since changing the version itself already changes the leaf... So probably not.

tapdb/assets_store.go
```diff
@@ -2513,6 +2514,9 @@ func (a *AssetStore) ConfirmParcelDelivery(ctx context.Context,
 					"output: %w", err)
 			}
 
+			// TODO(roasbeef): asset version needed above?
```
guggero, line 2517: Is our intention to make V1 the new default? I though this would just be op-in for cases where we actually need the flexibility for certain constructions?

tappsbt/interface.go
```diff
@@ -50,6 +50,7 @@ var (
 	PsbtKeyTypeOutputTapAsset                              = []byte{0x76}
 	PsbtKeyTypeOutputTapSplitAsset                         = []byte{0x77}
 	PsbtKeyTypeOutputTapAnchorTapscriptSibling             = []byte{0x78}
+	PsbtKeyTypeOutputAssetVersion                          = []byte{0x79}
```
guggero, line 53: Needs an update to the BIP.

tappsbt/mock.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/internal/test"
 	"github.com/lightningnetwork/lnd/keychain"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/rand"
```
guggero, line 17: Wrong import? I think we instead want to use the `test.RandIntn()` method here for stable/deterministic test vectors.

guggero: But even then, the test vectors will need to be updated _once_ but will become stable after that.

tappsbt/decode.go
```diff
@@ -230,39 +230,60 @@ func (o *VOutput) decode(pOut psbt.POutput, txOut *wire.TxOut) error {
...
-		decoder: assetDecoder(&o.SplitAsset),
-	}, {
-		key: PsbtKeyTypeOutputTapAnchorTapscriptSibling,
-		decoder: tlvDecoder(
-			&o.AnchorOutputTapscriptSibling,
-			commitment.TapscriptPreimageDecoder,
-		),
-	}}
+	mapping := []decoderMapping{
+		{
```
guggero, line 234: Is this to make editing easier in vim? Because tbh, I like the more compact form, as it also saves one level of indentation.

guggero (commented):
Did another pass, looks good!

asset/asset.go
```diff
@@ -1038,6 +1084,17 @@ func (a *Asset) Encode(w io.Writer) error {
 	return stream.Encode(w)
 }
 
+// EncodeNoWitness encodes the asset without the witness into a TLV stream.
+// This is used for serializing on an asset as a leaf within a TAP MS-SMT tree.
+// This only applies when the asset version is v1.
+func (a *Asset) EncodeNoWitness(w io.Writer) error {
+	stream, err := tlv.NewStream(a.encodeRecords(EncodeSegwit)...)
```
ffranr, line 1120: How about switching on the asset version in `Asset.Encode(io.Writer)` and then dropping `EncodeNoWitness`? We can also then get rid of doc comments like `This only applies when the asset version is v1.`.

asset/asset.go
```diff
@@ -741,9 +741,38 @@ type Asset struct {
...
+	scriptKey ScriptKey, groupKey *GroupKey,
+	opts ...NewAssetOpt) (*Asset, error) {
+
+	options := defaultNewAssetOptions()
+	for _, opt := range opts {
+		opt(options)
+	}
 
 	// Collectible assets can only ever be issued once.
 	if genesis.Type != Normal && amount != 1 {
```
ffranr, line 778: At this point we check that the genesis type corresponds to the `amount` field value. Maybe we should check that the asset value is valid at this point also.

address/address.go
```diff
@@ -125,6 +125,31 @@ type Tap struct {
...
+// defaultNewAddrOptions returns a newAddrOptions struct with default values.`
+func defaultNewAddrOptions() *newAddrOptions {
+	return &newAddrOptions{
+		assetVersion: asset.V0,
+	}
+}
+
+// NewAddrOpt is a functional option that allows callers to modify how a new
+// address will be created.
+type NewAddrOpt func(*newAddrOptions)
```
ffranr, line 143: :+1:

rpcserver.go
```diff
@@ -916,6 +916,23 @@ func (r *rpcServer) QueryAddrs(ctx context.Context,
 	}, nil
 }
 
+// parseAssetVersion parses an asset version from the RPC variant.
+func parseAssetVersion(version taprpc.AssetVersion) (asset.Version, error) {
```
ffranr, line 926: In many other places in our code we call this "unmarshal". We should consider doing the same here.

I think this function should be in this file: `taprpc/marshal.go`

cmd/tapcli/addrs.go
```diff
@@ -65,9 +71,22 @@ func newAddr(ctx *cli.Context) error {
...
+	switch ctx.Uint64(assetVersionName) {
+	case 0:
+		assetVersion = taprpc.AssetVersion_ASSET_VERSION_V0
+	case 1:
+		assetVersion = taprpc.AssetVersion_ASSET_VERSION_V1
+
+	default:
+		return fmt.Errorf("invalid asset version: %v",
+			ctx.Uint64(assetVersionName))
+	}
```
ffranr, line 84: This is a RPC "marshal" function that could live in `taprpc/marshal.go`.

commitment/commitment_test.go
```diff
@@ -1177,3 +1184,52 @@ func TestIsTaprootAssetCommitmentScript(t *testing.T) {
...
+	// We'll start by generating a random asset.
+	genesis1 := asset.RandGenesis(t, asset.Normal)
+	groupKey1 := asset.RandGroupKey(t, genesis1)
+	asset1 := randAsset(t, genesis1, groupKey1)
+
+	// We'll modify the asset version so it uses the segwit encoding for
+	// the leaf commitment.
+	asset1.Version = asset.V1
+
+	// We'll modify the asset witness to ontain a valid prev ID, along with
```
ffranr, line 1202: nit: `ontain` -> `contain`

tapdb/assets_store.go
ffranr: If `V1` is the new default, can't we just make these changes as breaking changes in `V0`? Though I appreciate the new version handling logic. I guess we want to support existing assets.

taprpc/mintrpc/mint.proto
```diff
@@ -62,6 +62,11 @@ message MintAsset {
     This asset will be minted with the same group key as the anchor asset.
     */
     string group_anchor = 6;
+
+    /*
+    The version of asset to mint.
+    */
+    taprpc.AssetVersion asset_version = 7;
```
ffranr, line 69: Do we need this field here? I don't see the value in being able to mint an old asset version. Do we mean to signify a different asset type instead perhaps?

asset/asset.go
Roasbeef: So we don't want to unilaterally change the asset encoding. As an example, we want to ensure the asset serialized in proofs still has the witness. The goal here is only to change the serialization when we insert into the SMT tree which is committed to in the top level taproot output.

taprpc/mintrpc/mint.proto
Roasbeef: So we want to sort of leave the path open for either making this version default, or continuing to use the old version. So far I imagine this'll mainly be used when doing more custom higher level protocols, that'll use the vPSBT APIs directly.

asset/asset.go
ffranr: > we want to ensure the asset serialized in proofs still has the witness

Even `V1` assets serialized in proofs still has the witness, is what I think you're saying. :+1:

asset/encoding.go
Roasbeef: Yeah I think that makes sense, actually had a bug here earlier due to missing a default case.

commitment/commitment_test.go
Roasbeef: Why not, tacked that on!

tapdb/assets_store.go
Roasbeef: Yeah agree it's just about having this be opt-in, so fine that change isn't encumbered. 

> If V1 is the new default, can't we just make these changes as breaking changes in V0? 

The current implementation is a sort of a hedge, if it turns out that we were wrong, then we can disallow V1 in the verifier.

tappsbt/interface.go
Roasbeef: https://github.com/Roasbeef/bips/pull/42

tappsbt/decode.go
Roasbeef: Don't think it's a vim specific thing, imo this way I can just copy/paste a row to add another one, then edit that. Before you had to mind the double curly braces at the end, and also the way each row was sort of folded under itself.

asset/asset.go
Roasbeef: What do you mean? That the value doesn't overflow, or something else?

rpcserver.go
Roasbeef: Yep this has been moved, as needed to add some other funcs, and now in the `taprpc` pacakge.

Roasbeef:
Addressed initial round of comments! 

Test vector check seems to be failing still, I modified the random asset gen to also generate random asset versions, but maybe this is tripping it up?

guggero:
> Addressed initial round of comments!
> 
> Test vector check seems to be failing still, I modified the random asset gen to also generate random asset versions, but maybe this is tripping it up?

As mentioned [here](https://github.com/lightninglabs/taproot-assets/pull/520#discussion_r1335809171) you need to use `test.RandIntn()` instead of `rand.Intn()` for test vectors, otherwise the output won't be deterministic.

asset/mock.go
```diff
@@ -103,7 +104,15 @@ func RandAssetWithValues(t testing.TB, genesis Genesis, groupKey *GroupKey,
 		t.Fatal("unhandled asset type", genesis.Type)
 	}
 
-	a, err := New(genesis, units, 0, 0, scriptKey, groupKey)
+	var assetVersion Version
+	if prand.Int()%2 == 0 {
```
guggero, line 108: Need to use `test.RandInt[uint8]()%2 == 0` to achieve stable/deterministic test vectors.

tappsbt/mock.go
```diff
@@ -88,6 +89,7 @@ func RandPacket(t testing.TB) *VPacket {
 		}},
 		Outputs: []*VOutput{{
 			Amount:                             123,
+			AssetVersion:                       asset.Version(rand.Intn(2)),
```
guggero, line 92: Need to use `test.RandIntn(2)` here to get stable/deterministic test vectors. Also, nit: line too long.

tapfreighter/wallet.go
```diff
@@ -784,6 +785,8 @@ func (f *AssetWallet) fundPacketWithInputs(ctx context.Context,
 				ScriptKey: asset.NUMSScriptKey,
 			}
 
+			// TODO(roasbeef): inherit version from which input?
```
guggero, line 788: How about the max version of all inputs? E.g. if one is V1 but the rest isn't, we'll probably not want to downgrade one but instead upgrade the others?

guggero (commented):
Did another pass after the update, looks very good! Main comments are about test vectors, which should be a very easy fix.

asset/asset.go
ffranr: I think I meant asset version (rather than "value"). I think we should check the version here also. We have an opportunity to make sure that `options.assetVersion` is sensible before even constructing the asset.

Roasbeef:
@ffranr @guggero PTAL! 

Final issues I had to hunt down: 
  * Heterogenous sends not possible, needed the asset version in the split locator 
  * `Upsert` and `Delete` in tap/asset commitments not accounting for heterogeneous versions 
  * Change vOut needs to account for version updates. Rn uses the max, but I think we want to re-work the control flow here eventually. Since when you need to set the version for the change output, you don't yet know what the asset will actually be. 


Some of the changes here might also be useful to update your mental model of the relevant sites @jharveyb.

itest/assertions.go
```diff
@@ -716,7 +716,6 @@ func AssertNonInteractiveRecvComplete(t *testing.T,
 func AssertAddr(t *testing.T, expected *taprpc.Asset, actual *taprpc.Addr) {
 	require.Equal(t, expected.AssetGenesis.AssetId, actual.AssetId)
 	require.Equal(t, expected.AssetType, actual.AssetType)
-	require.Equal(t, expected.Version, actual.AssetVersion)
```
guggero: Shouldn't we keep this assertion?

rpcserver.go
```diff
@@ -673,7 +673,6 @@ func (r *rpcServer) listBalancesByAsset(ctx context.Context,
 
 		resp.AssetBalances[assetIDStr] = &taprpc.AssetBalance{
 			AssetGenesis: &taprpc.GenesisInfo{
-				Version:      int32(balance.Version),
```
guggero, line 668: Why are we removing this?

tapdb/asset_minting_test.go
```diff
@@ -611,21 +644,23 @@ func TestCommitBatchChainActions(t *testing.T) {
 
 	// First, we'll create a new batch, then add some sample seedlings, and
 	// then those seedlings as assets.
-	batchKey, groupKey, groupGenAmt, genesisPkt, scriptRoot, assetRoot :=
-		addRandAssets(t, ctx, assetStore, numSeedlings)
+	//
+	// nolint:lll
```
guggero, line 648: nit: linter directive not needed anymore?

guggero (approved):
Very nice, just two questions and a nit, otherwise looks ready to go :100:

rpcserver.go
Roasbeef: Ah yeah this was left over. Basically this conflated the version of the tap commitment and the asset itself. IIRC, an assertion started to fail somewhere, but I think this was before I fixed the issue re max version tracking in the asset/tap commitment, will circle back then report.

itest/assertions.go
Roasbeef: So this ended up being misplaced: originally intent was to ensure that when you made an addr, it matched the asset created. However, in the tests, we want to ensure we can make an addr a v0 asset for a v1 asset (since they're more or less isomorphic) rn. Otherwise, we would never be able to make a v0 addr for a v1 asset with the itests the way tehy're set up rn.

Roasbeef: Turns out just forgot to _also_ set the version in the `AssetGenesis` struct, we have it duplicated across the main asset and this genesis version field as well.

PR #521, opened 2023-09-21
tapgarden: enhance Seedling.validateFields() doc and remove TODO

ffranr:
This commit removes a TODO which is already satisfied. The checks are already enforced on the DB level.

### DIFF
--- a/tapgarden/seedling.go
+++ b/tapgarden/seedling.go
@@ -101,7 +101,8 @@ type Seedling struct {
 // validateFields attempts to validate the set of input fields for the passed
 // seedling, an error is returned if any of the fields are out of spec.
 //
-// TODO(roasbeef): have this series of check be a DB level constraint?
+// NOTE: This function does not check the group key. That check is performed in
+// the validateGroupKey method.
 func (c Seedling) validateFields() error {
 	switch {
 	// Only normal and collectible asset types are supported.


tapgarden/seedling.go
```diff
@@ -101,7 +101,8 @@ type Seedling struct {
 // validateFields attempts to validate the set of input fields for the passed
 // seedling, an error is returned if any of the fields are out of spec.
 //
-// TODO(roasbeef): have this series of check be a DB level constraint?
+// NOTE: This function does not check the group key. That check is performed in
```
guggero, line 104: I think what the TODO actually meant was to add a `CHECK(length(asset_name) > 0)` and `CHECK(asset_supply > 0)` and `CHECK(asset_type = 0 OR asset_type = 1)` to the DB. But IMO we shouldn't have too much logic on the DB level, so I'm fine with this change.

guggero (approved):

tapgarden/seedling.go
ffranr: I saw that we have `NOT NULL` which should cover `CHECK(length(asset_name) > 0)` and `CHECK(asset_supply > 0)`.

Would you be opposed to me adding `CHECK(asset_type = 0 OR asset_type = 1)`?

guggero: No, in SQL an empty string is not the same as `NULL`. Neither is the value 0. So there is a diff here.

ffranr: Good point.

I'm keen on SQL level checks myself, because they can be reused in a different implementation of Taproot Assets. But that probably won't happen for a while!

Roasbeef (approved):
LGTM 🫖

Issue #510, opened 2023-09-19
Limit size of meta reveal


ffranr:
Before mainnet we should limit the size of meta reveal to something reasonable.

https://github.com/lightninglabs/taproot-assets/blob/0c6cf23cd7f991a29bbcdf2420fea3f3775a8ae9/proof/verifier.go#L275-L287

Issue #502, opened 2023-09-13
[testing]: Fuzz test user-input fields & Fix fuzzing crashes


Crypt-iQ:
### Fuzzing user-input fields scope: @Crypt-iQ 
- [ ] addr
- [ ] Proof file parsing and verification 
- [ ] RPC inputs

### Fix known bugs discovered from fuzzing: @guggero 
1. OOM via allocations: https://github.com/lightninglabs/taproot-assets/blob/c64acbeb27ae900dd5003cdf7caa6754d896996f/proof/file.go#L176

2. OOM via allocations: https://github.com/lightninglabs/taproot-assets/blob/c64acbeb27ae900dd5003cdf7caa6754d896996f/proof/file.go#L187

3. OOM via allocations (use DecodeP2P if record sizes are < 65535): https://github.com/lightninglabs/taproot-assets/blob/c64acbeb27ae900dd5003cdf7caa6754d896996f/proof/proof.go#L295 causes this https://github.com/lightninglabs/taproot-assets/blob/c64acbeb27ae900dd5003cdf7caa6754d896996f/proof/encoding.go#L92

4. OOM via allocations: https://github.com/lightninglabs/taproot-assets/blob/c64acbeb27ae900dd5003cdf7caa6754d896996f/proof/encoding.go#L157

5. OOM via allocations: https://github.com/lightninglabs/taproot-assets/blob/c64acbeb27ae900dd5003cdf7caa6754d896996f/proof/encoding.go#L206

6. OOM via allocations (use DecodeP2P): https://github.com/lightninglabs/taproot-assets/blob/c64acbeb27ae900dd5003cdf7caa6754d896996f/proof/taproot.go#L205

7. OOM via allocations + underflow (l may be 0): https://github.com/lightninglabs/taproot-assets/blob/c64acbeb27ae900dd5003cdf7caa6754d896996f/commitment/encoding.go#L176

8. OOM via allocations: https://github.com/lightninglabs/taproot-assets/blob/c64acbeb27ae900dd5003cdf7caa6754d896996f/proof/tx.go#L144

9. OOM via allocations (use DecodeP2P): https://github.com/lightninglabs/taproot-assets/blob/c64acbeb27ae900dd5003cdf7caa6754d896996f/address/address.go#L361

Crypt-iQ:
[...]

PR #525, opened 2023-09-21
decoding: fix panics by limiting max decode sizes in proofs, commitments and assets

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/510.
Fixes https://github.com/lightninglabs/taproot-assets/issues/502.

Contains a commit (6174f9d5) that is already in the `v0-3-0-staging` branch, but was incomplete, so I added itests for it as well.

This PR fixes a number of proof and asset decoding related issues discovered through fuzz tests. Both `FuzzFile` and `FuzzProof` were run for an hour each after the changes and they didn't discover any additional panics.

To help any reviewer, here are all the magic numbers in one place:

```
	// MaxAssetNameLength is the maximum byte length of an asset's name.
	// This byte length is equivalent to character count for single-byte
	// UTF-8 characters.
	MaxAssetNameLength = 64

	// MaxAssetEncodeSize is the size we expect an asset to not exceed in
	// its encoded form. This is used to prevent OOMs when decoding assets.
	// The main contributing factor to this size are the previous witnesses
	// which we currently allow to number up to 65k witnesses.
	MaxAssetEncodeSize = blockchain.MaxBlockWeight

	// FileMaxNumProofs is the maximum number of proofs we expect/allow to
	// be encoded within a single proof file. Given that there can only be
	// one transfer per block, this value would be enough to transfer an
	// asset every 10 minutes for 8 years straight. This limitation might be
	// lifted at some point when proofs can be compressed into a single
	// zero-knowledge proof.
	FileMaxNumProofs = 420000

	// FileMaxProofSize is the maximum size of a single proof in a proof
	// file. The maximum size of a meta reveal is 1 MB, so this value would
	// cap the number of additional inputs within a proof to roughly 128 of
	// assets with such large meta data.
	FileMaxProofSize = 128 * MetaDataMaxLen

	// FileMaxSize is the maximum size of a single proof file. This is not
	// just FileMaxNumProofs * FileMaxProofSize as only the minting proof
	// can commit to a large chunk of meta data. The other proofs are much
	// smaller, assuming they don't all have additional inputs. But we must
	// cap this value somewhere to avoid OOM attacks.
	FileMaxSize = 500 * 1024 * 1024

	// MetaDataMaxLen is the maximum length of the meta data. We limit this
	// to 1MB for now. This should be of sufficient size to commit to any
	// JSON data or even medium resolution images. If there is need to
	// commit to even more data, it would make sense to instead commit to an
	// annotated hash of the data instead. The reason for the limit is that
	// the meta data will be part of the genesis proof, which is stored in
	// the universe and needs to be validated by all senders and receivers
	// of the asset.
	MetaDataMaxLen = 1024 * 1024

	// MaxNumTaprootProofs is the maximum number of Taproot proofs there can
	// be in a proof. This limit represents the maximum block size in vBytes
	// divided by the size of a single P2TR output and is therefore only a
	// theoretical limit that can never be reached in practice.
	MaxNumTaprootProofs = blockchain.MaxBlockBaseSize / input.P2TRSize

	// MaxTaprootProofSize is the maximum size of a single Taproot proof.
	// A Taproot proof can contain a commitment proof which at maximum can
	// contain two MS-SMT proofs that max out at around 10k bytes each (in
	// the worst case).
	MaxTaprootProofSize = tlv.MaxRecordSize

	// MerkleProofMaxNodes is the maximum number of nodes a merkle proof can
	// contain. This is log2(max_num_txs_in_block) + 1, where max number of
	// transactions in a block is assumed to be 17k (theoretical value).
	MerkleProofMaxNodes = 15
```

We use `math.MaxUint16` for the number of previous witnesses and the number of TX witness elements (as well as the max length for each individual witness stack element).

Everything else uses the `tlv.MaxRecordSize` (65k) size as that should fit all of those elements.

cc @Crypt-iQ


scripts/fuzz.sh
```diff
@@ -0,0 +1,45 @@
...
+  RUN_TIME=$2
+  NUM_WORKERS=$3
+  TIMEOUT=$4
+
+  for pkg in $PACKAGES; do
+    pushd $pkg
+
+    go test -list=Fuzz.* | grep Fuzz | while read line; do
+        echo ----- Fuzz testing $pkg:$line for $2 with $3 workers -----
+        go test -fuzz=^$line\$ -test.timeout=$4 -fuzztime=$2 -parallel=$3
```
Roasbeef, line 16: Do we need to set up an Actions cache so the generated corpus goes into there?

proof/meta.go
```diff
@@ -73,3 +95,30 @@ func (m *MetaReveal) Decode(r io.Reader) error {
...
+	if r == nil {
+		return nil
+	}
+
+	switch r.Type {
+	case MetaOpaque:
+		// Valid type, fall through.
+
+	default:
+		return ErrInvalidMetaType
```
Roasbeef, line 111: I think we want to still allow the other type, so this way people can start to experiment with how they should be defined.

rpcserver.go
```diff
@@ -3499,3 +3512,25 @@ func (r *rpcServer) RemoveUTXOLease(ctx context.Context,
...
+}
+
+// unmarshalMetaType maps an RPC meta type into a concrete type.
+func unmarshalMetaType(rpcMeta taprpc.AssetMetaType) (proof.MetaType, error) {
+	switch rpcMeta {
+	case taprpc.AssetMetaType_META_TYPE_OPAQUE:
+		return proof.MetaOpaque, nil
+
+	default:
+		return 0, fmt.Errorf("unknown meta type: %d", rpcMeta)
```
Roasbeef, line 3540: Similar here re allowing the unknown meta types to be encoded/decoded.

commitment/encoding.go
```diff
@@ -161,6 +161,10 @@ func TapscriptPreimageDecoder(r io.Reader, val any, buf *[8]byte,
 		return tlv.ErrRecordTooLarge
 	}
 
+	if l < 1 {
```
Roasbeef, line 164: This mean to check `== 0`?

proof/file.go
```diff
@@ -32,6 +36,27 @@ type Version uint32
...
 	// V0 is the first version of the proof file.
 	V0 Version = 0
+
+	// FileMaxNumProofs is the maximum number of proofs we expect/allow to
+	// be encoded within a single proof file. Given that there can only be
+	// one transfer per block, this value would be enough to transfer an
+	// asset every 10 minutes for 1 year straight. This limitation might be
+	// lifted at some point when proofs can be compressed into a single
+	// zero-knowledge proof.
+	FileMaxNumProofs = 52500
```
Roasbeef, line 46: I think maybe we'll want a bit more breathing room here. 

Mid-tern, we can also move to a route where we just keep them on disk and memory map, so then we never have to load the entire thing into user space memory at once.

proof/tx.go
```diff
@@ -141,6 +142,11 @@ func (p *TxMerkleProof) Decode(r io.Reader) error {
 	if err != nil {
 		return err
 	}
+
+	if numNodes > math.MaxUint8 {
```
ffranr, line 146: I think we should have a named constant in our code instead of `math.MaxUint8` here. Then we can have a comment to explain the constant.

ffranr (commented):
Great to see these checks implemented!

scripts/fuzz.sh
guggero: We don't currently run anything in CI yet, this is just for easily running things locally. I guess the question is: Do we want to use the same pattern as in `lnd` where the corpus is in a separate repo?
IMO we don't need to make it that complicated and can just commit them to this repo directly? Since everything will be below `testdata` folders anyway.

proof/meta.go
guggero: Okay, I see the point about being able to add other types (without changing the proto _or_ the proof version).
But then shouldn't we begin with a 1:1 mapping of the RPC enum to the native type so we _don't_ need any `switch` statements in any place? So basically changing `MetaOpaque MetaType = 1` to `MetaOpaque MetaType = 0` so it matches `META_TYPE_OPAQUE = 0;`?

rpcserver.go
guggero: Okay, I think I'm going to do what I suggested [above](https://github.com/lightninglabs/taproot-assets/pull/525#discussion_r1338527288) and see how that looks.

commitment/encoding.go
guggero: Yes. I guess that's more clear, changed it.

proof/file.go
guggero: Yeah, makes sense. Bumped it to 420000 (roughly 8 years).

rpcserver.go
```diff
@@ -3499,3 +3512,25 @@ func (r *rpcServer) RemoveUTXOLease(ctx context.Context,
 
 	return &wrpc.RemoveUTXOLeaseResponse{}, nil
 }
+
+// marshalMetaType map the native meta type to the RPC type.
+func marshalMetaType(metaType proof.MetaType) (taprpc.AssetMetaType, error) {
```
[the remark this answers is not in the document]
guggero: Removed both functions per the discussion above. But would've agreed with you, just forgot we had that file.

itest/assets_test.go
```diff
@@ -110,6 +114,16 @@ func testMintAssets(t *harnessTest) {
 	// Check that we can retrieve the group keys for the issuable assets.
 	assertGroups(t, issuableAssets)
 
+	// Make sure that the minting proofs reflect the correct state.
+	AssertMintingProofs(t.t, t.tapd, simpleAssets, rpcSimpleAssets)
+	AssertMintingProofs(t.t, t.tapd, issuableAssets, rpcIssuableAssets)
+
+	// Make sure we can't mint assets with too much meta data.
+	invalidRequest := CopyRequest(simpleAssets[0])
+	invalidRequest.Asset.AssetMeta.Data = make([]byte, 1024*1024+1)
```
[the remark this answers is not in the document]
guggero: Yep, good point. Fixed.

asset/encoding.go
```diff
@@ -596,6 +604,10 @@ func LeafEncoder(w io.Writer, val any, buf *[8]byte) error {
 }
 
 func LeafDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
+	if l > MaxAssetEncodeSize {
+		return tlv.ErrRecordTooLarge
+	}
+
```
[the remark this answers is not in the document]
guggero: We basically have that with `DecodeP2P`. But we can't use that here because some records within the structs are allowed to be bigger, so we need to do it manually. But I agree, if we use TLV everywhere, then that would be a nice feature to have baked into the library.

proof/meta.go
```diff
@@ -73,3 +95,30 @@ func (m *MetaReveal) Decode(r io.Reader) error {
 	}
 	return stream.Decode(r)
 }
+
+// ValidateMetaReveal validates the meta reveal.
+func ValidateMetaReveal(r *MetaReveal) error {
```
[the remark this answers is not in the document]
guggero: Makes sense, changed.

proof/tx.go
guggero: Good point. Turns out the value can actually be way lower.

commitment/encoding.go
```diff
@@ -25,6 +25,12 @@ func AssetProofEncoder(w io.Writer, val any, buf *[8]byte) error {
 }
 
 func AssetProofDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
+	// We currently only use this with tlv.DecodeP2P, but in case we ever
+	// don't, we still want to enforce a limit.
+	if l > tlv.MaxRecordSize {
+		return tlv.ErrRecordTooLarge
+	}
+
```
ffranr, line 33: Should we add a `if l == 0 {` check similar to what we have in `TapscriptPreimageDecoder`?

asset/encoding.go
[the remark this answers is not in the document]
ffranr: Didn't know about `DecodeP2P`, thanks.

proof/encoding.go
```diff
@@ -110,6 +128,10 @@ func SplitRootProofEncoder(w io.Writer, val any, buf *[8]byte) error {
 }
 
 func SplitRootProofDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
+	if l > tlv.MaxRecordSize+asset.MaxAssetEncodeSize {
```
ffranr, line 131: Would you mind please adding a comment here to explain why we use `MaxAssetEncodeSize` at this point.

proof/file.go
```diff
@@ -172,6 +197,13 @@ func (f *File) Decode(r io.Reader) error {
 		return err
 	}
 
+	// Cap the number of proofs there can be within a single file to avoid
+	// OOM attacks. See the comment for FileMaxNumProofs for the reasoning
+	// behind the value chosen.
+	if numProofs > FileMaxNumProofs {
+		return ErrProofFileInvalid
+	}
```
ffranr, line 205: Might be a good idea to add context here that the error refers to the number of proofs and not the size of the proof file. Or a specific error type. Not a blocker though, IMO.

ffranr (commented):
I appreciate the small commits. I don't think this is very close to being ready. I just have some concerns with MB vs MiB, but that's about it.

scripts/fuzz.sh
jharveyb: Shellcheck mentions that all the variables in here need double quoting. And that args $2, $3, and $4 above are unused.

proof/file.go
jharveyb: I imagine we won't write proof files for each update in an LN channel right?

proof/meta.go
```diff
@@ -16,6 +17,24 @@ const (
...
+
+	// MetaDataMaxLen is the maximum length of the meta data. We limit this
+	// to 1MB for now. This should be of sufficient size to commit to any
+	// JSON data or even medium resolution images. If there is need to
+	// commit to even more data, it would make sense to instead commit to an
+	// annotated hash of the data instead. The reason for the limit is that
+	// the meta data will be part of the genesis proof, which is stored in
+	// the universe and needs to be validated by all senders and receivers
+	// of the asset.
+	MetaDataMaxLen = 1024 * 1024
```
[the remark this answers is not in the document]
jharveyb: Agree, I've been bitten by MB vs MiB many times. Would vote for a separate constant that's just MB tbh, and then alias from there as needed.

proof/encoding.go
jharveyb: IIUC it's because the split root is just the full root asset right?

jharveyb (commented):
Looking pretty good! I think some of the new limits are missing unit tests + this PR could also include similar mitigation for address decoding:

https://github.com/lightninglabs/taproot-assets/blob/bf4af96a55af655b02d1af89f9fabbbf15be8013/address/address.go#L369

I think just using `DecodeP2P` here is good enough?

proof/file.go
guggero: No, I don't think so. We'll just create a new single transition proof suffix for each update, and each update would be at the same "height" (same location in the proof file, so the file wouldn't be extended with every update).

commitment/encoding.go
guggero: No, since we don't do `l-1` which could then be negative. If `l` is `0` then `tlv.DVarBytes` will just read 0 bytes and the stream below will return `EOF`.

proof/encoding.go
guggero: Hmm, actually, taking a second look, this is just a `TaprootProof`, so it should have the same max size as that record. The thinking was around the split root asset, yes. But that's in the split commitment within previous witness, not in the split root proof.

guggero:
> Looking pretty good! I think some of the new limits are missing unit tests + this PR could also include similar mitigation for address decoding:
> 
> https://github.com/lightninglabs/taproot-assets/blob/bf4af96a55af655b02d1af89f9fabbbf15be8013/address/address.go#L369
> 
> I think just using `DecodeP2P` here is good enough?

Ah, good catch. Added!

scripts/fuzz.sh
Roasbeef: > IMO we don't need to make it that complicated and can just commit them to this repo directly? Since everything will be below testdata folders anyway.



Np committing to the same repo for now. If the corpus gets massive, then we may want to externalize it.

proof/meta.go
Roasbeef: > FYI, with the code as is, adding a new meta reveal type would involve bumping the proof.Proof version.

Why's that the case? No interpretation is actually needed. You just need to re-create the TLV serialization then hash. In reality, the type here is more of a client side thing.

proof/file.go
Roasbeef: > I imagine we won't write proof files for each update in an LN channel right?

They'll get replaced basically each time. Both sides also have everything they need to make the proof files that result from on-chain channel operations.

scripts/fuzz.sh
guggero: Cool, going to let them run for a full hour each and will then commit the corpus.

proof/meta.go
guggero: Yeah, he meant because of the marshal and un-marshal functions that returned an error for unknown meta types. The way I solved it now (direct cast, no `switch` statement) should allow users to use custom types.

proof/file.go
jharveyb: Hmm so if rewrite + update DB num proofs would be static 👌🏽

proof/proof.go
```diff
@@ -124,6 +130,27 @@ const (
...
+	// maximum can contain two MS-SMT proofs that max out at around 10k
+	// bytes each (in the worst case).
+	MaxTaprootProofSizeBytes = tlv.MaxRecordSize
+
+	// BlockHeaderSizeBytes is the size of a Bitcoin block header in bytes.
+	BlockHeaderSizeBytes = 80
+
+	// MerkleProofMaxNodes is the maximum number of nodes a merkle proof can
+	// contain. This is log2(max_num_txs_in_block) + 1, where max number of
+	// transactions in a block is assumed to be 17k (theoretical value).
```
jharveyb, line 152: Would be good to have more explanation / documentation here on where the 17k comes from; I saw slightly different numbers from a cursory search (but the end result was still 14 or 15).

proof/proof.go
```diff
@@ -391,6 +418,10 @@ func (p *Proof) Decode(r io.Reader) error {
 	if err != nil {
 		return err
 	}
+
+	// Note, we can't use the DecodeP2P method here, because the additional
+	// inputs records might be larger than 64k each. Instead, we add
+	// individual limits to each record.
```
jharveyb, line 424: 👍🏽

proof/encoding.go
```diff
@@ -37,6 +40,10 @@ func BlockHeaderEncoder(w io.Writer, val any, buf *[8]byte) error {
 }
 
 func BlockHeaderDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
+	if l > BlockHeaderSizeBytes {
```
jharveyb, line 43: If block headers are fixed-size, why is this set up for variable length at all?

proof/file.go
```diff
@@ -36,6 +40,27 @@ type Version uint32
...
+	// asset every 10 minutes for 8 years straight. This limitation might be
+	// lifted at some point when proofs can be compressed into a single
+	// zero-knowledge proof.
+	FileMaxNumProofs = 420000
+
+	// FileMaxProofSizeBytes is the maximum size of a single proof in a
+	// proof file. The maximum size of a meta reveal is 1 MiB, so this value
+	// would cap the number of additional inputs within a proof to roughly
+	// 16 of assets with such large meta data.
+	FileMaxProofSizeBytes = 16 * MetaDataMaxSizeBytes
```
jharveyb, line 56: Not sure about this one tbh. Since proof size is meta reveal + size of each transition proof. Even with a small meta reveal, if an asset was part of a 'wide' TX like an exchange batched withdraw or coinjoin then its proof file would be a fair bit bigger. IIUC this limit would apply to a single transition proof including proofs for `AdditionalInputs`, but wondering if it could be applied to each `AdditionalInput` indvidually.

Not urgent / can be bumped later client-side.

jharveyb (commented):
Looking pretty good! Only large nit is to add test cases for the changes where we have custom limits vs. `DecodeP2P()`. So `TxMerkleProofRecord` with 16+ nodes, `MetaRevealDataRecord`, `AnchorTxRecord`, etc. Though the change is simple enough that maybe that's overkill and fuzzing time is good enough.

Ready to approve if so.

commitment/proof.go
```diff
@@ -91,7 +91,7 @@ func (p *Proof) Decode(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	return stream.Decode(r)
```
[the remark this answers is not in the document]
guggero: Hmm, good question that I don't have an answer to. Any idea, @Roasbeef?

asset/encoding.go
```diff
@@ -64,16 +64,16 @@ func VarBytesDecoder(r io.Reader, val any, buf *[8]byte, _ uint64) error {
...
-		if bytesLen > (2<<24)-1 {
+		if bytesLen > maxLen {
 			return fmt.Errorf("%w: %v", ErrByteSliceTooLarge,
 				bytesLen)
 		}
 
-		var bytes []byte
-		if err := tlv.DVarBytes(r, &bytes, buf, bytesLen); err != nil {
+		var decoded []byte
+		if err := tlv.DVarBytes(r, &decoded, buf, bytesLen); err != nil {
```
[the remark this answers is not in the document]
guggero: Yeah, so `maxLen` is basically the hard coded (most of the times, sometimes implied by the parent record's max size) upper limit we allow for any allocation. Because `bytesLen` is part of the (potentially malicious) payload, we need to limit its maximum. But yes, if the `bytesLen` is below the `maxLen`, we allow allocating as many bytes as requested. If there aren't that many to read, we'll just end up in a `io.EOF` error which is okay.

proof/records.go
```diff
@@ -293,7 +294,31 @@ func MetaRevealTypeRecord(metaType *MetaType) tlv.Record {
...
+		return uint64(len(*data))
+	}
+	return tlv.MakeDynamicRecord(
+		MetaRevealDataType, data, sizeFunc, tlv.EVarBytes,
+		DVarBytesWithLimit(MetaDataMaxSizeBytes),
+	)
+}
+
+func DVarBytesWithLimit(limit uint64) tlv.Decoder {
+	return func(r io.Reader, val interface{}, _ *[8]byte, l uint64) error {
```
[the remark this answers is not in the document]
guggero: The naming here isn't optimal, I guess. `VarBytesDecoder` is used for a `[]byte` field within a composite records (see for example `GenesisDecoder` where a single TLV record contains multiple fields such as the outpoint, tag, meta hash and so on), so we need to explicitly write the number of bytes with `tlv.WriteVarInt()`.
If we're encoding a `[]byte` field directly as a TLV record, then the size of the record is already encoded as part of the TLV serialization itself and we don't need to write that again.
I'm going to rename `VarBytesEncoder => InlineVarBytesEncoder` and `VarBytesDecoder => InlineVarBytesDecoder` to hopefully make the distinction more clear.
What do you think?

proof/encoding.go
guggero: It is a static size record. But you're right, we can rewrite it somewhat to be more static.

proof/file.go
guggero: Yeah, it's quite hard to pick good numbers... I bumped this to `128 MiB` for now. But still, worst case that means we allocate 128 mebibytes for every attacker that uploads a malicious proof file to an RPC. Currently there is no such RPC that isn't protected by macaroon or by dev mode (the universe RPCs all only accept a single proof, not a file). So we should be okay, but still...

commitment/proof.go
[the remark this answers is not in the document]
jharveyb: Resolved after further reading, the type is not `CompressedProof` but we compress the proof in the encoder, before actual encoding (and mirror this for decode).

proof/records.go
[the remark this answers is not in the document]
jharveyb: That makes sense, I didn't notice that difference between these functions.

proof/file.go
jharveyb: Ah, yeah I would think that maybe the RPC limit is lower than this value, which is global IIUC. E.x. I think this limit would prevent creating transfers with large resulting proofs, even though transfers are macaroon-protected.

jharveyb:
Looking really good! Only nit is now a separate proof size limit for RPC input. I think instead of adding that in decode it could actually be somewhere like `proof.IsProofFile()` as a guard before `proofFile.Decode()`.

proof/file.go
guggero: You mean the general RPC message size? We use 200MB as we do in `lnd`: https://github.com/lightninglabs/taproot-assets/blob/a488519854bb278db82de686629d4fdd2abd226f/server.go#L263

ffranr (approved):
I've reviewed the changes since my last review. All seem reasonable.

jharveyb (approved):
Discussed other places to limit proof sizes offline, these changes look good 🚀

Issue #527, opened 2023-09-21
multi: audit usage of fn.ParSlice


Roasbeef:
Today we use `fn.ParSlice` to do parallel operations on a slice. The routine uses an `errgroup` under the covers to handle the concurrency synchronization: https://github.com/lightninglabs/taproot-assets/blob/a201d48152b274e8faec390dab969391171af27e/fn/concurrency.go#L16-L34

Looking at the `godoc` a bit more closely, I realized that:
  * The first goroutine to return an error will cancel the context. 
  * If that context is checked by any of the other goroutines, then they'll all halt and not finish. 

This may have unintended consequences when we want all the goroutines to run to completion. 

We should audit all call sites to make sure that this behavior is acceptable in the scenarios that it's used.

dstadulis:
Still needs to be started

ffranr:
Uses of `fn.ParSlice`:

1. Used in `MultiArchiver.ImportProofs` to verify a set of proofs. I think this use is correct. We want to fail for all proofs if any proof in the given set fails verification.
2. Used in `ChainPorter.transferReceiverProof` to deliver multiple proofs via proof courier services. I think this use is correct for the moment. If we fail to deliver any one proof then we should re-attempt delivery for all output proofs. I think we currently recognise a send package as complete or not in its entirety.

ffranr:
More uses of `fn.ParSlice`:

3. `BatchCaretaker`, state `BatchStateConfirmed`. Function `fn.ParSlice` is used twice in the `BatchStateConfirmed` state. Both cases involve submitting proofs individually for storage via `ProofFiles.ImportProofs`. If one of the parallel attempts fail then `fn.ParSlice` would cancel the remaining parallel instances. However, some instances may have succeeded. IMO, we should get rid of `fn.ParSlice` in both of these cases and submit proofs atomically to `ProofFiles.ImportProofs`.
4. There are two uses in `FederationEnvoy`. One for syncing across servers and another for syncing across proofs for a given server. In both cases we currently avoid returning an error from the callbacks passed into `fn.ParSlice`. These instances should work as they are. However, we should probably create a new parallelization function that doesn't quit all instances on error as a more appropriate version of `fn.ParSlice`.
5. `MintingArchive.RegisterNewIssuanceBatch` used in verifying batch proofs. This use is correct. We want to quite verification if we fail to verify any proof in the batch.
[...]

PR #528, opened 2023-09-21
universe: update pushProofToFederation to make sure all servers are tried

Roasbeef:
In this commit, we fix a bug that would cause universe proof push to fail for all servers, if anyone of them was unreachable, or rejected the proof.

If we return `err` here, then eventually the context will be read by other goroutines, which is already cancelled at this point. This'll cause the push attempt to stop short.

We now long and return `nil` so the other goroutines will continue.

See https://github.com/lightninglabs/taproot-assets/issues/527 for more details.

### DIFF
--- a/universe/auto_syncer.go
+++ b/universe/auto_syncer.go
@@ -257,14 +257,20 @@ func (f *FederationEnvoy) pushProofToFederation(uniID Identifier, key BaseKey,
 	pushNewProof := func(ctx context.Context, addr ServerAddr) error {
 		remoteUniverseServer, err := f.cfg.NewRemoteRegistrar(addr)
 		if err != nil {
-			return fmt.Errorf("unable to connect to remote "+
-				"server(%v): %v", addr.HostStr(), err)
+			log.Warnf("cannot push proof unable to connect "+
+				"to remote server(%v): %v", addr.HostStr(),
+				err)
+			return nil
 		}
 
 		_, err = remoteUniverseServer.RegisterIssuance(
 			ctx, uniID, key, leaf,
 		)
-		return err
+		if err != nil {
+			log.Warnf("cannot push proof to remote "+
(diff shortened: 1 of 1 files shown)


guggero (approved):
LGTM :tada:

universe/auto_syncer.go
```diff
@@ -257,8 +257,10 @@ func (f *FederationEnvoy) pushProofToFederation(uniID Identifier, key BaseKey,
...
 		if err != nil {
-			return fmt.Errorf("unable to connect to remote "+
-				"server(%v): %v", addr.HostStr(), err)
+			log.Warnf("cannot push proof unable to connect "+
+				"to remote server(%v): %v", addr.HostStr(),
+				err)
+			return nil
 		}
 
 		_, err = remoteUniverseServer.RegisterIssuance(
```
ffranr, line 266: I think that if we encounter and return an error at this point then the `errgroup.Group` of the `fn.ParSlice` function call below will cancel the context for all servers.

We should probably warn rather than return an error here also.

guggero: I think if you can't register a proof with an universe _after_ being able to connect to that universe, I think that might warrant bubbling up the error. As that could point to something being actually wrong (as opposed to the server just being offline).

ffranr: An attacker might run a universe server that would accept a connection but maliciously returns a failure when attempting to submit proofs to it (perhaps for only certain coins etc). Which would interrupt a user's ability to register a proof with all other servers.

I think we should bubble up an error if we fail in any way for **every** universe server. I don't think `fn.ParSlice` (because of `errgroup.Group`) is the right way to go in the long term. But i think further changes belong in a different PR.

ffranr (approved):
I think this is the most robust approach for now. We should probably add better error analysis across all attempts in a future PR.

PR #529, opened 2023-09-22
Backoff procedure erroneously resumed between two different send events

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/508

This PR fixes the initial state of the proof courier's backoff procedure. Before this change, the initial state of the backoff procedure would erroneously include the previous send event(s). The fix involves using the outpoint to distinguish between send events.

### DIFF
--- a/proof/archive.go
+++ b/proof/archive.go
@@ -17,6 +17,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lnwire"
 )
 
 const (
@@ -64,7 +65,7 @@ type Locator struct {
 }
 
 // Hash returns a SHA256 hash of the bytes serialized locator.
-func (l *Locator) Hash() [32]byte {
+func (l *Locator) Hash() ([32]byte, error) {
 	var buf bytes.Buffer
 	if l.AssetID != nil {
 		buf.Write(l.AssetID[:])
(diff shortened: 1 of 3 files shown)


proof/archive.go
```diff
@@ -74,6 +78,16 @@ func (l *Locator) Hash() [32]byte {
 	}
 	buf.Write(l.ScriptKey.SerializeCompressed())
 
+	if l.OutPoint != nil {
+		// Include outpoint hash in generated hash.
+		buf.Write(l.OutPoint.Hash.CloneBytes())
```
guggero, line 83: Can use `lnwire.WriteOutPoint`.

tapfreighter/chain_porter.go
```diff
@@ -451,6 +451,7 @@ func (p *ChainPorter) storeProofs(sendPkg *sendPackage) error {
 		outputProofLocator := proof.Locator{
 			AssetID:   &firstInput.ID,
 			ScriptKey: *out.ScriptKey.PubKey,
+			OutPoint:  &firstInput.OutPoint,
```
guggero, line 454: Aaah, nice catch!

guggero (approved):
LGTM, just two minor comments :tada:

jharveyb (approved):
Good find! 💯

proof/archive.go
```diff
@@ -74,8 +75,16 @@ func (l *Locator) Hash() [32]byte {
 	}
 	buf.Write(l.ScriptKey.SerializeCompressed())
 
+	if l.OutPoint != nil {
+		err := lnwire.WriteOutPoint(&buf, *l.OutPoint)
```
guggero, line 79: Ah, forgot about the error return type. Since we ignore any errors returned by `buf.Write()` above (which AFAIK can never return an actual error on a `bytes.Buffer` but is part of the `io.Writer` interface anyway), we probably can do the same here as well.

ffranr: The was merged before I saw your comment! Let me know if you want me to open another PR to go back on the err handling (I'm guessing this is fine as is?).

PR #532, opened 2023-09-25
tapdb: fix time zone issue with timestamps

guggero:
Fixes an issue where unit tests would fail when being run in certain time zones.


tapdb/assets_store.go
```diff
@@ -1682,7 +1682,7 @@ func (a *AssetStore) LeaseCoins(ctx context.Context, leaseOwner [32]byte,
 			err = q.UpdateUTXOLease(ctx, UpdateUTXOLease{
 				LeaseOwner: leaseOwner[:],
 				LeaseExpiry: sql.NullTime{
-					Time:  expiry,
+					Time:  expiry.UTC(),
```
Roasbeef, line 1685: 👍 for doing this at the lowest level.

Roasbeef (approved):
LGTM 🦚

Only request is some additional detail in the commit message body: 
  * Why did the comparison of the UTC vs localized timestamps encounter an issue here?
  * Why didn't the db translation layer handle normalizing the timestamps? 
  * For all the other areas we use timestamps, do we always need to map them to UTC before inserting?

guggero:
Added a commit message to clarify on this.

It looks like we could customize this by setting `?_time_format=xyz` in the DSN query string when initializing the SQLite database (https://gitlab.com/cznic/sqlite/-/blob/v1.25.0/sqlite.go#L347-355). But then we would have different behavior between SQLite and Postgres.
So I think we'll eventually want to solve this on the `sqlc` level?

jharveyb (approved):
Looks good - didn't see any other use of `UTC()` in other tests.

PR #534, opened 2023-09-26
Generalise universe structs/methods/variables to support transfer proofs in addition to issuance proofs

ffranr:
In this PR we change instances of "minting", "base", and "issuance" to more general terminology which will clarify that the code is also used for transition proofs (non-issuance).

Associated issue: https://github.com/lightninglabs/taproot-assets/issues/535
I may create additional PRs so I don't necessarily want this PR to close the above issue.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -2674,7 +2674,7 @@ func (r *rpcServer) AssetLeafKeys(ctx context.Context,
 }
 
 func marshalAssetLeaf(ctx context.Context, keys taprpc.KeyLookup,
-	assetLeaf *universe.MintingLeaf) (*unirpc.AssetLeaf, error) {
+	assetLeaf *universe.Leaf) (*unirpc.AssetLeaf, error) {
 
 	// In order to display the full asset, we'll also encode the genesis
 	// proof.
@@ -2698,7 +2698,7 @@ func marshalAssetLeaf(ctx context.Context, keys taprpc.KeyLookup,
 
 // marshalAssetLeaf marshals an asset leaf into the RPC form.
 func (r *rpcServer) marshalAssetLeaf(ctx context.Context,
-	assetLeaf *universe.MintingLeaf) (*unirpc.AssetLeaf, error) {
+	assetLeaf *universe.Leaf) (*unirpc.AssetLeaf, error) {
 
 	return marshalAssetLeaf(ctx, r.cfg.AddrBook, assetLeaf)
 }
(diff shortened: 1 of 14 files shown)


universe/interface.go
```diff
@@ -122,13 +122,16 @@ type LeafKey struct {
 	// TODO(roasbeef): add asset type too?
 }
 
-// UniverseKey is the key for a universe.
-func (b LeafKey) UniverseKey() [32]byte {
-	// key = sha256(mintingOutpoint || scriptKey)
+// Hash returns the hash bytes of a LeafKey. The returned value is used as a key
+// within a universe's MS-SMT tree.
+func (b LeafKey) Hash() [32]byte {
```
Roasbeef, line 128: I don't think this rename is warranted: it isn't the hash of the value, it's the key to use when inserting into a universe server.

tapdb/universe.go
```diff
@@ -447,14 +447,14 @@ func (b *BaseUniverseTree) FetchIssuanceProof(ctx context.Context,
...
 //
-// If the given universe key doesn't have a script key specified, then a proof
-// will be returned for each minting outpoint.
+// If the given universe leaf key doesn't have a script key specified, then a
+// proof will be returned for each minting outpoint.
 //
 // NOTE: This function accepts a database transaction and is called when making
 // broader DB updates.
-func universeFetchIssuanceProof(ctx context.Context,
+func universeFetchProofLeaf(ctx context.Context,
```
Roasbeef, line 457: `universeFetchLeafProof` instead?

rpcserver.go
```diff
@@ -2657,7 +2657,7 @@ func (r *rpcServer) AssetLeafKeys(ctx context.Context,
 	// TODO(roasbeef): tell above if was tring or not, then would set
 	// below diff
 
-	leafKeys, err := r.cfg.BaseUniverse.MintingKeys(ctx, universeID)
+	leafKeys, err := r.cfg.BaseUniverse.UniverseLeafKeys(ctx, universeID)
```
Roasbeef, line 2660: Still refs a `BaseUniverse` attribute name FWIW. IMO not a big deal though, would rather spend cycles elsewhere on more critical matters, as we can always change the names around with minimal ramifications down the line.

Roasbeef (changes requested):

rpcserver.go
ffranr: Yes, this PR doesn't address all of the changes that should be made eventually.

tapdb/universe.go
ffranr: This function is called in `MultiverseStore.FetchProofLeaf`. The new name is just copying that function name.

universe/interface.go
ffranr: I've removed this rename.

guggero (approved):
I don't have any strong opinions on any of the new names, so LGTM :tada:

Roasbeef (approved):
LGTM 🦇

PR #536, opened 2023-09-27
build: update to lnd 0.17.0

Roasbeef:
The latest version of btcd had an API change that we adapt to in our logic for merkle tree verification.


guggero (approved):
LGTM :tada:

go.mod
```diff
@@ -25,12 +25,12 @@ require (
 	github.com/lightninglabs/lightning-node-connect/hashmailrpc v1.0.2
 	github.com/lightninglabs/lndclient v0.16.0-15
 	github.com/lightninglabs/neutrino/cache v1.1.1
-	github.com/lightningnetwork/lnd v0.17.0-beta.rc2
-	github.com/lightningnetwork/lnd/cert v1.2.1
-	github.com/lightningnetwork/lnd/clock v1.1.0
-	github.com/lightningnetwork/lnd/ticker v1.1.0
-	github.com/lightningnetwork/lnd/tlv v1.1.0
-	github.com/lightningnetwork/lnd/tor v1.1.1
+	github.com/lightningnetwork/lnd v0.17.0-beta.rc5
```
guggero, line 28: Can we also bump `aperture` to `v0.2-beta` (will resolve to `v0.1.21-beta.0.20230705004936-87bb996a4030` for some reason, maybe because we didn't follow semver versioning?) and `lndclient` to `v0.17.0-0` please?

guggero (changes requested):
Did the same bump in all other projects to prepare, noticed two things that we need to bump to propagate to `litd` in the end.

go.mod
Roasbeef: > maybe because we didn't follow semver versioning?)

Yeah I think it needs an extra 0 at the end there in the tag? so v0.2.0-beta.

guggero: I think so, yes. No idea what other reason there could be for go modules turning a proper tag into a pseudo version...

go.mod
```diff
@@ -20,17 +20,17 @@ require (
 	github.com/jackc/pgconn v1.14.0
 	github.com/jackc/pgerrcode v0.0.0-20220416144525-469b46aa5efa
 	github.com/jessevdk/go-flags v1.4.0
-	github.com/lib/pq v1.10.3
-	github.com/lightninglabs/aperture v0.1.20-beta
+	github.com/lib/pq v1.10.7
+	github.com/lightninglabs/aperture v0.1.21-beta.0.20230705004936-87bb996a4030
 	github.com/lightninglabs/lightning-node-connect/hashmailrpc v1.0.2
-	github.com/lightninglabs/lndclient v0.16.0-15
+	github.com/lightninglabs/lndclient v0.17.0-0
```
guggero, line 26: Should bump this to `v0.17.0-1` to avoid `go mod tidy` to revert `lnd` from non-RC to `.rc5`.

guggero:
The `unknown database backend` is coming from Aperture, due to the version bump. Just need to apply the following diff:

```diff
diff --git a/itest/aperture_harness.go b/itest/aperture_harness.go
index df341aea..95500363 100644
--- a/itest/aperture_harness.go
+++ b/itest/aperture_harness.go
@@ -42,7 +42,8 @@ func NewApertureHarness(t *testing.T, port int) ApertureHarness {
                Authenticator: &aperture.AuthConfig{
                        Disable: true,
                },
-               Etcd: &aperture.EtcdConfig{},
+               DatabaseBackend: "sqlite",
+               Sqlite:          aperture.DefaultSqliteConfig(),
                HashMail: &aperture.HashMailConfig{
                        Enabled:               true,
                        MessageRate:           time.Millisecond,

```

Issue #530, opened 2023-09-25
Add config to restrict universe federation sync to specific assets


ffranr:
In this issue we will add RPC endpoint(s) and CLI arguments to facilitate the configuration of universe curation/sync such that only specific assets are included. Effectively constructing an asset inclusion list for the multiverse (and universe set).

This is a sub-issue of https://github.com/lightninglabs/taproot-assets/issues/476

Add a RPC endpoint called `UniverseFederationSyncConfig`. This endpoint accepts a list of assets and their asset specific configs (exclude, include, issuance only, etc.)

Multiverse federation config is stored in an SQL table.

And then CLI changes to make use of the new RPC via command line. We already have a `universe` command, we should add `federation config` subcommands.

PR #537, opened 2023-09-28
Add federation sync configuration SQL table and RPC endpoints

ffranr:
This PR adds functionality to configure the universe federation sync routine. It facilitates both a general configuration and a universe/asset specific configuration.

This PR is part of the work necessary to complete https://github.com/lightninglabs/taproot-assets/issues/530 .

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/537)


tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -81,3 +81,29 @@ CREATE VIEW universe_stats AS
...
     GROUP BY roots.asset_id, roots.group_key, roots.namespace_root;
+
+-- This table contains general configuration for universe federation syncing.
+CREATE TABLE IF NOT EXISTS federation_general_sync_config(
+    -- This ID column is used to ensure that only one row exists in this table.
+    id INTEGER PRIMARY KEY,
+
+    -- This flag indicates that, in general, only issuance proofs should be
+    -- synced.
+    only_sync_issuance_proofs BOOLEAN NOT NULL
```
guggero, line 92: IMO this is a bit overkill for a single boolean value... Should we perhaps instead create a generic `config` table where we can have key/value (and maybe a field that indicates the type to parse the value into)? That way we don't need to create new tables for any new config items we want to store.

ffranr: Please say more about the generic config table you have in mind.

What would be the types of the key and value columns?

How would we differentiate general config from universe specific?

guggero: Just something like this:
```
create table config (
key text not null,
value text not null
)
```

And thinking about it, we probably don't even need to add the type, the caller would know what type of value to expect.
Example usage would be:
```
tapdb.StoreConfigBool("federation.sync.only_issuance_proofs", true) // would convert the value to `"true"`
tapdb.StoreConfigString("foo.bar", "myvalue")

myBool := tapdb.LoadConfigBool("universe.sync.only_issuance_proofs") // would use `strconv.ParseBool()`
myString := tapdb.LoadConfigString("foo.bar")
```

That would just replace the single boolean value in `federation_general_sync_config` but give us a way to store future config values more easily.

For universe specific stuff like `federation_uni_sync_config` we'd probably still need a table.
But yeah, maybe it's total overkill for now, was just an idea.

guggero: Or the simple alternative for this specific case would be to only have the `federation_uni_sync_config` table but allow the `id` field to be `NULL` which would then represent the general config.

universe/interface.go
```diff
@@ -74,6 +81,55 @@ func (i *Identifier) StringForLog() string {
 		i.String(), i.AssetID[:], groupKey)
 }
 
+// Encode encodes an identifier as a byte array.
+func (i *Identifier) Encode(w io.Writer) error {
```
Roasbeef, line 85: Why this vs encoding directly in the SQL table? Ideal to not pin down the identifier in the schema itself?

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -81,3 +81,29 @@ CREATE VIEW universe_stats AS
     FROM universe_events u
     JOIN universe_roots roots ON u.universe_root_id = roots.id
     GROUP BY roots.asset_id, roots.group_key, roots.namespace_root;
+
+-- This table contains general configuration for universe federation syncing.
+CREATE TABLE IF NOT EXISTS federation_general_sync_config(
```
Roasbeef, line 86: Style nit: should have space before `(`

tapdb/sqlc/migrations/000007_universe.up.sql
Roasbeef: Wouldn't that config table also need some sort of name spacing? Though I _do_ see value in having a nice abstraction of persistent configuration. Not sure where else we'd use it at this point though. Arguably, the other config items like the federation members could also use this. IMO ultimately nice to have some specific structure though. 

So is this a case of YAGNI given we have a set use case for now?

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -81,3 +81,29 @@ CREATE VIEW universe_stats AS
...
+    -- synced.
+    only_sync_issuance_proofs BOOLEAN NOT NULL
+);
+
+-- This table contains universe (asset/asset group) specific federation sync
+-- configuration.
+CREATE TABLE IF NOT EXISTS federation_uni_sync_config (
+    -- This column contains the byte serialized ID of the universe that this
+    -- configuration applies to.
+    universe_id BLOB NOT NULL PRIMARY KEY,
```
Roasbeef, line 100: Re that other TLV, we have the namespace root here, which I think serves the same purpose? https://github.com/lightninglabs/taproot-assets/blob/580c7037f58db3dc91d9c73d1b17e151383b62df/tapdb/sqlc/migrations/000007_universe.up.sql#L5-L9

Then we have a method that takes the ID and maps it to a hex string.

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -81,3 +81,29 @@ CREATE VIEW universe_stats AS
...
+    id INTEGER PRIMARY KEY,
+
+    -- This flag indicates that, in general, only issuance proofs should be
+    -- synced.
+    only_sync_issuance_proofs BOOLEAN NOT NULL
+);
+
+-- This table contains universe (asset/asset group) specific federation sync
+-- configuration.
+CREATE TABLE IF NOT EXISTS federation_uni_sync_config (
```
Roasbeef, line 97: So is the idea that: 
  * We start blank with no sync config. 
  * This means sync everything, but general config says if it should be issuance or not. 
  * As you add a sync config, then we flip to not syncing everything, and only the items with a specified config. 

What about the case of you not necessarily syncing, but accepting insertions? Is this config used as well? So I only care about `A, B, C` and nothing else, someone tries to insert `D`, so then I reject that (with a predicable error).

tapdb/sqlc/migrations/000007_universe.up.sql
Roasbeef, line 92: Also I think we want this value to be true initially? So default is just issuance, then you can flip this (so maybe rename it to accept everything). Then you can create sync configs for sets of assets, and that supercedes this general config?

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -81,3 +81,29 @@ CREATE VIEW universe_stats AS
...
+    -- This column contains the byte serialized ID of the universe that this
+    -- configuration applies to.
+    universe_id BLOB NOT NULL PRIMARY KEY,
+
+    -- This flag indicates that the given universe should be excluded from the
+    -- federation sync process.
+    exclude BOOLEAN NOT NULL,
```
Roasbeef, line 104: See above comment re expected user/config worfklows.

tapdb/universe_federation.go
```diff
@@ -174,4 +187,138 @@ func (u *UniverseFederationDB) LogNewSyncs(ctx context.Context,
...
+	if err != nil {
+		return nil, err
+	}
+
+	return configs, nil
+}
+
+// FederationSyncConfigStore is used to manage the set of Universe servers as part
+// of a federation.
+type FederationSyncConfigStore interface {
```
Roasbeef, line 299: Let's move this to the top before it's referenced?

rpcserver.go
```diff
@@ -3204,6 +3204,98 @@ func (r *rpcServer) DeleteFederationServer(ctx context.Context,
...
+		config := universe.FedUniSyncConfig{
+			UniverseID:             uniID,
+			Exclude:                assetSyncConfig.Exclude,
+			OnlySyncIssuanceProofs: assetSyncConfig.OnlySyncIssuanceProofs,
+		}
+		err = r.cfg.FederationDB.UpsertFederationUniSyncConfig(
+			ctx, config,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to set universe specific "+
```
Roasbeef, line 3244: Print out which uni ID failed to apply?

Roasbeef: I wonder if we should try to make the config modification atomic as well. Reminds me of something we did on the lnd side: if ppl try to update routing configs, we tell them which of them applied vs not.

Roasbeef (changes requested):

ffranr:
All set/query db transactions are now atomic.

I've added an enum to signify which proofs should be synced. I think it's the cleanest solution because each sync process will either exclude proofs, sync issuance, or sync all proofs. We can also expand the enum (which comes close to Oli's key-value idea.)

Let me know what you guys think @guggero @Roasbeef

universe/interface.go
```diff
@@ -552,6 +552,44 @@ type FederationLog interface {
...
 }
 
+// ProofType is an enum that describes the type of proof which can be stored in
+// a given universe.
+type ProofType int
+
+const (
+	// ProofTypeNone indicates that all proofs should be excluded from the
+	// universe.
+	ProofTypeNone ProofType = iota
```
Roasbeef, line 562: I think maybe instead issuance should be the default? 

In what scenario would you imagine this is used, and doesn't return an error?

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -81,3 +81,33 @@ CREATE VIEW universe_stats AS
...
+
+-- This table contains general configuration for universe federation syncing.
+CREATE TABLE IF NOT EXISTS federation_general_sync_config (
+    -- This ID column is used to ensure that only one row exists in this table.
+    id INTEGER PRIMARY KEY,
+
+    -- This field is an enum representing the "maximum" proof type that should
+    -- be synced by default for all universes. Maximum here follows the
+    -- following ordering from min to max: none < issuance < transfer.
+    maximum_proof_type TEXT CHECK(maximum_proof_type IN ('none', 'issuance', 'transfer')) NOT NULL DEFAULT 'issuance'
```
Roasbeef, line 93: max? I think we just want this to be the `sync_type`? 

Or I guess this is trying to exploit the lexicographical ordering?

tapdb/sqlc/migrations/000007_universe.up.sql
Roasbeef: Just wanting to confirm we have the same mental model here.

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -81,3 +81,33 @@ CREATE VIEW universe_stats AS
...
+);
+
+-- This table contains universe (asset/asset group) specific federation sync
+-- configuration.
+CREATE TABLE IF NOT EXISTS federation_uni_sync_config (
+    -- This field contains the byte serialized ID of the asset that this
+    -- configuration applies to.
+    asset_id  BLOB NOT NULL UNIQUE CHECK(length(asset_id) = 32),
```
Roasbeef, line 101: I think since we have a composite primary key below, we don't actually need `UNIQUE` here?

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -81,3 +81,33 @@ CREATE VIEW universe_stats AS
...
+
+    -- This field contains the byte serialized group key public key of the asset
+    -- group that this configuration applies to.
+    group_key BLOB,
```
Roasbeef, line 105: We should also add a size assertion here.

tapdb/universe_federation.go
```diff
@@ -174,4 +211,142 @@ func (u *UniverseFederationDB) LogNewSyncs(ctx context.Context,
...
+// UpsertFederationSyncConfig upserts both the general and universe specific
+// federation sync configs.
+func (u *UniverseFederationDB) UpsertFederationSyncConfig(
+	ctx context.Context, generalSyncConfig *universe.FedGeneralSyncConfig,
+	uniSyncConfigs []*universe.FedUniSyncConfig) error {
+
+	var writeTx UniverseFederationOptions
+	return u.db.ExecTx(ctx, &writeTx, func(db UniverseServerStore) error {
+		// Upsert general federation sync config.
+		err := db.SetFederationGeneralSyncConfig(ctx,
```
Roasbeef, line 223: Odd line folding.

tapdb/universe_federation.go
```diff
@@ -174,4 +211,142 @@ func (u *UniverseFederationDB) LogNewSyncs(ctx context.Context,
...
+				config = uniSyncConfigs[i]
+
+				uniID        = config.UniverseID
+				assetIDBytes = uniID.AssetID[:]
+			)
+
+			// If the group key is set, then we'll serialize it
+			// into bytes.
+			var pubKey []byte
+			if uniID.GroupKey != nil {
```
Roasbeef, line 241: I think we want the group key to supersede all in this case?

tapdb/universe_federation.go
```diff
@@ -174,4 +211,142 @@ func (u *UniverseFederationDB) LogNewSyncs(ctx context.Context,
 	})
 }
 
-var _ universe.FederationLog = (*UniverseFederationDB)(nil)
+// UpsertFederationSyncConfig upserts both the general and universe specific
+// federation sync configs.
+func (u *UniverseFederationDB) UpsertFederationSyncConfig(
+	ctx context.Context, generalSyncConfig *universe.FedGeneralSyncConfig,
+	uniSyncConfigs []*universe.FedUniSyncConfig) error {
```
Roasbeef, line 218: Can use var args here and below if you want to save a few lines when you map from the gRPC proto to this struct.

tapdb/universe_federation.go
```diff
@@ -174,4 +211,142 @@ func (u *UniverseFederationDB) LogNewSyncs(ctx context.Context,
...
+			return err
+		}
+
+		generalConfig = &universe.FedGeneralSyncConfig{
+			ProofTypes: proofTypes,
+		}
+
+		// If no rows in result set, then we haven't set the config
+		// yet, so we'll default to true.
+		if errors.Is(err, sql.ErrNoRows) {
```
Roasbeef, line 294: I think you want this check above? Since we use the value there without checking the error.

rpcserver.go
```diff
@@ -3204,6 +3231,93 @@ func (r *rpcServer) DeleteFederationServer(ctx context.Context,
...
+				"types: %w", err)
+		}
+		syncGeneralConfig = &universe.FedGeneralSyncConfig{
+			ProofTypes: proofTypes,
+		}
+	}
+
+	// Unmarshal asset (asset/asset group) specific sync configs.
+	assetSyncConfigs := make(
+		[]*universe.FedUniSyncConfig, len(req.AssetSyncConfig))
```
Roasbeef, line 3257: Paren on new line at the end here?

rpcserver.go
```diff
@@ -3204,6 +3231,93 @@ func (r *rpcServer) DeleteFederationServer(ctx context.Context,
...
+
+// QueryFederationSyncConfig queries the universe federation sync configuration
+// settings.
+func (r *rpcServer) QueryFederationSyncConfig(ctx context.Context,
+	_ *unirpc.QueryFederationSyncConfigRequest,
+) (*unirpc.QueryFederationSyncConfigResponse, error) {
+
+	// Obtain the general and universe specific federation sync configs.
+	generalConfig, uniSyncConfigs, err :=
+		r.cfg.FederationDB.QueryFederationSyncConfigs(ctx)
```
Roasbeef, line 3289: Can use a temp variable to allow a proper line fold?

taprpc/universerpc/universe.proto
```diff
@@ -475,3 +489,53 @@ message GroupedUniverseEvents {
...
+    repeated AssetFederationSyncConfig asset_sync_config = 2;
+}
+
+message SetFederationSyncConfigResponse {
+}
+
+// GeneralFederationSyncConfig is a general configuration for universe
+// federation syncing.
+message GeneralFederationSyncConfig {
+    string proof_types = 1;
```
Roasbeef, line 505: I think we want this to be a proper proto enum instead.

Roasbeef: Also IIUC, this is a singular value so maybe `universe_sync_mode`?

taprpc/universerpc/universe.proto
```diff
@@ -475,3 +489,53 @@ message GroupedUniverseEvents {
...
+    //    bool only_sync_issuance_proofs = 1;
+}
+
+// AssetFederationSyncConfig is an asset universe specific configuration for
+// federation syncing.
+message AssetFederationSyncConfig {
+    // The ID of the universe to configure.
+    ID id = 1;
+
+    string proof_types = 2;
```
Roasbeef, line 519: Same here re enum.

Roasbeef (changes requested):

universe/interface.go
```diff
@@ -552,6 +552,44 @@ type FederationLog interface {
...
+)
+
+// String returns a human-readable string representation of the sync proof types
+// enum.
+func (t ProofType) String() string {
+	return []string{"none", "issuance", "transfer"}[t]
+}
+
+// NewSyncProofTypesFromString returns a new proof type given a string.
+func NewSyncProofTypesFromString(s string) (ProofType, error) {
```
guggero, line 580: nit: `ParseProofType`?

universe/interface.go
guggero: I think this will be used for the setting of "what proofs do I want to sync with my universe", which can be "none". But also still going through the diff.

tapdb/universe_federation.go
```diff
@@ -33,6 +68,8 @@ type UniverseServerStore interface {
 
 	// ListUniverseServers returns the total set of all universe servers.
 	ListUniverseServers(ctx context.Context) ([]sqlc.UniverseServer, error)
+
+	FederationSyncConfigStore
```
guggero, line 72: nit: We normally add embedded interfaces to the top.

rpcserver.go
```diff
@@ -3499,3 +3613,25 @@ func (r *rpcServer) RemoveUTXOLease(ctx context.Context,
 
 	return &wrpc.RemoveUTXOLeaseResponse{}, nil
 }
+
+// MarshalAssetFederationSyncConfig returns an RPC ready asset specific
+// federation sync config.
+func MarshalAssetFederationSyncConfig(
```
guggero, line 3619: Should we call this `MarshalAssetFedSyncCfg` to shorten?

guggero: Also, shouldn't this be in `taprpc/marshal.go`?

taprpc/universerpc/universe.proto
```diff
@@ -126,6 +126,20 @@ service Universe {
     period, grouped by day.
     */
     rpc QueryEvents (QueryEventsRequest) returns (QueryEventsResponse);
+
+    /* tapcli `universe sync config`
```
guggero, line 130: If this CLI sub command doesn't actually exist, this will break the API generator. So we should just remove the whole line. Or add the corresponding commands.

tapdb/universe_federation.go
```diff
@@ -174,4 +211,142 @@ func (u *UniverseFederationDB) LogNewSyncs(ctx context.Context,
...
+			}
+		}
+
+		return nil
+	})
+}
+
+// QueryFederationSyncConfigs returns the general and universe specific
+// federation sync configs.
+func (u *UniverseFederationDB) QueryFederationSyncConfigs(
```
guggero, line 262: A standalone unit test for DB operations would be nice.

taprpc/universerpc/universe.proto
```diff
@@ -475,3 +489,53 @@ message GroupedUniverseEvents {
...
+
+message SetFederationSyncConfigResponse {
+}
+
+// GeneralFederationSyncConfig is a general configuration for universe
+// federation syncing.
+message GeneralFederationSyncConfig {
+    string proof_types = 1;
+
+    //    // This flag indicates that, in general, only issuance proofs should
```
guggero, line 507: nit: commented out code, here and below.

itest/universe_test.go
```diff
@@ -6,7 +6,10 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"github.com/lightninglabs/taproot-assets/asset"
```
guggero, line 9: nit: Goland auto import issue. Can solve by adding an entry to "Settings -> Tools -> Actions on Save" that executes `gosimports`.

guggero (commented):
Looks pretty good. I think the level of configurability makes sense for the moment.
I assume the code to actually sync according to the setting will be done in a follow-up PR?

universe/interface.go
ffranr: > "what proofs do I want to sync with my universe", which can be "none".

Yes, that's what I had in mind. `ProofTypeNone` also allows us to not specify the proof type in an RPC call.

tapdb/sqlc/migrations/000007_universe.up.sql
ffranr: > As you add a sync config, then we flip to not syncing everything, and only the items with a specified config.

I have a different thought on this point. In my mind, a universe specific config should supersede the general config.

So if we only wanted to sync those universes which had specific configuration settings, then the general config should be set to "no syncing". And then the universe specific configs would supersede that general stance.

> What about the case of you not necessarily syncing, but accepting insertions? Is this config used as well? So I only care about A, B, C and nothing else, someone tries to insert D, so then I reject that (with a predicable error).

Where syncing is concerned, in its current state, this PR captures that behaviour with a general "no sync" config and then universe specific configs for `A`, `B`, `C`.

We could conceivably reuse these tables as permissions around universe proof couriering (universe specific permissions on `InsertProof`). I guess I can name the tables in a more general way in case we want to go down that route. What do you think @Roasbeef ?

rpcserver.go
ffranr: Like this?
```
	queryFedSyncConfigs := r.cfg.FederationDB.QueryFederationSyncConfigs
	generalConfig, uniSyncConfigs, err := queryFedSyncConfigs(ctx)
```

rpcserver.go
ffranr: > shouldn't this be in taprpc/marshal.go?

It makes use of a struct from the `universe` package which will be a import cycle in the `taprpc` package:
```
	imports github.com/lightninglabs/taproot-assets/taprpc
	imports github.com/lightninglabs/taproot-assets/taprpc/universerpc
	imports github.com/lightninglabs/taproot-assets/taprpc: import cycle not allowed
```

This is because the universe proof courier imports `universerpc` into the `universe` package. That probably needs to be disentangled.

taprpc/universerpc/universe.proto
ffranr: I've removed for now, will add back in a future PR.

Roasbeef:
> I assume the code to actually sync according to the setting will be done in a follow-up PR?

Yep plan is to do this as we work through https://github.com/lightninglabs/taproot-assets/issues/544.

Roasbeef:
Needs a rebase!

tapdb/universe_federation.go
Roasbeef: Bumping this so not forgotten, or maybe needs a slight rework to the control flow?

rpcserver.go
Roasbeef: Yeah I think that work work?

tapdb/sqlc/migrations/000007_universe.up.sql
Roasbeef: >  then the general config should be set to "no syncing". And then the universe specific configs would supersede that general stance.

I think that makes sense. 

> We could conceivably reuse these tables as permissions around universe proof couriering (universe specific permissions on InsertProof). I guess I can name the tables in a more general way in case we want to go down that route. 

I'm down with that approach. Seems like just a bit more logic to give us fine grained control from day one. Then as above, the specific config supersedes the general. 

```sql
-- This table contains universe (asset/asset group) specific remote proof
-- insert/export configuration.
CREATE TABLE IF NOT EXISTS uni_remote_permissions (
    -- This field contains the byte serialized ID of the asset that this
    -- configuration applies to.
    asset_id  BLOB CHECK(length(asset_id) = 32),

    -- This field contains the byte serialized group key public key of the asset
    -- group that this configuration applies to.
    group_key BLOB CHECK(LENGTH(group_key) = 32),

    -- This field is an enum representing the proof type stored in the given
    -- universe.
    proof_type TEXT NOT NULL CHECK(proof_type IN ('issuance', 'transfer')),

    -- This field is a boolean that indicates whether or not the given universe
    -- should accept remote proof insertion (via sync or proof courier).
    allow_remote_insert BOOLEAN NOT NULL,

    -- This field is a boolean that indicates whether or not the given universe
    -- should accept remote proof export (via sync or proof courier).
    allow_remote_export BOOLEAN NOT NULL,

    -- Both the asset ID and group key cannot be null at the same time.
    CHECK (asset_id IS NOT NULL OR group_key IS NOT NULL),

    -- The primary key is the universe identifier: the asset ID and group key.
    PRIMARY KEY (asset_id, group_key, proof_type)
);
```

IIUC, with this mode, if I want to allow all insertions, then we need mirrored bools of `allow_remote_insert` and `allow_remote_export` in the general config?

Roasbeef (commented):
> Reviewed 20 of 20 files at r1, 3 of 14 files at r2, 11 of 11 files at r3, 15 of 15 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/537)* status: all files reviewed, 29 unresolved discussions (waiting on @ffranr, @guggero, and @jharveyb)

___
*[`tapdb/universe_federation.go` line 267 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/537#-Ng6aEUjEL4QylqTZJL2:-Ng6aEUjEL4QylqTZJL3:b7m6g8s) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/2af624b6d0fdab5365e9ce2e737d85c922b6e476/tapdb/universe_federation.go#L267)):*
> ```Go
> 						AssetID:   assetIDBytes,
> 						GroupKey:  groupPubKey,
> 						ProofType: proofType.String(),
> ```

Looks like this doesn't also insert the export/insert bool?

taprpc/universerpc/universe.proto
```diff
@@ -145,6 +145,12 @@ message MerkleSumNode {
     // exposed via iterative queries
 }
 
+enum ProofType {
+    UNSPECIFIED = 0;
```
guggero, line 163: Due to a weird constraint in gRPC the individual enum value names must be unique across a whole RPC package. So we couldn't use `TRANSFER` within another enum (even if the enum itself is called differently) without getting a conflict. That's why we normally prefix each enum value name by its enum name. For example: `PROOF_TYPE_TRANSFER`.

universe/interface.go
```diff
@@ -573,6 +573,19 @@ const (
 	ProofTypeTransfer
 )
 
+// NewProofTypeFromAssetProof returns the proof type for the given asset proof.
+func NewProofTypeFromAssetProof(proof *proof.Proof) (ProofType, error) {
+	if proof == nil {
+		return ProofTypeUnspecified, fmt.Errorf("proof is nil")
```
guggero, line 595: micro-nit: for numerical values of a custom type we usually explicitly return `0` just to visually focus more on the error being non-nil.

tapdb/sqlc/queries/universe.sql
```diff
@@ -260,3 +260,33 @@ WHERE event_type IN ('SYNC', 'NEW_PROOF') AND
       event_timestamp >= @start_time AND event_timestamp <= @end_time
 GROUP BY day
 ORDER BY day;
+
+-- name: SetFederationGlobalSyncConfig :exec
+INSERT INTO federation_global_sync_config (
+    proof_type, allow_sync_insert, allow_sync_export
+)
+VALUES (@proof_type, @allow_sync_insert, @allow_sync_export)
+ON CONFLICT(proof_type)
```
guggero, line 269: Since we're overwriting on conflict, I think this should be called `UpsertFederationGlobalSyncConfig`. Would also better match the name used for the universe specific one below.

tapdb/sqlc/queries/universe.sql
```diff
@@ -260,3 +260,33 @@ WHERE event_type IN ('SYNC', 'NEW_PROOF') AND
...
+    DO UPDATE SET
+    allow_sync_insert = @allow_sync_insert,
+    allow_sync_export = @allow_sync_export;
+
+-- name: QueryFederationGlobalSyncConfig :many
```
guggero, line 274: nit: should this also be plural since we're fetching all?

tapdb/universe_federation.go
```diff
@@ -174,4 +214,148 @@ func (u *UniverseFederationDB) LogNewSyncs(ctx context.Context,
...
+			uniID := universe.Identifier{
+				AssetID:  assetID,
+				GroupKey: pubKey,
+			}
+
+			uniConfig := universe.FedUniSyncConfig{
+				UniverseID: uniID,
+				ProofTypes: proofTypes,
+			}
+			uniConfigs = append(uniConfigs, &uniConfig)
```
guggero, line 382: Should we pre-allocate the `uniConfigs` right before the `for` loop and then use index based insertion, since we know how many elements we expect?

guggero (commented):
Getting very close! Main comments are nits (and a couple of not yet resolved ones from my previous review).
I like how the new SQL/table structure looks!

taprpc/universerpc/universe.proto
ffranr: Good call :+1: I've addressed this here https://github.com/lightninglabs/taproot-assets/pull/540

universe/interface.go
ffranr: modified here: https://github.com/lightninglabs/taproot-assets/pull/540

tapdb/sqlc/migrations/000007_universe.down.sql
```diff
@@ -9,3 +9,5 @@ DROP INDEX IF EXISTS universe_roots_asset_id_idx;
 DROP INDEX IF EXISTS universe_roots_group_key_idx;
 DROP INDEX IF EXISTS universe_leaves_key_idx;
 DROP INDEX IF EXISTS universe_leaves_namespace;
+DROP INDEX IF EXISTS federation_general_sync_config;
```
guggero, line 12: There are no such indexes. Did you mean the tables? If yes, then the names are out of sync.

guggero (approved):
LGTM :tada:

tapdb/universe_federation.go
```diff
@@ -174,4 +220,190 @@ func (u *UniverseFederationDB) LogNewSyncs(ctx context.Context,
...
+		)
+		switch {
+		case errors.Is(err, sql.ErrNoRows):
+			// If the query does not return any rows then a global
+			// config for each proof type has not yet been set. We
+			// will return a default config for each proof type.
+			globalConfigs = []*universe.FedGlobalSyncConfig{
+				{
+					ProofType:       universe.ProofTypeIssuance,
+					AllowSyncInsert: true,
```
Roasbeef, line 312: This needs to be reconciled with the command line args we have. Perhaps we eliminate them and have them map here instead? 

Eg: rn the default is to not allow inserting if not specified (even for issuance), but then this flips that.

ffranr: In the latest commits, I've modified such that we default to not allow globally for issuance and transfer proof universes.

Roasbeef (commented):
> Reviewed 3 of 16 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/537)* status: 9 of 22 files reviewed, 37 unresolved discussions (waiting on @ffranr, @guggero, and @jharveyb)

___
*[`rpcserver.go` line 2598 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/537#-NgR_cNz2VdLtfbPVZOR:-NgR_cNz2VdLtfbPVZOS:bvv4pla) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/2af624b6d0fdab5365e9ce2e737d85c922b6e476/rpcserver.go#L2598)):*
> ```Go
> 	return &universe.FedUniSyncConfig{
> 		UniverseID: uniID,
> 		ProofTypes: proofTypes,
> ```

I think we still want this here to distinguish between issuance/transfer trees for insert/export?

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -85,3 +85,52 @@ CREATE VIEW universe_stats AS
...
+    allow_sync_insert BOOLEAN NOT NULL,
+
+    -- This field is a boolean that indicates whether or not a universe of the
+    -- given proof type should accept remote proof export via federation sync.
+    allow_sync_export BOOLEAN NOT NULL
+);
+
+-- This table contains universe (asset/asset group) specific federation sync
+-- configuration.
+CREATE TABLE IF NOT EXISTS federation_uni_sync_config (
```
Roasbeef, line 107: IIUC, in order for something to later on have a foreign key that refs this table, it _does_ in fact still need a primary key. 

I think we just want a composite primary key here. Though I saw you had to re-work things slightly tho, non-blocking comment.

Roasbeef (approved):
LGTM 🧲

PR #540, opened 2023-10-03
Add proof type to universe identifer

ffranr:
Towards the end result of two separate universe trees for each asset/asset group. One tree for issuance proofs and one for transfer proofs.


Roasbeef (changes requested):

universe/interface.go
```diff
@@ -552,6 +552,44 @@ type FederationLog interface {
...
+
+	// ProofTypeTransfer indicates that transfer proofs should be included
+	// in the universe.
+	ProofTypeTransfer
+)
+
+// String returns a human-readable string representation of the sync proof types
+// enum.
+func (t ProofType) String() string {
+	return []string{"none", "issuance", "transfer"}[t]
```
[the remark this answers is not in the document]
guggero: Right, good catch. Since in Goland a type isn't actually an enum, nothing is stopping you from creating `foo := ProofType(1234)`. So maybe just go with the usual switch that returns `unknown proof type <%d>` as the default?

universe/interface.go
```diff
@@ -552,6 +552,44 @@ type FederationLog interface {
 	LogNewSyncs(ctx context.Context, addrs ...ServerAddr) error
 }
 
+// ProofType is an enum that describes the type of proof which can be stored in
+// a given universe.
+type ProofType int
```
guggero, line 596: Can use `uint8` to avoid different types being used on different CPU architectures.

universe/interface.go
```diff
@@ -58,7 +61,21 @@ func (i *Identifier) Bytes() [32]byte {
 
 // MsSmtNamespace returns an MS-SMT namespace which corresponds to the ID.
 func (i *Identifier) MsSmtNamespace() string {
-	return hex.EncodeToString(fn.ByteSlice(i.Bytes()))
+	// Byte serials the group key if it's set, otherwise use the asset ID.
+	var groupKeyAssetIDBytes [32]byte = i.AssetID
```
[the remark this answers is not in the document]
guggero: Yeah, I think we can just use `fmt.Sprintf("%s-%x", i.ProofType, i.Bytes())` and should get the same result.

rpcserver.go
```diff
@@ -2523,15 +2554,41 @@ func (r *rpcServer) AssetRoots(ctx context.Context,
 	return resp, nil
 }
 
+func UnmarshalUniProofType(rpcType unirpc.ProofType) (universe.ProofType,
```
guggero, line 2557: nit: missing Godoc. Also, should we move these marshaling functions into `taprpc/marshal.go`?

guggero (commented):
Did a quick first pass, changes make sense, nice work!

rpcserver.go
ffranr: Can't move into `taprpc/marshal.go` because of `universe` package import.

universe/interface.go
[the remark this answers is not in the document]
ffranr: > I think we can still use Bytes() here to get the initial value we want.

With the addition of the `ProofType` field, `Bytes()` will now return the serialized bytes for two of the tree fields in the `Identifier`. What does that even mean? It's a bit of a strange method. In this PR I re-name it to `MultiverseLeafNodeKey`, which is the only way it's used outside of the `UniverseNamespace` method.

universe/interface.go
```diff
@@ -56,8 +56,8 @@ func (i *Identifier) Bytes() [32]byte {
 	return i.AssetID
 }
 
-// String returns a string representation of the ID.
-func (i *Identifier) String() string {
+// MsSmtNamespace returns an MS-SMT namespace which corresponds to the ID.
```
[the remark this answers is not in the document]
ffranr: I've reverted this change in the latest commits. This rename isn't directly in line with the goal of this PR.

universe/interface.go
```diff
@@ -571,6 +571,19 @@ const (
 	ProofTypeTransfer
 )
 
+// NewProofTypeFromAssetProof returns the proof type for the given asset proof.
+func NewProofTypeFromAssetProof(proof *proof.Proof) (ProofType, error) {
+	if proof == nil {
+		return ProofTypeUnspecified, fmt.Errorf("proof is nil")
+	}
+
+	if proof.Asset.HasGenesisWitness() {
```
Roasbeef, line 596: Also needs `HasGenesisWitnessForGroup` as well. I think we should have some wrapper function/method to unify them so we don't always forget.

universe/interface.go
```diff
@@ -80,6 +80,22 @@ func (i *Identifier) StringForLog() string {
...
 
+// ValidateProofUniverseType validates that the proof type matches the universe
+// identifier proof type.
+func ValidateProofUniverseType(proof *proof.Proof, uniID Identifier) error {
+	expectedProofType, err := NewProofTypeFromAssetProof(proof)
+	if err != nil {
+		return err
+	}
+
+	if expectedProofType != uniID.ProofType {
```
Roasbeef, line 91: 👍

universe/base.go
```diff
@@ -418,6 +418,16 @@ func (a *MintingArchive) getPrevAssetSnapshot(ctx context.Context,
 	prevProofs, err := a.cfg.Multiverse.FetchProofLeaf(
 		ctx, uniID, prevLeafKey,
 	)
+	// If we've failed in finding the previous proof in the transfer
+	// universe, we will try to find it in the issuance universe.
+	if uniID.ProofType == ProofTypeTransfer &&
+		errors.Is(err, ErrNoUniverseProofFound) {
+
+		uniID.ProofType = ProofTypeIssuance
+		prevProofs, err = a.cfg.Multiverse.FetchProofLeaf(
```
Roasbeef, line 447: 👍

universe/base.go
```diff
@@ -298,6 +298,26 @@ func (a *MintingArchive) RegisterNewIssuanceBatch(ctx context.Context,
 	nonAnchorItems := make([]*IssuanceItem, 0, len(items))
 	for ind := range items {
 		item := items[ind]
+
+		// If unspecified, set universe ID proof type based on leaf
+		// proof type.
+		if item.ID.ProofType == ProofTypeUnspecified {
```
Roasbeef, line 304: Much clearer now where/how the `ProofTypeUnspecified` type would be used!

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -12,7 +12,11 @@ CREATE TABLE IF NOT EXISTS universe_roots (
 
     -- We use the 32 byte schnorr key here as this is what's used to derive the
     -- top-level Taproot Asset commitment key.
-    group_key BLOB CHECK(LENGTH(group_key) = 32)
+    group_key BLOB CHECK(LENGTH(group_key) = 32),
+
+    -- This field is an enum representing the proof type stored in the given
+    -- universe.
+    proof_type TEXT NOT NULL CHECK(proof_type IN ('issuance', 'transfer'))
```
Roasbeef, line 19: Also IIUC, `namespace_root` now also includes this as a prefix. May want to update some of the comments in the schema file.

universe/interface.go
```diff
@@ -614,6 +614,22 @@ func (t ProofType) String() string {
 	return fmt.Sprintf("unknown(%v)", int(t))
 }
 
+// NewProofTypeFromString returns the proof type corresponding to the given
+// string.
+func NewProofTypeFromString(typeStr string) (ProofType, error) {
+	switch typeStr {
+	case "unspecified":
```
Roasbeef, line 621: IIUC, we never actually allow the "unspecified" type in the db. After all we need to know _where_ to put the proof.

tapdb/universe_test.go
```diff
@@ -34,6 +34,10 @@ func randUniverseID(t *testing.T, forceGroup bool) universe.Identifier {
 		id.GroupKey = groupKey.PubKey()
 	}
 
+	// Set universe proof type. This is the leaf proof type that will be
+	// used for all leaves in this universe.
+	id.ProofType = universe.ProofTypeIssuance
```
Roasbeef, line 39: May want to. be able to control this later? Could use a functional option here.

rpcserver.go
```diff
@@ -2679,20 +2679,47 @@ func (r *rpcServer) QueryAssetRoots(ctx context.Context,
 		return nil, err
 	}
 
-	rpcsLog.Debugf("Querying for asset root for %v", spew.Sdump(universeID))
-
-	assetRoot, err := r.cfg.BaseUniverse.RootNode(ctx, universeID)
+	// Attempt to retrieve the issuance universe root.
+	rpcsLog.Debugf("Querying for asset (group) issuance universe root "+
+		"for %v", spew.Sdump(universeID))
+	universeID.ProofType = universe.ProofTypeIssuance
```
Roasbeef, line 2685: Style nit: could use a bit of breathing room between stanzas here, eg: 
```go
rpcsLog.Debugf("Querying for asset (group) issuance universe root "

universeID.ProofType = universe.ProofTypeIssuance

	var issuanceRootRPC *unirpc.UniverseRoot
	issuanceRoot, err := r.cfg.BaseUniverse.RootNode(ctx, universeID)
```

rpcserver.go
```diff
@@ -2679,20 +2679,47 @@ func (r *rpcServer) QueryAssetRoots(ctx context.Context,
...
+	var issuanceRootRPC *unirpc.UniverseRoot
```
Roasbeef, line 2686: Does this need to be predeclared?

rpcserver.go
```diff
@@ -2679,20 +2679,47 @@ func (r *rpcServer) QueryAssetRoots(ctx context.Context,
...
+		}
+	}
+	issuanceRootRPC, err = marshalUniverseRoot(issuanceRoot)
 	if err != nil {
 		return nil, err
 	}
 
-	uniRoot, err := marshalUniverseRoot(assetRoot)
+	// Attempt to retrieve the transfer universe root.
+	rpcsLog.Debugf("Querying for asset (group) transfer universe root "+
```
Roasbeef, line 2702: Same here re spacing and predeclaration.

rpcserver.go
```diff
@@ -2679,20 +2679,47 @@ func (r *rpcServer) QueryAssetRoots(ctx context.Context,
...
+		}
+	}
+	transferRootRPC, err = marshalUniverseRoot(transferRoot)
 	if err != nil {
 		return nil, err
 	}
 
 	return &unirpc.QueryRootResponse{
-		AssetRoot: uniRoot,
+		IssuanceRoot: issuanceRootRPC,
```
Roasbeef, line 2721: 👍

rpcserver.go
```diff
@@ -3001,9 +3028,44 @@ func (r *rpcServer) QueryProof(ctx context.Context,
 		"(universeID=%x, leafKey=%x)", universeID,
 		leafKey.UniverseKey())
 
+	// Keep a record of whether the proof type was specified by the client.
+	unspecifiedArgProofType :=
+		universeID.ProofType == universe.ProofTypeUnspecified
```
Roasbeef, line 3053: Can use a parens to avoid the awkward line break here: 
```go
unspecifiedArgProofType := (universeID.ProofType == 
    universe.ProofTypeUnspecified
)
```

rpcserver.go
```diff
@@ -3001,9 +3028,44 @@ func (r *rpcServer) QueryProof(ctx context.Context,
...
+	if unspecifiedArgProofType {
+		// The target proof type was unspecified. Assume that the proof
+		// type is a transfer proof (more transfer proofs exist then
+		// issuance proofs).
+		universeID.ProofType = universe.ProofTypeTransfer
```
Roasbeef, line 3058: Hmm, maybe we should just return an error in this case? Otherwise if they want issuance, then we error out?

rpcserver.go
```diff
@@ -3001,9 +3028,44 @@ func (r *rpcServer) QueryProof(ctx context.Context,
...
 		ctx, universeID, leafKey,
 	)
+	// If we were unable to find a proof, and the proof type was originally
+	// unspecified, then we'll try again with the other proof type.
+	if unspecifiedArgProofType &&
+		errors.Is(err, universe.ErrNoUniverseProofFound) {
+
+		// The proof type was originally unspecified by the client. We
+		// will therefore try again with the other proof type.
+		switch universeID.ProofType {
```
Roasbeef, line 3072: Ah ok, I see the switcheroo here. 

I do wonder if we should just hide this one layer deeper, so at the layer right above the db where methods like `FetchIssuanceProof` are implemented.

itest/assertions.go
```diff
@@ -1038,6 +1038,11 @@ func AssertUniverseRoot(t *testing.T, client unirpc.UniverseClient,
 }
 
 func AssertUniverseRootEqual(a, b *unirpc.UniverseRoot) bool {
+	// Basic RPC form sanity checks.
+	if (a.Id != nil && b.Id == nil) || (a.Id == nil && b.Id != nil) {
```
Roasbeef, line 1042: Wouldn't this be taken care of below?

rpcserver.go
```diff
@@ -2735,6 +2735,26 @@ func (r *rpcServer) DeleteAssetRoot(ctx context.Context,
 
 	rpcsLog.Debugf("Deleting asset root for %v", spew.Sdump(universeID))
 
+	// If the universe proof type is unspecified, we'll delete both the
```
Roasbeef, line 2738: Heh forgot we even had this call....

Good to have though as gives the ability to properly prune a tree if needed (delete all transfer and just go back to issuance, etc). 

Don't think we need to do it here, but I think in the future we'd also want to add a DB record if something should be deleted, then just never added back again.

itest/universe_test.go
```diff
@@ -110,7 +112,12 @@ func testUniverseSync(t *harnessTest) {
 			}
 		}()
 
-		srcRoot, ok := universeRoots.UniverseRoots[uniKey]
+		// Construct universe namespace.
+		proofType, err := tap.UnmarshalUniProofType(newRoot.Id.ProofType)
+		require.NoError(t.t, err)
+		uniNamespace := fmt.Sprintf("%s-%s", proofType, uniKey)
```
Roasbeef, line 118: Hmm, still thinking about if we want to also manifest the prefix on this level, which arguably could/should just be something internal. 

I think will get a better feel of how it all looks on the CLI.

Roasbeef (changes requested):

universe/interface.go
ffranr: Yes that's true. `case "unspecified":` is just dead code right now. I included it to make `NewProofTypeFromString` complete.

rpcserver.go
ffranr: Will remove.

rpcserver.go
ffranr: I get a `syntax error: unexpected newline, expected )` error with the syntax you've suggested.

rpcserver.go
ffranr: The caller (client calling `QueryProof`) may not know if the proof they're querying for is a transfer or issuance proof. The universe RPC proof courier is one such caller. And the caller shouldn't need to specify the proof type. Proof type specification only helps the server figure out where the proof is stored. In other words, the client already provides the necessary args for proof identification.

itest/assertions.go
ffranr: We don't check for `a.Id == nil` below this point which can cause a nil pointer dereference panic.

tapdb/sqlc/migrations/000007_universe.up.sql
ffranr: Right now, the schema doesn't explain what's in `namespace_root`:
```
    -- For the namespace root, we set the foreign key constraint evaluation to
    -- be deferred until after the database transaction ends. Otherwise, if the
    -- root of the SMT is deleted temporarily before inserting a new root, then
    -- this constraint is violated as there's no longer a root that this
    -- universe tree can point to.
    namespace_root VARCHAR UNIQUE NOT NULL REFERENCES mssmt_roots(namespace) DEFERRABLE INITIALLY DEFERRED,
 ```

File: universe/interface.go [hunk 23]
```diff
@@ -92,6 +92,27 @@ func (i *Identifier) StringForLog() string {
 		i.MsSmtNamespace(), i.AssetID[:], groupKey, i.ProofType)
 }
 
+// ValidateProofUniverseType validates that the proof type matches the universe
+// identifier proof type.
+func ValidateProofUniverseType(proof proof.Proof, uniID Identifier) error {
```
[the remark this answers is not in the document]
ffranr: I've added a unit test for this. `ValidateProofUniverseType` is already covered in itests for the positive case. Should I add more on the itest front for this PR?

ffranr:
@Roasbeef @guggero There are universe events stats changes that we need following the addition of the proof type to the universe identifier. I'll open a new PR (based on this PR) to address those changes.

ffranr:
This needs a fixup rebase before it's merged.

rpcserver.go
```diff
@@ -2535,7 +2535,27 @@ func (r *rpcServer) FetchAssetMeta(ctx context.Context,
...
+	switch proofType {
+	case universe.ProofTypeUnspecified:
+		return unirpc.ProofType_PROOF_TYPE_UNSPECIFIED, nil
+	case universe.ProofTypeIssuance:
+		return unirpc.ProofType_PROOF_TYPE_ISSUANCE, nil
+	case universe.ProofTypeTransfer:
+		return unirpc.ProofType_PROOF_TYPE_TRANSFER, nil
+
+	default:
+		return unirpc.ProofType_PROOF_TYPE_UNSPECIFIED,
```
guggero, line 2552: nit: same here re returning an explicit `0` in error cases, would shorten the call as well.

asset/asset.go
```diff
@@ -1077,6 +1077,11 @@ func (a *Asset) HasGenesisWitnessForGroup() bool {
 	return *witness.PrevID == ZeroPrevID
 }
 
+// IsGenesisAsset returns true if an asset is a genesis asset.
+func (a *Asset) IsGenesisAsset() bool {
```
guggero, line 1081: non-blocking nit: there are at least three places where we have this exact condition that we could shorten with this new method.

guggero (approved):
Needs a rebase as mentioned. Other than that only two (very optional/non-blocking) nits, so LGTM :100:

ffranr:
@Roasbeef Please ignore the latest 5 commits, they've already been reviewed here: https://github.com/lightninglabs/taproot-assets/pull/537

Roasbeef (approved):
LGTM 🎎

PR #543, opened 2023-10-03
tapgarden: use lru.Cache for various new group caches

Roasbeef:
This also fixes a bug in `GenRawGroupAnchorVerifier` where the `groupAnchor` value wasn't being properly re assigned.

### DIFF
--- a/tapgarden/caretaker.go
+++ b/tapgarden/caretaker.go
@@ -14,6 +14,8 @@ import (
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/davecgh/go-spew/spew"
+	"github.com/lightninglabs/neutrino/cache/lru"
+
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/commitment"
 	"github.com/lightninglabs/taproot-assets/fn"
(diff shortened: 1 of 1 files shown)


tapgarden/caretaker.go
```diff
@@ -1381,12 +1400,14 @@ func GenRawGroupAnchorVerifier(ctx context.Context) func(*asset.Genesis,
 				return ErrGenesisNotGroupAnchor
 			}
 
-			groupAnchors[assetGroupKey] = gen
```
Roasbeef, line 1384: So here `groupAnchor` wasn't re assigned.

tapgarden/caretaker.go
```diff
@@ -1263,81 +1265,103 @@ func GenHeaderVerifier(ctx context.Context,
...
+
+// emptyVal is a simple type def around struct{} to use as a dummy value in in
+// the cache.
+type emptyVal struct{}
+
+// singleCacheValue is a dummy value that can be used to add an element to the
+// cache. This should be used when the cache just needs to worry aobut the
+// total number of elements, and not also the size (in bytes) of the elements.
+type singleCacheValue[T any] struct {
+	val T
```
Roasbeef, line 1279: You can't embed type param values, so had to use this here.

tapgarden/caretaker.go
```diff
@@ -1263,81 +1265,103 @@ func GenHeaderVerifier(ctx context.Context,
...
+// newSingleValue creates a new single cache value.
+func newSingleValue[T any](v T) singleCacheValue[T] {
+	return singleCacheValue[T]{
+		val: v,
+	}
+}
+
+// emptyCacheVal is a type def for an empty cache value. In this case the cache
+// is used more as a set.
+type emptyCacheVal = singleCacheValue[emptyVal]
```
Roasbeef, line 1296: Type alias used here to it inherits all the methods.

tapgarden/caretaker.go
Roasbeef: Ah returns nil anyway, so np.

tapgarden/caretaker.go
```diff
@@ -1263,81 +1265,103 @@ func GenHeaderVerifier(ctx context.Context,
...
 
 	// Cache anchors for groups that were previously fetched.
-	groupAnchors := make(map[asset.SerializedKey]*asset.Genesis)
-
-	// verifierLock is a mutex that must be held when using the group
-	// anchor verifier.
-	var verifierLock sync.Mutex
+	groupAnchors := lru.NewCache[
+		asset.SerializedKey, singleCacheValue[*asset.Genesis]](
+		assetGroupCacheSize,
```
jharveyb, line 1345: IIUC with a 10k element cache  this would be a few MB max?

jharveyb (approved):
LGTM  🚀 

Will be useful pattern for safely adding caching to some other callbacks as well.

Issue #544, opened 2023-10-03
universe: resolve overflow issues, use distinct issuance + transfer trees


Roasbeef:
This is a simple variant to: https://github.com/lightninglabs/taproot-assets/issues/479. Rather than introduce 128-bit integers, which'll need some custom DB extensions to support properly, we instead aim to go with a simpler approach, leaving room for future protocol versions to go with a 128-bit sum value. 

## Steps To Completion

- [ ] Enforce a max issuance value of a 32-bit value. This just restricts the amount that a single issuance UTXO can create. A single transaction can create more units with bundled UTXOs in the some genesis transaction. Most assets won't feasibly run through the entire 64-bit unit space. For the Universe issuance trees for grouped assets, the limit applies there as well. 
- [ ] Modify the Universe transfer tree to use a sum value of 1 everywhere. This makes the value more useful, as it's no longer summing different units. Instead it represents a count of the total number of registered transfers. 
- [ ] Split the issuance + transfer trees across the RPC API and database. This simplifies syncing, as you have two roots, if one isn't present, then you know that they don't have transfers or otherwise.

Roasbeef:
Fixed by #564

PR #545, opened 2023-10-03
asset: limit max value of issuance leaf to math.MaxUint32

Roasbeef:
This is a step towards fixing:
https://github.com/lightninglabs/taproot-assets/issues/544.

In this commit, we limit the max amount of an asset that can be issued
at one time to math.MaxUint32. The existing 32-bit sum is left in place.
This gives us enough buffer room, as we'd need 4 billion UTXOs of the
same assets to overflow the root sum. In the future, we can lift this
limit with future asset versions.

An asset can have more than 4 billion ish units if they issue with
distinct UTXOs. This value can be viewed as the max amt that can be
issued in a single tranche.


jharveyb:
Thinking that the amount field for addresses should now also be a uint32? And receiving more than that requires the interactive flow or multiple addresses.

Roasbeef:
> Thinking that the amount field for addresses should now also be a uint32?

Hmm, yeah good point, hadn't thought about that interaction. In that case, it's possible to actually just re-use the address if needed.

jharveyb:
Should also update the overflow checks we have for tree leaves so we can fail earlier than `asset.New()`, e.x. in the porter:

https://github.com/lightninglabs/taproot-assets/blob/5373ccceaa92a6b4e32ff55c2e272608429fbbc5/tapscript/send.go#L323

RPC types also need to be updated, like `MintAssetRequest`, so we don't ingest uint64s where relevant.

jharveyb:
> > Thinking that the amount field for addresses should now also be a uint32?
> 
> Hmm, yeah good point, hadn't thought about that interaction. In that case, it's possible to actually just re-use the address if needed.

Unrelated but would be neat to signal to the sender that those two asset leaves could be merged into the same anchor output; maybe if they share the same internal key?

Could also be useful for handling receiving for asset groups.

PR #546, opened 2023-10-04
build: prep for Go 1.22 by using GOEXPERIMENT=loopvar for tests 

Roasbeef:
In this commit, [we prep for some upcoming changes in
Go](https://go.dev/blog/loopvar-preview) by running our tests with
`GOEXPERIMENT=loopvar`. This changes the loop semantics to fix a common
bug where a scoping issue causes a variable to be re-used, which can
cause unintended bugs down the line.

If everything passes with this flag on, then we can keep it on, and also
rest a bit easier knowing that the compiler will fix a class of bugs
that would previously pop up for us.


guggero:
> `go: go.mod file indicates go 1.21, but maximum version supported by tidy is 1.20`

I think we need to use an explicit version of Go in the `mod-check` job in `.github/workflows/main.yaml`, as the default of the GitHub Actions runner seems to be `v1.20`.
Adding this should fix:

```
      - name: Setup go environment
        uses: ./.github/actions/setup-go
```

Roasbeef:
Updated with that. Will land this if CI green, one less thing to have to think about when writing+reviewing code.

Issue #347, opened 2023-06-08
Add comprehensive versioning to all protocol elements and even/odd compliance structure


dstadulis:
### Background
---
As tapd is upgraded and the protocol evolves, there will be a need by clients to deduce if the versions of various protocol elements (addresses, proof files, etc) need to known about by taproot-asset clients prescriptively or if the version is merely an indication of an optional update.  In order for the prescription functionality of version to be "self authenticated" (i.e. knowable without being previously known), the versioning method of "it's ok to be odd" (e.g. odd version are optional, even version are mandatory) will be used (See lightning BOLTs: https://github.com/lightning/bolts/commit/296075ec96d7788c0cf598e0ac24ef2de8b4a3a4)

### Compatibility
---
Implementing this feature will be a breaking change which we tolerate because of tapd's alpha status and will render large upgrading and correctness benefits over all operations when the daemon progresses to beta status.

### Deliverables
---
- [ ] Add comprehensive versioning to all protocol elements and even/odd compliance structure
  - [x] Add versioning to `addr`
https://github.com/Roasbeef/bips/blob/da3049f2f65b1c1c2b2caa7164449192b791bd29/bip-tap-addr.mediawiki?plain=1#L87-L96
 - [x] #392
  https://github.com/lightninglabs/taproot-assets/blob/7a5d5cae25f980c0ead87c0578584eafee689624/proof/proof.go#L80

### Necessary preceding work
- [ ] https://github.com/Roasbeef/bips/pull/33
  - [ ] implementing a new version for `Asset`, V1, #381 which would include:
  - [ ] Replace the `Genesis` field with a 32-byte `AssetID`, where the genesis data is provided in the genesis proof.
  - [ ] Update the `GroupKey` field to be 32 bytes instead of 96, to store only the tweaked group key.
  - [ ] Update encode/decode for Witness to follow even/odd rules for v1 assets
  - [ ] Update the TLV types to be even/odd compliant instead of sequential as they are now.

### Open Questions
---
#### 1
>Would a revision to an existing odd version, which isn't necessary, mean that the version would skip the even amount and progress directly to the next odd number?

Assume it would.

#### [2](https://github.com/lightninglabs/taproot-assets/issues/347#issuecomment-1611678264)
>for any existing v0 asset, because a v1 asset requires a new proof format that includes asset_genesis_reveal and group_key_reveal, all proofs for a v0 asset would have to be recomputed to move this data out of the asset and into the proof file.

[...]

PR #547, opened 2023-10-04
Use even/odd TLV numbers everywhere

guggero:
Definitely and finally closes https://github.com/lightninglabs/taproot-assets/issues/347.

Corresponding BIP PR is here: https://github.com/Roasbeef/bips/pull/43


commitment/records.go
```diff
@@ -10,14 +10,14 @@ import (
...
+	AssetProofAssetIDType tlv.Type = 2
+	AssetProofType        tlv.Type = 4
 
 	TaprootAssetProofVersionType tlv.Type = 0
-	TaprootAssetProofType        tlv.Type = 1
+	TaprootAssetProofType        tlv.Type = 2
 
-	ProofAssetProofType        tlv.Type = 0
-	ProofTaprootAssetProofType tlv.Type = 1
+	ProofAssetProofType        tlv.Type = 1
```
Roasbeef, line 19: Both should be even?

proof/records.go
```diff
@@ -12,37 +12,37 @@ import (
...
-	ChallengeWitnessType tlv.Type = 11
-	BlockHeightType      tlv.Type = 12
-	GenesisRevealType    tlv.Type = 13
-	GroupKeyRevealType   tlv.Type = 14
+	PrevOutType          tlv.Type = 2
+	BlockHeaderType      tlv.Type = 4
+	AnchorTxType         tlv.Type = 6
+	TxMerkleProofType    tlv.Type = 8
+	AssetLeafType        tlv.Type = 10
+	InclusionProofType   tlv.Type = 12
```
Roasbeef, line 20: One thing we even odd that we could do here is actually assign in pairs. We do this for feature bits in LN so we can flip them to required if needed. 

In this case, we would have `ExclusionProofsType` actually be `15`. 

Not sure if this is useful though?

Roasbeef (commented):
Looks pretty good, just a few areas for discussion re if they should be even or odd.

asset/records.go
```diff
@@ -15,16 +15,16 @@ type LeafTlvType = tlv.Type
...
-	LeafSplitCommitmentRoot LeafTlvType = 7
-	LeafScriptVersion       LeafTlvType = 8
-	LeafScriptKey           LeafTlvType = 9
-	LeafGroupKey            LeafTlvType = 10
+	LeafGenesis             LeafTlvType = 2
+	LeafType                LeafTlvType = 4
+	LeafAmount              LeafTlvType = 6
+	LeafLockTime            LeafTlvType = 7
+	LeafRelativeLockTime    LeafTlvType = 9
+	LeafPrevWitness         LeafTlvType = 11
```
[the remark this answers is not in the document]
guggero: My answer is going to look very similar for most of the comments I think: I mainly went off of the code in `EncodeRecords` and if there was a condition around the field, I declared it as optional.
For example the previous witness: https://github.com/lightninglabs/taproot-assets/blob/5373ccce/asset/asset.go#L1202

I can of course change that code to make the previous witness field mandatory. It seems like we already have a check for that in the VM (`validateStateTransition`).

asset/records.go
```diff
@@ -35,9 +35,9 @@ const (
 type WitnessTlvType = tlv.Type
 
 const (
-	WitnessPrevID          WitnessTlvType = 0
-	WitnessTxWitness       WitnessTlvType = 1
-	WitnessSplitCommitment WitnessTlvType = 2
+	WitnessPrevID          WitnessTlvType = 1
+	WitnessTxWitness       WitnessTlvType = 3
+	WitnessSplitCommitment WitnessTlvType = 5
```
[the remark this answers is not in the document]
guggero: Same here, both the `PrevID` and `TxWitness` fields have an explicit check in `EncodeRecords`: https://github.com/lightninglabs/taproot-assets/blob/5373ccceaa92a6b4e32ff55c2e272608429fbbc5/asset/asset.go#L340

I guess if we declare `PrevID` as required, it would no longer need to be a pointer? Not sure how to decide what to take as the source of truth for the decision: our current implementation or the fact that it should in practice never be empty...

commitment/records.go
guggero: Oops, no, I think they should be the other way around: `ProofAssetProofType` should be odd and `ProofTaprootAssetProofType` should be even: https://github.com/lightninglabs/taproot-assets/blob/5373ccceaa92a6b4e32ff55c2e272608429fbbc5/commitment/proof.go#L64

proof/records.go
guggero: Ah, I see. Yes, that could be nice so we could turn things on/off when it comes to being required. Happy to do the change if you think we should. But then I'd say we should update the address records as well.

asset/records.go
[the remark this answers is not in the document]
jharveyb: FWIW I'm ok with leaving the locktimes as 'inferred 0'.

asset/records.go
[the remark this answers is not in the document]
jharveyb: Indeed, its the `ZeroPrevID` in both forms of the genesis witness, and the VM already fails if its nil. In favor of promoting to not-pointer.

TxWitness is fine as-is though IIUC should only be empty for genesis assets.

proof/records.go
Roasbeef: Don't feel super strongly about this, as in the future if we wish we can use a new version to redo all the types. So far we don't have a concrete need for this as is though.

Roasbeef (approved):
LGTM 📣

Roasbeef:
Lost track of if was this PR or another, but: I think we can pull out the changes to the restriction/validation of the meta field into another PR.

proof/records.go
guggero: Yeah, me neither... So I guess let's keep it the way they are now?

guggero:
> Lost track of if was this PR or another, but: I think we can pull out the changes to the restriction/validation of the meta field into another PR.

I think what you mean was in https://github.com/lightninglabs/taproot-assets/pull/525, which was merged by now.

jharveyb (approved):
Great to have this sorted 💯

PR #549, opened 2023-10-04
Increase group witness test coverage

### LINKED ISSUE #511: Add script-path validation unit tests for group_key witness scenarios
@jharveyb to list issue deliverables

jharveyb:
Fixes #511. Tracking the canonical checklist here:

- [x] Manually generate group witness with a non-empty tapscript root (BIP86, key spend, script spend)
- [x] Verify custom witnesses in VM
- [ ] Validate proof with assets with custom witnesses
- [ ] Exercise new VM validation branches
- [ ] Exercise group and group anchor verifiers + sortassets
- [ ] Coverage for `GroupPubKey()`, `DeriveGroupKey()`, and new Has/NeedsGenesisWitnessForGroup helpers
- [ ] Exercise new branches in `proof.Verify()` concerning group key validation
- [ ] Validate group key reveal handling in `committedProofs()`
- [ ] Validate store & load of multi-element group witnesses


jharveyb:
Moved to `mint-rpc-fixes` as a base to inherit other fixes related to group key handling.

Roasbeef:
Would be good to thread through some of the new unit tests into test vectors as well.

asset/asset.go
```diff
@@ -861,6 +862,130 @@ func DeriveGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
...
+
+	if signDesc.DoubleTweak != nil {
+		return nil, fmt.Errorf("bad sign descriptor for group key")
+	}
+
+	// Compute the tweaked group key and set it in the asset before
+	// creating the virtual minting transaction.
+	genesisTweak := initialGen.ID()
+	tweakedGroupKey, err := GroupPubKey(
+		signDesc.KeyDesc.PubKey, genesisTweak[:], signDesc.TapTweak,
```
Roasbeef, line 908: Is the `signDesc` actually needed here if we're already obtaining the witness? IIUC, the last arg here should be the tapscript _root_ for the script group tree.

asset/asset.go
```diff
@@ -861,6 +862,130 @@ func DeriveGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
...
+			RawKey:        signDesc.KeyDesc,
+			GroupPubKey:   *tweakedGroupKey,
+			TapscriptRoot: signDesc.TapTweak,
+			Witness:       witness,
+		}, nil
+	}
+
+	// Populate the remaining fields of the signing descriptor needed to
+	// sign the virtual minting transaction.
+	signDesc.SingleTweak = genesisTweak[:]
```
Roasbeef, line 970: IIUC, the double tweak is also needed here?

tapdb/asset_minting_test.go
```diff
@@ -421,17 +422,18 @@ func seedlingsToAssetRoot(t *testing.T, genesisPoint wire.OutPoint,
...
 		require.NoError(t, err)
 
-		// Finally make a new asset commitment (the inner SMT tree) for
-		// this newly created asset.
-		assetRoot, err := commitment.NewAssetCommitment(
-			newAsset,
-		)
-		require.NoError(t, err)
+		if groupKey != nil {
+			engine, err := vm.New(newAsset, nil, nil)
```
Roasbeef, line 426: Hmm perhaps the unit tests on the higher level are enough? Feels strange to run VM checks at the DB abstraction level.

asset/asset.go
Roasbeef: IIUC, needed if it's a keypath spend where the group key commits to an actual script tree.

asset/asset.go
```diff
@@ -861,6 +866,134 @@ func DeriveGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
...
+		signDesc.SignMethod = input.TaprootScriptSpendSignMethod
+		signDesc.WitnessScript = tapLeaf.Script
+
+	default:
+		return nil, fmt.Errorf("bad sign descriptor for group key")
+	}
+
+	// Exit early if a group witness is already given, since we don't need
+	// to produce a signature.
+	if scriptWitness != nil {
```
Roasbeef, line 950: I think we can move this up even further: if the witness is already set, then all we need to do is set that field. We can go a step further and run validation here to make sure it's property formatted. 

Also wouldn't we assume the witness is fully formed, so includes the leaf script and the control block already?

tapdb/asset_minting_test.go
jharveyb: Agree, though this mirrors the sprouting step in the caretaker which also now runs the VM. Though we don't make interesting witnesses here so could remove.

asset/asset.go
jharveyb: I think there is no double tweak, only the single + tap tweak? Where the tap tweak was set by the caller.

asset/asset.go
jharveyb: True re: first point.

On the second, I think you're right - I had it this way to mimic the existing test utils vs. a real user.

jharveyb:
Split off fixes and simple test cases that don't make use of a custom group key to #598.

The taproot utils used for these tests should be made more general so they could be used in the eventually-exposed functions for making complex group keys.

jharveyb:
Rebased, this PR also has some unit test coverage unrelated to group key witnesses that will be split off into a separate PR.

The problematic group key witness tests are `collectible_group_anchor_BIP86_key` and `normal_group_anchor_key_spend`, still hunting down the cause.

jharveyb:
Rebased on #717 , depends on some helpers added there.

tapdb/asset_minting_test.go
jharveyb: Removed in latest version.

jharveyb:
Strangely, this failure only triggers for me when running tests from the IDE, not via Makefile.

dstadulis:
@jharveyb What else needs to be added to this before the PR is out of draft phase?

asset/asset.go
jharveyb: True, I missed that it's converted into the tweak inside `ComputeTaprootOutputKey`.

asset/asset.go
jharveyb: From the docs it looks like `DoubleTweak` is only used in the LN context and mutually exclusive with `SingleTweak`?

I think for key path spend we need a different tapscript tree rootHash as we aren't revealing any leaves.

jharveyb:
> @jharveyb What else needs to be added to this before the PR is out of draft phase?

Need to fix my tapscript building test code so all these test cases pass; don't think we need new test cases outside of what I added here.

asset/asset.go
```diff
@@ -798,39 +820,57 @@ func NewScriptKeyBip86(rawKey keychain.KeyDescriptor) ScriptKey {
...
 }
 
-// DeriveGroupKey derives an asset's group key based on an internal public
-// key descriptor, the original group asset genesis, and the asset's genesis.
-func DeriveGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
-	rawKey keychain.KeyDescriptor, initialGen Genesis,
-	newAsset *Asset) (*GroupKey, error) {
+// ValidateGroupKeyRequest ensures that the asset intended to be a member of an
+// asset group is well-formed.
+func (req *GroupKeyRequest) ValidateGroupKeyRequest() error {
```
ffranr, line 826: We could create a `NewGroupKeyRequest` function and add these validation checks within that function. That way we would avoid the risk of forgetting to call `ValidateGroupKeyRequest`. Or we could do both because the validation is cheap.

Either way, we should probably name this method `Validate`.

asset/asset.go
```diff
@@ -905,6 +906,116 @@ func DeriveGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
...
+
+	sig, err := genSigner.SignVirtualTx(signDesc, genesisTx, prevOut)
+	if err != nil {
+		return nil, err
+	}
+
+	// If this was a script spend, we also have to add the script itself and
+	// the control block to the witness, otherwise the verifier will reject
+	// the generated witness.
+	witness := wire.TxWitness{sig.Serialize()}
```
ffranr, line 1005: This line should probably be moved up above the comment. Because the comment relates to `signDesc.SignMethod == input.TaprootScriptSpendSignMethod` but this line does not.

ffranr (changes requested):
Looking good so far.

asset/asset.go
```diff
@@ -798,39 +820,57 @@ func NewScriptKeyBip86(rawKey keychain.KeyDescriptor) ScriptKey {
...
+	// membership.
+	if req.NewAsset == nil {
+		return fmt.Errorf("grouped asset cannot be nil")
+	}
 
-	// First, perform the final checks on the asset being authorized for
-	// group membership.
-	if newAsset == nil {
-		return nil, fmt.Errorf("grouped asset cannot be nil")
+	if !req.NewAsset.HasGenesisWitness() {
```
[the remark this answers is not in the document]
jharveyb: That doesn't have the same meaning; `IsGenesisAsset()` allows for an asset to have a group key set (if it also has a non-empty `TxWitness`). Here we should only accept an asset with no group key nor genesis witness set, as we generate both after handling this request.

The input asset here should be created when promoting asset seedlings to sprouts in the caretaker, so it's mostly just double-checking that the wrong type of asset isn't being passed here.

Added further comments around these checks.

asset/asset.go
jharveyb: Updated the name; fair point re: having this happen in some constructor instead of a separate func as it is now, I guess the `TapscriptRoot` is the only optional field.

jharveyb:
Rebased and added a constructor that also does request validation, ready for re-review.

ffranr (approved):
Nice work, great opportunity for me to review a few concepts, thanks.

asset/mock.go
```diff
@@ -309,6 +309,90 @@ func SignVirtualTx(priv *btcec.PrivateKey, signDesc *lndclient.SignDescriptor,
 	return sig, nil
 }
 
+// AssetCustomGroupKey constructs a group anchor that uses a group key with
+// a tapscript root that includes a hash lock script and signature script.
+func AssetCustomGroupKey(t *testing.T, useHashLock, BIP86, keySpend, valid bool,
```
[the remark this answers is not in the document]
jharveyb: Yes, your doc is correct! Will update this one, feels like a stub. Thinking we should add the definition of group anchor into the BIPs somewhere tbh.

The group anchor the first asset of the group, whose genesis information is used as the single tweak for deriving the group key. This function does indeed only create group anchors, as the asset we're signing,`protoAsset`, has the same genesis as is used in the `GroupKeyRequest`.

`keySpend` includes a non-empty tapscript root in the request, but the group key witness is created via key spend.

The `BIP86` option sets the signing method before we produce a witness with the `GenesisSigner`, and is equivalent to having an empty tweak in the group key request (the internal key is tweaked only with itself).

Will add further comments throughout.

vm/vm_test.go
```diff
@@ -444,6 +444,19 @@ func splitCollectibleStateTransition(validRoot bool) stateTransitionFunc {
 	}
 }
 
+func complexGroupAnchorStateTransition(useHashLock, BIP86, keySpend, valid bool,
```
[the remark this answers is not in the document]
jharveyb: Was following the naming for the other helpers here; all of these just create closures that return input variables for VM execution, and those inputs represent a state transition for the input asset.

For this specific helper it ended up as a thin wrapper around the group key derivation logic, which is fine as we just want to validate group anchor witnesses.

asset/mock.go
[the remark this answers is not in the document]
jharveyb: Expanded the comment s.t. it should be easier for others to pick up for their own tests later on; some flags are ignored if others are set (like `useHashLock` if `BIP86` is set).

Issue #539, opened 2023-10-02
[feature]: allow the `loadtest` binary to run repeatedly against persistent `tapd` instances


calvinrzachman:
### Background

We are curious to use the `loadtest` binary to put the `tapd` binary under stress and see how it performs. The test, as currently structured, can't be run multiple times in succession without wiping the backing stores/databases of the `tapd` instances under test.

It looks like the [current test](https://github.com/lightninglabs/taproot-assets/blob/main/itest/loadtest/mint_batch_test.go#L62) ran by the orchestrator binary is modeled heavily off an [existing itest](https://github.com/lightninglabs/taproot-assets/blob/main/itest/mint_batch_stress_test.go#L95) for batch minting assets. My guess is that the itest assumes that it will be dealing with fresh instances of tapd on each test iteration. We would prefer to allow repeated test iterations to run against persistent/long-lived tapd instances to track the performance of the daemon over time as multiple batches of assets (possibly of varying size, content, etc.) are minted. This may require a few changes to the test binary.

I can dig a bit deeper into the changes which might be necessary for allowing this, but for now here are the relevant portions of the testing binary code, based on the errors we run into (outlined below):
- Wait for a single TX in the mempool: https://github.com/lightninglabs/taproot-assets/blob/main/itest/utils.go#L186
- Assert that the single TX in the mempool has a certain “asset anchor” (TODO: not sure what this is quite yet): https://github.com/lightninglabs/taproot-assets/blob/main/itest/utils.go#L208
    - Do we need this assertion? If so, we will probably need to take care after each test failure that we clear the mempool (mine a block) so that we always only have the single expected TX in the mempool.
        - Even if we clear the mempool we keep getting this error. So far have only been able to get past this by wiping the DB of the primary `tapd` instance. 

### Your environment

- btcd version `0.23.3-beta`
```sh
/ # tapcli --network=regtest --tapddir=/data getinfo
{
    "version": "0.2.3-alpha commit=v0.2.3-228-g56d5c50",
    "lnd_version": "0.17.0-beta.rc2",
    "network": "regtest"
}
/ # uname -mrsv
Linux 5.4.209-116.367.amzn2.x86_64 #1 SMP Wed Aug 31 00:09:52 UTC 2022 x86_64
```

### Steps to reproduce

- Deploy and run the `loadtest` binary against a test environment consisting of two `tapd` instances + `btcd/lnd` stack.
- Observe the batch mint assets test cannot be executed multiple times without failure
[...]

PR #550, opened 2023-10-05
loadtest: prepare for multiple consecutive runs

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/539.

This PR allows the load test to be run against both `btcd` and `bitcoind` and also makes it ready for being run multiple times in a row against the same node.

cc @calvinrzachman

### DIFF
--- a/itest/assertions.go
+++ b/itest/assertions.go
@@ -932,12 +932,17 @@ func AssertSplitTombstoneTransfer(t *testing.T,
 func AssertNumGroups(t *testing.T, client taprpc.TaprootAssetsClient,
 	num int) {
 
+	require.Equal(t, num, NumGroups(t, client))
+}
+
+// NumGroups returns the current number of asset groups present.
+func NumGroups(t *testing.T, client taprpc.TaprootAssetsClient) int {
 	ctxb := context.Background()
 	groupResp, err := client.ListGroups(
 		ctxb, &taprpc.ListGroupsRequest{},
 	)
 	require.NoError(t, err)
-	require.Equal(t, num, len(groupResp.Groups))
+	return len(groupResp.Groups)
 }
 
 // AssertGroupSizes asserts that a set of groups the daemon is aware of contain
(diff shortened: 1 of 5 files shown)


itest/utils.go
```diff
@@ -20,7 +22,9 @@ import (
 )
 
 var (
-	zeroHash chainhash.Hash
+	zeroHash          chainhash.Hash
+	regtestMiningAddr = "n1VgRjYDzJT2TV72PnungWgWu18SWorXZS"
```
Roasbeef, line 26: Is seed generation by the nodes deterministic? 

Do we need to also move to simnet, so bitcoind won't wipe the chain on start up?

itest/mint_batch_stress_test.go
```diff
@@ -183,10 +183,7 @@ func mintBatchStressTest(
 	require.NoError(t, err)
 	require.Len(t, uniRoots.UniverseRoots, groupCount)
 
-	err = AssertUniverseRoot(
-		t, alice, groupBalance, nil, collectGroupKey,
-	)
-	require.NoError(t, err)
+	AssertUniverseRoot(t, alice, groupBalance, nil, collectGroupKey)
```
Roasbeef, line 186: Wouldn't this just fail the test either way?

itest/utils.go
guggero: This is just to send to `bitcoind` to specify _where_ to mine the the coins to. We don't really ever use that address for anything else.

No, it's `btcd` that drops the chain on startup, not `bitcoind`. So we can switch batch to `bitcoind` and `regtest`.

itest/mint_batch_stress_test.go
guggero: Yeah, but before it was a mixture of `require` and returning an error. And now we just do `t.Fatalf()` directly instead of returning the error and then doing `require.NoError()` outside.

calvinrzachman (commented):
Thanks for helping me out here! Just had a similar question to Laolu re: the hardcoded mining address. I will give the changes a try in `staging`

Roasbeef (approved):
LGTM 🪁

itest/loadtest/mint_batch_test.go
```diff
@@ -194,7 +196,18 @@ func mintBatchStressTest(t *testing.T, ctx context.Context,
 			},
 		},
 	)
-	require.NoError(t, err)
+	if err != nil {
+		require.ErrorContains(t, err, "universe server already added")
```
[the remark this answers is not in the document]
guggero: Good point! Though we cannot use `ErrorIs` because the error is being sent over RPC and cannot directly be unwrapped/matched. But we can still use the error constant in `ErrorContains`, which I'll fix.

itest/assertions.go
```diff
@@ -990,13 +990,15 @@ func AssertUniverseRootEquality(t *testing.T,
...
 func AssertUniverseRoot(t *testing.T, client unirpc.UniverseClient,
-	sum int, assetID []byte, groupKey []byte) error {
+	sum int, assetID []byte, groupKey []byte) {
 
 	bothSet := assetID != nil && groupKey != nil
 	neitherSet := assetID == nil && groupKey == nil
 	if bothSet || neitherSet {
-		return fmt.Errorf("only set one of assetID or groupKey")
+		t.Fatalf("only set one of assetID or groupKey")
 	}
```
ffranr: If we're not going to return an error at this point then we can just use `require.True` and populate the message arg.

itest/loadtest/mint_batch_test.go
```diff
@@ -194,7 +197,22 @@ func mintBatchStressTest(t *testing.T, ctx context.Context,
 			},
 		},
 	)
-	require.NoError(t, err)
+	if err != nil {
+		// Only fail the test for duplicate universe errors, as we might
+		// have already added the server in a previous run.
+		require.ErrorContains(
+			t, err, universe.ErrDuplicateUniverse.Error(),
+		)
```
ffranr, line 205: Do the the comment and `require` statement align here?

> fail the test for duplicate universe errors

But then the test pass if the error contains `universe.ErrDuplicateUniverse`. I think the comment should be:
```
		// Only fail the test if the error is not a duplicate universe error, as we might
		// have already added the server in a previous run.
```

ffranr (approved):
Looks ok to me, just two nits.

itest/loadtest/mint_batch_test.go
guggero: Ah, you're right, wanted to write `Only fail the test for other errors than duplicate universe errors`, fixed.

Issue #541, opened 2023-10-03
Ensure all instances of `clock.Now()` are `u.clock.Now().UTC()`


ffranr:
I'm pretty sure this unit test fails between 00:00 and 01:00
https://github.com/lightninglabs/taproot-assets/blob/main/tapdb/universe_stats_test.go#L199-L199

From what I understand the solution is to ensure that every instance of `clock.Now()` in code becomes `u.clock.Now().UTC()`.

Also, related to this unit test failure: apparently there are also `EventTime` db fields which are not used and can be removed. But perhaps we can push that change to another issue for now and keep things simple.

PR #551, opened 2023-10-05
Fix instances of `time.Now()` which should be set to UTC

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/541

I have confirmed that the failing unit test now passes.

I think we should enforce UTC location in `clock.NewTestClock` (LND). Thoughts on this @guggero ?

### DIFF
--- a/tapdb/universe_stats.go
+++ b/tapdb/universe_stats.go
@@ -137,7 +137,7 @@ func (u *UniverseStats) LogSyncEvent(ctx context.Context,
 		}
 
 		return db.InsertNewSyncEvent(ctx, NewSyncEvent{
-			EventTime:      u.clock.Now(),
+			EventTime:      u.clock.Now().UTC(),
 			EventTimestamp: u.clock.Now().UTC().Unix(),
 			AssetID:        uniID.AssetID[:],
 			GroupKeyXOnly:  groupKeyXOnly,
@@ -162,7 +162,7 @@ func (u *UniverseStats) LogSyncEvents(ctx context.Context,
 			}
 
 			err := db.InsertNewSyncEvent(ctx, NewSyncEvent{
-				EventTime:      u.clock.Now(),
+				EventTime:      u.clock.Now().UTC(),
 				EventTimestamp: u.clock.Now().UTC().Unix(),
 				AssetID:        uniID.AssetID[:],
 				GroupKeyXOnly:  groupKeyXOnly,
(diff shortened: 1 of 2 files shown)


guggero (approved):
LGTM :tada:

guggero:
> I think we should enforce UTC location in clock.NewTestClock (LND). Thoughts on this @guggero ?

Yes, that sounds good. Perhaps we can simply create a `UTCClock` implementation of the clock interface in Tapd and then replace `Now()` with `Now().UTC()`.

jharveyb:
We have some other writes that should probably be UTC?

https://github.com/lightninglabs/taproot-assets/blob/5373ccceaa92a6b4e32ff55c2e272608429fbbc5/tapfreighter/parcel.go#L353

Coin lease duration seems like local time rn instead of UTC also:

https://github.com/lightninglabs/taproot-assets/blob/5373ccceaa92a6b4e32ff55c2e272608429fbbc5/tapfreighter/wallet.go#L236
https://github.com/lightninglabs/taproot-assets/blob/5373ccceaa92a6b4e32ff55c2e272608429fbbc5/tapfreighter/chain_porter.go#L974

IIUC we should query the DB in UTC as well:

https://github.com/lightninglabs/taproot-assets/blob/5373ccceaa92a6b4e32ff55c2e272608429fbbc5/rpcserver.go#L3467
https://github.com/lightninglabs/taproot-assets/blob/5373ccceaa92a6b4e32ff55c2e272608429fbbc5/cmd/tapcli/addrs.go#L127

https://github.com/lightninglabs/taproot-assets/blob/5373ccceaa92a6b4e32ff55c2e272608429fbbc5/tapdb/universe_federation.go#L118

https://github.com/lightninglabs/taproot-assets/blob/5373ccceaa92a6b4e32ff55c2e272608429fbbc5/tapgarden/custodian.go#L531

We also still have a bunch of non-UTC usage in tests and the test mock utilities.

guggero:
@jharveyb I don't think we need to do anything outside of the `tapdb` package. Because for example in the RPC interface we use `.Unix()` which implicitly does UTC. This is specifically a database issue (the way sqlc and the DB driver format the time), not a general issue in my opinion.

So we should only make sure we use `.UTC()` before storing a timestamp in the DB.

jharveyb:
Ah I see, the timezone we use elsewhere won't matter since the DB normalizes to UTC before write I guess?

I looked through the DB logic for other tables where we have timestamps and they look ok.

jharveyb (approved):
Good catch 👍🏽

PR #554, opened 2023-10-09
feat: Extend GetInfo RPC with LND info.

ben2077:
Added additional fields to the GetInfo RPC response to provide richer details about the blockchain and the LND node. This includes the LND node's public key (identity), its alias, the current block height, and the latest block's hash.


rpcserver.go
```diff
@@ -286,10 +286,29 @@ func (r *rpcServer) DebugLevel(ctx context.Context,
 func (r *rpcServer) GetInfo(context.Context,
 	*taprpc.GetInfoRequest) (*taprpc.GetInfoResponse, error) {
 
+	// Retrieve the best block hash and height from the chain backend.
+	blockHash, blockHeight, err := r.cfg.Lnd.ChainKit.GetBestBlock(context.Background())
```
guggero, line 290: We should use the inbound context (just need to give it and the `*taprpc.GetInfoRequest` a parameter name). Here and below.

rpcserver.go
```diff
@@ -286,10 +286,29 @@ func (r *rpcServer) DebugLevel(ctx context.Context,
...
+	if err != nil {
+		// Handle error appropriately
```
guggero, line 292: This comment doesn't give any additional info and should just be removed, here and below.

rpcserver.go
```diff
@@ -308,6 +308,7 @@ func (r *rpcServer) GetInfo(context.Context,
 		NodeAlias:         info.Alias,
 		BlockHeight:       uint32(blockHeight),
 		BlockHash:         blockHash.String(),
+		SyncStatus:        info.SyncedToChain,
```
guggero, line 311: This commit should be squashed with the previous one.

guggero (commented):
Thanks for taking this over! Needs a couple of small fixes but is otherwise almost ready to go!

dstadulis:
could include in a subsequent v0.3 RC if we get the requested changes

rpcserver.go
```diff
@@ -283,13 +283,30 @@ func (r *rpcServer) DebugLevel(ctx context.Context,
...
-		LndVersion: r.cfg.Lnd.Version.Version,
-		Network:    r.cfg.ChainParams.Name,
+		Version:           Version(),
+		LndVersion:        r.cfg.Lnd.Version.Version,
+		Network:           r.cfg.ChainParams.Name,
+		LndIdentityPubkey: r.cfg.Lnd.NodePubkey.String(),
+		NodeAlias:         info.Alias,
+		BlockHeight:       uint32(blockHeight),
+		BlockHash:         blockHash.String(),
+		SyncStatus:        info.SyncedToChain,
```
jamaljsr, line 309: I think it would be better to change `SyncStatus` to `SyncedToChain`. As a user, seeing `"sync_status": true` in the output is pretty vague. It's not clear whether this is related to the chain or graph or something new in `tapd` regarding universes.

guggero:
Included a rebased version of the commit in this PR: https://github.com/lightninglabs/taproot-assets/pull/589 (keeping author attribution intact).

PR #553, opened 2023-10-06
Improve clarity on asset mint and universe stats RPC responses

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/522.

~Depends on https://github.com/lightninglabs/taproot-assets/pull/556.~

Addresses the following items:
1. Show group anchor when minting grouped assets
2. Serialize sprouts of finalized batch (this also discovered a bug with the `TapCommitment` for grouped assets which we fix)
3. Make distinction between group anchor and single asset more clear in `QueryAssetStats` by putting them into different fields (see below)
4. Show total supply of asset group both in overall universe stats and individual stats.

Example output of `tapcli universe stats`:

```
{
    "num_total_assets": "707037",
    "num_total_groups": "4",
    "num_total_syncs": "200",
    "num_total_proofs": "204"
}
```

Example output of `tapcli universe stats assets --sort_by asset_name`:

```
{
    "asset_stats": [
        {
            "group_key": "02e46677e636646628ae8d5850faafa137dbad2d7fee2079193c700430c50c818f",
            "group_supply": "705500",
            "group_anchor": {
                "asset_id": "7080c11a75ee6f286e1ed2c043ada20ad566c76b463b183111c7b7f8ce49a3b7",
                "genesis_point": "db198dea904ba078cfbf7738312351488b77b790ea2630d1caa5843793c36ea0:1",
                "total_supply": "500",
                "asset_name": "OLIBUX1",
                "asset_type": "NORMAL",
                "genesis_height": 192,
                "genesis_timestamp": "1696609501"
            },
            "asset": null,
            "total_syncs": "0",
            "total_proofs": "3"
        },
        {
            "group_key": "",
            "group_supply": "0",
            "group_anchor": null,
            "asset": {
                "asset_id": "7f3f37d2afbccb90abf2d56036aa64adce432c65d686e705cf6a3401d8321ca4",
                "genesis_point": "e912509957ab4a5e5e4a4c21fd33eda85601a71e58c21a14e449626d7ea12d1d:1",
                "total_supply": "1337",
                "asset_name": "OLIBUX1337",
                "asset_type": "NORMAL",
                "genesis_height": 193,
                "genesis_timestamp": "1696609572"
            },
            "total_syncs": "0",
            "total_proofs": "1"
        },
        {
            "group_key": "03e20eee073814638b134062948c0d37b93e2d45ca18066ee86440d305cec1f711",
            "group_supply": "100",
            "group_anchor": {
[...]

PR #556, opened 2023-10-09
tapfreighter: don't send proofs for tombstones, remove tombstones and burns from commitments

guggero:
Don't attempt to send proofs for tombstone outputs.

This wasn't discovered before because we don't have the proof courier enabled by default in all our integration tests.

EDIT: I pulled in the bug fix commits from https://github.com/lightninglabs/taproot-assets/pull/553 since it makes more sense to have them all bundled in the same PR.

So this PR now contains all fixes for bugs recently discovered and the unit or integration test to cover them.


tapfreighter/chain_porter.go
```diff
@@ -605,6 +605,18 @@ func (p *ChainPorter) transferReceiverProof(pkg *sendPackage) error {
 			return nil
 		}
 
+		unSpendable, err := out.ScriptKey.IsUnSpendable()
+		if err != nil {
+			return fmt.Errorf("error checking if script key is "+
+				"unspendable: %w", err)
+		}
+		if unSpendable {
```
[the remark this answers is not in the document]
guggero: Yeah, I can add that. I found this while writing and running the load test send test case. Also found another bug I'm currently investigating. So load testing for the win :sweat_smile:

commitment/tap.go
```diff
@@ -452,18 +452,14 @@ func (c *TapCommitment) Merge(other *TapCommitment) error {
 
 	// Otherwise, we'll need to merge the other asset commitments into
 	// this commitment.
-	for key, otherCommitment := range other.assetCommitments {
-		existingCommitment, ok := c.assetCommitments[key]
+	for key := range other.assetCommitments {
```
ffranr, line 455: `key` might not be the correct variable name after this change. I think it's an index now. Should we just use `i` instead?

ffranr (approved):
Good to see the variable renames :+1:

tapdb/asset_minting.go
```diff
@@ -543,7 +543,7 @@ func fetchAssetSprouts(ctx context.Context, q PendingAssetStore,
 
 	// For each sprout, we'll create a new asset commitment which will be a
 	// leaf at the top-level Taproot Asset commitment.
-	assetCommitments := make([]*commitment.AssetCommitment, len(dbSprout))
+	assetSprouts := make([]*asset.Asset, len(dbSprout))
```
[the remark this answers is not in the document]
guggero: Yeah, makes sense. Fixed.

commitment/tap.go
guggero: No, `key` is still correct IMO, since `assetCommitments` is a map, not a slice.

itest/multi_asset_group_test.go
```diff
@@ -190,14 +192,12 @@ func createMultiAssetGroup(anchor *mintrpc.MintAssetRequest,
...
-	nameModifier := "-tranche-"
 	groupSum := uint64(0)
-	for i := uint64(1); i < numAssets+1; i++ {
+	for i := uint64(1); i <= numAssets; i++ {
 		assetReq := CopyRequest(anchor)
 		assetReq.EnableEmission = false
 		assetReq.Asset.GroupAnchor = anchorName
-		reqName := anchorName + nameModifier + strconv.FormatUint(i, 10)
-		assetReq.Asset.Name = reqName
+		assetReq.Asset.Name = fmt.Sprintf("%s-trance-%d", anchorName, i)
```
jharveyb, line 200: super nit: tranche->trance

itest/multi_asset_group_test.go
```diff
@@ -270,3 +270,144 @@ func testMintMultiAssetGroupErrors(t *harnessTest) {
 		validAnchor.Asset.Amount
 	AssertBalanceByGroup(t.t, t.tapd, groupKeyHex, expectedGroupBalance)
 }
+
+// testMultiAssetGroupSend tests that we can randomly send assets from a group
+// of collectibles back and forth between two nodes.
```
jharveyb, line 275: IIUC this is only a set of one-way sends and not two-way?

jharveyb (approved):
Great set of fixes and readability improvements from older sections 🚀

Roasbeef (approved):
LGTM 🧨

PR #557, opened 2023-10-10
asset: update parsing to return error if unknown odd type encountered

Roasbeef:
In this commit, as a follow up to the recent even/odd type split, we add some helper functions to assert that when we decode an asset TLV stream, we bail out if we find any unknown types.

### DIFF
--- a/asset/records.go
+++ b/asset/records.go
@@ -3,11 +3,15 @@ package asset
 import (
 	"bytes"
 	"crypto/sha256"
+	"fmt"
+	"io"
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightninglabs/taproot-assets/mssmt"
 	"github.com/lightningnetwork/lnd/tlv"
+	"golang.org/x/exp/maps"
 )
 
 // LeafTlvType represents the different TLV types for Asset Leaf TLV records.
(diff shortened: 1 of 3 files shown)


asset/records.go
```diff
@@ -30,6 +34,71 @@ const (
 	LeafAssetID LeafTlvType = 11
 )
 
+// KnownAssetLeafTypes is a set of all known leaf types.
+var KnownAssetLeafTypes = fn.NewSet(
```
Roasbeef, line 38: Ideally we can set up some compile or `init` level thing to make sure that each time the set of types are updated, then this is as well.

asset/records.go
```diff
@@ -30,6 +34,71 @@ const (
...
+	// UnknownType is the type that was unknown.
+	UnknownType tlv.Type
+
+	// ValueBytes is the raw bytes of the value that was unknown.
+	ValueBytes []byte
+}
+
+// Erorr returns the error message for the ErrUnknownType.
+func (e ErrUnknownType) Error() string {
+	return fmt.Errorf("unknown type %d", e.UnknownType).Error()
```
guggero, line 56: nit: why not just `fmt.Sprintf()`?

guggero (approved):
Nice, LGTM :tada:

asset/records_test.go
```diff
@@ -0,0 +1,66 @@
...
+			knownTypes: fn.NewSet[tlv.Type](0, 2),
+			err: ErrUnknownType{
+				UnknownType: 5,
+				ValueBytes:  []byte{},
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+
```
jharveyb, line 58: linter fail on this extra newline.

dstadulis:
!lightninglabs-deploy mute 720h00m

dstadulis:
!lightninglabs-deploy mute 720h00m

guggero:
Replaced by https://github.com/lightninglabs/taproot-assets/pull/1037.

PR #558, opened 2023-10-10
loadtest: add multi send test

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/469.

Depends on https://github.com/lightninglabs/taproot-assets/pull/556.

This is the last piece to address our initial requirements for the load test.

cc @calvinrzachman:

I changed some things around and it's now possible to either specify a test case in the config (`test-case=xxx`, if you don't specify anything, all two tests will run) or in the command line (`-test.run="TestPerformance/send"`).
There are a couple of new test specific settings we can specify (and I renamed the `batch-size` config to `mint-test-batch-size`).
Here are the default values for these new config values:

```
mint-test-batch-size=200
send-test-num-sends=50
send-test-num-assets=1
send-test-send-type=1
```

### DIFF
--- a/itest/loadtest/load_test.go
+++ b/itest/loadtest/load_test.go
@@ -9,6 +9,18 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type testCase struct {
+	name string
+	fn   func(t *testing.T, ctx context.Context, cfg *Config)
+}
+
+var loadTestCases = []testCase{
+	{
+		name: "mint_batch_stress",
+		fn:   execMintBatchStressTest,
+	},
+}
+
 // TestPerformance executes the configured performance tests.
 func TestPerformance(t *testing.T) {
 	cfg, err := LoadConfig()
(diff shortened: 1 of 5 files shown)


itest/loadtest/load_test.go
```diff
@@ -18,23 +30,44 @@ func TestPerformance(t *testing.T) {
...
 
-	ctxt, cancel := context.WithTimeout(ctx, cfg.TestTimeout)
-	defer cancel()
+			continue
+		}
 
-	switch testName {
-	case "mint_batch_stress":
-		execMintBatchStressTest(t, ctxt, cfg)
+		success := t.Run(tc.name, func(tt *testing.T) {
```
Roasbeef, line 47: Wouldn't `t.Run` already handle the filter since we're using `test.run` now?

Roasbeef (approved):
LGTM 🔩

PR #559, opened 2023-10-10
rpc: unconstrain meta field for MintAssets

Roasbeef:
In this commit, we stop trying to strictly parse the meta field. Instead, users can set it to w/e they want, with the default being opaque. This lets users start to add structure to the meta field from day one. One example is a set meta with fields that dictate how the asset unit is to be displayed. This can be used to do things like leverage the current fix point format to add decimal places for display.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -7,6 +7,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


jharveyb:
The underlying `proof.MetaReveal.MetaType` field is `uint8`, but the matching field on the RPC request is `int32` - we should still clamp it to uint8 before conversion.

Roasbeef:
> but the matching field on the RPC request is int32 - we should still clamp it to uint8 before conversion.

Aren't we effectively doing that just with the cast? 

https://go.dev/play/p/sSqSvPQEDi-

guggero:
> Aren't we effectively doing that just with the cast?

Yes, but that would just truncate the value and lead to unexpected behavior. I think what @jharveyb meant was to return an error if the value was bigger than `math.MaxUint8`.

rpcserver.go
```diff
@@ -372,13 +372,8 @@ func (r *rpcServer) MintAsset(ctx context.Context,
...
 
 	if req.Asset.AssetMeta != nil {
-		metaType, err := unmarshalMetaType(req.Asset.AssetMeta.Type)
-		if err != nil {
-			return nil, err
-		}
-
 		seedling.Meta = &proof.MetaReveal{
-			Type: metaType,
+			Type: proof.MetaType(req.Asset.AssetMeta.Type),
```
guggero, line 376: Hmm... I could've sworn I already had this diff in a PR of mine... Or maybe I just forgot one instance of it?

guggero (approved):
LGTM pending the mentioned check for value overflow.

jharveyb:
> > Aren't we effectively doing that just with the cast?
> 
> Yes, but that would just truncate the value and lead to unexpected behavior. I think what @jharveyb meant was to return an error if the value was bigger than `math.MaxUint8`.

Or negative

guggero:
Ah, I guess we should at least make it `uint32` on the RPC level then.

jharveyb:
Yeah I could find where it was defined as `int32`, just that the concrete type is an enum, but I would expect that to be `uint32` and not `int32` anyways.

Also may want a test to make sure that the same strict meta type handling isn't also present elsewhere in the minting logic.

guggero (approved):

PR #560, opened 2023-10-10
tapdb: add unit test for universe leaf overflow

Roasbeef:
In this commit, we add a test for the universe leaf overflow. This tests that if we try to register a leaf that'll overflow the tree, then we'll just return an error and bail out of the transaction.


jharveyb (approved):
LGTM - could be good to add a fetch after trying to cause the overflow to check that we can successfully fetch the same leaf / data stored before that point.

PR #561, opened 2023-10-11
Universe events stats handle universe identity proof type

ffranr:
Depends on https://github.com/lightninglabs/taproot-assets/pull/540 . Do not merge this PR before merging https://github.com/lightninglabs/taproot-assets/pull/540 I've converted this PR to draft until all parent PRs have been merged.

This PR adds universe identity proof type support to universe events stats database SQL statements.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -3200,9 +3200,8 @@ func (r *rpcServer) QueryProof(ctx context.Context,
 		return nil, err
 	}
 
-	rpcsLog.Debugf("[QueryProof]: fetching proof at "+
-		"(universeID=%x, leafKey=%x)", universeID,
-		leafKey.UniverseKey())
+	rpcsLog.Debugf("[QueryProof]: fetching proof at (universeID=%v, "+
+		"leafKey=%x)", universeID, leafKey.UniverseKey())
 
 	// Keep a record of whether the proof type was specified by the client.
 	unspecifiedArgProofType :=
@@ -3237,14 +3236,14 @@ func (r *rpcServer) QueryProof(ctx context.Context,
 		)
 		if err != nil {
 			rpcsLog.Debugf("[QueryProof]: error querying for "+
-				"proof at (universeID=%x, leafKey=%x)",
+				"proof at (universeID=%v, leafKey=%x)",
 				universeID, leafKey.UniverseKey())
 			return nil, err
 		}
 	}
 	if err != nil {
 		rpcsLog.Debugf("[QueryProof]: error querying for proof at "+
-			"(universeID=%x, leafKey=%x)", universeID,
+			"(universeID=%v, leafKey=%x)", universeID,
 			leafKey.UniverseKey())
 		return nil, err
 	}
@@ -3253,9 +3252,8 @@ func (r *rpcServer) QueryProof(ctx context.Context,
 	// not be fully specified
 	proof := proofs[0]
 
-	rpcsLog.Debugf("[QueryProof]: found proof at "+
-		"(universeID=%x, leafKey=%x)", universeID,
-		leafKey.UniverseKey())
+	rpcsLog.Debugf("[QueryProof]: found proof at (universeID=%v, "+
+		"leafKey=%x)", universeID, leafKey.UniverseKey())
 
 	return r.marshalIssuanceProof(ctx, req, proof)
 }
@@ -3328,7 +3326,7 @@ func (r *rpcServer) InsertProof(ctx context.Context,
 	}
 
 	rpcsLog.Debugf("[InsertProof]: inserting proof at "+
-		"(universeID=%x, leafKey=%x)", universeID,
+		"(universeID=%v, leafKey=%x)", universeID,
 		leafKey.UniverseKey())
 
 	newUniverseState, err := r.cfg.BaseUniverse.RegisterIssuance(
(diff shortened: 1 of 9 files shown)


tapdb/universe_stats.go
```diff
@@ -141,6 +141,7 @@ func (u *UniverseStats) LogSyncEvent(ctx context.Context,
 			EventTimestamp: u.clock.Now().UTC().Unix(),
 			AssetID:        uniID.AssetID[:],
 			GroupKeyXOnly:  groupKeyXOnly,
+			ProofType:      uniID.ProofType.String(),
```
Roasbeef, line 144: Would've expected these test run on top of that branch to trigger unit test failures within it?

tapdb/sqlc/queries/universe.sql
```diff
@@ -113,14 +113,18 @@ SELECT * FROM universe_servers;
...
 WITH group_key_root_id AS (
     SELECT id
-    FROM universe_roots
+    FROM universe_roots roots
     WHERE group_key = @group_key_x_only
+      AND roots.proof_type = @proof_type
 ), asset_id_root_id AS (
     SELECT leaves.universe_root_id AS id
     FROM universe_leaves leaves
+    JOIN universe_roots roots
```
Roasbeef, line 122: Ah actually a bug fix? Or I guess we'd still have the same return value, this join is just slightly more correct.

tapdb/sqlc/queries/universe.sql
```diff
@@ -142,14 +142,18 @@ INSERT INTO universe_events (
...
+        AND roots.proof_type = @proof_type
 ), asset_id_root_id AS (
     SELECT leaves.universe_root_id AS id
     FROM universe_leaves leaves
-             JOIN genesis_info_view gen
-                  ON leaves.asset_genesis_id = gen.gen_asset_id
+    JOIN universe_roots roots
+        ON leaves.universe_root_id = roots.id
+    JOIN genesis_info_view gen
+        ON leaves.asset_genesis_id = gen.gen_asset_id
```
Roasbeef, line 154: Out of place, but I think we need to decide how we want to display things on the RPC level here. If we're going with the split approach, do we also make further RPC changes or just try to unify things? Some input from @jamaljsr likely useful here.

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -80,8 +80,7 @@ CREATE VIEW universe_stats AS
         COUNT(CASE WHEN u.event_type = 'SYNC' THEN 1 ELSE NULL END) AS total_asset_syncs,
         COUNT(CASE WHEN u.event_type = 'NEW_PROOF' THEN 1 ELSE NULL END) AS total_asset_proofs,
         roots.asset_id,
-        roots.group_key,
-        roots.namespace_root
```
Roasbeef, line 84: It wasn't used elsewhere?

Roasbeef (commented):
SGTM for the split approach. 

I think we still need to decide if we go with split or unified on the RPC layer.

tapdb/sqlc/migrations/000007_universe.up.sql
guggero: I guess it was just in the list for completeness, but the actual change here is that it's removed from the `GROUP BY` clause.

guggero (commented):
Looks good as far as I can tell. Will do some manual testing once the base PR has been merged.

tapdb/universe_stats.go
ffranr: Yes me too, but the stats must be aggregated correctly regardless, or we were missing coverage.

tapdb/sqlc/queries/universe.sql
ffranr: This additional `join` ensures that we select the asset ID from the row which has the correct proof type, but functionally it makes no difference.

tapdb/sqlc/queries/universe.sql
ffranr: @Roasbeef the changes in this PR don't lead to a split approach from @jamaljsr 's perspective. The query functions still aggregate across universe proof type.

ffranr:
@Roasbeef stats are still aggregated as before. This PR adds proof type resolution to how we store stats, but stats are still queried and aggregated as before.

guggero (commented):
Did a bunch of manual testing and found a bunch of small things that should be fixed.
Summarized them in this commit (feel free to just take over/split into fixup commits): https://github.com/lightninglabs/taproot-assets/tree/561-fixups

With those fixes applied, I think this is ready to go.

rpcserver.go
```diff
@@ -3200,9 +3200,8 @@ func (r *rpcServer) QueryProof(ctx context.Context,
 		return nil, err
 	}
 
-	rpcsLog.Debugf("[QueryProof]: fetching proof at "+
-		"(universeID=%x, leafKey=%x)", universeID,
-		leafKey.UniverseKey())
+	rpcsLog.Debugf("[QueryProof]: fetching proof at (universeID=%x, "+
```
Roasbeef, line 3203: This should also be `%v`.

Roasbeef (approved):
LGTM

guggero (approved):

PR #564, opened 2023-10-11
Use federation sync config in syncer

ffranr:
In this PR we modify the simple syncer and the federation envoy such that they use the federation sync configuration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/564)


ffranr:
In the fed sync config we differentiate between import and export syncing. This PR doesn't currently handle that nuance. We sync if both import and export is enabled.

universe/auto_syncer.go
```diff
@@ -183,7 +183,7 @@ func (f *FederationEnvoy) reportErr(err error) {
 // syncUniverseState attempts to sync Universe state with the target server.
 // If the sync is successful (even if no diff is generated), then a new sync
 // event will be logged.
-func (f *FederationEnvoy) syncUniverseState(ctx context.Context,
+func (f *FederationEnvoy) syncServerState(ctx context.Context,
```
Roasbeef, line 186: `godoc` needs to be updated.

universe/auto_syncer.go
```diff
@@ -184,14 +184,14 @@ func (f *FederationEnvoy) reportErr(err error) {
...
-	addr ServerAddr) error {
+	addr ServerAddr, syncConfigs SyncConfigs) error {
 
 	log.Infof("Syncing Universe state with server=%v", spew.Sdump(addr))
 
 	// Attempt to sync with the remote Universe server, if this errors then
 	// we'll bail out early as something wrong happened.
 	diff, err := f.cfg.UniverseSyncer.SyncUniverse(
-		ctx, addr, SyncIssuance,
+		ctx, addr, SyncFull, syncConfigs,
```
Roasbeef, line 194: This should pass through the sync type.

universe/auto_syncer.go
```diff
@@ -468,10 +468,60 @@ func (f *FederationEnvoy) SyncServers(serverAddrs []ServerAddr) error {
...
+			"config(s): %w", err)
+	}
+
+	syncConfigs := SyncConfigs{
+		GlobalSyncConfigs: globalConfigs,
+		UniSyncConfigs:    uniSyncConfigs,
+	}
+
+	syncServer := func(ctx context.Context, serverAddr ServerAddr) error {
+		return f.syncServerState(ctx, serverAddr, syncConfigs)
```
Roasbeef, line 485: We should have this just log the error instead.

universe/auto_syncer.go
```diff
@@ -468,10 +468,60 @@ func (f *FederationEnvoy) SyncServers(serverAddrs []ServerAddr) error {
...
+// IsSyncEnabled returns true if the given universe is configured to be
+// synchronized with the federation.
+//
+// TODO(ffranr): Add support for insert/export only sync enabled status.
+func (s *SyncConfigs) IsSyncEnabled(id Identifier) bool {
+	// Check for universe specific config. This takes precedence over the
+	// global config.
+	for _, cfg := range s.UniSyncConfigs {
+		if cfg.UniverseID == id {
+			return cfg.AllowSyncInsert && cfg.AllowSyncExport
```
Roasbeef, line 515: I think the export should be eliminated, as we only need to check that when we go to process RPC requests.

universe/auto_syncer.go
```diff
@@ -468,10 +468,60 @@ func (f *FederationEnvoy) SyncServers(serverAddrs []ServerAddr) error {
...
+		}
+	}
+
+	// Check for global config.
+	for _, cfg := range s.GlobalSyncConfigs {
+		if cfg.ProofType == id.ProofType {
+			return cfg.AllowSyncInsert && cfg.AllowSyncExport
```
Roasbeef, line 522: Same here as well.

universe/auto_syncer.go
Roasbeef: I think this is ok, as in this instance it isn't directly used for this main call path after all.

Roasbeef (changes requested):

Roasbeef:
- [ ] Examine export/insert in the rpc calls, reject read for export, opposite for import 
- [ ] Only examine insert for the syncer 
- [ ] Scratchup config, if only for itests -- `--universe.public`, allows insert+export global -- (`--universe.public_issuance`, `--universe.public_proofs`)

Roasbeef (changes requested):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/564)* status: 7 of 9 files reviewed, 6 unresolved discussions (waiting on @ffranr and @guggero)

___
*[`rpcserver.go` line 3272 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/564#-NgW0tcd71xL3hn3RZV0:-NgW0tcd71xL3hn3RZV1:btt0etq) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/34bf7e8b397c750de110224b0593848dd7c933cb/rpcserver.go#L3272)):*
> ```Go
> 	// TODO(roasbeef): query may return multiple proofs, if allow key to
> 	// not be fully specified
> 	proof := proofs[0]
> ```

if len(proofs) == 0 {
     return universe.ErrNoNUniverseProofFound
}]



_Code quote:_
```Go
	// TODO(roasbeef): query may return multiple proofs, if allow key to
	// not be fully specified
	proof := proofs[0]
```

Roasbeef (changes requested):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/564)* status: 8 of 9 files reviewed, 7 unresolved discussions (waiting on @ffranr and @guggero)

___
*[`universe/auto_syncer.go` line 490 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/564#-NgW1Uii7h4HaaVQLm_d:-NgW1Uii7h4HaaVQLm_e:btfntr7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/34bf7e8b397c750de110224b0593848dd7c933cb/universe/auto_syncer.go#L490)):*
> ```Go
> 
> 	// Obtain the general and universe specific federation sync configs.
> 	syncConfigs, err := f.QuerySyncConfigs(ctx)
> ```

Missing error check.

Issue #566, opened 2023-10-11
cmd/lncli: make new CLI commands for the universe federation config commanda 


Roasbeef:
New protos here: https://github.com/lightninglabs/taproot-assets/pull/563.

jharveyb:
From offline discussion:

-CLI args should take precedent over config from file or config stored in DB.
-Need some way to flag or resolve config conflicts directly (what if config file and DB differ on startup?)
-May want to expose only a subset of options via CLI

jharveyb:
Resolved with #568

PR #568, opened 2023-10-12
tapcli: expose universe sync config settings

jharveyb:
Fixes https://github.com/lightninglabs/taproot-assets/issues/566.

In draft as still needs some manual testing.

Would also like to add a universe federation config info command before merge.

Intended supported usage would be to set the insert or export bools independently (one or both at the same time), and only modify the config for a single scope at a time (global config or config for a specific asset).

### DIFF
--- a/cmd/tapcli/universe.go
+++ b/cmd/tapcli/universe.go
@@ -6,9 +6,9 @@ import (
 	"fmt"
 
 	tap "github.com/lightninglabs/taproot-assets"
+	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightninglabs/taproot-assets/proof"
 	"github.com/lightninglabs/taproot-assets/taprpc"
-	"github.com/lightninglabs/taproot-assets/taprpc/universerpc"
 	unirpc "github.com/lightninglabs/taproot-assets/taprpc/universerpc"
 	"github.com/lightninglabs/taproot-assets/universe"
 	"github.com/lightningnetwork/lnd/lncfg"
@@ -19,14 +19,14 @@ const (
 	proofTypeName = "proof_type"
 )
 
-func getUniverseClient(ctx *cli.Context) (universerpc.UniverseClient, func()) {
+func getUniverseClient(ctx *cli.Context) (unirpc.UniverseClient, func()) {
 	conn := getClientConn(ctx, false)
 
 	cleanUp := func() {
 		conn.Close()
 	}
 
-	return universerpc.NewUniverseClient(conn), cleanUp
+	return unirpc.NewUniverseClient(conn), cleanUp
 }
 
 // TODO(roasbeef): all should be able to connect to remote uni
@@ -74,12 +74,25 @@ var universeRootsCommand = cli.Command{
 	Action: universeRoots,
 }
 
-func parseUniverseID(ctx *cli.Context, mustParse bool) (*universerpc.ID, error) {
+func parseProofType(ctx *cli.Context) (*unirpc.ProofType, error) {
 	proofType, err := universe.ParseStrProofType(ctx.String(proofTypeName))
 	if err != nil {
 		return nil, err
 	}
 	rpcProofType, err := tap.MarshalUniProofType(proofType)
+	if err != nil {
+		return nil, err
+	}
+
+	if rpcProofType == unirpc.ProofType_PROOF_TYPE_UNSPECIFIED {
+		return nil, fmt.Errorf("invalid proof type")
+	}
+
+	return &rpcProofType, nil
+}
+
+func parseUniverseID(ctx *cli.Context, mustParse bool) (*unirpc.ID, error) {
+	rpcProofType, err := parseProofType(ctx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


cmd/tapcli/universe.go
```diff
@@ -8,7 +8,6 @@ import (
 	tap "github.com/lightninglabs/taproot-assets"
 	"github.com/lightninglabs/taproot-assets/proof"
 	"github.com/lightninglabs/taproot-assets/taprpc"
-	"github.com/lightninglabs/taproot-assets/taprpc/universerpc"
```
Roasbeef, line 11: Weird that linter didn't complain with both instances...

cmd/tapcli/universe.go
```diff
@@ -74,7 +73,11 @@ var universeRootsCommand = cli.Command{
 	Action: universeRoots,
 }
 
-func parseUniverseID(ctx *cli.Context, mustParse bool) (*universerpc.ID, error) {
+func parseProofType(ctx *cli.Context) (*unirpc.ProofType, error) {
+	if !ctx.IsSet(proofTypeName) {
```
Roasbeef, line 77: IIUC we set a default for issuance above, it always _should_ be set?

cmd/tapcli/universe.go
```diff
@@ -698,6 +715,218 @@ func universeFederationDel(ctx *cli.Context) error {
...
+	isGlobal bool) (*unirpc.GlobalFederationSyncConfig,
+	*unirpc.AssetFederationSyncConfig, error) {
+
+	var (
+		insertOpt *bool
+		exportOpt *bool
+		err       error
+	)
+
+	// The cli.BoolFlag works for variables that are false by default,
```
Roasbeef, line 765: I think we can use `BoolTFlag` here? If we want it to be true by default.

cmd/tapcli/universe.go
```diff
@@ -698,6 +715,218 @@ func universeFederationDel(ctx *cli.Context) error {
...
+		cli.StringFlag{
+			Name:  proofTypeName,
+			Usage: "the type of proof",
+		},
+		cli.StringFlag{
+			Name: proofInsertName,
+			Usage: "if true, remote Universes can push proofs to" +
+				"the local Universe",
+		},
+		cli.StringFlag{
```
Roasbeef, line 747: Why can't this be a `bool`? Or is that we want to actually allow it to not bet set (so you don't have to always set both)?

cmd/tapcli/universe.go
```diff
@@ -698,6 +715,218 @@ func universeFederationDel(ctx *cli.Context) error {
...
+	scope (all assets or specific assets).
+        `,
+	Subcommands: []cli.Command{
+		universeFederationGlobalConfig,
+		universeFederationLocalConfig,
+	},
+}
+
+var (
+	proofInsertName        = "insert"
```
Roasbeef, line 735: `allow_insert` and `allow_export`?

cmd/tapcli/universe.go
```diff
@@ -698,6 +715,218 @@ func universeFederationDel(ctx *cli.Context) error {
...
+	// The insert and export flags are not mutually exclusive.
+	if ctx.IsSet(proofInsertName) {
+		*insertOpt, err = isValidBool(ctx.String(proofInsertName))
+		if err != nil {
+			return nil, nil, fmt.Errorf("invalid proof insert flag")
+		}
+	}
+
+	if ctx.IsSet(proofExportName) {
+		*exportOpt, err = isValidBool(ctx.String(proofExportName))
```
Roasbeef, line 792: Unclear why we need special proof handling here.

Roasbeef (commented):
Looks pretty good, only question is re special handling of the bool fields.

cmd/tapcli/universe.go
jharveyb: I think we don't want it to have a default at all and have it be fully explicit.

Something like `tapcli universe federation config global --proof_type=transfer --insert true --export false`

Where `insert` and `export` without a bool fail.

cmd/tapcli/universe.go
jharveyb: Also that, I think it would make the most sense to be able to set either field or both with one command, vs one command per field.

cmd/tapcli/universe.go
jharveyb: Ah that came in after a rebase, will remove.

cmd/tapcli/universe.go
jharveyb: Yeah updated it to follow the abbrevation we use elsewhere in `cmd`.

cmd/tapcli/universe.go
jharveyb: Also for `ctx.BoolFlag`, actually trying to resolve the flag returns `false` if parsing fails, instead of a separate error.

cmd/tapcli/universe.go
Roasbeef: Ah sure I think that makes sense.

cmd/tapcli/universe.go
jharveyb: Not sure what you meant?

cmd/tapcli/universe.go
jharveyb: Moving to no default for these commands to match the other flags.

jharveyb:
Added the info command with a few flags to keep the output human-readable in the case of many configs present.

Tested manually on mainnet to display the default config correctly, successfully change settings individually, and reject universe sync for a specific asset.

Fixed an issue with how default configs are applied when we have some custom configs.

cmd/tapcli/universe.go
Roasbeef: Nvm, makes sense re decision to force a user to specify everything.

PR #570, opened 2023-10-12
make: enable monitoring build tag by default for release builds

guggero:
The official docker image and binary release build should have the Prometheus monitoring capability built in by default. We enable that by adding the `monitoring` tag to the list of release build tags.


Roasbeef (approved):
LGTM 🐑

PR #580, opened 2023-10-12
tapcfg: fix defaults for testnet universe courier

Roasbeef:
When we made testnet the default, the default config logic wasn't updated to properly implement the fallback logic. As a result, on testnet, the mainnet courier was still being used.


jharveyb:
Fixes work on testnet but hitting this issue for mainnet:

`[tapcli] could not load global options: unknown network: mainnet`

jharveyb (approved):

PR #583, opened 2023-10-13
tapdb: Federation DB fixes

jharveyb:
Fixes multiple issues with sync config load and store.

The first issue was the federation DB handle not returning the correct global config when the global config is modified for one proof type - this is resolved by loading the default config, and keying configs by proof type. Configs read from the DB then replace the default config with the matching proof type.

The second issue affected grouped assets, and led to group member asset IDs not being shown on the CLI.

The third issue was the UNIQUE constraint on universe sync configs not being enforced, since NULL values are distinct in UNIQUE constraints by default for SQLite and Postgres. This could lead to a bad configuration state with multiple entries for the same universe. This behavior is not configurable for SQLite, so extra columns were added that have a default value when one field of the universe ID is NULL. This should fix the DB constraint.

### DIFF
--- a/tapdb/universe_federation.go
+++ b/tapdb/universe_federation.go
@@ -263,10 +263,8 @@ func (u *UniverseFederationDB) UpsertFederationSyncConfig(
 		}
 
 		// Upsert universe specific sync configs.
-		for i := range uniSyncConfigs {
+		for _, config := range uniSyncConfigs {
 			var (
-				config = uniSyncConfigs[i]
-
 				uniID        = config.UniverseID
 				groupPubKey  []byte
 				assetIDBytes []byte
@@ -285,6 +283,7 @@ func (u *UniverseFederationDB) UpsertFederationSyncConfig(
 
 			err := db.UpsertFederationUniSyncConfig(
 				ctx, UpsertFedUniSyncConfigParams{
+					Namespace:       uniID.String(),
 					AssetID:         assetIDBytes,
 					GroupKey:        groupPubKey,
 					ProofType:       uniID.ProofType.String(),
@@ -379,11 +378,9 @@ func (u *UniverseFederationDB) QueryFederationSyncConfigs(
 			[]*universe.FedUniSyncConfig, len(uniDbConfigs),
 		)
 
-		for i := range uniDbConfigs {
-			conf := uniDbConfigs[i]
-
+		for i, config := range uniDbConfigs {
 			proofType, err := universe.ParseStrProofType(
-				conf.ProofType,
+				config.ProofType,
 			)
 			if err != nil {
 				return err
@@ -391,8 +388,8 @@ func (u *UniverseFederationDB) QueryFederationSyncConfigs(
 
 			// Construct group key public key from bytes.
 			var pubKey *btcec.PublicKey
-			if conf.GroupKey != nil {
-				pubKey, err = btcec.ParsePubKey(conf.GroupKey)
+			if config.GroupKey != nil {
+				pubKey, err = btcec.ParsePubKey(config.GroupKey)
 				if err != nil {
 					return fmt.Errorf("unable to parse "+
 						"group key: %v", err)
@@ -401,7 +398,7 @@ func (u *UniverseFederationDB) QueryFederationSyncConfigs(
 
 			// Construct asset ID from bytes.
 			var assetID asset.ID
-			copy(assetID[:], conf.AssetID)
+			copy(assetID[:], config.AssetID)
 
 			uniID := universe.Identifier{
 				AssetID:   assetID,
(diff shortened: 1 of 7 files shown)


tapdb/universe_federation.go
```diff
@@ -328,11 +333,15 @@ func (u *UniverseFederationDB) QueryFederationSyncConfigs(
 			return err
 
 		default:
-			// Parse global db sync configs.
-			globalConfigs = make(
-				[]*universe.FedGlobalSyncConfig,
-				len(globalDbConfigs),
-			)
+			// Start with the default global sync configs.
+			for i := range defaultGlobalSyncConfigs {
```
Roasbeef, line 337: Is this required given that there should only actually be two global configs in the DB? IIUC, we should always override with what's in the DB.

Roasbeef: Also loop scoping stuff isn't needed anymore. We build+test with that new flag.

tapdb/universe_federation.go
```diff
@@ -328,11 +333,15 @@ func (u *UniverseFederationDB) QueryFederationSyncConfigs(
...
+				config := defaultGlobalSyncConfigs[i]
+
+				globalConfigSet[config.ProofType] = config
```
Roasbeef, line 340: Ah ok, so it's for if a user sets one, and not the other. 

We should have a unit test here, should be at most 20 lines.

tapdb/multiverse.go
```diff
@@ -152,6 +152,15 @@ func (b *MultiverseStore) RootNodes(
 				groupedAssets map[asset.ID]uint64
 			)
 
+			// Parse universe proof type and populate the universe
+			// ID.
+			id.ProofType, err = universe.ParseStrProofType(
```
Roasbeef, line 157: Makes sense, lead to blank values for grouped assets for the map we added. 

Also needs a test.

tapdb/sqlc/migrations/000007_universe.up.sql
```diff
@@ -130,6 +130,14 @@ CREATE TABLE IF NOT EXISTS federation_uni_sync_config (
         (asset_id IS NULL AND group_key IS NOT NULL)
     ),
 
+    -- This field is similar to asset_id but is never NULL, and only used for
+    -- the unique constraint below.
+    populated_asset_id BLOB GENERATED ALWAYS AS (COALESCE(asset_id, 0)) STORED,
+
+    -- This field is similar to group_key but is never NULL, and only used for
+    -- the unique constraint below.
+    populated_group_key BLOB GENERATED ALWAYS AS (COALESCE(group_key, 0)) STORED,
```
Roasbeef, line 139: TIL about generated columns!

Roasbeef: IIUC, w/ sqlite `ALTER TABLE`, we can add columns, but _not_ drop prior `UNIQUE` constraints. 

Also the query needs to be updated here, since these new generated columns will be the ones that conflict.

Roasbeef (changes requested):

tapdb/universe_federation.go
jharveyb: I agree it's sort of overkill for only two types, but I figured if we plan to extended these types later on I'd rather write it this way and not have to rework it later.

tapdb/sqlc/migrations/000007_universe.up.sql
jharveyb: IIUC we don't actually need to drop the existing constraint, but add the new stricter constraint?

Good point re: query.

Roasbeef:
Needs a rebase. 

> We can also add a migration to DROP any contents in this table. Since no deployment should have a local Universe config set, including the Universe servers we run, this should not be an issue.

I think this could work. Drop the old, then add on new.

Roasbeef (approved):
LGTM 🛩️

Changes look good, just need to verify the upgrade path works out like we think it does (old config wiped, new one in place, everything works from there).

jharveyb:
Verified locally on mainnet setup with the original issue that old local configs were dropped and upsert now works via CLI.

tapdb/universe_federation.go
```diff
@@ -421,6 +418,13 @@ func (u *UniverseFederationDB) QueryFederationSyncConfigs(
 		return nil, nil, err
 	}
 
+	// TODO(jhb): Should we also sort by asset ID?
+	// Return config options in a semi-stable order.
+	sort.SliceStable(uniConfigs, func(i, j int) bool {
```
guggero, line 423: Why not just an `ORDER BY` clause? That's why we do SQL in the first place, isn't it?

guggero (commented):
Approach looks good! Should replace the TODO with some SQL magic, then this is good to go!

guggero (approved):
Nice :100:

tapdb/universe_federation.go
guggero: Looks like the order is still different on Postgres...

jharveyb: Fixed, wasn't backend-specific but removing the sort after calling `maps.Values()` for the global config.

Issue #576, opened 2023-10-12
mint: reject requests for collectibles with a `supply` value


jharveyb:
Collectible assets are always amount 1, so users trying to make a different-valued collectible probably want to make an asset group with multiple members instead. However, right now the CLI silently accepts `mint` commands with a non-one amount in this case.

Issue #573, opened 2023-10-12
[bug]: identical runtime_id on two different nodes running in Docker


jamaljsr:
### Background

I'm noticing an issue in Polar where sometimes the `runtime_id` value is the same on two different nodes. This prevents federating the two of them.

![paste](https://github.com/lightninglabs/taproot-assets/assets/1356600/6dff2236-8759-4893-855c-74db776f5fc0)

### Your environment

```sh
tapcli getinfo                                    # version of `tapd`, `lnd`, and network
{
    "version": "0.3.0-alpha.rc1 commit=v0.3.0-alpha.rc1",
    "lnd_version": "0.17.0-beta",
    "network": "regtest"
}

uname -mrsv                                          # operating system 
Linux 6.4.16-linuxkit #1 SMP PREEMPT Sat Sep 23 13:36:48 UTC 2023 aarch64

bitcoind --version || btcd --version # version of `btcd`, `bitcoind`, or other backend
Bitcoin Core version v25.0.0

Running nodes in Docker v24.0.6 via Polar v2.1.0-rc1
```

### Steps to reproduce

Note: you may need to repeat these steps a few times to reproduce the issue. In my testing, it's happened approx 25% of the time. 

1. In Polar v2.1.0-rc1, create a new network with 2 `lnd` nodes. Add 2 `tapd` nodes.
2. Start the network
3. Open the terminal for `alice-tap` and `bob-tap`
4. Run `tapcli universe info`
5. Check to see if the `runtime_id` is identical
6. If so, on `alice-tap`, run `tapcli universe federation add --universe_host bob-tap` to produce the error
    ```
    [tapcli] rpc error: code = Unknown desc = cannot add ourselves as a federation member
    ```

### Expected behavior

The `runtime_id` should always be unique.

### Actual behavior

The `runtime_id` is the same on two different nodes.

PR #589, opened 2023-10-16
Fix runtime ID, check amount when minting collectibles, rebase GetInfo RPC fixes commit

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/576.
Fixes https://github.com/lightninglabs/taproot-assets/issues/573.

Replaces https://github.com/lightninglabs/taproot-assets/pull/554.

Fixes multiple small issues.


jharveyb:
Would be good to have an itest for the runtime ID change; something like starting 100 `tapd` instances, pulling the runtime ID via `universe info`, and making sure that they are all unique.

tapcfg/server.go
```diff
@@ -275,7 +276,13 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 		SyncBatchSize:       defaultUniverseSyncBatchSize,
 	})
 
-	runtimeID := prand.Int63() // nolint:gosec
+	var runtimeIDBytes [8]byte
```
Roasbeef, line 279: Any reason to not increase to 32 bytes so we don't have to think about it?

Roasbeef (changes requested):

Roasbeef:
> Would be good to have an itest for the runtime ID change; something like starting 100 tapd instances, pulling the runtime ID via universe info, and making sure that they are all unique.


Unit test here (forces us to abstract it a bit more) would prob be good enough here. Or we increase increase to 32 bytes and call it a day.

tapcfg/server.go
jharveyb: Could do both but p sure root cause was the RNG.

cmd/tapcli/assets.go
```diff
@@ -178,16 +183,26 @@ func mintAsset(ctx *cli.Context) error {
 		}
 	}
 
+	assetType, err := parseAssetType(ctx)
+	if err != nil {
+		return err
+	}
+
+	amount := ctx.Uint64(assetSupplyName)
```
Roasbeef, line 191: What if we just set it to `1` for the user here? Can't be any other value anyway.

Roasbeef (changes requested):

tapcfg/server.go
guggero: > Any reason to not increase to 32 bytes so we don't have to think about it?

Yeah, just to make things non-breaking on the RPC interface.

I think 8 bytes should be waaaay enough for this. I'm pretty sure this was a combination of weak RNG+docker+macOS that lead to duplicates. I tried to reproduce this on my machine with a dummy docker setup and wasn't able to.

Using the better RNG should definitely make this almost impossible to hit.

Unfortunately neither a unit nor an integration test can help here, since as long as things are in the same process, we can never get a collision as even the weak RNG is a package-level variable. So if code from the same process request randomness, it will be different for each call as the calls are serialized. This can only be reproduced by spawning multiple processes within a very short time (which isn't super easy to do with our current setup).

I'm voting to just leave things the way they are and rely on the CPRNG being much better for this.

cmd/tapcli/assets.go
guggero: That's what we currently do and is what lead to confusion: The user entered "5" and then didn't know why just one collectible was minted.

jharveyb: Agree, this is one of the recurring reports - I think the only acceptable 'silent' behavior to have here would be to accept a mint request for a Collectible even if the supply flag is not set. But rejecting for amounts != 1 reduces surprise later on ("oh I guess I need to use an asset group to represent what i'd like to create")

guggero: I changed the logic in the latest push in that we only silently update the amount to 1 if the `--supply` flag wasn't specified at all for collectible assets.

Roasbeef (approved):
LGTM 🗞️

Issue #591, opened 2023-10-16
universe: implement topological sort for syncing from the universe transfers tree


Roasbeef:
Today when we client is configured in "full sync" mode (so issuance and transfer trees for any or all assets), we'll run into an issue (dependent on the key sorting) where we'll fail to insert a transfer proof into our local universe. 

When we go to validate the inputs for a transfer proof, we'll try to fetch all the inputs so we can get the amt+script for each of them: https://github.com/lightninglabs/taproot-assets/blob/e695071f24be98c845b898eff811d58c57d4ed5a/universe/base.go#L198-L212

The `getPrevAssetSnapshot` routine will then attempt to fetch the leaf proofs from the local multiverse:
https://github.com/lightninglabs/taproot-assets/blob/e695071f24be98c845b898eff811d58c57d4ed5a/universe/base.go#L436-L452

If we're validating the transfer of asset `X`, _before_ the issuance of asset `X`, then this will fail. 

To resolve this, when we're syncing items from a Universe, if we're syncing the transfer root, then we need to make sure we do a _topological sort_ of the leaves before we attempt to validate them. In order to stage the items during sync, a set to build up the set of relevant leaves. Then we'll validate them in chunks, based on the set of disjointed graph for an input cluster, using topological sort to validate them in order.

jharveyb:
Continuation of https://github.com/lightninglabs/taproot-assets/pull/592#discussion_r1361638607 :

I think we could extend the proof file format to include the # of proofs, and then also sort with that.

Once processing is sorted we could also deduplicate some proof verification (if I'm verifying a spend, and a spend of some outputs from the first spend, then these proof files would only differ by one proof).

dstadulis:
Fixed in #594 https://github.com/lightninglabs/taproot-assets/pull/594/commits/6cc83003a7989b016fc7795fd1a04f2a2d97759c

Issue #575, opened 2023-10-12
proof: enable early termination of the universe proof courier 


Roasbeef:
Today we can use the Universe tree and the gRPC/REST API as a source for issuance and transfer proofs. The main loop walks backwards from the transaction is received on chain, using the prev outpoints to look up the next leaf in the Universe trees: https://github.com/lightninglabs/taproot-assets/blob/a990fdf18c2ca13cbb825c9aa95866831ccfadf8/proof/courier.go#L1045-L1095

Each time we use the gRPC client to query for the proof. Instead, we should add a step right before that, that attempts to look in the local proof acrhive. This'll allow us to short circuit, as all assets with the same asset ID eventually have a common ancestor. Once we know about this ancestor, then we have the rest of the proof so we can exit early. 

## Steps To Completion 

- [ ] Pass in the proof archive into the `UniverseRpcCourier`. Use that local archive to query before making a network query to the universe.

Issue #571, opened 2023-10-12
load test: use universe RPC proof courier for send load test


guggero:
I wanted to update our load test case that sends assets back and forth to use a local universe RPC proof courier by adding
```
		ProofCourierAddr: fmt.Sprintf(
			"%s://%s:%d", proof.UniverseRpcCourierType,
			send.cfg.Host, send.cfg.Port,
		),
```
here: https://github.com/lightninglabs/taproot-assets/blob/ba9fa8ca77267f0f2c5ca8235e82baa938b76d0d/itest/loadtest/send_test.go#L66

For some reason that resulted in the transfers not being executed anymore, failing with the following errors on the sender's side:

```
2023-10-12 14:06:20.890 [ERR] RPCS: [/universerpc.Universe/InsertProof]: unable to fetch previous asset snapshot: unable to fetch previous proof: no universe proof found
2023-10-12 14:06:20.891 [ERR] FRTR: unable to transfer receiver proof: error delivering proof(s): failed to deliver proof via courier service: error inserting proof into universe courier service: rpc error: code = Unknown desc = unable to fetch previous asset snapshot: unable to fetch previous proof: no universe proof found
```

Didn't have any time to look into it in more detail, so I'm not sure if this is a problem of the send test case code or an actual problem with the RPC proof courier.

Some perhaps important context is that the send test sends individual collectibles out of a grouped set of 200 assets. So perhaps an `assetID` vs. `groupKey` problem?

PR #592, opened 2023-10-17
universe: sort incoming transfer proofs by block height

Roasbeef:
In this commit, we attempt a simple sorting routine to address: https://github.com/lightninglabs/taproot-assets/issues/591.

Without these, we'll try to validate transfer proofs out of order, which is akin verifying an orphan transaction in the Bitcoin mempool. Rather than do a full on topological sort, we instead just sort by block height. As we don't support spending unconfirmed change, this should be enough to serialize the contents of a block.

### DIFF
--- a/universe/syncer.go
+++ b/universe/syncer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 	"sync/atomic"
 
 	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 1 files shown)


universe/syncer.go
```diff
@@ -210,13 +211,21 @@ func (s *SimpleSyncer) syncRoot(ctx context.Context, remoteRoot BaseRoot,
 	)
 
 	// We use an error group to simply the error handling of a goroutine.
+	// This goroutine will handle reading in batches of new leaves to
+	// insert into the DB. We'll fee the output of the goroutines below
```
guggero, line 215: nit: s/fee/feed

universe/syncer.go
```diff
@@ -243,22 +252,42 @@ func (s *SimpleSyncer) syncRoot(ctx context.Context, remoteRoot BaseRoot,
...
+		})
 	if err != nil {
 		return err
 	}
 
+	// If this is a tranfer tree, then we'll collect all the items as we
+	// need to sort them to ensure we can validate them in dep order.
+	if !isIssuanceTree {
+		transferLeaves := fn.Collect(transferLeafProofs)
+		sort.Slice(transferLeaves, func(i, j int) bool {
```
guggero, line 283: I think this'll work for now. Though I wonder: Will we ever want to make it possible to have subsequent transfers within the same block? Assuming an interactive scheme where participants would accept proofs as being valid before they have the block header populated (e.g. in a pocket universe construction)?

guggero (approved):
LGTM :tada: 

An integration test would've been nice. But I'll add one when addressing https://github.com/lightninglabs/taproot-assets/issues/575 or https://github.com/lightninglabs/taproot-assets/issues/571 anyway, so not a blocker.

jharveyb (approved):
Looks good - added some other detail to #591.

universe/syncer.go
Roasbeef: Yeah I think so, then we would use something like an `scid` here, that takes into account: tx index, and output index as well.

dstadulis: Will this sort be run again if there's a block reorg?

Roasbeef:
Superseded by #594

PR #598, opened 2023-10-18
make+vm: increase vm test coverage, improve the coverage target

jharveyb:
Spinoff of #549.

Improves test coverage for the VM and ports Makefile improvements from LND that allow for measuring test coverage per-package and per-test-case.


Roasbeef (approved):
LGTM 🍌

Issue #599, opened 2023-10-18
[bug]: Unable to burn asset and at the same time I lose all units of that asset


kulisekf:
### Background

I tried to burn a certain amount of asset units (always < 100%). After executing the burn command output, this asset completely disappeared from the tapcli assets list, but the asset record in tapcli assets balances and tapcli universe roots are still fully populated.

### Your environment

tapcli getinfo   
```json
{
    "version":  "0.3.0-alpha.rc2 commit=",
    "lnd_version":  "0.16.3-beta",
    "network":  "testnet3"
}
```
uname -mrsv   
```
Linux 5.4.0-164-generic
```
bitcoind --version || btcd --version  
```
Bitcoin Core version v24.0.0
```
lsb_release -a
```
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.6 LTS
Release:        20.04
Codename:       focal
```

### Steps to reproduce

First I minted asset type normal. 
`tapcli assets mint --type normal --name fantasycoin --supply 100 --meta_bytes "fantastic money"`
`tapcli assets mint f`
tapcli assets list:
[assets_list_before_burn_output.txt](https://github.com/lightninglabs/taproot-assets/files/12999368/assets_list_before_burn_output.txt)

after mining several blocks is time for burn
`tapcli assets burn --asset_id {asset_id} --amount 50`
output:
```
Please confirm destructive action.
Asset ID: 23a7f87be15b96427286d2bf0c4960b7f3bff2c713792ad728818e5f93b82de7
Current available balance: 100
Amount to burn: 50
 Are you sure you want to irreversibly burn (destroy, remove from circulation) the specified amount of assets?
Please answer 'yes' or 'no' and press enter:
```
I entered `yes`
output:
[burn_command_output.txt](https://github.com/lightninglabs/taproot-assets/files/12999362/burn_command_output.txt)

### Expected behavior

Maximum amount of the asset will be reduced, I will lose the same amount along with it, but I will have access to the rest

### Actual behavior
log:
[log_output.txt](https://github.com/lightninglabs/taproot-assets/files/12999377/log_output.txt)

tapcli universe roots | grep -A 4 -B 9 fantasycoin:
[universe_roots_output.txt](https://github.com/lightninglabs/taproot-assets/files/12999387/universe_roots_output.txt)

tapcli assets transfers:
[assets_transfers_output.txt](https://github.com/lightninglabs/taproot-assets/files/12999389/assets_transfers_output.txt)

[...]

PR #600, opened 2023-10-18
tapfreighter: don't attempt to send burns with courier

guggero:
We forgot to exclude burn outputs when it comes to burns, which then resulted in the `invalid proof courier URI address: parse "": empty url` error, failing to confirm burns.
So the transfer was pending forever.

Fixes https://github.com/lightninglabs/taproot-assets/pull/599.


dstadulis (approved):
Thanks for the catch

PR #603, opened 2023-10-18
multi: add txid of batch when returning the MintingBatch

Roasbeef:
In this PR, we thread through the `txid` of a minting batch if available. We also add an assertion in the itests to make sure it's always populated.


taprpc/mintrpc/mint.proto
```diff
@@ -102,11 +102,17 @@ message MintingBatch {
     */
     bytes batch_key = 1;
 
-    // The assets that are part of the batch.
-    repeated MintAsset assets = 2;
+    /*
+    The transaction ID of the batch. Only populated if the batch has been
+    committed.
+    */
+    string batch_txid = 2;
```
guggero, line 109: Do we care about breaking gRPC backward compatibility at this point or not yet?

guggero (approved):
LGTM :tada: 

Just a question about we want to preserve gRPC backward compatibility from v0.3.0 onward or not yet.

taprpc/mintrpc/mint.proto
Roasbeef: I think not necessarily yet? Just felt aesthetically better to group the singular fields, then have the repeated fields at the end.

Issue #118, opened 2022-09-13
Send: improve PSBT funding by using new lnd RPC


jharveyb:
Right now the send state machine computes TX outputs and constructs a PSBT; we need the connected LND instance to add an input to cover TX fees, and we also need to add the input that carries the input TaroCommitment.

Limitations & solutions:

- Right now the FundPsbt() call will always BIP69 sort the outputs, which would break existing split commitments. However, we have enough information to restore the original output layout.
- FundPsbt() will return an error if any inputs are attached that are not owned by the wallet, so we must use FundPsbt() before attaching our asset-bearing input.
- Without the asset-bearing inputs attached, FundPsbt() will underpay on fees. We can work around this by computing the fee needed to cover the asset-bearing inputs, increase the change output value by this amount, and later decrement the change output value.

Given that, we have some options for implementation:

1. Pass the PSBT as-is to the wallet via FundPsbt(), attach our input and key derivation info to the funded PSBT, reshuffle outputs, have the wallet sign via SignPsbt(), finalize with FinalizePsbt(). This would result in underpaying on fees.
2. Same as 1, but compute the fee impact of our asset-bearing input and increment the dummy output values before passing the PSBT to FundPsbt().
3. Update the FundPsbt() walletrpc call to make BIP69 compliance optional, so we don't have to reshuffle outputs in option 1.
4. Perform coin selection on the tarod side via ListUnspent(), attach the asset-bearing inputs, then use SignPsbt() and FinalizePsbt().

The initial implementation is using option 1, but it may be worth revising the FundPsbt() RPC call in the future to simplify this flow.

nope78787:
Would this allow `tarod` to operate in "watch-only" mode? 

If would be nice to hold the keys to Taro assets outside the daemon. Perhaps having multiple keys which can produce schnorr signatures using s threshold scheme.

Roasbeef:
@nope78787 so this is the case already: `tapd` doesn't hold any private keys itself, instead the wallet abstraction handles that. The default implementation today is one backed by lnd's wallet RPC service.

Roasbeef:
I think let's generalize this issue to supporting fully external PSBT inputs from RPC? So this would let a caller provide a PSBT for a custom spend, instead of `tapd` making the entire template as is.

dstadulis:
>fully external PSBT inputs from RPC? So this would let a caller provide a PSBT for a custom spend, instead of tapd making the entire template as is.

[...]

PR #605, opened 2023-10-19
multi: anchor fee test coverage

### LINKED ISSUE #671: [feature]: further test coverage for manual fee rates and fee rate estimation
Continuation of work from https://github.com/lightninglabs/taproot-assets/pull/595 , being implemented in https://github.com/lightninglabs/taproot-assets/pull/605 .

- [ ] change units for manual fees to sat/vB to better match LND etc.
- [ ] test negative cases for manual fees below the minimum fee, exceeding wallet funds
- [ ] negative cases for fee estimation exceeding wallet funds or excessively low estimate
- [ ] increase logging of fees used and TX weights
- [ ] investigate mis-adjustment of final absolute fees for anchor TXs for minting and transfers

jharveyb:
Fixes #671 

Verifies fixes for various issues with fee estimation.

There are also some issues with cleaning up caretaker state if the caretaker returns an error, which is not recoverable from inside the caretaker.

Batch finalization could be manually reattempted by the user.

Also added extra logging of fee rates and transaction weights.

Should also change units exposed to the user.

> In line with our other software, this flag could be called `--sat_per_vbyte` and be measured in sat/vB. We use the term `fee_rate` a bit differently in LND.

### DIFF
--- a/cmd/tapcli/assets.go
+++ b/cmd/tapcli/assets.go
@@ -10,6 +10,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/tapcfg"
 	"github.com/lightninglabs/taproot-assets/taprpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/urfave/cli"
 )
 
@@ -50,7 +51,7 @@ var (
 	groupByGroupName             = "by_group"
 	assetIDName                  = "asset_id"
 	shortResponseName            = "short"
-	feeRateName                  = "fee_rate"
+	feeRateName                  = "sat_per_vbyte"
 	assetAmountName              = "amount"
 	burnOverrideConfirmationName = "override_confirmation_destroy_assets"
 )
@@ -261,7 +262,7 @@ var finalizeBatchCommand = cli.Command{
 		},
 		cli.Uint64Flag{
 			Name: feeRateName,
-			Usage: "if set, the fee rate in sat/kw to use for " +
+			Usage: "if set, the fee rate in sat/vB to use for " +
 				"the minting transaction",
 		},
 	},
@@ -270,9 +271,18 @@ var finalizeBatchCommand = cli.Command{
 
 func parseFeeRate(ctx *cli.Context) (uint32, error) {
 	if ctx.IsSet(feeRateName) {
-		feeRate := ctx.Uint64(feeRateName)
-		if feeRate > math.MaxUint32 {
+		userFeeRate := ctx.Uint64(feeRateName)
+		if userFeeRate > math.MaxUint32 {
 			return 0, fmt.Errorf("fee rate exceeds 2^32")
+		}
+
+		// Convert from sat/vB to sat/kw. Round up to the fee floor if
+		// the specified feerate is too low.
+		feeRate := chainfee.SatPerKVByte(userFeeRate * 1000).
+			FeePerKWeight()
+
+		if feeRate < chainfee.FeePerKwFloor {
+			feeRate = chainfee.FeePerKwFloor
 		}
 
 		return uint32(feeRate), nil
@@ -531,7 +541,7 @@ var sendAssetsCommand = cli.Command{
 		},
 		cli.Uint64Flag{
 			Name: feeRateName,
-			Usage: "if set, the fee rate in sat/kw to use for " +
+			Usage: "if set, the fee rate in sat/vB to use for " +
 				"the anchor transaction",
 		},
 		// TODO(roasbeef): add arg for file name to write sender proof
--- a/itest/assertions.go
+++ b/itest/assertions.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/rpcclient"
 	"github.com/btcsuite/btcd/wire"
@@ -24,6 +25,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/universe"
 	"github.com/lightningnetwork/lnd/lnrpc/chainrpc"
 	"github.com/lightningnetwork/lnd/lntest/wait"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/exp/maps"
 )
@@ -217,6 +219,75 @@ func AssertTxInBlock(t *testing.T, block *wire.MsgBlock,
 	require.Fail(t, "tx was not included in block")
 
 	return nil
+}
+
+// AssertTransferFeeRate checks that fee paid for the TX anchoring an asset
+// transfer is close to the expected fee for that TX, at a given fee rate.
+func AssertTransferFeeRate(t *testing.T, minerClient *rpcclient.Client,
+	transferResp *taprpc.SendAssetResponse, inputAmt int64,
+	feeRate chainfee.SatPerKWeight, roundFee bool) {
+
+	txid, err := chainhash.NewHash(transferResp.Transfer.AnchorTxHash)
+	require.NoError(t, err)
+
+	AssertFeeRate(t, minerClient, inputAmt, txid, feeRate, roundFee)
+}
+
+// AssertFeeRate checks that the fee paid for a given TX is close to the
+// expected fee for the same TX, at a given fee rate.
+func AssertFeeRate(t *testing.T, minerClient *rpcclient.Client, inputAmt int64,
+	txid *chainhash.Hash, feeRate chainfee.SatPerKWeight, roundFee bool) {
+
+	var (
+		outputValue                 float64
+		expectedFee, maxOverpayment btcutil.Amount
+		maxVsizeDifference          = int64(2)
+	)
+
+	verboseTx, err := minerClient.GetRawTransactionVerbose(txid)
+	require.NoError(t, err)
+
+	vsize := verboseTx.Vsize
+	for _, vout := range verboseTx.Vout {
+		outputValue += vout.Value
+	}
+
+	t.Logf("TX vsize of %d bytes", vsize)
+
+	btcOutputValue, err := btcutil.NewAmount(outputValue)
+	require.NoError(t, err)
+
+	actualFee := inputAmt - int64(btcOutputValue)
+
+	switch {
+	case roundFee:
+		// Replicate the rounding performed when calling `FundPsbt`.
+		feeSatPerVbyte := uint64(feeRate.FeePerKVByte()) / 1000
+		roundedFeeRate := chainfee.SatPerKVByte(
+			feeSatPerVbyte * 1000,
+		).FeePerKWeight()
+
+		expectedFee = roundedFeeRate.FeePerKVByte().
+			FeeForVSize(int64(vsize))
+		maxOverpayment = roundedFeeRate.FeePerKVByte().
+			FeeForVSize(maxVsizeDifference)
+
+	default:
+		expectedFee = feeRate.FeePerKVByte().
+			FeeForVSize(int64(vsize))
+		maxOverpayment = feeRate.FeePerKVByte().
+			FeeForVSize(maxVsizeDifference)
+	}
+
+	// The actual fee may be higher than the expected fee after
+	// confirmation, as the freighter makes a worst-case estimate of the TX
+	// vsize. The gap between these two fees should still be small.
+	require.GreaterOrEqual(t, actualFee, int64(expectedFee))
+
+	overpaidFee := actualFee - int64(expectedFee)
+	require.LessOrEqual(t, overpaidFee, int64(maxOverpayment))
+
+	t.Logf("Correct fee of %d sats", actualFee)
 }
 
 // WaitForBatchState polls until the planter has reached the desired state with
--- a/itest/assertions.go
+++ b/itest/assertions.go
@@ -640,16 +640,16 @@ func ConfirmAndAssertOutboundTransfer(t *testing.T,
 	expectedAmounts []uint64, currentTransferIdx,
 	numTransfers int) *wire.MsgBlock {
 
-	return ConfirmAndAssetOutboundTransferWithOutputs(
+	return ConfirmAndAssertOutboundTransferWithOutputs(
 		t, minerClient, sender, sendResp, assetID, expectedAmounts,
 		currentTransferIdx, numTransfers, 2,
 	)
 }
 
-// ConfirmAndAssetOutboundTransferWithOutputs makes sure the given outbound
+// ConfirmAndAssertOutboundTransferWithOutputs makes sure the given outbound
 // transfer has the correct state and number of outputs before confirming it and
 // then asserting the confirmed state with the node.
-func ConfirmAndAssetOutboundTransferWithOutputs(t *testing.T,
+func ConfirmAndAssertOutboundTransferWithOutputs(t *testing.T,
 	minerClient *rpcclient.Client, sender TapdClient,
 	sendResp *taprpc.SendAssetResponse, assetID []byte,
 	expectedAmounts []uint64, currentTransferIdx,
--- a/itest/fee_estimation_test.go
+++ b/itest/fee_estimation_test.go
@@ -1,0 +1,181 @@
+package itest
+
+import (
+	"context"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/taproot-assets/taprpc"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+	"github.com/stretchr/testify/require"
+)
+
+// testFeeEstimation tests that we're able to spend outputs of various script
+// types, and that the fee estimator and TX size estimator used during asset
+// transfers are accurate.
+func testFeeEstimation(t *harnessTest) {
+	var (
+		// Make a ladder of UTXO values so use order is deterministic.
+		anchorAmounts = []int64{10000, 9990, 9980, 9970}
+
+		// The default feerate in the itests is 12.5 sat/vB, but we
+		// define it here explicitly to use for assertions.
+		defaultFeeRate   = chainfee.SatPerKWeight(3125)
+		higherFeeRate    = defaultFeeRate * 2
+		excessiveFeeRate = defaultFeeRate * 8
+		lowFeeRate       = chainfee.SatPerKWeight(500)
+
+		// We will mint assets using the largest NP2WKH output, and then
+		// use all three output types for transfers.
+		initialUTXOs = []*UTXORequest{
+			{
+				Type:   lnrpc.AddressType_NESTED_PUBKEY_HASH,
+				Amount: anchorAmounts[0],
+			},
+			{
+				Type:   lnrpc.AddressType_NESTED_PUBKEY_HASH,
+				Amount: anchorAmounts[1],
+			},
+			{
+				Type:   lnrpc.AddressType_WITNESS_PUBKEY_HASH,
+				Amount: anchorAmounts[2],
+			},
+			{
+				Type:   lnrpc.AddressType_TAPROOT_PUBKEY,
+				Amount: anchorAmounts[3],
+			},
+		}
+	)
+
+	ctxb := context.Background()
+	ctxt, cancel := context.WithTimeout(ctxb, defaultWaitTimeout)
+	defer cancel()
+
+	// Set the initial state of the wallet of the first node. The wallet
+	// state will reset at the end of this test.
+	SetNodeUTXOs(t, t.lndHarness.Alice, btcutil.Amount(1), initialUTXOs)
+	defer ResetNodeWallet(t, t.lndHarness.Alice)
+
+	// Mint some assets with a NP2WPKH input, which will give us an anchor
+	// output to spend for a transfer.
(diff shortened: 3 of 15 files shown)


jharveyb:
Will split out the caretaker shutdown issue discovered here into a separate PR. I think we can move testing for that fix into the existing unit tests for the planter, where we trigger the error with a different mock wallet anchor that always fails.

jharveyb:
Split caretaker fixes into #693 , the fee estimation itest here depends on fixes added there.

jharveyb:
Acceptance Criteria:

- [ ] new weight and fee estimator that reuses code from `btcwallet`
- [ ] refactor minter and freighter to both use the same weight & fee estimator
- [ ] itest that exercises the new manual feerate sanity-checking logic in the relevant RPC calls
- [ ] update RPC calls to use sat/vB
- [ ] itest that exercises errors for minter and freighter when the backing wallet is underfunded
- [ ] itest that exercises correct minting and transfer when backing wallet only has old (P2PKH) input types

dstadulis:
There's a round issue that has caused the tests to fail.  @jharveyb has identified root cause and addressing it.

jharveyb:
Specifically this unit conversion from sat/kw to sat/vB when constructing the `FundPsbt` call:

https://github.com/lightninglabs/taproot-assets/blob/564795a0e4c1d1034493b498f58a552920f2e39c/wallet_anchor.go#L61

Where e.x. a feerate of 3125 sat/kw or 12.5 sat/vB is rounded down to 12 sat/vB.

jharveyb:
Pushed up my latest state; discovered a separate issue about dust change handling in the freighter.

jharveyb:
Made tracking issue for dusty change: #738. Going to defer that for a separate PR.

jharveyb:
One Q I have is if we should actually accept feerates as `float` / float32 instead of uints now - with the change to sat/vB as the unit, as user cannot manually specify 1 sat/vB as a feerate, nor a rate 'close' to it like 1.5 sat/vB. This wasn't as issue before with sat/kw.

jharveyb:
From offline discussion: the standard for fee unit handling is to use sat/kW on the RPC interface, but expose sat/vB as the unit on the CLI. We then convert from sat/vB to sat/kW before making our RPC call.

tapgarden/caretaker.go
```diff
@@ -411,26 +403,28 @@ func (b *BatchCaretaker) fundGenesisPsbt(ctx context.Context) (*FundedPsbt, erro
...
-	// of a fee rate estimate.
 	var feeRate chainfee.SatPerKWeight
 	switch {
+	// If a fee rate was manually assigned for this batch, use that instead
+	// of a fee rate estimate.
 	case b.cfg.BatchFeeRate != nil:
 		feeRate = *b.cfg.BatchFeeRate
-		log.Infof("BatchCaretaker(%x): using manual fee rate",
-			b.batchKey[:])
+		log.Infof("BatchCaretaker(%x): using manual fee rate: %s, %d "+
```
guggero, line 412: I think we should do `feeRate.FeePerKVByte() / 1000` instead of `feeRate.FeePerKVByte().FeeForVSize(1)` which is a bit confusing IMO.

tapfreighter/wallet.go
```diff
@@ -1654,61 +1654,25 @@ func addAnchorPsbtInputs(btcPkt *psbt.Packet, vPkt *tappsbt.VPacket,
...
-			)
-		default:
-			return fmt.Errorf("unknown pkScript: %x",
-				inputPkScript)
+	// total vsize of the transaction and the necessary fee at the desired
+	// fee rate.
+	inputScripts := make([][]byte, 0, len(btcPkt.Inputs))
+	for inputIdx, input := range btcPkt.Inputs {
+		if input.WitnessUtxo.PkScript == nil {
+			return fmt.Errorf("input %d on psbt missing"+
```
guggero, line 1662: nit: missing space in error message

tapfreighter/wallet.go
```diff
@@ -1654,61 +1654,25 @@ func addAnchorPsbtInputs(btcPkt *psbt.Packet, vPkt *tappsbt.VPacket,
...
-				txscript.SigHashDefault,
-			)
-		default:
-			return fmt.Errorf("unknown pkScript: %x",
-				inputPkScript)
+	// total vsize of the transaction and the necessary fee at the desired
+	// fee rate.
+	inputScripts := make([][]byte, 0, len(btcPkt.Inputs))
+	for inputIdx, input := range btcPkt.Inputs {
+		if input.WitnessUtxo.PkScript == nil {
```
guggero, line 1661: Should check `input.WitnessUtxo` being `nil` first, as that's not mandatory.

tapfreighter/wallet.go
```diff
@@ -1654,61 +1654,25 @@ func addAnchorPsbtInputs(btcPkt *psbt.Packet, vPkt *tappsbt.VPacket,
...
-	// With this, we can now calculate the total fee we need to pay. We'll
-	// also make sure to round up the required fee to the floor.
-	totalWeight := int64(weightEstimator.Weight())
-	requiredFee := feeRate.FeeForWeight(totalWeight)
+	estimatedSize, requiredFee := tapscript.EstimateFee(
+		inputScripts, btcPkt.UnsignedTx.TxOut, feeRate,
+	)
+	log.Infof("Estimated TX vsize: %d, fee %d", estimatedSize,
+		requiredFee)
+	log.Infof("TX required fee before change adjustment: %d at feerate "+
```
guggero, line 1674: Same here re `FeeForVSize(1)`.

itest/assertions.go
```diff
@@ -640,16 +640,16 @@ func ConfirmAndAssertOutboundTransfer(t *testing.T,
...
 		currentTransferIdx, numTransfers, 2,
 	)
 }
 
-// ConfirmAndAssetOutboundTransferWithOutputs makes sure the given outbound
+// ConfirmAndAssertOutboundTransferWithOutputs makes sure the given outbound
 // transfer has the correct state and number of outputs before confirming it and
 // then asserting the confirmed state with the node.
-func ConfirmAndAssetOutboundTransferWithOutputs(t *testing.T,
+func ConfirmAndAssertOutboundTransferWithOutputs(t *testing.T,
```
guggero, line 723: :sweat_smile:

itest/fee_estimation_test.go
```diff
@@ -0,0 +1,182 @@
...
+		anchorAmounts = []int64{10000, 9990, 9980, 9970}
+
+		// The default feerate in the itests is 12.5 sat/vB, but we
+		// define it here explicitly to use for assertions.
+		defaultFeeRate   = chainfee.SatPerKWeight(3125)
+		higherFeeRate    = defaultFeeRate * 2
+		excessiveFeeRate = defaultFeeRate * 8
+		lowFeeRate       = chainfee.SatPerKWeight(500)
+
+		// We will mint assets using the larget NP2WKH output, and then
```
guggero, line 29: nit: s/larget/largest/?

itest/utils.go
```diff
@@ -148,6 +152,73 @@ func MineBlocks(t *testing.T, client *rpcclient.Client,
...
 }
 
+type UTXORequest struct {
+	Type   lnrpc.AddressType
+	Amount int64
+}
+
+// SetAnchorUTXOs sets the wallet state for the sending node to a set of UTXOs
+// of a specific type and value.
+func SetNodeUTXOs(t *harnessTest, wallet *node.HarnessNode,
```
guggero, line 162: nit: Godoc name mismatch.

itest/utils.go
```diff
@@ -148,6 +152,73 @@ func MineBlocks(t *testing.T, client *rpcclient.Client,
...
+		return &wire.TxOut{
+			PkScript: addrScript,
+			Value:    req.Amount,
+		}
+	}
+
+	aliceOutputs := fn.Map(reqs, makeOutputs)
+
+	_ = t.lndHarness.Miner.SendOutputsWithoutChange(aliceOutputs, feeRate)
+	t.lndHarness.MineBlocksAndAssertNumTxes(6, 1)
```
guggero, line 200: Do we need to mine 6 blocks? 1 should be enough for most RPCs that require a min conf of 1.

rpcserver.go
```diff
@@ -491,26 +491,30 @@ func (r *rpcServer) MintAsset(ctx context.Context,
...
 }
 
-// checkFeeRateSanity ensures that the provided fee rate is above the same
-// minimum fee used as a floor in the fee estimator.
+// checkFeeRateSanity ensures that the provided fee rate, in sat/vB, is above
+// the same minimum fee used as a floor in the fee estimator.
 func checkFeeRateSanity(rpcFeeRate uint32) (*chainfee.SatPerKWeight, error) {
-	var feeRate *chainfee.SatPerKWeight
-	switch {
+	feeFloor := chainfee.FeePerKwFloor.FeePerKVByte()
```
guggero, line 497: I think we should calculate in sat/kw only in here, so we should rather convert the manual fee rate below to sat/kw.

rpcserver.go
```diff
@@ -491,26 +491,30 @@ func (r *rpcServer) MintAsset(ctx context.Context,
...
+	// Cast the fee rate from sat/vB to sat/kb below comparing against our
+	// fee floor.
+	manualFeeRate := chainfee.SatPerKVByte(rpcFeeRate * 1000)
+	switch {
 	// A manual fee was set but is below a reasonable floor.
-	case rpcFeeRate < uint32(chainfee.FeePerKwFloor):
-		return nil, fmt.Errorf("manual fee rate %d below floor of %d",
-			rpcFeeRate, uint32(chainfee.FeePerKwFloor))
+	case manualFeeRate < feeFloor:
+		return nil, fmt.Errorf("manual fee rate %s below floor of %s",
```
guggero, line 510: I think it's okay to round up to the floor rate here instead of returning an error.

taprpc/taprootassets.proto
```diff
@@ -915,7 +915,7 @@ message AddrReceivesResponse {
 message SendAssetRequest {
     repeated string tap_addrs = 1;
 
-    // The optional fee rate to use for the minting transaction, in sat/kw.
+    // The optional fee rate to use for the minting transaction, in sat/vB.
```
guggero, line 918: This should stay `sat/kw`! The conversion should only be done on the CLI side.

rpcserver.go
guggero: Or rather, the manual fee rate should be in sat/kw on the RPC, see comment below.

cmd/tapcli/assets.go
```diff
@@ -261,7 +261,7 @@ var finalizeBatchCommand = cli.Command{
 		},
 		cli.Uint64Flag{
 			Name: feeRateName,
-			Usage: "if set, the fee rate in sat/kw to use for " +
+			Usage: "if set, the fee rate in sat/vB to use for " +
```
guggero, line 264: This is correct, but the RPC should still be sat/kw. So we need to do the conversion within the CLI as a convenience to the user.

guggero (commented):
Very nice improvements and test additions!
Needs one change for the RPC to remain at sat/kw, the rest looks good :100:

itest/fee_estimation_test.go
```diff
@@ -0,0 +1,182 @@
...
+	AssertNonInteractiveRecvComplete(t.t, t.tapd, transferIdx)
+
+	sendInputAmt = anchorAmounts[2] + 1000
+	AssertTransferFeeRate(
+		t.t, t.lndHarness.Miner.Client, sendResp, sendInputAmt,
+		higherFeeRate, false,
+	)
+
+	// If we quadruple the fee rate, the freighter should fail during input
+	// input selection.
```
ffranr, line 135: no need for extra `input` in comment

rpcserver.go
ffranr: I'm not sure if I agree with you Oli. If the fee rate was set manually maybe the user strictly didn't want to pay a higher fee. I think we should add a flag to the relevant RPC calls if we want to automatically round up. Could be a future PR.

If this error does remain, then I think we should roll it an `error` type and then check for it in the itest.

itest/fee_estimation_test.go
```diff
@@ -0,0 +1,182 @@
...
+	})
+	require.ErrorContains(t.t, err, "insufficient funds available")
+
+	// The transfer should also be rejected if the manually-specified
+	// feerate fails the sanity check against the fee estimator's fee floor.
+	_, err = t.tapd.SendAsset(ctxt, &taprpc.SendAssetRequest{
+		TapAddrs: []string{addr3.Encoded},
+		FeeRate:  1,
+	})
+	require.ErrorContains(t.t, err, "below floor of")
```
ffranr, line 159: Consider adding an error type for "below floor" error?

Also, with this in mind, perhaps this itest commit should come after the rpc server commit (next commit).

itest/fee_estimation_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+	sendInputAmt = anchorAmounts[3] + 1000
+	AssertTransferFeeRate(
+		t.t, t.lndHarness.Miner.Client, sendResp, sendInputAmt,
+		lowFeeRate, false,
+	)
+
+	// Reset the state of the first node's wallet so that following tests
+	// are not affected.
+	ResetNodeWallet(t, t.lndHarness.Alice)
```
ffranr, line 181: Perhaps execute this as a `defer` at the point where these values are first modified within this test?

itest/utils.go
```diff
@@ -148,6 +152,73 @@ func MineBlocks(t *testing.T, client *rpcclient.Client,
 	return blocks
 }
 
+type UTXORequest struct {
+	Type   lnrpc.AddressType
+	Amount int64
+}
+
+// SetAnchorUTXOs sets the wallet state for the sending node to a set of UTXOs
```
ffranr, line 160: nit: comment mentions "sending node" but it could be generalised to any node.

ffranr (commented):
Great itest and clear changes. Nice work.

I can only think of small things that might need addressing. My second review should be quick.

jharveyb:
https://github.com/lightninglabs/taproot-assets/pull/605#discussion_r1431559617

(Can't reply to comment directly) - the rounding up here is a difference of 3 sat/kw or 12 sat/kvB; so extremely small, even for a hypothetical large transfer TX.

This is also what we do in Pool:

https://github.com/lightninglabs/pool/blob/59c7c2b40009503851de6da3b3f9b4d90b34e208/cmd/pool/account.go#L172

Without the rounding, the TX would get rejected later on by LND before broadcast anyways, so the alternative (no rounding) is not allowing the user to specify a fee below 2 sat/vB.

tapfreighter/wallet.go
```diff
@@ -1654,61 +1654,25 @@ func addAnchorPsbtInputs(btcPkt *psbt.Packet, vPkt *tappsbt.VPacket,
...
+		"%d sat/vB", requiredFee, feeRate.FeePerKVByte().FeeForVSize(1))
```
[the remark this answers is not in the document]
jharveyb: Changed, its effectivey just `/1000`.

tapfreighter/wallet.go
jharveyb: Agree, but is there a 'correct' case where `input.WitnessUtxo` is non-nil? IIUC any change outputs added should have this field populated, and earlier in this func we populate it for anchor inputs.

rpcserver.go
jharveyb: The rounding up is a difference of 3 sat/kw or 12 sat/kvB; so extremely small, even for a hypothetical large transfer TX.

This is also what we do in Pool:

https://github.com/lightninglabs/pool/blob/59c7c2b40009503851de6da3b3f9b4d90b34e208/cmd/pool/account.go#L172

Without the rounding, the TX would get rejected later on by LND before broadcast anyways, so the alternative (no rounding) is not allowing the user to specify a fee below 2 sat/vB.

I think the suggestion is to round on the CLI where the user cannot specify something above the fee floor but below 2 sat/vB, and then return an error on the actual RPC interface where a user can specify sat/kw directly.

itest/fee_estimation_test.go
jharveyb: Do we have a place to put error types that can be imported to both the RPC server and the CLI?

jharveyb:
Updated, would like to resolve the detail on possible inputs for transfers that don't have `WitnessUtxo` set.

rpcserver.go
guggero: The change from 250 to 253 as the floor fee rate was a security fix we did in `lnd`. So devs don't necessarily know about it, and we do them a favor by doing it that way, otherwise their TX could get stuck.
And as mentioned above, we do this in all our projects. Basically you should never publish a TX with just 250 sat/kWU.

tapfreighter/wallet.go
guggero: In `lnd` we always set `WitnessUtxo`, but it's not strictly mandatory according to the BIP. But since most hardware wallets also require it, we can assume that any software creating a PSBT would set it. So we can error out if it's `nil`.

itest/fee_estimation_test.go
guggero: I'd say in the RPC server directly, the CLI can just import the main package. Though since this error crosses the gRPC/TCP chasm, it won't be parseable as a type and we'll need to do string matching again. So not sure if it's worth the effort.

tapfreighter/wallet.go
jharveyb: Sounds good, I imagine this would also come up if the funding input is not-segwit like P2SH?

(Which wouldn't come from lnd 0.17+, but may be a consideration once we support external inputs).

guggero: `lnd` doesn't create any non-witness addresses, so it should never have non-witness UTXOs. Even nested-p2wkh (wrapped in p2wh) is segwit and therefore needs to have the witness UTXO field set.

jharveyb: Added a TODO + note on #118 where I expect this to be most relevant.

rpcserver.go
jharveyb: Confirmed manually that the CLI rounding works.

jharveyb:
Made final updates re: non-witness inputs, ready for final approval.

tapgarden/mock.go
```diff
@@ -101,17 +103,24 @@ func (m *MockWalletAnchor) FundPsbt(_ context.Context, packet *psbt.Packet,
 			Index: rand.Uint32(),
 		},
 	})
-	packet.Inputs = append(packet.Inputs, psbt.PInput{
+	// Use a P2TR input by default.
```
guggero, line 106: micor-nit: add newline before comment.

tapfreighter/wallet.go
```diff
@@ -1654,61 +1654,31 @@ func addAnchorPsbtInputs(btcPkt *psbt.Packet, vPkt *tappsbt.VPacket,
...
-				txscript.SigHashDefault,
-			)
-		default:
-			return fmt.Errorf("unknown pkScript: %x",
-				inputPkScript)
+	// total vsize of the transaction and the necessary fee at the desired
+	// fee rate.
+	inputScripts := make([][]byte, 0, len(btcPkt.Inputs))
+	for inputIdx, input := range btcPkt.Inputs {
+		// TODO(jhb): Support non-witness UTXOs as inputs by fetching
```
guggero, line 1661: I don't think we need this TODO here as I don't think we'll ever go through the effort of supporting non-witness inputs in our products. Or do you mean the `input.NonWitnessUtxo` field that might also be set?

tapfreighter/wallet.go
```diff
@@ -1654,61 +1654,31 @@ func addAnchorPsbtInputs(btcPkt *psbt.Packet, vPkt *tappsbt.VPacket,
...
+		// their PkScripts here.
+		if input.WitnessUtxo == nil {
+			return fmt.Errorf("input %d has no witness, not "+
```
guggero, line 1664: I would update the error message to read: "PSBT input %d doesn't specify witness UTXO, which is not supported".

tapfreighter/wallet.go
```diff
@@ -1654,61 +1654,31 @@ func addAnchorPsbtInputs(btcPkt *psbt.Packet, vPkt *tappsbt.VPacket,
...
-	}
-	for _, txOut := range btcPkt.UnsignedTx.TxOut {
-		outputAmt += txOut.Value
 
-		addrType, _, _, err := txscript.ExtractPkScriptAddrs(
-			txOut.PkScript, params,
-		)
-		if err != nil {
-			return err
+		if input.WitnessUtxo.PkScript == nil {
```
guggero, line 1668: Ah, and I guess we should do a `len() == 0` check here instead of nil.

cmd/tapcli/assets.go
```diff
@@ -261,7 +262,7 @@ var finalizeBatchCommand = cli.Command{
...
```
guggero, line 265: Sorry, I only saw this now: but we should update the flag name to `fee_rate_sat_per_vbyte` or just `sat_per_vbyte` as well. And then update all other places where we use the `parseFeeRate` as well (for example `finalizebatch` still mentions `sat/kw` but it would now be `sat/vByte`.

guggero (commented):
Nice, this is very close! CLI needs to be updated since the `parseFeeRate()` is used for other commands too. Other than that this looks good.

tapfreighter/wallet.go
jharveyb: IIUC a non-witness input would have `NonWitnessUtxo` set, and `WitnessUtxo` unset, so this method for gathering the scripts would fail. If we don't want to support non-witness inputs (makes sense to me), then I agree the TODO can be removed. AFAICT it shouldn't be a problem if both are set.

tapfreighter/wallet.go
jharveyb: Wish there was a linter for this!

cmd/tapcli/assets.go
jharveyb: I think both callers, `finalizeBatch` and `sendAssets`, use sat/vB rn?

guggero (approved):

Issue #608, opened 2023-10-23
[bug]:Universe Sync Meet bug "received message larger than max (5028775 vs. 4194304)"


MaxBipal:
### Background

** universe sync error **

[tapcli] rpc error: code = Unknown desc = unable to sync universe: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (5028775 vs. 4194304)

I thought this is a bug because the universe has a giant asset list data and is larger than the default value

### environment
```
41535046832b:~/.taprooot-assets# tapcli --network=mainnet --tapddir=/root/.taprooot-assets getinfo
{
    "version": "0.3.0-alpha commit=v0.3.0",
    "lnd_version": "0.17.0-beta",
    "network": "mainnet",
    "lnd_identity_pubkey": "02bedb301348812343c947d743ba42c5c5b170ea0f5359d5d945b568009f22ef21",
    "node_alias": "mar-uxuy-lnd",
    "block_height": 813449,
    "block_hash": "00000000000000000003ea9097e1d0ab27582d3339cf1afcdab566564632b0c3",
    "sync_to_chain": true
}

41535046832b:~/.taprooot-assets# uname -mrsv
Linux 6.4.16-linuxkit #1 SMP PREEMPT Sat Sep 23 13:36:48 UTC 2023 x86_64
```

### Expected behavior

Sync Data with Universe

### Actual behavior

```
41535046832b:~/.taprooot-assets# tapcli --network=mainnet --tapddir=/root/.taprooot-assets universe s --universe_host universe.lightning.finance:10029
[tapcli] rpc error: code = Unknown desc = unable to sync universe: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (5028775 vs. 4194304)
```

guggero:
Thanks for the report. Looks like we forgot to use the default message size in some places. Created a fix here: #609.

PR #609, opened 2023-10-23
multi: add proof courier addr to CLI, fix gRPC max message size, fix linter

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/608.

### DIFF
--- a/cmd/tapcli/addrs.go
+++ b/cmd/tapcli/addrs.go
@@ -27,11 +27,10 @@ var addrCommands = []cli.Command{
 }
 
 const (
-	groupKeyName = "group_key"
-
-	amtName = "amt"
-
-	assetVersionName = "asset_version"
+	groupKeyName         = "group_key"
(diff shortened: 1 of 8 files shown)


ffranr (approved):
Nice :100:

Roasbeef:
Thing is, we'll run into this in short order, as the issue is we don't have pagination.

guggero:
> Thing is, we'll run into this in short order, as the issue is we don't have pagination.
> 
> EDIT: looked the recv limit is pretty wumbo, which should hold us over, but do we want to continue sending over these gobs of data over w/o pagination?

I agree, this isn't the long-term fix. But since we had different max sizes and the same constant copied everywhere, I think we should still apply this fix just for clarity's sake.

jharveyb (approved):
Looks good! Didn't find any spots where the constant wasn't replaced.

MaxBipal:
Cool

PR #610, opened 2023-10-23
Rename universe functionality to accomodate transfer proofs.

ffranr:
Our universe functionality now handles transfer proofs in addition to issuance proofs. This PR renames functions and structures for clarity and correctness.


dstadulis:
Just code clean up,
we're using transfer proofs through this code areas, so this increases the accuracy

universe/auto_syncer.go
```diff
@@ -76,7 +76,7 @@ type FederationPushReq struct {
 // FederationIssuanceBatchPushReq is used to push out a batch of new issuance
 // events to all or some members of the federation.
 type FederationIssuanceBatchPushReq struct {
-	IssuanceBatch []*IssuanceItem
+	IssuanceBatch []*Item
```
guggero, line 79: Do we want to rename this (and the struct name) as well? Or is this actually just for issuances?

guggero (approved):
LGTM :tada:

universe/auto_syncer.go
ffranr: Yes, I've added more commits to address this.

ffranr:
There are probably more changes that we need to make along the same lines, but I would prefer to handle those in another PR unless they're pressing.

guggero (approved):

jharveyb (approved):
LGTM 👍🏽 

Will definitely be helpful when debugging these areas in the future.

PR #612, opened 2023-10-23
universe: make additional details optional to speed up query

guggero:
This makes the additional DB query for each grouped asset optional and hides it behind an additional gRPC/REST parameter.


taprpc/universerpc/universe.swagger.json
```diff
@@ -791,6 +791,15 @@
             }
           }
         },
+        "parameters": [
+          {
+            "name": "with_amounts_by_id",
+            "description": "TODO(roasbeef): filter by asset ID, etc?.",
```
dstadulis, line 797: Is this TODO supposed to be here?

guggero: Nice catch! The comment was there previously but wasn't mapped to a specific field. Updated that.

jharveyb (approved):
LGTM! I see the linter issue is fixed in #609.

Do we want to update the CLI to maintain the old behavior, or expose this setting? In favor of just keeping the old behavior imo.

https://github.com/lightninglabs/taproot-assets/blob/c7d57c597dfaadf0732a9ffb4bf63152c427f8f8/cmd/tapcli/universe.go#L157

Roasbeef:
Was this guided optimization via `EXPLAIN ANALYZE`, or something else? Do you have that dump we can use to add more detail to the commit body?

Roasbeef:
Ah reading more now, I see it's to just leave off some extra queries if/when only the roots are needed. 

I think we'll want to have the REST interface display by default, or just make sure the new query param there is documented.

guggero:
> I think we'll want to have the REST interface display by default, or just make sure the new query param there is documented.

Yeah, I should've added more info here, sorry.
So this extra loop to fetch all assets within a group was a user experience improvement added in https://github.com/lightninglabs/taproot-assets/pull/492 to better understand the output of the roots call.
But because this does an extra DB call for each group, it is very expensive. So we definitely should **NOT** turn this on for universe sync _and_ the REST call. I checked with @jamaljsr and we don't use the extra information on Terminal Web.
The extra info is only required in Polar where Jamal is going to add the query parameter (but there it will just query the local universe).

But I think we do want the CLI to remain at the previous behavior, so I'm going to fix that.

cmd/tapcli/universe.go
```diff
@@ -70,6 +71,11 @@ var universeRootsCommand = cli.Command{
 				"either 'issuance' or 'transfer'",
 			Value: universe.ProofTypeIssuance.String(),
 		},
+		cli.BoolFlag{
+			Name: skipAmountsByIdName,
+			Usage: "skip showing the amounts by ID for grouped " +
+				"assets to optimize response size and speed",
+		},
 	},
```
ffranr, line 79: Doesn't this default to `false`? And so we wont skip? I think our intent is to skip by default. I think the flag should capture the positive case here: `withAmountsById`.

ffranr: > But I think we do want the CLI to remain at the previous behavior, so I'm going to fix that.

I see this comment from you now.

ffranr (approved):
Looks good :fireworks:

Roasbeef (approved):
LGTM 🐌

Issue #512, opened 2023-09-19
Implement backoff procedure in universe proof courier


ffranr:
The hashmail proof courier uses a backoff procedure to re-attempt proof transfer if an attempt fails. We should use a similar procedure in the universe proof courier.

This change involves abstracting the backoff functionality from the hashmail proof courier and then using it in the universe proof courier.

Roasbeef:
Keeping this open, as we still need receiver back off.

dstadulis:
Receiver side will land by EoD Monday
SQL tables will be changed and scope is complete

ffranr:
We should add an itest for the universe proof courier which is similar to `testOfflineReceiverEventuallyReceives`.

dstadulis:
objective is to get the PR in now and then add test in `v0.3.2`

Issue #478, opened 2023-08-30
proof: implement encrypted proof payload for hashmail and other mailbox based couriers 


Roasbeef:
### Background

Today we use the hashmail courier to send proofs from sender to receiver. One unresolved TODO is to use an encryption scheme to preserve privacy between the sender and receiver: https://github.com/lightninglabs/taproot-assets/blob/4a0a16cc166f4bab1819947a7eba4ec368779b15/proof/courier.go#L571-L583

This can be generalized to other mailbox-like transports such as nostr. 

In the TODO above, I propose using the Elliptic Curve Integrated Encryption Scheme (ECIES), which is a hybrid encryption scheme. In our domain, we can use the script key of the receiver as the main DH point, with the sender generating an ephemeral secret themselves, then doing DH between this points to derive a secret key. The secret key would then be used to encrypt the payload with an AEAD scheme. The final payload is the ephemeral DH point, concatenated with the AEAD payload. 

## Steps To Completion 

- [ ] Decide if the courier interface should make this opaque, or create a new superset interface, `EncryptedProofCourier`
    * No big trade offs here imo, just if we want to have the encryption step be a distinct phase or not. It _will_ require the ability to do a DH with a key, and for the receiver the same as well. 
    * Control-flow wise, the diff would be handling the encryption before/after passing the payload to the courier. 
    

- [ ] Implement [ECIES](https://en.wikipedia.org/wiki/Integrated_Encryption_Scheme) in a new module (maybe to be upstreamed to `btcd/btcec` eventually? For the symmetric encryption scheme, we'd likely go with ChaChaPoly. 
    * Importantly, we'd want to include the DH point in side the associated data. This serves to fully authenticate the entire payload. 

- [ ] Update the existing hashmail impl to use this new scheme. We may want to use a new URI for this? Depends on timing w.r.t v0.3 IMO. Either way, clients should have a way to detect if this is being used or not.

jharveyb:
Instead of an ephemeral secret on the sender side, could we use the private key from one of the keys already being used by the sender during the transfer? Like the top-level taproot key for the first input to the TX.

Then  we could address #740 by both parties computing a DH point from those keys and use that to derive a hashmail stream ID / mailbox address.

Roasbeef:
Fixed by https://github.com/lightninglabs/taproot-assets/pull/1587

PR #617, opened 2023-10-24
Add backoff procedure to universe RPC proof courier

ffranr:
This PR extracts the backoff procedure from the hashmail proof courier and makes use of it in the universe RPC proof courier.

Fixes https://github.com/lightninglabs/taproot-assets/issues/512

### DIFF
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -398,9 +398,10 @@ func testBasicSendPassiveAsset(t *harnessTest) {
 	test.WriteTestVectors(t.t, proof.RegtestTestVectorName, testVectors)
 }
 
-// testReattemptFailedAssetSend tests that a failed attempt at sending an asset
-// proof will be reattempted by the tapd node.
-func testReattemptFailedAssetSend(t *harnessTest) {
+// testReattemptFailedAssetSendHashmailCourier tests that a failed attempt at
+// sending an asset proof will be reattempted by the tapd node. This test
+// targets the hashmail courier.
+func testReattemptFailedAssetSendHashmailCourier(t *harnessTest) {
 	var (
 		ctxb = context.Background()
 		wg   sync.WaitGroup
@@ -444,10 +445,11 @@ func testReattemptFailedAssetSend(t *harnessTest) {
 			return false
 		}
 
-		// Default number of proof delivery attempts in tests is 3,
-		// therefore expect at least 2 backoff wait events
-		// (not waiting on first attempt).
-		expectedEventCount := 2
+		// Expected number of events is one less than the number of
+		// tries because the first attempt does not count as a backoff
+		// event.
+		nodeBackoffCfg := t.tapd.clientCfg.HashMailCourier.BackoffCfg
+		expectedEventCount := nodeBackoffCfg.NumTries - 1
 
 		// Context timeout scales with expected number of events.
 		timeout := time.Duration(expectedEventCount) *
@@ -485,6 +487,11 @@ func testReattemptFailedAssetSend(t *harnessTest) {
 
 	// Simulate a failed attempt at sending the asset proof by stopping
 	// the receiver node.
+	//
+	// The receiving tapd node does not return a proof received confirmation
+	// message via the universe RPC courier. We can simulate a proof
+	// transfer failure by stopping the courier service directly and not the
+	// receiving tapd node.
 	require.NoError(t.t, t.tapd.stop(false))
 
 	// Send asset and then mine to confirm the associated on-chain tx.
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -501,6 +501,109 @@ func testReattemptFailedAssetSendHashmailCourier(t *harnessTest) {
 	wg.Wait()
 }
 
+// testReattemptFailedAssetSendUniCourier tests that a failed attempt at
+// sending an asset proof will be reattempted by the tapd node. This test
+// targets the universe proof courier.
+func testReattemptFailedAssetSendUniCourier(t *harnessTest) {
+	var (
+		ctxb = context.Background()
+		wg   sync.WaitGroup
+	)
+
+	// Make a new node which will send the asset to the primary tapd node.
+	// We expect this node to fail because our send call will time out
+	// whilst the porter continues to attempt to send the asset.
+	sendTapd := setupTapdHarness(
+		t.t, t, t.lndHarness.Bob, t.universeServer,
+		func(params *tapdHarnessParams) {
+			params.expectErrExit = true
+		},
+	)
+
+	// Subscribe to receive asset send events from primary tapd node.
+	eventNtfns, err := sendTapd.SubscribeSendAssetEventNtfns(
+		ctxb, &taprpc.SubscribeSendAssetEventNtfnsRequest{},
+	)
+	require.NoError(t.t, err)
+
+	// Test to ensure that we receive the expected number of backoff wait
+	// event notifications.
+	// This test is executed in a goroutine to ensure that we can receive
+	// the event notification(s) from the tapd node as the rest of the test
+	// proceeds.
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		// Define a target event selector to match the backoff wait
+		// event. This function selects for a specific event type.
+		targetEventSelector := func(event *taprpc.SendAssetEvent) bool {
+			switch eventTyped := event.Event.(type) {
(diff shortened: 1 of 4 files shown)


proof/courier.go
```diff
@@ -513,15 +514,16 @@ type Recipient struct {
...
+// or not.
+type BackoffExecError struct {
+	execErr error
+}
 
-	// BackoffCfg configures the behaviour of the proof delivery
-	// functionality.
-	BackoffCfg *BackoffCfg
+func (e *BackoffExecError) Error() string {
+	return fmt.Sprintf("backoff exec error: %s", e.execErr.Error())
```
guggero, line 536: Do we need a `nil` check on the `execErr` here? Or can we be reasonably sure it is always set?

ffranr:
I haven't added an itest which covers this new use of the backoff procedure in the universe proof courier. Let me know if you guys have any ideas along those lines.

proof/courier.go
```diff
@@ -543,21 +545,190 @@ type BackoffCfg struct {
 	MaxBackoff time.Duration `long:"maxbackoff" description:"The maximum backoff time to wait before retrying to deliver the proof to the receiver."`
 }
 
+// BackoffHandle is a handle to the backoff procedure.
+type BackoffHandle struct {
```
guggero, line 564: nit: since this is a concrete type/struct, wouldn't it be a handle**r** (instead of just handle)? I initially thought this was a typo but it seems very consistent throughout the diff.

proof/courier.go
```diff
@@ -543,21 +545,190 @@ type BackoffCfg struct {
...
+	timestamps, err := b.deliveryLog.QueryProofDeliveryLog(
+		ctx, proofLocator,
+	)
+	if err != nil {
+		return fmt.Errorf("unable to retrieve proof delivery "+
+			"logs: %w", err)
+	}
+
+	if len(timestamps) == 0 {
+		log.Infof("No previous proof delivery attempts found "+
```
guggero, line 601: This sounds like a `Debugf` or even `Tracef` log.

guggero: Same for the other log outputs in this method (info should be important enough for the user to actually make it into the log).

proof/courier.go
```diff
@@ -543,21 +545,190 @@ type BackoffCfg struct {
...
+	BackoffCfg *BackoffCfg
+}
+
 // HashMailCourier is a hashmail proof courier service handle. It implements the
 // Courier interface.
 type HashMailCourier struct {
 	// cfg contains the courier's configuration parameters.
 	cfg *HashMailCourierCfg
 
+	// BackoffHandle handles the backoff procedure for the proof delivery
```
guggero, line 758: nit: Godoc comment mismatch (case).

proof/courier.go
```diff
@@ -581,89 +752,50 @@ func (h *HashMailCourier) DeliverProof(ctx context.Context,
...
-	//
-	// Only wait if we have a non-zero number of past delivery attempts.
-	timeSinceLastAttempt := timeSinceLastDeliveryAttempt(timestamps)
-	backoffResetWait := h.cfg.BackoffCfg.BackoffResetWait
-	if len(timestamps) > 0 &&
-		timeSinceLastAttempt < backoffResetWait {
+		// Now that the stream has been initialized, we'll write
+		// the proof over the stream.
+		//
+		// TODO(roasbeef): do ecies here
```
guggero, line 804: What could the TODO here mean? Maybe a typo of "retries"? Then I guess it would be addressed now.

proof/courier.go
```diff
@@ -622,10 +639,17 @@ func (b *BackoffHandle) Exec(ctx context.Context, proofLocator Locator,
 		return fmt.Errorf("backoff config not specified")
 	}
 
+	locatorHash, err := proofLocator.Hash()
+	if err != nil {
+		return err
+	}
+	log.Infof("Starting proof delivery backoff procedure for proof: "+
+		"(locator_hash=%x)", locatorHash)
```
guggero, line 647: I think this logs as double hashed, since it's a hash that has a `String()` method. So should proably be `locatorHash[:]` instead.

proof/courier.go
```diff
@@ -984,6 +1008,10 @@ type UniverseRpcCourier struct {
 	// the universe RPC server.
 	client unirpc.UniverseClient
 
+	// BackoffHandle handles the backoff procedure for the proof delivery
```
guggero, line 1011: nit: same here re godoc comment case mismatch.

guggero (approved):
Very nice! Looks great, nothing much to add other than nits.

I think as a follow-up PR we should enable the RPC universe courier by default in the itests and remove all calls to `sendProof()` (which circumvents any courier at the moment).

proof/courier.go
jharveyb: AFAICT `BackoffHandle.Exec()` is the only place we're building such an error. But a nil check would be good anyways, could just return "backoff exec error" imo.

proof/courier.go
jharveyb: Agree, I think with that change details about the backoff would only get logged on failure, with the final error about # of attempts.

proof/courier.go
jharveyb: Not addressed, related to encrypting proofs for the receiver before uploading to the courier:

https://github.com/lightninglabs/taproot-assets/issues/478

jharveyb (commented):
Looks good, nice refactor to make this reusable!

I think one way to add this into the itests would just be to have a test that's a duplicate of `testReattemptFailedAssetSend`, but with `proofCourierType: proof.UniverseRpcCourierType`.

When I do that, the test hangs as the transferEvents are never broadcast. Issue may be in `deliverFunc()` but I didn't dig in further.

proof/courier.go
ffranr: I've added an additional comment to this TODO

ffranr:
The latest commits address the review comments and add an itest to cover the changes introduced in this PR.

jharveyb (approved):
Good to see the itest changes were small! LGTM 👍🏽

proof/courier.go
```diff
@@ -1012,20 +1106,27 @@ func (c *UniverseRpcCourier) DeliverProof(ctx context.Context,
...
 
-		// Submit proof to courier.
-		_, err = c.client.InsertProof(ctx, &unirpc.AssetProof{
-			Key:       &universeKey,
-			AssetLeaf: &assetLeaf,
-		})
+		// Setup delivery routine and start backoff procedure.
+		deliverFunc := func() error {
+			// Submit proof to courier.
+			_, err = c.client.InsertProof(ctx, &unirpc.AssetProof{
```
Roasbeef, line 1113: This only adds back off for proof insertion. What we've seen fail is that the receiver attempts to look up, but maybe got the block before the sender, then never retries again.

We also need the receiver to properly retry.

itest/send_test.go
```diff
@@ -492,6 +501,109 @@ func testReattemptFailedAssetSend(t *harnessTest) {
...
+	AssertAddrCreated(t.t, t.tapd, rpcAssets[0], recvAddr)
+
+	// Simulate a failed attempt at sending the asset proof by stopping
+	// the proof courier service.
+	//
+	// In following the hashmail proof courier protocol, the receiver node
+	// returns a proof received confirmation message via the courier.
+	// We can simulate a proof transfer failure by stopping the receiving
+	// tapd node. The courier service should still be operational.
+	require.NoError(t.t, t.proofCourier.Stop())
```
Roasbeef, line 598: How is this testing the back off? We want to make sure that if the receiver gets there too early, they continue to retry in the background.

Roasbeef (changes requested):

itest/send_test.go
jharveyb: I think this does properly test it for the sender as-is.

guggero: I was a bit confused at the beginning as well. But basically what this test does is test that we get a `ReceiverProofBackoffWaitEvent` event if the proof courier isn't started. Meaning the sender tries to send, fails then backs off and tries _again_.
But the test doesn't currently assert the successful submission.

IMO we can add the backoff-retry mechanism for the receiver in a separate PR and there extend both of these two tests to make sure that the transmission eventually works if the proof courier is turned on.

guggero (approved):
Nice, LGTM :tada: 

I suggest doing the backoff-retry for the _receiver_ in a separate PR and get this in as is.

PR #622, opened 2023-10-24
tapd: allow configuring all connection params for Postgres

guggero:
We want to be able to specify the number of idle connections, and other lifetime parameters for Postgres as well.


Roasbeef (approved):
LGTM 🎿

Issue #621, opened 2023-10-24
universe: create ephemeral multiverse roots to reduce bandwidth for specific syncs


Roasbeef:
When a client goes to fetch a multiverse, we should read the relevant roots, and create an in-memory multiverse root to send to the client.

Roasbeef:
Keeping this one open, as the syncer hasn't been updated to use the new methods yet.

dstadulis:
TODO:
Decision point on if
- [ ] Streaming 
vs
- [ ] Not streaming: enables HTTP functional

Issue #574, opened 2023-10-12
universe: add overlay points for the multiverse root and leaves


Roasbeef:
The way the universe tree works today is that we have the base SMT, then we have an overlay that points to each of the leaves, and also the root. This makes it easy to query for a given universe trees, as we use the SMT namespacing underneath.

On disk, this looks something like: https://github.com/lightninglabs/taproot-assets/blob/a990fdf18c2ca13cbb825c9aa95866831ccfadf8/tapdb/sqlc/migrations/000007_universe.up.sql#L1-L41

We also want this for the multiverse tree as well. Otherwise, we can't easily do things like query for all the leaves in a multiverse tree (issuance or transfer). 

We should replicate the above structure for for the two multiverse trees we maintain. 

## Steps to Completion 

- [ ] Create a new `multiverse_roots` and `multiverse_leaves` table. For the root, we just care if it's issuance or transfer. We should copy over the `namespace_root` field for the root itself. For the leaves, we just need either `asset_id` and `group_key`, we can drop items like the minting point, the script key, etc. 

- [ ] Each time we go to modify the multiverse tree, upsert and root and leaf after the operation. 

- [ ] Add new methods to query the set of multiverse leaves and also fetch arbitrary proofs. Ofc, add unit tests for all this.

PR #624, opened 2023-10-25
multiverse: add overlay points for universe trees

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/574.
Fixes https://github.com/lightninglabs/taproot-assets/issues/621.

Depends on the caching PR (not up yet).


tapdb/sqlc/migrations/000010_multiverse_tree.up.sql
```diff
@@ -49,6 +49,12 @@ WHERE EXISTS (
     SELECT 1 FROM mssmt_roots WHERE namespace = 'multiverse-transfer'
 );
 
+-- Before we turn the existing universe_roots into multiverse_leaves, we need to
+-- fix a bug in the data where the asset ID was set for grouped assets.
+UPDATE universe_roots
```
Roasbeef, line 54: Will this resolve that issue where we saw we had two roots for an asset: with for the asset ID, and the other for the actual asset group?

tapdb/universe.go
```diff
@@ -405,7 +407,7 @@ func universeUpsertProofLeaf(ctx context.Context, dbTx BaseUniverseStore,
 	// root ID that we'll use to insert the universe leaf overlay.
 	universeRootID, err := dbTx.UpsertUniverseRoot(ctx, NewUniverseRoot{
 		NamespaceRoot: namespace,
-		AssetID:       fn.ByteSlice(leaf.ID()),
+		AssetID:       assetIDBytes,
```
Roasbeef, line 410: I think this is needed for the existing queries to work, as this is how we link it to a genesis asset.

tapdb/sqlc/migrations/000010_multiverse_tree.up.sql
```diff
@@ -0,0 +1,80 @@
...
+    -- Both the asset ID and group key cannot be null at the same time.
+    CHECK (asset_id IS NOT NULL OR group_key IS NOT NULL),
+    
+    UNIQUE(asset_id, group_key)
+);
+
+-- If there already is a multiverse root entry in the mssmt_roots for the
+-- issuance or transfer multiverses, add them to the multiverse_roots table as
+-- well. Both statements are no-ops if the root doesn't exist yet.
+INSERT INTO multiverse_roots (namespace_root, proof_type)
```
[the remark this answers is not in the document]
guggero: Yes, exactly. We only insert them for universes we already have in the DB. If there are none, nothing happens and they are upserted when they are first pushed during a sync.

tapdb/sqlc/migrations/000010_multiverse_tree.up.sql
guggero: I thought so at first, but it turns out that was an issue with transfer proofs showing up in the stats. See fix here: https://github.com/lightninglabs/taproot-assets/pull/628.
So I'm going to remove that "asset ID fix" commit again, as it broke some queries as you mentioned below.
I'm just using `CASE WHEN ur.group_key IS NULL THEN ur.asset_id ELSE NULL END,` now for the `multiverse_leaves` insert.

tapdb/universe.go
guggero: Yes, you are right. Removed this commit again and made sure we only use either value in `multiverse_leaves`.

tapdb/sqlc/queries/universe.sql
```diff
@@ -345,4 +345,29 @@ ON CONFLICT(namespace)
...
+) VALUES (
+    @multiverse_root_id, @asset_id, @group_key, @leaf_node_key,
+    @leaf_node_namespace
+)
+ON CONFLICT (asset_id, group_key)
+    DO UPDATE SET leaf_node_key = EXCLUDED.leaf_node_key
+RETURNING id;
+
+-- name: QueryMultiverseLeaves :many
+SELECT r.namespace_root, r.proof_type, l.asset_id, l.group_key, l.leaf_node_key
```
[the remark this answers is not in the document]
guggero: Yep, makes sense.

guggero:
Addressed all comments so far and added unit tests.
Main thing missing now are just the RPC methods.

guggero:
Added RPCs and integration tests.
This now also fixes https://github.com/lightninglabs/taproot-assets/issues/621 I believe. Or did you have more specific functionality in mind for that, @Roasbeef?

universe/base.go
```diff
@@ -126,6 +127,71 @@ func (a *Archive) RootNodes(ctx context.Context,
...
+// MultiverseRoot returns the root node of the multiverse for the specified
+// proof type. If the given list of universe IDs is non-empty, then the root
+// will be calculated just for those universes.
+func (a *Archive) MultiverseRoot(ctx context.Context, proofType ProofType,
+	filterByIDs []Identifier) (mssmt.Node, error) {
+
+	log.Debugf("Fetching multiverse root for proof type: %v", proofType)
+
+	leaveIDs, err := a.cfg.Multiverse.FetchMultiverseLeaves(
+		ctx, nil, nil, proofType,
```
Roasbeef, line 139: Rather than fetch everything all the time, we can instead pass the identifier in (if they exist) and use db filtering for it.

universe/base.go
```diff
@@ -126,6 +127,71 @@ func (a *Archive) RootNodes(ctx context.Context,
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to fetch universe "+
+				"root: %w", err)
+		}
+
+		rootHash := uniRoot.NodeHash()
+		rootSum := uniRoot.NodeSum()
+
+		if id.ProofType == ProofTypeIssuance {
+			rootSum = 1
```
Roasbeef, line 182: We _should_ just be able to use the existing root sum here.

Roasbeef:
Will re-open this against my main branch. I think the only thing that needs to change is that: the leaf value for the multiverse tree is the root of a universe tree itself. So when we go to try and make the in memory version of the tree, we can just use the value directly (hash+sum).

tapdb/sqlc/migrations/000012_multiverse_tree.up.sql
```diff
@@ -0,0 +1,83 @@
...
+
+INSERT INTO multiverse_roots (namespace_root, proof_type)
+SELECT 'multiverse-transfer', 'transfer'
+WHERE EXISTS (
+    SELECT 1 FROM mssmt_roots WHERE namespace = 'multiverse-transfer'
+);
+
+-- And now we create the multiverse_leaves entries for the multiverse roots.
+-- This is a no-op if the multiverse root doesn't exist yet.
+INSERT INTO multiverse_leaves (
```
Roasbeef, line 55: This works on the overlay level, but then the underlying SMT tree wouldn't have the proper leaves/branches inserted. 

I think in addition to this, we also need a runtime/startup check where we insert the items into the tree if no actual leaves exist. We can check this with a query that properly joins against `mssmt_nodes`

Roasbeef: Actually this is ok. The leaves already exist in the multiverse tree, so the new join allows us to obtain that value and optimize some of the added logic (reduce number of DB transactions).

Roasbeef:
@ffranr @jharveyb PTAL

ffranr (approved):
Nice to have those new `fn` types! :100:

fn/option.go
```diff
@@ -0,0 +1,149 @@
+package fn
+
+// Option[A] represents a value which may or may not be there. This is very
+// often preferable to nil-able pointers.
+type Option[A any] struct {
+	isSome bool
+	some   A
+}
```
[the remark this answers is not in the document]
Roasbeef: I think once we figure out all the useful patterns yeah, this workflow just lets us add stuff easily as we go, w/o having to first touch multiple repos w/ tags+modules, etc.

universe/base.go
```diff
@@ -134,6 +135,71 @@ func (a *Archive) RootNodes(ctx context.Context,
...
+	}
+
+	// If a filter list is provided, then we'll only include the leaves
+	// that are in the filter list.
+	includeUniverse := func(id Identifier) bool {
+		if len(filterByIDs) == 0 {
+			return true
+		}
+
+		for _, filterID := range filterByIDs {
```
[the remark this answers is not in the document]
Roasbeef: I left this in place for oli's commits, the newer version doesn't do this type of scan.

tapdb/sqlc/migrations/000012_multiverse_tree.up.sql
```diff
@@ -0,0 +1,83 @@
+CREATE TABLE IF NOT EXISTS multiverse_roots (
+    id BIGINT PRIMARY KEY,
+
+    -- For the namespace root, we set the foreign key constraint evaluation to
+    -- be deferred until after the database transaction ends. Otherwise, if the
+    -- root of the SMT is deleted temporarily before inserting a new root, then
+    -- this constraint is violated as there's no longer a root that this
+    -- universe tree can point to.
+    namespace_root VARCHAR UNIQUE NOT NULL REFERENCES mssmt_roots(namespace) DEFERRABLE INITIALLY DEFERRED,
```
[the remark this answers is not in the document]
jharveyb: Agree but maybe better done by the  caller of insert/upsert? i don't think we consistently do one or the other but that approach seems better IMO.

tapdb/sqlc/migrations/000012_multiverse_tree.down.sql
```diff
@@ -0,0 +1,3 @@
+DROP INDEX IF EXISTS multiverse_leaves_unique;
```
jharveyb, line 1: I don't see this index ever constructed?

tapdb/sqlc/migrations/000012_multiverse_tree.up.sql
```diff
@@ -0,0 +1,83 @@
...
+    multiverse_root_id, asset_id, group_key, leaf_node_key, leaf_node_namespace
+) SELECT
+      (SELECT id from multiverse_roots where namespace_root = 'multiverse-issuance'),
+      CASE WHEN ur.group_key IS NULL THEN ur.asset_id ELSE NULL END,
+      ur.group_key,
+      -- UNHEX() only exists in SQLite and it doesn't take a second argument
+      -- (the 'hex' part). But it also doesn't complain about it, so we can
+      -- leave it in for the Postgres version which is replaced in-memory to
+      -- DECODE() which needs the 'hex' argument.
+      UNHEX(REPLACE(namespace_root, 'issuance-', ''), 'hex'),
```
jharveyb, line 65: https://sqlite.org/lang_corefunc.html#unhex

Unhex does take a second arg, which seem like it's intended to restrict the set of hex characters that can be decoded to what's in the second arg? Or am I misreading the docs here.

Related, I'm not sure why this field should be a `BLOB` here but `VARCHAR` in the universe tables.

jharveyb, line 65: Should this be `ur.namespace_root`?

Related, I'm not sure how this works as-is. IIUC, we are trimming the prefix `issuance-` or `transfer-` from the universe namespace root. That leaves us with the universe ID serialized as bytes, which will be the same for the issuance and transfer universe trees.

Since `leaf_node_key` is unique, we wouldn't be able to store two entries in the `multiverse_leaves` to represent both the issuance and transfer universe trees for a given asset.

I think in the added tests we only test with trees where we insert universe leaves of one proof type and not both.

jharveyb (commented):
Only reviewed the first few commits but a bit confused on the table structure overall and the mapping for the universe roots to multiverse leaves here.

jharveyb:
Finished review, the following commits look good. Excited for the `Option` and `Either` types.

universe/base.go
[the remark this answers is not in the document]
guggero: Yeah, now we make multiple DB queries instead :sweat_smile: Not sure which will be faster in the end, I guess it depends on how many we query explicitly.

tapdb/sqlc/migrations/000012_multiverse_tree.down.sql
guggero: No, you're right, it doesn't exist. Left over from when I tried something.

tapdb/sqlc/migrations/000012_multiverse_tree.up.sql
guggero: Ah, you're right, there is a second argument. But it looks like we're fine with the value `'hex'` being passed in as `Y`:
> Hexadecimal digits in Y have no affect on the translation of X. Only characters in Y that are not hexadecimal digits are ignored in X. 

It does mean that the hex string `00xx11xx22` would be decoded as `0x00, 0x11, 0x22` because of the `x` in the second parameter. But I think that's okay since we know we only have valid hex.

> Related, I'm not sure why this field should be a BLOB here but VARCHAR in the universe tables.

I think SQLite internally doesn't really make a difference between BLOB and VARCHAR and can seamlessly convert between the two.

dstadulis:
Need to add another unit test that @jharveyb highlighted

guggero:
Okay, updated unit test is in, this would now catch the issue with the unique index. Ready for final review.

Roasbeef (approved):
LGTM 🔧

jharveyb:
Post-merge approval 💯

PR #626, opened 2023-10-26
multi: add new caches for universe proofs, roots, keys, and config

Roasbeef:
In this PR, we add a series of LRU caches to the Universe data for a public server. Rather than use write through caches, we simply purge the cache once new information is available, leaving it to the next caller to grab a consistent view of the cache. 

We also maintain an in memory version of our configs as today, each time we want to fetch/insert a proof, we need to also hit the db (before this PR) to obtain that information.


tapdb/multiverse.go
```diff
@@ -61,21 +70,238 @@ type BatchedMultiverse interface {
...
+	return fn.ToArray[ProofKey](h.Sum(nil))
+}
+
+// numCachedProofs is the number of universe proofs we'll cache.
+const numCachedProofs = 25_000
+
+// cachedProof is a single cached proof.
+type cachedProof []*universe.Proof
+
+// Size just returns 1 as we're limiting based on the total number of proofs.
```
guggero, line 99: Related to the comment above this would the number of groups in the cache but the actual number of proofs can well exceed 25k if there are a lot of large groups.

tapdb/multiverse.go
```diff
@@ -61,21 +70,238 @@ type BatchedMultiverse interface {
...
+func (c *cachedProof) Size() (uint64, error) {
+	return 1, nil
+}
+
+// leafProofCache is used to cache proofs for issuance leaves for assets w/o a
+// group key.
+type leafProofCache = *lru.Cache[ProofKey, *cachedProof]
```
guggero, line 106: IMO it's quite hard to reason about type aliases that are referencing a pointer type. Should we just do `type leafProofCache = *lru.Cache[ProofKey, *cachedProof]` and then return a pointer of `leafProofCache` below?

guggero: This would also better match the pattern of the `proofCache` below.

tapdb/multiverse.go
```diff
@@ -61,21 +70,238 @@ type BatchedMultiverse interface {
...
+
+// atomicRootIndex is an atomic pointer to a root index.
+type atomicRootIndex = atomic.Pointer[rootIndex]
+
+// newAtomicRootIndex creates a new atomic root index.
+func newAtomicRootIndex() atomicRootIndex {
+	var a atomicRootIndex
+	a.Store(&rootIndex{})
+
+	return a
```
guggero, line 202: Same here re copying a struct with an embedded mutex.

tapdb/multiverse.go
```diff
@@ -61,21 +70,238 @@ type BatchedMultiverse interface {
...
+	// Attempt to read directly from the root node cache.
+	rootNodeCache := r.allRoots.Load()
+	rootNodes := *rootNodeCache
+
+	return rootNodes
+}
+
+// fetchRoot reads the cached root for the given ID.
+func (r *rootNodeCache) fetchRoot(id universe.Identifier,
+) *universe.Root {
```
guggero, line 254: nit: fits on previous line.

tapdb/universe_federation.go
```diff
@@ -124,22 +126,43 @@ type BatchedUniverseServerStore interface {
...
+
+	globalCfgPtr.Store(&globalSyncCfgs{})
+	assetCfgsPtr.Store(&assetSyncCfgs{})
+
 	return &UniverseFederationDB{
-		db:    db,
-		clock: clock,
+		db:        db,
+		clock:     clock,
+		globalCfg: globalCfgPtr,
```
guggero, line 164: Also a linter issue with copying something with an embedded Mutex.

tapdb/multiverse.go
```diff
@@ -283,6 +283,69 @@ func (r *rootNodeCache) wipeCache() {
...
+	}
+}
+
+// fetchLeafKeys reads the cached leaf keys for the given ID.
+func (u *universeLeafCache) fetchLeafKeys(id universe.Identifier,
+) []universe.LeafKey {
+
+	leafKeys, err := u.leafCache.Get(treeID(id.String()))
+	if err == nil {
+		log.Debugf("read leaf keys for %v from cache",
```
guggero, line 325: nit: demote to trace to avoid spamming the log? As we'd expect to hit this quite often I think.

tapdb/multiverse.go
```diff
@@ -356,6 +422,81 @@ func (b *MultiverseStore) RootNode(ctx context.Context,
...
+		return universe.Root{}, dbErr
+	}
+
+	var nodeHash mssmt.NodeHash
+	copy(nodeHash[:], universeRoot.RootHash[:])
+	smtNode := mssmt.NewComputedNode(
+		nodeHash, uint64(universeRoot.RootSum),
+	)
+
+	return universe.Root{
```
guggero, line 463: Since we fetched the root node from the DB, do we also want to cache it now? Or is that done on another layer?

tapdb/multiverse.go
```diff
@@ -61,21 +70,238 @@ type BatchedMultiverse interface {
...
+
+// fetchRoots reads the cached roots for the given proof type. If the amounts
+// are needed, then we return nothing so we go to the database to fetch the
+// information.
+func (r *rootNodeCache) fetchRoots(withAmts, haveWriteLock bool,
+) []universe.Root {
+
+	// We don't cache roots with amounts, so if the caller wants the
+	// amounts, they'll go to disk.
+	if withAmts {
```
guggero, line 233: Non-blocking for this PR, but I'm wondering if we should add the option to disallow this call on public universe instances, since it bypasses the cache?

guggero (commented):
Great work with the cache! Makes a huge difference :100:

tapdb/multiverse.go
Roasbeef: So this actually just always maps to one value, the API on the DB level lets you try to get stuff like all the proofs for a given outpoint (not binding script key). We go `treeID -> proofKey`. Each of the new group instances will have a diff `proofKey`.

tapdb/universe_test.go
```diff
@@ -513,6 +513,13 @@ func TestUniverseTreeIsolation(t *testing.T) {
 	require.Nil(t, normalRoot)
 	require.ErrorIs(t, err, universe.ErrNoUniverseRoot)
 
+	for _, rootNode := range rootNodes {
+		// TODO(roasbeef): need base universe -> universe cache
+		// invalidation or delete thru multiverse
+		multiverse.rootNodeCache.Load().Delete(treeID(rootNode.ID.String()))
```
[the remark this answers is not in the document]
jharveyb: Flagged as outdated but still applies.

tapdb/multiverse.go
jharveyb: Do you mean that `FetchProofLeaf` is always called with the `universeKey.ScriptKey` set, so the returned array is always one element?

If so we could assert that before updating the cache and then make `cachedProof` not an array - was also confused by this and the impact on cache size.

tapdb/multiverse.go
```diff
@@ -61,21 +70,306 @@ type BatchedMultiverse interface {
...
+}
+
+// fetchProof reads the cached proof for the given ID and leaf key.
+func (p *proofCache) fetchProof(id universe.Identifier,
+	leafKey universe.LeafKey) []*universe.Proof {
+
+	// First, get the sub-cache for this universe ID from the map of
+	// caches.
+	idStr := treeID(id.String())
+	proofCache, _ := p.LoadOrStore(idStr, newLeafCache())
```
jharveyb, line 133: nit" could exit early here based on the dropped return value from `LoadOrStore`; if we had no leaf cache before there would be nothing to fetch.

tapdb/multiverse.go
```diff
@@ -61,21 +70,306 @@ type BatchedMultiverse interface {
...
+	assetProofCache := proofCache.(leafProofCache)
+
+	// With that lower level cache obtained, we can check to see if we have
+	// a hit or not.
+	proofKey := NewProofKey(id, leafKey)
+	proof, err := assetProofCache.Get(proofKey)
+	if err == nil {
+		log.Debugf("read proof for %v+%v from cache, key=%v",
```
jharveyb, line 141: Would demote this (and most other new debug log statements0 to Tracef.

tapdb/multiverse.go
```diff
@@ -61,21 +70,306 @@ type BatchedMultiverse interface {
...
+	a.Store(treeCache)
+
+	return a
+}
+
+// rootIndex maps a tree ID to a universe root.
+type rootIndex = lnutils.SyncMap[treeID, *universe.Root]
+
+// atomicRootIndex is an atomic pointer to a root index.
+type atomicRootIndex = atomic.Pointer[rootIndex]
```
jharveyb, line 195: I guess these indexes backed by `SyncMap`s don't necessarily need a size limit as they are getting wiped p often?

Also, here the `atomic.Pointer` is basically acting as a mutex + pointer? Vs. only needing a `Mutex` for the new `lru.Cache` objects.

Roasbeef: Yeah exactly, using the atomic pointer we can just hot swap the pointer value to nil/empty, which then triggers the cache repopulation.

tapdb/multiverse.go
```diff
@@ -61,21 +70,238 @@ type BatchedMultiverse interface {
...
+	h.Write(leafKeyBytes[:])
+
+	return fn.ToArray[ProofKey](h.Sum(nil))
+}
+
+// numCachedProofs is the number of universe proofs we'll cache.
+const numCachedProofs = 25_000
+
+// cachedProof is a single cached proof.
+type cachedProof []*universe.Proof
```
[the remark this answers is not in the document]
Roasbeef: This is just the issuance proof for a proof key, which uniquely maps to a leaf. Kept the slice as that's the current return value.

tapdb/multiverse.go
Roasbeef: > Do you mean that FetchProofLeaf is always called with the universeKey.ScriptKey set, so the returned array is always one element?


Yeah it's always a single element. The functionality to expose the partial queries isn't exposed over RPC. 

The current limits were set just to make sure we could cache everything as is to stop hammering the disk. W/ the current "sync all behavior" and some of the problematic clients, we'd end up caching everything as the LRU size would quickly be exceeded. Agree we can tune some of these values down some though. At present, there're ~42k total assets.

tapdb/multiverse.go
Roasbeef: This is needed to have some of the declarations compile cleanly. Otherwise it needs to do some intermediate caching.

tapdb/multiverse.go
Roasbeef: Will look into this, but this is only used for initialization.

tapdb/multiverse.go
Roasbeef: Removed these in favor of the cache logger in the latest diff.

tapdb/universe.go
```diff
@@ -614,6 +613,62 @@ func universeFetchProofLeaf(ctx context.Context,
...
+
+				return q.Limit
+			}(),
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	var leafKeys []universe.LeafKey
```
jharveyb, line 639: maybe a place we'd want iterators over preallocated slices in addition to what we have now.

tapdb/multiverse.go
```diff
@@ -64,21 +73,420 @@ type BatchedMultiverse interface {
...
+// Size just returns 1, as we cache based on the total number of assets, but
+// not the sum of their leaves.
+func (c cachedLeafKeys) Size() (uint64, error) {
+	return uint64(1), nil
+}
+
+// numMaxCachedPages is the maximum number of pages we'll cache for a given
+// page cache. Each page is 512 items, so we'll cache 10 of them, up to 5,120
+// for a given namespace.
+const numMaxCachedPages = 1000
```
jharveyb, line 361: nit: godoc incorrect here?

tapdb/multiverse.go
```diff
@@ -64,21 +73,420 @@ type BatchedMultiverse interface {
...
+// leafKeysCache is used to cache the set of leaf keys for a given universe.
+// For each treeID we store an inner cache for the paginated responses.
+type leafKeysCache = lru.Cache[treeID, *leafPageCache]
+
+// universeLeafCaches is used to cache the set of leaf keys for a given
+// universe.
+type universeLeafCache struct {
+	sync.Mutex
+
+	leafCache *leafKeysCache
```
jharveyb, line 399: IIUC there is no limit on how many leaf page caches we make, but one limit on elements shared over all leaf page caches?

tapdb/multiverse.go
```diff
@@ -133,11 +541,133 @@ func (b *MultiverseStore) RootNode(ctx context.Context,
...
+func (b *MultiverseStore) UniverseRootNode(ctx context.Context,
+	id universe.Identifier) (universe.Root, error) {
+
+	// First, we'll check the root node cache to see if we already have it.
+	rootNode := b.rootNodeCache.fetchRoot(id)
+	if rootNode != nil {
+		return *rootNode, nil
+	}
+
+	b.rootNodeCache.Lock()
```
jharveyb, line 554: Was going to suggest something like this in the earlier commits; I think some of the other caches aren't using a RWMutex though?

Can address in follow-up PR.

tapdb/universe_stats.go
```diff
@@ -104,22 +109,238 @@ type BatchedUniverseStats interface {
...
+
+// eventQueryBucket is the interval that we'll use to bucket similar queries
+// into.
+const eventQueryBucket = time.Hour
+
+// newEventQuery creates a new event query from the given query.
+func newEventQuery(q universe.GroupedStatsQuery) eventQuery {
+	// For both the start and time time, we'll round down to the nearest
+	// hour. This'll serve to bucket queries into hourly buckets.
+	startTime := q.StartTime.UTC().Truncate(eventQueryBucket).Unix()
```
jharveyb, line 129: Do we want to implement some minimum query timespan of one hour here? Especially since the caller doesn't explicitly know we're truncating.

IIUC right now start and end times < 1 hr apart would return nothing.

tapdb/universe_stats.go
```diff
@@ -273,10 +494,88 @@ func (u *UniverseStats) AggregateSyncStats(
...
+		return
+	}
+
+	log.Debugf("Refreshed stats cache, interval=%v, took=%v",
+		u.opts.cacheDuration, time.Since(now))
+
+	u.statsSnapshot.Store(&dbStats)
+
+	// Reset the timer so we'll refresh again after the cache duration.
+	if !u.statsRefresh.Stop() {
```
jharveyb, line 534: 🚀

tapdb/universe_stats.go
```diff
@@ -401,39 +726,59 @@ func (u *UniverseStats) QuerySyncStats(ctx context.Context,
...
+			}
+
+			return sqlInt16(*q.AssetTypeFilter)
+		}(),
+		SortBy:        sqlStr(sortTypeToOrderBy(q.SortBy)),
+		SortDirection: sqlInt16(q.SortDirection),
+		NumOffset:     int32(q.Offset),
+		NumLimit: func() int32 {
+			if q.Limit == 0 {
+				return int32(math.MaxInt32)
```
jharveyb, line 766: Is this max because the stats pagination interprets limit differently from the latest pagination PR?

proof/proof.go
```diff
@@ -446,3 +446,29 @@ func (p *Proof) IsUnknownVersion() bool {
 		return true
 	}
 }
+
+// SparseDecode can be used to decode a proof from a reader without decoding
+// and parsing the entire thing. This handles ignoring the magic bytes, and
+// will decode directly into the target records.
+func SparseDecode(r io.Reader, records ...tlv.Record) error {
+	// The very first byte of the serialized proof is a magic byte, so
```
jharveyb, line 454: Why ignore magic bytes here? IIUC it costs us little to actually validate them here.


Or is it unneeded b/c internal use only.

tapgarden/caretaker.go
```diff
@@ -1150,6 +1152,11 @@ func (b *BatchCaretaker) storeMintingProof(ctx context.Context,
 		ScriptKey: &a.ScriptKey,
 	}
 
+	var proofBuf bytes.Buffer
+	if err = mintingProof.Encode(&proofBuf); err != nil {
```
jharveyb, line 1156: I think we have some lingering double encoding here and in the planter - on L1102 above we encode the same proof as a file with `EncodeAsProofFile`, which calls `Enccode` with the proof internally.

If we reworked `NewFile()` and `EncodeAsProofFile()` to accept already-encoded proofs we could save on one `Encode()` call. Can add to a follow-up PR.

tapdb/universe_test.go
```diff
@@ -562,13 +580,14 @@ func TestUniverseLeafQuery(t *testing.T) {
...
-				leaf.GroupKey.Witness =
-					sharedWitness
-				leaf.Proof.Asset.GroupKey.Witness =
-					sharedWitness
+				leaf.GroupKey.Witness = sharedWitness
+
+				//nolint:lll
+				//leaf.Proof.Asset.GroupKey.Witness = sharedWitness
+				leaf.Asset.GroupKey.Witness = sharedWitness
+				// TODO(roasbeef): circle back
```
jharveyb, line 590: What's the circle here, adding the group key to the proof? Confused why it doesn't fail as-is.

jharveyb (approved):
LGTM, big lift with all the new caches + rewiring DB calls + other misc fixes!

Excited to have this in, I think we have some smaller tasks to spin off re: generic caches and proof encode.

ffranr (approved):

tapdb/multiverse.go
Roasbeef: Fixed.

tapdb/multiverse.go
Roasbeef: So there's a limit on the total amount of active caches for a given universe (`treeID`), this is set to 50k rn. Then for each of chose, due to the pagination, we'll cache different pages (up to 1k total pages). I think we can tune these values further, once most clients are upgraded, that way the LRU will actually cache the most relevant items, rather than needing to cache it all.

tapdb/multiverse.go
Roasbeef: The mutex here is in place so that when the cache needs to be populated, only one goroutine does, and not like 10k.

tapdb/universe_stats.go
Roasbeef: > IIUC right now start and end times < 1 hr apart would return nothing.



Hmm, def need to tighten up with some tests. The intention here is that if you have start+end > 1 hr apart, then it's just the end+start that get truncated down. So `start=11:15`, and `end=2:19` goes to `start=11:00`, `end=2:00`.

tapdb/universe_stats.go
Roasbeef: Yeah this is kinda legacy. I think we want to enforce another limit here as well. For the stats site it relies on this not being populated, but a future deploy will start to end the limit itself.

proof/proof.go
Roasbeef: Later commit ends up validating it, might have been a fixup mixup.

tapgarden/caretaker.go
Roasbeef: Ah yeah I think that can indeed return the main proof there. I think ultimately it's worth it more to optimize the proof fetching vs minting here. But we can do another path to basically re-use the serialization more.

tapdb/universe_test.go
Roasbeef: Was a stray comment to remind myself to fix it after a rebase and tests started failing. The line above was for the old API where we stored the proof directly. The actual line just can access the asset in the leaf.

tapdb/multiverse.go
```diff
@@ -64,21 +73,236 @@ type BatchedMultiverse interface {
...
+	return &proofCache{}
+}
+
+// fetchProof reads the cached proof for the given ID and leaf key.
+func (p *proofCache) fetchProof(id universe.Identifier,
+	leafKey universe.LeafKey) []*universe.Proof {
+
+	// First, get the sub-cache for this universe ID from the map of
+	// caches.
+	idStr := treeID(id.String())
```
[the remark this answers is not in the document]
Roasbeef: It's just a type def to make the types a bit more readable. a `treeID` is just a string. We can't use the full `universe.Identifier` here as it has a pointer, so that can mess up map key comparison.

tapdb/multiverse.go
```diff
@@ -64,21 +73,236 @@ type BatchedMultiverse interface {
...
+	proofCache, _ := p.LoadOrStore(idStr, newLeafCache())
+	assetProofCache := proofCache.(leafProofCache)
+
+	// With that lower level cache obtained, we can check to see if we have
+	// a hit or not.
+	proofKey := NewProofKey(id, leafKey)
+	proof, err := assetProofCache.Get(proofKey)
```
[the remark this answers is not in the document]
Roasbeef: Agree, I also want to modify the cache itself to have methods that allow purging the entire thing. In lieu of that, atomics are used to allow hot swapping.

tapdb/multiverse.go
```diff
@@ -304,6 +561,9 @@ func (b *MultiverseStore) FetchProofLeaf(ctx context.Context,
 		return nil, dbErr
 	}
 
+	// Insert the proof we just read up into the main cache.
+	b.proofCache.insertProof(id, universeKey, proofs)
```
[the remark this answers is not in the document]
Roasbeef: Mentioned a bit above in one of the first review passes, but there's only ever a single proof. The API does support the query params, but these aren't ever exposed over RPC.

PR #628, opened 2023-10-26
universe stats: only select issuance roots for asset related stats

guggero:
This fixes a bug where the output of `tapcli universe stats assets` would show weird duplicates because the query would also include transfer related universe roots.
This also fixes the load test to work with the latest version of the `main` branch.

### DIFF
--- a/itest/loadtest/mint_batch_test.go
+++ b/itest/loadtest/mint_batch_test.go
@@ -127,7 +127,12 @@ func mintTest(t *testing.T, ctx context.Context, cfg *Config) {
 	// we asserted previously.
 	uniRoots, err := alice.AssetRoots(ctx, &unirpc.AssetRootRequest{})
 	require.NoError(t, err)
-	require.Len(t, uniRoots.UniverseRoots, groupCount)
+	issuanceRoots := fn.FilterMap(
+		uniRoots.UniverseRoots, func(root *unirpc.UniverseRoot) bool {
+			return root.Id.ProofType == unirpc.ProofType_PROOF_TYPE_ISSUANCE
+		},
+	)
+	require.Len(t, issuanceRoots, groupCount)
 
 	itest.AssertUniverseRoot(t, alice, groupBalance, nil, collectGroupKey)
 
--- a/tapdb/sqlc/migrations/000010_universe_stats.down.sql
+++ b/tapdb/sqlc/migrations/000010_universe_stats.down.sql
@@ -1,0 +1,2 @@
+-- This file is empty on purpose. There is nothing to roll back for this
+-- migration.
--- a/fn/func.go
+++ b/fn/func.go
@@ -45,6 +45,21 @@ func Filter[T any](s []T, f func(T) bool) []T {
 	return output
 }
 
+// FilterMap applies the given predicate function to each element of the given
+// map and generates a new slice containing only the elements for which the
+// predicate returned true.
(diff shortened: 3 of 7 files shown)


jharveyb (approved):
LGTM 🚀 

Would be good to add test coverage in a follow-on PR; rn in unit tests we default to inserting issuance proofs, so for `tapdb/universe_stats/QuerySyncStats()` we could add a transfer proof and assert that the returned stats don't change.

tapdb/sqlc/migrations/000010_universe_stats.down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file is empty on purpose. There is nothing to roll back for this
```
Roasbeef, line 1: So re stats for transfers vs assets, the intent was that we started to use the other enum type to differentiate syncing of transfer vs issuance proof: https://github.com/lightninglabs/taproot-assets/blob/81eb8d72e75dd567ba4e2ff3f15a1226a2d20fdc/universe/base.go#L535-L543

Roasbeef: We also store the proof type here as well: https://github.com/lightninglabs/taproot-assets/blob/main/tapdb/universe_stats.go#L143C31-L143C31

Roasbeef: We're sorta just not using the total "syncs" count anymore, but we can still obtain stats about the number of transfers.

itest/loadtest/mint_batch_test.go
```diff
@@ -127,7 +127,12 @@ func mintTest(t *testing.T, ctx context.Context, cfg *Config) {
 	// we asserted previously.
 	uniRoots, err := alice.AssetRoots(ctx, &unirpc.AssetRootRequest{})
 	require.NoError(t, err)
-	require.Len(t, uniRoots.UniverseRoots, groupCount)
+	issuanceRoots := fn.FilterMap(
+		uniRoots.UniverseRoots, func(root *unirpc.UniverseRoot) bool {
```
Roasbeef, line 131: Ahh I think this is what you were running into @calvinrzachman ?

guggero: Yes, exactly.

tapdb/sqlc/migrations/000010_universe_stats.down.sql
guggero: Hmm, not 100% sure what you mean. The proof type here https://github.com/lightninglabs/taproot-assets/blob/main/tapdb/universe_stats.go#L143C31-L143C31 refers to the [`roots.proof_type`](https://github.com/lightninglabs/taproot-assets/blob/81eb8d72/tapdb/sqlc/queries/universe.sql#L127) of the `InsertNewSyncEvent` query.

But you're saying we should start distinguishing between `NEW_PROOF` (for transfers?) and `NEW_ROOT` (for issuances)?

Does sound like a fix for another PR though. Also, inserting a row per event does look like a scaling issue. Maybe an upsert on a bucketized time stamp, so we'd only have one entry per 5 minutes and event type?

itest/loadtest/mint_batch_test.go
calvinrzachman: Yea, deployed to `staging` and this change seems to have fixed the issue! 🚀

Roasbeef (approved):
LGTM 🪥

Issue #618, opened 2023-10-24
universe: flip presence of global vs specific config 


Roasbeef:

dstadulis:
simple syncer takes a slice 
consider address handling (if you don't have an asset), if group_key no

if try add and we don't have it then sync with default universe

PR #631, opened 2023-10-26
universe: stop syncer from syncing all known universes by default

ffranr:
If a target set of universes is not specified the syncer will now sync universes which have associated specific sync configs, instead of all known universes.

Fixes https://github.com/lightninglabs/taproot-assets/issues/618

### DIFF
--- a/universe/syncer.go
+++ b/universe/syncer.go
@@ -74,62 +74,85 @@ func (s *SimpleSyncer) executeSync(ctx context.Context, diffEngine DiffEngine,
 		s.isSyncing.Store(false)
 	}()
 
+	// Examine config to ascertain whether global insertion for either proof
+	// type is allowed.
+	globalInsertEnabled := fn.Any(
+		syncConfigs.GlobalSyncConfigs,
+		func(config *FedGlobalSyncConfig) bool {
+			return config.AllowSyncInsert
+		},
+	)
+
+	// Define a custom filter function that will be used to filter for
+	// Universes that we want to sync. This filter makes use of the
+	// given syncType.
+	uniIdSyncFilter := func(id Identifier) bool {
+		// If we're syncing issuance proofs, then the universe must
+		// relate to issuance proofs.
+		if syncType == SyncIssuance &&
+			id.ProofType != ProofTypeIssuance {
+
+			return false
+		}
+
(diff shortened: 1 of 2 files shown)


universe/syncer.go
```diff
@@ -74,14 +74,39 @@ func (s *SimpleSyncer) executeSync(ctx context.Context, diffEngine DiffEngine,
...
+		idsToSync, func(id Identifier) bool {
+			// If we're syncing issuance proofs, then we'll only
+			// sync issuance roots.
+			if syncType == SyncIssuance &&
+				id.ProofType != ProofTypeIssuance {
+
+				return false
+			}
+
+			return syncConfigs.IsSyncInsertEnabled(id)
```
jharveyb, line 99: nit (for another PR): `IsSyncInsertEnabled()` loops over all configs we have; maybe those configs should be a `Set` instead of a list if we need to check membership often.

universe/syncer.go
```diff
@@ -74,14 +74,39 @@ func (s *SimpleSyncer) executeSync(ctx context.Context, diffEngine DiffEngine,
 		s.isSyncing.Store(false)
 	}()
 
+	// If we're not syncing a specific set of Universes, then we'll
+	// sync all Universes which have an associated specific sync config.
+	// These universes are filtered based on the actual config content
+	// later.
+	if len(idsToSync) == 0 {
```
jharveyb, line 81: I think for allowing opt-in global sync, we would still want the old logic that uses `RootNodes` (the old default case of the switch), but we could gate on having a global sync config with insert set to true.

Then switching default behavior from opt-out to opt-in should be just setting this flag to false?

https://github.com/lightninglabs/taproot-assets/blob/81eb8d72e75dd567ba4e2ff3f15a1226a2d20fdc/tapdb/universe_federation.go#L49

And we wouldn't need a new daemon flag.

ffranr: I think what you're suggesting will work. Is a user hasn't explicitly modified the global config in the db then they should still be using the default. And so modifying the default global config should affect most users when they upgrade.

guggero (approved):
Nice, LGTM :tada:

jharveyb (approved):
Small diff, nice! 💯

PR #633, opened 2023-10-27
address: optimistic asset sync

### LINKED ISSUE #620: addrs: attempt to sync from universe if asset ID unknown
As a we move away from the global sync config, we want to make UX more streamlined, by attempting to query for the main root before error'ing out to the user that we don't know of the asset ID.

jharveyb:
Fixes #620.

This PR adds support for creating addresses for assets that  a node has not bootstrapped or synced. If the asset info is not found, the node tries to sync asset info from all members of their universe federation. If the asset info is found, this asset is also added to the universe sync config to keep track of future possible issuance.

This helps address the UX changes from switching global sync to opt-in, since you can also make an address for an asset that's a member of an asset group, without having bootstrapped the asset group creation itself.

A related change is that the `QueryAssetRoots` RPC will now try and look up an asset group for a given universe ID, iff the ID specified an asset ID and no other roots were found. This is what allows a user to bootstrap an asset group without knowing the group key.

TODOs:

- [x] extend itest to cover a grouped asset and verify asset universe config changes + sync behavior with the config changes

### DIFF
--- a/address/book.go
+++ b/address/book.go
@@ -70,6 +70,19 @@ type QueryParams struct {
 	UnmanagedOnly bool
 }
 
+// AssetSyncer is an interface that allows the address.Book to look up asset
+// genesis and group information from both the local asset store and assets
+// known to universe servers in our federation.
+type AssetSyncer interface {
+	// SyncAssetInfo queries the universes in our federation for genesis
+	// and asset group information about the given asset ID.
+	SyncAssetInfo(ctx context.Context, assetID *asset.ID) error
+
+	// EnableAssetSync updates the sync config for the given asset so that
+	// we sync future issuance proofs.
+	EnableAssetSync(ctx context.Context, groupInfo *asset.AssetGroup) error
+}
+
 // Storage is the main storage interface for the address book.
 type Storage interface {
 	EventStorage
@@ -131,6 +144,10 @@ type KeyRing interface {
 type BookConfig struct {
 	// Store holds the set of created addresses.
 	Store Storage
+
+	// Syncer allows the address.Book to sync issuance information for
+	// assets from universe servers in our federation.
+	Syncer AssetSyncer
 
 	// KeyRing points to an active key ring instance.
 	KeyRing KeyRing
--- a/address/book.go
+++ b/address/book.go
@@ -2,12 +2,14 @@ package address
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/url"
 	"sync"
 	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/taproot-assets/asset"
@@ -188,6 +190,64 @@ func NewBook(cfg BookConfig) *Book {
 	}
 }
 
+// queryAssetInfo attempts to locate asset genesis information by querying
+// geneses already known to this node. If asset issuance was not previously
+// verified, we then query universes in our federation for issuance proofs.
+func (b *Book) queryAssetInfo(ctx context.Context,
+	id asset.ID) (*asset.AssetGroup, error) {
+
+	// Check if we know of this asset ID already.
+	assetGroup, err := b.cfg.Store.QueryAssetGroup(ctx, id)
+	switch {
+	case assetGroup != nil:
+		return assetGroup, nil
+
+	// Asset lookup failed gracefully; continue to asset lookup using the
+	// AssetSyncer if enabled.
+	case errors.Is(err, ErrAssetGroupUnknown):
+		if b.cfg.Syncer == nil {
+			return nil, ErrAssetGroupUnknown
+		}
+
+	case err != nil:
+		return nil, err
+	}
+
+	log.Debugf("asset %v is unknown, attempting to bootstrap", id.String())
+
+	// Use the AssetSyncer to query our universe federation for the asset.
+	err = b.cfg.Syncer.SyncAssetInfo(ctx, &id)
+	if err != nil {
+		return nil, err
+	}
+
+	// The asset genesis info may have been synced from a universe
+	// server; query for the asset ID again.
+	assetGroup, err = b.cfg.Store.QueryAssetGroup(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Debugf("bootstrap succeeded for asset %v", id.String())
+
+	// If the asset was found after sync, and has an asset group, update our
+	// universe sync config to ensure that we sync future issuance proofs.
+	// Ungrouped assets will have no new issuance proofs so do not need a
+	// universe sync config at all.
+	if assetGroup.GroupKey != nil {
+		log.Debugf("enabling asset sync for asset group %x",
+			schnorr.SerializePubKey(
+				&assetGroup.GroupKey.GroupPubKey,
+			))
+		err = b.cfg.Syncer.EnableAssetSync(ctx, assetGroup)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return assetGroup, nil
+}
+
 // NewAddress creates a new Taproot Asset address based on the input parameters.
 func (b *Book) NewAddress(ctx context.Context, assetID asset.ID, amount uint64,
 	tapscriptSibling *commitment.TapscriptPreimage,
@@ -195,8 +255,8 @@ func (b *Book) NewAddress(ctx context.Context, assetID asset.ID, amount uint64,
 ) (*AddrWithKeyInfo, error) {
 
 	// Before we proceed and make new keys, make sure that we actually know
-	// of this asset ID already.
-	if _, err := b.cfg.Store.QueryAssetGroup(ctx, assetID); err != nil {
+	// of this asset ID, or can import it.
+	if _, err := b.queryAssetInfo(ctx, assetID); err != nil {
 		return nil, fmt.Errorf("unable to make address for unknown "+
 			"asset %x: %w", assetID[:], err)
 	}
@@ -234,7 +294,7 @@ func (b *Book) NewAddressWithKeys(ctx context.Context, assetID asset.ID,
 	// Before we proceed, we'll make sure that the asset group is known to
 	// the local store. Otherwise, we can't make an address as we haven't
 	// bootstrapped it.
-	assetGroup, err := b.cfg.Store.QueryAssetGroup(ctx, assetID)
+	assetGroup, err := b.queryAssetInfo(ctx, assetID)
 	if err != nil {
 		return nil, err
 	}
--- a/itest/addrs_test.go
+++ b/itest/addrs_test.go
@@ -4,11 +4,17 @@ import (
 	"bytes"
 	"context"
 
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	tap "github.com/lightninglabs/taproot-assets"
+	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/internal/test"
 	"github.com/lightninglabs/taproot-assets/tappsbt"
 	"github.com/lightninglabs/taproot-assets/taprpc"
 	wrpc "github.com/lightninglabs/taproot-assets/taprpc/assetwalletrpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/tapdevrpc"
+	unirpc "github.com/lightninglabs/taproot-assets/taprpc/universerpc"
+	"github.com/lightninglabs/taproot-assets/universe"
 	"github.com/lightningnetwork/lnd/lntest/wait"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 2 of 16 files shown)


rpcserver.go
```diff
@@ -2993,6 +2995,109 @@ func (r *rpcServer) QueryAssetRoots(ctx context.Context,
...
+		foundGroupKey  *btcec.PublicKey
+	)
+
+	// Attempt to look up a matching asset group.
+	if mayBeGrouped {
+		groupedAssetID = universeID.AssetID
+		rpcsLog.Debugf("No roots found for asset %v, checking if this "+
+			"asset is part of a group", groupedAssetID.String())
+
+		assetGroup, err := r.cfg.TapAddrBook.QueryAssetGroup(
```
Roasbeef, line 3016: Doing it at this level means another round trip (actually 3x) and set of DB queries each time this path is exercised. Alternatively, we can do this from within the main DB query, as we have the asset group view available. Doing it at that lower level woudl also mean we can test this more effectively, which is something we generally want to improve on now that we have the initial alpha release out.

itest/universe_test.go
```diff
@@ -497,6 +498,52 @@ func testUniverseFederation(t *harnessTest) {
...
+	uniIDNoGroupKey := &unirpc.ID{
+		Id: &unirpc.ID_AssetId{
+			AssetId: groupedAsset[0].AssetGenesis.AssetId,
+		},
+	}
+	var groupUniRoots *unirpc.QueryRootResponse
+
+	waitErr := wait.NoError(func() error {
+		var err error
+		groupUniRoots, err = t.tapd.QueryAssetRoots(
```
Roasbeef, line 518: Shouldn't this succeed on the very first try? Why do we need `wait.NoError` wrapped around the call.

address/book.go
```diff
@@ -70,6 +70,20 @@ type QueryParams struct {
 	UnmanagedOnly bool
 }
 
+// AssetSyncer is an interface that allows the address.Book to look up asset
+// genesis and group information from both the local asset store and assets
+// known to universe servers in our federation.
+type AssetSyncer interface {
+	// FetchAssetInfo queries the universes in our federation for genesis
+	// and asset group information about the given asset ID.
+	FetchAssetInfo(ctx context.Context, assetID *asset.ID) error
```
Roasbeef, line 80: IIUC, we just want to attempt to sync it we get an error that we don't know of the asset ID on disk. So we don't need to check to see if we know of it, if we just try and then detect the error when we don't.

address/book.go
```diff
@@ -70,6 +70,20 @@ type QueryParams struct {
...
+
+	// UpdateSyncConfig updates the sync config for the given asset so that
+	// we sync future issuance proofs.
+	UpdateSyncConfig(ctx context.Context, assetID *asset.ID,
```
Roasbeef, line 84: Can use `AssetGroup` here. 

I think we also only care about updating the sync config if the asset itself has a group.

tapcfg/server.go
```diff
@@ -301,6 +294,16 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 		},
 	)
 
+	// TODO(jhb): expose `SyncerActive` setting elsewhere
+	addrBook := address.NewBook(address.BookConfig{
+		Store:        tapdbAddrBook,
+		Syncer:       universeFederation,
+		SyncerActive: true,
```
Roasbeef, line 301: Why's this needed? Can't we just say if the passed interface is nil or not?

universe/auto_syncer.go
```diff
@@ -228,27 +235,16 @@ func (f *FederationEnvoy) syncServerState(ctx context.Context,
...
-	ctx, cancel := f.WithCtxQuit()
-	defer cancel()
-
-	fedServers, err := f.cfg.FederationDB.UniverseServers(
-		ctx,
-	)
-	if err != nil {
-		err := fmt.Errorf("unable to fetch set of universe "+
-			"servers: %v", err)
-		f.reportErr(err)
```
Roasbeef, line 240: I have a commit that removes this report, otherwise it causes the daemon to shutdown.

universe/auto_syncer.go
```diff
@@ -531,6 +519,147 @@ func (f *FederationEnvoy) SetAllowPublicAccess() error {
...
+				return nil
+			}
+
+			returnedSyncDiffs <- syncDiff[0]
+		}
+
+		return nil
+	}
+
+	// Sync with the federation Universe servers in parallel.
```
Roasbeef, line 609: Why do we need to sync from all the severs in parallel, when we only want a single asset ID here? 

Instead I think we can just select one (at random?), then try to sync with just that instance.

universe/auto_syncer.go
```diff
@@ -531,6 +519,147 @@ func (f *FederationEnvoy) SetAllowPublicAccess() error {
...
+		return fmt.Errorf("asset lookup failed for asset: %v",
+			assetID.String())
+	}
+
+	return nil
+}
+
+// UpdateSyncConfig updates the sync config for the given asset to that we sync
+// future issuance proofs.
+func (f *FederationEnvoy) UpdateSyncConfig(ctx context.Context,
```
Roasbeef, line 633: `EnableAssetSync`? or `EnableSyncForAsset`?

address/book.go
```diff
@@ -193,15 +194,70 @@ func NewBook(cfg BookConfig) *Book {
...
+	// Check if we know of this asset ID already.
+	assetGroup, err := b.cfg.Store.QueryAssetGroup(ctx, id)
+	switch {
+	case assetGroup != nil:
+		return assetGroup, nil
+
+	// Asset lookup failed gracefully; continue to asset lookup using the
+	// AssetSyncer if enabled.
+	case errors.Is(err, ErrAssetGroupUnknown):
+		if !b.cfg.SyncerActive {
```
Roasbeef, line 212: Why would this ever be false in practice?

address/book.go
```diff
@@ -193,15 +194,70 @@ func NewBook(cfg BookConfig) *Book {
...
+		}
+
+	case err != nil:
+		return nil, err
+	}
+
+	log.Debugf("asset %v is unknown, attempting to bootstrap", id.String())
+
+	// Use the AssetSyncer to query our universe federation for the asset.
+	err = b.cfg.Syncer.FetchAssetInfo(ctx, &id)
```
Roasbeef, line 223: Ah I see here we use sync vs fetch somewhat interchangeably. Perhaps use a `Sync` prefix here since we're reaching out to a remove server to try to obtain the information?

address/book.go
```diff
@@ -193,15 +194,70 @@ func NewBook(cfg BookConfig) *Book {
...
+	log.Debugf("bootstrap succeeded for asset %v", id.String())
+
+	// If the asset was found after sync, update our universe sync config
+	// to ensure that we sync future issuance proofs.
+	var groupKey *btcec.PublicKey
+	if assetGroup.GroupKey != nil {
+		groupKey = &assetGroup.GroupKey.GroupPubKey
+	}
+
+	err = b.cfg.Syncer.UpdateSyncConfig(ctx, &id, groupKey)
```
Roasbeef, line 244: I think we can get away with just a single method here that'll try to sync, and if it's able to, it'll add to the config. That way the addr book needs to know less information about syncing/configs, etc.

Roasbeef (changes requested):
Nice work! IMO, this is the syncing model we should've gone with the first time around: only sync stuff on demand when you actually need it. 

Dropped to main comments: 
  1. Looking into moving the query fall back logic deeper into the DB. This can help us cut down on extra DB transactions. 
  2. Only syncing from one server at a time, and simplifying the new duties of the addr book a bit.

address/book.go
jharveyb: Resolved later on? That's what we do with `queryAssetInfo`, intent was to keep `QueryAssetGroup` calls separate from those backend by the universe interfaces, and leave room to turn off this JIT bootstrapping. Also I think sync would fail if we minted the asset and never pushed issuance proofs to any universe in our federation.

address/book.go
jharveyb: Yeah, I guess an ungrouped asset only ever needs to be synced once and therefore never needs a universe config option? For issuance proofs at least.

universe/auto_syncer.go
jharveyb: Intent was that you get the 'latest' asset state, e.x. for an asset group one server may only have record of 5 issuances while the official issuer server has a record of 10.

Also just to mirror what we do for the background sync. I think random single sync may not work well if we expect users to have universes with non-overlapping assets in their federation (one host wallet provider, one explorer, one stablecoin issuer, etc.).

address/book.go
jharveyb: Intended as user config option; setting syncer to `nil` on startup would also work.

address/book.go
jharveyb: Was thinking of that initially as well, but it's a bit awkward without adding group lookup to the various universe and federation universes (which I figured we didn't want to do).

From the context of `FetchAssetInfo`, when a sync succeeds, we don't know which proof is for the group anchor, and we would need to use that proof to return results equivalent to what `QueryAssetGroup` returns.

address/book.go
```diff
@@ -193,15 +194,70 @@ func NewBook(cfg BookConfig) *Book {
...
+
+	// Use the AssetSyncer to query our universe federation for the asset.
+	err = b.cfg.Syncer.FetchAssetInfo(ctx, &id)
+	if err != nil {
+		return nil, err
+	}
+
+	// The asset genesis info may have been synced from a universe
+	// server; query for the asset ID again.
+	assetGroup, err = b.cfg.Store.QueryAssetGroup(ctx, id)
```
GeorgeTsagk, line 230: Could we have `Syncer.FetchAssetInfo` also return some results above? Could save the extra call & store roundtrip

tapgarden/custodian_test.go
```diff
@@ -266,6 +300,80 @@ func TestCustodianNewAddr(t *testing.T) {
...
+	newAddr, err := h.addrBook.NewAddress(
+		ctx, newAsset.ID(), 1, nil, proofCourierAddr,
+	)
+	require.NoError(t, err)
+	require.NotNil(t, newAddr)
+
+	numAddrs := 1
+	validAddrs := fn.MakeSlice(newAddr)
+
+	// TODO(jhb): remove asset from syncer, assert that address creation
```
GeorgeTsagk, line 349: what is the expected effect of removing an asset from the syncer?

itest/addrs_test.go
```diff
@@ -184,6 +187,151 @@ func testMultiAddress(t *harnessTest) {
...
+		},
+	)
+	require.NoError(t.t, err)
+
+	// Bob's Universe stats should show that he now has two assets.
+	AssertUniverseStats(t.t, bob, 2, 0, 2, 1)
+
+	// Bob should not be able to make an address for a random asset ID
+	// that he nor Alice are aware of.
+
```
GeorgeTsagk, line 245: nit: extra newline

itest/addrs_test.go
```diff
@@ -184,6 +187,151 @@ func testMultiAddress(t *harnessTest) {
...
+	syncDiff, err := bob.SyncUniverse(ctxt, &unirpc.SyncRequest{
+		UniverseHost: t.tapd.rpcHost(),
+		SyncMode:     unirpc.UniverseSyncMode_SYNC_ISSUANCE_ONLY,
+	})
+	require.NoError(t.t, err)
+	require.Len(t.t, syncDiff.SyncedUniverses, 0)
+
+	// Bob should be able to make an address for both new assets minted by
+	// Alice, even though he has not synced their issuance proofs.
+
```
GeorgeTsagk, line 321: nit: extra nl

GeorgeTsagk (commented):
nice PR! had the chance to dive into a lot of code while reading, JIT syncing looks promising

tapgarden/custodian_test.go
jharveyb: Here, the `address.Book` should still succeed to make an address, as `QueryAssetGroup` should find the asset genesis (since the address should have been imported after the `AddAsset` call above.

address/book.go
jharveyb: Yes, but contingent on doing the group lookup on the multiverse side / when looking up the root node itself.

rpcserver.go
Roasbeef: Discussed offline that we'll defer the optimization to a future PR, but clean up the call site a bit here with some helper functions.

tapcfg/server.go
Roasbeef: Ah sure agree re being able to opt out of it, in which case, I think the server would just not set `Syncer` at all. Then if that's nil, all the new behavior is skipped.

address/book.go
Roasbeef: Gotcha, I think `nil` here matches how we manage optional interface impls elsewhere.

address/book.go
Roasbeef: Not sure I follow: why does it matter which proof gets returned here? IIUC, we just care that we have the genesis information for the asset ID in question. Assuming sync has completed, we should have the genesis info on disk, so we repeat the query that failed to get the info we need.

Roasbeef:
Removed from draft so the itests can run as well.

address/book.go
jharveyb: For a group we also need the group key and the witness from the group anchor to match `QueryAssetGroup`. So those would only be in the group anchor proof.

Or do you mean perform the lookup again in `EnableAssetSync`?

universe/auto_syncer.go
```diff
@@ -531,6 +518,137 @@ func (f *FederationEnvoy) SetAllowPublicAccess() error {
...
+	// Sync with the federation Universe servers in parallel.
+	err = fn.ParSlice(ctx, fedServers, syncFromUni)
+	if err != nil {
+		// We should never receive a non-nil error from the sync above.
+		log.Errorf("unable to perform asset lookup with federation: "+
+			"%v", err)
+		return err
+	}
+
+	syncDiffs := fn.Collect(returnedSyncDiffs)
```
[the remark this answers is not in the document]
jharveyb: Yeah I think once the caller syncs the full state from any federation syncer, the other calls would have another path to exit early from the local and remote root matching. But also not sure.

address/book.go
jharveyb: And that works without this being `*AssetSyncer`?

rpcserver.go
```diff
@@ -2993,6 +2995,109 @@ func (r *rpcServer) QueryAssetRoots(ctx context.Context,
...
+				groupedAssetID.String())
+
+		// The asset is not a grouped asset; we will return empty
+		// universe roots below.
+		default:
+		}
+	}
+
+	// If an asset group was found, fetch the roots for that group.
+	if foundGroupKey != nil {
```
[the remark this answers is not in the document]
jharveyb: Yup considered that initially but I think current version simplifies the err handling a bit, it works out.

jharveyb:
Reworked the config for disabling the asset syncer. itest extended to test disabling the asset syncer + asserts that the config update succeeds + tests the syncer with a grouped asset. Also refactored the itest harness a bit to make adding future args for test setup a bit easier.

rpcserver.go
```diff
@@ -2974,21 +3046,21 @@ func (r *rpcServer) QueryAssetRoots(ctx context.Context,
...
 		// above.
-		if !errors.Is(err, universe.ErrNoUniverseRoot) {
-			return nil, err
+		if !errors.Is(transferErr, universe.ErrNoUniverseRoot) {
+			return nil, transferErr
 		}
 	}
 
-	transferRootRPC, err := marshalUniverseRoot(transferRoot)
+	transferRootRPC, err = marshalUniverseRoot(transferRoot)
```
Roasbeef, line 3063: Won't this panic if nothing was found above?

rpcserver.go
```diff
@@ -2957,13 +2957,85 @@ func (r *rpcServer) QueryAssetRoots(ctx context.Context,
 			"given universe")
 	}
 
-	issuanceRoot, err := r.cfg.UniverseArchive.RootNode(ctx, universeID)
+	// Query for both a issaunce and transfer universe root.
+	assetRoots, err := r.queryAssetProofRoots(ctx, universeID)
 	if err != nil {
+		return nil, err
+	}
+
```
Roasbeef, line 2965: If we have an actual root here, then shouldn't we just return early rather than relying on the default behavior below?

universe/auto_syncer.go
```diff
@@ -531,6 +518,134 @@ func (f *FederationEnvoy) SetAllowPublicAccess() error {
...
+
+	fedServers, err := f.cfg.FederationDB.UniverseServers(
+		ctx,
+	)
+	if err != nil {
+		cancel()
+
+		err := fmt.Errorf("unable to fetch set of universe servers: %v",
+			err)
+		f.reportErr(err)
```
Roasbeef, line 534: We should remove this (depends on order of which PRs land first), as is it'll make the entire daemon shutdown if we can't read the set of servers for w/e reason. So we should just log a `Warnf` here.

tapcfg/config.go
```diff
@@ -253,6 +253,12 @@ type UniverseConfig struct {
 	PublicAccess bool `long:"public-access" description:"If true, and the Universe server is on a public interface, valid proof from remote parties will be accepted, and proofs will be queryable by remote parties. This applies to federation syncing as well as RPC insert and query."`
 }
 
+// AddressConfig is the config that houses any address Book related config
+// values.
+type AddrBookConfig struct {
+	DisableSyncer bool `long:"disable-syncer" description:"If true, tapd will not try to sync issuance proofs for unknown assets when creating an address."`
```
Roasbeef, line 259: 👍

itest/tapd_harness.go
```diff
@@ -80,12 +80,28 @@ type tapdConfig struct {
...
 }
 
+type harnessOpts struct {
+	proofSendBackoffCfg     *proof.BackoffCfg
+	proofReceiverAckTimeout *time.Duration
+	proofCourier            proof.CourierHarness
+	addrAssetSyncerDisable  bool
+}
+
+type harnessOption func(*harnessOpts)
```
Roasbeef, line 90: 👍

Roasbeef (approved):
LGTM 🎡

Just one lingering questions about early return and if we want to return an error on unknown root, or just serialize an empty proto.

rpcserver.go
jharveyb: Hmm yeah I think the `RootNode` call above would err,  return a real root, or `Root{}`. But `marshalUniverseRoot has an empty nil-ness check that exits early on `Root{}`:

https://github.com/lightninglabs/taproot-assets/blob/e569cf553d324297c303d190678320fcc50b9c10/rpcserver.go#L2751

 And we already hit this as the unknown assets would have no transfer root.

rpcserver.go
jharveyb: We can; in practice, `mayBeGrouped` would be false, so we skip the group lookup.

Looked it and early exit does reduce indention though, so switched to that.

universe/auto_syncer.go
jharveyb: K will mirror what you have up.

jharveyb:
Looks like new flake on Postgres?

dstadulis:
itests: if you modify the global state of [something] it'll cause dysfunction.  FIxes are known and will arrive today

jharveyb:
Ready for final 2nd review.

itest/addrs_test.go
```diff
@@ -184,6 +190,243 @@ func testMultiAddress(t *harnessTest) {
...
+	_, err := bob.NewAddr(ctxt, &taprpc.NewAddrRequest{
+		AssetId:      firstAsset.AssetGenesis.AssetId,
+		Amt:          firstAsset.Amount / 2,
+		AssetVersion: firstAsset.Version,
+	})
+	require.ErrorContains(t.t, err, "asset lookup failed for asset")
+
+	// We'll now add the main node, as a member of Bob's Universe
+	// federation. We expect that their state is synchronized shortly after
+	// the call returns.
```
GeorgeTsagk, line 229: nit: We haven't had any flakes here, but should we wait with `Eventualy` to ensure nodes are sync'd instead?

itest/addrs_test.go
```diff
@@ -184,6 +190,243 @@ func testMultiAddress(t *harnessTest) {
...
+		AssetId:      firstAsset.AssetGenesis.AssetId,
+		Amt:          firstAsset.Amount,
+		AssetVersion: firstAsset.Version,
+	})
+	require.ErrorContains(t.t, err, "asset group is unknown")
+
+	// Restart Bob again with the syncer enabled. Bob should be able to make
+	// an address for both new assets minted by Alice, even though he has
+	// not synced their issuance proofs.
+	restartBobNoUniSync(false)
```
GeorgeTsagk, line 365: :+1: 
Last note w.r.t being able to enable/disable JIT sync: Is the "global sync" (fetch all) and JIT sync going to live side-by-side in the future? Feels to me that if JIT is disabled then global sync must be present, and if JIT is enabled global has to be off, as these two sort of overlap

GeorgeTsagk: does that lead us to 2 future modes for sync?

GeorgeTsagk (approved):
LGTM :fire: big step for smart sync :eyes:

itest/addrs_test.go
jharveyb: Good Q: So if global sync is on, the syncer setting sort of doesn't matter; the only way to not know about an asset would be if it was pushed to a federation member after your last sync (up to 10 mins ago).

The default with this PR and the others in main is "JIT sync on, global sync off". With manual global sync still possible.

With JIT on anything you sync from a global sync, you haven't interacted with yet, so I think that's a reasonable default.

PR #634, opened 2023-10-30
Add pagination to universerpc calls

### LINKED ISSUE #590: rpc: add pagination support to all relevant RPC APIs
Today we have pagination support for some calls like the `QueryAssetStats` call. We should also add them where relevant to the set of other calls that can return lists of data. 

One example is the `QueryAssetRoots` and `AssetLeafKeys` call. Eventually both of these will exceed the default gRPC message limit, so we need to add pagination support here to ensure syncing, etc can still work.

GeorgeTsagk:
## Description

This PR enables pagination for some of the universe rpc calls, specifically `AssetRoots` and `AssetLeafKeys`.

Closes #590


rpcserver.go
```diff
@@ -3058,17 +3063,24 @@ func marshalLeafKey(leafKey universe.LeafKey) *unirpc.AssetKey {
...
+	universeID, err := UnmarshalUniID(req.Id)
 	if err != nil {
 		return nil, err
 	}
 
 	// TODO(roasbeef): tell above if was tring or not, then would set
 	// below diff
 
-	leafKeys, err := r.cfg.UniverseArchive.UniverseLeafKeys(ctx, universeID)
+	leafKeys, err := r.cfg.UniverseArchive.UniverseLeafKeys(
```
jharveyb, line 3076: I think requests with a limit over the maximum page size (which we can define in `universe/interface` or elsewhere) should be rejected here at the RPC layer. Should also assert that offset is non-negative.

tapdb/multiverse.go
```diff
@@ -31,7 +32,8 @@ type (
 type BaseMultiverseStore interface {
 	BaseUniverseStore
 
-	UniverseRoots(ctx context.Context) ([]BaseUniverseRoot, error)
+	UniverseRoots(ctx context.Context,
```
jharveyb, line 35: I think `sqlc.UniverseRootsParams` should get a type alias, like `BaseUniverseRoot` above.

tapdb/universe.go
```diff
@@ -612,13 +615,27 @@ func universeFetchProofLeaf(ctx context.Context,
...
 	readTx := NewBaseUniverseReadTx()
 	dbErr := b.db.ExecTx(ctx, &readTx, func(db BaseUniverseStore) error {
-		universeKeys, err := db.FetchUniverseKeys(ctx, b.smtNamespace)
+		universeKeys, err := db.FetchUniverseKeys(
+			ctx, UniverseLeafKeysQuery{
+				Namespace:     b.smtNamespace,
+				SortDirection: sqlInt16(direction),
+				NumOffset:     int32(offset),
+				NumLimit: func() int32 {
+					if limit == 0 {
```
jharveyb, line 631: Discussed offline that this should default to the maximum page size, not `maxint32`. so an old client gets only the first page, not as much data as possible.

universe/base.go
```diff
@@ -127,14 +127,22 @@ func (a *Archive) RootNode(ctx context.Context,
 	})
 }
 
+type RootNodesQuery struct {
```
jharveyb, line 130: These types should get added a commit earlier? IIUC they are used in commit 2 but added in commit 3.

Also, since both the RPC and the DB are using `int32` for these fields, the type here could be `int32` as well and eliminate some conversions elsewhere.

universe/syncer.go
```diff
@@ -13,6 +13,12 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const (
+	// defaultPageLimit is the default page limit used when fetching
+	// multiple pages of data from the remote Universe.
+	defaultPageLimit = 512
```
jharveyb, line 19: 👍🏽

universe/syncer.go
```diff
@@ -235,15 +263,55 @@ func (s *SimpleSyncer) syncRoot(ctx context.Context, remoteRoot Root,
 	log.Infof("UniverseRoot(%v) diverges, performing leaf diff...",
 		uniID.String())
 
-	// Otherwise, we'll need to perform a diff operation to find the set of
-	// keys we need to fetch.
-	remoteUniKeys, err := diffEngine.UniverseLeafKeys(ctx, uniID)
-	if err != nil {
-		return err
+	var (
```
jharveyb, line 266: Restore comment explaining the diff operation?

tapdb/universe.go
```diff
@@ -38,6 +39,8 @@ type (
 
 	// UniverseLeaf is a universe leaf.
 	UniverseLeaf = sqlc.QueryUniverseLeavesRow
+
```
jharveyb, line 42: nit: missing godoc

jharveyb (commented):
Looking pretty close!

`fetchRootsForIDs` also need to be updated to use pagination now:

https://github.com/lightninglabs/taproot-assets/blob/e569cf553d324297c303d190678320fcc50b9c10/universe/syncer.go#L179

rpcserver.go
```diff
@@ -3058,17 +3063,24 @@ func marshalLeafKey(leafKey universe.LeafKey) *unirpc.AssetKey {
 // Taproot Asset commitment, and script_key is the script_key of the asset
 // within the Taproot Asset commitment for the given asset_id or group_key.
 func (r *rpcServer) AssetLeafKeys(ctx context.Context,
-	req *unirpc.ID) (*unirpc.AssetLeafKeyResponse, error) {
+	req *unirpc.AssetLeafKeysRequest) (*unirpc.AssetLeafKeyResponse, error) {
 
-	universeID, err := UnmarshalUniID(req)
+	universeID, err := UnmarshalUniID(req.Id)
```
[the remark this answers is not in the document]
jharveyb: Yeah there's a tracking issue for that, I think for some it's more acceptable rn since other funcs called in the RPC handler will deal with it, or we don't access struct members directly.

universe/base.go
GeorgeTsagk: > These types should get added a commit earlier? IIUC they are used in commit 2 but added in commit 3.

My bad, gonna rebase the commits

tapdb/multiverse.go
```diff
@@ -133,15 +136,27 @@ func (b *MultiverseStore) RootNode(ctx context.Context,
...
 		uniRoots []universe.Root
 		readTx   = NewBaseMultiverseReadTx()
 	)
 
+	params := sqlc.UniverseRootsParams{
+		SortDirection: sqlInt16(q.SortDirection),
+		NumOffset:     int32(q.Offset),
+		NumLimit: func() int32 {
+			if q.Limit == 0 {
+				return universe.MaxPageSize
```
Roasbeef, line 151: 👍 

We should prob scan elsewhere to implement a default max page size instead of going wumbo by default.

tapdb/universe.go
```diff
@@ -612,13 +616,27 @@ func universeFetchProofLeaf(ctx context.Context,
 
 // MintingKeys returns all the keys inserted in the universe.
 func (b *BaseUniverseTree) MintingKeys(ctx context.Context,
+	direction universe.SortDirection, offset, limit int32,
```
Roasbeef, line 619: Pull out into a `QueryParam` struct like we do elsewhere?

cmd/tapcli/universe.go
```diff
@@ -18,6 +18,7 @@ import (
 const (
 	proofTypeName       = "proof_type"
 	skipAmountsByIdName = "skip_amounts_by_id"
+	defaultPageSize     = 512
```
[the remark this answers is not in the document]
Roasbeef: We can use the constant directly from the `universe` package.

universe/syncer.go
```diff
@@ -116,9 +122,31 @@ func (s *SimpleSyncer) executeSync(ctx context.Context, diffEngine DiffEngine,
 	// remote servers.
 	case globalInsertEnabled:
 		log.Infof("Fetching all roots for remote Universe server...")
-		targetRoots, err = diffEngine.RootNodes(ctx, false)
-		if err != nil {
-			return nil, err
+
+		offset := int32(0)
+
+		for {
```
Roasbeef, line 128: Ah cool, didn't realize we went ahead to also update the syncers in this version. We should make sure that once we extend past the page size, this logic works properly. The load tests can be a good sanity check, but eventually I think we'll want to add some unit tests here (as the`diffEngine`) is already abstracted to ensure proper termination.

universe/syncer.go
```diff
@@ -116,9 +122,31 @@ func (s *SimpleSyncer) executeSync(ctx context.Context, diffEngine DiffEngine,
...
+			log.Debugf("Fetching roots in range: %v to %v", offset,
+				offset+defaultPageLimit)
+			tempRoots, err := diffEngine.RootNodes(
```
Roasbeef, line 131: I think we can extract this into a `fetchAllRoots` function, which also sets us up to be able to add more unit tests in an isolated fashion later.

universe/syncer.go
```diff
@@ -236,14 +264,59 @@ func (s *SimpleSyncer) syncRoot(ctx context.Context, remoteRoot Root,
...
 	}
-	localUniKeys, err := s.cfg.LocalDiffEngine.UniverseLeafKeys(ctx, uniID)
-	if err != nil {
-		return err
+
+	// Initialize the offset to be used for the pages.
+	offset = 0
+
+	for {
+		tempLocalKeys, err := s.cfg.LocalDiffEngine.UniverseLeafKeys(
```
Roasbeef, line 302: Same here re extracting into a `fetchAllLeafkeys` function. Can eliminate the duplication above.

Roasbeef (changes requested):
Nice work! Just a few comments re some refactoring we can do to make the new pagination awareness logic in the syncer easier to unit tests in the future.

universe/syncer.go
Roasbeef: Yeah, as is with the itests rn, we'll only have a single page of assets worth. If we extract this out into a diff function, then we can write tests for the pagination/termination logic, and we should also eventually add itests and unit tests at the DB level for the behavior as well.

dstadulis:
@GeorgeTsagk is attempting to make the batch mint itests with pagination being enabled
But the batch mint itests is also failing on main branch
The linter may not be running on this itests so there could be some rot

tapdb/multiverse.go
dstadulis: @GeorgeTsagk Can you summarize what information you'll need from @Roasbeef to get clarity?

GeorgeTsagk: > We should prob scan elsewhere to implement a default max page size instead of going wumbo by default.

Are you referring to the iterative page fetching process as scanning? If yes that's done on the client side, and it default to the max page, which is "wumbo" but actually reasonably small (512 items)

GeorgeTsagk:
Pushed an itest that covers pagination

itest/universe_pagination_test.go
```diff
@@ -0,0 +1,246 @@
...
+	imageMetadataBytes := GetImageMetadataBytes(t.t, ImageMetadataFileName)
+
+	mintBatchAssetsTest(
+		t.t, t.tapd, bob, t.tapd.rpcHost(), mintSize, mintBatches,
+		imageMetadataBytes, timeout,
+	)
+}
+
+// mintBatchAssetsTest mints many assets
+func mintBatchAssetsTest(
```
Roasbeef, line 61: IIUC this was lifted from another test, so no need to fully copy/paste, may need some intermediate refactor.

Roasbeef (approved):
LGTM 💽

tapdb/multiverse.go
Roasbeef: I mean other places for the pagination calls (existing ones) where we don't have a default max page size. I think non-blocking for this PR tho, can be done in another pass.

itest/universe_pagination_test.go
GeorgeTsagk: the func body is changed so that multiple groups are created, this func creates multiple batches

tapdb/multiverse.go
```diff
@@ -133,15 +136,27 @@ func (b *MultiverseStore) RootNode(ctx context.Context,
...
+	q universe.RootNodesQuery) ([]universe.Root, error) {
 
 	var (
 		uniRoots []universe.Root
 		readTx   = NewBaseMultiverseReadTx()
 	)
 
+	params := sqlc.UniverseRootsParams{
+		SortDirection: sqlInt16(q.SortDirection),
+		NumOffset:     int32(q.Offset),
```
jharveyb, line 148: nit: unneeded int32 casts.

universe_rpc_diff.go
```diff
@@ -119,13 +122,20 @@ func (r *RpcUniverseDiff) RootNode(ctx context.Context,
...
+	uniID, err := MarshalUniID(q.Id)
 	if err != nil {
 		return nil, err
 	}
-	assetKeys, err := r.conn.AssetLeafKeys(ctx, uniID)
+	assetKeys, err := r.conn.AssetLeafKeys(
+		ctx, &unirpc.AssetLeafKeysRequest{
+			Id:        uniID,
+			Direction: unirpc.SortDirection(q.SortDirection),
+			Offset:    int32(q.Offset),
```
jharveyb, line 135: nit: unneeded int32 casts.

rpcserver.go
```diff
@@ -3058,17 +3063,32 @@ func marshalLeafKey(leafKey universe.LeafKey) *unirpc.AssetKey {
...
 	if err != nil {
 		return nil, err
 	}
 
 	// TODO(roasbeef): tell above if was tring or not, then would set
 	// below diff
 
-	leafKeys, err := r.cfg.UniverseArchive.UniverseLeafKeys(ctx, universeID)
+	if req.Limit > universe.MaxPageSize || req.Limit < 0 {
+		return nil, fmt.Errorf("invalid request limit")
```
jharveyb, line 3081: nit: include max page size in error message.

jharveyb (approved):
LGTM! Great to have this in, can play with the specific limits later on.

rpcserver.go
GeorgeTsagk: oh that was my previous message, thought of leaving it out in order to not expose server params

PR #637, opened 2023-10-30
Add backoff procedure to the proof courier receive procedure

ffranr:
Fixes https://github.com/lightninglabs/taproot-assets/issues/512

This PR generalises the proof courier proof transfer backoff procedure so that it can be used for both retrieving and delivering proofs. It also extends the universe proof courier's proof retrieval functionality using the backoff handler.

The new itest is currently failing because the custodian does not correctly populate the correct event subscribers list within the proof courier instance.


tapdb/assets_store.go
```diff
@@ -294,7 +293,8 @@ type ActiveAssetsStore interface {
 		assetID []byte) (sqlc.FetchAssetMetaForAssetRow, error)
 }
 
-type InsertRecvProofTxAttemptParams = sqlc.InsertReceiverProofTransferAttemptParams
+type LogProofTransferAttemptParams = sqlc.LogProofTransferAttemptParams
```
Roasbeef, line 296: Style nit: can nest under the same `type` statement. Also might as well make a shorter name, as otherwise only saving on the the `sqlc` qualifier.

tapdb/sqlc/migrations/000011_transfers.up.sql
```diff
@@ -0,0 +1,17 @@
+DROP TABLE IF EXISTS receiver_proof_transfer_attempts;
```
Roasbeef, line 1: How will this affect clients that are updating mid send/recv?

proof/courier.go
```diff
@@ -632,53 +632,56 @@ func (b *BackoffHandler) initialDelay(ctx context.Context,
...
-// time delayed strategy. The backoff strategy is used to ensure that we don't
-// spam the courier service with proof delivery attempts.
-func (b *BackoffHandler) Exec(ctx context.Context, proofLocator Locator,
-	targetFunc func() error, subscriberEvent func(fn.Event)) error {
+// Exec attempts to execute the given proof transfer function using a repeating
+// backoff time delayed strategy. The backoff strategy is used to ensure that we
+// don't spam the courier service with proof transfer attempts.
+func (b *BackoffHandler) Exec(ctx context.Context,
+	proofLocator Locator, transferType TransferType,
+	transferFunc func() ([]byte, error),
```
Roasbeef, line 640: Perhaps pull out transfer func and into a top level type def so this is a bit easier to read? Eg: why can it now also return a blob?

proof/courier.go
```diff
@@ -632,53 +632,56 @@ func (b *BackoffHandler) initialDelay(ctx context.Context,
...
-			return fmt.Errorf("unable to log proof "+
-				"delivery attempt: %w", err)
+			return nil, fmt.Errorf("unable to log proof delivery "+
+				"attempt: %w", err)
 		}
 
-		// Execute target function.
-		errExec = targetFunc()
+		// Execute target proof transfer function.
+		proofBlob, errExec = transferFunc()
```
Roasbeef, line 685: Why does back off exec need to know anything about proof blobs?

proof/courier.go
```diff
@@ -1187,13 +1187,32 @@ func (c *UniverseRpcCourier) ReceiveProof(ctx context.Context,
 			LeafKey: assetKey,
 		}
 
-		resp, err := c.client.QueryProof(ctx, &universeKey)
+		// Setup proof receive/query routine and start backoff
+		// procedure.
+		receiveFunc := func() ([]byte, error) {
```
Roasbeef, line 1195: Ah I see where the blob comes in now. Alternatively, it could just assign to some closure variable? Just thinking about symmetry/uniformity with the interface here.

Roasbeef (changes requested):

tapdb/sqlc/migrations/000011_transfers.up.sql
ffranr: I think that in the worst case they'll just skip a delay in their interaction with the courier service.

proof/courier.go
ffranr: I'll change it to a closure variable. That way the target function signature will always be minimal.

proof/courier.go
ffranr: I'll use a closure variable and keep the target function signature minimal.

GeorgeTsagk (commented):
lgtm overall, waiting for the itest update :mailbox:

proof/courier.go
```diff
@@ -832,14 +836,15 @@ func (h *HashMailCourier) DeliverProof(ctx context.Context,
 		defer cancel()
 		err = h.mailbox.RecvAck(ctxTimeout, receiverStreamID)
 		if err != nil {
-			return fmt.Errorf("failed to receive ACK "+
+			return nil, fmt.Errorf("failed to receive ACK "+
 				"from receiver within timeout: %w", err)
 		}
 
-		return nil
+		return nil, nil
```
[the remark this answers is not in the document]
ffranr: The delivery function did not need to return the proof blob that was delivered. I've reformulated the "target function" callback in this PR's latest commits. If there is a proof blob that needs to be obtained from the closure we can use a closure variable rather than a returned variable. So this "returning nil for `[]byte` situation" has been resolved as well.

proof/courier.go
```diff
@@ -1110,21 +1115,23 @@ func (c *UniverseRpcCourier) DeliverProof(ctx context.Context,
...
 			if err != nil {
-				return fmt.Errorf("error inserting proof into "+
-					"universe courier service: %w", err)
+				return nil, fmt.Errorf("error inserting "+
+					"proof into universe courier "+
+					"service: %w", err)
 			}
 
-			return nil
+			return nil, nil
```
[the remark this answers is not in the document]
ffranr: See here: https://github.com/lightninglabs/taproot-assets/pull/637#discussion_r1377661872

ffranr:
I've updated this PR and the new itest is now passing.

This PR does not add the backoff procedure to the hashmail proof courier receive process.

This PR does not add an itest to ensure that the asset is eventually correctly received by the receiving node when the universe proof courier is used.  Such a test does exist for the hashmail proof courier (see `testOfflineReceiverEventuallyReceives`). I don't think that the lack of that test should block this PR.

ffranr:
I've made a few changes to this PR today. I've added a new RPC endpoint for subscribing to asset receive event notifications specifically.

tapgarden/custodian.go
```diff
@@ -86,6 +86,13 @@ type Custodian struct {
 	// events about new proofs being imported.
 	proofSubscription *fn.EventReceiver[proof.Blob]
 
+	// statusEventsSubs is a map of subscribers that want to be notified on
+	// new status events, keyed by their subscription ID.
+	statusEventsSubs map[uint64]*fn.EventReceiver[fn.Event]
```
Roasbeef, line 91: Non-blocking, but we could make this a `lntuils.SyncMap` which'll abstract away from some of the mutex usage.

itest/send_test.go
```diff
@@ -779,6 +893,48 @@ func assertRecvNtfsEvent(t *harnessTest, ctx context.Context,
 	require.Equal(t.t, expectedCount, countFound)
 }
 
+// assertRecvNtfsEvent asserts that the given asset receive event notification
+// was received. This function will block until the event is received or the
+// event stream is closed.
+func assertAssetRecvNtfsEvent(t *harnessTest, ctx context.Context,
```
Roasbeef, line 899: `godoc` mismatch

Roasbeef (approved):
LGTM 🎉

tapgarden/custodian.go
ffranr: I'll look into implementing `SyncMap` in a separate PR.

jharveyb: May allow us to reuse this subscriber setup + publish + etc. logic across multiple packages / state machines as well (if it becomes a bit more generic).

ffranr:
> One bit I'm confused about is teh new ntfn channel in the custodian - I don't see additional code for actually broadcasting events on that channel, is it rebroadcasting backoff events from somewhere? I see it in the new itest but am missing how its wired up.

@jharveyb The custodian executes 
```
courier.SetSubscribers(c.statusEventsSubs)
```

And then the courier broadcasts backoff handler publishes events via the `Exec` function's `subscriberEvent` callback.

In other words, all we had to do to hookup the custodian was to register its subscribers with the courier.

GeorgeTsagk (approved):
LGTM :mailbox:

Issue #638, opened 2023-10-31
Add testing coverage for backoff procedure in universe proof courier


dstadulis:
### Background

#512 will be merged for `v0.3.1` but tests will be added in `v0.3.2`

### Necessary test coverage

- [ ] itests
- [ ] unit test

Will elaborate more with Robyn

PR #644, opened 2023-11-02
Ensure receiver node backoff retries finally succeed in asset send via universe proof courier

ffranr:
Fixes https://github.com/lightninglabs/taproot-assets/issues/638

Before the changes introduced in this PR, the universe proof courier proof retrieval itest did not ensure that the proof was actually retrievable after multiple retrieval attempts were conducted. This PR ensures that the proof can actually be retrieved after repeated retrieval attempts using the backoff procedure.

This PR is ready for review.

### DIFF
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/hex"
+	"fmt"
 	"io"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


Roasbeef:
Needs a rebase!

itest/send_test.go
```diff
@@ -621,30 +622,90 @@ func testReattemptFailedReceiveUniCourier(t *harnessTest) {
...
+	// the receiver node. The receiver node will fail to retrieve the proof
+	// from the proof courier. We simulate this failure by stopping the
+	// proof courier service and then restarting the receiver tapd node.
+	t.Logf("Stopping the proof courier service")
+	require.NoError(t.t, t.proofCourier.Stop())
+
+	// Restart receiving tapd node.
+	t.Logf("Re-starting receiving tapd node")
+	require.NoError(t.t, receiveTapd.start(false))
+	// Defer stopping the receiving tapd node to ensure that it is stopped
```
Roasbeef, line 688: Style nit: missing newline above.

itest/send_test.go
```diff
@@ -621,30 +622,90 @@ func testReattemptFailedReceiveUniCourier(t *harnessTest) {
...
+	require.NoError(t.t, receiveTapd.stop(false))
+
+	// Send asset and then mine to confirm the associated on-chain tx.
+	sendAssetsToAddr(t, sendTapd, recvAddr)
+	_ = MineBlocks(t.t, t.lndHarness.Miner.Client, 1, 1)
+
+	// At this point, the proof courier service is running. We will
+	// therefore pause to allow the sender to transfer the proof to the
+	// proof courier service.
+	time.Sleep(2 * time.Second)
```
Roasbeef, line 676: Can we instead hook this into the sender side events to eliminate the sleep?

jharveyb:
> Needs a rebase!

🎵  Another one 🎵

itest/send_test.go
jharveyb: Also use smaller timeouts for `BackoffResetWait` and `InitialBackoff` to drop total test time? Or is there some floor.

jharveyb: Actually those don't seem to be the blockers for test time, I think it may be the receiver-side settings? And the params are applying to proof send not recv. Not sure.

jharveyb (commented):
Looks good! Ideally we drop the timeouts but does indeed test receiver backoff.

itest/send_test.go
ffranr: I think we could listen for a `SendStateComplete` state event from the sender node's chain porter. But I don't think it's ideal. Ideally we'd have a "proof delivered" event which contained the right data to tie it back to the particular send we're interested in.

I think there's room to add more event listening/assertion itest helper functions also. Listening for an event requires many lines of code currently.

Maybe we circle back once we've got better events in place? I can add a TODO if you agree.

jharveyb (approved):

itest/send_test.go
ffranr: > Using timeouts of:

@jharveyb doesn't always pass on my computer with those numbers. I think they're cutting it too close.

Roasbeef (approved):
LGTM 🧞

Issue #643, opened 2023-11-02
[bug]: Not adding server to federation


AnYing6067:
### Background

when i tried to start tapd, couldn`t start successfully

### Your environment

{
    "version": "0.3.0-alpha commit=v0.3.0",
    "lnd_version": "0.17.0-beta",
    "network": "mainnet",
    "lnd_identity_pubkey": "0374d031900f514526ccf25bace30da82a0dee925415c25287846a09ac334ca55c",
    "node_alias": "0374d031900f514526cc",
    "block_height": 814916,
    "block_hash": "00000000000000000001b74f4b02ed6afca7ae080ee1fb75e212f4736c9d6f96",
    "sync_to_chain": true
}

lnd is running normal

![image](https://github.com/lightninglabs/taproot-assets/assets/34560626/794d8106-7361-4291-8d00-646b3e21c34e)

thanks

PR #647, opened 2023-11-02
rpc: remove num assets from universe Info call

Roasbeef:
The call it just used as a quick sanity check, so doesn't need to depend on a potentially expensive db transaction.

Fixes https://github.com/lightninglabs/taproot-assets/issues/643


jharveyb (approved):
Makes sense, LGTM - Could update after the stats rework.

PR #657, opened 2023-11-06
taprpc: Move asset type to genesis info

### LINKED ISSUE #655: Move RPC message field `asset_type` into field `asset_genesis` within the `Asset` RPC message
The `asset_type` RPC field can be found here: https://github.com/lightninglabs/taproot-assets/blob/c3718c035a8e54ed8d07feabfd106c243775d09e/taprpc/taprootassets.proto#L294-L294

We should move it into the `asset_genesis` field found within the same RPC message (`Asset`). This change makes sense because the asset type is part of the immutable genesis metadata.

In our golang code we can see that the asset type is correctly an asset genesis field, see:
https://github.com/lightninglabs/taproot-assets/blob/1e24b9997431cb14786550109292b8d6c90525c0/asset/asset.go#L170-L170

GeorgeTsagk:
Closes #655


jharveyb (approved):
LGTM 👍🏽  Smaller than expected, didn't realize we use `AssetTypt` outside of Genesis in so many places.

ffranr (approved):
Nice! :100: I think this will make more sense to users.

PR #658, opened 2023-11-07
tapdb: ensure universe asset stats ticker is stopped before refresh

Roasbeef:
In this commit, we fix a bug that would cause the universe stats cache to be wiped more than once every 30 minutes. Before this commit, we'd accumulate a new background `time.AfterFunc` ticker for each unique request set in the cache.

To fix this, we'll always first stop the old ticker before making a new one.

### DIFF
--- a/tapdb/universe_stats.go
+++ b/tapdb/universe_stats.go
@@ -512,6 +512,7 @@ func (u *UniverseStats) populateSyncStatsCache() {
 
 	// If this is a test, then we'll just purge the items.
 	if u.opts.cacheDuration == 0 {
+		log.Debugf("nil state cache duration, wiping cache")
 		u.statsSnapshot.Store(nil)
 		return
 	}
@@ -535,7 +536,10 @@ func (u *UniverseStats) populateSyncStatsCache() {
 
 	// Reset the timer so we'll refresh again after the cache duration.
 	if !u.statsRefresh.Stop() {
-		<-u.statsRefresh.C
+		select {
+		case <-u.statsRefresh.C:
+		default:
+		}
 	}
 
 	u.statsRefresh.Reset(u.opts.cacheDuration)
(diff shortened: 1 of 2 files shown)


tapdb/universe_stats.go
```diff
@@ -837,6 +840,13 @@ func (u *UniverseStats) QuerySyncStats(ctx context.Context,
 
 	// Finally, we'll create the time after function that'll wipe the
 	// cache, forcing a refresh.
+	//
+	// If we already have a timer active, then stop it, so we only have a
+	// single timer going at any given time.
+	if u.syncStatsRefresh != nil && !u.syncStatsRefresh.Stop() {
+		<-u.syncStatsRefresh.C
```
GeorgeTsagk, line 847: what are the guarantees of reading immediately from `u.syncStatsRefresh.C`?
could also do
```suggestion
                select {
		case <-u.syncStatsRefresh.C:
		case <-ctx.Done():
			return nil, ctx.Err()
		}
```

tapdb/universe_stats.go
```diff
@@ -152,7 +153,7 @@ type assetEventsCache = *lru.Cache[eventQuery, cachedAssetEvents]
 
 // statsQueryCacheSize is the total number of asset query responses that we'll
 // hold inside the cache.
-const statsQueryCacheSize = 80_0000
```
GeorgeTsagk, line 155: TIL: you can place underscore separators anywhere

jharveyb (approved):
LGTM, was initially confused about why we use this timer differently from `statsRefresh` but makes sense. I think this is the only place this change is needed.

tapdb/universe_stats.go
Roasbeef: Yep, super useful for numbers like this! In this case I still miscounted the amount of zeros lol.

Issue #650, opened 2023-11-03
Custodian should emit an RPC notification status event when the recieve process completes sucessfully


ffranr:
The custodian handles receiving assets. It should emit an RPC notification status event over the subscribers channel which is initialised via the RPC endpoint `SubscribeReceiveAssetEventNtfns`. This feature was requested by a user on slack [here](https://lightningcommunity.slack.com/archives/C03B3556HQ8/p1698818230588589).

Currently, the custodian does emit events emanating from the proof courier. This is an example of an event notification sent to subscribers:
https://github.com/lightninglabs/taproot-assets/blob/1e24b9997431cb14786550109292b8d6c90525c0/proof/courier.go#L704-L704

The new event should implement the `fn.Event` interface. It should contain (as fields) the corresponding address by which the asset was sent.

The new event should be emitted at this point: https://github.com/lightninglabs/taproot-assets/blob/1e24b9997431cb14786550109292b8d6c90525c0/tapgarden/custodian.go#L459-L459

dstadulis:
Would like to consider what other event notifications could be added that may be helpful applications to use

ffranr:
> Would like to consider what other event notifications could be added that may be helpful applications to use

A user on Slack requested an event notification which would be published once a sufficient number of confirmation block had been mined. (Confirming an asset transfer.)

dstadulis:
How about just subscribe to confirmation count or reporting when the tap client meets it's "confirmed" threshold

PR #659, opened 2023-11-07
Custodian emits a new asset-receive-complete event to notification subscribers

ffranr:
Fixes https://github.com/lightninglabs/taproot-assets/issues/650


tapgarden/custodian.go
```diff
@@ -698,6 +733,19 @@ func (c *Custodian) RegisterSubscriber(receiver *fn.EventReceiver[fn.Event],
...
+// publishSubscriberStatusEvent publishes an event to all status events
+// subscribers.
+func (c *Custodian) publishSubscriberStatusEvent(event fn.Event) {
+	// Lock the subscriber mutex to ensure that we don't modify the
+	// subscriber map while we're iterating over it.
+	c.statusEventsSubsMtx.Lock()
+	defer c.statusEventsSubsMtx.Unlock()
+
+	for _, sub := range c.statusEventsSubs {
+		sub.NewItemCreated.ChanIn() <- event
```
GeorgeTsagk, line 745: is this non-blocking?

if it's blocking then maybe worth considering bringing down the ctx to this function to ensure healthy exit

itest/send_test.go
```diff
@@ -754,6 +754,10 @@ func testReattemptFailedReceiveUniCourier(t *harnessTest) {
 	// proof(s).
 	t.Logf("Attempting to confirm asset received by receiver node")
 	AssertNonInteractiveRecvComplete(t.t, receiveTapd, 1)
+
+	// Confirm that the sender tapd node eventually receives the asset
+	// transfer and publishes an asset recv complete event.
+	assertAssetRecvCompleteEvent(t, ctxb, 5*time.Second, eventNtfns)
```
GeorgeTsagk, line 760: Maybe could go a tiny bit further and also check receiving address?

GeorgeTsagk (approved):
Looks good! couple of minor comments

tapgarden/custodian.go
ffranr: Good point! I'll use `fn.SendOrQuit`.

itest/send_test.go
ffranr: Good idea! I've added the recv address as an arg to `assertAssetRecvCompleteEvent`.

tapgarden/custodian.go
jharveyb: Looks like we'd want to use `SendOrQuit` for all event publishing, not just here? Could be follow-up PR.

tapgarden/custodian.go
```diff
@@ -458,6 +488,16 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 				log.Errorf("unable to import proofs: %v", err)
 				return
 			}
+
+			// At this point the "receive" process is complete. We
+			// will now notify all status event subscribers.
+			recvCompleteEvent := NewAssetReceiveCompleteEvent(*addr, op)
```
jharveyb, line 494: nit: line length; could use recv instead of receive in some places here

jharveyb (approved):
Looks good, clean PR! 💯 

Needs another rebase.

tapgarden/custodian.go
ffranr: > SendOrQuit

Good idea!

Issue #613, opened 2023-10-23
[bug]: neutrino header verifier times out on mainnet proof verification/sync


jharveyb:
### Background

Describe your issue here.

Mainnet tapd was offline for about a week. After start, I triggered a manual sync, which failed with error:

```sh
2023-10-23 16:13:24.392 [WRN] UNIV: encountered an error whilst syncing with server=(universe.ServerAddr) {
 ID: (int64) 1,
 addrStr: (string) (len=32) "universe.lightning.finance:10029",
 addr: (net.Addr) <nil>
}
: %!w(*fmt.wrapError=&{unable to register issuance proofs: unable to register issuance proofs: unable to verify issuance proofs: unable to verify proof: failed to validate proof block header: unable to retrieve block: rpc error: code = Unknown desc = did not get response before timeout 0xc0003d5f40})
```

Also observed a similar error from the scheduled background sync.

### Your environment

```sh
tapcli getinfo                                    # version of `tapd`, `lnd`, and network
uname -mrsv                                          # operating system 
bitcoind --version || btcd --version # version of `btcd`, `bitcoind`, or other backend
```
* any other relevant environment details

Testing PR https://github.com/lightninglabs/taproot-assets/pull/609 , lnd 0.17.0 with neutrino.

### Steps to reproduce

Leave tapd offline for a bit, then try to sync with the default universe.

Roasbeef:
Looks like we make the header verifier here: https://github.com/lightninglabs/taproot-assets/blob/c7d57c597dfaadf0732a9ffb4bf63152c427f8f8/tapcfg/server.go#L129 

I think it's a timeout issue on the neutrino side of things. The block takes to long to fetch, or if all your peers are pruned then maybe they don't have the block.

dstadulis:
https://github.com/lightningnetwork/lnd/pull/8111 will be used to retrieve the header information even if the neutrino backend doesn't have the full block.

jharveyb:
Tracking changes here: #315.

dstadulis:
fixed with #660

PR #660, opened 2023-11-08
chain_bridge: fetch block headers over blocks when verifying proofs

### LINKED ISSUE #315: tarogarden: cache block header verifier results during batch confirmation

jharveyb:
Fixes #315 , #613 . Depends on https://github.com/lightninglabs/lndclient/pull/161 .

This makes use of the `GetBlockHeader` RPC now available in LND v0.17.1 to avoid fetching a full block during proof verification.

`VerifyBlock` is used in the `HeaderVerifier`, which never made use of the full block. This should be safe to change for all places where a `HeaderVerifier` is used, so this is added without adding a separate implementation of `HeaderVerifier`.

Confirmed that the `GetBlock` calls are no longer made by running `tapd` with this PR backed by LND v0.17.1-beta-rc1 and manually syncing.

### DIFF
--- a/chain_bridge.go
+++ b/chain_bridge.go
@@ -9,6 +9,7 @@ import (
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/taproot-assets/tapgarden"
 	"github.com/lightningnetwork/lnd/chainntnfs"
+	"github.com/lightningnetwork/lnd/lnrpc/verrpc"
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 )
 
@@ -16,6 +17,8 @@ import (
 // interface backed by an active remote lnd node.
 type LndRpcChainBridge struct {
 	lnd *lndclient.LndServices
+
+	getBlockHeaderSupported *bool
 }
 
 // NewLndRpcChainBridge creates a new chain bridge from an active lnd services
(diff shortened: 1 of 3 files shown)


dstadulis:
- [ ] test with old version to test fall back

jharveyb:
Have tested that this uses `GetBlockHeader()` with lnd `0.17.1`, still need to manually test that this will fall back to `GetBlock()` with lnd `0.17.0`.

jharveyb:
Bumped to appropriate `lndclient` and `lnd` versions, ready for merge.

GeorgeTsagk (approved):
Looks good :100:

ffranr (approved):
Looks good to me! I appreciate the small diff. :+1:

PR #663, opened 2023-11-08
RPC: Remove `anchor_txid` from `AnchorInfo`

### LINKED ISSUE #654: Remove field `anchor_txid` from RPC message `AnchorInfo`.
The txid is included in field `anchor_outpoint`. We should therefore remove field `anchor_txid` from the `AnchorInfo` RPC message.

GeorgeTsagk:
Closes #654


itest/assertions.go
```diff
@@ -69,9 +70,12 @@ func AssetAnchorCheck(txid, blockHash chainhash.Hash) AssetCheck {
 			return fmt.Errorf("asset is missing chain anchor field")
 		}
 
-		if a.ChainAnchor.AnchorTxid != txid.String() {
+		anchorTxid :=
+			strings.Split(a.ChainAnchor.AnchorOutpoint, ":")[0]
```
GeorgeTsagk, line 74: Since the txid info was duplicate I now retrieve it by splitting the outpoint string

GeorgeTsagk:
There was a flake on a batch mint assets test, kicked it and passed, just flagging

itest/assertions.go
ffranr: I think you should use `wire.NewOutPointFromString` here.

ffranr: I think you should use `wire.NewOutPointFromString` here.

ffranr (changes requested):

itest/assertions.go
GeorgeTsagk: Good call, is the hash reversed in the raw outpoint string?

itest/assertions.go
```diff
@@ -69,9 +69,18 @@ func AssetAnchorCheck(txid, blockHash chainhash.Hash) AssetCheck {
 			return fmt.Errorf("asset is missing chain anchor field")
 		}
 
-		if a.ChainAnchor.AnchorTxid != txid.String() {
+		out, err :=
+			wire.NewOutPointFromString(a.ChainAnchor.AnchorOutpoint)
+
```
ffranr, line 74: nit: this empty line is unnecessary

itest/assertions.go
```diff
@@ -69,9 +69,18 @@ func AssetAnchorCheck(txid, blockHash chainhash.Hash) AssetCheck {
...
+		if err != nil {
+			return fmt.Errorf("unable to parse outpoint: %v", err)
+		}
+
+		anchorTxid := out.Hash.String()
+
+		if anchorTxid != txid.String() {
 			return fmt.Errorf("unexpected asset anchor TXID, got "+
-				"%v wanted %x", a.ChainAnchor.AnchorTxid,
+				"%v wanted %x", anchorTxid,
```
ffranr, line 83: nit: the `txid[:]` can be on the same line as `"%v wanted %x", anchorTxid,`

itest/assets_test.go
```diff
@@ -173,11 +174,15 @@ func transferAssetProofs(t *harnessTest, src, dst *tapdHarness,
...
-		anchorTxHash, err := chainhash.NewHashFromStr(
-			existingAsset.ChainAnchor.AnchorTxid,
+		out, err := wire.NewOutPointFromString(
+			existingAsset.ChainAnchor.AnchorOutpoint,
 		)
 		require.NoError(t.t, err)
 
+		anchorTxHash := out.Hash
+
+		require.NoError(t.t, err)
```
ffranr, line 184: This line is redundant. We're doing the same thing two statements above.

ffranr (changes requested):
Looks good to me! Just two or so nits.

I'm a fan of this change because it reduces our test surface.

itest/assertions.go
ffranr: Yes, the outpoint string consists of the `txid` and the `vout` index. The `txid` in the outpoint string is a hex encoding of the byte reversed hash.

ffranr (approved):
Nice! :100:

jharveyb (approved):
LGTM 👍🏽

PR #667, opened 2023-11-14
Expose anchor point in AssetStats

### LINKED ISSUE #641: rpc: add `anchor_point` to the `QueryAssetStats` endpoint
Today for the `QueryAssetStats` endpoint, we return the `genesis_point `, and `genesis_height` and `genesis_timestamp`. The `genesis_point` is the outpoint of the first input in the transaction that created an asset. The `genesis_height` is the height of the block that included the transaction, and the `genesis_timestamp` is the timestamp in the block header for that block. 

Some UIs were confused about `genesis_point`: they thought it was the output that created the asset. To clarify this, we should add `anchor_point`, which'll be the outpoint of the transaction that created the asset. We can get this information from the `genesis_assets` table, as we already return the `output_index`, so the query juts needs to obtain the `txid` of that anchor transaction.

GeorgeTsagk:
Closes #641


ffranr (approved):
Looks good to me! :congratulations:

jharveyb:
Looks good! Main note is to assert that this is being populated in itests, could add that here:

https://github.com/lightninglabs/taproot-assets/blob/6dd69b2f9ed5e8875cfc087ba4a4cc0ca4d1fc11/itest/assertions.go#L1302

jharveyb (approved):
LGTM 🚀

Issue #665, opened 2023-11-13
[bug]: Troubleshoot discrepancy between command results: `asset balance` and `assets list`


ffranr:
@HannahMR reported the following:

> It seems there is an error in the tapcli assets balance command on v0.3.1. My balance is 10, but appears as 30. 

> So I was making demo videos and trasfering assets between an Alice node and a Bob node. The Alice node is my tapd demo node that I've been using for the demo videos, and my Bob node is the Velas Commerce testnet node. Both are running LND 0.17 and tapd 0.3.1

> I minted two rounds of two different assets, 100 units each time, on my Alice node, one of which I named APIdemobux. On Friday I had sent two transactions of 10 APIdemobux to Leo, he send 10 of those back leaving me with a total of 190 APIdemobux. 

> Today, for a demo video, using the command line, I transfered 10 APIdemobux to the Bob node. At the time of recording, those assets were showing up as expected when using the $ tapcli assets balance command. 

> However, an hour or so later, when running that command on the Bob node, the balance showed as 20. And after restarting the tapd service on the server, the balances shows as 30! 

> I can however run the assets list command and the balance apears appropriately as 10. 

> The output of the tapcli assets balance and the assets list command on the Alice node are still showing the correct amounts. 

> I only put that transaction through one time... just went and double checked the video. 

> I can send logs if that's helpful. I also have the output of the balance and list commands from each node.

Just to summaries the issue I think you've encountered: the reported balance for an asset is different between these two commands:
`tapcli asset balance` and `tapcli assets list`. And also the result returned by `tapcli asset balance` changes over time after receiving an asset.

I've got a copy of Bob's `tapd.db`. I will inspect.

HannahMR:
[tapd.log](https://github.com/lightninglabs/taproot-assets/files/13351934/tapd.log)
[tapd.log.5.gz](https://github.com/lightninglabs/taproot-assets/files/13351935/tapd.log.5.gz)
[tapd.log.6.gz](https://github.com/lightninglabs/taproot-assets/files/13351936/tapd.log.6.gz)
[tapd.log.7.gz](https://github.com/lightninglabs/taproot-assets/files/13351938/tapd.log.7.gz)

ffranr:
Thank you Hannah for the logs. Unfortunately they aren't very useful for this issue. The syncer log messages are so numerous I think they've squeezed out any potentially interesting messages.

dstadulis:
plan
- [ ] turn off debugger
[...]

PR #679, opened 2023-11-16
itest: attempt to reproduce asset balance bug report

ffranr:
Bug report:
https://github.com/lightninglabs/taproot-assets/issues/665

This PR attempts to reproduce the bug reported in the issue above. It does not succeed. These are the rough steps taken in this PR's bug reproduction attempt:
1. Send an asset to a receiver node.
2. Assert that the receiver node's asset balance is correct.
3. Restart the receiver node. Repeat 2.
4. Mine during receiver node restart. Repeat 2.

This PR is based on the commit corresponding to release tag `v0.3.1`. I think the receiving node was at release `v0.3.1`.

This PR shouldn't be merged.

### DIFF
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -269,7 +269,7 @@ func testRestartReceiverCheckBalance(t *harnessTest) {
 			break
 		}
 
-		require.Equal(t.t, assetGroupBalance.Balance, uint64(10))
+		require.Equal(t.t, int(10), int(assetGroupBalance.Balance))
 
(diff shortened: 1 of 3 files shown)


ffranr:
I've added a commit to this PR. It now reproduces the asset balance bug reported here: https://github.com/lightninglabs/taproot-assets/issues/665

The latest commit removes a delay which effectively hid the bug from the itests.

ffranr:
These commits are now in use in two other PRs that we can potentially merge.

Issue #627, opened 2023-10-26
[bug]: When using --group_key AND --enable_emission the assets have different tweaked_group_key


straumat:
### Background

When emiting a second emissions of an asset, if you set `--group_key` AND `--enable_emission`, the second asset emited will have a different `tweaked_group_key`.

### Your environment

```
tapd v0.3-rc2
```

### Steps to reproduce
I mint my first asset
```
tapcli assets mint --type normal \ 
                --name unlimitedRoylloCoin1 \
                --enable_emission true
```
I get this result:
```
"name": "unlimitedRoylloCoin1"
"asset_id": "ce2d57a6bef7668a781bc149f5491aab7392b30a2f4368f013722f46a57abc70"
  "asset_group": {
    "raw_group_key": "036c1fb8f682b877b9caa27ddb7068a242f662332ea7eb99e3c32c12170c15a0bc",
    "tweaked_group_key": "0349d60c6689bbbb2ffc9b30b45b96a3d6e5fd5fd01c867344c0054f6105048241",
    "asset_witness": "0140e0c383aaecabb44bee72683d825118b7ef7124a9c8be149c82b7c4c525340aceb4194a177bc1e13e0bee7df5425720f4b5f30335faed586bef129780146b7a2c"
  },
```
I mint a second a second asset:
```
tapcli assets mint --type normal \
                --name unlimitedRoylloCoin2 \
                --enable_emission true \
                --group_key 0349d60c6689bbbb2ffc9b30b45b96a3d6e5fd5fd01c867344c0054f6105048241
```
And as result:
```
"name": "unlimitedRoylloCoin2",
"asset_id": "8deae8ae25e1e46f8d74ba47b28860001965628d2c011f8f19f9e07dbecec7d0"
  "asset_group": {
    "raw_group_key": "02a97752df0a27efc44bf7bf0dcf7113b391ba6e1884cd5a78fe493f4da91e8f99",
    "tweaked_group_key": "02807f3a0009f9ed5c3ba0e3662c511619beb216e7f165b9ba77668c4835776b96",
    "asset_witness": "0140d5cccebf06fb9e26a8baa1fccf01f59c3c411a87dde8ff16e710fe0193eded77153dfb5d954f810b7f5920192f72f856b024e89f4eba1099087fb3dc4b3efe35"
  },
```
tweaked_group_key are different for both emissions.

### Expected behavior

[...]

PR #681, opened 2023-11-16
split `enableEmission` into `newGroupedAsset` and `groupedAsset`

### LINKED ISSUE #236: add ability to set the emissions CLI flag to false
Currently, IFF the emissions CLI flag is set to true will the taro daemon be set to mint an asset with possible, subsequent emissions.  Users are currently assuming that the same field can be set to false to prescribe an asset not have subsequent emissions.  This is causing repeated user errors and is is a reasonable so assumption.

The issue will be satisfied by tarod allowing the emissions CLI field to also accept `false`

jharveyb:
Fixes #236 , #627 .

We replace the `enableEmission` flag with two flags, `newGroupedAsset` and `groupedAsset`, at both the RPC and CLI level.

Correct usage:

Asset that creates an asset group, allows future emission / reissuance: `--new_grouped_asset`

Asset minted into an asset group specified by group key: `--grouped_asset --group_key`

Asset minted into the same group as another asset in the same minting batch, specified by asset name: `--grouped_asset --group_anchor`

We also fix a small bug where displaying a pending batch would not show if an asset had `--new_grouped_asset` set, and port a fix for a logging issue that could cause a crash.


GeorgeTsagk (commented):
PR looks good, will also do some manual testing

vectors seem to be failing? (kicked it, will see if it passes)

taprpc/mintrpc/mint.proto
```diff
@@ -33,40 +33,74 @@ service Mint {
     rpc ListBatches (ListBatchRequest) returns (ListBatchResponse);
 }
 
-message MintAsset {
+message MintSeedling {
```
ffranr, line 36: Is this the first time we're exposing users to the concept of a "seedling"? Perhaps we should add a doc comment to explain what this message type is.

I don't think we should hold up this PR on terminology. But perhaps we should think carefully about "seepling" if this is the first time a user might encounter that word in this context. `MintCandidateAsset`?

ffranr (approved):
Great to see this cleared up :+1:

tapgarden/caretaker.go
```diff
@@ -785,7 +785,7 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
 		case strings.Contains(err.Error(), "already exists"):
 			break
 
-		case err != nil:
+		default:
```
Roasbeef, line 788: isn't this equiv to the above? A non-nil error ends up here w/ the prior version that had explicit err check. A commit body message could help here re rationale/understanding.

tapgarden/custodian.go
```diff
@@ -546,12 +546,15 @@ func (c *Custodian) importAddrToWallet(addr *address.AddrWithKeyInfo) error {
 	// along if so.
 	case strings.Contains(err.Error(), "already exists"):
 
-	case err != nil:
+	default:
```
Roasbeef, line 549: Same here, or there's a condition that `ptrAddr` is nil but the error non-nil? Either way, seems equiv.

taprpc/mintrpc/mint.proto
Roasbeef: IIUC, this will also trigger a larger breaking change as well (proto compilation failing). Seems non-essential?

taprpc/mintrpc/mint.proto
```diff
@@ -33,40 +33,74 @@ service Mint {
...
-    uint64 amount = 4;
+    uint64 amount = 5;
 
     /*
-    The specific group key this asset should be minted with.
+    If true, then the asset will be created with a group key, which allows for
+    future asset issuance.
     */
-    bytes group_key = 5;
+    bool enable_emission = 6;
```
Roasbeef, line 62: I thought we were doing away with this terminology?

taprpc/mintrpc/mint.proto
```diff
@@ -79,14 +113,20 @@ message MintAssetRequest {
     If true, then the asset will be created with a group key, which allows for
     future asset issuance.
     */
-    bool enable_emission = 2;
+    bool new_grouped_asset = 2;
```
Roasbeef, line 116: Any reason to _not_ move this into `MintAsset`?

taprpc/mintrpc/mint.proto
Roasbeef: Looks like just for display, could be `PendingBatchAsset`, or just `PendingAsset`.

rpcserver.go
```diff
@@ -357,12 +357,37 @@ func (r *rpcServer) MintAsset(ctx context.Context,
...
-	if req.EnableEmission {
+	case req.NewGroupedAsset:
 		if specificGroupKey || specificGroupAnchor {
 			return nil, fmt.Errorf("must disable emission to " +
 				"specify a group")
 		}
+
+	// If the asset is intended to be part of an existing group, a group key
+	// or anchor must be specified, but not both.
+	case req.GroupedAsset:
```
Roasbeef, line 376: Nice to see all the cases enumerated here, then compare that to the single catch all that could allow for unintended behavior.

rpcserver.go
```diff
@@ -377,29 +402,24 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 		AssetType:      asset.Type(req.Asset.AssetType),
 		AssetName:      req.Asset.Name,
 		Amount:         req.Asset.Amount,
-		EnableEmission: req.EnableEmission,
+		EnableEmission: req.NewGroupedAsset,
```
Roasbeef, line 405: Could also change the attribute to match as well.

rpcserver.go
```diff
@@ -2534,25 +2554,31 @@ func marshalSeedlings(
...
 			Name:         seedling.AssetName,
 			AssetMeta:    seedlingMeta,
 			Amount:       seedling.Amount,
 			GroupKey:     groupKeyBytes,
 			GroupAnchor:  groupAnchor,
-		})
+		}
+
+		if seedling.EnableEmission {
+			nextSeedling.EnableEmission = true
```
Roasbeef, line 2568: See re doing away with `EnableEmission` on the RPC level.

taprpc/mintrpc/mint.proto
jharveyb: There was a reason when `MintAsset` was used within `MintingBatch` - now that that isn't the case, these could be moved into `MintAsset`.

rpcserver.go
jharveyb: Attribute as in field in `tapgarden.Seedling` ?

jharveyb:
Rebased + renamed `MintSeedling` to `PendingAsset` + renamed the last public `EnableEmission` field to `NewGroupedAsset` + moved the new asset group flags into `MintAsset`.

Roasbeef (approved):
LGTM 🧜‍♀️

PR #684, opened 2023-11-17
Balance bug db asset insert query fix

ffranr:
This PR changes an


ffranr:
Even if this PR works. There's a bunch more things that we need to do to make tapd robust to this problem:
* Unit tests for the `UpsertAsset` query.
* Stopping the custodian from even attempting a duplicate insert.

ffranr:
A less risk solution: https://github.com/lightninglabs/taproot-assets/pull/687

ffranr:
I've just rebased and bumped SQL migration numbers. The itest was removed from this PR because it's already merged.

tapdb/sqlc/migrations/000013_assets_unique_constraint.up.sql
```diff
@@ -0,0 +1,31 @@
+-- We will apply a new unique constraint on the assets table. This constraint
```
guggero, line 1: nit: we should add a comment to the `.down.sql` file that it's empty on purpose.

tapdb/sqlc/migrations/000013_assets_unique_constraint.up.sql
```diff
@@ -0,0 +1,31 @@
...
+    SELECT
+        asset_id,
+        anchor_utxo_id,
+        genesis_id,
+        script_key_id,
+
+        -- This is the row number of the row within the set of rows that violate
+        -- the constraint. We will delete all rows with a row number greater
+        -- than 1.
+        ROW_NUMBER() OVER (PARTITION BY anchor_utxo_id, genesis_id, script_key_id ORDER BY asset_id) AS row_num
```
guggero, line 15: Ah, I see now why we'd want to have a way to test migration scripts...
I think this probably works, but we definitely need to make sure it does the correct thing for both Postgres and SQLite.

tapdb/sqlc/migrations/000013_assets_unique_constraint.up.sql
```diff
@@ -0,0 +1,31 @@
...
+    SELECT anchor_utxo_id, genesis_id, script_key_id
+    FROM duplicate_rows
+    -- Delete all rows with a row number greater than 1. This should leave
+    -- exactly one row (row number 0) for each set of rows that violate the
+    -- constraint.
+    WHERE row_num > 1
+);
+
+-- Create a unique index on the new table
+CREATE UNIQUE INDEX assets_uniqueness_index_anchor_utxo_id_genesis_id_script_key_id
```
guggero, line 29: Do we have a way to remove the index created by the previous `UNIQUE(asset_id, genesis_id, script_key_id)` part? I guess that's probably dependent on the actual DB implementation.
This is likely a downside of using the "inline" unique key creation, that it doesn't have a name and therefore can't easily be updated.

guggero: Hmm, I think I google that before at some point... And if I remember correctly, we basically need to re-create the table to get rid of the original unique constraint. But I think we'd probably want to do that, as otherwise this is an expensive but useless index we're carrying along.
IIRC this is how you replace a table "correctly":
1. `CREATE TABLE assets_tmp ... (full definition, but without the inline UNIQUE constraint)`
2. `INSERT INTO assets_tmp AS SELECT * from assets`
3. Alter any foreign keys that point to the old `assets` table
4. `DROP TABLE assets`
5. `ALTER TABLE assets_tmp RENAME TO assets`

guggero (commented):
Did a first pass. I think the approach is correct, but might need some more involved SQL magic to get rid of the old unique key.

ffranr:
Now that the [db migration work](https://github.com/lightninglabs/taproot-assets/pull/707) has been merged. We should test this PR using that framework.

PR #687, opened 2023-11-20
Check for existing asset in db before attempting to insert a new asset.

ffranr:
This PR is another solution for https://github.com/lightninglabs/taproot-assets/issues/665 .

In these commits we add a check to ensure that if a confirmed asset already exists in our db we do not attempt to insert another copy.

This solution is a better solution than https://github.com/lightninglabs/taproot-assets/pull/684 because it is simpler and less risky.


tapdb/assets_common.go
```diff
@@ -200,6 +204,24 @@ func upsertAssetsWithGenesis(ctx context.Context, q UpsertAssetStore,
...
+			AnchorUtxoID: anchorUtxoID,
+			GenesisID:    sqlInt64(genAssetID),
+			ScriptKeyID:  sqlInt64(scriptKeyID),
+		})
+		if err != nil {
+			return 0, nil, fmt.Errorf("unable to query assets: %w",
+				err)
+		}
+
+		if len(existingAssets) > 0 {
```
jharveyb, line 220: IIUC the returned asset primary keys here are not in a stable order, e.x. if I already have duplicate assets, we don't have a guarantee about the same asset ID always being returned here.

However, I don't think this is problem, as the only place I see the `assetIDs` array returned by `upsertAssetsWithGenesis` being used is here:

https://github.com/lightninglabs/taproot-assets/blob/b8a007d3cb14de85b03dabb606c248491a57b491/tapdb/assets_store.go#L1445

Where we're inserted one asset, which should be new. TL;DR the unstable ordering should be fine.

itest/tapd_harness.go
```diff
@@ -223,6 +224,11 @@ func newTapdHarness(t *testing.T, ht *harnessTest, cfg tapdConfig,
 		finalCfg.HashMailCourier = nil
 	}
 
+	// Set the custodian proof retrieval delay if it was specified.
+	if opts.custodianProofRetrievalDelay != nil {
+		finalCfg.CustodianProofRetrievalDelay = *opts.custodianProofRetrievalDelay
```
jharveyb, line 229: nit: line length

jharveyb (approved):
Looks good! Confirmed that the itest does test this issue of displayed balance.

Sounds like another angle for a fix is to have the custodian not perform these duplicate proof fetches on restart in the first place?

itest/send_test.go
```diff
@@ -135,6 +135,196 @@ func testBasicSendUnidirectional(t *harnessTest) {
...
+
+		timeout := 2 * defaultProofTransferReceiverAckTimeout
+		ctx, cancel := context.WithTimeout(ctxb, timeout)
+		defer cancel()
+		assertAssetSendNtfsEvent(
+			t, ctx, eventNtfns, targetEventSelector, 1,
+		)
+	}()
+
+	// First, we'll make a normal assets with enough units to allow us to
```
GeorgeTsagk, line 195:
```suggestion
	// First, we'll make a normal asset with enough units to allow us to
```

GeorgeTsagk (approved):
Lgtm

ffranr:
> Sounds like another angle for a fix is to have the custodian not perform these duplicate proof fetches on restart in the first place?

@jharveyb yes, I would like to refactor the custodian in another PR. It doesn't handle the events cache very cleanly.

Roasbeef:
> Sounds like another angle for a fix is to have the custodian not perform these duplicate proof fetches on restart in the first place?

Yep, this also needs to be resolved. We have some candidate fixes in mind, and can exercise the fixes at the unit test level.

tapdb/sqlc/queries/assets.sql
```diff
@@ -430,7 +430,10 @@ WHERE (
     assets.amount >= COALESCE(sqlc.narg('min_amt'), assets.amount) AND
     assets.spent = COALESCE(sqlc.narg('spent'), assets.spent) AND
     (key_group_info_view.tweaked_group_key = sqlc.narg('key_group_filter') OR
-      sqlc.narg('key_group_filter') IS NULL)
+      sqlc.narg('key_group_filter') IS NULL) AND
+    assets.anchor_utxo_id = COALESCE(sqlc.narg('anchor_utxo_id'), assets.anchor_utxo_id) AND
+    assets.genesis_id = COALESCE(sqlc.narg('genesis_id'), assets.genesis_id) AND
+    assets.script_key_id = COALESCE(sqlc.narg('script_key_id'), assets.script_key_id)
```
Roasbeef, line 436: We were already able to filter on script key above, see `tweaked_script_key`.

Roasbeef: Are any of these actually useful? Not much detail re rationale int he commit body message.

tapdb/assets_common.go
```diff
@@ -200,6 +204,24 @@ func upsertAssetsWithGenesis(ctx context.Context, q UpsertAssetStore,
 			anchorUtxoID = anchorUtxoIDs[idx]
 		}
 
+		// Check for matching assets in the database. If we find one,
+		// we'll just use its primary key ID, and we won't attempt to
+		// insert the asset again.
+		existingAssets, err := q.QueryAssets(ctx, QueryAssetFilters{
```
Roasbeef, line 210: Ah ok see use now, as a sanity check before insertion.

tapgarden/custodian.go
```diff
@@ -420,7 +421,7 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 			// the proof will very likely fail. We should expect
 			// retrieval success before this delay.
 			select {
-			case <-time.After(defaultProofRetrievalDelay):
+			case <-time.After(c.cfg.ProofRetrievalDelay):
```
Roasbeef, line 424: Is this actually needed, now that we have proper retry?

ffranr: @Roasbeef Using this delay is one methodology which works. We could use another methodology without this delay.

I've documented this delay as follows:
> The number of seconds the custodian waits after identifying an asset transfer on-chain and before retrieving the corresponding proof.

My reasoning for this delay is that we shouldn't expect the first attempt at proof retrieval to work if we're immediately reacting to an on-chain tx. Because the sending node may not have had sufficient time to upload the proof to the proof courier. So, rather than try and expect a guaranteed failure, we delay so that our first attempt at retrieval might succeed.

I think we'll delay either way. Either with the backoff procedure after a first failed attempt. Or before hand, pre-emptively, to avoid failure on our first attempt.

PR #696, opened 2023-11-22
build: bump Go version to Go 1.21.4

Roasbeef:


ffranr:
`GO_VERSION = 1.21.0` in `Makefile` needs to be updated otherwise the linter won't pass.

Issue #526, opened 2023-09-21
universe: add background loop for optimistic asset proof to FederationEnvoy


Roasbeef:
Today after we create an asset, if we have an active universe federation, we'll attempt to push out the new proofs to them. However, if this fails for w/e reason, then we won't attempt to every push out proofs again, unless someone connects to our Universe server and then obtains our proofs, and pushes out to others. 

We should add a background reconciliation loop, that if configured, will push out our proofs to our universe federation. A simple version of this would just be to grab our issuance proofs (or full, etc) then attempt to call an push out to our federation members. This would be somewhat inefficient. Another take would be to instead maintain an on disk queue, and work through the queue ensuring that each time the proof is accepted. If so, then we can stop trying to publish it. 

## Steps To Completion

- [x] Decide in if we'll do best effort basis (not efficient), or use an on disk queue, and treat pushing to a universe like we do sending proof files after a transfer. 
- [ ] Implement option above, figure out how to expose as configuration

Roasbeef:
This TODO can be removed once this is fixed: https://github.com/lightninglabs/taproot-assets/blob/a201d48152b274e8faec390dab969391171af27e/universe/auto_syncer.go#L274-L276

dstadulis:
Is the background loop currently pushing? may be only receiving

dstadulis:
This will ensure that the proof is more properly distributed during the minting process because there isn't a second attempt on the proof publishing.

ffranr:
> maintain an on disk queue, and work through the queue ensuring that each time the proof is accepted. If so, then we can stop trying to publish it.

I'm in favour of this approach.

I think we only need to care about syncing our issuance proofs. And not transfer proofs.

Perhaps a table with columns:
```
[issuance_proof, universe_server, timestamp, proof_sync_status]
```

Where `proof_sync_status` is an enum with values `unknown`, `accepted`, and `rejected`.

This will hopefully give us enough flexibility to see our latest sync history. And also to force re-try based on the sync timestamp, the sync status, and specific to a particular universe server.

Configuration wise, we already have a default sync interval of 10 mins. I think that config parameter can be re-used for this new case. When combined with our issuance proof sync history SQL table we shouldn't spam other universe servers excessively.

One bit of work that we might be missing is the response from the other universe servers. It would be useful if a peer universe server could include in its response whether the proof was rejected (server not interested in proof) or accepted (no need to try again, proof has been recognised).

jharveyb:
[...]

PR #698, opened 2023-11-27
Federation envoy re-attempts sync push for issuance proofs

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/526

This PR adds a new SQL table which logs the federation sync status of select proofs. In this PR we use the new table to log and re-attempt failed asset issuance proof push sync events.

The federation envoy is modified such that tick events kick-off syncing for pending proofs found in the new log table.

I would be happy to receive reviews to double check that I'm on the right track with these changes.

This PR is currently missing tests.

### DIFF
--- a/tapdb/assets_store_test.go
+++ b/tapdb/assets_store_test.go
@@ -246,105 +246,17 @@ func assertAssetEqual(t *testing.T, a, b *asset.Asset) {
 func TestImportAssetProof(t *testing.T) {
 	t.Parallel()
 
-	// First, we'll create a new instance of the database.
-	_, assetStore, db := newAssetStore(t)
+	var (
+		ctxb = context.Background()
 
-	// Next, we'll make a new random asset that also has a few inputs with
-	// dummy witness information.
-	testAsset := randAsset(t)
+		dbHandle   = NewDbHandle(t)
+		assetStore = dbHandle.AssetStore
+	)
 
-	assetRoot, err := commitment.NewAssetCommitment(testAsset)
-	require.NoError(t, err)
-
-	taprootAssetRoot, err := commitment.NewTapCommitment(assetRoot)
-	require.NoError(t, err)
-
-	// With our asset created, we can now create the AnnotatedProof we use
-	// to import assets into the database.
-	var blockHash chainhash.Hash
-	_, err = rand.Read(blockHash[:])
-	require.NoError(t, err)
-
-	anchorTx := wire.NewMsgTx(2)
-	anchorTx.AddTxIn(&wire.TxIn{})
-	anchorTx.AddTxOut(&wire.TxOut{
-		PkScript: bytes.Repeat([]byte{0x01}, 34),
-		Value:    10,
-	})
-
+	// Add a random asset and corresponding proof into the database.
+	testAsset, testProof := dbHandle.AddRandomAssetProof(t)
 	assetID := testAsset.ID()
-	anchorPoint := wire.OutPoint{
-		Hash:  anchorTx.TxHash(),
-		Index: 0,
-	}
-	initialBlob := bytes.Repeat([]byte{0x0}, 100)
-	updatedBlob := bytes.Repeat([]byte{0x77}, 100)
-	testProof := &proof.AnnotatedProof{
-		Locator: proof.Locator{
-			AssetID:   &assetID,
-			ScriptKey: *testAsset.ScriptKey.PubKey,
-		},
-		Blob: initialBlob,
-		AssetSnapshot: &proof.AssetSnapshot{
-			Asset:             testAsset,
-			OutPoint:          anchorPoint,
-			AnchorBlockHash:   blockHash,
-			AnchorBlockHeight: test.RandInt[uint32](),
-			AnchorTxIndex:     test.RandInt[uint32](),
-			AnchorTx:          anchorTx,
-			OutputIndex:       0,
-			InternalKey:       test.RandPubKey(t),
-			ScriptRoot:        taprootAssetRoot,
-		},
-	}
-	if testAsset.GroupKey != nil {
-		testProof.GroupKey = &testAsset.GroupKey.GroupPubKey
-	}
-
-	// We'll now insert the internal key information as well as the script
-	// key ahead of time to reflect the address creation that happens
-	// elsewhere.
-	ctxb := context.Background()
-	_, err = db.UpsertInternalKey(ctxb, InternalKey{
-		RawKey:    testProof.InternalKey.SerializeCompressed(),
-		KeyFamily: test.RandInt[int32](),
-		KeyIndex:  test.RandInt[int32](),
-	})
-	require.NoError(t, err)
-	rawScriptKeyID, err := db.UpsertInternalKey(ctxb, InternalKey{
-		RawKey:    testAsset.ScriptKey.RawKey.PubKey.SerializeCompressed(),
-		KeyFamily: int32(testAsset.ScriptKey.RawKey.Family),
-		KeyIndex:  int32(testAsset.ScriptKey.RawKey.Index),
-	})
-	require.NoError(t, err)
-	_, err = db.UpsertScriptKey(ctxb, NewScriptKey{
-		InternalKeyID:    rawScriptKeyID,
-		TweakedScriptKey: testAsset.ScriptKey.PubKey.SerializeCompressed(),
-		Tweak:            nil,
-	})
-	require.NoError(t, err)
-
-	// We'll add the chain transaction of the proof now to simulate a
-	// batched transfer on a higher layer.
-	var anchorTxBuf bytes.Buffer
-	err = testProof.AnchorTx.Serialize(&anchorTxBuf)
-	require.NoError(t, err)
-	anchorTXID := testProof.AnchorTx.TxHash()
-	_, err = db.UpsertChainTx(ctxb, ChainTxParams{
-		Txid:        anchorTXID[:],
-		RawTx:       anchorTxBuf.Bytes(),
-		BlockHeight: sqlInt32(testProof.AnchorBlockHeight),
-		BlockHash:   testProof.AnchorBlockHash[:],
-		TxIndex:     sqlInt32(testProof.AnchorTxIndex),
-	})
-	require.NoError(t, err, "unable to insert chain tx: %w", err)
-
-	// With all our test data constructed, we'll now attempt to import the
-	// asset into the database.
-	require.NoError(t, assetStore.ImportProofs(
-		ctxb, proof.MockHeaderVerifier, proof.MockGroupVerifier, false,
-		testProof,
-	))
+	initialBlob := testProof.Blob
 
 	// We should now be able to retrieve the set of all assets inserted on
 	// disk.
@@ -371,7 +283,7 @@ func TestImportAssetProof(t *testing.T) {
 		ScriptKey: *testAsset.ScriptKey.PubKey,
 	})
 	require.NoError(t, err)
-	require.Equal(t, initialBlob, []byte(currentBlob))
+	require.Equal(t, initialBlob, currentBlob)
 
 	// We should also be able to fetch the created asset above based on
 	// either the asset ID, or key group via the main coin selection
@@ -391,6 +303,8 @@ func TestImportAssetProof(t *testing.T) {
 
 	// We'll now attempt to overwrite the proof with one that has different
 	// block information (simulating a re-org).
+	updatedBlob := bytes.Repeat([]byte{0x77}, 100)
+
 	testProof.AnchorBlockHash = chainhash.Hash{12, 34, 56}
 	testProof.AnchorBlockHeight = 1234
 	testProof.AnchorTxIndex = 5678
--- a/tapdb/sqlc/migrations/000013_universe_fed_proof_sync_log.up.sql
+++ b/tapdb/sqlc/migrations/000013_universe_fed_proof_sync_log.up.sql
@@ -1,0 +1,36 @@
+-- This table stores the log of federation universe proof sync attempts. Rows
+-- in this table are specific to a given proof leaf, server, and sync direction.
+CREATE TABLE IF NOT EXISTS federation_proof_sync_log (
+    id BIGINT PRIMARY KEY,
+
+    -- The status of the proof sync attempt.
+    status TEXT NOT NULL CHECK(status IN ('pending', 'complete')),
+
+    -- The timestamp of when the log entry for the associated proof was last
+    -- updated.
+    timestamp TIMESTAMP NOT NULL,
+
+    -- The number of attempts that have been made to sync the proof.
+    attempt_counter BIGINT NOT NULL DEFAULT 0,
+
+    -- The direction of the proof sync attempt.
+    sync_direction TEXT NOT NULL CHECK(sync_direction IN ('push', 'pull')),
+
+    -- The ID of the subject proof leaf.
+    proof_leaf_id BIGINT NOT NULL REFERENCES universe_leaves(id),
+
+    -- The ID of the universe that the proof leaf belongs to.
+    universe_root_id BIGINT NOT NULL REFERENCES universe_roots(id),
+
+    -- The ID of the server that the proof will be/was synced to.
+    servers_id BIGINT NOT NULL REFERENCES universe_servers(id)
+);
+
+-- Create a unique index on table federation_proof_sync_log
+CREATE UNIQUE INDEX federation_proof_sync_log_unique_index_proof_leaf_id_servers_id
+ON federation_proof_sync_log (
+    sync_direction,
+    proof_leaf_id,
+    universe_root_id,
+    servers_id
+);
--- a/tapdb/sqlc/queries/universe.sql
+++ b/tapdb/sqlc/queries/universe.sql
@@ -365,3 +365,99 @@ ON CONFLICT(namespace)
 SELECT namespace, asset_id, group_key, proof_type, allow_sync_insert, allow_sync_export
 FROM federation_uni_sync_config
 ORDER BY group_key NULLS LAST, asset_id NULLS LAST, proof_type;
+
+-- name: UpsertFederationProofSyncLog :one
+INSERT INTO federation_proof_sync_log as log (
+    status, timestamp, sync_direction, proof_leaf_id, universe_root_id,
+    servers_id
+) VALUES (
+    @status, @timestamp, @sync_direction,
+    (
+        -- Select the leaf id from the universe_leaves table.
+        SELECT id
+        FROM universe_leaves
+        WHERE leaf_node_namespace = @leaf_namespace
+            AND minting_point = @leaf_minting_point_bytes
+            AND script_key_bytes = @leaf_script_key_bytes
+        LIMIT 1
+    ),
+    (
+        -- Select the universe root id from the universe_roots table.
+        SELECT id
+        FROM universe_roots
+        WHERE namespace_root = @universe_id_namespace
+        LIMIT 1
+    ),
+    (
+        -- Select the server id from the universe_servers table.
+        SELECT id
+        FROM universe_servers
+        WHERE server_host = @server_host
+        LIMIT 1
+    )
+) ON CONFLICT (sync_direction, proof_leaf_id, universe_root_id, servers_id)
+DO UPDATE SET
+    status = EXCLUDED.status,
+    timestamp = EXCLUDED.timestamp,
+    -- Increment the attempt counter.
+    attempt_counter = CASE
+       WHEN @bump_sync_attempt_counter = true THEN log.attempt_counter + 1
+       ELSE log.attempt_counter
+    END
+RETURNING id;
+
+-- name: QueryFederationProofSyncLog :many
+SELECT
+    log.id, status, timestamp, sync_direction, attempt_counter,
+
+    -- Select fields from the universe_servers table.
+    server.id as server_id,
+    server.server_host,
+
+    -- Select universe leaf related fields.
+    leaf.minting_point as leaf_minting_point_bytes,
+    leaf.script_key_bytes as leaf_script_key_bytes,
+    mssmt_node.value as leaf_genesis_proof,
+    genesis.gen_asset_id as leaf_gen_asset_id,
+    genesis.asset_id as leaf_asset_id,
+
+    -- Select fields from the universe_roots table.
+    root.asset_id as uni_asset_id,
+    root.group_key as uni_group_key,
+    root.proof_type as uni_proof_type
+
+FROM federation_proof_sync_log as log
+
+JOIN universe_leaves as leaf
+    ON leaf.id = log.proof_leaf_id
+
+-- Join on mssmt_nodes to get leaf related fields.
+JOIN mssmt_nodes mssmt_node
+     ON leaf.leaf_node_key = mssmt_node.key AND
+        leaf.leaf_node_namespace = mssmt_node.namespace
+
+-- Join on genesis_info_view to get leaf related fields.
+JOIN genesis_info_view genesis
+     ON leaf.asset_genesis_id = genesis.gen_asset_id
+
+JOIN universe_servers as server
+    ON server.id = log.servers_id
+
+JOIN universe_roots as root
+    ON root.id = log.universe_root_id
+
+WHERE (log.sync_direction = sqlc.narg('sync_direction')
+           OR sqlc.narg('sync_direction') IS NULL)
+        AND
+      (log.status = sqlc.narg('status') OR sqlc.narg('status') IS NULL)
+        AND
+
+      -- Universe leaves WHERE clauses.
+      (leaf.leaf_node_namespace = sqlc.narg('leaf_namespace')
+           OR sqlc.narg('leaf_namespace') IS NULL)
+        AND
+      (leaf.minting_point = sqlc.narg('leaf_minting_point_bytes')
+           OR sqlc.narg('leaf_minting_point_bytes') IS NULL)
+        AND
+      (leaf.script_key_bytes = sqlc.narg('leaf_script_key_bytes')
+           OR sqlc.narg('leaf_script_key_bytes') IS NULL);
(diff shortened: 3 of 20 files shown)


universe/interface.go
```diff
@@ -84,6 +84,66 @@ func (i *Identifier) StringForLog() string {
...
+	proofTypeStr string) (Identifier, error) {
+
+	proofType, err := ParseStrProofType(proofTypeStr)
+	if err != nil {
+		return Identifier{}, err
+	}
+
+	// If the group key bytes are set, then we'll preferentially populate
+	// the universe ID with that and not the asset ID.
+	if groupKeyBytes != nil {
```
guggero, line 102: Should always use `len() != 0` with byte slices IMO, to avoid `nil` vs `empty slice` issue, especially if things come out of the DB. Same below with the asset ID bytes.

universe/interface.go
```diff
@@ -84,6 +84,66 @@ func (i *Identifier) StringForLog() string {
...
+	var assetID asset.ID
+	copy(assetID[:], assetIDBytes)
+
+	return Identifier{
+		AssetID:   assetID,
+		ProofType: proofType,
+	}, nil
+}
+
+// parseUserKey parses a group key from bytes, which can be in either the
```
guggero, line 130: nit: Godoc comment mismatch.

tapdb/sqlc/migrations/000012_universe_fed_proof_sync_log.up.sql
```diff
@@ -0,0 +1,36 @@
...
+
+    -- The ID of the universe that the proof leaf belongs to.
+    universe_root_id BIGINT NOT NULL REFERENCES universe_roots(id),
+
+    -- The ID of the server that the proof will be/was synced to.
+    servers_id BIGINT NOT NULL REFERENCES universe_servers(id)
+);
+
+-- Create a unique index on table federation_proof_sync_log
+CREATE UNIQUE INDEX federation_proof_sync_log_unique_index_proof_leaf_id_servers_id
```
guggero, line 30: I think this can be simplified by using `UNIQUE (sync_direction, proof_leaf_id, universe_root_id, servers_id)` in the table definition directly, then we don't have to name the index (and things look more similar to other table definitions we have). Or do we need a named index?

tapdb/sqlc/queries/universe.sql
```diff
@@ -370,4 +370,87 @@ ON CONFLICT(namespace)
...
+    status = EXCLUDED.status,
+    timestamp = EXCLUDED.timestamp,
+    -- Increment the attempt counter.
+    attempt_counter = CASE
+       WHEN @bump_sync_attempt_counter = true THEN log.attempt_counter + 1
+       ELSE log.attempt_counter
+    END
+RETURNING id;
+
+-- name: QueryFederationProofSyncLog :many
```
guggero, line 415: We should add some basic unit tests for the upsert and query to make sure they work as expected, since they look a bit more involved (especially with the `ON CONFLICT ... DO UPDATE`, that always tends to have surprising side effects, at least in my experience).

universe/auto_syncer.go
```diff
@@ -313,6 +452,51 @@ func (f *FederationEnvoy) syncer() {
 				continue
 			}
 
+			// After we've synced with the federation, we'll
+			// attempt to push out any pending proofs that we
+			// haven't yet completed.
+			ctxFetchLog, cancelFetchLog := f.WithCtxQuitNoTimeout()
+			syncDirection := SyncDirectionPush
+			logEntries, err :=
```
guggero, line 460: nit: extract `db := f.cfg.FederationDB` instead of breaking after the `:=` part for better readability?

universe/auto_syncer.go
```diff
@@ -313,6 +452,51 @@ func (f *FederationEnvoy) syncer() {
...
+				f.cfg.FederationDB.FetchPendingProofsSyncLog(
+					ctxFetchLog, &syncDirection,
+				)
+			if err != nil {
+				log.Warnf("unable to query pending push "+
+					"sync log: %w", err)
+				continue
+			}
+			cancelFetchLog()
```
guggero, line 469: We should also cancel right before the `continue` above. Or just move this cancel to directly after the DB call.

universe/auto_syncer.go
```diff
@@ -313,6 +452,51 @@ func (f *FederationEnvoy) syncer() {
...
+			//  log for a given proof leaf. Pending proof sync log
+			//  entries are only relevant for the set of servers
+			//  that existed at the time the log entry was created.
+			//  If a new server is added, then we should create a
+			//  new log entry for the new server.
+
+			for idx := range logEntries {
+				entry := logEntries[idx]
+
+				go func() {
```
guggero, line 482: Aren't we already in a goroutine here? So can't we just wait for these operations to finish synchronously?

guggero (commented):
Did a first pass, looks pretty good so far. Nice work!

universe/auto_syncer.go
ffranr: If at this point we had a lot of pending log entries that we needed to handle, blocking at this point would stop the federation envoy from accepting new push requests. I think that's the only motivator I can think of. What do you think?

guggero: Right, we would block new requests from coming in. But we aren't just logging something, we are pushing proofs to the universe. So we could run into a situation where we have a dependency or timing problem if something we rely on the universe knowing is not yet pushed because it's in a goroutine.
Also, `pushProofToFederation()` itself creates more goroutines, so we're spawning a lot of them potentially... I'd vote for blocking here instead.

ffranr: I agree with you that `pushProofToFederation` creates new go routines using `fn.ParSlice`. So I agree that we should remove the goroutine in the `for idx := range logEntries` loop.

> So we could run into a situation where we have a dependency or timing problem if something we rely on the universe knowing is not yet pushed because it's in a goroutine.

But at this point we're handling pending log entries kicked off by the federation envoy ticker. We've already waited until the next tick. We aren't populating proofs in the local universe we're pushing them to remote servers.

guggero: Okay, I see that the timing concern probably isn't an issue. But still, creating goroutines in a loop is usually a red flag. And I guess in this instance, if there is reason to do it asynchronously, there should only be a single goroutine that does the whole loop. Not a goroutine per loop entry.

tapdb/sqlc/migrations/000012_universe_fed_proof_sync_log.up.sql
ffranr: I thought a named index would make it easier to modify the constraint in a future SQL migration. Based on what I encountered here: https://github.com/lightninglabs/taproot-assets/pull/684#pullrequestreview-1754083528

@guggero What do you think of the migration point re named indexes?

universe/auto_syncer.go
ffranr: Sure no problem, will do. In a future PR I hope to break down each `case` in this `select` statement into three methods. This will simplify error handling (avoid the need to log everywhere rather than just return) and will save on indentation. I think it will also help with variable name shadowing.

universe/auto_syncer.go
ffranr: > But still, creating goroutines in a loop is usually a red flag.

Ok, I'll keep that in mind :+1:

universe/auto_syncer.go
ffranr: Good point.

tapdb/sqlc/migrations/000012_universe_fed_proof_sync_log.up.sql
guggero: Yeah, after reviewing that other PR, I think we should indeed start to use named indexes. Otherwise we have to re-create the whole table just to change an index...

ffranr:
I think, to keep this PR simple, I'll add the functionality (and tests) to cleanup stale entries from the proof sync log in a separate PR. I'm confident that this PR will add the necessary columns to the new log table to make that possible.

ffranr:
I think this would be a good idea where sql query unit testing is concerned:
```golang
package tapdb

import (
	"database/sql"
	"testing"

	"github.com/lightninglabs/taproot-assets/tapdb/sqlc"
	"github.com/lightningnetwork/lnd/clock"
)

// testDbStores is a helper struct that contains all the database stores that
// are needed for testing.
type testDbStores struct {
	FedDb *UniverseFederationDB

	MintingDb *AssetMintingStore

	AssetDb *AssetStore
	
	Db sqlc.Querier
}

// newTestDbStores creates a new set of test database stores.
func newTestDbStores(t *testing.T, clock clock.Clock) testDbStores {

	db := NewTestDB(t)

	dbTransactionExecutor := NewTransactionExecutor(db,
		func(tx *sql.Tx) UniverseServerStore {
			return db.WithTx(tx)
		},
	)

	txCreator := func(tx *sql.Tx) PendingAssetStore {
		return db.WithTx(tx)
	}
	activeTxCreator := func(tx *sql.Tx) ActiveAssetsStore {
		return db.WithTx(tx)
	}

	assetMintingDB := NewTransactionExecutor(db, txCreator)
	assetsDB := NewTransactionExecutor(db, activeTxCreator)

	return testDbStores {
		FedDb: NewUniverseFederationDB(dbTransactionExecutor, clock),
		MintingDb: NewAssetMintingStore(assetMintingDB),
		AssetDb: NewAssetStore(assetsDB, clock),
		Db: db,
	}
}
```

And then we can have access to each store for the same db in any unit test. And we wont need to extend functions like `newTestFederationDb` at each call site just to add a store that we need.

tapdb/sqlc/queries/universe.sql
ffranr: Unit tests added :+1:

tapdb/universe_federation.go
```diff
@@ -267,6 +299,261 @@ func (u *UniverseFederationDB) LogNewSyncs(ctx context.Context,
...
+	if err != nil {
+		return nil, err
+	}
+
+	return proofSyncLogs, nil
+}
+
+// parseSqlProofSyncLogEntry returns a parsed proof sync log entry given a DB
+// row.
+func parseSqlProofSyncLogEntry(ctx context.Context, entry ProofSyncLogEntry,
```
guggero, line 479: Related to the comment above: Since we're doing another DB call here, the function name is a bit misleading, as we're doing way more than just parsing (that's why I missed this in the first review iteration).

universe/auto_syncer.go
```diff
@@ -227,55 +233,178 @@ func (f *FederationEnvoy) syncServerState(ctx context.Context,
 	return nil
 }
 
-// pushProofToFederation attempts to push out a new proof to the current
-// federation in parallel.
-func (f *FederationEnvoy) pushProofToFederation(uniID Identifier, key LeafKey,
-	leaf *Leaf) {
+func (f *FederationEnvoy) pushProofToServer(ctx context.Context,
```
guggero, line 236: nit: Godoc missing here and below.

universe/auto_syncer.go
```diff
@@ -227,55 +233,178 @@ func (f *FederationEnvoy) syncServerState(ctx context.Context,
...
+	// Ensure that we have a pending sync log entry for this
+	// leaf and server pair. This will allow us to handle all
+	// pending syncs in the event of a restart or at a different
+	// point in the envoy.
+	_, err := f.cfg.FederationDB.UpsertFederationProofSyncLog(
+		ctx, uniID, key, addr, SyncDirectionPush,
+		ProofSyncStatusPending, true,
+	)
+	if err != nil {
+		log.Warnf("unable to log proof sync as pending: %v",
```
guggero, line 268: nit: log messages should be capitalized, here and below.

universe/auto_syncer.go
```diff
@@ -227,55 +233,178 @@ func (f *FederationEnvoy) syncServerState(ctx context.Context,
...
+		// If we are logging proof sync attempts, we will use the
+		// logged version of the push function.
+		if proofSyncLogged {
+			err := f.pushProofToServerLogged(
+				ctx, uniID, key, leaf, addr,
+			)
+			if err != nil {
+				log.Warnf("cannot push proof via logged "+
+					"server push: %v", err)
+				return nil
```
guggero, line 327: nit: can remove the `return nil` as we're returning just below anyway.

universe/auto_syncer.go
```diff
@@ -227,55 +233,178 @@ func (f *FederationEnvoy) syncServerState(ctx context.Context,
...
-			ctx, uniID, key, leaf,
-		)
+		// If we are not logging proof sync attempts, we will use the
+		// non-logged version of the push function.
+		err := f.pushProofToServer(ctx, uniID, key, leaf, addr)
 		if err != nil {
-			log.Warnf("cannot push proof to remote "+
-				"server(%v): %v", addr.HostStr(), err)
+			log.Warnf("cannot push proof: %v", err)
+			return nil
```
guggero, line 338: nit: same here re `return nil`.

tapdb/sqlc/queries/universe.sql
```diff
@@ -453,4 +453,24 @@ WHERE (log.sync_direction = sqlc.narg('sync_direction')
...
+    WHERE
+        (server_host = sqlc.narg('server_host')
+            OR sqlc.narg('server_host') IS NULL)
+)
+DELETE FROM federation_proof_sync_log
+WHERE
+    servers_id IN (SELECT id FROM selected_server_id) AND
+    (status = sqlc.narg('status')
+        OR sqlc.narg('status') IS NULL) AND
+    (timestamp >= sqlc.narg('timestamp')
```
guggero, line 473: nit: rename to `min_timestamp` since we do a greater than or equal comparison?

tapdb/sqlc/queries/universe.sql
```diff
@@ -453,4 +453,24 @@ WHERE (log.sync_direction = sqlc.narg('sync_direction')
...
+        OR sqlc.narg('timestamp') IS NULL) AND
+    (attempt_counter >= sqlc.narg('attempt_counter')
```
guggero, line 475: nit: same here, `min_attempt_counter`?

tapdb/test_sqlutils.go
```diff
@@ -0,0 +1,78 @@
...
+	"testing"
+	"time"
+
+	"github.com/lightninglabs/taproot-assets/tapdb/sqlc"
+	"github.com/lightningnetwork/lnd/clock"
+)
+
+// TestDbHandler is a helper struct that contains all the database stores that
+// are needed for testing.
+type TestDbHandler struct {
```
guggero, line 14: nit: just name this `DbHarness`?

universe/auto_syncer.go
ffranr: I did it this way because it makes it safer for any changes which add code after this `if` clause. I think my IDE collapses `if err != nil` clauses sometimes, which means I might miss the lack of a `return`. What do you think of that idea? Is the extra `return nil` dead code a bother?

universe/auto_syncer.go
ffranr: See: https://github.com/lightninglabs/taproot-assets/pull/698#discussion_r1414006833

tapdb/universe_federation.go
ffranr: What name should I use instead? `marshalSqlProofSyncLogEntry`?

I wish we could do `universe.ProofSyncLogEntry::FromDbType(...)`, that way it should be easy to find `FromDbType` on the type namespace.

tapdb/assets_store_test.go
```diff
@@ -240,14 +240,15 @@ func assertAssetEqual(t *testing.T, a, b *asset.Asset) {
...
 }
 
-// TestImportAssetProof tests that given a valid asset proof (mainly the final
-// snapshot information), we're able to properly import all the components on
-// disk, then retrieve the asset as if it were ours.
-func TestImportAssetProof(t *testing.T) {
-	t.Parallel()
+// testDbAddRandomAssetProof generates a random asset and corresponding proof
+// and inserts them into the given test database.
+func testDbAddRandomAssetProof(t *testing.T, ctx context.Context,
```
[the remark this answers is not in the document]
guggero: Well, it sounds like those test data generation code might be useful in other places too. So finding it is easier if it is centralized or a method on a harness that you might need to use anyway. But no strong opinion about either way. It's just usually if you have multiple functions that all take the same struct as a parameter (here the DB harness) that can be a sign that a method could make sense.

tapdb/universe_federation_test.go
```diff
@@ -112,6 +112,192 @@ func TestUniverseFederationCRUD(t *testing.T) {
...
+
+	// Check that the remaining log entry is as expected.
+	pendingEntry := pendingLogEntries[0]
+	require.Equal(t, servers[2], pendingEntry.ServerAddr)
+}
+
+func assertProofSyncLogLeafKey(t *testing.T, actualLeafKey universe.LeafKey,
+	logLeafKey universe.LeafKey) {
+
+	// TODO(ffranr): Can we really ignore the tweaked script key?
```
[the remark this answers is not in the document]
guggero: Ah, I wasn't sure if those were questions to the reviewer or just things you wanted to figure out before finalizing the PR.

> // TODO(ffranr): Can we really ignore the tweaked script key?

Yes, since the tweaked key is the derivation information of the script key, that is only ever known to the "owner" of an asset and is never serialized in a proof. So it is correct to blank it out before comparing it here.

> // TODO(ffranr): Do we need to check all the proof leaf fields?

Yes, I think it makes sense to also compare the other fields. Does something fail if we just do `require.Equal(t, actualLeaf, logLeaf)`?

universe/auto_syncer.go
ffranr: I've removed `return nil`. Don't think we can return error because of `fn.ParSlice`.

tapdb/assets_store_test.go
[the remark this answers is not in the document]
ffranr: I might leave this for a future PR. I hope that's ok with you. I think I will add methods on the DB harness which point to something in `assets_store_test.go` at least. Good point on the discovery front.

tapdb/universe_federation_test.go
```diff
@@ -30,15 +30,41 @@ func newTestFederationDb(t *testing.T,
 	return NewUniverseFederationDB(dbTxer, clock), db
 }
 
+// testDbAddRandomAssetProof is a helper function that will create server addresses
+// and add them to the database.
+func testDbAddRandomServerAddrs(t *testing.T, ctx context.Context,
```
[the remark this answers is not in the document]
ffranr: Discussed here: https://github.com/lightninglabs/taproot-assets/pull/698#discussion_r1416028772

tapdb/universe_federation_test.go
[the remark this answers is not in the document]
ffranr: When it comes to comparing proof leaf fields: the asset (decoded from the leaf genesis proof) does not contain the correct `GroupKey.Witness`.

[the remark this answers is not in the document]
guggero: Ah, that can also be set to `nil`, as it's basically just extracted from the asset witness (which probably only happens at DB level and not when parsing a proof). So that's okay to blank out before comparison.

guggero (approved):
Two small things that we should fix, otherwise LGTM :tada:

tapdb/sqlc/migrations/000013_universe_fed_proof_sync_log.up.sql
```diff
@@ -0,0 +1,36 @@
...
+-- in this table are specific to a given proof leaf, server, and sync direction.
+CREATE TABLE IF NOT EXISTS federation_proof_sync_log (
+    id BIGINT PRIMARY KEY,
+
+    -- The status of the proof sync attempt.
+    status TEXT NOT NULL CHECK(status IN ('pending', 'complete')),
+
+    -- The timestamp of when the log entry for the associated proof was last
+    -- updated.
+    timestamp TIMESTAMP NOT NULL,
```
jharveyb, line 11: Wondering if this should be a BIGINT and store Unix time like we did for universe events:

https://github.com/lightninglabs/taproot-assets/pull/396#discussion_r1263860576

Though I see we also have queries with comparisons to values using this `TIMESTAMP` type already?

universe/auto_syncer.go
```diff
@@ -219,55 +233,173 @@ func (f *FederationEnvoy) syncServerState(ctx context.Context,
...
+	}
 
-	ctx, cancel := f.WithCtxQuitNoTimeout()
-	defer cancel()
+	return nil
+}
+
+// pushProofToServerLogged attempts to push out a new proof to the target
+// server, and logs the sync attempt.
+func (f *FederationEnvoy) pushProofToServerLogged(ctx context.Context,
```
jharveyb, line 259: nit: can use `pushProofToServer` in here

universe/auto_syncer.go
```diff
@@ -219,55 +233,173 @@ func (f *FederationEnvoy) syncServerState(ctx context.Context,
...
+		ProofSyncStatusComplete,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("unable to query federation sync log: %w",
+			err)
+	}
+
+	// Filter out servers that we've already pushed to.
+	filteredFedServers := fn.Filter(fedServers, func(a ServerAddr) bool {
+		for idx := range logs {
```
jharveyb, line 377: nit: could be a lot of looping here if we have a lot of servers (or log entries). IIUC we could iterate over the `logs` once to build a `Set` of host strings, and then check for membership in that.

Not a required change though.

tapdb/universe_federation_test.go
```diff
@@ -98,6 +87,232 @@ func TestUniverseFederationCRUD(t *testing.T) {
...
+}
+
+// assertProofSyncLogLeaf asserts that a leaf derived from a proof sync log
+// entry is equal to a given universe leaf.
+func assertProofSyncLogLeaf(t *testing.T, actualLeaf universe.Leaf,
+	logLeaf universe.Leaf) {
+
+	if actualLeaf.GenesisWithGroup.GroupKey != nil {
+		actualLeaf.GenesisWithGroup.GroupKey.Witness = nil
+		actualLeaf.GenesisWithGroup.GroupKey.RawKey.PubKey = nil
```
jharveyb, line 302: the raw pubkey here is also populated via some other means I assume?

tapdb/universe_federation_test.go
```diff
@@ -98,6 +87,232 @@ func TestUniverseFederationCRUD(t *testing.T) {
...
+	for i := range servers {
+		server := servers[i]
+
+		// Set sync status to pending for all servers except the first.
+		syncStatus := universe.ProofSyncStatusPending
+		if i == 0 {
+			syncStatus = universe.ProofSyncStatusComplete
+		}
+
+		_, err := fedStore.UpsertFederationProofSyncLog(
```
jharveyb, line 131: Sould add one upsert here with `SyncDirectionPull` to exercise that as well.

jharveyb (commented):
Looks really good! Just one note on the unit test coverage.

tapdb/sqlc/migrations/000013_universe_fed_proof_sync_log.up.sql
guggero: I think it's okay as long as the comparison in the queries is just greater than or less than (and not some fancy calculations), since in the end, `TIMESTAMP` in SQLite is just a string, formatted on the Golang side. So it works for ordering and filtering but parsing it in a SQL query is near impossible to get done for both SQLite and Postgres. So if we need to do any calculation (e.g. creating time buckets or things like that), we should use a unix timestamp.

universe/auto_syncer.go
ffranr: Good point, thanks.

universe/auto_syncer.go
ffranr: I've made this change in the latest commits.

tapdb/universe_federation_test.go
ffranr: I don't think so. In the leaf derived from the log, the `RawKey.PubKey` field is not set.

I think that's OK because the distributed leaf doesn't need to include any data that was used to derive the group key. As far as I understand, the the `RawKey` is the pre-teaked group key. Do you agree?

tapdb/universe_federation_test.go
```diff
@@ -98,6 +87,284 @@ func TestUniverseFederationCRUD(t *testing.T) {
...
+	)
+	require.Equal(t, uniId.String(), completePushEntry.UniID.String())
+	require.Equal(t, int64(0), completePushEntry.AttemptCounter)
+
+	assertProofSyncLogLeafKey(
+		t, uniProof.LeafKey, completePushEntry.LeafKey,
+	)
+	assertProofSyncLogLeaf(t, *uniProof.Leaf, completePushEntry.Leaf)
+
+	// Retrieve all push sync status complete log entries.
```
jharveyb, line 206: nit: s/push/pull/g

tapdb/universe_federation_test.go
```diff
@@ -98,6 +87,284 @@ func TestUniverseFederationCRUD(t *testing.T) {
...
+	completePullLogEntries, err := fedStore.QueryFederationProofSyncLog(
```
jharveyb, line 207: Should be `SyncDirectionPull` here.

tapdb/universe_federation_test.go
```diff
@@ -98,6 +87,284 @@ func TestUniverseFederationCRUD(t *testing.T) {
...
+	// Check that the complete log entry is as expected.
+	completePullEntry := completePullLogEntries[0]
+
+	require.Equal(t, servers[0], completePullEntry.ServerAddr)
+	require.Equal(
+		t, universe.ProofSyncStatusComplete,
+		completePullEntry.SyncStatus,
+	)
+	require.Equal(
+		t, universe.SyncDirectionPush, completePullEntry.SyncDirection,
```
jharveyb, line 225: This should also be `SyncDirectionPull`.

tapdb/universe_federation_test.go
ffranr: Thank you!

PR #703, opened 2023-11-28
taprpc: add missing cli autogen docs

jharveyb:
Adds some missing tags in the proto files that prevented parts of our API docs page from showing CLI commands when one exists.

Right now this page, https://lightning.engineering/api-docs/api/taproot-assets/taproot-assets/fetch-asset-meta , shows no `Shell` example even though the command exists.

Also fixes some help comments.

I don't think we can easily add a lint for these comments since some RPC endpoints intentionally don't have a matching CLI cmd, like the PSBT endpoints.

### DIFF
--- a/cmd/tapcli/assets.go
+++ b/cmd/tapcli/assets.go
@@ -261,8 +261,8 @@ var finalizeBatchCommand = cli.Command{
 		},
 		cli.Uint64Flag{
 			Name: feeRateName,
-			Usage: "if set, the fee rate in sat/kw to use for the" +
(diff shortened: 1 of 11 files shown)


guggero (approved):
Nice, LGTM :tada:

ffranr (approved):
I'm not sure about `universe federation config info` vs just `universe federation config`. I can see that we do use `info` in one other place though.

guggero:
We should really fix that race unit test flake... @Roasbeef can you override merge (without queue) please?

Issue #694, opened 2023-11-21
Add test framework for SQL DB migrations


ffranr:
This project currently lacks a test framework for SQL database migration changes.

Our migrations are incremental and are stored in order here: `./tapdb/sqlc/migrations/*.sql`

Currently, in testing, we simply apply all migrations and then use the resultant DB. We should add the capability to apply a subset of the migration changes and then allow an opportunity for test code to examine the DB.

guggero:
I think we can make this quite simple for us: In `NewTestPostgresDB` (or `NewTestSqliteDB`), if an optional `initialDataFileName` parameter is specified, we would read that file as SQL and run the statements inside against the DB before creating the store (which then applies the migrations).
So if the initial data file contains a full SQL DB dump (including the table that the migrations library uses to keep track of what migrations have already been applied), then we can test certain incremental migrations.
The initial data files might be a bit hard to maintain manually, but ideally we create them from an actual DB once and then don't have to update them.
Maybe we need separate files for Postgres and SQLite, or maybe we can apply the same text search/replace mappings we use during the migrations themselves.

PR #707, opened 2023-11-29
tapdb: add minimalistic test framework for testing DB migrations

guggero:
This is an attempt at solving https://github.com/lightninglabs/taproot-assets/issues/694.
There are many ways to solve this, and this is IMO the most easy to implement one.

So this is mainly a proposal to get the discussion started. If we feel like we need something more involved, we can discuss based on this example code.

I mainly had this PR in mind when coming up with the design for this: https://github.com/lightninglabs/taproot-assets/pull/684
There we need to have a way to insert existing data into the DB before applying the latest migration.

Here are the pros/cons of the current approach:

Pros:
 - Simple to implement, understand, use and maintain
 - Should allow us to fully test data manipulation migrations
 - No need to carry along multiple versions of Golang model code
 - No impact on production code or run time

Cons:
 - Not very easy to test schema manipulation migrations, since Golang model is always the very latest version
 - If we need to test changes to the actual schema, we need to do it with raw SQL queries
 - Need to manually create test data SQL code

Fixes https://github.com/lightninglabs/taproot-assets/issues/694


bhandras:
Super cool!

ffranr (approved):
Migration changes make sense to me. Nice PR, very clear! I'm not sure where you're going with the last commit.

guggero:
> e.g if we already have 10 migrations in the DB and a new PR is trying to add n.11, are we gonna test that migration against all possible states a tapd could be upgrading from?

The idea is that we have a way to test migrations that do data manipulations. Those are quite hard to test with just a unit test that looks at the state _after_ all migrations have been applied. One of those PRs is https://github.com/lightninglabs/taproot-assets/pull/684 (which was taken as the inspiring PR to influence some of the design in this PR).

jharveyb (approved):
Very neat! I think this would have worked for validating the latest changes to the `anchor_txid` view, and repro'ing that user issue.

Re: test data; I imagine we could generate a 'golden' DB state from one of itests and use that for testing future migrations. Feels like using this in future tests shouldn't be too much work.

Needs rebase.

tapdb/assets_store.go
```diff
@@ -861,7 +861,7 @@ func fetchAssetsWithWitness(ctx context.Context, q ActiveAssetsStore,
 	// First, we'll fetch all the assets we know of on disk.
 	dbAssets, err := q.QueryAssets(ctx, assetFilter)
 	if err != nil {
-		return nil, nil, fmt.Errorf("unable to read db assets: %v", err)
+		return nil, nil, fmt.Errorf("unable to read db assets: %w", err)
```
Roasbeef, line 864: 👍

Roasbeef (approved):
Was a bit skeptical initially, but I think this nicely resolves issues that would pop up in a _purely_ unit test approach, namely: we wouldn't be able to programmatically create older versions of certain tables as the Go code is generated from the _latest_ version of the migration. 

It's also the case that all the generated code already has SQL fragments in them, and you can look at the older versions to extract raw SQL used for the insertion files. 

LGTM 🥕

PR #708, opened 2023-11-29
workflows: fix release name

ffranr:
Release name should be `Taproot Assets v...` rather than `tapd v...`.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -39,7 +39,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           tag_name: ${{ env.RELEASE_VERSION }}
-          name: tapd ${{ env.RELEASE_VERSION }}
+          name: ${{ env.RELEASE_VERSION }}
           draft: true
           prerelease: false
           files: taproot-assets-${{ env.RELEASE_VERSION }}/*


.github/workflows/release.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           tag_name: ${{ env.RELEASE_VERSION }}
-          name: tapd ${{ env.RELEASE_VERSION }}
+          name: Taproot Assets ${{ env.RELEASE_VERSION }}
```
guggero, line 42: Shouldn't it just be the version name? Taking https://github.com/lightninglabs/taproot-assets/releases/tag/v0.3.1 as an example?

guggero: Or, I guess the tag name plus the `-alpha` suffix. Which probably will be wrong for the RC versions again... Perhaps we should also use `rc2-alpha` in `version.go` to make things more uniform...

GeorgeTsagk: > Shouldn't it just be the version name?

It seems like we do it both ways in other repos. In [`loop`](https://github.com/lightninglabs/loop/releases) for example we only have the semver, but in [`lnd`](https://github.com/lightningnetwork/lnd/releases) there is also the "lnd" prefix

I would suggest staying consistent with what we've done so far in taproot-assets repo and only include the semver as the release name
```suggestion
          name: ${{ env.RELEASE_VERSION }}
```

ffranr: I like George's suggestion because it's easiest to keep consistent across all LL projects. Less work in copy and pasting the release workflow yaml.

guggero (approved):

jharveyb (approved):

PR #713, opened 2023-12-04
rpcserver+tapscript: err shadowing fixes

jharveyb:
This removes some instances of error shadowing, which caught one minor bug where we would have returned a `nil` error after an error was `non-nil`.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -1050,8 +1050,6 @@ func (r *rpcServer) QueryAddrs(ctx context.Context,
 func (r *rpcServer) NewAddr(ctx context.Context,
 	req *taprpc.NewAddrRequest) (*taprpc.Addr, error) {
 
-	var err error
-
 	// Parse the proof courier address if one was provided, otherwise use
 	// the default specified in the config.
 	courierAddr := r.cfg.DefaultProofCourierAddr
@@ -1089,7 +1087,7 @@ func (r *rpcServer) NewAddr(ctx context.Context,
 	rpcsLog.Infof("[NewAddr]: making new addr: asset_id=%x, amt=%v",
 		assetID[:], req.Amt)
 
-	err = r.checkBalanceOverflow(ctx, &assetID, nil, req.Amt)
+	err := r.checkBalanceOverflow(ctx, &assetID, nil, req.Amt)
 	if err != nil {
 		return nil, err
 	}
@@ -1376,7 +1374,7 @@ func (r *rpcServer) marshalProof(ctx context.Context, p *proof.Proof,
 	var exclusionProofs [][]byte
 	for _, exclusionProof := range p.ExclusionProofs {
 		var exclusionProofBuf bytes.Buffer
-		err := exclusionProof.Encode(&exclusionProofBuf)
+		err = exclusionProof.Encode(&exclusionProofBuf)
 		if err != nil {
 			return nil, fmt.Errorf("unable to encode exclusion "+
 				"proofs: %w", err)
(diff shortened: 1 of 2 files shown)


rpcserver.go
```diff
@@ -456,7 +456,7 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 
 		// If the asset meta field was specified, then the data inside
 		// must be valid. Let's check that now.
-		if err := seedling.Meta.Validate(); err != nil {
+		if err = seedling.Meta.Validate(); err != nil {
```
guggero, line 459: I don't really understand why we don't want shadowing here? We tend to do this everywhere and as far as I'm aware there shouldn't be any side effect.

rpcserver.go
```diff
@@ -585,7 +586,7 @@ func (r *rpcServer) ListBatches(_ context.Context,
 		}
 
 	case len(req.GetBatchKeyStr()) > 0:
-		batchKeyBytes, err := hex.DecodeString(req.GetBatchKeyStr())
+		batchKeyBytes, err = hex.DecodeString(req.GetBatchKeyStr())
```
guggero, line 589: IMO it's fine to shadow the `err` variable here, as we check it within the same scope. It would be different if we had a shared `if err != nil` after the `switch`. So this moves the `batchKeyBytes` outside of this scope which IMO just makes it a bit harder to understand (since it's not used anywhere else).

rpcserver.go
```diff
@@ -3459,34 +3458,38 @@ func UnmarshalOutpoint(outpoint string) (*wire.OutPoint, error) {
 // unmarshalLeafKey un-marshals a leaf key from the RPC form.
 func unmarshalLeafKey(key *unirpc.AssetKey) (universe.LeafKey, error) {
 	var (
-		leafKey universe.LeafKey
-		err     error
+		leafKey        universe.LeafKey
```
guggero, line 3461: Actually, I think we should do the opposite here and just declare the `leafKey` here, since that's all we're interested in after the `switch`. All other variables can be inline (which solves the `err` shadowing too, as that wouldn't be declared here anymore either).

rpcserver.go
```diff
@@ -3459,34 +3458,38 @@ func UnmarshalOutpoint(outpoint string) (*wire.OutPoint, error) {
...
 
 		leafKey.ScriptKey = &asset.ScriptKey{
-			PubKey: pubKey,
+			PubKey: scriptKey,
 		}
 
 	case key.GetScriptKeyStr() != "":
-		scriptKeyBytes, sErr := hex.DecodeString(key.GetScriptKeyStr())
-		if sErr != nil {
+		scriptKeyBytes, err = hex.DecodeString(key.GetScriptKeyStr())
```
guggero, line 3481: I guess this is the only problem that we should really fix, since before this change we assigned `sErr` but returned `err`.

guggero (commented):
Not really sure about the reasons for most of the diff in this PR, but I'm fine either way. There is at least one area where we have a shadowing problem but my proposed solution would look differently (see inline comment).

rpcserver.go
jharveyb: Yeah, I guess shadowing is OK given that we always check errs right after assignment.

rpcserver.go
```diff
@@ -3509,7 +3512,7 @@ func unmarshalLeafKey(key *unirpc.AssetKey) (universe.LeafKey, error) {
 	case key.GetOutpoint() != nil:
 		op := key.GetOp()
 
-		hash, err := chainhash.NewHashFromStr(op.HashStr)
+		hash, err = chainhash.NewHashFromStr(op.HashStr)
```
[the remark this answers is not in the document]
guggero: Interesting... Nice catch!

Roasbeef (approved):
LGTM 🚆

PR #716, opened 2023-12-04
Add enhanced prometheus metrics for tapd

GeorgeTsagk:
## Description

This PR polishes the previous prometheus framework and adds a few new collectors in order to monitor tapd instances.


GeorgeTsagk:
Not for reviewers: Please take some time to consider what other metrics would be nice-to-have.
We already expose a variety of go runtime / GRPC related metrics, so in this PR we enhance this set with application-specific metrics to further improve our insight on a running tapd

monitoring/config.go
```diff
@@ -17,6 +20,9 @@ type PrometheusConfig struct {
 	// generic RPC metrics to monitor the health of the service.
 	RPCServer *grpc.Server
 
+	// UniverseStats ...
```
bhandras: nit: missing godoc

monitoring/asset_collector.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if a.cfg.UniverseStats == nil {
+		log.Error("UniverseStats is nil")
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), dbTimeout)
+	defer cancel()
+
+	universeStats, err := a.cfg.UniverseStats.AggregateSyncStats(ctx)
```
bhandras, line 100: Is there a need at all for such collector? Couldn't we just use SQL for the dash and create the data on demand for the requested time interval?

monitoring/asset_balances_collector.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if a.cfg.AssetStore == nil {
+		log.Error("asset store is nil")
+		return
+	}
+
+	ctxdb, cancel := context.WithTimeout(context.Background(), dbTimeout)
+	defer cancel()
+
+	assets, err := a.cfg.AssetStore.FetchAllAssets(ctxdb, false, false, nil)
```
bhandras, line 79: I think this is a bad idea as the number of assets will always just go up and this query will put an extra load on the DB as well as might just timeout if takes long. Perhaps better to build a specialized query that joins the asset and managed utxo tables and returns exactly what we want. Even then I'd say it's better to do this on-demand if only for a grafana dash and not for alerting.

GeorgeTsagk: `FetchAllAssets` fetches the assets owned by the daemon. It basically gives you all your balances for all your owned assets

GeorgeTsagk: > I think this is a bad idea as the number of assets will always just go up and this query will put an extra load on the DB as well as might just timeout if takes long. 

So yeah if you're hoarding an enormous amount of different asset types you could run into this, but don't think it's reasonable to think of that possibility now

monitoring/asset_collector.go
GeorgeTsagk: > and create the data on demand for the requested time interval?

How would you gain access to past states?

Prometheus scrapes on a fixed interval, who would do the scraping with manual SQL on the dash?

monitoring/config.go
GeorgeTsagk: It's added later on 6de7742db05861980b46079ccdc93e620a58b2a5 , could fixup?

monitoring/asset_collector.go
bhandras: Looking at the query here it makes sense to scrape as this doesn't look very expensive and provides a better user experience as users don't need to write complicated SQL queries.

The alternative would be to use the `event_time` in the `universe_events` table as the time source and create a query generating the same stats bucketed by time intervals. (This could be accomplished using windowing and the `PARTITION BY` clause). Bucketing is useful to provide a level of detail in Grafana.

monitoring/asset_balances_collector.go
GeorgeTsagk: Yeah if we want to maximise performance over everything we'd add something like `FetchTagsOutpointsAmountsOfAssets` but kind of seems like polluting the store interface. We are using 3 of the most important fields already returned by `FetchAllAssets` so the justification for custom store method is bit weaker

what do others think @ffranr @guggero

guggero: Since we're passing in `false` for `includeSpent`, this will only return assets owned by the daemon that are currently not spent. So this number shouldn't increase massively and IMO it should be fine.
I agree that we should try to keep the number of queries to a reasonable count as they are rather complex in some cases and hard to maintain (and test) if they have a lot of joins.

GeorgeTsagk:
Last push adds a new commit which introduces the `garden_collector.go`

monitoring/garden_collector.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Collect is called by the Prometheus registry when collecting metrics.
+//
+// NOTE: Part of the prometheus.Collector interface.
+func (a *gardenCollector) Collect(ch chan<- prometheus.Metric) {
+	a.collectMx.Lock()
+	defer a.collectMx.Unlock()
+
+	if a.cfg == nil {
```
bhandras, line 63: These checks could be done in `newGardenCollector` too, otherwise it's easy to just deploy something that'll not collect (but rather just error log).

monitoring/garden_collector.go
```diff
@@ -0,0 +1,120 @@
...
+
+	for _, batch := range batches {
+		state := batch.State()
+
+		switch {
+		case state == tapgarden.BatchStatePending ||
+			state == tapgarden.BatchStateFrozen ||
+			state == tapgarden.BatchStateCommitted ||
+			state == tapgarden.BatchStateBroadcast ||
+			state == tapgarden.BatchStateConfirmed:
```
bhandras, line 92: Why not just check for `BatchStatePending`? (and remove the `if` below).

bhandras (commented):
Also ptal at the linter error.

monitoring/garden_collector.go
GeorgeTsagk: So the gauge value may only be updated while state is `BatchStatePending`. The rest of the states are not final, and they may not change the value of the gauge so I need to just register them without changing the gauge (I can't do `len(batch.Seedlings)` again as the array is emptied). The reason I register them is because these states mean that the batch is still "in-flight"

monitoring/garden_collector.go
GeorgeTsagk: Also did this for the rest of the collectors

GeorgeTsagk:
@bhandras PTAL, also merged a lot of commits to a total of 2

monitoring/asset_collector.go
```diff
@@ -0,0 +1,125 @@
...
+	registry *prometheus.Registry
+
+	gauges map[string]prometheus.Gauge
+}
+
+func newUniverseStatsCollector(cfg *PrometheusConfig,
+	registry *prometheus.Registry) *universeStatsCollector {
+
+	if cfg == nil {
+		log.Error("cfg is nil")
```
bhandras, line 35: Instead of logging, just return an `fmt.Errorf` which will then prevent the daemon from starting (instead of logging something we'd skip over). (For all other checks as well).

GeorgeTsagk: that's what I'm currently doing by returning `nil` (caller checks if returned struct is `nil`, if yes it errors out)

GeorgeTsagk: can switch to `error` ofcs if it's more readable

monitoring/prometheus.go
```diff
@@ -54,28 +47,42 @@ func (p *PrometheusExporter) Start() error {
...
+	uniStatsCollector, err := newUniverseStatsCollector(p.config, p.registry)
+	if err != nil {
+		return err
+	}
+	p.registry.MustRegister(uniStatsCollector)
 
-	// Next, we'll attempt to register all our metrics. If we fail to
-	// register ANY metric, then we'll fail all together.
-	if err := p.registerMetrics(); err != nil {
+	assetBalancesCollecor, err := newAssetBalancesCollector(p.config, p.registry)
```
guggero, line 64: nit: line length.

guggero (approved):
tACK, LGTM :tada:

monitoring/asset_balances_collector.go
```diff
@@ -0,0 +1,112 @@
...
+
+func newAssetBalancesCollector(cfg *PrometheusConfig,
+	registry *prometheus.Registry) (*assetBalancesCollector, error) {
+
+	if cfg == nil {
+		return nil, errors.New("asset collector prometheus cfg is nil")
+	}
+
+	if cfg.AssetStore == nil {
+		log.Error("asset store is nil")
```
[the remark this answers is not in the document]
bhandras: +1, no need for the xtra log.

bhandras (approved):
LGTM, pending Oli's comments! 🎉

PR #719, opened 2023-12-05
multi: attempt to fix issue of high number of TCP connections by explicitly closing connections and implementing an idle timer

Roasbeef:
On the instance we've run, we've been seeing a very high number of active TCP connections at any given time. After tweaking some Aperture related settings, it seems that the issue is with our own gRPC server, and the way we make outbound connections. 

This PR implements to fixes to attempt to mitigate this issue:
   * Close out any outbound connections we make (courier, gRPC syncing, etc) once we're done. 
   * Add a default idle timer that'll close out connections that haven't received any requests/responses after an interval of time.

### DIFF
--- a/proof/courier.go
+++ b/proof/courier.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/davecgh/go-spew/spew"
 	"github.com/lightninglabs/lightning-node-connect/hashmailrpc"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/fn"
@@ -66,6 +67,9 @@ type Courier interface {
 	// SetSubscribers sets the set of subscribers that will be notified
 	// of proof courier related events.
 	SetSubscribers(map[uint64]*fn.EventReceiver[fn.Event])
+
+	// Close stops the courier instance.
+	Close() error
 }
 
 // CourierAddr is a fully validated courier address (including protocol specific
@@ -223,6 +227,7 @@ func (h *UniverseRpcCourierAddr) NewCourier(_ context.Context,
 		backoffHandle: backoffHandle,
 		transfer:      cfg.TransferLog,
 		subscribers:   subscribers,
+		rawConn:       conn,
 	}, nil
 }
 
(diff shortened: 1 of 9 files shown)


guggero (commented):
Very nice! I think this should fix our high number of connection issue.
Some CI steps fail and I have a question around logging, otherwise this looks good!

proof/courier.go
```diff
@@ -928,6 +945,11 @@ func (h *HashMailCourier) publishSubscriberEvent(event fn.Event) {
 	}
 }
 
+// Close closes the underlying connection to the hashmail server.
+func (h *HashMailCourier) Close() error {
+	return h.mailbox.Close()
```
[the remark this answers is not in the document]
Roasbeef: Opted to have the logging happen in this routine.

Roasbeef:
PTAL!

proof/courier.go
```diff
@@ -467,6 +478,8 @@ func (h *HashMailBox) RecvAck(ctx context.Context, sid streamID) error {
 
 // CleanUp atempts to tear down the mailbox as specified by the passed sid.
 func (h *HashMailBox) CleanUp(ctx context.Context, sid streamID) error {
+	defer h.rawConn.Close()
```
guggero, line 481: Non-blocking: I guess my request was less about actually logging the error but rather logging it instead of having it as a return value in the interface. Because swallowing the error here does show up as a code smell issue in my IDE (and will probably be flagged by the linter if we decide to turn on some of the rules we have in `lnd`).

guggero (approved):
Nice, LGTM :tada:

proof/courier.go
GeorgeTsagk: Maybe over-engineering:

In such use-cases we could try returning a named variable
```golang
func (h *HashMailBox) CleanUp(ctx context.Context, sid streamID) (err error) {
```
and assign to `err` the error we want to return within the body.
Then our `defer` would look like this
```golang
defer func(){
    if err != nil {
        err = h.rawConn.Close()
    }
}()
```

GeorgeTsagk: Or yeah we could just omit returning `error`. Do we care about consistency on the `Close` func signature?

server.go
```diff
@@ -269,6 +271,12 @@ func (s *Server) RunUntilShutdown(mainErrChan <-chan error) error {
 	serverOpts = append(serverOpts, rpcServerOpts...)
 	serverOpts = append(serverOpts, ServerMaxMsgReceiveSize)
 
+	keepAliveParams := keepalive.ServerParameters{
+		MaxConnectionIdle: time.Minute * 2,
```
GeorgeTsagk, line 275: I expected to see smth like 15 - 30 sec here. Is this like a widely accepted default-ish value?

GeorgeTsagk (approved):
Looks Good! :spider_web:

server.go
Roasbeef: Just kinda chose an initial value tbh...

Roasbeef:
itest seems to be failing consistently, will dig in.

proof/courier.go
guggero, line 481: We need to move this `h.rawConn.Close()` out of the `CleanUp()` method, since that is being called for two mailboxes (sender+receiver). So it needs to be done in `DeliverProof` as a `defer` before cleaning up the mailboxes.

guggero (commented):
Found the issue with the itest, see inline comment.

PR #720, opened 2023-12-05
.github/workflows: ensure release tag version string is valid SemVer

ffranr:
This commit adds a check to this project's GitHub release workflow to ensure that the release version string is SemVer compliant.


ffranr:
@guggero had the idea of adding a `make release-tag` command. The command would derive the correct release tag from the version specified in `version.go`. Then, it could check that the version string is SemVer compliant. And then finally create the git tag release tag.

We could then also compare the output of `make release-tag` to the git tag in the release workflow as a final consistency check.

I'll update this PR with that command.

scripts/get-release-tag-str.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+# This script derives a version string from the version fields in a target Go
+# file. It also checks if the derived version string is SemVer compliant.
+
+# get_git_tag_version_string reads the version fields from the given file and
+# constructs a git tag version string.
+get_git_tag_version_string() {
```
guggero, line 8: Doesn't bash or the command line only really know strings? IMO the "_string" suffix of the function and the "-str" suffix of the file are a bit redundant.

Makefile
```diff
@@ -115,6 +115,18 @@ release:
 	$(VERSION_CHECK)
 	./scripts/release.sh build-release "$(VERSION_TAG)" "$(BUILD_SYSTEM)" "$(RELEASE_TAGS)" "$(RELEASE_LDFLAGS)"
 
+release-tag:
+	@$(call print, "Adding release tag.")
+
+	@version_file_path="version.go" && \
+	tag=$$(./scripts/get-release-tag-str.sh $$version_file_path); \
```
guggero, line 122: This won't work, since we echo other stuff in the script. Should we just move all the logic into the script and call it `create-release-tag.sh`?

guggero: Or to re-use the version parsing script, just have a second script that creates the tag itself? Just because the script syntax in the Makefile itself isn't as portable as `bash` I think.

guggero (commented):
Nice, looks pretty good! Just a couple of suggestions to fix the actual make target.

Makefile
ffranr: I think it does work, I may be misunderstanding what you mean:
```shell
user@laptop:~/dev/taproot-assets$ make release-tag
 Adding release tag.
Reading version fields from version.go
Derived git tag name: v0.5.2
Git tag name version string is SemVer compliant
Adding git tag: v0.5.2
user@laptop:~/dev/taproot-assets$ git log
commit 781d1609316193b246988c5e368dbd96f00e92b5 (HEAD -> validate-release-version, tag: v0.5.2)
```

In the script I write log messages to stderr. Only the git tag name goes into stdout.

guggero: Oops, I missed the `>&2` indeed. Sorry about that. Looks good to me then.

guggero (approved):
Cool, looking forward to the standardized tagging procedure in all our projects :100:

version.go
```diff
@@ -162,13 +174,33 @@ func semanticVersion() string {
...
+	// hyphen called for by the semantic versioning spec is automatically
+	// appended and should not be contained in the pre-release string.
+	// The pre-release version is not appended if it contains invalid
+	// characters.
 	preRelease := normalizeVerString(AppPreRelease, semanticAlphabet)
-	if preRelease != "" {
-		version = fmt.Sprintf("%s-%s", version, preRelease)
+
+	// Append any status and pre-release strings to the version string.
+	switch {
```
jharveyb, line 191: Bit confused on this switch, do we ever want appStatus without preRelease? From the tapd tags it looks like we only do both or neither, ex. `v0.3.2`, `v0.3.0-alpha.rc3`, but no `v0.3.0-alpha` nor `v0.3.0-rc3`.

Or is this to also match a different pattern used for a different project?

jharveyb (commented):
Great PR to improve our tooling and release flow!

Some nits and clarifying comments but looks close.

scripts/get-git-tag-name.sh
```diff
@@ -0,0 +1,74 @@
...
+  local file_path="$1"
+
+  # Check if the file exists
+  if [ ! -f "$file_path" ]; then
+      echo "Error: File not found at $file_path" >&2
+      exit 1
+  fi
+
+  # Read and parse the version fields.
+  local app_major=$(grep -oP 'AppMajor\s*uint\s*=\s*\K\d+' "$file_path")
```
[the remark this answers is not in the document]
ffranr: The variables could be auto-formatted as a block:
```
	AppMajor  uint = 0
	AppMinor  uint = 3
	AppPatch  uint = 2
	AppStatus      = "alpha"
```

which adds a variable number of white spaces everywhere except for after the `=`.

Broadly, my idea here was to try to make the matching as robust as possible. With this idea of matching any legitimate golang variable deceleration of a given name, rather than any particular formatting.

Should I make the matching more specific?

scripts/get-git-tag-name.sh
```diff
@@ -0,0 +1,74 @@
...
+  local app_minor=$(grep -oP 'AppMinor\s*uint\s*=\s*\K\d+' "$file_path")
+  local app_patch=$(grep -oP 'AppPatch\s*uint\s*=\s*\K\d+' "$file_path")
+  local app_status=$(grep -oP 'AppStatus\s*=\s*"([^"]*)"' "$file_path" | cut -d '"' -f 2)
```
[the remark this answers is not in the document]
ffranr: Yes, that sounds like a good idea to me.

Makefile
```diff
@@ -115,6 +115,18 @@ release:
...
+	@$(call print, "Adding release tag.")
+
+	@version_file_path="version.go" && \
+	tag=$$(./scripts/get-git-tag-name.sh $$version_file_path); \
+	exit_status=$$?; \
+	if [ $$exit_status -ne 0 ]; then \
+		echo "Script encountered an error with exit status $$exit_status."; \
+	fi; \
+	echo "Adding git tag: $$tag"; \
+	git tag -s -a "$$tag" -m "Release $$tag";
```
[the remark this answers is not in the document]
ffranr: I think that the git tag message is a git artefact only and isn't used in a github release.

Here we create the git tag message as part of the `release-tag` make command which is release specific. The message is just a bit of extra documentation. The `$$tag` is probably redundant. We could possibly expand the message to say that the tag was created using the `release-tag` command? I think I'll make that change.  Let me know what you think.

scripts/get-git-tag-name.sh
```diff
@@ -0,0 +1,74 @@
...
+  fi
+
+  echo "$tag_name"
+}
+
+# check_git_tag_name checks if the given git tag name is a valid SemVer
+# compliant version string.
+check_git_tag_name() {
+  local tag_name="$1"
+  semver_regex="^v([0-9]+)\.([0-9]+)\.([0-9]+)(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+)?$"
```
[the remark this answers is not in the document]
ffranr: The regex that you're referring to wasn't specific enough. It allowed leading zeros for example. I've generated a better regex using AI. And also documentation.

version.go
ffranr: > Or is this to also match a different pattern used for a different project?

I think we should make sure that possibility is available to us. I don't think we have a strict org wide rule around the use of status without pre-release. `v0.3.0-alpha` seems like a reasonable tag to me for instance.

version.go
```diff
@@ -162,13 +174,33 @@ func semanticVersion() string {
...
 	version := fmt.Sprintf("%d.%d.%d", AppMajor, AppMinor, AppPatch)
 
-	// Append pre-release version if there is one. The hyphen called for
-	// by the semantic versioning spec is automatically appended and should
-	// not be contained in the pre-release string. The pre-release version
+	// If defined, we will now sanitise the release status string. The
+	// hyphen called for by the semantic versioning spec is automatically
+	// appended and should not be contained in the status string. The status
 	// is not appended if it contains invalid characters.
+	appStatus := normalizeVerString(AppStatus, semanticAlphabet)
```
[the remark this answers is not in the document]
ffranr: I agree with you on removing `.` and `-`.

I think that the idea behind the semantic alphabet is to be check that the version substrings are compatible with the semver standard, which is broader than what we use. The broader standard does allow digits and upper characters in app status.

I'll try to make new alphabets relevant to what we might want in our org.

scripts/get-git-tag-name.sh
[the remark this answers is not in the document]
jharveyb: Mm that makes sense, didn't think of these spacing changes `gofmt` sometimes makes.

scripts/get-git-tag-name.sh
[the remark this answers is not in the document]
jharveyb: I'm wondering also if there is a golang tagging one we could find somewhere?

version.go
jharveyb: K, I thought leaving out the `rc1` suffix when using `alpha` or `beta` is what caused issues previously? I don't recall exactly which tag caused those though.

guggero: What caused the issue previously was leaving out `-alpha` while just using an `rc` tag (e.g. `0.3.2-rc2`), because then the version check of the release script didn't work as it just did a substring comparison with what the app version string reported (`v0.3.2-alpha.rc2`).
So I think we should also update the `check_tag_correct` function in `scripts/release.sh` to use the new `get-git-tag-name.sh` script and compare it to the checked out tag.

And to answer your other question, just having the status is what we want for non-RC releases in all other projects than this one (e.g. lnd).

version.go
```diff
@@ -162,13 +184,33 @@ func semanticVersion() string {
...
-		version = fmt.Sprintf("%s-%s", version, preRelease)
+
+	// Append any status and pre-release strings to the version string.
+	switch {
+	case appStatus != "" && preRelease != "":
+		version = fmt.Sprintf(
+			"%s-%s.%s", version, appStatus, preRelease,
+		)
+	case appStatus != "":
+		version = fmt.Sprintf(
```
guggero, line 209: nit: fits on one line, here and below.

Makefile
```diff
@@ -115,6 +115,18 @@ release:
...
+	@$(call print, "Adding release tag.")
+
+	@version_file_path="version.go" && \
+	tag=$$(./scripts/get-git-tag-name.sh $$version_file_path); \
+	exit_status=$$?; \
+	if [ $$exit_status -ne 0 ]; then \
+		echo "Script encountered an error with exit status $$exit_status."; \
+	fi; \
+	echo "Adding git tag: $$tag"; \
+	git tag -s -a "$$tag" -m "Tag generated using command \`make release-tag\`.";
```
guggero, line 128: I think the `-a` is implied by `-s` and `-m`.

guggero: Also, should we move the "$$tag" to the very end? For a moment I thought it was the value for `-a` which I then had to look up first.

scripts/get-git-tag-name.sh
```diff
@@ -0,0 +1,122 @@
...
+  app_major=$(grep -oP 'AppMajor\s*uint\s*=\s*\K\d+' "$file_path")
+
+  local app_minor
+  app_minor=$(grep -oP 'AppMinor\s*uint\s*=\s*\K\d+' "$file_path")
+
+  local app_patch
+  app_patch=$(grep -oP 'AppPatch\s*uint\s*=\s*\K\d+' "$file_path")
+
+  local app_status
+  app_status=$(grep -oP 'AppStatus\s*=\s*"([a-z]*)"' "$file_path" | cut -d '"' -f 2)
```
guggero, line 29: We could use `grep -oP 'AppStatus\s*=\s*"\K([a-z]*)'` here to avoid needing to use `cut`.

scripts/get-git-tag-name.sh
```diff
@@ -0,0 +1,122 @@
...
+
+  local app_pre_release
+  app_pre_release=$(grep -oP 'AppPreRelease\s*=\s*"([a-z0-9]*)"' "$file_path" | cut -d '"' -f 2)
```
guggero, line 32: This could be simplified to `grep -oP 'AppPreRelease\s*=\s*"\K([a-z0-9]*)'`

scripts/get-git-tag-name.sh
```diff
@@ -0,0 +1,122 @@
...
+
+  # Parse the GitTagIncludeStatus field.
+  gitTagIncludeStatus=false
```
guggero, line 35: nit: use snake_case for this variable to match any other variable in the file? Also, do we need the `local` prefix?

scripts/get-git-tag-name.sh
```diff
@@ -0,0 +1,122 @@
...
+      echo "Error: GitTagIncludeStatus is not present in the file."
+      exit 1
+  fi
+
+  # Construct the git tag name with conditional inclusion of app_status and
+  # app_pre_release
+  tag_name="v${app_major}.${app_minor}.${app_patch}"
+
+  # Append app_status if gitTagIncludeStatus is true and app_status is not empty
+  if [ "$gitTagIncludeStatus" = true ] && [ -n "$app_status" ]; then
```
guggero, line 51: This doesn't match what we're doing in `version.go`...
For

```
	AppMajor uint = 0
	AppMinor uint = 3
	AppPatch uint = 2
	AppStatus = "alpha"
	AppPreRelease = "rc2"
```
the shell script would produce `v0.3.2-alpha-rc2` while the `version.go` would output `v0.3.2-alpha.rc2`.

scripts/get-git-tag-name.sh
```diff
@@ -0,0 +1,122 @@
...
+  if [ -n "$app_pre_release" ]; then
+      tag_name+="-${app_pre_release}"
+  fi
+
+  echo "$tag_name"
+}
+
+# check_git_tag_name checks if the given git tag name is a valid SemVer
+# compliant version string.
+check_git_tag_name() {
```
guggero, line 65: Do we even need this function anymore? Since we assemble the git tag in this script, we know it should produce a valid semver tag (and if either of the string variables contain something breaking, it would panic when starting the binary, so unit tests would break immediately).

.github/workflows/release.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
           go-version: '${{ env.GO_VERSION }}'
 
       - name: build release for all architectures
-        run: SKIP_VERSION_CHECK=1 make release tag=${{ env.RELEASE_VERSION }}
+        run: make release tag=${{ env.RELEASE_VERSION }}
```
guggero, line 46: To make this work, we need to update the version check in `scripts/release.go`. We should replace the build and invocation with `--version` with just calling the tag given as a parameter matches the one output by `get_git_tag_name`.

guggero (changes requested):
Sorry, found a couple of more things while testing :sweat_smile:

Makefile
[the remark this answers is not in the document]
guggero: Ah, the convention for `-m` would be: `-m "<project_name> <git_tag>"`. So maybe we could pass in the project name from the command line as well? And hard code it in the Makefile as we do with the name of the `version.go` file.

scripts/get-git-tag-name.sh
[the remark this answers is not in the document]
ffranr: I've been thinking a bit more about this. I think that we don't need a regex which checks for semver compliance in all its possible forms. We just need a check for a Lightning Labs version compliance as a subset of semver compliance. We can just manually check that our simpler regex is a subset of semver.

So I think our regex here can be much simpler and we don't need a golang tagging regex check either.

Do you agree?

Makefile
ffranr: > I think the `-a` is implied by `-s` and `-m`.

I don't think that's the case. I think that an annotated tag contains more information than just message and signature. An annotated tag should also include the tagger's name, email, and tagging date. Which shouldn't be included in a non-annotated tag.

> Also, should we move the "$$tag" to the very end?

For clarity, I'll rearrange the command as:
```
git tag -as -m "Tag generated using command `make release-tag`." "$$tag"
```

Let me know if there's more I can do for clarity.

scripts/get-git-tag-name.sh
ffranr: Nice, thanks.

scripts/get-git-tag-name.sh
ffranr: Thanks, I've changed to snake case and added the `local` keyword to narrow the scope.

scripts/get-git-tag-name.sh
ffranr: I've gone with `v0.3.2-alpha.rc2` from both. So the pre-release is prefixed with `.`

scripts/get-git-tag-name.sh
ffranr: Yes I think you're right.

ffranr:
Addressed (responded to) all review comments. New code pushed. Re-requested reviews.

scripts/get-git-tag-name.sh
guggero: I think we need another `else if` case for this project where `git_tag_include_status` is `false`. I tried it locally and the tag created was: `v0.3.2.rc1`.

guggero:
One last issue with the tag name (https://github.com/lightninglabs/taproot-assets/pull/720#discussion_r1469567794), then we're good to go :tada:

scripts/get-git-tag-name.sh
ffranr: OK, we spoke via DM.

My understanding: in the latest commits, `git_tag_include_status` is `false` and `AppPreRelease` is specified then we hope to see git tags of this sort: `v0.3.2-rc1`.

Oli said:
> Yes, anything not part of the actual version number should be separated by a dash. But then the individual parts of the suffix (e.g. alpha and rcX) should be separated by dots to keep the format we have in other projects

jharveyb: Looking at our current tags, I think we never use an `rc` suffix without `alpha` or `beta`? And our release tags have neither (though our release page adds `alpha` to the release name & notes).

I understand that if we want to reuse the work in this PR, we don't want to disallow tags with only one suffix set, but for this repo I think we'd want the option to force these suffixes to be linked / paired.

guggero (approved):
tACK, LGTM :tada:

scripts/get-git-tag-name.sh
guggero: We do use the `-alpha.rcX` currently, otherwise the version check in the release script would fail. But IIRC we do **not** want the `alpha` to be in the git tag at all. So now that we use the same logic for the release version check as we do when creating the tag, we can get rid of it.

jharveyb: Double checked the golang rules, I agree that we could satisfy those rules without the `alpha` or `beta` keywords.

jharveyb (approved):
LGTM, great to see the alphabet changes + Makefile updates! 💯

Issue #578, opened 2023-10-12
tapgarden/custodian: check the local proof archive/universe when looking to receipt proofs


Roasbeef:
Today the custodian looks at the local proof archive (in the database and flat flies on disk) to figure out if it has the receipt proof it needs to spend received assets. It should also create a notification service that uses the local universe instance to detect if it has received proofs. This'll allow users to use `tapcli universe proofs insert` to grab proofs from w/e and insert locally. 

## Steps To Completion 

- [ ] Give the custodian a new multi archive wrapper that'll return inserted proofs at this point: https://github.com/lightninglabs/taproot-assets/blob/a990fdf18c2ca13cbb825c9aa95866831ccfadf8/tapgarden/custodian.go#L540-L551. 
- [ ] This should be allowed to short circuit the normal courier here: https://github.com/lightninglabs/taproot-assets/blob/a990fdf18c2ca13cbb825c9aa95866831ccfadf8/tapgarden/custodian.go#L403-L415.

dstadulis:
Need to add another notifcation

PR #726, opened 2023-12-06
[1/2] custodian: look up proofs in local universe as well

guggero:
~Depends on https://github.com/lightninglabs/taproot-assets/pull/712.~

Fixes https://github.com/lightninglabs/taproot-assets/issues/578.

This PR updates the proof notification publish/subscribe mechanism in a way that allows us to pass in multiple proof sources to the custodian so it can detect incoming proofs from multiple sources.

### DIFF
--- a/fn/func.go
+++ b/fn/func.go
@@ -130,9 +130,33 @@ func All[T any](xs []T, pred func(T) bool) bool {
 	return true
 }
 
+// AllMapItems returns true if the passed predicate returns true for all items
+// in the map.
+func AllMapItems[T any, K comparable](xs map[K]T, pred func(T) bool) bool {
+	for i := range xs {
+		if !pred(xs[i]) {
+			return false
+		}
+	}
+
+	return true
+}
+
 // Any returns true if the passed predicate returns true for any item in the
 // slice.
 func Any[T any](xs []T, pred func(T) bool) bool {
+	for i := range xs {
+		if pred(xs[i]) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// AnyMapItem returns true if the passed predicate returns true for any item in
+// the map.
+func AnyMapItem[T any, K comparable](xs map[K]T, pred func(T) bool) bool {
 	for i := range xs {
 		if pred(xs[i]) {
 			return true
@@ -148,8 +172,28 @@ func NotAny[T any](xs []T, pred func(T) bool) bool {
 	return !Any(xs, pred)
 }
 
+// NotAnyMapItem returns true if the passed predicate returns false for all
+// items in the map.
+func NotAnyMapItem[T any, K comparable](xs map[K]T, pred func(T) bool) bool {
+	return !AnyMapItem(xs, pred)
+}
+
 // Count returns the number of items in the slice that match the predicate.
 func Count[T any](xs []T, pred func(T) bool) int {
+	var count int
+
+	for i := range xs {
+		if pred(xs[i]) {
+			count++
+		}
+	}
+
+	return count
+}
+
+// CountMapItems returns the number of items in the map that match the
+// predicate.
+func CountMapItems[T any, K comparable](xs map[K]T, pred func(T) bool) int {
 	var count int
 
 	for i := range xs {
--- a/fn/iter.go
+++ b/fn/iter.go
@@ -7,10 +7,8 @@ package fn
 // This function can be used instead of the normal range loop to ensure that a
 // loop scoping bug isn't introduced.
 func ForEachErr[T any](s []T, f func(T) error) error {
-	for _, item := range s {
-		item := item
-
-		if err := f(item); err != nil {
+	for i := range s {
+		if err := f(s[i]); err != nil {
 			return err
 		}
 	}
@@ -22,9 +20,17 @@ func ForEachErr[T any](s []T, f func(T) error) error {
 // This can be used to ensure that any normal for-loop don't run into bugs due
 // to loop variable scoping.
 func ForEach[T any](items []T, f func(T)) {
-	for _, item := range items {
-		item := item
-		f(item)
+	for i := range items {
+		f(items[i])
+	}
+}
+
+// ForEachMapItem is a generic implementation of a for-each (map with side
+// effects). This can be used to ensure that any normal for-loop don't run into
+// bugs due to loop variable scoping.
+func ForEachMapItem[T any, K comparable](items map[K]T, f func(T)) {
+	for i := range items {
+		f(items[i])
 	}
 }
 
@@ -35,6 +41,14 @@ func Enumerate[T any](items []T, f func(int, T)) {
 	for i := 0; i < len(items)-1; i++ {
 		item := items[i]
 		f(i, item)
+	}
+}
+
+// EnumerateMap is a generic enumeration function. The closure will be called
+// for each key and item in the passed-in map.
+func EnumerateMap[T any, K comparable](items map[K]T, f func(K, T)) {
+	for key := range items {
+		f(key, items[key])
 	}
 }
 
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -1374,6 +1374,67 @@ func testSendMultipleCoins(t *harnessTest) {
 	AssertNonInteractiveRecvComplete(t.t, secondTapd, 5)
 }
 
+// testSendNoCourierUniverseImport tests that we can send assets to a node that
+// has no courier, and then manually transfer the proof to the receiving using
+// the universe proof import RPC method.
+func testSendNoCourierUniverseImport(t *harnessTest) {
+	ctxb := context.Background()
+
+	// First, we'll make a normal assets with enough units.
+	rpcAssets := MintAssetsConfirmBatch(
+		t.t, t.lndHarness.Miner.Client, t.tapd,
+		[]*mintrpc.MintAssetRequest{simpleAssets[0]},
+	)
+
+	firstAsset := rpcAssets[0]
+	genInfo := firstAsset.AssetGenesis
+
+	// Now that we have the asset created, we'll make a new node that'll
+	// serve as the node which'll receive the assets. We turn off the proof
+	// courier by supplying a dummy implementation.
+	secondTapd := setupTapdHarness(
+		t.t, t, t.lndHarness.Bob, t.universeServer,
+		func(params *tapdHarnessParams) {
+			params.proofCourier = &proof.MockProofCourier{}
+		},
+	)
+	defer func() {
+		require.NoError(t.t, secondTapd.stop(!*noDelete))
+	}()
+
+	// Next, we'll attempt to transfer some amount of assets[0] to the
+	// receiving node.
+	numUnitsSend := uint64(1200)
+
+	// Get a new address (which accepts the first asset) from the
+	// receiving node.
+	receiveAddr, err := secondTapd.NewAddr(ctxb, &taprpc.NewAddrRequest{
+		AssetId: genInfo.AssetId,
+		Amt:     numUnitsSend,
+	})
+	require.NoError(t.t, err)
+	AssertAddrCreated(t.t, secondTapd, firstAsset, receiveAddr)
+
+	// Send the assets to the receiving node.
+	sendResp := sendAssetsToAddr(t, t.tapd, receiveAddr)
+
+	// Assert that the outbound transfer was confirmed.
+	expectedAmtAfterSend := firstAsset.Amount - numUnitsSend
+	ConfirmAndAssertOutboundTransfer(
+		t.t, t.lndHarness.Miner.Client, t.tapd, sendResp,
+		genInfo.AssetId,
+		[]uint64{expectedAmtAfterSend, numUnitsSend}, 0, 1,
+	)
+
+	// Since we disabled proof couriers, we need to manually transfer the
+	// proof from the sender to the receiver now. We use the universe RPC
+	// InsertProof method to do this.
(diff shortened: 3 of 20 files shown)


proof/courier.go
```diff
@@ -1308,24 +1308,25 @@ func FetchProofProvenance(ctx context.Context, originLocator Locator,
 				"failed: %w", err)
 		}
 
-		// Decode transition proof from query response.
-		var currentProof Proof
-		err = currentProof.Decode(bytes.NewReader(proofBlob))
+		// Decode just the asset leaf record from the proof.
+		var proofAsset asset.Asset
+		assetRecord := AssetLeafRecord(&proofAsset)
+		err = SparseDecode(bytes.NewReader(proofBlob), assetRecord)
```
Roasbeef, line 1314: Nice usage of `SparseDecode`! I added it for this exact reason, TLV ftw!

tapdb/assets_store.go
```diff
@@ -1238,6 +1238,37 @@ func (a *AssetStore) FetchProof(ctx context.Context,
...
+// and checking for ErrProofNotFound.
+func (a *AssetStore) HasProof(ctx context.Context, locator proof.Locator) (bool,
+	error) {
+
+	// We don't need anything else but the script key since we have an
+	// on-disk index for all proofs we store.
+	scriptKey := locator.ScriptKey
+	readOpts := NewAssetStoreReadTx()
+	dbErr := a.db.ExecTx(ctx, &readOpts, func(q ActiveAssetsStore) error {
+		_, err := q.FetchAssetProof(
```
Roasbeef, line 1252: If we want this to be the more optimized version (no need to read the blob, etc), then we can have this just check the table for the primary key, instead of reading out he entire thing.

Also IIUC, this won't return true if we have the proof in the Universe tree, but not in our normal asset database, not sure if that's the intention.

tapdb/multiverse.go
```diff
@@ -476,15 +477,20 @@ type MultiverseStore struct {
 	proofCache *proofCache
 
 	leafKeysCache *universeLeafCache
+
+	// eventDistributor is an event distributor that will be used to notify
+	// subscribers about new proof leaves that are added to the multiverse.
+	eventDistributor *fn.EventDistributor[proof.Blob]
```
Roasbeef, line 483: Mention that as is, this is only used for transfer proofs?

proof/mint.go
```diff
@@ -11,9 +11,86 @@ import (
 	"github.com/lightninglabs/taproot-assets/commitment"
 )
 
-// Blob represents a serialized proof file, including the checksum.
+// Blob either represents a serialized proof file, including the checksum or a
+// single serialized issuance/transition proof. Which one it is can be found out
+// from the leading magic bytes (or the helper methods that inspect those).
```
Roasbeef, line 16: Hmm, I wonder if we would be better of making these into different types instead. We added the magic prefix bytes partially to avoid confusing them in the first place. With the type route, we'd break compilation if one was attempted to be passed in where the other was accepted. As is, we reply on runtime checks to avoid such classes of bugs.

Roasbeef: This could use the `Either` type added in the multiverse root PR as an example. Then you can still use `Blob` everywhere, but rely on type safety re protection of trying to use one vs the other.

tapgarden/custodian.go
```diff
@@ -654,9 +663,91 @@ func (c *Custodian) checkProofAvailable(event *address.Event) error {
...
+		log.Debugf("Received single proof, inspecting if matches event")
+		lastProof, err = p.AsSingleProof()
+		if err != nil {
+			return fmt.Errorf("error decoding proof: %w", err)
+		}
+
+		// Before we go ahead and fetch the full file, let's make sure
+		// we are actually interested in this proof.
+		matchesEvent := false
+		for _, event := range c.events {
```
Roasbeef, line 686: This is duplicated with the logic below.

tapgarden/custodian.go
```diff
@@ -654,9 +663,91 @@ func (c *Custodian) checkProofAvailable(event *address.Event) error {
...
+		if err != nil {
+			return fmt.Errorf("error fetching full proof file for "+
+				"event: %w", err)
+		}
+
+		// Do we already have this proof in our main archive? This
+		// should only be false if we got the notification from our
+		// local universe instead of the local proof archive (which the
+		// couriers use).
+		haveProof, err := c.cfg.ProofArchive.HasProof(ctxt, loc)
```
Roasbeef, line 721: Why do we need to check if the archive has it or not? Is this for those that later turn on a public universe so we catch up the on disk state?

itest/send_test.go
```diff
@@ -1370,6 +1370,67 @@ func testSendMultipleCoins(t *harnessTest) {
...
+		[]uint64{expectedAmtAfterSend, numUnitsSend}, 0, 1,
+	)
+
+	// Since we disabled proof couriers, we need to manually transfer the
+	// proof from the sender to the receiver now. We use the universe RPC
+	// InsertProof method to do this.
+	sendProofUniRPC(t, t.tapd, secondTapd, receiveAddr.ScriptKey, genInfo)
+
+	// And now, the transfer should be completed on the receiver side too.
+	AssertNonInteractiveRecvComplete(t.t, secondTapd, 1)
```
Roasbeef, line 1431: 😈

Roasbeef (changes requested):
Nice work! Main comment is if we _really_ want to overload `Blob` (tho it already kinda is) vs using distinct types, and then an `Either` to allow them both be used in instances where either of them will work. 

Reading over this change, it struck me (once again), that we should probably try to consolidate all the different proof storage interfaces we have. At times it isn't super easy to follow where a proof if coming from or why, but the `multi` wrappers do help a bit.

tapdb/assets_store.go
guggero: Yeah, good point. Added a `HasAssetProof` query to speed up.

> Also IIUC, this won't return true if we have the proof in the Universe tree, but not in our normal asset database, not sure if that's the intention.

Correct. The intention is to have a method for finding out which backend has the proof and which doesn't. That's why the multi archiver will return `false` if one of its backends does not have the proof, so we can import it there.

tapgarden/custodian.go
guggero: This is mainly an optimization. Since we get notifications from multiple subsystems, we don't want to waste time overwriting a proof we already have in our main backend (which here is the multi-archiver, which means `HasProof` would only return `true`if we already have it in all backed backends).
Added an additional sentence to make it more clear.

tapgarden/custodian.go
guggero: Yes. The first is to short-cut any calls that are made because of transfer proofs imported into the local universe that we might not be interested in anyway. Added some generic function magic to make it more compact.

proof/mint.go
guggero: Yes, I think distinguishing between a single proof or a proof file would help a lot in the clarity of the code. But I think that would be a bigger refactor and should probably be done in a separate PR.
Do you think this is a reasonable temporary step for this PR?

jharveyb:
Actually, given that some parts of this only handle transfer proofs, do these changes enable 'sideloaded' issuance proofs to get added to our local universe state?

I think for transfers it's implicit as the transfer proof includes the issuance proof, but for those who want to manually sync issuance i'm not sure if we test that rn.

ffranr:
This PR has two reviewers. @guggero would you like me to review also?

guggero:
> This PR has two reviewers. I've had a review request for this. @guggero would you like me to review also?

No, it's fine, thank you. You basically already reviewed this PR since the commits of this one used to be included in https://github.com/lightninglabs/taproot-assets/pull/730 which you did look at.

fn/iter.go
```diff
@@ -22,9 +20,17 @@ func ForEachErr[T any](s []T, f func(T) error) error {
...
-		item := item
-		f(item)
+	for i := range items {
+		f(items[i])
+	}
+}
+
+// ForEachMapItem is a generic implementation of a for-each (map with side
+// effects). This can be used to ensure that any normal for-loop don't run into
+// bugs due to loop variable scoping.
```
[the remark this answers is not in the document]
guggero: Correct, but this change makes everything a bit more consistent IMO.

tapdb/multiverse.go
```diff
@@ -1109,3 +1210,61 @@ func (b *MultiverseStore) FetchLeaves(ctx context.Context,
...
+
+	// No delivery of existing items requested, we're done here.
+	if !deliverExisting {
+		return nil
+	}
+
+	ctx := context.Background()
+	for _, loc := range deliverFrom {
+		id := universe.Identifier{
+			AssetID:   *loc.AssetID,
```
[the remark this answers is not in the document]
guggero: Good point! I guess we don't actually use this part of the code yet, but doesn't hurt to already check things properly.

tapdb/assets_store.go
```diff
@@ -1238,6 +1242,38 @@ func (a *AssetStore) FetchProof(ctx context.Context,
...
+// HasProof returns true if the proof for the given locator exists. This is
+// intended to be a performance optimized lookup compared to fetching a proof
+// and checking for ErrProofNotFound.
+func (a *AssetStore) HasProof(ctx context.Context, locator proof.Locator) (bool,
+	error) {
+
+	// We don't need anything else but the script key since we have an
+	// on-disk index for all proofs we store.
+	var (
+		scriptKey = locator.ScriptKey
```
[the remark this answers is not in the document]
guggero: Hmm, you're right! But seems to be an issue independent of the changes in this PR. Going to create a new tracking issue for it.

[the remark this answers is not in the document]
guggero: https://github.com/lightninglabs/taproot-assets/issues/745

guggero:
> Actually, given that some parts of this only handle transfer proofs, do these changes enable 'sideloaded' issuance proofs to get added to our local universe state?

Yes, since the only way to side load a proof currently (without using dev mode), is the universe RPC's `InsertProof` method which does allow you to insert an issuance proof directly into the universe.
And that does insert the meta data and genesis information into the asset DB as well.
But it does sound like a good integration test case, so I'll add it.

jharveyb:
> > Actually, given that some parts of this only handle transfer proofs, do these changes enable 'sideloaded' issuance proofs to get added to our local universe state?
> 
> Yes, since the only way to side load a proof currently (without using dev mode), is the universe RPC's `InsertProof` method which does allow you to insert an issuance proof directly into the universe. And that does insert the meta data and genesis information into the asset DB as well. But it does sound like a good integration test case, so I'll add it.

With the itest change, I think we still don't exercise import on the genesis proof?

The second node has the default universe server set, so IIUC they sync the genesis proof at that point. Then they receive only the transfer proof, and import that successfully.

What I meant earlier was to actually have the universe server unset for the second node and import the issuance proof manually via `ImportProof`. Then, perform the transfer and import that proof manually as well (which the test does do right now).

guggero:
> 

You mean because `t.universeServer` is still passed in?
We only discussed changing the behavior of this, but currently that variable is required. We do set `params.noDefaultUniverseSync = true` though, which does disable syncing of any proofs with the universe.

So I think the test does exactly what you propose?
With `sendProofUniRPC(t, t.tapd, bob, firstAsset.ScriptKey, firstAssetGen)` manually syncing the issuance proof.

jharveyb (approved):
Good point re: testing issuance insertion as-is. LGTM!

guggero:
Rebased, ready for final review @Roasbeef.

proof/mint.go
Roasbeef: SGTM, down to make an issue for it so we can move on here. Main thing is we can supplement the byte level prefix with a type, and then accept that type contextually to cut down on possible bugs in this area.

fn/func.go
```diff
@@ -130,6 +130,18 @@ func All[T any](xs []T, pred func(T) bool) bool {
 	return true
 }
 
+// AllMapItems returns true if the passed predicate returns true for all items
+// in the map.
+func AllMapItems[T any, K comparable](xs map[K]T, pred func(T) bool) bool {
```
Roasbeef, line 135: Cool! Looking forward to when/if they add more direct iterator stuff to the Go stdlib, then we'd only need one version of this.

proof/archive.go
```diff
@@ -140,6 +140,93 @@ type NotifyArchiver interface {
...
+		return proofBlob, nil
+	}
+
+	return nil, ErrProofNotFound
+}
+
+// RegisterSubscriber adds a new subscriber for receiving events. The
+// registration request is forwarded to all registered archives.
+func (m *MultiArchiveNotifier) RegisterSubscriber(
+	receiver *fn.EventReceiver[Blob], deliverExisting bool,
```
Roasbeef, line 189: Leaving review note to self: does the subscriber need to be ready to handle multiple subscriptions notifications from all the archives for the same proof file?

itest/test_list_on_test.go
```diff
@@ -76,6 +76,10 @@ var testCases = []*testCase{
 		test:             testOfflineReceiverEventuallyReceives,
 		proofCourierType: proof.HashmailCourierType,
 	},
+	{
+		name: "addr send no proof courier with local universe import",
+		test: testSendNoCourierUniverseImport,
```
Roasbeef, line 81: Not `gofmt`'d?

Roasbeef (approved):
LGTM 🍉

itest/test_list_on_test.go
Roasbeef: Linter says: it actually is!

PR #730, opened 2023-12-08
proof: allow proof courier to short cut with local archive

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/726, only the last commit is new.

Fixes https://github.com/lightninglabs/taproot-assets/issues/575.

With this PR we allow the universe RPC courier to short cut the
iterative proof retrieval process if we arrive at a proof in the chain
that we already have in the local archive. And since the local archive
only stores full chains of proofs, once we get one from it, the full
retrieval process is complete.


proof/file.go
```diff
@@ -393,6 +393,26 @@ func (f *File) AppendProof(proof Proof) error {
 	return nil
 }
 
+// AppendProofRaw appends a raw proof to the file and calculates its chained
+// hash.
+func (f *File) AppendProofRaw(proof []byte) error {
+	var prevHash [sha256.Size]byte
```
ffranr, line 399: `var prevHash` should be moved down to just above
```
if !f.IsEmpty() {
	prevHash = f.proofs[len(f.proofs)-1].hash
}
```

proof/archive.go
```diff
@@ -140,6 +140,89 @@ type NotifyArchiver interface {
 	fn.EventPublisher[Blob, []*Locator]
 }
 
+// MultiArchiveNotifier is a NotifyArchiver that wraps several other archives
+// and notifies subscribers about new proofs that are added to any of the
+// archives.
+type MultiArchiveNotifier struct {
```
ffranr, line 146: Why do we need this in addition to `MultiArchiver`? `MultiArchiver` supports the notification interface.

tapdb/multiverse.go
```diff
@@ -850,6 +856,77 @@ func (b *MultiverseStore) FetchProofLeaf(ctx context.Context,
...
+			leafKey.OutPoint = *loc.OutPoint
+		}
+
+		proofs, err := b.FetchProofLeaf(ctx, uniID, leafKey)
+		if errors.Is(err, universe.ErrNoUniverseProofFound) {
+			// If we didn't find a proof, maybe we arrived at the
+			// issuance proof, in which case we need to adjust the
+			// proof type.
+			uniID.ProofType = universe.ProofTypeIssuance
+			proofs, err = b.FetchProofLeaf(ctx, uniID, leafKey)
```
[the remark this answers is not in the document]
guggero: That's handled by the general `if err != nil {` of the outer block, since the `err` variable is declared in the outer scope.

tapdb/multiverse.go
```diff
@@ -850,6 +856,77 @@ func (b *MultiverseStore) FetchProofLeaf(ctx context.Context,
...
+
+			// If we still didn't find a proof, then we'll return
+			// the proof not found error, but the one the proof
+			// package as the Godoc for this method states.
```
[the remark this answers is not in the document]
guggero: Yeah, there was a word missing. Re-wrote the sentence to be more clear, I hope now it makes more sense.

tapdb/multiverse.go
```diff
@@ -850,6 +856,77 @@ func (b *MultiverseStore) FetchProofLeaf(ctx context.Context,
...
+		return nil, fmt.Errorf("error fetching proof from archive: %w",
+			err)
+	}
+
+	var buf bytes.Buffer
+	if err := file.Encode(&buf); err != nil {
+		return nil, fmt.Errorf("error encoding proof file: %w", err)
+	}
+
+	return buf.Bytes(), nil
```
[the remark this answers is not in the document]
guggero: True, but we're implementing a method from an existing interface. And changing that would require a rather big overall refactor. Also see the discussion [here](https://github.com/lightninglabs/taproot-assets/pull/726#discussion_r1421199458), we might do a general overhaul along those lines in a separate PR.

tapdb/multiverse.go
```diff
@@ -944,6 +1021,14 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 	b.proofCache.delProofsForAsset(id)
 	b.leafKeysCache.wipeCache(idStr)
 
+	// Notify subscribers about the new proof leaf, now that we're sure we
+	// have written it to the database. But we only care about transfer
+	// proofs, as the events are received by the custodian to finalize
+	// inbound transfers.
+	if id.ProofType == universe.ProofTypeTransfer {
+		b.eventDistributor.NotifySubscribers(leaf.RawProof)
+	}
```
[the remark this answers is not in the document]
guggero: While I agree in general in terms of separation of concern, this would then notify for each new issuance proof we sync from a universe, which could be a lot. So performance wise I think it's a reasonable tradeoff for now. I also made things more clear by renaming `eventDistributor` to `transferProofDistributor` since we only use it for that purpose right now.

tapdb/multiverse.go
```diff
@@ -1024,6 +1109,18 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 
 	b.rootNodeCache.wipeCache()
 
+	// Notify subscribers about the new proof leaves, now that we're sure we
+	// have written them to the database. But we only care about transfer
+	// proofs, as the events are received by the custodian to finalize
+	// inbound transfers.
+	for idx := range items {
```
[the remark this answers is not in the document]
guggero: See comment about performance above.

tapgarden/custodian.go
```diff
@@ -654,9 +663,91 @@ func (c *Custodian) checkProofAvailable(event *address.Event) error {
...
+		}
+
+		// Before we go ahead and fetch the full file, let's make sure
+		// we are actually interested in this proof.
+		matchesEvent := false
+		for _, event := range c.events {
+			if AddrMatchesAsset(event.Addr, &lastProof.Asset) &&
+				event.Outpoint == lastProof.OutPoint() {
+
+				matchesEvent = true
```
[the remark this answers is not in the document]
guggero: Yes, I refactored this a bit, using a function from `fn`. But there wasn't an `Any` function for maps, so I added that.

proof/archive.go
guggero: This just implements the reduced `NotifyArchiver` so we can support the multiverse store as an archive backend. If we wanted to use the existing `MultiArchiver`, we would need to implement the other methods from the `Archiver` interface in the multiverse DB.

guggero:
Thanks a lot for the reviews, @ffranr! Given this PR only has a single commit on top of #726, I assume you'll approve that one indirectly once you approve this one?

ffranr:
> Thanks a lot for the reviews, @ffranr! Given this PR only has a single commit on top of #726, I assume you'll approve that one indirectly once you approve this one?

I didn't realise that was the case. I think this is a good argument for basing PR's on top of one another rather than `main`. You mentioned in the PR body, I see now ("Depends on https://github.com/lightninglabs/taproot-assets/pull/726, only the last commit is new."). But right now my review isn't in the same place as Laolu's review of #726.

tapdb/multiverse.go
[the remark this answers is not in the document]
ffranr: > See comment about performance above.

Which comment are you referring to please?

If it's performance relating to `RawProof` encoding/decoding then perhaps we can carry along the proof type separately in our notification to the subscribers (tuple: `(ProofType, RawProof)`) to the subscribers.

tapdb/multiverse.go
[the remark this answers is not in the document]
ffranr: I like the rename! Sets expectations up nicely.

proof/courier.go
```diff
@@ -1300,6 +1306,55 @@ func FetchProofProvenance(ctx context.Context, originLocator Locator,
...
+		)
+		if localArchive != nil {
+			fullProof, err = localArchive.FetchProof(
+				ctx, currentLocator,
+			)
+		}
+
+		// If we don't have a local archive, then the IsFile() check
+		// returns false, and we skip this block.
+		if err == nil && fullProof.IsFile() {
```
jharveyb, line 1326: nit: may be useful to have more comment at the beginning of the block, took a bit to see how this lined up with the explanation in the commit msg.

jharveyb (commented):
Looks good, would be neat to have a test with the receiver where part of the proofs (issuance?) are already local.

guggero:
> I think this is a good argument for basing PR's on top of one another rather than main. 

Okay, I did that now (and also added the `[2/2]` title prefix). But I think that using this model will lead to more accidental premature merges because most of the team is used to the other model where the dependencies are outlined in the PR body instead of the base branch (which IMO is very easy to miss).

jharveyb:
Was thinking that someone could use 'request changes' to block the PR, but the author can't do that, only reviewers.

proof/courier.go
guggero: Makes sense, added!

tapdb/multiverse.go
```diff
@@ -960,7 +960,9 @@ func (b *MultiverseStore) FetchProof(ctx context.Context,
 		return proofs[0].Leaf.RawProof, nil
 	}
 
-	file, err := proof.FetchProofProvenance(ctx, originLocator, fetchProof)
+	file, err := proof.FetchProofProvenance(
+		ctx, nil, originLocator, fetchProof,
```
jharveyb, line 964: IIUC here the local archive is `nil` as we're already in the DB?

jharveyb, line 964: IIUC here the local archive is `nil` as we're already in the DB?

proof/courier.go
```diff
@@ -1300,6 +1310,57 @@ func FetchProofProvenance(ctx context.Context, originLocator Locator,
 	// is a reversal of that found in the proof file.
 	var reversedProofs []Blob
 	for {
+		// Before we attempt to fetch the proof from the potentially
```
jharveyb, line 1313: Easier to read now 💯

jharveyb (approved):
LGTM, excited for tap-over-avian-carrier 👍🏽

tapcfg/server.go
```diff
@@ -231,16 +231,11 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
...
-	//  do not yet enable a proof courier. Remove once all integration tests
-	//  support a proof courier.
-	var proofCourierCfg *proof.CourierCfg
-	if cfg.HashMailCourier != nil {
-		proofCourierCfg = &proof.CourierCfg{
-			ReceiverAckTimeout: cfg.HashMailCourier.ReceiverAckTimeout,
-			BackoffCfg:         cfg.HashMailCourier.BackoffCfg,
-			TransferLog:        assetStore,
-		}
+	proofCourierCfg := &proof.CourierCfg{
```
jharveyb, line 234: This is the case of the itest failure; by default we run the tests with no proof courier:

https://github.com/lightninglabs/taproot-assets/blob/becddbf4a658ef3523ebbb4570daaaac1066b944/itest/tapd_harness.go#L244

So we panic here as the nil check was removed.

tapdb/multiverse.go
guggero: Correct. Makes the code a bit harder to read but allows us to re-use the same algorithm for fetching proof individually from a universe in multiple places.

tapcfg/server.go
jharveyb: Looks like itest now failing on a mistmatched state for an address receive, maybe the sideloading now causes the receive state to not be changed somewhere in the custodian?

```bash
assertions.go:769: 
        	Error Trace:	/home/jhb/tap/review_repo/tap/itest/assertions.go:769
        	            				/home/jhb/tap/review_repo/tap/itest/send_test.go:124
        	            				/home/jhb/tap/review_repo/tap/itest/test_harness.go:160
        	            				/home/jhb/tap/review_repo/tap/itest/integration_test.go:79
        	            	got status ADDR_EVENT_STATUS_TRANSACTION_CONFIRMED, wanted ADDR_EVENT_STATUS_COMPLETED
```

From `AssertNonInteractiveRecvComplete` in the itests calling the `AddrReceives` RPC.

jharveyb: Related, we now have a name collision on these 'basic send unidirectional' itests which we should resolve.

proof/courier.go
```diff
@@ -1287,6 +1293,57 @@ func FetchProofProvenance(ctx context.Context, originLocator Locator,
...
+		// remote universe, we'll check our local archive to see if we
+		// already have it. This doesn't make sense in all cases, so we
+		// allow the local archive to be nil, in which case we just skip
+		// this step. The local universe might give us a single proof or
+		// a full file, depending on its type. If we get a full file, we
+		// are already done.
+		var (
+			fullProof Blob
```
Roasbeef, line 1304: So this is'll be the _file_ or the state transition?

Roasbeef (changes requested):
I think the only thing this is missing is a unit test to exercise the short circuit behavior. I think we _could_ do an itest as well, where that would assert than `N` queries were made to the remote universe server, instead of `N+M` or w/e, where `N` is the set of new state transitions, and `M` is that common prefix.

guggero:
Okay, I pushed the commits with the changes I think we need to make.
I know the unit and integration tests currently don't pass, I'll look into those on Monday. But if you want to give this another review round, that would be great @Roasbeef and @jharveyb.

proof/archive.go
```diff
@@ -366,6 +579,12 @@ func (f *FileArchiver) FetchProofs(_ context.Context,
 				"unable to parse script key: %w", err)
 		}
 
+		outPoint, err := ParseOutPoint(parts[1])
```
jharveyb, line 582: IIUC this would fail as we've written a truncated TXID, and the separator between the TXID and outpoint index is now `-` and not `:`.

Related, I don't think we can recover the proof locator from the proof path (as we do at the end of this loop), if we're truncating the TXID. One option is to use a larger alphabet like bech32 or base58, if we don't care about the paths matching the assetID and scriptkey displayed elsewhere.

tapdb/asset_minting_test.go
```diff
@@ -810,11 +810,21 @@ func TestCommitBatchChainActions(t *testing.T) {
 	// If we look up all the proofs by their specific script key, we should
 	// get the same set of proofs.
 	scriptKeys := fMapKeys(
```
jharveyb, line 812: nit: should probably rename this slice to match the new type.

jharveyb (commented):
Looks good overall! The changes to the queries look solid + migration seems reasonable.

proof/courier.go
Roasbeef: Also why do we assume the local archive can only give us the full file? If the local archive is actually our own local universe, then it can receive the state transition blob.

tapdb/addrs.go
```diff
@@ -887,11 +887,19 @@ func (t *TapAddressBook) CompleteEvent(ctx context.Context,
...
 
-	scriptKeyBytes := event.Addr.ScriptKey.SerializeCompressed()
+	outpoint, err := encodeOutpoint(anchorPoint)
+	if err != nil {
+		return fmt.Errorf("unable to encode outpoint: %w", err)
+	}
+
+	args := FetchAssetProof{
+		TweakedScriptKey: event.Addr.ScriptKey.SerializeCompressed(),
+		Outpoint:         outpoint,
```
Roasbeef, line 898: 👍

tapdb/assets_store.go
```diff
@@ -1212,20 +1217,32 @@ func (a *AssetStore) FetchAssetProofs(ctx context.Context,
...
-	scriptKey := locator.ScriptKey
+	// We have an on-disk index for all proofs we store, so we can use the
+	// script key as the primary identifier.
+	args := FetchAssetProof{
+		TweakedScriptKey: locator.ScriptKey.SerializeCompressed(),
+	}
+
+	// But script keys aren't unique, so if the locator explicitly specifies
+	// an outpoint, we'll use that as well.
+	if locator.OutPoint != nil {
```
Roasbeef: Do we need to _enforce_ this now? Eg: what happens if the script key is still used, but we have two blobs to return, which do we select to return back to the caller?

tapdb/sqlc/queries/assets.sql
```diff
@@ -631,15 +631,16 @@ WITH target_asset(asset_id) AS (
...
+    JOIN managed_utxos utxos
+      ON assets.anchor_utxo_id = utxos.utxo_id
     WHERE
         (script_keys.tweaked_script_key = sqlc.narg('tweaked_script_key')
              OR sqlc.narg('tweaked_script_key') IS NULL)
         AND (assets.asset_id = sqlc.narg('asset_id')
                  OR sqlc.narg('asset_id') IS NULL)
-    -- TODO(guggero): Fix this by disallowing multiple assets with the same
-    -- script key!
-    LIMIT 1
```
Roasbeef, line 642: ;)

tapdb/sqlc/queries/assets.sql
```diff
@@ -678,17 +679,21 @@ JOIN asset_info
...
-    SELECT assets.asset_id, script_keys.tweaked_script_key
+    SELECT assets.asset_id, script_keys.tweaked_script_key, utxos.outpoint
     FROM assets
     JOIN script_keys
-        ON assets.script_key_id = script_keys.script_key_id
-    WHERE script_keys.tweaked_script_key = $1
+      ON assets.script_key_id = script_keys.script_key_id
+    JOIN managed_utxos utxos
+      ON assets.anchor_utxo_id = utxos.utxo_id
+   WHERE script_keys.tweaked_script_key = $1
```
Roasbeef, line 688: Extra space on the left needed here?

proof/archive.go
```diff
@@ -231,10 +265,10 @@ var _ NotifyArchiver = (*MultiArchiveNotifier)(nil)
 // archiver takes a single root directory then creates the following overlap
 // mapping:
 //
-// proofs/
-// ├─ asset_id1/
-// │  ├─ script_key1
-// │  ├─ script_key2
+//	proofs/
+//	├─ asset_id1/
+//	│  ├─ script_key1-outpoint.assetproof
```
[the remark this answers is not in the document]
Roasbeef: Any particular rationale re having the script key be the prefix vs the outpoint? No blocker, just curious re if the alternative was considered. 

EDIT: nvm I see it's used for a prefix scan if only the outpoint is provided by a caller.

proof/archive.go
```diff
@@ -266,37 +286,105 @@ func NewFileArchiver(dirName string) (*FileArchiver, error) {
...
+// genProofFileStoragePath generates the full proof file path for storing a
+// proof based on a rootPath and a valid locator.
+// The final path is: root/assetID/scriptKey-outpoint.assetproof.
+//
+// NOTE: Because some operating systems have issues with paths longer than 256
+// characters, we don't use the full outpoint in the file name, but only the
+// first 16 bytes (32 hex characters) of the hash. That should be enough to
+// avoid collisions but saves us a full 32 characters (we already use 130 for
+// the hex encoded asset ID and script key).
+func genProofFileStoragePath(rootPath string, loc Locator) (string, error) {
```
Roasbeef, line 322: Re the root path, don't we also need to consider that if we're worrying about a 260-ish char limit? In other words, not sure if we can 100% ensure an issue w.r.t never happens. 

From a cursory search, looks like FAT32 file systems would be affected, old NTFS, older linux/mac.

proof/archive.go
```diff
@@ -266,37 +286,105 @@ func NewFileArchiver(dirName string) (*FileArchiver, error) {
...
+	// We have no proof for this script key.
+	case len(matches) == 0:
+		return "", ErrProofNotFound
+
+	// Exactly one proof for this script key, we'll return it.
+	case len(matches) == 1:
+		return matches[0], nil
+
+	// User needs to specify the outpoint as well, since we have multiple
+	// proofs for this script key.
```
Roasbeef, line 396: 👍

proof/archive.go
```diff
@@ -352,9 +558,16 @@ func (f *FileArchiver) FetchProofs(_ context.Context,
 			continue
 		}
 
-		scriptKeyBytes, err := hex.DecodeString(strings.ReplaceAll(
+		parts := strings.Split(strings.ReplaceAll(
 			fileName, TaprootAssetsFileSuffix, "",
-		))
+		), "-")
+		if len(parts) != 2 {
```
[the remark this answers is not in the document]
Roasbeef: Nice eye, I think so: `%x-%s-%d%s"`

proof/archive.go
```diff
@@ -376,6 +477,7 @@ func (f *FileArchiver) FetchProofs(_ context.Context,
 			Locator: Locator{
 				AssetID:   &id,
 				ScriptKey: *scriptKey,
+				OutPoint:  outPoint,
```
Roasbeef, line 598: This won't be the full outpoint since we're only storing the truncated version. 

Instead, we'll need to decode enough of the proof file to be able to extract the outpoint.

proof/archive.go
Roasbeef: Yeah this should fail as the outpoint also won't be large enough.

proof/archive_test.go
```diff
@@ -56,74 +112,101 @@ func TestFileArchiver(t *testing.T) {
...
 
 		// We should be able to insert a proof, then get it right back
 		// the same way we found it.
 		{
 			name: "proof happy path",
 			locator: Locator{
-				AssetID:   randAssetID(t),
+				AssetID:   randAssetID(),
 				ScriptKey: *test.RandPubKey(t),
+				OutPoint:  &wire.OutPoint{},
```
Roasbeef, line 174: If we add a slightly diff case (fetch proof, get annotated proof, assert locator), it should catch the issue mentioned above re not being able to recover the full outpoint from the new file prefix.

proof/archive.go
```diff
@@ -280,6 +292,16 @@ func NewFileArchiver(dirName string) (*FileArchiver, error) {
 		return nil, fmt.Errorf("unable to create proof dir: %w", err)
 	}
 
+	// We need to make sure that all our proof files have the new naming
+	// scheme. If they don't, we'll rename them now. This might take quite a
+	// while since we need to read each proof file and parse it to extract
+	// the outpoint and then rename it.
+	err := migrateOldFileNames(proofPath)
```
Roasbeef, line 301: So for now, we'll need to keep this partial migration check in place until we can make some other more breaking change? 

Alternatively, we could also move it to a diff prefix sub-folder, leaving the old proof files in tact. New versions would then only update the new proof files. This might be safer than trying to mass rename all the proof files users have on disk today.

proof/archive.go
```diff
@@ -310,8 +332,9 @@ func genProofFileStoragePath(rootPath string, loc Locator) (string, error) {
 	assetID := hex.EncodeToString(loc.AssetID[:])
 
 	truncatedHash := loc.OutPoint.Hash.String()[:outpointTruncateLength]
-	fileName := fmt.Sprintf("%x-%s-%d%s", loc.ScriptKey.SerializeCompressed(),
-		truncatedHash, loc.OutPoint.Index, TaprootAssetsFileSuffix)
+	fileName := fmt.Sprintf("%x-%s-%d.%s",
```
Roasbeef, line 337: Can refactor slightly to re-use the other function used for reading the file in the first place. This way, the template string is only in one place.

proof/archive.go
```diff
@@ -374,6 +397,96 @@ func lookupProofFilePath(rootPath string, loc Locator) (string, error) {
...
+
+		numFilesRenamed++
+		if numFilesRenamed%1000 == 0 {
+			log.Infof("Renamed %d of %d old files", numFilesRenamed,
+				len(oldProofs))
+		}
+	}
+
+	log.Infof("Done renaming  %d proof files, took %v", len(oldProofs),
+		time.Since(startTime))
```
Roasbeef, line 487: 👍 for logging timing here

asset/asset.go
```diff
@@ -1597,3 +1599,53 @@ func ValidateAssetName(name string) error {
 
 	return nil
 }
+
+// ChainAsset is a wrapper around the base asset struct that includes
+// information detailing where in the chain the asset is currently anchored.
+type ChainAsset struct {
```
Roasbeef, line 1605: 👍

Roasbeef (changes requested):

proof/util.go
```diff
@@ -97,3 +101,27 @@ func nextPowerOfTwo(n int) int {
 	exponent := uint(math.Log2(float64(n))) + 1
 	return 1 << exponent // 2^exponent
 }
+
+// ParseOutPoint parses an outpoint from a string in the format of txid:index.
+func ParseOutPoint(s string) (*wire.OutPoint, error) {
```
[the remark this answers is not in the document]
guggero: Ah nice, seems like we already have that version of `btcd` pulled in. Updated.

proof/archive.go
```diff
@@ -260,43 +294,215 @@ func NewFileArchiver(dirName string) (*FileArchiver, error) {
...
+		return "", ErrMultipleProofs
+	}
 }
 
-// FetchProof fetches a proof for an asset uniquely identified by the
-// passed ProofIdentifier.
+// migrateOldFileNames looks for proof files in the root path that don't conform
+// to the new naming scheme and renames them to the new scheme.
+func migrateOldFileNames(rootPath string) error {
+	// List all files matching rootPath/*/*.assetproof.
```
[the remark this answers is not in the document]
guggero: Actually, that was what the regular expression at the top of the file was supposed to be for, I just forgot to actually use it. Fixed now.

proof/archive.go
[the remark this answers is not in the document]
guggero: Yes, well spotted. Fixed.

proof/archive.go
guggero: Yeah, I forgot to update this after deciding to truncate the outpoint. It's fixed now (not super nice as we need to parse the proof file to extract the full outpoint, but the `FetchProofs` method is only used in a single place when a re-org causes multiple files to be updated, so it shouldn't be too bad).

proof/courier.go
guggero: This will be the full proof file. The `FetchProof` method of the `Archiver` dictates that it has to be the full file, not just the state transition. So even if the given implementation is the local universe, it will assemble and return the full proof file.

tapdb/assets_store.go
guggero: We will enforce this in a later commit, yes. Or rather, _if_ there are multiple proofs for just the script key, we return an error and require the user to also specify the outpoint.

proof/archive.go
guggero: Hmm, good point. Though I'm not sure we want to have logic in our code to detect if this would be an issue? And from what I can tell the error message one would get would be pretty clear (path or file name too long).
By truncating the outpoint our path is around 184 characters long (including the `proofs/` prefix) which leaves at least 66 characters for the folder name this would be located in, which should be enough. Without the truncation it would only be 34 characters which wouldn't be enough for `c:\users\guggero\appdata\roaming\tapd` for example (which is 38 characters).

proof/archive.go
guggero: Unfortunately we need to decode the full proof here. But the `FetchProofs` method is only used to fetch all proofs of an asset when a re-org happens, so we should hit this code path only very seldomly.

proof/archive.go
guggero: I think renaming is fine... I did add the check that we only try to rename files that actually have the old format. That way nothing should be logged in subsequent runs and the detection should be really fast. And then in a couple of versions we can probably just remove this code. Worst case situation would be that the file based archive doesn't find the proofs anymore if the migration code wasn't run for a user that skips many versions when upgrading.

proof/archive.go
guggero: I'm not sure what you mean by "other function". The `genProofFileStoragePath` function is the only one that has the template string... This commit just adds the dot to the template string to make it a bit easier to parse for human eyes (IMO).

guggero:
I addressed all comments. But still need to tackle the following TODOs:
 - Add test case for the case where we have a proof in the local multiverse but not yet in the DB (reproduce `v0.3.3-rc1` problem)
 - Fix integration tests
 - Add unit test that demonstrates the short-circuit behavior

guggero:
I finally addressed all TODOs :tada: 

I also tested the file migration. Though I forgot that in our testnet/prod `tapd` installation we only run a universe server, which doesn't store proof files (everything is in the multiverse database only).

So I instead went ahead and used the load test binary locally, which created 200 assets and sent a couple of them to a second node. After doing that on `main` I restarted the nodes on this branch and observed the log:

First node:
```
2024-02-02 17:18:50.339 [INF] PROF: Found 263 proof files in /home/guggero/.tapd/data/regtest/proofs with old naming scheme, renaming now (will take a while)
2024-02-02 17:18:50.427 [INF] PROF: Done renaming  263 proof files, took 88.173316ms
```

Second node:
```
2024-02-02 17:20:28.219 [INF] PROF: Found 63 proof files in /home/guggero/.tapd-bob/data/regtest/proofs with old naming scheme, renaming now (will take a while)
2024-02-02 17:20:28.239 [INF] PROF: Done renaming  63 proof files, took 20.181639ms
```

I then made sure further assets could be sent successfully, which worked as expected.

proof/archive.go
Roasbeef: SGTM, we can always restore directly from the DB if needed.

proof/archive.go
Roasbeef: I think I meant this one lol, got mixed up on diffs -- resolving

tapdb/assets_store.go
```diff
@@ -1275,10 +1275,26 @@ func (a *AssetStore) FetchProofs(ctx context.Context,
 						"script key: %w", err)
 				}
 
+				f := proof.File{}
+				err = f.Decode(bytes.NewReader(dbRow.ProofFile))
```
Roasbeef, line 1321: One thing we could do to optimize stuff like this would be to have the outpoint he in a predictable location at the end of a file. Not crucial now, just another idea re how we can make the proof files more usable.

tapfreighter/chain_porter.go
```diff
@@ -744,14 +744,13 @@ func (p *ChainPorter) transferReceiverProof(pkg *sendPackage) error {
 				"proof file: %w", err)
 		}
 
-		// Hash proof locator.
-		hash, err := proofLocator.Hash()
```
Roasbeef, line 742: So before for proofs with the same asset ID + script key, we would end up clobbering them in this map, causing us to not write all relevant data to disk?

proof/archive.go
```diff
@@ -140,7 +140,9 @@ type Archiver interface {
 	// passed ProofIdentifier.
 	//
 	// If a proof cannot be found, then ErrProofNotFound should be
-	// returned.
+	// returned. If multiple proofs exist for the given fields of the
```
Roasbeef, line 143: 👍

tapdb/assets_store.go
```diff
@@ -1404,13 +1402,14 @@ func (a *AssetStore) insertAssetWitnesses(ctx context.Context,
...
 
-		err = db.InsertAssetWitness(ctx, PrevInput{
+		err = db.UpsertAssetWitness(ctx, PrevInput{
 			AssetID:              assetID,
 			PrevOutPoint:         prevOutpoint,
 			PrevAssetID:          prevID.ID[:],
 			PrevScriptKey:        prevID.ScriptKey.CopyBytes(),
 			WitnessStack:         witnessStack,
 			SplitCommitmentProof: splitCommitmentProof,
+			WitnessIndex:         int32(idx),
```
Roasbeef, line 1412: Ah so we were missing this all along? Or we were relying on implicit read order here to keep the expected serialization the same?

tapdb/sqlc/migrations/000015_asset_witnesses.up.sql
```diff
@@ -0,0 +1,11 @@
+-- The witness index indicates the order of the witness in the list of witnesses
+-- for a given asset. We didn't really support more than one witness before, so
+-- the default value of 0 should be fine for all existing assets.
+ALTER TABLE asset_witnesses ADD COLUMN witness_index INTEGER NOT NULL DEFAULT 0;
```
Roasbeef, line 4: What about existing entries that don't have the `witness_index` field?

tapdb/assets_store.go
Roasbeef: Or seems the assumption is that no multi-input spends exist today in the DB?

Roasbeef (approved):
LGTM ⛷️

tapdb/sqlc/migrations/000015_asset_witnesses.up.sql
guggero: That's explained in the comment :sweat_smile:

tapdb/assets_store.go
```diff
@@ -2756,9 +2759,8 @@ func (a *AssetStore) reAnchorPassiveAssets(ctx context.Context,
 		}
 
 		// Update the asset proof.
-		err = q.UpsertAssetProof(ctx, ProofUpdate{
-			AssetID:   sqlInt64(passiveAsset.AssetID),
-			Outpoint:  passiveAsset.Outpoint,
```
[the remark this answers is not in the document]
guggero: The asset ID here is the database primary key of the `assets` table. And one asset can only ever have one proof. So we do want to overwrite the proofs here as these are the passive assets that are re-anchored.

tapfreighter/chain_porter.go
guggero: Yes, though since these were just the passive assets within the same commitment, the script keys were forced to be unique, so no clobbering should have happened.

The actual "hack" this commit is cleaning up is that we use the hash of the `Locator` struct that isn't fully populated (only 2 of the 4 members). And then when accessing the map again further down in the call stack, we need to make sure we only fill the same 2 members to arrive at the same hash, otherwise we get a mismatch.
It was good for performance, but bad for understanding what's going on. When I looked for all instances of the `Locator` struct and set the `OutPoint` everywhere, this just broke.
IMO the new way of structuring the proofs has a slight performance hit but has way less chance of us screwing things up by accident.

Also, I'm planning on eliminating the "passive assets" concept soon anyway (in a future refactor PR spun out from the learnings of the TAP channel prototype), so this should be gone eventually.

tapdb/assets_store.go
guggero: Yes, I think we were lucky and the implicit read order was correct so far. And we only have one itest that actually does use multiple inputs.

tapgarden/custodian.go
```diff
@@ -828,6 +826,35 @@ func (c *Custodian) mapProofToEvent(p proof.Blob) error {
...
+
+	// We don't have the proof yet, or not in all backends, so we
+	// need to import it now.
+	if !haveProof {
+		headerVerifier := GenHeaderVerifier(ctxt, c.cfg.ChainBridge)
+		err = c.cfg.ProofArchive.ImportProofs(
+			ctxt, headerVerifier, c.cfg.GroupVerifier, false, p,
+		)
+		if err != nil {
+			log.Errorf("ERROOOORRR: %v", err)
```
jharveyb, line 849: Are we leaving this in as-is? 😅

jharveyb (approved):
LGTM! New tests are solid, great to see all the fixes 🚀

tapfreighter/chain_porter.go
Roasbeef: > Also, I'm planning on eliminating the "passive assets" concept soon anyway (in a future refactor PR spun out from the learnings of the TAP channel prototype), so this should be gone eventually.

Interesting, perhaps make a tracking issue to get more cross-pollination of the idea? Is this related to the `SIGHASH_NOINPUT` ideas, or other?

tapgarden/custodian.go
Roasbeef: Can push a commit to resolve on our timezone before merging.

PR #731, opened 2023-12-11
Refactor fed envy event handling

ffranr:
This PR refactors the federation envoy event handling routine with the goal of code simplification and to gain some line length breathing room.

This PR shouldn't change any "business logic".

### DIFF
--- a/universe/auto_syncer.go
+++ b/universe/auto_syncer.go
@@ -429,59 +429,17 @@ func (f *FederationEnvoy) syncer() {
 					err)
 			}
 
+		// Handle a new batch push request.
 		case pushReq := <-f.batchPushRequests:
 			log.Debug("Federation envoy handling batch push " +
 				"request")
-			ctx, cancel := f.WithCtxQuitNoTimeout()
-
-			// First, we'll attempt to registrar the proof leaf with
-			// the local registrar server.
-			err := f.cfg.LocalRegistrar.UpsertProofLeafBatch(
-				ctx, pushReq.Batch,
-			)
-			cancel()
+			err := f.handleBatchPushRequest(pushReq)
 			if err != nil {
-				err := fmt.Errorf("unable to insert proof "+
-					"batch into local universe: %w", err)
-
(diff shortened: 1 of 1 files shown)


universe/auto_syncer.go
```diff
@@ -403,88 +403,18 @@ func (f *FederationEnvoy) syncer() {
...
-					)
-				f.pushProofToFederation(
-					ctxPush, entry.UniID, entry.LeafKey,
-					&entry.Leaf, servers, true,
-				)
-				cancelPush()
+				// Warn, but don't exit the syncer. The syncer
+				// should continue to run and attempt handle
+				// more events.
+				log.Warnf("unable to handle tick event: %v",
```
guggero, line 416: nit: log messages should start capitalized, here and in the other 2 commits.

universe/auto_syncer.go
```diff
@@ -624,6 +563,74 @@ func (f *FederationEnvoy) handleTickEvent() error {
...
+		return fmt.Errorf("nil push request")
+	}
+
+	// First, we'll attempt to registrar the proof leaf with the local
+	// registrar server.
+	ctx, cancel := f.WithCtxQuit()
+	newProof, err := f.cfg.LocalRegistrar.UpsertProofLeaf(
+		ctx, pushReq.ID, pushReq.Key, pushReq.Leaf,
+	)
+	cancel()
```
guggero, line 538: nit: we can use `defer` to cancel these contexts, since we're in a method now.

guggero (approved):
Nice refactor, LGTM :tada:

universe/auto_syncer.go
ffranr: I've added fixups which use `defer`. Please let me know if that's what you intended.

GeorgeTsagk (approved):
LGTM!
Nice cleanup, I'm always a fan of this approach as `syncer` now is simpler & more readable. No need to go through code you don't want to read :+1:

guggero (approved):

Issue #704, opened 2023-11-28
[bug]: Failed to burn asset: Input does not contain requested asset


ffranr:
Reported by @Liongrass 

> I seem to be unable to burn an asset:
> ```[tapcli] unable to send assets: rpc error: code = Unknown desc = error funding burn: input commitment does not contain asset_id=f04c2820492d2cca4df053cbf83d93eb99e1bc34f51da6acbe8a85e3db5d5914: send: Input does not contain requested asset```
> The asset is a grouped asset, but there exist 90 of them as part of the same asset id. It has 6+ confirmations since I last sent that asset

Command executed was:
```
tapcli assets burn --asset_id 53723837a882c2c3daf92c09406719055e0d1bf58b1c51e59672379ae4263bdc --amount 90
```

ffranr:
I've attempted to reproduce this bug here: https://github.com/lightninglabs/taproot-assets/pull/732

I couldn't reproduce the bug in the simplest case. I've requested logs from @Liongrass .

ffranr:
@Liongrass you say that the asset you were trying to burn was part of a group. But the burning functionality is indifferent to that point. It acts on a single asset identified by it's "asset ID" and not on an asset group. Did asset ID `53723837a882c2c3daf92c09406719055e0d1bf58b1c51e59672379ae4263bdc` have an amount of `90` or was that the total for the whole group?

ffranr:
Logs from Leon: 
```
2023-12-11 16:56:07.235 [INF] FRTR: Identified 1 eligible asset inputs for send of 90 to 53723837a882c2c3daf92c09406719055e0d1bf58b1c51e59672379ae4263bdc
2023-12-11 16:56:07.253 [INF] FRTR: Selected 1 asset inputs for send of 90 to 53723837a882c2c3daf92c09406719055e0d1bf58b1c51e59672379ae4263bdc
2023-12-11 16:56:07.266 [ERR] RPCS: [/taprpc.TaprootAssets/BurnAsset]: error funding burn: input commitment does not contain asset_id=53723837a882c2c3daf92c09406719055e0d1bf58b1c51e59672379ae4263bdc: send: Input does not contain requested asset
```

ffranr:
If I take these errors at face value, they indicate that our coin selector has selected anchored commitments which have associated tap commitments that do not include the target asset.

Leon has provided the tap db of the node which encountered this bug.

ffranr:
[...]

PR #732, opened 2023-12-11
Add asset group burn itest, logging, and missing error handling

ffranr:
This PR adds an itest which ensures that we can burn an asset from an asset group in the simplest case.

It makes the following changes to the `BurnAsset` RPC endpoint:
* Adds additional logging.
* Adds missing error handling when attempting to retrieve a group key.

This PR was created in an attempt to re-produce bug https://github.com/lightninglabs/taproot-assets/issues/704

### DIFF
--- a/itest/burn_test.go
+++ b/itest/burn_test.go
@@ -2,6 +2,7 @@ package itest
 
 import (
 	"context"
+	"encoding/hex"
 
 	taprootassets "github.com/lightninglabs/taproot-assets"
 	"github.com/lightninglabs/taproot-assets/address"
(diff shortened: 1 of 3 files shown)


rpcserver.go
```diff
@@ -2092,8 +2092,18 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
 
 	var groupKey *btcec.PublicKey
 	assetGroup, err := r.cfg.TapAddrBook.QueryAssetGroup(ctx, assetID)
-	if err == nil && assetGroup.GroupKey != nil {
+	switch {
+	case err == nil:
```
guggero, line 2098: I think we still want the `assetGroup.GroupKey != nil` safety check. Because looking at [this](https://github.com/lightninglabs/taproot-assets/blob/bc077ce8844fb1b2f0040979118193d01636c323/tapdb/addrs.go#L952), it can be that we return without an error but with the group being `nil`.

rpcserver.go
```diff
@@ -2106,6 +2108,15 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
 		return nil, fmt.Errorf("error querying asset group: %w", err)
 	}
 
+	var serializedGroupKey []byte
+	if groupKey != nil {
+		serializedGroupKey = groupKey.SerializeCompressed()
+	}
+
+	rpcsLog.Infof("Burning asset (asset_id=%x, group_key=%x, "+
+		"burn_amount=%d)",
```
guggero, line 2117: nit: can add some of the arguments on the same line.

itest/burn_test.go
```diff
@@ -301,3 +302,112 @@ func testBurnAssets(t *harnessTest) {
 	)
 	AssertBalanceByID(t.t, t.tapd, simpleGroupCollectGen.AssetId, 0)
 }
+
+// testBurnGroupedAssets tests that some amount of an asset from an asset group
+// can be burnt successfully.
+func testBurnGroupedAssets(t *harnessTest) {
```
guggero, line 308: So the main difference from this test to the existing one is that we burn assets from a second trance instead of the first one of a group?
https://github.com/lightninglabs/taproot-assets/blob/9943ff1390dca7965defc98cd0062e673b23e509/itest/burn_test.go#L257

itest/burn_test.go
```diff
@@ -301,3 +302,112 @@ func testBurnAssets(t *harnessTest) {
...
+	secondMintReq := simpleAssetsCopy[0]
+	secondMintReq.Asset.Amount = 1010
+	secondMintReq.Asset.GroupKey = assetGroupKey
+	secondMintReq.Asset.GroupedAsset = true
+
+	secondMintResponses := MintAssetsConfirmBatch(
+		t.t, miner, t.tapd,
+		[]*mintrpc.MintAssetRequest{secondMintReq},
+	)
+	require.Equal(t.t, 1, len(secondMintResponses))
```
guggero, line 344: nit: can use `require.Len()`.

guggero (changes requested):
Nice changes! Though I think the first commit is incomplete or can now fail, see inline comment.

itest/burn_test.go
ffranr: Yes, that's all it does.

itest/burn_test.go
```diff
@@ -301,3 +302,112 @@ func testBurnAssets(t *harnessTest) {
...
+	// Our asset balance should have been decreased by the burned amount.
+	AssertBalanceByID(t.t, t.tapd, burnAssetID, postBurnAmt)
+
+	// Confirm that the minted asset group still contains two assets.
+	assetGroups, err = t.tapd.ListGroups(ctxb, &taprpc.ListGroupsRequest{})
+	require.NoError(t.t, err)
+
+	encodedGroupKey = hex.EncodeToString(assetGroupKey)
+	assetGroup = assetGroups.Groups[encodedGroupKey]
+	require.Len(t.t, assetGroup.Assets, 2)
```
GeorgeTsagk, line 412: I wonder if we could (in a future PR) extend this. I am curious what would happen if we burned all the amount from the second mint. Would this then have to be `require.Len(t.t, assetGroup.Assets, 1)` ?

GeorgeTsagk (approved):
Very straight-forward test & easy to follow! :ladder:

GeorgeTsagk:
How are we dealing with the failing vector? I think it will keep this pr from merging

guggero:
> How are we dealing with the failing vector? I think it will keep this pr from merging

I think I might have found the issue, see https://github.com/lightninglabs/taproot-assets/pull/733#issuecomment-1850782675.

PR #733, opened 2023-12-11
Update test vectors

ffranr:
Includes files modified from running `make test-vector-check`.


guggero:
Hmm, I think I know what's going on... It's as simple as having a test name substring collision...
We're running the following commands in `gen-deterministic-test-vectors`:
```
	make unit gen-test-vectors=true pkg=mssmt case=TestReplaceWithEmptyBranch
	make unit gen-test-vectors=true pkg=mssmt case=TestReplace
``

But since the `case` variable ends up in `-test.run=Test...` variable, which is a regular expression, when we run `case=TestReplace` we run both `TestReplaceWithEmptyBranch` and `TestReplace` but within the same process, which then means they're pulling from the same randomness source.

So I think the following patch should make things stable again (after running another `make gen-deterministic-test-vectors`):
[diff.txt](https://github.com/lightninglabs/taproot-assets/files/13640336/diff.txt)

guggero (approved):
Nice, LGTM :tada:

guggero (approved):
Stricter matching vs. renaming is a good idea, I like the second approach too :+1:

GeorgeTsagk (approved):
LGTM!

Issue #597, opened 2023-10-18
[bug]: Proof file not transmitted upon confirmation for asset transfer


Liongrass:
### Background

I minted a normal (`--emission_enabled`) asset on the SENDER node. I synced RECEIVER node to the universe, created an address and sent funds from the SENDER node. Upon confirmation, both nodes sync to the universe, but proof files do not seem to get transmitted, and the RECEIVER does not show the assets in `tapcli assets list` or `tapcli assets balance`

### Your environment

```
    "version": "0.3.0-alpha.rc2 commit=v0.3.0-alpha.rc2-36-gb489d04",
    "lnd_version": "0.17.0-beta",
    "network": "testnet3",
```
* any other relevant environment details

### Steps to reproduce

- mint normal asset with emission_enabled (`tapcli assets mint --type normal --name leocoin --supply 21000 --enable_emission`)
- sync receiver to universe 
- receiver makes address (`tapcli2 addrs new --asset_id f2bdc9a75c4269eefc06bfb3762bb47964614aa612800cabe16eee4ca63a3c13 --amt 21`)
- sender sends coins (`tapcli assets send --addr taptb1qqqsqqspqqzzpu4aexn4csnfam7qd0anwc4mg7tyv992vy5qpj47zmhwfjnr50qnq5ssx424cy357lfll77gdmx3r2um33phj7d6yx76rv5w5admfknujvenqcssy5czsf4akh3y6yhsn6mp3t5rpt7tfrat9znegl9q52z52ldfnv28pqss9qaejy86wuuaf203n0c6nnvnhxeqmh0uxa6qanmnfr2twjsaeel4pgq32rpkw4hxjan9wfek2unsvvaz7tm5v4ehgmn9wsh82mnfwejhyum99ekxjemgw3hxjmn89enxjmnpde3k2w33xqcrywgedu59a`

### Expected behavior

Once the [transaction](https://mempool.space/testnet/tx/cbdf20d6764e216b625d1f622cb5b788ed7eeb29035826f299dca251ec7fbecf#flow=&vin=0) is confirmed I expect the sender to transmit their proofs to the recipient, who then knows about the assets.

### Actual behavior

No sync of proofs appear to happen.

Logs of both sender and receiver:

[sender_logs.txt](https://github.com/lightninglabs/taproot-assets/files/12979777/sender_logs.txt)
[receiver_logs.txt](https://github.com/lightninglabs/taproot-assets/files/12979780/receiver_logs.txt)

[...]

PR #734, opened 2023-12-11
proof courier: general cleanup and refactor, re-try after restart

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/597.

This PR removes a bunch of TODOs around the way proof couriers are instantiated and receive their configuration.
After the refactor we then fix a bug where we didn't attempt to fetch a proof using a proof courier again after restarting the daemon.

### DIFF
--- a/itest/tapd_harness.go
+++ b/itest/tapd_harness.go
@@ -76,7 +76,7 @@ const (
 	// defaultProofTransferReceiverAckTimeout is the default itest specific
 	// timeout we'll use for waiting for a receiver to acknowledge a proof
 	// transfer.
-	defaultProofTransferReceiverAckTimeout = 15 * time.Second
+	defaultProofTransferReceiverAckTimeout = 5 * time.Second
 )
 
 // tapdHarness is a test harness that holds everything that is needed to
--- a/itest/tapd_harness.go
+++ b/itest/tapd_harness.go
@@ -47,9 +47,19 @@ var (
 			"allow the postgres fixture to run in total. Needs "+
 			"to be increased for long-running tests.")
 
-	// defaultBackoffConfig is the default backoff config we'll use for
-	// sending proofs.
-	defaultBackoffConfig = proof.BackoffCfg{
+	// defaultHashmailBackoffConfig is the default backoff config we'll use
+	// for sending proofs with the hashmail courier.
+	defaultHashmailBackoffConfig = proof.BackoffCfg{
+		BackoffResetWait: time.Second,
+		NumTries:         5,
+		InitialBackoff:   300 * time.Millisecond,
+		MaxBackoff:       600 * time.Millisecond,
+	}
+
+	// defaultUniverseRpcBackoffConfig is the default backoff config we'll
+	// use for sending proofs with the universe RPC courier.
+	defaultUniverseRpcBackoffConfig = proof.BackoffCfg{
+		SkipInitDelay:    true,
 		BackoffResetWait: time.Second,
 		NumTries:         5,
 		InitialBackoff:   300 * time.Millisecond,
@@ -209,9 +219,11 @@ func newTapdHarness(t *testing.T, ht *harnessTest, cfg tapdConfig,
 	// Populate proof courier specific config fields.
 	//
 	// Use passed in backoff config or default config.
-	backoffCfg := defaultBackoffConfig
+	hashmailBackoffCfg := defaultHashmailBackoffConfig
+	universeRpcBackoffCfg := defaultUniverseRpcBackoffConfig
 	if opts.proofSendBackoffCfg != nil {
-		backoffCfg = *opts.proofSendBackoffCfg
+		hashmailBackoffCfg = *opts.proofSendBackoffCfg
+		universeRpcBackoffCfg = *opts.proofSendBackoffCfg
 	}
 
 	// Used passed in proof receiver ack timeout or default.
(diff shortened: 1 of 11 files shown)


proof/courier.go
```diff
@@ -541,11 +541,17 @@ func (e *BackoffExecError) Error() string {
...
+	// SkipInitDelay is a flag that indicates whether we should skip the
+	// initial delay before attempting to deliver the proof to the receiver
+	// or receiving from the sender.
 	SkipInitDelay bool
 
+	// InitialReceiveDelay, if not zero, is the amount of time we'll wait
+	// before attempting to receive a proof from the sender. If this is set,
+	// we'll wait for the specified duration regardless of what the transfer
+	// log contains.
+	InitialReceiveDelay time.Duration
```
ffranr: In my mind this delay is more appropriately part of the custodian. Because it relates to the method that the custodian uses to respond to on-chain transactions which indicate an asset transfer has taken place. That is to say, when should the custodian retrieve a proof having seen an on-chain transaction.

This delay isn't related to backoffs and repeat attempts (i.e. `BackoffCfg`).

How do you think about this delay?

proof/courier.go
```diff
@@ -578,6 +584,27 @@ type BackoffHandler struct {
 func (b *BackoffHandler) initialDelay(ctx context.Context,
 	proofLocator Locator, proofTransferType TransferType) error {
 
+	// If there is an initial delay set, we'll always wait for that amount
+	// of time, regardless of what the transfer log contains. This is to
+	// give the sender some time to transfer the proof to the universe if
+	// a universe RPC courier is being used.
```
ffranr, line 532: I think the same delay is technically relevant to the hashmail courier as well. I'm not sure.

ffranr (changes requested):
Nice to see the disentangled proof courier configs. I don't think moving the custodian delay makes sense to me.

proof/courier.go
guggero: Yes, you are correct that waiting for the sender to upload the proof using the courier is more of a function of the custodian. But imagine for some reason the delay is super short or super long depending on the type of proof courier, would we then go ahead and add a per-courier delay to the custodian?
I think it would be more easy to understand and configure if all time/backoff related configuration items that actually could be different per proof courier are configured on the courier itself.
But happy to change this back if you disagree.

tapgarden/custodian_test.go
```diff
@@ -402,21 +531,92 @@ func TestTransactionHandling(t *testing.T) {
...
-		}
+	dbProof, err := h.assetDB.FetchProof(ctx, mockProof.Locator)
+	require.NoError(t, err)
+	require.EqualValues(t, mockProof.Blob, dbProof)
+}
+
+// TestTransactionConfirmedOnly tests that the custodian only starts the proof
+// courier once a transaction has been confirmed. We also test that it correctly
+// re-tries fetching proofs using a proof courier after it has been restarted.
+func TestTransactionConfirmedOnly(t *testing.T) {
```
jharveyb, line 545: 💯 

This serves as a repro for #597 right?

itest/tapd_harness.go
```diff
@@ -47,26 +47,33 @@ var (
...
+		InitialReceiveDelay: 200 * time.Millisecond,
+	}
 )
 
 const (
 	// defaultProofTransferReceiverAckTimeout is the default itest specific
 	// timeout we'll use for waiting for a receiver to acknowledge a proof
 	// transfer.
-	defaultProofTransferReceiverAckTimeout = 15 * time.Second
+	defaultProofTransferReceiverAckTimeout = 5 * time.Second
```
jharveyb, line 76: Were there issues going below 5 here?

jharveyb (approved):
LGTM, nice to see the simpler courier handling + general cleanup.

proof/courier.go
jharveyb: I agree with Oli here, mostly because I think these timeouts should not be duplicated across multiple packages / subsystems, and since the courier configs already define timeout & backoff behavior, IMO it works better to have those values in those configs.

Another way to think about it would be, if I'm adding a courier with timeouts and backoff delays very different to what we have now, what would I have to change & where? IIUC this PR moves all of those potential changes into the courier config and internal logic.

tapgarden/custodian_test.go
guggero: Ah yes, correct.

itest/tapd_harness.go
guggero: Things became unstable, yes. But didn't want to invest more time as they ran very stable with 5 seconds.

proof/courier.go
ffranr: I'm not sure.

These changes hard code the delay into every new instance of the custodian's proof courier types. But if the custodian is [handling an old transaction](https://github.com/lightninglabs/taproot-assets/blob/becddbf4a658ef3523ebbb4570daaaac1066b944/tapgarden/custodian.go#L286) (catching up after being offline) then the proof courier will unnecessarily delay proof retrieval. I think this highlights my point that waiting on the timeout is a custodian specific feature.

Sure we could add precision by making the delay proof courier specific. But whether we want the delay or not is based on when we start handling a receive event.

For example, we might use the custodian to retrieve proofs as part of recovery. In which case we wouldn't want this delay at all.

I think the custodian needs to decide on whether to delay or not and the proof courier needs to decide on how long that delay is.

And I'm still not convinced this should be part of the backoff config. Right now, as far as I can tell, no part of the backoff procedure is sender/receiver specific. I think it would be good to retain and enhance that generality if possible. This delay could happen before starting the backoff procedure.

guggero: Okay, I guess that's a solid argument. I removed the change, please take a look.

ffranr:
@guggero just two nits, but otherwise I can approve quickly

tapcfg/server.go
```diff
@@ -321,11 +321,11 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 	// Addresses can have different proof couriers configured, but both
 	// types of couriers that currently exist will receive this config upon
 	// initialization.
-	proofCourierCfg := &proof.CourierCfg{
+	dispatcher := proof.NewDispatch(&proof.CourierCfg{
```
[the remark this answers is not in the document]
guggero: Updated.

tapgarden/custodian_test.go
```diff
@@ -114,6 +116,48 @@ func (h *custodianHarness) eventually(fn func() bool) {
...
 }
 
+// assertEventsPresent makes sure that the given number of events is present in
+// the address book, then returns those events.
+func (h *custodianHarness) assertEventsPresent(numEvents int,
+	status address.Status) []*address.Event {
+
+	ctx := context.Background()
+	ctxt, cancel := context.WithTimeout(ctx, testTimeout)
+	defer cancel()
```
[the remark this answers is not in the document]
guggero: No, we still need a timeout context. If the request gets stuck on the server side, `wait.NoError` won't cancel it, it just won't re-try if the given time has elapsed. So if we want to make sure a request doesn't hang, we do need to add a timeout to the context itself.

ffranr (approved):
Nice!

PR #735, opened 2023-12-13
Add means to specify sqlite db file path for tapd test harness and db unit tests

ffranr:
This PR will allow us to specify a particular tapd sqlite database file when creating an itest tapd harness or when unit testing our database store functionality. This should help us in debugging tapd. 

This PR was created as part of the work towards solving https://github.com/lightninglabs/taproot-assets/issues/704

### DIFF
--- a/tapdb/sqlite.go
+++ b/tapdb/sqlite.go
@@ -170,13 +170,21 @@ func (s *SqliteStore) ExecuteMigrations(target MigrationTarget) error {
 func NewTestSqliteDB(t *testing.T) *SqliteStore {
 	t.Helper()
 
-	dbFileName := filepath.Join(t.TempDir(), "tmp.db")
-	t.Logf("Creating new SQLite DB for testing: %s", dbFileName)
-
 	// TODO(roasbeef): if we pass :memory: for the file name, then we get
 	// an in mem version to speed up tests
+	dbPath := filepath.Join(t.TempDir(), "tmp.db")
+	t.Logf("Creating new SQLite DB handle for testing: %s", dbPath)
+
+	return NewTestSqliteDbHandleFromPath(t, dbPath)
+}
+
+// NewTestSqliteDbHandleFromPath is a helper function that creates a SQLite
+// database handle given a database file path.
+func NewTestSqliteDbHandleFromPath(t *testing.T, dbPath string) *SqliteStore {
+	t.Helper()
+
 	sqlDB, err := NewSqliteStore(&SqliteConfig{
-		DatabaseFileName: dbFileName,
+		DatabaseFileName: dbPath,
 		SkipMigrations:   false,
 	})
 	require.NoError(t, err)
--- a/tapdb/sqlutils_test.go
+++ b/tapdb/sqlutils_test.go
@@ -221,11 +221,8 @@ func (d *DbHandler) AddRandomServerAddrs(t *testing.T,
 	return addrs
 }
 
-// NewDbHandle creates a new store and query handle to the test database.
-func NewDbHandle(t *testing.T) *DbHandler {
-	// Create a new test database.
-	db := NewTestDB(t)
-
+// newDbHandleFromDb creates a new database store handle given a database store.
+func newDbHandleFromDb(db *BaseDB) *DbHandler {
 	testClock := clock.NewTestClock(time.Now())
 
 	// Gain a handle to the pending (minting) universe federation store.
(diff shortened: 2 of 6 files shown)


tapdb/sqlite.go
```diff
@@ -168,15 +168,21 @@ func (s *SqliteStore) ExecuteMigrations(target MigrationTarget) error {
...
 // testing.
 func NewTestSqliteDB(t *testing.T) *SqliteStore {
-	t.Helper()
-
-	dbFileName := filepath.Join(t.TempDir(), "tmp.db")
-	t.Logf("Creating new SQLite DB for testing: %s", dbFileName)
-
 	// TODO(roasbeef): if we pass :memory: for the file name, then we get
 	// an in mem version to speed up tests
+	dbPath := filepath.Join(t.TempDir(), "tmp.db")
```
guggero, line 173: I think the `t.Helper()` method is required in every helper method in the call stack to avoid it being attributed as the source of a test failure.

tapdb/sqlutils_test.go
```diff
@@ -221,11 +221,15 @@ func (d *DbHandler) AddRandomServerAddrs(t *testing.T,
 	return addrs
 }
 
-// NewDbHandle creates a new store and query handle to the test database.
-func NewDbHandle(t *testing.T) *DbHandler {
-	// Create a new test database.
-	db := NewTestDB(t)
+// DbStore is a database store interface.
+type DbStore interface {
```
guggero, line 225: I think instead of using yet another interface, we can just use `*BaseDB` for functions that need a generic handling of database related things. See https://github.com/lightninglabs/taproot-assets/blob/6a18d307e171d46180edf40d73f466bb514a23f5/tapdb/asset_minting_test.go#L45 for example.

tapdb/sqlutils_test.go
```diff
@@ -268,3 +272,17 @@ func NewDbHandle(t *testing.T) *DbHandler {
 		DirectQuery:             db,
 	}
 }
+
+// NewDbHandleFromPath creates a new database store handle given a database file
+// path.
+func NewDbHandleFromPath(t *testing.T, dbPath string) *DbHandler {
```
guggero, line 278: I'm a bit confused by the naming convention used here. If something is a handler, why would the variable names and functions to create such a handler only be referred to as "handle"? Or is that just a very consistent typo?
I still think "harness" is a better name here, since that's a term we very often use in all our projects for things that just hold test instances together and provide helper methods...

guggero (commented):
The code in the PR looks good (with just a few comments around naming, as usual :sweat_smile: ).
Though I'm not 100% sure we want or need to actually commit this code to the repo? Same with https://github.com/lightninglabs/taproot-assets/pull/736, it's great for debugging, but do we actually intend to merge that?

Also, just out of curiosity, wouldn't it be easier to just spin up a local node and place the DB there, then investigate RPC behavior using the CLI?

jharveyb:
Is this to get the itest DBs out of `/tmp/` and maybe have filenames that are easier to handle?

ffranr:
> Is this to get the itest DBs out of `/tmp/` and maybe have filenames that are easier to handle?

@jharveyb So far, I've tried debugging two different bugs using the bug reporter's tapd sqlite database file. The changes in this PR let me point a database unit test or an itest tapd harness towards an existing tapd sqlite database file.

ffranr:
> The code in the PR looks good (with just a few comments around naming, as usual 😅 ). Though I'm not 100% sure we want or need to actually commit this code to the repo? Same with #736, it's great for debugging, but do we actually intend to merge that?

#736 was just for debugging/troubleshooting, to show you guys the work I've been doing. It was never intended to be merged.

> Also, just out of curiosity, wouldn't it be easier to just spin up a local node and place the DB there, then investigate RPC behavior using the CLI?

I find it easier to write a test. That way I can commit and share. It's also easier to come back to later. And inspect state with golang dbg.

I've wanted `NewDbHandleFromPath` to exist before now (since looking at a previous bug).

tapdb/sqlutils_test.go
ffranr: I don't think we can use `*SqliteStore` as the type `*BaseDB`?

tapdb/sqlutils_test.go
ffranr: The way I think about it is that the type struct is a handler and an instance of that type is a handle.

We can go with harness, I don't mind either way.

I'll put up another PR to rename if you don't mind.

tapdb/sqlutils_test.go
guggero: Not directly, no. But `SqliteStore` has the `BaseDB` embedded.

tapdb/sqlutils_test.go
guggero: Okay, thanks for clarifying.
It would be nice to rename, but let's not make it a priority. Perhaps keep it in mind for the next time we add or change something to that struct?

guggero:
> https://github.com/lightninglabs/taproot-assets/pull/736 was just for debugging/troubleshooting, to show you guys the work I've been doing. It was never intended to be merged.

Okay, makes sense.

guggero (commented):
I'd still ike to get rid of the extra `DbStore` interface as IMO the number of different interfaces we have for DB stuff is already confusing enough (see inline comment). Other than that this is good to go.

tapdb/sqlutils_test.go
ffranr: Ah thanks for pointing that out, that works.

guggero (approved):

jharveyb (approved):
LGTM 💯

PR #662, opened 2023-11-08
loadtest: add support for push metrics

calvinrzachman:
Equip the test orchestrator with the ability to push metrics on execution time to a configurable remote prometheus push gateway.

Related to: https://github.com/lightninglabs/taproot-assets/pull/662 (`loadtest` metrics vs. `tapd` metrics)

PR #737, opened 2023-12-13
Loadtesting metrics, updated

GeorgeTsagk:
## Description

This PR is based on https://github.com/lightninglabs/taproot-assets/pull/662 by @calvinrzachman. It is an update / revisited version for our loadtesting setup and metrics collection.

### DIFF
--- a/itest/loadtest/config.go
+++ b/itest/loadtest/config.go
@@ -58,6 +58,17 @@ type BitcoinConfig struct {
 	TLSPath  string `long:"tlspath" description:"Path to btcd's TLS certificate, if TLS is enabled"`
 }
 
+// PrometheusGatewayConfig defines exported config options for connecting to the
+// Prometheus PushGateway.
+type PrometheusGatewayConfig struct {
+	// nolint: lll
+	Enabled bool `long:"enabled" description:"Enable pushing metrics to Prometheus PushGateway"`
+	// nolint: lll
+	Host    string `long:"host" description:"Prometheus PushGateway host address"`
+	Port    int    `long:"port" description:"Prometheus PushGateway port"`
+	PushURL string
+}
+
 // Config holds the main configuration for the performance testing binary.
 type Config struct {
 	// TestCases is a comma separated list of test cases that will be
@@ -97,6 +108,12 @@ type Config struct {
 
 	// TestTimeout is the timeout for each test.
 	TestTimeout time.Duration `long:"test-timeout" description:"the timeout for each test"`
+
+	// PrometheusGateway is the configuration for the Prometheus
+	// PushGateway.
+	//
+	// nolint: lll
+	PrometheusGateway *PrometheusGatewayConfig `group:"prometheus-gateway" namespace:"prometheus-gateway" description:"Prometheus PushGateway configuration"`
 }
 
 // DefaultConfig returns the default configuration for the performance testing
@@ -120,6 +137,11 @@ func DefaultConfig() Config {
 		SendType:         taprpc.AssetType_COLLECTIBLE,
 		TestSuiteTimeout: defaultSuiteTimeout,
 		TestTimeout:      defaultTestTimeout,
+		PrometheusGateway: &PrometheusGatewayConfig{
+			Enabled: false,
+			Host:    "localhost",
+			Port:    9091,
+		},
 	}
 }
 
@@ -156,6 +178,28 @@ func LoadConfig() (*Config, error) {
 // of it with sane defaults.
 func ValidateConfig(cfg Config) (*Config, error) {
 	// TODO (positiveblue): add validation logic.
+
+	// Validate Prometheus PushGateway configuration.
+	if cfg.PrometheusGateway.Enabled {
+		gatewayHost := cfg.PrometheusGateway.Host
+		gatewayPort := cfg.PrometheusGateway.Port
+
+		if gatewayHost == "" {
+			return nil, fmt.Errorf(
+				"gateway hostname may not be empty",
+			)
+		}
+
+		if gatewayPort == 0 {
+			return nil, fmt.Errorf("gateway port is not set")
+		}
+
+		// Construct the endpoint for Prometheus PushGateway.
+		cfg.PrometheusGateway.PushURL = fmt.Sprintf(
+			"%s:%d", gatewayHost, gatewayPort,
+		)
+	}
+
 	return &cfg, nil
 }
 
--- a/itest/loadtest/load_test.go
+++ b/itest/loadtest/load_test.go
@@ -4,10 +4,29 @@ package loadtest
 
 import (
 	"context"
+	"fmt"
 	"testing"
+	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/push"
 	"github.com/stretchr/testify/require"
 )
+
+var (
+	testDuration = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "test_duration_seconds",
+			Help: "Duration of the test execution, in seconds",
+		},
+		[]string{"test_name"},
+	)
+)
+
+func init() {
+	// Register the metric with Prometheus's default registry.
+	prometheus.MustRegister(testDuration)
+}
 
 type testCase struct {
 	name string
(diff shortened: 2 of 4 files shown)


GeorgeTsagk:
As discussed offline:
We will not include client-side metrics for resource usage as that's not that insightful, since the main load is placed on the universe server

itest/loadtest/load_test.go
```diff
@@ -55,6 +77,31 @@ func TestPerformance(t *testing.T) {
...
+			// Update the metric with the test duration.
+			testDuration.WithLabelValues(tc.name).Set(duration)
+
+			// Create a new pusher to push the metrics.
+			pushURL := cfg.PrometheusGateway.Host + ":" +
+				strconv.Itoa(cfg.PrometheusGateway.Port)
+
+			pusher := push.New(pushURL, "load_test").
+				Collector(testDuration).
+				Grouping("test_case", tc.name)
```
Roasbeef, line 179: Does the test name include factors such as the number of assets created, number before/after, etc? If not, we may want to add additional metrics, but as new time series rather than labels (don't want the labels to get too long).

itest/loadtest/config.go
```diff
@@ -154,5 +156,25 @@ func LoadConfig() (*Config, error) {
...
+		if gatewayHost == "" {
+			return nil, fmt.Errorf("gateway hostname may not be empty")
+		}
+
+		if gatewayPort == 0 {
+			return nil, fmt.Errorf("gateway port is not set")
+		}
+
+		// Construct the endpoint for Prometheus PushGateway.
+		cfg.PrometheusGateway.PushURL = fmt.Sprintf(
```
Roasbeef, line 174: Perhaps this can be part of the config creation? Otherwise, something meant to validate is actually _mutating_ the underlying config.

Roasbeef (changes requested):

itest/loadtest/load_test.go
```diff
@@ -20,11 +21,30 @@ var (
...
+
+	memAlloc = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "memory_alloc",
+			Help: "Memory usage of the test execution, in bytes",
+		},
+		[]string{"test_name"},
+	)
+
+	memTotalAlloc = prometheus.NewGaugeVec(
```
Roasbeef, line 33: Do we need this given the prom metrics already export memory related to the test? https://github.com/danielfm/prometheus-for-developers/blob/master/README.md#measuring-memorycpu-usage

Roasbeef (changes requested):

itest/loadtest/load_test.go
Roasbeef: Instead, I think we need to focus on metrics including (some of this is a matter of enableing the gRPC prom metrics, and utilizing what we've already exported w/ new queries) :
  * time to execute the various components of the test (list asset, coin selection, etc) 
    * this may require deeper telemetry within tapd itself, depending on the way the load tests are written 
  * proof size growth as a result of test instance (size after compared to size before) 
  * total db size (for both sqlite and postgres) 
  * total number of assets created
  * total number of assets sent 
  * 


cc @calvinrzachman as he may already have some of this posted in the dashboard

Roasbeef (changes requested):

itest/loadtest/load_test.go
GeorgeTsagk: > Do we need this given the prom metrics already export memory related to the test? https://github.com/danielfm/prometheus-for-developers/blob/master/README.md#measuring-memorycpu-usage

Probably not, will take a look at this

GeorgeTsagk: > time to execute the various components of the test (list asset, coin selection, etc)
this may require deeper telemetry within tapd itself, depending on the way the load tests are written

I think it's better to fully rely on the long-running tapd prometheus metrics that were updated on [tapd/pull/716](https://github.com/lightninglabs/taproot-assets/pull/716/files). We could probably come up with some fancy collectors to retrieve everything.

Given that the tapd instances that the loadtests run against are long-running, the metrics will persist across multiple loadtest runs.

GeorgeTsagk: So next steps:
- Strip any redundant changes from this PR: Let's keep the push metrics that are introduced here only for tracking the duration of the test
- Enhance tapd itself with any type of collector we deem appropriate (gRPC, proof related etc): Decorate our grafana dashboards with the prom metrics of the long-running alice & bob nodes

Tracking extra things in this layer (`itest/loadtest`) is kind of useless or very limiting as this is the client side and we don't really care about this code area, or it would be much harder extracting the tapd metrics on this level and then pushing it to prom

GeorgeTsagk:
With [this](https://github.com/lightninglabs/taproot-assets/pull/737#discussion_r1695503524) in mind, I'm marking this as ready for review. We will add more sophisticated tracking in tapd's prometheus monitoring, not in the loadtesting code

GeorgeTsagk:
Closes https://github.com/lightninglabs/lightning-infra/issues/1455

itest/loadtest/load_test.go
```diff
@@ -5,10 +5,28 @@ package loadtest
...
+	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/push"
 	"github.com/stretchr/testify/require"
 )
 
+var (
+	testDuration = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
```
Roasbeef, line 17: We should make this a histogram metric. Then we'll be able to do percentile plots, and heat maps, etc.

Roasbeef (changes requested):

itest/loadtest/load_test.go
GeorgeTsagk: So I looked into this and a histogram won't persist over multiple runs (it will be overwritten across runs). There's some tricks we can try to get the gateway to persist but really not worth the time & diff.

Given the frequency at which we run the loadtests, we can produce a histogram from the GaugeVec with PromQL directly in Grafana. This is not as performant as directly using a histogram, but will give us the same insights on percentiles etc

Roasbeef: Are you referring to [this](https://github.com/prometheus/pushgateway#:~:text=For%20the%20most,of%20the%20Pushgateway).)? 

IIUC that means that if the pushgateway is restart before it's scraped, the metrics won't persist, but IIUC we have the system always running.

GeorgeTsagk: yeah in our setup pushgateway should always be alive. I was referring to the part where the client side restarts (by design) and we push a fresh instance of "histogram". See [here](https://github.com/prometheus/pushgateway/issues/65)

By pushing a fresh histogram to the pushgateway we're overwriting the old one, effectively only keeping the values of our last run (last-write-wins)

I'm not sure if a HistogramVec, with unique label per test run, is going to be a fine workaround, less performant than a simple histogram nevertheless

GeorgeTsagk: There's also a community implementation of pushgateway that seems to be solving this issue (haven't tested it)
https://github.com/zapier/prom-aggregation-gateway

Roasbeef:
Has some linter failures: 
```
itest/loadtest/config.go:64: line is 102 characters (lll)
	Enabled bool   `long:"enabled" description:"Enable pushing metrics to Prometheus PushGateway"`
itest/loadtest/config.go:65: line is 86 characters (lll)
	Host    string `long:"host" description:"Prometheus PushGateway host address"`
itest/loadtest/config.go:110: line is 81 characters (lll)
	// PrometheusGateway is the configuration for the Prometheus PushGateway.
itest/loadtest/config.go:111: line is 161 characters (lll)
	PrometheusGateway *PrometheusGatewayConfig `group:"prometheus-gateway" namespace:"prometheus-gateway" description:"Prometheus PushGateway configuration"`
itest/loadtest/config.go:183: line is 83 characters (lll)
			return nil, fmt.Errorf("gateway hostname may not be empty")
```

Roasbeef (approved):
LGTM 🦜


g2g after linter fix

itest/loadtest/load_test.go
GeorgeTsagk: The label was updated so that each gauge gets a unique one, in order to distinguish the pushed metrics. On promQL it's fairly easy to aggregate everything into a single time series regardless of being individual gauges

The assets minted / sent are configured in the `loadtest.conf` file so that number is always stable (for now)

We can rely on the tapd prom metrics to relate loadtest durations with tapd behavior

Issue #297, opened 2023-04-26
Add full-value multi-input homogeneous asset IDs support to RPC endpoint AnchorVirtualPsbts


ffranr:
Currently, this RPC endpoint formulates a shipment request using a single input from the virtual package only. The endpoint should be extended to support multi input virtual packets.

ffranr:
As far as I can tell, the change necessary to the `AnchorVirtualPsbts` RPC endpoint is minimal. We basically just need to loop over all `vPacket` inputs and add them to the `tappsbt.InputCommitments` map in our call to `r.cfg.ChainPorter.RequestShipment`, just like we're doing now for `vPacket.Inputs[0]`.

Does that sound correct to you @guggero ?

(I'll take a look elsewhere in the code to see where else we need to make changes.)

```
	inputAsset := vPacket.Inputs[0].Asset()
	prevID := vPacket.Inputs[0].PrevID
	inputCommitment, err := r.cfg.AssetStore.FetchCommitment(
		ctx, inputAsset.ID(), prevID.OutPoint, inputAsset.GroupKey,
		&inputAsset.ScriptKey, true,
	)
	if err != nil {
		return nil, fmt.Errorf("error fetching input commitment: %w",
			err)
	}

	rpcsLog.Debugf("Selected commitment for anchor point %v, requesting "+
		"delivery", inputCommitment.AnchorPoint)

	resp, err := r.cfg.ChainPorter.RequestShipment(
		tapfreighter.NewPreSignedParcel(
			vPacket, tappsbt.InputCommitments{
				0: inputCommitment.Commitment,
			},
		),
	)
```

ffranr:
I see that we'll need to extend `FundVirtualPsbt` so that we can test our changes to `AnchorVirtualPsbts`.

ffranr:
We also need to address this TODO in this issue:
https://github.com/lightninglabs/taproot-assets/blob/0a63d81e7c0b3886161d2df6b07f6f54c1752dfd/tapscript/send.go#L514-L514

ffranr:
This PR should cover full value multi-input PSBT spend for homogenous input asset IDs: https://github.com/lightninglabs/taproot-assets/pull/741

ffranr:
I've narrowed down the scope of this issue to homogeneous input asset IDs. I'll create another issue to capture other sorts of inputs on Monday.

guggero:
> Does that sound correct to you @guggero ?

[...]

PR #741, opened 2023-12-16
Multi-input PSBT spend for homogeneous input asset ID

ffranr:
This PR solves https://github.com/lightninglabs/taproot-assets/issues/297 for the case were each input asset ID is the same.

This PR modifies our asset send procedure such that we can spend a multi-input full value PSBT where each input has the same asset ID. The PR adds an itest to test the new changes.

There are more things we should test:
* more than 2 inputs
* sending the received PSBT spend onwards once more.

But lets keep this PR small if we can and open another (if that makes sense).

### DIFF
--- a/tapscript/send.go
+++ b/tapscript/send.go
@@ -506,32 +506,51 @@ func PrepareOutputAssets(ctx context.Context, vPkt *tappsbt.VPacket) error {
 	)
 
 	if isFullValueInteractiveSend {
-		if len(inputs) != 1 {
-			return fmt.Errorf("full value interactive send " +
-				"must have exactly one input")
+		// Sum the total amount of the input assets.
+		inputsAmountSum := uint64(0)
+		for idx := range inputs {
+			input := inputs[idx]
+
+			// TODO(ffranr): At the moment, we need to ensure that
+			//  all inputs have the same asset ID. We've already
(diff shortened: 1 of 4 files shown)


tapscript/send.go
```diff
@@ -506,32 +506,51 @@ func PrepareOutputAssets(ctx context.Context, vPkt *tappsbt.VPacket) error {
...
 	if isFullValueInteractiveSend {
-		if len(inputs) != 1 {
-			return fmt.Errorf("full value interactive send " +
-				"must have exactly one input")
+		// Sum the total amount of the input assets.
+		inputsAmountSum := uint64(0)
+		for idx := range inputs {
+			input := inputs[idx]
+
+			// TODO(ffranr): At the moment, we need to ensure that
```
guggero, line 514: Does this need to be a TODO? I think we can just mention here that we do an additional check for clarity.

guggero (approved):
Very nice, LGTM :tada:

ffranr:
Replaced with https://github.com/lightninglabs/taproot-assets/pull/743

tapscript/send.go
ffranr: Sure, no problem :+1:

PR #743, opened 2023-12-19
Multi-input PSBT spend for homogeneous input asset ID

### LINKED ISSUE #297: Add full-value multi-input homogeneous asset IDs support to RPC endpoint AnchorVirtualPsbts

ffranr:
This PR solves https://github.com/lightninglabs/taproot-assets/issues/297 for the case were each input asset ID is the same.

This PR modifies our asset send procedure such that we can spend a multi-input full value PSBT where each input has the same asset ID. The PR adds an itest to test the new changes.

There are more things we should test:
* more than 2 inputs
* sending the received PSBT spend onwards once more.

But lets keep this PR small if we can and open another (if that makes sense).

This PR replaces https://github.com/lightninglabs/taproot-assets/pull/741
I've moved the branch to repo `lightninglabs/taproot-assets` so that I can base another branch on top of it.

Fixes #297

### DIFF
--- a/tapscript/send.go
+++ b/tapscript/send.go
@@ -506,32 +506,50 @@ func PrepareOutputAssets(ctx context.Context, vPkt *tappsbt.VPacket) error {
 	)
 
 	if isFullValueInteractiveSend {
-		if len(inputs) != 1 {
-			return fmt.Errorf("full value interactive send " +
-				"must have exactly one input")
+		// Sum the total amount of the input assets.
+		inputsAmountSum := uint64(0)
+		for idx := range inputs {
+			input := inputs[idx]
+
+			// At the moment, we need to ensure that all inputs have
+			// the same asset ID. We've already checked that above,
+			// but we will do it again here for clarity.
+			if inputs[idx].Asset().ID() != assetID {
+				return fmt.Errorf("multiple input assets " +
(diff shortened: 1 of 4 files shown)


tapscript/send.go
```diff
@@ -506,32 +506,51 @@ func PrepareOutputAssets(ctx context.Context, vPkt *tappsbt.VPacket) error {
...
 	if isFullValueInteractiveSend {
-		if len(inputs) != 1 {
-			return fmt.Errorf("full value interactive send " +
-				"must have exactly one input")
+		// Sum the total amount of the input assets.
+		inputsAmountSum := uint64(0)
+		for idx := range inputs {
+			input := inputs[idx]
+
+			// TODO(ffranr): At the moment, we need to ensure that
```
guggero, line 514: Same comment here re needing a TODO?

guggero (approved):
Re-ACK on new PR, LGTM :tada:

tapscript/send.go
ffranr: TODO removed.

jharveyb (approved):
Solid send improvements, LGTM 🎉

PR #746, opened 2023-12-21
version: remove dead code

ffranr:
I don't think we use this code.

### DIFF
--- a/version.go
+++ b/version.go
@@ -8,7 +8,6 @@ package taprootassets
 import (
 	"bytes"
 	"fmt"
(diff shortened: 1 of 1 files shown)


GeorgeTsagk (approved):

guggero:
Correct, we don't use this currently. But we will use it if we ever create a UI for taproot assets ourselves, then we'll want to be able to differentiate between actions from `tapcli` vs the UI, which is what the user agent is for... So we'll just add that code back in the future. Don't care either way.

PR #747, opened 2023-12-21
itest: extend multi-input PSBT spend test to include partial value spend

ffranr:
This commit extends the existing multi-input PSBT spend test to also include a partial value spend.

### DIFF
--- a/itest/psbt_test.go
+++ b/itest/psbt_test.go
@@ -954,14 +954,16 @@ func testPsbtMultiSend(t *harnessTest) {
 }
 
 // testMultiInputPsbtSingleAssetID tests to ensure that we can correctly
-// construct and spend a multi-input full value PSBT where each input has the
-// same asset ID.
+// construct and spend a multi-input partial value and full value PSBT where
+// each input has the same asset ID.
 //
 // The test works as follows:
 //  1. Mint an asset on the primary tapd node.
-//  2. Send the asset to a secondary tapd node in two different send events.
-//  3. Send the asset back to the primary tapd node in a single multi-input PSBT
-//     send event.
+//  2. Send the asset to a secondary tapd node in three different send events.
+//  3. Send a partial amount of the asset back to the primary tapd node in a
+//     single multi-input PSBT send event.
+//  4. Send the remaining amount of the asset back to the primary tapd node in
+//     a single full value multi-input PSBT send event.
 func testMultiInputPsbtSingleAssetID(t *harnessTest) {
 	var (
 		ctxb        = context.Background()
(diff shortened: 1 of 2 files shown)


itest/psbt_test.go
```diff
@@ -1022,13 +1032,38 @@ func testMultiInputPsbtSingleAssetID(t *harnessTest) {
...
 	AssertNonInteractiveRecvComplete(t.t, secondaryTapd, 2)
 
-	t.Logf("Two separate send events complete, now attempting to send " +
-		"back the full amount in a single multi-input PSBT send event")
+	// Third of three send events from primary (minting) node to the
+	// secondary node.
+	sendAmt = uint64(3000)
+	changeAmt = uint64(0)
+
+	addr, err = secondaryTapd.NewAddr(
```
guggero, line 1045: If the goal is to just create three different on-chain outputs, we _could_ use a multi-send (`sendAssetsToAddr()` supports multiple receiver addresses). Would shorten the code a bit.

guggero (approved):
Very nice! LGTM :tada:

itest/psbt_test.go
ffranr: Using `sendAssetsToAddr` I ran into problems reusing an address. (Not all output script keys were unique.) I'll investigate and try to get it working but I'll leave this PR as is for now.

jharveyb: Why would address reuse be necessary? IIUC each call to `NewAddr` should be returning addresses with unique script keys.

itest/psbt_test.go
```diff
@@ -1108,6 +1150,82 @@ func testMultiInputPsbtSingleAssetID(t *harnessTest) {
...
+	t.Logf("Partial amount multi-input PSBT send event complete. Now " +
+		"attempting to send the remaining amount in a full value " +
+		"multi-input PSBT send event back to the primary node from " +
+		"the secondary node.")
+
+	// Attempt a full value send of the rest of the asset back to the
+	// primary node.
+	sendAmt = uint64(1500)
+
+	vPkt = tappsbt.ForInteractiveSend(
```
jharveyb, line 1163: super-nit: we could assert the # of inputs on these PSBTs;  IIUC should be two in both cases?

jharveyb (approved):
LGTM, solid extension of coverage 💯

itest/psbt_test.go
```diff
@@ -1001,17 +1007,21 @@ func testMultiInputPsbtSingleAssetID(t *harnessTest) {
...
+		genInfo.AssetId, []uint64{changeAmt, sendAmt}, 0, 1,
 	)
 
 	AssertNonInteractiveRecvComplete(t.t, secondaryTapd, 1)
 
-	// Second of two send events from primary (minting) node to the
-	// secondary node.
+	// Second of three send events from primary (minting) node to secondary
+	// node.
+	sendAmt = uint64(1000)
```
GeorgeTsagk, line 1017: nit: instead of hard-coding the numbers here maybe calculate them from previous `changeAmt`?

GeorgeTsagk (approved):
Lgtm!

itest/psbt_test.go
ffranr: good idea, added

itest/psbt_test.go
ffranr: I've added a running total for more dynamic change amount calculation.

itest/psbt_test.go
ffranr: I might try again without address re-use. Thanks for pointing out the unique script key property. I'll look into that.

PR #756, opened 2024-01-04
Refactor and cleanup custodian events cache handling

ffranr:
This PR makes the following changes:
1. Refactor method `Custodian.receiveProof` such that it can be executed without a new goroutine (I see this as a simplification: putting all goroutine/sync code in the same scope).
2. Makes the custodian's event cache thread safe.
3. We then use the new thread safe event cache to cleanup events from within the courier proof retrieval goroutine.
4. Simplify the code by separating the event completion code from the `mapProofToEvent` method. In this way, `mapProofToEvent` only retrieves an event related to a given proof, and nothing more. Then, all event completion code is in `Custodian.setReceiveCompleted`.


tapgarden/custodian.go
```diff
@@ -125,9 +125,9 @@ type Custodian struct {
 	// statusEventsSubsMtx guards the general status events subscribers map.
 	statusEventsSubsMtx sync.Mutex
 
-	// events is a map of all transaction outpoints and their ongoing
-	// address events of inbound assets.
-	events map[wire.OutPoint]*address.Event
+	// events is a thread safe map of all transaction outpoints and their
+	// ongoing address events of inbound assets.
+	events sync.Map
```
guggero, line 130: Making this thread safe is only required because we want to clean up proofs in the `receiveProof()` method which is a different goroutine than the `watchInboundAssets()` main event loop.
Couldn't we instead just do:
```
c.proofSubscription.NewItemCreated.ChanIn() <- addrProof.Blob
```
at the very end of `receiveProof()` to feed it into the main event loop, removing the need for the `sync.Map` refactor that IMO makes the code a bit less readable with the type casts?

tapgarden/custodian.go
```diff
@@ -795,9 +771,72 @@ func (c *Custodian) setReceiveCompleted(event *address.Event,
...
+// address events and marking them as complete.
+func (c *Custodian) handleNewProof(newProof proof.Blob) error {
+	// Obtain proof file from a single proof blob or a proof file blob.
+	var (
+		file             *proof.File
+		proofFileVersion = proof.V0
+	)
+
+	switch {
+	case proof.IsProofFile(newProof):
```
guggero, line 806: Why is this logic needed? I think we'll only ever _not_ get a file once https://github.com/lightninglabs/taproot-assets/pull/726 is in, which already adds this code but in a slightly different way.

guggero (commented):
Thanks for the refactor!
I think the first commit definitely makes sense. But the rest IMO make the already hard to follow flow even trickier...
What do you think of my suggestion to instead just feed the new proof into the main event loop to assert cleanup of the events?

tapgarden/custodian.go
ffranr: That's a great idea! Thanks for taking a look. I've gone with the `c.proofSubscription` channel method.

tapgarden/custodian.go
ffranr: I've replaced this logic with a check to make sure that the proof blob is actually a file.

ffranr:
> Thanks for the refactor! I think the first commit definitely makes sense. But the rest IMO make the already hard to follow flow even trickier... What do you think of my suggestion to instead just feed the new proof into the main event loop to assert cleanup of the events?

Thanks for the review! I've gone with your idea. Reverted `c.events` and used the proof channel.

I'm still keen on the `mapProofToEvent` -> `handleNewProof` + `mapProofToEvent` + `setReceiveCompleted` refactor. I think `mapProofToEvent` is simplest if it only maps to an event, without side effects.

tapgarden/custodian.go
ffranr: I've just realised you used a negative "only ever _not_ get a file".

Whatever we're supposed to get at this point, I think we should add a check for it. Note that the blob to file conversion was there already, I didn't add that in this PR. I think we should check each time before we interpret a proof blob as a file or single proof.

I'll check for a single proof if https://github.com/lightninglabs/taproot-assets/pull/726 gets in first.

 I think it would make sense if the `c.proofSubscription` had a proper proof type and not just `Blob`. Or at least `SingleProofBlob` and `FileBlob`. Maybe one day a `VerifiedProof` type.

ffranr:
I've taken a look at #726 and I've reduced the scope of this PR so that I don't get in its way: https://github.com/lightninglabs/taproot-assets/pull/726/files#diff-47500066f2e4b70c65d3c7777f4478203365bdfda9f82ae6602e9e853792efeb

In other words, in this PR I've reverted my modifications to `mapProofToEvent`.

guggero (approved):
Very nice, LGTM :tada:

GeorgeTsagk (approved):
Nice clean-up, LGTM! :carrot:

PR #757, opened 2024-01-05
Add optional memo to TAP addresses

### LINKED ISSUE #751: [feature]: memo or description like field for generating TAP address
Can we have a feature that allows us to add a memo or description sort of field, to note what is the cause of creating a tap address or so.

sputn1ck:
This PR adds optional memos to TAP addresses. Memos are encoded in the address similiar to BOLT 11 Descriptions. 

I've tested it locally with a 0.3 Tap client. The implementation seems to be both backward and forward compatible.

Things I'm unsure about:
- Maximum length of description
- Where to add itests that make sense

Creation of address:
image

Decoding on old version:
image

Decoding of TAP addr without memo(addr created on tap version 0.3)
image

Fixes https://github.com/lightninglabs/taproot-assets/issues/751


guggero:
Thanks for the PR! I think having a memo is definitely useful on the receiving end to know what purpose an address was created for.
But I'm not sure whether it should be included in the address itself? Unfortunately the term is quite overloaded.
We use `--memo` for `lncli addinvoice` where the text does go into the invoice and can be seen by the sender.
But we also use `--memo` for `lncli openchannel` where it's purely a note/info for the node operator.

I would personally have expected the memo on a TAP address to follow the second pattern, where the memo is only for the creator of the address, not also the recipient.
If the intention is for this to be used like the memo in an invoice, we probably should make it more clear that it will become part of the address and anyone can decode it (so don't put privacy sensitive information in it). Also we would need to update the BIP and add the memo as a field there.

jtobin:
This one looks very stale, so going to close it for the time being.

PR #759, opened 2024-01-09
vPSBT: add sighash support (flattened virtual transactions)

### LINKED ISSUE #577: tapgarden/rpc: expose custom sighash flags for vPSBT funding to enable trustless asset swaps
### Background 

Today, we have the [vPSBT API](https://lightning.engineering/posts/2023-06-14-virtual-psbt/) which allows users to do custom spends using assets the daemon owns. Those custom spends can also have their own scripts, and also eventually custom sighash flags as well. Due to the unique nature of the TAP overlay, flags that normally are considered useless, like `SIGHASH_NONE`, are actually of use for _non-interactive_ swap scenarios. In this case, a seller can publish some data, and the "first" (subject to fee bidding, etc) buyer to add enough funds is able to claim those assets. 

#### The "One Little Trick" 🪄

The missing component to enable this today is allowing users to sign on the _TAP_ level with custom sighash flags. We also want to expose users bringing their own PSBT as well. Once this is exposed, `SIGHASH_NONE` can be used to sign "blank checks" (on the TAP layer). However, these blank checks cannot be cashed without also spending the input that encumbers the check input. That input can only be spent by completing a Bitcoin input/output "puzzle", by adding enough extra inputs to pay the seller and make the check claim valid. This process is non-custodial, and non-interactive 😎. 

### Sally The Seller's Signed Offer

Here's an example flow to illustrate the concept: 
  1. Sally decides to part with her 1-of-1 holographic beefzard card gifted to her by her grandpa so she can make a down payment on a condo in the new SpaceX Luna base. 
  2. She has an input `X`, that commits to the beefzard card (and nothing else for simplicity of this example). 
  3. She makes a PSBT (Bitcoin layer) that adds input `X`, and then an output with a bip 86 addr (doesn't commit to any assets). This 1-in-1 out input is signed with `SIGHASH_SINGLE | SIGHASH_ANYONE_CAN_PAY`. The output asks for `1 BTC` on top of the input amount.
       * At this point, the transaction isn't valid. The transaction is missing that 1 BTC input. 
  4. Sally now makes a vPSBT for the output that the buyer will use to claim the beefzard card. This vSPBT output is "loose", in that everything known is specified, but it doesn't commit to a script key, nor internal key. 
  5. Sally _signs_ (or w/e valid witness for the beefzard card) that vOutput, with `SIGHASH_NONE`. This means she doesn't commit to the script key the asset is to be sent to. This is where some of the magic happens. 
  6. Sally publishes her signed PSBT, and also the signed vPSBT. 
       * The only way someone can claim this is to:
            * Provide enough input to pay Sally. 
            * Anchor the vPSBT output into a new output under the control of the buyer. 

### Claiming the Swap Offer On-Chain

For someone to claim this:
   1. They see the offer from Sally on Nostr, or w/e.
   2. They add in their input on the Bitcoin level to complete the transaction. 
   3. They take the vOut, and anchor it in their new output, providing their own script key and internal key
   4. Broadcast, ???, profit.
      * Here ??? means they may need to compete in the mempool with the other would-be buyers. 

### Interactive & Non-Interactive Auctions 

To eliminate the ???, Alice can set up an open auction. She suspends her beefzard in a timelocked output (only needs to be on the TAP layer), then she has others send her offers, which come in the form of the 1-in-1 out described above. She can run things like a dutch auction using the `sequence` field of the execution transaction: as the price descends, so does the sequence field. Initially, only the highest price is available, then as time goes on, the lower prices are us available and unencumbered by the relative time lock.

You can make the process non-interactive again by publishing these time locked orders as outlined above.

GeorgeTsagk:
## Description

This PR adds sighash support for vPSBTs

## Todo list
- [ ] Decide on the set of sighash flags to support for witnesses
- [ ] Add sighash logic for validation
- [ ] Add extra vPSBT tests that cover the supported sighashes
- [ ] Add itest for trustless swap

Closes #577

### DIFF
--- a/asset/mock.go
+++ b/asset/mock.go
@@ -140,7 +140,7 @@ func virtualGenesisTxOut(newAsset *Asset) (*wire.TxOut, error) {
 // MockGroupTxBuilder.
 func virtualGenesisTx(newAsset *Asset) (*wire.MsgTx, error) {
 	var (
-		txIn *wire.TxIn
+		txIn []*wire.TxIn
 		err  error
 	)
 
@@ -159,7 +159,10 @@ func virtualGenesisTx(newAsset *Asset) (*wire.MsgTx, error) {
 	// With our single input and output mapped, we're ready to construct our
 	// virtual transaction.
 	virtualTx := wire.NewMsgTx(2)
-	virtualTx.AddTxIn(txIn)
+
+	for _, txIn := range txIn {
+		virtualTx.AddTxIn(txIn)
+	}
 	virtualTx.AddTxOut(txOut)
 	return virtualTx, nil
 }
--- a/asset/tx.go
+++ b/asset/tx.go
@@ -59,16 +59,16 @@ func VirtualTxWithInput(virtualTx *wire.MsgTx, input *Asset,
 
 	txCopy := virtualTx.Copy()
 	txCopy.LockTime = uint32(input.LockTime)
-	txCopy.TxIn[zeroIndex].PreviousOutPoint.Index = idx
-	txCopy.TxIn[zeroIndex].Sequence = uint32(input.RelativeLockTime)
-	txCopy.TxIn[zeroIndex].Witness = witness
+	txCopy.TxIn[idx].PreviousOutPoint.Index = idx
+	txCopy.TxIn[idx].Sequence = uint32(input.RelativeLockTime)
+	txCopy.TxIn[idx].Witness = witness
 	return txCopy
 }
 
 // VirtualGenesisTxIn computes the single input of a Taproot Asset virtual
 // transaction that represents a grouped asset genesis. The input prevout's hash
 // is the root of a MS-SMT committing to only the genesis asset.
-func VirtualGenesisTxIn(newAsset *Asset) (*wire.TxIn, mssmt.Tree, error) {
+func VirtualGenesisTxIn(newAsset *Asset) ([]*wire.TxIn, mssmt.Tree, error) {
 	inputTree := mssmt.NewCompactedTree(mssmt.NewDefaultStore())
 
 	// TODO(bhandras): thread the context through.
@@ -100,7 +100,11 @@ func VirtualGenesisTxIn(newAsset *Asset) (*wire.TxIn, mssmt.Tree, error) {
 
 	prevOut := VirtualTxInPrevOut(treeRoot)
 
-	return wire.NewTxIn(prevOut, nil, nil), inputTree, nil
+	txIns := []*wire.TxIn{
+		wire.NewTxIn(prevOut, nil, nil),
+	}
+
+	return txIns, inputTree, nil
 }
 
 // GenesisPrevOutFetcher returns a Taproot Asset input's `PrevOutFetcher` to be
--- a/tapscript/mint.go
+++ b/tapscript/mint.go
@@ -31,7 +31,7 @@ func BuildGenesisTx(newAsset *asset.Asset) (*wire.MsgTx,
 
 	// Now, create the virtual transaction that represents this asset
 	// minting.
-	virtualTx, _, err := VirtualTx(newAsset, nil)
+	virtualTx, _, err := VirtualTx(newAsset, nil, []*asset.Asset{newAsset})
 	if err != nil {
 		return nil, nil, err
 	}
--- a/tapscript/send.go
+++ b/tapscript/send.go
@@ -648,9 +648,14 @@ func SignVirtualTransaction(vPkt *tappsbt.VPacket, signer Signer,
 		prevAssets[input.PrevID] = input.Asset()
 	}
 
+	assetOutputs := make([]*asset.Asset, len(outputs))
+	for idx := range outputs {
+		assetOutputs[idx] = outputs[idx].Asset
+	}
+
 	// Create a Taproot Asset virtual transaction representing the asset
 	// transfer.
-	virtualTx, _, err := VirtualTx(newAsset, prevAssets)
+	virtualTx, _, err := VirtualTx(newAsset, prevAssets, assetOutputs)
 	if err != nil {
 		return err
 	}
@@ -669,7 +674,7 @@ func SignVirtualTransaction(vPkt *tappsbt.VPacket, signer Signer,
 		// Sign the virtual transaction based on the input script
 		// information (key spend or script spend).
 		newWitness, err := CreateTaprootSignature(
-			input, inputSpecificVirtualTx, 0, signer,
+			input, inputSpecificVirtualTx, idx, signer,
 		)
 		if err != nil {
 			return fmt.Errorf("error creating taproot "+
--- a/tapscript/tx.go
+++ b/tapscript/tx.go
@@ -40,13 +40,79 @@ const (
 	zeroIndex = 0
 )
 
-// virtualTxIn computes the single input of a Taproot Asset virtual transaction.
-// The input prevout's hash is the root of a MS-SMT committing to all inputs of
-// a state transition.
-func virtualTxIn(newAsset *asset.Asset, prevAssets commitment.InputSet) (
-	*wire.TxIn, mssmt.Tree, error) {
+// virtualTxInputTree places all of the inputs of a virtual transaction that
+// represents a taproot assets state transition into a MS-SMT. This tree is
+// returned as the result.
+func virtualTxInputTree(newAsset *asset.Asset, prevAssets commitment.InputSet) (
+	mssmt.Tree, error) {
 
 	inputTree := mssmt.NewCompactedTree(mssmt.NewDefaultStore())
+	// For each input we'll locate the asset UTXO being spent, then
+	// insert that into a new SMT, with the key being the hash of
+	// the prevID pointer, and the value being the leaf itself.
+	inputsConsumed := make(
+		map[asset.PrevID]struct{}, len(prevAssets),
+	)
+
+	// TODO(bhandras): thread the context through.
+	ctx := context.TODO()
+
+	for _, input := range newAsset.PrevWitnesses {
+		// At this point, each input MUST have a prev ID.
+		if input.PrevID == nil {
+			return nil, fmt.Errorf("%w: prevID is nil",
+				ErrNoInputs)
+		}
+
+		// The set of prev assets are similar to the prev
+		// output fetcher used in taproot.
+		prevAsset, ok := prevAssets[*input.PrevID]
+		if !ok {
+			return nil, fmt.Errorf("%w: unable to make "+
+				"virtual txIn %v", ErrNoInputs,
+				spew.Sdump(input.PrevID))
+		}
+
+		// Now we'll insert this prev asset leaf into the tree.
+		// The generated leaf includes the amount of the asset,
+		// so the sum of this tree will be the total amount
+		// being spent.
+		key := input.PrevID.Hash()
+		leaf, err := prevAsset.Leaf()
+		if err != nil {
+			return nil, err
+		}
+		_, err = inputTree.Insert(ctx, key, leaf)
+		if err != nil {
+			return nil, err
+		}
+
+		inputsConsumed[*input.PrevID] = struct{}{}
+	}
+
+	// In this context, the set of referenced inputs should match
+	// the set of previous assets. This ensures no duplicate inputs
+	// are being spent.
+	//
+	// TODO(roasbeef): make further explicit?
+	if len(inputsConsumed) != len(prevAssets) {
+		return nil, ErrInputMismatch
+	}
+
+	return inputTree, nil
+}
+
+// virtualTxIns computes the inputs of a Taproot Asset virtual transaction. Each
+// input's prevout hash is the root of a MS-SMT containing the asset input at
+// the same index.
+func virtualTxIns(newAsset *asset.Asset, prevAssets commitment.InputSet) (
+	[]*wire.TxIn, mssmt.Tree, error) {
+
+	var (
+		txIns    = make([]*wire.TxIn, len(newAsset.PrevWitnesses))
+		aggrTree = mssmt.NewCompactedTree(mssmt.NewDefaultStore())
+	)
+
 	// For each input we'll locate the asset UTXO being spent, then
 	// insert that into a new SMT, with the key being the hash of
 	// the prevID pointer, and the value being the leaf itself.
@@ -82,12 +148,33 @@ func virtualTxIn(newAsset *asset.Asset, prevAssets commitment.InputSet) (
 		if err != nil {
 			return nil, nil, err
 		}
+
+		// Create the ms smt that will hold the input.
+		inputTree := mssmt.NewCompactedTree(mssmt.NewDefaultStore())
+
+		// Insert the leaf into the tree holding just this input.
 		_, err = inputTree.Insert(ctx, key, leaf)
 		if err != nil {
 			return nil, nil, err
 		}
 
+		// Add the leaf to the aggregate input tree, which is going to
+		// hold all the inputs.
+		_, err = aggrTree.Insert(ctx, key, leaf)
+		if err != nil {
+			return nil, nil, err
+		}
+
 		inputsConsumed[*input.PrevID] = struct{}{}
+
+		treeRoot, err := inputTree.Root(context.Background())
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// TODO(roasbeef): document empty hash usage here
+		prevOut := asset.VirtualTxInPrevOut(treeRoot)
+		txIns = append(txIns, wire.NewTxIn(prevOut, nil, nil))
 	}
 
 	// In this context, the set of referenced inputs should match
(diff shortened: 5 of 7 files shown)


tappsbt/interface.go
```diff
@@ -40,6 +40,7 @@ var (
 	PsbtKeyTypeInputTapAnchorTapscriptSibling             = []byte{0x78}
 	PsbtKeyTypeInputTapAsset                              = []byte{0x79}
 	PsbtKeyTypeInputTapAssetProof                         = []byte{0x7a}
+	PsbtKeyTypeInputTapSigHashType                        = []byte{0x7b}
```
guggero, line 43: We don't need to encode this separately. Because the `i.SighashType` field is on the embedded `psbt.PInput`, it gets serialized automatically.

tapscript/tx.go
```diff
@@ -165,10 +167,25 @@ func virtualTxOut(txAsset *asset.Asset) (*wire.TxOut, error) {
...
+		// TODO(bhandras): thread the context through.
+		tree, err = outputTree.Insert(context.TODO(), key, leaf)
+		if err != nil {
+			return nil, err
+		}
+
+	case txscript.SigHashNone:
+		nullKey := [32]byte{}
+		nullLeaf := mssmt.LeafNode{}
+		tree, err = outputTree.Insert(context.TODO(), nullKey, &nullLeaf)
```
guggero, line 185: I think instead of inserting a null key/leaf, we just skip inserting anything? Leading to an empty tree instead? Since the null key is never really looked up anyway.

vm/vm.go
```diff
@@ -280,6 +280,20 @@ func (vm *Engine) validateStateTransition(virtualTx *wire.MsgTx) error {
 	}
 
 	for i, witness := range vm.newAsset.PrevWitnesses {
+		// TODO(george): Dummy assignment, need to figure out how to get
+		// the sighash byte from the signature.
+		sighashType := txscript.SigHashType(witness.TxWitness[0][33])
```
guggero, line 285: Convention is: If the signature is exactly 64 bytes, it's an implied `SigHashDefault`. If it's 65 bytes, it is a different sighash and the 65th byte is the sighash byte.
We probably also need some logic to find out which entry of the witness is an actual signature.

Hmm... and that just get me thinking. Maybe we need a different way of transporting the sighash flag than by appending it to the witness. Because we look at the sighash on the level "vPSBT -> dummy TX". But the underlying Taproot execution VM will look at the sighash on the level "dummy TX -> txsighash", where we'll probably want to always use SigHashDefault. So the actual signature shouldn't contain the sighash as on the lowest level we always want it to be SigHashDefault. But when validating the asset, we need to know what the "vPSBT -> dummy TX" sighash was. So we need to transport it somewhere.

Or maybe this is an argument for actually mapping the vPSBT inputs/outputs directly into dummy TX inputs/outputs, then we can use the same sighash logic on both levels.

guggero (commented):
Very nice, I think you're on the right track!

Though I think we might need the new script version @Roasbeef mentioned yesterday and map out the inputs/outputs directly instead of as a 1-in-one-out. See inline comment.

vm/vm.go
GeorgeTsagk: Right, thanks for writing this up, seems like there's loss of information by using the 1-1 dummy tx if we want to use more sighash flags.

One thing that I hadn't realized is that when the vPSBT is no longer around, a future verifier will have to rely on the [`ChallengeWitness`](https://github.com/lightninglabs/taproot-assets/blob/main/proof/proof.go#L297) field of `Proof` in order to check that a past transfer was valid. That `ChallengeWitness` currently is the single 1-1 dummy tx (which relied on tap always using "sighash all"). This means that the ephemeral dummy txs constructed by each vInput in order for the backend to sign will no longer be reconstructible by the verifier in order to check the signature.

As you mention, we can expand the `ChallengeWitness` and have multiple inputs and outputs. I believe we can align the inputs and outputs so that at input[i] / output[i] you'll find the roots of the trees into which the inputs and outputs were inserted, with respect to **that input's sighash flag**. So if on the vPSBT level we have 10 vInputs and 10 vOutputs, if the first (index `0`) vInput wants to use a sighash that commits to a single input and all the outputs then on the dummy tx level we'd have `input[0]` be produced by inserting to an ms-smt just `vInput[0]` and `output[0]` would be produced by inserting all outputs in an ms-smt.

Can we go even further than that and do a "raw" mapping?

We can have each input in the dummy tx be produced by inserting into an ms-smt just the one vInput that has the same index. We can do exactly the same for the outputs, with each one being produced by an ms-smt that only contains the single vOutput on the same index. At that point can we just leverage the sighashes on the btc level? Nobody knows that the inputs/outputs are dummy, but the signatures can normally sign them in any sighash fashion already available to us via the backend.

If the latter works I believe it's the preferable approach?
LMK what you think

guggero: I don't think this is an issue and is very likely just a misunderstanding around the `ChallengeWitness`.
That field is a special field that is only set in ownership proof files. An ownership proof is only created if you want/need to prove to someone that you do have the asset-level key to spend an asset. So it is just a signature showing you _could_ spend the asset, without actually spending it (so it's not a valid transfer proof).

So the field is only set for the output of `ProveAssetOwnership` and can only be verified with `VerifyAssetOwnership`.
The 1-in-1 out mentioned there is actually on the vPSBT level (not the dummy TX level), so it won't be affected by any changes we make due to the sighash implementation. Also, for the asset ownership proof we'll likely always use `SigHashDefault`.

guggero: > We can have each input in the dummy tx be produced by inserting into an ms-smt just the one vInput that has the same index. We can do exactly the same for the outputs, with each one being produced by an ms-smt that only contains the single vOutput on the same index. At that point can we just leverage the sighashes on the btc level? Nobody knows that the inputs/outputs are dummy, but the signatures can normally apply to them in any sighash fashion already available to us via the backend.

Yes, I think that is what we want. We'd do a raw mapping and then just use the BTC-VM level sighash functionality.

vm/vm.go
```diff
@@ -296,6 +322,14 @@ func (vm *Engine) validateStateTransition(virtualTx *wire.MsgTx) error {
 			if err != nil {
 				return err
 			}
+		case asset.ScriptV1:
```
GeorgeTsagk, line 325: Main motivation here is that `vm.ValidateWitnessV0` relies on constructing the virtual transaction with single input / single output. We can't break that now

tapscript/tx.go
```diff
@@ -99,125 +186,146 @@ func virtualTxIn(newAsset *asset.Asset, prevAssets commitment.InputSet) (
...
+	var (
+		txOuts = make([]*wire.TxOut, 0, len(assetOutputs))
+	)
 
-	return wire.NewTxIn(prevOut, nil, nil), inputTree, nil
-}
+	for _, txAsset := range assetOutputs {
+		// If we have any asset splits, then we'll indirectly commit to
+		// all of them through the SplitCommitmentRoot.
+		if txAsset.SplitCommitmentRoot != nil {
```
GeorgeTsagk, line 203: This can prove problematic for some sighash types. As this output implicitly commits to all outputs, altering the desired effect of the sighash

asset/mock.go
```diff
@@ -140,7 +140,7 @@ func virtualGenesisTxOut(newAsset *Asset) (*wire.TxOut, error) {
 // MockGroupTxBuilder.
 func virtualGenesisTx(newAsset *Asset) (*wire.MsgTx, error) {
 	var (
-		txIn *wire.TxIn
+		txIn []*wire.TxIn
```
guggero, line 143: nit: rename to `txIns` to not shadow the name in the loop below (`for _, txIn := range txIn`)?

tapscript/mint.go
```diff
@@ -31,7 +31,7 @@ func BuildGenesisTx(newAsset *asset.Asset) (*wire.MsgTx,
 
 	// Now, create the virtual transaction that represents this asset
 	// minting.
-	virtualTx, _, err := VirtualTx(newAsset, nil)
+	virtualTx, _, err := VirtualTx(newAsset, nil, []*asset.Asset{newAsset})
```
guggero, line 34: This looks a bit confusing, as if the inputs and outputs were the same... Maybe it would be a bit more clear if the first parameter would just be the `[]PreviousWitness` slice, since it looks like only that is being used for the input part of the virtual TX?

tapscript/send.go
```diff
@@ -648,9 +648,14 @@ func SignVirtualTransaction(vPkt *tappsbt.VPacket, signer Signer,
 		prevAssets[input.PrevID] = input.Asset()
 	}
 
+	assetOutputs := make([]*asset.Asset, len(outputs))
```
guggero, line 651: nit: could make use of the `fn.Map()` function here. Doesn't really make it shorter but makes review a bit easier IMO (since creation of the slice with `make` is abstracted away).

tapscript/tx.go
guggero: Yeah, you're right. I think we won't be able to support the `SIGHASH_SINGLE` use case on the asset level. Because when you create the outputs, you need to know _all_ outputs, otherwise you can't create a split commitment.
But looking at the example given in https://github.com/lightninglabs/taproot-assets/issues/577 it seems like we mostly want the `SIGHASH_NONE` functionality (at least on the output side). So that should still be achievable.

guggero (commented):
I think you're on the right track! Nice work so far :100: Left a couple of comments.

dstadulis:
Summarizing out-of-band updates:

PR will continue to implement:
**Spreading out the pseudo-transactions inputs/outputs s.t. they directly match those in the vPSBT level.**
Even though split commitment is a thorn atm w.r.t allowing to commit to individual outputs, we can still proceed with sighashes that make sense with what we have (sighash_all, sighash_none, anyone can pay)

## Status
- [ ] Add support for sighash checks
@GeorgeTsagk starting:
- [ ] Add itest that follows vPSBT signing flow and checks sighash_all / sighash_none

asset/mock.go
```diff
@@ -159,7 +159,10 @@ func virtualGenesisTx(newAsset *Asset) (*wire.MsgTx, error) {
 	// With our single input and output mapped, we're ready to construct our
 	// virtual transaction.
 	virtualTx := wire.NewMsgTx(2)
-	virtualTx.AddTxIn(txIn)
+
```
Roasbeef, line 162: So this can't be done unconditionally. We need to make a new TAP VM version, then flatten the virtual transaction _only_ for that version.

asset/mock.go
```diff
@@ -159,7 +159,10 @@ func virtualGenesisTx(newAsset *Asset) (*wire.MsgTx, error) {
...
+	for _, txIn := range txIn {
```
Roasbeef, line 163: FWIW, coming back to this, I think I'm still not _totally_ convinced we need to do this. Input level conditional signing can happen on the Bitcoin layer, in that you use the normal sig hash flags there to bind your Bitcoin level signature to the presence of one or all inputs. 

In TAP, we have a sort of output centric model: you validate the output (asset leaf TLV), and the witness data is also stored in the output. You can control the environment needed to validate that output based on the logical spend group. The set of inputs here can _only_ ever be the set of inputs referenced in the witness data for the output. Each input needs a witness, and is effectively always consolidating to that single output (ignoring splits for a second). 

With the framing above, in what case would we ever need to sign only _some_ of the inputs in that spend group? Changing the set of inputs actually changes the output, so those need to be fully bound. We don't have the ability to have loose binding of the inputs. 

If you consider splits, then that's an area that I think we can already have covered as the sighash flag can determine _which_ of the split out puts are inserted into the split output tree. For all, do everything. For none, do nothing. You can make single work somewhat

With all that said, I think it makes sense to make this an option, if we have indeed found a flow that can't work w/o a flattened virtual txn. Why wouldn't we also flatten the tx out as well? 

As mentioned above, this requires updates elsewhere.

asset/tx.go
```diff
@@ -59,16 +59,16 @@ func VirtualTxWithInput(virtualTx *wire.MsgTx, input *Asset,
 
 	txCopy := virtualTx.Copy()
 	txCopy.LockTime = uint32(input.LockTime)
-	txCopy.TxIn[zeroIndex].PreviousOutPoint.Index = idx
-	txCopy.TxIn[zeroIndex].Sequence = uint32(input.RelativeLockTime)
-	txCopy.TxIn[zeroIndex].Witness = witness
+	txCopy.TxIn[idx].PreviousOutPoint.Index = idx
```
Roasbeef, line 62: So if we're not doing the 1-in-1-out version, then we don't need to swap out information for each input. Instead, the virtual tx has all the inputs, and `idx` here is just passed to sign/verify.

asset/tx.go
```diff
@@ -100,7 +100,11 @@ func VirtualGenesisTxIn(newAsset *Asset) (*wire.TxIn, mssmt.Tree, error) {
 
 	prevOut := VirtualTxInPrevOut(treeRoot)
 
-	return wire.NewTxIn(prevOut, nil, nil), inputTree, nil
+	txIns := []*wire.TxIn{
```
Roasbeef, line 103: I don't think this needs to change. There's only ever a single virtual txin still.

tapscript/mint.go
Roasbeef: Yeah why do we need to pass in the asset twice? I don't see why we need to make any changes here. I think also shows the rationale re the design for the splits: the root split includes the commitment to them all, so you only need to pass around the root split for validation purposes.

tapscript/send.go
```diff
@@ -648,9 +648,14 @@ func SignVirtualTransaction(vPkt *tappsbt.VPacket, signer Signer,
...
+	for idx := range outputs {
+		assetOutputs[idx] = outputs[idx].Asset
+	}
+
 	// Create a Taproot Asset virtual transaction representing the asset
 	// transfer.
-	virtualTx, _, err := VirtualTx(newAsset, prevAssets)
+	virtualTx, _, err := VirtualTx(newAsset, prevAssets, assetOutputs)
```
Roasbeef, line 658: See above re asset outputs (they're already enumerated via the split commitment root).

tapscript/tx.go
```diff
@@ -57,6 +57,72 @@ func virtualTxIn(newAsset *asset.Asset, prevAssets commitment.InputSet) (
...
+		return nil, ErrInputMismatch
+	}
+
+	return inputTree, nil
+}
+
+// virtualTxIns computes the inputs of a Taproot Asset virtual transaction. Each
+// input's prevout hash is the root of a MS-SMT containing the asset input at
+// the same index.
+func virtualTxIns(newAsset *asset.Asset, prevAssets commitment.InputSet) (
```
Roasbeef, line 108: Func above looks to do pretty much the same thing?

tapscript/tx.go
Roasbeef: > Yeah, you're right. I think we won't be able to support the SIGHASH_SINGLE use case on the asset level.

Agreed. Though I think it doesn't matter, as the root bitcoin signature is what actually binds everything. See my comment above. We don't need to enumerate the outputs as the split does that for us already.

tapscript/tx.go
```diff
@@ -99,125 +186,146 @@ func virtualTxIn(newAsset *asset.Asset, prevAssets commitment.InputSet) (
...
-	virtualTx.AddTxOut(txOut)
+
+	// We add each input as a standard transaction input.
+	for _, txIn := range txIns {
+		virtualTx.AddTxIn(txIn)
+	}
+
+	// We add each output as a standard transaction output.
+	for _, txOut := range txOuts {
+		virtualTx.AddTxOut(txOut)
```
Roasbeef, line 326: Doesn't match the mock version in the tests above.

vm/vm.go
```diff
@@ -279,6 +279,32 @@ func (vm *Engine) validateStateTransition(virtualTx *wire.MsgTx) error {
 		return newErrKind(ErrInvalidTransferWitness)
 	}
 
+	// Initialite an array holding all the assets that appear in this state
+	// transition outputs.
+	assetOutputs := make([]*asset.Asset, 0, len(vm.splitAssets)+1)
+
+	// Add the root asset to the list of outputs.
+	// TODO(george): was the root asset placed 1st on the signer's side? is
```
Roasbeef, line 287: Yes, it's always first in the split commitment tree.

vm/vm.go
```diff
@@ -279,6 +279,32 @@ func (vm *Engine) validateStateTransition(virtualTx *wire.MsgTx) error {
...
+	// mean that following loop is a noop?
+	for _, splitAsset := range vm.splitAssets {
+		assetOutputs = append(assetOutputs, &splitAsset.Asset)
+	}
+
+	// Construct the virtual transaction that this input signed,
+	// based on the sighash flag of the signature that we acquired
+	// above.
+	virtualTx, _, err := tapscript.VirtualTx(
+		vm.newAsset, vm.prevAssets, assetOutputs,
```
Roasbeef, line 302: This should take the script version as input. We can't break all the old versions (but again, I don't think this level of changes are needed: output sighashes are all or nothing).

vm/vm.go
Roasbeef: See above. We can't unconditionally make the new tree.

Roasbeef (changes requested):

GeorgeTsagk:
Closing this, replaced by #779.

Issue #714, opened 2023-12-04
[feature]: Allow filtering in ListTransfers


FLiotta:
**Is your feature request related to a problem? Please describe.**
As an exchange who makes many transfers a day, we currently have no easy way to filter transfered assets other than requesting all transfers and then filtering client-side (which potentially can't scale)

**Describe the solution you'd like**
The ability to filter by `anchor_tx_hash` (or something alike) would help a lot

**Describe alternatives you've considered**
We're currently fetching all ListTransfers and filtering clientside

PR #761, opened 2024-01-09
taprpc+tapdb: `ListTransfers` RPC optionally filters on anchor tx hash

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/714

This commit enhances the `ListTransfers` RPC endpoint by adding support for filtering on anchor transaction hash.

This commit also extends the asset store `QueryParcels` method such that it also filters on anchor transaction hash.


tapdb/assets_store_test.go
```diff
@@ -1311,6 +1311,13 @@ func TestAssetExportLog(t *testing.T) {
 	require.Equal(t, 1, len(parcels))
 	require.Equal(t, spendDelta, parcels[0])
 
+	// We should also be able to query for the parcel when filtering on its
+	// anchor transaction hash.
+	parcels, err = assetsStore.QueryParcels(ctx, &anchorTxHash, true)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(parcels))
```
guggero, line 1318: nit: can use `require.Len()`.

taprpc/taprootassets.proto
```diff
@@ -460,6 +460,10 @@ message ListBalancesResponse {
 }
 
 message ListTransfersRequest {
+    // anchor_tx_hash specifies the hash of the anchor transaction for which to
+    // retrieve transfers. An empty value indicates that this parameter should
+    // be disregarded in transfer selection.
+    bytes anchor_tx_hash = 1;
```
guggero, line 466: I'm not sure if we should define this as a string. Otherwise when using REST, you would need to reverse the byte order in the URL. Or perhaps we can do:
```
oneof anchor_tx {
   bytes anchor_tx_hash = 1;
   string anchor_txid_str = 2;
}
```

And then in the REST mapping we could do:
```
    - selector: taprpc.TaprootAssets.ListTransfers
      get: "/v1/taproot-assets/assets/transfers"
      additional_bindings:
        - get: "/v1/taproot-assets/assets/transfers/{anchor_tx.anchor_txid_str}"
```

guggero (changes requested):
Very nice! One comment about the REST API, everything else looks good.

guggero (approved):
Nice, LGTM :tada: 

Non blocking, but we could still add the additional REST mapping for convenience:
```
      additional_bindings:
        - get: "/v1/taproot-assets/assets/transfers/{anchor_txid}"
```

ffranr:
@guggero thanks for the reviews, I've added the additional bindings.

GeorgeTsagk (approved):
Lgtm :heavy_check_mark:

PR #765, opened 2024-01-11
[tap channels preparation 1/?]: Add QueryInternalKey and QueryScriptKey RPCs, pre-fill script key info in vPSBT

guggero:
A couple of preparatory commits extracted from the TAP channel simulation branch.


ffranr (commented):
That was easy to review, thanks!

I think I'll be able to approve quickly on Monday.

perms/perms.go
```diff
@@ -108,6 +108,14 @@ var (
 			Entity: "assets",
 			Action: "write",
 		}},
+		"/assetwalletrpc.AssetWallet/QueryInternalKey": {{
+			Entity: "assets",
+			Action: "read",
+		}},
+		"/assetwalletrpc.AssetWallet/QueryScriptKey": {{
+			Entity: "assets",
+			Action: "rad",
```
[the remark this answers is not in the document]
guggero: Haha, nice catch!

rpcserver.go
```diff
@@ -1841,6 +1842,146 @@ func (r *rpcServer) NextScriptKey(ctx context.Context,
...
+			internalKey = oddParity(internalKey)
+
+			keyLocator, err = r.cfg.AssetWallet.FetchInternalKey(
+				ctx, internalKey,
+			)
+			if err != nil {
+				return nil, fmt.Errorf("error fetching "+
+					"internal key: %w", err)
+			}
+		} else if err != nil {
```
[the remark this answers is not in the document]
guggero: Changed to `switch`. Unfortunately the `fallthrough` keyword can only be used if it's the first statement of a `case` body.

rpcserver.go
```diff
@@ -1841,6 +1842,146 @@ func (r *rpcServer) NextScriptKey(ctx context.Context,
...
+				err)
+		}
+
+		tweakedKey, err = r.cfg.AssetWallet.FetchScriptKey(
+			ctx, scriptKey,
+		)
+
+		// If the key can't be found with the even parity, we'll try
+		// the odd parity.
+		if errors.Is(err, address.ErrScriptKeyNotFound) {
```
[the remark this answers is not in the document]
guggero: No, it is done after the `errors.Is()`. If it is a "not found" error, we don't want to handle the error but re-try first. For any other error, we won't go into the retry-logic.

rpcserver.go
```diff
@@ -4525,3 +4666,16 @@ func MarshalAssetFedSyncCfg(
 		AllowSyncExport: config.AllowSyncExport,
 	}, nil
 }
+
+// oddParity turns the given public key parsed as a Schnorr x-only public key
+// from the even parity into its odd parity counterpart.
+func oddParity(evenParityKey *btcec.PublicKey) *btcec.PublicKey {
```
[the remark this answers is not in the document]
guggero: I agree about moving it to `btced` at some point. But I don't want to have the whole review-merge-push-tag dependency on this PR for 6 lines of code.

Good idea with the parity flip though, implemented that and moved the function into `tapscript`.

tapfreighter/wallet.go
```diff
@@ -1494,6 +1514,24 @@ func (f *AssetWallet) SignOwnershipProof(
...
+}
+
+// FetchInternalKey attempts to fetch the key locator information for the given
+// raw internal key. If the key cannot be found, then
+// address.ErrInternalKeyNotFound is returned.
+func (f *AssetWallet) FetchInternalKey(ctx context.Context,
+	rawKey *btcec.PublicKey) (keychain.KeyLocator, error) {
+
+	return f.cfg.AddrBook.FetchInternalKey(ctx, rawKey)
+}
```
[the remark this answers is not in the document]
guggero: I think the address book fits for now, since those keys are mostly created explicitly for addresses. But the `rpcserver` doesn't currently get access to the `TapAddressBook` instance. And since the wallet also needs access to those methods, I decided to just expose them there.
I guess we can refactor those responsibilities and dependencies between components in a separate PR.

guggero:
Thank you for the thorough review, @ffranr. I addressed all your comments.

tapscript/util.go
```diff
@@ -36,6 +36,18 @@ func PayToTaprootScript(taprootKey *btcec.PublicKey) ([]byte, error) {
...
+// FlipParity turns the given public key from even to odd parity or vice versa.
+func FlipParity(evenParityKey *btcec.PublicKey) *btcec.PublicKey {
+	keyCompressed := evenParityKey.SerializeCompressed()
+	keyCompressed[0] ^= 1
+
+	// We already know the even key is a valid point on the curve, so we
+	// don't need to check the error here as the flipped key will also be
+	// valid.
+	oddKey, _ := btcec.ParsePubKey(keyCompressed)
+	return oddKey
```
ffranr, line 48: You use the names `evenParityKey` and `oddKey` here. But they could be generalised further because the argument key doesn't have to be even. `evenParityKey -> key`, and `oddKey -> flippedParityKey` ?

The comment in this function says "We already know the even key is a valid point on the curve". If this function is to be used generally, we might not know that. So we might want the comment to state an assumption here. Or check the error.

ffranr (approved):
:+1: 

Added a comment on generalisation, otherwise I think this looks good.

tapscript/util.go
guggero: Oops, fixed.

rpcserver.go
```diff
@@ -1722,6 +1722,43 @@ func (r *rpcServer) SignVirtualPsbt(_ context.Context,
...
+		}
+
+		scriptKey := input.Asset().ScriptKey
+
+		// If the full tweaked script key isn't set on the asset, we
+		// need to look it up in the local database, to make sure we'll
+		// be able to sign for it.
+		if scriptKey.TweakedScriptKey == nil {
+			tweakedScriptKey, err := r.cfg.AssetWallet.FetchScriptKey(
+				ctx, scriptKey.PubKey,
```
Roasbeef, line 1742: So we're making a distinction here between the internal key of the script key, and the final taproot output level public key? 

IIRC, this is for situations where some other related wallet knows the internal key, but _not_ the final tweaked version which might have an actual script leaf commitment?

asset/asset.go
```diff
@@ -426,7 +426,7 @@ func (w *Witness) Decode(r io.Reader) error {
 }
 
 // DeepEqual returns true if this witness is equal with the given witness.
-func (w *Witness) DeepEqual(o *Witness) bool {
+func (w *Witness) DeepEqual(assetVersion Version, o *Witness) bool {
```
Roasbeef, line 429: Perhaps the tests should be what changes here, or we make a new method for this specific short circuit behavior. I don't think the caller should be forced to know which asset version is used, as the function just seeks to do a direct comparison between the two witness fields.

fn/func.go
```diff
@@ -30,6 +30,27 @@ func Map[I, O any, S []I](s S, f func(I) O) []O {
 	return output
 }
 
+// Keys returns a slice containing all the keys of the given map.
+func Keys[T any, K comparable](s map[K]T) []K {
```
Roasbeef, line 34: This and the below are in an `x` package in the std lib: https://pkg.go.dev/golang.org/x/exp/maps#Values

Roasbeef (changes requested):

dstadulis:
@guggero should @ffranr review the latest commit or are more commits needed?

guggero:
Doesn't need a re-review, nothing major changed other than addressing @ffranr's comments.

asset/asset.go
guggero: Okay, I refactored this slightly. Not sure about the naming, but is this what you had in mind?

fn/func.go
guggero: Ah yeah, I keep forgetting. Cool, removed this whole commit.

guggero:
Comments addressed, should be good for final round of review.

Roasbeef (approved):
LGTM 🐍

Issue #766, opened 2024-01-15
[feature]: create `StreamHandler` RFQ service (send/recv lnd custom msg)


ffranr:
Parent issue: https://github.com/lightninglabs/taproot-assets/issues/683

In this issue we will add the functionality necessary to subscribe and receive request for quote (RFQ) messages from LND. These messages will be passed between tapd peers via their lnd instances.

This work will involve the creation of a `RfqStream` component:

> This module will use the `SendCustomMessage` and `SubscribeCustomMessages` [LND] RPC calls to funnel the RFQ messages (which use the custom message range) to the `RfqNegotiator`. It should maintain book keeping to keep track of the `RfqNegotiator` instance for all the active assets the daemon cares about at the time. It'll also need to properly map the requests to responses as well (see the `rfq_id` value in the messages).

The goal of this issue is to get to a point where we can write integration tests that send and receive RFQ messages via LND.

ffranr:
I'm going to use 20116 as the custom message type. Index positions of letters "t", "a", and "p" concatenated.

ffranr:
I noticed that LND sends custom messages with low priority and lazily:
https://github.com/lightningnetwork/lnd/blob/e7687005141701b9a609d12d53a0e84c713370a9/server.go#L4687-L4687

We might want to consider adding message priority customisation to the `SendCustomMessage` LND RPC endpoint so that we can better ensure timely tap RFQ message exchange.

ffranr:
Spun out the wire messages into a new issue: https://github.com/lightninglabs/taproot-assets/issues/780

PR #767, opened 2024-01-16
A minimal version of the RFQ system

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/766

This PR adds a minimal version of the RFQ system.


rfqservice/quote_negotiator.go
```diff
@@ -0,0 +1,77 @@
...
+		ContextGuard: &fn.ContextGuard{
+			DefaultTimeout: DefaultTimeout,
+			Quit:           make(chan struct{}),
+		},
+	}, nil
+}
+
+// HandleIncomingQuoteRequest handles an incoming quote request.
+func (h *QuoteNegotiator) HandleIncomingQuoteRequest(_ msg.QuoteRequest) error {
+	// TODO(ffranr): Push quote request onto queue. We will need to handle
```
guggero, line 47: So the pattern here would be to have two `select` statements. One for sending the request into `h.incomingQuoteRequests` and one for reading from a result queue. Both `select`s would also read on `h.quit` to make them shutdown when required.
Or we pass in a response channel with the request and wait for an answer there, similar to how we do it in https://github.com/lightninglabs/taproot-assets/blob/d8d50878267a35011c83322ef2d89ee3909a47bb/tapgarden/planter.go#L729.

rfqservice/rfq.go
```diff
@@ -0,0 +1,234 @@
...
+				// event loop exits.
+				err = m.stopSubsystems()
+				if err != nil {
+					log.Errorf("Error stopping RFQ "+
+						"subsystems: %v", err)
+				}
+
+				// The context can now be cancelled as all
+				// dependant components have been stopped.
+				defer cancel()
```
guggero, line 88: nit: defer within defer.

rfqservice/rfq.go
```diff
@@ -0,0 +1,234 @@
...
+			}()
+
+			log.Info("Starting RFQ manager main event loop")
+			err = m.mainEventLoop()
+			if err != nil {
+				startErr = err
```
guggero, line 94: This `startErr` will never be evaluated as we're in a goroutine. I think we should instead just do `go m.mainEventLoop()` and do all the `defer` and error handling there.

rfqservice/rfq.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// Initialise the RFQ raw message stream handler.
+	m.Wg.Add(1)
+	go func() {
+		defer m.Wg.Done()
+		m.rfqStreamHandle, err = NewStreamHandler(
+			ctx, m.cfg.PeerMessagePorter,
+		)
+	}()
+	if err != nil {
```
guggero, line 126: This `err` will always be `nil` here, since the goroutine won't be started immediately.
I guess my question would be: Can these two goroutines here (and the one for `mainEventLoop`) not all be a single event loop?

If not, I would suggest to turn the stream handler and quote negotiator into a fully startable/stoppable independent system that does take care of its own goroutine handling. Otherwise things can get very hard to understand very quickly.

rfqservice/stream.go
```diff
@@ -0,0 +1,239 @@
...
+
+		IncomingQuoteRequests: incomingQuoteRequests,
+
+		ContextGuard: &fn.ContextGuard{
+			DefaultTimeout: DefaultTimeout,
+			Quit:           make(chan struct{}),
+		},
+	}
+
+	err = streamHandler.Start()
```
guggero, line 70: Related to my comment above, but this will basically never return until shut down. So I think we should instead turn this into its own goroutine and start/stop management or call it `Run()` instead to make it clear it's going to block.

guggero (commented):
Very nice progress on this!
Doing things fully concurrently is pretty hard. I think the way the three main components are currently set up is probably quite hard to understand and error prone. So I think we should either merge the three components together (e.g. a single main event loop goroutine if possible) or make them more independent (both sub components have a start/stop method and manage their own goroutine, so be fully independent components).

rfqmessages/messages.go
```diff
@@ -0,0 +1,39 @@
...
+// greater than this value.
+//
+// This offset was chosen as the concatenation of the alphabetical index
+// positions of the letters "t" (20), "a" (1), and "p" (16).
+const TapMessageTypeBaseOffset = 20116 + uint32(lnwire.CustomTypeStart)
+
+const (
+	// MsgTypeQuoteRequest is the message type identifier for a quote
+	// request message.
+	MsgTypeQuoteRequest = TapMessageTypeBaseOffset + 1
```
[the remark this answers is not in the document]
ffranr: These are not TLV type values. These are lndclient custom message type integers. Do the even/odd rules still apply there?

chain_bridge.go
```diff
@@ -200,6 +200,22 @@ func (l *LndRpcChainBridge) EstimateFee(ctx context.Context,
 	return l.lnd.WalletKit.EstimateFeeRate(ctx, int32(confTarget))
 }
 
+// SubscribeCustomMessages creates a subscription to custom messages received
+// from our peers.
+func (l *LndRpcChainBridge) SubscribeCustomMessages(
```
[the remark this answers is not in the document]
ffranr: I'll add a TODO.

rfqmessages/messages.go
```diff
@@ -0,0 +1,39 @@
...
+	// reject message.
+	MsgTypeQuoteReject = TapMessageTypeBaseOffset + 3
+)
+
+// OutgoingMessage is an interface that represents an outbound message that can
+// be sent to a peer.
+type OutgoingMessage interface {
+	// LndClientCustomMsg returns a custom message that can be sent to a
+	// peer using the lndclient.
+	LndClientCustomMsg() (lndclient.CustomMessage, error)
```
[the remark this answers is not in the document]
ffranr: Good idea, thanks!

rfqservice/rfq.go
ffranr: > turn the stream handler and quote negotiator into a fully startable/stoppable independent system

I'm going for this. Thanks for taking a look!

rfqmessages/messages.go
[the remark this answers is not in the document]
ffranr: I've thought again about this. I still agree that removing any mention of `lndclient` from the `rfqmessages` package makes sense (a good interface generalisation). Adding `WireMessage` and so on.

I'm not sure it simplifies anything to remove the peer field from these messages. It just means that we have to map the peer to the message some other way. Do you have a nice way in mind of doing that? Seems simpler to just including it in the message.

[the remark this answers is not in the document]
ffranr: In standup today @guggero  mentioned that he was ok with the peer field in the messages. And that I should remove lndclient from any interface in `rfqmessages`.

chain_bridge.go
```diff
@@ -200,6 +200,37 @@ func (l *LndRpcChainBridge) EstimateFee(ctx context.Context,
 	return l.lnd.WalletKit.EstimateFeeRate(ctx, int32(confTarget))
 }
 
+// SubscribeCustomMessages creates a subscription to custom messages received
+// from our peers.
+//
+// TODO(ffranr): This should be moved into its own interface because it's not
```
Roasbeef, line 206: +1 

Can be like `MessageTransport` or something like that, same with the below.

chain_bridge.go
```diff
@@ -200,6 +200,37 @@ func (l *LndRpcChainBridge) EstimateFee(ctx context.Context,
...
+func (l *LndRpcChainBridge) SendCustomMessage(ctx context.Context,
+	msg lndclient.CustomMessage) error {
+
+	return l.lnd.Client.SendCustomMessage(ctx, msg)
+}
+
+// InterceptHtlcs intercepts all incoming HTLCs and calls the given handler
+// function with the HTLC details. The handler function can then decide whether
+// to accept or reject the HTLC.
+func (l *LndRpcChainBridge) InterceptHtlcs(
```
Roasbeef, line 228: This could be move to a like `HtlcSwapControl` interface or something like that.

config.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/address"
 	"github.com/lightninglabs/taproot-assets/monitoring"
 	"github.com/lightninglabs/taproot-assets/proof"
+	"github.com/lightninglabs/taproot-assets/rfqservice"
```
Roasbeef, line 13: > Drop the service, just `rfq`, it's cleaner

rfqmsg/accept.go
```diff
@@ -0,0 +1,225 @@
...
+
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+const (
+	// Accept message type field TLV types.
+
+	TypeAcceptID          tlv.Type = 0
```
Roasbeef, line 17: This being the `32` byte rfq ID?

rfqmsg/accept.go
```diff
@@ -0,0 +1,225 @@
...
+	TypeAcceptAskingPrice tlv.Type = 2
+	TypeAcceptExpiry      tlv.Type = 4
+	TypeAcceptSignature   tlv.Type = 6
```
Roasbeef, line 20: As mentioned, the sig may not be 100% necessary, but we can keep it on for now until we get further into pipeline.

rfqmsg/accept.go
```diff
@@ -0,0 +1,225 @@
...
+// accept message.
+type acceptMsgData struct {
+	// ID is the unique identifier of the request for quote (RFQ).
+	ID ID
+
+	// AskingPrice is the asking price of the quote.
+	AskingPrice lnwire.MilliSatoshi
+
+	// Expiry is the asking price expiry lifetime unix timestamp.
+	Expiry uint64
```
Roasbeef, line 81: If you use the new `tlv` stuff, then this can just be: 
```go
Expiry tlv.RecordT[tlv.Type4, uint64]
```

and so forth

Then you don't need to make thing like `TypeRecordAcceptID`, as those types _are_ records themselves (or rather can become one).

Roasbeef: Goes for all the new TLV related structs here. 

I think this also needs an LN message framing layer as well. So the wire messages that can be decoded by something using the BOLT protocol message headers.

rfqmsg/accept.go
```diff
@@ -0,0 +1,225 @@
...
+
+// ShortChannelId returns the short channel ID of the quote accept.
+func (q *Accept) ShortChannelId() SerialisedScid {
+	// Given valid RFQ message ID, we then define a RFQ short chain ID
+	// (SCID) by taking the last 8 bytes of the RFQ message ID and
+	// interpreting them as a 64-bit integer.
+	scidBytes := q.ID[24:]
+
+	// TODO(ffranr): Am I doing this right?
+	scidInteger := binary.BigEndian.Uint64(scidBytes)
```
Roasbeef, line 195: Yep, can write a unit test to ensure the round trip between the two works properly.

rfqmsg/accept.go
```diff
@@ -0,0 +1,225 @@
...
+		acceptMsgData: msgData,
+	}, nil
+}
+
+// ShortChannelId returns the short channel ID of the quote accept.
+func (q *Accept) ShortChannelId() SerialisedScid {
+	// Given valid RFQ message ID, we then define a RFQ short chain ID
+	// (SCID) by taking the last 8 bytes of the RFQ message ID and
+	// interpreting them as a 64-bit integer.
+	scidBytes := q.ID[24:]
```
Roasbeef, line 192: I think could lift into a method in the `ID` to do the slicing to encapsulate a bit more.

Alternatively, the struct could store the entire ID slice, then another attr that sub-slices into the main one. So pointing to the same backing slice.

rfqmsg/accept.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// TODO(ffranr): Am I doing this right?
+	scidInteger := binary.BigEndian.Uint64(scidBytes)
+
+	return SerialisedScid(scidInteger)
+}
+
+// ToWire returns a wire message with a serialized data field.
+//
+// TODO(ffranr): This method should accept a signer so that we can generate a
```
Roasbeef, line 202: For inspiration, you could check out the way we did it in lnd for the gossip wire messages: https://github.com/lightningnetwork/lnd/blob/master/lnwire/channel_update.go#L224-L226

Then follow that to the various call sites.

rfqmsg/messages.go
```diff
@@ -0,0 +1,58 @@
...
+// SerialisedScid is a serialised short channel id (SCID).
+type SerialisedScid uint64
+
+// TapMessageTypeBaseOffset is the taproot-assets specific message type
+// identifier base offset. All tap messages will have a type identifier that is
+// greater than this value.
+//
+// This offset was chosen as the concatenation of the alphabetical index
+// positions of the letters "t" (20), "a" (1), and "p" (16).
+const TapMessageTypeBaseOffset = 20116 + uint32(lnwire.CustomTypeStart)
```
Roasbeef, line 20: 😎

rfqmsg/messages.go
```diff
@@ -0,0 +1,58 @@
...
+	// MsgType is the protocol message type number.
+	MsgType uint32
+
+	// Data is the data exchanged.
+	Data []byte
+}
+
+// IncomingMsg is an interface that represents an inbound wire message
+// that has been received from a peer.
+type IncomingMsg interface {
```
Roasbeef, line 50: Empty interface? Will read on to see the usage...


May want to make this "sealed" so having a private function s.t new instances can only be defined in tis package.

rfqmsg/messages.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// OutgoingMsg is an interface that represents an outbound wire message
+// that can be sent to a peer.
+type OutgoingMsg interface {
+	// ToWire returns a wire message with a serialized data field.
+	ToWire() (WireMessage, error)
```
Roasbeef, line 57: 🧐

rfqmsg/reject.go
```diff
@@ -0,0 +1,213 @@
...
+			return err
+		}
+
+		var errMsgBytes []byte
+		err = tlv.DVarBytes(r, &errMsgBytes, buf, l-4)
+		if err != nil {
+			return err
+		}
+
+		rejectErr := &RejectErr{
```
Roasbeef, line 70: Hmm, still need to finish getting through the diff, but rather than encode these messages as TLV blobs _within_ TLV blobs, they can just be normal wire messages. So you'd define the message as if it were in the `lnwire` pacakge.

rfqmsg/request.go
```diff
@@ -0,0 +1,281 @@
...
+func TypeRecordRequestAssetID(assetID **asset.ID) tlv.Record {
+	const recordSize = sha256.Size
+
+	return tlv.MakeStaticRecord(
+		TypeRequestAssetID, assetID, recordSize,
+		IDEncoder, IDDecoder,
+	)
+}
+
+func IDEncoder(w io.Writer, val any, buf *[8]byte) error {
```
Roasbeef, line 41: Won't need stuff like this with `tlv.RecordT`.

rfqservice/negotiator.go
```diff
@@ -0,0 +1,279 @@
...
+type Negotiator struct {
+	startOnce sync.Once
+	stopOnce  sync.Once
+
+	// cfg holds the configuration parameters for the negotiator.
+	cfg NegotiatorCfg
+
+	// outgoingMessages is a channel which is populated with outgoing peer
+	// messages.
+	outgoingMessages chan<- rfqmsg.OutgoingMsg
```
Roasbeef, line 32: So messages to send _to_ the peer?

rfqservice/negotiator.go
```diff
@@ -0,0 +1,279 @@
...
+
+	// assetSellOffers is a map (keyed on asset) that holds the asset sell
+	// offers that the negotiator has received.
+	assetSellOffers map[string]AssetSellOffer
+
+	assetGroupSellOffers map[string]AssetSellOffer
```
Roasbeef, line 38: Any particular reason for a `string` over the asset ID byte array?

rfqservice/negotiator.go
```diff
@@ -0,0 +1,279 @@
...
+// queryPriceOracle queries the price oracle for the asking price.
+func (n *Negotiator) queryPriceOracle(
+	req rfqmsg.Request) (*OracleAskingPriceResponse, error) {
+
+	ctx, cancel := n.WithCtxQuitNoTimeout()
+	defer cancel()
+
+	res, err := n.cfg.PriceOracle.QueryAskingPrice(
+		ctx, req.AssetID, req.AssetGroupKey, req.AssetAmount,
+		&req.BidPrice,
```
Roasbeef, line 68: Why's a bit price needed? In both directions, the receiver/sender is the one that's asking the edge node for a price quote. Or is this the hint price that it sends along?

rfqservice/negotiator.go
```diff
@@ -0,0 +1,279 @@
...
+
+	// If the asking price is nil, then we will return the error message
+	// supplied by the price oracle.
+	if oracleResponse.AskingPrice == nil {
+		rejectMsg := rfqmsg.NewRejectMsg(
+			request.Peer, request.ID, *oracleResponse.Err,
+		)
+		var msg rfqmsg.OutgoingMsg = &rejectMsg
+
+		sendSuccess := fn.SendOrQuit(n.outgoingMessages, msg, n.Quit)
```
Roasbeef, line 89: Ah I see now. Alternatively, we can have the outgoing messages be some generic interface that we can use to send messages. then w/e concrete implementation can use an underlying channel for synchronization purposes.

rfqservice/negotiator.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	// TODO(ffranr): Ensure that the expiry time is valid and sufficient.
+
+	// If the asking price is not nil, then we can proceed to respond with
+	// an accept message.
+	acceptMsg := rfqmsg.NewAcceptFromRequest(
+		request, *oracleResponse.AskingPrice, oracleResponse.Expiry,
+	)
+	var msg rfqmsg.OutgoingMsg = &acceptMsg
```
Roasbeef, line 103: No need for `var` usage here?

rfqservice/negotiator.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		return fmt.Errorf("invalid asset sell offer: %w", err)
+	}
+
+	// Store the offer in the appropriate map.
+	//
+	// If the asset group key is not nil, then we will use it as the key for
+	// the offer. Otherwise, we will use the asset ID as the key.
+	switch {
+	case offer.AssetGroupKey != nil:
```
Roasbeef, line 200: If wanted to flatten/unify, could use an `fn.Either` here for a single map.

rfqservice/negotiator.go
```diff
@@ -0,0 +1,279 @@
...
+	// Remove the offer from the appropriate map.
+	//
+	// If the asset group key is not nil, then we will use it as the key for
+	// the offer. Otherwise, we will use the asset ID as the key.
+	switch {
+	case assetGroupKey != nil:
+		compressedKey := assetGroupKey.SerializeCompressed()
+		keyStr := hex.EncodeToString(compressedKey)
+
+		delete(n.assetGroupSellOffers, keyStr)
```
Roasbeef, line 227: What about auto expiry?

rfqservice/negotiator.go
```diff
@@ -0,0 +1,279 @@
...
+	return nil
+}
+
+// mainEventLoop executes the main event handling loop.
+func (n *Negotiator) mainEventLoop() {
+	log.Debug("Starting negotiator event loop")
+
+	for {
+		select {
+		case <-n.Quit:
```
Roasbeef, line 246: No actual event loop?

rfqservice/oracle.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Err is the error returned by the price oracle service.
+	Err *rfqmsg.RejectErr
+}
+
+// PriceOracle is an interface that provides exchange rate information for
+// assets.
+type PriceOracle interface {
+	// QueryAskingPrice returns the asking price for the given asset amount.
+	QueryAskingPrice(ctx context.Context, assetId *asset.ID,
```
Roasbeef, line 34: Can combine asset+group key into single identifier, see above.

rfqservice/order.go
```diff
@@ -0,0 +1,313 @@
...
+	// Scid is the serialised short channel ID (SCID) of the channel to
+	// which the remit applies.
+	Scid SerialisedScid
+
+	// AssetAmount is the amount of the tap asset that is being requested.
+	AssetAmount uint64
+
+	// MinimumChannelPayment is the minimum number of millisatoshis that
+	// must be sent in the HTLC.
+	MinimumChannelPayment lnwire.MilliSatoshi
```
Roasbeef, line 30: +1 as overpay would be accepted, and maybe necessary to handle certain types of price slippage

rfqservice/order.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// Expiry is the asking price expiry lifetime unix timestamp.
+	Expiry uint64
+}
+
+// NewChannelRemit creates a new channel remit.
+func NewChannelRemit(quoteAccept rfqmsg.Accept) (*ChannelRemit, error) {
```
Roasbeef, line 37: No need for an error?

rfqservice/order.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// TODO(ffranr): I think we need a way to silently dismiss HTLCs that
+	//  are not associated with tap.
+
+	scid := SerialisedScid(htlc.OutgoingChannelID.ToUint64())
+	channelRemit, ok := h.FetchChannelRemit(scid)
+
+	// If a channel remit does not exist for the channel SCID, we reject the
+	// HTLC.
+	if !ok {
```
Roasbeef, line 141: Rather than fail here, we just want to allow the HTLC to pass through. As there might be _normal_ HTLCs coming across, either to forward elsewhere, or if they're receiving an actual BTC payment.

rfqservice/order.go
```diff
@@ -0,0 +1,313 @@
...
+		return &lndclient.InterceptedHtlcResponse{
+			Action: lndclient.InterceptorActionFail,
+		}, nil
+	}
+
+	// At this point, we know that the channel remit exists and has not
+	// expired whilst sitting in the local cache. We can now check that the
+	// HTLC complies with the channel remit.
+	err := channelRemit.CheckHtlcCompliance(htlc)
```
Roasbeef, line 150: Also note that we'll need to distinguish between crossing _into_ BTC and _out_ of it. When we cross in, we're receiving an asset and sending out BTC, when we cross out the opposite. 

The current structs and flows seem to be concerned mostly with the out (recv flow for leaf node), but we also need to be concerned with the opposite. I think _this_ is where the distinction between an ask and a bid exists:
 * When we receive in an asset to send out BTC, we're effectively buying the asset and selling BTC. 
 * When we receive in BTC to send out an asset, we're effectively selling the asset and buying BTC.

rfqservice/stream.go
```diff
@@ -0,0 +1,284 @@
...
+	wireMsg rfqmsg.WireMessage) error {
+
+	quoteAccept, err := rfqmsg.NewAcceptFromWireMsg(wireMsg)
+	if err != nil {
+		return fmt.Errorf("unable to create a quote accept message "+
+			"from a wire message: %w", err)
+	}
+	quoteAccept = quoteAccept
+
+	// TODO(ffranr): At this point, we need to validate that the quote
```
Roasbeef, line 113: 👍

rfqservice/stream.go
```diff
@@ -0,0 +1,284 @@
...
+	wireMsg rfqmsg.WireMessage) error {
+
+	quoteReject, err := rfqmsg.NewQuoteRejectFromWireMsg(wireMsg)
+	if err != nil {
+		return fmt.Errorf("unable to create a quote reject message "+
+			"from a wire message: %w", err)
+	}
+	quoteReject = quoteReject
+
+	// TODO(ffranr): At this point, we need to validate that the quote
```
Roasbeef, line 132: Rather than handle it here, if this is meant to be a dumb message passing sub-system, it can delegate the actual handling of this message into a higher layer.

tapdb/sqlc/migrations/000015_rfq.up.sql
```diff
@@ -0,0 +1,9 @@
+-- asset_sell_offers is a table that stores the set of all asset sell offers
+-- that are applicable to the RFQ negotiation process.
```
Roasbeef, line 2: Repeating the framing presented above:
 * When we're the first hop in the transfer, we're buying the asset and selling BTC. 
 * When we're the last hop in the transfer, we're selling the asset and buying BTC.

Roasbeef (commented):
Amazing work! Completed an initial drive by, dropped some relevant comments and fresh framing to help us synchronize our mental models.

ffranr:
The latest commits include an itest which:

* negotiates a quote via RFQ system between two peers
* buy side peer uses an accepted quote to send HTLC over custom scid
* sell side peer validates HTLC based on accepted quote in RFQ system

rfqservice/negotiator.go
ffranr: Yes. I've added that to the comment for greater clarity.

rfqservice/negotiator.go
ffranr: Latest commits use:
```
	assetSellOffers map[asset.ID]SellOffer

	// assetGroupSellOffers is a map (keyed on asset group key) that holds
	// asset sell offers.
	assetGroupSellOffers map[btcec.PublicKey]SellOffer
```

rfqservice/negotiator.go
ffranr: This is the request bid price. It's the hint price provided by the peer which is requesting a quote. I thought the price oracle might as well have access to it. But we might not want to allow that because the quote requesting peer might be colluding with the price oracle.

rfqservice/negotiator.go
ffranr: I'm not sure what we want here ultimately. I think that the latest commits have the simplest case for now.

I think we need to link this to the asset wallet somehow.

rfqservice/negotiator.go
ffranr: Removed.

rfqservice/oracle.go
ffranr: Will do :+1:

rfqservice/order.go
ffranr: We may need to cross reference with known channels (query LND) here.

Otherwise, for increased robustness, we may want to add a tap specific custom record to the HTLC to identify it as targetting tap. Then, other services could filter HTLCs with private SCIDs.

rfqservice/stream.go
ffranr: Yes, i see what you mean. Maybe with a peer session service or something like that.

rfq/stream.go
```diff
@@ -0,0 +1,306 @@
...
+		err := h.handleIncomingQuoteReject(wireMsg)
+		if err != nil {
+			return fmt.Errorf("unable to handle incoming quote "+
+				"reject message: %w", err)
+		}
+
+	default:
+		// Silently disregard the message if we don't recognise the
+		// message type.
+		return nil
```
guggero, line 188: Should we at least log an info or debug line here? Or is this to avoid spamming if other custom wire messages are received? Maybe a debug line like "Received message %T which is unrelated to RFQ, ignoring" could still be helpful.

tapcfg/server.go
```diff
@@ -321,6 +324,19 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 
 	multiNotifier := proof.NewMultiArchiveNotifier(assetStore, multiverse)
 
+	// Construct the RFQ manager.
+	priceOracle := rfq.NewMockPriceOracle(3600)
```
guggero, line 328: Should we add a TODO here? Or somehow else make sure a mock price oracle doesn't run in production?

perms/perms.go
```diff
@@ -212,6 +212,22 @@ var (
 			Entity: "universe",
 			Action: "read",
 		}},
+		"/rfqrpc.Rfq/UpsertAssetBuyOrder": {{
```
guggero, line 215: Yeah, the "Upsert" definitely feels like the wrong word at this level...

rpcserver.go
```diff
@@ -4749,3 +4756,308 @@ func MarshalAssetFedSyncCfg(
...
+	}
+
+	// Unmarshal the peer if specified.
+	var peer *route.Vertex
+
+	if req.Peer != nil {
+		var peerVertexBytes [33]byte
+		copy(peerVertexBytes[:], req.Peer)
+
+		pv, err := route.NewVertexFromBytes(peerVertexBytes[:])
```
guggero, line 4844: `NewVertexFromBytes` checks the length, so can pass in `req.Peer` directly.

rpcserver.go
```diff
@@ -4749,3 +4756,308 @@ func MarshalAssetFedSyncCfg(
...
+
+	return &rfqrpc.UpsertAssetSellOfferResponse{}, nil
+}
+
+// marshalAcceptedQuotes marshals a map of accepted quotes into the RPC form.
+func marshalAcceptedQuotes(
+	acceptedQuotes map[rfq.SerialisedScid]rfqmsg.Accept) []*rfqrpc.AcceptedQuote {
+
+	// Marshal the accepted quotes into the RPC form.
+	rpcQuotes := make([]*rfqrpc.AcceptedQuote, 0)
```
guggero, line 4929: nit: pre-initialize with `len(acceptedQuotes)`.

taprpc/rfqrpc/rfq.proto
```diff
@@ -0,0 +1,141 @@
...
+service Rfq {
+    /* tapcli: `rfq buyorder`
+    UpsertAssetBuyOrder is used to upsert a buy order for a specific asset.
+    */
+    rpc UpsertAssetBuyOrder (UpsertAssetBuyOrderRequest)
+        returns (UpsertAssetBuyOrderResponse);
+
+    /* tapcli: `rfq selloffer`
+    UpsertAssetSellOrder is used to upsert a sell offer for a specific asset.
+     */
```
guggero, line 16: micro-nit: indentation of closing comment is off by one space.

taprpc/rfqrpc/rfq.proto
```diff
@@ -0,0 +1,141 @@
...
+    // The minimum amount of the asset to buy.
+    uint64 min_asset_amount = 2;
+
+    // The maximum amount BTC to spend (units: millisats).
+    uint64 max_bid = 3;
+
+    // The unix timestamp after which the order is no longer valid.
+    uint64 expiry = 4;
+
+    // peer is an optional field that can be used to specify the peer to
```
guggero, line 64: nit: "peer's public key" to make it more clear what the content of the field is (to not confuse it with a node alias or whatever).

taprpc/rfqrpc/rfq.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+    - selector: rfqrpc.Rfq.UpsertAssetSellOffer
+      post: "/v1/taproot-assets/rfq/selloffer/asset-id/{asset_specifier.asset_id_str}"
+      body: "*"
+      additional_bindings:
+        - post: "/v1/taproot-assets/rfq/selloffer/group-key/{asset_specifier.group_key_str}"
+          body: "*"
+
+    - selector: rfqrpc.Rfq.QueryRfqAcceptedQuotes
+      post: "/v1/taproot-assets/rfq/quotes/accepted"
```
guggero, line 21: I think this could just be a `get` instead?

itest/rfq_test.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// newLndNode creates a new lnd node with the given name and funds its wallet
+// with the specified outputs.
+func newLndNode(name string, outputFunds []btcutil.Amount,
+	ht *lntest.HarnessTest) *node.HarnessNode {
+
+	lndArgs := []string{
+		"--default-remote-max-htlcs=483",
+		"--dust-threshold=5000000",
```
guggero, line 220: Why is this necessary?

itest/rfq_test.go
```diff
@@ -0,0 +1,353 @@
...
+//  2. Bob's node accepts the quote.
+//  3. Carol uses the quote accept message to construct a lightning invoice which
+//     will pay for the quote accepted by Bob.
+//  4. Alice pays the invoice.
+//  5. Bob's node intercepts the lightning payment from Alice and validates it
+//     against the quote accepted between Bob and Carol.
+func testRfqHtlcIntercept(t *harnessTest) {
+	// Initialize a new test scenario.
+	ts := newRfqTestScenario(t)
+	defer ts.Cleanup()
```
guggero, line 37: nit: could do `t.t.Cleanup(ts.Cleanup)` instead. Or even do that directly in `newRfqTestScenario()`.

guggero (commented):
Beautiful code and very nice commit structure.
Truly amazing work on this, congrats :tada:  Reads very well!

rpcserver.go
```diff
@@ -4749,3 +4756,308 @@ func MarshalAssetFedSyncCfg(
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshalling asset specifier: "+
+			"%w", err)
+	}
+
+	// Unmarshal the peer if specified.
+	var peer *route.Vertex
+
+	if req.Peer != nil {
```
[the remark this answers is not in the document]
ffranr: I need to remember that.

rpcserver.go
ffranr: Thanks, I keep forgetting to do that.

taprpc/rfqrpc/rfq.proto
ffranr: And on the other comments also, thanks :+1:

itest/rfq_test.go
ffranr: Ah I was just matching the standby nodes to get this to work. I'll try to remove this now.

rpcserver.go
```diff
@@ -4749,3 +4756,308 @@ func MarshalAssetFedSyncCfg(
...
+
+	// Unmarshal the group key.
+	var (
+		groupKeyBytes []byte
+		groupKey      *btcec.PublicKey
+
+		err error
+	)
+	switch {
+	case len(req.GetGroupKey()) > 0:
```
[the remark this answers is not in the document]
ffranr: Good point. I need to use `fn.Either` here also.

rfq/manager.go
```diff
@@ -0,0 +1,583 @@
...
+	err := m.negotiator.RemoveAssetSellOffer(assetID, assetGroupKey)
+	if err != nil {
+		return fmt.Errorf("error removing asset sell offer: %w", err)
+	}
+
+	return nil
+}
+
+// BuyOrder is a struct that represents a buy order.
+type BuyOrder struct {
```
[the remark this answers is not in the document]
ffranr: In my mind I distinguish between the words "buy" and "bid" in the following way: I think of "buying" as the process of purchasing something. And a "bid" is an offer made as part of the buying process.

So what I'm trying to convey with the name `BuyOrder` is an order for tapd to go off and purchase an asset. The bid offer value made during that process might be determined later in the buying process (after querying a price oracle for example).

The `BuyOrder` structure can then contain fields like `MaxBid`. Which limits the bid that tapd might offer when buying.

taprpc/rfqrpc/rfq.proto
```diff
@@ -0,0 +1,141 @@
...
+
+    /*
+    SubscribeRfqEventNtfns is used to subscribe to RFQ events.
+     */
+    rpc SubscribeRfqEventNtfns (SubscribeRfqEventNtfnsRequest)
+        returns (stream RfqEvent);
+}
+
+message AssetSpecifier {
+    oneof id {
```
guggero, line 35: Just a thought I had while working on something else: Should we just re-use the universe ID here that looks exactly the same? Would semantically more or less make sense as well IMO. Then we could re-use all the parsing/marshaling functions.

rfq/negotiator.go
```diff
@@ -0,0 +1,437 @@
...
+			return fmt.Errorf("negotiator failed to send reject " +
+				"message")
+		}
+
+		return nil
+	}
+
+	// If there is no price oracle available, then we cannot proceed with
+	// the negotiation. We will reject the quote request with an error.
+	if n.cfg.PriceOracle == nil {
```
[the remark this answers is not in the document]
ffranr: Error now handled on negotiator construction :+1:

rfq/negotiator.go
```diff
@@ -0,0 +1,437 @@
...
+	// offers.
+	assetSellOffers map[asset.ID]SellOffer
+
+	// assetGroupSellOffers is a map (keyed on asset group key) that holds
+	// asset sell offers.
+	assetGroupSellOffers map[btcec.PublicKey]SellOffer
+
+	// ErrChan is a channel that is used to communicate goroutine errors
+	// to the parent service.
+	ErrChan chan error
```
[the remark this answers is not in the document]
ffranr: Good idea, thanks.

rfq/stream.go
ffranr: I was trying to avoid scaling the number of log messages with the number of peer messages received by the LND node. Perhaps ee should add support in LND to be able to subscribe to certain message types only.

I've added a trace level log message. Let me know if you'd prefer debug level.

tapcfg/server.go
ffranr: I've added a TODO.

taprpc/rfqrpc/rfq.proto
ffranr: I think that the universe ID structure should contain the asset specifier structure as a field. The universe ID contains a proof type field which is not relevant when specifying an asset. So I think we should have:

```
message UniverseId {
    AssetSpecifier asset_specifier = 1;

    ProofType proof_type = 2;
}
``` 

Would be good to be able to re-use parsing/marshaling functions :+1:

rfq/manager.go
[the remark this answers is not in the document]
guggero: Okay, sounds good to me. Thanks for explaining your reasoning.

rfq/manager.go
```diff
@@ -0,0 +1,583 @@
...
+		case <-m.Quit:
+			log.Debug("Manager main event loop has received the " +
+				"shutdown signal")
+			return
+		}
+	}
+}
+
+// UpsertAssetSellOffer upserts an asset sell offer with the RFQ manager.
+func (m *Manager) UpsertAssetSellOffer(offer SellOffer) error {
```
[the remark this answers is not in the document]
ffranr: I don't understand you're thinking on the verb "upsert". Do you think that we should only use it when naming SQL statements?

> since we don't know on what parameters of the offer the "collision" (replacement) is happening

On the asset. If a sell offer exists for an asset then its updated. I'll add that in the doc.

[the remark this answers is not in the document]
guggero: Yes, I mean the verb "upsert". It's not an actual English word. And anyone unfamiliar with relational databases perhaps won't know the meaning. So I think it's fine to use it within the `tapd` package. But everywhere else I wouldn't use it, especially not in user-facing RPC names.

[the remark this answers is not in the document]
ffranr: If you think the programmers who will use the RPC endpoints won't understand/be comfortable with "upsert" when trying to create or update a sell offer/buy order then we should rename.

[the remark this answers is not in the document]
ffranr: In the latest commit I've changed "upsert" to "add" in the RPC endpoints. And I've described the update behaviour in the comment.

rfqmsg/accept.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+}
+
+// Decode decodes the structure from a TLV stream.
+func (q *acceptMsgData) Decode(r io.Reader) error {
+	stream, err := tlv.NewStream(q.decodeRecords()...)
+	if err != nil {
+		return err
+	}
+	return stream.Decode(r)
```
jharveyb, line 142: I think we'd want `stream.DecodeP2P()` for everything in `rfq`.

rfqmsg/accept_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/stretchr/testify/require"
+)
+
+// TestAcceptShortChannelId tests the ShortChannelId method of a quote accept
+// message.
+func TestAcceptShortChannelId(t *testing.T) {
+	t.Parallel()
+
+	scidInt := uint64(123456)
```
jharveyb, line 17: nit: random instead of static?

rfq/order.go
```diff
@@ -0,0 +1,320 @@
...
+	"github.com/lightninglabs/taproot-assets/rfqmsg"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// SerialisedScid is a serialised short channel id (SCID).
+type SerialisedScid uint64
+
+// ChannelRemit is a struct that holds the terms which determine whether a
+// channel HTLC is accepted or rejected.
+type ChannelRemit struct {
```
jharveyb, line 20: TIL remit 😅

rfq/order.go
```diff
@@ -0,0 +1,320 @@
...
+type OrderHandler struct {
+	startOnce sync.Once
+	stopOnce  sync.Once
+
+	// cfg holds the configuration parameters for the RFQ order handler.
+	cfg OrderHandlerCfg
+
+	// channelRemits is a map of serialised short channel IDs (SCIDs) to
+	// associated active channel remits.
+	channelRemits map[SerialisedScid]*ChannelRemit
```
jharveyb, line 105: nit: could probably use `lnutils.SyncMap` in a few of the new subsystems instead of a manual map + mtx.

rfq/order.go
```diff
@@ -0,0 +1,320 @@
...
+
+	h.channelRemits[channelRemit.Scid] = channelRemit
+
+	return nil
+}
+
+// FetchChannelRemit fetches a channel remit given a serialised SCID. If a
+// channel remit is not found, false is returned. Expired channel remits are
+// not returned and are removed from the cache.
+func (h *OrderHandler) FetchChannelRemit(scid SerialisedScid) (*ChannelRemit,
```
jharveyb, line 262: nit: here (and some other places) we don't distinguish between expired vs. non-existent - I'm not sure why we  want to erase that info? At a levels at least vs. just external RPCs.

rfq/oracle.go
```diff
@@ -0,0 +1,156 @@
...
+		suggestedBidPrice *lnwire.MilliSatoshi) (*OracleAskResponse,
+		error)
+
+	// QueryBidPrice returns a bid price for the given asset amount.
+	QueryBidPrice(ctx context.Context, assetId *asset.ID,
+		assetGroupKey *btcec.PublicKey,
+		assetAmount uint64) (*OracleBidResponse, error)
+}
+
+//// RpcPriceOracle is a price oracle that uses an external RPC server to get
```
jharveyb, line 63: To be resurrected in a follow-up PR?

chain_bridge.go
```diff
@@ -203,3 +204,61 @@ func (l *LndRpcChainBridge) EstimateFee(ctx context.Context,
...
+func NewLndRouterClient(lnd *lndclient.LndServices) *LndRouterClient {
+	return &LndRouterClient{
+		lnd: lnd,
+	}
+}
+
+// InterceptHtlcs intercepts all incoming HTLCs and calls the given handler
+// function with the HTLC details. The handler function can then decide whether
+// to accept or reject the HTLC.
+func (l *LndRouterClient) InterceptHtlcs(
```
jharveyb, line 257: 🔥

rfq/stream.go
```diff
@@ -0,0 +1,306 @@
...
+
+	return nil
+}
+
+// handleIncomingWireMessage handles an incoming wire message.
+func (h *StreamHandler) handleIncomingWireMessage(
+	wireMsg rfqmsg.WireMessage) error {
+
+	switch wireMsg.MsgType {
+	case rfqmsg.MsgTypeRequest:
```
jharveyb, line 163: nit: There's a fair amount of duplicate code here for each message type. Could be simplified to something like:

```go
// parse incoming message
var msg rfqmsg.IncomingMsg

switch wireMsg.MsgType {
  case rfqmsg.MsgTypeRequest:
    quoteReject, err := rfqmsg.NewQuoteRejectFromWireMsg(wireMsg)
    // handle err
    // debug log message
    
    msg = rfqmsg.IncomingMsg(quoteReject)
    
    case rfqmsg.MsgTypeAccept:
    // ....
    
    // handle final case + default

}

sendSuccess := fn.SendOrQuit(h.cfg.IncomingMessages, msg, h.Quit)
if !sendSuccess {
	return fmt.Errorf("RFQ stream handler shutting down")
}
```

I think the only missing piece there is a `String()` method for `rfq.MsgType` to help with accurate log messages.

rfq/negotiator.go
```diff
@@ -0,0 +1,424 @@
...
+		ctx, assetId, assetGroupKey, assetAmount,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query price oracle for "+
+			"bid: %w", err)
+	}
+
+	// If the bid price is nil, then we will return the error message
+	// supplied by the price oracle.
+	if oracleResponse.BidPrice == nil {
```
jharveyb, line 94: This feels inverted; we're ignoring the `oracleResponse.Err`, unless the price is nil? Seems like we should always check that response err, and maybe modify our behavior if the price also happened to be unset.

rfq/negotiator.go
```diff
@@ -0,0 +1,424 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query price oracle for ask "+
+			"price: %w", err)
+	}
+
+	// If the asking price is nil, then we will return a quote reject
+	// message which contains the error message supplied by the price
+	// oracle.
+	if oracleResponse.AskPrice == nil {
```
jharveyb, line 173: Same note: re inverted error checking order.

rfq/negotiator.go
```diff
@@ -0,0 +1,424 @@
...
+	default:
+		return fmt.Errorf("asset ID and asset group key are both nil")
+	}
+
+	return nil
+}
+
+// HasAssetSellOffer returns true if the negotiator has an asset sell offer
+// which matches the given asset ID/group and asset amount.
+func (n *Negotiator) HasAssetSellOffer(assetID *asset.ID,
```
jharveyb, line 359: nit: similar note from elsewhere re: distinguishing between non-existent sell offer vs. undersized sell offer.

rfq/manager.go
```diff
@@ -0,0 +1,594 @@
...
+	}
+
+	return nil
+}
+
+// handleIncomingMessage handles an incoming message. These are messages that
+// have been received from a peer.
+func (m *Manager) handleIncomingMessage(incomingMsg rfqmsg.IncomingMsg) error {
+	// Perform type specific handling of the incoming message.
+	switch msg := incomingMsg.(type) {
```
jharveyb, line 265: nit: missing a default case

rfq/manager.go
```diff
@@ -0,0 +1,594 @@
...
+	}
+
+	return nil
+}
+
+// handleOutgoingMessage handles an outgoing message. Outgoing messages are
+// messages that will be sent to a peer.
+func (m *Manager) handleOutgoingMessage(outgoingMsg rfqmsg.OutgoingMsg) error {
+	// Perform type specific handling of the outgoing message.
+	switch msg := outgoingMsg.(type) {
```
jharveyb, line 304: nit: should be an if, or will be expanded later?

rpcserver.go
```diff
@@ -4749,3 +4756,310 @@ func MarshalAssetFedSyncCfg(
...
+		assetID = &id
+
+	case len(req.GetAssetIdStr()) > 0:
+		assetIDBytes, err := hex.DecodeString(req.GetAssetIdStr())
+		if err != nil {
+			return nil, nil, fmt.Errorf("error decoding asset "+
+				"ID: %w", err)
+		}
+
+		copy(assetID[:], assetIDBytes)
```
jharveyb, line 4792: nit: nil deref warning, something like this should work:

```go
var id asset.ID
copy(id[:], assetIDBytes)
assetID = &id
```

rpcserver.go
```diff
@@ -4749,3 +4756,310 @@ func MarshalAssetFedSyncCfg(
...
+	return &rfqrpc.QueryRfqAcceptedQuotesResponse{
+		AcceptedQuotes: rpcQuotes,
+	}, nil
+}
+
+// marshallRfqEvent marshals an RFQ event into the RPC form.
+func marshallRfqEvent(eventInterface fn.Event) (*rfqrpc.RfqEvent, error) {
+	timestamp := eventInterface.Timestamp().UTC().Unix()
+
+	switch event := eventInterface.(type) {
```
jharveyb, line 4966: nit: missing default case.

rfqmsg/accept.go
guggero: Yes, good point (main difference is not allowing any messages longer than 65k bytes to prevent memory exhaustion attacks).

rfqmsg/request_test.go
```diff
@@ -0,0 +1,70 @@
...
+		testName string
+
+		id            ID
+		assetId       *asset.ID
+		assetGroupKey *btcec.PublicKey
+		assetAmount   uint64
+		bidPrice      lnwire.MilliSatoshi
+	}{
+		{
+			testName:      "asset_group_key_nil",
```
guggero, line 37: nit: I know this is wrong in some existing places, but basically the test names should just use spaces which are then turned into underlines by the test framework for the actual name (a bit easier to read for humans if it's spaces here IMO).

rfq/order.go
```diff
@@ -0,0 +1,320 @@
...
+	h.channelRemitsMtx.Lock()
+	defer h.channelRemitsMtx.Unlock()
+
+	remit, ok := h.channelRemits[scid]
+	if !ok {
+		return nil, false
+	}
+
+	// If the remit has expired, return false and clear it from the cache.
+	if time.Now().Unix() > int64(remit.Expiry) {
```
guggero, line 274: nit: maybe this would be slightly easier to read? `time.Unix(remit.Expiry, 0).Before(time.Now())`

rfq/order.go
```diff
@@ -0,0 +1,320 @@
...
+// cleanupStaleChannelRemits removes expired channel remits from the local
+// cache.
+func (h *OrderHandler) cleanupStaleChannelRemits() {
+	h.channelRemitsMtx.Lock()
+	defer h.channelRemitsMtx.Unlock()
+
+	// Iterate over the channel remits and remove any that have expired.
+	staleCounter := 0
+	for scid, remit := range h.channelRemits {
+		if time.Now().Unix() > int64(remit.Expiry) {
```
guggero, line 291: re: same here about using the `Before()` method.

server.go
```diff
@@ -158,6 +158,11 @@ func (s *Server) initialize(interceptorChain *rpcperms.InterceptorChain) error {
 			"federation: %v", err)
 	}
 
+	// Start the request for quote (RFQ) manager.
+	if err := s.cfg.RfqManager.Start(); err != nil {
+		return fmt.Errorf("unable to start RFQ manager: %v", err)
```
guggero, line 163: New linter rule complains about non-wrapped error.

guggero (approved):
LGTM pending nits and existing comments from @jharveyb.

rfqmsg/messages.go
```diff
@@ -0,0 +1,70 @@
...
+// SerialisedScid is a serialised short channel id (SCID).
+type SerialisedScid uint64
+
+// TapMessageTypeBaseOffset is the taproot-assets specific message type
+// identifier base offset. All tap messages will have a type identifier that is
+// greater than this value.
+//
+// This offset was chosen as the concatenation of the alphabetical index
+// positions of the letters "t" (20), "a" (1), and "p" (16).
+const TapMessageTypeBaseOffset = 20116 + uint32(lnwire.CustomTypeStart)
```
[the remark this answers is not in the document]
ffranr: The `lndclient.CustomMessage.MsgType` field has type `uint32`. So I made message types `uint32` to ease comparison. But I'll change the message type field to have type `lnwire.MessageType`, and I think `lndclient.CustomMessage.MsgType` is wrong and should be changed.

I'll check for overflow in the down conversion.

rfqmsg/request.go
```diff
@@ -0,0 +1,316 @@
...
+			AssetID:       assetID,
+			AssetGroupKey: assetGroupKey,
+			AssetAmount:   assetAmount,
+			BidPrice:      bidPrice,
+		},
+	}, nil
+}
+
+// NewRequestMsgFromWire instantiates a new instance from a wire message.
+func NewRequestMsgFromWire(wireMsg WireMessage) (*Request, error) {
```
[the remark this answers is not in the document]
ffranr: Done. And for accept and reject.

rfqmsg/accept.go
ffranr: Good to know, thanks.

rfq/oracle.go
ffranr: Yes, exactly.

rfq/stream.go
ffranr: I've simplified this code in the latest commits. I wasn't sure which way the `stream` service was going to go, so I made it verbose and repetitive. But now, I think that `stream` will only handle send/recv and message type casting.

rfq/manager.go
ffranr: Will almost certainly be extended later.

rpcserver.go
ffranr: Thanks for the code.

rfq/order.go
ffranr: Latest commits contain:
```
	expireTime := time.Unix(int64(remit.Expiry), 0).UTC()
	currentTime := time.Now().UTC()

	if currentTime.After(expireTime) {
```

Let me know if that doesn't work for you.

rfq/order.go
ffranr: See: https://github.com/lightninglabs/taproot-assets/pull/767#discussion_r1506734562

rfqservice/oracle.go
ffranr: I'll get that done in a future PR.

perms/perms.go
ffranr: Changed to `Add...`

rfqmsg/accept.go
ffranr: I'll handle this in a future PR if that's Ok with you. I'm not sure if it's a good idea to derive the scid from the ID.

rfq/order.go
ffranr: In the `OrderHandler` my intention was to only have active channel remits in state. But I do think its important to be able to respond with a "quote expired" message. I'll think a bit about how to implement that, it might be something we want to do once we have better persistence in place.

We need to think about handling many short lived quotes and therefore many expired quotes. For example, how long do we keep them around? I'd like to address in a future PR if that makes sense to you too.

rfq/negotiator.go
ffranr: Good point, thanks!

I'm going to leave a TODO in this PR to modify `HasAssetSellOffer` (+ rename) such that it returns an error which will allow the caller to respond with a more informative reject message. And then I can have an error type for "undersized sell offer".

rfq/negotiator.go
ffranr: Added price oracle error check regardless of `AskPrice` field value.

rfq/negotiator.go
ffranr: I'm going to refactor this function in the next PR. It will deal exclusively with price and wont return an RFQ message. This is because I'll use it when both buying and selling an asset.

I've added a TODO to remember to check the price oracle error correctly.

rfq/order.go
guggero: I don't think we need the explicit `.UTC()` call here, so this can be shortened into one line perhaps (or maybe not because of the `int64` cast). I think we only need to be careful and use `.UTC()` when we use the string representation of a date (which happens when storing a date in the DB for example).

guggero: But non-blocking for this PR, let's get it merged!

jharveyb (approved):
Solid changes given the feedback, and all points not changed bot tagged with TODOs for follow-up PRs in this series / line of work 💯

PR #768, opened 2024-01-16
tapgarden: tapscript tree support for minting anchor outputs

### LINKED ISSUE #343: tapgarden/taprpc: permit providing a tapscript sibling hash for minting calls
As is today, users can use the vPSBT calls to add arbitrary scripts to encumber tap assets. Right now this needs to be done mostly manually, as [we currently only have low-level calls](https://github.com/lightninglabs/taproot-assets/blob/main/itest/psbt_test.go#L26-L28). 

In order to allow minting directly into a multi-sig like script on both the Bitcoin and Tap layer, we should allow users to pass in this extra script information either during batch creation, or finalization. With this in place, users would be able to mint directly into an arbitrary on-chain script, as well as tapscript. After this, one missing piece would still be being able to pass in PSBTs into the normal send call, as those PSBT witnesses would be used to carry out the on chain Bitcoin spend. 

This issue is meant to track only the minting of assets directly into tapscripts. 

# Steps To Completion 

- [ ] Optionally allow a user to specify a tapscript sibling hash+pre-image [for each created asset](https://github.com/lightninglabs/taproot-assets/blob/main/taprpc/mintrpc/mint.proto#L33-L62). 
    * For the bitcoin anchor output, assuming we keep with the single batch anchor approach, we can accept tapscript root value in the `Finalize` call. 
    * In order to use this, users would need to create their entire script tree, or just that leaf, then make it into a tap branch with the pre-image also available. 

- [ ] During the minting process, rather than use a plain BIP 86 tweak key, instead use that script key as an internal key and commit to the tap script tweaked output key as the tweak key. 
     * This portion covers the script on the tap layer.

- [x] Using the parameters added above, rather than using the internal wallet key solely as the top level key with a blank tapscript sibling, the tapscript sibling specified by the user should be used in place. 

With the above, users can mint arbitrary asset to a top-level multi-sig anchor output, and also specify arbitrary scripts for each of the asset leaves created in that batch.

jharveyb:
Fixes #343.

Adding support for a batch key tapscript sibling first (BTC anchor-level), which would be supplied during the `Finalize` call for the batch. This should be a smaller change than per-asset support, while still using similar interfaces and DB changes.

TODOs:

- [x] unit tests for taptree en/decode
- [x] unit tests for tapscript sibling during Finalize call

I think given the related DB changes the per-asset sibling support should be in this PR (adding further TODOs) vs. a separate PR, but that may cause some commit reordering later on.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/768)


jharveyb:
itests are panicking as the DB-backed implementation of `TapscriptTreeStore` is not wired up yet.

dstadulis:
Next milestone for the PR:
Have the unit tests passing

We will try to split up the PR.  if the itest test be modified to be independent to the DB changes, PR would be possible to be split.

jharveyb:
Proof generation during minting now includes the tapscript sibling, and this includes the fixes for tapscript preimage verification.

The DB implementation will be a follow-on PR.

commitment/taproot.go
```diff
@@ -243,28 +293,17 @@ func TapBranchHash(preimage []byte) (*chainhash.Hash, error) {
...
+	// Decode the script version and the script itself.
+	version, script, err := ScriptFromLeafPreimage(preimage)
+	if err != nil {
+		return nil, err
 	}
 
-	// Enforce that it is not including another Taproot Asset commitment.
-	if bytes.Contains(preimage, TaprootAssetsMarker[:]) {
+	// Verify that the script is not including a Taproot Asset commitment.
+	if IsTaprootAssetCommitmentScript(script) {
```
Roasbeef, line 303: I think ok as is if we want to keep the exiting control flow, but in the future I'd like to re-work the control flow to assert all the preconditions up front, then have this just be a dumb method that computes the tap leaf hash.

commitment/taproot.go
```diff
@@ -150,11 +178,33 @@ func (t *TapscriptPreimage) TapHash() (*chainhash.Hash, error) {
...
-		return ErrPreimageIsTapCommitment
-	}
+	switch {
+	// A preimage smaller than a valid Taproot Asset commitment script needs
+	// no further inspection.
+	case len(t.SiblingPreimage) < TaprootAssetCommitmentScriptSize:
+		return nil
+
+	case len(t.SiblingPreimage) == TaprootAssetCommitmentScriptSize:
+		if IsTaprootAssetCommitmentScript(t.SiblingPreimage) {
```
Roasbeef, line 188: 👍

Roasbeef (changes requested):
Nice work! This helps us cover the ability to use custom tapscript trees at the top-level Bitcoin output. Spotted a few areas we can use some new types to avoid `nil` pointer checks when we want an arg to actually be optional w/o relying on any specific pointer semantics. 

I think to round out this PR, we'll want to make a concrete implementation for the new tree store in the DB. We can denormalize a bit by having a table that refs an internal key and then build out the tree storage in another table to be ref'd as a foreign key. This'll let us de-dup a bit, as we can have the tree storage primary key be the root hash. 

Then we can add an itest to ensure that we can spend from the output using normal keypend. From there we can opt to push out the final component of accepting a valid witness somehow (send call that uses a PSBT?) so we can make further progess on the group key component.

commitment/taproot.go
```diff
@@ -243,28 +293,17 @@ func TapBranchHash(preimage []byte) (*chainhash.Hash, error) {
 // TapLeafHash computes the TapHash of a TapLeaf node from its preimage
 // if possible, otherwise an error is returned.
 func TapLeafHash(preimage []byte) (*chainhash.Hash, error) {
-	// Empty preimage.
-	if len(preimage) == 0 {
-		return nil, ErrInvalidEmptyTapscriptPreimage
+	// Decode the script version and the script itself.
+	version, script, err := ScriptFromLeafPreimage(preimage)
```
JeremyRubin, line 297: Not quite correct here in how this is used. Please rename the function from "TapLeafHash" to something else, as I think it's confusing in how it's used, which seems to be closer to "ProofNotTapLeafHashCommitsToTapleafAssetCommitment" therefore an empty preimage is valid here

commitment/taproot.go
JeremyRubin: agree.. Dumb method "TapLeafHash" and smart method which checks whatever validity constraints on the leaf

commitment/taproot.go
```diff
@@ -150,11 +178,33 @@ func (t *TapscriptPreimage) TapHash() (*chainhash.Hash, error) {
...
+	case len(t.SiblingPreimage) == TaprootAssetCommitmentScriptSize:
+		if IsTaprootAssetCommitmentScript(t.SiblingPreimage) {
+			return ErrPreimageIsTapCommitment
+		}
 
-	return nil
+		return nil
+
+	default:
+		// The sibling may be an encoded tapleaf; remove the version
```
JeremyRubin, line 195: what is an "encoded tapleaf", i don't know that terminology

jharveyb:
Expanding this PR to include the DB implementation and a itest that will cover a key-spend & script spend of the minting output.

tapgarden/planter.go
```diff
@@ -127,6 +128,11 @@ type stateParamReq[T, S any] struct {
 	param S
 }
 
+type FinalizeParams struct {
+	FeeRate *chainfee.SatPerKWeight
```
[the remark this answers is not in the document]
jharveyb: The outer `Params` struct is optional, as well as each member. Agree `fn.Option` would make sense here.

tapgarden/planter.go
```diff
@@ -676,13 +684,40 @@ func (c *ChainPlanter) gardener() {
...
+
+	if params != nil {
+		tapTree = params.TapTree
+		feeRate = params.FeeRate
+	}
+
+	// If a tapscript tree was specified for this batch, we'll store it on
+	// disk. The caretaker we start for this batch will use it when deriving
+	// the final Taproot output key.
+	// TODO(jhb): Make this load call mandatory once we have a DB-backed
```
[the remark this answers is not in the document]
jharveyb: As in `c.cfg.TreeStore` would never be nil, which I think makes sense. This was just to have these changes used in unit tests without breaking the itests.

tapgarden/batch.go
```diff
@@ -104,7 +105,10 @@ func (m *MintingBatch) validateGroupAnchor(s *Seedling) error {
 
 // MintingOutputKey derives the output key that once mined, will commit to the
 // Taproot asset root, thereby creating the set of included assets.
-func (m *MintingBatch) MintingOutputKey() (*btcec.PublicKey, []byte, error) {
+func (m *MintingBatch) MintingOutputKey(
+	sibling *commitment.TapscriptPreimage) (*btcec.PublicKey, []byte,
```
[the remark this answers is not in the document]
jharveyb: As-is it would be persisted on `LoadTapscriptTree` right after we call `Finalize`. I agree we'd want to rework that if we want to be able to supply the sibling preimage earlier than the `Finalize` call (new RPC / CLI cmd?).

commitment/taproot.go
jharveyb: The byte slice `leafVersion || compactSizeof(script) || script`, which is the input for `h_tapleaf`.

commitment/taproot.go
jharveyb: Empty preimage would be rejected inside `ScriptFromLeafPreimage`?

Agree re: naming, could lift this validation to `TapscriptPreimage.TapHash()`.

commitment/taproot.go
JeremyRubin: This function needs to be made type safe, TapScriptPreimage should only ever be an encoded tapleaf or an unencoded one, and that should be handled before this is reached.

commitment/taproot.go
```diff
@@ -150,11 +178,33 @@ func (t *TapscriptPreimage) TapHash() (*chainhash.Hash, error) {
...
+		// prefix and inspect the script.
+		_, script, err := ScriptFromLeafPreimage(t.SiblingPreimage)
+		if err != nil {
+			return err
+		}
+
+		if IsTaprootAssetCommitmentScript(script) {
```
Roasbeef, line 202: Reminder that this func also needs to be updated. Then we should write a negative tests where we try to include a TAP v1 commitment as a sibling hash for the addr creation.

tapgarden/batch.go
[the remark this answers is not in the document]
jharveyb: Leaning that way to make it explicitly part of the batch struct vs. associated with the internal key.

asset/encoding.go
```diff
@@ -11,11 +11,26 @@ import (
...
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/lightninglabs/taproot-assets/mssmt"
 	"github.com/lightningnetwork/lnd/tlv"
 )
 
+// TLV types for tapscript tree encode/decode.
+const (
+	typeTapscriptType           tlv.Type = 1
```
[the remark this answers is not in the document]
jharveyb: It's `waddrmgr.TapscriptType`, or `waddrmgr.Tapscript.Type`, so the type of tapscript object (full tree, partial reveal, etc.)

Will remove in next update as we'll only support full trees.

commitment/taproot.go
jharveyb: The update would be to tie this version check to defined `asset.Version` s right? I don't recall any other changes needed.

asset/asset.go
```diff
@@ -462,6 +464,66 @@ const (
...
+	return fn.NewLeft[TapBranchPreimage, []txscript.TapLeaf](b)
+}
+
+func NewTreePreimageFromLeaves(l []txscript.TapLeaf) (*TapscriptTreePreimage,
+	error) {
+
+	if len(l) == 0 {
+		return nil, fmt.Errorf("no leaves given")
+	}
+	// TODO(jhb): assert leaf versions?
```
Roasbeef, line 484: Can assert that it's the default leaf version. Based on some recent investigation, we'll sorta  move away from having any real requirements re the leaf version at all.

asset/asset.go
```diff
@@ -462,6 +464,66 @@ const (
...
+
+func NewTapBranchPreimage(branch txscript.TapBranch) TapBranchPreimage {
+	return TapBranchPreimage{
+		left:  branch.Left().TapHash(),
+		right: branch.Right().TapHash(),
+	}
+}
+
+func (b TapBranchPreimage) TapBranchHash() chainhash.Hash {
+	l, r := b.left[:], b.right[:]
```
Roasbeef, line 508: I think can lift the `tapBranchHash` function from elsewhere.

asset/asset.go
```diff
@@ -462,6 +464,66 @@ const (
...
+	}
+
+	return *chainhash.TaggedHash(chainhash.TagTapBranch, l, r)
+}
+
+// TODO(jhb): redundant?
+// NewTapBranchHash takes the raw tap hashes of the left and right nodes and
+// hashes them into a branch.
+func NewTapBranchHash(l, r chainhash.Hash) chainhash.Hash {
+	if bytes.Compare(l[:], r[:]) > 0 {
```
Roasbeef, line 520: Can use above?

asset/encoding.go
```diff
@@ -11,11 +11,18 @@ import (
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/taproot-assets/mssmt"
 	"github.com/lightningnetwork/lnd/tlv"
 )
 
+// TLV types for TapLeaf encode/decode.
```
Roasbeef, line 20: This isn't an asset specific encoding right? Instead, we'll use it in the db, but stashing it here to avoid possible import cycles?

asset/encoding.go
```diff
@@ -664,3 +671,131 @@ func LeafDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
...
+
+	return &TapBranchPreimage{
+		left:  leftHash,
+		right: rightHash,
+	}, nil
+}
+
+// The following TapLeaf {en,de}coders are a duplicate of those used in
+// btcwallet. Specifically, the inner loop logic for handling []TapLeaf.
+// https://github.com/btcsuite/btcwallet/blob/master/waddrmgr/tlv.go#L160
```
Roasbeef, line 701: Comment slightly out of date now (we store a simplified version now).

asset/encoding.go
```diff
@@ -664,3 +671,131 @@ func LeafDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
...
+	}
+
+	leafVersion := uint8(leaf.LeafVersion)
+	tlvRecords := []tlv.Record{
+		tlv.MakePrimitiveRecord(
+			typeTapLeafVersion, &leafVersion,
+		),
+	}
+
+	if len(leaf.Script) > 0 {
```
Roasbeef, line 718: Why would the version be specified, but not also the script?

asset/encoding.go
```diff
@@ -664,3 +671,131 @@ func LeafDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
...
+
+	var leafTLVBytes bytes.Buffer
+	err = tlvStream.Encode(&leafTLVBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	// We encode the record with a varint length followed by
+	// the _raw_ TLV bytes.
+	var buf bytes.Buffer
```
Roasbeef, line 739: This is so it's a nested TLV? Or exterior length prefix for the DB? 

If it's a nested TLV, then the value already has a length prefix, so we can use the limit reader trick we use elsewhere.

asset/interface.go
```diff
@@ -24,3 +28,28 @@ type GenesisTxBuilder interface {
...
+// stored prior to storing the associated Tapscript tree. The associated
+// Tapscript tree cannot be updated once stored.
+type TapscriptTreeStore interface {
+	// StoreTapscriptTree persists a Tapscript tree associated with an
+	// internal key. Storing an empty Tapscript tree is a no-op, and storing
+	// a Tapscript tree for a key that already has a stored Tapscript tree
+	// returns a concrete error. The internal key for the control block of
+	// the Tapscript object must be set to the internal key.
+	StoreTapscriptTree(ctx context.Context,
+		tapTree TapscriptTreePreimage) (*chainhash.Hash, error)
```
Roasbeef, line 44: Return value being the root hash here?

asset/mock.go
```diff
@@ -205,6 +206,114 @@ func (m *MockGroupTxBuilder) BuildGenesisTx(newAsset *Asset) (*wire.MsgTx,
...
+		rootHash    chainhash.Hash
+		fullTreeErr error
+	)
+
+	tapTree.WhenLeft(func(b TapBranchPreimage) {
+		rootHash = b.TapBranchHash()
+		branchBytes := EncodeTapBranchPreimage(b)
+		m.Store[rootHash] = storedTapTree{true, [][]byte{branchBytes}}
+	})
+	tapTree.WhenRight(func(leaves []txscript.TapLeaf) {
```
Roasbeef, line 242: 👍

tapgarden/planter.go
```diff
@@ -676,13 +684,42 @@ func (c *ChainPlanter) gardener() {
...
+	// implementation of the tapscript tree store.
+	if params != nil {
+		params.FeeRate.WhenSome(func(fr chainfee.SatPerKWeight) {
+			feeRate = &fr
+		})
+		params.TapTree.WhenSome(func(p asset.TapscriptTreePreimage) {
+			if c.cfg.TreeStore != nil {
```
Roasbeef, line 707: Ah I see what you mean here re that extra nil check at times. In this case, I think it'll go away once we have a proper DB version.

Roasbeef: Params could similarly be `fn.Option`? Nesting would be the same.

commitment/taproot.go
```diff
@@ -73,6 +74,7 @@ func (t TapscriptPreimageType) String() string {
 // TapscriptPreimage wraps a pre-image byte slice with a type byte that self
 // identifies what type of pre-image it is.
 type TapscriptPreimage struct {
+	// TODO(jhb): SiblingPreimage need not be 32 bytes?
```
Roasbeef, line 77: So the sibling preimage here assumes a `tapHash` round has already been done when it's a leaf. So if it was actually a leaf, then it'll be 32 bytes. If two branches, then 64 bytes.

commitment/taproot.go
```diff
@@ -113,6 +115,49 @@ func NewPreimageFromBranch(branch txscript.TapBranch) *TapscriptPreimage {
 	}
 }
 
+// TapTreeToSibling constucts a taproot sibling hash from a Tapscript tree,
```
Roasbeef, line 146: `godoc` name mismatch.

commitment/taproot.go
Roasbeef: We could also try to unroll/enumerate all the possible versions here in diff types instead of the enum as well. I think we can do that in another pass though, as then we can tackle decomposing some of the optional/inner structs in the proof object.

commitment/taproot.go
```diff
@@ -113,6 +115,49 @@ func NewPreimageFromBranch(branch txscript.TapBranch) *TapscriptPreimage {
...
+			// another Taproot Asset commitment before use.
+			err = preimage.VerifyNoCommitment()
+			return
+		}
+
+		tree := txscript.AssembleTaprootScriptTree(tl...)
+		rootChildren := txscript.NewTapBranch(
+			tree.RootNode.Left(), tree.RootNode.Right(),
+		)
+		preimage = NewPreimageFromBranch(rootChildren)
```
Roasbeef, line 179: Hmm, yeah that comment above is funky. The original logic was written assuming that the `SiblingPreimage` bytes here is actually 32-bytes, as it's already been hashed into something that goes into the actual tapscript tree. 

To remedy this, I think we need to partition from `NewPreimageFromLeaf` slightly to make a version that hashes one level. We should also consult the existing itest here: https://github.com/lightninglabs/taproot-assets/blob/eb3ed937f3c92e0d6a7315d914274ad0d4186cef/tapscript/send_test.go#L2169

tapgarden/batch.go
```diff
@@ -125,8 +145,10 @@ func (m *MintingBatch) MintingOutputKey() (*btcec.PublicKey, []byte, error) {
 
 // genesisScript returns the script that should be placed in the minting output
 // within the genesis transaction.
-func (m *MintingBatch) genesisScript() ([]byte, error) {
-	mintingOutputKey, _, err := m.MintingOutputKey()
+func (m *MintingBatch) genesisScript(
+	sibling *commitment.TapscriptPreimage) ([]byte, error) {
```
Roasbeef, line 149: Similarly, we should already have this be bound to `MintingBatch` assuming we've read both from disk (or was created in memory).

Roasbeef: Ah, tho I guess as is, we'll read from the tree store, then supplement the tap sibling read as needed.

commitment/taproot.go
```diff
@@ -99,6 +99,34 @@ func NewPreimageFromLeaf(leaf txscript.TapLeaf) *TapscriptPreimage {
 	}
 }
 
+// ScriptFromLeafPreimage decodes the leafVersion and script from a taproot
+// leaf preimage.
+func ScriptFromLeafPreimage(preimage []byte) (*txscript.TapscriptLeafVersion,
+	[]byte, error) {
```
Roasbeef, line 105: I think we have an opportunity to use something other than `[]byte` here as the return type. It could be something like: `type Tapscript []byte`. Then we'd have `type TapscriptLeaf []byte` for the case that we expect the leaf version and length prefix to be there.

commitment/taproot.go
```diff
@@ -99,6 +99,34 @@ func NewPreimageFromLeaf(leaf txscript.TapLeaf) *TapscriptPreimage {
 	}
 }
 
+// ScriptFromLeafPreimage decodes the leafVersion and script from a taproot
+// leaf preimage.
+func ScriptFromLeafPreimage(preimage []byte) (*txscript.TapscriptLeafVersion,
```
Roasbeef, line 104: Re the comment below, we can have this accept `TapscriptLeaf`, then this becomes a transformation that extracts the versions, and returns that along with the encoded script.

commitment/taproot.go
```diff
@@ -150,11 +178,33 @@ func (t *TapscriptPreimage) TapHash() (*chainhash.Hash, error) {
...
 // commitment.
 func (t *TapscriptPreimage) VerifyNoCommitment() error {
-	if IsTaprootAssetCommitmentScript(t.SiblingPreimage) {
-		return ErrPreimageIsTapCommitment
-	}
+	switch {
+	// A preimage smaller than a valid Taproot Asset commitment script needs
+	// no further inspection.
+	case len(t.SiblingPreimage) < TaprootAssetCommitmentScriptSize:
+		return nil
```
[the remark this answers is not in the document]
Roasbeef: I like `!=` here as well. Then it's more obviously the dual of the `==` check below.

commitment/taproot.go
Roasbeef: This is the "v0 commitment secretly has a v1 commitment as a sibling" scenario. 

What I think we want to do is loosen the check here: https://github.com/lightninglabs/taproot-assets/blob/eb3ed937f3c92e0d6a7315d914274ad0d4186cef/commitment/tap.go#L316-L318

So it asserts length, then has the equality check left in.

commitment/taproot.go
Roasbeef: Made a suggestion above, then I think the name of this could be like:
```
func ExtractTapLeafscript(t TapscriptLeaf) (uint32, Tapscript, error) {
}
```

or w/e 

Or `NewTapscriptFromLeaf`. 

Then similar to this, we'd have a constructor to create `TapscriptLeaf` (maybe force a private var?) that asserts that the version and prefix are in place.

rpcserver.go
```diff
@@ -516,12 +516,22 @@ func (r *rpcServer) FinalizeBatch(_ context.Context,
...
 
+	feeRateOpt := fn.None[chainfee.SatPerKWeight]()
 	feeRate, err := checkFeeRateSanity(req.FeeRate)
 	if err != nil {
 		return nil, err
 	}
 
-	batch, err := r.cfg.AssetMinter.FinalizeBatch(feeRate)
+	if feeRate != nil {
+		feeRateOpt = fn.Some(*feeRate)
```
Roasbeef, line 526: Opt here could be option here, then it's just an assignment. Partially pref.

rpcserver.go
```diff
@@ -516,12 +516,22 @@ func (r *rpcServer) FinalizeBatch(_ context.Context,
...
+	}
+
+	// TODO(jhb): Add tapscript tree as an arg in FinalizeBatchRequest.
```
Roasbeef, line 529: Comment to mark TODO as blocking before merge.

tapgarden/planter_test.go
```diff
@@ -1238,6 +1371,11 @@ var testCases = []mintingStoreTestCase{
 		interval: minterInterval,
 		testFunc: testFinalizeBatch,
 	},
+	{
+		name:     "finalize_with_tapscript_tree",
```
Roasbeef, line 1375: FWIW if you leave out the `_` here, it'll insert them automatically as it replaces spaces.

Roasbeef (changes requested):

commitment/taproot.go
jharveyb: Talked to Oli, and the comment here is wrong - if the preimage is created from a leaf, it must be the full leaf. Rather, the preimage cannot be a taphash, because then we  wouldn't be able to verify that it's not a tap asset commitment.

This was the issue I hit before when modifying the proof verification code, where i'd already taphashed the preimage, but the verifier needs to do that hashing itself.

asset/interface.go
jharveyb: Yes exactly, which would then be added to the batch struct and stored in the DB as part of the batch itself.

asset/asset.go
jharveyb: I thought so too, but it was already reimplemented in the `commitment` package; the original in `txscript` is private.

asset/encoding.go
jharveyb: Exactly, this seemed like a type that would cause import issues.

asset/encoding.go
jharveyb: Not sure, but I think you could call `txscript.NewTapLead` with an empty script but not an empty version.

asset/encoding.go
jharveyb: Yeah, I wasn't certain about carrying this bit over from the original code - there the size prefix is used to make a `limitedReader` for each `TapLeaf`, where those readers are children of a reader made with a size hint about the full `[]TapLeaf` object. For a single leaf, I don't see the utility of the size prefix.

> so we can use the limit reader trick we use elsewhere

Do you have a ref for this? Or maybe it's not needed here at all (but we should probably enforce a max tapleaf script size before encode?)

tapgarden/planter.go
jharveyb: I think we'd have to switch all the public funcs. to be `fn.Option`. We could use an option as the return from `typedParam` and get rid of one level of nesting though.

jharveyb:
Latest push adds a DB implementation of `TapscriptTreeManager`, without unit tests. We also flesh out the `TapscriptTreeNodes` type, which represents the two formats we'll accept for tapscript trees on the RPC, and store tapscript trees on disk.

Remaining TODOs, in descending priority order:

- [ ] tapdb unit tests for implementation of `TapscriptTreeManager`
- [ ] DB changes to add tapscript sibling hash for minting batches + accompanying unit test coverage
- [ ] Itest with a non-empty minting batch tapscript tree + script path spend of minting UTXO
- [ ] Refactor `commitment.TapscriptPreimage` type to use safe constructors and getters, extend test coverage for tapscript sibling validation

asset/asset.go
```diff
@@ -462,6 +464,136 @@ const (
...
+type TapscriptTreeNodes struct {
+	// An ordered list of TapLeaf objects, representing a Tapscript tree.
+	leaves *TapLeafNodes
+
+	// The TapHashes of the child nodes of a TapBranch.
+	branch *TapBranchNodes
+}
+
+type TapLeafNodes struct {
+	v []txscript.TapLeaf
```
guggero, line 481: Could this instead be a direct type alias for a slice? E.g. `type TapLeafNodes []txscript.TapLeaf`?

asset/asset.go
```diff
@@ -462,6 +464,136 @@ const (
...
+
+func (t TapscriptTreeNodes) GetBranch() fn.Option[TapBranchNodes] {
+	if t.branch == nil {
+		return fn.None[TapBranchNodes]()
+	}
+
+	return fn.Some[TapBranchNodes](*t.branch)
+}
+
+func (t TapscriptTreeNodes) FromBranch(tbn TapBranchNodes) TapscriptTreeNodes {
```
guggero, line 508: Why is this a method on `TapscriptTreeNodes`? Same below. Those should probably just be standalone functions?

asset/asset.go
```diff
@@ -462,6 +464,136 @@ const (
...
+func TapTreeNodesFromBranch(branch txscript.TapBranch) TapscriptTreeNodes {
+	return TapscriptTreeNodes{
+		branch: &TapBranchNodes{
+			left:  branch.Left().TapHash(),
+			right: branch.Right().TapHash(),
+		},
+	}
+}
+
+func (l TapLeafNodes) ToLeaves() []txscript.TapLeaf {
```
guggero, line 571: IMO the methods of a type should be declared right after the type declaration.

asset/encoding.go
guggero: Yeah, I also don't think we need this double encoding.
The nice thing about the TLV stream is that it handles `EOF` correctly if it's at the end of a record. So we only need to use the limit reader trick if we are reading a nested record within a larger stream. If a stream occupies the full bytes we have available, we don't need to do this.

asset/mock.go
```diff
@@ -205,6 +206,98 @@ func (m *MockGroupTxBuilder) BuildGenesisTx(newAsset *Asset) (*wire.MsgTx,
...
+func (m *MockTapscriptTreeStore) StoreTapscriptTree(_ context.Context,
+	tapTree TapscriptTreeNodes) (*chainhash.Hash, error) {
+
+	if m.FailStore {
+		return nil, fmt.Errorf("failed to store tapscript tree")
+	}
+
+	var rootHash chainhash.Hash
+
+	tapLeaves := tapTree.GetLeaves().UnwrapToPtr()
```
guggero, line 234: Wouldn't this be a prime example for `WhenSome()`?
Although I have to say I haven't fully warmed up to that functional style of `Option` and `Either`, as it breaks one of the Golang patterns I've become very fond of: The happy path is non-indented in most cases...

But probably not something to discuss on this PR specifically.

tapdb/sqlc/migrations/000015_tapscript_trees.down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file is empty on purpose. There is nothing to roll back for this
+-- migration.
```
guggero, line 2: nit: if there is an inverse operation, it should be added here (which isn't always the case). But the inverse of a `CREATE TABLE` would be a `DROP TABLE`.

tapdb/sqlc/migrations/000015_tapscript_trees.up.sql
```diff
@@ -0,0 +1,28 @@
...
+
+        root_hash_id BIGINT NOT NULL REFERENCES tapscript_roots(root_id),
+
+        node_index BIGINT NOT NULL,
+
+        raw_node_id BIGINT NOT NULL REFERENCES tapscript_nodes(node_id),
+
+        -- A leaf can be repeated within a tree, and shared amongst trees, but
+        -- there can only be one leaf at a given index in a tree.
+        UNIQUE(root_hash_id, node_index, raw_node_id)
```
guggero, line 27: I know this inline style for unique indexes is very nice. But the downside is that we cannot change such a unique constraint. So if we ever needed to add or remove a column here, we'd need to create a whole new table, copy over the data and so on.
So perhaps for new code we want to create a named index instead? Then that can be dropped and re-created much more easily.

tapdb/sqlc/queries/assets.sql
```diff
@@ -776,6 +776,86 @@ SELECT key_family, key_index
...
+-- name: FetchTapscriptTreeRoots :many
+SELECT *
+FROM tapscript_roots;
+
+-- name: FetchTapscriptTree :many
+WITH tree_info AS (
+    -- We use this CTE to fetch all edges that link the given tapscript tree
+    -- root hash to child nodes. Each edge also contains the index of the child
+    -- node in the tapscript tree.
+    SELECT tapscript_roots.root_id, tapscript_roots.root_hash,
```
guggero, line 822: nit: since we only use the `branch_only`, `raw_node_id` and `node_index` in the actual outer select, do we need to select the other fields at all?

tapdb/sqlc/queries/assets.sql
```diff
@@ -776,6 +776,86 @@ SELECT key_family, key_index
...
+FROM tapscript_nodes
+JOIN tree_info
+    ON tree_info.raw_node_id = tapscript_nodes.node_id
+ORDER BY tree_info.node_index ASC;
+
+-- name: DeleteTapscriptTreeEdges :exec
+WITH tree_info AS (
+    -- We use this CTE to fetch all edges that link the given tapscript tree
+    -- root hash to child nodes.
+    SELECT tapscript_roots.root_id, tapscript_roots.root_hash,
```
guggero, line 842: Same here, would probably be enough to just select `edge_id`.

tapdb/sqlc/queries/assets.sql
```diff
@@ -776,6 +776,86 @@ SELECT key_family, key_index
...
+        tapscript_edges.edge_id
+    FROM tapscript_roots
+    JOIN tapscript_edges
+        ON tapscript_roots.root_id = tapscript_edges.root_hash_id
+    WHERE tapscript_roots.root_hash = @root_hash
+)
+-- TODO(jhb): This query does not delete any nodes. Separate query to delete
+-- nodes with no edges? (Which may be different from all child nodes of this
```
guggero, line 850: As discussed offline: Could use the `WHERE NOT EXISTS(...)` SQL feature for this.

tapdb/asset_minting.go
```diff
@@ -1206,6 +1208,114 @@ func (a *AssetMintingStore) FetchGroupByGroupKey(ctx context.Context,
...
+
+	var (
+		rootHash       chainhash.Hash
+		branchOnly     bool
+		treeNodesBytes [][]byte
+		err            error
+	)
+
+	tapLeaves := treeNodes.GetLeaves().UnwrapToPtr()
+	if tapLeaves != nil {
```
guggero, line 1222: nit: same here re `WhenSome()`.

tapdb/assets_common.go
```diff
@@ -596,3 +598,91 @@ func maybeUpsertAssetMeta(ctx context.Context, db UpsertAssetStore,
...
+	case nodeCount == 0:
+		return fmt.Errorf("no tapscript tree nodes provided")
+
+	case branchOnly && nodeCount != 2:
+		return fmt.Errorf("tapscript tree of branches must only have" +
+			"2 nodes")
+
+	// Our node index is an int64, so we can't store a tree with more than
+	// MaxInt64 nodes.
+	case nodeCount > math.MaxInt64:
```
guggero, line 640: I don't think we should allow anything even remotely close to this. Since we probably don't want the user to allow to store a whole BitVM commitment. So maybe something close to `math.MaxInt16` would be more reasonable? They could still just use the branches option if they really need more leaves.

tapgarden/interface.go
```diff
@@ -44,7 +44,7 @@ type Planter interface {
 
 	// FinalizeBatch signals that the asset minter should finalize
 	// the current batch, if one exists.
-	FinalizeBatch(feeRate *chainfee.SatPerKWeight) (*MintingBatch, error)
+	FinalizeBatch(params *FinalizeParams) (*MintingBatch, error)
```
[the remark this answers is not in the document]
guggero: Or just not a pointer at all? Since the content of the struct are options themselves.

commitment/tap.go
```diff
@@ -313,9 +313,6 @@ func IsTaprootAssetCommitmentScript(script []byte) bool {
 	if len(script) != TaprootAssetCommitmentScriptSize {
 		return false
 	}
-	if script[0] != byte(asset.V0) {
```
guggero, line 316: Should this just check for V0 _or_ V1 instead? Or why is this removed?

commitment/taproot.go
```diff
@@ -113,6 +115,56 @@ func NewPreimageFromBranch(branch txscript.TapBranch) *TapscriptPreimage {
...
+// TapTreeToSibling constucts a taproot sibling hash from Tapscript tree nodes,
+// to be used with a TapCommitment tree root to derive a tapscript root. This
+// could be multiple TapLeaf objects, or a representation of a TapBranch.
+func NewPreimageFromTapscriptTreeNodes(
+	tn asset.TapscriptTreeNodes) (*TapscriptPreimage, error) {
+
+	var preimage *TapscriptPreimage
+
+	maybeLeaves := tn.GetLeaves().UnwrapToPtr()
+	if maybeLeaves != nil {
```
guggero, line 155: nit: `WhenSome` instead?

tapgarden/caretaker.go
```diff
@@ -679,10 +679,32 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
...
+		var batchSibling *commitment.TapscriptPreimage
+		if b.cfg.TreeStore != nil && b.cfg.Batch.tapSiblingHash != nil {
+			tapSibling, err := b.cfg.TreeStore.LoadTapscriptTree(
+				ctx, *b.cfg.Batch.tapSiblingHash,
+			)
+			if err != nil {
+				return 0, err
+			}
+			if tapSibling == nil {
+				return 0, asset.TreeNotFound
```
guggero, line 695: Should the underlying implementation return `ErrNoRows` instead?

tapgarden/caretaker.go
```diff
@@ -776,11 +798,15 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
...
-		// disk, then import the public key into the wallet.
+		// disk, then import the public key into the wallet. The sibling
+		// here can always be nil as we'll fetch the output key computed
+		// in BatchStateFrozen.
 		//
 		// TODO(roasbeef): re-run during the broadcast phase to ensure
 		// it's fully imported?
-		mintingOutputKey, tapRoot, err := b.cfg.Batch.MintingOutputKey()
+		mintingOutputKey, tapRoot, err := b.cfg.Batch.MintingOutputKey(
+			nil,
```
guggero, line 808: If we make the parameter not a pointer, we can just pass in an empty params object with just `None()` for the two values.

guggero (commented):
Did a first pass, looks pretty good. Nice work! Will do a second round once all TODOs are addressed.

asset/asset.go
jharveyb: I think in that case we wouldn't be able to enforce using the constructors or decoders in `asset` as the only way to create an instance of `TapLeafNodes`.

asset/asset.go
jharveyb: Agree, leftover from previous approaches.

asset/mock.go
jharveyb: The problem I had with `WhenSome` was that it requires any logic using the value inside the `Option`, to be rewritten as a standalone function with side effects. Which is doable but reads awkwardly imo, for the changes needed in this PR at least.

tapdb/sqlc/migrations/000015_tapscript_trees.up.sql
jharveyb: SGTM, you mean like:

https://github.com/lightninglabs/taproot-assets/blob/a6415f8e5cfafb1d7e85f09428905b80607c9be7/tapdb/sqlc/migrations/000014_multiverse_tree.up.sql#L38

?

tapdb/assets_common.go
jharveyb: I'm fine with `MaxInt16`, and I imagine non-BitVM users aren't using such large trees.

tapgarden/caretaker.go
jharveyb: I thought that was only for queries expecting one result / tagged with `:one` ? I recall finding the docs for this before but haven't rediscovered them yet...

jharveyb:
Latest push includes an itest that demonstrates creating a minting batch, and finalizing it with a tapscript tree sibling. The genesis outpoint is then spent via the script path of that tapscript tree, validating that the Taproot tree in the genesis outpoint was constructed correctly.

There are 2+ new commits concerning adding a tapscript sibling field to batches in the DB, and fixes around storing all elements of the Taproot tree for genesis outputs.

Remaining tasks are mostly test coverage + updating to match review feedback + extending the delete query for `TapscriptTreeStore`.

rpcserver.go
jharveyb: TODO has been implemented in this PR now.

Roasbeef (changes requested):
> Reviewed 4 of 14 files at r1, 11 of 11 files at r2, 1 of 1 files at r3, 34 of 34 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/768)* status: all files reviewed, 83 unresolved discussions (waiting on @guggero, @JeremyRubin, and @jharveyb)

___
*[`rpcserver.go` line 519 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_OzQT02kRsf_30Tn2:-Nq_OzQT02kRsf_30Tn3:b-a16lpk) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/fb4df872329b223f7cc844355819fa8b92f54677/rpcserver.go#L519)):*
> ```Go
> 	error) {
> 
> 	feeRateOpt := fn.None[chainfee.SatPerKWeight]()
> ```

Looking forward to the next release when they'll actually make type inferencing work properly for type params...
___
*[`rpcserver.go` line 2967 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_V0gl-43C-eLvvcec:-Nq_V0gl-43C-eLvvced:b-pq49ez) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/rpcserver.go#L2967)):*
> ```Go
> 	// txscript.TapscriptLeafVersion is a uint8.
> 	rpcLeafVersion := leaf.LeafVersion
> 	if rpcLeafVersion > math.MaxUint8 {
> ```

Commented in the other revision, but I think we should just restrict the leaf version all together. 
___
*[`asset/asset.go` line 479 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_V9Co3NSZM77enzbV:-Nq_V9Co3NSZM77enzbW:b-19vluf) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/asset/asset.go#L479)):*
> ```Go
> 
> 	// The TapHashes of the child nodes of a TapBranch.
> 	branch *TapBranchNodes
> ```

Could `fn.Either` it, depends on what the actual ergonomics are though further down the line
___
*[`asset/asset.go` line 482 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_VF7UAPoUjxWArtz2:-Nq_VF7UAPoUjxWArtz3:btjdkjt) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/asset/asset.go#L482)):*
> ```Go
> }
> 
> type TapLeafNodes struct {
> ```

Could just be:
```go
type TapLeafNodes []txscript.TapLeaf
```
___
*[`asset/asset.go` line 582 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_VoZWCCjKY1bYs_wG:-Nq_VoZWCCjKY1bYs_wH:bk7vcfp) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/asset/asset.go#L582)):*
> ```Go
> }
> 
> func (b TapBranchNodes) ToBranch() [][]byte {
> ```

Maybe `type SerializedBranch []byte`, then `type SerializedBranches [2]SerializedBranch`?
___
*[`asset/encoding.go` line 677 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_W1-X21a7VFzPW7OI:-Nq_W1-X21a7VFzPW7OJ:b-fja6pm) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/asset/encoding.go#L677)):*
> ```Go
> 
> func EncodeTapBranchNodes(branch TapBranchNodes) [][]byte {
> 	var left, right bytes.Buffer
> ```

Buffer used to ensure a copy?
___
*[`asset/encoding.go` line 684 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_W67476QMkxUQTOmt:-Nq_W67476QMkxUQTOmu:b-ffzf32) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/asset/encoding.go#L684)):*
> ```Go
> }
> 
> func DecodeTapBranchNodes(branchData [][]byte) (*TapBranchNodes, error) {
> ```

Can use the `SerializedBranches` type mentioned above (could also do that basic private variable thing to force creation via constructor). 
___
*[`commitment/tap.go` line 316 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_PGY45_gQeM0GqWwI:-Nq_PGY45_gQeM0GqWwJ:b-ff6ilb) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a6415f8e5cfafb1d7e85f09428905b80607c9be7/commitment/tap.go#L316)):*
> ```Go
> 		return false
> 	}
> 	if script[0] != byte(asset.V0) {
> ```

:+1: 
___
*[`commitment/taproot.go` line 174 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_WZDa4decZeJQL8Wp:-Nq_WZDa4decZeJQL8Wq:bj6oo2o) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/commitment/taproot.go#L174)):*
> ```Go
> 		}
> 
> 		// Make a branch from the leaves.
> ```

I think we also need to check that none of the specified leaves are a commitment as well. Just for belt and suspenders coverage. 
___
*[`fn/option.go` line 53 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_PgbnAwxS9c8LqhQW:-Nq_PgbnAwxS9c8LqhQX:b-od02x5) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/fn/option.go#L53)):*
> ```Go
> // UnwrapPtr is used to extract a reference to a value from an option, and we
> // supply an empty pointer in the case when the Option is empty.
> func (o Option[A]) UnwrapToPtr() *A {
> ```

Alternatively, I think this could be written just using `WhenSome`? 

So:
```go
func (o Option[A]) UnwrapToPtr() *A {
    var v *A
    o.WhenSome(func(a A) {
           v = &a
    })

    return v
}
```
___
*[`fn/option.go` line 54 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_PZWN8CXqXyNRHJUa:-Nq_PZWN8CXqXyNRHJUb:bsr032t) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/fn/option.go#L54)):*
> ```Go
> // supply an empty pointer in the case when the Option is empty.
> func (o Option[A]) UnwrapToPtr() *A {
> 	UnwrapToNil := func() *A { return nil }
> ```

Variable name should start with a lower case char. 
___
*[`itest/assets_test.go` line 375 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_Q5jBCUdWF6kI0As-:-Nq_Q5jCDGUVc7G89kEt:bubw0p8) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/itest/assets_test.go#L375)):*
> ```Go
> 	)
> 	typedBranch := asset.TapTreeNodesFromBranch(siblingBranch)
> 	rawBranch := typedBranch.GetBranch().UnwrapToPtr().ToBranch()
> ```

Needs a nil pointer check to avoid deref?
___
*[`itest/assets_test.go` line 392 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_QGMV4zYqDaybQsCr:-Nq_QGMV4zYqDaybQsCs:blm0axs) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/itest/assets_test.go#L392)):*
> ```Go
> 
> 	AssertAssetBalances(t.t, t.tapd, rpcSimpleAssets, rpcIssuableAssets)
> 	// assert on taproot_asset_root, merkle_root, and internal_key
> ```

To be resolved before merging? 
___
*[`itest/assets_test.go` line 431 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_QWHV-QdbDLSGmVee:-Nq_QWHV-QdbDLSGmVef:b-69nf1l) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/itest/assets_test.go#L431)):*
> ```Go
> 	// Build the control block and witness.
> 	inclusionProof := bytes.Join(
> 		[][]byte{siblingScriptHash[:], mintTapTreeRoot}, []byte{},
> ```

So if you already have the tapscript tree above, then you can use the internal map (based on hash of the script) to retrieve the pre-computed control block 
___
*[`itest/assets_test.go` line 481 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_Qlwq6F2ZP5inIqn4:-Nq_Qlwq6F2ZP5inIqn5:b-8v0613) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/itest/assets_test.go#L481)):*
> ```Go
> 
> 	// Bob should detect the TX.
> 	t.lndHarness.Miner.AssertNumTxsInMempool(1)
> ```

I think once we get in the RPC calls to allow committing the vPSBT and handing off to the freighter, we'll want to call that in place here as then we can also ensure that the transfer shows up in the DB of the sender (also passive assets handled, etc). 
___
*[`tapdb/asset_minting.go` line 310 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_RI9k52GKrFvROH7p:-Nq_RI9k52GKrFvROH7q:b-qmb2z7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/asset_minting.go#L310)):*
> ```Go
> 
> 		newBatchSibling := newBatch.TapSibling()
> 		if newBatchSibling != nil {
> ```

Could be an option here to `WhenSome` (would lead to the same amount of nesting), but I understand in this scenario it's also a matter of style pref. 
___
*[`tapdb/asset_minting.go` line 1254 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_Rhvt6A5ck8aB65Nt:-Nq_Rhvt6A5ck8aB65Nu:b1bgse7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/asset_minting.go#L1254)):*
> ```Go
> }
> 
> func (a *AssetMintingStore) StoreTapscriptTree(ctx context.Context,
> ```

Missing `gogoc` comment. 
___
*[`tapdb/asset_minting.go` line 1264 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_Rno_ChOndbpbZ6Vl:-Nq_Rno_ChOndbpbZ6Vm:bkkjsua) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/asset_minting.go#L1264)):*
> ```Go
> 	)
> 
> 	tapLeaves := treeNodes.GetLeaves().UnwrapToPtr()
> ```

Alternatively, could use `MapOptionZ` here to implement the side obtain `treeNodeBytes`, then return an error. 
___
*[`tapdb/asset_minting.go` line 1273 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_RyAvCHsO0KbP4IgM:-Nq_RyAvCHsO0KbP4IgN:b-tllydv) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/asset_minting.go#L1273)):*
> ```Go
> 	}
> 
> 	tapBranch := treeNodes.GetBranch().UnwrapToPtr()
> ```

Same here re optionz 
___
*[`tapdb/asset_minting.go` line 1279 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_S0Ri3fwBQwdoCyhX:-Nq_S0Ri3fwBQwdoCyhY:b7ximrc) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/asset_minting.go#L1279)):*
> ```Go
> 		treeNodesBytes = asset.EncodeTapBranchNodes(*tapBranch)
> 	}
> ```

Should these be mutually exclusive? (branch vs leaves)
___
*[`tapdb/assets_common.go` line 648 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_SMRfB1KruDkSxAj5:-Nq_SMRfB1KruDkSxAj6:b-n64b6i) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/assets_common.go#L648)):*
> ```Go
> 
> 	for i := int64(0); i < int64(len(nodes)); i++ {
> 		nodeId, err := q.UpsertTapscriptTreeNode(ctx, nodes[i])
> ```

Node here being the leaf script, that may actually be re-used elsewhere so we can save on duplication? 
___
*[`tapdb/assets_common.go` line 653 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_STNS0ZlaF_oh64dv:-Nq_STNS0ZlaF_oh64dw:b-wrh5d3) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/assets_common.go#L653)):*
> ```Go
> 		}
> 
> 		edge := TapscriptTreeEdge{
> ```

Re above, could make the distinction re script vs node. Script is the free floating script, while the node binds it to a root ID and also enforces an index. 
___
*[`tapdb/sqlc/migrations/000016_tapscript_trees.up.sql` line 1 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_Sncb8sb5CJ4e4XJ2:-Nq_SnccCtSR1A4P8sxr:b6flmod) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/sqlc/migrations/000016_tapscript_trees.up.sql#L1)):*
> ```TSQL
> ALTER TABLE asset_minting_batches ADD COLUMN tapscript_sibling BLOB;
> ```

Then the addition within the script key table will come in a future PR? 

So associating a given output that we control with a tapscript tree that has been tweaked with the internal key to arrive at the public output. 
___
*[`tapdb/sqlc/migrations/000016_tapscript_trees.up.sql` line 28 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_T4EJA-sZxBuPJqAU:-Nq_T4EJA-sZxBuPJqAV:b-8fx2u7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/sqlc/migrations/000016_tapscript_trees.up.sql#L28)):*
> ```TSQL
> );
> 
> -- A leaf can be repeated within a tree, and shared amongst trees, but there can
> ```

:+1: 
___
*[`tapdb/sqlc/queries/assets.sql` line 855 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_Tcan52PSK6TktXX3:-Nq_Tcan52PSK6TktXX4:b6vboox) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/sqlc/queries/assets.sql#L855)):*
> ```TSQL
>     SELECT tapscript_roots.root_id, tapscript_roots.root_hash,
>         tapscript_roots.branch_only, tapscript_edges.node_index,
>         tapscript_edges.raw_node_id
> ```

I think we only need two of these attributes in the CTE as only the ID and `branch_only` fields are used. 
___
*[`tapdb/sqlc/queries/assets.sql` line 883 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_U65CCQUOBxf6ehVc:-Nq_U65CCQUOBxf6ehVd:b-a174z8) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/tapdb/sqlc/queries/assets.sql#L883)):*
> ```TSQL
> -- nodes with no edges? (Which may be different from all child nodes of this
> -- root, if a node is in multiple trees).
> DELETE FROM tapscript_edges
> ```

You can use `ON DELETE CASCADE` to achieve that
___
*[`taprpc/taprootassets.proto` line 761 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/768#-Nq_ULCK6iH8VgSt5iRc:-Nq_ULCK6iH8VgSt5iRd:b6vxr5v) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea6a5c84706b644f431694cd3ee9f0105f205ec9/taprpc/taprootassets.proto#L761)):*
> ```proto
> message TapLeaf {
>     // The leaf version. Should be 0xc0 (192) in case of a SegWit v1 script.
>     uint32 leaf_version = 1;
> ```

Perhaps we just don't expose it at all for now? Any other leaf version today is non-standard, and don't think we're suuuper close to repurposing one of them

jharveyb:
> > ```go
> > 	// txscript.TapscriptLeafVersion is a uint8.
> > 	rpcLeafVersion := leaf.LeafVersion
> > 	if rpcLeafVersion > math.MaxUint8 {
> > ```
> 
> Commented in the other revision, but I think we should just restrict the leaf version all together.

SGTM, we could open that up later as-needed.

> > ```go
> > }
> > 
> > type TapLeafNodes struct {
> > ```
> 
> Could just be:
> 
> ```go
> type TapLeafNodes []txscript.TapLeaf
> ```

This would allow creating a `TapLeafNodes` object without any validation from another package. IIUC we have to use a private struct member to prevent that.

> > ```go
> > }
> > 
> > func (b TapBranchNodes) ToBranch() [][]byte {
> > ```
> 
> Maybe `type SerializedBranch []byte`, then `type SerializedBranches [2]SerializedBranch`?

👍🏽 

> > ```go
> > func EncodeTapBranchNodes(branch TapBranchNodes) [][]byte {
> > 	var left, right bytes.Buffer
> > ```
> 
> Buffer used to ensure a copy?

Yes, though doing an explicit `bytes.Clone()` now.

> > ```go
> > }
> > 
> > func DecodeTapBranchNodes(branchData [][]byte) (*TapBranchNodes, error) {
> > ```
> 
> Can use the `SerializedBranches` type mentioned above (could also do that basic private variable thing to force creation via constructor).

This is effectively a second constructor, where the first is from `TapBranch` (where we don't need any of these sanity checks).

> > ```go
> > 		}
> > 
> > 		// Make a branch from the leaves.
> > ```
> 
> I think we also need to check that none of the specified leaves are a commitment as well. Just for belt and suspenders coverage.

Wouldn't that already be out-of-spec / rejected since it's not at the correct level in the tree? Happy to add though.

> > ```go
> > // UnwrapPtr is used to extract a reference to a value from an option, and we
> > // supply an empty pointer in the case when the Option is empty.
> > func (o Option[A]) UnwrapToPtr() *A {
> > ```
> 
> Alternatively, I think this could be written just using `WhenSome`?
> 
> So:
> 
> ```go
> func (o Option[A]) UnwrapToPtr() *A {
>     var v *A
>     o.WhenSome(func(a A) {
>            v = &a
>     })
> 
>     return v
> }
> ```

👍🏽 

> > ```go
> > 	AssertAssetBalances(t.t, t.tapd, rpcSimpleAssets, rpcIssuableAssets)
> > 	// assert on taproot_asset_root, merkle_root, and internal_key
> > ```
> 
> To be resolved before merging?

👍🏽 

> > ```go
> > 	// Bob should detect the TX.
> > 	t.lndHarness.Miner.AssertNumTxsInMempool(1)
> > ```
> 
> I think once we get in the RPC calls to allow committing the vPSBT and handing off to the freighter, we'll want to call that in place here as then we can also ensure that the transfer shows up in the DB of the sender (also passive assets handled, etc).

There is no asset transfer happening here (and neither tapd is aware of the TX being made).

> > ```go
> > 		treeNodesBytes = asset.EncodeTapBranchNodes(*tapBranch)
> > 	}
> > ```
> 
> Should these be mutually exclusive? (branch vs leaves)

Yes, the passed `treeNodes` should only have one of the two struct members populated. All the constructors for `TapscriptTreeNodes` only populate one member.

> > ```go
> > 	for i := int64(0); i < int64(len(nodes)); i++ {
> > 		nodeId, err := q.UpsertTapscriptTreeNode(ctx, nodes[i])
> > ```
> 
> Node here being the leaf script, that may actually be re-used elsewhere so we can save on duplication?

I named it node instead of leaf, as it can also be the child of a tapbranch / a taphash. And this table layout seemed better than separating taphashes into their own table. Nodes can indeed be reused in multiple trees.

> > ```tsql
> > ALTER TABLE asset_minting_batches ADD COLUMN tapscript_sibling BLOB;
> > ```
> 
> Then the addition within the script key table will come in a future PR?
> 
> So associating a given output that we control with a tapscript tree that has been tweaked with the internal key to arrive at the public output.

Yes, though the `script_keys` table already has a field for storing a tweak, so I don't think we'll need to modify that table. The `internal_keys` table has no such field so I needed to add it here.

> > ```tsql
> > -- nodes with no edges? (Which may be different from all child nodes of this
> > -- root, if a node is in multiple trees).
> > DELETE FROM tapscript_edges
> > ```
> 
> You can use `ON DELETE CASCADE` to achieve that

I looked into it a bit, and I think the many edges -> one node mapping makes that option (or RESTRICT) untenable? Or at least more complicated than what I have now (h/t Oli), which is: Delete the edges, and then use `WHEN NOT EXISTS` to delete any nodes not referenced by an edge.

> > ```proto
> > message TapLeaf {
> >     // The leaf version. Should be 0xc0 (192) in case of a SegWit v1 script.
> >     uint32 leaf_version = 1;
> > ```
> 
> Perhaps we just don't expose it at all for now? Any other leaf version today is non-standard, and don't think we're suuuper close to repurposing one of them

👍🏽

asset/mock.go
```diff
@@ -205,6 +206,76 @@ func (m *MockGroupTxBuilder) BuildGenesisTx(newAsset *Asset) (*wire.MsgTx,
 // GenesisTxBuilder interface.
 var _ GenesisTxBuilder = (*MockGroupTxBuilder)(nil)
 
+type MockTapscriptTreeStore struct {
+	Store               map[SerializedKey][]byte
+	FailLoad, FailStore bool
```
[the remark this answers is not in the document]
jharveyb: Removed that impl entirely since its no longer used, and I don't think we'll need it in tests for other packages either.

tapgarden/caretaker.go
jharveyb: Confirmed what I posted above in tests.

tapgarden/planter_test.go
```diff
@@ -1205,8 +1210,136 @@ func testFinalizeBatch(t *mintingTestHarness) {
...
+	t.treeStore.FailStore = true
+	t.finalizeBatch(&wg, respChan, &finalizeReq)
+	finalizeErr := <-t.errChan
+	require.ErrorContains(t, finalizeErr, "unable to store tapscript tree")
+
+	// Allow tapscript tree storage to succeed, but force tapscript tree
+	// loading to fail.
+	t.treeStore.FailStore = false
+	t.treeStore.FailLoad = true
+	wg = sync.WaitGroup{}
```
guggero, line 1257: nit: wait group doesn't need explicit initialization.

itest/assets_test.go
```diff
@@ -334,3 +343,170 @@ func testMintAssetNameCollisionError(t *harnessTest) {
...
+		return !bytes.Equal(utxo.TaprootAssetRoot, utxo.MerkleRoot)
+	}
+	mintingOutputWithSibling := fn.Filter(
+		maps.Values(utxos.ManagedUtxos), utxoWithSibling,
+	)
+	require.Len(t.t, mintingOutputWithSibling, 1)
+	genesisWithSibling := mintingOutputWithSibling[0]
+
+	// Extract the fields needed to construct a script path spend.
+	mintTapTweak := genesisWithSibling.MerkleRoot
```
guggero, line 415: Should we also re-compute the merkle root here and make sure it's sibling root + asset commitment root?

itest/assets_test.go
```diff
@@ -334,3 +343,170 @@ func testMintAssetNameCollisionError(t *harnessTest) {
...
+	hashLockWitness := wire.TxWitness{
+		hashLockPreimage, hashLockLeaf.Script, hashLockControlBlockBytes,
+	}
+
+	// Make a non-tap address from Bob to send Alice's genesis UTXO to.
+	burnAddrResp := t.lndHarness.Bob.RPC.NewAddress(&lnrpc.NewAddressRequest{
+		Type: lnrpc.AddressType_TAPROOT_PUBKEY,
+	})
+	burnAddr, err := btcutil.DecodeAddress(
+		burnAddrResp.Address, t.lndHarness.Miner.ActiveNet,
```
guggero, line 451: nit: can use `harnessNetParams`.

guggero (commented):
Awesome work! Looks very close, just a couple of new (and old IMO unaddressed) smaller things.

tapgarden/planter_test.go
jharveyb: Agree, this was a workaround for the planter hanging after failing to freeze a batch - found a fix.

tapgarden/planter_test.go
```diff
@@ -1205,8 +1210,136 @@ func testFinalizeBatch(t *mintingTestHarness) {
...
+	sendConfNtfn()
+
+	// Once the TX is broadcast, the caretaker should run to completion,
+	// storing issuance proofs and updating the batch state to finalized.
+	batchWithSibling := t.assertFinalizeBatch(&wg, respChan, "")
+	require.NotNil(t, batchWithSibling)
+	t.assertNoError()
+	t.assertNoPendingBatch()
+	t.assertNumCaretakersActive(0)
+	t.assertLastBatchState(batchCount, tapgarden.BatchStateFinalized)
```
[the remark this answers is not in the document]
jharveyb: Yes, we do it in the itest but worth doing here as well.

tapgarden/planter_test.go
Roasbeef: Maybe was the issue of "attempted to copy mutex"?

tapdb/sqlc/queries/assets.sql
```diff
@@ -795,6 +795,85 @@ SELECT key_family, key_index
...
+)
+DELETE FROM tapscript_edges
+WHERE edge_id IN (SELECT edge_id FROM tree_info);
+
+-- name: DeleteTapscriptTreeNodes :exec
+DELETE FROM tapscript_nodes
+WHERE NOT EXISTS (
+    SELECT 1
+        FROM tapscript_edges
+        -- Delete any node that is not referenced by any edge.
```
Roasbeef, line 881: 🤔

tapdb/asset_minting_test.go
```diff
@@ -1191,6 +1279,304 @@ func TestGroupAnchors(t *testing.T) {
...
+	err = deleteTapscriptTreeWrapper(t, ctx, tree4Hash[:], assetStore)
+	require.NoError(t, err)
+
+	assertStoredTreeEqual(t, ctx, true, assetStore, tree5Hash[:], tree5)
+}
+
+// TestTapscriptTreeManager tests the functions that implement the
+// TapscriptTreeManager interface. This follows the same actions as
+// TestTapscriptTreeStore, but with higher-level functions.
+func TestTapscriptTreeManager(t *testing.T) {
```
Roasbeef, line 1453: 😍

commitment/tap.go
Roasbeef: IIRC, was removed when we realized that this would actually fail for any v1 (segwit) commitments. 

I think the open question here is: do we need to enumerate all the current/known asset versions? Or is enough that we bind the greater structure (script size, and the asset market). 

My fear is that if we enumerate all the versions here, then we'll need to go back and update this each time for older clients, so not quite soft forkable. Instead, I think for the known versions, we can validate the requirements we know (in this case, the top level hash or w/e hasn't changed), then just return true otherwise. IMO we want to avoid the case where an old client can't swap or interact w/ a new client, as the new client wants their asset in a v3 commitment (uses sha3 or w/e).

commitment/taproot.go
Roasbeef: Would be same indentation fwiw, but can also chalk it up to a personal style thing. As this wants to return an error, `MapOptionZ` can be used as well.

taprpc/mintrpc/mint.proto
```diff
@@ -177,6 +177,12 @@ message FinalizeBatchRequest {
 
     // The optional fee rate to use for the minting transaction, in sat/kw.
     uint32 fee_rate = 2;
+
+    oneof batch_sibling {
```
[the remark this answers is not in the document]
Roasbeef: Yep, these'll show up in the API documentation

Roasbeef (approved):
LGTM 🦋


Once https://github.com/lightninglabs/taproot-assets/pull/787 lands, we can use it to augment the itest by making `tapd` aware of the transfer/spend from the minting output w/ a set of custom scripts.

tapgarden/planter_test.go
jharveyb: One branch of the planter logic on Finalize did not return an error to the caller, but only on `ErrChan` - so the caller would hang waiting for an error, but from the unit test context we can just read `ErrChan` and reset the waitGroup. Fixed the root cause in the planter, so this is no longer needed.

commitment/tap.go
jharveyb: Moved to a separate commit with notes.

guggero:
Needs a rebase, then I'll go for the final review.

tapdb/asset_minting_test.go
guggero: Very nice tests indeed!

guggero (approved):
Looks like the rebase will be trivial, so gave it another look.
Looks like some of my previous comments are still unaddressed. LGTM after those nits are addressed.

commitment/taproot.go
jharveyb: Went with always enforcing these checks before computing the TapHash, as sometimes we're initializing `TapscriptPreimage` from a decoded Proof vs. using a safe constructor.

jharveyb:
Pushed final update with some commits in the middle to finish making `commitment.TapscriptPreimage` a safer type. Added extra validation logic to constructors but also `TapHash()` itself, as that gets used in proof validation logic on a preimage decoded directly from a proof file (so no checks were performed before that call).

Looks like I've introduced a flake to the multi-asset group itest + changed the preimage encoding unintentionally, investigating now. The intent is to not change the preimage encoded format in this PR.

jharveyb:
> Pushed final update with some commits in the middle to finish making `commitment.TapscriptPreimage` a safer type. Added extra validation logic to constructors but also `TapHash()` itself, as that gets used in proof validation logic on a preimage decoded directly from a proof file (so no checks were performed before that call).
> 
> Looks like I've introduced a flake to the multi-asset group itest + changed the preimage encoding unintentionally, investigating now. The intent is to not change the preimage encoded format in this PR.

Rebased + test issues are resolved, ready for final review.

asset/interface.go
```diff
@@ -24,3 +28,29 @@ type GenesisTxBuilder interface {
...
+	LoadTapscriptTree(ctx context.Context,
+		rootHash chainhash.Hash) (*TapscriptTreeNodes, error)
+
+	// DeleteTapscriptTree deletes the Tapscript tree with the given root
+	// hash.
+	DeleteTapscriptTree(ctx context.Context, rootHash chainhash.Hash) error
+}
+
+var (
+	TreeNotFound     = errors.New("tapscript tree not found")
```
guggero, line 54: nit: missing Godoc.

tapgarden/interface.go
```diff
@@ -236,6 +238,9 @@ type MintingStore interface {
 	// key, including the genesis information used to create the group.
 	FetchGroupByGroupKey(ctx context.Context,
 		groupKey *btcec.PublicKey) (*asset.AssetGroup, error)
+
+	CommitBatchTapSibling(ctx context.Context, batchKey *btcec.PublicKey,
```
guggero, line 242: nit: still missing Godoc.

tapgarden/planter.go
```diff
@@ -127,6 +128,11 @@ type stateParamReq[T, S any] struct {
 	param S
 }
 
+type FinalizeParams struct {
```
guggero, line 131: nit: still missing Godoc.

guggero (approved):
Awesome, LGTM :tada:

Roasbeef (approved):
LGTM 🦖

PR #775, opened 2024-01-30
tools: exclude non-project files from go version consistency lint check

ffranr:
I noticed that the lint check (Go versions consistency check) was examining my local `vendor` directory. This directory contains project dependency files, which we shouldn't try to lint. Therefore this PR introduces an exclusion list to avoid checking certain directories.

### DIFF
--- a/tools/check-go-version-dockerfile.sh
+++ b/tools/check-go-version-dockerfile.sh
@@ -25,12 +25,29 @@ fi
 
 target_go_version="$1"
 
-# Search for Dockerfiles in the current directory and its subdirectories
-dockerfiles=$(find . -type f -name "*.Dockerfile" -o -name "Dockerfile")
+# We find target files using the 'find' command in conjunction with the 'read'
+# command. We exclude some directories from the search.
+#
+# We use the 'read' command to help ensure that we correctly handle filenames
+# with spaces, newlines, and special characters. The '-print0' option in 'find'
+# outputs filenames separated by a null character. This allows the 'read'
+# command in the while loop to accurately distinguish each filename. The
+# 'target_files' array is then populated, preserving the integrity of each
+# filename. This approach ensures safe handling of filenames, regardless of
+# their complexity.
+while IFS= read -r -d '' file; do
+    target_files+=("$file")
+done < <(find . \
+    -path ./vendor -prune -o \
+    -type f \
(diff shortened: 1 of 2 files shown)


tools/check-go-version-dockerfile.sh
```diff
@@ -28,8 +28,17 @@ target_go_version="$1"
 # Search for Dockerfiles in the current directory and its subdirectories
 dockerfiles=$(find . -type f -name "*.Dockerfile" -o -name "Dockerfile")
 
+# Exclude non-project files.
+exclusions=(./vendor)
```
guggero, line 32: Since we're doing a `find` above, we should use `-not -path` there (same for the YAML): https://stackoverflow.com/questions/4210042/how-do-i-exclude-a-directory-when-using-find

ffranr: I didn't go down that route because I thought this would crowd the find command if we were to add more elements to the exclusion list. Here are two more ways to do this, but I think both ways are more complicated than what we've got committed right now:

1. 
```
exclude_dirs=(./vendor ./another_directory)
exclusion_pattern=""

for dir in "${exclude_dirs[@]}"; do
    exclusion_pattern+=" -path $dir -prune -o"
done

dockerfiles=$(find . $exclusion_pattern -type f \( -name "*.Dockerfile" -o -name "Dockerfile" \) -print)
```

2.
```
exclusions=(./vendor ./another_directory)
exclusion_params=""

for dir in "${exclusions[@]}"; do
    exclusion_params+=" -e $dir"
done

dockerfiles=$(find . -type f \( -name "*.Dockerfile" -o -name "Dockerfile" \) | grep -v $exclusion_params)
```

guggero: I would go with option 1 then.
Although just to avoid the somewhat tricky syntax of bash lists, we could also just do:
```
dockerfiles=$(find . \
  -path ./vendor -prune -o \
  -path ./another_directory -o \
  -type f \
  \(-name "*.Dockerfile" -o -name "Dockerfile" \) \
  -print\
)
```

ffranr: I've gone down the route that you suggest in your last comment ^

guggero (approved):
tACK, LGTM :tada:

PR #779, opened 2024-01-31
vPSBT: add sighash support (no version bump)

GeorgeTsagk:
# Description 
This PR replaces https://github.com/lightninglabs/taproot-assets/pull/759 , as it introduces simple sighash functionality (`SIGHASH_ALL` and `SIGHASH_NONE`) for vPSBTs without changing the way we create the virtual transactions that are being signed. This allows us to be backwards compatible and does not require a new script version.

### DIFF
--- a/itest/psbt_test.go
+++ b/itest/psbt_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	tap "github.com/lightninglabs/taproot-assets"
 	"github.com/lightninglabs/taproot-assets/address"
@@ -18,8 +19,11 @@ import (
 	"github.com/lightninglabs/taproot-assets/taprpc"
 	wrpc "github.com/lightninglabs/taproot-assets/taprpc/assetwalletrpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	"github.com/lightninglabs/taproot-assets/taprpc/tapdevrpc"
+	"github.com/lightninglabs/taproot-assets/tapscript"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/keychain"
+	"github.com/lightningnetwork/lnd/lntest/wait"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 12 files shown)


itest/psbt_test.go
```diff
@@ -1265,6 +1267,162 @@ func testMultiInputPsbtSingleAssetID(t *harnessTest) {
...
+	)
+	require.NoError(t.t, err)
+
+	const (
+		magicAmtOverride = 42
+	)
+
+	// Edit the already signed PSBT and change the VOutput amount. This
+	// should be ok as we used SIGHASH_NONE for the generated signature.
+	signedPacket.Outputs[1].Asset.Amount = magicAmtOverride
```
GeorgeTsagk, line 1388: this is probably wrong, overriding the outputs should also update the root split commitment as that's where the amount inflation/deflation is checked

tapscript/send.go
```diff
@@ -648,16 +648,20 @@ func SignVirtualTransaction(vPkt *tappsbt.VPacket, signer Signer,
...
-	}
-
 	for idx := range inputs {
 		input := inputs[idx]
 
+		// Create a Taproot Asset virtual transaction representing the
+		// asset transfer. We provide this input's sighash flag and
+		// index in order to construct a virtual transaction that
+		// commits to the desired parts of the asset transfer.
+		virtualTx, _, err := VirtualTx(
```
Roasbeef, line 658: Why do we need to make a new virtual tx for each input? As we're concerned with signing, isn't just `VirtualTxWithInput` enough?

tapscript/send.go
```diff
@@ -666,6 +670,10 @@ func SignVirtualTransaction(vPkt *tappsbt.VPacket, signer Signer,
 			virtualTxCopy, input.Asset(), uint32(idx), nil,
 		)
 
+		fmt.Printf("SGHSH: signing %s:%x (%v)\n",
```
Roasbeef, line 673: Reminder to make it a `Debugf` in the future.

tapscript/tx.go
```diff
@@ -111,7 +111,28 @@ func virtualTxIn(newAsset *asset.Asset, prevAssets commitment.InputSet) (
 
 // virtualTxOut computes the single output of a Taproot Asset virtual
 // transaction based on whether an asset has a split commitment or not.
-func virtualTxOut(txAsset *asset.Asset) (*wire.TxOut, error) {
+func virtualTxOut(txAsset *asset.Asset, sighashFlag txscript.SigHashType) (
+	*wire.TxOut, error) {
+
+	// If the SigHashNone flag is used, we don't need to commit to any
+	// output, so we just create an empty tree.
+	if sighashFlag == txscript.SigHashNone {
```
Roasbeef, line 119: Alternatively, what stops us from just relying on Bitcoin here? We can make the same output as normal, but then we sign it using sighash none. So everything is still committed to as normal, but Bitcoin level part is what blanks out output when signing.

tapscript/tx.go
```diff
@@ -194,6 +219,16 @@ func VirtualTx(newAsset *asset.Asset, prevAssets commitment.InputSet) (
 		err       error
 	)
 
+	// Check that we support the provided sighash flag.
+	switch sighashFlag {
```
Roasbeef, line 223: Same here, sighash can be about the signing, not affect the tx construction persay.

vm/vm.go
```diff
@@ -288,6 +288,69 @@ func (vm *Engine) validateStateTransition(virtualTx *wire.MsgTx) error {
...
 
+		// Assume the default flag of SigHashAll.
+		sighashFlag := txscript.SigHashAll
+
+		const (
+			leafVersionEven = byte(txscript.BaseLeafVersion)
+			leafVersionOdd  = byte(txscript.BaseLeafVersion | 1)
+		)
+
+		for _, txWitness := range witness.TxWitness {
```
Roasbeef, line 299: I think we can pull this out into a helper function, and even just re-use `txscript` more. 

But even then, each signature within a control block can also have its own sighash flag. Thinking about it more, I don't think we need a new tx version? In that now we can sign it, but the Bitcoin layer is what would handle all the verification.

This PR is then more about the RPC level stuff, and the itest to confirm the case.

vm/vm.go
```diff
@@ -345,25 +408,5 @@ func (vm *Engine) Execute() error {
 		}
 	}
 
-	// Now that we know we're not dealing with a genesis state transition,
-	// we'll map our set of asset inputs and outputs to the 1-input 1-output
-	// virtual transaction.
-	virtualTx, inputTree, err := tapscript.VirtualTx(
```
Roasbeef, line 351: Still very important. 

See commetn above, we can retain this by leaning on Bitcoin for the sighash side of things.

itest/psbt_test.go
```diff
@@ -1265,6 +1268,182 @@ func testMultiInputPsbtSingleAssetID(t *harnessTest) {
...
+
+	// Now we deserialize the signed packet again in order to edit it
+	// and then anchor it.
+	signedPacket, err := tappsbt.NewFromRawBytes(
+		bytes.NewReader(signedResp.SignedPsbt), false,
+	)
+	require.NoError(t.t, err)
+
+	// Edit the already signed PSBT and change the output amounts. This
+	// should be ok as we used SIGHASH_NONE for the input's signature.
```
Roasbeef, line 1384: Should also test swapping out the dest script. This is an important component for the non-interactive swap demo.

itest/test_list_on_test.go
```diff
@@ -171,6 +171,10 @@ var testCases = []*testCase{
 		name: "psbt multi send",
 		test: testPsbtMultiSend,
 	},
+	{
+		name: "psbt_sighash_none",
```
Roasbeef, line 175: you can leave out the underscores, the test framework inserts it automatically

Roasbeef (changes requested):

itest/psbt_test.go
Roasbeef: IIUC you removed that check in the diff. 

What we want here is:
  * Alice signs input off to Bob, she doesn't know Bob's script key yet. 
  * Bob takes the signature, places it in a Bitcoin output commitment he controls, then swaps in his actual script key. 
  * Bob can now broadcast a state transition that transfers that asset to him.

The other portion is on the Bitcoin sighash side:
   * Returning to the above, Alice's prev asset outpoint must be spent for that transaction to be valid (Alice's prev input points to that prev outpoint). 
   * Alice signs the PSBT that has 1 BTC input, and 2 BTC output. This needs Bob to provide that 1 BTC input. This uses sighash single + anyonecanpay on the Bitcoin side. 
   * Bob can then take the sig, make a new commitment, attach his input, then broadcast the Bitcoin tx (signs with sighash all).

GeorgeTsagk: > IIUC you removed that check in the diff.

That was part of the closed PR. Here I am not removing any supply checks

> What we want here is:

Guess you're describing the swap flow? This itest is only about a simple check that sighashes work. The trustless swap will be a separate test.

tapscript/tx.go
GeorgeTsagk: Yeap! Makes sense

tapscript/tx.go
GeorgeTsagk: Moved this check to the VM validation level

vm/vm.go
GeorgeTsagk: Yes I don't believe we need a script version bump.

guggero (commented):
Nice, looks pretty good! Mostly nits now :tada:

vm/vm.go
```diff
@@ -288,6 +328,46 @@ func (vm *Engine) validateStateTransition(virtualTx *wire.MsgTx) error {
...
+		virtualTx, inputTree, err := tapscript.VirtualTx(
+			vm.newAsset, vm.prevAssets,
+		)
+		if err != nil {
+			return err
+		}
+
+		// If we are commiting to the outputs, we need to also perform
+		// an inflation check.
+		if sighashFlag != txscript.SigHashNone {
```
GeorgeTsagk, line 358: We probably don't want to **always** do this?

This `if` is useful when vPSBTs are partially constructed and passed around, as `tapd` will run the vm on psbts that are being signed.

But at the same time we don't want to have transfers that are never checked in terms of supply.

For a vPSBT that has only one input of `SigHashNone` we need to signal when the inflation check should be performed or not, when there's no intention of further adding more inputs that commit to all inputs/outputs.

GeorgeTsagk:
diff significantly reduced, also tidied up the commits a bit, PTAL

vm/vm_test.go
```diff
@@ -712,7 +712,7 @@ func TestVM(t *testing.T) {
 			f: scriptTreeSpendStateTransition(
 				t, false, true, txscript.SigHashSingle,
 			),
-			err: nil,
+			err: newErrKind(ErrInvalidSigHashFlag),
```
guggero, line 715: Should we also add a positive test vector with a supported sighash flag?

vm/vm.go
guggero: Copying my answer from chat:

> I think we only should skip the check at sign time but not at validation time. I think I mentioned this before, but the current check we have in the `SignVirtualPsbt` code where we validate the signature directly after signing probably needs to be removed or updated in any case. Otherwise we couldn't use that RPC for multi-signature as it would always fail for the first signer(s). So maybe we can pass in a boolean `isSigning` into the VM so we know we can skip some checks

Roasbeef: If it's about the validation flow for vPSBTs, then I think for that flow, we can just compute the tapscript sighash directly and validate the signature given the available information. This way, we don't need a special case in the VM that could possibly be exploited with some edge case (valid sig that actually_does_ do inflation).

I think by the time we get to the layer of fully validating a proof file, we want the strictness checks possible to avoid inadvertent vectors that allow side stepping the rule set.

vm/vm.go
```diff
@@ -288,6 +328,46 @@ func (vm *Engine) validateStateTransition(virtualTx *wire.MsgTx) error {
 				spew.Sdump(witness.PrevID))
 		}
 
+		// Assume the default flag of SigHashAll.
+		sighashFlag := extractSighashFromWitnessStack(witness.TxWitness)
+
+		// Check that we support the provided sighash flag.
+		switch sighashFlag {
+		case txscript.SigHashDefault, txscript.SigHashAll,
```
Roasbeef, line 336: Why do we need to officially support some flags vs others? Stuff like single or anyonecan pay has no bearing at this level. It's also the case that the logic above doesn't _exhaustively_ extract sighashes: it assumes keyspend only. All sigs within the witness can _also_ have sighashes, eg: someone doing `OP_CHECKSGIADD`.

vm/vm.go
```diff
@@ -102,6 +102,46 @@ func matchesAssetParams(newAsset, prevAsset *asset.Asset,
 	return nil
 }
 
+// extractSighashFromWitnessStack attempts to extract the sighash flag from a
+// witness stack. If no sighash flag is found, the default sighash flag is
+// returned.
+func extractSighashFromWitnessStack(
```
Roasbeef, line 108: I think we should eliminate this function, as it only assumes a keyspend sighash. A valid witness may have > 1 signature (multi-sig, etc) in the witness (script path).

vm/vm.go
Roasbeef: Reading in more of the diff: this won't work as intended if the virtual tx at this point is _incomplete_ (output doesn't have the true value/amount). 

Also the check (if it were needed, I don't think it is) needs to actually apply the sighash mask to extract the value rather than do a plain equality check.

vm/vm.go
Roasbeef: Also sighash values are masks, you can't just compare directly.

Roasbeef (changes requested):

vm/vm.go
GeorgeTsagk: > Stuff like single or anyonecan pay has no bearing at this level

So should I allow them even though they'll have no effect, or a malformed effect?

vm/vm.go
```diff
@@ -102,6 +102,69 @@ func matchesAssetParams(newAsset, prevAsset *asset.Asset,
 	return nil
 }
 
+// ValidateVPsbt is a helper method that checks a vPSBT for validity. This
+// includes checking the provided signatures against the fields of the transfer
+// they commit to.
+func ValidateVPsbt(newAsset *asset.Asset, splitAssets []*commitment.SplitAsset,
+	prevAssets commitment.InputSet) error {
```
GeorgeTsagk, line 109: Due to a circular dependency issue that would eventually be introducing by merging both this PR and https://github.com/lightninglabs/taproot-assets/pull/787 on `main`, this func was changed to not accept a `Vpkt` as that would import `taproot-assets/tappsbt`, which would be the source of the circular dep. issue

tapscript/interface.go
```diff
@@ -17,6 +17,15 @@ type TxValidator interface {
 		prevAssets commitment.InputSet) error
 }
 
+// VPsbtValidator is the interface used to validate a vPSBT. It is less strict
+// compared to the TxValidator, as it focused on witness validation. This method
+// may be used in partially constructed transfers.
+type VPsbtValidator interface {
+	ValidateVPsbt(newAsset *asset.Asset,
```
guggero, line 24: nit: interface methods also need Godoc comments. Even if they are kind of identical to the interface comment for 1-method interfaces.

tapscript/send_test.go
```diff
@@ -859,7 +863,7 @@ var prepareOutputAssetsTestCases = []testCase{{
 // TestSignVirtualTransaction tests edge cases around signing a witness for
 // an asset transfer and validating that transfer with the Taproot Asset VM.
 func TestSignVirtualTransaction(t *testing.T) {
-	t.Parallel()
+	// t.Parallel()
```
guggero, line 866: nit: should not comment this out.

vm/vm.go
guggero: Since this isn't a vPSBT we're validating anymore, maybe we should rename the interface and method. And while we're at it, give it a more descriptive name. Otherwise what's the difference in functionality between the `TxValidator` and the `VPsbtValidator`?
One verifies the finished transaction in a proof while the other one validates a (partial) signature that was just created. So maybe `SigningValidator` or something like that would be more descriptive?

guggero: Also maybe extending the comment a bit explaining _why_ exactly this is different from the `TxValidator` would probably be good. For example that in case of a multi-signature we can't fully verify the full witness, only part of it (for which we might add a TODO in the code somewhere).

guggero (commented):
Hmm, I'm confused. Did something get lost in a rebase somewhere? Because I don't see the actual changes for supporting `SigHashNone` anymore... Also makes me wonder why the unit/integration tests still run file? Or am I missing something?

guggero (approved):
> Hmm, I'm confused. Did something get lost in a rebase somewhere? Because I don't see the actual changes for supporting `SigHashNone` anymore... Also makes me wonder why the unit/integration tests still run file? Or am I missing something?

Okay, I guess I missed the part that [we already hooked in the vPSBTs sighash flag into the actual sign descriptor](https://github.com/lightninglabs/taproot-assets/blob/478e2bbbf1657a02eb159da5b440726c9e2bbefa/tapscript/tx.go#L343). So all good here (after addressing nits).

Roasbeef (approved):
LGTM 🕸️

PR #784, opened 2024-02-06
tapdb: fix migration 15

guggero:
We assumed nobody would have multiple witness entries at this point, as they could only be produced when using the vPSBT API. But a user reported problems when running the migration 15 which indicates they indeed had multiple witnesses per asset (possibly due to multiple imports of the same proof which would be a whole separate issue and likely needs to be fixed in their DB manually).

This commit makes sure we insert a non-duplicate default value (we just pick the witness ID as we just care about the order not the actual value).


Roasbeef (approved):
LGTM 🦖

Roasbeef (approved):
LGTM 🐍

PR #787, opened 2024-02-07
[psbt saga part 3/4]: taprpc: add CommitVirtualPsbts and PublishAndLogTransfer RPC methods

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/805.

This is a first draft towards a new series of RPCs that allows us to drive the send logic completely from outside of the daemon.
That means we'll likely want to bypass the freighter partially or completely, since for TAP channels not all state and/or proofs of all state updates should be reflected in the database.

We'll likely want an additional RPC that allows us to store the output of the `CommitVirtualPsbts` call in the database, realizing the actual entries in the transfers and assets tables.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/787)


guggero:
@GeorgeTsagk it's still in draft, needs a lot of cleanup and some decisions. But should have what you need to build the sighash swap itest.

guggero:
@Roasbeef I have the following flow in mind for driving an asset send mostly from outside of the daemon (for the `ASSET_VERSION_V1` use case, meaning segregated asset witness):

1. User calls `FundVirtualPsbt` or creates vPSBTs from a list of known assets they want to spend (e.g. TAP channels, the list of proofs from the `tx_asset_proof` funding messages).
2. User calls `CommitVirtualPsbts` to create the output commitments for the vPSBTs, this creates proof suffixes for each asset output (lacking the asset witness and the chain header information).
3. The user then finalizes the proofs by signing the vPSBTs and "injecting" the asset witness into the proof suffixes.
4. Optional step: for transactions that should be reflected in the TAP database, the user would call `DeliverVirtualPsbts` (or whatever, this method doesn't exist yet), that would then create the `asset_transfer` entries, wait for confirmation and then ship the proofs to recipients using the proof courier.
    1. This step would only be called for transactions that the user should be aware of. For example normal sends (interactive or non-interactive) or the funding transaction. It would however _not_ be called for commitment transactions, as they should IMO not be reflected in the TAP database.
    2. We would probably call this step if a channel commitment transaction is broadcast due to a force-close. Then it would make sense to realize the action as a transfer and see the balances.

My current question is mostly around step 4: To transport the unfinished proof suffixes and optional proof courier addresses from step 3 to step 4 we would need to place them somewhere. I would argue that the vPSBT outputs would be a good place for that. But I distinctly remember that you were against storing proofs or proof courier addresses in the vPSBTs. Do you remember why you argued that way? Mostly due to the size of the vPSBTs?

taprpc/assetwalletrpc/assetwallet.proto
```diff
@@ -180,6 +189,85 @@ message AnchorVirtualPsbtsRequest {
...
+    repeated bytes virtual_psbts = 1;
+
+    /*
+    The template of the BTC level anchor transaction that the virtual
+    transactions should be mapped to. The template is expected to already
+    contain all asset related inputs and outputs corresponding to the virtual
+    transactions given above. This can be achieved by using
+    tapfreighter.PrepareAnchoringTemplate for example.
+    */
+    bytes anchor_psbt = 2;
```
GeorgeTsagk, line 209: > is expected to already contain all asset related inputs and outputs corresponding to the virtual transactions given above.

Can it also include inputs or outputs that are unrelated to any asset?

rpcserver.go
```diff
@@ -1875,6 +1880,337 @@ func (r *rpcServer) AnchorVirtualPsbts(ctx context.Context,
...
+	}
+
+	// Make sure we decorate all asset inputs with the correct internal key
+	// derivation path (if it's indeed a key this daemon owns).
+	for idx := range pkt.Inputs {
+		pIn := &pkt.Inputs[idx]
+
+		// We only care about asset inputs which always specify a
+		// Taproot merkle root.
+		if len(pIn.TaprootMerkleRoot) == 0 {
```
GeorgeTsagk, line 1916: at this point, what does it mean if a `pIn` does not specify a taproot merkle root?

rpcserver.go
```diff
@@ -1875,6 +1880,337 @@ func (r *rpcServer) AnchorVirtualPsbts(ctx context.Context,
...
+		fundRequest.Fees = &walletrpc.FundPsbtRequest_SatPerVbyte{
+			SatPerVbyte: fee.SatPerVbyte,
+		}
+
+	default:
+		return nil, fmt.Errorf("unknown fee type")
+	}
+
+	lndWallet := r.cfg.Lnd.WalletKit
+	fundedPacket, changeIndex, lockedUTXO, err := lndWallet.FundPsbt(
```
GeorgeTsagk, line 2020: IIUC at this point there is a strict check that the inputs amount must equal the outputs amount.

If I want to create an incomplete anchorTx, could we somehow signal that there may be an input or output missing, hence no need to perform strict checks?

tappsbt/interface.go
```diff
@@ -529,6 +530,14 @@ type VOutput struct {
...
 	// TaprootDerivationPath fields of the PSBT output.
 	ScriptKey asset.ScriptKey
+
+	// ProofSuffix is the optional new transition proof blob that is created
+	// once the asset output was successfully committed to the anchor
+	// transaction referenced above. The proof suffix is not yet complete
+	// since the header information needs to be added once the anchor
+	// transaction was confirmed in a block. This field cannot be of type
+	// proof.Proof directly because that would cause a circular dependency.
+	ProofSuffix []byte
```
GeorgeTsagk, line 540: after reading the rest of the commits I think it's a bit unclear why this proof suffix needs to be exported now

previously we were only anchoring one vPSBT on a anchoringTx, why didn't that require this field too?

tappsbt/decode.go
```diff
@@ -278,11 +278,15 @@ func (o *VOutput) decode(pOut psbt.POutput, txOut *wire.TxOut) error {
 			),
 		},
 		{
-			key: PsbtKeyTypeOutputAssetVersion,
+			key: PsbtKeyTypeOutputTapAssetVersion,
```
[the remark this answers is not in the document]
guggero: This is just the variable name, nothing changed otherwise.

rpcserver.go
guggero: No, the input and output sum doesn't have to be equal here. That's exactly the task for `lndWallet.FundPsbt` to figure out if additional inputs are required to fund the outputs.
But that does mean that all _outputs_ need to be specified here, otherwise we wouldn't know what to fund.

tappsbt/interface.go
guggero: See my long-form comment above, just before your review. Before this change the proofs never "left" the daemon, they were copied from one state to the next in a separate data structure within the freighter. But now, because we return the incomplete state back to the user, we need to transport the proofs somewhere. I think the vPSBT itself isn't too bad a place, but we might need to discuss that.

Roasbeef:
> Optional step: for transactions that should be reflected in the TAP database, the user would call DeliverVirtualPsbts (or whatever, this method doesn't exist yet), that would then create the asset_transfer entries, wait for confirmation and then ship the proofs to recipients using the proof courier.

Yeah I think that makes sense. Alternatively, the user may wish to just have the set of participants side load the proofs as well. Though either way you'd want to make sure that the sender's DB is updated to reflect that those outputs are now spent. In theory, we could also just watch all the outpoints we believe to be unspent for spends. In that case though, we'd still want to know things like if a change output was created that _does_ still store outputs.

itest/utils.go
```diff
@@ -88,7 +88,9 @@ func MineBlocks(t *testing.T, client *rpcclient.Client,
 	var blockHashes []*chainhash.Hash
 
 	switch backend {
-	case rpcclient.BitcoindPost19:
+	case rpcclient.BitcoindPre19, rpcclient.BitcoindPre22,
```
Roasbeef, line 91: 👍

tappsbt/interface.go
Roasbeef: Yeah I think this is a part of the more interactive vPSBT ceremony. When you're operating mainly outside the daemon, you want to prove to someone after signing that the new outputs was properly constructed (no duplicate commitments, etc), so you need to transmit this data _somehow_, and it makes sense to just piggy back it on top of what you were already sending to each other.

tappsbt/interface.go
```diff
@@ -750,3 +759,19 @@ func deserializeTweakedScriptKey(pOut psbt.POutput) (*asset.TweakedScriptKey,
 		Tweak:  tweak,
 	}, nil
 }
+
+// Encode encodes the virtual packet into a byte slice.
+func Encode(vPkt *VPacket) ([]byte, error) {
```
Roasbeef, line 764: Kinda surprised we didn't have something like this already? Perhaps we should have the encode be a method to take a writer? Same below w/ a reader.

Roasbeef: Ah I see below just uses the decode that takes a reader.

tapfreighter/chain_porter.go
```diff
@@ -830,8 +830,11 @@ func (p *ChainPorter) importLocalAddresses(ctx context.Context,
 func createDummyOutput() *wire.TxOut {
 	// The dummy PkScript is the same size as an encoded P2TR output.
 	newOutput := wire.TxOut{
-		Value:    int64(tapscript.DummyAmtSats),
-		PkScript: make([]byte, 34),
+		Value: int64(tapscript.DummyAmtSats),
+		PkScript: append(
+			[]byte{txscript.OP_1, txscript.OP_DATA_32},
```
Roasbeef, line 835: 👍

taprpc/assetwalletrpc/assetwallet.proto
```diff
@@ -180,6 +189,85 @@ message AnchorVirtualPsbtsRequest {
     repeated bytes virtual_psbts = 1;
 }
 
+message CommitVirtualPsbtsRequest {
+    /*
+    The list of virtual transactions that should be mapped to the given BTC
+    level anchor transaction template. The virtual transactions are expected to
+    be signed (or use ASSET_VERSION_V1 with segregated witness to allow for
```
Roasbeef, line 196: 👍 

We should also validate that req when they're passed in.

taprpc/assetwalletrpc/assetwallet.proto
```diff
@@ -180,6 +189,85 @@ message AnchorVirtualPsbtsRequest {
...
+    contain all asset related inputs and outputs corresponding to the virtual
+    transactions given above. This can be achieved by using
+    tapfreighter.PrepareAnchoringTemplate for example.
+    */
+    bytes anchor_psbt = 2;
+
+    oneof anchor_change_output {
+        /*
+        Use the existing output within the anchor PSBT with the specified
+        index ax the change output. Any leftover change will be added to the
```
Roasbeef, line 214: ax the -> as the

rpcserver.go
```diff
@@ -56,6 +57,10 @@ var (
 	// ServerMaxMsgReceiveSize is the largest message our server will
 	// receive.
 	ServerMaxMsgReceiveSize = grpc.MaxRecvMsgSize(lnrpc.MaxGrpcMsgSize)
+
+	// P2TrChangeType is the type of change address that should be used for
+	// funding PSBTs, as we'll always want to use P2TR change addresses.
+	P2TrChangeType = walletrpc.ChangeAddressType_CHANGE_ADDRESS_TYPE_P2TR
```
Roasbeef, line 63: Make the `r` capitalized as well?

rpcserver.go
Roasbeef: ^ Could add that to the comment here.

rpcserver.go
```diff
@@ -1875,6 +1880,337 @@ func (r *rpcServer) AnchorVirtualPsbts(ctx context.Context,
...
+		derivation, trDerivation := tappsbt.Bip32DerivationFromKeyDesc(
+			keyDesc, r.cfg.ChainParams.HDCoinType,
+		)
+		pIn.Bip32Derivation = []*psbt.Bip32Derivation{derivation}
+		pIn.TaprootBip32Derivation = []*psbt.TaprootBip32Derivation{
+			trDerivation,
+		}
+	}
+
+	// TODO(guggero): Validate inputs/outputs by going through all vPSBTs.
```
Roasbeef, line 1961: IIUC, validation here would also include the segwit version thing.

rpcserver.go
```diff
@@ -1875,6 +1880,337 @@ func (r *rpcServer) AnchorVirtualPsbts(ctx context.Context,
...
+	}
+
+	// Validate the actual fee rate of the transaction.
+	txFees, err := fundedPacket.GetTxFee()
+	if err != nil {
+		return nil, fmt.Errorf("error calculating transaction fees: %w",
+			err)
+	}
+
+	// TODO(guggero): Create a defer that unlocks the UTXOs if we error out
```
Roasbeef, line 2034: 👍

rpcserver.go
```diff
@@ -1875,6 +1880,337 @@ func (r *rpcServer) AnchorVirtualPsbts(ctx context.Context,
 	}, nil
 }
 
+// CommitVirtualPsbts creates the output commitments and proofs for the given
+// virtual transactions by committing them to the BTC level anchor transaction.
+// In addition, the BTC level anchor transaction is funded and prepared up to
+// the point where it is ready to be signed.
+func (r *rpcServer) CommitVirtualPsbts(ctx context.Context,
+	req *wrpc.CommitVirtualPsbtsRequest) (*wrpc.CommitVirtualPsbtsResponse,
```
Roasbeef, line 1888: I think we could break up this method into a few helper functions to make it easier to scan at a glance. Can wait till finalization + tests, etc.

guggero:
Taking the PR out of draft as it is feature complete now. Still want to add more explicit integration and unit tests (and also need to address some of the review comments). But the PR should not require any additional functionality and is ready for a first round IMO.

guggero:
Just as a note to reviewers. The main integration test that currently validates the logic in these two PRs is in https://github.com/lightninglabs/taproot-assets/pull/796. But I will add more specific ones to this PR as well.

rpcserver.go
```diff
@@ -1726,10 +1731,30 @@ func (r *rpcServer) FundVirtualPsbt(ctx context.Context,
 			"specified")
 	}
 
-	var err error
+	// Extract the passive assets that are needed for the fully RPC driven
+	// flow.
+	passivePackets, err := r.cfg.AssetWallet.CreatePassiveAssets(
```
GeorgeTsagk, line 1736: maybe this hunk should be in a separate commit?

rpcserver.go
```diff
@@ -1867,6 +1892,296 @@ func (r *rpcServer) AnchorVirtualPsbts(ctx context.Context,
...
+	activePackets := make([]*tappsbt.VPacket, len(req.VirtualPsbts))
+	for idx := range req.VirtualPsbts {
+		activePackets[idx], err = tappsbt.Decode(req.VirtualPsbts[idx])
+		if err != nil {
+			return nil, fmt.Errorf("error decoding virtual "+
+				"packet at index %d: %w", idx, err)
+		}
+	}
+
+	passivePackets := make([]*tappsbt.VPacket, len(req.PassiveAssetPsbts))
```
GeorgeTsagk, line 1921: does this mean that the user has to be aware & responsible of which assets are being passively carried when creating a vPSBT?

tapfreighter/chain_porter.go
```diff
@@ -882,23 +886,20 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
...
-		vPacket := currentPkg.VirtualPacket
-		firstRecipient, err := vPacket.FirstNonSplitRootOutput()
-		if err != nil {
-			return nil, fmt.Errorf("unable to get first "+
-				"interactive output: %w", err)
+		for idx := range currentPkg.VirtualPackets {
+			vPkt := currentPkg.VirtualPackets[idx]
+
+			logPacket(vPkt, "Constructing new Taproot Asset "+
+				"commitments")
```
GeorgeTsagk, line 895: shouldn't this loop have more things in its body? Is logging the only purpose?

tapfreighter/parcel.go
```diff
@@ -286,7 +286,101 @@ func (p *PreSignedParcel) kit() *parcelKit {
...
+	}
+}
+
+// kit returns the parcel kit used for delivery.
+func (p *PreAnchoredParcel) kit() *parcelKit {
+	return p.parcelKit
+}
+
+// Validate validates the parcel.
+func (p *PreAnchoredParcel) Validate() error {
```
GeorgeTsagk, line 359: since we expect the signatures to exist, shouldn't we also do some vm/sig checks?

GeorgeTsagk: if that is not the scope of `Validate` then maybe define what exactly we're validating in the comment above?

itest/universe_pagination_test.go
```diff
@@ -22,15 +22,15 @@ const (
...
+	testPageSizeSmall = 10
 
 	// testGroupSize is the size of the asset group we mint in the
 	// testUniversePaginationSimple test.
 	testGroupSize = 79
 )
 
 func testUniversePaginationSimple(t *harnessTest) {
-	mintSize := 255
+	mintSize := 50
```
GeorgeTsagk, line 33: was the issue here that we had 255 mints and a page size of 2, resulting in 100++ pages (and calls)?

GeorgeTsagk (commented):
Looks good so far! +tACK :test_tube:

rpcserver.go
guggero: For the new flow (`FundVirtualPsbts -> CommitVirtualPsbts -> PublishAndLogTransfer`) yes. For any existing flow, no.

tapfreighter/parcel.go
guggero: Good point. Added the comment to the `Validate` method.

itest/universe_pagination_test.go
guggero: Correct. Reducing this took more than 10 seconds off the itest.

guggero:
Here's a quick update of what I would still like to get in to either this PR or part 4:

TODOs
 - [ ] validate asset input completeness
 - [ ] unlock lnd utxos on error: Should we unlock UTXOs in the porter if publishing the TX fails?
 - [ ] gate commitvirtualpsbts on lnd v0.18

guggero:
Validating the input assets turned out to be a bit more involved and require a couple of more refactor commits :sweat_smile: But I'm sure it's totally worth it as we now can also validate we properly spend all assets in an input commitment, even with the existing send flow of the freighter.

guggero:
All TODOs addressed, ready for final review :tada:

rpcserver.go
Roasbeef: Do we have any measures in place to protect a user from accidentally urning assets if they don't pass them in properly to `CommitVirtualPsbts`?

Roasbeef (commented):
> Reviewed 3 of 3 files at r1, 3 of 3 files at r2, 12 of 12 files at r3, 37 of 37 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/787)* status: all files reviewed, 25 unresolved discussions (waiting on @GeorgeTsagk and @guggero)

___
*[`fn/errors.go` line 39 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBMTG0-hIITJanwaA5:-NsBMTG0-hIITJanwaA6:b-kdda92) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/fn/errors.go#L39)):*
> ```Go
> 
> // IsRpcErr returns true if the given error is a gRPC error with the given
> // candidate error as the cause.
> ```

Comment talks about a gRPC error, but doesn't actually use the `status` package? 
___
*[`itest/psbt_test.go` line 1635 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBOGTN0kPPhiMD0SHv:-NsBOGTN0kPPhiMD0SHw:b-i1ns4d) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/itest/psbt_test.go#L1635)):*
> ```Go
> 		t.t, t.lndHarness.Miner.Client, t.tapd,
> 		[]*mintrpc.MintAssetRequest{
> 			issuableAssets[0],
> ```

Don't both of these need to be the v1 segwit version to work properly with the commit flow?
___
*[`itest/test_list_on_test.go` line 48 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBOzYW-SPEf6aAW74F:-NsBOzYW-SPEf6aAW74G:bja4pr9) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/itest/test_list_on_test.go#L48)):*
> ```Go
> 	},
> 	// Skipping due to re-org bug in btcd.
> 	// TODO(guggero): Re-enable this test once we have a re-org fix for
> ```

As a ref, this is the fix: https://github.com/btcsuite/btcd/pull/2123
___
*[`tapfreighter/wallet.go` line 428 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBPF-w7J0x0bDzavjB:-NsBPF-w7J0x0bDzavjC:b44u26) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/4035264470ad4d546da8056d4aa5b74861c2a524/tapfreighter/wallet.go#L428)):*
> ```Go
> 
> // passiveAssetVPacket creates a virtual packet for the given passive asset.
> func passiveAssetVPacket(params *address.ChainParams, passiveAsset *asset.Asset,
> ```

😆
___
*[`tapfreighter/wallet.go` line 463 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBPYk--ary6TwOVYYI:-NsBPYk--ary6TwOVYYJ:bh9wm9k) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/tapfreighter/wallet.go#L463)):*
> ```Go
> 	// If we didn't find the anchor in an active packet, something
> 	// definitely went wrong.
> 	emptyPrevID := asset.PrevID{}
> ```

Style nit: can be `var emptyPrevID asset.PrevID`
___
*[`tapfreighter/wallet.go` line 707 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBQ1Jq2lPNgrAmM8HK:-NsBQ1Jq2lPNgrAmM8HL:b-l7y6gp) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/4035264470ad4d546da8056d4aa5b74861c2a524/tapfreighter/wallet.go#L707)):*
> ```Go
> 	}
> 
> 	// Gather Taproot Asset commitments from the selected anchored assets.
> ```

Oh weird, looks like we never even used this variable and it was just left over. 
___
*[`tapsend/send.go` line 1352 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBTpXYBr3W7oCfdwwb:-NsBTpXZ5q0d7Vm-4YyR:b44u2e) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/tapsend/send.go#L1352)):*
> ```Go
> // commitment.
> func ExtractUnSpendable(c *commitment.TapCommitment) []*asset.Asset {
> 	return fn.Filter(c.CommittedAssets(), func(a *asset.Asset) bool {
> ```

😎
___
*[`tapsend/send.go` line 1389 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBTykq2FT4cnqj9u6h:-NsBTykq2FT4cnqj9u6i:b-rpmjgh) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/tapsend/send.go#L1389)):*
> ```Go
> // RemovePacketsFromCommitment removes all assets within the virtual
> // transactions from the given commitment.
> func RemovePacketsFromCommitment(c *commitment.TapCommitment,
> ```

Really dig the new approach here based on a pipeline of smaller composed functions!

guggero (commented):
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/787)* status: 8 of 37 files reviewed, 25 unresolved discussions (waiting on @GeorgeTsagk and @Roasbeef)

___
*[`fn/errors.go` line 39 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBMTG0-hIITJanwaA5:-NsDOaenELHbvLqowEVH:boac1ac) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/fn/errors.go#L39)):*

Comment talks about a gRPC error, but doesn't actually use the `status` package?
</blockquote></details>

Yes, I think if an error comes from gRPC but isn't wrapped properly (e.g. in `lndclient`) then the status package doesn't recognize it anymore and we need to do string comparison.
___
*[`itest/psbt_test.go` line 1635 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBOGTN0kPPhiMD0SHv:-NsDPK8-9PAzG_bh-n-M:b-ekhh4j) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/itest/psbt_test.go#L1635)):*

Don't both of these need to be the v1 segwit version to work properly with the commit flow?
</blockquote></details>

Only if we _don't_ sign them before calling the commit RPC. But we do manually sign the passive assets in this test. The TAP channel simulation itest will add a test case for unsigned V1 assets in the commit flow.
___
*[`tapfreighter/wallet.go` line 463 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBPYk--ary6TwOVYYI:-NsDQKM7BiI0BKjmNt2Y:b-896fix) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/tapfreighter/wallet.go#L463)):*

Style nit: can be `var emptyPrevID asset.PrevID`
</blockquote></details>

Done.
___
*[`tapfreighter/wallet.go` line 707 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBQ1Jq2lPNgrAmM8HK:-NsDQObK6mILdUDCFKBU:bga1p2f) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/4035264470ad4d546da8056d4aa5b74861c2a524/tapfreighter/wallet.go#L707)):*

Oh weird, looks like we never even used this variable and it was just left over.
</blockquote></details>

Yeah, I don't know why no linter found this earlier...
___
*[`tapsend/send.go` line 1389 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-NsBTykq2FT4cnqj9u6h:-NsDQXhYDbcvhQYMfoH4:buxgkh6) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/tapsend/send.go#L1389)):*

Really dig the new approach here based on a pipeline of smaller composed functions!
</blockquote></details>

Nice, thanks! Also much easier to unit test that way.

guggero:
I think I addressed all comments (not sure, still getting used to Reviewable).

guggero:
I was able to re-enable the re-org itest by bumping to the latest version of `btcd` :tada:

GeorgeTsagk (approved):
had another look due to the many force updates, still LGTM :100:

Roasbeef (approved):
:lgtm:

> Reviewed 30 of 30 files at r25, 4 of 4 files at r26, 2 of 2 files at r27, 2 of 2 files at r28, 2 of 2 files at r29, 5 of 5 files at r30, 3 of 3 files at r31, 3 of 3 files at r32, 2 of 2 files at r33, 2 of 2 files at r34, 3 of 3 files at r35, 3 of 3 files at r36, 11 of 13 files at r37, 22 of 22 files at r38, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/787#-:-NtTqNIS0vmydpfQsAo-:bnfp4nl)* status: all files reviewed, 24 unresolved discussions (waiting on @GeorgeTsagk and @guggero)

PR #794, opened 2024-02-07
multi: introduce new meta type for json strings

### LINKED ISSUE #783: spec+asset: move to officially designate asset meta type 1 as a JSON blob
The meta field was always intended to allow developers and users to specify custom meta data for assets. So far, we just have the first type, `0` which is an opaque blob of bytes. Rather than try to enumerate each new possilbe meta type, a strong contender for something more flexible is just a JSON blob here. 

By designating asset meta type `1` as a JSON blob, we also give wallets, explorers and other tools a hint on how to properly parse the asset meta data. 

### Steps To Completion 

- [ ] Update the spec and code to specify the new asset meta type. 

- [ ] Update RPCs and CLI as necessary to provide streamlined developer UX.

Roasbeef:
In this PR, we designate meta type `1` as a JSON string meta. We perform upfront validation at the CLI and RPC level, but retain the meta hash only validation at the protocol level (proof validity only cares about hashing, and not the structure)

Fixes #783


cmd/tapcli/assets.go
```diff
@@ -142,6 +145,30 @@ func parseAssetType(ctx *cli.Context) (taprpc.AssetType, error) {
 	}
 }
 
+func parseMetaType(metaType string,
+	metaBytes []byte) (taprpc.AssetMetaType, error) {
+
+	switch metaType {
+
+	case "opaque":
```
guggero, line 153: Doesn't this change disallow people to play around with custom meta types on the CLI?
Maybe in the default case we should try to parse the type as a number and then use that and only fail if it's not a valid number? So people can still experiment with types > 1.

itest/asset_meta_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	ctxb := context.Background()
+	for _, tc := range testCases {
+		t.t.Run(tc.asset.Asset.Name, func(tt *testing.T) {
+			_, err := t.tapd.MintAsset(ctxb, tc.asset)
+			if err != nil {
+				if tc.errString == "" {
+					tt.Fatalf("unexpected error: %v", err)
+				}
+				if !strings.Contains(
```
guggero, line 74: nit: use `require.ErrorContains`?

guggero (approved):
Nice, LGTM :tada:

itest/asset_meta_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// testAssetMeta tests the validation+parsing logic for asset meta data.
+func testAssetMeta(t *harnessTest) {
+	// In this test, we'll attempt to issue the following assets with
+	// distinct meta type. Within each test case, negative failure
+	// scenarios may exist.
+	jsonType := taprpc.AssetMetaType_META_TYPE_JSON
+	testCases := []struct {
```
jharveyb, line 19: We could test support for unknown meta types here as well.

jharveyb (commented):
LGTM! Some small lint failures.

cmd/tapcli/assets.go
Roasbeef: Ah sure, modified to retain that ability, but still allow these first class aliases.

itest/asset_meta_test.go
Roasbeef: Ah didn't know that existed, swapped to use

itest/asset_meta_test.go
Roasbeef: done!

PR #795, opened 2024-02-08
itest: use configurable timeout for load test

guggero:
As investigated by @calvinrzachman, here's a small fix to make the load test more reliable.

Because things can take a bit longer the more addresses and events there are, the initial proof retrieval attempt of the universe proof courier can fail, causing it to go into a 30 second backoff. Because we used a hard coded timeout of 30 seconds for the transfer to complete, this sometimes failed the load test.
By making sure we use the same, configurable, timeout where relevant, we make the send process more reliable.

In addition, the following lines should be added to the load test nodes to shorten the duration of the backoff:

```
universerpccourier.initialbackoff=2s
universerpccourier.maxbackoff=30s
```

### DIFF
--- a/itest/assertions.go
+++ b/itest/assertions.go
@@ -688,8 +688,19 @@ func AssertAddrEvent(t *testing.T, client taprpc.TaprootAssetsClient,
 	addr *taprpc.Addr, numEvents int,
 	expectedStatus taprpc.AddrEventStatus) {
 
+	AssertAddrEventCustomTimeout(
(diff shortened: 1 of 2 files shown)


calvinrzachman (approved):
Looks good to me! I gave this + your suggested configuration change a go in the load test environment and it resolves the flakiness I had been seeing. Thanks for looking at this one 🙏

Roasbeef (approved):
LGTM 🌂

PR #796, opened 2024-02-09
[psbt saga part 4/4]: Multisig demo

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/787.

This PR adds an integration test that demonstrates how multi signatures can be used on both the BTC and asset level.
In the demo we use the following Tapscript to encumber the BTC level anchor output:
```
<key_1> OP_CHECKSIG <key_2> OP_CHECKSIGADD 2 OP_EQUAL
```
This is basically a 2-of-2 multi signature. To make sure the key spend path cannot be taken, we use the NUMS public key as the internal key.

On the asset level, we use a script key that is the MuSig2 combined key of two script keys.

This means that the 1000 units of assets sent to the constructed TAP address will be co-owned by both `tapd` instances, on the asset and BTC level.

We then spend from that address again, just to demonstrate how the witnesses for such transfers would be created.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/796)


itest/multisig_test.go
```diff
@@ -0,0 +1,597 @@
...
+		AddOp(txscript.OP_EQUAL).
+		Script()
+	require.NoError(t.t, err)
+	btcTapLeaf := txscript.TapLeaf{
+		LeafVersion: txscript.BaseLeafVersion,
+		Script:      btcTapscript,
+	}
+
+	// The actual internal key of the BTC level Taproot output will be the
+	// provably un-spendable NUMS key.
```
GeorgeTsagk, line 88: are we using NUMS for a security reason, or just to keep things simple? Couldn't the internal key be a alice<>bob combined key?

GeorgeTsagk (approved):
Very well documented and easy-to-follow itest :100:

itest/multisig_test.go
guggero: Just for demonstration purposes, to disable the key spend. But yes, since this is a 2-of-2, the internal key could've just been a combined MuSig2 key of the two parties.
But since the demo is for showing how to execute a sibling tap leaf, we constructed it this way.

Roasbeef (approved):
:lgtm: 🌙

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 10 of 10 files at r4, 11 of 11 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/796#-:-NsBh9Xn0I5IPYpR_qaw:bkt6kgc)* status: all files reviewed, 2 unresolved discussions (waiting on @GeorgeTsagk and @guggero)

___
*[`tapdb/sqlc/queries/assets.sql` line 296 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/796#-NsBf_WiBwmGjwHzHxLd:-NsBf_WiBwmGjwHzHxLe:bpf0vvg) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/0e136b7846781832426944b7f56333992a897869/tapdb/sqlc/queries/assets.sql#L296)):*
> ```TSQL
> WHERE spent = FALSE
> GROUP BY assets.genesis_id, genesis_info_view.asset_id,
>          version, genesis_info_view.asset_tag, genesis_info_view.meta_hash,
> ```

Hmm, trying to remember why we also wanted to group by version in the first place 🤔

GeorgeTsagk (approved):
Looks good! :ok_hand:

lukegao209:
Hi @guggero  I saw the video about Multisig. 
I have one question :  Alice (tapd) + Bob(tapd) + Third party(tapd)     Multisig.

for David(no tapd with) : 
1. is it possible for alice to send an asset to Bob,and the target is multisig by (Bob,David,Third Party)?
2. can David and Bob transfer the asset by their Multisig?

guggero:
> 1. is it possible for alice to send an asset to Bob,and the target is multisig by (Bob,David,Third Party)?

Yes, definitely. Both on the BTC level (`internal_key` and `tapscript_sibling` of the `NewAddrRequest` message) as well as on the asset level (custom `script_key`). Of course Alice would need to know all the public keys involved, so she can construct the outputs. Alice would then send the proof to Bob as usual and Bob can import it to show the asset in `ListAssets`. I think the only caveat is that the `script_key_is_local` field in the `ListAssets` response would be `false`, because the `tadp` wallet wouldn't be able to spend the asset without externally produced signatures.

> 2. can David and Bob transfer the asset by their Multisig?

Yes, if they use the correct RPCs to create the correct signatures.

lukegao209:
> > 1. is it possible for alice to send an asset to Bob,and the target is multisig by (Bob,David,Third Party)?
> 
> Yes, definitely. Both on the BTC level (`internal_key` and `tapscript_sibling` of the `NewAddrRequest` message) as well as on the asset level (custom `script_key`). Of course Alice would need to know all the public keys involved, so she can construct the outputs. Alice would then send the proof to Bob as usual and Bob can import it to show the asset in `ListAssets`. I think the only caveat is that the `script_key_is_local` field in the `ListAssets` response would be `false`, because the `tadp` wallet wouldn't be able to spend the asset without externally produced signatures.
> 
> > 2. can David and Bob transfer the asset by their Multisig?
> 
> Yes, if they use the correct RPCs to create the correct signatures.

Its a great feature. its possible for btccoin address to receive taproot assets without tapd.

PR #797, opened 2024-02-10
make: Enhance/fix go version linting

dstadulis:
A portability bug in BSD vs GNU sed is remediated in this PR as well as tightening up when Go version validation fails.

- make: simplify calling Go version consistency checks
- make: add comments to describe purpose of the linter-supporting scripts
- make: replace non-portable use of sed with tr
- make: replace error condition's predicate Instead of returning an error if any go version exists other than the stipulated version, require the extracted Go version to be the stipulated version
- make: error reporting logic - Add error reporting logic   Error if false-negative version validation exist: e.g. find command   fails to induce check_go_version - Add validation-failure reporting - Capture output of check_go_version for validation processing
- make: Modify phrasing of error report to communicate violated version and stipulated version Remove exit status to be replaced with FAIL error-message presence

## Tests
Tests to validate true positive and true negative
```sh
echo "Validate help print"
./tools/check-go-version-dockerfile.sh
./tools/check-go-version-yaml.sh
echo "Validate pass condition"
./tools/check-go-version-dockerfile.sh 1.21.4
./tools/check-go-version-yaml.sh 1.21.4
echo "Validate failure condition"
./tools/check-go-version-dockerfile.sh 96
./tools/check-go-version-yaml.sh 96
```

### DIFF
--- a/tools/check-go-version-dockerfile.sh
+++ b/tools/check-go-version-dockerfile.sh
@@ -17,6 +17,9 @@ check_go_version() {
     fi
 }
 
+# Export function to be accessible by subshells e.g. `find -exec`
+export -f check_go_version
+
 # Check if the target Go version argument is provided
 if [ $# -eq 0 ]; then
     echo "Usage: $0 <target_go_version>"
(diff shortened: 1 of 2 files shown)


ffranr:
As far as I'm aware, this PR isn't ready for me to review. So I'm going to remove my review request to keep my review request list short. Please re-request when ready.

PR #800, opened 2024-02-14
itest: universe server harness uses a dedicated LND node

ffranr:
In this commit we ensure that the universe server harness uses a new dedicated LND node. The universe harness previously used the Bob standby LND node. This change allows us to use an in-test tapd node with LND node Bob without conflicting with the universe server tapd node.

This change will be important when passing messages between tapd nodes (e.g. RFQ system) via LND nodes.

This work is necessary to complete: https://github.com/lightninglabs/taproot-assets/issues/766

### DIFF
--- a/itest/integration_test.go
+++ b/itest/integration_test.go
@@ -50,6 +50,13 @@ func TestTaprootAssetsDaemon(t *testing.T) {
 
 	lndHarness.SetupStandbyNodes()
 
+	// Create a new LND node for use with the universe server.
+	t.Log("Starting universe server LND node")
+	uniServerLndHarness := lndHarness.NewNode("uni-server-lnd", nil)
+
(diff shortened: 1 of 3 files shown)


guggero (approved):
LGTM :tada: 

I think it makes sense to use a specific node for the universe.
But I'd still say that for a test where we require a small test network of three nodes it would make sense to spin up three extra `lnd`s, just to make sure we don't "pollute" the long-running nodes with channel state (which might have an impact on other tests).

GeorgeTsagk (approved):
LGTM

Just to make sure, this new harness node is introduced here in order to be used in future itests that require it to be a separate node? Are we going to also update all existing itests to use the `uniServerLndHarness` instead of Bob?

PR #804, opened 2024-02-15
itest: trustless swap based on vPSBT/PSBT sighash flags

### LINKED ISSUE #577: tapgarden/rpc: expose custom sighash flags for vPSBT funding to enable trustless asset swaps

GeorgeTsagk:
## Description

Based on #787 , #779 

Closes #577 

This PR introduces a basic itest that follows the flow of a trustless swap that is possible by correctly manipulating the sighash flags on the asset and bitcoin levels. We create an asymmetry between the vPSBT that describes the asset transfer and the PSBT that anchors it in a bitcoin transaction. The vPSBT only commits to the inputs, allowing anyone to define their own output, while the PSBT commits to the output (which asks for more bitcoin than what is provided in the anchor input) and allows someone to bring their own input to fulfill the transaction.

This way, the only way to "claim" the loose assets on the vPSBT layer is by bringing your own bitcoin on the PSBT layer, which would effectively be the pay-out for the asset holder who created this swap offer.

## Todos

- [x] Add swap redemption part
~~- [ ] Add malicious actor test coverage~~

### DIFF
--- a/itest/addrs_test.go
+++ b/itest/addrs_test.go
@@ -4,9 +4,11 @@ import (
 	"bytes"
 	"context"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/wire"
 	tap "github.com/lightninglabs/taproot-assets"
+	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightninglabs/taproot-assets/internal/test"
 	"github.com/lightninglabs/taproot-assets/proof"
@@ -568,6 +570,99 @@ func importProof(t *harnessTest, dst *tapdHarness, rawFile []byte,
 	importResp, err := dst.ImportProof(ctxb, &tapdevrpc.ImportProofRequest{
 		ProofFile:    rawFile,
 		GenesisPoint: genesisPoint,
+	})
+	require.NoError(t.t, err)
+
+	return importResp
+}
+
+// sendUniProof manually exports a proof from the given source using the
+// universe RPCs and then imports it into the destination node.
+func sendUniProof(t *harnessTest, src, dst *tapdHarness, scriptKey []byte,
+	genInfo *taprpc.GenesisInfo, group *taprpc.AssetGroup,
+	outpoint string) *tapdevrpc.ImportProofResponse {
+
+	ctxb := context.Background()
+	ctxt, cancel := context.WithTimeout(ctxb, defaultWaitTimeout)
+	defer cancel()
+
+	fetchUniProof := func(ctx context.Context,
+		loc proof.Locator) (proof.Blob, error) {
+
+		uniID := universe.Identifier{
+			AssetID: *loc.AssetID,
+		}
+		if loc.GroupKey != nil {
+			uniID.GroupKey = loc.GroupKey
+		}
+
+		rpcUniID, err := tap.MarshalUniID(uniID)
+		require.NoError(t.t, err)
+
+		op := &unirpc.Outpoint{
+			HashStr: loc.OutPoint.Hash.String(),
+			Index:   int32(loc.OutPoint.Index),
+		}
+		scriptKeyBytes := loc.ScriptKey.SerializeCompressed()
+
+		uniProof, err := src.QueryProof(ctx, &unirpc.UniverseKey{
+			Id: rpcUniID,
+			LeafKey: &unirpc.AssetKey{
+				Outpoint: &unirpc.AssetKey_Op{
+					Op: op,
+				},
+				ScriptKey: &unirpc.AssetKey_ScriptKeyBytes{
+					ScriptKeyBytes: scriptKeyBytes,
+				},
+			},
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		return uniProof.AssetLeaf.Proof, nil
+	}
+
+	var assetID asset.ID
+	copy(assetID[:], genInfo.AssetId)
+
+	scriptPubKey, err := btcec.ParsePubKey(scriptKey)
+	require.NoError(t.t, err)
+
+	op, err := wire.NewOutPointFromString(outpoint)
+	require.NoError(t.t, err)
+
+	loc := proof.Locator{
+		AssetID:   &assetID,
+		ScriptKey: *scriptPubKey,
+		OutPoint:  op,
+	}
+
+	if group != nil {
+		groupKey, err := btcec.ParsePubKey(group.TweakedGroupKey)
+		require.NoError(t.t, err)
+
+		loc.GroupKey = groupKey
+	}
+
+	var proofFile *proof.File
+	err = wait.NoError(func() error {
+		proofFile, err = proof.FetchProofProvenance(
+			ctxt, nil, loc, fetchUniProof,
+		)
+		return err
+	}, defaultWaitTimeout)
+	require.NoError(t.t, err)
+
+	var buf bytes.Buffer
+	err = proofFile.Encode(&buf)
+	require.NoError(t.t, err)
+
+	t.Logf("Importing proof %x", buf.Bytes())
+
+	importResp, err := dst.ImportProof(ctxb, &tapdevrpc.ImportProofRequest{
+		ProofFile:    buf.Bytes(),
+		GenesisPoint: genInfo.GenesisPoint,
 	})
 	require.NoError(t.t, err)
 
--- a/itest/psbt_test.go
+++ b/itest/psbt_test.go
@@ -4,16 +4,22 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"fmt"
+	"net/url"
 	"testing"
 
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/btcutil/psbt"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/davecgh/go-spew/spew"
 	"github.com/lightninglabs/taproot-assets/address"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/commitment"
 	"github.com/lightninglabs/taproot-assets/internal/test"
+	"github.com/lightninglabs/taproot-assets/proof"
 	"github.com/lightninglabs/taproot-assets/tappsbt"
 	"github.com/lightninglabs/taproot-assets/taprpc"
 	wrpc "github.com/lightninglabs/taproot-assets/taprpc/assetwalletrpc"
@@ -22,7 +28,9 @@ import (
 	"github.com/lightninglabs/taproot-assets/tapsend"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/keychain"
+	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/lnrpc/walletrpc"
+	"github.com/lightningnetwork/lnd/lntest"
 	"github.com/lightningnetwork/lnd/lntest/node"
 	"github.com/lightningnetwork/lnd/lntest/wait"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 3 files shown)


itest/psbt_test.go
```diff
@@ -1608,6 +1619,436 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+	require.NoError(t.t, err)
+
+	// Now we retrieve the bitcoin PSBT from the response.
+	btcpsbt, err = psbt.NewFromRawBytes(
+		bytes.NewReader(resp.AnchorPsbt), false,
+	)
+	require.NoError(t.t, err)
+
+	// The first input is the anchor that carries Alice's assets. This input
+	// will only commit to itself and Alice's output at same index (0), which
```
ffranr, line 1766: Maybe: "This input is committed to the first output" rather than "This input
	// will only commit to itself and Alice's output"
	
I think that we "commit an input to something" is correct rather than "the input commits itself"? I'm not sure how "commit" is commonly used in this context.

itest/psbt_test.go
```diff
@@ -1608,6 +1619,436 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+	)
+	btcpsbt.UnsignedTx.TxIn = append(
+		btcpsbt.UnsignedTx.TxIn[:1], btcpsbt.UnsignedTx.TxIn[2:]...,
+	)
+
+	// Let's get rid of the change output that we no longer need.
+	btcpsbt.Outputs = btcpsbt.Outputs[:2]
+	btcpsbt.UnsignedTx.TxOut = btcpsbt.UnsignedTx.TxOut[:2]
+
+	t.Logf("Funded BTC PSBT: %v", spew.Sdump(btcpsbt))
```
ffranr, line 1788: Is this still a "funded" PSBT? I think we've just stripped that funding. I think at this point it's funded with the taproot-asset anchoring only, is that correct?

itest/psbt_test.go
```diff
@@ -1608,6 +1619,436 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+	bobVOut.AnchorOutputBip32Derivation = nil
+	bobVOut.AnchorOutputTaprootBip32Derivation = nil
+	bobVOut.SetAnchorInternalKey(
+		bobAnchorInternalKey, harnessNetParams.HDCoinType,
+	)
+	deliveryAddrStr := fmt.Sprintf(
+		"%s://%s", proof.UniverseRpcCourierType,
+		t.universeServer.ListenAddr,
+	)
+	deliveryAddr, err := url.Parse(deliveryAddrStr)
```
ffranr, line 1846: Would be good to have this as a method on the universe server.

itest/psbt_test.go
```diff
@@ -1608,6 +1619,436 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+		bobVOut.AnchorOutputTaprootBip32Derivation
+
+	// Before Bob tidies up the output commitments he keeps a backup of the
+	// transfer witnesses. This is where Alice's SIGHASH_NONE signature
+	// lies.
+	witnessBackup := bobVPsbt.Outputs[0].Asset.PrevWitnesses
+
+	// Bob tidies up the outputs.
+	err = tapsend.PrepareOutputAssets(ctxb, bobVPsbt)
+	require.NoError(t.t, err)
```
ffranr, line 1867: What does tidying up mean here?

itest/psbt_test.go
```diff
@@ -1608,6 +1619,436 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+		},
+	)
+	require.NoError(t.t, err)
+
+	// Bob should sign exactly 1 input.
+	require.Len(t.t, signResp.SignedInputs, 1)
+
+	// The signed input should be at index 1, which is the bitcoin that Bob
+	// is bringing into the PSBT.
+	require.Equal(t.t, uint32(1), signResp.SignedInputs[0])
```
ffranr, line 1922: Might be worth naming these indexes, something like: `bobPsbtInFundingIdx := uint32(1)`

ffranr:
As requested, I've only reviewed the itest in the last commit.

itest/psbt_test.go
```diff
@@ -1608,6 +1619,436 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+	require.NoError(t.t, err)
+
+	// Alice signs the vPSBT.
+	signedResp, err := alice.SignVirtualPsbt(
+		ctxb, &wrpc.SignVirtualPsbtRequest{
+			FundedPsbt: fundedPsbtBytes,
+		},
+	)
+	require.NoError(t.t, err)
+	require.Contains(t.t, signedResp.SignedInputs, uint32(0))
```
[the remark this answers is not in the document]
GeorgeTsagk: `txscript.SigHashNone` is used so that we can sign the input and allow its spending while also committing to none of the outputs, allowing somebody (Bob) later to edit the outputs.

If we isolate on the vPSBT level, Alice is basically saying "i'm signing these assets, and anyone can define the output"

itest/psbt_test.go
```diff
@@ -1608,6 +1619,436 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+	require.NoError(t.t, err)
+
+	require.Len(t.t, vPkt.Inputs, 1)
+	require.Len(t.t, vPkt.Outputs, 1)
+
+	t.Logf("Funded vPSBT: %v", toBase64(t.t, vPkt))
+
+	// On the vPSBT level, which describes the assets transfer, we do not
+	// commit to any outputs.
+	vPkt.Inputs[0].SighashType = txscript.SigHashNone
```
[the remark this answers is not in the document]
GeorgeTsagk: `txscript.SigHashNone` commits to all inputs and none of the outputs

itest/psbt_test.go
```diff
@@ -1608,6 +1619,436 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+	derivation, trDerivation := getAddressBip32Derivation(
+		t.t, addrResp.Address, t.lndHarness.Alice,
+	)
+	btcpsbt.Outputs[0].Bip32Derivation = []*psbt.Bip32Derivation{
+		derivation,
+	}
+	btcpsbt.Outputs[0].TaprootBip32Derivation = []*psbt.TaprootBip32Derivation{
+		trDerivation,
+	}
+	btcpsbt.Outputs[0].TaprootInternalKey = trDerivation.XOnlyPubKey
```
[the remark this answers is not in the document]
GeorgeTsagk: The derivation/key related additions that follow these 2 lines are basically required for `TxOut[0]` to actually pay to `alicePkScript`. I decided to keep the first 2 lines together as they indeed express alice's "terms".

Perhaps worth adding a comment as to what these extra lines do.

itest/psbt_test.go
```diff
@@ -1608,6 +1619,436 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+
+	// Now we need to commit the vPSBT and PSBT, creating all the related
+	// proofs for this transfer to be valid.
+	resp, err := alice.CommitVirtualPsbts(
+		ctxb, &wrpc.CommitVirtualPsbtsRequest{
+			VirtualPsbts: [][]byte{signedResp.SignedPsbt},
+			AnchorPsbt:   b.Bytes(),
+			AnchorChangeOutput: &wrpc.CommitVirtualPsbtsRequest_Add{
+				Add: true,
+			},
```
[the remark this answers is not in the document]
GeorgeTsagk: Yes, this is essentially asking for a btc change output to be created.

Previously we checked for 2 outputs because one was Alice's btc "reward" (index `0`) and one was the anchor of the assets

Now that Bob is bringing in his bitcoin, he might want to add an explicit change at index `2`, since after paying alice at index`0` there may be change from his inputs. He could "merge" the btc change with the assets anchor. I believe keeping them separate is more clean and simple

itest/psbt_test.go
GeorgeTsagk: Hmm not sure about the exact phrasing here. But the comment should definitely capture the fact that the input actually commits to "itself", as it is valid only if it's present in the final tx and also at the same tx as the output it was signing.

itest/psbt_test.go
GeorgeTsagk: Right, will rename to something like "Alice BTC PSBT"

itest/psbt_test.go
GeorgeTsagk: Essentially prepares the asset outputs

itest/psbt_test.go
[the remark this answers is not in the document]
ffranr: Ok thanks. Does that mean that if you set `vPkt.Inputs[0].SighashType` then it applies to all `vPkt.Inputs`? I think we should add a comment mentioning how this line commits to all inputs.

itest/psbt_test.go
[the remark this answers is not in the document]
ffranr: Thanks for the explanation!

itest/psbt_test.go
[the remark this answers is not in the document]
ffranr: Thank you for the explanation.

Given what you said, I think the code comment above is incorrect, it says:
```
	// This bitcoin PSBT should have 1 input, which is the anchor of Alice's
	// assets, and 2 outputs that correspond to Alice's anchor that carries the
	// transfered assets, and one output that carries the bitcoin change.
```
but I think it should say:
```
	// This bitcoin PSBT should have 1 input, which is the anchor of Alice's
	// assets, and 2 outputs that correspond to the BTC payment that Alice
        // will claim (index 0), and Alice's anchor that carries the transferred
        // assets (index 1).
```

itest/psbt_test.go
ffranr: I understand it this way: The input itself does not commit to anything, as it is not a signature. Rather, it is the signature generated by Alice that commits to the input.

I think you're also saying that the signature generated by Alice actually commits to more than just the input. It commits to the input and also the output at index 0. And that the final tx is only valid if it includes this input and output.

ffranr (commented):
I've replied to the ongoing review discussions. They are all close to being resolved.

itest/psbt_test.go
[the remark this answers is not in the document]
GeorgeTsagk: > Does that mean that if you set vPkt.Inputs[0].SighashType then it applies to all vPkt.Inputs

Every different `vPkt.Inputs[x]` can carry its own sighash flag. For example, we could have a `vPkt` with 2 inputs where input `0` uses SigHashAll and input `1` uses SigHashNone.

> I think we should add a comment mentioning how this line commits to all inputs.

In the example above the `vPkt` has only one input. Perhaps I'm not getting what you mean?

itest/psbt_test.go
GeorgeTsagk: > I understand it this way: The input itself does not commit to anything, as it is not a signature. Rather, it is the signature generated by Alice that commits to the input.

Right, my phrasing is wrong there, it's the signature that commits not the input.

> I think you're also saying that the signature generated by Alice actually commits to more than just the input. It commits to the input and also the output at index 0. And that the final tx is only valid if it includes this input and output.

Yeah this input's signature will commit to that input and the output at the same index (`0`). The fact that this input & output need to be in the tx is more like stating the obvious, could just strip that part

bhandras:
Is this branch and the added test supposed to work already or needs changes from other branches? Tried running it but fails on my end (checked out this branch).

guggero:
I might have broken things in the base branch (`commit-virtual-psbts`), so a rebase might be needed.

guggero:
> I approve of the itest commit (last commit). Maybe we need to change the base branch such that this PR only contains that one commit?

I already did that, just needs a rebase. But to avoid rebasing too often, I'd say let's just park this one until the dependent PR is merged.

guggero (approved):
Needs a rebase, otherwise LGTM :tada:

itest/psbt_test.go
```diff
@@ -1610,6 +1618,352 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+	// Bob should have signed the input at the expected index.
+	require.Equal(t.t, bobInputIdx, signResp.SignedInputs[0])
+
+	require.NoError(t.t, finalPsbt.SanityCheck())
+
+	signedPkt := finalizePacket(t.t, t.lndHarness.Bob, finalPsbt)
+
+	require.True(t.t, signedPkt.IsComplete())
+
+	logResp := logAndPublish(
```
ffranr, line 1941: nit: too many empty lines here perhaps

itest/psbt_test.go
```diff
@@ -1610,6 +1618,352 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+		},
+	)
+	require.NoError(t.t, err)
+
+	bobVPsbt, err = tappsbt.Decode(resp.VirtualPsbts[0])
+	require.NoError(t.t, err)
+
+	// Since Bob brings in a new input to the bitcoin transaction, he needs
+	// to sign it. We do not care about the sighash flag here, that can be
+	// the default, as we're not further editing the bitcoin transaction.
```
ffranr, line 1909: nit: "as we're not further editing the bitcoin transaction" -> "as we will not edit the bitcoin transaction further"

itest/psbt_test.go
```diff
@@ -1610,6 +1618,352 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+		logResp.Transfer.Outputs[0].Anchor.Outpoint,
+	)
+
+	bobAssets, err := bob.ListAssets(ctxb, &taprpc.ListAssetRequest{})
+	require.NoError(t.t, err)
+
+	// Verify that Bob now holds the asset.
+	require.Len(t.t, bobAssets.Assets, 1)
+	require.Equal(t.t, bobAssets.Assets[0].Amount, numUnits)
+}
```
ffranr, line 1965: We've got assertion functions to ensure that the transfer was complete. Maybe consider using https://github.com/lightninglabs/taproot-assets/blob/98634ca6d552d8eec4bb16491dc288f869f529f4/itest/assertions.go#L876-L876

ffranr (approved):
That was clear and made sense to me. Nice work!

itest/psbt_test.go
```diff
@@ -1610,6 +1618,352 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+		chainParams = &address.RegressionNetTap
+		assetID     asset.ID
+	)
+	copy(assetID[:], genInfo.AssetId)
+
+	// Now Alice will derive the script and anchor internal keys that will
+	// be used to bootstrap an interactive full send of her assets.
+	aliceScriptKey, aliceAnchorInternalKey := DeriveKeys(t.t, alice)
+	vPkt := tappsbt.ForInteractiveSend(
+		assetID, numUnits, aliceScriptKey, 1, aliceAnchorInternalKey,
```
Roasbeef, line 1650: For clarify we could maybe have `aliceScriptKey` just be a blank pub key? In that later on Bob will swap this out in order to claim this himself.

itest/psbt_test.go
```diff
@@ -1610,6 +1618,352 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+
+	var b bytes.Buffer
+	err = btcpsbt.Serialize(&b)
+	require.NoError(t.t, err)
+
+	// Now we need to commit the vPSBT and PSBT, creating all the related
+	// proofs for this transfer to be valid.
+	resp, err := alice.CommitVirtualPsbts(
+		ctxb, &wrpc.CommitVirtualPsbtsRequest{
+			VirtualPsbts: [][]byte{signedResp.SignedPsbt},
```
Roasbeef, line 1746: Hmm, not sure if I'm following this properly, but it appears that we're anchoring the vPSBT into the transaction above that has 1-input-1-output. Instead, Bob should be the one anchoring the vPSBT (with the blank signed check from Alice) into an output that _he_ solely controls.

Roasbeef: Actually scratch the above, I see an output index of `1` is provided. However how can we anchor this ourselves if we don't yet know Bob's public key? In other words, Bob should be the one that takes the vPSBT, and anchors it. Alice's role is to give the signed PSBT that spends her input with an asset, and asks for X amt of BTC by providing less total inputs than the output amt.

Roasbeef: Ok scratch all that, I see this was just a round about way of doing this, and we'll strip things down to just 1 input 1 output. 

Leaving these comments in place just to demonstrate my thought process while reviewing this.

itest/psbt_test.go
```diff
@@ -1610,6 +1618,352 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
+	bobVOut.SetAnchorInternalKey(
+		bobAnchorInternalKey, harnessNetParams.HDCoinType,
+	)
+	deliveryAddrStr := fmt.Sprintf(
+		"%s://%s", proof.UniverseRpcCourierType,
+		t.universeServer.ListenAddr,
+	)
+	deliveryAddr, err := url.Parse(deliveryAddrStr)
+	require.NoError(t.t, err)
+	bobVPsbt.Outputs[0].ProofDeliveryAddress = deliveryAddr
```
Roasbeef, line 1847: Given this is present, why do we need to manually insert proofs above?

Roasbeef (approved):
Nice work! 

I think the flow can be simplified a bit (eg: why do we need to do the fund at the start just to strip it away, instead of just using alice's input), but this is working fine e2e as is. 

Main comment is that we should try to use the `InsertProof` method on the Universe instead of using the semi-deprecated `ImportProofs` call. Or do we use that method since the `InsertProof` will only be picked up if a receive address is known, and in this case Bob never makes one as this is a more adhoc swap?

itest/addrs_test.go
```diff
@@ -574,6 +576,99 @@ func importProof(t *harnessTest, dst *tapdHarness, rawFile []byte,
...
+	}, defaultWaitTimeout)
+	require.NoError(t.t, err)
+
+	var buf bytes.Buffer
+	err = proofFile.Encode(&buf)
+	require.NoError(t.t, err)
+
+	t.Logf("Importing proof %x", buf.Bytes())
+
+	importResp, err := dst.ImportProof(ctxb, &tapdevrpc.ImportProofRequest{
```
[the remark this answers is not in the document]
guggero: The thing is, if the receiver doesn't expect a transfer (e.g. there is no address that corresponds to the output), then `InsertProof` doesn't really do anything other than inserting the proof into the universe. It's not being picked up as an asset that belongs to the node. That's why I asked offline what mental model you had in mind for interactive transfers. Do we need a call that tells a node: "Hey, that proof you have there in your universe, that's actually an asset you can spend"?

itest/psbt_test.go
GeorgeTsagk: Tried using this but didn't seem to work? Will keep current checks for now

itest/psbt_test.go
GeorgeTsagk: We'll need to derive the anchor key too so I guess this is less lines overall?

Will rename the variable to describe intent

Issue #616, opened 2023-10-24
Refactor send logic


dstadulis:
### Background

Now that splitting assets, burning / tombstones features have been added, the daemon would benefit if sending logic was refactored.  There are complexities in how passive assets interact with these features and a holistic refactor will ensure security, and readability.

dstadulis:
- [ ] Asset send/receive procedure uploading / requesting proofs to trigger on transaction confirmation.

Roasbeef:
I think the other part of this is: making sure we enumerate all the rules properly in the BIP.

PR #799, opened 2024-02-13
[preparation 1/2]: refactor to simplify send logic, prepare for new RPCs

guggero:
This is a preparatory PR that does some of the refactoring required for https://github.com/lightninglabs/taproot-assets/pull/787. We also add two new fields to the vPSBT outputs: `ProofSuffix` and `ProofDeliveryAddress` which will be required to ship this missing information between different participants in a fully RPC driven send.

PR #805, opened 2024-02-19
[psbt saga part 2/4]: preparation: refactor send logic, make passive assets independent

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/799.

Fixes https://github.com/lightninglabs/taproot-assets/issues/616.

This is a preparatory PR that does some additional refactoring required for https://github.com/lightninglabs/taproot-assets/pull/787.

The preparations in this PR are mostly focused on making the active and passive virtual transactions independent from each other. And we also make sure we have everything we need to  store a transfer that happened outside of the freighter (through RPCs) in the database as a transfer entry.


guggero:
Okay, I was able to confirm this has all the functionality we need for https://github.com/lightninglabs/taproot-assets/pull/787 and https://github.com/lightninglabs/taproot-assets/pull/796.
I only had to add a single commit along the other 20: 665acfcd

Now I only need to fix all the tests and add a bunch of more unit and integration test coverage.
But I still think the PR is ready for a first look, it shouldn't change apart from test related findings.

tapfreighter/wallet.go
```diff
@@ -1231,28 +1211,139 @@ func removeActiveCommitments(inputCommitment *commitment.TapCommitment,
...
+// passive assets to. If no suitable output is found, a new anchor output is
+// created.
+func (f *AssetWallet) determinePassiveAssetAnchorOutput(ctx context.Context,
+	activePackets []*tappsbt.VPacket) (*keychain.KeyDescriptor, uint32,
+	error) {
+
+	var (
+		maxAnchorOutputIndex uint32
+		candidates           []*tappsbt.VOutput
+		candidateDescriptos  []*keychain.KeyDescriptor
```
GeorgeTsagk, line 1243:
```suggestion
		candidateDescriptors  []*keychain.KeyDescriptor
```

tapfreighter/wallet.go
```diff
@@ -1231,28 +1211,139 @@ func removeActiveCommitments(inputCommitment *commitment.TapCommitment,
...
+			anchorKeyDesc, err := vOut.AnchorKeyToDesc()
+			if err != nil {
+				// We can't determine the key descriptor for
+				// this output, so we'll skip it as it very
+				// likely doesn't belong to us then.
+				continue
+			}
+
+			// Ignore any anchor outputs that are not local to us.
+			if !f.cfg.KeyRing.IsLocalKey(ctx, anchorKeyDesc) {
```
GeorgeTsagk, line 1258: is this essentially checking if the anchor belongs to us?

GeorgeTsagk (approved):
LGTM + tACK :sparkles: 
ty for the clean & consistent commit messages :100:

GeorgeTsagk:
Not sure if this is an unrelated flake or introduced by this diff: [itest postgres](https://github.com/lightninglabs/taproot-assets/actions/runs/8023093944/job/21918863242?pr=805#step:4:3438)

tapfreighter/wallet.go
guggero: Correct! Since we'll want to re-anchor any passive assets back to our wallet.

guggero:
> Not sure if this is an unrelated flake or introduced by this diff: [itest postgres](https://github.com/lightninglabs/taproot-assets/actions/runs/8023093944/job/21918863242?pr=805#step:4:3438)

I ran this about 100 times locally and wasn't able to reproduce...
It was a deadlock in Postgres, possibly due to many proofs being imported at the same time. If we ever run into this again, I'd suggest adding a mutex around the proof import to avoid parallel updates to the same tables producing a deadlock.

tapfreighter/wallet.go
```diff
@@ -156,10 +156,6 @@ type AnchorVTxnsParams struct {
 	// anchored by the anchor transaction.
 	VPkts []*tappsbt.VPacket
 
-	// InputCommitments is a map from virtual package input index to its
-	// associated Taproot Assets commitment.
-	InputCommitments tappsbt.InputCommitments
```
Roasbeef, line 189: No longer needed as all this info is now in the `Vpkts` passed in?

tapfreighter/chain_porter.go
```diff
@@ -931,7 +931,6 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
 			ctx, &AnchorVTxnsParams{
 				FeeRate:            feeRate,
 				VPkts:              []*tappsbt.VPacket{vPacket},
-				InputCommitments:   currentPkg.InputCommitments,
```
Roasbeef, line 934: Re the commit message: IIRC we ended up special casing the passive assets so we wouldn't have to force a dev/caller to handle all that book keeping logic. With this new approach, will the freighter automatically create the relevant vPSBTs for the passive assets, or is that something the caller needs to be burdened with?

tapsend/send.go
```diff
@@ -975,65 +973,29 @@ func commitPacket(vPkt *tappsbt.VPacket,
...
-	// Calculate the number of outputs we need for our template TX.
-	maxOutputIndex := uint32(len(outputs))
+func CreateAnchorTx(vPackets []*tappsbt.VPacket) (*psbt.Packet, error) {
+	// We locate the highest anchor output index in all virtual packets to
+	// create a template TX with the correct number of outputs.
+	var maxOutputIndex uint32
+	for _, vPkt := range vPackets {
+		// Sanity check the output indexes provided by the sender. There
+		// must be at least one output.
+		if len(vPkt.Outputs) < 1 {
```
Roasbeef, line 968: Could be written as just `len(pkts) == 0` for clarity

tapfreighter/parcel.go
```diff
@@ -348,150 +348,201 @@ func ConvertToTransfer(currentHeight uint32, vPkt *tappsbt.VPacket,
...
+
+	return parcel, nil
+}
+
+// transferInput creates a TransferInput from a virtual input and the anchor
+// packet.
+func transferInput(anchorPacket *psbt.Packet,
+	vIn *tappsbt.VInput) (*TransferInput, error) {
+
+	// We don't know the actual outpoint the input is spending, so we need
```
Roasbeef, line 418: Isn't this just in the prev ID section of the asset included in the `vIn`?

Roasbeef: Or I guess it just might not be populated at this point?

tapdb/assets_store.go
```diff
@@ -2032,11 +2032,22 @@ func (a *AssetStore) LogPendingParcel(ctx context.Context,
 			}
 		}
 
+		// Then the passive assets.
+		if len(spend.PassiveAssets) > 0 {
```
Roasbeef, line 2036: Should also defensively check `PassiveAssetsAnchor` for nil-ness.

tapfreighter/wallet.go
```diff
@@ -1231,28 +1211,139 @@ func removeActiveCommitments(inputCommitment *commitment.TapCommitment,
...
 			continue
 		}
 
-		err := removeAsset(assetCommitment, vIn.Asset(), key)
-		if err != nil {
-			return nil, fmt.Errorf("unable to "+
-				"delete asset: %w", err)
+		fromKey := vOut.Asset.PrevWitnesses[0].PrevID.ScriptKey
+		toKey := asset.ToSerialized(vOut.ScriptKey.PubKey)
+		if fromKey != toKey {
```
Roasbeef, line 1305: So this is trying to rule out the case of anchoring to a self-send w/ addr re-use?

tapfreighter/wallet.go
```diff
@@ -1231,28 +1211,139 @@ func removeActiveCommitments(inputCommitment *commitment.TapCommitment,
...
+	// output for them.
+	anchorOutDesc, anchorOutIdx, err := f.determinePassiveAssetAnchorOutput(
+		ctx, activePackets,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("unable to determine passive asset "+
+			"anchor output: %w", err)
+	}
+
+	// We didn't find an output, let's create one.
```
Roasbeef, line 1345: Stray comment?

proof/verifier.go
```diff
@@ -425,10 +425,12 @@ func (p *Proof) Verify(ctx context.Context, prev *AssetSnapshot,
...
 	// merkle proof within a block in the chain.
 	if prev != nil && p.PrevOut != prev.OutPoint {
-		return nil, commitment.ErrInvalidTaprootProof // TODO
+		return nil, fmt.Errorf("%w: prev output mismatch",
+			commitment.ErrInvalidTaprootProof) // TODO
 	}
 	if !txSpendsPrevOut(&p.AnchorTx, &p.PrevOut) {
-		return nil, commitment.ErrInvalidTaprootProof // TODO
+		return nil, fmt.Errorf("%w: doesn't spend prev output",
+			commitment.ErrInvalidTaprootProof) // TODO
```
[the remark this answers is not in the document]
guggero: Wasn't sure what exactly the TODO referred to (the error being returned or the validation itself). But removed the TODO now.

tapfreighter/wallet.go
guggero: Not really. It wasn't really needed anymore once we started carrying along the passive assets explicitly. Tried to explain in the commit message:
> We always started with the input commitment, cloned it, removed the
active asset from the commitment, then added the output asset. And
finally we also committed the passive assets to that output.
But because the input commitment minus the active asset is equal to the
passive assets, we can simplify this whole procedure if we just assume
that the passive assets are in the list of the given virtual packets.

tapfreighter/chain_porter.go
guggero: Yes, they will still be created automatically. If you use the `FundVirtualPsbt -> AnchorVirtualPsbts` flow you don't have to deal with passive assets at all. If you use the `FundVirtualPsbt -> CommitVirtualPsbts -> PublishAndLogTransfer` flow (implemented in part 3 of the saga), you'll need to carry along the passive assets and sign them manually (since they might also be multisig or custom spend scripts so you need to do manual signing anyway) and feed them into the calls.

tapfreighter/parcel.go
guggero: Hmm, you're right. This should just be the prevID and it should be set at this point. Not sure why it was implemented this way initially.

tapfreighter/wallet.go
guggero: Yes... Fixed.

tapfreighter/wallet.go
guggero: Yes. Basically we're trying to find an "active" transfer that goes to an anchor internal key we own, so we can commit the passive assets there. And the only two heuristics for finding out if a transfer is "active" is that either it's a split or it goes to a new script key (full value send to yourself).
So worst case is we create a new anchor output even though we don't necessarily need to.

guggero: Although, thinking back to why I created this logic: The goal was to ultimately remove the conceptual separation of active and passive assets in the RPCs and just identify them based on the heuristic described above. But since we currently still have the distinction between active/passive in the RPCs, maybe we don't need this more complicated logic and just pick any anchor output that goes back to our wallet (or create one if there is none).

Roasbeef (approved):
LGTM 🎊

PR #812, opened 2024-02-23
Linter checks for `fmt.Errorf` error wrap misuse

ffranr:
This PR enables a linter which checks for `fmt.Errorf` error wrap misuse.

It disallows:
```
fmt.Errorf("%v", err)
```

And allows:
```
fmt.Errorf("%w", err)
```


proof/verifier.go
```diff
@@ -356,7 +356,7 @@ func (p *Proof) verifyGenesisGroupKey(groupVerifier GroupVerifier) error {
 	groupKey := p.Asset.GroupKey.GroupPubKey
 	err := groupVerifier(&groupKey)
 	if err != nil {
-		return fmt.Errorf("%w: %v", ErrGroupKeyUnknown, err)
+		return fmt.Errorf("%w: %w", ErrGroupKeyUnknown, err)
```
guggero, line 359: Does the linter complain about these? Because wrapping both IMO isn't what we want here. We want the caller to see it's an `ErrGroupKeyUnknown`, the `err` is just in the message to give more context. Same with the one below, i don't think we should wrap the `err`.

guggero (changes requested):
Very nice, thanks a lot!
Not sure about the last commit though.

proof/verifier.go
ffranr: The linter does complain about this. I'll see if I can get around this with some string formatting.

guggero (approved):

jharveyb (approved):
Looks good! Diff larger than expected (in a good way).

PR #814, opened 2024-02-27
Refactor the RFQ service to handle payment init via asset sale

ffranr:
Towards fixing https://github.com/lightninglabs/taproot-assets/issues/818

This PR refactors the RFQ system to make adding asset sale support easier.

### DIFF
--- a/rfq/negotiator.go
+++ b/rfq/negotiator.go
@@ -74,7 +74,7 @@ func NewNegotiator(cfg NegotiatorCfg) (*Negotiator, error) {
 // an appropriate outgoing response message which should be sent to the peer.
 func (n *Negotiator) queryBidFromPriceOracle(peer route.Vertex,
 	assetId *asset.ID, assetGroupKey *btcec.PublicKey,
-	assetAmount uint64) (rfqmsg.OutgoingMsg, error) {
+	assetAmount uint64) (lnwire.MilliSatoshi, uint64, error) {
 
 	ctx, cancel := n.WithCtxQuitNoTimeout()
 	defer cancel()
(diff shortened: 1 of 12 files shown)


rfq/negotiator.go
```diff
@@ -157,66 +157,53 @@ func (n *Negotiator) HandleOutgoingBuyOrder(buyOrder BuyOrder) error {
...
-	// the asking price is nil, then we will return a quote reject
-	// message indicating that the price oracle did not specify an error.
+	// By this point, the price oracle did not return an error. However, if
+	// the asking price is nil, then we will return early.
 	if oracleResponse.AskPrice == nil {
-		rejectErr := rfqmsg.ErrPriceOracleUnspecifiedError
-
-		reject := rfqmsg.NewReject(request, rejectErr)
-		return reject, nil
+		return nil, 0, nil
```
guggero, line 186: I think maybe the price should be a `fn.Option` then? Returning `nil` alongside a `nil` error is exactly the pattern we want to avoid at all cost I think.

guggero: Or alternatively create a specific error type to catch at the call site.

rfq/negotiator.go
```diff
@@ -248,31 +235,65 @@ func (n *Negotiator) HandleIncomingBuyRequest(
...
-		outgoingMsgResponse, err := n.queryAskFromPriceOracle(request)
+		askPrice, askExpiry, err := n.queryAskFromPriceOracle(
+			nil, request.AssetID, request.AssetGroupKey,
+			request.AssetAmount, &request.BidPrice,
+		)
 		if err != nil {
-			err = fmt.Errorf("negotiator failed to handle price "+
-				"oracle ask price response: %w", err)
+			// Send a reject message to the peer.
+			var msg rfqmsg.OutgoingMsg = rfqmsg.NewReject(
```
guggero, line 266: nit: can just be `msg := rfqmsg.NewReject(`? Same further below.

rfq/oracle.go
```diff
@@ -19,6 +20,19 @@ type OracleError struct {
 	Msg string
 }
 
+// String returns a human-readable string representation of the error.
+func (o *OracleError) String() string {
```
guggero, line 24: Doesn't a custom error need a `Error()` method that returns a string in order to satisfy the `error` interface? Maybe we should do that instead?

guggero (approved):
Nice, LGTM :tada:

rfq/negotiator.go
ffranr: I don't think returning a price of `nil` or `fn.None` in combination of `err == nil` should ever be correct here. I think we need a `Result` type, see: https://github.com/lightninglabs/taproot-assets/issues/832 What do you think?

guggero: I'm not sure I understand. If the combination of `fn.None` and `err == nil` should never be correct, why are we returning it then? Wouldn't the "never be correct" indicate it should actually be an error type?
Using the `Result` type is just a more formal contract of what we already implicitly want to do in Golang (never `return nil, nil`), just with new syntax, which IMO breaks a lot of existing visual pattern we already got used to over the years.

ffranr: > I'm not sure I understand. If the combination of fn.None and err == nil should never be correct, why are we returning it then? Wouldn't the "never be correct" indicate it should actually be an error type?

Ah, you're right, I wasn't thinking clearly on this. I'll return an error type instead of `return nil, 0, nil`.

dstadulis:
@GeorgeTsagk

GeorgeTsagk (approved):
Looking good! :heavy_check_mark:

PR #815, opened 2024-02-28
[psbt saga 5/4]: Multisig demo with 2-of-3 multisig in tapleaves

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/796.

This PR adds an integration test that demonstrates how multi signatures can be used on both the BTC and asset level.
In the demo we use the following Tapscript to encumber the BTC level anchor output:
```
<key_1> OP_CHECKSIG <key_2> OP_CHECKSIGADD 2 OP_EQUAL
```
This is basically a 2-of-2 multi signature. To make sure the key spend path cannot be taken, we use the NUMS public key as the internal key.

On the asset level, we use a script key that is similarly encumbered. But instead of a single leaf we use three different leaves with a 2-of-2 MuSig2 each:
 - Leaf 1: 2-of-2 MuSig2 between Alice and Bob
 - Leaf 2: 2-of-2 MuSig2 between Alice and a third party
 - Leaf 3: 2-of-2 MuSig2 between Bob and a third party

This means that the 1000 units of assets sent to the constructed TAP address will be co-owned by both `tapd` instances, on the asset and BTC level.

We then spend from that address again (using Alice and Bob's keys on the asset level), just to demonstrate how the witnesses for such transfers would be created.


guggero:
I've folded these changes into https://github.com/lightninglabs/taproot-assets/pull/796, so this PR is no longer necessary.

PR #817, opened 2024-02-28
multi: add linters for switch exhaustiveness, if-else chains, and err handling

jharveyb:
Inspired by #812. These are some lints that felt useful for keeping style consistent, as well as how we use `switch`.

I've added fixes for a few small issues, and lint runtime (both cold and hot module cache) on my machine increased from 13.5 to ~16 seconds.

I'm not sure which linters could be used to replicate the gopls analyzers I'm using locally, but those fixes were less important IMO.

### DIFF
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -54,20 +54,9 @@ func testBasicSendUnidirectional(t *harnessTest) {
 
 		broadcastState := tapfreighter.SendStateBroadcast.String()
 		targetEventSelector := func(event *taprpc.SendAssetEvent) bool {
-			switch eventTyped := event.Event.(type) {
-			case *taprpc.SendAssetEvent_ExecuteSendStateEvent:
-				ev := eventTyped.ExecuteSendStateEvent
-
-				// Log send state execution.
-				timestamp := time.UnixMicro(ev.Timestamp)
-				t.Logf("Executing send state (%v): %v",
-					timestamp.Format(time.RFC3339Nano),
-					ev.SendState)
-
-				return ev.SendState == broadcastState
-			}
-
-			return false
+			return AssertSendEventExecuteSendState(
+				t, event, broadcastState,
+			)
 		}
 
 		timeout := 2 * defaultProofTransferReceiverAckTimeout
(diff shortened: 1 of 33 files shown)


itest/send_test.go
```diff
@@ -65,6 +65,7 @@ func testBasicSendUnidirectional(t *harnessTest) {
 					ev.SendState)
 
 				return ev.SendState == broadcastState
+			default:
```
guggero, line 68: Should these default cases trigger a test failure? Sounds like there shouldn't be any other events here?

tapfreighter/chain_porter.go
```diff
@@ -670,7 +666,8 @@ func (p *ChainPorter) transferReceiverProof(pkg *sendPackage) error {
 			return nil
 		}
 		if err != nil {
-			return fmt.Errorf("error delivering proof: %w", err)
```
guggero, line 673: Nice catch! Too bad this couldn't be caught by any static linter. Because GoLand didn't catch this, even though I think under the hood it also runs gopls...

guggero (approved):
Very nice, LGTM :tada:

tapfreighter/chain_porter.go
jharveyb: Agree - it just got turned on by default for gopls 0.15 series, idk if adding a gopls run to CI would make sense.

itest/send_test.go
jharveyb: Good point, I think type switches are a special case where the `singleCaseSwitch` lint is not triggered, and a type assertion is not suggested.

For these types, a fitting type assertion is actually generated from proto compilation; switched to those.

proof/courier.go
```diff
@@ -1162,9 +1162,6 @@ func (c *UniverseRpcCourier) ReceiveProof(ctx context.Context,
 		receiveFunc := func() error {
 			// Retrieve proof from courier.
 			resp, err := c.client.QueryProof(ctx, &universeKey)
-			if err != nil {
```
Roasbeef, line 1165: kek

tapdb/assets_store.go
```diff
@@ -1153,7 +1153,8 @@ func (a *AssetStore) FetchAssetProofs(ctx context.Context,
 		// TODO(roasbeef): can modify the query to use IN somewhere
 		// instead? then would take input params and insert into
 		// virtual rows to use
-		for _, locator := range targetAssets {
+		for ind := range targetAssets {
```
Roasbeef, line 1156: Was this from an older linter before the new scoping rules became the default?

Roasbeef (approved):
LGTM 🪷

tapdb/assets_store.go
jharveyb: It's from gosec rule G601, but the new rules won't apply for this project for a while since we're pinned to go 1.20. With the `GOEXPERIMENT=loopvar` build option we should already be fine but would rather leave this lint on anyways.

Issue #124, opened 2022-09-15
tarodb: add additional unit test coverage for asset coin selection 


Roasbeef:
As described on the tittle, rn we have minimal coverage to being integrating other related sub-systems.

PR #823, opened 2024-03-06
Improve coin selection unit tests

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/124.


jharveyb:
Found some issues (in the unit tests) revealed by these test changes.

Put up a branch that could get pulled into this PR: https://github.com/lightninglabs/taproot-assets/tree/coin-selection-test-fixes

jharveyb (changes requested):
See above comment

guggero:
Thanks for the review and the fixes. I took over the commit with the group anchor improvement.

> IIUC the timeout would occur for any signal since the method call was complete.

That was kind of on purpose. We explicitly expect the `LeaseCoins` method not to be called. I agree that the timeout would occur on any other channel as well. But IMO selecting on the `listSignals` a second time doesn't really change that?

jharveyb:
> That was kind of on purpose. We explicitly expect the `LeaseCoins` method not to be called. I agree that the timeout would occur on any other channel as well. But IMO selecting on the `listSignals` a second time doesn't really change that?

Ah, fair enough - so any of the 3 calls timing out is sufficient.

jharveyb (approved):
LGTM, coverage go up 🚀

Roasbeef (approved):
LGTM 🎉

Issue #721, opened 2023-12-05
[feature]: Enhance caretaker w/ ability to set user-defined UTXOs and GenesisAmtSats values


dstadulis:
**Is your feature request related to a problem? Please describe.**
As a `taproot-assets user` I want to `select user-defined UTXOs and GenesisAmtSats values` in order to
* `be able to include my taproot-asset proofs in UTXOs I choose`
* `Reduce inputs / obviate CPFP, in order to update the proofs`
* `privacy: increase covertness of taproot-asset UTXOs to deny meta-data fingerprinting`

[Edit Capturing TODOs from #866]
- [ ] Extend tapdb/asset_minting tests to cover new query params for script key, group key, etc.
- [ ] Add tests for tapdb/asset_minting/CommitBatchTx, AddSeedlingGroups, FetchSeedlingGroups

https://github.com/lightninglabs/taproot-assets/blob/eb79a0766f593fe8bfa1866a212740ed8c6a34b2/tapgarden/caretaker.go#L53-L57

**Describe the solution you'd like**
tapd and lnd work together to do UTXO selection

**Describe alternatives you've considered**
recompiling the daemon any time a custom `GenesisAmtSats` value is desired

jharveyb:
With #858 and #866 , this would now involve:

- [ ] Update `FundBatch` to also accept a serialized PSBT (which should have 1+ inputs, any may specify a change output).
- [ ] Update `fundGenesisPsbt` to handle a PSBT that may already include a change output (would the new `FundPsbt` call handle adjusting the change value? I think so.).
- [ ] Extra checks to ensure that the genesis point is still computed correctly for genesis anchor TXs with 1+ inputs.

We may want to apply other limits on the user-provided PSBT.

dstadulis:
https://github.com/lightninglabs/taproot-assets/issues/721 implementation will be pushed to v0.4.1 -- justification is: gather more security assurances before tapd uses wumbo UTXOs.

SadiqJaf:
## Pinned capability result

We tested the custom-genesis input/PSBT controls described in this issue against a confirmed leafless two-party MuSig2 P2TR UTXO that LND did not own. The UTXO was independently spent as a control before the mint and transfer probes.

[...]

PR #827, opened 2024-03-11
tapgarden: add FundBatch and SealBatch requests

### LINKED ISSUE #820: tapgarden: earlier handling of asset script keys
Requirement for #722.

When issuing an asset into a group, we need to produce a valid group witness. Right now the only option is a signature with the group key itself, over a virtual TX that represents the asset genesis. This requires the assets in the batch to be fully-formed, with a genesis point for the batch selected and a script key derived.

Right now, asset script keys cannot be provided by the user, and they are always BIP86 keys. The user also cannot specify a genesis point for the batch. We derive a group key right after setting these fields in an asset.

To use a custom group key witness, we would need to have the asset fully-formed earlier, and expose the virtual TX representing the asset genesis to a user for them to produce a valid group witness.

Specifically:

- [ ] Add new options when submitting  a `MintAsset` request, to also specify a script key.
- [ ] Derive script keys for asset seedlings when they are submitted, before the batch is frozen.
- [ ] Derive internal keys for asset groups when seedlings are submitted, before the batch is frozen (or accept external keys as the internal key for a group key?)
- [ ] Select the genesis outpoint earlier when freezing the batch (or when creating a batch?)
- [ ] Expose the virtual TX for grouped asset genesis to the user, perhaps via the `ListBatches` RPC?
- [ ] Modify `ListBatches` and other RPCs to always represent seedlings in a batch in insertion order, so that users can provide witnesses in the same order.
- [ ] Accept group witnesses for assets in a batch, likely as part of the `Finalize` call.

jharveyb:
Closes #820

Update the minting state machines to move batch funding into a separate command that can execute before batch finalization. Add a new command to supply group anchor witnesses before batch finalization.

Remaining TODOs:

- [x] derive seedling script keys in `prepAssetSeedling`
- [x] DB migration to add seedling script key reference field
- [x] Update DB queries to write & fetch seedling script key
- [x] Update DB queries to write & fetch batch TX
- [x] derive group internal key in `prepAssetSeedling`
- [x] Update DB queries to write & fetch seedling group internal key
- [ ] Extend `tapdb/asset_minting` tests to cover new query params
- [ ] Add tests for `tapdb/asset_minting/CommitBatchTx`
- [x] Resolve `FinalizeParams` handling wrt. `fundBatch`
- [x] Extend `tapgarden/planter_test` test to cover `FundBatch`
- [x] Implement `SealBatch`
- [x] Remove batch ticker from caretaker and `tapgarden` tests

### DIFF
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -695,6 +695,30 @@ type GroupKeyRequest struct {
 	NewAsset *Asset
 }
 
+// GroupVirtualTx contains all the information needed to produce an asset group
+// witness, except for the group internal key descriptor (or private key). A
+// GroupVirtualTx is constructed from a GroupKeyRequest.
+type GroupVirtualTx struct {
+	// Tx is a virtual transaction that represents the genesis state
+	// transition of a grouped asset.
+	Tx wire.MsgTx
+
+	// PrevOut is a transaction output that represents a grouped asset.
+	// PrevOut uses the tweaked group key as its PkScript. This is used in
+	// combination with GroupVirtualTx.Tx as input for a GenesisSigner.
+	PrevOut wire.TxOut
+
+	// GenID is the asset ID of the grouped asset in a GroupKeyRequest. This
+	// ID is needed to construct a sign descriptor for a GenesisSigner, as
+	// it is the single tweak for the group internal key.
+	GenID ID
+
+	// TweakedKey is the tweaked group key for the given GroupKeyRequest.
+	// This is later used to construct a complete GroupKey, after a
+	// GenesisSigner has produced an asset group witness.
+	TweakedKey btcec.PublicKey
+}
+
 // GroupKeyReveal is a type for representing the data used to derive the tweaked
 // key used to identify an asset group. The final tweaked key is the result of:
 // TapTweak(groupInternalKey, tapscriptRoot)
@@ -959,6 +983,50 @@ func (s ScriptKey) IsUnSpendable() (bool, error) {
 	return NUMSPubKey.IsEqual(s.PubKey), nil
 }
 
+// IsEqual returns true is this tweaked script key is exactly equivalent to the
+// passed other tweaked script key.
+func (ts *TweakedScriptKey) IsEqual(other *TweakedScriptKey) bool {
+	if ts == nil {
+		return other == nil
+	}
+
+	if other == nil {
+		return false
+	}
+
+	if !bytes.Equal(ts.Tweak, other.Tweak) {
+		return false
+	}
+
+	return EqualKeyDescriptors(ts.RawKey, other.RawKey)
+}
+
+// IsEqual returns true is this script key is exactly equivalent to the passed
+// other script key.
+func (s *ScriptKey) IsEqual(otherScriptKey *ScriptKey) bool {
+	if s == nil {
+		return otherScriptKey == nil
+	}
+
+	if otherScriptKey == nil {
+		return false
+	}
+
+	if s.PubKey == nil {
+		return otherScriptKey.PubKey == nil
+	}
+
+	if otherScriptKey.PubKey == nil {
+		return false
+	}
+
+	if !s.TweakedScriptKey.IsEqual(otherScriptKey.TweakedScriptKey) {
+		return false
+	}
+
+	return s.PubKey.IsEqual(otherScriptKey.PubKey)
+}
+
 // NewScriptKey constructs a ScriptKey with only the publicly available
 // information. This resulting key may or may not have a tweak applied to it.
 func NewScriptKey(key *btcec.PublicKey) ScriptKey {
@@ -1045,72 +1113,20 @@ func (req *GroupKeyRequest) Validate() error {
 		return fmt.Errorf("missing group internal key")
 	}
 
+	tapscriptRootSize := len(req.TapscriptRoot)
+	if tapscriptRootSize != 0 && tapscriptRootSize != sha256.Size {
+		return fmt.Errorf("tapscript root must be %d bytes",
+			sha256.Size)
+	}
+
 	return nil
 }
 
-// DeriveGroupKey derives an asset's group key based on an internal public
-// key descriptor, the original group asset genesis, and the asset's genesis.
-func DeriveGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
-	req GroupKeyRequest) (*GroupKey, error) {
-
-	// First, perform the final checks on the asset being authorized for
-	// group membership.
-	err := req.Validate()
-	if err != nil {
-		return nil, err
-	}
-
-	// Compute the tweaked group key and set it in the asset before
-	// creating the virtual minting transaction.
-	genesisTweak := req.AnchorGen.ID()
-	tweakedGroupKey, err := GroupPubKey(
-		req.RawKey.PubKey, genesisTweak[:], nil,
-	)
-	if err != nil {
-		return nil, fmt.Errorf("cannot tweak group key: %w", err)
-	}
-
-	assetWithGroup := req.NewAsset.Copy()
-	assetWithGroup.GroupKey = &GroupKey{
-		GroupPubKey: *tweakedGroupKey,
-	}
-
-	// Build the virtual transaction that represents the minting of the new
-	// asset, which will be signed to generate the group witness.
-	genesisTx, prevOut, err := genBuilder.BuildGenesisTx(assetWithGroup)
-	if err != nil {
-		return nil, fmt.Errorf("cannot build virtual tx: %w", err)
-	}
-
-	// Build the static signing descriptor needed to sign the virtual
-	// minting transaction. This is restricted to group keys with an empty
-	// tapscript root and key path spends.
-	signDesc := &lndclient.SignDescriptor{
-		KeyDesc:     req.RawKey,
-		SingleTweak: genesisTweak[:],
-		SignMethod:  input.TaprootKeySpendBIP0086SignMethod,
-		Output:      prevOut,
-		HashType:    txscript.SigHashDefault,
-		InputIndex:  0,
-	}
-	sig, err := genSigner.SignVirtualTx(signDesc, genesisTx, prevOut)
-	if err != nil {
-		return nil, err
-	}
-
-	return &GroupKey{
-		RawKey:      req.RawKey,
-		GroupPubKey: *tweakedGroupKey,
-		Witness:     wire.TxWitness{sig.Serialize()},
-	}, nil
-}
-
-// DeriveCustomGroupKey derives an asset's group key based on a signing
-// descriptor, the original group asset genesis, and the asset's genesis.
-func DeriveCustomGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
-	req GroupKeyRequest, tapLeaf *psbt.TaprootTapLeafScript,
-	scriptWitness []byte) (*GroupKey, error) {
-
+// BuildGroupVirtualTx derives the tweaked group key for group key request,
+// and constructs the group virtual TX needed to construct a sign descriptor and
+// produce an asset group witness.
+func (req *GroupKeyRequest) BuildGroupVirtualTx(genBuilder GenesisTxBuilder) (
+	*GroupVirtualTx, error) {
 	// First, perform the final checks on the asset being authorized for
 	// group membership.
 	err := req.Validate()
@@ -1133,26 +1149,6 @@ func DeriveCustomGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
 		GroupPubKey: *tweakedGroupKey,
 	}
 
-	// Exit early if a group witness is already given, since we don't need
-	// to construct a virtual TX nor produce a signature.
-	if scriptWitness != nil {
-		if tapLeaf == nil {
-			return nil, fmt.Errorf("need tap leaf with group " +
-				"script witness")
-		}
-
-		witness := wire.TxWitness{
-			scriptWitness, tapLeaf.Script, tapLeaf.ControlBlock,
-		}
-
-		return &GroupKey{
-			RawKey:        req.RawKey,
-			GroupPubKey:   *tweakedGroupKey,
-			TapscriptRoot: req.TapscriptRoot,
-			Witness:       witness,
-		}, nil
-	}
-
 	// Build the virtual transaction that represents the minting of the new
 	// asset, which will be signed to generate the group witness.
 	genesisTx, prevOut, err := genBuilder.BuildGenesisTx(assetWithGroup)
@@ -1160,13 +1156,57 @@ func DeriveCustomGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
 		return nil, fmt.Errorf("cannot build virtual tx: %w", err)
 	}
 
+	return &GroupVirtualTx{
+		Tx:         *genesisTx,
+		PrevOut:    *prevOut,
+		GenID:      genesisTweak,
+		TweakedKey: *tweakedGroupKey,
+	}, nil
+}
+
+// AssembleGroupKeyFromWitness constructs a group key given a group witness
+// generated externally.
+func AssembleGroupKeyFromWitness(genTx GroupVirtualTx, req GroupKeyRequest,
+	tapLeaf *psbt.TaprootTapLeafScript, scriptWitness []byte) (*GroupKey,
+	error) {
+
+	if scriptWitness == nil {
+		return nil, fmt.Errorf("script witness cannot be nil")
+	}
+
+	groupKey := &GroupKey{
+		RawKey:        req.RawKey,
+		GroupPubKey:   genTx.TweakedKey,
+		TapscriptRoot: req.TapscriptRoot,
+		Witness:       wire.TxWitness{scriptWitness},
+	}
+
+	if tapLeaf != nil {
+		if tapLeaf.LeafVersion != txscript.BaseLeafVersion {
+			return nil, fmt.Errorf("unsupported script version")
+		}
+
+		groupKey.Witness = append(
+			groupKey.Witness, tapLeaf.Script, tapLeaf.ControlBlock,
+		)
+	}
+
+	return groupKey, nil
+}
+
+// DeriveGroupKey derives an asset's group key based on an internal public key
+// descriptor, the original group asset genesis, and the asset's genesis.
+func DeriveGroupKey(genSigner GenesisSigner, genTx GroupVirtualTx,
+	req GroupKeyRequest, tapLeaf *psbt.TaprootTapLeafScript) (*GroupKey,
+	error) {
+
 	// Populate the signing descriptor needed to sign the virtual minting
 	// transaction.
 	signDesc := &lndclient.SignDescriptor{
 		KeyDesc:     req.RawKey,
-		SingleTweak: genesisTweak[:],
+		SingleTweak: genTx.GenID[:],
 		TapTweak:    req.TapscriptRoot,
-		Output:      prevOut,
+		Output:      &genTx.PrevOut,
 		HashType:    txscript.SigHashDefault,
 		InputIndex:  0,
 	}
@@ -1193,7 +1233,7 @@ func DeriveCustomGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
 		return nil, fmt.Errorf("bad sign descriptor for group key")
 	}
 
-	sig, err := genSigner.SignVirtualTx(signDesc, genesisTx, prevOut)
+	sig, err := genSigner.SignVirtualTx(signDesc, &genTx.Tx, &genTx.PrevOut)
 	if err != nil {
 		return nil, err
 	}
@@ -1204,13 +1244,14 @@ func DeriveCustomGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
 	// the control block to the witness, otherwise the verifier will reject
 	// the generated witness.
 	if signDesc.SignMethod == input.TaprootScriptSpendSignMethod {
-		witness = append(witness, signDesc.WitnessScript)
-		witness = append(witness, tapLeaf.ControlBlock)
+		witness = append(
+			witness, signDesc.WitnessScript, tapLeaf.ControlBlock,
+		)
 	}
 
 	return &GroupKey{
 		RawKey:        signDesc.KeyDesc,
-		GroupPubKey:   *tweakedGroupKey,
+		GroupPubKey:   genTx.TweakedKey,
 		TapscriptRoot: signDesc.TapTweak,
 		Witness:       witness,
 	}, nil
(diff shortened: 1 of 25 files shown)


dstadulis:
Will be pulled out of draft and merged this sprint

jharveyb:
Feature work is done IMO, follow-on work can happen in a separate PR as this one has already ballooned quite a bit.

Remaining tasks are just unit test coverage AFAICT, should add some usage in an itest as well.

asset/asset.go
```diff
@@ -695,6 +695,16 @@ type GroupKeyRequest struct {
 	NewAsset *Asset
 }
 
+type GroupGenesisTx struct {
```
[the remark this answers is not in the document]
jharveyb: Good point - renamed to `GroupVirtualTx` and virtual feels more consistent with the methods also used in this process + our other language around vPSBTs. I think pseudo would also fit.

fn/iter.go
```diff
@@ -28,9 +28,9 @@ func ForEach[T any](items []T, f func(T)) {
 // ForEachMapItem is a generic implementation of a for-each (map with side
 // effects). This can be used to ensure that any normal for-loop don't run into
 // bugs due to loop variable scoping.
-func ForEachMapItem[T any, K comparable](items map[K]T, f func(T)) {
+func ForEachMapItem[T any, K comparable](items map[K]T, f func(K, T)) {
```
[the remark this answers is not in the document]
jharveyb: It was unused before this PR, and is now used in one place:

https://github.com/lightninglabs/taproot-assets/blob/1ab44287e243bfebd7a4a4636d7ef1244540ff71/tapgarden/planter.go#L484

I think a caller could ignore the map keys with `_` if they want the old behavior.

tapgarden/planter.go
```diff
@@ -769,9 +816,7 @@ func (c *ChainPlanter) finalizeBatch(params FinalizeParams) (*BatchCaretaker,
 
 	// At this point, we have a non-empty batch, so we'll first finalize it
 	// on disk. This means no further seedlings can be added to this batch.
-	ctx, cancel = c.WithCtxQuit()
```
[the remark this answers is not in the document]
jharveyb: I think this is resolved in later commits?

jharveyb:
One DB change is actually missing from this PR, related to group key handling. We need to also associate a `[32]byte` with a Seedling, which would later map to a `GroupKey.TapscriptRoot`. Without that, a user could not later provide a custom tapscript root for a group key, and if they generated witness using such a root they would not verify with the group virtual TXs made during `sealBatch`.

tapgarden/interface.go
```diff
@@ -43,6 +43,14 @@ type Planter interface {
 	// returned.
 	CancelSeedling() error
 
+	// FundBatch attempts to provide a genesis point for the current batch,
+	// or create a new funded batch.
+	FundBatch(params FundParams) (*MintingBatch, error)
```
Roasbeef, line 48: So with the the _user_ will select the genesis point? I thought it was just enough to do the coin selection earlier, then make it available to the user if needed so they can provide the custom witness?

tapgarden/planter.go
```diff
@@ -135,6 +135,27 @@ type FinalizeParams struct {
 	SiblingTapTree fn.Option[asset.TapscriptTreeNodes]
 }
 
+// FundParams are the options available to change how a batch is funded, and how
+// the genesis TX is constructed.
+type FundParams struct {
+	FeeRate        fn.Option[chainfee.SatPerKWeight]
+	SiblingTapTree fn.Option[asset.TapscriptTreeNodes]
```
Roasbeef, line 139: Comment above says a genesis point, but looks like this is mainly about proving the sibling tap tree? Also thought we already had that plumbing hooked up here.

tapgarden/planter.go
```diff
@@ -135,6 +135,27 @@ type FinalizeParams struct {
...
+	// TODO(jhb): follow-up PR: accept a PSBT here
+}
+
+// groupSeal specifies the group witness for a seedling in a funded batch.
+type groupSeal struct {
+	GroupMember  asset.ID
```
Roasbeef, line 145: Doesn't this also need the new asset name and the existing group key?

tapgarden/planter.go
```diff
@@ -644,6 +667,29 @@ func (c *ChainPlanter) gardener() {
...
+				err = c.fundBatch(ctx, *fundReqParams)
+				cancel()
+				if err != nil {
+					req.Error(fmt.Errorf("unable to fund "+
+						"minting batch: %w", err))
+					break
+				}
+
+			// TODO(jhb): follow-up PR: Implement SealBatch command
+			case reqTypeSealBatch:
```
Roasbeef, line 890: Have it just return an error in this temp state?

tapdb/asset_minting.go
```diff
@@ -328,6 +328,37 @@ func (a *AssetMintingStore) CommitMintingBatch(ctx context.Context,
...
+
+			genesisPointID, err := upsertGenesisPoint(
+				ctx, q, genesisOutpoint,
+			)
+			if err != nil {
+				return fmt.Errorf("unable to upsert genesis "+
+					"point: %w", err)
+			}
+
+			return q.BindMintingBatchWithTx(ctx, BatchChainUpdate{
```
Roasbeef, line 366: Passing thought: does this invalidate any existing logic that assumes that only a near finalized batch can have a batch tx bound to it?

tapgarden/caretaker.go
```diff
@@ -383,64 +383,6 @@ func (b *BatchCaretaker) assetCultivator() {
 	}
 }
 
-// fundGenesisPsbt generates a PSBT packet we'll use to create an asset.  In
-// order to be able to create an asset, we need an initial genesis outpoint. To
-// obtain this we'll ask the wallet to fund a PSBT template for GenesisAmtSats
-// (all outputs need to hold some BTC to not be dust), and with a dummy script.
-// We need to use a dummy script as we can't know the actual script key since
-// that's dependent on the genesis outpoint.
-func (b *BatchCaretaker) fundGenesisPsbt(
```
Roasbeef, line 392: Can you add the rationale for the move to the commit body?

tapgarden/planter.go
```diff
@@ -773,18 +835,14 @@ func (c *ChainPlanter) gardener() {
 	}
 }
 
-// finalizeBatch creates a new caretaker for the batch and starts it.
-func (c *ChainPlanter) finalizeBatch(params FinalizeParams) (*BatchCaretaker,
-	error) {
-
+// If funding fails should we delete the batch?
```
Roasbeef, line 975: stray comment

tapgarden/planter.go
```diff
@@ -773,18 +835,14 @@ func (c *ChainPlanter) gardener() {
...
+func (c *ChainPlanter) fundBatch(ctx context.Context, params FundParams) error {
```
Roasbeef, line 976: Will this routine be called multiple times now? Otherwise why would the batch already be funded before this call?

tapgarden/planter.go
```diff
@@ -797,32 +855,140 @@ func (c *ChainPlanter) finalizeBatch(params FinalizeParams) (*BatchCaretaker,
...
+	error) {
+
+	var (
+		feeRate *chainfee.SatPerKWeight
+		err     error
+	)
+
+	// Process the finalize parameters.
+	params.FeeRate.WhenSome(func(fr chainfee.SatPerKWeight) {
+		feeRate = &fr
```
Roasbeef, line 1167: Can write alternatively as: 
```go
feeRate := params.FeeRate.UnwrapOr(nil)
```

tapgarden/planter.go
```diff
@@ -797,32 +855,140 @@ func (c *ChainPlanter) finalizeBatch(params FinalizeParams) (*BatchCaretaker,
...
+	})
+
+	ctx, cancel := c.WithCtxQuit()
+	defer cancel()
+
+	params.SiblingTapTree.WhenSome(func(tn asset.TapscriptTreeNodes) {
```
Roasbeef, line 1173: Can write alternatively as: 
```go
err := fn.MapOptionZ(params.SiblingTapTree, func(tn asset.TapscriptTreeNodes) error {
       return c.cfg.TreeStore.StoreTapscriptTree(ctx, tn)
})
```

tapdb/sqlc/migrations/000017_seedling_script_group_keys.up.sql
```diff
@@ -0,0 +1,12 @@
...
+ALTER TABLE asset_seedlings ADD COLUMN script_key_id BIGINT REFERENCES script_keys(script_key_id);
+
+-- For a group anchor, we derive the internal key for the future group key early,
+-- to allow use of custom group witnesses.
+ALTER TABLE asset_seedlings ADD COLUMN group_internal_key_id BIGINT REFERENCES internal_keys(key_id);
+
+-- For a group key, the internal key can also be tweaked to commit to a
+-- tapscript tree. Once we finalize the batch, this tweak will also be stored
+-- as part of the asset group itself.
+ALTER TABLE asset_seedlings ADD COLUMN group_tapscript_root BLOB;
```
Roasbeef, line 12: Will this also eventually ref the tapscript trees we store on disk? 

Also wouldn't it be enough to just ref an group key here since that already have the sibling hash as a column?

tapdb/sqlc/queries/assets.sql
```diff
@@ -108,11 +111,14 @@ WITH target_key_id AS (
...
-    emission_enabled, batch_id, group_genesis_id, group_anchor_id
+    emission_enabled, batch_id, group_genesis_id, group_anchor_id,
+    script_key_id, group_internal_key_id, group_tapscript_root
 ) VALUES (
     $2, $3, $4, $5, $6, $7,
     (SELECT key_id FROM target_key_id),
-    sqlc.narg('group_genesis_id'), sqlc.narg('group_anchor_id')
+    sqlc.narg('group_genesis_id'), sqlc.narg('group_anchor_id'),
+    sqlc.narg('script_key_id'), sqlc.narg('group_internal_key_id'),
+   sqlc.narg('group_tapscript_root')
```
Roasbeef, line 121: Needs an extra space to line up w/ the lines above?

tapdb/asset_minting.go
```diff
@@ -429,6 +434,71 @@ func (a *AssetMintingStore) CommitMintingBatch(ctx context.Context,
 	return err
 }
 
+func insertOptionalSeedlingParams(ctx context.Context, q PendingAssetStore,
+	batchKey []byte, seedling *tapgarden.Seedling) ([4]sql.NullInt64,
```
Roasbeef, line 438: How about returning a struct instead of an array? That was we can ref each of them by a field instead of the index which might change over time.

tapdb/asset_minting.go
```diff
@@ -543,6 +606,70 @@ func fetchAssetSeedlings(ctx context.Context, q PendingAssetStore,
...
+					RawKey: keychain.KeyDescriptor{
+						KeyLocator: scriptKeyLocator,
+						PubKey:     scriptKeyInternalPub,
+					},
+					Tweak: dbSeedling.ScriptKeyTweak,
+				},
+			}
+		}
+
+		if dbSeedling.GroupKeyRaw != nil {
```
Roasbeef, line 645: We should update the unit tests assertions to make sure these are always set as expected.

tapgarden/planter.go
```diff
@@ -1133,6 +1133,40 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
 		}
 	}
 
+	if c.pendingBatch != nil {
+		if _, ok := c.pendingBatch.Seedlings[req.AssetName]; ok {
```
Roasbeef, line 1354: Thought we already had this check elsewhere?

tapgarden/seedling.go
```diff
@@ -95,6 +97,22 @@ type Seedling struct {
 
 	// update is used to send updates w.r.t the state of the batch.
 	updates SeedlingUpdates
+
+	// ScriptKey is the tweaked Taproot key that will be used to spend the
+	// asset after minting. By default, this key is constructed with a
+	// BIP-0086 style tweak.
+	ScriptKey *asset.ScriptKey
```
Roasbeef, line 104: IIUC, this'll always be set, so it doesn't need to be a pointer.

tapgarden/seedling.go
```diff
@@ -95,6 +97,22 @@ type Seedling struct {
...
+
+	// GroupInternalKey is the raw group key before the tweak with the
+	// genesis point or tapscript root has been applied.
+	GroupInternalKey *keychain.KeyDescriptor
+
+	// TapscriptRoot is the root of the Tapscript tree that commits to all
```
Roasbeef, line 110: `godoc` mismatch.

tapgarden/seedling.go
```diff
@@ -95,6 +97,22 @@ type Seedling struct {
...
+	// script spend conditions for the group key. Instead of spending an
+	// asset, these scripts are used to define witnesses more complex than
+	// a Schnorr signature for reissuing assets. A group key with an empty
+	// Tapscript root can only authorize reissuance with a signature.
+	GroupTapscriptRoot []byte
```
Roasbeef, line 115: Any reason to not make this a `[32]byte`? Then we won't need the checks below.

tapdb/asset_minting.go
```diff
@@ -1174,6 +1174,98 @@ func (a *AssetMintingStore) CommitBatchTx(ctx context.Context,
 	})
 }
 
+// AddSeedlingGroups stores the asset groups for seedlings associated with a
+// batch.
+func (a *AssetMintingStore) AddSeedlingGroups(ctx context.Context,
```
Roasbeef, line 1179: Can't this be done when we commit the batch to disk to mark it as finalized?

tapgarden/planter.go
```diff
@@ -477,6 +477,173 @@ func (c *ChainPlanter) fundGenesisPsbt(ctx context.Context,
 	return fundedGenesisPkt, nil
 }
 
+// filterSeedlingsWithGroup separates a set of seedlings into two sets based on
+// their relation to an asset group, which has not been constructed yet.
+func filterSeedlingsWithGroup(
+	seedlings map[string]*Seedling) (map[string]*Seedling,
+	map[string]*Seedling) {
+
+	WithGroup := make(map[string]*Seedling)
```
Roasbeef, line 486: Variable names should start with a lower case character.

tapgarden/planter.go
```diff
@@ -903,7 +1070,86 @@ func (c *ChainPlanter) fundBatch(ctx context.Context, params FundParams) error {
 	return nil
 }
 
-func (c *ChainPlanter) sealBatch(params SealParams) error {
+func (c *ChainPlanter) sealBatch(ctx context.Context, _ SealParams) error {
```
Roasbeef, line 1073: Doesn't use the param?

tapgarden/planter.go
```diff
@@ -903,7 +1070,86 @@ func (c *ChainPlanter) fundBatch(ctx context.Context, params FundParams) error {
...
+			protoAsset.Genesis, protoAsset.Amount,
+			protoAsset.LockTime, protoAsset.RelativeLockTime,
+			protoAsset.ScriptKey, groupKey,
+			asset.WithAssetVersion(protoAsset.Version),
+		)
+		if err != nil {
+			return err
+		}
+
+		err = c.cfg.TxValidator.Execute(groupedAsset, nil, nil)
```
Roasbeef, line 1133: Where's the witness? Also comments here aren't full sentences above.

Roasbeef (changes requested):

tapgarden/interface.go
jharveyb: The `FundBatch` call, in this PR, just triggers coin selection, yes. The `FundParams` are the same as `FinalizeParams`. In some future PR we can expand those params to include a template PSBT with a genesis point (which would address #721)

tapgarden/planter.go
jharveyb: At the point of providing witnesses, we can already display the asset ID for each seedling to the user, since the batch was already funded. And the info given when submitting the seedling is enough to map the seedling to the group it would be a part of (implemented in `buildGroupReqs`), so we don't need that info again.

tapdb/asset_minting.go
jharveyb: Good Q, so far my conclusion was no, as the funding happened in the first state of the caretaker and that field of the batch struct isn't modified before then. Since we don't unwind batch funding right now (#830), there's no logic around that to change.

We may want to delete genesis points from cancelled batches later on.

tapdb/sqlc/migrations/000017_seedling_script_group_keys.up.sql
jharveyb: > Will this also eventually ref the tapscript trees we store on disk?

The value will, but this specific field won't. It'll be copied over to the final entry in the `asset_groups` table. I don't think we need to require a stored tapscript tree with a matching root though. If the user is generating a key-path witness, we only need the root and not the full tree. We only need the tree to validate a script-path witness.

"Make a group key with this tapscript tree root. If I want to produce a script-path witness, I'll later provide the tapscript tree for this tapscript tree root."

We haven't exposed the tapscript tree store yet, so the tapscript tree could be a field in the params for `sealBatch`.

> Also wouldn't it be enough to just ref an group key here since that already have the sibling hash as a column?

No, as this happens before group keys have been fully formed and inserted into `asset_groups` at all. Before being able to seal the batch, we need both an internal key for the group and this optional tweak; the asset ID can be computed after batch funding, which is the missing piece for deriving the final tweaked key.

We can only insert the asset group once we have the group witness, so at the end of `sealBatch`.

tapgarden/planter.go
jharveyb: Yes, right now it's checked in `batch.addSeedling` - I added it here so it would happen earlier before we derive the script key for the seedling, but forgot to remove `addSeedling`.

tapgarden/planter.go
jharveyb: Deferred for a follow-up PR.

tapgarden/seedling.go
jharveyb: Good point, I don't think so.

tapdb/asset_minting.go
jharveyb: As in defer it to the end of `finalizeBatch` instead of at the end of `sealBatch`? If we don't write it at the end of `sealBatch` then we'd still need to persist the populated asset groups somewhere, like the batch state itself.

Though as-is we'd also want to delete the group if the batch gets cancelled, which isn't implemented.

Right now this data gets written at the end of the caretaker `BatchStateFrozen` in `AddSproutsToBatch`.

tapgarden/planter.go
jharveyb: The witness is a field of `groupKey` above, and we have to re-make the asset with the witness populated to actually validate.

tapgarden/seedling.go
jharveyb: Thinking about it further, the reason we use `[]byte` for the tapscript root elsewhere is to not store the all-zero byte slice by default / for all cases where there is no root.

tapgarden/planter.go
jharveyb: Since we're also returning the `rootHash` I think this doesn't work out.

jharveyb:
Updated to address some feedback and update tests to exercise query changes. Also fixed some issues with not accepting externally-set seedling script keys and group internal keys.

Hitting some issues with minting DB tests around using the seedling group tapscript root.

From discussion offline, I think the simplest behavior for `FundBatch` and `SealBatch` would be that they can be called exactly once before `Finalize`. Its straightforward to detect if a batch is funded or sealed, so it should also be straightforward to return an error here and require the batch to be resubmitted.

tapdb/asset_minting.go
```diff
@@ -408,35 +415,38 @@ func (a *AssetMintingStore) CommitMintingBatch(ctx context.Context,
...
+				GroupTapscriptRoot
+
+			optionalDbIDs, err := insertOptionalSeedlingParams(
+				ctx, q, rawBatchKey, seedling,
+			)
+			if err != nil {
+				return err
+			}
+
+			dbSeedling.GroupInternalKeyID =
```
guggero, line 445: Much nicer with the named struct!

tapdb/assets_common.go
```diff
@@ -84,21 +84,55 @@ type UpsertAssetStore interface {
...
+
 // upsertGenesis imports a new genesis point into the database or returns the
 // existing ID if that point already exists.
 func upsertGenesisPoint(ctx context.Context, q UpsertAssetStore,
 	genesisOutpoint wire.OutPoint) (int64, error) {
 
 	genesisPoint, err := encodeOutpoint(genesisOutpoint)
 	if err != nil {
-		return 0, fmt.Errorf("unable to encode genesis point: %w", err)
+		return 0, fmt.Errorf("%w: %w", ErrEncodeGenesisPoint, err)
```
guggero, line 128: I know newer Golang versions allow for double wrapping errors. But not sure how that works exactly? Do we really want to do that?

asset/asset.go
```diff
@@ -695,6 +695,16 @@ type GroupKeyRequest struct {
 	NewAsset *Asset
 }
 
+type GroupVirtualTx struct {
```
guggero, line 698: nit: Godoc missing (struct and members).

asset/asset.go
```diff
@@ -1045,14 +1055,18 @@ func (req *GroupKeyRequest) Validate() error {
...
+
 	return nil
 }
 
-// DeriveGroupKey derives an asset's group key based on an internal public
-// key descriptor, the original group asset genesis, and the asset's genesis.
-func DeriveGroupKey(genSigner GenesisSigner, genBuilder GenesisTxBuilder,
-	req GroupKeyRequest) (*GroupKey, error) {
-
+// Build the virtual TX needed to produce an asset group witness.
```
guggero, line 1067: nit: Godoc comment doesn't match function name.

guggero (approved):
LGTM, pending existing comments and successful CI run.

tapdb/assets_common.go
jharveyb: Good catch, I tested this in the playground and what I added doesn't work as expected. I mostly wanted to dedupe these error messages; AFAICT callers right now aren't inspecting the concrete error type.

So I think the fix would be to just define these new variables as strings instead of errors, and use:

`fmt.Errorf("%s: %w)`.

Callers would then have to use `strings.Contains()` instead of `errors.Is()`.

jharveyb:
Added logic to reject `FundBatch` and `FinalizeBatch` calls when the pending batch is already funded.
Updated minting DB tests for the new seedling key handling, and for better coverage.

We had some behavior around group verifiers that assumed that the group tapscript root was always empty, which is fixed.
I also removed the old static `DeriveGroupKey` func, as it is no longer appropriate with all the new seedling options (and it wasn't being tested in the VM tests either).

Also added a test around making a batch with external seedling keys and actually using `FundBatch`.

Moving on to the follow-up PR but will some spots here like new queries around seedling groups still need tests added.

jharveyb:
Rebased.

Roasbeef (approved):
LGTM 🦚

guggero:
Needs another rebase to apply new linter rules, then a couple of fixes for the new rules.

PR #829, opened 2024-03-12
Remove duplicate comment words

dstadulis:
Duplicate words are identified with the regex:

^[\s\t]*//.*\b(\w+)\b \1\b
^[\s\t]*#.*\b(\w+)\b \1\b

and removed

### DIFF
--- a/tapdb/interfaces.go
+++ b/tapdb/interfaces.go
@@ -49,7 +49,7 @@ type TxOptions interface {
 type BatchedTx[Q any] interface {
 	// ExecTx will execute the passed txBody, operating upon generic
 	// parameter Q (usually a storage interface) in a single transaction.
-	// The set of TxOptions are passed in in order to allow the caller to
+	// The set of TxOptions are passed in order to allow the caller to
 	// specify if a transaction should be read-only and optionally what
 	// type of concurrency control should be used.
 	ExecTx(ctx context.Context, txOptions TxOptions,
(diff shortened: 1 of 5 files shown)


tapdb/interfaces.go
```diff
@@ -49,7 +49,7 @@ type TxOptions interface {
 type BatchedTx[Q any] interface {
 	// ExecTx will execute the passed txBody, operating upon generic
 	// parameter Q (usually a storage interface) in a single transaction.
-	// The set of TxOptions are passed in in order to allow the caller to
+	// The set of TxOptions are passed in order to allow the caller to
```
guggero, line 52: I think this one is actually on purpose. Could perhaps be written as "passed in in-order" to be more clear.

GeorgeTsagk: "in-order" as in ordered in a specific way? after looking at `TxOptions` there doesn't seem to be any slice or array that holds entities in some order. So perhaps change the phrasing completely and do something like
```suggestion
	// The set of TxOptions are provided in order to allow the caller to
```

GeorgeTsagk (approved):
LGTM, pending the "in in order" nit

tapdb/interfaces.go
guggero: Ah, no, I misread it myself. So your suggestion is actually better.

guggero:
!lightning-labs-deploy mute

guggero:
@dstadulis are you ever going to update this PR?

Issue #818, opened 2024-02-29
[feature]: Extend the RFQ service to handle asset sell (send) initiated lightning payments


ffranr:
The [current RFQ service implementation](https://github.com/lightninglabs/taproot-assets/issues/683) supports asset purchase requests, enabling users to acquire tap assets at the end of a Lightning payment chain. This issue seeks to extend the RFQ service to encompass requests for asset sales. This enhancement enables users to settle a Lightning invoice by conducting an asset sale to the next hop in a Lightning payment route, based on mutual agreement.

Based on my assessment, the most straightforward method to accommodate asset sales within the RFQ service involves introducing new message types specifically designed for initiating and accepting asset sale requests. The existing request and accept message types will be designated for asset purchases only.

Upon completing this issue, the RFQ message types will be:

* `SellRequest` for initiating a sale,
* `SellAccept` for confirming the acceptance of a sale,
* `BuyRequest` for initiating a purchase,
* `BuyAccept` for confirming the acceptance of a purchase,
* `Reject` for declining any of the above requests.

Separate buy and sell request messages provides flexibility for the future, simplicity in naming and documenting fields, and unambiguous types in code.

PR #835, opened 2024-03-14
Add asset sell support to the RFQ service

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/818

This PR builds on top of the refactoring work from https://github.com/lightninglabs/taproot-assets/pull/814 .

It adds asset sell request and sell request accept support to the RFQ service.


dstadulis:
There's local code
need to rebase code
getting the itest

ffranr:
I'm not sure how to setup the itest for the case where we sell a tap asset as the payment to satisfy the first hop in a lightning payment route. 

Alice wants to pay Carol. Carol generates the invoice.

Alice wants to sell a tap asset to Bob, and Alice would like Bob to send the bid payment in that sale to Carol. In so doing Bob would satisfy the invoice that Carol generated for Alice.

I can currently:
* Reach a sale quote agreement between Alice and Bob.
* Generate an invoice with Carol.
* Generate a custom route from Alice, to Bob, and then to Carol.

Bob can use his RFQ service to listen for a HTLC from Alice that satisfies the agreed Alice sale quote. And Alice can create a custom route with an input amount of 0 to Bob, but then stipulate that Bob must pay Carol the balance agreed upon in Alice-Bob's sale quote.

Is that the way forward? @guggero ?

guggero:
There are important parts missing on the `lnd` side for this to work fully. So we need to find a compromise so we can test something. I think what we discussed offline at some point was that for now Alice would embed custom data to Bob (the same way you can embed `dest_custom_records` for the final receiver, we'd want to include a custom TLV record for Bob, I think we found that the `SendToRoute` might allow that or might need some additions to allow for it).
The custom data at this moment could just be the RFQ ID. Then Bob's `tapd` could intercept the HTLC, decode the custom record data and check that an offer with that RFQ ID exists, then check that the outgoing amount matches that offer.

ffranr:
> There are important parts missing on the `lnd` side for this to work fully. So we need to find a compromise so we can test something. I think what we discussed offline at some point was that for now Alice would embed custom data to Bob (the same way you can embed `dest_custom_records` for the final receiver, we'd want to include a custom TLV record for Bob, I think we found that the `SendToRoute` might allow that or might need some additions to allow for it). The custom data at this moment could just be the RFQ ID. Then Bob's `tapd` could intercept the HTLC, decode the custom record data and check that an offer with that RFQ ID exists, then check that the outgoing amount matches that offer.

@guggero In the asset buy payment initiation case the HTLC intercepting node uses the SCID to identify the relevant HTLC. Both peers are aware of the SCID as it is present in the agreed upon quote (derived from message ID). Why do we need custom records? Can't we just send forward a HTLC over the quote SCID?

guggero:
> @guggero In the asset buy payment initiation case the HTLC intercepting node uses the SCID to identify the relevant HTLC. Both peers are aware of the SCID as it is present in the accepted quote (SCID is derived from message ID). Why do we need custom records? Can't we just forward a HTLC over the accepted quote SCID?

Oh, you're right, we have the `incoming_channel_id`, I wasn't aware of that. I thought we'd only have the outgoing one, which in this case would be a "real" channel in the network. But then yes, we can just use the quote SCID to match things, no need for custom records.

ffranr:
> > @guggero In the asset buy payment initiation case the HTLC intercepting node uses the SCID to identify the relevant HTLC. Both peers are aware of the SCID as it is present in the accepted quote (SCID is derived from message ID). Why do we need custom records? Can't we just forward a HTLC over the accepted quote SCID?
> 
> Oh, you're right, we have the `incoming_channel_id`, I wasn't aware of that. I thought we'd only have the outgoing one, which in this case would be a "real" channel in the network. But then yes, we can just use the quote SCID to match things, no need for custom records.

I've managed to get the custom records method to work with a custom route. But I didn't manage to forward a HTLC when I set `ChanId` on the custom route or set the incoming channel on a hop hint. Not sure what's going on there, perhaps LND is rejecting the channel ID given that an actual channel exists.

ffranr:
In the last commit in this PR i bump lndclient so that I can inspect custom records with HTLC intercept. I'll spin out the bump commit into its own PR.

The first 9 commits can be split out into a separate PR. They are effectively refactor commits.

dstadulis:
itests are failing when running together

dstadulis:
@ffranr will split out PRs to get merge faster

ffranr:
please review https://github.com/lightninglabs/taproot-assets/pull/845 first

ffranr:
In today's standup, I talked about trying to fix a bug where two RFQ tests can't run at the same time. Turns out, the problem is with the lnd/lntest node manager – it just can't handle more than 7 nodes at once. If you try to add LND node numer 8, it'll just freeze up without giving you an error. This is probably because the default setting for the btcd miner RPC can only take 10 clients max, and adding another LND node would make it 11, which is one too many.

The simplest workaround for this issue, which likely necessitates modifying `lnd/lntest`, is to run the RFQ integration tests asynchronously. This approach allows us to progress to version 0.4 while planning to enhance `lntest` later. What are your thoughts on this strategy? Does anyone have deeper insights into `lntest`?

guggero:
To bypass the issue in `lnd/lntest`, did you try if re-using the already running standby nodes (Alice and Bob) would help? Then you need to start two fewer nodes. There are `tapd` nodes connected to those, but they should behave pretty independently when it comes to channels and on-chain funds.

ffranr:
> To bypass the issue in `lnd/lntest`, did you try if re-using the already running standby nodes (Alice and Bob) would help? Then you need to start two fewer nodes. There are `tapd` nodes connected to those, but they should behave pretty independently when it comes to channels and on-chain funds.

I will try what you've suggested. But I think just running the test asynchronously would be a simpler temporary solution. The RFQ tests create three new LND nodes and ensures they have the correct channel setup between them. Using one of the standby nodes may lead to unintended future side-effects from channel states or from HTLC interception.

ffranr:
> To bypass the issue in `lnd/lntest`, did you try if re-using the already running standby nodes (Alice and Bob) would help? Then you need to start two fewer nodes. There are `tapd` nodes connected to those, but they should behave pretty independently when it comes to channels and on-chain funds.

I've just tried using standby node Alice as the third node in the RFQ tests. When running one of the RFQ itests in conjunction with the rest of the itests, a LND peer message channel (which the RFQ stream handler is subscribed to) closed unexpectedly.

I'm going to try just running one of the RFQ tests asynchronously to the other itests.

ffranr:
Turns out that I wasn't cleaning up the RFQ test scenario LND nodes correctly. They were living on in the test harness LND node manager. I've corrected that and now the itests pass.

I still think that we should address the `lnd/lntest` node count limitation at one point, but that doesn't matter to tap too much right now. I don't think we'll need >=6 LND nodes simultaneously in a single itest anytime soon.

taprpc/rfqrpc/rfq.proto
```diff
@@ -71,6 +78,27 @@ message AddAssetBuyOrderRequest {
...
+    // asset_specifier is the subject asset.
+    AssetSpecifier asset_specifier = 1;
+
+    // The maximum amount of the asset to sell.
+    uint64 max_asset_amount = 2;
+
+    // The minimum amount of BTC to accept (units: millisats).
+    uint64 min_ask = 3;
+
+    // The unix timestamp after which the order is no longer valid.
```
guggero, line 91: nit: "Unix timestamp in seconds"?

rfq/manager.go
```diff
@@ -288,6 +297,16 @@ func (m *Manager) handleIncomingMessage(incomingMsg rfqmsg.IncomingMsg) error {
 				"request: %w", err)
 		}
 
+	case *rfqmsg.SellAccept:
+		// TODO(ffranr): The stream handler should ensure that the
```
guggero, line 301: What additional information do we need to implement this TODO? Any reason not to do it in this PR?

taprpc/rfqrpc/rfq.proto
```diff
@@ -151,6 +172,14 @@ message PeerAcceptedBuyQuoteEvent {
     PeerAcceptedBuyQuote peer_accepted_buy_quote = 2;
 }
 
+message PeerAcceptedSellQuoteEvent {
+    // Unix timestamp.
```
guggero, line 180: nit: "in microseconds"!

taprpc/rfqrpc/rfq.proto
```diff
@@ -134,6 +134,27 @@ message PeerAcceptedBuyQuote {
...
+    uint64 scid = 3;
+
+    // asset_amount is the amount of the subject asset.
+    uint64 asset_amount = 4;
+
+    // bid_price is the price in millisats for the entire asset amount.
+    uint64 bid_price = 5;
+
+    // The unix timestamp after which the quote is no longer valid.
+    uint64 expiry = 6;
```
guggero, line 155: nit: "in seconds".

rpcserver.go
```diff
@@ -5685,6 +5685,26 @@ func marshallRfqEvent(eventInterface fn.Event) (*rfqrpc.RfqEvent, error) {
...
+			Id:          event.ID[:],
+			Scid:        uint64(event.ShortChannelId()),
+			AssetAmount: event.AssetAmount,
+			BidPrice:    uint64(event.BidPrice),
+			Expiry:      event.Expiry,
+		}
+
+		eventRpc := &rfqrpc.RfqEvent_PeerAcceptedSellQuote{
+			PeerAcceptedSellQuote: &rfqrpc.PeerAcceptedSellQuoteEvent{
+				Timestamp:             uint64(timestamp),
```
guggero, line 5726: Don't we use microseconds in all our events?

itest/rfq_test.go
```diff
@@ -214,6 +215,164 @@ func testRfqAssetBuyHtlcIntercept(t *harnessTest) {
...
+			// tapd node to send a request for quote message to
+			// Bob's node.
+			PeerPubKey: ts.BobLnd.PubKey[:],
+		},
+	)
+	require.NoError(t.t, err, "unable to upsert asset sell order")
+
+	// Wait until Alice receives an incoming sell quote accept message (sent
+	// from Bob) RFQ event notification.
+	waitErr := wait.NoError(func() error {
```
guggero, line 273: I don't think that `wait.NoError()` is the right method here, since that will just re-try (without canceling anything on timeout), but `Recv()` will block until a message arrives.

itest/rfq_test.go
```diff
@@ -214,6 +215,164 @@ func testRfqAssetBuyHtlcIntercept(t *harnessTest) {
...
+	t.Log("Alice paying invoice")
+	ts.AliceLnd.RPC.SendToRouteV2(&routerrpc.SendToRouteRequest{
+		PaymentHash: invoice.RHash,
+		Route:       routeBuildResp.Route,
+	})
+
+	// At this point Bob should have received a HTLC with the asset transfer
+	// specific scid. We'll wait for Bob to publish an accept HTLC event and
+	// then validate it against the accepted quote.
+	waitErr = wait.NoError(func() error {
```
guggero, line 345: Same here. I think a goroutine that sends on a channel, followed by a `select` on the channel with a timeout is probably what we want.

itest/rfq_test.go
```diff
@@ -214,6 +215,164 @@ func testRfqAssetBuyHtlcIntercept(t *harnessTest) {
...
+			)
+			t.Log("Bob has accepted the HTLC")
+			return nil
+		}
+
+		return fmt.Errorf("unexpected event: %v", event)
+	}, defaultWaitTimeout)
+	require.NoError(t.t, waitErr)
+
+	// TODO(ffranr): Confirm that Carol receives the lightning payment from
```
guggero, line 365: Can we implement this TODO in this PR? Or anything still missing?

guggero (commented):
Looks pretty good, very nice work :tada: 
Just a few minor comments

rfq/negotiator.go
```diff
@@ -293,6 +293,57 @@ func (n *Negotiator) HandleIncomingBuyRequest(
...
+		if !sendSuccess {
+			err := fmt.Errorf("negotiator failed to add sell " +
+				"request message to the outgoing messages " +
+				"channel")
+			n.cfg.ErrChan <- err
+			return
+		}
+	}()
+
+	return nil
```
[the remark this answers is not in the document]
ffranr: There's a TODO in this function that will end up with an error return code path:
```
// TODO(ffranr): Ensure that we have a suitable buy offer for the asset
//  that our peer is trying to sell to us.
```

It will end up looking like `HandleIncomingBuyRequest`.

itest/rfq_test.go
ffranr: Can't we just block on the `bobEventNtfns.Recv()` call and depend on the itest timeout rather than using `select`? We expect to receive the event.

rfq/manager.go
ffranr: In this PR I'm just trying to add functionality for the asset sell payment initiation to match what we already have for the asset buy payment initiation. The functionality that this TODO refers to is also missing for the incoming `rfqmsg.BuyAccept` case.

I can add this functionality in a small future PR.

rfq/negotiator.go
```diff
@@ -293,6 +300,67 @@ func (n *Negotiator) HandleIncomingBuyRequest(
...
+		}
+
+		// Construct and send a sell accept message.
+		msg := rfqmsg.NewSellAcceptFromRequest(
+			request, bidPrice, bidExpiry,
+		)
+		sendOutgoingMsg(msg)
+	}()
+
+	return nil
```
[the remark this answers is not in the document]
ffranr: There's a TODO in this function that will end up with an error return code path:
```
// TODO(ffranr): Ensure that we have a suitable buy offer for the asset
//  that our peer is trying to sell to us.
```

It will end up looking like `HandleIncomingBuyRequest`.

itest/rfq_test.go
```diff
@@ -214,6 +215,164 @@ func testRfqAssetBuyHtlcIntercept(t *harnessTest) {
...
+		require.NoError(t.t, err)
+
+		_, ok := event.Event.(*rfqrpc.RfqEvent_PeerAcceptedSellQuote)
+		require.True(t.t, ok, "unexpected event: %v", event)
+
+		return nil
+	}, defaultWaitTimeout)
+	require.NoError(t.t, waitErr)
+
+	// Alice should have received an accepted quote from Bob. This accepted
```
GeorgeTsagk, line 284: What are Bob's acceptance criteria here?

rfqmsg/sell_accept.go
```diff
@@ -0,0 +1,217 @@
...
+	sig [64]byte
+}
+
+// encodeRecords determines the non-nil records to include when encoding at
+// runtime.
+func (q *sellAcceptMsgData) encodeRecords() []tlv.Record {
+	var records []tlv.Record
+
+	// Add id record.
+	records = append(records, TypeRecordSellAcceptID(&q.ID))
```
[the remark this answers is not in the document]
ffranr: Going with just `records()`.

itest/rfq_test.go
GeorgeTsagk: Could just assert the invoice here?

GeorgeTsagk (commented):
Looks good, a few Qs
very nice commit structure

rfq/manager.go
```diff
@@ -427,6 +427,50 @@ func (m *Manager) UpsertAssetBuyOrder(order BuyOrder) error {
...
+	// Peer is the peer that the buy order is intended for. This field is
+	// optional.
+	Peer *route.Vertex
+}
+
+// UpsertAssetSellOrder upserts an asset sell order for management.
+func (m *Manager) UpsertAssetSellOrder(order SellOrder) error {
+	// For now, a peer must be specified.
+	//
+	// TODO(ffranr): Add support for peerless sell orders. The negotiator
```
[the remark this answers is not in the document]
ffranr: The idea here is to (in the future) support submitting a sell order to the RFQ (Request for Quote) service without designating a specific peer. Consequently, the RFQ service would be responsible for identifying the most suitable peer, potentially by soliciting quotes from various peers to ascertain the most favorable offer.

rfq/negotiator.go
```diff
@@ -293,6 +300,67 @@ func (n *Negotiator) HandleIncomingBuyRequest(
 	return nil
 }
 
+// HandleIncomingSellRequest handles an incoming asset sell quote request.
+// This is a request by our peer to sell an asset to us.
+func (n *Negotiator) HandleIncomingSellRequest(
+	request rfqmsg.SellRequest) error {
```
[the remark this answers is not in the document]
ffranr: Going forward, we might send the ask price over to a price oracle so it can take that into account when coming up with a bid price. But since we haven't got a price oracle up and running yet, we're holding off on passing the ask price along for now.

rfq/negotiator.go
```diff
@@ -293,6 +293,57 @@ func (n *Negotiator) HandleIncomingBuyRequest(
...
 
+// HandleOutgoingSellOrder handles an outgoing sell order by constructing sell
+// requests and passing them to the outgoing messages channel. These requests
+// are sent to peers.
+func (n *Negotiator) HandleOutgoingSellOrder(order SellOrder) error {
+	// Query the price oracle for a reasonable ask price. We perform this
+	// query and response handling in a separate goroutine in case it is a
+	// remote service and takes a long time to respond.
+	n.Wg.Add(1)
+	go func() {
```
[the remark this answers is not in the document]
ffranr: `queryAskFromPriceOracle` is supposed to be the function for the oracle-asking logic. The rest of the code creates a new request and puts it on the outgoing channel.

This is where I think something like this might come in handy:
```golang
func execGoRoutine(cg *fn.ContextGuard, errChan chan<- error,
    exec func() error) error {

    // Ensure that the error channel has a buffer of at least 1.
    if cap(errChan) == 0 {
        return fmt.Errorf("error channel has no buffer")
    }

    cg.Wg.Add(1)
    go func() {
        defer cg.Wg.Done()

        err := exec()
        if err != nil {
            sendSuccess := fn.SendOrQuit(errChan, err, cg.Quit)
            if !sendSuccess {
                log.Warn("Failed to send error on error "+
                    "channel: %v", err)
            }
        }
    }()
    
    return nil
}
```

I'll see if i can refactor later. Is this nit blocking for you?

itest/rfq_test.go
ffranr: We don't have an acceptance criteria for Bob right now. Bob will just accept any offer.

There's a TODO in the `Negotiator` which will give Bob a criteria:
```
	// TODO(ffranr): Ensure that we have a suitable buy offer for the asset
	//  that our peer is trying to sell to us.
```

itest/rfq_test.go
ffranr: hmm I can't get this to work for some reason. Route is: `Alice -> Bob -> Carol`, the HTLC does seem to reach Bob. And Bob does accept the intercepted HTLC. But then the send function returns:
```
Failure_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (1)
```

And so when i lookup the invoice from Carol, it isn't settled.

ffranr: I've attempted to disable Bob's tap HTLC interception but the payment still doesn't succeed. Any ideas @guggero @GeorgeTsagk ?

ffranr: I've got the payment to succeed.

rfq/negotiator.go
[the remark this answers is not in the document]
guggero: Makes sense :+1:

rfq/manager.go
guggero: Sounds good to me.

rfqmsg/sell_request.go
```diff
@@ -0,0 +1,274 @@
...
+	)
+	records = append(
+		records, TypeRecordSellRequestAskPrice(&q.AskPrice),
+	)
+
+	return records
+}
+
+// Encode encodes the structure into a TLV stream.
+func (q *sellRequestMsgData) Encode(writer io.Writer) error {
```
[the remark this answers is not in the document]
ffranr: I've added basic encode/decode unit tests for the new message types.

ffranr:
Some of the review comments also apply to existing code that the PR doesn't touch. I've made a note of them and I'll create another PR address those code changes.

itest/rfq_test.go
guggero: Yes, but then we'd need to initiate `ts.AliceTapd.SubscribeRfqEventNtfns` with a timeout context. Otherwise it will just block forever, since there's nothing to time out (other than eventually `go test` timing out, which currently is set at 20 minutes with `-test.timeout`).

But I think the `BeforeTimeout` function works too!

guggero (approved):
Great work! LGTM pending unit tests for encode/decode of messages.

rfq/negotiator.go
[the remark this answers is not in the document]
GeorgeTsagk: nop, looks good!

GeorgeTsagk (approved):
Looks good!

are the encode/decode unit tests landing in a diff PR?
otherwise lgtm

ffranr:
@GeorgeTsagk thanks for taking a second look.

I've added these unit tests for encode/decode:
https://github.com/lightninglabs/taproot-assets/pull/835/files#diff-26f6b4e4b5386817099becddf822524f93feec1169c2f0901380dabb7c25d740

https://github.com/lightninglabs/taproot-assets/pull/835/files#diff-4ee5dc5905f2ad2d13635df68e16d1881e3ceedde4ab1f91529d7b95f0c7396a

Issue #806, opened 2024-02-22
[bug]: Empty events for addrs receives


vanditshah99:
We are receiving empty addrs receives for multiple tapd address.
Example tapd address : taptb1qqqsqqspqqzzqzz8zn67pdj5sllk6ugunea3ew2wg03nxm3k4z4sty4mdzy825h7qcss9czp7gtxq8s3d30l4hufunm7cwx67fctyhzsg2yufwum6lapheylpqssxv4lttmctkkuur6wjws6p35vaupcqk9pgfgar6fv54n6jt3lhjdapgqscrpkw4hxjan9wfek2unsvvaz7tm5v4ehgmn9wsh82mnfwejhyum99ekxjemgw3hxjmn89enxjmnpde3k2w33xqcrywghy0ewy

Address Generation Logs

```
2024-02-22 09:44:25.879 [INF] RPCS: [NewAddr]: making new addr: asset_id=084714f5e0b65487ff6d711c9e7b1cb94e43e3336e36a8ab0592bb68887552fe, amt=12
2024-02-22 09:44:25.924 [DBG] TAPD: Deriving new key for fam_family=212
2024-02-22 09:44:25.954 [DBG] TAPD: Deriving new key for fam_family=212
```

And Address Send Logs 

```
024-02-22 09:45:11.288 [INF] FRTR: Received to send request to 1 addrs: [TapAddr{id=084714f5e0b65487ff6d711c9e7b1cb94e43e3336e36a8ab0592bb68887552fe, amount=12, script_key=02e041f216601e116c5ffadf89e4f7ec38daf270b25c504289c4bb9bd7fa1be49f}]
2024-02-22 09:45:11.288 [INF] FRTR: ChainPorter executing state: SendStateVirtualCommitmentSelect
2024/02/22 09:45:12 http: TLS handshake error from 10.0.1.84:36685: EOF
2024-02-22 09:45:12.408 [INF] FRTR: Identified 39 eligible asset inputs for send of 12 to 084714f5e0b65487ff6d711c9e7b1cb94e43e3336e36a8ab0592bb68887552fe
2024-02-22 09:45:12.449 [INF] FRTR: Selected 1 asset inputs for send of 12 to 084714f5e0b65487ff6d711c9e7b1cb94e43e3336e36a8ab0592bb68887552fe
2024-02-22 09:45:12.509 [DBG] TAPD: Deriving new key for fam_family=212
[...]

Issue #781, opened 2024-02-01
[bug]: streamline event completion on custodian reboot / start up


dstadulis:
### Background

send-complete events that have occurred while tapd is offline aren't ever reported as completed

### Steps to reproduce

Untested observation from inspecting goroutines with @ffranr 

### Expected behavior

send complete events that have occurred after tapd comes back online after being offline

https://github.com/lightninglabs/taproot-assets/blob/1ca17f6faa8edcaccd7e5e3d7c4d638f8ca2c988/tapgarden/custodian.go#L279-L285

This commit contains TODOs that relate to streamlining receive event completion: https://github.com/lightninglabs/taproot-assets/pull/773/commits/b07b74f21dfd275aa78d3be5200800930151ae48

ffranr:
Another change that we should make at the same time is to change this into a `>` rather than a `==`:
https://github.com/lightninglabs/taproot-assets/blob/1ca17f6faa8edcaccd7e5e3d7c4d638f8ca2c988/tapgarden/custodian.go#L445

The confirmations field can be negative, uses signed integer type.

Issue #739, opened 2023-12-14
[feature]: Custodian should emit an RPC notification status event when the recieve process starts


ffranr:
The custodian should emit an RPC notification status event when the receive process starts. The event should be published to subscribers once a wallet transaction has been successfully associated with an address. Which happens here:

https://github.com/lightninglabs/taproot-assets/blob/7c0688e3bf84c2b9aeb4be3417cc766706ddaba7/tapgarden/custodian.go#L392-L392

Note that we already publish a transaction complete event to subscribers, see completed issue https://github.com/lightninglabs/taproot-assets/issues/650 .

This is where the custodian publishes that transaction complete event:

https://github.com/lightninglabs/taproot-assets/blob/7c0688e3bf84c2b9aeb4be3417cc766706ddaba7/tapgarden/custodian.go#L497-L497

This feature was requested on Slack: https://lightningcommunity.slack.com/archives/C03B3556HQ8/p1702546986057129

RajGoodF:
Yes much needed !

dstadulis:
## Receive-notification feature, acceptance criteria

tapd would emit a receive event
1. After a transaction has been initiated even if  the transaction hasn't yet been included in a block.
2. After the transaction is included in a block
3. After the block the transaction had been included in, is ejected from main chain. (reorged out)

Issue #822, opened 2024-03-05
[feature]: 


lukegao209:
**Is your feature request related to a problem? Please describe.**

SubscribeReceiveAssetEventNtfns : add  time filter in request params
SubscribeSendAssetEventNtfns : add time filter in request ; add addr and txhash in response

**Describe the solution you'd like**
I think SubscribeReceiveAssetEventNtfns need at least a time filter for subscribe or it will received all the events when the Subscribe service restart.

there is no addr and txhash info in SubscribeSendAssetEventNtfns response.
 I can do nothing after receiving the send events as there is not addr and txhash info , for example :  check the send event on chain; send notification to user

The features are very important for projects to support their users to send and receives.

Nsandomeno:
Having the `addr` on send notifications alone seems like it would be the quickest way to enable an application to link the event to a domain-specific transaction, account, or user database resource.

Very curious about any thoughts on this or other more secure/robust approaches.

guggero:
@lukegao209 The two notification RPCs currently do not give you events from past events, only new events from the time you subscribe. Are you requesting that you can retrieve events from the past, starting at a certain time?
Or do you mean you'd only like to receive events for addresses that were created a certain time ago? For example "only give me receive events for addresses that were created today"?

I'm currently re-designing the event APIs to be way more useful (not just for our integration tests).

guggero:
> The two notification RPCs currently do not give you events from past events

Okay, I just noticed this isn't always true. On startup of the daemon, if you already subscribe quickly enough, you get notifications about previously completed events. I'll attempt to fix that. Already completed events shouldn't re-notify.

lukegao209:
> @lukegao209 The two notification RPCs currently do not give you events from past events, only new events from the time you subscribe. Are you requesting that you can retrieve events from the past, starting at a certain time? Or do you mean you'd only like to receive events for addresses that were created a certain time ago? For example "only give me receive events for addresses that were created today"?
> 
> I'm currently re-designing the event APIs to be way more useful (not just for our integration tests).

it would be great to subscribe starting at a certain time

dstadulis:
@lukegao209 https://github.com/lightninglabs/taproot-assets/pull/853 should satisfy your need -- If you'd like to review for feature acceptance, that'd be great

PR #836, opened 2024-03-14
[events API 1/2]: Add new send and receive streaming RPCs

guggero:
Replaces https://github.com/lightninglabs/taproot-assets/pull/773.

Fixes https://github.com/lightninglabs/taproot-assets/issues/781.
Fixes https://github.com/lightninglabs/taproot-assets/issues/806.
Fixes https://github.com/lightninglabs/taproot-assets/issues/739.

This PR adds new RPCs for being notified about inbound and outbound asset transfers.
The existing `SubscribeSendAssetEventNtfns` and `SubscribeReceiveAssetEventNtfns` were mostly built for asserting the internal state during integration tests and weren't super helpful for app builders. That's why these old RPCs were moved to the `tapdevrpc` package and are now only used in the integration tests.

The new RPCs notify on all send and receive states instead of only a few of them. The events also contain way more information that's useful to map them to the user action. And both RPCs allow the events to be filtered (either by address for receive events and by script key for send events).

This is part 1 of the "events API" saga. The follow-up PRs will base on this one and address:
 - Allow to be notified about past events (e.g. reading events from the database back to a certain specified date): https://github.com/lightninglabs/taproot-assets/issues/822
 - Add event RPC for the minting process
 - Notify on send errors: https://github.com/lightninglabs/taproot-assets/issues/254


guggero:
> BTW, CI failed on RPC and lint check.

Fixed!

itest/assertions.go
```diff
@@ -766,6 +770,114 @@ func AssertAddrEventByStatus(t *testing.T, client taprpc.TaprootAssetsClient,
...
+			return
+		}
+
+		startStatus++
+	}
+}
+
+// AssertSendEvents makes sure all events with incremental status are sent
+// on the stream for the given script key.
+func AssertSendEvents(t *testing.T, scriptKey []byte,
```
jharveyb, line 813: nit: Seems like this is almost always used for states `StoreProofs` to `Complete`? Could wrap that in another assertion. Non-blocking / optional.

itest/send_test.go
```diff
@@ -1023,36 +1029,48 @@ func testOfflineReceiverEventuallyReceives(t *harnessTest) {
 // assertAssetSendNtfsEvent asserts that the given asset send event notification
 // was received. This function will block until the event is received or the
 // event stream is closed.
-func assertAssetSendNtfsEvent(t *harnessTest, ctx context.Context,
-	eventNtfns taprpc.TaprootAssets_SubscribeSendAssetEventNtfnsClient,
-	targetEventSelector func(*taprpc.SendAssetEvent) bool,
+func assertAssetSendNtfsEvent(t *harnessTest,
```
jharveyb, line 1032: This func and `assertAssetRecvNtfsEvent` look very similar,  and can be combined with a type constraint + parameter:

```go
type AssetNtfsEvent interface {
	tapdevrpc.SendAssetEvent | tapdevrpc.ReceiveAssetEvent
}

func assertAssetNtfsEvent[T AssetNtfsEvent](t *harnessTest,
	stream *EventSubscription[*T], timeout time.Duration,
	targetEventSelector func(*T) bool, expectedCount int) {
```

From local testing this actually works without specifying the type at the callsites which is kind of neat. Though the timeout arg has to be reordered for these two to match.

itest/send_test.go
```diff
@@ -1023,36 +1029,48 @@ func testOfflineReceiverEventuallyReceives(t *harnessTest) {
...
+		case <-stream.Context().Done():
+			require.NoError(t.t, stream.Context().Err())
 
-		if countFound == expectedCount {
 			break
+		default:
 		}
 
-		event, err := eventNtfns.Recv()
+		if countFound == expectedCount {
```
jharveyb, line 1063: missing `close(success)` IIUC

rpcserver.go
```diff
@@ -2544,226 +2574,478 @@ func marshalOutputType(outputType tappsbt.VOutputType) (taprpc.OutputType,
...
+		default:
+			return false, fmt.Errorf("unknown event type: %T", e)
+		}
 
-		// Handle the case where the RPC server is shutting down.
-		case <-r.quit:
-			return nil
+		// If we're not filtering on a specific address, we return all
+		// events.
+		if addrString == "" {
```
jharveyb, line 2693: super-nit: could be a one-line with `(addrString == "" || eventAddrString == addrString), nil`

tapdb/addrs.go
```diff
@@ -890,6 +924,52 @@ func fetchEvent(ctx context.Context, db AddrBook, eventID int64,
 	}, nil
 }
 
+// fetchEventByOutpoint fetches a single address event identified by its address
+// and outpoint.
+func fetchEventByOutpoint(ctx context.Context, db AddrBook,
+	addr *address.AddrWithKeyInfo, outpoint wire.OutPoint) (*address.Event,
+	error) {
+
+	dbEvent, err := db.FetchAddrEventByAddrKeyAndOutpoint(
```
jharveyb, line 933: We could return a concrete err to the caller if the query returns `ErrNoRows`, which it should when there's no match since it specifies `:one`.

tapdb/sqlc/queries/addrs.sql
```diff
@@ -105,6 +105,32 @@ LEFT JOIN internal_keys
...
 WHERE id = $1;
 
+-- name: FetchAddrEventByAddrKeyAndOutpoint :one
+WITH target_addr(addr_id) AS (
+    SELECT id
+    FROM addrs
+    WHERE addrs.taproot_output_key = $1
+)
+SELECT
+    addr_events.id, creation_time, status, asset_proof_id, asset_id,
```
jharveyb, line 115: Why expose `addr_events.id` here? I didn't spot a use of it later on (may be in some follow-on PR?)

tapgarden/custodian.go
```diff
@@ -582,6 +587,18 @@ func (c *Custodian) mapToTapAddr(walletTx *lndclient.Transaction,
 		return nil, fmt.Errorf("unable to encode address: %w", err)
 	}
 
+	// Skip already completed events.
+	ctxt, cancel = c.CtxBlocking()
+	existingEvent, err := c.cfg.AddrBook.QueryEvent(ctxt, addr, op)
+	cancel()
+	if err == nil && existingEvent.Status >= address.StatusCompleted {
```
jharveyb, line 594: I think we end up dropping the err here? We could safely ignore the `ErrNoRows` equivalent though.

rpcserver.go
```diff
@@ -2544,226 +2574,478 @@ func marshalOutputType(outputType tappsbt.VOutputType) (taprpc.OutputType,
...
+	// Create a new event subscriber and pass a copy to the event source.
+	// We will then read events from the subscriber.
 	eventSubscriber := fn.NewEventReceiver[fn.Event](fn.DefaultQueueSize)
 	defer eventSubscriber.Stop()
 
-	// Register the subscriber with the AssetCustodian.
-	err := r.cfg.AssetCustodian.RegisterSubscriber(
-		eventSubscriber, false, false,
-	)
+	var unused T
```
jharveyb, line 2779: nit: all small comment to explain why this isn't `nil` could be useful, took me a sec to figure it out.

jharveyb (commented):
Looks great! Seems to expose all the relevant info. Just had a few notes + nits.

itest/assertions.go
guggero: Makes sense, done.

rpcserver.go
guggero: True. But I try to avoid chaining too many boolean operators together, as they can sometimes be hard to read or understand at a quick glance. I'd rather e a bit more explicit here.

tapdb/addrs.go
guggero: Fair point! Also added this method to the unit tests.

tapdb/sqlc/queries/addrs.sql
guggero: It's part of the `address.Event` struct. It looks like it's currently only used in unit tests... Didn't want to also refactor them, as the PR was already growing quite a bit. But it looks like we could probably un-export the `address.Event.ID` field.

rpcserver.go
guggero: Makes sense, added.

jharveyb (approved):
Re-reviewed, LGTM! Rebase looks correct, I think it needs another.

tapgarden/custodian.go
```diff
@@ -579,6 +579,28 @@ func (c *Custodian) mapToTapAddr(walletTx *lndclient.Transaction,
 		return nil, fmt.Errorf("unable to encode address: %w", err)
 	}
 
+	// Skip already completed events.
```
Roasbeef, line 590: So this'll suppress duplicate notification events?

tapfreighter/chain_porter.go
```diff
@@ -1199,25 +1198,59 @@ func (p *ChainPorter) publishSubscriberEvent(event fn.Event) {
...
+func newAssetSendEvent(pkg sendPackage) *AssetSendEvent {
+	newSendEvent := &AssetSendEvent{
 		timestamp: time.Now().UTC(),
-		SendState: state,
+		SendState: pkg.SendState,
+		// The parcel remains static throughout the state machine, so we
+		// don't need to copy it, there can be no data race.
+		Parcel:         pkg.Parcel,
+		VirtualPackets: fn.CopyAll(pkg.VirtualPackets),
+		PassivePackets: fn.CopyAll(pkg.PassiveAssets),
```
Roasbeef, line 1260: 😎

taprpc/taprootassets.proto
```diff
@@ -999,3 +1006,23 @@ message OutPoint {
     */
     uint32 output_index = 2;
 }
+
+message SubscribeReceiveEventsRequest {
+    // Filter receives by a specific address. Leave empty to get all receive
+    // events for all addresses.
+    string filter_addr = 1;
```
Roasbeef, line 1020: What about exposing the event ID and/or timestamp to allow users to recv a backlog of events?

taprpc/taprootassets.proto
```diff
@@ -999,3 +1006,23 @@ message OutPoint {
...
+}
+
+message ReceiveEvent {
+    // Event creation timestamp.
+    int64 timestamp = 1;
+
```
Roasbeef, line 1026: Block height + hash?

rpcserver.go
```diff
@@ -3185,6 +3189,91 @@ func (r *rpcServer) SubscribeReceiveAssetEventNtfns(
...
+		// If we're not filtering on a specific address, we return all
+		// events.
+		if addrString == "" {
+			return true, nil
+		}
+
+		return eventAddrString == addrString, nil
+	}
+
+	return handleEvents[bool, *taprpc.ReceiveEvent](
```
Roasbeef, line 3276: Could also consider a another refactor where we make a new interface for the filter+marshal methods, so then the event handler accepts that struct/impl instead of these function closures. 

Non-blocking.

Roasbeef (changes requested):

tapgarden/custodian.go
guggero: Yes. Otherwise if we have a pending event at block X we would get notifications for all events between block X and now (even if they are already complete).

taprpc/taprootassets.proto
guggero: Yes, we'll do that in part 2 of the series.

taprpc/taprootassets.proto
guggero: We don't keep the block hash around in the event, so we wouldn't have that at all times. But I added the block height, maybe that's enough for now?

rpcserver.go
guggero: Do you mean a method on the actual event object (for example `tapgarden.AssetReceiveEvent`) itself so it knows how to serialize itself to RPC?
Or a new struct in the RPC server, something like `receiveEventFilterMarshaler`?
I'll consider that for part, but I feel like things are already pretty compact and readable the way they are now.

jharveyb:
Re-reviewed after rebases, still LGTM.

taprpc/taprootassets.proto
Roasbeef: Ah makes sense, yeah I think that's a good stop gap.

rpcserver.go
Roasbeef: So basically morph `handleEvents` into a a struct that can accept the generic params of filter type, marshaller, etc.

Roasbeef (approved):
LGTM 🦊

PR #845, opened 2024-03-19
RFQ refactor towards adding asset sell request support

ffranr:
Towards fixing https://github.com/lightninglabs/taproot-assets/issues/818

This PR contains commits which refactor the RFQ service towards solving https://github.com/lightninglabs/taproot-assets/issues/818 .

The changes in this PR shouldn't change the functionality of the RFQ service. They should rename methods, functions, types, and variables so that the changes necessary for https://github.com/lightninglabs/taproot-assets/issues/818 are easier to make in a different PR.


rpcserver.go
```diff
@@ -4999,18 +5001,20 @@ func marshalAcceptedQuotes(
 	return rpcQuotes
 }
 
-// QueryRfqAcceptedQuotes queries for accepted quotes from the RFQ system.
-func (r *rpcServer) QueryRfqAcceptedQuotes(_ context.Context,
-	_ *rfqrpc.QueryRfqAcceptedQuotesRequest) (
-	*rfqrpc.QueryRfqAcceptedQuotesResponse, error) {
+// QueryRfqPeerAcceptedQuotes is used to query for quotes that were requested by
+// our node and have been accepted our peers.
+func (r *rpcServer) QueryRfqPeerAcceptedQuotes(_ context.Context,
```
guggero, line 5006: nit: since we're renaming: This is located in the `rfqrpc` package, within the `Rfq` service. Do we really need the `Rfq` in this name as well? I think we can shorten some RPC names if we consider that there's always the package name context to help.

taprpc/rfqrpc/rfq.proto
```diff
@@ -106,19 +106,21 @@ message AcceptedQuote {
     uint64 expiry = 6;
 }
 
-message QueryRfqAcceptedQuotesResponse {
-    repeated AcceptedQuote accepted_quotes = 1;
+message QueryRfqPeerAcceptedQuotesResponse {
```
guggero, line 109: Same here, IMO we can remove the `Rfq` part from all message names.

rfqmsg/messages.go
```diff
@@ -17,8 +21,15 @@ func (id ID) String() string {
 	return hex.EncodeToString(id[:])
 }
 
-// SerialisedScid is a serialised short channel id (SCID).
-type SerialisedScid uint64
+func (id ID) Scid() SerialisedScid {
```
guggero, line 24: nit: missing Godoc.

chain_bridge.go
```diff
@@ -87,13 +87,13 @@ func (l *LndRpcChainBridge) GetBlock(ctx context.Context,
 func (l *LndRpcChainBridge) GetBlockHeader(ctx context.Context,
 	hash chainhash.Hash) (*wire.BlockHeader, error) {
 
-	header, err := l.lnd.ChainKit.GetBlockHeader(ctx, hash)
+	block, err := l.lnd.ChainKit.GetBlock(ctx, hash)
```
guggero, line 90: Why is this change needed? This is actually a performance regression, because fetching the full block is always more expensive than fetching the header (especially for Neutrino).

guggero (changes requested):
Changes look good with the exception of the last commit (see inline comment).

chain_bridge.go
ffranr: I've removed this change, thanks for pointing it out. I think I made this change because I was originally using a local lndclient repo.

rpcserver.go
ffranr: I'll shorten the name.

I included the RFQ package name because I wanted to make sure that the method would be unique in the `rpcServer` struct method namespace also.

guggero: That makes sense. Though I think if we ever start to experience naming collisions in the RPC server, then it's time to split that code up into several structs to mirror the packages they implement.

rfqmsg/messages.go
```diff
@@ -17,8 +21,16 @@ func (id ID) String() string {
 	return hex.EncodeToString(id[:])
 }
 
-// SerialisedScid is a serialised short channel id (SCID).
-type SerialisedScid uint64
+// Scid returns the short channel id (SCID) of the RFQ message.
+func (id ID) Scid() SerialisedScid {
+	// Given valid RFQ message id, we then define a RFQ short chain id
```
GeorgeTsagk, line 26:
```suggestion
	// Given valid RFQ message id, we then define a RFQ short channel id
```

rfqmsg/messages.go
```diff
@@ -17,8 +21,16 @@ func (id ID) String() string {
...
+	// (SCID) by taking the last 8 bytes of the RFQ message id and
+	// interpreting them as a 64-bit integer.
+	scidBytes := id[24:]
```
GeorgeTsagk, line 29: nit
```suggestion
	scidBytes := id[len(id)-8]
```

GeorgeTsagk (approved):
LGTM :zap: 
few nits

rfqmsg/messages.go
ffranr: I don't think your suggesting works. We want to select the last 8 elements from the `id` array, not the single element at `len(id)-8`.

ffranr:
Last commit (`lndclient` version bump) wasn't necessary after rebase.

Issue #254, opened 2023-02-03
Async asset send procedure reports errors via event notification stream.


ffranr:
Following on from PR https://github.com/lightninglabs/taro/pull/222 , the asset send procedure will be entirely async. Consequently, if the process is initiated from a client via RPC, that call will not return any potential error.

The work to be done: the [event streamer](https://github.com/lightninglabs/taro/blob/82daf05d4fd50d14e47010365c14a1416e806350/rpcserver.go#L1221) should support streaming error event messages which relate to the asset send procedure.

PR #853, opened 2024-03-22
[events API 2/2]: Add new mint streaming RPC, notify on errors, replay previous send/receive events

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/836.

Fixes https://github.com/lightninglabs/taproot-assets/issues/822.
Fixes https://github.com/lightninglabs/taproot-assets/issues/254.


taprpc/taprootassets.proto
```diff
@@ -1018,6 +1018,10 @@ message SubscribeReceiveEventsRequest {
     // Filter receives by a specific address. Leave empty to get all receive
     // events for all addresses.
     string filter_addr = 1;
+
+    // The start time as a unix timestamp in seconds. If not set (default value
+    // 0), the daemon will start streaming events from the current time.
+    uint64 start_time = 2;
```
Roasbeef, line 1024: Ah ok, it's in this PR. 

I do wonder if a monotonic event ID will be easier for devs to handle than a timestamp though.

guggero: I personally think a timestamp is more useful, as it contains some information about the records by itself already. And since we're using a timestamp with microsecond precision, we effectively get monotonically increasing values as well.
Happy to add the ID as well though. But we'd need to also change the event subscription interface for the `deliverFrom` parameter, which in this case would need to be `fn.Either[time.Time, uint64]`.

tapgarden/caretaker.go
```diff
@@ -195,8 +198,17 @@ func (b *BatchCaretaker) Cancel() error {
 		if err != nil {
 			err = fmt.Errorf("BatchCaretaker(%x), batch state(%v), "+
 				"cancel failed: %w", batchKey, batchState, err)
+
+			b.cfg.PublishMintEvent(newAssetMintErrorEvent(
+				cancelResp.err, BatchStateSeedlingCancelled,
```
jharveyb, line 203: P sure this should just be `err`, as `cancelResp` is only populated at the end of this scope?

tapgarden/caretaker.go
```diff
@@ -207,8 +219,17 @@ func (b *BatchCaretaker) Cancel() error {
 		if err != nil {
 			err = fmt.Errorf("BatchCaretaker(%x), batch state(%v), "+
 				"cancel failed: %w", batchKey, batchState, err)
+
+			b.cfg.PublishMintEvent(newAssetMintErrorEvent(
+				cancelResp.err, BatchStateSproutCancelled,
```
jharveyb, line 224: Same note here re: err being returned.

tapgarden/caretaker.go
guggero: Oops, left over from previous iteration. You're right of course, fixed. Same below.

jharveyb:
Re-reviewed after rebases, still LGTM.

Roasbeef (approved):
LGTM ⛷️

Roasbeef:
Docker build check looks to have failed here, but afaict was some internal error, and we don't get the full logs here for some reason.

Roasbeef:
Looks like this failed the new linter directives (merge queue failure): 
```
level=info msg="Memory: 426 samples, avg is 766.2MB, max is 2346.4MB"
level=info msg="Execution took 47.585519618s"
							NewAssetReceiveErrorEvent(
tapgarden/custodian.go:479: line is 89 characters (lll)
								event.ConfirmationHeight,
tapgarden/custodian.go:1066: line is 81 characters (lll)
			receiver.NewItemCreated.ChanIn() <- NewAssetReceiveEvent(
```

PR #854, opened 2024-03-22
tapd: temporarily increase query stats cache from 80k to 200k

Roasbeef:
We have a bit over 130k assets on mainnet now, so 80k isn't large enough
to cache the entire response assuming no pagination.

We also start to log cache insertion failures.

### DIFF
--- a/tapdb/multiverse.go
+++ b/tapdb/multiverse.go
@@ -201,7 +201,9 @@ func (p *proofCache) insertProof(id universe.Identifier,
 		id.StringForLog(), leafKey, proofKey[:])
 
 	proofVal := cachedProof(proof)
-	_, _ = assetProofCache.Put(proofKey, &proofVal)
+	if _, err := assetProofCache.Put(proofKey, &proofVal); err != nil {
+		log.Errorf("unable to insert into proof cache: %v", err)
+	}
 }
 
 // delProofsForAsset deletes all the proofs for the given asset.
@@ -350,7 +352,10 @@ func (r *rootNodeCache) cacheRoots(q universe.RootNodesQuery,
 
 	// Store the main root pointer, then update the root index.
 	rootPageCache := r.allRoots.Load()
-	_, _ = rootPageCache.Put(newRootPageQuery(q), rootNodes)
+	_, err := rootPageCache.Put(newRootPageQuery(q), rootNodes)
+	if err != nil {
+		log.Errorf("unable to insert into root cache: %v", err)
+	}
 
 	rootIndex := r.rootIndex.Load()
 	for _, rootNode := range rootNodes {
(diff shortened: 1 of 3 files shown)


tapdb/multiverse.go
```diff
@@ -201,7 +201,9 @@ func (p *proofCache) insertProof(id universe.Identifier,
 		id.StringForLog(), leafKey, proofKey[:])
 
 	proofVal := cachedProof(proof)
-	_, _ = assetProofCache.Put(proofKey, &proofVal)
+	if _, err := assetProofCache.Put(proofKey, &proofVal); err != nil {
+		log.Errorf("unable to insert into proof cache: %v", err)
```
guggero, line 205: micro-nit: Start sentences uppercase for log messages?

tapdb/schemas.go
```diff
@@ -5,5 +5,5 @@ import (
 	_ "embed"
 )
 
-//go:embed sqlc/migrations/*.up.sql
+//go:embed sqlc/migrations/*.*.sql
```
guggero, line 8: Nice catch!

guggero (approved):
Thanks, LGTM :tada:

tapdb/multiverse.go
```diff
@@ -482,11 +487,19 @@ func (u *universeLeafCache) cacheLeafKeys(q universe.UniverseLeafKeysQuery,
 		}
 
 		// Store the cache in the top level cache.
-		_, _ = u.leafCache.Put(idStr, pageCache)
+		if _, err := u.leafCache.Put(idStr, pageCache); err != nil {
```
ffranr, line 490: Above, we're attempting to get `pageCache` from `u.leafCache`. And at this point, if `err != nil`, we're attempting to construct a new `pageCache` and put it in the `u.leafCache`.

I think that we should have a more specific error check than `err != nil` above. We should check for `cache.ErrElementNotFound` when first attempting to get `pageCache` from `u.leafCache`. If the error only relates to a missing element, then we can legitimately try to put the missing element in `u.leafCache`. But if the error is anything else, then I don't think we should try to put into `u.leafCache` at this point.

tapdb/multiverse.go
```diff
@@ -482,11 +487,19 @@ func (u *universeLeafCache) cacheLeafKeys(q universe.UniverseLeafKeysQuery,
...
+			// If we encounter an error here, we'll exit to avoid a
+			// panic below.
+			log.Errorf("unable to store entry in page cache: %v",
+				err)
```
ffranr, line 494: This error message can perhaps be more accurately
```
log.Errorf("Unable to store new page cache in leaf cache: %v", err)
```

tapdb/universe_stats.go
```diff
@@ -728,7 +730,9 @@ func (u *UniverseStats) QueryAssetStatsPerDay(ctx context.Context,
 	}
 
 	// We have a fresh result, so we'll cache it now.
-	_, _ = u.assetEventsCache.Put(query, results)
+	if _, err := u.assetEventsCache.Put(query, results); err != nil {
+		log.Errorf("unable to query events cache: %v", err)
```
ffranr, line 734: I don't think the error message is correct here. The `Put` method call is not trying to query the events cache. Perhaps the error message should be:

```
log.Errorf("Unable to populate assets events cache: %v", err)
```

tapdb/universe_stats.go
```diff
@@ -155,7 +155,7 @@ type assetEventsCache = *lru.Cache[eventQuery, cachedAssetEvents]
 
 // statsQueryCacheSize is the total number of asset query responses that we'll
 // hold inside the cache.
-const statsQueryCacheSize = 80_000
+const statsQueryCacheSize = 200_000
```
ffranr, line 158: At 200k, I think the standby memory won't exceed 10 Mb (based on the cache key data structure size).

GeorgeTsagk:
Would it make sense to paginate the call? It's already been added in other similar calls

dstadulis:
!lightninglabs-deploy mute 720h30m

dstadulis:
!lightninglabs-deploy mute 360h

Issue #841, opened 2024-03-19
[bug]: Can't generate addr after sync asset from other universe


lukegao209:
### Background
Alice(Tapd V0.3.0) ,Bob(Tapd V0.3.3)
Step1 : Bob => excute cmd : tapcli -n mainnet universe sync --universe_host

            {
    "synced_universes":  [
        {
            "old_asset_root":  {
                "id":  null,
                "mssmt_root":  null,
                "asset_name":  "",
                "amounts_by_asset_id":  {}
            },
            "new_asset_root":  {
                "id":  {
                    "asset_id":  "475a642ed13bc44af6490c8571404988d7b514386cf8f3a603d04a0d2fa9f8f5",
                    "proof_type":  "PROOF_TYPE_ISSUANCE"
                },
                "mssmt_root":  {
                    "root_hash":  "f6b42e0cb6cf9de9cc83adfc645ec7d751a12bc402cbdcc2175750a6881c9461",
                    "root_sum":  "2100000000"
                },
                "asset_name":  "",
                "amounts_by_asset_id":  {}
            },
            "new_asset_leaves":  []
        }
    ]
}
Step2 : Bob=> tapcli -n mainnet addrs new --amt 11 --asset_id xxxx

`[tapcli] unable to make addr: rpc error: code = Unknown desc = unable to make new addr: unable to make address for unknown asset xxxx: asset group is unknown`

The asset is a normal type without group, but it fails to generate a new addrs.
I think the 'asset group is unknown' is a correct status,but it show as error.
The check code should be here in book.go line 252(V.0.3.3)

```
// NewAddress creates a new Taproot Asset address based on the input parameters.
func (b *Book) NewAddress(ctx context.Context, assetID asset.ID, amount uint64,
	tapscriptSibling *commitment.TapscriptPreimage,
	proofCourierAddr url.URL, addrOpts ...NewAddrOpt,
) (*AddrWithKeyInfo, error) {

	// Before we proceed and make new keys, make sure that we actually know
	// of this asset ID, or can import it.
	if _, err := b.queryAssetInfo(ctx, assetID); err != nil {
		return nil, fmt.Errorf("unable to make address for unknown "+
			"asset %x: %w", assetID[:], err)
	}

	rawScriptKeyDesc, err := b.cfg.KeyRing.DeriveNextTaprootAssetKey(ctx)
	if err != nil {
		return nil, fmt.Errorf("unable to gen key: %w", err)
	}
```
[...]

PR #856, opened 2024-03-25
cmd/tapcli: allow proof file or single proof on insert

guggero:
Related to https://github.com/lightninglabs/taproot-assets/issues/841.

In order to be able to directly import a proof leaf that was previously exported with `tapcli universe leaves` directly into the universe, we allow the proof to either be a proof file or a single transition proof.


ffranr (approved):

jharveyb (approved):
LGTM, nice catch!

Was going to suggest checked against our max filesize for proofs but it's already part of `proof.File.Decode()`.

Issue #852, opened 2024-03-21
[feature]: Upgrade freighter's send coin-selection method to use new FundPsbt method 


dstadulis:
https://github.com/lightninglabs/taproot-assets/issues/118#issuecomment-1978593804

- [ ] Use the new coin selection funding mode added to lnd's FundPsbt method in the freighter's send 
- [ ] Either make lnd v0.18.0 required or gate the use of the new funding mode on that version being available

PR #858, opened 2024-03-26
Use new `WalletKit.FundPsbt` coin selection option

GeorgeTsagk:
## Description

This PR bumps the lnd dependency in order to acquire the recently added coin selection functionality in the `FundPsbt` call, and also uses the new option in the tapd `FundPsbt` hook.

Fixes https://github.com/lightninglabs/taproot-assets/issues/852


wallet_anchor.go
```diff
@@ -54,8 +54,13 @@ func (l *LndRpcWalletAnchor) FundPsbt(ctx context.Context, packet *psbt.Packet,
 
 	pkt, changeIndex, leasedUtxos, err := l.lnd.WalletKit.FundPsbt(
 		ctx, &walletrpc.FundPsbtRequest{
-			Template: &walletrpc.FundPsbtRequest_Psbt{
-				Psbt: psbtBuf.Bytes(),
+			Template: &walletrpc.FundPsbtRequest_CoinSelect{
+				CoinSelect: &walletrpc.PsbtCoinSelect{
+					Psbt: psbtBuf.Bytes(),
+					ChangeOutput: &walletrpc.PsbtCoinSelect_Add{
```
guggero, line 60: There was some loss of context due to the creation of a new issue. But this isn't enough. What we want to get rid of is this code (or simplify it):

https://github.com/lightninglabs/taproot-assets/blob/52f174888ec4c92e619a3c996772700875afeb56/tapfreighter/wallet.go#L1445-L1486

and

https://github.com/lightninglabs/taproot-assets/blob/3305aef4b2b079685f218b786200f83545e900bc/tapgarden/caretaker.go#L657-L671

See original comment here: https://github.com/lightninglabs/taproot-assets/issues/118#issue-1372070303

In the first case (`wallet.go`), we can add the anchor PSBT inputs before funding and can then skip the `adjustFundedPsbt` step (because the new `FundPsbt` call doesn't do BIP-69 ordering anymore).

In the second case (`caretaker.go`) we are now sure that the change output index is always at index 1, so we don't need that extra logic anymore.

GeorgeTsagk: > In the second case (caretaker.go) we are now sure that the change output index is always at index 1, so we don't need that extra logic anymore.

If I ask `FundPsbt` to always place change on index 1 I'll cause many itests to fail on [`MintAssetsConfirmBatch`](https://github.com/lightninglabs/taproot-assets/blob/main/itest/assets_test.go#L99)

```
2024-03-26 19:37:30.119 [ERR] RPCS: [/mintrpc.Mint/FinalizeBatch]: unable to finalize batch: unable to advance state machine: unable to fund psbt: unable to fund psbt: rpc error: code = Unknown desc = change output index out of range: 1
```

do we want to create the new index, or check if it exists and place there?

guggero: You need to add a change output first (derive address) and add it, so there is something there at index 1.

GeorgeTsagk:
The new `WalletKit.FundPsbt` template option seems to be following a different path for selecting coins to fund the psbt
(as seen [here](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/walletrpc/walletkit_server.go#L1604-L1607))

This new path also introduces a fee sanity check that errors out when the fees are more than 20% of the total amount of the outputs
(see [here](https://github.com/lightningnetwork/lnd/blob/master/lnwallet/chanfunding/coin_select.go#L146-L157)) 

While this fee sanity check makes sense for channel funding cases, where you don't want to overpay, for tapd it may be the case that moving anchors around will burn very high fees compared to the anchors amount.

Not sure how we should proceed, but some options seem to be:
a) Add new exception flag on lnd `WalletKit.FundPsbt` for the fee sanity check to not be performed (feels like bloating the RPC..)
b) Edit the itest cases that trigger this so that bigger coins are involved in the tx and greater amounts are involved, so that we don't run into this case

dstadulis:
> Not sure how we should proceed, but some options seem to be: a) Add new exception flag on lnd `WalletKit.FundPsbt` for the fee sanity check to not be performed (feels like bloating the RPC..) b) Edit the itest cases that trigger this so that bigger coins are involved in the tx and greater amounts are involved, so that we don't run into this case

Additional remediation solution:
Implement https://github.com/lightninglabs/taproot-assets/issues/721 which consequently would increase the denominator of the fee check and causing it to pass.

itest/fee_estimation_test.go
```diff
@@ -17,13 +17,15 @@ import (
 func testFeeEstimation(t *harnessTest) {
 	var (
 		// Make a ladder of UTXO values so use order is deterministic.
-		anchorAmounts = []int64{10000, 9990, 9980, 9970}
+		anchorAmounts = []int64{1_000_000, 999_990, 999_980, 99_970}
```
GeorgeTsagk, line 20: We had to bump the values here in order to not run into the 20% `FundPsbt` fee check anymore. Also the last output was shrinked in order to return the expected errors which are fee calculation related (output must fail for excessive fee)

guggero:
Needs a rebase.

Also, we need to add the code that bumps the minimum required `lnd` version to `v0.17.99` (not `v0.18.0`, otherwise it won't work with current `lnd` `master`).

GeorgeTsagk:
@guggero @jharveyb I added the patch commit for the unit-race. This is now rebased and ready for another round of review! 

ty

tapcfg/config.go
```diff
@@ -180,7 +180,7 @@ var (
 	// required in lnd to run tapd.
 	minimalCompatibleVersion = &verrpc.Version{
 		AppMajor: 0,
-		AppMinor: 15,
+		AppMinor: 17,
```
jharveyb, line 183: Leave a TODO for the bump to 0.18 once tagged?

tapfreighter/wallet.go
```diff
@@ -1387,56 +1350,9 @@ func (f *AssetWallet) FetchInternalKeyLocator(ctx context.Context,
 	return f.cfg.AddrBook.FetchInternalKeyLocator(ctx, rawKey)
 }
```
jharveyb, line 1352: Awesome to have all these checks moved into the LND `FundPsbt` call!

I reviewed that a bit and confirmed that dust checks and removal of dust change are implemented s.t. it would match the behavior we want here.

tapgarden/mock.go
```diff
@@ -147,27 +148,47 @@ func FundGenesisTx(packet *psbt.Packet, feeRate chainfee.SatPerKWeight) {
 	packet.UnsignedTx.TxOut[1].Value -= int64(fee)
 }
 
+// FundPsbt funds a PSBT.
 func (m *MockWalletAnchor) FundPsbt(_ context.Context, packet *psbt.Packet,
-	_ uint32, feeRate chainfee.SatPerKWeight) (*tapsend.FundedPsbt, error) {
+	_ uint32, _ chainfee.SatPerKWeight,
```
jharveyb, line 153: super nit: can split above function to reuse here.

tapgarden/mock.go
```diff
@@ -147,27 +148,47 @@ func FundGenesisTx(packet *psbt.Packet, feeRate chainfee.SatPerKWeight) {
...
+	changeOutput.PkScript[0] = txscript.OP_0
+	packet.UnsignedTx.AddTxOut(&changeOutput)
+	packet.Outputs = append(packet.Outputs, psbt.POutput{})
 
 	// We always have the change output be the second output, so this means
 	// the Taproot Asset commitment will live in the first output.
 	pkt := &tapsend.FundedPsbt{
 		Pkt:               packet,
-		ChangeOutputIndex: 1,
+		ChangeOutputIndex: changeIdx,
```
jharveyb, line 188: Confirmed that this idx can be unchecked by the planter as its checked within `FundPsbt`.

jharveyb (approved):
Great work! 💯 

Not many PRs that are a net reduction in lines. Will be useful for future custom funding work.

tapgarden/planter_test.go
```diff
@@ -735,8 +736,18 @@ func (t *mintingTestHarness) assertBatchGenesisTx(
...
+		psbtBuf  bytes.Buffer
+		psbtCopy *psbt.Packet
+	)
+
+	err := pkt.Pkt.Serialize(&psbtBuf)
+	require.NoError(t, err)
+	psbtCopy, err = psbt.NewFromRawBytes(&psbtBuf, false)
+	require.NoError(t, err)
+
+	for _, txOut := range psbtCopy.UnsignedTx.TxOut {
```
guggero, line 750: Nice catch!

tapcfg/config.go
guggero: Yes. And I think we should either remove the very first commit of the PR (or change the versions to current `lnd master`) _OR_ set this to `0.18.0` directly and then update the `go.mod` once the RC1 has been tagged.
Either way we need to park this PR for a bit until v0.18 is at least in RC.

guggero (commented):
LGTM :tada: 

Not going to approve yet because this depends on `lnd v0.18.0` to be out (at least in RC1).

go.mod
```diff
@@ -1,48 +1,49 @@
 module github.com/lightninglabs/taproot-assets
 
-go 1.21
+go 1.21.4
```
guggero, line 3: I think this should remain `go 1.21`, since a minor version apparently isn't valid in some cases.

tapgarden/mock.go
guggero: There's a lot of code duplication between `FundPsbt` and `FundGenesisTx`, which could be extracted. But non-blocking IMO.

guggero (approved):
SHIP IT :tada:

tapgarden/mock.go
GeorgeTsagk: yeah I took a quick look and it wasn't very obvious to me, gonna skip for now as to not further delay merge

go.mod
GeorgeTsagk: I think lnd master requires `go 1.21.4`, which won't allow us to use `go 1.21` here

PR #861, opened 2024-03-28
Add RFQ buy offer

ffranr:
This PR enhances the RFQ (Request for Quote) system by integrating new capabilities for managing asset buy offers. Key updates include:

- A new `BuyOffer` RFQ offer type for improved screening of sell quotes.
- Adjustments to integration tests to pre-register buy offers, aligning with realistic transaction sequences.
- The negotiator RFQ service now verifies the presence of suitable buy offers when processing sell requests.

These changes mirror the existing `SellOffer` code.


rfq/negotiator.go
```diff
@@ -541,6 +578,119 @@ func (n *Negotiator) HasAssetSellOffer(assetID *asset.ID,
 	return true
 }
 
+// BuyOffer is a struct that represents an asset buy offer. This data structure
+// describes the maximum amount of an asset that this node is willing to
+// purchase.
+//
+// A buy offer is passive (unlike a buy order), meaning that it does not
```
jharveyb, line 585: Comment needs update for `BuyOffer` and not `SellOffer`.

rfq/negotiator.go
```diff
@@ -541,6 +578,119 @@ func (n *Negotiator) HasAssetSellOffer(assetID *asset.ID,
...
+		return fmt.Errorf("invalid asset buy offer: %w", err)
+	}
+
+	// Store the offer in the appropriate map.
+	//
+	// If the asset group key is not nil, then we will use it as the key for
+	// the offer. Otherwise, we will use the asset ID as the key.
+	switch {
+	case offer.AssetGroupKey != nil:
+		n.assetGroupBuyOffers.Store(*offer.AssetGroupKey, offer)
```
jharveyb, line 633: Didn't spot this on the prior PRs, but the group key should be serialized before being used as a map key.

jharveyb (commented):
Looks good! Should there be a case in the itest to test for a missing buy offer when a sell request is received?

rfq/negotiator.go
ffranr: I don't see the mistake here. This bit is correct:
```
it is used by the node to selectively accept or reject incoming asset sell quote requests
```

The counterparty is requesting a quote to sell their asset.

rfq/negotiator.go
ffranr: Could you please help me understand why the group key should be serialized before being used as a map key. The key type in this instance is `btcec.PublicKey`, and not a pointer.

ffranr:
> Should there be a case in the itest to test for a missing buy offer when a sell request is received?

@jharveyb I've been told to not do too much on the RFQ service before we actually start using it. It might be revised considerably. So I think we should skip adding anything further that's non-essential.

rfq/negotiator.go
```diff
@@ -541,6 +541,43 @@ func (n *Negotiator) HasAssetSellOffer(assetID *asset.ID,
...
+	// AssetID represents the identifier of the subject asset.
+	AssetID *asset.ID
+
+	// AssetGroupKey is the public group key of the subject asset.
+	AssetGroupKey *btcec.PublicKey
+
+	// MaxUnits is the maximum amount of the asset which this node is
+	// willing to purchase.
+	MaxUnits uint64
+}
```
GeorgeTsagk, line 599: Doesn't the buy offer include any info for price per unit? Just the unit amount?

rpcserver.go
```diff
@@ -5618,6 +5618,41 @@ func (r *rpcServer) AddAssetSellOffer(_ context.Context,
 	return &rfqrpc.AddAssetSellOfferResponse{}, nil
 }
 
+// AddAssetBuyOffer upserts a new buy offer for the given asset into the RFQ
+// manager. If the offer already exists for the given asset, it will be updated.
+//
+// A buy offer is used by the node to selectively accept or reject incoming
+// asset sell quote requests before price is considered.
```
GeorgeTsagk, line 5625: ok this answers my prior comment

GeorgeTsagk (commented):
solid, few Qs from my end

rfq/negotiator.go
ffranr: Just trying to capture the very simplest case for now. As far as I understand, we don't want to extend the RFQ service too far beyond a MVP right now.

rfq/negotiator.go
jharveyb: So `btcec.PublicKey` is just two `secp.FieldVal` values:

https://github.com/decred/dcrd/blob/08d8572807872f2b9737f8a118b16c320a04b077/dcrec/secp256k1/pubkey.go#L69

But the same public key can be represented by multiple combinations of `secp.FieldVal` values (as those values can be in different formats, non-normalized, etc). For map keys that are structs, they will be compared field-by-field:

https://go.dev/ref/spec#Comparison_operators

So the internal `x` and `y` values will be compared. Using the serialized compressed key avoids this issue (which has appeared before with pubkeys as map keys).

rfq/negotiator.go
jharveyb: Ah, ok - I was comparing this to the godoc for `SellOffer` and wasn't sure if this was correct.

rfq/negotiator.go
ffranr: Ok, I think I understand. Essentially, even though two public keys perform identically, their underlying structures may vary. Consequently, comparing these structures directly, especially the secp.FieldVal fields, might result in an unreliable assessment of equality.

ffranr:
I've added a commit which fixes this problem that @jharveyb identified: https://github.com/lightninglabs/taproot-assets/pull/861#discussion_r1544795064

rfq/negotiator.go
```diff
@@ -13,6 +13,12 @@ import (
 	"github.com/lightningnetwork/lnd/routing/route"
 )
 
+// pubKeyFixedBytes is a fixed size byte array that holds a compressed public
+// key.
+//
+// TODO(ffranr): Export this type from the btcec package.
+type pubKeyFixedBytes = [btcec.PubKeyBytesLenCompressed]byte
```
guggero, line 20: Can't we just use `asset.SerializedKey`?

ffranr: That's what I needed, thanks!

That type should probably be moved into `btcec`. Would be easier to find. I didn't think to look in `asset`.

guggero: Yeah, we have a couple of aliases for the same thing. For example `route.Vertex` in `lnd` and lots of just `[33]byte` types all over. So maybe a type in `btcec` would actually make sense.

rfq/negotiator.go
```diff
@@ -505,7 +555,11 @@ func (n *Negotiator) HasAssetSellOffer(assetID *asset.ID,
 	var sellOffer *SellOffer
 	switch {
 	case assetGroupKey != nil:
-		offer, ok := n.assetGroupSellOffers.Load(*assetGroupKey)
+		groupKeyBytes := assetGroupKey.SerializeCompressed()
```
jharveyb, line 558: Could be `SerializedKey.ToSerialized()`.

jharveyb (approved):
LGTM 💯

itest/rfq_test.go
```diff
@@ -234,8 +234,19 @@ func testRfqAssetSellHtlcIntercept(t *harnessTest) {
 	ctxt, cancel := context.WithTimeout(ctxb, defaultWaitTimeout)
 	defer cancel()
 
-	// TODO(ffranr): Add an asset buy offer to Bob's tapd node. This will
-	//  allow Alice to sell the newly minted asset to Bob.
+	// Upsert an asset buy offer to Bob's tapd node. This will allow Bob to
+	// buy the newly minted asset from Alice.
+	_, err := ts.BobTapd.AddAssetBuyOffer(
```
GeorgeTsagk, line 239: Is this diff here enough to also test the expected behavior of the itest?

Or I assume not all pieces are here yet for the e2e flow?

IIUC here we only test that the RFQ buy offer was placed, but not used in any way

GeorgeTsagk (approved):
Looks good, just a final Q on the e2e itest

itest/rfq_test.go
ffranr: With the updates in this PR, the integration test fails unless we introduce an asset buy offer to Bob's node, as demonstrated here. This is necessary because, in the absence of the buy offer, Alice's subsequent request would be instantly declined by Bob's node, owing to the unavailability of an appropriate buy offer.

The goal of this itest is to test the entire payment flow facilitated by the RFQ service. We are not explicitly testing that the buy offer is used.

PR #862, opened 2024-03-29
tapgarden: check for ErrProofNotFound when checking for local proof file

Roasbeef:
Otherwise, as `lookupProofFilePath` will return an error if the file isn't found (which wraps `proof.ErrProofNotFound`), then the custodian will potentially fail to start up.


GeorgeTsagk (approved):

jharveyb (approved):
Looks good, from a cursory look I don't think we need a similar change elsewhere for `HasProof` at least.

Issue #834, opened 2024-03-14
[feature]: Add type `AssetSpecifier`


ffranr:
We should add and make use of a new type called `AssetSpecifier` which is `fn.Either[asset.ID, btcec.PublicKey]`. With this new type we can specify an asset and ensure that either an asset ID is specified or an asset group key (but not both).

This type should be useful throughout our code base. For example:
https://github.com/lightninglabs/taproot-assets/blob/761ccca181599c40e56206869d57cff32346a983/rpcserver.go#L663-L663

The universe `Identifier` type could make use of this type:
https://github.com/lightninglabs/taproot-assets/blob/761ccca181599c40e56206869d57cff32346a983/universe/interface.go#L45-L52

We should put the type somewhere general where it can be imported and used throughout our codebase.

The RFQ RPC work has already introduced an `AssetSpecifier` RPC type. It should probably be moved somewhere more general. See https://github.com/lightninglabs/taproot-assets/blob/761ccca181599c40e56206869d57cff32346a983/rpcserver.go#L4814-L4814

ffranr:
After experimenting with this modification, I believe using `fn.Either` isn't the best approach. Often, we identify an asset by both its ID and group key, using the group key to assess if the asset can be issued.

From my perspective, using both an asset ID and a group key simply provides a more precise identification of an asset within its group, which is perfectly valid.

Therefore, I suggest that the asset specifier should accommodate either or both the asset ID and asset group key.

ffranr:
https://github.com/lightninglabs/taproot-assets/pull/865 adds the type to our code. It does not include RPC changes. I will create another issue to address RPC modifications.

PR #865, opened 2024-04-02
Add asset specifier

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/834

This PR introduces a new type named `AssetSpecifier` within the `asset` package. The `AssetSpecifier` type is designed to facilitate the conveyance of either an asset ID, an asset group public key, or both, throughout our codebase.

To demonstrate the utility of `AssetSpecifier`, this PR incorporates it in two key areas: coin selection and the derivation of tap commitment keys. While there are numerous potential applications for this new type across our project, this PR deliberately focuses on a select few to maintain brevity. The integration of `AssetSpecifier` sets the stage for future code contributions to leverage this type.

### DIFF
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -247,6 +247,124 @@ func DecodeGenesis(r io.Reader) (Genesis, error) {
 	return gen, err
 }
 
+var (
+	// ErrUnwrapAssetID is an error type which is returned when an asset ID
+	// cannot be unwrapped from a specifier.
+	ErrUnwrapAssetID = errors.New("unable to unwrap asset ID")
+)
+
+// Specifier is a type that can be used to specify an asset by its ID, its asset
+// group public key, or both.
+type Specifier struct {
+	// id is the asset ID.
+	id fn.Option[ID]
+
+	// groupKey is the asset group public key.
+	groupKey fn.Option[btcec.PublicKey]
+}
+
+// NewSpecifierOptionalGroupPubKey creates a new specifier that specifies an
+// asset by its ID and an optional group public key.
+func NewSpecifierOptionalGroupPubKey(id ID,
+	groupPubKey *btcec.PublicKey) Specifier {
+
+	s := Specifier{
+		id: fn.Some(id),
+	}
+
+	if groupPubKey != nil {
+		s.groupKey = fn.Some(*groupPubKey)
+	}
+
+	return s
+}
+
+// NewSpecifierOptionalGroupKey creates a new specifier that specifies an
+// asset by its ID and an optional group key.
+func NewSpecifierOptionalGroupKey(id ID, groupKey *GroupKey) Specifier {
+	s := Specifier{
+		id: fn.Some(id),
+	}
+
+	if groupKey != nil {
+		s.groupKey = fn.Some(groupKey.GroupPubKey)
+	}
+
+	return s
+}
+
+// NewSpecifierFromId creates a new specifier that specifies an asset by its ID.
+func NewSpecifierFromId(id ID) Specifier {
+	return Specifier{
+		id: fn.Some(id),
+	}
+}
+
+// NewSpecifierFromGroupKey creates a new specifier that specifies an asset by
+// its group public key.
+func NewSpecifierFromGroupKey(groupPubKey btcec.PublicKey) Specifier {
+	return Specifier{
+		groupKey: fn.Some(groupPubKey),
+	}
+}
+
+// AsBytes returns the asset ID and group public key as byte slices.
+func (s *Specifier) AsBytes() ([]byte, []byte) {
+	var assetIDBytes, groupKeyBytes []byte
+
+	s.WhenGroupPubKey(func(groupKey btcec.PublicKey) {
+		groupKeyBytes = groupKey.SerializeCompressed()
+	})
+
+	s.WhenId(func(id ID) {
+		assetIDBytes = id[:]
+	})
+
+	return assetIDBytes, groupKeyBytes
+}
+
+// HasId returns true if the asset ID field is specified.
+func (s *Specifier) HasId() bool {
+	return s.id.IsSome()
+}
+
+// HasGroupPubKey returns true if the asset group public key field is specified.
+func (s *Specifier) HasGroupPubKey() bool {
+	return s.groupKey.IsSome()
+}
+
+// WhenId executes the given function if the ID field is specified.
+func (s *Specifier) WhenId(f func(ID)) {
+	s.id.WhenSome(f)
+}
+
+// WhenGroupPubKey executes the given function if asset group public key field
+// is specified.
+func (s *Specifier) WhenGroupPubKey(f func(btcec.PublicKey)) {
+	s.groupKey.WhenSome(f)
+}
+
+// UnwrapIdOrErr unwraps the ID field or returns an error if it is not
+// specified.
+func (s *Specifier) UnwrapIdOrErr() (ID, error) {
+	id := s.id.UnwrapToPtr()
+	if id == nil {
+		return ID{}, ErrUnwrapAssetID
+	}
+
+	return *id, nil
+}
+
+// UnwrapIdToPtr unwraps the ID field to a pointer.
+func (s *Specifier) UnwrapIdToPtr() *ID {
+	return s.id.UnwrapToPtr()
+}
+
+// UnwrapGroupKeyToPtr unwraps the asset group public key field to a pointer.
+func (s *Specifier) UnwrapGroupKeyToPtr() *btcec.PublicKey {
+	return s.groupKey.UnwrapToPtr()
+}
+
 // Type denotes the asset types supported by the Taproot Asset protocol.
 type Type uint8
 
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -2011,6 +2011,12 @@ func (a *Asset) Leaf() (*mssmt.LeafNode, error) {
 	return mssmt.NewLeafNode(buf.Bytes(), a.Amount), nil
 }
 
+// Specifier returns the asset's specifier.
+func (a *Asset) Specifier() Specifier {
+	id := a.Genesis.ID()
+	return NewSpecifierOptionalGroupKey(id, a.GroupKey)
+}
+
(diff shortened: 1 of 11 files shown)


dstadulis:
!lightninglabs-deploy mute 720h00m

GeorgeTsagk:
what's the state / prio of this pr?

dstadulis:
low prio -- it's an enhancement that ffranr created to make asset specification easier, but isn't critical path for proof of concept

asset/asset.go
```diff
@@ -232,6 +232,71 @@ func DecodeGenesis(r io.Reader) (Genesis, error) {
 	return gen, err
 }
 
+// Specifier is a type that can be used to specify an asset by its ID, its asset
+// group public key, or both.
+type Specifier struct {
```
Roasbeef, line 237: Could also do:
```
type Specifier {
      fn.Either[ID, btcec.PublicKey]
}
```

Then we inherit all the methods of the either type.

ffranr: I'm not sure if `fn.Either` will do here, because sometimes the current code expects both fields to be specified.

asset/asset.go
```diff
@@ -247,6 +247,71 @@ func DecodeGenesis(r io.Reader) (Genesis, error) {
...
+	// id is the asset ID.
+	id fn.Option[ID]
+
+	// groupKey is the asset group public key.
+	groupKey fn.Option[btcec.PublicKey]
+}
+
+// NewSpecifier creates a new specifier that specifies an asset by its ID and
+// its group public key.
+func NewSpecifier(id ID, groupPubKey btcec.PublicKey) Specifier {
```
guggero, line 262: To avoid needing to do `if a.groupKey != nil {` in a couple of places, should we have two constructors instead, that take the group key as a pointer each?

What I mean:
```
// NewSpecifier creates a new specifier that specifies an asset by its ID and
// its group public key.
func NewSpecifier(id ID, groupPubKey *btcec.PublicKey) Specifier {
	s := Specifier{
		id: fn.Some(id),
	}

	if groupPubKey != nil {
		s.groupKey = fn.Some(*groupPubKey)
	}

	return s
}

// NewSpecifierWithGroup creates a new specifier that specifies an asset by its
// ID and its group key.
func NewSpecifierWithGroup(id ID, groupKey *GroupKey) Specifier {
	s := Specifier{
		id: fn.Some(id),
	}
	
	if groupKey != nil {
		s.groupKey = fn.Some(groupKey.GroupPubKey)
	}
	
	return s
}
```

Those two should simplify the code in a lot of places.

asset/asset.go
```diff
@@ -1958,6 +1958,18 @@ func (a *Asset) Leaf() (*mssmt.LeafNode, error) {
 	return mssmt.NewLeafNode(buf.Bytes(), a.Amount), nil
 }
 
+// Specifier returns the asset's specifier.
+func (a *Asset) Specifier() Specifier {
+	id := a.Genesis.ID()
+	specifier := NewSpecifierFromId(id)
+
+	if a.GroupKey != nil {
```
guggero, line 1975: See previous comment, could just be `return NewSpecifierWithGroup(a.genesis.ID(), a.GroupKey)`.

asset/asset.go
```diff
@@ -1499,23 +1499,32 @@ func New(genesis Genesis, amount, locktime, relativeLocktime uint64,
...
+	var commitmentKey [32]byte
+
+	switch {
+	case assetSpecifier.HasGroupPubKey():
+		assetSpecifier.WhenGroupPubKey(func(pubKey btcec.PublicKey) {
+			serializedPubKey := schnorr.SerializePubKey(&pubKey)
+			commitmentKey = sha256.Sum256(serializedPubKey)
+		})
+
+	case assetSpecifier.HasId():
```
guggero, line 1515: Since technically there could be an asset specifier that has neither the ID or the group key (of course only due to a mistake by a dev), should we make this a bit more fool proof and move the `TapCommitmentKey` to be a method on the asset specifier? And then panic if neither of the values are set?

tapfreighter/wallet.go
```diff
@@ -554,8 +570,13 @@ func (f *AssetWallet) fundPacketWithInputs(ctx context.Context,
 	fundDesc *tapsend.FundingDescriptor, vPkt *tappsbt.VPacket,
 	selectedCommitments []*AnchoredCommitment) (*FundedVPacket, error) {
 
+	assetId := fundDesc.AssetSpecifier.UnwrapIdToPtr()
```
guggero, line 561: Could also add an `UnwrapIdOrErr()` method, so we can return a named error constant for `unable to unwrap asset ID`.

tapdb/assets_store_test.go
```diff
@@ -659,16 +655,19 @@ func (a *assetGenerator) genAssets(t *testing.T, assetStore *AssetStore,
 	}
 }
 
-func (a *assetGenerator) bindAssetID(i int, op wire.OutPoint) *asset.ID {
+func (a *assetGenerator) assetSpecifierAssetID(
+	i int, op wire.OutPoint) asset.Specifier {
```
guggero, line 659: nit: argument `i` could remain on previous line, same in method below.

tapdb/assets_store_test.go
```diff
@@ -659,16 +655,19 @@ func (a *assetGenerator) genAssets(t *testing.T, assetStore *AssetStore,
...
+
 	gen := a.assetGens[i]
 	gen.FirstPrevOut = op
 
 	id := gen.ID()
-
-	return &id
+	return asset.NewSpecifierFromId(id)
```
guggero, line 665: nit: retain newline before `return`?

tapfreighter/interface.go
```diff
@@ -48,12 +43,15 @@ type CommitmentConstraints struct {
...
-	}
+
+	c.AssetSpecifier.WhenGroupPubKey(func(groupKey btcec.PublicKey) {
+		groupKeyBytes = groupKey.SerializeCompressed()
+	})
+
+	c.AssetSpecifier.WhenId(func(id asset.ID) {
+		assetIDBytes = id[:]
+	})
+
```
guggero, line 54: Feels like all of this could be a helper method as well, since it's being used at least twice in this commit.
Something like `groupKeyBytes, assetIDBytes := specifier.AsBytes()`.

guggero (commented):
Very nice concept, really dig it!

Left a couple of nits and simplification suggestions.

asset/asset.go
ffranr: I'm not a fan of this sort of signature:
```
func NewSpecifier(id ID, groupPubKey *btcec.PublicKey) Specifier
```
It doesn't make it clear that `groupPubKey` is optional. Pointer doesn't always mean optional. So as a developer using `NewSpecifier` I need to read the function body or the doc to figure out if I have to set that arg. Often times the function doc does not help.

That's why I went with:
```
func NewSpecifier(id ID, groupPubKey btcec.PublicKey) Specifier
func NewSpecifierFromId(id ID) Specifier
func NewSpecifierFromGroupKey(groupPubKey btcec.PublicKey) Specifier
```

---

I find `NewSpecifier` and  `NewSpecifierWithGroup` in your examples confusing because the names could be swapped without any effect. `NewSpecifierWithGroup` is just getting to `groupKey.GroupPubKey`, so it seems redundant to me, that field could be passed in at the call site?

---

How do you construct an asset specifier from a group key alone with your suggested `NewSpecifier` and  `NewSpecifierWithGroup`? I think that might be useful when we're doing RFQ for any asset in a group, for example. Or creating a tap address for any asset in a group. So I think the specifier should be able to refer to a group of assets without needing to refer to a single asset.

asset/asset.go
ffranr: I'm not sure. I think we should try to keep `Specifier` and its methods lean. To the same affect, what do you think about adding a default clause to this `switch` which panics?

asset/asset.go
guggero: Okay, I see how the naming might be confusing... But you see my point of trying to avoid all the `if groupKey != nil {` conditions all over the code base?
So maybe we can make it more obvious by putting the "optional" into the name? So `NewSpecifierGroupPubKeyOptional(id ID, groupPubKey *btcec.PublicKey)` and `NewSpecifierGroupKeyOptional(id ID, groupKey *asset.GroupKey)`?

asset/asset.go
guggero: Yeah, sounds good to me :+1:

asset/asset.go
ffranr: I've added those constructors in the latest commits.

tapsend/send.go
```diff
@@ -172,7 +172,12 @@ type FundingDescriptor struct {
 // TapCommitmentKey is the key that maps to the root commitment for the asset
 // group specified by a recipient descriptor.
 func (r *FundingDescriptor) TapCommitmentKey() [32]byte {
-	return asset.TapCommitmentKey(r.ID, r.GroupKey)
+	assetSpecifier := asset.NewIdSpecifier(r.ID)
```
guggero: nit: use the `*Optional` constructor here too?

guggero (approved):
Very nice! LGTM :tada:

tapsend/send.go
ffranr: I think you commented on outdated code and that I've made the right changes in the latest commits.

tapsend/send.go
```diff
@@ -172,7 +172,12 @@ type FundingDescriptor struct {
...
+	if r.GroupKey != nil {
+		assetSpecifier = asset.NewSpecifier(r.ID, *r.GroupKey)
+	}
+
```
Roasbeef: I think this can be simplified to: 
```
assetSpecifier := asset.NewSpecifierOptionalGroupKey(r.ID, r.GroupKey)
```

I think this was mentioned above, but perhaps there was some other resolution somewhere?

Roasbeef (approved):
LGTM ✂️

tapsend/send.go
Roasbeef: Scratch that, I see later on we modified this to use the specifier.

Issue #722, opened 2023-12-05
[feature]: custom Group Key API for minting


jharveyb:
Related to #343 .

We have support for group keys that include tapscript trees, but no way for users to actually provide a tapscript tree root during the minting process. Specifically:

- [x] Providing a specific internal key for a group key for an asset during minting.
- [x] Providing a custom witness for group key membership in a batch when finalizing a batch.
- [x] Storing the tapscript tree for future use (tracked in #343)

For the first point, we'll need to add a new field to the `Seedling` and `MintAssetRequest` structs for custom group keys. We likely also want to mark these seedlings as using a custom group key, so that we require an external witness when finalizing the batch.

For the second point, we'll need to extend the `FinalizeBatch` call to accept a list of witnesses for each seedling. Otherwise we shouldn't have many changes wrt. creating the minting proofs.

The last point is being tracked in #343.

jharveyb:
Depends on #768.

There are a few points in the minting flow where this feature could be exposed:

 - `AddSeedling` call, when providing other details about the asset that will become the group anchor
 - `Finalize` call, when specifying other batch-wide parameters like feerate
 - A new call, `Sign`, to provide a witness for a group key in the batch
 
Separately, there is the option to mandate that a tapscript tree is stored with `tapd` before minting begins (or least, with an RPC unrelated to minting). Then the tree could be referenced by root hash in minting-related calls.

More generally, right now we don't support an externally-managed key as the internal key for an asset group key. However, we also can't create a group key without an anchor asset. So we likely want to expand the `AddSeedling` call to be able to specify a custom internal key.

We'll also need more validation logic around group witnesses, to fail early for incorrect user-provided witnesses, but also to reject batch finalization if `tapd` can't produce a group witness directly (because a group internal key is externally managed).

dstadulis:
During initial implementation, investigation, @jharveyb uncovered, a few areas in which necessary changes will need to be implemented.  An corresponding issue to describe the necessary prerequisites before this issue is started is currently being written.
Roughly the prerequisite are: 
- [ ] Reordering minter logic
- [ ] Exposing group key creation phases

Roasbeef:
> We'll also need more validation logic around group witnesses, to fail early for incorrect user-provided witnesses, but also to reject batch finalization if tapd can't produce a group witness directly (because a group internal key is externally managed).

[...]

PR #866, opened 2024-04-02
tapgarden: implement SealBatch

jharveyb:
Addresses #722. Depends on #827.

Remaining TODOs:

- [ ] Extend `tapdb/asset_minting` tests to cover new query params for script key, group key, etc.
- [ ] Add tests for `tapdb/asset_minting/CommitBatchTx`, `AddSeedlingGroups`, `FetchSeedlingGroups`
- [x] Update SealBatch to reject being called more than once
- [x] Add itest that mints with custom asset group witnesses, custom script keys
- [x] Add itest that spends an asset minted with a custom script key
- [x] Update existing minting itests as needed
- [x] Add itest that spends an anchor amount with a custom script instead of a signature


tapgarden/batch.go
```diff
@@ -76,6 +76,62 @@ type MintingBatch struct {
...
+
+	// BatchKey is the unique identifier for a batch.
+	BatchKey keychain.KeyDescriptor
+
+	// Seedlings is the set of unsealed seedlings for this batch. This maps
+	// an asset's name to the seedling itself.
+	//
+	// NOTE: This field is only set if the state is BatchStateFrozen or
+	// BatchStatePending.
+	Seedlings map[string]*UnsealedSeedling
```
guggero, line 99: Should we perhaps just embed the `MintingBatch` struct and call this `UnsealedSeedlings` to avoid the code duplication? Because every other field is identical to the normal batch.
Or we could even go with `Seedlings fn.Either[map[string]*Seedling, map[string]*UnsealedSeedling]`. Or extend the `Seedling` itself to have a boolean `sealed` and an additional field `*PendingAssetGroup`?

tapgarden/planter_test.go
```diff
@@ -1646,6 +1659,25 @@ func testFundBeforeFinalize(t *mintingTestHarness) {
 	t.assertPendingBatchExists(numSeedlings)
 	t.assertSeedlingsExist(seedlings, nil)
 
+	observedGroupCount := 0
+	fundedSeedlings, err := t.planter.ListBatches(
+		tapgarden.ListBatchesParams{
+			Verbose: true,
+		},
+	)
+	t.Logf("%v", spew.Sdump(fundedSeedlings[0]))
```
guggero: nit: stray debug log. Maybe add some additional context if we want to keep it in for debugging purposes?

tapgarden/planter.go
```diff
@@ -155,15 +155,15 @@ type FundParams struct {
...
+// MemberWitness specifies the group witness for a seedling in a funded batch.
+type MemberWitness struct {
+	Member  asset.ID
+	Witness wire.TxWitness
 }
 
 // SealParams change how asset groups in a minting batch are created.
 type SealParams struct {
-	GroupWitnesses []groupSeal
+	GroupWitnesses []MemberWitness
```
guggero: nit: remove TODO below?

tapgarden/planter.go
```diff
@@ -1018,9 +1018,34 @@ func (c *ChainPlanter) gardener() {
 
 				req.Resolve(c.pendingBatch)
 
-			// TODO(jhb): follow-up PR: Implement SealBatch command
 			case reqTypeSealBatch:
-				req.Error(fmt.Errorf("not yet implemented"))
+				if c.pendingBatch == nil {
+					req.Error(fmt.Errorf(
```
guggero, line 1020: nit: format as:
```
					req.Error(fmt.Errorf("no pending " +
						"batch"))
```
?

taprpc/marshal.go
```diff
@@ -16,6 +21,97 @@ type KeyLookup interface {
...
+			// is derived using BIP-0086.
+			Tweak: rpcKey.TapTweak,
+		}
+	}
+
+	return &scriptKey, nil
+}
+
+// MarshalScriptKey marshals the native script key into the RPC counterpart.
+func MarshalScriptKey(scriptKey asset.ScriptKey) *ScriptKey {
```
guggero, line 100: :+1:

taprpc/mintrpc/mint.proto
```diff
@@ -16,6 +16,20 @@ service Mint {
     */
     rpc MintAsset (MintAssetRequest) returns (MintAssetResponse);
 
+    /* tapcli `assets mint fund`
+    FundBatch will attempt to fund the current pending batch with a genesis
+    input, or create a new funded batch if no batch exists yet.
+    */
+    rpc FundBatch (FundBatchRequest) returns (FundBatchResponse);
+
+    // sealbatch
```
guggero, line 25: nit: stray comment.

taprpc/mintrpc/mint.proto
```diff
@@ -16,6 +16,20 @@ service Mint {
     */
     rpc MintAsset (MintAssetRequest) returns (MintAssetResponse);
 
+    /* tapcli `assets mint fund`
```
guggero, line 19: Reminder to implement those two new commands in the CLI as well in this PR, otherwise this breaks the API docs generator.

taprpc/mintrpc/mint.proto
```diff
@@ -172,6 +244,27 @@ message MintingBatch {
...
+    The transaction ID of the batch. Only populated if the batch has been
+    committed.
+    */
+    string batch_txid = 2;
+
+    // The state of the batch.
+    BatchState state = 3;
+
+    // The assets that are part of the batch.
+    repeated UnsealedAsset assets = 4;
```
guggero, line 265: Same here re code duplication. Maybe we can also re-use the `MintingBatch` and just have two types of assets (`repeated PendingAsset pending_assets` and `repeated UnsealedAssets unsealed_assets`) of which only ever one is set?

tapgarden/planter.go
```diff
@@ -1277,6 +1274,10 @@ func (c *ChainPlanter) sealBatch(ctx context.Context,
 		return nil, fmt.Errorf("unable to build group requests: "+
 			"%w", err)
 	}
+	if len(groupReqs) != len(genTXs) {
+		return nil, fmt.Errorf("mismatched number of group requests" +
```
guggero, line 1278: nit: add space after word "requests".

guggero (commented):
Looks pretty good to me! Looking forward to seeing it in action with an itest :tada:
Main request is to potentially remove some code duplication by moving some of the complexity of the seedlings further down, instead of having two almost identical batch structs.

tapgarden/planter.go
jharveyb: Was related to #721 but I think that won't be a fast follow 👍🏽

jharveyb:
Looks like a possible new flake in custodian event handling?

```bash
--- FAIL: TestTransactionConfirmedOnly (123.22s)
    test_postgres.go:11: Creating new Postgres DB for testing
    custodian_test.go:220: 
        	Error Trace:	/home/runner/work/taproot-assets/taproot-assets/tapgarden/custodian_test.go:220
        	            				/home/runner/work/taproot-assets/taproot-assets/tapgarden/custodian_test.go:740
        	            				/home/runner/work/taproot-assets/taproot-assets/tapgarden/custodian_test.go:669
        	            	method did not return within the timeout
        	Test:       	TestTransactionConfirmedOnly
FAIL
FAIL	github.com/lightninglabs/taproot-assets/tapgarden	156.362s
FAIL
```

jharveyb:
Posted some updates, mostly fixes where the group tapscript root wasn't being propogated correctly.

Also some changes in the `MintAsset` RPC to accomodate all the new fields.

Still working out some marshalling issues when generating group witnesses.

jharveyb:
Posted my latest state. The issue with the version of the test I was working on before was internal keys not derived from the  backing tapd node.

The posted version generates a sig for a siglock external to the planter, but I could not get a similar witness working if that key was derived from the backing lnd node, for both the Taproot Assets key family, or a random key family.

I'm at the last steps of the PSBT flow, which is very similar to the multisig test.

jharveyb:
Remaining tasks:

- [ ] freeze batch at the end of SealBatch to prevent further calls to add seedlings, fund, or freeze
- [ ] Add CLI support

jharveyb:
I have a branch that resolves #883 , but the better fix for that would be switching to `SignPsbt` for the `VirtualTxSigner`.

From offline discussion, I think for this PR we'll continue to accept a non-local group internal key without any constraints on the `keyDescriptor` / won't require mutual exclusion of the pubkey and `keyLocator`. I think having the family and index in the DB for a non-local internal key is useful in some cases, and *shouldn't* cause issues if it's for the group internal key.

jharveyb:
CLI commands are added, but they omit the fields with non-trivial formatting like `KeyDescriptor` or `asset.ScriptKey`. Similar to the commands we have for generating addresses, some functionality is only available via the RPCs directly.

tapgarden/planter.go
```diff
@@ -1273,47 +1284,116 @@ func (c *ChainPlanter) sealBatch(ctx context.Context, _ SealParams) error {
...
+		ctx, genesisPoint, anchorOutputIndex, singleSeedling,
+	)
+
+	switch {
+	case len(existingGroups) != 0:
+		return nil, fmt.Errorf("batch is already sealed")
+	case err != nil:
+		// The only expected error is for a missing asset genesis.
+		errMsg := err.Error()
+		if !strings.Contains(errMsg, "unable to fetch genesis asset") {
```
guggero, line 1307: Feels like we should return a typed error here and use `errors.Is()` instead of string matching...

taprpc/marshal.go
```diff
@@ -50,12 +153,203 @@ func MarshalAssetVersion(version asset.Version) (AssetVersion, error) {
...
+}
+
+// UnmarshalGroupWitness parses an asset group witness from the RPC variant.
+func UnmarshalGroupWitness(wit *GroupWitness) (*asset.PendingGroupWitness,
+	error) {
+
+	// Assert that a given witness stack does not exceed the limit used by
+	// the VM.
+	checkWitnessSize := func(wit [][]byte) error {
+		witSize := fn.Reduce(wit,
```
guggero, line 386: nit: formatting.

taprpc/mintrpc/mint.proto
```diff
@@ -16,6 +16,19 @@ service Mint {
     */
     rpc MintAsset (MintAssetRequest) returns (MintAssetResponse);
 
+    /* tapcli `assets mint fund`
+    FundBatch will attempt to fund the current pending batch with a genesis
+    input, or create a new funded batch if no batch exists yet.
```
guggero, line 21: Feels like we should add a bit more information about when or if this needs to be called? I'd suggest something like this: "This RPC method only needs to be called if the funding transaction (e.g. fee rate) should be customized. Otherwise SealBatch or FinalizeBatch can be called directly."

taprpc/mintrpc/mint.proto
```diff
@@ -16,6 +16,19 @@ service Mint {
...
+    */
+    rpc FundBatch (FundBatchRequest) returns (FundBatchResponse);
+
+    /* tapcli `assets mint seal`
+    SealBatch will attempt to seal the current pending batch by creating and
+    validating asset group witness for all assets in the batch. If a witness
+    is not provided, a signature will be derived to serve as the witness.
```
guggero, line 28: Additional comment here could say "Calling this RPC method is only required if any of the assets in the batch have a custom group key that needs to be signed with externally. Otherwise FinalizeBatch can be called directly."

rpcserver.go
```diff
@@ -580,42 +650,97 @@ func checkFeeRateSanity(rpcFeeRate uint32) (*chainfee.SatPerKWeight, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &mintrpc.FundBatchResponse{
+		Batch: rpcBatch,
+	}, nil
+}
+
+func (r *rpcServer) SealBatch(ctx context.Context,
```
guggero, line 695: nit: missing Godoc.

itest/assertions.go
```diff
@@ -142,6 +142,72 @@ func AssetScriptKeyIsBurnCheck(isBurn bool) AssetCheck {
 	}
 }
 
+func AssetScriptKeyCheck(scriptKey *taprpc.ScriptKey) AssetCheck {
+	return func(a *taprpc.Asset) error {
+		if scriptKey != nil {
```
guggero, line 147: nit: invert condition and return early to continue un-indented below? Same for the other new checks below.

itest/mint_fund_seal_test.go
```diff
@@ -0,0 +1,706 @@
...
+	)
+
+	aliceLndClient, err := t.newLndClient(aliceLnd)
+	require.NoError(t.t, err)
+
+	defer func() {
+		require.NoError(t.t, bobTapd.stop(!*noDelete))
+	}()
+
+	displayGroupWitness := func(wit *taprpc.GroupWitness) string {
```
guggero, line 55: nit: since this is a gRPC message could instead use `toJSON()`.

itest/mint_fund_seal_test.go
```diff
@@ -0,0 +1,706 @@
...
+		)
+
+		// Ensure that we use a different key family from tapd.
+		for randFam == asset.TaprootAssetsKeyFamily {
+			randFam = test.RandInt31n(math.MaxInt32)
+		}
+
+		// Use DeriveNextKey to reach our desired key index.
+		for i := int32(0); i != randInd; i++ {
+			desc, err = keyRing.DeriveNextKey(
```
guggero, line 89: micro-nit: in case we already have a non-zero amount of keys for this family, we could set `i = desc.Index` to short cut and don't derive too many keys.

guggero: Would require a change to the condition though, so would need to be `i <= randInt`, otherwise it might turn into an endless loop.

itest/mint_fund_seal_test.go
```diff
@@ -0,0 +1,706 @@
...
+	tweakedScriptHashLock := test.ScriptHashLock(
+		t.t, bytes.Clone(test.DefaultHashLockWitness),
+	)
+	tweakedScriptSigLock := test.ScriptSchnorrSig(
+		t.t, tweakedScriptDesc.PubKey,
+	)
+	tweakedScriptTapTree := txscript.AssembleTaprootScriptTree(
+		tweakedScriptHashLock, tweakedScriptSigLock,
+	)
+	tweakedScriptTweak := txscript.NewTapBranch(
```
guggero, line 112: nit: this could just be `tweakedScriptTapTree.Root.TapHash()`, no need to create a branch. Same further below.

itest/mint_fund_seal_test.go
```diff
@@ -0,0 +1,706 @@
...
+		},
+	)
+	require.Equal(t.t, 3, unsealedAssetCount)
+
+	// Fetch the responses that match the three grouped asset requests.
+	assetGroupedInternalTweaked, err := fn.First(
+		unsealedAssets, func(a *mintrpc.UnsealedAsset) bool {
+			return a.Asset.Name ==
+				assetReqGroupedInternalTweaked.Asset.Name
+		})
```
guggero, line 267: nit: closing parenthesis on next line (here and a couple of instances further below).

itest/mint_fund_seal_test.go
```diff
@@ -0,0 +1,706 @@
...
+	)
+	assetTweakedScriptKey, err := fn.First(
+		batchAssets, func(a *taprpc.Asset) bool {
+			return assetReqWithScriptKey.Asset.Name ==
+				a.AssetGenesis.Name
+		})
+	require.NoError(t.t, err)
+
+	// We should have one group with two assets and a balance of 2. The
+	// The other should have one asset with a balance of 5000.
```
guggero, line 429: nit: duplicate word in second sentence.

itest/mint_fund_seal_test.go
```diff
@@ -0,0 +1,706 @@
...
+		OutputKeyYIsOdd: anchorKeyYIsOdd,
+		LeafVersion:     txscript.BaseLeafVersion,
+		InclusionProof:  anchorUTXO.TaprootAssetRoot,
+	}
+	anchorInputControlBlockBytes, err := anchorInputControlBlock.ToBytes()
+	require.NoError(t.t, err)
+
+	// With our control block, we can generate the two needed signatures
+	// and construct the anchor witness.
+	firstKeyPartialSig := partialSignWithKey(
```
guggero, line 621: Nice to have this flow tested as well :100:

itest/mint_fund_seal_test.go
```diff
@@ -0,0 +1,706 @@
...
+	// Bob should have detected the transfer, and now own half of the total
+	// asset supply.
+	AssertAddrEvent(t.t, bobTapd, bobAddr, 1, statusCompleted)
+	AssertNonInteractiveRecvComplete(t.t, bobTapd, 1)
+	AssertBalanceByID(
+		t.t, bobTapd, assetTweakedScriptKey.AssetGenesis.AssetId,
+		assetTweakedScriptKey.Amount/2,
+	)
+
+	time.Sleep(time.Second / 4)
```
guggero, line 664: Can we use `ConfirmAndAssertOutboundTransferWithOutputs` instead of the sleep to make sure Alice sees the outbound transfer as complete as well?

cmd/tapcli/assets.go
```diff
@@ -291,11 +315,10 @@ func mintAsset(ctx *cli.Context) error {
...
 
-var finalizeBatchCommand = cli.Command{
-	Name:        "finalize",
-	ShortName:   "f",
-	Usage:       "finalize a batch",
-	Description: "Attempt to finalize a pending batch.",
+var fundBatchCommand = cli.Command{
+	Name:        "fund",
+	Usage:       "fund a batch",
+	Description: "Attempt to fund a pending batch.",
```
guggero, line 321: Add some additional context here similar to my previous comment about _when_ this might be needed? Otherwise the additional commands might confuse the users.

cmd/tapcli/assets.go
```diff
@@ -310,29 +333,82 @@ var finalizeBatchCommand = cli.Command{
...
 
-		return uint32(feeRate), nil
+	printRespJSON(resp)
+	return nil
+}
+
+var sealBatchCommand = cli.Command{
+	Name:        "seal",
+	Usage:       "seal a batch",
+	Description: "Attempt to seal the pending batch.",
```
guggero, line 364: Does this command really do anything on the command line other than sanity check that we can generate witnesses for all group keys in the batch? Not sure if we should hide (`Hidden: true`) the command and only keep it for debugging? Since it probably doesn't make sense to support submitting the actual witnesses through the CLI (only makes sense through RPC).

rpcserver.go
```diff
@@ -580,42 +650,97 @@ func checkFeeRateSanity(rpcFeeRate uint32) (*chainfee.SatPerKWeight, error) {
...
+	req *mintrpc.SealBatchRequest) (*mintrpc.SealBatchResponse, error) {
+
```
Roasbeef, line 697: Shouldn't this require that group witnesses re provided? Otherwise, how can you seal a batch w/o providing the witness?

jharveyb: Seal with no arguments will try to produce the 'default' witness of a signature with a tapd-owned key. That would fail if tapd can't sign for all asset group keys. At that point the user could try `SealBatch` again with the required witnesses.

itest/mint_fund_seal_test.go
jharveyb: Good catch! I think using the random int as an offset from the current index should work.

tapgarden/planter.go
jharveyb: I think I tried that before, but pulling in the concrete error from `tapdb` causes an import cycle.

taprpc/mintrpc/mint.proto
jharveyb: Updated, it's narrower than that since you can provide a tapscript sibling and feerate with both `Fund` and `Finalize`. So it's really only if you plan to use `Seal` later.

cmd/tapcli/assets.go
jharveyb: Yeah I agree it doesn't have much use...

I suppose we want to still expose `fund` as we'll add other features to that later.

jharveyb: Update: added more to the descriptions and hid this cmd. In practice you could do `fund` + `finalize` from the CLI if we don't accept witness in `seal`.

jharveyb:
Updated to address all comments, itest should be more readable now + other small fixes.

Roasbeef (approved):
LGTM 🍵

PR #870, opened 2024-04-03
tools: update linter to latest, update linter config

jharveyb:
Bumps golangci-lint to 1.57.1 and updates our linter config to match changed upstream defaults.


gijswijs (commented):
Does it make sense to add `golangci-lint config verify` to the Makefile somewhere?

We have the option to check it, so it seems prudent to do so.

E.g. if I add a bs option to the config file and run the verifier I get

```
golangci-lint config verify
jsonschema: "linters-settings.gofmt" does not validate with "/properties/linters-settings/properties/gofmt/additionalProperties": additionalProperties 'tralalal' not allowed
Error: the configuration contains invalid elements
```

But if I then run `golangci-lint run` it just runs the linter without complaining. There is value in having a clean and validated `golangci.yml`, so I prefer adding it to the `lint` target in the Makefile.

ffranr (approved):
This change looks good as is IMO. I don't have a strong opinion on enable/disable linters by default.

I like the suggestion of `golangci-lint config verify` somewhere in the CI. Possibly as a first step in the `make lint` command? But I also think that that change can be a separate small PR.

gijswijs (commented):
Agreed on the `golangci-lint config verify` in a separate PR.

w.r.t. enabling/disabling linters by default. It ensures a more reproducible CI, so I'm in favor of it.

tapdb/migrations.go
```diff
@@ -173,7 +173,7 @@ func newReplacerFile(parent fs.File, replaces map[string]string) (*replacerFile,
 
 	contentStr := string(content)
 	for from, to := range replaces {
-		contentStr = strings.Replace(contentStr, from, to, -1)
```
gijswijs, line 176: nit: seems unrelated to the PR at hand?

.golangci.yml
```diff
@@ -26,18 +23,21 @@ linters-settings:
...
+        - unslice
+        - captLocal
+        - underef
+        - appendAssign
+        - exitAfterDefer
   staticcheck:
     checks: ["-SA1019"]
 
 linters:
   enable:
```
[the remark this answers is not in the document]
ffranr: An argument for enabling all linters by default in `golangci-lint` is to maintain code quality in alignment with evolving standards. As `golangci-lint` introduces new linters in future releases, enabling them by default ensures our codebase automatically benefits from the latest practices and checks without requiring manual intervention to discover and enable new linters with each version update. This approach proactively keeps our code compliant with the most current standards endorsed by `golangci-lint`. We will have an opportunity to encounter deviations from new linters when upgrading `golangci-lint`.

I think the version of `golangci-lint` we use in our project should be fixed. So default changes to `golangci-lint` will be dealt with when upgrading that package. That shouldn't cause instability on `main` or in the CI.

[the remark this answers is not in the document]
gijswijs: @ffranr Fair point.  your approach is also in line with golang-ci integration guidelines (https://golangci-lint.run/welcome/install/#github-actions)

.golangci.yml
```diff
@@ -26,18 +23,21 @@ linters-settings:
   exhaustive:
     default-signifies-exhaustive: true
   gocritic:
-      enabled-checks:
-        - singleCaseSwitch
-        - ifElseChain
-        - assignOp
-        - unlambda
+      disabled-checks:
```
guggero, line 26: I think this is also indented by a level too much?

guggero (approved):
Thanks, LGTM :tada:

tapdb/migrations.go
jharveyb: Yes, though it also causes a postgres test failure, which is surprising.

.golangci.yml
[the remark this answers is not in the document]
jharveyb: Worth noting that new linters that become defaults in future versions won't be enabled unless we bump the `golangci-lint` version in a future PR similar to this one.

tapdb/migrations.go
dstadulis: Help me understand why the test failure, post function change, is surprising

jharveyb: Should be no behavior change moving from `Replace` with a 3rd arg of `-1` to `ReplaceAll`, according to the `strings` package documentation. But one of the unit tests running under postgres failed on an SQL issue. Which I don't think was a known flake.

Will flakehunt locally.

Issue #864, opened 2024-04-02
channels: implement the initial version of the `AuxLeafStore`


Roasbeef:
This is related to: https://github.com/lightninglabs/taproot-assets/issues/863. 

Given the ability to encode/decode our book keeping data in TLV blobs, we now need a way to map that blob, and some additional channel information into the set of tapscript leaves that'll commit to the asset information in all the commitment outputs.

### Primary Abstraction Definition 

As is, the primary interface resembles something like: 
```go
// CommitAuxLeaves stores two potential auxiliary leaves for the remote and
// local output that made be used to argument the final tapscript trees of the
// commitment transaction.
type CommitAuxLeaves struct {
	// LocalAuxLeaf is the local party's auxiliary leaf.
	LocalAuxLeaf input.AuxTapLeaf

	// RemoteAuxLeaf is the remote party's auxiliary leaf.
	RemoteAuxLeaf input.AuxTapLeaf

	// OutgoingHTLCLeaves is the set of aux leaves for the outgoing HTLCs
	// on this commitment transaction.
	OutgoingHtlcLeaves input.AuxTapLeaves

	// IncomingHTLCLeaves is the set of aux leaves for the incoming HTLCs
	// on this commitment transaction.
	IncomingHtlcLeaves input.AuxTapLeaves
}

// ForRemoteCommit returns the local+remote aux leaves from the PoV of the
// remote party's commitment.
func (c *CommitAuxLeaves) ForRemoteCommit() CommitAuxLeaves {
	// TODO(roasbeef): remove?
	return CommitAuxLeaves{
		LocalAuxLeaf:       c.RemoteAuxLeaf,
		RemoteAuxLeaf:      c.LocalAuxLeaf,
		OutgoingHtlcLeaves: c.IncomingHtlcLeaves,
		IncomingHtlcLeaves: c.OutgoingHtlcLeaves,
	}
}

// AuxLeafStore is used to optionally fetch auxiliary tapscript leaves for the
// commitment transaction given an opaque blob. This is also used to implement
// a state transition function for the blobs to allow them to be refreshed with
// each state.
type AuxLeafStore interface {
	// FetchLeavesFromView attempts to fetch the auxiliary leaves that
	// correspond to the passed aux blob, and pending fully evaluated HTLC
	// view.
	FetchLeavesFromView(prevBlob tlv.Blob,
		view *HtlcView,
		keyRing CommitmentKeyRing) fn.Option[CommitAuxLeaves]

[...]

Issue #863, opened 2024-04-02
channels: create parsing routines for `tlv.Blob` auxiliary blobs 


Roasbeef:
For the integration in `lnd`, we'll need to store some meta data long side the channel state to be able to reconstruct all the tap leaves we need for signing and such. 

In the initial PR series, we've started to store `tlv.Blob` in the areas:
  * `OpenChannel`: This blob will store the _capacity_ information related to a channel. This entails all the `(asset_id, amt)` tuples, and potentially other information that we may need to be able to sign the TAP portion of the commitment transaction. 
  * `ChannelCommitment`: This blob stores the balance information for a given commitment/state. This includes the local+remote balance, and the asset mappings within them. We can also optionally store information related to any/all HTLCs in this blob space.
       * If HTLC information isn't stored here, then w/ the way the interfaces are set up rn, we can just obtain the asset information from the `ExtraData` TLV that'll store the ID+amt information. See the bLIP for more information. 
   * `RevocationLog`: Similar to the above, but used to store information related to a revoked state. 

## Steps to Completion 
- [ ] Define a new struct that'll store the attributes to be TLV encoded in each of the areas above. 
- [ ] Write encode/decode routines for this struct to map from the raw bytes back to the asset balance meta data.

PR #878, opened 2024-04-11
[custom channels 1/3]: create aux leaf store and tlv.Blob decoders

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/864.
Fixes https://github.com/lightninglabs/taproot-assets/issues/863.

A first implementation of the aux leaf store that decodes the custom `tlv.Blob`s from commitments and HTLCs and creates auxiliary leafs from them.

Link to all PRs in the saga:


tappsbt/create.go
```diff
@@ -144,3 +147,63 @@ func OwnershipProofPacket(ownedAsset *asset.Asset,
...
+// the packet.
+func FromProofs(proofs []*proof.Proof,
+	params *address.ChainParams) (*VPacket, error) {
+
+	pkt := &VPacket{
+		ChainParams: params,
+		Version:     0,
+	}
+
+	for idx := range proofs {
```
ffranr, line 161: We could spin the body of this loop into a `NewVInputFromProof` function. `FromProofs` would then be simpler and we might be able to reuse `NewVInputFromProof`.

asset/asset.go
```diff
@@ -193,6 +193,18 @@ func (i ID) String() string {
 	return hex.EncodeToString(i[:])
 }
 
+// Record returns a TLV record that can be used to encode/decode an ID to/from a
+// TLV stream.
+//
+// NOTE: This is part of the tlv.RecordProducer interface.
```
ffranr, line 199: Perhaps better than just a comment, we could ensure using:
```
// Ensure ID implements the tlv.RecordProducer interface.
var _ tlv.RecordProducer = (*ID)(nil)
```

proof/encoding.go
```diff
@@ -14,6 +14,33 @@ import (
 	"github.com/lightningnetwork/lnd/tlv"
 )
 
+func Encoder(w io.Writer, val any, buf *[8]byte) error {
```
ffranr, line 17: lint: missing empty lines and function doc comments.

proof/proof.go
```diff
@@ -430,6 +430,25 @@ func (p *Proof) Decode(r io.Reader) error {
 	return stream.Decode(r)
 }
 
+// Record returns a TLV record that can be used to encode/decode a Proof to/from
+// a TLV stream.
+//
+// NOTE: This is part of the tlv.RecordProducer interface.
```
ffranr, line 436: `var _ tlv.RecordProducer = (*ID)(nil)` instead of comment here also?

tapchannel/allocation.go
```diff
@@ -0,0 +1,485 @@
...
+	// CommitAllocationHtlcIncoming is an allocation type that is used for
+	// allocating assets to an incoming HTLC output.
+	CommitAllocationHtlcIncoming AllocationType = 3
+
+	// CommitAllocationHtlcOutgoing is an allocation type that is used for
+	// allocating assets to an outgoing HTLC output.
+	CommitAllocationHtlcOutgoing AllocationType = 4
+
+	// SecondLevelHtlcAllocation is an allocation type that is used for
+	// allocating assets to a second level HTLC output.
```
ffranr, line 73: What does "second level" mean here? Maybe we should add more to this comment to explain.

ffranr (commented):
I still need to review `tapchannel: add allocation struct` forward.

tappsbt/create.go
guggero: I see you've already included this commit in your PR (#896). I'd say let's do it when we actually need it (IMO because of the call to `pkt.SetInputAsset()` creating a `VInput` outside of the context of a `VPacket` would leave us with an incomplete `VInput`).

tapchannel/allocation.go
guggero: Extended the comment with the wording from the BOLTS: https://github.com/lightning/bolts/blob/master/03-transactions.md#commitment-transaction-outputs

tapchannel/allocation.go
```diff
@@ -0,0 +1,485 @@
...
+	HtlcIndex input.HtlcIndex
+
+	// OutputCommitment is the taproot output commitment that is set after
+	// fully distributing the coins and creating the asset and TAP trees.
+	OutputCommitment *commitment.TapCommitment
+}
+
+// TapscriptSibling returns the tapscript sibling preimage from the non-asset
+// leaves of the allocation. If there are no non-asset leaves, nil is returned.
+func (a *Allocation) TapscriptSibling() (*commitment.TapscriptPreimage, error) {
```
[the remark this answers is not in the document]
guggero: I changed it to an option in a local version and it made the two call sites of this method way less readable and also quite a bit longer (and in both cases we would transform the option into a nil or non-nil value again in the next step anyway). I think options are nice for library APIs where the caller might not know what to expect. So I instead decided to not export this method but keep the pointer.

tapchannel/allocation.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+
+	// We group the assets by asset ID, since we'll want to create a single
+	// virtual packet per asset ID (with each virtual packet potentially
+	// having multiple inputs and outputs).
+	assetIDs := fn.Map(inputs, func(input *proof.Proof) asset.ID {
+		return input.Asset.ID()
+	})
+	uniqueAssetIDs := fn.NewSet(assetIDs...).ToSlice()
+	pieces := make([]*piece, len(uniqueAssetIDs))
```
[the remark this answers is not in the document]
guggero: Correct. Added a comment above. I know "piece" is a terrible name. If you have a better suggestion, I'm all ears!

tapchannel/allocation.go
```diff
@@ -0,0 +1,485 @@
...
+	CommitAllocationHtlcOutgoing AllocationType = 4
+
+	// SecondLevelHtlcAllocation is an allocation type that is used for
+	// allocating assets to a second level HTLC output.
+	SecondLevelHtlcAllocation AllocationType = 5
+)
+
+// Allocation is a struct that tracks how many units of assets should be
+// allocated to a specific output of an on-chain transaction.
+type Allocation struct {
```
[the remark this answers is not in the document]
guggero: I would look at it the other way around: The allocation is the recipe/instruction, not the "result" of the distribution. It basically means: Assign me any available asset pieces until this amount is reached. And the whole reason we have to do this whole algorithms with allocations and coin distribution in the first place is **because** we can have UTXOs from different asset IDs. So by definition all of the code in this file is for handling multiple assets.

I've updated some comments on the struct, but would prefer to keep the shorter name here.

guggero:
Thanks for the review! I addressed all comments and also pulled in all changes from the PoC branch. Should be ready for secondary review.

ffranr (approved):
a lot of good work :+1:

rpcserver.go
```diff
@@ -2257,13 +2257,6 @@ func (r *rpcServer) CommitVirtualPsbts(ctx context.Context,
 		return nil, fmt.Errorf("error funding packet: %w", err)
 	}
 
-	// Validate the actual fee rate of the transaction.
-	txFees, err := fundedPacket.GetTxFee()
-	if err != nil {
-		return nil, fmt.Errorf("error calculating transaction fees: %w",
-			err)
-	}
-
```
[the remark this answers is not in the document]
guggero: Yes this was intentional, the `txFees` was only used to set a field in the `tapsend.AnchorTransaction` we populated below. But since we're only going to use the `fundedPacket.UnsignedTx` directly, everything else is not needed anymore.

proof/taproot.go
```diff
@@ -539,3 +539,113 @@ func AddExclusionProofs(baseProof *BaseProofParams, finalTx *wire.MsgTx,
...
+		// TapscriptProof. So we'll swap things below.
+		tree := txscript.AssembleTaprootScriptTree(leaves...)
+		topLeft := tree.RootNode.Left()
+		topRight := tree.RootNode.Right()
+
+		topLeftBranch, ok := topLeft.(txscript.TapBranch)
+		if !ok {
+			return nil, fmt.Errorf("expected left node to be a "+
+				"branch, got %T", topLeft)
+		}
```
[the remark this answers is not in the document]
guggero: Yes, the comment is correct in its meaning. Just "in the right" branch is perhaps a bit misleading, as on the right of the root there is no branch, just a single leaf. Updated the comment a bit.

tapchannelmsg/records.go
```diff
@@ -0,0 +1,1568 @@
...
+
+type (
+	// HtlcAmountRecordType is a type alias for the TLV type that is used to
+	// encode an asset ID and amount list within the custom records of an
+	// HTLC record on the wire.
+	HtlcAmountRecordType = tlv.TlvType65536
+
+	// HtlcSigsRecordType is a type alias for the TLV type that is used to
+	// encode the signatures of an HTLC record on the wire.
+	HtlcSigsRecordType = tlv.TlvType65537
```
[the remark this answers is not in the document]
guggero: It will be used in a later commit. And I think the order of the TLV numbers makes sense in terms of how they are introduced in the bLIP.

tapchannelmsg/records.go
```diff
@@ -0,0 +1,1568 @@
...
+	return tlvStream.Decode(r)
+}
+
+// eHtlcAuxLeafMapRecord is an encoder for HtlcAuxLeafMapRecord.
+func eHtlcAuxLeafMapRecord(w io.Writer, val interface{}, buf *[8]byte) error {
+	if v, ok := val.(*map[input.HtlcIndex]HtlcAuxLeaf); ok {
+		numHtlcs := uint64(len(*v))
+		if err := tlv.WriteVarInt(w, numHtlcs, buf); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
guggero: The max check when decoding is to avoid a memory allocation DoS, since we decode stuff sent to us over the wire. I don't think we need to defend against us creating too many HTLCs, at least not at this (wire) level.

tapchannel/commitment.go
```diff
@@ -0,0 +1,1129 @@
...
+	// PaymentDescriptor is the original payment descriptor.
+	*lnwallet.PaymentDescriptor
+
+	// AssetBalances is the decoded asset balances of the HTLC.
+	AssetBalances []*cmsg.AssetBalance
+}
+
+// DecodedView is a copy of the original HTLC view, but with the asset balances
+// of the HTLCs decoded.
+type DecodedView struct {
```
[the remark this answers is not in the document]
guggero: The goal here is to have a different name than the `lnwallet.HtlcView` and `lnwallet.PaymentDescriptor` to indicate that we have the same data, just in an already readily available manner. Using the same name but re-declared in this package would just lead to confusion and decrease readability. Would you like "Parsed..." instead?

tapchannel/commitment.go
```diff
@@ -0,0 +1,1129 @@
...
+				AssetBalances:     assetHtlc.Balances(),
+			}
+			local, remote = processAddEntry(
+				decodedEntry, local, remote, isOurCommit, false,
+				nextHeight,
+			)
+
+			newView.OurUpdates = append(
+				newView.OurUpdates, decodedEntry,
+			)
```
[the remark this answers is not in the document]
guggero: I tried this with the `processRemoveEntry` further up and it would look something like this:

```go
			var assetHtlc lfn.Option[rfqmsg.Htlc]
			err := lfn.MapOptionZ(
				parentEntry.CustomRecords,
				func(blob tlv.Blob) error {
					htlc, err := rfqmsg.DecodeHtlc(blob)
					if err != nil {
						return fmt.Errorf("unable to "+
							"decode asset htlc: %w",
							err)
					}

					assetHtlc = lfn.Some(*htlc)

					return nil
				},
			)
			if err != nil {
				return 0, 0, nil, fmt.Errorf("error "+
					"processing our remove entry: %w", err)
			}

			assetHtlc.WhenSome(func(a rfqmsg.Htlc) {
				decodedEntry := &DecodedDescriptor{
					PaymentDescriptor: entry,
					AssetBalances:     a.Balances(),
				}

				local, remote = processRemoveEntry(
					decodedEntry, local, remote,
					isOurCommit, true, nextHeight,
				)
			})
```

And in my opinion it requires more mental work to parse what's happening compared to having everything inside a single `WhenSome`.

tapchannel/commitment.go
```diff
@@ -0,0 +1,1129 @@
...
+	if isIncoming {
+		// If this is a new incoming (un-committed) HTLC, then we need
+		// to update their balance accordingly by subtracting the
+		// amount of the HTLC that are funds pending.
+		theirBalance -= amount
+	} else {
+		// Similarly, we need to debit our balance if this is an
+		// outgoing HTLC to reflect the pending balance.
+		ourBalance -= amount
+	}
```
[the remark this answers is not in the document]
guggero: I know this part is a bit hard to grasp, but I copied most of it from the `lnd` codebase, so I'm pretty sure it's correct.
The goal here is to calculate the _new_ balance of a new commitment transaction based on the previous commitment transaction's balances. And the HTLCs here are basically the diff between the two. So if I send you an HTLC it comes out of my balance and if you send me an HTLC it comes out of your pocket.

guggero:
Rebased on top of https://github.com/lightninglabs/taproot-assets/pull/904, so please skip the first 21 commits when reviewing @GeorgeTsagk  (will rebase again as soon as https://github.com/lightninglabs/taproot-assets/pull/904 makes it through the merge queue).

tapchannel/allocation.go
```diff
@@ -0,0 +1,526 @@
...
+
+// Allocation is a struct that tracks how many units of assets should be
+// allocated to a specific output of an on-chain transaction. An allocation can
+// be seen as a recipe/instruction to distribute a certain number of asset units
+// to a specific output of an on-chain transaction. The output is mainly
+// identified by its output index but also carries along additional information
+// that is required for making sure the resulting on-chain outputs can be sorted
+// in a deterministic way (that is almost but not exactly following the BIP-69
+// rules for sorting transaction outputs).
+type Allocation struct {
```
GeorgeTsagk, line 88: is this the equivalent of locking utxos on lnd?

after reading godoc looks like it's only about on-chain assets but the allocation types indicate that we're also going to reflect tap channel htlc balance changes via this `Allocation` struct?

tapchannel/allocation.go
```diff
@@ -0,0 +1,526 @@
...
+}
+
+// DistributeCoins allocates a set of inputs (extracted from the given input
+// proofs) to virtual outputs as specified by the allocations given. It returns
+// a list of virtual packets (one for each distinct asset ID) with virtual
+// outputs that sum up to the amounts specified in the allocations. The main
+// purpose of this function is to deterministically re-distribute heterogeneous
+// asset outputs (asset UTXOs of different sizes from different tranches/asset
+// IDs) according to the distribution rules provided as "allocations".
+func DistributeCoins(inputs []*proof.Proof, allocations []*Allocation,
```
GeorgeTsagk, line 315: this can be used outside of tap channel context too, correct?

tapchannelmsg/records.go
```diff
@@ -0,0 +1,1568 @@
...
+)
+
+const (
+	// MaxNumOutputs is the maximum number of asset outputs that are allowed
+	// in a single record. This mainly affects the maximum number of asset
+	// UTXOs that can reside within a single commitment. This number should
+	// in practice be very small (probably close to 1), as all outputs must
+	// be from the same asset group but from different tranches to be
+	// encoded as an individual record.
+	MaxNumOutputs = 2048
```
GeorgeTsagk, line 27: godoc recommends a value close to `1`, but isntead 2048 was chosen

what's the rationale behind this value?

GeorgeTsagk: nit: wrap it in a const

tapchannel/aux_leaf_creator.go
```diff
@@ -0,0 +1,282 @@
...
+		cfg: cfg,
+		ContextGuard: &fn.ContextGuard{
+			DefaultTimeout: DefaultTimeout,
+			Quit:           make(chan struct{}),
+		},
+	}
+}
+
+// Start attempts to start a new aux leaf creator.
+func (c *AuxLeafCreator) Start() error {
```
GeorgeTsagk, line 59: after a glimpse I don't get why does this need to be live?

tapchannelmsg/custom_channel_data.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// ChannelCustomData represents the data that is returned in the
+// CustomChannelData field of a lnrpc.Channel object.
+type ChannelCustomData struct {
+	OpenChan    OpenChannel
+	LocalCommit Commitment
+}
+
+// AsJson returns the JSON representation of the channel custom data.
```
GeorgeTsagk, line 62: why do we bother touching JSON? i'm missing something, this is not user-exposed data right?

GeorgeTsagk (approved):
LGTM :100: 
have a few Qs for more context

tapchannel/allocation.go
guggero: Sure. But the main reason to use this is to have a deterministic way of distribution that both parties can use (in LN only the TXID of the transactions are exchanged between the peers, so they need to come up with exactly the same TX down to the last bit). So I think this mostly makes sense in the context of Lightning channels. But it might turn out to be useful for other stuff too.

tapchannelmsg/records.go
guggero: Just didn't want to introduce an upfront limitation. This is just to make sure we aren't memory-DoSed when decoding p2p messages.

tapchannelmsg/custom_channel_data.go
guggero: It's to render the custom TLV encoded data as JSON on the CLI and RPC/REST interfaces. So it takes the TLV data and renders it as JSON on the RPC output, so on the CLI for example we just have to turn the raw bytes into a string and format the nested JSON.

tapchannel/aux_leaf_creator.go
guggero: Hmm, you're right. Currently we don't require any state. But I think that might come with some later changes, so I'd like to keep it in (also to keep all the components looking and behaving similarly).

PR #882, opened 2024-04-16
[custom channels 3/3]: add FundingController to manage channel asset funding 

Roasbeef:
In this PR, we add a new sub-system, that `FundingController` which is tasked with managing the asset channel funding process from start to finish. The new funding sub-system will communicate directly with an lnd node (in-process or via RPC) to drive the funding process with an externally funded PSBT template. We'll interact with the local `AssetWallet` to funding+sign the PSBT on all levels, then ask lnd to validate what we've created, before finally broadcasting the final transaction. Before we kick off the funding process, we'll send a series of input validation proofs to the remote party. They'll validate these proofs (sent as custom messages) to ensure that we can actually sign for the assets sent. 

Link to all PRs in the saga:
 - 4/3: https://github.com/lightninglabs/taproot-assets/pull/912

## TODO 

- [x] The PSBT returned from lnd will have the funding output append to it. As is, we're able to construct the funding output entirely ourselves, but the internal key isn't yet known, so the default anchoring process will arrive at the wrong output. We may need also have lnd return the musig2 internal key for this process to work properly. 
- [x] The default anchor txn logic will make a series of dummy outputs (1k sats). However, the channel may be funded with an output larger than that to be able to pay for co-op close fees. We'll need to be able to pass a custom amount. This is related to the above TODO
- [x] Add a concrete implementation of `PsbtChannelFunder`


tapchannel/aux_funding.go
```diff
@@ -0,0 +1,933 @@
...
+		}
+	})
+}
+
+// newCommitBlob creates a new commitment blob that'll be stored in the channel
+// state for the specified party.
+func newCommitBlob(chanAssets assetOutputListRecord,
+	local bool) ([]byte, error) {
+
+	var commit Commitment
```
guggero, line 307: We need to create proper proofs here, using the new commitment helper functions in my PR. Basically a single `Allocation` where all the funds go to one side. So we'll also need to know all funding keys and whether we're going to add (non-asset) commitment CPFP anchor outputs (which we'll need to create exclusion proofs for).

tapchannel/aux_funding.go
```diff
@@ -0,0 +1,933 @@
...
+			tempPID := assetProof.TempChanID.Val
+
+			// TODO(rosabeef): verify that has challenge witness
+			// before?
+
+			// Next, we'll validate this proof to make sure that
+			// the initiator is actually able to spend these
+			// outputs in the funding transaction.
+			ctxb := context.Background()
+			_, err := assetProof.Proof.Val.Verify(
```
guggero, line 764: Oh, I just noticed something: In `Verify` we only check _either_ the challenge witness _or_ the state transition witness, even if both is present. I guess we should change that, otherwise I could smuggle in an asset that I can sign for but that was never authorized to belong to me.

tapchannel/records.go
```diff
@@ -384,6 +386,17 @@ func (c *Commitment) Decode(r io.Reader) error {
 	return tlvStream.Decode(r)
 }
 
+// Bytes returns the serialized Commitment record.
+func (c *Commitment) Bytes() []byte {
```
guggero, line 390: Lots of rebase artifacts in this file, you can probably get rid of all the diffs except for this method.

guggero (commented):
Awesome, looks very good! Left some suggestions and clarifications to make sure we have the same flow in mind.

tapchannel/aux_funding.go
```diff
@@ -0,0 +1,933 @@
...
+	err = f.cfg.AssetWallet.SignPassiveAssets(passiveAssets)
+	if err != nil {
+		return nil, fmt.Errorf("unable to sign passive assets: %w", err)
+	}
+
+	// Now that all the passive+active assets have been signed, we'll call
+	// into the wallet to: do the PSBT funding, add all the extra meta data
+	// to the PSBT, ask lnd to sign it, and also generate any new proof
+	// suffixes needed.
+	finalFundingTx, err := f.cfg.AssetWallet.AnchorVirtualTransactions(
```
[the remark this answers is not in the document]
Roasbeef: Ah yeah, this'll do all the funding, but then exclude the PSBT output that lnd wants to add to it. 

What this _should_ do is:
  * With our asset PSBT inputs, and also the PSBT output `lnd` has added, we'll do funding on the Bitcoin level 
  * We'll then also ask lnd to sign (relying on the assumption that all the tapscript root + key desc info is set in the inputs) 

> But I assume the call to f.cfg.ChannelFunder.OpenChannel() will do the actual PSBT funding under the hood, right?

So this'll just call `OpenChannel`, then wait for the update that includes the updated PSBT template. In theory, we don't need to pass one in, as we just want to know the internal key + script from `lnd`. So we can pass in a blank template, then append that output information to the new PSBT we create.

tapchannel/aux_funding.go
```diff
@@ -0,0 +1,933 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to open channel: %v", err)
+	}
+
+	// With the asset intent back from lnd, we'll now be able to access the
+	// funding PSBT that includes the funding output lnd wants. We already
+	// computed this ourself, so we'll snip one off.
+	//
+	// TODO(roasbeef): PrepareAnchoringTemplate adds a dummy output, but
+	// lnd knows what it wants already
```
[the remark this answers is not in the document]
Roasbeef: > I'd imagine the logic there doing something like this:

> Call into lnrpc.OpenChannel with psbt: true and the custom blob

So we don't pass in the custom blob here, but specify the `tempPID`. Then `lnd` will use that to call into `DescFromPendingChanID` to arrive at the correct tapscript root. 

Other than that, yeah this is the flow I have in mind. 

> The funding PSBT template now has two outputs, funding output (committing to a split output) and asset change (committing to the asset split root) and one asset input

What currently takes care of adding the asset change output? I think there's a bit of a dependency there as if there _is_ going to be change, then that impacts the tapscript root we'll return from `DescFromPendingChanID`.

tapchannel/aux_funding.go
```diff
@@ -0,0 +1,933 @@
...
+// pending asset channel funding flow.
+type pendingAssetFunding struct {
+	peerPub btcec.PublicKey
+
+	pid funding.PendingChanID
+
+	initiator bool
+
+	// TODO(roasbeef): should be the new Either identifier?
+	assetID asset.ID
```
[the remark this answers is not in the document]
Roasbeef: Yeah just to simplify for now, single asset funding, will still make sure things can be generalized though after I get the initial version working e2e.

tapchannel/aux_funding.go
```diff
@@ -0,0 +1,933 @@
...
+
+// assetRootFromInputs computes the asset root from the set of inputs provided.
+// This'll be used to identify the set of assets that'll be used as funding
+// inputs into the channel.
+func assetRootFromInputs(inputs []*proof.Proof) (*mssmt.BranchNode, error) {
+	ctxb := context.Background()
+
+	// Insert all the assets into a new SMT that'll commit to all the
+	// assets we plan to use as input to funding.
+	fundingTree := mssmt.NewCompactedTree(mssmt.NewDefaultStore())
```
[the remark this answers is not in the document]
Roasbeef: So this is a root that's used as a sanity check during the funding process. It'll be sent along with open+accept channel which helps both sides verify that they're about to create/expect the same funding output. I'll rename it. 

When we go to create the actual TAP funding output, we do call `commitment.FromAssets`: https://github.com/lightninglabs/taproot-assets/pull/882/files/1bf0eef4261a8b36d989b01728db95f7129f58c3#diff-dfb077e35b4e69617e526023958915ea56d7eb703fd25890976b5a487831c5eeR402-R408

tapchannel/aux_funding.go
Roasbeef: Why are all the proofs needed at that level? My mental model is that here we just need to know which side owns which assets. 

The commitment key ring isn't yet known at this stage (that's where the aux leaf stuff comes into play).

tapchannel/records.go
Roasbeef: Yeah not sure what happened here, will clean it up.

tapchannel/aux_funding.go
Roasbeef: So here I always pass in `nil` for the prev asset, which _should_ force just the challenge witness validation. Can you give a more complete example of the smuggle edge case?

guggero: Discussed offline, we'll check the full asset provenance at another point, here we just need to make sure the remote party can sign for the assets.

tapchannel/allocation.go
```diff
@@ -424,6 +424,13 @@ func AssignOutputCommitments(allocations []*Allocation,
 
 	for idx := range allocations {
 		alloc := allocations[idx]
+
+		// Allocations without any assets won't be mapped to an output
+		// commitment.
+		if alloc.Type == AllocationTypeNoAssets {
```
guggero, line 430: Oops... Nice catch!

tapchannel/aux_funding.go
```diff
@@ -752,28 +752,39 @@ func (f *FundingController) anchorVPackets(fundedPkt *tapsend.FundedPsbt,
 func (f *FundingController) signAndFinalizePsbt(ctx context.Context,
 	pkt *psbt.Packet) (*wire.MsgTx, error) {
 
-	log.Infof("Signing and finalizing PSBT w/ lnd")
+	log.Infof("Signing and finalizing PSBT w/ lnd: %v", spew.Sdump(pkt))
 
-	signedPkt, err := f.cfg.ChainWallet.SignAndFinalizePsbt(ctx, pkt)
+	// By default the wallet won't try to finalize output it sees are watch
+	// only (like the asset input), so we'll have it sign ourselves first.
+	signedPkt, err := f.cfg.ChainWallet.SignPsbt(ctx, pkt)
```
guggero, line 759: Ah, yeah, this is probably something I could've improved in the documentation of the `SignPsbt` and `FinalizePsbt` calls. The former needs to be used for any custom signing where special instructions (like a merkle root) are present in the PSBT. The latter can only sign for "normal", keyspend wallet inputs.

guggero (commented):
Super exciting!

guggero:
Ready for full review.

tapchannelmsg/wire_msgs.go
```diff
@@ -0,0 +1,354 @@
...
+	TxAssetOutputProofType lnwire.MessageType = 32770
+
+	// AssetFundingCreatedType is the message type of the
+	// AssetFundingCreated message.
+	AssetFundingCreatedType lnwire.MessageType = 32771
+
+	// AssetFundingAckType is the message type of the AssetFundingAck
+	// message.
+	AssetFundingAckType lnwire.MessageType = 32772
+)
```
ffranr, line 32: These values need to be modified such that they satisfy the following criteria:
1. They need to be specific to taproot-assets (lower probability of collision with other apps).
2. They need to avoid collision with the RFQ wire messages.

See here fore how that's done for RFQ: https://github.com/lightninglabs/taproot-assets/blob/1f580d7b9583887dc215b21f30eb3498e4b4dc56/rfqmsg/messages.go#L53-L61

The methodology for RFQ wire message types is also broken. It should be:
```
// This offset was chosen as the concatenation of the alphabetical index
// positions of the letters "t" (20), "a" (1), and "p" (16).
const TapMessageTypeBaseOffset = 20116 + 18617 + lnwire.CustomTypeStart
```

where the 18617 is new and comes from alphabet index pos of `R`, `F`, and `Q`.

ffranr: For this PR we need 
```
const TapMessageTypeBaseOffset = 20116 + <package-specific-offset> + lnwire.CustomTypeStart
```

tapchannelmsg/wire_msgs.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+// AssetFundingAck is sent by the responder of the funding flow after they've
+// received the funding input and output proofs. If the responder is able to
+// fully validate the proposed funding parameters, then they'll send this method
+// with accept=true. Otherwise, they'll send with accept set to false.
+type AssetFundingAck struct {
+	// TempChanID is the temporary channel ID that was assigned to the
+	// channel.
+	TempChanID tlv.RecordT[tlv.TlvType0, funding.PendingChanID]
```
ffranr, line 296: is this always exactly the "pending" channel ID? If so can we just name it `PendingChanID` for consistency?

server.go
```diff
@@ -768,6 +783,31 @@ func (s *Server) InlineParseCustomData(msg proto.Message) error {
 	return cmsg.ParseCustomChannelData(msg)
 }
 
+// Name returns the name of this endpoint. This MUST be unique across all
+// registered endpoints.
+//
+// NOTE: This method is part of the protofsm.MsgEndpoint interface.
+func (s *Server) Name() protofsm.EndpointName {
+	return tapchannel.MsgEndpointName
```
ffranr, line 791: The return value here doesn't seem appropriate to me. `Server` is the main taproot asset server but `tapchannel.MsgEndpointName` is specific to channel funding: "taproot assets channel funding".

server.go
```diff
@@ -768,6 +783,31 @@ func (s *Server) InlineParseCustomData(msg proto.Message) error {
...
+}
+
+// CanHandle returns true if the target message can be routed to this endpoint.
+//
+// NOTE: This method is part of the protofsm.MsgEndpoint interface.
+func (s *Server) CanHandle(msg protofsm.PeerMsg) bool {
+	<-s.ready
+	return s.cfg.AuxFundingController.CanHandle(msg)
```
ffranr, line 799: This checks for handling for the channel funding wire messages but not the RFQ messages. Given that this method is on the tap server I would expect it to validate all tap relevant wire messages. Should it?

What's the motivation of having these methods as part of `Server` and not as part of a sub-service?

rpcserver.go
```diff
@@ -6390,6 +6391,46 @@ func (r *rpcServer) SubscribeRfqEventNtfns(
 	)
 }
 
+// FundChannel initiates the channel funding negotiation with a peer for the
+// creation of a channel that contains a specified amount of assets of a given
+// asset.
```
ffranr, line 6396:
```
specified amount of assets of a given
// asset.
```

should be?

```
specified amount of a given asset.
```

taprpc/taprootassets.proto
```diff
@@ -128,6 +128,13 @@ service TaprootAssets {
     */
     rpc SubscribeSendEvents (SubscribeSendEventsRequest)
         returns (stream SendEvent);
+
+    /*
+    FundChannel initiates the channel funding negotiation with a peer for the
+    creation of a channel that contains a specified amount of assets of a given
+    asset.
+    */
+    rpc FundChannel (FundChannelRequest) returns (FundChannelResponse);
```
ffranr, line 137: It would be nice if we could keep all channel specific RPC endpoints to a single channel RPC package.

ffranr (changes requested):

server.go
guggero: As discussed offline, the whole of `tapd` is just a single message endpoint. Ideally we'd want this value to be returned from the `AuxFundingController`. But that instance isn't ready to be called into at the moment we get called here, so we just return the same constant as the `AuxFundingController` also would.

server.go
guggero: The main reason we don't want to handle the RFQ messages in the same way is that it would mean we couldn't run `tapd` in standalone mode anymore, only bundled through `litd` (which connects this hook).
If we use the custom message RPC of `lnd` for the RFQ messages, we are independent of the mode `tapd` is running in (standalone with remote `lnd` or bundled in `litd` in integrated mode).

So this short cut is very specific to the funding process which is only really relevant in the litd integrated mode.

To your second question: The reason this is on the main `Server` instance is the way we set up those hooks in `litd`: https://github.com/lightninglabs/lightning-terminal/pull/761/commits/03479d12759d0600d33956c5df527dae97eb29e1
The `Server` struct is what we instantiate within `tapd`, so we want it to implement all the aux interfaces directly (which it will then proxy to the correct components once the correct state is achieved).
I agree that this isn't super nice and should probably be made a bit nicer in the future. But to be honest, for now I don't really have a much better idea on how to do this, without needing to add a lot more `tapd` specific code to `litd`.

tapchannelmsg/wire_msgs.go
guggero: I think it is, yes. Renamed :+1:

tapchannelmsg/wire_msgs.go
ffranr: As @guggero mentioned on keybase:
`32768 < p2p custom message type int < 65535`

Therefore, this won't work:
```
const TapMessageTypeBaseOffset = 20116 + 18617 + lnwire.CustomTypeStart
```

tapchannelmsg/wire_msgs.go
```diff
@@ -0,0 +1,354 @@
...
+	return t.TempChanID.Val
+}
+
+// FundingAssetID returns the asset ID of the underlying asset.
+func (t *TxAssetInputProof) FundingAssetID() fn.Option[asset.ID] {
+	return fn.Some(t.AssetID.Val)
+}
+
+// Amt returns the amount of the asset that this output represents.
+func (t *TxAssetInputProof) Amt() fn.Option[uint64] {
```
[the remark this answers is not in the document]
guggero: I think it's because some messages don't actually have an amount, but each message needs to implement this to satisfy the interface. IMO the interface itself is a bit awkward, at least since the addition of the `AssetFundingAck`. But changing this would lead to quite a bit of refactor, so I think we should do that at a later point in time.

tapchannelmsg/wire_msgs.go
guggero: Good catch, I forgot about those numbers. I updated them to be `256` higher than the RFQ messages, that should give us enough space to add new messages there.

perms/perms.go
```diff
@@ -256,6 +256,10 @@ var (
 			Entity: "rfq",
 			Action: "write",
 		}},
+		"/tapchannelrpc.TaprootAssetChannels/FundChannel": {{
+			Entity: "assets",
```
ffranr, line 260: Maybe `Entity` should be set to `channels` here?

ffranr (approved):
Just one comment here to consider: https://github.com/lightninglabs/taproot-assets/pull/882#discussion_r1616174342

GeorgeTsagk (approved):
:heavy_check_mark:

PR #885, opened 2024-04-17
[custom channels 2/3]: add aux leaf signer

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/878.

Implements a first version of the aux leaf signer.

Link to all PRs in the saga:
 - 4/3: https://github.com/lightninglabs/taproot-assets/pull/912


tapchannelmsg/records.go
```diff
@@ -736,6 +815,358 @@ func DecodeHtlcAuxLeaf(blob tlv.Blob) (*HtlcAuxLeaf, error) {
...
+type AssetSig struct {
+	// AssetID is the asset ID that the signature is for.
+	AssetID tlv.RecordT[tlv.TlvType0, asset.ID]
+
+	// Sig is the signature for the asset spend.
+	Sig tlv.RecordT[tlv.TlvType1, lnwire.Sig]
+
+	// SigHashType is the sigHash type that was used to create the
+	// signature.
+	SigHashType tlv.RecordT[tlv.TlvType2, uint32]
```
ffranr, line 829: Use `txscript.SigHashType` here instead of `uint32`?

tappsbt/decode_test.go
```diff
@@ -3,19 +3,26 @@ package tappsbt
...
+	"github.com/lightninglabs/taproot-assets/commitment"
 	"github.com/lightninglabs/taproot-assets/internal/test"
+	"github.com/lightninglabs/taproot-assets/mssmt"
 	"github.com/stretchr/testify/require"
 )
 
 var (
 	generatedTestVectorName = "psbt_encoding_generated.json"
 
+	packetHexFileName = filepath.Join(testDataFileName, "packet.hex")
```
ffranr, line 24: It might be useful to add a comment here to describe how the contents of `packet.hex` was derived (whether it is a real world example or not).

ffranr (approved):
Some simple lint errors, otherwise :+1:

tapchannelmsg/records.go
guggero: This would lead to `panic: unknown primitive type: *txscript.SigHashType [recovered]`. We either need to have types that have a `Record()` method or are from a list of defined primitive types (no type aliases allowed).

tappsbt/decode_test.go
guggero: Makes sense, added.

GeorgeTsagk (approved):
Lgtm :leafy_green:

tapchannel/auf_leaf_signer_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	case <-time.After(testTimeout):
+		t.Fatalf("timeout waiting for response")
+	}
+}
+
+type mockVirtualSigner struct {
+}
+
+func (m *mockVirtualSigner) SignVirtualPacket(vPkt *tappsbt.VPacket,
```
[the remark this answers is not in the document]
guggero: Fixed!

tapcfg/server.go
```diff
@@ -423,6 +429,7 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 		UniverseQueriesBurst:     cfg.Universe.UniverseQueriesBurst,
 		RfqManager:               rfqManager,
 		AuxLeafCreator:           auxLeafCreator,
+		AuxLeafSigner:            auxLeafSigner,
```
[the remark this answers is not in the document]
guggero: We're always creating and starting these components. Since they're pretty light-weight, it IMO doesn't make sense to only conditionally create and start them for the custom channel specific use case, as that would mean we'd need to deal with more `.WhenSome()` and `.IsSome()` constructs.

But maybe something to consider in this regard is to only start the HTLC and invoice acceptors if they're really needed, since both of them can only be registered once in `lnd`. So if we always occupy that single "slot" with `tapd` even if you don't use custom channels, that could create issues for some users. I'll look into that in the invoice PR.

Issue #908, opened 2024-05-20
[feature]: add version field to RFQ wire messages


ffranr:
### [Feature]: Add Version Field to RFQ Wire Messages

**Description:**
Request For Quote (RFQ) wire messages should include version fields.

**Rationale:**
Versioning for these messages is important because they are passed between peers that may be running incompatible `tapd` versions. Version fields will allow flexibility to modify the contents of RFQ messages in the future. Each RFQ message's version should be incremented separately, allowing the reject message to remain at a low version number for longer. The reject message can then be used to cleanly reject unsupported message versions. 

**Modifications:**
- Most modifications should be restricted to the `rfqmsg` package.
- Extend the `Validate` method for each message data to check the version number. Unsupported version numbers should fail validation. Refer to this example [`Validate` method](https://github.com/lightninglabs/taproot-assets/blob/057880e788247bf1417596a34b420e2ca3159e29/rfqmsg/buy_request.go#L147).
- The version TLV field should be the first field in a message data TLV stream.
- The BLIP does not currently specify version fields for RFQ messages. Refer to the [BLIP document](https://github.com/Roasbeef/blips/blob/tap-blip/blip-tap.md#request-for-quote-tap_rfq). The BLIP should be updated.

**Alternative Strategy:**
An alternative to version fields is to increment the p2p message type integer (similar to LND "custom messages"), defining completely new and distinct wire messages. However, this strategy complicates rejection of future versions by older `tapd` nodes. Therefore, IMO, we shouldn't depending on p2p message types.

dstadulis:
Thinking that this should be a bit field akin to verisonBits so RFQ version cross-compatibility can be more tailored between clients

PR #911, opened 2024-05-21
Add version fields to RFQ wire messages

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/908

### DIFF
--- a/rfqmsg/buy_accept.go
+++ b/rfqmsg/buy_accept.go
@@ -15,12 +15,22 @@ import (
 const (
 	// Buy accept message type field TLV types.
 
-	TypeBuyAcceptID        tlv.Type = 0
-	TypeBuyAcceptAskPrice  tlv.Type = 2
-	TypeBuyAcceptExpiry    tlv.Type = 4
-	TypeBuyAcceptSignature tlv.Type = 6
-	TypeBuyAcceptAssetID   tlv.Type = 8
+	TypeBuyAcceptVersion   tlv.Type = 0
+	TypeBuyAcceptID        tlv.Type = 2
+	TypeBuyAcceptAskPrice  tlv.Type = 4
+	TypeBuyAcceptExpiry    tlv.Type = 6
+	TypeBuyAcceptSignature tlv.Type = 8
+	TypeBuyAcceptAssetID   tlv.Type = 10
 )
+
+func TypeRecordBuyAcceptVersion(version *WireMsgDataVersion) tlv.Record {
+	const recordSize = 1
+
+	return tlv.MakeStaticRecord(
+		TypeBuyAcceptVersion, version, recordSize,
+		WireMsgDataVersionEncoder, WireMsgDataVersionDecoder,
+	)
+}
 
 func TypeRecordBuyAcceptID(id *ID) tlv.Record {
 	const recordSize = 32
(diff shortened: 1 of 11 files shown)


rfqmsg/reject.go
```diff
@@ -102,9 +112,18 @@ var (
 	}
 )
 
+const (
+	// latestRejectVersion is the latest supported reject wire message data
+	// field version.
+	latestRejectVersion = V0
```
guggero, line 118: nit: shouldn't this just be "lastKnownVersion"? Since we won't reject V0.

guggero: Or just `latestVersion` similar to the name used in the next commit.

guggero (commented):
Nice!

rfqmsg/buy_request_test.go
```diff
@@ -46,6 +48,7 @@ func TestBuyRequestMsgDataEncodeDecode(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.testName, func(tt *testing.T) {
 			req := buyRequestMsgData{
+				Version:       tc.version,
```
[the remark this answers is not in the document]
ffranr: In this PR, version checking happens after decoding (i.e. in `NewBuyRequestMsgFromWire`). So it shouldn't show up as an error in the encoding/decoding unit test. But I could add a non-zero value as version anyway here.

rfqmsg/reject.go
ffranr: Here I attempted to use "latest version" constants which were specific to each message type. They are all in the same package so I think they need unique names.

guggero: Oh, I see... The word "reject" refers to the message type, not that we reject this version... Okay, never mind.

rfqmsg/buy_request_test.go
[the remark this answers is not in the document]
guggero: Yeah, that's okay. I mean that if we only encode the default value of 0, we don't actually test that any data is really encoded. So only when we add a non-zero value do we know encoding/decoding works on that field.

rfqmsg/buy_accept.go
```diff
@@ -15,13 +15,23 @@ import (
...
-	TypeBuyAcceptAskPrice  tlv.Type = 2
-	TypeBuyAcceptExpiry    tlv.Type = 4
-	TypeBuyAcceptSignature tlv.Type = 6
-	TypeBuyAcceptAssetID   tlv.Type = 8
+	TypeBuyAcceptVersion   tlv.Type = 0
+	TypeBuyAcceptID        tlv.Type = 2
+	TypeBuyAcceptAskPrice  tlv.Type = 4
+	TypeBuyAcceptExpiry    tlv.Type = 6
+	TypeBuyAcceptSignature tlv.Type = 8
+	TypeBuyAcceptAssetID   tlv.Type = 10
```
Roasbeef, line 23: Can we consolidate the buy/sell messages with what we had outlined here? https://github.com/Roasbeef/blips/pull/3

ffranr: I'll spin that out into another PR. I think this PR is ready to be merged.

rfqmsg/messages.go
```diff
@@ -115,6 +116,45 @@ func NewIncomingMsgFromWire(wireMsg WireMessage) (IncomingMsg, error) {
...
+	// V0 represents version 0 of the contents in a wire message data field.
+	V0 WireMsgDataVersion = 0
+)
+
+// WireMsgDataVersionEncoder is a function that can be used to encode a
+// WireMsgDataVersion to a writer.
+func WireMsgDataVersionEncoder(w io.Writer, val any, buf *[8]byte) error {
+	if version, ok := val.(*WireMsgDataVersion); ok {
+		versionUint8 := uint8(*version)
+		return tlv.EUint8(w, &versionUint8, buf)
```
jharveyb, line 133: nit: why not `EUint8T` like in `asset.VersionEncoder`?

```go
return tlv.EUint8T(w, uint8(*version), buf)
```

jharveyb (approved):
LGTM 💯

guggero (approved):

rfqmsg/messages.go
ffranr: It should have been `tlv.EUint8T`. I'll try to keep that in mind in the future. I didn't get a chance to make that modification in this PR.

jharveyb: Sg, could be in the follow-up PR you mentioned above.

Issue #906, opened 2024-05-17
[feature]: add support for RFQ price oracle RPC service


ffranr:
We need to implement modifications related to the price oracle component of the RFQ service.

#### Modifications Required

1. **Command Line Flag `--rfq.acceptall`**
   - Add the command line flag `--rfq.acceptall`.
   - This flag is optionally specified on `tapd` startup.
   - When specified, the RFQ service will accept the peer-suggested rate tick during any RFQ negotiation. It will also accept any quote regardless of subject asset.
   - This should be a development flag and should be clearly indicated as such.

2. **Command Line Argument `--rfq.oracleurl`**
   - Add the command line argument `--rfq.oracleurl`.
   - This is an optional command line argument.
   - Users can specify a price oracle service URL using this argument.
   - The specified URL will be queried whenever the RFQ service needs to query a price oracle.
   - The price oracle URL must be an RPC service.

3. **Default Behavior**
   - If neither `--rfq.acceptall` nor `--rfq.oracleurl` are specified, the RFQ service should reject any RFQ requests.
   - Log a warning message indicating the lack of necessary configuration.

ffranr:
Any solution to this issue will need to build upon this (currently open) PR: https://github.com/lightninglabs/taproot-assets/pull/904

ffranr:
I think `--rfq.acceptall` should be a dev only flag. Should the flag be called `--dev.rfq.acceptall` or something like that? Is that a pattern that we use anywhere? And we can hide it behind the golang compile `dev` tag also.

guggero:
Discussed offline:

 - Add ShouldAcceptBidPrice and ShouldAcceptAskPrice methods to price oracle interface, mock implementation just returns true
 - if --experimental.rfq.acceptallprices is set, we _don't_ call into the above methods
 - add a new flag --experimental.rfq.priceoracleaddress (should match rfqrpc://<hostname>:<port>)
   - if the flag is not set, then we use an implemetations of the price oracle mock
   - the static usd/asset price of the mock oracle can be configured with --experimental.rfq.oracleusdperbtc
   - if the flag is set and is a proper URI with the rfqrpc:// prefix, then we use that implementation
   - the Oracle RPC is defined in `taprpc/rfqrpc/oracle.proto` with a service called Oracle
 - the default value for --experimental.rfq.priceoracleaddress is empty, which results in an error ("must specify price oracle address)
[...]

PR #915, opened 2024-05-23
Add RPC price oracle service

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/906

I have minimized changes to ensure this can be included in version 0.4 with minimal review effort.


dstadulis:
Scope One Endpoint

let's code work as it currently works
let's users use their own price oracle

Aperture will be a helpful testing apparatus
In unit tests, can define struct that set 
dummy grpc server

itest/tapd_harness.go
```diff
@@ -213,6 +214,14 @@ func newTapdHarness(t *testing.T, ht *harnessTest, cfg tapdConfig,
 	// was not set, this will be false, which is the default.
 	tapCfg.AddrBook.DisableSyncer = opts.addrAssetSyncerDisable
 
+	// Set the experimental config for the RFQ service.
+	tapCfg.Experimental = &tapcfg.ExperimentalConfig{
+		Rfq: rfq.CliConfig{
+			PriceOracleAddress:  rfq.MockPriceOracleServiceAddress,
+			MockOracleUsdPerBtc: 5820600,
```
Roasbeef, line 221: Style nit: 
```go
MockOracleUsdPerBtc: 5_820_600,
```

rfq/cli.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// CliConfig is a struct that holds tapd cli configuration options for the RFQ
+// service.
+//
+// nolint: lll
+type CliConfig struct {
+	AcceptAllPrices bool `long:"acceptallprices" description:"Skip price oracle validation and accept all prices"`
+
+	PriceOracleAddress string `long:"priceoracleaddress" description:"Price oracle gRPC server address (rfqrpc://<hostname>:<port>)"`
```
Roasbeef, line 19: Do we actually need the custom URI here?

rfq/oracle.go
```diff
@@ -60,6 +61,31 @@ type OracleBidResponse struct {
 	Err *OracleError
 }
 
+// ParsePriceOracleAddress parses a price oracle service address string and
+// returns a URL type instance.
+func ParsePriceOracleAddress(addrStr string) (*url.URL, error) {
```
Roasbeef, line 76: If we're keeping the custom URI, then we can add a type alias here to distinguish things a bit more: 
```
type OracleAddr = url.URL
```

tapcfg/config.go
```diff
@@ -317,6 +330,8 @@ type Config struct {
 
 	Prometheus monitoring.PrometheusConfig `group:"prometheus" namespace:"prometheus"`
 
+	Experimental *ExperimentalConfig `group:"experimental" namespace:"experimental"`
```
Roasbeef, line 333: So the full path will be: `--experimental.rfq.acceptallprices`?

rfq/negotiator.go
```diff
@@ -13,6 +14,18 @@ import (
...
 )
 
+const (
+	// defaultRateTickExpirySeconds is the default rate tick expiry lifetime
+	// in seconds.
+	//
+	// TODO(ffranr): This const is currently used in conjunction with the
+	//  AcceptAllPrices flag. It is used to set the expiry time of the rate
+	//  tick in the accept message. This is a temporary solution and should
+	//  be replaced with an expiry time provided by the peer in the quote
```
Roasbeef, line 24: IIRC, we already have the expiry included in the message? 

I think for the final release, we want this to be more like 1 - 10 minutes.

rfq/negotiator.go
```diff
@@ -281,6 +299,26 @@ func (n *Negotiator) HandleIncomingBuyRequest(
...
+			)
+
+			// Construct and send a sell accept message.
+			msg := rfqmsg.NewBuyAcceptFromRequest(
+				request, request.BidPrice, expiryTimestamp,
+			)
+			sendOutgoingMsg(msg)
+
+			return
+		}
```
Roasbeef, line 320: This should auto reject if `AcceptAllPrices` isn't specified. So we want 3 cases: 
  * no oracle specified, reject (default config) 
  * accept all 
  * query the oracle

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -0,0 +1,105 @@
...
+// we multiply that by 10,000 to arrive at the usd_rate_tick:
+// $61,234.95 * 10000 = 612,349,500. To convert back to our normal rate, we
+// decide by 10,000 to arrive back at $61,234.95.
+message RateTick {
+    // rate is the exchange rate between the subject asset and the payment
+    // asset.
+    uint64 rate = 1;
+
+    // precision is the number of decimal places in the exchange rate.
+    uint32 precision = 2;
```
Roasbeef, line 37: Why's this required? Just for display purposes somewhere?

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -0,0 +1,105 @@
...
+    // transaction_type indicates whether the transaction is a purchase or a
+    // sale.
+    TransactionType transaction_type = 1;
+
+    // subject_asset is the asset to be priced for purchase or sale.
+    AssetSpecifier subject_asset = 2;
+
+    // subject_asset_max_amount is the maximum amount of the subject asset that
+    // could be involved in the transaction.
+    uint64 subject_asset_max_amount = 3;
```
Roasbeef, line 74: 👍

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -0,0 +1,105 @@
...
+
+    // payment_asset is the asset used for purchasing or receiving from a sale.
+    AssetSpecifier payment_asset = 4;
```
Roasbeef, line 77: All zeroes being BTC?

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// QueryRateTickSuccessResponse is the successful response to a rate tick query.
+message QueryRateTickSuccessResponse {
+    // rate_tick is the rate tick for the transaction.
+    RateTick rate_tick = 1;
+}
+
+// QueryRateTickErrResponse is the error response to a rate tick query.
+message QueryRateTickErrResponse {
```
Roasbeef, line 91: 👍

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -0,0 +1,105 @@
...
+    // error is the error message.
+    string message = 1;
```
Roasbeef, line 93: Perhaps let's also add some error code here from the getgo?

rfq/oracle.go
```diff
@@ -105,55 +115,183 @@ type PriceOracle interface {
...
+	client := oraclerpc.NewPriceOracleClient(conn)
+
+	return &RpcPriceOracle{
+		client:  client,
+		rawConn: conn,
+	}, nil
+}
+
+// QueryAskPrice returns the ask price for the given asset amount.
+func (r *RpcPriceOracle) QueryAskPrice(ctx context.Context,
```
Roasbeef, line 172: We can move this into the main package along with the init information above, then the RFQ system accepts a fully initialized version of the service.

Roasbeef (changes requested):

guggero (commented):
Nice work, looks very good!

rfq/negotiator.go
ffranr: The RFQ quote request message type does not include an expiry currently. So the suggested rate tick is provided without an expiry.

taprpc/priceoraclerpc/price_oracle.proto
ffranr: Will remove. We can always add it back later if we need greater precision.

taprpc/priceoraclerpc/price_oracle.proto
ffranr: Added to doc.

rfq/oracle.go
ffranr: I'm going to work on a test for this. If there's time before the release I'll take a look at the change you suggest. Let me know if you'd like me to prioritise differently.

rfq/cli.go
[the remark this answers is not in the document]
ffranr: I'm not sure I understand what you mean. If the mock price isn't set the error message is
```
mockoraclecentpersat must be set when using the mock price oracle service
```

rfq/cli.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// CliConfig is a struct that holds tapd cli configuration options for the RFQ
+// service.
+//
+// nolint: lll
+type CliConfig struct {
+	PriceOracleAddress string `long:"priceoracleaddress" description:"Price oracle gRPC server address (rfqrpc://<hostname>:<port>). Mock address: use_mock_price_oracle_service_promise_to_not_use_on_mainnet"`
+
+	SkipAcceptQuotePriceCheck bool `long:"skipacceptquotepricecheck" description:"Skip price validation on RFQ accept quotes"`
```
guggero, line 19: Suggestion for description: "Accept any price quote returned by RFQ peer, skipping validation"?

rfq/cli.go
```diff
@@ -0,0 +1,61 @@
...
+	MockPriceOracleServiceAddress = "use_mock_price_oracle_service_" +
+		"promise_to_not_use_on_mainnet"
+)
+
+// CliConfig is a struct that holds tapd cli configuration options for the RFQ
+// service.
+//
+// nolint: lll
+type CliConfig struct {
+	PriceOracleAddress string `long:"priceoracleaddress" description:"Price oracle gRPC server address (rfqrpc://<hostname>:<port>). Mock address: use_mock_price_oracle_service_promise_to_not_use_on_mainnet"`
```
guggero, line 17: Suggestion for description suffix: "To use the integrated mock, use the following value: "?

rfq/manager.go
```diff
@@ -199,9 +204,10 @@ func (m *Manager) startSubsystems(ctx context.Context) error {
 	// Initialise and start the quote negotiator.
 	m.negotiator, err = NewNegotiator(
 		NegotiatorCfg{
-			PriceOracle:      m.cfg.PriceOracle,
-			OutgoingMessages: m.outgoingMessages,
-			ErrChan:          m.subsystemErrChan,
+			PriceOracle:               m.cfg.PriceOracle,
+			OutgoingMessages:          m.outgoingMessages,
+			SkipAcceptQuotePriceCheck: m.cfg.SkipAcceptQuotePriceCheck, // nolint: lll
```
guggero, line 209: nit: use nolint directive above this line instead of at the end? Below as well.

rfq/oracle.go
```diff
@@ -60,6 +62,43 @@ type OracleBidResponse struct {
...
+	RfqRpcOracleAddrScheme string = "rfqrpc"
+)
+
+// ParsePriceOracleAddress parses a price oracle service address string and
+// returns a URL type instance.
+func ParsePriceOracleAddress(addrStr string) (fn.Option[OracleAddr], error) {
+	addrNone := fn.None[OracleAddr]()
+
+	// Basic sanity check to ensure the address is not empty.
+	if addrStr == "" {
```
guggero, line 97: I think we should restrict the use of `fn.Option` to places where returning a nil error and `fn.None()` is actually possible. Since here we always either return an error or a valid address, we shouldn't use options.

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -0,0 +1,107 @@
...
+    // subject_asset is the asset to be priced for purchase or sale.
+    AssetSpecifier subject_asset = 2;
+
+    // subject_asset_max_amount is the maximum amount of the subject asset that
+    // could be involved in the transaction.
+    uint64 subject_asset_max_amount = 3;
+
+    // payment_asset is the asset used for purchasing or receiving from a sale.
+    //
+    // NOTE: An asset ID of all zeros indicates that the payment asset is BTC.
```
guggero, line 75: Still need to mention that BTC would be used in milli-satoshi for the rate tick to actually mean "milli-satoshi per asset unit".

tapcfg/server.go
```diff
@@ -326,8 +326,30 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
...
+	// real price oracle service.
+	var priceOracle rfq.PriceOracle
+
+	switch cfg.Experimental.Rfq.PriceOracleAddress {
+	case rfq.MockPriceOracleServiceAddress:
+		priceOracle = rfq.NewMockPriceOracle(
+			3600, cfg.Experimental.Rfq.MockOracleCentPerSat,
+		)
+
+	case "":
```
guggero, line 339: This is just for the itest? Because I think if the user leaves the CLI flag empty they would get an error from the `Validate()` method?

rfq/negotiator.go
```diff
@@ -157,16 +157,24 @@ func (n *Negotiator) HandleOutgoingBuyOrder(buyOrder BuyOrder) error {
...
+			// Query the price oracle for a bid price.
+			var err error
+			bidPrice, _, err = n.queryBidFromPriceOracle(
+				*buyOrder.Peer, buyOrder.AssetID,
+				buyOrder.AssetGroupKey, buyOrder.MinAssetAmount,
+			)
+			if err != nil {
+				err := fmt.Errorf("negotiator failed to "+
+					"handle price oracle response: %w", err)
+				n.cfg.ErrChan <- err
```
guggero, line 175: Not sure if it's worth erroring out here? Or just log and continue without a proposed bid price?

rfq/negotiator.go
```diff
@@ -278,6 +278,16 @@ func (n *Negotiator) HandleIncomingBuyRequest(
 		}
 	}
 
+	// Reject the quote request if a price oracle is unavailable.
+	if n.cfg.PriceOracle == nil {
```
guggero, line 282: Circling back to my previous question, shouldn't it be impossible to not have a price oracle configured? At least as a normal user? You either have to use an RPC implementation or the mock?
I think even if the end-user is running with the `--skip...` flag they should be using the mock oracle (it just isn't actually queried, bit it's always hooked up internally).

guggero (commented):
Nice work, seems to have everything we discussed offline.
Just a couple of minor questions and suggestions.

rfq/negotiator.go
```diff
@@ -445,6 +465,101 @@ func (n *Negotiator) HandleOutgoingSellOrder(order SellOrder) {
...
+			return
+		}
+
+		// Ensure that the peer provided price is reasonable given the
+		// price provided by the price oracle service.
+		if !n.pricesWithinBounds(msg.AskPrice, oraclePrice) {
+			// The price is not within the acceptable tolerance.
+			// We will return without calling the quote accept
+			// callback.
+			log.Debugf("buy accept price is not within "+
```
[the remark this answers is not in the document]
ffranr: I wouldn't think of it as an error if the price is unacceptable. Perhaps it should lead to an event being published by the RFQ manager to its subscribers?

rfq/manager.go
ffranr: I'm starting to see a pattern here. Never at the end right?

rfq/negotiator.go
ffranr: I think it should be valid for a user to start tapd without specifying a price oracle address (not even the mock string). A user might not care about lightning channel support.

Roas suggests here https://github.com/lightninglabs/taproot-assets/pull/915#discussion_r1615539072
that if the price oracle isn't specified we just reject all incoming quotes.

rfq/oracle.go
ffranr: Does that mean that the return value becomes a pointer I suppose?

tapcfg/server.go
ffranr: No, I don't think the `Validate` method will return an error if the price oracle address is unset. See https://github.com/lightninglabs/taproot-assets/pull/915#discussion_r1618686552

rfq/negotiator.go
```diff
@@ -23,6 +31,18 @@ type NegotiatorCfg struct {
 	// messages. These are messages which are destined to be sent to peers.
 	OutgoingMessages chan<- rfqmsg.OutgoingMsg
 
+	// AcceptPriceDeviationBp specifies the maximum allowable price
+	// deviation, in basis points. This parameter defines the maximum
+	// percentage deviation that the price returned by the price oracle
+	// service can have from a peer's quote accept price for this node to
+	// use the accepted quote.
+	AcceptPriceDeviationBp uint64
```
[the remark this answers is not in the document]
ffranr: Basis points is very commonly used in finance. Where as I would associate PPM with environmental science or chemistry. So I find LND's use of PPM instead of BP strange. I can see that it's used for the fee rate in LND.

As you suggest, I'll switch to PPM in this PR for consistency.

rfq/negotiator.go
```diff
@@ -445,6 +465,101 @@ func (n *Negotiator) HandleOutgoingSellOrder(order SellOrder) {
...
+	diff := actualExpiry.Unix() - time.Now().Unix()
+	return diff >= minRateTickExpiryLifetime
+}
+
+// priceWithinBounds returns true if the difference between the first price and
+// the second price is within the acceptable bounds.
+func (n *Negotiator) pricesWithinBounds(firstPrice lnwire.MilliSatoshi,
+	secondPrice lnwire.MilliSatoshi) bool {
+
+	deviation := 10000 * math.Abs(float64(secondPrice-firstPrice)) /
```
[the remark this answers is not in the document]
ffranr: Good call. It needed it.

rfq/negotiator.go
```diff
@@ -23,6 +37,18 @@ type NegotiatorCfg struct {
 	// messages. These are messages which are destined to be sent to peers.
 	OutgoingMessages chan<- rfqmsg.OutgoingMsg
 
+	// AcceptPriceDeviationPpm specifies the maximum allowable price
+	// deviation, in parts per million (PPM). This parameter defines the
+	// maximum percentage deviation that the price returned by the price
```
guggero, line 42: nit: s/percentage deviation/allowed deviation/?

rfq/negotiator.go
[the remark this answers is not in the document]
guggero: The problem is that when the user runs an asset channel related command (for example `litcli ln addinvoice` or `litcli ln payinvoice`), the code calls into the RFQ service to create a buy/sell order. That code then waits for either an accept or error to arrive. If nothing happens, because we simply drop/ignore the accepted quote, then the command will time out and the user needs to check the log to find out why.
So I do think we should return an error somehow. Maybe we can differentiate between an actual hard error (something didn't work on the protocol level) and a soft error (something didn't match my expectations).

rfq/negotiator_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"testing"
+
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/stretchr/testify/require"
+)
+
+// TestPricesWithinBounds tests the pricesWithinBounds function.
+func TestPricesWithinBounds(t *testing.T) {
```
guggero, line 11: :+1:

tapcfg/server.go
guggero: Ah, right, forgot about the standalone `tapd` users that don't care about RFQ at all. Makes sense.

rfq/negotiator.go
guggero: Makes sense :+1:

guggero (commented):
Very nice, getting really close! Only one question remains around user experience of returning an error vs. failing silently.
And also if there's time, having a validation test for the RFQ RPC client in a unit test would be nice.

rfq/negotiator.go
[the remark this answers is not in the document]
ffranr: Yes, I agree with what you've said. That's why returning an event would be appropriate. I think the RPC endpoint just blocks waiting for an event anyway. I could be wrong, will look into it.

rfq/negotiator.go
```diff
@@ -625,6 +625,115 @@ func (n *Negotiator) HandleIncomingBuyAccept(msg rfqmsg.BuyAccept,
...
+		finalise(
+			msg, fn.Some[InvalidQuoteRespEvent](
+				*invalidQuoteRespEvent,
+			),
+		)
+
+		return
+	}
+
+	if n.cfg.SkipAcceptQuotePriceCheck {
```
guggero, line 697: Since we allow the price oracle to be `nil` now, we'll run into a panic if the user hasn't set the flag but also hasn't configured the oracle (they just ran stuff with the minimum set they could get things to start with, which is what we need to expect).

In my tests I added the following:
```
	// Reject the quote request if a price oracle is unavailable.
	if n.cfg.PriceOracle == nil {
		invalidQuoteRespEvent := NewInvalidQuoteRespEvent(
			&msg, PriceOracleQueryErrQuoteRespStatus,
		)
		finalise(msg, fn.Some[InvalidQuoteRespEvent](
			*invalidQuoteRespEvent,
		))
		return
	}
```

Which works quite nicely. Though it makes me wish the invalid quote also had an error string that we could give a bit more context or an error message to the user with.

rfq/manager.go
```diff
@@ -756,6 +756,57 @@ func (q *PeerAcceptedBuyQuoteEvent) Timestamp() time.Time {
...
+type InvalidQuoteRespEvent struct {
+	// timestamp is the event creation UTC timestamp.
+	timestamp time.Time
+
+	// QuoteResponse is the quote response received from the peer which was
+	// deemed invalid.
+	QuoteResponse rfqmsg.QuoteResponse
+
+	// Status is the status of the quote response.
+	Status QuoteRespStatus
```
guggero, line 834: See my other comment regarding having a free text or error message here too to give some additional context to the user. But non-blocking, just something for the wish list to improve user experience.

What things look like from my tests for example:
```
[litcli] peer 02ad73dc4471e83179651add7096d238920fefe58ab721b31cc8a25ea575ffe324 sent back an invalid quote, status: INVALID_RATE_TICK
```

Would be nice if we could also display the message that's currently only in the logs (e.g. "Buy accept price is not within acceptable bounds").
But again, non-blocking.

rfq/negotiator.go
guggero: Ah, the same on the `HandleIncomingSellAccept` side, I guess.

tapcfg/server.go
```diff
@@ -326,8 +326,30 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
...
+		// Leave the price oracle as nil, which will cause the RFQ
```
guggero, line 340: Because we allow the price oracle to be nil, we must expect users to run with the minimum flags they can get `litd` to run with. So it will very likely be `nil` for most users. So we need to remove the `nil` check in the `NewNegotiator` function, otherwise `litd` will not start:

```
// NewNegotiator creates a new quote negotiator.
func NewNegotiator(cfg NegotiatorCfg) (*Negotiator, error) {
	// If the price oracle is nil, then we will return an error.
	if cfg.PriceOracle == nil {
		return nil, fmt.Errorf("price oracle is nil")
	}
```

guggero (commented):
Thank you very much for the new messages!
I was able to successfully test the following scenarios (after a couple of small fixes, see inline comments) and get proper feedback on the CLI:

- Both Zane and Yara running with no RFQ flags at all, get expected error: `[litcli] peer 02ad73dc4471e83179651add7096d238920fefe58ab721b31cc8a25ea575ffe324 rejected the quote, code: 1, error message: price oracle unavailable`
- Yara (RFQ service provider) running with mock oracle, Zane with default flags. Get expected error: `[litcli] peer 02ad73dc4471e83179651add7096d238920fefe58ab721b31cc8a25ea575ffe324 sent back an invalid quote, status: PRICE_ORACLE_QUERY_ERR`
- Yara and Zane running with mock oracle but with different prices, got expected error: `[litcli] peer 02ad73dc4471e83179651add7096d238920fefe58ab721b31cc8a25ea575ffe324 sent back an invalid quote, status: INVALID_RATE_TICK`
- Yara and Zane running with mock oracle and same price, everything works correctly

Awesome work, this is great. So as soon as the two mandatory things are fixed, I think we can merge this :rocket:

rfq/negotiator.go
ffranr: > Though it makes me wish the invalid quote also had an error string that we could give a bit more context or an error message to the user with.

ffranr:
Price oracle RPC client unit tests added. Thanks Oli for the TLS RPC server-client fix!

guggero (approved):
Great work, LGTM :tada:

ffranr:
Replaced with https://github.com/lightninglabs/taproot-assets/pull/926
because of merge blocking due to requested changes

PR #919, opened 2024-05-27
tapchannel: add new AuxChanCloser to enable co-op close for asset channels

Roasbeef:
In this Pr, we add a new `AuxChanCloser` that allows co-op close for asset channels. As we haven't fully unified the wallet yet, a "fully loaded" coop close transaction will have 4 output: 2 for the BTC balance, and 2 for the asset amounts. We use a new interface added in `lnd` to allow the aux chan closer to add additional outputs, and also ensure that sorting works out the same on both sides. 

Along the way, we also modify the chan funder to push the funding proof to a Universe if the proof courier addr is specified. This sets us up to ensure that the receiver can properly handle a proof discovery. 

For the initiator, we rely on the porter to push out the proofs for our coop close with the specified proof courier addr (currently global, should allow a side to modify). As the non-initiator, since all our asset outputs will be the result of a split (we force a split for full value sends), we simply import the asset output as an addr into the custodian. This way, the custodian is able to watch for on-chain confirmation, then handle obtaining all the proofs we need to be able to send again.


guggero:
I changed almost all commits and added a couple of my own, which I guess counts as a review as well. LGTM :tada:

guggero (approved):
I changed almost all commits and added a couple of my own, which I guess counts as a review as well. LGTM :tada:

GeorgeTsagk (approved):
My only Q while leaving the diff was how are we able to spend a `declared_known` script?

Looks good!

guggero:
> My only Q while leaving the diff was how are we able to spend a `declared_known` script?

There isn't an easy way yet, at least not without knowing some internals.
I think we'll probably add a new flag to the `FundVirtualPsbt` RPC in the future that allows you to select such an asset.
But for now you'd basically export a proof, then use the `tappsbt.FromProofs()` function to create a vPSBT that you can then spend (which is basically what we're doing in the custom channels use case).

rpcserver.go
```diff
@@ -6441,6 +6442,81 @@ func (r *rpcServer) FundChannel(ctx context.Context,
 	}, nil
 }
 
+// EncodeCustomRecords allows RPC users to encode Taproot Asset channel related
+// data into the TLV format that is used in the custom records of the lnd
+// payment or other channel related RPCs. This RPC is completely stateless and
+// does not perform any checks on the data provided, other than pure format
+// validation.
+func (r *rpcServer) EncodeCustomRecords(_ context.Context,
```
Roasbeef, line 6450: Do we still this as something that's temp? As in we'd just wrap the payment specific stuff in an entirely new RPC?

guggero: I think so, yes. Or maybe not, given that with this it should be pretty easy to do the RPC calls in any language. I guess the question is: how much effort is it for the app devs to call 3 RPCs (buy/sell order, encode records, send payment) with the correct values vs. us threading all the payment and channel flags through some new RPC methods in tapd...

Issue #914, opened 2024-05-23
[feature]: RFQ Wire message updates: split `asset_amt`, implicit asset_id types reduction


dstadulis:
# Background
https://github.com/Roasbeef/blips/pull/3#discussion_r1561808470

# TODO Scope
- [ ] Split the asset_amt field
- [ ] Use only 3 different wire message types (from in situ 5) to handle:
  - [ ] If `in_asset_id` (e.g. USD-X) is only specified
    - [ ] In LN setting, it's a quote for _receiving_ asset_id
  - [ ] If `out_asset_id` is only specified
    - [ ] In LN setting, it's a quote for _sending_ asset_id

Before this change, `asset_amt` was the amount of USD (or w/e) tied to the invoice. So we were asking for a price quote in BTC/asset_id.

# Example user swap flows to confirm wire message reduction
[edit at will]

PR #921, opened 2024-06-01
Reduce RFQ wire message type count

ffranr:
Issue: https://github.com/lightninglabs/taproot-assets/issues/914

This PR includes request message changes. It does not include accept message changes, however, should be ready for review and merge as is.


rfqmsg/messages.go
```diff
@@ -125,6 +125,13 @@ const (
 	V0 WireMsgDataVersion = 0
 )
 
+// Record returns a TLV record that can be used to encode/decode a
+// WireMsgDataVersion to/from a TLV stream.
+func (v *WireMsgDataVersion) Record() tlv.Record {
+	return tlv.MakeStaticRecord(0, v, 1, WireMsgDataVersionEncoder,
```
guggero, line 125: nit: formatting of multi-line function call.

guggero: Also, I know it's perhaps a bit too verbose, but I made it a habit to add the following comment wherever we use a 0 TLV type:
```
	// Note that we set the type here as zero, as when used with a
	// tlv.RecordT, the type param will be used as the type.
```

rfqmsg/request.go
```diff
@@ -0,0 +1,472 @@
...
+
+// newRequestWireMsgDataFromBuy creates a new requestWireMsgData from a buy
+// request.
+func newRequestWireMsgDataFromBuy(q BuyRequest) requestWireMsgData {
+	version := tlv.NewRecordT[tlv.TlvType0](q.Version)
+	id := tlv.NewRecordT[tlv.TlvType1](q.ID)
+
+	// Calculate the expiration unix timestamp in seconds.
+	expiry := tlv.NewPrimitiveRecord[tlv.TlvType2](
+		uint64(time.Now().Add(defaultRequestExpiry).Unix()),
```
guggero, line 108: Kind of feels like this is the wrong place for this logic. Maybe we can set the expiry on the request already? But non-blocking as we don't really use the expiry yet anyway.

rfqmsg/request.go
```diff
@@ -0,0 +1,472 @@
...
+	var suggestedRateTick requestSuggestedTickRate
+	if uint64(q.BidPrice) != 0 {
+		suggestedRateTick = tlv.SomeRecordT[tlv.TlvType4](
+			tlv.NewPrimitiveRecord[tlv.TlvType4](
+				uint64(q.BidPrice),
+			),
+		)
+	}
+
+	inAssetID := requestInAssetID{}
```
guggero, line 122: nit: `var inAsset requestInAssetID` instead? Same below with the group keys.

rfqmsg/buy_request.go
```diff
@@ -286,28 +286,61 @@ func NewBuyRequest(peer route.Vertex, assetID *asset.ID,
...
+	msgData.InAssetID.WhenSome(
+		func(inAssetID tlv.RecordT[tlv.TlvType5, asset.ID]) {
+			assetID = &inAssetID.Val
+		},
+	)
+
+	var assetGroupKey *btcec.PublicKey
+	msgData.InAssetGroupKey.WhenSome(
+		// nolint: lll
+		func(inAssetGroupKey tlv.RecordT[tlv.TlvType6, *btcec.PublicKey]) {
```
guggero, line 90: nit: feel free to use less descriptive parameter names for such closures if it helps to avoid running into the length limit. For example just `key` instead of `inAssetGroupKey`, since from the outer context it's clear what the variable means.
Same further below (`t` instead of `suggestedRateTick`).

guggero (commented):
Did a first pass and it looks great! Will probably want to run it through some manual and integration tests with the whole custom channel setup to make sure nothing breaks, so we might want to get in the other RFQ PR first.

rfqmsg/request.go
ffranr: I'll add a TODO to explain that this field should be populated on the `BuyRequest` and not just set here.

guggero (approved):
tACK, validated against the custom channel integration test, things still work as before.

GeorgeTsagk (approved):
ACK :heavy_check_mark:

Issue #839, opened 2024-03-18
[bug]: restart tapd will reset  FedGlobalSyncConfig when with the --universe.public-access flag.


lukegao209:
Start tapd with the --universe.public-access flag.
Modify the settings accordingly.
Restart tapd.
The FedGlobalSyncConfig will be reset to its initial values. And then tapd will cost a long time and cpu to sync assets from lightning labs's universe

To prevent this, I suggest implementing an initialization check in server.go. 

```
	if s.cfg.UniversePublicAccess {
		err := s.cfg.UniverseFederation.SetAllowPublicAccess()
		if err != nil {
			return fmt.Errorf("unable to set public access "+
				"for universe federation: %w", err)
		}
	}
```

dstadulis:
Thanks for the report

>Restart tapd
When the tapd daemon is started for a second time (the restart), is the `--universe.public-access` flag being set?

lukegao209:
yeah, --universe.public-access flag is being set.

Roasbeef:
Ran into this myself, pretty annoying. 

Here's a suggestion on how to resolve this:
   * Separate the config for allowed actions to your own personal universe server vs how to _sync_ with a remote universe server. 
   * The existing args stay, these govern your own personal universe server. When you do `--universe.public-access`, this means that you let people both read from your server, and also insert new proofs into your server. 
       * By default with this, you _won't_ actively sync all records from a remote/default universe server.
    * We add a new set of flags to govern how _sync_ works. The default is sync nothing. You can add `--universe.sync-all-assets` to indicate you want to attempt to sync any universe root you can find. Today we have `import`, and `export`. This adds a third option: `sync`. 
    * In the auto syncer, when we go to sync, we'll look at this flag to see if it should be done at all. If so, then we look at the global or local flags to determine _what_ to sync.

PR #927, opened 2024-06-05
Enhance universe server/syncing CLI config control

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/839

This PR modifies the `--universe.public-access` argument to restrict its impact to determining public access permissions for the universe server, eliminating its previous influence on universe federation syncing behavior. The `--universe.public-access` argument can now set read, write, or read-write public access permissions. Using `--universe.public-access` as a flag will result in a configuration error, ensuring that the old flag behavior is clearly rejected (see: `ParseUniversePublicAccessStatus`).

Additionally, this PR introduces a new command line argument: `--universe.sync-all-assets`. When set, this flag will ensure that the federation syncer defaults to syncing all assets (global config set to sync all assets).


config.go
```diff
@@ -76,6 +77,72 @@ type DatabaseConfig struct {
...
+
+	// If the CLI argument is nil, we assume that no public access is
+	// granted.
+	if cliArg == nil {
+		return UniversePublicAccessStatusNone, nil
+	}
+
+	s := *cliArg
+
+	switch {
```
guggero, line 128: nit: could do `switch s {`, then `case "rw", "wr":` and so on.

tapcfg/config.go
```diff
@@ -257,12 +257,14 @@ type LndConfig struct {
...
 // values.
+//
+// nolint: lll
 type UniverseConfig struct {
 	SyncInterval time.Duration `long:"syncinterval" description:"Amount of time to wait between universe syncs"`
 
 	FederationServers []string `long:"federationserver" description:"The host:port of a Universe server peer with. These servers will be added as the default set of federation servers. Can be specified multiple times."`
 
-	PublicAccess bool `long:"public-access" description:"If true, and the Universe server is on a public interface, valid proof from remote parties will be accepted, and proofs will be queryable by remote parties. This applies to federation syncing as well as RPC insert and query."`
+	PublicAccess *string `long:"public-access" description:"The public access mode for the universe server. This can be unset, 'r', 'w', or 'rw'. If unset, public access is not enabled for the universe server. If 'r' is included, public access is allowed for read-only endpoints. If 'w' is included, public access is allowed for write endpoints."`
```
guggero, line 269: Unfortunately using a pointer here doesn't have the desired effect of not requiring a value at all. If I just use the flag as it was set before (simulating an existing node) like `--universe.public-access`, I get the error message `expected argument for flag `--universe.public-access', but got option `--allow-public-uni-proof-courier'
failed to load config: expected argument for flag `--universe.public-access', but got option `--allow-public-uni-proof-courier'`

So if I explicitly wanted to pass an empty value, I'd need to set `--universe.public-access=""`, which will cause the argument not to be `nil` but an empty string, but then gives me the error `error creating server: unable to generate server config: unable to parse universe public access status: unknown universe public access status:`.

So I think we can just make this a string and just check for empty string. Too bad, would've been a nice way to remain backward compatible.

config.go
```diff
@@ -76,6 +77,72 @@ type DatabaseConfig struct {
 	FederationDB *tapdb.UniverseFederationDB
 }
 
+// UniversePublicAccessStatus is a type that indicates the status of public
+// access to the universe server.
+type UniversePublicAccessStatus string
```
guggero, line 82: Since this is all config related, should this go into the `tapcfg` package? Non-blocking, just kind of weird here as the file is otherwise empty of any logic.

tapcfg/config.go
guggero: Also, should we keep some of the explanation of what this flag actually influences from the original message?
Perhaps something along the lines of:
```go
"The public access mode for the universe server, controlling whether remote parties can read from and/or write to this universe server over RPC if exposed to a public network interface. This can be unset, 'r', 'w', or 'rw'. If unset, public access is not enabled for the universe server. If 'r' is included, public access is allowed for read-only endpoints. If 'w' is included, public access is allowed for write endpoints.
```

config.go
```diff
@@ -138,11 +205,12 @@ type Config struct {
...
-	// is on a public interface, valid proof from remote parties will be
-	// accepted, and proofs will be queryable by remote parties.
-	// This applies to federation syncing as well as RPC insert and query.
-	UniversePublicAccess bool
+	// UniversePublicAccess is a field that indicates the status of public
+	// access (i.e. read/write) to the universe server.
+	//
+	// NOTE: This field does not influence universe federation syncing
+	// behaviour.
+	UniversePublicAccess UniversePublicAccessStatus
```
guggero, line 217: Nice to untangle this and only have a single meaning behind the flag :100:

perms/perms.go
```diff
@@ -308,6 +309,16 @@ func MacaroonWhitelist(allowPublicUniProofCourier bool,
 		whitelist[k] = v
 	}
 
+	// Conditionally whitelist universe server read methods.
+	if allowUniPublicAccessRead {
```
guggero, line 313: nit: to make things a bit more explicit, we could do `if allowUniPublicAccessRead || allowPublicUniProofCourier {` since the proof courier flag overwrites/supersedes the other two.

universe/auto_syncer.go
```diff
@@ -764,8 +764,8 @@ func (f *FederationEnvoy) SyncServers(serverAddrs []ServerAddr) error {
 	return nil
 }
 
-// SetAllowPublicAccess sets the global sync config to allow public access
-// for proof insert and export across all universes.
+// SetAllowPublicAccess sets the global (default) sync config to allow public
+// access for proof insert and export across all universes.
```
guggero, line 768: Sorry, it's been a while since I last looked at this, so I might misunderstand something. But isn't this config only for what _our_ syncer does? And doesn't influence what others can do to our universe? If that's true, then we might want to rename this method as its name first threw me off due to its similarity to the old boolean flag we now replaced.

guggero (commented):
Very nice! Mostly nits and suggestions, the overall change is great and necessary, even if it will require some users to change their config ever so slightly.

Ah, just a thought: Can you perhaps include a small section in the PR text body that we should put into the release notes for the next release? E.g. what needs to be changed and what the suggested value (I assume `rw` for everyone who previously had the boolean flag set) to set is?

config.go
ffranr: I need to use that type in `taproot-assets/config.go`. I'll get a circular dep if it's in `tapcfg`.

itest/tapd_harness.go
```diff
@@ -200,10 +200,11 @@ func newTapdHarness(t *testing.T, ht *harnessTest, cfg tapdConfig,
...
 	}
 
-	// Ensure valid proof from tapd nodes will be accepted, and proofs will
-	// be queryable by other tapd nodes. This applies to federation syncing
-	// as well as RPC insert and query.
-	tapCfg.Universe.PublicAccess = true
+	// Configure the universe server to ensure that valid proofs from tapd
+	// nodes will be accepted, and proofs will be queryable by other tapd
+	// nodes.
+	tapCfg.Universe.PublicAccess =
```
guggero, line 206: micro-nit: treat the cast as a function call and wrap/format the lines accordingly.

guggero (approved):
Beautiful, LGTM :tada:

PR #928, opened 2024-06-06
rfq+tapchannel: track locally accepted quotes in manager

guggero:
To allow direct peer (zero-hop) payments, we also need to be able to look up quotes we've accepted that our peer requested. The example flow here is: Dave creates an invoice (either specifying a BTC amount or asset amount, in any case the invoice is always denominated in milli-satoshi), Charlie has a direct channel with Dave and wants to pay the invoice with assets. Charlie only sees the milli-satoshi amount of the invoice and needs to find out how many asset units to send. For that, Charlie asks Dave for a quote over the given amount of milli-satoshi, gets an asset unit amount back. Charlie then creates an HTLC referencing that asset sell quote RFQ ID. But the invoice interceptor on Dave's side would previously not recognize that RFQ ID, because it's not an ID for a quote Dave requested but one Dave accepted.

### DIFF
--- a/rfq/manager.go
+++ b/rfq/manager.go
@@ -124,8 +124,23 @@ type Manager struct {
 	// requested and that have been accepted by peer nodes. These quotes are
 	// exclusively used by our node for the sale of assets, as they
 	// represent agreed-upon terms for sale transactions with our peers.
-	peerAcceptedSellQuotes lnutils.SyncMap[SerialisedScid,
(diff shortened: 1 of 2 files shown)


GeorgeTsagk (approved):
Clean :sparkle: 

Could add some simple coverage post rc

ffranr (approved):

Issue #925, opened 2024-06-04
Use `meta_data` field to covey an asset's `characteristic`


dstadulis:
### Discussed in https://github.com/lightninglabs/taproot-assets/discussions/763

<div type='discussions-op-text'>

<sup>Originally posted by **dstadulis** January 10, 2024</sup>
### Background
Using integers in applications which require uncompromising accuracy (Money, accounting, settlement etc) is common in order to handle the onerous computer hardware and software demands to perform accurate operations on floating point numbers.  IEEE 754 is a standards example of the attempts to create uniformity in how to implement computation on floats.
Financial accounting software often uses integers to achieve accuracy without subjecting itself to FP arithmetic complexity.
https://en.wikipedia.org/wiki/IEEE_754
https://husobee.github.io/money/float/2016/09/23/never-use-floats-for-currency.html

tapd uses integers to ensure that it's precision and accuracy is high.  Bitcoin similarly uses integers for internal account and `COIN = 100000000` used to convert/present the decimal in the UI.

Such a variable merely instructs UIs on where to place the decimal point and referred to as a `characteristic`. [1]

### User / Market Needs
As a `application developer` I want an `programmatic, authenticated method to produce an asset's characteristic` in order to `make the presentation of tapd's internally-accounted integers be displayed in UIs without needing bespoke code`
As an `Asset Issuer` I want to `convey where the decimal place should be displayed to users` in order to `prevent confusion and calculation errors in handling amounts`

### Resources
[1] https://en.wikipedia.org/wiki/Characteristic_(exponent_notation)#Mantissa_and_characteristic
https://en.wikipedia.org/wiki/Significand</div>

Roasbeef:
Could also call it precision.

Roasbeef:
Current plan:
  * Add `precision` to the minting RPC. This is an integer, it's an integer in the JSON as well.
     * We'd also display this for any asset display call on RPC, including the Universe calls. 
     * Wallets are to use this to modify the way they display assets. Eg: if you have 1000 units on an asset, and the `precision` is set to 2, then it's displayed as `10.00` in UIs. 
     * This also has implications for the RFQ and channel layer as well. Ideally the smallest unit of the asset should be on par or less than the value of a satoshi, ideally a milli-satoshi. 
  * If this is set, then the meta type MUST be JSON, fail otherwise. 
  * If we have meta JSON, then we merge it with this new fields. 

[...]

PR #930, opened 2024-06-06
Add decimal display field for mints

GeorgeTsagk:
## Description

We add a new rpc parameter named `decimal_display` which expresses the granularity of the assets, expressed in number of decimal places.

> Example: If an asset named `testbuxx` is minted with a decimal display of `2` then we need 100 of those minted assets to compose a whole unit of `testbuxx`.

This field is only descriptive and does not change the flow of operation for the whole lifecycle of the asset. The decimal display is encoded in the asset metadata **iff** the metadata uses a json encoding.

Closes https://github.com/lightninglabs/taproot-assets/issues/925


itest/assets_test.go
```diff
@@ -473,3 +473,31 @@ func testMintAssetsWithTapscriptSibling(t *harnessTest) {
...
+	assetMint := mintrpc.MintAssetRequest{
+		Asset: &mintrpc.MintAsset{
+			AssetType: taprpc.AssetType_NORMAL,
+			Name:      mintName,
+			AssetMeta: &taprpc.AssetMeta{
+				Data: jsonData,
+				Type: taprpc.AssetMetaType_META_TYPE_JSON,
+			},
+			Amount: 500,
+			// Precision: 2,
```
GeorgeTsagk, line 492: this is temp as currently the itest seems to be reporting mismatching metahashes between the generated metadata hash and what is present in the genesis info

GeorgeTsagk: there was a hard-coded `TypeOpaque` metadata type in an itest helper method, this is fixed now

rpcserver.go
```diff
@@ -6570,3 +6600,26 @@ func encodeVirtualPackets(packets []*tappsbt.VPacket) ([][]byte, error) {
...
+func EncodePrecisionInJSON(precision uint64, jBytes []byte) ([]byte, error) {
+	var jMeta map[string]interface{}
+
+	err := json.Unmarshal(jBytes, &jMeta)
+	if err != nil {
+		return nil, fmt.Errorf("unable to unmarshal json asset meta: "+
+			"%w", err)
+	}
+
+	jMeta[MetadataPrecisionKey] = precision
```
GeorgeTsagk, line 6616: not sure if I should check for conflict here
it's not impossible that a user may have already defined this key (maybe intentionally, maybe by mistake :shrug: )

so perhaps on conflict maintain the user provided value? they *probably* know what they're doing
...or just error out

guggero: I think returning an error instead of doing something implicit is probably better. Then the user has the option to just use the JSON or remove the value from the JSON and use the flag.

taprpc/mintrpc/mint.proto
```diff
@@ -187,6 +187,14 @@ message MintAsset {
     a BIP-86 key will be derived from the underlying wallet.
     */
     taprpc.ScriptKey script_key = 12;
+
+    /*
+    Precision is a single integer value that represents how fine grained an
+    asset unit is. The value of this field is the number of decimal places that
+    this asset can be divided into. For example, if the precision is 2, a unit
```
Roasbeef, line 194: Thought we were going with an approach where the supply amount stays, and this is mainly for UI purposes? So a re-write here should be: 
> If the precision is 2, an a user has 1000 units of an asset, then this would be displayed as 10.00

rpcserver.go
```diff
@@ -109,6 +110,10 @@ const (
 	// proofTypeReceive is an alias for the proof type used for receiving
 	// assets.
 	proofTypeReceive = tapdevrpc.ProofTransferType_PROOF_TRANSFER_TYPE_RECEIVE
+
+	// MetadataPrecisionKey is the JSON key used in the metadata field of a
+	// minted asset to express the precision of the minted asset.
+	MetadataPrecisionKey = "tap_asset_precision"
```
Roasbeef, line 116: Could drop the prefix? Given it's JSON, which is already effectively namespaced (lives in the meta field), and we can save some here re space by doing so.

rpcserver.go
```diff
@@ -6570,3 +6600,31 @@ func encodeVirtualPackets(packets []*tappsbt.VPacket) ([][]byte, error) {
...
 	return rawPackets, nil
 }
+
+// EncodePrecisionInJSON takes a precision value and an encoded json object and
+// attempts to append a specific key-value pair in the json object, then encodes
+// it again and returns it.
+func EncodePrecisionInJSON(precision uint64, jBytes []byte) ([]byte, error) {
+	var jMeta map[string]interface{}
+
+	err := json.Unmarshal(jBytes, &jMeta)
```
Roasbeef, line 6610: Will this fail for empty `jBytes`? Thinking of the case where the user sets nothing, and then specifies the precision field.

rpcserver.go
```diff
@@ -6570,3 +6600,31 @@ func encodeVirtualPackets(packets []*tappsbt.VPacket) ([][]byte, error) {
 
 	return rawPackets, nil
 }
+
+// EncodePrecisionInJSON takes a precision value and an encoded json object and
+// attempts to append a specific key-value pair in the json object, then encodes
+// it again and returns it.
+func EncodePrecisionInJSON(precision uint64, jBytes []byte) ([]byte, error) {
```
Roasbeef, line 6607: Could alternatively have this live in the `assets` package, or `taprpc`.

itest/asset_meta_test.go
```diff
@@ -101,3 +101,31 @@ func testAssetMeta(t *harnessTest) {
...
+			},
+			Amount:    500,
+			Precision: 2,
+		},
+	}
+	rpcSimpleAssets := MintAssetsConfirmBatch(
+		t.t, t.lndHarness.Miner.Client, t.tapd,
+		[]*mintrpc.MintAssetRequest{&assetMint},
+	)
+
```
Roasbeef, line 127: Should assert that the precision field has indeed been encoded.

Roasbeef (changes requested):

taprpc/mintrpc/mint.proto
GeorgeTsagk: How would we ever deal with sub-units then? it would always be fixed to `.00`

rpcserver.go
GeorgeTsagk: ah yeah, def can drop
just used this to minimize conflict with some `precision` the user may have defined

but sgtm can just do `precision` here

rpcserver.go
GeorgeTsagk: nice catch, probably yeah, will test and adjust it accordingly
(we **should** be able to accept empty json)

itest/asset_meta_test.go
GeorgeTsagk: this is done indirectly by checking the MetaHash of the asset genesis with the generated MetaHash of the metadata in the request. This is nested inside the helpers, could surface here too

taprpc/mintrpc/mint.proto
Roasbeef: There are no sub units. It's always a whole unit. Integer arhietmetic everywhere. A whole unit may be part of a denomination however.

Roasbeef: It also wouldn't be fixed to `.00`. If I have 999 unit, then that's `9.99`.

rpcserver.go
GeorgeTsagk: moving to taprpc

itest/asset_meta_test.go
GeorgeTsagk: Added an explicit check too

cmd/tapcli/assets.go
```diff
@@ -43,6 +43,7 @@ var (
 	assetMetaBytesName           = "meta_bytes"
 	assetMetaFilePathName        = "meta_file_path"
 	assetMetaTypeName            = "meta_type"
+	assetDecimalPlacesName       = "decimal_places"
```
dstadulis, line 46:
```suggestion
	assetDecimalDisplayName       = "decimal_display"
```

taprpc/mintrpc/mint.proto
```diff
@@ -187,6 +187,14 @@ message MintAsset {
     a BIP-86 key will be derived from the underlying wallet.
     */
     taprpc.ScriptKey script_key = 12;
+
+    /*
+    Decimal display is a single integer value that represents how fine grained
+    an asset unit is. The value of this field is the number of decimal places
+    that this asset can be divided into. For example, if the decimal_display
+    value is 2, a unit is considered to be 100 of these minted assets.
```
guggero, line 195: Add something like "This is intended as information to wallets that display balances only and have no impact on the behavior or other parts of the protocol. The value will be added as a field to the meta data JSON of the asset (and is therefore only compatible with the JSON meta data type)."

taprpc/mintrpc/mint.proto
```diff
@@ -187,6 +187,14 @@ message MintAsset {
...
+    */
+    uint64 decimal_display = 13;
```
guggero, line 197: make this uint32?

rpcserver.go
```diff
@@ -517,6 +517,32 @@ func (r *rpcServer) MintAsset(ctx context.Context,
...
+	if decDisplay != 0 && metaType != taprpc.AssetMetaType_META_TYPE_JSON {
+		return nil, fmt.Errorf("decimal display requires JSON meta " +
+			"type")
+	}
+
+	// If a JSON meta field is specified, then we'll update the decimal
+	// display in the meta field. Even if the decimal display is not
+	// specified in the rpc request, we'll still encode a 0 value.
+	if req.Asset.AssetMeta.Type == taprpc.AssetMetaType_META_TYPE_JSON {
+		updatedMeta, err := taprpc.EncodeDecimalDisplayInJSON(
```
guggero, line 536: We should also cap the value of `decDisplay`. Given `uint64` is 19 decimal places, does it make sense to allow more than 12 to be after the decimal point?

cmd/tapcli/assets.go
```diff
@@ -77,6 +78,11 @@ var mintAssetCommand = cli.Command{
 			Name:  assetSupplyName,
 			Usage: "the target supply of the minted asset",
 		},
+		cli.Uint64Flag{
+			Name: assetDecimalPlacesName,
+			Usage: "the number of decimal places that this asset " +
+				"mint uses",
```
dstadulis, line 84:
```suggestion
			Name: assetDecimalDisplayName,
			Usage: "the number of decimal places to shift the " +
				"internal asset representation to the left", 
```

taprpc/taprpc_utils.go
```diff
@@ -1,6 +1,17 @@
...
+	"encoding/json"
+	"fmt"
+
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+const (
+	// MetadataDecDisplayKey is the JSON key used in the metadata field of a
+	// minted asset to express the decimal display of the minted asset.
+	MetadataDecDisplayKey = "tap_asset_decimal_display"
```
guggero, line 13: I don't think we need the `tap_asset_` prefix in the key.

cmd/tapcli/assets.go
```diff
@@ -296,6 +302,7 @@ func mintAsset(ctx *cli.Context) error {
 			AssetType:       assetType,
 			Name:            ctx.String(assetTagName),
 			AssetMeta:       assetMeta,
+			DecimalDisplay:  ctx.Uint64(assetDecimalPlacesName),
```
dstadulis, line 305:
```suggestion
			DecimalDisplay:  ctx.Uint64(assetDecimalDisplayName),
```

cmd/tapcli/assets.go
guggero, line 46: Update flag name to "decimal_display" to match the JSON key?

itest/test_list_on_test.go
```diff
@@ -25,6 +25,10 @@ var testCases = []*testCase{
 		name: "mint fund seal assets",
 		test: testMintFundSealAssets,
 	},
+	{
+		name: "mint asset decimal places",
+		test: testMintAssetWithDecimalPlaces,
```
dstadulis, line 30:
```suggestion
		test: testMintAssetWithDecimalDisplayMetaField,
```

cmd/tapcli/assets.go
```diff
@@ -77,6 +78,11 @@ var mintAssetCommand = cli.Command{
 			Name:  assetSupplyName,
 			Usage: "the target supply of the minted asset",
 		},
+		cli.Uint64Flag{
+			Name: assetDecimalPlacesName,
+			Usage: "the number of decimal places that this asset " +
```
guggero, line 83: I think this requires more info as well, can re-use most of what I suggested for the gRPC field description.

guggero (commented):
Looks pretty good to me, mostly UX related comments.

taprpc/mintrpc/mint.proto
dstadulis, line 195:
```suggestion
    DecimalDisplay: Converts Taproot Asset's internal, integer representation of
    assets amount to fractional quantities.

    DecimalDisplay dictates the number of decimal places to shift the amount to the
    left converting from Taproot Asset's internal integer representation to a
    UX-recognizable fractional quantity.

    E.g, if Amount is 1000 (representing 10 euros) and DecimalDisplay is 2, the
    displayed amount would be 10.00
```

taprpc/mintrpc/mint.swagger.json
```diff
@@ -427,6 +427,11 @@
         "script_key": {
           "$ref": "#/definitions/taprpcScriptKey",
           "description": "The optional script key to use for the new asset. If no script key is given,\na BIP-86 key will be derived from the underlying wallet."
+        },
+        "decimal_display": {
+          "type": "string",
+          "format": "uint64",
+          "description": "Decimal display is a single integer value that represents how fine grained\nan asset unit is. The value of this field is the number of decimal places\nthat this asset can be divided into. For example, if the decimal_display\nvalue is 2, a unit is considered to be 100 of these minted assets."
```
dstadulis, line 434:
```suggestion
          "description": "DecimalDisplay: Converts Taproot Asset's internal, integer representation of\nassets amount to fractional quantities.\n\nDecimalDisplay dictates the number of decimal places to shift the amount to the\nleft converting from Taproot Asset's internal integer representation to a\nUX-recognizable fractional quantity.\n\nE.g, if Amount is 1000 (representing 10 euros) and DecimalDisplay is 2, the\ndisplayed amount would be 10.00"
```

dstadulis (changes requested):
Added improved description and changes some names

rpcserver.go
GeorgeTsagk: Agree

GeorgeTsagk: Will add the above rationale as a comment next to the const

taprpc/taprpc_utils.go
GeorgeTsagk: This was only done to minimize risk of a random conflict with a user's keys

Since we now error out on conflict I'll simplify this key.

taprpc/mintrpc/mint.proto
```diff
@@ -187,6 +187,20 @@ message MintAsset {
     a BIP-86 key will be derived from the underlying wallet.
     */
     taprpc.ScriptKey script_key = 12;
+
+    /*
+    DecimalDisplay dictates the number of decimal places to shift the amount to
```
guggero, line 192: nit: either use just "Decimal display" or the actual field name "decimal_display" at beginning of sentence.

taprpc/mintrpc/mint.proto
```diff
@@ -187,6 +187,20 @@ message MintAsset {
...
+    the left converting from Taproot Asset integer representation to a
+    UX-recognizable fractional quantity.
+
+    For example, if the decimal_display value is 2 and there's 100 of those
+    assets a wallet would display the amount as "1.00". This field is intended
```
guggero, line 197: s/of those assets a wallet/of those assets and a wallet/

taprpc/mintrpc/mint.proto
guggero: nit: unaddressed.

guggero (approved):

taprpc/mintrpc/mint.proto
GeorgeTsagk: hm, not really a syntax that matches what I want to say here, maybe:
```
if the decimal_display value is 2 and there's 100 of those
assets, then a wallet would display the amount as "1.00"
```

dstadulis (changes requested):
Little wording tweak

PR #931, opened 2024-06-06
RFQ enrich incoming accept messages with sent requests

ffranr:
In this PR, we enrich incoming accept messages with corresponding sent requests. This involves storing outbound requests before sending and matching incoming accept message responses based on the RFQ quote `ID`. This matching process is performed in the RFQ stream service, ensuring that other RFQ services only encounter accept messages with fully populated fields, including the new request field.

As a result, transporting the asset ID in the accept message is no longer necessary. The asset ID can be found in the request message.

Removing the asset ID field from the accept wire message is important because it is redundant and isn't clear whether it is `out_asset` or `in_asset`. It also does not have a asset group key counterpart.


guggero (approved):
Very nice, LGTM :tada:

GeorgeTsagk (approved):
LGTM

Issue #886, opened 2024-04-17
[feature]: Exercise asset locktimes


jharveyb:
As part of asset channel support, we'd like to start using the asset-level locktimes (even though all the anchor transactions will be using locktimes as they are now for LN).

Specifically, we'd like to:

- Extract the locktimes of the anchor TX for asset inputs
- Have the VM check these locktimes when transferring an asset
- Plumb in locktime support for the virtual transactions we use to represent transfers
- Add assertions around all output locktimes being shorter than the transfer anchor TX locktimes (related to re-org safety)
- Ensure that these locktimes are also checked during proof verification

Some related pre-existing code that would be useful for this:

https://pkg.go.dev/github.com/btcsuite/btcd/blockchain#IsFinalizedTransaction

https://pkg.go.dev/github.com/btcsuite/btcd/blockchain#SequenceLockActive

I think we could work 'inside-out' to implement this, so:

- Wire up locktime support when constructing virtual TXs (& then validating)
- Add args to the VM to pass in the current block height, etc.
- Wire up locktime propagation when constructing virtual TXs for transfer
- Add locktime checking to coin selection at the beginning of the transfer logic
- Add args for the transfer flow (PSBT only?) to add locktimes to outputs
- Update proof verifier to propagate anchor locktimes and info all the way to the VM

gijswijs:
Considering the fact that we are talking about assets in a lightning channel, what happens if we don't support locktimes for in-channel asset transfers? You still have the locktimes of the anchor tx, so there is no chance of spending an asset before said locktime. The re-org vulnerability doesn't apply by definition, because the anchor Tx locktime is longer than the asset locktime (which has no locktime)

jharveyb:
True - actually, I'm not sure which situations an asset locktime would have meaning that wouldn't also be required in the anchor TX.

gijswijs:
If we start at the VM level, the following needs to occur:

- [ ] Add argument to the VM to pass in the block height of the block that includes the transaction.

We don't map the `relative_lock_time` or `lock_time` during the mapping of the inputs. Instead we'll map this during the validation process.

During validation of a state transition:
- For each `prev_input` within the set of referenced `prev_asset_witnesses`:
  - [ ] Set the sequence number to the `relative_lock_time` field of the input, if it exists.
  - [ ] Set the lock time of the transaction as the `lock_time` of the input TLV leaf being validated, if it exists.
  - Perform external lock time and relative lock time validation:
[...]

PR #933, opened 2024-06-07
Add lock time (CSV/CLTV) support to asset scripts, finish adding versions everywhere

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/886.

This PR allows asset-level Tapscript leaves to use `OP_CHECKSEQUENCE[VERIFY]` and `OP_CHECKLOCKTIME[VERIFY]` op codes and enforces them correctly in the proof verification process.
The fields [`LockTime`](https://github.com/lightninglabs/taproot-assets/blob/93da9d4f4353f7f50dc6511cc1201d9074d101d6/asset/asset.go#L1325) and [`RelativeLockTime`](https://github.com/lightninglabs/taproot-assets/blob/93da9d4f4353f7f50dc6511cc1201d9074d101d6/asset/asset.go#L1331) need to be set on the root output asset of a spending virtual transaction (`vOutput`, the same root asset that receives the transfer TX witness).
Asset-level lock times are also bubbled up to the BTC-level anchor transaction to prevent issues when mixing time locks on the asset and BTC level.

To allow for a smooth transition to lock time enabled assets, and to simplify protocol feature updates in the future, we also add a new commitment version to the Taproot Asset level proofs  (similar to how we added version support to [assets](https://github.com/lightninglabs/taproot-assets/pull/381) and [addresses](https://github.com/lightninglabs/taproot-assets/pull/501)).


vm/error.go
```diff
@@ -106,6 +110,8 @@ func (k ErrorKind) String() string {
 		return "invalid split commitment proof"
 	case ErrInvalidRootAsset:
 		return "invalid zero-value root asset"
+	case ErrUnFinalizedAsset:
+		return "un-finalized asset"
```
gijswijs, line 114: ultra-nit: unfinalized is a correct word and doesn't need a hyphen.

vm/vm.go
```diff
@@ -28,17 +31,74 @@ type Engine struct {
 	// prevAssets maps newAsset's inputs by the hash of their PrevID to
 	// their asset.
 	prevAssets commitment.InputSet
+
+	// validateTimeLocks is a flag that indicates whether the engine should
+	// validate lock times.
+	validateTimeLocks bool
```
gijswijs, line 37: Are we sure we want this flag to exist? Is there still a reason to ever validate without looking at timelocks?

vm/vm.go
```diff
@@ -437,3 +532,66 @@ func (vm *Engine) Execute() error {
...
+		// Otherwise, we can just compare the lock time to the current
+		// block height.
+		case blockHeight < uint32(newAsset.LockTime):
+			return newErrKind(ErrUnFinalizedAsset)
+		}
+	}
+
+	// Now check any relative lock time. For this we need to look up the
+	// height of the block the input's anchor transaction was confirmed in.
+	if newAsset.RelativeLockTime != 0 {
```
gijswijs, line 564: We assume `RelativeLockTime` is always expressed in blocks? The sequence field on btc anchorTx has a type flag for indicating blocks or seconds. (by the way, PR 3/3 also assumes blocks for `RelativeLockTime` but it also assumes blocks for `LockTime`!

tx_validator.go
```diff
@@ -47,3 +70,156 @@ func (v *WitnessValidatorV0) ValidateWitnesses(newAsset *asset.Asset,
...
+	assetStore *tapdb.AssetStore, proofFile *proof.File) *ProofChainLookup {
+
+	return &ProofChainLookup{
+		chainBridge: chainBridge,
+		assetStore:  assetStore,
+		proofFile:   proofFile,
+	}
+}
+
+// CurrentHeight return the current height of the main chain.
```
gijswijs, line 96:
```suggestion
// CurrentHeight returns the current height of the main chain.
```

tx_validator.go
```diff
@@ -47,3 +70,156 @@ func (v *WitnessValidatorV0) ValidateWitnesses(newAsset *asset.Asset,
...
+	}
+
+	// If we arrive here, we couldn't find the transaction in the proof
+	// file.
+	return 0, errTxNotFound
+}
+
+// MeanBlockTimestamp returns the timestamp of the block at the given height as
+// a Unix timestamp in seconds, taking into account the mean time elapsed over
+// the previous 10 blocks.
```
gijswijs, line 172:
```suggestion
// the previous 11 blocks.
```

tapgarden/mock.go
```diff
@@ -462,6 +466,40 @@ func (m *MockChainBridge) EstimateFee(ctx context.Context,
...
+// included in.
+func (m *MockChainBridge) TxBlockHeight(context.Context,
+	chainhash.Hash) (uint32, error) {
+
+	return 123, nil
+}
+
+// MeanBlockTimestamp returns the timestamp of the block at the given height as
+// a Unix timestamp in seconds, taking into account the mean time elapsed over
+// the previous 10 blocks.
```
gijswijs, line 479:
```suggestion
// the previous 11 blocks.
```

gijswijs (changes requested):
Looks like it's nearly there. Made some small comments.

Also, I think you need to still update the test vectors.

vm/error.go
guggero: Hmm, both my IDE and translation website of choice (dict.leo.cc) didn't want to know that word. But you're right, reverted.

vm/vm.go
guggero: Yeah, I think it's nice for unit tests. We would need to hook up way more to get all existing tests passed without the flag.

vm/vm.go
guggero: Hmm, good point. Not sure if we would want to also use a specific bit to indicate a time based sequence. Or if we also want to use the trick that everything above 500 million is a Unix timestamp... Added a TODO for now.

Roasbeef (changes requested):
Checkpoint at `a47890fbb438418d82941b01d26135d992902cc7`

vm/vm.go
```diff
@@ -437,3 +526,71 @@ func (vm *Engine) Execute() error {
 
 	return vm.validateStateTransition()
 }
+
+// checkLockTime checks the absolute and relative lock time of the previous
+// asset. `blockTimestamp` is ignored for now.
+func checkLockTime(ctx context.Context, prevAsset *asset.Asset,
```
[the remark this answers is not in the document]
guggero: Yes, I agree, the lock times (especially the relative one, as it applies to each input being spent) should've been defined on the `PrevWitness`. Though given the time we have available before we want to push this out, I think we can't just move them right now (given the implications on TLV encoding/versioning/proofs and so on).

tapdb/assets_store.go
```diff
@@ -344,6 +345,17 @@ type AssetGroupBalance struct {
 	Balance  uint64
 }
 
+// cacheableTimestamp is a wrapper around an int32 that can be used as a
+// value in an LRU cache.
+type cacheableBlockHeight uint32
```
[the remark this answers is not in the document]
guggero: No, the previous commit used a `cacheableTimestamp`, here we're introducing `cacheableBlockHeight`. So I think things should compile correctly (I even checked that manually at some point). But yes, we don't have that CI step active here (we probably should add that).

tx_validator.go
```diff
@@ -47,3 +69,156 @@ func (v *WitnessValidatorV0) ValidateWitnesses(newAsset *asset.Asset,
...
+func (l *ProofChainLookup) MeanBlockTimestamp(ctx context.Context,
+	blockHeight uint32) (time.Time, error) {
```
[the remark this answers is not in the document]
guggero: Yes, I lifted it. Because otherwise I would've needed to implement the `HeaderCtx` interface which would probably have resulted in way more lines of codes (and awkward either empty methods because they're not used by that function or "proper" implementations that don't serve a real purpose).
So just lifting the logic instead seemed way more straight-forward.

vm/vm.go
guggero: Okay, I flipped the semantics here. Turned out to not be a big change after all.

vm/vm.go
guggero: Okay, added the same rules as in `btcd`. Re-using those functions wasn't really possible without restructuring everything, so I lifted some more logic instead.

vm/vm.go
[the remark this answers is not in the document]
guggero: I reordered the commits so it's more clear for any reviewer that we're changing the semantics. Sorry about that.

vm/vm.go
```diff
@@ -437,3 +526,71 @@ func (vm *Engine) Execute() error {
...
+		// block height.
+		case blockHeight < uint32(prevAsset.LockTime):
+			return newErrKind(ErrUnfinalizedAsset)
+		}
+	}
+
+	// Now check any relative lock time. For this we need to look up the
+	// height of the block the input's anchor transaction was confirmed in.
+	if prevAsset.RelativeLockTime != 0 {
+		// TODO(guggero): Also allow time based relative lock times?
```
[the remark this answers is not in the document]
guggero: Added/copied the logic from `btcd`.

itest/psbt_test.go
```diff
@@ -2139,6 +2139,200 @@ func testPsbtExternalCommit(t *harnessTest) {
...
+	require.NoError(t.t, err)
+
+	// Now we'll attempt to complete the transfer.
+	sendResp2, err := bob.AnchorVirtualPsbts(
+		ctxt, &wrpc.AnchorVirtualPsbtsRequest{
+			VirtualPsbts: [][]byte{b.Bytes()},
+		},
+	)
+	require.NoError(t.t, err)
+	// require.ErrorContains(t.t, err, "non BIP68 final")
```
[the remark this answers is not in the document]
guggero: Yeah, looks like the itest wasn't fully finished yet. Fixed that, we now have negative and positive test for the anchor transaction lock time bubbling up.

vm/vm.go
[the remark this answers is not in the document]
guggero: Aight, so I ended up implementing it by looking at those functions you provided.
Unfortunately, as mentioned somewhere above, re-using the code directly wasn't possible without a lot of refactor.
So I'm not really sure there's no off-by-one in there or something. Would be great if you could take a look, as I'm not super familiar with the intricacies of the different BIPs.
If we see the risk of it being wrong as too great, we can still error out and declare time based time locks as non-supported yet.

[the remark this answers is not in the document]
gijswijs: At some point I hooked up `IsFInalizedTransaction`, which didn't require much of a refactor, but I ran into the off-by-one issue, that we discussed in the `btcd` chat. The scope of that function seems to be just ever so slightly not directly mappable onto this situation. Off course we could do the same as `btcd` and add one to the height (https://github.com/btcsuite/btcd/blob/b1b94202082b8e22160df31349992f37fb360f75/blockchain/validate.go#L803C3-L803C37) but that didn't make for very self-explanatory code.

directly using `SequenceLockActive` required a lot of refactor, that we chose not to do at this moment.

gijswijs (approved):

guggero:
Added a negative integration test for the asset-level lock time and found two small bugs in the process. Pushed two commits to address them.

vm/vm.go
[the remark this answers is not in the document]
Roasbeef: Gotcha, no need to use 100%, also given that the functions we have very closely mirror them.

tapsend/send.go
```diff
@@ -609,13 +611,19 @@ func PrepareOutputAssets(ctx context.Context, vPkt *tappsbt.VPacket) error {
 		// it's a zero value (tombstone) split output for the sender.
 		if vOut.Type.IsSplitRoot() {
 			vOut.Asset = splitCommitment.RootAsset.Copy()
+			vOut.Asset.LockTime = vOut.LockTime
+			vOut.Asset.RelativeLockTime = vOut.RelativeLockTime
+			vOut.Asset.ScriptKey = vOut.ScriptKey
```
Roasbeef, line 617: We weren't setting this before? 🤔

tapsend/send.go
```diff
@@ -1134,13 +1134,56 @@ func HasInput(tx *wire.MsgTx, outpoint wire.OutPoint) bool {
 	return false
 }
 
+// AssertAnchorTimeLocks makes sure the anchor transaction and its inputs have
+// the correct lock time and sequence set, according to the assets being spent.
+func AssertAnchorTimeLocks(btcPkt *psbt.Packet, vPkt *tappsbt.VPacket) {
```
Roasbeef, line 1155: IMO the func name doesn't quite match the body, would think it would return an error if the caller didn't do this assembly itself beforehand. 

Non-blocking naming suggestion: `UpdateAnchorTimeLocks`?

commitment/tap.go
```diff
@@ -54,9 +80,8 @@ type AssetCommitments map[[32]byte]*AssetCommitment
 // leaves represented as `asset_version || asset_tree_root || asset_sum`, are
 // keyed by their `asset_group_key` or `asset_id` otherwise.
 type TapCommitment struct {
-	// Version is the maximum Taproot Asset version found within all of the
-	// assets committed.
-	Version asset.Version
+	// Version is the version of the Taproot Asset commitment.
+	Version TapCommitmentVersion
```
Roasbeef, line 84: 👍

commitment/tap.go
```diff
@@ -120,17 +145,37 @@ func NewTapCommitment(newCommitments ...*AssetCommitment) (*TapCommitment,
...
-	}, nil
+	switch {
+	// If no version was specified, we'll use the max version of the asset
+	// commitments.
+	case version == nil:
+		return &TapCommitment{
+			TreeRoot:         root,
+			assetCommitments: assetCommitments,
+			tree:             tree,
+			Version:          TapCommitmentVersion(maxVersion),
```
Roasbeef, line 156: 👍 for bubbling this routine up, and also making it more explicit.

commitment/tap.go
```diff
@@ -288,16 +371,55 @@ func NewTapCommitmentWithRoot(version asset.Version,
...
+		}
 	}
+
 	leafScript := bytes.Join(leafParts, nil)
 	return txscript.NewBaseTapLeaf(leafScript)
 }
 
+// Downgrade returns a taproot asset commitment with the same root, but with
+// the asset commitment version downgraded to V0. This is useful for unit tests.
+func (c *TapCommitment) Downgrade() (*TapCommitment, error) {
```
Roasbeef, line 408: Non-blocking: could be a private method if just used for unit tests.

tappsbt/interface.go
```diff
@@ -173,6 +188,23 @@ func (p *VPacket) Copy() *VPacket {
 	}
 }
 
+// CommitmentVersion returns the Taproot Asset commitment version that matches
+// the VPacket version.
+func CommitmentVersion(vers VPacketVersion) (*commitment.TapCommitmentVersion,
```
Roasbeef, line 193: Doesn't seem to be used?

tapdb/sqlc/migrations/000019_managed_utxo_commitment_version.up.sql
```diff
@@ -0,0 +1,3 @@
+-- Add a field to store the version of the Taproot Asset commitment anchored in
+-- this UTXO. Existing UTXOs will have this set to NULL.
+ALTER TABLE managed_utxos ADD COLUMN root_version SMALLINT;
```
Roasbeef, line 3: 👍

tapdb/asset_minting.go
```diff
@@ -891,11 +893,66 @@ func fetchAssetSprouts(ctx context.Context, q PendingAssetStore,
...
+	tapCommitment, err := commitment.FromAssets(
+		fn.Ptr(commitment.TapCommitmentV2), assetSprouts...,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// If we don't have any sprouts yet, it doesn't make sense to verify the
+	// commitment.
+	if len(assetSprouts) == 0 {
```
Roasbeef, line 907: Can lift this above, don't need to make a commitment at all if no sprouts...?

tapfreighter/chain_porter.go
```diff
@@ -451,6 +451,18 @@ func (p *ChainPorter) storeProofs(sendPkg *sendPackage) error {
 		serializedScriptKey := asset.ToSerialized(out.ScriptKey.PubKey)
 		sendPkg.FinalProofs[serializedScriptKey] = outputProof
 
+		// Before we import the proof into the proof archive, we'll
+		// validate it.
+		verifier := &proof.BaseVerifier{}
+		_, err = verifier.Verify(
```
Roasbeef, line 457: 👍 

Should also help catch certain classes of bugs earlier. Like writing an invalid proof to disk, then it whens trying to post to a Universe.

Roasbeef (approved):
LGTM 🧤

commitment/tap.go
Roasbeef: Scratch that, used elsewhere.

commitment/tap.go
jharveyb: Ah the last case could be dropped here actually.

tappsbt/interface.go
jharveyb: Used in `tapsend/commitPacket`, `tapsend/ValidateAnchorOutputs`, `tapfreigher/wallet/FundPacket`. And the aux funding controller.

tapdb/asset_minting.go
jharveyb: Good point..I think `len(assetSprouts) == 0` should actually be an error here?

tapsend/send.go
jharveyb: I think they get set as part of `.Copy()`.

tapdb/asset_minting.go
jharveyb: Resolved this by updating the test to account for commitment root validation + moving this check earlier.

jharveyb:
Couldn't repro `unit-race` failure locally, may be a new flake or just timeout only on slower CI box.

PR #936, opened 2024-06-07
Custom channel bug fixes

guggero:
Fixes the following issues:
 - A panic during co-op close when one side didn't have any balance at all
 - Format custom channel data to JSON in `PendingChannels` RPC
 - Fix HTLC being rejected due to asset precision rounding issue
 - Allow forwarding an asset-to-asset HTLC (single hop payment with asset channels on both sides)

### DIFF
--- a/tapchannel/aux_closer.go
+++ b/tapchannel/aux_closer.go
@@ -258,7 +258,9 @@ func (a *AuxChanCloser) AuxCloseOutputs(
 		// anchor amt, then we'll just drop this allocation, and modify
 		// our asset allocation to match this value.
 		if amtAfterAnchor <= o.DustLimit {
-			localAlloc.BtcAmount = btcAmt
+			if localAlloc != nil {
+				localAlloc.BtcAmount = btcAmt
+			}
 			return
 		}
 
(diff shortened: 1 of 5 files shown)


tapchannel/aux_closer.go
```diff
@@ -258,7 +258,9 @@ func (a *AuxChanCloser) AuxCloseOutputs(
 		// anchor amt, then we'll just drop this allocation, and modify
 		// our asset allocation to match this value.
 		if amtAfterAnchor <= o.DustLimit {
-			localAlloc.BtcAmount = btcAmt
+			if localAlloc != nil {
+				localAlloc.BtcAmount = btcAmt
+			}
```
ffranr, line 263: If `btcAmt > 0` here then shouldn't `localAlloc != nil`? In other words, do we need to create an allocation here somehow?

ffranr (approved):
LGTM

tapchannel/aux_closer.go
guggero: No, the `localAlloc` is the allocation for assets.  And if there are more than 1000 sats in BTC available, they go into a separate, BTC only allocation (so to not mix the asset and BTC outputs too much). This `if` is for when we have neither BTC nor assets.

ffranr (approved):

PR #938, opened 2024-06-08
tapchannel: implement new AuxSweeper sub-system

Roasbeef:
In this PR, we implement yet another `tapchannel` sub-system, the `AuxSweeper`. This interface will be used by `lnd` to handle force close resolution on chain.


tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,402 @@
...
+	quit chan struct{}
+	wg   sync.WaitGroup
+}
+
+// NewAuxSweeper...
+func NewAuxSweeper(cfg *AuxSweeperCfg) *AuxSweeper {
+	return &AuxSweeper{
+		resolutionReqs: make(chan *resolutionReq),
+		cfg:            cfg,
+		quit:           make(chan struct{}),
```
guggero, line 106: Why not use the `ContextGuard`, with that we get the wait group, quit channel and all shutdown aware contexts for free.

tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,402 @@
...
+	sweepDesc tapscriptSweepDesc
+}
+
+// createAndSignSweepVpackets...
+func (a *AuxSweeper) createAndSignSweepVpackets(
+	sweepInputs []*cmsg.AssetOutput, signDesc input.SignDescriptor,
+	sweepDesc lfn.Result[tapscriptSweepDesc],
+) lfn.Result[[]*tappsbt.VPacket] {
+
+	// TODO(roasbeef): is this model actually worth it? maybe need AndThen
```
guggero: I'd say it doesn't really help readability... At least that's my first impression :sweat_smile:

tapchannel/aux_sweeper.go
```diff
@@ -322,53 +397,98 @@ func (a *AuxSweeper) resolveContract(req lnwallet.ResolutionReq,
...
+	var (
+		sweepDesc    lfn.Result[tapscriptSweepDesc]
+		assetOutputs []*cmsg.AssetOutput
+	)
 
 	switch req.Type {
 	// A non-delay output. This means the remote party force closed on
 	// chain.
-	case lnwallet.CommitNoDelay:
+	case input.TaprootRemoteCommitSpend:
```
guggero, line 649: Nice, this whole `switch` statement makes it very obvious and easy to understand what's going on :100: Cool that you were able to generalize the code so much.

fn/func.go
```diff
@@ -60,6 +60,18 @@ func FilterMap[T any, K comparable](s map[K]T, f func(T) bool) []T {
 	return output
 }
 
+// FlatMap applies the given mapping function to each element of the given
+// slice and concatenates the results into a new slice.
+func FlatMap[I, O any, S []I](s S, f func(I) []O) []O {
```
guggero, line 65: I think it would be very useful if (after the release is out, in a more quiet minute) to consolidate the `fn` package between `lnd` and `tapd`, then only use one. It happened to me several times that I didn't understand why a `WhenSome()` closure gave me compilation errors and it was just because the wrong `fn` package was being used :sweat_smile:

tapchannel/aux_sweeper.go
```diff
@@ -360,6 +384,190 @@ func commitRevokeSweepDesc(keyRing *lnwallet.CommitmentKeyRing,
...
+		)
+		if err != nil {
+			return err
+		}
+	}
+	for _, remoteAsset := range commitState.RemoteAssets.Val.Outputs {
+		err := assetOutputToVPacket(
+			fundingInputProofs, vPktsByAssetID, remoteAsset,
+			&a.cfg.ChainParams, req.CommitTx,
+			a.cfg.DefaultCourierAddr,
```
guggero, line 517: In the co-op close I added the desired delivery address of each party to the shutdown message. Since there is no interactivity here, I think we should perhaps negotiate that upfront, during channel funding?
Otherwise if I use a non-standard universe and force close, you as my peer wouldn't know what universe to fetch the proof from.

tapchannel/aux_sweeper.go
```diff
@@ -360,6 +384,190 @@ func commitRevokeSweepDesc(keyRing *lnwallet.CommitmentKeyRing,
...
+	if err != nil {
+		return fmt.Errorf("unable to create output "+
+			"commitments: %w", err)
+	}
+
+	// Before we ship things off, we'll also insert all the script keys
+	// (which have a tap leaf) associated with them into the db. This way
+	// we'll be able to recognize the output as belonging to the wallet.
+	//
+	// TODO(roasbeef): remote's output known, but only when it's a breach
```
guggero, line 569: Can we detect which ones are ours and which ones aren't? Otherwise we'll have unspent assets in our DB forever (because currently we don't mark assets as spent if their anchor UTXO is spent, only if the proof for spending the asset is imported into the same node again).

tapchannel/aux_sweeper.go
```diff
@@ -360,6 +384,190 @@ func commitRevokeSweepDesc(keyRing *lnwallet.CommitmentKeyRing,
...
+// created by the commitment transaction at a later step.
+func (a *AuxSweeper) importCommitTx(req lnwallet.ResolutionReq,
+	commitState *cmsg.Commitment, fundingInfo *cmsg.OpenChannel) error {
+
+	// To start, we'll re-create vPackets for all of the outputs of the
+	// commitment transaction.
+	//
+	// We'll use the fundingInfo proofs to create a vIn for each of them.
+	fundingInputProofs := make(map[asset.ID]*proof.Proof)
+	for _, fundingInput := range fundingInfo.FundedAssets.Val.Outputs {
```
guggero, line 495: Hmm, so with all these loops we basically unroll the result of `CreateAllocations` back into a virtual packet... I wonder if there was an easier way to do this, but honestly couldn't think of a better way off the top of my head.

guggero (commented):
Did a first pass, left a couple of comments.

jharveyb:
Scanned to make sure that this would sit cleanly alongisde #933 , looks like we have no new `tappsbt` constructors etc. so that aspect should be fine.

tapchannel/aux_sweeper.go
Roasbeef: Hehe, experimenting with some new stuff 🤓. Main thing I like here is that in cases when all you have are results, then you don't need any `if err != nil` in the function body, instead it goes only at the end. It also forces you to break up your logic into discrete computations/transformations, which are assembled together at the very end. 

Agree the syntax _does_ leave some to be desired though. 

Other benefit is that you can just accept/return `T` for generic funcs, and don't need to worry about: "oh, what if there's an error too?". I also added a way to map `(T, error)` into `Result[T]`.

tapchannel/aux_sweeper.go
Roasbeef: Yee, took a few iterations to get it into this current version. The logic is focused, as it just worries about how to make a sweep desc for a specific asset type, then the rest _should_ be able to stay the same. 

Note that it isn't aware yet of HTLC preimages, but my plan is to add that to the `tapscriptSweepDesc` struct. 

Initially I was going to refactor the lnd witness generation functions a bit more to re-use, but they're all mostly the same, and only the HTLC ones really deviate much. So opted to use as inspiration and lift what was identical.

fn/func.go
Roasbeef: Yeah I also started to have `lfn` vs `fn` at times once I ran into weird compile errors (func doesn't exist, what do you mean, ofc it does!). 

I also spoke with @ProofOfKeags elsewhere about just using a `go.work` in the top level package of lnd for `fn`. Then this way we don't need to do the module dance as much. Not sure how that interacts w/ external packages though.

tapchannel/aux_sweeper.go
Roasbeef: Yeah makes sense. 

FWIW, I added this here mainly so I could add an assertion that it was properly uploaded to the Universe. I saw in another commit, you started to do validation in the porter before moving on. Without that, I would have a transfer, but then universe upload fails as an inclusion proof was wrong or something. 

In the latest diff, both sides are able to complete+import the close proof themselves. Then the responder also complete the proof for the funding transaction itself. This would move to later live in the `chanvalidate` package. so as soon as the funding txn is confirmed, we complete the proof then try to validate before using it at all.

tapchannel/aux_sweeper.go
Roasbeef: Yeah I ended up resolving this for that very reason. Now we'll only import the outputs that we _know_ are ours.

tapchannel/aux_sweeper.go
Roasbeef: Heh yeh, there _def_ is some overlap there. I _did_ use the allocation creation for spending from the commitment outputs however. 

Was going to convert it, then just had a bit of sunken cost I guess. Not too late to consolidate either way. Would need to refactor a bit though, as we don't have things like the filtered htlc view here.

Roasbeef:
> Scanned to make sure that this would sit cleanly alongisde https://github.com/lightninglabs/taproot-assets/pull/933 , looks like we have no new tappsbt constructors etc. so that aspect should be fine.

Adding this on in the next push 👍

Roasbeef:
Pushed up a new set of commits.

tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,1475 @@
...
+			signDesc, vIn, &a.cfg.ChainParams,
+			scriptTree.TapTweak(),
+		)
+
+		// In this case, the witness isn't special, so we'll set the
+		// control block now for it.
+		vIn.TaprootLeafScript[0].ControlBlock = ctrlBlkBytes
+
+		log.Infof("signing vPacket for input=%v",
+			spew.Sdump(vIn.PrevID))
```
ffranr, line 267: Should this be a debug level log message?

tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,1475 @@
...
+// createAndSignSweepVpackets creates vPackets that sweep the funds from the
+// channel to the wallet, and then signs them as well.
+func (a *AuxSweeper) createAndSignSweepVpackets(
+	sweepInputs []*cmsg.AssetOutput, signDesc input.SignDescriptor,
+	sweepDesc lfn.Result[tapscriptSweepDesc],
+) lfn.Result[[]*tappsbt.VPacket] {
+
+	// Based on the sweep inputs, make vPackets that sweep all the inputs
+	// into a new output with a fresh script key. They won't have an
+	// intenral key set, we'll do that when we go to make the output to
```
ffranr, line 302: spelling: intenral

tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,1475 @@
...
+) lfn.Result[tapscriptSweepDesc] {
+
+	// We'll make the script tree for the to remote script (we're remote as
+	// this is their commitment transaction). We don't have an auxLeaf here
+	// as we're on the TAP layer.
+	toRemoteScriptTree, err := input.NewRemoteCommitScriptTree(
+		keyRing.ToRemoteKey, input.NoneTapLeaf(),
+	)
+	if err != nil {
+		return lfn.Err[tapscriptSweepDesc](err)
```
ffranr, line 347: I think we should enhance the error message here so that we can be sure to pin point its origin from the logs. Other similar opportunities in this file.

tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,1475 @@
...
+	return lfn.Ok(tapscriptSweepDesc{
+		scriptTree:     toLocalScriptTree,
+		ctrlBlockBytes: ctrlBlockBytes,
+	})
+}
+
+// addVPacket converts an asset outputs to the corresponding vPackets that can
+// be used to complete the proof needed to import a commitment transaction.
+// This new vPacket is added to the specified map.
+func assetOutputToVPacket(fundingInputProofs map[asset.ID]*proof.Proof,
```
ffranr, line 444: wrong function name in doc

tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,1475 @@
...
+				return tappsbt.TypeSimple
+			}
+			return tappsbt.TypeSplitRoot
+		}(),
+		Interactive:             true,
+		AnchorOutputIndex:       inclusionProof.OutputIndex,
+		AnchorOutputInternalKey: inclusionProof.InternalKey,
+		//nolint:ll
+		AnchorOutputTapscriptSibling: inclusionProof.CommitmentProof.TapSiblingPreimage,
+		// TODO(roasbeef): want full info here actually?
```
ffranr, line 497: Can this TODO be removed now? What extra info does it refer to?

tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,1475 @@
...
+		ProofSuffix: &assetProof,
+	}
+
+	// While we're here, we'll also replace the transaction stored in the
+	// proof with the correct one.
+	vOut.ProofSuffix.AnchorTx = *commitTx
+
+	// Finally, we'll set the delivery address to the default courier, so
+	// we publish the proof in the specified Universe.
+	vOut.ProofDeliveryAddress = courierAddr
```
ffranr, line 508: I don't think this comment is correct in general. The courier doesn't necessarily push to a universe server.

tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,1475 @@
...
+	}
+
+	return &localScriptTree, &remoteScriptTree, nil
+}
+
+// importCommitScriptKeys imports the script keys for the commitment outputs
+// into the local addr book.
+func (a *AuxSweeper) importCommitScriptKeys(req lnwallet.ResolutionReq) error {
+
+	ctxb := context.Background()
```
ffranr, line 647: This line can be moved to just above `for _, key := range keysToImport {` below.

tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,1475 @@
...
+		keysToImport = append(keysToImport, *localCommitScriptKey)
+
+	case req.CloseType == lnwallet.RemoteForceClose:
+		keysToImport = append(keysToImport, *remoteCommitScriptKey)
+
+	default:
+		return fmt.Errorf("unknown close type: %v", req.CloseType)
+	}
+
+	log.Infof("Importing script_keys=%v", spew.Sdump(keysToImport))
```
ffranr, line 676: debug log level instead perhaps?

tapchannel/aux_sweeper.go
Roasbeef: Yep, I've made the "full" information above by also specifying the script key.

tapchannel/aux_leaf_signer.go
```diff
@@ -355,6 +355,69 @@ func (s *AuxLeafSigner) verifyHtlcSignature(chanState *channeldb.OpenChannel,
...
+			Value: signDesc.DoubleTweak.Serialize(),
+		})
+
+		signingKey = input.DeriveRevocationPubkey(
+			signingKey, signDesc.DoubleTweak.PubKey(),
+		)
+	}
+
+	return *signingKey, leafToSign
+}
```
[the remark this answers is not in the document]
Roasbeef: Lint commit message?

tapchannel/aux_sweeper.go
```diff
@@ -0,0 +1,1475 @@
...
+// sweepAddrReq is a request to sweep a set of inputs to a specified address.
+type sweepAddrReq struct {
+	inputs []input.Input
+	change lnwallet.AddrWithKey
+
+	resp chan lfn.Result[sweep.SweepOutput]
+}
+
+// broadcastReq is used by the sweeper to notify us of a transaction broascast.
+type broadcastReq struct {
```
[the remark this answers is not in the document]
Roasbeef: So it's kinda both, it's letting us know the sweeper is about to broadcast something, which we'll use as an opportunity to finalize any relevant proofs.

tapchannel/aux_sweeper.go
Roasbeef: Good point. This is in place mainly so we can run assertions properly in the itests. Could consider making the param an option, then it would only be done with some other config guard.

proof/append.go
```diff
@@ -184,7 +188,8 @@ func CreateTransitionProof(prevOut wire.OutPoint,
 			OutputIndex: params.RootOutputIndex,
 			InternalKey: params.RootInternalKey,
 			CommitmentProof: &CommitmentProof{
-				Proof: *rootMerkleProof,
+				Proof:              *rootMerkleProof,
+				TapSiblingPreimage: params.RootTapscriptSibling,
```
guggero, line 192: Nice catch!

tapchannelmsg/records.go
```diff
@@ -1907,3 +1908,134 @@ func DecodeAuxShutdownMsg(blob tlv.Blob) (*AuxShutdownMsg, error) {
...
+}
+
+func dVpktList(r io.Reader, val interface{}, buf *[8]byte, _ uint64) error {
+	if typ, ok := val.(*VpktList); ok {
+		numPkts, err := tlv.ReadVarInt(r, buf)
+		if err != nil {
+			return err
+		}
+
+		if numPkts == 0 {
```
guggero, line 1968: Do we also need to sanity check the `numPkts` to avoid OOM below?

tapfreighter/chain_porter.go
```diff
@@ -527,7 +549,8 @@ func (p *ChainPorter) fetchInputProof(ctx context.Context,
 		ScriptKey: *scriptKey,
 		OutPoint:  &input.OutPoint,
 	}
-	inputProofBytes, err := p.cfg.AssetProofs.FetchProof(
+	log.Infof("fetching proof for input: ", spew.Sdump(inputProofLocator))
```
guggero, line 552: nit: degrade to `log.Tracef`?

asset/asset.go
```diff
@@ -1730,10 +1730,17 @@ func (a *Asset) deepEqual(allowSegWitIgnoreTxWitness bool, o *Asset) bool {
...
 
-	if !reflect.DeepEqual(a.ScriptKey, o.ScriptKey) {
+	// If both assets have a tweaked key
+	//
+	// TODO(roasbeef): fin
+	if !a.ScriptKey.PubKey.IsEqual(o.ScriptKey.PubKey) {
 		return false
 	}
 
+	/*if !reflect.DeepEqual(a.ScriptKey, o.ScriptKey) {
```
guggero, line 1740: Is this still temporary or should we actually continue just comparing the public key?

guggero (approved):
LGTM pending green CI :100:

proof/append.go
Roasbeef: Heh yeh this one took a while!

tapchannelmsg/records.go
Roasbeef: What would we limit it to? FWIW, we're always the ones generating these vPackets. 

I guess we can limit it to the total amount of outputs you can have on a commitment transaction. Will limit it to 1k for some breathing room.

tapfreighter/chain_porter.go
Roasbeef: Actually a left over debug log, will just remove.

PR #934, opened 2024-06-07
Add custom channels min remote reserve

GeorgeTsagk:
## Description

This PR adds a new const value of `1062` for the remote channel reserve for custom channels. We also use that const value as the push amount when opening the channel.

PR #939, opened 2024-06-08
More custom channel bugfixes

guggero:
Fixes the following issues found during testing:
 - Allow a push amount to be set in the RPC and CLI to improve bi-directional sending UX in regtest
 - Make sure we set the minimum possible channel reserve when opening the channel
 - Make sure the `FundChannel` RPC cannot be called when `tapd` is running in standalone mode (not integrated into `litd`)
 - Fix bug that caused channel funding to fail if we spent the full amount in the channel funding (and the input being spent was the split root output of a previous send)
 - Add proof courier delivery address to the shutdown messages and allocations, so individual proofs are uploaded to correct universes
 - Make sure the party that isn't the initiator of the channel shutdown correctly fetches their co-op close output proof


psbt_channel_funder.go
```diff
@@ -13,6 +13,12 @@ import (
 	"github.com/lightningnetwork/lnd/routing/route"
 )
 
+const (
+	// CustomChannelRemoteReserve is the custom channel minimum remote
+	// reserve that we'll use for our channels.
+	CustomChannelRemoteReserve = 1062
```
Roasbeef, line 19: Oddly precise number?

config.go
```diff
@@ -147,6 +147,11 @@ type Config struct {
 	// connecting to itself as a federation member.
 	RuntimeID int64
 
+	// EnableChannelFeatures indicates that tapd is running inside the
+	// Lightning Terminal daemon (litd) and can provide Taproot Asset
+	// channel functionality.
+	EnableChannelFeatures bool
```
Roasbeef, line 153: 👍

tapchannel/allocation.go
```diff
@@ -447,6 +453,7 @@ func DistributeCoins(inputs []*proof.Proof, allocations []*Allocation,
 				AnchorOutputInternalKey:      a.InternalKey,
 				AnchorOutputTapscriptSibling: sibling,
 				ScriptKey:                    a.ScriptKey,
+				ProofDeliveryAddress:         deliveryAddr,
```
Roasbeef, line 456: > In order for us to send the proofs to the correct proof courier address,
we need to be able to specify different delivery addresses for each
allocation.

In the future, or today?

tapchannel/aux_closer.go
```diff
@@ -110,13 +123,14 @@ func createCloseAlloc(isLocal, isInitiator bool, closeAsset *asset.Asset,
...
-		BtcAmount:           tapsend.DummyAmtSats,
-		SortTaprootKeyBytes: sortKeyBytes,
+		SplitRoot:            isInitiator,
+		InternalKey:          shutdownMsg.AssetInternalKey.Val,
+		ScriptKey:            asset.NewScriptKey(&scriptKey),
+		Amount:               closeAsset.Amount,
+		AssetVersion:         asset.V0,
+		BtcAmount:            tapsend.DummyAmtSats,
+		SortTaprootKeyBytes:  sortKeyBytes,
+		ProofDeliveryAddress: proofDeliveryUrl,
```
Roasbeef, line 133: Ah existing tests assumed that they were all using the same courier.

Roasbeef (approved):
LGTM 🐜

psbt_channel_funder.go
guggero: Yeah, not sure where exactly this is coming from. Took over the commit from https://github.com/lightninglabs/taproot-assets/pull/934, @GeorgeTsagk did some calculations to arrive at it.
Probably the 1% of our 100k default channel size plus some dust margin?

tapchannel/allocation.go
guggero: Today. Otherwise if the initiator of the shutdown process uses the non-default proof courier address, the respondent wouldn't ever learn that and wouldn't know where to download the proof from.

tapchannel/aux_closer.go
guggero: Before this commit the respondent to the shutdown process didn't actually know what proof courier the final proof would be uploaded to. Now the desired delivery address is sent along with the internal and script key, so the non-shutdown-initiator will know where to look for the proof.

guggero:
Added one more bugfix, kudos to @jamaljsr for the amazing testing.

Commit message reads:
```
tapchannel: account for BTC only balances

The allocation code incorrectly assumed that if there was no asset
balance on one side, there also wouldn't be a BTC balance. With that
incorrect assumptions, no commitment or anchor allocations were added if
there for example was a BTC push amount added in the channel opening
transaction.
```

Issue #948, opened 2024-06-12
[feature]: decimal display support pt. 2


jharveyb:
Addressing #925 , continuation of #930 .

- [x] Only encode non-default (non-zero) precision values.
- [x] Reject mints if the existing JSON metadata is too big to add the precision field to, if precision was specified.
- [x] On reissuance, fetch the precision from the asset metadata (if possible) of the asset group anchor, and assert equality with the reissuance request.
-- "Precision should be fixed for all issuances in an asset group"
- [x] display new precision field on some (?) RPC responses
-- ListAssets, ListTransfers, and the address-related RPC calls

dstadulis:
- [ ] Add guidance / explanation in documentation the rationale behind a decimal display / precision of four 
- [ ] Add recommended use of decimal display in documentation

Roasbeef:
https://github.com/lightninglabs/taproot-assets/pull/1059 will fix the last two bullet points here

PR #952, opened 2024-06-14
Precision field checking

jharveyb:
Addresses #948 .

The meta helpers need some unit tests.

### DIFF
--- a/proof/meta.go
+++ b/proof/meta.go
@@ -5,10 +5,14 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
+	"maps"
+	"math"
 
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightningnetwork/lnd/tlv"
+	"golang.org/x/exp/constraints"
 )
 
 // MetaType is the type of the meta data being revealed.
@@ -33,6 +37,15 @@ const (
 	// the universe and needs to be validated by all senders and receivers
 	// of the asset.
 	MetaDataMaxSizeBytes = 1024 * 1024
+
+	// MetadataDecDisplayKey is the JSON key used in the metadata field of a
+	// minted asset to express the decimal display of the minted asset.
+	MetadataDecDisplayKey = "decimal_display"
+
+	// maxDecDisplay is the maximum value of decimal display that a user can
+	// define when minting assets. Since the uint64 max value has 19 decimal
+	// places we will allow for a max of 12 decimal places.
+	MaxDecDisplay = uint32(12)
 )
 
 var (
@@ -42,8 +55,37 @@ var (
 	// ErrMetaDataTooLarge signals that the meta data is too large.
 	ErrMetaDataTooLarge = errors.New("meta data too large")
 
+	// ErrMetaTypeNegative signals that the given value is an invalid meta
+	// type because it is negative.
+	ErrMetaTypeNegative = errors.New("meta type cannot be negative")
+
+	// ErrMetaTypeTooLarge signals that the given value is an invalid meta
+	// type because it is too large.
+	ErrMetaTypeTooLarge = errors.New("meta type above limit of 255")
+
 	// ErrInvalidJSON signals that the meta data is not a valid JSON.
 	ErrInvalidJSON = errors.New("invalid JSON")
+
+	// ErrNotJSON is returned if the metadata is expected to be JSON but
+	// is another meta type.
+	ErrNotJSON = errors.New("metadata is not JSON")
+
+	// ErrDecDisplayInvalid is returned if the decimal display value is
+	// invalid.
+	ErrDecDisplayInvalid = errors.New("invalid decimal display")
+
+	// ErrDecDisplayTooLarge is returned if the decimal display value
+	// exceeds the current limit.
+	ErrDecDisplayTooLarge = errors.New("decimal display too large")
+
+	// ErrDecDisplayInvalidType is returned if the value in a JSON object
+	// assigned to the decimal display key is not a uint32.
+	ErrDecDisplayInvalidType = errors.New("decimal display JSON field is " +
+		"not a number")
+
+	// ErrDecDisplayMissing is returned if the decimal display key is
+	// not present in a JSON object.
+	ErrDecDisplayMissing = errors.New("decimal display field missing")
 )
 
 // MetaReveal is an optional TLV type that can be added to the proof of a
@@ -59,6 +101,11 @@ type MetaReveal struct {
 	Data []byte
 }
 
+// A subset of Integer that excludes int8, since we never use it in practice.
+type SizableInteger interface {
+	constraints.Unsigned | ~int | ~int16 | ~int32 | ~int64
+}
+
 // Validate validates the meta reveal.
 func (m *MetaReveal) Validate() error {
 	// A meta reveal is allowed to be nil.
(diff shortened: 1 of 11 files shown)


proof/meta.go
```diff
@@ -85,6 +131,188 @@ func (m *MetaReveal) Validate() error {
...
+	case currentDecDisplay == decDisplay:
+		return &MetaReveal{
+			Type: m.Type,
+			Data: m.Data,
+		}, nil
+
+	// Our metadata is invalid in another way.
+	default:
+		return nil, fmt.Errorf("%w: %d", ErrDecDisplayTooLarge,
+			decDisplay)
```
ffranr, line 312: The error here seems very specific given the switch clause.

rpcserver.go
```diff
@@ -510,36 +504,45 @@ func (r *rpcServer) MintAsset(ctx context.Context,
...
+		if metaType == proof.MetaJson && req.Asset.DecimalDisplay != 0 {
+			updatedMeta, err := seedlingMeta.SetDecDisplay(
+				req.Asset.DecimalDisplay,
+			)
+			if err != nil {
+				return nil, err
+			}
+
+			seedlingMeta = updatedMeta
+		}
```
ffranr, line 545: I think the comment promises that we execute `seedlingMeta.Validate()` again, but I don't think we do.

tapdb/asset_minting.go
```diff
@@ -964,27 +964,67 @@ func fetchAssetMetas(ctx context.Context, db PendingAssetStore,
...
 }
 
+// fetchMetaByAssetID fetches a single asset meta reveal.
+func fetchMetaByAssetID(ctx context.Context, db PendingAssetStore,
+	ID asset.ID) (*proof.MetaReveal, error) {
+
+	assetMeta, err := db.FetchAssetMetaForAsset(ctx, ID[:])
+	switch {
+	case errors.Is(err, sql.ErrNoRows):
+		return nil, nil
```
ffranr, line 992: I think this sort of case is where we want to `return fn.None[proof.MetaReveal](), nil`
But what you have should work so it might do for now. Maybe add a TODO?

tapgarden/batch.go
```diff
@@ -141,6 +141,18 @@ func (m *MintingBatch) validateGroupAnchor(s *Seedling) error {
...
+	// anchor. We already validated the seedling metadata, so we don't care
+	// if the value is explicit or if the metadata is JSON, but we must
+	// compute the same value for both assets.
+	_, seedlingDecDisplay, _ := s.Meta.GetDecDisplay()
+	_, anchorDecDisplay, _ := anchor.Meta.GetDecDisplay()
+	if seedlingDecDisplay != anchorDecDisplay {
+		return fmt.Errorf("seedling decimal display does not match "+
+			"group anchor: %d, %d", seedlingDecDisplay,
+			anchorDecDisplay)
+	}
```
ffranr, line 154: :+1:

tapgarden/planter.go
```diff
@@ -1653,18 +1653,28 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
...
+		anchorMeta, err := c.cfg.Log.FetchAssetMeta(
+			ctx, groupInfo.Genesis.ID(),
+		)
+		if err != nil {
+			return fmt.Errorf("group anchor genesis %x not found: "+
+				"%w", groupKeyBytes, err,
+			)
+		}
+
+		err = req.validateGroupKey(*groupInfo, *anchorMeta)
```
ffranr, line 1676: I might be wrong, but I think this exemplifies why `fetchMetaByAssetID` (and `FetchAssetMeta`) should return an `fn.Option[proof.MetaReveal]`. I think that at this point `anchorMeta` could be nil. And so `*anchorMeta` is trouble.

ffranr (changes requested):
We need to double check one line for a potential nil pointer dereference.

ffranr:
I've forced pushed a minimal change to address the nil pointer dereference issue.

rpcserver.go
ffranr: I've added the re-validate bit that the in-code comment called for.

ffranr:
I've force pushed once more to address: https://github.com/lightninglabs/taproot-assets/pull/952#discussion_r1642862777

ffranr (approved):
Approve, but I've made some modifications. See PR comments.

proof/meta.go
```diff
@@ -85,6 +131,188 @@ func (m *MetaReveal) Validate() error {
...
+
+	// Fetch the current decimal display value.
+	currentMetaJSON, currentDecDisplay, err := m.GetDecDisplay()
+	switch {
+	// The current metadata is valid JSON. Either no decimal display value
+	// is present, or it is but doesn't match the desired value.
+	case err == ErrDecDisplayMissing, currentDecDisplay != decDisplay:
+		// If the requested decimal display value is 0, we don't need to
+		// add the field at all, as that is the default.
+		if decDisplay == 0 {
```
GeorgeTsagk, line 277: I think it's better if we always set it, even if it is the default `0` value. Otherwise API users have to also assume the default value which may be unclear compared to the field always being present

proof/meta.go
GeorgeTsagk: +1, would expect a more generic `ErrInvalidDecDisplay` error ?

tapgarden/batch.go
```diff
@@ -141,6 +141,18 @@ func (m *MintingBatch) validateGroupAnchor(s *Seedling) error {
 			*s.GroupAnchor)
 	}
 
+	// The decimal display of the seedling must match that of the group
+	// anchor. We already validated the seedling metadata, so we don't care
+	// if the value is explicit or if the metadata is JSON, but we must
+	// compute the same value for both assets.
+	_, seedlingDecDisplay, _ := s.Meta.GetDecDisplay()
+	_, anchorDecDisplay, _ := anchor.Meta.GetDecDisplay()
```
GeorgeTsagk, line 149: do we always assume that the group anchor is in the same batch?

GeorgeTsagk (commented):
looking good

tapgarden/batch.go
ffranr: In `tapgarden.Seedling` (which is a bit like the batch minting request) the group anchor is optional. But if set, then an attempt is made to retrieved  the anchor from the batch. So, as far as I can tell, the group anchor doesn't have to be in the batch.

GeorgeTsagk: a) Should we also check if `anchor != nil` here?
b) So if the anchor is not in the batch, do we skip the decimal display check?

ffranr: In this method we've already retrieved the anchor using `anchor, ok := m.Seedlings[*s.GroupAnchor]`. And we return an error if it's missing. We should never really be adding a `nil` entry into `m.Seedlings`, but I could double check at this point. I will add that check.

ffranr: I've gone with
```
	if anchor == nil || !ok {
		return fmt.Errorf("group anchor %v not present in batch",
			s.GroupAnchor)
	}
```

tapgarden/planter.go
```diff
@@ -1653,18 +1653,28 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
...
+			ctx, groupInfo.Genesis.ID(),
+		)
+		if err != nil {
+			return fmt.Errorf("group anchor genesis %x not found: "+
+				"%w", groupKeyBytes, err,
+			)
+		}
+
+		err = req.validateGroupKey(*groupInfo, anchorMeta)
+		if err != nil {
```
ffranr, line 1677: @GeorgeTsagk here's where the metadata is validated against a group anchor that's not in the batch. And in the next clause down metadata validation is performed for instances where the group anchor **is** included in the batch.

tapgarden/batch.go
ffranr: > a) Should we also check if `anchor != nil` here?
> b) So if the anchor is not in the batch, do we skip the decimal display check?

(b) is covered here: https://github.com/lightninglabs/taproot-assets/pull/952#discussion_r1643136426

GeorgeTsagk (approved):
LGTM

PR #954, opened 2024-06-18
Fix proof storage clobber error for postgres backend

### LINKED ISSUE #951: [bug]/tapdb: proof storage clobber error for postgres backend
### Background

This failure has started to pop up on itests for the postgres backend: 
```
2024-06-12 16:42:33.793 [INF] GRDN: Received new proof file for asset ID 9bc313ea555c28227c86fa4a2b90f54dff46b14db4bbc56bd5d5e2e4dc3af34f, version=0,num_proofs=3
2024-06-12 16:42:33.807 [INF] FRTR: Marking parcel (txid=80bf98b4af11c552c95557806afa79468edec7b0a947f2c7b58306b0d4df4211) as confirmed!
2024-06-12 16:42:33.814 [ERR] FRTR: unable to transfer receiver proof: unable to log parcel delivery confirmation: failed to confirm transfer: unknown postgres error: ERROR: more than one row returned by a subquery used as an expression (SQLSTATE 21000)
```

It can be seen as a test failure in this itest run: https://github.com/lightninglabs/taproot-assets/actions/runs/9486428727/job/26140728303 
```
    --- FAIL: TestTaprootAssetsDaemon/multi_address (13.48s)
```

The issue appears to be with this query: https://github.com/lightninglabs/taproot-assets/blob/51e3512464e1efbed2abb6b3ea50f26fa41802f5/tapdb/sqlc/queries/assets.sql#L671-L691

If the CTE for `target_asset` returns more than one `asset_id`, then this issue will be triggered. AFAICT, sqlite will just take the very first asset ID. 

If we look at the table, we see that `asset_id` here is intended to be unique: https://github.com/lightninglabs/taproot-assets/blob/51e3512464e1efbed2abb6b3ea50f26fa41802f5/tapdb/sqlc/migrations/000002_assets.up.sql#L245-L256

This appears to be the root of the issue, as if you insert proofs for two diff script keys, but the same asset ID, then behavior is undefined across the two DB backends. However, this is only an issue for the db based proof file storage. The file based storage uses a full path of the outpoint+assetID+scriptKey (prev ID 3-tuple). The Universe based storage uses something similar. 

Most sub-systems end up using a `MultiArchive` that's based on some or all the proof storage backends, which my answer why this issue has been hidden for some time. 

### Suggested Fix

First we need to get a better feel of the impact. When we send out using the `ChainPorter`, the file based proofs are updated first, then later the DB proofs once proofs are delivered. 

Ultimately, I think we need a DB migration here to add the outpoint+script key as a composite 3-tuple primary key, similar to what we've done elsewhere. 

### Steps to reproduce

Run `make itest icase=multi_address` on postgres.

### Expected behavior

Should pass. 

### Actual behavior

Fails.

gijswijs:
Fixes #951.
An (intermitent) failure popped up when running itest on the postgres backend.

This draft PR tries to solve this error, initially by unifying the behavior of both backends.

Currenlty the status is that this leads to an error later in the process (only for postgres), where the given number of events rhat exist with the given status is of by one.

### DIFF
--- a/tapdb/asset_minting.go
+++ b/tapdb/asset_minting.go
@@ -113,12 +113,13 @@ type (
 	// MintingBatchInit is used to create a new minting batch.
 	MintingBatchInit = sqlc.NewMintingBatchParams
 
-	// ProofUpdate is used to update a proof file on disk.
-	ProofUpdate = sqlc.UpsertAssetProofParams
-
 	// ProofUpdateByID is used to update a proof file on disk by asset
 	// database ID.
 	ProofUpdateByID = sqlc.UpsertAssetProofByIDParams
+
+	// FetchAssetID is used to fetch the primary key ID of an asset, by
+	// outpoint and tweaked script key.
+	FetchAssetID = sqlc.FetchAssetIDParams
 
 	// NewScriptKey wraps the params needed to insert a new script key on
 	// disk.
@@ -222,12 +223,14 @@ type PendingAssetStore interface {
 	FetchAssetsForBatch(ctx context.Context, rawKey []byte) ([]AssetSprout,
 		error)
 
-	// UpsertAssetProof inserts a new or updates an existing asset proof on
-	// disk.
-	//
-	// TODO(roasbeef): move somewhere else??
-	UpsertAssetProof(ctx context.Context,
-		arg sqlc.UpsertAssetProofParams) error
+	// FetchAssetID fetches the `asset_id` (primary key) from the assets
+	// table for a given asset identified by `Outpoint` and
+	// `TweakedScriptKey`.
+	FetchAssetID(ctx context.Context, arg FetchAssetID) ([]int64, error)
+
+	// UpsertAssetProofByID inserts a new or updates an existing asset
+	// proof on disk.
+	UpsertAssetProofByID(ctx context.Context, arg ProofUpdateByID) error
 
 	// FetchAssetMetaForAsset fetches the asset meta for a given asset.
 	FetchAssetMetaForAsset(ctx context.Context,
@@ -1644,9 +1647,28 @@ func (a *AssetMintingStore) MarkBatchConfirmed(ctx context.Context,
 		// As a final act, we'll now insert the proof files for each of
 		// the assets that were fully confirmed with this block.
 		for scriptKey, proofBlob := range mintingProofs {
-			err := q.UpsertAssetProof(ctx, ProofUpdate{
-				TweakedScriptKey: scriptKey.CopyBytes(),
-				ProofFile:        proofBlob,
+			// We need to fetch the table primary key `asset_id`
+			// first, as we need it to update the proof.
+			dbAssetIds, err := q.FetchAssetID(ctx, FetchAssetID{
+				TweakedScriptKey: scriptKey[:],
+			})
+			if err != nil {
+				return err
+			}
+
+			// We should not have more than one `asset_id`.
+			if len(dbAssetIds) > 1 {
+				return fmt.Errorf("expected 1 asset id, found "+
+					"%d with asset ids %v",
+					len(dbAssetIds), dbAssetIds)
+			}
+
+			// Upload proof by the dbAssetId, which is the _primary
+			// key_ of the asset in table assets, not the BIPS
+			// concept of `asset_id`.
+			err = q.UpsertAssetProofByID(ctx, ProofUpdateByID{
+				AssetID:   dbAssetIds[0],
+				ProofFile: proofBlob,
 			})
 			if err != nil {
 				return fmt.Errorf("unable to insert proof "+
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -215,9 +215,10 @@ type ActiveAssetsStore interface {
 	UpsertManagedUTXO(ctx context.Context, arg RawManagedUTXO) (int64,
 		error)
 
-	// UpsertAssetProof inserts a new or updates an existing asset proof on
-	// disk.
-	UpsertAssetProof(ctx context.Context, arg ProofUpdate) error
+	// FetchAssetID fetches the `asset_id` (primary key) from the assets
+	// table for a given asset identified by `Outpoint` and
+	// `TweakedScriptKey`.
+	FetchAssetID(ctx context.Context, arg FetchAssetID) ([]int64, error)
 
 	// UpsertAssetProofByID inserts a new or updates an existing asset
 	// proof on disk.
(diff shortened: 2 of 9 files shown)


gijswijs:
Also fixes #961

tapdb/asset_minting.go
```diff
@@ -113,13 +113,18 @@ type (
 	// MintingBatchInit is used to create a new minting batch.
 	MintingBatchInit = sqlc.NewMintingBatchParams
 
+	// ProofUpdate is used to update a proof file on disk.
```
guggero, line 116: nit: duplicate line.

tapdb/assets_store.go
```diff
@@ -218,6 +218,9 @@ type ActiveAssetsStore interface {
 	// disk.
 	UpsertAssetProof(ctx context.Context, arg ProofUpdate) error
 
+	FetchAssetID(ctx context.Context,
```
guggero, line 222: nit: missing Godoc. Also, should this be called `QueryAssetPrimaryID`? To avoid confusion with the actual `AssetID` type (32-byte vs. numerical database ID). And query to signify we're querying by script key and/or outpoint?

tapdb/assets_store.go
```diff
@@ -1615,10 +1618,34 @@ func (a *AssetStore) upsertAssetProof(ctx context.Context,
...
-	return db.UpsertAssetProof(ctx, ProofUpdate{
-		TweakedScriptKey: scriptKeyBytes,
-		Outpoint:         outpointBytes,
-		ProofFile:        proof.Blob,
+
+	// We need to fetch the asset ID first, as we need it to update the
+	// proof. We could do this in one query, using `UpsertAssetProof` but
+	// this gave issues with a postgresql backend. See:
+	// https://github.com/lightninglabs/taproot-assets/issues/951
+	dbAssetIds, err := db.FetchAssetID(ctx,
```
guggero, line 1629: nit: formatting.

tapdb/assets_store.go
```diff
@@ -1615,10 +1618,34 @@ func (a *AssetStore) upsertAssetProof(ctx context.Context,
...
+	if err != nil {
+		return err
+	}
+
+	// We are erroring out if the first query returns more than one result.
+	// This is what the subquery of `UpsertAssetProof` did at times. We are
+	// not sure why this happens but if it happens with the separated
+	// queries we want to error out here as well.
+	if len(dbAssetIds) > 1 {
+		return fmt.Errorf("expected 1 asset id, found %d with asset"+
```
guggero, line 1643: nit: missing space within string before line break.

tapdb/sqlc/queries/assets.sql
```diff
@@ -682,6 +682,7 @@ WITH target_asset(asset_id) AS (
             OR sqlc.narg('tweaked_script_key') IS NULL)
         AND (utxos.outpoint = sqlc.narg('outpoint')
             OR sqlc.narg('outpoint') IS NULL)
+    LIMIT 1
```
guggero, line 698: Since the CTE seems to be the actual problem, should we change this query so it takes the asset primary DB ID? Since we're querying that ID in the next commit.

tapdb/assets_store.go
```diff
@@ -379,6 +380,8 @@ type AssetStore struct {
 	clock clock.Clock
 
 	txHeights *lru.Cache[chainhash.Hash, cacheableBlockHeight]
+
+	sync.Mutex
```
guggero, line 384: Should give this mutex a name, so it's more clear where it's used. For example: `proofInsertMtx`?

tapdb/assets_store.go
```diff
@@ -1572,6 +1575,8 @@ func (a *AssetStore) importAssetFromProof(ctx context.Context,
 	// As a final step, we'll insert the proof file we used to generate all
 	// the above information.
 	scriptKeyBytes := newAsset.ScriptKey.PubKey.SerializeCompressed()
+	a.Lock()
```
guggero: nit: move this to above the comment into its own visual block (empty line before and after) to stand out more clearly?

guggero: I guess this will be removed by the next commit anyway, but generally we try to make the use of mutexes stand out visually.

guggero (commented):
Nice fix! Have it running on my machine for multiple minutes as well now and so far no problem.

tapdb/sqlc/queries/assets.sql
gijswijs: We already have that query down below: `UpsertAssetProofByID`
Better to remove `UpsertAssetProof` and use two separate queries by default.

tapdb/asset_minting.go
```diff
@@ -1644,9 +1647,28 @@ func (a *AssetMintingStore) MarkBatchConfirmed(ctx context.Context,
 		// As a final act, we'll now insert the proof files for each of
 		// the assets that were fully confirmed with this block.
 		for scriptKey, proofBlob := range mintingProofs {
-			err := q.UpsertAssetProof(ctx, ProofUpdate{
-				TweakedScriptKey: scriptKey.CopyBytes(),
-				ProofFile:        proofBlob,
+			// We need to fetch the table primary key `asset_id`
+			// first, as we need it to update the proof.
+			dbAssetIds, err := q.FetchAssetID(ctx, FetchAssetID{
```
GeorgeTsagk, line 1652: IIUC I could have assets under the same scriptkey, so more than 1 values could be returned here? (or in a similar context)

do we need assets to be under exclusive script keys?

tapdb/assets_store.go
```diff
@@ -1567,13 +1568,11 @@ func (a *AssetStore) importAssetFromProof(ctx context.Context,
...
 
-	// As a final step, we'll insert the proof file we used to generate all
-	// the above information.
-	scriptKeyBytes := newAsset.ScriptKey.PubKey.SerializeCompressed()
-	return db.UpsertAssetProof(ctx, ProofUpdate{
-		TweakedScriptKey: scriptKeyBytes,
-		Outpoint:         anchorPoint,
-		ProofFile:        proof.Blob,
+	// Upload proof by the dbAssetId, which is the _primary key_ of the
+	// asset in table assets, not the BIPS concept of `asset_id`.
```
GeorgeTsagk, line 1572: :+1:

GeorgeTsagk (approved):
LGTM
was this originally detected via a flake? just asking whether we added coverage for it or not

tapdb/asset_minting.go
gijswijs: We will address this in a different PR. Made an issue to keep track. #966

Issue #942, opened 2024-06-10
Polar Test Flows Working/Not Working


levmi:
(h/t to @jamaljsr for running these tests and documenting it all) Jamal ran through test flows on Polar to exhaustively test what is currently working with the latest vs. what is not working. Those flows and their outcomes are outlined below. 

The network structure is the following with all `litd` nodes except for `erin` which is Core Lightning. I've re-tested using the latest `litd branch 0-19-staging` (https://github.com/lightninglabs/lightning-terminal/commit/980ee938aa1e7a45633216aff8775ca8d05c12aa)

```
alice --[asset]-> bob --[sats]-> carol --[asset]-> dave
                                   |
                                 [sats]
                                   |
                                   v
                                  erin (core-ln)
```

Working

- [x]  fund asset channel using full asset balance after mint
- [x] fund asset channel using partial asset balance after mint
- [x] fund asset channel using full asset balance after mint + onchain send
- [x] fund asset channel using partial asset balance after mint + onchain send
- [x] fund asset channel using full asset balance from onchain receive
- [x] fund asset channel using partial asset balance from onchain receive
- [x] fund asset channel with `push_sat=5000`
- [x] create asset invoice on dave, pay with asset on alice (multi-hop)
- [x] create asset invoice on bob, pay with asset on alice (direct peer)
- [x] create asset invoice on alice, pay with asset on bob (direct peer, opposite direction)
- [x] create asset invoice on dave, pay with sats on bob (sats -> asset)
- [x]  create sats invoice on bob, pay with asset on dave (asset -> sats)
- [x] create asset invoice on alice, pay with sats on erin (cln sats -> asset)
- [x] create sats invoice on erin, pay with asset on alice (asset -> cln sats)
- [x] send asset keysend payment from alice to bob

Not Working
- [x] create asset invoice on alice, pay with asset on dave (multi-hop, opposite direction after funding sending half asset amt. `FAILURE_REASON_NO_ROUTE`)
- [x] create sats invoice on bob, pay with sats on alice (direct peer, sats -> sats. `FAILURE_REASON_INSUFFICIENT_BALANCE`) --> Solved by https://github.com/lightninglabs/taproot-assets/pull/962
- [x] create sats invoice on dave, pay with sats on alice (multi-hop, sats -> sats. `FAILURE_REASON_INSUFFICIENT_BALANCE`) --> Solved by https://github.com/lightninglabs/taproot-assets/pull/962
[...]

PR #959, opened 2024-06-19
tapchannelmsg: increase channel data reader max record size

ffranr:
Testing indicated that when using large but valid asset amounts the channel custom records buffer size was greater than `tlv.MaxRecordSize`.

This commit introduces a new greater constant which will be used as the channel data max record size.

Addresses bug reported here: https://github.com/lightninglabs/taproot-assets/issues/942#issuecomment-2174129269

What is the correct value for this new constant?

NOTE: I haven't finished investigating why the record takes up more memory.

### DIFF
--- a/tapchannelmsg/custom_channel_data.go
+++ b/tapchannelmsg/custom_channel_data.go
@@ -10,15 +10,12 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/taproot-assets/rfqmsg"
 	"github.com/lightningnetwork/lnd/lnrpc"
-	"github.com/lightningnetwork/lnd/tlv"
 	"google.golang.org/protobuf/proto"
 )
 
 // ReadOpenChannel reads the content of an OpenChannel struct from a reader.
-func ReadOpenChannel(r io.Reader) (*OpenChannel, error) {
-	openChanData, err := wire.ReadVarBytes(
-		r, 0, tlv.MaxRecordSize, "chan data",
-	)
+func ReadOpenChannel(r io.Reader, maxReadSize uint32) (*OpenChannel, error) {
+	openChanData, err := wire.ReadVarBytes(r, 0, maxReadSize, "chan data")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


tapchannelmsg/custom_channel_data.go
```diff
@@ -14,10 +14,14 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// ChannelDataMaxRecordSize is the maximum size of the custom channel data that
+// can be read from a reader.
+const ChannelDataMaxRecordSize = tlv.MaxRecordSize * 2
```
Roasbeef, line 19: Have we been able to validate that this resolves the issue? 

The other thing to consider here is that w/o adding any proof level deduplication, this'll expand along with the number of active HTLCs. So perhaps it should be a multiplier based on the current max proof suffix size. We know the max amount of HTLCs in either direction, so we can base it off that. 

Also unlike other contexts, we're the one creating this value, so it's a bit more trusted.

Roasbeef (changes requested):

tapchannelmsg/custom_channel_data.go
ffranr: The data is already in memory. I've pushed up a new commit which adds a "max read size" argument which we can set based on the data that's already in memory.

I've tested this with the litd itest and it solves the max read size issue.

I've also created a unit test on this branch so that we can play around with the custom channel data which exceeded the previous max read limit:
https://github.com/lightninglabs/taproot-assets/commit/2a90bb366a0c03d68634723b6ae581e7c9266276

I didn't see anything particularly interesting in that custom channel data example. It's commitment portion just bigger than the default TLV record size.

guggero (approved):
LGTM :tada: 

Def makes sense to just use the length here as we exactly know how many bytes there are.
Thanks for validating with the linked unit test. Looking at the test data, I see there are only 14 HTLCs that cause the big size, probably mostly the proofs.

Roasbeef (approved):
LGTM 🎢

PR #960, opened 2024-06-19
Add decimal display to `taprpc.Asset`

GeorgeTsagk:
## Description

Recently we supported a new `decimal_display` field when minting which would be encoded in the asset metadata. This PR tries to expose that field back to the user via commonly used asset RPC calls


proof/courier.go
```diff
@@ -1053,8 +1053,11 @@ func (c *UniverseRpcCourier) DeliverProof(ctx context.Context,
 		proofAsset := transitionProof.Asset
 
 		// Construct asset leaf.
+		// TODO(george): We pass in a 0 dec display as we have no easy
+		// access to that value here, would have to pollute interfaces.
+		// Does this value matter in the proof flow?
 		rpcAsset, err := taprpc.MarshalAsset(
-			ctx, &proofAsset, true, true, nil,
+			ctx, &proofAsset, true, true, nil, 0,
```
Roasbeef, line 1060: Could make it an optional value? 

Also perhaps we should make it a new proto message. This way, we can distinguish it being set as `0` vs not being set at all.

taprpc/marshal.go
```diff
@@ -444,7 +444,8 @@ func UnmarshalGroupWitness(wit *GroupWitness) (*asset.PendingGroupWitness,
 
 // MarshalAsset converts an asset to its rpc representation.
 func MarshalAsset(ctx context.Context, a *asset.Asset,
-	isSpent, withWitness bool, keyRing KeyLookup) (*Asset, error) {
+	isSpent, withWitness bool, keyRing KeyLookup,
+	decDisplay uint32) (*Asset, error) {
```
Roasbeef, line 448: See comment. Above, I think we should either make this an option, and/or modify the proto to be its own message, that way we can check for nil vs not.

Roasbeef (changes requested):

taprpc/taprootassets.proto
```diff
@@ -348,6 +348,25 @@ message GenesisReveal {
     GenesisInfo genesis_base_reveal = 1;
 }
 
+message DecimalDisplayOpt {
```
guggero, line 351: nit: why "Opt"? And not just DecimalDisplay?

taprpc/taprootassets.proto
```diff
@@ -348,6 +348,25 @@ message GenesisReveal {
...
+    // This value signals that the decimal display value has been set.
+    bool is_set = 1;
```
guggero, line 353: Looking at the RPC code in the next commit, wouldn't the whole `decimal_display` message just be `null` if it wasn't set? So the boolean is kind of superfluous as it would always be `true` (if the parent message is there)?

taprpc/marshal.go
```diff
@@ -474,6 +475,13 @@ func MarshalAsset(ctx context.Context, a *asset.Asset,
 		IsBurn:                 a.IsBurn(),
 	}
 
+	decDisplay.WhenSome(func(u uint32) {
+		rpcAsset.DecimalDisplay = &DecimalDisplayOpt{
```
guggero, line 479: See my previous comment. If we want the boolean `IsSet` to be useful, we should set `DecimalDisplay: &DecimalDisplayOpt{},` above where we initialize `rpcAsset`.

proof/courier.go
```diff
@@ -1053,8 +1053,11 @@ func (c *UniverseRpcCourier) DeliverProof(ctx context.Context,
 		proofAsset := transitionProof.Asset
 
 		// Construct asset leaf.
+		// TODO(george): We pass in a 0 dec display as we have no easy
```
guggero, line 1056: Actually, the whole serialized asset doesn't really matter in the proof courier case. So IMO you can remove the TODO and just add a comment that the courier currently doesn't look at the RPC marshaled asset anyway.

rpcserver.go
```diff
@@ -4866,7 +4879,7 @@ func marshalAssetLeaf(ctx context.Context, keys taprpc.KeyLookup,
 	assetLeaf *universe.Leaf) (*unirpc.AssetLeaf, error) {
 
 	rpcAsset, err := taprpc.MarshalAsset(
-		ctx, assetLeaf.Asset, false, true, keys,
+		ctx, assetLeaf.Asset, false, true, keys, fn.None[uint32](),
```
guggero, line 4882: This (`marshalAssetLeaf`) is called from multiple places. Should we add the parameter here too and then decide at each call site if it would be possible to get the decimal display or not?
Mainly thinking about `marshalUniverseProofLeaf`, `marshalUniverseDiff` and `AssetLeaves` where the decimal display would be useful to see.

guggero (commented):
Nice :tada: 
Just a couple of suggestions where we might also want to expose the value.

taprpc/taprootassets.proto
GeorgeTsagk: because it's optional, as in "it might be there"

taprpc/taprootassets.proto
GeorgeTsagk: I thought `null` did not exist in the proto world? it will instead set the default values so `is_set` will be `false`

taprpc/marshal.go
GeorgeTsagk: could add this, to make it more explicit

rpcserver.go
GeorgeTsagk: Main thing I encountered when coming from universe related code paths is that there's no access to the meta reveal, which seems to require `AssetStore`. Unless there's some other (easier) way to acquire the meta reveal, should I add asset store access to all the subsystems that call this?

taprpc/taprootassets.proto
guggero: It does, but only for messages, not for primitive types.

rpcserver.go
guggero: Hmm, I see. Let's skip it then. But then I think we should keep the `DecimalDisplay` message as `null` on the RPC level. Otherwise it would look like such an asset on the universe RPC side doesn't have a decimal display set, but in reality we just don't know. So maybe, to be able to show the three different states (we don't know if decimal display is set, it is explicitly not set, it is explicitly set), we should pass in the `*taprpc.DecimalDisplay` message here directly. When it's `nil`, we don't know, when it's not `nil` the boolean inside has a meaning.

taprpc/taprootassets.proto
guggero: Yeah, but that is already signified by a child member of that message. So I think it's not necessary?

taprpc/taprootassets.proto
```diff
@@ -424,6 +443,11 @@ message Asset {
     // determined for script keys that aren't known to the wallet of the local
     // tapd node.
     bool script_key_has_script_path = 19;
+
+    // This field defines a decimal display value that may be present. This is
+    // a wrapper around the uint32 that signals whether the value has been set
+    // or not.
```
guggero, line 449: Add "If this field is null, it means the presence of a decimal display field is unknown in the current context".

taprpc/taprootassets.proto
GeorgeTsagk: ok, changing

taprpc/marshal.go
GeorgeTsagk: ok so I went with the approach where `is_set` is omitted, and we rely on the nested message being nil or not to signal whether a value has been set

Roasbeef (approved):
LGTM 🧤

Issue #691, opened 2023-11-21
Fix `assets` SQL table unique constrinat


ffranr:
The `assets` SQL table's unique constraint is ineffective (it's a no-op). It doesn't do what we intend it to do.

The current unique constraint is on the fields:
```
asset_id, genesis_id, script_key_id
```

but `asset_id` is the table's primary key integer and not the asset ID byte array. This means that each row is unique and that duplicate assets can be inserted into the table.

We add a unique constraint index with the following fields to the table:
```
anchor_utxo_id, genesis_id, script_key_id
```

With this new constraint we should not be able to add duplicate assets into the table.

Adding the new constraint will mean deleting rows from the `assets` table which violate this new constraint.

Another change that follows naturally from this modification is to change the `InsertNewAsset` SQL statement into an upsert.

This issue has caused this bug https://github.com/lightninglabs/taproot-assets/issues/665 but another solution has now addressed that bug.

PR https://github.com/lightninglabs/taproot-assets/pull/684 is an attempt at fixing this issue.

dstadulis:
I had started writing an issue for these needs.  Feel free to edit or lift any relevant elements.

### Issue title: 
Test duplicate address receive flow + DB uniqueness

### Description: 
This test will ensure that if a client tries to send to two duplicate addresses (say in the same transfer), that the daemon throw an error, we'll want to note if it gets clobbered in an internal map. It's important to make sure this is tested to avoid any potential issues with transactions being processed incorrectly. The unit tests should cover the scenarios:

1) Global/DB uniqueness: Test the `outpoint + assetID + scriptKey` are indeed *globally unique and sufficient to ensure unique DB location* and compatible with address reuse.

2) Batch collision: `internalKey + assetID + scriptKey` don't collide within a batch.

3) Outpoint collision: `assetID + scriptKey` don't collide within an outpoint.

4) Duplicate Sends: Test should also include a scenario where a client sends to multiple different addresses, but one of them is duplicated - to test if only the duplicates get rejected or if all transactions fail due to an error.

ffranr:
We should complete issue https://github.com/lightninglabs/taproot-assets/issues/694 before addressing this issue.

gijswijs:
Removing the old constraint is still an open issue #967

PR #965, opened 2024-06-24
Fix duplicate asset creation in itest

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/691.

Depends on https://github.com/lightninglabs/taproot-assets/pull/954.

It turns out that in one of our itests (`multi_address`), because we do an address self-transfer, we sometimes created duplicate asset entries, if the timing was unfortunate.
Basically both the send logic (freighter) as well as the address receive logic (custodian) inserted the same asset (one materializing it from the transfer output, the other from the proof found in the DB).


gijswijs (commented):
This PR touches `ApplyPendingOutput` which creates new assets. We still need to address Oli's comment about not reusing lock_time and relative_lock_time from the spent asset. See: https://github.com/lightninglabs/taproot-assets/blob/daf9676006b009624ea4b2910c8557d68a47c2e3/tapdb/assets_store.go#L2781-L2784

Although not relevant for this specific PR, I wanted to raise it here for future reference.

tapdb/assets_common.go
```diff
@@ -71,9 +71,10 @@ type UpsertAssetStore interface {
 	QueryAssets(context.Context, QueryAssetFilters) ([]ConfirmedAsset,
 		error)
 
-	// InsertNewAsset inserts a new asset on disk.
-	InsertNewAsset(ctx context.Context,
-		arg sqlc.InsertNewAssetParams) (int64, error)
+	// UpsertAsset inserts a new asset on disk, or returning the ID of the
```
gijswijs, line 74:
```suggestion
	// UpsertAsset inserts a new asset on disk, or returns the ID of the
```

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE UNIQUE INDEX assets_genesis_id_script_key_id_anchor_utxo_id_unique
```
gijswijs, line 1: Shouldn't we also drop the constraint `UNIQUE(asset_id, genesis_id, script_key_id)` ? Because that constraint makes no sense, and is still there: https://github.com/lightninglabs/taproot-assets/blob/dd8ba788757bd7d1b7c16ab0b97921283d4d0f5e/tapdb/sqlc/migrations/000002_assets.up.sql#L221

guggero: Unfortunately we can't easily remove that constraint, because it is not a named one (we should really avoid this in the future, exactly due to this reason).
The only way to get rid of this would be to create a completely new temporary asset table (without the constraint), copy the data, disable all foreign keys, drop the old table, rename the temporary table to the previous name, enable all foreign keys again.
We should probably do that at some point, but don't want to write such a script under time pressure.

gijswijs: Made an issue for this: #967

gijswijs:
lgtm :tada:

gijswijs (approved):

itest/addrs_test.go
```diff
@@ -649,6 +649,16 @@ func runMultiSendTest(ctxt context.Context, t *harnessTest, alice,
 		return nil
 	}, defaultTimeout/2)
 	require.NoError(t.t, err)
+
+	// We start out with two assets at Alice, one normal and one grouped. We
+	// then send it to Bob and to ourselves, with change being created. So
+	// in each round we turn one of the assets into 3 pieces (two
+	// self-transfers via addresses and one change output).
+	if runIdx == 0 {
```
jharveyb, line 657: super-nit: This could use the `runIdx` and `numRuns` values directly, as-is it would be incorrect for `runIdx > 1`.

But we don't call this func in that way for any existing tests, so no need to change right now.

itest/psbt_test.go
```diff
@@ -2719,15 +2734,29 @@ func testPsbtRelativeLockTimeSend(t *harnessTest) {
 		numTransfers, numOutputs,
 	)
 
+	// Make sure the transfer shows the lock time.
+	resp, err := bob.ListTransfers(ctxb, &taprpc.ListTransfersRequest{})
```
jharveyb, line 2738: 💯

jharveyb (approved):
LGTM, great catch 👍🏽

Roasbeef (approved):
LGTM 🦔

PR #968, opened 2024-06-24
tapdb: backup sqlite db before running database migration

ffranr:
This commit adds functionality to back up any existing SQLite database file before attempting a database migration.

The process involves creating a backup file, migrating the database, and removing the backup if no changes occurred to the 
database file. This approach avoids unnecessary backups but results in a backup being created and deleted each time tapd starts.

### DIFF
--- a/tapdb/migrations.go
+++ b/tapdb/migrations.go
@@ -26,20 +26,24 @@ const (
 )
 
 // MigrationTarget is a functional option that can be passed to applyMigrations
-// to specify a target version to migrate to.
-type MigrationTarget func(mig *migrate.Migrate) error
+// to specify a target version to migrate to. `currentDbVersion` is the current
+// (migration) version of the database, or None if unknown.
+// `maxMigrationVersion` is the maximum migration version known to the driver,
+// or None if unknown.
+type MigrationTarget func(mig *migrate.Migrate,
+	currentDbVersion int, maxMigrationVersion uint) error
 
 var (
 	// TargetLatest is a MigrationTarget that migrates to the latest
 	// version available.
-	TargetLatest = func(mig *migrate.Migrate) error {
+	TargetLatest = func(mig *migrate.Migrate, _ int, _ uint) error {
 		return mig.Up()
 	}
 
 	// TargetVersion is a MigrationTarget that migrates to the given
 	// version.
 	TargetVersion = func(version uint) MigrationTarget {
-		return func(mig *migrate.Migrate) error {
+		return func(mig *migrate.Migrate, _ int, _ uint) error {
 			return mig.Migrate(version)
 		}
 	}
@@ -146,16 +150,30 @@ func applyMigrations(fs fs.FS, driver database.Driver, path, dbName string,
 			ErrMigrationDowngrade, migrationVersion, latestVersion)
 	}
 
-	log.Infof("Applying migrations from version=%v", migrationVersion)
+	// Report the current version of the database before the migration.
+	currentDbVersion, _, err := driver.Version()
+	if err != nil {
+		return fmt.Errorf("unable to get current db version: %w", err)
+	}
+	log.Infof("Attempting to apply migration(s) "+
+		"(current_db_version=%v, latest_migration_version=%v)",
+		currentDbVersion, latestVersion)
 
 	// Apply our local logger to the migration instance.
 	sqlMigrate.Log = &migrationLogger{log}
 
 	// Execute the migration based on the target given.
-	err = targetVersion(sqlMigrate)
+	err = targetVersion(sqlMigrate, currentDbVersion, latestVersion)
 	if err != nil && !errors.Is(err, migrate.ErrNoChange) {
 		return err
 	}
+
+	// Report the current version of the database after the migration.
+	currentDbVersion, _, err = driver.Version()
+	if err != nil {
+		return fmt.Errorf("unable to get current db version: %w", err)
+	}
+	log.Infof("Database version after migration: %v", currentDbVersion)
 
 	return nil
 }
--- a/tapdb/migrations_test.go
+++ b/tapdb/migrations_test.go
@@ -2,6 +2,9 @@ package tapdb
 
 import (
 	"context"
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/btcsuite/btcd/wire"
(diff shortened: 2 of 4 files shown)


tapdb/sqlite.go
```diff
@@ -165,6 +203,48 @@ func (s *SqliteStore) ExecuteMigrations(target MigrationTarget) error {
...
+		return "", fmt.Errorf("error opening database file: %w", err)
+	}
+	defer srcFile.Close()
+
+	// Get the current time and format it as a Unix timestamp in
+	// nanoseconds.
+	timestamp := time.Now().UnixNano()
+
+	// Add the timestamp to the backup name.
+	backupFullFilePath := dbFullFilePath + "." + fmt.Sprint(timestamp) +
```
guggero, line 226: nit: `fmt.Sprintf("%s.%d.backup", dbFullFilePath, timestamp)`?

tapdb/sqlite.go
```diff
@@ -140,7 +144,41 @@ func NewSqliteStore(cfg *SqliteConfig) (*SqliteStore, error) {
 	// Now that the database is open, populate the database with our set of
 	// schemas based on our embedded in-memory file system.
 	if !cfg.SkipMigrations {
-		if err := s.ExecuteMigrations(TargetLatest); err != nil {
+		backupAndMigrate := func(mig *migrate.Migrate) error {
+			// Create a backup of the database before starting the
+			// migration.
+			backupFullFilePath, err := backupSqliteDatabase(
```
guggero, line 150: Maybe we should add a flag to disable this? Or specify the backup location? Thinking about universes with very large databases where we would now require the available disk space to at least be double.
Though I guess/hope those universes might not be running SQLite...

tapdb/sqlite.go
```diff
@@ -140,7 +144,41 @@ func NewSqliteStore(cfg *SqliteConfig) (*SqliteStore, error) {
 	// Now that the database is open, populate the database with our set of
 	// schemas based on our embedded in-memory file system.
 	if !cfg.SkipMigrations {
-		if err := s.ExecuteMigrations(TargetLatest); err != nil {
+		backupAndMigrate := func(mig *migrate.Migrate) error {
```
guggero, line 147: Should we keep the mechanism with the `MigrationTarget`? So we can do a backup while also being able to target a specific version? Would be handy for unit tests.

guggero (commented):
Very nice idea!

tapdb/sqlite.go
ffranr: Added flag.

tapdb/sqlite.go
ffranr: Hard for a skip backup flag to reach that point.

ffranr:
I think this PR should be ready for reviews and merge now. I got rid of the big downside risk I was worried about (naively copying the database file). We now execute a `VACUUM INTO` query from the source db. See https://sqlite.org/lang_vacuum.html . This is a documented sqlite database backup strategy.

tapdb/sqlite.go
```diff
@@ -149,6 +157,87 @@ func NewSqliteStore(cfg *SqliteConfig) (*SqliteStore, error) {
...
+
+	// Create a database backup file full path from the given source
+	// database full file path.
+	//
+	// Get the current time and format it as a Unix timestamp in
+	// nanoseconds.
+	timestamp := time.Now().UnixNano()
+
+	// Add the timestamp to the backup name.
+	backupFullFilePath := fmt.Sprintf(
```
jharveyb, line 174: From https://sqlite.org/lang_vacuum.html

'The file named by the INTO clause must not previously exist, or else it must be an empty file, or the VACUUM INTO command will fail with an error.'

So we should check that the path isn't non-empty before we try the backup, and make an empty file there before the actual VACUUM.

jharveyb (approved):
Looks good & great feature!

IIUC this won't delete any existing backups right? I think that's the correct behavior for now but just checking.

I imagine a similar feature is available for Postgres but can be deferred to another PR.

tapdb/migrations.go
```diff
@@ -109,6 +120,31 @@ func (m *migrationLogger) Verbose() bool {
 	return m.log.Level() <= btclog.LevelDebug
 }
 
+// maxKnownMigrationVersion returns the maximum migration version known to the
+// given source driver.
+func maxKnownMigrationVersion(sourceDriver source.Driver) (fn.Option[uint],
```
guggero, line 125: If `fn.None[uint](), nil` is never returned, why do we even need the option here? What's wrong with the well-established `return 0, err` pattern?
IMO we should only use the option type (which decreases readability) when `<none>, nil` is actually an option as a return value in a function.

tapdb/migrations.go
```diff
@@ -146,17 +182,43 @@ func applyMigrations(fs fs.FS, driver database.Driver, path, dbName string,
 			ErrMigrationDowngrade, migrationVersion, latestVersion)
 	}
 
-	log.Infof("Applying migrations from version=%v", migrationVersion)
+	// Compute the max migration version known to the driver.
+	maxMigrationVersion, err := maxKnownMigrationVersion(migrateFileServer)
```
guggero, line 186: Can't we just use the `LatestMigrationVersion` constant that is checked to be up to date with a CI script?

tapdb/sqlite.go
```diff
@@ -149,6 +157,87 @@ func NewSqliteStore(cfg *SqliteConfig) (*SqliteStore, error) {
...
+		"%s.%d.backup", dbFullFilePath, timestamp,
+	)
+
+	log.Infof("Creating backup of database file: %v -> %v",
+		dbFullFilePath, backupFullFilePath)
+
+	// Backup the database
```
guggero, line 181: nit: incomplete comment.

tapdb/sqlite.go
```diff
@@ -149,6 +157,87 @@ func NewSqliteStore(cfg *SqliteConfig) (*SqliteStore, error) {
...
+	vacuumIntoQuery := "VACUUM INTO ?;"
```
guggero, line 182: Wow, nice. TIL about `VACUUM INTO`...

tapdb/sqlite.go
```diff
@@ -53,6 +56,11 @@ type SqliteConfig struct {
 	// up if they don't already exist.
 	SkipMigrations bool `long:"skipmigrations" description:"Skip applying migrations on startup."`
 
+	// SkipMigrationDbBackup if true, then a backup of the database will not
+	// be created before applying migrations.
+	// nolint: lll
```
guggero, line 61: nit: move `nolint` comment to whole `SqliteConfig` struct?

tapdb/migrations.go
```diff
@@ -146,17 +182,43 @@ func applyMigrations(fs fs.FS, driver database.Driver, path, dbName string,
...
+	currentDbVersion, _, err := driver.Version()
+	if err != nil {
+		return fmt.Errorf("unable to get current db version: %w", err)
+	}
+	log.Infof("Attempting to apply migration(s) "+
+		"(current_db_version=%v, max_migration_version=%v)",
+		currentDbVersion, maxMigrationVersion)
+
+	var currentDbV fn.Option[uint]
+	if currentDbVersion >= 0 {
```
guggero, line 202: Isn't this always >= 0, since it's an uint?

tapdb/sqlite.go
```diff
@@ -149,6 +157,87 @@ func NewSqliteStore(cfg *SqliteConfig) (*SqliteStore, error) {
...
+	}
+
+	return nil
+}
+
+// backupAndMigrate is a helper function that creates a database backup before
+// initiating the migration, and then migrates the database to the latest
+// version.
+func (s *SqliteStore) backupAndMigrate(mig *migrate.Migrate,
+	currentDbVersion fn.Option[uint],
```
guggero, line 201: I think we can get rid of the options here. If no migration has ever been applied, `currentDbVersion` will just be 0 and `maxMigrationVersion` can be read from a constant. Should make the code a bit easier over all.

tapdb/sqlite.go
```diff
@@ -149,6 +157,87 @@ func NewSqliteStore(cfg *SqliteConfig) (*SqliteStore, error) {
...
+	versionUpgradePending := false
+	currentDbVersion.WhenSome(func(currentV uint) {
+		maxMigrationVersion.WhenSome(func(maxV uint) {
+			versionUpgradePending = currentV < maxV
+		})
+	})
+
+	// If a database migration version upgrade is pending, create a backup
+	// of the database before starting the migration.
+	if versionUpgradePending && !s.cfg.SkipMigrationDbBackup {
```
guggero, line 216: nit: could make this into a 3-way switch statement to distinguish between `versionUpgradePending && !s.cfg.SkipMigrationDbBackup`,  `versionUpgradePending && s.cfg.SkipMigrationDbBackup` and `!versionUpgradePending`, at least for the log message.

tapdb/sqlite.go
```diff
@@ -149,6 +157,87 @@ func NewSqliteStore(cfg *SqliteConfig) (*SqliteStore, error) {
...
+			return err
+		}
+	} else {
+		log.Infof("Skipping database backup before applying migrations")
+	}
+
+	// Migration to the target version.
+	err := mig.Up()
+	switch {
+	case errors.Is(err, migrate.ErrNoChange):
```
guggero, line 230: The `ErrNoChange` error is already handled by `applyMigrations`, so we can just do `return mig.Up()`.

tapdb/sqlite.go
ffranr: I read that as: if a file does exist, it must be empty, otherwise it will fail with an error.

If a file does not exist at the backup path at all, then VACUUM INTO will just create one. I don't think we need to add an empty file as a sort of placeholder. And I think we can skip checking for an existing file because of the nanosecond timestamp included in the file name.

tapdb/migrations_test.go
```diff
@@ -130,3 +133,89 @@ func TestMigrationDowngrade(t *testing.T) {
...
+
+	db := NewTestSqliteDBWithVersion(t, 14)
+
+	// We need to insert some test data that will be affected by the
+	// migration number 15.
+	InsertTestdata(t, db.BaseDB, "migrations_test_00015_dummy_data.sql")
+
+	// And now that we have test data inserted, we can create a backup and
+	// migrate to the latest version.
+	err := db.ExecuteMigrations(db.backupAndMigrate)
```
guggero, line 180: Nice test!

tapdb/migrations_test.go
```diff
@@ -130,3 +133,89 @@ func TestMigrationDowngrade(t *testing.T) {
...
 	require.ErrorIs(t, err, ErrMigrationDowngrade)
 }
+
+// findDbBackupFilePath walks the directory of the given database file path and
+// returns the path to the backup file.
+func findDbBackupFilePath(t *testing.T, dbFilePath string) string {
+	var dbBackupFilePath string
+	dir := filepath.Dir(dbFilePath)
+
+	err := filepath.Walk(dir,
```
guggero, line 143: nit: formatting.

guggero (commented):
Great idea!

I think with the latest update that disallows downgrades, we can simplify some things here. But really like the approach and test!

tapdb/sqlite.go
jharveyb: I read it the same way, should have tagged it with a nit. As in, the timestamp should indeed protect us from a file conflict.

Was suggesting the placeholder as a way to catch some errors before we call VACUUM vs. have SQLite err and possibly return less useful information on err.

Non-blocking.

tapdb/migrations.go
ffranr: Good idea. Latest commit use `latestVersion` from
```
latestVersion := opts.latestVersion.UnwrapOr(LatestMigrationVersion)
```
which is basically the same as what you've suggested, I think `opts.latestVersion` is just there for tests. But I thought I'd point that out just in case I was missing something.

tapdb/migrations.go
ffranr: I've removed `maxKnownMigrationVersion`. But I'll keep your `return 0, err` pattern comment in mind for next time.

tapdb/sqlite.go
ffranr: I've revised this function. More logical now IMO. Let me know what you think.

tapdb/sqlite.go
ffranr: > Was suggesting the placeholder as a way to catch some errors before we call VACUUM vs. have SQLite err and possibly return less useful information on err.

I'm not sure what error info that we might catch with a placeholder that we wouldn't have otherwise seen. Do you know of any? I appreciate that this is a nit now. Just curious.

guggero (approved):
Nice, LGTM :tada:

Issue #964, opened 2024-06-24
Asset change from channel open not accessible


ffranr:
This bug issue follows from a report made here: https://github.com/lightninglabs/taproot-assets/issues/942

Jamal report (see https://github.com/lightninglabs/taproot-assets/issues/942#issuecomment-2176663481):

There's an issue which I previously reported in chat that I forgot to mention here. 

After opening a channel with a capacity that's lower than the node's total asset balance, the remaining balance becomes inaccessible. `tapcli assets list` returns no assets and `tapcli assets balance` shows the asset with the pre-channel balance.

<details>
<summary>Expand for reproduction of the issue via CLI</summary>

```sh
litd@alice:/$ litcli --version
litcli version 0.13.0-alpha commit=v0.13.0-alpha-9-g980ee938aa1e7a45633216aff8775ca8d05c12aa
litd@alice:/$ tapcli --version
tapcli version 0.3.2-alpha commit=v0.3.3-0.20240609154533-2b3c60ae77a9
litd@alice:/$ tapcli assets mint --name PBUX --type normal --supply 1000
{
    "pending_batch":  {
        "batch_key":  "03f44cb28755bb1069a3b44d4af3cc49bbe9e433d56bd1aa05d49efbc997e91aa3",
        "batch_txid":  "",
        "state":  "BATCH_STATE_PENDING",
        "assets":  [
            {
                "asset_version":  "ASSET_VERSION_V0",
                "asset_type":  "NORMAL",
                "name":  "PBUX",
                "asset_meta":  null,
                "amount":  "1000",
                "new_grouped_asset":  false,
                "group_key":  "",
                "group_anchor":  "",
                "group_internal_key":  null,
                "group_tapscript_root":  "",
                "script_key":  {
                    "pub_key":  "3573758b2c7f60b6c6be2cf7f99d886c1d0d56a53da11e30a96eb926a63a3886",
                    "key_desc":  {
                        "raw_key_bytes":  "031a620843e9a4a8fdfa76bd27de4d118db6e2d45b23d408802d39057109e783e4",
                        "key_loc":  {
                            "key_family":  212,
                            "key_index":  0
                        }
                    },
                    "tap_tweak":  ""
                }
            }
        ],
        "created_at":  "1718732824",
        "height_hint":  107,
        "batch_psbt":  ""
    }
}
litd@alice:/$ tapcli assets mint finalize
{
    "batch":  {
[...]

Issue #645, opened 2023-11-02
[feature]: Extend reporting of transitory/limbo transaction states (CLI/RPC)


dstadulis:
When assets move into a transitory state (e.g. burning an asset), the daemon stops listing assets because it intends to remove them from the wallet's ownership.  There's then an apparent UX disparity between 1) what's being reported by tapd 2) the state that's included in blocks 3) The state that could confirm by a broadcasted transaction.  The daemon would benefit in terms of UX, and future functionality, by reporting these classes of transitory / limbo state.   The CLI should be intuitive to users in the scenario where the burn etc is in limbo while the daemon waits for the transaction to be included in a block.  

- User feedback has been given that transitory state not being available makes it confusing what's has or is about to happen
- **Important to payment-channel functionality, the daemon's architecture will benefit from being able to keep track of and report pending transaction**

## Deliverable
`tapd` should be enhanced to report transitory / in-flight / limbo transactions
lnd's `lncli pendingchannels` command could be a parallel:

https://github.com/lightningnetwork/lnd/blob/e6fbaafda4a9cfe08ccee5106aa6dc1da62bcebc/rpcserver.go#L3823-L3858

dstadulis:
this functionality might be already available

dstadulis:
In the state machine for _transfer_ there is some record keeping of the state of proofs, could be a starting point @jharveyb will follow up with more code references

jharveyb:
As an example for the transfer state machine (tapfreighter), we write info about the transfer before anchor TX broadcast:

https://github.com/lightninglabs/taproot-assets/blob/12d24398b8027b592fbf51d0ed17f4f21a3d30c0/tapfreighter/chain_porter.go#L1015

Which is later updated after receiver proofs are delivered:

https://github.com/lightninglabs/taproot-assets/blob/12d24398b8027b592fbf51d0ed17f4f21a3d30c0/tapfreighter/chain_porter.go#L744

It looks like we already expose these pending transfers:

https://github.com/lightninglabs/taproot-assets/blob/12d24398b8027b592fbf51d0ed17f4f21a3d30c0/rpcserver.go#L959

So we could maybe address this with doc updates, or decide to show those pending transfers outside of this `ListTransfers` call.

PR #969, opened 2024-06-24
Add field `UnconfirmedTransfers` to RPC endpoint `ListAssets`

ffranr:
Added a new field `UnconfirmedTransfers` to the `ListAssets` RPC endpoint response. The field is a count of unconfirmed transfers. This field helps users interpret asset list responses, especially when asset coins are unconfirmed.

This change will hopeful help users understand this sort of situation: https://github.com/lightninglabs/taproot-assets/issues/964


rpcserver.go
```diff
@@ -957,8 +957,25 @@ func (r *rpcServer) ListAssets(ctx context.Context,
 		return nil, err
 	}
 
+	// We will also report the number of unconfirmed transfers. This is
+	// useful for clients as unconfirmed asset coins are not included in the
+	// asset list.
+	transfersResp, err := r.ListTransfers(
```
guggero, line 963: By default, `ListTransfers` returns both confirmed and unconfirmed transfers.
So I think we should instead call into `r.cfg.AssetStore.QueryParcels(ctx, nil, true)` to get the unconfirmed transfers directly.

guggero (changes requested):
Very nice, should help a lot with the user experience.
Just need to make sure we only count actually unconfirmed transfers, not all of them.

dstadulis:
Existing issue that might be closed by this PR: https://github.com/lightninglabs/taproot-assets/issues/645

Would #969 support reporting `pendingchannels` in lnd?

ffranr:
> Existing issue that might be closed by this PR: #645
> 
> Would #969 support reporting `pendingchannels` in lnd?

I think pending channels count towards the unconfirmed transfers counter that's being added in this PR.

guggero (approved):
Nice, LGTM :tada:

gijswijs (commented):
Related to @guggero 's comment earlier (https://github.com/lightninglabs/taproot-assets/pull/969#discussion_r1652083113) I think that the Go Doc comment of `QueryParcels` is incorrect. 
https://github.com/lightninglabs/taproot-assets/blob/eeb8a5b1a7fa32fa0b40c6b04abfe7415853af50/tapdb/assets_store.go#L3020

It should probably read something like this:
```
// QueryParcels returns the set of confirmed and unconfirmed parcels or exclusively the unconfirmed parcels.
```

ffranr:
> Related to @guggero 's comment earlier ([#969 (comment)](https://github.com/lightninglabs/taproot-assets/pull/969#discussion_r1652083113)) I think that the Go Doc comment of `QueryParcels` is incorrect.
> 
> https://github.com/lightninglabs/taproot-assets/blob/eeb8a5b1a7fa32fa0b40c6b04abfe7415853af50/tapdb/assets_store.go#L3020
> 
> It should probably read something like this:
> 
> ```
> // QueryParcels returns the set of confirmed and unconfirmed parcels or exclusively the unconfirmed parcels.
> ```

Yes. But I think I'm using `QueryParcels` correctly in this PR.

gijswijs (approved):
LGTM! 🎉

PR #970, opened 2024-06-25
tapchannel: update lnd dep and fix bug in breach handling 

Roasbeef:
In this commit, we fix a bug in breach handling by ensuring we use the proper CSV delay when making the local commitment script key. Without this, we won't recognize an output as ours, and will fail to insert the justice transaction into the database.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1110,13 +1110,13 @@ func (s *Server) ResolveContract(req lnwallet.ResolutionReq,
 }
 
 // DeriveSweepAddr takes a set of inputs, and the change address we'd use to
-// sweep them, and maybe results an extra sweep output that we should add to
+// sweep them, and maybe results in an extra sweep output that we should add to
 // the sweeping transaction.
 func (s *Server) DeriveSweepAddr(inputs []input.Input,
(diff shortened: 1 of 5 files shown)


guggero (approved):
Nice, LGTM :tada:

guggero:
CI failure is just a missing `go mod tidy` run. Will fix after merging the `lnd` PR and updating to the final, merged commit.

server.go
```diff
@@ -1116,7 +1116,7 @@ func (s *Server) DeriveSweepAddr(inputs []input.Input,
 	change lnwallet.AddrWithKey) lfn.Result[sweep.SweepOutput] {
```
gijswijs, line 1116: ultranit: Go doc comment is slightly confusing to me. Shouldn't this read: "and maybe results in an extra sweep output" instead of "and maybe results an extra sweep output"? Same here: 
https://github.com/lightninglabs/taproot-assets/blob/a50c87ab4afade81ee75d59dfac64095479698c4/tapchannel/aux_sweeper.go#L1496-L1497

gijswijs (approved):
Nice, LGTM :tada:

PR #973, opened 2024-06-25
tapdb: implement basic migration downgrade protection

Roasbeef:
In this commit, we implement basic migration downgrade protection. This ensures that tapd will refuse to downgrade versions if a migration happened in the latest software version.

We achieved this by hard coding the latest version as a constant. Then when we go to migrate, we check this version against the latest version stored in the database. If the hard coded version is less than the latest version, then we conclude this is a downgrade attempt.


tapdb/migrations.go
```diff
@@ -12,6 +13,16 @@ import (
 	"github.com/golang-migrate/migrate/v4"
 	"github.com/golang-migrate/migrate/v4/database"
 	"github.com/golang-migrate/migrate/v4/source/httpfs"
+	"github.com/lightninglabs/taproot-assets/fn"
+)
+
+const (
+	// LatestMigrationVersion is the latest migration version of the
+	// database.  This is used to implement downgrade protection for hte
```
jharveyb, line 21: nit: typo, hte -> the

tapdb/migrations.go
```diff
@@ -34,6 +45,35 @@ var (
 	}
 )
 
+var (
+	// ErrMigrationDowngrade is returned when a database downgrade is detected.
```
jharveyb, line 49: line length over 80, fails linter now 🎉 

Not sure why this needs its own `var ( )` block either?

jharveyb (approved):
LGTM 👍🏽 

Confirmed that this failed in itest if I downgraded `LeatestMigrationVersion` to 20. `mint_batch_resume` failed as we restart a node there.

guggero:
Added a commit to check the hard coded version in the Golang code with a CI step.

guggero (approved):
tACK, LGTM :tada:

PR #975, opened 2024-06-25
tapchannel: use taproot overlay chan type for asset chans

Roasbeef:
With this change, we'll use explicit negotiation for the new asset chan
specific chan type.


guggero (approved):
LGTM pending merge of `lnd` PR.

gijswijs (approved):
The commit does what is says on the tin. No comments, let's go. :tada:

PR #978, opened 2024-06-26
docs: add `sample-tapd.conf`

Roasbeef:
Generated by Claude 3.5 Sonnet, should spot check accordingly.


guggero (approved):
Very nice!

Added some extra information and updated some values to look more similar to `lnd`'s sample conf, then also added the check script we have.

HannahMR:
It would be great to add a bit of guidance on when config needs to be in the tapd.conf file and when it needs to be in the lit.conf file.

Liongrass:
The only thing I found unclear about this configuration file was the RFQ section at the very bottom.

For example, does `use_mock_price_oracle_service_promise_to_not_use_on_mainnet` (lol) require a boolean or not?

And the next line, if I want to use the mock price oracle, do I also need to set `experimental.rfq.priceoracleaddress=`?

And finally I might benefit from having an example what a USD per sat rate would look like, especially since the configuration reads "cent per sat" but the description reads "USD per sat". 1 sat is currently 0.00061 USD, would I then enter 0.00061 below?

`experimental.rfq.mockoraclecentpersat=0.00061`

guggero:
`use_mock_price_oracle_service_promise_to_not_use_on_mainnet` is the value you set for the key `experimental.rfq.priceoracleaddress=`.
So:
```
experimental.rfq.priceoracleaddress=use_mock_price_oracle_service_promise_to_not_use_on_mainnet
```

And yes, it's that verbose on purpose.

PR #980, opened 2024-06-27
tapdb: remove duplicate assets before adding unique index

guggero:
This commit first removes duplicate assets that were mistakenly created due to self transfers in previous versions.
After removing the duplicates, the new unique index should be applied without issues.


tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
```diff
@@ -1,3 +1,43 @@
+-- First step is to make sure all duplicate assets are seen as spent, if their
+-- anchor transaction is spent. That way we can later also group on the spent
+-- column to remove duplicates.
+UPDATE assets
```
gijswijs, line 4: I'm not sure if I grok this first query. How are we setting all duplicate assets here as being spent? What is it in the CTE that selects for duplicates exclusively?

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
```diff
@@ -1,3 +1,43 @@
...
+                            JOIN chain_txns ct
+                                 ON mu.txn_id = ct.txn_id
+                            LEFT JOIN asset_transfer_inputs ati
+                                      ON ati.anchor_point = mu.outpoint
+                   WHERE a.spent = false
+                     AND ati.input_id IS NOT NULL);
+
+-- Now remove all asset witnesses for the assets we're going to remove in the
+-- next step.
+DELETE
```
gijswijs, line 19: Can we safely delete the witness of a duplicate asset? Are we sure the witness is also available for the asset that will remain? Otherwise we should UPDATE the foreign key of the asset witness.

gijswijs (commented):
There are four tables with a foreign key reference to `assets`:
- `asset_witnesses`
- `asset_proofs`
- `passive_assets`
- `addr_events`

This PR addresses `asset_witnesses` but ignores the other three tables. Are we sure we can safely do so?

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
guggero: Yes, I think so. We always insert the witness with the asset ID that was inserted in the previous step. So every asset should have a witness.

guggero:
> There are four tables with a foreign key reference to `assets`:
> 
>     * `asset_witnesses`
> 
>     * `asset_proofs`
> 
>     * `passive_assets`
> 
>     * `addr_events`
> 
> 
> This PR addresses `asset_witnesses` but ignores the other three tables. Are we sure we can safely do so?

Yeah, good point. In the example database I was given there were no entries in the `asset_proofs`, `passive_assets` and `addr_events` table. So I think due to how the other insert/upsert queries are structured, only the asset with the lowest ID would get an entry in those tables. So by keeping the asset with the lowest ID when removing duplicates should work.
But I'll look into some more safe queries with the help of LLMs.

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
gijswijs: Because the duplication happened during address self-transfer all assets should be considered spent in the first place?

guggero: No, the logic that updates the `spent` flag in the normal application flow will only do that for the "original" assets, not the duplicates. So in order for us to be able to properly de-duplicate the assets, we'll just make sure the `spent` flag is set correctly for all of them. Update the comment on the query as well.

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
guggero: Update the logic to make sure we fill any gaps for duplicates that for some reason didn't have a witness.

guggero:
@gijswijs I added some more involved logic to properly account for what to do with the `asset_witnesses` and `asset_proofs` table. Those should now be dealt with correctly, I also added test code for that.

That leaves the `passive_assets` and `addr_events` tables. I thought hard about the code that produced the duplicates and am 95% sure that we won't ever have entries in those two tables for the duplicate entries. And if for some reason there would be, then for those users the migration would fail and we could manually intervene.

tapdb/migrations_test.go
```diff
@@ -130,3 +132,64 @@ func TestMigrationDowngrade(t *testing.T) {
 	err := db.ExecuteMigrations(TargetLatest, WithLatestVersion(1))
 	require.ErrorIs(t, err, ErrMigrationDowngrade)
 }
+
+func TestMigration20(t *testing.T) {
```
ffranr, line 136: we should add a doc. It might describe on a high level which aspects of the version 20 migration we're testing.

tapdb/migrations_test.go
```diff
@@ -130,3 +132,64 @@ func TestMigrationDowngrade(t *testing.T) {
...
+	// We need to insert some test data that will be affected by the
+	// migration number 20.
+	InsertTestdata(t, db.BaseDB, "migrations_test_00020_dummy_data.sql")
+
+	// And now that we have test data inserted, we can migrate to the latest
+	// version.
+	err := db.ExecuteMigrations(TargetLatest)
+	require.NoError(t, err)
+
+	_, assetStore := newAssetStoreFromDB(db.BaseDB)
```
ffranr, line 150: At this point in the test, we are primarily verifying the state of the database after it has been loaded. We should add a comment to clarify this.

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
```diff
@@ -1,4 +1,107 @@
...
+WHERE asset_id IN (SELECT a.asset_id
+                   FROM assets a
+                            JOIN managed_utxos mu
+                                 ON a.anchor_utxo_id = mu.utxo_id
+                            JOIN chain_txns ct
+                                 ON mu.txn_id = ct.txn_id
+                            LEFT JOIN asset_transfer_inputs ati
+                                      ON ati.anchor_point = mu.outpoint
+                   WHERE a.spent = false
+                     AND ati.input_id IS NOT NULL);
```
ffranr, line 15: Please correct me if my understanding is wrong: if `ati.input_id IS NOT NULL` then the asset is spent? And those are the rows that we ensure are set to spent, is that correct?

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
```diff
@@ -1,4 +1,107 @@
...
+                  AND a.amount = tmp.amount
+                  AND a.anchor_utxo_id = tmp.anchor_utxo_id
+                  AND a.spent = tmp.spent;
+
+-- Step 4: To make the next step possible, we need to disable a unique index on
+-- the asset_witnesses table. We'll re-create it later.
+DROP INDEX IF EXISTS asset_witnesses_asset_id_witness_index_unique;
+
+-- Step 5: Update the asset_witnesses and asset_proofs tables to reference the
+-- new asset_ids. But skip any records that already reference the new asset_id.
```
ffranr, line 46: I think I would be more confident in this step if we didn't do "skip any records that already reference the new asset_id", which I think is unnecessary. Can't we just set again, value shouldn't change.

ffranr (approved):

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
guggero: Correct. We `LEFT JOIN` here, meaning that any field on `ati` is `NULL` if there is no matching spending transaction referencing the asset's on-chain outpoint.

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
guggero: Hmm, you're right. Since we have disabled/removed the unique constraint on the witness table, we can remove the `AND NOT EXISTS` for that table, since it doesn't matter if we create any collisions.

Unfortunately the `asset_proofs` table has an anonymous/inline `UNIQUE` constraint that we cannot disable or remove. So we need to keep that `AND NOT EXISTS` to skip re-assigning proofs if the asset we're going to keep already has one. Otherwise we run into a unique key constraint violation error.

tapdb/testdata/migrations_test_00020_dummy_data.sql
```diff
@@ -0,0 +1,50 @@
...
+
+INSERT INTO asset_witnesses VALUES(1,1,X'a1594fc379308b2a209f6d0bdb8602e9f87cf71fc232c69032b9a5fed28f933100000000',X'add7d0d7cc37e58a7c0d8ad40b6904050d2baa25a1829f00689c4b27b524dd04',X'02827d74858d152da1fae12010ad8d3c46b595c2d4480512a6575925424617124f',X'0101aa',NULL, 0);
+INSERT INTO asset_witnesses VALUES(2,1,X'a1594fc379308b2a209f6d0bdb8602e9f87cf71fc232c69032b9a5fed28f933101000000',X'add7d0d7cc37e58a7c0d8ad40b6904050d2baa25a1829f00689c4b27b524dd04',X'02827d74858d152da1fae12010ad8d3c46b595c2d4480512a6575925424617124f',X'0101bb',NULL, 1);
+
+INSERT INTO asset_witnesses VALUES(3,2,X'a1594fc379308b2a209f6d0bdb8602e9f87cf71fc232c69032b9a5fed28f933100000000',X'add7d0d7cc37e58a7c0d8ad40b6904050d2baa25a1829f00689c4b27b524dd04',X'02827d74858d152da1fae12010ad8d3c46b595c2d4480512a6575925424617124f',X'0101cc',NULL, 0);
+INSERT INTO asset_witnesses VALUES(4,2,X'a1594fc379308b2a209f6d0bdb8602e9f87cf71fc232c69032b9a5fed28f933101000000',X'add7d0d7cc37e58a7c0d8ad40b6904050d2baa25a1829f00689c4b27b524dd04',X'02827d74858d152da1fae12010ad8d3c46b595c2d4480512a6575925424617124f',X'0101dd',NULL, 1);
+
+INSERT INTO asset_proofs VALUES(1,1,X'aaaa');
+
+INSERT INTO asset_proofs VALUES(2,2,X'cccc');
```
gijswijs, line 32: Maybe insert only a single proof, linked to the asset we will delete. That way we test the replacement of old ID with new ID with regards to update query for `asset_proofs`

```suggestion
INSERT INTO asset_proofs VALUES(2,2,X'aaaa');
```

gijswijs (commented):
Very nice!! This should fix a lot of the cases of duplication. I think we might run into one or two edge cases where manual intervention is still needed, but the majority (if not all) of it is now covered. I have some questions, a nit and one suggestion that I think will improve the test.

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
guggero: For the proofs, we have the `AND NOT EXISTS`. So it only moves proofs for assets where there is no proof already for `new_asset_id`. If I remove the `AND NOT EXISTS`, I get a `UNIQUE constraint failed: asset_proofs.asset_id` error.

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
```diff
@@ -1,4 +1,105 @@
...
+-- the asset_witnesses table. We'll re-create it later.
+DROP INDEX IF EXISTS asset_witnesses_asset_id_witness_index_unique;
+
+-- Step 5: Update the asset_witnesses and asset_proofs tables to reference the
+-- new asset_ids.
+UPDATE asset_witnesses
+SET asset_id = (SELECT new_asset_id
+                FROM tmp_asset_id_mapping
+                WHERE old_asset_id = asset_witnesses.asset_id)
+WHERE EXISTS (SELECT 1
```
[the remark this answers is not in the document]
guggero: Hmm, that's way simpler, thanks. Seems to be supported by both engines too.

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
```diff
@@ -1,4 +1,105 @@
...
+                        asset_id, witness_index
+        );
+
+-- Step 8: Delete any duplicate proofs.
+DELETE
+FROM asset_proofs
+WHERE asset_id NOT IN (SELECT min_asset_id FROM tmp_min_assets);
+
+-- Step 9: Delete the duplicates from the assets table. This will then also
+-- delete dangling asset_witnesses, since we have a ON DELETE CASCADE clause on
```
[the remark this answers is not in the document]
guggero: No, since we update all witnesses to point to the asset that we're going to keep, removing the duplicate assets won't remove the now duplicate witnesses. I guess I don't even have to mention DELETE CASCADE here...

tapdb/testdata/migrations_test_00020_dummy_data.sql
guggero: Hmm, you're right. And doing that showed that the query for the proofs isn't correct.
So we need a better query...

gijswijs: I've pushed a commit with a better query.

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
```diff
@@ -1,4 +1,99 @@
...
+DELETE
+FROM asset_witnesses
+WHERE witness_id NOT IN (SELECT min(witness_id)
+                         FROM asset_witnesses
+                         GROUP BY asset_id, witness_index);
+
+-- Step 7: Re-enable the unique index on the asset_witnesses table.
+CREATE UNIQUE INDEX asset_witnesses_asset_id_witness_index_unique
+    ON asset_witnesses (
+                        asset_id, witness_index
```
jharveyb, line 76: nit: over-indented.

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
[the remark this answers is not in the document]
jharveyb: Would be good to remove that mention.

jharveyb (commented):
Looks good, just some lingering Qs on query specifics and explanatory comments.

tapdb/sqlc/migrations/000020_asset_unique_key.up.sql
```diff
@@ -1,4 +1,99 @@
...
+WHERE asset_witnesses.asset_id = tmp_asset_id_mapping.old_asset_id;
+
+-- For the proofs we need skip re-assigning them to the asset that we're going
+-- to keep if it already has a proof. This is because the unique index on the
+-- asset_proofs table would prevent us from doing so. And we can't disable the
+-- unique index, because it is an unnamed/inline index.
+UPDATE asset_proofs
+SET asset_id = filtered_mapping.new_asset_id
+FROM (                  
+        SELECT MIN(old_asset_id) AS old_asset_id, new_asset_id
```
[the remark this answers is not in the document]
gijswijs: `asset_id` in table `asset_proofs` has a `UNIQUE` constraint. So if we map `old_asset_id` a and b both to `new_asset_id` x we violate that constraint. So by grouping on `new_asset_id` and taking `MIN(old_asset_id)` we only map a single `old_asset_id` to the `new_asset_id`. (a to x in this case) The remaining asset_proof with foreign key b will be deleted in step 8.

The `filtered_mapping` table does the exact same thing as `tmp_asset_id_mapping` but taking into account the `UNIQUE` constraint on `asset_id` in table `asset_proofs`. Hence I reused the column names.

> So the end result would be that, for duplicate assets, if the proof was not already using the new_asset_id (which is the smaller asset ID), asset_proofs.asset_id would be updated to use that new_asset_id.

Correct.

> And any proof already referencing that new_asset_id should be unchanged.

It will get updated, but with the same values, so the result is unchanged.

jharveyb (approved):
LGTM, great fix and should be safe 👍🏽

PR #979, opened 2024-06-26
taprpc: Set dec display when metadata is empty

jharveyb:
Resolves #972 by setting the dec display as the only JSON object when the given metadata is empty. The RPC request that causes this behavior is tested in the meta itest, and I manually tested the CLI for:

- Non-zero dec display, empty metadata, JSON meta type -> correct `taprpc.AssetMeta` struct is passed to the RPC, asset listed correctly after minting (this was with the exact CLI cmd Leo posted).
- zero dec display, empty metadata, JSON meta type -> rejected for empty metadata
- unknown meta type, empty metadata -> rejected for empty metadata (IIUC this was the existing behavior)

PR #983, opened 2024-06-27
taprpc: decimal display requires JSON meta

jharveyb:
Follow-on to #979 .

For minting via RPC, there was a missing check to prevent the decimal display value from being dropped before reaching the minter.

### DIFF
--- a/itest/asset_meta_test.go
+++ b/itest/asset_meta_test.go
@@ -172,11 +172,20 @@ func testMintAssetWithDecimalDisplayMetaField(t *harnessTest) {
 	_, err = t.tapd.MintAsset(ctxt, secondAssetReq)
 	require.ErrorContains(t.t, err, "decimal display does not match")
 
+	// Requesting a decimal display without specifying the metadata type as
+	// JSON should fail.
+	secondAssetReq.Asset.DecimalDisplay = firstAsset.DecimalDisplay
+	secondAssetReq.Asset.AssetMeta = nil
+
+	_, err = t.tapd.MintAsset(ctxt, secondAssetReq)
+	require.ErrorContains(t.t, err, "decimal display requires JSON")
+
 	// If we update the decimal display to match the group anchor, minting
 	// should succeed. We also unset the metadata to ensure that the decimal
 	// display is set as the sole JSON object if needed.
-	secondAssetReq.Asset.AssetMeta.Data = []byte{}
-	secondAssetReq.Asset.DecimalDisplay = firstAsset.DecimalDisplay
+	secondAssetReq.Asset.AssetMeta = &taprpc.AssetMeta{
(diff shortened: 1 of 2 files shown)


itest/asset_meta_test.go
```diff
@@ -172,11 +172,20 @@ func testMintAssetWithDecimalDisplayMetaField(t *harnessTest) {
 	_, err = t.tapd.MintAsset(ctxt, secondAssetReq)
 	require.ErrorContains(t.t, err, "decimal display does not match")
 
+	// Requesting a decimal display without specifying the metadata type as
+	// JSON should fail.
+	secondAssetReq.Asset.DecimalDisplay = firstAsset.DecimalDisplay
+	secondAssetReq.Asset.AssetMeta = nil
+
+	_, err = t.tapd.MintAsset(ctxt, secondAssetReq)
+	require.ErrorContains(t.t, err, "decimal display requires JSON")
```
GeorgeTsagk, line 181: 👍 nice, good to check the bad path too

rpcserver.go
```diff
@@ -505,6 +505,14 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 	}
 
 	var seedlingMeta *proof.MetaReveal
+
+	// If a custom decimal display is set, the meta type must also be set to
+	// JSON.
+	if req.Asset.DecimalDisplay != 0 && req.Asset.AssetMeta == nil {
+		return nil, fmt.Errorf("decimal display requires JSON asset " +
```
GeorgeTsagk, line 512: Might be better to instead bootstrap the meta json with just the decimal display field? otherwise a user will have to provide an empty json `{}` for this to work, which might not be intuitive

GeorgeTsagk (approved):

rpcserver.go
ffranr: I agree with setting to an empty meta json if the meta field nil in the request but a decimal display is given.

ffranr (approved):
I think we should do this: https://github.com/lightninglabs/taproot-assets/pull/983/commits/5fa519d92163dd049d77070d0f28fb6d9e9b4e55#r1658411119

And then should be good to go.

rpcserver.go
jharveyb: The `AssetMeta` field here is both `Data` and `Type`. So we require `Type` to be set t JSON for decimal display to get set as the only JSON field if the metadata data field was empty.

Was implemented in https://github.com/lightninglabs/taproot-assets/pull/979 , seen here:

https://github.com/lightninglabs/taproot-assets/blob/311231d67693983b78f48edee1f9fac60efe032d/rpcserver.go#L533

This check is added before that bootstrapping to enforce that the user at least set the correct metadata type.

PR #984, opened 2024-06-28
tapdb: add anchor point to SetSpent, remove LIMIT 1

guggero:
Discovered while adding more assertions in the `litd` custom channel itests.

This commit fixes two things:
 - Because we could have multiple assets with the same script key, but in different anchor transactions, we also need to include the anchor outpoint of the input we want to mark as spent. Otherwise a duplicate script key (e.g. a second TAP channel funding script key, which is identical for every channel as it's just OP_TRUE) couldn't be set as spent anymore.
 - Remove the TODO and LIMIT 1 workaround, since we now have a proper unique index preventing duplicate script keys _WITHIN THE SAME ANCHOR OUTPUT_.

### DIFF
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -2727,8 +2727,9 @@ func (a *AssetStore) ConfirmParcelDelivery(ctx context.Context,
 		for idx := range inputs {
 			spentAssetIDs[idx], err = q.SetAssetSpent(
 				ctx, SetAssetSpentParams{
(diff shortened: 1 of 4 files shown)


Roasbeef (approved):
LGTM 💃🏾

jharveyb (approved):
Also LGTM :thumb:

Scoping by outpoint is probably going to useful in other spots where script keys can be duplicated / reused.

Issue #985, opened 2024-06-28
[bug]: responder of taproot assets channel doesn't have proof for commitment transaction


guggero:
Found while adding more assertions to the custom channel integration tests in `litd`: https://github.com/lightninglabs/lightning-terminal/pull/784

If the responder of a channel initiates the co-op close process, they seem to not have the proofs for the commitment transaction available in the local proof archive.
This results in the following error:

```
2024-06-28 17:50:31.085 [ERR] FRTR: Error evaluating state (SendStateStoreProofs): failed to generate an updated proof file for output 0: error fetching input proof: error fetching input proof -- locator=(proof.Locator) {
 AssetID: (*asset.ID)(0xc001034fe4)((len=32 cap=32) 452d67bfd17372691f86c3410ec836fae677f5a2119097c50680d463a2c58b1f),
 GroupKey: (*secp256k1.PublicKey)(<nil>),
 ScriptKey: (secp256k1.PublicKey) {
  x: (secp256k1.FieldVal) 50aaeb166f4234650d84a2d8a130987aeaf6950206e0905401ee74ff3f8d18e6,
  y: (secp256k1.FieldVal) 8e5b090f56a33ca5081a001b402f774ebd0690abed9273a47dec5c3ae92d9786
 },
 OutPoint: (*wire.OutPoint)(0xc001034fc0)(c26984a76cec3f420e59cef8f8faa8057c5f382cad999b7c9861219095d7e33b:0)
}
: unable to find proof
```

That's why the transfer then never fully completes and the final balance isn't correct.

Roasbeef:
I think we need to call `auxSweeper.importOutputProof` when the co-op close process starts. Better yet, with a slightly larger change, we can hook into the `chanvalidate` module for lnd, and make sure that once the funding transaction is confirmed, all the proofs are imported.

PR #987, opened 2024-07-02
tapchannel: import funding output proofs for responder during co-op close

Roasbeef:
Otherwise, the transfer insert for the responder may fail as they don't
have the funding output in their database, so shipping the transfer
fails as it can't find the proof.

Fixes https://github.com/lightninglabs/taproot-assets/issues/985.


jharveyb:
Post-merge LGTM 👍🏽

Issue #997, opened 2024-07-09
[feature]: Implement "Asset TLV Placeholder" Design for Decimal Display


dstadulis:
## Context

Given the scope of https://github.com/lightninglabs/taproot-assets/issues/956 is board, @roasbeef proposed an alternative strategy to achieve, eventual, protocol-level enforcement of an asset decimal display TLV

## Description

As proposed by Roasbeef, this issue aims to specify an unserialized asset TLV (Type-Length-Value) placeholder that will reserve space for decimal display without fully delivering the feature. This design will ensure that the TLV is bound to the asset while allowing for future development and interpretation of the stored opaque blob.

#### Tasks to complete

- [ ] Bump version: Increment the version number to reflect the changes.
- [ ] Opaque blob storage: Store an opaque blob in the asset, which will be interpreted later.
- [ ] Old client compatibility: Ensure that old clients can recognize and serialize the asset using normal asset serialization methods, even if they don't understand the new TLV placeholder.

## Inspiration

This implementation should draw inspiration from how Lnd wiremessage reads past the point of concrete values.

### Deliverables

- [ ] Unserialized asset TLV placeholder that reserves space for decimal display
- [ ] Version number increment
- [ ] Opaque blob storage and later interpretation mechanism
- [ ] Compatibility with old clients

**Describe alternatives you've considered**
If decimal display is solely specified via JSON then there will likely be on going tech debt necessary to support it and alternative taproot asset implementations could fault.  While decimal display isn't protocol state machine consensus critical, it does have a drastic effect on the human interpreted values.

Issue #956, opened 2024-06-18
[enhancement] Migrate `Decimal Display` into TLV to induce availability and protocol-level enforcement


dstadulis:
## Description:

To ensuring correct decimal display usage the daemon has recently added checks inside the planter (added in #952) this should catch inadvertent user errors (if a user is using a tapd daemon).  This is substandard as the protocol-design need is to entwine the value with the endogenous properties of an asset -- e.g. inextricably link `decimal_display` into the asset.

## Proposed Solution:

Move the decimal display value into the asset's TLV. This would make it an integral part of the asset's property and enforce on the protocol level consistency.

## Benefits

1. This will enforce consistency decimal display value across any taproot asset implementation (or hacky uses of tapd) and light client by making it inherent to the asset

**Implementation Plan

To implement this solution, we will:

- [ ] Update the TLV to include decimal display information.
- [ ] Modify the RPC/API interfaces to accommodate the updated TLV structure.
- [ ] Refactor the verification logic to utilize the TLV-based decimal display.

dstadulis:
Moving this issues out of v0.4 milestone in favor of https://github.com/lightninglabs/taproot-assets/issues/997

dstadulis:
Removing from Sprint 30
Available to be reassigned if others want to take it from @guggero

PR #998, opened 2024-07-09
proof: add decimal display as TLV field to meta reveal

guggero:
Minimal implementation that fixes https://github.com/lightninglabs/taproot-assets/issues/956.
Also fixes https://github.com/lightninglabs/taproot-assets/issues/997.


proof/meta.go
```diff
@@ -66,26 +66,9 @@ var (
...
-	// invalid.
-	ErrDecDisplayInvalid = errors.New("invalid decimal display")
-
 	// ErrDecDisplayTooLarge is returned if the decimal display value
 	// exceeds the current limit.
 	ErrDecDisplayTooLarge = errors.New("decimal display too large")
-
-	// ErrDecDisplayInvalidType is returned if the value in a JSON object
-	// assigned to the decimal display key is not a uint32.
-	ErrDecDisplayInvalidType = errors.New("decimal display JSON field is " +
```
Roasbeef, line 83: Nice to get rid of all these error cases that no longer apply!

proof/meta_test.go
```diff
@@ -59,6 +62,24 @@ func TestValidateMetaReveal(t *testing.T) {
 			},
 			expectedErr: nil,
 		},
+		{
+			name: "invalid decimal display",
+			reveal: &MetaReveal{
+				Type:           MetaJson,
+				Data:           []byte(`{"key": "value"}`),
+				DecimalDisplay: 999,
```
Roasbeef, line 70: 👍

proof/meta_test.go
```diff
@@ -78,3 +99,137 @@ func TestValidateMetaReveal(t *testing.T) {
...
+			require.NoError(tt, err)
+
+			require.Equal(tt, tc.reveal, decoded)
+		})
+	}
+}
+
+// TestDecodeOldMetaReveal tests decoding an old MetaReveal encoding that does
+// not include the decimal display.
+func TestDecodeOldMetaReveal(t *testing.T) {
```
Roasbeef, line 188: Nice client compat tests 👍

Roasbeef (approved):
LGTM 💃🏾

rpcserver.go
```diff
@@ -552,15 +552,13 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 		// If a custom decimal display was requested, add that to the
 		// metadata and re-validate it.
 		if metaType == proof.MetaJson && req.Asset.DecimalDisplay != 0 {
```
jharveyb, line 554: I think this condition now needs to change and not require the `metaType` to be JSON?

itest/asset_meta_test.go
```diff
@@ -149,10 +149,10 @@ func testMintAssetWithDecimalDisplayMetaField(t *harnessTest) {
 	}
 
 	// Manually update the requested metadata and compute the expected hash.
-	updatedMeta, err := mintedMeta.SetDecDisplay(firstAsset.DecimalDisplay)
+	err := mintedMeta.SetDecDisplay(firstAsset.DecimalDisplay)
```
jharveyb, line 152: Should also cover non-JSON metadata to make sure the dec display gets passed through the minter (and then list assets) correctly.

jharveyb (commented):
Looks good overall! Def cleaner than what we had before.

Added comments + I think we want to add this field to the `AssetMeta` proto as well + pipe it through there? If it's actually going to be a TLV field.

https://github.com/lightninglabs/taproot-assets/blob/5236deae2e3b3c73610063062643c5f32ab10ca4/taprpc/taprootassets.proto#L163

jharveyb:
Also, I don't see the utility of adding this field as JSON in the metadata any more?

IIRC the actual user request was 'it shows up in RPC responses', and if its also discoverable from decoding proofs then it seems redundant. AFAICT we never read it after setting it.

Roasbeef:
> Also, I don't see the utility of adding this field as JSON in the metadata any more?

Good point, we can just leave it out there.

Roasbeef:
> I think we want to add this field to the AssetMeta proto as well 

Technically it's part of the the metal reveal, which contains the raw asset meta, and now these additional bytes. I guess it's useful though to have it there if people want to use `FetchAssetMeta` to query the decimal value.

jharveyb:
> > I think we want to add this field to the AssetMeta proto as well
> 
> Technically it's part of the the metal reveal, which contains the raw asset meta, and now these additional bytes. I guess it's useful though to have it there if people want to use `FetchAssetMeta` to query the decimal value.

Just realized it's now also input for `MetaReveal.MetaHash()` - so definitely 'consensus' if we're going to leave that as is.

Roasbeef (changes requested):
Discussed offline, and concluded that this approach was a bit too slick to be true:
  * Old clients won't know of this meta reveal field. 
  * They'll parse the meta reveal w/o the new field, then go to check against the meta hash.
  * That fails as they didn't include the portion they didn't know about.

proof/meta_test.go
```diff
@@ -59,6 +62,24 @@ func TestValidateMetaReveal(t *testing.T) {
...
+			},
+			expectedErr: ErrDecDisplayTooLarge,
+		},
+		{
+			name: "correct decimal display",
+			reveal: &MetaReveal{
+				Type:           MetaJson,
+				Data:           []byte(`{"key": "value"}`),
+				DecimalDisplay: 8,
+			},
```
GeorgeTsagk, line 80: could also check against Opaque data type and also with 0 dec display

GeorgeTsagk (commented):
So the purpose here was to drop this field to the `MetaReveal` level in order to support dec display for custom meta data?

guggero:
> Discussed offline, and concluded that this approach was a bit too slick to be true:
> 
>     * Old clients won't know of this meta reveal field.
> 
>     * They'll parse the meta reveal w/o the new field, then go to check against the meta hash.
> 
>     * That fails as they didn't include the portion they didn't know about.

Ah, yes, nice catch.

Adding a summary of my comments made offline here for reference:

I see two possibilities for something we could do now:
1. don't add the decimal display to the meta hash (e.g. don't commit to it, just put it there as informational piece of data). not super nice and probably opens the door for shenanigans
2. add a "feature bit" to the meta type that indicates the meta data is nested TLV. then old nodes just read in the meta data as a blob and arrive at the correct hash. and new nodes would see the feature bit and do a second round of TLV decode on the data, being able to read and interpret extra fields that we put in there (the "actual" raw meta data being one of the sub fields). should work but feels veeery hacky...

I guess doing the two-level TLV encoding would qualify as "carveout/placeholder for future fields" as described in the issue #997. it just feels very hacky if we need to describe it in the BIP as "two levels of TLV encoding, depending on a feature bit"... but happy to go ahead and implement it, if we think it's the way to go

rpcserver.go
guggero: Good point! Fixed this and a couple of other occurrences of similar checks.

guggero:
Okay, the current approach doesn't work and any quick hacks are too hacky. I'll implement the proper carve out to be able to do this properly in the next major version in an upcoming PR.

Issue #1001, opened 2024-07-09
[bug]: 2024-07-09 21:26:10.034 [ERR] RPCS: [/taprpc.TaprootAssets/ListAssets]: unable to marshal asset: unable to extract decimal display for asset_id=dc68dcc370f15f3c318cc947e0315ad849a75cfb3ad6ef177933a8873d7dfe25 :decimal display field missing


weltitob:
Cant really explain myself this error it occurs on version  lightninglabs/lightning-terminal:v0.13.2-alpha.rc2 when importing my .tapd folder and thus my assets: 2024-07-09 21:26:10.034 [ERR] RPCS: [/taprpc.TaprootAssets/ListAssets]: unable to marshal asset: unable to extract decimal display for asset_id=dc68dcc370f15f3c318cc947e0315ad849a75cfb3ad6ef177933a8873d7dfe25 :decimal display field missing

Roasbeef:
Looks like the culprit is here: https://github.com/lightninglabs/taproot-assets/blob/aa3d671f4bc4b7e14c44e4e172011d3740065bd3/proof/meta.go#L236-L240

Will have a fix up shortly.

Roasbeef:
This'll be fixed by https://github.com/lightninglabs/taproot-assets/pull/998

PR #1002, opened 2024-07-10
proof: relax parsing in `DecDisplayForAssetID`

Roasbeef:
If an old asset didn't have this field, then we can just assume that the value is zero, rather than error out.

Fixes https://github.com/lightninglabs/taproot-assets/issues/1001


jharveyb:
That return was used to enter this case:

https://github.com/lightninglabs/taproot-assets/blob/5236deae2e3b3c73610063062643c5f32ab10ca4/proof/meta.go#L284

Which would trigger adding the display to the JSON. So that should get updated as well.

jharveyb (approved):
LGTM 👌🏽

Issue #995, opened 2024-07-09
[feature]: MVP Price-Oracle Service Demonstration


dstadulis:
**Is your feature request related to a problem? Please describe.**

As a `potential edge node operator` I want to `understand how tapd's price oracle calls interact with external services` in order to `have a conceptual stepping stone for how to build my fully fledged price proffering system`

**Describe the solution you'd like**
- Deliverable:
	- Bare-minimum code to `solicit price information`, of a `currency pair`, from a `price feed source`
	- The context being in the future users may build systems which digest multiple types of pairs from numerous sources (internal or external)
- Estimated size
	- 20 line Go or Python of a rough implementation

ffranr:
I've added a basic example price oracle service here: https://gist.github.com/ffranr/9d18e8f122e25b4bb435365b30c31456

PR #1004, opened 2024-07-10
Add example basic price oracle service

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/995

This PR introduces a basic price oracle service example and ensures its compilation during our CI checks.


docs/examples/basic-price-oracle/main.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// getRateTick returns a rate tick for a given transaction type and subject
+// asset max amount.
+func getRateTick(transactionType oraclerpc.TransactionType,
+	subjectAssetMaxAmount uint64) oraclerpc.RateTick {
+
+	// Determine the rate based on the transaction type.
+	var rate uint64
+	if transactionType == oraclerpc.TransactionType_PURCHASE {
```
Roasbeef, line 52: Can we add more godoc details to the diff of the two transaction types? Perhaps with a brief walkthrough example? So something like: Alice has `X` and wants to pay an invoice, she queries the oracle for a quote with `transactionType` `Z`. 

On second thought, this sounds like something we should add as normal mark down docs, and also in the builder's guide.

docs/examples/basic-price-oracle/main.go
```diff
@@ -0,0 +1,163 @@
...
+// QueryRateTick queries the rate tick for a given transaction type, subject
+// asset, and payment asset. The rate tick is the exchange rate between the
+// subject asset and the payment asset.
+func (p *RpcPriceOracleServer) QueryRateTick(_ context.Context,
+	req *oraclerpc.QueryRateTickRequest) (
+	*oraclerpc.QueryRateTickResponse, error) {
+
+	// Ensure that the payment asset is BTC. We only support BTC as the
+	// payment asset in this example.
+	if !oraclerpc.IsAssetBtc(req.PaymentAsset) {
```
Roasbeef, line 86: Is my understanding correct that `PaymentAsset`is the asset that'll be sent _outgoing_? Though if I think about the send vs recv flows, that means in the send flow the `PaymentAsset` is BTC, but in the recv flow (receiving some other asset) it'll be that asset? 

Or is it that this is the asset the penultimate hop will be receiving?

Roasbeef (changes requested):

docs/examples/basic-price-oracle/main.go
ffranr: I conceptualize the price oracle service as follows: it does not know on whether an asset is inbound or outbound on a channel; its primary function is to return a price. This applies regardless of whether the node is attempting to formulate a rate tick hint price or a quote accept rate tick.

Hence, it is the client's responsibility to select the `SubjectAsset` and `PaymentAsset` according to their needs. The `TransactionType` determines the trade side the client is interested in.

The price oracle is currently called in two places:
https://github.com/lightninglabs/taproot-assets/blob/eeca76c074bbc91197cfb67fb03cf384cd008b49/rfq/oracle.go#L310-L326

https://github.com/lightninglabs/taproot-assets/blob/eeca76c074bbc91197cfb67fb03cf384cd008b49/rfq/oracle.go#L228-L257

At both call sites the payment asset is BTC and it's the transaction type that varies.

I've tried to develop a basic price oracle in a generalized manner, such that it returns a price without any awareness of asset channels. I think that makes sense for light tap clients but maybe not for edge nodes?

docs/examples/basic-price-oracle/main.go
ffranr: In the latest commits I've included an example in the doc for `QueryRateTick`.

docs/examples/basic-price-oracle/main.go
```diff
@@ -0,0 +1,163 @@
...
+// asset max amount.
+func getRateTick(transactionType oraclerpc.TransactionType,
+	subjectAssetMaxAmount uint64) oraclerpc.RateTick {
+
+	// Determine the rate based on the transaction type.
+	var rate uint64
+	if transactionType == oraclerpc.TransactionType_PURCHASE {
+		// The rate for a purchase transaction is 42,000 milli-satoshi
+		// per asset unit
+		rate = 42_000
```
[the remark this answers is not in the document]
ffranr: I've updated this comment (and the one below) to state "42,000 asset units per mSAT."

docs/examples/basic-price-oracle/main.go
```diff
@@ -0,0 +1,185 @@
...
+		expiry = time.Now().Add(1 * time.Minute).Unix()
+	}
+
+	return oraclerpc.RateTick{
+		Rate:            rate,
+		ExpiryTimestamp: uint64(expiry),
+	}
+}
+
+// QueryRateTick queries the rate tick for a given transaction type, subject
```
Roasbeef, line 77: I think it would be good to further lift this into a high level markdown doc (along with how to set one up, the flags, etc). Don't think we need to block merging this based on that though.

Roasbeef (approved):
LGTM 🦄

docs/examples/basic-price-oracle/main.go
ffranr: I think I've mentioned this to @Liongrass .

Issue #989, opened 2024-07-02
[feature]/universe+rpc: add new command to push up proofs to universes


Roasbeef:
**Is your feature request related to a problem? Please describe.**

Today we have `tapcli universe sync` which'll attempt to _fetch_ proofs from the target universe. 

When we mint, we'll also push out to all registered universe federation members. We'll also [maintain a log to ensure we can push out to the set of federation members reliably](https://github.com/lightninglabs/taproot-assets/blob/2d8d4849df66c9632038018a8da01a882be39baf/universe/auto_syncer.go#L300-L345). 

However there isn't an easy way to just push out an arbitrary proof. 

**Describe the solution you'd like**

We should add `tapcli universe push` or `tapcli universe insert` that targets a remote universe rather than the local on.

dstadulis:
`tapcli universe push`
or
`tapcli universe broadcast`

appeal to me

gijswijs:
What does `universe proofs insert` do if not this?

guggero:
`tapcli universe proofs insert` inserts a proof into the _local_ universe. I think the goal of this issue is to have a command that uploads a proof from the local universe to a remove one (a single, specific proof instead of doing a full sync).

jharveyb:
Another option is to use the `RpcUniverseRegistrar.UpsertProofLeaf` func, with a connection made to a server specified via CLI flags. You could connect via `NewRpcUniverseRegistrar()`.

gijswijs:
As discussed with @jharveyb:

New rpc endpoint for the universe rpc: 'PushProof`, tapcli: `universe push`
The request should contain a `UniverseKey` so that we can query the issuance or transfer proof for a given asset from the local universe and a `ServerAddr` that we can use to connect to and push the proof. Alternatively we could assume that the universe is always a registered federation server and we can identify it by an ID.

We then use `universeProofQuery` to fetch the proof from the local universe.
https://github.com/lightninglabs/taproot-assets/blob/aa3d671f4bc4b7e14c44e4e172011d3740065bd3/cmd/tapcli/universe.go#L477-L478

And we push the proof to the remote universe, using `RpcUniverseRegistrar.UpsertProofLeaf`
https://github.com/lightninglabs/taproot-assets/blob/aa3d671f4bc4b7e14c44e4e172011d3740065bd3/universe_rpc_registrar.go#L83

PR #1014, opened 2024-07-11
taprpc+tapcli: add universe proof push command

jharveyb:
Addresses #989 .

I'm not sure if exposing the endpoint with a `/push` suffix in the best option; maybe it doesn't make sense to expose over REST at all.

Testing manually now.

### DIFF
--- a/cmd/tapcli/universe.go
+++ b/cmd/tapcli/universe.go
@@ -436,7 +436,8 @@ var universeProofCommand = cli.Command{
 	`,
 	Subcommands: []cli.Command{
 		universeProofQueryCommand,
-		universeProofInsertInsert,
+		universeProofInsertCommand,
+		universeProofPushCommand,
 	},
 }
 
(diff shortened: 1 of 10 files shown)


rpcserver.go
```diff
@@ -5340,6 +5340,78 @@ func (r *rpcServer) InsertProof(ctx context.Context,
...
+// attempt to upload the asset leaf.
+func (r *rpcServer) PushProof(ctx context.Context,
+	req *unirpc.PushProofRequest) (*unirpc.PushProofResponse, error) {
+
+	switch {
+	case req.Server == nil:
+		return nil, fmt.Errorf("remote Universe must be specified")
+
+	case req.Key == nil:
+		return nil, fmt.Errorf("Universe key must be specified")
```
guggero, line 5354: nit: error message should start lower case.

rpcserver.go
```diff
@@ -5340,6 +5340,78 @@ func (r *rpcServer) InsertProof(ctx context.Context,
...
+	}
+
+	switch {
+	case req.Server.Host == "" && req.Server.Id == 0:
```
guggero, line 5358: I think Golang switch statements are ordered and the first that matches is executed. So could combine this with the above.

taprpc/universerpc/universe.yaml
```diff
@@ -37,6 +37,13 @@ http:
         - post: "/v1/taproot-assets/universe/proofs/group-key/{key.id.group_key_str}/{key.leaf_key.op.hash_str}/{key.leaf_key.op.index}/{key.leaf_key.script_key_str}"
           body: "*"
 
+    - selector: universerpc.Universe.PushProof
```
guggero, line 40: I probably would've put the "verb" of the action a bit earlier. So for example `/v1/taproot-assets/universe/proofs/push/asset-id/.....` and `/v1/taproot-assets/universe/proofs/push/group-key/.....`.

cmd/tapcli/universe.go
```diff
@@ -581,6 +585,59 @@ func universeProofInsert(ctx *cli.Context) error {
...
 }
 
+var universeProofPushCommand = cli.Command{
+	Name:      "push",
+	ShortName: "p",
+	Usage:     "push a proof to a remote Universe",
+	Description: `
+	Push a proof present in the local Universe to a remote Universe.
+	`,
+	Flags:  append(universeProofArgs, universeServerArgs...),
```
guggero, line 595: The `--server_id` flag has the text `the ID of the universe server to delete` which is a bit weird in this context. Might need to re-define it here.

guggero (approved):
tACK, very nice. Just a couple of nits.

GeorgeTsagk (approved):
LGTM! ✔️

taprpc/universerpc/universe.yaml
jharveyb: That was my first choice but I was undecided if that was too similar to Query and Insert. IIUC the `body` having a different name & structure should prevent any mixups as well as the path.

Roasbeef: Looks like this doesn't explicitly include the universe server ID. Will the REST mapping just make this zero?

guggero: There's still the body of the REST request that takes any variable that wasn't specified in the URL. And since the universe server ID is optional, it makes sense to have it there IMO.

Issue #1026, opened 2024-07-16
[bug]: FundChannel UX improvements


vanditshah99:
We are facing weird issue with lit 
```
{
    "version": "0.13.99-alpharc4 commit=v0.13.99-alpha.rc4"
}
```

We minted two assets having genesis transactions as
```
https://mempool.space/testnet/tx/2d74f4358ab17ddbc296afe2272e3b82d1be395b0c7dddd6e43275f024aa06c3 
```
```
https://mempool.space/testnet/tx/2d74f4358ab17ddbc296afe2272e3b82d1be395b0c7dddd6e43275f024aa06c3
```

This two transaction are confirmed, but when we try to do a fund channel with peer using the following command
```
litcli -n testnet --macaroonpath /root/.lnd/data/chain/bitcoin/testnet/admin.macaroon ln fundchannel --node_key 020c3b59b1854d31b94c0dd9df1019d251b4cd5565129f7ad16e3ec8c4671afd8c --sat_per_vbyte 20 --asset_amount 500000000 --asset_id 9f697c53c52ad5b80740caf3b9e299c8491c04625b224632eb85a9633f2a1522
```

We receive this error 
```
[litcli] error funding channel: rpc error: code = Unknown desc = error funding channel: unable to fund vPacket: error fetching input proof: cannot fetch proof for input asset: unable to find proof
```

Also post this error asset list API of tapcli doesn't show the newly minted asset anymore!

[df29cf30b0e641d26120bcecd43847edc0842b4dcb6cc858a2c5f3483688954b.txt](https://github.com/user-attachments/files/16248012/df29cf30b0e641d26120bcecd43847edc0842b4dcb6cc858a2c5f3483688954b.txt)
[2d74f4358ab17ddbc296afe2272e3b82d1be395b0c7dddd6e43275f024aa06c3.txt](https://github.com/user-attachments/files/16248023/2d74f4358ab17ddbc296afe2272e3b82d1be395b0c7dddd6e43275f024aa06c3.txt)

guggero:
Thanks for the logs. Does the node that received the channel (the one not opening it) have universe sync turned on?
I think you need the new `--taproot-assets.universe.sync-all-assets` flag and it needs to be connected to the same federation the channel creator is. Because in the logs I don't see any proof sync messages on the receiver.

guggero:
[...]

PR #1029, opened 2024-07-16
tapchannel: ensure genesis proofs for input assets already known

Roasbeef:
With this commit, we partially address https://github.com/lightninglabs/taproot-assets/issues/1026 by ensuring that when we
receive an incoming funding request, then we've already verified the
genesis proofs of the incoming asset.

If we haven't verified the proofs yet, then we'll try to fetch from a
remote universe. Otherwise, this check will be a noop.

### DIFF
--- a/tapfreighter/wallet.go
+++ b/tapfreighter/wallet.go
@@ -388,7 +388,35 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
 		return nil, err
 	}
 
-	return f.fundPacketWithInputs(ctx, fundDesc, vPkt, selectedCommitments)
+	// If we return with an error, we want to release the coins we've
+	// selected.
+	success := false
+	defer func() {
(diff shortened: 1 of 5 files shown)


guggero (approved):
Nice. LGTM pending CI fix.

tapfreighter/wallet.go
```diff
@@ -388,7 +388,35 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
 		return nil, err
 	}
 
-	return f.fundPacketWithInputs(ctx, fundDesc, vPkt, selectedCommitments)
+	// If we return with an error, we want to release the coins we've
+	// selected.
+	success := false
+	defer func() {
```
jharveyb, line 394: neat pattern!

jharveyb (approved):
LGTM pending linter fix.

Issue #1019, opened 2024-07-13
[bug]: In Ver lightning-terminal:v0.13.99-alpha.rc4


lukegao209:
### Background

I understand that you guys might be working on bugs in this version, but I wanted to share the issues I’ve encountered in case it helps.

	•	Version: lightning-terminal:v0.13.99-alpha.rc4
	•	Litds: Alice and Bob
	•	Network: Regtest
	•	Bitcoin Node: Bitcoind (mint 1 block/min), BlockHeight: 10000+
### Issues:
* Long Sync Time: After the initial startup, Litd takes a long time to sync before it becomes usable, at least 30 minutes. This seems excessively long. 

```sh
2024-07-13 10:21:11 2024-07-13 02:21:11.691 [INF] LTND: Waiting for wallet encryption password. Use `lncli create` to create a wallet, `lncli unlock` to unlock an existing wallet, or `lncli changepassword` to change the password of an existing wallet and unlock it.
2024-07-13 10:21:45 2024-07-13 02:21:45.611 [INF] LNWL: Opened wallet
2024-07-13 10:21:45 2024-07-13 02:21:45.612 [INF] LTND: Wallet recovery mode enabled with address lookahead of 12345678 addresses
2024-07-13 10:21:45 2024-07-13 02:21:45.640 [INF] LITD: Connecting basic lnd client
2024-07-13 10:21:45 2024-07-13 02:21:45.641 [INF] LITD: Basic lnd client connected
2024-07-13 10:21:45 2024-07-13 02:21:45.641 [INF] LITD: Connecting full lnd client
2024-07-13 10:21:45 2024-07-13 02:21:45.641 [INF] LNDC: Creating lnd connection to 127.0.0.1:10009
2024-07-13 10:21:45 2024-07-13 02:21:45.642 [INF] LNDC: Connected to lnd
2024-07-13 10:21:45 2024-07-13 02:21:45.643 [INF] LNDC: Waiting for lnd to unlock
2024-07-13 10:21:45 2024-07-13 02:21:45.643 [INF] LNDC: Wallet state of lnd is now: Wallet is unlocked
2024-07-13 10:21:46 2024-07-13 02:21:46.702 [INF] LNWL: Started listening for bitcoind block notifications via ZMQ on 54.92.19.81:28334
2024-07-13 10:21:46 2024-07-13 02:21:46.702 [INF] LNWL: Started listening for bitcoind transaction notifications via ZMQ on 54.92.19.81:28335
[...]

PR #1030, opened 2024-07-17
custom channels: more bug fixes

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1019.
Fixes https://github.com/lightninglabs/taproot-assets/issues/1026.

Fixes another bug found by increasing the test coverage in https://github.com/lightninglabs/lightning-terminal/pull/797: When there is no balance and no HTLC going toward the initiator of a channel, we still need to create a split root output.

We then also fix an issue that we didn't send a funding error to the remote party when we couldn't fund the channel. So they had it pending until it timed out after 10 minutes (or restart) and we couldn't try again ("number of pending channel exceeds maximum").

And finally, we make sure we don't allow not yet confirmed freshly minted assets to be used for funding any asset transaction, including channel openings. To make the user experience around those unconfirmed assets we also filter them by default and add a new count and flag to the RPC interface.


tapchannel/commitment.go
```diff
@@ -688,34 +687,14 @@ func CreateAllocations(chanState *channeldb.OpenChannel, ourBalance,
...
-		// channel initiator, then the HTLC is for the channel
-		// initiator.
-		case !chanState.IsInitiator && !isIncoming:
-			htlcIsForInitiator = true
-		}
-
-		// We should always just have a single split root on the side
-		// of the initiator. So if the initiator doesn't have any normal
-		// asset output to house the split root, then we'll take the
-		// _first_ HTLC that pays to the initiator.
```
Roasbeef, line 716: In theory we could also enforce an asset reserve requirement as well. I think we'd need some extra TLVs in the `open_channel` and `accept_channel` message to do this properly.

tapsend/send.go
```diff
@@ -590,6 +590,14 @@ func PrepareOutputAssets(ctx context.Context, vPkt *tappsbt.VPacket) error {
 		splitLocators = append(splitLocators, &locator)
 	}
 
+	// If we're creating a split commitment, there must be a split root.
+	// If none of the outputs was designated as the split root, then
+	// something was done incorrectly.
+	if rootLocator == nil {
+		return fmt.Errorf("cannot create split commitment without " +
```
Roasbeef, line 597: If we make this into a concrete error like `ErrNoRootLocator`, then in the future we can match against it within unit tests.

tapchannel/aux_funding_controller.go
```diff
@@ -1264,7 +1264,7 @@ func (f *FundingController) processFundingMsg(ctx context.Context,
 		)
 		if err != nil {
 			return tempPID, fmt.Errorf("unable to create "+
-				"commitment: %v", err)
+				"commitment: %w", err)
```
Roasbeef, line 1268: Weird that the linter didn't catch this, thought I had addressed all these areas.

tapchannel/aux_funding_controller.go
```diff
@@ -1442,6 +1443,19 @@ func (f *FundingController) processFundingReq(fundingFlows fundingFlowIndex,
 					uErr)
 			}
 
+			// If anything went wrong during the funding process,
```
Roasbeef, line 1446: 👍

rpcserver.go
```diff
@@ -975,6 +975,34 @@ func (r *rpcServer) ListAssets(ctx context.Context,
...
+		unconfirmedMints uint64
+	)
+
+	// We now count and filter the assets according to the
+	// IncludeUnconfirmedMints flag.
+	for idx := range rpcAssets {
+		switch {
+		// If the asset isn't confirmed yet, we count it but only
+		// include it in the output list if the client requested it.
+		case rpcAssets[idx].ChainAnchor.BlockHeight == 0:
```
Roasbeef, line 989: Any particular reason we're opting to do this here instead of within the db query? Mainly just for a smaller diff?

tapdb/assets_store.go
```diff
@@ -1823,6 +1823,11 @@ func (a *AssetStore) ListEligibleCoins(ctx context.Context,
 	assetFilter.Spent = sqlBool(false)
 	assetFilter.Leased = sqlBool(false)
 
+	// We also only want to select confirmed commitments (freshly minted
+	// unconfirmed assets would otherwise be included). Unconfirmed assets
+	// have a block height of 0, so we set the minimum block height to 1.
+	assetFilter.MinAnchorHeight = sqlInt32(1)
```
Roasbeef, line 1829: 👍

rpcserver.go
guggero: Mainly so we can _count_ how many assets we filter out. Which is purely a user experience thing, as the user would get a hint that their freshly minted assets might not appear in the list because they're unconfirmed.

I guess for now this works and avoids us needing to do two queries. But once we add pagination, we might need to change the approach here... Can do it now if you think it's worth doing.
Kinda forgot about our plans to add pagination, so happy to change it.

tapchannel/aux_funding_controller.go
guggero: Yeah, I also thought we turned on that linter...

tapchannel/commitment.go
guggero: True. But I don't see why we'd want to do that? I understand that on the BTC level it's so both parties have something at stake. And since we still have the reserve requirement on the BTC level, that holds true even if we don't also enforce it on the asset level...

gijswijs (commented):
Codewise, lgtm.
I do have a few questions so that I can confirm my understanding.

This PR specifically fixes the issue where the split root becomes nil with commit cb19dd62bb0fc841bda2713ce00c51bcc556fb4e.

This issue came to light with this report #1019

It also fixes the issue of poor UX that resulted in selecting unconfirmed minting transactions mentioned here #1026

The are 2 todos mentioned in that thread
> As a funder, release locked/leased UTXOs if funding fails
> As a recipient, do an asset specific universe sync on incoming funding request, to make sure we know at least the meta info and other issuance proof

Those 2 todos are not fixed in this PR but in PR #1029, Correct?

I want to be sure that we can close that issue.

guggero:
@gijswijs correct, those two TODOs were already addressed in https://github.com/lightninglabs/taproot-assets/pull/1029.

Roasbeef (approved):
LGTM 🕍

PR #1032, opened 2024-07-18
tapdb+rpc: return lease owner and expiry for managed UTXOs

Roasbeef:
This should help us track down issues when assets aren't unlocked even
though they should be.

We also extend an existing unit test to ensure these values are properly
set.


tapdb/assets_store.go
```diff
@@ -417,6 +417,14 @@ type ManagedUTXO struct {
 	// TapscriptSibling is the serialized tapscript sibling preimage of
 	// this asset. This will usually be blank.
 	TapscriptSibling []byte
+
+	// LeaseOwner is the identifeir of the lease owner of this UTXO. If
```
guggero, line 421: nit: typo in identifier.

guggero (approved):
Very nice, LGTM :tada:

GeorgeTsagk (approved):
LGTM :green_circle:

PR #1033, opened 2024-07-19
tapchannel: use defer to ensure lease unlocked if funding failure

Roasbeef:
Once we call `fundVirtualPacket`, we've locked inputs assuming it succeeds. If for w/e reason, the remote peer fails to send us the first ack (before the call to `completeChannelFunding`), then before this commit, we would fail to unlock the inputs. Only after the call to `completeChannelFunding` would we unlock the UTXOs.

In this commit, we use two new defers to ensure that we always unlock the UTXOs if the pre-setup fails, and also if `completeChannelFunding` fails.

Fixes https://github.com/lightninglabs/taproot-assets/issues/1019

### DIFF
--- a/tapchannel/aux_funding_controller.go
+++ b/tapchannel/aux_funding_controller.go
@@ -505,6 +505,9 @@ func (p *pendingAssetFunding) unlockInputs(ctx context.Context,
 func (p *pendingAssetFunding) unlockAssetInputs(ctx context.Context,
 	coinSelect tapfreighter.CoinSelector) error {
 
+	log.Debugf("unlocking asset inputs: %v",
+		spew.Sdump(p.lockedAssetInputs))
+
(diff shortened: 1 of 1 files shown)


tapchannel/aux_funding_controller.go
```diff
@@ -1028,13 +1031,6 @@ func (f *FundingController) completeChannelFunding(ctx context.Context,
 		return nil, err
 	}
 
-	fundingState.lockedAssetInputs = fn.Map(
```
Roasbeef, line 1031: I moved this down so we track the locked inputs as soon as the initial vPSBT funding finishes.

guggero (approved):
Nice, LGTM :tada:

GeorgeTsagk (approved):
LGTM
nice trick with the `setupSuccess`/`completeSuccess` flags

Issue #1009, opened 2024-07-11
tapfreighter/[bug]: universe proof transfer not resumed after restart


Roasbeef:
### Background

After we confirm a transfer, and update proofs (only for the file based store), we'll attempt to upload any proofs (if needed) to a universe: https://github.com/lightninglabs/taproot-assets/blob/fca9d11f027cdb5b81a36c4a8999c6b010ac0c44/tapfreighter/chain_porter.go#L1165-L1190

The issue here is that this is run in a goroutine, with the state step function returning immediately. Upon restart, we won't attempt to publish the proofs again. Even if we don't have any proofs, this is important, as only once we transfer all the proofs, do we call `ConfirmParcelDelivery`: https://github.com/lightninglabs/taproot-assets/blob/main/tapfreighter/chain_porter.go#L795-L808. 

The function name is slightly overloaded, but `ConfirmParcelDelivery` is what will actually finalize the transfer by applying the pending outputs (make new assets w/ the relevant details): https://github.com/lightninglabs/taproot-assets/blob/fca9d11f027cdb5b81a36c4a8999c6b010ac0c44/tapdb/assets_store.go#L2816-L2842. 

Note that we _do_ have the transfer log, and can resume from that, but this doesn't allow us to also execute this call back that finalizes the transfer. 

One other relevant detail is that `transferReceiverProof` sets the state to `SendStateComplete`: https://github.com/lightninglabs/taproot-assets/blob/fca9d11f027cdb5b81a36c4a8999c6b010ac0c44/tapfreighter/chain_porter.go#L814, but only after it has completed. There's a line before the call that prematurely updates this state: https://github.com/lightninglabs/taproot-assets/blob/fca9d11f027cdb5b81a36c4a8999c6b010ac0c44/tapfreighter/chain_porter.go#L1169-L1172

### Steps to reproduce

Update an itest to initiate a transfer, confirm, but then cause the proof transfer to fail. 

### Expected behavior

It should continue to both transmit the proofs, but also fix the logic above to call `ConfirmParcelDelivery` as soon as we get the confirmation notification. 

We should also remove the premature update of `SendStateComplete`. This'll ensure that that case gets re-executed on start up until the proofs are actually sent. 

With the above, we also need to read the parcels from disk, to relaunch the state machine for parcels that weren't finalized. 

### Actual behavior

On restart, lnd won't reattempt proof transfer. However, the proof files on disk _are_ properly updated.

Roasbeef:
[...]

PR #1035, opened 2024-07-19
Store transfer output proof delivery status

ffranr:
### Summary

This PR is work towards fixing https://github.com/lightninglabs/taproot-assets/issues/1009 (making the proof transfer process more robust).

This pull request adds a new `proof_deliver_complete` column to the `asset_transfer_outputs` table. It includes updates to the table schema, modifications to SQL statements, and adjustments to unit tests. Additionally, it integrates changes to the `tapfreighter` component for logging proof delivery confirmations.

### Changes

1. **Schema Update**:
   - Added the `proof_deliver_complete` column to the `asset_transfer_outputs` table.
   
2. **SQL Statement Modifications**:
   - Updated the row insertion and retrieval SQL statements to include the `proof_deliver_complete` column.
   - Added a new SQL statement specifically for modifying the `proof_deliver_complete` column.
   
3. **Field Population**:
   - Set the `proof_deliver_complete` field for new transfer output rows.
   - Ensured the `proof_deliver_complete` field is populated when reading from the database.

6. **Proof Delivery Logging**:
   - Added a `ConfirmProofDelivery` method to the `tapfreighter` export log.
   - The `ConfirmProofDelivery` function is called after a proof has been successfully delivered to a remote peer.
   
7. **Unit Test Adjustment**:
   - Modified a unit test to verify that a transfer output proof is correctly flagged as confirmed.

### DIFF
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -2401,20 +2401,50 @@ func insertAssetTransferOutput(ctx context.Context, q ActiveAssetsStore,
 		return fmt.Errorf("unable to insert script key: %w", err)
 	}
 
+	// Now we will mark the output proof as undelivered if it is intended
+	// for a counterpart.
+	//
+	// If the transfer output proof is not intended for a counterpart the
+	// `proofDeliveryComplete` field will be left as NULL. Otherwise, we set
+	// it to false to indicate that the proof has not been delivered yet.
+	shouldDeliverProof, err := output.ShouldDeliverProof()
+	if err != nil {
+		return fmt.Errorf("unable to determine if proof should be "+
+			"delivery for given transfer output: %w", err)
+	}
+
+	var proofDeliveryComplete sql.NullBool
+	if shouldDeliverProof {
+		proofDeliveryComplete = sql.NullBool{
+			Bool:  false,
+			Valid: true,
+		}
+	}
+
+	// Check if position value can be stored in a 32-bit integer. Type cast
+	// if possible, otherwise return an error.
+	if output.Position > math.MaxInt32 {
+		return fmt.Errorf("position value %d is too large for db "+
+			"storage", output.Position)
+	}
+	position := int32(output.Position)
+
 	dbOutput := NewTransferOutput{
-		TransferID:          transferID,
-		AnchorUtxo:          newUtxoID,
-		ScriptKey:           scriptKeyID,
-		ScriptKeyLocal:      output.ScriptKeyLocal,
-		Amount:              int64(output.Amount),
-		LockTime:            sqlInt32(output.LockTime),
-		RelativeLockTime:    sqlInt32(output.RelativeLockTime),
-		AssetVersion:        int32(output.AssetVersion),
-		SerializedWitnesses: witnessBuf.Bytes(),
-		ProofSuffix:         output.ProofSuffix,
-		NumPassiveAssets:    int32(output.Anchor.NumPassiveAssets),
-		OutputType:          int16(output.Type),
-		ProofCourierAddr:    output.ProofCourierAddr,
+		TransferID:            transferID,
+		AnchorUtxo:            newUtxoID,
+		ScriptKey:             scriptKeyID,
+		ScriptKeyLocal:        output.ScriptKeyLocal,
+		Amount:                int64(output.Amount),
+		LockTime:              sqlInt32(output.LockTime),
+		RelativeLockTime:      sqlInt32(output.RelativeLockTime),
+		AssetVersion:          int32(output.AssetVersion),
+		SerializedWitnesses:   witnessBuf.Bytes(),
+		ProofSuffix:           output.ProofSuffix,
+		NumPassiveAssets:      int32(output.Anchor.NumPassiveAssets),
+		OutputType:            int16(output.Type),
+		ProofCourierAddr:      output.ProofCourierAddr,
+		ProofDeliveryComplete: proofDeliveryComplete,
+		Position:              position,
 	}
 
 	// There might not have been a split, so we can't rely on the split root
@@ -2526,6 +2556,22 @@ func fetchAssetTransferOutputs(ctx context.Context, q ActiveAssetsStore,
 			outputAnchor.CommitmentVersion = fn.Ptr(dbRootVersion)
 		}
 
+		// Parse the proof deliver complete flag from the database.
+		var proofDeliveryComplete fn.Option[bool]
+		if dbOut.ProofDeliveryComplete.Valid {
+			proofDeliveryComplete = fn.Some(
+				dbOut.ProofDeliveryComplete.Bool,
+			)
+		}
+
+		vOutputType := tappsbt.VOutputType(dbOut.OutputType)
+
+		// Ensure the position value is valid.
+		if dbOut.Position < 0 {
+			return nil, fmt.Errorf("invalid position value in "+
+				"db: %d", dbOut.Position)
+		}
+
 		outputs[idx] = tapfreighter.TransferOutput{
 			Anchor:           outputAnchor,
 			Amount:           uint64(dbOut.Amount),
@@ -2549,9 +2595,11 @@ func fetchAssetTransferOutputs(ctx context.Context, q ActiveAssetsStore,
 				splitRootHash,
 				uint64(dbOut.SplitCommitmentRootValue.Int64),
 			),
-			ProofSuffix:      dbOut.ProofSuffix,
-			Type:             tappsbt.VOutputType(dbOut.OutputType),
-			ProofCourierAddr: dbOut.ProofCourierAddr,
+			ProofSuffix:           dbOut.ProofSuffix,
+			Type:                  vOutputType,
+			ProofCourierAddr:      dbOut.ProofCourierAddr,
+			ProofDeliveryComplete: proofDeliveryComplete,
+			Position:              uint64(dbOut.Position),
 		}
 
 		err = readOutPoint(
@@ -2794,10 +2842,10 @@ func (a *AssetStore) ConfirmParcelDelivery(ctx context.Context,
 				!out.ScriptKeyLocal && !isKnown
 
 			log.Tracef("Skip asset creation for "+
-				"output %d?: %v,  scriptKey=%x, "+
+				"output %d?: %v,  position=%v, scriptKey=%x, "+
 				"isTombstone=%v, isBurn=%v, "+
 				"scriptKeyLocal=%v, scriptKeyKnown=%v",
-				idx, skipAssetCreation,
+				idx, skipAssetCreation, out.Position,
 				scriptPubKey.SerializeCompressed(),
 				isTombstone, isBurn, out.ScriptKeyLocal,
 				isKnown)
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -126,6 +126,12 @@ type (
 	// output.
 	NewTransferOutput = sqlc.InsertAssetTransferOutputParams
 
+	// OutputProofDeliveryStatus wraps the params needed to set the delivery
+	// status of a given output proof.
+	//
+	// nolint: lll
+	OutputProofDeliveryStatus = sqlc.SetTransferOutputProofDeliveryStatusParams
+
 	// NewPassiveAsset wraps the params needed to insert a new passive
 	// asset.
 	NewPassiveAsset = sqlc.InsertPassiveAssetParams
@@ -277,6 +283,11 @@ type ActiveAssetsStore interface {
 	// the DB.
 	InsertAssetTransferOutput(ctx context.Context,
 		arg NewTransferOutput) error
+
+	// SetTransferOutputProofDeliveryStatus sets the delivery status of a
+	// given transfer output proof.
+	SetTransferOutputProofDeliveryStatus(ctx context.Context,
+		arg OutputProofDeliveryStatus) error
 
 	// FetchTransferInputs fetches the inputs to a given asset transfer.
 	FetchTransferInputs(ctx context.Context,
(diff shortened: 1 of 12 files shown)


tapdb/sqlc/queries/transfers.sql
```diff
@@ -23,11 +23,36 @@ INSERT INTO asset_transfer_outputs (
...
+WITH target(output_id) AS (
+    SELECT output_id
+    FROM asset_transfer_outputs output
+    JOIN asset_transfers
+      ON output.transfer_id = asset_transfers.id
+    JOIN chain_txns
+      ON asset_transfers.anchor_txn_id = chain_txns.txn_id
+    JOIN script_keys
+      ON output.script_key = script_keys.script_key_id
+    WHERE (chain_txns.txid = sqlc.narg('anchor_tx_id') OR sqlc.narg('anchor_tx_id') IS NULL)
```
jharveyb, line 41: Was thinking that we may not need every field here, but after finishing the review that seems to make sense / we don't actually have to use them all when using this query.

tapdb/assets_store.go
```diff
@@ -2702,6 +2739,70 @@ func (a *AssetStore) QueryProofTransferLog(ctx context.Context,
...
+	err := asset.WitnessEncoder(&witnessBuf, &args.WitnessData, &scratch)
+	if err != nil {
+		return fmt.Errorf("unable to encode witness: %w", err)
+	}
+
+	// Serialize the script key.
+	serialisedScriptKey := args.ScriptKey.PubKey.SerializeCompressed()
+
+	// Deconstruct the split commitment root if provided.
+	var (
```
jharveyb, line 2764: super-nit: can just these `var` blocks together and remove `splitCommitmentRootHash`.

Non-blocking.

jharveyb (approved):
Looks good overall!

AFAICT this is missing the reordering of when proof delivery occurs mentioned here:

https://github.com/lightninglabs/taproot-assets/issues/1009#issuecomment-2226193293

Is that intended for a follow-up PR?

ffranr:
> Looks good overall!
> 
> AFAICT this is missing the reordering of when proof delivery occurs mentioned here:
> 
> [#1009 (comment)](https://github.com/lightninglabs/taproot-assets/issues/1009#issuecomment-2226193293)
> 
> Is that intended for a follow-up PR?

@jharveyb Thanks for the review. Yes, will definitely get done. I was thinking in a follow up PR.

tapdb/sqlc/queries/transfers.sql
```diff
@@ -23,11 +23,36 @@ INSERT INTO asset_transfer_outputs (
...
+      AND (script_keys.tweaked_script_key = sqlc.narg('script_key') OR sqlc.narg('script_key') IS NULL)
+      AND (output.script_key_local = sqlc.narg('script_key_local') OR sqlc.narg('script_key_local') IS NULL)
+      AND (output.amount = sqlc.narg('amount') OR sqlc.narg('amount') IS NULL)
+      AND (output.serialized_witnesses = sqlc.narg('serialized_witnesses') OR sqlc.narg('serialized_witnesses') IS NULL)
```
guggero, line 45: I don't think we should use any of these serialized values as query elements.
What I think we need is a way to uniquely address a specific transfer output. So we should add a new column `position` (trying to avoid the use of the word "index" to not make this easy to confuse with the output index of a transaction) on the `asset_transfer_outputs` table that for new transfer outputs corresponds to the index in the `OutboundParcel.Outputs` slice. And for existing transfer outputs we can just use the primary key to back fill, since we just need a value that's unique per transfer.

Then this whole query can just query on the `anchor_utxo` ID (outpoint instead of just TXID) and the `position`.

So basically this:
```sql
-- The output position indicates the order of the output in the list of outputs
-- for a given transfer. We'll be inserting an actual value in the next query,
-- so we just start with -1.
ALTER TABLE asset_transfer_outputs ADD COLUMN position INTEGER NOT NULL DEFAULT -1;

-- Update the position to be the same as the output id for existing entries.
-- We'll use the position integer as a uniquely identifiable number of an output
-- within a transfer, so setting the default to the output_id is just to make
-- sure we have a unique value that also satisfies the unique constraint we add
-- below.
UPDATE asset_transfer_outputs SET position = CAST(output_id AS INTEGER)
WHERE position = -1;

-- We need to be able to upsert outputs, so we need a unique constraint on
-- (output_id, position).
CREATE UNIQUE INDEX asset_transfer_outputs_output_id_position_unique
ON asset_transfer_outputs (
    output_id, position
);
```

tapdb/assets_store.go
```diff
@@ -2728,6 +2739,70 @@ func (a *AssetStore) QueryProofTransferLog(ctx context.Context,
 	return timestamps, err
 }
 
+// ConfirmProofDelivery marks a transfer output proof as successfully
+// delivered to counterparty.
+func (a *AssetStore) ConfirmProofDelivery(ctx context.Context,
+	args tapfreighter.ProofDeliveryConfirmation) error {
```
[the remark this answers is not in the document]
guggero: Actually, this should become way easier with the other comment I made. Main point here is to use the output's `Anchor.Outpoint` instead of just the `anchorTXID` and the new `position` to make this update bullet proof.

guggero (changes requested):
Nice approach!
Makes sense to me but I think the current implementation on the SQL level is a bit brittle and will likely lead to issues down the line.
Added a suggestion in there that should solve it in a nice way (and with probably a smaller diff too).

tapdb/assets_store.go
```diff
@@ -2401,20 +2401,35 @@ func insertAssetTransferOutput(ctx context.Context, q ActiveAssetsStore,
...
 	}
 
+	// Mark the output proof as undelivered if it is intended for a
+	// counterpart. A proof is considered destined for a remote peer if a
+	// proof courier address is present or if the script key is not local.
+	//
+	// If the transfer output proof is not intended for a counterpart the
+	// `proofDeliverComplete` field will be left as NULL.
+	var proofDeliverComplete sql.NullBool
+	if len(output.ProofCourierAddr) > 0 || !output.ScriptKeyLocal {
```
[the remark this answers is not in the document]
ffranr: Here we need the conditions under which a proof will be delivered. If a proof is to be delivered then `proofDeliverComplete` is set to `false` at this point. Which means: "should be delivered, but hasn't yet". Otherwise, if there is no proof to be delivered then we leave it as `null`.

So maybe we only need `if len(output.ProofCourierAddr) > 0` here?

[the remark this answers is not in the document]
ffranr: I've added more to the comment to make that clear.

tapdb/sqlc/queries/transfers.sql
ffranr: I'm not sure if the new unique constraint `asset_transfer_outputs_output_id_position_unique` is necessary. `output_id` is already a primary key, so its unique. And so (output_id, position) will be unique because `output_id` is a primary key.

ffranr: The deliver subroutine doesn't currently have access to the output index in `OutboundParcel.Outputs`. We use:

```
err := fn.ParSlice(ctx, pkg.OutboundPkg.Outputs, deliver)
```

Would be necessary to have a version of `fn.ParSlice` which passes in the index. So we can get the position to `p.cfg.ExportLog.ConfirmProofDelivery`. That's why I didn't go down the `position` route but rather specify everything. But I agree that ultimately a position index would be more solid. I will work on this.

guggero: The unique index is more of a safety net to make sure we don't screw anything up, since we rely on the position being unique for each transfer.

> The deliver subroutine doesn't currently have access to the output index ...

Ah, yes, because of the `fn.ParSlice`. I think it makes sense to just add a new `Position` field on the `OutboundParcel.Output` struct, since it will be stored in the database as well. Then we can still make use of the async/unordered nature of `fn.ParSlice`

ffranr: > The unique index is more of a safety net to make sure we don't screw anything up, since we rely on the position being unique for each transfer.

Ah got it. A unique constraint on transaction id and position perhaps.

guggero: Ah, yes, of course... Sorry, seems I was a bit confused here. `output_id` doesn't make any sense, should be `transfer_id, position`.

tapdb/assets_store.go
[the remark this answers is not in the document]
guggero: I'm not sure we can just rely on the proof delivery address being set. It's true that we never set the delivery address for tombstone outputs or burns. But an RPC user might, mistakenly.
So I really think we should extract all the conditions linked above into a method on the `TransferOutput` struct, something like `ShouldDeliverProof`. That also makes the `ChainPorter.transferReceiverProof` way slimmer.

tapdb/assets_store.go
```diff
@@ -2745,6 +2756,47 @@ func (a *AssetStore) QueryProofTransferLog(ctx context.Context,
...
+	if err != nil {
+		return fmt.Errorf("unable to encode anchor outpoint: %w", err)
+	}
+
+	// Ensure that the position value can be stored in a 32-bit integer.
+	// Type cast if possible, otherwise return an error.
+	if outputPosition > math.MaxInt32 {
+		return fmt.Errorf("position value is too large for db: %d ",
+			outputPosition,
+		)
```
guggero, line 2775: nit: closing parenthesis on previous line for function call with formatting directive (also, there's an extra space at the end of the string).

tapdb/assets_store.go
```diff
@@ -2745,6 +2756,47 @@ func (a *AssetStore) QueryProofTransferLog(ctx context.Context,
...
+
+	err = a.db.ExecTx(ctx, &writeTxOpts, func(q ActiveAssetsStore) error {
+		params := OutputProofDeliveryStatus{
+			DeliveryComplete:         sqlBool(true),
+			SerializedAnchorOutpoint: anchorOutpointBytes,
+			Position:                 outPosition,
+		}
+		return q.SetTransferOutputProofDeliveryStatus(ctx, params)
+	},
+	)
```
guggero, line 2791: nit: closing parenthesis on previous line.

tapdb/assets_store.go
```diff
@@ -2745,6 +2756,47 @@ func (a *AssetStore) QueryProofTransferLog(ctx context.Context,
...
+	// Ensure that the position value can be stored in a 32-bit integer.
+	// Type cast if possible, otherwise return an error.
+	if outputPosition > math.MaxInt32 {
+		return fmt.Errorf("position value is too large for db: %d ",
+			outputPosition,
+		)
+	}
+	outPosition := int32(outputPosition)
+
+	var (
```
guggero, line 2779: nit: single line `var` statement? Or move `outPosition = int32(outputPosition)` into the `var` block?

guggero (commented):
Very happy with the SQL part, should make things more reliable.
Almost there on the rest!

guggero (approved):
Very nice, LGTM :rocket:

tapdb/sqlc/migrations/000022_transfer_outputs_proof_delivered.up.sql
```diff
@@ -0,0 +1,28 @@
...
+ADD COLUMN proof_delivery_complete BOOL;
+
+-- Add column `position` which indicates the index of the output in the list of
+-- outputs for a given transfer. This index position in conjunction with the
+-- transfer id can be used to uniquely identify a transfer output.
+--
+-- We'll be inserting an actual value in the next query, so we just start
+-- with -1.
+ALTER TABLE asset_transfer_outputs
+ADD COLUMN position INTEGER NOT NULL DEFAULT -1;
```
jharveyb, line 13: Why not use a `BIGINT` here? Which should translate to an `int64` on the Go side, and reduce the need for some of the type checking elsewhere in the PR.

tapfreighter/chain_porter.go
```diff
@@ -640,50 +640,21 @@ func (p *ChainPorter) transferReceiverProof(pkg *sendPackage) error {
...
-			log.Debugf("Not transferring proof for local output "+
-				"script key %x", key.SerializeCompressed())
-			return nil
-		}
-
-		// Un-spendable means this is a tombstone output resulting from
-		// a split.
-		unSpendable, err := out.ScriptKey.IsUnSpendable()
+		// We'll first check to see if the proof should be delivered.
+		shouldDeliverProof, err := out.ShouldDeliverProof()
```
jharveyb, line 649: 👍🏽

tapdb/assets_store.go
```diff
@@ -2401,20 +2412,50 @@ func insertAssetTransferOutput(ctx context.Context, q ActiveAssetsStore,
...
+	if shouldDeliverProof {
+		proofDeliveryComplete = sql.NullBool{
+			Bool:  false,
+			Valid: true,
+		}
+	}
+
+	// Check if position value can be stored in a 32-bit integer. Type cast
+	// if possible, otherwise return an error.
+	if output.Position > math.MaxInt32 {
```
jharveyb, line 2437: See note elsewhere re: switching to `int64`.

jharveyb (commented):
Looks good overall! The newer version is definitely a smaller / simpler diff. Just one note on the position type itself.

tapdb/sqlc/migrations/000022_transfer_outputs_proof_delivered.up.sql
jharveyb: As a counterargument, I don't think we can reasonably perform a transfer with `2^32` outputs anyways.  I'm not sure if there's an argument for supporting that.

tapdb/sqlc/migrations/000022_transfer_outputs_proof_delivered.up.sql
```diff
@@ -0,0 +1,28 @@
+-- Add a column to track if the proof has been delivered for an asset transfer
+-- output.
+ALTER TABLE asset_transfer_outputs
+ADD COLUMN proof_delivery_complete BOOL;
+
+-- Add column `position` which indicates the index of the output in the list of
+-- outputs for a given transfer. This index position in conjunction with the
```
Roasbeef, line 7: Couldn't we just use the existing primary key for this?

tapfreighter/interface.go
```diff
@@ -248,6 +248,48 @@ type TransferOutput struct {
...
+			"unspendable: %w", err)
+	}
+
+	if unSpendable {
+		return false, nil
+	}
+
+	// If this is an output that is going to our own node/wallet, we don't
+	// need to deliver a proof.
+	if out.ScriptKey.TweakedScriptKey != nil && out.ScriptKeyLocal {
```
Roasbeef, line 289: We should double check this with some of the litd itests we have for channels. At times, we verify that a proof was uploaded even if it was a local script key.

tapfreighter/interface.go
```diff
@@ -248,6 +248,48 @@ type TransferOutput struct {
...
+		return false, nil
+	}
+
+	// If the script key is a burn key, we don't need to deliver a proof.
+	if len(out.WitnessData) > 0 && asset.IsBurnKey(
+		out.ScriptKey.PubKey, out.WitnessData[0],
```
Roasbeef, line 295: Aren't burns still relevant? So to proof to someone that you've burnt an asset.

tapfreighter/chain_porter.go
```diff
@@ -643,47 +643,16 @@ func (p *ChainPorter) transferReceiverProof(pkg *sendPackage) error {
...
-		if unSpendable {
-			log.Debugf("Not transferring proof for un-spendable "+
-				"output script key %x",
-				key.SerializeCompressed())
-			return nil
-		}
-
-		// Burns are also always kept local and not sent to any
-		// receiver.
-		if len(out.WitnessData) > 0 && asset.IsBurnKey(
```
Roasbeef, line 670: Ah I see this was also just existing behavior, feel free to ignore those other comments.

PR #1037, opened 2024-07-19
TLV encoding: add forward-compatibility

guggero:
Replaces https://github.com/lightninglabs/taproot-assets/pull/557.

Fixes https://github.com/lightninglabs/taproot-assets/issues/997.

With this PR we add forward-compatibility that will allow us in the future to add new optional (odd-numbered) TLV types to any of our TLV structs without breaking old clients.
This will allow us to implement https://github.com/lightninglabs/taproot-assets/issues/956 in the future, once everyone has upgraded to `v0.4.1` (or whichever version this lands in).

I opted to not go with any version increases, since this doesn't really change anything in the current behavior yet...

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1037)


guggero:
Rebased.

asset/records.go
```diff
@@ -30,6 +34,87 @@ const (
 	LeafAssetID LeafTlvType = 11
 )
 
+// KnownAssetLeafTypes is a set of all known asset leaf TLV types. This set is
+// asserted to be complete by a check in the BIP test vector unit tests.
```
Roasbeef, line 38: 👍

asset/records.go
```diff
@@ -30,6 +34,87 @@ const (
...
+		}
+	}
+
+	return nil
+}
+
+// TlvStrictDecode attempts to decode the passed buffer into the TLV stream. It
+// takes the set of known types for a given stream, and returns an error if the
+// buffer includes any unknown even types.
+func TlvStrictDecode(stream *tlv.Stream, r io.Reader,
```
Roasbeef, line 113: 👍

asset/asset.go
```diff
@@ -2000,7 +2000,8 @@ func (a *Asset) Decode(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	return stream.Decode(r)
+
+	return TlvStrictDecode(stream, r, KnownAssetLeafTypes)
```
Roasbeef: I think to be slightly more conservative, we should put this behind a flag that's disabled for now. This protects us in the off chain that someone is using a modified version of the daemon with their own custom flag, or is using a fresh implementation. 

We could check for this in the wild by running the new decode method on the existing assets in proofs uploaded to our Universe.

asset/asset.go
```diff
@@ -2001,7 +2013,14 @@ func (a *Asset) Decode(r io.Reader) error {
 		return err
 	}
 
-	return TlvStrictDecode(stream, r, KnownAssetLeafTypes)
+	unknownOddTypes, err := TlvStrictDecode(stream, r, KnownAssetLeafTypes)
+	if err != nil {
+		return err
+	}
+
+	a.unknownOddTypes = unknownOddTypes
```
Roasbeef, line 2032: Dig the implementation! 

One gap (still getting through the PR) that may pop up is when we encode a protocol object within the database as a fully featured row vs encoding an encoded blob.

asset/records.go
```diff
@@ -87,32 +87,74 @@ func AssertNoUnknownEvenTypes(parsedTypes tlv.TypeMap,
 	return nil
 }
 
+// FilterUnknownTypes filters out all types that are unknown from the given
+// parsed types. The known types are specified as a set.
+func FilterUnknownTypes(parsedTypes tlv.TypeMap,
+	knownTypes fn.Set[tlv.Type]) tlv.TypeMap {
+
+	result := make(tlv.TypeMap, len(parsedTypes))
```
Roasbeef, line 95: We should be able to use `Set.Diff` here: https://github.com/lightningnetwork/lnd/blob/b7c59b36a74975c4e710a02ea42959053735402e/fn/set.go#L34-L43

Roasbeef: Then we'd need to run through again to read the keys in the output of the diff from the `parsedTypes` argument. So LoC wise, the current version may actually be the most minimal.

asset/asset.go
```diff
@@ -1873,16 +1873,27 @@ func (a *Asset) deepEqual(allowSegWitIgnoreTxWitness bool, o *Asset) bool {
...
+	switch {
+	// If only one of the keys is nil, they are not equal.
+	case (a.ScriptKey.PubKey == nil && o.ScriptKey.PubKey != nil) ||
+		(a.ScriptKey.PubKey != nil && o.ScriptKey.PubKey == nil):
+
 		return false
-	}
 
-	/*if !reflect.DeepEqual(a.ScriptKey, o.ScriptKey) {
+	// If both are non-nil, we compare the public keys.
```
Roasbeef, line 1887: 👍

Roasbeef (commented):
Solid approach! 

No major comments, other than one re how we can de-risk enforcing strict decode in a short time horizon. 

Also need to circle back re the interaction with the test vectors and the new type constants, as well as some of the forward compat unit tests.

jharveyb (commented):
Looks good overall! I think it could be useful to have some of this exercised in itest, especially for the composite / nested types like proofs.

asset/asset.go
jharveyb: Isn't that what we do for most fields?

guggero:
Will get back to this PR next week.

asset/asset.go
guggero: But wouldn't that break the intended forward compatibility?
Isn't the intention that when we ship `v0.4.2` with this code enabled (e.g. no flag, allow new unknown odd types) that we could then with `v0.5.0` add new fields with odd types, being backward compatible to `v0.4.2`?
With the intention that users can remain on `v0.4.2` without their nodes complaining when syncing new proofs created with `v0.5.0`. If they first have to turn on a flag to allow that, doesn't that kind of defeat the purpose?

asset/asset.go
guggero: We wouldn't actually _store_ the information in the unknown types. Just keep it in memory so we can validate correctly (e.g. arrive at the correct leaf hash for assets or the correct meta hash for meta reveals).
Only when you upgrade to the version that implements the new value, would we store it, with the corresponding database migration. But then it would be a known type and the value wouldn't end up in `unknownOddTypes` anymore.

asset/records.go
```diff
@@ -30,6 +34,129 @@ const (
...
+
+	return FilterUnknownTypes(parsedTypes, knownTypes), nil
+}
+
+// CombineRecords returns a new slice of records that combines the given records
+// with the unparsed types converted to static records.
+func CombineRecords(records []tlv.Record, unparsed tlv.TypeMap) []tlv.Record {
+	for k, v := range unparsed {
+		records = append(records, tlv.MakeStaticRecord(
+			k, nil, uint64(len(v)), tlv.StubEncoder(v), nil,
```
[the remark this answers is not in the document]
guggero: We only ever **encode** the records returned here. Decoding from those static/stub records isn't supported anyway, so no need passing in a decoder.

address/address_test.go
```diff
@@ -544,3 +555,57 @@ func FuzzAddressDecode(f *testing.F) {
...
+	require.NoError(t, err)
+
+	var buf bytes.Buffer
+	err = knownAddr.Encode(&buf)
+	require.NoError(t, err)
+
+	// With the known address now encoded, we can add an unknown even type
+	// to the encoded bytes. That should provoke an error when parsed again.
+	unknownTypeValue := []byte("I could be anything, really")
+	unknownEvenType := append([]byte{
```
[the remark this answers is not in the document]
guggero: Yeah, I thought the same when creating the tests but didn't get around to doing it. But have now made all tests as generic as possible.

proof/records.go
```diff
@@ -45,6 +46,45 @@ const (
 	MetaRevealDataType     tlv.Type = 2
```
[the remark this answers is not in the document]
guggero: No, I think 5 is correct. The way I understood the TLV numbering scheme is that you always have pairs, one even, one odd (for required/optional). So `commitment.ProofTaprootAssetProofType` defines the `0/1` and `2/3` pairs, then `CommitmentProofTapSiblingPreimageType` continues with the `4/5` pair.

Maybe this doesn't really make sense when we don't have both required and optional. But at least it's done consistently in that way throughout the code base (at least I think so).

guggero:
> Also need to circle back re the interaction with the test vectors and the new type constants, as well as some of the forward compat unit tests.

Will look into that tomorrow, but I think for now this is ready for a re-review.

asset/records.go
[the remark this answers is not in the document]
jharveyb: Makes sense 👍🏽

asset/asset.go
jharveyb: Related, a useful itest could be that, if a proof is created and then extended with an unknown odd field, checking that a client could successfully store that proof (via local Universe, DB, or flat file), and then return the identical proof to another client.

internal/test/generic.go
```diff
@@ -0,0 +1,59 @@
...
+// function returns the expected item and unknown types.
+func RunUnknownOddTypeTest[T any](t *testing.T, knownItem T,
+	unknownTypeErr error, encode func(*bytes.Buffer, T) error,
+	decode func(*bytes.Buffer) (T, error), verify func(T, tlv.TypeMap)) {
+
+	var buf bytes.Buffer
+	err := encode(&buf, knownItem)
+	require.NoError(t, err)
+
+	// With the known asset now encoded, we can add an unknown even type to
```
jharveyb, line 25: nit: asset -> item

internal/test/generic.go
```diff
@@ -0,0 +1,59 @@
...
+	// the encoded bytes. That should provoke an error when parsed again.
+	unknownTypeValue := []byte("I could be anything, really")
+	unknownEvenType := append([]byte{
+		byte(40),                    // Type 40 is unknown.
+		byte(len(unknownTypeValue)), // Length of the value.
+	}, unknownTypeValue...)
+	buf.Write(unknownEvenType)
+
+	// Try to parse it again.
+	var parsedAsset T
```
jharveyb, line 35: Linter complains as the output of this `decode()` is not actually being used. The declaration of `parsedAsset` (should be parsedItem?) could be moved down to the last `decode()` call.

asset/records_test.go
```diff
@@ -0,0 +1,164 @@
...
+				t, knownAssetLeaf.NodeSum(),
+				parsedAssetLeaf.NodeSum(),
+			)
+			require.NotEqual(
+				t, knownAssetLeaf.NodeHash(),
+				parsedAssetLeaf.NodeHash(),
+			)
+
+			parsedAsset.unknownOddTypes = nil
+			knownAsset.GroupKey = nil
```
jharveyb, line 159: Why clear the group key here? `Asset.deepEqual()` should be able to handle it correctly IIRC.

address/address_test.go
```diff
@@ -544,3 +555,42 @@ func FuzzAddressDecode(f *testing.F) {
...
+			// The address should've changed, to make sure the
+			// unknown value was taken into account when creating
+			// the serialized address.
+			parsedAddrString, err := parsedAddr.EncodeAddress()
+			require.NoError(t, err)
+
+			require.NotEqual(t, knownAddrString, parsedAddrString)
+
+			parsedAddr.unknownOddTypes = nil
+			parsedAddr.assetGen = asset.Genesis{}
```
jharveyb, line 591: Same Q, why does the genesis need to be cleared for the comparison? Explanatory comment would be useful if it is needed.

jharveyb (commented):
Looks good! Left some comment in existing threads on itest behavior we may want to get better assurances on forwards-compat.

Some of the unit tests already extend data from the BIP test vectors, so I'm not sure if we want/need to add data with unknown types as test vectors themselves.

guggero:
I've added test vectors and an integration test.

itest/addrs_test.go
```diff
@@ -661,6 +663,179 @@ func runMultiSendTest(ctxt context.Context, t *harnessTest, alice,
...
+	lastProof.UnknownOddTypes = proofCustomTypes
+	lastProof.InclusionProof.UnknownOddTypes = tlv.TypeMap{
+		345: []byte("it's included"),
+	}
+	cp := lastProof.InclusionProof.CommitmentProof
+	cp.UnknownOddTypes = tlv.TypeMap{
+		567: []byte("it's committed"),
+	}
+	cp.TaprootAssetProof.UnknownOddTypes = tlv.TypeMap{
+		789: []byte("there's assets in here..."),
```
jharveyb, line 752: 😆

itest/addrs_test.go
```diff
@@ -661,6 +663,179 @@ func runMultiSendTest(ctxt context.Context, t *harnessTest, alice,
...
+
+	lastProof2, err := f2.LastProof()
+	require.NoError(t.t, err)
+
+	// If the contents are identical to what we uploaded, we just need to
+	// check a single value in the proof to be sure all the custom types are
+	// there.
+	require.Equal(t.t, modifiedBlob, transferProof2.RawProofFile)
+	require.True(t.t, bytes.Contains(modifiedBlob, []byte("it's included")))
+	require.True(
```
jharveyb, line 792: 💯

jharveyb (approved):
Awesome! With the itest, I'm convinced that this should work out.

Since that includes proof decode with an unknown type, I don't think there's particular value on running this over the proofs of our testnet and mainnet universes. IIUC any unknown types should have been dropped when those servers were storing the proofs anyway.

Roasbeef (approved):
> Reviewed 18 of 18 files at r1, 28 of 28 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1037)* status: all files reviewed, 23 unresolved discussions (waiting on @guggero and @jharveyb)

___
*[`commitment/commitment_test.go` line 1707 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/1037#-O4SUuXQ9a1F-XSIK_yk:-O4SUuXQ9a1F-XSIK_yl:b-nutwqo) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/d06206e0e88e8ad71df0a9c24af1fa54ccc51122/commitment/commitment_test.go#L1707)):*
> ```Go
> 
> 	var knownProofBytes []byte
> 	test.RunUnknownOddTypeTest(
> ```

Nice generalization :+1:

Issue #1038, opened 2024-07-22
[bug]: FundChannel Ack Timeout Issue


vanditshah99:
We tried the following scenario

We minted an asset on Tapd Standalone Instance (T1) 
```
{
    "version": "0.4.0-alpha.rc4 commit=v0.4.0-rc4",
    "lnd_version": "0.18.1-beta",
    "network": "testnet3",
    "lnd_identity_pubkey": "02ac55f686b46ae6fc60924fe2115b1a7b64d4d71c0de2265a1ef0136d680b1c61",
    "node_alias": "SpeedDev2",
    "block_height": 2869499,
    "block_hash": "00000000000000076bad3bf7db76e38f1842c50b3b32bd104bab93b056c778c6",
    "sync_to_chain": true
}
```
And transferred few Bucks to LiT-Tapd Instance (T2)

Now from T2 we raise an fundchannel request
```
litcli -n testnet --rpcserver=localhost:8443 --macaroonpath=~/.lnd/data/chain/bitcoin/testnet/admin.macaroon ln fundchannel --node_key=02ac55f686b46ae6fc60924fe2115b1a7b64d4d71c0de2265a1ef0136d680b1c61 --asset_amount=50000000 --asset_id=f8ffc554dad8a3780c408297a70b70af1c8d96cae3c167ff55cbd6308edb2a08 --push_amt=50000
```
This fails stating
```
[litcli] error funding channel: rpc error: code = Unknown desc = error funding channel: didn't receive funding ack after 30s
```
The logs of t1 are as follows
[t1 tapd node.txt](https://github.com/user-attachments/files/16332009/t1.tapd.node.txt)
[t1 lnd node.txt](https://github.com/user-attachments/files/16332006/t1.lnd.node.txt)

The logs of T2 node are following
[t2_ack.txt](https://github.com/user-attachments/files/16331983/t2_ack.txt)

Also, when this error is encountered, we face this issue of asset not showing in asset list grpc.
```
https://mempool.space/testnet/tx/6718cd84a2a5104b5ce7e1f6765750c1a6bf3f885248a9cae4bb5a71ab239e17#vout=0

```
5 conf already at this moment

Also litcli getinfo(T2 Node)
```
{
    "version": "0.13.99-alpharc4 commit=v0.13.99-alpha.rc4"
}
```

Roasbeef:
Working on adding a fast fail check, so we just bail out if the other nodes doesn't support the channel type sooner with a nice error message.

[...]

PR #1041, opened 2024-07-22
tapchannel: check for feature bits before opening chans

Roasbeef:
With the way the protocol works as is, we signal the optional feature bit, so connections can be made with older peers. Rather than timeout after they don't recognize our custom message, we'll instead use the feature bits to detect early if they don't understand the new protocol.

Fixes https://github.com/lightninglabs/taproot-assets/issues/1038

### DIFF
--- a/feature_bit_verifier.go
+++ b/feature_bit_verifier.go
@@ -1,0 +1,54 @@
+package taprootassets
+
+import (
+	"context"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightninglabs/taproot-assets/tapchannel"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// LndFeatureBitVerifier is a struct that verifies that the feature bits of a
+// target connected peer, using our registered lnd node.
+type LndFeatureBitVerifier struct {
+	lnd *lndclient.LndServices
+}
+
+// NewLndFeatureBitVerifier creates a new LndFeatureBitVerifier instance.
+func NewLndFeatureBitVerifier(
+	lnd *lndclient.LndServices) *LndFeatureBitVerifier {
+
+	return &LndFeatureBitVerifier{
+		lnd: lnd,
+	}
+}
+
+// HasFeature returns true if the peer has the given feature bit set. If the
+// peer can't be found, then ErrNoPeer is returned.
+func (l *LndFeatureBitVerifier) HasFeature(ctx context.Context,
+	peerPub btcec.PublicKey, bit lnwire.FeatureBit) (bool, error) {
+
+	peerBytes := route.NewVertex(&peerPub)
+
+	peers, err := l.lnd.Client.ListPeers(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	for _, peer := range peers {
+		if peer.Pubkey != peerBytes {
+			continue
+		}
+
+		return peer.Features.HasFeature(bit), nil
+	}
+
+	// If we get to this point, we weren't able to find the peer.
+	return false, tapchannel.ErrNoPeer
+}
+
+// A compile-time check to ensure that LndFeatureBitVerifier implements the
+// FeatureBitVerifier interface.
+var _ tapchannel.FeatureBitVerifer = (*LndFeatureBitVerifier)(nil)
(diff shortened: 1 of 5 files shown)


feature_bit_verifier.go
```diff
@@ -0,0 +1,54 @@
...
+	for _, peer := range peers {
+		if peer.Pubkey != peerBytes {
+			continue
+		}
+
+		return peer.Features.HasFeature(bit), nil
+	}
+
+	// If we get to this point, we weren't able to find the peer.
+	return false, tapchannel.ErrNoPeer
```
guggero, line 49: Seems weird to be returning an error from a package using this code... I know it's probably unavoidable with the current way the packages are depending on each other. But we should likely move all the lnd client related code into its own package.

tapchannel/aux_funding_controller.go
```diff
@@ -200,6 +212,10 @@ type FundingControllerCfg struct {
 	// AssetSyncer is used to ensure that we've already verified the asset
 	// genesis for any assets used within channels.
 	AssetSyncer AssetSyncer
+
+	// FeatureBits is used to verify that the peer has the required feature
+	// to fund asset channels.
+	FeatureBits FeatureBitVerifer
```
guggero, line 218: This needs to be hooked up:

```diff
diff --git a/tapcfg/server.go b/tapcfg/server.go
index 302a5982..5adc8648 100644
--- a/tapcfg/server.go
+++ b/tapcfg/server.go
@@ -102,6 +102,7 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
        msgTransportClient := tap.NewLndMsgTransportClient(lndServices)
        lndRouterClient := tap.NewLndRouterClient(lndServices)
        lndInvoicesClient := tap.NewLndInvoicesClient(lndServices)
+       lndFeatureBitsVerifier := tap.NewLndFeatureBitVerifier(lndServices)
 
        uniDB := tapdb.NewTransactionExecutor(
                db, func(tx *sql.Tx) tapdb.BaseUniverseStore {
@@ -434,6 +435,7 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
                        TxSender:           chainPorter,
                        DefaultCourierAddr: proofCourierAddr,
                        AssetSyncer:        addrBook,
+                       FeatureBits:        lndFeatureBitsVerifier,
                },
        )
        auxTrafficShaper := tapchannel.NewAuxTrafficShaper(

```

feature_bit_verifier.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// LndFeatureBitVerifier is a struct that verifies that the feature bits of a
+// target connected peer, using our registered lnd node.
+type LndFeatureBitVerifier struct {
+	lnd *lndclient.LndServices
+}
+
+// NewLndFeatureBitVerifier creates a new LndFeatureBitVerifier instance.
+func NewLndFeatureBitVerifier(lnd *lndclient.LndServices,
```
guggero, line 20: nit: break before argument instead?

guggero (commented):
tACK, once the feature bit verifier is updated this works as advertised.

feature_bit_verifier.go
Roasbeef: Yeah not a bad idea to move them. 

Re returning the error from another package: I think it can be a useful pattern (assuming we use a new type for the errors) to restrict the type of errors that an implementation can return.

feature_bit_verifier.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// HasFeature returns true if the peer has the given feature bit set. If the
+// peer can't be found, then ErrNoPeer is returned.
+func (l *LndFeatureBitVerifier) HasFeature(ctx context.Context,
+	peerPub btcec.PublicKey, bit lnwire.FeatureBit) (bool, error) {
+
+	peerBytes := route.NewVertex(&peerPub)
+
+	peers, err := l.lnd.Client.ListPeers(ctx)
```
jharveyb, line 35: Would `GetNodeInfo` also work here? IIUC the info returned from that call should be as up-to-date and what's returned from `ListPeers`, for connected nodes. But AFAICT it doesn't show if we're actually _connected_ to that node.

Roasbeef: So each time we connect to a peer, we'll get the fresh set of feature bits from the `init` message. Compared to waiting for the next `node_announcement`, this _should_ be more up to date.

PR #1044, opened 2024-07-23
Pass 'pkg' argument to `flake-unit-race` Makefile target

gijswijs:
This PR changes the behavior of the target `flake-unit-race` in the Makefile, to allow to run it for a specific package continuously.

Usage example: 'make flake-unit-race pkg=tapgarden'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -226,7 +226,7 @@ flake-unit-trace:
 
 flake-unit-race:
 	@$(call print, "Flake hunting races in unit tests.")
-	while [ $$? -eq 0 ]; do env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOLIST) | $(XARGS) env $(GOTEST) -race -test.timeout=20m -count=1; done
+	while [ $$? -eq 0 ]; do $(GOLIST) | $(XARGS) env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOTEST) -race -test.timeout=20m -count=1; done
 
 # =============
 # FUZZING
--- a/make/testing_flags.mk
+++ b/make/testing_flags.mk
@@ -4,10 +4,11 @@ RPC_TAGS = autopilotrpc chainrpc invoicesrpc peersrpc routerrpc signrpc verrpc w
 LOG_TAGS =
 TEST_FLAGS =
 ITEST_FLAGS = -logoutput
-COVER_PKG = $$(go list -deps -tags="$(DEV_TAGS)" ./... | grep '$(PKG)' | grep -v lnrpc)
-RACE_PKG = go list -deps -tags="$(DEV_TAGS)" ./... | grep '$(PKG)'
 COVER_HTML = go tool cover -html=coverage.txt -o coverage.html
 POSTGRES_START_DELAY = 5
+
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -226,7 +226,7 @@ flake-unit-trace:
 
 flake-unit-race:
 	@$(call print, "Flake hunting races in unit tests.")
-	while [ $$? -eq 0 ]; do env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOLIST) | $(XARGS) env $(GOTEST) -race -test.timeout=20m -count=1; done
+	while [ $$? -eq 0 ]; do $(RACE_PKG) | $(XARGS) env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOTEST) -race -test.timeout=20m -count=1; done
```
guggero, line 229: The variable name `RACE_PKG` threw me off here... Since it doesn't actually contain the package(s) but the command to list them. Should we rename it perhaps?

guggero (approved):
tACK, LGTM :tada:

make/testing_flags.mk
```diff
@@ -20,7 +20,7 @@ ifneq ($(pkg),)
 UNITPKG := $(PKG)/$(pkg)
 UNIT_TARGETED = yes
 COVER_PKG = $(PKG)/$(pkg)
-RACE_PKG = $(PKG)/$(pkg)
+RACE_PKG = echo $(PKG)/$(pkg)
```
jharveyb, line 23: Was the echo needed here or a leftover?

Makefile
jharveyb: IIUC it's a separate name / var since we replace that list with one entry when we specify the package?

So we have the listing from `GOLIST` duplicated in `RACE_PKG` and `COVER_PKG`.

make/testing_flags.mk
gijswijs: Needed, we need it so that we can pipe the result into the next bash command, since that's what we did with `GOLIST` as well.

Makefile
gijswijs: Yeah, and we don't use `RACE_PKG` or `COVER_PKG` anywhere, nor do we have those variables in `lnd`. So I'm not sure why we have them. So I thought, let's repurpose one of them.

On second thought, what would be the better approach is to adjust `GOLIST` here (similar to `RACE_PKG`):

https://github.com/lightninglabs/taproot-assets/blob/9e92b52aef3362cbc1ddd9ccea7b7e82e719ff57/make/testing_flags.mk#L19-L25

gijswijs: But `GOLIST` does also contains the *command* to list the packages. It's weird, because the assignment is with `:=`, but if I run the Makefile target I see this in the cli:

```
gijs@gijs-Legion:~/git/taproot-assets$ make flake-unit-race
 Flake hunting races in unit tests.
while [ $? -eq 0 ]; do go list -deps -tags="dev monitoring" ./... | grep 'github.com/lightninglabs/taproot-assets' | xargs -L 1 env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" GOEXPERIMENT=loopvar GO111MODULE=on go test -race -test.timeout=20m -count=1; done
```

`:=` doesn't seem to do what I think it does.

guggero (approved):
Nice, I like the new approach. re-ACK :tada:

Roasbeef (approved):
LGTM

PR #1047, opened 2024-07-24
multi: remove RFQ and Taproot Asset Channel functionality for production build of lightning-terminal

guggero:
We remove any experimental functionality in order for tapd being compile time compatible with lnd 0.18.2-beta for integration into a production build of lightning-terminal.

NOTE: This PR is being merged into the `v0-4-0-lnd-18` branch, which will only be used for integration with the upcoming `lightning-terminal v0.13.3-alpha`.


ffranr (approved):
The diff doesn't seem minimal to me, but I guess that doesn't matter?

How did you arrive at this diff? I would assume you just needed to set LND's version minimally remove code until it builds, and then disable tap channel support at the interfaces.

guggero:
> The diff doesn't seem minimal to me, but I guess that doesn't matter?

Yeah, the diff isn't small. But IMO it's minimal in the number of lines it _changes_ as most of the diff is just cleanly deleting code. So IMO minimal in review effort.

> How did you arrive at this diff? I would assume you just needed to set LND's version minimally remove code until it builds, and then disable tap channel support at the interfaces.

That's pretty much what I did, yes :sweat_smile:

PR #1048, opened 2024-07-24
[custom channels]: add new RPC methods for asset channel specific calls

guggero:
Adds standalone RPC calls to the `tapchannelrpc` package to simplify usage for RPC users.
These RPCs mirror the functionality that is currently only available in the command line in `litd`:
 - `litcli ln addinvoice` -> `tapchannelrpc.AddInvoice`
 - `litcli ln payinvoice` -> `tapchannelrpc.SendPayment`
 - `litcli ln sendpayment` -> `tapchannelrpc.SendPayment`


guggero:
Ready for review. Will do manual and integration test validation in https://github.com/lightninglabs/lightning-terminal/pull/807 (since these new RPCs can't really be tested well in `tapd` standalone).

guggero:
Looking at the failure in the `itest` (https://github.com/lightninglabs/lightning-terminal/actions/runs/10148102664/job/28060046297?pr=807), I think we need to pass in the RFQ peer's public key to the new calls, so we know what peer to ask for a quote if there are multiple channels with the given asset.
So I'll need to update this PR slightly.

rpcserver.go
```diff
@@ -6841,3 +7185,133 @@ func getDecimalDisplayNonStrict(
...
+			assetBalances, func(c channelWithAsset) bool {
+				return c.channelInfo.PubKeyBytes == *peerPubKey
+			},
+		)
+	}
+
+	switch {
+	// If there are multiple asset channels for the same asset, we need to
+	// ask the user to specify the peer public key. Otherwise, we don't know
+	// who to ask for a quote.
```
GeorgeTsagk, line 7221: nit: we could convert this into a feature in next versions, the user may not want to pick a single peer here so instead we could ask for a quote from all peers and choose the "best" one

rpcserver.go
```diff
@@ -6841,3 +7185,133 @@ func getDecimalDisplayNonStrict(
...
+	case len(assetBalances) > 1 && peerPubKey == nil:
+		return nil, fmt.Errorf("multiple asset channels found for "+
+			"asset %s, please specify the peer pubkey", id.String())
+
+	// If the user specified a peer public key, and we still have multiple
+	// channels, it means we have multiple channels with the same asset and
+	// the same peer. So we just return the first channel.
+	case len(assetBalances) >= 1:
+		return &assetBalances[0], nil
```
GeorgeTsagk, line 7230: nit: the first channel may be depleted, need to think of rotation

rpcserver.go
```diff
@@ -6730,6 +6733,347 @@ func (r *rpcServer) EncodeCustomRecords(_ context.Context,
...
+		// Even if the user didn't specify the peer public key before,
+		// we definitely know it now. So let's make sure it's always
+		// set.
+		peerPubKey = &assetChan.channelInfo.PubKeyBytes
+
+		// TODO(guggero): This should actually be the max BTC amount
+		// (invoice amount plus fee limit) in milli-satoshi, not the
+		// asset amount. Need to change the whole RFQ API to do that
+		// though.
+		maxAssetAmount := assetChan.assetInfo.LocalBalance
```
GeorgeTsagk, line 6819: I don't get the TODO
Isn't this the asset local balance?

rpcserver.go
```diff
@@ -6730,6 +6733,347 @@ func (r *rpcServer) EncodeCustomRecords(_ context.Context,
...
+				"for keysend payment")
+		}
+
+		balances := []*rfqmsg.AssetBalance{
+			rfqmsg.NewAssetBalance(assetID, req.AssetAmount),
+		}
+		htlc := rfqmsg.NewHtlc(balances, fn.None[rfqmsg.ID]())
+
+		// We'll now map the HTLC struct into a set of TLV records,
+		// which we can then encode into the map format expected.
```
GeorgeTsagk, line 6906:
```suggestion
		// which we can then encode into the expected map format.
```

GeorgeTsagk (approved):
looks good!
have some Qs

rpcserver.go
```diff
@@ -6946,6 +6946,134 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+
+	// Now that we have the accepted quote, we know the amount in Satoshi
+	// that we need to pay. We can now update the invoice with this amount.
+	mSatPerUnit := acceptedQuote.AskPrice
+	iReq.ValueMsat = int64(req.AssetAmount * mSatPerUnit)
+
+	// The last step is to create a hop hint that includes the fake SCID of
+	// the quote, alongside the channel's routing policy.
+	channelID := assetChan.channelInfo.ChannelID
+	ourPolicy, err := r.getOurPolicy(ctx, channelID, peerPubKey.String())
```
ffranr, line 7044: Perhaps instead of `ourPolicy` in both variable name and method name, just `channelPolicy`? Because "their" policy isn't something we deal with. I have the same sort of thoughts around the `getOurPolicy` method name and doc.

ffranr (commented):
Would it be valuable to add itests which exercise the new endpoints in this PR? Just checking for example that the outgoing HTLC contains the correct fields.

rpcserver.go
ffranr: It is currently, but will the threshold will be lowered to match the payment rather than just the entire channel outgoing balance.

rpcserver.go
guggero: I agree. But that would be a larger overhaul of the whole RFQ system, so I'd rather do that in a follow-up PR.

rpcserver.go
guggero: At this level the actual channel balance doesn't matter. We only need the `channel` object to get the peer public key. We don't want to do balance checks at this level really, as that should be handled by the sending logic.

rpcserver.go
guggero: We set it to the asset balance because the unit is wrong (right now it's the maximum amount in _assets_, but the user doesn't know how many assets they need to send, that's the question the RFQ call should answer in the first place). We'll do some refactor and eventually this should be the maximum amount the user is willing to pay in **satoshis** (invoice amount plus the user-defined maximum routing fee they're willing to pay).

guggero:
> Would it be valuable to add itests which exercise the new endpoints in this PR? Just checking for example that the outgoing HTLC contains the correct fields.

As mentioned in the PR body, the full functionality test is done in the `litd` integration tests.
But I added minimal calls to the RPCs to make sure they're hooked up correctly.

Roasbeef (approved):
LGTM 🌛

Issue #1045, opened 2024-07-23
[feature]: Ensure Golang Version 1.22.3 for `make release` Command to Reproduce Release Manifest


ffranr:
**Description:**

The overall goal is to be able to reproduce a release manifest file. In order to reproduce the digest hashes in the manifest file of a given release, the correct Golang version must be used when building the binaries. Therefore, the `make release` command needs to assert the correct Golang version. If the Golang version is incorrect, the command should exit with an error message.

**Considerations and Modifications:**

- The makefile already includes `GO_VERSION = 1.22.3`.
- Pass the `GO_VERSION` to `./scripts/release.sh build-release` in `make release`.
- Modify `./scripts/release.sh` to ensure that the correct version of Golang is used.
- If the required Golang version isn't passed into `./scripts/release.sh`, the script should skip the version check.
- The linter should ensure that the golang version passed into `./scripts/release.sh build-release` in `make release` is consistent with the version used in `.github/workflows/release.yaml` (this check is already in place and does not require modifications as part of this issue).

**Acceptance Criteria:**

- Running `make release` should use Golang version 1.22.3 (current go version stipulated in the `Makefile`).
- If the Golang version is not 1.22.3, the script should exit with an error message.
- If the required Golang version isn't passed, the script should skip the version check.
- The release manifest file can be reproduced with identical digest hashes by using the correct Golang version.

ffranr:
Useful snippets:

### Modification to `Makefile`

```makefile
release:
	@$(call print, "Releasing tapd and tapcli binaries.")
	$(VERSION_CHECK)
	./scripts/release.sh build-release "$(VERSION_TAG)" "$(BUILD_SYSTEM)" "$(RELEASE_TAGS)" "$(RELEASE_LDFLAGS)" "$(GO_VERSION)"
```

### Modification to `./scripts/release.sh`

```bash
#!/bin/bash

REQUIRED_GO_VERSION=$6

if [ -n "$REQUIRED_GO_VERSION" ]; then
  # Check the current Golang version
  CURRENT_GO_VERSION=$(go version | grep -oP 'go\K[0-9]+\.[0-9]+\.[0-9]+')

  if [ "$CURRENT_GO_VERSION" != "$REQUIRED_GO_VERSION" ]; then
    echo "Error: Golang version $CURRENT_GO_VERSION found, but $REQUIRED_GO_VERSION is required."
    exit 1
  fi
fi

# Proceed with the rest of the script
```

[...]

PR #1051, opened 2024-07-26
makefile+scripts: specify Go version for building release binaries

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/1045

To ensure our release binaries are reproducible, they must be built with a specific version of Go. This commit sets the `GOTOOLCHAIN` environment variable when building release binaries, ensuring that the build command uses the targeted Go version.

The `GOTOOLCHAIN` environment variable is supported only in Go version 1.21 or newer. This commit verifies that the invoked Go version meets this requirement.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -118,11 +118,23 @@ function check_tag_correct() {
 
 # build_release builds the actual release binaries.
 #   arguments: <version-tag> <build-system(s)> <build-tags> <ldflags>
+#              <go-version>
 function build_release() {
   local tag=$1
   local sys=$2
   local buildtags=$3
   local ldflags=$4
+  local goversion=$5
+
+  # Check if the active Go version matches the specified Go version.
+  active_go_version=$(go version | awk '{print $3}' | sed 's/go//')
+  if [ "$active_go_version" != "$goversion" ]; then
+    echo "Error: active Go version ($active_go_version) does not match \
+required Go version ($goversion)."
+    exit 1
+  fi
+
+  echo "Building release for tag $tag with Go version $goversion"
 
   green " - Packaging vendor"
   go mod vendor
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,7 @@ release-install:
 release:
 	@$(call print, "Releasing tapd and tapcli binaries.")
 	$(VERSION_CHECK)
-	./scripts/release.sh build-release "$(VERSION_TAG)" "$(BUILD_SYSTEM)" "$(RELEASE_TAGS)" "$(RELEASE_LDFLAGS)"
+	./scripts/release.sh build-release "$(VERSION_TAG)" "$(BUILD_SYSTEM)" "$(RELEASE_TAGS)" "$(RELEASE_LDFLAGS)" "$(GO_VERSION)"
 
 release-tag:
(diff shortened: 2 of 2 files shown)


ffranr:
I'll add https://github.com/lightninglabs/taproot-assets/issues/1045#issuecomment-2245581043 in a different PR.

scripts/release.sh
```diff
@@ -117,12 +117,32 @@ function check_tag_correct() {
...
+
+  # We will use the `GOTOOLCHAIN` environment variable to specify the Go version
+  # to use by the Go build command. Only go1.21 and newer support this
+  # environment variable.
+  #
+  # Now we will check to ensure that the installed Go version is 1.21 or newer.
+  min_required_installed_go_version="1.21"
+  installed_go_version=$(go version | awk '{print $3}' | sed 's/go//')
+
+  if [ "$(printf '%s\n' "$min_required_installed_go_version" "$installed_go_version" | sort -V | head -n1)" = "$min_required_installed_go_version" ]; then
```
guggero, line 136: This line is very long. Perhaps we should extract the first part into a variable as well?

scripts/release.sh
```diff
@@ -166,9 +186,9 @@ function build_release() {
     mkdir "${dir}"
     pushd "${dir}"
 
-    green " - Building: ${os} ${arch} ${arm} with build tags '${buildtags}'"
-    env GOEXPERIMENT=loopvar CGO_ENABLED=0 GOOS="$os" GOARCH="$arch" GOARM=$arm GOAMD64="v1" go build -v -trimpath -ldflags="${ldflags}" -tags="${buildtags}" ${PKG}/cmd/tapd
-    env GOEXPERIMENT=loopvar CGO_ENABLED=0 GOOS="$os" GOARCH="$arch" GOARM=$arm GOAMD64="v1" go build -v -trimpath -ldflags="${ldflags}" -tags="${buildtags}" ${PKG}/cmd/tapcli
+    green " - Building: ${gotoolchain} ${os} ${arch} ${arm} with build tags '${buildtags}'"
```
guggero, line 189: I'm very skeptical of using the `GOTOOLCHAIN` approach for this specific problem. Reproducibility is quite a fragile and difficult thing to achieve (took us many attempts on the `lnd` side to get the initial script right).
And any external mechanism (that is fairly new as well, only introduced in Go 1.21) will make it super hard to debug. The interaction with the docker based build and/or macOS also isn't clear to me.

Also, what's even the point of reproducing a build's signature if you don't have to do any manual check/verification and just blindly let the script do everything automatically?

So because I think this might introduce more problems than it solves, can't we just simply do a **check** for the right version and error out if it doesn't match?

guggero (commented):
While the code itself looks good, I'm very skeptical of the approach... See inline comment.

scripts/release.sh
ffranr: Latest commit is just a check which returns an error if the incorrect Go version is used.

ffranr:
Latest commit is just a check which returns an error if the incorrect Go version is used.

guggero (approved):
tACK, LGTM :tada:

Roasbeef (approved):
LGTM 🧊

PR #1054, opened 2024-08-01
Add more prometheus metrics

GeorgeTsagk:
## Description

This PR adds any missing metrics for the tapd monitoring package based on this [comment](https://github.com/lightninglabs/taproot-assets/pull/737#discussion_r1688809081)

Specifically, in this PR we add:
- Asset proof collector: gather a vector for all proof sizes, with script key as the identifier
- DB collector: gathers db related data, currently only gathering the total db size


tapdb/assets_store.go
```diff
@@ -337,6 +337,9 @@ type ActiveAssetsStore interface {
 	// FetchAssetMetaForAsset fetches the asset meta for a given asset.
 	FetchAssetMetaForAsset(ctx context.Context,
 		assetID []byte) (sqlc.FetchAssetMetaForAssetRow, error)
+
+	// AssetsDBSize returns the total size of the taproot assets database.
+	AssetsDBSize(ctx context.Context) (int64, error)
```
Roasbeef, line 342: Perhaps we should move this to another interface? So like `Meta` which could also be extended to return information about the current database migration.

tapdb/sqlc/queries/metadata.sql
```diff
@@ -0,0 +1,2 @@
+-- name: AssetsDBSize :one
+SELECT pg_catalog.pg_database_size(current_database()) AS size;
```
Roasbeef, line 2: We'll need a solution that also works w/ sqlite. 

Got this from Claude (untested): 
```
SELECT page_count * page_size AS size_in_bytes 
FROM pragma_page_count(), pragma_page_size();
```

Roasbeef, line 2: Is this query precise enough? What if the database is multi tenant and `tapdb` is just one of the users?

monitoring/asset_proof_collector.go
```diff
@@ -0,0 +1,100 @@
...
+// NOTE: Part of the prometheus.Collector interface.
+func (a *assetProofCollector) Collect(ch chan<- prometheus.Metric) {
+	a.collectMx.Lock()
+	defer a.collectMx.Unlock()
+
+	ctxdb, cancel := context.WithTimeout(context.Background(), dbTimeout)
+	defer cancel()
+
+	// Fetch all proofs.
+	proofs, err := a.cfg.AssetStore.FetchAssetProofs(ctxdb)
```
Roasbeef, line 76: With the default scraping interval, I think this'll run every 60 seconds or so. Extrapolating out to say 100x the current scale, will this be fast enough?

We may want to consider a more precise query than what we have here.

monitoring/asset_proof_collector.go
```diff
@@ -0,0 +1,100 @@
...
+	for scriptKey, _ := range proofs {
+		key := fmt.Sprintf("%x", scriptKey)
+		proofMap[key] = a.proofsVec.WithLabelValues(key)
+	}
+
+	// We'll track the size of the proof blob.
+	for key, proof := range proofs {
+		key := fmt.Sprintf("%x", key)
+		a.proofsVec.WithLabelValues(key).
+			Set(float64(len(proof)))
```
Roasbeef, line 96: Looking at this section, I think the file system may be a better way to run this query. We can use `stat` directly, though unless we can somehow batch the system calls, that may also prove to be somewhat slow depending on the underlying virtualization set up.

Roasbeef: This could be useful: https://pkg.go.dev/io/fs#WalkDir

Roasbeef (changes requested):

monitoring/asset_proof_collector.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, errors.New("asset proof collector asset store is " +
+			"nil")
+	}
+
+	return &assetProofCollector{
+		cfg:      cfg,
+		registry: registry,
+		proofsVec: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: numAssetProofsMetric,
```
Roasbeef, line 45: This would be another good candidate for a histogram. We'd then be able to plot percentiles to find outliers.

monitoring/asset_proof_collector.go
```diff
@@ -0,0 +1,100 @@
...
+				Help: "Proofs of all assets",
+			},
+			[]string{"asset_name"},
```
Roasbeef, line 48: Asset name isn't unique, so I think we want the full asset ID here.

Roasbeef (changes requested):

tapdb/sqlc/queries/metadata.sql
GeorgeTsagk: the above recommendation from Claude seems to be working fine

GeorgeTsagk: Given that it's built-in I trust it to a certain degree. It seems to be accounting for every byte at the time of execution, including unused allocated space,  data marked as deleted etc

The query should return the same size for the db regardless of the user and what they have access to within the db

monitoring/asset_proof_collector.go
GeorgeTsagk: Agree

If we want to stick with taking a snapshot of **all** asset proofs and keep track of them, we don't have anything better currently. I believe we can stick to this for now, and our loadtests will indicate the next step.

Also, if we want to keep the loadtest behavior "random" (i.e send a random asset X times) we can't really know which asset is going to be used, so it kind of makes sense to keep track of everything.

If we change the loadtest behavior so that specific assets are being sent around in specific patterns, then we can change the above query and only track what we're interested in

monitoring/asset_proof_collector.go
GeorgeTsagk: I don't get how looking at the fs directly is feasible? proofs are stored in the db, and in the case of postgres backend that's a service running anywhere, and for sqlite the db could actually be in your file system, but still how can you pin-point which bytes refer to proofs?

monitoring/asset_proof_collector.go
GeorgeTsagk: this is the script key, which is unique

label outdated, needs change

tapdb/assets_store.go
```diff
@@ -378,29 +390,43 @@ type BatchedAssetStore interface {
 	BatchedTx[ActiveAssetsStore]
 }
 
+type BatchedMetaStore interface {
```
guggero, line 393: nit: missing Godoc.

tapcfg/server.go
```diff
@@ -94,7 +107,7 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 	tapdbAddrBook := tapdb.NewTapAddressBook(
 		addrBookDB, &tapChainParams, defaultClock,
 	)
-	assetStore := tapdb.NewAssetStore(assetDB, defaultClock)
+	assetStore := tapdb.NewAssetStore(assetDB, metaDB, defaultClock, dbType)
```
guggero, line 110: Seems like we need to add these new arguments in some unit tests as well, see CI failures.

tapcfg/server.go
```diff
@@ -42,18 +42,25 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
...
+		db     databaseBackend
+		dbType sqlc.BackendType
+	)
 
 	// Now that we know where the database will live, we'll go ahead and
 	// open up the default implementation of it.
-	var db databaseBackend
 	switch cfg.DatabaseBackend {
 	case DatabaseBackendSqlite:
+		dbType = sqlc.BackendTypeSqlite
```
guggero, line 55: :+1:

tapdb/assets_store.go
```diff
@@ -3267,6 +3293,42 @@ func (a *AssetStore) FetchAssetMetaForAsset(ctx context.Context,
 	return assetMeta, nil
 }
 
+// AssetsDBSize returns the total size of the taproot assets database.
+func (a *AssetStore) AssetsDBSize(ctx context.Context) (int64, error) {
+	var totalSize int64
+
+	readOpts := NewAssetStoreReadTx()
+	dbErr := a.metaDb.ExecTx(ctx, &readOpts, func(q MetaStore) error {
+
```
guggero, line 3302: nit: unnecessary empty line. Same below before the `if dbErr != nil {` check.

monitoring/asset_proof_collector.go
guggero: We also store the proofs as individual files, in addition to the database. See `FileArchiver` in `proof/archive.go`. There we have the script key in the filename (among other things) and the file size would directly correspond to the proof size.

Another alternative would be a DB query that just returns the script key and `length(proof_file)` instead of the full proof, which should already be quite a bit faster.

guggero (commented):
Looks pretty good. I agree that we should optimize the call that fetches the proof sizes and put it into a histogram. The rest seems good, will do a manual test once comments are addressed.

tapcfg/server.go
GeorgeTsagk: yeap the MetaDB was not instantiated there, added :+1:

monitoring/asset_proof_collector.go
GeorgeTsagk: I implemented the query that @guggero recommended, as it seemed more straight forward and was within explored grounds

tapdb/sqlc/queries/metadata.sql
Roasbeef: What I mean is that if we're also running `lnd` in the same postgres DB, will it return the total size, or just the size of the DB w/ the attached user?

guggero: `current_database()` would refer to the database/schema you're using. Assuming there would be a different one for `lnd` and `taproot-assets`, I think this should work.

tapdb/sqlc/queries/assets.sql
```diff
@@ -709,6 +709,19 @@ FROM asset_proofs
...
+WITH asset_info AS (
+    SELECT assets.asset_id, script_keys.tweaked_script_key
+    FROM assets
+    JOIN script_keys
+        ON assets.script_key_id = script_keys.script_key_id
+)
+SELECT asset_info.tweaked_script_key AS script_key, 
+LENGTH(asset_proofs.proof_file) AS proof_file_length
+FROM asset_proofs
+JOIN asset_info
```
guggero, line 722: Instead of joining with a CTE, I think we should just do a normal three-way join here between asset_proofs, assets and script_keys. IMO that's a bit easier to read.

monitoring/asset_proof_collector.go
```diff
@@ -0,0 +1,110 @@
...
+
+	if cfg.AssetStore == nil {
+		return nil, errors.New("asset proof collector asset store is " +
+			"nil")
+	}
+
+	return &assetProofCollector{
+		cfg:      cfg,
+		registry: registry,
+		proofsHistogram: prometheus.NewHistogram(
```
guggero, line 42: Shouldn't we use `newProofSizesHistogram` here? Seems like the values are different here as well for the buckets...

monitoring/db_collector.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	dbSizeMetric = "total_db_size"
+)
+
+// dbCollector is a Prometheus collector that exports metrics related to the
+// daemon's database.
+type dbCollector struct {
```
guggero, line 17: Do we really need two collector structs and files? IMO we could merge the asset proof and db collector into one (`dbCollector`) as both read things from the DB...

guggero (commented):
A couple of optimization suggestions. But this works, tested it by running the load test locally with one sqlite and one postgres backed instance.

tapdb/sqlc/queries/metadata.sql
guggero: Btw, I tested this locally where I have a single development timescale/postgres instance with different databases.
And I get distinct values for each of the following queries:

```
SELECT pg_catalog.pg_database_size('taprootassets') AS size;
--> 16802671

SELECT pg_catalog.pg_database_size('loop') AS size;
--> 8135535

SELECT pg_catalog.pg_database_size('pool') AS size;
--> 8135535

select current_database();
--> taprootassets
```

monitoring/asset_proof_collector.go
GeorgeTsagk: oh yes, will use that here too

guggero (approved):
tACK, LGTM :tada:

Roasbeef (approved):
LGTM 🧞‍♂️

PR #1055, opened 2024-08-01
Reattempt proof delivery on node restart

ffranr:
Some of the commits in this PR refactor the code for clarity and don't modify functionality.

The overall goal of this PR is to ensure that we don't try to deliver proofs which have already been delivered. And that we reattempt proof delivery on node restart.

Related to https://github.com/lightninglabs/taproot-assets/issues/1009

### DIFF
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -656,8 +656,10 @@ func (p *ChainPorter) transferReceiverProof(pkg *sendPackage) error {
 		}
 
 		if !shouldDeliverProof {
-			log.Debugf("Not delivering proof for output with "+
-				"script key %x", key.SerializeCompressed())
+			log.Debugf("Not delivering transfer ouput proof "+
+				"(proof_delivery_status=%v, script_key=%x)",
+				out.ProofDeliveryComplete,
+				key.SerializeCompressed())
 			return nil
 		}
 
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -148,7 +148,7 @@ func (p *ChainPorter) Start() error {
 
 		// Start the main chain porter goroutine.
 		p.Wg.Add(1)
-		go p.assetsPorter()
+		go p.mainEventLoop()
 
 		// Identify any pending parcels that need to be resumed and add
 		// them to the exportReqs channel so they can be processed by
@@ -239,10 +239,10 @@ func (p *ChainPorter) QueryParcels(ctx context.Context,
 	)
 }
 
-// assetsPorter is the main goroutine of the ChainPorter. This takes in incoming
+// mainEventLoop is the main goroutine of the ChainPorter. This takes a parcel
 // requests, and attempt to complete a transfer. A response is sent back to the
 // caller if a transfer can be completed. Otherwise, an error is returned.
-func (p *ChainPorter) assetsPorter() {
+func (p *ChainPorter) mainEventLoop() {
 	defer p.Wg.Done()
 
 	for {
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -171,20 +171,80 @@ func (p *ChainPorter) resumePendingParcels() error {
 		return err
 	}
 
+	// Return early if there are no pending parcels to resume.
+	if len(outboundParcels) == 0 {
+		log.Info("No pending parcels to resume")
+		return nil
+	}
+
+	log.Infof("Attempting to resume asset transfer for %d parcels",
+		len(outboundParcels))
+
 	// We resume delivery using the normal parcel delivery mechanism by
 	// converting the outbound parcels into pending parcels.
 	for idx := range outboundParcels {
 		outboundParcel := outboundParcels[idx]
-		log.Infof("Attempting to resume delivery for anchor_txid=%v",
-			outboundParcel.AnchorTx.TxHash().String())
+
+		pendingParcel := NewPendingParcel(outboundParcel)
+		reportPendingParcel(*pendingParcel)
 
 		// At this point the asset porter should be running. It should
 		// therefore pick up the pending parcels from the channel and
 		// attempt to deliver them.
-		p.outboundParcels <- NewPendingParcel(outboundParcel)
+		p.outboundParcels <- pendingParcel
 	}
 
 	return nil
+}
+
+// reportPendingParcel logs information about a pending parcel.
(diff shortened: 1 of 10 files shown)


tapfreighter/chain_porter.go
```diff
@@ -171,22 +171,76 @@ func (p *ChainPorter) resumePendingParcels() error {
 		return err
 	}
 
+	if len(outboundParcels) == 0 {
+		log.Infof("Attempting to resume asset transfer for %d parcels",
```
guggero, line 175: Was the `if len(...) == 0 {` check intentional here? Sounds like we might want to always log that?
Otherwise the log message might directly read "No pending parcels to resume"...

tapfreighter/chain_porter.go
```diff
@@ -171,22 +171,76 @@ func (p *ChainPorter) resumePendingParcels() error {
...
+	// Formulate a log entry for each proof delivery pending transfer output
+	// for the pending parcel.
+	var outputLogStrings []string
+
+	for idx := range outboundParcel.Outputs {
+		transferOut := outboundParcel.Outputs[idx]
+
+		// We are only interested in proof outputs that have not yet
+		// been delivered successfully (therefore: some and false).
+		if transferOut.ProofDeliveryComplete != fn.Some(false) {
```
guggero, line 209: Comparison with `fn.Some()` here. Should use `UnwrapOr()` here too.

tapfreighter/chain_porter.go
```diff
@@ -171,22 +171,76 @@ func (p *ChainPorter) resumePendingParcels() error {
...
 		// attempt to deliver them.
-		p.outboundParcels <- NewPendingParcel(outboundParcel)
+		p.outboundParcels <- pendingParcel
 	}
 
 	return nil
 }
 
+// reportPendingParcel logs information about a pending parcel.
+func (p *ChainPorter) reportPendingParcel(pendingParcel PendingParcel) {
```
guggero, line 197: This can be a function and not a method, since `p` is never actually used.

tapfreighter/chain_porter.go
```diff
@@ -171,22 +171,76 @@ func (p *ChainPorter) resumePendingParcels() error {
...
+				"proof_courier_addr=%s",
+			idx, skBytes, proofCourierAddr,
+		)
+		outputLogStrings = append(
+			outputLogStrings, outputLog,
+		)
+	}
+
+	log.Infof("Encountered pending parcel "+
+		"(anchor_txid=%v, count_undeliverd_proofs=%d)",
```
guggero, line 230: nit: typo in "undelivered".

guggero (commented):
Code looks good, nice work.

Just to clarify: There will be a follow-up PR that actually addresses the issue that a transfer isn't completed (e.g. a call to `ConfirmProofDelivery`) while the proof transfer wasn't successful?
Or was the goal to address this here (might need some re-ordering of actions)?

tapfreighter/interface.go
```diff
@@ -252,6 +252,12 @@ type TransferOutput struct {
 // ShouldDeliverProof returns true if a proof corresponding to the subject
 // transfer output should be delivered to a peer.
 func (out *TransferOutput) ShouldDeliverProof() (bool, error) {
+	// If the proof delivery is already complete, we don't need to deliver
+	// a proof.
+	if out.ProofDeliveryComplete == fn.Some(true) {
```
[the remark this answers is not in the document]
ffranr: `if out.ProofDeliveryComplete..UnwrapOr(false) {` is more complicated in my mind. I'm not sure if it's correct? Shouldn't it be:
```
if out.ProofDeliveryComplete.UnwrapOr(true) {
```
Because if `out.ProofDeliveryComplete` is `None` then we want the conditional to evaluate to true (so that `ShouldDeliverProof` returns false).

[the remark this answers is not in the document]
guggero: Yeah, depending on whether the desired outcome for `None` is true or false that should be used.
I agree that it isn't necessarily more readable. I'm just not sure if comparing values of `fn.Option` always returns the expected result. For instance if the value is not a comparable value, it would compare pointers instead of the content. So even though we use primitive/comparable values here, I think it's bad practice to do that.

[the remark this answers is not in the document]
ffranr: I was wrong, `if out.ProofDeliveryComplete..UnwrapOr(false) {` is correct because we're using `ShouldDeliverProof` to set the value of `ProofDeliveryComplete` and we only return false if `ProofDeliveryComplete` shows that the proof has been delivered. And not if `ProofDeliveryComplete` is none/unset.

ffranr:
> Code looks good, nice work.
> 
> Just to clarify: There will be a follow-up PR that actually addresses the issue that a transfer isn't completed (e.g. a call to `ConfirmProofDelivery`) while the proof transfer wasn't successful? Or was the goal to address this here (might need some re-ordering of actions)?

@guggero I will put up another PR for the `ConfirmProofDelivery` changes.

guggero (approved):
Nice, LGTM :tada:

tapfreighter/chain_porter.go
```diff
@@ -148,37 +150,103 @@ func (p *ChainPorter) Start() error {
...
 	})
 
 	return startErr
 }
 
+// resumePendingParcels attempts to resume delivery for any pending parcels that
+// were previously interrupted. This is done by querying the export log for any
+// pending parcels and adding them to the outboundParcels channel so they can be
+// processed by the main porter goroutine.
+func (p *ChainPorter) resumePendingParcels() error {
```
jharveyb, line 165: Non-blocking note: should this logic be in its own goroutine?

IIUC, right now this blocks `Start()` from returning. If there are multiple outbound parcels to resume, this will also block until the caretaker has handled all of them, as the `outboundParcels` channel is unbuffered.

I'm not sure if `Start()` being blocked on parcel resume would cause any issues right now though.

jharveyb (approved):
LGTM 💯 

Solid clarifications & progress to the next PR. The `if` blocks with `UnwrapOr` were a bit hard to read at first but the comments cleared that up.

tapfreighter/chain_porter.go
ffranr: Thanks for picking up on that. Good points. I think I'll put up a separate PR to run `resumePendingParcels` in a separate goroutine. I don't think this PR changes the sync problem you're outlining (I think it existed before this PR).

ffranr: https://github.com/lightninglabs/taproot-assets/pull/1068

PR #1064, opened 2024-08-06
Coveralls support

jharveyb:
First attempt to update CI so that coverage info for unit tests + itests is uploaded to Coveralls (and could be displayed for future PRs).

I looked at the setup for btcd:

https://github.com/btcsuite/btcd/blob/master/.github/workflows/go.yml

https://github.com/btcsuite/btcd/blob/b161cd6a199b4e35acec66afc5aad221f05fe1e3/Makefile#L109

I think we don't need to change directory into subfolders for unit tests since we don't have sub-packages in this repo?

For the itest coverage, I looked at the PR adding that for LND:

https://github.com/lightningnetwork/lnd/pull/7364

But the way itests are run there is different (via script), so I think the coverage output ends up in a different place.

When playing with this locally I couldn't get the itest coverage data to be written in the correct location, so I think that's the last missing piece.

### DIFF
--- a/make/testing_flags.mk
+++ b/make/testing_flags.mk
@@ -4,6 +4,8 @@ RPC_TAGS = autopilotrpc chainrpc invoicesrpc peersrpc routerrpc signrpc verrpc w
 LOG_TAGS =
 TEST_FLAGS =
 ITEST_FLAGS = -logoutput
+ITEST_COVERAGE =
+COLLECT_ITEST_COVERAGE =
 COVER_HTML = go tool cover -html=coverage.txt -o coverage.html
 POSTGRES_START_DELAY = 5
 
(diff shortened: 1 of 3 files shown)


make/testing_flags.mk
```diff
@@ -57,6 +59,12 @@ ifneq ($(tags),)
 DEV_TAGS += ${tags}
 endif
 
+# Enable integration test coverage (requires Go >= 1.20.0).
+ifneq ($(cover),)
+ITEST_COVERAGE = -cover
```
guggero, line 64: Things work a bit differently in our integration test as compared to `lnd`'s.
In `lnd` we compile a binary of the integration test (`go build -cover`), then run that binary multiple times in parallel, which generates the "mergeable" cover format.

Since we here run everything in a single invocation of `go test`, we can actually just specify `-coverprofile=coverage-itest.txt` and use that directly in `path-to-profile: coverage-itest.txt` of the coverage plugin.

So there's no need for the `COLLECT_ITEST_COVERAGE` step.

guggero (commented):
Nice! Needs a change to actually work, but great to have coverage being sent/completed now.

make/testing_flags.mk
jharveyb: From offline update, that change didn't work out.

I pulled the itest changes out of this PR and put them on the branch https://github.com/lightninglabs/taproot-assets/tree/coveralls_unit_itest so we can pick that up independently.

jharveyb:
Closing in favor of #1069

PR #1068, opened 2024-08-08
Optimize Parcel Resumption and Reduce Blocking in ChainPorter

ffranr:
This PR addresses this concern by @jharveyb : https://github.com/lightninglabs/taproot-assets/pull/1055#discussion_r1708183895

### Changes

- **Improved Goroutine Management:** The `WaitGroup.Done()` call has been repositioned to align closely with the corresponding `WaitGroup.Add(1)` call, making the code easier to read and reducing the risk of errors.
  
- **Buffered Channel for Outbound Parcels:** A buffer has been added to the `outboundParcels` channel, allowing the system to handle new parcels without being blocked by pending ones, which enhances overall performance.

- **Concurrent Resumption of Pending Parcels:** The `ChainPorter.resumePendingParcels` method is now executed in a separate goroutine, ensuring that resuming pending parcels does not delay the startup process.

### DIFF
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -132,7 +132,7 @@ func NewChainPorter(cfg *ChainPorterConfig) *ChainPorter {
 	)
 	return &ChainPorter{
 		cfg:             cfg,
-		outboundParcels: make(chan Parcel),
+		outboundParcels: make(chan Parcel, 10),
 		subscribers:     subscribers,
 		ContextGuard: &fn.ContextGuard{
 			DefaultTimeout: tapgarden.DefaultTimeout,
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -155,7 +155,15 @@ func (p *ChainPorter) Start() error {
 			p.mainEventLoop()
 		}()
 
-		startErr = p.resumePendingParcels()
+		// Resume any pending parcels in a new goroutine so that we
+		// don't delay returning from the Start method. Without this
+		// goroutine the resumePendingParcels method would block on the
+		// buffering constraint of the outboundParcels channel.
+		p.Wg.Add(1)
+		go func() {
+			defer p.Wg.Done()
+			startErr = p.resumePendingParcels()
+		}()
 	})
 
 	return startErr
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -150,7 +150,10 @@ func (p *ChainPorter) Start() error {
 
 		// Start the main chain porter goroutine.
 		p.Wg.Add(1)
-		go p.mainEventLoop()
+		go func() {
+			defer p.Wg.Done()
+			p.mainEventLoop()
+		}()
 
 		startErr = p.resumePendingParcels()
 	})
@@ -311,8 +314,6 @@ func (p *ChainPorter) QueryParcels(ctx context.Context,
 // requests, and attempt to complete a transfer. A response is sent back to the
 // caller if a transfer can be completed. Otherwise, an error is returned.
 func (p *ChainPorter) mainEventLoop() {
-	defer p.Wg.Done()
-
 	for {
 		select {
 		case outboundParcel := <-p.outboundParcels:


tapfreighter/chain_porter.go
```diff
@@ -132,7 +132,7 @@ func NewChainPorter(cfg *ChainPorterConfig) *ChainPorter {
 	)
 	return &ChainPorter{
 		cfg:             cfg,
-		outboundParcels: make(chan Parcel),
+		outboundParcels: make(chan Parcel, 10),
```
jharveyb, line 135: I think we want to leave this as unbuffered? Otherwise we would end up trying to start the state machine with multiple packets at once:

https://github.com/lightninglabs/taproot-assets/blob/1b5a4ef392dcad4906272b1ef805d7333f5d4080/tapfreighter/chain_porter.go#L318

tapfreighter/chain_porter.go
```diff
@@ -150,9 +150,20 @@ func (p *ChainPorter) Start() error {
...
 
-		startErr = p.resumePendingParcels()
+		// Resume any pending parcels in a new goroutine so that we
+		// don't delay returning from the Start method. Without this
+		// goroutine the resumePendingParcels method would block on the
+		// buffering constraint of the outboundParcels channel.
+		p.Wg.Add(1)
+		go func() {
+			defer p.Wg.Done()
+			startErr = p.resumePendingParcels()
```
jharveyb, line 165: I think `resumePendingParcels()` needs to be modified to receive Quit signals?

Like this case:

https://github.com/lightninglabs/taproot-assets/blob/1b5a4ef392dcad4906272b1ef805d7333f5d4080/tapfreighter/chain_porter.go#L330

But here:

https://github.com/lightninglabs/taproot-assets/blob/1b5a4ef392dcad4906272b1ef805d7333f5d4080/tapfreighter/chain_porter.go#L194

tapfreighter/chain_porter.go
```diff
@@ -311,8 +314,6 @@ func (p *ChainPorter) QueryParcels(ctx context.Context,
 // requests, and attempt to complete a transfer. A response is sent back to the
 // caller if a transfer can be completed. Otherwise, an error is returned.
 func (p *ChainPorter) mainEventLoop() {
-	defer p.Wg.Done()
```
guggero, line 314: We should add a `NOTE` comment to this method that it MUST be run as a goroutine.
Normally, the `defer p.Wg.Done()` at the start of a method indicates this to someone reading the code.

tapfreighter/chain_porter.go
guggero: Why is it even a problem if we delay startup by waiting for pending parcels? Anything that takes a while (e.g. proof transfer) will be done in a goroutine anyway. So I don't see a pressing reason to do things async here.

Also, if we do this in another goroutine, we don't need the buffered channel as already mentioned by @jharveyb. I think it just makes the behavior less deterministic (e.g. with 9 parcels the goroutine finishes almost immediately but with 11 it blocks until complete)...

jharveyb: IIUC the start of each 'subsystem' is sequential and blocking on any one `Start()` call would block the rest?

https://github.com/lightninglabs/taproot-assets/blob/a8d8b5a46552c98fe6ef54c796dffa17927b8e33/server.go#L185

Even with proof transfer in a goroutine, if we're resuming more than one parcel I think the transfer process for the first one would block resumption of the second? And these wouldn't happen in parallel.

The concrete case I was thinking of was: "If I have a wide transfer (with many recipients), what happens on restart?"

I think they would attempt to be transferred, sequentially, and any issues with proof upload at that point would block the caretaker startup. Maybe I'm wrong about which errors would cause which functions to block though.

guggero: We already spin up a goroutine for each individual parcel: https://github.com/lightninglabs/taproot-assets/blob/9ac1baacd996c20ce6b9c51fa77b4b84e0d3fe5d/tapfreighter/chain_porter.go#L337
So at startup, because the main event loop already is a goroutine and just starts another one for each parcel, we can feed in the parcels to resume synchronously and block on that.

jharveyb: Ah, so we can handle parcels in parallel then?

In that case, IIUC, the only startup delay would be calling `advanceState()` for each resumed parcel, which should be fast.

And then we actually don't need to modify the current behavior, and the problem I was thinking of doesn't exist.

I'm not sure if we have an existing test that handles multiple parcels at once, that would be good to have to validate this.

ffranr:
Closing, see https://github.com/lightninglabs/taproot-assets/pull/1068#discussion_r1714017168

jharveyb:
> Closing, see [#1068 (comment)](https://github.com/lightninglabs/taproot-assets/pull/1068#discussion_r1714017168)

It would still be good to have an itest that exercised having multiple transfers running concurrently. A simple way to get that would be just delaying mining a block; so submit a transfer, then a second transfer - both should end up waiting for the confirmation, but proceed as normal after a block is mined.

ffranr:
> > Closing, see [#1068 (comment)](https://github.com/lightninglabs/taproot-assets/pull/1068#discussion_r1714017168)
> 
> @ffranr
> 
> It would still be good to have an itest that exercised having multiple transfers running concurrently. A simple way to get that would be just delaying mining a block; so submit a transfer, then a second transfer - both should end up waiting for the confirmation, but proceed as normal after a block is mined.

@jharveyb good idea! I'll spin that into an issue.

ffranr:
issue: https://github.com/lightninglabs/taproot-assets/issues/1081

PR #1069, opened 2024-08-08
CI: push coverage file for unit tests

jharveyb:
Second attempt to update CI so that coverage info for unit tests is uploaded to Coveralls (and could be displayed for future PRs).

I looked at the setup for btcd:

https://github.com/btcsuite/btcd/blob/master/.github/workflows/go.yml

https://github.com/btcsuite/btcd/blob/b161cd6a199b4e35acec66afc5aad221f05fe1e3/Makefile#L109

I think we don't need to change directory into subfolders for unit tests since we don't have sub-packages in this repo?

Itest coverage may more involved so deferring that for a separate PR.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -232,6 +232,7 @@ jobs:
         if: matrix.unit_type == 'unit-cover'
         with:
           path-to-profile: coverage.txt
+          flag-name: 'unit'
           parallel: true
 
   ########################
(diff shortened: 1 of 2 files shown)


Roasbeef:
Perhaps we should base it off what `lnd` does since it's more updated? Eg: I don't see where in this PR we start to run `go test` with the relevanta coverage args (`-coverprofile`).

guggero:
> Perhaps we should base it off what `lnd` does since it's more updated? Eg: I don't see where in this PR we start to run `go test` with the relevanta coverage args (`-coverprofile`).

We still use `goacc` here and in `lnd`:
https://github.com/lightninglabs/taproot-assets/blob/b3a0095eb01181a3bc957986c8b0568398cd399d/make/testing_flags.mk#L116
https://github.com/lightningnetwork/lnd/blob/8a5160e1de6568699d2244fc63162b05febf380c/make/testing_flags.mk#L147

So coverage has always been collected, just not uploaded/finished correctly.

guggero (approved):
Nice! This now looks the same as in `lnd` and seems to be working (get a report from the bot on the PR).

GeorgeTsagk (approved):
Looks good!

PR #1074, opened 2024-08-09
Spend transaction change outputs even if undelivered proof(s)

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/1009

Refactor `ChainPorter` state machine such that change outputs can be spent even if proof delivery is incomplete.

### Changes

1. **Integration Test: Spend Change on Failed Proof Delivery**
   - A new integration test is added to ensure that a tapd node can spend a change output even if the proof transfer for the previous transaction fails.

2. **Renaming Methods and States for Clarity**
   - The `ConfirmParcelDelivery` method is renamed to `LogAnchorTxConfirm` to more accurately reflect its functionality. This method now focuses on updating the send package on disk to indicate the anchoring transaction's on-chain confirmation, without "confirming parcel delivery" to disk (it never did that anyway).
   - The `ChainPorter` state is renamed from `SendStateStoreProofs` to `SendStateStorePostAnchorTxConf`. This new name better represents the broader storage updates performed after the anchor transaction confirmation, beyond just proof storage.

3. **State Management Enhancements**
   - The `ChainPorter` state machine is updated to store the package state after the anchor transaction confirmation, but before initiating the proof transfer. This change ensures that the transfer change output coins remain spendable even if the proof delivery fails.
   - The anchor transaction block hash is now stored in the outbound package, providing additional context for the state of pending transfers.

4. **Database Query Generalization**
   - The `QueryAssetTransfers` SQL query argument `unconf_only` is generalized and renamed to `pending_transfers_only`. This updated flag now includes pending transfers with unconfirmed anchoring transactions or undelivered transfer output proofs. The query now also returns the anchor transaction block hash, allowing users to determine if the transfer's anchoring transaction has been confirmed on-chain.


ffranr:
`FAIL: TestTaprootAssetsDaemon/fee_estimation` just failed. That test might be flaky. I don't think it's related to the changes in this PR.

itest/send_test.go
```diff
@@ -951,6 +951,161 @@ func testReattemptFailedSendUniCourier(t *harnessTest) {
...
+	// asset proof transfer to be unsuccessful.
+	sendAssetsToAddr(t, sendTapd, recvAddr)
+	MineBlocks(t.t, t.lndHarness.Miner.Client, 1, 1)
+
+	// Ensure that the transaction took place as expected.
+	listTransfersResp, err := sendTapd.ListTransfers(
+		ctxb, &taprpc.ListTransfersRequest{},
+	)
+	require.NoError(t.t, err)
+	require.Len(t.t, listTransfersResp.Transfers, 1)
```
guggero, line 1074: Can we assert the transfer's TXID here? And that it has a block height and block hash set?
And that the output with the address is showing as not yet transferred? Might need to expose that status on the gRPC side.

itest/send_test.go
```diff
@@ -951,6 +951,161 @@ func testReattemptFailedSendUniCourier(t *harnessTest) {
...
+	// Ensure that the transaction took place as expected.
+	listTransfersResp, err = sendTapd.ListTransfers(
+		ctxb, &taprpc.ListTransfersRequest{},
+	)
+	require.NoError(t.t, err)
+	require.Len(t.t, listTransfersResp.Transfers, 2)
+
+	// Restart the proof courier service.
+	t.Log("Starting proof courier service")
+	require.NoError(t.t, proofCourier.Start(nil))
```
guggero, line 1106: After this also assert the transfer output status shows delivery complete?

guggero (commented):
Very nice!

tapdb/sqlc/queries/transfers.sql
```diff
@@ -42,20 +42,25 @@ SET proof_delivery_complete = @delivery_complete
 WHERE output_id = (SELECT output_id FROM target);
 
 -- name: QueryAssetTransfers :many
-SELECT
-    id, height_hint, txns.txid, transfer_time_unix
+SELECT DISTINCT
```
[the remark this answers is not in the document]
guggero: I'm just saying that whenever I see a `DISTINCT` in a query, from experience, it's a sign that most likely something is fishy with the query. We're joining over an additional table, even though we don't need values from that table, so it should probably only be used in a sub-query (or maybe a CTE would fit here as well).
Also, I could imagine `DISTINCT` not being super fast, but that's more of a gut feeling.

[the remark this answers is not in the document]
guggero: Also, I think we should have a basic unit tests covering the new query. I myself made multiple mistakes in queries just because I didn't add proper unit tests.

itest/send_test.go
ffranr: I would like to add a TODO for that. I think that if the proof transfer is stuck in the backoff procedure then it might need a new connection to the proof courier if the courier restarts. Or something like that. I will look into it, but I prefer not to hold up the PR on this.

In this PR, I would like to assert the delivery status of those proofs as pending and then show that the change can be spent.

Let me know if you disagree.

guggero: Sounds good to me. I think we should create a new PR that allows users to edit/update the delivery address of a transfer. In that PR we can assert the state of the outputs.

tapdb/sqlc/queries/transfers.sql
```diff
@@ -42,20 +42,25 @@ SET proof_delivery_complete = @delivery_complete
...
+        OR sqlc.narg('anchor_tx_hash') IS NULL)
 
--- Here we have another optional query clause to select a given transfer
--- based on the anchor_tx_hash, but only if it's specified.
-AND (txns.txid = sqlc.narg('anchor_tx_hash') OR
-    sqlc.narg('anchor_tx_hash') IS NULL)
+    -- Filter for pending transfers only if requested.
+    AND (
+        (@pending_transfers_only = true
+             AND (txns.block_hash IS NULL OR outputs.proof_delivery_complete = false))
```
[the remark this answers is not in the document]
ffranr: Can't use `COALESCE` in postgres with two different types. I propose `txns.block_hash IS NULL OR txns.block_hash = X'00'` as an alternative.

tapdb/sqlc/queries/transfers.sql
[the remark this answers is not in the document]
ffranr: Thinking about this a bit more, perhaps we want both `DISTINCT` and a better join? Here's my reasoning:

without `DISTINCT` we face two errors: duplicates and missing rows. But with `DISTINCT` we have to worry about missing rows. In other words, we do actually want the final rows to be `DISTINCT`.

As for performance, I think we need to optimise against an actual problem, and I don't think we've seen a performance problem just yet. And it's not clear to me that performance is actually significantly affected.

But I also see your point that the rest of the query has to remove duplicates also. Otherwise we might lose precision in what we want the statement to achieve.

Do you think using `DISTINCT` will hide problems though?

ffranr:
This PR also adds extra useful fields to the `ListTransfers` RPC endpoint response.

tapdb/sqlc/queries/transfers.sql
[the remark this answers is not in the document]
ffranr: OK. `txns.block_hash = X'00'` doesn't work in Postgres. And `txns.block_hash = '\x00'::bytea` doesn't work in sqlite. Any ideas on this? Are we trying to stop identify when `txns.block_hash` is an all zeros `[32]byte`? We should ensure that it isn't set instead?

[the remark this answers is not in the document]
jharveyb: Agree, since it's nullable I think we shouldn't need to be storing a blob of `0`?

tapdb/sqlc/queries/transfers.sql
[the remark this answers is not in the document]
jharveyb: I also don't understand what the `DISTINCT` is doing here - how would we end up with duplicate or missing rows for this query?

itest/send_test.go
```diff
@@ -951,6 +951,259 @@ func testReattemptFailedSendUniCourier(t *harnessTest) {
...
+	// attempt to transfer the asset proof should fail.
+	//
+	// We will soon start the asset transfer process. However, before we
+	// start, we subscribe to the send events from the sending tapd node so
+	// that we can be sure that a proof delivery has been attempted
+	// unsuccessfully.
+	events := SubscribeSendEvents(t.t, sendTapd)
+
+	wg.Add(1)
+	go func() {
```
jharveyb, line 1029: nit: this could be pulled into the `assertions` file if we're going to use it across multiple tests.

itest/send_test.go
```diff
@@ -951,6 +951,259 @@ func testReattemptFailedSendUniCourier(t *harnessTest) {
...
+	AssertAddrCreated(t.t, recvTapd, rpcAssets[0], recvAddr)
+
+	sendAssetsToAddr(t, sendTapd, recvAddr)
+	MineBlocks(t.t, t.lndHarness.Miner.Client, 1, 1)
+
+	// There may be a delay between mining the anchoring transaction and
+	// recognizing its on-chain confirmation. To handle this potential
+	// delay, we use require.Eventually to ensure the transfer details are
+	// correctly listed after confirmation.
+	require.Eventually(t.t, func() bool {
```
jharveyb, line 1135: nit: this could be pulled into the `assertions` file if we're going to use it across multiple tests.

non-blocking / could be in the follow-up PR.

itest/send_test.go
```diff
@@ -951,6 +951,259 @@ func testReattemptFailedSendUniCourier(t *harnessTest) {
...
+		require.NotEmpty(t.t, firstTransfer.AnchorTxBlockHash)
+
+		// Assert proof transfer status for each transfer output.
+		require.Len(t.t, firstTransfer.Outputs, 2)
+
+		// First output should have a proof delivery status of not
+		// applicable. This indicates that a proof will not be delivered
+		// for this output.
+		firstOutput := firstTransfer.Outputs[0]
+		require.Equal(
```
jharveyb, line 1156: 💯

jharveyb (commented):
Looks generally good! Seems like the follow-up PR to assert that restart works should be pretty small.

Only outstanding Q is on the DB query changes.

tapdb/sqlc/queries/transfers.sql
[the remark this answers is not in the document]
guggero: Yeah, I'm pretty sure we don't need the `DISTINCT` anymore with the sub query. Before it was needed to de-duplicate because we selected/joined transfers **and** outputs but were only interested in fields of the transfers, so duplicates were possible.
But now we only select over transfers (and chain transactions, which should be a 1:1 relationship). So there shouldn't be any duplicates.
And I'm not sure what you mean by "missing rows"? Missing from what table?

tapdb/sqlc/queries/transfers.sql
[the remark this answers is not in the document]
guggero: Sorry, I now see that my whole comment was super wrong... Even though I clearly even wrote `block_hash`, in my mind I was reading `block_height` the whole time... And because I recently did something in that area I knew that we basically treated `NULL` and `0` as the same there. But of course this doesn't make any sense for the **hash**.
Sorry, I'm all over the place lately, can't concentrate properly with the heat and lack of sleep...

tapdb/sqlc/queries/transfers.sql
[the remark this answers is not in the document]
ffranr: > And I'm not sure what you mean by "missing rows"? Missing from what table?

I mean that one error type is that the query doesn't return some rows that should be there. And that `DISTINCT` guarantees the absence of another error type: duplicate rows. It's nice to have that guarantee.

I agree that we don't **need** `DISTINCT` but we might as well have it because it is what we want. We want the returned rows from the query to be distinct.

Anyway, I'm going to remove `DISTINCT` for now because I don't think it matters very much. The JOIN has been modified as you suggested.

tapdb/sqlc/queries/transfers.sql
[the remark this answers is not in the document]
ffranr: No problem! :+1: Null check only it is.

tapfreighter/chain_porter.go
```diff
@@ -408,7 +408,16 @@ func (p *ChainPorter) waitForTransferTxConf(pkg *sendPackage) error {
 	case confEvent = <-confNtfn.Confirmed:
 		log.Debugf("Got chain confirmation: %v", confEvent.Tx.TxHash())
 		pkg.TransferTxConfEvent = confEvent
-		pkg.SendState = SendStateStoreProofs
+
+		// If the anchoring tx block hash is given, we'll also store it
+		// in the outbound package.
+		var anchorTxBlockHash fn.Option[chainhash.Hash]
```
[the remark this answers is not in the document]
ffranr: Nice, thanks!

itest/send_test.go
ffranr: I'm not sure if I agree with you that this should be pulled into `assertions`. All I'm doing here is calling the `ListTransfers` endpoint and then inspecting the results. And the results are confirmed anchor transactions. And for each transaction, one not applicable proof delivery status output and one delivery pending output.

Is that really a repeatable state that we can package up into an assertion? What would we even name that assertion?

itest/send_test.go
ffranr: I've managed to simplify this assertion a bit because we only wanted to detect a single event anyway. It's basically just a function call now. Let me know if you still think it should be moved into `assertions`.

tapdb/assets_store_test.go
```diff
@@ -2140,4 +2164,14 @@ func TestTransferOutputProofDeliveryStatus(t *testing.T) {
 	require.Equal(
 		t, sql.NullBool{}, transferOutputs[1].ProofDeliveryComplete,
 	)
+
+	// At this point the anchoring transaction has been confirmed on-chain
+	// and the proof delivery status shows complete for all applicable
+	// transfer outputs. Therefore, we should not be able to find any
+	// pending transfers.
+	assetTransfers, err = db.QueryAssetTransfers(ctx, TransferQuery{
```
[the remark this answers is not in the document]
ffranr: :+1: added!

itest/send_test.go
jharveyb: We have a similar helper func that inspects the output of `ListTransfers`:

https://github.com/lightninglabs/taproot-assets/blob/39c1f2786bd54526165a671078663b8ddfba913b/itest/assertions.go#L1060

https://github.com/lightninglabs/taproot-assets/blob/39c1f2786bd54526165a671078663b8ddfba913b/itest/assertions.go#L1095

The more flexible version we use for inspecting slices of assets:

https://github.com/lightninglabs/taproot-assets/blob/39c1f2786bd54526165a671078663b8ddfba913b/itest/assertions.go#L276

Which is called as part of larger assertions or helper funcs.

ffranr: OK, thanks for the example. I'll come back to this in a future PR.

tapdb/assets_store_test.go
```diff
@@ -2088,7 +2090,29 @@ func TestTransferOutputProofDeliveryStatus(t *testing.T) {
 
 	// At this point, we should be able to query for the log parcel, by
 	// looking for all unconfirmed transfers.
-	assetTransfers, err := db.QueryAssetTransfers(ctx, TransferQuery{})
+	assetTransfers, err := db.QueryAssetTransfers(ctx, TransferQuery{
```
jharveyb, line 2093: Didn't spot this the first time; I think we want to add assertions on the new `AnchorTxBlockHash` field for the asset transfers here?

IIUC before `ConfirmChainAnchorTx`, it should be len == 0. And therefore we should see the same transfer if  `PendingTransfersOnly` was true.

After the TX confirmation the block hash should match `randBlockHash`.

jharveyb (commented):
Updates look good! Just blocking on that one unit test update.

jharveyb (approved):
LGTM 👍🏽

taprpc/taprootassets.proto
```diff
@@ -627,6 +627,10 @@ message AssetTransfer {
 
     // Describes the set of newly created asset outputs.
     repeated TransferOutput outputs = 6;
+
+    // The block hash the contains the anchor transaction above. If unset, the
+    // anchor transaction is unconfirmed.
+    bytes anchor_tx_block_hash = 7;
```
guggero, line 633: Same request as here: https://github.com/lightninglabs/taproot-assets/pull/1075#discussion_r1713581013
Whenever we reference block or transaction hashes, I think we should also return the human-readable (byte-reversed) variant that can be copy/pasted easily from the command line. So just add an additional `anchor_tx_block_hash_str` that returns the `.String()` variant.

guggero (approved):
Very nice, LGTM :tada: 

One tiny request, but can also be deferred to follow-up PR, so non-blocking.

taprpc/taprootassets.proto
ffranr: OK, I'll add this in a future PR (you mention in the review that you're ok with that). I think I've got a good pattern in mind.

PR #1075, opened 2024-08-12
Expand VerifyOwnershipProof

sputn1ck:
This PR expands VerifyOwnershipProof by returning the outpoint, block height and blockhash the proof commits to.
This eases asset ownership verification, as clients can check if the outpoint has been spent.


taprpc/assetwalletrpc/assetwallet.proto
```diff
@@ -410,6 +410,15 @@ message VerifyAssetOwnershipRequest {
 
 message VerifyAssetOwnershipResponse {
     bool valid_proof = 1;
+
+    // The outpoint the proof commits to.
+    taprpc.OutPoint outpoint = 2;
+
+    // The block hash the output is part of.
+    bytes block_hash = 3;
```
guggero, line 418: Should we output this as `block_hash_str` as well that has the human-readable (byte-inversed) representation?

And maybe have `outpoint_str` as well that shows the string notation of the outpoint (`<txid>:<index>`)?
Just because the raw transaction hash bytes might confuse a lot of devs that aren't familiar with that specific quirk of bitcoin.

guggero (commented):
Very nice, thanks for the enhancement. One request concerning user/developer friendliness of the new fields.

guggero (approved):
LGTM :tada:

GeorgeTsagk (approved):
:raised_hands: :green_circle:

PR #1076, opened 2024-08-12
rfq: actually set subject asset in QueryRateTick

guggero:
Fixes the problem that because of the missing allocation, the subject asset ID was always set to an empty slice.

Reported by Luke in Slack, thanks a lot for pointing out the problem!

### DIFF
--- a/rfq/oracle.go
+++ b/rfq/oracle.go
@@ -221,11 +221,14 @@ func (r *RpcPriceOracle) QueryAskPrice(ctx context.Context,
 		return nil, fmt.Errorf("asset ID is nil")
 	}
 
-	// Construct query request.
-	var subjectAssetId []byte
-	copy(subjectAssetId, assetId[:])
+	var (
+		subjectAssetId = make([]byte, 32)
+		paymentAssetId = make([]byte, 32)
+	)
 
-	paymentAssetId := make([]byte, 32)
+	// The payment asset ID is BTC, so we leave it at all zeroes. We only
+	// set the subject asset ID.
+	copy(subjectAssetId, assetId[:])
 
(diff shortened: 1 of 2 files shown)


rfq/oracle_test.go
```diff
@@ -95,7 +117,7 @@ func runQueryAskPriceTest(t *testing.T, tc *testCaseQueryAskPrice) {
 	defer backendService.Stop()
 
 	// Wait for the server to start.
-	time.Sleep(2 * time.Second)
+	time.Sleep(200 * time.Millisecond)
```
dstadulis, line 120: Is it intended to reduce the time, rather than be equivalent?

guggero: The whole test took 14 seconds because of this wait, which is unnecessary (it's an in-memory server which should be spun up in a couple of milliseconds, even on slow systems).

rfq/oracle.go
```diff
@@ -221,11 +221,14 @@ func (r *RpcPriceOracle) QueryAskPrice(ctx context.Context,
...
-	copy(subjectAssetId, assetId[:])
+	var (
+		subjectAssetId = make([]byte, 32)
+		paymentAssetId = make([]byte, 32)
+	)
 
-	paymentAssetId := make([]byte, 32)
+	// The payment asset ID is BTC, so we leave it at all zeroes. We only
+	// set the subject asset ID.
+	copy(subjectAssetId, assetId[:])
```
jharveyb, line 231: super-nit: could be `fn.CopySlice(assetId[:])`.

jharveyb (approved):
LGTM 👍🏽

Issue #819, opened 2024-03-01
[feature]: ownership/prove & ownership/verify should accept random challenge


straumat:
**Is your feature request related to a problem? Please describe.**
On [Royllo](https://explorer.royllo.org), I have one page for each taproot assets. On that page I display all the information I could find about the asset. In the next release, peoples will be able to create a Royllo account and, there, they will be able to claim they created an asset.
After validating that the user is really the owner, Royllo will let the user manage the public royllo page of its asset  (add/remove content, add news, change alias...)

**Describe the solution you'd like**
For an asset, I should be able to generate a challenge that only the person who created the asset could reply.
I display this challenge on royllo.
The user should be able, on his node, to enter the challenge and get the correct reply.
He sends me the reply so I can be sure he is the owner.

(Like signmessage in bitcoin or k1 in lnurl auth)

**Describe alternatives you've considered**
I've tried to implement `wallet/ownership/prove` and `wallet/ownership/verify` but something is wrong.
Example: a user wanting to manage an asset page on Royllo will login on Royllo. There, he will find find an explanation : "call the `wallet/ownership/prove` method on your node, retrieve the proof_with_witness value and paste the proof_with_witness here".
On royllo back end, I call the decode and `wallet/ownership/verify` methods to check if the ownership method returns true.
But now that the user gave me its `proof_with_witness`, I can go anywhere else with it and pretend i'm the asset creator!
Nobody can challenge me

Roasbeef:
Given the above, here's a rough implementation plan:
  * Add a random challenge hash to the `prove` RPC call
  * When we go to generate the ownership proof, if the challenge hash is present, we should use that instead of the NUMs key here: https://github.com/lightninglabs/taproot-assets/blob/main/tappsbt/address.go#L139
      * We can modify to bind something like `NUMS + challenge*G`
  * Then `verify` will also accept tihs challenge hash as well.

With that you can create unique challenge hashes for a given context/action. It's still the verifies job to ensure that a challenge hash is never re-used.

straumat:
Yes!
I don't know yet all the details of taproot assets but from my point of view, I would like something like lnurl auth but I'm not sure exactly what is a scriptkey (it looks like a public key to me for now).

- Anyone having the issuance proof (bob) (retrieved from a universe server for example) can call `wallet/ownership/prove` with a parameter that is a random value he chose; The prove method returns a challenge (a bit like when lnurl auth generates an encrypted value of k1 with the user public key). 

[...]

PR #1077, opened 2024-08-12
Add challenge to proof ownership flows

GeorgeTsagk:
## Description

Currently, our `ProveAssetOwnership` / `VerifyAssetOwnership` calls create an ownership proof by creating a vPSBT that spends the asset to a NUMS key, and that challenge witness would be appended to the proof for the verifier to check.

In this PR we add a new `[32]byte challenge` field on both methods.
- `ProveAssetOwnership`: if a challenge is defined we instead spend it to `NUMS + challenge*G`, otherwise just `NUMS` (noop)
- `VerifyAssetOwnership`: if a challenge is defined we will construct a virtual tx that spends to `NUMS +challenge*G` and verify that state transition, otherwise just `NUMS` (noop)

## Comment

In the original issue there's the expectation to encode the challenge in the proof. Since the verifier is supposed to have knowledge of the challenge, we instead add the challenge as a parameter to the `VerifyAssetOwnership`. This way we don't have to change the `Proof` TLV encoders to accept an optional `[32]byte` challenge, which significantly reduces code pollution and minimizes diff.

Closes https://github.com/lightninglabs/taproot-assets/issues/819


GeorgeTsagk:
Needs rebase after merging https://github.com/lightninglabs/taproot-assets/pull/1075
Will with 1st round of fixes

address/address.go
```diff
@@ -517,3 +518,50 @@ func DecodeAddress(addr string, net *ChainParams) (*Tap, error) {
 
 	return &a, nil
 }
+
+// genChallengeNUMSAddress generates a NUMS address that is multiplied by the
```
guggero, line 522: nit: Godoc doesn't match function name.

guggero: Also, it's not an "address" that's being created but a script key.

address/address.go
```diff
@@ -517,3 +518,50 @@ func DecodeAddress(addr string, net *ChainParams) (*Tap, error) {
...
+	secp256k1.ScalarMultNonConst(&challenge, &g, &res)
+
+	// Calculate res = nums + res.
+	secp256k1.AddNonConst(&nums, &res, &res)
+
+	res.ToAffine()
+
+	resultPubKey := btcec.NewPublicKey(&res.X, &res.Y)
+
+	return asset.ScriptKey{
```
guggero, line 564: Should use `asset.NewScriptKey()` here to make sure it's a proper (even) Schnorr key.

address/address.go
```diff
@@ -517,3 +518,50 @@ func DecodeAddress(addr string, net *ChainParams) (*Tap, error) {
...
+		challenge    secp256k1.ModNScalar
+	)
+
+	if challengeBytesOpt.IsNone() {
+		return asset.NUMSScriptKey
+	}
+
+	var challengeBytes [32]byte
+
+	challengeBytesOpt.WhenSome(func(b [32]byte) {
```
guggero, line 539: We should also add a simple unit test to cover this function:

```
// TestGenChallengeNUMS tests the generation of NUMS challenges.
func TestGenChallengeNUMS(t *testing.T) {
	t.Parallel()

	gx, gy := secp256k1.Params().Gx, secp256k1.Params().Gy

	// addG is a helper function that adds G to the given public key.
	addG := func(p *btcec.PublicKey) *btcec.PublicKey {
		x, y := secp256k1.S256().Add(p.X(), p.Y(), gx, gy)
		var xFieldVal, yFieldVal secp256k1.FieldVal
		xFieldVal.SetByteSlice(x.Bytes())
		yFieldVal.SetByteSlice(y.Bytes())
		return btcec.NewPublicKey(&xFieldVal, &yFieldVal)
	}

	testCases := []struct {
		name        string
		challenge   fn.Option[[32]byte]
		expectedKey asset.ScriptKey
	}{
		{
			name:        "no challenge",
			challenge:   fn.None[[32]byte](),
			expectedKey: asset.NUMSScriptKey,
		},
		{
			name: "challenge is scalar 1",
			challenge: fn.Some([32]byte{
				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01,
			}),
			expectedKey: asset.NewScriptKey(addG(asset.NUMSPubKey)),
		},
		{
			name: "challenge is scalar 2",
			challenge: fn.Some([32]byte{
				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02,
			}),
			expectedKey: asset.NewScriptKey(
				addG(addG(asset.NUMSPubKey)),
			),
		},
	}

		result := GenChallengeNUMS(tc.challenge)
		require.Equal(
			t, tc.expectedKey.PubKey.SerializeCompressed(),
			result.PubKey.SerializeCompressed(),
		)
	}
}
```

proof/proof.go
```diff
@@ -296,6 +297,12 @@ type Proof struct {
 	// a valid signature to spend the asset.
 	ChallengeWitness wire.TxWitness
 
+	// ChallengeBytes is an optional field of the proof that is used when
+	// verifying an ownership proof. This fiels is only populated when the
+	// corresponding ProofVerificationOption option is defined. This is
+	// not encoded in the proof's records.
+	ChallengeBytes fn.Option[[32]byte]
```
guggero, line 304: I think we shouldn't transport this in the `Proof` struct, as that will be confusing since we don't encode it.

Instead we can just have a struct `proofVerificationParams` with the challenge bytes that is being set by the `ProofVerificationOption` function.

proof/verifier.go
```diff
@@ -511,7 +527,12 @@ type GroupAnchorVerifier func(gen *asset.Genesis,
 func (p *Proof) Verify(ctx context.Context, prev *AssetSnapshot,
 	headerVerifier HeaderVerifier, merkleVerifier MerkleVerifier,
 	groupVerifier GroupVerifier,
-	chainLookup asset.ChainLookup) (*AssetSnapshot, error) {
+	chainLookup asset.ChainLookup,
+	opts ...ProofVerificationOption) (*AssetSnapshot, error) {
+
+	for _, opt := range opts {
```
guggero, line 533: See comment above. This wouldn't actually modify the `Proof` struct but an instance of `proofVerificationParams` that stores the values. Then we can pass in `params.ChallengeBytes` below to `p.verifyChallengeWitness`.

rpcserver.go
```diff
@@ -5874,8 +5878,20 @@ func (r *rpcServer) ProveAssetOwnership(ctx context.Context,
 		return nil, fmt.Errorf("error fetching commitment: %w", err)
 	}
 
+	var challengeOpt fn.Option[[32]byte]
+
+	switch len(req.Challenge) {
+	case 0:
+		challengeOpt = fn.None[[32]byte]()
```
guggero, line 5885: No need to set `fn.None()`, that's the default value of a `var ... fn.Option[...]`.

taprpc/assetwalletrpc/assetwallet.proto
```diff
@@ -398,6 +398,10 @@ message ProveAssetOwnershipRequest {
     bytes script_key = 2;
 
     taprpc.OutPoint outpoint = 3;
+
+    // An optional 32-byte challenge that may be used to bind the generated
+    // proof.
```
guggero, line 403: Should we add something to the effect of "This challenge also needs to be presented when verifying the ownership proof, so it should be passed to the verifier, similar to how the message is needed when verifying a Bitcoin signature."?

itest/ownership_test.go
```diff
@@ -0,0 +1,121 @@
...
+	AssertNonInteractiveRecvComplete(t.t, bobTapd, 1)
+	AssertSendEventsComplete(t.t, bobAddr.ScriptKey, sendEvents)
+
+	// Now bob generates an ownership proof for the received assets. This
+	// proof does not contain a challenge.
+	proof, err := bobTapd.ProveAssetOwnership(
+		ctxb, &assetwalletrpc.ProveAssetOwnershipRequest{
+			AssetId:   rpcAssets[0].AssetGenesis.AssetId,
+			ScriptKey: bobAddr.ScriptKey,
+		})
```
guggero, line 62: nit: closing parenthesis, here and again further below.

itest/ownership_test.go
```diff
@@ -0,0 +1,121 @@
...
+	// Now alice edits a byte of the challenge. This challenge should not
+	// match the ownership proof, therefore a failure is expected.
+	ownershipChallenge[0] = 8
+	_, err = t.tapd.VerifyAssetOwnership(
+		ctxb, &assetwalletrpc.VerifyAssetOwnershipRequest{
+			ProofWithWitness: proof.ProofWithWitness,
+			Challenge:        ownershipChallenge[:],
+		},
+	)
+	require.Error(t.t, err)
```
guggero, line 110: Should we also match part what error we get? Something like `require.ErrorContains(t.t, err, "invalid challenge witness")` here and below?

guggero (commented):
Looks very good, thanks a lot!
Have a couple of improvements and nits, nothing major.

guggero (approved):
Very nice, LGTM :tada:

guggero:
Oh, linter is unhappy. A line is too long.

address/address.go
```diff
@@ -517,3 +518,48 @@ func DecodeAddress(addr string, net *ChainParams) (*Tap, error) {
...
+	var (
+		nums, g, res btcec.JacobianPoint
+		challenge    secp256k1.ModNScalar
+	)
+
+	if challengeBytesOpt.IsNone() {
+		return asset.NUMSScriptKey
+	}
+
+	var challengeBytes [32]byte
```
Roasbeef, line 537: Can can also further bind the challenge if we use `h(nums || challenge)` instead of just `challenge`.

PR #1080, opened 2024-08-13
linter: copy over Go version check scripts from LND

ffranr:
Replaced the existing Go version check scripts with those from the LND repository. These updated scripts are confirmed to work on macOS. The commit also relocates the scripts to the `scripts` directory from the `tools` directory.

The goal is to have the same scripts in both repositories.


guggero (approved):
Thanks, LGTM :tada:

jharveyb (approved):
LGTM 💯

Issue #1078, opened 2024-08-12
Increase test coverage for RPC endpoints


jharveyb:
Currently, we have itests that exercise the RPC endpoints, but not tests that check their safety; specifically that parsing gRPC `Request` types is panic-free and enforces the invariants that we require, like the size of certain fields or that only certain combinations of fields are valid.

One way to improve that is to introduce property-based tests for the initial logic in our RPC endpoints that handles request unmarshalling and sanity checking.

A useful tool for this is https://github.com/flyingmutant/rapid , which seems better-maintained and more capable than alternatives like https://pkg.go.dev/testing/quick . `rapid` tests can also be used as fuzz tests.

IMO the highest-priority endpoints are those related to `universerpc`, specifically those involved in sync operations as they would be exposed to public use by universe server operators. Others like `MintAsset` would also be good to cover since they involve may fields and invariants wrt. field combinations.

This may require minor refactoring of these endpoints to separate unmarshalling logic from the core logic of the endpoint itself. Also, due to import cycles for some logic like the `universerpc` unmarshalling functions, I think these tests would have to be put in the top-level scope / at the same point as `rpcserver` itself. That may require some changes to have these tests run as part of the existing `make unit` target.

jtobin:
I'm going to close this. I experimented with generating _tons_ of property tests some months ago, but (to my surprise) they didn't actually flush out any errors whatsoever.

In general I think the RPC endpoints are reasonably well-tested. More tests are always good (and property tests are especially good), but I don't think there's necessarily an issue here at present.

PR #1083, opened 2024-08-13
poc: rpc prop tests

jharveyb:
Starts to address #1078 .

I think this logic can probably be simplified a fair amount wrt. reducing the # of switches. I think adding wrapper types with their own validity-checking functions would also help for that; handling proto enums makes the test a lot larger, but I think most of the REST endpoints will have enum fields to support arguments as strings.

Rapid has some functions to autogenerate values via reflection: https://pkg.go.dev/pgregory.net/rapid#Make

I'm not sure if that will populate the gRPC private fields like `state`, `unknownFields`, etc. IIUC we don't want those to be populated.

This small test found some missing checks, which would _probably_ get rejected later on. Will peek at other RPCs with more fields to test out more complex invariants.


jharveyb:
I think this is in a decent state for now - this change to `rapid` made gRPC message generation much more ergonomic:

https://github.com/flyingmutant/rapid/pull/72

I'd prefer to block merging this until that PR (or something similar enough) lands in a `rapid` release. We could pull in the fixes in a separate PR if needed.

jtobin:
Closing as stale.

PR #1087, opened 2024-08-14
rpc: add ChainHash message with both bytes and string representations

ffranr:
This is a follow up PR for this review comment: https://github.com/lightninglabs/taproot-assets/pull/1074#discussion_r1716470965

This commit introduces the `ChainHash` RPC message, which contains two fields:
- A raw byte representation of the hash, ideal for direct use in Go RPC calls.
- A byte-reversed hexadecimal string representation of the hash, convenient for users who need to copy and paste the value, such as when passing it as a command-line argument.


ffranr:
@Roasbeef, I'm wondering if there's a more general location where we can define this type of RPC message, so it can be utilized across all our RPC subpackages. Any ideas on that?

taprpc/taprootassets.proto
```diff
@@ -611,6 +611,23 @@ message ListTransfersResponse {
...
 
+// BlockHash represents a hash of a Bitcoin (BTC) blockchain block.
+//
+// This message includes two separate fields for the same hash:
+// - The `hash` field provides the raw byte format, which is convenient for Go
+//   developers to use directly in other RPC calls.
+// - The `reversed_hash_hex` field provides the byte-reversed hash in
+//   hexadecimal string format, which is useful for users who need to copy and
+//   paste the value, such as passing it as a command-line argument.
+message BlockHash {
```
guggero, line 622: If we do this, it should actually be called `ChainHash`, as this representation also applies to transaction hashes.

I think this file is the correct place for it, so it can be included in all sub RPC packages.

taprpc/taprootassets.proto
```diff
@@ -611,6 +611,23 @@ message ListTransfersResponse {
...
+    // The block hash in raw byte format.
+    bytes hash = 1;
+
+    // The block hash as a hexadecimal string of the byte-reversed hash,
+    // commonly used in Bitcoin for human-readable representation.
+    string reversed_hash_hex = 2;
```
guggero, line 628: The convention for naming this would just be `hash_str` both in other projects and this one:
 - https://github.com/lightningnetwork/lnd/blob/5e84ba92afe4cd13d1360fb76eb751ec328ee1b5/lnrpc/lightning.proto#L1070
 - https://github.com/lightningnetwork/lnd/blob/5e84ba92afe4cd13d1360fb76eb751ec328ee1b5/lnrpc/lightning.proto#L1058
 - https://github.com/lightninglabs/taproot-assets/blob/faea364ddc0a1c9359fadfd9b9b3570ed7a727c7/taprpc/assetwalletrpc/assetwallet.proto#L424
 - https://github.com/lightninglabs/taproot-assets/blob/2f369531aab60b141a8ec8fb8205574ec6412a9f/taprpc/universerpc/universe.proto#L279

We should aim for some consistency across our repositories to make developers' lives easier. The comment above it should be explanation enough.

taprpc/taprootassets.proto
```diff
@@ -628,9 +645,10 @@ message AssetTransfer {
     // Describes the set of newly created asset outputs.
     repeated TransferOutput outputs = 6;
 
-    // The block hash the contains the anchor transaction above. If unset, the
-    // anchor transaction is unconfirmed.
-    bytes anchor_tx_block_hash = 7;
+    // The block hash of the blockchain block that contains the anchor
+    // transaction. If this value is unset, the anchor transaction is
+    // unconfirmed.
+    BlockHash anchor_tx_block_hash = 7;
```
guggero, line 651: If we want consistency across the project, I think we should use this for all occurrences of a block or transaction hash. But that would also mean a breaking change for developers...

I know you weigh the urge to "make things nice and clean and proper" quite high.
My tendency in general is more along the lines of "never break APIs" and "at least be consistent across APIs, even if it doesn't feel super clean"...
So my _slight_ tendency would be to just go with two fields here as well and not introduce a new message. Because that would just mean yet another pattern for developers to get used to (even if I agree that it's probably how things should've been from the start). And since we don't normally change any of our APIs (at least definitely in `lnd` that's the policy), we couldn't upgrade all other APIs to the new pattern...

guggero (commented):
While I like the idea, see my inline comment for why we _might_ not want to do it.
But happy to be overruled...

taprpc/taprootassets.proto
ffranr: The two fields still exist though, so I don't see this as breaking the two fields pattern. It just wraps them in a single message. Gives us a chance to document a bit better, and re-use for un/marshalling, and so on. I don't think we should break any existing fields (the field in the PR is an exception because it hasn't been released yet).

guggero: Yeah, okay, I guess you're right. Let's do it for this field and new fields.

guggero (approved):
LGTM :tada:

jharveyb (approved):
LGTM - wondering if we should swap this in in other relevant spots / RPC messages before the next major release?

Non-blocking.

guggero:
> LGTM - wondering if we should swap this in in other relevant spots / RPC messages before the next major release?

I think we don't want to change any APIs that have already been in a released version. So the field changed in this PR is the only one we can change without potentially breaking anyone's application.

But we should keep this in mind for future fields.

Issue #1071, opened 2024-08-08
[bug]: when trying to sync proofs we get either sqllite error or empty array 


starbackr-dev:
### Background

 Issue when trying to sync the proof for USDTL. 

tapcli universe sync --asset_id d75dae162be7cd0a2df76831ee1cce4eaeb74a32fabb0d46be7a0ad70fd6ad27  --universe_host universe.lightning.finance:10029
{
    "synced_universes":  []
}

### Your environment

mainnet

```sh
tapcli getinfo                       # version of `tapd`, `lnd`, and network
uname -mrsv                          # operating system 
bitcoind --version || btcd --version # version of `btcd`, `bitcoind`, or other backend
```
* any other relevant environment details

log files

2024-08-07 21:13:21.562 [INF] UNIV: Attempting to sync universe: host=universe.lightning.finance, sync_type=issuance, ids=([]universe.Identifier) (len=1 cap=1) {
 (universe.Identifier) issuance-d75dae162be7cd0a2df76831ee1cce4eaeb74a32fabb0d46be7a0ad70fd6ad27
}

2024-08-07 21:13:21.733 [INF] UNIV: Fetching 1 roots
2024-08-07 21:13:22.438 [INF] UNIV: Obtained 1 roots from remote Universe server
2024-08-07 21:13:22.446 [INF] UNIV: UniverseRoot(issuance-8e783d128da97b0c3fca189b7c04839c2f1dd7505b993ac7bc54f1bec22cec95) diverges, performing leaf diff...
2024-08-07 21:13:22.665 [INF] UNIV: UniverseRoot(issuance-8e783d128da97b0c3fca189b7c04839c2f1dd7505b993ac7bc54f1bec22cec95): diff_size=1
2024-08-07 21:13:22.787 [INF] UNIV: Verifying 1 new proofs for insertion into Universe
2024-08-07 21:13:22.807 [INF] UNIV: Inserting 1 verified group anchor proofs into Universe
2024-08-07 21:13:28.178 [ERR] RPCS: [/universerpc.Universe/SyncUniverse]: unable to sync universe: unable to register proofs: unable to register proofs: unable to register new group anchor issuance proofs: unable to upsert multiverse leaf: sql: Scan error on column index 0, name "id": converting NULL to int64 is unsupported

guggero:
What version are you running?

habibitcoin:
[...]

PR #1089, opened 2024-08-15
tapdb: fix multiple issues around migration file handling

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1071.

This PR is the result after almost two days of debugging #1071.
The TL;DR is: due to a git branch cherry-pick mishap, the migration number 14 wasn't included in the `v0.3.3` release version. Some users then applied the migration directly as it was in the source directory, which lead to them not having properly working auto incrementing primary keys.

The first commit in this PR adds a CI check that verifies migration numbers in the file names are continuous.

The second commit re-applies the migration 14 (somewhat modified) as migration 23.

The last commit then fixes the workaround with the primary key type replacements we used. See comment in `scripts/gen_sqlc_docker.sh` if you want to understand the full problem and solution.

### DIFF
--- a/scripts/check-migration-numbering.sh
+++ b/scripts/check-migration-numbering.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+# The directory containing migration files.
+migrations_path="tapdb/sqlc/migrations"
+
+# Check if the directory exists
+if [ ! -d "$migrations_path" ]; then
+  echo "Directory $migrations_path does not exist."
+  exit 1
+fi
+
+# Get all unique prefixes (e.g., 000001, always 6 digits) from .up.sql files.
+prefixes=($(ls "$migrations_path"/*.up.sql 2>/dev/null | \
+    sed -E 's/.*\/([0-9]{6})_.*\.up\.sql/\1/' | sort))
+
+# Check if no prefixes were found.
+if [ ${#prefixes[@]} -eq 0 ]; then
+  echo "No .up.sql migration files found in $migrations_path."
+  exit 1
+fi
+
+# Iterate over prefixes to ensure that there are no gaps and that each prefix
+# has a corresponding .down.sql file. Because the prefixes are sorted, and the
+# index starts at 0, we expect each prefix to be the index plus one.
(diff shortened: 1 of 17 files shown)


ffranr (approved):
Looks good :+1: I've tested the script locally on Linux.

ffranr:
Another fee estimation itest here:
```
--- FAIL: TestTaprootAssetsDaemon/fee_estimation (16.48s)
```

scripts/check-migration-numbering.sh
```diff
@@ -0,0 +1,43 @@
...
+migrations_path="tapdb/sqlc/migrations"
+
+# Check if the directory exists
+if [ ! -d "$migrations_path" ]; then
+  echo "Directory $migrations_path does not exist."
+  exit 1
+fi
+
+# Get all unique prefixes (e.g., 000001, always 6 digits) from .up.sql files.
+prefixes=($(ls "$migrations_path"/*.up.sql 2>/dev/null | \
```
jharveyb, line 13: Super-nit; a simpler command:

`ls -1 tapdb/sqlc/migrations/*.up.sql | cut -d'_' -f1 | cut -d'/' -f4 | sort -n`

jharveyb (approved):
LGTM!

Tested the script with a fake migration of prefix `000025`, both only with `up` and `up`/`down`.

Also checked on the `BIGINT PRIMARY KEY` -> `INTEGER PRIMARY KEY` replacement, and that `BIGSERIAL` is the right Postgres type to use here.

scripts/check-migration-numbering.sh
guggero: That seems a bit brittle if someone uses unexpected characters in the file name...

guggero:
Going to verify the postgres itest suceeds locally before merging. I'm also going to build a `litd` image with this and test it against my own node.

guggero:
> Going to verify the postgres itest suceeds locally before merging. I'm also going to build a `litd` image with this and test it against my own node.

Applied this to my node without problems and also ran the full postgres itest locally.

PR #1092, opened 2024-08-19
Enhance Proof Courier Handlers with Lazy Connection Attempts for HashMail and UniverseRPC

ffranr:
This PR introduces updates to the HashMail and UniverseRPC proof courier handlers, specifically making connection attempts optionally lazy. The changes are aimed at enhancing the robustness of the connection handling process by integrating these attempts into the backoff procedure where feasible.

### DIFF
--- a/proof/courier.go
+++ b/proof/courier.go
@@ -185,9 +185,6 @@ func NewCourierDispatch(cfg *CourierCfg) *URLDispatch {
 // address.
 func (u *URLDispatch) NewCourier(ctx context.Context, addr *url.URL,
 	lazyConnect bool) (Courier, error) {
-
-	subscribers := make(map[uint64]*fn.EventReceiver[fn.Event])
-
 	// Create new courier addr based on URL scheme.
 	switch addr.Scheme {
 	case HashmailCourierType:
@@ -197,32 +194,10 @@ func (u *URLDispatch) NewCourier(ctx context.Context, addr *url.URL,
 		)
 
 	case UniverseRpcCourierType:
-		cfg := u.cfg.UniverseRpcCfg
-		backoffHandler := NewBackoffHandler(
-			cfg.BackoffCfg, u.cfg.TransferLog,
+		return NewUniverseRpcCourier(
+			ctx, u.cfg.UniverseRpcCfg, u.cfg.TransferLog,
+			u.cfg.LocalArchive, addr, lazyConnect,
 		)
-
-		// Connect to the universe RPC server.
-		dialOpts, err := serverDialOpts()
-		if err != nil {
-			return nil, err
-		}
-
-		serverAddr := fmt.Sprintf("%s:%s", addr.Hostname(), addr.Port())
-		conn, err := grpc.Dial(serverAddr, dialOpts...)
-		if err != nil {
-			return nil, err
-		}
-
-		client := unirpc.NewUniverseClient(conn)
-
-		return &UniverseRpcCourier{
-			client:        client,
-			backoffHandle: backoffHandler,
-			cfg:           u.cfg,
-			subscribers:   subscribers,
-			rawConn:       conn,
-		}, nil
 
 	default:
 		return nil, fmt.Errorf("unknown courier address protocol "+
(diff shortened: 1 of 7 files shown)


proof/courier.go
```diff
@@ -91,12 +92,72 @@ type CourierCfg struct {
 	LocalArchive Archiver
 }
 
+// CourierConnStatus is an enum that represents the different states a courier
+// connection can be in.
+type CourierConnStatus int
```
guggero, line 97: Is there a specific reason we need to duplicate this type here? Why can't we just use `grpcconn.State` directly?

proof/courier.go
```diff
@@ -737,6 +809,77 @@ type HashMailCourier struct {
...
+		// to determine if the connection is ready.
+		connStatus, err := h.mailbox.ConnectionStatus()
+		if err != nil {
+			return fmt.Errorf("unable to determine connection "+
+				"status: %w", err)
+		}
+
+		// Return early and don't attempt to establish a new connection
+		// if the connection status is idle, ready, or connecting.
+		if connStatus == CourierConnStatusIdle ||
```
guggero, line 830: **If** we add a new type (see comment above), we might as well add a method `IsPending()` or something like that.

proof/courier.go
guggero: Ah, I see it's because we don't have a "disconnected" state in `grpcconn.State`. Might be worth mentioning here.

proof/courier.go
```diff
@@ -1188,6 +1170,77 @@ type UniverseRpcCourier struct {
...
+//
+// This method does nothing if a service connection is already established.
+func (c *UniverseRpcCourier) ensureConnect(ctx context.Context) error {
+	// If we're already connected, we'll return early.
+	if c.rawConn != nil && c.client != nil {
+		// Check the gRPC connection state to determine if the
+		// connection is ready.
+		gRpcConnState := c.rawConn.GetState()
+
+		if gRpcConnState == grpcconn.Idle ||
```
guggero, line 1212: nit: same here re method.

guggero (approved):
Nice refactor! Looks good to me :tada:

proof/courier.go
```diff
@@ -91,12 +92,80 @@ type CourierCfg struct {
...
 }
 
+// CourierConnStatus is an enum that represents the different states a courier
+// connection can be in.
+type CourierConnStatus int
+
+const (
+	// CourierConnStatusUnknown indicates that the connection status is
+	// unknown.
+	CourierConnStatusUnknown CourierConnStatus = iota
```
jharveyb, line 102: non-blocking nit: may be good to have `unknown` after `disconnected`, to follow the enum pattern we have elsewhere / make comparison a bit safer.

Though with exhaustive switches I suppose the order doesn't matter.

proof/courier.go
```diff
@@ -1019,6 +1175,82 @@ type UniverseRpcCourier struct {
...
+		}
+
+		if connStatus.IsPending() {
+			return nil
+		}
+	}
+
+	// At this point, we know that the connection is not ready. We'll now
+	// attempt to establish a new connection to the courier service.
+	dialOpts, err := serverDialOpts()
```
jharveyb, line 1232: You mentioned offline that you wanted the option for the connection setup to be blocking; I think adding the `WithBlock()` `grpc.DialOption` would work for that.

https://pkg.go.dev/google.golang.org/grpc@v1.59.0#WithBlock

jharveyb (commented):
Seems solid! Few comments:

- Would be good for the initial sentence for commit 2 to be different from commit 1.
- What's the recommended setup once this PR is in; should `lazyConnect` always be true?

From itests passing (including those using backoff) this is covered in testing, which is nice.

jharveyb (approved):

proof/courier.go
ffranr: I think it's a good idea to have `unknown` as the first enum value, ensuring that the default for a `CourierConnStatus` variable is `unknown`. Which means that if we don't set the field, it won't default to something opinionated.

proof/courier.go
ffranr: That's what I've been playing around with in conjunction with `DialContext`

ffranr:
> Would be good for the initial sentence for commit 2 to be different from commit 1.

@jharveyb I'm not sure what you mean. Should I make them different for clarity?

> What's the recommended setup once this PR is in; should lazyConnect always be true?

`lazyConnect` is currently `true` at each call site. I have a plan for a use case where it will be `false`.

Issue #675, opened 2023-11-15
[bug]: DB TX retries exceeded during global universe sync


jharveyb:
### Background

Describe your issue here.

When running with tapd v0.3.0 and a global sync of all issuances on, I get numerous log messages about DB TX retries due to serialization errors. In some cases, these operations hit the 10 retry limit and cause sync to fail.

I know that for 0.3.1 onwards global sync is not the default, but IMO this is an underlying issue that I think could appear even without global sync.

### Your environment

```sh
tapcli getinfo                       # version of `tapd`, `lnd`, and network

tapd v0.3.0, lnd v0.17.0, mainnet

uname -mrsv                          # operating system 

debian testing
16-core CPU, NVMe SSD

bitcoind --version || btcd --version # version of `btcd`, `bitcoind`, or other backend

default neutrino backend with a local bitcoind v0.25 peer
```
* any other relevant environment details

### Steps to reproduce

Tell us how to reproduce this issue. Please provide stacktraces and links to code in question.

Perform global issuance sync with the Lightning Labs default universe, with a fresh tapd.

From the log lines preceding these DB issues, I'm pretty sure this is triggered when multiple large (3000+ leaf) asset groups are being synced, and those proofs are being written to disk. The default batch size here is 200 elements, which can be a max size of at least 200 MiB. Since the sync is parallel, we can have multiple goroutines attempting writes in parallel.

The log lines aren't tagged with a goroutine ID or any way to distinguish writers, but based on the attempt counter it seems like 10+ concurrent writers.

Partial logs, generated with `cat tapd.log | rg serializ -C20`:

```sh
2023-11-06 20:19:47.207 [INF] UNIV: Verifying 1 new proofs for insertion into Universe
2023-11-06 20:19:47.207 [DBG] UNIV: UniverseRoot(issuance-766ca5fe4347d5e2218e90387d813c8959e9eaa20e7d777896d1326bad5581f8): Inserting 1 new leaves (1 of 1)
2023-11-06 20:19:47.207 [INF] UNIV: Verifying 1 new proofs for insertion into Universe
2023-11-06 20:19:47.207 [INF] UNIV: Inserting 0 verified group anchor proofs into Universe
2023-11-06 20:19:47.207 [INF] UNIV: Inserting 0 verified group anchor proofs into Universe
[...]

PR #1093, opened 2024-08-19
[universe]: fix proof sync for large asset minting batches

guggero:
I encountered this on `testnet` while trying to reproduce https://github.com/lightninglabs/taproot-assets/issues/675.

Can be reproduced with:
```
--universe.sync-all-assets
--universe.federationserver=testnet.universe.lightning.finance:10029
```

And `tapcli u s --universe_host testnet.universe.lightning.finance:10029`.

This PR fixes the issue by making sure any group anchors are inserted into the DB first.

### DIFF
--- a/universe/syncer.go
+++ b/universe/syncer.go
@@ -272,20 +272,9 @@ func (s *SimpleSyncer) syncRoot(ctx context.Context, remoteRoot Root,
 	// local registrar as they're fetched.
 	var (
 		fetchedLeaves = make(chan *Item, len(keysToFetch))
-		newLeafProofs []*Leaf
+		newLeafProofs = make([]*Leaf, 0, len(keysToFetch))
 		batchSyncEG   errgroup.Group
 	)
-
-	// We use an error group to simply the error handling of a goroutine.
-	// This goroutine will handle reading in batches of new leaves to
-	// insert into the DB. We'll fee the output of the goroutines below
-	// into the input fetchedLeaves channel.
-	batchSyncEG.Go(func() error {
-		newLeafProofs, err = s.batchStreamNewItems(
-			ctx, uniID, fetchedLeaves, len(keysToFetch),
-		)
-		return err
-	})
 
 	// If this is a transfer tree, then we'll use these channels to sort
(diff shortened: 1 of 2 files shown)


universe/syncer.go
```diff
@@ -319,6 +308,28 @@ func (s *SimpleSyncer) syncRoot(ctx context.Context, remoteRoot Root,
...
+					)
+					if err != nil {
+						return fmt.Errorf("unable to "+
+							"register group "+
+							"anchor proof: %w", err)
+					}
+
+					// No need to batch this further, we've
+					// already inserted it.
+					return nil
```
jharveyb, line 330: Suggested fix:

```go
newAnchorProofs  = append(
     newAnchorProofs, leafProof.Leaf,
)
```

universe/syncer.go
```diff
@@ -338,6 +349,17 @@ func (s *SimpleSyncer) syncRoot(ctx context.Context, remoteRoot Root,
 		return err
 	}
 
+	// We use an error group to simply the error handling of a goroutine.
+	// This goroutine will handle reading in batches of new leaves to
+	// insert into the DB. We'll fee the output of the goroutines below
+	// into the input fetchedLeaves channel.
+	batchSyncEG.Go(func() error {
+		newLeafProofs, err = s.batchStreamNewItems(
```
jharveyb, line 357: Itest failure is caused by the leaves being inserted above not being added to `newLeafProofs`, which is used to build the RPC response.

I verified a fix locally by declaring `newAnchorProofs`, populating that as suggested above, and then joining the two lists before sending the `AssetSyncDiff`:

```go
newLeafProofs = append(newAnchorProofs, newLeafProofs...)
```

jharveyb (commented):
Looks good! Just need that one fix + possible logging improvements.

universe/syncer.go
guggero: Thanks, great catch!

jharveyb (approved):
LGTM 👍🏽

Roasbeef (approved):
LGTM 🚜

Issue #1082, opened 2024-08-13
[feature]: Handle Invalid Proof Courier Address During Ongoing Asset Transfer


ffranr:
The proof courier service encoded in a tap address may become permanently offline or its address may be misspent. In such cases, if a user has already initiated an asset transfer to that tap address, the anchoring transaction might confirm on-chain, but the associated proofs cannot be delivered to the intended peer due to the invalid courier address.

**Problem:**
- The proof courier address in the tap address could be invalid or unreachable after an asset transfer has begun.
- The anchoring transaction may confirm on-chain, but the proofs required for the transfer cannot be delivered, potentially causing issues in completing the transfer.

**Proposed Solutions:**
To handle this scenario, we need to implement a mechanism that allows the user to specify an alternative strategy for delivering the proofs. Here are some potential options:

1. **Override Proof Courier Address:**
   - Allow the user to override the proof courier address after the transfer has started.
   - This would require adding a new RPC endpoint that the user can call to specify a new, valid courier address.

2. **Force Delivery to Public Universe Server:**
   - Provide an option for the user to force the tapd to deliver the proof to a public universe server instead.
   - Optionally, the user might be able to configure tapd to do this automatically if the proof courier address is detected as invalid.

3. **Pre-Transfer Proof Courier Ping:**
   - Before the asset transfer process begins, implement a check that pings the proof courier service.
   - If the service does not respond, the transfer process should not proceed, preventing potential issues with undeliverable proofs later on.

Implementing these strategies will ensure that the proofs can still be delivered and the asset transfer completed, even if the initial proof courier service is no longer available.

PR #1097, opened 2024-08-21
rpc: add SetPendingTransferProofCourierAddr endpoint to fix erroneous courier addresses

ffranr:
This PR introduces a new RPC endpoint, SetPendingTransferProofCourierAddr, which allows overriding the proof courier address for a pending asset transfer output. This enhancement directly addresses the issue of correcting erroneous courier addresses during asset transfers as described in [issue #1082](https://github.com/lightninglabs/taproot-assets/issues/1082).


ffranr:
This PR is in draft because it's missing an itest. I'm looking for early feedback.

taprpc/taprootassets.proto
```diff
@@ -611,6 +615,25 @@ message ListTransfersResponse {
     repeated AssetTransfer transfers = 1;
 }
 
+message SetPendingTransferProofCourierAddrRequest {
+    // The new proof courier address to use for the target transfer output proof
+    // delivery.
+    string new_proof_courier_addr = 1;
```
guggero, line 621: Maybe to make it more clear we can only update this field (for now?), we could put that into a new sub-message.

So something like this:

```
message UpdateTransferOutput {
    // The hexadecimal encoded transaction ID (txid) of the anchor transaction
    // for the target transfer.
    string anchor_txid = 1;

    // Script public key of the target transfer output.
    bytes transfer_output_script_pub_key = 2;

    // The position of the output in the transfer output list.
    uint64 transfer_output_position = 3;

    TransferOutputModify modify_params = 4;
}

message TransferOutputModify {
    string proof_courier_addr = 1;
```

rpcserver.go
```diff
@@ -1357,6 +1362,114 @@ func (r *rpcServer) ListTransfers(ctx context.Context,
...
+	}
+
+	// Validate the requested new proof courier address.
+	_, err = proof.ParseCourierAddress(req.NewProofCourierAddr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid proof courier address: %w", err)
+	}
+
+	// Unmarshal the transfer output script key public key.
+	outputScriptKey, err := parseUserKey(req.TransferOutputScriptPubKey)
```
guggero, line 1388: So the pubkey is a safety net to make sure the user really specified the right one?

rpcserver.go
```diff
@@ -1357,6 +1362,114 @@ func (r *rpcServer) ListTransfers(ctx context.Context,
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid script pub key: %w", err)
+	}
+
+	// Query for pending transfers only.
+	pendingParcels, err := r.cfg.AssetStore.QueryParcels(
+		ctx, anchorTxHash, true,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query parcels: %w", err)
```
guggero, line 1398: The user will run into this if they specify a completed transfer. Perhaps we should mention this in the error message, so something like "parcel not found in pending parcels"?

rpcserver.go
```diff
@@ -1357,6 +1362,114 @@ func (r *rpcServer) ListTransfers(ctx context.Context,
...
+
+	// Identify the target parcel and modify the proof courier address.
+	var targetParcel *tapfreighter.OutboundParcel
+	for idxParcel := range pendingParcels {
+		pendingParcel := pendingParcels[idxParcel]
+
+		// The anchoring transaction of target parcel must have already
+		// been confirmed on-chain.
+		if pendingParcel.AnchorTxBlockHash.IsNone() {
+			continue
```
guggero, line 1409: Does this matter? Maybe we want to update before we reached a confirmation?
I assume it has something to do with how we insert a new parcel, see comment below.

rpcserver.go
```diff
@@ -1357,6 +1362,114 @@ func (r *rpcServer) ListTransfers(ctx context.Context,
...
+
+	// Return an error if the target parcel was not found.
+	if targetParcel == nil {
+		return nil, fmt.Errorf("target pending transfer output not " +
+			"found")
+	}
+
+	// Request delivery of the updated parcel.
+	pendingParcel := tapfreighter.NewPendingParcel(targetParcel)
+	_, err = r.cfg.ChainPorter.RequestShipment(pendingParcel)
```
guggero, line 1464: Will this be handled properly and not lead to duplicates? Shouldn't we add the possibility to update an existing parcel instead?

guggero (commented):
Did a first pass, looks like what we need. Have some naming suggestions though.

taprpc/taprootassets.proto
jharveyb: With the current definition I think this is the only reasonable field to change rn, but I agree that using a sub-message is a worthwhile change.

taprpc/taprootassets.proto
```diff
@@ -611,6 +615,25 @@ message ListTransfersResponse {
...
+
+    // The hexadecimal encoded transaction ID (txid) of the anchor transaction
+    // for the target transfer.
+    string anchor_txid = 2;
```
jharveyb, line 625: I think this should be the full outpoint, not just the TXID? Otherwise we aren't uniquely identifying the transfer output.

The transfer output list is per-transfer, but IIUC there is no requirement of exactly 1 transfer per TX (can I combine multiple transfers into one TX?)

So the suggested field would be `anchor_outpoint`, then in the RPC call implementation you could add a filter on `pendingParcel.Outputs` for outputs with a matching `Anchor.OutPoint`.

jtobin:
Closing as stale.

PR #1100, opened 2024-08-22
Enhance Proof Delivery Resilience with ParSliceErrCollect in ChainPorter

ffranr:
This PR updates the proof delivery process in `ChainPorter` by utilizing the newly introduced `ParSliceErrCollect` function. This change ensures that proof delivery continues even if an individual transfer output proof encounters an error, improving the reliability of the delivery process.

This change ensures that as many transfer output proofs as possible are delivered, even in the presence of errors.

### Changes
- **New Function**: Introduced `ParSliceErrCollect`, which collects errors during parallel processing without halting the entire operation.
- **ChainPorter**: Modified to use `ParSliceErrCollect` for parallel proof delivery, allowing the process to continue despite individual delivery failures.

Related to: https://github.com/lightninglabs/taproot-assets/issues/1082

### DIFF
--- a/fn/concurrency.go
+++ b/fn/concurrency.go
@@ -2,7 +2,9 @@ package fn
 
 import (
 	"context"
+	"fmt"
 	"runtime"
+	"sync"
 
 	"golang.org/x/sync/errgroup"
 )
@@ -31,4 +33,49 @@ func ParSlice[V any](ctx context.Context, s []V, f ErrFunc[V]) error {
 	}
 
 	return errGroup.Wait()
+}
+
+// ParSliceErrCollect can be used to execute a function on each element of a
+// slice in parallel. This function is fully blocking and will wait for all
+// goroutines to finish (subject to context cancellation/timeout). Any errors
+// will be collected and returned as a map of slice element index to error.
+// Active goroutines limited with number of CPU.
+func ParSliceErrCollect[V any](ctx context.Context, s []V,
+	f ErrFunc[V]) (map[int]error, error) {
+
+	errGroup, ctx := errgroup.WithContext(ctx)
+	errGroup.SetLimit(runtime.NumCPU())
+
+	var instanceErrorsMutex sync.Mutex
+	instanceErrors := make(map[int]error, len(s))
+
+	for idx := range s {
+		errGroup.Go(func() error {
+			err := f(ctx, s[idx])
+			if err != nil {
+				instanceErrorsMutex.Lock()
+				instanceErrors[idx] = err
+				instanceErrorsMutex.Unlock()
+			}
+
+			// Avoid returning an error here, as that would cancel
+			// the errGroup and terminate all slice element
+			// processing instances. Instead, collect the error and
+			// return it later.
+			return nil
+		})
+	}
+
+	// Now we will wait/block for all goroutines to finish.
+	//
+	// The goroutines that are executing in parallel should not return an
+	// error, but the Wait call may return an error if the context is
+	// canceled or timed out.
+	err := errGroup.Wait()
+	if err != nil {
+		return nil, fmt.Errorf("failed to wait on error group in "+
+			"ParSliceErrorCollect: %w", err)
+	}
+
+	return instanceErrors, nil
 }
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -870,9 +870,41 @@ func (p *ChainPorter) transferReceiverProof(pkg *sendPackage) error {
 
 	// If we have a non-interactive proof, then we'll launch several
 	// goroutines to deliver the proof(s) to the receiver(s).
-	err := fn.ParSlice(ctx, pkg.OutboundPkg.Outputs, deliver)
+	instanceErrors, err := fn.ParSliceErrCollect(
+		ctx, pkg.OutboundPkg.Outputs, deliver,
+	)
 	if err != nil {
 		return fmt.Errorf("error delivering proof(s): %w", err)
+	}
+
+	// If there were any errors during the proof delivery process, we'll
+	// log them all here.
+	for idx := range instanceErrors {
+		output := pkg.OutboundPkg.Outputs[idx]
+		instanceErr := instanceErrors[idx]
+
+		scriptPubKey := output.ScriptKey.PubKey.SerializeCompressed()
+		anchorOutpoint := output.Anchor.OutPoint.String()
+		courierAddr := string(output.ProofCourierAddr)
+
+		log.Errorf("Error delivering transfer output proof "+
+			"(anchor_outpoint=%s, script_pub_key=%v, "+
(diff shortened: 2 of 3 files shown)


dstadulis:
Ensure proofs are delivered given partial proof courier addresses availability

fn/concurrency.go
```diff
@@ -32,3 +34,43 @@ func ParSlice[V any](ctx context.Context, s []V, f ErrFunc[V]) error {
 
 	return errGroup.Wait()
 }
+
+// ParSliceErrCollect can be used to execute a function on each element of a
+// slice in parallel. This function is fully blocking and will wait for all
+// goroutines to finish. Error will be collected and returned as a map of slice
```
guggero, line 40: nit: s/Error will be/Errors will be/

fn/concurrency.go
```diff
@@ -32,3 +34,43 @@ func ParSlice[V any](ctx context.Context, s []V, f ErrFunc[V]) error {
...
+	f ErrFunc[V]) (map[int]error, error) {
+
+	errGroup, ctx := errgroup.WithContext(ctx)
+	errGroup.SetLimit(runtime.NumCPU())
+
+	var instanceErrorsMutex sync.Mutex
+	instanceErrors := make(map[int]error, len(s))
+
+	for idx, v := range s {
+		v := v
```
guggero, line 52: nit: this is no longer required, as we're now using Go 1.22.

fn/concurrency.go
```diff
@@ -32,3 +34,43 @@ func ParSlice[V any](ctx context.Context, s []V, f ErrFunc[V]) error {
...
+			// the errGroup and terminate all slice element
+			// processing instances. Instead, collect the error and
+			// return it later.
+			return nil
+		})
+	}
+
+	err := errGroup.Wait()
+	if err != nil {
+		return nil, fmt.Errorf("failed to wait on error group in "+
```
guggero, line 71: This would only happen when the context is canceled or times out? Might be worth mentioning in the Godoc comment.

fn/concurrency.go
guggero, line 40: "will wait for all goroutines to finish.", except for when the context is canceled from the outside, right?

fn/concurrency.go
```diff
@@ -32,3 +34,43 @@ func ParSlice[V any](ctx context.Context, s []V, f ErrFunc[V]) error {
 
 	return errGroup.Wait()
 }
+
+// ParSliceErrCollect can be used to execute a function on each element of a
```
guggero, line 38: Very nice function, should be handy in other places I can imagine.

tapfreighter/chain_porter.go
```diff
@@ -870,11 +870,45 @@ func (p *ChainPorter) transferReceiverProof(pkg *sendPackage) error {
...
+			log.Errorf("Error delivering transfer output proof "+
+				"(anchor_outpoint=%s, script_pub_key=%v, "+
+				"position=%d, proof_courier_addr=%s, "+
+				"proof_delivery_status=%v): %v",
+				anchorOutpoint, scriptPubKey, output.Position,
+				courierAddr, output.ProofDeliveryComplete,
+				instanceErr)
+		}
+
+		if firstErr != nil {
```
guggero, line 907: Instead of tracking this separately, I think we could add a function `fn.PeekMap()` that returns the first element of a map?

guggero (commented):
Very nice improvement!

Just nits and suggestions.

fn/concurrency.go
```diff
@@ -32,3 +34,43 @@ func ParSlice[V any](ctx context.Context, s []V, f ErrFunc[V]) error {
...
+// goroutines to finish. Error will be collected and returned as a map of slice
+// element index to error. Active goroutines limited with number of CPU.
+func ParSliceErrCollect[V any](ctx context.Context, s []V,
+	f ErrFunc[V]) (map[int]error, error) {
+
+	errGroup, ctx := errgroup.WithContext(ctx)
+	errGroup.SetLimit(runtime.NumCPU())
+
+	var instanceErrorsMutex sync.Mutex
+	instanceErrors := make(map[int]error, len(s))
```
jharveyb, line 49: nit: can use `lnutils.SyncMap` here. Though fetching only the keys would be more ergonomic with what you have now IMO.

tapfreighter/chain_porter.go
```diff
@@ -870,11 +870,45 @@ func (p *ChainPorter) transferReceiverProof(pkg *sendPackage) error {
...
+	)
 	if err != nil {
 		return fmt.Errorf("error delivering proof(s): %w", err)
 	}
 
+	// If there were any errors during the proof delivery process, we'll
+	// report them here and return the first error we encountered.
+	if len(instanceErrors) > 0 {
+		var firstErr error
+		for idx := range instanceErrors {
```
jharveyb, line 884: `range` over a map will be unordered:

https://go.dev/blog/maps

But you could build a slice of ordered keys with something like:

`slices.Sort(maps.Keys(instanceErrors))`

IIUC you want to log all errors but only return the first one.

tapfreighter/chain_porter.go
jharveyb: Also, shouldn't `firstErr` always be non-nil at this point, as the map of errs is non-empty and there should be no entry will a nil value?

jharveyb (commented):
Looks good, a useful func! I think we have some spots in the universe code, specifically federations, where this could be used as well.

Would be good to add a test for this alongside the test for `ParSlice`.

fn/concurrency.go
ffranr: I've gone with `s[idx]` in the latest commits. Are you saying that `v := v` isn't required because of Go 1.22? Why is that, can you say more please?

tapfreighter/chain_porter.go
ffranr: > Also, shouldn't firstErr always be non-nil at this point, as the map of errs is non-empty and there should be no entry will a nil value?

I've removed `if len(instanceErrors) > 0 {` in the latest commit, you were right @jharveyb that some of the code was redundant in the previous version.

fn/concurrency.go
ffranr: I figured I could return `map[int]error` directly rather than use `lnutils.SyncMap`. And that the sync handling mutex would be restricted to the scope of this small function anyway.

That's why I went with a plain map over anything else. Any thoughts on my reasoning here?

fn/concurrency.go
guggero: https://go.dev/blog/loopvar-preview

tapfreighter/chain_porter.go
ffranr: I've added `fn.PeekMap` to select on of the errors. It doesn't select deterministically but I think that's OK in this case.

jharveyb (approved):
LGTM 👍🏽

Issue #1099, opened 2024-08-22
[custom channels]: proof invalid after sweeping delayed to_local from commitment TX


guggero:
Reported by @vanditshah99 on Slack.

Error message:
```
2024-08-22 05:23:15.914 [INF] TCHN: Funding new vPacket channel, asset_id=d75dae162be7cd0a2df76831ee1cce4eaeb74a32fabb0d46be7a0ad70fd6ad27, amt=2500000000
2024-08-22 05:23:16.005 [INF] FRTR: Identified 4 eligible asset inputs for send of 2500000000 to {{{true [215 93 174 22 43 231 205 10 45 247 104 49 238 28 206 78 174 183 74 50 250 187 13 70 190 122 10 215 15 214 173 39]} {true {{[34110241 1949323 48120036 40959240 24868899 21497998 20063886 41859510 24713038 2748180]} {[7917944 17651436 37337177 9952462 25770521 11959941 57927709 49520190 57251631 4178363]}}}} 2500000000 true}
2024-08-22 05:23:16.036 [INF] FRTR: Selected 1 asset inputs for send of 2500000000 to d75dae162be7cd0a2df76831ee1cce4eaeb74a32fabb0d46be7a0ad70fd6ad27
2024-08-22 05:23:16.071 [DBG] TAPD: Deriving new key for fam_family=212
2024-08-22 05:23:16.071 [DBG] RPCS: [/walletrpc.WalletKit/DeriveNextKey] requested
2024-08-22 05:23:16.089 [INF] TCHN: Generating input ownership proofs for 1 inputs
2024-08-22 05:23:16.089 [INF] FRTR: Generating ownership proof for asset d75dae162be7cd0a2df76831ee1cce4eaeb74a32fabb0d46be7a0ad70fd6ad27
2024-08-22 05:23:16.090 [DBG] RPCS: [/signrpc.Signer/SignOutputRaw] requested
2024-08-22 05:23:16.090 [DBG] SGNR: Generating sigs for 1 inputs:
2024-08-22 05:23:16.092 [DBG] TCHN: unlocking asset inputs: ([]wire.OutPoint) (len=1 cap=1) {
 (wire.OutPoint) f75b88f0dad035f6b8ad5c6b43081ee245eb24b76e24237fc322f083b4f9d1ee:0
}

[...]

PR #1103, opened 2024-08-23
commitment: splits shouldn't inherit locktime from inputs

Roasbeef:
In this commit, we fix an existing bug related to lock times and split commitments. Before this commit, if an input had a relative lock time, then when we went to make the new split assets, we would _copy_ that value into the split. This isn't correct as the input is valid/confirmed, so we don't need to copy over the lock time information.

The prior behavior would cause certain classes of spends to fail, as we would be validating a new root asset that has no lock time, but the root asset split inserted into the split commitment would be carrying the old lock time. When verifying the split, we would set the lock times of the split to that of the new asset:
https://github.com/lightninglabs/taproot-assets/blob/e893dee87e9d8f0de53b8ee9e2527add80df6491/vm/vm.go#L305-L307.

As we copied over the lock time from the input, we would now effectively invalid the split commitment.

Fixes https://github.com/lightninglabs/taproot-assets/issues/1099

### DIFF
--- a/tapfreighter/wallet.go
+++ b/tapfreighter/wallet.go
@@ -315,6 +315,11 @@ func createPassivePacket(params *address.ChainParams, passiveAsset *asset.Asset,
 	// asset.
 	outputAsset.SplitCommitmentRoot = nil
 
+	// If this was previously a time locked asset, it's now a simple asset.
+	// So we need to clear out the time locks as well.
+	outputAsset.LockTime = 0
+	outputAsset.RelativeLockTime = 0
+
 	// Clear the output asset witness data. We'll be creating a new witness.
 	outputAsset.PrevWitnesses = []asset.Witness{{
 		PrevID: &vInput.PrevID,
--- a/tapfreighter/wallet.go
+++ b/tapfreighter/wallet.go
@@ -309,16 +309,7 @@ func createPassivePacket(params *address.ChainParams, passiveAsset *asset.Asset,
 	}
 
 	// Specify virtual output.
-	outputAsset := passiveAsset.Copy()
-
-	// Clear the split commitment root, as we'll be transferring the whole
-	// asset.
-	outputAsset.SplitCommitmentRoot = nil
-
-	// If this was previously a time locked asset, it's now a simple asset.
-	// So we need to clear out the time locks as well.
-	outputAsset.LockTime = 0
-	outputAsset.RelativeLockTime = 0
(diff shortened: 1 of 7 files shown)


guggero (approved):
Nice find! I pushed an updated commit that also updates some other call sites.
Perhaps we should extract this "copy for descendant asset" functionality into a generic function?

guggero:
Updated https://github.com/lightninglabs/lightning-terminal/pull/825 to confirm this fully fixes the issue.

tapfreighter/wallet.go
```diff
@@ -315,6 +315,11 @@ func createPassivePacket(params *address.ChainParams, passiveAsset *asset.Asset,
 	// asset.
 	outputAsset.SplitCommitmentRoot = nil
 
+	// If this was previously a time locked asset, it's now a simple asset.
+	// So we need to clear out the time locks as well.
+	outputAsset.LockTime = 0
```
jharveyb, line 320: IIUC, wrt. tapchannels, this would cover the case of the input for the funding TX anchoring 2+ assets, where 1 asset is used for channel funding, and 1+ assets would be change (passive) and sent back to the funder in a separate output?

I think we don't cover that on the Terminal itests rn, will work on that.

jharveyb (approved):
LGTM! Wondering if we shouldn't bundle the `SplitCommitment` and locktime resetting into another method similar to `Copy()`, to make this more explicit / easier in the future.

Roasbeef:
> Perhaps we should extract this "copy for descendant asset" functionality into a generic function?

SGTM, was thinking that too, can tack it on.

tapfreighter/wallet.go
Roasbeef: I think Oli tacked this on. 

At a glance, makes sense as the passive assets are always 1-in-1-out transactions so it doesn't go through the normal split creation case. 

Based on his comment, we can make a like `CopyFreshAssetTemplate` function that knows to blank out these fields, and we use it in both contexts. 

 > this would cover the case of the input for the funding TX anchoring 2+ assets, where 1 asset is used for channel funding, and 1+ assets would be change (passive) and sent back to the funder in a separate output?

Good question...today we still only support a single asset in the anchor. I think this is more precautionary for a case that can arise with custom vPSBT flows? 

cc @guggero

Roasbeef:
Implemented here: https://github.com/lightninglabs/taproot-assets/pull/1103/commits/881ecafbeae15a0058f6187a03dbd3632da230be

PR #1107, opened 2024-08-26
Add AcceptPriceDeviationPpm configuration

lukegao209:
The AcceptPriceDeviationPpm is set at 5% and cannot be configured. This presents a certain price slippage risk for RFQ Swaps. Request to add this configuration option. The default value should be 50,000 (5%)

### DIFF
--- a/sample-tapd.conf
+++ b/sample-tapd.conf
@@ -317,7 +317,8 @@
 ; use_mock_price_oracle_service_promise_to_not_use_on_mainnet
 ; experimental.rfq.priceoracleaddress=
 
-; The default price deviation inparts per million that is accepted by the RFQ negotiator.
+; The default price deviation inparts per million that is accepted by 
+; the RFQ negotiator.
 ; Example: 50,000 ppm => price deviation is set to 5% .
 ; experimental.rfq.acceptpricedeviationppm=50000
 
--- a/sample-tapd.conf
+++ b/sample-tapd.conf
@@ -310,12 +310,16 @@
 ; (latency, etc)
 ; prometheus.perfhistograms=false
 
-[experimental.rfq]
+[experimental]
 
 ; Price oracle gRPC server address (rfqrpc://<hostname>:<port>)
 ; To use the integrated mock, use the following value:
 ; use_mock_price_oracle_service_promise_to_not_use_on_mainnet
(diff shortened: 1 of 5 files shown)


sample-tapd.conf
```diff
@@ -310,13 +310,17 @@
 ; (latency, etc)
 ; prometheus.perfhistograms=false
 
-[experimental.rfq]
+[experimental]
```
guggero, line 313: I'm not sure this change is correct. I think the section headers always need to reflect the full prefix to be valid. The config file parsing library we use is quite particular about these things...

sample-tapd.conf
```diff
@@ -310,13 +310,17 @@
...
 
 ; Price oracle gRPC server address (rfqrpc://<hostname>:<port>)
 ; To use the integrated mock, use the following value:
 ; use_mock_price_oracle_service_promise_to_not_use_on_mainnet
 ; experimental.rfq.priceoracleaddress=
 
+; The default price deviation inparts per million that is accepted by the RFQ negotiator.
```
guggero, line 320: nit: break before 80 characters.

tapcfg/server.go
```diff
@@ -373,6 +373,7 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 			ChannelLister:   walletAnchor,
 			AliasManager:    lndRouterClient,
 			// nolint: lll
+			AcceptPriceDeviationPpm:   rfqCfg.AcceptPriceDeviationPpm,
```
guggero, line 376: If a `//nolint` comment is above a line, it only counts for that line. But both of these are too long. So either needs a comment each, or put the comment before a block, so it counts for the whole block (so for example before `rfq.ManagerCfg{` on line 369).

guggero (commented):
Thanks for the change! Makes sense, just a couple of nits.

sample-tapd.conf
lukegao209: I’ve tested it. The experimental RFQ feature doesn’t work with [experimental.rfq] ; it shows a ‘no group found’ warning

guggero (approved):
Linter still not happy. And would be nice if you could squash all the commits into a single one with a message that makes sense outside of the context of PR reviews.
Other than that LGTM :tada:

PR #1108, opened 2024-08-26
Accept price deviation ppm configuration

lukegao209:
The AcceptPriceDeviationPpm is set at 5% and cannot be configured. This presents a certain price slippage risk for RFQ Swaps. Request to add this configuration option. The default value should be 50,000 (5%)


guggero:
Comments from previous PR still apply: Linter needs fixing and commits should be squashed.

PR #1115, opened 2024-09-03
[wallet]: only select BIP-86 script keys for channel funding, add flag to FundVirtualPsbt

guggero:
Fixes a bug that caused channel funding transactions to be mistakenly used for coin selection, even though they are known to be script path only spend assets.


tapfreighter/chain_porter.go
```diff
@@ -994,7 +994,7 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
 				"address parcel")
 		}
 		fundSendRes, err := p.cfg.AssetWallet.FundAddressSend(
-			ctx, addrParcel.destAddrs...,
+			ctx, true, addrParcel.destAddrs...,
```
Roasbeef, line 997: I think we can make this diff/change and future call sites more readable if we introduce an enum early here instead of a `bool`. 

It could be something like (random names, can def be improved): 
```go
type CoinSelectType uint8

const (
    DefaultCoinSelect CoinSelectType
    Bip86Only
    ScriptTreesAllowed 
)
```

Then we can make `DefaultCoinSelect = Bip86Only` when parsing, which is typically the behavior you want unless doing something fancy.

tapfreighter/wallet.go
```diff
@@ -67,14 +67,15 @@ type Wallet interface {
...
+	FundAddressSend(ctx context.Context, bip86ScriptKeysOnly bool,
 		receiverAddrs ...*address.Tap) (*FundedVPacket, error)
 
 	// FundPacket funds a virtual transaction, selecting assets to spend
 	// in order to pay the given recipient. The selected input is then added
 	// to the given virtual transaction.
 	FundPacket(ctx context.Context, fundDesc *tapsend.FundingDescriptor,
-		vPkt *tappsbt.VPacket) (*FundedVPacket, error)
+		vPkt *tappsbt.VPacket,
+		bip86ScriptKeysOnly bool) (*FundedVPacket, error)
```
Roasbeef, line 78: With the comment above, the new enum can possibly go in the `tapsend.FundingDescriptor`.

itest/mint_fund_seal_test.go
```diff
@@ -385,6 +385,26 @@ func testMintFundSealAssets(t *harnessTest) {
 		Amt:     assetTweakedScriptKey.Amount / 2,
 	})
 	require.NoError(t.t, err)
+
+	// We only have script encumbered assets now, so selecting BIP-086 only
+	// assets should result in an error.
```
Roasbeef, line 390: 👍

Roasbeef (changes requested):
Nice fix!

tapfreighter/chain_porter.go
guggero: Makes sense! Changed the type. But had to keep the default value at `ScriptTreesAllowed` to not break existing users of the `FundVirtualPsbts` RPC call (which is primarily used for doing script sends, at least in our itests).

jharveyb (approved):
LGTM, AFAICT the default for `FundVirtualPsbt` will be the same but `FundAddressSend` will be BIP86 only.

Would be nice to have less negations in the path from the enum to the actual DB filter, but minor nit / non-blocking.

guggero:
> have less negations in the path from the enum to the actual DB filter

I'm not sure what you mean. Are you referencing the naming/wording from "only" and "allowed" that is later mapped to a single boolean?

Roasbeef (approved):
LGTM 📈

PR #1116, opened 2024-09-05
Small fixes after recent PRs

guggero:
Fixes formatting of build tag excluded files the linter didn't pick up on.

And fixes the new itest to actually run against the correct `litd` branch (I missed that during review). At the same time we slightly optimize the litd itest build.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -20,6 +20,8 @@ env:
   GOPATH: /home/runner/work/go
 
   GO_VERSION: '1.22.3'
+  
+  LITD_ITEST_BRANCH: '0-19-staging'
 
 jobs:
   #######################
@@ -306,7 +308,11 @@ jobs:
         uses: ./.github/actions/setup-go
 
       - name: Clone Lit repository
-        run: git clone https://github.com/lightninglabs/lightning-terminal.git
+        uses: actions/checkout@v3
+        with:
+          repository: lightninglabs/lightning-terminal
+          ref: ${{ env.LITD_ITEST_BRANCH }}
+          path: lightning-terminal
 
       - name: Update go.mod to use the local Tap repository
         working-directory: ./lightning-terminal
@@ -316,7 +322,7 @@ jobs:
 
       - name: Install yarn
         run: npm install -g yarn
-          
+
       - name: setup nodejs
         uses: ./lightning-terminal/.github/actions/setup-node
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yaml
```diff
@@ -306,7 +308,11 @@ jobs:
         uses: ./.github/actions/setup-go
 
       - name: Clone Lit repository
-        run: git clone https://github.com/lightninglabs/lightning-terminal.git
+        uses: actions/checkout@v3
+        with:
+          repository: lightninglabs/lightning-terminal
+          ref: ${{ env.LITD_ITEST_BRANCH }}
```
GeorgeTsagk, line 314: not sure if related, but there's a tiny misbehavior according to an itest
https://github.com/lightninglabs/taproot-assets/actions/runs/10715230733/job/29710290030?pr=1116#step:4:52159

we get a "status completed" instead of "proofs transferred"

GeorgeTsagk (approved):
Lgtm! thanks for the fixes

for some reason my `make lint` run didn't catch these yesterday

.github/workflows/main.yaml
guggero: Hmm, that seems to be a GitHub only flake... I ran this test a bunch of time locally and it never failed for me...
Going to re-run CI and just hope it goes away.

gijswijs: Man, this has been driving me bananas! I had something similar way back and never understood why that happens on GitHub.

The weird thing is that in my case when the test ran on Github it expected a different state than what the test was actually supposed to be expecting, and I think that's the case hear as well, right?

The test expects `SendStateWaitTxConf`
https://github.com/lightninglabs/taproot-assets/blob/6f4d65fa7ad1c0b7e40b543467c4dea0ebbde9aa/itest/psbt_test.go#L1148-L1152

But according to the run on GitHub the test expects `SendStateTransferProofs` which makes no sense whatsoever:
https://github.com/lightninglabs/taproot-assets/actions/runs/10715230733/job/29710290030?pr=1116#step:4:52158

Maybe an enumeration is off by one?

gijswijs (approved):

.github/workflows/main.yaml
guggero: The enum values passed into `AssertSendEvents()` is a "from" and "to". So it's expecting all events from `SendStateWaitTxConf` in order up until `SendStateComplete`. And `SendStateTransferProofs` just happens to be an in-between one.
So I think it's a pure timing issue, perhaps because GitHub is slower than the local machine.

gijswijs: Oh, I misunderstood that. Thanks!

PR #1117, opened 2024-09-06
[rfq]: Refactor according to proposed changes for better precision

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/1059.

Implements the suggested changes from the document in https://github.com/lightninglabs/taproot-assets/pull/1059.

This is a HUGE refactor and will be a breaking change both on the RFQ p2p message level as well as on the price oracle RPC level.
Because it's not breaking on any persistence level, it's simply enough to upgrade both peers of a channel.
Since we know it's going to be a breaking change anyway and we'll need to just rip off the bandage, we might as well put in as many refactor and cleanup changes as possible.


rfqmsg/reject.go
```diff
@@ -125,54 +110,42 @@ const (
...
 
 	// ID represents the unique identifier of the quote request message that
 	// this response is associated with.
-	ID ID
+	ID tlv.RecordT[tlv.TlvType2, ID]
 
 	// Err is the error code and message that provides the reason for the
 	// rejection.
-	Err RejectErr
+	Err tlv.RecordT[tlv.TlvType4, RejectErr]
```
ffranr, line 125: In the previous TLV setup the err field was optional. I think we want the same here. The peer shouldn't have to provide an error. I think we should use `tlv.TlvType3` here.

rfqmsg/reject.go
```diff
@@ -230,9 +205,9 @@ func NewQuoteRejectFromWireMsg(wireMsg WireMessage) (*Reject, error) {
 	}
 
 	// Ensure that the message version is supported.
-	if msgData.Version > latestRejectVersion {
+	if msgData.Version.Val > latestRejectVersion {
```
ffranr: with this update I think that we can handle precisely the latest version (bumped in this PR) and no other. Including no earlier version. So Should this be `!=` rather than `>`?

rfqmsg/messages.go
```diff
@@ -76,6 +79,10 @@ var (
 	// ErrUnknownMessageType is an error that is returned when an unknown
 	// message type is encountered.
 	ErrUnknownMessageType = errors.New("unknown message type")
+
+	// MilliSatPerBtc is the number of milli-satoshis in a bitcoin.
+	// This is 100 billion, which is 10^11.
+	MilliSatPerBtc = NewUint64FixedPoint(1, 11)
```
ffranr, line 85: Perhaps consider the name `MsatPerBtc` instead? I think it's shorter but still clear.

rfqmsg/messages.go
```diff
@@ -182,3 +189,55 @@ type QuoteResponse interface {
...
+// NewUint64FixedPoint creates a new fixed point record with a uint64
+// value and a scale.
+func NewUint64FixedPoint(value uint64, scale int) Uint64FixedPoint {
+	return Uint64FixedPoint{
+		Value: rfqmath.NewGoInt(value),
+		Scale: scale,
+	}
+}
+
+// Record returns a TLV record that can be used to encode/decode an ID to/from a
```
ffranr, line 212: `an ID` should be modified to `a uint64 fixed point`

rfqmsg/messages.go
```diff
@@ -182,3 +189,55 @@ type QuoteResponse interface {
...
+	// asset is BTC, this is set to MilliSatPerBtc.
+	InAssetPrice Uint64FixedPoint
+
+	// InAssetPrice is the price of the output asset, expressed in asset
+	// units per BTC, represented as a fixed point number. If the output
+	// asset is BTC, this is set to MilliSatPerBtc.
+	OutAssetPrice Uint64FixedPoint
+
+	// Expiry is the price's expiry.
+	Expiry time.Time
```
ffranr, line 242: We should clarify in the document what we mean by "expiry." Specifically, it refers to the Unix timestamp (in seconds) after which the quote will expire. This represents an absolute point in time, not a time offset.

rfqmsg/messages.go
```diff
@@ -182,3 +189,55 @@ type QuoteResponse interface {
...
+	// Note that we set the type here as zero, as when used with a
+	// tlv.RecordT, the type param will be used as the type.
+	return tlv.MakeStaticRecord(
+		0, r, recordSize, Uint64FixedPointEncoder,
+		Uint64FixedPointDecoder,
+	)
+}
+
+// PriceQuote is a struct that holds the price quote by an oracle for a swap
+// between two assets.
```
ffranr, line 229: I think we should remove "by an oracle". Where a price quote comes from or what generates it is not relevant to what it is. We might store/retrieve via an archive. Maybe something like this:

```
// PriceQuote represents a price quote for an asset exchange operation.
//
// It contains the prices of both the input and output assets, expressed in
// asset units per BTC, using a fixed point representation. These prices can be
// considered as the exchange rates for swapping between the input and output
// assets. The quote is valid for a limited time and will expire.
```

rfqmsg/messages.go
```diff
@@ -182,3 +189,55 @@ type QuoteResponse interface {
...
 	String() string
 }
+
+// Uint64FixedPoint is a fixed point record that can be used to encode/decode a
+// fixed point number with a uint64 value to/from a TLV stream.
+type Uint64FixedPoint rfqmath.FixedPoint[rfqmath.GoInt[uint64]]
+
+// NewUint64FixedPoint creates a new fixed point record with a uint64
+// value and a scale.
+func NewUint64FixedPoint(value uint64, scale int) Uint64FixedPoint {
```
ffranr, line 205: `scale` here should be a `uint` perhaps? negative values should not be allowed from what I understand

Maybe it should be a `uint8` because that's what it encodes into. Using that type will avoid overflows during encoding.

PR #1118, opened 2024-09-09
tapchannel: improve aux signer signal handling

### LINKED ISSUE #1114: tapchannel - [bug]: potential deadlock when signing second level HTLCs
### Background

A user has noticed that occasionally the daemon may freeze up when signing second level HTLCs. 

We we're able to obtain a trace to confirm that the daemon was indeed deadlocked, with `lnd` waiting for a channel response from `tapd`, while `tapd` is attempting to send a response to `lnd`:
```
goroutine 2042 [chan receive, 44 minutes]:
github.com/lightningnetwork/lnd/lnwallet.(*LightningChannel).SignNextCommitment(0xc006d40360)
	github.com/lightningnetwork/lnd@v0.18.0-beta.rc4.0.20240730143253-1b353b0bfd58/lnwallet/channel.go:4706 +0x124c
github.com/lightningnetwork/lnd/htlcswitch.(*channelLink).updateCommitTx(0xc00280ec08)
	github.com/lightningnetwork/lnd@v0.18.0-beta.rc4.0.20240730143253-1b353b0bfd58/htlcswitch/link.go:2534 +0x6a
github.com/lightningnetwork/lnd/htlcswitch.(*channelLink).updateCommitTxOrFail(0xc00280ec08)
	github.com/lightningnetwork/lnd@v0.18.0-beta.rc4.0.20240730143253-1b353b0bfd58/htlcswitch/link.go:2485 +0x1f
github.com/lightningnetwork/lnd/htlcswitch.(*channelLink).handleUpstreamMsg(0xc00280ec08, {0x389c7b0, 0xc00faed740})
	github.com/lightningnetwork/lnd@v0.18.0-beta.rc4.0.20240730143253-1b353b0bfd58/htlcswitch/link.go:2367 +0xa96
github.com/lightningnetwork/lnd/htlcswitch.(*channelLink).htlcManager(0xc00280ec08)
	github.com/lightningnetwork/lnd@v0.18.0-beta.rc4.0.20240730143253-1b353b0bfd58/htlcswitch/link.go:1415 +0x1596
created by github.com/lightningnetwork/lnd/htlcswitch.(*channelLink).Start in goroutine 1680
	github.com/lightningnetwork/lnd@v0.18.0-beta.rc4.0.20240730143253-1b353b0bfd58/htlcswitch/link.go:541 +0x47d
```
```
goroutine 1324695 [chan send, 44 minutes]:
github.com/lightninglabs/taproot-assets/tapchannel.(*AuxLeafSigner).processAuxSigBatch(0xc005532930, 0xc0049c1008, 0xc00583e800, {0xc0178b0008, 0x2, 0x2?})
	github.com/lightninglabs/taproot-assets@v0.4.2-0.20240824000229-881ecafbeae1/tapchannel/aux_leaf_signer.go:269 +0x1eb
created by github.com/lightninglabs/taproot-assets/tapchannel.(*AuxLeafSigner).SubmitSecondLevelSigBatch in goroutine 2042
	github.com/lightninglabs/taproot-assets@v0.4.2-0.20240824000229-881ecafbeae1/tapchannel/aux_leaf_signer.go:104 +0xcf
```

Here's a diagram from Clade based on the trace above: 
```mermaid
graph TD
    subgraph "goroutine 2042 (lnd)"
        E[htlcManager]
        D[handleUpstreamMsg]
        C[updateCommitTxOrFail]
        B[updateCommitTx]
        A[SignNextCommitment]
        
        E --> D
        D --> C
        C --> B
        B -->|"chan receive (44 min)"| A
    end

    subgraph "goroutine 1324695 (taproot-assets)"
        G[SubmitSecondLevelSigBatch]
        F[processAuxSigBatch]
        
        G -->|"chan send (44 min)"| F
    end

    G -->|Creates| F
    G -->|Waiting for| A
    E -->|Waiting for| G

    classDef waiting fill:#FF9999,stroke:#333,stroke-width:2px,color:#000;
    classDef normal fill:#D9EAD3,stroke:#333,stroke-width:1px,color:#000;
    class A,F waiting;
    class B,C,D,E,G normal;
```

From the above, we can see that we have circular waiting dependancy. 

As is, the channel created for the transfer of sigs is always buffered:  https://github.com/lightningnetwork/lnd/blob/e78b0bb8532dd181cfa112790113335d65937e37/lnwallet/aux_signer.go#L58-L76

We're then blocking here in `tapd`: https://github.com/lightninglabs/taproot-assets/blob/0551a3f6c147c085b1e520ae192806f15873e24f/tapchannel/aux_leaf_signer.go#L265-L273

One thing to note is that in `tapd`, this is the send when there's **no aux blob for a channel**. 

### Expected behavior

Buffered channel send never blocks. 

### Actual behavior

Buffered channel send blocks. Potentially there's some underlying mutation here.

jharveyb:
Fixes #1114 .

This PR updates the aux signer to handle cancel and quit signals in a similar fashion to the 'primary' sig job handler in `lnwallet/sigpool`. This also improves safety as the cancel channel can now be closed by both `tapd` and `lnd`.

### DIFF
--- a/tapchannel/auf_leaf_signer_test.go
+++ b/tapchannel/auf_leaf_signer_test.go
@@ -22,7 +22,6 @@ import (
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwire"
-	"github.com/lightningnetwork/lnd/tlv"
 	"github.com/stretchr/testify/require"
 )
 
@@ -40,10 +39,58 @@ var (
 	)
 
 	testTimeout = time.Second
+
+	chanState = &channeldb.OpenChannel{
+		ChanType: channeldb.AnchorOutputsBit |
+			channeldb.ScidAliasChanBit | channeldb.SingleFunderBit |
+			channeldb.SimpleTaprootFeatureBit |
+			channeldb.TapscriptRootBit,
+		IsInitiator: true,
+	}
+
+	// sig job batch size when making more that one sig job.
+	numSigJobs = int32(10)
+
+	// Threshold for trying to cancel or quit the aux leaf signer (allow
+	// the signer to complete a third of the batch).
+	sigJobCancelThreshold = numSigJobs / 3
 )
 
-// TestAuxLeafSigner tests the AuxLeafSigner implementation.
-func TestAuxLeafSigner(t *testing.T) {
+// RandAuxSigJob generates a basic aux signer job with random key material.
+func RandAuxSigJob(t *testing.T, cancelChan chan struct{},
+	commitBlob lfn.Option[[]byte], outputIdx int32) lnwallet.AuxSigJob {
+
+	keyDesc, _ := test.RandKeyDesc(t)
+	keyRing := test.RandCommitmentKeyRing(t)
+
+	return lnwallet.AuxSigJob{
+		SignDesc: input.SignDescriptor{
+			KeyDesc: keyDesc,
+		},
+		BaseAuxJob: lnwallet.BaseAuxJob{
+			OutputIndex: outputIdx,
+			KeyRing:     keyRing,
+			HTLC: lnwallet.PaymentDescriptor{
+				HtlcIndex: 0,
+				Amount: lnwire.NewMSatFromSatoshis(
+					354,
+				),
+				EntryType: lnwallet.Add,
+			},
+			Incoming:   false,
+			CommitBlob: commitBlob,
+			HtlcLeaf:   input.AuxTapLeaf{},
+		},
+		Resp:   make(chan lnwallet.AuxSigJobResp, 1),
+		Cancel: cancelChan,
+	}
+}
+
+// setupAuxLeafSigner sets up an AuxLeafSigner instance and a batch of sig jobs
+// to use in unit tests.
+func setupAuxLeafSigner(t *testing.T, numJobs int32) (*AuxLeafSigner,
+	chan struct{}, *wire.MsgTx, []lnwallet.AuxSigJob) {
+
 	cfg := &LeafSignerConfig{
 		ChainParams: testChainParams,
 		Signer:      &mockVirtualSigner{},
@@ -52,30 +99,8 @@ func TestAuxLeafSigner(t *testing.T) {
 	signer := NewAuxLeafSigner(cfg)
 	require.NoError(t, signer.Start())
 
-	defer func() {
-		require.NoError(t, signer.Stop())
-	}()
-
-	chanState := &channeldb.OpenChannel{
-		ChanType: channeldb.AnchorOutputsBit |
-			channeldb.ScidAliasChanBit | channeldb.SingleFunderBit |
-			channeldb.SimpleTaprootFeatureBit |
-			channeldb.TapscriptRootBit,
-		IsInitiator: true,
-	}
 	randInputProof := randProof(t)
 	commitTx := &randInputProof.AnchorTx
-	keyRing := lnwallet.CommitmentKeyRing{
-		CommitPoint:         test.RandPubKey(t),
-		LocalCommitKeyTweak: test.RandBytes(32),
-		LocalHtlcKeyTweak:   test.RandBytes(32),
-		LocalHtlcKey:        test.RandPubKey(t),
-		RemoteHtlcKey:       test.RandPubKey(t),
-		ToLocalKey:          test.RandPubKey(t),
-		ToRemoteKey:         test.RandPubKey(t),
-		RevocationKey:       test.RandPubKey(t),
-	}
-
 	outgoingHtlcs := make(map[input.HtlcIndex][]*cmsg.AssetOutput)
 	outgoingHtlcs[0] = []*cmsg.AssetOutput{
 		cmsg.NewAssetOutput(
@@ -87,32 +112,27 @@ func TestAuxLeafSigner(t *testing.T) {
 	com := cmsg.NewCommitment(
 		nil, nil, outgoingHtlcs, nil, lnwallet.CommitAuxLeaves{},
 	)
+	cancelChan := make(chan struct{})
 
-	randKeyDesc, _ := test.RandKeyDesc(t)
-
-	jobs := []lnwallet.AuxSigJob{
-		{
-			SignDesc: input.SignDescriptor{
-				KeyDesc: randKeyDesc,
-			},
-			BaseAuxJob: lnwallet.BaseAuxJob{
-				OutputIndex: 0,
-				KeyRing:     keyRing,
-				HTLC: lnwallet.PaymentDescriptor{
-					HtlcIndex: 0,
-					Amount: lnwire.NewMSatFromSatoshis(
-						354,
-					),
-					EntryType: lnwallet.Add,
-				},
-				Incoming:   false,
-				CommitBlob: lfn.Some[tlv.Blob](com.Bytes()),
-				HtlcLeaf:   input.AuxTapLeaf{},
-			},
-			Resp:   make(chan lnwallet.AuxSigJobResp),
-			Cancel: make(chan struct{}),
-		},
+	// Constructing multiple jobs will allow us to assert that later jobs
+	// are cancelled successfully.
+	jobs := make([]lnwallet.AuxSigJob, 0, numJobs)
+	for idx := range numJobs {
+		newJob := RandAuxSigJob(
+			t, cancelChan, lfn.Some(com.Bytes()), idx,
+		)
+		jobs = append(jobs, newJob)
 	}
+
+	return signer, cancelChan, commitTx, jobs
+}
+
+// TestAuxLeafSigner tests the AuxLeafSigner implementation.
+func TestAuxLeafSigner(t *testing.T) {
+	signer, _, commitTx, jobs := setupAuxLeafSigner(t, 1)
+	defer func() {
+		require.NoError(t, signer.Stop())
+	}()
 
 	err := signer.SubmitSecondLevelSigBatch(chanState, commitTx, jobs)
 	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


tapchannel/auf_leaf_signer_test.go
```diff
@@ -42,6 +42,51 @@ var (
 	testTimeout = time.Second
 )
 
+// RandCommitmentKeyRing generates a commitment key ring used to derive HTLC
+// signatures.
+func RandCommitmentKeyRing(t *testing.T) lnwallet.CommitmentKeyRing {
```
guggero, line 47: This could be useful in other tests as well. Should we move it to `internal/test/helpers.go`?

tapchannel/auf_leaf_signer_test.go
```diff
@@ -131,6 +142,171 @@ func TestAuxLeafSigner(t *testing.T) {
...
+	)
+	cancelChan := make(chan struct{})
+
+	// Constructing multiple jobs will allow us to assert that later jobs
+	// are cancelled successfully.
+	numSigJobs := 10
+	jobs := make([]lnwallet.AuxSigJob, 0, numSigJobs)
+	outputIdxCounter := int32(0)
+
+	for range numSigJobs {
```
guggero, line 188: nit: could use loop index here instead of output index counter? Same in other the test below.

tapchannel/auf_leaf_signer_test.go
```diff
@@ -131,6 +142,171 @@ func TestAuxLeafSigner(t *testing.T) {
...
+	case <-jobs[cancelThreshold].Resp:
+		// Send the cancel signal; jobs at the end of the batch should
+		// not be processed.
+		close(cancelChan)
+	}
+
+	signer.Wg.Wait()
+
+	// Once the aux signer finishes handling the batch, the last job of the
+	// batch should have an empty response. Otherwise, the signer failed
```
guggero, line 216: nit: should say "empty response channel"?

tapchannel/aux_leaf_signer.go
```diff
@@ -240,36 +240,53 @@ func (s *AuxLeafSigner) processAuxSigBatch(chanState *channeldb.OpenChannel,
 
 	defer s.Wg.Done()
 
+	shutdownErr := fmt.Errorf("tapd is shutting down")
```
guggero, line 243: nit: Could even be declared as a public variable at the start of the file.

tapchannel/aux_leaf_signer.go
```diff
@@ -240,36 +240,53 @@ func (s *AuxLeafSigner) processAuxSigBatch(chanState *channeldb.OpenChannel,
...
 	log.Tracef("Processing %d aux sig jobs", len(sigJobs))
 
 	for idx := range sigJobs {
 		sigJob := sigJobs[idx]
 		cancelAndErr := func(err error) {
 			log.Errorf("Error processing aux sig job: %v", err)
 
-			close(sigJob.Cancel)
+			// Check that the cancel signal was not already sent
```
guggero, line 251: Since we're expected to be the only ones closing the `sigJob.Cancel` channel, we could also do this with a `sync.Once`. But I guess this works too and would also work in case someone else closes the channel.

guggero (approved):
Nice bug hunting and fix!

tapchannel/auf_leaf_signer_test.go
```diff
@@ -131,6 +142,171 @@ func TestAuxLeafSigner(t *testing.T) {
 	}
 }
 
+// TestAuxLeafSignerCancel tests that the AuxLeafSigner will handle a cancel
```
Roasbeef, line 145: If I run this vanilla in master, I get this error: 
```
 FAIL: TestAuxLeafSignerCancel (0.05s)
=== RUN   TestAuxLeafSignerCancelAndQuit
panic: close of closed channel

goroutine 5 [running]:
github.com/lightninglabs/taproot-assets/tapchannel.(*AuxLeafSigner).processAuxSigBatch.func1(...)
        /Users/roasbeef/gocode/src/github.com/lightninglabs/taproot-assets/tapchannel/aux_leaf_signer.go:250
github.com/lightninglabs/taproot-assets/tapchannel.(*AuxLeafSigner).processAuxSigBatch(0x140006c7080, 0x140001c6008, 0x140012a9220, {0x14000748008, 0xa, 0x1039c0840?})
        /Users/roasbeef/gocode/src/github.com/lightninglabs/taproot-assets/tapchannel/aux_leaf_signer.go:259 +0x588
created by github.com/lightninglabs/taproot-assets/tapchannel.(*AuxLeafSigner).SubmitSecondLevelSigBatch in goroutine 4
        /Users/roasbeef/gocode/src/github.com/lightninglabs/taproot-assets/tapchannel/aux_leaf_signer.go:104 +0xc8
exit status 2
```

tapchannel/auf_leaf_signer_test.go
```diff
@@ -131,6 +142,171 @@ func TestAuxLeafSigner(t *testing.T) {
...
+	err := signer.SubmitSecondLevelSigBatch(chanState, commitTx, jobs)
+	require.NoError(t, err)
+
+	select {
+	case <-time.After(testTimeout):
+		t.Fatalf("timeout waiting for response")
+	case <-jobs[cancelThreshold].Resp:
+		// Send the cancel signal; jobs at the end of the batch should
+		// not be processed.
+		close(cancelChan)
```
Roasbeef, line 210: Won't this always lead to a panic in the test? As `cancelChan` will be closed more than once. 

As mentioned earlier, perhaps what we want is to wrap it in some light abstraction over a `sync.Once`?

tapchannel/auf_leaf_signer_test.go
```diff
@@ -131,6 +142,171 @@ func TestAuxLeafSigner(t *testing.T) {
...
+	// a third of the batch.
+	cancelThreshold := numSigJobs / 3
+
+	err := signer.SubmitSecondLevelSigBatch(chanState, commitTx, jobs)
+	require.NoError(t, err)
+
+	select {
+	case <-time.After(testTimeout):
+		t.Fatalf("timeout waiting for response")
+	case <-jobs[cancelThreshold].Resp:
```
Roasbeef, line 207: Don't we actually want all the jobs to _share_ a response channel? Also potentially making the channel unbuffered so we can cancel as soon as the first job is received.

tapchannel/auf_leaf_signer_test.go
```diff
@@ -131,6 +142,171 @@ func TestAuxLeafSigner(t *testing.T) {
...
+		cmsg.NewAssetOutput(
+			randInputProof.Asset.ID(), randInputProof.Asset.Amount,
+			randInputProof,
+		),
+	}
+
+	com := cmsg.NewCommitment(
+		nil, nil, outgoingHtlcs, nil, lnwallet.CommitAuxLeaves{},
+	)
+	cancelChan := make(chan struct{})
```
Roasbeef, line 262: Looks like we can consolidate the boiler plate above this into a shared func for both tests.

tapchannel/auf_leaf_signer_test.go
```diff
@@ -131,6 +142,171 @@ func TestAuxLeafSigner(t *testing.T) {
...
+	err := signer.SubmitSecondLevelSigBatch(chanState, commitTx, jobs)
+	require.NoError(t, err)
+
+	select {
+	case <-time.After(testTimeout):
+		t.Fatalf("timeout waiting for response")
+	case <-jobs[cancelThreshold].Resp:
+		// Another component could have sent the cancel signal; we'll
+		// send that before the quit signal.
+		close(cancelChan)
```
Roasbeef, line 292: Same comment above re a light abstraction around `sync.Once`, if we want to hold minimal assumptions w.r.t _how_ the sig jobs can be cancelled in the first place.

tapchannel/aux_leaf_signer.go
Roasbeef: Yeah I think we need to straighten out exactly who/what can close the quit channel, otherwise we'll invariably run into a double close panic.

tapchannel/aux_leaf_signer.go
```diff
@@ -240,36 +240,53 @@ func (s *AuxLeafSigner) processAuxSigBatch(chanState *channeldb.OpenChannel,
...
 			return
-
 		default:
 		}
 
 		// If there is no commit blob, this isn't a custom channel. We
 		// still need to signal the job as done though, even if we don't
 		// have a signature to return.
 		if sigJob.CommitBlob.IsNone() {
-			sigJob.Resp <- lnwallet.AuxSigJobResp{
```
Roasbeef, line 269: So did we ever figure out for certain, why in the trace a goroutine could be blocked here, assuming the channel is always buffered?

Roasbeef (changes requested):
Nice job digging into this! 

I think the changes re the quit/cancel operations make sense, and one of the added tests does demonstrate a problem in this area.

The one thing I'm not sure if we've figured out is: why was the goroutine blocking on a channel that _should_ always be buffered?

tapchannel/auf_leaf_signer_test.go
jharveyb: Yes, that's the expected outcome - added the tests separately from the fix to show the problem if lnd hit an error and closed the channel.

tapchannel/auf_leaf_signer_test.go
jharveyb: I don't think that's needed? / don't see the benefit, we'd still have to be able  to cancel from here.

tapchannel/aux_leaf_signer.go
jharveyb: No, I imagine it's just a forced shutdown that happened to be when this send was happening. I double checked and that `Resp` chan is definitely constructed as buffered.

tapchannel/aux_leaf_signer.go
jharveyb: lnd can close the channel, it is shared amongst the 'classic' sig jobs and these aux sig jobs.

jharveyb:
Ok, updated to address initial comments.

I think this should get merged as-is and then have a follow-up PR, following the sketch here:

https://github.com/lightninglabs/taproot-assets/issues/1114#issuecomment-2343280960

tapchannel/aux_leaf_signer.go
jharveyb: Updated s.t. lnd is the _only_ one that will be closing the channel (simpler flow, having tapd close was not really needed)

tapchannel/aux_leaf_signer.go
```diff
@@ -241,43 +244,49 @@ func (s *AuxLeafSigner) processAuxSigBatch(chanState *channeldb.OpenChannel,
...
 	log.Tracef("Processing %d aux sig jobs", len(sigJobs))
-
 	for idx := range sigJobs {
 		sigJob := sigJobs[idx]
-		cancelAndErr := func(err error) {
+		respondErr := func(err error) {
 			log.Errorf("Error processing aux sig job: %v", err)
 
-			close(sigJob.Cancel)
 			sigJob.Resp <- lnwallet.AuxSigJobResp{
```
guggero, line 252: Wherever we send to a channel, we should also read on the `Quit` channel to make sure we never block on a send (even if it's buffered).

jharveyb: hmm, so if we receive `Quit` we abandon sending the err? that may then block `lnd` then IIUC.

Also, if we select on `Quit`, and we entered this function from a `Quit` case statement, we would never send the error, since we're reading an already closed channel.

guggero: Hmm, yeah, you're right. We have to assume the `Resp` channel is buffered here though (which it is), then this should never block.

guggero (approved):

tapchannel/aux_leaf_signer.go
Roasbeef: > We have to assume the Resp channel is buffered here though (which it is), then this should never block.

Yeah this is what was puzzling w.r.t the original stack trace that led us down this line of inquiry: unless some weird mutation happened, why would `tapd` block on the channel send?

Roasbeef (approved):
LGTM

PR #1119, opened 2024-09-10
Add 'include_leased' flag to ListBalances

### LINKED ISSUE #1106: [feature]: Enhance `asset_sum` report to clarify conservation
## Background
Strict, asset-amount conservation checks exists in the tapd VM to prevent undue inflation of asset quantities. However, there may be instances where the daemon reports asset amounts that appear to be double counted, even though they are being properly conserved.

### Asset Balance Clarification

To avoid confusion, it is essential to clearly delineate between payment-channel balances and L1-only asset balances. This will ensure that users have a comprehensive understanding of their asset holdings.

#### Task 1: Ensure Mutual Exclusivity
- [ ] Verify that the sum of L1-only asset balances and payment-channel balances are mutually exclusive, i.e., there is
 no overlap or double counting between the two types of balances.
  - [ ] Ensure that asset balance sum reports, sum(non-channel asset balances + local_balance)

### Comprehensive Asset Sum Calculations

To ensure accuracy in asset sum calculations, we need to investigate the comprehensiveness of these calculations.

#### Task 2: Investigate `asset sum` Calculation
- [x] https://github.com/lightninglabs/taproot-assets/issues/1106#issuecomment-2391574186 ~~Check if the `script_key_has_script_path": true` flag is being incorrectly included in the asset sum. (Intuition is no)~~
  - [x] In situ: Assets have been incorporated in the db (due to some other reason).  Confirm that these assets are not being counted erroneously in asset sum.
- [x] Verify that leased assets are not showing up in the asset sum, as they may be subject to different conservation rules. [Completed in https://github.com/lightninglabs/taproot-assets/pull/1119] 

## Background
In the scenario where two nodes have a channel
Node 1 force closes
Node 2 (the responder) -- even if it didn't have the channel's entire balance, the node insert the force close transaction, into it's own database
These inserted transaction are added to the database fields that are importing the outputs as well
While these outputs cannot be spent due to the presence of a script path.
But litd/tapd displays these as part of the node's balance today

## Solution
`make another delineated name space, to track the amounts` in order for the `wallet balance commands` to `by default not show these balance, by default`

## Considerations
These values are used during tests to assert certain outcomes have been met
`add a flag to enable reporting` in order for `assertion tests` to `have the necessary data available to run tests`

gijswijs:
This PR adds `include_leased` flag to ListBalances rpc endpoint.
With this flag you show balances that include the leased assets, which are otherwise ignored. In doing so the behavior is now synced with the `ListAssets` endpoint.

Fixes #1106 task 2 (2nd checkbox, and probably also 1st checkbox)

Fixes https://github.com/lightninglabs/taproot-assets/issues/1106


cmd/tapcli/assets.go
```diff
@@ -693,6 +698,8 @@ func listAssetBalances(ctx *cli.Context) error {
 		}
 	}
 
+	req.IncludeLeased = ctx.Bool(assetIncludeLeasedName)
```
guggero, line 701: nit: could just add this to where the struct is initialized:
```
	req := &taprpc.ListBalancesRequest{
		IncludeLeased: ctx.Bool(assetIncludeLeasedName),
	}
```

rpcserver.go
```diff
@@ -938,7 +938,7 @@ func (r *rpcServer) checkBalanceOverflow(ctx context.Context,
 	case groupPubKey != nil:
 		// Retrieve the current balance of the group.
 		balances, err := r.cfg.AssetStore.QueryAssetBalancesByGroup(
-			ctx, groupPubKey,
+			ctx, groupPubKey, false,
```
guggero, line 941: Same here, I think we'll want `true`.

tapdb/assets_store.go
```diff
@@ -1014,20 +1042,37 @@ func (a *AssetStore) QueryBalancesByAsset(ctx context.Context,
...
+	// filtering our database query understands.
+	assetBalancesFilter := QueryAssetBalancesByGroupFilters{
+		Now: sql.NullTime{
+			Time:  a.clock.Now().UTC(),
+			Valid: true,
+		},
+	}
+
+	// By default, we only show assets that are not leased.
+	if !includeLeased {
```
guggero, line 1059: Would be nice to also add some unit tests for the flag, especially around the three different values for the `leased` flag on the DB level (`NULL` = no filter, `true`  = only leased, `false` = only non-leased).

tapdb/assets_store.go
```diff
@@ -1014,20 +1042,37 @@ func (a *AssetStore) QueryBalancesByAsset(ctx context.Context,
...
+		assetBalancesFilter.Leased = sqlBool(false)
+	}
+
+	// Only show specific group if a groupKey has been passed
```
guggero, line 1063: nit: missing full stop.

itest/assertions.go
```diff
@@ -1886,7 +1886,7 @@ func AssertGenesisOutput(t *testing.T, output *taprpc.ManagedUtxo,
 }
 
 func AssertAssetBalances(t *testing.T, client taprpc.TaprootAssetsClient,
-	simpleAssets, issuableAssets []*taprpc.Asset) {
+	simpleAssets, issuableAssets []*taprpc.Asset, includeLeased bool) {
```
guggero, line 1889: Since we should now have the same flags for the `ListAssets` and `ListBalances` calls, I wonder if we should also add a check here that makes sure the sum of all assets returned by `ListAssets` matches the sum returned in `ListBalances`?

rpcserver.go
```diff
@@ -922,7 +922,7 @@ func (r *rpcServer) checkBalanceOverflow(ctx context.Context,
 	case assetID != nil:
 		// Retrieve the current asset balance.
 		balances, err := r.cfg.AssetStore.QueryBalancesByAsset(
-			ctx, assetID,
+			ctx, assetID, false,
```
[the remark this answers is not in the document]
GeorgeTsagk: shouldn't we preserve the previous behavior? unless it was wrong before, for not including leased assets in the check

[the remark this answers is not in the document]
guggero: Yes, we want the previous behavior, which means include leased assets. Which would be `true` at this level (where it means `includeLeased`).

itest/assertions.go
gijswijs: Your wish is my command.

tapdb/sqlc/queries/assets.sql
```diff
@@ -321,7 +321,18 @@ JOIN genesis_info_view
 -- around that needs to be used with this query until a sqlc bug is fixed.
 LEFT JOIN key_group_info_view
     ON assets.genesis_id = key_group_info_view.gen_asset_id
-WHERE spent = FALSE
+LEFT JOIN managed_utxos utxos
```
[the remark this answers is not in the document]
gijswijs: Correct, this was done on purpose. The initial approach was having a normal join with the `leased CASE` in the join clause, but this failed the `TestCommitBatchChainActions` unit test here: 
https://github.com/lightninglabs/taproot-assets/blob/c7fe6ef4afc9848cde8e19ee8e11f795fcaa0003/tapdb/asset_minting_test.go#L1044

The reason it failed is because that unit test expects to see the group genesis asset. But group genesis assets have `anchor_utxo_id` set to `NULL`, so they don't show up in the join unless it is a left join. Having the left join meant that I had to change the semantics of the query ever so slightly.

I now changed the join to have the semantics be consistent with `QueryAssets`while still making sure that the group genesis assets show up.

tapdb/assets_store.go
gijswijs: Added a unit test. Although the state 'only leased' can't be hit, because we never run the query with `leased = true`. 
```
if !includeLeased {
	assetBalancesFilter.Leased = sqlBool(false)
}
```
All other states are extensively tested with grouped and leased assets mixed.

tapdb/assets_store_test.go
```diff
@@ -2205,3 +2205,147 @@ func TestTransferOutputProofDeliveryStatus(t *testing.T) {
...
 	)
 }
+
+func TestQueryAssetBalances(t *testing.T) {
+	t.Parallel()
+
+	_, assetsStore, _ := newAssetStore(t)
+	ctx := context.Background()
+
+	// First, we'll generate 3 assets, two of them sharing the same anchor
```
guggero, line 2215: nit: 4 assets?

tapdb/assets_store_test.go
```diff
@@ -2205,3 +2205,147 @@ func TestTransferOutputProofDeliveryStatus(t *testing.T) {
...
+	for _, balance := range balancesByGroup {
+		balanceByGroupSum += balance.Balance
+	}
+	require.Equal(t, totalGroupedBalances, balanceByGroupSum)
+
+	// Now we lease the first asset for 1 hour. This will cause the second
+	// one also to be leased, since it's on the same anchor transaction. The
+	// second asset is in its own group, so when leased, the entire group is
+	// leased.
+	leaseOwner := fn.ToArray[[32]byte](test.RandBytes(32))
```
guggero, line 2286: :+1:

tapdb/sqlc/queries/assets.sql
```diff
@@ -321,7 +321,18 @@ JOIN genesis_info_view
 -- around that needs to be used with this query until a sqlc bug is fixed.
 LEFT JOIN key_group_info_view
     ON assets.genesis_id = key_group_info_view.gen_asset_id
-WHERE spent = FALSE
+JOIN managed_utxos utxos
+    ON (assets.anchor_utxo_id = utxos.utxo_id OR assets.anchor_utxo_id IS NULL) AND
```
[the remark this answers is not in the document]
gijswijs: The issue is that when anchoring pending assets with `AnchorPendingAssets` we select the assets on `prev_out`, but the genesis asset has a different `prev_out` than the other assets in the batch. So all assets _except_ the genesis asset is anchored.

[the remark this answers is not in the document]
gijswijs: So `addRandGroupToBatch` creates a random asset genesis and stores it with `storeGroupGenesis`. But that random asset doesn't have a managed utxo in the database. It's inserted with `upsertAssetsWithGenesis` and although that func does allow for a `utxo_id` being passed, in this case `nil` is passed.

I have changed it, so that it _also_ creates a managed utxo in the database, and then passes the `utxo_id` it gets back from that operation to `upsertAssetsWithGenesis`. It's quite involved.

Because that random asset is now a proper asset that returns in `FetchAllAssets` it does change some of the assumptions in the original failing test `TestCommitBatchChainActions`.

What this all means is that we can have proper JOINS without the trickery from before.

tapdb/asset_minting.go
```diff
@@ -1566,6 +1566,7 @@ func (a *AssetMintingStore) CommitSignedGenesisTx(ctx context.Context,
 			PrevOut:      genesisOutpoint,
 			AnchorUtxoID: sqlInt64(utxoID),
 		})
+
```
guggero, line 1569: nit: unnecessary change?

tapdb/asset_minting_test.go
```diff
@@ -149,10 +149,13 @@ func storeGroupGenesis(t *testing.T, ctx context.Context, initGen asset.Genesis,
 	upsertAsset := func(q PendingAssetStore) error {
 		_, err = maybeUpsertAssetMeta(ctx, q, &assetGen, nil)
 		require.NoError(t, err)
+		// Insert a random managed UTXO.
```
guggero, line 152: nit: new line before comment.

tapdb/asset_minting_test.go
```diff
@@ -168,6 +171,85 @@ func storeGroupGenesis(t *testing.T, ctx context.Context, initGen asset.Genesis,
 	}
 }
 
+// AddRandomManagedUTXO is a helper function that will create a random managed
```
guggero, line 174: nit: Capitalization of Godoc function name doesn't match actual function name.

tapdb/asset_minting_test.go
```diff
@@ -168,6 +171,85 @@ func storeGroupGenesis(t *testing.T, ctx context.Context, initGen asset.Genesis,
...
+	// Insert an internal key.
+	_, err = db.UpsertInternalKey(ctx, InternalKey{
+		RawKey:    randPubKey.SerializeCompressed(),
+		KeyFamily: 1,
+		KeyIndex:  2,
+	})
+	require.NoError(t, err)
+
+	// Insert the managed UTXO.
+	managedUTXO := RawManagedUTXO{
```
guggero, line 234: Nice, this looks good!

tapdb/asset_minting_test.go
```diff
@@ -790,6 +872,7 @@ type randAssetCtx struct {
 	tapSiblingBytes []byte
 	tapSiblingHash  chainhash.Hash
 	mintingBatch    *tapgarden.MintingBatch
+	randomGenesis   *asset.Genesis
```
guggero, line 875: nit: since this is the group's genesis, should we call this `groupGenesis` instead?

tapdb/asset_minting_test.go
```diff
@@ -967,13 +1051,14 @@ func TestCommitBatchChainActions(t *testing.T) {
 	)
 	require.Equal(t, txIndex, extractSqlInt32[uint32](dbGenTx.TxIndex))
 
-	// If we query for the set of all active assets, then we should get
-	// back the same number of seedlings.
+	// If we query for the set of all active assets, then we should get back
```
guggero, line 1054: :+1:

tapdb/asset_minting_test.go
```diff
@@ -995,9 +1080,15 @@ func TestCommitBatchChainActions(t *testing.T) {
 	// All the assets should also have a matching asset version as the
 	// seedlings we created.
 	mintingBatch := randAssetCtx.mintingBatch
+	randomGenesisTag := randAssetCtx.randomGenesis.Tag
 	require.True(t, fn.All(assets, func(dbAsset *asset.ChainAsset) bool {
 		seedling, ok := mintingBatch.Seedlings[dbAsset.Genesis.Tag]
 		if !ok {
+			// The only asset that doesn't have a seedling is the
```
guggero, line 1087: :+1:

guggero (approved):
LGTM, pending a rebase/squash and some more nits :tada:

tapdb/asset_minting_test.go
gijswijs: Good catch. I had it first defined someplace else.

tapdb/asset_minting_test.go
gijswijs: :+1:

itest/assertions.go
```diff
@@ -1913,7 +1913,9 @@ func AssertAssetBalances(t *testing.T, client taprpc.TaprootAssetsClient,
 
 	require.Equal(t, len(allAssets), len(assetIDBalances.AssetBalances))
 
+	var totalBalance uint64
```
GeorgeTsagk, line 1916: nit: shouldn't this commit precede the previous `itest: test balances with an without leased flag`

GeorgeTsagk (approved):
Lgtm!
very nice commit structure and thorough unit test 💯

PR #1120, opened 2024-09-10
Tap channels liquidity fixes

GeorgeTsagk:
This PR adds some small fixes for liquidity related bugs that were reported on tap channels.

### DIFF
--- a/tapchannel/aux_traffic_shaper.go
+++ b/tapchannel/aux_traffic_shaper.go
@@ -120,8 +120,8 @@ func (s *AuxTrafficShaper) HandleTraffic(_ lnwire.ShortChannelID,
 // should be handled by the traffic shaper, the HandleTraffic method should be
 // called first.
 func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
-	commitmentBlob lfn.Option[tlv.Blob],
-	linkBandwidth lnwire.MilliSatoshi) (lnwire.MilliSatoshi, error) {
+	commitmentBlob lfn.Option[tlv.Blob], linkBandwidth,
+	htlcAmt lnwire.MilliSatoshi) (lnwire.MilliSatoshi, error) {
 
 	// If the commitment or HTLC blob is not set, we don't have any
 	// information about the channel and cannot determine the available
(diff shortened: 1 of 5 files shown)


GeorgeTsagk:
"Parent" PR (itest in litd) for more context & linked issues:
https://github.com/lightninglabs/lightning-terminal/pull/842

tapchannel/aux_traffic_shaper.go
```diff
@@ -140,6 +140,33 @@ func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
 		return linkBandwidth, nil
 	}
 
+	// Get the minimum HTLC amount, which is just above dust.
+	minHtlcAmt := lnwire.NewMSatFromSatoshis(DefaultOnChainHtlcAmount)
```
guggero, line 144: Hmm, looking at the previous commit, I wondered if this value needed to be dynamic, based on the channel config's dust limit.
But luckily for `lnd` to `lnd` this value is always static: https://github.com/lightningnetwork/lnd/blob/b58fcf73fff36a0ac45e89fac2d7aa589fa6c47b/lnwallet/wallet.go#L1419

guggero (commented):
Nice, I think this should work.

tapchannel/aux_traffic_shaper.go
GeorgeTsagk: I see, perhaps worth keeping track of somehow, in case we ever have tap channel between different LN node implementations

tapchannel/commitment.go
```diff
@@ -707,6 +707,24 @@ func CreateAllocations(chanState *channeldb.OpenChannel, ourBalance,
 			allocType = CommitAllocationHtlcIncoming
 		}
 
+		dustLimit := chanState.LocalChanCfg.DustLimit
```
guggero, line 710: This can be done at the top of the `CreateAllocations` function (after where we set `leaseExpiry`), then it can be re-used for both closures.

tapchannel/aux_traffic_shaper.go
```diff
@@ -219,6 +220,14 @@ func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
 
 	mSatPerAssetUnit := quote.BidPrice
 
+	// At this point we have acquired what we need to express the asset
+	// bandwidth expressed in satoshis. Before we return the result, we need
+	// to check if the link bandwidth can afford sending a dust htlc to the
```
guggero, line 225: nit: non-dust htlc?

tapchannel/aux_traffic_shaper.go
```diff
@@ -196,6 +189,14 @@ func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
 		// means in terms of satoshis. But the satoshi amount doesn't
 		// matter too much here, we just want to signal that this
 		// channel _does_ have available bandwidth.
+
+		// Check if the current link bandwidth can afford sending out the htlc
```
guggero, line 193: nit: I think this block should probably be above the other comment in this block? Since that comment explicitly references the `btcutil.MaxSatoshi` that is returned.

guggero (approved):
Only nits, other than that LGTM :taco:

GeorgeTsagk:
`go.mod` updated to point to the latest commit of `lnd`/`0-19-staging`, which includes the fixes this PR depends on

guggero (approved):
Great fixes, thank you! LGTM :tada:

ffranr (approved):

PR #1123, opened 2024-09-16
fn+proof: use GOMAXPROCS for worker pool size

jharveyb:
Noticed while reading #675 .

The impact of a mismatch between `runtime.NumCPU()` and `GOMAXPROCS` is a known issue:

https://github.com/golang/go/issues/33803

https://github.com/uber-go/automaxprocs

Tl;dr, if a user wants to limit the OS threads used by `tapd` and sets `GOMAXPROCS`, that will have no effect right now. Reading `GOMAXPROCS` is preferred to `runtime.NumCPU()` for setting the size of a CPU-bound worker pool. This matches `runtime.NumCPU()` by default.

https://pkg.go.dev/runtime#GOMAXPROCS


ffranr (approved):

GeorgeTsagk (approved):
Lgtm

guggero (approved):

Issue #1122, opened 2024-08-29
Block RPC calls if not properly started or errored out during start


dstadulis:
In a recent user report, litd dysfunction was shown because the universe RPC call ran before the universe sub service was online.  

See TODO summary here:
https://github.com/lightninglabs/taproot-assets/issues/1122#issuecomment-2345928938

PR #1126, opened 2024-09-17
Startup related fixes

GeorgeTsagk:
This PR fixes some startup related bugs that were discovered when integrating tapd with litd.

Read more [here](https://github.com/lightninglabs/taproot-assets/issues/1122)

Closes https://github.com/lightninglabs/taproot-assets/issues/1122

### DIFF
--- a/rpcperms/interceptor.go
+++ b/rpcperms/interceptor.go
@@ -413,6 +413,11 @@ func (r *InterceptorChain) rpcStateUnaryServerInterceptor() grpc.UnaryServerInte
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo,
 		handler grpc.UnaryHandler) (interface{}, error) {
 
+		if info.Server == nil {
+			return nil, fmt.Errorf("cannot handle call, server " +
+				"not ready")
+		}
+
 		if err := r.checkRPCState(info.Server); err != nil {
 			return nil, err
 		}
@@ -426,6 +431,10 @@ func (r *InterceptorChain) rpcStateUnaryServerInterceptor() grpc.UnaryServerInte
 func (r *InterceptorChain) rpcStateStreamServerInterceptor() grpc.StreamServerInterceptor {
 	return func(srv interface{}, ss grpc.ServerStream,
 		info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+
+		if srv == nil {
+			return fmt.Errorf("srv is nil, can't check RPC state")
+		}
 
 		if err := r.checkRPCState(srv); err != nil {
 			return err
--- a/server.go
+++ b/server.go
@@ -90,6 +90,16 @@ func (s *Server) UpdateConfig(cfg *Config) {
 //
 // NOTE: the rpc server is not registered with any grpc server in this function.
 func (s *Server) initialize(interceptorChain *rpcperms.InterceptorChain) error {
+	var ready bool
+
+	// If by the time this function exits we haven't yet given the ready
+	// signal, we detect it here and signal that the daemon should quit.
+	defer func() {
+		if !ready {
+			close(s.quit)
+		}
+	}()
+
 	// Show version at startup.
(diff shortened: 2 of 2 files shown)


server.go
```diff
@@ -702,11 +702,16 @@ func (s *Server) waitForReady() error {
...
-		return nil
-
 	case <-s.quit:
 		return fmt.Errorf("tapd is shutting down")
+
+	default:
+		select {
+		case <-s.ready:
+			return nil
+		case <-s.quit:
```
guggero, line 712: I think this doesn't solve the `tapd: don't set initialized=true when startup (StartAsSubserver) fails` part of the issue. We don't close the `quit` channel when we fail to start there (which is the main entrypoint for `litd`).

rpcperms/interceptor.go
```diff
@@ -413,6 +413,12 @@ func (r *InterceptorChain) rpcStateUnaryServerInterceptor() grpc.UnaryServerInte
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo,
 		handler grpc.UnaryHandler) (interface{}, error) {
 
+		if info.Server == nil {
+			return nil, fmt.Errorf("server is nil, can't check " +
+				"RPC state")
+
```
guggero, line 419: nit: linter complains about superfluous empty line.

rpcperms/interceptor.go
```diff
@@ -413,6 +413,12 @@ func (r *InterceptorChain) rpcStateUnaryServerInterceptor() grpc.UnaryServerInte
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo,
 		handler grpc.UnaryHandler) (interface{}, error) {
 
+		if info.Server == nil {
+			return nil, fmt.Errorf("server is nil, can't check " +
```
guggero, line 417: Can probably be a bit less specific in the error message here, perhaps just "cannot handle call, server not ready"?

server.go
```diff
@@ -702,11 +702,16 @@ func (s *Server) waitForReady() error {
 	// this part of the code, the called component will handle the quit
 	// signal.
 	select {
-	case <-s.ready:
-		return nil
-
 	case <-s.quit:
 		return fmt.Errorf("tapd is shutting down")
+
+	default:
```
guggero, line 708: Nice fix. This probably explains one class of potential errors we had.

server.go
```diff
@@ -702,11 +702,16 @@ func (s *Server) waitForReady() error {
...
+		select {
+		case <-s.ready:
```
guggero, line 710: Maybe mention here that we now _wait_ until the server becomes ready or shuts down (meaning this _is_ blocking).

server.go
```diff
@@ -824,17 +824,25 @@ func (s *Server) Name() protofsm.EndpointName {
 // CanHandle returns true if the target message can be routed to this endpoint.
 //
 // NOTE: This method is part of the protofsm.MsgEndpoint interface.
-func (s *Server) CanHandle(msg protofsm.PeerMsg) bool {
-	<-s.ready
+func (s *Server) CanHandle(msg protofsm.PeerMsg) (bool, error) {
+	err := s.waitForReady()
```
guggero, line 828: Depending on the discussion on the `lnd` PR it _might_ be okay to just return `false` if `s.waitForReady()` returns an error. But calling `s.waitForReady()` should definitely be done so we don't block indefinitely.

guggero (commented):
Nice fixes, thanks. Needs some additional checks though.

server.go
GeorgeTsagk: Ah right, someone waiting at the `waitForReady` endpoint will stay hanging

Will enhance the `StartAsSubserver` method to conditionally close the `s.quit` chan, if initialization faced any err

server.go
GeorgeTsagk: Right, we don't really want unnecessary error decorations in this interface.

Instead can just log a debug msg and return false

Roasbeef (approved):
LGTM 🎨

Issue #871, opened 2024-04-04
[feature]: update RFQ service to align blip


ffranr:
### Objective
The goal of this issue is to revise the RFQ (Request for Quote) service following the recent updates in the Tap blip, detailed in [this pull request](https://github.com/Roasbeef/blips/pull/2).

### Current Implementation
The RFQ service currently uses two separate wire message types to differentiate quote requests for buying and selling. Whereas, the BLIP uses a unified request wire message type in conjunction with a new swap direction field.

### Required Changes
To ensure compatibility with the updated BLIP standards, our RFQ service needs to reduce its wire message types to three, incorporating the swap direction field as necessary. The current RFQ wire message types are listed [here](https://github.com/lightninglabs/taproot-assets/blob/c723abd3c9db8a6a2f3f1eaa85ce5aefb52c8170/rfqmsg/messages.go#L46-L66).

### Implementation Notes
- We should wait until https://github.com/Roasbeef/blips/pull/2 is merged before starting this issue. That PR may have implications for how this issue is resolved.
- We can retain the  exported `rfqmsg` types `BuyRequest`, `SellRequest`, `BuyAccept`, and `SellAccept`. Only the wire messages need to change.

Roasbeef:
Now fully up to date.

PR #1128, opened 2024-09-18
RFQ session lookup during quote accept message parsing

ffranr:
Future RFQ quote accept messages will not contain enough information for the RFQ service to classify them as buy or sell. Therefore, before these messages can be fully interpreted, the corresponding quote request message must be retrieved and inspected.

This commit modifies the parsing of incoming quote accept messages so they can be accurately classified as buy or sell by looking up the associated quote request message.

As a beneficial side effect, parsed quote accept message fields are now fully populated within the parsing routine.

This PR is not a breaking change. It's related to https://github.com/lightninglabs/taproot-assets/issues/871


rfq/stream.go
```diff
@@ -92,8 +92,15 @@ func NewStreamHandler(ctx context.Context,
...
+	// Define a function to look up an RFQ session. It is necessary to
+	// utilize the corresponding outgoing request when parsing some
+	// incoming wire messages.
+	sessionLookup := func(id rfqmsg.ID) (rfqmsg.OutgoingMsg, bool) {
+		return h.outgoingRequests.Load(id)
+	}
+
 	// Parse the wire message as an RFQ message.
-	msg, err := rfqmsg.NewIncomingMsgFromWire(wireMsg)
+	msg, err := rfqmsg.NewIncomingMsgFromWire(wireMsg, sessionLookup)
```
guggero, line 103: Could we directly pass in a reference to `h.outgoingRequests.Load` here, instead of creating the `sessionLookup` wrapper? Or does that not work because of type assertions?

rfqmsg/accept.go
```diff
@@ -248,17 +250,25 @@ func NewIncomingAcceptFromWire(wireMsg WireMessage) (IncomingMsg, error) {
...
+	switch typedRequest := request.(type) {
+	case *BuyRequest:
+		return newBuyAcceptFromWireMsg(wireMsg, msgData, *typedRequest)
+	case *SellRequest:
+		return newSellAcceptFromWireMsg(wireMsg, msgData, *typedRequest)
 	}
 
-	// Otherwise, this is a sell request.
-	return newSellAcceptFromWireMsg(wireMsg, msgData)
+	return nil, fmt.Errorf("unknown request type for incoming accept "+
```
guggero, line 272: nit: since there's now a `switch`, this could be the `default` case.

rfqmsg/messages.go
```diff
@@ -90,13 +91,28 @@ type WireMessage struct {
...
+
 	switch wireMsg.MsgType {
 	case MsgTypeRequest:
 		return NewIncomingRequestFromWire(wireMsg)
+
 	case MsgTypeAccept:
-		return NewIncomingAcceptFromWire(wireMsg)
+		// A session lookup is necessary when parsing quote accept type
+		// messages.
+		if sessionLookup == nil {
```
guggero, line 109: nit: move this check to where we actually access/use the callback function?

guggero (approved):
Nice refactor! Will come in handy with the following RFQ message changes, I'm sure.

rfq/stream.go
ffranr: that works! thanks for the suggestion

GeorgeTsagk (approved):
:sparkles:

PR #1130, opened 2024-09-19
[custom channels¨

guggero:


guggero:
Similar to https://github.com/lightninglabs/loop/pull/828, I bumped to `lndclient v0.18.4-0` which references https://github.com/lightningnetwork/lnd/pull/9183 (commit https://github.com/lightningnetwork/lnd/pull/9183/commits/ca3bde901eb86638256a72fd17b75ce5951d119f) which should remain stable, so we can start the main review work here.

rfqmsg/messages.go
```diff
@@ -32,7 +33,27 @@ func (id ID) Scid() SerialisedScid {
...
+	}
+
+	// Normalize the value relative to min, then wrap within blockRange, and
+	// finally shift back by min. Generated by ChatGPT and I couldn't find a
+	// better way to do this.
+	blockRange := maxBlock - minBlock + 1
+	scid.BlockHeight = ((scid.BlockHeight-minBlock)%blockRange+blockRange)%
+		blockRange + minBlock
+
+	return SerialisedScid(scid.ToUint64())
```
ffranr: I think this shifting will need to be described in the BLIP:
```
Given a valid `rfq_id`, the RFQ specific SCID `tap_rfq_scid` is defined by
taking the last `8` bytes of the `rfq_id` and interpreting them as a 64-bit
integer.
```

It would probably be simpler at this point to just generate a SCID independent of RFQ ID and pass it to our peer in the RFQ request message.

There might be an advantage in separating the RFQ message ID from the SCID. We might be able to re-use the same SCID across multiple quotes for example. So every payment would correspond to a single RFQ quote but then multiple payments can share the same SCID.

There may be an advantage there when we come up with a formulation that avoids quote re-use.

rfqmsg/messages_test.go
```diff
@@ -0,0 +1,129 @@
...
+	return intToID(scid.ToUint64())
+}
+
+// TestIDSerializedScid tests that an RFQ ID can be correctly converted into a
+// custom SCID alias that lies within the valid SCID block height range.
+func TestIDSerializedScid(t *testing.T) {
+	tests := []struct {
+		id       ID
+		expected lnwire.ShortChannelID
+	}{
```
ffranr, line 28: Consider adding a `label` field to describe what the test case is trying to investigate.

rfqmsg/records.go
```diff
@@ -137,6 +139,31 @@ func (h *Htlc) Bytes() []byte {
...
+
+	for idx, balance := range h.Balances() {
+		j.Balances[idx] = &JsonHtlcBalance{
+			AssetID: hex.EncodeToString(balance.AssetID.Val[:]),
+			Amount:  balance.Amount.Val,
+		}
+	}
+
+	return json.Marshal(j)
+}
```
ffranr, line 165: Do we need unit tests for this?

tapchannel/aux_closer.go
```diff
@@ -162,8 +162,8 @@ func fundingSpendWitness() lfn.Result[wire.TxWitness] {
 	)
 	ctrlBlockBytes, err := ctrlBlock.ToBytes()
 	if err != nil {
-		return lfn.Errf[wire.TxWitness]("unable to serialize control "+
-			"block: %w", err)
+		return lfn.Err[wire.TxWitness](fmt.Errorf("unable to "+
+			"serialize control block: %w", err))
```
ffranr, line 166: It seems to me that `lfn.Errf` was cleaner? Why are we going to `lfn.Err` here?

> This commit unifies the use of the lfn.Result return type as a
preparation for the next commit, where we change a bunch of
function/method signatures to use the Result type as well.

The next commit doesn't do that.

rfqmsg/messages.go
guggero: I don't think separating the two values is a good idea at this point. I went with the approach to generate random IDs until they fall within the correct range.

So the bLIP could just say: "The `rfq_id` must be drawn in a way that the derived `tap_rfq_scid` falls within the correct range for an SCID alias (`16_000_000 <= block height < 16_250_000`).

tapchannel/aux_closer.go
guggero: My IDE shows an "issue" with the code otherwise: `This verb can be used only in 'fmt. Errorf' calls (%w) ` which annoys me quite a bit...
Also the use of `lnf.Err()` vs. `lnf.Errf()` wasn't really consistent.
But I'll change it back if you feel strongly about it...

Updated the commit message to reference the correct commit that introduces the changes.

ffranr: We can exclude `lnf.Errf` from that IDE check by clicking on the lightbulb:

image

It removes all formatting checks for `lnf.Errf`, but won't we only use `%w` mostly anyway perhaps.

Keeping `lfn.Errf` does seem worth it to me because we wrap errors very often in our codebase. And if we can avoid the internal `fmt.Errorf(` then I think our code would be more clear and succinct.

ffranr (approved):

tapchannel/aux_closer.go
guggero: Thanks for the tip. I reverted the `lnf.Err(fmt.Errorf(...))` occurrences with `lnf.Errorf(...)` again.

GeorgeTsagk (approved):
Lgtm https://github.com/lightninglabs/taproot-assets/pull/1130/commits/fc1a4309c5095e8371cbb90bca07daa2dee8213a

i think the htlc intercept itest might be flaky, kicked it

Is this PR going to get more commits as we move forward, or is this it's final-ish state?

guggero:
> Is this PR going to get more commits as we move forward, or is this it's final-ish state?

There will be one more round of `go.mod` bumps across all projects once the final/tagged version of `v0.18.4-beta` is out. But this PR is now complete as is, as soon as the CI is fixed (which I will look into momentarily).

guggero:
I had to add 2 commits to address some of the RFQ changes (and to make the itests pass with the new way asset to BTC conversions now happen with more precision).
Re-requesting review from @ffranr with the ask to just take a look at the last two commits.

ffranr (approved):

Issue #1028, opened 2024-07-16
[feature]: Enhance RFQ system to accept quotes by group key


dstadulis:
Currently the RFQ system handles quote specified by asset ID but ultimately the market will submit and reply to quotes by group key functionality.  This is a place holder issue to be shaped further

Related to this issue is the same need to expand a feature to group key exists in: https://github.com/lightninglabs/taproot-assets/issues/291

ffranr:
Latest in standup is that we may punt this until we add asset group key support across the board for tap payment channels.

PR #1132, opened 2024-09-20
Limit number of HTLCs in custom channel

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1149

Depends on https://github.com/lightninglabs/taproot-assets/pull/1130.

Addresses comment in https://github.com/lightningnetwork/lnd/pull/9072#discussion_r1765319065.

### DIFF
--- a/psbt_channel_funder.go
+++ b/psbt_channel_funder.go
@@ -156,6 +156,16 @@ func (l *LndPbstChannelFunder) OpenChannel(ctx context.Context,
 	}
 }
 
+// ChannelAcceptor is used to accept and potentially influence parameters of
+// incoming channels.
+func (l *LndPbstChannelFunder) ChannelAcceptor(ctx context.Context,
+	acceptor lndclient.AcceptorFunction) (chan error, error) {
+
+	return l.lnd.Client.ChannelAcceptor(
+		ctx, tapchannel.DefaultTimeout/2, acceptor,
+	)
+}
+
 // A compile-time check to ensure that LndPbstChannelFunder fully implements
 // the tapchannel.PsbtChannelFunder interface.
 var _ tapchannel.PsbtChannelFunder = (*LndPbstChannelFunder)(nil)
--- a/psbt_channel_funder.go
+++ b/psbt_channel_funder.go
@@ -97,9 +97,8 @@ func (l *LndPbstChannelFunder) OpenChannel(ctx context.Context,
 	// We'll map our high level params into a request for a: private,
 	// taproot channel, that uses the PSBT funding flow.
 	taprootCommitType := lnrpc.CommitmentType_SIMPLE_TAPROOT_OVERLAY
-	openChanStream, errChan, err := l.lnd.Client.OpenChannelStream(
-		ctx, route.NewVertex(&req.PeerPub), req.ChanAmt, req.PushAmt,
-		true, lndclient.WithCommitmentType(&taprootCommitType),
+	channelOpenOptions := []lndclient.OpenChannelOption{
+		lndclient.WithCommitmentType(&taprootCommitType),
 		lndclient.WithFundingShim(&lnrpc.FundingShim{
 			Shim: &lnrpc.FundingShim_PsbtShim{
 				PsbtShim: &lnrpc.PsbtShim{
@@ -110,6 +109,20 @@ func (l *LndPbstChannelFunder) OpenChannel(ctx context.Context,
 			},
 		}),
 		lndclient.WithRemoteReserve(CustomChannelRemoteReserve),
+	}
+
+	// Limit the number of HTLCs that can be added to the channel by the
+	// remote party.
+	if req.RemoteMaxHtlc > 0 {
+		channelOpenOptions = append(
+			channelOpenOptions,
+			lndclient.WithRemoteMaxHtlc(req.RemoteMaxHtlc),
+		)
+	}
+
+	openChanStream, errChan, err := l.lnd.Client.OpenChannelStream(
+		ctx, route.NewVertex(&req.PeerPub), req.ChanAmt, req.PushAmt,
+		true, channelOpenOptions...,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("unable to open channel with "+
--- a/tapchannel/aux_funding_controller.go
+++ b/tapchannel/aux_funding_controller.go
@@ -20,6 +20,7 @@ import (
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/davecgh/go-spew/spew"
+	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/taproot-assets/address"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/commitment"
@@ -52,6 +53,28 @@ const (
 	// ackTimeout is the amount of time we'll wait to receive the protocol
 	// level ACK from the remote party before timing out.
 	ackTimeout = time.Second * 30
+
+	// maxNumAssetIDs is the maximum number of fungible asset pieces (asset
+	// IDs) that can be committed to a single channel. The number needs to
+	// be limited to prevent the number of required HTLC signatures to be
+	// too large for a single CommitSig wire message to carry them. This
+	// value is tightly coupled with the number of HTLCs that can be added
+	// to a channel at the same time (maxNumHTLCs). The values were
+	// determined with the TestMaxCommitSigMsgSize test in
+	// aux_leaf_signer_test.go then a set was chosen that would allow for
+	// a decent number of HTLCs (and also a number that is divisible by two
+	// because each side will only be allowed to add half of the total).
+	maxNumAssetIDs = 3
+
+	// maxNumHTLCs is the maximum number of HTLCs there can be in an asset
+	// channel to avoid the number of signatures exceeding the maximum
+	// message size of a CommitSig message. See maxNumAssetIDs for more
+	// information.
+	maxNumHTLCs = 166
+
+	// maxNumHTLCsPerParty is the maximum number of HTLCs that can be added
+	// by a single party to a channel.
+	maxNumHTLCsPerParty = maxNumHTLCs / 2
 )
 
 // ErrorReporter is used to report an error back to the caller and/or peer that
@@ -138,6 +161,11 @@ type PsbtChannelFunder interface {
 	// process. Afterward, the funding transaction should be signed and
 	// broadcast.
 	OpenChannel(context.Context, OpenChanReq) (AssetChanIntent, error)
+
+	// ChannelAcceptor is used to accept and potentially influence
+	// parameters of incoming channels.
+	ChannelAcceptor(ctx context.Context,
+		acceptor lndclient.AcceptorFunction) (chan error, error)
 }
 
 // TxPublisher is an interface used to publish transactions.
@@ -222,6 +250,9 @@ type FundingControllerCfg struct {
 	// FeatureBits is used to verify that the peer has the required feature
 	// to fund asset channels.
 	FeatureBits FeatureBitVerifer
+
+	// ErrChan is used to report errors back to the main server.
+	ErrChan chan<- error
 }
 
 // bindFundingReq is a request to bind a pending channel ID to a complete aux
@@ -297,6 +328,36 @@ func (f *FundingController) Start() error {
 
 	f.Wg.Add(1)
 	go f.chanFunder()
+
+	f.Wg.Add(1)
+	go func() {
+		defer f.Wg.Done()
+
+		ctx, cancel := f.WithCtxQuitNoTimeout()
+		defer cancel()
+
+		errChan, err := f.cfg.ChannelFunder.ChannelAcceptor(
+			ctx, f.channelAcceptor,
+		)
+		if err != nil {
+			err = fmt.Errorf("unable to start channel acceptor: %w",
+				err)
+			f.cfg.ErrChan <- err
+			return
+		}
+
+		// We'll accept channels for as long as the funding controller
+		// is running or until we receive an error.
+		select {
+		case err := <-errChan:
+			err = fmt.Errorf("channel acceptor error: %w", err)
+			f.cfg.ErrChan <- err
+
+		case <-f.Quit:
+			log.Infof("Stopping channel acceptor, funding " +
+				"controller shutting down")
+		}
+	}()
 
 	return nil
 }
@@ -1008,10 +1069,11 @@ func (f *FundingController) completeChannelFunding(ctx context.Context,
 	// Now that we have the initial PSBT template, we can start the funding
 	// flow with lnd.
 	fundingReq := OpenChanReq{
-		ChanAmt: 100_000,
-		PushAmt: fundingState.pushAmt,
-		PeerPub: fundingState.peerPub,
-		TempPID: fundingState.pid,
+		ChanAmt:       100_000,
+		PushAmt:       fundingState.pushAmt,
+		PeerPub:       fundingState.peerPub,
+		TempPID:       fundingState.pid,
+		RemoteMaxHtlc: maxNumHTLCsPerParty,
 	}
 	assetChanIntent, err := f.cfg.ChannelFunder.OpenChannel(ctx, fundingReq)
 	if err != nil {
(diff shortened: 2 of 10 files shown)


guggero:
Calculated max values from `TestMaxCommitSigMsgSize`:

```
    aux_leaf_signer_test.go:111: Last valid commit sig msg size with: numAssetIDs=0, numHTLCs=966
    aux_leaf_signer_test.go:103: Last valid commit sig msg size with: numAssetIDs=1, numHTLCs=373
    aux_leaf_signer_test.go:103: Last valid commit sig msg size with: numAssetIDs=2, numHTLCs=231
    aux_leaf_signer_test.go:103: Last valid commit sig msg size with: numAssetIDs=3, numHTLCs=166
    aux_leaf_signer_test.go:103: Last valid commit sig msg size with: numAssetIDs=4, numHTLCs=130
    aux_leaf_signer_test.go:103: Last valid commit sig msg size with: numAssetIDs=5, numHTLCs=107
    aux_leaf_signer_test.go:103: Last valid commit sig msg size with: numAssetIDs=6, numHTLCs=91
    aux_leaf_signer_test.go:103: Last valid commit sig msg size with: numAssetIDs=7, numHTLCs=79
    aux_leaf_signer_test.go:103: Last valid commit sig msg size with: numAssetIDs=8, numHTLCs=70
    aux_leaf_signer_test.go:103: Last valid commit sig msg size with: numAssetIDs=9, numHTLCs=63
    aux_leaf_signer_test.go:103: Last valid commit sig msg size with: numAssetIDs=10, numHTLCs=57
```

tapchannel/aux_funding_controller.go
```diff
@@ -1417,6 +1484,26 @@ func (f *FundingController) processFundingReq(fundingFlows fundingFlowIndex,
...
+	// We need to limit the number of different fungible assets (asset IDs)
+	// we allow to be commited to a single channel. This is to make sure we
+	// have a decent number of HTLCs available. See Godoc of maxNumAssetIDs
+	// for more information.
+	//
+	// TODO(guggero): This following code is obviously wrong and needs to be
+	// changed when we support committing fungible assets into a channel. To
+	// avoid this TODO from being overlooked, we add a dummy implementation
+	// with a condition that currently will never be true (since there's
+	// only a single vPacket being selected currently anyway).
```
dstadulis, line 1496: :+1:

tapchannel/aux_funding_controller.go
```diff
@@ -51,6 +52,28 @@ const (
 	// ackTimeout is the amount of time we'll wait to receive the protocol
 	// level ACK from the remote party before timing out.
 	ackTimeout = time.Second * 30
+
+	// maxNumAssetIDs is the maximum number of fungible asset pieces (asset
+	// IDs) that can be committed to a single channel. The number needs to
```
dstadulis, line 57: In the near future, the design goal will be to have [group-key based invoices.](https://github.com/lightninglabs/taproot-assets/issues/1028).  Does this future inform any design choices?

Thinking outloud:
If an invoice is generated with a group-key,
- the number of asset IDs, associated with that group-key, might be in excess of `maxNumAssetIDs`

Doesn't seem like the `maxNumAssetIDs` effects the ability to invoice/RFQ via group key but might constrain route construction for the sender.

Let's discuss how to handle the fact that

tapchannel/aux_funding_controller.go
```diff
@@ -51,6 +52,28 @@ const (
...
+
+	// maxNumHTLCs is the maximum number of HTLCs there can be in an asset
+	// channel to avoid the number of signatures exceeding the maximum
+	// message size of a CommitSig message. See maxNumAssetIDs for more
+	// information.
+	maxNumHTLCs = 166
+
+	// maxNumHTLCsPerParty is the maximum number of HTLCs that can be added
+	// by a single party to a channel.
+	maxNumHTLCsPerParty = maxNumHTLCs / 2
```
dstadulis, line 76: Would allowing these values to be configured and setting these values during funding negotiation be possible?  Even if these values are shipped as a MVP.

The motivation being, senders and receivers may have different needs, some might desire numerous AssetIDs but minimal concurrent HTLCs or the converse, numerous concurrent HTLCs and minimal number of AssetIDs.
Maybe these values could be the defaults

tapchannel/aux_leaf_signer_test.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// TestMaxCommitSigMsgSize attempts to find values for the max number of asset
+// IDs we want to allow per channel and the resulting maximum number of HTLCs
+// that channel can allow. The maximum number of different asset IDs that can be
+// committed to a channel directly impacts the number of HTLCs that can be
+// created on that channel, because we have a limited message size to exchange
+// the second-stage HTLC signatures. The goal of this test is to find the right
+// number of asset IDs we should allow per channel to still give us a reasonable
+// amount of HTLCs.
```
dstadulis, line 89: ❤️ 🧠

dstadulis (commented):
Added a few design-option questions

tapchannel/aux_funding_controller.go
```diff
@@ -51,6 +52,28 @@ const (
...
+	// IDs) that can be committed to a single channel. The number needs to
+	// be limited to prevent the number of required HTLC signatures to be
+	// too large for a single CommitSig wire message to carry them. This
+	// value is tightly coupled with the number of HTLCs that can be added
+	// to a channel at the same time (maxNumHTLCs). The values were
+	// determined with the TestMaxCommitSigMsgSize test in
+	// aux_leaf_signer_test.go then a set was chosen that would allow for
+	// a decent number of HTLCs (and also a number that is divisible by two
+	// because each side will only be allowed to add half of the total).
+	maxNumAssetIDs = 3
```
Roasbeef, line 66: Why do we need to limit this, instead of just the total amount of HTLCs?

Roasbeef: In that if we have 10 passive asset IDs in a channel, there're no issues as long as the total amount of HTLCs isn't exceeded.

dstadulis: Given a certain number of asset IDs would a proof-size limit be encountered?

Roasbeef: IIUC no, the real limit is the HTLCs. This PR seems to take an approach to optimize equal HTLC utilization across all asset IDs, but I think we can just limit HLTCs instead. 

If we move the TAP level HTLC sigs to a new message, then we don't need a limit at all. As we'll fall back to the existing limit at the base LN level.

Roasbeef:
Ok, discussed offline a bit, and I understand now the need to limit on the asset ID basis. Rather than asset ID, it's more accurate to say that we need to limit the total number of UTXOs that can exist within a funding output. Each time we create an HTLC, we may need to reference one or more asset UTXOs from the funding output. Each of those means a input referenced, which itself needs a signature. As a result, a single HTLC may actually need several HTLCs to be transmitted over the wire. Without moving to a more effecient encoding (not much room can be gained), or to a new way of transmitting the signatures, then we need to limit the total amount of asset UTXOs that can reside within a funding output, or we run into message size limits on the protocol level. 

One alternative path is to add the equivalent of `SIGHASH_NOINPUT` on the protocol layer. If we add this, then when we send HTLCs, we can actually send the second-level signatrues alongside them, as we just need the pkScript to be stable, which is the case for the _next_ commitment. I think this breaks down though, as for the commitment after that, a different per-commitment-secret is used (new revocation secret), so that original signature can no longer be used.

guggero:
Rebased to make sure CI runs against new `litd` branch. Also taking out of draft state since the discussions showed we want this safety feature as a short-term stop-gap solution.

tapchannel/aux_leaf_signer_test.go
```diff
@@ -0,0 +1,166 @@
...
+func TestMaxCommitSigMsgSize(t *testing.T) {
+	const maxNumAssetIDs = 10
```
Roasbeef, line 91: Will we remove this test before merge? 

I guess we can leave it in and add a `Skipf`?

psbt_channel_funder.go
```diff
@@ -110,6 +110,7 @@ func (l *LndPbstChannelFunder) OpenChannel(ctx context.Context,
 			},
 		}),
 		lndclient.WithRemoteReserve(CustomChannelRemoteReserve),
+		lndclient.WithRemoteMaxHtlc(req.RemoteMaxHtlc),
```
Roasbeef, line 113: Should we only set it if the value is above zero? 

Or I guess lnd will just pick that up as not set, and apply the default.

tapchannel/aux_funding_controller.go
```diff
@@ -1682,6 +1744,42 @@ func (f *FundingController) chanFunder() {
...
+
+	// Ignore any non-asset channels, just accept them.
+	if *req.CommitmentType != lnwallet.CommitmentTypeSimpleTaprootOverlay {
+		return &lndclient.AcceptorResponse{
+			Accept: true,
+		}, nil
+	}
+
+	// Reject custom channels that don't observe the max HTLC limit.
+	if req.MaxAcceptedHtlcs != maxNumHTLCsPerParty {
```
Roasbeef, line 1787: Should we relax the constraint here? So rather than an exact match, we reject any value _above_ `maxNumHTLCsPerParty`.

tapchannel/aux_funding_controller.go
```diff
@@ -1497,6 +1497,26 @@ func (f *FundingController) processFundingReq(fundingFlows fundingFlowIndex,
...
+	// TODO(guggero): This following code is obviously wrong and needs to be
+	// changed when we support committing fungible assets into a channel. To
+	// avoid this TODO from being overlooked, we add a dummy implementation
+	// with a condition that currently will never be true (since there's
+	// only a single vPacket being selected currently anyway).
+	assetIDSet := lfn.NewSet[asset.ID]()
+	for _, out := range fundingVpkt.VPacket.Outputs {
+		assetIDSet.Add(out.Asset.ID())
+	}
+	if len(assetIDSet.ToSlice()) > maxNumAssetIDs {
```
Roasbeef, line 1514: Can use `assetIDSet.Size()` here.

tapchannel/aux_funding_controller.go
```diff
@@ -1497,6 +1497,26 @@ func (f *FundingController) processFundingReq(fundingFlows fundingFlowIndex,
 		}
 	}()
 
+	// We need to limit the number of different fungible assets (asset IDs)
+	// we allow to be commited to a single channel. This is to make sure we
+	// have a decent number of HTLCs available. See Godoc of maxNumAssetIDs
+	// for more information.
+	//
+	// TODO(guggero): This following code is obviously wrong and needs to be
```
Roasbeef, line 1505: Is this TODO still applicable?

psbt_channel_funder.go
guggero: I guess that's what would happen, yes. But just to make sure I added a condition.

tapchannel/aux_funding_controller.go
guggero: Yes, we need to actually count the number of different asset IDs (tranches) that are used for funding a channel, once we allow grouped assets to be used. So this will change in 0.6.

tapchannel/aux_leaf_signer_test.go
```diff
@@ -0,0 +1,172 @@
...
+			"e49e2a25e92208678a2df86970da91b03a8af8815a8a60498b35" +
+			"8daf560b347aa557",
+	)
+	testSchnorrSig, _ = lnwire.NewSigFromSchnorrRawSignature(
+		testSchnorrSigStr,
+	)
+
+	startNumHTLCs = 5
+	endNumHTLCs   = input.MaxHTLCNumber
+)
```
ffranr, line 48: I recommend adding comments to explain the purpose of these fields. For instance, what does `testRBytes` represent?

tapchannel/aux_leaf_signer_test.go
ffranr: I vote for leaving in and skip. It's probably easiest to find here IMO.

tapchannel/aux_leaf_signer_test.go
ffranr: Does `startNumHTLCs` really need a global scope? I think it's only used in `TestMaxCommitSigMsgSize`.

tapchannel/aux_leaf_signer_test.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	commitSig := cmsg.NewCommitSig(assetSigs)
+
+	var buf bytes.Buffer
+	err = commitSig.Encode(&buf)
+	require.NoError(t, err)
+
+	msg.CustomRecords = lnwire.CustomRecords{
+		65543: buf.Bytes(),
```
ffranr, line 168: There must be a const for 65543 somewhere? I couldn't find one. But I think it's the custom records TLV int lower bound.

tapchannel/aux_funding_controller.go
ffranr: This conversation was resolved, but I agree with what Roas says above. Why are we moving forward with the tap level HTLC limitation in this PR?

ffranr (commented):
Clean commits! Thanks.

Why are we limiting HTLCs based on the TAP level content? Basically, what's the resolution here: https://github.com/lightninglabs/taproot-assets/pull/1132/commits/f807b5ff43b4769c5d6a0a31b8d03288b509d6f0#r1776473336

tapchannel/aux_funding_controller.go
guggero: I resolved the comment based on this: https://github.com/lightninglabs/taproot-assets/pull/1132#issuecomment-2399339907

Tried my best to explain why this is needed in the code comments, but let me try again:
 - By our definition, within an on-chain transaction of multiple different involved asset IDs (e.g. multiple tranches of the same grouped asset), each asset ID needs to get its own virtual transaction.
 - A virtual transaction can combine multiple inputs of the _same_ asset ID into a single output. But if there are multiple input from different asset IDs (and because of that multiple virtual transactions), that also means there are multiple asset outputs in such a transaction.
 - If multiple different asset IDs are committed to a TAP channel, even if the on-chain funding output is one _logical_ shared UTXO, there would be an asset-level UTXO for each asset ID.
 - Depending on the amounts of those asset-level UTXOs and the size of the HTLC in question, each HTLC in a channel commitment transaction would spend between one and all of these UTXOs, therefore potentially requiring multiple asset-level signatures per HTLC.
 - We can't predict how many "pieces" of the channel inputs are consumed by each HTLC, so we have to make some sort of worst-case assumption. That worst case assumption is that each HTLC would need to carry an asset-level signature for each of the spent asset ID "piece". This is obviously an incorrect assumption as the worst-case can't really ever be true. But because we can't easily come up with a heuristic that always works, we might as well use the worst case and draw our conclusions from there.
 - All Lightning p2p messages are limited to 65k payload.
 - The question with everything above now is: How many worst-case asset HTLCs can we fit into a single [`commitment_signed`](https://github.com/lightning/bolts/blob/master/02-peer-protocol.md#the-commitment_signed-message) message? That's what the test answers.
 - Because there currently is no way to reject accepting an HTLC when noticing it would be too large at run time, we have to agree on a maximum number of in-flight HTLCs upfront, which means at channel open time.

I hope the above makes sense. Let me know if you think this would be helpful as a standalone document somewhere (not sure where would make sense).

But, keep in mind: This is kind of a short-term stop-gap solution, as mentioned below. The "real" solution would be to circumvent the p2p message payload size limit by allowing a single message to be split up into multiple smaller ones (which requires a protocol update).

ffranr: Thanks for the detailed response. Very valuable.

> But, keep in mind: This is kind of a short-term stop-gap solution, as mentioned below. The "real" solution would be to circumvent the p2p message payload size limit by allowing a single message to be split up into multiple smaller ones (which requires a protocol update).

Seems like a complex solution for a temporary problem. Does it make more sense to limit naively (without care for worst case asset utxo count) for now and work on the split message side of things?

Does this really require a protocol update? Can't we just use our own message types that support splitting?

guggero: > Does it make more sense to limit naively

What values would you suggest? Sure, for now we could go with just 1 asset ID since we can't yet commit more than one anyway. But as soon as we add grouped asset support, this _will_ become an issue.

And remember, picking the wrong value means force close as worst case scenario, which is definitely not what we want.

> Can't we just use our own tap message types that support splitting?

Sure, since we do lnd-to-lnd only at the moment, we can do it without a protocol update. But that's still quite a big refactor on the `lnd` side. So not sure we can wait for that.

Roasbeef: So we _can_ do a protocol upgrade, but those changes would be more invasive. Right now we piggy back on top of all the existing LN related wire messages, so we can rely on `lnd` to do serialization and also retransmission for us. The state machine will also handle all messages, and pass over context to `tapd` as needed. 

If we add a new message, then we need to make sure that we process this in lock step with the existing LN channel state machine. Eg: if we add `asset_htlc_sigs`, we need to make sure that we validate that fully _before_ we revoke a state, or we can end up with an invalid commitment. 

I think this is fine as a stopgap, as the 483 limit is relatively high in practice. Most other implementations have a global limit of this value to like 50 or so. We can monitor to see if this is too low in practice, but it's also the case that right now with the way the protocol works, anyone can offer us a smaller limit. 

I think the path we can pursue after we land this is: revive the old message fragmentation logic in lnd. This keeps the 65 KB protocol limit, but lets an application level message span across multiple logical transport messages. With that, we can lift these limits, and things work as normal for the most part. We'd just need to make sure that re-introducing that logic doesn't break things for normal LN, in particular cross impl compat.

Roasbeef (approved):
LGTM 🔋

guggero:
Looks like we forgot to add the commitment type to the channel acceptor. Didn't notice this litd itest failure before, since it ran against the wrong branch.
So we'll need to wait for https://github.com/lightningnetwork/lnd/pull/9288 and then bump the `lnd` version.

ffranr (approved):

PR #1134, opened 2024-09-23
rfqmath: `FixedPoint.Mul` handles fixed-points with different scales

ffranr:
The previous implementation of FixedPoint.Mul required both fixed-point numbers to have the same scale, which led to silent errors when scales differed. This update ensures the function properly handles multiplication of fixed-points with different scales.

### DIFF
--- a/rfqmath/fixed_point.go
+++ b/rfqmath/fixed_point.go
@@ -72,16 +72,45 @@ func (f FixedPoint[T]) ToFloat64() float64 {
 	return float
 }
 
-// Mul returns a new FixedPoint that is the result of multiplying the existing
-// int by the passed one.
+// Mul multiplies the current FixedPoint value by another and returns the result
+// as a new FixedPoint value.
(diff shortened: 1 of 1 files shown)


dstadulis:
There would have been error if the points weren't the same

ffranr:
I believe this approach could help make `FixedPoint.Mul` more robust. Currently, the method on `main` requires the caller to manually scale the fixed-point argument before invoking `FixedPoint.Mul`. Without this scaling, the calculation is likely to be incorrect.

I'm closing this PR since it's not immediately necessary. I'll create an issue to outline the proposed changes, which we can de-prioritize relative to the current workload.

One weakness in this PR is the line:

```go
downScaleProduct := coefficientProduct.Div(divisor)
```

This could introduce precision loss, which is something we want to avoid. Additionally, the PR lacks unit tests for cases involving differing scales, which should be addressed.

PR #1136, opened 2024-09-24
Part 1: RFQ uses fixed-points

ffranr:
The high-level goal of this change is to represent asset-to-BTC conversion rates using the new fixed-point number type. This PR updates several "price" related fields that were previously typed as milli-satoshis, which were used as a workaround in the Proof of Concept (PoC).

It is part of the ongoing effort to resolve: https://github.com/lightninglabs/taproot-assets/issues/871.

The ultimate goal is to align tapd with these BLIP changes: https://github.com/Roasbeef/blips/pull/4

---

### Note to Reviewers

This is the first phase of the fixed-point refactor, targeting the `approved-rfq-fixedpoint` branch. This branch will accumulate all approved changes required to close https://github.com/lightninglabs/taproot-assets/issues/871 before being merged into `main`.

There are several `TODO(ffranr): Temp solution.` comments in this PR, which will be fully addressed in subsequent PRs before merging `approved-rfq-fixedpoint` into `main`.

Please review the commits in sequence. Note that some commit diffs may overlap slightly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1136)


dstadulis:
Current progress of feeding fixedpoints through the code, and building the commit structure through

ffranr:
I need to re-write this so that we use `BigIntFixedPoint` for all "internal" tapd fixed-points and then use `Uint64FixedPoint` when we export/import via RPC or wire messages.

rfq/oracle.go
```diff
@@ -52,26 +51,15 @@ func (o *OracleError) Error() string {
...
-	// Expiry is the price expiryDelay lifetime unix timestamp.
-	Expiry uint64
-
-	// Err is an optional error returned by the price oracle service.
-	Err *OracleError
-}
-
-// OracleBidResponse is a struct that holds the price oracle's suggested bid
-// price for an asset.
-type OracleBidResponse struct {
```
Roasbeef, line 70: So we've unified the bid vs ask structs to simplify? Would a type alias still be useful in certain contexts? So one for bid and the other for asks.

rfq/oracle.go
```diff
@@ -232,13 +220,28 @@ func (r *RpcPriceOracle) QueryAskPrice(ctx context.Context,
...
+		}
+
 		rateTickHint = &oraclerpc.RateTick{
-			Rate:            uint64(*bidPrice),
+			SubjectAssetRate: subjectAssetRate,
+
+			// For now, we only support BTC as the payment asset.
+			PaymentAssetRate: &oraclerpc.FixedPoint{
+				Coefficient: rfqmsg.MilliSatPerBtc.Uint64(),
+				Scale:       0,
```
Roasbeef, line 242: Assuming a scale factor again.

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -22,6 +22,37 @@ enum TransactionType {
...
+//
+// This type consists of two integer fields: a coefficient and a scale.
+// Using this format enables precise and consistent representation of fractional
+// numbers while avoiding floating-point data types, which are prone to
+// precision errors.
+//
+// The relationship between the fractional representation and its fixed-point
+// representation is expressed as:
+// ```
+// V = F_c / (10^F_s)
```
Roasbeef, line 35: 👍

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -32,13 +63,21 @@ enum TransactionType {
 // NOTE: That if the payment asset is BTC, the rate tick will be given as
 // milli-satoshi per asset unit.
 message RateTick {
-    // rate is the exchange rate between the subject asset and the payment
-    // asset.
-    uint64 rate = 1;
+    // subjectAssetRate is the subject asset to BTC conversion rate
```
Roasbeef, line 66: The comment above about the tick is now outdated.

Roasbeef: Also I think we should find a place in the proto comments to insert the formulas we have for conversion. Will serve as a guide for those implementing the new oracle version.

rfq/negotiator.go
```diff
@@ -161,12 +161,12 @@ func (n *Negotiator) HandleOutgoingBuyOrder(buyOrder BuyOrder) error {
 		// We calculate a proposed bid price for our peer's
 		// consideration. If a price oracle is not specified we will
 		// skip this step.
-		var bidPrice lnwire.MilliSatoshi
+		var assetRateBid *rfqmsg.BigIntFixedPoint
```
Roasbeef, line 164: `fn.Option` since specifying the suggested rate is optional?

rfqmsg/buy_request.go
```diff
@@ -40,14 +39,17 @@ type BuyRequest struct {
 	// requesting a quote.
 	AssetAmount uint64
 
-	// BidPrice is the peer's proposed bid price for the asset amount.
-	BidPrice lnwire.MilliSatoshi
+	// SuggestedAssetRate represents a proposed conversion rate between the
+	// subject asset and BTC. This rate is an initial suggestion intended to
+	// initiate the RFQ negotiation process and may differ from the final
+	// agreed rate.
+	SuggestedAssetRate *BigIntFixedPoint
```
Roasbeef, line 46: Same here re opt.

rfqmsg/buy_request.go
```diff
@@ -99,22 +101,25 @@ func NewBuyRequestMsgFromWire(wireMsg WireMessage,
...
-	// Extract the suggested rate tick if provided.
-	var bidPrice lnwire.MilliSatoshi
+	// Extract the suggested asset to BTC rate if provided.
+	//
+	// TODO(ffranr): Temp solution.
+	var suggestedAssetRate *BigIntFixedPoint
 	msgData.SuggestedRateTick.WhenSome(
 		func(rate tlv.RecordT[tlv.TlvType4, uint64]) {
-			bidPrice = lnwire.MilliSatoshi(rate.Val)
+			r := NewBigIntFixedPoint(rate.Val, 0)
```
Roasbeef, line 110: Assumes scale.

rpcserver.go
```diff
@@ -6591,14 +6592,18 @@ func marshalPeerAcceptedBuyQuotes(
 		[]*rfqrpc.PeerAcceptedBuyQuote, 0, len(quotes),
 	)
 	for scid, quote := range quotes {
+		rpcAskAssetRate := &rfqrpc.FixedPoint{
+			Coefficient: quote.AssetRate.Coefficient.ToUint64(),
```
Roasbeef, line 6596: With sufficient scale (this value is already multiplied by the scale), then this won't always fit into a uint64. 

We should be using the main constructor when ever possible to create these integers, using a base value (like in dollars/cents), then scaling up from there: https://github.com/lightninglabs/taproot-assets/blob/65539b6fd8e11657f96353a87714ed2d3e4399a6/rfqmath/fixed_point.go#L159-L169

We want ppl to think in the normal values, adding the scale to basically normalize things, not think in the largest possible number representation.

rpcserver.go
```diff
@@ -7136,8 +7141,19 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 
 	// Now that we have the accepted quote, we know the amount in Satoshi
 	// that we need to pay. We can now update the invoice with this amount.
-	mSatPerUnit := acceptedQuote.AskPrice
-	iReq.ValueMsat = int64(req.AssetAmount * mSatPerUnit)
+	askAssetRate := rfqmath.FixedPoint[rfqmath.GoInt[uint64]]{
+		Coefficient: rfqmath.NewGoInt(
```
Roasbeef, line 7146: I don't think we want to be using the go int variant anywhere. In this case, you may be silently losing precision.

rpcserver.go
```diff
@@ -7136,8 +7141,19 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+			acceptedQuote.AskAssetRate.Coefficient,
+		),
+		Scale: uint8(acceptedQuote.AskAssetRate.Scale),
+	}
+	assetAmount := rfqmath.FixedPoint[rfqmath.GoInt[uint64]]{
+		Coefficient: rfqmath.NewGoInt(req.AssetAmount),
+		Scale:       0,
```
Roasbeef, line 7153: Same here again re assuming scale.

taprpc/marshal.go
```diff
@@ -585,9 +585,11 @@ func MarshalAcceptedBuyQuoteEvent(
 		Id:          event.ID[:],
 		Scid:        uint64(event.ShortChannelId()),
 		AssetAmount: event.Request.AssetAmount,
-		// TODO(ffranr): Temp solution.
-		AskPrice: event.AssetRate.Coefficient.ToUint64(),
-		Expiry:   event.Expiry,
+		AskAssetRate: &rfqrpc.FixedPoint{
```
Roasbeef, line 589: Same here, the constructors should always be used. This is how we make sure that `Coefficient` is the right size internally. We want to scale up with the constructor, not before.

rfqmath/convert.go
```diff
@@ -62,6 +62,10 @@ func MilliSatoshiToUnits[N Int[N]](milliSat lnwire.MilliSatoshi,
 // compute the total amount of mSAT (X) as follows:
 //   - X = (U / Y) * M
 //   - where M is the number of mSAT in a BTC (100,000,000,000).
+//
+// TODO(ffranr): This function only works with BigInt as the underlying
```
Roasbeef, line 66: We only want to use the big int version. The other was added just to have another implementation, and also demonstrate where the unit64 will lose precision.

rfqmsg/sell_request.go
```diff
@@ -40,18 +39,19 @@ type SellRequest struct {
...
 
-	// AskPrice is the peer's proposed ask price for the asset amount. This
-	// is not the final price, but a suggested price that the requesting
-	// peer is willing to accept.
-	AskPrice lnwire.MilliSatoshi
+	// SuggestedAssetRate represents a proposed conversion rate between the
+	// subject asset and BTC. This rate is an initial suggestion intended to
+	// initiate the RFQ negotiation process and may differ from the final
+	// agreed rate.
+	SuggestedAssetRate *BigIntFixedPoint
```
Roasbeef, line 46: Same here re opt.

Roasbeef (changes requested):
Nice work! In particular, I found the commit structure very easy to follow. 

I think my main comments are related to the usage of the new `FixedPoint` integer, namely:
  * The main constructor should always be used. We take a value, then scale that, to arrive at the coefficient. At times I see a value (which may have already been scaled) used directly to create the struct, by passing the constructor. I think we should make the fields private to force everything to use the constructor. 
     * If this isn't done properly, then we'll end up manipulating a doubly or even triply scaled value. The eventual unit tests should also serve to bind this to make sure conversion is happening properly everywhere. 
  * We only want to use the big int version. The uint64 version was added mainly for completions sake.

rfqmsg/buy_request.go
GeorgeTsagk: part of temp solution

It will be fully threaded through in follow-up PR as noted in the [PR desc](https://github.com/lightninglabs/taproot-assets/pull/1136#issue-2546061589)

GeorgeTsagk (approved):
Lgtm

rfqmath/convert.go
ffranr: With this TODO, in a later PR, I hope to make `UnitsToMilliSatoshi` specific to `BigInt` fixed-points by removing the generic type parameter.

rfqmsg/messages.go
```diff
@@ -194,3 +199,34 @@ type QuoteResponse interface {
 	// String returns a human-readable string representation of the message.
 	String() string
 }
+
+// Uint64FixedPoint is a fixed point record that can be used to encode/decode a
+// fixed point number with a `uint64` coefficient to/from a TLV stream.
+type Uint64FixedPoint rfqmath.FixedPoint[rfqmath.GoInt[uint64]]
```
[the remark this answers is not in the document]
ffranr: I've removed `Uint64FixedPoint` in the latest commits. We only need `BigIntFixedPoint` in this PR.

rfqmsg/messages.go
```diff
@@ -194,3 +199,34 @@ type QuoteResponse interface {
...
+
+// NewUint64FixedPoint creates a new fixed point record given a fixed-point
+// coefficient and scale.
+func NewUint64FixedPoint(coefficient uint64, scale uint8) Uint64FixedPoint {
```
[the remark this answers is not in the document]
ffranr: I've removed `Uint64FixedPoint` in the latest commits. We only need `BigIntFixedPoint` in this PR.

rfqmsg/messages_test.go
```diff
@@ -0,0 +1,40 @@
...
+// TestUint64FixedPoint tests encoding and decoding of the Uint64FixedPoint
+// struct.
+func TestUint64FixedPoint(t *testing.T) {
+	type testStruct struct {
+		ExchangeRate tlv.RecordT[tlv.TlvType1, Uint64FixedPoint]
+	}
+
+	foo := &testStruct{
+		ExchangeRate: tlv.NewRecordT[tlv.TlvType1](
+			NewUint64FixedPoint(123456789, 7),
```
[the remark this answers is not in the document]
ffranr: I've removed `Uint64FixedPoint` in the latest commits. We only need `BigIntFixedPoint` in this PR.

rfq/negotiator.go
```diff
@@ -178,6 +178,13 @@ func (n *Negotiator) HandleOutgoingBuyOrder(buyOrder BuyOrder) error {
 					"price oracle for outgoing buy "+
 					"request: %v", err)
 			}
+
+			// TODO(ffranr): This is a temporary solution which will
+			//  be re-written once RFQ quote request messages are
+			//  updated to include a suggested asset rate.
```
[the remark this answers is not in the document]
ffranr: Yes

docs/examples/basic-price-oracle/go.mod
```diff
@@ -1,24 +0,0 @@
-module basic-price-oracle
-
-go 1.22
-
-toolchain go1.22.3
-
-replace github.com/lightninglabs/taproot-assets => ../../../
```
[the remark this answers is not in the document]
ffranr: My mistake, I will add back in. Not sure why the CI is green...

rfq/negotiator.go
```diff
@@ -322,10 +315,17 @@ func (n *Negotiator) HandleIncomingBuyRequest(
 	go func() {
 		defer n.Wg.Done()
 
+		// TODO(ffranr): This is a temporary solution which will be
+		//  re-written once RFQ quote request messages are updated to
+		//  include a suggested asset rate.
+		suggestedAssetRate := rfqmsg.NewUint64FixedPoint(
+			uint64(request.BidPrice), 4,
```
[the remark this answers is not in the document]
ffranr: The assumed scale factor is now 0 (in the latest commits). But this will soon be replaced with carrying the fixed-point type through.

taprpc/priceoraclerpc/price_oracle.proto
ffranr: A lot of `price_oracle.proto` is out of date. The rate tick terminology for example. I will be sure to change all that before I consider this work complete. But for this PR I hope to just focus on getting the fixed-point type into the right fields.

I'll be sure to add a lot to the proto comments :+1:

rfqmsg/buy_request.go
ffranr: Yes. Everything marked with `// TODO(ffranr): Temp solution.` in the latest commits for this PR.

rfq/negotiator.go
```diff
@@ -123,28 +123,28 @@ func (n *Negotiator) queryBidFromPriceOracle(peer route.Vertex,
...
+		return nil, 0, fmt.Errorf("failed to query price oracle for "+
 			"bid price: %s", oracleResponse.Err)
 	}
 
 	// By this point, the price oracle did not return an error or a bid
 	// price. We will therefore return an error.
-	if oracleResponse.BidPrice == nil {
-		return 0, 0, fmt.Errorf("price oracle did not specify a bid " +
-			"price")
+	if oracleResponse.AssetRate.Coefficient.ToUint64() == 0 {
```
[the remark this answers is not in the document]
ffranr: > I think it can just compare the coefficient directly here?

Latest commit compare `BigInt` with `0`. So we need the `ToUint64`.

> Or if the value is nil if there's no response, then we can continue to rely on that. If it's nil, won't we trigger a nil pointer here if we always deref?

If `oracleResponse == nil` and `err != nil` then we have an internal coding error in the implementation of `n.cfg.PriceOracle.QueryBidPrice`. That return situation goes against a convention I thought we had. I'm not sure if I'm answering your question here.

rfq/oracle.go
ffranr: > So we've unified the bid vs ask structs to simplify?

Yes. The caller knows what they'll get back because the function that they call is specific to bid/ask.

> Would a type alias still be useful in certain contexts? So one for bid and the other for asks.

I don't think there's much use for an alias right now. Usually we pass the returned value straight into a request or accept with a field name that indicates bid/ask.

rpcserver.go
ffranr: I think we should add something to minimise the coefficient and scale of a fixed-point before trying to export as `uint64`:

```golang
// countTrailingZeros returns the number of trailing zeros in the given integer.
func countTrailingZeros[T Int[T]](v T) uint8 {
	var (
		val = v

		// count will be the number of trailing zeros in the given
		// integer.
		count = uint8(0)
	)

	// We'll use these constants to avoid creating new instances
	// unnecessarily.
	var (
		zero = NewInt[T]().FromUint64(0)
		ten  = NewInt[T]().FromUint64(10)
	)

	// Count the number of trailing zeros in the given integer `val`.
	// The counting process continues as long as `val` is greater than zero
	// and dividing `val` by 10 does not result in a loss of precision.
	// This means that dividing by 10 successfully removes a trailing zero
	// from the number.
	for val.Gt(zero) && val.Div(ten).Mul(ten).Equals(val) {
		val = val.Div(ten)
		count++
	}

	return count
}

// MinParams returns a new FixedPoint with the smallest possible scale
// and coefficient, preserving precision. Useful for converting a fixed-point
// coefficient to a `uint64` representation.
func (f FixedPoint[T]) MinParams() FixedPoint[T] {
	trailingZeros := countTrailingZeros(f.Coefficient)

	// Determine the new scale based on the number of trailing zeros in
	// the coefficient.
	var newScale uint8
	if trailingZeros > f.Scale {
		// If the number of trailing zeros is greater than the current
		// scale, we can effectively use a minimal scale of 0.
		newScale = 0
	} else {
		// Otherwise we can only scale down to a new scale that is the
		// difference between the current scale and the number of
		// trailing zeros. Scaling down further will result in a loss
		// of precision in the coefficient.
		newScale = f.Scale - trailingZeros
	}

	// Return a new FixedPoint scaled to the smallest possible scale.
	return f.ScaleTo(newScale)
}
```

ffranr: In the latest commits, when marshalling into RPC, I've changed `quote.AssetRate.Coefficient.ToUint64()` into `ToUint64Safe`.

rpcserver.go
ffranr: Yes, good catch. I've since tried to use `BigInt` fixed-point everywhere. This got through by mistake.

rfqmsg/sell_request.go
ffranr: Fixed in latest commits.

rpcserver.go
ffranr: Don't we have to assume a scale of 0 when converting an asset amount to a fixed-point?

taprpc/marshal.go
ffranr: Would you like me to create a marshal constructor for the fixed-point RPC type?

rfqmsg/messages.go
```diff
@@ -75,9 +75,13 @@ const (
...
 var (
-	// ErrUnknownMessageType is an error that is returned when an unknown
-	// message type is encountered.
+	// ErrUnknownMessageType is an error returned when an unknown message
+	// type is encountered.
 	ErrUnknownMessageType = errors.New("unknown message type")
+
+	// MilliSatPerBtc is the number of milli-satoshis in one bitcoin, equal
+	// to 100 billion (10^11).
+	MilliSatPerBtc = big.NewInt(100_000_000_000)
```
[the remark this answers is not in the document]
ffranr: this is now a big int fixed-point

rfq/oracle.go
ffranr: `rfqmsg.MilliSatPerBtc` is of type big int fixed-point in the latest commits. It's now defined as:
```
MilliSatPerBtc = rfqmath.FixedPointFromUint64[rfqmath.BigInt](100, 9)
```
does that address your concern?

Roasbeef:
Ok, discussed offline, and for now re format, we'll:
  * Figure out a proper format for the byte serialization of the fixed point instance. 
  * As a stop gap, we'll:
      * Write a property based test for the new intermediate value, to ensure that we never lose scale. 
      * On the wire, just use the `big.Int` serialization. 
      * For the RPC interface, write the scale and coefficient values a strings.

Roasbeef (changes requested):
> Reviewed 30 of 30 files at r1, 1 of 29 files at r4.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1136)* status: 2 of 30 files reviewed, 27 unresolved discussions (waiting on @ffranr)

___
*[`rpcserver.go` line 7174 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/1136#-O8vU6Hd30ucVNrlfPZC:-O8vU6He7pZn3jt1yBrB:b4y9sb0) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/c5a94c38ebb191e3d823180b60de68bbe811cefa/rpcserver.go#L7174)):*
> ```Go
> 
> 	// Convert the asset amount into a fixed-point.
> 	assetAmount := rfqmsg.NewBigIntFixedPoint(req.AssetAmount, 0)
> ```

This should be using the decimal display of the given asset. 
___
*[`rfq/order.go` line 128 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/1136#-O8vUbloDqNoulDo_iFc:-O8vUbloDqNoulDo_iFd:b-qssa7u) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/c5a94c38ebb191e3d823180b60de68bbe811cefa/rfq/order.go#L128)):*
> ```Go
> 	// Check that the HTLC amount is not greater than the negotiated maximum
> 	// amount.
> 	maxAssetAmount := rfqmsg.NewBigIntFixedPoint(c.MaxAssetAmount, 0)
> ```

Same here, should be using the proper scale value (from the decimal display).

ffranr:
> Ok, discussed offline, and for now re format, we'll:
> 
>     * Figure out a proper format for the byte serialization of the fixed point instance.
> 
>     * As a stop gap, we'll:
>       
>       * Write a property based test for the new intermediate value, to ensure that we never lose scale.
>       * On the wire, just use the `big.Int` serialization.
>       * For the RPC interface, write the scale and coefficient values a strings.


>       * Write a property based test for the new intermediate value, to ensure that we never lose scale.
>       * On the wire, just use the `big.Int` serialization.
This two points are relevant to the part 2 PR: https://github.com/lightninglabs/taproot-assets/pull/1141
This PR doesn't include any any RFQ wire messages changes.


> * For the RPC interface, write the scale and coefficient values a strings.

This change is included in this PR.

ffranr:
> > Reviewed 30 of 30 files at r1, 1 of 29 files at r4.
> > _[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1136)_ status: 2 of 30 files reviewed, 27 unresolved discussions (waiting on @ffranr)
> 
> _[`rpcserver.go` line 7174 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/1136#-O8vU6Hd30ucVNrlfPZC:-O8vU6He7pZn3jt1yBrB:b4y9sb0) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/c5a94c38ebb191e3d823180b60de68bbe811cefa/rpcserver.go#L7174)):_
> 
> > ```go
> > 	// Convert the asset amount into a fixed-point.
> > 	assetAmount := rfqmsg.NewBigIntFixedPoint(req.AssetAmount, 0)
> > ```
> 
> This should be using the decimal display of the given asset.
> 
> _[`rfq/order.go` line 128 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/1136#-O8vUbloDqNoulDo_iFc:-O8vUbloDqNoulDo_iFd:b-qssa7u) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/c5a94c38ebb191e3d823180b60de68bbe811cefa/rfq/order.go#L128)):_
> 
> > ```go
> > 	// Check that the HTLC amount is not greater than the negotiated maximum
> > 	// amount.
> > 	maxAssetAmount := rfqmsg.NewBigIntFixedPoint(c.MaxAssetAmount, 0)
> > ```
> 
> Same here, should be using the proper scale value (from the decimal display).

@Roasbeef

In both of these comments I think you're asking me to construct the fixed-points using the asset's decimal display (instead of just using a scale of 0 and setting the coefficient directly).

I don't see why we should care about the decimal display value here. The value of the fixed-points remains the same, and the calculation is unaffected. Looking up the decimal display introduces a potential for error.

taprpc/marshal.go
guggero: Yeah, maybe just a `rfqrpc.MarshalFixedPoint` that does exactly this. Just to make sure we don't accidentally forget the scale or coefficient (however unlikely that is).

rpcserver.go
```diff
@@ -7136,8 +7152,22 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+	askAssetRate, err := rfqrpc.UnmarshalFixedPoint(
+		acceptedQuote.AskAssetRate,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshalling ask asset rate: %w",
+			err)
+	}
+
+	// Convert the asset amount into a fixed-point.
+	assetAmount := rfqmsg.NewBigIntFixedPoint(req.AssetAmount, 0)
```
guggero, line 7167: I thought about this and what @Roasbeef said a lot, and I think I agree with him that we can't use 0 here.
Because we are going to calculate a price for the asset, it is important that we know the inherent representation of the asset. So basically if we use 0, we would imply one asset unit being 1 dollar, even though the asset might have a decimal display of 3 or 6.

So we do need to query the decimal display value here and use it. That also makes sure we have synced the issuance proof for that asset when attempting to create an invoice.

rfq/order.go
```diff
@@ -128,7 +125,12 @@ func (c *AssetSalePolicy) CheckHtlcCompliance(
 
 	// Check that the HTLC amount is not greater than the negotiated maximum
 	// amount.
-	maxOutboundAmount := lnwire.MilliSatoshi(c.MaxAssetAmount) * c.AskPrice
+	maxAssetAmount := rfqmsg.NewBigIntFixedPoint(c.MaxAssetAmount, 0)
```
guggero, line 128: See previous comment, we need to fetch the decimal display unit here and use it.

rfq/order.go
```diff
@@ -175,8 +177,15 @@ func (c *AssetSalePolicy) GenerateInterceptorResponse(
...
 	}
 
-	outgoingAssetAmount := uint64(htlc.AmountOutMsat / c.AskPrice)
-	htlcBalance := rfqmsg.NewAssetBalance(*c.assetID, outgoingAssetAmount)
+	// Compute the outgoing asset amount given the msat outgoing amount and
+	// the asset to BTC rate.
+	outgoingAssetAmount := rfqmath.MilliSatoshiToUnits(
+		htlc.AmountOutMsat, c.AskAssetRate,
+	)
+	amt := outgoingAssetAmount.ScaleTo(0).ToUint64()
```
guggero, line 185: Hmm, from looking at the Godoc of `ToUint64()` the `ScaleTo(0)` _shouldn't_ be required, as it mentions "scaled down from the existing scale". But it looks like the implementation doesn't match that. Perhaps something to fix later on.

guggero (commented):
Nice work! I think this is pretty close, with the exception of two uses of 0 as the scale that aren't marked with `// TODO: temp solution`.

Perhaps those instances could be partially recreated in a unit test to show they do the right thing. Properly executing the specific logic within a unit or integration test is probably pretty hard. So just a demo of the conversion part in a unit test should give us some confidence with some real-world number examples that are easy to follow as mere humans.

rpcserver.go
ffranr: This is how I see it:

We're calculating msat value from asset units and an asset unit to BTC rate. The decimal display is not relevant to the calculation. I've added this unit test to show that scale 0 works here: https://github.com/lightninglabs/taproot-assets/pull/1136/commits/875c0e2c1c94b8cb912a42ee91fcfdd6e1db1e43

Why is decimal display useful here for converting asset units to msats?

ffranr: Are we saying that `req.AssetAmount` is not in asset units but in real world currency units here? I think not because it's a `uint64`, right?

rfq/oracle.go
```diff
@@ -374,56 +402,59 @@ var _ PriceOracle = (*RpcPriceOracle)(nil)
...
+			assetRateCoefficient, 0,
+		),
 	}
 }
 
 // NewMockPriceOracleSatPerAsset creates a new mock price oracle with a
 // specified satoshis per asset rate.
 func NewMockPriceOracleSatPerAsset(expiryDelay uint64,
-	satPerAsset btcutil.Amount) *MockPriceOracle {
+	assetRateCoefficient uint64) *MockPriceOracle {
```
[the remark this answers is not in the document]
ffranr: I've added TODOs to remove this and related config param.

ffranr:
I think our confusion might come down to how we define fields. Consider this part of the code which is contentious:
```
	// Check that the HTLC amount is not greater than the negotiated maximum
	// amount.
	maxAssetAmount := rfqmath.NewBigIntFixedPoint(c.MaxAssetAmount, 0)

	maxOutboundAmount := rfqmath.UnitsToMilliSatoshi(
		maxAssetAmount, c.AskAssetRate,
	)
```

From my POV:
* `maxAssetAmount` is in asset units uint64 (with not scaled down for decimal display in the UI). This isn't a cent float value. It is a fixed-point with coefficient `c.MaxAssetAmount` and scale `0`.

* `c.AskAssetRate` this is the asset units to BTC rate. It is not the same as the cents to BTC rate. It is the number of asset units per BTC. Before displaying this in a UI it would need to be scaled with decimal display.

* `maxOutboundAmount` is in msats.

rpcserver.go
Roasbeef: Ultimately it _is_ in asset units, just unscaled. I think that's where they divergence is in our thought models here, and also where I want to make sure we're very careful (re use of constructors, etc). My fear was that if this _was_ already a scaled value (output from some other operation that didn't scale down), then we'd compound an error by not carrying along what the true scale should be. 

So far now, I think this seems to be fine, but ultimately the further unit and itests will be the true judge. 

I think my scenario above can still happen, if we ever write a scaled value to wire/rpc w/o the scale itself. Will think about how we can add some sanity checks or assertions elsewhere to make sure we catch instances like this.

docs/examples/basic-price-oracle/go.mod
```diff
@@ -4,21 +4,185 @@ go 1.22
 
 toolchain go1.22.3
 
-replace github.com/lightninglabs/taproot-assets => ../../../
+replace (
+	github.com/lightninglabs/taproot-assets => ../../../
```
Roasbeef, line 8: I think this should be removed before merge?

Roasbeef (approved):
LGTM 🐒

PR #1137, opened 2024-09-25
rfqmath: clarify scale requirement in FixedPoint Mul and Div docs

ffranr:
This commit updates the documentation for `FixedPoint.Mul` and `FixedPoint.Div` to specify that these methods cannot be used with arguments having different scale values from the target fixed-point. This clarification helps prevent misuse of these methods.


dstadulis (approved):
lgtm, simple documentation clarification

GeorgeTsagk (approved):

PR #1139, opened 2024-09-30
taprpc: move `ImportProof` from `tapdevrpc` to `taprpc`

bhandras:
This commit turns the`ImportProof` API to be publicly available as it is required for some use cases.


jharveyb:
Relevant existing discourse:

https://github.com/lightninglabs/taproot-assets/pull/804#discussion_r1535160155

Is the use case manually importing a proof for an asset that the importing node actually owns?

In that case, maybe we can improve the existing `InsertProof` call to 'take ownership' of the inserted assets, if we know that we own the relevant internal and script keys.

Not sure if that covers your use or you're looking at something more involved? Either way, sidestepping the local universe will probably cause issues later.

dstadulis:
Core user need: when importing a proof and recognize and if the node owns the asset that's being imported
Custom channels side steps the introspection need

bhandras:
> Relevant existing discourse:
> 
> [#804 (comment)](https://github.com/lightninglabs/taproot-assets/pull/804#discussion_r1535160155)
> 
> Is the use case manually importing a proof for an asset that the importing node actually owns?
> 
> In that case, maybe we can improve the existing `InsertProof` call to 'take ownership' of the inserted assets, if we know that we own the relevant internal and script keys.
> 
> Not sure if that covers your use or you're looking at something more involved? Either way, sidestepping the local universe will probably cause issues later.

IIUC `InsertProof` wouldn't work in our case as what we need is a receiver who doesn't own the internal and script keys. The actual use case is a swap HTLC sweep. wdyt?

bhandras:
Closing this as we managed to emulate the same functionality by iterating on the whole proof with `InsertProof`  plus setting the anchor derivation upon sweep. Thanks again to @guggero for all the insights and debugging help!

PR #1140, opened 2024-10-03
examples: add TLS support to basic price oracle service

ffranr:
This commit adds basic TLS support to the example price oracle service. When connecting to the RPC price oracle, Tapd now assumes TLS. TLS certificate authentication is skipped by Tapd.

Simplify integration of the example price oracle with Tapd. (Hopefully helpful for @HannahMR )

### DIFF
--- a/docs/examples/basic-price-oracle/main.go
+++ b/docs/examples/basic-price-oracle/main.go
@@ -6,6 +6,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"crypto/ecdsa"
 	"crypto/elliptic"
@@ -13,6 +14,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"encoding/hex"
 	"encoding/pem"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


docs/examples/basic-price-oracle/main.go
```diff
@@ -174,12 +183,72 @@ func startService(grpcServer *grpc.Server) error {
...
-	backendService := grpc.NewServer(serverOpts...)
+
+	// Create the gRPC server with TLS
+	transportCredentials := credentials.NewTLS(&tls.Config{
+		Certificates: []tls.Certificate{tlsCert},
+	})
+
+	backendService := grpc.NewServer(grpc.Creds(transportCredentials))
 	_ = startService(backendService)
 	backendService.Stop()
```
gijswijs, line 253: Shouldn't we handle a returned error?
Also, why not use `backendService.GracefulStop()`?

gijswijs (approved):
LGTM! :tada:
I can still see some magic numbers in there, but that's expected for essentially a mock. But I just wanted to flag it.

Also, this server now _only_ works with TLS. That's what we want, right?

Made a nit comment, but nothing that would block merging.

ffranr:
> Also, this server now only works with TLS. That's what we want, right?

@gijswijs yes, I think it will do for now. And tapd expects that right now.

gijswijs (approved):

HannahMR:
This does solve the communication error between litd and the oracle in my testing.

ffranr:
> Also, this server now only works with TLS. That's what we want, right?

That's fine IMO. We just want to make sure that it can work with tapd out of the box.

docs/examples/basic-price-oracle/main.go
```diff
@@ -174,12 +183,72 @@ func startService(grpcServer *grpc.Server) error {
...
+
+	keyUsage := x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature
+	extKeyUsage := []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth}
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		Subject: pkix.Name{
+			Organization: []string{"basic-price-oracle"},
+		},
+		NotBefore: time.Now(),
+		NotAfter:  time.Now().Add(24 * time.Hour), // Valid for 1 day
```
dstadulis, line 244: What's the rational for expiring the cert in 24 hours?

docs/examples/basic-price-oracle/main.go
```diff
@@ -174,12 +226,78 @@ func startService(grpcServer *grpc.Server) error {
 	return grpcServer.Serve(grpcListener)
 }
 
+// Generate a self-signed TLS certificate and private key.
+func generateSelfSignedCert() (tls.Certificate, error) {
+	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
dstadulis, line 231: Would other curves be more desired? p256 vs https://pkg.go.dev/crypto/elliptic#P521

ffranr:
I've added logging to this and the format is sub-optimal. So I will improve before merge.

docs/examples/basic-price-oracle/main.go
ffranr: Just a reasonable number that won't get in the way of running the basic example.

docs/examples/basic-price-oracle/main.go
ffranr: This is just to get TLS working with the example service. Its just example code and not security critical.

Issue #1008, opened 2024-07-10
tapchannel/[feature]: update `AuxInvoiceManager` to enforce strict forwarding for invoices


Roasbeef:
**Is your feature request related to a problem? Please describe.**

Today if a user has one private asset channel, and a public Bitcoin channel, then due to fees or jitter otherwise, a sender may actually pay over the _BTC_ channel. From the receivers side, the invoice _has_ been paid, but potentially not in the desired currency. 

**Describe the solution you'd like**

Today we'll based the custom invoice hook based on if an HTLC has custom data or not: https://github.com/lightninglabs/taproot-assets/blob/fca9d11f027cdb5b81a36c4a8999c6b010ac0c44/tapchannel/aux_invoice_manager.go#L121-L124. 

We should also do the opposite check: if we get an HTLC over a channel, for an invoice with custom data, but the HTLC has no custom data, then we reject. 

**Describe alternatives you've considered**

Make the behavior above optional, all a lax receive mode (any currency works). 

>

dstadulis:
>[should] reject the invoice if it doesn't come across the asset chan we expect
>logic would be in the invoice acceptor

Handles the case where an external node attempts to route payment to node ID rather than over a discrete set of SCIDs

guggero:
Probably also need to do something similar on the `SendPayment` side.
Example for paying a 500 sat invoice with assets, but the pathfinding choosing arbitrary (non-TAP channel) first hops:

```
d5fac4f7358e:/# litcli --macaroonpath /root/.lnd/data/chain/bitcoin/mainnet/admin.macaroon ln payinvoice --pay_req lnbc5u1pnvgudppp52g3pn35s5ewzu2guahyv44mparkxg2y7r3egqpflumc6yt5xdd8sdqqcqzzsxqyz5vqsp5w64str3ac0s4z6h4fhzry5gcj483mvlz8n5l6uajcfsl2uvscvys9qxpqysgqwmw58axr96xln4srf4wgp69aqkrfgfl7cv79h83cs56s6zke07yhmr672xeyj8e6sz4k0w45dhcw559mtu5k3vwkv35dw6sk0qqlkugpjwcrpf --asset_id d75dae162be7cd0a2df76831ee1cce4eaeb74a32fabb0d46be7a0ad70fd6ad27
Payment hash: 522219c690a65c2e291cedc8cad761e8ec64289e1c7280053fe6f1a22e866b4f
Description:
Amount (in satoshis): 500
[...]

PR #1144, opened 2024-10-08
tapchannel: enforce strict forwarding for asset invoices

GeorgeTsagk:
This PR adds a few extra steps to the `AuxInvoiceManager` before it settles an invoice. Previously we wouldn't check that we received an asset HTLC for an asset invoice, leading to some [undesired behavior](https://github.com/lightninglabs/taproot-assets/issues/1008#issue-2401933487).

With these changes we enforce that if the invoice corresponds to an asset invoice then we only settle it if we received assets.

Todo:

- [x] Tests

Closes https://github.com/lightninglabs/taproot-assets/issues/1008


dstadulis:
Before assigning reviewers, tests will need to be added.

guggero (commented):
Nice approach! And great to have a unit test. Though I wonder what would actually happen on the sender and receiver side with this change? My assumption is that things would just time out (rather than rejecting the HTLC).
So I think we should also have an integration test for this.

tapchannel/aux_invoice_manager.go
```diff
@@ -120,6 +123,14 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(_ context.Context,
 
 	// No custom record on the HTLC, so we have nothing to do.
 	if len(req.WireCustomRecords) == 0 {
+		// If there's no wire custom records and the invoice is an asset
+		// invoice do not settle the invoice.
+		//
+		// TODO(george): Strict-forwarding could be configurable?
+		if s.isAssetInvoice(req.Invoice) {
+			resp.AmtPaid = 0
```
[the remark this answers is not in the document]
GeorgeTsagk: We're intercepting the invoice at the htlc level, which means that for mpp payments we'll intercept N times for N shards. Within the context of intercepting a single htlc we can only signal what the effective value of that single htlc is (i.e we can only ignore the "value contribution")

In the payment context we'll see a timeout, as we're still waiting for other htlcs that may sum up to the expected value

GeorgeTsagk:
An itest that covers this behavior has been added in a [litd PR](https://github.com/lightninglabs/lightning-terminal/pull/875)

GeorgeTsagk:
Will enhance the `HtlcModify` interface to explicitly allow cancelling of HTLCs

tapchannel/aux_invoice_manager.go
```diff
@@ -27,6 +30,20 @@ type InvoiceHtlcModifier interface {
 		handler lndclient.InvoiceHtlcModifyHandler) error
 }
 
+// RfqManager is an interface that abstracts the functionalities of the rfq
+// manager that are needed by AuxInvoiceManager.
+type RfqManager interface {
```
jharveyb, line 35: Add compile-time check that `rfq.Manager` satisfies this new interface?

tapchannel/aux_invoice_manager.go
```diff
@@ -227,6 +252,59 @@ func (s *AuxInvoiceManager) priceFromQuote(rfqID rfqmsg.ID) (
...
+// the scid matches an existing quote.
+func (s *AuxInvoiceManager) isAssetInvoice(invoice *lnrpc.Invoice) bool {
+	hints := invoice.RouteHints
+
+	if len(hints) != 1 {
+		return false
+	}
+
+	hint := hints[0]
+	if len(hint.HopHints) != 1 {
```
jharveyb, line 289: Should this be a minimum of 1, instead of exactly 1? Seems like this would cause issues if we wanted to use hop hints for other reasons with this payment.

Also, AFAICT we only need to inspect one hint (the 'last' hint) for this use case.

tapchannel/aux_invoice_manager.go
```diff
@@ -227,6 +252,59 @@ func (s *AuxInvoiceManager) priceFromQuote(rfqID rfqmsg.ID) (
...
+	if err != nil {
+		log.Debugf("scid %v does not correspond to a valid RFQ quote",
+			scid)
+
+		return false
+	}
+
+	// We also want the RFQ peer and the node ID of the hop hint to match.
+	nodeId := hop.NodeId
+	return strings.Compare(peer.String(), nodeId) == 0
```
jharveyb, line 305: nit: p sure this can just use `==`

jharveyb:
A failing case (with the PR as-is, I think this seed is invalid if we change any of the generators):

`go test -run TestAuxInvoiceManagerProperty -rapid.seed 11951275822997898772`

Which fails here:

https://github.com/GeorgeTsagk/taproot-assets/blob/a1e4157e0b0905bbb76595485cee268402428fd2/tapchannel/aux_invoice_manager_test.go#L161

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,571 @@
...
+	return rfqID
+}
+
+// genInvoice generates an invoice that may have a random amount, and may have
+// routing hints.
+func genInvoice(t *rapid.T) *lnrpc.Invoice {
+	res := &lnrpc.Invoice{}
+
+	// Do a random draw with a 1/8 chance of introducing an empty invoice.
+	nullInv := rapid.Uint16().Draw(t, "invoice_exists")
```
jharveyb, line 340: I think the places where you to pick a random, probably invalid, value with some frequency, we can just use `rapid.Bool()` instead.

Also, if you separate the Generator for `lnrpc.Invoice` into another function, ideally that matches the `Generator` function signature:

https://pkg.go.dev/pgregory.net/rapid#Custom

Then a generator for `*lnrpc.Invoice` using https://pkg.go.dev/pgregory.net/rapid#Ptr is straightforward:

```go
func invoiceGen(t *rapid.T) lnrpc.Invoice {
/// implementation
}

func requestGen(t *rapid.T) (.....) {
/// implementation 

request.Invoice = rapid.Ptr(invoiceGen, true).Draw(t, "request invoice")

/// implementation
}
```

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,571 @@
...
+	if nullInv%8 == 0 {
+		return res
+	}
+
+	// Generate a random invoice value.
+	valueMsat := rapid.Int64Range(
```
jharveyb, line 346: I would recommend making the ranges for these numerical values like invoice value, units, and price much smaller; at least to start. IIUC this allows the test framework to cover more edge cases more quickly (and combinations of these variables that may be edge cases).

IIUC for these tests we really care about the relationship between these three values and not the absolute magnitudes.

Would also be good to have comments explaining why each range is the size it is, and why `0` may or may not be a meaningful value, etc.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,571 @@
...
+
+	res.ValueMsat = valueMsat
+
+	// Do a random draw with a 1/8 chance to introduce empty routing hints.
+	nullHints := rapid.Uint16().Draw(t, "invoice_hints_exist")
+	if nullHints%8 == 0 {
+		return res
+	}
+
+	res.RouteHints = testRouteHints()
```
jharveyb, line 358: I think we can also have a generator for route hints of a random length and with random content. Right now I don't think the tests cover route hints that exist but are not matches to the expected route hint. Nor does it cover cases of more than 1 hint.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,571 @@
...
+
+	return res
+}
+
+// genCustomRecords generates custom records that have a random amount of random
+// asset units, and may have an scid as routing hint.
+func genCustomRecords(t *rapid.T,
+	rfqID rfqmsg.ID) (lnwire.CustomRecords, uint64) {
+
+	assetUnits := rapid.Uint64Range(10_000, 100_000).Draw(t, "asset_units")
```
jharveyb, line 368: Same note re: smaller range.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,571 @@
...
+	return customRecords, assetUnits
+}
+
+// genHtlc generates an istance of rfqmsg.Htlc with the provided asset amounts
+// and rfqID.
+func genHtlc(t *rapid.T, balance []*rfqmsg.AssetBalance,
+	rfqID rfqmsg.ID) *rfqmsg.Htlc {
+
+	// Introduce a 1/5 chance of no rfqID in this htlc.
+	noRfqID := rapid.Uint16().Draw(t, "has_rfqid_probability")
```
jharveyb, line 390: Same notes re: replacing `Uint16()` and mod with `Bool()`.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,571 @@
...
+	rfqMap := rfq.BuyAcceptMap{}
+
+	len := rapid.IntRange(0, 5).Draw(t, "requestsLen")
+	requests := make([]lndclient.InvoiceHtlcModifyRequest, len)
+
+	for i := range requests {
+		req, numAssets, scid := genRequest(t)
+		requests[i] = req
+
+		genBuyQuotes(
```
jharveyb, line 443: 👍🏽

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,571 @@
...
+func genBuyQuotes(t *rapid.T, rfqMap rfq.BuyAcceptMap, units, amtMsat uint64,
+	scid rfqmsg.ID) {
+
+	var peer route.Vertex
+	askPrice := lnwire.MilliSatoshi(0)
+
+	// Introduce a 1/8 chance that the quote's peerID is not correct.
+	noPeerMatch := rapid.Uint16().Draw(t, "nodeID_mismatch_probability")
+	if noPeerMatch%8 == 0 {
+		peer = route.Vertex{3, 1, 4}
```
jharveyb, line 463: Would be good to further randomize the negative vertex values we're using as well. I think just generating 2 random bytes and converting that to a vertex would be sufficient.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,571 @@
...
+	} else {
+		peer = testNodeID
+	}
+
+	// Introduce a 1/5 chance that the askPrice of this asset will result in
+	// a random total asset value.
+	noValueMatch := rapid.Uint16().Draw(t, "no_asset_value_match")
+	if noValueMatch%5 == 0 {
+		askPrice = lnwire.MilliSatoshi(
+			rapid.Uint64Range(
```
jharveyb, line 473: Same note re: smaller range.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,571 @@
...
+	return h.Sum(nil)
+}
+
+func assetID(i byte) asset.ID {
+	return asset.ID(newHash([]byte{i}))
+}
+
+func dummyRfqID(value int) rfqmsg.ID {
+	var rfqID rfqmsg.ID
+	for i := 0; i < len(rfqID); i++ {
```
jharveyb, line 543: super-nit: could use `bytes.Repeat()` here.

jharveyb (commented):
Solid start with the property tests! Definitely headed in the right direction.

The actual implementation changes look reasonable, left some notes on test improvements.

tapchannel/aux_invoice_manager.go
```diff
@@ -121,6 +124,14 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(_ context.Context,
 
 	// No custom record on the HTLC, so we have nothing to do.
 	if len(req.WireCustomRecords) == 0 {
+		// If there's no wire custom records and the invoice is an asset
+		// invoice do not settle the invoice.
+		//
+		// TODO(george): Strict-forwarding could be configurable?
```
guggero, line 144: Not sure we need this TODO here. IMO if you want assets you create an asset invoice. If not, you create a normal one.

tapchannel/aux_invoice_manager.go
```diff
@@ -121,6 +124,14 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(_ context.Context,
 
 	// No custom record on the HTLC, so we have nothing to do.
 	if len(req.WireCustomRecords) == 0 {
+		// If there's no wire custom records and the invoice is an asset
+		// invoice do not settle the invoice.
```
guggero, line 142: nit: Should we be a bit more specific in what happens here? We cancel all HTLCs of the current set.

tapchannel/aux_invoice_manager.go
guggero: Right, we could have multiple inbound channels with the same asset. So I guess we should just check if _one_ of the hop hints has an SCID that corresponds to an RFQ ID.

rfq/manager.go
```diff
@@ -784,7 +789,7 @@ func (m *Manager) PeerAcceptedBuyQuotes() map[SerialisedScid]rfqmsg.BuyAccept {
 // to our node for the sale of assets.
 //
 //nolint:lll
-func (m *Manager) PeerAcceptedSellQuotes() map[SerialisedScid]rfqmsg.SellAccept {
+func (m *Manager) PeerAcceptedSellQuotes() SellAcceptMap {
```
guggero, line 792: nit: can remove the `//nolint` directives here and below now.

guggero (commented):
Looks good. Not approving yet because of dependencies and unaddressed comments. But once those are in, this is good to go.

tapchannel/aux_invoice_manager.go
GeorgeTsagk: Updated to embrace various routing hints / hop hints

we now look for at least 1 hop hint that related to a valid rfq

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,571 @@
...
+		select {
+		case <-done:
+		case <-time.After(testTimeout):
+			t.Fail()
+		}
+	}
+}
+
+// genRandomRfqID generates a random rfqmsg.ID value.
+func genRandomRfqID(t *rapid.T) rfqmsg.ID {
```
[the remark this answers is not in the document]
GeorgeTsagk: hmm I tried this but ended up always getting
```
        aux_invoice_manager_test.go:501: [rapid] draw rfq_id: rfqmsg.ID{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}
```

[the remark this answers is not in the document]
GeorgeTsagk: could something be off with the reflection?

tapchannel/aux_invoice_manager_test.go
GeorgeTsagk: Given the many fields that **may** be invalid, we would be gravitated towards spending the majority of the test runs in early failures.

For example, if the above `"invoice_exists"` draw was given a 50% chance, we'd effectively dedicate 50% of our total runs into the same invoice check in the code.

tapchannel/aux_invoice_manager_test.go
GeorgeTsagk: > IIUC this allows the test framework to cover more edge cases more quickly (and combinations of these variables that may be edge cases).

The rapid framework itself tends to stick more to the ends of specified ranges, in order to expose such edge case behavior.

> IIUC for these tests we really care about the relationship between these three values and not the absolute magnitudes.

Right, I changed the invoice amt / asset units drawn values to have a range which relates to each other. Added some docs on the related section

tapchannel/aux_invoice_manager_test.go
GeorgeTsagk: > Right now I don't think the tests cover route hints that exist but are not matches to the expected route hint.

We actually check that, but the mismatch is always generated on the rfq quote generation, it's the `"nodeID_mismatch_probability"` value

> Nor does it cover cases of more than 1 hint.

Extended code to have variable route/hop hint lengths

tapchannel/aux_invoice_manager_test.go
GeorgeTsagk: https://github.com/lightninglabs/taproot-assets/pull/1144#discussion_r1813478955

tapchannel/aux_invoice_manager_test.go
GeorgeTsagk: I will expose all those "magic numbers" that somehow express probability of smth happening as top level const values, so it's easier to inspect and change

tapchannel/aux_invoice_manager_test.go
GeorgeTsagk: https://github.com/lightninglabs/taproot-assets/pull/1144#discussion_r1813925080

tapchannel/aux_invoice_manager_test.go
GeorgeTsagk: https://github.com/lightninglabs/taproot-assets/pull/1144#discussion_r1813478955

tapchannel/aux_invoice_manager_test.go
GeorgeTsagk: true, casting between types makes this into more diff though

tapchannel/aux_invoice_manager.go
GeorgeTsagk: This was done with this feature case in mind: I create an invoice of value X, and i may receive X via sats or via assets in any combination

tapchannel/aux_invoice_manager.go
```diff
@@ -121,9 +153,25 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(_ context.Context,
...
+		//
+		// TODO(george): Strict-forwarding could be configurable?
+		if isAssetInvoice(req.Invoice, s) {
+			resp.CancelSet = true
+		}
+
 		return resp, nil
 	}
 
+	if req.Invoice.PaymentAddr == nil {
```
guggero, line 171: Why is this needed? I'm not sure if the payment address is always set for AMP payments (or just for MPP)...
If this is because of unit tests, perhaps just checking the invoice for being non-nil?

tapchannel/aux_invoice_manager.go
```diff
@@ -234,6 +282,63 @@ func (s *AuxInvoiceManager) priceFromQuote(rfqID rfqmsg.ID) (
...
+
+// isAssetInvoice checks whether the provided invoice is an asset invoice. This
+// method checks whether the routing hints of the invoice match those created
+// when generating an asset invoice, and if that's the case we then check that
+// the scid matches an existing quote.
+func isAssetInvoice(invoice *lnrpc.Invoice, rfqLookup RfqLookup) bool {
+	hints := invoice.RouteHints
+
+	for _, rh := range hints {
+		hint := rh
```
guggero, line 315: nit: no longer required with Golang 1.22 and later.

tapchannel/aux_invoice_manager.go
```diff
@@ -227,6 +238,64 @@ func (s *AuxInvoiceManager) priceFromQuote(rfqID rfqmsg.ID) (
...
+// then it returns the RFQ peer's node id.
+func (s *AuxInvoiceManager) rfqPeerFromScid(scid uint64) (route.Vertex, error) {
+	acceptedBuyQuotes := s.cfg.RfqManager.PeerAcceptedBuyQuotes()
+	acceptedSellQuotes := s.cfg.RfqManager.LocalAcceptedSellQuotes()
+
+	buyQuote, isBuy := acceptedBuyQuotes[rfqmsg.SerialisedScid(scid)]
+	sellQuote, isSell := acceptedSellQuotes[rfqmsg.SerialisedScid(scid)]
+
+	switch {
+	case isBuy:
```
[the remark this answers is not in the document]
guggero: Still unaddressed. If there's no good reason why we'd eve want to check a sell order, I think we should change this to only look at buy orders.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,362 @@
...
+
+// mockRfqManager mocks the interface of the rfq manager required by the aux
+// invoice manager. It also holds some internal state to return the desired
+// quotes.
+type mockRfqManager struct {
+	peerBuyQuotes   rfq.BuyAcceptMap
+	localSellQuotes rfq.SellAcceptMap
+}
+
+//nolint:lll
```
guggero, line 100: nit: nolint not required here and below.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,362 @@
...
+func (m *mockRfqManager) PeerAcceptedBuyQuotes() rfq.BuyAcceptMap {
+	return m.peerBuyQuotes
+}
+
+//nolint:lll
+func (m *mockRfqManager) LocalAcceptedSellQuotes() rfq.SellAcceptMap {
+	return m.localSellQuotes
+}
+
+// mockHtlcModifier mocks the HtlcModifier interface that is require by the Aux
```
guggero, line 110: nit: incomplete sentence.

guggero (commented):
Looks very close! Will do final review once dependencies are in.

tapchannel/aux_invoice_manager_test.go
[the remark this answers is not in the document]
jharveyb: Hmm, maybe - could instead generate a byte slice and cast that I suppose.

tapchannel/aux_invoice_manager_test.go
jharveyb: > For example, if the above `"invoice_exists"` draw was given a 50% chance, we'd effectively dedicate 50% of our total runs into the same invoice check in the code.

I don't think that's true; the framework should determine that this variable causes early failures and then leave this as `non-nil` for future iterations, after some iterations where this was `nil`.

Are you observing an issue with this version in practice?

go.sum
```diff
@@ -89,8 +89,9 @@ github.com/btcsuite/btcd/chaincfg/chainhash v1.0.0/go.mod h1:7SFka0XMvUgj3hfZtyd
 github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
 github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0 h1:59Kx4K6lzOW5w6nFlA0v5+lk/6sjybR934QNHSJZPTQ=
 github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
-github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f h1:bAs4lUbRJpnnkd9VhRV3jjAVU7DJVjMaK+IsvSeZvFo=
 github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f/go.mod h1:TdznJufoqS23FtqVCzL0ZqgP5MqXbb4fg/WgDys70nA=
+github.com/btcsuite/btclog v0.0.0-20241003133417-09c4e92e319c h1:4HxD1lBUGUddhzgaNgrCPsFWd7cGYNpeFUgd9ZIgyM0=
```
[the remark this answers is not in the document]
GeorgeTsagk: the issue was in `docs/examples/basic-price-oracle/go.mod` , I did perform a `go mod tidy` there and fixed it up with the first commit

jharveyb (commented):
Resetting my reviewer state.

tapchannel/aux_invoice_manager_test.go
guggero: I agree, we should let the framework decide what it "spends time on". That seems to be the whole purpose that it knows how to most efficiently discover code paths.

tapchannel/aux_invoice_manager.go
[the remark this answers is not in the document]
guggero: Turns out I was wrong here! In case of a direct-peer payment through an asset invoice, it can actually be a sell order SCID. The itest failed here, so will update it.

tapchannel/aux_invoice_manager_test.go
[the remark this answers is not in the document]
jharveyb: I think this is a generic issue with arrays; generating a slice and casting def works tho.

guggero:
Okay, I rolled back the `isBuy/isSell` change since that made the itest fail.

I should've just read the Godoc of the method, it was always right there :sweat_smile: 

```
// priceFromQuote retrieves the price from the accepted quote for the given RFQ
// ID. We allow the quote to either be a buy or a sell quote, since we don't
// know if this is a direct peer payment or a payment that is routed through the
// multiple hops. If it's a direct peer payment, then the quote will be a sell
// quote, since that's what the peer created to find out how many units to send
// for an invoice denominated in BTC.
```

tapchannel/aux_invoice_manager_test.go
```diff
@@ -0,0 +1,798 @@
...
+	case <-done:
+	case <-time.After(testTimeout):
+		t.Fail()
+	}
+}
+
+// TestAuxInvoiceManagerProperty runs property based tests on the
+// AuxInvoiceManager.
+func TestAuxInvoiceManagerProperty(t *testing.T) {
+	t.Parallel()
```
jharveyb, line 715: I think `rapid` tests are not thread safe? Not sure if that meant between multiple `rapid` tests though, or just any other test.

jharveyb (approved):
LGTM! Generators look a lot better.

I think we can still simplify them more with other rapid built-ins but this definitely works for now.

jharveyb:
Tested for 2,000,000 iterations, looks good!

tapchannel/aux_invoice_manager_test.go
guggero: Yeah, that's possible. But the `t.Parallel()` only means other tests in this file can run in parallel to the rapid test. And since we don't have any other rapid tests running there, this should be fine.

Roasbeef:
Wanted to provide a bit of drive by review here, re the usage of property based tests. 

I think the best way to approach property bsaed tests is from a black-box angle. So you come up with some invariants a given function/method should hold, then generate random inputs, prune that space to be relevant, and assert the output based on input. 

A trivial example of this is encoding, we want `decode(encode(x)) == x`. That just asserts that the functions should do, now how they do it (which internal fields they update, if they use a buffer or not, etc, etc). 

Glancing at this PR, I think the property tests are too white box, and too coarse. They use state in between each test to execute assertions, and [re-compute a lot of intermediate values](https://github.com/lightninglabs/taproot-assets/pull/1144/commits/d4d2af3e41ae88eef7d08bc92cb8258d376c77a0#diff-6110e3c3cc5e29a72d89835da9f62387ea715cb7d27c5ed288986b54a11501d9R212-R231) to attempt to assert correct behavior. In short, they're written in "normal" white box manner. 

If we take a step back and think about what we wanted here, we want an invariant that goes something like:
  * Given an incoming HTLC, and an invoice that may correspond to that HTLC, we only accept the HTLC if the payload in the HTLC matches the asset ID expectation in the invoice. 

This is the invariant that would prevent a known payment hash from being accepted with a normal HTLC, rather than an asset HTLC. 

If we look at the commit that added this logic, we see this section was changed: https://github.com/lightninglabs/taproot-assets/pull/1144/commits/ad057184918dfb8d7d485db6aef4c97a7e531c08#diff-5761d8e06dd213a5ff31aca2abe70153485d5d7832da64b35d371c0303e42bbdR156-R167. `isAssetInvoice` is what's meant to enforce that invariant. So that's the smallest minimal unit we need to test here. Instead of trying to mock/test the much alrge system of the invoice manager, we could just focus on proving the input needed for that routine, and assert the output. 

Then we can go another step up to also test the htlc wire custom records recognition. To do this, we'd need to break up the `handleInvoiceAccept` function into smaller components. Then we can write property tests on each of the components, along with the relevant composition of each component. 

If we wanted to test the bit that [handles summing up the set of HTLCs](https://github.com/GeorgeTsagk/taproot-assets/blob/ad057184918dfb8d7d485db6aef4c97a7e531c08/tapchannel/aux_invoice_manager.go#L223-L239), then rather than re-implement the logic to do a white box test, we can use another invariant. This might be somewhere along the lines of:
  * Only once a set of HTLCs that sum up to the invoice amount are extended do we accept the HTLC.

Roasbeef:
The package also has some support for testing more stateful machines which may be useful in the future, haven't tried it out myself though: https://pkg.go.dev/pgregory.net/rapid#T.Repeat

PR #1147, opened 2024-10-08
tapchannel: relax timing for quit and cancel tests

### LINKED ISSUE #1143: [bug]: `unit-test` `TestAuxLeafSignerCancelAndQuit` produces inconsistent results
### Background

`TestAuxLeafSignerCancelAndQuit` produces inconsistent tests results / flakes
https://github.com/lightninglabs/taproot-assets/actions/runs/11217147728/job/31178047667?pr=1140#step:189:207

jharveyb:
Fixes #1143 . I had replicated the original failure locally, race tested this change for 100 iterations locally.

### DIFF
--- a/tapchannel/auf_leaf_signer_test.go
+++ b/tapchannel/auf_leaf_signer_test.go
@@ -49,7 +49,7 @@ var (
 	}
 
 	// sig job batch size when making more that one sig job.
-	numSigJobs = int32(10)
+	numSigJobs = int32(25)
 
 	// Threshold for trying to cancel or quit the aux leaf signer (allow
 	// the signer to complete a third of the batch).
@@ -207,7 +207,6 @@ func TestAuxLeafSignerCancelAndQuit(t *testing.T) {
 		// Another component could have sent the cancel signal; we'll
 		// send that before the quit signal.
 		close(cancelChan)
-		time.Sleep(time.Millisecond)
 
 		// Send the quit signal; jobs at the end of the batch should not
 		// be processed.


GeorgeTsagk (approved):

tapchannel/auf_leaf_signer_test.go
```diff
@@ -207,7 +207,7 @@ func TestAuxLeafSignerCancelAndQuit(t *testing.T) {
 		// Another component could have sent the cancel signal; we'll
 		// send that before the quit signal.
 		close(cancelChan)
-		time.Sleep(time.Millisecond)
+		time.Sleep(time.Microsecond * 250)
```
guggero, line 210: Isn't this the opposite of "relaxing" the timing? Since this is now 0.25 milliseconds as before it was 1 millisecond?

gijswijs: What Oli says. :point_up:

Roasbeef: Can we get rid of this sleep?

jharveyb: The wording is sort of flipped; the 'timing' here is how long the aux signer will take to skip the remaining sig jobs. 

Ideally, the last job is never processed after the quit signal is sent. I didn't account for the fact that either case of the quit-cancel switch may fire, and that jobs will get processed a lot faster once one signal is received.

After adding jobs and decreasing the timeout, I can no longer repro this flake.

jharveyb: > Can we get rid of this sleep?

Actually, yes - I think there doesn't need to be known delay between the cancel and the quit, it was a leftover from fixing the initial issue with these signals.

guggero (approved):

PR #1151, opened 2024-10-16
tapd: fix L1 assets balance calculation

gijswijs:
This commits modifies the SQL queries `QueryAssetBalancesByGroup` and
`QueryAssetBalancesByAsset` to allow an `exclude_key` parameter to
exclude assets with that key from the balance calculation.

`assetsStore.QueryBalancesByAsset` and
`assetsStore.QueryAssetBalancesByGroup` use this parameter to exclude
assets that are specifically used for funding custom channels. The
balance of these assets should be reported through the L2 channel
balance and not through the L1 asset balance.

### DIFF
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -23,6 +23,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/tapdb/sqlc"
 	"github.com/lightninglabs/taproot-assets/tapfreighter"
 	"github.com/lightninglabs/taproot-assets/tappsbt"
+	"github.com/lightninglabs/taproot-assets/tapscript"
 	"github.com/lightninglabs/taproot-assets/tapsend"
 	"github.com/lightningnetwork/lnd/clock"
 	"github.com/lightningnetwork/lnd/keychain"
@@ -983,6 +984,15 @@ func (a *AssetStore) QueryBalancesByAsset(ctx context.Context,
 		},
 	}
 
+	// We exclude the assets that are specifically used for funding custom
+	// channels. The balance of those assets is reported through lnd channel
+	// balance. Those assets are identified by the funding script tree for a
+	// custom channel asset-level script key.
+	excludeKey := asset.NewScriptKey(
+		tapscript.NewChannelFundingScriptTree().TaprootKey,
+	)
+	assetBalancesFilter.ExcludeKey = excludeKey.PubKey.SerializeCompressed()
+
 	// By default, we only show assets that are not leased.
 	if !includeLeased {
 		assetBalancesFilter.Leased = sqlBool(false)
@@ -1054,6 +1064,15 @@ func (a *AssetStore) QueryAssetBalancesByGroup(ctx context.Context,
 			Valid: true,
 		},
 	}
+
+	// We exclude the assets that are specifically used for funding custom
+	// channels. The balance of those assets is reported through lnd channel
+	// balance. Those assets are identified by the funding script tree for a
+	// custom channel asset-level script key.
+	excludeKey := asset.NewScriptKey(
+		tapscript.NewChannelFundingScriptTree().TaprootKey,
+	)
+	assetBalancesFilter.ExcludeKey = excludeKey.PubKey.SerializeCompressed()
 
 	// By default, we only show assets that are not leased.
 	if !includeLeased {
--- a/tapdb/assets_store_test.go
+++ b/tapdb/assets_store_test.go
@@ -2349,3 +2349,64 @@ func TestQueryAssetBalances(t *testing.T) {
 	}
 	require.Equal(t, totalGroupedBalances, leasedBalanceByGroupSum)
 }
+
+func TestQueryAssetBalancesCustomChannelFunding(t *testing.T) {
+	t.Parallel()
+
+	_, assetsStore, _ := newAssetStore(t)
+	ctx := context.Background()
+
+	// First, we'll generate 2 assets, one of them having a script key that
+	// is the typical funding script key.
+	const numAssets = 2
+	const numGroups = 1
+	assetGen := newAssetGenerator(t, numAssets, numGroups)
+
+	fundingScriptKey := asset.NewScriptKey(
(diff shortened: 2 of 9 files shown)


tapdb/assets_store_test.go
```diff
@@ -2349,3 +2349,109 @@ func TestQueryAssetBalances(t *testing.T) {
...
+	require.Equal(t, assetDesc[1].amt, balanceByGroupSum)
+}
+
+// newFundingScriptTree creates a new funding script tree for a custom channel
+// asset-level script key. The script tree is constructed with a simple OP_TRUE
+// script that allows anyone to spend the output. This simplifies the funding
+// process as no signatures for the asset-level witnesses need to be exchanged.
+// This is still safe because the BTC level multi-sig output is still protected
+// by a 2-of-2 MuSig2 output.
+func newFundingScriptTree() *FundingScriptTree {
```
guggero, line 2421: Instead of creating a duplicate here, let's just move the original type to `tapscript`, which _should_ allow us to access it from `tapdb` without creating a circular package dependency.

tapdb/assets_store.go
```diff
@@ -1055,6 +1068,18 @@ func (a *AssetStore) QueryAssetBalancesByGroup(ctx context.Context,
 		},
 	}
 
+	// We exclude the assets that are specifically used for funding custom
+	// channels. The balance of those assets is reported through lnd channel
+	// balance. Those assets are identified by the funding script tree for a
+	// custom channel asset-level script key.
+	excludeKeyStr := "0350aaeb166f4234650d84a2d8a130987aeaf6950206e090540" +
```
guggero, line 1075: We should try to derive this from `newFundingScriptTree()` instead. See my other comment on how to do that without getting into package dependency hell.

guggero (commented):
Nice!

tapdb/assets_store.go
gijswijs: If we do that, we should add a unit test whether `newFundingScriptTree()` returns that value though. We have to be sure that value does not accidentally change, don't you think?

tapdb/sqlc/queries/assets.sql
```diff
@@ -332,7 +332,11 @@ JOIN managed_utxos utxos
                 utxos.lease_expiry <= @now)
            ELSE TRUE
        END
-WHERE spent = FALSE 
+JOIN script_keys
+    ON assets.script_key_id = script_keys.script_key_id
+WHERE spent = FALSE AND 
+        (script_keys.tweaked_script_key <> sqlc.narg('exclude_key') OR
```
guggero, line 338: nit: looks like we're using the equivalent `!=` operator in another place. Since that seems to be widely supported in most modern databases, I actually prefer that because of its same meaning in Go.

tapscript/script_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestNewFundingScriptTree tests the NewFundingScriptTree function. We use this
+// function to filter out assets used for custom channel funding from the
+// balance queries. Therefor the script key cannot change since this would lead
+// to inconsistencies in the balance queries.
+func TestNewFundingScriptTree(t *testing.T) {
+	expectedTaprootKeyHex := "0350aaeb166f4234650d84a2d8a130987aeaf695020" +
```
guggero, line 15: See my previous comment, but this key, threaded through `asset.NewScriptKey()` would start with `02`. Not sure if it makes sense to change this code after all, but perhaps a comment would be useful?

guggero (commented):
Very nice! One small (but important) technicality, other than that looks great :tada:

tapdb/sqlc/queries/assets.sql
gijswijs: That's a force of my old SQL habits seeping through right there.

tapscript/script.go
```diff
@@ -7,7 +7,7 @@ import (
 
 // anyoneCanSpendScript is a simple script that allows anyone to spend the
 // output.
-func anyoneCanSpendScript() []byte {
+func AnyoneCanSpendScript() []byte {
```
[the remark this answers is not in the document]
gijswijs: VsCode shill: F2 also renames the Godoc comment. :smile:

guggero (approved):
Nice, LGTM :tada:

tapscript/script.go
```diff
@@ -25,7 +25,7 @@ type FundingScriptTree struct {
 // by a 2-of-2 MuSig2 output.
 func NewFundingScriptTree() *FundingScriptTree {
```
GeorgeTsagk, line 26: since this is no longer in `tapchannel` we could indicate in the func name that this is channel specific

tapdb/assets_store.go
```diff
@@ -983,6 +984,15 @@ func (a *AssetStore) QueryBalancesByAsset(ctx context.Context,
 		},
 	}
 
+	// We exclude the assets that are specifically used for funding custom
+	// channels. The balance of those assets is reported through lnd channel
+	// balance. Those assets are identified by the funding script tree for a
+	// custom channel asset-level script key.
+	excludeKey := asset.NewScriptKey(
```
GeorgeTsagk, line 991: if I pay assets to an OP_TRUE scriptkey for other purposes (i.e an in-flight trustless swap) it will also ignore them (in that case, it's desired as a pending swap is not part of my balance)

Maybe some other use-case (a not-so-secure wallet..?) is going to hold assets in OP_TRUE script keys and still control the btc level utxos, in that case we would report 0 balance

non-blocking, just worth a mention

GeorgeTsagk (approved):
LGTM :ringed_planet:

tapdb/assets_store.go
gijswijs: Well, in the case of a not-so-secure wallet it's also correct behavior, since you should assume it's also not part of your balance anymore. :smile:

guggero:
Uh, linter is unhappy because of a too-long line... Will merge once fixed.

PR #1157, opened 2024-10-22
rfqmsg: request fields blip align

ffranr:
In this PR, we align the request RFQ wire message fields with the tap BLIP.

All TLV record values are already aligned with the BLIP due to previous PRs. This PR introduces internal changes to `tapd` to ensure the `request` message fields are handled, named, and interpreted correctly. These changes are non-interface and non-breaking.

This will be the last RFQ BLIP alignment PR.


Roasbeef:
We should also make sure the min+max fields are set properly, along with the transfer type.

guggero:
We should probably also remove the `--experimental.rfq.mockoraclesatsperasset=` flag in this PR (or a related PR).

ffranr:
> We should probably also remove the `--experimental.rfq.mockoraclesatsperasset=` flag in this PR (or a related PR).

@guggero I would prefer to do that in a future PR. I want to prioritize BLIP alignment and address any bugs arising from misalignment issues.

rfq/order.go
```diff
@@ -260,19 +261,28 @@ func (c *AssetPurchasePolicy) CheckHtlcCompliance(
...
+	// Sum the asset balance in the HTLC record.
+	assetAmt, err := htlcRecord.SumAssetBalance(c.AssetSpecifier)
+	if err != nil {
+		return fmt.Errorf("error summing asset balance: %w", err)
+	}
+
 	// Convert the inbound asset amount to millisatoshis and ensure that the
 	// outgoing HTLC amount is not more than the inbound asset amount.
-	assetAmt := rfqmath.NewBigIntFixedPoint(c.AssetAmount, 0)
+	assetAmtFp := rfqmath.BigIntFixedPoint{
```
Roasbeef, line 272: I think the constructor can still be used here?

rfqmsg/request.go
```diff
@@ -99,6 +88,17 @@ type requestWireMsgData struct {
 	// outbound to the requesting peer (therefore inbound to the
 	// counterparty peer).
 	OutAssetGroupKey requestOutAssetGroupKey
+
+	// AssetMaxAmount represents the maximum asset amount that the target
+	// peer is expected to accept/divest.
+	AssetMaxAmount tlv.RecordT[tlv.TlvType16, uint64]
```
guggero: :pray: thanks for re-ordering.

asset/asset.go
```diff
@@ -263,6 +263,47 @@ type Specifier struct {
...
+	if !isAnySpecified && mustBeSpecified {
+		return Specifier{}, fmt.Errorf("at least one of the asset ID "+
+			"or asset group key fields must be specified "+
+			"(id=%v, groupPubKey=%v, groupKey=%v)",
+			id, groupPubKey, groupKey)
+	}
+
+	// Create an option for the asset ID.
+	var optId fn.Option[ID]
+	if id != nil {
```
guggero, line 287: We have `fn.MaybeSome()` (which we should port over to `lnd` as well since it's quite useful).

asset/asset.go
```diff
@@ -349,6 +349,22 @@ func NewSpecifierFromGroupKey(groupPubKey btcec.PublicKey) Specifier {
...
+	s.WhenId(func(id ID) {
+		assetID = id
+	})
+
+	var groupKeyBytes []byte
+	s.WhenGroupPubKey(func(key btcec.PublicKey) {
+		groupKeyBytes = key.SerializeCompressed()
+	})
+
+	return fmt.Sprintf("AssetSpecifier(id=%s, group_pub_key=%x)",
```
guggero, line 364: Since we now kind of have a semantic distinction between the asset being specified (`fn.Option`) or all zero (for specifying an asset is BTC), shouldn't we also make the distinction here?
Meaning that if the asset ID isn't set, we'd print `id=, group_pub_key`?

rfq/negotiator.go
```diff
@@ -405,8 +405,9 @@ func (n *Negotiator) HandleIncomingSellRequest(
 		// Query the price oracle for a bid price. This is the price we
 		// are willing to pay for the asset that our peer is trying to
 		// sell to us.
+		assetID, assetGroupKey := request.AssetSpecifier.UnwrapToPtr()
 		assetRate, rateExpiry, err := n.queryBidFromPriceOracle(
-			request.Peer, request.AssetID, request.AssetGroupKey,
+			request.Peer, assetID, assetGroupKey,
```
guggero, line 409: Should we instead pass in the `Specifier` into `queryBidFromPriceOracle` to avoid some duplication? Then we likely also don't have to unwrap to pointer?

rfqmath/convert_test.go
```diff
@@ -498,11 +497,9 @@ func TestPriceOracleRateExample(t *testing.T) {
 	// BTC rate into tap asset units per BTC by applying a multiplier
 	// (`dollarToTap`) based on the asset’s decimal display.
 	decimalDisplay := 4
-	dollarToTap := NewBigInt(
-		big.NewInt(0).SetUint64(uint64(
-			math.Pow(float64(10), float64(decimalDisplay)),
-		)),
-	)
+	dollarToTap := NewBigIntFromUint64(uint64(
+		math.Pow(float64(10), float64(decimalDisplay)),
```
guggero, line 501: nit: can use `math.Pow10`.

rfqmsg/records_test.go
```diff
@@ -39,6 +45,21 @@ func assetHtlcTestCase(t *testing.T, tc htlcTestCase) {
...
 	}
+
+	// Check expected asset sum balances if specified in test case.
+	if tc.sumBalances == nil {
+		tc.sumBalances = make(map[asset.ID]rfqmath.BigInt)
+	}
+
+	for assetID, expectedBalance := range tc.sumBalances {
+		assetSpecifier := asset.NewSpecifierFromId(
+			assetID,
```
guggero, line 56: nit: looks like this would fit on previous line?

rfq/negotiator.go
ffranr: `queryBidFromPriceOracle` is called given a `BuyRequest` in another part of the code. And the `BuyRequest` does not have an `AssetSpecifier` field yet. I will add and then `queryBidFromPriceOracle` can accept an `asset.Specifier`.

rfq/order.go
ffranr: In the latest commits, I've introduced a new fixed-point method, `SetIntValue`, which enables constructing a fixed-point instance from any integer value of the same type as the fixed-point coefficient's type:
```
assetAmtFp := new(rfqmath.BigIntFixedPoint).SetIntValue(assetAmt)
```

ffranr:
I think we should drop the `TransferType` field from the request message. As I was adding it I found that it was pulling in unnecessary context which I found made things more complicated. It relates different request messages to what there intended big picture uses might be. Which is not something the request/accept/reject RFQ messages need to be aware of IMO. It's a bit like a "note" or "label" but doesn't seem to add much when applied to the message level.

What do you think @guggero ?

ffranr:
Things left to do in this PR:

I've added the `MinInAsset` and `MinOutAsset` request message fields. I think we need to add reject message error types for these fields as the accept message doesn't propose other lower bounds.

And we need to reject based on the dust limit in relation to `MinInAsset` or `MinOutAsset` depending on whichever one is BTC. Not sure where the dust limit is defined so might add my own msats dust limit const.

Any other use of these fields will probably be rewritten by the MPP OKR.

rfq/negotiator.go
ffranr: I notice now that `BuyOrder` will need the asset specifier upgrade at the same time as `queryBidFromPriceOracle`. I'm just going to leave that for a future PR. It distracts from the main goal of this PR, IMO.

ffranr:
I think we should merge this PR as is, subject to reviews.

rfq/order.go
```diff
@@ -80,13 +80,15 @@ type Policy interface {
...
+	AcceptedQuoteId rfqmsg.ID
 
-	// MaxAssetAmount is the maximum amount of the asset that is being
-	// requested.
-	MaxAssetAmount uint64
+	// MaxOutboundAssetAmount represents the maximum asset amount permitted
+	// by policy for outbound transactions. It sets an upper limit on the
+	// amount of assets this node is willing to divest within the remit of
+	// the policy.
+	MaxOutboundAssetAmount uint64
```
guggero, line 95: IMO this name is a bit confusing. Since in the RFQ doc and on the wire we actually talk about the `max_in_asset`... 

In my mind it would be more consistent if this was called `MaxInAsset` here, the asset buy policy as well as in the `BuyRequest` and `SellRequest` messages.

OR do the switch at the transition from wire to request struct, but then have the units correct on the request structs (one being `uint64` and the other being `lnwire.MilliSatoshi`).
And the policies should 100% match the request structs (both in name and unit type), otherwise it'll be even more confusing.

guggero (commented):
Thanks for all the updates. I think we are fairly close here, but I have a couple of things that we should try to get in with this PR, as IMO the changes are within the scope of the PR:

1. Rename the `SellRequest`'s `AssetAmount` either to `MaxInAsset` with `uint64` or to `MaxInboundAmount` with `lnwire.MilliSatoshi` (depending on the decision where to make the switch, see inline comment).
2. Enforce the above max amount in the asset purchase policy.
3. Since we don't have the "straightening out" of the buy/ask/purchase terminology I had in my WIP branch, and we still turn a buy request into an ask request for the oracle (which internally is then called a sale transaction), we should add that to the RFQ doc, best with a sequence diagram. Otherwise it will endlessly confuse me, and probably any authors of oracle logic as well.

guggero:
> I think we should drop the `TransferType` field from the request message.

We currently decide whether it's a buy or sell request based on which asset is BTC here:
https://github.com/lightninglabs/taproot-assets/blob/e1e8268049f1dcc46b29df48c699ad5891bfa3c4/rfqmsg/request.go#L503

Once we support asset-to-asset conversion directly, how would we distinguish between buy and sell? Just the _direction_ of the in vs. out asset?

ffranr:
> > I think we should drop the `TransferType` field from the request message.
> 
> We currently decide whether it's a buy or sell request based on which asset is BTC here:
> 
> https://github.com/lightninglabs/taproot-assets/blob/e1e8268049f1dcc46b29df48c699ad5891bfa3c4/rfqmsg/request.go#L503
> 
> Once we support asset-to-asset conversion directly, how would we distinguish between buy and sell? Just the _direction_ of the in vs. out asset?

IMO we should also drop the buy/sell types when we move to support asset-to-asset. So the transfer type is not needed then either.

rfq/order.go
ffranr: `order.go` is completely from the point of view of an edge node. That's one reason why we're using "policy" structs here and copying over fields rather than just carrying along an entire accepted quote.

I think things would be more confusing if we write everything from a wallet node's PoV including the edge node only routines.

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -121,9 +121,14 @@ message QueryAssetRatesRequest {
     // unit
     AssetSpecifier payment_asset = 4;
 
+    // payment_asset_max_amount is the maximum amount of the payment asset that
+    // could be involved in the transaction. This field is optional. If set to
+    // zero, it is considered unset.
+    uint64 payment_asset_max_amount = 5;
```
guggero, line 127: Do we need this field for an actual use case today? I think we should just remove it to make things easier to understand... See my other comments as well.

rfq/negotiator.go
```diff
@@ -177,9 +177,12 @@ func (n *Negotiator) HandleOutgoingBuyOrder(buyOrder BuyOrder) error {
 
 		if n.cfg.PriceOracle != nil && assetSpecifier.IsSome() {
 			// Query the price oracle for a bid price.
+			//
+			// TODO(ffranr): Add assetMaxAmt to BuyOrder and use as
```
guggero, line 181: This TODO remains unresolved and we don't pass any amount to the oracle.

rpcserver.go
```diff
@@ -6950,11 +6948,12 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
-		// (invoice amount plus fee limit) in milli-satoshi, not the
-		// asset amount. Need to change the whole RFQ API to do that
-		// though.
-		maxAssetAmount := assetChan.assetInfo.LocalBalance
+		// paymentMaxAmt is the maximum amount that the counterparty is
+		// expected to pay. This is the amount that the invoice is
+		// asking for plus the fee limit in milli-satoshis.
+		//
+		// TODO(ffranr): Add the fee limit to this amount.
+		paymentMaxAmt := *invoice.MilliSat
```
guggero, line 6956:
```
		// paymentMaxAmt is the maximum amount that the counterparty is
		// expected to pay. This is the amount that the invoice is
		// asking for plus the fee limit in milli-satoshis.
		var paymentMaxAmt lnwire.MilliSatoshi
		if invoice.MilliSat == nil {
			amt, err := lnrpc.UnmarshallAmt(pReq.Amt, pReq.AmtMsat)
			if err != nil {
				return fmt.Errorf("error unmarshalling " +
					"amount: %w", err)
			}
			if amt == 0 {
				return errors.New("amount must be " +
					"specified when paying a zero amount " +
					"invoice")
			}

			paymentMaxAmt = amt
		} else {
			paymentMaxAmt = *invoice.MilliSat
		}

		// Calculate the fee limit that should be used for this payment.
		feeLimit, err := lnrpc.UnmarshallAmt(
			pReq.FeeLimitSat, pReq.FeeLimitMsat,
		)
		if err != nil {
			return fmt.Errorf("error unmarshalling fee limit: %w",
				err)
		}

		paymentMaxAmt += feeLimit
```

guggero (commented):
Almost there, thanks a lot for sticking to this huge undertaking :100:

rfq/negotiator.go
```diff
@@ -162,11 +162,22 @@ func (n *Negotiator) HandleOutgoingBuyOrder(buyOrder BuyOrder) error {
 		// skip this step.
 		var assetRateHint fn.Option[rfqmsg.AssetRate]
 
-		if n.cfg.PriceOracle != nil {
+		// Construct an asset specifier from the order.
+		// TODO(ffranr): The order should have an asset specifier.
```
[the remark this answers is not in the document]
ffranr: This TODO is trying to say that `buyOrder` should have an `asset.Specifier` type field rather than constructing an asset specifier at this point. I will clarify the todo in code.

rfq/negotiator.go
```diff
@@ -106,7 +106,7 @@ func NewNegotiator(cfg NegotiatorCfg) (*Negotiator, error) {
 // queryBidFromPriceOracle queries the price oracle for a bid price. It returns
 // an appropriate outgoing response message which should be sent to the peer.
 func (n *Negotiator) queryBidFromPriceOracle(peer route.Vertex,
-	assetSpecifier asset.Specifier, assetAmount uint64,
+	assetSpecifier asset.Specifier, assetMaxAmt fn.Option[uint64],
```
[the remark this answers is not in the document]
ffranr: > Is the assetMaxAmount used anywhere anymore after this series of commits?

`assetMaxAmt` is passed on to `n.cfg.PriceOracle.QueryBidPrice` below. It gets passed to the price oracle service.

> Can't we just remove it, and make the `paymentMaxAmount` a non-option and just use that everywhere?

These two fields relate to two different assets.

rfq/negotiator.go
ffranr: Adding and populating that field in the `BuyOrder` would require an RPC change. I mean to do that in a future PR. I think this PR is getting too long.

taprpc/priceoraclerpc/price_oracle.proto
ffranr: `MaxInAsset` sometimes relates to BTC and sometimes to a tap asset.

The two different fields in the price oracle relate to the two different assets that we're tying to find an exchange rate for. Instead of a single field that is sometimes the upper bound on the tap asset and sometimes an upper bound on the BTC side, I've got two different fields.

rpcserver.go
ffranr: Thank you!

taprpc/priceoraclerpc/price_oracle.proto
guggero: Okay, I guess I see where my confusion is coming from. It's because we still do the "flip" of the transaction type for the oracle (which I eliminated in my WIP refactor PR).

What I mean is:
 - On the wallet end user side, a `SellOrder` goes into the oracle as an `ask` (transaction type `sale` with `payment_assset_max_amount` set)
 - On the edge node, that same `SellOrder` goes into the oracle as a `bid` (transaction type `purchase` with `payment_max_amount` set)

And then for a `BuyOrder` it would be kind of the other way around.
Can we please add all of this to the RFQ documentation?

And perhaps update the comments in this proto file to make it more clear to the developer of an oracle when what field is set? For example that never both `_max_amount` fields are set at the same time?

rfq/negotiator.go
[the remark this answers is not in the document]
guggero: Okay, I see. Confusion on my part, see comment below with a request for documenting this a bit more.

rfq/negotiator.go
guggero: But at this specific call site, we only pass in `fn.None` for all the values. So we're not passing any amount when requesting a rate for the hint. Is that an allowed way to call into the oracle, with none of the `_max_amount` fields set?

guggero (approved):
Okay, I think I finally understand all the changes and aligned my own mental model to them. Would love to have a bit more documentation around that, but not blocking this PR on it. So LGTM, nice work! :tada:

rfq/negotiator.go
ffranr: I think it should technically be a valid way of calling the price oracle. If upper bounds aren't given then the price would just be less accurate. But we should complete the TODO. I'll add the correct field to `BuyOrder` in a separate PR.

ffranr:
Further work to be included in a separate PR:

* Better documentation for `SellRequest`, `BuyRequest`, `BuyOrder`, `SellOrder`.
* Add missing field to `BuyOrder` and populate in RPC endpoint.

guggero:
> Further work to be included in a separate PR:
> 
>     * Better documentation for `SellRequest`, `BuyRequest`, `BuyOrder`, `SellOrder`.
> 
>     * Add missing field to `BuyOrder` and populate in RPC endpoint.


Sure, let's do all of that in a follow-up PR, then we can discuss there.

> @guggero are you also ok with the following:
> 
>     * Remove transfer type field from `request` RFQ message in BLIP and revise TLV integers in BLIP and tapd.

I still don't fully see how that would work, given that we'd still need to know whether to ask the oracle for an ask or bid. Or would we also remove that directionality there and only have a single RPC method?

ffranr:
> I still don't fully see how that would work, given that we'd still need to know whether to ask the oracle for an ask or bid. Or would we also remove that directionality there and only have a single RPC method?

Yes, I think we only need a single oracle interface method. The oracle needs to know what asset is inbound and which is outbound and then it can return the ask/bid based on that. We can do that now with the oracle RPC endpoint by setting payment asset, subject asset, and the transfer type. But we can simplify the RPC endpoint's messages further also.

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -121,9 +121,14 @@ message QueryAssetRatesRequest {
...
+
     // asset_rates_hint is an optional suggestion of asset rates for the
     // transaction, intended to provide guidance on expected pricing.
-    AssetRates asset_rates_hint = 5;
+    AssetRates asset_rates_hint = 6;
```
GeorgeTsagk, line 131: are we ok with breaking this?

rpcserver.go
```diff
@@ -6601,11 +6599,12 @@ func marshalPeerAcceptedSellQuotes(quotes map[rfq.SerialisedScid]rfqmsg.SellAcce
 			Scale:       uint32(quote.AssetRate.Scale),
 		}
 
+		// TODO(ffranr): Add SellRequest payment max amount to
+		//  PeerAcceptedSellQuote.
```
GeorgeTsagk, line 6603: shouldn't this TODO be addressed before merging?

GeorgeTsagk (approved):
Very clean commits :100: 

LGTM, straight forward changes :white_check_mark: 

have some minor Qs/nits, non-blocking

rpcserver.go
ffranr: I'll put up another PR to address this. This PR is getting a bit long.

PR #1160, opened 2024-10-28
Ensure that asset rate hint expiry timestamps are passed to price oracle

ffranr:
This PR refines the RFQ code to ensure that asset rate expiry timestamps are correctly passed to and retrieved from the price oracle service.

Previously, rate expiry timestamps were not consistently provided to the price oracle.

### Key Changes

1. **New `AssetRate` Type**:  
   Introduces the `AssetRate` type to represent an asset's exchange rate to BTC, along with its expiration timestamp.
   
   ```go
   // AssetRate represents the exchange rate of an asset to BTC, encapsulating
   // both the rate in fixed-point format and an expiration timestamp.
   //
   // Combining these fields ensures that each rate is inherently tied to an
   // expiry, clarifying its validity period.
   type AssetRate struct {
       // Rate defines the exchange rate of asset units to BTC using a
       // fixed-point representation, ensuring precision for fractional asset
       // rates.
       Rate rfqmath.BigIntFixedPoint

       // Expiry indicates the UTC timestamp when this rate expires and should
       // no longer be considered valid.
       Expiry time.Time
   }
   ```

2. **Integration in Function Signatures and Request Structures**:  
   The `AssetRate` type is now used in the signatures of price oracle query functions and request message structures. This change standardizes the handling of asset rates with expiry, reducing ambiguity and potential errors related to rate validity.

By encapsulating asset rates with expiry timestamps, this PR improves the reliability and clarity of rate management in the RFQ flow.

### DIFF
--- a/rfq/negotiator.go
+++ b/rfq/negotiator.go
@@ -450,11 +450,11 @@ func (n *Negotiator) HandleOutgoingSellOrder(order SellOrder) {
 		// We calculate a proposed ask price for our peer's
 		// consideration. If a price oracle is not specified we will
 		// skip this step.
-		var assetRate fn.Option[rfqmath.BigIntFixedPoint]
+		var assetRate fn.Option[rfqmsg.AssetRate]
 
 		if n.cfg.PriceOracle != nil {
 			// Query the price oracle for an asking price.
-			rate, _, err := n.queryAskFromPriceOracle(
+			rate, expiryUnix, err := n.queryAskFromPriceOracle(
 				order.Peer, order.AssetID, order.AssetGroupKey,
 				order.MaxAssetAmount,
 				fn.None[rfqmsg.AssetRate](),
@@ -466,7 +466,10 @@ func (n *Negotiator) HandleOutgoingSellOrder(order SellOrder) {
 				return
 			}
 
-			assetRate = fn.Some[rfqmath.BigIntFixedPoint](*rate)
+			expiry := time.Unix(int64(expiryUnix), 0)
+			assetRate = fn.Some[rfqmsg.AssetRate](
+				rfqmsg.NewAssetRate(*rate, expiry),
+			)
 		}
 
 		request, err := rfqmsg.NewSellRequest(
--- a/rfq/negotiator.go
+++ b/rfq/negotiator.go
@@ -162,11 +162,11 @@ func (n *Negotiator) HandleOutgoingBuyOrder(buyOrder BuyOrder) error {
 		// We calculate a proposed bid price for our peer's
 		// consideration. If a price oracle is not specified we will
 		// skip this step.
-		var assetRateBid fn.Option[rfqmath.BigIntFixedPoint]
+		var assetRateHint fn.Option[rfqmsg.AssetRate]
 
 		if n.cfg.PriceOracle != nil {
 			// Query the price oracle for a bid price.
-			rate, _, err := n.queryBidFromPriceOracle(
+			rate, expiryUnix, err := n.queryBidFromPriceOracle(
 				*buyOrder.Peer, buyOrder.AssetID,
 				buyOrder.AssetGroupKey, buyOrder.MinAssetAmount,
 				fn.None[rfqmsg.AssetRate](),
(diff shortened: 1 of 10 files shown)


rfqmsg/buy_request.go
```diff
@@ -148,6 +162,17 @@ func (q *BuyRequest) Validate() error {
 			q.Version)
 	}
 
+	// Ensure that the suggested asset rate has not expired.
+	var err error
+	q.AssetRateHint.WhenSome(func(rate AssetRate) {
```
guggero, line 159: nit: should use `fn.MapOptionZ` to return the error and not have the closure modify outside variables.

rfqmsg/buy_request.go
```diff
@@ -40,33 +42,39 @@ type BuyRequest struct {
...
 
 	id, err := NewID()
 	if err != nil {
 		return nil, fmt.Errorf("unable to generate random "+
 			"quote request id: %w", err)
 	}
 
+	// Construct a suggested asset rate if a rate hint is provided.
+	var assetRateHint fn.Option[AssetRate]
+	rateHint.WhenSome(func(rate rfqmath.BigIntFixedPoint) {
```
guggero: nit: can use `MapOptionZ` here.

rfqmsg/sell_request.go
```diff
@@ -199,9 +206,9 @@ func (q *SellRequest) String() string {
 	}
 
 	return fmt.Sprintf("SellRequest(peer=%x, id=%x, asset_id=%s, "+
-		"asset_group_key=%x, asset_amount=%d, ask_asset_rate=%v)",
+		"asset_group_key=%x, asset_amount=%d, asset_rate_hint=%v)",
 		q.Peer[:], q.ID[:], q.AssetID, groupKeyBytes, q.AssetAmount,
-		q.SuggestedAssetRate)
+		q.AssetRateHint)
```
guggero, line 202: Do we have a `String()` method on the `Rate` struct?

rfq/oracle.go
```diff
@@ -324,6 +325,42 @@ func (r *RpcPriceOracle) QueryBidPrice(ctx context.Context, assetId *asset.ID,
 	// set the subject asset ID.
 	copy(subjectAssetId, assetId[:])
 
+	// Construct the RPC asset rates hint.
+	var (
+		rpcAssetRatesHint *oraclerpc.AssetRates
+		err               error
+	)
+	assetRateHint.WhenSome(func(assetRate rfqmsg.AssetRate) {
```
guggero, line 339: Should use `fn.MapOptionZ` if there's an error return variable.

rfq/negotiator.go
```diff
@@ -180,9 +179,8 @@ func (n *Negotiator) HandleOutgoingBuyOrder(buyOrder BuyOrder) error {
 					"request: %v", err)
 			}
 
-			expiry := time.Unix(int64(expiryUnix), 0)
 			assetRateHint = fn.Some[rfqmsg.AssetRate](
-				rfqmsg.NewAssetRate(*rate, expiry),
+				*assetRate,
```
guggero, line 183: nit: could now fit on previous line.

guggero (approved):
Nice cleanup, LGTM :tada:

rfq/oracle.go
ffranr: Even if I use `fn.MapOptionZ` here I'll still need to modify one variable in the outer scope. Using `fn.Result` doesn't seem to work because the "zero" value will be an error at best I think.

guggero: Here are my two proposals for a different approach (preferring variant 1):

[variant1.diff.txt](https://github.com/user-attachments/files/17562008/variant1.diff.txt)
[variant2.diff.txt](https://github.com/user-attachments/files/17562007/variant2.diff.txt)

ffranr: Unit tests don't pass on variant 1. That's what I ran into also.

When `assetRateHint` is `None`, we want to end up with `rpcAssetRatesHint` with value of `nil`. But instead we end up with `rpcAssetRatesHint` with value of `AssetRates{}`.

And when the price oracle service inspects the `QueryAssetRatesRequest.AssetRatesHint` field it is present but unset instead of just `nil`. In my mind that's an invalid request because the expiry timestamp is 0.

ffranr: In [variant2.diff.txt](https://github.com/user-attachments/files/17562007/variant2.diff.txt) the err is returned but the result is applied to an out of scope variable. I don't see the win there.

guggero: Okay, I guess variant 1 would then need to have a pointer in the result, then it would properly be `nil`. Not super nice to have a pointer within an option/result, but at least we'd keep the composability of those functional/optional uses and don't go the "side effect" route...

gijswijs (approved):
Very nice improvement. I've run the unit tests on _all_ commitments and they pass with flying colours. (We should probably make that part of the CI, testing each individual commit)

No changes required. :tada:

Issue #1105, opened 2024-08-23
[bug]: Validate transaction's fee for min-relay suitability before broadcast


HannahMR:
### Background

[Maintainer edit: Transactions' fees may be created which are insufficient to meet the backend's stipulated minimum relay amount]
If an asset channel is opened without specifying the sats per vbyte amount, a too low fee may be used. The channel then gets stuck as pending and another channel can't be opened as there is already a pending channel for that asset. 

### My environment

I believe I was running litd v0.13.99 in integrated mode on a Ubuntu server at the time this error occurred. 

### Steps to reproduce

Open an asset channel using the standard litcli command while not specifying the sats per vbyte amount. 

Command used to open channel:

    $ litcli --network=testnet --tlscertpath ~/.lit/tls.cert --macaroonpath=~/.lnd/data/chain/bitcoin/testnet/admin.macaroon ln fundchannel --node_key [peer node pub key]  --asset_amount [amount] --asset_id [id of asset to be included in the channel]

### Expected behavior

An appropriate fee is used for the tx, the tx is confirmed and the channel is opened. 

### Actual behavior

The fee is insufficient and the channel is not broadcast and stays in a pending state preventing the opening of another channel using that asset. 

Error received: 

    [litcli] error funding channel: rpc error: code = Unknown desc = error funding channel: error requesting delivery: unable to broadcast transaction f1f72cf96a37b677798392824744112da8f09015f6fc89dfe6de335cdf8a3a0e: rpc error: code = Unknown desc = undefined: min relay fee not met

Appearing in the logs:

    2024-07-17  17:57:57.885  [ERR] BTCN: Broadcast attempt failed: undefined: -26: min relay fee not met, 254 < 255

dstadulis:
Thanks!  Will tweak the issue a bit

jharveyb:
AFAICT we're funding the TX without sanity checking the resulting feerate in `tapd` before we sign the TX and pass it to LND. I think we actually have a similar issue elsewhere, where we fund without immediately checking the resulting TX against some floor feerate. We could probably just add that check to `WalletAnchor`.

Funding for channel open:

https://github.com/lightninglabs/taproot-assets/blob/72b93f84a0afa08e01c99d582357672133fc9b20/tapchannel/aux_funding_controller.go#L1063

A heavyweight check of TX validity:

[...]

PR #1163, opened 2024-10-30
min_relay_fee check during channel funding

gijswijs:
This PR adds `MinRelayFee` to the `WalletAnchor` and uses it during tapchannel funding to check whether the feerate meets the minimum relay fee.

fixes: #1105

~~This PR depends on  https://github.com/lightninglabs/lndclient/pull/200 and uses a pseudoversion of that package.~~

~~go.mod should be changed before merging this PR,~~

### DIFF
--- a/tapchannel/aux_funding_controller.go
+++ b/tapchannel/aux_funding_controller.go
@@ -1354,6 +1354,18 @@ func (f *FundingController) processFundingReq(fundingFlows fundingFlowIndex,
 			fundReq.PeerPub.SerializeCompressed())
 	}
 
+	// Before we proceed, we'll make sure the fee rate we're using is above
+	// the min relay fee.
+	minRelayFee, err := f.cfg.ChainWallet.MinRelayFee(fundReq.ctx)
+	if err != nil {
+		return fmt.Errorf("unable to establish min_relay_fee: %w",
+			err)
+	}
+	if fundReq.FeeRate.FeePerKWeight() < minRelayFee {
+		return fmt.Errorf("fee rate %v too low, min_relay_fee: %v",
+			fundReq.FeeRate.FeePerKWeight(), minRelayFee)
+	}
+
 	// To start, we'll make a new pending asset funding desc. This'll be
 	// our scratch pad during the asset funding process.
 	tempPID, err := newPendingChanID()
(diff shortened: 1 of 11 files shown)


guggero:
Something I forgot to add in my "update to 0.18.4" PR which we should add here (since we're using another RPC field of `lnd` that's only available in `v0.18.4-beta` and later: We should bump [this version](https://github.com/lightninglabs/taproot-assets/blob/main/tapcfg/config.go#L187) to `0.18.4`.

tapchannel/aux_funding_controller.go
```diff
@@ -1058,6 +1058,18 @@ func (f *FundingController) completeChannelFunding(ctx context.Context,
...
 	log.Debugf("Funding PSBT pre funding: %s", spew.Sdump(fundingPsbt))
 
+	// Before we proceed, we'll make sure the fee rate we're using is above
+	// the min relay fee.
+	minRelayFee, err := f.cfg.ChainWallet.MinRelayFee(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to establish min_relay_fee: %w",
+			err)
+	}
+	if minRelayFee > fundingState.feeRate.FeePerKWeight() {
```
guggero, line 1068: micro-nit: I would expect the condition to read "if my transaction's fee is _smaller_ than the minimum relay fee, I have a problem". So perhaps we can rewrite it that way?

tapchannel/aux_funding_controller.go
```diff
@@ -1058,6 +1058,18 @@ func (f *FundingController) completeChannelFunding(ctx context.Context,
 
 	log.Debugf("Funding PSBT pre funding: %s", spew.Sdump(fundingPsbt))
 
+	// Before we proceed, we'll make sure the fee rate we're using is above
+	// the min relay fee.
+	minRelayFee, err := f.cfg.ChainWallet.MinRelayFee(ctx)
```
guggero, line 1063: I think we should move this check into `processFundingReq`, directly after the `!supportsAssetChans` check.
Otherwise we kick off a bunch of temporary state and p2p communication, just to find out something we already knew earlier.

guggero (commented):
Very nice and clean! One request and a nit, but we're super close!

guggero (approved):
Very nice, LGTM :tada:

jharveyb (approved):
Nice to remove all these checks - LGTM!

I think we should have a follow-up PR to add similar logic to the minter, I know there have been issues there also.

I think this is the right spot, or at least related:

https://github.com/lightninglabs/taproot-assets/blob/420f2467a6770a9c6042159dcd4c3ce194db25c2/tapgarden/planter.go#L552

Issue #880, opened 2024-04-13
[bug]: `universe_roots` cache-hit ratio, gross underperformance


dstadulis:
### Background
This issue relates to the cache performance of `universe_roots` with respect to interacting with the DB cache. Based on the logs from Lab's universe, the **hit-to-misses ratio for this cache is exhibiting under performance. Remediating the issue will ensure performance and efficiency of universe functionality.**

>│ 2024-04-01 23:32:09.632 [INF] TADB: db cache universe_roots: **260271 hits, 2225329 misses, 10.47% hit ratio** 

### Steps to reproduce [edit at will]

1. Initialize a universe
2. Let the `universe_roots` cache be populated with roots
3. Observe the cache usage over a period of time (e.g., minutes)
4. Check the cache hit-to-misses ratio in the logs or monitoring dashboard

### Expected result:
The hit-to-misses ratio should be higher than [?80%]

### Actual result:
The cache hit-to-misses ratio is insufficient -- at ~10%.
Ramification question: Do the missed hits cause additional/supplemental/superfluous requests?

### Possible solutions:

1. Analyze the current cache configuration to identify any optimization opportunities.
2. Investigate if inefficiencies in the caching logic that need to be fixed.

----
Please provide feedback or suggestions for further improvement after reviewing this issue.

dstadulis:
to be answered:
Are the elements that are being missed, actually being cached?

jharveyb:
From offline discussion, I think cache was just too small. Worst-case size is 2x # of asset IDs / groups since there are issuance and transfer roots.

jharveyb:
This was also slow to populate for me locally, which caused a timeout on the CLI command `universe roots`.

Running 0.4.0 rc3, with the old "full sync" mode (I think?) connected to the testnet LL universe server:

```bash
2024-07-10 13:29:40.121 [INF] TADB: populating root cache...
2024-07-10 13:35:45.524 [INF] TADB: populating root cache...
2024-07-10 13:37:20.565 [DBG] TADB: Populating 512 root nodes into cache, took=1m35.040313762s
```

I think the cache gets dumped repeatedly during the process of sync for being too small?

Even after a successful `universe roots` call, future calls claim to be repopulated the root cache and take the same amount of time.

jharveyb:
Doubling the page size `universe.MaxPageSize` helped a lot; not sure what % of the data returned is coming out of cache now though.

Roasbeef:
[...]

PR #1169, opened 2024-11-04
[multiverse RPC]: add better universe root node cache

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/880.

Opening as draft, mainly looking for concept ACK.

@Roasbeef the idea here is the following:
 - Since the universe sync is what needs to be fast, we add a completely new cache for those queries
 - The syncer client always queries with `sort=ascending` and `WithAmountsById=false`, which allows us to optimize the new cache and keep _all_ root nodes in memory for those queries
 - The old/existing page based root node cache will only be used for custom user queries that want descending order or `WithAmountsById=true`, which allows us to keep a much smaller cache
 - All the other caches remain untouched in their functionality (just renamed and made configurable).

TODOs:
 - Test new cache
 - Update comments in RPC proto file that order of results is dependent on query parameters (depending on whether they hit the syncer cache or not).


tapdb/multiverse_cache_test.go
```diff
@@ -0,0 +1,143 @@
...
+	const pageSize = 64
+	roots := queryRoots(t, multiverse, pageSize)
+	require.Len(t, roots, numAssets)
+	assertAllLeavesInRoots(t, allLeaves, roots)
+
+	// We need to round up, since we need to make another query for the
+	// remainder. And the way the cache query is built (query, if not found,
+	// acquire lock, query again), we always make two queries for each page.
+	numMisses := ((numAssets / pageSize) + 1) * 2
+	require.EqualValues(t, 0, multiverse.rootNodeCache.hit.Load())
```
Roasbeef, line 50: Nice usage of the internal stat counter here to test the his/misses!

tapdb/multiverse_cache.go
```diff
@@ -187,6 +189,212 @@ func (r *rootIndex) wipe() {
...
+	// of these universes.
+	universeKeyList []universe.IdentifierKey
+
+	// universeRoots is a map of universe ID key to the root of the
+	// universe. This map is needed to look up the root of an individual
+	// universe quickly, or to look up the roots of all the universes when
+	// paging through them. This map never needs to be cleared, as universe
+	// roots are only added (new issuance or transfer events) or modified
+	// (re-issuance into grouped asset or new transfer) but never removed.
+	// Therefore, when the cache needs to be wiped, we only need to clear
```
Roasbeef, line 272: >  but never removed.

It's not exposed over the RPC interface rn, but they _can_ be removed (see the `DeleteUniverse` method).

tapdb/multiverse_cache.go
```diff
@@ -187,6 +189,212 @@ func (r *rootIndex) wipe() {
...
+
+	*cacheLogger
+}
+
+// newSyncerRootNodeCache creates a new root node cache.
+func newSyncerRootNodeCache() *syncerRootNodeCache {
+	return &syncerRootNodeCache{
+		universeRoots: make(
+			map[universe.IdentifierKey]universe.Root,
+			numCachedProofs,
```
Roasbeef: I think we can pre-size this to a larger size, as rn it's set to 50k, but we have ~175k assets on mainnet rn.

tapdb/multiverse_cache.go
```diff
@@ -187,6 +189,212 @@ func (r *rootIndex) wipe() {
...
+
+	// We shouldn't be called for a query that can't be served from the
+	// cache. But in case we are, we'll just short-cut here.
+	if !isQueryForSyncerCache(q) {
+		return nil
+	}
+
+	// If we have the write lock already, no need to fetch it.
+	if !haveWriteLock {
+		r.RLock()
```
Roasbeef, line 326: This grabs the read lock, but seems the intention is to grab the _write_ lock.

tapdb/multiverse_cache.go
```diff
@@ -187,6 +189,212 @@ func (r *rootIndex) wipe() {
...
+		r.Miss()
+
+		return nil
+	}
+
+	offset := q.Offset
+	limit := q.Limit
+
+	// Is the page valid?
+	if offset < 0 || offset >= int32(len(r.universeRoots)) || limit <= 0 {
```
Roasbeef, line 343: Log this w/ a warning?

tapdb/multiverse_cache.go
```diff
@@ -187,6 +189,212 @@ func (r *rootIndex) wipe() {
...
+//
+// NOTE: This method must be called while holding the syncer cache lock.
+func (r *syncerRootNodeCache) replaceCache(newRoots []universe.Root) {
+	r.universeKeyList = make([]universe.IdentifierKey, len(newRoots))
+	for idx, root := range newRoots {
+		r.universeKeyList[idx] = root.ID.Key()
+		r.universeRoots[root.ID.Key()] = root
+	}
+
+	// To make sue we can easily add and remove entries, we sort the
```
Roasbeef, line 409: sue -> sure

tapdb/multiverse_cache.go
```diff
@@ -187,6 +189,212 @@ func (r *rootIndex) wipe() {
...
+		// never changes.
+		r.universeRoots[root.ID.Key()] = root
+
+		return
+	}
+
+	r.universeKeyList = append(r.universeKeyList, root.ID.Key())
+	r.universeRoots[root.ID.Key()] = root
+
+	// To make sue we can easily add and remove entries, we sort the
```
Roasbeef, line 440: Can pull into a helper method, as used above verbatim.

tapdb/multiverse_cache.go
```diff
@@ -187,6 +189,212 @@ func (r *rootIndex) wipe() {
...
+			r.universeKeyList[i][:], r.universeKeyList[j][:],
+		) < 0
+	})
+}
+
+// remove removes a single root from the cache.
+//
+// NOTE: This method must be called while holding the syncer cache lock.
+func (r *syncerRootNodeCache) remove(key universe.IdentifierKey) {
+	idx := sort.Search(len(r.universeKeyList), func(i int) bool {
```
Roasbeef, line 458: Nice usage of `sort.Search` here 👍

tapdb/multiverse_cache.go
```diff
@@ -187,6 +189,212 @@ func (r *rootIndex) wipe() {
...
+		return bytes.Compare(r.universeKeyList[i][:], key[:]) >= 0
+	})
+	if idx < len(r.universeKeyList) && r.universeKeyList[idx] == key {
+		// Remove the entry from the list.
+		r.universeKeyList = append(
```
Roasbeef, line 463: Can use the `slices` package here: https://pkg.go.dev/slices#Delete

tapdb/multiverse_cache.go
```diff
@@ -13,6 +13,59 @@ import (
 	"github.com/lightningnetwork/lnd/lnutils"
 )
 
+// MultiverseCacheConfig is the configuration for the different multiverse
+// caches that exist.
+//
+//nolint:lll
+type MultiverseCacheConfig struct {
```
Roasbeef, line 20: 👍

tapdb/multiverse_cache.go
```diff
@@ -13,6 +13,59 @@ import (
...
+	// serves all paginated queries for root nodes that use different
+	// parameters than the syncer cache.
+	RootNodePageCacheSize uint64 `long:"root-node-page-cache-size" description:"The size of the root node page cache for all requests that aren't served by the syncer cache."`
+}
+
+// DefaultMultiverseCacheConfig returns the default configuration for the
+// multiverse cache.
+func DefaultMultiverseCacheConfig() MultiverseCacheConfig {
+	return MultiverseCacheConfig{
+		ProofsPerUniverse:        5,
```
Roasbeef, line 60: Temp for this commit? 

We should inherit the defaults defined in the existing package here.

Roasbeef (commented):
Approach ACK!

tapdb/multiverse_cache.go
guggero: No, we're really just reading here. But the comment is a bit misleading, fixed it.

tapdb/multiverse_cache.go
guggero: I think the original value for this was way too high. Because this is the number of proofs we cache **per universe** (and we don't have a limit on the number of universes we cache proofs for, it's a `lnutils.SyncMap` here and not an `lru.Cache`), we should be quite conservative here.

Perhaps we should turn this into an `lru.Cache` too, same as with the `universeLeafPageCache`?

guggero:
Addressed all comments and added proper unit tests. Good for full review now.

tapdb/multiverse_cache_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+		require.Len(t, cache.universeKeyList, numAssets)
+		require.Len(t, cache.universeRoots, numAssets)
+		require.Contains(t, cache.universeKeyList, root.ID.Key())
+		require.Contains(t, cache.universeRoots, root.ID.Key())
+
+		roots := queryRoots(t, multiverse, pageSize)
+		require.Len(t, roots, numAssets)
+		require.Equal(t, originalRoots, roots)
+	}
```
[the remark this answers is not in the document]
GeorgeTsagk: we could at least `fetchRoots` again and verify they got served from the cache

GeorgeTsagk (commented):
looking good!

do we have a way of verifying/testing:
- new resource usage overhead
- performance difference of having a new syncer cache (an itest?)

otherwise ACK, plus some minor Qs

tapdb/multiverse_cache.go
```diff
@@ -0,0 +1,417 @@
...
+	// First, get the sub-cache for this universe ID from the map of
+	// caches.
+	idStr := treeID(id.String())
+	assetProofCache, _ := p.LoadOrStore(idStr, newLeafCache())
+
+	// With that lower level cache obtained, we can check to see if we have
+	// a hit or not.
+	proofKey := NewProofKey(id, leafKey)
+	proofFromCache, err := assetProofCache.Get(proofKey)
+	if err == nil {
```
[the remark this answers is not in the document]
guggero: This is a pure code move commit, don't want to change things.

tapdb/multiverse_cache_test.go
```diff
@@ -0,0 +1,219 @@
...
+	require.EqualValues(t, numHits, multiverse.rootNodeCache.hit.Load())
+	require.EqualValues(
+		t, numMisses, multiverse.rootNodeCache.miss.Load(),
+	)
+
+	// The new syncer cache should only have two misses, one from when the
+	// cache was empty, one from after acquiring the write lock and all
+	// other queries should be hits.
+	require.EqualValues(t, 2, multiverse.syncerCache.miss.Load())
+	require.EqualValues(t, numHits, multiverse.syncerCache.hit.Load())
```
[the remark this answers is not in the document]
guggero: Yeah, you either have something in the cache for a page query or you don't, there's no partial hit.

guggero:
I added a test that calculates the approximate size of the entries in the cache:

```
=== RUN   TestSyncerCacheMemoryUsage
    multiverse_cache_test.go:214: Generated 500 roots in 26.583318ms
=== RUN   TestSyncerCacheMemoryUsage/500_roots
    multiverse_cache_test.go:224: Memory usage for 500 roots: 139496 bytes
    multiverse_cache_test.go:226: Memory usage per root: 278 bytes
    multiverse_cache_test.go:228: Benchmark took 8.836µs
--- PASS: TestSyncerCacheMemoryUsage/500_roots (0.01s)
    multiverse_cache_test.go:214: Generated 5000 roots in 257.823179ms
=== RUN   TestSyncerCacheMemoryUsage/5000_roots
    multiverse_cache_test.go:224: Memory usage for 5000 roots: 1073384 bytes
    multiverse_cache_test.go:226: Memory usage per root: 214 bytes
    multiverse_cache_test.go:228: Benchmark took 104.568µs
--- PASS: TestSyncerCacheMemoryUsage/5000_roots (0.01s)
    multiverse_cache_test.go:214: Generated 170000 roots in 8.847171795s
=== RUN   TestSyncerCacheMemoryUsage/170000_roots
    multiverse_cache_test.go:224: Memory usage for 170000 roots: 34259176 bytes
    multiverse_cache_test.go:226: Memory usage per root: 201 bytes
    multiverse_cache_test.go:228: Benchmark took 1.820929ms
--- PASS: TestSyncerCacheMemoryUsage/170000_roots (0.08s)
--- PASS: TestSyncerCacheMemoryUsage (9.23s)
```

With around 170k roots currently in existence on mainnet, that means the new cache would only use ~34 MBytes of memory.

GeorgeTsagk (approved):
Looks good!

tapdb/multiverse.go
```diff
@@ -443,6 +528,44 @@ func (b *MultiverseStore) queryRootNodes(ctx context.Context,
...
+//
+// NOTE: This method must be called while holding the syncer cache lock.
+func (b *MultiverseStore) fillSyncerCache(ctx context.Context) error {
+	now := time.Now()
+	log.Infof("Populating syncer root cache...")
+
+	params := sqlc.UniverseRootsParams{
+		SortDirection: sqlInt16(universe.SortAscending),
+		NumOffset:     0,
+		NumLimit:      universe.MaxPageSize,
```
Roasbeef, line 576: In cases like this, I wonder if it's better to just read the entire set of roots in one go. So one single query, instead of a bunch of smaller ones. With the current amount of assets on mainnet, we'll end up doing 300 or so queries here to read them all out. Some benchmarking would likely help us find the sweet spot here re page size to read. 

Non-blocking.

tapdb/multiverse_cache.go
```diff
@@ -271,6 +475,7 @@ func (r *rootNodeCache) cacheRoots(q universe.RootNodesQuery,
 
 	rootIndex := r.rootIndex.Load()
 	for _, rootNode := range rootNodes {
+		rootNode := rootNode
```
Roasbeef, line 563: FWIW, with the latest version of Go, we should no longer need to do this reassignment trick.

tapdb/multiverse_cache.go
Roasbeef: Ah ok, the original intent was for that to be a instance wide value (eg: only cache 50k proofs total). 

I think before merge, we should deploy a version of this on testnet/mainnet to observe the cache hit/miss rate.

Roasbeef (approved):
LGTM 🦠

Issue #1164, opened 2024-10-30
[bug/feature]: Implement PrevId[] UTXO selection for FundVirtualPSBT RPC method 


habibitcoin:
### Background

When operating a pocket universe, an operator may want to assist users in funding a PSBT to fulfill a Taproot Asset address transfer. In the case of Tajfi, we do this by taking the user's public key and then searching for UTXOs we control that have a script key matching the user, and then including an array of user-controlled-UTXOs that match, or exceed, the decoded amount of the Taproot Asset address invoice to fulfill.

Currently, we cannot provide the PrevId because the RPC method is not yet implemented, but it does seem that future support was already planned: https://github.com/lightninglabs/taproot-assets/blob/main/rpcserver.go#L2065
```
case req.GetRaw() != nil:
		raw := req.GetRaw()
		if len(raw.Inputs) > 0 {
			return nil, fmt.Errorf("template inputs not yet " +
				"supported")
		}
```

So I'd say this is kinda inbetween a bug and feature request. I think this would probably be a great beginner user task, and I'll probably take a stab at it soon.

### Your environment

Using tapd v0.4.1. Although probably not relevant since as mentioned above, this behavior/error path seems to be expected regardless of path.

### Steps to reproduce
This should be reproducible by sending a valid RPC payload for FundVirtualPSBT. An example of this payload for the HTTP request is provided below:

```
{
  "raw": {
    "recipients": {
      "tapabcxyz": 0 
    },
    "inputs": [
      {
        "outpoint": {
          "txid": "da742200b442ab7d6b8d5e39c2b1203f74b1ff3fcb5237d0418771b960e5243f",
          "output_index": 1
        },
        "id": "0a17de7503cc844f3a4fbf964af818266cd5f8369b2927971e8bafd62d897fdf",
        "script_key": "02d4d18c8f36edad6f3fba2407ac4a8b5d9416071febac4016adeeff8dd862e83a"
      }
    ]
  }
}
```

Removing the inputs object will allow the call to continue, but can result in a problem where the UTXOs selected might not have the desired script_key the pocket universe user intends on using to fulfill their transfer.

### Expected behavior

The daemon should attempt to only fund the virtual PSBT using the inputs provided in the raw payload.

habibitcoin:
Gave a stab at this in https://github.com/lightninglabs/taproot-assets/pull/1172 and have something that has at least unblocked our internal testing.

guggero:
Fixed by #1172.

PR #1172, opened 2024-11-04
Implement PrevId[] UTXO selection for FundVirtualPSBT RPC method; valid for single PrevId at first

habibitcoin:
Related to https://github.com/lightninglabs/taproot-assets/issues/1164

This PR simply implements the plumbing required to utilize the PrevIds payload for FundVirtualPSBT. Similar to how we can only handle 1 address for a raw input, this PR also only implements support for a single PrevId in the PrevIds payload.

At a high level, we perform the following:

- Update FundAddressSend() to take in an optional array of PrevID pointers
- If we see the PrevID pointers array is provided, and the length is 1, we add these values to the CommitmentConstraints object. CommitmentConstraints is also updated as a struct to hold an array of PrevIDs
- We update `constraintsToDbFilter` to utilize the prevID if the length is 1:
```
if len(query.PrevIDs) == 1 {
			prevID := query.PrevIDs[0]
			encodedOutpoint, err := encodeOutpoint(prevID.OutPoint)
			if err == nil {
				assetFilter.AnchorPoint = encodedOutpoint
				assetFilter.TweakedScriptKey = prevID.ScriptKey.CopyBytes()
			}
			// If there's an error, we simply skip setting the fields
			// TODO, maybe log error or return it?
		}
```
- Return an error to the RPC if an attempt is made to send more than 1 input

To-dos:
- [x] For some reason, my outpoint is not being encoded properly, but during testing if I simply comment out `assetFilter.AnchorPoint = encodedOutpoint`, everything else works as expected and is serving our needs in our demo environment
- [x] itest

### DIFF
--- a/itest/assets_test.go
+++ b/itest/assets_test.go
@@ -627,16 +627,27 @@ func testMintBatchAndTransfer(t *harnessTest) {
 	require.True(t.t, proto.Equal(originalBatch, afterBatch))
 }
 
-// testAssetBalances tests the balance retrieval functionality for issued
-// assets. The function mints two batches of assets and asserts if the tapcli
-// `assets balance` returns the correct balances. It then funds a vPSBT, putting
-// a lease on one of the two batches. It then asserts whether the endpoint still
-// returns the correct balances, taking into account the `include_leased` flag.
+// testAssetBalances validates the balance retrieval and virtual PSBT funding
+// functionality for issued assets. The test performs the following steps:
+//  1. Mints two batches of assets and verifies that the tapcli
+//     `assets balance` returns the correct balances.
+//  2. Tests funding a vPSBT, putting a lease on one of the two batches, with
+//     the `FundVirtualPsbt` RPC for various scenarios:
+//     - Fails if the Inputs field contains a nil Outpoint.
+//     - Fails if the Inputs field contains an invalid Outpoint.
+//     - Fails if the Inputs field contains an invalid short AssetId.
+//     - Fails if the Inputs field contains an invalid short ScriptKey.
+//     - Succeeds if a valid Outpoint is provided in the Inputs field.
+//     - Succeeds if the Inputs field is not provided at all, using asset
+//     coin selection instead.
+//  3. Ensures that leased assets are reflected correctly in the balance
+//     retrieval, with and without the `include_leased` flag.
 func testAssetBalances(t *harnessTest) {
 	ctxb := context.Background()
 	ctxt, cancel := context.WithTimeout(ctxb, defaultWaitTimeout)
 	defer cancel()
 
+	// Mint assets for testing.
 	rpcSimpleAssets := MintAssetsConfirmBatch(
 		t.t, t.lndHarness.Miner().Client, t.tapd, simpleAssets,
 	)
@@ -652,8 +663,18 @@ func testAssetBalances(t *harnessTest) {
 		t.t, t.tapd, rpcSimpleAssets, rpcIssuableAssets, false,
 	)
 
+	// Check chain anchor for the target asset.
+	out, err := wire.NewOutPointFromString(
+		targetAsset.ChainAnchor.AnchorOutpoint,
+	)
+	require.NoError(t.t, err)
+
 	var (
 		targetAssetGenesis = targetAsset.AssetGenesis
+		assetId            = targetAssetGenesis.AssetId
+		scriptKey          = targetAsset.ScriptKey
+		anchorTxid         = out.Hash.CloneBytes()
+		anchorVout         = out.Index
 		aliceTapd          = t.tapd
 		bobLnd             = t.lndHarness.Bob
 	)
@@ -672,31 +693,154 @@ func testAssetBalances(t *harnessTest) {
 	})
 	require.NoError(t.t, err)
 
-	// Now we can create our virtual transaction and ask Alice's tapd to
-	// fund it.
 	recipients := map[string]uint64{
 		bobAddr.Encoded: bobAddr.Amount,
 	}
-	_, err = aliceTapd.FundVirtualPsbt(
-		ctxt, &wrpc.FundVirtualPsbtRequest{
-			Template: &wrpc.FundVirtualPsbtRequest_Raw{
-				Raw: &wrpc.TxTemplate{
-					Recipients: recipients,
+
+	subTests := []struct {
+		name           string
+		inputs         []*wrpc.PrevId
+		expectError    bool
+		expectedErrMsg string
+	}{
+		{
+			name: "Fail if Inputs are provided but Outpoint is nil",
+			inputs: []*wrpc.PrevId{
+				{
+					Outpoint:  nil,
+					Id:        assetId,
+					ScriptKey: scriptKey,
 				},
 			},
+			expectError:    true,
+			expectedErrMsg: "index 0 has a nil Outpoint",
 		},
-	)
-	require.NoError(t.t, err)
+		{
+			name: "Fail if Inputs contain an invalid Outpoint",
+			inputs: []*wrpc.PrevId{
+				{
+					Outpoint: &taprpc.OutPoint{
+						Txid: []byte(
+							"invalid_txid",
+						),
+						OutputIndex: anchorVout,
+					},
+					Id:        assetId,
+					ScriptKey: scriptKey,
+				},
+			},
+			expectError:    true,
+			expectedErrMsg: "invalid Txid",
+		},
+		{
+			name: "Fail if AssetId is too short",
+			inputs: []*wrpc.PrevId{
+				{
+					Outpoint: &taprpc.OutPoint{
+						Txid:        anchorTxid,
+						OutputIndex: anchorVout,
+					},
+					Id:        []byte{1, 2},
+					ScriptKey: scriptKey,
+				},
+			},
+			expectError:    true,
+			expectedErrMsg: "invalid asset ID",
+		},
+		{
+			name: "Fail if ScriptKey is too short",
+			inputs: []*wrpc.PrevId{
+				{
+					Outpoint: &taprpc.OutPoint{
+						Txid:        anchorTxid,
+						OutputIndex: anchorVout,
+					},
+					Id:        assetId,
+					ScriptKey: []byte{1, 2},
+				},
+			},
+			expectError:    true,
+			expectedErrMsg: "invalid script key",
+		},
+		{
+			name: "Fail if valid, unfindable Outpoint is provided",
+			inputs: []*wrpc.PrevId{
+				{
+					Outpoint: &taprpc.OutPoint{
+						Txid:        anchorTxid,
+						OutputIndex: anchorVout + 1,
+					},
+					Id:        assetId,
+					ScriptKey: scriptKey,
+				},
+			},
+			expectError:    true,
+			expectedErrMsg: "failed to find coin(s)",
+		},
+		{
+			name: "Succeed if a valid Outpoint is provided",
+			inputs: []*wrpc.PrevId{
+				{
+					Outpoint: &taprpc.OutPoint{
+						Txid:        anchorTxid,
+						OutputIndex: anchorVout,
+					},
+					Id:        assetId,
+					ScriptKey: scriptKey,
+				},
+			},
+			expectError: false,
+		},
+		{
+			name:        "Succeed if no Inputs are provided",
+			inputs:      nil,
+			expectError: false,
+		},
+	}
 
-	// With a transaction funding should have led to a lease on the simple
-	// assets, we'll use the balance calls to ensure that we're able to
-	// retrieve the proper balances.
-	rpcEmptyAssets := []*taprpc.Asset{}
-	AssertAssetBalances(
-		t.t, t.tapd, rpcEmptyAssets, rpcIssuableAssets, false,
-	)
+	for _, tt := range subTests {
+		// Now we can create our virtual transaction and ask Alice's
+		// tapd to fund it.
+		_, err := aliceTapd.FundVirtualPsbt(
+			ctxt, &wrpc.FundVirtualPsbtRequest{
+				Template: &wrpc.FundVirtualPsbtRequest_Raw{
+					Raw: &wrpc.TxTemplate{
+						Recipients: recipients,
+						Inputs:     tt.inputs,
+					},
+				},
+			},
+		)
 
-	AssertAssetBalances(
-		t.t, t.tapd, rpcSimpleAssets, rpcIssuableAssets, true,
-	)
+		if tt.expectError {
+			require.ErrorContains(t.t, err, tt.expectedErrMsg)
+		} else {
+			require.NoError(t.t, err)
+
+			// With a transaction funding should have led to a
+			// lease on the simple assets, we'll use the balance
+			// calls to ensure that we're able to retrieve the
+			// proper balances.
+			rpcEmptyAssets := []*taprpc.Asset{}
+			AssertAssetBalances(
+				t.t, t.tapd, rpcEmptyAssets, rpcIssuableAssets,
+				false,
+			)
+			AssertAssetBalances(
+				t.t, t.tapd, rpcSimpleAssets, rpcIssuableAssets,
+				true,
+			)
+
+			// Unlock the input if provided so it can be reused.
+			for _, input := range tt.inputs {
+				_, err = aliceTapd.RemoveUTXOLease(
+					ctxb,
+					&wrpc.RemoveUTXOLeaseRequest{
+						Outpoint: input.Outpoint,
+					},
+				)
+				require.NoError(t.t, err)
+			}
+		}
+	}
 }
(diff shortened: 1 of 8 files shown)


tapfreighter/interface.go
```diff
@@ -35,6 +35,9 @@ type CommitmentConstraints struct {
 	// to satisfy the constraints.
 	MinAmt uint64
 
+	// PrevIDs are the set of inputs allowed to be used
+	PrevIDs []*asset.PrevID
```
guggero, line 39: These shouldn't be pointers, but instead just `[]asset.PrevID`.

tapdb/assets_store.go
```diff
@@ -883,6 +883,18 @@ func (a *AssetStore) constraintsToDbFilter(
 		assetFilter.AssetIDFilter = assetIDBytes
 		assetFilter.KeyGroupFilter = groupKeyBytes
 
+		// Check if there's exactly one PrevID and set the filters accordingly.
+		if len(query.PrevIDs) == 1 {
```
guggero, line 887: With this alternative approach we wouldn't need to restrict this to a single ID:

```go
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -883,18 +883,6 @@ func (a *AssetStore) constraintsToDbFilter(
                assetFilter.AssetIDFilter = assetIDBytes
                assetFilter.KeyGroupFilter = groupKeyBytes
 
-               // Check if there's exactly one PrevID and set the filters accordingly.
-               if len(query.PrevIDs) == 1 {
-                       prevID := query.PrevIDs[0]
-                       encodedOutpoint, err := encodeOutpoint(prevID.OutPoint)
-                       if err == nil {
-                               assetFilter.AnchorPoint = encodedOutpoint
-                               assetFilter.TweakedScriptKey = prevID.ScriptKey.CopyBytes()
-                       }
-                       // If there's an error, we simply skip setting the fields
-                       // TODO, maybe log error or return it?
-               }
-
                // TODO(roasbeef): only want to allow asset ID or other and not
                // both?
 
@@ -2003,7 +1991,48 @@ func (a *AssetStore) ListEligibleCoins(ctx context.Context,
        // have a block height of 0, so we set the minimum block height to 1.
        assetFilter.MinAnchorHeight = sqlInt32(1)
 
-       return a.queryCommitments(ctx, assetFilter)
+       selectedCommitments, err := a.queryCommitments(ctx, assetFilter)
+       if err != nil {
+               return nil, fmt.Errorf("unable to query commitments: %w", err)
+       }
+
+       // If we want to restrict on specific inputs, we do the filtering now.
+       if len(constraints.PrevIDs) > 0 {
+               selectedCommitments = filterCommitmentsByPrevIDs(
+                       selectedCommitments, constraints.PrevIDs,
+               )
+
+               // If this results in an empty list, we return the same error we
+               // would if there were no coins found without the filter.
+               if len(selectedCommitments) == 0 {
+                       return nil, tapfreighter.ErrMatchingAssetsNotFound
+               }
+       }
+
+       return selectedCommitments, nil
+}
+
+// filterCommitmentsByPrevIDs filters the given commitments by the previous IDs
+// given.
+func filterCommitmentsByPrevIDs(commitments []*tapfreighter.AnchoredCommitment,
+       prevIDs []*asset.PrevID) []*tapfreighter.AnchoredCommitment {
+
+       prevIDMatches := func(p *asset.PrevID,
+               c *tapfreighter.AnchoredCommitment) bool {
+
+               return p.OutPoint == c.AnchorPoint && p.ID == c.Asset.ID() &&
+                       p.ScriptKey == asset.ToSerialized(
+                               c.Asset.ScriptKey.PubKey,
+                       )
+       }
+
+       commitmentInList := func(c *tapfreighter.AnchoredCommitment) bool {
+               return fn.Any(prevIDs, func(p *asset.PrevID) bool {
+                       return prevIDMatches(p, c)
+               })
+       }
+
+       return fn.Filter(commitments, commitmentInList)
 }
```

guggero (commented):
Thanks for the PR! This is definitely something we want as well and just didn't get to yet.
See inline comment for alternative and slightly more generic approach.

tapdb/assets_store.go
habibitcoin: love it! Great idea, will implement. thanks for the feedback

habibitcoin:
@guggero addressed! Any thoughts on modifying or adding a new itest? I was thinking maybe to modify
```
// testPsbtExternalCommit tests the ability to fully customize the BTC level of
// an asset transfer using a PSBT. This exercises the CommitVirtualPsbts and
// PublishAndLogTransfer RPCs. The test case moves some assets into an output
// that has a hash lock tapscript.
func testPsbtExternalCommit(t *harnessTest) {
```

Or copy + update the scenario to include an array of 2 inputs perhaps

rpcserver.go
```diff
@@ -2049,9 +2049,31 @@ func (r *rpcServer) FundVirtualPsbt(ctx context.Context,
 
 	case req.GetRaw() != nil:
 		raw := req.GetRaw()
+		prevIDs := []asset.PrevID{}
 		if len(raw.Inputs) > 0 {
-			return nil, fmt.Errorf("template inputs not yet " +
-				"supported")
+			for _, input := range raw.Inputs {
```
guggero, line 2054: Can remove a whole level of indentation by removing the `if len(raw.Inputs) > 0 {` since `for ... range` loops are `nil` safe.

rpcserver.go
```diff
@@ -2049,9 +2049,31 @@ func (r *rpcServer) FundVirtualPsbt(ctx context.Context,
...
+				// Create a new chainhash.Hash and set its bytes
```
guggero, line 2055: nit: comments should always be full sentences and end with punctuation.

rpcserver.go
```diff
@@ -2049,9 +2049,31 @@ func (r *rpcServer) FundVirtualPsbt(ctx context.Context,
...
+				var hash chainhash.Hash
+				err := hash.SetBytes(input.Outpoint.Txid)
```
guggero, line 2057: nit: Should use `chainhash.NewHash()` instead.

rpcserver.go
```diff
@@ -2049,9 +2049,31 @@ func (r *rpcServer) FundVirtualPsbt(ctx context.Context,
...
+				if err != nil {
+					return nil, fmt.Errorf(
+						"invalid Txid length: %w", err,
+					)
+				}
+				// Decode the input into an asset.PrevID
```
guggero, line 2063: nit: if a comment isn't at the beginning of a new block/indentation, we usually add an empty line before to give some breathing space.

rpcserver.go
```diff
@@ -2049,9 +2049,31 @@ func (r *rpcServer) FundVirtualPsbt(ctx context.Context,
...
+		prevIDs := []asset.PrevID{}
 		if len(raw.Inputs) > 0 {
-			return nil, fmt.Errorf("template inputs not yet " +
-				"supported")
+			for _, input := range raw.Inputs {
+				// Create a new chainhash.Hash and set its bytes
+				var hash chainhash.Hash
+				err := hash.SetBytes(input.Outpoint.Txid)
+				if err != nil {
+					return nil, fmt.Errorf(
```
guggero, line 2059: nit: Multi-line calls to formatting function should follow the different form to more easily distinguish them from non-error/non-logging calls. See https://github.com/lightningnetwork/lnd/blob/master/docs/code_formatting_rules.md#exception-for-log-and-error-message-formatting.

tapfreighter/wallet.go
```diff
@@ -69,6 +69,7 @@ type Wallet interface {
 	// selected assets.
 	FundAddressSend(ctx context.Context,
 		coinSelectType tapsend.CoinSelectType,
+		prevIDs []asset.PrevID,
```
guggero, line 72: nit: seems to fit on previous line, here and below.

tapfreighter/interface.go
```diff
@@ -35,6 +35,9 @@ type CommitmentConstraints struct {
 	// to satisfy the constraints.
 	MinAmt uint64
 
+	// PrevIDs are the set of inputs allowed to be used
```
guggero, line 38: nit: missing full stop.

rpcserver.go
guggero, line 2057: We should probably check that `input.Outpoint` is not `nil`.

rpcserver.go
```diff
@@ -2049,9 +2049,31 @@ func (r *rpcServer) FundVirtualPsbt(ctx context.Context,
 
 	case req.GetRaw() != nil:
 		raw := req.GetRaw()
+		prevIDs := []asset.PrevID{}
```
guggero, line 2052: nit: `var prevIDs []asset.PrevID` style is preferred over empty slice initialization.

guggero (commented):
Thanks a lot for the updates!
Mostly nits remaining. Would be nice to perhaps add an integration test? Do you have time to do that?

habibitcoin:
@guggero I decided to modify the testAssetBalances itest, as it seemed to be a suitable candidate to test this functionality.

I am having trouble however testing both the FundVirtualPSBT RPC with a valid PrevId, release the input lease, and then trying again without any PrevId specified, confirming we succeed

I was trying to use tapgarden.NewMockWalletAnchor(), but that doesn't seem to work. I was also seeing if maybe I could create a taprootassets.NewLndRpcWalletAnchor(), but wasn't sure what to use as the input. Curious if you had any recommendations there? I think once that is fixed, all the tests will pass and this should be ready to ship!

```

	// 4. Succeed if no Inputs are provided.
	// Unlock the input so that it can be spent again.
	walletAnchor := tapgarden.NewMockWalletAnchor()
	err = walletAnchor.UnlockInput(ctxb, *out)
	require.NoError(t.t, err)

	_, err = aliceTapd.FundVirtualPsbt(
		ctxt, &wrpc.FundVirtualPsbtRequest{
			Template: &wrpc.FundVirtualPsbtRequest_Raw{
				Raw: &wrpc.TxTemplate{
					Recipients: recipients,
				},
			},
		},
	)
	require.NoError(t.t, err)
	
	```

**Comment by guggero** [comment]
Yeah, that sounds good as an integration test.

You can't use a mock in itests, as that will literally not do anything. What you'll want to call into instead is [`assetwalletrpc.RemoveUTXOLease`](https://lightning.engineering/api-docs/api/taproot-assets/asset-wallet/remove-u-t-x-o-lease/) to release the asset-level input lease.
It's possible that any BTC on-chain funds added to pay for the fees then also remain leased. Those you need to release on the `lnd` RPC with [`walletrpc.ReleaseOutput`](https://lightning.engineering/api-docs/api/lnd/wallet-kit/release-output/).
Or you just make sure the `lnd` that's backing Alice has enough on-chain UTXOs so it doesn't become an issue in the first place.

File: rpcserver.go [hunk 11]
```diff
@@ -2048,10 +2048,36 @@ func (r *rpcServer) FundVirtualPsbt(ctx context.Context,
 		}
 
 	case req.GetRaw() != nil:
-		raw := req.GetRaw()
-		if len(raw.Inputs) > 0 {
-			return nil, fmt.Errorf("template inputs not yet " +
-				"supported")
+		var (
+			raw     = req.GetRaw()
+			prevIDs = []asset.PrevID{}
```
guggero, line 2053: nit: to avoid allocating a zero-length slice, can just be `prevIDs []asset.PrevID` (so without the equal sign, just a variable declaration that ends up being `nil`, which is fine for `append()`).

itest/assets_test.go
```diff
@@ -673,6 +683,86 @@ func testAssetBalances(t *harnessTest) {
...
+					Recipients: recipients,
+					Inputs: []*wrpc.PrevId{
+						emptyOutpointPrevId,
+					},
+				},
+			},
+		},
+	)
+	require.Error(t.t, err)
+	require.Contains(t.t, err.Error(), "index 0 has a nil Outpoint")
```
guggero, line 710: nit: can combine this and `require.Error()` into a single call to `require.ErrorContains()`.

itest/assets_test.go
```diff
@@ -673,6 +683,86 @@ func testAssetBalances(t *harnessTest) {
...
+
+	// 2. Fail if Inputs contain an invalid Outpoint.
+	invalidTxidPrevId := &wrpc.PrevId{
```
guggero, line 713: Could convert all of this into a table driven test (and then add some test cases for too short asset ID and script key):

```
subTests := []struct {
		name           string
		inputs         []*wrpc.PrevId
		expectError    bool
		expectedErrMsg string
	}{
		{
			name: "Fail if Inputs are provided but Outpoint is nil",
			inputs: []*wrpc.PrevId{
				{
					Outpoint:  nil,
					Id:        assetId,
					ScriptKey: scriptKey,
				},
			},
			expectError:    true,
			expectedErrMsg: "index 0 has a nil Outpoint",
		},
		{
			name: "Fail if Inputs contain an invalid Outpoint",
			inputs: []*wrpc.PrevId{
				{
					Outpoint: &taprpc.OutPoint{
						Txid: []byte(
							"invalid_txid",
						),
						OutputIndex: 0,
					},
					Id:        assetId,
					ScriptKey: scriptKey,
				},
			},
			expectError:    true,
			expectedErrMsg: "invalid Txid length",
		},
		{
			name: "Succeed if a valid Outpoint is provided",
			inputs: []*wrpc.PrevId{
				{
					Outpoint: &taprpc.OutPoint{
						Txid:        out.Hash[:],
						OutputIndex: out.Index,
					},
					Id:        assetId,
					ScriptKey: scriptKey,
				},
			},
			expectError: false,
		},
	}

	for _, tt := range subTests {
		_, err := aliceTapd.FundVirtualPsbt(
			ctxt, &wrpc.FundVirtualPsbtRequest{
				Template: &wrpc.FundVirtualPsbtRequest_Raw{
					Raw: &wrpc.TxTemplate{
						Recipients: recipients,
						Inputs:     tt.inputs,
					},
				},
			},
		)

		if tt.expectError {
			require.ErrorContains(t.t, err, tt.expectedErrMsg)
		} else {
			require.NoError(t.t, err)

			// Unlock the input so that it can be spent again.
			for _, input := range tt.inputs {
				_, err = aliceTapd.RemoveUTXOLease(
					ctxb, &wrpc.RemoveUTXOLeaseRequest{
						Outpoint: input.Outpoint,
					},
				)
				require.NoError(t.t, err)
			}
		}
	}
```

guggero (commented):
Thanks a lot for the update and test!
Looks very close, just a couple of suggestions and fixes.

guggero (approved):
Thanks a lot, LGTM :tada:

habibitcoin:
@dstadulis just wanted to bump this before it gets stale

itest/assets_test.go
```diff
@@ -668,31 +689,139 @@ func testAssetBalances(t *harnessTest) {
...
+					},
+					Id:        assetId,
+					ScriptKey: []byte{1, 2},
+				},
+			},
+			expectError:    true,
+			expectedErrMsg: "invalid script key",
+		},
+		{
+			name: "Succeed if a valid Outpoint is provided",
```
gijswijs, line 762: Would add a test case here that fails if a valid outpoint isn't found.

```
diff --git a/itest/assets_test.go b/itest/assets_test.go
index 00633c65..944f01d0 100644
--- a/itest/assets_test.go
+++ b/itest/assets_test.go
@@ -758,6 +758,21 @@ func testAssetBalances(t *harnessTest) {
                        expectError:    true,
                        expectedErrMsg: "invalid script key",
                },
+               {
+                       name: "Fail if valid, unfindable Outpoint is provided",
+                       inputs: []*wrpc.PrevId{
+                               {
+                                       Outpoint: &taprpc.OutPoint{
+                                               Txid:        anchorTxid,
+                                               OutputIndex: anchorVout + 1,
+                                       },
+                                       Id:        assetId,
+                                       ScriptKey: scriptKey,
+                               },
+                       },
+                       expectError:    true,
+                       expectedErrMsg: "failed to find coin(s)",
+               },
                {
                        name: "Succeed if a valid Outpoint is provided",
                        inputs: []*wrpc.PrevId{
```

gijswijs (changes requested):
Great work! I made some comments and a clarifying question and a single change request. Very nice to see this addition.

This also needs a rebase, before we can merge.

tapdb/assets_store.go
```diff
@@ -1991,7 +1991,48 @@ func (a *AssetStore) ListEligibleCoins(ctx context.Context,
 	// have a block height of 0, so we set the minimum block height to 1.
 	assetFilter.MinAnchorHeight = sqlInt32(1)
 
-	return a.queryCommitments(ctx, assetFilter)
+	selectedCommitments, err := a.queryCommitments(ctx, assetFilter)
+	if err != nil {
+		return nil, fmt.Errorf("unable to query commitments: %w", err)
+	}
+
+	// If we want to restrict on specific inputs, we do the filtering now.
```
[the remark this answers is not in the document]
habibitcoin: Exactly yep!

habibitcoin:
@gijswijs thank you for the review and appreciate your patience - this should be in good shape now!

gijswijs (approved):
LGTM :tada:

Issue #1094, opened 2024-08-20
universe - [feature]: finish implementation of on chain universe commitments


Roasbeef:
**Is your feature request related to a problem? Please describe.**

Today we don't have a commitment maintained by an issuer on chain that can be used to authenticate new issuance, [as outlined in the BIP](https://github.com/Roasbeef/bips/blob/bip-tap-pr/bip-tap-universe.mediawiki#:~:text=genesis%20outpoint%20spent.-,Canonical%20Issuance%20Universe%20State%20Transition%20Rules,-In%20order%20to).  

**Describe the solution you'd like**

Revamp and review the section of the BIP listed above. 

In addition, we want to expand the scope of the chain commitment to include 3 commitemtns:
  * issued 
  * burned
  * ignore 

Issued is the root of the normal issuance universe tree we maintain today. 

Burned is a new root that stores any relevant burning transaction as state transition proofs. 

Ignore is another new root for outputs that should be ignored, meaning if an asset being verified has the specified asset outpoint/scriptkey in it's history, then it should be ignored. 

Some very early interface definitions and structs can be found here which should prove to be useful: https://github.com/lightninglabs/taproot-assets/blob/420f2467a6770a9c6042159dcd4c3ce194db25c2/universe/interface.go#L661-L696. 

In particular, we'll need to decide on the finer details of the "canonical issuance" scheme, which as proposed, would reuse the `asset_group_key` as an on-chain taproot public key, in order to tie the off-chain asset group key to an on-chain UTXO. 

With this, the total amount outstanding is: `issued - burned - ignore`. 

**Additional context**

Such a feature can be used to authenticate any supply level information, useful for transparency pages involving bridges.

dstadulis:
@ffranr this is the issue related to replicating "gettxoutsetinfo" type behavior in tapd

jharveyb:
Writing back offline discussion:

The design as outlined in the BIP linked above has an issue with a circular dependency. Tl;dr:

- The proofs for any issuance event can only be created once the anchor TX for the issuance is confirmed.
- The canonical universe commitment should have the proofs for issuance events as leaves in an MS-SMT.
- Therefore, any commitment that includes an issuance event as a leaf must be anchored _after_ the issuance event.
- This also means that the commitment update must happen in a TX separate from the issuance itself.

This presents a new question of how to link issuance events to the chain of canonical universe commitments.

[...]

PR #1173, opened 2024-11-05
mintrpc+tapgarden: add universe announcement flag to asset minting

ffranr:
This PR adds a new flag to the `MintAsset` RPC request message. When set to true, this flag enables the universe announcement feature for the minting batch. To comply with this feature, the minting batch must be restricted to assets that share the same group key.

---

This PR is part of the work towards closing: https://github.com/lightninglabs/taproot-assets/issues/1094


rpcserver.go
```diff
@@ -574,7 +597,7 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 		}
 	}
 
-	if specificGroupInternalKey {
+	if req.Asset.GroupInternalKey != nil {
```
jharveyb, line 600: This doesn't have the same meaning as before. `specificGroupInternalKey` was `true` if the seedling is being minted into an existing group.

`GroupInternalKey` is only non-nil if you are creating a _new_ group, _and_ you want to use a specific internal key instead of whichever key `tapd` will generate automatically.

tapgarden/batch.go
```diff
@@ -309,6 +315,35 @@ func (m *MintingBatch) HasSeedlings() bool {
 	return len(m.Seedlings) != 0
 }
 
+// ValidateSeedling checks if a seedling is valid for the batch.
+func (m *MintingBatch) ValidateSeedling(newSeedling Seedling) error {
```
jharveyb, line 319: IIUC we want this to be a logical AND between this flag for each seedling, and the matching flag for the batch.

So the first seedling of the batch should update the batch flag, and every other seedling should match. This implies that every seedling would have `EnableUniAnnounce` set to `true`.

tapgarden/batch.go
```diff
@@ -309,6 +315,35 @@ func (m *MintingBatch) HasSeedlings() bool {
...
+	// Ensure that the seedling and batch agree on the enabled universe
+	// announcements.
+	if m.EnableUniAnnounce != newSeedling.EnableUniAnnounce {
+		return fmt.Errorf("batch and seedling do not agree on " +
+			"enabled universe announcements")
+	}
+
+	// If the seedling supports universe announcements, it must have a group
+	// anchor or the same group key as all the other seedlings in the batch.
+	if newSeedling.EnableUniAnnounce {
```
jharveyb, line 329: 👌🏽 

This constraint could be loosened in the future but it seems reasonable for this initial version, assuming that it doesn't affect the proof format or on-chain TX pattern.

tapgarden/planter.go
```diff
@@ -1978,12 +1985,26 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
...
+		// batch will still be valid after adding the seedling.
+		err := c.pendingBatch.ValidateSeedling(*req)
+		if err != nil {
+			return fmt.Errorf("seedling can not be added to "+
+				"pending batch: %w", err)
+		}
+
 		c.pendingBatch.Seedlings[req.AssetName] = req
 
+		// The batch may already exist, but it may not have any
```
jharveyb, line 1999: Could mention that this would happen if a batch got funded before having seedlings added.

tapgarden/planter_test.go
```diff
@@ -1654,6 +1654,12 @@ func testFundSealBeforeFinalize(t *mintingTestHarness) {
 	// harness.
 	t.refreshChainPlanter()
 
+	// A pending batch should not exist yet. Therefore, `PendingBatch`
+	// should return nil and no error.
+	batch, err := t.planter.PendingBatch()
```
jharveyb, line 1659: 💯

taprpc/mintrpc/mint.proto
```diff
@@ -215,6 +215,11 @@ message MintAssetRequest {
     possibly printed on the command line in the case of a very large batch.
     */
     bool short_response = 2;
+
+    // Set to true to enable universe announcements for the asset(s) mint event.
+    // This option restricts the minting batch to assets that are part of the
+    // same universe (i.e., share the same group key).
+    bool enable_uni_announce = 3;
```
jharveyb, line 222: Should this just be in the `MintAsset` message instead, alongside all the other Seedling members?

jharveyb (commented):
Looks good overall! Still some bits to figure out about exactly which constraints will apply to the batches here.

tapgarden/batch.go
ffranr: Yes, that's what I'm going for. `EnableUniAnnounce` set to `true` on each seedling if it's true on the batch. They need to correspond basically. We can relax that sort of thing in the future. Let me know if you have thoughts on that sort of approach.

Roasbeef:
I think this is out dated, as we have a new design we went with. Closing for now.

PR #1174, opened 2024-11-05
tapdb: Add Universe indices, optimize SQL queries, `RWMutex` for cache

jbrill:
# Universe Query Performance Optimization and Lock Safety

This PR introduces two key improvements:
1. Strategic composite indices to optimize `tapdb` universe queries
2. Enhanced lock safety for concurrent stats access

## Lock Safety Improvements
```go
// Added finer-grained locking for stats access
statsCacheLogger    *cacheLogger
statsRefresh        *time.Timer
eventsMtx          sync.RWMutex    // Upgraded from sync.Mutex
assetEventsCache   assetEventsCache
eventsCacheLogger  *cacheLogger
syncStatsMtx      sync.RWMutex    // Upgraded from sync.Mutex
syncStatsCache    *atomicSyncStatsCache
syncStatsRefresh  *time.Timer
```

Key changes:
- Upgraded mutexes to RWMutex for better concurrency
- Separated read/write lock paths in QueryAssetStatsPerDay and QuerySyncStats
- Proper lock ordering to prevent deadlocks
- Added atomic cache for sync stats

### Key Changes
```sql
-- Optimizes universe root lookups and filtering
CREATE INDEX idx_universe_roots_composite
ON universe_roots(namespace_root, proof_type, asset_id);

-- Enhances universe leaf queries and joins
CREATE INDEX idx_universe_leaves_composite
ON universe_leaves(leaf_node_namespace, universe_root_id, leaf_node_key);

-- Improves MSSMT node lookups and aggregations 
CREATE INDEX idx_mssmt_nodes_composite 
ON mssmt_nodes(namespace, key, hash_key, sum);
```

#### PostgreSQL 
- Asset Stats Query: 4.58x improvement (205ms → 44.8ms)
  - Query cost reduced from 49.17 to 23.58
  - Eliminated sequence scans
  - Optimized sort operations and hash joins

- Universe Leaves Query: 1.24x improvement (20.9ms → 16.8ms)
  - Query cost reduced from 59.91 to 20.99
  - Replaced sequence scans with efficient index usage

- Root Query: 1.05x improvement (17.7ms → 16.9ms)
  - Maintained efficient lookups with better join patterns

#### SQLite
- Asset Stats Query: 2.02x improvement (55.9ms → 27.7ms)
  - Replaced BLOOM FILTER with covering index
  - More efficient join strategy

- Universe Leaves and Root Queries: Maintain baseline performance
  - Slight overhead (~5%) from index maintenance
  - Already optimized by SQLite's automatic indexing

#### Write Performance
- Single inserts: <5% overhead
- Batch inserts (100+): 10-15% overhead
- Storage impact: ~25% additional space across all indices

#### Query Cost Analysis
PostgreSQL shows significant cost reductions:
- Complex aggregations: 52% reduction in total cost
- Join operations: 65% reduction in materialization overhead
- Sort operations: ~50% reduction in cost

The improvements come from:
- Better join order selection through index availability
- Reduced materialization steps
- Elimination of table scans
- More efficient sort operations through index coverage

### Implementation Considerations
- Indices designed to balance read/write performance
- Column ordering optimized for common query patterns
- Careful consideration of storage overhead vs performance gains
- Support for both PostgreSQL and SQLite query planners

## Testing
- Comprehensive benchmarks across both databases
- Test dataset: 100 assets, 300 leaves per tree
- 100 iterations per query type
- Full query plan analysis for optimization verification

## Next Steps
1. Monitor index usage patterns in production
2. Gather metrics on maintenance overhead
3. Consider database-specific optimizations
4. Evaluate impact on write-heavy workflows
5. Monitor lock contention in production

The PR delivers two key improvements: significantly better query performance through strategic indexing and enhanced concurrency safety through improved locking patterns. Both changes maintain reasonable overhead while providing substantial benefits for production deployments.

### DIFF
--- a/make/testing_flags.mk
+++ b/make/testing_flags.mk
@@ -54,6 +54,11 @@ ifeq ($(dbbackend),postgres)
 DEV_TAGS += test_db_postgres
 endif
 
+# Run universe tests with increased scale for performance testing.
+ifneq ($(long-tests),)
+DEV_TAGS += longtests
+endif
+
 ifneq ($(tags),)
 DEV_TAGS += ${tags}
 endif
--- a/tapdb/sqlc/migrations/000024_universe_optimization_indexes_queries.up.sql
+++ b/tapdb/sqlc/migrations/000024_universe_optimization_indexes_queries.up.sql
@@ -1,0 +1,14 @@
+-- Most impactful for query_asset_stats which currently has highest latency
+-- Supports the common join pattern and filters on proof_type.
+CREATE INDEX IF NOT EXISTS idx_universe_leaves_asset 
+ON universe_leaves(asset_genesis_id, universe_root_id);
+
+-- Helps with the join conditions we frequently see
+-- This is especially useful for query_universe_leaves and improves join efficiency.
+CREATE INDEX IF NOT EXISTS idx_mssmt_nodes_composite 
+ON mssmt_nodes(namespace, key, hash_key, sum);
+
+-- Optimizes the common namespace_root lookups along with proof_type filtering
+-- This helps with fetch_universe_root and roots-related queries.
+CREATE INDEX IF NOT EXISTS idx_universe_roots_composite
+ON universe_roots(namespace_root, proof_type, asset_id);
--- a/tapdb/sqlc/queries/universe.sql
+++ b/tapdb/sqlc/queries/universe.sql
@@ -6,10 +6,10 @@ FROM universe_roots
 JOIN mssmt_roots 
     ON universe_roots.namespace_root = mssmt_roots.namespace
 JOIN mssmt_nodes 
-    ON mssmt_nodes.hash_key = mssmt_roots.root_hash AND
-       mssmt_nodes.namespace = mssmt_roots.namespace
+    ON mssmt_nodes.hash_key = mssmt_roots.root_hash 
+       AND mssmt_nodes.namespace = mssmt_roots.namespace
 JOIN genesis_assets
-     ON genesis_assets.asset_id = universe_roots.asset_id
+    ON genesis_assets.asset_id = universe_roots.asset_id
 WHERE mssmt_nodes.namespace = @namespace;
 
 -- name: UpsertUniverseRoot :one
@@ -30,7 +30,7 @@ WITH root_id AS (
     WHERE namespace_root = @namespace_root
 )
 DELETE FROM universe_events
-WHERE universe_root_id = (SELECT id from root_id);
+WHERE universe_root_id = (SELECT id FROM root_id);
 
 -- name: DeleteUniverseRoot :exec
 DELETE FROM universe_roots
@@ -54,25 +54,23 @@ DELETE FROM universe_leaves
 WHERE leaf_node_namespace = @namespace;
 
 -- name: QueryUniverseLeaves :many
-SELECT leaves.script_key_bytes, gen.gen_asset_id, nodes.value genesis_proof, 
-       nodes.sum sum_amt, gen.asset_id
-FROM universe_leaves leaves
-JOIN mssmt_nodes nodes
-    ON leaves.leaf_node_key = nodes.key AND
-        leaves.leaf_node_namespace = nodes.namespace
-JOIN genesis_info_view gen
+SELECT leaves.script_key_bytes, gen.gen_asset_id, nodes.value AS genesis_proof, 
+       nodes.sum AS sum_amt, gen.asset_id
+FROM universe_leaves AS leaves
+JOIN mssmt_nodes AS nodes
+    ON leaves.leaf_node_key = nodes.key 
+       AND leaves.leaf_node_namespace = nodes.namespace
+JOIN genesis_info_view AS gen
     ON leaves.asset_genesis_id = gen.gen_asset_id
 WHERE leaves.leaf_node_namespace = @namespace 
-        AND 
-    (leaves.minting_point = sqlc.narg('minting_point_bytes') OR 
-        sqlc.narg('minting_point_bytes') IS NULL) 
-        AND
-    (leaves.script_key_bytes = sqlc.narg('script_key_bytes') OR 
-        sqlc.narg('script_key_bytes') IS NULL);
+      AND (leaves.minting_point = sqlc.narg('minting_point_bytes') OR 
+           sqlc.narg('minting_point_bytes') IS NULL) 
+      AND (leaves.script_key_bytes = sqlc.narg('script_key_bytes') OR 
+           sqlc.narg('script_key_bytes') IS NULL);
 
 -- name: FetchUniverseKeys :many
 SELECT leaves.minting_point, leaves.script_key_bytes
-FROM universe_leaves leaves
+FROM universe_leaves AS leaves
 WHERE leaves.leaf_node_namespace = @namespace
 ORDER BY 
     CASE WHEN sqlc.narg('sort_direction') = 0 THEN leaves.id END ASC,
@@ -84,14 +82,14 @@ SELECT * FROM universe_leaves;
 
 -- name: UniverseRoots :many
 SELECT universe_roots.asset_id, group_key, proof_type,
-       mssmt_roots.root_hash root_hash, mssmt_nodes.sum root_sum,
-       genesis_assets.asset_tag asset_name
+       mssmt_roots.root_hash AS root_hash, mssmt_nodes.sum AS root_sum,
+       genesis_assets.asset_tag AS asset_name
 FROM universe_roots
 JOIN mssmt_roots
     ON universe_roots.namespace_root = mssmt_roots.namespace
 JOIN mssmt_nodes
-    ON mssmt_nodes.hash_key = mssmt_roots.root_hash AND
-       mssmt_nodes.namespace = mssmt_roots.namespace
+    ON mssmt_nodes.hash_key = mssmt_roots.root_hash 
+       AND mssmt_nodes.namespace = mssmt_roots.namespace
 JOIN genesis_assets
     ON genesis_assets.asset_id = universe_roots.asset_id
 ORDER BY 
@@ -329,8 +327,9 @@ SELECT
     SUM(CASE WHEN event_type = 'SYNC' THEN 1 ELSE 0 END) AS sync_events,
     SUM(CASE WHEN event_type = 'NEW_PROOF' THEN 1 ELSE 0 END) AS new_proof_events
 FROM universe_events
-WHERE event_type IN ('SYNC', 'NEW_PROOF') AND
-      event_timestamp >= @start_time AND event_timestamp <= @end_time
+-- BETWEEN is inclusive for both start and end values.
+WHERE event_type IN ('SYNC', 'NEW_PROOF') 
+      AND event_timestamp BETWEEN @start_time AND @end_time
 GROUP BY day
 ORDER BY day;
 
@@ -367,7 +366,7 @@ FROM federation_uni_sync_config
 ORDER BY group_key NULLS LAST, asset_id NULLS LAST, proof_type;
 
 -- name: UpsertFederationProofSyncLog :one
-INSERT INTO federation_proof_sync_log as log (
+INSERT INTO federation_proof_sync_log AS log (
     status, timestamp, sync_direction, proof_leaf_id, universe_root_id,
     servers_id
 ) VALUES (
@@ -401,7 +400,7 @@ DO UPDATE SET
     timestamp = EXCLUDED.timestamp,
     -- Increment the attempt counter.
     attempt_counter = CASE
-       WHEN @bump_sync_attempt_counter = true THEN log.attempt_counter + 1
+       WHEN @bump_sync_attempt_counter = TRUE THEN log.attempt_counter + 1
        ELSE log.attempt_counter
     END
 RETURNING id;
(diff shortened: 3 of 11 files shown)


jbrill:
note: incorporating postgres functionality now for the test

tapdb/sqlc/migrations/000024_universe_optimization_indexes_queries.up.sql
```diff
@@ -0,0 +1,53 @@
...
+);
+
+-- Multiverse optimization
+CREATE INDEX IF NOT EXISTS idx_multiverse_leaves_composite 
+ON multiverse_leaves(
+    multiverse_root_id, leaf_node_namespace,
+    asset_id, group_key, leaf_node_key
+);
+
+-- Analyze existing tables
```
guggero, line 46: Could add a quick explanation for those previously unaware of the `ANALYZE` command.
From the docs:

> The ANALYZE command gathers statistics about tables and indices and stores the collected information in [internal tables](https://www.sqlite.org/fileformat2.html#intschema) of the database where the query optimizer can access the information and use it to help make better query planning choices.

tapdb/sqlc/migrations/000024_universe_optimization_indexes_queries.up.sql
```diff
@@ -0,0 +1,53 @@
+-- Core universe_roots indexes
```
guggero, line 1: nit: we generally prefer all comments to be full sentences (including punctuation).

tapdb/sqlc/migrations/000024_universe_optimization_indexes_queries.up.sql
```diff
@@ -0,0 +1,53 @@
...
+    leaf_node_namespace, minting_point, script_key_bytes,
+    id, leaf_node_key, universe_root_id, asset_genesis_id
+);
+
+CREATE INDEX IF NOT EXISTS idx_universe_leaves_sort 
+ON universe_leaves(leaf_node_namespace, id, minting_point, script_key_bytes);
+
+-- MSSMT nodes optimization
+CREATE INDEX IF NOT EXISTS idx_mssmt_nodes_namespace 
+ON mssmt_nodes(namespace, hash_key, key, value, sum);
```
guggero, line 20: We do we include the value and sum in the index? AFAIK we never use those in an actual query, or do we?

tapdb/sqlc/queries/universe.sql
```diff
@@ -329,8 +327,8 @@ SELECT
     SUM(CASE WHEN event_type = 'SYNC' THEN 1 ELSE 0 END) AS sync_events,
     SUM(CASE WHEN event_type = 'NEW_PROOF' THEN 1 ELSE 0 END) AS new_proof_events
 FROM universe_events
-WHERE event_type IN ('SYNC', 'NEW_PROOF') AND
-      event_timestamp >= @start_time AND event_timestamp <= @end_time
+WHERE event_type IN ('SYNC', 'NEW_PROOF') 
+      AND event_timestamp BETWEEN @start_time AND @end_time
```
guggero, line 331: :+1: makes sense. A comment that `BETWEEN` is inclusive for both start and end value might be helpful, as I assume the more explicit form was used to avoid needing a documentation lookup.

tapdb/universe_perf_test.go
```diff
@@ -191,10 +176,10 @@ func TestUniverseIndexPerformance(t *testing.T) {
 	t.Parallel()
 
 	const (
-		numAssets         = 25
+		numAssets         = 50
```
guggero, line 179: We might want to have a testing flag in which we can manually significantly bump these numbers to see how the tradeoffs work out at scales more close to what we see in our main universe.

guggero (commented):
Awesome work on analyzing and optimizing the universe tables! Did a first pass just to get some early feedback in.

tapdb/sqlc/migrations/000024_universe_optimization_indexes_queries.up.sql
jbrill: Did some digging and it looks like we `SELECT` either `sum` or `value`:

`value`:
- Used in the `QueryUniverseLeaves` query, selected as `genesis_proof`.

`sum`:
- Used in the `FetchUniverseRoot` query, selected as `root_sum`.
- Used in the `QueryUniverseLeaves` query, selected as `sum_amt`.
- Used in the `QueryUniverseAssetStats` query in an aggregation to calculate the total supply.

My understanding is using this `covered` index helps improve these queries by not accessing the main tables:
https://www.postgresql.org/docs/current/indexes-index-only-scans.html

tapdb/sqlc/migrations/000024_universe_optimization_indexes_queries.up.sql
jbrill: So I'm thinking we could scrap the `analyze` from the migration, and instead utilize it during maintenance windows potentially. I'm reading that the `analyze` command can be resource intensive for users with many rows.

tapdb/sqlc/migrations/000024_universe_optimization_indexes_queries.up.sql
guggero: Sorry, I was a bit imprecise. I mean we never use those values in the `WHERE` or `GROUP BY` clauses of a query, so an index shouldn't really be crucial.
Or is the idea to have a [covering index](https://www.postgresql.org/docs/current/indexes-index-only-scans.html) so we don't have to do a data area lookup for these queries at all?

jbrill: Yes, I _think_ here that with a covering index we could bypass table scans at all. I think though that it might not be worth the storage overhead, so I went ahead and made this a minimal index instead.

tapdb/universe_perf_test.go
jbrill: I went ahead and created a `universe-scale` testing flag. I'm not sure if this is the correct approach here

tapdb/universe_perf_test.go
```diff
@@ -180,19 +181,20 @@ func formatSize(bytes int64) string {
 
 // TestUniverseIndexPerformance tests query performance with indices.
 func TestUniverseIndexPerformance(t *testing.T) {
-	if testing.Short() {
-		t.Skip("skipping index performance test in short mode")
-	}
+    if testing.Short() {
```
guggero, line 184: Incorrect formatting. Probably need to configure your IDE to use 8 spaces for tabs by default.

tapdb/universe_perf_test.go
```diff
@@ -180,19 +181,20 @@ func formatSize(bytes int64) string {
...
 
-	const (
-		numAssets         = 50
-		numLeavesPerTree  = 10
-		numEventsPerAsset = 25
-		numQueries        = 50
-		batchSize         = 5
-	)
+    // Check for universe-scale test flag
+    if flag.Lookup("test.universe-scale") != nil {
```
guggero, line 191: Nice. I think we should rename the flag though, since anything `test.*` is kind of reserved for Golang internal testing harness. So can perhaps just be `long-tests` and we can re-use that flag for other places too?

tapdb/universe_perf_test.go
```diff
@@ -358,15 +359,21 @@ func TestUniverseIndexPerformance(t *testing.T) {
 	}
 }
 
-// TestUniverseQuerySyncStatsSorting checks that query results are sorted correctly
+// TestUniversePerfQuerySyncStatsSorting checks that query results are sorted correctly
```
guggero, line 362: Can probably squash this commit with the previous one.

tapdb/universe_stats.go
```diff
@@ -285,16 +285,16 @@ type UniverseStats struct {
 
 	clock clock.Clock
 
-	statsMtx         sync.Mutex
+	statsMtx         sync.RWMutex
```
guggero, line 288: nit: probably don't need to change this one, since we're not using any read-only locking for it.

tapdb/universe_perf_test.go
```diff
@@ -0,0 +1,679 @@
...
+				} else {
+					stat.withoutIndices = avgTime
+				}
+				stat.queries = numQueries
+			}
+		})
+	}
+
+	// Execute tests without and with indices
+	runTest(false)
```
guggero, line 348: Could shed a whole level of indentation above by doing:
```
	for _, indices := range []bool{false, true} {
		t.Run(fmt.Sprintf("indices=%v", indices), func(tt *testing.T) {
			runTest(tt, indices)
		})
	}
```

tapdb/universe_perf_test.go
```diff
@@ -0,0 +1,679 @@
...
+					ON universe_events (universe_root_id);
+					CREATE INDEX IF NOT EXISTS idx_universe_events_sync 
+					ON universe_events (event_type);
+					CREATE INDEX IF NOT EXISTS idx_universe_leaves_universe_root_id 
+					ON universe_leaves (universe_root_id);
+					CREATE INDEX IF NOT EXISTS idx_universe_leaves_asset_genesis_id 
+					ON universe_leaves (asset_genesis_id);
+				`)
+				require.NoError(t, err)
+			}
```
guggero: This should probably be a helper function since it looks like the same code is used three times (but different versions of it?)

guggero (commented):
Looks good, thanks for the updates!
We're mostly down to code style, so getting closer.

jbrill:
@guggero thanks so much for the feedback. I've made those changes, which include some sweeping changes to both the indices and the performance test.

I opted to use the `EXPLAIN ANALYZE` statement to get more insightful details of our indices and how they are altering our main queries.

Here is the output for [postgres](https://pastebin.com/dRfQP5J4) and [sqlite](https://pastebin.com/VrRL73mf).

guggero:
Can you please squash the two middle commits so it's a bit easier to review? Going to take another look next week.

jbrill:
> Can you please squash the two middle commits so it's a bit easier to review? Going to take another look next week.

Sounds good. I've also updated the PR description to reflect the lock changes and analysis from the query plans.

make/testing_flags.mk
```diff
@@ -54,6 +54,11 @@ ifeq ($(dbbackend),postgres)
 DEV_TAGS += test_db_postgres
 endif
 
+# Run universe tests with increased scale for performance testing.
+ifneq ($(long-tests),)
+TEST_FLAGS += long-tests
```
jbrill, line 59: NOTE: Unsure if this is the correct approach here.

tapdb/universe_stats.go
```diff
@@ -750,13 +752,18 @@ func (u *UniverseStats) QuerySyncStats(ctx context.Context,
 
 	// First, check the cache to see if we already have a cached result for
 	// this query.
-	syncSnapshots := u.syncStatsCache.fetchQuery(q)
+	var syncSnapshots []universe.AssetSyncSnapshot
+
+	u.syncStatsMtx.RLock()
+	syncSnapshots = u.syncStatsCache.fetchQuery(q)
```
guggero, line 758: nit: can just use the `:=` syntax here, then you don't need to declare the `var` above.

make/testing_flags.mk
guggero: Sorry, I think I lead you on the wrong path here... Turns out flags aren't as easy to use as build tags.
Have a working version here, feel free to just apply this to your commit directly: https://github.com/guggero/taproot-assets/commit/34a0a2824e8869d9390ea8913761f26df937fe57

Can test with: `make unit pkg=tapdb` vs. `make unit pkg=tapdb long-tests=true`.

tapdb/universe_perf_test.go
```diff
@@ -0,0 +1,485 @@
...
+			namespace,
+			key,
+			hash_key,
+			sum
+		);`
+)
+
+// executeSQLStatements executes a list of SQL statements with logging.
+func executeSQLStatements(
+	t *testing.T,
```
guggero, line 59: See this file for our preferred formatting style (which implicitly we try to observe in all our Golang projects): https://github.com/lightningnetwork/lnd/blob/master/docs/code_formatting_rules.md#wrapping-long-function-definitions

guggero (commented):
Very nice! Extremely close, going to assign another reviewer as soon as the build tags are working (see commit I pushed).

jbrill:
@guggero thanks for the round of feedback! I've patched that commit onto the performance test commit in this PR. I've also updated the code style based on the formatting rules doc. Thanks a bunch for this reference.

tapdb/universe_perf_test.go
```diff
@@ -0,0 +1,402 @@
...
+				ON leaves.leaf_node_key = nodes.key 
+				AND leaves.leaf_node_namespace = nodes.namespace
+			JOIN genesis_info_view gen
+				ON leaves.asset_genesis_id = gen.gen_asset_id
+			WHERE roots.proof_type = 'issuance'
+			GROUP BY gen.asset_id
+		)
+		SELECT asset_id, supply, COUNT(*) as num_leaves
+		FROM asset_supply
+		GROUP BY asset_id, supply`
```
GeorgeTsagk, line 141: why not compile these queries with sqlc?

GeorgeTsagk: (nit, non blocking)

GeorgeTsagk (approved):

GeorgeTsagk:
kicking itests, probably flake

tapdb/universe_perf_test.go
guggero: They would then be part of the "production" code, which we don't necessarily want. So I think in tests it's fine to have pure SQL, as long as it's compatible with both SQLite and Postgres.

PR #1176, opened 2024-11-07
Hashmail logging cleanup

ffranr:


tapfreighter/chain_porter.go
```diff
@@ -763,15 +763,73 @@ func (p *ChainPorter) updateAssetProofFile(ctx context.Context,
 	}, nil
 }
 
+// reportProofTransfers logs a summary of the transfer outputs that require
+// proof delivery and those that do not.
+func reportProofTransfers(
+	notDeliveringOutputs []TransferOutput,
```
guggero, line 769: nit: first parameter seems to fit on previous line?

proof/courier.go
```diff
@@ -439,6 +439,17 @@ func (h *HashMailBox) RecvAck(ctx context.Context, sid streamID) error {
 		return fmt.Errorf("unable to create read stream: %w", err)
 	}
 
+	// Ensure the read stream is closed before returning.
+	defer func() {
+		log.Debugf("Closing hashmail read stream (sid=%x)", sid[:])
+
+		err := readStream.CloseSend()
```
guggero, line 446: I don't think this is necessary. If you look into `*hashMailClient.RecvStream()`, you see that `x.ClientStream.CloseSend()` is already called. So I think this might even lead to an error?

guggero (commented):
Thanks for the logging improvements! One question about the last commit.

proof/courier.go
ffranr: Dropped last commit in latest push.

guggero (approved):

jharveyb (approved):
LGTM 👍🏽

PR #1177, opened 2024-11-07
tapcfg: allow to configure lnd RPC timeout, increase default val

guggero:
This fixes an issue where tapd would shut down because querying channels on a slow lnd would time out:

```
2024-10-30 10:06:21.535 [DBG] STAT: Setting the taproot-assets sub-server as errored: received critical error from sub-server (taproot-assets), shutting down: failed to handle outgoing message: error adding local alias: add alias: error listing local channels: rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

Will add a similar PR to `litd`, as there the `lndclient` is created in a single place for all sub servers, so this daemon specific value would be ignored.

### DIFF
--- a/proof/courier.go
+++ b/proof/courier.go
@@ -550,7 +550,7 @@ type BackoffCfg struct {
 
 	// BackoffResetWait is the amount of time we'll wait before
 	// resetting the backoff counter to its initial state.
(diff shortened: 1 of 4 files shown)


GeorgeTsagk:
pending lint / sample conf

Roasbeef (approved):
LGTM 🕸️

PR #1178, opened 2024-11-07
Add `ListBurns` RPC 

### LINKED ISSUE #1095: tapdb - [feature]: add new associative table to track asset burns `ListBurns`

Today we have an easy way to [burn](https://lightning.engineering/api-docs/api/taproot-assets/taproot-assets/burn-asset) assets, but no easy way to track all burns we've made for a given asset. 

With the way things work today, burns are just another transfer: https://github.com/lightninglabs/taproot-assets/blob/420f2467a6770a9c6042159dcd4c3ce194db25c2/rpcserver.go#L3258-L3282

However, we don't have an easy/efficient way to scan the transfers table for all the burns we've done: https://github.com/lightninglabs/taproot-assets/blob/420f2467a6770a9c6042159dcd4c3ce194db25c2/tapdb/sqlc/migrations/000005_transfers.up.sql#L1-L9

We should add a new associative table (that can store other meta data) that we add to each time we burn an asset. We should then create a set of queries to easily look up the set of burnt assets all time, for a given asset, etc, etc. 

A draft table would look something like: 
```sql
CREATE TABLE IF NOT EXISTS asset_burn_transfers (
     burn_id INTEGER PRIMARY KEY, 

     transfer_id BIGINT NOT NULL REFERENCES asset_transfers(id),
     
     note TEXT,

     asset_id BIGINT NOT NULL REFERENCES assets(asset_id),

     group_key_id BIGINT NOT NULL REFERENCES asset_groups(group_id) 
)
```

The `note` field can be used to store other metadata related to a burn. 

A user can scan all the proof files on disk, to find those that end in a bug suffix, but this doesn't lend well to the creation of an automated system.

GeorgeTsagk:
## Description

We're currently able to provably burn assets with tapd but we have no (easy) way of retrieving burn related historical data. This PR adds a DB table and exposes some new RPC methods/parameters to help keep track of burns.

When completing an asset burn, we add an entry to the new burns table. We also add a new `ListBurns` method which returns a list of all burns that pass the user-provided filters (asset id, group id, anchor txid)

Closes #1095

### DIFF
--- a/cmd/tapcli/assets.go
+++ b/cmd/tapcli/assets.go
@@ -29,6 +29,7 @@ var assetsCommands = []cli.Command{
 			listAssetBalancesCommand,
 			sendAssetsCommand,
 			burnAssetsCommand,
+			listBurnsCommand,
 			listTransfersCommand,
 			fetchMetaCommand,
 		},
@@ -52,6 +53,7 @@ var (
 	assetShowUnconfMintsName     = "show_unconfirmed_mints"
 	assetGroupKeyName            = "group_key"
 	assetGroupAnchorName         = "group_anchor"
+	anchorTxidName               = "anchor_txid"
 	batchKeyName                 = "batch_key"
 	groupByGroupName             = "by_group"
 	assetIDName                  = "asset_id"
@@ -852,6 +854,71 @@ func burnAssets(ctx *cli.Context) error {
 	})
 	if err != nil {
 		return fmt.Errorf("unable to send assets: %w", err)
+	}
+
+	printRespJSON(resp)
+	return nil
+}
+
+var listBurnsCommand = cli.Command{
+	Name:  "listburns",
+	Usage: "list burnt assets",
+	Description: `
+	List assets that have been burned by this daemon. These are assets that
+	have been destroyed and are no longer spendable.
+
+	Some filters may be used to return more specific results.
+	`,
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  assetIDName,
+			Usage: "the asset ID of the burnt asset",
+		},
+		cli.StringFlag{
+			Name:  assetGroupKeyName,
+			Usage: "the group key of the burnt asset",
+		},
+		cli.StringFlag{
+			Name: anchorTxidName,
+			Usage: "the txid of the transaction the burn was " +
+				"anchored to",
+		},
+	},
+	Action: listBurns,
+}
+
+func listBurns(ctx *cli.Context) error {
+	assetIDHex := ctx.String(assetIDName)
+	assetIDBytes, err := hex.DecodeString(assetIDHex)
+	if err != nil {
+		return fmt.Errorf("invalid asset ID: %w", err)
+	}
+
+	groupKeyHex := ctx.String(assetGroupKeyName)
+	groupKeyBytes, err := hex.DecodeString(groupKeyHex)
+	if err != nil {
+		return fmt.Errorf("invalid group key: %w", err)
+	}
+
+	anchorTxidStr := ctx.String(anchorTxidName)
+	anchorTxid, err := hex.DecodeString(anchorTxidStr)
+	if err != nil {
+		return fmt.Errorf("invalid anchor txid: %w", err)
+	}
+
+	ctxc := getContext()
+	client, cleanUp := getClient(ctx)
+	defer cleanUp()
+
+	resp, err := client.ListBurns(
+		ctxc, &taprpc.ListBurnsRequest{
+			AssetId:         assetIDBytes,
+			TweakedGroupKey: groupKeyBytes,
+			AnchorTxid:      anchorTxid,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("could not list burns: %w", err)
 	}
 
 	printRespJSON(resp)
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -2275,7 +2275,7 @@ func (r *rpcServer) AnchorVirtualPsbts(ctx context.Context,
 	}
 
 	resp, err := r.cfg.ChainPorter.RequestShipment(
-		tapfreighter.NewPreSignedParcel(vPackets, inputCommitments),
+		tapfreighter.NewPreSignedParcel(vPackets, inputCommitments, ""),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("error requesting delivery: %w", err)
@@ -3290,7 +3290,7 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
 	resp, err := r.cfg.ChainPorter.RequestShipment(
 		tapfreighter.NewPreSignedParcel(
 			[]*tappsbt.VPacket{fundResp.VPacket},
-			fundResp.InputCommitments,
+			fundResp.InputCommitments, "",
 		),
 	)
 	if err != nil {
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -37,6 +37,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/rfqmsg"
 	"github.com/lightninglabs/taproot-assets/rpcperms"
 	"github.com/lightninglabs/taproot-assets/tapchannel"
+	"github.com/lightninglabs/taproot-assets/tapdb"
 	"github.com/lightninglabs/taproot-assets/tapfreighter"
 	"github.com/lightninglabs/taproot-assets/tapgarden"
 	"github.com/lightninglabs/taproot-assets/tappsbt"
@@ -3290,7 +3291,7 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
 	resp, err := r.cfg.ChainPorter.RequestShipment(
 		tapfreighter.NewPreSignedParcel(
 			[]*tappsbt.VPacket{fundResp.VPacket},
-			fundResp.InputCommitments, "",
+			fundResp.InputCommitments, in.Note,
 		),
 	)
 	if err != nil {
@@ -3326,6 +3327,40 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
 		BurnTransfer: parcel,
 		BurnProof:    burnProof,
 	}, nil
+}
+
+// ListBurns returns a list of burnt assets. Some filters may be defined in the
+// request to return more specific results.
+func (r *rpcServer) ListBurns(ctx context.Context,
+	in *taprpc.ListBurnsRequest) (*taprpc.ListBurnsResponse, error) {
+
+	burns, err := r.cfg.AssetStore.QueryBurns(
+		ctx, tapdb.QueryBurnsFilters{
+			AssetID:    in.AssetId,
+			GroupKey:   in.TweakedGroupKey,
+			AnchorTxid: in.AnchorTxid,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	rpcBurns := fn.Map(burns, marshalRpcBurn)
+
+	return &taprpc.ListBurnsResponse{
+		Burns: rpcBurns,
+	}, nil
+}
+
+// marshalRpcBurn creates an instance of *taprpc.AssetBurn from the tapdb model.
+func marshalRpcBurn(b *tapfreighter.AssetBurn) *taprpc.AssetBurn {
+	return &taprpc.AssetBurn{
+		Note:            b.Note,
+		AssetId:         b.AssetID,
+		TweakedGroupKey: b.GroupKey,
+		Amount:          b.Amount,
+		AnchorTxid:      b.AnchorTxid[:],
+	}
 }
 
 // marshalOutboundParcel turns a pending parcel into its RPC counterpart.
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -179,6 +179,10 @@ type (
 	// when fetching a tapscript tree, which includes the serialized node
 	// and the node index in the tree.
 	TapscriptTreeNode = sqlc.FetchTapscriptTreeRow
+
+	// QueryBurnsFilters is a set of filters that is applied on the set of
+	// the returned burns.
+	QueryBurnsFilters = sqlc.QueryBurnsParams
 )
 
 // ActiveAssetsStore is a sub-set of the main sqlc.Querier interface that
@@ -358,6 +362,15 @@ type ActiveAssetsStore interface {
 	// FetchAssetMetaForAsset fetches the asset meta for a given asset.
 	FetchAssetMetaForAsset(ctx context.Context,
 		assetID []byte) (sqlc.FetchAssetMetaForAssetRow, error)
+
+	// InsertBurn inserts a new row to the asset burns table which
+	// includes all important data related to the burn.
+	InsertBurn(ctx context.Context, arg sqlc.InsertBurnParams) (int64,
+		error)
+
+	// QueryBurns returns all burn entries that match the passed filters.
+	QueryBurns(ctx context.Context,
+		arg sqlc.QueryBurnsParams) ([]sqlc.QueryBurnsRow, error)
 }
 
 // MetaStore is a sub-set of the main sqlc.Querier interface that contains
@@ -2951,7 +2964,8 @@ func (a *AssetStore) ConfirmProofDelivery(ctx context.Context,
 // confirmation of the anchor transaction, ensuring the on-chain reference
 // information is up to date.
 func (a *AssetStore) LogAnchorTxConfirm(ctx context.Context,
-	conf *tapfreighter.AssetConfirmEvent) error {
+	conf *tapfreighter.AssetConfirmEvent,
+	burns []*tapfreighter.AssetBurn) error {
 
 	var (
 		writeTxOpts    AssetStoreTxOptions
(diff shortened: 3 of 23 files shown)


tapdb/assets_store.go
```diff
@@ -402,6 +415,14 @@ func (c cacheableBlockHeight) Size() (uint64, error) {
 	return 1, nil
 }
 
+type AssetBurn struct {
```
guggero, line 418: nit: missing Godoc on struct and members.

tapdb/assets_store.go
```diff
@@ -3526,6 +3547,90 @@ func (a *AssetStore) TxHeight(ctx context.Context, txid chainhash.Hash) (uint32,
...
+
+		return nil
+	})
+}
+
+// QueryBurns queries burnt assets based on the passed filters.
+func (a *AssetStore) QueryBurns(ctx context.Context,
+	filters QueryBurnsFilters) ([]*AssetBurn, error) {
+
+	res := make([]*AssetBurn, 0)
```
guggero, line 3594: nit: don't make with 0, just do `var res []*AssetBurn` instead.

rpcserver.go
```diff
@@ -3312,12 +3314,58 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
 		}
 	}
 
+	// At this point everything completed correctly, so we log this burn in
+	// our db.
+	err = r.cfg.AssetStore.InsertBurn(
```
guggero, line 3319: nit: can fit more params per line, making a bit more compact.

rpcserver.go
```diff
@@ -3312,12 +3314,58 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
+		ctx, parcel.AnchorTxHash, in.Note,
+		assetID[:], serializedGroupKey,
+		in.AmountToBurn,
+	)
```
guggero, line 3323: It's customary to return a `nil` response on error.

rpcserver.go
```diff
@@ -3312,12 +3314,58 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	rpcBurns := make([]*taprpc.AssetBurn, len(burns))
+
+	for i, b := range burns {
+		rpcBurns[i] = marshalRpcBurn(b)
```
guggero, line 3350: nit: since there's no error return value, can use `fn.Map()`.

taprpc/taprootassets.proto
```diff
@@ -104,6 +104,8 @@ service TaprootAssets {
     */
     rpc BurnAsset (BurnAssetRequest) returns (BurnAssetResponse);
 
+    rpc ListBurns (ListBurnsRequest) returns (ListBurnsResponse);
```
guggero, line 107: nit: missing description and CLI name for API docs generator.

taprpc/taprootassets.yaml
```diff
@@ -61,6 +61,9 @@ http:
       post: "/v1/taproot-assets/burn"
       body: "*"
 
+    - selector: taprpc.TaprootAssets.ListBurns
+      get: "/v1/taproot-assets/listburns"
```
guggero, line 65: nit: more RESTy would be `get: "/v1/taproot-assets/burns`.

cmd/tapcli/assets.go
```diff
@@ -858,6 +860,75 @@ func burnAssets(ctx *cli.Context) error {
...
+
+	resp, err := client.ListBurns(
+		ctxc, &taprpc.ListBurnsRequest{
+			AssetId:         assetIDBytes,
+			TweakedGroupKey: groupKeyBytes,
+			AnchorTxid:      anchorTxid,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("could not list burns")
```
guggero, line 925: nit: include error in message, here and above.

cmd/tapcli/assets.go
```diff
@@ -858,6 +860,75 @@ func burnAssets(ctx *cli.Context) error {
...
+			Name: anchorTxidName,
+			Usage: "the txid of the transaction the burn was " +
+				"anchored to",
+		},
+	},
+	Action: listBurns,
+}
+
+func listBurns(ctx *cli.Context) error {
+	if ctx.NArg() != 0 || ctx.NumFlags() == 0 {
```
guggero, line 891: I think calling this without any flags should be okay, if I just want all burns?

tapdb/sqlc/queries/transfers.sql
```diff
@@ -194,3 +194,33 @@ FROM passive_assets as passive
...
+    abt.asset_id,
+    abt.group_key,
+    abt.amount,
+    ct.txid AS anchor_txid -- Retrieving the txid from chain_txns
+FROM asset_burn_transfers abt
+JOIN asset_transfers at ON abt.transfer_id = at.id
+JOIN chain_txns ct ON at.anchor_txn_id = ct.txn_id
+WHERE
+    -- Optionally filter by asset_id.
+    (abt.asset_id = @asset_id OR @asset_id IS NULL)
```
guggero, line 219: Not sure if these should be nullable arguments (e.g. `sqlc.narg('asset_id')`) or things work correctly because the underlying type (`[]byte`) is nullable...
I'd say a basic DB level CRUD unit test should answer the question.

guggero (commented):
Looks pretty good. I wonder if we should add a Golang based "migration" of sort that queries existing assets in the DB, checks if they're burns (cannot be done on SQL level alone) and then inserts them into the new table retroactively?
Perhaps we could add a mechanism that detects if any DB level migrations were executed. And if they were, we can assume an update happened and could run these Golang-level checks (which would need to be idempotent).

tapdb/assets_store.go
```diff
@@ -3526,6 +3548,90 @@ func (a *AssetStore) TxHeight(ctx context.Context, txid chainhash.Hash) (uint32,
...
 
+// InsertBurn inserts a burnt asset to the db which includes all related
+// metadata.
+func (a *AssetStore) InsertBurn(ctx context.Context, txid []byte, note string,
+	assetID, groupKey []byte, amount uint64) error {
+
+	var writeTxOpts AssetStoreTxOptions
+	return a.db.ExecTx(ctx, &writeTxOpts, func(q ActiveAssetsStore) error {
+		// Look up the transfer that relates to this txid.
+		// TODO(george): We assume there's a single burn in every txn,
```
guggero, line 3559: I don't think we need this TODO. If there are multiple burns in a transaction, we'd call `InsertBurn` multiple times with the same `txid`. But because they would all still be in the same transfer, we'd query the same transfer ID which is correct.

tapdb/assets_store.go
```diff
@@ -3526,6 +3548,90 @@ func (a *AssetStore) TxHeight(ctx context.Context, txid chainhash.Hash) (uint32,
...
+		// this will need to be fixed in the future.
+		assetTransfers, err := q.QueryAssetTransfers(ctx, TransferQuery{
+			AnchorTxHash: txid,
+		})
+		if err != nil {
+			return fmt.Errorf("unable to query asset transfers: %w",
+				err)
+		}
+		assetTransfer := assetTransfers[0]
```
guggero, line 3568: Should check that we have exactly one result (which should always be the case when we query with an anchor TX hash).

tapdb/sqlc/queries/transfers.sql
```diff
@@ -194,3 +194,33 @@ FROM passive_assets as passive
...
+)
+RETURNING burn_id;
+
+-- name: QueryBurns :many
+SELECT
+    abt.note,
+    abt.asset_id,
+    abt.group_key,
+    abt.amount,
+    ct.txid AS anchor_txid -- Retrieving the txid from chain_txns
```
guggero, line 213: nit: missing full stop, here and below.

tapdb/sqlc/queries/transfers.sql
```diff
@@ -194,3 +194,33 @@ FROM passive_assets as passive
...
+FROM asset_burn_transfers abt
+JOIN asset_transfers at ON abt.transfer_id = at.id
```
guggero, line 215: Looks like we don't need to join on `asset_transfers`? Since we don't seem to be using `at` anywhere.

tapdb/assets_store_test.go
```diff
@@ -2204,6 +2205,44 @@ func TestTransferOutputProofDeliveryStatus(t *testing.T) {
 	require.Equal(
 		t, randBlockHash[:], assetTransfers[0].AnchorTxBlockHash,
 	)
+
+	// Let's insert a burn.
+	assetID := inputAsset.ID()
+	transferID := assetTransfers[0].ID
+	_, err = assetsStore.db.InsertBurn(ctx, sqlc.InsertBurnParams{
```
guggero, line 2212: We'll want to test the actual `AssetStore`'s `QueryBurns` and `InsertBurn` methods, not directly call into the DB-level methods.

rpcserver.go
```diff
@@ -3312,12 +3314,56 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
 	}, nil
 }
 
+// ListBurns returns a list of burnt assets. Some filters may be defined in the
+// request to return more specific results.
+func (r *rpcServer) ListBurns(ctx context.Context,
+	in *taprpc.ListBurnsRequest) (*taprpc.ListBurnsResponse, error) {
+
+	burns, err := r.cfg.AssetStore.QueryBurns(
+		ctx, sqlc.QueryBurnsParams{
```
guggero, line 3339: We should use the alias `tapdb.QueryBurnsFilters` here.

cmd/tapcli/assets.go
```diff
@@ -858,6 +860,71 @@ func burnAssets(ctx *cli.Context) error {
...
+	groupKeyHex := ctx.String(assetGroupKeyName)
+	groupKeyBytes, err := hex.DecodeString(groupKeyHex)
+	if err != nil {
+		return fmt.Errorf("invalid group key: %w", err)
+	}
+
+	anchorTxidStr := ctx.String(anchorTxidName)
+	anchorTxid, err := hex.DecodeString(anchorTxidStr)
+	if err != nil {
+		return fmt.Errorf("invalid anchor txid")
```
guggero, line 906: nit: missing wrapped error.

guggero (commented):
Super close!

tapdb/assets_store.go
Roasbeef: ^ makes sense

tapdb/assets_store.go
Roasbeef: Yeah otherwise, we'll panic here.

tapdb/assets_store.go
```diff
@@ -3526,6 +3548,90 @@ func (a *AssetStore) TxHeight(ctx context.Context, txid chainhash.Hash) (uint32,
...
+
+		for _, b := range burns {
+			burn := marshalAssetBurnTransfer(b)
+
+			res = append(res, burn)
+		}
+
+		return nil
+	})
+
```
Roasbeef, line 3616: Style nit: extra space.

tapdb/assets_store_test.go
```diff
@@ -2204,6 +2205,44 @@ func TestTransferOutputProofDeliveryStatus(t *testing.T) {
 	require.Equal(
 		t, randBlockHash[:], assetTransfers[0].AnchorTxBlockHash,
 	)
+
+	// Let's insert a burn.
```
Roasbeef, line 2209: How about we make another isolated unit test for this? This exiting test is already rather long/large.

tapdb/sqlc/queries/transfers.sql
Roasbeef: Perhaps the idea was to return the full transfer information alongside the burn? 

From the PoV of a future query CLI, I think that would be desirable.

rpcserver.go
```diff
@@ -3312,12 +3314,56 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
 		}
 	}
 
+	// At this point everything completed correctly, so we log this burn in
```
Roasbeef, line 3317: Hmm, perhaps we should do this further in the pipeline? So in the same db transaction that we insert the transfer. 

Otherwise, if we crash here, then the burn is never inserted in the db, and we exit in an inconsistent state.

rpcserver.go
```diff
@@ -3312,12 +3314,56 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	rpcBurns := fn.Map(burns, marshalRpcBurn)
+
+	return &taprpc.ListBurnsResponse{
+		Burns: rpcBurns,
```
Roasbeef, line 3352: Re other comment, I think it would be useful to include the full transfer here, also maybe the raw tx itself?

Roasbeef (changes requested):

tapdb/assets_store.go
GeorgeTsagk: > But because they would all still be in the same transfer, we'd query the same transfer ID which is correct.

Ah right, I believed that there would be multiple transfers which is not the way it work

tapdb/sqlc/queries/transfers.sql
GeorgeTsagk: I need to join in order to grab the asset_transfers row, and then link to the chain_txns row to get the txid

rpcserver.go
GeorgeTsagk: that seems to happen within the state machine of the chain porter, deep in the `r.cfg.ChainPorter.RequestShipment` step above

threading through all the params needed for the `InsertBurn` call seems like a bad way to go

rpcserver.go
GeorgeTsagk: a bit overblown? I'd expect the common usage of this call being about accounting, where only assetid/groupid and amount matters

if for any other reason (perhaps debugging related) the anchor tx / transfer is needed you could grab the anchor_txid and query them directly from the respective RPCs

rpcserver.go
Roasbeef: > threading through all the params needed for the InsertBurn call seems like a bad way to go

Can you think of an alternative? 

We want this visibility to be air tight. If it's possible to just not record a burn all together, then that's problematic. 

Perhaps we need a hybrid migration route where we scan the db to insert the new burn metadata into transfers, then can rely on that information being specified in a transfer? 

Can you explain the issue you see with adding this meta data at the transfers level to it can be inserted when we go to log the parcel?

rpcserver.go
Roasbeef: That's reasonable, if we need to expose it later, then it's on disk already i the `chain_txns` table.

rpcserver.go
```diff
@@ -3314,12 +3315,56 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
 		}
 	}
 
+	// At this point everything completed correctly, so we log this burn in
+	// our db.
+	err = r.cfg.AssetStore.InsertBurn(
```
Roasbeef, line 3320: I still think this is problematic. We make a call to `RequestShipment` above. What stops us from marking the shipment as a burn? We would also inherit the logic that only marks something as a finalized transfer once fully confirmed. `InsertBurn` would also not need to read the transfer at all, since it would be run in the same db txn that the transfer is inserted within.

Roasbeef (changes requested):

rpcserver.go
GeorgeTsagk: Seems like most of the burn info is available in the Parcel itself, so with the latest push I"m inserting the Burn in the same transaction as the transfer.

Only needed to add the `note string` argument as an extra to some interfaces, but it's minimal

bhandras:
> Looks pretty good. I wonder if we should add a Golang based "migration" of sort that queries existing assets in the DB, checks if they're burns (cannot be done on SQL level alone) and then inserts them into the new table retroactively? Perhaps we could add a mechanism that detects if any DB level migrations were executed. And if they were, we can assume an update happened and could run these Golang-level checks (which would need to be idempotent).

Just saw this comment, and thought this might come in handy: https://github.com/lightningnetwork/lnd/pull/8831/commits/fba2b5d24fb91d7c547640816d8175f843328b99

tapdb/assets_store.go
```diff
@@ -3526,6 +3568,94 @@ func (a *AssetStore) TxHeight(ctx context.Context, txid chainhash.Hash) (uint32,
 	return uint32(dbBlockHeight), nil
 }
 
+// InsertBurn inserts a burnt asset to the db which includes all related
+// metadata.
+func (a *AssetStore) InsertBurn(ctx context.Context, txid []byte, note string,
```
guggero, line 3573: This is no longer needed (outside of the unit test). Should we just remove it?

tapfreighter/chain_porter.go
```diff
@@ -646,6 +646,36 @@ func (p *ChainPorter) storePackageAnchorTxConf(pkg *sendPackage) error {
...
+			// We prepare the burn and add it to the list.
+			b := &AssetBurn{
+				AssetID:    assetID[:],
+				Amount:     o.Amount,
+				AnchorTxid: pkg.OutboundPkg.AnchorTx.TxHash(),
+				Note:       pkg.Note,
+			}
+
+			if o.Asset.GroupKey != nil &&
+				!o.Asset.GroupKey.RawKey.IsEmpty() {
```
guggero, line 669: This is wrong. `IsEmpty()` tests the key locator. But we don't access the `o.Asset.GroupKey.RawKey` below. So just remove this second condition.

tapdb/assets_store.go
```diff
@@ -402,6 +415,15 @@ func (c cacheableBlockHeight) Size() (uint64, error) {
 	return 1, nil
 }
 
+// AssetBurn holds data related to a burn of an asset.
+type AssetBurn struct {
```
guggero, line 419: Seems like we could use `tapfreighter.AssetBurn` instead, since we need to re-declare it there. Can move that declaration to this commit.

tapfreighter/parcel.go
```diff
@@ -464,6 +471,9 @@ type sendPackage struct {
 	// TransferTxConfEvent contains transfer transaction on-chain
 	// confirmation data.
 	TransferTxConfEvent *chainntnfs.TxConfirmation
+
+	// Note is a user provided description for this transfer.
```
guggero, line 475: nit: add something to the effect of "this is currently only used for burns", since we don't store it in the transfer table currently.

guggero (approved):
Nice, LGTM :tada:

Roasbeef (approved):
LGTM 👟

GeorgeTsagk:
rebased, waiting CI then merging

PR #1180, opened 2024-11-07
tappsbt: add AltLeaf support to vPacket

jharveyb:
Introduce the AltLeaves field for vPackets, which can be used to carry non-asset data alongside a Taproot Assets transfer. This data will be anchored alongside any assets involved in a transfer, allowing for other protocols to operate in the same anchor TXs as Taproot Assets.


jharveyb:
Remaining TODOs from existing feedback:

- [ ] Add explicit negative test cases for `[]AltLeaf` with either duplicate script keys or total size over the 64K limit.
- [ ] Explore a better interface definition for `AltLeaf` using `*T` to save on save manual type annotation.
- [ ] Update `AltLeaf` constructor to accept a script key, as a field that is not allowed to be nil.
- [ ] Update `AltLeaves` en/decoder to reject duplicate script keys.
- [ ] Add a type def for `AltLeaf[Asset]`.

Worth noting that with this more minimal encoder, the minimal size for a leaf is 42 bytes vs. 128 for a normal `Asset`.

jharveyb:
Rebased, addressed all TODO items, and added some further changes to the definition of an `AltLeaf`.

Specifically, the `lockTime` and `relativeLocktime` must be 0. The `SplitCommitmentRoot` must also be empty.

Prop tests were updated to reflect this, and there are failing cases for each condition of `ValidateAltLeaf()` committed as well.

asset/asset.go
```diff
@@ -2238,3 +2241,134 @@ type ChainAsset struct {
...
+
+	// EncodeAltLeaf encodes an AltLeaf into a TLV stream.
+	EncodeAltLeaf(w io.Writer) error
+
+	// DecodeAltLeaf decodes an AltLeaf from a TLV stream.
+	DecodeAltLeaf(r io.Reader) error
+}
+
+// NewAltLeaf instantiates a new valid AltLeaf.
+func NewAltLeaf(key ScriptKey) (*Asset, error) {
```
guggero, line 2264: If we create a constructor, should we take all params that make up the alt leaf? E.g. the prev witnesses and the script version?

asset/asset.go
```diff
@@ -2238,3 +2241,134 @@ type ChainAsset struct {
...
+
+	return nil
+}
+
+// encodeAltLeafRecords determines the set of non-nil records to include when
+// encoding an AltLeaf. Since the Genesis, Group Key, Amount, and Version fields
+// are static, we can omit those fields.
+func (a *Asset) encodeAltLeafRecords() []tlv.Record {
+	records := make([]tlv.Record, 0, 4)
+	records = append(records, NewLeafTypeRecord(&a.Type))
```
guggero, line 2341: Do we need the type? Since that's part of the Genesis, which is always the empty genesis?

tappsbt/mock.go
```diff
@@ -338,6 +338,29 @@ func NewTestFromVInput(t testing.TB, i *VInput) *TestVInput {
 		ti.Proof = proof.NewTestFromProof(t, i.Proof)
 	}
 
+	if len(i.AltLeaves) > 0 {
+		// Assert that the concrete type of AltLeaf is supported.
+		switch i.AltLeaves[0].(type) {
```
guggero, line 343: Too bad the property testing framework doesn't play nice with `require`. Otherwise this could just be `require.IsType(t, &asset.Asset{}, v.AltLeaves[0])`.

tappsbt/mock.go
```diff
@@ -747,5 +807,18 @@ func (to *TestVOutput) ToVOutput(t testing.TB) *VOutput {
 		)
 	}
 
+	if len(to.AltLeaves) > 0 {
+		v.AltLeaves = make(
+			[]AltLeafAsset, 0, len(to.AltLeaves),
+		)
+
+		for idx := range to.AltLeaves {
+			v.AltLeaves = append(
```
guggero, line 816: nit: could use index based assignment here and above, then the `make` call also gets short enough to fit a single line.

guggero (approved):
Just a couple of nits, otherwise LGTM :tada:

asset/encoding.go
```diff
@@ -803,3 +807,91 @@ func DecodeTapLeaf(leafData []byte) (*txscript.TapLeaf, error) {
...
+
+func AltLeavesDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
+	// There is no limit on the number of alt leaves, but the total size of
+	// all alt leaves must be below 64 KiB.
+	if l > math.MaxUint16 {
+		return tlv.ErrRecordTooLarge
+	}
+
+	if typ, ok := val.(*[]AltLeaf[*Asset]); ok {
+		// Each alt leaf is at least 42 bytes, which limits the total
```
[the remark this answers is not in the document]
jharveyb: Those are optional though - the required fields are Type, ScriptVersion, and ScriptKey. The scriptKey will be 33 bytes, Type is 1 byte, and ScriptVersion is 2 bytes (TIL).

2 byte overhead per field is 33+1+2+(3*2) = 42

tappsbt/mock.go
jharveyb: Ah didn't know about that in `require`!

FWIW I think usage of `require` can be replaced with `assert` + `Errorf` in most cases.

asset/asset.go
jharveyb: Good point - I don't think so, since it would always be `Normal`.

asset/encoding.go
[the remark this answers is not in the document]
guggero: Ah, yeah, you mention "at least". And we do limit the maximum total size above. Never mind then.

tappsbt/mock.go
jharveyb: I think the `make` line break was needed before adding the `AltLeafAsset` type alias.

jharveyb:
Addressed nits, updated encoding to drop the `Type` field, rebased.

asset/asset.go
```diff
@@ -2238,3 +2241,137 @@ type ChainAsset struct {
...
+
+// encodeAltLeafRecords determines the set of non-nil records to include when
+// encoding an AltLeaf. Since the Genesis, Group Key, Amount, and Version fields
+// are static, we can omit those fields.
+func (a *Asset) encodeAltLeafRecords() []tlv.Record {
+	records := make([]tlv.Record, 0, 3)
+
+	// Always use the normal witness encoding, since the asset version is
+	// always V0.
+	if len(a.PrevWitnesses) > 0 {
```
Roasbeef, line 2348: Will the alt leaf version ever have a use for the prev witness value?

tappsbt/interface.go
```diff
@@ -104,6 +106,9 @@ type bip32DerivationPredicate func(*psbt.Bip32Derivation) bool
 // BIP-0032 derivation paths.
 type taprootBip32DerivationPredicate func(*psbt.TaprootBip32Derivation) bool
 
+// AltLeafAsset is an AltLeaf backed by an Asset object.
+type AltLeafAsset = asset.AltLeaf[*asset.Asset]
```
Roasbeef, line 110: 👍

internal/test/helpers.go
```diff
@@ -98,7 +98,7 @@ func RandOp(t testing.TB) wire.OutPoint {
 	return op
 }
 
-func RandPrivKey(_ testing.TB) *btcec.PrivateKey {
```
Roasbeef, line 101: 👍

asset/asset_test.go
```diff
@@ -514,6 +516,79 @@ func TestAssetEncoding(t *testing.T) {
...
 
+// TestAltLeafEncoding runs a property test for AltLeaf validation, encoding,
+// and decoding.
+func TestAltLeafEncoding(t *testing.T) {
+	t.Run("alt leaf encode/decode", rapid.MakeCheck(testAltLeafEncoding))
+}
+
+// testAltLeafEncoding tests the AltLeaf validation logic, and that a valid
+// AltLeaf can be encoded and decoded correctly.
+func testAltLeafEncoding(t *rapid.T) {
```
Roasbeef, line 527: 👏

asset/generators.go
```diff
@@ -0,0 +1,319 @@
...
+
+	// MaxTxWitnessElementSize is the maximum size of a single object in a
+	// generated witness stack.
+	MaxTxWitnessElementSize = 64
+)
+
+var (
+	// Simple generators.
+
+	VersionGen = rapid.SampledFrom([]Version{V0, V1})
```
Roasbeef, line 35: Nice! Was writing some tests with `rapid` in another PR and nearly reached to write custom gens for everything, but decided not to, but glad you did

Roasbeef (approved):
LGTM 🦚

Nice work! Rapid `testdata` dir adds a bit of bloat, but I think in the short term, it's best to continue committing the test cases to the repo. In the future, we may end up reaching for an external corpus like we do with fuzzing stuff, or perhaps there's a way to _minimize_ the test cases it adds after the fact.

asset/asset.go
jharveyb: Hmm I don't expect it to, but that's also the only field left here that's 'consensus' wrt. being handled by parts of `tapd`, compared to the unknown odd field. Which may be useful for future features.

Given the size limits for `[]AltLeaves` I'm not too worried about it causing dysfunction.

jharveyb:
> In the future, we may end up reaching for an external corpus like we do with fuzzing stuff, or perhaps there's a way to _minimize_ the test cases it adds after the fact.

Yeah, the framework has some settings around minimization but I think that does not account for the size of the input, but maybe how quickly the code under test fails?

The default shrinking time is 30 seconds, but reaching failing cases was much faster than that. I'm not sure exactly how we can shrink the fail cases, especially which structs that have value fields and not pointers.

Issue #1168, opened 2024-11-01
[bug]: Not all transfer proofs generated


youngzhenhao:
### Background

Only a portion of proofs is generated locally after SendAsset to 12 addresses, this is also true after 6 blocks have been generated in bitcoin mainnet.

Bitcoin of asset's this UTXO  has actually been spent, but asset's this UTXO is still existing, can not be used, and no output index 0's asset UTXO available.

### Your environment

```sh
litd: 0.13.3-alpha
lnd: v0.18.2-beta
tapd: 0.4.1-alpha
network: mainnet
system: Linux 4.19.157
lnd use neutrino
```
* any other relevant environment details

### Steps to reproduce

1. Create a new instance of the lightning-terminal daemon and run it.
2. Unlock lnd and wait it synced to chain.
3. Call SendAsset of taprpc to send assets to 12 addresses are from different tapd.
4. Wait several new block generates.
5. Check proof files of transfer in directory.

### Expected behavior

All proof file with output index 0 to 12 are generated in the `{tapddir}\data\mainnet\proofs\{asset_id}\`.

### Actual behavior

Only proof file with output index 0 was generated in the `{tapddir}\data\mainnet\proofs\{asset_id}`.

Logs [here](https://github.com/youngzhenhao/backup-log-files/blob/main/2024-10-31/.lnd/logs/bitcoin/mainnet/lnd.log)

Proof file with index 0 [here](https://github.com/youngzhenhao/backup-log-files/blob/main/2024-10-31/.tapd/data/mainnet/proofs/47e9a860b665cc094c75b2c0d720aa950bafed5238a99f4edf8c6a1ad57d9a45/021a79c9b213e0eb1addaba079ae9119501ebe017d59ade215d783e1f66cc0c68e-3161fc663d42415b217dd56b8f727072-0.assetproof)

The same thing happened two weeks ago, in one of the previous transfers, until now, only proof files with output indexes 0,1,2 were generated after SendAsset to other 12 addresses.

### Expectation

1. What would cause only a portion of proof files be generated?
2. How to make asset's this invalid UTXO usable?
3. How to use the existing data to regenerate all the proof files?

guggero:
I see this:
```
2024-10-31 09:12:07.415 [ERR] FRTR: Error evaluating state (SendStateStoreProofs): error verifying proof: error verifying proofs: invalid exclusion proof: invalid taproot proof: derived_keys=([]asset.SerializedKey) (len=2 cap=2) {
[...]

PR #1181, opened 2024-11-08
[tapsend]: Enforce unique script keys

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1168 (or at least prevents it from happening in the future).

This PR enforces script keys to be unique per top-level MS-SMT tree (assetID/groupKey) within a single transfer.

To be able to test this properly, we also make sure selected UTXOs are released by the freighter if the shipment of a parcel results in an error.
That change then impacts some assumptions in the integration tests, which requires additional commits to fix.

### DIFF
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/tapdevrpc"
 	unirpc "github.com/lightninglabs/taproot-assets/taprpc/universerpc"
+	"github.com/lightninglabs/taproot-assets/tapsend"
 	"github.com/lightningnetwork/lnd/lntest/wait"
 	"github.com/stretchr/testify/require"
 )
@@ -101,6 +102,16 @@ func testBasicSendUnidirectional(t *harnessTest) {
 		AssetVersion: rpcAssets[0].Version,
 	})
 	require.NoError(t.t, err)
+
+	// Before we start sending, we test that we aren't allowed to send to
+	// the same address more than once within the same transfer.
+	_, err = t.tapd.SendAsset(ctxb, &taprpc.SendAssetRequest{
+		TapAddrs: []string{
+			bobAddr.Encoded,
+			bobAddr.Encoded,
+		},
+	})
+	require.ErrorContains(t.t, err, tapsend.ErrDuplicateScriptKeys.Error())
 
 	for i := 0; i < numSends; i++ {
 		t.t.Logf("Performing send procedure: %d", i)
--- a/tapfreighter/coin_select.go
+++ b/tapfreighter/coin_select.go
@@ -67,8 +67,14 @@ func (s *CoinSelect) SelectCoins(ctx context.Context,
 		return nil, ErrMatchingAssetsNotFound
 	}
 
-	log.Infof("Identified %v eligible asset inputs for send of %d to %v",
-		len(eligibleCommitments), constraints.MinAmt, constraints)
+	anchorInputs := fn.Map(
+		eligibleCommitments, func(c *AnchoredCommitment) string {
+			return c.AnchorPoint.String()
+		},
+	)
+	log.Infof("Identified %v eligible asset inputs for send of %d to %v: "+
+		"%v", len(anchorInputs), constraints.MinAmt,
+		constraints.String(), anchorInputs)
 
 	// Only select coins anchored in a compatible commitment.
 	compatibleCommitments := fn.Filter(
--- a/tapsend/send.go
+++ b/tapsend/send.go
@@ -130,6 +130,14 @@ var (
 	ErrNoRootLocator = errors.New(
 		"cannot create split commitment without split root output",
 	)
+
+	// ErrDuplicateScriptKeys is an error that is returned when the outputs
+	// of a virtual packet contain duplicate script keys.
+	ErrDuplicateScriptKeys = errors.New(
+		"send: duplicate script keys in outputs - cannot re-use same " +
+			"script key in same transaction (e.g. cannot send to " +
+			"same address more than once)",
+	)
 )
 
 var (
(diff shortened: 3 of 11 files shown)


guggero:
I'll look into the failed itest on Monday, looks like a test is using duplicate keys.

tapsend/send.go
```diff
@@ -130,6 +130,14 @@ var (
 	ErrNoRootLocator = errors.New(
 		"cannot create split commitment without split root output",
 	)
+
```
Roasbeef, line 133: In commit message: collidingon -> colliding

tapsend/send.go
```diff
@@ -364,6 +372,46 @@ func ValidateInputs(inputCommitments tappsbt.InputCommitments,
...
+		perAssetMap, ok := scriptKeys[tapCommitmentKey]
+		if !ok {
+			perAssetMap = make(map[asset.SerializedKey]struct{})
+			scriptKeys[tapCommitmentKey] = perAssetMap
+		}
+
+		for idx := range vPkt.Outputs {
+			vOut := vPkt.Outputs[idx]
+
+			// Check if the script key has already been used in this
```
Roasbeef, line 401: Isn't it ok for us to have a script key duplicate _across_ Bitcoin transaction outputs?

tapsend/send_test.go
```diff
@@ -2952,3 +2952,151 @@ func TestValidateAnchorInputs(t *testing.T) {
...
+			Outputs: make([]*tappsbt.VOutput, len(outputKeys)),
+		}
+		var a asset.Asset
+		if groupKey != nil {
+			a.GroupKey = groupKey
+		}
+		vPkt.SetInputAsset(0, &a)
+
+		for i, outputKey := range outputKeys {
+			vPkt.Outputs[i] = &tappsbt.VOutput{
```
Roasbeef, line 2995: Shouldn't the output index be considered here when determining uniqueness?

tapsend/send.go
jharveyb: I think a duplicate script key for the same asset would be an issue, even without those assets being children of an asset split. If that occurred, the exclusion proof would be a proof of inclusion, but for a _different_ leaf vs. the existing model where the exclusion proof terminates in a nil leaf.

tapsend/send.go
```diff
@@ -364,6 +372,46 @@ func ValidateInputs(inputCommitments tappsbt.InputCommitments,
 	return isFullValueSpend, nil
 }
 
+// ValidateScriptKeysUnique makes sure the outputs of a set of virtual packets
+// don't lead to collisions in and of the trees (e.g. two asset outputs with the
+// same script key in the same anchor output) or with exclusion proofs (e.g. two
+// asset outputs with the same script key in different anchor outputs).
+func ValidateScriptKeysUnique(packets []*tappsbt.VPacket) error {
+	// We'll keep track of all script keys (per top commitment key) that
```
jharveyb, line 380: :100:

tapsend/send_test.go
```diff
@@ -2952,3 +2952,151 @@ func TestValidateAnchorInputs(t *testing.T) {
...
+				makeVPacket(
+					assetID2, groupKey1,
+					[]asset.ScriptKey{
+						scriptKey3,
+					},
+				),
+			},
+		},
+		{
+			name: "collision, same group packets, same keys",
```
jharveyb, line 3071: would be good to test collision based on asset ID + script key in addition to group key + script key.

tapfreighter/coin_select.go
```diff
@@ -67,8 +67,14 @@ func (s *CoinSelect) SelectCoins(ctx context.Context,
...
 	}
 
-	log.Infof("Identified %v eligible asset inputs for send of %d to %v",
-		len(eligibleCommitments), constraints.MinAmt, constraints)
+	anchorInputs := fn.Map(
+		eligibleCommitments, func(c *AnchoredCommitment) string {
+			return c.AnchorPoint.String()
+		},
+	)
+	log.Infof("Identified %v eligible asset inputs for send of %d to %v: "+
```
jharveyb, line 75: nit: this is counting anchor inputs and not asset inputs?

IIUC the anchor inputs may contain multiple eligible asset inputs, which we will select from later.

tapsend/send_test.go
guggero: No, that's kind of the point of this fix. See comment(s) above.

itest/send_test.go
```diff
@@ -102,6 +103,16 @@ func testBasicSendUnidirectional(t *harnessTest) {
 	})
 	require.NoError(t.t, err)
 
+	// Before we start sending, we test that we aren't allowed to send to
+	// the same address more than once within the same transfer.
+	_, err = t.tapd.SendAsset(ctxb, &taprpc.SendAssetRequest{
```
jharveyb, line 108: 👍🏽

jharveyb (approved):
Looks solid, thanks for catching this!

We may want to add similar checks earlier in the freighter eventually but this should address the root cause IMO.

gijswijs (approved):
Very nice!

I feel that the tapfreighter commits should be their own PR? They don't have anything to do with enforcing unique script keys. Same thing goes fee estimation itest.

Or maybe just mention it in the PR that a number of other fixes were implemented as well, just for future reference.

tapsend/send.go
```diff
@@ -364,6 +372,46 @@ func ValidateInputs(inputCommitments tappsbt.InputCommitments,
...
+			scriptKeys[tapCommitmentKey] = perAssetMap
+		}
+
+		for idx := range vPkt.Outputs {
+			vOut := vPkt.Outputs[idx]
+
+			// Check if the script key has already been used in this
+			// transaction.
+			scriptKey := asset.ToSerialized(vOut.ScriptKey.PubKey)
+			if _, ok := perAssetMap[scriptKey]; ok {
```
gijswijs, line 404: Let me see if I understand this:
In theory it would be ok to reuse scriptKeys over different tapCommitmentKeys?

jharveyb: Yes, since those would have different paths in the anchored `tapCommitment`.

I think this statement is correct:

"For a given transfer, across all anchor outputs, assets can share a `TapCommitmentKey()`, or an `AssetCommitmentKey()`, but not both."

If both are shared, for any two assets, then we can't generate valid exclusion proofs.

tapsend/send.go
Roasbeef: Hmm yeh, I do wonder if there's a small change we can make the the exclusion proofs here to patch this portion.

tapsend/send_test.go
guggero: Thanks, added two more test cases.

tapfreighter/coin_select.go
guggero: I think if we have multiple eligible assets in the same anchor, we'd still get multiple `*AnchoredCommitment` structs returned. So IMO this is correct.

guggero:
> They don't have anything to do with enforcing unique script keys. Same thing goes fee estimation itest.

Actually, they kind of do. To be able to test the check introduced with the PR, we want to be able to try again. That's why we added the fix for releasing the leased inputs. Which then changes a bunch of assumptions in the itests.
Updated the PR body and commit messages to make that more clear.

guggero:
Thanks to the failing integration test I noticed two things that I needed to fix (with the latest push):
 - For grouped assets, the asset ID goes into the asset-level commitment key, not just the script key. So the commitment keys actually look like this (and uniqueness needs to be guaranteed across those within a single transfer):

|                   | TAP level           | asset level                        |
|-------------------|---------------------|------------------------------------|
| grouped asset     | `SHA256(group_key)` | `SHA256(asset_id \|\| script_key)` |
| non-grouped asset | `asset_id`          | `SHA256(script_key)`               |

 - When spending multiple input assets from the same on-chain input where other passive assets were present, we would duplicate those passive assets when creating passive virtual packets for them. This wasn't really an issue before since within the commitment trees they'd just be collapsed/overwritten into a single asset. But because the new uniqueness test operates on the slice of virtual packets, this was discovered. So a simple de-duplication fix was added in the last commit.

jharveyb:
>     * For grouped assets, the asset ID goes into the asset-level commitment key, not just the script key. So the commitment keys actually look like this (and uniqueness needs to be guaranteed across those within a single transfer):
> 
> 
> 	TAP level 	asset level
> grouped asset 	`SHA256(group_key)` 	`SHA256(asset_id || script_key)`
> non-grouped asset 	`asset_id` 	`SHA256(script_key)`

May be a motivation to use the `Specifier` type in more places.

guggero:
@jharveyb with the plan for getting rid of the collisions, I think we should go ahead and merge this. Would appreciate a review, thanks in advance!

jharveyb (approved):
Re-reviewed, LGTM 👍🏽

PR #1184, opened 2024-11-09
rpc: add asset hodl invoice support

Roasbeef:
In this commit, we add hodl invoice support when creating invoice for TAP channels. The invoice creation proceeds as normal, but a special HodlInvoice field is set which contains a payment hash which must be set.

When invoices are created using this endpoint. They MUST be settled using the SettleInvoice RPC on the invoicesrpc server in lnd.


rpcserver.go
```diff
@@ -7218,6 +7220,57 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+		}
+
+		payReq, err := r.cfg.Lnd.Invoices.AddHoldInvoice(
+			ctx, &invoicesrpc.AddInvoiceData{
+				Memo:            iReq.Memo,
+				Value:           lnwire.MilliSatoshi(iReq.ValueMsat),
+				Hash:            &payHash,
+				DescriptionHash: iReq.DescriptionHash,
+				Expiry:          iReq.Expiry,
+				Private:         true,
```
guggero, line 7255: Should we take `private` from `iReq` as well?

guggero (approved):
Linter isn't happy about line length and one question about a flag, otherwise LGTM :tada:

rpcserver.go
GeorgeTsagk: not sure of the side effects, but since we're here we assume we're using the scid hophint flow?

is it even a valid move from user perspective to set `Private: false`?

GeorgeTsagk (approved):
Lgtm

rpcserver.go
Roasbeef: So revisiting things, I think we actually want to set `Private` to false: https://github.com/lightningnetwork/lnd/blob/4f6b510869abd03a2defabd57257b0bc94f4e557/lnrpc/invoicesrpc/addinvoice.go#L442-L460

Notice how if `Private` is true, then we'll try to continue adding hop hints up the max limit. However, in this case, we only care about our singular hop hint, as we want to make sure the incoming HTLC goes over the specified channel. 

Pushed a fix up to set this to false.

ZZiigguurraatt:
why call it hodl invoice whereas everywhere in LND documentation hold invoice is used?

ZZiigguurraatt:
Why do you have a separate `HodlInvoice` (https://lightning.engineering/api-docs/api/taproot-assets/taproot-asset-channels/add-invoice#tapchannelrpchodlinvoice) object with just a `payment_hash` in it? Do you expect to add more fields? LND's `AddHoldInvoice` (https://lightning.engineering/api-docs/api/lnd/invoices/add-hold-invoice) does not have this.

Also, `AddHoldInvoice` in LND uses `hash` instead of `payment_hash`. For new users, it is confusing if there is a difference. I think they should have common names everywhere. I think `payment_hash` is better to use than just `hash`.

ZZiigguurraatt:
Also, at https://lightning.engineering/api-docs/api/taproot-assets/taproot-asset-channels/add-invoice/index.html#tapchannelrpcaddinvoicerequest it says to use an `invoice_request` generated from an  `Invoice` (https://lightning.engineering/api-docs/api/lnd/lightning/add-invoice/index.html#lnrpcinvoice). Do we use this one for a HOLD invoice too, or an `AddHoldInvoiceRequest` (https://lightning.engineering/api-docs/api/lnd/invoices/add-hold-invoice#invoicesrpcaddholdinvoicerequest)? Note: `AddHoldInvoiceRequest` has the `hash` field, so that would be defined twice if it were to be used.

PR #1185, opened 2024-11-09
tapdb: fix bug w.r.t null bool handling, allow InsertScriptKey to allow flipping known to true 

Roasbeef:
In this commit, we fix some incorrect handling on null bool in SQL, that
looks to be mostly benign. Before this commit, instead of doing `.Bool`,
we did `.Valid`. The latter is only useful to check if something is
zero, or NULL. In this case, we can just go with the bool value
directly.

Then, we update `InsertScriptKey` to allow flipping known to
true, if it was false before.

This is useful as at times a script key from a smart contract might have
been inserting on disk, but with declared known as false. Then if we
tried to insert it again, with known as true, the upsert logic would end
up making no change on disk.

Note that without the initial bug fix, the test added in the last commit 
would fail.

### DIFF
--- a/tapdb/addrs.go
+++ b/tapdb/addrs.go
@@ -467,7 +467,9 @@ func (t *TapAddressBook) QueryAddrs(ctx context.Context,
 				return fmt.Errorf("unable to make addr: %w", err)
 			}
 
-			declaredKnown := addr.ScriptKeyDeclaredKnown.Valid
+			declaredKnown := extractBool(
+				addr.ScriptKeyDeclaredKnown,
+			)
 			addrs = append(addrs, address.AddrWithKeyInfo{
 				Tap: tapAddr,
 				ScriptKeyTweak: asset.TweakedScriptKey{
@@ -620,9 +622,11 @@ func fetchAddr(ctx context.Context, db AddrBook, params *address.ChainParams,
 	return &address.AddrWithKeyInfo{
 		Tap: tapAddr,
 		ScriptKeyTweak: asset.TweakedScriptKey{
-			RawKey:        scriptKeyDesc,
-			Tweak:         dbAddr.ScriptKeyTweak,
-			DeclaredKnown: dbAddr.ScriptKeyDeclaredKnown.Valid,
+			RawKey: scriptKeyDesc,
+			Tweak:  dbAddr.ScriptKeyTweak,
+			DeclaredKnown: extractBool(
+				dbAddr.ScriptKeyDeclaredKnown,
+			),
 		},
 		InternalKeyDesc:  internalKeyDesc,
 		TaprootOutputKey: *taprootOutputKey,
--- a/tapdb/addrs.go
+++ b/tapdb/addrs.go
@@ -689,6 +689,7 @@ func (t *TapAddressBook) InsertScriptKey(ctx context.Context,
 			return fmt.Errorf("error inserting internal key: %w",
 				err)
 		}
+
 		_, err = q.UpsertScriptKey(ctx, NewScriptKey{
 			InternalKeyID:    internalKeyID,
 			TweakedScriptKey: scriptKey.PubKey.SerializeCompressed(),
--- a/tapdb/sqlc/queries/assets.sql
+++ b/tapdb/sqlc/queries/assets.sql
@@ -854,7 +854,15 @@ INSERT INTO script_keys (
 )  ON CONFLICT (tweaked_script_key)
     -- As a NOP, we just set the script key to the one that triggered the
     -- conflict.
-    DO UPDATE SET tweaked_script_key = EXCLUDED.tweaked_script_key
+    DO UPDATE SET 
+      tweaked_script_key = EXCLUDED.tweaked_script_key,
+      -- If the script key was previously unknown, we'll update to the new
+      -- value.
+      declared_known = CASE
+                         WHEN script_keys.declared_known IS NULL OR script_keys.declared_known = FALSE
+                         THEN COALESCE(EXCLUDED.declared_known, script_keys.declared_known)
+                         ELSE script_keys.declared_known
+                       END
 RETURNING script_key_id;
 
 -- name: FetchScriptKeyIDByTweakedKey :one
--- a/tapdb/addrs_test.go
+++ b/tapdb/addrs_test.go
@@ -3,6 +3,7 @@ package tapdb
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"math/rand"
 	"testing"
 	"time"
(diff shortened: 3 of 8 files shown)


tapdb/addrs.go
```diff
@@ -467,7 +467,9 @@ func (t *TapAddressBook) QueryAddrs(ctx context.Context,
 				return fmt.Errorf("unable to make addr: %w", err)
 			}
 
-			declaredKnown := addr.ScriptKeyDeclaredKnown.Valid
+			declaredKnown := extractBool(
```
guggero, line 473: Oops, don't know what I was thinking at that moment... Nice catch!

tapdb/sqlc/queries/assets.sql
```diff
@@ -857,6 +857,10 @@ INSERT INTO script_keys (
     DO UPDATE SET tweaked_script_key = EXCLUDED.tweaked_script_key
 RETURNING script_key_id;
 
+-- name: DeclareScriptKeyKnown :exec
+UPDATE script_keys
+SET declared_known = TRUE
+WHERE tweaked_script_key = $1;
```
guggero, line 863: nit: add empty line between queries?

tapdb/addrs.go
```diff
@@ -681,22 +684,41 @@ func (t *TapAddressBook) InsertScriptKey(ctx context.Context,
...
-		return err
+		if err != nil {
+			return fmt.Errorf("error inserting script key: %w", err)
+		}
+
+		// This key might already exist, so we'll make sure that
+		// declared_known is set to true. Otherwise, if the key already
+		// existed, and we need to view it as ours, we'll fail to do so.
+		if declaredKnown {
+			return q.DeclareScriptKeyKnown(
```
guggero, line 710: Argh, another oversight on my part... An alternative fix (and perhaps the one expected) could be to update the `ON CONFLICT` clause to set

```
  -- This is not a NOP, we overwrite the declared_known value.
  DO UPDATE SET  tweaked_script_key = EXCLUDED.tweaked_script_key,
                               declared_known = EXCLUDED.declared_known
```

guggero (commented):
Thanks a lot for the fix!

tapdb/addrs.go
jharveyb: I was thinking the same - in that case idk if a separate query would be useful?

jharveyb:
> In this commit, we fix some incorrect handling on null bool in SQL, that looks to be mostly benign. Before this commit, instead of doing `.Bool`, we did `.Valid`. The latter is only useful to check if something is zero, or NULL. In this case, we can just go with the bool value directly.

This only works for bools that are always set? Vs. other bools that are optional.

I think this would be correct for any spot where we insert data created with `sqlBool()`.

jharveyb (commented):
Looks reasonable, though I wonder if we can have an equivalent fix just by changing the existing query; commented separately.

tapdb/addrs.go
Roasbeef: Thx! Likely need to do some more analysis, but in most cases this would've just resulted in an RPC display being wrong. There is the case though when we decide to create a transfer output or not, that relied on this erroneous check.

tapdb/addrs.go
Roasbeef: Mhmm, we could also go a step further and also allow a tweak to be updated as well. 

IIUC, we have some other logic that skips an asset in the default balance set if the asset has a script key tweak. In my testing, that wasn't triggering as we'd insert the transfer for the force close (insert script keys w/o tweak), then I'd want to insert the script keys again before trying to make contract resolutions (which failed to update). 

My rationale for doing it with this new method was that it only affected the call path I was concerned with. Will switch to the more general method and see if that breaks anything elsewhere. Agree the upsert extension is cleaner.

tapdb/sqlc/queries/assets.sql
```diff
@@ -854,7 +854,14 @@ INSERT INTO script_keys (
 )  ON CONFLICT (tweaked_script_key)
     -- As a NOP, we just set the script key to the one that triggered the
     -- conflict.
-    DO UPDATE SET tweaked_script_key = EXCLUDED.tweaked_script_key
+    DO UPDATE SET 
+      tweaked_script_key = EXCLUDED.tweaked_script_key,
+      -- If the script key was previously unknown, we'll update to the new
+      -- value.
+      declared_known = CASE
+                         WHEN script_keys.declared_known = FALSE THEN EXCLUDED.declared_known
```
Roasbeef, line 862: Opted for this case statement to make sure that if the ordering gets reversed, we don't end up making something as unknown after it was marked as known.

Roasbeef:
@guggero @jharveyb PTAL.

tapdb/sqlc/queries/assets.sql
guggero: Shouldn't this be `WHEN script_keys.declared_known != TRUE`? Otherwise we won't set it when it's `NULL`.

Roasbeef: So I wrote a test to try to make things fail before fixing (pushed up now), but but it still passed. I copied some code so I could insert a `NULL` for the bool, but without any change, the extra case still passed.

Roasbeef: Ah, so we don't allow [it to be null in the first place](https://github.com/lightninglabs/taproot-assets/blob/7c1cd9a32937415d1014bb16b39d4c4512bfa713/tapdb/sqlc/migrations/000018_script_key_declared_known.up.sql#L5), just at times the response after a join can be NULL, and a query might use `sqlc.narg`. Either way I updated to the new version as it's more explicit, and slightly more compact.

Roasbeef: Nvm, my test was missing a statement 🤦 (fails when that missing statement is added). I also misinterpreted the initial migration: you need to specify `NOT NULL` otherwise it's allowed. 

Updated, and now accounts for `NULL`. Test passes!

guggero (approved):
LGTM :tada:

jharveyb (approved):
LGTM 💯

tapdb/sqlc/queries/assets.sql
```diff
@@ -854,7 +854,15 @@ INSERT INTO script_keys (
 )  ON CONFLICT (tweaked_script_key)
     -- As a NOP, we just set the script key to the one that triggered the
     -- conflict.
-    DO UPDATE SET tweaked_script_key = EXCLUDED.tweaked_script_key
+    DO UPDATE SET 
+      tweaked_script_key = EXCLUDED.tweaked_script_key,
+      -- If the script key was previously unknown, we'll update to the new
+      -- value.
+      declared_known = CASE
```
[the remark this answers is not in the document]
guggero: I don't think we'd ever want to "un-know" a key. So I think as a safety feature this makes a lot of sense.

PR #1186, opened 2024-11-11
Track HTLCs in rfq policies

GeorgeTsagk:
## Description

This PR adds a simple in-memory tracking of accepted htlcs in the RFQ policy level. HTLCs that passed the policy check are now tracked by each policy, possibly affecting future policy check outcomes (depends on individual policy's implementation)

### DIFF
--- a/chain_bridge.go
+++ b/chain_bridge.go
@@ -18,6 +18,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/tapgarden"
 	"github.com/lightningnetwork/lnd/chainntnfs"
 	"github.com/lightningnetwork/lnd/funding"
+	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/routing/route"
@@ -376,9 +377,19 @@ func (l *LndRouterClient) DeleteLocalAlias(ctx context.Context, alias,
 	return l.lnd.Router.XDeleteLocalChanAlias(ctx, alias, baseScid)
 }
 
+// SubscribeHtlcEvents subscribes to a stream of events related to
+// HTLC updates.
+func (l *LndRouterClient) SubscribeHtlcEvents(
+	ctx context.Context) (<-chan *routerrpc.HtlcEvent,
+	<-chan error, error) {
+
+	return l.lnd.Router.SubscribeHtlcEvents(ctx)
+}
+
 // Ensure LndRouterClient implements the rfq.HtlcInterceptor interface.
 var _ rfq.HtlcInterceptor = (*LndRouterClient)(nil)
 var _ rfq.ScidAliasManager = (*LndRouterClient)(nil)
+var _ rfq.HtlcSubscriber = (*LndRouterClient)(nil)
 
 // LndInvoicesClient is an LND invoices RPC client.
 type LndInvoicesClient struct {
--- a/rfq/order.go
+++ b/rfq/order.go
@@ -14,6 +14,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/rfqmath"
 	"github.com/lightninglabs/taproot-assets/rfqmsg"
 	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
 	"github.com/lightningnetwork/lnd/lnutils"
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwire"
@@ -842,4 +843,13 @@ type HtlcInterceptor interface {
 	// InterceptHtlcs intercepts HTLCs, using the handling function provided
 	// to respond to HTLCs.
 	InterceptHtlcs(context.Context, lndclient.HtlcInterceptHandler) error
+}
+
+// HtlcSubscriber is an interface that contains the function necessary for
+// retrieving live HTLC event updates.
+type HtlcSubscriber interface {
+	// SubscribeHtlcEvents subscribes to a stream of events related to
+	// HTLC updates.
+	SubscribeHtlcEvents(ctx context.Context) (<-chan *routerrpc.HtlcEvent,
+		<-chan error, error)
 }
--- a/rfq/order.go
+++ b/rfq/order.go
@@ -13,6 +13,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightninglabs/taproot-assets/rfqmath"
 	"github.com/lightninglabs/taproot-assets/rfqmsg"
+	"github.com/lightningnetwork/lnd/channeldb/models"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
 	"github.com/lightningnetwork/lnd/lnutils"
@@ -71,6 +72,14 @@ type Policy interface {
 	// which the policy applies.
 	Scid() uint64
 
+	// TrackAcceptedHtlc makes the policy aware of this new accepted HTLC.
+	// This is important in cases where the set of existing HTLCs may affect
+	// whether the next compliance check passes.
+	TrackAcceptedHtlc(circuitKey models.CircuitKey, amt lnwire.MilliSatoshi)
+
+	// UntrackHtlc stops tracking the uniquely identified HTLC.
+	UntrackHtlc(circuitKey models.CircuitKey)
+
 	// GenerateInterceptorResponse generates an interceptor response for the
 	// HTLC interceptor from the policy.
 	GenerateInterceptorResponse(
@@ -95,8 +104,21 @@ type AssetSalePolicy struct {
 	// the policy.
 	MaxOutboundAssetAmount uint64
 
+	// CurrentAssetAmountMsat is the total amount that is held currently in
+	// accepted HTLCs.
+	CurrentAmountMsat lnwire.MilliSatoshi
+
+	// stateMutex is a mutex that locks access to this policy's internal
+	// state. This is needed as state is updated asynchronously by each
+	// routine that handles an intercepted HTLC.
+	stateMutex sync.RWMutex
+
 	// AskAssetRate is the quote's asking asset unit to BTC conversion rate.
 	AskAssetRate rfqmath.BigIntFixedPoint
+
+	// htlcToAmt maps the unique HTLC identifiers to the effective amount
+	// that they carry.
+	htlcToAmt map[models.CircuitKey]lnwire.MilliSatoshi
 
 	// expiry is the policy's expiry unix timestamp after which the policy
 	// is no longer valid.
@@ -105,12 +127,15 @@ type AssetSalePolicy struct {
 
 // NewAssetSalePolicy creates a new asset sale policy.
 func NewAssetSalePolicy(quote rfqmsg.BuyAccept) *AssetSalePolicy {
+	htlcToAmtMap := make(map[models.CircuitKey]lnwire.MilliSatoshi)
+
 	return &AssetSalePolicy{
 		AssetSpecifier:         quote.Request.AssetSpecifier,
 		AcceptedQuoteId:        quote.ID,
 		MaxOutboundAssetAmount: quote.Request.AssetMaxAmt,
 		AskAssetRate:           quote.AssetRate.Rate,
 		expiry:                 uint64(quote.AssetRate.Expiry.Unix()),
+		htlcToAmt:              htlcToAmtMap,
 	}
 }
 
@@ -125,10 +150,15 @@ func NewAssetSalePolicy(quote rfqmsg.BuyAccept) *AssetSalePolicy {
 func (c *AssetSalePolicy) CheckHtlcCompliance(
 	htlc lndclient.InterceptedHtlc) error {
 
+	// Since we will be reading CurrentAmountMsat value we acquire a read
+	// lock.
+	c.stateMutex.RLock()
+	defer c.stateMutex.RUnlock()
+
 	// Check that the channel SCID is as expected.
 	htlcScid := SerialisedScid(htlc.OutgoingChannelID.ToUint64())
 	if htlcScid != c.AcceptedQuoteId.Scid() {
-		return fmt.Errorf("htlc outgoing channel ID does not match "+
+		return fmt.Errorf("HTLC outgoing channel ID does not match "+
 			"policy's SCID (htlc_scid=%d, policy_scid=%d)",
 			htlcScid, c.AcceptedQuoteId.Scid())
 	}
@@ -152,8 +182,8 @@ func (c *AssetSalePolicy) CheckHtlcCompliance(
 		maxAssetAmount, c.AskAssetRate,
 	)
 
-	if htlc.AmountOutMsat > policyMaxOutMsat {
-		return fmt.Errorf("htlc out amount is greater than the policy "+
+	if (c.CurrentAmountMsat + htlc.AmountOutMsat) > policyMaxOutMsat {
+		return fmt.Errorf("HTLC out amount is greater than the policy "+
 			"maximum (htlc_out_msat=%d, policy_max_out_msat=%d)",
 			htlc.AmountOutMsat, policyMaxOutMsat)
 	}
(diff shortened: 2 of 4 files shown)


rfq/order.go
```diff
@@ -166,6 +175,12 @@ func (c *AssetSalePolicy) CheckHtlcCompliance(
 	return nil
 }
 
+// TrackAcceptedHtlc accounts for the newly accepted htlc. This may affect the
+// acceptance of future htlcs.
+func (c *AssetSalePolicy) TrackAcceptedHtlc(htlc lndclient.InterceptedHtlc) {
+	c.CurrentAmountMsat += htlc.AmountOutMsat
```
guggero, line 181: Not sure if these need to be atomic values? Since multiple goroutines could manipulate them concurrently.

rfq/order.go
```diff
@@ -314,7 +333,7 @@ func (c *AssetPurchasePolicy) CheckHtlcCompliance(
 
 	// Ensure that the outbound HTLC amount is less than the maximum agreed
 	// BTC payment.
-	if htlc.AmountOutMsat > c.PaymentMaxAmt {
+	if (c.CurrentAmountMsat + htlc.AmountOutMsat) > c.PaymentMaxAmt {
```
guggero, line 336: This would work for a single attempt. But in case where multiple attempts over the same edge node are done, this would quickly result in the HTLCs being rejected, because we would never reduce the current amount based on the HTLCs that were canceled from previous attempts.

guggero (changes requested):
As mentioned offline, we also need to track and take into account canceled HTLCs from previous attempts.
These can be subscribed to with https://lightning.engineering/api-docs/api/lnd/router/subscribe-htlc-events.

ffranr (commented):
I'm up-to-date with this PR. I agree with Oli's comments.

GeorgeTsagk:
@ffranr @guggero 
I updated the PR to now include an htlc event subscription. 
We now maintain a map of htlc circuit keys to rfq policy
When an htlc forward is failed we retrieve the htlc circuit key, fetch the corresponding policy and have it to stop tracking the htlc.

ffranr:
> Will also see how to add some coverage here

@GeorgeTsagk should i review this before you add coverage or wait for that extra changes?

GeorgeTsagk:
> > Will also see how to add some coverage here
> 
> @GeorgeTsagk should i review this before you add coverage or wait for that extra changes?

You can review
Coverage will soon be added in the LitD itests

tapchannel/aux_leaf_signer.go
```diff
@@ -735,3 +737,82 @@ func (v *schnorrSigValidator) validateSchnorrSig(virtualTx *wire.MsgTx,
...
+	// index. Otherwise, we'd multiply G by zero.
+	index++
+
+	// If we wrapped around from math.MaxUint64 to 0, we need to make sure
+	// the tweak is 1 to not cause a multiplication by zero. This should
+	// never happen, as it would mean we have more than math.MaxUint64
+	// updates in a channel, which exceeds the protocol's maximum.
+	if index == 0 {
+		return new(secp256k1.ModNScalar).SetInt(1)
+	}
```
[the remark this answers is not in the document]
ffranr: Also, on the function name, perhaps `ScriptKeyTweakFromHtlcIndex`.

tapchannel/aux_leaf_signer.go
```diff
@@ -735,3 +737,82 @@ func (v *schnorrSigValidator) validateSchnorrSig(virtualTx *wire.MsgTx,
...
+	return indexAsScalar
+}
+
+// TweakPubKeyWithIndex tweaks the given internal public key with the given
+// HTLC index. The tweak is derived from the index in a way that never results
+// in a zero tweak. The value of 1 is always added to the index to make sure
+// this value is always non-zero. The public key is tweaked like this:
+//
+//	tweakedKey = key + (index+1) * G
+func TweakPubKeyWithIndex(pubKey *btcec.PublicKey,
```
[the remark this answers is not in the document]
GeorgeTsagk: this was from a diff PR this was based on, ignore

tapchannel/aux_leaf_signer_test.go
```diff
@@ -394,3 +396,61 @@ func TestTweakHtlcTree(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tweak := AddTweakWithIndex(tc.tweak, tc.index)
+			resultBytes := tc.result.Bytes()
+			resultBigInt := new(big.Int).SetBytes(resultBytes[:])
+			tweakBigInt := new(big.Int).SetBytes(tweak)
+			// 36893488147419103230
+			// 55340232221128654845
```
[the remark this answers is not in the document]
GeorgeTsagk: this was from a diff PR this was based on, ignore

tapchannel/aux_leaf_signer.go
```diff
@@ -735,3 +737,82 @@ func (v *schnorrSigValidator) validateSchnorrSig(virtualTx *wire.MsgTx,
...
+	index input.HtlcIndex) *btcec.PublicKey {
+
+	// We need to operate on Jacobian points, which is just a different
+	// representation of the public key that allows us to do scalar
+	// multiplication.
+	var (
+		pubKeyJacobian, tweakTimesG, tweakedKey btcec.JacobianPoint
+	)
+	pubKey.AsJacobian(&pubKeyJacobian)
```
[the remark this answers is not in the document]
GeorgeTsagk: this was from a diff PR this was based on, ignore

chain_bridge.go
```diff
@@ -376,9 +377,17 @@ func (l *LndRouterClient) DeleteLocalAlias(ctx context.Context, alias,
 	return l.lnd.Router.XDeleteLocalChanAlias(ctx, alias, baseScid)
 }
 
+func (l *LndRouterClient) SubscribeHtlcEvents(
```
guggero, line 380: nit: requires Godoc comment (for the same method in the `HtlcSubscriber` interface as well).

rfq/order.go
```diff
@@ -95,9 +103,17 @@ type AssetSalePolicy struct {
...
+	// CurrentAssetAmountMsat is the total amount that is held currently in
+	// accepted htlcs.
+	CurrentAmountMsat lnwire.MilliSatoshi
+
 	// AskAssetRate is the quote's asking asset unit to BTC conversion rate.
 	AskAssetRate rfqmath.BigIntFixedPoint
 
+	// htlcToAmt maps the unique htlc identifiers to the effective amount
+	// that they carry.
+	htlcToAmt lnutils.SyncMap[string, lnwire.MilliSatoshi]
```
guggero, line 115: So both the `htlcToAmt` and `CurrentAmountMsat` need to be in sync. If we use an atomic value for one and a sync map for the other, then it's still possible for a _read_ from another goroutine to get a different value while we're updating things (e.g. the map was updated but the total not yet).

So I think because we have two separate values, we need a `sync.RWMutex` instead. But then we can use a normal map here at least.

rfq/order.go
```diff
@@ -152,7 +168,8 @@ func (c *AssetSalePolicy) CheckHtlcCompliance(
 		maxAssetAmount, c.AskAssetRate,
 	)
 
-	if htlc.AmountOutMsat > policyMaxOutMsat {
+	if (c.CurrentAmountMsat + htlc.AmountOutMsat) > policyMaxOutMsat {
+		// if htlc.AmountOutMsat > policyMaxOutMsat {
```
guggero, line 172: nit: commented out code.

rfq/order.go
```diff
@@ -338,6 +386,31 @@ func (c *AssetPurchasePolicy) CheckHtlcCompliance(
...
 }
 
+// TrackAcceptedHtlc accounts for the newly accepted htlc. This may affect the
+// acceptance of future htlcs.
+func (c *AssetPurchasePolicy) TrackAcceptedHtlc(
+	htlc lndclient.InterceptedHtlc) {
+
+	c.CurrentAmountMsat += htlc.AmountOutMsat
+
+	htlcIDStr := htlcIdentifierStr(
```
guggero, line 396: I think we can probably simplify the interface of this `TrackAcceptedHtlc` a bit by just passing in `circuitKey` and `amount`. Then it's IMO a bit more apparent what's happening (looks more like a map add and delete) and the policy doesn't need to know how the key is calculated.

rfq/order.go
```diff
@@ -167,6 +184,29 @@ func (c *AssetSalePolicy) CheckHtlcCompliance(
 	return nil
 }
 
+// TrackAcceptedHtlc accounts for the newly accepted htlc. This may affect the
+// acceptance of future htlcs.
+func (c *AssetSalePolicy) TrackAcceptedHtlc(htlc lndclient.InterceptedHtlc) {
+	c.CurrentAmountMsat += htlc.AmountOutMsat
+
+	htlcIDStr := htlcIdentifierStr(
```
guggero, line 192: We can use structs as map keys directly. And as long as there are no pointer values in a struct (meaning the struct can be directly compared with `==` without the pointers leading to side effects), things work out as expected. So we can just use the circuit key directly as the map key, no need to convert to string.

rfq/order.go
```diff
@@ -167,6 +184,29 @@ func (c *AssetSalePolicy) CheckHtlcCompliance(
...
+}
+
+// UntrackHtlc stops tracking the uniquely identified htlc.
+func (c *AssetSalePolicy) UntrackHtlc(htlcIDStr string) {
+	amt, found := c.htlcToAmt.LoadAndDelete(htlcIDStr)
+	if !found {
+		return
+	}
+
+	c.CurrentAmountMsat -= amt
```
guggero, line 207: Yeah, we definitely need a mutex here. Otherwise another goroutine might read this value while we're waiting on the `LoadAndDelete` above to complete.

rfq/order.go
```diff
@@ -95,9 +103,17 @@ type AssetSalePolicy struct {
 	// the policy.
 	MaxOutboundAssetAmount uint64
 
+	// CurrentAssetAmountMsat is the total amount that is held currently in
+	// accepted htlcs.
```
guggero, line 107: nit: use the same capitalization for HTLC everywhere? So `s/htlcs/HTLCs` here and throughout the file.

rfq/order.go
```diff
@@ -530,6 +626,12 @@ type OrderHandler struct {
 	// associated asset transaction policies.
 	policies lnutils.SyncMap[SerialisedScid, Policy]
 
+	// htlcToPolicy maps a unique htlc identifier encoded as a string, to
+	// the policy that applies to it. We need this map because for failed
+	// HTLCs we don't have the RFQ data available, so we need to cache this
+	// info.
+	htlcToPolicy lnutils.SyncMap[string, Policy]
```
guggero, line 633: Same here, can use the circuit key as the map key directly. Here a sync map is enough and we don't need a mutex, as it's just a single map that needs to be synced.

rfq/order.go
```diff
@@ -640,12 +753,64 @@ func (h *OrderHandler) mainEventLoop() {
...
+				event.IncomingChannelId, event.IncomingHtlcId,
+			)
+
+			switch {
+			case failEvent != nil:
+				fallthrough
+			case linkFail != nil:
+				// Fetch the policy that is related to this
+				// htlc.
+				policy, found :=
```
guggero, line 790: nit: format as:

```
				policy, found := h.htlcToPolicy.LoadAndDelete(
					htlcIDStr,
				)
```
instead.

guggero (commented):
Looks pretty good. Left a couple of comments to simplify the code and make it more concurrency safe.

rfq/order.go
GeorgeTsagk: > So both the htlcToAmt and CurrentAmountMsat need to be in sync

What do you mean by "in sync"? The `htlcToAmt` is only used to cache the htlc's effective amount when the time to "cancel" it comes (i.e call `UntrackHtlc`)

On the other hand, `CurrentAmountMsat` is incremented/decremented by multiple goroutines, so def agree that we have to guard that.

So could just use a mutex that protects `CurrentAmountMsat` alone

ffranr (commented):
I'm up to date on this. Will wait for the revised commits based on Oli's review.

GeorgeTsagk:
Thank you @ffranr @guggero 

I just pushed the commits that address your feedback, also rebased on latest #1224

rfq/order.go
```diff
@@ -585,13 +714,19 @@ func (h *OrderHandler) handleIncomingHtlc(_ context.Context,
...
 		log.Warnf("HTLC does not comply with policy: %v "+
-			"(htlc=%v, policy=%v)", err, htlc, policy)
+			"(HTLC=%v, policy=%v)", err, htlc, policy)
 
 		return &lndclient.InterceptedHtlcResponse{
 			Action: lndclient.InterceptorActionFail,
 		}, nil
 	}
 
+	h.htlcToPolicy.Store(htlc.IncomingCircuitKey, policy)
```
guggero, line 724: This is already done by `TrackAcceptedHtlc` below.

rfq/order.go
```diff
@@ -166,6 +194,32 @@ func (c *AssetSalePolicy) CheckHtlcCompliance(
...
 }
 
+// TrackAcceptedHtlc accounts for the newly accepted HTLC. This may affect the
+// acceptance of future HTLCs.
+func (c *AssetSalePolicy) TrackAcceptedHtlc(circuitKey models.CircuitKey,
+	amt lnwire.MilliSatoshi) {
+
+	c.CurrentAmountMsat += amt
+
+	c.currentAmountMutex.Lock()
```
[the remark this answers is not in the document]
GeorgeTsagk: oopsie

rfq/order.go
GeorgeTsagk: that's a diff map called `htlcToAmt`

rfq/order.go
```diff
@@ -639,12 +774,66 @@ func (h *OrderHandler) mainEventLoop() {
...
+				HtlcID: event.IncomingHtlcId,
+			}
+
+			switch {
+			case failEvent != nil:
+				fallthrough
+			case linkFail != nil:
+				// Fetch the policy that is related to this
+				// HTLC.
+				policy, found := h.htlcToPolicy.LoadAndDelete(
```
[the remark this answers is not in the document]
GeorgeTsagk: dif map called `htlcToAmt`

rfq/order.go
```diff
@@ -435,6 +533,27 @@ func (a *AssetForwardPolicy) CheckHtlcCompliance(
 	return nil
 }
 
+// TrackAcceptedHtlc accounts for the newly accepted HTLC. This may affect the
+// acceptance of future HTLCs.
+func (a *AssetForwardPolicy) TrackAcceptedHtlc(circuitKey models.CircuitKey,
+	amt lnwire.MilliSatoshi) {
+
+	// Track accepted HTLC in the incoming policy.
+	a.incomingPolicy.TrackAcceptedHtlc(circuitKey, amt)
```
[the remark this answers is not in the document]
GeorgeTsagk: Not necessary

Every policy needs to be atomic w.r.t it's own state, if we have two callers in the composite `TrackAcceptedHtlc` above it doesn't really matter if you guarantee that in&out policy will be updated atomically. If you have 2 HTLCs being intercepted at the same time (while only one can make it through) then you're just not sure which one is gonna get cancelled, which is the case anyway

rfq/order.go
guggero: Ah yes, my bad.

guggero (commented):
LGTM pending the dependent PR.

GeorgeTsagk:
Rebased on `main` after merging dependent PR, PTAL

guggero (approved):
LGTM :tada:

ffranr (commented):
I'm not sure if we need
```
htlcToPolicy lnutils.SyncMap[models.CircuitKey, Policy]
```
Please see review comments.

rfq/order.go
```diff
@@ -530,6 +662,11 @@ type OrderHandler struct {
 	// associated asset transaction policies.
 	policies lnutils.SyncMap[SerialisedScid, Policy]
 
+	// htlcToPolicy maps an HTLC circuit key to the policy that applies to
+	// it. We need this map because for failed HTLCs we don't have the RFQ
+	// data available, so we need to cache this info.
+	htlcToPolicy lnutils.SyncMap[models.CircuitKey, Policy]
```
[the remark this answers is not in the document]
ffranr: If we can manage to do that then all the HTLC states can be nested under the relevant policies.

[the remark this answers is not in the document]
GeorgeTsagk: Yes `OrderHandler.fetchPolicy` does find the scid but that's not done based on the `CircuitKey` alone, it also uses the `InterceptedHtlc.InWireCustomRecords` field to extract the scid directly. Unfortunately this field is not available when you get notified for the HTLC cancel event*, so that's why I'm caching this info on this layer.

\*It's not a value that's simply not exposed on the API, but more like data that is not available in the lnd channel link by the time we process the HTLC cancel. To change LND and have that data available by the API directly would involve doing the same kind of HTLC-caching on the LND link level (or something similar), which could get much more complicated.

I def agree that eliminating HTLC related state here would be a win, but can't see a workaround.

rfq/order.go
```diff
@@ -313,17 +387,22 @@ func (c *AssetPurchasePolicy) CheckHtlcCompliance(
...
 			"asset amount in millisatoshis (htlc_out_msat=%d, "+
 			"inbound_asset_amount=%s, "+
 			"inbound_asset_amount_msat=%v)", htlc.AmountOutMsat,
 			assetAmt.String(), inboundAmountMSat)
 	}
 
+	// Since we will be reading CurrentAmountMsat value we acquire a read
+	// lock.
+	c.stateMutex.RLock()
+	defer c.stateMutex.RUnlock()
```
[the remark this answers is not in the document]
GeorgeTsagk: Sure, performance overhead is too small to battle readability

rfq/order.go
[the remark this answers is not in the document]
ffranr: I'm trying to remember how this work. Are there HTLCs where we can only get the SCID from their custom records? Is the custom records check strictly necessary for some HTLCs?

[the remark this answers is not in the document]
ffranr: For incoming asset payment HTLCs we inspect custom records. From what I understand and according to @guggero .

So I will approve this PR as is. I will leave merging down to you @GeorgeTsagk . I'll put up a PR to clarify `OrderHandler.fetchPolicy` with comments.

This PR should work as it is either way IMO.

ffranr (approved):
Will leave merging down to @GeorgeTsagk .

rfq/order.go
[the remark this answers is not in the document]
GeorgeTsagk: Yes, we can only retrieve the SCID from the custom records. The ChanID of the circuit keys correspond to the actual scids of the channels that carry the HTLCs

PR #1187, opened 2024-11-12
scripts: use `trap` to make sure old files always restored

Roasbeef:
In this commit, we fix an issue I observed on my machine after updating many command line/dev utilities.

I observed that if the script failed for w/e reason, then the old files weren't properly restored.

The repro on my machine is something like:
  1. Shutdown docker, make sure the daemon isn't accessible.
  2. Run `mac sqlc`. The script should fail.
  3. Observe that the files in your local working tree are still altered.

We fix this by using the `trap` command, [which is basically like Go's `defer`, but for bash
scripts](https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_12_02.html)!

### DIFF
--- a/scripts/gen_sqlc_docker.sh
+++ b/scripts/gen_sqlc_docker.sh
@@ -2,11 +2,23 @@
 
 set -e
 
+# restore_files is a function to restore original schema files.
+restore_files() {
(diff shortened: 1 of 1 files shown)


jharveyb (approved):
LGTM 👍🏽

PR #1188, opened 2024-11-12
proof: add code for stitching together failed proof suffixes

guggero:
This helped me verify and fix proofs in https://github.com/lightninglabs/taproot-assets/issues/1168 and I think the code might be useful for other things too. It won't run in CI, so no harm in keeping it IMO.


guggero:
Merging, as this is just a test that's skipped by default anyway.

Issue #1024, opened 2024-07-15
[feature]: config flag to disable adding the default universes to our federation


jharveyb:
**Is your feature request related to a problem? Please describe.**

If I start `tapd` with a default config, a default universe run by LL will be added to my universe federation. If I later want to remove the universe from my federation, I could use a command like `tapcli u f d`. If I restart my node, the default universe is re-added to my config:

https://github.com/lightninglabs/taproot-assets/blob/d70bccd2714a3f808e070a080c510cf396a11284/tapcfg/server.go#L197

Discovered while manually testing #1014 

**Describe the solution you'd like**

A new config flag to gate this behavior / turn off these defaults, so `tapd` is usable with an empty federation.

**Describe alternatives you've considered**

**Additional context**

PR #1190, opened 2024-11-14
tapcfg: add flag to disable default federation server

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1024.

This commit adds a new universe.no-default-federation boolean flag that allows the user to disable adding the default federation server.

### DIFF
--- a/sample-tapd.conf
+++ b/sample-tapd.conf
@@ -282,6 +282,10 @@
 ; Can be specified multiple times
 ; universe.federationserver=
 
+; If set, the default Universe server (available for testnet and mainnet) will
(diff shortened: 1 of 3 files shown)


jharveyb (approved):
Tested manually on mainnet with CLI flag `--universe.no-default-federation`, worked as expected.

As is it won't delete the default universe(s) if it's already been added, but if you do delete the default, and reboot with the flag set, the default will not be added to the config.

Thanks for taking care of this! 🎉

Roasbeef (approved):
LGTM 💣

PR #1191, opened 2024-11-14
Fee bumping when fee estimation doesn't meet min relay fee

### LINKED ISSUE #1171: [bug]: On signet, min-relay-fee is not met if fee is not specified. Fee will be 1 sat short
### Background

When using signet, it seems that calls to AnchorVirtualPSBT RPC and to SendAsset RPC will result in a transaction that does not meet the min-relay-fee. The vsize of the transaction ends up being 254.5 vB usually (see https://mempool.space/signet/tx/aff4e70f1816ec611da03a89889be79b81b57067355abe99c14e89f0f2085ba8#vin=0 as an example). 

Here is an example txn that will not relay with a fee rate of **barely** under 1.00 sats/vB (size is 254.5, fee is 254)

```
02000000000102aee61b60b329cd99950230a146a11b16ff53c85109d352aee5a5e87088f62b990000000000000000000c6c3c8409df4c95942069945a9cd867d657d13b635814d7884580d9c74a64be01000000000000000003e803000000000000225120b5b88d39d530a564e1b6e2b714e0d4c16bc284b49778898f98faffd8db135b13e803000000000000225120b59492b62792ca980cd5dd087099897d63b6f85fea0680a43e792d70191ee54d00b5eb0b00000000225120d430711c2a954904278394c6c9378713534a2d1d5c37fbea58deebef95596cea0140510398932e106f72156dc13213a7999ef77544a8bf5ee11b7b24859555908b8bd18b30c5889233c95abcc4f7cd2dddc2eab36266f89c38adb137c6e6bdda3a960140a2ccf5834b81952fe1c3eac8276743d6c4571ba232cc98ea0129e7b3582cca21dab09f0f953b98e5f0f8768ec0a4fd400a6b36cc496cb96cef7c47ba75fde17300000000
```

I'm not sure exactly the problem here, perhaps something related to the nature of signet transactions (I havent seen this problem on other networks). Also unfortunately it seems not everyone has adopted package relay, so unable to use CPFP via package relay to unstuck the above txn using the below child:
```
02000000000101f18d47aae097b364ad0dc9409c8b917db47cf1f1535606283614b943d84ed4c202000000000000000001a784eb0b0000000022512044be1fcf86a056efc21c6775b3b348756f3e099a3da70bbe5726558db2b595ac0140b16fcf7770feb3190b6a1bc524c3f80244e7d09399e66808a91594b16289d2b25d341e597834fc7622dc5857576c1a3d72592fce3105636cd5b7a2aad3438e597f5d0300
```

### Your environment

```sh
tapcli getinfo                       # version of `tapd`, `lnd`, and network
{
    "version": "0.4.1-alpha commit=622c7929",
    "lnd_version": "0.18.3-beta",
    "network": "signet",
    "lnd_identity_pubkey": "026580b3316c7b10f487235d8e3247ba960f4d96952aa0edb85a4e096501bb4acc",
    "node_alias": "Tajfi",
    "block_height": 220574,
    "block_hash": "00000109ab0a6ba26dd8b8fb078368e5af4d8ad6fbaca796933d1d1cc4ab6a82",
    "sync_to_chain": true
}
uname -mrsv                          # operating system 
Linux 6.8.0-1015-gcp #17~22.04.1-Ubuntu SMP Tue Sep  3 16:11:52 UTC 2024 x86_64

bitcoind --version  # version of `btcd`, `bitcoind`, or other backend
Bitcoin Core version v28.0.0
```

### Steps to reproduce

Mint any asset and finalize the batch. Then try to spend the asset using the CLI and do NOT specify a fee rate.

### Expected behavior

Transaction should be generated with 254.5 sats --> round up to 255 sats

### Actual behavior

Transaction generated is 254 sats

gijswijs:
Both the minting transaction and normal on chain transactions suffer from edge cases where the fee estimation comes up with a fee that doesn't meet the current min relay fee. This PR changes that behavior by checking the estimated fee against the min relay fee, and bumps the fee if it doesn't clear the minimum fee height implied by min relay fee.

fixes #1171

### DIFF
--- a/itest/addrs_test.go
+++ b/itest/addrs_test.go
@@ -1078,23 +1078,69 @@ func sendProofUniRPC(t *harnessTest, src, dst *tapdHarness, scriptKey []byte,
 	return importResp
 }
 
-// sendAssetsToAddr spends the given input asset and sends the amount specified
+// sendOptions is a struct that holds a SendAssetRequest and an
+// optional error string that should be tested against.
+type sendOptions struct {
+	sendAssetRequest taprpc.SendAssetRequest
+	errText          string
+}
+
+// sendOption is a functional option for configuring the sendAssets call.
+type sendOption func(*sendOptions)
+
+// withReceiverAddresses is an option to specify the receiver addresses for the
+// send.
+func withReceiverAddresses(addrs ...*taprpc.Addr) sendOption {
+	return func(options *sendOptions) {
+		encodedAddrs := make([]string, len(addrs))
+		for i, addr := range addrs {
+			encodedAddrs[i] = addr.Encoded
+		}
+		options.sendAssetRequest.TapAddrs = encodedAddrs
+	}
+}
+
+// withFeeRate is an option to specify the fee rate for the send.
+func withFeeRate(feeRate uint32) sendOption {
+	return func(options *sendOptions) {
+		options.sendAssetRequest.FeeRate = feeRate
+	}
+}
+
+// withError is an option to specify the string that is expected in the error
+// returned by the SendAsset call.
+func withError(errorText string) sendOption {
+	return func(options *sendOptions) {
+		options.errText = errorText
+	}
+}
+
+// sendAsset spends the given input asset and sends the amount specified
 // in the address to the Taproot output derived from the address.
-func sendAssetsToAddr(t *harnessTest, sender *tapdHarness,
-	receiverAddrs ...*taprpc.Addr) (*taprpc.SendAssetResponse,
+func sendAsset(t *harnessTest, sender *tapdHarness,
+	opts ...sendOption) (*taprpc.SendAssetResponse,
 	*EventSubscription[*taprpc.SendEvent]) {
 
 	ctxb := context.Background()
 	ctxt, cancel := context.WithTimeout(ctxb, defaultWaitTimeout)
 	defer cancel()
 
-	require.NotEmpty(t.t, receiverAddrs)
-	scriptKey := receiverAddrs[0].ScriptKey
+	// Create base request that will be modified by options.
+	options := &sendOptions{}
 
-	encodedAddrs := make([]string, len(receiverAddrs))
-	for i, addr := range receiverAddrs {
-		encodedAddrs[i] = addr.Encoded
+	// Apply all the functional options.
+	for _, opt := range opts {
+		opt(options)
 	}
+
+	require.NotEmpty(t.t, options.sendAssetRequest.TapAddrs)
+
+	// We need the first address's scriptkey to subscribe to events.
+	firstAddr, err := address.DecodeAddress(
+		options.sendAssetRequest.TapAddrs[0], &address.RegressionNetTap,
+	)
+	require.NoError(t.t, err)
+	scriptKey := firstAddr.ScriptKey.SerializeCompressed()
 
 	ctxc, streamCancel := context.WithCancel(ctxb)
 	stream, err := sender.SubscribeSendEvents(
@@ -1108,9 +1154,12 @@ func sendAssetsToAddr(t *harnessTest, sender *tapdHarness,
 		Cancel:            streamCancel,
 	}
 
-	resp, err := sender.SendAsset(ctxt, &taprpc.SendAssetRequest{
-		TapAddrs: encodedAddrs,
-	})
+	resp, err := sender.SendAsset(ctxt, &options.sendAssetRequest)
+	if options.errText != "" {
+		require.ErrorContains(t.t, err, options.errText)
+		return nil, nil
+	}
+
 	require.NoError(t.t, err)
 
 	// We'll get events up to the point where we broadcast the transaction.
@@ -1121,6 +1170,15 @@ func sendAssetsToAddr(t *harnessTest, sender *tapdHarness,
 	)
 
 	return resp, sub
+}
+
+// sendAssetsToAddr is a variadic wrapper around sendAsset that enables passsing
+// a multitude of addresses.
+func sendAssetsToAddr(t *harnessTest, sender *tapdHarness,
+	receiverAddrs ...*taprpc.Addr) (*taprpc.SendAssetResponse,
+	*EventSubscription[*taprpc.SendEvent]) {
+
+	return sendAsset(t, sender, withReceiverAddresses(receiverAddrs...))
 }
 
 // fundAddressSendPacket asks the wallet to fund a new virtual packet with the
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/taproot-assets/internal/test"
 	"github.com/lightninglabs/taproot-assets/proof"
 	"github.com/lightninglabs/taproot-assets/tapfreighter"
@@ -18,7 +20,9 @@ import (
 	"github.com/lightninglabs/taproot-assets/taprpc/tapdevrpc"
 	unirpc "github.com/lightninglabs/taproot-assets/taprpc/universerpc"
 	"github.com/lightninglabs/taproot-assets/tapsend"
+	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/lntest/wait"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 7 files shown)


dstadulis:
itest failing due to feeEstimation changes
some values aren't reset -- will investigate more

tapgarden/planter.go
```diff
@@ -566,6 +566,17 @@ func (c *ChainPlanter) fundGenesisPsbt(ctx context.Context,
 			return nil, fmt.Errorf("unable to estimate fee: %w",
 				err)
 		}
+		minRelayFee, err := c.cfg.Wallet.MinRelayFee(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("unable to obtain minimum relay"+
```
guggero, line 571: nit: for ease of review, we standardize putting the space between words on the previous line if we have to wrap a string.

guggero (approved):
Very nice, LGTM :tada:

tapgarden/planter.go
```diff
@@ -566,6 +566,17 @@ func (c *ChainPlanter) fundGenesisPsbt(ctx context.Context,
 			return nil, fmt.Errorf("unable to estimate fee: %w",
 				err)
 		}
+		minRelayFee, err := c.cfg.Wallet.MinRelayFee(ctx)
```
jharveyb, line 569: As-is this won't catch the case of the user manually specifying a feerate below minrelayfee.

Also, in that case IMO you'd want to fail with an error message reporting the minrelayfee, instead of bumping the fee silently.

So I'd suggest moving this check after the switch instead of in the default case; then you can switch behavior based on `manualFeeRate != nil`.

tapfreighter/chain_porter.go
```diff
@@ -1128,6 +1128,17 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
 				return nil, fmt.Errorf("unable to estimate "+
 					"fee: %w", err)
 			}
+			minRelayFee, err := p.cfg.Wallet.MinRelayFee(ctx)
```
jharveyb, line 1131: Same note re: separate switch to catch manual feerate.

itest/send_test.go
```diff
@@ -129,6 +133,123 @@ func testBasicSendUnidirectional(t *harnessTest) {
...
+	// be returned by the fee estimation.
+	lowFeeRate := chainfee.SatPerVByte(1).FeePerKWeight()
+	highMinRelayFeeRate := chainfee.SatPerVByte(2).FeePerKVByte()
+	defaultMinRelayFeeRate := chainfee.SatPerVByte(1).FeePerKVByte()
+	defaultFeeRate := chainfee.SatPerKWeight(3125)
+	t.lndHarness.SetFeeEstimateWithConf(lowFeeRate, 6)
+	t.lndHarness.SetMinRelayFeerate(highMinRelayFeeRate)
+
+	// Reset all fee rates to their default value at the end of this test.
+	defer t.lndHarness.SetMinRelayFeerate(defaultMinRelayFeeRate)
```
jharveyb, line 178: 👍🏽 

good pattern re: defer!

itest/send_test.go
```diff
@@ -129,6 +133,123 @@ func testBasicSendUnidirectional(t *harnessTest) {
...
+	)
+
+	AssertNonInteractiveRecvComplete(t.t, secondTapd, 1)
+	AssertSendEventsComplete(t.t, bobAddr.ScriptKey, sendEvents)
+
+	// Close event stream.
+	err = events.CloseSend()
+	require.NoError(t.t, err)
+
+	wg.Wait()
```
jharveyb, line 250: `wg` not used in this test?

jharveyb (commented):
Looks close! Just one note: re handling manual feerates.

The itest can be expanded a bit to check for failure on an insufficient manual feerate for both mint and transfer as well.

tapgarden/planter.go
gijswijs: Correct. The issue was raised specifically in a scenario where the user did *not* specify a feerate. The reasoning for putting the check inside the switch was that if you do specify the feerate manually, you are fully responsible for it. You could *even* specify a feerate below minrelayfee.

gijswijs:
> The itest can be expanded a bit

Famous last words. :smile: 

Anyways, I've changed both `sendAssetsToAddr` and `MintAssetsConfirmBatch` to be able to test with specific feerates and to be able to test for specific errors. I hope we can benefit from this more granular control in the future for other itests as well.

Please also have a look at 

https://github.com/lightninglabs/taproot-assets/blob/1159a3fe2911b0dc4d5e1f7816daa8d560da58bc/rpcserver.go#L685-L686

If I'm right we could remove a check there.

rpcserver.go
```diff
@@ -664,8 +664,13 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 
 // checkFeeRateSanity ensures that the provided fee rate, in sat/kw, is above
 // the same minimum fee used as a floor in the fee estimator.
-func checkFeeRateSanity(rpcFeeRate uint32) (*chainfee.SatPerKWeight, error) {
+func checkFeeRateSanity(ctx context.Context, rpcFeeRate uint32,
+	lndWallet lndclient.WalletKitClient) (*chainfee.SatPerKWeight, error) {
 	feeFloor := uint32(chainfee.FeePerKwFloor)
```
guggero, line 669: nit: need empty line after multi-line function call definition.

rpcserver.go
```diff
@@ -676,17 +681,27 @@ func checkFeeRateSanity(rpcFeeRate uint32) (*chainfee.SatPerKWeight, error) {
 		return nil, fmt.Errorf("manual fee rate below floor: "+
 			"(fee_rate=%d, floor=%d sat/kw)", rpcFeeRate, feeFloor)
 
+	// A manual fee was set but is below minRelayFee.
+	// TODO (gijs):  Can minRelayFee ever be lower than feeFloor? If not, we
```
guggero, line 685: No, that would be a mistake. I can't speak for all backend implementations, whether they would respect that. But if we want to be sure, we'd check `minRelayFee` returned from lnd (which is just passed forward from the chain backend) and check it against the fee flow. If it's below, we just use the fee floor.

tapgarden/planter.go
```diff
@@ -571,6 +571,32 @@ func (c *ChainPlanter) fundGenesisPsbt(ctx context.Context,
...
+		// If a fee rate was manually assigned for this batch, we err
+		// out, otherwise we silently bump the feerate.
+		case manualFeeRate != nil:
+			// This case should already be handled by the
+			// `checkFeeRateSanity` of `rpcserver.go`. We check here
+			// again to be safe.
+			return nil, fmt.Errorf("feerate does not meet " +
+				"minrelayfee")
+		default:
+			log.Infof("bump fee rate for batch %x to meet "+
```
guggero, line 592: nit: Log statements usually start capitalized (vs. error messages that should be lower case according to Golang defaults and linter preferences).

itest/addrs_test.go
```diff
@@ -1078,24 +1078,70 @@ func sendProofUniRPC(t *harnessTest, src, dst *tapdHarness, scriptKey []byte,
 	return importResp
 }
 
-// sendAssetsToAddr spends the given input asset and sends the amount specified
+// sendOptions is a struct that holds a SendAssetRequest and an
+// optional error string that should be tested against.
+type sendOptions struct {
```
guggero, line 1083: Nice :+1:

itest/addrs_test.go
```diff
@@ -1078,24 +1078,70 @@ func sendProofUniRPC(t *harnessTest, src, dst *tapdHarness, scriptKey []byte,
...
-	for i, addr := range receiverAddrs {
-		encodedAddrs[i] = addr.Encoded
+	// Apply all the functional options
+	for _, opt := range opts {
+		opt(options)
 	}
 
+	require.NotEmpty(t.t, options.sendAssetRequest.TapAddrs)
+
+	// We need the first address's scriptkey to subscribe to events
```
guggero, line 1138: nit: missing full stop at end of sentence.

guggero (approved):
Nice, LGTM :tada:

itest/addrs_test.go
gijswijs: Fun fact: This came from Claude LLM.

rpcserver.go
gijswijs: Resolved it slightly different, so that we can still throw different errors.

tapfreighter/chain_porter.go
```diff
@@ -1128,6 +1128,40 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
...
+			switch {
+			// If a fee rate was manually assigned for this parcel,
+			// we err out, otherwise we silently bump the feerate.
+			case addrParcel.transferFeeRate != nil:
+				// This case should already be handled by the
+				// `checkFeeRateSanity` of `rpcserver.go`. We
+				// check here again to be safe.
+				p.unlockInputs(ctx, &currentPkg)
+
+				return nil, fmt.Errorf("feerate does not " +
```
jharveyb, line 1156: super-nit: should return the `minrelayfee` here so the user can reattempt above that feerate.

tapgarden/planter.go
```diff
@@ -571,6 +571,32 @@ func (c *ChainPlanter) fundGenesisPsbt(ctx context.Context,
...
+	// bump it up.
+	if feeRate < minRelayFee {
+		switch {
+		// If a fee rate was manually assigned for this batch, we err
+		// out, otherwise we silently bump the feerate.
+		case manualFeeRate != nil:
+			// This case should already be handled by the
+			// `checkFeeRateSanity` of `rpcserver.go`. We check here
+			// again to be safe.
+			return nil, fmt.Errorf("feerate does not meet " +
```
jharveyb, line 589: same nit re: displaying `minrelayfee`.

jharveyb (approved):
Great changes + nice to see itest improvements 🎉

Issue #1053, opened 2024-07-31
[feature]: add channel conditional validation to `AddAsset*Order` RPC endpoints


ffranr:
Currently, the `AddAssetBuyOrder` and `AddAssetSellOrder` RPC endpoints initiates the RFQ communication process with a specified target peer. This issue proposes a modification to the endpoint to include a pre-condition check. Specifically, the RFQ process should only commence if there is an existing tap channel for the specified asset between the local node and the target peer, and that channel has sufficient funds. If these conditions are not met, the endpoints should return a helpful error.

dstadulis:
RFQ rework is a prerequisite

PR #1192, opened 2024-11-14
[rfq]: add validation to `AddAssetBuyOrder` and `AddAssetSellOrder` RPCs

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1053.

Adds basic validation to `AddAssetBuyOrder` and `AddAssetSellOrder` RPCs: We want to have a channel with the given asset to the given peer when adding an order. Otherwise RFQ negotiation doesn't make sense.

We add an exception for our integration tests, where we can't create full-blown asset channels (in the `tapd` standalone integration tests). A new flag can be specified that then loosens the requirement to just a normal, active channel with the peer.


ffranr (commented):
I think we should consider dropping the term "balance" from the new req arg name. 

We can just call it `SkipAssetChannelCheck` and the "channel check" can include checking for an existing channel and any balance check that we may add later (is absent from this PR i think?).

rfq/negotiator.go
```diff
@@ -467,6 +467,15 @@ func (n *Negotiator) HandleOutgoingSellOrder(order SellOrder) {
 	go func() {
 		defer n.Wg.Done()
 
+		// Unwrap the peer from the buy order. For now, we can assume
+		// that the peer is always specified.
+		peer, err := order.Peer.UnwrapOrErr(
+			fmt.Errorf("buy order peer must be specified"),
```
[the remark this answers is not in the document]
guggero: Fixed by rebasing on your merged commit.

taprpc/rfqrpc/rfq.proto
```diff
@@ -138,6 +138,12 @@ message AddAssetBuyOrderRequest {
     // timeout_seconds is the number of seconds to wait for the peer to respond
     // with an accepted quote (or a rejection).
     uint32 timeout_seconds = 5;
+
+    // If set, the check if a channel with the given asset exists with the peer
+    // will be skipped. An active channel with the peer is still required for
+    // the RFQ negotiation to work. This flag shouldn't be set outside of test
+    // scenarios.
+    bool skip_asset_balance_check = 6;
```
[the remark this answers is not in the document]
guggero: Yeah, you're right. We do check that the balance isn't zero, but nothing more. Renamed.

taprpc/rfqrpc/rfq.proto
```diff
@@ -174,6 +180,12 @@ message AddAssetSellOrderRequest {
...
```
ffranr, line 188: I think this also needs to be renamed to `skip_asset_channel_check`.

guggero: Hmm, I did, with the latest push. But forgot the itest... Which is fixed now too.

guggero: Ah, there are two RPC calls... Sorry, getting too late. Will fix.

ffranr (approved):

gijswijs (approved):
tACK :tada:

Just one nit.

I would like to see a future where `tapd` can be tested with some sort of a `lnd`-mock so that we can mimick proper asset channels, and not need to depend on flags like this.

PR #1197, opened 2024-11-15
More RFQ improvments

ffranr:
Improve docs for: `SellOrder`, `SellRequest`, `BuyOrder`, `BuyRequest`.
Populate transfer type field in request.
Make use of `AssetRate` type.

### DIFF
--- a/rfqmsg/buy_accept.go
+++ b/rfqmsg/buy_accept.go
@@ -2,8 +2,8 @@ package rfqmsg
 
 import (
 	"fmt"
+	"time"
 
-	"github.com/lightninglabs/taproot-assets/rfqmath"
 	"github.com/lightningnetwork/lnd/routing/route"
 )
 
@@ -30,10 +30,7 @@ type BuyAccept struct {
 	ID ID
 
 	// AssetRate is the accepted asset to BTC rate.
-	AssetRate rfqmath.BigIntFixedPoint
-
-	// Expiry is the asking price expiry lifetime unix timestamp.
-	Expiry uint64
+	AssetRate AssetRate
 
 	// sig is a signature over the serialized contents of the message.
 	sig [64]byte
@@ -41,10 +38,8 @@ type BuyAccept struct {
 
 // NewBuyAcceptFromRequest creates a new instance of a quote accept message
 // given a quote request message.
-//
-// TODO(ffranr): Use new AssetRate type for assetRate arg.
 func NewBuyAcceptFromRequest(request BuyRequest,
-	assetRate rfqmath.BigIntFixedPoint, expiry uint64) *BuyAccept {
+	assetRate AssetRate) *BuyAccept {
 
 	return &BuyAccept{
 		Peer:      request.Peer,
@@ -52,7 +47,6 @@ func NewBuyAcceptFromRequest(request BuyRequest,
 		Version:   latestBuyAcceptVersion,
 		ID:        request.ID,
 		AssetRate: assetRate,
-		Expiry:    expiry,
 	}
 }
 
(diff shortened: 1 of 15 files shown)


ffranr:
Things left to think about:
* request fields: min in-asset and min out-asset.
* Use transfer type to determine buy/sell. Remove need for all zeros asset ID to signify BTC if we can just use transfer type? BLIP?

rfqmsg/messages.go
```diff
@@ -90,6 +90,63 @@ func (id *ID) Record() tlv.Record {
 	return tlv.MakeStaticRecord(0, id, recordSize, IdEncoder, IdDecoder)
 }
 
+// TransferType defines the type of transaction which will be performed if the
+// quote request leads to an accepted agreement.
+type TransferType uint8
```
GeorgeTsagk, line 95: why are we adding this? I don't see why we need to describe the purpose of the transfer via this enum

GeorgeTsagk (commented):
Very nice cleanup

Could you rebase so that we can see the LitD itests running against this too?

rfqmsg/messages.go
ffranr: This field may not be strictly necessary, in my opinion. However, after discussing with @guggero, I believe we're including it for two reasons:  
1. Debugging: For asset-to-asset quotes (rather than asset-to-BTC) later on, it could provide useful context.  
2. Future use: This field might support adding other transfer types in the future.

guggero: Correct. But we should also actually set and use it. My goal is to replace the following code with the transfer type detection:

https://github.com/lightninglabs/taproot-assets/blob/c7273399e009a4b6b710c59b818e34c559773130/rfqmsg/request.go#L512-L551

rfqmsg/buy_accept.go
```diff
@@ -44,15 +42,14 @@ type BuyAccept struct {
 //
 // TODO(ffranr): Use new AssetRate type for assetRate arg.
 func NewBuyAcceptFromRequest(request BuyRequest,
-	assetRate rfqmath.BigIntFixedPoint, expiry uint64) *BuyAccept {
+	assetRate rfqmath.BigIntFixedPoint, expiry time.Time) *BuyAccept {
```
guggero, line 45: nit: feels like we could also fix the TODO above with this commit? Since we now have the correct data types in place.

guggero (commented):
Very nice, thanks for the additional cleanup!
One request and one nit, so we're quite close :pray:

GeorgeTsagk (approved):
Lgtm!

rfqmsg/messages.go
```diff
@@ -90,6 +90,63 @@ func (id *ID) Record() tlv.Record {
...
+
+const (
+	// UnspecifiedTransferType represents an undefined or transfer type.
+	UnspecifiedTransferType TransferType = iota
```
guggero, line 99: Constants that are stored in the DB or are transferred over the network should never use `iota`, otherwise it's too easy to accidentally add a value somewhere in the middle and changing the meaning of all following constants.

rfqmsg/request.go
```diff
@@ -509,44 +509,20 @@ func NewIncomingRequestFromWire(wireMsg WireMessage) (IncomingMsg, error) {
...
-	if isBuyRequest {
+	// Classify the incoming request as a buy or sell.
+	//
+	// When the requesting peer attempts to pay an invoice using a Tap
+	// asset, they are "selling" the Tap asset to the edge node. Conversely,
+	// when the requesting peer attempts to receive a Tap asset as payment
+	// to settle an invoice, they are "buying" the Tap asset from the edge
+	// node.
+	switch msgData.TransferType.Val {
+	case PayInvoiceTransferType:
```
guggero, line 520: Yayy :tada:

guggero (approved):
Sorry, found one last thing. Other than that LGTM :tada:

Issue #1162, opened 2024-10-29
`tapdb` - [feature]: Modify asset-creation criteria to distinguish between spendable / watch-only assets


dstadulis:
## Background
The in situ, stopgap solution for handling asset balances is to whitelists the funding script key. However, this approach has limitations, as there are scenarios where the node needs to recognize an output as spendable even if the asset can't be spent at present or future conditions would change which would make the asset spendable by the node.

This is critical for crediting new outputs in the database, specifically when creating assets
https://github.com/lightninglabs/taproot-assets/blob/2639a99bcad7ddf7ac84b742175077ffb682b673/tapdb/assets_store.go#L3042-L3060

## Problem
By skiping asset creation for outputs with fancy script keys, `tapdb` won't create an input asset to spend later on. This leads to a failed sweep operation when trying to spend these assets.

## Proposed Solution
- Insert assets in the database as normal. By default, don't return any asset with a fancy script key as balance. Instead, return these assets with a separate flag, effectively treating them as watch-only.

This approach allows:

- Mark assets as belonging to the node -- even if the node can't spend them yet (e.g. when a preimage is remains unknown)
- Import these assets later when they become spendable
- Avoid failing sweep operations due to missing input assets

## Expected Outcome
By distinguishing between spendable and watch-only assets, the wallet has the necessary predicates to correctly handle asset balances and avoid errors during sweep operations.

## Alternatives Considered
TINA

PR #1198, opened 2024-11-15
[tapdb]: add script key type enum, run Golang based post migration checks

guggero:
Depends on https://github.com/golang-migrate/migrate/pull/1253 (might need to maintain our own fork if we decide to go with the approach).

Fixes https://github.com/lightninglabs/taproot-assets/issues/1458.
Fixes https://github.com/lightninglabs/taproot-assets/issues/1162.

Adds a specific `key_type` field to the script key table, which allows us to distinguish between the following types of keys:
 - Unknown (--> replaces the previous `declared_known` boolean, which can now be synthesized with `key_type != Unknown`)
 - Bip-86
 - Script path external (user-defined key with script paths)
 - Burn
 - Tombstone
 - Taproot Asset Channel keys

This allows us to correctly filter out channel related keys when showing balances (e.g. `ListBalances`, `ListUtxos` or `ListAssets`) and also allows us to do more custom coin selection.

This is required for cleanly filter out channel related script keys implemented in https://github.com/lightninglabs/taproot-assets/pull/1413.

cc @GeorgeTsagk: this mechanism can be used to add asset burns to the new table retroactively.


guggero:
I finally got this down to a single TODO: Add more integration tests.
But other than that this contains all functionality that I wanted to add (and that we need).
So asking for a first round of review now.

guggero:
Addressed the final TODO by asserting the new balances in multiple integration test steps.

ffranr (commented):
I don't think we need to fork the `migrate` package (https://github.com/guggero/migrate/commit/f82a98804d82bffa60b61d3a2e802cd1ae3d7ce0) It would be nice not to have to maintain a fork.

Can't we get away with just reformulating `applyMigrations` in `tapdb/migrations.go`?

Roughly:
```go
sqlMigrate, err := migrate.NewWithInstance(
    "migrations", migrateFileServer, dbName, driver,
)

// We will walk `opts.postStepCallbacks` in ascending version index.
// Get next post migration callback index from `opts.postStepCallbacks`:
// `nextVersion`.
//
// Migrate to that version.
sqlMigrate.Migrate(nextVersion)

// Run callback function using the same driver that we passed into
// `migrate.NewWithInstance`.
opts.postStepCallbacks[nextVersion](driver)

// Handle the next version in `opts.postStepCallbacks` or until target migration version is reached.
```

The high level idea is to migrate using the `migrate` in stages and pause to apply post migration callbacks.

We can also set the version to dirty before exec callback using `driver.SetVersion(version int, dirty bool)` and then back not dirty if successful.

ffranr:
> > The high level idea is to migrate using the migrate in stages and pause to apply post migration callbacks.
> 
> With your formulation, what happens if we upgrade to version 3 from version 2 (2->3), then as we're executing the call back for post version 2, the daemon crashes?
> 
> IIUC, we'll start from version 3, and skip the post step call back of version 2.
> 
> What we want here is that the call back is executed in the exact same context as the migration to version 2 in order to retain atomicity. The db shouldn't move to version 2 until the call back has successfully been executed.

@Roasbeef I think you're right. I now see that we want to execute the callback while holding the same driver lock that was used to migrate. Given that, I agree the fork methodology is likely the best path forward.

ffranr:
> > The high level idea is to migrate using the migrate in stages and pause to apply post migration callbacks.
> 
> With your formulation, what happens if we upgrade to version 3 from version 2 (2->3), then as we're executing the call back for post version 2, the daemon crashes?
> 
> IIUC, we'll start from version 3, and skip the post step call back of version 2.
> 
> What we want here is that the call back is executed in the exact same context as the migration to version 2 in order to retain atomicity. The db shouldn't move to version 2 until the call back has successfully been executed.

@Roasbeef I agree that my approach isn't atomic, since the callback wouldn't run within the same database transaction. But I believe that's also the case with the current state of the migration package fork solution.

That said, the fork solution does allow the callback to run while holding the same driver lock, which is still valuable. So I think forking the package is probably the best path forward.

asset/asset.go
```diff
@@ -1118,12 +1109,9 @@ func (s *ScriptKey) IsEqual(otherScriptKey *ScriptKey) bool {
...
 // DeclareScriptKey RPC. Knowing the key conceptually means the key belongs to
 // the local wallet or is at least known by a software that operates on the
-// local wallet. The DeclaredAsKnown flag is never serialized in proofs, so this
-// is never explicitly set for keys foreign to the local wallet. Therefore, if
-// this method returns true for a script key, it means the asset with the script
-// key will be shown in the wallet balance.
+// local wallet.
 func (s *ScriptKey) DeclaredAsKnown() bool {
-	return s.TweakedScriptKey != nil && s.TweakedScriptKey.DeclaredKnown
+	return s.TweakedScriptKey != nil && s.Type != ScriptKeyUnknown
```
ffranr, line 1114: Maybe `ScriptKeyUnknown` should be renamed to `ScriptKeyTypeUnknown`, because the type of script key is the thing that is unknown, even for a tracked script key. Have i understood that correctly?

rpcserver.go
```diff
@@ -2302,21 +2302,21 @@ func (r *rpcServer) FundVirtualPsbt(ctx context.Context,
 
 // unmarshalCoinSelectType converts an RPC select type into a native one.
 func unmarshalCoinSelectType(
-	coinSelectType wrpc.CoinSelectType) (tapsend.CoinSelectType, error) {
+	coinSelectType wrpc.CoinSelectType) (fn.Option[asset.ScriptKeyType],
```
ffranr, line 2305: func doc comment could be updated

tapdb/assets_store.go
```diff
@@ -968,19 +967,25 @@ func (a *AssetStore) QueryBalancesByAsset(ctx context.Context,
...
 	}
 
+	// We exclude the assets that are specifically used for funding custom
+	// channels. The balance of those assets is reported through lnd channel
+	// balance. Those assets are identified by the specific script key type
+	// for channel keys. We exclude them unless explicitly queried for.
+	assetBalancesFilter.ExcludeScriptKeyType = sqlInt16(
+		asset.ScriptKeyScriptPathChannel,
+	)
+
```
ffranr: Do we also need to exclude `burn` and `tombstone` here? I would expect `ExcludeScriptKeyType` to be a list—same in the SQL query. But perhaps we can defer making it a list for now and still remain correct?

tapfreighter/chain_porter.go
```diff
@@ -1296,6 +1296,12 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
 				"%w", err)
 		}
 
+		// Burn and tombstone keys are the only keys that we don't
+		// explicitly store in the DB before this point. But we'll want
+		// them to have the correct type when creating the transfer, so
+		// we'll set that now.
+		detectUnSpendableKeys(currentPkg.VirtualPackets)
```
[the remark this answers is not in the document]
guggero: Such a key might come from the RPC interface, set by the user. So it's hard to determine an exact point before that we could reliably do this. But I moved the logic even closer to the database logic, so it is more clear that this only affects what is stored in the DB and not really the transfer itself.

asset/asset.go
guggero: Yeah, I guess so. But then as a consequence we should also put the word `Type` into the values of the pseudo-enum, so `ScriptKeyTypeBip86`, `ScriptKeyTypeScriptPathExternal` and so on, which makes them even longer.
But since we have an explicit Golang type that also has the word `Type` in it, that also provides some context to it. So as long as we don't have any naming collision I'm opting to keep the current naming.

tapdb/assets_store.go
guggero: I would argue no. Burn keys are something the user explicitly creates, so IMO they should be shown if the user selects `with_spent`. And tombstones are created behind the scenes, _BUT_ they have have a relevance to the user as they might lock up BTC in an otherwise empty UTXO (see https://github.com/lightninglabs/taproot-assets/issues/1458 that is fixed by this PR).
The only type of script key that IMO should always be hidden from the user (as otherwise they _will_ cause confusion and questions) are channel related keys and their assets. Sure, the user creates those channels, so it is the result of a user action. But for any balance related calls we should not include them, as otherwise the assets might be counted twice (once off-chain and once on-chain). And especially force close output related script keys should be hidden, as they are always swept into "normal" outputs anyway under the hood.

guggero:
Thanks for the review! Addressed all comments, ready for another round.

tapdb/sqlc/queries/addrs.sql
```diff
@@ -10,12 +10,8 @@ SELECT
     version, asset_version, genesis_asset_id, group_key, tapscript_sibling,
     taproot_output_key, amount, asset_type, creation_time, managed_from,
     proof_courier_addr,
-    script_keys.tweaked_script_key,
-    script_keys.tweak AS script_key_tweak,
-    script_keys.declared_known AS script_key_declared_known,
-    raw_script_keys.raw_key AS raw_script_key,
-    raw_script_keys.key_family AS script_key_family,
-    raw_script_keys.key_index AS script_key_index,
+    sqlc.embed(script_keys),
```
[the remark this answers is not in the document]
guggero: Yes. As you can see in the diff of the generated code.

tapdb/sqlc/migrations/000033_script_key_type.up.sql
```diff
@@ -0,0 +1,8 @@
+-- The key_type column is used to store the type of key that is stored in the
+-- script_keys table. The type is a Golang numeric type that will have values
+-- such as BIP-0086, script path with custom (externally defined) script, script
+-- path with Taproot Asset Channel related script, etc. The NULL value
+-- will mean the type is not known. Existing script keys will be inspected by
+-- the post-migration step Golang code that determines the type of each script
+-- key.
+ALTER TABLE script_keys ADD COLUMN key_type SMALLINT;
```
Roasbeef, line 8: Non-blocking, but an alternative here would be the enum table pattern. So a new `script_key_type` table w/ set values (burn, transfer, channel, etc) and this new column refs that. This let's us put the mapping of string type to integer within the db itself. 

Tradeoff is potentially a join to match thing sup if just the primary key is stored here. Adding new types also then is a migration, albeit just one that inserts a single row into this new table. 

Just using an integer type here and putting the mapping in the Go layer as is don here certainly simpler.

tapdb/sqlc/queries/assets.sql
```diff
@@ -864,27 +865,37 @@ WHERE txid = $1;
...
-                 END
+      tweak =
+          CASE
+             WHEN script_keys.tweak IS NULL
+             THEN COALESCE(EXCLUDED.tweak, script_keys.tweak)
+             ELSE script_keys.tweak
+           END,
+      -- We only overwrite the key type with a value that does not mean
+      -- "unknown" (0 or NULL).
+        key_type =
```
Roasbeef, line 885: 👍

asset/asset.go
```diff
@@ -108,6 +108,50 @@ const (
 	EncodeSegwit
 )
 
+// ScriptKeyType denotes the type of script key used for an asset. This type is
+// serialized to the database, so we don't use iota for the values to ensure
+// they don't change by accident.
+type ScriptKeyType uint8
```
Roasbeef, line 114: 👍

asset/asset.go
```diff
@@ -1098,15 +1145,55 @@ func (s *ScriptKey) HasScriptPath() bool {
...
+	if s.TweakedScriptKey != nil &&
+		s.TweakedScriptKey.Type != ScriptKeyUnknown {
+
+		return s.TweakedScriptKey.Type
+	}
+
+	// If there is a known tweak, then we know that this is a script path
+	// key. We never return the channel type, since those keys should always
+	// be declared properly, and we never should need to guess their type.
+	if s.HasScriptPath() {
```
Roasbeef, line 1155: Seems that for tap channel keys, we'd miss out on a more precise label here. Unless we always take a code path that skips this routine.

rpcserver.go
```diff
@@ -8134,7 +8134,12 @@ func (r *rpcServer) DeclareScriptKey(ctx context.Context,
 			err)
 	}
 
-	err = r.cfg.TapAddrBook.InsertScriptKey(ctx, *scriptKey, true)
+	// Because we've been given the key over the RPC interface, we can't be
+	// 100% sure of the type. But we can make a best effort guess based on
+	// the fields the user has set.
+	keyType := scriptKey.DetermineType()
```
Roasbeef: Any reason to not pass in the key type? That would give the caller a chance to do this determination.

tapdb/post_migration_checks.go
```diff
@@ -0,0 +1,204 @@
...
+
+				newType = assumedType
+
+			// Previous channel funding script keys (OP_TRUE) can
+			// be detected, since they are the same key. New, unique
+			// script keys for grouped asset channels are only
+			// introduced with the same version as this migration
+			// ships in, so they should be stored with the correct
+			// type already.
+			case assumedType == asset.ScriptKeyScriptPathExternal &&
```
Roasbeef, line 172: 👍

tapdb/post_migration_checks.go
```diff
@@ -0,0 +1,204 @@
...
+				TweakedScriptKey: k.ScriptKey.TweakedScriptKey,
+				Tweak:            k.ScriptKey.Tweak,
+				DeclaredKnown:    k.ScriptKey.DeclaredKnown,
+				KeyType:          sqlInt16(newType),
+			})
+			if err != nil {
+				return fmt.Errorf("error updating script key "+
+					"type: %w", err)
+			}
+		}
```
Roasbeef, line 199: Log if we aren't able to detect the key type? 

Don't think we have a dry run migration mode right now, but may be something we want to add, as can be an extra sanity check here.

rpcserver.go
```diff
@@ -8135,11 +8135,21 @@ func (r *rpcServer) DeclareScriptKey(ctx context.Context,
...
+	// 100% sure of the type, if it wasn't declared. But we can make a
+	// best-effort guess based on the fields the user has set. This is a
+	// no-op if the type is already set.
+	scriptKey.Type = scriptKey.DetermineType()
 
-	err = r.cfg.TapAddrBook.InsertScriptKey(ctx, *scriptKey, true, keyType)
+	// The user is declaring the key, so they should know what type it is.
+	// So if they didn't set it, and it wasn't an obvious one, we'll require
+	// them to set it.
+	if scriptKey.Type == asset.ScriptKeyUnknown {
```
Roasbeef, line 8181: Cool, can ignore my other comment.

taprpc/marshal.go
```diff
@@ -692,3 +692,37 @@ func MarshalAsset(ctx context.Context, a *asset.Asset,
...
+	case *ScriptKeyTypeQuery_ExplicitType:
+		explicitType, err := UnmarshalScriptKeyType(t.ExplicitType)
+		if err != nil {
+			return fn.None[asset.ScriptKeyType](), false, err
+		}
+
+		includeSpent := false
+		switch explicitType {
+		case asset.ScriptKeyTombstone, asset.ScriptKeyBurn:
+			includeSpent = true
```
Roasbeef, line 715: Why do we always want to include spent for burns and tombstones?

Roasbeef: Ah I see here from the proto comments, can add there here to imo: 
```
If the type
    // is set to SCRIPT_KEY_BURN or SCRIPT_KEY_TOMBSTONE the include_spent flag
    // will automatically be set to true, because assets of that type are always
    // marked as spent.
```

tapdb/assets_store.go
```diff
@@ -999,19 +998,25 @@ func (a *AssetStore) QueryBalancesByAsset(ctx context.Context,
...
-	// channels. The balance of those assets is reported through lnd channel
-	// balance. Those assets are identified by the funding script tree for a
-	// custom channel asset-level script key.
-	excludeKey := asset.NewScriptKey(
-		tapscript.NewChannelFundingScriptTree().TaprootKey,
-	)
-	assetBalancesFilter.ExcludeKey = excludeKey.PubKey.SerializeCompressed()
+		// If the user explicitly wants to see the channel related asset
+		// balances, we need to set the exclude type to NULL.
+		if t == asset.ScriptKeyScriptPathChannel {
```
Roasbeef, line 1015: But this doesn't allow them to _only_ query for the channel related balances?

tapdb/sqlc/queries/assets.sql
```diff
@@ -339,8 +339,8 @@ JOIN managed_utxos utxos
 JOIN script_keys
     ON assets.script_key_id = script_keys.script_key_id
 WHERE spent = FALSE AND 
-      (script_keys.tweaked_script_key != sqlc.narg('exclude_key') OR
-        sqlc.narg('exclude_key') IS NULL) AND
+      (script_keys.key_type != sqlc.narg('exclude_script_key_type') OR
```
Roasbeef, line 342: 👍

itest/burn_test.go
```diff
@@ -157,9 +157,10 @@ func testBurnAssets(t *harnessTest) {
 
 	// We'll now assert that the burned asset has the correct state.
 	burnedAsset := burnResp.BurnProof.Asset
-	allAssets, err := t.tapd.ListAssets(
-		ctxt, &taprpc.ListAssetRequest{IncludeSpent: true},
-	)
+	allAssets, err := t.tapd.ListAssets(ctxt, &taprpc.ListAssetRequest{
+		IncludeSpent:  true,
+		ScriptKeyType: allScriptKeysQuery,
```
Roasbeef, line 162: Do we need to change CLI defaults as a result of this? 

Also something to call out in the release notes as a change in behavior here.

Roasbeef (approved):
Great work! Very comprehensive change, and now we have a method of doing optional Go migrations as well. 

Only potential blocker I identified is if we need to change the defaults on the CLI due to the new filtration semantics. We may also need to modify some balance checks/assertions in the litd channel tests as well. Can do both/either of those as a follow up. 

LGTM 🦃

tapdb/sqlc/migrations/000033_script_key_type.up.sql
guggero: Right, but what would the benefits of that be? IMO it doesn't make the data more easy to read "by eye", meaning that without the join it would still be a non-self-explanatory number.
And on top of that we'd need to string matching to map the values to Golang constants. Or would the idea be to use string-based constants on the Golang level?

asset/asset.go
guggero: Yes, that's correct. For already existing channel keys (excluding the funding `OP_TRUE` key, as that's detectable), we couldn't assign the correct value.
But those should only include force close outputs. And they should be spent already, so hopefully not causing too much confusion for users.

rpcserver.go
guggero: You mean from the RPC request? We'll do that in a later commit.

taprpc/marshal.go
guggero: Good point. Added that comment.

tapdb/assets_store.go
guggero: Hmm, not sure I understand... We always set the `ExcludeScriptKeyType` to channel keys above. But if the user explicitly wants to see channel keys, we set the `ExcludeScriptKeyType` to `NULL`, disabling the exclude.

tapdb/sqlc/queries/assets.sql
```diff
@@ -135,6 +135,9 @@ SELECT seedling_id, asset_name, asset_type, asset_version, asset_supply,
     sqlc.embed(assets_meta),
     emission_enabled, batch_id, 
     group_genesis_id, group_anchor_id, group_tapscript_root,
+    -- TODO(guggero): We should use sqlc.embed() for the script key and internal
+    -- key fields, but we can't because it's a LEFT JOIN. We should check if the
+    -- LEFT JOIN is actually necessary or if we always have keys for seedlings.
```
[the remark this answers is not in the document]
guggero: I think this is a quirk of the minting logic. At some point during the minting process there can be seedlings that don't yet have a script key defined, but we still want to be able to store them. I changed this to a normal join and tests started failing, so I didn't look into it more closely.

itest/burn_test.go
guggero: No, the defaults remain the same. But I added a new `--script_key_type` flag to all relevant commands so the users can also filter on the command line.

guggero:
Had to add a commit to point to the `script-key-migrations` branch in `litd`: https://github.com/lightninglabs/lightning-terminal/pull/1053
Was able to re-use a lot of the new itest balance assertion code, which is nice.

@Roasbeef feel free to merge this PR if you feel like your questions have been addressed satisfactorily.

PR #1199, opened 2024-11-16
Updated ListAssetRequest to include additional filter options

### LINKED ISSUE #876: [feature]: Add filters to `ListAssets`
## Description

We should add filters to the `ListAssets` call, so that only assets that satisfy certain criteria are returned.
Examples of such filters could be:
- `GroupKey`: only return assets that belong to the group with this key
- `MinAmount`: only return assets with amount greater or equal to this value
- `MaxAmount`: only return assets with amount less or equal to this value
- `ScriptKey`: return all assets that use this script key
- `AnchorOutpoint`: return all assets that are currently anchored on this outpoint

itsrachelfish:
Resolves #876 

This PR adds new filter options to the `ListAssets` RPC call.

- The `ListAssetRequest` proto definition was updated to include the new filter options.
- The `fetchRpcAssets` function was refactored to accept a `AssetQueryFilters` argument so that separate arguments don't need to be added for each filter option.
- The assets store `constraintsToDbFilter` function was updated to use the new filter options.
- And the `assets.sql` file was updated to include a new `maxAmount` option.


rpcserver.go
```diff
@@ -953,8 +954,13 @@ func (r *rpcServer) ListAssets(ctx context.Context,
 	}
 
 	rpcAssets, err := r.fetchRpcAssets(
-		ctx, req.WithWitness, req.IncludeSpent, req.IncludeLeased,
+		ctx,
+		req.WithWitness,
+		req.IncludeSpent,
+		req.IncludeLeased,
+		req.MinAmount,
```
GeorgeTsagk, line 961:
```suggestion
		ctx, req.WithWitness, req.IncludeSpent,	req.IncludeLeased,
		req.MinAmount,
```

rpcserver.go
```diff
@@ -953,8 +954,13 @@ func (r *rpcServer) ListAssets(ctx context.Context,
...
 	)
+
```
GeorgeTsagk, line 963: extra newline

rpcserver.go
```diff
@@ -1007,10 +1013,20 @@ func (r *rpcServer) ListAssets(ctx context.Context,
 }
 
 func (r *rpcServer) fetchRpcAssets(ctx context.Context, withWitness,
-	includeSpent, includeLeased bool) ([]*taprpc.Asset, error) {
+	includeSpent, includeLeased bool,
+	minAmountFilter uint64) ([]*taprpc.Asset, error) {
+
+	constraints := tapfreighter.CommitmentConstraints{
+		CoinSelectType: tapsend.ScriptTreesAllowed,
+		MinAmt:         minAmountFilter,
```
GeorgeTsagk, line 1021: for the final version of this PR we'd want the extra filters to be present here

sample-tapd.conf
```diff
@@ -328,8 +328,8 @@
 ; universe.multiverse-caches.syncer-cache-pre-alloc-size=100000
 
 ; The size of the root node page cache for all requests that aren't served by
-; the syncer cache. (default: 10240)
-; universe.multiverse-caches.root-node-page-cache-size=10240
+; the syncer cache. (default: 327680)
```
GeorgeTsagk, line 331: since this was an unrelated bugfix it's better to have it in a separate commit

taprpc/taprootassets.proto
```diff
@@ -189,6 +189,21 @@ message ListAssetRequest {
...
+    uint64 max_amount = 6;
+
+    // Only return assets that belong to the group with this key
+    bytes group_key = 7;
+
+    // Return all assets that use this script key
+    bytes script_key = 8;
+
+    // Return all assets that are currently anchored on this outpoint
+    string anchor_outpoint = 9;
```
GeorgeTsagk, line 206: missing period `.` on comments

GeorgeTsagk (commented):
looking good!
it's in the right direction :100:

rpcserver.go
itsrachelfish: What's the style guide here? No newlines after parentheses, but yes to newlines after curly braces? :thinking:

GeorgeTsagk: it's specifically for when we do `if err != nil` right after a func call

even if the call is multiline we stich the `if err` statement right after

tapdb/assets_store.go
```diff
@@ -980,6 +1008,10 @@ type AssetQueryFilters struct {
 	// MinAnchorHeight is the minimum block height the asset's anchor tx
 	// must have been confirmed at.
 	MinAnchorHeight int32
+
+	ScriptKeyID *int64
```
GeorgeTsagk, line 1012: missing godocs (see `MinAnchorHeight` above)

tapdb/assets_store.go
```diff
@@ -883,12 +883,40 @@ func (a *AssetStore) constraintsToDbFilter(
...
+			}
+		}
+
 		if query.MinAnchorHeight != 0 {
 			assetFilter.MinAnchorHeight = sqlInt32(
 				query.MinAnchorHeight,
 			)
 		}
 
+		if query.ScriptKeyID != nil {
```
GeorgeTsagk, line 906: we can't really have user-facing arguments that reference database IDs

the `ScriptKeyID` is the DB identifier for a script key and there's no possible way a user could (...easily) get their hands on one of them in order to filter the assets

Instead we could have `*AssetQueryFilters` expose `scriptKey []byte` and `anchorOutpoint []byte` filters, which the user can easily acquire, then on this point in the code we could map these values to the unique db ID that the underlying layer requires for filtering (that is `QueryAssetFilters` struct)

GeorgeTsagk: if you add some tests with the current way things work, you'll see that when you'll try to query a specific script key you'll have trouble finding the ID to filter by

rpcserver.go
```diff
@@ -1024,10 +1064,11 @@ func (r *rpcServer) ListAssets(ctx context.Context,
 }
 
 func (r *rpcServer) fetchRpcAssets(ctx context.Context, withWitness,
-	includeSpent, includeLeased bool) ([]*taprpc.Asset, error) {
+	includeSpent, includeLeased bool,
+	queryFilters *tapdb.AssetQueryFilters) ([]*taprpc.Asset, error) {
 
 	assets, err := r.cfg.AssetStore.FetchAllAssets(
-		ctx, includeSpent, includeLeased, nil,
+		ctx, includeSpent, includeLeased, queryFilters,
```
GeorgeTsagk, line 1071: you should add some unit test coverage in `tapdb` to test the behavior of `AssetStore.FetchAllAssets`

GeorgeTsagk (commented):
lots of improvements! :arrow_up: 
I have one comment on the types of filters

We'll also need that unit test coverage :heavy_check_mark:

Roasbeef:
I think this function can help resolve the issue w.r.t getting the internal ID, you can just specify all the normal bytes (script key, anchor point, etc): https://github.com/lightninglabs/taproot-assets/blob/ec84760f1f665d16cf730c7e6ee8b30261023d9c/tapdb/assets_store.go#L904-L935

itsrachelfish:
@GeorgeTsagk @guggero I updated the code to filter assets based on the tweaked script key instead of the internal script key ID. I ran into some weird issues with the automated tests, for some reason I was unable to filter the 10th asset:

```
assetGen:    assetGen.assetGens[9],
anchorPoint: assetGen.anchorPoints[4],
amt:         777,
scriptKey:   scriptKeyWithScript,
```

Oli said this might have something to do with an underlying issue with the specific anchor point or asset generation, but I'm not sure why this particular asset is "cursed" :sweat_smile: 

I worked around the issue by adding a script key to the 3rd asset and was able to filter it as expected.

tapdb/assets_store.go
```diff
@@ -873,12 +873,38 @@ func (a *AssetStore) constraintsToDbFilter(
 				Valid: true,
 			}
 		}
+
+		if query.MaxAmt != 0 {
+			assetFilter.MaxAmt = sql.NullInt64{
```
guggero, line 878: nit: can use `sqlInt64(query.MaxAmt)` (and change it above for the `MinAmt` as well).

tapdb/assets_store.go
```diff
@@ -873,12 +873,38 @@ func (a *AssetStore) constraintsToDbFilter(
...
+		}
+
+		if query.AnchorPoint != nil {
+			anchorPointBytes, err := encodeOutpoint(
+				*query.AnchorPoint,
+			)
+			if err != nil {
+				return QueryAssetFilters{}, fmt.Errorf(
+					"unable to encode outpoint: %w", err,
+				)
```
guggero, line 902: nit: calls to `Errorf()` can have the closing parenthesis on the same line as the arguments (see formatting exception in the formatting doc).

rpcserver.go
```diff
@@ -1059,9 +1059,65 @@ func (r *rpcServer) ListAssets(ctx context.Context,
...
+			return nil, fmt.Errorf("error parsing group key: %w",
+				err)
+		}
+
+		// Construct an asset specifier using the requested group key.
+		groupSpecifier, err := asset.NewSpecifier(
+			nil, groupKey, nil, false,
+		)
+		if err != nil {
+			return nil, fmt.Errorf(
```
guggero, line 1080: nit: inconsistent with other formatting. Can be:
```
			return nil, fmt.Errorf("error creating asset "+
				"specifier: %w", err)
```

rpcserver.go
```diff
@@ -1059,9 +1059,65 @@ func (r *rpcServer) ListAssets(ctx context.Context,
 			"and include_leased")
 	}
 
+	constraints := tapfreighter.CommitmentConstraints{
+		MinAmt:         req.MinAmount,
+		MaxAmt:         req.MaxAmount,
+		CoinSelectType: tapsend.ScriptTreesAllowed,
+	}
+
+	if req.GroupKey != nil {
```
guggero, line 1068: Should check the length here instead, which asserts `nil` as well as an empty slice.

rpcserver.go
```diff
@@ -1059,9 +1059,65 @@ func (r *rpcServer) ListAssets(ctx context.Context,
...
+		groupKey, err := btcec.ParsePubKey(req.GroupKey[:])
```
guggero, line 1069: `req.GroupKey` is already a slice, no need to use slice operator.

rpcserver.go
```diff
@@ -1059,9 +1059,65 @@ func (r *rpcServer) ListAssets(ctx context.Context,
...
+		if err != nil {
+			return nil, fmt.Errorf("error parsing group key: %w",
+				err)
+		}
+
+		// Construct an asset specifier using the requested group key.
+		groupSpecifier, err := asset.NewSpecifier(
```
guggero, line 1076: Use `NewSpecifierFromGroupKey` instead.

rpcserver.go
```diff
@@ -1059,9 +1059,65 @@ func (r *rpcServer) ListAssets(ctx context.Context,
...
+		}
+
+		constraints.AssetSpecifier = groupSpecifier
+	}
+
+	filters := &tapdb.AssetQueryFilters{
+		CommitmentConstraints: constraints,
+	}
+
+	if req.ScriptKey != nil {
```
guggero, line 1092: Same here, should be length check, not `nil` check.

rpcserver.go
```diff
@@ -1059,9 +1059,65 @@ func (r *rpcServer) ListAssets(ctx context.Context,
 			"and include_leased")
 	}
 
+	constraints := tapfreighter.CommitmentConstraints{
+		MinAmt:         req.MinAmount,
+		MaxAmt:         req.MaxAmount,
+		CoinSelectType: tapsend.ScriptTreesAllowed,
```
guggero, line 1065: Should probably use `tapsend.DefaultCoinSelectType` here.

guggero (commented):
Almost there.

rpcserver.go
itsrachelfish: This results in a compiler error:

```
./rpcserver.go:1082:9: invalid argument: req.ScriptKey (variable of type *taprpc.ScriptKey) for built-in len
```

Other RPC methods such as `ExportProofRequest` define `ScriptKey` as bytes whereas `ListAssetRequest` uses the tap RPC `ScriptKey` struct.

Should `ListAssetRequest` be updated to use bytes instead of the `ScriptKey` struct? Or is there a different way I should approach this?

guggero: No, it's fine. Didn't see this was a struct when reviewing. So the `nil` check is okay, the `taprpc.UnmarshalScriptKey` will do the required length checks.

guggero (approved):
Nice, LGTM :tada:

PR #1203, opened 2024-11-19
Block on Proof Courier Service Connection Attempt

ffranr:
This PR enhances the reliability of the proof courier service and the robustness of the proof transfer process:  

- **Blocking Courier Service Connection:** Ensures courier service connection attempts are blocking, preventing failures from premature connection usage and simplifying debugging.  
- **Proof Transfer Resilience:**  
  - Implements logic to re-attempt proof transfers when backoff attempts are exhausted, avoiding delays until the next service restart.  
  - Refines logging and error messages to improve debugging.  

These updates strengthen the courier service's reliability and make proof transfers more fault-tolerant.

### DIFF
--- a/itest/tapd_harness.go
+++ b/itest/tapd_harness.go
@@ -270,7 +270,8 @@ func newTapdHarness(t *testing.T, ht *harnessTest, cfg tapdConfig,
 		BackoffCfg:         &hashmailBackoffCfg,
 	}
 	finalCfg.UniverseRpcCourier = &proof.UniverseRpcCourierCfg{
-		BackoffCfg: &universeRpcBackoffCfg,
+		BackoffCfg:            &universeRpcBackoffCfg,
+		ServiceRequestTimeout: 50 * time.Millisecond,
 	}
 
 	switch typedProofCourier := (opts.proofCourier).(type) {


itest/tapd_harness.go
```diff
@@ -257,7 +257,8 @@ func newTapdHarness(t *testing.T, ht *harnessTest, cfg tapdConfig,
 		BackoffCfg:         &hashmailBackoffCfg,
 	}
 	finalCfg.UniverseRpcCourier = &proof.UniverseRpcCourierCfg{
-		BackoffCfg: &universeRpcBackoffCfg,
+		BackoffCfg:            &universeRpcBackoffCfg,
+		ServiceRequestTimeout: 2 * time.Second,
```
guggero, line 261: If I understand the changes in this PR correctly, _this_ timeout should actually trigger (instead of the 5s in the previous commit) if the hashmail server isn't running or responding. So perhaps we can also get some of the "lost" seconds back by using a lower value here?

guggero (commented):
Changes to the logic make sense, thanks for the added robustness!
Some questions around some of the chosen values, since the integration tests take quite a while longer now.

itest/tapd_harness.go
```diff
@@ -80,7 +80,7 @@ const (
 	// defaultProofTransferReceiverAckTimeout is the default itest specific
 	// timeout we'll use for waiting for a receiver to acknowledge a proof
 	// transfer.
-	defaultProofTransferReceiverAckTimeout = 500 * time.Millisecond
+	defaultProofTransferReceiverAckTimeout = 5 * time.Second
```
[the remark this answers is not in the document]
ffranr: I think I've managed to remove this change in the latest commits.

itest/tapd_harness.go
ffranr: I think I've managed to lower this significantly in the latest commits.

guggero (approved):
Nice, LGTM :tada: 

Current CI failures are known flakes or fixed in another PR (LiT itest).

Roasbeef:
IIRC, the issue last time was that an address contained an invalid courier addr, so the proof could never be delivered, meaning the send was never completed. 

Does this change resolve that? IIUC now, we'll just block when trying to connect, but _after_ the send has already been confirmed on chain?

ffranr:
> IIRC, the issue last time was that an address contained an invalid courier addr, so the proof could never be delivered, meaning the send was never completed.
> 
> Does this change resolve that? IIUC now, we'll just block when trying to connect, but _after_ the send has already been confirmed on chain?

@Roasbeef The PR does not address the invalid courier address problem. This PR ensures that we don't try to use a proof courier service connection before it's ready and so potentially avoid an unnecessary backoff and log messages.

proof/courier.go
```diff
@@ -293,6 +293,7 @@ func serverDialOpts() ([]grpc.DialOption, error) {
 	tlsConfig := tls.Config{InsecureSkipVerify: true}
 	transportCredentials := credentials.NewTLS(&tlsConfig)
 	opts = append(opts, grpc.WithTransportCredentials(transportCredentials))
+	opts = append(opts, grpc.WithBlock())
```
jharveyb, line 296: 💯

proof/courier.go
```diff
@@ -1154,10 +1155,17 @@ func (h *HashMailCourier) SetSubscribers(
...
+// nolint:lll
 type UniverseRpcCourierCfg struct {
 	// BackoffCfg configures the behaviour of the proof delivery
 	// functionality.
 	BackoffCfg *BackoffCfg
+
+	// ServiceRequestTimeout defines the maximum duration we'll wait for
+	// a courier service to handle our outgoing request during a connection
+	// attempt, or when delivering or retrieving a proof.
+	ServiceRequestTimeout time.Duration `long:"servicerequestimeout" description:"The maximum duration we'll wait for a courier service to handle our outgoing request during a connection attempt, or when delivering or retrieving a proof."`
```
jharveyb, line 1168: Shouldn't the hashmail courier / every courier implementation have a similar timeout? And then use it on initialization.

For hashmail I think that would be in `NewHashMailBox`, you could add the same child `ctx` object as you're doing for the Universe courier.

jharveyb (commented):
Looks mostly good! Just one note.

Also, for the third commit, the comment should say `async` not `sync` ?

proof/courier.go
ffranr: Yes you're right! Nice catch. Hahsmail is blocking also now, so would be good to have that timeout.

dstadulis:
!lightninglabs-deploy mute 720h30m

guggero (approved):
I've rebased this after https://github.com/lightninglabs/taproot-assets/pull/1412 already made the connection blocking.
So this PR now just adds a configurable timeout to the initial blocking attempt and also a re-try mechanism when all backoff attempts have failed.

guggero:
Hmm, not sure what happened in the merge queue... Rebased to re-trigger CI.

PR #1209, opened 2024-11-21
[custom channels]: generate unique script keys for HTLCs

guggero:
Fixes the LiT integration tests that were broken after merging https://github.com/lightninglabs/taproot-assets/pull/1181.
But that basically showed us that colliding script keys for MPP shard HTLCs are actually a problem.

This PR tweaks each script key with the HTLC index, making them unique again, even if they have the same payment hash and timeout (which results in the same BTC level taproot output key).

To make the accompanying [`litd` integration tests](https://github.com/lightninglabs/lightning-terminal/pull/908) pass, we also fix the following two bugs:
 - Calculate the sweep transaction change output index correctly and dynamically to avoid invalid exclusion proofs when there are multiple HTLCs being swept at the same time.
 - Fix an issue with `UpsertScriptKey` that wouldn't allow the tweak of a script key to be updated when it is learned after the key was already added to the DB.

### DIFF
--- a/tapchannel/allocation_sort.go
+++ b/tapchannel/allocation_sort.go
@@ -2,7 +2,8 @@ package tapchannel
 
 import (
 	"bytes"
-	"sort"
+	"cmp"
+	"slices"
 )
 
 // InPlaceAllocationSort performs an in-place sort of output allocations.
@@ -14,51 +15,21 @@ import (
 // transactions, the script does not directly commit to them. Instead, the CLTVs
 // must be supplied separately to act as a tie-breaker, otherwise we may produce
 // invalid HTLC signatures if the receiver produces an alternative ordering
-// during verification.
+// during verification. Because multiple shards of the same MPP payment can be
+// identical in all other fields, we also use the HtlcIndex as a final
+// tie-breaker.
 //
-// NOTE: Commitment and commitment anchor outputs should have a 0 CLTV value.
+// NOTE: Commitment and commitment anchor outputs should have a 0 CLTV and
+// HtlcIndex value.
 func InPlaceAllocationSort(allocations []*Allocation) {
-	sort.Sort(sortableAllocationSlice{allocations})
-}
-
-// sortableAllocationSlice is a slice of allocations and the corresponding CLTV
-// values of any HTLCs. Commitment and commitment anchor outputs should have a
-// CLTV of 0.
-type sortableAllocationSlice struct {
-	allocations []*Allocation
-}
-
-// Len returns the length of the sortableAllocationSlice.
-//
-// NOTE: Part of the sort.Interface interface.
-func (s sortableAllocationSlice) Len() int {
-	return len(s.allocations)
-}
-
-// Swap exchanges the position of outputs i and j.
-//
-// NOTE: Part of the sort.Interface interface.
-func (s sortableAllocationSlice) Swap(i, j int) {
-	s.allocations[i], s.allocations[j] = s.allocations[j], s.allocations[i]
-}
-
-// Less is a modified BIP69 output comparison, that sorts based on value, then
-// pkScript, then CLTV value.
-//
-// NOTE: Part of the sort.Interface interface.
-func (s sortableAllocationSlice) Less(i, j int) bool {
-	allocI, allocJ := s.allocations[i], s.allocations[j]
-
-	if allocI.BtcAmount != allocJ.BtcAmount {
-		return allocI.BtcAmount < allocJ.BtcAmount
-	}
-
-	pkScriptCmp := bytes.Compare(
-		allocI.SortTaprootKeyBytes, allocJ.SortTaprootKeyBytes,
-	)
-	if pkScriptCmp != 0 {
-		return pkScriptCmp < 0
-	}
-
-	return allocI.CLTV < allocJ.CLTV
+	slices.SortFunc(allocations, func(i, j *Allocation) int {
+		return cmp.Or(
+			cmp.Compare(i.BtcAmount, j.BtcAmount),
+			bytes.Compare(
+				i.SortTaprootKeyBytes, j.SortTaprootKeyBytes,
+			),
+			cmp.Compare(i.CLTV, j.CLTV),
+			cmp.Compare(i.HtlcIndex, j.HtlcIndex),
+		)
+	})
 }
--- a/tapchannel/aux_leaf_signer.go
+++ b/tapchannel/aux_leaf_signer.go
@@ -3,12 +3,15 @@ package tapchannel
 import (
 	"bytes"
 	"fmt"
+	"math"
+	"math/big"
 	"sync"
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcutil/psbt"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 	"github.com/lightninglabs/taproot-assets/address"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/commitment"
@@ -750,4 +753,109 @@ func (v *schnorrSigValidator) validateSchnorrSig(virtualTx *wire.MsgTx,
 	}
 
 	return nil
+}
+
+// ScriptKeyTweakFromHtlcIndex converts the given HTLC index into a modulo N
+// scalar that can be used to tweak the internal key of the HTLC script key on
+// the asset level. The value of 1 is always added to the index to make sure
+// this value is always non-zero.
+func ScriptKeyTweakFromHtlcIndex(index input.HtlcIndex) *secp256k1.ModNScalar {
+	// If we're at math.MaxUint64, we'd wrap around to 0 if we incremented
+	// by 1, but we need to make sure the tweak is 1 to not cause a
+	// multiplication by zero. This should never happen, as it would mean we
+	// have more than math.MaxUint64 updates in a channel, which exceeds the
+	// protocol's maximum.
+	if index == math.MaxUint64 {
+		return new(secp256k1.ModNScalar).SetInt(1)
+	}
+
+	// We need to avoid the tweak being zero, so we always add 1 to the
+	// index. Otherwise, we'd multiply G by zero.
+	index++
+
+	indexAsBytes := new(big.Int).SetUint64(index).Bytes()
+	indexAsScalar := new(secp256k1.ModNScalar)
+	_ = indexAsScalar.SetByteSlice(indexAsBytes)
+
+	return indexAsScalar
+}
+
+// TweakPubKeyWithIndex tweaks the given internal public key with the given
+// HTLC index. The tweak is derived from the index in a way that never results
+// in a zero tweak. The value of 1 is always added to the index to make sure
+// this value is always non-zero. The public key is tweaked like this:
+//
+//	tweakedKey = key + (index+1) * G
+func TweakPubKeyWithIndex(pubKey *btcec.PublicKey,
+	index input.HtlcIndex) *btcec.PublicKey {
+
+	// Avoid panic if input is nil.
+	if pubKey == nil {
+		return nil
+	}
+
+	// We need to operate on Jacobian points, which is just a different
+	// representation of the public key that allows us to do scalar
+	// multiplication.
+	var (
+		pubKeyJacobian, tweakTimesG, tweakedKey btcec.JacobianPoint
+	)
+	pubKey.AsJacobian(&pubKeyJacobian)
+
+	// Derive the tweak from the HTLC index in a way that never results in
+	// a zero tweak. Then we multiply G by the tweak.
+	tweak := ScriptKeyTweakFromHtlcIndex(index)
+	secp256k1.ScalarBaseMultNonConst(tweak, &tweakTimesG)
+
+	// And finally we add the result to the key to get the tweaked key.
+	secp256k1.AddNonConst(&pubKeyJacobian, &tweakTimesG, &tweakedKey)
+
+	// Convert the tweaked key back to an affine point and create a new
+	// taproot key from it.
+	tweakedKey.ToAffine()
+	return btcec.NewPublicKey(&tweakedKey.X, &tweakedKey.Y)
+}
+
+// TweakHtlcTree tweaks the internal key of the given HTLC script tree with the
+// given index, then returns the tweaked tree with the updated taproot key.
+// The tapscript tree and tapscript root are not modified.
+// The internal key is tweaked like this:
+//
+//	tweakedInternalKey = internalKey + (index+1) * G
+func TweakHtlcTree(tree input.ScriptTree,
+	index input.HtlcIndex) input.ScriptTree {
+
+	// The tapscript tree and root are not modified, only the internal key
+	// is tweaked, which inherently modifies the taproot key.
+	tweakedInternalPubKey := TweakPubKeyWithIndex(tree.InternalKey, index)
+	newTaprootKey := txscript.ComputeTaprootOutputKey(
+		tweakedInternalPubKey, tree.TapscriptRoot,
+	)
+
+	return input.ScriptTree{
+		InternalKey:   tweakedInternalPubKey,
+		TaprootKey:    newTaprootKey,
+		TapscriptTree: tree.TapscriptTree,
+		TapscriptRoot: tree.TapscriptRoot,
+	}
+}
+
+// AddTweakWithIndex adds the given index to the given tweak. If the tweak is
+// empty, the index is used as the tweak directly. The value of 1 is always
+// added to the index to make sure this value is always non-zero.
+func AddTweakWithIndex(maybeTweak []byte, index input.HtlcIndex) []byte {
+	indexTweak := ScriptKeyTweakFromHtlcIndex(index)
+
+	// If we don't already have a tweak, we just use the index as the tweak.
+	if len(maybeTweak) == 0 {
+		return fn.ByteSlice(indexTweak.Bytes())
+	}
+
+	// If we have a tweak, we need to parse/decode it as a scalar, then add
+	// the index as a scalar, and encode it back to a byte slice.
+	tweak := new(secp256k1.ModNScalar)
+	_ = tweak.SetByteSlice(maybeTweak)
+	newTweak := tweak.Add(indexTweak)
+
+	return fn.ByteSlice(newTweak.Bytes())
 }
--- a/tapchannel/aux_leaf_signer.go
+++ b/tapchannel/aux_leaf_signer.go
@@ -370,7 +370,7 @@ func verifyHtlcSignature(chainParams *address.ChainParams,
 
 	vPackets, err := htlcSecondLevelPacketsFromCommit(
 		chainParams, chanState, commitTx, baseJob.KeyRing, htlcOutputs,
-		baseJob, htlcTimeout,
+		baseJob, htlcTimeout, baseJob.HTLC.HtlcIndex,
 	)
 	if err != nil {
 		return fmt.Errorf("error generating second level packets: %w",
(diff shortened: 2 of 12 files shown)


tapchannel/commitment.go
```diff
@@ -730,6 +730,13 @@ func CreateAllocations(chanState lnwallet.AuxChanState, ourBalance,
 				"allocation, HTLC is dust")
 		}
 
+		// To ensure uniqueness of the script key across HTLCs with the
+		// same payment hash and timeout (which would be equal
+		// otherwise), we tweak the asset level internal key of the
+		// script key with the HTLC index. We'll ONLY use this for the
+		// asset level, NOT for the BTC level.
+		tweakedTree := TweakHtlcTree(htlcTree, htlc.HtlcIndex)
```
Roasbeef, line 738: The other relevant area here is in the aux sweeper, when we go to derive the control block and witness scripts.

tapchannel/aux_leaf_signer_test.go
```diff
@@ -394,3 +396,61 @@ func TestTweakHtlcTree(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tweak := AddTweakWithIndex(tc.tweak, tc.index)
+			resultBytes := tc.result.Bytes()
+			resultBigInt := new(big.Int).SetBytes(resultBytes[:])
+			tweakBigInt := new(big.Int).SetBytes(tweak)
+			// 36893488147419103230
```
Roasbeef, line 448: Stray comment?

tapchannel/aux_sweeper.go
```diff
@@ -314,7 +314,7 @@ func (a *AuxSweeper) signSweepVpackets(vPackets []*tappsbt.VPacket,
 		// signature.
 		signingKey, leafToSign := applySignDescToVIn(
 			signDesc, vIn, &a.cfg.ChainParams,
-			tapscriptDesc.scriptTree.TapTweak(),
+			tapscriptDesc.scriptTree.TapTweak(), 0, false,
```
gijswijs, line 317: IIUC the `index` value here has no meaning since `tweakWithIndex` is set to `false`.
Wouldn't the Functional Options Pattern be more elegant for `applySignDescToVIn`? You would get something like...
```go
signingKey, leafToSign := applySignDescToVIn(
	signDesc, vIn, s.cfg.ChainParams, tapscriptRoot,
	WithIndexTweak(baseJob.HTLC.HtlcIndex)
)

```
when you *do* want to tweak, and...
```go
signingKey, leafToSign := applySignDescToVIn(
	signDesc, vIn, &a.cfg.ChainParams,
	tapscriptDesc.scriptTree.TapTweak()
)
```
when you don't.

tapchannel/aux_leaf_signer.go
```diff
@@ -735,3 +753,102 @@ func (v *schnorrSigValidator) validateSchnorrSig(virtualTx *wire.MsgTx,
 
 	return nil
 }
+
+// HtlcIndexAsScriptKeyTweak converts the given HTLC index into a modulo N
+// scalar that can be used to tweak the internal key of the HTLC script key on
+// the asset level. The value of 1 is always added to the index to make sure
+// this value is always non-zero.
+func HtlcIndexAsScriptKeyTweak(index input.HtlcIndex) *secp256k1.ModNScalar {
```
gijswijs, line 761: All four functions below are exported, but aren't used outside of the `tapchannel` package in this PR. Do we expect the need of those functions outside of this package?

gijswijs (commented):
Very nice. I did a thorough review and left a bunch of comments throughout, but no blockers AFAICT.

tapchannel/commitment.go
guggero: Yes. I've now based this PR on your sweeper PR and added that code.

tapchannel/aux_leaf_signer_test.go
```diff
@@ -163,3 +169,288 @@ func makeCommitSig(t *testing.T, numAssetIDs, numHTLCs int) *lnwire.CommitSig {
...
+			tweakedTree := TweakHtlcTree(tc.tree, tc.index)
+			require.Equal(
+				t, tc.result.InternalKey.SerializeCompressed(),
+				tweakedTree.InternalKey.SerializeCompressed(),
+			)
+			require.Equal(
+				t, tc.result.TaprootKey.SerializeCompressed(),
+				tweakedTree.TaprootKey.SerializeCompressed(),
+			)
+			require.Equal(t, tc.result, tweakedTree)
```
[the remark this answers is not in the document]
guggero: There is some overlap in the two checks. But comparing the serialized keys just made debugging way easier. And the `require.Equal()` check then still make sure the other fields (tapscript tree and root) are equal.

tapchannel/aux_sweeper.go
guggero: Removed this whole code as it turned out to be wrong anyway.

tapchannel/aux_leaf_signer.go
guggero: You're correct. But such stateless and potentially general purpose functions IMO don't hurt to be exported.

guggero:
I rebased this on top of https://github.com/lightninglabs/taproot-assets/pull/1154 to verify everything works.
Only the last two commits are for this PR!

@Roasbeef I have run a full itest suite on top of this PR to validate it works.
Since we're tweaking the HTLC revocation key (which is the internal key of the script key), we currently don't need to tweak anything at the **signing** level, since revocation sweeps aren't currently implemented. Left a big TODO in there so we don't forget when we add that.
So I guess this approach is even more minimal as initially thought (at least for now).

tapchannel/aux_leaf_signer_test.go
```diff
@@ -163,3 +169,288 @@ func makeCommitSig(t *testing.T, numAssetIDs, numHTLCs int) *lnwire.CommitSig {
...
+			require.Equal(t, tc.result, tweak)
+		})
+	}
+}
+
+// TestTweakPubKeyWithIndex tests the TweakPubKeyWithIndex function.
+func TestTweakPubKeyWithIndex(t *testing.T) {
+	randNum := test.RandInt[uint32]()
+
+	makePubKey := func(tweak uint64) *btcec.PublicKey {
```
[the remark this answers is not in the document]
gijswijs: The below test case for `TestTweakPubKeyWithIndex` would still break the test because of the overflow:
```go
{
        name:   "break test",
        pubKey: startKey,
        index:  math.MaxUint64 - randNum,
        result: pubKeyFromUint64(randNum + math.MaxUint64 - randNum + 1),
},
```
It's such a weird edge case (in a unit test no less), we might as well leave it like this. I'm just proud that I thought of an overflow to break stuff. :wink:

tapchannel/aux_sweeper.go
gijswijs: IIUC you created the func `TweakedRevocationKeyRing` which tweaks the `RevocationKey` and then use that keyring in the local and remote sweeps, both for timeout and success. Since we are *in* the aux_sweeper anyway, we know for a fact we need to tweak it with the `index` so no `tweakWithIndex` flag is needed.

Have I got that right?

gijswijs:
We are nearly there. Just two minor nits and a comment to check whether I understand the code.

jharveyb:
> I rebased this on top of #1154 to verify everything works. Only the last two commits are for this PR!
> 

Should the base branch for PR be changed then, or does this supercede #1154 directly?

tapchannel/aux_sweeper.go
guggero: Yes, we need to tweak it for every sweep of an HTLC output. Since we currently only implement timeout and success sweeps, the tweak will only affect the internal key of the control block (because we're executing a script path in both cases). Once we implement revocation, then we'll also need to tweak the signing key, as then we'll use the key path.
Not sure if that's what you meant here?

tapchannel/aux_leaf_signer_test.go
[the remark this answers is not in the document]
guggero: Yes, good catch! Added a simplified version of your test case in, alongside with a comment why we're currently okay with this.

tapchannel/aux_sweeper.go
gijswijs: Thanks for the clarification. Yes, that's what I meant.

gijswijs (approved):
LGTM! :tada:

guggero:
Thanks for the review!

Unfortunately while testing this in an itest, I ran into a bit of a road block.
Turns out the tweaking works well. BUT, the thing we tweak the keys with (`HtlcIndex`) is not stable across the two peers of the channel.
Meaning that for peer A, its `HtlcIndex` of the **outgoing** HTLCs is not necessarily exactly the same as for peer B's **incoming** HTLCs. So after a certain number of HTLCs (the number seems to be around 10 in the itest, but could be pure coincidence), the ordering is out of sync and because of that the two peers have different commitment transactions, which leads to a force close.

I think we might actually need to introduce a stable sort index as a custom records to the HTLC, so both parties can use the correct value when sorting.

Roasbeef:
Does the in-place allocation sort need to factor in the htlc index now? 

Before this change, they were identical for a given `(amt, payhash, cltv)` HTLC tuple. Now as the internal key is tweaked by the index, they'll no longer be identical. That can cause us to mix up two allocations that have the same HTLC tuple, but differ based on their index. By default `sort.Sort` isn't stable, so the slice ordering can be changed if all the keys match.

tapchannel/allocation_sort.go
```diff
@@ -43,7 +43,7 @@ func (s sortableAllocationSlice) Swap(i, j int) {
 }
 
 // Less is a modified BIP69 output comparison, that sorts based on value, then
-// pkScript, then CLTV value.
+// pkScript, then CLTV value and finally HtlcIndex.
 //
 // NOTE: Part of the sort.Interface interface.
 func (s sortableAllocationSlice) Less(i, j int) bool {
```
jharveyb, line 49: non-blocking nit: AFAICT the stdlib recommendation now is to use `slices.SortFunc` instead of `sort.Sort`.

tapchannel/aux_leaf_signer.go
```diff
@@ -751,3 +753,102 @@ func (v *schnorrSigValidator) validateSchnorrSig(virtualTx *wire.MsgTx,
...
+	tweak := HtlcIndexAsScriptKeyTweak(index)
+	secp256k1.ScalarBaseMultNonConst(tweak, &tweakTimesG)
+
+	// And finally we add the result to the key to get the tweaked key.
+	secp256k1.AddNonConst(&pubKeyJacobian, &tweakTimesG, &tweakedKey)
+
+	// Convert the tweaked key back to an affine point and create a new
+	// taproot key from it.
+	tweakedKey.ToAffine()
+	return btcec.NewPublicKey(&tweakedKey.X, &tweakedKey.Y)
```
jharveyb, line 809: From the godoc of `NewPublicKey`; does this output key need to be checked with `IsOnCurve`?

My guess is no since we know the input was on the curve.

tapchannel/aux_leaf_signer_test.go
```diff
@@ -163,3 +179,287 @@ func makeCommitSig(t *testing.T, numAssetIDs, numHTLCs int) *lnwire.CommitSig {
...
+
+	testCases := []struct {
+		name   string
+		tree   input.ScriptTree
+		index  uint64
+		result input.ScriptTree
+	}{
+		{
+			name: "index 0",
+			tree: input.ScriptTree{
```
jharveyb, line 327: super-nit: `tree` is already the same here and handled as read-only, so could be deduplicated.

jharveyb (approved):
LGTM! Massive lift 🚀 

I think the commit message for the first commit is now wrong btw? Wrt. mentioning that the tweaking for revocation is not implemented.

tapchannel/aux_leaf_signer.go
```diff
@@ -751,3 +753,102 @@ func (v *schnorrSigValidator) validateSchnorrSig(virtualTx *wire.MsgTx,
...
+	index++
+
+	// If we wrapped around from math.MaxUint64 to 0, we need to make sure
+	// the tweak is 1 to not cause a multiplication by zero. This should
+	// never happen, as it would mean we have more than math.MaxUint64
+	// updates in a channel, which exceeds the protocol's maximum.
+	if index == 0 {
+		return new(secp256k1.ModNScalar).SetInt(1)
+	}
+
```
ffranr, line 773: To avoid overflowing `index` we could do something like this instead:

```
	if index == math.MaxUint64 {
		return new(secp256k1.ModNScalar).SetInt(1)
	}

        index++

        ...
```

ffranr: Also, on the function name, perhaps `ScriptKeyTweakFromHtlcIndex`.

tapchannel/aux_leaf_signer.go
```diff
@@ -751,3 +753,102 @@ func (v *schnorrSigValidator) validateSchnorrSig(virtualTx *wire.MsgTx,
...
+func TweakPubKeyWithIndex(pubKey *btcec.PublicKey,
+	index input.HtlcIndex) *btcec.PublicKey {
+
+	// We need to operate on Jacobian points, which is just a different
+	// representation of the public key that allows us to do scalar
+	// multiplication.
+	var (
+		pubKeyJacobian, tweakTimesG, tweakedKey btcec.JacobianPoint
+	)
+	pubKey.AsJacobian(&pubKeyJacobian)
```
ffranr, line 796: nit: ensure that `pubKey` is not `nil`.

tapchannel/aux_leaf_signer.go
```diff
@@ -751,3 +753,102 @@ func (v *schnorrSigValidator) validateSchnorrSig(virtualTx *wire.MsgTx,
...
+}
+
+// TweakPubKeyWithIndex tweaks the given internal public key with the given
+// HTLC index. The tweak is derived from the index in a way that never results
+// in a zero tweak. The value of 1 is always added to the index to make sure
+// this value is always non-zero. The public key is tweaked like this:
+//
+//	tweakedKey = key + (index+1) * G
+func TweakPubKeyWithIndex(pubKey *btcec.PublicKey,
+	index input.HtlcIndex) *btcec.PublicKey {
```
ffranr, line 788: I feel as if this function and the one above should be generalised into "tweak public key by integer" function(s). Which can live in btcec eventually. None of this logic is unique to taproot assets.

tapchannel/allocation_sort.go
guggero: Makes sense. Also found a way to simplify the code quite a bit.

tapchannel/aux_leaf_signer.go
guggero: Yeah, we assume the input key is on the curve.

guggero:
> I think the commit message for the first commit is now wrong btw? Wrt. mentioning that the tweaking for revocation is not implemented.

No, it's correct. Sweeping an HTLC of a revoked commitment (old state) isn't implemented yet. In that case we would need to tweak the signing key at sign time. In any non-revocation (e.g. normal, latest-state force close), we only need to tweak the internal key when we create the output and then again when we create the control block. But no _signing_ private key needs to be tweaked as of now.

gijswijs (approved):
No further remarks, only questions to see if I understand it. Also the last two commits fix problems that are unrelated to HTLC script key tweak, am I right?
The change output index calculation and the upsert of tweaks are unrelated to the initial problem AFAICT.

tl;dr: LGTM! :tada:

tapchannel/commitment.go
```diff
@@ -766,13 +780,13 @@ func CreateAllocations(chanState lnwallet.AuxChanState, ourBalance,
 			NonAssetLeaves: sibling,
 			ScriptKey: asset.ScriptKey{
 				PubKey: asset.NewScriptKey(
-					htlcTree.TaprootKey,
+					tweakedTree.TaprootKey,
```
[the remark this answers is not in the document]
guggero: Yes, correct. We need to make sure the sort order is stable _before_ applying any tweaks (TAP leaf or HtlcIndex). Added a comment to make this clear.

tapchannel/commitment.go
```diff
@@ -1289,12 +1317,12 @@ func createSecondLevelHtlcAllocations(chanType channeldb.ChannelType,
 		),
 		InternalKey:    htlcTree.InternalKey,
 		NonAssetLeaves: sibling,
-		ScriptKey:      asset.NewScriptKey(htlcTree.TaprootKey),
+		ScriptKey:      asset.NewScriptKey(tweakedTree.TaprootKey),
 		SortTaprootKeyBytes: schnorr.SerializePubKey(
```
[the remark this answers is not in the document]
guggero: Added a comment here too.

guggero:
> Also the last two commits fix problems that are unrelated to HTLC script key tweak, am I right?

Correct. But those changes were required to get the integration tests passing, so they were required follow-up fixes.
Going to update the PR body with what else is fixed for future reference :+1:

tapchannel/allocation_sort.go
```diff
@@ -14,51 +15,21 @@ import (
...
-	pkScriptCmp := bytes.Compare(
-		allocI.SortTaprootKeyBytes, allocJ.SortTaprootKeyBytes,
-	)
-	if pkScriptCmp != 0 {
-		return pkScriptCmp < 0
-	}
-
-	return allocI.CLTV < allocJ.CLTV
+	slices.SortFunc(allocations, func(i, j *Allocation) int {
+		return cmp.Or(
```
Roasbeef, line 26: 👍

tapchannel/aux_sweeper.go
```diff
@@ -2256,26 +2256,20 @@ func (a *AuxSweeper) sweepContracts(inputs []input.Input,
...
 		// We only need to generate an exclusion proof for the second
 		// output in the commitment transaction.
-		//
-		// TODO(roasbeef): case of no change?
 		target.ExclusionProofs = append(
 			target.ExclusionProofs, proof.TaprootProof{
-				OutputIndex:    1,
-				InternalKey:    sweepInternalKey.PubKey,
-				TapscriptProof: tsProof,
+				OutputIndex: changeOutputIndex,
```
Roasbeef, line 2268: Ahh good catch!

tapchannel/aux_sweeper.go
```diff
@@ -2256,26 +2256,20 @@ func (a *AuxSweeper) sweepContracts(inputs []input.Input,
 // sweepExclusionProofGen is a helper function that generates an exclusion
 // proof for the internal key of the change output.
 func sweepExclusionProofGen(sweepInternalKey keychain.KeyDescriptor,
-) tapsend.ExclusionProofGenerator {
+	changeOutputIndex uint32) tapsend.ExclusionProofGenerator {
 
 	return func(target *proof.BaseProofParams,
 		isAnchor tapsend.IsAnchor) error {
 
-		tsProof, err := proof.CreateTapscriptProof(nil)
```
Roasbeef, line 2243: Why don't we need to call this any longer? Did it just amount to: 
```go
&proof.TapscriptProof{
					Bip86: true,
				},
```
?

Roasbeef (approved):
LGTM 🦋

tapdb/sqlc/queries/assets.sql
```diff
@@ -862,7 +862,13 @@ INSERT INTO script_keys (
                          WHEN script_keys.declared_known IS NULL OR script_keys.declared_known = FALSE
                          THEN COALESCE(EXCLUDED.declared_known, script_keys.declared_known)
                          ELSE script_keys.declared_known
-                       END
+                       END,
+      -- If the tweak was previously unknown, we'll update to the new value.
+      tweak = CASE
```
Roasbeef, line 867: Non-blocking, but can extend the recently added TestScriptKeyKnownUpsert test to cover this

tapchannel/aux_sweeper.go
guggero: Yeah, it's equivalent. But I just liked the more explicit version more, since we explicitly _know_ it's a BIP-86 here.

tapdb/sqlc/queries/assets.sql
guggero: Yeah, makes sense. Added.

Issue #1207, opened 2024-11-21
[feature]: Enable asset group witness creation using external PSBT signer during minting


ffranr:
This issue must be completed first: https://github.com/lightninglabs/taproot-assets/issues/1226

---

This issue proposes a new feature to enhance the asset minting process. It introduces the option for a minter to generate the asset group witness by signing a PSBT using an external signer, independent of tapd. This will enable the minter to securely store the private asset group key in cold storage throughout the minting process.

### Batch Minting Overview

At a high level, the current batch minting process involves the following steps:

1. The `MintAsset` RPC endpoint is called to add a new asset to a minting batch. This batch remains open across multiple calls to this endpoint.
2. The `SealBatch` RPC endpoint is invoked, triggering the `tapgarden.ChainPlanter`, which generates asset group witnesses where applicable. This step blocks the `tapgarden.ChainPlanter` instance until complete.
3. The `FinalizeBatch` RPC endpoint is called, initiating the `tapgarden.ChainPlanter`. If the batch has not already been sealed, it is sealed at this stage. A new `tapgarden.BatchCaretaker` state machine is then started, anchoring the Taproot artifacts to a Bitcoin transaction, which is subsequently broadcast.

If a batch has not been sealed using the `SealBatch` RPC endpoint then the `FinalizeBatch` RPC endpoint will call into the same batch sealing routine.

### Minimal Solution

A minimal solution could take the following approach:

- Extend the `Seedling` (asset minting request) to include a new flag when passed into the `MintAsset` RPC endpoint. This flag indicates that the asset group witness should be generated using the new external PSBT signing flow. This ensures that if the `FinalizeBatch` RPC endpoint is called prematurely, the default batch sealing process fails, avoiding unintended behavior.

- Introduce a new RPC endpoint, `QuerySealBatchPSBTs`, which returns all PSBTs required for the minter to externally sign and generate asset group witnesses for each flagged seedling in the batch. This endpoint should send a new request type to `tapgarden.ChainPlanter` which will need to be refactored where relevant. The endpoint name seems appropriate, as "seal" appears to refer almost exclusively to asset witness creation based on the code.

- The signed PSBTs are then submitted by the signer through the `SealBatch` RPC endpoint. The endpoint uses the newly generated group witnesses to seal the batch as usual.

---

See also this LND signing gist: https://gist.github.com/guggero/569241aa9fec57e287101187bd28d1c5

ffranr:
[...]

PR #1210, opened 2024-11-21
Mint with external asset group PSBT signer

ffranr:
This PR updates the `ChainPlanter` to enable the extraction of data required for an external signer to generate and produce the asset group witness.

This is a work in progress, and I am seeking early feedback on the approach. It follows the solution implementation proposed here: https://github.com/lightninglabs/taproot-assets/issues/1207

I think `SeedlingExternalSealPkg` should contain everything that the external signer needs to sign the PSBT.


ffranr:
We can probably close this in favour of https://github.com/lightninglabs/taproot-assets/pull/1214

guggero:
> We can probably close this in favour of #1214

Yes, I agree. Looks like #1214 is farther along.

Issue #1208, opened 2024-11-21
[bug]: Unable to send assets after creating a wallet using an existing cipher seed mnemonic.


youngzhenhao:
### Background

Sending assets after creating a wallet using an existing cipher seed mnemonic, this error returns: `[tapcli] unable to send assets: rpc error: code = Unknown desc = unable to sign and commit virtual packet: unable to generate Taproot Asset witness data: invalid transfer asset witness:`.

#### lncli walletbalance
```json
{
    "total_balance": "7473",
    "confirmed_balance": "7473",
    "unconfirmed_balance": "0",
    "locked_balance": "0",
    "reserved_balance_anchor_chan": "0",
    "account_balance": {
        "default": {
            "confirmed_balance": "7473",
            "unconfirmed_balance": "0"
        }
    }
}
```

#### tapcli a l
```json
{
    "assets":  [
        {
            "version":  "ASSET_VERSION_V0",
            "asset_genesis":  {
                "genesis_point":  "e27eef0b269ad501df2bb10f2eb78ebbc6b7d50b0fd7b863d01d93da8a65f1ce:0",
                "name":  "乖乖🐉",
                "meta_hash":  "3361645a88c4dde3a723f3353355c2a70321b9de6cc1da55260a52eb0aaebc79",
                "asset_id":  "b832d1c158e198609941a3be5e646504363fbee8d87d89750a5436fa4d4c8d77",
                "asset_type":  "NORMAL",
                "output_index":  0
            },
            "amount":  "50",
            "lock_time":  0,
            "relative_lock_time":  0,
            "script_version":  0,
            "script_key":  "02a4d1248dbeab217f3238424b83b012640170be6f25207c6680aa1ab8392c6dc6",
            "script_key_is_local":  true,
            "asset_group":  null,
            "chain_anchor":  {
[...]

PR #1213, opened 2024-11-22
[wallet]: fix signing interaction with lnd, allowing restore from seed

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1208.

Depends on https://github.com/lightninglabs/lndclient/pull/203.

Fixes an issue with how we tell `lnd` what key to sign. This basically boils down to a weird behavior in `lndclient` that we now circumvent with a new signing method.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,9 +26,9 @@ require (
 	github.com/lib/pq v1.10.9
 	github.com/lightninglabs/aperture v0.3.2-beta.0.20241015115230-d59b5514c19a
 	github.com/lightninglabs/lightning-node-connect/hashmailrpc v1.0.2
-	github.com/lightninglabs/lndclient v0.18.4-5
+	github.com/lightninglabs/lndclient v0.18.4-7
 	github.com/lightninglabs/neutrino/cache v1.1.2
-	github.com/lightningnetwork/lnd v0.18.3-beta.rc3.0.20241120143113-9246d5c51cd2
+	github.com/lightningnetwork/lnd v0.18.4-beta.rc1
 	github.com/lightningnetwork/lnd/cert v1.2.2
 	github.com/lightningnetwork/lnd/clock v1.1.1
 	github.com/lightningnetwork/lnd/fn v1.2.3
--- a/itest/send_test.go
+++ b/itest/send_test.go
@@ -1954,6 +1954,91 @@ func testSendNoCourierUniverseImport(t *harnessTest) {
 	AssertSendEventsComplete(t.t, receiveAddr.ScriptKey, sendEvents)
 }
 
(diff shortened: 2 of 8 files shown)


itest/send_test.go
```diff
@@ -1954,6 +1954,87 @@ func testSendNoCourierUniverseImport(t *harnessTest) {
...
+	// Send more assets after restoring the node.
+	aliceAddr, err = alice.NewAddr(ctxb, &taprpc.NewAddrRequest{
+		AssetId: genInfo.AssetId,
+		Amt:     sendAmount,
+	})
+	require.NoError(t.t, err)
+
+	AssertAddrCreated(t.t, alice, rpcAsset, aliceAddr)
+
+	sendAsset(
```
Roasbeef: So when this fails, it's because the wrong key was derived? Is it possibly us (`tapd`) passing in the incorrect information? Like an issue in a db query somewhere or something.

guggero: Yeah, the wrong key is derived. The default case is family 0 and index 0 if the key isn't found by pubkey only.
So nothing wrong in `tapd`, just weird `lndclient` behavior.

go.mod
```diff
@@ -26,9 +26,9 @@ require (
 	github.com/lib/pq v1.10.9
 	github.com/lightninglabs/aperture v0.3.2-beta.0.20241015115230-d59b5514c19a
 	github.com/lightninglabs/lightning-node-connect/hashmailrpc v1.0.2
-	github.com/lightninglabs/lndclient v0.18.4-5
+	github.com/lightninglabs/lndclient v0.18.4-7
 	github.com/lightninglabs/neutrino/cache v1.1.2
-	github.com/lightningnetwork/lnd v0.18.3-beta.rc3.0.20241120143113-9246d5c51cd2
+	github.com/lightningnetwork/lnd v0.18.4-beta.rc1
```
Roasbeef, line 31: 👍

Roasbeef (approved):
LGTM 🌹

itest/send_test.go
```diff
@@ -1954,6 +1954,92 @@ func testSendNoCourierUniverseImport(t *harnessTest) {
...
+	// We create a new lnd node from a seed, so we can restore it with the
+	// same seed later.
+	password := []byte("somepassword")
+	seedLnd, mnemonic, _ := t.lndHarness.NewNodeWithSeed(
+		"seed-lnd", lndDefaultArgs, password, false,
+	)
+	t.lndHarness.FundCoins(btcutil.SatoshiPerBitcoin, seedLnd)
+
+	bob := setupTapdHarness(t.t, t, seedLnd, t.universeServer)
+	defer func() {
```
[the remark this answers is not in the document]
guggero: Ah, yes. Makes sense. Changed.

PR #1214, opened 2024-11-22
Add PSBT field to UnsealedAsset RPC message

ffranr:
The unsealed asset RPC message already includes a `group_virtual_tx` field. This PR enhances the message by adding a new field that contains the byte-serialized PSBT equivalent. This addition is intended to assist external signers in producing a group witness.

For an example of the external signer workflow, refer to `testMintFundSealAssets`.

This PR addresses the request in [#1207](https://github.com/lightninglabs/taproot-assets/issues/1207).


rpcserver.go
```diff
@@ -4218,12 +4219,32 @@ func marshalUnsealedSeedling(verbose bool,
 		if err != nil {
 			return nil, err
 		}
+
+		// Generate PSBT equivalent of the group virtual tx.
+		groupVirtualPsbt, err := psbt.NewFromUnsignedTx(
```
guggero, line 4224: This doesn't really give us anything... This can trivially also be `bitcoin-cli converttopsbt <tx>`.
If we want to give a signer any chance of being able to sign this, we need to add the minimum amount of metadata we have available:

```
		groupVirtualPsbt.Inputs[0].WitnessUtxo = &wire.TxOut{
			Value:    0, // Probably needs to be set to the equal amount as the VM TX's output value?
			PkScript: nil, // Needs to be the p2tr script for the script key Taproot key.
		}
		groupVirtualPsbt.Inputs[0].TaprootInternalKey = ? // I think we can set this from the script key's internal key, which I assume we need to know at this point?
		groupVirtualPsbt.Inputs[0].TaprootMerkleRoot = ? // Same as above, should be the tweak of the script key.

		groupVirtualPsbt.Outputs[0].TaprootInternalKey = ? // Same as above.
```

guggero (commented):
Thanks for the PR. Though I'm not sure this alone gives us anything in terms of being able to _sign_ for such a key.
PSBTs are all about the meta information contained to tell an offline device _what_ it's signing and _what key_ it has to sign with.
We have some of that information available (see inline comment).
But other stuff we just don't know. Like the derivation path for the signing key.

And because you can't easily edit a PSBT (or at least I haven't found a good editor that doesn't suck), it's not trivial to add that information if you're just a user and not in a Golang code environment.

IMO we have a couple of options here:
 - Go down the path we talked about offline: force the user to provide a **template** PSBT (generated using `lnd` for example, see gist linked above) that contains the derivation path. The template would be provided in a previous step and then merged into the actual group VM TX. --> Maybe a bit too clunky for now?
 - Force the group key to be known to `lnd`'s wallet, for example by importing the xpub as a watch-only wallet. That would tie things very much into `lnd`, but it would also simplify some things (for example you wouldn't need to do a dummy on-chain TX as shown in the Gist, just to be able to create a template, we could instead just query the derivation path of the key through an `RPC` when the PSBT group key option is requested). --> Perhaps the easiest solution to go for right now, wouldn't make other solutions impossible in the future.
 - Force the user to provide all that meta information through the RPC (kind of defeats the purpose of the PSBT, as that's the "transportation medium" for such meta info.

Whatever route we go with, we definitely MUST accompany it with a document that runs you through the process step-by-step.
Feel free to use whatever you seem useful from this gist mentioned above: https://gist.github.com/guggero/569241aa9fec57e287101187bd28d1c5

Issue #1216, opened 2024-11-26
[bug]: SendPayment fails with EOF error after successful payment


ZZiigguurraatt:
I am trying to make a payment using the SendPayment (https://lightning.engineering/api-docs/api/taproot-assets/taproot-asset-channels/send-payment/) gRPC in python where alice pays bob over a direct taproot asset channel (no routing or exchanges to sats). The payment succeeds, but after that I get an EOF error from the gRPC. I'm getting this error when using litd with commit d82cb9d275ad706207eebaa86ad68edb5ef055d3 as well as tag v0.13.993-exp . Note: this is not a problem with my python gRPC because no other calls have this error and LND's SendPaymentV2 (https://lightning.engineering/api-docs/api/lnd/router/send-payment-v2/) works fine.

```
payment_result {
  payment_hash: "62afeac0474399f2ca8687c2deab04b5b4aeaf7f3514ab9896ec4cb39f18e8fb"
  value: 500
  creation_date: 1732645337
  payment_preimage: "765dde18d7518b67ebd7f3f2f25ae21bb12ef60bcd3438a62dd909bd07486f50"
  value_sat: 500
  value_msat: 500000
  payment_request: "lnbcrt5u1pn5vywepp5v2h74sz8gwvl9j5xslpda2cykk62atmlx522hxyka3xt88ccarasdqqcqzzsxqyz5vqrzjqwedrfsykk87jmxmy4hf93usnc3ywawqst7uclt3hz602z2965ma0aa8ecy32drt8qqqqqlgqqqqqqgq2qsp5du78r6hcay0ayklukp40d2mjyagmkxchrr40ce4vw3zktrz2ch5q9qxpqysgqx3zhe5jvssletg9a596zw06q5p687lwdzhgnddkjtsucyw8zrzdrz2ptgatj6ds0u6jqya35a4gmu65zq60rwkrzrvpp94w33ajhqycqstx4qd"
  status: SUCCEEDED
  creation_time_ns: 1732645337576337406
  htlcs {
    status: SUCCEEDED
    route {
      total_time_lock: 14923
      total_amt: 500
      hops {
        chan_id: 16251881370222592
        chan_capacity: 100000
        amt_to_forward: 500
        expiry: 14923
        amt_to_forward_msat: 500000
        pub_key: "03295e89e0eb6f166f9d285b4d00409f83c3c266a5e6ceb7e20ad26fb719d65aaa"
        tlv_payload: true
        mpp_record {
[...]

PR #1217, opened 2024-11-26
rpc: try to catch EOF error from lnd sendpayment

Roasbeef:
A user observed that when using Python, they get an EOF even if the payment succeeds. My guess is that `lnd` has hung up, then we try to recv and get an EOF error. With this PR, we try to catch that EOF error and just return `nil`. 

One other idea: do we need to call `CloseSend` at the very end or in a defer?

Fixes: https://github.com/lightninglabs/taproot-assets/issues/1216

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -7193,7 +7193,12 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
 
 		update, err := updateStream.Recv()
 		if err != nil {
-			return err
+			// Stream is closed; no more updates.
+			if err == io.EOF {
+				return nil
+			}
+			return fmt.Errorf("failed to receive payment "+
+				"update: %w", err)
 		}
 
 		err = stream.Send(&tchrpc.SendPaymentResponse{


ZZiigguurraatt:
Not sure if I'm doing something wrong, but I'm getting this error.

```
unknown flag `taproot-assets.experimental.rfq.priceoracleaddress'
could not load config: error parsing flags: unknown flag `taproot-assets.experimental.rfq.priceoracleaddress'
```

ZZiigguurraatt:
Has `taproot-assets.experimental.rfq.priceoracleaddress` been depreciated for something else?

guggero:
Are you running `tapd` directly? Then you need to remove the `taproot-assets` prefix as that's only valid in `litd`.

ZZiigguurraatt:
I'm running litd. I didn't think you could use sendpayment with tapd directly.

guggero:
> you could use sendpayment with tapd directly

Yeah, you can't. Just trying to find out what's wrong here... Is it possible you have an old `litd` somewhere in your `$PATH` that's being picked up?
The flag still exists on the `update-to-lnd-18-4` branch (but it probably doesn't on `master`):

```
litd --help | grep oracle
      --taproot-assets.experimental.rfq.priceoracleaddress=                    Price oracle gRPC server address (rfqrpc://<hostname>:<port>). To use the
                                                                               use_mock_price_oracle_service_promise_to_not_use_on_mainnet
```

ZZiigguurraatt:
> > you could use sendpayment with tapd directly
> 
> Yeah, you can't. Just trying to find out what's wrong here... Is it possible you have an old `litd` somewhere in your `$PATH` that's being picked up? The flag still exists on the `update-to-lnd-18-4` branch (but it probably doesn't on `master`):
> 
> ```
> litd --help | grep oracle
>       --taproot-assets.experimental.rfq.priceoracleaddress=                    Price oracle gRPC server address (rfqrpc://<hostname>:<port>). To use the
>                                                                                use_mock_price_oracle_service_promise_to_not_use_on_mainnet
> ```

I was running inside docker, so I don't think that was the issue.

Roasbeef:
Updated the PR body with more context.

ZZiigguurraatt:
Sorry, I accidentally was not using `dev.Dockerfile`, so it was just pulling from master instead of using my locally checked out branch.

I can confirm this fix works!

rpcserver.go
```diff
@@ -7193,7 +7193,12 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
 
 		update, err := updateStream.Recv()
 		if err != nil {
-			return err
+			// Stream is closed; no more updates.
+			if err == io.EOF {
```
guggero, line 7197: nit: Use `errors.Is()` instead? Though it shouldn't really be wrapped at this point, so doesn't matter too much.

guggero (approved):
Thanks for the fix! LGTM :tada:

Issue #1219, opened 2024-11-27
[bug]: REST FundChannel Method Incorrectly Implemented as GET Instead of POST and Does Not Work


V0nMis3s:
### Background

The `FundChannel` method in the REST API is currently implemented as a `GET` method, but this is semantically incorrect as it performs a state-changing operation (funding a channel). Moreover, the `GET` implementation fails to function even when used as intended, returning a malformed public key error.

This issue renders the method unusable and violates RESTful principles. The method should be implemented as a `POST` and accept the `FundChannelRequest` object in the body of the request.

### Your environment

```sh
tapcli getinfo
# tapd version: 0.4.1-alpha commit=v0.4.2-0.20240815180811-2110839696cb
# lnd version: 0.18.0-beta
# network: regtest

uname -mrsv
# inux 6.8.0-49-generic #49-Ubuntu SMP x86_64

bitcoind --version
# Bitcoin Core version v27.0

```

### Steps to reproduce

1. Start in [integrated mode](https://docs.lightning.engineering/lightning-network-tools/lightning-terminal/integrating-litd), and ensure it is running.
2. Attempt to fund a channel using a GET request with query parameters:
```sh
curl -k -X GET \
  https://{host}:{port}/v1/taproot-assets/channels/fund \
  -H "Grpc-Metadata-macaroon: $(xxd -ps -c 10000 /path/to/admin.macaroon)" \
  -H "Content-Type: application/json" \
  --data-urlencode "asset_amount=100" \
  --data-urlencode "peer_pubkey=base64_pubkey" \
  --data-urlencode "asset_id=base64_asset_id="
```
3. Observe the response:
```json
{
  "code": 2,
  "message": "error parsing peer pubkey: malformed public key: invalid length: 0",
  "details": []
}
```
4. Attempt the same operation as a POST request with the request body:
 ```sh
curl -k -X POST \
  https://{host}:{port}/v1/taproot-assets/channels/fund \
  -H "Grpc-Metadata-macaroon: $(xxd -ps -c 10000 /path/to/admin.macaroon)" \
  -H "Content-Type: application/json" \
  -d '{
        "asset_amount": "100",
        "peer_pubkey": "base64_pubkey",
        "asset_id": "base64_asset_id="
      }'
```
5.Observe the response:
```json
{
  "code": 12,
  "message": "Method Not Allowed",
  "details": []
}
``` 

### Expected behavior

The `FundChannel` method should:
[...]

PR #1220, opened 2024-11-27
tchrpc: change FundChannel endpoint to POST

GeorgeTsagk:
Closes https://github.com/lightninglabs/taproot-assets/issues/1219

### DIFF
--- a/taprpc/tapchannelrpc/tapchannel.pb.gw.go
+++ b/taprpc/tapchannelrpc/tapchannel.pb.gw.go
@@ -31,18 +31,15 @@ var _ = runtime.String
 var _ = utilities.NewDoubleArray
 var _ = metadata.Join
 
-var (
-	filter_TaprootAssetChannels_FundChannel_0 = &utilities.DoubleArray{Encoding: map[string]int{}, Base: []int(nil), Check: []int(nil)}
-)
-
 func request_TaprootAssetChannels_FundChannel_0(ctx context.Context, marshaler runtime.Marshaler, client TaprootAssetChannelsClient, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {
 	var protoReq FundChannelRequest
 	var metadata runtime.ServerMetadata
 
-	if err := req.ParseForm(); err != nil {
-		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)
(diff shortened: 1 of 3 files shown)


V0nMis3s:
Hi @GeorgeTsagk, thank you for addressing the issue so quickly! However, I noticed that the changes in this PR directly modify the `taprpc/tapchannelrpc/tapchannel.pb.gw.go` file, which, as indicated in the first line (`Code generated by protoc-gen-grpc-gateway. DO NOT EDIT.`), is an auto-generated file. 
Direct edits to this file are not recommended because:
1. These changes will be overwritten the next time the file is regenerated using `protoc`.
2. Maintaining changes directly in generated files makes it harder to track and ensure consistency across versions.

### Proposed Structural Solution
Instead of directly modifying the `.pb.gw.go` file, the appropriate place to make these changes is in the `.proto` file (`taprpc/tapchannelrpc/tapchannel.proto`). Specifically:
1. Add the explicit method annotation to the `FundChannel` RPC method.
2. Regenerate the gateway code by running `protoc` with the necessary flags, which will update the `.pb.gw.go` file automatically.

This approach ensures that:
- The `FundChannel` method is correctly mapped to a `POST` in the REST API.
- The generated files remain consistent with the source `.proto` definitions.
- The changes are maintainable and won't be accidentally overwritten.

Please let me know what you think!

GeorgeTsagk:
Hey @V0nMis3s, thanks for taking a look at the PR

> Instead of directly modifying the .pb.gw.go file, the appropriate place to make these changes is in the .proto file (taprpc/tapchannelrpc/tapchannel.proto). Specifically:

This was actually automated and was generated by running `make rpc` (check the [Makefile](https://github.com/lightninglabs/taproot-assets/blob/main/Makefile#L253-L255) section). The only change that had to be made was in the [`tapchannel.yaml`](https://github.com/lightninglabs/taproot-assets/blob/main/taprpc/tapchannelrpc/tapchannel.yaml) file, which we use to map the rpc methods to rest endpoints

ffranr (approved):

V0nMis3s:
Thank you for the explanation! I assumed the mapping was automatically handled by the default rules of grpc-gateway, and I didn’t realize the tapchannel.yaml file was being used to map the RPC methods to REST endpoints. Thanks for pointing that out!

gijswijs (approved):
LGTM :tada:

Issue #1215, opened 2024-11-23
[feature]: Allow specifying a custom RFQ scid on `tchrpc.SendPayment`


GeorgeTsagk:
## Description

Currently the [`tchrpc.SendPayment`](https://github.com/lightninglabs/taproot-assets/blob/main/taprpc/tapchannelrpc/tapchannel.proto#L34) RPC allows specifying the peer pubkey to be used for the rfq negotiation. This is the only way to influence the rfq operations that will take place in order to complete the payment.

## Problem

A super-user may not want to rely on the internal mechanism of our RPC for the RFQ negotiation. If a quote was somehow negotiated via an external system they will want to somehow guide `tchrpc.SendPayment` into using it, instead of generating a new one.

## Proposed solution

Add a new argument to the `tchrpcSendPaymentRequest` which is the scid of an existing quote. The method will then blindly try to use that rfq scid to complete the payment, skipping all the rfq negotiation parts.

PR #1224, opened 2024-11-28
Support user-generated RFQ on `tchrpc.SendPayment`

GeorgeTsagk:
Closes https://github.com/lightninglabs/taproot-assets/issues/1215


taprpc/tapchannelrpc/tapchannel.proto
```diff
@@ -121,6 +121,11 @@ message SendPaymentRequest {
     // contain a valid keysend record (key 5482373484 and a 32-byte preimage
     // that corresponds to the payment hash).
     routerrpc.SendPaymentRequest payment_request = 4;
+
+    // The rfq scid to use for this payment. If the user sets this value then
+    // the payment will immediately be dispatched, skipping the rfq negotiation
+    // phase, and using the following rfq scid.
+    uint64 scid = 5;
```
ffranr, line 128: Should this skip RFQ completely or just override scid? Where does the asset rate come from?

guggero: I think we'll instead want the full 32-byte RFQ ID to be passed in. Then we can validate it exists and hasn't expired yet. If all is okay, we can easily derive the SCID from it.

rpcserver.go
```diff
@@ -7034,6 +7034,27 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+	case req.Scid != 0:
+		var quote rfqmsg.SellAccept
+		for id, q := range r.cfg.RfqManager.PeerAcceptedSellQuotes() {
+			if id == rfqmsg.SerialisedScid(req.Scid) {
+				quote = q
+				break
+			}
+		}
+
+		htlc := rfqmsg.NewHtlc(nil, fn.Some(quote.ID))
```
guggero, line 7046: We should probably do this part as well to remain RPC compatible:

```
		// Send out the information about the quote on the stream.
		err = stream.Send(&tchrpc.SendPaymentResponse{
			Result: &tchrpc.SendPaymentResponse_AcceptedSellOrder{
				AcceptedSellOrder: quote,
			},
		})
		if err != nil {
			return err
		}

		// Unmarshall the accepted quote's asset rate.
		assetRate, err := rfqrpc.UnmarshalFixedPoint(
			acceptedQuote.BidAssetRate,
		)
		if err != nil {
			return fmt.Errorf("error unmarshalling asset rate: %w",
				err)
		}

		// Calculate the equivalent asset units for the given invoice
		// amount based on the asset-to-BTC conversion rate.
		numAssetUnits := rfqmath.MilliSatoshiToUnits(
			*invoice.MilliSat, *assetRate,
		)

		rpcsLog.Infof("Using quote for %v asset units at %v asset/BTC "+
			"from peer %x with SCID %d", numAssetUnits, assetRate,
			peerPubKey, acceptedQuote.Scid)

		var rfqID rfqmsg.ID
		copy(rfqID[:], acceptedQuote.Id)
```

guggero (commented):
Very nice! Simple enough approach. Have two suggestions though, otherwise looks good, thank you :tada:

taprpc/tapchannelrpc/tapchannel.proto
GeorgeTsagk: I don't really see the need to pass the full `[32]byte` rfq id?

We'll still have to perform the lookup with `r.cfg.RfqManager.PeerAcceptedSellQuotes()` which uses the scid as the key

Also, if an entry exists in that list it means that:
- Quote exists
- Quote hasn't expired

ffranr: I would argue that the RFQ ID provides a more general and accurate way to reference an RFQ quote. The SCID is downstream of the RFQ quote and serves as a specific property of it. Using the RFQ ID in the interface ensures clarity and reinforces that we are referring to the entire quote (including the asset rate) rather than just the SCID.

rpcserver.go
```diff
@@ -7034,6 +7034,27 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
 		// Continue below.
 
+	case req.Scid != 0:
+		var quote rfqmsg.BuyAccept
+		for id, q := range r.cfg.RfqManager.PeerAcceptedBuyQuotes() {
+			if id == rfqmsg.SerialisedScid(req.Scid) {
+				quote = q
+				break
+			}
+		}
```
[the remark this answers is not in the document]
ffranr: I see that the latest commits return an error :+1:

ffranr (commented):
I think Oli has a good point and that we should pass in the RFQ quote ID and not the SCID. I think that would be clearer, see comment thread.

taprpc/tapchannelrpc/tapchannel.proto
```diff
@@ -122,10 +122,10 @@ message SendPaymentRequest {
...
     routerrpc.SendPaymentRequest payment_request = 4;
 
-    // The rfq scid to use for this payment. If the user sets this value then
-    // the payment will immediately be dispatched, skipping the rfq negotiation
-    // phase, and using the following rfq scid.
-    uint64 scid = 5;
+    // The rfq id to use for this payment. If the user sets this value then the
+    // payment will immediately be dispatched, skipping the rfq negotiation
+    // phase, and using the following rfq id instead.
+    bytes rfq_id = 5;
```
guggero, line 128: This belongs into the first commit.

rpcserver.go
```diff
@@ -7034,6 +7034,92 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+		if len(req.RfqId) != 32 {
+			return fmt.Errorf("rfq id length must be 32")
+		}
+
+		// Now let's try to perform an internal lookup to see if there's
+		// an actual quote on this ID.
+		var rfqID rfqmsg.ID
+		copy(rfqID[:], req.RfqId)
+
+		var quote rfqmsg.SellAccept
```
guggero, line 7048: nit: can have this as a pointer, then below we can do a `nil` check instead of comparing the ID against an empty ID.

rpcserver.go
```diff
@@ -7034,6 +7034,92 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+		for id, q := range r.cfg.RfqManager.PeerAcceptedSellQuotes() {
+			if id == rfqID.Scid() {
```
guggero, line 7050: Why not just `rfqID == q.ID`?

rpcserver.go
```diff
@@ -7034,6 +7034,92 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+		}
+
+		// This quote ID's scid did not match anything.
+		if quote.ID == [32]byte{} {
+			return fmt.Errorf("scid did not match an accepted " +
+				"quote")
+		}
+
+		// This quote ID's scid matched a different quote.
+		if !bytes.Equal(quote.ID[:], req.RfqId) {
```
guggero, line 7063: This would not be needed if we did the comparison on the RFQ ID in the first place.

guggero (changes requested):

rpcserver.go
GeorgeTsagk: Right, in that case I can get rid of the check below too

GeorgeTsagk:
not sure why litd is failing, kicked it

locally the itests are passing

guggero (approved):
LGTM :tada:

rpcserver.go
```diff
@@ -7034,6 +7034,86 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
 	case len(firstHopRecords) > 0:
 		// Continue below.
 
+	case req.RfqId != nil:
+		// Check if the provided rfq ID matches the expected length.
+		if len(req.RfqId) != 32 {
+			return fmt.Errorf("rfq id length must be 32")
```
ffranr, line 7040: `must be 32 bytes in length`

rpcserver.go
```diff
@@ -7034,6 +7034,86 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+				quote = &qCopy
+				break
+			}
+		}
+
+		// This quote ID's scid did not match anything.
+		if quote == nil {
+			return fmt.Errorf("scid did not match an accepted " +
+				"quote")
+		}
```
ffranr, line 7061: we're matching on quote IDs now. So comment and error needs update.

rpcserver.go
```diff
@@ -7034,6 +7034,86 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+			Result: &tchrpc.SendPaymentResponse_AcceptedSellOrder{
+				AcceptedSellOrder: sellOrder,
+			},
+		})
+		if err != nil {
+			return err
+		}
+
+		rpcsLog.Infof("Using quote for %v asset units at %v asset/BTC "+
+			"from peer %x with SCID %d", numAssetUnits, rate,
```
ffranr, line 7102: We should use `rate.String()` here. (I don't think that will happen automatically here with %v)

rpcserver.go
```diff
@@ -7034,6 +7034,86 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+		}
+
+		// Send out the information about the quote on the stream.
+		err = stream.Send(&tchrpc.SendPaymentResponse{
+			Result: &tchrpc.SendPaymentResponse_AcceptedSellOrder{
+				AcceptedSellOrder: sellOrder,
+			},
+		})
+		if err != nil {
+			return err
```
ffranr, line 7098: consider wrapping this error with a `fmt.Errorf` description to give context.

ffranr (approved):

PR #1225, opened 2024-11-29
multi: roll back change in request page size

guggero:
This fixes an issue with a new tapd (v0.5.0-rc1) attempting to sync with an old (v0.4.1) universe server. Because we now serve all the root nodes from a cache, we allow a much larger page size when requesting the root nodes. But old universe servers will not yet allow that higher request limit, resulting in the following error:

[/universerpc.Universe/AssetLeafKeys]: invalid request limit

So we need to use the old limit when _requesting_ pages. And once most of the universe servers have been updated (e.g. v0.5.1), we can then also use the larger limit on requests.


GeorgeTsagk (approved):
lgtm

Roasbeef (approved):
LGTM 🚁

PR #1229, opened 2024-12-02
scripts/keys: add roasbeef's new key

Roasbeef:
In this commit we add roasbeef's new GPG signing key.


guggero (approved):
Can confirm this is the same key as [in `lnd`](https://github.com/lightningnetwork/lnd/blob/master/scripts/keys/roasbeef.asc).

GeorgeTsagk (approved):

Issue #1227, opened 2024-12-02
[bug]: Terminal crashes


lukegao209:
### Background
Version ： lightninglabs/lightning-terminal:v0.14.0-alpha.rc1-path-prefix

When the RFQ accesses the Price Oracle, if it returns an error indicating that the asset is not found, the terminal crashes.

ffranr:
This should probably be moved to the terminal repo.

@lukegao209 do you have any logs just in case tapd is the cause of the crash?

lukegao209:
You can add our Price Oracle server and call addinvoice with an assetId that does not exist on the Price Oracle. This will cause the terminal to crash.

--taproot-assets.experimental.rfq.priceoracleaddress=rfqrpc://oracle.ln.exchange:443

guggero:
Thanks. But if you have access to the stack trace of the crash, that would help us to fix the bug even quicker.

lukegao209:
> Thanks. But if you have access to the stack trace of the crash, that would help us to fix the bug even quicker.
here is the crash logs : 
2024-12-03 01:03:01.202 [INF] LITD: Handling gRPC request: /litrpc.Proxy/BakeSuperMacaroon
2024-12-03 01:03:01.216 [INF] LITD: Handling gRPC request: /tapchannelrpc.TaprootAssetChannels/AddInvoice
2024-12-03 01:03:01.790 [WRN] RFQS: failed to query bid price from price oracle for outgoing buy request: failed to query price oracle for bid price: OracleError(code=0, msg=unsupported subject asset)
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1672b80]

goroutine 26389 [running]:
github.com/lightninglabs/taproot-assets/rfq.(*Negotiator).HandleOutgoingBuyOrder.func1()
        github.com/lightninglabs/taproot-assets@v0.5.0-rc1/rfq/negotiator.go:197 +0x370
created by github.com/lightninglabs/taproot-assets/rfq.(*Negotiator).HandleOutgoingBuyOrder in goroutine 26378
        github.com/lightninglabs/taproot-assets@v0.5.0-rc1/rfq/negotiator.go:158 +0xd8

guggero:
Great, thank you! Will create a fix today.

[...]

PR #1232, opened 2024-12-03
rfq: fix potential nil pointer panic

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1227.

Because we don't return if the error returned from the price oracle is non-nil, the asset rate might be nil in case we couldn't talk to the oracle. So we need to use fn.MaybeSome() instead of fn.Some() with a dereference.


Roasbeef (approved):
LGTM 🎉

PR #1233, opened 2024-12-03
Add AltLeaf support to tapfreighter

jharveyb:
This continues the work from #1180 .

Specifically:

- Adds the `AltLeaves` field to `Proof`.
- Tests correct en/decode of a `Proof` with `AltLeaves` populated.
- Propagates `AltLeaves` from `vOutputs`, to anchor outputs for a transfer, and the corresponding transition proof.
- Fetches `AltLeaves` for asset inputs during coin selection, and uses these `AltLeaves` when verifying the input anchor pubkey.

Test-wise, the added unit tests cover the new validation methods around `[]AltLeaves`, and the new funcs that act on Tap commitments. The changes in later commits in `tapsend` and `tapfrieghter` are mostly only exercised via the PSBT itest.


jharveyb:
From some offline review, there are a few clear TODOs:

- Convert the new funcs that operate on Tap commitments to methods
- Convert new errors to named errors
- Add some more altleaf generation in existing unit tests
- Add new asserts for itests wrt. checking for alt leaves

jharveyb:
The test vectors for `Proof` also need to be updated.

commitment/tap.go
```diff
@@ -662,3 +662,114 @@ func TrimSplitWitnesses(version *TapCommitmentVersion,
...
+		if newLeafKeys.Contains(leafKey) {
+			return fmt.Errorf("%w: existing alt leaf: %x",
+				asset.ErrDuplicateAltLeafKey, leafKey)
+		}
+	}
+
+	// None of the new or existing alt leaves collide; we can now update
+	// the AltCommitment and Tap commitment.
+	if currentAltCommit == nil {
+		currentAltCommit, err = NewAssetCommitment(
```
guggero, line 754: move this into the `else` case above?

tapfreighter/wallet.go
```diff
@@ -910,6 +910,16 @@ func createAndSetInput(vPkt *tappsbt.VPacket, idx int,
 	}
 	vPkt.SetInputAsset(idx, assetInput.Asset)
 
+	inputAltLeaves, err := assetInput.Commitment.FetchAltLeaves()
+	if err != nil {
+		return fmt.Errorf("cannot fetch alt leaves from input: %w", err)
+	}
+
+	err = vPkt.Inputs[idx].SetAltLeaves(inputAltLeaves)
```
guggero, line 930: Hmm, I thought we wouldn't start producing alt leaves just yet, only in the next version? Or I guess since the alt leaves would only be in the root output's commitment, a non-interactive receiver wouldn't come into contact with them yet.

rpcserver.go
```diff
@@ -1837,6 +1851,7 @@ func (r *rpcServer) marshalProof(ctx context.Context, p *proof.Proof,
 		IsBurn:              p.Asset.IsBurn(),
 		GenesisReveal:       genesisReveal,
 		GroupKeyReveal:      &GroupKeyReveal,
+		AltLeaves:           altLeavesBuf.Bytes(),
```
guggero, line 1854: Not sure if we should instead encode this just as a list of `taprpc.Asset` instead?

rpcserver.go
```diff
@@ -1758,6 +1759,19 @@ func (r *rpcServer) marshalProof(ctx context.Context, p *proof.Proof,
 		}
 	}
 
+	var altLeavesBuf bytes.Buffer
+	if altLeaves != nil {
```
guggero, line 1763: Check for length instead?

guggero (commented):
Very clean! Did a first pass but will want to take another one to take a closer look at backward compatibility, with some manual testing.

tapfreighter/wallet.go
Roasbeef: So IIUC, we aren't quite _producing_ them yet, but we always make sure that if they're present, they're passed along/validated.

Roasbeef: We'll also need to do this (`SetAltLeaves`) once we're constructing a `vOut` when prepping to send.

rpcserver.go
Roasbeef: Hmm good question...I think that would be a bit more useful, as then we can assert based on that in the future itests once everything is strung up.

Roasbeef (changes requested):
Nice work with this change! Lots of little nooks and crannies to identify+update. Will sit with it for a bit to try and identify any missing areas or deficiencies in the tests

rpcserver.go
jharveyb: I thought about that too and decided against it, since most fields would be at some default value. And the user should never handle them as assets - this is mostly added for 3rd party tools that would need access to the leaves.

proof/append.go
```diff
@@ -135,6 +135,19 @@ func CreateTransitionProof(prevOut wire.OutPoint,
...
 
+	// Copy any AltLeaves from the anchor commitment to the proof.
+	altLeaves, err := params.TaprootAssetRoot.FetchAltLeaves()
+	if err != nil {
+		return nil, err
+	}
+
+	if len(altLeaves) > 0 {
+		proof.AltLeaves = make([]asset.AltLeafAsset, len(altLeaves))
+		for idx, altLeaf := range altLeaves {
```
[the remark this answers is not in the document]
jharveyb: I think that doesn't work because we're 'casting' into `asset.AltLeafAsset` here, which a copy wouldn't do.

commitment/tap.go
```diff
@@ -662,3 +662,114 @@ func TrimSplitWitnesses(version *TapCommitmentVersion,
...
+// MergeAltLeaves adds a set of AltLeaves to an existing TapCommitment. Merging
+// fails if the new AltLeaves collide with any existing AltLeaves.
+func (c *TapCommitment) MergeAltLeaves(altLeaves []asset.AltLeafAsset) error {
+	if len(altLeaves) == 0 {
+		return nil
+	}
+
+	// First, check that the given alt leaves have unique
+	// AssetCommitmentKeys.
+	newLeafKeys := fn.NewSet[[32]byte]()
```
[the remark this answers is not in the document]
jharveyb: Good point - one may have come before the other when working on this, not certain.

jharveyb:
Added a new itest to check that the tapfreighter and the vPSBT RPCs are rejecting invalid combinations of `AltLeaves` correctly (which surfaced some small need tweaking). The new test also demonstates that `AltLeaves` are merged into the same anchor output correctly.

gijswijs (commented):
Why did we choose to trim alt leaves from passive vPkts? I tried to think of a scenario where we would want those alt leaves to persist. I couldn't really think of one, but I could see a future where we would find a use case for that.

proof/proof_test.go
```diff
@@ -283,6 +283,13 @@ func genRandomGenesisWithProof(t testing.TB, assetType asset.Type,
 		},
 	)
 	require.NoError(t, err)
+
+	// Include 1 or more alt leaves in the anchor output Tap commitment.
```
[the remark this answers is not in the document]
guggero: You're not wrong. But this specific part of the code is also used for the test vectors, so we definitely want to include some alt leaves there. Updated the comment to make that more clear.

tappsbt/interface.go
```diff
@@ -660,6 +692,35 @@ func (o *VOutput) SetAnchorInternalKey(keyDesc keychain.KeyDescriptor,
 	)
 }
 
+// SetAltLeaves asserts that a set of AltLeaves are valid, and updates a VOutput
+// to set the AltLeaves. Setting the output's AltLeaves twice is disallowed.
+func (o *VOutput) SetAltLeaves(altLeaves []*asset.Asset) error {
+	// AltLeaves can be set exactly once on a VOutput.
```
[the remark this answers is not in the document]
guggero: I agree, this was quite a bit of duplicated code. But I was able to trim it down quite a bit, so it's not as bad anymore. IMO it's not enough to resort to dark pointer magic (`*target = newLeaves`). Hope you agree...

commitment/tap.go
```diff
@@ -662,3 +662,110 @@ func TrimSplitWitnesses(version *TapCommitmentVersion,
...
+						"leaf: %x",
+						asset.ErrDuplicateAltLeafKey,
+						leafKey)
+				}
+			}
+
+		}
+	}
+
+	// None of the new or existing alt leaves collide; we can now update
```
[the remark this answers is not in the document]
guggero: Correct. Each `VInput` can only reference a single input. And each `VOutput` can also only have its alt leaves be set once. But in case there are multiple virtual packets (e.g. multiple asset transfers are committed to the same on-chain anchor transaction) the alt leaves from multiple `VOutput`s might need to be merged into a single commitment (if two or more `VOutput`s reference the same on-chain output index).

guggero:
> Why did we choose to trim alt leaves from passive vPkts?

We don't trim the alt leaves from passive assets. The two places where we use the `commitment.TrimAltLeaves()` function is where we check if any assets remain after removing a certain asset from a commitment (e.g. the asset being burned or the active asset of a transfer). And for those checks the alt leaves would be counted as "normal" assets and give a wrong result. So just for those two checks we trim the alt leaves (but in the case of passive assets we add them back further below).

guggero (approved):
As far as I can tell this gives us everything we need to be forward compatible with alt leaves, so giving my LGTM.

tappsbt/interface.go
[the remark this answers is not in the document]
gijswijs: Steer clear of Rust then. :smile:

gijswijs:
LGTM :tada:

gijswijs (approved):

PR #1235, opened 2024-12-04
Refactor GroupKeyReveal for multi-version support

ffranr:
Rename the existing `GroupKeyReveal` struct to `GroupKeyRevealV0` and introduce a new `GroupKeyReveal` interface. This change enables adding new versions of `GroupKeyReveal` while retaining the V0 type for proper documentation of its behavior.

Support for `GroupKeyReveal` V0 will be preserved.

### DIFF
--- a/proof/proof.go
+++ b/proof/proof.go
@@ -301,11 +301,13 @@ type Proof struct {
 	// re-derivation of the asset group key.
 	GenesisReveal *asset.Genesis
 
-	// GroupKeyReveal is an optional set of bytes that represent the public
-	// key and Tapscript root used to derive the final tweaked group key for
-	// the asset group. This field must be provided for issuance proofs of
-	// grouped assets.
-	GroupKeyReveal *asset.GroupKeyReveal
+	// GroupKeyReveal contains the data required to derive the final tweaked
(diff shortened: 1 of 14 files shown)


GeorgeTsagk (approved):
Clean :sparkles: 
LGTM :heavy_check_mark:

proof/proof.go
```diff
@@ -301,11 +301,11 @@ type Proof struct {
...
 
-	// GroupKeyReveal is an optional set of bytes that represent the public
-	// key and Tapscript root used to derive the final tweaked group key for
-	// the asset group. This field must be provided for issuance proofs of
-	// grouped assets.
-	GroupKeyReveal *asset.GroupKeyReveal
+	// GroupKeyReveal contains the data required to derive the final tweaked
+	// group key for an asset group.
+	//
+	// NOTE: This field is mandatory for issuance proofs of grouped assets.
```
guggero, line 307: IIRC this only needs to be set for the group anchor. Then each following tranche just needs to have a valid signature for the previously revealed group key. Just in case we want to add some extra precision to this sentence.

guggero (approved):
Nice and spotless rename/refactor :100:

Issue #1234, opened 2024-12-03
[bug]: asset_amount not showing up in SendPayment response


ZZiigguurraatt:
When I use `SendPayment` (https://lightning.engineering/api-docs/api/taproot-assets/taproot-asset-channels/send-payment/) I get a `SendPaymentResponse` object that includes an `accepted_sell_order` field of type `PeerAcceptedSellQuote`. When I look at the definition of `PeerAcceptedSellQuote` (https://github.com/lightninglabs/taproot-assets/blob/f35de11be21295685f8fb56d0e0d394c48b3bd4a/taprpc/rfqrpc/rfq.proto#L265-L266), it includes an `asset_amount` field, but I am not getting that in my response. Instead, I'm just getting

```
    accepted_sell_order {
      peer: "037ef030abc42123d69ed904897262267a6c2c06d91cf2397d7bed193763fc16fe"
      id: "f\3571\037Vg I\224\275\014\226\360N\332}\253\200<7y\007j/\366\250\240\246\036G\327\276"
      scid: 17773632564750702526
      bid_asset_rate {
        coefficient: "1000000"
      }
      expiry: 1733261726
    }
```

Looking at the log file, I see

```
    2024-12-03 20:35:26.456 [INF] RPCS: Got quote for 10 asset units at 1000000 asset/BTC from peer 30333765663033306162633432313233643639656439303438393732363232
    3637613663326330366439316366323339376437626564313933373633666331366665 with SCID 17604825916137069597
```

PR #1236, opened 2024-12-04
[custom channels]: Fix bandwidth manager and other bugs

guggero:
Depends on https://github.com/lightningnetwork/lnd/pull/9333.

Fixes the issue that would lead to a force close if the channel bandwidth of a target channel during a forward event isn't sufficient.

Also fixes https://github.com/lightninglabs/taproot-assets/issues/1234 (drive-by fix) and another edge case found during integration tests.

### DIFF
--- a/tapchannel/commitment.go
+++ b/tapchannel/commitment.go
@@ -363,9 +363,10 @@ func processAddEntry(htlc *DecodedDescriptor, ourBalance, theirBalance uint64,
 // non-dust satoshi balance. It also checks and returns whether we need a local
 // and/or remote anchor output.
 func SanityCheckAmounts(ourBalance, theirBalance btcutil.Amount,
-	ourAssetBalance, theirAssetBalance uint64, view *DecodedView,
-	chanType channeldb.ChannelType, whoseCommit lntypes.ChannelParty,
-	dustLimit btcutil.Amount) (bool, bool, error) {
+	ourAssetBalance, theirAssetBalance uint64, assetView,
+	nonAssetView *DecodedView, chanType channeldb.ChannelType,
+	whoseCommit lntypes.ChannelParty, dustLimit btcutil.Amount) (bool, bool,
+	error) {
 
 	log.Tracef("Sanity checking amounts, whoseCommit=%v, ourBalance=%d, "+
 		"theirBalance=%d, ourAssetBalance=%d, theirAssetBalance=%d",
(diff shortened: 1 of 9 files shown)


tapchannel/commitment.go
```diff
@@ -363,20 +363,24 @@ func processAddEntry(htlc *DecodedDescriptor, ourBalance, theirBalance uint64,
...
 		"theirBalance=%d, ourAssetBalance=%d, theirAssetBalance=%d",
 		whoseCommit, ourBalance, theirBalance, ourAssetBalance,
 		theirAssetBalance)
 
 	var (
-		numHTLCs int64
-		feePerKw = view.FeePerKw
+		// When determining whether we need an anchor output, we need to
+		// count the non-asset HTLCs as well.
+		numHTLCs = uint64(len(nonAssetView.OurUpdates) +
```
GeorgeTsagk, line 379: would this lead to stuck force-closes? (covered in the [LitD itest](https://github.com/lightninglabs/lightning-terminal/blob/98c9819b81d03f926f72656d9c0fb33c010e933e/itest/litd_custom_channels_test.go#L3533-L3537)?)

trying to understand if we ever had coverage for this

GeorgeTsagk (approved):
LGTM! :goat:

GeorgeTsagk:
when we do "breaking version bumps" I wonder how can we get the [`Lit Tests`](https://github.com/lightninglabs/taproot-assets/actions/runs/12162324676/job/33918992171?pr=1236) to run correctly and show :green_circle: instead of compilation errors

tapchannel/commitment.go
guggero: No, the channel will not force close (at least from my observations), but the HTLC will be rejected. This is covered in the new `litd` itest, where we pay sats over an asset channel.

guggero:
> when we do "breaking version bumps" I wonder how can we get the [`Lit Tests`](https://github.com/lightninglabs/taproot-assets/actions/runs/12162324676/job/33918992171?pr=1236) to run correctly and show 🟢 instead of compilation errors
> 
> Could maybe pin the LitD commit that we want to use to run the itests, then CI would use that

Yeah, it's not optimal. We could point the litd itest to an unmerged commit, but IMO that's also not great, since we need to merge `tapd` first, then update `litd` which leads to a new commit...

So I think the best strategy is just to always follow up a `tapd` change that doesn't pass the LiT itests with a PR that shows it will eventually pass.

tapchannel/commitment.go
Roasbeef: Won't this result in us _over counting_? The non asset view will still have an manifest HTLC for a given asset HTLC. I think it's enough to just have this combined sum here, and drop the `numHTLCs` increment in the inner loop.

Roasbeef (changes requested):

tapchannel/commitment.go
guggero: No, the non-asset view and the asset view are fully distinct. And even if we were over counting, we only use `numHTLCs > 0` as a condition for determining whether we need a commitment anchor output.

guggero: I thought about this again and my initial approach wasn't 100% correct. It is correct that the `assetView` and
`nonAssetView` are non-overlapping. But we need to replicate this behavior:
https://github.com/lightningnetwork/lnd/blob/ab7aae07085ec244f7888e990e8d8cc6ae6dfa83/lnwallet/commitment.go#L706

So we also need to check for dust, which I now added.

GeorgeTsagk (approved):
LGTM pt.2 :star2:

ffranr (approved):

Issue #1226, opened 2024-12-02
[feature]: optionally generate asset group key without internal key tweak during minting


ffranr:
The asset group key is derived by applying two distinct tweaks to a raw public key: the Internal Key Tweak (known as the "single tweak" in code) and the Tapscript Root Tweak. The Internal Key Tweak ensures the asset group key is unique to the genesis asset ID, while the Tapscript Root Tweak binds the key to the conditions encoded in the Taproot script tree. The resulting key allows for securely identifying and reissuing assets within a group.

### Problem: Asset Group PSBT External Signing

Applying the Internal Key Tweak to the raw signing public key prevents an external PSBT hardware wallet signer from signing the PSBT, which is necessary for generating the asset group witness and for validation within the tap VM. This limitation arises because the PSBT signer cannot recover the raw signing public key from the tweaked internal key, as the Internal Key Tweak obscures the original public key.

In addition, the PSBT format, along with most hardware wallets, currently lacks support for specifying the Internal Key Tweak. Such support would be necessary to identify the underlying raw signing key and correctly derive the tweaked Taproot internal key.

### Solution Part 1: Adjust Asset Group Key Generation Process

The proposed solution introduces a flag to allow skipping the Internal Key Tweak step when generating an asset group key. The implementation will proceed as follows:

1. Add a flag to the `MintAsset` RPC endpoint to specify whether the Internal Key Tweak step should be skipped. This flag defaults to `false`.
2. Populate the flag in the `tapgarden.Seedling` struct to carry the configuration through the minting process.
3. Pass the flag to the `GroupKeyRequest` type struct to ensure the choice persists through the relevant logic.
4. Update the `asset.GroupPubKey` function to respect the flag, bypassing the Internal Key Tweak step if the flag is set.

Proposed flag name: `SkipInternalKeyTweak`.

### Solution Part 2: Adjust GroupKeyReveal Verification Process

To ensure backward compatibility while supporting asset group keys generated without the Internal Key Tweak, the `GroupKeyReveal` verification process needs to be adjusted as follows:

1. Attempt to derive the asset group key by skipping the Internal Key Tweak. Apply only the second tweak (Tapscript Root Tweak) to the raw signing key.
2. If the derived key matches the asset group key, the verification is complete.
3. If the derived key does not match, apply the Internal Key Tweak as usual, followed by the second tweak, and check against the asset group key.
4. If neither derivation matches, the verification fails.

[...]

PR #1246, opened 2024-12-09
Add `GroupKeyReveal` V1

ffranr:
This PR introduces a new group reveal structure in which the tapscript tree is tweaked instead of the internal key being modified with the asset ID. As a result, an external PSBT signer no longer needs to handle the tweaking of an internal key prior to signing.

Work towards solving: https://github.com/lightninglabs/taproot-assets/issues/1226

### DIFF
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -918,6 +918,25 @@ type GroupVirtualTx struct {
 	TweakedKey btcec.PublicKey
 }
 
+// GroupKeyRevealVersion specifies the version of the group key reveal data.
+type GroupKeyRevealVersion uint8
+
+const (
+	// GroupKeyRevealVersion0 specifies version number 0 for a group key
+	// reveal data structure.
+	GroupKeyRevealVersion0 GroupKeyRevealVersion = 0
+
+	// GroupKeyRevealVersion1 specifies version number 1 for a group key
+	// reveal data structure.
+	GroupKeyRevealVersion1 GroupKeyRevealVersion = 1
+)
+
+const (
+	// LatestGroupKeyRevealVersion is the latest supported group key reveal
+	// version.
+	LatestGroupKeyRevealVersion = GroupKeyRevealVersion1
+)
+
 // GroupKeyReveal represents the data used to derive the adjusted key that
 // uniquely identifies an asset group.
 type GroupKeyReveal interface {
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -957,6 +957,193 @@ type GroupKeyReveal interface {
 	GroupPubKey(assetID ID) (*btcec.PublicKey, error)
 }
 
+// GroupKeyRevealTapscript represents the data structure used to derive the
+// tweaked tapscript root, which is subsequently used to compute the asset
+// group key.
+//
+// The tapscript tree ensures that the derived asset group key is unique
+// to a specific genesis asset ID. This design prevents asset group keys from
+// being reused across different genesis assets or non-compliant asset minting
+// tranches (e.g., tranches of a different asset type).
+//
+// The uniqueness of the asset group key to a specific genesis asset ID is
+// guaranteed by including the genesis asset ID as a leaf in the tapscript tree.
+// Additionally, we ensure that the tapscript tree cannot include multiple
+// recognizable genesis asset IDs. To achieve this, we enforce that the genesis
+// asset ID must appear in the first leaf layer of the tapscript tree. At this
+// tree level, there are exactly two leaf positions. We use the remaining
+// revealed structure of the tree to prove that the genesis asset ID is
+// correctly committed and that no conflicting or duplicate genesis asset IDs
+// are committed.
+//
+// The tapscript tree can include a custom tapscript subtree, enabling users
+// to incorporate custom script spend paths. However, in the simplest case,
+// if a custom tapscript tree is not included, the tapscript tree adopts the
+// following structure:
+//
+//	           [tapscript root]
+//	               /       \
+//	[genesis asset ID]   [internal key hash]
+//
+// Here, the tapscript tree consists of two leaf nodes: the genesis asset ID
+// and the hash of the taproot internal key. In this case, we can prove that
+// the genesis asset ID is committed to the tree root by deriving the tapscript
+// root hash using the hash of the internal key. Further, the internal key hash
+// is not a valid asset ID because its pre-image is 33 bytes long (the
+// serialized compressed public key), making it cryptographically improbable
+// to serve as a valid pre-image for an asset ID (which requires a longer
+// pre-image).
+//
+// When a custom tapscript tree is provided, the tapscript tree adopts the
+// following structure:
+//
+//	           [tapscript root]
+//	               /       \
+//	[genesis asset ID]   [tweaked custom tree root]
+//	                            /        \
+//	             [internal key hash]   [custom tree root]
+//
+// In this scenario, we can prove that the root hash commits to the genesis
+// asset ID by deriving the tapscript root from the tweaked custom tree root
+// and the genesis asset ID. Specifically, the tapscript root hash is
+// reconstructed by concatenating the genesis asset ID and the tweaked custom
+// tree root, and then hashing them.
+//
+// Additionally, we can prove that the tweaked custom tree root is not a valid
+// asset ID because it is derived from the internal key hash and the custom
+// tree root. By including the internal key hash as a leaf, the structure
+// ensures that the tweaked custom tree root cannot be misinterpreted as a
+// valid genesis asset ID.
+//
+// Note that the scripts in the tapscript tree will be made non-executable using
+// OP_RETURN, except for the custom tree root.
+type GroupKeyRevealTapscript struct {
+	// finalRoot is the final tapscript root after all tapscript tweaks have
+	// been applied. The asset group key is derived from this root and the
+	// internal key.
+	finalRoot chainhash.Hash
+
+	// customTapscriptRoot is the optional root of a custom tapscript tree
+	// that includes script spend conditions for the group key.
+	customTapscriptRoot fn.Option[chainhash.Hash]
+}
+
+// Validate checks that the group key reveal tapscript is well-formed and
+// compliant.
+func (g *GroupKeyRevealTapscript) Validate(assetID ID,
+	internalKey btcec.PublicKey) error {
+
+	var emptyHash chainhash.Hash
+
+	// Ensure that the final root is not empty.
+	if g.finalRoot == emptyHash {
+		return fmt.Errorf("group key reveal final tapscript root is " +
+			"empty")
+	}
+
+	// If an exclusion proof is specified, ensure that it is not empty.
+	err := fn.MapOptionZ(
+		g.customTapscriptRoot,
+		func(root chainhash.Hash) error {
+			if root == emptyHash {
+				return fmt.Errorf("group key reveal " +
+					"tapscript root asset ID exclusion " +
+					"proof is specified but empty")
+			}
+			return nil
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	// TODO(ffranr): Verify that the inclusion and exclusion proofs are
+	//  valid using the internal key and asset ID.
+
+	return nil
+}
+
+// GroupKeyRevealV1 is a version 1 group key reveal type for representing the
+// data used to derive and verify the tweaked key used to identify an asset
+// group.
+type GroupKeyRevealV1 struct {
+	// tapInternalKey refers to the internal key used to derive the asset
+	// group key. Typically, this internal key is the user's signing public
+	// key.
+	tapInternalKey SerializedKey
+
+	// tapscriptTree is the tapscript tree that commits to the genesis asset
+	// ID and any script spend conditions for the group key.
+	tapscriptTree GroupKeyRevealTapscript
+}
+
+// Ensure that GroupKeyRevealV1 implements the GroupKeyReveal interface.
+var _ GroupKeyReveal = (*GroupKeyRevealV1)(nil)
+
+// NewGroupKeyRevealV1 creates a new version 1 group key reveal instance.
+func NewGroupKeyRevealV1(tapInternalKey SerializedKey,
+	finalRoot chainhash.Hash,
+	customTapscriptRoot fn.Option[chainhash.Hash]) GroupKeyReveal {
+
+	return &GroupKeyRevealV1{
+		tapInternalKey: tapInternalKey,
+		tapscriptTree: GroupKeyRevealTapscript{
+			finalRoot:           finalRoot,
+			customTapscriptRoot: customTapscriptRoot,
+		},
+	}
+}
+
+// RawKey returns the raw key of the group key reveal.
+func (g *GroupKeyRevealV1) RawKey() SerializedKey {
+	return g.tapInternalKey
+}
+
+// SetRawKey sets the raw key of the group key reveal.
+func (g *GroupKeyRevealV1) SetRawKey(rawKey SerializedKey) {
+	g.tapInternalKey = rawKey
+}
+
+// TapscriptRoot returns the tapscript root of the group key reveal.
+func (g *GroupKeyRevealV1) TapscriptRoot() []byte {
+	return g.tapscriptTree.finalRoot[:]
+}
+
+// SetTapscriptRoot sets the tapscript root of the group key reveal.
+func (g *GroupKeyRevealV1) SetTapscriptRoot(tapscriptRootBytes []byte) {
+	var tapscriptRoot chainhash.Hash
+	copy(tapscriptRoot[:], tapscriptRootBytes)
+
+	g.tapscriptTree.finalRoot = tapscriptRoot
+}
+
+// GroupPubKey returns the group public key derived from the group key reveal.
+func (g *GroupKeyRevealV1) GroupPubKey(assetID ID) (*btcec.PublicKey, error) {
+	internalKey, err := g.RawKey().ToPubKey()
+	if err != nil {
+		return nil, fmt.Errorf("group reveal raw key invalid: %w", err)
+	}
+
+	return GroupPubKeyV1(internalKey, assetID, g.tapscriptTree)
+}
+
+// GroupPubKeyV1 derives a version 1 asset group key from a signing public key
+// and a tapscript tree.
+func GroupPubKeyV1(internalKey *btcec.PublicKey, assetID ID,
+	tapscriptTree GroupKeyRevealTapscript) (*btcec.PublicKey, error) {
+
+	err := tapscriptTree.Validate(assetID, *internalKey)
+	if err != nil {
+		return nil, fmt.Errorf("group key reveal tapscript tree "+
+			"invalid: %w", err)
+	}
+
+	tapOutputKey := txscript.ComputeTaprootOutputKey(
+		internalKey, tapscriptTree.finalRoot[:],
+	)
+	return tapOutputKey, nil
+}
+
 // GroupKeyRevealV0 is a version 0 group key reveal type for representing the
 // data used to derive the tweaked key used to identify an asset group. The
 // final tweaked key is the result of: TapTweak(groupInternalKey, tapscriptRoot)
--- a/proof/encoding.go
+++ b/proof/encoding.go
@@ -8,8 +8,10 @@ import (
 
 	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/taproot-assets/asset"
+	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightningnetwork/lnd/tlv"
 )
 
@@ -466,6 +468,11 @@ func GenesisRevealDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
 }
 
 func GroupKeyRevealEncoder(w io.Writer, val any, buf *[8]byte) error {
+	// TODO(ffranr): When encoding V1 and onwards, we must fill rawKey,
+	//  tapscriptRoot, and version. Ensuring these fields are populated will
+	//  mean that older tapd nodes will reject the group key reveal cleanly
+	//  (and not try to erroneously parse an unsupported group key reveal).
+
 	if t, ok := val.(*asset.GroupKeyReveal); ok {
 		key := (*t).RawKey()
 		if err := asset.SerializedKeyEncoder(w, &key, buf); err != nil {
@@ -492,14 +499,108 @@ func GroupKeyRevealDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
 		if err != nil {
 			return err
 		}
+
+		// Compute remaining bytes. This calculation will not underflow
+		// because we have already verified that the length is at least
+		// the size of the compressed public key.
 		remaining := l - btcec.PubKeyBytesLenCompressed
-		var tapscriptRoot []byte
-		err = tlv.DVarBytes(r, &tapscriptRoot, buf, remaining)
-		if err != nil {
-			return err
+
+		// Attempt to read the tapscript root bytes if they are present.
+		var tapscriptRootBytes []byte
+		if remaining >= 32 {
+			// At this point, there are at least 32 bytes remaining
+			// which means that there is a tapscript root present.
+			// Read the tapscript root bytes.
+			err = tlv.DVarBytes(r, &tapscriptRootBytes, buf, 32)
+			if err != nil {
+				return err
+			}
+
+			// Update the remaining bytes length counter.
+			remaining -= 32
 		}
 
-		*typ = asset.NewGroupKeyRevealV0(rawKey, tapscriptRoot)
+		// Set a nil taproot root to an empty slice. This ensures that
+		// the encoding/decoding round trip is consistent.
+		if tapscriptRootBytes == nil {
+			tapscriptRootBytes = []byte{}
+		}
+
+		// If there are still bytes remaining, then the next byte should
+		// be the group key reveal version.
+		var version asset.GroupKeyRevealVersion
+		if remaining > 0 {
+			var v uint64
+			err = tlv.DUint8(r, &v, buf, 1)
+			if err != nil {
+				return err
+			}
+
+			version = asset.GroupKeyRevealVersion(v)
+
+			// Update the remaining bytes length counter.
+			remaining -= 1
+		}
+
+		// If the parsed version is greater the latest group key reveal
+		// version, then we reject the group key reveal.
+		//
+		// It is important to cleanly reject future versions of group
+		// key reveals that are not supported by this version of tapd.
+		// This safeguards compatibility for future upgrades to the
+		// group key reveal format.
+		if version > asset.LatestGroupKeyRevealVersion {
+			return fmt.Errorf("unsupported group key reveal "+
+				"version %d", version)
+		}
+
+		// If this is a version 0 group key reveal, then we can return
+		// the group key reveal now.
+		if version == asset.GroupKeyRevealVersion0 {
+			*typ = asset.NewGroupKeyRevealV0(
+				rawKey, tapscriptRootBytes,
+			)
+			return nil
+		}
+
+		// At this point, we know this is a version 1 group key reveal.
+		// Future versions could parse different fields from this point
+		// on.
+		//
+		// For clarity and robustness, we explicitly check for version
+		// 1.
+		if version != asset.GroupKeyRevealVersion1 {
+			return fmt.Errorf("code error: expected group reveal "+
+				"version 1, got %d", version)
+		}
+
+		// We can now cast the tapscript root bytes to a hash.
+		var tapscriptRoot chainhash.Hash
+		copy(tapscriptRoot[:], tapscriptRootBytes)
+
+		// The remaining bytes should constitute the custom tapscript
+		// tree root.
+		var customTapscriptRoot fn.Option[chainhash.Hash]
+		if remaining >= chainhash.HashSize {
+			var rootBytes [32]byte
+			err = tlv.DBytes32(
+				r, &rootBytes, buf,
+				chainhash.HashSize,
+			)
+			if err != nil {
+				return fmt.Errorf("unable to read custom "+
+					"tapscript tree root: %w", err)
+			}
+
+			var root chainhash.Hash
+			copy(root[:], rootBytes[:])
+			customTapscriptRoot = fn.Some(root)
+		}
+
+		*typ = asset.NewGroupKeyRevealV1(
+			rawKey, tapscriptRoot, customTapscriptRoot,
+		)
+
 		return nil
 	}
 	return tlv.NewTypeForEncodingErr(val, "GroupKeyReveal")
--- a/proof/encoding.go
+++ b/proof/encoding.go
@@ -2,7 +2,6 @@ package proof
 
 import (
 	"bytes"
-	"crypto/sha256"
 	"fmt"
 	"io"
 	"math"
@@ -480,10 +479,6 @@ func GroupKeyRevealEncoder(w io.Writer, val any, buf *[8]byte) error {
 }
 
 func GroupKeyRevealDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
-	if l > btcec.PubKeyBytesLenCompressed+sha256.Size {
-		return tlv.ErrRecordTooLarge
-	}
-
 	if l < btcec.PubKeyBytesLenCompressed {
 		return fmt.Errorf("%w: group key reveal too short",
 			ErrProofInvalid)


ffranr:
I've added a long doc comment to `GroupKeyRevealTapscript`. Please review what I've described there. Let me know if you agree/disagree.

asset/asset.go
```diff
@@ -938,6 +957,193 @@ type GroupKeyReveal interface {
...
+// The tapscript tree ensures that the derived asset group key is unique
+// to a specific genesis asset ID. This design prevents asset group keys from
+// being reused across different genesis assets or non-compliant asset minting
+// tranches (e.g., tranches of a different asset type).
+//
+// The uniqueness of the asset group key to a specific genesis asset ID is
+// guaranteed by including the genesis asset ID as a leaf in the tapscript tree.
+// Additionally, we ensure that the tapscript tree cannot include multiple
+// recognizable genesis asset IDs. To achieve this, we enforce that the genesis
+// asset ID must appear in the first leaf layer of the tapscript tree. At this
```
guggero, line 973: The BIP for TAP mentions "the leftmost or rightmost" leaf, making sure it's at the highest level. And I think in practice we have it in the rightmost (last) position in the lnd channel scripts. So I think we should probably use the same wording here, to make sure it's not confusing? Especially if we end up using the same code?

asset/asset.go
```diff
@@ -938,6 +957,193 @@ type GroupKeyReveal interface {
...
+// revealed structure of the tree to prove that the genesis asset ID is
+// correctly committed and that no conflicting or duplicate genesis asset IDs
+// are committed.
+//
+// The tapscript tree can include a custom tapscript subtree, enabling users
+// to incorporate custom script spend paths. However, in the simplest case,
+// if a custom tapscript tree is not included, the tapscript tree adopts the
+// following structure:
+//
+//	           [tapscript root]
```
guggero, line 984: I think there might be some confusion on how the Taproot key creation works today.

This is my sketch of how things work with just Taproot:
```
              (taproot output key)                                 
                       |                                           
       +---------------+----------+                                
       |                          |                                
 (internal key)            [tapscript root]                        
                                  |                                
                     +------------+-------------+                  
                     |                          |                  
               [leaf 1 hash]             [leaf 2 hash]             
                   <leaf 1>                  <leaf 2> 

[] = hash
() = pubkey
<> = script

Where (taproot output key) = (internal key) + TapTweak((internal key) || [tapscript root])*G
```

But with the asset ID as a top-level tree element, it would look like this:

```
             (taproot output key)                                  
                      |                                            
      +---------------+----------+                                 
      |                          |                                 
(internal key)            [tapscript root]                         
                                 |                                 
                    +------------+-------------+                   
                    |                          |                   
              [gen asset ID]                                       
                                        [tapscript sibling hash]   
                                                                   
                                  +------------+-------------+     
                                  |                          |     
                            [leaf 1 hash]             [leaf 2 hash]
                                <leaf 1>                  <leaf 2> 
```

So not sure if the rest of the text below still works with this corrected view?

guggero (commented):
Only looked at the Godoc comment of `GroupKeyRevealTapscript`.

asset/asset.go
Roasbeef: With either diagram, I think we should make it clear where the tapscript merkle tree ends, and the EC tweaking operations start. At a glance for both, it looks like the internal key is actually part of the merkle tree.

asset/asset.go
```diff
@@ -957,6 +957,193 @@ type GroupKeyReveal interface {
...
+// tree root, and then hashing them.
+//
+// Additionally, we can prove that the tweaked custom tree root is not a valid
+// asset ID because it is derived from the internal key hash and the custom
+// tree root. By including the internal key hash as a leaf, the structure
+// ensures that the tweaked custom tree root cannot be misinterpreted as a
+// valid genesis asset ID.
+//
+// Note that the scripts in the tapscript tree will be made non-executable using
+// OP_RETURN, except for the custom tree root.
```
Roasbeef, line 1019: `OP_RETURN` for the genesis ID leaf?

asset/asset.go
```diff
@@ -957,6 +957,193 @@ type GroupKeyReveal interface {
...
+					"proof is specified but empty")
+			}
+			return nil
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	// TODO(ffranr): Verify that the inclusion and exclusion proofs are
```
Roasbeef, line 1060: I think for this, we can re-use the existing control block routines. You can treat the branch of the custom scripts as another element (need to check that it won't do the leaf hash instead of the branch hash). If it does assume a leaf hash, then it may be worth it to refactor things slightly to allow specifying that. 

Then the reveal portion comes down to:
  * validate the control block proof (we have routines for that already)
  * use the script tokenizer from `btcsuite/txscript` to verify the proepr asset ID commitment

proof/encoding.go
```diff
@@ -492,14 +499,108 @@ func GroupKeyRevealDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
...
+		// the size of the compressed public key.
 		remaining := l - btcec.PubKeyBytesLenCompressed
-		var tapscriptRoot []byte
-		err = tlv.DVarBytes(r, &tapscriptRoot, buf, remaining)
-		if err != nil {
-			return err
+
+		// Attempt to read the tapscript root bytes if they are present.
+		var tapscriptRootBytes []byte
+		if remaining >= 32 {
```
Roasbeef, line 510: In retrospect, I think we should've gone with a nest TLV here, so one value of the key, and another of the tapscript root. Then we wouldn't need this special casing to try and detect what was encoded.

Roasbeef: We also have an option of using a new TLV type here, making them mutually exclusive. 

Do we know how an old client will react if the old reveal is just _missing_ all together (no type specified).

proof/encoding.go
```diff
@@ -480,10 +479,6 @@ func GroupKeyRevealEncoder(w io.Writer, val any, buf *[8]byte) error {
 }
 
 func GroupKeyRevealDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
-	if l > btcec.PubKeyBytesLenCompressed+sha256.Size {
```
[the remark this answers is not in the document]
ffranr: Older nodes will have to update, they won't be able to validate the new group key reveals. However, updated nodes will still be able to handle the older group key reveal versions.

asset/asset.go
ffranr: Yes, and for the internal key leaf.

proof/encoding.go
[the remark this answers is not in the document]
Roasbeef: What I mean is that won't older nodes reject the new proofs all together due to a parsing failure? Which will cause temporary Universe state divergence.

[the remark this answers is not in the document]
ffranr: Yes, the older nodes won't be able to correctly parse the new group key reveal. But they should receive an error message rather than decode incorrectly. Older nodes will have to upgrade to handle newly minted assets. And so will universe servers.

taprpc/group_key_tapscript_test.go
```diff
@@ -0,0 +1,237 @@
...
+	rootCheckBytes = userSubtreeControlBlock.RootHash(
+		customScriptLeaf.Script,
+	)
+	copy(rootCheck[:], rootCheckBytes)
+
+	require.Equal(t, actualTaprootHash2, rootCheck)
+
+	// ---------------------------------------------------------------------
+	//
+	// Example with an optional user-provided root hash only.
```
guggero, line 155: Nice, thank you for outlining the three different approaches.

IMO the decision boils down to:
Proof size, privacy and BIP complexity: The "preimage based" approach is slightly larger and would expose the script (if a single one is used) in the reveal, but describing it would be slightly easier in the BIP (since we already have the tapscript sibling approach in other places).

So I'm okay with either of them. Thanks for your patience and persistence in explaining your approach to me :see_no_evil:

guggero (commented):
Great work! I think if we can just get this part into v0.5.0 we should be set for future upgrades.

asset/asset.go
```diff
@@ -944,6 +944,364 @@ type GroupKeyReveal interface {
...
+	// Ensure that the final root is not all zeros.
+	if g.finalRoot == zeroHashBytes {
+		return fmt.Errorf("group key reveal final tapscript root is " +
+			"empty")
+	}
+
+	// Extract custom tapscript subtree root if specified. This can be a
+	// hash of a leaf or branch node. We use an empty hash (all zero byte
+	// array) if the custom tapscript root is not specified.
+	customTapscriptRoot := g.customTapscriptRoot.UnwrapOr(chainhash.Hash{})
```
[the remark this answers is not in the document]
ffranr: > Should we really be inserting a "fake" root here? IMO it would feel more natural to just promote the internal key leaf to the same level as the asset ID leaf instead of inserting a fake, all-zero branch.

Using all zero as the custom tapscript root if absent has no cost to the user, because there isn't a script spend path at that point. My goal with that approach was to have a consistent tree topology whether a custom branch was specified or not. Which means there is only one tapscript root to try to reproduce during verification.

> Either way, I think this approach requires a bit more explanation in a BIP than the "there's a sibling you need to add to the root of your tree" approach I was suggesting.

This means that the user has to modify their tree before tapd can accept it. In my opinion, it would be simpler for the user if they didn’t have to modify their tree at all. I think Tapd should be able to accept any user branch as-is and then provide a control block proof in return.

asset/asset.go
```diff
@@ -944,6 +944,364 @@ type GroupKeyReveal interface {
...
+	var tapscriptRootBytes [32]byte
+	copy(tapscriptRootBytes[:], g.tapscriptTree.finalRoot[:])
+
+	if err := tlv.EBytes32(w, &tapscriptRootBytes, buf); err != nil {
+		return fmt.Errorf("group key reveal final tapscript root "+
+			"encode error: %w", err)
+	}
+
+	// Encode the version into the writer.
+	groupKeyRevealVersion := uint8(1)
```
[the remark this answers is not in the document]
ffranr: Yes, the new group key reveal will be TLV encoded.

asset/asset.go
```diff
@@ -944,6 +944,364 @@ type GroupKeyReveal interface {
...
+	g.tapInternalKey = rawKey
+	g.tapscriptTree.finalRoot = finalTapscriptRoot
+	g.tapscriptTree.customTapscriptRoot = customTapscriptRoot
+
+	return nil
+}
+
+// RawKey returns the raw key of the group key reveal.
+func (g *GroupKeyRevealV1) RawKey() SerializedKey {
+	return g.tapInternalKey
```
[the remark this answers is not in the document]
ffranr: In my previous formulation I wasn't doing dispatch on decode/encode so I added these methods, but we can remove them now.

asset/asset.go
```diff
@@ -944,6 +944,364 @@ type GroupKeyReveal interface {
...
+// Note that the scripts in the tapscript tree will be made non-executable using
+// OP_RETURN, except for the custom tree root.
+type GroupKeyRevealTapscript struct {
+	// finalRoot is the final tapscript root after all tapscript tweaks have
+	// been applied. The asset group key is derived from this root and the
+	// internal key.
+	finalRoot chainhash.Hash
+
+	// customTapscriptRoot is the optional root of a custom tapscript tree
+	// that includes script spend conditions for the group key.
```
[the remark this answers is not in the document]
ffranr: `customTapscriptRoot` is the user provided subtree hash. And `finalRoot` is the tapscript tree root hash which incorporates the asset ID and the user provided subtree.

`finalRoot` is never none going forwards because the tree will at least include asset ID.

`customTapscriptRoot` can be none because there may not be any user provided subtree.

asset/asset.go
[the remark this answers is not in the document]
ffranr: I really don't think tapd needs to do nor should do any of the following:

* user script validation (i.e. user single script leaf can't be asset ID)
* user subtree validation (i.e. user subtree must include sibling node)
* multiple possible validation paths (i.e. final tapscript tree could have different topologies)
* or even produce a user control block proof that can vary in size.

They are all just unnecessary UX problems and potential errors IMO.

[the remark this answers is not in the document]
ffranr: We can likely simplify the tapscript tree by removing the internal key leaf and using the following structure:
```
//	           [tapscript root]
//	               /       \
//	[genesis asset ID]    [tweaked user subtree]
//	                            /        \
//	                   [OP_RETURN 0]   [user tapscript subtree]
```

A spending script cannot be derived for `[tweaked user subtree]`. Moreover, it is computationally infeasible for a user to construct `[user tapscript subtree]` in a way that would make `[tweaked user subtree]` include a valid asset ID leaf script.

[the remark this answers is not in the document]
ffranr: We can likely simplify the tapscript tree by removing the internal key leaf and using the following structure:
```
//	           [tapscript root]
//	               /       \
//	[genesis asset ID]    [tweaked user subtree]
//	                            /        \
//	                   [OP_RETURN 0]   [user subtree]
```

A spending script cannot be derived for `[tweaked user subtree]`. Moreover, it is computationally infeasible for a user to construct `[user tapscript subtree]` in a way that would make `[tweaked user subtree]` a valid asset ID leaf.

I think this formulation might be simpler to document, have the same security properties, and be easier to implement.

ffranr:
The one thing we need to deal with (probably) is:

```
func GroupKeyRevealRecord(reveal *asset.GroupKeyReveal) tlv.Record {
	recordSize := func() uint64 {
		// TODO(ffranr): Do we need dispatch on record size? I think
		//  this is used for encoding only. And perhaps only to
		//  determine whether there's anything to encode at all?
		if reveal == nil || *reveal == nil {
			return 0
		}
		r := *reveal
		return uint64(
			btcec.PubKeyBytesLenCompressed + len(r.TapscriptRoot()),
		)
	}
	return tlv.MakeDynamicRecord(
		GroupKeyRevealType, reveal, recordSize,
		asset.GroupKeyRevealEncoder, asset.GroupKeyRevealDecoder,
	)
}
```

I didn't resolve this TODO.

ffranr:
> The one thing we need to deal with (probably) is:
> 
> ```
> func GroupKeyRevealRecord(reveal *asset.GroupKeyReveal) tlv.Record {
> 	recordSize := func() uint64 {
> 		// TODO(ffranr): Do we need dispatch on record size? I think
> 		//  this is used for encoding only. And perhaps only to
> 		//  determine whether there's anything to encode at all?
> 		if reveal == nil || *reveal == nil {
> 			return 0
> 		}
> 		r := *reveal
> 		return uint64(
> 			btcec.PubKeyBytesLenCompressed + len(r.TapscriptRoot()),
> 		)
> 	}
> 	return tlv.MakeDynamicRecord(
> 		GroupKeyRevealType, reveal, recordSize,
> 		asset.GroupKeyRevealEncoder, asset.GroupKeyRevealDecoder,
> 	)
> }
> ```
> 
> I didn't resolve this TODO.

Yes, not sure what to do here.

I think recordSize is only relevant for encoding. The problem is that GroupKeyRevealV1 is encoded as a TLV stream, so I’m not sure how to calculate the size of the encoded byte array. I don’t think it’s just the sum of the field sizes because of the TLV metadata.

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+func NewGroupKeyTapscriptRoot(genesisAssetID ID,
+	customRoot fn.Option[chainhash.Hash]) lfn.Result[GroupKeyRevealTapscript] {
+
+	// First, we compute the tweaked custom branch hash. This hash is
+	// derived by combining the hash of a non-spendable leaf and the root
+	// hash of the custom tapscript subtree.
+	//
+	// If a custom tapscript subtree root hash is provided, we use it.
+	// Otherwise, we default to an empty hash (a zero-filled byte array).
+	emptyNonSpendLeaf, err := NewNonSpendableScriptLeaf(nil).Unpack()
```
guggero, line 1112: For these helper functions where `nil, nil` isn't really a valid return combination, I don't really see the benefit of using a result type. Especially if we immediately call `Unpack()` in all call sites.
But non-blocking.

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+	return tlv.MakePrimitiveRecord((*GKRVersion).TypeVal(), version)
+}
+
+func NewGKRInternalKeyRecord(internalKey *SerializedKey) tlv.Record {
+	return tlv.MakePrimitiveRecord(
+		(*GKRInternalKey).TypeVal(), (*[33]byte)(internalKey),
+	)
+}
+
+func NewGKRTapsciptRootRecord(root *chainhash.Hash) tlv.Record {
```
guggero, line 996: nit: typo in word Tapscript, here and in the TLV type above.

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+	internalKey, err := btcec.ParsePubKey(g.internalKey[:])
+	if err != nil {
+		return lfn.Errf[txscript.ControlBlock]("failed to parse "+
+			"internal key: %w", err)
+	}
+
+	outputKey := txscript.ComputeTaprootOutputKey(
+		internalKey, g.tapscript.root[:],
+	)
+	outputKeyIsOdd := outputKey.SerializeCompressed()[0] == 0x03
```
guggero, line 1244: nit: use the `secp256k1.PubKeyFormatCompressedOdd` constant (can create a local copy if things get too long otherwise). Both here and in the test.

proof/records.go
```diff
@@ -380,6 +380,9 @@ func GenesisRevealRecord(genesis **asset.Genesis) tlv.Record {
 
 func GroupKeyRevealRecord(reveal *asset.GroupKeyReveal) tlv.Record {
 	recordSize := func() uint64 {
+		// TODO(ffranr): Do we need dispatch on record size? I think
```
guggero, line 383: No, this is only required for encoding. For decoding, the length is encoded in the record (the L in TLV). So can just panic here (which we usually do in these record size functions).

guggero (commented):
Very nice work on this, thanks for the extra push to completion!!!

Approach/Concept ACK. Code looks very good too. But will want to do another pass tomorrow with a fresh mind (but don't expect any significant changes).

Roasbeef:
> I think recordSize is only relevant for encoding. The problem is that GroupKeyRevealV1 is encoded as a TLV stream, so I’m not sure how to calculate the size of the encoded byte array

So that's just used to know what the outer length prefix should be. You can either compute a static value as you did there, or just encode it (without the type and length prefix) to see what the size would be if written as a normal TLV stream.

asset/asset.go
```diff
@@ -980,6 +983,41 @@ func (g *GroupKeyRevealV0) Encode(w io.Writer) error {
...
+// Decode decodes the group key reveal from the reader.
+func (g *GroupKeyRevealV0) Decode(r io.Reader, buf *[8]byte, l uint64) error {
+	// Verify that the group key reveal is not excessively long. This check
+	// is essential to prevent misinterpreting V1 and later group key
+	// reveals as V0.
+	if l > btcec.PubKeyBytesLenCompressed+sha256.Size {
+		return tlv.ErrRecordTooLarge
+	}
+
+	if l < btcec.PubKeyBytesLenCompressed {
```
Roasbeef, line 1446: Stye nit: could be combined above into a switch statement.

proof/encoding.go
```diff
@@ -480,32 +478,19 @@ func GroupKeyRevealEncoder(w io.Writer, val any, _ *[8]byte) error {
...
-		return tlv.ErrRecordTooLarge
+	typ, ok := val.(*asset.GroupKeyReveal)
+	if !ok {
```
Roasbeef: I like this short circuit approach in the TLV decoder, leads to less nesting!

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+			"non-spendable script: %w", err)
+	}
+
+	// Create a new tapscript leaf from the script.
+	leaf := txscript.NewBaseTapLeaf(script)
+	return lfn.Ok(leaf)
+}
+
+// TLV types for GroupKeyReveal.
+var (
```
Roasbeef, line 977: So you can just make these type aliases instead of make a new pointer. 

Here's an example from `lnd`: https://github.com/lightningnetwork/lnd/blob/8ecef03315f749ae8820b4735978ca2e0552de2e/lnwire/musig2.go#L10-L15

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+	GKRVersion           = new(tlv.TlvType0)
+	GKRInternalKey       = new(tlv.TlvType2)
+	GKRTapsciptRoot      = new(tlv.TlvType4)
+	GKRCustomSubtreeRoot = new(tlv.TlvType7)
+)
+
+func NewGKRVersionRecord(
+	version *uint8) tlv.Record {
```
Roasbeef, line 985: FWIW, for just about all these fields (primitive fields) you can use the new `RecordT` type, here's an example: https://github.com/lightningnetwork/lnd/blob/8ecef03315f749ae8820b4735978ca2e0552de2e/lnwire/musig2.go#L22-L24

So you can just have a single struct, and declare all the fields as attributes, it would look something like: 
```go

type (
   GKRVersionType           = new(tlv.TlvType0)
   GKRInternalKeyType       = new(tlv.TlvType2)
   GKRTapsciptRootType     = new(tlv.TlvType4)
   GKRCustomSubtreeRootType = new(tlv.TlvType7)
)
type GRoupKeyRevealV1 struct {
    Version tlv.RecordT[GKRVersionType, uint32]

    InternalKey tlv.RecordT[GKRInternalKeyType, *btcec.PublicKey]

   TapscriptRoot tlv.RecordT[GKRTapsciptRootType, [sha256.Size]byte]
  
   CustomTreeRoot tlv.Record[GKRCustomSubtreeRootType, [sha256.Size]byte]
}
```

Then your encode+decode is something like: 
```go
func (g *GRoupKeyRevealV1) Encode(w io.Writer) error {
    	stream, err := tlv.NewStream(
		g.Version.Record(), g.InternalKey.Record(), g.TapscriptRoot.Record(),g.CustomTreeRoot.Record(),
	)
	if err != nil {
		return err
	}

	return stream.Encode(w)
}

func (g *GRoupKeyRevealV1) Decode(r io.Reader) error {
  	stream, err := tlv.NewStream(
		g.Version.Record(), g.InternalKey.Record(), g.TapscriptRoot.Record(),g.CustomTreeRoot.Record(),
	)
	if err != nil {
		return err
	}

	return stream.Decode(r)
}

func eGkrV1(w io.Writer, val interface{}, _ *[8]byte) error {
	if v, ok := val.(*GRoupKeyRevealV1); ok {
		return v.Encode(w)
	}

	return tlv.NewTypeForEncodingErr(val, "*GRoupKeyRevealV1")
}

func dGkrV1(r io.Reader, val interface{}, _ *[8]byte, l uint64) error {
	if v, ok := val.(*GRoupKeyRevealV1); ok {
		return v.Decode(r)
	}

	return tlv.NewTypeForDecodingErr(val, "*GRoupKeyRevealV1", l, l)
}

func (g *GRoupKeyRevealV1) Record() tlv.Record {
	sizeFunc := func() uint64 {
		var buf bytes.Buffer
		err := pgEncode(&buf)
		if err != nil {
			panic(err)
		}
		return uint64(len(buf.Bytes()))
	}

	return tlv.MakeDynamicRecord(
		0, p, sizeFunc, eGkrV1, dGkrv1,
	)
}
```

Roasbeef: Then I think we can even add the `Record()` method to the old interface. So you just call that when makign the stream to handle encode/decode.

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+
+	// customSubtreeInclusionProof provides the inclusion proof for the
+	// custom tapscript subtree. It is required to spend the custom
+	// tapscript leaves within the tree.
+	//
+	// NOTE: This field should not be serialized as part of the group key
+	// reveal. It is included here to ensure it is constructed concurrently
+	// with the tapscript root, maintaining consistency and minimizing
+	// errors.
+	customSubtreeInclusionProof []byte
```
Roasbeef, line 1094: Any reason this isn't just `txscript.ControlBlock`? 

Why wouldn't it be serialized as part of the reveal? It includes the internal key and also the inclusion proof all the way up.

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+	//
+	// If a custom tapscript subtree root hash is provided, we use it.
+	// Otherwise, we default to an empty hash (a zero-filled byte array).
+	emptyNonSpendLeaf, err := NewNonSpendableScriptLeaf(nil).Unpack()
+	if err != nil {
+		return lfn.Err[GroupKeyRevealTapscript](err)
+	}
+
+	// Compute the tweaked custom branch hash.
+	tweakedCustomBranchHash := TapBranchHash(
```
Roasbeef, line 1118: If this is the only item we _could_ just make the asset ID commitment the sole element in the tree. This would add more cases we need to verify though, so I think this approach is fine.

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+		return fmt.Errorf("failed to compute tapscript root hash: %w",
+			err)
+	}
+
+	// Ensure that the final tapscript root hash matches the computed root
+	// hash.
+	customRoot := g.customSubtreeRoot.UnwrapOr(chainhash.Hash{})
+
+	if !g.root.IsEqual(&tapscript.root) {
+		return fmt.Errorf("failed to derive tapscript root from "+
```
Roasbeef, line 1177: Nice error message, will certainly save some debugging time later 👍

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+	}
+
+	return nil
+}
+
+// GroupKeyRevealV1 is a version 1 group key reveal type for representing the
+// data used to derive and verify the tweaked key used to identify an asset
+// group.
+type GroupKeyRevealV1 struct {
+	// version is the version of the group key reveal.
```
Roasbeef, line 1192: See my struct sketch above that combines the main struct definition and also TLV fields. Having to always do `.Val` can be seen as cumbersome, in which case we can expose getter methods.

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+		tapscript:   gkrTapscript,
+	})
+}
+
+// ScriptSpendControlBlock returns the control block for the script spending
+// path in the custom tapscript subtree.
+//
+// nolint: lll
+func (g *GroupKeyRevealV1) ScriptSpendControlBlock(
+	genesisAssetID ID) lfn.Result[txscript.ControlBlock] {
```
Roasbeef, line 1233: Maybe wrap this in another type of a `PartialControl` block? This gives the proof up to the custom tapscript root, but a spender will still need to provide the other merkle authentication hashes.

asset/asset.go
```diff
@@ -943,6 +944,447 @@ type GroupKeyReveal interface {
...
+		return err
+	}
+	return stream.Encode(w)
+}
+
+// Decode decodes the group key reveal from a reader.
+func (g *GroupKeyRevealV1) Decode(r io.Reader, buf *[8]byte, l uint64) error {
+	var customSubtreeRoot chainhash.Hash
+
+	tlvStream, err := tlv.NewStream(
```
Roasbeef, line 1310: Ah ok, I see now the decomposition of the reveal struct, and the auxiliary tapscript proof information.

proof/encoding.go
```diff
@@ -478,19 +480,40 @@ func GroupKeyRevealEncoder(w io.Writer, val any, _ *[8]byte) error {
 }
 
 func GroupKeyRevealDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
+	// Return early if the val is not a pointer to a GroupKeyReveal.
```
Roasbeef: So this is now actually a _pointer_ to an interface?

proof/records.go
Roasbeef: Is this the generic version? if so, it should dynamically determine the size just by encoding.

asset/group_key_reveal_test.go
```diff
@@ -0,0 +1,144 @@
...
+// GroupKeyReveal generates a GroupKeyReveal instance from the test case.
+func (tc testCaseGkrEncodeDecode) GroupKeyReveal() (GroupKeyReveal, error) {
+	gkr, err := NewGroupKeyRevealV1(
+		tc.internalKey, tc.genesisAssetID, tc.customSubtreeRoot,
+	).Unpack()
+
+	return &gkr, err
+}
+
+func TestGroupKeyTapscriptEncodeDecode(t *testing.T) {
```
Roasbeef, line 32: Here's a quick mafs version (may need some tweaking) of this test that uses `rapid`: 
```go
func TestGroupKeyTapscriptEncodeDecode(tt *testing.T) {
	ttt.Parallel()

	rapid.Check(tt, func(t *rapid.T) {
		// Generate random test inputs using rapid generators
		internalKey := *rapid.Custom(func(t *rapid.T) *btcec.PublicKey {
			priv, err := btcec.NewPrivateKey()
			if err != nil {
				t.Fatal(err)
			}
			return priv.PubKey()
		}).Draw(t, "internal_key")

		genesisAssetID := ID(rapid.SliceOfN([]byte{}, 32, 32).Draw(t, "genesis_id"))

		// Randomly decide whether to include a custom script
		hasCustomScript := rapid.Bool().Draw(t, "has_custom_script")

		var customSubtreeRoot fn.Option[chainhash.Hash]
		var customScriptLeaf *txscript.TapLeaf

		if hasCustomScript {
			// Generate random script between 1-100 bytes
			scriptSize := rapid.IntRange(1, 100).Draw(t, "script_size")
			customScript := rapid.SliceOfN([]byte{}, scriptSize, scriptSize).Draw(t, "custom_script")
			
			customScriptLeaf = txscript.NewBaseTapLeaf(customScript)
			customSubtreeRoot = fn.Some(customScriptLeaf.TapHash())
		} else {
			customSubtreeRoot = fn.None[chainhash.Hash]()
		}

		// Create and test GroupKeyReveal
		gkr, err := NewGroupKeyRevealV1(
			internalKey, 
			genesisAssetID,
			customSubtreeRoot,
		).Unpack()
		require.NoError(t, err)

		groupPubKey, err := gkr.GroupPubKey(genesisAssetID)
		require.NoError(t, err)

		// Test encoding/decoding
		var buffer bytes.Buffer
		var scratchBuffEncode [8]byte
		err = GroupKeyRevealEncoder(&buffer, &gkr, &scratchBuffEncode)
		require.NoError(t, err)

		var gkrDecoded GroupKeyReveal
		var scratchBuffDecode [8]byte
		err = GroupKeyRevealDecoder(
			&buffer,
			&gkrDecoded,
			&scratchBuffDecode,
			uint64(buffer.Len()),
		)
		require.NoError(t, err)

		// Prepare for comparison by removing non-encoded fields
		gkrV1, ok := gkr.(*GroupKeyRevealV1)
		require.True(t, ok)
		gkrV1.tapscript.customSubtreeInclusionProof = nil

		// Compare decoded with original
		require.Equal(t, gkrV1, gkrDecoded)

		// Verify decoded group public key
		groupPubKeyDecoded, err := gkrDecoded.GroupPubKey(genesisAssetID)
		require.NoError(t, err)
		require.Equal(t, groupPubKey, groupPubKeyDecoded)

		// Test custom subtree if present
		if customSubtreeRoot.IsSome() && customScriptLeaf != nil {
			gkrDecodedV1, ok := gkrDecoded.(*GroupKeyRevealV1)
			require.True(t, ok)

			ctrlBlock, err := gkrDecodedV1.ScriptSpendControlBlock(
				genesisAssetID,
			).Unpack()
			require.NoError(t, err)

			computedRoot := chainhash.Hash(
				ctrlBlock.RootHash(customScriptLeaf.Script),
			)
			require.Equal(t, gkrDecodedV1.tapscript.root, computedRoot)
		}
	})
}
```

Could be made more succinct with custom generators.

Roasbeef (changes requested):
Excellent work! 

Did an initial pass through the diff, at times suggesting changes with in-line code comments. I think the only thing missing here is some tests at the `proof.Proof` level, to make sure things can be properly encoded/decoded with the new reveal type. We should be able to tack this onto the existing randomized tests and test vector generation.

asset/asset.go
ffranr: I've gone with:
```
type GroupKeyRevealTlvType = tlv.Type

const (
	GKRVersion           GroupKeyRevealTlvType = 0
	GKRInternalKey       GroupKeyRevealTlvType = 2
	GKRTapscriptRoot     GroupKeyRevealTlvType = 4
	GKRCustomSubtreeRoot GroupKeyRevealTlvType = 7
)
```
for now because I've seen that pattern used quite a bit in the old TLV setups (before `tlv.RecordT`).

proof/encoding.go
ffranr: If `val` is asserted to be of type `GroupKeyReveal` rather than `*GroupKeyReveal`, and `val == nil`, the type assertion inside `GroupKeyRevealDecoder` (e.g., `typ, ok := val.(GroupKeyReveal)`) will fail. This happens because Go's type assertion mechanism requires both a dynamic type and a value to confirm the type, and a `nil` value with no dynamic type results in `ok == false`.

For example:

```go
var gkrDecoded GroupKeyReveal
var scratchBuffDecode [8]byte
err := GroupKeyRevealDecoder(
    &buffer, gkrDecoded, &scratchBuffDecode,
    uint64(buffer.Len()),
)
```

This will fail because `GroupKeyRevealDecoder` cannot determine whether `val` implements the `GroupKeyReveal` interface when `val == nil`. Since `gkrDecoded` is not a pointer, it has no dynamic type in this context, making the assertion invalid.

By contrast, if `val` is of type `*GroupKeyReveal`, the dynamic type is always present, even when the value is `nil`. This allows `GroupKeyRevealDecoder` to qualify `val == nil` correctly.

I think there might be a way around this using the `reflect` package but I haven't gotten it to work yet.

asset/asset.go
ffranr: I played around with using `tlv.RecordT` fields for a while but I couldn't get everything to work correctly.

I got lost in the weeds a bit trying to change all the types and modifying how the fields were used. I got thrown converting `customSubtreeRoot fn.Option[chainhash.Hash]` into a `tlv.OptionalRecordT`. Maybe the key was to drop `chainhash.Hash` in favour of `[sha256.Size]byte`.

I wouldn't mind giving the `tlv.RecordT` approach another go, I've managed to use it in RFQ `requestWireMsgData`, but would you mind if I do that after rc3?

asset/asset.go
ffranr: My thinking here is not to include `txscript.ControlBlock` in the serialization to avoid redundancy. Breakdown:

- **`InternalKey`**: Already serialized separately.  
- **`OutputKeyYIsOdd`**: Derivable from internal key and tapscript root.  
- **`LeafVersion`**: Typically constant; serialization might not be necessary?
- **`InclusionProof`**: Derivable from the genesis asset ID which should be available separately whenever group key reveal is at hand.  

I'm aiming to eliminate data duplication, ensuring that each field serves as a definitive "source of truth." And also reduce serialization size.

I suppose if the default value of `LeafVersion` changes we could version bump group key reveal? Maybe we should just include it now?

asset/group_key_reveal_test.go
ffranr: Latest commit will include this test :+1: Thanks!

ffranr:
> Must be added to tapgarden/planter_test.go to work.

Works great, thank you!

asset/asset.go
ffranr: hmm good point.

I would like to revise this for 0.6. Right now, it's useful for testing (checking the tapscript root).

asset/asset.go
ffranr: This commit is a code move/copy. I was trying not to make any code changes other than copying to maximise clarity and minimise the review burden. I agree that switch statement would be nice! Perhaps after this I can do some refactoring.

ffranr:
I've added a small unit test which covers `Proof` encoding/decoding with `GroupKeyRevealV1`.

I think I've addressed all review comments.

Roasbeef (commented):
> Reviewed 2 of 3 files at r1, 3 of 3 files at r2, 1 of 1 files at r3, 1 of 2 files at r5, 2 of 4 files at r6, 3 of 3 files at r7, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1246)* status: all files reviewed, 34 unresolved discussions (waiting on @ffranr and @guggero)

guggero (approved):
Awesome work, LGTM :tada:

Issue #888, opened 2024-04-23
[enhancement]: Premature Taproot-Asset Channel Exhaustion Due to Above-dust `lnd` Requirement


dstadulis:
# Background
In scenarios where taproot assets are contained within bitcoin lightning channels,
LND has the requirement to push an above-dust amount of satoshis to the remote side of the channel to complete a HTLC payment.  

Potential complication with this LND requirement is: 
In the scenario a taproot-assets-enabled lightning channel contains a de minimis amount of satoshis, (in situ taproot-assets outputs have limits -- until #721 is fixed) the bitcoin satoshi-balance of the channel will be exhausted after a few (3-5?) payments (assuming a payment in the `to_local` direction doesn't occur).

# Deliverables
A method to ensure that a taproot-assets lightning channel can send and receive HTLCs without being constrained by the LND above-dust requirement.

# Proposed Remediation Options
* add an inbound fee discount of the incoming channel w/ a base fee the same as the asset carrier HTLC amt
* ?Alternatives

# Necessity
This functionality isn't required to demonstrate tapd payment-channel functionality but will be necessary to meet user-expectations before LL endorses Taproot-assets payment-channel functionality.

jharveyb:
You can also set up a matching HTLC (with no anchored assets) in the opposite direction to net out the BTC amount transferred.

Roasbeef:
> You can also set up a matching HTLC (with no anchored assets) in the opposite direction to net out the BTC amount transferred.

Yep, this was a very early idea for LN, but we just ended up doing the trimmed dust route instead. The bi-directional HTLC trick _does_ require one side to have some funds to begin with, but the `push_amt` feature does exist, so that could be utilized. Mechanically, I think we'd want to add a new special rule where we insert that HTLC in the other direction automatically (special evaluation rule).

GeorgeTsagk:
> add an inbound fee discount of the incoming channel w/ a base fee the same as the asset carrier HTLC amt

I don't think this works well. It assumes there's some symmetry in inbound/outbound traffic, and if that was the case (assuming that remote side also uses non-dust htlcs) then a fix would not even be needed as the sats would auto-balance

>  I think we'd want to add a new special rule where we insert that HTLC in the other direction automatically (special evaluation rule).

given that the channel lifecycle is totally LND's job, then that would have to be exposed via LND API?
IIUC we want a special case where if we want to send just an above dust amount to the other side, the commitment will also add an incoming htlc of the same amt.

[...]

PR #1252, opened 2024-12-12
[custom channels]: enforce minimum amounts

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1240.

This PR introduces new minimum transportable amounts (asset units for invoice creation/buy quotes and milli-satoshi for sending payments/sell quotes) and then enforces them in the corresponding Taproot Asset Channel RPCs.

Because attempting to create an invoice for a too low amount would result in a payment that cannot be forwarded by the edge node (`FeeInsufficient` error on forwarding, see Godoc comment on `rfqmath.MinTransportableUnits` function), creating such invoices is prevented.
On the other side, when attempting to pay a small invoice with assets, the payment is rejected. But because it _can_ succeed, the user has the option to override and send the payment anyway, even if that would be uneconomical.

For reference, here's the Godoc (and code) of `rfqmath.MinTransportableUnits` that explains the `FeeInsufficient` problem:

```go
// MinTransportableUnits computes the minimum number of transportable units
// of an asset given its asset rate and the constant HTLC dust limit. This
// function can be used to enforce a minimum invoice amount to prevent
// forwarding failures due to invalid fees.
//
// Given a wallet end user A, an edge node B, an asset rate of 100 milli-
// satoshi per asset unit and a flat 0.1% routing fee (to simplify the
// scenario), the following invoice based receive events can occur:
//  1. Success case: User A creates an invoice over 5,000 units (500,000 milli-
//     satoshis) that is paid by the network. An HTLC over 500,500 milli-satoshis
//     arrives at B. B converts the HTLC to 5,000 units and sends 354,000 milli-
//     satoshis to A.
//     A receives a total "worth" of 854,000 milli-satoshis, which is already
//     more than the invoice amount. But at least the forwarding rule in `lnd`
//     for B is not violated (outgoing amount mSat < incoming amount mSat).
//  2. Failure case: User A creates an invoice over 3,530 units (353,000 milli-
//     satoshis) that is paid by the network. An HTLC over 353,530 milli-satoshis
//     arrives at B. B converts the HTLC to 3,530 units and sends 354,000 milli-
//     satoshis to A.
//     This fails in the `lnd` forwarding logic, because the outgoing amount
//     (354,000 milli-satoshis) is greater than the incoming amount (353,530
//     milli-satoshis).
func MinTransportableUnits(dustLimit lnwire.MilliSatoshi,
	rate BigIntFixedPoint) BigIntFixedPoint {

	// We can only transport an asset unit equivalent amount that's greater
	// than the dust limit for an HTLC, since we'll always want an HTLC that
	// carries an HTLC to be reflected in an on-chain output.
	units := MilliSatoshiToUnits(dustLimit, rate)

	// If the asset's rate is such that a single unit represents more than
	// the dust limit in satoshi, then the above calculation will come out
	// as 0. But we can't transport zero units, so we'll set the minimum to
	// one unit.
	if units.ScaleTo(0).ToUint64() == 0 {
		units = NewBigIntFixedPoint(1, 0)
	}

	return units
}
```

Just for clarity: This enforcement will no longer be necessary once we address https://github.com/lightninglabs/taproot-assets/issues/888 with the symmetric HTLC construction.


guggero:
Taking this out of draft, since `litd` itests pass in the updated branch (PR to follow shortly).
The `litd` itests here are expected to fail, since we need some compile time fixes as well as new assertions.

GeorgeTsagk (commented):
Looks good

Don't we also want to enforce a min-amt when creating an invoice?

Current commits seem to only protect the payer, would be useful to know that my invoice is uneconomical when I create it too

guggero:
> Looks good
> 
> Don't we also want to enforce a min-amt when creating an invoice?
> 
> Current commits seem to only protect the payer, would be useful to know that my invoice is uneconomical when I create it too

Thanks for the review. Isn't that what https://github.com/lightninglabs/taproot-assets/pull/1252/commits/9abedabde35756df171ec4fc7a0803d0864a4569 does?
Or do you mean that the user should be able to override there as well? That won't be possible since the payment will never make it through the network.

guggero:
`litd` itest is here: https://github.com/lightninglabs/lightning-terminal/pull/921.

GeorgeTsagk:
> Thanks for the review. Isn't that what https://github.com/lightninglabs/taproot-assets/commit/9abedabde35756df171ec4fc7a0803d0864a4569 does?

Oh, probs missed it. :+1:

GeorgeTsagk (approved):
🏌️ ⛳

ZZiigguurraatt:
Generally, I'm wondering if the root cause of this should be fixed instead. Pretty soon, 354 sat is going to buy a soda or candy bar. How can we not make small payments? Seems like a real limitation. There is discussion of symmetric/bidirectional HTLC in https://github.com/lightninglabs/taproot-assets/issues/888 . Why introduce this this additional code complexity when it isn't going to even be needed when #888 is fixed?

rfqmath/convert.go
```diff
@@ -4,9 +4,27 @@ import (
...
+	DefaultOnChainHtlcAmount = lnwallet.DustLimitForSize(
+		input.UnknownWitnessSize,
+	)
+
+	// DefaultOnChainHtlcAmountMSat is the default amount that we consider
+	// as the smallest HTLC amount that can be sent on-chain in
+	// milli-satoshis.
+	DefaultOnChainHtlcAmountMSat = lnwire.NewMSatFromSatoshis(
+		DefaultOnChainHtlcAmount,
+	)
```
ffranr, line 25: maybe slightly shorter name `DefaultOnChainHtlcMSat`

taprpc/rfqrpc/rfq.proto
```diff
@@ -236,12 +236,15 @@ message PeerAcceptedBuyQuote {
...
+    // should be made.
     uint64 scid = 3;
 
-    // asset_amount is the amount of the subject asset.
-    uint64 asset_amount = 4;
+    // The maximum exchange amount denoted in the subject asset. This includes
+    // the user-configured maximum routing fees, so the actual payment amount
+    // will be less than this. This just defines the maximum volume that the
+    // edge node has accepted to divest with the given rate.
+    uint64 asset_max_amount = 4;
```
ffranr, line 247: :+1:

ffranr (commented):
Looks good so far. I still have a little more to review...

Roasbeef:
Nice work tracking down this issue! 

Thought about it a bit, and one issue I see with this solution is that even once we remove the behavior where we push the 354 sat HTLC across each time, this issue (trying to pay an invoice < 354 sats) still exists. Due to the NgU technology, 354 sats == ~35 cents, meaning that users won't be able to send amounts below that value with this change. Assuming NgU continues, the smallest invoice a user can pay will only grow. 

I think an alternative way to resolve this, would be to modify the `link.CheckHtlcForward` method. IIUC, we're failing here: https://github.com/lightningnetwork/lnd/blob/bb9c680a48cd1075d793cfc97b85f3676b2812c2/htlcswitch/link.go#L3267-L3288. Referring to @guggero's example above, we fail as the incoming amount is less than the outgoing amt, once we swap in the carrier HTLC amount of 354 sats. 

This fails as we're still expecting the normal BTC routing fee to be paid on the last hop. If we relax this for HTLCs w/ the proper set of custom records (adding stricter checks/parsing), then we'd enable smaller invoices to be paid. This isn't without a tradeoff though, as naively the last hop will still lose money (eg: 1 sat incoming, but send 354 out, a loss of 353 sats). One way to side step this would be to _also_ pay that last forwarding node a portion of the routing fee in the asset unit. Unfortunately, this doesn't work when the receiver doesn't yet have any asset units to give to the other side. 

TL;DR: NgU makes dust actually worth something, which hinders the smallest invoice that can be paid until the dust policy is abolished...

rfqmath/convert.go
```diff
@@ -156,3 +156,20 @@ func MinTransportableUnits(dustLimit lnwire.MilliSatoshi,
...
+// be represented in a Lightning Network payment when transferring an asset,
+// given the asset rate and the constant HTLC dust limit. This function can be
+// used to enforce a minimum payable amount with assets, as any invoice amount
+// below this value would be uneconomical as the total amount sent would exceed
+// the total invoice amount.
+func MinTransportableMSat(dustLimit lnwire.MilliSatoshi,
+	rate BigIntFixedPoint) lnwire.MilliSatoshi {
+
+	// We can only transport at least one asset unit in an HTLC. And we
+	// always have to send out an HTLC with a BTC amount of 354 satoshi. So
```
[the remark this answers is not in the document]
guggero: Yeah, this calculation only applies until we add the symmetric HTLC trick. Then we can delete both of these functions (I think).

guggero:
> There is discussion of symmetric/bidirectional HTLC in https://github.com/lightninglabs/taproot-assets/issues/888 . Why introduce this this additional code complexity when it isn't going to even be needed when https://github.com/lightninglabs/taproot-assets/issues/888 is fixed?

Because without #888 things will fail in unexpected ways that are hard to understand for the user when small amounts are involved. So instead of running into weird timeouts or `no_route` errors, we disallow payments that we _know_ are going to fail with the current situation and warn the user about situations where their expectations might be invalidated (e.g. sending more sats out than the invoice wanted).

Waiting for #888 isn't really an option for RC3 since that is quite complicated and involved to implement.

rpcserver.go
```diff
@@ -7074,8 +7074,24 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
 
+		// paymentMaxAmt is the maximum amount that the counterparty is
+		// expected to pay. This is the amount that the invoice is
+		// asking for plus the fee limit in milli-satoshis.
+		paymentMaxAmt, _, err := r.parseRequest(pReq)
+		if err != nil {
+			return err
+		}
+
+		// Check if the payment is economical based on the quote.
```
ffranr, line 7085: "Payment is economical" doesn’t seem quite right here. We’re not evaluating whether the payment is a good value or fair price relative to what’s received. Instead, we’re specifically checking for overpayment—meaning the total payment value exceeds the invoice amount.

rpcserver.go
```diff
@@ -7320,6 +7344,55 @@ func (r *rpcServer) parseRequest(
...
 
+// checkEconomicalPayment checks if paying a certain invoice amount is
+// economical to pay using assets, given the rate from the accepted quote and
+// the minimum non-dust HTLC amount dictated by the protocol.
+func checkEconomicalPayment(quote *rfqrpc.PeerAcceptedSellQuote,
+	paymentAmount lnwire.MilliSatoshi, allowOverpay bool) error {
+
+	rateFP, err := rfqrpc.UnmarshalFixedPoint(quote.BidAssetRate)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling asset rate: %w", err)
```
ffranr, line 7355: don't need word "error" here.

ffranr (approved):

Issue #1250, opened 2024-12-12
[bug]: Unable to fundchannel - message payload is too large 


vanditshah99:
### Background

We are trying to create a channel from speed4 to speed3, using fundchannel command, receiving the following error
```
[litcli] error funding channel: rpc error: code = Unknown desc = error funding channel: unable to send input ownership proofs: unable to send proof to peer: rpc error: code = Unknown desc = message payload is too large - encoded 66255 bytes, but maximum message payload is 65533 bytes
```

### Your environment
Mainnet

```sh
tapcli getinfo
{
    "version": "0.5.0-alpha.rc2 commit=v0.5.0-rc2",
    "lnd_version": "0.18.4-beta.rc2",
    "network": "mainnet",
    "lnd_identity_pubkey": "038dbb70e3484da0bed806156f5270d39b480a659c82fb4b56ba3dcbe0edd74d55",
    "node_alias": "speed4",
    "block_height": 874391,
    "block_hash": "000000000000000000005d8175570bdf70133d57b2c7820a1445082a9c53eb39",
   
}
lncli getinfo
67503f3564b2:/src# lncli getinfo
{
    "version":  "0.18.4-beta.rc2 commit=lightning-terminal-v0.14.0-alpha.rc2",
    "commit_hash":  "890764a28af254243c64856c4addf6b9dc674716",
    "identity_pubkey":  "038dbb70e3484da0bed806156f5270d39b480a659c82fb4b56ba3dcbe0edd74d55",
    "alias":  "speed4",
    "color":  "#f2f2f2",
    "num_pending_channels":  0,
    "num_active_channels":  1,
    "num_inactive_channels":  0,
    "num_peers":  3,
    "block_height":  874391,
    "block_hash":  "000000000000000000005d8175570bdf70133d57b2c7820a1445082a9c53eb39",
    "best_header_timestamp":  "1734005631",
    "synced_to_chain":  true,
    "synced_to_graph":  true,
    "testnet":  false,
    "chains":  [
        {
            "chain":  "bitcoin",
            "network":  "mainnet"
        }
    ],
    "uris":  [
        "038dbb70e3484da0bed806156f5270d39b480a659c82fb4b56ba3dcbe0edd74d55@44.207.210.37:8737"
    ],
    "features":  {
        "0":  {
            "name":  "data-loss-protect",
            "is_required":  true,
            "is_known":  true
        },
        "5":  {
            "name":  "upfront-shutdown-script",
[...]

PR #1259, opened 2024-12-17
tapchannel: send+recv chunks of the input proof to stay under max msg limit

Roasbeef:
In this PR, we start to send+recv chunks of the input proofs. This
ensures that if a suffix proof is larger than the `lnwire` message size,
then we'll be able to still send+recv it

Fixes https://github.com/lightninglabs/taproot-assets/issues/1250


tapchannelmsg/wire_msgs.go
```diff
@@ -237,6 +319,8 @@ type AssetFundingCreated struct {
 	// funding output to be able to create the aux funding+commitment
 	// blobs.
 	FundingOutputs tlv.RecordT[tlv.TlvType1, AssetOutputListRecord]
+
+	// TODO(roasbeef): need to chunk this??
```
guggero, line 459: I would say no, or not yet. We currently only allow 3 asset IDs to be committed to a channel (`maxNumAssetIDs`). And because the proofs for this output should be quite small (only one on-chain funding output and a change output), and only proof suffixes, I doubt we'd get up to 64k.

tapchannel/aux_funding_controller.go
```diff
@@ -1295,9 +1348,27 @@ func (f *FundingController) processFundingMsg(ctx context.Context,
...
+		// have all the chunks.
+		finalProof, err := assetFunding.addInputProofChunk(
+			assetProof.ProofChunk.Val,
+		)
+		if err != nil {
+			return tempPID, fmt.Errorf("unable to add input proof "+
+				"chunk: %w", err)
+		}
+
+		// If there's no final proof yet, we can just return>
```
guggero, line 1362: nit: comment artifact.

tapchannel/aux_funding_controller.go
```diff
@@ -461,6 +467,34 @@ func (p *pendingAssetFunding) addToFundingCommitment(a *asset.Asset) error {
...
+// returned.
+func (p *pendingAssetFunding) addInputProofChunk(chunk cmsg.ProofChunk,
+) (*proof.Proof, error) {
+
+	// Collect this proof chunk with the rest of the proofs.
+	chunkID := chunk.ChunkSumID.Val
+
+	proofChunks := p.inputProofChunks[chunkID]
+	proofChunks = append(proofChunks, chunk)
+	p.inputProofChunks[chunkID] = proofChunks
```
guggero, line 481: Can just be `p.inputProofChunks[chunkID] = append(p.inputProofChunks[chunkID], chunk)`?

tapchannel/aux_funding_controller.go
```diff
@@ -461,6 +467,34 @@ func (p *pendingAssetFunding) addToFundingCommitment(a *asset.Asset) error {
...
+
+	// If this isn't the last chunk, then we can just return nil and exit
+	if !chunk.Last.Val {
+		return nil, nil
```
guggero, line 485: Kind of looks like a use case for an option as a return value?

tapchannelmsg/wire_msgs.go
```diff
@@ -138,6 +160,122 @@ func NewProofChunk(sum [32]byte, chunk []byte, last bool) ProofChunk {
...
+		}
+	}
+
+	// With the digest sum validated, we'll now concatenate all the chunks
+	// together to obtain the full proof.
+	var proofBuf bytes.Buffer
+	for _, chunk := range chunks {
+		_, err := proofBuf.Write(chunk.Chunk.Val)
+		if err != nil {
+			return nil, fmt.Errorf("unable to write chunk: %v", err)
```
guggero, line 256: nit: linter wants `%w` for errors, here and below.

guggero (commented):
Very nice! Thanks for the quick bug fix.

What's the backward compatibility here? I assume channel funding would just result in an error (or timeout) if an rc3 client attempts to open with an rc2 one?
Sounds good to me, just thinking aloud.

Roasbeef:
> What's the backward compatibility here? I assume channel funding would just result in an error (or timeout) if an rc3 client attempts to open with an rc2 one?

Yeah if rc2 opens with rc3, there'd be an error/timeout as the verifier now relies on reading that `Last` bool to decide if it should combine the chunks and decode into the entire proof. If we want to, we can add more backwards compat if we assume a chunk size. So with this approach, if the chunk we get is below the chunk size, then we assume that it's the entire proof as proceed as normal.

tapchannel/aux_funding_controller.go
Roasbeef: Need to reference `proofChunks` below, so ends up being the same amount of lines.

guggero:
Btw, the `litd` itest failure seems to be a flake, verified locally that all itests still pass with this.

Roasbeef:
Pushed up a new version.

GeorgeTsagk (commented):
looks good!

GeorgeTsagk:
> Btw, the `litd` itest failure seems to be a flake, verified locally that all itests still pass with this.

Should we add itest coverage including an input proof that is > 60k bytes?

guggero (approved):
LGTM pending other review comments :tada:

GeorgeTsagk (approved):
良さそうだ ✅

PR #1263, opened 2024-12-19
asset: change all-zero hash to empty leaf

guggero:
Changes the behavior of the group key reveal v1.


GeorgeTsagk:
the legacy group key reveal is V0, therefore there's nothing breaking with this PR, correct?

ffranr (approved):
Nice! Now the user can always specify all leaf scripts.

ffranr:
> the legacy group key reveal is V0, therefore there's nothing breaking with this PR, correct?

Yes, exactly.

PR #1272, opened 2025-01-03
Part 1: Asset Minting with V1 Asset Group Key and Chantools Cold Storage Support

ffranr:
This PR builds on Oli's WIP to demonstrate asset minting using the new V1 asset group key formulation. It integrates `chantools` to mint assets with an external asset group signing key (taproot internal key), offering a viable option for users who wish to keep their asset group signing key in cold storage.

A new integration test is included to simulate user behavior by leveraging `chantools` via the command line. The test verifies that the minter can successfully sign an asset into the asset group using the external asset group signing key.

---

### Notes for Reviewers

Work which will be included in a future PR(s), ordered by most important first:
* Tweaked NUMS key asset group key formulation (https://github.com/lightninglabs/taproot-assets/pull/1290#pullrequestreview-2549690279)
* Doc TODO (https://github.com/lightninglabs/taproot-assets/pull/1295)
```
TODO(guggero/ffranr): Replace this step by allowing the CLI to take the signed
PSBT instead. We can list the batch to get the asset ID of each pending asset
and match the PSBT's input previous output to find out what signed PSBT belongs
to what asset (in case there are multiple).
```
* itest: ensure minter can spend asset minted using external key.
* Unit tests for `ExternalKey` type and methods (draft: https://github.com/lightninglabs/taproot-assets/pull/1284)
* Unit tests for method `PendingAssetGroup.PSBT`.
* Add unit tests for CLI arg parsing before calling into RPC.


ffranr:
I think these are the changes necessary. This PR can't really be made smaller in any meaningful way.

ffranr:
As mentioned during our last call, I’ve updated `taproot-assets/tapdb/sqlc/migrations/000026_asset_group_version_customsubtree.down.sql` to include the recreation of the previous version of `key_group_info_view` as part of the down migration. While we’re not actively using down migrations, it’s still beneficial to have this in place as a precaution.

cmd/tapcli/assets.go
```diff
@@ -138,6 +139,21 @@ var mintAssetCommand = cli.Command{
 				"in order to avoid printing a large amount " +
 				"of data in case of large batches",
 		},
+		cli.StringFlag{
+			Name: "group_key_xpub",
```
Roasbeef, line 143: Do we also need to be able to specify optional leaves for the custom tapscript root?

cmd/tapcli/assets.go
```diff
@@ -425,9 +473,37 @@ func sealBatch(ctx *cli.Context) error {
 	client, cleanUp := getMintClient(ctx)
 	defer cleanUp()
 
-	resp, err := client.SealBatch(ctxc, &mintrpc.SealBatchRequest{
+	req := &mintrpc.SealBatchRequest{
 		ShortResponse: ctx.Bool(shortResponseName),
-	})
+	}
+
+	// TODO(guggero): Actually just ask for the signed PSBT back and extract
```
Roasbeef, line 480: Will this be addressed in this PR? 

So accepting the signed PSBT to extract the signatures manually ourselves.

Roasbeef: One other related thing is that we want to add some basic meta data to the group key minting PSBT to allow external software to display information such as the asset ID, name, and units to be minted.

taprpc/mintrpc/mint.proto
```diff
@@ -176,9 +176,16 @@ message MintAsset {
...
 
     /*
-    The optional root of a tapscript tree that will be used when constructing a
-    new asset group key. This enables future issuance authorized with a script
-    witness.
+    An optional root of a Tapscript tree used when constructing a new asset
+    group key. This allows for future asset issuance authorized using a
+    script witness.
+
+    If an external group key is provided, the V1 scheme for group key script
```
Roasbeef, line 186: Why this assumption vs a version + default?

asset/asset.go
```diff
@@ -946,13 +946,23 @@ type GroupKey struct {
...
+	//
+	// In the V1 group key construction, this root is never empty. It always
+	// includes two layers of script leaves that commit to the group
+	// anchor's (genesis) asset ID, ensuring any user-provided Tapscript
+	// root is positioned at level 2.
 	TapscriptRoot []byte
 
+	// CustomTapscriptRoot is an optional tapscript root to graft at the
+	// second level of the tapscript tree, if specified.
+	CustomTapscriptRoot fn.Option[chainhash.Hash]
```
Roasbeef, line 1025: Perhaps we should combine these two fields into a new encapsulated struct? This way we hide some of the details, and callers only need to interact with a single field instead of two. 

In particular, for the v1 case, IMO the `TapscriptRoot` field should be private and generated as needed.

tapgarden/planter.go
```diff
@@ -745,15 +746,75 @@ func buildGroupReqs(genesisPoint wire.OutPoint, assetOutputIndex uint32,
 		// already be specified. Use that to derive the key group
 		// signature along with the tweaked key group.
 		if seedling.EnableEmission {
-			if seedling.GroupInternalKey == nil {
+			if seedling.GroupInternalKey == nil &&
```
Roasbeef, line 812: We can cut down on the indentation and line length a bit here by moving this into a new function. This new function can also be more easily tested to isolated the new functionality.

asset/asset.go
```diff
@@ -92,6 +92,30 @@ const (
 	V1 Version = 1
 )
 
+// GroupKeyVersion denotes the version of the group key construction.
+type GroupKeyVersion uint8
```
Roasbeef, line 96: Why isn't the version we added to just the group key reveal sufficient?

asset/asset.go
```diff
@@ -2017,6 +2055,25 @@ func (req *GroupKeyRequest) Validate() error {
 			sha256.Size)
 	}
 
+	// Version 1 specific checks.
+	if req.Version == GroupKeyV1 {
+		tapscriptRoot, err := chainhash.NewHash(req.TapscriptRoot)
```
Roasbeef, line 2122: Ah ok, it's a matter of if we'er making our tree grafting explicit or implicit...

Still gathering context, but would lean towards making it implicit (so not modifying/mutating the script tree root in place). This way it's unambiguous if a user provided a custom root or not (it'll be blank if not), and we don't need to lug around that extra field to determine if we _actually_ have a custom root or not.

asset/asset.go
```diff
@@ -2077,35 +2111,74 @@ func (req *GroupKeyRequest) Validate() error {
 	return nil
 }
 
+// NewGroupPubKey derives a group key for the asset group based on the group key
+// request and the genesis asset ID.
+func (req *GroupKeyRequest) NewGroupPubKey(genesisAssetID ID) (btcec.PublicKey,
+	error) {
+
+	// If the external key is not specified, we will construct a version 0
```
Roasbeef, line 2147: Or do we just want to center on a new default version?

tapgarden/planter.go
```diff
@@ -141,6 +144,66 @@ type PendingAssetGroup struct {
...
+func (p *PendingAssetGroup) PSBT(
+	params chaincfg.Params) (*psbt.Packet, error) {
+
+	// Generate PSBT equivalent of the group virtual tx.
+	packet, err := psbt.NewFromUnsignedTx(&p.GroupVirtualTx.Tx)
+	if err != nil {
+		return nil, fmt.Errorf("error producing group virtual PSBT "+
+			"from tx: %w", err)
+	}
+
```
Roasbeef, line 158: Re my other comment, this is where we'd add the new custom global fields to allow external signers/coordinates to display meta data w.r.t what it actually being minted. We can also provide the information needed to derive the virtual tx in the first place (for additional validation).

taprpc/mintrpc/mint.proto
```diff
@@ -124,6 +124,9 @@ message UnsealedAsset {
 
     // The group virtual transaction for the asset.
     taprpc.GroupVirtualTx group_virtual_tx = 3;
+
+    // The byte serialized PSBT equivalent of the group virtual transaction.
+    string group_virtual_psbt = 4;
```
Roasbeef, line 129: `SealBatch` can also optionally accept this.

tapdb/sqlc/migrations/000026_asset_group_version_customsubtree.up.sql
```diff
@@ -0,0 +1,31 @@
+-- Add version field to table asset_groups.
+ALTER TABLE asset_groups ADD COLUMN version INTEGER NOT NULL DEFAULT 0;
+
+-- Add custom_subtree_root field to the table asset_groups. This optional field
+-- is used to store the root of the custom subtree for the asset group. The
+-- custom subtree, if provided, represents a subtree of the final tapscript
+-- tree.
+ALTER TABLE asset_groups ADD COLUMN custom_subtree_root BLOB;
```
Roasbeef, line 8: Why isn't the existing field sufficient? 

https://github.com/lightninglabs/taproot-assets/blob/a00fd22d4e009dc83cec5a093d91e0db1f65ef50/tapdb/sqlc/migrations/000002_assets.up.sql#L95

Roasbeef: I think we only need the version here added for this migration.

tapdb/assets_common.go
```diff
@@ -629,11 +633,28 @@ func parseGroupKeyInfo(tweakedKey, rawKey, witness, tapscriptRoot []byte,
...
-		RawKey:        groupRawKey,
-		GroupPubKey:   *tweakedGroupKey,
-		TapscriptRoot: tapscriptRoot,
-		Witness:       groupWitness,
+		Version:             groupKeyVersion,
+		RawKey:              groupRawKey,
+		GroupPubKey:         *tweakedGroupKey,
+		TapscriptRoot:       tapscriptRoot,
+		Witness:             groupWitness,
+		CustomTapscriptRoot: customSubtreeRootHash,
```
Roasbeef, line 657: Should expand assertions for db tests to make sure these new fields are read/write properly. 

Although I don't think we need the new custom tapscript root at all. We can get away with just one field, the existing one, instead of having to lug around two fields where one is a transformation of the other.

itest/mint_fund_seal_test.go
```diff
@@ -501,6 +505,188 @@ func testMintFundSealAssets(t *harnessTest) {
 	})
 }
 
+// testMintExternalGroupKeyChantools tests that we're able to mint an asset
+// using an external asset signing group key derived and managed by chantools.
+func testMintExternalGroupKeyChantools(t *harnessTest) {
```
Roasbeef, line 510: 😎

Roasbeef (changes requested):

asset/asset.go
```diff
@@ -1904,38 +2117,96 @@ func (req *GroupKeyRequest) Validate() error {
...
 
-	// Compute the tweaked group key and set it in the asset before
-	// creating the virtual minting transaction.
-	genesisTweak := req.AnchorGen.ID()
-	tweakedGroupKey, err := GroupPubKeyV0(
-		req.RawKey.PubKey, genesisTweak[:], req.TapscriptRoot,
-	)
+	// Construct an asset group pub key.
+	genesisAssetID := req.AnchorGen.ID()
+	groupPubKey, err := req.NewGroupPubKey(genesisAssetID)
```
gijswijs, line 2198: :+1: I don't mind these small commits. Your commit narrative is very clear to me.

tapdb/sqlc/queries/assets.sql
```diff
@@ -403,20 +406,24 @@ LIMIT 1;
...
 WHERE (
     key_group_info_view.gen_asset_id = @genesis_id
 );
 
 -- name: QueryAssets :many
 SELECT
-    assets.asset_id AS asset_primary_key, assets.genesis_id, version, spent,
+    assets.asset_id AS asset_primary_key,
+    assets.genesis_id, assets.version,
+    spent,
```
gijswijs, line 426: ultra-nit: What are our coding style guidelines re line breaks and such for SQL?

tapdb/sqlc/migrations/000026_asset_group_version_customsubtree.up.sql
```diff
@@ -0,0 +1,31 @@
...
+-- We're going to recreate key_group_info_view to include the new columns.
+-- Therefore, we need to drop the existing view.
+DROP VIEW IF EXISTS key_group_info_view;
+
+-- Recreate the key_group_info_view to include the new columns.
+--
+-- This view is used to perform a series of joins that allow us to extract
+-- the group key information, as well as the group sigs for the series of
+-- assets we care about. We obtain only the assets found in the batch
+-- above, with the WHERE query at the bottom.
```
gijswijs, line 19: This is copy pasta from the original migration file. The comment `assets found in the batch above` doesn't make sense here.

```suggestion
-- assets we care about. We obtain the assets with the WHERE query at the
-- bottom.
```

gijswijs (commented):
Great work. I have added some small comments throughout.

nit: the commit message of commit a993070 claims the commit populates the external key in the seedling instance, but I don't think that happens in that commit.

I would love to see a itest that tests the re-issuance of an asset with cold minting.

Would also like to see a unit test that tests both the V0 and the V1 flow re GroupTapscriptRoot and a custom tapscript root. Whatever approach we take regarding implicit or explicit flow @roasbeef suggested, we will end up with two separate flows that are currently not fully covered by unit tests.

tapdb/sqlc/queries/assets.sql
ffranr: Good question. I'm not sure. @guggero do we have a doc on this?

I would think that if we use `AS`, example `assets.asset_id AS asset_primary_key,` then that gets its own line. And otherwise we wrap at column 80.

So I've moved `spent,` to the line above.

taprpc/mintrpc/mint.proto
ffranr: This PR will handle that change: https://github.com/lightninglabs/taproot-assets/pull/1295

cmd/tapcli/assets.go
ffranr: > Will this be addressed in this PR?
> 
> So accepting the signed PSBT to extract the signatures manually ourselves.

It will be addressed here: https://github.com/lightninglabs/taproot-assets/pull/1295

ffranr: > One other related thing is that we want to add some basic meta data to the group key minting PSBT to allow external software to display information such as the asset ID, name, and units to be minted.

I've put up this issue: https://github.com/lightninglabs/taproot-assets/issues/1298

taprpc/mintrpc/mint.proto
```diff
@@ -201,6 +201,14 @@ message MintAsset {
...
     */
     uint32 decimal_display = 13;
+
+    /*
+    The external group key is an optional field that allows specifying an
+    external signing key for the group virtual transaction during minting.
+    This key enables signing operations to be performed externally, outside
+    the daemon.
+    */
+    taprpc.ExternalKey external_group_key = 14;
```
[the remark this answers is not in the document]
ffranr: I think, maybe no? Because you might want to still specify a group key that you've used in the past `group_key` but then also specify that you're going to sign externally (so setting `external_group_key`).

But I think that if you set `external_group_key` then you shouldn't need to set `group_key`.

And conversely, if we store the `external_group_key` data at the point of creating the group key. Then we shouldn't need to set `external_group_key` if `group_key` is specified.

cmd/tapcli/assets.go
ffranr: I would argue that we do eventually but not for the next release. So we could potentially save that for a later PR. What do you think?

tapdb/sqlc/migrations/000026_asset_group_version_customsubtree.up.sql
ffranr: We need to maintain the custom subtree root separately from the tapscript root because this table will later be used to construct the mint proof group key reveal, which requires specifying the custom subtree root.

ffranr: We need the `version` column because the group key reveal formalisation might differ in the future even if we still have a separate custom subtree root.

tapdb/assets_common.go
ffranr: > Although I don't think we need the new custom tapscript root at all. We can get away with just one field, the existing one, instead of having to lug around two fields where one is a transformation of the other.

I think we need the custom tapscript root column so that we can read from it later to construct the group key reveal for the minting proof.

tapgarden/planter.go
ffranr: Great idea!

I've spun this into a new issue: https://github.com/lightninglabs/taproot-assets/issues/1298
I did that because I would like to shape it a little and I don't think it's critical right now. But would be good to get it into the same release.

asset/asset.go
ffranr: Here's how I see it:

The upside of keeping the default on V0 is that it's well tested and shifting to V1 wouldn't win anything unless the minter requires external signing.

The downside of keeping the default on V0 is that the debugging/user troubleshooting space is bigger.

The other downside is that not all our itests will cover V1 like they do V0.

I think, if we had the time, we should default to V1. And update all itests and unit tests to V1. The problem is that we still need almost the same test set for V0 because we will still support V0 going forwards.

asset/asset.go
ffranr: I think you're right. I think we can re-use group key reveal version type here. Will update.

asset/asset.go
ffranr: Yes, that sounds like a great idea to me.

I'm trying to make the minimal changes necessary in this PR, so I would like to make the change you suggested in a new PR. Let me know if you'd like me to work another way.

asset/asset.go
ffranr: I think we have to carry around the extra custom root so that we can populate it in the group key reveal. Otherwise our peer won't be able to validate our group key.

asset/asset.go
```diff
@@ -883,6 +966,10 @@ type GroupKeyRequest struct {
 	// has been applied.
 	RawKey keychain.KeyDescriptor
 
+	// ExternalKey specifies a public key that, when provided, is used to
+	// externally sign the group virtual transaction outside of tapd.
+	ExternalKey fn.Option[ExternalKey]
```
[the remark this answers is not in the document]
ffranr: I think these fields should be combined. They should effectively hold the same data if `ExternalKey` is specified. With the exception that `RawKey` has the public key and `ExternalKey` has the xpub. Of course, it's also possible to derive the public key from `ExternalKey`.

Ah but I also notice that `RawKey`'s type `KeyLocator` assumes derivation path `m/ 1017'` where as we need `m/ 86'`. So maybe we need to re-think `RawKey`.

Something we can do for now is not specify `RawKey` if `ExternalKey` is specified. And point that out in the doc.

ffranr:
We're missing db unit tests for the new columns. I'll see if I can add them to this PR.

ffranr:
Problem with this PR:
```
CREATE TABLE IF NOT EXISTS asset_groups (
    group_id INTEGER PRIMARY KEY,

    tweaked_group_key BLOB UNIQUE NOT NULL CHECK(length(tweaked_group_key) = 33),
```

For group key V1, `tweaked_group_key` could be the same across minting events because asset group key is no longer tweaked with the asset ID. This means that a minter may fail to mint a second asset using the same external signing key (because the taproot internal key of the asset group key will be the same) and therefore `UNIQUE` constraint will error.

ffranr:
> Problem with this PR:
> 
> ```
> CREATE TABLE IF NOT EXISTS asset_groups (
>     group_id INTEGER PRIMARY KEY,
> 
>     tweaked_group_key BLOB UNIQUE NOT NULL CHECK(length(tweaked_group_key) = 33),
> ```
> 
> For group key V1, `tweaked_group_key` could be the same across minting events because asset group key is no longer tweaked with the asset ID. This means that a minter may fail to mint a second asset using the same external signing key (because the taproot internal key of the asset group key will be the same) and therefore `UNIQUE` constraint will error.
> 
> And the minter can't just use a different external signing key because the asset group key will be different.

I've extended the integration test to demonstrate that the same external key can be used to mint two different assets into the same asset group across two separate batches. Therefore, the potential database issue I mentioned earlier isn't actually a problem. I believe we're writing the final asset group key into that field, which remains unique for each mint.

ffranr:
I've improved the itest and i've added a new DB unit test for the new asset group fields.

itest/chantools_harness.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// workingDir is the chantools harness working directory.
+	workingDir string
+
+	// walletDbPath is the path to the wallet.db file created by chantools.
+	walletDbPath string
+}
+
+// NewChantoolsHarness creates a new instance of the ChantoolsHarness struct.
+func NewChantoolsHarness(t *testing.T) ChantoolsHarness {
```
guggero, line 50: Very cool! Awesome to have this tested automatically.
Makes me think we should probably expose these functions in `chantools` so we can use them directly from Go instead of needing to invoke the binary.

itest/mint_fund_seal_test.go
```diff
@@ -501,6 +501,91 @@ func testMintFundSealAssets(t *harnessTest) {
...
+	chantools := NewChantoolsHarness(t.t)
+
+	// Use chantools to create a new wallet and derive a new key.
+	chantools.CreateWallet(t.t)
```
guggero, line 510: :heart_eyes:

itest/utils.go
```diff
@@ -735,6 +739,182 @@ func ManualMintSimpleAsset(t *harnessTest, lndNode *node.HarnessNode,
...
+	)
+
+	return batchAssets
+}
+
+// DeserializeWitnessStack deserializes a serialized witness stack into a
+// [][]byte.
+//
+// TODO(ffranr): Do we already have this function somewhere else? Move to btcd?
+func DeserializeWitnessStack(serialized []byte) ([][]byte, error) {
```
guggero, line 852: Could probably use `asset.TxWitnessDecoder()` somehow.

guggero (approved):
Amazing work on this! Thank you so much for picking up my WIP work and cleaning it up this nicely :pray:

LGTM with just some minor comments :tada:

taprpc/taprootassets.proto
```diff
@@ -255,6 +255,35 @@ message GenesisInfo {
     uint32 output_index = 6;
 }
 
+/*
+This message represents an external key used for deriving and managing
+hierarchical deterministic (HD) wallet addresses according to BIP-86.
+ */
```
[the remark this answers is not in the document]
ffranr: My IDE does this automatically. I can disable if it's against our style guide. I think the point is to align the `*` vertically.

asset/asset.go
```diff
@@ -92,6 +92,30 @@ const (
...
+
+	// GroupKeyV1 is the version of the group key that uses a construction
+	// that is compatible with PSBT signing where the group anchor's asset
+	// ID is appended as a sibling to any user-provided tapscript tree.
+	GroupKeyV1 GroupKeyVersion = 1
+)
+
+// NewGroupKeyVersionFromInt32 creates a new GroupKeyVersion from an int32. This
+// function is useful for decoding GroupKeyVersions from the SQL database.
+func NewGroupKeyVersionFromInt32(v int32) (GroupKeyVersion, error) {
```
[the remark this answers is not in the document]
ffranr: Would you like to see `NewGroupKeyVersionInt32` or `NewGroupKeyVersion` or something else?

My idea is that we may need another `NewGroupKeyVersionFrom...` in the future so making it more specific now.

itest/chantools_harness.go
ffranr: While you were away, I attempted to expose those functions in `chantools`, but it turned out to be a significant change that interfered with our short-term priorities.

I agree that exporting that functionality from `chantools` is a good idea. However, one advantage of the command-line harness is that it more closely mimics typical user behavior.

itest/utils.go
ffranr: I will reformulate in https://github.com/lightninglabs/taproot-assets/pull/1295
where we will extract the witness stack from the signed PSBT in tapd.

taprpc/taprootassets.proto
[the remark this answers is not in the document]
guggero: Not a style guide per se, but it's formatted that way across all our proto files in all projects.

itest/chantools_harness.go
gijswijs: If we choose to import chantools as a library here we can don't need the chantools check in the  Makefile target anymore.

itest/mint_fund_seal_test.go
```diff
@@ -500,6 +501,91 @@ func testMintFundSealAssets(t *harnessTest) {
...
+	mintReq.Asset.ExternalGroupKey = externalGroupKey
+
+	assetReqs := []*mintrpc.MintAssetRequest{mintReq}
+
+	batchAssets := MintAssetExternalSigner(
+		t, t.tapd, assetReqs, signerCallback,
+	)
+	t.Logf("First batch asset: %v", batchAssets)
+
+	// Formulate a second asset mint request with the same external group
```
gijswijs, line 573: 🧡

itest/utils.go
```diff
@@ -735,6 +739,182 @@ func ManualMintSimpleAsset(t *harnessTest, lndNode *node.HarnessNode,
...
+				"element data: %w", err)
+		}
+		stack = append(stack, element)
+	}
+
+	return stack, nil
+}
+
+// readCompactSize reads a compact size integer from the given reader.
+func readCompactSize(r *bytes.Reader) (uint64, error) {
```
gijswijs, line 883: I think you can use `ReadVarInt` here. See https://pkg.go.dev/github.com/btcsuite/btcd/wire#ReadVarInt

gijswijs (approved):
This is a milestone! Really tremendous work.
LGTM :tada: only some final nits and remarks.

asset/asset.go
```diff
@@ -1040,7 +1207,7 @@ func NewGKRCustomSubtreeRootRecord(root *chainhash.Hash) tlv.Record {
 // hash may represent either a single leaf or the root hash of an entire
 // subtree.
 //
-// If `custom_root_hash` is not provided, it defaults to a bare `[OP_RETURN` as
+// If `custom_root_hash` is not provided, it defaults to a bare `[OP_RETURN]` as
```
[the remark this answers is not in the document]
guggero: Yes, I updated this comment in https://github.com/lightninglabs/taproot-assets/pull/1290 to reflect that (which actually adds the pedersen commitment).

tapdb/sqlc/queries/assets.sql
guggero: > I would think that if we use AS, example assets.asset_id AS asset_primary_key, then that gets its own line. And > otherwise we wrap at column 80.

> So I've moved spent, to the line above.

There's no style guide yet. But what you describe above matches what I normally do.

itest/chantools_harness.go
ffranr: Yes would be nice to not need another external binary. As long as the exported functions correspond to the CLI commands correctly.

taprpc/taprootassets.proto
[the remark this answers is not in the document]
ffranr: > Not a style guide per se, but it's formatted that way across all our proto files in all projects.

Sounds pretty ubiquitous!

taprpc/taprootassets.proto
```diff
@@ -285,7 +285,10 @@ message ExternalKey {
 }
 
 message GroupKeyRequest {
-    // The internal key for the asset group before any tweaks have been applied.
+    /*
+    The internal key for the asset group before any tweaks have been applied.
+    If this field is set then external_key must be empty, and vice versa.
+     */
```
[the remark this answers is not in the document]
ffranr: Ok, I think I've fixed every instance like this.

taprpc/taprootassets.proto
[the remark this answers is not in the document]
ffranr: I think I've managed to fix ever instance introduced by this PR.

asset/asset.go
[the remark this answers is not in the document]
guggero: I'd say `NewGroupKeyVersion`. Creating and validating a version from an int seems to be the standard. If we ever need another one, then we can use `NewGroupKeyVersionFrom...`.

tapgarden/planter.go
```diff
@@ -723,6 +734,10 @@ func buildGroupReqs(genesisPoint wire.OutPoint, assetOutputIndex uint32,
 
 			groupReqs = append(groupReqs, *groupReq)
 			genTXs = append(genTXs, *genTx)
+
+			// TODO(ffranr): Should we continue to the next seedling
```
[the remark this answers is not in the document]
ffranr: I'll come back to this later. I mean to refactor `buildGroupReqs`.

tapdb/sqlc/migrations/000026_asset_group_version_customsubtree.up.sql
```diff
@@ -0,0 +1,28 @@
+-- Add version field to table asset_groups.
+ALTER TABLE asset_groups ADD COLUMN version INTEGER NOT NULL DEFAULT 0;
+
+-- Add custom_subtree_root field to the table asset_groups. This optional field
+-- is used to store the root of the custom subtree for the asset group. The
+-- custom subtree, if provided, represents a subtree of the final tapscript
+-- tree.
+ALTER TABLE asset_groups ADD COLUMN custom_subtree_root BLOB;
```
Roasbeef, line 8: So we have this table which was created when we added initial support to specify a custom tapscript sibling for the Bitcoin UTXO that anchors a new minting batch: https://github.com/lightninglabs/taproot-assets/blob/b5eda453a41ade01dee7346fe5fb3f895afaa4dd/tapdb/sqlc/migrations/000016_tapscript_trees.up.sql#L6-L17

I think we have an opportunity here to prep for future changes where a user provides `tapd` with the tapscript leaf information, which we've learned is required to make a valid wallet descriptor to feed into hardware wallets. Specifically, we can modify `BLOB` here to instead point to the `tapscript_roots` primary key and store `custom_subtree_root` as a `root_hash` in the table. In the future, users can pass in the leaves of their tapscript tree, to be inserted at creation, or after the fact. 

We'll also need this leaf information to be able to generate control blocks for script path spends from the group key.

ffranr: Nice idea :+1: 

I think if `custom_subtree_root` is `None` we can set `root_hash` to all zeros. Then we can keep the `NOT NULL` on that row.

ffranr: I guess we can set `branch_only` to `FALSE` always for `custom_subtree_root`.

GeorgeTsagk (approved):
LGTM, +tACK :1st_place_medal: 
great work!

PR #1286, opened 2025-01-09
asset: add more test cases for ExternalKey.PubKey

starius:
Added more derivation paths and corresponding public keys, check in the test that the values returned by the method match to expected values. Also added tpub (testnet xpub).

To run "chantools derivekey" with xpubs, use PR https://github.com/lightninglabs/chantools/pull/179

### DIFF
--- a/asset/asset_test.go
+++ b/asset/asset_test.go
@@ -1309,11 +1309,23 @@ func TestExternalKeyPubKey(t *testing.T) {
 		return *xpub
 	}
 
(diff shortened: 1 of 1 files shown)


ffranr:
Thank you very much for this! I've cherry picked your commit into https://github.com/lightninglabs/taproot-assets/pull/1284 so that we can avoid the merge commit. The commit still has you as the author and so on.

PR #1290, opened 2025-01-10
multi: add new v2 version of GKR based on Pedersen commitments 

Roasbeef:
In this PR, we add a new internal package that can be used to create
and verify Pedersen commitments. We aim to make the API very simple to
use, while providing configuration that's useful for tests and a general
set of use cases.

We also add property based tests to achieve 100% test coverage of all
relevant invariants.

We then this that to add a new version of the new more flexible GKR type.
This version uses an `OP_CHECKSIG` with an un-spendable public key. We
derive such a public key by using the Pedersen commitment of the asset
ID, or blank message (for the dummy leaf case). As a Pedersen commitment
uses an aux generator H, that no one knows the private key to, it's
impossible to generate a signature using the commitment point.

An informal sketch of a proof by contradiction goes something like:
  * Starting with a Pedersen commitment, `P = m*G + r*H`.
  * Let's assume that we can actually sign with `P`. This means that
    there's some `x`, where `P = x*G` that we know.
  * If we re-write `x` in terms of our commitment, we have: `P = m*G + r*H`.
  * Wlog, we'll simplify by removing `r`, the blinding factor, so: `P =
    m*G + H`.
  * In terms of private values we have `x*G = m*G + H`
  * If we solve for `H`, then we have `H = (x-m)*G`
  * If this is a Pedersen commitment, then the private key of `H` must be
    unknown to all parties.
  * If we can sign with `P`, then this means that we know the value `(x-m)`.
  * However, since this is a Pedersen commitment, it's impossible to
    know the value `x`, which is a function of the private value `H` of
    the Pedersen commitment (`P= (x-m)*G`). This is a contradiction, as
    we see the that if we know `x-m`, then we also know `h`.


dstadulis:
>Wlog, we'll simplify by removing r, the blinding factor, so: P = m*G + H.
Does Wlog = WLOG?
aka "without loss of generality"

asset/asset.go
```diff
@@ -953,17 +954,62 @@ type GroupKeyReveal interface {
...
-	// of the script to ensure that the script is non-executable.
-	scriptBuilder := txscript.NewScriptBuilder().AddOp(txscript.OP_RETURN)
+// The script leaf is made non-spendable by including an OP_RETURN/OP_CHECKSIG
+// opcode at the start of the script. While the script can still be executed, it
+// will always fail and cannot be used to spend funds.
+func NewNonSpendableScriptLeaf(version GkrV1Version,
+	data []byte) (txscript.TapLeaf, error) {
+
+	var firstOp byte
+	switch version {
```
dstadulis, line 964: helpful switch to create flexibility

asset/asset.go
```diff
@@ -953,17 +954,62 @@ type GroupKeyReveal interface {
...
-// The script leaf is made non-spendable by including an OP_RETURN opcode at the
-// start of the script. While the script can still be executed, it will always
-// fail and cannot be used to spend funds.
-func NewNonSpendableScriptLeaf(data []byte) (txscript.TapLeaf, error) {
-	// Construct a script builder and add the OP_RETURN opcode to the start
-	// of the script to ensure that the script is non-executable.
-	scriptBuilder := txscript.NewScriptBuilder().AddOp(txscript.OP_RETURN)
+// The script leaf is made non-spendable by including an OP_RETURN/OP_CHECKSIG
+// opcode at the start of the script. While the script can still be executed, it
+// will always fail and cannot be used to spend funds.
```
dstadulis, line 959: Thinking abstractly, about failsafe methods to prevent any accidental insertion of a NUMS without the OP_CHECKSIG:

maybe: a mutex that's called obtained when NUMS generation occurs and then is only released after prefixing the key with OP_CHECKSIG?

asset/asset.go
```diff
@@ -953,17 +954,62 @@ type GroupKeyReveal interface {
...
+	default:
+		// TODO(roasbeef): auto accept of new version for forwards
+		// compat?
+		return txscript.TapLeaf{}, fmt.Errorf("unknown "+
+			"version %v", version)
+	}
+
+	// Construct a script builder and add the firstOp opcode to the start of
+	// the script to ensure that the script is non-executable.
+	scriptBuilder := txscript.NewScriptBuilder().AddOp(firstOp)
```
dstadulis, line 983: Sound defensive positioning

internal/pedersen/commitment.go
```diff
@@ -0,0 +1,151 @@
...
+	// means that the commitment loses the hiding attribute, but still
+	// remains computationally binding.
+	Mask fn.Option[[sha256.Size]byte]
+
+	// NUMs is an optional value that should be used to verify the
+	// commitment if a custom NUMs point was used.
+	NUMs fn.Option[btcec.PublicKey]
+}
+
+// Commitment is a Pederson commitment of the form: m*G + r*H, where:
```
dstadulis, line 43:
```suggestion
// Commitment is a Pedersen commitment of the form: m*G + r*H, where:
```
branch is also labeled `son` rather than `sen`

internal/pedersen/commitment.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Mask is the mask used to blind the message. This is typically
+	// referred to as `r` in the Pedersen commitment literature.
+	//
+	// We make this optional, as without it we'll default to no value, which
+	// means that the commitment loses the hiding attribute, but still
+	// remains computationally binding.
+	Mask fn.Option[[sha256.Size]byte]
+
+	// NUMs is an optional value that should be used to verify the
```
dstadulis, line 38:
```suggestion
	// NUMS is an optional value that should be used to verify the
```

internal/pedersen/commitment.go
```diff
@@ -0,0 +1,151 @@
...
+	return &commitOpts{}
+}
+
+// commitOpt is a functional option that can be used to modify the default set
+// of options.
+type commitOpt func(*commitOpts)
+
+// WithCustomNUMs is a functional option that can be used to set a custom NUMs
+// point.
+func WithCustomNUMs(h btcec.PublicKey) commitOpt {
```
dstadulis, line 74: WRT NUMS style is traditional acronym style preferred of capitalizing every letter of the acronym?

asset/asset.go
```diff
@@ -953,17 +954,62 @@ type GroupKeyReveal interface {
...
+
+	switch version {
```
ffranr, line 985: I don't think we should compose the script in parts and switch on `version` twice. I think it adds code paths unnecessarily and makes it harder to see the final script which corresponds to any particular version. (The reader has to compose the script in their head.)

ffranr (changes requested):
I hadn't considered Pedersen commitments as a solution for our need before reviewing this PR. Really interesting subject to review/consider.

I think that we can re-use the existing cryptography functionality to achieve the same goal.

I think we can do:
```
	nonSpendableKey := input.TaprootNUMSKey
	if data != nil {
		dataHash := chainhash.HashH(data)

		tweakedNumsKey := input.TweakPubKeyWithTweak(
			&input.TaprootNUMSKey, dataHash[:],
		)
		nonSpendableKey = *tweakedNumsKey
	}

	scriptBuilder := txscript.NewScriptBuilder()
	scriptBuilder.AddData(schnorr.SerializePubKey(&nonSpendableKey))
	scriptBuilder.AddOp(txscript.OP_CHECKSIG)
	script, err := scriptBuilder.Script()
```

IMO we can re-use the existing function `input.TweakPubKeyWithTweak`. The function is short, so I'll quote it here with extra comments to show that it does the same thing in terms of the Pedersen commitment formula `P = m*G + r*H`.

```
func TweakPubKeyWithTweak(pubKey *btcec.PublicKey,
	tweakBytes []byte) *btcec.PublicKey {

	var (
		pubKeyJacobian btcec.JacobianPoint
		tweakJacobian  btcec.JacobianPoint
		resultJacobian btcec.JacobianPoint
	)
	// Form tweak scalar from tweak bytes. This is `m` in the equation
	// `P = m*G + r*H`.
	tweakKey, _ := btcec.PrivKeyFromBytes(tweakBytes)

	// Compute the Jacobian point for the tweak. This is `m*G`.
	// Crucially, the base point is the generator point G.
	btcec.ScalarBaseMultNonConst(&tweakKey.Key, &tweakJacobian)

	// At this point we can just assume `r` is 1, as we're only using `H` as
	// the NUMS point. The public key here is the NUMS key.
	pubKey.AsJacobian(&pubKeyJacobian)
	
	// Perform the addition: m*G + r*H
	// where r = 1 and H is the NUMS point.
	btcec.AddNonConst(&pubKeyJacobian, &tweakJacobian, &resultJacobian)

	// Return P.
	resultJacobian.ToAffine()
	return btcec.NewPublicKey(&resultJacobian.X, &resultJacobian.Y)
}
```

I think `P` will remain non-spendable for the reason you've outlined in the PR body. Specifically, the NUMS generator point cannot be expressed as a linear combination of the standard generator point `G`.

Since we won’t be utilizing the privacy feature of Pedersen commitments, I think we should reconsider using Pedersen commitments altogether. From what I understand after reading [Non-Interactive and Information-Theoretic Secure Verifiable Secret Sharing](https://link.springer.com/chapter/10.1007/3-540-46766-1_9), the privacy aspect seems to be a fundamental part of Pedersen commitments. Referring to Pedersen commitments while essentially creating a NUMS key commitment strikes me as unnecessarily complex.

asset/asset.go
guggero: I agree. Also, for the `CHECKSIG` operation, I think the order needs to be reversed (e.g. `<pubkey> OP_CHECKSIG`.

Here's the version I'm going to test with on bitbox2 (to see if they already support `pk()` tree elements):

```
	var builder *txscript.ScriptBuilder
	switch version {
	// For the OP_RETURN based version, we'll use a single OP_RETURN opcode.
	case OpReturnVersion:
		builder = txscript.NewScriptBuilder().AddOp(txscript.OP_RETURN)
		if data != nil {
			builder = builder.AddData(data)
		}

	// For the Pedersen commitment based version, we'll use a single
	// OP_CEHCKSIG with an un-spendable key.
	case PedersenVersion:
		var msg [sha256.Size]byte
		copy(msg[:], data)

		// Make a Pedersen opening that uses no mask (we don't carry on
		// the random value, as we don't care about hiding here). We'll
		// also use the existing NUMs point.
		op := pedersen.Opening{
			Msg: msg,
		}
		commitPoint := pedersen.NewCommitment(op).Point()

		commitBytes := schnorr.SerializePubKey(&commitPoint)

		builder = txscript.NewScriptBuilder().AddData(commitBytes).
			AddOp(txscript.OP_CHECKSIG)

	default:
		// TODO(roasbeef): auto accept of new version for forwards
		// compat?
		return txscript.TapLeaf{}, fmt.Errorf("unknown "+
			"version %v", version)
	}

	// Construct script from the script builder.
	script, err := builder.Script()
	if err != nil {
		return txscript.TapLeaf{}, fmt.Errorf("failed to construct "+
			"non-spendable script: %w", err)
	}
```

asset/asset.go
Roasbeef: The checksig is always added first, later in the body of this function.

Roasbeef:
> I think that we can re-use the existing cryptography functionality to achieve the same goal.

It's certainly similar, but that function was created for a different purpose, namely for the randomized per commitment point for a LN commitment. The goal here with the `internal` package was to make a higher level wrapper with a single purpose that's less error prone (eg: explicit `Opening` struct, ability to use non-randomized commitment, etc). 

> the privacy aspect seems to be a fundamental part of Pedersen commitments

A cryptographic commitment can be hiding (no information about the opening revealed from the commitment). It can also be binding (not possbile to commit to one value and prevent an opening of a different value. Both those properties can rely on either a computational or information theoretic assumption. 

Pedersen commitments are computationally binding (reduces to the DH problem), and information theoretically (perfect) hiding. 

The mini library I created allows the commitment to be optionally non-hiding. In our case, we don't really need the hiding property as when the commitment is presented, we already know what the opening is (the asset ID). 

> Referring to Pedersen commitments while essentially creating a NUMS key commitment strikes me as unnecessarily complex.

Not sure what you mean, that's the very construction of a Pedersen commitment. It relies on a aux generator `H` with an unknown discrete log. A NUMS key by definition has an unknown discrete log. This is just a precise implementation of the construct.

asset/asset.go
Roasbeef: Great suggestion. I think I tried a bit too hard to fit the new logic into the existing flow. 

I've also added a test here for both versions to ensure that they're actually unspendable.

ffranr:
> > Referring to Pedersen commitments while essentially creating a NUMS key commitment strikes me as unnecessarily complex.
> 
> Not sure what you mean, that's the very construction of a Pedersen commitment. It relies on a aux generator `H` with an unknown discrete log. A NUMS key by definition has an unknown discrete log. This is just a precise implementation of the construct.

What I mean is that a Pedersen commitment includes the separate component of the unknown discrete log, the blinding factor `r`, which IMO is an unnecessary complexity for our use case since we won't use it.

From my perspective, the only functional change introduced by this PR is the added support for `r`. Otherwise, `input.TweakPubKeyWithTweak` already computes exactly (see: https://github.com/lightninglabs/taproot-assets/commit/d6d7a7112593a683ac383d321c0d53909bf4b52e) what we need and seems sufficiently general for our purposes.

But I do like that the Pedersen commitment library formalises the `G` tweak term and NUMS point relationship where as with `input.TweakPubKeyWithTweak` it just seems more happenstance. `input.TweakPubKeyWithTweak` could have just assume that the public key had the generator point `G` for example.

asset/asset.go
```diff
@@ -953,21 +954,49 @@ type GroupKeyReveal interface {
...
+		var msg [sha256.Size]byte
+		copy(msg[:], data)
+
+		// Make a Pedersen opening that uses no mask (we don't carry on
+		// the random value, as we don't care about hiding here). We'll
+		// also use the existing NUMs point.
+		op := pedersen.Opening{
+			Msg: msg,
+		}
+		commitPoint := pedersen.NewCommitment(op).Point()
```
ffranr, line 984: I think this sets `r = 1` here. :+1: 

Which IMO is simpler than setting it to asset ID (which an earlier revision might have done?)

asset/asset.go
```diff
@@ -953,21 +954,49 @@ type GroupKeyReveal interface {
...
+			AddOp(txscript.OP_CHECKSIG)
 
-	// Add the data to the script if it is provided.
-	if data != nil {
-		scriptBuilder.AddData(data)
+	default:
+		// TODO(roasbeef): auto accept of new version for forwards
+		// compat?
+		return txscript.TapLeaf{}, fmt.Errorf("unknown "+
+			"version %v", version)
```
ffranr, line 995: We need to ensure that the correct leaf script is constructed for the given version. What does "auto accept" mean here? I think returning an error is the behaviour we want.

asset/asset.go
```diff
@@ -1072,7 +1101,12 @@ func NewGKRCustomSubtreeRootRecord(root *chainhash.Hash) tlv.Record {
 //   - [OP_RETURN] is a non-spendable script leaf containing the script
 //     `OP_RETURN`. Its presence ensures that [tweaked_custom_branch] remains
 //     a branch node and cannot be a valid genesis asset ID leaf.
+//
+// TODO(roasbeef): upgrade diagram w/ v1 and v2
```
ffranr, line 1105: In updating the diagram maybe we can just do:
```
//	                       [tapscript_root]
//	                         /          \
//	[non_spend(<genesis asset ID>)]   [tweaked_custom_branch]
//	                                        /        \
//                                    [non_spend()]    <custom_root_hash>
```

And we can describe `non_spend(...)` in the doc for `NewNonSpendableScriptLeaf`.

`non_spend` would be a nice (Taproot) Miniscript primitive type.

asset/asset.go
```diff
@@ -1182,12 +1219,29 @@ func (g *GroupKeyRevealTapscript) Validate(assetID ID) error {
 	return nil
 }
 
+// GkrV1Version is the version of the group key reveal.
+//
+// Version 1 is the original version that's based on an OP_RETURN.
+//
+// Version 2 is a follow up version that instead uses a Pedersen commitment.
+type GkrV1Version = uint8
```
ffranr, line 1227: The type name is confusing to me in two different ways here: it conflates with "version 1" of the group key reveal, and that there's `V1` in the name.

I think we should name this `NonSpendLeafVersion`. There's no need to tie this version type (or non-sepnd leaf fucntionality) directly to Group Key Reveal V1, as the non-spend leaf version is downstream of GKR V1. In fact, GKR depends on the non-spend leaf version, not the other way around.

ffranr (commented):
How do we feed the non-spend leaf version from the interface to this point? Do we need to serialise it into the proof group key reveal?

Alternatively, instead of adding a non-spend leaf version, we could, during group key reveal verification just attempt to reconstruct the tapscript root using both non-spend leaf versions. We just want one of them to validate, doesn't matter which.

So in the next release could have this functionality:
* Construct group keys with Pedersen commitment.
* Be able to verify group keys with either OP_RETURN or Pedersen commitment leaves. And just tries both verification methods without looking up any non-spend leaf version.

Roasbeef:
> How do we feed the non-spend leaf version from the interface to this point? Do we need to serialise it into the proof group key reveal?

So what I've done here is just inherit the value from the enclosing struct during decoding: https://github.com/lightninglabs/taproot-assets/pull/1290/files#diff-98586afaee9ef4f1d46ba2b4af7ce1fd31d28bc83e6df2117e6483532221f8a7R1388-R1391. They're the exact same version, which is only really used when we need to make the unspendable leaf.

asset/asset.go
Roasbeef: So since we don't specify the option of the masking/blinding value, `r` gets set to one during the option handling in `NewCommitment`.

asset/asset.go
Roasbeef: Yeah you're right, this was just a fleeting thought I had during implementation thinking about future/possible changes.

asset/asset.go
Roasbeef: Good idea, done. 

I had some trouble with the way `gofmt` seems to want to format comments. At least with my editor set up, I just keeps removing the tabs I add to create some indentation for the new section.

asset/asset.go
Roasbeef: SGTM, swapped to be `NonSpendLeafVersion`. It's just a `uint8` on the wire, so in the future, we can update this constant name if we need to.

guggero:
I've pushed up some fixes and additional commits for what we need for the hardware wallet support.

> Alternatively, instead of adding a non-spend leaf version, we could, during group key reveal verification just attempt to reconstruct the tapscript root using both non-spend leaf versions. We just want one of them to validate, doesn't matter which.

I think the way the version is encoded now I prefer the more explicit one. And since we already encoded the version anyway, we don't store more data in the proof.

guggero (approved):
Provided my review in form of direct patches.

asset/asset.go
```diff
@@ -1202,6 +1197,68 @@ func NewGKRCustomSubtreeRootRecord(root *chainhash.Hash) tlv.Record {
 	return tlv.MakePrimitiveRecord(GKRCustomSubtreeRoot, (*[32]byte)(root))
 }
 
+// NumsXPub turns the given NUMS key into an extended public key (using the x
+// coordinate of the public key as the chain code), then derives the actual key
+// to use from the derivation path 0/0. The extended key always has the mainnet
+// version, but can be converted to any network on demand by the caller with
+// CloneWithVersion().
+func NumsXPub(numsKey *btcec.PublicKey) (*hdkeychain.ExtendedKey,
```
ffranr, line 1205: nit: Accept arg as just `numsKey btcec.PublicKey`. Why risk `nil` here?

I think this is just one of those patterns, but I don't see the point of it.

asset/asset.go
```diff
@@ -1202,6 +1197,68 @@ func NewGKRCustomSubtreeRootRecord(root *chainhash.Hash) tlv.Record {
...
+
+	return extendedNumsKey, actualKey, nil
+}
+
+// TweakedNumsKey derives the NUMS key from the given data, then creates the
+// extended key from it and derives the actual (derived child) key to use from
+// the derivation path 0/0. The extended key always has the mainnet version, but
+// can be converted to any network on demand by the caller with
+// CloneWithVersion().
+func TweakedNumsKey(data []byte) (*hdkeychain.ExtendedKey, *btcec.PublicKey,
```
ffranr, line 1245: Arg should just be `data [sha256.Size]byte` IMO

asset/asset.go
```diff
@@ -1224,24 +1252,36 @@ func NewGKRCustomSubtreeRootRecord(root *chainhash.Hash) tlv.Record {
 //
 //	                       [tapscript_root]
 //	                         /          \
-//	[OP_RETURN <genesis asset ID>]   [tweaked_custom_branch]
+//	[non_spend(<genesis asset ID>)]   [tweaked_custom_branch]
```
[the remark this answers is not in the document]
ffranr: Also is the tree struct just the genesis asset ID node now for HW signing without custom spend subtree?

asset/asset.go
```diff
@@ -1122,21 +1124,47 @@ type GroupKeyReveal interface {
...
+		builder = txscript.NewScriptBuilder().AddOp(txscript.OP_RETURN)
+		if data != nil {
+			builder = builder.AddData(data)
+		}
+
+	// For the Pedersen commitment based version, we'll use a single
+	// OP_CEHCKSIG with an un-spendable key.
+	case PedersenVersion:
+		var msg [sha256.Size]byte
+		copy(msg[:], data)
```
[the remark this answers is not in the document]
guggero: Added a check to the last commit.

asset/asset.go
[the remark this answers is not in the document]
guggero: Fixed the references to `[OP_RETURN]`.

> Also is the tree struct just the genesis asset ID node now for HW signing without custom spend subtree?

If there are no custom scripts, then yes, there's only a single leaf with the `non_spend(<genesis asset ID>)` script. Updated some references in the comments.

asset/testdata/rapid/TestGroupKeyRevealEncodeDecodeRapid/TestGroupKeyRevealEncodeDecodeRapid-20250113154507-29836.fail
```diff
@@ -0,0 +1,222 @@
+# 2025/01/13 15:45:07.315685 [TestGroupKeyRevealEncodeDecodeRapid] [rapid] draw internal_key_bytes: []byte{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}
```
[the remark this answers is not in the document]
guggero: See comment above.

asset/group_key_reveal_test.go
```diff
@@ -273,3 +281,118 @@ func TestGroupKeyRevealEncodeDecodeRapid(tt *testing.T) {
...
+			prevOuts := txscript.NewCannedPrevOutputFetcher(
+				leafScript, amt,
+			)
+			sigHash := txscript.NewTxSigHashes(spendTx, prevOuts)
+
+			// If this is the Pedersen variant, then we'll actually
+			// need to generate a signature.
+			var sig []byte
+			if testCase.version == PedersenVersion {
+				privKey, _ := btcec.PrivKeyFromBytes(assetID[:])
```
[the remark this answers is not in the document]
guggero: Yes, that's how I understood it as well.

asset/asset.go
```diff
@@ -1316,26 +1316,15 @@ type GroupKeyRevealTapscript struct {
...
-
-	// customSubtreeInclusionProof provides the inclusion proof for the
-	// custom tapscript subtree. It is required to spend the custom
-	// tapscript leaves within the tree.
-	//
-	// NOTE: This field should not be serialized as part of the group key
-	// reveal. It is included here to ensure it is constructed concurrently
-	// with the tapscript root, maintaining consistency and minimizing
-	// errors.
-	customSubtreeInclusionProof []byte
```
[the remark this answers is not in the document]
guggero: But isn't this exactly what the `if !g.root.IsEqual(&tapscript.root) {` check is for?

gijswijs:
Really cool stuff, especially the mask being 1.

I have no remarks on the code.

I do have some problems with the proof sketch.

> * If we can sign with `P`, then this means that we know the value `(x-m)`.
> 
> * However, since this is a Pedersen commitment, it's impossible to
>   know the value `x`, which is a function of the private value `H` of
>   the Pedersen commitment (`P= (x-m)*G`). This is a contradiction, as
>   we see the that if we know `x-m`, then we also know `h`.

I find this part of the sketch confusing. The assumption that knowing how to sign with `P` directly implies knowledge of `x - m` is problematic, iiuc. If you can sign with `P`, it means you know the private key `x`, not `x-m`.

The contradiction arises not from knowing `x-m`, but from the assumption that `H` can be expressed as a known scalar multiple of `G`. In a Pedersen commitment, `H` is chosen such that its discrete log with respect to `G` is unknown and cannot be derived from the commitment. `P` can never be expressed in terms of `G`. Assuming that you can sign with `P` implies that you *can* do that, and that's where the contradiction arises.

Maybe I'm being too academic. It does not have any implications for the code, which looks sound to me.

ffranr (approved):
Nice!

Issue #1293, opened 2025-01-13
[bug]: Can't create address with asset from remote universe.


kallerosenbaum:
### Background

I have two litd instances running on regtest in kubernetes. I have minted an asset on node 0, and it's available in my local universe there. However, when I try to create an address on node 1, it fails with `[ERR] RPCS: [/taprpc.TaprootAssets/NewAddr]: unable to make new addr: unable to make address for unknown asset 431d08d06ad89fb719dd3bc37b84a08db9381e
1b840c9e74101404926fc44a45: asset lookup failed for asset: 431d08d06ad89fb719dd3bc37b84a08db9381e1b840c9e74101404926fc44a45`

Am I doing this wrong, or have I been hit by a bug?

### Your environment

I'm running litd docker image `lightninglabs/lightning-terminal:v0.14.0-alpha@sha256:1fec11bdf96a1c628f29b8f6f106decebc68d03b1bf240779d419290a003aa95` in a kubernetes environment.

bitcoind version 27,0

### Steps to reproduce

Node 0:
```
lnd-assets-regtest-0:/# tapcli --tlscertpath ~/.lit/tls.cert --rpcserver localhost:8443 --network=regtest assets mint --decimal_display 6 --meta_bytes '{"hello":true}' --type normal --name USDB --supply 1000000000000 --meta_type json --new_grouped_asset
lnd-assets-regtest-0:/# tapcli --tlscertpath ~/.lit/tls.cert --rpcserver localhost:8443 --network=regtest assets mint finalize
lnd-assets-regtest-0:/# tapcli --tlscertpath ~/.lit/tls.cert --rpcserver localhost:8443 --network=regtest universe roots
{
    "universe_roots": {
        "issuance-5855ac2d9761fb379b93a6a55b98ad86661ee96ad4de682b5c6dafb8dfafdd6f": {
            "id": {
                "group_key": "f1453960e1500763fb96de3458826240d59e386404463843219f4c9ba8c5b34f",
                "proof_type": "PROOF_TYPE_ISSUANCE"
            },
            "mssmt_root": {
                "root_hash": "fca80b63e8c8ef872c1e73dfd7f3b6c92a3d64b1aa9a7c6d34f59bb2f64b59f1",
                "root_sum": "1000000000000"
            },
            "asset_name": "USDB",
            "amounts_by_asset_id": {
[...]

PR #1294, opened 2025-01-13
universe: improve logging for single asset sync attempt

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1293.

When syncing a single asset (for example for creating an address) fails, the cause isn't easily visible in the logs. First, because the actual remote error wasn't logged and second because it was logged as debug. This commit fixes both.

### DIFF
--- a/universe/auto_syncer.go
+++ b/universe/auto_syncer.go
@@ -846,9 +846,14 @@ func (f *FederationEnvoy) SyncAssetInfo(ctx context.Context,
 		// Sync failures are expected from Universe servers that do not
 		// have a relevant universe root.
 		if err != nil {
-			log.Debugf("asset lookup for %v failed with remote"+
-				"server: %v", assetID.String(), addr.HostStr())
-			//lint:ignore nilerr failure is expected and logged
+			log.Warnf("Asset lookup failed: asset_id=%v, "+
+				"remote_server=%v: %v", assetID.String(),
+				addr.HostStr(), err)
+
+			// We don't want to abort syncing here, as this might
+			// just be one server in our list and returning an error
+			// would cause us to stop trying the other servers.
+			// lint:ignore nilerr failure is expected and logged.
 			return nil
 		}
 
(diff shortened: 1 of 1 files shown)


ffranr (approved):
Nice :+1:

I prefer all log params to go on the end of the message, but no big deal.

universe/auto_syncer.go
```diff
@@ -846,18 +846,30 @@ func (f *FederationEnvoy) SyncAssetInfo(ctx context.Context,
 		// Sync failures are expected from Universe servers that do not
 		// have a relevant universe root.
 		if err != nil {
-			log.Debugf("asset lookup for %v failed with remote"+
-				"server: %v", assetID.String(), addr.HostStr())
-			//lint:ignore nilerr failure is expected and logged
+			log.Warnf("Asset lookup for %v failed with remote"+
+				"server '%v': %v", assetID.String(),
```
starius, line 850: A space is missing between "remote" and "server" in the log message.

universe/auto_syncer.go
```diff
@@ -846,18 +846,30 @@ func (f *FederationEnvoy) SyncAssetInfo(ctx context.Context,
...
 		// There should only be one sync diff since we're only syncing
 		// one universe root.
 		if syncDiff != nil {
 			if len(syncDiff) != 1 {
-				log.Debugf("unexpected number of sync diffs: "+
-					"%v", len(syncDiff))
+				log.Warnf("Unexpected number of sync diffs %d "+
+					"when looking up asset %v on remote "+
+					"server '%v': %v", len(syncDiff),
+					assetID.String(), addr.HostStr(), err)
```
starius, line 867: Here `err` is always `nil`, because non-nil err would cause `return` in the branching at line 857.

starius (approved):
LGTM! 🏆
A couple of comments.

Also, maybe it is sufficient to use `Infof` to make it visible? Does `Warnf` cause an alert?

universe/auto_syncer.go
dstadulis: nice catch

guggero: Oops, thanks for noticing! Fixed.

guggero:
> Also, maybe it is sufficient to use `Infof` to make it visible? Does `Warnf` cause an alert?

Warnings don't trigger alarms AFAIK. At least not in our infrastructure, so I think this warrants a bit more than just info level.

PR #1295, opened 2025-01-13
SealBatch RPC endpoint accepts signed group PSBT

ffranr:
This PR modifies the `SealBatch` RPC endpoint such that it accepts a complete signed PSBT directly, instead of requiring the witness to be extracted and passed separately.

The goal of this PR is to simplify the UX for the external key minting procedure. It's a pretty radical simplification because it replaces the `bitcoin-cli decodepsbt` call from the `docs/external-group-key.md` doc. I think we should include something along these lines in the next release. (cc @dstadulis )

---

This work isn't complete. I ran into a problem: the signed PSBT provided by chantools does not contain the right flags/metadata (such as tapscript root) to be able to map the PSBT to a group key request. The PSBT seems to only contain the witness.

So we might need to include the asset ID mapping in the RPC request. Thoughts on this?

I think I'm going to go with requiring the asset ID to signed PSBT mapping passed in as an arg. Perhaps like this:

```proto
message WitnessStack {
    // The witness stack for the asset group.
    repeated bytes witness = 1;
}

message GroupSeal {
    // The asset ID of the pending asset that should be assigned this asset
    // group witness.
    bytes genesis_id = 1;

    oneof group_seal {
        // The serialized witness stack for the asset group.
        WitnessStack witness = 2;

        // The base64 encoded signed group virtual PSBT.
        string signed_group_virtual_psbt = 3;
    }
}

message SealBatchRequest {
    /*
    If true, then the assets currently in the batch won't be returned in the
    response. This is mainly to avoid a lot of data being transmitted and
    possibly printed on the command line in the case of a very large batch.
    */
    bool short_response = 1;

    // This field is deprecated. Use the `seals` field instead.
    // The assetID, witness pairs that authorize asset membership in a group.
    repeated taprpc.GroupWitness group_witnesses = 2 [deprecated = true];

    // Group seals that authorize asset membership in a group.
    repeated GroupSeal seals = 3;
}
```

I think it's wise to require the genesis asset ID because we never know what might be included in the signed PSBT.


guggero:
I think we should have designed the API like this in the first place. Not having the asset ID in there and only relying on the order of things is suboptimal.
But now that we've already have that in place it's a bit unfortunate that we have to deprecate things and look in both places for a few version.

My initial idea was to do everything in the CLI only: Call `ListBatches`, extract the PSBTs, then compare the previous outpoints of the inputs to match the signed PSBTs with the unsigned ones to get the correct order.
But perhaps that's too ugly. So I'm okay with your approach.

But in any case, the previous outpoint should be the thing that can be matched against, as that should be unique per asset.

tapgarden/planter.go
```diff
@@ -204,6 +205,81 @@ func (p *PendingAssetGroup) PSBT(
 	return packet, nil
 }
 
+// DeserializeWitnessStack deserializes a serialized witness stack into a
+// [][]byte.
+//
+// TODO(ffranr): Where should this function live? Do we have an existing
+// function that does this?
+func DeserializeWitnessStack(serialized []byte) ([][]byte, error) {
```
Roasbeef, line 213: If you use `wire.TxWitness`, then you can use the encode/decode methods from there: https://pkg.go.dev/github.com/btcsuite/btcd/wire#TxWitness

Roasbeef: Hmm, or actually looks like not much of it is exported: https://github.com/btcsuite/btcd/blob/v0.24.2/wire/msgtx.go#L588-L626

tapgarden/planter.go
```diff
@@ -1698,6 +1774,38 @@ func (c *ChainPlanter) fundBatch(ctx context.Context, params FundParams,
...
+// matchPsbtToGroupReq attempts to match a signed group virtual PSBT to a
+// corresponding group key request. This is done by comparing the tapscript root
+// in the PSBT to the tapscript root in the group key request.
+func matchPsbtToGroupReq(psbt psbt.Packet,
+	groupReqs []asset.GroupKeyRequest) fn.Option[asset.GroupKeyRequest] {
+
+	// If there is only one group request, we can assume that the PSBT
+	// matches it.
+	//
+	// TODO(ffranr): This is a workaround because the PSBT does not contain
```
Roasbeef, line 1786: I think we should add some minimal information to the group key minting PSBT. This can be used by UIs to display additional information w.r.t what is being requested for mint.

tapgarden/planter.go
```diff
@@ -1787,9 +1895,50 @@ func (c *ChainPlanter) sealBatch(ctx context.Context, params SealParams,
...
+		genesisAssetID := groupReqMatch.NewAsset.ID()
+
+		if _, ok := externalWitnesses[genesisAssetID]; ok {
+			return nil, fmt.Errorf("signed PSBT is a duplicate "+
+				"witness for asset ID: %v", genesisAssetID)
+		}
+
+		// Extract the witness stack from the signed PSBT.
+		witnessStack, err := DeserializeWitnessStack(
+			psbtPacket.Inputs[0].FinalScriptWitness,
```
Roasbeef, line 1928: Alternatively, we should be able to finalize the PSBT at this point. Then we can just read the `Witness` attribute as normal.

ffranr:
Needs rebase.

tapgarden/planter.go
guggero: We should be able to match on the `input[0].PrevOutpoint`.
We can generate the genesis TX from the `GroupKeyRequest` using this:
```
tx, err := req.BuildGroupVirtualTx(&tapscript.GroupTxBuilder{})
```

tapgarden/planter.go
guggero: Yes, I think we should do that in any case, to make sure the TX is actually finalizable, which adds some additional validation.
Also, Ledger puts the signature in a different field (`Inputs[0].TaprootKeySpendSig`), which the following two calls take care of:
```
		psbt.MaybeFinalizeAll(packet)
		psbt.Extract(packet)
```

tapgarden/planter.go
```diff
@@ -1698,6 +1774,38 @@ func (c *ChainPlanter) fundBatch(ctx context.Context, params FundParams,
 	return nil
 }
 
+// matchPsbtToGroupReq attempts to match a signed group virtual PSBT to a
+// corresponding group key request. This is done by comparing the tapscript root
+// in the PSBT to the tapscript root in the group key request.
+func matchPsbtToGroupReq(psbt psbt.Packet,
```
guggero, line 1780: Could we perhaps do the matching of this on the `rpcserver` level at least? Then the planter wouldn't need any new logic and the new field vs. deprecated field would only be visible on the RPC level?
I think we can get the current group key requests in the RPC server by querying the pending batch. Or am I missing something?

guggero (commented):
Nice work!

ZZiigguurraatt:
> I think I'm going to go with requiring the asset ID to signed PSBT mapping passed in as an arg

Instead of requiring the asset id to be passed, if it can't be included in the PSBT, what about just looking through and testing all unsealed assets and see if the witness is valid for any of them? Might be slower, but would be simpler for the user.

guggero:
Changed the base branch to `main`, needs a rebase.

tapgarden/planter.go
ffranr: `DeserializeWitnessStack` removed in latest commits.

tapgarden/planter.go
ffranr: I think I looked into that approach but the pending batch didn't include everything I needed. I think it may not be verbose? I can look into this again.

ffranr:
> > I think I'm going to go with requiring the asset ID to signed PSBT mapping passed in as an arg
> 
> Instead of requiring the asset id to be passed, if it can't be included in the PSBT, what about just looking through and testing all unsealed assets and see if the witness is valid for any of them? Might be slower, but would be simpler for the user.

Oli suggested another way which didn't require asset ID in psbt or passed in via arg, we can match on prevout.

ffranr:
I haven't tried out the CLI arg or the new doc modification. I can do that when I'm online later.

ZZiigguurraatt:
I just tested this and I'm getting `unable to parse signed group virtual PSBT: unexpected EOF` .

rpcserver.go
```diff
@@ -780,6 +780,7 @@ func (r *rpcServer) FundBatch(ctx context.Context,
 func (r *rpcServer) SealBatch(ctx context.Context,
 	req *mintrpc.SealBatchRequest) (*mintrpc.SealBatchResponse, error) {
 
+	// Unmarshal group witnesses from the request.
```
guggero, line 783: So the idea is that the old way and new way could even be mixed? Since both seem to be parsed and then merged (at least with a collision check, so nothing is overwritten).

Not sure if that's more confusing to the user rather than helpful? I think just from an API perspective it makes sense to keep the old way (as that's perhaps easier to use programmatically). But perhaps we should disallow using both at the same time, to avoid some confusion?

tapgarden/planter.go
guggero: With the latest version the matching seems to be quite slim and contained to this file, so don't spend too much time on it.

tapgarden/planter.go
```diff
@@ -1702,6 +1702,52 @@ func (c *ChainPlanter) fundBatch(ctx context.Context, params FundParams,
...
+		if len(tx.Tx.TxIn) != 1 {
+			return fn.None[asset.GroupKeyRequest](), fmt.Errorf(
+				"group virtual TX must have a single input")
+		}
+		vTxInPrevOut := tx.Tx.TxIn[0].PreviousOutPoint
+
+		// If the previous output of the signed PSBT matches the
+		// previous output of the group virtual TX, we have a match.
+		if vTxInPrevOut.Hash.IsEqual(&psbtInPrevOut.Hash) {
+			groupReqMatch = fn.Some(req)
```
guggero, line 1743: Return directly here? Then with a more explicit `fn.None()` after the loop?

tapgarden/planter.go
```diff
@@ -1702,6 +1702,52 @@ func (c *ChainPlanter) fundBatch(ctx context.Context, params FundParams,
...
+		// Sanity check that the group virtual TX.
+		if len(tx.Tx.TxIn) != 1 {
+			return fn.None[asset.GroupKeyRequest](), fmt.Errorf(
+				"group virtual TX must have a single input")
+		}
+		vTxInPrevOut := tx.Tx.TxIn[0].PreviousOutPoint
+
+		// If the previous output of the signed PSBT matches the
+		// previous output of the group virtual TX, we have a match.
+		if vTxInPrevOut.Hash.IsEqual(&psbtInPrevOut.Hash) {
```
guggero, line 1742: nit: Hashes can be compared directly with `==`.

cmd/tapcli/assets.go
```diff
@@ -463,6 +463,11 @@ var sealBatchCommand = cli.Command{
 			Usage: "the asset ID and signature, separated by a " +
 				"colon",
 		},
+		cli.StringSliceFlag{
+			Name: "signed_group_psbt",
+			Usage: "a signed group PSBT for a single asset group " +
```
guggero, line 468: Same UX question here: should we make it more clear that `group_signatures` and `signed_group_psbt` probably shouldn't be used at the same time? Or at least mention that they both do the same thing, just with a different format.

docs/external-group-key.md
```diff
@@ -249,99 +249,10 @@ Successfully signed PSBT:
...
-
 ## Step 6: Seal the batch with the signature
 
-TODO(guggero/ffranr): Use signed PSBT instead, currently it's
-`--group_signatures <asset_id>:<signature>`, but should be
-`--signed_group_psbt` or something like that.
-
 ```shell
-$ tapcli assets mint seal --group_signatures c4b0771c1bd1334bf20df5204c162702a6dc765a9cb15b1bc9e3c91e0282061b:0bff5f83c9a86c8ed98ccbdae26020afa00c51824c8164cc0ab6f53856701b96e6614873b6905f5081b711d25e217cfe39fe179b1678bd29bded592b163b2f81
+$ tapcli assets mint seal --signed-group-psbt cHNidP8BAF4CAAAAATKZro+KSjqg4YQvE0bqBCbWuii3ekKAdufOGo57L8lwAAAAAAAAAAAAAQBlzR0AAAAAIlEgxoSpM86Pyu/bCRKhOc6/2TLXDGXnUnXn69FQqD8gw7cAAAAAAAEBKwBlzR0AAAAAIlEgqUflbsA2uA/P2qYe6qclsUox7koFCR3h1xkwh4+M1wQBCEIBQAv/X4PJqGyO2YzL2uJgIK+gDFGCTIFkzAq29ThWcBuW5mFIc7aQX1CBtxHSXiF8/jn+F5sWeL0pve1ZKxY7L4EAAA==
```
guggero, line 255: nit: actual flag name uses underscores: `signed_group_psbt`.

guggero (approved):
tACK, very nice!

Just a couple user experience questions.

rpcserver.go
guggero: Did you switch between branches? Or use my Ledger branch before I rebased it on top of this one?
Not being able to parse the PSBT probably means it's because of the bump in the `psbt` library that isn't in this branch. So can you try again with the current version of https://github.com/lightninglabs/taproot-assets/pull/1301 that includes this PR?

ZZiigguurraatt: Still getting `unable to parse signed group virtual PSBT: unexpected EOF` with `cold-group-key-pedersen`. I'm using `chantools` to generate the PSBT, by the way. Maybe I'm not supposed to use that?

guggero: Hmm, weird. Can you paste the PSBT here please?

ZZiigguurraatt: cHNidP8BAF4CAAAAAXDjbryJmCvHJRSApZdIx5hXNK2GP9UbAUerZY4GQAKRAAAAAAAAAAAAAQAA1SrjEeoCIlEgENtlBgJGs/eIb0m0qPnftFKJV35Qx99lNuMVlf9/MskAAAAATwEENYfPA4er6h2AAAAALGhINzvEPLXWuCFxp02C9aQjgemKj+QY0XAegOagP/UDkYbhV/i3qKVvtfTAtnnYqIOqj4TwFCDmYGsrG+L/2tsQEGCLuVYAAIABAACAAAAAgE8BBDWHzwMAAAAAAAAAAIUaE1HYiSMu1VX5efx/Y6Fm5nQuXXhBzr+rtftT5UedAoUaE1HYiSMu1VX5efx/Y6Fm5nQuXXhBzr+rtftT5UedEHqaVedWAACAAQAAgAAAAIAAAQErAADVKuMR6gIiUSBhvDnSXEPqw//vGBlJWwSUed/YNc1ywTv23B7kUysE6gEIQgFAIjoG0o9tjBGmQOuiPb//cl+/Q0XmlRQ0aukVK3qfiIM8AkWUCi1fqDiMWj3QRIyjF7eRXhC+CTO8lTw3BrV+eAAA

ffranr: This unit test passes (uses the PSBT from the message above):
```
func TestDecodePsbt(t *testing.T) {
	groupPsbt := "cHNidP8BAF4CAAAAAXDjbryJmCvHJRSApZdIx5hXNK2GP9UbAUerZY4GQAKRAAAAAAAAAAAAAQAA1SrjEeoCIlEgENtlBgJGs/eIb0m0qPnftFKJV35Qx99lNuMVlf9/MskAAAAATwEENYfPA4er6h2AAAAALGhINzvEPLXWuCFxp02C9aQjgemKj+QY0XAegOagP/UDkYbhV/i3qKVvtfTAtnnYqIOqj4TwFCDmYGsrG+L/2tsQEGCLuVYAAIABAACAAAAAgE8BBDWHzwMAAAAAAAAAAIUaE1HYiSMu1VX5efx/Y6Fm5nQuXXhBzr+rtftT5UedAoUaE1HYiSMu1VX5efx/Y6Fm5nQuXXhBzr+rtftT5UedEHqaVedWAACAAQAAgAAAAIAAAQErAADVKuMR6gIiUSBhvDnSXEPqw//vGBlJWwSUed/YNc1ywTv23B7kUysE6gEIQgFAIjoG0o9tjBGmQOuiPb//cl+/Q0XmlRQ0aukVK3qfiIM8AkWUCi1fqDiMWj3QRIyjF7eRXhC+CTO8lTw3BrV+eAAA"

	r := bytes.NewReader([]byte(groupPsbt))
	psbtPacket, err := psbt.NewFromRawBytes(r, true)
	require.NoError(t, err)

	t.Logf("psbtPacket: %v", psbtPacket)
}
```

So i would guess that either the argument is set incorrectly in the command line or there is an error in going from cli arg to RPC message string.

I'm going to modify the error so that we can see what it's actually trying to decode:
```
		if err != nil {
			return nil, fmt.Errorf("unable to parse signed "+
				"group virtual PSBT (signed_psbt=%s): %w",
				groupPsbt, err)
		}
```

ZZiigguurraatt: I'm using the gRPC passing the PSBT into `signed_group_virtual_psbts`, not the CLI.

ZZiigguurraatt: You can give it a try here: https://github.com/lightninglabs/tapdvalidation/tree/sealbatch-signed-group-psbt-test . Follow `README.md`, except you'll have to add an extra step to checkout this branch before you run `./start_mini_META`.

ZZiigguurraatt: You can see here the difference between the old way that worked and the new way that doesn't work: https://github.com/lightninglabs/tapdvalidation/commit/e95ee5b432443570632cc36202c14fb902b4cc9a.

ZZiigguurraatt: Okay, figure out my issue. It has been fixed here: https://github.com/lightninglabs/tapdvalidation/commit/2aa7f2f05eb43c7f3e11eb1381759f240057c231 . The problem is I passed a string as `signed_group_virtual_psbts` instead of an array of strings.

Can we get a more useful test and error for this mistake?

cmd/tapcli/assets.go
ffranr: I've modified the doc here and on the RPC fields to specify that these options are mutually exclusive. But I've not used `oneof` to avoid a breaking change. I suppose we might make a radical simplification at one point and make many breaking changes at once or V2.

GeorgeTsagk (approved):
LGTM!
:fire: this was def needed

rpcserver.go
ZZiigguurraatt: Using Python.

It seems to do more clear error messages if I mix up a string , bytes, or int, but for some reason does not check arrays very well.

PR #1297, opened 2025-01-13
tapcfg: read multiverse cache from cfg instead of using default

Roasbeef:


gijswijs (approved):
LGTM! :tada:

Just to check: with this change we only use `DefaultMultiverseStoreConfig()` in the context of tests

Roasbeef:
> Just to check: with this change we only use DefaultMultiverseStoreConfig() in the context of tests

Correct. When we make the default config, we use a version of that that only makes the inner cache struct: https://github.com/lightninglabs/taproot-assets/blob/main/tapcfg/config.go#L437-L439

Issue #1255, opened 2024-12-16
Assert invoice-RFQ asset ID and HTLC asset ID match


dstadulis:
Link-level robustness to ensure invoice settling logic is preserved

Slated for v0.5.1

PR #1299, opened 2025-01-14
Invoice Acceptor: ensure asset ID match between RFQ and HTLC 

GeorgeTsagk:
## Description

This PR adds an extra strictness check which requires the asset ID of the RFQ quote and HTLC records to match. This is done as an extra strict check to guard against HTLC and RFQ asset ID mismatch, which can lead to malicious behavior where a quote for a different asset is being accounted for when accepting an asset HTLC.

Closes https://github.com/lightninglabs/taproot-assets/issues/1255

### DIFF
--- a/tapchannel/aux_invoice_manager.go
+++ b/tapchannel/aux_invoice_manager.go
@@ -195,6 +195,16 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(_ context.Context,
 		return resp, nil
 	}
 
+	// We now run some validation checks on the asset HTLC.
+	err = s.validateAssetHTLC(htlc)
+	if err != nil {
+		log.Errorf("Failed to validate asset HTLC: %v", err)
+
+		resp.CancelSet = true
+
+		return resp, nil
+	}
+
 	// Convert the total asset amount to milli-satoshis using the price from
 	// the accepted quote.
 	rfqID := htlc.RfqID.ValOpt().UnsafeFromSome()
@@ -365,6 +375,43 @@ func isAssetInvoice(invoice *lnrpc.Invoice, rfqLookup RfqLookup) bool {
 	}
 
 	return false
+}
+
+// validateAssetHTLC runs a couple of checks on the provided asset HTLC.
+func (s *AuxInvoiceManager) validateAssetHTLC(htlc *rfqmsg.Htlc) error {
+	rfqID := htlc.RfqID.ValOpt().UnsafeFromSome()
+
+	// Retrieve the asset identifier from the RFQ quote.
+	identifier, err := s.identifierFromQuote(rfqID)
+	if err != nil {
+		return fmt.Errorf("could not extract assetID from "+
+			"quote: %v", err)
+	}
+
+	if !identifier.HasId() {
+		return fmt.Errorf("asset specifier has empty assetID")
+	}
+
+	// Check for each of the asset balances of the HTLC that the identifier
+	// matches that of the RFQ quote.
+	for _, v := range htlc.Balances() {
+		err := fn.MapOptionZ(
+			identifier.ID(), func(id asset.ID) error {
+				if v.AssetID.Val != id {
+					return fmt.Errorf("mismatch between " +
+						"htlc asset ID and rfq asset " +
+						"ID")
+				}
+
+				return nil
+			},
+		)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // Stop signals for an aux invoice manager to gracefully exit.
--- a/tapchannel/aux_invoice_manager.go
+++ b/tapchannel/aux_invoice_manager.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/taproot-assets/address"
+	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightninglabs/taproot-assets/rfq"
 	"github.com/lightninglabs/taproot-assets/rfqmath"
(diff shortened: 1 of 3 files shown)


tapchannel/aux_invoice_manager.go
```diff
@@ -197,6 +198,15 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(_ context.Context,
 	// Convert the total asset amount to milli-satoshis using the price from
 	// the accepted quote.
 	rfqID := htlc.RfqID.ValOpt().UnsafeFromSome()
+
+	assetID := s.assetIDFromQuote(rfqID)
+	for _, v := range htlc.Balances() {
+		if v.AssetID.Val != *assetID {
```
ffranr, line 204: To avoid dereference of nil at `*assetID` I think `assetIDFromQuote` should return `asset.ID, error` (error in case of missing policy for HTLC).

ffranr (changes requested):
Good idea!

dstadulis:
TODO
- [ ] itest to cover the negative cases

tapchannel/aux_invoice_manager_test.go
```diff
@@ -422,6 +431,42 @@ func TestAuxInvoiceManager(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "asset invoice, wrong asset htlc",
```
GeorgeTsagk, line 435: Added coverage for this in our unit test cases

tapchannel/aux_invoice_manager.go
```diff
@@ -247,6 +248,34 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(_ context.Context,
 	return resp, nil
 }
 
+// assetIDFromQuote retrieves the quote by looking up the rfq manager's maps of
+// accepted quotes based on the passed rfq ID. If there's a match, the asset
+// ID is returned.
+func (s *AuxInvoiceManager) assetIDFromQuote(
```
guggero, line 271: To avoid us needing to touch this again once we add group key support, should we turn this into `identifierFromQuote()` instead and return the identifier directly?

tapchannel/aux_invoice_manager_test.go
```diff
@@ -100,6 +100,7 @@ func (m *mockHtlcModifier) HtlcModifier(ctx context.Context,
 		res, err := handler(ctx, r)
 
 		if err != nil {
+			fmt.Printf("### got handler err: %v\n", err)
```
guggero, line 103: Stray log.

tapchannel/aux_invoice_manager.go
guggero: This is unaddressed. By returning the identifier directly, we can do `if !ident.HasId() {`.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -754,7 +799,7 @@ func testInvoiceManager(t *rapid.T) {
 
 // TestAuxInvoiceManagerProperty runs property based tests on the
 // AuxInvoiceManager.
-func TestAuxInvoiceManagerProperty(t *testing.T) {
+func TestAuxIndvoiceManagerProperty(t *testing.T) {
```
guggero, line 802: This seems to be adding a typo instead of fixing one.

tapchannel/aux_invoice_manager.go
```diff
@@ -198,6 +198,23 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(_ context.Context,
...
 	// the accepted quote.
 	rfqID := htlc.RfqID.ValOpt().UnsafeFromSome()
+
+	assetID, err := s.assetIDFromQuote(rfqID)
+	if err != nil {
+		return nil, fmt.Errorf("could not extract assetID from "+
+			"quote: %v", err)
+	}
+	for _, v := range htlc.Balances() {
+		if v.AssetID.Val != *assetID {
```
guggero, line 208: This probably works for now as we only have a single asset ID in an HTLC. But this will definitely change with group key support. So perhaps we can add minimum forward compatibility by extracting this into a function, something like `htlcMatchesIdentifier(i asset.identifier)`. Then we can just change the logic there.

tapchannel/aux_invoice_manager.go
```diff
@@ -263,12 +280,14 @@ func (s *AuxInvoiceManager) assetIDFromQuote(
 	switch {
 	case isBuy:
 		if buyQuote.Request.AssetSpecifier.HasId() {
-			return buyQuote.Request.AssetSpecifier.UnwrapIdToPtr(), nil
+			req := buyQuote.Request
```
guggero, line 283: This belongs to the previous commit.

guggero (commented):
Careful self review of the commits would've prevented 50-60% of my comments...
Approach looks okay to me but can be perhaps made a bit more generic.

tapchannel/aux_invoice_manager.go
```diff
@@ -197,6 +198,28 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(_ context.Context,
...
+		err := checkHTLCAssetSpecifierMatch(v, identifier)
+		if err != nil {
+			log.Errorf("asset specifier mismatch, cancelling "+
+				"htlc set: %v", err)
+
+			resp.CancelSet = true
+
+			return resp, nil
+		}
+	}
```
ffranr, line 221: I think we should consider rolling this functionality into a method called:
```
func (s *AuxInvoiceManager) validateHTLC(htlc rfqmsg.Htlc) error
```

* That method could extract `rfqID` from the given `htlc`.
* Call `s.identifierFromQuote` to get the associated asset specifier.
* Validate HTLC balances.

And then we can drop `checkHTLCAssetSpecifierMatch` which is a less general HTLC verifier.

ffranr (commented):
Nice! This is pretty much there IMO.

I'm just going to suggest a potential simplification in this review.

tapchannel/aux_invoice_manager.go
GeorgeTsagk: yeah, agree with this simplification

I think @guggero 's recommendation already leaned towards this direction, so we might as well fully abstract this away (since it's gonna change a lot soon-ish anyway)

tapchannel/aux_invoice_manager.go
```diff
@@ -367,6 +377,33 @@ func isAssetInvoice(invoice *lnrpc.Invoice, rfqLookup RfqLookup) bool {
...
+		return fmt.Errorf("could not extract assetID from "+
+			"quote: %v", err)
+	}
+
+	// Check for each of the asset balances of the HTLC that the identifier
+	// matches that of the RFQ quote.
+	for _, v := range htlc.Balances() {
+		if !identifier.HasId() {
+			return fmt.Errorf("asset specifier has empty assetID")
+		}
```
guggero, line 396: This can be moved to before the loop.

tapchannel/aux_invoice_manager.go
```diff
@@ -367,6 +377,33 @@ func isAssetInvoice(invoice *lnrpc.Invoice, rfqLookup RfqLookup) bool {
...
+
+		if v.AssetID.Val != *identifier.UnwrapIdToPtr() {
+			return fmt.Errorf("mismatch between htlc asset ID " +
+				"and rfq asset ID")
+		}
```
guggero, line 401: This feels very much against the whole `Option` paradigm... We could add an `ID()` or `IDOption()` method to the specifier that returns the underlying `fn.Option` directly, then we could do this the "intended" way:

```
		err := fn.MapOptionZ(
			identifier.IdOption(), func(id asset.ID) error {
				if v.AssetID.Val != id {
					return fmt.Errorf("mismatch between " +
						"htlc asset ID and rfq asset " +
						"ID")
				}

				return nil
			},
		)
		if err != nil {
			return err
		}
```

But non-blocking.

guggero (approved):
Looks good. One more suggestion, but non-blocking.

ffranr (approved):
Good addition

PR #1300, opened 2025-01-14
monitoring: export collector timeout as prometheus config

GeorgeTsagk:
This PR exports the prometheus collector timeout as a prometheus config parameter.

With this timeout we will be able to observe values on our loadtest tapds for longer periods, as the timeout will be harder to reach.

### DIFF
--- a/monitoring/config.go
+++ b/monitoring/config.go
@@ -1,6 +1,8 @@
 package monitoring
 
 import (
+	"time"
+
 	"github.com/lightninglabs/taproot-assets/tapdb"
 	"github.com/lightninglabs/taproot-assets/tapgarden"
 	"github.com/lightninglabs/taproot-assets/universe"
@@ -10,6 +12,8 @@ import (
 // PrometheusConfig is the set of configuration data that specifies if
 // Prometheus metric exporting is activated, and if so the listening address of
 // the Prometheus server.
+//
+// nolint: lll
 type PrometheusConfig struct {
 	// Active, if true, then Prometheus metrics will be exported.
 	Active bool `long:"active" description:"if true prometheus metrics will be exported"`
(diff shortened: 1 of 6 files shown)


Roasbeef (changes requested):
How about we make this a config option? This needs to be set with the scape timeout in mind. Otherwise, it's possible that scrapes just continue to build up if the interval is << the scrape latency.

monitoring/prometheus.go
```diff
@@ -16,11 +16,8 @@ var (
 	// serverMetrics is a global variable that holds the Prometheus metrics
 	// for the gRPC server.
 	serverMetrics *grpc_prometheus.ServerMetrics
-)
 
-const (
-	// dbTimeout is the default database timeout.
-	dbTimeout = 20 * time.Second
```
guggero, line 23: Let's keep the constant and rename it to `defaultTimeout`? Then we don't have to have the magic number below.

GeorgeTsagk: Sure, makes sense

monitoring/config.go
```diff
@@ -45,7 +54,8 @@ type PrometheusConfig struct {
 // metrics exporter.
 func DefaultPrometheusConfig() PrometheusConfig {
 	return PrometheusConfig{
-		ListenAddr: "127.0.0.1:8989",
-		Active:     false,
+		ListenAddr:          "127.0.0.1:8989",
+		Active:              false,
+		CollectorRPCTimeout: time.Second * 25,
```
guggero, line 59: nit: `defaultTimeout`.

monitoring/prometheus.go
```diff
@@ -12,15 +12,17 @@ import (
...
 var (
 	// serverMetrics is a global variable that holds the Prometheus metrics
 	// for the gRPC server.
 	serverMetrics *grpc_prometheus.ServerMetrics
-)
 
-const (
-	// dbTimeout is the default database timeout.
-	dbTimeout = 20 * time.Second
+	promTimeout time.Duration
```
guggero, line 25: nit: Godoc comment would be nice.

monitoring/prometheus.go
```diff
@@ -46,6 +48,12 @@ func (p *PrometheusExporter) Start() error {
 		return fmt.Errorf("server metrics not set")
 	}
 
+	if p.config.CollectorRPCTimeout.Seconds() != 0 {
```
guggero, line 51: nit: how about:
```
	promTimeout = defaultTimeout
	if p.config.CollectorRPCTimeout.Seconds() != 0 {
		promTimeout = p.config.CollectorRPCTimeout
	}
```

guggero (approved):

ffranr (approved):

PR #1302, opened 2025-01-16
Aggregate Universe Stats: always store result of querySyncStats

GeorgeTsagk:
Our universe stats could run into a case where if in an un-cached state (like on system start-up) the db query takes longer than the default client timeout (usually 30s) then we'd fail the query and return early. This would cause the next client to also receive an error instead of the cached result, making this codepath stuck in an endless failing query. With this change we parallelize the db query and we always wait for the result in order to store it in cache. We may fail the function earlier, but the result will always be retrieved & persisted. This way, even if the 1st client retrieves an error, the next one will receive the cached (late) result of the previous run.

Closes https://github.com/lightninglabs/taproot-assets/issues/1304

### DIFF
--- a/tapdb/universe_stats.go
+++ b/tapdb/universe_stats.go
@@ -575,14 +575,50 @@ func (u *UniverseStats) AggregateSyncStats(
 
 	log.Debugf("Populating aggregate sync stats")
 
-	dbStats, err := u.querySyncStats(ctx)
-	if err != nil {
-		return dbStats, err
-	}
+	var (
+		resChan = make(chan universe.AggregateStats, 1)
+		errChan = make(chan error, 1)
+	)
 
-	// We'll store the DB stats then start our time after function to wipe
-	// the stats pointer so we'll refresh it after a period of time.
-	u.statsSnapshot.Store(&dbStats)
+	// We'll fire the db query in a separate go routine, with an undefined
+	// timeout. This way, we'll always retrieve the result regardless of
+	// what happens in the current context. This way we're always waiting
+	// for the call to complete and caching the result even if this
+	// function's result is an error.
+	go func() {
+		// Note: we have the statsMtx held, so even if a burst of
+		// requests took place in an un-cached state, this call would
+		// not be triggered multiple times.
+		dbStats, err := u.querySyncStats(context.Background())
+		if err != nil {
+			errChan <- err
+			return
+		}
+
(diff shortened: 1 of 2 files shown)


tapdb/universe_stats.go
```diff
@@ -575,14 +575,48 @@ func (u *UniverseStats) AggregateSyncStats(
 
 	log.Debugf("Populating aggregate sync stats")
 
-	dbStats, err := u.querySyncStats(ctx)
-	if err != nil {
+	var (
+		resChan = make(chan universe.AggregateStats, 1)
+		errChan = make(chan error, 1)
+		dbStats universe.AggregateStats
```
guggero, line 581: nit: move this just before the `select`, otherwise it might look like it's shadowed in the Goroutine.

tapdb/universe_stats.go
```diff
@@ -575,14 +575,48 @@ func (u *UniverseStats) AggregateSyncStats(
...
+		// We'll store the DB stats so that it can be read from cache
+		// later.
+		u.statsSnapshot.Store(&dbStats)
+
+		resChan <- dbStats
+	}()
+
+	select {
+	case <-ctx.Done():
+		log.Debugf("ctx timeout before retrieving aggregate sync stats")
```
guggero, line 610: nit: Log statements should always start capitalized. Also, make it clear that the outer context timed out?

guggero (commented):
Concept ACK. Some kind of test case (unit or integration) would be nice.

tapdb/universe_stats_test.go
```diff
@@ -255,6 +255,61 @@ func TestUniverseStatsEvents(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(
+		context.Background(), quickTimeoutDuration,
+	)
+	defer cancel()
+
+	// The tiny timeout duration should make the following call result in a
+	// context deadline related error.
+	_, err := sh.db.AggregateSyncStats(ctx)
+	require.Error(t, err)
+	require.ErrorContains(t, err, "context deadline exceeded")
```
guggero, line 303: micro-nit: `ErrorContains()` already implicitly does `Error()`, so the above line is not strictly needed.

guggero (approved):
Very nice!

ffranr (approved):
Nice!

PR #1307, opened 2025-01-17
Remove `.fail` files and update `.gitignore`

ffranr:
This PR removes `.fail` files generated by the Rapid testing framework and updates the `.gitignore` to prevent them from being committed in the future.

### Why

`.fail` files are used locally to reproduce test failures and are not intended for version control. They are environment-specific and provide no value in the repository.  The presence of a `.fail` file does not suppress or void unit tests.

### Changes

- Removed existing `.fail` files.
- Updated `.gitignore` to exclude `*.fail`.


guggero (approved):
Offline discussion seems relevant here. I don't really mind having the files around if they have a tiny bit of usefulness. But also don't have super strong arguments for keeping them.

GeorgeTsagk (approved):
LGTM
Rationale for keeping them is that these fail cases represent inputs that failed once in the past, so they are can being retried when running prop tests

Problem with that is:
a) Many of these fail cases were generated during development (during the shaping of the prop test itself), so they might not really represent actual edge cases
b) Diff seems quite big, so would def go for a 16k line cleanup unless there's a strong reason not to

PR #1310, opened 2025-01-20
tapdb: fix performance of aggregate stats

guggero:
Addition to https://github.com/lightninglabs/taproot-assets/pull/1302.

Difference can best be seen when running `make unit-trace pkg=tapdb long-tests=1 case=TestUniverseIndexPerformance/name=query_aggregated`:

```
    universe_perf_test.go:279: No indices exec time: 28.961577804s
    universe_perf_test.go:280: With indices exec time: 11.647093789s
    universe_perf_test.go:289: Improvement: 2.49x
```

I also confirmed on mainnet that the old and new query return the exact same result:
[old-query.txt](https://github.com/user-attachments/files/18479924/old-query.txt)
[new-query.txt](https://github.com/user-attachments/files/18479925/new-query.txt)

The speedup on mainnet is also quite significant. The old query took 2.5 minutes while the new one only took 15 seconds.

Hopefully this should make https://github.com/lightninglabs/taproot-assets/pull/1302 not necessary (but let's keep it anyway, since it's a good idea as well).

Shout out to @jbrill who's performance test setup allowed me to test this quite easily.

### DIFF
--- a/tapdb/sqlc/migrations/000027_better_universe_stats.down.sql
+++ b/tapdb/sqlc/migrations/000027_better_universe_stats.down.sql
@@ -1,0 +1,16 @@
+-- We restore the previous version of the universe_stats view, as it was created
+-- in the 000010_universe_stats.up.sql migration file.
+
+DROP VIEW universe_stats;
+
+CREATE VIEW universe_stats AS
+SELECT
+    COUNT(CASE WHEN u.event_type = 'SYNC' THEN 1 ELSE NULL END) AS total_asset_syncs,
+    COUNT(CASE WHEN u.event_type = 'NEW_PROOF' THEN 1 ELSE NULL END) AS total_asset_proofs,
+    roots.asset_id,
+    roots.group_key,
+    roots.proof_type
+FROM universe_events u
+JOIN universe_roots roots
+  ON u.universe_root_id = roots.id
+GROUP BY roots.asset_id, roots.group_key, roots.proof_type;
--- a/tapdb/sqlc/migrations/000027_better_universe_stats.up.sql
+++ b/tapdb/sqlc/migrations/000027_better_universe_stats.up.sql
@@ -1,0 +1,37 @@
+DROP VIEW universe_stats;
+
+CREATE VIEW universe_stats AS
+WITH sync_counts AS (
+    SELECT universe_root_id, COUNT(*) AS count
+    FROM universe_events
+    WHERE event_type = 'SYNC'
+    GROUP BY universe_root_id
+), proof_counts AS (
+    SELECT universe_root_id, event_type, COUNT(*) AS count
+    FROM universe_events
+    WHERE event_type = 'NEW_PROOF'
+    GROUP BY universe_root_id, event_type
+), aggregated AS (
+    SELECT COALESCE(SUM(count), 0) as total_asset_syncs,
+           0 AS total_asset_proofs,
+           universe_root_id
+    FROM sync_counts
+    GROUP BY universe_root_id
+    UNION ALL
+    SELECT 0 AS total_asset_syncs,
+           COALESCE(SUM(count), 0) as total_asset_proofs,
+           universe_root_id
+    FROM proof_counts
+    GROUP BY universe_root_id
+)
+SELECT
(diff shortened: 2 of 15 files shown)


tapdb/sqlc/migrations/000027_better_universe_stats.down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file is empty on purpose. There is nothing to roll back for this
+-- migration.
```
jbrill, line 2: Do we not want to restore the old `universe_view` in this rollback?

tapdb/sqlc/migrations/000027_better_universe_stats.up.sql
```diff
@@ -0,0 +1,37 @@
...
+SELECT
+    SUM(ag.total_asset_syncs) AS total_asset_syncs,
+    SUM(ag.total_asset_proofs) AS total_asset_proofs,
+    roots.asset_id,
+    roots.group_key,
+    roots.proof_type
+FROM aggregated ag
+JOIN universe_roots roots
+    ON ag.universe_root_id = roots.id
+GROUP BY roots.asset_id, roots.group_key, roots.proof_type
```
jbrill, line 36: Should we not tweak the index here like in the test?
```
		CREATE INDEX IF NOT EXISTS idx_universe_leaves_asset
		ON universe_leaves(
			asset_genesis_id, universe_root_id
		);`
```

tapdb/universe_perf_long_test.go
```diff
@@ -6,7 +6,7 @@ package tapdb
 const longTestScale = 5
 
 var (
-	numAssets         = 100 * longTestScale
-	numLeavesPerTree  = 300 * longTestScale
-	numQueries        = 100 * longTestScale
+	numAssets        = 100 * longTestScale
+	numLeavesPerTree = 300 * longTestScale
+	numQueries       = 100 * longTestScale
```
jbrill, line 11: Should the test timeout be a bit longer for the `_long_test.go` or do you think the 5 minute timeout is ok?

tapdb/sqlc/migrations/000027_better_universe_stats.up.sql
jbrill: Same with `idx_mssmt_nodes_composite`

jbrill (approved):
Nice!! Huge optimization 👏 just added a few comments/questions

tapdb/universe_perf_test.go
```diff
@@ -139,6 +155,52 @@ const (
 		SELECT asset_id, supply, COUNT(*) as num_leaves
 		FROM asset_supply
 		GROUP BY asset_id, supply`
+
+	// queryAggregatedStats gets aggregated stats for the universe. This
+	// corresponds to the QueryUniverseStats query.
+	queryAggregatedStats = `
+		WITH stats AS (
```
GeorgeTsagk, line 162: why re-write this query? can't we grab it from some sqlc artifact?

GeorgeTsagk (approved):
:zap:  LGTM, tACK

tapdb/sqlc/migrations/000027_better_universe_stats.up.sql
guggero: I'm not sure what you mean? Extend the indexes to cover the fields mentioned in `GROUP BY`? Those are all covered already be pre-existing indexes.

tapdb/universe_perf_test.go
guggero: Hmm, good idea. Turns out we actually can.

tapdb/sqlc/migrations/000027_better_universe_stats.up.sql
guggero: No I just noticed that the names of some of the indexes that were removed in the test didn't actually exist under those names, that's the only thing changed in the indexes. All other commits are about having a more optimized version of the `universe_stats` view.

jbrill: Got it -- thanks for the clarification 👍

Issue #990, opened 2024-07-02
[bug] - universe: QueryAssetRoots returns an empty error when no asset root can be found 


Roasbeef:
### Background

Today `QueryAssetRoots` will return an empty error when no asset root can be found: https://github.com/lightninglabs/taproot-assets/blob/2d8d4849df66c9632038018a8da01a882be39baf/rpcserver.go#L4707-L4710

If a user attempts to query for a non-existent asset ID, the'll get an error that:
```
 [tapcli] rpc error: code = Unknown desc = unable to sync universe: unable to fetch roots for universe sync: missing universe id
``` 

Which is confusing, as the real error lies in the call to unmarshall the empty root: https://github.com/lightninglabs/taproot-assets/blob/2d8d4849df66c9632038018a8da01a882be39baf/universe_rpc_diff.go#L120

We'll find that  there's no asset ID as no real root was returned: https://github.com/lightninglabs/taproot-assets/blob/2d8d4849df66c9632038018a8da01a882be39baf/universe_rpc_diff.go#L44-L56

Historically, I think this behavior exists to just return an empty response rather than error out for REST clients? Either way, the CLI should be able to properly detect that no roots are found. This may have some implications to the sync and auto syncer in the `universe` package. 

### Steps to reproduce

Run `tapcli universe sync --asset_id=X` with an asset ID that doesn't exist. 

### Expected behavior

CLI says the asset ID doesn't exist on the universe. 

### Actual behavior

Obscure decoding error.

jharveyb:
I have a similar recollection re: empty responses for REST clients.

On a fresh look, I'm not sure why we don't have a nil-ness check here before trying to unmarshal?

https://github.com/lightninglabs/taproot-assets/blob/2d8d4849df66c9632038018a8da01a882be39baf/universe_rpc_diff.go#L116

It seems like the 'more-correct' option there would be to return something like `universe.ErrNoUniverseRoot` instead of unmarshalling something empty. Or skip unmarshalling and return an empty `universe.Root{}`.

In that case, I think we'd want to extend `executeSync` to validate the contents of `targetRoots` before actually executing the sync, and return a more useful error re: missing roots on the target server for the sync.

I'm not sure if other users of the sync interface like the scheduled sync rely on that unmarshal failing as a signal of a missing root though.

dstadulis:
Roas mentioned
>[down stream systems] expect no error when no root, and checks the length, so we'll need to update that as well
CC @jharveyb

Issue #1315, opened 2025-01-22
[universe]: return universe.ErrNoUniverseRoot when no root is found


guggero:
This is a follow-up issue after fixing https://github.com/lightninglabs/taproot-assets/issues/990 in a temporary manner to not break backward compatibility.

Once we shipped the fix for https://github.com/lightninglabs/taproot-assets/issues/990 (which is #1314) in `v0.5.1` and most clients updated, we can start actually returning a proper error in the `QueryAssetRoots` RPC if no root is found (the current response is just an empty root).

So this fix can be done for 0.6 or 0.7.

PR #1316, opened 2025-01-22
universe: sync "missing universe id" when syncing unknown asset ID

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/990, at least partially.

We'll need another fix in future versions and actually return an error properly (https://github.com/lightninglabs/taproot-assets/issues/1315).
But to not break older clients, we'll only do that after a full version (e.g. we ship this fix in 0.5.1, then with 0.6 or 0.7 we return an actual error, knowing that enough clients will understand to not interrupt the whole sync process).

### DIFF
--- a/itest/universe_test.go
+++ b/itest/universe_test.go
@@ -717,16 +717,16 @@ func testUniverseFederation(t *harnessTest) {
 	require.Empty(t.t, rootResp.TransferRoot.AssetName)
 	require.Empty(t.t, rootResp.TransferRoot.AmountsByAssetId)
 
-	// Because the above QueryAssetRoots doesn't return an error, this
-	// currently leads to an error further down in the sync logic, which can
-	// be shown by syncing an invalid asset ID.
+	// The sync logic should detect that the response from the universe
+	// server was empty and just skip that root, not resulting in an error
+	// anymore.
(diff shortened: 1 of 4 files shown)


ffranr:
How does this change affect older nodes?

If an older node attempts to sync an asset ID that is unknown to a universe server implementing this change, the older node will receive an error. Does this terminate the universe syncing process for that server?

Maybe we should add a version field to RPC endpoint's `QueryAssetRoots` request message `universerpc.AssetRootQuery`?

guggero:
There are no errors returned yet. I tried to make that very clear in the commit messages and all the comments. We're _preparing_ for being able to deal with an error being returned but we don't return it yet. We'll return errors with https://github.com/lightninglabs/taproot-assets/issues/1315, which will be done in 0.6 or 0.7.

ffranr (approved):
Once we start returning this new error, syncing might break for v0.5.0 nodes and older. But we're ok with that because the new group key formulation is only understood by nodes v0.5.1 and newer anyway.

So we expect everyone to upgrade their nodes past v0.5.0 soon enough.

GeorgeTsagk (approved):
Lgtm 🚢

PR #1320, opened 2025-01-23
gitignore: add rule to ignore btcd-itest log files

ffranr:
Update `.gitignore` to exclude `itest` directory log files matching the pattern `.log.<num>.gz`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ cmd/tapd/tapd
 /itest/chantools
 /itest/regtest/*
 /itest/*.log
+/itest/*.log.*
 /itest/.backendlogs/*
 /itest/.minerlogs/*
 /itest/tapd-itest


guggero (approved):

.gitignore
```diff
@@ -16,6 +16,7 @@ cmd/tapd/tapd
 /itest/chantools
 /itest/regtest/*
 /itest/*.log
+/itest/*.log.*
```
GeorgeTsagk, line 19: nit: was the above line just ineffective? in that case could only keep the new pattern

GeorgeTsagk (approved):

.gitignore
ffranr: I think that in `.gitignore` files, patterns are interpreted using globbing syntax, not regular expressions.

And therefore `/itest/*.log.*` does not match on `/itest/file.log`

Issue #1318, opened 2025-01-23
[feature]: Minimal Code Changes for Future Universe Commitment Transaction Support


ffranr:
We need to make minimal changes now to enable seamless support for fully-featured universe commitment transactions in the future. The following steps outline the required adjustments:

1. **Taproot Locking for Minting Change Output**  
   Modify the minting transaction so that the change output is locked in a way that it can only be spent via `tapd`. This will involve re-using the existing Pedersen commitment work to commit to the genesis asset ID in the tapscript of the output. The output will use a Taproot format, with the internal key temporarily set to an LND hot wallet key. This setup ensures we can later sign the universe commitment transaction spending this output using an external cold signing key.

2. **User-Enabled Universe Commitment Support**  
   Add a user-facing option to enable universe commitment during minting. If enabled:
   - Warn the user that the change output will remain unspendable until full universe commitment support is implemented.
   - Restrict the minting batch to include only a single asset group for events with universe commitment enabled.

3. **Metadata Update for Minting Proofs**
   Update the asset minting proof metadata to indicate whether universe commitment is enabled for the minting event.

This change should will allow retroactive application of the feature to past minting events which have uni commitment support enabled.

Roasbeef:
> This will involve re-using the existing Pedersen commitment work to commit to the genesis asset ID in the tapscript of the output.

Why do we need to commit to the genesis ID here? IIRC, we never finalized the design of _what_ would actually go into this new change output. 

> The output will use a Taproot format, with the internal key temporarily set to an LND hot wallet key. This setup ensures we can later sign the universe commitment transaction spending this output using an external cold signing key.


These two statements appear to contradict each other. 

In an earlier design, the group key was intended to be used for the internal key of this commitment/change output. At that point, we were still tweaking the group key directly with the asset ID, so we would've had to deal with the double tweak issues. With GKR v1, the internal key is "bare", so I think we should be able to just re-use it for the internal key of this Bitcoin commitment anchor output. 

One issue with the above idea is that then if the group key is actually a cold key, user interaction may be needed to update the on chain commitment. But perhaps that'll just be a part of the normal minting process?

> Warn the user that the change output will remain unspendable until full universe commitment support is implemented.


Seems like something that requires a UI to be truly effective? 

> Restrict the minting batch to include only a single asset group for events with universe commitment enabled.


[...]

PR #1325, opened 2025-01-24
Add pre-commitment output to mint anchor TX

ffranr:
This is a work-in-progress draft that will eventually address [Minimal Code Changes for Future Universe Commitment Transaction Support](https://github.com/lightninglabs/taproot-assets/issues/1318).

At a high level, this PR introduces the following functionality: if the "enable universe commitment" flag is set on a batch, the change output will be locked with a script that only `tapd` can spend. Additionally, the PR includes validations and restrictions to ensure that the "enable universe commitment" feature can only be applied to minting batches containing assets from the same asset group. For more details, refer to the linked issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325)


ffranr:
This PR only locks the pre-commitment output. Before release, we should include code that spends that output in a universe commitment transaction. So that change doesn't get locked up unnecessarily.

cmd/commands/assets.go
```diff
@@ -154,6 +155,15 @@ var mintAssetCommand = cli.Command{
 			Usage: "the master fingerprint of the key the xpub " +
 				"was derived from",
 		},
+		cli.BoolFlag{
+			Name: universeCommitmentsName,
+			Usage: "if set, the asset group will be minted with " +
```
guggero, line 160: I think we also need to mention that this locks up the BTC change output of the batch. Might even be worth asking for confirmation in the CLI (e.g. `You requested to use universe commitments. This feature is not fully implemented yet and the on-chain commitments aren't produced at the moment. To make sure the change output created by this batch isn't used up in the meantime, it will be locked to a key that can't be spent by `lnd` automatically. In a future version when the commitments are produce on-chain, this change will be unlocked again. Are you sure you want to proceed?`)

guggero (commented):
Very nice and clean work!
Definitely on the right track, got just a few smaller things.

tapgarden/planter.go
```diff
@@ -2238,11 +2238,91 @@ func (c *ChainPlanter) CancelBatch() (*btcec.PublicKey, error) {
...
+	// If the universe commitments feature is disabled for this seedling,
+	// we can skip this any further modifications. The seedling will be
+	// validated relative to the batch later.
+	if !req.UniverseCommitments {
+		return nil
+	}
+
+	// Set the universe commitments flag in the seedling metadata.
+	//
+	// TODO(ffranr): Ensure that this does not conflict with the
```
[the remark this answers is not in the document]
ffranr: Ah I see now that the meta TLV fields are never merged into the `Data` field. I'll remove this TODO.

tapgarden/planter.go
```diff
@@ -2238,11 +2238,91 @@ func (c *ChainPlanter) CancelBatch() (*btcec.PublicKey, error) {
...
+				"in batch (anchor_seedling_name=%s)",
+				anchorSeedlingName)
+		}
+
+		if anchor.Meta == nil {
+			return fmt.Errorf("group anchor seedling has no meta "+
+				"(anchor_seedling_name=%s)", anchorSeedlingName)
+		}
+
+		if anchor.Meta.DelegationKey.IsNone() {
```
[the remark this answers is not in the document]
ffranr: I'll add. It's a bit overkill but I like that each function looks after itself and doesn't make assumptions, to an extent.

cmd/commands/assets.go
ffranr: Instead of just issuing a warning, I was hoping we could ensure that the release containing this PR also includes the code responsible for creating and broadcasting the universe commitment transaction. That way, the change-locking issue would never be a concern.

Do you think that's reasonable given where we are today?

tapgarden/planter.go
```diff
@@ -745,6 +807,40 @@ func (c *ChainPlanter) fundGenesisPsbt(ctx context.Context,
...
+				"funded batch but no seedlings in batch")
+		}
+
+		// Extract delegated key from the first encountered seedling in
+		// the batch. For simplicity, the delegation key should be set
+		// for all seedlings in the batch.
+		var delegationKey keychain.KeyDescriptor
+		for _, seedling := range c.pendingBatch.Seedlings {
+			var err error
+			delegationKey, err = seedling.DelegationKey.UnwrapOrErr(
```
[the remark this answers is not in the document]
ffranr: Yes, good idea.

tapgarden/planter.go
```diff
@@ -1779,6 +1875,9 @@ func (c *ChainPlanter) fundBatch(ctx context.Context, params FundParams,
 		}
 	}
 
+	// TODO(ffranr): Do we need to write the pre-commitment output key to
```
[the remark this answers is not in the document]
ffranr: I'll take a look at adding that to this PR.

cmd/commands/assets.go
Roasbeef: I wonder if we should try to make an escape hatch here for the funds. So something similar to the way anchor outputs in LN work today (anyone can spend it after N blocks, but instead of anyone a sole lnd key).

Roasbeef (changes requested):
> Reviewed 4 of 4 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, 2 of 2 files at r9, 1 of 1 files at r10, 2 of 2 files at r11, 3 of 3 files at r12, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325)* status: all files reviewed, 21 unresolved discussions (waiting on @ffranr and @guggero)

___
*[`tapgarden/planter.go` line 676 at r13](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWNJmVE8KHE0euhWAm:-OJWNJmVE8KHE0euhWAn:b6uhhzc) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/918cc162de1835b1707b13bfe7795e86d6f4e78d/tapgarden/planter.go#L676)):*
> ```Go
> 
> 	// Override PSBT change output.
> 	psbtChangeOut := &fundedPsbt.Pkt.Outputs[fundedPsbt.ChangeOutputIndex]
> ```

I think we should _add_ an output, rather than modify the existing change output. Otherwise if a single high value input is spent to create a batch, then we'll end up locking that in the pre-commitment output. 

With our current model, when we got to spend the pre-commitment output, what happens to the output value? Will it be rolled into the commitment output, or will we ask the wallet for a change address to refund the pre-commitment funds back to?
___
*[`tapgarden/planter.go` line 818 at r13](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWNkh5DThMI3W0tnZw:-OJWNkh5DThMI3W0tnZx:bqeh48z) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/918cc162de1835b1707b13bfe7795e86d6f4e78d/tapgarden/planter.go#L818)):*
> ```Go
> 		for _, seedling := range c.pendingBatch.Seedlings {
> 			var err error
> 			delegationKey, err = seedling.DelegationKey.UnwrapOrErr(
> ```

I think we can just do something like: `fn.All(seedlings, DelegationKey.Some()` 
___
*[`tapgarden/planter.go` line 1869 at r13](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWNrsF9yRWFFKUnUpV:-OJWNrsF9yRWFFKUnUpW:b3qm6fm) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/918cc162de1835b1707b13bfe7795e86d6f4e78d/tapgarden/planter.go#L1869)):*
> ```Go
> 	}
> 
> 	// TODO(ffranr): Do we need to write the pre-commitment output key to
> ```

For each batch on disk, we should track if there's a pre-commitment output or not. I think it would be good to start building out that new table structure now. This way in the future once we implement stuff like catch up, or the spending functionality, we can easily look up all the pre-commitment outputs that are related to a given group key (and mint/burn delta). 
___
*[`tapgarden/batch.go` line 331 at r5](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWLR9q484d4bPmChdZ:-OJWLR9q484d4bPmChd_:bh9ym34) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/6ce78687e415995103f33f40e1a794ea305f0dd7/tapgarden/batch.go#L331)):*
> ```Go
> 	// At this point, we know that the universe commitment flag is enabled
> 	// for the seedling. Therefore, the delegation key must be set.
> 	delegationKey, err := newSeedling.DelegationKey.UnwrapOrErr(
> ```

Perhaps I've missed where this delegation key was added? 

Also I think we can refine the name here a bit to make the purpose clearer. Is this the key that can sign the new minting payload, or is this the key that can be used in place of the internal group key for the pre-commitment output? 
___
*[`tapgarden/batch.go` line 362 at r5](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWLvPp6DQa6CdE3fTu:-OJWLvPp6DQa6CdE3fTv:b-ps6xfs) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/6ce78687e415995103f33f40e1a794ea305f0dd7/tapgarden/batch.go#L362)):*
> ```Go
> 	// batch.
> 	for _, seedling := range m.Seedlings {
> 		// Ensure that the delegation key matches that of the candidate
> ```

May be missing some context here, but we'd allow all the assets created in a batch to share a pre-commitment key? Does this mean that they'll then create N outputs, one for each distinct asset group, or will they be combined? 
___
*[`tapgarden/planter.go` line 2281 at r11](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWN0TuE3IgubCD73XJ:-OJWN0Tv5mJE5bhLoJK1:b-uqo158) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea57e7ad8ec665cbe777e8f813a76f4876824ed9/tapgarden/planter.go#L2281)):*
> ```Go
> 		}
> 
> 		if anchor.DelegationKey.IsNone() {
> ```

Why do we need the delegation key in two places? 

`anchor.Meta.DelegationKey` and `anchor.DelegationKey`
___
*[`tapgarden/batch.go` line 383 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWKnkj614XQKELoPDG:-OJWKnkj614XQKELoPDH:b1isko8) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/95a5940bdcca38c83385031bf3a32c67486f2cdd/tapgarden/batch.go#L383)):*
> ```Go
> 		return nil
> 	}
> ```

I thought the plan was to isolate mints that require a universe commitment from others? So I would expect to see some logic here that asserts the only contents of the batch now have the exact same group key. 

That also brings a new potential requirement to mind: do we support multi-tranche issuance with this scheme? So I want to create new tranches A, B, C (diff UTXO sizes and diff asset IDs) in the same transaction.

cmd/commands/assets.go
guggero: I don't think that's necessary. If anyone really needs to spend this, they could use the `lnd` PSBT API to sweep it back.

ffranr (commented):
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325)* status: 3 of 24 files reviewed, 21 unresolved discussions (waiting on @guggero and @Roasbeef)

___
*[`tapgarden/batch.go` line 383 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWKnkj614XQKELoPDG:-OK2ONcB-7t2I6oYEEUB:b-ifjmh0) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/95a5940bdcca38c83385031bf3a32c67486f2cdd/tapgarden/batch.go#L383)):*

I thought the plan was to isolate mints that require a universe commitment from others? So I would expect to see some logic here that asserts the only contents of the batch now have the exact same group key. 

That also brings a new potential requirement to mind: do we support multi-tranche issuance with this scheme? So I want to create new tranches A, B, C (diff UTXO sizes and diff asset IDs) in the same transaction.
</blockquote></details>

Currently, this code should restrict a batch to a single asset group when the universe commitment feature is enabled. However, I believe we can support multiple seedlings within the batch under this scheme. I'm not sure if I'm overlooking anything.
___
*[`tapgarden/batch.go` line 331 at r5](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWLR9q484d4bPmChdZ:-OJYL7aW0P3wMUoegh2d:b-rqciuy) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/6ce78687e415995103f33f40e1a794ea305f0dd7/tapgarden/batch.go#L331)):*

Perhaps I've missed where this delegation key was added? 

Also I think we can refine the name here a bit to make the purpose clearer. Is this the key that can sign the new minting payload, or is this the key that can be used in place of the internal group key for the pre-commitment output?
</blockquote></details>

As I understand it, we intended to use the same delegation key for both the Taproot internal key of the pre-commitment output and the mint payload signing key.
___
*[`tapgarden/batch.go` line 362 at r5](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWLvPp6DQa6CdE3fTu:-OK2NrSpEDRQ68jbaxfF:bggsnd) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/6ce78687e415995103f33f40e1a794ea305f0dd7/tapgarden/batch.go#L362)):*

May be missing some context here, but we'd allow all the assets created in a batch to share a pre-commitment key? Does this mean that they'll then create N outputs, one for each distinct asset group, or will they be combined?
</blockquote></details>

My understanding is that each asset group in the batch, when the universe commitment feature is enabled, will have a distinct pre-commitment output. However, currently, if the pre-commitment feature is enabled, the minting batch is limited to a single asset group.
___
*[`tapgarden/planter.go` line 2281 at r11](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWN0TuE3IgubCD73XJ:-OK2PoD16IDSG5BEnHL7:b-3hj3bg) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ea57e7ad8ec665cbe777e8f813a76f4876824ed9/tapgarden/planter.go#L2281)):*

Why do we need the delegation key in two places? 

`anchor.Meta.DelegationKey` and `anchor.DelegationKey`
</blockquote></details>

I think I've managed to push handling of metadata into a single spot so that we can just make use of `Seedling`. Metadata is inflexible as it represents what we intend on serializing.
___
*[`tapgarden/planter.go` line 676 at r13](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWNJmVE8KHE0euhWAm:-OK2OxgA4kPNE4kuWf6I:b-yj4j72) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/918cc162de1835b1707b13bfe7795e86d6f4e78d/tapgarden/planter.go#L676)):*

I think we should _add_ an output, rather than modify the existing change output. Otherwise if a single high value input is spent to create a batch, then we'll end up locking that in the pre-commitment output. 

With our current model, when we got to spend the pre-commitment output, what happens to the output value? Will it be rolled into the commitment output, or will we ask the wallet for a change address to refund the pre-commitment funds back to?
</blockquote></details>

I've updated this PR so that the pre-commitment output is now a distinct new output rather than a reuse of the change output.
___
*[`tapgarden/planter.go` line 818 at r13](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWNkh5DThMI3W0tnZw:-OK2PMIE0K-JXaFWI7YW:bh5f5zu) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/918cc162de1835b1707b13bfe7795e86d6f4e78d/tapgarden/planter.go#L818)):*

I think we can just do something like: `fn.All(seedlings, DelegationKey.Some()`
</blockquote></details>

I've revised this such that we only concern ourselves with the group anchor seedling.
___
*[`tapgarden/planter.go` line 1869 at r13](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OJWNrsF9yRWFFKUnUpV:-OK2PelJ2zlNcj3INV8D:b-v7ey8y) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/918cc162de1835b1707b13bfe7795e86d6f4e78d/tapgarden/planter.go#L1869)):*

For each batch on disk, we should track if there's a pre-commitment output or not. I think it would be good to start building out that new table structure now. This way in the future once we implement stuff like catch up, or the spending functionality, we can easily look up all the pre-commitment outputs that are related to a given group key (and mint/burn delta).
</blockquote></details>

I've added a new SQL table to this PR.

ffranr:
Next steps here:
* a new PR for unit tests and refactor to facilitate unit tests
* only support internal key for now? In which case we can SQL foreign key ref into the internal key table.

tapgarden/planter.go
```diff
@@ -764,19 +977,71 @@ func (c *ChainPlanter) fundGenesisPsbt(ctx context.Context,
...
 
 	// Sanity check the funded PSBT.
 	if fundedGenesisPkt.ChangeOutputIndex == -1 {
-		return nil, fmt.Errorf("undefined change output index in " +
+		return zero, fmt.Errorf("undefined change output index in " +
 			"funded anchor transaction")
 	}
 
+	// Classify anchor transaction output indexes.
+	anchorOutIndexes, err :=
```
guggero, line 990: nit: wrap function arguments instead.

tapgarden/planter.go
```diff
@@ -669,9 +670,95 @@ func (c *ChainPlanter) newBatch() (*MintingBatch, error) {
 	return newBatch, nil
 }
 
+// preCommitmentOutput creates the pre-commitment output for a batch that uses
+// universe commitments.
+func preCommitmentOutput(pendingBatch *MintingBatch) (*wire.TxOut,
+	*btcec.PublicKey, *btcec.PublicKey, error) {
```
guggero, line 676: We could add the `wire.TxOut` of the output to the `PreCommitmentOutput` struct and return that here. Then we can have a single option in `fundGenesisPsbt` and do the validation in a `Validate()` method to reduce the size of that already long function a bit.

tapdb/sqlc/migrations/000030_mint_anchor_uni_commitments.down.sql
```diff
@@ -0,0 +1,9 @@
+-- Drop the table mint_anchor_uni_commitments.
+    DROP TABLE IF EXISTS mint_anchor_uni_commitments;
```
guggero, line 2: nit: unnecessary indentation.

tapdb/sqlc/queries/assets.sql
```diff
@@ -1016,3 +1017,22 @@ FROM genesis_assets assets
...
+
+-- Upsert a record into the mint_anchor_uni_commitments table.
+-- If a record with the same batch_id and group_key already exists, update the
+-- existing record. Otherwise, insert a new record.
+-- name: UpsertMintAnchorUniCommitment :one
+INSERT INTO mint_anchor_uni_commitments (
+    id, batch_id, tx_output_index, taproot_internal_key, group_key
+)
+VALUES ($1, $2, $3, $4, $5)
+ON CONFLICT(batch_id, group_key) DO UPDATE SET
```
guggero, line 1029: There is no unique key that spans over these two fields, so this will never trigger.
I wonder why SQLite or sqlc didn't pick that up?
We'll need an explicit unique key if we want this combination to be unique.
Also makes me think we should have a unit test that covers the specific case.

tapdb/asset_minting.go
```diff
@@ -239,6 +239,9 @@ type PendingAssetStore interface {
 	// FetchAssetMetaForAsset fetches the asset meta for a given asset.
 	FetchAssetMetaForAsset(ctx context.Context,
 		assetID []byte) (sqlc.FetchAssetMetaForAssetRow, error)
+
+	FetchMintAnchorUniCommitment(ctx context.Context,
```
guggero, line 247: nit: missing Godoc comment.

tapgarden/caretaker.go
```diff
@@ -612,18 +594,11 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
...
-		// TODO(jhb): return the anchor index instead of change? or both
-		// so this works for N outputs
-		b.anchorOutputIndex, err = extractAnchorOutputIndex(
-			b.cfg.Batch.GenesisPacket,
-		)
-		if err != nil {
-			return 0, err
-		}
+		// Unpack output indexes.
+		changeOutputIndex := b.cfg.Batch.GenesisPacket.ChangeOutputIndex
```
guggero, line 599: nit: move `genesisPacket := b.cfg.Batch.GenesisPacket` up here to shorten these?

tapdb/asset_minting.go
```diff
@@ -382,18 +451,18 @@ func (a *AssetMintingStore) CommitMintingBatch(ctx context.Context,
 				ctx, q, genesisOutpoint,
 			)
 			if err != nil {
-				return fmt.Errorf("%w: %w",
-					ErrUpsertGenesisPoint, err)
+				return fmt.Errorf("unable to insert mint "+
```
guggero, line 454: nit: can get rid of the packet serialization above now.

tapdb/asset_minting.go
```diff
@@ -1354,15 +1422,16 @@ func (a *AssetMintingStore) CommitBatchTx(ctx context.Context,
...
-			RawKey:        rawBatchKey,
-			MintingTxPsbt: psbtBuf.Bytes(),
-			ChangeOutputIndex: sqlInt32(
-				genesisPacket.ChangeOutputIndex,
-			),
-			GenesisID: sqlInt64(genesisPointID),
-		})
-		return err
+		// Insert the batch transaction.
+		err = insertMintAnchorTx(
```
guggero, line 1426: nit: another packet serialization call that can be removed here.

tapdb/asset_minting.go
```diff
@@ -1499,6 +1568,8 @@ func (a *AssetMintingStore) AddSproutsToBatch(ctx context.Context,
 
 	var writeTxOpts AssetStoreTxOptions
 	return a.db.ExecTx(ctx, &writeTxOpts, func(q PendingAssetStore) error {
+		// Upsert the assets with genesis and get the database ID of the
```
guggero, line 1571: We could add the `upsertGenesisPoint` call to `insertMintAnchorTx` function as well. That would mean it would be called twice in this specific case (since `upsertAssetsWithGenesis` also does that), but we'd be able to remove two instances of it further above...

guggero (commented):
Very nice work! Looks great :tada:

tapgarden/planter.go
ffranr: Elsewhere, `PreCommitmentOutput` is populated from the database where the tx ouput is missing. We only have the index at that point.

guggero: Yes, I saw that. But can't we re-create the tx output from the database fields? Then that could actually even be a method on the struct that generates the `txOut`.

tapdb/sqlc/migrations/000030_mint_anchor_uni_commitments.down.sql
ffranr: we should add a SQL linter...

tapgarden/planter.go
```diff
@@ -669,6 +669,26 @@ func (c *ChainPlanter) newBatch() (*MintingBatch, error) {
...
+	// Add one output to anchor all assets which are being minted.
+	txTemplate.AddTxOut(tapsend.CreateDummyOutput())
+
+	// Formulate the PSBT packet from the template transaction.
+	genesisPkt, err := psbt.NewFromUnsignedTx(txTemplate)
+	if err != nil {
+		return zero, fmt.Errorf("unable to make psbt packet: %w", err)
+	}
+
+	return *genesisPkt, nil
```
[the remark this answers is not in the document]
ffranr: My understanding is/was:
1) Avoid pointer types as function arguments where possible.
2) Functions returning pointer types is fine. But if possible, and not too cumbersome, just return non-pointer.

So that's why, in this PR, I've added a method with this sig (because it's handy to use `nil` when returning `err != nil`):
```
func preCommitmentOutput(pendingBatch *MintingBatch) (*wire.TxOut,
	*btcec.PublicKey, *btcec.PublicKey, error)
```

And also a method like this:
```
func unfundedAnchorPsbt(preCommitmentTxOut fn.Option[wire.TxOut]) (psbt.Packet,
	error) {

	var zero psbt.Packet
```

tapdb/sqlc/queries/assets.sql
ffranr: I think that I'm going to pair on `(batch_id, tx_output_index)` instead with `UNIQUE` and `CONFLICT`. Which effectively means: that each batch could have multiple different pre-commitment outputs and each pre-commitment output would relate to a single asset group.

tapgarden/planter.go
ffranr: I've revised `preCommitmentOutput` along the lines you suggested. Please let me know if that's what you had in mind.

tapgarden/planter.go
```diff
@@ -2680,3 +2939,117 @@ var _ Planter = (*ChainPlanter)(nil)
...
+
+	return PreCommitmentOutput{
+		OutIdx:      outIdx,
+		InternalKey: internalKey,
+		GroupPubKey: groupPubKey,
+	}
+}
+
+// TxOut returns the pre-commitment output as a wire.TxOut instance.
+func (p *PreCommitmentOutput) TxOut() (wire.TxOut, error) {
```
guggero, line 2960: Nice!

tapdb/sqlc/migrations/000030_mint_anchor_uni_commitments.up.sql
```diff
@@ -0,0 +1,42 @@
...
+    tx_output_index INTEGER NOT NULL,
+
+    -- The Taproot output internal key for the pre-commitment output.
+    taproot_internal_key BLOB,
+
+    -- The asset group key associated with the universe commitment.
+    group_key BLOB,
+
+    -- Ensure that each batch_id, tx_output_index pair is unique.
+    UNIQUE(batch_id, tx_output_index)
```
guggero, line 41: Even though this is nice and easy to understand, I think we should get into the habit of declaring any indexes as proper named indexes outside of the table definition.
Because otherwise if we eve want to change this index, we'd need to drop and re-create the whole table.

tapdb/sqlc/queries/assets.sql
```diff
@@ -1016,3 +1017,22 @@ FROM genesis_assets assets
...
+-- Upsert a record into the mint_anchor_uni_commitments table.
+-- If a record with the same batch_id and group_key already exists, update the
+-- existing record. Otherwise, insert a new record.
+-- name: UpsertMintAnchorUniCommitment :one
+INSERT INTO mint_anchor_uni_commitments (
+    id, batch_id, tx_output_index, taproot_internal_key, group_key
+)
+VALUES ($1, $2, $3, $4, $5)
+ON CONFLICT(batch_id, tx_output_index) DO UPDATE SET
+    tx_output_index = EXCLUDED.group_key,
```
guggero, line 1030: `EXCLUDED.group_key,` seems to be wrong! Or you meant to do `group_key = EXCLUDED.group_key`?

Also, since IMO these `ON CONFLICT ... EXCLUDED ...` queries are sometimes a bit hard to read, I got into the habit of adding a small comment above the fields, just mentioning whether it's a NOP (https://github.com/lightninglabs/taproot-assets/blob/74f52ac31c6f9c36ee45e65a85637133c6ac0aed/tapdb/sqlc/queries/assets.sql#L164) or not (https://github.com/lightninglabs/taproot-assets/blob/74f52ac31c6f9c36ee45e65a85637133c6ac0aed/tapdb/sqlc/queries/assets.sql#L175).
I know, in this case it's quite obvious.

tapdb/sqlc/queries/assets.sql
```diff
@@ -1016,3 +1017,22 @@ FROM genesis_assets assets
...
+    taproot_internal_key = EXCLUDED.taproot_internal_key
+RETURNING id;
```
guggero, line 1032: I still think we should cover this with a unit test. But can be done in a follow-up PR.

guggero (commented):
Very nice!

One last request regarding the unique index. Other than that looks great :100:

tapgarden/planter.go
guggero: Yeah, this is exactly what I had in mind. Nice!

guggero (approved):

tapdb/sqlc/migrations/000030_mint_anchor_uni_commitments.up.sql
```diff
@@ -0,0 +1,44 @@
...
+
+    -- The index of the mint batch anchor transaction pre-commitment output.
+    tx_output_index INTEGER NOT NULL,
+
+    -- The Taproot output internal key for the pre-commitment output.
+    taproot_internal_key BLOB,
+
+    -- The asset group key associated with the universe commitment.
+    group_key BLOB
+);
```
Roasbeef, line 39: During offline discussion, we discussed adding a bool  that denotes if it has been spent. Alternatively, this can be a foreign key ref as well.

Roasbeef: I'll add this as a migration in an upcoming PR. Nothing spends the pre-commitment yet, so we don't lose any context if we do it in a new migration.

Roasbeef (changes requested):
Changes are looking pretty good!

The one thing that appears to be missing is proper restrictions to ensure that we won't create a batch with multiple seedlings with distinct group keys that each want a universe commitment. The checks in `validateUniCommitment` don't appear to be strong enough. 

AFAICT, we'll allow `AddSeedling` to succeed if the group anchor name is specified, but it's a different group, albeit with the same dec display value. A few unit tests here for the added validation logic would go a long way re increasing confidence that we're properly implementing this restriction. 

> Reviewed 4 of 8 files at r14, 1 of 1 files at r15, 1 of 2 files at r16, 2 of 2 files at r17, 1 of 1 files at r18, 2 of 2 files at r19, 1 of 1 files at r20, 1 of 1 files at r21, 1 of 1 files at r22, 1 of 1 files at r23, 1 of 1 files at r25, 1 of 1 files at r26, 3 of 3 files at r27, 2 of 2 files at r28, 1 of 1 files at r29, 16 of 16 files at r35, 4 of 4 files at r36, 8 of 8 files at r37, 8 of 8 files at r38, 1 of 1 files at r39, 1 of 1 files at r40, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-:-OKZ-eQV16-Oa9HK7wr_:b-7z0hyd)* status: all files reviewed, 39 unresolved discussions (waiting on @ffranr and @guggero)

___
*[`tapgarden/batch.go` line 457 at r38](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OKYzR3O9jChKOFenDG5:-OKYzR3O9jChKOFenDG6:b-ugnql6) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/5fb69eb7f994fb88a52760116cc22f45e8244bb1/tapgarden/batch.go#L457)):*
> ```Go
> 	// already part of the batch. The group anchor must have been added to
> 	// the batch before the candidate seedling.
> 	if newSeedling.GroupAnchor == nil {
> ```

Why is this relevant? 

This set of checks doesn't appear to be strong enough. IIUC, what we want is that:
  * If you have new asset issuance in a batch _and_ it has opted into universe commitments, then this is the only one of its kind in the batch (we don't support multiple pre-commitment outputs). 
___
*[`tapgarden/planter.go` line 775 at r35](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OKYucS_0BACsVSxyOc4:-OKYucS_0BACsVSxyOc5:b-8fx2u7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/3625751125c56cfff687bcae6e3cdaf8a2e433f9/tapgarden/planter.go#L775)):*
> ```Go
> // AnchorTxOutputIndexes specifies the output indexes of the batch mint anchor
> // transaction.
> type AnchorTxOutputIndexes struct {
> ```

:+1: 
___
*[`tapgarden/planter.go` line 903 at r35](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OKYutbpDJ4yFCu6l06o:-OKYutbpDJ4yFCu6l06p:b-pbs2hc) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/3625751125c56cfff687bcae6e3cdaf8a2e433f9/tapgarden/planter.go#L903)):*
> ```Go
> 			preCommitmentOut,
> 			func(p PreCommitmentOutput) lfn.Result[wire.TxOut] {
> 				txOut, err := p.TxOut()
> ```

Style nit: you could actually just pass it in as: 
```go
lfn.NewResult(p.TxOut())
```

ffranr (commented):
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325)* status: 4 of 25 files reviewed, 39 unresolved discussions (waiting on @guggero and @Roasbeef)

___
*[`tapgarden/batch.go` line 457 at r38](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OKYzR3O9jChKOFenDG5:-OKcyCBu87hNf3mSaiXL:b-gfw1fg) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/5fb69eb7f994fb88a52760116cc22f45e8244bb1/tapgarden/batch.go#L457)):*

Why is this relevant? 

This set of checks doesn't appear to be strong enough. IIUC, what we want is that:
  * If you have new asset issuance in a batch _and_ it has opted into universe commitments, then this is the only one of its kind in the batch (we don't support multiple pre-commitment outputs).
</blockquote></details>

I've added further assertions and unit tests for this validation method.
___
*[`tapgarden/planter.go` line 775 at r35](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OKYucS_0BACsVSxyOc4:-OKcyR7GCdbmVCdLzrWD:b-896fix) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/3625751125c56cfff687bcae6e3cdaf8a2e433f9/tapgarden/planter.go#L775)):*

:+1:
</blockquote></details>

Done.
___
*[`tapgarden/planter.go` line 903 at r35](https://reviewable.io/reviews/lightninglabs/taproot-assets/1325#-OKYutbpDJ4yFCu6l06o:-OKcyPq77bZXXQ124cye:b-896fix) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/3625751125c56cfff687bcae6e3cdaf8a2e433f9/tapgarden/planter.go#L903)):*

Style nit: you could actually just pass it in as: 
```go
lfn.NewResult(p.TxOut())
```
</blockquote></details>

Done.

guggero (approved):
Very nice, LGTM :tada:

Issue #1321, opened 2025-01-24
[bug]: make custom record detection more specific


guggero:
Payments that have custom records (such as the experimental endorsement bit) are currently rejected when `lnd` is running within `litd`, because of this code: https://github.com/lightninglabs/taproot-assets/blob/3c22b8465fd52c2e59bdefdd0f99b0d75f840c57/tapchannel/aux_invoice_manager.go#L155

The check should be more specific and count Taproot Asset channel specific custom records only. A new helper function to do that should be created in `tapchannelmsg` that checks for the possible TLV types defined there that can occur in an HTLC.

PR #1330, opened 2025-01-27
tapchannel: detect asset related custom records correctly

guggero:
Fixes more occurrences related to #1321.

### DIFF
--- a/tapchannel/commitment.go
+++ b/tapchannel/commitment.go
@@ -127,7 +127,10 @@ func ComputeView(ourBalance, theirBalance uint64,
 					entry.ParentIndex)
 			}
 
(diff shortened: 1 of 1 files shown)


gijswijs (approved):
LGTM! :tada:

I've checked the repo with the following regex `\.CustomRecords.*\d` and couldn't find any other occurences apart from this one:

https://github.com/lightninglabs/taproot-assets/blob/4d9c148cfdd735475f40adc097b8c0a3aadf24cb/itest/channels_test.go#L38-L42

That one already checks on the custom record type, so that won't be a problem.

GeorgeTsagk (approved):
LGTM

PR #1285, opened 2025-01-09
Add `mintV2` loadtest

GeorgeTsagk:
## Description

This is a lighter version of the mintTest, which also uses a few more configuration parameters to control how minting occurs. We now mint normal assets into a fixed total number of groups.

This was originally intended as a rewrite of the existing mint test, but it is introduced as a totally separate function. We could keep both functions and even try to run them in parallel at some point.

The minter is fixed to Alice node, as Bob could not mint into a group that is controlled by Alice. So fixing this role to Alice makes things much simpler.

PR #1331, opened 2025-01-27
Add loadtest `sendV2`

GeorgeTsagk:
Based on #1285 

This PR adds the last 2 commits.

## Description

This PR adds a new refactored send test for our loadtest suite, `sendV2. 

The new send test uses normal assets which may have balances greater than `1` (case for collectibles), which offers a bit more coverage for our code (coin selection, psbt signing, etc).

Compared to the original send test, we strip away any non-mandatory assertions and RPC calls, keeping them to the minimum required to make things happen.

### DIFF
--- a/itest/loadtest/loadtest-sample.conf
+++ b/itest/loadtest/loadtest-sample.conf
@@ -22,6 +22,10 @@ send-test-num-sends=5
 # Number of assets to send in each send operation for send test
 send-test-num-assets=1
 
+# Type of asset to use in the test case. For V2 test cases, a "normal" type is
+# required.
+send-asset-type="normal"
+
 # Timeout for the entire test suite
 test-suite-timeout=120m
 
--- a/itest/loadtest/send_test.go
+++ b/itest/loadtest/send_test.go
@@ -29,22 +29,24 @@ func sendTest(t *testing.T, ctx context.Context, cfg *Config) {
 	ctxt, cancel := context.WithTimeout(ctxb, cfg.TestTimeout)
 	defer cancel()
 
+	sendType := stringToAssetType(cfg.SendAssetType)
+
 	t.Logf("Running send test, sending %d asset(s) of type %v %d times",
-		cfg.NumAssets, cfg.SendType, cfg.NumSends)
+		cfg.NumAssets, sendType, cfg.NumSends)
 	for i := 1; i <= cfg.NumSends; i++ {
 		send, receive, ok := pickSendNode(
-			t, ctx, cfg.NumAssets, cfg.SendType, alice, bob,
+			t, ctx, cfg.NumAssets, sendType, alice, bob,
 		)
 		if !ok {
 			t.Fatalf("Aborting send test at attempt %d of %d as "+
 				"no node has enough balance to send %d "+
 				"assets of type %v", i, cfg.NumSends,
-				cfg.NumAssets, cfg.SendType)
+				cfg.NumAssets, sendType)
 			return
 		}
 
 		sendAssets(
-			t, ctxt, cfg.NumAssets, cfg.SendType, send, receive,
+			t, ctxt, cfg.NumAssets, sendType, send, receive,
 			bitcoinClient, cfg.TestTimeout,
 		)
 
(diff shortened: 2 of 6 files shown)


itest/loadtest/loadtest-sample.conf
```diff
@@ -22,6 +22,10 @@ send-test-num-sends=5
 # Number of assets to send in each send operation for send test
 send-test-num-assets=1
 
+# Type of asset to use in the test case. For V2 test cases, a 0 type is
+# required.
+send-test-send-type=0
+
```
ffranr, line 28: Is 0 here the `asset.Type` `Normal`? I think this should be a string rather than an int. Maybe something like:
```
send-test-asset-type=normal
```

itest/loadtest/send_test.go
```diff
@@ -29,30 +30,181 @@ func sendTest(t *testing.T, ctx context.Context, cfg *Config) {
...
+	// Alice is set to be the minter in mintV2, so we use Alice's universe.
+	var (
+		uniHost = fmt.Sprintf(
+			"%s:%d", alice.cfg.Host, alice.cfg.Port,
+		)
+	)
+
+	// Let's make sure Bob is aware of all the assets that Alice may have
+	// minted.
+	itest.SyncUniverses(ctx, t, bob, alice, uniHost, cfg.TestTimeout)
```
guggero, line 80: I think we probably have to add a new parameter to this function that allows us to sync transfer proofs as well.
When I ran this test a second time after running 1 mintV2 and 1 sendV2 test, this sync never completed, because it only tried syncing issuance proofs.

itest/loadtest/send_test.go
```diff
@@ -29,30 +30,181 @@ func sendTest(t *testing.T, ctx context.Context, cfg *Config) {
...
+
+	ctxb := context.Background()
+	ctxt, cancel := context.WithTimeout(ctxb, cfg.TestTimeout)
+	defer cancel()
+
+	sendType := stringToAssetType(cfg.SendAssetType)
+
+	// Alice is set to be the minter in mintV2, so we use Alice's universe.
+	var (
+		uniHost = fmt.Sprintf(
```
guggero, line 73: nit: if just `uniHost := fmt.Sprintf(...)` then can fit on one line.

guggero (commented):
One question/concern about repeatability of this test, other than that looks good to me.

guggero:
> `LiT` tests are failing, as they are using the `itest.SyncUniverses` helper
> 
> [Lit PR](https://github.com/lightninglabs/lightning-terminal/pull/964) that updates to this latest commit of tapd

We could make it a functional option to not break other code...

itest/utils.go
```diff
@@ -834,15 +839,45 @@ func MintAssetExternalSigner(t *harnessTest, tapNode *tapdHarness,
...
+		so.syncMode = mode
+	}
+}
+
 // SyncUniverses syncs the universes of two tapd instances and waits until they
 // are in sync.
 func SyncUniverses(ctx context.Context, t *testing.T, clientTapd,
 	universeTapd commands.RpcClientsBundle, universeHost string,
-	timeout time.Duration) {
+	timeout time.Duration, opts ...SyncUniverseOpt) {
```
guggero, line 870: :+1:

itest/loadtest/utils.go
```diff
@@ -289,3 +289,18 @@ func getBitcoinConn(t *testing.T, cfg *BitcoinConfig) *rpcclient.Client {
 
 	return client
 }
+
+// stringToAssetType converts a string of an asset type to its respective taprpc
+// type enum value.
+func stringToAssetType(t string) taprpc.AssetType {
+	switch t {
+	case "normal":
+		return taprpc.AssetType_NORMAL
```
guggero, line 298: nit: could just have this as the `default` case, since that's what we do in the end anyway.

itest/loadtest/send_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	prand "math/rand"
+	"math/rand/v2"
```
guggero, line 7: There's a `v2`? TIL...

guggero (approved):
tACK, LGTM :tada:

PR #1335, opened 2025-01-28
proof: add new universe commitment TLV fields to asset meta reveal

guggero:
~Depends on https://github.com/lightninglabs/taproot-assets/pull/1337.~

Fixes https://github.com/lightninglabs/taproot-assets/issues/956.

Adds two new fields to the TLV record of the asset's meta reveal:
 - Decimal display: Allows minting of assets with non-JSON meta data as we don't need to rely on writing the field there)
 - Canonical universe URL: Define the "home" universe for an asset that universe commitments will be synced to

**Important for reviewers**:
 - This PR does not add the RPC or CLI level flags for the two new entries. Those only really make sense with https://github.com/lightninglabs/taproot-assets/pull/1325 merged as well, so we'll add those later.
 - ~While writing this PR a bug was found in the backward compatible way we use odd TLV records. See e9286f0d245b09cfda79d9e00d5acd0508e6dbf7. Unfortunately that means these new fields will only be backward compatible to `v0.5.1` (this PR) and not all the way back to `v0.5.0`.~
 - ~We prepare for future backward compatibility tests with 293c60f22cddce014e4f5a6e6bccbeda3a5eba28. The idea is to be able to have a CI step that check out an older version (e.g. `v0.5.1`) in the future and run that test with updated test vectors of the future version, to make sure they can still be parsed and validated by that older version.~

### DIFF
--- a/itest/asset_meta_test.go
+++ b/itest/asset_meta_test.go
@@ -167,31 +167,83 @@ func testMintAssetWithDecimalDisplayMetaField(t *harnessTest) {
 	secondAssetReq.Asset.GroupKey = groupKey
 	secondAssetReq.Asset.DecimalDisplay = 0
 
-	// Reissuance should fail if the decimal display does not match the
+	// Re-issuance should fail if the decimal display does not match the
 	// group anchor.
 	_, err = t.tapd.MintAsset(ctxt, secondAssetReq)
 	require.ErrorContains(t.t, err, "decimal display does not match")
 
-	// Requesting a decimal display without specifying the metadata type as
-	// JSON should fail.
+	// Requesting a decimal display without specifying the metadata field
+	// with at least the type should fail.
 	secondAssetReq.Asset.DecimalDisplay = firstAsset.DecimalDisplay
 	secondAssetReq.Asset.AssetMeta = nil
 
 	_, err = t.tapd.MintAsset(ctxt, secondAssetReq)
-	require.ErrorContains(t.t, err, "decimal display requires JSON")
+	require.ErrorContains(
+		t.t, err, "decimal display requires asset metadata",
+	)
 
-	// If we update the decimal display to match the group anchor, minting
-	// should succeed. We also unset the metadata to ensure that the decimal
-	// display is set as the sole JSON object if needed.
+	// Attempting to set a different decimal display in the JSON meta data
+	// as in the new RPC request field should give us an error as well.
 	secondAssetReq.Asset.AssetMeta = &taprpc.AssetMeta{
 		Type: taprpc.AssetMetaType_META_TYPE_JSON,
+		Data: []byte(`{"foo": "bar", "decimal_display": 3}`),
 	}
-	MintAssetsConfirmBatch(
+	_, err = t.tapd.MintAsset(ctxt, secondAssetReq)
+	require.ErrorContains(
+		t.t, err, "decimal display in JSON asset meta does not match",
+	)
+
+	// If we set a valid asset meta again, minting should succeed, using the
+	// same decimal display as the group anchor.
+	secondAssetReq.Asset.AssetMeta.Data = []byte(`{"foo": "bar"}`)
+	secondAssets := MintAssetsConfirmBatch(
 		t.t, t.lndHarness.Miner().Client, t.tapd,
 		[]*mintrpc.MintAssetRequest{secondAssetReq},
 	)
+	require.Len(t.t, secondAssets, 1)
+	require.NotNil(t.t, secondAssets[0].DecimalDisplay)
+	require.EqualValues(
+		t.t, 2, secondAssets[0].DecimalDisplay.DecimalDisplay,
+	)
+
+	// For an asset with a JSON meta data type, we also expect the decimal
+	// display to be encoded in the meta data JSON.
+	metaResp, err := t.tapd.FetchAssetMeta(
+		ctxt, &taprpc.FetchAssetMetaRequest{
+			Asset: &taprpc.FetchAssetMetaRequest_AssetId{
+				AssetId: secondAssets[0].AssetGenesis.AssetId,
+			},
+		},
+	)
+	require.NoError(t.t, err)
+	require.Contains(t.t, string(metaResp.Data), `"foo":"bar"`)
+	require.Contains(t.t, string(metaResp.Data), `"decimal_display":2`)
 
 	AssertGroupSizes(
 		t.t, t.tapd, []string{hex.EncodeToString(groupKey)}, []int{2},
+	)
+
+	// Now we also test minting an asset that uses the opaque meta data type
+	// and check that the decimal display is correctly encoded as well.
+	thirdAsset := &mintrpc.MintAsset{
+		AssetType: taprpc.AssetType_NORMAL,
+		Name:      "test-asset-opaque-decimal-display",
+		AssetMeta: &taprpc.AssetMeta{
+			Type: taprpc.AssetMetaType_META_TYPE_OPAQUE,
+			Data: []byte("some opaque data"),
+		},
+		Amount:         123,
+		DecimalDisplay: 7,
+	}
+	thirdAssetReq := &mintrpc.MintAssetRequest{Asset: thirdAsset}
+	thirdAssets := MintAssetsConfirmBatch(
+		t.t, t.lndHarness.Miner().Client, t.tapd,
+		[]*mintrpc.MintAssetRequest{thirdAssetReq},
+	)
+
+	require.Len(t.t, thirdAssets, 1)
+	require.NotNil(t.t, thirdAssets[0].DecimalDisplay)
+	require.EqualValues(
+		t.t, 7, thirdAssets[0].DecimalDisplay.DecimalDisplay,
 	)
 }
--- a/proof/encoding.go
+++ b/proof/encoding.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"net/url"
 
 	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/fn"
@@ -499,4 +501,144 @@ func DUint32Option(r io.Reader, val interface{}, buf *[8]byte, l uint64) error {
 		return nil
 	}
 	return tlv.NewTypeForDecodingErr(val, "*fn.Option[uint32]", l, l)
+}
+
+// UrlSliceOptionEncoder encodes a URL option. If the value is not set, or the
+// slice is empty, we'll encode it as a zero-length record. But that means that
+// the type and length fields will still be encoded, which is different from the
+// record not being present at all. If the distinction should be made (e.g. to
+// not re-encode old records that didn't have that field at all with the new
+// zero-length record), the caller needs to handle that by conditionally
+// including or not including the record.
+func UrlSliceOptionEncoder(w io.Writer, val any, buf *[8]byte) error {
+	if t, ok := val.(*fn.Option[[]url.URL]); ok {
+		return fn.MapOptionZ(*t, func(value []url.URL) error {
+			numValues := uint64(len(value))
+			if numValues == 0 {
+				return nil
+			}
+
+			err := tlv.WriteVarInt(w, numValues, buf)
+			if err != nil {
+				return err
+			}
+
+			for _, addr := range value {
+				addrBytes := []byte(addr.String())
+				err := asset.InlineVarBytesEncoder(
+					w, &addrBytes, buf,
+				)
+				if err != nil {
+					return err
+				}
+			}
+
+			return nil
+		})
+	}
+	return tlv.NewTypeForEncodingErr(val, "*fn.Option[[]url.URL]")
+}
+
+// UrlSliceOptionDecoder decodes a URL option.
+func UrlSliceOptionDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
+	if l > MaxNumCanonicalUniverseURLs*MaxCanonicalUniverseURLLength {
+		return tlv.ErrRecordTooLarge
+	}
+
+	if t, ok := val.(*fn.Option[[]url.URL]); ok {
+		if l == 0 {
+			*t = fn.None[[]url.URL]()
+
+			return nil
+		}
+
+		numValues, err := tlv.ReadVarInt(r, buf)
+		if err != nil {
+			return err
+		}
+
+		if numValues > MaxNumCanonicalUniverseURLs {
+			return tlv.ErrRecordTooLarge
+		}
+
+		urls := make([]url.URL, 0, numValues)
+		for i := uint64(0); i < numValues; i++ {
+			var urlBytes []byte
+			err := asset.InlineVarBytesDecoder(
+				r, &urlBytes, buf,
+				MaxCanonicalUniverseURLLength,
+			)
+			if err != nil {
+				return err
+			}
+
+			addr, err := url.ParseRequestURI(string(urlBytes))
+			if err != nil {
+				return err
+			}
+			urls = append(urls, *addr)
+		}
+
+		*t = fn.Some(urls)
+
+		return nil
+	}
+	return tlv.NewTypeForDecodingErr(val, "*fn.Option[[]url.URL]", l, l)
+}
+
+// PublicKeyOptionEncoder encodes a public key option. If the value is not set,
+// we'll encode it as a zero-length record. But that means that the type and
+// length fields will still be encoded, which is different from the record not
+// being present at all. If the distinction should be made (e.g. to not
+// re-encode old records that didn't have that field at all with the new
+// zero-length record), the caller needs to handle that by conditionally
+// including or not including the record.
+func PublicKeyOptionEncoder(w io.Writer, val any, buf *[8]byte) error {
+	if t, ok := val.(*fn.Option[btcec.PublicKey]); ok {
+		return fn.MapOptionZ(*t, func(value btcec.PublicKey) error {
+			if value == emptyKey {
+				return nil
+			}
+
+			ptr := &value
+			return asset.CompressedPubKeyEncoder(w, &ptr, buf)
+		})
+	}
+	return tlv.NewTypeForEncodingErr(val, "*fn.Option[btcec.PublicKey]")
+}
+
+// PublicKeyOptionDecoder decodes a public key option.
+func PublicKeyOptionDecoder(r io.Reader, val any, buf *[8]byte,
+	l uint64) error {
+
+	if l > btcec.PubKeyBytesLenCompressed {
+		return tlv.ErrRecordTooLarge
+	}
+
+	if t, ok := val.(*fn.Option[btcec.PublicKey]); ok {
+		if l == 0 {
+			*t = fn.None[btcec.PublicKey]()
+
+			return nil
+		}
+
+		var result *btcec.PublicKey
+		err := asset.CompressedPubKeyDecoder(r, &result, buf, l)
+		if err != nil {
+			return err
+		}
+
+		if result == nil || *result == emptyKey {
+			*t = fn.None[btcec.PublicKey]()
+
+			return nil
+		}
+
+		*t = fn.Some(*result)
+
+		return nil
+	}
+	return tlv.NewTypeForDecodingErr(
+		val, "*fn.Option[btcec.PublicKey]", l, l,
+	)
 }
--- a/proof/encoding.go
+++ b/proof/encoding.go
@@ -9,6 +9,7 @@ import (
 	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/taproot-assets/asset"
+	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightningnetwork/lnd/tlv"
 )
 
@@ -462,4 +463,40 @@ func GenesisRevealDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error {
 	}
 
 	return tlv.NewTypeForEncodingErr(val, "GenesisReveal")
+}
+
+// EUint32Option encodes a uint32 option. If the value is not set, we'll encode
+// it as a zero-length record. But that means that the type and length fields
+// will still be encoded, which is different from the record not being present
+// at all. If the distinction should be made (e.g. to not re-encode old records
+// that didn't have that field at all with the new zero-length record), the
+// caller needs to handle that by conditionally including or not including the
+// record.
+func EUint32Option(w io.Writer, val interface{}, buf *[8]byte) error {
+	if t, ok := val.(*fn.Option[uint32]); ok {
+		return fn.MapOptionZ(*t, func(value uint32) error {
+			return tlv.EUint32T(w, value, buf)
+		})
+	}
+	return tlv.NewTypeForEncodingErr(val, "*fn.Option[uint32]")
+}
+
+// DUint32Option decodes a uint32 option.
+func DUint32Option(r io.Reader, val interface{}, buf *[8]byte, l uint64) error {
+	if t, ok := val.(*fn.Option[uint32]); ok {
+		if l == 0 {
+			*t = fn.None[uint32]()
+			return nil
+		}
+
+		var newVal uint32
+		if err := tlv.DUint32(r, &newVal, buf, l); err != nil {
+			return err
+		}
+
+		*t = fn.Some(newVal)
+
+		return nil
+	}
+	return tlv.NewTypeForDecodingErr(val, "*fn.Option[uint32]", l, l)
 }
--- a/proof/meta.go
+++ b/proof/meta.go
@@ -8,7 +8,9 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"net/url"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightningnetwork/lnd/tlv"
@@ -46,6 +48,14 @@ const (
 	// define when minting assets. Since the uint64 max value has 19 decimal
 	// places we will allow for a max of 12 decimal places.
 	MaxDecDisplay = uint32(12)
+
+	// MaxNumCanonicalUniverseURLs is the maximum number of canonical
+	// universe URLs that can be set.
+	MaxNumCanonicalUniverseURLs = 16
+
+	// MaxCanonicalUniverseURLLength is the maximum length of the canonical
+	// universe URL.
+	MaxCanonicalUniverseURLLength = 255
 )
 
 var (
@@ -86,6 +96,35 @@ var (
 	// ErrDecDisplayMissing is returned if the decimal display key is
 	// not present in a JSON object.
 	ErrDecDisplayMissing = errors.New("decimal display field missing")
+
+	// ErrCanonicalUniverseInvalid is returned if the canonical universe
+	// URL is invalid.
+	ErrCanonicalUniverseInvalid = errors.New(
+		"canonical universe URL invalid",
+	)
+
+	// ErrTooManyCanonicalUniverseURLs is returned if the number of
+	// canonical universe URLs exceeds the maximum.
+	ErrTooManyCanonicalUniverseURLs = fmt.Errorf(
+		"too many canonical universe URLs, max %d",
+		MaxNumCanonicalUniverseURLs,
+	)
+
+	// ErrCanonicalUniverseURLTooLong is returned if the canonical universe
+	// URL is too long.
+	ErrCanonicalUniverseURLTooLong = fmt.Errorf(
+		"canonical universe URL too long, max %d characters",
+		MaxCanonicalUniverseURLLength,
+	)
+
+	// ErrDelegationKeyEmpty is returned if the delegation key is empty.
+	ErrDelegationKeyEmpty = errors.New("delegation key is empty")
+
+	// ErrDelegationKeyNotOnCurve is returned if the delegation key is not
+	// on the curve.
+	ErrDelegationKeyNotOnCurve = errors.New(
+		"delegation key is not on curve",
+	)
 )
 
 // MetaReveal is an optional TLV type that can be added to the proof of a
@@ -109,6 +148,22 @@ type MetaReveal struct {
 	// this value is not zero, then the decimal display is also added as a
 	// field to the JSON object for backward compatibility.
 	DecimalDisplay fn.Option[uint32]
+
+	// UniverseCommitments indicates that the asset group this asset belongs
+	// to will create and push universe commitments to the canonical
+	// universe. A universe commitment is a "proof of inventory" that
+	// commits the issuer's current total asset balance (sum of all mints
+	// minus burns or ignored assets) on-chain.
+	UniverseCommitments bool
+
+	// CanonicalUniverses is a list of URLs of the canonical (approved,
+	// authoritative) universe where the asset minting and universe
+	// commitment proofs will be pushed to.
+	CanonicalUniverses fn.Option[[]url.URL]
+
+	// DelegationKey is the public key that is used to verify universe
+	// commitment related on-chain outputs and proofs.
+	DelegationKey fn.Option[btcec.PublicKey]
 
 	// UnknownOddTypes is a map of unknown odd types that were encountered
 	// during decoding. This map is used to preserve unknown types that we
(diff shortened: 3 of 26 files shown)


ffranr:
I've been thinking more about universe commitments. I think we should get further along in the design before deciding on a universe commitment asset metadata TLV field. It might be wise to split out the universe commitment field from this PR.

ffranr (commented):
I haven't reviewed these changes but:

I've been thinking more about universe commitments. I think we should get further along in the design of that feature before deciding on a universe commitment asset metadata TLV field. It might be wise to split out the universe commitment field from this PR. We can catchup on this.

guggero:
@ffranr I've removed the `UniverseCommitments` boolean field as you suggested.

tapdb/sqlc/migrations/000028_asset_meta_decimal_display.up.sql
```diff
@@ -0,0 +1,5 @@
+-- We add a set of new minting/universe related columns. Those will all be
+-- added to the meta record as odd (optional) TLV fields. Therefore, they will
+-- not be set on older assets, which is why we need to make them nullable here.
+ALTER TABLE assets_meta ADD COLUMN meta_decimal_display INTEGER;
+ALTER TABLE assets_meta ADD COLUMN meta_canonical_universe VARCHAR;
```
ffranr, line 5: `meta_canonical_universe` entries will be size limited on TLV record p2p limit, right?

ffranr (changes requested):
The canonical universe check might hinder us unnecessarily.

Great to see unit tests covering the previous metadata formulation :+1:

proof/meta.go
```diff
@@ -100,6 +100,14 @@ type MetaReveal struct {
 	// Data is the committed data being revealed.
 	Data []byte
 
+	// DecimalDisplay is the decimal display value of the asset. This is
+	// used to determine the number of decimal places to display when
+	// presenting the asset amount to the user. If this field is not
+	// explicitly encoded in the TLV, the default value of 0 is used. If the
+	// meta type is JSON then the decimal display is also added as a field
+	// to the JSON object.
+	DecimalDisplay *uint32
```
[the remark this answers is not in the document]
guggero: I turned it into an option. We need to have it optional, otherwise we can't distinguish between the record not being there (e.g. when calculating the meta hash for old assets) and the default value of 0. But for new assets we'll always encode the default value of 0.

proof/meta_test.go
```diff
@@ -23,6 +25,7 @@ var (
 	)
 )
 
+// TestValidateMetaReveal tests the validation of a MetaReveal.
```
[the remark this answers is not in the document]
guggero: Added.

tapdb/sqlc/migrations/000028_asset_meta_decimal_display.up.sql
guggero: Good catch. It was not. Also added length validation in the `MetaReveal.Validate()` method.

tapgarden/seedling.go
```diff
@@ -204,6 +210,25 @@ func (c Seedling) validateGroupKey(group asset.AssetGroup,
...
+	if anchorMeta != nil && anchorMeta.CanonicalUniverse != nil {
+		anchorCanonicalURL = anchorMeta.CanonicalUniverse.String()
+	}
+
+	if seedlingCanonicalURL != anchorCanonicalURL {
+		return fmt.Errorf("seedling canonical universe URL does not "+
+			"match group anchor: %v, %v", seedlingCanonicalURL,
+			anchorCanonicalURL)
+	}
+
```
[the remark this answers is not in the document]
ffranr: > This will also help make sure the user doesn't forget to set the URL by accident. Or has a typo on a subsequent mint.

Isn't that precisely why we should allow updating the URL in a subsequent tranche? What’s our current plan if a user mints an asset on-chain and later discovers the URL is incorrect? (We've encountered a similar issue with the proof courier URL in tap addresses.)

Maybe we should consider a universe server (if that's the URL scheme) connection check as part of minting.

From my perspective, when a client attempts to evaluate a mint and has access to a single asset ID within a given group, they also have three potential canonical universe URLs:  

1. The canonical URL of the tranche they are evaluating.  
2. The genesis canonical URL.  
3. The default URL hardcoded into `tapd`.  

IMO `tapd` should attempt to sync with these URLs in the given order.

fn/option.go
```diff
@@ -58,6 +58,12 @@ func ElimOption[A, B any](o Option[A], b func() B, f func(A) B) B {
 	return b()
 }
 
+// Unwrap is used to extract the value from an option. If the option is empty,
+// then the zero value of the type that parameterizes the option is returned.
+func (o Option[A]) Unwrap() (A, bool) {
```
Roasbeef, line 63: Makes it easier to do unsafe access (eg: ignoring the second arg), but we basically already have that exposed with `UnwrapToPtr` as is 🤷

itest/asset_meta_test.go
```diff
@@ -167,31 +167,61 @@ func testMintAssetWithDecimalDisplayMetaField(t *harnessTest) {
...
+	}
+	thirdAssetReq := &mintrpc.MintAssetRequest{Asset: thirdAsset}
+	thirdAssets := MintAssetsConfirmBatch(
+		t.t, t.lndHarness.Miner().Client, t.tapd,
+		[]*mintrpc.MintAssetRequest{thirdAssetReq},
+	)
+
+	require.Len(t.t, thirdAssets, 1)
+	require.NotNil(t.t, thirdAssets[0].DecimalDisplay)
+	require.EqualValues(
```
Roasbeef, line 224: Should we also assert that the dec display is _still_ encoded in the asset meta JSON here (still getting through the PR).

Roasbeef: Also what should be the behavior if the user sets the field manually in the JSON meta? What about if that value differs from the one here in the proto.

proof/meta.go
```diff
@@ -232,6 +243,12 @@ func (m *MetaReveal) GetDecDisplay() (map[string]interface{}, uint32, error) {
 		return nil, 0, nil
 	}
 
+	// If the decimal display is set as the new TLV value, we can use that
+	// directly.
+	if decimalDisplay, isSet := m.DecimalDisplay.Unwrap(); isSet {
+		return nil, decimalDisplay, nil
```
Roasbeef, line 277: Before returning, should we do a sanity check that:
  * The value isn't already encoded in the asset meta JSON
  * That if it is then it matches this value

?

proof/records.go
```diff
@@ -43,8 +44,9 @@ const (
 	TapscriptProofTapPreimage2 tlv.Type = 3
 	TapscriptProofBip86        tlv.Type = 4
 
-	MetaRevealEncodingType tlv.Type = 0
-	MetaRevealDataType     tlv.Type = 2
+	MetaRevealEncodingType   tlv.Type = 0
+	MetaRevealDataType       tlv.Type = 2
+	MetaRevealDecimalDisplay tlv.Type = 5
```
Roasbeef: Should this actually be an even value? As if you don't understand it, then you won't be able to reconstruct the meta hash properly.

proof/records.go
```diff
@@ -362,6 +364,59 @@ func MetaRevealDataRecord(data *[]byte) tlv.Record {
 	)
 }
 
+func MetaRevealDecimalDisplayRecord(
+	decimalDisplay *fn.Option[uint32]) tlv.Record {
```
Roasbeef, line 371: Why a _pointer_ to the option vs just the option?

rpcserver.go
```diff
@@ -494,64 +494,46 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 
 	var seedlingMeta *proof.MetaReveal
 
-	// If a custom decimal display is set, the meta type must also be set to
-	// JSON.
+	// If a custom decimal display is set, we require the AssetMeta to be
+	// set. That means the user has to at least specify the meta type.
```
Roasbeef, line 498: What about assuming it's just the opaque type if the dec display is set? Given that by default (opaque asset type, not JSON), we won't encode this in the asset meta.

rpcserver.go
```diff
@@ -494,64 +494,46 @@ func (r *rpcServer) MintAsset(ctx context.Context,
...
 	if req.Asset.DecimalDisplay != 0 && req.Asset.AssetMeta == nil {
-		return nil, fmt.Errorf("decimal display requires JSON asset " +
+		return nil, fmt.Errorf("decimal display requires asset " +
 			"metadata")
 	}
 
+	// Decimal display doesn't really make sense for collectibles.
+	if req.Asset.DecimalDisplay != 0 &&
```
Roasbeef, line 505: So we'll reject this on RPC, but still allow the possibility on the proof/protocol level?

proof/meta.go
```diff
@@ -110,6 +122,11 @@ type MetaReveal struct {
 	// field to the JSON object for backward compatibility.
 	DecimalDisplay fn.Option[uint32]
 
+	// CanonicalUniverse is the URL of the canonical (approved,
+	// authoritative) universe where the asset minting and universe
+	// commitment proofs will be pushed to.
+	CanonicalUniverse fn.Option[url.URL]
```
Roasbeef, line 128: What about instead lifting this to the issuance proof itself? We can require that this be signed by the group key so it can't just be swapped out by a 3rd party. 

This way, one can update this filed with a new issuance event (or some sort of off-chain proof), and not be locked in from the very start (DNS addr changes, etc). 

Thinking about it further, we can even tie this into the unified universe commitment structure. We can add a 4th commitment/field, that can be updated via a spend on chain to distribute new meta/param updates to all the verifiers.

Roasbeef: Another idea is we just put an authentication/delegation key here. This key can be used in the pre-commitment output, and then can also be used to delegate authentication capabilities to another key. Such abilities can include updating the canonical universe value.

Roasbeef (changes requested):

tapgarden/seedling.go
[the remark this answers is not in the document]
guggero: I see your point. Not sure about the connection check, could be a bit of a hurdle (perhaps you first want to make sure the mint succeeds before setting up the universe server?), I'd suggest adding that to the PR that adds the CLI/RPC flags for the field in the first place.

Good idea about syncing in order and with decreasing priority if there are multiple sources for a potential canonical universe URL. Going to remove the check for now then.

fn/option.go
guggero: Yeah, I agree. But what other pattern would you recommend for "if the option is set, return that value from the current function, if not, continue"?

Without any "unsafe" operations, it would be as clunky as this (unless I'm missing something):
```
func doSomething(foo fn.Option[bar]) (bar, error) {
    var (
        priorityBar bar
        fooSet bool
    )
    foo.WhenSome(func(b bar) {
        priorityBar = b
        fooSet = true
    })

    if fooSet {
       return priorityBar, nil
    }

    // Find another source for bar...
}
```

proof/records.go
guggero: No, even would mean you have to understand it or error out. With the "unknown odd types" functionality, you'll be able to construct the meta hash correctly, even if you don't understand each value. That was the whole point of allowing those unknown odd types. You can't decode their content if you haven't upgraded. But you can still calculate the hash of the TLV stream as a whole correctly.

proof/records.go
guggero: Those `...Record()` functions are also used for decoding. So we need to be able to **write** to the parameter, that's why it's always a pointer.
We should upgrade all those records to the new TLV pattern (using `tlv.RecordT`), but that's a bigger refactor that should be its own PR.

rpcserver.go
guggero: Hmm, yeah, I guess we could do that now. But that would mean that both the meta data blob and type would be empty/unset and the meta hash would just be TLV record of the decimal display?

rpcserver.go
guggero: Yeah. We only validate the meta data at creation time. At proof validation time we don't validate anything other than the hash of the reveal matching the committed meta hash.
And IMO that's probably okay, as the meta data fields aren't necessarily things we need to enforce at the protocol level.

proof/meta.go
guggero: Also see discussion here: https://github.com/lightninglabs/taproot-assets/pull/1335#discussion_r1939010717

I guess at this point the field in the meta reveal would be more of a "universe hint" than the "authoritative" universe. The use for it would mostly be bootstrapping:
1. New `tapd` comes online, only syncs to the default universe (e.g. the hard coded one)
2. User wants to learn about grouped asset X, turns on issuance proof sync for just asset X
3. `tapd` syncs the issuance proofs for asset X, sees universe URL in meta reveal of those syncs
4. `tapd` turns on universe commitment sync for asset X using the universe URL(s) found in the meta reveals

But maybe you have a different flow in mind... Should we just remove the field then if we're not really sure yet how we're going to use it?

> Another idea is we just put an authentication/delegation key here.

We are going to use an authenticated key for the pre-commitment key. So perhaps we could also include the universe URL in the actual universe commitment proof, where it would be more "authoritative" and could be updated. But IMO that URL can't **only** be in the universe commitment proof itself, otherwise you we have a bootstrap issue (you can only get the universe commitment proofs from the universe that then tells you what URL to use to get universe commitment proofs...)

guggero: > We can require that this be signed by the group key so it can't just be swapped out by a 3rd party.

Not sure if that's a good idea UX wise. In case of a cold wallet mint, you'd already need to sign at least 2 PSBTs per mint event (one for the group key VM TX, one spending the previous universe commitment output). So needing to sign yet another transaction just makes things clunky IMO.

proof/records.go
Roasbeef: Yep, I was mistaken. I forgot about the addition where we'll just add the unknown TLV types to the buffer before we serialize.

proof/meta.go
Roasbeef: > I guess at this point the field in the meta reveal would be more of a "universe hint" than the "authoritative" universe.

The way I've been viewing things is that this is the _primary_ Universe that should be checked on spend to obtain the pre-image of the commitment data, which is this new unified commitment. It should be what people check tomake sure they have all the data, but ofc this data can still be mirrored elsewhere. 

> tapd syncs the issuance proofs for asset X, sees universe URL in meta reveal of those syncs
> tapd turns on universe commitment sync for asset X using the universe URL(s) found in the meta reveals

So with my suggestion, rather than seeing it in the meta, they'd see it in a singed section in the proof metadata. To go a step further, we might also augment the genesis virtual tx to commit to this data. 

One thing that we still haven't spec'd out, is the RPC API for fetching this new unified commitment root, the sub-trees and the leaves of those sub-trees. The "normal" Universes, may not offer this specialized API, though to be maximally useful, they should mirror it from the canonical universe. 

> So perhaps we could also include the universe URL in the actual universe commitment proof, where it would be more "authoritative" and could be updated

This sounds similar to the idea I mentioned earlier about having a param sub-tree/commitment. This is effectively authenticated, as it can only be updated by being able to spend the pre-commitment tree. 

However, as verifiers will first sync the normal issuance proofs, we need to provide them with enough information to then be able to locate + authenticate the canonical Universe. The issuance proofs can include a signed payload that contains this bootstrap + discovery information (including what key to use to verify. 

> But IMO that URL can't only be in the universe commitment proof itself, otherwise you we have a bootstrap issue (you can only get the universe commitment proofs from the universe that then tells you what URL to use to get universe commitment proofs...)

I don't see how a bootstrap issue arises. You use the existing normal Universes to fetch the issuance proofs, etc. But these may not have burn, and the new meta/param tree we've been ideating. To be able to properly verify the commitment spends on chain, you look to this canonical Universe.

Roasbeef: > Not sure if that's a good idea UX wise. In case of a cold wallet mint, you'd already need to sign at least 2 PSBTs per mint event (one for the group key VM TX, one spending the previous universe commitment output). So needing to sign yet another transaction just makes things clunky IMO.

Agreed. So one way we can avoid that clunkiness is allowing a delegation key to be presented in the initial meta. This key is then the one that signs the set of booststrap params, and can even include yet another delegation key in the set of params, etc.

Roasbeef:
Bringing this outside the thread so it doesn't get lost. 

Here's my idea w.r.t what should go into the meta, how bootstrapping would work, and this new bootstrap params field:
  * ### What goes inside the meta:
      * A bool that is used to communicate to verifiers that the issuer is opting into the new on-chain commitment rules. 
      * A delegation key. This is used to create+verify the initial pre-commitment, and also to sign a new payload that goes into the issuance proof itself. 
   * ### New issuance proof payload:
       * We add a new TLV field to the issuance proof, call this `commitment_params` for now (random name, open to other suggestions). 
       * This can be updated with every new issuance proof. 
          * Alternatively, this goes in the unified commitment, either as a new sub-tree, or is the first use of the "reserved" opaque hash I've mentioned before. I think @guggero suggested a form of this above. 
       * To start with, it includes the following fields:
          * `canonical_universe_url`: same thing we've added here in this PR. Can be updated with another issuance, or a param spend of the current on-chain commitment.
             * The param spend is an off-chain event that triggers an on-chain spend, just like updating the ignore list. 
          * `delegate_key`: specifies which key to use for the pre-commitment creation+verification. If blank, use the group key. 
              * We can support many of these, or add some other field later re a threshold or w/e. May also be a musig key, etc. 
       * If it goes into the issuance proof, then we also include a signature of this payload:
          * If it has a delegation key, you verify with that, otherwise use the group key (allows for the external flow with minimal round trips for issuance). 
          * To avoid hot swap attacks, the meta may need to commit to this field. If this is the case, then it isn't too far off to just include this new payload into the meta itself.

guggero:
Trying to summarize this thread and combine it with the internal design doc.
This is my current understanding of what we'd want to do:

## Initial mint
 * Asset metadata
   * `UniverseCommitments: bool`
   * `IssuanceDelegationKey: PubKey` (currently called `UniCommitTRKey` in #1325)
 * Minting batch
   * Change is locked to `IssuanceDelegationKey` (`UniCommitTRKey`) or group key if not present
 * Issuance proof
   * Commitment params (hashed and signed with `IssuanceDelegationKey` or group key if not present)
     * `CanonicalUniverseURL: string`
     * `CommitmentDelegationKey: PubKey` --> Delegation key for universe commitment proof
     * `PreCommitmentIndex: uint32` --> Output index of the pre-commitment (minting batch change output)
   * `CommitmentParamsSig: [64]byte` 
 * Universe commitment proof
   * Commitment data (burns, ignores and so on, hashed and signed with `CommitmentDelegationKey` or group key if ont present)
 * Universe commitment TX
   * Spends pre-commitment output
   * Output is locked to group key + commitment leaf
   * Change output goes back to wallet

## Subsequent mint
* Asset metadata
    * `UniverseCommitments: bool` (enforced if group anchor has enabled)
    * `IssuanceDelegationKey: PubKey` (currently called `UniCommitTRKey` in #1325) --> could be a new key or re-use the same as in the group anchor
* Minting batch
    * Change is locked to `IssuanceDelegationKey` (`UniCommitTRKey`) or group key if not present
* Issuance proof
    * Commitment params (hashed and signed with `IssuanceDelegationKey` or group key if not present)
        * `CanonicalUniverseURL: string` --> Could be a new URL
        * `CommitmentDelegationKey: PubKey` --> Delegation key for universe commitment proof, could be a new one or re-use the same as in the group anchor
        * `PreCommitmentIndex: uint32` --> Output index of the pre-commitment (minting batch change output)
    * `CommitmentParamsSig: [64]byte`
* Universe commitment proof
    * Commitment data (burns, ignores and so on, hashed and signed with `CommitmentDelegationKey` or group key if ont present)
* Universe commitment TX
    * Spends group anchor's universe commitment output (needs to sign with group internal key)
    * Output is locked to group internal key + commitment leaf
    * Change output goes back to wallet

My main question here is: Do we really need two different delegation keys? One in the asset metadata and then a new one in the issuance proof's commitment params? I only differentiate between those two keys because @Roasbeef mentioned two delegation keys above. But I think this could be the same one? So just one delegation key defined in the asset meta that signs for both the issuance proof payload and the universe commitment payload.
But each minting tranche could have its new delegation key, so it can be changed with every mint, along with the canonical universe URL.

Second question: Do we still want to lock the actual universe commitment transaction output to the group key (plus a leaf that contains the commitment roots)? Or would that also lock to the delegated key?

ffranr:
I think I pretty much agree with Oli's summary. Except that I don't think we should ever lock any uni commitment related output with the group key. Because of the external cold key signing requirements.

And I don't see the point in two different delegation keys also.

I think the canonical universe URL should go in the issuance proof (signed payload) and then we should be able to update it in the universe commitment proof. So I don't think the bootstrapping issue is a blocker.

> Thinking about it further, we can even tie this into the unified universe commitment structure. We can add a 4th commitment/field, that can be updated via a spend on chain to distribute new meta/param updates to all the verifiers.

I agree with this sort of idea. I think we should generalise the whole universe commitment notion to something that the minter can use to issue general updates (update canonical uni URL, update ignore list, update total issuance, etc.)

ffranr:
Just thinking outloud:

I think what we need in the asset meta data is:
* the delegation key
* bool flag for: does the issuance proof contain a payload signed with that delegation key. (we need this encase a melicious universe server strips the signed payload out completely. This is a more general UniverseCommitments=true)

And then in the signed payload in the issuance proof we can have:
* uni pre-commitment output index: fn.Option[uint]
* (first/latest) canonical uni URL
* asset ID/anchor first prev out txid to ensure sig is unique?

And then the signature field for the payload.
So it's like a general signed payload with a general delegated key. And all we're doing is linking the delegation key to the group key. And we assure that the payload sig is unique to that issuence
using the asset ID as the nonce

Roasbeef:
Latest iteration:
  * Move the universe URL back into the meta:
     * Potentially add a revoke field 
     * Potentially make it slice so can add many 
  * Pick GKR V1 or the normal TAP commitment structure for the universe commitment:
     * Need to ensure its unique

itest/asset_meta_test.go
guggero: Added a check in the itest and more validation to make sure the two values match when creating the meta reveal.

proof/meta.go
guggero: I added a check for that in the RPC server, so for when we create the reveal. If someone uses a different implementation and allows encoding of different values in the JSON vs. the TLV field, IMO there's not much we can do about it and the TLV value should probably take precedence.

guggero:
I've updated the code according to the latest state as discussed offline.

proof/encoding.go
```diff
@@ -499,3 +501,141 @@ func DUint32Option(r io.Reader, val interface{}, buf *[8]byte, l uint64) error {
...
+// UrlSliceOptionEncoder encodes a URL option. If the value is not set, or the
+// slice is empty, we'll encode it as a zero-length record. But that means that
+// the type and length fields will still be encoded, which is different from the
+// record not being present at all. If the distinction should be made (e.g. to
+// not re-encode old records that didn't have that field at all with the new
+// zero-length record), the caller needs to handle that by conditionally
+// including or not including the record.
+func UrlSliceOptionEncoder(w io.Writer, val any, buf *[8]byte) error {
+	if t, ok := val.(*fn.Option[[]url.URL]); ok {
+		return fn.MapOptionZ(*t, func(value []url.URL) error {
```
Roasbeef, line 514: 😎

proof/meta.go
```diff
@@ -152,7 +207,44 @@ func (m *MetaReveal) Validate() error {
...
+	if err != nil {
+		return err
+	}
+
+	return fn.MapOptionZ(m.DelegationKey, func(key btcec.PublicKey) error {
+		if key == emptyKey {
+			return ErrDelegationKeyEmpty
+		}
+
+		if !key.IsOnCurve() {
```
Roasbeef, line 242: 👍

tapdb/asset_meta.go
```diff
@@ -11,13 +15,48 @@ import (
...
+
 	if len(meta.MetaDataHash) == 0 {
-		return fn.None[proof.MetaReveal]()
+		return fn.None[proof.MetaReveal](), nil
+	}
+
+	var canonicalUniverse fn.Option[[]url.URL]
+	if len(meta.MetaCanonicalUniverses) > 0 {
+		urls := strings.Split(
+			string(meta.MetaCanonicalUniverses), "\x00",
```
Roasbeef, line 28: Any particular reason `\x00` was chosen here? Still getting through the commits, but iiuc we chose to store them as a single strings vs create a new foreign key table via a new table back to the meta. The latter seems a bit more flexible, though these fields are sort of immutable from the PoV of the protocol as is. 

Perhaps we just should verify that it's possible to create a migration in the future that splits on this character, and then makes the new tables with the updated refs?

tapdb/sqlc/migrations/000028_asset_meta_tlv_fields.up.sql
```diff
@@ -0,0 +1,12 @@
...
+-- not be set on older assets, which is why we need to make them nullable here.
+ALTER TABLE assets_meta ADD COLUMN meta_decimal_display INTEGER;
+
+ALTER TABLE assets_meta ADD COLUMN meta_universe_commitments BOOL;
+
+-- If there's more than one URL, we're going to join them using the 0x00 control
+-- character (which is invalid in a URL itself).
+ALTER TABLE assets_meta ADD COLUMN meta_canonical_universes BLOB;
+
+ALTER TABLE assets_meta ADD COLUMN meta_delegation_key BLOB;
```
Roasbeef, line 12: Restrict size to be 33 bytes?

Roasbeef:
```
CHECK(length(data) <= 33)
```

tapdb/sqlc/migrations/000028_asset_meta_tlv_fields.up.sql
```diff
@@ -0,0 +1,12 @@
+-- We add a set of new minting/universe related columns. Those will all be
+-- added to the meta record as odd (optional) TLV fields. Therefore, they will
+-- not be set on older assets, which is why we need to make them nullable here.
+ALTER TABLE assets_meta ADD COLUMN meta_decimal_display INTEGER;
+
+ALTER TABLE assets_meta ADD COLUMN meta_universe_commitments BOOL;
+
+-- If there's more than one URL, we're going to join them using the 0x00 control
+-- character (which is invalid in a URL itself).
+ALTER TABLE assets_meta ADD COLUMN meta_canonical_universes BLOB;
```
Roasbeef, line 10: Similar to comment below, should we also restrict this to the the max size, or we'll rely on the RPC parsing/validation logic for that?

tapdb/sqlutils.go
```diff
@@ -48,6 +49,17 @@ func sqlInt32[T constraints.Integer](num T) sql.NullInt32 {
 	}
 }
 
+// sqlOptInt32 turns an option of a numerical integer type into the NullInt32
+// that sql/sqlc uses when an integer field can be permitted to be NULL.
+func sqlOptInt32[T constraints.Integer](num fn.Option[T]) sql.NullInt32 {
```
Roasbeef, line 54: 👍

Roasbeef (approved):
LGTM 🦂

tapdb/asset_meta.go
guggero: Ah, yeah, I meant to comment on that in the commit message but forgot.
I just chose any character that is illegal in a `url.URL` during parsing. And apparently all control characters (anything in ASCII before the space character).
Since the chance of there being multiple values IMO is quite low, making an extra query every time seems quite a performance hit vs. the "ugliness" of having delimiter-separated entries in a column.

And yes, it seems possible to extract this into its own table in the future using SQL only: https://chatgpt.com/share/67ab12e6-e8a0-8003-a26e-40264ae887a9

tapdb/sqlc/migrations/000028_asset_meta_tlv_fields.up.sql
guggero: We do have the parsing logic that makes sure there can't be more than 16 URLs and that each URL isn't longer than 255 characters. But I still added the length limit here, makes sense to me.

fn/option.go
ffranr: I think we should add this function to `fn`:
```golang
// Deref safely dereferences a pointer.
// If the pointer is nil, it returns the zero value of type T and an error.
func Deref[T any](ptr *T) (T, error) {
	if ptr == nil {
		var zero T
		return zero, errors.New("nil pointer dereference")
	}
	return *ptr, nil
}
```

It would be useful outside this use case.

And then we can just do:

```golang
func doSomething(foo fn.Option[bar]) (bar, error) {
   if v := foo.UnwrapToPtr(); v != nil {
       return fn.Deref(v)
   }

   // Find another source for bar...
}
```

Eventually we could get to a place where pointer dereference only happens in `Deref`. And it's pretty clear that `Deref` will never dereference a `nil` pointer.

ffranr: Maybe we can also add DerefPanic which just panics when the pointer is nil. That way it's clear in our code what we expect when dereferencing pointers. And we always dereference with either `Deref` or `DerefPanic`.

ffranr: IMO unsafe code is only a problem when it's all over the codebase. If it's just in a few select well though out places it isn't a problem.

guggero: I agree that we should discuss this. But can we please not block this PR on it?
Do you want me to leave things in and add a TODO or revert to an "ugly" but safe method, then we fix things later on?

guggero: I've removed the `Unwrap` method from `fn` and solved this in a different way (`UnwrapOr()`).

proof/encoding.go
```diff
@@ -499,3 +501,141 @@ func DUint32Option(r io.Reader, val interface{}, buf *[8]byte, l uint64) error {
...
 }
+
+// UrlSliceOptionEncoder encodes a URL option. If the value is not set, or the
+// slice is empty, we'll encode it as a zero-length record. But that means that
+// the type and length fields will still be encoded, which is different from the
+// record not being present at all. If the distinction should be made (e.g. to
+// not re-encode old records that didn't have that field at all with the new
+// zero-length record), the caller needs to handle that by conditionally
+// including or not including the record.
+func UrlSliceOptionEncoder(w io.Writer, val any, buf *[8]byte) error {
```
ffranr, line 512:
```
If the value is not set, or the
// slice is empty, we'll encode it as a zero-length record.
```

Why are we doing it that way? That's not what I would expect from encoding an `Option`. If it's `None` then the record shouldn't be included.

What's the point in encoding a `None` with as a zero length record?

tapdb/sqlc/migrations/000028_asset_meta_tlv_fields.up.sql
```diff
@@ -0,0 +1,21 @@
+-- We add a set of new minting/universe related columns. Those will all be
+-- added to the meta record as odd (optional) TLV fields. Therefore, they will
+-- not be set on older assets, which is why we need to make them nullable here.
+ALTER TABLE assets_meta ADD COLUMN meta_decimal_display INTEGER;
+
+ALTER TABLE assets_meta ADD COLUMN meta_universe_commitments BOOL;
```
ffranr, line 6: I think there should be a comment here explaining what this column means.

tapgarden/seedling.go
```diff
@@ -181,6 +181,14 @@ func (c Seedling) validateGroupKey(group asset.AssetGroup,
 			"group asset type %v", group.Genesis.Type)
 	}
 
+	return validateAnchorMeta(c.Meta, anchorMeta)
+}
+
+// validateAnchorMeta checks that the metadata of the seedling matches that of
+// the group anchor, if there is a group anchor.
+func validateAnchorMeta(seedlingMeta *proof.MetaReveal,
```
ffranr, line 189: We're not restricting seedlings to genesis canonical URLs here anymore, which I think is correct and what we need.

proof/encoding.go
guggero: The thing is, once you're in the `...Encoder` function, you can't decide to _not_ include a record anymore. That's why we only conditionally add records in `EncodeRecords()`. That's what I mean with the part in the comment that says `the caller needs to handle that by conditionally including or not including the record.`.

All of this goes away once we use the new TLV encoding way, when this would become a `tlv.OptionalRecordT`.

ffranr: Why can't we not encode within the `...Encoder`?

Does this have strange implications for reproducing old proofs in the decode/encode cycle?

guggero: We can decide to not encode any _payload_ within the `...Encoder` function. But when this function is called, then the T and L of TLV are already written to the stream (assuming we didn't skip the record altogether as we do in https://github.com/lightninglabs/taproot-assets/blob/6c68c9f81fec2dd5967d454918a7c904a74cf652/proof/meta.go#L492). So once we are in the encoder function, we cannot decide whether to include the record or not. We can only decide whether we write any data or cause it to be a zero-length record (but the type and length would still be written).

ffranr: The other way I think about this is why not just have `UrlSliceEncoder` and decoder and then relectively call based on some/none. So that the encoder/decoder functionality (and its doc comment) doesn't have to express the None case complexity.

PR #1338, opened 2025-01-29
tapd: fix mutex deadlock in universe cache

guggero:
We tried to acquire the read lock after already having acquired the write lock.


GeorgeTsagk:
This [CI failure](https://github.com/lightninglabs/taproot-assets/actions/runs/13032474953/job/36354713020?pr=1338#step:4:213) is being fixed by #1339

gijswijs (approved):
LGTM :tada:

PR #1339, opened 2025-01-29
tapchannel: add specifier in property test rfq generator

GeorgeTsagk:
We include the asset specifier in the generated quotes of the property test framework. This is needed as now the aux invoice HTLC interceptor performs an extra check, according to which the asset ID in the rfq and the wire records must match.

PR #1299 introduced this check


guggero (approved):
tACK, a couple dozen iterations of `make flake-unit pkg=tapchannel` didn't fail anymore, whereas on master it would fail within a couple of iterations.

Thanks for the fix!

PR #1345, opened 2025-01-31
tapdb: fixes another issue with the syncer cache

guggero:
This time we activate the syncer cache by default in the integration tests to avoid further issues.

The actual bug fix is quite trivial, the wrong root node was inserted into the cache.


GeorgeTsagk (approved):

gijswijs (approved):
Uni, Multi...
Potato, potato!
:tada:

PR #1348, opened 2025-02-03
Unlock Coins on Anchor TX Broadcast Failure  

ffranr:
This PR ensures that asset coins are properly unlocked if an error occurs during the anchor transaction broadcast process. Previously, failed broadcasts could leave coins locked, causing send attempts to be stuck.  

Key changes:  
- Unlock asset coins before returning an error when publishing an anchor transaction.  
- Ensure inputs are unlocked if an error occurs before reaching the broadcast complete state.  
- Introduce `SendStateBroadcastComplete` to better define when a transfer can be canceled.  
- Improve logging by reporting the final fee rate of the anchoring transaction and logging the wallet's minimum relay fee for better debugging.

### DIFF
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -1491,6 +1491,9 @@ func (p *ChainPorter) unlockInputs(ctx context.Context, pkg *sendPackage) {
 		return
 	}
 
+	// TODO(ffranr): Make use of CheckMempoolAccept to ensure we don't
+	//  unlock inputs for transactions that are in the mempool.
+
 	// If we haven't even attempted to broadcast yet, we're still in a state
 	// where we give feedback to the user synchronously, as we haven't
 	// created an on-chain transaction that we need to await confirmation.
@@ -1530,6 +1533,9 @@ func (p *ChainPorter) unlockInputs(ctx context.Context, pkg *sendPackage) {
 			log.Warnf("Unable to unlock input %v: %v", op, err)
 		}
 	}
+
+	// TODO(ffranr): Remove pending asset transfer and chain tx from the
+	//  database.
 }
 
 // logPacket logs the virtual packet to the debug log.
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -1178,6 +1178,9 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
 				"minimum relay fee: %w", err)
 		}
 
+		log.Infof("Min relay fee: %v",
+			minRelayFee.FeePerKVByte().String())
+
 		// If the fee rate is below the minimum relay fee, we'll
 		// bump it up.
 		if feeRate < minRelayFee {
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -1391,6 +1391,18 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
 				"transaction %v: %w", txHash, err)
 
 		case err != nil:
+			// If the error is due to the min relay fee not being
+			// met, we'll unlock the inputs we locked for this
+			// transfer before returning the error.
+			if strings.Contains(
+				err.Error(), "min relay fee not met",
+			) {
+				p.unlockInputs(ctx, &currentPkg)
+			}
+
+			// We exercise caution by not unlocking the inputs in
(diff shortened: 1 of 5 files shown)


tapfreighter/chain_porter.go
```diff
@@ -1474,7 +1491,7 @@ func (p *ChainPorter) unlockInputs(ctx context.Context, pkg *sendPackage) {
 	// sanity-check that we have known input commitments to unlock, since
 	// that might not always be the case (for example if another party
 	// contributes inputs).
-	if pkg.SendState < SendStateStorePreBroadcast &&
+	if pkg.SendState < SendStateBroadcastComplete &&
```
gijswijs, line 1494: If we remove the lease on assets when we are in state `SendStateStorePreBroadcast` or `SendStateBroadcast`, we don't clean up the parcel. Which seems wrong to me. This wasn't fully the case in the old situation because we only released coins when the parcel wasn't persisted to disk. So at least upon restart, the parcel was lost, and the assets would truly be free again. In this PR, we get situations were we remove the lease on coins, but still have parcels persisted to disk, that are loaded after a restart.

gijswijs: A similar problem occurs a few lines down (and wasn't introduced by this PR, but should still be handled)

https://github.com/lightninglabs/taproot-assets/blob/e0c96614f7f5b667f4890f6467f6f793dce99111/tapfreighter/chain_porter.go#L1520

In states `SendStateStorePreBroadcast` and `SendStateBroadcast` (which are the final two states that call into `unlockInputs`) we try to release the inputs of the _anchor tx_ but we this leaves us with a situation where we still have the parcel. This PR changes the behavior and at least releases the assets, but we are still stuck with a pending parcel.

Also, we should inform the user what to do if `p.cfg.Wallet.UnlockInput` fails, because now he has a potential asset burning tx in his lnd node that we couldn't cancel.

gijswijs (changes requested):
I think we need to rethink this at a few levels. I think this might be a good time to introduce a failing state that properly handles releasing coins, releasing acnhor tx inputs and cleans up any lingering parcels. I've made some comments throughout where I see issues with this.

There are some points where I think the specific error message we get back from `WalletKit` informs us on how to proceed. I've mentioned this throughout as well. 

Thirdly, we could also use a RPC endpoint that allows you to cancel stuck parcels. But mayby that's something for another PR.

tapfreighter/chain_porter.go
```diff
@@ -1388,10 +1391,24 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
 				"transaction %v: %w", txHash, err)
 
 		case err != nil:
+			// Unlock the inputs we locked for this transfer before
+			// returning the error.
+			p.unlockInputs(ctx, &currentPkg)
```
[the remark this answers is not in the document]
ffranr: We could explicitly check whether the error message contains "min relay fee not met" to specifically address this coin-locking issue when the wallet is `bitcoind`.  

I agree that unlocking for a general error might be too risky. In a future update, we could consider checking the mempool from `unlockInputs`.

tapfreighter/chain_porter.go
ffranr: I agree that we should cleanup the pending parcel as well.

> Also, we should inform the user what to do if `p.cfg.Wallet.UnlockInput` fails, because now he has a potential asset burning tx in his lnd node that we couldn't cancel.

I don't think we currently call `unlockInputs` in state broadcast complete or beyond. So LND won't have a burning tx.

ffranr: Perhaps, in the worst case, if we don't cancel the pending parcel then upon resumption, the asset coins would have been spent and the transaction would fail because the UTXO would have been spent.

Roasbeef:
I think before broadcast, we can call into `CheckMempoolAccept`, then only broadcast if that returns that it's likely to be accepted. 

Here's an example in `lnd`: https://github.com/lightningnetwork/lnd/blob/6bf895aeb929fb64265bfb226f796392921d8d3a/lnwallet/btcwallet/btcwallet.go#L1928-L1957

litbot-9000:

guggero:
!lightninglabs-deploy mute

jtobin:
The underlying issue here still seems to exist, so this would be worth following up on.

jtobin:
Closing. The underlying issue is now captured in #2295, which also proposes a natural fix based on the new watcher architecture.

PR #1352, opened 2025-02-04
Add `sync` loadtest

### LINKED ISSUE #1313: [feature]:  Add `synctest` to our loadtest suite
## Description

We currently have loadtests for minting and sending assets. One other aspect of tapd we want to loadtest is the universe. The purpose of a `synctest` is to fetch data from a universe server based on certain configurations.

To keep the topology simple, we won't have on-demand tapd nodes performing the client requests to the universe, instead we'll use lite golang client code that performs the RPC requests directly. We don't really care about handling the responses, we just want to make sure we perform them. The purpose is to emulate a full-sync by multiple clients.

The configuration could include things like whether we're doing a full sync or not, number of clients, or number of syncs per client.

GeorgeTsagk:
## Description

This PR adds a new type of loadtest, which is focused around the syncing endpoints.

We add a new test that emulates a full `AssetRoots` sync by a configurable number of clients. The page size may also be configured (although I believe we better use the max).

Since the dummy syncers don't care about the actual data and all we're trying to do is stress the universe server, we don't even use a tap related client. We hit the public REST endpoint for simplicity and ignore the response's details.

Also added a TODO for future expansions and configurations for this test. For now, this test is only restricted to performing an AssetRoots sync from multiple clients (from the PoV of the server).

Closes #1313

### DIFF
--- a/itest/loadtest/sync_test.go
+++ b/itest/loadtest/sync_test.go
@@ -1,0 +1,157 @@
+package loadtest
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"sync"
+	"testing"
+
+	tap "github.com/lightninglabs/taproot-assets"
+	"github.com/lightninglabs/taproot-assets/taprpc/universerpc"
+	"github.com/lightninglabs/taproot-assets/universe"
+	"github.com/stretchr/testify/require"
+)
+
+// syncTest checks that the universe server can handle multiple requests to its
+// public universe endpoints. The number of clients is configurable.
+func syncTest(t *testing.T, ctx context.Context, cfg *Config) {
+	alice := initAlice(t, ctx, cfg)
+
+	// Let's start by logging the aggregate universe stats.
+	res, err := alice.UniverseStats(ctx, &universerpc.StatsRequest{})
+	require.NoError(t, err)
+
+	t.Logf("Universe Aggregate Stats: %+v", res)
+
+	// We'll use this wait group to block until all clients are done.
+	var wg sync.WaitGroup
+
+	// We dispatch a client sync for the configured number of clients.
+	for i := range cfg.SyncNumClients {
+		switch cfg.SyncType {
+		case "simplesyncer":
+			wg.Add(1)
+			go simpleSyncer(t, ctx, cfg, i, &wg)
+
+		case "rest":
+			wg.Add(1)
+			go syncAssetRootsREST(t, ctx, cfg, i, &wg)
+		}
+	}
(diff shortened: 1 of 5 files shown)


itest/loadtest/sync_test.go
```diff
@@ -0,0 +1,114 @@
...
+		// Break if we reached the end. This is signalled by retrieving
+		// less entities than what was defined as the max limit,
+		// meaning that there's nothing left to consume.
+		if len < limit {
+			break
+		}
+
+		offset += limit
+	}
+
```
GeorgeTsagk, line 90: linter complains here, will fix with first round of feedback

Roasbeef:
Here's an idea for how we can get broad coverage in the sync tests _without_ needing to spin up an entire lnd instance (don't commit to disk, etc):
  * We can spin up a new syncer instance: 
  https://github.com/lightninglabs/taproot-assets/blob/74f8e485ee562fe311b6e9759f53cd884d663f9a/universe/syncer.go#L29-L65
    * The only _real_ interface we need to specify there is the `NewRemoteDiffEngine` closure. 
       * We have a default implementation that uses the RPC system, which can be found here: https://github.com/lightninglabs/taproot-assets/blob/74f8e485ee562fe311b6e9759f53cd884d663f9a/universe_rpc_diff.go
    * For `LocalDiffEngine` and `LocalRegistrar` we can have a stub implementation that does nothing. So we just say we don't have any of the roots/information (local diff engine), and then we don't actually write to disk (no registrar).  

With this, we have a light weight way of generating _real_ load of a historical sync (issuance, transfers, or both). If we want to add some more incremental load (new assets minted, thundering herd to sync), then we can modify `LocalDiffEngine` to pretend like it already knows about leafs+keys.

guggero (approved):
tACK, LGTM :tada:

itest/loadtest/utils.go
```diff
@@ -311,3 +314,80 @@ func stringToAssetType(t string) taprpc.AssetType {
 		return taprpc.AssetType_NORMAL
 	}
 }
+
+// noopBaseUni is a dummy implementation of the universe.DiffEngine and
+// universe.LocalRegistrar interfaces. This is meant to be used by the simple
+// syncer used in the sync loadtest. As we don't care about persistence and we
+// always want to do a full sync, we always return an empty root node to trigger
+// a sync.
+type noopBaseUni struct{}
```
Roasbeef, line 323: 😎

Roasbeef (approved):
LGTM 🦖

Roasbeef:
Failing the linter, but should be g2g once that's resolved.

PR #1357, opened 2025-02-06
rpcserver: better rfq channel selection

GeorgeTsagk:
Previously, callers of rfqChannel would just get the first channel from the list of candidates. This is sufficient, but we could easily optimise this depending on the intention. Depending on whether we want to receive or send over the channel, we return the channel with the best remote or local balance accordingly.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -6576,8 +6576,9 @@ func (r *rpcServer) checkPeerChannel(ctx context.Context, peer route.Vertex,
 		}
 
 		// If we don't get an error here, it means we do have an asset
-		// channel with the peer.
-		_, err = r.rfqChannel(ctx, assetID, &peer)
+		// channel with the peer. The intention doesn't matter as we're
+		// just checking whether a channel exists.
+		_, err = r.rfqChannel(ctx, assetID, &peer, NoIntention)
 		if err != nil {
 			return fmt.Errorf("error checking asset channel: %w",
 				err)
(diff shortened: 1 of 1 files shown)


rpcserver.go
```diff
@@ -7825,9 +7848,40 @@ func (r *rpcServer) rfqChannel(ctx context.Context, id asset.ID,
...
+					bestBalance.assetInfo.RemoteBalance {
+
+					bestBalance = b
+				}
+			})
+
+		case SendIntention:
+			// If the intention is to send, return the channel with
+			// the biggest local balance.
+			fn.ForEach(assetBalances, func(b channelWithAsset) {
```
guggero, line 7871: Just to check my understanding: When sending, this code path only really matters if the `peerPubKey` isn't set and you have multiple channels with a given asset to different peers. Then we select the peer with the best inbound liquidity to us (the actual channel is ignored on the send related call sites).
But if `peerPubKey` is specified, this is basically a no-op, since it doesn't matter which channel of a peer we choose, as we only need the pubKey anyway.

guggero (commented):
Nice and easy fix for better invoice peer selection!

rpcserver.go
```diff
@@ -6587,8 +6587,9 @@ func (r *rpcServer) checkPeerChannel(ctx context.Context, peer route.Vertex,
 		}
 
 		// If we don't get an error here, it means we do have an asset
-		// channel with the peer.
-		_, err = r.rfqChannel(ctx, assetID, &peer)
+		// channel with the peer. The intention doesn't matter as we're
+		// just checking whether a channel exists.
+		_, err = r.rfqChannel(ctx, assetID, &peer, ReceiveIntention)
```
ffranr, line 6592: use `NoIntention` here

rpcserver.go
```diff
@@ -7825,9 +7848,40 @@ func (r *rpcServer) rfqChannel(ctx context.Context, id asset.ID,
...
 	// channels, it means we have multiple channels with the same asset and
-	// the same peer. So we just return the first channel.
+	// the same peer.
 	case len(assetBalances) >= 1:
-		return &assetBalances[0], nil
+		// Initialize to first element.
+		bestBalance := assetBalances[0]
+
+		switch intention {
+		case ReceiveIntention:
```
ffranr, line 7857: instead of nesting another `switch` here, I think you should return early above if `len(assetBalances) == 0`

ffranr (approved):

Issue #1361, opened 2025-02-06
[bug]: `acceptpricedeviationppm` not enforced


ZZiigguurraatt:
`sample-tapd.conf` shows the default of `experimental.rfq.acceptpricedeviationppm=50000` (https://github.com/lightninglabs/taproot-assets/blob/9202f08c68c1b4d97b0f7521784983ce920226b9/sample-tapd.conf#L363-L366), which is equal to 5%. This is defined at https://github.com/lightninglabs/taproot-assets/blob/74f8e485ee562fe311b6e9759f53cd884d663f9a/rfq/negotiator.go#L22-L26 .

I have a price oracle running. Here's what happens when the price oracle is queried directly outside of `tapd`:

```
In [7]: QueryPriceOracle(asset_id_str=TheAssetID.hex(),hostname='price_oracle',transaction_type_str='PURCHASE')
time: 1738876742.2311115
request:
subject_asset {
  asset_id_str: "0c5d1e47c3f2302c8fb15164989de56f9e1dfee2357788e2fce4551cbd6b23dd"
}
payment_asset {
  asset_id: "\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000"
}


client received:
ok {
  asset_rates {
    subjectAssetRate {
      coefficient: "11000000000"
    }
    paymentAssetRate {
      coefficient: "100000000000"
      scale: 9
    }
    expiry_timestamp: 1738877042
  }
}


In [8]: QueryPriceOracle(asset_id_str=TheAssetID.hex(),hostname='price_oracle',transaction_type_str='SALE')
time: 1738876748.3672757
request:
transaction_type: SALE
subject_asset {
  asset_id_str: "0c5d1e47c3f2302c8fb15164989de56f9e1dfee2357788e2fce4551cbd6b23dd"
}
payment_asset {
  asset_id: "\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000"
}


client received:
ok {
  asset_rates {
    subjectAssetRate {
      coefficient: "10000000000"
    }
    paymentAssetRate {
      coefficient: "100000000000"
      scale: 9
    }
    expiry_timestamp: 1738877048
  }
}
```

You can see that the `PURCHASE` rate is `10%` higher than the `SALE` price.

[...]

PR #1365, opened 2025-02-10
RFQ accept price qualify bug

ffranr:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1361  

This PR corrects a bug where we called the wrong price oracle endpoints when evaluating the price from a quote accept message sent in response to a quote request message.  

Additionally, we enhance the mock price oracle to allow method call hooking.  

See the commit messages for details on the original issue and the correct approach.

### DIFF
--- a/rfq/negotiator_test.go
+++ b/rfq/negotiator_test.go
@@ -1,0 +1,385 @@
+package rfq
+
+import (
+	"testing"
+	"time"
+
+	"github.com/lightninglabs/taproot-assets/asset"
+	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/rfqmath"
+	"github.com/lightninglabs/taproot-assets/rfqmsg"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+// testCaseIncomingSellAccept is a test case for the handling of an incoming
+// sell accept message.
+type testCaseIncomingSellAccept struct {
+	// name is the name of the test case.
+	name string
+
+	// incomingSellAcceptRate is the rate in the incoming sell accept
+	// message.
+	incomingSellAcceptRate rfqmsg.AssetRate
+
+	// priceOracleAskPrice is the rate returned by the price oracle.
+	priceOracleAskPrice rfqmsg.AssetRate
+
+	// acceptPriceDeviationPpm is the acceptable price deviation in ppm.
+	acceptPriceDeviationPpm uint64
+
+	// quoteRespStatus is the expected status of the quote check.
+	quoteRespStatus fn.Option[QuoteRespStatus]
+}
+
+// assertIncomingSellAcceptTestCase asserts the handling of an incoming sell
+// accept message for a test case.
+func assertIncomingSellAcceptTestCase(
+	t *testing.T, tc testCaseIncomingSellAccept) {
+
+	// Create a mock price oracle.
+	mockPriceOracle := &MockPriceOracle{}
+
+	// Register an expected call and response for price oracle method
+	// QueryAskPrice.
+	mockPriceOracle.On(
+		"QueryAskPrice", mock.Anything, mock.Anything,
+		mock.Anything, mock.Anything, mock.Anything,
+	).Return(
+		&OracleResponse{
+			AssetRate: tc.priceOracleAskPrice,
+		}, nil,
+	)
+
+	// Define sell request and sell accept messages.
+	var (
+		assetSpecifier = asset.NewSpecifierFromId(asset.ID{1, 2, 3})
+		peerID         = route.Vertex{1, 2, 3}
+		msgID          = rfqmsg.ID{1, 2, 3}
+	)
+
+	sellRequest := rfqmsg.SellRequest{
+		Peer:           peerID,
+		ID:             msgID,
+		AssetSpecifier: assetSpecifier,
+		PaymentMaxAmt:  lnwire.MilliSatoshi(1000),
+		AssetRateHint:  fn.None[rfqmsg.AssetRate](),
+	}
+
+	sellAccept := rfqmsg.SellAccept{
+		Request:   sellRequest,
+		AssetRate: tc.incomingSellAcceptRate,
+	}
+
+	// Create the negotiator.
+	errChan := make(chan error, 1)
+	negotiator, err := NewNegotiator(NegotiatorCfg{
+		PriceOracle:             mockPriceOracle,
+		OutgoingMessages:        make(chan rfqmsg.OutgoingMsg, 1),
+		AcceptPriceDeviationPpm: tc.acceptPriceDeviationPpm,
+		ErrChan:                 errChan,
+	})
+	require.NoError(t, err)
+
+	// Define the finalise callback function.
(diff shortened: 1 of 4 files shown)


ZZiigguurraatt:
One thing I realized when testing this fix is that when selling an asset, with `experimental.rfq.acceptpricedeviationppm=50000`, SellPrice>=(1-.05)*BuyPrice instead of BuyPrice<=(1+.05)*SellPrice . Wondering if we should make this clearer in the documentation?

ZZiigguurraatt:
Can we get a more informative error message than

`peer 03bb1c09a26fbb1b5799b0263eed8e516fb5fbe141d93104ed19211a429f6065bd sent back an invalid quote, status: INVALID_ASSET_RATES`

That says that the rate was over the `experimental.rfq.acceptpricedeviationppm` limit? Like I can't tell if it is invalid because the quote isn't for a large enough amount of assets, the expiry time is past, etc..

ZZiigguurraatt:
> One thing I realized when testing this fix is that when selling an asset with `SendPayment`, with `experimental.rfq.acceptpricedeviationppm=50000`, SellPrice>=(1-.05)*BuyPrice instead of BuyPrice<=(1+.05)*SellPrice . Wondering if we should make this clearer in the documentation?


Note, when buying an asset with `AddInvoice`, it works the same as above and SellPrice>=(1-.05)*BuyPrice. Given how SendPayment works, I would have thought it would be the opposite and be BuyPrice<=(1+.05)*SellPrice. I guess the seller is always the one that trigger's the error and the seller's error is just propagated back to the buyer when `AddInvoice` is run?

When using `AddInvoice` does the buyer give the edge node/seller a "price" or just a "price hint" when the RFQ initially starts? When using `AddInvoice` and the seller rejects, does the seller have any opportunity to make a counter offer? Just trying to understand so I know I'm testing properly.

rfq/oracle.go
```diff
@@ -428,21 +428,6 @@ func NewMockPriceOracle(expiryDelay,
 	}
 }
 
-// NewMockPriceOracleSatPerAsset creates a new mock price oracle with a
-// specified satoshis per asset rate.
-//
-// TODO(ffranr): Remove in favour of NewMockPriceOracle.
-func NewMockPriceOracleSatPerAsset(expiryDelay uint64,
```
guggero, line 435: Hmm, so this was actually broken in 4e291e5e and we didn't realize back then. But the meaning between `rfqCfg.MockOracleAssetsPerBTC` and `rfqCfg.MockOracleSatsPerAsset` is different, so one of these functions is now incorrect, probably this one (since a rate is assets per BTC, so we'd need to convert things from sats per asset to that rate, which will be imprecise).
Or we completely remove the whole config flag for the sats per asset way of configuring the mock oracle.

rfq/negotiator_test.go
```diff
@@ -0,0 +1,349 @@
...
+	negotiator, err := NewNegotiator(NegotiatorCfg{
+		PriceOracle:             mockPriceOracle,
+		OutgoingMessages:        make(chan rfqmsg.OutgoingMsg, 1),
+		AcceptPriceDeviationPpm: tc.AcceptPriceDeviationPpm,
+		ErrChan:                 errChan,
+	})
+	require.NoError(t, err)
+
+	// Define the finalise callback function.
+	finalise := func(
```
guggero, line 87: nit: format like other function definitions instead? Her and in other test below.

```
	finalise := func(msg rfqmsg.BuyAccept,
		event fn.Option[InvalidQuoteRespEvent]) {
```

rfq/negotiator_test.go
```diff
@@ -0,0 +1,349 @@
...
+	"github.com/lightningnetwork/lnd/routing/route"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+// testCaseIncomingSellAccept is a test case for the handling of an incoming
+// sell accept message.
+type testCaseIncomingSellAccept struct {
+	// Name is the name of the test case.
+	Name string
```
guggero, line 21: nit: do all of these fields need to be exported? Since the main struct isn't?

rfq/negotiator_test.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// TestHandleIncomingBuyAccept tests the handling of an incoming buy accept
+// message.
+func TestHandleIncomingBuyAccept(t *testing.T) {
+	defaultQuoteExpiry := time.Now().Add(time.Hour)
+
+	testCases := []testCaseIncomingBuyAccept{
+		{
+			Name: "accept price within bounds",
```
guggero, line 302: Should we also add a test case where the price is within the bounds, but just barely? Instead of only being exactly the same? Here and in the test above.

guggero (commented):
Changes to the RFQ code look good to me. Nice investigation and fixes!

I noticed we broke something with the mock oracle in an earlier PR and I think we should fix that again or remove things completely.

rfq/negotiator_test.go
ffranr: I'll make them all non-exported.

rfq/oracle.go
ffranr: I'll leave the old constructor in and I'll fix the bug. We need to convert sats per asset to asset per btc correctly.

ffranr: I've added a TODO in this PR to fix this oversight.

rfq/negotiator_test.go
ffranr: Done.

But see this PR for an explanation of the numbers: https://github.com/lightninglabs/taproot-assets/pull/1377

rfq/negotiator_test.go
```diff
@@ -0,0 +1,349 @@
...
+	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/rfqmath"
+	"github.com/lightninglabs/taproot-assets/rfqmsg"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+// testCaseIncomingSellAccept is a test case for the handling of an incoming
```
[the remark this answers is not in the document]
ffranr: Good idea. I might look into that later. We have rapid tests for the tolerance method already. They might be handy for the expiry.

ffranr:
@ZZiigguurraatt Thank you for your comments, I've gone through them. I think we will need to consider them separately from this bug. I'll think about what you've said some more tomorrow.

rfq/negotiator_test.go
```diff
@@ -0,0 +1,385 @@
...
+	defaultQuoteExpiry := time.Now().Add(time.Hour)
+
+	testCases := []testCaseIncomingSellAccept{
+		{
+			name: "accept price just within bounds 1",
+			incomingSellAcceptRate: rfqmsg.AssetRate{
+				Rate:   rfqmath.NewBigIntFixedPoint(1000, 0),
+				Expiry: defaultQuoteExpiry,
+			},
+			priceOracleAskPrice: rfqmsg.AssetRate{
```
guggero, line 140: Nice, seems like we caught some edge cases in the tolerance calculation with these test cases?

guggero (approved):

rfq/negotiator_test.go
ffranr: From my POV, it forced us to be clear about the methodology we use to check tolerance.

PR #1371, opened 2025-02-11
make+itest: don't use locally installed chantools

guggero:
A new version of chantools is available. But we want to use a specific version, so we make sure to always build it for the itest to not interfere with any other version that might already be installed on the system.


ffranr (approved):
Nice! That's how it should have been from the beginning.

PR #1375, opened 2025-02-11
scripts: exclude ./itest/chantools from YAML Go version linter

ffranr:
After running integration tests locally, the chantools repo is pulled, causing the linter to mistakenly analyze that repo. This commit updates the linter exclusion list to prevent this issue.

### DIFF
--- a/scripts/check-go-version-yaml.sh
+++ b/scripts/check-go-version-yaml.sh
@@ -63,11 +63,35 @@ fi
 
 target_go_version="$1"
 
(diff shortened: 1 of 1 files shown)


ffranr:
We might introduce exclusion pattern matching later, but an overly broad pattern could unintentionally exclude important files.  

Perhaps the exclusion list should support both individual files and entire directories?

Roasbeef (approved):
LGTM ☘️

PR #1377, opened 2025-02-12
rfqmath: Improve WithinTolerance doc and handle zero coefficient edge case

ffranr:
This PR enhances the documentation for `FixedPoint.WithinTolerance` to clarify that the tolerance check is max-based, meaning it is applied relative to the larger of the two values. This ensures that a 100% tolerance (1,000,000 PPM) always results in a match for any two nonzero values.  

Additionally, the function now returns an error if either value is zero, as tolerance is undefined in such cases.  

#### Alternative Tolerance Strategies For Consideration
These are alternative strategies that we don't use in our code.
  
- **Bid-based (stricter):** Uses the smaller value as the reference. This enforces a tighter threshold but does not guarantee a match at 100% tolerance. It may also be more intuitive in some bid-ask spread scenarios, so it is something to keep in mind.  
- **Midpoint-based (balanced):** Uses the average of both values. This smooths out variations but is less intuitive for bid-ask spread comparisons.  

The max-based approach was chosen for its predictability and leniency in bid-ask scenarios, ensuring a 100% tolerance always results in a match while still maintaining a meaningful spread constraint at lower tolerances.

### DIFF
--- a/rfq/negotiator.go
+++ b/rfq/negotiator.go
@@ -652,9 +652,30 @@ func (n *Negotiator) HandleIncomingBuyAccept(msg rfqmsg.BuyAccept,
 		tolerance := rfqmath.NewBigIntFromUint64(
 			n.cfg.AcceptPriceDeviationPpm,
 		)
-		acceptablePrice := msg.AssetRate.Rate.WithinTolerance(
+		acceptablePrice, err := msg.AssetRate.Rate.WithinTolerance(
 			assetRate.Rate, tolerance,
 		)
+		if err != nil {
+			// The tolerance check failed. We will return without
+			// calling the quote accept callback.
+			err = fmt.Errorf("failed to check tolerance: %w", err)
+			log.Errorf("Error checking tolerance: %v", err)
+
+			// Construct an invalid quote response event so that we
+			// can inform the peer that the quote response has not
+			// validated successfully.
+			invalidQuoteRespEvent := NewInvalidQuoteRespEvent(
+				&msg, InvalidAssetRatesQuoteRespStatus,
+			)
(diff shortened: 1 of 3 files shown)


ffranr:
cc @ZZiigguurraatt

Perhaps you might find the extra documentation useful in explaining some of your testing outcomes?

rfq/negotiator.go
```diff
@@ -769,9 +791,31 @@ func (n *Negotiator) HandleIncomingSellAccept(msg rfqmsg.SellAccept,
...
-		acceptablePrice := msg.AssetRate.Rate.WithinTolerance(
+		acceptablePrice, err := msg.AssetRate.Rate.WithinTolerance(
 			assetRate.Rate, tolerance,
 		)
+		if err != nil {
+			// The tolerance check failed. We will return without
+			// calling the quote accept callback.
+			err = fmt.Errorf("failed to check tolerance: %w", err)
+			log.Errorf("Error checking tolerance: %v", err)
+			n.cfg.ErrChan <- err
```
guggero, line 802: I'm not sure we should shut down here (which happens when we send an error on the main channel). Otherwise a "rogue" oracle can cause your node to go down.

guggero (commented):
LGTM with the caveat of not sending the error on the main err chan.

guggero (approved):
I removed the send on the main error channel. I know we also send there in case we can't call the oracle at all, which might be more okay (as the node can't function without an oracle). But perhaps we need to look at error/shutdown conditions more generally in a follow-up PR.

gijswijs (approved):
LGTM :tada:

PR #1382, opened 2025-02-12
Support group keys for RFQ negotiation flows

GeorgeTsagk:
## Description

This PR introduces the minimum functionality for two nodes to negotiate and register a buy or sell quote, which identifies the underlying asset via a group key instead of an ID.

It includes a basic itest which lets Alice & Bob negotiate a quote based on a specifier which only includes a group key.

Related to https://github.com/lightninglabs/taproot-assets/issues/1028, but does not close it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1382)

### DIFF
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -438,6 +438,16 @@ func (s *Specifier) UnwrapToPtr() (*ID, *btcec.PublicKey) {
 	return s.UnwrapIdToPtr(), s.UnwrapGroupKeyToPtr()
 }
 
+// AssertNotEmpty checks whether the specifier is empty, returning an error if
+// so.
+func (s *Specifier) AssertNotEmpty() error {
+	if !s.HasId() && !s.HasGroupPubKey() {
+		return fmt.Errorf("asset specifier is empty")
+	}
+
+	return nil
+}
+
 // Type denotes the asset types supported by the Taproot Asset protocol.
 type Type uint8
 
--- a/itest/rfq_test.go
+++ b/itest/rfq_test.go
@@ -424,6 +424,112 @@ func testRfqAssetSellHtlcIntercept(t *harnessTest) {
 	require.NoError(t.t, err)
 }
 
+// testRfqNegotiationGroupKey checks that two nodes can negotiate and register
+// quotes based on a specifier that only uses a group key.
+func testRfqNegotiationGroupKey(t *harnessTest) {
+	// Initialize a new test scenario.
+	ts := newRfqTestScenario(t)
+
+	// Mint an asset with Alice's tapd node.
+	rpcAssets := MintAssetsConfirmBatch(
+		t.t, t.lndHarness.Miner().Client, ts.AliceTapd,
+		[]*mintrpc.MintAssetRequest{issuableAssets[0]},
+	)
+
+	mintedAssetGroupKey := rpcAssets[0].AssetGroup.TweakedGroupKey
+
+	ctxb := context.Background()
+	ctxt, cancel := context.WithTimeout(ctxb, defaultWaitTimeout)
+	defer cancel()
+
+	// Subscribe to Alice's RFQ events stream.
+	aliceEventNtfns, err := ts.AliceTapd.SubscribeRfqEventNtfns(
+		ctxb, &rfqrpc.SubscribeRfqEventNtfnsRequest{},
+	)
+	require.NoError(t.t, err)
+
+	// Alice sends a sell order to Bob for some amount of the newly minted
+	// asset.
+	askAmt := uint64(42000)
+	sellOrderExpiry := uint64(time.Now().Add(24 * time.Hour).Unix())
+
+	// We first try to add a sell order without specifying the asset skip
+	// flag. That should result in an error, since we only have a normal
+	// channel and not an asset channel.
+	sellReq := &rfqrpc.AddAssetSellOrderRequest{
+		AssetSpecifier: &rfqrpc.AssetSpecifier{
+			Id: &rfqrpc.AssetSpecifier_GroupKey{
+				GroupKey: mintedAssetGroupKey,
+			},
+		},
+		PaymentMaxAmt: askAmt,
+		Expiry:        sellOrderExpiry,
+
+		// Here we explicitly specify Bob as the destination
+		// peer for the sell order. This will prompt Alice's
+		// tapd node to send a request for quote message to
+		// Bob's node.
+		PeerPubKey: ts.BobLnd.PubKey[:],
+
+		TimeoutSeconds: uint32(rfqTimeout.Seconds()),
+	}
+	_, err = ts.AliceTapd.AddAssetSellOrder(ctxt, sellReq)
+	require.ErrorContains(
+		t.t, err, "no asset channel balance found",
+	)
+
+	// Now we set the skip flag and we shouldn't get an error anymore.
+	sellReq.SkipAssetChannelCheck = true
+	_, err = ts.AliceTapd.AddAssetSellOrder(ctxt, sellReq)
+	require.NoError(t.t, err, "unable to upsert asset sell order")
+
+	// Wait until Alice receives an incoming sell quote accept message (sent
+	// from Bob) RFQ event notification.
+	BeforeTimeout(t.t, func() {
+		event, err := aliceEventNtfns.Recv()
+		require.NoError(t.t, err)
+
+		_, ok := event.Event.(*rfqrpc.RfqEvent_PeerAcceptedSellQuote)
+		require.True(t.t, ok, "unexpected event: %v", event)
+	}, rfqTimeout)
+
+	// We now repeat the same flow, where Alice is making a BuyOrderRequest.
+	assetMaxAmt := uint64(1000)
+	buyOrderExpiry := sellOrderExpiry
+
+	buyReq := &rfqrpc.AddAssetBuyOrderRequest{
+		AssetSpecifier: &rfqrpc.AssetSpecifier{
+			Id: &rfqrpc.AssetSpecifier_GroupKey{
+				GroupKey: mintedAssetGroupKey,
+			},
+		},
+		AssetMaxAmt:    assetMaxAmt,
+		Expiry:         buyOrderExpiry,
+		PeerPubKey:     ts.BobLnd.PubKey[:],
+		TimeoutSeconds: uint32(rfqTimeout.Seconds()),
+	}
+
+	_, err = ts.AliceTapd.AddAssetBuyOrder(ctxt, buyReq)
+	require.ErrorContains(
+		t.t, err, "no asset channel balance found",
+	)
+
+	// Now we set the skip flag and we shouldn't get an error anymore.
+	buyReq.SkipAssetChannelCheck = true
+	_, err = ts.AliceTapd.AddAssetBuyOrder(ctxt, buyReq)
+	require.NoError(t.t, err)
+
+	// Wait until Alice receives an incoming buy quote accept message (sent
+	// from Bob) RFQ event notification.
+	BeforeTimeout(t.t, func() {
+		event, err := aliceEventNtfns.Recv()
+		require.NoError(t.t, err)
+
+		_, ok := event.Event.(*rfqrpc.RfqEvent_PeerAcceptedBuyQuote)
+		require.True(t.t, ok, "unexpected event: %v", event)
+	}, rfqTimeout)
+}
+
 // rfqTestScenario is a struct which holds test scenario helper infra.
 type rfqTestScenario struct {
 	testHarness *harnessTest
--- a/rfq/manager.go
+++ b/rfq/manager.go
@@ -558,18 +558,7 @@ func (m *Manager) addScidAlias(scidAlias uint64, assetSpecifier asset.Specifier,
 		return c.PubKeyBytes == peer
 	}, localChans)
 
-	// Identify the correct channel to use as the base SCID for the alias
-	// by inspecting the asset data in the custom channel data.
-	assetID, err := assetSpecifier.UnwrapIdOrErr()
-	if err != nil {
-		return fmt.Errorf("asset ID must be specified when adding "+
-			"alias: %w", err)
-	}
-
-	var (
-		assetIDStr = assetID.String()
-		baseSCID   uint64
-	)
+	var baseSCID uint64
 	for _, localChan := range peerChannels {
 		if len(localChan.CustomChannelData) == 0 {
 			continue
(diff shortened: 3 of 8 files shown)


Roasbeef (changes requested):
Great set of commits! Minimal, yet impactful diff. 

I think I've spotted a bug where we'll start to compare sats balance instead of assets balance when trying to select the "best" channel. We should follow up there with a unit test to confirm the error, then correct it (if it does indeed exist). 

In order to properly test this e2e, I think we may need to expose a group key arg during asset invoice creation. I don't think this requires the full on multi asset group channels (we just want to verify that we can make invoices based no the group key). 

> Reviewed 1 of 1 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1382#-:-OJVzw0k55kbX503y3wv:b-7mncvx)* status: all files reviewed, 3 unresolved discussions (waiting on @GeorgeTsagk)

___
*[`rfq/manager.go` line 963 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1382#-OJVwJ9YEn2tGeeijQvX:-OJVwJ9YEn2tGeeijQvY:bjgiwcz) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/9cdc424eb4ac407a38b42c26d7a64b309acf74e5/rfq/manager.go#L963)):*
> ```Go
> 
> 	// Store the result for future calls.
> 	m.assetIDToGroup.Store(id, groupKeyBytes)
> ```

Not something we need to worry about today, but eventually we'll want have some sort of eviction policy here. 
___
*[`rpcserver.go` line 7865 at r6](https://reviewable.io/reviews/lightninglabs/taproot-assets/1382#-OJVyibp-67EHQFMlcBM:-OJVyibp-67EHQFMlcBN:b-v9etjj) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/4b125be505fed27924b9658311c4ecd05b21d91d/rpcserver.go#L7865)):*
> ```Go
> 		// the best local balance.
> 		fn.ForEach(balances, func(b channelWithSpecifier) {
> 			if b.channelInfo.LocalBalance >
> ```

Is `.LocalBalance` here actually the asset, or sats balance? I ask as it's `lnclient.ChannelInfo`, so I assume it would be a sats balance. 

We should ensure that for the best balance here, we're looking at the _asset balance_ 
___
*[`rpcserver.go` line 7909 at r6](https://reviewable.io/reviews/lightninglabs/taproot-assets/1382#-OJVyxqe8Qom06dtLExH:-OJVyxqe8Qom06dtLExI:b-mgzjry) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/4b125be505fed27924b9658311c4ecd05b21d91d/rpcserver.go#L7909)):*
> ```Go
> 
> 		for assetIdx := range assets {
> 			assetOutput := assets[assetIdx]
> ```

Looks like the inner loop here could be an actual `Filter`, you'd need to contend with the extra return error value, but that's where the `Result` type can be nice when working with higher order functions/types. 
___
*[`itest/rfq_test.go` line 429 at r7](https://reviewable.io/reviews/lightninglabs/taproot-assets/1382#-OJVzNDq6PU0uFaZP2k1:-OJVzNDq6PU0uFaZP2k2:bkzb2b9) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/a76775af0e094ee910de06b9073aefea4652b59a/itest/rfq_test.go#L429)):*
> ```Go
> // testRfqNegotiationGroupKey checks that two nodes can negotiate and register
> // quotes based on a specifier that only uses a group key.
> func testRfqNegotiationGroupKey(t *harnessTest) {
> ```

We should complement this test with one on `litd` that makes RFQ quotes with a group key. 

Think aloud, in order to properly test that, I think we may need to support creating asset invoices with a group key at the very least. 
___
*[`rpcserver.go` line 7886 at r6](https://reviewable.io/reviews/lightninglabs/taproot-assets/1382#-OJVya8d5L51vHDe8SVy:-OJVya8d5L51vHDe8SVz:bgc7vrs) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/28c4ea8e019bb5a52e59401f66b0a5eebf0a0eee/rpcserver.go#L7886)):*
> ```Go
> type channelWithAsset struct {
> 	// assetInfo is the information about one of the assets in a channel.
> 	assetInfo rfqmsg.JsonAssetChanInfo
> ```

Interesting how this seems to duplicate the information in the attribute below it?

asset/asset.go
```diff
@@ -245,6 +245,11 @@ func (i ID) String() string {
 	return hex.EncodeToString(i[:])
 }
 
+// IsEqual returns true if the ID matches the provided ID.
+func (i ID) IsEqual(a ID) bool {
```
guggero, line 249: just a drive-by comment: `asset.ID` is an array, so it supports the `==` operator.

rfqmsg/sell_request.go
```diff
@@ -152,12 +152,10 @@ func NewSellRequestFromWire(wireMsg WireMessage,
...
 func (q *SellRequest) Validate() error {
-	// Ensure that the asset specifier is set.
-	//
-	// TODO(ffranr): For now, the asset ID must be set. We do not currently
-	//  support group keys.
-	if !q.AssetSpecifier.HasId() {
-		return fmt.Errorf("asset id not specified in SellRequest")
+	// Ensure that the asset specifier has either an asset ID or a group
+	// key set.
+	if !q.AssetSpecifier.HasId() && !q.AssetSpecifier.HasGroupPubKey() {
```
guggero, line 157: Maybe add `IsEmpty()` method to specifier? Or `AssertNotEmpty()` that then also returns the error that we can just return here?

guggero (commented):
Just a drive-by review, will take a closer look once marked as ready for review by assigning reviewers.

rfq/manager.go
```diff
@@ -965,6 +966,33 @@ func (m *Manager) GetAssetGroupKey(ctx context.Context,
...
+		return bytes.Equal(group, specifierGK.SerializeCompressed()),
+			nil
+
+	case specifier.HasId():
+		specifierID := specifier.UnwrapIdToPtr()
+
+		return specifierID.IsEqual(id), nil
+	}
+
+	return false, fmt.Errorf("specifier is empty")
```
guggero, line 1005: style nit: I find it easier to read if a `switch` has a `default` case that has the error return. But personal preference I guess.

rfq/manager.go
```diff
@@ -602,8 +614,8 @@ func (m *Manager) addScidAlias(scidAlias uint64, assetSpecifier asset.Specifier,
 	// At this point, if the base SCID is still not found, we return an
 	// error. We can't map the SCID alias to a base SCID.
 	if baseSCID == 0 {
-		return fmt.Errorf("add alias: base SCID not found for asset: "+
-			"%v", assetID)
+		return fmt.Errorf("add alias: base SCID not found for %v",
```
guggero, line 617: I think we need to use `%s` and `&assetSpecifier` to invoke the `.String()` method here.

rpcserver.go
```diff
@@ -7799,60 +7797,60 @@ const (
...
 			"channel balance: %w", err)
 	}
 
-	assetBalances, haveBalance := balances[id]
-	if !haveBalance || len(assetBalances) == 0 {
-		return nil, fmt.Errorf("no asset channel balance found for "+
-			"asset %s", id.String())
+	if len(balances) == 0 {
+		return nil, fmt.Errorf("no asset channel balance found for %s",
+			specifier.String())
```
guggero, line 7814: I think because `.String()` has a pointer receiver, it's enough to just do `&specifier`.

guggero: Can be applied to the other error messages below as well (`%v` and pointer).

rpcserver.go
```diff
@@ -7879,28 +7877,67 @@ func (r *rpcServer) rfqChannel(ctx context.Context, id asset.ID,
...
+
+			var assetID asset.ID
+			copy(assetID[:], assetIDBytes)
+
+			match, err := r.cfg.RfqManager.AssetMatchesSpecifier(
+				ctx, specifier, assetID,
+			)
+			if err != nil {
+				return false, err
+			}
```
guggero, line 7930: This is the exact same code as in the previous commit. Perhaps we should extract that into a re-usable method in the `Manager`:

```
func (m *Manager) ChannelCompatible(ctx context.Context,
	jsonAsset rfqmsg.JsonAssetChanInfo, specifier asset.Specifier) (bool,
	error) {

	gen := jsonAsset.AssetInfo.AssetGenesis
	assetIDBytes, err := hex.DecodeString(
		gen.AssetID,
	)
	if err != nil {
		return false, fmt.Errorf("error decoding asset ID: %w", err)
	}

	var assetID asset.ID
	copy(assetID[:], assetIDBytes)

	match, err := m.AssetMatchesSpecifier(ctx, specifier, assetID)
	if err != nil {
		return false, err
	}

	// TODO(george): Instead of returning the first result,
	// try to pick the best channel for what we're trying to
	// do (receive/send). Binding a baseSCID means we're
	// also binding the asset liquidity on that channel.
	return match, nil
}
```

itest/rfq_test.go
```diff
@@ -424,6 +424,112 @@ func testRfqAssetSellHtlcIntercept(t *harnessTest) {
...
+	_, err = ts.AliceTapd.AddAssetSellOrder(ctxt, sellReq)
+	require.NoError(t.t, err, "unable to upsert asset sell order")
+
+	// Wait until Alice receives an incoming sell quote accept message (sent
+	// from Bob) RFQ event notification.
+	BeforeTimeout(t.t, func() {
+		event, err := aliceEventNtfns.Recv()
+		require.NoError(t.t, err)
+
+		_, ok := event.Event.(*rfqrpc.RfqEvent_PeerAcceptedSellQuote)
```
guggero, line 492: Since we now have the initial request message available in the event, I wonder if we should add the asset specifier to the RFQ quote messages and verify here that the group key is set properly.

See https://github.com/lightninglabs/taproot-assets/blob/7d7cec6bf5938b4b3b06e59c4e608bd26f985146/rpcserver.go#L6889

guggero (commented):
Nice work! Have a refactor/simplification suggestion but nothing major.

Roasbeef (approved):
LGTM :sailboat: 

> Reviewed 1 of 1 files at r6, 8 of 8 files at r8, 2 of 2 files at r9, 2 of 2 files at r10, 1 of 1 files at r11, 1 of 1 files at r12, 1 of 1 files at r13, 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1382#-:-OKYkBqk4PFzBjft6GA5:bwyu0ru)* status: all files reviewed, 15 unresolved discussions (waiting on @GeorgeTsagk and @guggero)

___
*[`rpcserver.go` line 8017 at r13](https://reviewable.io/reviews/lightninglabs/taproot-assets/1382#-OKYjXxXAUzjQbBr083o:-OKYjXxXAUzjQbBr083p:b-8fx2u7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/3715ef86bb0fbd640bffe39f1fd5ace5f000b427/rpcserver.go#L8017)):*
> ```Go
> 		// the best local balance.
> 		fn.ForEach(balances, func(b channelWithSpecifier) {
> 			if b.assetInfo.LocalBalance >
> ```

:+1:

rpcserver.go
```diff
@@ -8067,27 +8069,36 @@ func (r *rpcServer) computeChannelAssetBalance(
...
 
-			channelsByID[assetID] = append(
-				channelsByID[assetID], channelWithAsset{
-					assetInfo:   assetOutput,
-					channelInfo: openChan,
-				},
-			)
+		// Since the assets of the channel passed the above filter,
+		// we're safe to aggregate their info to be represented as a
+		// single entity.
```
Roasbeef, line 8083: Based on this comment, looks like these lines should actually reside within the conditional below.

rfq/manager.go
```diff
@@ -917,6 +935,34 @@ func (m *Manager) RemoveSubscriber(
...
+	// Perform the DB query.
+	group, err := m.cfg.GroupLookup.QueryAssetGroup(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	// If the asset does not belong to a group, return early with no error
+	// or response.
+	if group == nil || group.GroupKey == nil {
+		return nil, nil
```
guggero, line 955: Hmm, this pattern suggests to me that this should probably return `(fn.Option[btcec.PublicKey], error)` instead to avoid accidental nil pointer panics.

rfq/manager.go
```diff
@@ -963,6 +964,67 @@ func (m *Manager) GetAssetGroupKey(ctx context.Context,
...
+	switch {
+	case specifier.HasGroupPubKey():
+		group, err := m.GetAssetGroupKey(ctx, id)
+		if err != nil {
+			return false, err
+		}
+
+		specifierGK := specifier.UnwrapGroupKeyToPtr()
+
+		return group.IsEqual(specifierGK), nil
```
guggero, line 979: Yep, see previous comment, this would nil pointer panic on a non-grouped asset.

rfq/manager.go
```diff
@@ -917,6 +935,34 @@ func (m *Manager) RemoveSubscriber(
 	return nil
 }
 
+// GetAssetGroupKey retrieves the group key of an asset based on its ID.
+func (m *Manager) GetAssetGroupKey(ctx context.Context,
```
guggero, line 937: I think this could be non-exported now, since it's only used through `AssetMatchesSpecifier()` at the moment.

rpcserver.go
```diff
@@ -8067,27 +8069,36 @@ func (r *rpcServer) computeChannelAssetBalance(
...
+		var aggrAssetInfo rfqmsg.JsonAssetChanInfo
+
+		for _, assetInfo := range assetData.Assets {
```
guggero, line 8086: The whole JSON structure is very weird at the moment and I'm refactoring this to be correct in https://github.com/lightninglabs/taproot-assets/pull/1413. Can you please add a TODO here that we fix this?

itest/rfq_test.go
guggero: Okay, but can you add it to your TODO list for another PR please? Since IMO that is now an important information about a quote, whether it was created for an asset ID or group key.

guggero (commented):
Almost there!

guggero (approved):
LGTM :tada:

PR #1386, opened 2025-02-14
tapdb: expand allowed proof types in universe-related tables

Roasbeef:
This commit introduces a migration that expands the allowed values for the "proof_type" column in the universe-related tables by adding new proof types: "burn" and "ignore".

Key changes include:
- Creation of a new table `proof_types` to store valid proof type values. Before this commit, we used a `CHECK` statement. With the way sqlite works, we can't _alter_ the `CHECK` statement, and just dropping the row would lose data.

  So instead, we add a new enum-like table instead. This can be updated in future migrations to add new proof types. We then do a migration to update all tables to use this instead.

- Updates to the `universe_roots`, `federation_global_sync_config`, and `federation_uni_sync_config` tables to accommodate the new proof types. This involves adding a new column `proof_type_ext` that references the `proof_types` table, copying existing data, and renaming columns to maintain the original structure.

- A new test case `TestMigration29` has been added to ensure the migration works as expected, verifying both the migration of existing data and the insertion of new rows with the new proof types.

- The `.gitignore` file has been updated to exclude files matching the pattern `.aider*`.

Additionally, a down migration script has been created to revert these changes if necessary. This ensures that the database schema can be rolled back to its previous state without data loss.


tapdb/testdata/migrations_test_00029_dummy_data.sql
```diff
@@ -0,0 +1,28 @@
...
+INSERT INTO mssmt_roots (namespace, root_hash)
+VALUES ('n1', X'abcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdef');
+
+-- Insert into universe_roots with the original schema.
+INSERT INTO universe_roots (id, namespace_root, asset_id, group_key, proof_type)
+VALUES (1, 'n1', NULL, X'000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f', 'issuance');
+
+-- Insert into federation_global_sync_config.
+INSERT INTO federation_global_sync_config (proof_type, allow_sync_insert, allow_sync_export)
+VALUES ('transfer', 1, 0);
```
ffranr, line 23: Use `true` and `false` instead of integers (`1` and `0`) for PostgreSQL compatibility. SQLite accepts both, but PostgreSQL requires explicit booleans.

Same for query in `TestMigration29`.

tapdb/sqlc/migrations/000029_ignore_burn_universe.down.sql
```diff
@@ -0,0 +1,39 @@
...
+-- Changes:
+-- 1. For federation_global_sync_config, re-create the table using the original schema since
+--    the proof_type column is a primary key. This avoids dropping a primary key column.
+-- 2. For other tables, revert changes via ALTER statements.
+-- 3. Drop the 'proof_types' enum table.
+-- 4. Recreate the 'universe_stats' view using the latest definition to match the reverted schema.
+-- ******************************************************************************************
+
+-- For universe_roots
+ALTER TABLE universe_roots ADD COLUMN proof_type TEXT NOT NULL CHECK(proof_type IN ('issuance', 'transfer'));
```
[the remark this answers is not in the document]
Roasbeef: So I don't think we currently have that rule for SQL queries. Eg: `queries/mssmt.sql`

tapdb/migrations_test.go
```diff
@@ -293,3 +293,156 @@ func TestMigration20(t *testing.T) {
...
+	//   - Insert a unique mssmt_nodes row and a corresponding mssmt_roots
+	//   row with namespace 'ns_burn'.
+	//
+	//   - Then insert a universe_roots row referencing namespace 'ns_burn'
+	_, err = db.ExecContext(ctx, ` 
+	  INSERT INTO mssmt_nodes (
+	      hash_key, l_hash_key, r_hash_key, key, value, sum, namespace
+	  )
+	  VALUES (
+	    X'1111111111111111111111111111111111111111111111111111111111111111',
```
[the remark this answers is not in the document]
Roasbeef: This took me waaay to long to figure out. But the issue was that for the migration set up, we always map the byte literal from to postgres as needed. I added a helper function to do this.

tapdb/sqlc/schemas/generated_schema.sql
```diff
@@ -632,6 +619,10 @@ CREATE TABLE proof_transfer_log (
     time_unix TIMESTAMP NOT NULL
 );
 
+CREATE TABLE proof_types (
```
Roasbeef, line 622: Nice part about this is that you can spot check the _end_ result of a migration.

tapdb/sqlc/schemas/generated_schema.sql
```diff
@@ -389,15 +380,11 @@ CREATE TABLE federation_uni_sync_config (
...
-    proof_type TEXT NOT NULL CHECK(proof_type IN ('issuance', 'transfer')),
-
-    -- This field is a boolean that indicates whether or not the given universe
-    -- should accept remote proof insertion via federation sync.
     allow_sync_insert BOOLEAN NOT NULL,
 
     -- This field is a boolean that indicates whether or not the given universe
     -- should accept remote proof export via federation sync.
-    allow_sync_export BOOLEAN NOT NULL,
+    allow_sync_export BOOLEAN NOT NULL, proof_type TEXT REFERENCES proof_types(proof_type),
```
Roasbeef, line 387: Not sure why it didn't do a new line here.

tapdb/sqlc/schemas/generated_schema.sql
ffranr: yes! very valuable

ffranr (approved):
Nice refactor :+1:

tapdb/migrations_test.go
```diff
@@ -4,16 +4,29 @@ import (
...
+// backend is Postgres. In particular, it transforms occurrences of hex literals
+// formatted as X'...' into the format '\x...'.
+func transformByteLiterals(t *testing.T, db *BaseDB, query string) string {
+	if db.Backend() == sqlc.BackendTypePostgres {
+		re := regexp.MustCompile(`X'([0-9A-Fa-f]+?)'`)
+		query = re.ReplaceAllString(query, `'\x$1'`)
+	}
+	return query
+}
+
```
guggero, line 29: This cut off the Godoc of the function below...

Issue #1062, opened 2024-08-05
[feature]: Consolidate SQL Schema into a Single SQL File and Integrate with CI


ffranr:
Our current SQL database schema is distributed across multiple SQL migration scripts. This structure complicates understanding the entire schema, as it requires examining multiple migration scripts individually.

To simplify this, we should generate a complete database schema as a single SQL file and place it in the `docs` directory for easier reference. Additionally, similar to our test vectors, this schema SQL file should be autogenerated and checked for determinism as part of our Continuous Integration (CI) process.

**Tasks:**
1. Use SQLite to dump the database schema from a fully migrated database.
2. Save the dumped schema as a SQL file in the `docs` directory.
3. Integrate the schema generation process into CI:
   - Autogenerate the schema during the CI build process.
   - Implement checks to ensure the schema dump is deterministic.

**Command to use:**
```sql
sqlite3 <database_file> .schema > docs/database_schema.sql
```

**Note:**
- Ensure that the schema reflects the fully migrated state of the database.
- Implement the CI checks to ensure any changes to the schema are intentional and deterministic.

PR #1387, opened 2025-02-14
tapdb+sqlc: sqlc: add script to merge SQL migrations into consolidated schemas 

Roasbeef:
In this PR, we introduce a new Go script located in
`cmd/merge-sql-schemas/main.go` that consolidates SQL migration files
into a single schema file. The script reads all `.up.sql` migration
files from the `tapdb/sqlc/migrations` directory, executes them in order
on an in-memory SQLite database, and retrieves the final schema
definitions for tables and views.

Additionally, the Makefile is updated to include a call to this new
script during the SQL code generation process. The `.gitignore` file is
that temporary or unwanted files do not clutter the repository.
also modified to exclude files matching the pattern `.aider*`, ensuring

When I was making #1386, I made a few mistakes along the way as I was looking at the _initial_ migration file, but later migration files ended up modifying a schema I was trying to migrate. With this, devs can look in a single place for the latest set of schemas. 

Fixes https://github.com/lightninglabs/taproot-assets/issues/1062

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,6 +249,8 @@ gen: rpc sqlc
 sqlc:
 	@$(call print, "Generating sql models and queries in Go")
 	./scripts/gen_sqlc_docker.sh
+	@$(call print, "Merging SQL migrations into consolidated schemas")
+	go run ./cmd/merge-sql-schemas/main.go
 
 sqlc-check: sqlc
 	@$(call print, "Verifying sql code generation.")
--- a/Makefile
+++ b/Makefile
@@ -252,9 +252,17 @@ sqlc:
 	@$(call print, "Merging SQL migrations into consolidated schemas")
 	go run ./cmd/merge-sql-schemas/main.go
 
-sqlc-check: sqlc
+sqlc-check: 
 	@$(call print, "Verifying sql code generation.")
-	if test -n "$$(git status --porcelain '*.go')"; then echo "SQL models not properly generated!"; git status --porcelain '*.go'; exit 1; fi
+	@if [ ! -f tapdb/sqlc/schemas/generated_schema.sql ]; then \
+		echo "Missing file: tapdb/sqlc/schemas/generated_schema.sql"; \
+		exit 1; \
+	fi
+	@if test -n "$$(git status --porcelain '*.go')"; then \
+		echo "SQL models not properly generated!"; \
+		git status --porcelain '*.go'; \
+		exit 1; \
+	fi
 
 rpc:
 	@$(call print, "Compiling protos.")
--- a/cmd/merge-sql-schemas/main.go
+++ b/cmd/merge-sql-schemas/main.go
@@ -1,0 +1,106 @@
+package main
+
+import (
+	"database/sql"
+	"log"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+
+	_ "modernc.org/sqlite" // Register the pure-Go SQLite driver.
+)
+
+func main() {
+	// Open an in-memory SQLite database.
+	db, err := sql.Open("sqlite", ":memory:")
+	if err != nil {
+		log.Fatalf("failed to open in-memory db: %v", err)
+	}
+	defer db.Close()
+
+	migrationDir := "tapdb/sqlc/migrations"
+	files, err := os.ReadDir(migrationDir)
+	if err != nil {
+		log.Fatalf("failed to read migration dir: %v", err)
+	}
+
+	var upFiles []string
+	upRegex := regexp.MustCompile(`\.up\.sql$`)
+	for _, f := range files {
+		if !f.IsDir() && upRegex.MatchString(f.Name()) {
+			upFiles = append(upFiles, f.Name())
+		}
+	}
+	sort.Strings(upFiles)
+
+	// Execute each up migration in order.
+	for _, fname := range upFiles {
+		path := filepath.Join(migrationDir, fname)
+		data, err := os.ReadFile(path)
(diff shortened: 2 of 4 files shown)


cmd/merge-sql-schemas/main.go
```diff
@@ -0,0 +1,101 @@
...
+	//
+	// In our case, after running all the migration files on an in‑memory
+	// database, we execute the following query to extract only the schema
+	// definitions for tables and views. Ordering by name ensures the output
+	// is stable across runs.
+	//
+	// This way, we can consolidate and export the complete database schema
+	// as it stands after all migrations have been applied.
+	// ---------------------------------------------------------------------
+	rows, err := db.Query(`SELECT type, name, sql FROM sqlite_master WHERE type IN ('table','view') ORDER BY name`) //nolint:lll
```
guggero, line 70: nit: why not just break the string into two lines instead of making things even longer with the `nolint` comment?

cmd/merge-sql-schemas/main.go
```diff
@@ -0,0 +1,101 @@
...
+	defer rows.Close()
+
+	var finalSchema string
+	for rows.Next() {
+		var typ, name, sqlDef string
+		if err := rows.Scan(&typ, &name, &sqlDef); err != nil {
+			log.Fatalf("error scanning row: %v", err)
+		}
+
+		// Append the retrieved CREATE statement. We add a semicolon and a
```
guggero, line 83: nit: line length.

cmd/merge-sql-schemas/main.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"database/sql"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+
+	_ "modernc.org/sqlite" // pure Go SQLite driver
```
guggero, line 12: nit: Should be full sentence, including capitalization and punctuation.

guggero (approved):
Nice! Very useful :tada:

cmd/merge-sql-schemas/main.go
```diff
@@ -0,0 +1,101 @@
...
+func main() {
+	// Open an in-memory SQLite database.
+	db, err := sql.Open("sqlite", ":memory:")
+	if err != nil {
+		log.Fatalf("failed to open in-memory db: %v", err)
+	}
+	defer db.Close()
+
+	migrationDir := "tapdb/sqlc/migrations"
+	files, err := ioutil.ReadDir(migrationDir)
```
guggero, line 24: `ioutil` shouldn't be used anymore. Use `os.ReadDir()` and `os.ReadFile` instead.

guggero (changes requested):

cmd/merge-sql-schemas/main.go
Roasbeef: Ended up making it multi-line given it's using a multi-line string literal.

guggero (approved):

cmd/merge-sql-schemas/main.go
```diff
@@ -0,0 +1,105 @@
...
+	"path/filepath"
+	"regexp"
+	"sort"
+
+	_ "modernc.org/sqlite" // Register the pure-Go SQLite driver.
+)
+
+func main() {
+	// Open an in-memory SQLite database.
+	db, err := sql.Open("sqlite", ":memory:")
```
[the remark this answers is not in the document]
Roasbeef: Yup, my initial idea was to use some SQL parser thing to add all the statements, then generate a final table from that. Our `aider` AI-Overloads instead came up with this method. This also ensures that w/e we produce can be accepted by `sqlite` as well.

Makefile
```diff
@@ -254,6 +254,8 @@ gen: rpc sqlc
 sqlc:
 	@$(call print, "Generating sql models and queries in Go")
 	./scripts/gen_sqlc_docker.sh
+	@$(call print, "Merging SQL migrations into consolidated schemas")
+	go run ./cmd/merge-sql-schemas/main.go
```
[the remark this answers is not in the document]
Roasbeef: Done.

cmd/merge-sql-schemas/main.go
```diff
@@ -0,0 +1,105 @@
...
+		log.Fatalf("error iterating rows: %v", err)
+	}
+
+	// Finally, we'll write out the new schema, taking care to ensure that
+	// that output dir exists.
+	outDir := "tapdb/sqlc/schemas"
+	if err = os.MkdirAll(outDir, 0755); err != nil {
+		log.Fatalf("failed to create schema output dir: %v", err)
+	}
+	outFile := filepath.Join(outDir, "final_schema.sql")
```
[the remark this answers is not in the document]
Roasbeef: Done.

Makefile
```diff
@@ -252,9 +252,17 @@ sqlc:
 	@$(call print, "Merging SQL migrations into consolidated schemas")
 	go run ./cmd/merge-sql-schemas/main.go
 
-sqlc-check: sqlc
+sqlc-check:
```
guggero, line 255: Why did you remove the dependency on the `sqlc` target? Now it won't actually generate the schema anymore to see if there's a diff.

guggero (changes requested):

Makefile
Roasbeef: My bad, missed this. I removed it to verify that the bash fragment I had worked (so I could remove the file, then run the check w/o it adding the file again).

PR #1394, opened 2025-02-17
scripts: update georgetsagk.asc gpg key

GeorgeTsagk:
## Description

Updates the expected pub key fingerprint to the correct one, and strips the old pub key from the ASCII armored format in `georgetsagk.asc` which contained both old and new key.


guggero (approved):
tACK, LGTM :tada:

ffranr (approved):

PR #1395, opened 2025-02-17
Add itest oracle harness

GeorgeTsagk:
## Description

This PR adds the oracle harness that is used in the [LitD itests](https://github.com/lightninglabs/lightning-terminal/blob/master/itest/oracle_test.go) to our tapd repo. This will allow us to run an actual dummy grpc oracle server that the tapd instances of our itests can use. This will offer a slightly bigger code coverage in itests plus enable more complex scenarios with moving prices.

This branch is based on `rfq-negotiation-groupkey`, which is also set as the base of this GH pull request in order to not see duplicate commits.

- [ ] Fix bug: The `SubscribeRfqEventNtfns` RPC endpoint will crash if it tries to marshal the rfq message sent by our peer if they consider our price unacceptable.

### DIFF
--- a/itest/rfq_test.go
+++ b/itest/rfq_test.go
@@ -8,9 +8,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/rfqmath"
 	"github.com/lightninglabs/taproot-assets/rfqmsg"
 	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/rfqrpc"
@@ -19,6 +21,7 @@ import (
 	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
 	"github.com/lightningnetwork/lnd/lntest"
 	"github.com/lightningnetwork/lnd/lntest/node"
+	"github.com/lightningnetwork/lnd/lntest/port"
 	"github.com/lightningnetwork/lnd/lntest/wait"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/tlv"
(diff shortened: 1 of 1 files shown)


guggero (approved):
Linter is unhappy, otherwise LGTM :tada:

rfq/oracle.go
```diff
@@ -335,13 +321,25 @@ func (r *RpcPriceOracle) QueryBidPrice(ctx context.Context,
...
+		subjectSpecifier.Id = &oraclerpc.AssetSpecifier_AssetId{
+			AssetId: assetID[:],
+		}
+
+	case assetSpecifier.HasGroupPubKey():
+		groupKey := assetSpecifier.UnwrapGroupKeyToPtr()
+		subjectSpecifier.Id = &oraclerpc.AssetSpecifier_GroupKey{
+			GroupKey: groupKey.SerializeCompressed(),
+		}
+	}
```
ffranr, line 338: this could be an RPC marshal function in the RFQ package, could be called in `QueryAskPrice` also:
```
rpcMarshalAssetSpecifier(asset.Specifier) oraclerpc.AssetSpecifier
```

itest/oracle_harness.go
```diff
@@ -0,0 +1,313 @@
...
+	if o.grpcServer != nil {
+		o.grpcServer.Stop()
+	}
+	if o.grpcListener != nil {
+		_ = o.grpcListener.Close()
+	}
+}
+
+// getAssetRates returns the asset rates for a given transaction type and
+// subject asset max amount.
```
ffranr, line 99: `subject asset max amount` <- I don't think this is true?

itest/rfq_test.go
```diff
@@ -546,13 +546,41 @@ type rfqTestScenario struct {
 	CarolTapd *tapdHarness
 }
 
+// rfqTestScenarioOpts is a struct that holds options related to creating an rfq
+// test scenario.
+type rfqTestScenarioOpts struct {
+	oracleServerAddr string
+}
```
ffranr, line 577: it would be better here if we could specify different oracle servers for alice, bob, and carol.

itest/oracle_harness.go
```diff
@@ -0,0 +1,313 @@
...
+type oracleHarness struct {
+	oraclerpc.UnimplementedPriceOracleServer
+
+	listenAddr string
+
+	grpcListener net.Listener
+	grpcServer   *grpc.Server
+
+	bidPrices map[string]rfqmath.BigIntFixedPoint
+	askPrices map[string]rfqmath.BigIntFixedPoint
```
ffranr, line 33: i think you should add a comment to explain what `string` is here. in other words, I think we should have field level doc comments here (even though it's test code).

ffranr (commented):
Could we specify different oracle servers for alice, bob, and carol from the get go perhaps?

ffranr (approved):
I think you can also use `asset.Specifier` as a map key. I think all fields are comparable.

guggero:
> I think you can use `asset.Specifier` as a map key. I think all fields are comparable.

We should be careful when comparing fields that have a `btcec.PublicKey`. Even though its members are comparable, it's possible that the same point on a curve might be encoded in different ways (see the `Normalize()` method of the `FieldVal` member).

PR #1399, opened 2025-02-18
rfq+rfqmath: fix sats-per-asset conversion in mock price oracle

ffranr:
* Add a fixed-point conversion function for converting sats per asset to asset units per BTC.
* Add a unit test.
* Use the new function to fix the conversion in `NewMockPriceOracleSatPerAsset`.


guggero (approved):
Thanks for the fix, LGTM :tada:

PR #1400, opened 2025-02-19
proof: add new IgnoreChecker proof rejection cache 

Roasbeef:
In this PR, we introduce a new proof.VerifierCtx struct to cut down
on line noise when verifying single proofs and proof files. This also
lets us centralize the creation of the various interfaces we need to
verify a proof.

We then add a new interface, the IgnoreChecker. This is to be
used as a proof rejection cache, which allows us to ensure that we don't
continue to validate proofs that we already know to be invalid. The
underlying cache implementation is also intended to support its own
invalidation to handle re-org edge cases.

### DIFF
--- a/proof/append_test.go
+++ b/proof/append_test.go
@@ -212,8 +212,7 @@ func runAppendTransitionTest(t *testing.T, assetType asset.Type, amt uint64,
 
 	// Append the new transition to the genesis blob.
 	transitionBlob, transitionProof, err := AppendTransition(
-		genesisBlob, transitionParams, MockHeaderVerifier,
-		MockMerkleVerifier, MockGroupVerifier, MockChainLookup,
+		genesisBlob, transitionParams, MockVerifierCtx,
 	)
 	require.NoError(t, err)
 	require.Greater(t, len(transitionBlob), len(genesisBlob))
@@ -421,8 +420,7 @@ func runAppendTransitionTest(t *testing.T, assetType asset.Type, amt uint64,
 	}
 
 	split1Blob, split1Proof, err := AppendTransition(
-		transitionBlob, split1Params, MockHeaderVerifier,
-		MockMerkleVerifier, MockGroupVerifier, MockChainLookup,
+		transitionBlob, split1Params, MockVerifierCtx,
 	)
 	require.NoError(t, err)
 	require.Greater(t, len(split1Blob), len(transitionBlob))
@@ -464,8 +462,7 @@ func runAppendTransitionTest(t *testing.T, assetType asset.Type, amt uint64,
 	}
 
 	split2Blob, split2Proof, err := AppendTransition(
-		transitionBlob, split2Params, MockHeaderVerifier,
-		MockMerkleVerifier, MockGroupVerifier, MockChainLookup,
+		transitionBlob, split2Params, MockVerifierCtx,
 	)
 	require.NoError(t, err)
 	require.Greater(t, len(split2Blob), len(transitionBlob))
@@ -508,8 +505,7 @@ func runAppendTransitionTest(t *testing.T, assetType asset.Type, amt uint64,
 	}
 
 	split3Blob, split3Proof, err := AppendTransition(
-		transitionBlob, split3Params, MockHeaderVerifier,
-		MockMerkleVerifier, MockGroupVerifier, MockChainLookup,
+		transitionBlob, split3Params, MockVerifierCtx,
 	)
 	require.NoError(t, err)
 	require.Greater(t, len(split3Blob), len(transitionBlob))
@@ -569,10 +565,7 @@ func verifyBlob(t testing.TB, blob Blob) *AssetSnapshot {
 	f := NewEmptyFile(V0)
 	require.NoError(t, f.Decode(bytes.NewReader(blob)))
 
-	finalSnapshot, err := f.Verify(
-		context.Background(), MockHeaderVerifier, MockMerkleVerifier,
-		MockGroupVerifier, MockChainLookup,
-	)
+	finalSnapshot, err := f.Verify(context.Background(), MockVerifierCtx)
 	require.NoError(t, err)
 
 	return finalSnapshot
--- a/proof/mock.go
+++ b/proof/mock.go
@@ -27,6 +27,18 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var (
+	// MockVerifierCtx is a verifier context that uses mock implementations
+	// for all the verifier interfaces.
+	MockVerifierCtx = VerifierCtx{
+		HeaderVerifier:      MockHeaderVerifier,
+		MerkleVerifier:      MockMerkleVerifier,
+		GroupVerifier:       MockGroupVerifier,
+		GroupAnchorVerifier: MockGroupAnchorVerifier,
+		ChainLookupGen:      MockChainLookup,
+	}
+)
+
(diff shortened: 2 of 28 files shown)


proof/append_test.go
```diff
@@ -569,9 +565,15 @@ func verifyBlob(t testing.TB, blob Blob) *AssetSnapshot {
 	f := NewEmptyFile(V0)
 	require.NoError(t, f.Decode(bytes.NewReader(blob)))
 
+	vCtx := VerifierCtx{
```
guggero, line 568: Can use `MockVerifierCtx` here too?

proof/proof_test.go
```diff
@@ -812,10 +812,15 @@ func TestGenesisProofVerification(t *testing.T) {
 				tc.genesisRevealMutator, tc.groupRevealMutator,
 				tc.assetVersion,
 			)
+
+			vCtx := VerifierCtx{
```
guggero, line 819: nit: `MockVerifierCtx`?

proof/verifier.go
```diff
@@ -53,20 +65,15 @@ type BaseVerifier struct {
...
 	err := proofFile.Decode(blobReader)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse proof: %w", err)
 	}
 
-	return proofFile.Verify(
-		ctx, headerVerifier, merkleVerifier, groupVerifier,
-		chainLookupGen.GenFileChainLookup(&proofFile),
-	)
+	return proofFile.Verify(ctx, vCtx)
```
guggero, line 94: This changes the semantics on the use of `ProofChainLookup`. Before, we passed in the whole file, which actually made it useful as a cache, since it could look up `TxBlockHeight` in a previous proof of the file. Now that we always just pass in a single proof (within `verifyAssetStateTransition`), this caching effect isn't there anymore and we always need to hit the DB...

Perhaps we can keep the old signature on `Proof.Verify()` and just change the signature of the `Verifier.Verify`?

proof/verifier.go
```diff
@@ -681,11 +696,8 @@ func (p *Proof) Verify(ctx context.Context, prev *AssetSnapshot,
 	// TODO(roasbeef): need tx index as well
 
 	return &AssetSnapshot{
-		Asset: &p.Asset,
-		OutPoint: wire.OutPoint{
-			Hash:  p.AnchorTx.TxHash(),
-			Index: p.InclusionProof.OutputIndex,
-		},
+		Asset:             &p.Asset,
+		OutPoint:          anchorOutPoint,
```
guggero, line 744: This commit doesn't compile, this variable doesn't exist yet.

proof/mock.go
```diff
@@ -1026,3 +1026,21 @@ func (tmr *TestMetaReveal) ToMetaReveal(t testing.TB) *MetaReveal {
 		UnknownOddTypes:     tmr.UnknownOddTypes,
 	}
 }
+
+type mockIgnoreChecker struct {
+	ignoredAssetPoints fn.Set[AssetPoint]
+	ignoreAll          bool
+}
+
+func newMockIgnoreChecker(ignoreAll bool, ignorePoints ...AssetPoint,
```
guggero, line 1035: nit: break before `ignorePoints` to avoid awkward wrapping?

proof/verifier.go
```diff
@@ -571,6 +590,31 @@ func (p *Proof) Verify(ctx context.Context, prev *AssetSnapshot,
...
+				ScriptKey: asset.ToSerialized(
+					p.Asset.ScriptKey.PubKey,
+				),
+			}
+
+			return checker.IsIgnored(assetPoint)
+		},
+	)
+	if fail {
+		return prev, ErrProofInvalid
```
guggero, line 615: nit: wrap error with some context? So we know it failed due to being ignored?

guggero (changes requested):
Thanks for the refactor, that was definitely needed!
I think this changes the caching behavior quite a bit, so we might want to revert small parts of it (see inline comment).

proof/append_test.go
Roasbeef: Yep forgot about this area, fixed.

proof/verifier.go
guggero: So the chain lookup is used for three different things during (relative) lock time validation:
 - Get the current block height (easy, just a "cheap" call to the backend, no caching required)
 - Block mean time ("expensive" and not really cacheable, multiple calls to the backend required, but only required for unix timestamp based lock times)
 - Get block height of input transaction (requires txindex or chain re-scan, so considered too expensive), see https://github.com/lightninglabs/taproot-assets/blob/e2107c8977eb727a2406862b41ca564279453891/vm/vm.go#L586

The third one is the problem, as with Neutrino for example we can't do it in a reasonable time. BUT, that's where the proof file comes in: The previous proof in the file contains exactly the transaction we need plus its mined height. That's why we cache the full file whenever possible.

.aider.conf.yml
```diff
@@ -0,0 +1,18 @@
+read: CONVENTIONS.md
```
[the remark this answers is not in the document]
Roasbeef: Didn't mean to commit this, was meant to be as part of the `.gitignore`.

.aider.chat.history.md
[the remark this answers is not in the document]
Roasbeef: Yep a mistake, will purge all these and properly add the gitignore.

msg.txt
```diff
@@ -0,0 +1 @@
+In this commit we add roasbeef's new GPG signing key.
```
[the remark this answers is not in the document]
Roasbeef: Shouldn't have been committed, removed.

proof/verifier.go
Roasbeef: Thanks, for that explanation. 

Pushed up a new version. I landed somewhere in the middle: `VerifierCtx` is still in place, but for `proof.Verify` it accepts the `ChainLookup` as another arg. 

I considered making a new `fn.Either[ChainLookup, ChainLookupGen]` to make it clearer that they're used un a mutually exclusive way. Wasn't sure how people would feel about the ergonomics of that kind of approach though. I think we can mask some of the boiler plate with some helper methods. I lean towards this approach myself as it's less ambiguous _which_ you should use (the generator or the ready instance).

proof/proof_stitching_test.go
```diff
@@ -121,9 +127,7 @@ func TestProofStitching(t *testing.T) {
 		// get an "invalid exclusion proof" error when validating, we
 		// need to skip this proof.
 		_, err = f.Verify(
-			context.Background(), verifier.VerifyHeader,
-			DefaultMerkleVerifier, MockGroupVerifier,
-			MockChainLookup,
+			context.Background(), vCtx,
```
guggero: Here we actually switch to the `DefaultMerkleVerifier` instead of the `MockMerkleVerifier`, since we now have a valid merkle proof (which IIRC is part of the test case).

proof/verifier.go
```diff
@@ -304,6 +308,12 @@ func (p *Proof) verifyAssetStateTransition(ctx context.Context,
 		splitAssets = append(splitAssets, splitAsset)
 	}
 
+	chainLookup, err := vCtx.ChainLookupGen.GenProofChainLookup(p)
```
guggero: We should actually pass in the `chainLookup` from `Verify`. Otherwise the caching still doesn't work.

guggero (approved):
Pushed up a couple of fixes. With that, LGTM :tada:

PR #1405, opened 2025-02-20
Support Directory Exclusion in Golang Version Linter Check

ffranr:
Enhance the Golang version check linter by adding support for directory exclusions.  

**Changes:**  
- Updated the `is_exclude` function to allow excluding entire directories, not just individual files.  
- Applied this to exclude all files in `./itest/chantools` from the version check.  

This improves flexibility in managing linter exclusions for project directories.


guggero (approved):
Thanks, LGTM :tada:

PR #1410, opened 2025-02-25
Enable channel features via flag

Jossec101:
This PR removes the hardcoding on enabling support channel funding via flag

### DIFF
--- a/cmd/commands/commands.go
+++ b/cmd/commands/commands.go
@@ -84,6 +84,11 @@ func NewApp(actionOpts ...ActionOption) cli.App {
 				"macaroon jar instead of the default one. " +
 				"Can only be used if profiles are defined.",
 		},
+		cli.BoolFlag{
+			Name: "enable-channels-features",
+			Usage: "allows tapd to be able to" +
+				"fund channels without litd",
+		},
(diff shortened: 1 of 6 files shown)


ffranr (commented):
I think this is fine as is, but would probably be better if `EnableChannelFeatures` was in the general server config rather than the RPC specific. I see that it only applies to an RPC endpoint, but there's nothing RPC specific in the flag.

If we add this to tapd cli, will this mislead users towards danger somehow? What happens if a user calls `FundChannel` without litd.

I would be ok to add this flag without further checks to `ExperimentalConfig` with a warning in the description.

guggero:
!lightninglabs-deploy mute

Jossec101:
> The reason this flag is hard-coded is that it can only be set to `true` when `tapd` is running in integrated mode within `litd`. So setting this flag will just make `tapd` return an error at a later state. You currently cannot create asset channels with `tapd` and `litd` running in separate processes (also known as "remote" mode).

Fair enough, is remote mode far on the roadmap? we dont really see a good aproach to have a "Integrated" binary as a good way for production deployments.

Closing

PR #1411, opened 2025-02-25
multi: Add label to `PublishTransaction`

gijswijs:
This PR chanegs the label used for Txs published to the network. Before, *all* Txs got the label `tapd-asset-minting`. Now, the label depends on the context. Currently we use two labels:

- `tapd-asset-minting`
- `tapd-asset-send`

In the future we could even allow for more granular information being passed to the label.

### DIFF
--- a/chain_bridge.go
+++ b/chain_bridge.go
@@ -227,9 +227,8 @@ func (l *LndRpcChainBridge) GetBlockTimestamp(ctx context.Context,
 // PublishTransaction attempts to publish a new transaction to the
 // network.
 func (l *LndRpcChainBridge) PublishTransaction(ctx context.Context,
-	tx *wire.MsgTx) error {
+	tx *wire.MsgTx, label string) error {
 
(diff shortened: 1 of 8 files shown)


ffranr (commented):
IMO, we should use

`tapd-asset-transfer`
`tapd-asset-issuance`

to align with our proof type terminology.

And maybe even?

`TAP-transfer`
`TAP-issuance`

guggero:
I vote for:
```
tapd-asset-transfer
tapd-asset-issuance
```

Roasbeef:
Can we go with one of the above ideas so we can merge this?

ffranr:
> Can we go with one of the above ideas so we can merge this?

I've gone with:
```
tapd-asset-transfer
tapd-asset-issuance
```

I've also made the labels constants and rebased.

Roasbeef (approved):
LGTM 🦺

PR #1413, opened 2025-02-26
[wallet 3/3]: group key support for channel funding

guggero:
Adds the ability to fund multiple tranches of a grouped asset into a channel.

Corresponding itest: https://github.com/lightninglabs/lightning-terminal/pull/987


guggero:
TODO for myself: add group key to channel funding blob, see https://github.com/lightninglabs/lightning-terminal/pull/995/commits/4768ba244d8eaecce7dafacbeacd1d68d827d0cf#r1982366865.

Roasbeef:
> TODO for myself: add group key to channel funding blob, see

To be done in this PR?

tapsend/send.go
```diff
@@ -208,6 +208,12 @@ type FundingDescriptor struct {
 
 	// CoinSelectType specifies the type of coins that should be selected.
 	CoinSelectType CoinSelectType
+
+	// DistinctSpecifier indicates whether we _only_ look at either the
+	// group key _or_ the asset ID but not both. That means, if the group
+	// key is set, we ignore the asset ID and allow multiple inputs of the
+	// same group to be selected.
+	DistinctSpecifier bool
```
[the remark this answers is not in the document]
Roasbeef: Alternatively, for the use case we see in this PR, couldn't we have the caller create a specifier that _only_ has a group key if funding is initiated using a group key?

tapdb/assets_store.go
```diff
@@ -886,8 +886,13 @@ func (a *AssetStore) constraintsToDbFilter(
...
 		assetFilter.KeyGroupFilter = groupKeyBytes
 
-		// TODO(roasbeef): only want to allow asset ID or other and not
-		// both?
+		// If we query by group key, we don't also include the asset ID,
+		// otherwise we'd only get assets from that specific tranche.
+		if query.DistinctSpecifier &&
+			len(assetFilter.KeyGroupFilter) > 0 {
+
+			assetFilter.AssetIDFilter = nil
```
Roasbeef, line 894: Should add a test case for this, assuming we don't go with one of the other options I mentioned in my other comment.

rpcserver.go
```diff
@@ -7073,7 +7073,7 @@ func (r *rpcServer) FundChannel(ctx context.Context,
 	}
 
 	assetID, groupKey, err := parseAssetSpecifier(
-		req.GetAssetId(), "", nil, "",
+		req.GetAssetId(), "", req.GetGroupKey(), "",
```
Roasbeef, line 7087: Re the other comment, this is where we can implement the either or behavior: 
  * We only allow asset ID or group key to be set
  * We make a specifier that only has one half populated 

With that change, I don't think we'd need that new bool and behavior change in the db?

rfqmsg/custom_channel_data.go
```diff
@@ -28,19 +28,19 @@ type JsonAssetUtxo struct {
...
+	FundingAssets       []JsonAssetUtxo    `json:"funding_assets"`
+	LocalAssets         []JsonAssetTranche `json:"local_assets"`
+	RemoteAssets        []JsonAssetTranche `json:"remote_assets"`
+	OutgoingHtlcs       []JsonAssetTranche `json:"outgoing_htlcs"`
+	IncomingHtlcs       []JsonAssetTranche `json:"incoming_htlcs"`
+	Capacity            uint64             `json:"capacity"`
+	LocalBalance        uint64             `json:"local_balance"`
+	RemoteBalance       uint64             `json:"remote_balance"`
+	OutgoingHtlcBalance uint64             `json:"outgoing_htlc_balance"`
+	IncomingHtlcBalance uint64             `json:"incoming_htlc_balance"`
```
Roasbeef, line 43: This is the sum of the incoming HTLCs above?

rfqmsg/custom_channel_data.go
```diff
@@ -28,19 +28,19 @@ type JsonAssetUtxo struct {
...
 // assets within a channel.
 type JsonAssetChannel struct {
-	Assets []JsonAssetChanInfo `json:"assets"`
+	FundingAssets       []JsonAssetUtxo    `json:"funding_assets"`
+	LocalAssets         []JsonAssetTranche `json:"local_assets"`
+	RemoteAssets        []JsonAssetTranche `json:"remote_assets"`
+	OutgoingHtlcs       []JsonAssetTranche `json:"outgoing_htlcs"`
+	IncomingHtlcs       []JsonAssetTranche `json:"incoming_htlcs"`
+	Capacity            uint64             `json:"capacity"`
+	LocalBalance        uint64             `json:"local_balance"`
```
Roasbeef, line 40: Will there be one of these new structs for each future asset group? 

Also when it comes to balances, will we show a sum of the total grouped assets, but then individual balances for each asset ID included?

tapdb/assets_store.go
```diff
@@ -1477,6 +1477,10 @@ func locatorToProofQuery(locator proof.Locator) (FetchAssetProof, error) {
 		args.Outpoint = outpoint
 	}
 
+	if locator.AssetID != nil {
+		args.AssetID = locator.AssetID[:]
```
Roasbeef, line 1481: Hmm, yeh looks like another instance where we forgot to handle/prevent the collision case that can happen here. 

Can we add a test case here to make sure we can properly select based on the final 3-tuple (assetID, scriptKey, outpoint)?

tapdb/assets_store.go
```diff
@@ -3062,7 +3062,7 @@ func (a *AssetStore) LogAnchorTxConfirm(ctx context.Context,
 
 	var (
 		writeTxOpts    AssetStoreTxOptions
-		localProofKeys []asset.SerializedKey
+		localProofKeys []tapfreighter.OutputIdentifier
```
Roasbeef, line 3065: > We recently changed the semantics around transfers. We now allow
> multiple virtual packets to be combined in a single transfer.

Wasn't this always the case if a user used the vPSBT APIs?

tapdb/assets_store.go
```diff
@@ -3088,9 +3088,11 @@ func (a *AssetStore) LogAnchorTxConfirm(ctx context.Context,
 		// We'll keep around the IDs of the assets that we set to being
 		// spent. We'll need one of them as our template to create the
 		// new assets.
-		spentAssetIDs := make([]int64, len(inputs))
+		spentAssetIDs := make(map[asset.ID]int64, len(inputs))
```
Roasbeef, line 3091: Isn't this also vulnerable to a collision? If we spend multiple inputs that are the same asset ID, but distinct script key.

tapfreighter/interface.go
```diff
@@ -218,6 +221,35 @@ type Anchor struct {
...
+		bytes.NewReader(proofSuffix),
+		proof.AssetLeafRecord(&outProofAsset),
+	)
+	if err != nil {
+		return zero, fmt.Errorf("unable to sparse decode proof: %w",
+			err)
+	}
+
+	keyData := append([]byte{}, fn.ByteSlice(outProofAsset.ID())...)
+	keyData = append(keyData, scriptKey.SerializeCompressed()...)
```
Roasbeef, line 249: No outpoint needed?

tapchannel/aux_closer.go
```diff
@@ -528,38 +529,34 @@ func (a *AuxChanCloser) ShutdownBlob(
 		return none, err
 	}
 
-	// Next, we'll collect all the assets that we own in this channel.
-	assets := commitState.LocalAssets.Val.Outputs
+	// Next, we'll collect all the asset IDs that were committed to the
+	// channel.
+	assetIDs := fn.Map(
```
Roasbeef, line 544: IIUC, there can potentially be duplicates here: multiple UTXOs of a given asset ID were used to fund the channel? Perhaps it should be run through as a set first.

tapchannel/aux_closer.go
```diff
@@ -528,38 +529,34 @@ func (a *AuxChanCloser) ShutdownBlob(
...
+		err = a.cfg.AddrBook.InsertScriptKey(ctx, newKey, true)
 		if err != nil {
-			return none, fmt.Errorf("error adding new address: %w",
-				err)
+			return none, fmt.Errorf("error declaring script key: "+
+				"%w", err)
 		}
 
-		scriptKeys[channelAsset.AssetID.Val] = *newKey.PubKey
+		scriptKeys[assetID] = *newKey.PubKey
```
Roasbeef, line 569: Based on the comment above, we may end up with script keys thgat was created/inserted but were never used as we overrode them here in the map.

tapchannel/aux_closer.go
```diff
@@ -156,6 +156,53 @@ func createCloseAlloc(isLocal bool, outputSum uint64,
 	}, nil
 }
 
+// signCommitVirtualPackets signs the commit virtual packets with the funding
+// witness, which is just the script and control block for the OP_TRUE spend.
+func signCommitVirtualPackets(ctx context.Context,
```
Roasbeef, line 161: 👍

tapchannel/aux_sweeper.go
```diff
@@ -324,16 +318,6 @@ func (a *AuxSweeper) createSweepVpackets(sweepInputs []*cmsg.AssetOutput,
 				"assets: %w", err)
 		}
 
-		// Next before we sign, we'll make sure to update the witness
-		// of the prev asset's root asset. Otherwise, we'll be signing
-		// the wrong input leaf.
-		vIn := vPackets[idx].Inputs[0]
-		if vIn.Asset().HasSplitCommitmentWitness() {
-			//nolint:lll
-			rootAsset := vIn.Asset().PrevWitnesses[0].SplitCommitment.RootAsset
```
Roasbeef, line 333: Hmm, I recall this was one of those bugs that took be a while to track down properly. 

The root asset (funding output) in this case needs to have the updated witness w/ the control block in tact, as we don't apply the segwit encoding to root assets stored as part of a split commitment.

tapscript/script.go
```diff
@@ -48,3 +52,87 @@ func NewChannelFundingScriptTree() *FundingScriptTree {
...
 }
+
+// NewChannelFundingScriptTreeUniqueID creates a new funding script tree for a
+// custom channel asset-level script key. The script tree is constructed with a
+// simple OP_TRUE script that allows anyone to spend the output and another
+// OP_RETURN script that makes the resulting script key unique. This simplifies
+// the funding process as no signatures for the asset-level witnesses need to be
+// exchanged. This is still safe because the BTC level multi-sig output is still
+// protected by a 2-of-2 MuSig2 output.
+func NewChannelFundingScriptTreeUniqueID(id asset.ID) (*FundingScriptTree,
```
Roasbeef, line 63: Do we also need to worry about the case where multiple asset UTXOs w/ the same asset ID exist in the funding output? Or is the assumption that the funder won't create such an output, as they can make a "smaller" one with the UTXO value consolidated (as tehy're the same assetID after all).

Roasbeef (changes requested):

rfqmsg/custom_channel_data.go
guggero: There will only be a single `JsonAssetChannel` per channel. But that will now more clearly show the difference between the funding blob (e.g. the assets that were committed to the channel at funding time --> `FundingAssets`, for example `25k units of asset A, 30k units of asset B`) and the current commitment blob (e.g. the current distribution of assets --> `Local/RemoteBalance/Outgoing/IncomingHtlcs`, for example `LocalBalance: 10k units of asset A, RemoteBalance: 15k units of asset A, 25k units of asset B, OutgoingHtlc: 5k units of asset B`).

tapdb/assets_store.go
guggero: I guess it was always the case **since** we implemented committing multiple vPackets to a single on-chain transaction, yes. Which was a while back, so perhaps "we recently changed" isn't super accurate.

tapfreighter/interface.go
guggero: Hmm, I guess you're right, we should probably also add the output index (the TXID is superfluous because a transfer always represents one on-chain TX), since you could theoretically have the same asset ID with the same script key in multiple outputs (although I think that's what we currently disallow to fix the exclusion proof collision issue?). But adding the output index definitely shouldn't hurt, so going to add that.

tapchannel/aux_closer.go
guggero: No, I don't think that should ever be the case. Even if you have multiple inputs of the same asset ID, we'd merge them into the same UTXO. Only for different asset IDs we need to keep distinct UTXOs around (because they'll be represented by distinct vPackets). Will add a comment to make that more clear.

tapchannel/aux_closer.go
guggero: Yeah, you're right... So perhaps for the very first one we should try to re-use the one we already derived?

tapscript/script.go
guggero: Exactly, the same assetIDs would always be merged into a single UTXO (see comment above as well).

tapchannel/aux_sweeper.go
guggero: Hmm, yeah, I thought this was probably the case. But I removed these lines and ran a full integration test suite and nothing broke. Is it possible that this was fixed indirectly in a later version by adding the witness earlier in the flow? Because I checked and the commitment transaction should already have the witness (both in the root asset as well as in each split commitment's root asset copy). But I will double check again.

tapscript/script.go
```diff
@@ -48,3 +52,87 @@ func NewChannelFundingScriptTree() *FundingScriptTree {
...
+	error) {
+
+	// First, we'll generate our OP_TRUE script.
+	fundingScript := AnyoneCanSpendScript()
+	fundingTapLeaf := txscript.NewBaseTapLeaf(fundingScript)
+
+	// Then we'll create the OP_RETURN leaf with the asset ID to make the
+	// resulting script key unique.
+	opReturnLeaf, err := asset.NewNonSpendableScriptLeaf(
+		asset.OpReturnVersion, id[:],
```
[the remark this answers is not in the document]
guggero: This is only within a single on-chain outpoint (e.g. the funding transaction). The universe has uniqueness over the tuple `(group_key_or_asset_id, script_key, outpoint)`. Which for multiple tranches of the same group key would all be the same for the individual UTXOs committed to the channel (because `group_key` would be set since it's a grouped asset, see how the universe ID works). So we just need to de-duplicate the script key by adding the asset ID to its tapscript path. And since we'd only at most have one UTXO per asset ID, there should be no additional collision possible (see comments above about merging multiple inputs of the same asset ID into a single one within the channel).

tapchannel/aux_closer.go
```diff
@@ -157,24 +156,58 @@ func createCloseAlloc(isLocal, isInitiator bool, outputSum uint64,
...
 
-	tapscriptTree := fundingScriptTree.TapscriptTree
-	ctrlBlock := tapscriptTree.LeafMerkleProofs[0].ToControlBlock(
-		&input.TaprootNUMSKey,
-	)
-	ctrlBlockBytes, err := ctrlBlock.ToBytes()
-	if err != nil {
-		return lfn.Errf[wire.TxWitness]("unable to serialize control "+
-			"block: %w", err)
+	useUniqueScriptKey := len(vPackets) > 1
```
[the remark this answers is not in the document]
guggero: Correct. The definition for a vPacket that we arrived at is that it reflects all the changes of a single assetID within one on-chain transaction. Meaning that all the inputs and all the outputs of one asset ID within a transfer would be merged into the same vPacket. Will add a comment to make more clear.

rpcserver.go
guggero: Hmm, yeah, maybe we can introduce a sub type `ExclusiveSpecifier` that embeds the original identifier (just so we don't need to change things everywhere)... Will think about it.

tapsend/send.go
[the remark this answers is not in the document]
guggero: The problem is, that even today we have both use cases where we specify the `AND` case and the `XOR` case. Both are valid use cases IMO and depending on what the user wants this boolean basically switches between them.
See https://github.com/lightninglabs/taproot-assets/pull/1413#discussion_r2023114154.

tapdb/assets_store.go
guggero: Added unit tests. Had to simplify and refactor the `TestFetchAllAssets` test a bit, but should be more easy to add more test cases in the future.

tapdb/assets_store.go
guggero: Added a test case.

tapdb/assets_store.go
guggero: Yes, there can be collisions if we spend multiple inputs of the same asset ID.
But for this case it doesn't matter, since we just need one template to create a copy from. And as long as that template comes from the correct asset ID (making sure that the references to the genesis asset and group key are correct to be copied), it doesn't matter _which_ one we use. Clarified this with comments and by renaming the variable to `copyTemplateIDs`.

tapchannel/aux_closer.go
guggero: Okay, my comment right above this was incorrect. Because we shouldn't have duplicate asset IDs (as explained in https://github.com/lightninglabs/taproot-assets/pull/1413#discussion_r2023105021), we shouldn't overwrite keys here. So we don't derive keys that we then don't end up using.

tapchannel/aux_sweeper.go
guggero: Okay, I found exactly why this isn't needed anymore.
This was fixed with c072d3f2, which was in #1154.
Which was quite a while after de5bed63 (in #938) which introduced the temporary fix in the first place.
So it seems like it was indeed needed at some point but could've been removed in #1154 already.

guggero:
Addressed all comments.

PR #1416, opened 2025-02-27
Support group keys on HTLC Interceptor & Invoice HTLC Modifier

GeorgeTsagk:
## Description

With this PR we extend the `rfq.OrderHandler` and `tapchannel.AuxInvoiceManager` sub systems to be able to read groupkey specifiers and handle HTLCs with multiple asset balances.

Based on #1382

### DIFF
--- a/tapchannel/aux_invoice_manager_test.go
+++ b/tapchannel/aux_invoice_manager_test.go
@@ -63,6 +63,22 @@ var (
 
 	// The test RFQ SCID that is derived from testRfqID.
 	testScid = testRfqID.Scid()
+
+	// The common asset ID used on test cases.
+	assetID = dummyAssetID(1)
+
+	// The asset ID of the first asset that is considered part of the group.
+	groupAssetID_1 = dummyAssetID(41)
+
+	// The asset ID of the second asset that is considered part of the
+	// group.
+	groupAssetID_2 = dummyAssetID(42)
+
+	// The specifier based on the common asset ID.
+	assetSpecifier = asset.NewSpecifierFromId(assetID)
+
+	// The specifier based on a dummy generated group key.
+	groupSpecifier = asset.NewSpecifierFromGroupKey(*pubKeyFromUint64(1337))
 )
 
 // mockRfqManager mocks the interface of the rfq manager required by the aux
@@ -95,7 +111,11 @@ func (m *mockRfqManager) AssetMatchesSpecifier(ctx context.Context,
 
 	switch {
 	case specifier.HasGroupPubKey():
-		return true, nil
+		if id == groupAssetID_1 || id == groupAssetID_2 {
+			return true, nil
+		}
+
+		return false, nil
 
 	case specifier.HasId():
 		return *specifier.UnwrapIdToPtr() == id, nil
@@ -292,11 +312,6 @@ func (m *mockHtlcModifierProperty) HtlcModifier(ctx context.Context,
 // TestAuxInvoiceManager tests that the htlc modifications of the aux invoice
 // manager align with our expectations.
 func TestAuxInvoiceManager(t *testing.T) {
-	var (
-		assetID        = dummyAssetID(1)
-		assetSpecifier = asset.NewSpecifierFromId(assetID)
-	)
-
 	testCases := []struct {
 		name            string
 		buyQuotes       rfq.BuyAcceptMap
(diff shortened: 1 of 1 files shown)


rfqmsg/records.go
```diff
@@ -82,22 +84,39 @@ func (h *Htlc) Balances() []*AssetBalance {
...
 	for idx := range h.Amounts.Val.Balances {
 		balance := h.Amounts.Val.Balances[idx]
 
-		if balance.AssetID.Val != targetAssetID {
+		ctxt, cancel := context.WithTimeout(
+			context.Background(), wait.DefaultTimeout,
+		)
+		defer cancel()
+
+		match, err := checker.AssetMatchesSpecifier(
```
Roasbeef, line 112: The current call site added in this commit (234e6440c938aadf6945f6c10c94f4ae06c96b2b) passes in a nil value, but we don't have a check for nilness here.

rfq/order.go
```diff
@@ -53,12 +54,23 @@ func parseHtlcCustomRecords(customRecords map[uint64][]byte) (*rfqmsg.Htlc,
 // SerialisedScid is a serialised short channel id (SCID).
 type SerialisedScid = rfqmsg.SerialisedScid
 
+// AssetSpecifierChecker is an interface that contains methods for checking
+// certain properties related to asset specifiers.
+type AssetSpecifierChecker interface {
+	// AssetMatchesSpecifier checks whether the passed specifier and asset
+	// ID match. If the specifier contains a group key, it will check
+	// whether the asset belongs to that group.
+	AssetMatchesSpecifier(ctx context.Context,
```
Roasbeef, line 63: I think we should remove the other version of this interface added (in rfqmsg) and just keep this one.

rfq/order.go
```diff
@@ -248,11 +261,22 @@ func (c *AssetSalePolicy) GenerateInterceptorResponse(
...
+	switch {
+	case c.AssetSpecifier.HasGroupPubKey():
+		groupKey := c.AssetSpecifier.UnwrapGroupKeyToPtr()
+
+		// We have performed checks for the asset IDs inside the HTLC
+		// against the specifier's group key in a previous step. Here
+		// we just need to provide a dummy value as the asset ID. The
+		// real asset IDs will be carefully picked in a later step in
+		// the process. What really matters now is the total amount.
+		assetID = sha256.Sum256(groupKey.SerializeCompressed())
```
Roasbeef, line 275: Why isn't the asset ID relevant here? It looks like it's used to pass along which custom records to use for the outgoing HTLC transformation. 

IIUC, we need to select the asset ID amongst the set of active channels the outgoing link has that may all be diff, but have the same asset group.

tapchannel/aux_invoice_manager.go
```diff
@@ -42,6 +42,12 @@ type RfqManager interface {
 	// node and have been requested by our peers. These quotes are
 	// exclusively available to our node for the sale of assets.
 	LocalAcceptedSellQuotes() rfq.SellAcceptMap
+
+	// AssetMatchesSpecifier checks if the provided asset satisfies the
+	// provided specifier. If the specifier includes a group key, we will
+	// check if the asset belongs to that group.
+	AssetMatchesSpecifier(ctx context.Context, specifier asset.Specifier,
```
Roasbeef, line 49: Seems the `rfq` package is already imported here, so we can embed that interface. Mainly a question re style.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -73,14 +73,37 @@ type mockRfqManager struct {
...
 
+// AssetMatchesSpecifier checks if the provided asset satisfies the provided
+// specifier. If the specifier includes a group key, we will check if the asset
+// belongs to that group.
+func (m *mockRfqManager) AssetMatchesSpecifier(ctx context.Context,
+	specifier asset.Specifier, id asset.ID) (bool, error) {
+
+	switch {
+	case specifier.HasGroupPubKey():
+		return true, nil
```
Roasbeef: Why do we just return true for group keys? We should expand the unit tests here and earlier in the PR to make sure to matches properly re same group key but diff asset IDs.

tapchannel/aux_invoice_manager_test.go
```diff
@@ -63,6 +63,22 @@ var (
 
 	// The test RFQ SCID that is derived from testRfqID.
 	testScid = testRfqID.Scid()
+
+	// The common asset ID used on test cases.
+	assetID = dummyAssetID(1)
+
+	// The asset ID of the first asset that is considered part of the group.
+	groupAssetID_1 = dummyAssetID(41)
```
Roasbeef, line 71: nix the under score ?

Roasbeef (changes requested):

PR #1418, opened 2025-03-03
Unit tests for mint anchor transaction pre-commitment output

ffranr:
Add unit tests to provide coverage for the changes made in https://github.com/lightninglabs/taproot-assets/pull/1325


ffranr:
Needs rebase.

tapgarden/mock.go
```diff
@@ -346,7 +360,7 @@ func NewGenesisTx(t testing.TB, feeRate chainfee.SatPerKWeight) psbt.Packet {
 }
 
 // FundGenesisTx add a genesis input and change output to a 1-output TX.
```
jtobin: Nit: though it's obvious from the code, it might still be worth tacking on ".., returning the output index." to the comment here. Otherwise I see uint32 when scanning the function definition & header comment, and have to spend a second calculating what it means.

ffranr: Good point. I've enhanced the doc comment. In a better world we would have a change output index type. Maybe I'll add that in a separate PR.

PR #1420, opened 2025-03-05
tapdb+universe: implement initial version of universe ignore tree

Roasbeef:
In this PR, we implement the initial version of the universe Ignore Tree. 

Along the way, we need a migration to the universe leaves, as the current unique index structure prevents us from adding two leaves that have the same script key, but distinct namespace (eg: transfer to a script key, then a burn of that). 

With that in place, we're able to implement the ignore tree using the existing MS-SMT/Universe routines we already have in the database. 

TODO 

- [ ] Consistent interfaces. 
- [ ] Unit tests for the database layer


Roasbeef:
Removed from draft.

Roasbeef:
Pushed up a new version (CI was failing as we had two migration 30's). PTAL.

Roasbeef:
Will fix the collision issues in the burn tree PR. 

This is ready for review (no changes since my last push to fix CI after the rebase).

universe/interface.go
```diff
@@ -1176,3 +1192,59 @@ type Telemetry interface {
 	QueryAssetStatsPerDay(ctx context.Context,
 		q GroupedStatsQuery) ([]*GroupedStats, error)
 }
+
+// IgnoreTuple....
```
guggero: nit: move this into the next commit where things are actually defined?

tapdb/ignore_tree.go
```diff
@@ -0,0 +1,380 @@
...
+		// Get the root of the tree to retrieve the sum.
+		root, err := tree.Root(ctx)
+		if err != nil {
+			return err
+		}
+
+		// If root is empty, return empty sum.
+		//
+		// TODO(roasbeef): verify behavior
+		if root.NodeHash() == mssmt.EmptyTreeRootHash {
```
guggero, line 198: Or, if we expect there to be no zero amount leaves in the tree, could also just do `root.NodeSum() == 0`.

tapdb/ignore_tree.go
```diff
@@ -0,0 +1,380 @@
...
+	id.GroupKey = spec.UnwrapGroupKeyToPtr()
+
+	id.ProofType = universe.ProofTypeIgnore
+	return id, nil
+}
+
+// AddTuple adds a new ignore tuples to the ignore tree.
+func (it *IgnoreUniverseTree) AddTuples(ctx context.Context,
+	spec asset.Specifier, tuples ...*universe.SignedIgnoreTuple,
+) lfn.Result[[]universe.AuthenticatedIgnoreTuple] {
```
guggero, line 44: Should we create a local alias `IgnoreTuples = []universe.AuthenticatedIgnoreTuple` to shorten some of these definitions?

guggero (commented):
Nice!
Most of my comments are around the SQL migration, the rest looks great.
Love all the rapid tests :100:

ffranr (commented):
I feel as if "tuple" is used as a a bit of a filler word here. We haven't used the word "coin" much, but this might be the right time to use it.

`IgnoreCoin`, `AddCoins(...)`,

```
// AddCoin adds a new coin to the ignore tree.
```

I think the programming rule of thumb is "Name the thing, not the container.", basically encapsulation which makes it cheaper to change the underlying data type if we ever need to.

tapdb/sqlc/migrations/000031_ignore_tree.up.sql
```diff
@@ -0,0 +1,155 @@
...
+-- Create a new universe_leaves table with the enhanced 3-column unique constraint.
+CREATE TABLE new_universe_leaves (
+    id INTEGER PRIMARY KEY,
+    asset_genesis_id BIGINT NOT NULL REFERENCES genesis_assets(gen_asset_id),
+    minting_point BLOB NOT NULL,
+    script_key_bytes BLOB NOT NULL CHECK(LENGTH(script_key_bytes) = 32),
+    universe_root_id BIGINT NOT NULL REFERENCES universe_roots(id),
+    leaf_node_key BLOB,
+    leaf_node_namespace VARCHAR NOT NULL,
+    UNIQUE(minting_point, script_key_bytes, leaf_node_namespace)
```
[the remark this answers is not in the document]
Roasbeef: Ah so have it be a named unique index instead? Makes sense, will add that.

tapdb/sqlc/schemas/generated_schema.sql
```diff
@@ -699,22 +699,15 @@ CREATE TABLE universe_events (
     event_time TIMESTAMP NOT NULL
 , event_timestamp BIGINT NOT NULL DEFAULT 0);
 
-CREATE TABLE universe_leaves (
+CREATE TABLE "universe_leaves" (
```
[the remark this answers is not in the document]
Roasbeef: Yeah not sure why that was added. The way the tool works, it just reads the schema directly from the meta table in sqlite itself... 

Perhaps there was some formatting or extra white space in the migration that resulted in this? Will fiddle with it a bit.

tapdb/sqlc/migrations/000031_ignore_tree.up.sql
```diff
@@ -0,0 +1,155 @@
...
+-- has a foreign key to universe_leaves.id that would prevent dropping the table.
+
+-- Create a temporary backup table for federation_proof_sync_log.
+CREATE TABLE new_federation_proof_sync_log (
+    id INTEGER PRIMARY KEY,
+    status TEXT NOT NULL CHECK(status IN ('pending', 'complete')),
+    timestamp TIMESTAMP NOT NULL,
+    attempt_counter BIGINT NOT NULL DEFAULT 0,
+    sync_direction TEXT NOT NULL CHECK(sync_direction IN ('push', 'pull')),
+    proof_leaf_id BIGINT NOT NULL, -- FK constraint intentionally omitted for now.
```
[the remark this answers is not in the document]
Roasbeef: IIUC with sqlite, we can't do `ALTER TABLE ADD FOREIGN KEY`. We can only: rename a table/column, add/drop a table/column. 

Need to revisit this, but the reason I needed to make a version w/o the constraint was that the migration fails otherwise either due to the rename of one of the tables above, or because we didn't update all the refs to point to the new version.

Roasbeef:
> I feel as if "tuple" is used as a a bit of a filler word here.

Tuple as in: https://en.wikipedia.org/wiki/Tuple. 

It's not really a "coin" in the context we usually use that term within Bitcoin.

universe/ignore_records.go
```diff
@@ -0,0 +1,217 @@
...
+	// SignedIgnoreTuple.
+	IgnoreSignatureType = tlv.TlvType2
+)
+
+// IgnoreTuple represents an asset previous ID that we want to ignore.
+type IgnoreTuple struct {
+	asset.PrevID
+
+	// Amount is the amount of units that we want to ignore.
+	Amount uint64
```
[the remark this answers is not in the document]
Roasbeef: The amount field here assumes that the creator knows about the actual asset UTXO and uses the correct amount. We need the amount field to make the leaf properly. 

I like the suggestion to modify the godoc string, adding in the next push.

tapdb/ignore_tree.go
```diff
@@ -0,0 +1,380 @@
...
+	// ignore tree context.
+	if !spec.HasGroupPubKey() {
+		return id, fmt.Errorf("group key must be set")
+	}
+
+	id.GroupKey = spec.UnwrapGroupKeyToPtr()
+
+	id.ProofType = universe.ProofTypeIgnore
+	return id, nil
+}
```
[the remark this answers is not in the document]
Roasbeef: I split the middle a bit and made a new `UnwrapGroupKeyOrErr()` method, which is now used.

tapdb/ignore_tree.go
Roasbeef: Can remove this TODO as it's verified with the unit tests, and also the impl: https://github.com/lightninglabs/taproot-assets/blob/581c8819aa95093494064f42f40adf700e3e4149/mssmt/store.go#L199-L201

https://github.com/lightninglabs/taproot-assets/blob/581c8819aa95093494064f42f40adf700e3e4149/tapdb/mssmt.go#L365-L369

tapdb/sqlc/migrations/000031_ignore_tree.up.sql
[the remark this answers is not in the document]
Roasbeef: So along the way I found a bug in the script to render the final generated schema: it didn't include named indexes! I added a new commit that fixes this issue.

ffranr:
> > I feel as if "tuple" is used as a a bit of a filler word here.
> 
> Tuple as in: https://en.wikipedia.org/wiki/Tuple.
> 
> It's not really a "coin" in the context we usually use that term within Bitcoin.

I think a term like “asset coin” could be useful. Or something similar, I'm not sure what the convention is around "coin". Basically a higher level term for an asset leaf in a Taproot Asset MS-SMT—something that conveys spendability.

tapdb/ignore_tree.go
```diff
@@ -0,0 +1,377 @@
...
+
+	var writeTx BaseUniverseStoreOptions
+	txErr := it.db.ExecTx(ctx, &writeTx, func(db BaseUniverseStore) error {
+		tree := mssmt.NewCompactedTree(
+			newTreeStoreWrapperTx(db, namespace),
+		)
+
+		// First, insert all tuples into the tree. This'll create a new
+		// insert universe leaves to reference the SMT leafs. set of
+		// normal SMT leafs. Once inserted, we'll then also
```
ffranr, line 72: incomplete comment

universe/interface.go
```diff
@@ -1190,3 +1191,46 @@ type Telemetry interface {
...
+	// resides within.
+	IgnoreTreeRoot mssmt.Node
+
+	// InclusionProof is the universe inclusion proof for the ignore tuple
+	// within the universe tree.
+	InclusionProof *mssmt.Proof
+}
+
+// TupleQueryResp is the response to a query for ignore tuples.
+type TupleQueryResp = lfn.Result[lfn.Option[[]AuthenticatedIgnoreTuple]]
```
ffranr, line 1213: Why include `lfn.Option` here? Wouldn't an empty slice signal the same thing as a `None`?

I think a similar sort of question can be asked for `SumQueryResp` below. I assume that if there are any leaves in the tree then the sum is always `> 0`, am I wrong? In which case `None` and `0` mean the same thing.

tapdb/sqlc/queries/assets.sql
```diff
@@ -670,6 +670,11 @@ WHERE (
     genesis_assets.asset_type = @asset_type
 );
 
+-- name: LookupGenesisID :one
+SELECT gen_asset_id
+FROM genesis_assets
+WHERE asset_id = $1;
+
```
ffranr, line 677: Instead of adding this new query, we could generalize `FetchGenesisID` above it to ignore unset fields in the WHERE clause. If that’s too large a change and we want to keep `LookupGenesisID`, then I’d suggest renaming it to `FetchGenesisIDByAssetID` or similar.  
  
I don’t believe we use the term `lookup` in any other SQL queries in this file—`fetch` seems to be our usual convention.

universe/ignore_records.go
```diff
@@ -186,6 +187,23 @@ func (s *SignedIgnoreTuple) Bytes() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// UniverseLeafNode returns an SMT leaf for the SignedIgnoreTuple.
```
ffranr, line 190: `returns an SMT` -> `returns an MS-SMT`

universe/interface.go
```diff
@@ -1190,3 +1191,46 @@ type Telemetry interface {
...
+type AuthenticatedIgnoreTuple struct {
+	*SignedIgnoreTuple
+
+	// IgnoreTreeRoot is the root of the ignore tree that the ignore tuple
+	// resides within.
+	IgnoreTreeRoot mssmt.Node
+
+	// InclusionProof is the universe inclusion proof for the ignore tuple
+	// within the universe tree.
+	InclusionProof *mssmt.Proof
```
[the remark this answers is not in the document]
Roasbeef: > And why is InclusionProof a pointer? Is it a pointer because the ignore could be the tree root and so no inclusion proof is necassary

We have 256 levels of the tree, and 32 bytes for each sibling hash at a given level, so it can be up to 8 KB of _just_ the space occupied by the hash (not counting any of the other fields and runtime over head). This is pretty big for a struct field to be openly coded, so I opted to to make it a pointer instead.

universe/interface.go
Roasbeef: I went back and forth a bit here: the `None` value makes it explicit when the query has no response value, while the empty slice is an implicit indicator. 

> I assume that if there are any leaves in the tree then the sum is always > 0, am I wrong? In which case None and 0 mean the same thing.

Same thing here re implicit vs explicit. If there's no tree, then a sum value isn't defined. An implicit interpretaiton is that if there's no tree, then the sum is zero.

tapdb/sqlc/queries/assets.sql
Roasbeef: The current formulation of `LookupGenesisID` is pretty simple: it doesn't involve an joins nor any of the extra fields, no CTE, etc. I prefer to have a simple query here as it'll also be very efficient. 

I used look up here, as fetch was already taken (`FetchGenesisID`). Will go with `FetchGenesisIDByAssetID`.

universe/interface.go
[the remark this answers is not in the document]
Roasbeef: `SignedIgnoreTuple` is pretty small, so will just embed it.

guggero (approved):
LGTM :tada:

ffranr (approved):

PR #1422, opened 2025-03-05
Update lnd compile-time dependency to v0.19

guggero:
This updates the compile-time dependency of `lnd` to the current `master` branch, which will be 0.19 RC1 very soon.
The diff is quite large because of the following major changes in `lnd`:
 - `lnd/fn` was updated to `lnd/fn/v2` which included some breaking API changes
 - The logging system was updated to `github.com/btcsuite/btclog/v2`
 - The `lnd` itest framework removed the standby nodes

NOTE: The `LiT` integration test is expected to fail for this PR, since this PR is required for `litd` to update. So we have a circular dependency here.


cmd/commands/commands.go
```diff
@@ -9,7 +9,7 @@ import (
 
 	tap "github.com/lightninglabs/taproot-assets"
```
ellemouton, line 10: nice split of commits, thanks 🙏

tapcfg/config.go
```diff
@@ -318,8 +315,10 @@ type Config struct {
 
 	DataDir        string `long:"datadir" description:"The directory to store tapd's data within"`
 	LogDir         string `long:"logdir" description:"Directory to log output."`
-	MaxLogFiles    int    `long:"maxlogfiles" description:"Maximum logfiles to keep (0 for no rotation)"`
-	MaxLogFileSize int    `long:"maxlogfilesize" description:"Maximum logfile size in MB"`
+	MaxLogFiles    int    `long:"maxlogfiles" hidden:"true" description:"DEPRECATED! Use logging.file.max-files instead. Maximum logfiles to keep (0 for no rotation)"`
+	MaxLogFileSize int    `long:"maxlogfilesize" hidden:"true" description:"DEPRECATED! Use logging.file.max-file-size instead. Maximum logfile size in MB"`
+
```
ellemouton, line 320: 👍

tapchannel/aux_sweeper.go
```diff
@@ -569,10 +572,16 @@ func (a *AuxSweeper) createAndSignSweepVpackets(
 		return lfn.Ok(vPkts)
 	}
 
-	return lfn.AndThen2(
+	return lfn.FlatMapResult(
```
ellemouton, line 575: nice 🙏  glad we didnt need to do `AndThen(AndThen())`

proof/proof_test.go
```diff
@@ -1467,7 +1467,10 @@ func TestProofUnknownOddType(t *testing.T) {
...
 func init() {
 	logWriter := build.NewRotatingLogWriter()
-	logger := logWriter.GenSubLogger(Subsystem, func() {})
-	logWriter.RegisterSubLogger(Subsystem, logger)
+	logConfig := build.DefaultLogConfig()
+	logMgr := build.NewSubLoggerManager(
+		build.NewDefaultLogHandlers(logConfig, logWriter)...,
+	)
+	logger := logMgr.GenSubLogger(Subsystem, func() {})
 	UseLogger(logger)
```
ellemouton, line 1475: I dont think we actually want to write to a file here right? ie, we dont need to rotating log writer anymore - we used to cause we needed that just to get the stdout logger to work but now these have been separated. So I think we can just:

```
func init() {
	logger := btclog.NewSLogger(btclog.NewDefaultHandler(os.Stdout))
	UseLogger(logger.SubSystem(Subsystem))
}

```

itest/test_harness.go
```diff
@@ -200,14 +202,19 @@ func (h *harnessTest) shutdown(_ *testing.T) error {
 func (h *harnessTest) setupLogging() {
 	h.logWriter = build.NewRotatingLogWriter()
 
+	logConfig := build.DefaultLogConfig()
+	h.logMgr = build.NewSubLoggerManager(
+		build.NewDefaultLogHandlers(logConfig, h.logWriter)...,
+	)
+
```
ellemouton, line 209: i think same thing here as previous comment - i think we just want to print to console here so dont need to setup the rotating log writer. The binaries actually used in the itest will have their own rotating log writers which will write to a file but i think normal itest logs we just write to stdout usuall?

ellemouton (approved):
LGTM! Thanks for this - huge lift fosho 🙏 

Left comments on 2 spots where the logging init can be simplified now that the rotating log writer and std-out logger have been divorced

ellemouton:
> seems like we need to do some bump in LiT too?

@GeorgeTsagk  - see the PR description

proof/proof_test.go
guggero: Ah yes, thanks! Updated.

tapchannel/aux_sweeper.go
guggero: We kind of had to... Just that `AndThen` is now called `FlatMapResult` :sweat_smile:

guggero:
Since this will break the `lit` itest on `main` when merged, I'm going to leave the PR open until we have at least the branch/PR up in `lit`, then we can point to that temporarily.

ellemouton:
opened this pr against this branch: https://github.com/lightninglabs/taproot-assets/pull/1427

guggero:
The CI isn't 100% flake free but close enough I think. So I'm going to merge this so we can continue resolving the PR dependency tree.

PR #1423, opened 2025-03-05
Support group keys on `SendPayment` & `AddInvoice`

GeorgeTsagk:
## Description

This PR takes care of the rest of the work required to create and pay invoices by using group keys.

- Using a `group_key` on `SendPayment` means that only channels that contain assets which belong to that group are going to be considered for sending this payment.
- Using a `group_key` on `AddInvoice` means that only channels that contain assets which belong to that group may be considered for generating an RFQ quote and encoding it in the invoice.

[LiT PR](https://github.com/lightninglabs/lightning-terminal/pull/995) for itest

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -7679,18 +7679,26 @@ func checkOverpayment(quote *rfqrpc.PeerAcceptedSellQuote,
 func (r *rpcServer) AddInvoice(ctx context.Context,
 	req *tchrpc.AddInvoiceRequest) (*tchrpc.AddInvoiceResponse, error) {
 
+	if len(req.AssetId) > 0 && len(req.GroupKey) > 0 {
+		return nil, fmt.Errorf("cannot set both asset id and group key")
+	}
+
 	if req.InvoiceRequest == nil {
 		return nil, fmt.Errorf("invoice request must be specified")
 	}
 	iReq := req.InvoiceRequest
 
-	// Do some preliminary checks on the asset ID and make sure we have any
-	// balance for that asset.
-	if len(req.AssetId) != sha256.Size {
-		return nil, fmt.Errorf("asset ID must be 32 bytes")
+	assetID, groupKey, err := parseAssetSpecifier(
+		req.AssetId, "", req.GroupKey, "",
+	)
+	if err != nil {
+		return nil, err
 	}
-	var assetID asset.ID
-	copy(assetID[:], req.AssetId)
+
+	specifier, err := asset.NewExclusiveSpecifier(assetID, groupKey)
+	if err != nil {
+		return nil, err
+	}
 
 	// The peer public key is optional if there is only a single asset
 	// channel.
@@ -7704,8 +7712,6 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 
 		peerPubKey = &parsedKey
 	}
-
-	specifier := asset.NewSpecifierFromId(assetID)
 
 	// We can now query the asset channels we have.
 	assetChan, err := r.rfqChannel(
@@ -7728,15 +7734,13 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 		time.Duration(expirySeconds) * time.Second,
 	)
 
+	rpcSpecifier := marshalAssetSpecifier(specifier)
+
 	resp, err := r.AddAssetBuyOrder(ctx, &rfqrpc.AddAssetBuyOrderRequest{
-		AssetSpecifier: &rfqrpc.AssetSpecifier{
-			Id: &rfqrpc.AssetSpecifier_AssetId{
-				AssetId: assetID[:],
-			},
-		},
-		AssetMaxAmt: req.AssetAmount,
-		Expiry:      uint64(expiryTimestamp.Unix()),
-		PeerPubKey:  peerPubKey[:],
+		AssetSpecifier: &rpcSpecifier,
+		AssetMaxAmt:    req.AssetAmount,
+		Expiry:         uint64(expiryTimestamp.Unix()),
+		PeerPubKey:     peerPubKey[:],
 		TimeoutSeconds: uint32(
 			rfq.DefaultTimeout.Seconds(),
 		),


rfqmsg/records.go
```diff
@@ -82,22 +84,43 @@ func (h *Htlc) Balances() []*AssetBalance {
 	return h.Amounts.Val.Balances
 }
 
+// SpecifierChecker is an interface that contains methods for checking certain
+// properties related to asset specifiers.
+type SpecifierChecker interface {
```
Roasbeef, line 89: IIRC we have another instances of this elsewhere in the codebase? 

Also is this something that can just be a function, or do we expect the actual implementation to vary depending on context?

rfq/order.go
```diff
@@ -53,12 +54,23 @@ func parseHtlcCustomRecords(customRecords map[uint64][]byte) (*rfqmsg.Htlc,
 // SerialisedScid is a serialised short channel id (SCID).
 type SerialisedScid = rfqmsg.SerialisedScid
 
+// AssetSpecifierChecker is an interface that contains methods for checking
+// certain properties related to asset specifiers.
+type AssetSpecifierChecker interface {
```
Roasbeef, line 59: Same here re duplication, I think we only need one instance, unless some of these are in place to avoid import cycles (doesn't appear so).

tapchannel/aux_invoice_manager.go
```diff
@@ -42,6 +42,12 @@ type RfqManager interface {
 	// node and have been requested by our peers. These quotes are
 	// exclusively available to our node for the sale of assets.
 	LocalAcceptedSellQuotes() rfq.SellAcceptMap
+
+	// AssetMatchesSpecifier checks if the provided asset satisfies the
+	// provided specifier. If the specifier includes a group key, we will
+	// check if the asset belongs to that group.
+	AssetMatchesSpecifier(ctx context.Context, specifier asset.Specifier,
```
Roasbeef, line 49: Similar interface comment here.

rfq/manager.go
```diff
@@ -949,6 +950,10 @@ func (m *Manager) getAssetGroupKey(ctx context.Context,
 	// Perform the DB query.
 	group, err := m.cfg.GroupLookup.QueryAssetGroup(ctx, id)
 	if err != nil {
+		if strings.Contains(err.Error(), "asset group is unknown") {
```
Roasbeef, line 954: Can we do a more specific error check here so we don't need to rely on string matching?

Roasbeef: Also when would this fail? Instances where we haven't synced the group key yet?

rfq/manager.go
```diff
@@ -982,11 +983,19 @@ func (m *Manager) AssetMatchesSpecifier(ctx context.Context,
 			return false, err
 		}
 
+		specifierGK := specifier.UnwrapGroupKeyToPtr()
+
+		// If the group lookup failed, let's see if the provided assetID
+		// is the hash of the group key, which is used by the sender to
```
Roasbeef, line 989: This is a bit confusing, why aren't we just passing around the group key isntead of this special case where a hash of it can be provided?

Roasbeef: If the look up fails, doesn't that mean we haven't verified this asset at all yet? That should have happened during funding, when we send the transition proofs of each of the assets.

tapchannel/aux_traffic_shaper.go
```diff
@@ -308,11 +310,19 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
...
+	case quote.Request.AssetSpecifier.HasId():
+		assetId = *quote.Request.AssetSpecifier.UnwrapIdToPtr()
+
+	case quote.Request.AssetSpecifier.HasGroupPubKey():
+		// If a group key is defined in the quote we place the hash of
+		// the group key as the dummy asset ID in the HTLC. This asset
+		// balance in the HTLC is just a hint and the actual asset IDs
+		// will be picked later in the process.
+		groupKey := quote.Request.AssetSpecifier.UnwrapGroupKeyToPtr()
+		assetId = sha256.Sum256(groupKey.SerializeCompressed())
```
Roasbeef, line 325: Do we actually need to start using a specifier for the HTLC balance TLVs? 

Otherwise, we're introducing some overloading here, which is likely to be the source of confusion/bugs down the line. 

As an example, if we're using a fake asset ID (which is actually the group key hash), won't all the logic to create the allocations for th actual TAP tree in the HTLC/commitment outputs fail?

rpcserver.go
```diff
@@ -7474,9 +7504,34 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+				groupKey.SerializeCompressed(),
+			)
+
+			// We can't distribute the amount over distinct asset ID
+			// balances, so we provide the total amount under the
+			// dummy asset ID that is produced by hashing the group
+			// key.
+			balances = []*rfqmsg.AssetBalance{
+				rfqmsg.NewAssetBalance(
+					groupHash, req.AssetAmount,
```
Roasbeef, line 7530: Same here again re using a group key hash instead of an actual asset ID.

rpcserver.go
```diff
@@ -7687,13 +7687,30 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+	case len(req.AssetId) > 0:
+		if len(req.AssetId) != sha256.Size {
+			return nil, fmt.Errorf("asset ID must be 32 bytes")
+		}
+		var assetID asset.ID
+		copy(assetID[:], req.AssetId)
+		specifier = asset.NewSpecifierFromId(assetID)
+
+	case len(req.GroupKey) > 0:
+		// asset.NewSpecifierFromGroupKey()
```
Roasbeef, line 7705: stray comment

Roasbeef (changes requested):

rfq/order.go
```diff
@@ -53,12 +54,19 @@ func parseHtlcCustomRecords(customRecords map[uint64][]byte) (*rfqmsg.Htlc,
 // SerialisedScid is a serialised short channel id (SCID).
 type SerialisedScid = rfqmsg.SerialisedScid
 
+// SpecifierChecker checks whether the passed specifier and asset ID match. If
+// the specifier contains a group key, it will check whether the asset belongs
+// to that group.
+type SpecifierChecker func(ctx context.Context, specifier asset.Specifier,
```
guggero, line 60: Can just use the one from `rfqmsg` IMO.

rfq/order.go
```diff
@@ -248,11 +256,22 @@ func (c *AssetSalePolicy) GenerateInterceptorResponse(
...
+		// We have performed checks for the asset IDs inside the HTLC
+		// against the specifier's group key in a previous step. Here
+		// we just need to provide a dummy value as the asset ID. The
+		// real asset IDs will be carefully picked in a later step in
+		// the process. What really matters now is the total amount.
+		assetID = sha256.Sum256(groupKey.SerializeCompressed())
+
+	case c.AssetSpecifier.HasId():
+		specifierID := *c.AssetSpecifier.UnwrapIdToPtr()
+		copy(assetID[:], specifierID[:])
```
guggero, line 274: This change seems unrelated to the rest of the commit. I think this should probably go into its own commit with its own commit message (explaining that the asset ID in the balance record doesn't really matter as the allocation code will actually take care of assigning specific asset IDs).

rfq/order.go
```diff
@@ -248,11 +256,22 @@ func (c *AssetSalePolicy) GenerateInterceptorResponse(
...
+	switch {
+	case c.AssetSpecifier.HasGroupPubKey():
+		groupKey := c.AssetSpecifier.UnwrapGroupKeyToPtr()
+
+		// We have performed checks for the asset IDs inside the HTLC
+		// against the specifier's group key in a previous step. Here
+		// we just need to provide a dummy value as the asset ID. The
+		// real asset IDs will be carefully picked in a later step in
+		// the process. What really matters now is the total amount.
+		assetID = sha256.Sum256(groupKey.SerializeCompressed())
```
guggero, line 270: I know I probably mentioned this at some point to just hash the group key. But I wonder if it makes sense to instead leave the asset ID empty? Or use a fixed, dummy value in both cases? Since I don't think we ever try to resolve this back...

rfq/order.go
```diff
@@ -53,12 +54,19 @@ func parseHtlcCustomRecords(customRecords map[uint64][]byte) (*rfqmsg.Htlc,
...
+type SpecifierChecker func(ctx context.Context, specifier asset.Specifier,
+	id asset.ID) (bool, error)
+
 // Policy is an interface that abstracts the terms which determine whether an
 // asset sale/purchase channel HTLC is accepted or rejected.
 type Policy interface {
 	// CheckHtlcCompliance returns an error if the given HTLC intercept
 	// descriptor does not satisfy the subject policy.
-	CheckHtlcCompliance(htlc lndclient.InterceptedHtlc) error
+	CheckHtlcCompliance(htlc lndclient.InterceptedHtlc,
```
guggero, line 68: I think since we're touching this method signature anyway, let's add the context here.

rfq/manager.go
guggero: We should resolve that circular dependency. Can be done by applying https://github.com/lightninglabs/taproot-assets/commit/fad16bf8171c4650dd60518d090d20aa2548561e (in the upstream branch `tinygo`).

rfq/manager.go
guggero: Hmm, so see my other inline comment. So we do look up the value we put into the dummy ID again here.
Not sure what's better, doing an explicit hash or just use the x-coordinates of the group key directly.

But what we can do in any case is do the `groupHash == id` comparison **before** querying the database. Because if that matches, we **know** there isn't an asset ID in the DB. That would probably also allow us to get rid of the previous commit.

rpcserver.go
```diff
@@ -7227,13 +7227,28 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
 	ctx := stream.Context()
 
-	// Do some preliminary checks on the asset ID and make sure we have any
-	// balance for that asset.
-	if len(req.AssetId) != sha256.Size {
-		return fmt.Errorf("asset ID must be 32 bytes")
+	var specifier asset.Specifier
+
+	switch {
+	case len(req.AssetId) > 0 && len(req.GroupKey) > 0:
```
guggero, line 7233: Could probably use `parseAssetSpecifier` here? Combined with `asset.NewSpecifier()`. Or maybe even `specifierWithGroupKeyLookup` if that makes sense here?

rpcserver.go
```diff
@@ -7358,7 +7373,26 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+
+		case specifier.HasGroupPubKey():
+			groupKey := specifier.UnwrapGroupKeyToPtr()
+			groupKeyBytes := groupKey.SerializeCompressed()
+			rpcSpecifier = rfqrpc.AssetSpecifier{
+				Id: &rfqrpc.AssetSpecifier_GroupKey{
+					GroupKey: groupKeyBytes,
+				},
+			}
+		}
```
guggero, line 7395: Probably makes sense to extract this into a `marshalAssetSpecifier` function, as this seems to appear twice in the code now.

rpcserver.go
```diff
@@ -7687,13 +7687,29 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
 	iReq := req.InvoiceRequest
 
-	// Do some preliminary checks on the asset ID and make sure we have any
-	// balance for that asset.
-	if len(req.AssetId) != sha256.Size {
-		return nil, fmt.Errorf("asset ID must be 32 bytes")
+	var specifier asset.Specifier
+
+	switch {
+	case len(req.AssetId) > 0 && len(req.GroupKey) > 0:
```
guggero, line 7693: Same here and below re: using existing functions and/or extracting into function.

guggero (commented):
Did a first pass.

Concerning the discussion around group key in the asset sum entry.
I failed to consider the group key send case when designing the p2p messages.

Once an HTLC is locked in, the `allocation` code makes sure each HTLC has a specific, clearly defined set of asset pieces, all with IDs defined.
But as long as the HTLC is just transmitted (e.g. on the `SendPayment` RPC, in `ProduceHtlcExtraData` function of the aux traffic shaper, or even in the `UpdateAddHtlc` wire message), we don't actually know the ID of the piece(s) that are going to be used for the HTLC.
So in those scenarios we only really check two things in the policies:
 - The sum of the assets being transferred in the HTLC message
 - Whether the assets match the policy

For the second check, I think it makes sense to indirectly put the group key in there. Whether we hash it first (as we do for the universe key) or whether we just use the x-coordinate of the group key doesn't matter too much IMO. But I think we need to be very clear wherever we do that and definitely add a comment on _why_ we're doing that (feel free to copy my explanation above).

rfq/order.go
guggero: With the current allocation logic, you wouldn't be able to send a specific `assetID` in an HTLC, since that would mess with the distribution algorithm.
So I think we decided in the meeting yesterday that we're going to use the x-coordinate of the group key as the asset ID in case of a grouped asset. And we'd do that everywhere where we don't have an allocated/concrete asset ID yet.

tapchannel/aux_traffic_shaper.go
```diff
@@ -308,11 +310,20 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
...
-		return 0, nil, fmt.Errorf("quote has no asset ID: %w", err)
+	var assetId asset.ID
+
+	switch {
+	case quote.Request.AssetSpecifier.HasId():
+		assetId = *quote.Request.AssetSpecifier.UnwrapIdToPtr()
+
+	case quote.Request.AssetSpecifier.HasGroupPubKey():
+		// If a group key is defined in the quote we place the hash of
+		// the group key as the dummy asset ID in the HTLC. This asset
```
guggero, line 321: nit: comment not accurate anymore.

taprpc/tapchannelrpc/tapchannel.proto
```diff
@@ -142,6 +142,10 @@ message SendPaymentRequest {
     // are sent out to the network than the invoice amount plus routing fees
     // require to be paid.
     bool allow_overpay = 6;
+
+    // The group key which dictates which assets may be used for this payment.
+    // Mutually exclusive to asset_id.
```
guggero, line 147: nit: add same "mutually exclusive" comment to asset_id as well?

rpcserver.go
```diff
@@ -7246,19 +7246,32 @@ func (r *rpcServer) EncodeCustomRecords(_ context.Context,
...
+	var specifier asset.Specifier
+
+	assetID, groupKey, err := parseAssetSpecifier(
+		req.AssetId, "", req.GroupKey, "",
+	)
+	if err != nil {
+		return err
+	}
+
+	switch {
```
guggero, line 7268: Should we add this `switch` and the error check above as a new function `asset.NewExclusiveSpecifier()`?
To make sure we use the same semantics everywhere.

taprpc/tapchannelrpc/tapchannel.proto
```diff
@@ -192,6 +192,11 @@ message AddInvoiceRequest {
     // won't be settled automatically. Instead, users will need to use the
     // invoicesrpc.SettleInvoice call to manually settle the invoice.
     HodlInvoice hodl_invoice = 5;
+
+    // The group key which dictates which assets may be accepted for this
+    // invoice. If set, any asset that belongs to this group may be accepted to
+    // settle this invoice. Mutually exclusive to asset_id.
```
guggero, line 198: Same here re adding the same comment on asset_id.

rpcserver.go
```diff
@@ -7684,18 +7684,31 @@ func checkOverpayment(quote *rfqrpc.PeerAcceptedSellQuote,
 func (r *rpcServer) AddInvoice(ctx context.Context,
 	req *tchrpc.AddInvoiceRequest) (*tchrpc.AddInvoiceResponse, error) {
 
+	if len(req.AssetId) > 0 && len(req.GroupKey) > 0 {
```
guggero, line 7687: Re-use potential, see my previous comment for `NewExclusiveSpecifier`.

guggero (commented):
Very nice!
Super super close, just a couple of nits and improvement requests.

rfq/order.go
```diff
@@ -249,11 +250,23 @@ func (c *AssetSalePolicy) GenerateInterceptorResponse(
...
+	// the specifier's group key in a previous step. Here we just need to
+	// provide a dummy value as the asset ID. The real asset IDs will be
+	// carefully picked in a later step in the process. What really matters
+	// now is the total amount.
+	switch {
+	case c.AssetSpecifier.HasGroupPubKey():
+		groupKey := c.AssetSpecifier.UnwrapGroupKeyToPtr()
+		groupKeyX := schnorr.SerializePubKey(groupKey)
+
+		assetID = asset.ID(groupKeyX)
```
Roasbeef, line 265: This is nice compared to the hash version, as this way we aren't actually losing any information. 

However, it's possible that a normal asset ID, can actually be interpreted as a valid x coordinate. To avoid this confusion, when we decode on the other side, and conclude it _might_ be a pubkey, we should check our local db to make sure it's actually a group key that we've validated.

rfq/manager.go
```diff
@@ -949,6 +951,10 @@ func (m *Manager) getAssetGroupKey(ctx context.Context,
 	// Perform the DB query.
 	group, err := m.cfg.GroupLookup.QueryAssetGroup(ctx, id)
 	if err != nil {
+		if errors.Is(err, address.ErrAssetGroupUnknown) {
```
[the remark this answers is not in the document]
Roasbeef: > An actual group lookup would have to be performed for that set of inputs. If that asset doesn't belong to a group we would run into the above line.

Seems related to the first comment I left in this round of review. Looks like we _would_ do a DB look up to make sure that we didn't accidentally interpret an asset ID as a group key.

rfq/manager.go
```diff
@@ -978,6 +979,18 @@ func (m *Manager) AssetMatchesSpecifier(ctx context.Context,
 
 	switch {
 	case specifier.HasGroupPubKey():
+		specifierGK := specifier.UnwrapGroupKeyToPtr()
+
+		// Let's directly check if the ID is equal to the hash of the
```
Roasbeef, line 984: hash -> encoded group key

Roasbeef (approved):
LGTM ❄️


I dig the change to move to using the x-coord instead of the hash. 

Changes look good pending the final comments from the other reviewer. 

Will check out the corresponding itest changes now.

rpcserver.go
```diff
@@ -7246,19 +7246,27 @@ func (r *rpcServer) EncodeCustomRecords(_ context.Context,
 func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
 	stream tchrpc.TaprootAssetChannels_SendPaymentServer) error {
 
+	if len(req.AssetId) > 0 && len(req.GroupKey) > 0 {
```
guggero, line 7249: Should we move this check into `NewExclusiveSpecifier` for when both is set?

rpcserver.go
guggero: This check can be moved as well.

guggero (approved):
One last simplification request, otherwise LGTM :tada: 

Pending @Roasbeef's other comments.

rpcserver.go
guggero: Yeah, I guess that makes sense :+1:

PR #1427, opened 2025-03-07
tapchannel: check custom records contain assets

ellemouton:
Currently (*AuxTrafficShaper).ProduceHtlcExtraData exits early if the htlcCustomRecords are empty. However, we need to cater for the case when htlc custom records exist from other contexts (such as endorsement signaling) and so we change this check to be more strict and to check that the records dont contain any asset records.


guggero (approved):
Thanks for the fix, LGTM :tada:

PR #1447, opened 2025-03-19
tapcli: Global flags to accept ENV vars overrides.

gcaracuel:
Adds ability to global flags to de config using ENV variables.
This is useful when running `tapcli` in containers and for any reason not using default values or running lightning-terminal in integrated mode.


guggero (approved):
Nice, thank you! LGTM :Tada:

guggero:
The linter isn't happy about a line that's too long. The rest of the failures look like flakes.

PR #1448, opened 2025-03-21
Allow setting sats/msats to `taprpc.AddInvoice`

### LINKED ISSUE #1306: fix typo in error message
At

https://github.com/lightninglabs/taproot-assets/blob/2e343d75ce575362aae997a1150c5a28418f6f74/rpcserver.go#L7565

"over" should be replaced with "of".

### LINKED ISSUE #1440: [feature]: generate asset invoice with fix amount of SATs instead of asset amount
We are working on receiving assets on LNURL /LN Address, currently, when a user requests an LN invoice for a fixed amount in SATs (e.g., 1100 SATs) for given LNURL , since we know that this LNURL is linked to asset we convert this amount into an equivalent USDT-L (Taproot Asset) based on our exchange rate and generate a USDT-L invoice in the background. However, due to slight fluctuations in the exchange rate or rounding mechanisms at price oracle, the SATs value of the generated invoice may slightly differ (e.g., 1105 SATs instead of 1100).

This discrepancy causes an issue where the payer's wallet (expecting an invoice of exactly 1100 SATs) rejects the payment since the decoded invoice shows a different SATs amount than initially requested.

We propose enhancing the "generate invoice" API in Tapd to allow specifying the invoice amount in SATs rather then asset amount while still issuing an asset-backed (USDT-L) invoice. This means:

The requested SATs amount remains fixed (e.g., 1100 SATs).
The backend handles the conversion and ensures the asset invoice (USDT-L) internally matches the requested SATs amount.
The payer sees an invoice with the expected SATs amount, preventing wallet rejection.
This would allow seamless integration of USDT-L payments within LNURL without breaking existing user expectations.

Wallet-side adaptation: Requesting LN wallets to support a small margin of deviation in invoice amounts, but this is not widely supported and would require broad ecosystem changes.

**Additional context**
This issue primarily affects Lightning Address (LNURL) payments where users expect fixed SATs amounts. If TAPD allows asset-backed invoices with a fixed SATs value, it would enhance stablecoin interoperability in Lightning without breaking user expectations.

Below is the current journey of payment
1) Alice wants to send 1100 SATs on LN Address of Bob
2) Alice Pings our wallet for an LN Invoice of 1100 SATs
3) At wallet we know that this LN Address created by Bob is meant for USDT-L so he is expecting his asset (USDT-L) balance to increase.
4) We convert the 1100 SATs to say 1.1 USDT-L based on exchange rate and generate invoice of it, and due to price oracle at node the SATs value of invoice become 1105 SATs.
5) When Alice receives invoice and decode it it sees that amount in invoice is 1105 SATs
6) Now since the amount is different (1105)  then what was requested (1100) the wallet of Alice denies the operation.

GeorgeTsagk:
## Description

This PR allows for the user to set the invoice value via the `Value`/`ValueMsat` fields.

Previously we'd take the asset units and the agreed upon quote, then we'd convert that to an msat amount. Now the value of the invoice can be directly set via the previously ignored fields, while also still acquiring a sufficient quote and returning that to the user.

Closes #1440 
Fixes #1306

### DIFF
--- a/rfq/manager.go
+++ b/rfq/manager.go
@@ -15,6 +15,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/address"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/rfqmath"
 	"github.com/lightninglabs/taproot-assets/rfqmsg"
 	lfn "github.com/lightningnetwork/lnd/fn/v2"
 	"github.com/lightningnetwork/lnd/lnutils"
@@ -1060,6 +1061,33 @@ func (m *Manager) publishSubscriberEvent(event fn.Event) {
 			return true
 		},
 	)
+}
+
+// EstimateAssetUnits is a helper function that queries our price oracle to find
+// out how many units of an asset are needed to evaluate to the provided amount
+// in milli satoshi.
+func EstimateAssetUnits(ctx context.Context, oracle PriceOracle,
+	specifier asset.Specifier,
+	amtMsat lnwire.MilliSatoshi) (uint64, error) {
+
+	oracleRes, err := oracle.QueryBidPrice(
+		ctx, specifier, fn.None[uint64](), fn.Some(amtMsat),
+		fn.None[rfqmsg.AssetRate](),
+	)
+	if err != nil {
+		return 0, err
+	}
+
+	if oracleRes.Err != nil {
+		return 0, fmt.Errorf("cannot query oracle: %v",
+			oracleRes.Err.Error())
+	}
+
+	assetUnits := rfqmath.MilliSatoshiToUnits(
+		amtMsat, oracleRes.AssetRate.Rate,
+	)
+
+	return assetUnits.ScaleTo(0).ToUint64(), nil
 }
 
 // PeerAcceptedBuyQuoteEvent is an event that is broadcast when the RFQ manager
--- a/rfq/manager.go
+++ b/rfq/manager.go
@@ -1012,6 +1012,12 @@ func (m *Manager) AssetMatchesSpecifier(ctx context.Context,
 	}
 }
 
+// GetPriceDeviationPpm returns the configured price deviation in ppm that is
+// used in rfq negotiations.
+func (m *Manager) GetPriceDeviationPpm() uint64 {
+	return m.cfg.AcceptPriceDeviationPpm
+}
+
 // ChannelCompatible checks a channel's assets against an asset specifier. If
 // the specifier is an asset ID, then all assets must be of that specific ID,
 // if the specifier is a group key, then all assets in the channel must belong
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -7898,17 +7898,6 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 func validateInvoiceAmount(acceptedQuote *rfqrpc.PeerAcceptedBuyQuote,
 	requestAssetAmount uint64) (lnwire.MilliSatoshi, error) {
 
-	// If the invoice is for an asset unit amount smaller than the minimal
-	// transportable amount, we'll return an error, as it wouldn't be
-	// payable by the network.
-	if acceptedQuote.MinTransportableUnits > requestAssetAmount {
-		return 0, fmt.Errorf("cannot create invoice over %d asset "+
-			"units, as the minimal transportable amount is %d "+
-			"units with the current rate of %v units/BTC",
-			requestAssetAmount, acceptedQuote.MinTransportableUnits,
-			acceptedQuote.AskAssetRate)
-	}
-
 	// Now that we have the accepted quote, we know the amount in Satoshi
 	// that we need to pay. We can now update the invoice with this amount.
 	//
@@ -7928,6 +7917,17 @@ func validateInvoiceAmount(acceptedQuote *rfqrpc.PeerAcceptedBuyQuote,
 	newInvoiceAmtMsat := rfqmath.UnitsToMilliSatoshi(
 		assetAmount, *askAssetRate,
 	)
+
+	// If the invoice is for an asset unit amount smaller than the minimal
+	// transportable amount, we'll return an error, as it wouldn't be
+	// payable by the network.
+	if acceptedQuote.MinTransportableUnits > requestAssetAmount {
+		return 0, fmt.Errorf("cannot create invoice for %d asset "+
+			"units, as the minimal transportable amount is %d "+
+			"units with the current rate of %v units/BTC",
+			requestAssetAmount, acceptedQuote.MinTransportableUnits,
+			acceptedQuote.AskAssetRate)
+	}
 
 	return newInvoiceAmtMsat, nil
 }
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -7746,11 +7746,24 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 		time.Duration(expirySeconds) * time.Second,
 	)
 
+	// We now want to calculate the upper bound of the RFQ order, which
+	// either is the asset amount specified by the user, or the converted
+	// satoshi amount of the invoice, expressed in asset units, using the
+	// local price oracle's conversion rate.
+	maxUnits, err := calculateAssetMaxAmount(
+		ctx, r.cfg.PriceOracle, specifier, req.AssetAmount, iReq,
+		r.cfg.RfqManager.GetPriceDeviationPpm(),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error calculating asset max "+
+			"amount: %w", err)
+	}
+
 	rpcSpecifier := marshalAssetSpecifier(specifier)
 
 	resp, err := r.AddAssetBuyOrder(ctx, &rfqrpc.AddAssetBuyOrderRequest{
 		AssetSpecifier: &rpcSpecifier,
-		AssetMaxAmt:    req.AssetAmount,
+		AssetMaxAmt:    maxUnits,
 		Expiry:         uint64(expiryTimestamp.Unix()),
 		PeerPubKey:     peerPubKey[:],
 		TimeoutSeconds: uint32(
@@ -7784,7 +7797,7 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 	// Satoshi that we need to pay. We can now update the invoice with this
 	// amount.
 	invoiceAmtMsat, err := validateInvoiceAmount(
-		acceptedQuote, req.AssetAmount,
+		acceptedQuote, req.AssetAmount, iReq,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("error validating invoice amount: %w",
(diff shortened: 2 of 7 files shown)


rpcserver.go
```diff
@@ -7676,13 +7699,52 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+		maxUnits = assetUnits.ToUint64()
+	}
+
+	// Since we used a different oracle price query above calculate the max
+	// amount of units, we want to add some breathing room to account for
+	// price fluctuations caused by the small time delay, plus the fact that
+	// the agreed upon quote may be different. If we don't add this safety
+	// window the peer may allow a routable amount that evaluates to less
+	// than what we ask for.
+	maxUnits = (maxUnits * 105) / 100
```
ZZiigguurraatt, line 7739: Seems like this should be greater than or equal to `experimental.rfq.acceptpricedeviationppm` instead of being hard coded.

rpcserver.go
```diff
@@ -7676,13 +7699,44 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+			fn.None[rfqmsg.AssetRate](),
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		if oracleRes.Err != nil {
+			return nil, fmt.Errorf("cannot query oracle: %v",
+				oracleRes.Err.Error())
+		}
```
Roasbeef, line 7724: Would be good to start to split up some of this logic into helper functions. Looks like we have the distinct phases of: 
  * initial parsing/validation 
  * creation of the receive/buy order 
  * invoice creation

rpcserver.go
```diff
@@ -7735,12 +7778,44 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+		return nil, fmt.Errorf("cannot create invoice over %d asset "+
+			"units, as the minimal transportable amount is %d "+
+			"units with the current rate of %v units/BTC",
+			invUnits, acceptedQuote.MinTransportableUnits,
+			acceptedQuote.AskAssetRate)
+	}
+
+	switch {
+	case satsMode:
+		amtMsat := lnwire.MilliSatoshi(req.InvoiceRequest.ValueMsat +
```
Roasbeef, line 7812: Same comment here re just accepting one or the other.

Roasbeef: We also already computed this up above.

Roasbeef (changes requested):
Nice work! Pretty straight forward implementation. The main thing we need to ensure is that the negotiated quote has a total amount that's greater than the sats input of the user.

rpcserver.go
```diff
@@ -7637,6 +7637,29 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 	if len(req.AssetId) != sha256.Size {
 		return nil, fmt.Errorf("asset ID must be 32 bytes")
 	}
+
+	// Let's make sure that only one type of amount is set, in order to
+	// avoid ambiguous behavior. This field dictates the actual value of the
+	// invoice so let's be strict and only allow one possible value to be
+	// set.
+	if req.AssetAmount > 0 && (req.InvoiceRequest.ValueMsat > 0 ||
```
saraogiraj94, line 7645: shouldn't the proto files be updated in the same Merge Request ? like adding of new keys ValueMsat

ZZiigguurraatt:
OK, looks like I am having the same problem if I try to build litd with the taproot assets main branch tip too.

guggero:
> @guggero Not sure if this issue is the cause?
> 
> https://github.com/lightninglabs/taproot-assets/blame/5290f9c3e811bb327a8f8a446d7b3519daf2b453/rfq/order.go#L16

This will be required first: https://github.com/lightninglabs/lightning-terminal/pull/998

ZZiigguurraatt:
> > @guggero Not sure if this issue is the cause?
> > https://github.com/lightninglabs/taproot-assets/blame/5290f9c3e811bb327a8f8a446d7b3519daf2b453/rfq/order.go#L16
> 
> This will be required first: [lightninglabs/lightning-terminal#998](https://github.com/lightninglabs/lightning-terminal/pull/998)

OK, so https://github.com/lightninglabs/taproot-assets/blame/5290f9c3e811bb327a8f8a446d7b3519daf2b453/rfq/order.go#L16 is not the cause, but instead, the fix for taproot-assets and other dependencies are basically missing that same fix?

guggero:
The update to lnd 19 was already done in `taproot-assets`, which came with a bunch of changes. Those changes also need to be done in `litd` for things to be compile-time compatible. Which is what the PR I linked to does.

guggero:
Correct. I'm working on it.

rpcserver.go
```diff
@@ -7637,6 +7637,29 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+		req.InvoiceRequest.Value > 0) {
+
+		return nil, fmt.Errorf("cannot set both asset amount and sats" +
+			"amount")
+	}
+
+	if req.InvoiceRequest.ValueMsat > 0 && req.InvoiceRequest.Value > 0 {
```
guggero, line 7652: We should use `lnrpc.UnmarshallAmt()`, since it already does all this validation and conversion. That way we also know we're using the same rules as the actual `lnd` backend the payment request will be handed to.

rpcserver.go
```diff
@@ -7735,12 +7778,44 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
-	// Convert the asset amount into a fixed-point.
-	assetAmount := rfqmath.NewBigIntFixedPoint(req.AssetAmount, 0)
+	invUnits := req.AssetAmount
+
+	// If the invoice was created over a satoshi amount, we need to
+	// calculate the units.
+	if satsMode {
+		invUnits = rfqmath.MilliSatoshiToUnits(
+			amtMsat, *askAssetRate,
+		).ToUint64()
```
guggero, line 7816: We always need to use `ScaleTo(0).` before calling `ToUint64()`, otherwise we get incorrect results!

rpcserver.go
guggero: nit: `lnwire.NewMSatFromSatoshis()` instead of multiplying. Also, formatting.

rpcserver.go
```diff
@@ -7735,12 +7778,44 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+	}
+
+	switch {
+	case satsMode:
+		amtMsat := lnwire.MilliSatoshi(req.InvoiceRequest.ValueMsat +
+			req.InvoiceRequest.Value*(1000))
+
+		iReq.ValueMsat = int64(amtMsat)
+
+	default:
```
guggero, line 7849: IMO all this calculation logic with different possible paths has become quite complex and hard to read and hard to test as well. Should we extract the logic into a method and write some unit tests for it?

rpcserver.go
```diff
@@ -7730,6 +7730,34 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+	// than what we ask for.
+	tolerance := rfqmath.NewBigIntFromUint64(
+		r.cfg.RfqManager.GetPriceDeviationPpm(),
+	)
+
+	// Parse the max asset units to the rfqmath type.
+	maxMathUnits := rfqmath.NewBigIntFromUint64(maxUnits)
+
+	// Calculate the tolerance margin.
+	toleranceUnits := maxMathUnits.Mul(tolerance).Div(
```
guggero, line 7747: See `WithinTolerance()`.

guggero (commented):
Did a quick pass.

rpcserver.go
Roasbeef: Sounds like a similar sentiment to this comment: https://github.com/lightninglabs/taproot-assets/pull/1448/commits/e5ea7bcf227b9fe6569f618ceec309fa09dcbae1#r2008167666

rpcserver.go
Roasbeef: The advantage of using the existing method is that it's a standardized check elsewhere, and the method has tests (also note the recent bug fix in that area).

Roasbeef: Ah I think the difference here is that we want the absolute value that would be in tolerance vs determining if something is already within our tolerance range. 

Either way, I think we should extract this to a function so we can write tests for it. Generally we should try to minimize the amount of new biz logic we add to the rpc server functinos.

Roasbeef: We should also add some logging here so we can pick up on any abnormalities w.r.t the calculation.

Roasbeef (changes requested):

rfq/manager.go
```diff
@@ -1044,6 +1045,33 @@ func (m *Manager) publishSubscriberEvent(event fn.Event) {
 	)
 }
 
+// EstimateAssetUnits is a helper function that queries our price oracle to find
+// out how many units of an asset are needed to evaluate to the provided amount
+// in msats.
+func EstimateAssetUnits(ctx context.Context, oracle PriceOracle,
```
guggero, line 1051: I think we need to add the word "bid" to this function, since we call into the `QueryBidPrice`.

But I'm not sure that even is the correct RPC to call into?
Because today for the `AddInvoice` flow with asset units, we do the following:
 - Query the local oracle for a rate hint, using `queryBidFromPriceOracle`.
 - Create a `BuyRequest` that we send to our peer, including the hint we got from our local oracle
 - The peer will call into `queryAskFromPriceOracle` (see `HandleIncomingBuyRequest`)
 - The ask price will likely have a spread on it, since it's on the "opposite" side of the spread gap

So if we add the call to `QueryBidPrice` to our local oracle to find out the initial asset amount we start the flow above with, I think we end up with at least the spread percentage of more sats in the invoice than the user wanted.

Therefore I think we should call into the `QueryAskPrice` here instead?

rfq/manager.go
```diff
@@ -1044,6 +1045,33 @@ func (m *Manager) publishSubscriberEvent(event fn.Event) {
...
+	if oracleRes.Err != nil {
+		return 0, fmt.Errorf("cannot query oracle: %v",
+			oracleRes.Err.Error())
+	}
+
+	assetUnits := rfqmath.MilliSatoshiToUnits(
+		amtMsat, oracleRes.AssetRate.Rate,
+	)
+
+	return assetUnits.ToUint64(), nil
```
guggero, line 1072: Always scale down to 0 before using `ToUint64()`.

guggero (commented):
Thanks for the update! Have some questions around what rate to use for the calculation.

taprpc/tapchannelrpc/tapchannel.proto
```diff
@@ -175,13 +175,16 @@ message AddInvoiceRequest {
...
-    // for more details). The value/value_msat fields will be overwritten by the
-    // satoshi (or milli-satoshi) equivalent of the asset amount, after
-    // negotiating a quote with a peer that supports the given asset ID.
+    // for more details).
+    //
+    // Only one of the asset_amount/value/value_msat may be set to dictate the
+    // value of the invoice. When using asset_amount, the value/value_msat
+    // fields will be overwritten by the satoshi (or milli-satoshi) equivalent
+    // of the asset amount, after negotiating a quote with a peer that supports
+    // the given asset ID.
```
ZZiigguurraatt, line 187: Same comment as below. Do we want to clarify that when specifying value/value_msat that we still really receive assets?

ZZiigguurraatt:
When specifying value/value_msat, would it be beneficial to return `asset_amount` in `tapchannelrpc.AddInvoiceResponse` so people don't have to calculate it from the `accepted_buy_quote`? Might just make it clearer to people what this is doing.

rpcserver.go
```diff
@@ -7621,6 +7621,147 @@ func checkOverpayment(quote *rfqrpc.PeerAcceptedSellQuote,
...
+			assetAmount, *askAssetRate,
+		)
+		invoiceValueMsat = int64(valMsat)
+	}
+
+	// If the invoice is for an asset unit amount smaller than the minimal
+	// transportable amount, we'll return an error, as it wouldn't be
+	// payable by the network.
+	if acceptedQuote.MinTransportableUnits > invUnits {
+		return 0, fmt.Errorf("cannot create invoice over %d asset "+
```
ZZiigguurraatt, line 7755: Please fix this issue now instead of carry it over: https://github.com/lightninglabs/taproot-assets/issues/1306

rpcserver.go
```diff
@@ -7621,6 +7621,147 @@ func checkOverpayment(quote *rfqrpc.PeerAcceptedSellQuote,
...
+	var invoiceValueMsat int64
+	switch {
+	case amtMsat != 0:
+		// If the invoice was created over a satoshi amount, we need to
+		// calculate the units.
+		invUnits = rfqmath.MilliSatoshiToUnits(
+			amtMsat, *askAssetRate,
+		).ScaleTo(0).ToUint64()
+
+		// Now let's see if the negotiated quote can actually route the
```
ZZiigguurraatt, line 7722: are we actually checking if this quote can "route" the payment? seems like this has nothing to do with routing through channels, more if the RFQ is big enough to accept it? or do we still consider this part of "routing" because the HTLC will be rejected if it doesn't fit within the RFQ?

ZZiigguurraatt: Yes, we have https://github.com/lightningnetwork/lnd/issues/9380 that could add such checks for AddInvoice.

Just thinking the word "route" should not be used here so we don't confuse people to think we are actually checking if the payment could really be routed.

guggero:
I pushed up a version that simplifies the commit structure, so we don't touch the same code multiple times. Should also make it easier to spot what exactly in the logic is changed (don't want to introduce a potential siphon or other kind of exploitable vulnerability here, so yes, I'm extra strict with reviewer friendliness of commit structure).

guggero (approved):

guggero:
Litd test will succeed after merging https://github.com/lightninglabs/lightning-terminal/pull/1040.

guggero:
Also fixes https://github.com/lightninglabs/taproot-assets/issues/1306.

Roasbeef:
Seeing this failure in the litd itest;, not sure if related: 
```
    assets_test.go:1428: Asking peer 03a024e1ee071b31d510d75a7794f979ed1131b1b021492f7613f82fda9530c84a for quote to buy assets to receive for invoice over 1 units; waiting up to 300s
    assets_test.go:1445: 
        	Error Trace:	/home/runner/work/taproot-assets/taproot-assets/lightning-terminal/itest/assets_test.go:1445
        	            				/home/runner/work/taproot-assets/taproot-assets/lightning-terminal/itest/litd_custom_channels_test.go:2163
        	            				/home/runner/work/taproot-assets/taproot-assets/lightning-terminal/itest/test_harness.go:103
        	            				/home/runner/work/taproot-assets/taproot-assets/lightning-terminal/itest/litd_test.go:87
        	Error:      	Error "rpc error: code = Unknown desc = error validating invoice amount: cannot create invoice for 1 asset units, as the minimal transportable amount is 20 units with the current rate of coefficient:\"5820600\" units/BTC" does not contain "cannot create invoice over 1 asset units, as the minimal transportable amount"
        	Test:       	TestLightningTerminal/custom_channels_liquidity
    harness.go:381: finished test: , start height=438, end height=483, mined blocks=45
    harness.go:403: !============================================!
        Too many blocks (45) mined in one test! Tips:
        1. break test into smaller individual tests, especially if this is a table-drive test.
        2. use smaller CSV via `--bitcoin.defaultremotedelay=1.`
        3. use smaller CLTV via `--bitcoin.timelockdelta=18.`
        4. remove unnecessary CloseChannel when test ends.
        5. use `CreateSimpleNetwork` for efficient channel creation.
 ```

Roasbeef:
Ah I see [this comment now ](https://github.com/lightninglabs/taproot-assets/pull/1448#issuecomment-2808956546), so does this just need a rebase now?

Roasbeef (approved):
LGTM 🥦

guggero:
Litd itest is expected to fail now, because the referenced branch requires https://github.com/lightninglabs/taproot-assets/pull/1478 to be merged. So this will become stable again soon.

PR #1453, opened 2025-03-22
Proof-system-v1

gijswijs:
This pull request introduces several enhancements and refactors related to inclusion and exclusion proof handling. The key changes include:

- STXO Commitment and Proofs:
  - Implemented commitment to spent assets within Taproot Assets using `AltLeaves`.
  - Added encoders and decoders for CommitmentProof structures that are used for the STXO proofs.
  - Introduced STXO inclusion proofs and proof verification for non-split assets.
  - Added STXO exclusion proofs and verification for the outputs that don't involve the asset spending the STXOs
- Code Refactoring and Improvements:
  - Minor refactors of code throughout
  - Improved wording in comments for clarity.
  - Removed unnecessary parameters and assertions
  -  Refactor the code of `proof/verifier.go`
- Testing:
  - Added integration tests for STXO inclusion proofs.
  - Added integration tests for STXO exclusion proofs.
  - Implemented unit tests for CommitmentProofs encoding and decoding.

Considerations
- Potentially remove the commit `multi: remove AssertInputsUnique`. Although with this proof system we are able to parse a proof for non-unique inputs, older nodes cannot. So for now this check should probably remain in place.
- Explore adding proofs to asset splits.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1453)


proof/append.go
```diff
@@ -160,6 +160,47 @@ func CreateTransitionProof(prevOut wire.OutPoint,
...
+			map[asset.SerializedKey]commitment.Proof,
+			len(proof.Asset.PrevWitnesses),
+		)
+		for _, wit := range proof.Asset.PrevWitnesses {
+			if wit.PrevID == nil {
+				return nil, fmt.Errorf("witness %v has no "+
+					"prevID", wit)
+			}
+
+			spentAsset := &asset.Asset{
```
Roasbeef: Shouldn't this be using the constructor to make alt leaves?

proof/append.go
```diff
@@ -160,6 +160,47 @@ func CreateTransitionProof(prevOut wire.OutPoint,
...
+				asset.EmptyGenesisID,
+				spentAsset.AssetCommitmentKey(),
+			)
+			if err != nil {
+				return nil, err
+			}
+
+			key := spentAsset.ScriptKey.PubKey.SerializeCompressed()
+
+			stxoInclusionProofs[asset.SerializedKey(key)] =
```
Roasbeef, line 196: Can make the `SerializedKey` above to fix this dangling line.

proof/append.go
```diff
@@ -191,6 +232,7 @@ func CreateTransitionProof(prevOut wire.OutPoint,
 			return nil, fmt.Errorf("root asset mismatch")
 		}
 
+		// TODO(jhb): add STXO inclusion proof for root prevIDs
```
Roasbeef, line 239: Marker to be resolved before merge?

proof/append.go
```diff
@@ -160,6 +160,47 @@ func CreateTransitionProof(prevOut wire.OutPoint,
...
+
+			stxoInclusionProofs[asset.SerializedKey(key)] =
+				*stxoProof
+		}
+
+		if len(stxoInclusionProofs) == 0 {
+			return nil, fmt.Errorf("no stxo inclusion proofs")
+		}
+
+		proof.InclusionProof.CommitmentProof.STXOProofs =
```
Roasbeef, line 204: Still getting through the PR, but a reminder to also add exclusion proofs: a prevID in a given output needs to prove that it isn't in the STXO for any other output (that has a TAP commitment).

proof/mint.go
```diff
@@ -158,6 +158,23 @@ func (p *BaseProofParams) HaveInclusionProof(anchorOutputIndex uint32) bool {
...
 
+// HaveSTXOInclusionProof returns true if the inclusion proof is for the given
+// anchor output index.
+func (p *BaseProofParams) HaveSTXOExclusionProof(
+	anchorOutputIndex uint32) bool {
+
+	for _, proof := range p.ExclusionProofs {
+		if proof.OutputIndex == anchorOutputIndex &&
+			proof.CommitmentProof != nil &&
+			len(proof.CommitmentProof.STXOProofs) > 0 {
```
Roasbeef, line 169: How are we distinguishing between the inclusion and exclusion proofs? Don't we need to maps here? 

Also this might be a chance to clean up some of the structure of the proofs. Eg: being more explicit about when something is an inclusion proof or exclusion proof, etc.

Roasbeef: Or is the idea that for a given output all the prevIDs are unique as their based on outpoints? 

I think we should try to be as explicit as possible with the logic here to avoid inadvertent bugs.

tapsend/proof.go
```diff
@@ -326,9 +339,86 @@ func proofParams(finalTx *wire.MsgTx, vPkt *tappsbt.VPacket,
...
+	stxoAssets, err := asset.CollectSTXO(newAsset)
+	if err != nil {
+		return fmt.Errorf("error collecting STXO assets: %w", err)
+	}
+
+	for idx := range outputs {
+		vOut := outputs[idx]
+
+		outIndex := vOut.AnchorOutputIndex
+		// We can use `HaveInclusionProof` here because it is just a
```
Roasbeef, line 360: Style nit: missing space above.

tapsend/proof.go
```diff
@@ -326,9 +339,86 @@ func proofParams(finalTx *wire.MsgTx, vPkt *tappsbt.VPacket,
...
+// addSTXOExclusionProofs adds exclusion proofs for all the STXOs of the asset,
+// for all the outputs that are asset outputs but haven't been processed yet,
+// otherwise they'll be skipped. This should only be called after
+// `addOtherOutputExclusionProofs` because it depends on
+// `params.ExclusionProofs` already being set.
+func addSTXOExclusionProofs(outputs []*tappsbt.VOutput,
+	newAsset *asset.Asset, params *proof.TransitionParams,
+	outputCommitments map[uint32]*commitment.TapCommitment) error {
+
+	stxoAssets, err := asset.CollectSTXO(newAsset)
```
Roasbeef, line 351: 👍

tapsend/proof.go
```diff
@@ -326,9 +339,86 @@ func proofParams(finalTx *wire.MsgTx, vPkt *tappsbt.VPacket,
...
+			identifier := asset.ToSerialized(pubKey)
+
+			_, exclusionProof, err := tapTree.Proof(
+				stxoAsset.TapCommitmentKey(),
+				stxoAsset.AssetCommitmentKey(),
+			)
+			if err != nil {
+				return err
+			}
+
```
Roasbeef, line 382: Perhaps we should have a sanity check that this is actually an exclusion proof as a sanity check? I think we can rely on the empty hash abstraction here.

itest/test_list_on_test.go
```diff
@@ -5,6 +5,10 @@ package itest
 import "github.com/lightninglabs/taproot-assets/proof"
 
 var testCases = []*testCase{
+	{
+		name: "psbt stxo exclusion proofs",
+		test: testPsbtSTXOExclusionProofs,
```
Roasbeef, line 10: Don't see this new test added in this commit.

proof/verifier.go
```diff
@@ -180,39 +180,56 @@ func (p *Proof) verifySplitRootProof() error {
...
-	for _, exclusionProof := range p.ExclusionProofs {
-		exclusionProof := exclusionProof
-		derivedCommitment, err := verifyTaprootProof(
-			&p.AnchorTx, &exclusionProof, &p.Asset, false,
-		)
-		if err != nil {
-			return nil, err
-		}
+	// Early pass to determine if we're dealing with v1 or v2 proofs.
+	hasV2Proofs := len(p.ExclusionProofs) > 0 &&
```
Roasbeef, line 253: We should add an explicit version here.

proof/verifier.go
```diff
@@ -221,30 +238,203 @@ func (p *Proof) verifyExclusionProofs() (*commitment.TapCommitmentVersion,
...
 
-	// All ExclusionProofs must have similar versions.
-	firstCommitVersion := maps.Values(commitVersions)[0]
-	for outputIdx, commitVersion := range commitVersions {
-		outputCommitVersion := commitVersion
-		if !commitment.IsSimilarTapCommitmentVersion(
-			&firstCommitVersion, &outputCommitVersion,
-		) {
+	// All proofs must have similar versions.
+	firstVersion := maps.Values(commitVersions)[0][0]
```
Roasbeef, line 292: Verify this with an assertion here? `lfn.All`

proof/verifier.go
```diff
@@ -180,39 +180,56 @@ func (p *Proof) verifySplitRootProof() error {
...
+	commitVersions := make(map[uint32][]commitment.TapCommitmentVersion)
 
-		// Store the commitment version. If there was no Taproot Asset
-		// commitment present, then there is nothing to store.
-		if derivedCommitment != nil {
-			commitVersions[outputIdx] = derivedCommitment.Version
+	// Verify all of the encoded v1 exclusion proofs.
+	for idx := range p.ExclusionProofs {
+		exclusionProof := p.ExclusionProofs[idx]
+		err := p.handleBasicExclusionProof(&exclusionProof,
```
Roasbeef, line 274: Style nit: Should make a new ilne after the first variable, close parens on the line after.

proof/verifier.go
```diff
@@ -221,30 +238,203 @@ func (p *Proof) verifyExclusionProofs() (*commitment.TapCommitmentVersion,
...
+		stxoAsset := stxoAssets[idx].(*asset.Asset)
+		key := asset.ToSerialized(stxoAsset.ScriptKey.PubKey)
+		assetMap[key] = stxoAsset
+		for outIdx := range p2trOutputs {
+			p2trOutputs[outIdx].Add(key)
+		}
+	}
+
+	for idx := range p.ExclusionProofs {
+		baseProof := p.ExclusionProofs[idx]
```
Roasbeef, line 321: Ah that's right, we do have an explicit field for `ExclusionProofs` here.

proof/verifier.go
```diff
@@ -221,30 +238,203 @@ func (p *Proof) verifyExclusionProofs() (*commitment.TapCommitmentVersion,
...
+		if baseProof.CommitmentProof == nil {
```
Roasbeef, line 322: We're ok with this being nil here as we assume v1 vs v2 proofs to be mutually exclusive?

proof/verifier.go
```diff
@@ -221,30 +238,203 @@ func (p *Proof) verifyExclusionProofs() (*commitment.TapCommitmentVersion,
...
+			err := p.handleBasicExclusionProof(&baseProof,
```
Roasbeef, line 323: Same line wrapping comment here, won't add others as to not spam things.

itest/psbt_test.go
```diff
@@ -2454,6 +2455,196 @@ func testPsbtTrustlessSwap(t *harnessTest) {
 	require.Equal(t.t, bobScriptKeyBytes, bobAssets.Assets[0].ScriptKey)
 }
 
+// testPsbtSTXOExclusionProofs tests that we can properly send normal assets
+// back and forth, using partial amounts, between nodes with the use of PSBTs,
+// and that we see the expected STXO exclusion proofs.
+func testPsbtSTXOExclusionProofs(t *harnessTest) {
```
Roasbeef, line 2461: Looks this this test should have gone in a prior commit.

proof/taproot.go
```diff
@@ -429,7 +429,7 @@ func (p TaprootProof) DeriveByAssetExclusion(assetCommitmentKey,
 
 	return deriveCommitmentKeys(
 		tapCommitment, p.InternalKey,
-		p.CommitmentProof.TapSiblingPreimage, true, true,
+		p.CommitmentProof.TapSiblingPreimage, false, true,
```
Roasbeef, line 432: Nice catch!

Roasbeef (changes requested):

asset/asset.go
```diff
@@ -1366,6 +1366,13 @@ func (a *Asset) HasSplitCommitmentWitness() bool {
 	return IsSplitCommitWitness(a.PrevWitnesses[0])
 }
 
+// IsTransferRoot returns true if this asset represents a root transfer. A root
+// transfer is an asset that is neither a genesis asset nor contains split
+// commitment witness data.
+func (a *Asset) IsTransferRoot() bool {
```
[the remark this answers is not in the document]
gijswijs: Correct. However, there is the todo of adding the v2 proofs to the `SplitRootProof`. You commented on that yourself as well: https://github.com/lightninglabs/taproot-assets/blob/af9066c4aeb34207f7be4ca3de28da92c4f42836/proof/append.go#L238-L239

proof/encoding.go
```diff
@@ -346,6 +347,87 @@ func CommitmentProofDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error
...
+				return err
+			}
+			proofBuf.Reset()
+		}
+		return nil
+	}
+	return tlv.NewTypeForEncodingErr(val, "[]CommitmentProof")
+}
+
+func CommitmentProofsDecoder(r io.Reader, val any, buf *[8]byte,
```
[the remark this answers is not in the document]
gijswijs: I'm not sure if JHB considered re-use of the encode/decode for `[]TaprootProof`. But re v2-proof there would never be a situation where you would want to prove that we didn't commit to _any_ altleaves: _If_ there is a v2-proof then by definition we commit to the assets being spend.

tl;dr I think we are saying the same thing.

proof/taproot.go
```diff
@@ -38,6 +38,13 @@ type CommitmentProof struct {
 	// at the tapscript root of the expected output.
 	TapSiblingPreimage *commitment.TapscriptPreimage
 
+	// STXOProofs is a list of proofs that either prove the spend of the
+	// inputs as referenced in the asset's previous witnesses' prevIDs when
+	// this is in an inclusion proof, or the non-spend of an input if this
+	// is in an exclusion proof. Only the root assets in a transfer (so no
+	// minted or split assets) will have STXO inclusion or exclusion proofs.
+	STXOProofs map[asset.SerializedKey]commitment.Proof
```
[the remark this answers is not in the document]
gijswijs: Yes, we derive the script key from the burn key that is in its turn derived from the `prevID`. Which is considered to be unique, otherwise you would be able to spend the same asset twice.

We ensure that _all_ `prevID`s are mapped (perfect bijection) during proof verification here: https://github.com/lightninglabs/taproot-assets/blob/1a3011c7ae0b6323e40818c4278f7ce75dbdf433/proof/verifier.go#L181-L196

proof/append.go
gijswijs: :100: I think that was already the case when you reviewed this, as it says this code was already outdated even before I made revisions... weird.
Anyway, yes. :+1:

proof/mint.go
gijswijs: This func only checks for _exclusion_ proofs and does so by cycling through the proofs in the `ExclusionProofs` field of the `BaseProofParams`. We decided to store stxo proofs in the `commitmentProof` struct as a nice solution for not having all sorts of `UnkownOddType`s, which was the initial approach JHB and I took. That does mean that the context informs whether we are working with an in- or exclusion proof.

We do rely on `prevID`s being unique, but I fail to see how that affects the code at the point where you placed this comment. See my other comment on `prevID`s here: https://github.com/lightninglabs/taproot-assets/pull/1453#discussion_r2024950538

itest/test_list_on_test.go
gijswijs: Cherry-picking gone wrong.

itest/psbt_test.go
gijswijs: yup! Cherry-picking went off rails.

proof/verifier.go
gijswijs: No, I don't think that's a particular good fit right here. Two reasons:
- `verifySTXOVersions` has two nested loops, so that would mean two nested `lfn.All`, which is kinda ugly.
- I want to log the exact proof that has the version that is different from the version of the first proof.

It's probably all possible with lfn.All, but it doesn't improve readability imo.

proof/verifier.go
gijswijs: :tada:

proof/verifier.go
gijswijs: /me shakes fist at Aider

proof/verifier.go
gijswijs: No, this is intended for a different case where we encounter a BIP-0086 wallet output (likely a change output) that does not commit to any script or Taproot Asset root. There is no special stxo proof handling needed for those outputs, so we default to `handleBasicExclusionProof`.

I've changed the code to better reflect the intent, and I've added a explanatory comment.

proof/append.go
guggero: This landed in the last commit, should probably be moved to this one.

proof/append.go
```diff
@@ -160,6 +160,46 @@ func CreateTransitionProof(prevOut wire.OutPoint,
...
+			// witness.
+			_, stxoProof, err := params.TaprootAssetRoot.Proof(
+				asset.EmptyGenesisID,
+				spentAsset.AssetCommitmentKey(),
+			)
+			if err != nil {
+				return nil, err
+			}
+
+			key := spentAsset.ScriptKey.PubKey.SerializeCompressed()
```
guggero, line 194: If the code is moved from the last commit here, then this can just be: `keySerialized := asset.ToSerialized(scriptKey.PubKey)`

proof/mint.go
```diff
@@ -158,6 +158,23 @@ func (p *BaseProofParams) HaveInclusionProof(anchorOutputIndex uint32) bool {
 	return p.OutputIndex == int(anchorOutputIndex)
 }
 
+// HaveSTXOInclusionProof returns true if the inclusion proof is for the given
```
guggero, line 161: nit: Godoc name and description doesn't match implementation.

itest/psbt_test.go
```diff
@@ -2454,6 +2455,197 @@ func testPsbtTrustlessSwap(t *harnessTest) {
...
+
+	mintedAsset := rpcAssets[0]
+	genInfo := rpcAssets[0].AssetGenesis
+	var assetId asset.ID
+	copy(assetId[:], genInfo.AssetId)
+
+	// Now that we have the asset created, we'll make a new node that'll
+	// serve as the node which'll receive the assets.
+	bobLnd := t.lndHarness.NewNodeWithCoins("Bob", nil)
+	bob := setupTapdHarness(
```
guggero, line 2495: nit: can fit on a single line.

itest/psbt_test.go
```diff
@@ -2454,6 +2455,197 @@ func testPsbtTrustlessSwap(t *harnessTest) {
...
+	)
+	defer func() {
+		require.NoError(t.t, bob.stop(!*noDelete))
+	}()
+
+	alice := t.tapd
+
+	// We need to derive two keys, one for the new script key and
+	// one for the internal key.
+	bobScriptKey, bobAnchorIntKeyDesc := DeriveKeys(
```
guggero, line 2506: nit: can also fit on one line.

proof/verifier.go
```diff
@@ -180,71 +226,276 @@ func (p *Proof) verifySplitRootProof() error {
...
+			commitVersions[outIdx], commitment.Version,
+		)
+	}
+
+	return commitment, nil
+}
+
+// makeSTXOProof creates a new proof for an STXO reusing the base proof while
+// replacing commitmentProof bu the stxoProof.
+func (p *Proof) makeSTXOProof(baseProof TaprootProof,
```
guggero, line 427: This can be a pure function, looks like we don't use `p` anywhere. Then we can export it and also use it in the itest in the previous commit. Though it feels a bit weird that we have to fake a whole `TaprootProof` just to verify the STXO proof. But let's not refactor that right now, just thinking aloud.

proof/verifier.go
```diff
@@ -180,71 +226,276 @@ func (p *Proof) verifySplitRootProof() error {
...
+	return nil
+}
+
+// verifySTXOProofSet verifies a set of STXO proofs.
+func (p *Proof) verifySTXOProofSet(baseProof TaprootProof,
+	assetMap map[asset.SerializedKey]*asset.Asset,
+	p2trOutputs map[uint32]fn.Set[asset.SerializedKey],
+	commitVersions map[uint32][]commitment.TapCommitmentVersion,
+	inclusion bool,
+) (*commitment.TapCommitment, error) {
```
guggero, line 394: nit: can fit on previous line to avoid weird formatting.

proof/verifier.go
```diff
@@ -180,71 +226,276 @@ func (p *Proof) verifySplitRootProof() error {
...
-		exclusionProof := exclusionProof
-		derivedCommitment, err := verifyTaprootProof(
-			&p.AnchorTx, &exclusionProof, &p.Asset, false,
+	return p.verifyV1ExclusionProofs(p2trOutputs)
+}
+
+// verifyV1ExclusionProofs verifies all version 1 exclusion proofs.
+func (p *Proof) verifyV1ExclusionProofs(
+	p2trOutputs map[uint32]fn.Set[asset.SerializedKey],
+) (*commitment.TapCommitmentVersion, error) {
```
guggero, line 268: Create a named type for `map[uint32]fn.Set[asset.SerializedKey],` (e.g. `OutputKeys`) to avoid weird formatting.

proof/verifier.go
```diff
@@ -180,71 +226,276 @@ func (p *Proof) verifySplitRootProof() error {
...
 			continue
 		}
-		if txscript.IsPayToTaproot(txOut.PkScript) {
-			p2trOutputs[uint32(i)] = struct{}{}
+		if !txscript.IsPayToTaproot(txOut.PkScript) {
+			continue
 		}
+
+		outIdx := uint32(i)
+		p2trOutputs[outIdx] = make(fn.Set[asset.SerializedKey])
```
guggero, line 250: Hmm, we don't actually add any keys to this set here?

proof/verifier.go
guggero: Or, actually, since as the **input** to the v1 and v2 function we're only really interested in the set of output indexes we want to check exclusion for, wouldn't it be enough to just use `fn.Set[uint32]`?

proof/verifier.go
```diff
@@ -180,71 +226,276 @@ func (p *Proof) verifySplitRootProof() error {
...
+			commitVersions[outputIdx],
+			derivedCommitment.Version,
+		)
+	}
+
+	return nil
+}
+
+// verifySTXOProofSet verifies a set of STXO proofs.
+func (p *Proof) verifySTXOProofSet(baseProof TaprootProof,
```
guggero, line 389: Could probably pass in the anchor TX here and make this a pure function as well.

proof/verifier.go
```diff
@@ -180,71 +226,276 @@ func (p *Proof) verifySplitRootProof() error {
...
+
+	if len(p2trOutputs) > 0 {
+		return ErrMissingExclusionProofs
+	}
+
+	return nil
+}
+
+// verifySTXOVersions verifies all STXO versions match.
+func (p *Proof) verifySTXOVersions(firstVersion commitment.TapCommitmentVersion,
```
guggero, line 477: nit: doesn't need to be a method.

proof/verifier.go
```diff
@@ -165,6 +165,52 @@ func verifyTaprootProof(anchor *wire.MsgTx, proof *TaprootProof,
 
 // verifyInclusionProof verifies the InclusionProof is valid.
 func (p *Proof) verifyInclusionProof() (*commitment.TapCommitment, error) {
+
+	// Determine if we're dealing with v1 or v2 proofs.
+	hasV2Proofs := p.InclusionProof.CommitmentProof != nil &&
+		len(p.InclusionProof.CommitmentProof.STXOProofs) > 0
+
+	if hasV2Proofs {
```
guggero, line 173: Can return early if this is false, then continue un-indented.

proof/verifier.go
```diff
@@ -767,6 +1018,7 @@ func (p *Proof) VerifyProofs() (*commitment.TapCommitment, error) {
 
 	// The inclusion proof must have a similar version to all exclusion
 	// proofs.
+
```
guggero, line 1021: nit: unnecessary newline.

proof/append.go
```diff
@@ -171,10 +171,14 @@ func CreateTransitionProof(prevOut wire.OutPoint,
 					"prevID", wit)
 			}
 
-			spentAsset := &asset.Asset{
-				ScriptKey: asset.NewScriptKey(
-					asset.DeriveBurnKey(*wit.PrevID),
-				),
+			prevIdKey := asset.DeriveBurnKey(*wit.PrevID)
```
guggero, line 174: nit: move this to earlier commit.

proof/append_test.go
```diff
@@ -135,6 +135,17 @@ func runAppendTransitionTest(t *testing.T, assetType asset.Type, amt uint64,
 
 	// Add some alt leaves to the commitment anchoring the asset transfer.
 	altLeaves := asset.ToAltLeaves(asset.RandAltLeaves(t, true))
+
+	// Commit to the stxo of the previous asset. Otherwise the inclusion
+	// proofs will fail.
+	prevIdKey := asset.DeriveBurnKey(*newAsset.PrevWitnesses[0].PrevID)
+	scriptKey := asset.NewScriptKey(prevIdKey)
+	stxoAsset, err := asset.NewAltLeaf(
+		scriptKey, asset.ScriptV0, nil,
```
guggero, line 144: Since we're only ever passing `nil` (and we use the previous witness in the script/burn key), I think we can just remove the witness parameter completely.

tapsend/proof_test.go
```diff
@@ -222,27 +222,38 @@ func addOutputCommitment(t *testing.T, anchorTx *AnchorTransaction,
 		}
 	}
 
-	for idx, assets := range assetsByOutput {
-		for idx := range assets {
-			if !assets[idx].HasSplitCommitmentWitness() {
+	stxoAssetsByOutput := make(map[uint32][]asset.AltLeaf[asset.Asset])
```
guggero, line 225: Should we also add some negative tests where we expect a failure?

guggero (commented):
Nice work, this looks pretty good :tada: 
Have a couple of suggestions, after which we can probably take this out of draft.

tapsend/proof.go
gijswijs: I think I can do one better.

I've added code some 20 lines below that confirms that we are creating the stxo proofs for the asset for which there already is an (old style) exclusion proof. So this confirms that this is actually an exclusion proof, but also confirms that it's a valid exclusion proof for this exact asset.

proof/verifier.go
gijswijs: Don't get me started on the `TaprootProof` construction. I don't like it _at all_.

guggero: Yeah, me neither... Probably something we can refactor at some point.

proof/verifier.go
gijswijs: With the refactor of `p2trOutputs` we now use a simple `fn.Set[uint32]` here and only set the keys once we are verifying and tracking stxo proofs.

proof/encoding.go
```diff
@@ -346,6 +347,98 @@ func CommitmentProofDecoder(r io.Reader, val any, buf *[8]byte, l uint64) error
...
+			if err != nil {
+				return err
+			}
+
+			proofBuf.Reset()
+		}
+		return nil
+	}
+
+	return tlv.NewTypeForEncodingErr(val, "[]CommitmentProof")
```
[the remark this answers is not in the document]
gijswijs: Laser-eyed-Oli :eye:

proof/append_test.go
gijswijs: :+1:

guggero:
Extracted some commits into #1465 to ease review on this one. Pushed up a rebase.

proof/verifier.go
```diff
@@ -184,67 +188,273 @@ func (p *Proof) verifySplitRootProof() error {
 func (p *Proof) verifyExclusionProofs() (*commitment.TapCommitmentVersion,
 	error) {
 
-	// Gather all P2TR outputs in the on-chain transaction.
-	p2trOutputs := make(map[uint32]struct{})
+	// There is nothing to check so return early.
+	if p.ExclusionProofs == nil {
```
guggero, line 236: I think this is incorrect. Because that would mean I can just create a proof with no exclusion proof and I wouldn't have to prove anything.
The code below proves that we actually have an exclusion proof **per p2tr output**. So the exit early would only be if `len(p2trOutputs) == 0`.

proof/verifier.go
```diff
@@ -184,67 +188,273 @@ func (p *Proof) verifySplitRootProof() error {
...
 
-	// If there were no commitments in any exclusion proofs, then there is
-	// no version to return.
+	// If no commitments in proofs, no version to return.
+	if len(commitVersions) == 0 {
+		return nil, nil
+	}
+
+	// All proofs must have similar versions.
+	firstVersion := maps.Values(commitVersions)[0][0]
```
guggero, line 296: Move this extraction of the first version into the `verifySTXOVersions`, since it's done the same way twice?

proof/verifier.go
```diff
@@ -184,67 +188,273 @@ func (p *Proof) verifySplitRootProof() error {
...
-	for _, exclusionProof := range p.ExclusionProofs {
-		exclusionProof := exclusionProof
-		derivedCommitment, err := verifyTaprootProof(
-			&p.AnchorTx, &exclusionProof, &p.Asset, false,
+	// Early pass to determine if we're dealing with v0 or v1 proofs.
+	hasV1Proofs := len(p.ExclusionProofs) > 0 &&
+		p.ExclusionProofs[0].CommitmentProof != nil &&
+		len(p.ExclusionProofs[0].CommitmentProof.STXOProofs) > 0
+
+	if hasV1Proofs {
```
guggero, line 261: This whole logic is pretty hard to follow. And because of that, it's hard to reason about not removing any existing verification.

Can we perhaps change the flow that we always verify the V0 exclusion proofs.
And then in `verifyV1ExclusionProofs` we would _only_ verify the STXO proofs?

Here's a quick diff of what I mean:
```diff
diff --git a/proof/verifier.go b/proof/verifier.go
index 166e81a0..f25e0e0e 100644
--- a/proof/verifier.go
+++ b/proof/verifier.go
@@ -83,6 +83,8 @@ type BaseVerifier struct {
 // transaction.
 type P2TROutputsSTXOs = map[uint32]fn.Set[asset.SerializedKey]
 
+type CommittedVersions = map[uint32][]commitment.TapCommitmentVersion
+
 // Verify takes the passed serialized proof file, and returns a nil
 // error if the proof file is valid. A valid file should return an
 // AssetSnapshot of the final state transition of the file.
@@ -232,11 +234,6 @@ func (p *Proof) verifySplitRootProof() error {
 func (p *Proof) verifyExclusionProofs() (*commitment.TapCommitmentVersion,
        error) {
 
-       // There is nothing to check so return early.
-       if p.ExclusionProofs == nil {
-               return nil, nil
-       }
-
        // Gather all P2TR outputs in the on-chain transaction. The STXO proofs
        // are tracked using a set of serialized keys. We ignore that set when
        // verifying V0 exclusion proofs.
@@ -252,6 +249,14 @@ func (p *Proof) verifyExclusionProofs() (*commitment.TapCommitmentVersion,
                p2trOutputs.Add(uint32(i))
        }
 
+       commitVersions := make(CommittedVersions)
+
+       p2trOutputCopy := fn.NewSet(p2trOutputs.ToSlice()...)
+       err := p.verifyV0ExclusionProofs(p2trOutputCopy, commitVersions)
+       if err != nil {
+               return nil, err
+       }
+
        // Early pass to determine if we're dealing with v0 or v1 proofs.
        hasV1Proofs := p.IsVersionV1() &&
                len(p.ExclusionProofs) > 0 &&
@@ -259,54 +264,66 @@ func (p *Proof) verifyExclusionProofs() (*commitment.TapCommitmentVersion,
                len(p.ExclusionProofs[0].CommitmentProof.STXOProofs) > 0
 
        if hasV1Proofs {
-               return p.verifyV1ExclusionProofs(p2trOutputs)
+               p2trOutputCopy = fn.NewSet(p2trOutputs.ToSlice()...)
+               err := p.verifyV1ExclusionProofs(p2trOutputCopy, commitVersions)
+               if err != nil {
+                       return nil, err
+               }
+       }
+
+       // If no commitments in proofs, no version to return.
+       if len(commitVersions) == 0 {
+               return nil, nil
        }
 
-       return p.verifyV0ExclusionProofs(p2trOutputs)
+       // All proofs must have similar versions.
+       firstVersion := maps.Values(commitVersions)[0][0]
+       return verifySTXOVersions(firstVersion, commitVersions)
 }
 
 // verifyV0ExclusionProofs verifies all version 1 exclusion proofs.
-func (p *Proof) verifyV0ExclusionProofs(p2trOutputs fn.Set[uint32]) (
-       *commitment.TapCommitmentVersion, error) {
-
-       commitVersions := make(map[uint32][]commitment.TapCommitmentVersion)
+func (p *Proof) verifyV0ExclusionProofs(p2trOutputs fn.Set[uint32],
+       commitVersions CommittedVersions) error {
 
        // Verify all of the encoded v0 exclusion proofs.
        for idx := range p.ExclusionProofs {
                exclusionProof := p.ExclusionProofs[idx]
-               err := p.handleBasicExclusionProof(
-                       &exclusionProof, p2trOutputs, commitVersions,
+
+               derivedCommitment, err := verifyTaprootProof(
+                       &p.AnchorTx, &exclusionProof, &p.Asset, false,
                )
                if err != nil {
-                       return nil, err
+                       return fmt.Errorf("error verifying STXO proof: %w", err)
+               }
+
+               outputIdx := exclusionProof.OutputIndex
+               delete(p2trOutputs, exclusionProof.OutputIndex)
+
+               // Store commitment version if Taproot Asset commitment present.
+               if derivedCommitment != nil {
+                       commitVersions[outputIdx] = append(
+                               commitVersions[outputIdx],
+                               derivedCommitment.Version,
+                       )
                }
        }
 
        // If any outputs are missing a proof, fail.
        if len(p2trOutputs) > 0 {
-               return nil, ErrMissingExclusionProofs
-       }
-
-       // If no commitments in proofs, no version to return.
-       if len(commitVersions) == 0 {
-               return nil, nil
+               return ErrMissingExclusionProofs
        }
 
-       // All proofs must have similar versions.
-       firstVersion := maps.Values(commitVersions)[0][0]
-       return verifySTXOVersions(firstVersion, commitVersions)
+       return nil
 }
 
 // verifyV1ExclusionProofs verifies all version 2 exclusion proofs.
-func (p *Proof) verifyV1ExclusionProofs(
-       p2trOutputs fn.Set[uint32]) (*commitment.TapCommitmentVersion, error) {
-
-       commitVersions := make(map[uint32][]commitment.TapCommitmentVersion)
+func (p *Proof) verifyV1ExclusionProofs(p2trOutputs fn.Set[uint32],
+       commitVersions CommittedVersions) error {
 
        // Collect the STXOs from the new asset.
        stxoAssets, err := asset.CollectSTXO(&p.Asset)
        if err != nil {
-               return nil, fmt.Errorf("error collecting STXO assets: %w", err)
+               return fmt.Errorf("error collecting STXO assets: %w", err)
        }
 
        // Create a P2TROutputsSTXOs from p2trOutputs and map STXOs by
@@ -327,68 +344,22 @@ func (p *Proof) verifyV1ExclusionProofs(
        }
 
        for idx := range p.ExclusionProofs {
-               baseProof := p.ExclusionProofs[idx]
-               if baseProof.TapscriptProof != nil &&
-                       baseProof.TapscriptProof.Bip86 {
-
-                       // This is a BIP-0086 wallet output (likely a change
-                       // output) that does not commit to any script or Taproot
-                       // Asset root. There is no special stxo proof handling
-                       // needed, so we default to `handleBasicExclusionProof`.
-                       err := p.handleBasicExclusionProof(
-                               &baseProof, p2trOutputs, commitVersions,
-                       )
-                       if err != nil {
-                               return nil, err
-                       }
+               exclusionProof := p.ExclusionProofs[idx]
+               if exclusionProof.TapscriptProof != nil &&
+                       exclusionProof.TapscriptProof.Bip86 {
+
                        continue
                }
 
                _, err := verifySTXOProofSet(
-                       &p.AnchorTx, baseProof, assetMap, p2trOutputsSTXOs,
+                       &p.AnchorTx, exclusionProof, assetMap, p2trOutputsSTXOs,
                        commitVersions, false,
                )
                if err != nil {
-                       return nil, err
+                       return err
                }
        }
 
-       err = p.verifyRemainingOutputs(p2trOutputsSTXOs)
-       if err != nil {
-               return nil, err
-       }
-
-       if len(commitVersions) == 0 {
-               return nil, nil
-       }
-
-       // All proofs must have similar versions.
-       firstVersion := maps.Values(commitVersions)[0][0]
-       return verifySTXOVersions(firstVersion, commitVersions)
-}
-
-// handleBasicExclusionProof handles a basic (non-STXO) exclusion proof.
-func (p *Proof) handleBasicExclusionProof(baseProof *TaprootProof,
-       p2trOutputs fn.Set[uint32],
-       commitVersions map[uint32][]commitment.TapCommitmentVersion) error {
-
-       derivedCommitment, err := verifyTaprootProof(
-               &p.AnchorTx, baseProof, &p.Asset, false,
-       )
-       if err != nil {
-               return fmt.Errorf("error verifying STXO proof: %w", err)
-       }
-
-       outputIdx := baseProof.OutputIndex
-       delete(p2trOutputs, baseProof.OutputIndex)
-
-       // Store commitment version if Taproot Asset commitment present.
-       if derivedCommitment != nil {
-               commitVersions[outputIdx] = append(
-                       commitVersions[outputIdx], derivedCommitment.Version,
-               )
-       }
-
        return nil
 }
 
@@ -450,33 +421,6 @@ func MakeSTXOProof(baseProof TaprootProof,
        }
 }
 
-// verifyRemainingOutputs verifies any remaining outputs are BIP-86 outputs.
-func (p *Proof) verifyRemainingOutputs(p2trOutputs P2TROutputsSTXOs) error {
-       if len(p2trOutputs) == 0 {
-               return nil
-       }
-
-       // Check remaining outputs are BIP-86 outputs.
-       for outIdx := range p2trOutputs {
-               for _, proof := range p.ExclusionProofs {
-                       if proof.OutputIndex != outIdx {
-                               continue
-                       }
-                       if proof.TapscriptProof != nil &&
-                               proof.TapscriptProof.Bip86 {
-
-                               delete(p2trOutputs, outIdx)
-                       }
-               }
-       }
-
-       if len(p2trOutputs) > 0 {
-               return ErrMissingExclusionProofs
-       }
-
-       return nil
-}
-
 // verifySTXOVersions verifies all STXO versions match.
 func verifySTXOVersions(firstVersion commitment.TapCommitmentVersion,
        versions map[uint32][]commitment.TapCommitmentVersion) (
diff --git a/tapsend/proof.go b/tapsend/proof.go
index 018857fd..4773ef9a 100644
--- a/tapsend/proof.go
+++ b/tapsend/proof.go
@@ -366,6 +366,33 @@ func addSTXOExclusionProofs(outputs []*tappsbt.VOutput,
                        continue
                }
 
+               // Find the exclusion proofs for this output.
+               var eProof *proof.TaprootProof
+               for idx := range params.ExclusionProofs {
+                       e := params.ExclusionProofs[idx]
+                       if e.OutputIndex == outIndex {
+                               eProof = &params.ExclusionProofs[idx]
+                               break
+                       }
+               }
+               if eProof == nil {
+                       return fmt.Errorf("no exclusion proof for "+
+                               "output %d", outIndex)
+               }
+
+               // There aren't any assets in that output, we can skip
+               // creating exclusion proofs for it.
+               if eProof.CommitmentProof == nil {
+                       continue
+               }
+
+               commitmentProof := eProof.CommitmentProof
+               if commitmentProof.STXOProofs == nil {
+                       commitmentProof.STXOProofs = make(
+                               map[asset.SerializedKey]commitment.Proof,
+                       )
+               }
+
                tapTree := outputCommitments[outIndex]
 
                for idx := range stxoAssets {
@@ -381,28 +408,6 @@ func addSTXOExclusionProofs(outputs []*tappsbt.VOutput,
                                return err
                        }
 
-                       // Find the exclusion proofs for this output.
-                       var eProof *proof.TaprootProof
-                       for idx := range params.ExclusionProofs {
-                               e := params.ExclusionProofs[idx]
-                               if e.OutputIndex == outIndex {
-                                       eProof = &params.ExclusionProofs[idx]
-                                       break
-                               }
-                       }
-                       if eProof == nil {
-                               return fmt.Errorf("no exclusion proof for "+
-                                       "output %d", outIndex)
-                       }
-
-                       // There aren't any assets in that output, we can skip
-                       // creating exclusion proofs for it.
-                       if eProof.CommitmentProof == nil {
-                               continue
-                       }
-
-                       commitmentProof := eProof.CommitmentProof
-
                        // Confirm that we are creating the stxo proofs for the
                        // asset that is being created. We do this by confirming
                        // that the exclusion proof for the newly created asset
@@ -417,13 +422,6 @@ func addSTXOExclusionProofs(outputs []*tappsbt.VOutput,
                                        "stxo proofs: %w", err)
                        }
 
-                       //nolint:lll
-                       if commitmentProof.STXOProofs == nil {
-                               commitmentProof.STXOProofs = make(
-                                       map[asset.SerializedKey]commitment.Proof,
-                               )
-                       }
-
                        commitmentProof.STXOProofs[identifier] = *exclusionProof
                }
        }
```

Or maybe it's easier as a wip commit: https://github.com/guggero/taproot-assets/commit/a2a1633b0161818b1143572eeff7961f7c79e379

proof/verifier.go
```diff
@@ -169,9 +169,52 @@ func verifyTaprootProof(anchor *wire.MsgTx, proof *TaprootProof,
...
+
+	commitment, err := verifySTXOProofSet(
+		&p.AnchorTx, baseProof, assetMap, p2trOutputs, commitVersions,
+		true,
 	)
+	if err != nil {
+		return nil, err
+	}
+
+	err = p.verifyRemainingOutputs(p2trOutputs)
```
guggero, line 213: I don't think we need to call into this. We only added a single output index to the map and we expect that to be empty now, after the STXO verification. So we just need to check for map emptiness.

tapsend/proof_test.go
guggero: Yeah, that would be good.
Can assist with that test, have a couple of ideas.

proof/verifier.go
```diff
@@ -252,7 +253,8 @@ func (p *Proof) verifyExclusionProofs() (*commitment.TapCommitmentVersion,
 	}
 
 	// Early pass to determine if we're dealing with v0 or v1 proofs.
-	hasV1Proofs := len(p.ExclusionProofs) > 0 &&
+	hasV1Proofs := p.IsVersionV1() &&
```
guggero, line 256: If this is a V1 proof and there are no STXO proofs, should that result in an error?
Same for the inclusion proofs above.

guggero (commented):
Did another pass. Definitely looks better, but I think we still need an iteration on the core verification logic, to make it easier to understand.

guggero:
Sorry, I don't think I made things easier for you as an author by splitting the PR into two. And now I noticed that the commit https://github.com/lightninglabs/taproot-assets/commit/c9b1d54430cdd7c4cdf3e1e65e7a33babc428afd does need to go into this one instead...
Let me know if I should fix my own mess by rebasing this one.

proof/append.go
gijswijs: Oli has the same train of thought as do I. But I want this to be confirmed by @Roasbeef

tapsend/proof.go
```diff
@@ -326,9 +339,101 @@ func proofParams(finalTx *wire.MsgTx, vPkt *tappsbt.VPacket,
...
+				continue
+			}
+
+			commitmentProof := eProof.CommitmentProof
+
+			// Confirm that we are creating the stxo proofs for the
+			// asset that is being created. We do this by confirming
+			// that the exclusion proof for the newly created asset
+			// is already present.
+			_, err = eProof.DeriveByAssetExclusion(
```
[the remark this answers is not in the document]
gijswijs: This was in reaction to an [earlier comment by Roas](https://github.com/lightninglabs/taproot-assets/pull/1453#discussion_r2017817356). We can (currently) only create `STXOExclusionProofs` after the old `v0` exclusion proofs are created. We want to be sure we are creating the stxo proofs for the same asset that we created the old exlcusion proofs for. So we verify those `v0` proofs here with the `newAsset` so that we know for sure that we are creating the stxo exclusion proofs for the _same_ asset. It's a sanity check.

proof/verifier.go
gijswijs: Yes, but there need not necessarily be any exclusion proofs, right? We can have a single output anchor tx that wouldn't necessitate any exclusion proofs. So we can't err out if we don't find them.

Roasbeef:
> I think we can remove the TODO, since only the root assets need to have an inclusion proof, since the split outputs themselves should be covered by the split commitment and exclusion proof.
But wanted to confirm with both of you first...

Correct. We if an asset is a split (has split commitment proof), then we don't want to add the input to the STXO at all. Otherwise, we'll get a different tree depending on the order that we inserted them (we derive the alt leaf using the script key of an asset). 

Related to that that if a prev ID is all zeroes, then we also don't want to add them. Otherwise many issued assets in an output would introduce a similar issue as above.

Roasbeef:
Is this PR ready to leave draft? We should start to run CI against it.

gijswijs:
> Related to that that if a prev ID is all zeroes, then we also don't want to add them. Otherwise many issued assets in an output would introduce a similar issue as above.

I think that is covered by:

https://github.com/lightninglabs/taproot-assets/blob/79e199feca0512b9cd26977a0ab4ffd673ca9385/asset/asset.go#L1472-L1477

We use this func to determine if we need to collect  and commit to STXOs here:

https://github.com/lightninglabs/taproot-assets/blob/79e199feca0512b9cd26977a0ab4ffd673ca9385/asset/asset.go#L2175-L2180

proof/append.go
gijswijs: Resolved. See: https://github.com/lightninglabs/taproot-assets/pull/1453#issuecomment-2857973135

proof/append.go
```diff
@@ -160,6 +160,49 @@ func CreateTransitionProof(prevOut wire.OutPoint,
...
+			map[asset.SerializedKey]commitment.Proof,
+			len(proof.Asset.PrevWitnesses),
+		)
+		for _, wit := range proof.Asset.PrevWitnesses {
+			if wit.PrevID == nil {
+				return nil, fmt.Errorf("witness %v has no "+
+					"prevID", wit)
+			}
+
+			prevIdKey := asset.DeriveBurnKey(*wit.PrevID)
```
Roasbeef, line 174: Taking a prevID, then creating a burn key from that, a script key from that, and finally the actual alt leaf is done in many areas in the diff. It may be worth considering the addition of a new helper function to take care of this. It'll cut down on 5-ish lines per instances across the diff.

proof/append_test.go
```diff
@@ -293,8 +303,19 @@ func runAppendTransitionTest(t *testing.T, assetType asset.Type, amt uint64,
 		nil, split1Commitment,
 	)
 	require.NoError(t, err)
+
+	// Commit to the stxo of the previous asset. Otherwise, the inclusion
+	// proofs will fail. With splits this is only needed for the root asset.
+	prevIdKey1 := asset.DeriveBurnKey(*split1Asset.PrevWitnesses[0].PrevID)
```
Roasbeef, line 309: Same comment here re the alt leaf creation operations. Won't comment elsewhere as to not litter.

proof/verifier.go
```diff
@@ -165,9 +176,71 @@ func verifyTaprootProof(anchor *wire.MsgTx, proof *TaprootProof,
...
+	// are no STXO proofs present (because they're not needed for this type
+	// of asset).
+	if p.IsVersionV0() || !hasStxoProofs {
+		return v0Commitment, nil
+	}
+
+	// For an inclusion proof, there is only one output index that we need
+	// to check.
+	outIdx := p.InclusionProof.OutputIndex
+	p2trOutputs := map[uint32]fn.Set[asset.SerializedKey]{
```
Roasbeef, line 209: You can use the `P2TROutputsSTXOs` type alias defined above to shorten this slightly.

proof/verifier.go
```diff
@@ -184,67 +257,232 @@ func (p *Proof) verifySplitRootProof() error {
...
+		p2trOutputs.Add(uint32(i))
+	}
+
+	// There is nothing to check so return early.
+	if len(p2trOutputs) == 0 {
+		return nil, nil
+	}
+
+	commitVersions, err := p.verifyV0ExclusionProofs(
+		maps.Clone(p2trOutputs),
```
Roasbeef, line 281: Nice usage of `maps.Clone` here 👍

proof/verifier.go
```diff
@@ -184,67 +257,232 @@ func (p *Proof) verifySplitRootProof() error {
...
+			TapSiblingPreimage: baseProof.CommitmentProof.
+				TapSiblingPreimage,
+		},
+		TapscriptProof:  baseProof.TapscriptProof,
+		UnknownOddTypes: baseProof.UnknownOddTypes,
+	}
+}
+
+// verifySTXOVersions verifies all STXO versions match.
+func verifySTXOVersions(versions map[uint32][]commitment.TapCommitmentVersion) (
```
Roasbeef, line 463: IIUC, this isn't STXO specific, it's generic to the asset commitment: we want to make sure all the asset commitments have the same version. Looks like this was refactored out of the old logic, so I think we can just rename it to: `assertUniformCommitVersions` or something like that?

proof/verifier_test.go
```diff
@@ -0,0 +1,613 @@
...
+
+// outputScript returns the output script for the given taproot commitment
+// and internal key.
+func outputScript(t *testing.T, c *commitment.TapCommitment,
+	i *btcec.PublicKey) []byte {
+
+	tapscriptRoot := c.TapscriptRoot(nil)
+	expectedKey, err := schnorr.ParsePubKey(schnorr.SerializePubKey(
+		txscript.ComputeTaprootOutputKey(i, tapscriptRoot[:]),
+	))
```
Roasbeef, line 61: The first error here isn't checked.

proof/verifier_test.go
```diff
@@ -0,0 +1,613 @@
...
+	pkScript, err := txscript.PayToTaprootScript(expectedKey)
+	require.NoError(t, err)
+
+	return pkScript
+}
+
+// makeSpentAsset creates a random asset that spends the given witness.
+func makeSpentAsset(t *testing.T, witness asset.Witness) *asset.Asset {
```
Roasbeef, line 69: Cool, this is the exact function I think we can lift to live outside the test package, for use elsewhere.

proof/verifier_test.go
```diff
@@ -0,0 +1,613 @@
...
+	}
+
+	randAsset := asset.RandAsset(t, asset.Normal)
+
+	// Make sure it's a transfer asset and not a genesis.
+	randAsset.PrevWitnesses[0].PrevID.ID = asset.RandID(t)
+
+	for _, w := range cfg.additionalPrevWitnesses {
+		randAsset.PrevWitnesses = append(randAsset.PrevWitnesses, w)
+	}
```
Roasbeef, line 94: Can do just do: 
```
randAsset.PrevWitnesses = append(randAsset.PrevWitnesses, cfg.additionalPrevWitnesses...)
```

proof/verifier_test.go
```diff
@@ -0,0 +1,613 @@
...
+
+// TestVerifyV0V1InclusionProofs tests that we properly verify the inclusion
+// proofs for both v0 and v1.
+func TestVerifyV1InclusionProof(t *testing.T) {
+	testCases := []struct {
+		name                string
+		makeProof           func(t *testing.T) *Proof
+		expectedErr         error
+		expectedErrContains string
+	}{{
```
Roasbeef, line 258: Style nit: can put the frist case on a new line, then each section after that adds a new test, so like: 
```go
t = []struct {
}{
  case
}, 
{
  case
}, 
}
  case
}, 
```

proof/verifier_test.go
```diff
@@ -0,0 +1,613 @@
...
+			// The second one is missing, so the proof should be
+			// invalid.
+			_, _ = addV1InclusionProof(
+				t, p, rootCommitment, withAddStxoProof(),
+				withUpdateAnchorTx(),
+			)
+
+			return p
+		},
+		expectedErr:         ErrInvalidCommitmentProof,
```
Roasbeef, line 385: Nice set of tests! I think for the sake of extneding the test vectors here in the future, we should make some new specific errors for these sub-cases. So in this case, it would be like `ErrStxoInputProofMissing` or something like that.

proof/verifier_test.go
```diff
@@ -0,0 +1,613 @@
...
+						// The output with the inclusion
+						// proof.
+						PkScript: dummyP2TR,
+					}},
+				},
+			}
+
+			return p
+		},
+	}, {
```
Roasbeef, line 489: Same here re one brace per line.

proof/verifier_test.go
```diff
@@ -0,0 +1,613 @@
...
+			// We now have a V1 proof, so we expect the proof to
+			// verify successfully.
+			p.Version = TransitionV1
+
+			return p
+		},
+	}, {
+		name: "multiple assets with an exclusion proof each",
+		makeProof: func(t *testing.T) *Proof {
+			// TODO(guggero): implement
```
Roasbeef, line 572: Do we plan on resolving this before merging?

tapsend/send_test.go
```diff
@@ -682,6 +682,8 @@ func checkOutputCommitments(t *testing.T, vPkt *tappsbt.VPacket,
 	default:
 		require.Fail(t, "unknown vPacket version")
 	}
+
+	// TODO(jhb): Check for correct STXOs
```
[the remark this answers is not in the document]
Roasbeef: TODO is still here.

Roasbeef (changes requested):
Did a clean review, focusing on the tests in particular. Feels like we're pretty close. 

The linter is failing, and also the unit tests panic as one of the test cases is incomplete: 
```
--- FAIL: TestVerifyV1ExclusionProof (0.12s)
    --- FAIL: TestVerifyV1ExclusionProof/multiple_assets_with_an_exclusion_proof_each (0.00s)
```

proof/verifier_test.go
gijswijs: I'm going to vote in favor of less indentation here, to prevent line wrapping hell.

Roasbeef (commented):
> Reviewed 25 of 25 files at r1, 28 of 28 files at r2, 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1453)* status: all files reviewed, 61 unresolved discussions (waiting on @gijswijs and @guggero)

guggero (commented):
I pushed a single fixup commit that adds more unit tests (and also uncovered a small bug that is fixed too). Feel free to check and apply that commit directly (not attribution required).

Other than that only nits are left in the code (which I didn't fix, I leave that to you).

The litd itest failure is related to the changes here though:
```
error funding channel: unable to bind PSBT: rpc error: code = Unknown desc = error verifying PSBT: funding output not found in PSBT
```

So I'll need to take a look into that. But hopefully it's something small (will take a closer look tomorrow).

proof/verifier.go
```diff
@@ -184,67 +257,242 @@ func (p *Proof) verifySplitRootProof() error {
...
+		},
+		TapscriptProof:  baseProof.TapscriptProof,
+		UnknownOddTypes: baseProof.UnknownOddTypes,
+	}
+}
+
+// assertVersionConsistency verifies all STXO versions match.
+func assertVersionConsistency(
+	versions map[uint32][]commitment.TapCommitmentVersion,
+) (*commitment.TapCommitmentVersion, error) {
```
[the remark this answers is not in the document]
gijswijs: I think the docs are ambiguous. They state:

>  Also, lines should not end with an open parenthesis if the function definition isn't finished yet.

But also

> RIGHT
> ```
> func longFunctionName(
> 	a, b, c) (d, error) {
> ```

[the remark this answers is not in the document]
gijswijs: I did take the approach of the `longFunctionName` example to resolve this comment tho.

Roasbeef:
> So I'll need to take a look into that. But hopefully it's something small (will take a closer look tomorrow).

I think what's missing is that we need to mind the alt leaves when we manually create the funding output our selves.

guggero (approved):
With the above comment and a (hopefully) green CI, I approve this PR. Great work, super happy to finally see this getting over the finish line.

Roasbeef:
> We do NOT produce any V1 proofs in this PR (outside of unit tests). Otherwise newly minted or transferred assets would not be recognized by older nodes ("unknown proof version 1")

What if we still set all the fields, but don't yet bump the version? The new TLVs would all be odd. We then flip the new version at a point in the future once the new fields are showing up more often in public proofs. 

We'd then still verify as normal. 

> Which means we'll need a specific upgrade plan for channels in the future.

Makes sense, we can add another version to the initial set of pre funding messages that we send to the other side.

> STXO commitments are only added to root outputs so on-chain sends to older nodes still work as expected. So we can test that we can correctly spend any inputs that have STXO commitments in them.

Is this actually a change in behavior from the earlier versions? IIUC, we never set it for splits anyway, as they spend a prevID of zero.

gijswijs (commented):
Also, for full clarification and confirmation:

> The new TLVs would all be odd.

This PR only introduces a single new TLV, which is indeed odd.

`CommitmentProofSTXOProofsType tlv.Type = 7`

Roasbeef (approved):
LGTM 🐄

> Reviewed 1 of 9 files at r3, 20 of 20 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1453#-:-OQjkrcS77NEMr2iuOIH:b-lilaky)* status: all files reviewed, 67 unresolved discussions (waiting on @gijswijs and @guggero)

___
*[`tapchannel/aux_closer.go` line 457 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/1453#-OQjjePyAUWOokedAuT6:-OQjjePyAUWOokedAuT7:b-8fx2u7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/76c070601d34785ff5fcf6e726a13272db895f85/tapchannel/aux_closer.go#L457)):*
> ```Go
> 	// the output indexes in the allocations.
> 	outCommitments, err := tapsend.CreateOutputCommitments(
> 		vPackets, tapsend.WithNoSTXOProofs(),
> ```

:+1:

PR #1456, opened 2025-03-27
tapdb+universe: implement new Universe tree for 1st party burns

Roasbeef:
In this commit, we add a new interface, the BurnTree. This interface
will be used by a future sub-system to maintain the on-chain asset
commitment for a given asset group.

The structure of this tree is very similar to the existing issuance
tree. The main difference is that we'll only insert burn proofs into
this tree.

We then implement a concrete implementation of this interface 
using the existing universe tree routines.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -5148,11 +5148,11 @@ func (r *rpcServer) DeleteAssetRoot(ctx context.Context,
 func marshalLeafKey(leafKey universe.LeafKey) *unirpc.AssetKey {
 	return &unirpc.AssetKey{
 		Outpoint: &unirpc.AssetKey_OpStr{
-			OpStr: leafKey.OutPoint.String(),
+			OpStr: leafKey.LeafOutPoint().String(),
 		},
 		ScriptKey: &unirpc.AssetKey_ScriptKeyBytes{
 			ScriptKeyBytes: schnorr.SerializePubKey(
-				leafKey.ScriptKey.PubKey,
+				leafKey.LeafScriptKey().PubKey,
 			),
 		},
 	}
@@ -5287,7 +5287,7 @@ func (r *rpcServer) AssetLeaves(ctx context.Context,
 
 // unmarshalLeafKey un-marshals a leaf key from the RPC form.
 func unmarshalLeafKey(key *unirpc.AssetKey) (universe.LeafKey, error) {
-	var leafKey universe.LeafKey
+	var leafKey universe.BaseLeafKey
 
 	switch {
 	case key.GetScriptKeyBytes() != nil:
@@ -5537,7 +5537,7 @@ func unmarshalUniverseKey(key *unirpc.UniverseKey) (universe.Identifier,
 
 	var (
 		uniID  = universe.Identifier{}
-		uniKey = universe.LeafKey{}
+		uniKey = universe.BaseLeafKey{}
 		err    error
 	)
 
@@ -5550,12 +5550,12 @@ func unmarshalUniverseKey(key *unirpc.UniverseKey) (universe.Identifier,
 		return uniID, uniKey, err
 	}
 
-	uniKey, err = unmarshalLeafKey(key.LeafKey)
+	leafKey, err := unmarshalLeafKey(key.LeafKey)
 	if err != nil {
 		return uniID, uniKey, err
 	}
 
-	return uniID, uniKey, nil
+	return uniID, leafKey, nil
 }
 
 // unmarshalAssetLeaf unmarshals an asset leaf from the RPC form.
--- a/tapdb/burn_tree.go
+++ b/tapdb/burn_tree.go
@@ -27,65 +27,18 @@ func NewBurnUniverseTree(db BatchedUniverseTree) *BurnUniverseTree {
 	return &BurnUniverseTree{db: db}
 }
 
-// burnSpecifierToIdentifier converts an asset.Specifier into a
-// universe.Identifier for the burn tree.
-func burnSpecifierToIdentifier(spec asset.Specifier) (universe.Identifier,
-	error) {
-
-	var id universe.Identifier
-
-	// The specifier must have a group key to be able to be used within the
-	// ignore tree context.
-	if !spec.HasGroupPubKey() {
-		return id, fmt.Errorf("group key must be set")
-	}
-
-	id.GroupKey = spec.UnwrapGroupKeyToPtr()
-	id.ProofType = universe.ProofTypeBurn
-
-	return id, nil
-}
-
 // Sum returns the sum of the burn leaves for the given asset.
 func (bt *BurnUniverseTree) Sum(ctx context.Context,
 	spec asset.Specifier) universe.BurnTreeSum {
 
 	// Derive identifier from the asset.Specifier.
-	id, err := burnSpecifierToIdentifier(spec)
+	id, err := specifierToIdentifier(spec, universe.ProofTypeBurn)
 	if err != nil {
 		return lfn.Err[lfn.Option[uint64]](err)
 	}
-	namespace := id.String()
 
-	var sumOpt lfn.Option[uint64]
-
-	readTx := NewBaseUniverseReadTx()
-	txErr := bt.db.ExecTx(ctx, &readTx, func(db BaseUniverseStore) error {
-		tree := mssmt.NewCompactedTree(
-			newTreeStoreWrapperTx(db, namespace),
-		)
-
-		// Get the root of the tree to retrieve the sum.
-		root, err := tree.Root(ctx)
-		if err != nil {
-			return err
-		}
-
-		// If root is empty, return empty sum.
-		if root.NodeHash() == mssmt.EmptyTreeRootHash {
-			return nil
-		}
-
-		// Return the sum from the root.
-		sumOpt = lfn.Some(root.NodeSum())
-
-		return nil
-	})
-	if txErr != nil {
-		return lfn.Err[lfn.Option[uint64]](txErr)
-	}
-
-	return lfn.Ok(sumOpt)
+	// Use the generic helper to get the sum.
+	return getUniverseTreeSum(ctx, bt.db, id)
 }
 
 // ErrNotBurn is returned when a proof is not a burn proof.
@@ -108,7 +61,7 @@ func (bt *BurnUniverseTree) InsertBurns(ctx context.Context,
 
 	// Derive identifier (and thereby the namespace) from the
 	// asset.Specifier.
-	id, err := burnSpecifierToIdentifier(spec)
+	id, err := specifierToIdentifier(spec, universe.ProofTypeBurn)
 	if err != nil {
 		return lfn.Err[[]*universe.AuthenticatedBurnLeaf](err)
 	}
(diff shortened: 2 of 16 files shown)


universe/interface.go
```diff
@@ -1235,3 +1236,90 @@ type IgnoreTree interface {
 	QueryTuples(context.Context, asset.Specifier,
 		...IgnoreTuple) TupleQueryResp
 }
+
+// BurnLeaf is a type that represents a burn leaf within the universe tree.
+type BurnLeaf struct {
+	// UniverseKey is the key that the burn leaf is stored at.
+	UniverseKey LeafKey
```
guggero, line 1243: Since we're creating a new tree, should we make sure we don't introduce the same collision issue again? Meaning we should use a fixes leaf key (that also includes the asset ID in case of a grouped asset) here?
Not sure if that changes the overall semantics though or would cause us to not be able to re-use some of the existing code. Mostly just thinking aloud.

universe/interface.go
```diff
@@ -1235,3 +1236,90 @@ type IgnoreTree interface {
...
+	// BurnProof is the universe inclusion proof for the burn leaf within
+	// the universe tree.
+	BurnProof *mssmt.Proof
+}
+
+// BurnDesc is a type that represents a burn leaf within the universe tree. This
+// is useful for querying the state without needing the proof itself.
+type BurnDesc struct {
+	// AssetSpec is the asset specifier for the burn leaf.
+	AssetSpec asset.Specifier
```
guggero, line 1279: Ah, I see we're using the specifier here. So perhaps the collision issue isn't present here, but still going through the diff...

tapdb/burn_tree.go
```diff
@@ -0,0 +1,425 @@
...
+				RawProof: rawProofBytes,
+				Asset:    &burnLeaf.BurnProof.Asset,
+				Amt:      burnLeaf.BurnProof.Asset.Amount,
+			}
+
+			// Call the generic upsert function. MetaReveal is nil
+			// for burns, as this isn't an issuance instance. We
+			// also skip inserting into the multi-verse tree for
+			// now.
+			uniProof, err := universeUpsertProofLeaf(
```
guggero, line 167: Okay, I see we can't just fix the collision issue here by using a different `leafKey` construction, otherwise we couldn't re-use this code.
Then I guess we'll just have to fix the universe collision issue as a whole in a separate PR.

tapdb/burn_tree.go
```diff
@@ -0,0 +1,425 @@
...
+				return fmt.Errorf("unable to upsert burn "+
+					"leaf for key %v: %w", leafKey, err)
+			}
+
+			insertedProofs[leafKey] = uniProof
+		}
+
+		// Now, construct the AuthenticatedBurnLeaf results.
+		for _, burnLeaf := range burnLeaves {
+			uniProof, ok := insertedProofs[burnLeaf.UniverseKey]
```
guggero, line 180: I think we should use `burnLeaf.UniverseKey.UniverseKey()` instead, which serializes and hashes the key's content to avoid issues with how public keys can be represented differently even though they are the same point in a curve

tapdb/burn_tree_test.go
```diff
@@ -0,0 +1,611 @@
...
+		require.NoError(t, err)
+		queryLeaves := queryOpt.UnwrapOrFail(t)
+
+		// There should be exactly one leaf for this outpoint+scriptkey
+		// combination, not multiple.
+		count := 0
+		for _, leaf := range queryLeaves {
+			if leaf.BurnLeaf.UniverseKey.OutPoint ==
+				burnLeaf.UniverseKey.OutPoint &&
+				leaf.BurnLeaf.UniverseKey.ScriptKey.PubKey.IsEqual( //nolint:lll
```
guggero, line 274: nit: can add `//nolint` directives above a block (e.g. above the `if`) so it counts for the whole section.

guggero (commented):
Did a first pass, very nice! Nothing major, so should be pretty close.
Nice unit tests too :100: 

CI failures seem to be flakes.

Roasbeef:
Pushed up a commit to consolidate a lot of the duplication between the ignore and burn trees.

tapdb/burn_tree.go
Roasbeef: I think what we can do is make a slim interface that's just `UniverseKey()`, then we make a wrapper version that includes `asset.ID`, then update the function to use that.

I think that should resolve the issue here? Then that's a bit of prep for making this change in other areas.

Roasbeef:
Pushed up a commit to solve the collision issues. We add a new interface for `LeafKey`, then make a new concrete `BaseLeafKey` implementation. This way any methods that needs the old leaf key can continue to accept it, so we're able to re-use many helper functions. We then add a `UniqueLeafKey` interface, along with a concrete implementation that'll be used by the burn tree. 

For the ignore tree, we use the hash of the prev ID as a key, which already includes the asset ID, so we're safe there.

tapdb/burn_tree.go
```diff
@@ -204,221 +157,189 @@ func (bt *BurnUniverseTree) InsertBurns(ctx context.Context,
...
-
-				if len(dbLeaves) == 0 {
-					continue
-				}
-
-				leavesToQuery = append(
-					leavesToQuery, dbLeaves...,
-				)
+	// If specific burn points are provided, query for each of them.
+	if len(burnPoints) > 0 {
```
guggero, line 177: nit: since we're in a helper function now, we could invert the condition of this `if`, do the general `QueryUniverseLeaves` query and return. Then we could do this block un-indented further down, saving a whole indentation-level for the nested loop.

tapdb/burn_tree.go
```diff
@@ -204,221 +157,189 @@ func (bt *BurnUniverseTree) InsertBurns(ctx context.Context,
...
+	}
+
+	return burnLeaf, leafKey.UniverseKey(), nil
 }
 
-// ListBurns attempts to list all burn leaves for the given asset.
-func (bt *BurnUniverseTree) ListBurns(ctx context.Context,
-	spec asset.Specifier) universe.ListBurnsResp {
+// buildAuthBurnLeaf constructs the final AuthenticatedBurnLeaf.
+func buildAuthBurnLeaf(decodedLeaf *universe.BurnLeaf,
```
guggero, line 267: nit: remove error return value since it's never used?

tapdb/ignore_tree.go
```diff
@@ -171,100 +157,124 @@ func (it *IgnoreUniverseTree) Sum(ctx context.Context,
...
-		// leaves for this namespace. The namespace is derived from the
-		// group key, and the proof type, which in this case is ignore.
-		universeLeaves, err := db.QueryUniverseLeaves(
-			ctx, UniverseLeafQuery{
-				Namespace: namespace,
-			},
-		)
+		leaves, err := dbtx.QueryUniverseLeaves(ctx, leafQuery)
+		if errors.Is(err, sql.ErrNoRows) {
+			continue // Skip if no leaves found for this tuple
```
guggero, line 230: nit: code style.

tapdb/universe.go
```diff
@@ -112,6 +113,243 @@ type BaseUniverseStore interface {
...
+// The LeafType is the concrete type of the leaf, AuthType is the type of the
+// wrapper of the LeafType that includes MS-SMT merkle proof info, and finally
+// the QueryType is the type that is used to query the leaves.
+func queryUniverseLeavesAndProofs[LeafType any, AuthType any, QueryType any](
+	ctx context.Context, db BatchedUniverseTree, assetSpec asset.Specifier,
+	id universe.Identifier,
+
+	// queryFunc retrieves the relevant UniverseLeaf records based on the
+	// specific query parameters.
+	queryFunc func(context.Context, BaseUniverseStore, asset.Specifier,
```
guggero, line 196: Should we extract these function definitions into their own types to a) shorten this and b) have their descriptive comments moved to the type. Since these inline function parameter comments are really unusual for any of our repos.

tapdb/universe.go
```diff
@@ -112,6 +113,243 @@ type BaseUniverseStore interface {
...
+	// decodeFunc decodes the raw proof bytes from a UniverseLeaf into a
+	// domain-specific type and reconstructs the LeafKey.
+	decodeFunc func(UniverseLeaf) (LeafType, uniKey, error),
+
+	// buildAuthFunc constructs the final authenticated result type using
+	// the decoded leaf, its inclusion proof, and the tree root.
+	buildAuthFunc func(LeafType, *mssmt.Proof, mssmt.Node) (AuthType, error), //nolint:lll
+
+	// queryParams are the specific parameters for the queryFunc.
+	queryParams ...QueryType) (lfn.Result[lfn.Option[[]AuthType]], error) {
```
guggero, line 208: Why would we have both a result _and_ an error return value?

tapdb/burn_tree.go
```diff
@@ -204,221 +157,189 @@ func (bt *BurnUniverseTree) InsertBurns(ctx context.Context,
...
-					BurnProof:   &burnProof,
-				},
-				BurnTreeRoot: root,
-				BurnProof:    inclusionProof,
-			}
-			resultLeaves = append(resultLeaves, authLeaf)
-		}
+// decodeAndBuildAuthBurnLeaf decodes the raw leaf, reconstructs the key, and
+// builds the AuthenticatedBurnLeaf.
+func decodeAndBuildAuthBurnLeaf(dbLeaf UniverseLeaf) (
```
guggero, line 232: So here we're using the:
```
func myFuncName(param, param) (
	returnVal, returnVal) {
```

style, while further down we're using

```
func myFuncName(param, param
) (returnVal, returnVal) {
```

Can we please at least get consistency in terms of this decision? I way prefer the first one, for reasons previously stated.

tapdb/ignore_tree.go
```diff
@@ -171,100 +157,124 @@ func (it *IgnoreUniverseTree) Sum(ctx context.Context,
...
+// buildAuthIgnoreTuple constructs the final AuthenticatedIgnoreTuple.
+func buildAuthIgnoreTuple(decodedLeaf *universe.SignedIgnoreTuple,
+	proof *mssmt.Proof,
+	root mssmt.Node) (universe.AuthenticatedIgnoreTuple, error) {
+
+	return universe.AuthenticatedIgnoreTuple{
+		SignedIgnoreTuple: decodedLeaf,
+		InclusionProof:    proof,
+		IgnoreTreeRoot:    root,
+	}, nil
```
guggero, line 277: nit: another error return value we can remove.

tapdb/universe.go
```diff
@@ -112,6 +113,243 @@ type BaseUniverseStore interface {
...
+//
+// We accept and return an abstract type T, which can be created by reading the
+// raw value of the universe leaf, and decoding that.
+func listUniverseLeaves[T any](ctx context.Context, db BatchedUniverseTree,
+	id universe.Identifier,
+
+	// decodeFunc decodes the raw proof bytes from a UniverseLeaf into the
+	// desired domain-specific type.
+	decodeFunc func(UniverseLeaf) (T, error),
+) (lfn.Result[lfn.Option[[]T]], error) {
```
guggero, line 306: Again, why result _and_ error?

universe/interface.go
```diff
@@ -262,11 +262,24 @@ func (m *Leaf) SmtLeafNode() *mssmt.LeafNode {
...
 }
 
-// LeafKey is the top level leaf key for a universe. This will be used to key
-// into a universe's MS-SMT data structure. The final serialized key is:
-// sha256(mintingOutpoint || scriptKey). This ensures that all
-// leaves for a given asset will be uniquely keyed in the universe tree.
-type LeafKey struct {
+// LeafKey is an interface that allows us to obtain the universe key for a leaf
+// within a universe.
+type LeafKey interface {
```
guggero, line 267: Very nice solution! Awesome to be able to leave behind the universe collision issue for new trees :100:

guggero (approved):
Very nice refactor, code re-use and collision avoidance!
Code looks good to me (apart from a few mostly code style related nits, Aider still seems to struggle with our preferred code style. Or I struggle with its decisions, however you want to see it :sweat_smile: )

tapdb/burn_tree.go
```diff
@@ -0,0 +1,426 @@
...
+				return fmt.Errorf("unable to upsert burn "+
+					"leaf for key %v: %w", leafKey, err)
+			}
+
+			insertedProofs[leafKey.UniverseKey()] = uniProof
+		}
+
+		// Now, construct the AuthenticatedBurnLeaf results.
+		for _, burnLeaf := range burnLeaves {
+			leafKey := burnLeaf.UniverseKey.UniverseKey()
```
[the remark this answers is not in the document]
Roasbeef: Ah sure, we don't need need this second loop. Collapsed the creation of `AuthenticatedBurnLeaf` into the prev loop.

tapdb/universe.go
Roasbeef: No good reason...

Removed.

tapdb/universe.go
Roasbeef: Sure. Earlier I went back and forth between making this a struct w/ an interface vs the closure approach. Ended up with the closure approach.

Roasbeef:
Pushed up latest iteration. This is dependent on the ignore tree PR.

ffranr:
I've changed base branch to `main` now that `ignore-tree-impl` PR is merged.

tapdb/ignore_tree.go
```diff
@@ -171,98 +158,121 @@ func (it *IgnoreUniverseTree) Sum(ctx context.Context,
...
+	}
 
-	if txErr != nil {
-		return lfn.Err[[]*universe.IgnoreTuple](txErr)
+	return allLeaves, nil
+}
+
+// decodeAndBuildAuthIgnoreTuple decodes the raw leaf, reconstructs the key,
+// and builds the AuthenticatedIgnoreTuple.
+func decodeAndBuildAuthIgnoreTuple(dbLeaf UniverseLeaf,
```
ffranr, line 253: maybe function name could be: `parseDbSignedIgnoreTuple`

tapdb/ignore_tree.go
```diff
@@ -171,98 +158,121 @@ func (it *IgnoreUniverseTree) Sum(ctx context.Context,
...
+		return nil, uniKey{}, fmt.Errorf("error decoding tuple: %w",
+			err)
 	}
 
-	return lfn.Ok(tuples)
+	return &signedTuple, signedTuple.UniverseKey(), nil
+}
+
+// buildAuthIgnoreTuple constructs the final AuthenticatedIgnoreTuple.
+func buildAuthIgnoreTuple(decodedLeaf *universe.SignedIgnoreTuple,
```
ffranr, line 268: Perhaps this can become a `universe.AuthenticatedIgnoreTuple` constructor called `NewAuthIgnoreTuple`

ffranr: BTW, I'm getting an error in my IDE because `decodedLeaf` is a pointer where as field `SignedIgnoreTuple` is not.

rpcserver.go
```diff
@@ -5276,7 +5276,7 @@ func (r *rpcServer) AssetLeaves(ctx context.Context,
 
 // unmarshalLeafKey un-marshals a leaf key from the RPC form.
 func unmarshalLeafKey(key *unirpc.AssetKey) (universe.LeafKey, error) {
-	var leafKey universe.LeafKey
+	var leafKey universe.BaseLeafKey
```
ffranr, line 5279: Perhaps we can follow the same pattern we used for the `GroupKeyReveal` upgrade from V0 to V1. That would mean introducing a new interface, renaming `universe.LeafKey` to `universe.LeafKeyV0`, and adding a new `universe.LeafKeyV1`, rather than layering changes on top of a shared base. Might help keep the original V0 implementation clearly identifiable.

universe/interface.go
```diff
@@ -1237,3 +1238,90 @@ type IgnoreTree interface {
...
 		...IgnoreTuple) TupleQueryResp
 }
+
+// BurnLeaf is a type that represents a burn leaf within the universe tree.
+type BurnLeaf struct {
+	// UniverseKey is the key that the burn leaf is stored at.
+	UniverseKey LeafKey
+
+	// BurnProof is the burn proof that is stored within the burn leaf.
+	BurnProof *proof.Proof
```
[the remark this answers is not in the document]
Roasbeef: Mainly following convention here with the rest of the codebase. The proof can be up to 8 KB ignoring run time overhead, the heap is an apt place for the contents IMO.

tapdb/ignore_tree.go
Roasbeef: Good suggestion, much shorter as well.

tapdb/ignore_tree.go
Roasbeef: > BTW, I'm getting an error in my IDE because decodedLeaf is a pointer where as field SignedIgnoreTuple is not.

This happened post rebase, should be resolved in the latest push.

Roasbeef: Done!

rpcserver.go
Roasbeef: I think the `GroupKeyReveal` case is different, as that changed affected the proof/wire protocol. With this change that you're commenting on, there's no semantic change here: the universe key is untouched. This just positions us to update it later to use the new unique universe key variant. 

For the move, rather than p2p changes, the changes will be based on the RPC interface. We'd add a version to the various queries, then add additional validation based on that version. We'd keep a snapshot of the legacy tree, migrate things over to the new tree, but then only support new _insertions_ into the new tree.

tapdb/burn_tree.go
Roasbeef: I know it can seem a bit arbitrary, but it's just based on personal aesthetics. I like the second option when we have multiple return values: IMO the delineation between the return value and args is clear. Breaking early at say the second return value is also an option, but I prefer to have them be on the same line mostly.

ffranr (approved):
LGTM!

PR #1457, opened 2025-03-27
Multi rfq receive (`AddInvoice` multiple hop hints)

### LINKED ISSUE #1359: [feature]: Multi-RFQ receive
## Description

When generating an invoice, we are restricted to a specific assetID and a single peer. Even if that peer agreed on an RFQ quote with us, they may never be able to forward the payment to us because of their inbound. We should investigate ways of encoding multiple RFQs / entry-points to tapd and escape the single link bottleneck.

Similar to [Multi-RFQ send](https://github.com/lightninglabs/taproot-assets/issues/1358), but a bit more difficult as we need to create a bolt11 compatible invoice that will be payable by all LN implementations, so we can't pull a lot of custom tricks. Multi-RFQ send has the convenience of being resolved within the scope of the local `tapd`<>`lnd`.

Items to investigate:

- [x] Even if we did acquire multiple RFQs before creating the invoice, how can we encode them all in the Bolt11 format
- [ ] ~Use blinded paths, which can give up to 3 entry points (possibility for 3 RFQs?)~
- [ ] ~Present multiple RFQ with interactive bolt12 offer communication~

GeorgeTsagk:
## Description

This PR introduces the ability for tapd nodes to create invoices which involve multiple peer quotes. When the `peerPubKey` is left unspecified in the `AddInvoice` RPC we no longer return an error, but instead acquire quotes with all peers that have a valid asset channel with us.

Within this PR we also extract some rfq/liquidity related functions to the `rfq` package to keep `rpcserver.go` more clean.

Closes #1359 

Based on #1423 (using that as base branch for now to avoid bloated diff)

### DIFF
--- a/rfq/manager.go
+++ b/rfq/manager.go
@@ -41,7 +41,9 @@ const (
 type ChannelLister interface {
 	// ListChannels returns a list of channels that are available for
 	// routing.
-	ListChannels(ctx context.Context) ([]lndclient.ChannelInfo, error)
+	ListChannels(ctx context.Context, activeOnly, publicOnly bool,
+		_ ...lndclient.ListChannelsOption) ([]lndclient.ChannelInfo,
+		error)
 }
 
 // ScidAliasManager is an interface that can add short channel ID (SCID) aliases
@@ -548,7 +550,7 @@ func (m *Manager) addScidAlias(scidAlias uint64, assetSpecifier asset.Specifier,
 
 	// Retrieve all local channels.
 	ctxb := context.Background()
-	localChans, err := m.cfg.ChannelLister.ListChannels(ctxb)
+	localChans, err := m.cfg.ChannelLister.ListChannels(ctxb, true, false)
 	if err != nil {
 		// Not being able to call lnd to add the alias is a critical
 		// error, which warrants shutting down, as something is wrong.
--- a/rfq/manager.go
+++ b/rfq/manager.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sort"
 	"sync"
 	"time"
 
@@ -1070,62 +1071,83 @@ type ChannelWithSpecifier struct {
 	AssetInfo rfqmsg.JsonAssetChannel
 }
 
+// PeerChanMap is a structure that maps peers to channels. This is used for
+// filtering asset channels against an asset specifier.
+type PeerChanMap map[route.Vertex][]ChannelWithSpecifier
+
 // ComputeCompatibleChannelAssetBalance computes the total local and remote
 // balance for each asset channel that matches the provided asset specifier.
 func (m *Manager) ComputeCompatibleChannelAssetBalance(ctx context.Context,
 	activeChannels []lndclient.ChannelInfo,
-	specifier asset.Specifier) ([]ChannelWithSpecifier, bool, error) {
+	specifier asset.Specifier) (PeerChanMap, bool, error) {
 
 	var (
-		channels                 = make([]ChannelWithSpecifier, 0)
+		peerChanMap              = make(PeerChanMap)
 		haveGroupedAssetChannels bool
 	)
 	for chanIdx := range activeChannels {
-		openChan := activeChannels[chanIdx]
-		if len(openChan.CustomChannelData) == 0 {
-			continue
-		}
-
-		var assetData rfqmsg.JsonAssetChannel
-		err := json.Unmarshal(openChan.CustomChannelData, &assetData)
-		if err != nil {
-			return nil, false, fmt.Errorf("unable to unmarshal "+
-				"asset data: %w", err)
-		}
-
-		if len(assetData.GroupKey) > 0 {
-			haveGroupedAssetChannels = true
-		}
-
-		// Check if the assets of this channel match the provided
-		// specifier.
-		pass, err := m.ChannelMatchesFully(
-			ctx, assetData, specifier,
+		var (
+			pass      bool
+			assetData rfqmsg.JsonAssetChannel
 		)
-		if err != nil {
-			return nil, false, err
+
+		openChan := activeChannels[chanIdx]
+
+		// If there specifier is empty, we skip all asset balance
+		// related checks.
+		if specifier.IsSome() {
+			if len(openChan.CustomChannelData) == 0 {
+				continue
+			}
+
+			err := json.Unmarshal(
+				openChan.CustomChannelData, &assetData,
+			)
+			if err != nil {
+				return nil, false, fmt.Errorf("unable to "+
+					"unmarshal asset data: %w", err)
+			}
+
+			if len(assetData.GroupKey) > 0 {
+				haveGroupedAssetChannels = true
+			}
+
+			// Check if the assets of this channel match the
+			// provided specifier.
+			pass, err = m.ChannelMatchesFully(
+				ctx, assetData, specifier,
+			)
+			if err != nil {
+				return nil, false, err
+			}
 		}
 
-		if pass {
-			channels = append(channels, ChannelWithSpecifier{
-				Specifier:   specifier,
-				ChannelInfo: openChan,
-				AssetInfo:   assetData,
-			})
+		// We also append the channel in the case where the specifier is
+		// empty. This means that the caller doesn't really care about
+		// the type of balance.
+		if pass || !specifier.IsSome() {
+			peerChanMap[openChan.PubKeyBytes] = append(
+				peerChanMap[openChan.PubKeyBytes],
+				ChannelWithSpecifier{
+					Specifier:   specifier,
+					ChannelInfo: openChan,
+					AssetInfo:   assetData,
+				},
+			)
 		}
 	}
 
-	return channels, haveGroupedAssetChannels, nil
+	return peerChanMap, haveGroupedAssetChannels, nil
 }
 
-// chanIntention defines the intention of calling rfqChannel. This helps with
+// ChanIntention defines the intention of calling rfqChannel. This helps with
 // returning the channel that is most suitable for what we want to do.
-type chanIntention uint8
+type ChanIntention uint8
 
 const (
 	// NoIntention defines the absence of any intention, signalling that we
 	// don't really care which channel is returned.
-	NoIntention chanIntention = iota
+	NoIntention ChanIntention = iota
 
 	// SendIntention defines the intention to send over an asset channel.
 	SendIntention
@@ -1135,13 +1157,12 @@ const (
 	ReceiveIntention
 )
 
-// RfqChannel returns the channel to use for RFQ operations. If a peer public
-// key is specified, the channels are filtered by that peer. If there are
-// multiple channels for the same specifier, the user must specify the peer
-// public key.
-func (m *Manager) RfqChannel(ctx context.Context,
-	specifier asset.Specifier, peerPubKey *route.Vertex,
-	intention chanIntention) (*ChannelWithSpecifier, error) {
+// RfqChannel returns the channel to use for RFQ operations. It returns a map of
+// peers and their eligible channels. If a peerPubKey is specified then the map
+// will only contain one entry for that peer.
+func (m *Manager) RfqChannel(ctx context.Context, specifier asset.Specifier,
+	peerPubKey *route.Vertex,
+	intention ChanIntention) (PeerChanMap, error) {
 
 	activeChannels, err := m.cfg.ChannelLister.ListChannels(
 		ctx, true, false,
@@ -1150,7 +1171,8 @@ func (m *Manager) RfqChannel(ctx context.Context,
 		return nil, err
 	}
 
-	balances, haveGroupChans, err := m.ComputeCompatibleChannelAssetBalance(
+	// nolint:lll
+	balancesMap, haveGroupChans, err := m.ComputeCompatibleChannelAssetBalance(
 		ctx, activeChannels, specifier,
 	)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


tapchannel/aux_traffic_shaper.go
```diff
@@ -308,11 +310,20 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
...
+	case quote.Request.AssetSpecifier.HasId():
+		assetId = *quote.Request.AssetSpecifier.UnwrapIdToPtr()
+
+	case quote.Request.AssetSpecifier.HasGroupPubKey():
+		// If a group key is defined in the quote we place the hash of
+		// the group key as the dummy asset ID in the HTLC. This asset
+		// balance in the HTLC is just a hint and the actual asset IDs
+		// will be picked later in the process.
+		groupKey := quote.Request.AssetSpecifier.UnwrapGroupKeyToPtr()
+		groupKeyX := schnorr.SerializePubKey(groupKey)
```
Roasbeef, line 325: Commit says hash, but we use a serialized pubkey here.

Roasbeef: On the decoding side, do we take any steps to ensure that when we try to figure out if it's an asset ID or group key, we don't mistakenly think an asset ID is actually a group key? To elaborate a bit, there's a small probability that a given hash is actually a valid x-coord. This is how the NUMS key derivation works. 

I guess we can check our local DB or some other source to verify that it's actually a group key or asset ID?

taprpc/tapchannelrpc/tapchannel.proto
```diff
@@ -142,6 +142,10 @@ message SendPaymentRequest {
     // are sent out to the network than the invoice amount plus routing fees
     // require to be paid.
     bool allow_overpay = 6;
+
+    // The group key which dictates which assets may be used for this payment.
+    // Mutually exclusive to asset_id.
+    bytes group_key = 7;
```
Roasbeef, line 150: Thought this was already in the other PR as a commit?

rpcserver.go
```diff
@@ -6432,6 +6432,31 @@ func MarshalAssetFedSyncCfg(
 	}, nil
 }
 
+// marshalAssetSpecifier marshals an asset specifier to the RPC form.
```
Roasbeef, line 6435: So we don't allow the combo here? Of group key _and_ an asset ID.

taprpc/tapchannelrpc/tapchannel.proto
Roasbeef: Ah ok, it appears that this one wasn't rebased after changes to the base PR.

rpcserver.go
```diff
@@ -7991,177 +7995,6 @@ func (r *rpcServer) DecDisplayForAssetID(ctx context.Context,
...
-		// specifier.
-		pass, err := r.cfg.RfqManager.ChannelCompatible(
-			ctx, assetData.Assets, specifier,
-		)
-		if err != nil {
-			return nil, err
-		}
-
-		if pass {
-			// Since the assets of the channel passed the above
```
Roasbeef, line 8076: 👍

rfq/manager.go
```diff
@@ -1066,9 +1067,11 @@ type ChannelWithSpecifier struct {
 // each asset channel that matches the provided asset specifier.
 func (m *Manager) ComputeChannelAssetBalance(ctx context.Context,
 	activeChannels []lndclient.ChannelInfo,
-	specifier asset.Specifier) ([]ChannelWithSpecifier, error) {
+	specifier asset.Specifier) (map[route.Vertex][]ChannelWithSpecifier,
```
Roasbeef, line 1073: Style nit: can use a type def here to give the data structure a more descriptive name/type, and also cut down on the amt of chars needed to ref it a bit.

rfq/manager.go
```diff
@@ -1104,25 +1107,39 @@ func (m *Manager) ComputeChannelAssetBalance(ctx context.Context,
 				aggrInfo.RemoteBalance += info.RemoteBalance
 			}
 
-			channels = append(channels, ChannelWithSpecifier{
+			_, ok := peerChanMap[openChan.PubKeyBytes]
+			if !ok {
+				peerChanMap[openChan.PubKeyBytes] =
```
Roasbeef, line 1115: I think since this isn't a nested map, you can actually just ignore `ok` here and append to the return value, which will automatically allocate space as needed.

rfq/manager.go
```diff
@@ -1046,6 +1046,185 @@ func (m *Manager) ChannelCompatible(ctx context.Context,
...
+	// committed to.
+	ChannelInfo lndclient.ChannelInfo
+
+	// AssetInfo contains the asset related info of the channel.
+	AssetInfo rfqmsg.JsonAssetChanInfo
+}
+
+// ComputeChannelAssetBalance computes the total local and remote balance for
+// each asset channel that matches the provided asset specifier.
+func (m *Manager) ComputeChannelAssetBalance(ctx context.Context,
```
Roasbeef, line 1071: I think this code move is also a good opportunity to add some test coverage to these functions. Once we add varying RFQ selection strategies, the scope of the functions will expand, so good to nail down some test coverage now while we're at it. 

a lil `aider` can prob go a long way here

rfq/manager.go
```diff
@@ -1132,97 +1149,67 @@ const (
...
-				bestBalance.AssetInfo.LocalBalance {
-
-				bestBalance = b
-			}
-		})
-
+		// When sending we care about the volume of our local balances,
+		// so we sort by local balances in descending order.
+		for _, v := range balancesMap {
+			sort.Slice(v, func(i, j int) bool {
```
Roasbeef, line 1186: Hmm, will this actually mutate the underlying value here (the slice) captured by the loop? 

Would favor making the re-assignment explicit here, which may guard against future foot guns.

rfq/manager.go
```diff
@@ -1132,97 +1149,67 @@ const (
...
-		// Do nothing. Just return the first element that was
-		// assigned above.
+		// We don't care about sending or receiving, this means that
+		// the method was called as a dry check. Do nothing.
+	}
+
+	// If a peer public key was specified, we always want to use that to
+	// filter the asset channels.
+	if peerPubKey != nil {
+		_, ok := balancesMap[*peerPubKey]
```
Roasbeef, line 1208: 👍

rfq/manager.go
```diff
@@ -1212,6 +1215,61 @@ func (m *Manager) RfqChannel(ctx context.Context,
...
+type InboundPolicyFetcher func(ctx context.Context, chanID uint64,
+	remotePubStr string) (*lnrpc.RoutingPolicy, error)
+
+// RfqToHopHint creates the hop hint representation which encapsulates certain
+// quote information along with some other data required by the payment to
+// succeed.
+func (m *Manager) RfqToHopHint(ctx context.Context,
+	policyFetcher InboundPolicyFetcher, channelID uint64,
+	peerPubKey route.Vertex, quote *rfqrpc.PeerAcceptedBuyQuote,
+	hold bool) (*lnrpc.HopHint, []zpay32.HopHint, error) {
```
Roasbeef, line 1229: We do we return a single RPC hop hint, but a slice of normal hop hints? 

I think the `godoc` comment could use some expansion. Going a bit futher, we may want to consider making a new wrapper type here depending on the end use case (does the client sometimes need to map between the two versions?).

rfq/manager.go
```diff
@@ -1212,6 +1215,61 @@ func (m *Manager) RfqChannel(ctx context.Context,
...
+				FeeProportionalMillionths: uint32(
+					inboundPolicy.FeeRateMilliMsat,
+				),
+				CLTVExpiryDelta: uint16(
+					inboundPolicy.TimeLockDelta,
+				),
+			},
+		}
+
+		return nil, hopHint, nil
```
Roasbeef, line 1257: As we only ever return a single hop hint here (right now), seems we can just return it directly.

Roasbeef: If we return that directly, then maybe we can make a thin wrapper struct which can take our "rfq" hop hint, and map it into either the `zpay32` version (just return the embedded version), or map it into the RPC version.

rpcserver.go
```diff
@@ -7926,11 +7926,64 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 	}
 
 	return &tchrpc.AddInvoiceResponse{
-		AcceptedBuyQuote: acceptedQuote,
+		// TODO(george): For now we just return the expensive quote.
+		AcceptedBuyQuote: expensiveQuote,
```
Roasbeef, line 7901: I don't think this compiles, `expensiveQuote` isn't created/assigned above.

rpcserver.go
```diff
@@ -7926,11 +7926,64 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
 		InvoiceResult:    invoiceResp,
 	}, nil
 }
 
+func (r *rpcServer) AcquireBuyOrder(ctx context.Context,
```
Roasbeef, line 7906: Missing `godoc` comment.

Roasbeef: Also does this _need_ to live in the RPC server? If not, then just like the refactor earlier, we can move it into the `rfq` package. This way we can add tests w/o needing to mock out the various RPC sever deps.

rpcserver.go
```diff
@@ -7768,81 +7759,62 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+		quote   *rfqrpc.PeerAcceptedBuyQuote
+		rate    *rfqmath.BigIntFixedPoint
+		channel rfq.ChannelWithSpecifier
 	}
 
-	var acceptedQuote *rfqrpc.PeerAcceptedBuyQuote
-	switch r := resp.Response.(type) {
-	case *rfqrpc.AddAssetBuyOrderResponse_AcceptedQuote:
-		acceptedQuote = r.AcceptedQuote
+	acquiredQuotes := make([]quoteWithInfo, 0)
```
Roasbeef, line 7768: Style nit: can just declare as a variable instead: 
```go
var acquiredQuotes []quoteWithInfo
```

If it needs to escape, then Go will handle that.

rpcserver.go
```diff
@@ -7768,81 +7759,62 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
 
-	case *rfqrpc.AddAssetBuyOrderResponse_InvalidQuote:
-		return nil, fmt.Errorf("peer %v sent back an invalid quote, "+
-			"status: %v", r.InvalidQuote.Peer,
-			r.InvalidQuote.Status.String())
+	// TODO(george): should we limit the max number of acquired quotes?
```
Roasbeef, line 7770: So there's an upper limit when it comes to invoices (one in terms of the QR code, the other in terms of the max bolt11 invoice size). 

lnd will return an error if a created invoice is about to exceed the max size: https://github.com/lightningnetwork/lnd/blob/b01f4e5148cc3c77d66681df817caef13eaa3e81/zpay32/invoice.go#L84-L87. 

So based on the knowledge of the size of other params, we can derive a max hop hint value (lnd might already have such a derivation somewhere). 

Either way, we'll want to pick an upper limit. Given that upper limit, we'll also want to make a decision w.r.t which set of accepted quotes we actually want to insert into the hop hint. 

Thinking about it a bit more, if we pipeline an RFQ request to 50 peers, but we only end up accepting 10 of them (better price or w/e), we should probably cancel that set of accepted quotes on the other peers.

rpcserver.go
```diff
@@ -7768,81 +7759,62 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+		if err != nil {
+			return nil, err
+		}
 
-	default:
-		return nil, fmt.Errorf("unexpected response type: %T", r)
+		acquiredQuotes = append(acquiredQuotes, quoteWithInfo{
+			quote: quote,
+			rate:  rate,
+			// Since the channels are sorted, we now the value with
```
Roasbeef, line 7792: now -> know

rpcserver.go
```diff
@@ -7768,81 +7759,62 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+			// the greatest remote balance is at index 0.
+			channel: channels[0],
```
Roasbeef, line 7794: Shouldn't we retain _all_ the channels? As we can add many hop hints for a given edge to the invoice. Or is the idea that we want to start doing that in another change?

rpcserver.go
```diff
@@ -7768,81 +7759,62 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
-	// payable by the network.
-	if acceptedQuote.MinTransportableUnits > req.AssetAmount {
-		return nil, fmt.Errorf("cannot create invoice over %d asset "+
-			"units, as the minimal transportable amount is %d "+
-			"units with the current rate of %v units/BTC",
-			req.AssetAmount, acceptedQuote.MinTransportableUnits,
-			acceptedQuote.AskAssetRate)
-	}
+	// Let's sort the ask rate of the quotes in ascending order.
+	sort.Slice(acquiredQuotes, func(i, j int) bool {
```
Roasbeef, line 7799: Solid initial strategy. 

In the future, we'll also want to take into account what the allotted amount of volume is. For now, since we basically never have standing orders (we req just enough to recv), this isn't strictly required.

rpcserver.go
```diff
@@ -7768,81 +7759,62 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
-	)
-	if err != nil {
-		return nil, fmt.Errorf("error unmarshalling ask asset rate: %w",
-			err)
-	}
+	// We grab the most expensive rate to use as reference for the total
+	// invoice amount. Since peers have varying prices for the assets, we
+	// pick the most expensive rate in order to allow for any combination of
+	// MPP shards through our set of chosen peers.
+	expensiveRate := acquiredQuotes[0].rate
```
Roasbeef, line 7808: Why should we pick the most expensive quote? This is effectively the same as giving the sender an option for the most expensive route in terms of fees.

rpcserver.go
```diff
@@ -7768,81 +7759,62 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+	expensiveQuote := acquiredQuotes[0].quote
 
 	// Convert the asset amount into a fixed-point.
 	assetAmount := rfqmath.NewBigIntFixedPoint(req.AssetAmount, 0)
 
 	// Calculate the invoice amount in msat.
-	valMsat := rfqmath.UnitsToMilliSatoshi(assetAmount, *askAssetRate)
+	valMsat := rfqmath.UnitsToMilliSatoshi(assetAmount, *expensiveRate)
```
Roasbeef, line 7815: Ah ok, I think this is making a bit more sense now: ultimately we need to put _some_ value into the invoice. I think we'll want to do another level of filtering here to ensure that we don't force the user to overpay. Or will the acceptance deviation check handle that?

rpcserver.go
```diff
@@ -7894,31 +7861,35 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
-		FeeProportionalMillionths: uint32(
-			inboundPolicy.FeeRateMilliMsat,
-		),
-		CltvExpiryDelta: inboundPolicy.TimeLockDelta,
-	}
-	iReq.RouteHints = []*lnrpc.RouteHint{
-		{
+	routeHints := make([]*lnrpc.RouteHint, 0)
+	for _, v := range acquiredQuotes {
+		hopHint, _, err := r.cfg.RfqManager.RfqToHopHint(
```
Roasbeef, line 7875: 👍

rpcserver.go
```diff
@@ -7894,31 +7861,35 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
-		ChanId:      acceptedQuote.Scid,
-		FeeBaseMsat: uint32(inboundPolicy.FeeBaseMsat),
-		FeeProportionalMillionths: uint32(
-			inboundPolicy.FeeRateMilliMsat,
-		),
-		CltvExpiryDelta: inboundPolicy.TimeLockDelta,
-	}
-	iReq.RouteHints = []*lnrpc.RouteHint{
-		{
+	routeHints := make([]*lnrpc.RouteHint, 0)
```
Roasbeef, line 7873: Same here re var vs make.

Roasbeef (changes requested):

rpcserver.go
Roasbeef: Gotcha, I think we can at least make it private then? Otherwise it looks like an RPC we're implementing vs a helper method.

rpcserver.go
```diff
@@ -7785,6 +7790,7 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 		return nil, fmt.Errorf("invoice request must be specified")
 	}
 	iReq := req.InvoiceRequest
+	existingQuotes := iReq.RouteHints != nil
```
Roasbeef, line 7793: Hmm, perhaps we should actually make validation stricter here and _disallow_ adding extra route hints? Otherwise a user can add hop hints that affect our reachability, or encodes the wrong details, etc.

rpcserver.go
```diff
@@ -7853,63 +7849,88 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
-	case *rfqrpc.AddAssetBuyOrderResponse_AcceptedQuote:
-		acceptedQuote = r.AcceptedQuote
+	var acquiredQuotes []quoteWithInfo
 
-	case *rfqrpc.AddAssetBuyOrderResponse_InvalidQuote:
-		return nil, fmt.Errorf("peer %v sent back an invalid quote, "+
-			"status: %v", r.InvalidQuote.Peer,
-			r.InvalidQuote.Status.String())
+	for peer, channels := range chanMap {
+		if existingQuotes {
```
Roasbeef, line 7861: Odd control flow, I think it makes more sense to just skip the loop all together if we have existing quotes. See comment above re if that should ever even be the case tho.

rpcserver.go
```diff
@@ -7853,63 +7849,88 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
...
+			"invoice")
+	}
+
+	// We need to trim any extra quotes that cannot make it into the bolt11
+	// invoice due to size limitations.
+	if len(acquiredQuotes) > maxRfqHopHints {
+		acquiredQuotes = acquiredQuotes[:maxRfqHopHints]
+	}
+
+	// TODO(george): We want to cancel back quotes that didn't make it into
```
Roasbeef, line 7909: Let's spin this into an issue post merge so it doesn't get lost.

rpcserver.go
Roasbeef: > Something to keep in mind here is that when receiving shards our peer does non-strict forwarding on the LND level. The channel we pick here does not bind the actual link that will be used for the HTLC.

With the recent set of liquidity fixes, IIUC we're much stricter here now (being too lax was itself a bug). 

Looks like actually just the channel ID and pubkey are explicitly needed, and we fetch the policy using one of lnd's RPC calls.

Roasbeef (commented):
Changes are looking really good!

Only blocking factor now IMO is the addition of unit test coverage for the new helper functions that were moved from the rpcserver to the rfq package (which are also expanded in this PR).

rpcserver.go
Roasbeef: Reminder here.

Roasbeef (approved):
LGTM 🦮

Roasbeef:
There's an itest failure. Not sure if it's a flake though, can't view the logs from the UI as they're truncated.

rfq/manager.go
```diff
@@ -1054,6 +1054,190 @@ func (m *Manager) ChannelMatchesFully(ctx context.Context,
...
+	// ReceiveIntention defines the intention to receive over an asset
+	// channel.
+	ReceiveIntention
+)
+
+// RfqChannel returns the channel to use for RFQ operations. If a peer public
+// key is specified, the channels are filtered by that peer. If there are
+// multiple channels for the same specifier, the user must specify the peer
+// public key.
+func (m *Manager) RfqChannel(ctx context.Context,
```
ffranr: `Rfq` in method name here is redundant. We could call this `FetchChannel` for example.

rfq/manager.go
```diff
@@ -1054,6 +1054,190 @@ func (m *Manager) ChannelMatchesFully(ctx context.Context,
...
+			})
+		}
+	}
+
+	return channels, haveGroupedAssetChannels, nil
+}
+
+// chanIntention defines the intention of calling rfqChannel. This helps with
+// returning the channel that is most suitable for what we want to do.
+type chanIntention uint8
```
ffranr: We should export this since `RfqChannel` is exported. Also, `RfqChannel` method name is not exact in doc comment. Note also my comment on naming `RfqChannel`.

rfq/manager.go
```diff
@@ -1054,6 +1054,190 @@ func (m *Manager) ChannelMatchesFully(ctx context.Context,
 	return true, nil
 }
 
+// ChannelWithSpecifier is a helper struct that combines the information of an
+// asset specifier that is satisfied by a channel with the channels' general
+// information.
+type ChannelWithSpecifier struct {
```
ffranr, line 1020: Perhaps we should just call this `TapChannel`?

rfq/manager_test.go
```diff
@@ -0,0 +1,352 @@
...
+	proof1       = proof.Proof{
+		Asset: asset1,
+	}
+	proof2 = proof.Proof{
+		Asset: asset2,
+	}
+	proof3 = proof.Proof{
+		Asset: asset3,
+	}
+	testGroupKey = pubKeyFromUint64(2121)
```
ffranr, line 57: Perhaps we should just set this to a random group key? I think we have functions for generating random group keys. And then we don't need `pubKeyFromUint64`. Although, I do like `pubKeyFromUint64`, random key is better practice for more coverage IMO.

ffranr (commented):
I've added naming suggestions.

ffranr (approved):

PR #1462, opened 2025-04-08
`PaymentBandwidth` uses HtlcView

GeorgeTsagk:
## Description

When `AuxTrafficShaper.PaymentBandwidth` is called asynchonously we could previously report the same bandwidth for 2 different HTLCs that are validating whether they can be added to the channel. This could lead to false-positives and payment failures as the latter HTLC should be checked against the correct bandwidth.

Instead of always consulting the commitment blob on what the current asset balance is, we now use the existing helper `ComputeView` to get a more precise representation of the asset balance, which is enabled by the new argument `lnwallet.HtlcView`.

### DIFF
--- a/tapchannel/aux_traffic_shaper.go
+++ b/tapchannel/aux_traffic_shaper.go
@@ -15,6 +15,7 @@ import (
 	cmsg "github.com/lightninglabs/taproot-assets/tapchannelmsg"
 	lfn "github.com/lightningnetwork/lnd/fn/v2"
 	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnutils"
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/tlv"
@@ -180,10 +181,20 @@ func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
 	// never be settled. Other HTLCs that may also call into this method are
 	// not yet registered to the commitment, so we need to account for them
 	// manually.
-	computedLocal, _, err := ComputeLocalBalance(*commitment, htlcView)
+	computedLocal, decodedView, err := ComputeLocalBalance(
+		*commitment, htlcView,
+	)
 	if err != nil {
 		return 0, err
 	}
+
+	log.Tracef("Computed asset HTLC View: commitmentLocal=%v, "+
+		"computedLocal=%v, nextHeight=%v, thisHtlc=%v, newView=%v",
+		cmsg.OutputSum(commitment.LocalOutputs()), computedLocal,
+		htlcView.NextHeight, htlc.Amounts.Val.Sum(),
+		lnutils.NewLogClosure(func() string {
+			return prettyPrintLocalView(*decodedView)
+		}))
 
 	// If the HTLC carries asset units (keysend, forwarding), then there's
 	// no need to do any RFQ related math. We can directly compare the asset
@@ -436,4 +447,32 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
 	}
 
 	return htlcAmountMSat, updatedRecords, nil
+}
+
+// prettyPrintLocalView returns a string that pretty-prints the local update log
+// of an HTLC view.
+func prettyPrintLocalView(view DecodedView) string {
+	var res string
+	res = "\nHtlcView Local Updates:\n"
+	for _, v := range view.OurUpdates {
+		assetAmt := uint64(0)
+		if rfqmsg.HasAssetHTLCCustomRecords(v.CustomRecords) {
+			assetHtlc, err := rfqmsg.HtlcFromCustomRecords(
+				v.CustomRecords,
+			)
+			if err != nil {
+				res = fmt.Sprintf("%s\terror: could not "+
+					"decode htlc custom records\n", res)
+				continue
+			}
+
+			assetAmt = rfqmsg.Sum(assetHtlc.Balances())
+		}
+
+		res = fmt.Sprintf("%s\thtlcIndex=%v: amt=%v, assets=%v, "+
+			"addHeight=%v\n", res, v.HtlcIndex, v.Amount, assetAmt,
+			v.AddHeight(lntypes.Local))
+	}
+
+	return res
 }


guggero (commented):
Code wise this looks good to me. But IMO we'd want a bit more certainty around whether this really works in all cases.

tapchannel/aux_funding_controller.go
```diff
@@ -549,7 +549,7 @@ func newCommitBlobAndLeaves(pendingFunding *pendingAssetFunding,
 
 	// Just like above, we don't have a real HTLC view here, so we'll pass
 	// in a blank view.
-	fakeView := &lnwallet.HtlcView{}
+	fakeView := lnwallet.AuxHtlcView{}
```
guggero, line 552: nit: since this isn't a pointer to a struct anymore, this could just be `var fakeView lnwallet.AuxHtlcView`.

tapchannel/commitment.go
```diff
@@ -93,19 +94,19 @@ func ComputeView(ourBalance, theirBalance uint64,
 	localHtlcIndex := make(map[uint64]lnwallet.AuxHtlcDescriptor)
 	remoteHtlcIndex := make(map[uint64]lnwallet.AuxHtlcDescriptor)
 
-	for _, entry := range original.AuxOurUpdates() {
+	for _, entry := range original.Updates.Local {
```
guggero, line 97: This change and the new parameter in payment bandwidth should probably be in the same commit as the lnd bump commit, if we want each commit to compile individually (which at least in lnd we want but I think we haven't ported over that CI step to `tapd` yet).

tapchannel/aux_traffic_shaper.go
```diff
@@ -192,13 +193,19 @@ func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
...
 	)
 	if err != nil {
 		return 0, err
 	}
 
+	log.Tracef("Computed asset HTLC View: commitmentLocal=%v, "+
+		"computedLocal=%v, nextHeight=%v, thisHtlc=%v, newView=%v\n",
+		cmsg.OutputSum(commitment.LocalOutputs()), computedLocal,
+		htlcView.NextHeight, htlc.Amounts.Val.Sum(),
+		localViewLogClosure(*decodedView))
```
guggero, line 202: We don't need to pull in `lnutils` here, that was just an example.
Can just be:
```
	log.Tracef("Computed asset HTLC View: commitmentLocal=%v, "+
		"computedLocal=%v, nextHeight=%v, thisHtlc=%v, newView=%v\n",
		cmsg.OutputSum(commitment.LocalOutputs()), computedLocal,
		htlcView.NextHeight, htlc.Amounts.Val.Sum(), func() string {
			return prettyPrintLocalView(*decodedView)
		})
```

guggero: Also, I don't think we need the `\n` at the end here. If the output needs an extra line at the end, that should come from `prettyPrintLocalView()`.

tapchannel/aux_traffic_shaper.go
```diff
@@ -193,8 +188,8 @@ func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
 	// never be settled. Other HTLCs that may also call into this method are
 	// not yet registered to the commitment, so we need to account for them
 	// manually.
-	computedLocal, _, decodedView, _, err := ComputeView(
-		localBalance, remoteBalance, lntypes.Local, htlcView,
+	computedLocal, decodedView, err := ComputeLocalBalance(
```
guggero, line 191: For ease of review (for future PRs), I always find it easier to do any refactoring first, before adding new functionality or changing things. That way the refactor commit can be validated with the old mental model and the reviewer simply needs to assure that nothing changes in the behavior. And then often the diff of the actual changes are smaller in the refactored code, so also easier to digest. And the same lines aren't touched twice.

tapchannel/aux_traffic_shaper.go
```diff
@@ -235,15 +246,26 @@ func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
...
 		return lnwire.NewMSatFromSatoshis(btcutil.MaxSatoshi), nil
 
-	case htlcAssetAmount != 0 && htlcAssetAmount > computedLocal:
+	case htlcAssetAmount > computedLocal:
 		// The asset balance of the channel is simply not enough to
 		// route the asset units, we report 0 bandwidth in order for the
 		// HTLC to fail back.
 		return 0, nil
+
+	default:
```
guggero, line 255: This should not be possible, so it's just confusing to have the `default` case.
In fact, now that we definitely return from this function, we can turn the first `case` into an `if`. Then have the second `case` non-indented as the implicit `else`.

tapchannel/aux_traffic_shaper.go
```diff
@@ -206,24 +201,40 @@ func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
...
+
 	switch {
-	case htlcAssetAmount != 0 && htlcAssetAmount <= computedLocal:
+	case htlcAssetAmount <= computedLocal:
 		// Check if the current link bandwidth can afford sending out
 		// the htlc amount without dipping into the channel reserve. If
 		// it goes below the reserve, we report zero bandwidth as we
 		// cannot push the HTLC amount.
-		if linkBandwidth < htlcAmt {
+		if linkBdwth < htlcAmt {
```
guggero, line 234: nit: let's keep the same param name to reduce the diff even more?

tapchannel/aux_traffic_shaper.go
```diff
@@ -297,6 +319,31 @@ func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
...
+	// as reported by the latest commitment.
+	localBalance := cmsg.OutputSum(commitment.LocalOutputs())
+	remoteBalance := cmsg.OutputSum(commitment.RemoteOutputs())
+
+	// With the help of the latest HtlcView, let's calculate a more precise
+	// local balance. This is useful in order to not forward HTLCs that may
+	// never be settled. Other HTLCs that may also call into this method are
+	// not yet registered to the commitment, so we need to account for them
+	// manually.
+	computedLocal, _, decodedView, _, err := ComputeView(
```
guggero, line 337: This is still missing the explicit setting of `htlcView.NextHeight = 0` and the explanation around that.

guggero (commented):
Nice, I think the code is correct now for our use case.
Now it's just code style and reviewer friendliness that can be improved somewhat.

guggero (approved):
LGTM, thanks!

tapchannel/aux_traffic_shaper.go
```diff
@@ -172,18 +175,52 @@ func (s *AuxTrafficShaper) PaymentBandwidth(htlcBlob,
...
-	// need an RFQ ID as we can just compare the local balance and the
-	// required HTLC amount. If there is no amount set, we need to look up
-	// the RFQ ID in the HTLC blob and use the accepted quote to determine
-	// the amount.
+	log.Tracef("Computed asset HTLC View: commitmentLocal=%v, "+
+		"computedLocal=%v, nextHeight=%v, thisHtlc=%v, newView=%v",
+		cmsg.OutputSum(commitment.LocalOutputs()), computedLocal,
+		htlcView.NextHeight, htlc.Amounts.Val.Sum(), func() string {
+			return prettyPrintLocalView(*decodedView)
+		}())
```
guggero, line 195: Why? This removes the whole benefit of the function callback in the first place...

guggero: Hmm... Very weird. It seems like it's not enough that it's the correct function type, it must be of type `LogClosure` (the reason for which I don't understand, must be some explicit configuration somewhere).
So can just be:
```
lnutils.NewLogClosure(func() string {
			return prettyPrintLocalView(*decodedView)
		})
```

guggero: Ah, of course... Because the `LogClosure` type has a `String()` method:
```
// String invokes the underlying function and returns the result.
func (c LogClosure) String() string {
	return c()
}
```

Makes sense now. Sorry, gave you bad advice before then. I thought being able to pass in a function was a feature of the logging system. Turns out it was actually a clever trick.

tapchannel/aux_traffic_shaper.go
ffranr: It's should return an error if it ever reaches a point it shouldn't.

PR #1464, opened 2025-04-09
universe/supplycommit: create new state machine responsible for maintaining the supply commitment for an asset

Roasbeef:
In this commit, we add a new sub-package to the `universe` package which adds a new state machine responsible for managing the supply commitment for a given grouped asset. 

A supply commitment is a special MS-SMT tree that tracks the total supply of a grouped asset. A supply can be modified by: minting new assets, doing an official burn of some assets, or adding some assets to the ignore proof cache. 

To create a supply commitment (for an asset that opts into this), we'll spend the pre-commitment from all past minting batches that have been confirmed on chain. We'll then stage all the pending updates, to create new sub-trees (ignore, burn mint) for the grouped asset. With those trees created, we'll create the new root supply commitment. 

With the root supply commitment created, we'll make a new transaction and commit to that using the non spendable script leaf that we use elsewhere to commit to asset IDs for a group key. 

During this process, we'll continue to serve the _old_ supply commitments. Only once the supply commitments have been confirmed on disk will we _apply_ all the state transitions on disk. After we apply these state transitions, we'll start to serve the new commitments, and go back to our default state. 

See the last commit's readme for more details.


universe/supplycommit/supply_commit_states.go
```diff
@@ -0,0 +1,553 @@
...
+
+// CommitTickEvent is a special event that is used to trigger an update of the
+// on-chain commitment.
+type CommitTickEvent struct {
+}
+
+// eventSealed is a special method that is used to seal the interface.
+func (c *CommitTickEvent) eventSealed() {}
+
+// UpdatePendingState is the state of the state machine when we have From this
```
ffranr, line 285: unfinished sentence: `when we have From`.

universe/supplycommit/supply_commit_states.go
```diff
@@ -0,0 +1,553 @@
...
+
+// StateMachine is a state machine that handles creating and updating the
+// on-chain universe supply commitment for a given asset.
+//
+// nolint:lll
+type StateMachine = protofsm.StateMachine[Event, *Environment]
+
+// Config is a configuration struct that is used to initialize a new supply
+// commit state machine.
+type Config = protofsm.StateMachineCfg[Event, *Environment]
```
ffranr, line 73: I don't think these are used and they might be the only reason for the `lnd` replace in `go.mod`. I'm not sure.

universe/supplycommit/supply_commit_states.go
```diff
@@ -0,0 +1,553 @@
...
+func (s *SignTxEvent) eventSealed() {}
+
+// CommitTxSignState is the state of the state machine we'll transition to once
+// we've created the tx that commits to the latest supply tree. In this state,
+// we'll sign the transaction and all its inputs, then write our state to disk
+// so we'll be able to recover it on restart.
+//
+// State transitions:
+//   - SignTxEvent -> CommitBroadcastState
+type CommitTxSignState struct {
```
ffranr, line 406: I'm not sure if I understand the state transition in the doc comment. Should this be:
```
// State transitions:
//   - SignTxEvent -> CommitTxSignState
//   - CommitTxSignState -> BroadcastEvent
```

Are we describing to and from states in doc `State transitions`?

universe/supplycommit/supply_commit_states.go
```diff
@@ -0,0 +1,553 @@
...
+func (c *ConfEvent) eventSealed() {}
+
+// CommitBroadcastState is the state of the state machine we'll transitions
+// to once we've signed the transaction. In this state, we'll broadcast the
+// transaction, then wait for a confirmation event.
+//
+// State transitions:
+//   - BroadcastEvent -> CommitBroadcastState
+//   - ConfEvent -> DefaultState
+type CommitBroadcastState struct {
```
ffranr, line 464: I don't understand `ConfEvent -> DefaultState` in the doc here.

Once our latest commitment transaction confirms on chain the state machine could transition to `DefaultState`?

universe/supplycommit/README.md
```diff
@@ -0,0 +1,135 @@
...
+It does *not* handle:
+
+*   Discovering supply update events (this is done by other parts of the system
+    feeding events into the state machine).
+
+*   Serving supply proofs to external requesters (this is handled by the broader
+    Universe server components).
+
+*   Managing commitments for multiple different assets simultaneously (each
+    asset specifier typically gets its own state machine instance).
```
ffranr, line 64: `asset specifier` -> `asset group`

ffranr (commented):
I reviewed this PR to see how I might copy its state‑machine architecture for the verification work.

I saw your note about still adding unit tests and doc comments, so I'm leaving these review comments with that in mind and hoping they're helpful.

go.mod
```diff
@@ -216,3 +216,5 @@ replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate
 
 // Note this is a temproary replace and will be removed when taprpc is tagged.
 replace github.com/lightninglabs/taproot-assets/taprpc => ./taprpc
+
+replace github.com/lightningnetwork/lnd => github.com/roasbeef/lnd v0.2.1-alpha.0.20250505223159-b58451faad16
```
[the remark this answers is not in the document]
Roasbeef: The lnd PRs are here: 
  * https://github.com/lightningnetwork/lnd/pull/9726
  * https://github.com/lightningnetwork/lnd/pull/9725

universe/supplycommit/supply_commit_states.go
Roasbeef: So the state machine isn't instantiated externally yet, once it is these will  be used. 

The lnd fork dep is needed to we can access the block block after confirmation, and also pass that into the state machine.

universe/supplycommit/supply_commit_states.go
Roasbeef: This is meant to indicate that from this state, we can go to the sign state.

universe/supplycommit/supply_commit_states.go
Roasbeef: This is slightly out of date, a new `CommitFinalizeState` state is was transitions us to the default.

go.mod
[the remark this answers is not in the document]
guggero: Can be bumped to `lnd v0.19.1-beta.rc1`.

universe/supplycommit/env.go
```diff
@@ -0,0 +1,407 @@
...
+func (r *RootCommitment) TxOut() (*wire.TxOut, error) {
+	// First, obtain the root hash of the supply tree.
+	supplyRootHash := r.SupplyRoot.NodeHash()
+
+	// We'll create a new unspendable output that contains a commitment to
+	// the root.
+	//
+	// TODO(roasbeef): need other version info here/
+	tapLeaf, err := asset.NewNonSpendableScriptLeaf(
+		asset.PedersenVersion, supplyRootHash[:],
```
guggero: Mostly unrelated to the changes here, but I'm wondering if some time in the future (commitment version 3?) we should also use this construction for the normal asset output commitments, instead of the pseudo script we use today.

universe/supplycommit/env.go
```diff
@@ -0,0 +1,407 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create leaf: %w", err)
+	}
+
+	var taprootKey *btcec.PublicKey
+	if r.OutputKey == nil {
+		tapscriptTree := txscript.AssembleTaprootScriptTree(tapLeaf)
```
guggero, line 192: If this is meant as a caching mechanism, should we then move the root hash and tap leaf calculation inside this block to save some CPU cycles?

universe/supplycommit/supply_commit_transitions.go
```diff
@@ -0,0 +1,804 @@
...
+		case BurnTreeType:
+			burns++
+		case IgnoreTreeType:
+			ignores++
+		}
+	}
+
+	// Format the label string including the asset specifier, root hash, and
+	// counts of each update type.
+	label := fmt.Sprintf(
```
guggero: Nice!

universe/supplycommit/supply_commit_transitions.go
```diff
@@ -0,0 +1,804 @@
...
+	default:
+		return nil, fmt.Errorf("%w: received %T while in %T",
+			ErrInvalidStateTransition, event, d)
+	}
+}
+
+// ProcessEvent processes incoming events for the UpdatesPendingState. From
+// here, we'll either be staging new events, or we'll attempt to start the
+// commitment update process once a new commitment is received.
+func (s *UpdatesPendingState) ProcessEvent(event Event, env *Environment) (
```
guggero: micro nit: use `u` as the receiver name to match what's in `supply_commit_states.go`?

universe/supplycommit/supply_commit_transitions.go
```diff
@@ -0,0 +1,804 @@
...
+
+	switch newEvent := event.(type) {
+	// We've received a new update event, we'll stage this in our local
+	// state, and do a self transition.
+	case SupplyUpdateEvent:
+
+		// We just got a new pending update in addition to the one that
+		// made us transition to this state. This ensures that we'll
+		// remember to process from this state after a restart.
+		ctx := context.Background()
```
guggero: I feel like we should make the context part of the `ProcessEvent` argument list, so we can pass it in instead of needing to create a new one every time.

universe/supplycommit/README.md
```diff
@@ -0,0 +1,135 @@
...
+2.  **Batch Updates:** Collect these supply-altering events over time.
+
+3.  **Commit Periodically:** On a regular trigger (`CommitTickEvent`), create a
+    new cryptographic commitment (an MS-SMT root) representing the current state
+    of the asset's supply. This involves separate sub-trees for mints, burns, and
+    ignores, which are then committed into a single root supply tree.
+
+4.  **Anchor On-Chain:** Construct, fund, sign, and broadcast a Bitcoin
+    transaction that anchors this new supply root commitment into the
+    blockchain. This transaction spends the previous commitment output (if one
```
guggero, line 28: So the way I understand the flow proposed in the current iteration is that we can only spend pre-commitment outputs from minting transactions that have already confirmed?
Which would mean we would either need to block the user from creating two mint transactions for the same group in the same block (to make sure there's a "linear" transaction chain), or allow one commitment transaction to spend multiple pre-commitment outputs.

If I understand this sentence correctly, we're aiming for the second option, is that right?

guggero (commented):
Looks great!

universe/supplycommit/env.go
Roasbeef: Yeah I think that would be a lot cleaner, and we'd get to cut down on the amount of total commitment types we have.

universe/supplycommit/env.go
Roasbeef: SGTM.

universe/supplycommit/supply_commit_transitions.go
Roasbeef: I think there's actually a pending PR open on `lnd` to do this. Agree it makes sense to pass in generically.

Roasbeef:
FWIW this won't build due to the replace until we fix the tags in lnd.

litbot-9000:

ffranr:
This patch fixes the lint and `unit-race` CI errors:

```
Index: universe/supplycommit/supply_commit_test.go
IDEA additional info:
<+>UTF-8
diff --git a/universe/supplycommit/supply_commit_test.go b/universe/supplycommit/supply_commit_test.go
--- a/universe/supplycommit/supply_commit_test.go	(revision bab9a94047f48e54580beaf6d7e1fbf94ed5efc2)
+++ b/universe/supplycommit/supply_commit_test.go	(date 1750254593179)
@@ -292,7 +292,7 @@
 	h.sendEvent(event)
 	h.assertNoStateTransitions()
 
-	require.ErrorIs(h.t, h.mockErrReporter.reportedError, expectedErr)
+	require.ErrorIs(h.t, h.mockErrReporter.GetReportedError(), expectedErr)
 }
 
 // assertAndGetCurrentState fetches the current state from the state machine,
@@ -635,7 +635,7 @@
 		h.assertNoStateTransitions()
 
 		require.ErrorIs(
-			t, h.mockErrReporter.reportedError,
+			t, h.mockErrReporter.GetReportedError(),
 			ErrInvalidStateTransition,
 		)
 	})
@@ -938,7 +938,9 @@
 		h.start()
 		defer h.stopAndAssert()
 
-		h.assertHandlesInvalidEvent(&unknownEvent{}, ErrInvalidStateTransition)
+		h.assertHandlesInvalidEvent(
+			&unknownEvent{}, ErrInvalidStateTransition,
+		)
 	})
 }
 
@@ -988,7 +990,9 @@
 		h.start()
 		defer h.stopAndAssert()
 
-		h.assertHandlesInvalidEvent(&unknownEvent{}, ErrInvalidStateTransition)
+		h.assertHandlesInvalidEvent(
+			&unknownEvent{}, ErrInvalidStateTransition,
+		)
 	})
 }
 
Index: universe/supplycommit/mock.go
IDEA additional info:
<+>UTF-8
diff --git a/universe/supplycommit/mock.go b/universe/supplycommit/mock.go
--- a/universe/supplycommit/mock.go	(revision bab9a94047f48e54580beaf6d7e1fbf94ed5efc2)
+++ b/universe/supplycommit/mock.go	(date 1750254094424)
@@ -2,6 +2,7 @@
 
 import (
 	"context"
+	"sync"
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcutil"
@@ -373,9 +374,18 @@
 type mockErrorReporter struct {
 	mock.Mock
 	reportedError error
+	mu            sync.Mutex
 }
 
 func (m *mockErrorReporter) ReportError(err error) {
-	m.Called(err)
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	m.reportedError = err
+	m.Called(err)
+}
+
+func (m *mockErrorReporter) GetReportedError() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return m.reportedError
 }
```

universe/supplycommit/README.md
ffranr: > If I understand this sentence correctly, we're aiming for the second option, is that right?

Yes, that's how i see it. The supply commitment transaction can spend multiple pre-commitment outputs.

universe/supplycommit/supply_commit_states.go
```diff
@@ -0,0 +1,553 @@
...
+	return n.BurnLeaf.UniverseLeafNode()
+}
+
+// A compile time assertion to ensure that NewBurnEvent implements the
+// SupplyUpdateEvent interface.
+var _ SupplyUpdateEvent = (*NewBurnEvent)(nil)
+
+// NewMintEvent signals that a caller wishes to update the mint portion of the
+// supply tree with a new minted asset.
+type NewMintEvent struct {
```
ffranr: Could drop the `New` prefix from these event type names.

universe/supplycommit/supply_commit_states.go
```diff
@@ -0,0 +1,553 @@
+package supplycommit
```
ffranr: files in `supplycommit` directory start with `supply_commit_`. That prefix is probably redundant?

universe/supplycommit/supply_commit_states.go
```diff
@@ -0,0 +1,553 @@
...
+func (c *CommitFinalizeState) IsTerminal() bool {
+	return true
+}
+
+// String returns the name of the state.
+func (c *CommitFinalizeState) String() string {
+	return "CommitFinalizeState"
+}
+
+// SpendEvent....
```
ffranr: missing doc comment

universe/supplycommit/supply_commit_test.go
```diff
@@ -0,0 +1,1488 @@
...
+		h := newSupplyCommitTestHarness(t, &harnessCfg{
+			initialState: &CommitFinalizeState{
+				SupplyTransition: initialTransition,
+			},
+			assetSpec: defaultAssetSpec,
+		})
+		h.start()
+		defer h.stopAndAssert()
+
+		h.assertHandlesInvalidEvent(&unknownEvent{}, ErrInvalidStateTransition)
```
ffranr: my patch should fix this

universe/supplycommit/supply_commit_transitions.go
```diff
@@ -0,0 +1,817 @@
...
+			burns++
+		case IgnoreTreeType:
+			ignores++
+		}
+	}
+
+	// Format the label string including the asset specifier, root hash, and
+	// counts of each update type.
+	label := fmt.Sprintf(
+		"supply_commit(%v):root=%x:sum=%v:m=%d,b=%d,i=%d",
```
ffranr: add `tapd_` prefix to the label here and above so that it's clear the tx came from tapd? Also, `supply_commit(%v):root` could be in a variable because it's repeated in the `rootNode == nil` if-clause above.

We also might want to use dashes rather than underscores. Note that out transfer tx label is:
```
TransferTxLabel = "tapd-asset-transfer"
```

universe/supplycommit/supply_commit_states.go
Roasbeef: Could be interpreted as implicit, but I prefer the new prefix to make things a bit more explicit.

universe/supplycommit/supply_commit_test.go
Roasbeef: Applied! Thanks.

universe/supplycommit/supply_commit_transitions.go
Roasbeef: Ok I switched to underscores, and added the `tapd-` prefix. 

>  Also, supply_commit(%v):root could be in a variable because it's repeated in the rootNode == nil if-clause above.

There's som eoverlap, but it's slightly different as once uses the `no_root`.

go.mod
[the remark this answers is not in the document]
Roasbeef: I don't think these made it into the minor release, but they are in master.

universe/supplycommit/README.md
Roasbeef: > or allow one commitment transaction to spend multiple pre-commitment outputs.
> If I understand this sentence correctly, we're aiming for the second option, is that right?

Correct. Each time we go to make a new commitment, we'll query for _all_ the unspent pre-commitment outputs to spend in the next commitment transaction.

go.mod
```diff
@@ -30,11 +30,11 @@ require (
 	github.com/lightninglabs/lndclient v0.19.0-7
 	github.com/lightninglabs/neutrino/cache v1.1.2
 	github.com/lightninglabs/taproot-assets/taprpc v1.0.7
-	github.com/lightningnetwork/lnd v0.19.0-beta
+	github.com/lightningnetwork/lnd v0.19.0-beta.rc5.0.20250611041824-9e9524766c8a
```
guggero, line 33: nit: can use `v0.19.1-beta` now.

universe/interface.go
```diff
@@ -247,12 +247,20 @@ type Leaf struct {
 
 	// Amt is the amount of units associated with the coin.
 	Amt uint64
+
+	// IsBurn is a boolean that indicates whether the leaf represents a burn
```
guggero, line 251: Is the `[temp]` commit message suffix still needed?

universe/supplycommit/supply_commit_states.go
guggero: +1

universe/supplycommit/supply_commit_transitions.go
```diff
@@ -0,0 +1,818 @@
...
+			return nil, fmt.Errorf("unable to fund supply "+
+				"commitment tx: %w", err)
+		}
+
+		// With the funded PSBT ready, transition to the signing state,
+		// passing the PSBT and the new commitment via an event. Create
+		// a new SupplyStateTransition for the next state.
+		nextSupplyTransition := SupplyStateTransition{
+			OldCommitment: oldCommitment,
+			//OldSupplyRoots:    c.SupplyTransition.OldSupplyRoots,
```
guggero: nit: remove commented out code? Or is there a TODO to address here still?

universe/supplycommit/supply_commit_transitions.go
```diff
@@ -0,0 +1,818 @@
...
+	env *Environment) (*tapsend.FundedPsbt, error) {
+
+	// Create a new PSBT from the unsigned transaction.
+	commitPkt, err := psbt.NewFromUnsignedTx(supplyCommit.Txn)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create PSBT: %w", err)
+	}
+
+	// Estimate the required fee rate.
+	feeRate, err := env.Chain.EstimateFee(ctx, CommitConfTarget)
```
guggero: The conf target should probably be part of the `Environment` so it can be configured later.

universe/supplycommit/supply_commit_transitions.go
```diff
@@ -0,0 +1,818 @@
...
+		}
+
+		var pkScript []byte
+		if len(commitTx.TxOut) > 0 {
+			pkScript = commitTx.TxOut[0].PkScript
+		}
+		confEvent := &protofsm.RegisterConf[Event]{
+			Txid:     commitTx.TxHash(),
+			PkScript: pkScript,
+			// TODO(roasbeef): need to add a height hint here
```
guggero: Can just call into `env.Chain.CurrentHeight()` and set it here?

go.mod
Roasbeef: https://github.com/lightninglabs/taproot-assets/pull/1464#discussion_r2155690260

universe/interface.go
Roasbeef: I don't think so, but I consider this a sort of hack. Basically we need this to use the full value of the burnt amount instead of just `1` (as we do for transfers). 

It think an alternative here is either to make a new wrapper struct, or modify this logic, but still only conditionally (we still insert just `1` for the multiverse transfer tree.

universe/supplycommit/supply_commit_transitions.go
Roasbeef: SGTM.

go.mod
guggero: Ah, damn, I thought I added them. But then we have to move them to the 0.19.2 milestone to include them there, otherwise we'll have a problem in litd.

universe/supplycommit/env.go
```diff
@@ -0,0 +1,407 @@
...
+	// BlockHeight is the block height of the transaction that contains the
+	// pre-commitment.
+	BlockHeight uint32
+
+	// MintingTxn is the minting transaction itself that created the pre
+	// commitment.
+	MintingTxn *wire.MsgTx
+
+	tapgarden.PreCommitmentOutput
+}
```
ffranr: I don't think we should re-use `tapgarden.PreCommitmentOutput` here. I think we should just copy over the fields into this new `PreCommitment` type.

https://github.com/lightninglabs/taproot-assets/pull/1589 will make the group key field an `Option` which only makes sense during minting (in `tapgarden`). But in `supplycommit` we expect that field to never be none.

Basically, I think the code will be easier to maintain over the long run if `PreCommitmentOutput` from `tapgarden` isn't re-used in `supplycommit`.

ffranr: I think the TODO in the struct doc can be removed? What does the TODO mean?

Roasbeef: Yeah the TODO is out of date, removed.

Roasbeef: Remove the embedded struct, using the fields in line.

universe/supplycommit/supply_commit_states.go
Roasbeef: Done!

PR #1465, opened 2025-04-11
Extract non-controversion commits from Proof-system-v1 PR

guggero:
To make review easier on #1453, I'm extracting a couple of commits that aren't directly related to that PR (or are certain we want them, like the removal of the alt leaves from the inputs).


guggero:
Need to look into the failing itest...

guggero:
Removed commit https://github.com/lightninglabs/taproot-assets/commit/c9b1d54430cdd7c4cdf3e1e65e7a33babc428afd as that only makes sense in the context of #1453. So we'll have to add that again there.

Roasbeef (approved):
LGTM 🦍

ffranr (approved):

PR #1347, opened 2025-02-03
 mssmt: add new batched ms-smt leaf insertion with recursive merge update 

Roasbeef:
# Implement Batched Insertion for MS-SMT to Optimize Database Operations (Fix #451)

 ## Aider Usage

 **All modifications in these files were produced using `aider`** (including this PR description). The changes were generated with aider v0.73.0 using:
 - Main model: **o3-mini** (architect edit format)
 - Editor model: **gpt-4o** (editor-diff edit format)
 - Weak model: **gpt-4o-mini**

 The command executed was similar to:

`aider --model o3-mini --reasoning-effort high --architect`

This ensured that all modifications to the batched insertion algorithm, merge logic, and recursive update functionality were handled consistently across the codebase.

 ## Overview

 This PR introduces a new batched insertion routine for the Merkle-Sum Sparse Merkle Tree (MS-SMT), significantly reducing the number of database transactions by updating the root and internal nodes only once per batch. The new algorithm recursively partitions a sorted batch of leaf insertions by
 examining individual bits of the insertion keys at each tree level. When a subtree is empty and a single leaf is inserted, a compacted leaf is created using `NewCompactedLeafNode`. 

In collision cases—when an existing compacted leaf is encountered with a different key—the algorithm calls the
 `merge` function to combine the new and existing leaves into a newly reconstructed subtree. For multiple inbound insertions, the algorithm recursively processes left and right partitions and finally creates a branch node to combine the updated children.

 ## Recursive Merge & Time Complexity

 - **Recursive Partitioning:**
   The routine processes the tree level by level (from height 0 to last bit index) and partitions the sorted batch based on the bit value (0 or 1) at that level. This recursive process ensures that the insertion position is determined in O(log U) time per leaf (where U is the number of bits in th
 key, ~256).

 - **Merge Optimization:**
   Instead of performing N separate insertions (each incurring O(log U) database operations), the new batched algorithm processes the entire batch within one transaction. Although it still performs O(log U) work per leaf internally, the overall database overhead is reduced from `O(N log U)` to `O(log
 U) + O(1)` per batch, greatly improving efficiency for large batches.

 ## Conclusion

 This change improves the efficiency of bulk insertions in the MS-SMT by reducing database write transactions and by applying a robust recursive merge/update algorithm. The new implementation replaces multiple single insert calls with one batched update, leading to significant performance gains in
 high-load scenarios. 

This PR fixes issue [#451](https://github.com/lightninglabs/taproot-assets/issues/451).

PR #1467, opened 2025-04-12
mssmt: add new Copy method and InsertMany to optimize slightly 

Roasbeef:
In this commit, we add a new `Copy` method that takes as an arg another `mssmt.Tree` and inserts all the keys+leaves from the target tree into the source tree. We then optimize a bit by adding an `InsertMany` method that will call `insert` N times in the same `TreeStoreUpdateTx`. This doesn't yet implement any fancy optimizations to ensure the root is only updated once as https://github.com/lightninglabs/taproot-assets/pull/1347 does. 

This was spun off a work unit related to https://github.com/lightninglabs/taproot-assets/pull/1464.

### DIFF
--- a/mssmt/compacted_tree.go
+++ b/mssmt/compacted_tree.go
@@ -510,15 +510,73 @@ func (t *CompactedTree) Copy(ctx context.Context, targetTree Tree) error {
 		return err
 	}
 
-	// Insert all found leaves into the target tree.
-	for key, leaf := range leaves {
-		// Use the target tree's Insert method.
-		_, err := targetTree.Insert(ctx, key, leaf)
-		if err != nil {
-			return fmt.Errorf("error inserting leaf with key %x "+
-				"into target tree: %w", key, err)
-		}
+	// Insert all found leaves into the target tree using InsertMany for
+	// efficiency.
+	_, err = targetTree.InsertMany(ctx, leaves)
+	if err != nil {
+		return fmt.Errorf("error inserting leaves into "+
+			"target tree: %w", err)
 	}
 
 	return nil
+}
+
+// InsertMany inserts multiple leaf nodes provided in the leaves map within a
+// single database transaction.
+func (t *CompactedTree) InsertMany(ctx context.Context,
+	leaves map[[hashSize]byte]*LeafNode) (Tree, error) {
+
+	if len(leaves) == 0 {
+		return t, nil
+	}
+
+	dbErr := t.store.Update(ctx, func(tx TreeStoreUpdateTx) error {
+		currentRoot, err := tx.RootNode()
+		if err != nil {
+			return err
+		}
+		rootBranch := currentRoot.(*BranchNode)
+
+		for key, leaf := range leaves {
+			// Check for potential sum overflow before each
+			// insertion.
+			sumRoot := rootBranch.NodeSum()
+			sumLeaf := leaf.NodeSum()
+			err = CheckSumOverflowUint64(sumRoot, sumLeaf)
+			if err != nil {
+				return fmt.Errorf("compact tree leaf insert "+
+					"sum overflow, root: %d, leaf: %d; %w",
+					sumRoot, sumLeaf, err)
+			}
+
+			// Insert the leaf using the internal helper.
+			newRoot, err := t.insert(
+				tx, &key, 0, rootBranch, leaf,
+			)
+			if err != nil {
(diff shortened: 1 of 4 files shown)


mssmt/compacted_tree.go
```diff
@@ -392,3 +392,166 @@ func (t *CompactedTree) MerkleProof(ctx context.Context, key [hashSize]byte) (
 
 	return NewProof(proof), nil
 }
+
+// collectLeaves recursively traverses the compacted tree represented by the
+// given node and collects all non-empty leaf nodes along with their keys.
+func collectLeaves(ctx context.Context, tx TreeStoreViewTx, node Node) (
```
ffranr, line 398: I don't think this function is used. Is it included intentionally?

mssmt/compacted_tree.go
```diff
@@ -392,3 +392,166 @@ func (t *CompactedTree) MerkleProof(ctx context.Context, key [hashSize]byte) (
...
+	}
+
+	// Check against EmptyTree branches requires depth. If we encounter an
+	// unexpected node that isn't identifiable as empty, return error.
+	return nil, fmt.Errorf("unexpected node type %T encountered "+
+		"during leaf collection", node)
+}
+
+// collectLeavesRecursive is the helper for collectLeaves that includes depth.
+func collectLeavesRecursive(ctx context.Context, tx TreeStoreViewTx, node Node,
```
ffranr, line 432: I think this function used to be a helper for `collectLeaves`, but now it is the go-to. So I think doc is stale.

mssmt/compacted_tree.go
```diff
@@ -392,3 +392,166 @@ func (t *CompactedTree) MerkleProof(ctx context.Context, key [hashSize]byte) (
...
+		return make(map[[hashSize]byte]*LeafNode), nil
+	}
+
+	return nil, fmt.Errorf("unexpected node type %T encountered "+
+		"during leaf collection at depth %d", node, depth)
+}
+
+// Copy copies all the key-value pairs from the source tree into the target
+// tree. The target tree is assumed to be empty.
+func (t *CompactedTree) Copy(ctx context.Context, targetTree Tree) error {
```
ffranr, line 519: > The target tree is assumed to be empty.

I don't think the target tree has to be empty. So maybe we can remove this from the doc. Otherwise, if it has to be empty, then I think we should assert that in code in addition to having this note in the doc.

mssmt/tree.go
```diff
@@ -333,6 +333,112 @@ func (t *FullTree) MerkleProof(ctx context.Context, key [hashSize]byte) (
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Recursively find leaves in the right subtree. We need to set
+		// the bit corresponding to the current depth in the key prefix.
+		rightKeyPrefix := keyPrefix
+		byteIndex := depth / 8
+		bitIndex := depth % 8
+		rightKeyPrefix[byteIndex] |= (1 << bitIndex)
```
ffranr, line 378: I think that this should be spun into its own function so that we can add a bit more docs to explain what's happening here. I'm surprised that this function doesn't exist already. It's repeated in `PackBits` for example.

mssmt/tree_test.go
```diff
@@ -822,6 +822,103 @@ func TestBIPTestVectors(t *testing.T) {
 	}
 }
 
+// TestTreeCopy tests the Copy method for both FullTree and CompactedTree,
+// including copying between different tree types.
+func TestTreeCopy(t *testing.T) {
```
ffranr, line 827: All these test cases assume that the target tree is empty. We might want to assert that in the copy routine or add test coverage for non-empty.

ffranr (commented):
Looks good to me!

Some of the comments are missing punctuation, but no big deal.

mssmt/compacted_tree.go
Roasbeef: Yep not needed, can remove.

mssmt/compacted_tree.go
Roasbeef: There's no need for it to be empty. Will update the `godoc` here, and also add additional test cases of non-empty roots.

mssmt/tree_test.go
Roasbeef: Expanded to include a non-empty target tree.

ffranr (approved):
LGTM :+1:

Issue #1466, opened 2025-04-11
[feature]: Support optional labels for asset send transfers


ffranr:
We should add support for assigning a string label to each asset send transfer. This label would be stored locally in the `tapd` DB.

### Motivation

Currently, we subscribe to send events using the recipient’s script key. This suboptimal—there may be multiple recipients, and script keys may be reused or ambiguous. A label provides more general means to track and reference a specific transfer.

### Proposal

- Add an optional `label` field to asset send requests.
- Persist the label in the local DB alongside the transfer state.
- Allow event subscriptions to target transfers by label.
- Users can optionally use the label field to attach human-readable context to the transfer (i.e. a note).

### Use Case

The primary goal is to enable filtering send events based on transfer labels instead of script keys, allowing event subscriptions to be established from the very start of the asset send process—unlike the current setup, which only allows this after coin selection.

PR #1468, opened 2025-04-14
Add Transfer Label Support for Asset Sends

ffranr:
Closes: https://github.com/lightninglabs/taproot-assets/issues/1466

This PR introduces support for assigning and filtering transfer labels when sending assets. The label provides a way to track and correlate asset transfers across different subsystems.

**Summary of Changes:**
- **`taprpc`**:  
  - Added optional `label` field to `SendAssetRequest`.  
  - Updated `SubscribeSendEvents` to support filtering by transfer label.
- **`tapfreighter`**:  
  - Added label support to `Package`, `Parcel`, and `AssetSendEvent`.
- **`tapdb`**:  
  - Enabled reading and writing of transfer labels in the database.
- **`itest`**:  
  - Extended integration tests to validate filtering of send events using labels.


tapdb/sqlc/migrations/000031_asset_transfer_label.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Add a column `label` to table `asset_transfer`.
+ALTER TABLE asset_transfers ADD COLUMN label VARCHAR DEFAULT '';
```
guggero, line 2: Should this be `DEFAULT NULL` instead? Since `sqlStr()` will turn an empty string into a `NULL` value.

itest/utils.go
```diff
@@ -61,6 +62,51 @@ type ClientEventStream[T any] interface {
...
 	ClientEventStream[T]
 	Cancel context.CancelFunc
+
+	// Filter is an optional filter function that can be used to filter
+	// events received from the client stream.
+	//
+	// If set, it will be called for each event received from the stream. If
+	// it returns true, the event is returned. If it returns false, the
+	// event is ignored and the next event is received from the stream.
+	Filter func(T) (bool, error)
```
guggero, line 72: Same here re naming. Should we call this `ShouldNotify`? Or whatever we choose to call the function(s) in `rpcserver`.

taprpc/taprootassets.proto
```diff
@@ -731,6 +731,11 @@ message AssetTransfer {
     // transaction. If the anchor transaction is still unconfirmed, this value
     // will be 0.
     uint32 anchor_tx_block_height = 8;
+
+    // An optional short label for the transfer. This label can be used to track
+    // the progress of the transfer via the logs or an event subscription.
+    // Multiple transfers can share the same label.
+    string label = 9;
```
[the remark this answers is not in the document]
ffranr: I think I'd prefer to handle that in a future PR. Perhaps instead of exposing the transfer's integer primary key, we could add a UUIDv4. That way we leave open the possibility of exporting and merging transfer records without collision, and also of pruning and inserting new transfers safely.

Transfer output position is fine as an int IMO because it's nested under the transfer ID.

guggero:
Needs a rebase due to previously merged PR.

guggero:
Needs a rebase. Will merge directly to bypass queue afterward.

PR #1470, opened 2025-04-15
Add preliminary proof courier connection check to ChainPorter

ffranr:
Depends on https://github.com/lightninglabs/taproot-assets/pull/1468
Will be in draft until that PR is merged. However, this PR is ready to review now.

---

This PR introduces an early-stage connection check to the asset send process:

- **ChainPorter update:** Adds `SendStateStartHandleAddrParcel` as the new initial state for address parcels. In this state, proof courier connectivity is verified to prevent unnecessary coin selection or transaction broadcasting if delivery is likely to fail.
- **Flag support:** Adds CLI and RPC plumbing for a flag to skip the courier ping check (logic not yet wired).
- **itest adjustments:** Disables the new check in select integration test cases where the courier is kept offline until anchor broadcast.

These changes improve failure predictability and test flexibility in proof transfer workflows.


cmd/commands/assets.go
```diff
@@ -233,6 +234,18 @@ func parseFeeRate(ctx *cli.Context) (uint32, error) {
 	return uint32(0), nil
 }
 
+// parseSkipProofCourierPingCheck parses the skip proof courier ping check flag
+// from the CLI context.
+func parseSkipProofCourierPingCheck(ctx *cli.Context) bool {
+	if ctx.IsSet(skipProofCourierPingCheckName) {
```
guggero, line 240: nit: since a boolean flag can only either be set or not, I don't think we need this helper function.
Can just use:
```
	resp, err := client.SendAsset(ctxc, &taprpc.SendAssetRequest{
		TapAddrs:                  addrs,
		FeeRate:                   feeRate,
		SkipProofCourierPingCheck: ctx.Bool(
			skipProofCourierPingCheckName,
		),
	})
```
below.

rpcserver.go
```diff
@@ -3324,7 +3324,8 @@ func (r *rpcServer) SendAsset(ctx context.Context,
 	}
 
 	resp, err := r.cfg.ChainPorter.RequestShipment(
-		tapfreighter.NewAddressParcel(feeRate, req.Label, tapAddrs...),
+		tapfreighter.NewAddressParcel(feeRate, req.Label,
```
guggero, line 3327: nit: formatting.

itest/addrs_test.go
```diff
@@ -931,6 +932,14 @@ func withReceiverAddresses(addrs ...*taprpc.Addr) sendOption {
 	}
 }
 
+// withSkipProofCourierPingCheck is an option to skip the proof courier ping
+// check. This is useful for testing purposes.
+func withSkipProofCourierPingCheck(skipCheck bool) sendOption {
```
guggero, line 937: nit: since the default value of a bool is `false`, I think we don't need a parameter here and can just set this to `true`. Since I think we'd never want to turn it on and later off again?

guggero (approved):
Very nice! Just a couple of simplification suggestions, otherwise LGTM :tada:

ffranr:
I will rebase once `send-asset-label` is merged.

PR #1474, opened 2025-04-17
Add method `Proof.Bytes`

ffranr:
Add method `Proof.Bytes` and then use in many places. Removes redundant buffer variable and two redundant standalone functions.

### DIFF
--- a/proof/proof.go
+++ b/proof/proof.go
@@ -460,6 +460,16 @@ func (p *Proof) Decode(r io.Reader) error {
 	return nil
 }
 
+// Bytes returns the serialized proof.
+func (p *Proof) Bytes() ([]byte, error) {
+	var buf bytes.Buffer
+	err := p.Encode(&buf)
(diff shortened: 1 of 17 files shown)


proof/proof.go
```diff
@@ -476,12 +476,11 @@ func (p *Proof) Bytes() ([]byte, error) {
 // NOTE: This is part of the tlv.RecordProducer interface.
 func (p *Proof) Record() tlv.Record {
 	sizeFunc := func() uint64 {
-		var buf bytes.Buffer
-		err := p.Encode(&buf)
+		pBytes, err := p.Bytes()
```
guggero, line 479: nit: use `proofBytes` instead of `pBytes`, just to unify naming a bit? Here and two more places in the diff below (`proof/proof_test.go` and `tappsbt/encode.go`).

guggero (approved):
Very nice, thanks a lot!

PR #1475, opened 2025-04-17
taprpc: support custom lock ID and expiration in CommitVirtualPsbts

bhandras:
This change extends the CommitVirtualPsbts RPC to accept an optional custom UTXO lock ID and lease expiration. By enabling callers to define these parameters explicitly, the wallet can better coordinate UTXO locking across concurrent workflows and provide safer, more flexible transaction retargeting during PSBT construction.

If unset, the default internal lock ID and 10-minute expiration remain in effect, preserving backward compatibility.

depends on: https://github.com/lightningnetwork/lnd/pull/9724


bhandras:
Build is failing as it is pending updates from https://github.com/lightninglabs/taproot-assets/pull/1478
Please do not review yet.

guggero (approved):
LGTM :tada:

Issue #1385, opened 2025-02-13
[feature]: add a function, `rfqmath.FixedPointFromUint64_Decimal_Display`


ZZiigguurraatt:
There is a lot of confusion when running a price oracle server on how to properly input the exchange rate. One example of this is here: https://lightningcommunity.slack.com/archives/C03B3556HQ8/p1738866875640909?thread_ts=1738748330.746709&cid=C03B3556HQ8 .

The basic-price-oracle example uses the `rfqmath.FixedPointFromUint64` function, and that is where a lot of the confusions arises.

https://github.com/lightninglabs/taproot-assets/blob/28c4ea8e019bb5a52e59401f66b0a5eebf0a0eee/docs/examples/basic-price-oracle/main.go#L111-L126

https://github.com/lightninglabs/taproot-assets/blob/28c4ea8e019bb5a52e59401f66b0a5eebf0a0eee/rfqmath/fixed_point.go#L189-L199

What is very confusing to new users is the "scale". It is recommended that this be the same as the decimal display that was defined when creating the asset. This leads users to think that the "coefficient" be defined in terms of decimal display asset units and then the exchange rate can be off by an order of magnitude of the decimal display. The "scale" parameter is recommended to be the same as the decimal display in order to reduce arithmetic round off error, but it doesn't really change the exchange rate itself.

I'd recommend a new function be created, `rfqmath.FixedPointFromUint64_Decimal_Display`, where the the exchange rate in terms of decimal display assets can be input, and then the decimal display digits as the second input. Then, convert the exchange rate from decimal display asset units to base asset units and use that as the `coefficient` input to `rfqmath.FixedPointFromUint64`. Then, also use the decimal display digits as the `scale` input to `rfqmath.FixedPointFromUint64`.

See also, the definition of FixedPoint: https://github.com/lightninglabs/taproot-assets/blob/3a0556c25d5c196cd7abd690d4cbb1c91369a4f9/taprpc/priceoraclerpc/price_oracle.proto#L26-L57 .

kallerosenbaum:
Thanks for looking into this. What would have helped me most would be better examples in the price-oracle-example code. So I suggest to keep the code as is but improve the comments. For example:
```
 	// As an example, the purchase rate is $42,000 per BTC.
        // The returned value from this function must represent the
        // number of base asset units per BTC. With a decimal display
        // of 6, we should return the amount 42,000,000,000.
        // But we need to return the number as a FixedPoint, which
[...]

PR #1479, opened 2025-04-17
basic-price-oracle: clarify price oracle demo asset rate documentation

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/1385

Add explanation of the difference between real-world prices and tap asset rates, and relate to asset decimal display. Also include a FixedPoint representation example to improve clarity.

### DIFF
--- a/docs/examples/basic-price-oracle/main.go
+++ b/docs/examples/basic-price-oracle/main.go
@@ -101,6 +101,73 @@ func isSupportedSubjectAsset(subjectAsset *oraclerpc.AssetSpecifier) bool {
 	return false
 }
 
+// getPurchaseRate returns the buy (purchase) rate for the asset. The unit of
+// the rate is the number of TAP asset units per BTC.
+//
+// Suppose our TAP asset is a USD stablecoin. To support liquidity and precision
+// in rate conversion, we mint 1,000,000 TAP asset units per 1.00 USD. Wallet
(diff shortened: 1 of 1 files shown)


docs/examples/basic-price-oracle/main.go
```diff
@@ -101,6 +101,69 @@ func isSupportedSubjectAsset(subjectAsset *oraclerpc.AssetSpecifier) bool {
...
+// asset units of our valuable asset.
+//
+// To maintain integer-based communication, representing such small fractional
+// rates directly is impractical due to floating-point precision issues.
+// Instead, the price oracle returns a FixedPoint representation with integer
+// components.
+//
+// A FixedPoint number `F` consists of a coefficient `C` and scale exponent `s`:
+//
+// F = C * (10^-s)
```
guggero, line 146: This confused me for a bit, since the actual calculation we do is: $$F = \frac {C} {10^s}$$
Which is of course equivalent to $$F = C * 10^{-s}$$
Perhaps we can write both here?

guggero (approved):
Very nice explanation, LGTM :tada:

docs/examples/basic-price-oracle/main.go
timkaebisch: I agree, it also took me a while. Writing both equations definitely helps make it clearer.

PR #1480, opened 2025-04-19
Add chain_anchor with block_timestamp to AssetLeaves RPC

### LINKED ISSUE #1477: [bug] `ChainAnchor` field missing in `AssetLeaves` RPC response
The `ChainAnchor` field in the `AssetLeaves` RPC response is currently unpopulated, but it should be.

To fix this, construct a chain asset from the corresponding leaf proof and marshal it using existing logic to populate `ChainAnchor`.

Note: it may not be possible to construct chain assets from all proof types. We should at least support this for issuance proofs.

jamaljsr:
Closes #1477 

We would like to update Terminal to display all of the asset mints/burns for a group key. We also want to display the txid and timestamp of the mint txn. I'm using the universe `AssetLeaves` RPC to get the list of issued assets. This is when I realized that the `chain_anchor` field was not being set in this response.

I have taken a stab at implementing the fix to include the `chain_anchor` field in the `AssetLeaves` RPC response. I updated  `marshalAssetLeaf` to call `MarshalChainAsset` which includes the chain anchor info. I also added a new `block_timestamp` field to the RPC message, which gets set from the block header which is encoded in the proof.

Note: This is my first contribution to this repo and would appreciate any guidance on whether the approach I've taken here is good. There may be better ways to get the desired result which I have not discovered yet.


rpcserver.go
```diff
@@ -1163,9 +1163,26 @@ func (r *rpcServer) MarshalChainAsset(ctx context.Context, a asset.ChainAsset,
...
+	if rpcAsset.ChainAnchor != nil &&
+		rpcAsset.ChainAnchor.BlockHeight != 0 &&
+		rpcAsset.ChainAnchor.BlockTimestamp == 0 {
+
+		ts := r.cfg.ChainBridge.GetBlockTimestamp(
+			ctx,
+			rpcAsset.ChainAnchor.BlockHeight,
+		)
+		rpcAsset.ChainAnchor.BlockTimestamp = ts
+	}
```
jamaljsr, line 1183: This block doesn't feel quite right here, but I added it because I saw that the `block_timestamp` field was `0` when running `tapcli assets list`. I didn't see how to grab the timestamp from within the assets store [dbAssetsToChainAssets](https://github.com/jamaljsr/taproot-assets/blob/leaves-chain-anchor/tapdb/assets_store.go#L829-L841) function. I didn't feel like it would be appropriate to somehow call `ChainBridge.GetBlockTimestamp()` from the db level. Please let me know if there's a better way to do this.

Roasbeef: > I didn't see how to grab the timestamp from within the assets store dbAssetsToChainAssets function since it isn't stored in the D

Yeah the timestamp isn't stored in the DB, as we only store the transaction and not _also_ the header. 

However the header is stored in the `proof.Proof` struct, which is a state transition proof (new txn properly followed all the rules). Looks like you start to decode the proof below here in the diff, so you should be able to get the header, then the timestamp from that.

Roasbeef: https://github.com/lightninglabs/taproot-assets/blob/db28a03c23ed908dd7cb7ae5e411e921cdc421c0/proof/proof.go#L243-L246

Roasbeef: https://pkg.go.dev/github.com/btcsuite/btcd/wire#BlockHeader

jamaljsr: Nice! I was able to fetch the proof and grab the timestamp from the block header. Thanks for the pointer. This implementation feels much better.

jamaljsr: Eww! Looks like this change broke some of the tests. Going to try to resolve those issues.

rpcserver.go
```diff
@@ -5218,8 +5218,18 @@ func marshalAssetLeaf(ctx context.Context, keys taprpc.KeyLookup,
...
+	// Decode the single proof to extract on-chain anchor info.
+	p, err := proof.Decode(assetLeaf.RawProof)
+	if err != nil {
+		return nil, err
+	}
+	chainAsset, err := p.ToChainAsset()
+	if err != nil {
+		return nil, err
+	}
+	// Marshal as a chain asset to include chain_anchor details.
```
guggero, line 5230: nit: add empty line before new comment.

tapdb/assets_store.go
```diff
@@ -826,6 +827,24 @@ func (a *AssetStore) dbAssetsToChainAssets(dbAssets []ConfirmedAsset,
...
+			AssetID:   &assetID,
+			ScriptKey: *scriptKeyPub,
+			OutPoint:  &anchorOutpoint,
+		}
+		blob, err := a.FetchProof(ctx, locator)
+		if err != nil {
+			return nil, fmt.Errorf("unable to fetch proof for "+
+				"asset %s: %w", assetID, err)
+		}
+		p, err := blob.AsSingleProof()
```
guggero, line 842: This should never work, as the proofs in the database are always proof files.

And anyway, this feels a bit much to do for each asset. I know Laolu mentioned this approach.
But my counter suggestion would be to add this block to `rpcServer.MarshalChainAsset()`:
```
	if a.AnchorBlockTimestamp == 0 && a.AnchorBlockHeight > 0 {
		a.AnchorBlockTimestamp = r.cfg.ChainBridge.GetBlockTimestamp(
			ctx, a.AnchorBlockHeight,
		)
	}
```

I know it's a bit of a hack. But it fixes the one call (from `fetchRpcAssets()`) that doesn't use `someProof.ToChainAsset()` (which sets the timestamp from the block header in the proof).

guggero (commented):
Very useful feature. See inline comment for a bit more light-weight approach suggestion that should also fix the tests.

guggero (approved):
Nice, LGTM :tada:

ffranr (approved):
Nice!

PR #1487, opened 2025-04-24
taprpc: extract as standalone Go submodule for WASM clients

ffranr:
Reformulate `taprpc` as a standalone Go submodule. This allows WASM clients to depend on a smaller, more focused module, rather than importing the entire `taproot-assets` package.

Related issue: https://github.com/lightninglabs/lightning-node-connect/issues/113

Overall plan: https://github.com/lightninglabs/lightning-node-connect/issues/113#issuecomment-2832804329


ffranr:
This is a minimal replacement of https://github.com/lightninglabs/taproot-assets/pull/1090

ffranr:
Overall plan: https://github.com/lightninglabs/lightning-node-connect/issues/113#issuecomment-2832804329

server.go
```diff
@@ -133,7 +132,7 @@ func (s *Server) initialize(interceptorChain *rpcperms.InterceptorChain) error {
 				Checkers: []macaroons.Checker{
 					macaroons.IPLockChecker,
 				},
-				RequiredPerms: perms.RequiredPermissions,
+				RequiredPerms: taprpc.RequiredPermissions,
```
Roasbeef, line 135: Have tried to get things building with _just_ the move to have `perms` be a sub-module? 

If for whatever reason that isn't sufficient, then I think in order to adopt this RPC sub-module, we'd want to relax some rules we have re replace directives during development. Otherwise every RPC related change will need a PR for updating the protos, then another one for using the new set of protos.

cmd/commands/addrs.go
```diff
@@ -82,7 +83,7 @@ func newAddr(ctx *cli.Context) error {
 	client, cleanUp := getClient(ctx)
 	defer cleanUp()
 
-	assetVersion, err := taprpc.MarshalAssetVersion(
+	assetVersion, err := rpcutils.MarshalAssetVersion(
```
guggero, line 86: While this approach definitely works, I'm not sure if having a single `rpcutils` package is a very good idea. Because it will contain code that references many different packages, it will be very easy to create more circular dependencies later on.

What we instead should do in general, to avoid these "centralized" (from a dependency PoV) packages is to create a `marshal.go` file in each package and place these functions there.
And the functions would always go into the package that declares the non-RPC type of the marshal/unmarshal functions. So in this example it would be `asset`, because `asset.Version` is declared there.
Then we can also shorten the type a bit, as it can just be `asset.MarshalVersion`.

I know it's a bit more work but I'm sure we won't regret it in the future.

docs/examples/basic-price-oracle/go.mod
```diff
@@ -20,7 +20,10 @@ require (
 	dario.cat/mergo v1.0.1 // indirect
 	github.com/Azure/go-ansiterm v0.0.0-20230124172434-306776ec8161 // indirect
 	github.com/Microsoft/go-winio v0.6.1 // indirect
+	github.com/NebulousLabs/fastrand v0.0.0-20181203155948-6fb6489aac4e // indirect
```
guggero, line 25: While we're at it: can we please replace the `github.com/lightninglabs/taproot-assets => ../../../` at the top with a fixed version so we don't have to update these two files with almost every PR?
I know it risks us forgetting to update it if we ever actually change anything that impacts the example. But the number of force pushes and CI cycles that can be avoided by forgetting to update these two files IMO is worth it.

rpcutils/price_oracle_marshal_test.go
```diff
@@ -1,17 +1,18 @@
-package priceoraclerpc
+package rpcutils
```
guggero, line 1: Same as my previous comment, this would IMO better fit in the `oracle` package.

rpcutils/rfq_marshal.go
```diff
@@ -1,14 +1,17 @@
...
 	"math/big"
 
 	"github.com/lightninglabs/taproot-assets/rfqmath"
+	"github.com/lightninglabs/taproot-assets/taprpc/rfqrpc"
 )
 
-// UnmarshalFixedPoint converts an RPC FixedPoint to a BigIntFixedPoint.
-func UnmarshalFixedPoint(fp *FixedPoint) (*rfqmath.BigIntFixedPoint, error) {
+// UnmarshalRfqFixedPoint converts an RPC FixedPoint to a BigIntFixedPoint.
+func UnmarshalRfqFixedPoint(fp *rfqrpc.FixedPoint) (*rfqmath.BigIntFixedPoint,
```
guggero, line 12: Looking at my previously proposed rule "put these functions into the package that defines the non-RPC type" should perhaps be a soft rule or "rule of thumb". Don't think placing this in the `rfqmath` package makes that much sense. But IMO it definitely makes sense in the `rfq` package.

taprpc/go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/lightningnetwork/lnd v0.18.0-beta.rc4.0.20240723043204-f09d4042aee4
 	google.golang.org/grpc v1.59.0
 	google.golang.org/protobuf v1.33.0
+	gopkg.in/macaroon-bakery.v2 v2.0.1
```
guggero, line 12: Same here, let's not downgrade things:

```
	gopkg.in/macaroon-bakery.v2 v2.1.0
	gopkg.in/macaroon.v2 v2.1.0
```

guggero (commented):
Thanks for looking into this!
Looks pretty good, but have a couple of suggestions.

cmd/commands/addrs.go
Roasbeef: One tradeoff with that approach is that then these packages that are mainly "pure" protocol biz logic (eg the group key verification logic in the `asset` package) will start to depend on the RPC sub-modules.

Roasbeef: Re circular deps, as long as we only import the package from the `main` package (rpc server logic, just encoding/decoding before we pass of into routines that live elsewhere), then I think we should be able to minimize the potential for circular deps. 

However if packages like say `rfq` do end up importing these new routines, then agree that it would be pretty easy to introduce one.

taprpc/mintrpc/mint.swagger.json
```diff
@@ -663,11 +659,14 @@
     "protobufAny": {
       "type": "object",
       "properties": {
-        "@type": {
+        "type_url": {
```
Roasbeef, line 662: This shifted due to a version bump?

Roasbeef (commented):
Overall concept ACK. 

No strong feelings on if we go with the `rpcutil` package vs putting RPC serialization routines in select packages. The main pitfall to avoid as mentioned is taking care to not introduce circular dep cycles. 

If `main` is the only package that imports this new one, then we should be ok. However if any other package imports it, then it'll likely be easy to run into a circular dep.

ffranr:
@guggero re `rpcutils`: We still need somewhere to store functions such as:
```
func MarshalKeyDescriptor(desc keychain.KeyDescriptor) *taprpc.KeyDescriptor
```

Would you mind if I handle rpc marshal functionality placement in a separate PR? I just want to carve out `taprpc` in this PR if possible.

docs/examples/basic-price-oracle/go.mod
ffranr: I've pinned to the latest `main` commit of taproot-assets for now because that was the easiest to build. Once the next release is out, I will pin to that. See commit message for more. Let me know if there's a better strategy.

taprpc/go.mod
ffranr: Will fix.

ffranr:
`rpcserver.go` can import anything, nothing depends on it. If we treated `rpcutils` similarly, then maybe we could move some marshalling functions from `rpcserver.go` to `rpcutils`. For example:

```
marshalAssetLeaf
marshalAssetSpecifier
marshalBatchState
marshalLeafKey
marshallReceiveAssetEvent
marshallRfqEvent
marshallSendAssetEvent
marshalMintingBatch
marshalMssmtNode
marshalMssmtProof
marshalOutboundParcel
marshalOutputProofDeliveryStatus
marshalOutputType
marshalPeerAcceptedBuyQuotes
marshalPeerAcceptedSellQuotes
marshalRpcBurn
marshalSeedling
marshalSeedlings
marshalSendEvent
marshalSprouts
MarshalUniID
MarshalUniProofType
marshalUniverseRoot
marshalUniverseServer
marshalUnsealedSeedling
marshalUnsealedSeedlings
marshalVerboseBatch
```

Which might be useful for outside developers. And also slim down `rpcserver.go`.

cmd/commands/addrs.go
guggero: Good point re depending on gRPC stuff from "pure" protocol logic packages. Which would make things hard if/when we decide to expose that logic as WASM ourselves.

guggero (approved):
Nice, LGTM :tada: (pending green CI which can be achieved through rebasing).

taprpc/mintrpc/mint.swagger.json
ffranr: Should be fixed now in my latest commits. I've checked to make sure that we require the same dep versions in taprpc and the main go.mod

ffranr:
Rebasing...

Roasbeef (approved):
LGTM 🥦

Only comment is if we want to move over this marshal code into the new package created: https://github.com/lightninglabs/taproot-assets/blob/1741fd4a0e9ad57b7cbad33cc4235fb795f221db/taprpc/universerpc/marshal.go

ffranr:
> LGTM 🥦
> 
> Only comment is if we want to move over this marshal code into the new package created: https://github.com/lightninglabs/taproot-assets/blob/1741fd4a0e9ad57b7cbad33cc4235fb795f221db/taprpc/universerpc/marshal.go

I've moved those marshalling functions into `rpcutils` also, to be consistent.

guggero:
I pushed up a fix to the module creation commit and bumped some dependency versions. This removed the diff in the generated RPC code and should also prevent Dependabot to go crazy after we merge the PR.

PR #1491, opened 2025-04-25
tapdb: correctly fill pkScript for outputs of QueryParcels

bhandras:
Previously, QueryParcels did not populate the pkScript field of the anchor outputs. This commit addresses that issue.

### DIFF
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -3542,6 +3542,13 @@ func (a *AssetStore) QueryParcels(ctx context.Context,
 					"anchor tx: %w", err)
 			}
 
+			// Fill in the anchor transaction's output pkScripts.
+			for i, out := range outputs {
+				outIdx := out.Anchor.OutPoint.Index
+				pkScript := anchorTx.TxOut[outIdx].PkScript
+				outputs[i].Anchor.PkScript = pkScript
+			}
+
 			// Marshal anchor tx block hash from the database to a
 			// Hash type.
 			var anchorTxBlockHash fn.Option[chainhash.Hash]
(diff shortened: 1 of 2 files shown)


tapdb/assets_store.go
```diff
@@ -3542,6 +3542,12 @@ func (a *AssetStore) QueryParcels(ctx context.Context,
 					"anchor tx: %w", err)
 			}
 
+			// Fill in the anchor transaction's output pkScripts.
+			for i := range outputs {
+				outputs[i].Anchor.PkScript =
+					anchorTx.TxOut[i].PkScript
```
guggero, line 3548: This is not correct, the indexes won't always match up.
This would be correct:
```
			for i, out := range outputs {
				outIdx := out.Anchor.OutPoint.Index
				pkScript := anchorTx.TxOut[outIdx].PkScript
				outputs[i].Anchor.PkScript = pkScript
			}
```

guggero (changes requested):
Thanks for the fix, but needs some tweaking.

tapdb/assets_store.go
bhandras: Thanks, fixed!

guggero (approved):
LGTM :tada:

tapdb/assets_store.go
```diff
@@ -3542,6 +3542,13 @@ func (a *AssetStore) QueryParcels(ctx context.Context,
 					"anchor tx: %w", err)
 			}
 
+			// Fill in the anchor transaction's output pkScripts.
+			for i, out := range outputs {
+				outIdx := out.Anchor.OutPoint.Index
+				pkScript := anchorTx.TxOut[outIdx].PkScript
+				outputs[i].Anchor.PkScript = pkScript
+			}
```
ffranr, line 3550: I think some unit tests are panicking on index out of range.

bhandras: Ah yes indeed, the test needed to be updated. PTAL.

bhandras:
> LGTM 🎉

Re-requested as I added some changes to the test.

guggero (approved):

Issue #1496, opened 2025-04-29
[bug]: add CLI examples for TaprootAssetChannels Service to web API docs


ZZiigguurraatt:
We have the TaprootAssetChannels Service (https://lightning.engineering/api-docs/category/taprootassetchannels-service/) with the following commands:

1. AddInvoice
2. SendPayment
3. EncodeCustomRecords
4. FundChannel
5. DecodeAssetPayReq

Documentation for all of these say `# There is no CLI command for this RPC`.

I can get the docs with

```
4d44bef66ecb:/$ litcli ln
NAME:
   litcli ln - Interact with the Lightning Network.

USAGE:
   litcli ln command [command options] [arguments...]

COMMANDS:

   Channels:
     fundchannel  Open a Taproot Asset channel with a node on the Lightning Network.

   Invoices:
     addinvoice  Add a new invoice to receive Taproot Assets.

   Payments:
     sendpayment         Send a payment over Lightning, potentially using a mulit-asset channel as the first hop
     payinvoice          Pay an invoice over lightning using an asset.
     decodeassetinvoice  Decodes an LN invoice and displays the invoice's amount in asset units specified by an asset ID

OPTIONS:
   --help, -h  show help
   
4d44bef66ecb:/$
```

and then more specific info with something like

```
4d44bef66ecb:/$ litcli ln addinvoice -h
NAME:
   litcli ln addinvoice - Add a new invoice to receive Taproot Assets.

USAGE:
   litcli ln addinvoice [command options] asset_id asset_amount

CATEGORY:
   Invoices

DESCRIPTION:
   
  Add a new invoice, expressing intent for a future payment, received in
  Taproot Assets.
  

OPTIONS:
   --memo value                     a description of the payment to attach along with the invoice (default="")
   --preimage value                 the hex-encoded preimage (32 byte) which will allow settling an incoming HTLC payable to this preimage. If not set, a random preimage will be created.
   --amt value                      the amt of satoshis in this invoice (default: 0)
   --amt_msat value                 the amt of millisatoshis in this invoice (default: 0)
   --description_hash value         SHA-256 hash of the description of the payment. Used if the purpose of payment cannot naturally fit within the memo. If provided this will be used instead of the description(memo) field in the encoded invoice.
   --fallback_addr value            fallback on-chain address that can be used in case the lightning payment fails
[...]

Issue #1493, opened 2025-04-28
[bug]: add `group_key` to `AssetPayReq` in `DecodeAssetPayReq`


ZZiigguurraatt:
In `AddInvoice`, `SendPayment`, and `FundChannel` we have a `group_key` that can be input in the request in place of the `asset_id` for sending any `asset_id` within the same grouped asset.

The RPC `DecodeAssetPayReq` does not have such an option for `group_key` and should be added to the request ( `AssetPayReq` ) maintain generality amoung all RPCs available in the TaprootAssetChannels Service.

Issue #1485, opened 2025-04-24
[feature]: heterogeneous multi asset LN channels


guggero:
This is a placeholder issue to brainstorm what changes we need to support committing multiple different assets that are not related (e.g. from different asset groups altogether, for example USD and EUR stablecoins) into the same Lightning Network channel.

Conceptually this should be pretty easy to understand: Each channel can hold different balances of assets. Each payment (or to be more precise each HTLC) shifts the balance of a single asset (or asset group).
The user interface also shouldn't need to change, as the user already specifies what asset ID (or asset group) an invoice should be created for or what asset should be used to make a payment.

There are a couple of technical changes required though.
This is a non-complete list that we'll want to extend as more things come to mind:
 - Channel allocation code (`tapsend/allocation.go`): Currently all assets are treated as fungible. The concept of distinct "classes" of assets needs to be introduced and implemented throughout that allocation code
 - RFQ system: balance checks in the different forwarding interceptors need to be updated to check for only the balances of the asset class mentioned in an HTLC

In a first version one payment/invoice would only be able to send or request assets from a single class of assets in a channel. But it is conceivable that a payment could send or request multiple such assets at the same time. But whether that is an actually useful feature or how that would look like on the user interface is not addressed in this issue.

PR #1498, opened 2025-04-30
misc: fix litcli command docs, add group key support to DecodeAssetPayReq

guggero:
Fixes a couple of smaller issues.

Fixes https://github.com/lightninglabs/taproot-assets/issues/1496.
Fixes https://github.com/lightninglabs/taproot-assets/issues/1493.

PR with itests is here:


ZZiigguurraatt:
While we are updating docs, can we add a comment `This command is a shortcut for 'sendpayment --pay_req='` when we use `litcli ln payinvoice -h` so that people understand the difference with `litcli ln sendpayment`? We already get such a message when we run `lncli payinvoice -h`.

guggero:
> While we are updating docs, can we add a comment `This command is a shortcut for 'sendpayment --pay_req='` when we use `litcli ln payinvoice -h` so that people understand the difference with `litcli ln sendpayment`? We already get such a message when we run `lncli payinvoice -h`.

Yes, added that here: https://github.com/lightninglabs/lightning-terminal/pull/1052.

ZZiigguurraatt:
> Yes, added that here: [lightninglabs/lightning-terminal#1052](https://github.com/lightninglabs/lightning-terminal/pull/1052).


Not seeing that. Here is what I get


```
de20a9034ac0:/$ litcli ln payinvoice -h
NAME:
   litcli ln payinvoice - Pay an invoice over lightning using an asset.

USAGE:
   litcli ln payinvoice [command options] pay_req [--asset_id=X | --group_key=X] [--rfq_peer_pubkey=y] [--allow_overpay]

CATEGORY:
   Payments

DESCRIPTION:
   
  This command attempts to pay an invoice using an asset channel as the
  source of the payment. The asset ID of the channel must be specified
  using the --asset_id flag.
  

OPTIONS:
   --pay_req value              a zpay32 encoded payment request to fulfill
   --fee_limit value            maximum fee allowed in satoshis when sending the payment (default: 0)
   --fee_limit_percent value    percentage of the payment's amount used as the maximum fee allowed when sending the payment (default: 0)
   --timeout value              the maximum amount of time we should spend trying to fulfill the payment, failing after the timeout has elapsed (default: 1m0s)
   --cancelable                 if set to true, the payment loop can be interrupted by manually canceling the payment context, even before the payment timeout is reached. Note that the payment may still succeed after cancellation, as in-flight attempts can still settle afterwards. Canceling will only prevent further attempts from being sent
   --cltv_limit value           the maximum time lock that may be used for this payment (default: 0)
   --last_hop value             pubkey of the last hop (penultimate node in the path) to route through for this payment
   --outgoing_chan_id value     short channel id of the outgoing channel to use for the first hop of the payment; can be specified multiple times in the same command
   --force, -f                  will skip payment request confirmation
   --allow_self_payment         allow sending a circular payment to self
   --data value                 attach custom data to the payment. The required format is: <record_id>=<hex_value>,<record_id>=<hex_value>,.. For example: --data 3438382=0a21ff. Custom record ids start from 65536.
   --inflight_updates           if set, intermediate payment state updates will be displayed. Only valid in combination with --json.
   --max_parts value            the maximum number of partial payments that may be used (default: 16)
   --json                       if set, payment updates are printed as json messages. Set by default on Windows because table formatting is unsupported.
   --max_shard_size_sat value   the largest payment split that should be attempted if payment splitting is required to attempt a payment, specified in satoshis (default: 0)
   --max_shard_size_msat value  the largest payment split that should be attempted if payment splitting is required to attempt a payment, specified in milli-satoshis (default: 0)
   --amp                        if set to true, then AMP will be used to complete the payment
   --time_pref value            (optional) expresses time preference (range -1 to 1) (default: 0)
   --amt value                  (optional) number of satoshis to fulfill the invoice (default: 0)
   --asset_id value             the asset ID of the asset to use when sending payments with assets; cannot be used at the same time as --group_key
   --group_key value            the group key of the asset to use when sending payments with assets; cannot be used at the same time as --asset_id
   --rfq_peer_pubkey value      (optional) the public key of the peer to ask for a quote when converting from assets to sats; must be set if there are multiple channels with the same asset ID present
   --allow_overpay              allow sending asset payments that are uneconomical because the required non-dust amount for an asset carrier HTLC plus one asset unit is higher than the total invoice/payment amount that arrives at the destination; meaning that the total amount sent exceeds the total amount received plus routing fees
   
de20a9034ac0:/$ 
```

but I do here, which is a mistake

```
de20a9034ac0:/$ litcli ln addinvoice -h
NAME:
   litcli ln addinvoice - Add a new invoice to receive Taproot Assets.

USAGE:
   litcli ln addinvoice [command options] [--asset_id=X | --group_key=X] --asset_amount=Y [--rfq_peer_pubkey=Z] 

CATEGORY:
   Invoices

DESCRIPTION:
   
  Add a new invoice, expressing intent for a future payment, received in
  Taproot Assets.

  This command is a shortcut for 'sendpayment --pay_req='.
  

OPTIONS:
   --memo value                     a description of the payment to attach along with the invoice (default="")
   --preimage value                 the hex-encoded preimage (32 byte) which will allow settling an incoming HTLC payable to this preimage. If not set, a random preimage will be created.
   --amt value                      the amt of satoshis in this invoice (default: 0)
   --amt_msat value                 the amt of millisatoshis in this invoice (default: 0)
   --description_hash value         SHA-256 hash of the description of the payment. Used if the purpose of payment cannot naturally fit within the memo. If provided this will be used instead of the description(memo) field in the encoded invoice.
   --fallback_addr value            fallback on-chain address that can be used in case the lightning payment fails
   --expiry value                   the invoice's expiry time in seconds. If not specified, an expiry of 86400 seconds (24 hours) is implied. (default: 0)
   --cltv_expiry_delta value        The minimum CLTV delta to use for the final hop. If this is set to 0, the default value is used. The default value for cltv_expiry_delta is configured by the 'bitcoin.timelockdelta' option. (default: 0)
   --private                        encode routing hints in the invoice with private channels in order to assist the payer in reaching you. If amt and amt_msat are zero, a large number of hints with these channels can be included, which might not be desirable.
   --amp                            creates an AMP invoice. If true, preimage should not be set.
   --blind                          creates an invoice that contains blinded paths. Note that invoices with blinded paths will be signed using a random ephemeral key so as not to reveal the real node ID of this node.
   --min_real_blinded_hops --blind  The minimum number of real hops to use in a blinded path. This option will only be used if --blind has also been set. (default: 0)
   --num_blinded_hops --blind       The number of hops to use for each blinded path included in the invoice. This option will only be used if --blind has also been set. Dummy hops will be used to pad paths shorter than this. (default: 0)
   --max_blinded_paths --blind      The maximum number of blinded paths to add to an invoice. This option will only be used if --blind has also been set. (default: 0)
   --blinded_path_omit_node value   The pub key (in hex) of a node not to use on a blinded path. The flag may be specified multiple times.
   --asset_id value                 the asset ID of the asset to receive; cannot be used at the same time as --group_key
   --group_key value                the group key of the asset to receive; cannot be used at the same time as --asset_id
   --asset_amount value             the amount of assets to receive (default: 0)
   --rfq_peer_pubkey value          (optional) the public key of the peer to ask for a quote when converting from assets to sats for the invoice; must be set if there are multiple channels with the same asset ID present
   
de20a9034ac0:/$ 
```

guggero:
Oops, yeah, added it to the wrong command. Updated in https://github.com/lightninglabs/lightning-terminal/pull/1052 (can you please comment CLI related issues there so we don't mix responsibilities in the PRs?).

ZZiigguurraatt:
I run

```
c63e5f7232b3:/$ lncli addinvoice --amt 44444
{
    "r_hash":  "2e5b78ca6916e72401cbfb74003a38d58feb4b48f3cd76fcac25faaf9230b29a",
    "payment_request":  "lnbcrt444440n1p5pfkknpp59edh3jnfzmnjgqwtld6qqw3c6k87kj6g70xhdl9vyha2ly3sk2dqdqqcqzzsxqyz5vqsp5z7x8c8h8n8ndr3zu5kgvwp337lnmmtn3n9qs4504zvrgqhlkj7rs9qxpqysgqdyfwpsqcs4dp20m9u5luqavyld3u5378r4tuhh8730rs2d3ry85z46xlxxna9d0ty5jy0sh83lvlg36uxkgjykxny798684frz5phrgpsk0tkn",
    "add_index":  "71",
    "payment_addr":  "178c7c1ee799e6d1c45ca590c70631f7e7bdae7199410ad1f51306805ff69787"
}
c63e5f7232b3:/$ 
```

Then I run

```
e97cc576779c:/$ litcli ln decodeassetinvoice --pay_req lnbcrt444440n1p5pfkknpp59edh3jnfzmnjgqwtld6qqw3c6k87kj6g70xhdl9vyha2ly3sk2dqdqqcqzzsxqyz5vqsp5z7x8c8h8n8ndr3zu5kgvwp337lnmmtn3n9qs4504zvrgqhlkj7rs9qxpqysgqdyfwpsqcs4dp20m9u5luqavyld3u5378r4tuhh8730rs2d3ry85z46xlxxna9d0ty5jy0sh83lvlg36uxkgjykxny798684frz5phrgpsk0tkn --group_key 03fef77c717c7017bd29326a71d0f7f3829439d43419ec7bd5cf017d240308f29c
{
    "asset_amount":  "4444400",
    "decimal_display":  {
        "decimal_display":  3
    },
    "asset_group":  {
        "raw_group_key":  "03a6553ff1aa8bb1dc91b35e1f8428f99e6dfc3a66e39945ad9c0c22fffec677ff",
        "tweaked_group_key":  "03fef77c717c7017bd29326a71d0f7f3829439d43419ec7bd5cf017d240308f29c",
        "asset_witness":  "01403e51e1e3ff0f3efe9c5bcb3cf6aa50f627b7f81df0080c19a74033631cdeb290bfa209612a74e1826d2bb8143aaa56dd3ff7a95ca22b9de650bc8e98fe1d76dd",
        "tapscript_root":  "a4e839c197da2cf98ca52789ddd714d2f7eb09702c0119c90bf8242fe55ca609"
    },
    "genesis_info":  {
        "genesis_point":  "6791c03d2ce28b1cfa6a8211615a9e7778592e3c5464ecd1e566e8c8f919850f:1",
        "name":  "Asset1",
        "meta_hash":  "827c23a4b1a5f4f9a9a42e61aa7bf7bf7e7925b4723728b60ad1918d632e2b0a",
        "asset_id":  "90aee2e3468a20be13dc5023d630ac201c95fdbc72ddea16653b39fe04a8d954",
        "asset_type":  "NORMAL",
        "output_index":  0
    },
    "pay_req":  {
        "destination":  "0230d12aed73b0b55dc84e4d21248fa7fb1d6ce73910d979946c69e16de7511370",
        "payment_hash":  "2e5b78ca6916e72401cbfb74003a38d58feb4b48f3cd76fcac25faaf9230b29a",
        "num_satoshis":  "44444",
        "timestamp":  "1746197203",
        "expiry":  "86400",
        "description":  "",
        "description_hash":  "",
        "fallback_addr":  "",
        "cltv_expiry":  "80",
        "route_hints":  [],
        "payment_addr":  "178c7c1ee799e6d1c45ca590c70631f7e7bdae7199410ad1f51306805ff69787",
        "num_msat":  "44444000",
        "features":  {
            "8":  {
                "name":  "tlv-onion",
                "is_required":  true,
                "is_known":  true
            },
            "14":  {
                "name":  "payment-addr",
                "is_required":  true,
                "is_known":  true
            },
            "17":  {
                "name":  "multi-path-payments",
                "is_required":  false,
                "is_known":  true
            },
            "25":  {
                "name":  "route-blinding",
                "is_required":  false,
                "is_known":  true
            }
        },
        "blinded_paths":  []
    }
}
e97cc576779c:/$ 
```

and `asset_amount` is off by a factor of `100`.

ZZiigguurraatt:
When running `litcli ln decodeassetinvoice --group_key`, does it make sense to return an `asset_id`, `meta_hash`, etc.? Seems like this might mislead users into thinking there is a single `asset_id`, but really it seems you are just responding with the first one.

guggero:
> When running `litcli ln decodeassetinvoice --group_key`, does it make sense to return an `asset_id`, `meta_hash`, etc.? Seems like this might mislead users into thinking there is a single `asset_id` for the group, but really it seems you are just responding with the first one.

Makes sense. I've added a condition that only prints that information if we requested to decode the invoice with an asset ID and not a group key.

address/book.go
```diff
@@ -82,8 +82,9 @@ type QueryParams struct {
 // known to universe servers in our federation.
 type AssetSyncer interface {
 	// SyncAssetInfo queries the universes in our federation for genesis
-	// and asset group information about the given asset ID.
-	SyncAssetInfo(ctx context.Context, assetID *asset.ID) error
+	// and asset group information about the given asset ID or group key.
+	SyncAssetInfo(ctx context.Context, assetID *asset.ID,
```
Roasbeef, line 86: Make it an `asset.Specifier`?

address/book.go
```diff
@@ -264,6 +265,25 @@ func (b *Book) QueryAssetInfo(ctx context.Context,
 	return assetGroup, nil
 }
 
+// SyncAssetGroup attempts to enable asset sync for the given asset group, then
+// perform an initial sync with the federation for that group.
+func (b *Book) SyncAssetGroup(ctx context.Context,
```
Roasbeef, line 270: Alternatively, we can have `QueryAssetInfo` accept an `asset.Specifier`.

Roasbeef: The way the function is setup, if you call it twice, only the first time will it actually try to fetch data.

universe/auto_syncer.go
```diff
@@ -807,10 +808,21 @@ func (f *FederationEnvoy) tryFetchServers() ([]ServerAddr, error) {
...
-	assetID *asset.ID) error {
+	assetID *asset.ID, groupKey *btcec.PublicKey) error {
 
-	if assetID == nil {
-		return fmt.Errorf("no asset ID provided")
+	uniID := Identifier{
+		ProofType: ProofTypeIssuance,
+	}
+
+	switch {
```
Roasbeef, line 817: Implicit assumption made here re if the user _only_ sets group key (or not).

rpcserver.go
```diff
@@ -8600,6 +8651,96 @@ func (r *rpcServer) DecodeAssetPayReq(ctx context.Context,
...
+	}
+
+	// Since we know we have at least one leaf, we can just take leaf ID and
+	// query the keys with it. We just need one so we can fetch the actual
+	// proof to find out the asset ID.
+	leafID := leaves[0]
+	leafKeys, err := r.cfg.Multiverse.UniverseLeafKeys(
+		ctx, universe.UniverseLeafKeysQuery{
+			Id:    leafID.ID,
+			Limit: 1,
```
Roasbeef, line 8712: What order are we actually relying on here?

ZZiigguurraatt:
@guggero  do you see my review comments for this one too now?

address/book.go
guggero: But `QueryAssetInfo` does the opposite of what we want: Given an asset ID it finds out what the group key is.
But what we need for this use case is: Given a group key, what asset IDs does that group contain?

rpcserver.go
```diff
@@ -8527,6 +8573,96 @@ func (r *rpcServer) DecodeAssetPayReq(ctx context.Context,
...
+		universe.ProofTypeIssuance,
+	)
+	if err != nil {
+		return asset.ID{}, fmt.Errorf("error fetching leaves: %w", err)
+	}
+
+	tapdLog.Tracef("Found %d leaves for group key %x",
+		len(leaves), groupKey.SerializeCompressed())
+
+	// If there are no leaves, then we need to sync the asset group.
```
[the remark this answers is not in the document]
guggero: We'll enforce that all assets minted will have the same decimal display. But that will be a different PR that fixes #1503. At this point we need to be able to assume that any asset of the same group key has the same properties in this regard. Finding out which one was the actual first tranche isn't easy to do here.

rpcserver.go
guggero: No specific order (meaning natural database order), since it's not really important which asset we get.

rpcserver.go
```diff
@@ -8581,29 +8581,73 @@ func (r *rpcServer) assetInvoiceAmt(ctx context.Context,
...
+	// We made sure that only one is set, so let's now use the asset ID
+	// or group key.
+	switch {
+	// The asset ID is easy, we can just copy the bytes.
+	case len(payReq.AssetId) != 0:
+		copy(assetID[:], payReq.AssetId)
+
+	// The group key is a bit more involved. We first need to sync the asset
+	// group, then fetch the leaves that are associated with this group key.
+	// From there, we can look up the asset ID of one of the group's
```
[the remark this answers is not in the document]
gijswijs: I guess this answers my question

https://github.com/lightninglabs/taproot-assets/blob/41650473b6927677f44674eca617f44837a3eaf7/rpcserver.go#L8798

gijswijs (approved):
Just one nit.  No blockers, so LGTM!

rfq/manager.go
```diff
@@ -1044,12 +1044,12 @@ func (m *Manager) ChannelCompatible(ctx context.Context,
 			return false, err
 		}
 
-		if !match {
-			return false, err
+		if match {
+			return true, err
```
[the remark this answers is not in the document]
guggero: So for now (until https://github.com/lightninglabs/taproot-assets/issues/1485 is implemented), we assume that each channel only has fungible assets. Meaning that we'd never have `bucks` and `yen`.

But I guess the question here is similar to what @ZZiigguurraatt asked offline: Are you supposed to be able to send an asset HTLC over a grouped asset channel by only specifying `--asset_id` (but picking an asset ID that's actually in the channel).

My argumentation is: Yes, as long as the `--asset_id` is actually part of the channel's funded asset IDs (and not one that is part of the group but not present in the channel).
BUT: That would only be a UX thing and wouldn't actually dictate what in-channel asset ID is really assigned to the HTLC, that would only be decided in the allocation code and is never up to the user to decide (because things there need to be strictly deterministic and predictable).

But perhaps that's then too confusing to the user and we might want to allow it only in the RFQ-internal code but not give the user that choice on the user interface?

The second discussion is about the `bool` return value of `ChannelCompatible`:
It is currently used in two places:
 1. https://github.com/lightninglabs/taproot-assets/blob/4b108f3b46dca9774f502eb669fa098c734879db/rpcserver.go#L8513: For checking what channel/peer to choose for a payment or invoice.
 2. https://github.com/lightninglabs/taproot-assets/blob/2655a69094606196a97c98989448e78098b4e8c3/rfq/manager.go#L582: For binding a quote to a specific channel

IMO both of these cases come down to the decision if we want to allow the user to initiate the RFQ process with an asset ID even for a channel that was funded with a group key (and has more than one asset ID in it).
If the answer is yes, then we need to allow a partial match.
If no, then this commit isn't needed.

tapgarden/mock.go
```diff
@@ -823,8 +823,8 @@ func (m *MockAssetSyncer) FetchAsset(id asset.ID) (*asset.AssetGroup, error) {
 	}
 }
 
-func (m *MockAssetSyncer) SyncAssetInfo(_ context.Context,
-	id *asset.ID) error {
+func (m *MockAssetSyncer) SyncAssetInfo(_ context.Context, id *asset.ID,
+	_ *btcec.PublicKey) error {
```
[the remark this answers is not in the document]
guggero: Oops, yeah, fixed.

ZZiigguurraatt (approved):

rfq/manager.go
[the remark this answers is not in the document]
guggero: We decided offline that we don't want to support using `--asset_id` on grouped channels. So I'm going to add a new error message so it's more clear to the user what they need to do.

guggero:
Okay, as discussed, I reverted the matching logic in the channel compatibility test.
We now return a more informative error (`no compatible asset channel found for %s, make sure to use group key for grouped asset channels`) in the following cases:
 - The user uses `--asset_id` on send or receive
 - There are grouped asset channels present
 - The channel that contains the asset with the ID specified contains other asset IDs

Which means, one can still use `--asset_id` for a grouped channel that only has assets from a single tranche. But as soon as multiple tranches are in a channel, only the `--group_key` flag will work.

guggero:
The litd itest is expected to fail btw. It will be fixed once https://github.com/lightninglabs/lightning-terminal/pull/1052 is merged. So as long as that PR is green, this PR can be merged.

Issue #1495, opened 2025-04-29
[feature]: Configure custom histogram buckets for tapd gRPC latency metrics


ffranr:
Currently, tapd is using the default histogram configuration for gRPC latency metrics, which caps the largest bucket at 10 seconds. As a result, any RPCs taking longer than 10s are incorrectly classified, making it difficult to accurately monitor long-running operations.

We should explicitly configure larger histogram buckets (e.g., 20s, 30s, 60s) in the tapd code to improve observability.

The universe server RPC endpoints we need to track more accurately are:
- `UniverseStats`
- `QueryAssetStats`
- `QueryProof`
- `InsertProof`

---

I think we just need to pass in better bucket sizes here:

https://github.com/lightninglabs/taproot-assets/blob/2173dc4700fe5678acbac2cb563e0815b8945917/monitoring/monitoring_on.go#L31-L34

ffranr:
Take a look also at `WithHistogramBuckets` in `github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus@v1.0.0-rc.0/options.go`

ffranr:
Terminal Web team report that a unary endpoint took 45mins to run one day and 4mins the next. We should be able to track that sort of variability.

PR #1500, opened 2025-04-30
monitoring: expand Prometheus buckets for RPC endpoint latency

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/1495

---

Default bucket sizes top out at 10s, causing longer-running RPC calls to be grouped into the largest bucket. This commit adds more buckets to provide finer-grained visibility into endpoint latency.

Terminal Web team report that a unary endpoint took 45mins to run one day and 4mins the next. We should be able to track that sort of variability.

### DIFF
--- a/monitoring/monitoring_on.go
+++ b/monitoring/monitoring_on.go
@@ -29,7 +29,14 @@ func GetPromInterceptors(cfg *PrometheusConfig) ([]grpc.UnaryServerInterceptor,
 	}
 
 	if cfg.PerfHistograms {
-		opt := grpc_prometheus.WithServerHandlingTimeHistogram()
+		// Set the histogram buckets in seconds.
+		histogramBuckets := []float64{
+			0.01, 0.1, 0.5, 1, 5, 10, 60, 120, 240, 600, 1200,
+		}
+		opt := grpc_prometheus.WithServerHandlingTimeHistogram(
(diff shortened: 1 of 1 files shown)


joao-krauchenco-mottu (approved):

monitoring/monitoring_on.go
```diff
@@ -29,7 +29,12 @@ func GetPromInterceptors(cfg *PrometheusConfig) ([]grpc.UnaryServerInterceptor,
 	}
 
 	if cfg.PerfHistograms {
-		opt := grpc_prometheus.WithServerHandlingTimeHistogram()
+		// Set the histogram buckets in seconds.
+		histogramBuckets := []float64{1, 5, 10, 60, 120, 240, 600, 1200}
```
guggero, line 33: Should we keep some of the smaller (sub-second) values? Like `0.01, 0.1, 0.5`? Because one second is quite long for any gRPC call, so we wouldn't notice faster calls to get slower until they actually hit a full second (I hope you know what I mean).

guggero (approved):
One question/suggestion, otherwise LGTM :tada:

monitoring/monitoring_on.go
ffranr: Sure why not, I was thinking of the data cost, but we can scale back later if it seems too much.

Roasbeef (approved):
LGTM 🧼

Issue #874, opened 2024-04-05
[feature]: add support for re-useable group key based addresses


jharveyb:
Spinoff of #291.

Invoices are an extension of our existing address format, that include an expiry time. Using invoices requires more interaction with a hashmail courier service than addresses, but they allow for receiving a balance of a grouped asset.

From an invoice, we can compute a unique ID. This commits to most invoice fields, but excludes `asset.AssetCommitmentKey()` so that any member of an asset group could satisfy an invoice.

```bash
invoiceID := H(tapscriptRoot || asset.TapCommitmentKey() || chain_params || asset_version 
    || internal_key || script_key || amount || courier_addr)
```

To send a grouped asset, the sender needs to communicates a `LeafMap` to the receiver. This provides the missing information to compute the on-chain output for the transfer.

```go
outputLeafMap := map[asset.ID]uint64
```

With an invoice, we can compute a unique ID. The sender can use the ID + information from the asset transfer inputs to upload a leaf map to a hashmail courier. The receiver polls the hashmail courier, and once they fetch the leaf map they can reconstruct the on-chain output for the send.

Specifically, on the sender side:

```go
firstScriptKey := transfer.inputs[0].ScriptKey
invoiceID := invoice.ID()
invoiceSharedSecret := ECDH(firstScriptKey, invoice.ScriptKey)

// upload sender pubkey so the receiver can derive invoiceSharedSecret
courier.WriteSenderKey(invoiceID, firstScriptKey)

// build the output leaf map from the set of all transfer outputs
outputLeafMap := make(map[asset.ID]uint64, len(transfer.inputs))
for _, vOutput := range transfer.outputs {
    // filter out change outputs, asset splits, etc.
    if invoice.Satisfies(vOutput.Asset) {
        outputLeafMap[vOutput.Asset.ID] = vOutput.Asset.Amount
    }
}

// upload the leaf map the receiver needs to find the on-chain TX
courier.WriteLeafMap(invoiceSharedSecret, outputLeafMap)
```

For the receiver:

```go
newBlockChan := cfg.ChainBridge.RegisterBlockEpochNtfn()

for {
    select {
    // On each new block, check for new data at the hashmail courier
    // for all unexpired invoices. If we received some data, complete
    // the receive process.
    case newBlock := <- newBlockChan:
       invoices := store.LoadInvoices()
       invoices := invoices.RemoveExpired()
       for _, invoice := range invoices {
[...]

PR #1502, opened 2025-05-01
[group key addrs 1/?]: add new authmailbox RPC server and client

guggero:
Implements the mailbox server and client as described here: https://github.com/lightninglabs/taproot-assets/issues/874#issuecomment-2752783389

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1502)


rpcutils/proof.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/taproot-assets/proof"
+	mboxrpc "github.com/lightninglabs/taproot-assets/taprpc/authmailboxrpc"
```
ffranr, line 11: I think file name here should be more specific to `authmailboxrpc`. IMO `proof.go` is too general.

proof/tx.go
```diff
@@ -167,3 +181,65 @@ func (p *TxMerkleProof) Decode(r io.Reader) error {
...
+	BlockHeight uint32
+
+	// MerkleProof is the proof that the transaction is included in the
+	// block and its merkle root.
+	MerkleProof TxMerkleProof
+
+	// ClaimedOutPoint is the outpoint that is being proved to exist in the
+	// transaction.
+	ClaimedOutPoint wire.OutPoint
+}
```
ffranr, line 204: Anyone can prove an outpoint in a block. The UTXO set is public. Any sender could use up all UTXOs.

Maybe I'm missing something, please see https://github.com/lightninglabs/taproot-assets/issues/874#issuecomment-2858564454

taprpc/authmailboxrpc/mailbox.proto
```diff
@@ -0,0 +1,221 @@
...
+message SendMessageRequest {
+    // The public key identifier of the intended receiver (ReceiverID), encoded
+    // as the raw bytes of the compressed public key.
+    bytes receiver_id = 1;
+
+    // The ECIES encrypted message payload.
+    bytes encrypted_payload = 2;
+
+    // The sender's ephemeral public key used for ECIES.
+    bytes sender_ephemeral_pubkey = 3;
```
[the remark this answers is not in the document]
Roasbeef: Ah I see we have `receiver_id` above. But with that said, I think we can omit this.

taprpc/authmailboxrpc/mailbox.proto
```diff
@@ -0,0 +1,221 @@
...
+    bytes signature = 1;
+}
+
+// Wrapper message for responses sent FROM the server TO the client
+// during the ReceiveMessages stream.
+message ReceiveMessagesResponse {
+    oneof response_type {
+        Challenge challenge = 1;
+        bool auth_success = 2;
+        MailboxMessage message = 3;
```
Roasbeef, line 183: Should this be `repeated`? That way we can send more than one message over a single message. 

We should also limit the size of the message to something reasonable to start with also.

authmailbox/message.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+)
+
+const (
+	// MsgMaxSize is the maximum size of a message in bytes.
+	MsgMaxSize = 65536
```
Roasbeef, line 12: 👍

authmailbox/message.go
```diff
@@ -0,0 +1,102 @@
...
+	// public key of the receiver.
+	ReceiverKey btcec.PublicKey
+
+	// EncryptedPayload is the encrypted message payload. This is the actual
+	// content of the message, encrypted with EICS.
+	EncryptedPayload []byte
+
+	// SenderEphemeralKey is the ephemeral public key of the sender used to
+	// encrypt the message.
+	SenderEphemeralKey btcec.PublicKey
```
Roasbeef, line 38: Mentioned in the proto file, but I don't think we need to store/ref this anywhere.

taprpc/authmailboxrpc/mailbox.proto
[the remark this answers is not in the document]
guggero: If you look at https://github.com/lightninglabs/taproot-assets/pull/1512, the sender's ephemeral key (the one used for the ECDH to arrive at the shared key) isn't actually part of the encrypted payload (only the nonce and the tag). Are you saying we should move the sender's ECDH key from this message into the encrypted payload itself?

authmailbox/message.go
guggero: See message above. We need this to decrypt the message. But we can move it to be part of the encrypted payload instead.

rpcutils/proof.go
guggero: Moved things around now, the (un)marshal function is now in the proof package.

proof/tx.go
guggero: Addressed this by adding the required internal key (and optional merkle root) fields to the TX proof.

authmailbox/message.go
Roasbeef: Ah ok, yeah that makes sense. In my mental model it was "part" of the payload, see this comment: https://github.com/lightninglabs/taproot-assets/pull/1502#discussion_r2110683393.

taprpc/authmailboxrpc/mailbox.proto
```diff
@@ -0,0 +1,232 @@
...
+    // proof "token". The output at the given index must be a P2TR output.
+    // The server must verify that the txid_hex matches the hash of the provided
+    // transaction data, and that this specific outpoint index exists in the
+    // transaction.
+    taprpc.OutPoint claimed_outpoint = 5;
+
+    // The Taproot internal key used to construct the P2TR output that is
+    // claimed by the outpoint above. Must be provided alongside the Taproot
+    // Merkle root to prove knowledge of the output's construction.
+    bytes internal_key = 6;
```
Roasbeef, line 90: 👍

taprpc/authmailboxrpc/mailbox.proto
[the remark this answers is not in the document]
guggero: Okay, makes sense, I'll move the sender ephemeral key into the blob/payload.

taprpc/authmailboxrpc/mailbox.proto
```diff
@@ -0,0 +1,229 @@
...
+    repeated bytes sibling_hashes = 1;
+
+    // The bitmask indicating the direction (left/right) of each sibling hash
+    // in the Merkle tree. Each bit corresponds to a sibling hash in the
+    // sibling_hashes list. 0 indicates left, 1 indicates right.
+    repeated bool bits = 2;
+}
+
+// Encapsulates the full proof required for sender authentication, demonstrating
+// that a specific P2TR transaction output was included in a Bitcoin block.
```
ffranr: `P2TR transaction output was` -> `P2TR transaction output and internal key were`

or otherwise mention internal key here?

taprpc/authmailboxrpc/mailbox.proto
```diff
@@ -0,0 +1,229 @@
...
+
+// Represents a structured error encountered during the ReceiveMessages stream.
+message ReceiveError {
+    enum Code {
+        CODE_UNSPECIFIED = 0;
+        INVALID_RECEIVER_ID = 1;
+        INVALID_SIGNATURE = 2;
+        AUTH_TIMEOUT = 3;
+        INTERNAL_ERROR = 4;
+        TEMPORARILY_UNAVAILABLE = 5;
```
ffranr: I think we should add doc comments for these enum values and also for the fields in the other messages.

taprpc/tapcommon.proto
```diff
@@ -0,0 +1,18 @@
...
+package taprpc;
+
+option go_package = "github.com/lightninglabs/taproot-assets/taprpc";
+
+// Represents a Bitcoin transaction outpoint.
+message OutPoint {
+    /*
+    Raw bytes representing the transaction id.
+    */
+    bytes txid = 1;
```
ffranr, line 12: for the txid specifically, should we use string here?

ffranr: or i guess this is a refactor of a message that's already in use

authmailbox/client.go
```diff
@@ -0,0 +1,258 @@
...
+)
+
+var (
+	// ErrServerShutdown is the error returned if the mailbox server signals
+	// it's going to shut down.
+	ErrServerShutdown = errors.New("server shutting down")
+
+	// ErrServerErrored is the error returned if the mailbox server
+	// sends back an error instead of a proper message.
+	ErrServerErrored = errors.New("server sent unexpected error")
```
ffranr: may be `ErrServerInternal`?

authmailbox/client.go
```diff
@@ -0,0 +1,258 @@
...
+	// MaxBackoff is the maximum time waited between connection attempts.
+	MaxBackoff time.Duration
+}
+
+// Client performs the client side part of mailbox message exchange.
+type Client struct {
+	cfg *ClientConfig
+
+	started atomic.Bool
+	stopped atomic.Bool
```
ffranr: consider:
```
startOnce sync.Once
stopOnce  sync.Once
```

authmailbox/client.go
```diff
@@ -0,0 +1,258 @@
...
+
+	return c.connectAndAuthenticate(ctxl, receiverKey, filter)
+}
+
+// connectAndAuthenticate opens a stream to the server and authenticates the
+// account to receive updates. It returns the subscription and a bool that
+// indicates if recovery can be continued. That value can safely be ignored if
+// recovery is not requested. Checking the returned error must take precedence
+// to the boolean flag.
+func (c *Client) connectAndAuthenticate(ctx context.Context,
```
ffranr: what is "recovery" here? I think the `bool` return doesn't exist.

taprpc/tapcommon.proto
guggero: Yeah, it's an existing message, don't think we should refactor it in this PR.

taprpc/authmailboxrpc/mailbox.proto
guggero: Actually, we're not using these error codes at the moment, so I removed them instead.

litbot-9000:
@roasbeef: review reminder
@ffranr: review reminder

Roasbeef (approved):
LGTM 🦋

> Reviewed 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1502#-:-OT-bUAW57XGXlsrXese:b-lila98)* status: all files reviewed, 19 unresolved discussions (waiting on @ffranr and @guggero)

___
*[`authmailbox/client.go` line 133 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/1502#-OT-aku_1IOzzywcVh5a:-OT-aku_1IOzzywcVh5b:ba1i6ko) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/937db827133f47728cba07709b5cbee8f0cc8a1b/authmailbox/client.go#L133)):*
> ```Go
> func (c *Client) Stop() error {
> 	var stopErr error
> 	c.stopOnce.Do(func() {
> ```

:+1:  for `sync.Once`. 
___
*[`authmailbox/server.go` line 99 at r2](https://reviewable.io/reviews/lightninglabs/taproot-assets/1502#-OT-bBKL5W4gL7quY151:-OT-bBKL5W4gL7quY152:bhrmmt5) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/937db827133f47728cba07709b5cbee8f0cc8a1b/authmailbox/server.go#L99)):*
> ```Go
> // Start signals that the RPC server starts accepting requests.
> func (s *Server) Start() error {
> 	if !s.started.CompareAndSwap(false, true) {
> ```

`sync.Once` here again to be consistent with the rest of the PR?

PR #1505, opened 2025-05-02
channels: fix coop close asset distribution for grouped asset channels

guggero:
What this issue fixes is a bit hard to explain. So I'll just copy the comment that's above the fix:

```golang
	// [...] When we [coop] close a channel, the output of the
	// `lncli closedchannels` command will show the last commitment state of
	// the channel as the closing asset balance. Which is correct in terms
	// of balances. But if there are multiple different asset IDs (e.g., in
	// a grouped asset channel), then _how_ those pieces are distributed
	// within the commitment transaction depends on the order of the
	// allocations. And the order of the allocations is dependent on the
	// BTC amount and the pkScript of the BTC-level output. Both of which
	// are different in the coop close output (we set the asset-level output
	// BTC amount to the dummy amount, and the pkScript will be a newly
	// derived internal key with no sibling script path).
	// So, long story short: If we used the tapsend.DistributeCoins method
	// here, it could happen that the actual asset output distribution shown
	// in the `lncli closedchannels` command would be different from the
	// actual distribution in the co-op close transaction.
	// This could mostly be seen as an UX-only issue, but was actually
	// discovered while attempting to assert the final closing balance of
	// grouped asset channels in the litd integration test.
```

### DIFF
--- a/tapchannel/commitment.go
+++ b/tapchannel/commitment.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"net/url"
 	"sort"
 
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
@@ -28,6 +29,7 @@ import (
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"golang.org/x/exp/maps"
 )
 
 // DecodedDescriptor is a wrapper around a PaymentDescriptor that also includes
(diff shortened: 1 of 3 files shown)


guggero:
Corresponding itest case that found this issue is here: https://github.com/lightninglabs/lightning-terminal/pull/1056

Roasbeef (approved):
LGTM 📩

tapchannel/commitment.go
```diff
@@ -1505,6 +1507,131 @@ func deriveFundingScriptKey(ctx context.Context, addrBook address.Storage,
...
+		}
+
+		vPackets[assetID] = pkt
+	}
+
+	localOutputs := c.LocalOutputs()
+	remoteOutputs := c.RemoteOutputs()
+
+	// We now distribute the outputs to the vPackets.
+	if len(localOutputs) > 0 {
```
ffranr, line 1547: this conditional is unnecessary when wrapping a for loop, similarly below

tapchannel/commitment.go
```diff
@@ -1505,6 +1507,131 @@ func deriveFundingScriptKey(ctx context.Context, addrBook address.Storage,
...
+		shutdownMsg.ProofDeliveryAddr.ValOpt(), func(u []byte) error {
+			var err error
+			proofDeliveryUrl, err = url.Parse(string(u))
+			return err
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode proof delivery "+
+			"address: %w", err)
+	}
```
ffranr, line 1616: you can avoid the side effect `proofDeliveryUrl` write here by returning `Result` from `MapOptionZ`. We have examples of that in our code.

tapchannel/commitment.go
```diff
@@ -1505,6 +1507,131 @@ func deriveFundingScriptKey(ctx context.Context, addrBook address.Storage,
...
+
+	return &tappsbt.VOutput{
+		Amount:       a.Amount,
+		AssetVersion: version,
+		Type: func() tappsbt.VOutputType {
+			if a.SplitCommitmentRoot == nil {
+				return tappsbt.TypeSimple
+			}
+			return tappsbt.TypeSplitRoot
+		}(),
```
ffranr, line 1626: We could just set this using a variable above without needing a closure

tapsend/allocation.go
```diff
@@ -486,19 +486,14 @@ func DistributeCoins(inputs []*proof.Proof, allocations []*Allocation,
...
 	// Each "piece" keeps track of how many assets of a specific asset ID
 	// we have already distributed. The pieces are also the main way to
 	// reference an asset ID's virtual packet.
-	pieces := make([]*piece, len(uniqueAssetIDs))
-	for i, assetID := range uniqueAssetIDs {
-		proofsByID := fn.Filter(inputs, func(i *proof.Proof) bool {
-			return i.Asset.ID() == assetID
-		})
+	pieces := make([]*piece, len(groupedProofs))
+	var idx int
```
ffranr, line 495: By using `pieces := make([]*piece, 0, len(groupedProofs))`, we can append elements without reallocations and avoid managing an explicit index `idx`.

tapchannel/commitment.go
guggero: Ah, was left over from a previous version. But you're right, not needed anymore, removed.

tapchannel/commitment.go
guggero: Makes sense, done.

ffranr (approved):
Thanks for the added context in 1:1

Issue #1460, opened 2025-04-03
[itest-flake] taproot asset channels itest flakes


ellemouton:
## Issue 1

[Link to build](https://github.com/lightninglabs/lightning-terminal/actions/runs/14243459629/job/39918534182?pr=1023)

```
    assets_test.go:1068: 
        	Error Trace:	/home/runner/work/lightning-terminal/lightning-terminal/itest/assets_test.go:1068
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/assets_test.go:851
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/litd_custom_channels_test.go:487
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/test_harness.go:103
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/litd_test.go:105
        	Error:      	Received unexpected error:
        	            	rpc error: code = DeadlineExceeded desc = context deadline exceeded
        	Test:       	TestLightningTerminal/test_custom_channels
```

possibly relevant log line in the `charlie` log:
```
[ERR] RPCS interceptor.go:624: [/tapchannelrpc.TaprootAssetChannels/SendPayment]: error adding sell order: timeout waiting for response from peer 021c0119f498b527ada7760297c8df3ce22912d9ca5bd4b0d5bca96186e955cb35
```

## Issue 2:

[Link to build](https://github.com/lightninglabs/lightning-terminal/actions/runs/14488490130/job/40639240900?pr=1039)

```
    assets_test.go:1055: 
        	Error Trace:	/home/runner/work/lightning-terminal/lightning-terminal/itest/assets_test.go:1055
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/litd_custom_channels_test.go:1997
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/test_harness.go:103
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/litd_test.go:105
        	Error:      	Error "rpc error: code = Unknown desc = error adding sell order: timeout waiting for response from peer 0215bdefbf36ac375fd9a2835c9d876127f572809da65d4fd7dc0ca67c935b442a" does not contain "rejecting payment of 20000 mSAT"
        	Test:       	TestLightningTerminal/test_custom_channels_liquidity
```

ZZiigguurraatt:
So this is random or repeatable?

PR #1506, opened 2025-05-02
rfq: store outgoing request before sending message

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1460.

We store the outgoing request in the request map, so we can interpret any response and map it to the original request. We do this before sending the message, because otherwise we could potentially receive the response while we're waiting for the atomic lock to be released, which would lead to us not knowing about the request in the first place, and we couldn't map the response.

### DIFF
--- a/rfq/stream.go
+++ b/rfq/stream.go
@@ -154,13 +154,12 @@ func (h *StreamHandler) HandleOutgoingMessage(
 	ctx, cancel := h.WithCtxQuitNoTimeout()
 	defer cancel()
 
-	err = h.cfg.PeerMessenger.SendCustomMessage(ctx, lndClientCustomMsg)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


ffranr (approved):

rfq/stream.go
```diff
@@ -154,13 +154,12 @@ func (h *StreamHandler) HandleOutgoingMessage(
...
 	defer cancel()
 
-	err = h.cfg.PeerMessenger.SendCustomMessage(ctx, lndClientCustomMsg)
-	if err != nil {
-		return fmt.Errorf("unable to send message to peer: %w",
-			err)
-	}
-
-	// Store outgoing requests.
+	// We store the outgoing request in the request map, so we can interpret
```
Roasbeef, line 157: 👍

Roasbeef (approved):
LGTM 🪙

PR #1507, opened 2025-05-03
tapdb: add implementation of supplycommit.SupplyTreeView 

Roasbeef:
In this commit, we add an implementation of the `supplycommitment.SupplyTreeView` interface to tapdb. This provides us with a concrete interface to be used along side #1464. 

The structure is similar to the existing multi-verse trees, but we instead of scoping by proof type (issuance or transfer), we instead scope by the _group key_ of a given asset. 

Along the way we do some refactors to allow us to do atomic inserts across all the sub-trees, and also the root tree.


tapdb/supply_tree.go
```diff
@@ -39,3 +54,535 @@ func subTreeNamespace(groupKey *btcec.PublicKey,
...
+	// Upsert the universe_supply_leaves entry to link the root tree,
+	// sub-tree type, and the sub-tree's namespace.
+	_, err = dbTx.UpsertUniverseSupplyLeaf(ctx, UpsertUniverseSupplyLeaf{
+		SupplyRootID:      rootID,
+		SubTreeType:       subTreeTypeStr,
+		LeafNodeKey:       leafKey[:],
+		LeafNodeNamespace: subNs,
+	})
+	if err != nil {
+		fmt.Println("sub tree: ", subTreeType.String())
```
ffranr, line 118: `fmt.Println` left over from debugging?

tapdb/sqlc/migrations/000035_supply_tree.up.sql
```diff
@@ -0,0 +1,46 @@
+-- Ensure the required proof types for supply sub-trees exist.
+-- Note: 'issuance' is assumed to exist from previous migrations.
+INSERT INTO proof_types (proof_type) VALUES ('burn'), ('ignore'), ('mint_supply')
+    ON CONFLICT (proof_type) DO NOTHING;
```
ffranr, line 4: > We also add a new proof type of mint_supply, to distinguish from the normal issuance type.

Is the new proof type (`mint_supply`) necessary here? I thought we already added supply commit specific prefixes to the relevant namespaces? If true, then I think we could just reuse the `issuance` proof type.

tapdb/burn_tree.go
```diff
@@ -31,110 +36,153 @@ func NewBurnUniverseTree(db BatchedUniverseTree) *BurnUniverseTree {
 func (bt *BurnUniverseTree) Sum(ctx context.Context,
 	spec asset.Specifier) universe.BurnTreeSum {
 
-	// Derive identifier from the asset.Specifier.
-	id, err := specifierToIdentifier(spec, universe.ProofTypeBurn)
-	if err != nil {
-		return lfn.Err[lfn.Option[uint64]](err)
+	groupKey := spec.UnwrapGroupKeyToPtr()
```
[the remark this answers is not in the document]
Roasbeef: Sure, I've changed to use this everywhere across all 3 PRs.

tapdb/burn_tree.go
```diff
@@ -31,110 +36,153 @@ func NewBurnUniverseTree(db BatchedUniverseTree) *BurnUniverseTree {
...
+		finalResults []*universe.AuthenticatedBurnLeaf
+		err          error
+	)
+	txErr := bt.db.ExecTx(ctx, &writeTx, func(db BaseUniverseStore) error {
+		finalResults, err = insertBurnsInternal(
+			ctx, db, spec, burnLeaves...,
+		)
+
+		// TODO(roasbeef): also update the root supply tree?
+		return err
```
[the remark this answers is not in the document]
Roasbeef: So this is done in the final PR of the series. We broke out these internal functions to add to each supply tree as the final PR uses these to update the root supply tree all at once. Just like we have methods to update an individual universe tree, but in practice we always update the multiverse trees afterwards. 

The reason I isolated this logic to the state machine is to ensure that via that interface, the set of supply roots are always consistent.

tapdb/ignore_tree.go
```diff
@@ -24,132 +25,159 @@ func NewIgnoreUniverseTree(db BatchedUniverseTree) *IgnoreUniverseTree {
...
-// AddTuple adds a new ignore tuples to the ignore tree.
-func (it *IgnoreUniverseTree) AddTuples(ctx context.Context,
-	spec asset.Specifier, tuples ...universe.SignedIgnoreTuple,
-) lfn.Result[universe.AuthIgnoreTuples] {
+// addTuplesInternal performs the insertion of ignore tuples within a database
+// transaction. It also updates the main supply tree with the new ignore
+// sub-tree root.
+//
+// NOTE: This function must be called within a database transaction.
+func addTuplesInternal(ctx context.Context, db BaseUniverseStore,
```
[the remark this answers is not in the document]
Roasbeef: Nice catch, I had some hold over logic in here to update the supply sub-tree, but shelved it to pursue the idea re the atomic updates.

tapdb/ignore_tree.go
```diff
@@ -24,132 +25,159 @@ func NewIgnoreUniverseTree(db BatchedUniverseTree) *IgnoreUniverseTree {
...
+
+	var (
+		writeTx      BaseUniverseStoreOptions
+		finalResults []universe.AuthenticatedIgnoreTuple
+		err          error
+	)
+	txErr := it.db.ExecTx(ctx, &writeTx, func(db BaseUniverseStore) error {
+		finalResults, err = addTuplesInternal(ctx, db, spec, tuples...)
+		return err
 	})
```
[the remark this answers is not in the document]
Roasbeef: `SupplyTreeStore.ApplySupplyUpdates` is what will update the supply roots. This is the main interface used by the state machine from the first and last PRs.

tapdb/sqlc/migrations/000035_supply_tree.up.sql
```diff
@@ -0,0 +1,46 @@
...
+
+-- Table representing the leaves within a root supply tree.
+-- Each leaf corresponds to the root of a sub-tree (mint, burn, ignore).
+CREATE TABLE universe_supply_leaves (
+    id INTEGER PRIMARY KEY,
+
+    -- Reference to the root supply tree this leaf belongs to.
+    supply_root_id BIGINT NOT NULL REFERENCES universe_supply_roots(id) ON DELETE CASCADE,
+
+    -- The type of sub-tree this leaf represents (issuance, burn, ignore).
```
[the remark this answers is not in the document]
Roasbeef: Indeed, I had it as issuance initially, but ran into a bug, related to uniqueness of leaves. I opted for a new proof type as: not all assets will have this special proof type, and even though the structure is similar, they're stored in distinct trees. A new proof type (though I agree we can use a better name here), helps to properly isolate the leaves of the two similar, but distinct trees.

tapdb/supply_tree.go
Roasbeef: Indeed 🤭, removed!

tapdb/sqlc/migrations/000035_supply_tree.up.sql
Roasbeef: So I ended up using an entirely distinct namespace schema for this set of trees. See the `rootSupplyNamespace` and `subTreeNamespace` functions. This helped me to track down some bugs where because a distinct namespace wasn't used everywhere, sometimes a leaf would end up in the wrong tree, causing a test to fail. 

I added this extra proof type to distinguish the supply tree leaf, from the normal issuance leaf. Any public asset will have an issuance leaf, but only a subset of those will have a supply tree leaf.

litbot-9000:

tapdb/sqlc/migrations/000036_supply_tree.up.sql
```diff
@@ -0,0 +1,46 @@
...
+    leaf_node_namespace VARCHAR NOT NULL,
+
+    -- Ensure each supply root has only one leaf per sub-tree type.
+    UNIQUE(supply_root_id, sub_tree_type)
+);
+
+-- Add indexes for frequent lookups.
+CREATE INDEX universe_supply_roots_group_key_idx ON universe_supply_roots(group_key);
+CREATE INDEX universe_supply_leaves_supply_root_id_idx ON universe_supply_leaves(supply_root_id);
+CREATE INDEX universe_supply_leaves_supply_root_id_type_idx ON universe_supply_leaves(supply_root_id, sub_tree_type);
```
guggero: We can just turn this into a `UNIQUE` index, then we don't need the unnamed `UNIQUE` clause above (that can't be changed because it doesn't have a name).

tapdb/sqlc/queries/supply_tree.sql
```diff
@@ -0,0 +1,55 @@
...
+WHERE r.namespace_root = @namespace_root;
+
+-- name: UpsertUniverseSupplyLeaf :one
+INSERT INTO universe_supply_leaves (
+    supply_root_id, sub_tree_type, leaf_node_key, leaf_node_namespace
+) VALUES (
+    @supply_root_id, @sub_tree_type, @leaf_node_key, @leaf_node_namespace
+)
+ON CONFLICT (supply_root_id, sub_tree_type)
+    -- This is a no-op to allow returning the ID.
```
guggero, line 26: This isn't a no-op, it updates the leaf node key and namespace.

ffranr:
This patch to the base PR might fix the CI: https://github.com/lightninglabs/taproot-assets/pull/1464#issuecomment-2984282589

tapdb/sqlc/migrations/000036_supply_tree.up.sql
Roasbeef: Makes sense, switched over.

tapdb/sqlc/queries/supply_tree.sql
Roasbeef: Ah, good catch. Modified to just set the sub tree type.

tapdb/burn_tree.go
```diff
@@ -31,110 +36,154 @@ func NewBurnUniverseTree(db BatchedUniverseTree) *BurnUniverseTree {
...
 
-		return nil
+		// TODO(roasbeef): also update the root supply tree?
+		return err
 	})
 	if txErr != nil {
 		return lfn.Err[[]*universe.AuthenticatedBurnLeaf](txErr)
 	}
 
+	// TODO(roasbeef): cache invalidation?
```
ffranr, line 185: What cache might need to be invalidated here?

tapdb/sqlc/migrations/000035_supply_tree.up.sql
ffranr: given that the supply tree namespace is prefixed with `supply-` do we still need a new proof type (`mint_supply`) for issuance?

Roasbeef: We still need to distinguish on the `proof_types` level as that table is referenced from elsewhere.

tapdb/burn_tree.go
Roasbeef: Was sort of a bread crumb during development. I was thinking of the caches that we added to make serving the RPC calls faster. As is, these new trees aren't a part of those caches at all. 

I think we can examine the performance of these new calsl after we add the RPC serving, then extend the cache at that point in a bew PR if needed.

tapdb/sqlc/migrations/000036_supply_tree.up.sql
```diff
@@ -0,0 +1,45 @@
...
+    -- The namespace within mssmt_nodes where the actual sub-tree root node resides.
+    leaf_node_namespace VARCHAR NOT NULL
+);
+
+-- Add indexes for frequent lookups.
+CREATE INDEX universe_supply_roots_group_key_idx ON universe_supply_roots(group_key);
+CREATE INDEX universe_supply_leaves_supply_root_id_idx ON universe_supply_leaves(supply_root_id);
+
+  -- Ensure each supply root has only one leaf per sub-tree type.
+CREATE UNIQUE INDEX universe_supply_leaves_supply_root_id_type_idx ON universe_supply_leaves(supply_root_id, sub_tree_type);
```
guggero: This index already covers the `supply_root_id` so we don't need the explicit `universe_supply_leaves_supply_root_id_idx` index above.

ffranr:
This patch fixes lint errors:

```
Subject: [PATCH] wallet: add method LndRpcWalletAnchor.DeriveNextKey()

Add `DeriveNextKey` method to `LndRpcWalletAnchor`, enabling it to
satisfy the `supplycommit.Wallet` interface.
---
Index: tapdb/universe.go
IDEA additional info:
<+>UTF-8
diff --git a/tapdb/universe.go b/tapdb/universe.go
--- a/tapdb/universe.go	(revision b8988d3a5a2c8f5a5b5c252698955c1b4b14f5cb)
+++ b/tapdb/universe.go	(date 1750685612847)
@@ -130,31 +130,6 @@
 		arg UpsertUniverseSupplyRoot) (int64, error)
 }
 
-// specifierToIdentifier converts an asset.Specifier into a universe.Identifier
-// for a specific proof type.
-//
-// NOTE: This makes an assumption that only specifiers with a group key are
-// valid for the ignore and burn proof types.
-func specifierToIdentifier(spec asset.Specifier,
-	proofType universe.ProofType) (universe.Identifier, error) {
-
-	var id universe.Identifier
-
-	// The specifier must have a group key to be able to be used within the
-	// ignore or burn tree context.
-	requireGroupKey := proofType == universe.ProofTypeIgnore ||
-		proofType == universe.ProofTypeBurn
-	if requireGroupKey && !spec.HasGroupPubKey() {
-		return id, fmt.Errorf("group key must be set for proof type %v",
-			proofType)
-	}
-
-	id.GroupKey = spec.UnwrapGroupKeyToPtr()
-	id.ProofType = proofType
-
-	return id, nil
-}
-
 // getUniverseTreeSum retrieves the sum of a universe tree specified by its
 // identifier.
 func getUniverseTreeSum(ctx context.Context, db BatchedUniverseTree,
Index: tapdb/supply_tree.go
IDEA additional info:
<+>UTF-8
diff --git a/tapdb/supply_tree.go b/tapdb/supply_tree.go
--- a/tapdb/supply_tree.go	(revision b8988d3a5a2c8f5a5b5c252698955c1b4b14f5cb)
+++ b/tapdb/supply_tree.go	(date 1750685675569)
@@ -278,7 +278,8 @@
 
 	groupKey, err := assetSpec.UnwrapGroupKeyOrErr()
 	if err != nil {
-		return nil, fmt.Errorf("group key must be specified for mint supply: %w", err)
+		return nil, fmt.Errorf("group key must be specified for "+
+			"mint supply: %w", err)
 	}
 	subNs := subTreeNamespace(groupKey, supplycommit.MintTreeType)
 
@@ -361,9 +362,8 @@
 
 	groupKey, err := spec.UnwrapGroupKeyOrErr()
 	if err != nil {
-		return nil, fmt.Errorf(
-			"group key must be specified for supply updates: %w", err,
-		)
+		return nil, fmt.Errorf("group key must be specified for "+
+			"supply updates: %w", err)
 	}
 
 	// Group updates by their sub-tree type, this'll simplify the logic

```

ffranr:
If helpful, I cleaned up the linter errors and rebased the branch here: https://github.com/ffranr/taproot-assets/tree/rebased-universe-supply-trees

tapdb/sqlc/migrations/000036_supply_tree.up.sql
Roasbeef: Ah that makes sense, will remove.

Roasbeef:
Merging as the unit test failure is unrelated: 
```
--- FAIL: TestBatchedAssetIssuance (86.69s)
    test_sqlite.go:11: Creating new SQLite DB handle for testing: /tmp/TestBatchedAssetIssuance1396529981/001/tmp.db
    test_sqlite.go:11: Creating new SQLite DB handle for testing: /tmp/TestBatchedAssetIssuance1396529981/002/tmp.db
    test_sqlite.go:11: Creating new SQLite DB handle for testing: /tmp/TestBatchedAssetIssuance1396529981/003/tmp.db
    --- FAIL: TestBatchedAssetIssuance/minting_with_cancellation (61.28s)
        planter_test.go:312: 
            	Error Trace:	/home/runner/work/taproot-assets/taproot-assets/tapgarden/planter_test.go:312
            	            				/home/runner/work/taproot-assets/taproot-assets/tapgarden/planter_test.go:1312
            	            				/home/runner/work/taproot-assets/taproot-assets/tapgarden/planter_test.go:2048
            	Error:      	Condition never satisfied
            	Test:       	TestBatchedAssetIssuance/minting_with_cancellation
    test_sqlite.go:11: Creating new SQLite DB handle for testing: /tmp/TestBatchedAssetIssuance1396529981/004/tmp.db
    test_sqlite.go:11: Creating new SQLite DB handle for testing: /tmp/TestBatchedAssetIssuance1396529981/005/tmp.db
    test_sqlite.go:11: Creating new SQLite DB handle for testing: /tmp/TestBatchedAssetIssuance1396529981/006/tmp.db
    test_sqlite.go:11: Creating new SQLite DB handle for testing: /tmp/TestBatchedAssetIssuance1396529981/007/tmp.db
```

PR #1508, opened 2025-05-03
tapdb: add implementation of supplycommit.CommitmentTracker and supplycommit.StateMachineStore

Roasbeef:
In this PR, we add a concrete implementation of the two fundamental interfaces that will drive the state machine implemented in #1464. 

We start with a state machine on disk for a given group key that also points to the latest supply commitment. When we want to add new elements to the supply tree, we create a new state transition, and then queue pending updates associated with this state transition. Once we've signed and broadcasted the commitment transaction, we then apply the state transition: insert all the updates into the tree, update the state machine to point to that new tree, and finalize the state transition. 

See the readme for more details.

### DIFF
--- a/tapdb/supply_commit.md
+++ b/tapdb/supply_commit.md
@@ -1,0 +1,190 @@
+# Taproot Asset Supply Commitment Persistence
+
+This document details the database schema and Go implementation (`tapdb/supply_commit.go`) that provides persistent storage for the Taproot Asset Supply Commitment State Machine (`universe/supplycommit`). This layer ensures the state machine's progress is durable across restarts and provides mechanisms for tracking historical and pending supply commitments.
+
+## Purpose
+
+The primary function of this persistence layer is to reliably store the state and associated data for the `universe/supplycommit` state machine. This state machine manages the process of creating and anchoring cryptographic commitments to the supply changes (mints, burns, ignores) of a specific asset group onto the Bitcoin blockchain.
+
+This involves storing:
+
+1.  The current operational state of the state machine for each asset group.
+2.  Details of past, successfully confirmed supply commitments.
+3.  Information about pending state transitions, including the specific supply update events being processed.
+4.  Details of the Bitcoin transactions used to anchor commitments.
+
+## Schema Overview
+
+Migration `000036_asset_commit.up.sql` introduces several tables to manage the state machine's lifecycle and data.
+
+1.  **`supply_commit_states`**: An enum-like table defining the possible states of the state machine (e.g., `DefaultState`, `UpdatesPendingState`, `CommitTxCreateState`).
+    *   `id` (INTEGER PK): Numeric ID for the state.
+    *   `state_name` (TEXT UNIQUE): Human-readable name of the state.
+
+2.  **`supply_commit_update_types`**: An enum-like table defining the types of supply updates (e.g., `mint`, `burn`, `ignore`).
+    *   `id` (INTEGER PK): Numeric ID for the update type.
+    *   `update_type_name` (TEXT UNIQUE): Human-readable name of the update type.
+
+3.  **`supply_commitments`**: Stores the details of a specific, potentially confirmed, supply commitment anchored on-chain.
+    *   `commit_id` (INTEGER PK): Unique identifier for this commitment instance.
+    *   `group_key` (BLOB): The tweaked group key identifying the asset group.
+    *   `chain_txn_id` (BIGINT FK -> `chain_txns.txn_id`): Reference to the Bitcoin transaction containing this commitment.
+    *   `output_index` (INTEGER): The output index within the `chain_txn_id` transaction.
+    *   `internal_key_id` (BIGINT FK -> `internal_keys.key_id`): The internal key used for the commitment output's Taproot derivation.
+    *   `output_key` (BLOB): The final tweaked Taproot output key.
+    *   `block_header` (BLOB): Header of the block confirming the commitment (NULL if unconfirmed).
+    *   `block_height` (INTEGER): Height of the confirming block (NULL if unconfirmed).
+    *   `merkle_proof` (BLOB): Merkle proof of the transaction's inclusion in the block (NULL if unconfirmed).
+    *   `supply_root_hash` (BLOB): The MS-SMT root hash of the supply tree at this commitment (NULL until finalized).
+    *   `supply_root_sum` (BIGINT): The MS-SMT root sum (total supply value) at this commitment (NULL until finalized).
+
+4.  **`supply_commit_state_machines`**: Tracks the current state for each asset group's state machine instance.
+    *   `group_key` (BLOB PK): The tweaked group key, uniquely identifying the state machine instance.
+    *   `current_state_id` (INTEGER FK -> `supply_commit_states.id`): The current operational state.
+    *   `latest_commitment_id` (BIGINT FK -> `supply_commitments.commit_id`): Reference to the most recently *finalized* and confirmed commitment (NULL if none).
+
+5.  **`supply_commit_transitions`**: Records an active attempt to transition the supply state. Acts like a Write-Ahead Log (WAL) entry for the state machine.
+    *   `transition_id` (INTEGER PK): Unique identifier for this transition attempt.
+    *   `state_machine_group_key` (BLOB FK -> `supply_commit_state_machines.group_key`): Links back to the state machine instance.
+    *   `old_commitment_id` (BIGINT FK -> `supply_commitments.commit_id`): The commitment being replaced (NULL for the first commitment).
+    *   `new_commitment_id` (BIGINT FK -> `supply_commitments.commit_id`): The new commitment being created by this transition (NULL initially).
+    *   `pending_commit_txn_id` (BIGINT FK -> `chain_txns.txn_id`): The Bitcoin transaction intended to confirm this transition (NULL until created/signed).
+    *   `finalized` (BOOLEAN): Indicates if the transition completed successfully and was applied. Defaults to `FALSE`.
+    *   `creation_time` (TIMESTAMP): When the transition was initiated.
+    *   `UNIQUE INDEX ... WHERE finalized = 0`: Crucially ensures only *one* non-finalized (pending) transition can exist per `state_machine_group_key` at any time.
+
+6.  **`supply_update_events`**: Stores the individual mint, burn, or ignore events associated with a *pending* transition.
+    *   `event_id` (INTEGER PK): Unique identifier for the event record.
+    *   `transition_id` (BIGINT FK -> `supply_commit_transitions.transition_id` ON DELETE CASCADE): Links to the parent transition. Cascade delete ensures events are cleaned up if a transition is aborted/deleted.
+    *   `update_type_id` (INTEGER FK -> `supply_commit_update_types.id`): Type of the update event.
+    *   `event_data` (BLOB): Serialized data specific to the event type (e.g., `NewMintEvent`, `NewBurnEvent`).
+
+7.  **`ALTER TABLE mint_anchor_uni_commitments`**: Adds a `spent_by` column (FK -> `supply_commitments.commit_id`) to track which supply commitment transaction spent a given minting pre-commitment output.
+
+### Schema Relationships (Mermaid Diagram)
+
+```mermaid
+erDiagram
+    supply_commit_state_machines ||--o{ supply_commit_transitions : "has_pending"
+    supply_commit_state_machines }|--|| supply_commit_states : "uses_state"
+    supply_commit_state_machines }|--|| supply_commitments : "tracks_latest"
+    supply_commit_transitions ||--o{ supply_update_events : "includes_events"
+    supply_commit_transitions }|--|| supply_commitments : "replaces_old"
+    supply_commit_transitions }|--|| supply_commitments : "creates_new"
+    supply_commit_transitions }|--|| chain_txns : "uses_pending_tx"
+    supply_update_events ||--|| supply_commit_update_types : "uses_type"
+    supply_commitments ||--|| chain_txns : "included_in_tx"
+    supply_commitments ||--|| internal_keys : "uses_internal_key"
+    mint_anchor_uni_commitments }|--|| supply_commitments : "spent_by_commit"
+    
+    supply_commit_states {
+        INTEGER id PK
+        TEXT state_name UK
+    }
+    
+    supply_commit_update_types {
+        INTEGER id PK
+        TEXT update_type_name UK
+    }
+    
+    supply_commitments {
+        INTEGER commit_id PK
+        BLOB group_key "Idx"
+        BIGINT chain_txn_id FK
+        INTEGER output_index
+        BIGINT internal_key_id FK
+        BLOB output_key
+        BLOB supply_root_hash
+        BIGINT supply_root_sum
+    }
+    
+    supply_commit_state_machines {
+        BLOB group_key PK
+        INTEGER current_state_id FK
+        BIGINT latest_commitment_id FK
+    }
+    
+    supply_commit_transitions {
+        INTEGER transition_id PK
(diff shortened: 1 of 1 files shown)


tapdb/sqlc/migrations/000036_asset_commit.up.sql
```diff
@@ -0,0 +1,133 @@
...
+    -- Timestamp when this transition was initiated.
+    creation_time TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+-- Table storing individual update events associated with a pending transition.
+CREATE TABLE supply_update_events (
+    event_id INTEGER PRIMARY KEY,
+
+    -- Reference to the state transition this event is part of.
+    transition_id BIGINT NOT NULL REFERENCES supply_commit_transitions(transition_id) ON DELETE CASCADE,
```
ffranr, line 110: an update event can't be persisted unless it references an existing transition, is that correct?
If so, then if we're waiting for tx conf then new update events can't be persisted?

If true, then I don't think it matters a great deal, but from a UX perspective, we will need to return an error to the user along the lines of: unconfirmed supply tx, can't accept new asset ignore right now.

tapdb/sqlc/migrations/000036_asset_commit.up.sql
```diff
@@ -0,0 +1,133 @@
...
+
+    -- The chain transaction that, once confirmed, will finalize this transition.
+    -- Can be NULL until the transaction is created and signed.
+    pending_commit_txn_id BIGINT REFERENCES chain_txns(txn_id),
+
+    -- Indicates if this transition has been successfully completed and committed.
+    finalized BOOLEAN NOT NULL DEFAULT FALSE,
+
+    -- Timestamp when this transition was initiated.
+    creation_time TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
```
guggero: Does `CURRENT_TIMESTAMP` work with `sqlc` which parses the string into a `time.Time`?

tapdb/sqlc/queries/supply_commit.sql
```diff
@@ -0,0 +1,191 @@
...
+    SELECT id
+    FROM supply_commit_states s2
+    WHERE s2.state_name = 'DefaultState'
+)
+INSERT INTO supply_commit_state_machines (
+    group_key, current_state_id, latest_commitment_id
+) VALUES (
+    @group_key,
+    -- Use the target state ID if found, otherwise use the default state ID.
+    coalesce((SELECT id FROM target_state), (SELECT id FROM default_state)),
```
guggero, line 18: Hmm, I'm a bit surprised this works. So the actual value returned from a sub query is `NULL` if there was no row found?

tapdb/supply_commit.go
```diff
@@ -0,0 +1,1303 @@
...
+	// FetchChainTxByID fetches a chain transaction by its primary key ID.
+	FetchChainTxByID(ctx context.Context,
+		txnID int64) (FetchChainTxByIDRow, error)
+
+	// FetchUniverseSupplyRoot fetches the root hash and sum for a supply
+	// tree namespace.
+	FetchUniverseSupplyRoot(ctx context.Context,
+		namespaceRoot string) (FetchUniverseSupplyRootRow, error)
+
+	TreeStore
```
guggero: Should we move the embedded interfaces to the top to make it more apparent and align with other such embedded interfaces?

tapdb/supply_commit.go
```diff
@@ -0,0 +1,1303 @@
...
+}
+
+// UnspentPrecommits returns the set of unspent pre-commitments for a given
+// asset spec. The asset spec will only specify a group key, and not also an
+// asset ID.
+func (s *SupplyCommitMachine) UnspentPrecommits(ctx context.Context,
+	assetSpec asset.Specifier) lfn.Result[supplycommit.PreCommits] {
+
+	groupKey := assetSpec.UnwrapGroupKeyToPtr()
+	if groupKey == nil {
```
guggero, line 225: nit: `UnwrapGroupKeyOrErr()`? Not that it shortens the code by any meaningful amount, just a bit more explicit.

tapdb/supply_commit.go
```diff
@@ -0,0 +1,1303 @@
...
+	}
+	groupKeyBytes := groupKey.SerializeCompressed()
+
+	var preCommits supplycommit.PreCommits
+	readTx := NewSupplyCommitReadTx()
+	dbErr := s.db.ExecTx(ctx, &readTx, func(db SupplyCommitStore) error {
+		rows, err := db.FetchUnspentPrecommits(ctx, groupKeyBytes)
+		if err != nil {
+			// It's okay if there are no unspent pre-commits.
+			if err == sql.ErrNoRows {
```
guggero: Should use `errors.Is()`.

guggero: Same everywhere below.

tapdb/supply_commit.go
guggero: Or, I guess since we're only ever using the group key part of the specifier, why don't we change the param to be a public key directly? Just to have some semantic meaning around what the key is?

tapdb/supply_commit.go
```diff
@@ -0,0 +1,1303 @@
...
+		// Construct the root node directly from the stored hash and
+		// sum. Handle potential NULL values if the root wasn't set yet
+		// (though FetchSupplyCommit filters for confirmed TX, so it
+		// should be set).
+		var (
+			rootHash mssmt.NodeHash
+			rootSum  uint64
+			rootNode *mssmt.BranchNode
+		)
+		if row.RootHash != nil && row.RootSum.Valid {
```
guggero: Check length of root hash instead to avoid empty slice giving false negative?

tapdb/supply_commit.go
```diff
@@ -0,0 +1,1303 @@
...
+				"update in state: %s", currentStateName)
+		}
+
+		// Now that we know the state machine is in the proper state,
+		// we'll fetch the transition ID, which will be needed below.
+		var transitionID int64
+		existingTransitionID, err := db.QueryExistingPendingTransition(
+			ctx, groupKeyBytes,
+		)
+		if err != nil {
```
guggero: Feels like this whole block could be simplified with a `switch`, instead of the `else` cases.

tapdb/supply_commit.go
```diff
@@ -0,0 +1,1303 @@
...
+		}
+
+		// Insert the new commitment record. Chain details (block
+		// height, header, proof, output index) are NULL at this stage.
+		//nolint:lll
+		newCommitmentID, err := db.InsertSupplyCommitment(ctx, sqlc.InsertSupplyCommitmentParams{
+			GroupKey:       groupKeyBytes,
+			ChainTxnID:     chainTxID,
+			InternalKeyID:  internalKeyID,
+			OutputKey:      outputKey.SerializeCompressed(), //nolint:lll
```
guggero: nit: nolint comment not needed if already above block.

tapdb/supply_commit.go
```diff
@@ -0,0 +1,1303 @@
...
+		}
+		return nil
+	})
+}
+
+// fetchCommitment is a helper to fetch and reconstruct a RootCommitment and
+// its associated chain confirmation details.
+func fetchCommitment(ctx context.Context, db SupplyCommitStore,
+	commitID sql.NullInt64, groupKeyBytes []byte,
+) (lfn.Option[supplycommit.RootCommitment], lfn.Option[commitmentChainInfo], error) { //nolint:lll
```
guggero: Could just break between the return values instead of using the nolint comment?

tapdb/supply_commit.go
```diff
@@ -0,0 +1,1303 @@
...
+
+	if !commitID.Valid {
+		return lfn.None[supplycommit.RootCommitment](),
```
guggero: We could create those `lfn.None[supplycommit.RootCommitment]()` and `lfn.None[commitmentChainInfo]()` as variables above, since they're used so many times across this function.

tapdb/supply_commit.go
```diff
@@ -0,0 +1,1303 @@
...
+			lfn.None[commitmentChainInfo](),
+			fmt.Errorf("failed to deserialize commit tx "+
+				"for commit %d: %w", commitID.Int64, err)
+	}
+
+	// Construct the SMT root node from the stored hash and sum. If they are
+	// NULL (e.g., initial commit before ApplyStateTransition ran), use the
+	// empty root.
+	var rootNode *mssmt.BranchNode
+	if commit.SupplyRootHash == nil || !commit.SupplyRootSum.Valid {
```
guggero, line 863: Check length instead of nil?

tapdb/supply_commit.go
guggero: Does the root commitment here really need to be an option? Doesn't look like it's ever `None` in a non-error case.

tapdb/supply_commit_test.go
```diff
@@ -0,0 +1,1793 @@
...
+
+	_, transitionOpt, err := h.commitMachine.FetchState(h.ctx, h.assetSpec)
+	require.NoError(h.t, err)
+	return transitionOpt
+}
+
+// assertPendingTransitionExists asserts that a pending (non-finalized)
+// transition exists.
+//
+//nolint:lll
```
guggero, line 270: We might as well add this to the top of the file so it counts for its entirety :sweat_smile:

tapdb/supply_commit.md
```diff
@@ -0,0 +1,190 @@
...
+        INTEGER id PK
+        BIGINT batch_id FK
+        INTEGER tx_output_index
+        BLOB taproot_internal_key
+        BLOB group_key
+        BIGINT spent_by FK
+    }
+```
+
+## State Machine Persistence Logic
```
guggero, line 148: Very nice doc!

tapdb/sqlc/migrations/000036_asset_commit.up.sql
Roasbeef: That's a great insight. As of right now, if everything is spun up together and a user attempts to add a new update _while_ we're creating a new commitment on-disk, then this'll fail. 

I think we can fix this just by removing the `NOT NULL` here. We'll need some additional follow ups in the state machine though:
  * Any time an update comes through in a state, attempt to insert it to disk w/ a nil `transition_id`. 
  * Once we circle back to the default state, then read all the "dangling" update from disk, and assign them to a state transition (in a single db txn). 

Alternatively, we can create some other table, then re-insert them in the default state. 

I can either do this in another PR (as it affects this one, and the next), or do the incremental change here. Which is preferred?

tapdb/sqlc/migrations/000036_asset_commit.up.sql
Roasbeef: Good q, I didn't think anything of it at the time, but I can see how there could be some weird interaction here with the way `sqlc` handle strings/timestamps. 

I've added an additional assertion to `assertTransitionInitialState` that makes sure the timestamp is set properly, and that it's "close" to what we'd get from `time.Now()`, Also TIL that [`require.WithinDuration`](https://pkg.go.dev/github.com/stretchr/testify/require#WithinDuration) exists.

tapdb/sqlc/queries/supply_commit.sql
Roasbeef: Hehe, yeah I went through a few versions of this. I had one which used a CTE as above as part of the update (so a CTE that inserts, then returns a value), but `sqlite` doesn't allow CTE statements in insert/update like say postgres does. 

To answer the question, we basically want either the ID of the target state, or the default if a target isn't set. So on the caller side, we just don't specify the `state_name` field if we want the default state to be inserted. If no state is set, then `target_state` has no rows, so trying to select on it returns nothing. In that case the `default_state` query will _always_ return a value (as we set up the table as such in the migration).

tapdb/supply_commit.go
Roasbeef: Switched everything over to use `UnwrapGroupKeyOrErr`. 

I used `asset.Specifier` as most other similar interfaces across the codebase are starting to take that value, to force an author/writer to make sure there's logic to handle the various combos.

tapdb/supply_commit.go
Roasbeef: SGTM.

tapdb/supply_commit.go
Roasbeef: Fixed!

tapdb/supply_commit.go
Roasbeef: done

tapdb/supply_commit.go
Roasbeef: Nice suggestion, was also able to get rid of some of the linter overrides as we've eliminated a layer of nesting.

tapdb/supply_commit.go
Roasbeef: I ended up just folding it in. Usually I reach for a new struct once a function cross 2 return args (ignoring the error). Adding an extra layer of indirection to avoid a linting comment doesn't seem like a worthy tradeoff to me.

Roasbeef: I have it as an option as in `supplycommit.SupplyStateTransition`, it's an option. It's an option here as the very first time we go to create a state transition, `OldCommitment` won't yet exist. 

We can return a value of `None` in the very fist error case: if we query and it doesn't exist (`ErrNoRows`), then `noneRootCommit` with a `nil` error is returned.

tapdb/supply_commit_test.go
Roasbeef: Heh, I prefer to just sprinkle it as needed. The linter comments don't really both me when reading/reviewing code. 

In this instance, I added it as there's no good place to split the line w/o it looking awkward. I'd rather just throw this on and move along instead of spending additional time on the more marginal aspects of our code style.

litbot-9000:

Roasbeef:
Changed the base to target main.

ffranr:
@Roasbeef here's a rebased version of this PR with additional fixups to address the unit test failures and SQL migration conflict:
https://github.com/ffranr/taproot-assets/commits/wip/1534-ignore-tree/supply-commit-machine/

tapdb/sqlc/migrations/000039_asset_commit.up.sql
```diff
@@ -0,0 +1,133 @@
+-- Enum-like table for state machine states.
```
guggero: Needs to be named with `000040_...` prefix now.

tapdb/sqlc/queries/assets.sql
```diff
@@ -1052,26 +1052,27 @@ WITH target_batch AS (
...
 )
 INSERT INTO mint_anchor_uni_commitments (
-    batch_id, tx_output_index, taproot_internal_key_id, group_key
+    batch_id, tx_output_index, taproot_internal_key_id, group_key, spent_by
 )
 VALUES (
-    (SELECT batch_id FROM target_batch), @tx_output_index,
-    @taproot_internal_key_id, @group_key
+    (SELECT batch_id FROM target_batch), @tx_output_index, 
+    @taproot_internal_key_id, @group_key, sqlc.narg('spent-by')
```
guggero: nit: this is probably resulting in the same Golang variable name, but I think we should consistently use `_` characters for word delimiters.

tapdb/supply_commit.go
```diff
@@ -0,0 +1,1290 @@
...
+
+	// FetchUniverseSupplyRoot fetches the root hash and sum for a supply
+	// tree namespace.
+	FetchUniverseSupplyRoot(ctx context.Context,
+		namespaceRoot string) (FetchUniverseSupplyRootRow, error)
+}
+
+// SupplyCommitTxOptions defines the set of functional options that can be used
+// to control the transaction behavior of the SupplyCommitMachine.
+type SupplyCommitTxOptions struct {
```
guggero: nit: could use the generic options that exist now.

PR #1511, opened 2025-05-05
build: bump version to v0.6 rc1

Roasbeef:


guggero (approved):

guggero:
Just a heads up that we have a slightly different tag naming scheme in the `tapd` repo than everywhere else. The tag name is also validated by the CI release process, so the release binaries won't be created if the tag name is incorrect. So best to use the `make release-tag` command to create the tag (which also verifies it is made on the correct commit).

guggero (approved):

PR #1512, opened 2025-05-06
[group key addrs 2/?]: internal/ecies: add encrypt/decrypt with ECIES

guggero:
Implements the ECIES encryption scheme as mentioned here: https://github.com/lightninglabs/taproot-assets/issues/874#issuecomment-2752783389

This commit creates a simple encryption and decryption function that uses AES256-GCM for tne symmetric encryption and HKDF with SHA256 for the key derivation.
The shared key generation is not part of these functions, because we'll need to use lnd RPCs in some cases to be able to derive it.

### DIFF
--- a/internal/ecies/ecies.go
+++ b/internal/ecies/ecies.go
@@ -1,0 +1,144 @@
+// This package implements an ECIES (Elliptic Curve Integrated Encryption
+// Scheme) encryption. It uses AES256-GCM for encryption and HKDF with SHA256
+// for key derivation. The package provides functions to encrypt and decrypt
+// messages using a shared secret derived between two parties using ECDH
+// (Elliptic Curve Diffie-Hellman).
+// Inspiration for parts of the code in this package was taken from
+// https://github.com/ecies/go.
+
+package ecies
+
+import (
+	"bytes"
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"crypto/sha256"
+	"fmt"
+	"io"
+
+	"golang.org/x/crypto/hkdf"
+)
+
+const (
+	// GCMNonceSize is the size of the nonce used in AES-GCM mode.
+	GCMNonceSize = 16
+)
+
+// EncryptSha256Aes256 encrypts the given message using AES256-GCM with a shared
+// secret (usually derived using ECDH between the sender's ephemeral key and the
+// receiver's public key) that is stretched using HKDF with SHA256. The output
+// is a byte slice containing:
+//
+//	<16 bytes nonce> <16 bytes tag> <... bytes ciphertext>
+func EncryptSha256Aes256(sharedSecret [32]byte, msg []byte) ([]byte, error) {
+	// We begin by stretching the shared secret using HKDF with SHA256.
+	stretchedKey, err := HkdfSha256(sharedSecret[:])
+	if err != nil {
+		return nil, fmt.Errorf("cannot derive hkdf key: %w", err)
+	}
+
+	// Then we create a new AES block cipher using the stretched key. With
+	// the stretched key length of 32 bytes, this will be AES256.
+	block, err := aes.NewCipher(stretchedKey[:])
+	if err != nil {
+		return nil, fmt.Errorf("cannot create new aes block: %w", err)
+	}
+
+	nonce := make([]byte, GCMNonceSize)
+	if _, err := rand.Read(nonce); err != nil {
+		return nil, fmt.Errorf("cannot read random bytes for nonce: "+
+			"%w", err)
+	}
+
+	// The buffer will contain the nonce, tag and ciphertext. We preallocate
+	// the buffer to the final size to avoid multiple allocations.
+	var result bytes.Buffer
+	result.Write(nonce)
+
+	gcm, err := cipher.NewGCMWithNonceSize(block, GCMNonceSize)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create aes gcm: %w", err)
+	}
+
+	ciphertext := gcm.Seal(nil, nonce, msg, nil)
+
+	// The tag is the last 16 bytes of the ciphertext. We want to move it
+	// to before the ciphertext, so we extract it and write it into the
+	// result buffer first.
+	tag := ciphertext[len(ciphertext)-gcm.NonceSize():]
+	result.Write(tag)
+	ciphertext = ciphertext[:len(ciphertext)-len(tag)]
+	result.Write(ciphertext)
+
+	return result.Bytes(), nil
+}
+
+// DecryptSha256Aes256 decrypts the given ciphertext using AES256-GCM with a
+// shared secret (usually derived using ECDH between the sender's ephemeral key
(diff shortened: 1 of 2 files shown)


internal/ecies/ecies_test.go
```diff
@@ -0,0 +1,187 @@
...
+func FuzzEncryptSha256Aes256(f *testing.F) {
+	f.Fuzz(func(t *testing.T, secretBytes []byte, msg []byte) {
+		var sharedSecret [32]byte
+		copy(sharedSecret[:], secretBytes)
+		_, _ = EncryptSha256Aes256(sharedSecret, msg)
+	})
+}
+
+// FuzzDecryptSha256Aes256 is a fuzz test for the DecryptSha256Aes256 function.
+func FuzzDecryptSha256Aes256(f *testing.F) {
```
Roasbeef, line 160: 👍

internal/ecies/ecies.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, fmt.Errorf("cannot read random bytes for nonce: "+
+			"%w", err)
+	}
+
+	// The buffer will contain the nonce, tag and ciphertext. We preallocate
+	// the buffer to the final size to avoid multiple allocations.
+	var result bytes.Buffer
+	result.Write(nonce)
+
+	gcm, err := cipher.NewGCMWithNonceSize(block, GCMNonceSize)
```
Roasbeef, line 59: Any particular reason to use this over `NewGCM`? 12-byte nonces are standardized, it's also what [the underlying Go packages uses](https://cs.opensource.google/go/go/+/refs/tags/go1.24.3:src/crypto/cipher/gcm.go;l=19) with the default constructor.

Roasbeef: See section 5.2.1 of this PDF: https://nvlpubs.nist.gov/nistpubs/legacy/sp/nistspecialpublication800-38d.pdf

> For IVs, it is recommended that implementations restrict support to the length of 96 bits, to
promote interoperability, efficiency, and simplicity of design. 

IV here is basically a nonce.

Roasbeef: One alternative crypto system to consider is `chacha20poly1305`: https://pkg.go.dev/golang.org/x/crypto/chacha20poly1305#New. This is what we use for encryption in LN. It also natively supports adding associative data (plaintext data included in the MAC).

guggero: Hmm, I just used an example I found somewhere.
But I like ChaCha20Poly1305 better, changed the code to use that. We'll use the sender's ephemeral public key as the additional data, to authenticate the ciphertext. It will be part of the outer (authmailbox message) envelope, not the ciphertext itself. Does that make sense?

gijswijs (commented):
Made some small comments throughout.

internal/ecies/ecies.go
```diff
@@ -0,0 +1,136 @@
...
+	// We can now create a new ChaCha20Poly1305 AEAD cipher using the
+	// stretched key.
+	aead, err := chacha20poly1305.New(stretchedKey[:])
+	if err != nil {
+		return nil, fmt.Errorf("cannot create new chacha20poly1305 "+
+			"cipher: %w", err)
+	}
+
+	// Select a random nonce, and leave capacity for the ciphertext.
+	nonce := make(
```
Roasbeef, line 45: As we're using a [random nonce here](https://en.wikipedia.org/wiki/ChaCha20-Poly1305#:~:text=%5Bedit%5D-,XChaCha20%2DPoly1305%20%E2%80%93%20extended%20nonce%20variant,-%5Bedit%5D), we should use [`NewX`](https://pkg.go.dev/golang.org/x/crypto/chacha20poly1305#NewX), which is meant for cases where a counter-like nonce isn't used. Basically some extra security margin for when nonces are generated randomly.

internal/ecies/ecies.go
```diff
@@ -0,0 +1,136 @@
...
+		[]byte, aead.NonceSize(),
+		aead.NonceSize()+len(msg)+aead.Overhead(),
+	)
+
+	if _, err := crand.Read(nonce); err != nil {
+		return nil, fmt.Errorf("cannot read random nonce: %w", err)
+	}
+
+	ciphertext := aead.Seal(nonce, nonce, msg, additionalData)
```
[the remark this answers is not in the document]
Roasbeef: Ah nvm we're appending the cipher text to the nonce here.

Roasbeef (changes requested):
Looking pretty good!

internal/ecies/ecies.go
guggero: Ah, yes, makes sense. Changed to X.

gijswijs (approved):
LGTM!

Should we be worried about the `TestLightningTerminal/custom_channels_multi_rfq (119.50s)` litd itest failing?

guggero:
> Should we be worried about the `TestLightningTerminal/custom_channels_multi_rfq (119.50s)` litd itest failing?

No, that looks like a flake. Re-triggered it.

litbot-9000:
@roasbeef: review reminder

guggero:
@Roasbeef friendly review ping.

internal/ecies/ecies.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, fmt.Errorf("additional data too long: %d bytes "+
+			"given, 255 bytes maximum", len(additionalData))
+	}
+
+	// We begin by hardening the shared secret against brute forcing by
+	// using HKDF with SHA256.
+	stretchedKey, err := HkdfSha256(sharedSecret[:], []byte(protocolName))
+	if err != nil {
+		return nil, fmt.Errorf("cannot derive hkdf key: %w", err)
+	}
```
ffranr, line 49: Does this actually add "hardening"? I think that HKDF with a constant salt doesn’t add brute-force hardening.

I wander if we can't just use the nonce below as the salt here. And then `protocolName` can be the `info` arg in `HkdfSha256` 's `hkdf.New` call.

If we use the nonce as the salt then `Encrypt...` and `Decrypt...` will then have access to the same (random) salt and the serialization format doesn't change.

Roasbeef: At a high level, you can view it as just _binding_ the shared secret we create to our particular context (eg: if we change the protocol name, for the same shared secret we get a diff stretched key). The [Noise Protocol ](http://www.noiseprotocol.org/noise.html#:~:text=InitializeSymmetric(protocol_name),ck%20%3D%20h.) does something similar to create an initial hash accumulator value which gets mixed into the initial shared secrets.

Roasbeef (approved):
LGTM 🎊

internal/ecies/ecies.go
ffranr: Yes, I see this as just a form of binding as well. I think we should update the comments in `ecies.go` to clarify that this isn't providing hardening, but rather serving as a binding mechanism.

Thanks for the link!

guggero: IMO it does harden against brute force because you need to use more CPU cycles per shared secret you want to try. But I guess in this context that's not really relevant as you'd attack the encryption in different manners than doing brute force.

PR #1514, opened 2025-05-06
itest flake: fixes flake in burn test

guggero:
Fixes a flake in the integration test that happens since we refactored the burn funding logic to use the allocation code for creating virtual packets.


tapsend/allocation.go
```diff
@@ -389,27 +391,29 @@ func sortPiecesWithProofs(pieces []*piece) {
...
+		// Sort amounts in reverse order so that the largest amounts are
+		// first.
+		cmp.Compare(j.Amount, i.Amount),
+		bytes.Compare(
+			i.ScriptKey.PubKey.SerializeCompressed(),
+			j.ScriptKey.PubKey.SerializeCompressed(),
+		),
+	)
+}
+
```
ffranr, line 416: You could generalise the function name and docstring to `AssetSort`, as it just sorts assets. IMO since it's a pure function, there's no need to explain its usage in the doc.

tapsend/allocation.go
```diff
@@ -509,6 +509,14 @@ func DistributeCoins(inputs []*proof.Proof, allocations []*Allocation,
 			},
 		)
 
+		// Before we create the virtual packet, we need to sort the
+		// proofs by amount and then by script key. This is important to
+		// ensure that the proofs are sorted in a deterministic way,
+		// before we assign them to the inputs of the virtual packet.
+		slices.SortFunc(proofsByID, func(i, j *proof.Proof) int {
+			return InputSort(i.Asset, j.Asset)
+		})
```
ffranr, line 518: How about:
```
// Before creating the virtual packet, sort the proofs by amount,
// then by script key. This ensures deterministic ordering before
// assigning them to the virtual packet inputs.
```

ffranr (approved):
nice :+1:

tapsend/allocation.go
guggero: The way we sort the assets is kind of specific to inputs. And the idea is that we should use this same function everywhere we deal with inputs, that why I wanted to have the word "Input" in there. I renamed to `AssetSortForInputs`, hope that makes things more clear.

tapsend/allocation_test.go
```diff
@@ -1230,3 +1230,139 @@ func TestAllocationsFromTemplate(t *testing.T) {
 		})
 	}
 }
+
+// TestSortPiecesWithProofs tests that we can sort pieces of assets with their
+// proofs. The sorting is done first by asset ID, then by the amount of the
+// proofs in descending order.
```
gijswijs, line 1236:
```suggestion
// proofs in descending order and then by script key.
```

tapsend/allocation.go
```diff
@@ -505,6 +509,14 @@ func DistributeCoins(inputs []*proof.Proof, allocations []*Allocation,
 			},
 		)
 
+		// Before we create the virtual packet, we need to sort the
+		// proofs by amount and then by script key. This is important to
+		// ensure that the proofs are sorted in a deterministic way,
+		// before we assign them to the inputs of the virtual packet.
+		slices.SortFunc(proofsByID, func(i, j *proof.Proof) int {
```
gijswijs, line 516: Not a comment, but just a confirmation that I'm understanding this PR.
This change (combined with the sorting in `wallet.go`) is the _actual_ fix right?

And you randomize the order of the assets in the unit test to check this is actually working.

:+1:

gijswijs (approved):
One ultra nit suggestion, one clarifying question.

Apart from that, LGTM!

tapsend/allocation.go
guggero: Correct, this is the fix.
The change in the unit test shows that it fixes things, as we now need to re-order the expected inputs to be descending in their amounts (each tranche has a larger amount than the previous one, so with the new sort applied, the tranches with the higher number is expected to come first).

Issue #1516, opened 2025-05-06
[bug]: can't mint second tranch of a grouped asset using the cold minting workflow


ZZiigguurraatt:
I'm trying to mint a second tranch of a grouped asset with the cold minting workflow. I'm getting

```
unable to mint asset: can't sign with group key 024aafcf69f2bc0b4aca4424591c8876de4df907f6a328a95d852b32dbce1dc14c
```

when I try to run `MintAsset`.

This issue is not caught in itest because of https://github.com/lightninglabs/taproot-assets/issues/1515 .

Issue #1515, opened 2025-05-06
[bug]: itest: testMintExternalGroupKeyChantools mints two assets in two different groups using the same private key instead of two different assets in the same group


ZZiigguurraatt:
`testMintExternalGroupKeyChantools` mints two assets in two different groups using the same private key instead of two different assets in the same group.

Here https://github.com/lightninglabs/taproot-assets/blob/52257227c8f87e2f0b7bde0bdc2b9af65228b058/itest/assets_test.go#L67 `NewGroupedAsset: true` but then nowhere in https://github.com/lightninglabs/taproot-assets/blame/52257227c8f87e2f0b7bde0bdc2b9af65228b058/itest/mint_fund_seal_test.go#L573-L587 is `GroupedAsset: true` which is what I would think you need to use according to https://github.com/lightninglabs/taproot-assets/blob/52257227c8f87e2f0b7bde0bdc2b9af65228b058/taprpc/mintrpc/mint.proto#L153-L165 for an existing asset.

The same `ExternalGroupKey` is used, but that is the signing key from the hardware wallet, not the same `group_key`, which I think uses the key from the hardware wallet and the UTXO to make it unique. I think you still need to use the same `ExternalGroupKey`, but you should not use `GroupedAsset: true` and you should need to pass a `group_key`.

PR #1517, opened 2025-05-06
minting: fix re-issuing into existing group with external key

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1515.
Fixes https://github.com/lightninglabs/taproot-assets/issues/1516.

### DIFF
--- a/itest/mint_fund_seal_test.go
+++ b/itest/mint_fund_seal_test.go
@@ -577,6 +577,9 @@ func testMintExternalGroupKeyChantools(t *harnessTest) {
 	mintReq2 := CopyRequest(issuableAssets[0])
 	mintReq2.Asset.Name = "itestbuxx-money-printer-brrr-tranche-2"
 	mintReq2.Asset.ExternalGroupKey = externalGroupKey
+	mintReq2.Asset.GroupedAsset = true
+	mintReq2.Asset.NewGroupedAsset = false
+	mintReq2.Asset.GroupKey = batchAssets[0].AssetGroup.TweakedGroupKey
 
 	assetReqs2 := []*mintrpc.MintAssetRequest{mintReq2}
 
--- a/tapgarden/planter.go
+++ b/tapgarden/planter.go
@@ -2668,7 +2668,11 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
 	// If a group internal key or tapscript root is specified, emission must
 	// also be enabled.
 	if !req.EnableEmission {
-		if req.GroupInternalKey != nil {
+		// For re-issuance or regular assets, the group internal key
+		// shouldn't be set. It is, however, set for re-issuance with
+		// an external key, because the internal group key is the key
+		// we compare the external key against.
+		if req.GroupInternalKey != nil && req.ExternalKey.IsNone() {
 			return fmt.Errorf("cannot specify group internal key " +
 				"without enabling emission")
 		}
--- a/tapgarden/seedling.go
+++ b/tapgarden/seedling.go
@@ -183,10 +183,49 @@ func (c Seedling) validateFields() error {
 func (c Seedling) validateGroupKey(group asset.AssetGroup,
 	anchorMeta *proof.MetaReveal) error {
 
-	// We must be able to sign with the group key.
-	if !group.GroupKey.IsLocal() {
-		groupKeyBytes := c.GroupInfo.GroupPubKey.SerializeCompressed()
-		return fmt.Errorf("can't sign with group key %x", groupKeyBytes)
+	switch {
+	// If we have an external key, we need to check that the group key
+	// matches the external key.
+	case c.ExternalKey.IsSome():
+		err := fn.MapOptionZ(
+			c.ExternalKey, func(extKey asset.ExternalKey) error {
+				if group.GroupKey == nil {
+					return fmt.Errorf("group key is nil")
+				}
+
+				if group.GroupKey.RawKey.PubKey == nil {
+					return fmt.Errorf("group raw key is " +
+						"nil")
+				}
+
+				pk, err := extKey.PubKey()
+				if err != nil {
+					return fmt.Errorf("error getting "+
+						"external key: %w", err)
+				}
+
+				if !pk.IsEqual(group.RawKey.PubKey) {
+					return fmt.Errorf("external key " +
+						"does not match group key")
+				}
+
+				return nil
+			},
+		)
+		if err != nil {
+			return fmt.Errorf("error validating external key: %w",
+				err)
+		}
+
+	// If it's not an external key, we need to check that we can actually
+	// sign with the group key.
+	default:
+		if !group.GroupKey.IsLocal() {
+			groupPubKey := c.GroupInfo.GroupPubKey
+			groupKeyBytes := groupPubKey.SerializeCompressed()
+			return fmt.Errorf("can't sign with group key %x",
+				groupKeyBytes)
+		}
 	}
 
 	// The seedling asset type must match the group asset type.


ZZiigguurraatt:
are you able to put these commits on top of https://github.com/lightninglabs/taproot-assets/pull/1498 as that is what I'm currently testing with ?

is that going to work okay as long as you merge https://github.com/lightninglabs/taproot-assets/pull/1498 first?

tapgarden/seedling.go
```diff
@@ -183,10 +183,49 @@ func (c Seedling) validateFields() error {
...
+	// matches the external key.
+	case c.ExternalKey.IsSome():
+		err := fn.MapOptionZ(
+			c.ExternalKey, func(extKey asset.ExternalKey) error {
+				if group.GroupKey == nil {
+					return fmt.Errorf("group key is nil")
+				}
+
+				if group.GroupKey.RawKey.PubKey == nil {
+					return fmt.Errorf("group raw key is " +
```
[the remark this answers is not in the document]
guggero: No, this is the `pub_key` in `group_internal_key`.

tapgarden/seedling.go
```diff
@@ -183,10 +183,49 @@ func (c Seedling) validateFields() error {
...
-		groupKeyBytes := c.GroupInfo.GroupPubKey.SerializeCompressed()
-		return fmt.Errorf("can't sign with group key %x", groupKeyBytes)
+	switch {
+	// If we have an external key, we need to check that the group key
+	// matches the external key.
+	case c.ExternalKey.IsSome():
+		err := fn.MapOptionZ(
+			c.ExternalKey, func(extKey asset.ExternalKey) error {
+				if group.GroupKey == nil {
+					return fmt.Errorf("group key is nil")
```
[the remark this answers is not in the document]
guggero: No, this is just the outer struct holding the info, it doesn't have an RPC couterpart. This error should never happen, it's just defensive to avoid a panic.

tapgarden/seedling.go
```diff
@@ -183,10 +183,49 @@ func (c Seedling) validateFields() error {
...
+
+				return nil
+			},
+		)
+		if err != nil {
+			return fmt.Errorf("error validating external key: %w",
+				err)
+		}
+
+	// If it's not an external key, we need to check that we can actually
```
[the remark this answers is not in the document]
guggero: Not necessarily. You only need to specify the `group_internal_key` when you _start_ a new group and want to tell `tapd` what internal key (the "raw" keypair from `lnd`) should be used for the group key construction.
When you mint into an existing group, then the internal key is already defined. So here we check that we can actually sign with that key (meaning the internal key defined in the initial mint is actually owned by the connected `lnd` node).

tapgarden/planter.go
```diff
@@ -2668,7 +2668,11 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
 	// If a group internal key or tapscript root is specified, emission must
 	// also be enabled.
 	if !req.EnableEmission {
-		if req.GroupInternalKey != nil {
+		// For re-issuance or regular assets, the group internal key
```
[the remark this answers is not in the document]
guggero: No, I did mean "or". Restructured the sentence to make it more clear.

tapgarden/planter.go
```diff
@@ -2668,7 +2668,11 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
...
+		// shouldn't be set. It is, however, set for re-issuance with
+		// an external key, because the internal group key is the key
+		// we compare the external key against.
+		if req.GroupInternalKey != nil && req.ExternalKey.IsNone() {
 			return fmt.Errorf("cannot specify group internal key " +
 				"without enabling emission")
```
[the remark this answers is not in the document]
guggero: We renamed `EnableEmission` to `NewGroupedAsset` and forgot to update some of the error messages. I did that in a new commit.

tapgarden/planter.go
```diff
@@ -2668,7 +2668,11 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
 	// If a group internal key or tapscript root is specified, emission must
 	// also be enabled.
 	if !req.EnableEmission {
-		if req.GroupInternalKey != nil {
+		// For re-issuance or regular assets, the group internal key
+		// shouldn't be set. It is, however, set for re-issuance with
+		// an external key, because the internal group key is the key
```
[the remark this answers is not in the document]
guggero: It's an internal thing about what key must be set in the request struct when we get to this point. Shouldn't matter at the user level.

ZZiigguurraatt:
I'm able to get this PR to work.

One additional thing I've now tested is the `group_anchor` option. Does that only work to mint two tranches in the same batch (without having n `group_key` first)? I'm trying to use it in a second batch, but I get

```
unable to mint asset: batch empty, group anchor Asset1 invalid
```

If it is only for minting two tranches in the same batch, why do we need to do that? Seems to me like if someone is going to be minting the same asset in the same batch, they would never need two tranches? I'm just wondering if this option should be removed if that is its only purpose. If it should not be removed, I think we should change the description from

```
The name of the asset in the batch that will anchor a new asset group.
This asset will be minted with the same group key as the anchor asset.
```

to

```
The name of a new asset in the same batch that will anchor a new asset group.
This asset will be minted with the same group key as the anchor asset.
This option cannot be used to anchor to an asset that has already been finalized in a previous batch.
```

If it is supposed to work from a second batch, I think we should fix it.

guggero:
> Does that only work to mint two tranches in the same batch (without having a group_key first)?

Yes, the group anchor (string) argument is only for tranches in the same batch.
I know we need to improve the documentation in this area. But basically the flags/fields work this way:

- `new_grouped_asset`: I want to create a new group and this asset starts the group.
- grouped_asset`: This asset belongs to a group and I'll specify what group that is with other params (can either be `group_anchor` or `group_key`)
- `group_internal_key`: When starting a new group, instruct `lnd` to use this specific key from the keychain to sign for the group (only works with `new_grouped_asset`).
- `external_group_key`: Don't use a group internal key that is owned by `lnd`, use the externally owned key specified in this struct. This is required for both starting a group as well as re-issuing into an existing group that uses the external group (these are basically signing instructions).
- `group_anchor`: Another asset in the current batch started the group, _this_ asset should be put in the same group as the one in the batch with the name given (use case: multiple collectibles that belong to the same group but are minted in the same transaction). Reason for this being a string: In the case where the user doesn't want to define what group key they want to assign, one is created automatically. But this is done when sealing the batch, so we can't reference it as an actual public key here. That's why we specify the group anchor by its name (string).
- `group_key`: Another asset in another batch/transaction started the group. This asset should join the group that has this group public key.

Feel free to use this in any way to update the docs...

tapgarden/planter.go
```diff
@@ -2668,14 +2668,19 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
 	// If a group internal key or tapscript root is specified, emission must
 	// also be enabled.
 	if !req.EnableEmission {
-		if req.GroupInternalKey != nil {
+		// For re-issuing grouped assets or regular (non-grouped)
```
ZZiigguurraatt, line 2671: should this say "NEW regular (non-grouped)?

tapgarden/planter.go
```diff
@@ -2668,14 +2668,19 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
 	// If a group internal key or tapscript root is specified, emission must
```
ZZiigguurraatt, line 2668: you want to remove "emission" here too?

tapgarden/planter.go
```diff
@@ -2668,14 +2668,19 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
...
-		if req.GroupInternalKey != nil {
+		// For re-issuing grouped assets or regular (non-grouped)
+		// assets, the group internal key shouldn't be set. It is,
+		// however, set for re-issuance with an external key, because
+		// the internal group key is the key we compare the external key
+		// against.
+		if req.GroupInternalKey != nil && req.ExternalKey.IsNone() {
 			return fmt.Errorf("cannot specify group internal key " +
-				"without enabling emission")
+				"without creating a new grouped asset")
```
ZZiigguurraatt, line 2678: should you add "first" on the end of this line?

tapgarden/planter.go
```diff
@@ -2668,14 +2668,19 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
 	// If a group internal key or tapscript root is specified, emission must
 	// also be enabled.
 	if !req.EnableEmission {
-		if req.GroupInternalKey != nil {
+		// For re-issuing grouped assets or regular (non-grouped)
+		// assets, the group internal key shouldn't be set. It is,
+		// however, set for re-issuance with an external key, because
```
ZZiigguurraatt, line 2673: is it required or optional? I'm able to do a second tranche using external signer without setting it.

tapgarden/seedling.go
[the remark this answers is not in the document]
ZZiigguurraatt: not understanding it, but may be outside of my knowledge scope, so maybe it doesn't need more clarification. not sure.

tapgarden/planter.go
```diff
@@ -2668,14 +2668,19 @@ func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
 	// If a group internal key or tapscript root is specified, emission must
 	// also be enabled.
 	if !req.EnableEmission {
-		if req.GroupInternalKey != nil {
+		// For re-issuing grouped assets or regular (non-grouped)
+		// assets, the group internal key shouldn't be set. It is,
```
ZZiigguurraatt, line 2672: seems contradictory to your comment https://github.com/lightninglabs/taproot-assets/pull/1517/files#r2077342801 ?

tapgarden/planter.go
guggero: We're in the minting process, so it's always a new asset.

tapgarden/planter.go
guggero: We'll do that once we rename the actual variable as well. Didn't want to do that as it will touch a loooot of code lines.

tapgarden/planter.go
guggero: No, it means "without the new_grouped_asset field". Perhaps I should mention the field name then. But then we have to make sure that it is called the same on the CLI.

tapgarden/planter.go
guggero: Again, the code path here doesn't necessarily match the requirements for the flags on the RPC/CLI level. Things happen before we get here. So this comment isn't meant for the user.

tapgarden/seedling.go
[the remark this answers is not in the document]
guggero: Yes, this is several layers down in the code from the RPC call, things look different here and might be pre-populated with stuff queried from the database. So you can ignore this.

tapgarden/seedling.go
```diff
@@ -183,10 +183,49 @@ func (c Seedling) validateFields() error {
 func (c Seedling) validateGroupKey(group asset.AssetGroup,
 	anchorMeta *proof.MetaReveal) error {
 
-	// We must be able to sign with the group key.
-	if !group.GroupKey.IsLocal() {
-		groupKeyBytes := c.GroupInfo.GroupPubKey.SerializeCompressed()
-		return fmt.Errorf("can't sign with group key %x", groupKeyBytes)
+	switch {
```
ffranr, line 186: Instead of this switch you can just return early with:
```golang
if c.ExternalKey.IsNone() && !group.GroupKey.IsLocal() {
	groupPubKey := c.GroupInfo.GroupPubKey
	groupKeyBytes := groupPubKey.SerializeCompressed()
	return fmt.Errorf("can't sign with group key %x",
		groupKeyBytes)
}
```

And you don't need to wrap the `fn.MapOptionZ` call in `case c.ExternalKey.IsSome():`. Because `fn.MapOptionZ` will just return `err == nil` if `c.ExternalKey.IsNone()`. Maybe we can save on indentation like that.

itest/mint_fund_seal_test.go
```diff
@@ -577,6 +577,9 @@ func testMintExternalGroupKeyChantools(t *harnessTest) {
 	mintReq2 := CopyRequest(issuableAssets[0])
 	mintReq2.Asset.Name = "itestbuxx-money-printer-brrr-tranche-2"
 	mintReq2.Asset.ExternalGroupKey = externalGroupKey
+	mintReq2.Asset.GroupedAsset = true
+	mintReq2.Asset.NewGroupedAsset = false
+	mintReq2.Asset.GroupKey = batchAssets[0].AssetGroup.TweakedGroupKey
```
ffranr, line 582: Why is `mintReq2.Asset.GroupedAsset = true` necessary here? I would have expected setting `mintReq2.Asset.ExternalGroupKey` to imply group asset sufficiently.

Is setting these fields necessary or are we just covering all possibilities here?

guggero: The RPC (and by extension the CLI) interface for the minting call is not optimal and quite confusing. See https://github.com/lightninglabs/taproot-assets/pull/1517#issuecomment-2859009609

So you need to set at least one of the `GroupedAsset` or `NewGroupedAsset` to `true` if you want to start or issue into a group, independent of the other fields.

tapgarden/seedling.go
guggero: Ah nice, that's actually much nicer. Updated, thanks.

ffranr (approved):
thanks for the fix

Issue #1535, opened 2025-05-13
[bug]: nil pointer on ListBatches


darioAnongba:
### Background

When `ListBatches` is called with `Verbose: true`. The daemon breaks with a nil pointer.

### Your environment

```sh
version:\"0.5.1-alpha commit=v0.5.1\" lnd_version:\"0.18.5-beta\" network:\"regtest\" sync_to_chain:true"
```

### Steps to reproduce

- Run Polar with `tapd`.
- Mint an asset with MintAsset (tried with Collectible)
- Call `ListBatches` with `Verbose: true`

### Expected behavior

Returns sucessfully

### Actual behavior

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x12deb6c]
goroutine 1390 [running]:
github.com/lightninglabs/taproot-assets.marshalVerboseBatch({{0x1925830, 0x7}, 0xdab5bffa, {0x192284c, 0x5}, {0x2e95520, 0x0, 0x0}, 0x2df17e0, 0x2ca8ae0, ...}, ...)
	github.com/lightninglabs/taproot-assets/rpcserver.go:4123 +0x2c
github.com/lightninglabs/taproot-assets.(*rpcServer).ListBatches.func1(0x2?)
	github.com/lightninglabs/taproot-assets/rpcserver.go:954 +0x8c
github.com/lightninglabs/taproot-assets/fn.MapErr[...]({0x40009de770?, 0x2, 0x1db84}, 0x4000a8b720?)
	github.com/lightninglabs/taproot-assets/fn/func.go:87 +0x68
github.com/lightninglabs/taproot-assets.(*rpcServer).ListBatches(0x40006dc8a0, {0x19091e0?, 0x40006dc640?}, 0x4000b3a200)
	github.com/lightninglabs/taproot-assets/rpcserver.go:950 +0x3a8
github.com/lightninglabs/taproot-assets/taprpc/mintrpc._Mint_ListBatches_Handler.func1({0x1cd06a8?, 0x40006e3320?}, {0x17ef4c0?, 0x4000b3a200?})
	github.com/lightninglabs/taproot-assets/taprpc/mintrpc/mint_grpc.pb.go:335 +0xd0
[...]

PR #1536, opened 2025-05-13
Fix panic when listing unfunded pending batch with verbose

ffranr:
Closes: https://github.com/lightninglabs/taproot-assets/issues/1535

---

This PR fixes a bug in `ListBatches` where a single nil element was left in the `verboseBatches` list, causing a panic during RPC marshalling. The bug occurred when a single unfunded pending batch was present.

Changes:

* Set `verboseBatches` length to zero to avoid uninitialized elements.
* Add a comment explaining why funding is required for verbose batch info.
* Add an itest to cover this edge case.
* Improve proto documentation for the `verbose` field in `ListBatches`.

### DIFF
--- a/itest/assets_test.go
+++ b/itest/assets_test.go
@@ -174,7 +174,17 @@ func testMintBatchResume(t *harnessTest) {
 
 	// Build one batch, but leave it as pending.
 	BuildMintingBatch(t.t, t.tapd, simpleAssets)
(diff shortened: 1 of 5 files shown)


Roasbeef (approved):
LGTM 🐧

guggero (approved):
Oops, my bad. Thanks for the fix!

PR #1537, opened 2025-05-14
tapchannel: don't break backward compatibility

guggero:
This partially reverts commit 3ed142ffdc59f53a3d405b960f5d24641c5cc51c to restore backward compatibility with nodes running on v0.5.x.

The intention with removing this explicit split root assignment was to simplify the code. But it actually would lead to force closes between nodes on v0.6.0 and v0.5.x, because they wouldn't be able to agree on what output should be chosen for the split root (since the behavior was changed).

### DIFF
--- a/tapsend/allocation_test.go
+++ b/tapsend/allocation_test.go
@@ -877,6 +877,129 @@ func TestDistributeCoins(t *testing.T) {
 				},
 			},
 		},
+		{
+			// This tests the backward compatibility case where the
+			// split root output is defined explicitly, even for an
+			// interactive packet (which is the case for channels).
+			name: "single asset, distributed to three outputs",
+			inputs: []*proof.Proof{
+				makeProof(t, assetID1Tranche1),
+			},
+			interactive: true,
+			//nolint:lll
+			allocations: []*Allocation{
+				{
+					Type:        CommitAllocationHtlcOutgoing,
+					Amount:      50,
+					OutputIndex: 2,
+				},
+				{
+					Type:        CommitAllocationToLocal,
+					Amount:      20,
+					OutputIndex: 3,
+				},
+				{
+					Type:        CommitAllocationToRemote,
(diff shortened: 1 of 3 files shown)


tapsend/allocation_test.go
```diff
@@ -877,6 +877,129 @@ func TestDistributeCoins(t *testing.T) {
...
+				},
+				assetID5.ID(): {
+					assetID5Tranche1.ScriptKey,
+				},
+			},
+			expectedOutputs: map[asset.ID][]*tappsbt.VOutput{
+				assetID4.ID(): {
+					{
+						Amount:            5000,
+						Type:              split,
```
Roasbeef, line 982: So this output is the split because it has a lower output index, even though we have a settled to-self output present in the commitment?

Roasbeef (commented):
LGRT

tapsend/allocation_test.go
guggero: This shows why the reverted commit was added initially: If there are multiple asset pieces in a channel, you can't guarantee that every on-chain output gets assets from every piece (depending on the amount distribution). So even if you say the to_remote output should house split root, there might be virtual packets (in this example asset ID 4) will not have any split output in the to_remote output.
In that case we just fall back to having the split at index 0.

But, since this new fallback case can only happen for new, grouped asset channels, it's okay since both nodes need to be on the newer version to understand.
And for the cases where there's only one asset in the channel, we know that each output will get a piece of that asset and we can guarantee that the split output is at the location it was in previous versions, hence restoring backward compatibility.

guggero:
I've added a backward compatibility test here: https://github.com/lightninglabs/lightning-terminal/pull/1074
As expected, that test broke and channels force closed before this fix. After updating to this commit, the test passes.

Roasbeef (approved):
LGTM 💎

Issue #1534, opened 2025-05-13
Add `IgnoreAsset` RPC endpoint


ffranr:
[Placeholder, will add description soon.]

PR #1541, opened 2025-05-15
Refactor in Preparation for Ignore Tree RPC Endpoints

ffranr:
This PR includes refactoring work as part of the effort to address [[issue #1534](https://github.com/lightninglabs/taproot-assets/issues/1534)](https://github.com/lightninglabs/taproot-assets/issues/1534).

---

* Extend the universe archive with ignore tree archive to gather early feedback.
* Improve/fix comments, documentation, and code robustness.


universe/archive.go
```diff
@@ -698,6 +698,13 @@ func (a *Archive) UniverseLeafKeys(ctx context.Context,
 func (a *Archive) MintingLeaves(ctx context.Context,
 	id Identifier) ([]Leaf, error) {
 
+	// Sanity check that the proof type is set to issuance. Only this proof
+	// type is valid for minting leaves.
+	if id.ProofType != ProofTypeIssuance {
```
Roasbeef, line 703: Surprised the itests aren't failing here. AFAICT, we still use this for normal transfer proofs as well: https://github.com/lightninglabs/taproot-assets/blob/1660f9ccd5d40717216a93fea16fa97491f7429f/rpcserver.go#L5399. 

So perhaps we just change the name of the method, and also the cal stack all the way down to the db?

asset/asset.go
```diff
@@ -546,21 +546,24 @@ func (t Type) String() string {
 	}
 }
 
-// PrevID serves as a reference to an asset's previous input.
+// PrevID identifies the anchoring state of an asset. It includes:
```
guggero, line 549: I'm not really sure this makes its use more clear. We mainly use this to point to a previous asset state (in order to spend it). So perhaps something along the lines of "PrevID identifies the anchoring state of a previous asset when spending it. This is conceptually equal to the PreviousOutPoint in a Bitcoin input."

tapdb/universe.go
```diff
@@ -117,15 +117,17 @@ type BaseUniverseStore interface {
...
+// valid for the ignore and burn proof types.
 func specifierToIdentifier(spec asset.Specifier,
 	proofType universe.ProofType) (universe.Identifier, error) {
 
 	var id universe.Identifier
 
 	// The specifier must have a group key to be able to be used within the
 	// ignore or burn tree context.
-	if !spec.HasGroupPubKey() {
+	requireGroupKey := proofType == universe.ProofTypeIgnore ||
```
guggero, line 128: This looks like a useful method that we could add on the `universe.ProofType` type directly.

universe/archive.go
ffranr: I believe that we have insufficient itest coverage. Method `MintingLeaves` is actually used to retrieve transfer proof leaves.

Our universe sync itest does not cover transfer proof syncing sufficiently, that's why my proof type check passed CI.

In the latest commits to this PR I've renamed method `MintingLeaves` to `FetchLeaves`. I'll put up an issue to improve itest coverage of transfer proof universe syncing.

asset/asset.go
ffranr: My PoV:

The problem/confusion here is that type `PrevID` is named poorly. Based on its fields, it could represent either the current or previous state of an asset.

The distinction between previous and current should be conveyed by the name of the field that uses this type—not by the type name itself. A more general type name would therefore be clearer and simpler.

What `PrevID` actually is is a bitcoin outpoint decorated with TAP data. Whether it's previous or current is not relevant at the type level.

So we could name it something neutral like `TapOutPoint` or `AssetOutPoint` for example.

> We mainly use this to point to a previous asset state (in order to spend it)

Yes, but, we also use it in every unspent proof leaf to refer to the current unspent outpoint. Which confused me for a while when getting familiar with our code/TAP.

I will try to improve the doc as you suggest. I'm not going to rename `PrevID` in this PR.

guggero: > Yes, but, we also use it in every unspent proof leaf to refer to the current unspent outpoint.

Hmm, okay, I wasn't aware of this. I see it's being used in the `IgnoreTuple` where I agree it's confusing. Do you have other examples? I couldn't really find others at a quick glance (perhaps they're in in-flight PRs?).

Maybe the solution is to add a comment to those fields where the `Prev` in the name adds confusion (as with the ignore tuple), and just mention that we use that type because of its properties and not because of its name.
Because I think renaming `PrevID` will be quite a large diff with minimal benefits in my opinion.

ffranr: More complaints I have about the current name:
* `PrevID` is not the previous asset ID.
* It does not refer to an asset's "previous input" during spending. I think it refers to the **current** inputs in the spend.

ffranr:
I've added more commits to rename `universe.BaseBackend`.

universe/archive.go
ffranr: Issue: https://github.com/lightninglabs/taproot-assets/issues/1543

tapdb/universe.go
guggero: It's less about re-usability and more about signifying that there's a specific restriction for the `universe.ProofTypeIgnore` and `universe.ProofTypeBurn` types, in that they can/should only be produced for grouped assets.

So yeah, a method `func (t ProofType) RequireGroupKey() bool {` would make that more clear in my opinion.
But just a suggestion, non-blocking.

asset/asset.go
guggero: > `PrevID` here is the outpoint of the asset which is an input in the spend. It is not the previous input of the input which is being spent. And it definitely isn't the previous asset ID.

Yes, but that's in line with how Bitcoin inputs work. An input has the `PrevOutPoint` field, which is what the input points to. So I think just having the reference to Bitcoin's `PrevOutPoint` in the type description might help, as it's the same concept.

> Note also that `proof/verifier.go` has `type AssetPoint = asset.PrevID`. Which might be the way to go? 

Yes, I like that. Wherever we don't use it as a previous ID (input), we could create a type alias.

PR #1549, opened 2025-05-20
   tapdb: allow reinsertion of an existing address the the book

bhandras:
With this change we allow "upserting" an address. This is useful in the case when a tapd user might not know if they've created a specific address already (as there's no easy way to look it up).


tapdb/sqlc/queries/addrs.sql
```diff
@@ -1,9 +1,13 @@
...
 INSERT INTO addrs (
     version, asset_version, genesis_asset_id, group_key, script_key_id,
     taproot_key_id, tapscript_sibling, taproot_output_key, amount, asset_type,
     creation_time, proof_courier_addr
-) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12) RETURNING id;
+) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12)
+ON CONFLICT(taproot_output_key) DO UPDATE SET
+    -- no-op update to trigger RETURNING clause
+    taproot_output_key = excluded.taproot_output_key
+RETURNING id;
```
ffranr, line 10: If there's a conflict on the taproot output key, I think the query should update the proof courier address, since that's not part of the taproot output key.

Otherwise, I think this should be compatible with how `tapgarden/custodian` currently operates.

ffranr: Had a 1:1 with Oli. My idea to allow updating the proof courier address here isn't sound. If a user has already shared their Tap address, we shouldn’t allow them to modify it in the database afterward. They will look to the wrong courier for the proof.

tapdb/addrs_test.go
```diff
@@ -172,6 +172,9 @@ func TestAddressInsertion(t *testing.T) {
 	}
 	require.NoError(t, addrBook.InsertAddrs(ctx, addrs...))
 
+	// Now insert the addresses again, to test that upsert works as well.
+	require.NoError(t, addrBook.InsertAddrs(ctx, addrs...))
```
guggero, line 176: What's your exact use case here? Because I'm not sure what the user experience of this change will be. Since we won't update any fields of the address, we will silently ignore any intended changes by the user.
Do you perhaps just need a `QueryAddrByKey` RPC or something along those lines?

bhandras: An address that is created with a given amount, script and internal keys, etc would need to be stored by the caller of `NewAddr` and fetched with `QueryAddrs` if we expect the caller application to restart at any point in time. This means that we'd need to guess when the address was created and scan through the results of `QueryAddrs`.

Maybe instead of adding `QueryAddrByKey` (which imo would be useful), we could make upsertion such that it will not change any of the fields if the address is already present. wdyt?

guggero: Ah, okay. So how about we add a check that compares all fields of the upserted address to the one provided in the RPC request of `NewAddr`. If all fields are equal, just return the address.
If any field is different (e.g. the user expects to be able to change the proof courier address for example), we'd return an error with a message like "address with given key already exists with different properties"?

That way we can serve your use case but also not have confusing/silent behavior that the user might not understand.

bhandras: Done, PTAL!

tapdb/sqlc/queries/addrs.sql
```diff
@@ -1,9 +1,45 @@
...
+) 
+ON CONFLICT (taproot_output_key) DO UPDATE
+SET
+    -- If the WHERE clause below is true (exact match on all other fields,
+    -- except for creation_time), we set taproot_output_key to its current
+    -- conflicting value. This is a no-op in terms of data change but allows
+    -- RETURNING id to work on the existing row.
+    taproot_output_key = excluded.taproot_output_key
+WHERE 
+    addrs.version = excluded.version
```
guggero, line 26: Hmm, interesting approach. 

So this now returns `unable to insert addr: sql: no rows in result set` if some fields mismatch. Sounds good to me, but I think we should catch that on the RPC server and return a more user friendly error. Otherwise it's hard to understand what's wrong.

tapdb/addrs_test.go
```diff
@@ -172,6 +172,21 @@ func TestAddressInsertion(t *testing.T) {
...
+	// Changing an address' creation time should not affect the insertion
+	// process but should not change the database either.
+	addrTime := addrs[0].CreationTime
+	addrs[0].CreationTime = addrTime.Add(time.Second)
+	require.NoError(t, addrBook.InsertAddrs(ctx, addrs[0]))
+	addrs[0].CreationTime = addrTime
+
+	// Now change the amount, which upon upsert should trigger an error.
+	addrs[0].Amount += 1
+	require.Error(t, addrBook.InsertAddrs(ctx, addrs[0]))
```
guggero, line 187: nit: `require.ErrorIs(t, addrBook.InsertAddrs(ctx, addrs[0]), sql.ErrNoRows)`?

guggero (commented):
Nice! One request and a nit, otherwise LGTM.

tapdb/sqlc/queries/addrs.sql
bhandras: ptal

tapdb/addrs_test.go
bhandras: ptal

guggero (approved):
Thanks, LGTM :tada:

ffranr (approved):

PR #1554, opened 2025-05-21
RPC: add ignore tree RPC endpoints

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/1534

---

Sharing early to get early feedback on naming and architecture. Still missing itests and query RPC endpoint.

This PR adds RPC endpoints to populate and query from ignore trees.


universe/archive.go
```diff
@@ -24,6 +25,10 @@ type ArchiveConfig struct {
 	// if the identifier has never been seen before.
 	NewBaseTree func(id Identifier) StorageBackend
 
+	// IgnoreTreeArchive is an archive of all known ignore trees across all
+	// assets.
+	IgnoreTreeArchive IgnoreTreeArchive
```
Roasbeef, line 30: FWIW, the I envisioned things originally is that you almost never serve the ignore tree by itself. It's always accompanied with two additional layers of inclusion proofs: first into the mains supply tree, and then the block header proof into the blockchain.

universe/archive.go
```diff
@@ -711,6 +716,21 @@ func (a *Archive) FetchLeaves(ctx context.Context,
 	)
 }
 
+// UpsertIgnoreTuples inserts the passed ignore tuples into the ignore tree
+// archive. If the tuples already exist, then no action is taken.
+//
+// NOTE: Ignore proofs are not included in the multiverse, as they are purely
+// off-chain. Only issuance and transfer proofs are included in the multiverse.
+func (a *Archive) UpsertIgnoreTuples(ctx context.Context,
```
Roasbeef, line 724: In the context of the greater system, we wouldn't want to insert ignore tuples directly into the tree. We want to make sure that we're always serving information that's consistent with the latest commitment on chain. Hence the way we stage updates as state transitions in #1508.

universe/ignore_records.go
```diff
@@ -91,6 +97,85 @@ func (i *IgnoreTuple) Record() tlv.Record {
...
+		return zero, err
+	}
+
+	digest := sha256.Sum256(ignoreBytes)
+	return digest, nil
+}
+
+// GenSig generates a Schnorr signature over the IgnoreTuple using the
+// provided SignerClient and key locator.
+func (i *IgnoreTuple) GenSig(ctx context.Context, signer lndclient.SignerClient,
```
Roasbeef, line 136: What about instead having a method that accepts the same params, but returns a [`SIgnedIgnoreTuple`](https://github.com/lightninglabs/taproot-assets/blob/49f92bdcd42ff753cc0747a55e6b33991f8e28fc/universe/ignore_records.go#L218-L225)?

asset/asset.go
```diff
@@ -569,6 +606,10 @@ func (id PrevID) Hash() [sha256.Size]byte {
 	return *(*[sha256.Size]byte)(h.Sum(nil))
 }
 
+// OutPoint is a type alias for an asset outpoint. It links the anchor outpoint
+// to the asset it secures.
+type OutPoint = PrevID
```
Roasbeef, line 611: `PrevOut`?

rpcserver.go
```diff
@@ -3769,6 +3769,168 @@ func marshalOutputType(outputType tappsbt.VOutputType) (taprpc.OutputType,
...
+		ctx, r.cfg.Lnd.Signer, delegationKey,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign ignore tuple: %w", err)
+	}
+
+	// Upsert a signed ignore tuple into the ignore tree archive and
+	// get back the authenticated ignore tuples.
+	signedIgnore := universe.NewSignedIgnoreTuple(ignoreTuple, ignoreSig)
+	authIgnoreTuples, err := r.cfg.UniverseArchive.UpsertIgnoreTuples(
```
Roasbeef, line 3892: Rather than insert directly into the DB, we should call into the supply commit state machine: https://github.com/lightninglabs/taproot-assets/blob/648cb72d8b3de9d3f3324ee24b777f918121693a/universe/supplycommit/supply_commit_states.go#L96-L101

Roasbeef (changes requested):

asset/asset.go
guggero: Or `AnchorPoint` (to be used as `asset.AnchorPoint` which reads nicely IMO) to not confuse it with a purely on-chain `wire.OutPoint`?

taprpc/taprootassets.proto
```diff
@@ -1492,6 +1505,32 @@ message OutPoint {
...
+
+    // The asset ID of the asset anchored at the outpoint.
+    AssetID asset_id = 2;
+
+    // The script key of the asset. This is the taproot output key that the
+    // asset is locked to.
+    oneof script_key {
+        // The script key in raw byte form. Use this field when interacting
+        // via gRPC.
+        bytes script_key_bytes = 4;
```
guggero, line 1526: Same question here: Why do we need both forms? On the CLI it should look the same. But in an itest I'd never easily know which field would actually be set by the RPC server (without looking up the code or calling the RPC).

guggero (commented):
Looks pretty good, did a quick first pass.

taprpc/taprootassets.proto
```diff
@@ -157,6 +157,19 @@ service TaprootAssets {
...
+// The asset ID is a 32-byte identifier that uniquely identifies an asset.
+message AssetID {
+    oneof id {
+        // The 32-byte asset ID in raw byte form. Use this field when
+        // interacting via gRPC.
+        bytes asset_id_bytes = 1;
+
+        // The 32-byte asset ID encoded as a hex string. Use this field when
+        // interacting via REST.
+        string asset_id_str = 2;
```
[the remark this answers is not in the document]
ffranr: I've gone for bytes where appropriate except for txid/outpoint which is string.

ffranr:
I've rebased on to branch `supply-commit-machine`.

litbot-9000:

ffranr:
**Current status**

* Introduced required changes to LND; PR opened at lightningnetwork/lnd#10028.
* Successfully published a supply-commitment transaction, but still need to verify that it commits to the intended state.

**Remaining work before we can merge this PR**

1. Merge lightningnetwork/lnd#10028 and bump the LND dependency in tapd.
2. DONE: ~~Populate `ErrorReporter` in `protofsm.StateMachineCfg` (see TODO).~~
3. DONE: ~~Replace the `InternalKey` field in `oldCommitment` (within `newRootCommitment`) with the full `KeyDescriptor` so we can spend old commitments (we can already spend pre-commitments in this PR).~~

ffranr:
I'm going to split this PR into smaller PRs. Here's the first: https://github.com/lightninglabs/taproot-assets/pull/1638

ffranr:
I've set this PR to "draft". Depends on PR https://github.com/lightninglabs/taproot-assets/pull/1643

ffranr:
I've split the `WIP` commit into commits that we can merge.

I haven't added command line changes to this PR. We might need to add a `universe supplycommit` subcommand?

ffranr:
Depends on https://github.com/lightninglabs/taproot-assets/pull/1655
Please no merge until the base branch has been merged into `main`.

asset/asset.go
```diff
@@ -55,6 +55,21 @@ const (
 // compressed, 33-byte form.
 type SerializedKey [33]byte
 
+// NewSerializedKeyFromBytes creates a new SerializedKey from the given byte
+// slice.
+func NewSerializedKeyFromBytes(key []byte) (SerializedKey, error) {
+	var zero SerializedKey
+
+	if len(key) != 33 {
```
guggero: Since we have an error return value, I think we should check that the point is actually on the curve by parsing it as `btcec.PublicKey`.

universe/ignore_records.go
```diff
@@ -92,6 +98,88 @@ func (i *IgnoreTuple) Record() tlv.Record {
...
+		return zero, err
+	}
+
+	digest := sha256.Sum256(ignoreBytes)
+	return digest, nil
+}
+
+// GenSignedIgnore generates a Schnorr signature over the IgnoreTuple using the
+// provided SignerClient and key locator.
+func (i *IgnoreTuple) GenSignedIgnore(ctx context.Context,
```
guggero, line 137: Some basic unit tests around this would be nice. Can take code from here, which is very similar: https://github.com/lightninglabs/taproot-assets/blob/a132dea7b64ca5ec2c1cd4bad27a9acc7e885566/proof/uni_commitments_test.go#L119

And perhaps add a `Verify` method as well, since that will certainly be used at some point?

taprpc/universerpc/universe.proto
```diff
@@ -169,6 +169,13 @@ service Universe {
     */
     rpc UpdateSupplyCommit (UpdateSupplyCommitRequest)
         returns (UpdateSupplyCommitResponse);
+
+    /* tapcli: `universe fetchsupplycommit`
```
guggero, line 174: We need actual command counterparts for all three RPCs in the `cmd/tapcli` package.

taprpc/universerpc/universe.yaml
```diff
@@ -86,4 +86,8 @@ http:
 
     - selector: universerpc.Universe.UpdateSupplyCommit
       post: "/v1/taproot-assets/universe/update-supply-commit"
+      body: "*"
```
guggero, line 89: To make these three endpoints a bit more RESTy, should we do this instead?

```
    - selector: universerpc.Universe.IgnoreAssetOutPoint
      post: "/v1/taproot-assets/universe/supply/ignore"
      body: "*"

    - selector: universerpc.Universe.UpdateSupplyCommit
      post: "/v1/taproot-assets/universe/supply/{group_key}"
      body: "*"

    - selector: universerpc.Universe.FetchSupplyCommit
      get: "/v1/taproot-assets/universe/supply/{group_key}"
```

Not sure if using the `{group_key}` placeholder that is of type `bytes` will work, but happy to test that if you agree to change the endpoints to the above.

itest/psbt_test.go
```diff
@@ -3619,10 +3619,19 @@ func sendToTapscriptAddr(ctx context.Context, t *harnessTest, alice,
 	AssertSendEventsComplete(t.t, bobAddr.ScriptKey, sendEvents)
 }
 
+// sendAssetAndAssertResponse is a struct that holds the response of a
+// sendAssetsToAddr call.
+type sendAssetAndAssertResponse struct {
+	// RpcResp is the response message returned by the SendAsset RPC
+	// endpoint.
+	RpcResp *taprpc.SendAssetResponse
```
guggero, line 3627: Why do we need this wrapper around the existing type? Can't we just return the RPC response directly?

itest/supply_commit_test.go
```diff
@@ -132,3 +136,238 @@ func testPreCommitOutput(t *harnessTest) {
...
+		t, expectedIgnoreSubtreeRootHash, inclusionProofBytes,
+		ignoreLeafKey, ignoreLeaf,
+	)
+
+	// Verify that the mined supply commitment transaction commits to the
+	// supply commitment tree.
+	require.Len(t.t, minedBlocks, 1)
+
+	block := minedBlocks[0]
+	// TODO(ffranr): Why are there two transactions in this block?
```
guggero: A block always has the coinbase transaction.

itest/supply_commit_test.go
```diff
@@ -132,3 +136,238 @@ func testPreCommitOutput(t *harnessTest) {
...
+						"commitment tx outputs in " +
+						"block")
+				}
+
+				foundCommitTxOut = true
+			}
+		}
+	}
+
+	require.True(t.t, foundCommitTxOut)
```
guggero, line 385: We can't assert here that we wouldn't accept the ignored asset again because that part isn't hooked up fully in `tapd` yet, right?

guggero (commented):
Great work! Very nice to see that many stacked PRs come together in an integration test :100: 

Just a couple of small things from my end, this is very close!

itest/psbt_test.go
ffranr: My thinking here is that since `sendAssetAndAssert` is used in several places, returning a typed struct allows us to extend it with additional fields later without changing all the call sites. That keeps future changes cheap. Given that this is a helper, I figured that tradeoff was acceptable. What do you think?

taprpc/universerpc/universe.yaml
ffranr: Much better paths! Thanks. Lets try.

taprpc/universerpc/universe.proto
ffranr: :+1: I was waiting for a first review to check that you guys were happy with my formulation. Will add CLI now.

rpcserver.go
```diff
@@ -3796,6 +3797,125 @@ func marshalOutputType(outputType tappsbt.VOutputType) (taprpc.OutputType,
...
 }
 
+// IgnoreAssetOutPointRequest is the request type for the IgnoreAssetOutPoint
+// RPC method.
+type IgnoreAssetOutPointRequest struct {
+	// AssetAnchorPoint is the asset anchor point to ignore.
+	AssetAnchorPoint asset.AnchorPoint
+
+	// Amount is the amount of the asset at the anchor outpoint.
+	Amount uint64
```
Roasbeef, line 3807: I wonder if we should just try to extract/assume the amount. Is this obtainable in all cases? Or should we just try to do a sanity check at some point in the pipeline?

rpcserver.go
```diff
@@ -3796,6 +3797,125 @@ func marshalOutputType(outputType tappsbt.VOutputType) (taprpc.OutputType,
...
+	delegationPubKey, err := metaReveal.DelegationKey.UnwrapOrErr(
+		fmt.Errorf("delegation key not found for given asset"),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Fetch the full delegation key locator.
+	delegationKey, err := r.cfg.TapAddrBook.FetchInternalKeyLocator(
+		ctx, &delegationPubKey,
```
Roasbeef, line 3877: So this'll reliably fail if we weren't the ones that actually issued the asset right? Or will signing fail below? 

Either way, we should amke thsi check and the related failure explicit.

rpcserver.go
```diff
@@ -3916,6 +3916,58 @@ func (r *rpcServer) IgnoreAssetOutPoint(ctx context.Context,
...
+	if !metaReveal.UniverseCommitments {
+		return nil, fmt.Errorf("universe commitment is not " +
+			"enabled for asset group anchor asset")
+	}
+
+	// Formulate an asset specifier from the asset group key.
+	assetSpec := asset.NewSpecifierFromGroupKey(*groupPubKey)
+
+	// Send a commit tick event to the supply commitment manager.
+	event := supplycommit.CommitTickEvent{}
```
Roasbeef, line 3988: It occurs to me that we didn't ever add the time baed ticker to the multi commitment manager. I remember we ended up removing it from the minting state machine in the past. Is this our new preferred route? So the caller must also "finalize" the batch, instead of the existing auto aggregration?

I think the manual finalize makes sense here personally.

taprpc/universerpc/universe.proto
```diff
@@ -703,4 +710,68 @@ message UpdateSupplyCommitRequest {
...
+    bytes supply_tree_leaf_key = 3;
+
+    // The inclusion proof for the supply commit subtree root node.
+    bytes supply_tree_inclusion_proof = 4;
+}
+
+message FetchSupplyCommitResponse {
+    // The supply commitment merkle sum root node for the specified asset.
+    MerkleSumNode supply_commitment_root = 1;
+
```
Roasbeef, line 766: I think we should also include some basic block information here. So: confirmation height, header hash, etc. 

This should also be useful in itests.

itest/supply_commit_test.go
```diff
@@ -132,3 +136,238 @@ func testPreCommitOutput(t *harnessTest) {
...
+//  1. Mints an asset group with universe supply commitments enabled.
+//  2. Transfers a portion of the asset to a secondary node.
+//  3. Instructs the primary node to ignore the outpoint now owned by the
+//     secondary node.
+//  4. Updates the asset group’s supply commitment, which should now include
+//     the ignored outpoint in the “ignore” subtree.
+//  5. Mines the commitment transaction.
+//  6. Retrieves the updated supply commitment transaction and asserts that the
+//     ignored subtree contains the expected outpoint.
+func testSupplyCommitIgnoreAsset(t *harnessTest) {
```
Roasbeef, line 153: Nice e2e test!

itest/supply_commit_test.go
Roasbeef: Can remove (extra is the coinbase txn as commented).

itest/supply_commit_test.go
Roasbeef: Yeah the syncer/verifier isn't hooked up yet.

itest/supply_commit_test.go
Roasbeef: Some extra subcases or distinct cases we can add:
  * Verify error if a 3rd party nodes attempts to add something to the ignore set
  * Caller inputs incorrect amount (can we always fully verify this?)
  * Do add another leaf to the ignore tree, verify its spent properly, etc.

universe/ignore_records.go
ffranr: I've added basic unit tests for `GenSignedIgnore`. I mean to add the `Verify` side of things with the verifier component in the next PR, if that works for you.

rpcserver.go
ffranr: I've made the endpoint's error message more explicit and extended the integration test to include a case where a non-owner node attempts to create a signed ignore tuple and correctly fails.

rpcserver.go
ffranr: > I think the manual finalize makes sense here personally.

I agree. I think, for now, the simplest MVP approach is to require manual triggering of the supply commit update publication.

rpcserver.go
ffranr: I'll ping you in chat on this. I've been thinking about this a bit. But, yes, I think we need to at least cross reference the amount against a proof.

taprpc/universerpc/universe.yaml
guggero: Ah, too bad. I though we fixed this, but apparently it's still the case that if there are placeholders of type `bytes` in the URI, they need to be base64url encoded. So `0265c8cee9f4e2bdd440fbe87211ad928ed17a37cca8e5ac13feb87ca59025cd4d` would need to become `AmXIzun04r3UQPvochGtko7RejfMqOWsE_64fKWQJc1N` which is kind of unfortunate. I guess that's why we did this for the other universe calls:

```
oneof key {
        // The 32-byte asset group key specified as raw bytes (gRPC only).
        bytes group_key = 1;

        // The 32-byte asset group key encoded as hex string (use this for
        // REST).
        string group_key_str = 2;
}
```

Then the mapping can become `/v1/taproot-assets/universe/supply/{key.group_key_str}`.
Would be nice if we could do it here as well, but leave it up to you if you want to revert to the previous mapping (without any placeholders in the URI) or add the `oneof` that requires a few more changes in the parameter parsing logic.

taprpc/universerpc/universe.proto
ffranr: I've added. And also assertions in itest.

itest/supply_commit_test.go
ffranr: I've extended the test to cover this
> Verify error if a 3rd party nodes attempts to add something to the ignore set

I'll add more edge cases to the itest in a followup PR.

taprpc/universerpc/universe.yaml
ffranr: I've gone with the `oneof` method.

taprpc/universerpc/universe.yaml
```diff
@@ -82,4 +82,8 @@ http:
 
     - selector: universerpc.Universe.IgnoreAssetOutPoint
       post: "/v1/taproot-assets/universe/supply/ignore"
+      body: "*"
+
+    - selector: universerpc.Universe.UpdateSupplyCommit
+      post: "/v1/taproot-assets/universe/supply/{group_key_str}"
```
guggero, line 88: This needs to be `{group_key.group_key_str}` to account for the name of the `oneof`. Same on the other one.

ffranr: hmmm I get error:
```
Generating protos from universerpc/universe.proto, into universerpc
--grpc-gateway_out: no field "group_key.group_key_str" found in UpdateSupplyCommitRequest
```

when i use `{group_key.group_key_str}`

guggero: Ah yes, sorry, my bad. The `id` in `{id.asset_id_str}` comes from the `ID` proto message, not the `oneof` :see_no_evil:

cmd/commands/universe.go
```diff
@@ -1503,3 +1505,190 @@ func universeEventStatsQueryCommand(ctx *cli.Context) error {
...
+
+var ignoreAssetOutPointCmd = cli.Command{
+	Name:  "ignore",
+	Usage: "ignores an asset outpoint when committing to the asset supply",
+	Description: "This command allows us to ignore a specific asset UTXO " +
+		"when creating a new supply commitment. This is useful if " +
+		"an otherwise valid asset should not be counted towards the " +
+		"total asset group supply.",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
```
guggero, line 1533: Don't we also need the asset ID and script key here?

taprpc/taprootassets.proto
```diff
@@ -1541,6 +1541,24 @@ message ListBurnsResponse {
     repeated AssetBurn burns = 1;
 }
 
+// A transaction outpoint annotated with TAP-level asset metadata. It uniquely
+// identifies an asset anchored at a specific outpoint.
+message AssetOutPoint {
```
guggero: nit: since we're adding this as a new type, this could go into the `tapcommon.proto` instead.

guggero (approved):
Very nice, LGTM :tada:

cmd/commands/universe.go
ffranr: Added, thanks.

rpcserver.go
Roasbeef: Did you forget to push this up maybe? I don't see the addition to the itests.

ffranr: Here: https://github.com/lightninglabs/taproot-assets/pull/1554/files#diff-0a6ab6eac652089a797c6a8ad1a73fa59bb983d6422044ef62c8b4cd1baa68a8R388-R392

PR #1555, opened 2025-05-21
sample-tapd.conf+tapcfg: add default signet/testnet4 uni servers

guggero:
Uses our universe servers as proof couriers by default for the networks testnet4 and signet.

NOTE: The required CNAME entries don't exist yet (but should soon), just putting this up so we don't forget for the final release.


Roasbeef (approved):
LGTM 🌈

guggero:
The ports for testnet4 and signet actually need to be `443`, since the infrastructure is set up slightly differently from the testnet and mainnet universes.

PR #1563, opened 2025-05-23
tapd: allow skipping the funding step in CommitVirtualPsbts

bhandras:
In order to be able to construct zero fee transactions (which do not need funding inputs) we want to allow `CommitVirtualPsbts` to skip the funding step altogether.


rpcserver.go
```diff
@@ -2570,80 +2570,92 @@ func (r *rpcServer) CommitVirtualPsbts(ctx context.Context,
...
+			MinConfs:              1,
+			ChangeType:            P2TRChangeType,
+			CustomLockId:          req.CustomLockId,
+			LockExpirationSeconds: req.LockExpirationSeconds,
+		}
+
+		// Unfortunately we can't use the same RPC types, so we have to
+		// do a 1:1 mapping to the walletrpc types for the anchor change
+		// output and fee "oneof" fields.
+		type existingIndex = walletrpc.PsbtCoinSelect_ExistingOutputIndex
```
guggero, line 2602: We have a `type()` block at the beginning of the file where we create shorter definitions/aliases to avoid these long lines. Could move this and the other offending `walletrpc` types there to avoid the linter complaining.

rpcserver.go
```diff
@@ -2570,80 +2570,92 @@ func (r *rpcServer) CommitVirtualPsbts(ctx context.Context,
...
-	// fee "oneof" fields.
-	type existingIndex = walletrpc.PsbtCoinSelect_ExistingOutputIndex
-	switch change := req.AnchorChangeOutput.(type) {
-	case *wrpc.CommitVirtualPsbtsRequest_ExistingOutputIndex:
-		coinSelect.ChangeOutput = &existingIndex{
-			ExistingOutputIndex: change.ExistingOutputIndex,
-		}
+	fundedPacket := pkt
+	changeIndex := int32(-1)
+	success := false
```
guggero, line 2580: nit: merge this block into the `var` statement above?

guggero (approved):
LGTM, just need to address the linter complaints.

rpcserver.go
bhandras: done

rpcserver.go
bhandras: done

PR #1567, opened 2025-05-27
Use new NoopAddHtlc TLV when sending assets with default above-dust anchor amt

### LINKED ISSUE #888: [enhancement]: Premature Taproot-Asset Channel Exhaustion Due to Above-dust `lnd` Requirement

### LINKED ISSUE #1305: [feature]: Custom HTLC settlement on LND to compensate anchor sats amount
## Description

Currently every tap channel payment carries the default onchain anchor amt which is about 354 sats.

This may cause UX trouble when users pay small invoices, or generally in cases where the value of 354 sats weighs too much.

We want to somehow make LND's `lightningChannel` state machine to eliminate this one-sided balance change in satoshis.

## Solution

We could modify the way HTLCs are settled in LND, so that satoshis are never shifted upon settling an HTLC, but the custom aux blobs of the commitment tx of the channel do get updated. This would allow for the asset balance change to be reflected on the channel, without shifting the satoshi balance at all. There would only be a locked/in-flight amount of satoshis as long as the HTLC itself is in-flight, but as soon as the HTLC gets settled the settlement logic would return the sats to the sender and only really accept the assets.

## Considerations

- We should find a correct way of filtering which HTLCs should be treated this way (aux blob existence?), as we want to avoid accidentally returning larger amounts of satoshis carried by an HTLC. (example: HTLC carrying 1btc + 1M$)
- We should keep in mind that even though there's no balance change in terms of sats, the sats of the in-flight HTLCs are exposed to malicious grabbing by the receiver (via the force close path / revealing HTLCs onchain). We should only allow small sats amounts for such HTLCs to disincentivize such behavior.

## Related Issues
https://github.com/lightninglabs/taproot-assets/issues/888

GeorgeTsagk:
## Description

When sending HTLCs that carry assets we now set the noop flag whenever the amount is the default anchor amount. Since that amount is mandatory for the anchoring of assets we want to keep it around while HTLCs are inflight, but not actually send it to the channel counter party when the HTLC settles. This is exactly what the noop flag triggers.

The feature is currently hidden behind the `go:build dev` flag, as it will break backwards compatibility if used by default. This will be the case until we introduce feature related peer messages for custom channels.

Closes #888
Closes #1305

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -73,6 +73,14 @@
   [related PR](https://github.com/lightninglabs/taproot-assets/pull/1613) for
   more info.
 
+- Asset HTLCs that are received or sent do not shift the satoshi balance of the
+  channel. This is because now the default behavior is to use
+  [Noop HTLCs](https://github.com/lightninglabs/taproot-assets/pull/1567) which
+  lock some above-dust sats amount with the HTLC as long as it's in-flight, but
+  nullifies it upon settlement. This is currently hidden behind the dev build
+  tag and also needs to be toggled on via the `channel.noop-htlcs` configuration
+  option.
+
 ## RPC Additions
 
 - The [price oracle RPC calls now have an intent, optional peer ID and metadata


rfqmsg/noop_add_active.go
```diff
@@ -0,0 +1,8 @@
+//go:build dev
```
Roasbeef, line 1: 👍

rfqmsg/records.go
```diff
@@ -135,6 +143,11 @@ func (h *Htlc) Records() []tlv.Record {
 		records = append(records, r.Record())
 	})
 
+	if h.NoopAdd {
+		r := tlv.NewPrimitiveRecord[lnwallet.NoopAddHtlcType](true)
```
Roasbeef, line 147: Another thing to update in the bLIPs.

tapchannel/aux_traffic_shaper.go
```diff
@@ -395,7 +395,14 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
 	if htlc.Amounts.Val.Sum() > 0 {
 		log.Tracef("Already have asset amount (sum %d) in HTLC, not "+
 			"producing extra data", htlc.Amounts.Val.Sum())
-		return totalAmount, htlcCustomRecords, nil
+
+		htlc.SetNoopAdd()
+		updatedRecords, err := htlc.ToCustomRecords()
```
Roasbeef, line 400: Seems the aspect related to the reserve that you mentioned isn't in this PR yet?

Roasbeef (changes requested):

rfqmsg/records.go
```diff
@@ -59,6 +60,13 @@ type Htlc struct {
 
 	// RfqID is the RFQ ID that corresponds to the HTLC.
 	RfqID tlv.OptionalRecordT[HtlcRfqIDType, ID]
+
+	// NoopAdd is a flag that indicates whether this HTLC should be marked
+	// as a noop_add for LND. A noop_add HTLC behaves identically to a
+	// normal HTLC except for the settlement step, where the satoshi amount
+	// is returned back to the sender, but the commitment blob is still
+	// updated to reflect the asset balance changes.
+	NoopAdd bool
```
guggero, line 179: Shouldn't this be an optional record directly? Then below we can simply do:
```
	h.NoopAdd.WhenSome(func(r tlv.RecordT[NoopAddHtlcType, bool]) {
		records = append(records, r.Record())
	})
```

Then decoding will also work correctly, which I think currently it doesn't (should also add unit test cases for this).

tapchannel/aux_traffic_shaper.go
```diff
@@ -431,7 +431,14 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
 	if htlc.Amounts.Val.Sum() > 0 {
 		log.Tracef("Already have asset amount (sum %d) in HTLC, not "+
 			"producing extra data", htlc.Amounts.Val.Sum())
-		return totalAmount, htlcCustomRecords, nil
+
+		htlc.SetNoopAdd()
```
guggero: Doing the noop build flag this way works, but tbh the `rfqmsg` package wouldn't necessarily be the place I'd look for a build flag switch.
Should we do a `aux_features_dev.go` and `aux_features.go` in this package that defines a `UseNoOpHTLCs bool` variable that we then set here? E.g.: `htlc.SetNoopAdd(UseNoOpHTLCs)`?

rfqmsg/records.go
guggero: Hmm, looking at the rest of the diff, it looks like we never actually detect this flag in `tapd`. Going to circle back to this after taking a look at the `lnd` PR.

GeorgeTsagk: within tapd we check for the existence of this flag a bit further in this file, but only for the purpose of setting the record

The record is then only detected in LND, here we only care about whether we set it or not

ZZiigguurraatt:
and this is dependent on https://github.com/lightningnetwork/lnd/pull/9871 ?

ZZiigguurraatt:
Are you going to revert https://github.com/lightninglabs/taproot-assets/pull/1252 as part of this or some separate PR? We need to remove the `allow_overpay` option (https://github.com/lightninglabs/taproot-assets/blob/723311b77566d8209d6cb0962888a1c6f0b0b7d1/taprpc/tapchannelrpc/tapchannel.proto#L144-L152).

tapchannel/aux_traffic_shaper.go
GeorgeTsagk: Agree the `aux_features.go` approach looks better, we can also bundle things up in there if we introduce more dev specific features.

ZZiigguurraatt:
> Are you going to revert #1252 as part of this or some separate PR? We need to remove the `allow_overpay` option (
> 
> https://github.com/lightninglabs/taproot-assets/blob/723311b77566d8209d6cb0962888a1c6f0b0b7d1/taprpc/tapchannelrpc/tapchannel.proto#L144-L152
> ).

In some discussions offline yesterday it sounds like we will be depreciating this eventually, but maybe wait a release or two for all nodes to upgrade. Otherwise we will get failures if both nodes aren't aware of the new protocol. So, I don't think it will be in this PR.

GeorgeTsagk:
> In some discussions offline yesterday it sounds like we will be depreciating this eventually, but maybe wait a release or two for all nodes to upgrade. Otherwise we will get failures if both nodes aren't aware of the new protocol. So, I don't think it will be in this PR.

I'm not sure this is the case, this is a valid sanity check that should apply per quote. If a quote violates the overpayment, it will be rejected therefore not going to be used by LND.

ZZiigguurraatt:
> > In some discussions offline yesterday it sounds like we will be depreciating this eventually, but maybe wait a release or two for all nodes to upgrade. Otherwise we will get failures if both nodes aren't aware of the new protocol. So, I don't think it will be in this PR.
> 
> I'm not sure this is the case, this is a valid sanity check that should apply per quote. If a quote violates the overpayment, it will be rejected therefore not going to be used by LND.

I don't think this is generally about overpaying a quote, I think it is just about overpaying in order to meet the 354 sat minimum.

rfqmsg/records.go
```diff
@@ -135,9 +143,19 @@ func (h *Htlc) Records() []tlv.Record {
...
+	if h.NoopAdd {
+		r := tlv.NewPrimitiveRecord[lnwallet.NoopAddHtlcType](true)
+		records = append(records, r.Record())
+	}
+
 	return records
 }
 
+// SetNoopAdd flags the HTLC as a noop_add.
+func (h *Htlc) SetNoopAdd(noopActive bool) {
```
guggero, line 278: Do we need this method, since we can just set the field directly, since it's exported?

ZZiigguurraatt:
> > Are you going to revert #1252 as part of this or some separate PR? We need to remove the `allow_overpay` option (
> > https://github.com/lightninglabs/taproot-assets/blob/723311b77566d8209d6cb0962888a1c6f0b0b7d1/taprpc/tapchannelrpc/tapchannel.proto#L144-L152
> > 
> > ).
> 
> In some discussions offline yesterday it sounds like we will be depreciating this eventually, but maybe wait a release or two for all nodes to upgrade. Otherwise we will get failures if both nodes aren't aware of the new protocol. So, I don't think it will be in this PR.


We have `min_transportable_units`. Do we want to populate that if both nodes are already upgraded to this new capability? If we still want to populate it, it seems like `min_transportable_units` should always be 1 if both have upgraded. I'm not sure what litd commit to use to test this PR, so maybe you already do that.


```
0c60be93a29f:/$ litcli ln addinvoice --amt 444 --group_key 036bd8b9572f3c8de9e8cd4f6df56752e7616c3930de84744182da9315c0cfff34
{
    "accepted_buy_quote":  {
        "peer":  "024997ab7e02bd8d7cd257a0c4a846f9f099f7126422c64f0f435a47fdcd3a0b3d",
        "id":  "b47c26a4ac74906863059f7d1b5ada90aed24fc438934f01f6ef94d8ed288f3c",
        "scid":  "17793604312176824124",
        "asset_max_amount":  "54528",
        "ask_asset_rate":  {
            "coefficient":  "98522167480000000000",
            "scale":  10
        },
        "expiry":  "1752081002",
        "min_transportable_units":  "34876"
    },
    "invoice_result":  {
        "r_hash":  "295594304cbf7347242507b9c2986ae7b5a7aab9b950339eb747222f60212e00",
        "payment_request":  "lnbcrt4440n1p5xagj8pp5992egvzvhae5wfp9q7uu9xr2u766024eh9gr884hgu3z7cpp9cqqdqqcqzzsxqzfvrzjqfye02m7q27c6lxj27svf2zxl8cfnacjvs3vvnc0gddy0lwd8g9nmah0jnvw62y08sqqqqlgqqqqqqgq2qsp5pfm92s4avpk024vj5v9px9qfky60uzxpxfhexsm55smvf00jrnhs9qxpqysgq7cxnm3n4a60ecycc5cgw8cwhktuh604k3et3jtkras05k9jadk5jgk2slvmedq8eta94n6qeyemydlanp9fc234p4zq9s7l93um0szsqqywphd",
        "add_index":  "1",
        "payment_addr":  "0a765542bd606cf55592a30a131409b134fe08c1326f934374a436c4bdf21cef"
    }
}
0c60be93a29f:/$
```

rfqmsg/records.go
GeorgeTsagk: I generally prefer wrapping things in methods, even if it's trivial, as it allows for easy extensions if needed.
Just a personal preference, fine with changing it.

litbot-9000:
@roasbeef: review reminder

GeorgeTsagk:
Rebased and updated to latest ref of lnd noop HTLCs PR

tapchannel/commitment.go
```diff
@@ -162,6 +165,9 @@ func ComputeView(ourBalance, theirBalance uint64,
 		case lnwallet.Add:
 			continue
 
+		case lnwallet.NoOpAdd:
```
Roasbeef, line 176: Do we also need to replicate the awareness of how this (doesn't) impact the sat balances when evaluated? IIRC, we started to replicate more of that logic here so we derive the correct local/remote balance information.

tapchannel/aux_traffic_shaper.go
Roasbeef: Ok, I see how this fits in now, resolving this comment.

tapchannel/aux_traffic_shaper.go
```diff
@@ -569,6 +576,12 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
 	// amount that should be sent on-chain, which is a value in satoshi that
 	// is just above the dust limit.
 	htlcAmountMSat := rfqmath.DefaultOnChainHtlcMSat
+
+	// Now we set the flag that marks this HTLC as a noop_add, which means
+	// that the above dust will eventually return to us. This means that
+	// only the assets will be sent and not any btc balance.
+	htlc.SetNoopAdd(rfqmsg.UseNoOpHTLCs)
```
Roasbeef: 👍

rfq/order.go
```diff
@@ -284,6 +285,8 @@ func (c *AssetSalePolicy) GenerateInterceptorResponse(
 		fn.None[[]rfqmsg.ID](),
 	)
 
+	htlcRecord.SetNoopAdd(rfqmsg.UseNoOpHTLCs)
```
Roasbeef: Can add a comment here that: we set this to ensure that when we send on the outgoing link, we don't push any sats over if we don't need to.

tapchannel/aux_invoice_manager.go
```diff
@@ -196,7 +197,26 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(ctx context.Context,
 
 			resp.CancelSet = true
 		} else {
-			iLog.Tracef("has no asset custom records, ignoring")
+			noopTLV := uint64(lnwallet.NoOpHtlcTLVEntry.TypeVal())
+			_, isNoop := req.WireCustomRecords[noopTLV]
+
+			// If the HTLC does not carry assets and the invoice is
```
Roasbeef, line 203: Great defensive check 👍

Roasbeef (approved):
LGTM 🎓

Merge just blocked on the `lnd` PR now.

tapchannel/commitment.go
GeorgeTsagk: For the tapd side it seems that when computing views and balances we only look at the asset side of things. So given that the noop doesn't change anything on the asset level maybe can leave as is? Can just throw a comment there that describes it on a high level.

GeorgeTsagk:
Changed the LND dep to point to the latest [`v0.19.3-rc2` commit](https://github.com/lightningnetwork/lnd/pull/10152)

GeorgeTsagk:
Release note was pointing to LND PR, changed

GeorgeTsagk:
The LiT itests are expected to fail, since we don't have a way to smoothly introduce this feature yet on the channel level. That means that any itest involving backwards compatibility will lead to a channel force close, failing that test.

The corresponding [LiT PR](https://github.com/lightninglabs/lightning-terminal/pull/1080) will remove it temporarily, until we add a way to introduce new features to the channel.

GeorgeTsagk:
> The LiT itests are expected to fail, since we don't have a way to smoothly introduce this feature yet on the channel level. That means that any itest involving backwards compatibility will lead to a channel force close, failing that test.
> 
> The corresponding [LiT PR](https://github.com/lightninglabs/lightning-terminal/pull/1080) will remove it temporarily, until we add a way to introduce new features to the channel.

Ok so I changed the approach here:

The itests are now expected to pass, as the newly added `noop-htlcs` configuration option defaults to `false`.

PR #1580, opened 2025-06-02
tapfreighter: fix `OutboundParcel.Copy` and add generic `Copy` fn test

bhandras:
The main goal of this commit is to give us a tool to be able to catch regressions when extending a copyable struct. We also fix `OutboundParcel.Copy()` along the way.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	github.com/lightningnetwork/lnd/tlv v1.3.1
 	github.com/lightningnetwork/lnd/tor v1.1.6
 	github.com/ory/dockertest/v3 v3.10.0
+	github.com/pmezard/go-difflib v1.0.0
 	github.com/prometheus/client_golang v1.14.0
 	github.com/stretchr/testify v1.10.0
 	github.com/urfave/cli v1.22.14
@@ -148,7 +149,6 @@ require (
 	github.com/opencontainers/image-spec v1.1.0 // indirect
 	github.com/opencontainers/runc v1.2.0 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
-	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/prometheus/client_model v0.3.0 // indirect
 	github.com/prometheus/common v0.37.0 // indirect
 	github.com/prometheus/procfs v0.8.0 // indirect
--- a/internal/test/copy.go
+++ b/internal/test/copy.go
@@ -1,0 +1,334 @@
+package test
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/pmezard/go-difflib/difflib"
+)
+
+// FillFakeData recursively fills a struct with dummy values.
+func FillFakeData[T any](t *testing.T, debug bool, maxDepth int, v T) {
+	if t != nil {
+		t.Helper()
+	}
+
+	val := reflect.ValueOf(v)
+	name := val.Type().Elem().Name()
+	fillFakeData(t, debug, 0, maxDepth, val, name)
+}
+
+// fillFakeData is the recursive helper to fill a value with fake data.
+func fillFakeData(t *testing.T, debug bool, depth, maxDepth int,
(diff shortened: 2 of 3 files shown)


copy_test.go
```diff
@@ -0,0 +1,355 @@
...
+	"testing"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/tapfreighter"
+	"github.com/pmezard/go-difflib/difflib"
+)
+
+// FillFakeData recursively fills a struct with dummy values.
+func FillFakeData[T any](t *testing.T, debug bool, maxDepth int, v T) {
```
guggero: nit: should we move these helper functions to the `internal/test` package? Then we can more easily re-use them in other places. And the copy test could actually be in the `tapfreighter` test.

copy_test.go
```diff
@@ -0,0 +1,355 @@
...
+		}
+
+	default:
+	}
+}
+
+// CopyFnTest checks that the Copy method returns a value that:
+// 1) is deeply equal
+// 2) does not alias mutable fields (pointers, slices, maps)
+func CopyFnTest[T fn.Copyable[T]](t *testing.T, debug, strict bool,
```
guggero: nit: rename to `AssertCopyEqual()` or something like that? Was initially not very clear to me what this function really does, at least from its name.

copy_test.go
```diff
@@ -0,0 +1,355 @@
...
+		t.Helper()
+	}
+
+	val := reflect.ValueOf(v)
+	name := val.Type().Elem().Name()
+	fillFakeData(t, debug, 0, maxDepth, val, name)
+}
+
+// fillFakeData is the recursive helper to fill a value with fake data.
+func fillFakeData(t *testing.T, debug bool, depth, maxDepth int,
```
guggero: Very nice! I wish the `fuzz` library exposed some of their helpers that are used to generate such random values. But nice to have such a compact version we can use here.

tapfreighter/interface.go
```diff
@@ -458,29 +458,56 @@ type OutboundParcel struct {
...
-		Inputs:             fn.CopySlice(o.Inputs),
-		Outputs:            fn.CopySlice(o.Outputs),
+		AnchorTxHeightHint:    o.AnchorTxHeightHint,
+		AnchorTxBlockHeight:   o.AnchorTxBlockHeight,
+		TransferTime:          o.TransferTime,
+		ChainFees:             o.ChainFees,
+		Inputs:                fn.CopySlice(o.Inputs),
+		Outputs:               fn.CopySlice(o.Outputs),
+		Label:                 o.Label,
+		SkipAnchorTxBroadcast: o.SkipAnchorTxBroadcast,
```
guggero, line 468: Nice catch!

tapfreighter/interface.go
```diff
@@ -458,29 +458,56 @@ type OutboundParcel struct {
...
-			MerkleRoot:        oldAnchor.MerkleRoot,
-			TapscriptSibling:  oldAnchor.TapscriptSibling,
-			NumPassiveAssets:  oldAnchor.NumPassiveAssets,
+			OutPoint:    oldAnchor.OutPoint,
+			Value:       oldAnchor.Value,
+			InternalKey: oldAnchor.InternalKey,
+			TaprootAssetRoot: fn.CopySlice(
+				oldAnchor.TaprootAssetRoot,
+			),
+			CommitmentVersion: &commitmentVersion,
```
guggero: This should remain nil (instead of becoming a pointer to a `0` value) if the original is nil.

tapfreighter/interface.go
bhandras: The initial bug that i found was that subscribing to send events did not correctly fill the confirmation height upon finished transfer.

tapfreighter/interface.go
bhandras: Thanks, indeed! Done.

copy_test.go
bhandras: Good idea, done!

copy_test.go
bhandras: Agreed, ideally something similar should be in the standard library.

copy_test.go
bhandras: Done.

PR #1583, opened 2025-06-04
channels: check RFQ quote compatibility with channel at pathfind time

guggero:
This fixes an edge case in the pathfinding logic of `lnd` (reported on Slack):

1. There are two channels with different assets:

```
2025-06-03 12:23:32.439 [DBG] HSWC: ShortChannelID=3607099:1470:0: aux traffic shaper reported available bandwidth: 53340397469 mSAT (<-- asset1)

2025-06-03 12:23:32.437 [DBG] HSWC: ShortChannelID=4148366:42:0: aux traffic shaper reported available bandwidth: 41166477370 mSAT (<-- asset2)
```

2. When trying to send a payment using `asset2`, both channels report bandwidth as shown above, because we didn't have the check in this PR.
3. Because the channel with the **wrong** asset reported a higher available bandwidth, `lnd` skipped the second channel with the log message `2025-06-03 12:23:32.296 [DBG] CRTR: Skipped edge 4561176653273366528: not max bandwidth, bandwidth=41166477370 mSAT, maxBandwidth=53340397469 mSAT`
4. The HTLC is sent over the wrong channel, then rejected in a later stage when we actually check the compatibility
5. The pathfinding logic doesn't know what's going on and re-tries over and over again, using high CPU

This PR breaks this cycle by not reporting any bandwidth for the first channel, since from the quote we can figure out the channel isn't actually compatible.

### DIFF
--- a/tapchannel/aux_traffic_shaper.go
+++ b/tapchannel/aux_traffic_shaper.go
@@ -1,6 +1,7 @@
 package tapchannel
 
 import (
+	"context"
 	"fmt"
 	"sync"
 
@@ -228,7 +229,7 @@ func (s *AuxTrafficShaper) PaymentBandwidth(fundingBlob, htlcBlob,
 	// Otherwise, we derive the available bandwidth from the HTLC's RFQ and
 	// the asset units in our local balance.
 	return s.paymentBandwidth(
-		htlc, computedLocal, linkBandwidth, minHtlcAmt,
+		htlc, computedLocal, linkBandwidth, minHtlcAmt, fundingChan,
 	)
 }
 
@@ -279,8 +280,8 @@ func paymentBandwidthAssetUnits(htlcAssetAmount, computedLocal uint64,
 // on the asset rate of the RFQ quote that is included in the HTLC and the asset
 // units of the local balance.
 func (s *AuxTrafficShaper) paymentBandwidth(htlc *rfqmsg.Htlc,
-	localBalance uint64, linkBandwidth,
-	minHtlcAmt lnwire.MilliSatoshi) (lnwire.MilliSatoshi, error) {
+	localBalance uint64, linkBandwidth, minHtlcAmt lnwire.MilliSatoshi,
+	fundingChan *cmsg.OpenChannel) (lnwire.MilliSatoshi, error) {
 
 	// If the HTLC doesn't have an RFQ ID, it's incomplete, and we cannot
 	// determine the bandwidth.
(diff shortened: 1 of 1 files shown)


tapchannelmsg/wire_msgs_test.go
```diff
@@ -258,3 +263,132 @@ func TestProofChunkErrorCases(t *testing.T) {
 		)
 	})
 }
+
+// ExtractHexDump extracts the hex bytes from a hex dump string, which is
+// typically formatted with an offset, hex bytes, and ASCII representation.
+func ExtractHexDump(input string) ([]byte, error) {
```
ffranr: hex bytes?

```
// ExtractHexDump parses a hex dump string (typically formatted with offsets,
// hex-encoded byte values, and ASCII representation) and returns the decoded bytes.
```

guggero:
I've verified this by adding a commit to https://github.com/lightninglabs/lightning-terminal/pull/1082, which currently makes the tests fail.
Once this PR is merged and the `go.mod` in the litd PR is updated, the tests will succeed again.

tapchannelmsg/wire_msgs_test.go
ffranr: Also, how were these `.hexdump` files generated? Maybe we should write that down somewhere. A `README.md` in `testdata`?

tapchannelmsg/records.go
```diff
@@ -229,6 +229,34 @@ func (o *OpenChannel) HasAllAssetIDs(ids fn.Set[asset.ID]) bool {
...
 
+// Compatible checks if the OpenChannel is compatible with the given asset
+// specifier.
+func (o *OpenChannel) Compatible(specifier asset.Specifier) bool {
+	switch {
+	case specifier.HasGroupPubKey():
+		targetGroupKey := specifier.UnwrapGroupKeyToPtr()
+		if targetGroupKey == nil {
+			return false
+		}
```
ffranr: if the specifier has a group pub key but then we end up with `targetGroupKey == nil`, that's an error. I don't think it means that the channel is incompatible with the specifier necessarily because asset ID might still match.

maybe use `specifier.UnwrapGroupKeyOrErr` here and return an error?

guggero: We check with `specifier.HasGroupPubKey()` above, so the nil case really should never happen. It's just defensive. I don't think it makes sense to return an error just for that.

tapchannelmsg/wire_msgs_test.go
guggero: I mentioned in the commit message that these are helpers to decode hex dumps from a trace log of lnd/litd. But I guess it's useful to have that in the Godoc comment too.

tapchannelmsg/records.go
```diff
@@ -229,6 +229,34 @@ func (o *OpenChannel) HasAllAssetIDs(ids fn.Set[asset.ID]) bool {
...
+
+		// If the specifier has a group key, then we must have a group
+		// key set on the OpenChannel.
+		return lfn.MapOptionZ(
+			o.GroupKey.ValOpt(),
```
GeorgeTsagk: I'm not sure this is correct, if the specifier has a group key, we could still use the channel even if it was funded with a single asset ID, if that asset ID is part of the group

tapchannelmsg/records.go
```diff
@@ -229,6 +229,34 @@ func (o *OpenChannel) HasAllAssetIDs(ids fn.Set[asset.ID]) bool {
...
+			func(groupKey *btcec.PublicKey) bool {
+				return targetGroupKey.IsEqual(groupKey)
```
GeorgeTsagk: instead we should run a check against all assets in `OpenChannel` and see if they satisfy `AssetMatchesSpecifier`

tapchannelmsg/records.go
guggero: But the specifier here is what was used by the user to create the quote. And if the quote was created for a group key, then the channel needs to have a group defined as well.
If the user creates a quote for a single asset ID in a grouped channel that only has that asset ID, then the specifier here will have the asset ID.

guggero: Hmm, but I guess just re-using the `AssetMatchesSpecifier` function does make sense... Will try it.

tapchannelmsg/records.go
guggero: This approach makes more sense, thanks!

tapchannel/aux_traffic_shaper.go
```diff
@@ -301,19 +302,54 @@ func (s *AuxTrafficShaper) paymentBandwidth(htlc *rfqmsg.Htlc,
...
+		)
+		if err != nil {
+			return 0, fmt.Errorf("error checking if asset ID %x "+
+				"matches specifier %s: %w", b.AssetID.Val[:],
+				specifier.String(), err)
+		}
+
+		// One of the asset IDs in the channel does not match the quote,
+		// we don't want to route this HTLC over this channel.
+		if !match {
```
GeorgeTsagk, line 345: :+1:

PR #1588, opened 2025-06-09
rpc: Replace AssetMeta with FetchAssetMetaResponse in FetchAssetMeta RPC endpoint

ffranr:
This PR introduces a breaking change, replacing the `AssetMeta` proto message with `FetchAssetMetaResponse` as the response for the `FetchAssetMeta` RPC endpoint.

This decouples the setting of asset metadata during minting from its retrieval, preventing users from setting tapd-controlled fields and removing duplicate fields from the minting interface. The new response now includes default supported metadata fields, allowing for the retrieval of necessary information like the delegation key during supply commitment testing.


ffranr:
I've moved these commits into https://github.com/lightninglabs/taproot-assets/pull/1589

ffranr:
I've moved these commits into https://github.com/lightninglabs/taproot-assets/pull/1589

Issue #1586, opened 2025-06-06
[feature]: add tests to verify correct pre-commitment output generation as part of supply commit feature


ffranr:
Add test coverage to ensure that pre-commitment outputs are generated correctly as part of the supply commit feature.

PR #1589, opened 2025-06-09
Refactor: Split Pre-Commitment Logic to Support New Asset Groups

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/1586

---

### Problem

The entire pre-commitment process—both adding the output to the transaction and linking it to an asset group—was executed in the `funding` step. This approach was flawed because it incorrectly assumed the asset group already existed on disk. This caused a failure when handling universe commitments for newly created assets, as the required group data was not yet available.

### Solution

This PR refactors the logic to align with the minting batch lifecycle, separating transaction construction from data association.

* **During the `funding` step:** The pre-commitment output is still created and added to the minting transaction, as before. This part of the logic, which deals with the transaction itself, in principle remains unchanged.

* **During the `seal` step:** We now introduce logic to `upsert` the relationship between the pre-commitment output and the asset group. This association is intentionally deferred to the `seal` step because this is the first point at which we can guarantee a new asset group has been successfully created and its data is available.

This separation of concerns ensures the feature works reliably for both new and existing asset groups.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1589)

### DIFF
--- a/itest/supply_commit_test.go
+++ b/itest/supply_commit_test.go
@@ -1,0 +1,134 @@
+package itest
+
+import (
+	"bytes"
+	"context"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/tapgarden"
+	"github.com/lightninglabs/taproot-assets/taprpc"
+	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	"github.com/stretchr/testify/require"
+)
+
+// assertAnchorTxPreCommitOut checks that the anchor transaction for the
+// minted asset includes a pre-commitment output for the supply commitment.
+// If an expected delegation key is provided, it verifies that it matches
+// the one used in the pre-commitment output. The function returns the
+// delegation key found in the asset metadata.
+func assertAnchorTxPreCommitOut(
+	t *harnessTest, tapd *tapdHarness, rpcAsset *taprpc.Asset,
+	expectedDelegationKey fn.Option[btcec.PublicKey]) btcec.PublicKey {
+
+	// Fetch metadata for the minted asset.
+	ctxb := context.Background()
+
+	metaResp, err := tapd.FetchAssetMeta(
+		ctxb, &taprpc.FetchAssetMetaRequest{
+			Asset: &taprpc.FetchAssetMetaRequest_AssetId{
+				AssetId: rpcAsset.AssetGenesis.AssetId,
+			},
+		},
+	)
+	require.NoError(t.t, err)
+
+	delegationKey, err := btcec.ParsePubKey(metaResp.DelegationKey)
+	require.NoError(t.t, err)
+
+	// If a specific delegation key is expected, verify it matches the one
+	// retrieved from the asset metadata.
+	expectedDelegationKey.WhenSome(func(expectedKey btcec.PublicKey) {
+		require.True(t.t, expectedKey.IsEqual(delegationKey))
+	})
+
+	// Parse anchor tx and confirm that one output is a supply commitment
+	// pre-commitment output.
+	var msgTx wire.MsgTx
+	err = msgTx.Deserialize(
+		bytes.NewReader(rpcAsset.ChainAnchor.AnchorTx),
+	)
+	require.NoError(t.t, err)
+
+	expectedTxOut, err := tapgarden.PreCommitTxOut(*delegationKey)
+	require.NoError(t.t, err)
+
+	// The pre-commitment output should be present in the anchor tx exactly
+	// once.
+	foundOnce := false
+	for idx := range msgTx.TxOut {
+		txOut := msgTx.TxOut[idx]
+		if txOut.Value != expectedTxOut.Value {
+			continue
+		}
+		if !bytes.Equal(txOut.PkScript, expectedTxOut.PkScript) {
+			continue
+		}
+
+		// We found a pre-commitment output, but it should only be
+		// present once.
+		if foundOnce {
+			t.t.Fatalf("found pre-commitment output more than once")
+		}
+
+		foundOnce = true
+	}
+	require.True(t.t, foundOnce)
+
+	return *delegationKey
+}
+
+// testPreCommitOutput tests that the pre-commitment output is correctly
+// included in the anchor transaction when minting an asset group with
+// universe/supply commitments enabled.
+func testPreCommitOutput(t *harnessTest) {
+	mintReq := CopyRequest(issuableAssets[0])
+	mintReq.Asset.UniverseCommitments = true
+	rpcAssets := MintAssetsConfirmBatch(
+		t.t, t.lndHarness.Miner().Client, t.tapd,
+		[]*mintrpc.MintAssetRequest{mintReq},
+	)
+	require.Len(t.t, rpcAssets, 1, "expected one minted asset")
+
+	rpcFirstTrancheAsset := rpcAssets[0]
+	delegationKey := assertAnchorTxPreCommitOut(
+		t, t.tapd, rpcFirstTrancheAsset, fn.None[btcec.PublicKey](),
+	)
+
+	// Mint another tranche into the same asset group to ensure that
+	// the pre-commitment output is still included in the anchor tx when a
+	// pre-existing asset group key is used.
+	tweakedGroupKey := rpcFirstTrancheAsset.AssetGroup.TweakedGroupKey
+
+	mintReq = &mintrpc.MintAssetRequest{
+		Asset: &mintrpc.MintAsset{
+			AssetType: taprpc.AssetType_NORMAL,
+			Name:      "itestbuxx-money-printer-brrr-tranche-2",
+			AssetMeta: &taprpc.AssetMeta{
+				Data: []byte("some metadata"),
+			},
+			Amount:          6000,
+			AssetVersion:    taprpc.AssetVersion_ASSET_VERSION_V1,
+			NewGroupedAsset: false,
+			GroupedAsset:    true,
+			GroupKey:        tweakedGroupKey,
+
+			UniverseCommitments: true,
+		},
+	}
+	rpcAssets = MintAssetsConfirmBatch(
+		t.t, t.lndHarness.Miner().Client, t.tapd,
+		[]*mintrpc.MintAssetRequest{mintReq},
+	)
+
+	rpcSecondTrancheAsset := rpcAssets[0]
+
+	assertAnchorTxPreCommitOut(
+		t, t.tapd, rpcSecondTrancheAsset, fn.Some(delegationKey),
+	)
+
+	secondAssetGroupKey := rpcSecondTrancheAsset.AssetGroup.TweakedGroupKey
+	// Ensure that the second tranche asset is part of the same group.
+	require.EqualValues(t.t, tweakedGroupKey, secondAssetGroupKey)
+}
--- a/itest/test_list_on_test.go
+++ b/itest/test_list_on_test.go
@@ -343,6 +343,10 @@ var testCases = []*testCase{
 		name: "asset signing after lnd restore from seed",
 		test: testRestoreLndFromSeed,
 	},
+	{
+		name: "pre commit output",
+		test: testPreCommitOutput,
+	},
 }
 
 var optionalTestCases = []*testCase{
--- a/itest/utils.go
+++ b/itest/utils.go
@@ -532,6 +532,9 @@ func MintAssetsConfirmBatch(t *testing.T, minerClient *rpcclient.Client,
 	assetRequests []*mintrpc.MintAssetRequest,
 	opts ...MintOption) []*taprpc.Asset {
 
+	// Sanity check that we have at least one asset request.
+	require.NotEmpty(t, assetRequests)
+
 	options := DefaultMintOptions()
 	for _, opt := range opts {
 		opt(options)


tapgarden/batch.go
```diff
@@ -101,6 +101,7 @@ func (m *MintingBatch) Copy() *MintingBatch {
 		// set, so a shallow copy is sufficient.
 		BatchKey:            m.BatchKey,
 		RootAssetCommitment: m.RootAssetCommitment,
+		UniverseCommitments: m.UniverseCommitments,
```
Roasbeef, line 113: 👍

tapgarden/planter.go
```diff
@@ -2976,14 +3043,20 @@ type PreCommitmentOutput struct {
...
-	// GroupPubKey is the asset group public key associated with this
-	// pre-commitment output.
-	GroupPubKey btcec.PublicKey
+	// GroupPubKey is the asset-group public key for this pre-commitment.
+	//
+	// Optional:
+	//   - Present when the group key is already known—either reused from an
+	//     existing group at funding time or generated once the batch is
+	//     sealed.
+	//   - Absent while an unsealed batch without a prior group key is still
```
Roasbeef, line 3177: So this handles the case where you're creating the very first tranche, that's unconfirmed, but then you attempt to create a _subsequent_ tranche?

tapgarden/planter.go
```diff
@@ -2671,10 +2671,13 @@ func (c *ChainPlanter) prepSeedlingDelegationKey(ctx context.Context,
 func (c *ChainPlanter) prepAssetSeedling(ctx context.Context,
 	req *Seedling) error {
 
-	// Finalise the seedling delegation key.
-	err := c.prepSeedlingDelegationKey(ctx, req)
-	if err != nil {
-		return err
+	// If the seedling has the universe/supply commitment feature enabled,
+	// finalize the delegation key.
+	if req.UniverseCommitments {
```
Roasbeef, line 2797: 👍

tapdb/sqlc/queries/assets.sql
```diff
@@ -1035,21 +1035,40 @@ JOIN assets_meta
...
     -- The following fields are updated if a conflict occurs.
     taproot_internal_key = EXCLUDED.taproot_internal_key,
     group_key = EXCLUDED.group_key
 RETURNING id;
 
--- Fetch a record from the mint_anchor_uni_commitments table by id.
+-- Fetch records from the mint_anchor_uni_commitments table by batch key.
 -- name: FetchMintAnchorUniCommitment :one
-SELECT id, batch_id, tx_output_index, taproot_internal_key, group_key
+WITH target_batch AS (
```
Roasbeef: 👍

tapgarden/planter.go
```diff
@@ -2431,13 +2431,69 @@ func (c *ChainPlanter) sealBatch(ctx context.Context, params SealParams,
...
+			return nil, err
+		}
+
+		if preCommitDesc.GroupPubKey.IsSome() {
+			return nil, fmt.Errorf("pre-commitment output " +
+				"group key is unexpectedly set, cannot " +
+				"update mint anchor pre-commitment output")
+		}
+
+		preCommitDesc.GroupPubKey = fn.Some(groupKey)
```
Roasbeef: A this point, won't the group key _always_ be known? Since this is the OG group key and not one of the tweaked versions that we use to ensure commitment uniqueness elsewhere.

tapdb/asset_minting.go
```diff
@@ -1409,13 +1409,75 @@ func (a *AssetMintingStore) CommitBatchTx(ctx context.Context,
...
+		return err
+	}
+
+	// Serialize batch key.
+	batchKey := batch.BatchKeyBytes()
+	if len(batchKey) == 0 {
+		return fmt.Errorf("batch key is empty")
+	}
+
+	// Serialize the internal key.
```
Roasbeef: Comments here above above seem somewhat marginal.

tapgarden/batch.go
```diff
@@ -429,12 +429,6 @@ func (m *MintingBatch) validateUniCommitment(newSeedling Seedling) error {
 				"with the universe commitment flag enabled")
 		}
 
-		if !newSeedling.HasGroupKey() {
```
Roasbeef, line 422: > At the time of validation, the group key may not yet exist and could be
generated later during batch sealing.

How can the group key be generated later on? We need to know the group key upfront to create any of the set of relevant commitments.

itest/assertions.go
```diff
@@ -292,20 +292,32 @@ func GroupAssetsByName(assets []*taprpc.Asset) map[string][]*taprpc.Asset {
...
-				err := check(rpcAsset)
-				require.NoError(t, err)
+		switch {
+		case len(assets[name]) == 1:
+			a = rpcAsset
+		default:
+			// If there are more than one asset with the same
+			// name, we need to check the metadata hash to find the
+			// correct one.
+			if bytes.Equal(rpcGen.MetaHash, metaHash[:]) {
```
Roasbeef, line 310: Does this change violate the assumptions of any of the other tests?

itest/supply_commit_test.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	"github.com/stretchr/testify/require"
+)
+
+func testPreCommitOutput(t *harnessTest) {
+	t.Log("Minting asset group with a single normal asset and " +
+		"universe/supply commitments enabled")
+	mintReq := issuableAssets[0]
+	mintReq.Asset.UniverseCommitments = true
```
Roasbeef: Before this would panic somewhere?

tapgarden/batch.go
ffranr: Previously, the `if !newSeedling.HasGroupKey()` check within `ChainPlanter.prepAssetSeedling` was used to verify that a seedling had its `GroupInfo` set upon being added to a batch. This worked correctly for seedlings requested through `c.seedlingReqs` that already had a group key. However, this check created an issue for seedlings whose group key is generated later, during the batch seal step. In this scenario, the seedlings would not yet have their `GroupInfo` set, causing the `HasGroupKey()` check to fail incorrectly.

tapdb/asset_minting.go
ffranr: I'll remove.

tapgarden/planter.go
ffranr: Yes, I agree. I will remove the `IsSome` check. If think it would have block reissuance with pre-commitment anyway.

tapgarden/planter.go
ffranr: Yes, it can handle issuing into an existing asset group and also creating a new asset group.

itest/assertions.go
ffranr: This check was somewhat odd. The assertion logic attempted to reconstruct what it believed the asset’s metadata reveal should be, then asserted that the resulting metadata hash matched. However, if the supply commitment delegation key is generated during minting and not specified in the minting request, the assertion logic can't reliably know the full metadata reveal in advance. We need to clarify what this metadata hash assertion is actually verifying, given that some metadata fields are only determined after minting completes.

itest/supply_commit_test.go
ffranr: I think it would just error during validation because the `GroupInfo` was expected to be set early, which was impossible when minting a new asset group.

tapgarden/batch.go
ffranr: I've revised this once more in the latest commits. It's now:
```
if !newSeedling.EnableEmission && !newSeedling.HasGroupKey() {
			return fmt.Errorf("universe commitment enabled: " +
				"seedling must either create a new asset " +
				"group or issue into an existing one")
}
```

which i think is what we want.

ffranr:
Failing itest is: testMintFundSealAssets. It's flakey, it doesn't always fail.
Earliest commit I've been able to reproduce so far:
```
tapdb+tapgarden: update pre-commit output db entry as part of batch seal ffranr 06/06/2025, 23:57 39674350
```
Error message:
```
2025-06-11 17:15:55.399 [ERR] GRDN caretaker.go:390: unable to advance state machine: unable to construct minting proofs: invalid proof file generated: group key reveal doesn't match group key
```
It fails locally and in github CI.

tapgarden/batch.go
ffranr: Done.

tapgarden/planter.go
ffranr: Done.

tapdb/sqlc/queries/assets.sql
ffranr: Done.

tapgarden/batch.go
ffranr: (This comment came from me pressing "Done" on Reviewable, BTW.)

rpcserver.go
```diff
@@ -4807,11 +4815,53 @@ func (r *rpcServer) FetchAssetMeta(ctx context.Context,
...
+		assetMeta.CanonicalUniverses, func(urls []url.URL) []string {
+			urlStrings := make([]string, 0, len(urls))
+			for idx := range urls {
+				urlAddr := urls[idx]
+				urlStrings = append(
+					urlStrings, urlAddr.String(),
+				)
+			}
+
+			return urlStrings
```
guggero: Use:
```
			return fn.Map(urls, func(urlAddr url.URL) string {
				return urlAddr.String()
			})
```
?

rpcserver.go
```diff
@@ -4807,11 +4815,53 @@ func (r *rpcServer) FetchAssetMeta(ctx context.Context,
...
+	)
+
 	metaHash := assetMeta.MetaHash()
-	return &taprpc.AssetMeta{
-		Data:     assetMeta.Data,
-		Type:     taprpc.AssetMetaType(assetMeta.Type),
-		MetaHash: metaHash[:],
+	return &taprpc.FetchAssetMetaResponse{
+		Data:                  assetMeta.Data,
+		Type:                  decDisplay,
```
guggero: Type should be `taprpc.AssetMetaType(assetMeta.Type),`!

itest/asset_meta_test.go
```diff
@@ -219,6 +219,12 @@ func testMintAssetWithDecimalDisplayMetaField(t *harnessTest) {
 	require.Contains(t.t, string(metaResp.Data), `"foo":"bar"`)
 	require.Contains(t.t, string(metaResp.Data), `"decimal_display":2`)
 
+	require.Empty(t.t, metaResp.UnknownOddTypes)
+	require.EqualValues(t.t, metaResp.DecimalDisplay, 2)
```
guggero: nit: order of expected vs. actual values.

Also, we should assert the type here too, which would've caught my previous comment.

tapgarden/batch.go
guggero, line 113: Can we please also add a copy test here, analog to https://github.com/lightninglabs/taproot-assets/blob/7ce15016366c38df1670ce6e1928ffa4b30fe9b7/tapfreighter/copy_test.go#L10?
So we can catch new fields not being copied in the future?

tapgarden/caretaker.go
```diff
@@ -596,12 +590,18 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
...
+		genesisPkt := b.cfg.Batch.GenesisPacket
 
-		changeOutputIndex := genesisPacket.ChangeOutputIndex
-		b.anchorOutputIndex = genesisPacket.AssetAnchorOutIdx
+		changeOutputIndex := genesisPkt.ChangeOutputIndex
+		b.anchorOutputIndex = genesisPkt.AssetAnchorOutIdx
 
-		genesisPoint := extractGenesisOutpoint(genesisTxPkt.UnsignedTx)
+		genesisPoint, err := genesisPkt.GenesisOutpoint().UnwrapOrErr(
+			fmt.Errorf("batch genesis outpoint is missing from " +
```
guggero: Since we have a very similar error that's being used 4 times, does it make sense to extract it into a declared error variable? Which would also make this a bit shorter.

tapgarden/planter.go
```diff
@@ -3036,6 +3054,26 @@ func NewFundedMintAnchorPsbt(
 	}, nil
 }
 
+// GenesisOutpoint returns the genesis outpoint of the mint anchor PSBT, which
+// is the first input in the genesis transaction.
+func (f *FundedMintAnchorPsbt) GenesisOutpoint() fn.Option[wire.OutPoint] {
```
guggero, line 3257: Since we're calling this method exclusively in combination with `UnwrapOrErr` does it really make sense to have an option here? IMO it could be `(*wire.OutPoint, error)` directly, or `fn.Result[wire.OutPoint]`.

tapdb/asset_minting.go
```diff
@@ -244,10 +248,10 @@ type PendingAssetStore interface {
 	FetchAssetMetaForAsset(ctx context.Context,
 		assetID []byte) (sqlc.FetchAssetMetaForAssetRow, error)
 
-	// FetchMintAnchorUniCommitment fetches the mint anchor uni commitment
-	// for a given batch.
+	// FetchMintAnchorUniCommitment fetches mint anchor pre-commitments.
 	FetchMintAnchorUniCommitment(ctx context.Context,
-		batchKey []byte) (sqlc.FetchMintAnchorUniCommitmentRow, error)
+		arg sqlc.FetchMintAnchorUniCommitmentParams) (
```
guggero: nit: use type aliases here to shorten slightly?

tapdb/sqlc/queries/assets.sql
```diff
@@ -1046,29 +1052,40 @@ WITH target_batch AS (
...
+    mint_anchor_uni_commitments.tx_output_index,
+    mint_anchor_uni_commitments.taproot_internal_key_id,
+    mint_anchor_uni_commitments.group_key,
+    batch_internal_keys.raw_key AS batch_key,
+    taproot_internal_keys.raw_key AS taproot_internal_key_raw,
+    taproot_internal_keys.key_family AS taproot_internal_key_family,
+    taproot_internal_keys.key_index AS taproot_internal_key_index
 FROM mint_anchor_uni_commitments
-WHERE batch_id = (SELECT batch_id FROM target_batch);
+    LEFT JOIN internal_keys taproot_internal_keys
```
guggero: Why is this a left join? Is it possible that a uni commitment doesn't have an internal key? If we make this a normal JOIN, then we can use `sqlc.embed()` and simplify the key parsing (create a function similar to `parseScriptKey()` that just parses an internal key).

tapdb/sqlc/migrations/000036_mint_supply_commit_key_refs.up.sql
```diff
@@ -0,0 +1,15 @@
+-- Add delegation_key_id column to asset_seedlings table as a foreign key
+-- to internal_keys table
+ALTER TABLE asset_seedlings
+ADD COLUMN delegation_key_id
+BIGINT REFERENCES internal_keys(key_id);
+
+-- Replace taproot_internal_key column with taproot_internal_key_id 
+-- as a foreign key to internal_keys table in mint_anchor_uni_commitments
```
guggero: Should we mention here why we're not doing a data migration from one column to the other? That we don't expect anyone to have any data in these tables yet?

tapdb/asset_minting.go
```diff
@@ -1228,23 +1242,70 @@ func marshalMintingBatch(ctx context.Context, q PendingAssetStore,
...
+					"pre-commitment output, got %d",
+					len(fetchRes))
+			}
+
+			res := fetchRes[0]
+
 			// Parse the internal key from the database.
-			internalKey, err := btcec.ParsePubKey(
-				res.TaprootInternalKey,
+			if res.TaprootInternalKeyRaw == nil {
```
guggero: Can avoid a lot of these `if` by not making the internal key a LEFT join (see my other comment).

tapgarden/planter.go
```diff
@@ -2203,6 +2203,69 @@ func matchPsbtToGroupReq(psbt psbt.Packet,
...
+	preCommitDesc, err :=
+		fundedAnchor.PreCommitmentOutput.UnwrapOrErr(
+			fmt.Errorf("pre-commitment output is unexpectedly " +
+				"absent"),
+		)
+	if err != nil {
+		return err
+	}
+
+	preCommitDesc.GroupPubKey = fn.Some(groupKey)
```
guggero: Why do we unpack this when we put it back as an option? If we just want to validate it's there, we can do `if !groupKeyOpt.IsSome()`. Same for the preCommitDesc. Probably doesn't matter too much, but might visually simplify the logic a little bit.

tapdb/asset_minting.go
```diff
@@ -1470,6 +1470,70 @@ func (a *AssetMintingStore) CommitBatchTx(ctx context.Context,
...
+			},
+			PubKey: pubKey,
+		}
+
+		delegationKey = fn.Some(dKey)
+
+		return nil
+	})
+	if dbErr != nil {
+		return zero, dbErr
```
guggero, line 1480: At this point `zero` and `delegationKey` should have the same, empty value. I think we can simplify to just use one.

tapdb/asset_minting.go
```diff
@@ -1470,6 +1470,70 @@ func (a *AssetMintingStore) CommitBatchTx(ctx context.Context,
...
+		}
+
+		// If we didn't find any pre-commitment outputs, then
+		// we can return early.
+		if len(fetchRow) == 0 {
+			return nil
+		}
+
+		// If there are multiple pre-commitment outputs, just choose any
+		// single one, we assume they all have the same delegation key.
```
guggero: Is this something we need to validate? Or was this validated at another point and we can make a more or less safe assumption here?

guggero: Ah, I guess this is more or less what this whole commit is doing. So perhaps re-formulate this comment to "we choose the first one because we can assume all commitment outputs of the same group have the same delegation key"?

itest/supply_commit_test.go
```diff
@@ -0,0 +1,134 @@
...
+	require.True(t.t, foundOnce)
+
+	return *delegationKey
+}
+
+// testPreCommitOutput tests that the pre-commitment output is correctly
+// included in the anchor transaction when minting an asset group with
+// universe/supply commitments enabled.
+func testPreCommitOutput(t *harnessTest) {
+	mintReq := issuableAssets[0]
```
guggero: Use `mintReq := CopyRequest(issuableAssets[0])` here to make sure we don't influence the following tests?

guggero: I know, these probably shouldn't have been pointers in the first place...

tapgarden/planter.go
ffranr: My point of view: just because a missing outpoint is an error at the call sites doesn't mean that it's an error in `FundedMintAnchorPsbt.GenesisOutpoint`. So what the caller sees as an error isn't necessarily an error in the function being called.

tapdb/asset_minting.go
ffranr:
```
if dbErr != nil {
    return delegationKey, dbErr
}
```

^ this looks odd to me because it looks like we're trying to return a delegation key and an error. That's why I use `return zero, dbErr`

tapdb/asset_minting.go
ffranr: Yes, for now the user doesn't have the option to specify a delegation key for a mint event — we reuse the existing one if present. We can extend the feature later to support specifying a new delegation key.

I've updated the comment.

tapgarden/planter.go
ffranr: I've gone with:
```
	// Formulate the pre-commitment output descriptor with the group key.
	preCommitDesc := fn.MapOptionZ(
		fundedAnchor.PreCommitmentOutput,
		// nolint: lll
		func(preCommit PreCommitmentOutput) fn.Option[PreCommitmentOutput] {
			preCommit.GroupPubKey = fn.Some(groupKey)
			return fn.Some(preCommit)
		},
	)
```
does that work for you?

tapdb/sqlc/migrations/000036_mint_supply_commit_key_refs.up.sql
ffranr: I mentioned in the commit message but will include here also, good idea.

tapdb/sqlc/queries/assets.sql
ffranr: you're right, internal key must be present. I've removed the `LEFT JOIN` like you suggest and embedded the internal key in the query response.

ffranr:
rebased because I thought SQL migration might conflict, should be mergeable in latest state.

ffranr:
I don't think [run LiT tests](https://github.com/lightninglabs/taproot-assets/actions/runs/15825565932/job/44611118714?pr=1589#logs) fail is related to this PR.

Roasbeef (approved):
:lgtm: 

> Reviewed 6 of 6 files at r1, 1 of 19 files at r2, 15 of 15 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1589#-:-OTUWrLd8GJDTrDMIsD2:bg7igkv)* status: all files reviewed, 28 unresolved discussions (waiting on @ffranr and @guggero)

Issue #1504, opened 2025-05-01
[feature]: `FetchAssetMeta` should accept `group_key` as an input alternative to `asset_id`


ZZiigguurraatt:
`FetchAssetMeta` currently requires one of the following

https://github.com/lightninglabs/taproot-assets/blob/e621beefce9d80a5fed091b52534876c23db5e15/taprpc/taprootassets.proto#L1301-L1317

as in input.

We should also accept `group_key` as there can be many `asset_id` in a group. Different each `asset_id` in a group can have different metadata, so we should return an array of metadata for all `asset_id` in the `group_key`.

However, there is a special case of data that is normally embedded in the metadata, the `decimal_display`. This is not supposed to change, so we should decode and return that value in a separate property. See also, https://github.com/lightninglabs/taproot-assets/issues/1503 . Alternatively, if we don't want to include the group's `decimal_display` value in the response to `FetchAssetMeta`, we should create a new RPC that does.

PR #1591, opened 2025-06-09
rpc: update FetchAssetMeta to allow fetch by group key

Roasbeef:
This commit refactors the `FetchAssetMeta` RPC implementation in `rpcserver.go`. The main function now uses a top-level switch statement that dispatches to dedicated helper functions for handling requests by asset ID, meta hash, and group key.

The following helper functions were introduced:
- `fetchMetaByAssetID`
- `fetchMetaByMetaHash`
- `fetchMetaByGroupKey`

Each helper encapsulates the logic for fetching and processing asset metadata based on the specific identifier type (including handling both raw byte and hex-encoded string inputs via the main dispatcher).

This change improves code organization and readability of the `FetchAssetMeta` method without altering its external behavior.

Unit tests in `rpcserver_test.go` have been reviewed and updated to ensure continued coverage of all cases, including new sub-tests for raw byte inputs for each identifier type.

Fixes https://github.com/lightninglabs/taproot-assets/issues/1504


Roasbeef:
Looks like it updated the `protoc` version along the way, so a lot of the diff is small updates in all the generated files.

PR #1592, opened 2025-06-09
rpc: allow minting assets with empty metadata

Roasbeef:
In this PR, we allow minting assets with an empty asset meta, and also permit minting an asset with a decimal display specified, but not asset meta. Both of these flows we previously disallowed. From v0.5 and prior, we allowed minting with an empty asset meta, but some stricter validation we added prevented it. 

Without these series of changes, a user must _always_ specify an asset meta when minting, even if they don't intend to using the meta feature at all. We also now permit just a decimal display without an asset meta for a similar reason. After moving the asset meta to a top level TLV, we no longer need to serialize a JSON blob.

### DIFF
--- a/itest/asset_meta_test.go
+++ b/itest/asset_meta_test.go
@@ -197,18 +197,9 @@ func testMintAssetWithDecimalDisplayMetaField(t *harnessTest) {
 	_, err = t.tapd.MintAsset(ctxt, secondAssetReq)
 	require.ErrorContains(t.t, err, "decimal display does not match")
 
-	// Requesting a decimal display without specifying the metadata field
-	// with at least the type should fail.
-	secondAssetReq.Asset.DecimalDisplay = firstAsset.DecimalDisplay
-	secondAssetReq.Asset.AssetMeta = nil
-
-	_, err = t.tapd.MintAsset(ctxt, secondAssetReq)
-	require.ErrorContains(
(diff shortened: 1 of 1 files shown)


rpcserver.go
```diff
@@ -533,6 +533,14 @@ func (r *rpcServer) MintAsset(ctx context.Context,
 			return nil, err
 		}
 
+		// If the meta type is JSON and no data is provided, default to
+		// an empty JSON object.
+		if metaType == proof.MetaJson &&
+			len(req.Asset.AssetMeta.Data) == 0 {
+
+			req.Asset.AssetMeta.Data = []byte("{}")
```
Roasbeef: Was this the prior behavior? Or did we act stricter checks?

Roasbeef:
Revamped the approach. 

We know allow users to:
  * specify no asset meta at all 
  * specify a decimal display, but no asset meta 

The issue was that `meta.Validate()` will call `IsValidMetaSize`. This means a user _must_ always specify an asset meta, even if they don't want to use the feature. I've fixed this just by allowing a zero length asset meta. Related to this is that we'll now encode the dec display as a TLV in the asset meta. A user should be able to specify a dec display, but not also a custom meta data. To allow this, I've relaxed the RPC server logic.

ffranr (approved):

PR #1596, opened 2025-06-13
universe+tapdb: add new `block_height` field to `universe_leaves` use that to implement block height based supply tree queries

Roasbeef:
In this PR, we add a new method `FetchSupplyLeavesByHeight`, that can be used to query for the leaves of a supply tree by height. This will be useful for writing the new syncing state machine and the sub-system that serves the supply tree syncer. Such a syncer would start from the initially created pre-commitment, then fetch all the leaves for the future updates, assemble those into the supply tree, then verify that everything matches up. 

<details>

<summary> Prompts used during development (aider+Gemini 2.5-pro-06-5) </summary>

### Initial draft creation

Based on by working memory of this new feature, I scanned the relevant files, then made mental nodes of what needed to be updated. I then drafted this fairly detailed prompt that guided `aider` w.r.t which files to update, and the set of high level changes to make. 
 
```
#### Today we assemble the individual updates (burn, ignore, mint) for a given asset  
#### specifier (group key is what we care about) into distinct trees for each type.  
#### We then further aggregrate each of those roots into a unified supply tree for  
#### the entire asset specifier. This tree commits to a root sum of the supply (you  
#### can subtract out any of the roots from the root sum to derive a snapshot of the  
#### supply such as the amt burnt, etc).  
####   
#### We then also have a system to assemble these updates, then commit then in a  
#### batch, this exists in: tapdb/supply_commit.go and tapdb/supply_tree.go.  
####   
#### For each of the supply sub-trees, we actually re-use the existing  
#### universe_leaves table to point to each of the respective leaves.  
####   
#### In this task, we want to add a block height to this leaf, as this'll allow us  
#### to scan the leaves in order, based on the type, so we can send out the set of  
#### leaves added in a state transition to a party that wants to sync the set of  
#### leaves.  
####   
####   
#### For this, we may need to update the set of queries in:  
#### tapdb/sqlc/queries/universe.sql. Of interest is also the  
#### applySupplyUpdatesInternal helper function, as we'll want to make sure that the  
#### helper functions that it calls properly inserts the height into the ultimately  
#### universe leaf. For the mint and burn types, we can obtain this height from the  
#### issuance proof.  
####   
#### For the ignore type, we'll need to add a block height field to  
#### universe.IgnoreTuple (universe/ignore_records.go).  
####   
#### For the leaves, we'll likely need to add a new migration (38) which adds the  
#### new height field (can be NULL) to the universe_leaves table.  
####   
#### We'll likely also want to add a new BlockHeight() method to the  
#### SupplyUpdateEvent interface.  
####   
#### Next, we'll want a new method on the SupplyTreeStore that given a start+end  
#### height and a specifier, the set of updates whose universe_leaves have a block  
#### height between that param. This may need to update  
#### tapdb/sqlc/queries/supply_tree.sql.  
####   
#### Also see tapdb/querier.go.  
####   
#### Ruminate on my sketch above, then let's create an execution plan remembering  
#### the ultimate goal. From there we'll implement each portion.  
```

### Subsequent iterations 

From here I made an edit or two to get things compiling. Then I executed a series of prompts to refine the code, refactor slightly, then add unit tests. 

> #### inside multiverse.UpsertProof Leaf, let's use `SparseDecode` instead along with `BlockHeightRecord` as we don't need to decode the entire proof just to get the block height  

> #### let's use the same sparse decode in `BaseUniverseStore`.UpsertProofLeaf  

> #### let's do the same for `SupplyTreeStore`.RegisterMintSupply (sparse decode)  

> #### can we morph that into a function closure invocation that itself returns lfn.Option[uint32]? so we don't leak that intermediate variable to the higher scope  

> #### great! but now we have a bit of duplication, can we make a helper function can call that intead of having the same duplication everywhere? we can put the helper function in tapdb/sqlutils.go  

> #### nice! now let's extend the unit tests in tapdb/supply_tree_test.go to cover the new `FetchSupplyLeavesByHeight` method.  

I messed when committing and dropped a commit entirely (lol), so I had `aider` re-add it: 
> #### help me ree-add `QuerySupplyLeavesByHeight` back to tapdb/sqlc/queries/supply_tree.sql, I deleted it accidentally  

</details>

### DIFF
--- a/tapdb/ignore_tree.go
+++ b/tapdb/ignore_tree.go
@@ -113,6 +113,21 @@ func addTuplesInternal(ctx context.Context, db BaseUniverseStore,
 				"universe root: %w", err)
 		}
 
+		var blockHeight lfn.Option[uint32]
+		height := tup.IgnoreTuple.Val.BlockHeight
+		if height > 0 {
+			blockHeight = lfn.Some(height)
+		}
+
+		sqlBlockHeight := lfn.MapOptionZ(
+			blockHeight, func(num uint32) sql.NullInt32 {
+				return sql.NullInt32{
+					Int32: int32(num),
+					Valid: true,
+				}
+			},
+		)
+
 		scriptKey := ignoreTup.ScriptKey
 		err = db.UpsertUniverseLeaf(ctx, UpsertUniverseLeaf{
 			AssetGenesisID:    assetGenID,
@@ -121,6 +136,7 @@ func addTuplesInternal(ctx context.Context, db BaseUniverseStore,
 			LeafNodeKey:       smtKey[:],
 			LeafNodeNamespace: namespace,
 			MintingPoint:      ignorePointBytes,
+			BlockHeight:       sqlBlockHeight,
 		})
 		if err != nil {
 			return nil, fmt.Errorf("failed to upsert ignore "+
--- a/tapdb/sqlc/queries/supply_tree.sql
+++ b/tapdb/sqlc/queries/supply_tree.sql
@@ -44,6 +44,25 @@ JOIN universe_supply_roots r
 WHERE r.id = @supply_root_id AND
       (l.sub_tree_type = sqlc.narg('sub_tree_type') OR sqlc.narg('sub_tree_type') IS NULL);
 
+-- name: QuerySupplyLeavesByHeight :many
+SELECT
+    leaves.script_key_bytes,
+    gen.gen_asset_id,
+    nodes.value AS supply_leaf_bytes,
+    nodes.sum AS sum_amt,
+    gen.asset_id,
+    leaves.block_height
+FROM universe_leaves AS leaves
+JOIN mssmt_nodes AS nodes
+    ON leaves.leaf_node_key = nodes.key
+    AND leaves.leaf_node_namespace = nodes.namespace
+JOIN genesis_info_view AS gen
+    ON leaves.asset_genesis_id = gen.gen_asset_id
+WHERE
+    leaves.leaf_node_namespace = @namespace AND
+    (leaves.block_height >= sqlc.narg('start_height') OR sqlc.narg('start_height') IS NULL) AND
+    (leaves.block_height <= sqlc.narg('end_height') OR sqlc.narg('end_height') IS NULL);
+
 -- name: DeleteUniverseSupplyLeaves :exec
 DELETE FROM universe_supply_leaves
 WHERE supply_root_id = (
--- a/tapdb/sqlc/queries/universe.sql
+++ b/tapdb/sqlc/queries/universe.sql
@@ -38,17 +38,18 @@ WHERE namespace_root = @namespace_root;
 
 -- name: UpsertUniverseLeaf :exec
 INSERT INTO universe_leaves (
-    asset_genesis_id, script_key_bytes, universe_root_id, leaf_node_key, 
-    leaf_node_namespace, minting_point
+    asset_genesis_id, script_key_bytes, universe_root_id, leaf_node_key,
+    leaf_node_namespace, minting_point, block_height
 ) VALUES (
     @asset_genesis_id, @script_key_bytes, @universe_root_id, @leaf_node_key,
-    @leaf_node_namespace, @minting_point
+    @leaf_node_namespace, @minting_point, @block_height
 ) ON CONFLICT (minting_point, script_key_bytes, leaf_node_namespace)
     -- This is a NOP, minting_point and script_key_bytes are the unique fields
     -- that caused the conflict.
     DO UPDATE SET minting_point = EXCLUDED.minting_point,
                   script_key_bytes = EXCLUDED.script_key_bytes,
-                  leaf_node_namespace = EXCLUDED.leaf_node_namespace;
+                  leaf_node_namespace = EXCLUDED.leaf_node_namespace,
+                  block_height = EXCLUDED.block_height;
 
 -- name: DeleteUniverseLeaves :exec
 DELETE FROM universe_leaves
--- a/tapdb/supply_tree.go
+++ b/tapdb/supply_tree.go
@@ -1,14 +1,18 @@
 package tapdb
 
 import (
+	"bytes"
 	"context"
+	"database/sql"
 	"encoding/hex"
+	"errors"
 	"fmt"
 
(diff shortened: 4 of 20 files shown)


tapdb/supply_tree.go
```diff
@@ -596,3 +596,111 @@ func (s *SupplyTreeStore) ApplySupplyUpdates(ctx context.Context,
 	return finalRoot, nil
 }
 
+// SupplyUpdate is a struct that holds a supply update event and its block
+// height.
+type SupplyUpdate struct {
+	supplycommit.SupplyUpdateEvent
+	BlockHeight uint32
+}
+
```
ffranr: `supplycommit.SupplyUpdateEvent` already includes 
```
// BlockHeight returns the block height of the update.
BlockHeight() uint32
```

so I don't think we need the separate `BlockHeight uint32` field in `SupplyUpdate`.

Roasbeef:
Rebased.

litbot-9000:
@ffranr: review reminder

tapdb/ignore_tree.go
```diff
@@ -113,6 +113,21 @@ func addTuplesInternal(ctx context.Context, db BaseUniverseStore,
 				"universe root: %w", err)
 		}
 
+		var blockHeight lfn.Option[uint32]
+		height := tup.IgnoreTuple.Val.BlockHeight
+		if height > 0 {
+			blockHeight = lfn.Some(height)
+		}
+
+		sqlBlockHeight := lfn.MapOptionZ(
```
guggero, line 122: Looks like we already have a helper for this: `sqlOptInt32`.

tapdb/sqlc/queries/universe.sql
```diff
@@ -38,17 +38,18 @@ WHERE namespace_root = @namespace_root;
...
-    @leaf_node_namespace, @minting_point
+    @leaf_node_namespace, @minting_point, @block_height
 ) ON CONFLICT (minting_point, script_key_bytes, leaf_node_namespace)
     -- This is a NOP, minting_point and script_key_bytes are the unique fields
     -- that caused the conflict.
     DO UPDATE SET minting_point = EXCLUDED.minting_point,
                   script_key_bytes = EXCLUDED.script_key_bytes,
-                  leaf_node_namespace = EXCLUDED.leaf_node_namespace;
+                  leaf_node_namespace = EXCLUDED.leaf_node_namespace,
+                  block_height = EXCLUDED.block_height;
```
guggero, line 52: This invalidates the `This is a NOP,` comment above! This would allow updating the block height specifically. Is that the intention? If yes, then we should update the comment.

tapdb/universe.go
```diff
@@ -773,6 +792,23 @@ func universeUpsertProofLeaf(ctx context.Context, dbTx BaseUniverseStore,
 		return nil, err
 	}
 
+	// If the block height isn't specified, then we'll attempt to extract it
+	// from the proof itself.
+	if blockHeight.IsNone() {
+		if leafProof.BlockHeight > 0 {
```
guggero: nit: could combine in a single `&&` condition?

tapdb/universe.go
```diff
@@ -773,6 +792,23 @@ func universeUpsertProofLeaf(ctx context.Context, dbTx BaseUniverseStore,
...
+			blockHeight = lfn.Some(leafProof.BlockHeight)
+		}
+	}
+
+	sqlBlockHeight := lfn.MapOptionZ(
```
guggero, line 800: nit: use `sqlOptInt32`.

tapdb/supply_tree.go
```diff
@@ -599,3 +603,107 @@ func (s *SupplyTreeStore) ApplySupplyUpdates(ctx context.Context,
...
+	if err != nil {
+		return nil, fmt.Errorf("group key must be "+
+			"specified for supply tree: %w", err)
+	}
+
+	var updates []SupplyUpdate
+
+	readTx := NewBaseUniverseReadTx()
+	dbErr := s.db.ExecTx(ctx, &readTx, func(db BaseUniverseStore) error {
+		for _, treeType := range []supplycommit.SupplySubTree{
```
guggero: nit: feels like this would make a good package-level variable with the list of tree types we often iterate over? Would also help a bit with the formatting/readability.

tapdb/sqlc/queries/supply_tree.sql
```diff
@@ -43,6 +43,25 @@ JOIN universe_supply_roots r
...
+    leaves.script_key_bytes,
+    gen.gen_asset_id,
+    nodes.value AS supply_leaf_bytes,
+    nodes.sum AS sum_amt,
+    gen.asset_id,
+    leaves.block_height
+FROM universe_leaves AS leaves
+JOIN mssmt_nodes AS nodes
+    ON leaves.leaf_node_key = nodes.key
+    AND leaves.leaf_node_namespace = nodes.namespace
```
ffranr, line 57: We already have a `where` clause in `QueryUniverseSupplyLeaves`. Instead of adding a new query here, can't we just extend `QueryUniverseSupplyLeaves` with an optional block height?

Roasbeef: Good eye. These queries differ slightly though: the above is based on always knowing the `root_id` to pass in (hence the join with universe supply roots). Here we query based on just the namespace, so we can skip the extra join on the root.

tapdb/sqlc/queries/universe.sql
Roasbeef: Nice catch, will update the comment here.

tapdb/universe.go
Roasbeef: Done.

tapdb/universe.go
Roasbeef: Done.

tapdb/supply_tree.go
Roasbeef: Good idea, added.

tapdb/universe.go
Roasbeef: Actually I realize the diff here: one uses `lfn`, while the other uses `fn`.

Roasbeef:
This should be ready for a final pass. The unit test failure is being addressed in another PR to improve the set of property based tests.

Issue #1598, opened 2025-06-13
[bug]:  The fee in the asset channel was not successfully collected.


shuishihang:
I created three nodes (A, B, C). Node A and B established an asset channel, while node B and C have a BTC channel. I set the asset price at 1 asset = 100 sats. When I created a 10,000 sats invoice using node C, node A paid it with assets but only transferred 354 sats + 100 assets.

However, when I set the price to 1 asset = 1 sat, an extra 1 asset was charged as a fee. At the 1 asset = 100 sats rate with a routing fee of 10 sats, after dozens of transactions, no asset-based fee was deducted. Shouldn’t the system charge 1 asset as a fee every 10 payments (since 10 × 10 sats = 100 sats = 1 asset)? Is there a way to collect fees in sats instead of assets in the asset channel?

guggero:
Fees are always collected in milli-satoshis. But because assets aren't divisible, a round-down mechanism is implemented.
Which means, the fees can only be paid if the assets have a high enough "resolution", meaning that one asset unit can represent one or more milli-satoshis.
If that isn't the case, then things are simply rounded down. And there's no over time accounting, which means "a fee every 10 payments" isn't what is happening. A fee can either be expressed in assets and be paid or it won't.

See https://github.com/lightninglabs/taproot-assets/blob/main/docs/rfq-and-decimal-display.md for more explanation.

If you want to test that this works, I suggest changing the following parameters:
 - Issue an asset with the recommended decimal display of 6
 - Set the rate as 1 asset = 1 milli-satoshi

But bear in mind that the 354 sats that are required to send a non-dust HTLC will always be paid, which might be larger than any possible fee that's accrued. But that's a current protocol limitation that we're going to [remove in future versions](https://github.com/lightninglabs/taproot-assets/pull/1567).

shuishihang:
Thank you for your response. I have another question: Suppose there are three nodes (A-B-C), where A and B are connected via a BTC Lightning channel, and B and C are also connected via a BTC Lightning channel. If node B sets its base_fee_msat to 2000 using `lncli updatechanpolicy`, then when node C generates an invoice and node A makes a payment, node B will earn a 2-sat routing fee. However, if these two BTC channels are instead asset channels (e.g., for a different asset), node B’s routing fee is determined by node C’s fee policy. Why does this happen?

guggero:
> node B’s routing fee is determined by node C’s fee policy. Why does this happen?

[...]

PR #1601, opened 2025-06-16
bugfix: fix btc-only balance force close, fix incorrect policy in invoice hop hint

guggero:
Fixes root cause of https://github.com/lightninglabs/taproot-assets/issues/1595.
Also fixes https://github.com/lightninglabs/taproot-assets/issues/1598.

Corresponding integration tests can be found here: https://github.com/lightninglabs/lightning-terminal/pull/1089

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -8553,8 +8553,10 @@ func (r *rpcServer) getInboundPolicy(ctx context.Context, chanID uint64,
 		return nil, fmt.Errorf("unable to fetch channel: %w", err)
 	}
 
+	// We want the policy that corresponds to the node that is the remote
+	// peer in the channel.
 	policy := edge.Node2Policy
-	if edge.Node2Pub == remotePubStr {
+	if edge.Node1Pub == remotePubStr {
 		policy = edge.Node1Policy
 	}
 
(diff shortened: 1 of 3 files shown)


guggero:
> We could consider introducing distinct types for inbound and outbound policies to prevent misuse—e.g., accidentally passing an outbound policy where an inbound one is expected. That said, probably not something for this PR.

Yeah, that would be nice. But that would require changes in lnd and its RPC.

> Also, do we need any regression tests (changes to tests) to cover this fix?

Yes, I've added them in the linked litd PR: https://github.com/lightninglabs/lightning-terminal/pull/1089
Gonna check why they're currently failing.

ffranr:
> Yeah, that would be nice. But that would require changes in lnd and its RPC.

@guggero I was thinking of just wrapper types in tap.

guggero:
> @guggero I was thinking of just wrapper types in tap.

Ah, okay. But we currently only use the policies in a single place, so not sure if that's a bit overkill for now.

rpcserver.go
```diff
@@ -8553,8 +8553,10 @@ func (r *rpcServer) getInboundPolicy(ctx context.Context, chanID uint64,
 		return nil, fmt.Errorf("unable to fetch channel: %w", err)
 	}
 
+	// We want the policy that corresponds to the node that is the remote
+	// peer in the channel.
 	policy := edge.Node2Policy
-	if edge.Node2Pub == remotePubStr {
+	if edge.Node1Pub == remotePubStr {
```
Roasbeef, line 8559: Ah yeah, the direction of the policy here is relative

PR #1602, opened 2025-06-16
Improve robustness of `GenGroupAnchorVerifier` with anchor checks

ffranr:
This PR strengthens the `GenGroupAnchorVerifier` by ensuring that only valid group anchor assets are accepted. It introduces helper methods `IsGenesisAsset` and `IsGroupAnchor` to clearly identify group anchors, and adds verification logic to handle edge cases where `FetchGroupByGroupKey` may return non-anchor assets without error.

This change was motivated by extensive debugging whilst developing another PR that revealed non-anchor asset geneses had been inserted into the database, and `GenGroupAnchorVerifier` was incorrectly treating them as valid group anchors.

### DIFF
--- a/tapgarden/caretaker.go
+++ b/tapgarden/caretaker.go
@@ -1512,6 +1512,20 @@ func GenGroupAnchorVerifier(ctx context.Context,
 					ErrGroupKeyUnknown)
 			}
 
+			isGroupAnchor, err := storedGroup.IsGroupAnchor()
+			if err != nil {
+				return fmt.Errorf("%x: group anchor verifier: "+
+					"unable to check if genesis is "+
+					"group anchor: %w", assetGroupKey[:],
+					err)
+			}
+
+			if !isGroupAnchor {
+				return fmt.Errorf("%x: group anchor verifier: "+
+					"genesis is not a group anchor: %w",
+					assetGroupKey[:], err)
+			}
+
 			groupAnchor = newSingleValue(storedGroup.Genesis)
 
 			_, _ = groupAnchors.Put(assetGroupKey, groupAnchor)


ffranr:
I don't expect the CI to fail. But if it does, please feel free not to review.

asset/group_key.go
```diff
@@ -92,6 +92,60 @@ type GroupKey struct {
 	Witness wire.TxWitness
 }
 
+// IsGenesisAsset returns true if the provided asset ID matches the group's
+// genesis ID, indicating it was the first asset minted into the group.
```
guggero: I'm not sure we should use the term "genesis" here, as IMO it doesn't mean what we are trying to detect here.
The "genesis" of an asset is always its minting event. But that also goes for additional tranches of a group, each tranche has its own genesis.
That's why we went with the term "group anchor", as it signifies that it created the group (a better name in hindsight might have been "group starter", "group initiator" or whatever).
So IMO this should be called `IsGroupAnchor()`? Or even more verbose `IsFirstGroupTranche()`?

asset/asset.go
```diff
@@ -937,6 +937,12 @@ type AssetGroup struct {
 	*GroupKey
 }
 
+// IsGroupAnchor returns true if this genesis matches the group genesis ID,
+// meaning it corresponds to the first tranche that established the group.
+func (a *AssetGroup) IsGroupAnchor() (bool, error) {
```
guggero, line 942: Yeah, given this, I think my previous comment makes even more sense. IMO we can have the same `IsGroupAnchor()` method on both the asset group and the group key. One has a parameter and the other does not. But the purpose is the same.

asset/group_key.go
```diff
@@ -92,6 +92,60 @@ type GroupKey struct {
...
+			return zero, fmt.Errorf("cannot derive group "+
+				"public key: %w", err)
+		}
+
+		derivedGroupPubKey = *gk
+
+	default:
+		return zero, fmt.Errorf("unsupported group key version: %d",
+			g.Version)
+	}
```
guggero: Simplify this to:
```
	gkr, err := NewGroupKeyReveal(*g, assetID)
	if err != nil {
		return zero, err
	}

	gk, err := gkr.GroupPubKey(assetID)
	if err != nil {
		return zero, fmt.Errorf("cannot derive group "+
			"public key: %w", err)
	}

	return expectedGroupPubKey.IsEqual(gk), nil
```

asset/group_key_test.go
```diff
@@ -679,3 +679,198 @@ func TestGroupKeyDerivationInvalidAsset(t *testing.T) {
...
+	// Helper function to generate a random asset ID.
+	randAssetID := func() ID {
+		assetBytes := test.RandBytes(32)
+		var assetID ID
+		copy(assetID[:], assetBytes)
+		return assetID
+	}
+
+	// Generate a sample asset ID to use as the genesis asset ID.
+	firstAssetID := randAssetID()
```
guggero: Can use `RandID(t)`, defined in `mock.go`.

asset/group_key_test.go
```diff
@@ -679,3 +679,198 @@ func TestGroupKeyDerivationInvalidAsset(t *testing.T) {
...
+			"root non-matching",
+		groupKeyVersion:      GroupKeyV1,
+		customTapscriptRoot:  fn.Some(test.RandHash()),
+		genesisAssetID:       randAssetID(),
+		candidateAssetID:     randAssetID(),
+		expectedIsGenesisRes: false,
+	}}
+
+	for _, tc := range testCases {
+		tc := tc
```
guggero: nit: no longer required. Can just use `tc` directly in Go 1.24 and later.

asset/group_key.go
ffranr: I thought of this simplification also. And I suspected you might notice it.

I deliberately choose not to use `NewGroupKeyReveal` because that passes the group key into the constructor, so it's not clear, IMO, if the group pub key is derived anew. Which is what we want here: we're trying to reproduce `g.GroupPubKey`.

The reader needs to know how `NewGroupKeyReveal` works to be convinced that it recreates the group pub key and won't just use `g.GroupPubKey`. So by going one level down I mean to help convince the reader that `derivedGroupPubKey` isn't just a plain copy of `g.GroupPubKey`.

guggero: Okay, I get that. But isn't that something a comment can explain? I mainly noticed this was copy/paste because of the TODO that's also copied. So in the future we need to update two places to address the TODO.
Or we could also rename `NewGroupKeyReveal` to `DeriveGroupKeyReveal` to make it more clear?
I think we should avoid code duplication where possible, even when it requires things to be explained a bit more in comments.

asset/group_key.go
ffranr: I've gone with `IsGroupAnchor` :+1:

asset/group_key.go
ffranr: I've gone for the simpler version + comment.

What did you mean by this TODO?
```
gkr, err := NewGroupKeyRevealV1(
			// TODO(guggero): Make this configurable in the future.
```

guggero: Okay cool. The TODO means: Right now we use Pedersen by default for v1 keys. But we could allow the user to choose the OP_RETURN variant in the future, if they don't need hardware wallet compatibility (or have a wallet that can display and sign that properly).

PR #1603, opened 2025-06-16
tapchannel: add additional logging for commitment sort error cases

Roasbeef:


tapchannel/commitment.go
```diff
@@ -1672,12 +1672,34 @@ func InPlaceCustomCommitSort(tx *wire.MsgTx, cltvs []uint32,
...
+		return fmt.Errorf("output and allocation size mismatch: "+
+			"len(tx.TxOut)=%d, len(allocations)=%d",
+			len(tx.TxOut), len(allocations))
 	}
 
 	if len(tx.TxOut) != len(cltvs) {
-		return fmt.Errorf("output and cltv list size mismatch")
+		log.Errorf("Output and cltv list size mismatch: "+
+			"len(tx.TxOut)=%d, len(cltvs)=%d. "+
+			"allocations=%s, htlc_indexes=%v",
```
GeorgeTsagk: we're using `allocations=%s` here and `allocations=%v` above, not sure if both behave the same but at least let's have the same notation

guggero (approved):

PR #1606, opened 2025-06-17
rpcserver: enforce keysend minimum carrier HTLC amount

guggero:
Fixes an issue where attempting to pay a below-dust carrier HTLC satoshi amount would knock a channel offline.

Found after attempting the negative test mentioned here: https://github.com/lightninglabs/lightning-terminal/pull/1089#discussion_r2150457951
Nice catch, @GeorgeTsagk.

### DIFF
--- a/itest/channels_test.go
+++ b/itest/channels_test.go
@@ -3,6 +3,8 @@ package itest
 import (
 	"context"
 
+	"github.com/lightninglabs/taproot-assets/internal/test"
+	"github.com/lightninglabs/taproot-assets/rfqmath"
 	"github.com/lightninglabs/taproot-assets/rfqmsg"
 	tchrpc "github.com/lightninglabs/taproot-assets/taprpc/tapchannelrpc"
 	"github.com/lightningnetwork/lnd/lnrpc"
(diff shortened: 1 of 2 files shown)


rpcserver.go
```diff
@@ -7798,6 +7798,16 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
 				"pubkey: %w", err)
 		}
 
+		// We must enforce a minimum satoshi amount to make sure the
+		// carrier HTLC is above dust. This is handled by the traffic
+		// shaper for invoice-based payments, but for keysend payments
+		// we need to do it here.
+		if pReq.Amt < int64(rfqmath.DefaultOnChainHtlcSat) {
+			return fmt.Errorf("keysend payment satoshi amount "+
```
GeorgeTsagk, line 7806: I wonder if this is correct
```
// ...but for keysend payments
// we need to do it here.
```

We could do this directly in the [`AuxTrafficShaper`](https://github.com/lightninglabs/taproot-assets/blob/3243db8c9028d7affa693e4db6d30410850632b7/tapchannel/aux_traffic_shaper.go#L434), where instead of blindly returning the msat we first clamp it to the `rfqmath.DefaultOnChainHtlcSat` amount if smaller.

Given that we're soon going to support noop, this is fine from the perspective of "assuming a greater amount" as it will not end up settling on the remote peer's side, as seen in the noop PR [related lines](https://github.com/lightninglabs/taproot-assets/blob/723311b77566d8209d6cb0962888a1c6f0b0b7d1/tapchannel/aux_traffic_shaper.go#L431-L442).

But noop is not yet merged, maybe let's keep it as is (i.e error out if below dust), and we could revert this check and apply the behavior in `ProduceHtlcExtraData` when noop gets merged.

So I guess there's no correct/incorrect approach after all

guggero: Yeah, good point. We could just auto-bump it to the minimum amount. We set the fixed amount of 354 sats when you use the CLI command as well.
But, I think for a developer using the API, an explicit error is probably more easy to understand. The user on the CLI doesn't have the option to set the satoshi amount, so IMO it's okay for us to choose. But a dev will see the parameter in the API docs and will wonder what to set it to. And it automatically using a higher value than they set might be unexpected.

So I also think we should keep the PR as is and address this again once the noop PR is merged.

itest/channels_test.go
```diff
@@ -66,6 +67,25 @@ func testChannelRPCs(t *harnessTest) {
 			"payment",
 	)
 
+	// Make sure that the minimum satoshi amount for a keysend payment is
+	// enforced.
+	stream, err = t.tapd.SendPayment(ctx, &tchrpc.SendPaymentRequest{
+		AssetAmount: 123,
```
ffranr, line 74: instead of `123`, make this `rfqmath.DefaultOnChainHtlcSat - 1`?

ffranr: Would also be nice to have a `MinOnChainHtlcSat` constant for this threshold. Just to have a const by name for this minimum.

PR #1609, opened 2025-06-18
multi: allow supply commit state machine to accept pending updates during state transition

Roasbeef:
In this PR, we update the state machine to accept pending updates during a state transition. This takes a burden off of callers, as they can just continue to send in new events at will without needing to handle the error case. 

To achieve this, we add a stage where we explicitly freeze a pending state transition. Once frozen, any other updates will be added as a dangling update (not pointing to a state transition). We then update the state machine to do a new transition where we'll query for dangling updates and bind them to a new state transition in an atomic step. We'll then create a new state transition with these dangling updates. 

TODO 

- [ ] Update docs and state machine diagram. 
- [ ] Additional tests for bind in db, and loop back in state machine


tapdb/sqlc/queries/supply_commit.sql
```diff
@@ -55,9 +55,9 @@ WHERE commit_id = @commit_id;
 -- name: InsertSupplyCommitTransition :one
 INSERT INTO supply_commit_transitions (
     state_machine_group_key, old_commitment_id, new_commitment_id,
-    pending_commit_txn_id, finalized, creation_time
+    pending_commit_txn_id, finalized, frozen, creation_time
```
guggero, line 58: nit: IMO it's a bit easier to see that nothing is missing when all the changes to the SQL schema and queries (and updating the generated code) are in a single commit.

tapdb/sqlc/queries/supply_commit.sql
```diff
@@ -95,6 +95,7 @@ SELECT
     t.old_commitment_id,
     t.new_commitment_id,
     t.pending_commit_txn_id,
+    t.frozen,
```
guggero: We could use `sqlc.embed(supply_commit_transitions)`, then new fields would be added automatically.

tapdb/supply_commit.go
```diff
@@ -522,6 +543,75 @@ func (s *SupplyCommitMachine) InsertPendingUpdate(ctx context.Context,
...
+		// If a pending transition exists, check if it's frozen.
+		if err == nil {
+			// If the transition is frozen, we should insert a
+			// dangling update that will be processed after the
+			// current transition is finalized.
+			if pendingTransition.Frozen {
+				var b bytes.Buffer
+				err := serializeSupplyUpdateEvent(&b, event)
+				if err != nil {
+					return fmt.Errorf("failed to serialize event "+
```
guggero: nit: seems like your LLM instruction markdown needs a section about line length.

tapdb/supply_commit.go
```diff
@@ -522,6 +543,75 @@ func (s *SupplyCommitMachine) InsertPendingUpdate(ctx context.Context,
...
+						"data: %w", err)
+				}
+				updateTypeID, err := updateTypeToInt(
+					event.SupplySubTreeType(),
+				)
+				if err != nil {
+					return fmt.Errorf(
```
guggero: And our error formatting exception.

tapdb/supply_commit.go
```diff
@@ -522,6 +543,75 @@ func (s *SupplyCommitMachine) InsertPendingUpdate(ctx context.Context,
...
+		)
+
+		// If a pending transition exists, check if it's frozen.
+		if err == nil {
+			// If the transition is frozen, we should insert a
+			// dangling update that will be processed after the
+			// current transition is finalized.
+			if pendingTransition.Frozen {
+				var b bytes.Buffer
+				err := serializeSupplyUpdateEvent(&b, event)
```
guggero: Hmm, there doesn't really seem to be any difference between this branch of `if pendingTransition.Frozen {` and the code below other than the `TransitionID: sql.NullInt64{},` being set or not. Feels like this could be simplified quite a bit?

tapdb/supply_commit.go
```diff
@@ -1086,7 +1182,10 @@ func (s *SupplyCommitMachine) FetchState(ctx context.Context,
 		// Now that we know we have a state transition, we'll query for
 		// all the pending updates related to the state transition.
 		eventRows, err := db.QuerySupplyUpdateEvents(
-			ctx, dbTransition.TransitionID,
+			ctx, sql.NullInt64{
```
guggero: nit: use `sqlInt64`?

tapdb/supply_commit.go
Roasbeef: Hadn't fully run the linter yet. The commit hash pin we used doesn't work in git work trees, will address the linting issues before removing from draft.

Roasbeef:
Removed from draft.

tapdb/supply_commit.go
Roasbeef: Ah yeah you're right, I went through a few iterations here with some fancier SQL queries before going this route, but didn't consolidate afterwards. 

Simplified this section with a helper function.

guggero (approved):
Very nice, LGTM :tada:

PR #1610, opened 2025-06-18
universe: use correct ID comparison for sync config

guggero:
Fixes a bug where the error "proof insert is disabled for the given universe" is returned, even though group key specific configs have been set. This was because the == operator checks pointer equality instead of content equality for pointer struct members.


Roasbeef (approved):
LGTM 🚝

PR #1587, opened 2025-06-06
[group key addrs 3/?]: send and receive support for V2 addresses

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/1502 and https://github.com/lightninglabs/taproot-assets/pull/1512.

This is a very early prototype that barely works.
But it shows all the areas that need to be touched to support V2 addresses.

This will likely be split into two or three PRs, once fully finished and cleaned up.

PR #1611, opened 2025-06-19
[group key addrs 3/4]: misc refactor

guggero:
Extracts some refactor commits out from https://github.com/lightninglabs/taproot-assets/pull/1587, which are mostly independent of the functionality in that PR and can be reviewed and merged individually.

### DIFF
--- a/fn/memory.go
+++ b/fn/memory.go
@@ -1,6 +1,10 @@
 package fn
 
-import "errors"
+import (
+	"bytes"
+	"errors"
+	"io"
+)
 
 var (
 	// ErrNilPointerDeference is returned when a nil pointer is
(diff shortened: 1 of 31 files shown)


fn/memory.go
```diff
@@ -68,3 +72,46 @@ func DerefPanic[T any](ptr *T) T {
...
+	}
+
+	var buf bytes.Buffer
+	err := s.Serialize(&buf)
+	if err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
```
ffranr: Drop these in favour of the encode style above? I think that having both in a core package like `fn` just leads to unnecessary confusion. Fundamental packages like `fn` provide future dev guidance.

Or perhaps there is some fundamental meaning difference between the terms "encode" and "serialize" that I don't know about? If there is please add to doc comments so that I know what one to use.

ffranr: Maybe also add a new file called `encoding.go` to `fn` to accommodate these changes. From my pov, I don't think they're related to memory.

key_ring.go
```diff
@@ -88,6 +89,23 @@ func (l *LndRpcKeyRing) IsLocalKey(ctx context.Context,
...
+// key if no key locator is specified):
+//
+//	P_shared = privKeyNode * ephemeralPubkey
+//
+// The resulting shared public key is serialized in the compressed
+// format and hashed with SHA256, resulting in a final key length of 256
+// bits.
+func (l *LndRpcKeyRing) DeriveSharedKey(ctx context.Context,
+	ephemeralPubKey *btcec.PublicKey,
+	keyLocator *keychain.KeyLocator) ([32]byte, error) {
```
ffranr: nit: return `[sha256.Size]byte`?

fn/memory.go
guggero: I'm not sure what you mean by "in favor of the encoding style above"? These are helper functions to easily convert existing types to bytes. And some types (e.g. `wire.MsgTx`, `psbt.Packet`) define a `Serialize()` method while other types have an `Encode()` method. And we want to support both.

Moved the definitions to `encoding.go`.

PR #1612, opened 2025-06-19
tapdb: retroactively insert old asset burns into burn table

guggero:
We added the dedicated burn table after introducing the burn functionality. Which means, some asset burns users made before the table was added didn't show up when querying burns with `tapcli assets listburns` (only when listing all assets, the `is_burn: true` would show in the output).

This fixes the problem by inserting any asset burns into the dedicated table.
We also fix an issue with the type of the foreign key.

@Roasbeef does this also have an impact on the burn tree? Should we also insert the burns there?


Roasbeef:
> does this also have an impact on the burn tree? Should we also insert the burns there?

Not _yet_, as we haven't hooked up inserting elements into the tree yet. We also haven't implemented retroactively creation of supply updates into the tree.

tapdb/sqlc/migrations/000036_insert_asset_burns_migration.up.sql
```diff
@@ -0,0 +1,32 @@
+-- We need to modify the transfer ID to be a BIGINT instead of INTEGER,
+-- otherwise at some point things will break unexpectedly.
+
+CREATE TABLE IF NOT EXISTS asset_burn_transfers_corrected (
+    -- The auto-incrementing integer that identifies this burn transfer.
+    burn_id INTEGER PRIMARY KEY,
+
+    -- A reference to the primary key of the transfer that includes this burn.
+    transfer_id BIGINT NOT NULL REFERENCES asset_transfers(id),
```
Roasbeef: So we only need to make this `BIGINT` for foreign key refs? Do we need to do this anywhere else? '

I think in some of my recent PRs, I don't declare foreign key refs as `BIGINT`.

tapdb/post_migration_checks.go
```diff
@@ -216,3 +223,95 @@ func determineAndAssignScriptKeyType(ctx context.Context,
...
+	burnAssets := fn.Filter(chainAssets, func(a *asset.ChainAsset) bool {
+		return a.IsBurn()
+	})
+
+	burnsInTable, err := q.QueryBurns(ctx, sqlc.QueryBurnsParams{})
+	if err != nil {
+		return err
+	}
+
+	burnsMatch := func(b sqlc.QueryBurnsRow, a *asset.ChainAsset) bool {
```
Roasbeef, line 266: Very clean 👍

tapdb/sqlc/migrations/000036_insert_asset_burns_migration.up.sql
Roasbeef: Perhaps we should add a new linting script for this?

Roasbeef (approved):
LGTM 🩱

tapdb/sqlc/migrations/000036_insert_asset_burns_migration.up.sql
guggero: Yes, if the primary key that is referenced is `INTEGER PRIMARY KEY` then it will become a `BIGSERIAL PRIMARY KEY` on postgres and the foreign key should be BIGINT to match that (you'll mainly notice because you have to cast the ID from int32 to int64 and back in the `tapdb` Golang code.
I did a quick scan and the only other instance I found was `mint_anchor_uni_commitments.batch_id`. Perhaps this can be updated in one of the in-flight PRs?

I also started looking into [`sqlfluff`](https://docs.sqlfluff.com/en/stable/index.html) that can do linting and also custom plugins, so we could add our own rule. Will create a PR soon.

Issue #1358, opened 2025-02-06
[feature]: Multi-RFQ send


GeorgeTsagk:
## Description

The daemon currently grabs the first available asset channel when trying to send a payment, and will acquire a single RFQ for that asset channel alone. If that fails for any reason, the operation fails, without trying any other candidate channels or quotes.

We want to change that, so that when sending a payment multiple asset channels are considered, and multiple RFQs are acquired. The payment lifecycle from the PoV of tapd should also support a dynamic selection of asset channels and RFQs, which is adaptive to failures.

A brief summary of what needs to be done:
- [x] Acquire quotes for all suitable asset channels, and consider all of them for sending the payment
- [x] Payment lifecycle should take into account the multiple asset channels and quotes, and be able to adapt the next attempt when previous one fails

Investigation:

- [x] Can we somehow re-use LND's sharding algorithm in order to not re-write something similar when fragmenting the asset amount over multiple channels/quotes?

Issue #1565, opened 2025-05-27
[bug]: `AddInvoiceResponse` only returns a single `accepted_buy_quote`


ZZiigguurraatt:
`AddInvoiceResponse` only returns a single `accepted_buy_quote`. Now that we have https://github.com/lightninglabs/taproot-assets/pull/1457 , this does not make sense. I'm not sure if this is the worst `accepted_buy_quote`, the best, or the first one. We should have all `accepted_buy_quote` returned, and there should be an indication which one is used to set the sats amount in the invoice that is generated.

https://github.com/lightninglabs/taproot-assets/blob/e323b9e72ce6b7c3421a0aaa78d697c310dc4f63/taprpc/tapchannelrpc/tapchannel.proto#L214-L218

I can decode the returned invoice to get a list of SCID and peer pubkeys used for each quote, but that does not tell me the quote amount, only which peers I also got quotes with. Also, I don't think a user should need to decode the returned invoice just to get the SCID and peer pubkeys.

ZZiigguurraatt:
We've introduced the same problem for `SendPayment` in https://github.com/lightninglabs/taproot-assets/pull/1613 with `accepted_sell_order`:


https://github.com/lightninglabs/taproot-assets/blob/aa55b42538dd9c64718259a64e45ed24cd9bc545/taprpc/tapchannelrpc/tapchannel.proto#L160-L165

ZZiigguurraatt:
> We've introduced the same problem for `SendPayment` in [#1613](https://github.com/lightninglabs/taproot-assets/pull/1613) with `accepted_sell_order`:
> 
> [taproot-assets/taprpc/tapchannelrpc/tapchannel.proto](https://github.com/lightninglabs/taproot-assets/blob/aa55b42538dd9c64718259a64e45ed24cd9bc545/taprpc/tapchannelrpc/tapchannel.proto#L160-L165)
> 
> Lines 160 to 165 in [aa55b42](/lightninglabs/taproot-assets/commit/aa55b42538dd9c64718259a64e45ed24cd9bc545)
>  // In case channel assets need to be swapped to another asset, an asset 
>  // sell order is negotiated with the channel peer. The result will be 
>  // the first message in the response stream. If no swap is needed, the 
>  // payment results will be streamed directly. 
>  rfqrpc.PeerAcceptedSellQuote accepted_sell_order = 1;

[...]

Issue #1442, opened 2025-03-14
[feature]: `AddInvoice`: allow specifying an `rfq_id`


ZZiigguurraatt:
For [SendPayment](https://lightning.engineering/api-docs/api/taproot-assets/taproot-asset-channels/send-payment/) we have the ability to manually pass an `rfq_id` from [AddAssetSellOrder](https://lightning.engineering/api-docs/api/taproot-assets/rfq/add-asset-sell-order/).

In [AddInvoice](https://lightning.engineering/api-docs/api/taproot-assets/taproot-asset-channels/add-invoice/)  we do not have the ability to manually pass an `rfq_id` from [AddAssetBuyOrder](https://lightning.engineering/api-docs/api/taproot-assets/rfq/add-asset-buy-order/) . This issue requests that capability to be added.

Motivation for the issue can be found in https://github.com/lightninglabs/taproot-assets/issues/1440 .

Here we have a test case that does manually create the asset invoice and uses a manually passed `rfq_id`: https://github.com/lightninglabs/lightning-terminal/blob/f082579252e4b6d1f9dd84e74646ec3e9b7e8e33/itest/litd_custom_channels_test.go#L2112-L2166 . That test needs to be updated once the capability is added in to the AddInvoice function (https://github.com/lightninglabs/taproot-assets/blob/331ac78538a3db5ca40826f82c5d3f6944744073/rpcserver.go#L7623-L7845).

Also, when fixing this issue, need to keep in mind https://github.com/lightninglabs/taproot-assets/issues/1428

PR #1613, opened 2025-06-23
Add Multi-RFQ Send

GeorgeTsagk:
## Description

Allows payments to use multiple quotes across multiple peers in order to pay an invoice. Previously we had to define a specific peer to negotiate a quote with, with this PR this is no longer required (but still supported) as Tapd will automatically scan our peers and establish quotes with all valid ones for the asset/amount of this payment.

The signature of `ProduceHtlcExtraData` had to be changed, as it's not possible to distinguish which of the quotes in the `rfqmsg.Htlc` should be used. We now provide the pubkey of the peer this HTLC is being sent to, in order to help Tapd extract the corresponding quote and calculate the correct amount of asset units.

Closes https://github.com/lightninglabs/taproot-assets/issues/1358

Depends on: https://github.com/lightningnetwork/lnd/pull/9980

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -47,6 +47,13 @@
   migration](https://github.com/lightninglabs/taproot-assets/pull/1612) was
   added that retroactively inserts all burned assets into that table.
 
+- Sending a payment now supports multi-rfq. This new feature allows for multiple
+  quotes to be used in order to carry out a payment. With multiple quotes, we
+  can use liquidity that is spread across different channels and also use
+  multiple rates. See
+  [related PR](https://github.com/lightninglabs/taproot-assets/pull/1613) for
+  more info.
+
 ## RPC Additions
 
 - The [price oracle RPC calls now have an intent, optional peer ID and metadata


rfqmsg/records.go
```diff
@@ -42,13 +42,119 @@ type (
...
+
+	return tlv.MakeDynamicRecord(
+		tlvType.TypeVal(), r, size, encodeAvailableIDs, decodeRfqIDs,
+	)
+}
+
+// encodeAvailableIDs encodes the RfqIDs struct on the wire.
+func encodeAvailableIDs(w io.Writer, val any, buf *[8]byte) error {
+	if ids, ok := val.(*RfqIDs); ok {
+		encodeIDs(w, &ids.IDs, buf)
```
guggero: Need to return the error here.

rfqmsg/records.go
```diff
@@ -42,13 +42,119 @@ type (
...
+		)
+		err := encodeAvailableIDs(&buf, r, &scratch)
+		if err != nil {
+			panic(fmt.Sprintf("unable to encode RFQ IDs: %v", err))
+		}
+		return uint64(buf.Len())
+	}
+
+	return tlv.MakeDynamicRecord(
+		tlvType.TypeVal(), r, size, encodeAvailableIDs, decodeRfqIDs,
```
guggero: Can remove the `tlvType.TypeVal()` here and just set 0, with the comment that we have in other places too:
```
	// Note that we set the type here as zero, as when used with a
	// tlv.RecordT, the type param will be used as the type.
```

rfqmsg/records.go
```diff
@@ -42,13 +42,119 @@ type (
...
+	ids, ok := val.(*[]ID)
+	if !ok {
+		return tlv.NewTypeForEncodingErr(val, "[]ID")
+	}
+	num := uint64(len(*ids))
+	if err := tlv.WriteVarInt(w, num, buf); err != nil {
+		return err
+	}
+	for _, id := range *ids {
+		idCopy := id // avoid pointer aliasing
```
guggero: nit: we don't really do end-of-line comments

rfqmsg/records.go
```diff
@@ -42,13 +42,119 @@ type (
...
+		}
+		ids.IDs = rfqIDs
+		return nil
+	}
+	return tlv.NewTypeForDecodingErr(val, "*RfqIDs", 0, 0)
+}
+
+// encodeIDs encodes the concatenated RFQ IDs into the given writer.
+func encodeIDs(w io.Writer, val any, buf *[8]byte) error {
+	ids, ok := val.(*[]ID)
```
guggero: Since we only use the `encodeIDs` and `decodeIDs` in this context, we can use a typed argument for `val` and skip this additional type check. Or we can directly inline both functions in the respective `...RfqIDs` functions.

rfqmsg/records.go
```diff
@@ -42,13 +42,119 @@ type (
...
+func decodeIDs(r io.Reader, val any, buf *[8]byte, _ uint64) error {
+	ids, ok := val.(*[]ID)
+	if !ok {
+		return tlv.NewTypeForDecodingErr(val, "[]ID", 0, 0)
+	}
+	num, err := tlv.ReadVarInt(r, buf)
+	if err != nil {
+		return err
+	}
+	list := make([]ID, num)
```
guggero: We allocate memory from a number we received over the wire. Need to check and error out the length before to make sure we limit the number of bytes we allocate.

rfqmsg/records.go
```diff
@@ -42,13 +42,119 @@ type (
...
+		if err := IdDecoder(r, &id, buf, 32); err != nil {
+			return err
+		}
+		list[i] = id
+	}
+	*ids = list
+	return nil
+}
+
+// SomeRfqIDsRecord creates an optional record that represens the array of
```
guggero: nit: typo.

rfqmsg/records.go
```diff
@@ -42,13 +42,119 @@ type (
...
+// available RFQ IDs.
+func SomeRfqIDsRecord(ids []ID) tlv.OptionalRecordT[AvailableRfqIDsType, RfqIDs] {
```
guggero: nit: line length.

tapchannel/aux_traffic_shaper.go
```diff
@@ -277,50 +277,72 @@ func paymentBandwidthAssetUnits(htlcAssetAmount, computedLocal uint64,
...
 	fundingChan *cmsg.OpenChannel) (lnwire.MilliSatoshi, error) {
 
-	// If the HTLC doesn't have an RFQ ID, it's incomplete, and we cannot
-	// determine the bandwidth.
-	if htlc.RfqID.ValOpt().IsNone() {
-		log.Tracef("No RFQ ID in HTLC, cannot determine matching " +
-			"outgoing channel")
+	// If the HTLC doesn't have any available RFQ IDs, it's incomplete, and
+	// we cannot determine the bandwidth.
+	if htlc.AvailableRfqIDs.IsNone() {
```
guggero, line 293: I think we still need to check if the old RFQ ID is set, to avoid backward compatibility issues.

tapchannel/aux_traffic_shaper.go
```diff
@@ -277,50 +277,72 @@ func paymentBandwidthAssetUnits(htlcAssetAmount, computedLocal uint64,
...
 		return 0, nil
 	}
 
-	// For every other use case (i.e. a normal payment with a negotiated
-	// quote or a multi-hop keysend that also uses a quote), we need to look
-	// up the accepted quote and determine the outgoing bandwidth in
-	// satoshis based on the local asset balance.
-	rfqID := htlc.RfqID.ValOpt().UnsafeFromSome()
+	// Retrieve the available RFQ IDs.
+	avaliableIDs := htlc.AvailableRfqIDs.UnsafeFromSome().Val.IDs
```
guggero: nit: typo (GoLand would show you those ;-) )

tapchannel/aux_traffic_shaper.go
```diff
@@ -277,50 +277,72 @@ func paymentBandwidthAssetUnits(htlcAssetAmount, computedLocal uint64,
 }
 
 // paymentBandwidth returns the available payment bandwidth of the channel based
-// on the asset rate of the RFQ quote that is included in the HTLC and the asset
-// units of the local balance.
+// on the list of availalbe RFQ IDs. If any of those IDs matches the channel, we
```
guggero: nit: typo

tapchannel/aux_traffic_shaper.go
```diff
@@ -277,50 +277,72 @@ func paymentBandwidthAssetUnits(htlcAssetAmount, computedLocal uint64,
...
+			minHtlcAmt, fundingChan,
+		)
+		if err != nil {
+			return 0, err
+		}
 
-	default:
-		return 0, fmt.Errorf("no accepted quote found for RFQ ID "+
-			"%x (SCID %d)", rfqID[:], rfqID.Scid())
+		if bandwidth > 0 {
```
guggero, line 351: So we explicitly just return the first bandwidth we find? To kind of almost randomize which channel we use first? Might be worth a comment here or in the method Godoc about _why_ we're taking the first vs. the largest for example.

server.go
```diff
@@ -1044,7 +1045,7 @@ func (s *Server) PaymentBandwidth(fundingBlob, htlcBlob,
 //
 // NOTE: This method is part of the routing.TlvTrafficShaper interface.
 func (s *Server) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
-	htlcCustomRecords lnwire.CustomRecords) (lnwire.MilliSatoshi,
+	htlcCustomRecords lnwire.CustomRecords, peer *route.Vertex) (lnwire.MilliSatoshi,
```
guggero: nit: line length

tapchannel/aux_traffic_shaper.go
```diff
@@ -456,16 +457,43 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
...
+	availableIDs := htlc.AvailableRfqIDs.UnsafeFromSome().Val.IDs
+
+	var (
+		rfqID rfqmsg.ID
+		quote rfqmsg.SellAccept
+	)
+
+	// Let's find the quote that matches this peer. This will be the quote
+	// that we'll use to calculate the asset units. Given that we may only
+	// establish a maximum of 1 quote per peer, this check is safe to
```
guggero: nit: to clarify my previous misunderstanding, change this to "establish a maximum of 1 quote per peer per payment, this check..."?

tapchannel/aux_traffic_shaper.go
```diff
@@ -456,16 +457,43 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
...
+	// perform as there are no competing quotes for a certain peer.
+	for _, id := range availableIDs {
+		q, ok := acceptedQuotes[id.Scid()]
+		if !ok {
+			continue
+		}
+
+		// We found the quote to use, now let's set the related fields.
+		if q.Peer.String() == peer.String() {
+			rfqID = id
```
guggero: nit: move this to after the comment or add newline before comment.

tapchannel/aux_traffic_shaper.go
```diff
@@ -456,16 +457,43 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
 		return totalAmount, htlcCustomRecords, nil
 	}
 
-	if htlc.RfqID.ValOpt().IsNone() {
-		return 0, nil, fmt.Errorf("no RFQ ID present in HTLC blob")
+	if htlc.AvailableRfqIDs.IsNone() {
```
guggero, line 483: Same here re backward compatibility. And also just to make sure we don't miss any edge case. After all, we explicitly set the singular `htlc.RfqID` in this function, so are we a 100% sure there is no code path where either the `ProduceHtlcExtraData` or `PaymentBandwidth` would be called with  `htlc.RfqID` set?
IMO we should check both and only error out if none of them (or both) are set at the same time, with a comment that we'd only really expect the list to be set in the future.

rpcserver.go
```diff
@@ -7703,77 +7685,76 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
-			return fmt.Errorf("error adding sell order: %w", err)
-		}
+		// Since we don't have any pre-negotiated rfq IDs available, we
+		// need to initiate the negotiation procedure for this payment.
 
-		var acceptedQuote *rfqrpc.PeerAcceptedSellQuote
-		switch r := resp.Response.(type) {
-		case *rfqrpc.AddAssetSellOrderResponse_AcceptedQuote:
-			acceptedQuote = r.AcceptedQuote
+		// We'll store here all the quotes we acquired successfully.
```
guggero, line 8188: nit: remove empty line between comment blocks? Sounds like they belong together.

rpcserver.go
```diff
@@ -7703,77 +7685,76 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+				return err
+			}
 
-		default:
-			return fmt.Errorf("unexpected response type: %T", r)
-		}
+			// Send out the information about the quote on the
+			// stream.
+			//
+			// nolint:ll
```
guggero: nit: we don't have the custom `ll` linter here (that's lnd only), we still use the default `lll`.

rpcserver.go
```diff
@@ -7703,77 +7685,76 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+					"a sell order for payment: %v", err)
+				continue
+			}
 
-		case *rfqrpc.AddAssetSellOrderResponse_RejectedQuote:
-			return fmt.Errorf("peer %v rejected the quote, code: "+
-				"%v, error message: %v", r.RejectedQuote.Peer,
-				r.RejectedQuote.ErrorCode,
-				r.RejectedQuote.ErrorMessage)
+			err = checkOverpayment(
```
guggero: So if a single peer gives us such a bad quote that we error out on this, the whole payment attempt is aborted. Not sure if we should instead log a warning here as well and continue? And perhaps store the latest error we logged and return that on `len(acquiredQuotes) == 0` so the user knows why no quotes could be found without needing to look at the log? IMO that would be useful for both the `r.acquireSellOrder` and `checkOverpayment` errors.

rpcserver.go
```diff
@@ -8463,6 +8444,52 @@ func (r *rpcServer) acquireBuyOrder(ctx context.Context,
...
+			),
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error adding sell order: %w", err)
+	}
+
+	switch r := resp.Response.(type) {
+	case *rfqrpc.AddAssetSellOrderResponse_AcceptedQuote:
+		quote = r.AcceptedQuote
```
guggero: nit: can return quote here directly to shorten things a bit (no var or final return statement needed then).

guggero (commented):
Looks pretty good! Have a couple of questions and suggestions, nothing major though.

rfqmsg/records.go
GeorgeTsagk: Good point, so far the total length of available RFQ IDs is open ended.

We should introduce a static limit (which would also limit the max number of quotes we acquire) and enforce it everywhere.

tapchannel/aux_traffic_shaper.go
GeorgeTsagk: this is a local Tapd<>LND interaction, would have to support old behavior if those were upgraded separately, but AFAIK that is not the case?

tapchannel/aux_traffic_shaper.go
GeorgeTsagk: Realized there's a flaw in this logic.

To answer your Q
> So we explicitly just return the first bandwidth we find? To kind of almost randomize which channel we use first?

Yes this is explicit. Given we establish 1 quote per peer, only one of the RFQ IDs in the array should produce some bandwidth. BUT, this is currently not an exhaustive check, we only check if the channel and the quote's specifier match... we also need to check that the channel peer & quote peer match. Otherwise the 1-quote-per-peer assumption is useless.

The reason this didn't trigger in the itests is probably because we're using the same assetID/groupID across quotes so the bandwidth report was roughly equal to reality (plus we use fixed oracle rates).

So with the current codebase there can be a mismatch between the reported bandwidth and the actually producable bandwidth.

The `ProduceHtlcExtraData` endpoint accepts the peer argument, which correctly guides it into picking the right quote. That's why the payments succeed. We just need to make sure the `PaymentBandwidth` calculates the bandwidth on the exact same selected quote.

tapchannel/aux_traffic_shaper.go
GeorgeTsagk: For the `PaymentBandwidth` and `ProduceHtlcExtraData` the `htlc.RfqID` is replaced by `htlc.AvailableRfqIDs`

The only use of `htlc.RfqID` is when `ProduceHtlcExtraData` locks into a specific RFQ, which will be used to produce the actual HTLC that will be sent out to our peer

From that point forward everything is the same (we only use the `htlc.RfqID` to route HTLCs / accept HTLCs to invoices)

rpcserver.go
GeorgeTsagk: Changed to instead `continue` in the loop and store the error in a var

The last errors (if any) of `r.acquireSellOrder` and `checkOverpayment` are joined and returned if no quotes were acquired

rfqmsg/records.go
```diff
@@ -42,13 +49,107 @@ type (
 	// encode an RFQ id within the custom records of an HTLC record on the
 	// wire.
 	HtlcRfqIDType = tlv.TlvType65538
+
+	// AvailableRfqIDsType is the type alias for the TLV type that is used
+	// to encode the list of available RFQ IDs that can be used for an HTLC.
+	// This list is only meant to be handled by the complementary LND
+	// instance via the AuxTrafficShaper hooks.
+	AvailableRfqIDsType = tlv.TlvType65540
```
ffranr, line 57: I think we should drop the term "available" from these changes. From my point of view it's implied.

We could name this TLV type: `HtlcRfqIDsType`. As I understand it, it's basically `HtlcRfqIDType` but it can hold multiple IDs.

rfqmsg/records.go
```diff
@@ -42,13 +49,107 @@ type (
...
 
 // SomeRfqIDRecord creates an optional record that represents an RFQ ID.
 func SomeRfqIDRecord(id ID) tlv.OptionalRecordT[HtlcRfqIDType, ID] {
 	return tlv.SomeRecordT(tlv.NewPrimitiveRecord[HtlcRfqIDType, ID](id))
 }
 
+// RfqIDs is a helper wrapper around the array of IDs that can be encoded as
+// records.
+type RfqIDs struct {
+	// IDs is a list of RFQ IDs that are associated with the HTLC.
```
ffranr: I think we should call this struct `HtlcRfqIDs` because the TLV type used will be specific to HTLC.

rfqmsg/records.go
```diff
@@ -59,21 +160,33 @@ type Htlc struct {
...
+	// This is used by the traffic shaper when querying for available
+	// bandwidth.
+	AvailableRfqIDs tlv.OptionalRecordT[AvailableRfqIDsType, RfqIDs]
 }
 
 // NewHtlc creates a new Htlc record with the given funded assets.
-func NewHtlc(amounts []*AssetBalance, rfqID fn.Option[ID]) *Htlc {
+func NewHtlc(amounts []*AssetBalance, rfqID fn.Option[ID],
+	availableRfqIDs fn.Option[[]ID]) *Htlc {
+
```
ffranr, line 177: I think we should change this function signature to 
```
func NewHtlc(amounts []*AssetBalance, rfqIDs []ID) *Htlc {
```

And if one ID is given set `htlc.RfqID` otherwise, if multiple, set the multi ID field on HTLC.

If `rfqIDs` is empty we should return an error here? Or to put that Q in the context of the current PR state, if `rfqID` and `availableRfqIDs` are both `None`, is that an error? I would expect RFQ ID is mandatory for `NewHtlc` in `rfqmsg` package.

ZZiigguurraatt:
Please update https://github.com/lightninglabs/taproot-assets/blob/df8ce528d7ff0d9554ee54ad114b788af7a8d51b/taprpc/tapchannelrpc/tapchannel.proto#L125-L130 to reflect this new capability. Also, I think we need to be more specific to include `group_key` in addition to `asset_id` in there.

ZZiigguurraatt:
What is this expected to do with an invoice with no amount specified?

ZZiigguurraatt:
`rfq_id` is of type bytes

https://github.com/lightninglabs/taproot-assets/blob/df8ce528d7ff0d9554ee54ad114b788af7a8d51b/taprpc/tapchannelrpc/tapchannel.proto#L139-L144

but with multi-rfq send, I would expect the ability to provide an array. Not sure that you want to fix that in this PR, but maybe it should be a separate issue? I did not make this comment for multi-rfq receive because of https://github.com/lightninglabs/taproot-assets/issues/1442 .

rfqmsg/records.go
GeorgeTsagk: Well yes, but I find 2 issues with this change (it is actually how I named this originally):
a) It's very easy to misread one for the other, as there's only 1 char diff between `HtlcRfqIDsType` and `HtlcRfqIDType`
b) The name IMO **should** be different to reflect the different semantics of the field. One is the actual RFQ we're using to send out an HTLC to our peer, the other is nothing but a hint towards LND as to what RFQs are available in order to query bandwidth & create first hop records

rfqmsg/records.go
GeorgeTsagk: I think we should keep the signature as is, in order to make sure each field is clear and expressiveness is preserved:

- If user sets `rfqID` then this RFQ will stick to the HTLC until it reaches our peer
- If user sets `rfqIDs` then those RFQs will be used later to lock in a specific `rfqID`

tapchannel/aux_traffic_shaper.go
guggero: So you're 100% certain for all the code paths that lead into `paymentBandwidth` there isn't a single one where potentially we only have the old RFQ ID set? What about this for example: https://github.com/lightninglabs/taproot-assets/blob/2e7ae0bc4f98d4f8d0d310d0755e5e36535cbe6f/rpcserver.go#L7674

tapchannel/aux_traffic_shaper.go
guggero: I think "Given we establish 1 quote per peer, only one of the RFQ IDs in the array should produce some bandwidth." would probably be a good comment to add above this statement.

tapchannel/aux_traffic_shaper.go
```diff
@@ -277,11 +280,12 @@ func paymentBandwidthAssetUnits(htlcAssetAmount, computedLocal uint64,
...
 
 // paymentBandwidth returns the available payment bandwidth of the channel based
-// on the list of availalbe RFQ IDs. If any of those IDs matches the channel, we
+// on the list of available RFQ IDs. If any of those IDs matches the channel, we
 // calculate the bandwidth based on its asset rate.
 func (s *AuxTrafficShaper) paymentBandwidth(htlc *rfqmsg.Htlc,
 	localBalance uint64, linkBandwidth, minHtlcAmt lnwire.MilliSatoshi,
-	fundingChan *cmsg.OpenChannel) (lnwire.MilliSatoshi, error) {
+	fundingChan *cmsg.OpenChannel,
+	peer *route.Vertex) (lnwire.MilliSatoshi, error) {
```
guggero: The `peer` should NOT be a pointer here and below, then we can use `==` to compare (which IIRC was the whole point for adding the `route.Vertex` array type).

tapchannel/aux_traffic_shaper.go
```diff
@@ -320,6 +327,12 @@ func (s *AuxTrafficShaper) paymentBandwidth(htlc *rfqmsg.Htlc,
 				"ID %x (SCID %d)", rfqID[:], rfqID.Scid())
 		}
 
+		// If the channel peer does not match the quote peer, continue
+		// to the next available quote.
+		if !bytes.Equal(quotePeer[:], peer[:]) {
```
guggero: See my comment above. Can just use `quotePeer route.Vertex`, then use the `==` comparison.

tapchannel/aux_traffic_shaper.go
```diff
@@ -456,16 +469,43 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
...
+	// establish a maximum of 1 quote per peer per payment, this check is
+	// safe to perform as there are no competing quotes for a certain peer.
+	for _, id := range availableIDs {
+		q, ok := acceptedQuotes[id.Scid()]
+		if !ok {
+			continue
+		}
+
+		// We found the quote to use, now let's set the related fields.
+		if q.Peer.String() == peer.String() {
```
guggero: No need to use `.String()` here if both are non-pointers.

tapchannel/aux_traffic_shaper.go
guggero: Same here re using an existing quote in the `SendPayment` RPC... Don't we have that covered by an itest?

guggero (commented):
Did another pass, getting very close here too.

tapchannel/aux_traffic_shaper.go
GeorgeTsagk: The above line indeed has to change into:
```golang
htlc := rfqmsg.NewHtlc(nil, fn.None[rfqmsg.ID](), fn.Some([]rfqmsg.ID{quote.ID}))
```

tapchannel/aux_traffic_shaper.go
GeorgeTsagk: Yeap, we have it covered, and the itest was failing! so +1 for our coverage, but I realize I never noticed the itest failing

So the reason it failed is because we were setting the existing RPC user provided RFQ as the single RfqID, instead it had to change to instead set it as the only available RFQ ID in the new field

rpcserver.go
```diff
@@ -8490,6 +8491,48 @@ func (r *rpcServer) acquireBuyOrder(ctx context.Context,
 	return quote, nil
 }
 
+// acquireSellOrder performs an RFQ negotiation with the target peer and quote
+// parameters and returns the quote if the negotiation was successful.
+func (r *rpcServer) acquireSellOrder(ctx context.Context,
```
ffranr: Here, I think we're placing a sell order to acquire a sell quote. So I think this method should be renamed to `acquireSellQuote`.

rpcserver.go
```diff
@@ -7734,76 +7721,88 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
-				"quote, status: %v", r.InvalidQuote.Peer,
-				r.InvalidQuote.Status.String())
+		// For each peer that satisfies our query above, we'll try and
+		// establish an RFQ quote for the asset specifier and max amount
+		// of this payment.
+		for peer := range chanMap {
+			quote, err := r.acquireSellOrder(
+				ctx, &rpcSpecifier, paymentMaxAmt, expiry,
+				&peer,
+			)
```
ffranr: I think the code would be better if this for-loop happened in a separate function.

I think `acquireSellOrder` is blocking, am I right? If so then the first peer could block until `expiry` (which i think is the invoice expiry) and then we don't have a chance to query any other peer.

We might need to call `r.acquireSellOrder` for each peer synchronously here. And the quote response timeout should be much less than the invoice expiry.

rpcserver.go
```diff
@@ -7734,76 +7721,88 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
-				r.RejectedQuote.ErrorCode,
-				r.RejectedQuote.ErrorMessage)
+			err = checkOverpayment(
+				quote, paymentMaxAmt, req.AllowOverpay,
+			)
+			if err != nil {
+				overpaymentErr = err
+				rpcsLog.Error(err)
+				continue
+			}
```
ffranr: I think we should add more context to the log message here.

rpcserver.go
```diff
@@ -7734,76 +7721,88 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
-			return err
-		}
+			// Unmarshall the accepted quote's asset rate.
+			assetRate, err := rpcutils.UnmarshalRfqFixedPoint(
+				quote.BidAssetRate,
+			)
+			if err != nil {
+				return fmt.Errorf("error unmarshalling asset "+
+					"rate: %w", err)
+			}
```
ffranr: If one of our peers returns a malformed RFQ fixed point (say if the "scale" was `> 255`) then this will fail for all peers, which is unnecessarily fragile.

rpcserver.go
```diff
@@ -7734,76 +7721,88 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
-			assetRate, peerPubKey, acceptedQuote.Scid)
+			acquiredQuotes = append(acquiredQuotes, rfqID)
 
-		var rfqID rfqmsg.ID
-		copy(rfqID[:], acceptedQuote.Id)
+			// If we reached the limit of quotes we can establish,
+			// break and continue.
+			if len(acquiredQuotes) >= rfqmsg.MaxSendPaymentQuotes {
+				break
+			}
```
ffranr: are we sampling from `chanMap` randomly? or are we going to query the first `rfqmsg.MaxSendPaymentQuotes` peers every time?

Maybe we should be more explicit about sampling from `chanMap`. If we had a peer subset selection function then we could be explicit about selection and we could expand later to weight selection based on past performance or whatever.

rpcserver.go
```diff
@@ -7734,76 +7721,88 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+		}
+
+		if len(acquiredQuotes) == 0 {
+			joinErr := errors.Join(quoteErr, overpaymentErr)
+			return fmt.Errorf("failed to acquire any quotes for "+
+				"payment: %v", joinErr)
```
ffranr: error formatting: `%v` -> `%w` here?

rpcserver.go
GeorgeTsagk: Isn't this a sanity check already handled by the rfq manager?

i.e if the quote is accepted then the asset rate is correctly formatted and also within our tolerance bounds

(reminder, the endpoint i'm calling here to get the quote is the RPC endpoint, that has all the surrounding checks included)

rpcserver.go
GeorgeTsagk: > are we sampling from chanMap randomly?

Yeah, since this is a map we're going to traverse the elements in a random order.

> If we had a peer subset selection function then we could be explicit about selection and we could expand later to weight selection based on past performance or whatever.

That's something I actually wanted to avoid. If we introduce the intelligence for picking the right RFQ peers and tracking past performance we'll end up duplicating logic that is already part of LND's mission control. I wanted to be as "raw" and transparent as possible. That's also why I placed the `MaxSendPaymentQuotes` limit quite high (`50` -- could bump higher, especially if we make the quote fetching async), so that in most cases we'll capture ALL our rfq peers in that set, and LND will do the rest.

rpcserver.go
ffranr: > Isn't this a sanity check already handled by the rfq manager?

Probably. But does that matter here? We can make this part of the code robust regardless of what the RFQ manager does.

From what I can see, the payment is abandoned if we fail to unmarshal a quote from a single peer. Maybe we should continue instead?

We should make it clear in this part of the code that a single peer can't cause grief. A reader shouldn't need to understand the RFQ package to see that.

ZZiigguurraatt:
Getting a vague error with `litcli`. Not sure why. Has anyone else tested mutli-RFQ send with `litcli` or only in itest?

```
24165ac30ed6:/$ litcli ln payinvoice --pay_req lnbcrt4440n1p5xdzkcpp5rk3tglvgl3nc8krwxv28qx4h6cvys33ta4ayhmzjcw97hs8hyj7qdqqcqzzsxqyz5vqsp5968vj7g9ngxv639jk4p9lu9mexcwlqp7r992xaj2645n0x8ay4ys9qxpqysgq46z9q5tgnq0t0e50vzsxwf7aq6cmwrajqyy0f79fe2s6jl3r40m3kxdcw8cd2q56eu6hvjdfkt8amyg0xznncejkn8mq4gp97xdsgngq6k3tgy --asset_id 12ef620e67294ac18f218e7bec769299d3eb33a3915aa3122fb1baafc14d0e53
Payment hash: 1da2b47d88fc6783d86e3314701ab7d61848462bed7a4bec52c38bebc0f724bc
Description: 
Amount (in satoshis): 444
Fee limit (in satoshis): 444
Destination: 02f57d125bb94141add3a82b9209858989000e70849005e9e031fdd6dbb388d6dd
Confirm payment (yes/no): yes
Got quote for 45065 asset units at 9 msat/unit from peer 02d63eec2b2dd001241e40a93d2b90c693f2efd8f011015accc0f73da0b22ad51e with SCID 17596861635263353475
[litcli] unexpected nil result
24165ac30ed6:/$ 
```

rpcserver.go
ffranr: > Yeah, since this is a map we're going to traverse the elements in a random order.

I think map iteration is unordered by Go spec but I don't think it's guaranteed to be random. But it's probably most likely to be random enough for us now.

Maybe we should warn log the count of peers that we never even queried? Our count of peers might be greater than `MaxSendPaymentQuotes`. Might help a super user make sense of why sometimes some of their peers are not queried.

rpcserver.go
GeorgeTsagk: Yeah, realized I totally agree, since this conversion is also only needed for a log further down, it's not critical for the actual payment

It makes sense to only log and continue

rpcserver.go
GeorgeTsagk: Changed the quote acquisition to be async, also used a shorter timeout for the RFQ negotiation

tapchannel/aux_traffic_shaper.go
guggero: I think we should add a commit at the end of this PR that removes the single RFQ ID from the `rfqmsg.NewHtlc()` function. That way we're sure we don't accidentally have this problem in other places.

tapchannel/aux_traffic_shaper.go
guggero: Comment still missing.

go.mod
```diff
@@ -217,3 +217,7 @@ replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate
 
 // Note this is a temproary replace and will be removed when taprpc is tagged.
 replace github.com/lightninglabs/taproot-assets/taprpc => ./taprpc
+
+replace github.com/lightningnetwork/lnd => github.com/GeorgeTsagk/lnd v0.0.0-20250702100924-4ad84627e108
```
guggero: Can bump this to `v0.19.2-beta.rc1` (which should not need the `sqldb` rewrite anymore).

tapchannel/aux_traffic_shaper.go
```diff
@@ -328,6 +340,10 @@ func (s *AuxTrafficShaper) paymentBandwidth(htlc *rfqmsg.Htlc,
 			return 0, err
 		}
 
+		// We know that we establish 1 quote per peer in the scope of
```
guggero, line 347: Ah, here it is. Please move this to the previous commit.

rpcserver.go
```diff
@@ -109,6 +109,11 @@ const (
 	// maxRfqHopHints is the maximum number of RFQ quotes that may be
 	// encoded as hop hints in a bolt11 invoice.
 	maxRfqHopHints = 20
+
+	// multiRfqNegotiationTimeout is the timeout within which we expect our
+	// peer and our node to reach an agreement over a quote. This timeout
+	// is a bit shorter
```
guggero: nit: incomplete comment.

rpcserver.go
```diff
@@ -8498,6 +8481,124 @@ func (r *rpcServer) acquireBuyOrder(ctx context.Context,
...
+	errorCollector := make(chan error)
+
+	// For each candidate peer in our list let's try to establish a
+	// sufficient quote for this payment. Any result that we collect will be
+	// reported back to the collector via the quote/err channels.
+	for peer := range chanMap {
+		// We make sure to always write the result to a channel before
+		// returning. This way the collector can expect a certain number
+		// of items via the channels.
+		go func(peer route.Vertex) {
```
guggero, line 9004: I think we should use `errgroup.Group` instead, would make the error cases a bit more explicit.

rpcserver.go
```diff
@@ -7738,77 +7726,72 @@ func (r *rpcServer) SendPayment(req *tchrpc.SendPaymentRequest,
...
+		)
 
-		case *rfqrpc.AddAssetSellOrderResponse_RejectedQuote:
-			return fmt.Errorf("peer %v rejected the quote, code: "+
-				"%v, error message: %v", r.RejectedQuote.Peer,
-				r.RejectedQuote.ErrorCode,
-				r.RejectedQuote.ErrorMessage)
+		// Send out the information about the acquired quotes on the
+		// stream.
+		for _, quote := range acquiredQuotes {
```
guggero: This could also be extracted into a function, would make it a bit more explicit that this only informs the user and that call to `rpcutils.UnmarshalRfqFixedPoin()` is only for the log statement.

ZZiigguurraatt:
> Getting a vague error with `litcli`. Not sure why. Has anyone else tested mutli-RFQ send with `litcli` or only in itest?
> 
> ```
> 24165ac30ed6:/$ litcli ln payinvoice --pay_req lnbcrt4440n1p5xdzkcpp5rk3tglvgl3nc8krwxv28qx4h6cvys33ta4ayhmzjcw97hs8hyj7qdqqcqzzsxqyz5vqsp5968vj7g9ngxv639jk4p9lu9mexcwlqp7r992xaj2645n0x8ay4ys9qxpqysgq46z9q5tgnq0t0e50vzsxwf7aq6cmwrajqyy0f79fe2s6jl3r40m3kxdcw8cd2q56eu6hvjdfkt8amyg0xznncejkn8mq4gp97xdsgngq6k3tgy --asset_id 12ef620e67294ac18f218e7bec769299d3eb33a3915aa3122fb1baafc14d0e53
> Payment hash: 1da2b47d88fc6783d86e3314701ab7d61848462bed7a4bec52c38bebc0f724bc
> Description: 
> Amount (in satoshis): 444
> Fee limit (in satoshis): 444
> Destination: 02f57d125bb94141add3a82b9209858989000e70849005e9e031fdd6dbb388d6dd
> Confirm payment (yes/no): yes
> Got quote for 45065 asset units at 9 msat/unit from peer 02d63eec2b2dd001241e40a93d2b90c693f2efd8f011015accc0f73da0b22ad51e with SCID 17596861635263353475
> [litcli] unexpected nil result
> 24165ac30ed6:/$ 
> ```

I seem to get this issue if the sender has multiple taproot asset channels. If the sender is a node with only a single taproot asset channel, it succeeds.

ZZiigguurraatt:
I was having a problem where I could not make a direct asset payment from alice to bob over an asset channel to an asset invoice. I was getting

```
Invoice(hash=6a1272796e8d8fb7fdb9f29669c2e94a92bb1196847275c6deafb21484207bb1, amt=4466):  has asset custom records, but invoice does not require assets, canceling HTLCs
```

then I would also get `FAILURE_REASON_INCORRECT_PAYMENT_DETAILS` on the sender's side.


this went on for a while, but now I cannot reproduce anymore, the direct asset payment now succeeds.

Just want to make a comment here in case someone else sees this behavior, or as an initial data point if I get the problem to come back somehow.

ZZiigguurraatt:
FYI: https://github.com/lightningnetwork/lnd/issues/10056 is limiting my ability to do much complicated testing with this PR.

ZZiigguurraatt:
> I was having a problem where I could not make a direct asset payment from alice to bob over an asset channel to an asset invoice. I was getting
> 
> ```
> Invoice(hash=6a1272796e8d8fb7fdb9f29669c2e94a92bb1196847275c6deafb21484207bb1, amt=4466):  has asset custom records, but invoice does not require assets, canceling HTLCs
> ```
> 
> then I would also get `FAILURE_REASON_INCORRECT_PAYMENT_DETAILS` on the sender's side.
> 
> this went on for a while, but now I cannot reproduce anymore, the direct asset payment now succeeds.
> 
> Just want to make a comment here in case someone else sees this behavior, or as an initial data point if I get the problem to come back somehow.

OK, problem came back again. This time I was on another node and used

```
litcli ln payinvoice --group_key 036bd8b9572f3c8de9e8cd4f6df56752e7616c3930de84744182da9315c0cfff34 --rfq_peer_pubkey 03db56650f632360f465e39933f3771f5285c5e18f10cb93574406ed6e720aea4b lnbcrt4440n1p5xagj8pp5992egvzvhae5wfp9q7uu9xr2u766024eh9gr884hgu3z7cpp9cqqdqqcqzzsxqzfvrzjqfye02m7q27c6lxj27svf2zxl8cfnacjvs3vvnc0gddy0lwd8g9nmah0jnvw62y08sqqqqlgqqqqqqgq2qsp5pfm92s4avpk024vj5v9px9qfky60uzxpxfhexsm55smvf00jrnhs9qxpqysgq7cxnm3n4a60ecycc5cgw8cwhktuh604k3et3jtkras05k9jadk5jgk2slvmedq8eta94n6qeyemydlanp9fc234p4zq9s7l93um0szsqqywphd
```

to avoid https://github.com/lightninglabs/taproot-assets/pull/1613#issuecomment-3032390484 , so it seems the problem is on the receiver's side and not sender's side.

On the receiver's side:

```
0c60be93a29f:/$ lncli getinfo|jq -r .identity_pubkey
03db56650f632360f465e39933f3771f5285c5e18f10cb93574406ed6e720aea4b
0c60be93a29f:/$ 
```

On the sender's side

```
c335aa6bc5a7:/$ lncli getinfo|jq -r .identity_pubkey
024997ab7e02bd8d7cd257a0c4a846f9f099f7126422c64f0f435a47fdcd3a0b3d
c335aa6bc5a7:/$ 
```

and also for reference

```
0c60be93a29f:/$ lncli decodepayreq lnbcrt4440n1p5xagj8pp5992egvzvhae5wfp9q7uu9xr2u766024eh9gr884hgu3z7cpp9cqqdqqcqzzsxqzfvrzjqfye02m7q27c6lxj27svf2zxl8cfnacjvs3vvnc0gddy0lwd8g9nmah0jnvw62y08sqqqqlgqqqqqqgq2qsp5pfm92s4avpk024vj5v9px9qfky60uzxpxfhexsm55smvf00jrnhs9qxpqysgq7cxnm3n4a60ecycc5cgw8cwhktuh604k3et3jtkras05k9jadk5jgk2slvmedq8eta94n6qeyemydlanp9fc234p4zq9s7l93um0szsqqywphd
{
    "destination": "03db56650f632360f465e39933f3771f5285c5e18f10cb93574406ed6e720aea4b",
    "payment_hash": "295594304cbf7347242507b9c2986ae7b5a7aab9b950339eb747222f60212e00",
    "num_satoshis": "444",
    "timestamp": "1752080967",
    "expiry": "300",
    "description": "",
    "description_hash": "",
    "fallback_addr": "",
    "cltv_expiry": "80",
    "route_hints": [
        {
            "hop_hints": [
                {
                    "node_id": "024997ab7e02bd8d7cd257a0c4a846f9f099f7126422c64f0f435a47fdcd3a0b3d",
                    "chan_id": "17793604312176824124",
                    "fee_base_msat": 1000,
                    "fee_proportional_millionths": 1,
                    "cltv_expiry_delta": 80
                }
            ]
        }
    ],
    "payment_addr": "0a765542bd606cf55592a30a131409b134fe08c1326f934374a436c4bdf21cef",
    "num_msat": "444000",
    "features": {
        "8": {
            "name": "tlv-onion",
            "is_required": true,
            "is_known": true
        },
        "14": {
            "name": "payment-addr",
            "is_required": true,
            "is_known": true
        },
        "17": {
            "name": "multi-path-payments",
            "is_required": false,
            "is_known": true
        },
        "25": {
            "name": "route-blinding",
            "is_required": false,
            "is_known": true
        }
    },
    "blinded_paths": []
}
0c60be93a29f:/$ 


```

ZZiigguurraatt:
> > Getting a vague error with `litcli`. Not sure why. Has anyone else tested mutli-RFQ send with `litcli` or only in itest?
> > ```
> > 24165ac30ed6:/$ litcli ln payinvoice --pay_req lnbcrt4440n1p5xdzkcpp5rk3tglvgl3nc8krwxv28qx4h6cvys33ta4ayhmzjcw97hs8hyj7qdqqcqzzsxqyz5vqsp5968vj7g9ngxv639jk4p9lu9mexcwlqp7r992xaj2645n0x8ay4ys9qxpqysgq46z9q5tgnq0t0e50vzsxwf7aq6cmwrajqyy0f79fe2s6jl3r40m3kxdcw8cd2q56eu6hvjdfkt8amyg0xznncejkn8mq4gp97xdsgngq6k3tgy --asset_id 12ef620e67294ac18f218e7bec769299d3eb33a3915aa3122fb1baafc14d0e53
> > Payment hash: 1da2b47d88fc6783d86e3314701ab7d61848462bed7a4bec52c38bebc0f724bc
> > Description: 
> > Amount (in satoshis): 444
> > Fee limit (in satoshis): 444
> > Destination: 02f57d125bb94141add3a82b9209858989000e70849005e9e031fdd6dbb388d6dd
> > Confirm payment (yes/no): yes
> > Got quote for 45065 asset units at 9 msat/unit from peer 02d63eec2b2dd001241e40a93d2b90c693f2efd8f011015accc0f73da0b22ad51e with SCID 17596861635263353475
> > [litcli] unexpected nil result
> > 24165ac30ed6:/$ 
> > ```
> 
> I seem to get this issue if the sender has multiple taproot asset channels. If the sender is a node with only a single taproot asset channel, it succeeds.

As implied in https://github.com/lightninglabs/taproot-assets/pull/1613#issuecomment-3053416318 , using `--rfq_peer_pubkey` also seems like it might be avoiding this issue on a node with multiple channels.

ZZiigguurraatt:
> > > Getting a vague error with `litcli`. Not sure why. Has anyone else tested mutli-RFQ send with `litcli` or only in itest?
> > > ```
> > > 24165ac30ed6:/$ litcli ln payinvoice --pay_req lnbcrt4440n1p5xdzkcpp5rk3tglvgl3nc8krwxv28qx4h6cvys33ta4ayhmzjcw97hs8hyj7qdqqcqzzsxqyz5vqsp5968vj7g9ngxv639jk4p9lu9mexcwlqp7r992xaj2645n0x8ay4ys9qxpqysgq46z9q5tgnq0t0e50vzsxwf7aq6cmwrajqyy0f79fe2s6jl3r40m3kxdcw8cd2q56eu6hvjdfkt8amyg0xznncejkn8mq4gp97xdsgngq6k3tgy --asset_id 12ef620e67294ac18f218e7bec769299d3eb33a3915aa3122fb1baafc14d0e53
> > > Payment hash: 1da2b47d88fc6783d86e3314701ab7d61848462bed7a4bec52c38bebc0f724bc
> > > Description: 
> > > Amount (in satoshis): 444
> > > Fee limit (in satoshis): 444
> > > Destination: 02f57d125bb94141add3a82b9209858989000e70849005e9e031fdd6dbb388d6dd
> > > Confirm payment (yes/no): yes
> > > Got quote for 45065 asset units at 9 msat/unit from peer 02d63eec2b2dd001241e40a93d2b90c693f2efd8f011015accc0f73da0b22ad51e with SCID 17596861635263353475
> > > [litcli] unexpected nil result
> > > 24165ac30ed6:/$ 
> > > ```
> > 
> > 
> > I seem to get this issue if the sender has multiple taproot asset channels. If the sender is a node with only a single taproot asset channel, it succeeds.
> 
> As implied in [#1613 (comment)](https://github.com/lightninglabs/taproot-assets/pull/1613#issuecomment-3053416318) , using `--rfq_peer_pubkey` also seems like it might be avoiding this issue on a node with multiple channels.

https://github.com/lightninglabs/taproot-assets/issues/1565#issuecomment-3025292013 might explain why `litcli ln payinvoice` is giving this weird error?

ZZiigguurraatt:
Before we merge this PR, I think you all should look at https://github.com/lightninglabs/taproot-assets/issues/1565#issuecomment-3025292013 and make sure that this is the way we want to do it so that we know that the fix for https://github.com/lightninglabs/taproot-assets/issues/1565 for `AddInvoice` is going to be consistent.

ZZiigguurraatt:
> Before we merge this PR, I think you all should look at [#1565 (comment)](https://github.com/lightninglabs/taproot-assets/issues/1565#issuecomment-3025292013) and make sure that this is the way we want to do it so that we know that the fix for #1565 for `AddInvoice` is going to be consistent.

From conversations offline, it sounds like we probably do want to change how this works and add a new field which is an array and eventually deprecate the old `accepted_sell_order` field. Reasons for this are:

- the way `SendPayment` keeps streaming a new value of the `accepted_sell_order` field, the client doesn't know when it is done.
- we don't have any plans to make an async return of `accepted_sell_order`, currently we wait for all quotes to be received and that is what we should continue to do.

litbot-9000:
@ffranr: review reminder

go.mod
```diff
@@ -217,3 +217,7 @@ replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate
 
 // Note this is a temproary replace and will be removed when taprpc is tagged.
```
ZZiigguurraatt: is this still needed?

go.mod
ZZiigguurraatt: > Can bump this to `v0.19.2-beta.rc1` (which should not need the `sqldb` rewrite anymore).

now it's `v0.19.2-beta` !

ZZiigguurraatt: seems like really this should remove the replace and update https://github.com/lightninglabs/taproot-assets/blob/d90193a88226fa87fb3c4da9fd85f7b1c0a4c030/go.mod#L33 instead?

go.mod
```diff
@@ -217,3 +217,7 @@ replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate
...
 replace github.com/lightninglabs/taproot-assets/taprpc => ./taprpc
+
+replace github.com/lightningnetwork/lnd => github.com/GeorgeTsagk/lnd v0.0.0-20250702100924-4ad84627e108
+
+replace github.com/lightningnetwork/lnd/sqldb => github.com/GeorgeTsagk/lnd/sqldb v0.0.0-20250702100924-4ad84627e108
```
ZZiigguurraatt: same with this one https://github.com/lightninglabs/taproot-assets/blob/d90193a88226fa87fb3c4da9fd85f7b1c0a4c030/go.mod#L135

go.mod
ZZiigguurraatt: like can we change https://github.com/lightninglabs/taproot-assets/blob/d90193a88226fa87fb3c4da9fd85f7b1c0a4c030/go.mod#L32 to `v1.0.9` ?

go.mod
```diff
@@ -128,7 +128,7 @@ require (
 	github.com/lightninglabs/lightning-node-connect/gbn v1.0.1 // indirect
 	github.com/lightninglabs/lightning-node-connect/mailbox v1.0.1 // indirect
 	github.com/lightninglabs/neutrino v0.16.1 // indirect
-	github.com/lightningnetwork/lightning-onion v1.2.1-0.20240712235311-98bd56499dfb // indirect
+	github.com/lightningnetwork/lightning-onion v1.2.1-0.20240815225420-8b40adf04ab9 // indirect
```
ZZiigguurraatt: What is the purpose of this change? If I rebase this PR onto main and keep this line, I get

```
# github.com/lightningnetwork/lnd/routing/blindedpath
/home/zzz/go/pkg/mod/github.com/lightningnetwork/lnd@v0.19.2-beta.rc2/routing/blindedpath/blinded_path.go:291:21: blindedPath.BlindedHops undefined (type *sphinx.BlindedPathInfo has no field or method BlindedHops)
/home/zzz/go/pkg/mod/github.com/lightningnetwork/lnd@v0.19.2-beta.rc2/routing/blindedpath/blinded_path.go:300:14: blindedPath.BlindedHops undefined (type *sphinx.BlindedPathInfo has no field or method BlindedHops)
/home/zzz/go/pkg/mod/github.com/lightningnetwork/lnd@v0.19.2-beta.rc2/routing/blindedpath/blinded_path.go:301:15: blindedPath.IntroductionPoint undefined (type *sphinx.BlindedPathInfo has no field or method IntroductionPoint)
/home/zzz/go/pkg/mod/github.com/lightningnetwork/lnd@v0.19.2-beta.rc2/routing/blindedpath/blinded_path.go:311:44: blindedPath.BlindingPoint undefined (type *sphinx.BlindedPathInfo has no field or method BlindingPoint)
/home/zzz/go/pkg/mod/github.com/lightningnetwork/lnd@v0.19.2-beta.rc2/routing/blindedpath/blinded_path.go:312:44: blindedPath.BlindedHops undefined (type *sphinx.BlindedPathInfo has no field or method BlindedHops)
```

go.mod
ZZiigguurraatt: > > Can bump this to `v0.19.2-beta.rc1` (which should not need the `sqldb` rewrite anymore).
> 
> now it's `v0.19.2-beta` !

actually lnd `v0.19.2-beta` is not compatible with this PR because it does not use `github.com/lightningnetwork/lightning-onion v1.2.1-0.20240815225420-8b40adf04ab9`

go.mod
ZZiigguurraatt: it seems tapd main branch is on lnd `v0.19.2-beta`, but lnd `v0.19.2-beta` does not yet use `github.com/lightningnetwork/lightning-onion v1.2.1-0.20240815225420-8b40adf04ab9`, so that is the cause of the error messages. it seems https://github.com/lightningnetwork/lnd/pull/9980 was cherry picked onto lnd `v0.19.2-beta` off of master and that is why I got confused.

taprpc/tapchannelrpc/tapchannel.proto
```diff
@@ -17,14 +17,17 @@ service TaprootAssetChannels {
     rpc FundChannel (FundChannelRequest) returns (FundChannelResponse);
 
     /*
+    Deprecated.
```
ffranr, line 20: We should consider adding a comment here to guide the user on the appropriate alternative action.

tapchannel/aux_traffic_shaper.go
GeorgeTsagk: The single RFQ ID still plays a role, let me refresh the semantics of the two:

`RfqID` is the RFQ that is being used to convert msats<->units and is present when the HTLC gets added over the channel

`AvailableRfqIDs` is an array of Rfq IDs that is used by LND to query bandwidth. When a channel with bandwidth (that corresponds to one of the RFQs in the array) is selected and an HTLC is sent over it then we produce the HtlcExtraData that include that RFQ ID as an `RfqID`

There's 2 reasons to keep both fields:

- Separation of concerns: `RfqID` is the RFQ of the HTLC that is sent to our peer, it's a 1-1 relationship (i.e we're committed to it in the context of this HTLC being sent over). The `AvailableRfqIDs` is an array of hints (to discover more bandwidth across different channels), one of which may eventually be encoded as an `RfqID` when an HTLC attempt is made

- Backwards compatibility: If only one peer upgrades to the latest tapd version and they're now using the new `AvailableRfqIDs` field then we won't run into any issues as that record is only present in the LND<>Tapd communication via the `AuxTrafficShaper`. Once the HTLC leaves our local daemon then we'll write the RFQ as an `RfqID` which is what our peer expects to read.

rpcserver.go
GeorgeTsagk: Oh, interesting pkg, was not aware of it. Seems though that `errgroup.Group` will return 1st error and stop all other routines. In order to collect all errors I'd still have to use some channel or mutex-protected array of errors.

IMO it doesn't look like it has that much to offer. (perhaps I'm not looking at the right way of using this structure?)

Switching to mutex-protected arrays for responses/errors and then using a simple wait group can also do the job

go.mod
GeorgeTsagk: this was probably a leftover of the old dependency tree

we can use `98bd56499dfb` after rebasing

taprpc/tapchannelrpc/tapchannel.proto
GeorgeTsagk: I'm not sure there even is an alternative action? We're deprecating this because as far as we know nobody is using this anymore

go.mod
GeorgeTsagk: we should be able to, will see if I can actually remove this replacement

go.mod
ZZiigguurraatt: i think after tapd and litd move off of lnd `v0.19.2-beta` and onto lnd master, we will have to put `20240815225420-8b40adf04ab9` back in though?

taprpc/tapchannelrpc/tapchannel.proto
ffranr: Ah, was this RPC endpoint only used internally? And since we don’t think it’s useful to users, we’re removing it—is that the idea?

go.mod
```diff
@@ -214,6 +214,3 @@ replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-d
 // We are using a fork of the migration library with custom functionality that
 // did not yet make it into the upstream repository.
 replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate/v4 v4.18.2-9023d66a-fork-pr-2
-
-// Note this is a temproary replace and will be removed when taprpc is tagged.
-replace github.com/lightninglabs/taproot-assets/taprpc => ./taprpc
```
guggero: This will make future PRs that change anything in the `taprpc` package super hard (needs to be two PRs). So I think we decided to leave this in. But we should update the version on line 32 regularly so external consumers will pull in the correct version.

tapchannel/aux_traffic_shaper.go
guggero: Okay, makes sense. Thanks for the explanation. I think the content of this comment should go into the Godoc of the two fields. Otherwise this semantic will be hard to get by in the future.

guggero: And the comment here should also explain more why it's okay that we only check the _available_ HTLCs. Currently the comments here are quite sparse and mostly describe the code, not the _why_ behind the code.

tapchannel/aux_traffic_shaper.go
guggero: We should add the same comment here that explains the _why_ for this check and why it's okay to only look at the available IDs.

rpcserver.go
```diff
@@ -8359,6 +8327,42 @@ func calculateAssetMaxAmount(ctx context.Context, priceOracle rfq.PriceOracle,
...
+func notifyUserForPaymentQuotes(
+	stream tchrpc.TaprootAssetChannels_SendPaymentServer,
+	quote *rfqrpc.PeerAcceptedSellQuote, payHash string,
+	peerPubKey *route.Vertex) error {
+
+	var err error
+
+	err = stream.Send(&tchrpc.SendPaymentResponse{
+		Result: &tchrpc.SendPaymentResponse_AcceptedSellOrder{
+			AcceptedSellOrder: quote,
```
guggero: I think streaming multiple quotes to the user is probably not the best idea, since there is no way for the user to know when they're done reading (other than the stream being closed).
So my idea was to add a second field to this RPC message, `AcceptedSellOrders` that is a slice of quotes. Then we can deprecate the old field (and update its documentation to say it will only ever contain the _first_ quote and that clients must update to receive the full list).

rpcserver.go
guggero: Ah, you're right, if you want all errors then the `errgroup.Group` isn't useful.
But I think we already have a helper for exactly what you need, that does all the mutex locking and stuff: `fn.ParSliceErrCollect()`.

rpcserver.go
ZZiigguurraatt: > I think streaming multiple quotes to the user is probably not the best idea, since there is no way for the user to know when they're done reading (other than the stream being closed). So my idea was to add a second field to this RPC message, `AcceptedSellOrders` that is a slice of quotes. Then we can deprecate the old field (and update its documentation to say it will only ever contain the _first_ quote and that clients must update to receive the full list).

yeah, summary was here https://github.com/lightninglabs/taproot-assets/pull/1613#issuecomment-3058019551 . not sure if @GeorgeTsagk  was part of that conversation though??

ZZiigguurraatt:
Would like to see some release notes for this PR similar to https://github.com/lightninglabs/taproot-assets/pull/1587/commits/a74f887e876599956829effd41089e520dd02206 in https://github.com/lightninglabs/taproot-assets/pull/1587

rpcserver.go
GeorgeTsagk: I think the clients know when to terminate the stream once they receive the payment updates (since the RFQ responses precede the payment update)

Regardless, I went ahead with this change and it's added in the new diff

GeorgeTsagk:
Changed the RPC response stream to now return all the quotes in an array.
Also added a release note for the introduced multi-rfq feature.

rpcserver.go
GeorgeTsagk: Hmm that would be useful, but then I'd still have to maintain some other channel to collect the responses

Could try to expand `ParSliceErrCollect` to also include returned responses?

guggero: Yes, perhaps something like this?

```

// ResultFunc is a type def for a function that takes a context (to allow early
// cancellation) and a series of value returning a Result. This is typically
// used a closure to perform concurrent work over a homogeneous slice of
// values.
type ResultFunc[V any, R fn.Result[any]] func(context.Context, V) R

// ParSliceResultCollect can be used to execute a function on each element of a
// slice in parallel. This function is fully blocking and will wait for all
// goroutines to finish (subject to context cancellation/timeout). Any results
// will be collected and returned as a map of slice element index to Result.
// Active goroutines limited with number of CPU.
func ParSliceResultCollect[V any, R fn.Result[any]](ctx context.Context, s []V,
	f ResultFunc[V, R]) (map[int]R, error) {

	errGroup, ctx := errgroup.WithContext(ctx)
	errGroup.SetLimit(runtime.GOMAXPROCS(0))

	var instanceResultsMutex sync.Mutex
	instanceResults := make(map[int]R, len(s))

	for idx := range s {
		errGroup.Go(func() error {
			result := f(ctx, s[idx])
			instanceResultsMutex.Lock()
			instanceResults[idx] = result
			instanceResultsMutex.Unlock()

			// Avoid returning an error here, as that would cancel
			// the errGroup and terminate all slice element
			// processing instances. Instead, collect the error and
			// return it later.
			return nil
		})
	}

	// Now we will wait/block for all goroutines to finish.
	//
	// The goroutines that are executing in parallel should not return an
	// error, but the Wait call may return an error if the context is
	// canceled or timed out.
	err := errGroup.Wait()
	if err != nil {
		return nil, fmt.Errorf("failed to wait on error group in "+
			"ParSliceErrorCollect: %w", err)
	}

	return instanceResults, nil
}
```

guggero: But can do in a follow-up PR.

rpcserver.go
```diff
@@ -8794,35 +8792,57 @@ func calculateAssetMaxAmount(ctx context.Context, priceOracle rfq.PriceOracle,
...
+		})
+		if err != nil {
+			return err
+		}
+	}
 
-	return nil
+	// We now stream the full array of negotiated quotes to the response
+	// stream.
+	return stream.Send(&tchrpc.SendPaymentResponse{
```
guggero, line 8842: :+1: looks good. This will require a small adjustment in `litcli` as well, just as a reminder.

guggero (approved):
LGTM :tada:

rfqmsg/records.go
```diff
@@ -42,13 +49,107 @@ type (
...
+	// Note that we set the type here as zero, as when used with a
+	// tlv.RecordT, the type param will be used as the type.
+	return tlv.MakeDynamicRecord(
+		0, r, size, encodeAvailableIDs, decodeRfqIDs,
+	)
+}
+
+// encodeAvailableIDs encodes the RfqIDs struct on the wire.
+func encodeAvailableIDs(w io.Writer, val any, buf *[8]byte) error {
```
ffranr: i think we should rename this function to mirror the `decodeRfqIDs` function below, so: `encodeRfqIDs`

rpcserver.go
```diff
@@ -8821,6 +8789,42 @@ func calculateAssetMaxAmount(ctx context.Context, priceOracle rfq.PriceOracle,
 	return maxMathUnits.ToUint64(), nil
 }
 
+// notifyUserForPaymentQuotes sends the accepted quotes to the RPC user of
+// SendPayment via the response stream. It also logs info on each accepted
+// quote.
+func notifyUserForPaymentQuotes(
```
ffranr, line 8793: I think you should squash commits:
* [taprpc+rpcserver: SendPayment returns quotes in new rpc field](https://github.com/lightninglabs/taproot-assets/pull/1613/commits/87d44714291fba4e1690f907d4d3b982190aae41)
* [rpcserver: SendPayment uses multiple RFQ quotes](https://github.com/lightninglabs/taproot-assets/pull/1613/commits/2e0b138a4e15670938b4aaa095e3e3ad6db105ec)

rpcserver.go
```diff
@@ -8794,35 +8792,57 @@ func calculateAssetMaxAmount(ctx context.Context, priceOracle rfq.PriceOracle,
...
-	if err != nil {
-		rpcsLog.Errorf("error unmarshalling asset "+
-			"rate for payment %v: %w", payHash, err)
+		// Unmarshall the accepted quote's asset rate.
+		assetRate, err := rpcutils.UnmarshalRfqFixedPoint(
+			quote.BidAssetRate,
+		)
+		if err != nil {
+			rpcsLog.Errorf("error unmarshalling asset rate for "+
+				"payment %v: %w", payHash, err)
```
ffranr: my IDE gives a lint error for `%w` used in anything but `fmt.Errorf`. We should probably not use `%w` here?

ffranr: `assetRate` is porbably `nil` at this point.

I think we should log more info in the `rpcsLog.Errorf` call and then `continue` so that we don't even attempt the `rpcsLog.Infof` call below. Would be good to include `peerPubKey` and `quote.Scid` in this error message.

PR #1614, opened 2025-06-23
[group key addrs 4/5]: implement mailbox server and message store database

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/1611.

Extracted some commits from https://github.com/lightninglabs/taproot-assets/pull/1587.

Implements the SQL store for the auth mailbox server and fully hooks up the server in `tapd`.
The implementation is covered by a basic message send and receive integration test.

### DIFF
--- a/asset/generators.go
+++ b/asset/generators.go
@@ -2,6 +2,7 @@ package asset
 
 import (
 	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"math"
@@ -114,11 +115,12 @@ var (
 		}
 	})
 	GenesisGen = rapid.Custom(func(t *rapid.T) Genesis {
+		maxTagLength := hex.DecodedLen(MaxAssetNameLength)
 		return Genesis{
 			FirstPrevOut: OutPointGen.Draw(t, "first_prev_out"),
-			Tag: rapid.StringN(
-				-1, -1, MaxAssetNameLength,
-			).Draw(t, "tag"),
+			Tag: hex.EncodeToString(rapid.SliceOfN(
+				rapid.Byte(), maxTagLength, maxTagLength,
+			).Draw(t, "tag")),
 			MetaHash: rapid.Make[[32]byte]().Draw(
 				t, "meta_hash",
 			),
--- a/tapdb/supply_tree_test.go
+++ b/tapdb/supply_tree_test.go
@@ -372,16 +372,12 @@ func setupSupplyTreeTestForProps(t *testing.T) (*SupplyTreeStore,
 func TestSupplyTreeStoreApplySupplyUpdates(t *testing.T) {
 	t.Parallel()
 
-	supplyStore, spec, eventGen := setupSupplyTreeTestForProps(
-		t,
-	)
+	supplyStore, spec, eventGen := setupSupplyTreeTestForProps(t)
 	ctxb := context.Background()
 
 	// Draw a random list of supply update events. Limit the number
 	// of events to keep test execution time reasonable.
-	updates := rapid.SliceOfN(
-		eventGen, 1, 20,
-	).Example()
+	updates := rapid.SliceOfN(eventGen, 1, 20).Example()
 
 	// Apply the updates.
 	finalRootSupplyRoot, err := supplyStore.ApplySupplyUpdates(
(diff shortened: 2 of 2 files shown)


authmailbox/server.go
```diff
@@ -257,18 +260,19 @@ func (s *Server) SendMessage(ctx context.Context,
...
-	err = s.cfg.MsgStore.StoreMessage(msg)
+	msgID, err := s.cfg.MsgStore.StoreMessage(ctx, claimedOp, msg)
 	if err != nil {
 		return nil, fmt.Errorf("error storing message: %w", err)
 	}
 
 	// Publish the message to all subscribers, meaning it will be
 	// distributed to all subscribed streams. Each stream will filter by
 	// recipient ID by itself.
+	msg.ID = msgID
```
ffranr, line 271: We assign `msg.ID` to the db table’s auto-incrementing primary key. Exposing that surrogate key externally couples the database to the API.

Another possible solution: generate a UUIDv4 for each message instead, store it in the db row, and keep the integer primary key internal to the db only. This decouples external identifiers from the DB’s bookkeeping (makes table rebuilds/merge safer because message IDs never collide?).

Not a big deal, but something that might be worth exploring.

authmailbox/server.go
```diff
@@ -234,13 +236,14 @@ func (s *Server) SendMessage(ctx context.Context,
 		// We didn't have the proof before, so we store it now. If at
 		// the same time a different goroutine is trying to store the
 		// same proof, we expect the database to handle the concurrency.
-		err = s.cfg.TxProofStore.StoreProof(txProof.ClaimedOutPoint)
+		err = s.cfg.TxProofStore.StoreProof(ctx, *txProof)
```
ffranr, line 239: :+1:

Roasbeef:
Needs to be rebased due to recent merges increasing the highest migration number.

authmailbox/server.go
Roasbeef: Do we actually even use this ID on the proto layer? 

IIUC, the client uses a block height range to scan for new messages.

tapdb/sqlc/migrations/000036_authmailbox.up.sql
```diff
@@ -0,0 +1,54 @@
...
+    
+    -- The encrypted payload of the message, which contains the sender's
+    -- ephemeral public key as part of the additional data (along with the nonce
+    -- used for the ECIES-HKDF-SHA256-XCHA20POLY1305 encryption).
+    encrypted_payload BLOB NOT NULL,
+    
+    -- The timestamp when the message was created on the server. This is a unix
+    -- timestamp in seconds to allow for easy querying and sorting of messages
+    -- based on their arrival time, without time zone complications.
+    arrival_timestamp BIGINT NOT NULL,
```
Roasbeef: 👍

tapdb/authmailbox.go
```diff
@@ -0,0 +1,335 @@
...
+			return fmt.Errorf("error inserting message: %w", err)
+		}
+
+		return nil
+	})
+	if dbErr != nil {
+		// Add context to unique constraint errors.
+		var uniqueConstraintErr *ErrSqlUniqueConstraintViolation
+		if errors.As(dbErr, &uniqueConstraintErr) {
+			return 0, proof.ErrTxMerkleProofExists
```
Roasbeef, line 133: 👍

tapdb/interfaces.go
```diff
@@ -35,12 +35,39 @@ const (
...
-// database transaction is created. Transaction can wither be read or write.
+// database transaction is created. Transaction can either be read or write.
 type TxOptions interface {
-	// ReadOnly returns true if the transaction should be read only.
+	// ReadOnly returns true if the transaction should be read-only.
 	ReadOnly() bool
 }
 
+// BaseTxOptions defines the set of db txn options the database understands.
+type BaseTxOptions struct {
```
Roasbeef, line 45: Nice. TBH, we can prob get rid of all the specific tx options we have like `BaseUniverseStoreOptions` and just use this instead. Non-blocking can do it in some other PR as it'll increase the diff size and scope here.

proof/tx.go
```diff
@@ -418,9 +419,9 @@ func UnmarshalTxProof(
 type TxProofStore interface {
 	// HaveProof returns true if the proof for the given outpoint exists in
 	// the store.
-	HaveProof(wire.OutPoint) (bool, error)
+	HaveProof(context.Context, wire.OutPoint) (bool, error)
```
Roasbeef, line 422: 👍

authmailbox/server.go
guggero: We use the ID to query as well. That makes things a bit easier if you know exactly which ID you received last. Because there could be multiple messages per block, so just using the block height might give you messages you've already seen.

Having a comparable (e.g. numeric) ID is useful here, exactly for the above case. Why do you think exposing the database primary ID is a bad thing on the API level, @ffranr? It's really just a number...

PR #1619, opened 2025-06-25
ecies: add version byte prefix to encoding format

ffranr:
Extend the encoding format to include a version byte prefix. This helps facilitate cleaner and safer upgrades in the future.

### DIFF
--- a/internal/ecies/ecies.go
+++ b/internal/ecies/ecies.go
@@ -25,6 +25,35 @@ const (
 	protocolName = "ECIES-HKDF-SHA256-XCHA20POLY1305"
 )
 
+// Version represents the version of the ECIES encoding format.
+type Version uint8
+
+const (
+	// VersionUndefined is the undefined version of the ECIES encoding
+	// format. It is used to indicate that the version is not set or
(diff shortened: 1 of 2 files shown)


guggero (approved):
Very nice, thank you! LGTM :tada: 

The litd itest is going to be fixed by https://github.com/lightninglabs/taproot-assets/pull/1614, so we can ignore that here.

internal/ecies/ecies.go
```diff
@@ -69,44 +99,59 @@ func EncryptSha256ChaCha20Poly1305(sharedSecret [32]byte, msg []byte,
 	ciphertext := aead.Seal(nonce, nonce, msg, additionalData)
 
 	var result bytes.Buffer
+	result.WriteByte(byte(latestVersion))
```
GeorgeTsagk, line 102: i don't get the bigger picture here, does adding this extra byte to describe the version break compatibility in any way?

ffranr: The `ecies` code is new and hasn't been used yet. It was added as part of the group key TAP address work that's still ongoing.

PR #1621, opened 2025-06-26
[group key addrs 5/6]: Add Pedersen unique script key type

guggero:
More commits extracted from https://github.com/lightninglabs/taproot-assets/pull/1587 that contain the last (mostly) non-functional changes that can be merged independently before we're starting to change the address behavior.


guggero:
Going to look into the test failures (I think the mock just needs some mutexes).
But requesting review on this anyway so folks can load up on context.

tapdb/addrs.go
```diff
@@ -117,6 +125,13 @@ type AddrBook interface {
 	FetchAddrByTaprootOutputKey(ctx context.Context,
 		arg []byte) (AddrByTaprootOutput, error)
 
+	// FetchAddrByScriptKeyAndVersion returns a single address based on its
+	// script key and version or a sql.ErrNoRows error if no such address
+	// exists.
+	FetchAddrByScriptKeyAndVersion(ctx context.Context,
+		arg AddrByScriptKeyAndVersion) (AddrByScriptKeyAndVersionRow,
+		error)
+
```
ffranr: Maybe we should just add the script key and version to `AddrQuery` and re-use `FetchAddrs`?

tapdb/addrs.go
```diff
@@ -866,8 +927,20 @@ func (t *TapAddressBook) QueryAddrEvents(
...
+			case errors.Is(err, sql.ErrNoRows):
+				return address.ErrNoAddr
+
+			case err != nil:
+				return err
+			}
+
+			addr, err := parseAddr(
+				ctx, db, t.params, row.Addr, row.ScriptKey,
+				row.InternalKey, row.InternalKey_2,
```
ffranr, line 935: error message context below these lines is now stale, should relate to parsing.

tapdb/sqlc/queries/addrs.sql
```diff
@@ -84,6 +80,21 @@ JOIN internal_keys taproot_keys
   ON addrs.taproot_key_id = taproot_keys.key_id
 WHERE taproot_output_key = $1;
 
+-- name: FetchAddrByScriptKeyAndVersion :one
```
ffranr: we could add documentation to this query also (but not very standard in our codebase)

asset/asset.go
```diff
@@ -152,6 +152,61 @@ const (
...
+var (
+	// AllScriptKeyTypes is a slice of all known script key types.
+	AllScriptKeyTypes = []ScriptKeyType{
+		ScriptKeyUnknown,
+		ScriptKeyBip86,
+		ScriptKeyScriptPathExternal,
+		ScriptKeyBurn,
+		ScriptKeyTombstone,
+		ScriptKeyScriptPathChannel,
+	}
```
ffranr: :+1:

asset/asset.go
```diff
@@ -152,6 +152,61 @@ const (
...
+	dbTypes := fn.CopySlice(AllScriptKeyTypes)
+
+	// For some RPCs (mostly balance related), we exclude the assets that
+	// are specifically used for funding custom channels by default. The
+	// balance of those assets is reported through lnd channel balance.
+	// Those assets are identified by the specific script key type for
+	// channel keys. We exclude them unless explicitly queried for.
+	if filterChannelRelated {
+		dbTypes = fn.CopySlice(ScriptKeyTypesNoChannel)
+	}
```
ffranr: I think `filterChannelRelated` should be called `excludeChannels`. "Filter" to me sounds like they might be included.

asset/asset.go
```diff
@@ -1098,6 +1115,65 @@ func EqualKeyDescriptors(a, o keychain.KeyDescriptor) bool {
...
+const (
+	// ScriptKeyDerivationUniquePedersen means the script key is derived
+	// using the address's recipient ID key and a single leaf that contains
+	// an un-spendable Pedersen commitment key
+	// (OP_CHECKSIG <NUMS_key + asset_id * G>). This can be used to
+	// create unique script keys for each virtual packet in the fragment,
+	// to avoid proof collisions in the universe, where the script keys
+	// should be spendable by a hardware wallet that only supports
+	// miniscript policies for signing P2TR outputs.
+	ScriptKeyDerivationUniquePedersen ScriptKeyDerivationMethod = 0
```
ffranr, line 1138: why not use `OP_RETURN` method here rather than pedersen? Are we leaving the door open for hardware wallets, is that the idea here still?

authmailbox/multi_subscription.go
```diff
@@ -0,0 +1,147 @@
...
+		msgQueue:         queue,
+	}
+}
+
+// Subscribe adds a new subscription for the specified client URL and receiver
+// key. It starts a new mailbox client if one does not already exist for the
+// given URL. The subscription will receive messages that match the provided
+// filter and will send them to the shared message queue.
+func (m *MultiSubscription) Subscribe(serverURL url.URL,
+	receiverKey keychain.KeyDescriptor, filter MessageFilter) error {
```
ffranr: Maybe this should be thread safe? I'm not sure how this might be used. But if it's going to be called from an RPC endpoint then it probably needs to be thread safe.

proof/send.go
```diff
@@ -0,0 +1,187 @@
...
+	MaxSendFragmentOutputs = 256
+)
+
+// SendFragmentVersion is the version of the send fragment.
+type SendFragmentVersion uint8
+
+const (
+	// SendFragmentV0 is the first version of the send fragment.
+	SendFragmentV0 SendFragmentVersion = 0
+)
```
ffranr: Might be useful to reserve `0` for "unspecified version". Useful for returning errors on version parsing.

Might also be useful to have a "latest version" var here.

tapdb/sqlc/queries/addrs.sql
```diff
@@ -67,14 +67,10 @@ LIMIT @num_limit OFFSET @num_offset;
 
 -- name: FetchAddrByTaprootOutputKey :one
 SELECT
-    version, asset_version, genesis_asset_id, group_key, tapscript_sibling,
-    taproot_output_key, amount, asset_type, creation_time, managed_from,
-    proof_courier_addr,
+    sqlc.Embed(addrs),
```
Roasbeef: Style nit: `Embed` is capitalized here but lower case below. 

Doesn't seem to affect the `sqlc` mapping though.

scripts/gen_sqlc_docker.sh
```diff
@@ -45,4 +45,4 @@ docker run \
 	-e UID=$UID \
 	-v "$DIR/../:/build" \
 	-w /build \
-	sqlc/sqlc:1.28.0 generate
+	sqlc/sqlc:1.29.0 generate
```
Roasbeef, line 49: 👍

asset/asset.go
```diff
@@ -152,6 +152,61 @@ const (
...
+	// Those assets are identified by the specific script key type for
+	// channel keys. We exclude them unless explicitly queried for.
+	if filterChannelRelated {
+		dbTypes = fn.CopySlice(ScriptKeyTypesNoChannel)
+	}
+
+	// If the user specified a script key type, we use that to filter the
+	// results.
+	userSpecified.WhenSome(func(t ScriptKeyType) {
+		dbTypes = []ScriptKeyType{t}
```
Roasbeef: Since this overrides the logic above, we can move it up to the top. 

Can also re-work to be something like: 
```go
dbTypes := userSpecified.UnwrapOr(nil)
if len(dbTypes) != 0  {
    return dbTypes
}
```

Then the rest of the logic that'll copy/filter the types.

scripts/gen_sqlc_docker.sh
```diff
@@ -46,3 +47,25 @@ docker run \
...
+# comment into the actual SQL query. But we add the comment ourselves and now
+# just need to replace the '$X/*SLICE:<field_name>*/' placeholders with the
+# actual placeholder that's going to be replaced by the sqlc generated code.
+echo "Applying sqlc.slice() workaround..."
+for file in tapdb/sqlc/*.sql.go; do
+  echo "Patching $file"
+
+  # First, we replace the `$X/*SLICE:<field_name>*/` placeholders with
+  # the actual placeholder that sqlc will use: `/*SLICE:<field_name>*/?`.
+  sed -i.bak -E 's/\$([0-9]+)\/\*SLICE:([a-zA-Z_][a-zA-Z0-9_]*)\*\//\/\*SLICE:\2\*\/\?/g' "$file"
```
Roasbeef, line 63: 😵

asset/asset.go
```diff
@@ -152,6 +152,61 @@ const (
...
+	}
+)
+
+// ScriptKeyTypeForDatabaseQuery returns a slice of script key types that should
+// be used when querying the database for assets. The returned slice will either
+// contain all script key types or only those that are not related to channels,
+// depending on the `filterChannelRelated` parameter. Unless the user specifies
+// a specific script key type, in which case the returned slice will only
+// contain that specific script key type.
+func ScriptKeyTypeForDatabaseQuery(filterChannelRelated bool,
```
Roasbeef: Perhaps this should instead live in the DB? As it's only used there for now, and makes some assumptions w.r.t _how_ the types are stored in the db. 

Eg: IIUC, if someone re-orders the script key types in the definition above (eg: by accident when adding a new type), queries will break.

This is generally why I lean towards making these db enum types into their own table, as then we lean on the database engine to help us enforce referential integrity.

proof/records.go
```diff
@@ -54,6 +54,12 @@ const (
 	MetaRevealUniverseCommitments    tlv.Type = 7
 	MetaRevealCanonicalUniversesType tlv.Type = 9
 	MetaRevealDelegationKeyType      tlv.Type = 11
+
+	SendFragmentVersionType     tlv.Type = 0
```
Roasbeef: Arguably could live in the `address` package, as IIUC we're not placing this information into proofs _yet_.

Roasbeef (changes requested):

asset/asset.go
guggero: Yes, mainly for the slightly easier hardware wallet support of the Pedersen construction.

scripts/gen_sqlc_docker.sh
guggero: Yeah, regular expressions (combined with the required escaping needed for the shell script and `sed` syntax) aren't easy to read. But I guess we can always ask Gemini to explain it to us :stuck_out_tongue_closed_eyes: I did provide before and after examples to hopefully make them a bit easier to understand.

asset/asset.go
guggero: I moved this specifically here to be close to the different script key types. The idea is that we don't forget to add new script keys to the two lists (`AllScriptKeyTypes` and `ScriptKeyTypesNoChannel`).
I also felt it was weird having this business logic in the database itself.

> assumptions w.r.t how the types are stored in the db

No, we don't make any assumptions, the type is a simple numeric type. Sure, the database doesn't know what possible types there are, but other than that there are no assumptions that I can think of.

> IIUC, if someone re-orders the script key types in the definition above (eg: by accident when adding a new type), queries will break.

That luckily isn't the case. Because we use `WHERE script_key_type IN (?, ?, ?, ...)`, the order of the enum-like values doesn't matter.
Perhaps you have the event status in mind where we still do this `status >= @status_from AND status <= @status_to` in 
https://github.com/lightninglabs/taproot-assets/blob/main/tapdb/sqlc/queries/addrs.sql#L170

We can turn that into a `WHERE status IN (?, ?, ?, ...)` query now too to make the query order-independent.

authmailbox/multi_subscription.go
guggero: Currently we only access this from a single Goroutine in the asset custodian. But doesn't hurt to make it thread safe, so I did that.

proof/records.go
guggero: Ah yes, makes sense.

guggero:
I addressed all comments, should be ready for another round of review.

Roasbeef (approved):
LGTM 💃🏾

proof/send.go
```diff
@@ -0,0 +1,218 @@
...
+type SendFragmentVersion uint8
+
+const (
+	// SendFragmentVersionUnknown is the version of the send fragment that
+	// indicates an unknown version. This is used to signal that the
+	// fragment version is not recognized.
+	SendFragmentVersionUnknown SendFragmentVersion = 0
+
+	// SendFragmentV0 is the first version of the send fragment.
+	SendFragmentV0 SendFragmentVersion = 1
```
ffranr, line 36: might be nice to call this `SendFragmentV1` so that the enum value corresponds to the version in code, might be confusing otherwise

guggero: Yeah, makes sense. I'll rename it in the follow-up PR.

PR #1627, opened 2025-06-30
`AddInvoice` correctly handles existing route hints

### LINKED ISSUE #1622: [bug]: Runtime error when validating asset invoice amount
### Background

I am creating an asset invoice using `tapchannelrpc.AddInvoice` where the `AddInvoiceRequest` passed in uses a `lnrpc.Invoice` in the  `InvoiceRequest` field that contains `RouteHints`. In these route hints I have 1 hop hint which contains the `NodeId` of a direct peer and a `ChanId` with a value that came from a `PeerAcceptedBuyQuote`  where I grabbed its `Scid` value. This `AddInvoice` calls fails with:

```
rpc error: code = Unknown desc = unexpected HTTP status code received from server: 500 (Internal Server Error); transport: received unexpected content-type "text/plain; charset=utf-8
```
and the node crashes.

Here are logs I was able to gather of the runtime error:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x60 pc=0x20f7258]

goroutine 599249 [running]:
github.com/lightninglabs/taproot-assets.validateInvoiceAmount(0x0, 0x125116e3, 0x4014897240?)
	github.com/lightninglabs/taproot-assets@v0.6.0/rpcserver.go:8344 +0x98
github.com/lightninglabs/taproot-assets.(*rpcServer).AddInvoice(0x4003231080, {0x43764b8, 0x4008657200}, 0x4014550c60)
	github.com/lightninglabs/taproot-assets@v0.6.0/rpcserver.go:8156 +0x870
github.com/lightninglabs/taproot-assets/taprpc/tapchannelrpc._TaprootAssetChannels_AddInvoice_Handler.func1({0x43764b8?, 0x4008657200?}, {0x2b36140?, 0x4014550c60?})
	github.com/lightninglabs/taproot-assets/taprpc@v1.0.7/tapchannelrpc/tapchannel_grpc.pb.go:264 +0xd0
github.com/lightningnetwork/lnd/rpcperms.(*InterceptorChain).CreateServerOpts.(*InterceptorChain).middlewareUnaryServerInterceptor.func7({0x43764b8, 0x4008657200}, {0x2b36140, 0x4014550c60}, 0x4009b9aa00, 0x400e6fcba0)
	github.com/lightningnetwork/lnd@v0.19.1-beta/rpcperms/interceptor.go:832 +0xec
google.golang.org/grpc.getChainUnaryHandler.func1({0x43764b8, 0x4008657200}, {0x2b36140, 0x4014550c60})
	google.golang.org/grpc@v1.65.0/server.go:1196 +0xa0
github.com/lightningnetwork/lnd/rpcperms.(*InterceptorChain).CreateServerOpts.(*InterceptorChain).MacaroonUnaryServerInterceptor.func5({0x43764b8, 0x4008657200}, {0x2b36140, 0x4014550c60}, 0x4009b9aa00?, 0x4007300e80)
	github.com/lightningnetwork/lnd@v0.19.1-beta/rpcperms/interceptor.go:689 +0x90
google.golang.org/grpc.getChainUnaryHandler.func1({0x43764b8, 0x4008657200}, {0x2b36140, 0x4014550c60})
	google.golang.org/grpc@v1.65.0/server.go:1196 +0xa0
github.com/lightningnetwork/lnd/rpcperms.(*InterceptorChain).CreateServerOpts.(*InterceptorChain).rpcStateUnaryServerInterceptor.func3({0x43764b8, 0x4008657200}, {0x2b36140, 0x4014550c60}, 0x4009b9aa00, 0x4007300e40)
	github.com/lightningnetwork/lnd@v0.19.1-beta/rpcperms/interceptor.go:781 +0x108
google.golang.org/grpc.getChainUnaryHandler.func1({0x43764b8, 0x4008657200}, {0x2b36140, 0x4014550c60})
	google.golang.org/grpc@v1.65.0/server.go:1196 +0xa0
github.com/lightningnetwork/lnd/rpcperms.(*InterceptorChain).CreateServerOpts.errorLogUnaryServerInterceptor.func1({0x43764b8?, 0x4008657200?}, {0x2b36140?, 0x4014550c60?}, 0x4009b9aa00, 0x400e6fcba0?)
	github.com/lightningnetwork/lnd@v0.19.1-beta/rpcperms/interceptor.go:605 +0x48
google.golang.org/grpc.NewServer.chainUnaryServerInterceptors.chainUnaryInterceptors.func1({0x43764b8, 0x4008657200}, {0x2b36140, 0x4014550c60}, 0x4009b9aa00, 0x2beef60?)
	google.golang.org/grpc@v1.65.0/server.go:1187 +0x88
github.com/lightninglabs/taproot-assets/taprpc/tapchannelrpc._TaprootAssetChannels_AddInvoice_Handler({0x2d27f00, 0x4000acf560}, {0x43764b8, 0x4008657200}, 0x4003ddd880, 0x400015bdc0)
	github.com/lightninglabs/taproot-assets/taprpc@v1.0.7/tapchannelrpc/tapchannel_grpc.pb.go:266 +0x148
google.golang.org/grpc.(*Server).processUnaryRPC(0x400004f600, {0x43764b8, 0x4008657020}, {0x4394080, 0x400032a000}, 0x4000b05d40, 0x4000a8c390, 0x63b4190, 0x0)
	google.golang.org/grpc@v1.65.0/server.go:1379 +0xb58
google.golang.org/grpc.(*Server).handleStream(0x400004f600, {0x4394080, 0x400032a000}, 0x4000b05d40)
	google.golang.org/grpc@v1.65.0/server.go:1790 +0xb20
google.golang.org/grpc.(*Server).serveStreams.func2.1()
	google.golang.org/grpc@v1.65.0/server.go:1029 +0x84
created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 493
	google.golang.org/grpc@v1.65.0/server.go:1040 +0x13c
```

My 2 cents is that [this check is failing](https://github.com/lightninglabs/taproot-assets/blob/cf2a5e5bb47cdfb3b8a6a570113ca5923dbbf805/rpcserver.go#L8344) since `existingQuotes` should be `true` in my case, and when that is true then[ this hits:](https://github.com/lightninglabs/taproot-assets/blob/cf2a5e5bb47cdfb3b8a6a570113ca5923dbbf805/rpcserver.go#L8087) which means `acquiredQuotes` may not get populated and this `expensiveQuote` would be nil. With `expensiveQuote` being nil we would run into this runtime error when we call `UnmarshalRfqFixedPoint` on it inside `validateInvoiceAmount`.

### Your environment

```sh
taproot assets v0.6.0
terminal v0.15.0-alpha
LND 0.19.1-beta
```

### Steps to reproduce

Get an accepted quote with `AddAssetBuyOrder`, create an asset invoice using `tapchannelrpc.AddInvoice` , and in the `tapchannelrpc.AddInvoiceRequest` inside the `InvoiceRequest` field, include route hints  that have a hop hint that looks like:

```
invoiceHopHints := []*lnrpc.HopHint{
			{
				NodeId: myPeersPubkey,
				ChanId:                    acceptedBuyQuoteScid,
				CltvExpiryDelta:           cltvExpiry,
				FeeBaseMsat:               feeBaseMsat,
				FeeProportionalMillionths: feeProportionalMillionths,
			},
		}
```

### Expected behavior

A runtime error should not occur

### Actual behavior

Asset invoice creation yields a runtime error, the node crashes.

GeorgeTsagk:
## Description

After merging #1359 we changed some of the logic around acquiring quotes in `taprpc.AddInvoice`. A special edge case where the user has already negotiated a quote and included it in the `lnrpc.Invoice` route hints would lead into a crash as a variable would remain unset.

Closes #1622

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -8163,9 +8163,19 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 	var expensiveQuote *rfqrpc.PeerAcceptedBuyQuote
 	if !existingQuotes {
 		expensiveQuote = acquiredQuotes[0].quote
+	} else {
+		mgr := r.cfg.AuxInvoiceManager
+		buyQuote, err := mgr.GetBuyQuoteFromRouteHints(
+			iReq, specifier,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find matching buy "+
+				"quote in accepted quotes: %w", err)
+		}
+
+		expensiveQuote = marshalPeerAcceptedBuyQuote(*buyQuote)
 	}
 
-	// replace with above
 	// Now that we have the accepted quote, we know the amount in (milli)
 	// Satoshi that we need to pay. We can now update the invoice with this
 	// amount.


tapchannel/aux_invoice_manager.go
```diff
@@ -419,6 +419,34 @@ func IsAssetInvoice(invoice *lnrpc.Invoice, rfqLookup RfqLookup) bool {
...
+				continue
+			}
+
+			quoteSpecifier := buyQuote.Request.AssetSpecifier
+
+			if quoteSpecifier.String() == specifier.String() {
+				return &buyQuote
+			}
+		}
+	}
```
ffranr: Should we check that there are no other quotes here? So exhaustively check: find the quote, but then keep looking and error if we find another matching quote.

GeorgeTsagk: It's not really crucial to not have competing quotes here, as we're just picking one to create the invoice over.

For the case where the user wants to provide multiple quotes for a multi-rfq receive we're going to introduce the RPC level `rfq_ids` field, which would also eliminate the need to manually create a `[]RouteHints` argument

tapchannel/aux_invoice_manager.go
```diff
@@ -419,6 +419,34 @@ func IsAssetInvoice(invoice *lnrpc.Invoice, rfqLookup RfqLookup) bool {
...
+		for _, h := range hint.HopHints {
+			scid := h.ChanId
+			buyQuote, ok := buyQuotes[rfqmsg.SerialisedScid(scid)]
+			if !ok {
+				continue
+			}
+
+			quoteSpecifier := buyQuote.Request.AssetSpecifier
+
+			if quoteSpecifier.String() == specifier.String() {
```
guggero: I don't think we should compare specifiers using the `String()` method. This will fail for grouped assets if one has an asset ID set and the other one doesn't (which can be the case depending on user input).
So  a new `Equal(exactMatch bool)` method probably makes sense.
Where `exactMatch` requires both fields to be identical. If set to `false`, it could just be `both asset IDs set and IDs equal or both group keys set and group keys equal`.

rpcserver.go
```diff
@@ -7233,27 +7233,37 @@ func marshalPeerAcceptedBuyQuotes(
...
-		rpcQuote := &rfqrpc.PeerAcceptedBuyQuote{
-			Peer:           quote.Peer.String(),
-			Id:             quote.ID[:],
-			Scid:           uint64(scid),
-			AssetMaxAmount: quote.Request.AssetMaxAmt,
-			AskAssetRate:   rpcAskAssetRate,
-			Expiry:         uint64(quote.AssetRate.Expiry.Unix()),
-		}
-		rpcQuotes = append(rpcQuotes, rpcQuote)
+	for _, quote := range quotes {
```
guggero: nit: `rpcQuotes := fn.Map(quotes, marshalPeerAcceptBuyQuote)`.

rpcserver.go
```diff
@@ -8192,9 +8192,19 @@ func (r *rpcServer) AddInvoice(ctx context.Context,
 	var expensiveQuote *rfqrpc.PeerAcceptedBuyQuote
 	if !existingQuotes {
 		expensiveQuote = acquiredQuotes[0].quote
+	} else {
+		buyQuote := r.cfg.AuxInvoiceManager.GetBuyQuoteFromRouteHints(
+			iReq, specifier,
+		)
+
+		if buyQuote == nil {
```
guggero: I think `GetBuyQuoteFromRouteHints()` should return this error instead, makes its use more explicit.

litbot-9000:

guggero:
Pushed up the fixes I suggested.

guggero (approved):

tapchannel/aux_invoice_manager.go
```diff
@@ -419,6 +419,40 @@ func IsAssetInvoice(invoice *lnrpc.Invoice, rfqLookup RfqLookup) bool {
...
+// GetBuyQuoteFromRouteHints is a helper method that extracts a buy quote from
+// the route hints of an invoice, given that the quote is accepted and has a
+// matching specifier. If no matching quote is found, it returns an error.
+func (s *AuxInvoiceManager) GetBuyQuoteFromRouteHints(invoice *lnrpc.Invoice,
+	specifier asset.Specifier) (*rfqmsg.BuyAccept, error) {
+
+	hints := invoice.RouteHints
+	buyQuotes := s.cfg.RfqManager.PeerAcceptedBuyQuotes()
+
+	for _, hint := range hints {
```
bhandras: nit: for better readability consider the following:
```
for _, routeHint := range invoice.RouteHints {
    for _, hopHint := range routeHint.HopHints {
```

tapchannel/aux_invoice_manager.go
```diff
@@ -419,6 +419,40 @@ func IsAssetInvoice(invoice *lnrpc.Invoice, rfqLookup RfqLookup) bool {
...
+					"specifiers: %w", err)
+			}
+
+			if areEqual {
+				return &buyQuote, nil
+			}
+		}
+	}
+
+	return nil, fmt.Errorf("no buy quote found specifier %s",
```
bhandras: nit: missing "for"?

tapchannel/aux_invoice_manager.go
```diff
@@ -419,6 +419,40 @@ func IsAssetInvoice(invoice *lnrpc.Invoice, rfqLookup RfqLookup) bool {
...
+
+			quoteSpecifier := buyQuote.Request.AssetSpecifier
+			areEqual, err := quoteSpecifier.Equal(&specifier, false)
+			if err != nil {
+				return nil, fmt.Errorf("error comparing "+
+					"specifiers: %w", err)
+			}
+
+			if areEqual {
+				return &buyQuote, nil
```
bhandras, line 445: nit: not an issue right now, just flagging the pattern. We should be cautious about returning pointers to values whose lifecycle we don’t fully control (in this case, `RfqManager` returns the quotes), as that gives the receiver write access to the returned object. I realize `PeerAcceptedBuyQuotes()` currently copies the items, but it’s still a bit risky in case that behavior ever changes.

bhandras (approved):
LGTM 🎉

tapchannel/aux_invoice_manager.go
guggero: Agree, fixed.

tapchannel/aux_invoice_manager.go
guggero: Good point. But the result from `PeerAcceptedBuyQuotes` is of type `map[SerialisedScid]rfqmsg.BuyAccept`, so it doesn't house pointers. Which means `buyQuote, ok := buyQuotes[scid]` creates a copy of the struct to which the pointer is then returned? And all the elements of the `rfqmsg.BuyAccept` struct are non-pointers.
So I don't think there's anything to consider here? Or am I missing something?

bhandras: Yes like I said, this is not an issue right now, just a pattern to be carful with.

bhandras (approved):

PR #1630, opened 2025-07-01
tapdb: sanitize db errors before returning

Roasbeef:
In this PR, we start to sanitize unexpected DB errors before returning to the caller. This is important, as ultimately for our gRPC calls, the caller is some client-side software. For now, we take the simple approach and match on some obvious strings, mapping them to a generic "db connection failed" error.

### DIFF
--- a/tapdb/postgres_test.go
+++ b/tapdb/postgres_test.go
@@ -1,0 +1,47 @@
+//go:build test_db_postgres
+
+package tapdb
+
+import (
+	"context"
+	"database/sql"
+	"testing"
+
+	"github.com/lightningnetwork/lnd/macaroons"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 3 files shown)


Roasbeef:
Thanks for that added test case! PTAL.

tapdb/postgres_test.go
```diff
@@ -0,0 +1,51 @@
...
+	require.NoError(t, err)
+
+	rksDB := NewTransactionExecutor(store, func(tx *sql.Tx) KeyStore {
+		return store.WithTx(tx)
+	})
+	rks := NewRootKeyStore(rksDB)
+
+	// Now before we attempt a query, we sneakily shut down the Postgres
+	// fixture, resulting in a connection error.
+	// sqlFixture.TearDown(t)
```
guggero: Oops, this was left in my diff from another attempt...

guggero (approved):
LGTM 🎉

Issue #1631, opened 2025-07-02
QuerySyncStats purges cache instead of refreshing in background


Roasbeef:
## Problem

The `QuerySyncStats` method in `tapdb/universe_stats.go:930-936` currently purges the entire sync stats cache after the cache duration expires. This leads to cold cache hits for subsequent requests, which can take a significant time to load.

## Current Behavior

When the cache duration expires, the code simply wipes the entire cache:
```go
u.syncStatsRefresh = time.AfterFunc(u.opts.cacheDuration, func() {
    log.Infof("Purging sync stats cache, duration=%v",
        u.opts.cacheDuration)
    
    u.syncStatsCache.wipe()
})
```

## Expected Behavior

Instead of purging the cache entirely, the system should:
1. Load fresh data in the background
2. Continue serving the existing cached data while the refresh is happening
3. Atomically swap to the new data once the background refresh is complete

This would prevent cold cache hits and improve response times for users.

## Impact

Users experience slow response times after cache expiration, as the entire dataset needs to be reloaded from scratch on the next request.

guggero:
Even though this behavior is still in place, I'm not sure we need to change it after https://github.com/lightninglabs/taproot-assets/pull/1650.
The query is now relatively fast and re-populating the cache only takes a minute or two. So perhaps we can close this?

levmi:
Given the above comment from Oli, I guess we can close this issue.

PR #1632, opened 2025-07-02
tapdb: don't wipe sync stat cache, refresh in background

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1631.

Opening as draft, we need unit and integration tests to make sure nothing breaks.


guggero:
Turns out the root cause isn't the stats query itself but the loop that marshals the assets which does a DB lookup and chain backend call per asset (for which there are ~100k on mainnet). So this will be fixed by https://github.com/lightninglabs/taproot-assets/pull/1650 instead.

Closing this one as not needed.

PR #1639, opened 2025-07-04
tapchannel: add a one unit extra tolerance to invoice acceptor

GeorgeTsagk:
## Description

When converting an mSAT amount to units and vice versa we frequently get an off-by-1-unit result. This is already accounted for in our invoice acceptor by adding the value of as many asset units as HTLCs in the invoice. What we didn't account for is that the UnitsToMilliSatoshi calculation for the margin can also be off by 1 msat.

So the result of `UnitsToMilliSatoshi` for the `totalUnits - 1` can itself be off by 1mSAT, compared to the result of `UnitsToMilliSatoshi` for a single unit

### DIFF
--- a/tapchannel/aux_invoice_manager.go
+++ b/tapchannel/aux_invoice_manager.go
@@ -264,8 +264,19 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(ctx context.Context,
 
 	// We assume that each shard can have a rounding error of up to 1 asset
 	// unit. So we allow the final amount to be off by up to 1 asset unit
-	// per accepted HTLC (plus the one we're currently processing).
+	// per accepted HTLC (plus the one we're currently processing). The
+	// plus one here is because the invoice only has previously accepted
+	// HTLCs committed to it, but we're processing the current HTLC which
+	// is not yet in that list.
 	allowedMarginAssetUnits := uint64(len(req.Invoice.Htlcs) + 1)
+
+	// Because we do a round trip of units -> milli-sats, then split into
+	// shards, then back to units and then back to milli-sats again, we lose
+	// up to numHTLCs+1 asset units in the process. So we allow for an
+	// additional unit here.
+	allowedMarginAssetUnits++
+
+	// Convert the allowed margin asset units to milli-satoshis.
 	marginAssetUnits := rfqmath.NewBigIntFixedPoint(
 		allowedMarginAssetUnits, 0,
 	)
--- a/tapchannel/aux_traffic_shaper_test.go
+++ b/tapchannel/aux_traffic_shaper_test.go
@@ -1,0 +1,114 @@
+package tapchannel
+
+import (
+	"testing"
+
+	"github.com/lightninglabs/taproot-assets/rfqmath"
+	"github.com/lightningnetwork/lnd/lnwire"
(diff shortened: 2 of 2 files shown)


litbot-9000:
@ffranr: review reminder

tapchannel/aux_invoice_manager.go
```diff
@@ -273,6 +273,10 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(ctx context.Context,
 		marginAssetUnits, *assetRate,
 	)
 
+	// We add a 1mSAT per HTLC margin which is the inprecision tolerance of
```
bhandras: nit: inprecision => imprecision

tapchannel/aux_invoice_manager.go
```diff
@@ -273,6 +273,10 @@ func (s *AuxInvoiceManager) handleInvoiceAccept(ctx context.Context,
...
+	// our rfqmath library.
+	allowedMarginMSat += lnwire.MilliSatoshi(allowedMarginAssetUnits)
```
bhandras: Is this the result of `UnitsToMilliSatoshi` flooring numbers less than 1? In essence should we fix it there (by adding 0.5 always)?

guggero: Yes, this is the actual cause of the issue. But we can't just change this behavior without potentially breaking older nodes in the process. So for now we just add some additional slack here (which is receiver-side only, doesn't affect routing nodes), then at a later date we decide on the exact rounding behavior we want and roll it out in a (version-triggered) safe manner.

guggero:
I've added a unit test that demonstrates the behavior in its output:

```
=== RUN   TestUnitConversionTolerance
    aux_traffic_shaper_test.go:27: Initial amount in asset units: 6000000
    aux_traffic_shaper_test.go:32: Invoice amount in msat: 60900000
    aux_traffic_shaper_test.go:38: Calculated msat per asset unit: 10.15
    aux_traffic_shaper_test.go:40: Number of asset units after round trip: 5999999
    aux_traffic_shaper_test.go:46: Sum of 2 shards in asset units: 5999998
    aux_traffic_shaper_test.go:52: Inbound amount in msat: 60899979
    aux_traffic_shaper_test.go:53: Total tolerance required in msat: 21
    aux_traffic_shaper_test.go:64: Old tolerance allowed in msat: 20
    aux_traffic_shaper_test.go:65: New tolerance allowed in msat: 30
--- PASS: TestUnitConversionTolerance (0.00s)
PASS
```

The previous code from @GeorgeTsagk would've allowed for a total of 40 msat of slack (double the old tolerance). But I was able to optimize it to `numHTLCs+1` which is 30 msat in this case. The accompanying rapid test makes sure we don't run into a lack of tolerance again.

GeorgeTsagk:
> The previous code from @GeorgeTsagk would've allowed for a total of 40 msat of slack (double the old tolerance). But I was able to optimize it to numHTLCs+1 which is 30 msat in this case. The accompanying rapid test makes sure we don't run into a lack of tolerance again.

in your example shouldn't the previous approach produce a margin of `old + num_shards * 1msat`

so for
- old tolerance = 20 msat
- num_shards = 2

that would be 22 msat?

PR #1643, opened 2025-07-09
supplycommit: add MultiStateMachineManager for managing group state machines

ffranr:
Part of the work towards completing https://github.com/lightninglabs/taproot-assets/issues/1534

These commits are split out from https://github.com/lightninglabs/taproot-assets/pull/1554

---

This PR introduces the `MultiStateMachineManager`, a service that manages supply commitment state machines for multiple asset groups. Each state machine is scoped to a specific asset group. The `MultiStateMachineManager` caches and orchestrates these group-specific machines, providing a unified interface for the RPC server to start, query, and stop them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1643)

### DIFF
--- a/universe/supplycommit/multi_sm_manager.go
+++ b/universe/supplycommit/multi_sm_manager.go
@@ -160,8 +160,15 @@ func (m *MultiStateMachineManager) fetchStateMachine(
 		ChainParams:      m.cfg.ChainParams,
 	}
 
-	// TODO(ffranr): Get initial state from disk here.
-	initialState := &DefaultState{}
+	// Before we start the state machine, we'll need to fetch the current
+	// state from disk, to see if we need to emit any new events.
+	ctx, cancel := m.WithCtxQuitNoTimeout()
+	defer cancel()
+
+	initialState, _, err := m.cfg.StateLog.FetchState(ctx, assetSpec)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch current state: %w", err)
+	}
 
 	// Create a new error reporter for the state machine.
 	errorReporter := NewErrorReporter(assetSpec)
@@ -179,6 +186,23 @@ func (m *MultiStateMachineManager) fetchStateMachine(
 	// the manager is stopped.
 	smCtx, _ := m.WithCtxQuitNoTimeout()
 	newSm.Start(smCtx)
+
+	// If specific initial states are provided, we send the corresponding
+	// events to the state machine to ensure it begins ticking as expected.
+	switch initialState.(type) {
+	// Once we write the commitment transaction to disk in CommitTxSign,
+	// then on restart, we'll be in the broadcast state. From this point,
+	// we'll trigger the broadcast event so we can resume the state machine.
+	case *CommitBroadcastState:
+		newSm.SendEvent(ctx, &BroadcastEvent{})
+
+	// Once we get a confirmation, then we'll transition to the
+	// CommitFinalizeState. If we crashed right after that, then
+	// we'll also send the finalize event so we can apply
+	// everything, and transition back to the normal default state.
+	case *CommitFinalizeState:
+		newSm.SendEvent(ctx, &FinalizeEvent{})
+	}
 
 	m.smCache.Set(*groupKey, &newSm)
 
--- a/universe/supplycommit/transitions.go
+++ b/universe/supplycommit/transitions.go
@@ -630,9 +630,7 @@ func (c *CommitTxSignState) ProcessEvent(event Event,
 				SupplyTransition: stateTransition,
 			},
 			NewEvents: lfn.Some(FsmEvent{
-				InternalEvent: []Event{&BroadcastEvent{
-					SignedCommitPkt: signedPsbt,
-				}},
+				InternalEvent: []Event{&BroadcastEvent{}},
 			}),
 		}, nil
 
@@ -654,13 +652,11 @@ func (c *CommitBroadcastState) ProcessEvent(event Event,
 	// signed commit tx, then register for a confirmation for when it
 	// confirms.
 	case *BroadcastEvent:
-		// First, we'll extract the final signed transaction from the
-		// PSBT.
-		commitTx, err := psbt.Extract(newEvent.SignedCommitPkt)
-		if err != nil {
-			return nil, fmt.Errorf("unable to extract "+
-				"psbt: %w", err)
+		if c.SupplyTransition.NewCommitment.Txn == nil {
+			return nil, fmt.Errorf("commitment transaction is nil")
 		}
+
+		commitTx := c.SupplyTransition.NewCommitment.Txn
 
 		// Construct a detailed label for the broadcast request using
 		// the helper function.


ffranr:
@guggero have i defined `LndFsmDaemonAdapters.RegisterConfirmationsNtfn` reasonably here? I think the linter will complain.

wallet_anchor.go
```diff
@@ -231,6 +233,23 @@ func (l *LndRpcWalletAnchor) MinRelayFee(
 	return l.lnd.WalletKit.MinRelayFee(ctx)
 }
 
+// DeriveNextKey derives the next key in the taproot assets key family.
+func (l *LndRpcWalletAnchor) DeriveNextKey(
```
Roasbeef: Interesting, I thought we already had this in place...

universe/supplycommit/multi_sm_manager.go
```diff
@@ -0,0 +1,394 @@
...
+	"github.com/lightningnetwork/lnd/msgmux"
+	"github.com/lightningnetwork/lnd/protofsm"
+)
+
+const (
+	// DefaultTimeout is the context guard default timeout.
+	DefaultTimeout = 30 * time.Second
+)
+
+type MultiStateMachineManagerCfg struct {
```
Roasbeef, line 39: Missing `godoc` comment.

universe/supplycommit/multi_sm_manager.go
```diff
@@ -0,0 +1,394 @@
...
+	return fn.Some(FetchCommitmentResp{
+		SupplyTree:      supplyTree,
+		Subtrees:        subtrees,
+		ChainCommitment: chainCommit,
+	}), nil
+}
+
+// StateMachineCache is a thread-safe cache mapping an asset group's public key
+// to its supply commitment state machine.
+type StateMachineCache struct {
```
Roasbeef: Does this need to be public exported?

universe/supplycommit/multi_sm_manager.go
```diff
@@ -0,0 +1,394 @@
...
+		AssetSpec:        assetSpec,
+		TreeView:         m.cfg.TreeView,
+		Commitments:      m.cfg.Commitments,
+		Wallet:           m.cfg.Wallet,
+		Chain:            m.cfg.Chain,
+		StateLog:         m.cfg.StateLog,
+		CommitConfTarget: DefaultCommitConfTarget,
+	}
+
+	// TODO(ffranr): Get initial state from disk here.
```
Roasbeef: IIRC I gave a commit to do this in the original PR before we started to break it up into sub-PRs.

Roasbeef: Found it!

Here's the commit: https://github.com/lightninglabs/taproot-assets/commit/d033b3c8d6da5c175616aaeb696a6746d52f7d01

universe/supplycommit/multi_sm_manager.go
```diff
@@ -0,0 +1,394 @@
...
+		return zero, fmt.Errorf("unable to fetch supply commit: %w",
+			err)
+	}
+
+	if chainCommitOpt.IsNone() {
+		// If the chain commitment is not present, we return an empty
+		// response.
+		return zero, nil
+	}
+	chainCommit, err := chainCommitOpt.UnwrapOrErr(fmt.Errorf("unable to "+
```
Roasbeef: Style nit: line folding.

server.go
```diff
@@ -210,6 +210,9 @@ func (s *Server) initialize(interceptorChain *rpcperms.InterceptorChain) error {
 		return fmt.Errorf("unable to start RFQ manager: %w", err)
 	}
 
+	// Start universe supply commitment manager.
+	s.cfg.SupplyCommitManager.Start()
```
Roasbeef: Any need to also stop it?

chain_bridge.go
```diff
@@ -458,3 +459,153 @@ func (l *LndInvoicesClient) HtlcModifier(ctx context.Context,
 // Ensure LndInvoicesClient implements the tapchannel.InvoiceHtlcModifier
 // interface.
 var _ tapchannel.InvoiceHtlcModifier = (*LndInvoicesClient)(nil)
+
+// LndFsmDaemonAdapters is a struct that implements the protofsm.DaemonAdapters
+// interface.
+type LndFsmDaemonAdapters struct {
```
Roasbeef, line 471: 👍

Roasbeef (changes requested):
Straight forward diff!

No major comments, nicely done.

universe/supplycommit/multi_sm_manager.go
ffranr: It does not. I will make it private.

universe/supplycommit/multi_sm_manager.go
ffranr: Nice, thanks! I've added that commit to this PR.

wallet_anchor.go
guggero: What about `key_ring.go`?

universe/supplycommit/multi_sm_manager.go
```diff
@@ -0,0 +1,402 @@
...
+	mu sync.RWMutex
+
+	// cache is a map of serialized asset group public keys to their
+	// supply commitment state machines.
+	cache map[asset.SerializedKey]*StateMachine
+}
+
+// newStateMachineCache creates a new supply commit state machine cache.
+func newStateMachineCache() stateMachineCache {
+	return stateMachineCache{
```
guggero: I'm surprised the linter didn't complain about this. But I think anything that contains a `sync.RWMutex` (or the plain one) should not be copied for safety. So this should return a pointer and it should be kept as a pointer in the parent struct.

universe/supplycommit/multi_sm_manager.go
```diff
@@ -87,25 +103,50 @@ func NewMultiStateMachineManager(
 
 // Start starts the multi state machine manager.
 func (m *MultiStateMachineManager) Start() error {
+	var err error
+
 	m.startOnce.Do(func() {
+		err = m.cfg.DaemonAdapters.Start()
```
guggero: Hmm, is it really the responsibility of the multi state machine manager to start something that it gets passed as a config item? I would expect this to be handled by the entity that instantiates and starts the multi state machine manager (so probably the main server?).

universe/supplycommit/multi_sm_manager.go
```diff
@@ -87,25 +103,50 @@ func NewMultiStateMachineManager(
 
 // Start starts the multi state machine manager.
 func (m *MultiStateMachineManager) Start() error {
+	var err error
```
guggero: nit: we usually call these `startErr` and `stopErr` respectively, to make the pattern more obvious.
So would be
```
		if err != nil {
			startErr = fmt.Errorf("unable to start daemon adapters: %w",
				err)
			return
		}
```
below.

lnd_services.go
guggero, line 1: Should we also merge the `psbt_channel_funder.go`, `key_ring.go` and `wallet_anchor.go` files into this one as well, since otherwise it's easy to miss what `lndclient` wrappers we've already added? Will be a big file, but at least everything is in one place.

universe/supplycommit/multi_sm_manager.go
```diff
@@ -182,8 +182,32 @@ func (m *MultiStateMachineManager) fetchStateMachine(
...
+	}
+
+	var triggerEvent fn.Option[Event]
+
+	switch initialState.(type) {
+	// Once we write the commitment transaction to disk in CommitTxSign,
+	// then on restart, we'll be in the broadcast state. From this point,
+	// we'll trigger the broadcast event so we can resume the state machine.
+	case *CommitBroadcastState:
+		triggerEvent = fn.Some[Event](&BroadcastEvent{})
```
guggero: Why don't we move this below the initialization of the state machine and send the event directly? No need for the option value here...

guggero (commented):
Very clean! Got a couple of suggestions, nothing too involved I think.

universe/supplycommit/multi_sm_manager.go
Roasbeef: That works too, it can either send directly in the switch, or use a similar temp var pattern.

wallet_anchor.go
ffranr: goal of this commit is to satisfy the `supplycommit.Wallet` interface. `LndRpcWalletAnchor` is nearly there where as `LndRpcKeyRing` is missing many methods. Did I misunderstand your reference to `key_ring.go`? Did you mean something else?

universe/supplycommit/multi_sm_manager.go
ffranr: From the doc for `sync.RWMutex`:
```
A RWMutex must not be copied after first use.
```

And from AI:
```
Linters such as go vet -copylocks or Staticcheck’s SA2002 therefore fire only when they can see the copy happens after first use.

In your constructor the mutex is still at its zero value
```

so we can copy in the constructor since it's not used yet. But I will go with pointers since that's what you're used to seeing.

wallet_anchor.go
guggero: Ah, I see. Perhaps in that case we should split the `supplycommit.Wallet` into `supplycommit.Wallet` and `supplycommit.KeyRing` to avoid this duplication? Or merge the `NewLndRpcKeyRing` and `LndRpcWalletAnchor` wallet types into a single type that implements both?

lnd_services.go
ffranr: Yes, we should refactor/cleanup this code. I will put up a separate PR for that.

universe/supplycommit/multi_sm_manager.go
ffranr: I don't see any benefit in using `startErr` and `stopErr` as error variable names in `Start()` and `Stop()`. The standard `err` seems sufficient. That said, I see we use these names elsewhere in the code as well, though it's not clear why to me in those cases either.
 
 I will modify this PR to use `startErr` and `stopErr`.

universe/supplycommit/multi_sm_manager.go
ffranr: In latest commit the daemon adapters component is started by the main server. It can be re-used by other state machines.

wallet_anchor.go
ffranr: Good idea. I've added `supplycommit.KeyRing`.

Roasbeef:
Can land after a rebase!

PR #1650, opened 2025-07-14
universe: fix asset stats being extremely slow

guggero:
It turns out it's not actually the stats query itself that is slow but the two pieces of information we need to fetch for each asset returned in the result:
 - The genesis timestamp: Needs to be fetched by looking up a block in the chain backend. Is already cached, but we increase the cache. https://github.com/lightninglabs/taproot-assets/blob/eff99b84184dbe72dccdbcafb6915e07e76572f2/rpcserver.go#L6545
 - The decimal display value: We used to do a DB lookup for each asset (which takes forever with a cardinality of ~100k on mainnet). We now cache this. https://github.com/lightninglabs/taproot-assets/blob/eff99b84184dbe72dccdbcafb6915e07e76572f2/rpcserver.go#L6551

### DIFF
--- a/address/book.go
+++ b/address/book.go
@@ -150,6 +150,11 @@ type Storage interface {
 	// transfer comes in later on.
 	InsertScriptKey(ctx context.Context, scriptKey asset.ScriptKey,
 		keyType asset.ScriptKeyType) error
+
+	// FetchAllAssetMeta attempts to fetch all asset meta known to the
+	// database.
+	FetchAllAssetMeta(
+		ctx context.Context) (map[asset.ID]*proof.MetaReveal, error)
 }
 
 // KeyRing is used to create script and internal keys for Taproot Asset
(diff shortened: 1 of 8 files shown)


address/book.go
```diff
@@ -423,6 +423,20 @@ func (b *Book) FetchAssetMetaForAsset(ctx context.Context,
 	return meta, nil
 }
 
+// DecDisplayForAssetID attempts to fetch the meta reveal for a specific asset
+// ID and extract the decimal display value from it.
+func (b *Book) DecDisplayForAssetID(ctx context.Context,
```
Roasbeef, line 442: 👍

Roasbeef (approved):
LGTM 🪡

We have exported gRPC states for each endpoint, so we should be able to consult that monitoring to lift concrete numbers for the speed up here.

bhandras (approved):

PR #1658, opened 2025-07-18
[group key addrs 6/7]: bug fixes and refactors

guggero:
Extracted some commits from https://github.com/lightninglabs/taproot-assets/pull/1587 as suggested.
Addressed comments on existing review in https://github.com/lightninglabs/taproot-assets/pull/1587 that were meant for these commits.

### DIFF
--- a/taprpc/authmailboxrpc/mailbox.proto
+++ b/taprpc/authmailboxrpc/mailbox.proto
@@ -165,9 +165,33 @@ message ReceiveMessagesRequest {
 
 // Carries the initial parameters from the client to start receiving messages.
 message InitReceive {
+    // The public key identifier of the receiver of the messages that should be
+    // received through a message receive stream, encoded as the raw bytes of
+    // the compressed public key.
     bytes receiver_id = 1;
+
+    // The exclusive start message ID, meaning messages with this ID or higher
+    // will be included in the response. This allows the client to resume
+    // receiving messages from a specific point without missing any. One of
+    // start_message_id_exclusive, start_block_height_inclusive or
+    // start_timestamp_exclusive must be at their non-default values for any
+    // existing messages to be returned!
     uint64 start_message_id_exclusive = 2;
+
+    // The inclusive start block height, meaning messages from this block height
+    // or higher will be included in the response. This allows the client to
+    // filter messages based on the block height at which they were produced.
+    // One of start_message_id_exclusive, start_block_height_inclusive or
+    // start_timestamp_exclusive must be at their non-default values for any
+    // existing messages to be returned!
     uint32 start_block_height_inclusive = 3;
+
+    // The exclusive start timestamp in Unix epoch seconds, meaning messages
+    // with a timestamp strictly greater than this value will be included in the
+    // response. This allows the client to filter messages based on their
+    // arrival time at the server. One of start_message_id_exclusive,
+    // start_block_height_inclusive or start_timestamp_exclusive must be at
+    // their non-default values for any existing messages to be returned!
     int64 start_timestamp_exclusive = 4;
 }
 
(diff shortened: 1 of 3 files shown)


authmailbox/server.go
```diff
@@ -222,28 +218,45 @@ func (s *Server) SendMessage(ctx context.Context,
...
+			// algorithm will produce a different ciphertext for the
+			// same message each time, we cannot compare the actual
+			// message itself. So we have to assume that using the
+			// same outpoint in the proof for the same recipient
+			// means it's also the same message.
+			if !existingMsg.ReceiverKey.IsEqual(&msg.ReceiverKey) {
+				// It's a different recipient, so someone is
+				// attempting to re-use a proof for a different
+				// recipient.
+				return nil, proof.ErrTxMerkleProofExists
```
ffranr: returned error type here doesn't make sense to me. I would expect an error type along the lines of: attempting to reuse outpoint for a different receiver.

tapdb/sqlc/queries/addrs.sql
```diff
@@ -174,3 +174,12 @@ WHERE addr_events.status >= @status_from
   AND COALESCE(@addr_taproot_key, addrs.taproot_output_key) = addrs.taproot_output_key
   AND addr_events.creation_time >= @created_after
 ORDER by addr_events.creation_time;
+
+-- name: QueryLastEventHeightByAddrVersion :one
```
ffranr: I think we should just call this `QueryLastEventHeight` and the addr version could be thought of as an optional argument. But not a big deal.

taprpc/authmailboxrpc/mailbox.proto
```diff
@@ -166,8 +166,29 @@ message ReceiveMessagesRequest {
 // Carries the initial parameters from the client to start receiving messages.
 message InitReceive {
     bytes receiver_id = 1;
+
+    // The exclusive start message ID, meaning messages with this ID or higher
+    // will be included in the response. This allows the client to resume
+    // receiving messages from a specific point without missing any. One of
+    // start_message_id_exclusive, start_block_height_inclusive or
```
ffranr, line 176: Would be good to add doc comment for `bytes receiver_id = 1;` above also

authmailbox/server.go
guggero: Two reasons for this: 1. Since this is potentially someone trying to re-use a proof and spam the server, I wanted to not give them too much information. 2. Be able to easily detect the error in unit tests, hence the constant.
But wrapped the error and added a little bit more context: `outpoint in proof alread in use: %w`.

tapdb/assets_store.go
```diff
@@ -3167,17 +3167,22 @@ func (a *AssetStore) LogAnchorTxConfirm(ctx context.Context,
 			isBurn := !isNumsKey && len(witnessData) > 0 &&
 				asset.IsBurnKey(scriptPubKey, witnessData[0])
 			isKnown := fullScriptKey.Type != asset.ScriptKeyUnknown
+			isRemotePedersen := fullScriptKey.Type ==
+				asset.ScriptKeyUniquePedersen &&
+				!out.ScriptKeyLocal
 			skipAssetCreation := !isTombstone && !isBurn &&
-				!out.ScriptKeyLocal && !isKnown
+				!out.ScriptKeyLocal &&
+				(!isKnown || isRemotePedersen)
```
Roasbeef: Could use a comment here re the diff re handling pederson types vs all the rest.

commitment/tap.go
```diff
@@ -457,8 +457,7 @@ func IsTaprootAssetCommitmentScript(script []byte) bool {
 func (c *TapCommitment) TapscriptRoot(sibling *chainhash.Hash) chainhash.Hash {
 	commitmentLeaf := c.TapLeaf()
 	if sibling == nil {
-		return txscript.AssembleTaprootScriptTree(commitmentLeaf).
-			RootNode.TapHash()
```
Roasbeef, line 461: 👍

Roasbeef (approved):
LGTM 🍽️

tapdb/assets_store.go
guggero: Yeah, good point. I did that and also extracted the logic into a standalone function and added unit tests for it: 20fa3815

PR #1669, opened 2025-07-21
Use nonce as HKDF salt and protocol name as info in ECIES encryption

ffranr:
This PR updates the HKDF parameters used in ECIES encryption to improve key derivation security:

- The per-message nonce is now used as the `salt` input to `HkdfSha256`. This improves brute-force resistance by introducing entropy that changes with each encryption.
- The protocol name, previously used as a constant salt, is now passed as the `info` parameter. This better aligns with the intended purpose of `info`, which is to provide domain/context separation.

Using a random, per-message salt ensures that the derived keys are unique even if the input key material is reused or predictable, and prevents pre-computation attacks. The change follows HKDF best practices.

### DIFF
--- a/internal/ecies/ecies.go
+++ b/internal/ecies/ecies.go
@@ -79,8 +79,9 @@ func EncryptSha256ChaCha20Poly1305(sharedSecret [32]byte, msg []byte,
 		return nil, fmt.Errorf("cannot read random nonce: %w", err)
 	}
 
-	// We begin by hardening the shared secret against brute forcing by
-	// using HKDF with SHA256.
+	// Derive a strong session key from the shared secret using HKDF-SHA256.
+	// The nonce is used as the salt, and the protocol name as the info
+	// label. This mitigates risks from weak shared secrets.
 	stretchedKey, err := HkdfSha256(
 		sharedSecret[:], nonce, []byte(protocolName),
 	)
@@ -197,8 +198,9 @@ func DecryptSha256ChaCha20Poly1305(sharedSecret [32]byte,
 	nonce := remainder[:nonceSize]
 	ciphertext := remainder[nonceSize:]
 
-	// We begin by hardening the shared secret against brute forcing by
-	// using HKDF with SHA256.
+	// Derive a strong session key from the shared secret using HKDF-SHA256.
+	// The nonce is used as the salt, and the protocol name as the info
+	// label. This mitigates risks from weak shared secrets.
 	stretchedKey, err := HkdfSha256(
 		sharedSecret[:], nonce, []byte(protocolName),
 	)
--- a/internal/ecies/ecies.go
+++ b/internal/ecies/ecies.go
@@ -71,6 +71,14 @@ func EncryptSha256ChaCha20Poly1305(sharedSecret [32]byte, msg []byte,
 			"given, 255 bytes maximum", len(additionalData))
 	}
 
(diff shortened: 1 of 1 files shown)


ffranr:
I don't think we've released since adding ECIES package, so this change shouldn't require version bump.

internal/ecies/ecies.go
```diff
@@ -81,7 +81,9 @@ func EncryptSha256ChaCha20Poly1305(sharedSecret [32]byte, msg []byte,
 
 	// We begin by hardening the shared secret against brute forcing by
 	// using HKDF with SHA256.
-	stretchedKey, err := HkdfSha256(sharedSecret[:], []byte(protocolName))
+	stretchedKey, err := HkdfSha256(
```
Roasbeef, line 85: What type of brute force resistance are we protecting against here? The shared secret is a 256-bit value derived from ECDH. A brute force attempt with or without the nonce would reduce down to either violating the pre-image resistance property of the hash function, or randomly finding a shared secret input that maps to the same key. A pre-image attack would take an expected 2^256 invocations, while a collision by the birthday bound would take 2^128 work, which is the same amount of bits of security as widely used params for ECC. 

Also note that the nonce here is public information for an attacker, as it's in plaintext alongside the final ciphertext. So if they're trying to brute force, they already have this nonce value. 

IMO, the only relevant change here is passing in the protocol name as info instead of salt below. Salt is used in the extract phase, while info is used in the expand phase.

ffranr: Yeah, I think you're right, the brute-force argument is pretty weak here.

Perhaps a stronger justification is that this might gives us some defense in depth. If the sender's ephemeral key is ever reused (intentionally or due to a bug), the shared secret might repeat. Using a per-message nonce as the HKDF salt ensures the derived key is still unique (hardening). That way, `EncryptSha256ChaCha20Poly1305` doesn't have to assume the given shared secret is solid. 

Perhaps we can think of it as cheap insurance? Stronger separation between messages, avoids any linkage in case of shared secret reuse, and reduces reliance on the shared secret always being perfect.

internal/ecies/ecies.go
```diff
@@ -192,9 +197,10 @@ func DecryptSha256ChaCha20Poly1305(sharedSecret [32]byte,
 			"cipher: %w", err)
 	}
 
-	// Split additional data, nonce and ciphertext.
-	nonce := remainder[:aead.NonceSize()]
-	ciphertext := remainder[aead.NonceSize():]
+	// Sanity check the nonce size used.
+	if len(nonce) != aead.NonceSize() {
```
guggero, line 220: Should we use the same `nonceSize` variable as above to make things more explicit? Same in next commit.

ffranr: I'm explicitly using `aead` here to validate that the actual `nonce` matches the expected size defined by `xchacha20poly1305`, rather than relying on `nonceSize := chacha20poly1305.NonceSizeX`.

litbot-9000:
@gijswijs: review reminder
@roasbeef: review reminder

ffranr:
Added `no-changelog` label because this code shouldn't be in use yet.

I've added a separate commit that updates the HKDF-related comments. It removes the misleading reference to brute force resistance and clarifies that HKDF helps mitigate issues arising from weak shared secrets.

PR #1671, opened 2025-07-22
ci: fix litd itest

guggero:
Attempts to fix the litd itests after adding a new gRPC server that needs to be exposed to be used in `litd`.
This PR works in conjunction with this commit in `litd`: https://github.com/lightninglabs/lightning-terminal/commit/f4b07907

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -21,7 +21,7 @@ env:
 
   GO_VERSION: '1.23.9'
   
-  LITD_ITEST_BRANCH: 'lnd-19-2'
+  LITD_ITEST_BRANCH: 'tapd-main-branch'
 
 jobs:
   #######################


authmailbox/server.go
```diff
@@ -95,6 +94,12 @@ func NewServer(cfg *ServerConfig) *Server {
 	}
 }
 
+// UpdateConfig sets the configuration for the mailbox server. This should be
+// called before starting the server.
+func (s *Server) UpdateConfig(cfg *ServerConfig) {
+	s.cfg = cfg
+}
+
```
ffranr: I don't understand why we need to take this approach. Could you say more on this please?

But if we must, perhaps we should add a check to `Start` to error early if the config isn't set.

guggero: As mentioned in the commit message, we need to register the struct that implements the gRPC server with the gRPC registry early in the `litd` startup process. So we need to be able to create the `Server` before we have everything available that is needed to build the configuration (`lndclient` stuff, interceptor and so on).

guggero: But the check in `Start` is a good idea, I'll add that.

ffranr: Another possibility perhaps is to pass the config into `Start`.

Because I don't think we should expect `UpdateConfig` to be used once the service is running, right? In fact, we might want to disallow calling `UpdateConfig` once the service is running.

But I would be more in favour of just passing the config into `Start` if possible. Makes things clear.

guggero: Okay, I did that for the RPC and authmailbox server. For the main server, I really don't want to break anything in the call stack from `litd`, so I kept things as they were.

Issue #1670, opened 2025-07-21
[Test Flake] TestAuxInvoiceManagerProperty: unexpected final asset value


Roasbeef:
## Description

The property-based test `TestAuxInvoiceManagerProperty` is flaking in CI with an unexpected final asset value.

## Error Details

```
--- FAIL: TestAuxInvoiceManagerProperty (5.33s)
    --- FAIL: TestAuxInvoiceManagerProperty/invoice_manager (5.33s)
        aux_invoice_manager_test.go:1040: [rapid] failed after 88 tests: unexpected final asset value, wanted 0, got 2
            To reproduce, specify -run="TestAuxInvoiceManagerProperty/invoice_manager" -rapid.failfile="testdata/rapid/TestAuxInvoiceManagerProperty_invoice_manager/TestAuxInvoiceManagerProperty_invoice_manager-20250721212415-40486.fail" (or -rapid.seed=14792004971574732895)
```

## Test Parameters (from rapid draw)

- **requestsLen**: 1
- **chan_id**: 0x0
- **rfq_id**: [32]uint8{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}
- **inv_exists**: true
- **invoice_value_msat**: 2
- **route_hints_len**: 1
- **hop_hints_len**: 1
- **hop_hint_bad_scid**: false
- **incorrect_peer**: false
- **no_wire_custom_records**: false
- **asset_units**: 0x1
- **asset_id**: 0x0
- **has_rfqid**: false
- **rfqid_match**: false
- **nodeID_mismatch**: false
- **scid_mismatch**: false
- **no_asset_value_match**: true
- **asset_msat_value**: 100000000001

## Reproduction

To reproduce locally, run:
```bash
go test -v -run="TestAuxInvoiceManagerProperty/invoice_manager" -rapid.seed=14792004971574732895
```

Or with the failfile:
```bash
[...]

PR #1672, opened 2025-07-22
Fix AuxInvoiceManager prop tests

GeorgeTsagk:
## Description

On the top level, this PR fixes the flake reported in https://github.com/lightninglabs/taproot-assets/issues/1670.

The actual moment in our development history where this flake could have been discovered was here https://github.com/lightninglabs/taproot-assets/pull/1478

By merging the above PR, we added an extra check to the HTLC validation by the aux invoice manager. That check compared the assets of the HTLC against the assets of the channel. We would draw the channel ID from `rapid.Uint64`, but due to the nature of rapid draws this would cause conflicts in the channel ID. That lead to failures when checking the HTLC against the channel.

> In detail, the channel asset info would source from `lndclient.ListChannels` which is mocked with the rapid-generated channels. After retrieving the responses we would check if the chanIDs match, which means that in many cases we could grab the wrong channel entry.

#### Why this is a flake

The current design of the aux invoice manager prop test allows rapid to draw many values randomly. Many of these values are flags that indicate whether a certain type of error should be caused. Due to the large number of draws related to "things that can go wrong" we would rarely run into the case where "nothing is going wrong, except for a chan ID collision".

> The above is true at least for our current default number of checks, which is 100. Even without the above fix (and without any rapid failure files present) you can run the tests with `-rapid.checks=100` a couple dozens of times before this error pops up. This allowed for the error introduction to be buried sufficiently deep for it to be ignored as a flake onwards.

With the assumption that we don't really want to account for non-sane cases where we have colliding channel IDs, we change the draw to be done from `rand.Uint64` instead. This way intentional collisions, which are not relevant, are avoided.

Closes https://github.com/lightninglabs/taproot-assets/issues/1670

> Given that this PR changes the set of drawn values, the seed referenced in the above issue becomes irrelevant. We can still draw and ignore the old value just to preserve the previously failing test case.

### DIFF
--- a/tapchannel/aux_invoice_manager_test.go
+++ b/tapchannel/aux_invoice_manager_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/big"
+	"math/rand/v2"
 	"testing"
 	"time"
 
@@ -820,11 +821,10 @@ func genHtlc(t *rapid.T, balance []*rfqmsg.AssetBalance,
 // method also returns the assetUnits and the rfqID used by the htlc.
 func genRequest(t *rapid.T) (lndclient.InvoiceHtlcModifyRequest, uint64,
 	asset.ID, rfqmsg.ID) {
-
 	request := lndclient.InvoiceHtlcModifyRequest{
 		CircuitKey: invoices.CircuitKey{
 			ChanID: lnwire.NewShortChanIDFromInt(
-				rapid.Uint64().Draw(t, "chan_id"),
+				rand.Uint64(),
 			),
 		},
 	}


ffranr:
@GeorgeTsagk This is the rapid fail I have locally. Not sure if this is the same bug.

[TestAuxInvoiceManagerProperty_invoice_manager-20250501230723-340947.zip](https://github.com/user-attachments/files/21375658/TestAuxInvoiceManagerProperty_invoice_manager-20250501230723-340947.zip)

tapchannel/aux_invoice_manager_test.go
```diff
@@ -820,11 +821,10 @@ func genHtlc(t *rapid.T, balance []*rfqmsg.AssetBalance,
 // method also returns the assetUnits and the rfqID used by the htlc.
 func genRequest(t *rapid.T) (lndclient.InvoiceHtlcModifyRequest, uint64,
 	asset.ID, rfqmsg.ID) {
-
 	request := lndclient.InvoiceHtlcModifyRequest{
 		CircuitKey: invoices.CircuitKey{
 			ChanID: lnwire.NewShortChanIDFromInt(
-				rapid.Uint64().Draw(t, "chan_id"),
+				rand.Uint64(),
```
ffranr, line 827: So the proposed solution is to draw a random number from a different generator? I don’t see how that addresses the flakiness.

The benefit of using Rapid is that we can set the seed once and have all random sampling follow it. What’s the advantage of switching to `rand` in this case?

ffranr: What I mean is: the set of possible drawn values hasn’t changed, right? So isn’t a collision just as likely as before?

GeorgeTsagk: > What I mean is: the set of possible drawn values hasn’t changed, right? So isn’t a collision just as likely as before?

The `rapid` framework involved quite a lot of intelligence in how it draws the values, causing many conflicts in most of the test cases. Given this is a chan_id we don't really want to have a conflict in that field, it's meant to be random but also different.

So to summarize: yes the solution is to change the generator, as this one is actually random and effectively cannot produce a conflict (chance of collision is `~5*(10^-20)`)

ffranr: I see what you mean when it comes to shrinking now. I can't think of a better approach.

PR #1673, opened 2025-07-22
Invoice manager tolerance tweaks

GeorgeTsagk:
## Description

First commit exposes our current tolerance, leading to new rounding errors.

There was also an inaccuracy in the way we sharded the amount across the number of HTLCs. We had to also carry the remainder of `invoiceAmt / numHTLCs`, which is currently added to the last shard.

Then we modify the tolerance model to instead be `(1unit + 1msat)*numHTLCs`

Let's use this topology to put things in order
```
Alice --[units]--> Bob --[msat]--> Carol --[units]--> Dave 
```

Dave creates an asset invoice, Alice pays to it (with `N` HTLCs)

- When an HTLC is forwarded to Carol: Carol converts msats to units, this justifies `N units` of tolerance
- When an HTLC is received by Dave: Dave converts the units of the HTLC back to msats, as the invoice is measured in msats, this justifies `N msats` of tolerance

The current rapid test tolerance model follows the above model and seems to be happy (for `rapid.checks` up to 10M)

### DIFF
--- a/tapchannel/aux_traffic_shaper_test.go
+++ b/tapchannel/aux_traffic_shaper_test.go
@@ -89,26 +89,46 @@ func TestUnitConversionToleranceRapid(t *testing.T) {
 
 		shardSizeMSat := invoiceAmtMsat / lnwire.MilliSatoshi(numHTLCs)
 		shardSizeFP := rfqmath.MilliSatoshiToUnits(shardSizeMSat, rate)
-		shardSizeUnit := shardSizeFP.ScaleTo(0).ToUint64()
 
-		shardSumFP := rfqmath.NewBigIntFixedPoint(
-			shardSizeUnit*numHTLCs, 0,
+		// In order to account for the full invoice amt we need to also
+		// somehow carry the remainder of invoiceAmt/numHTLCs. We do so
+		// by appending it to the last shard.
+		invoiceMod := invoiceAmtMsat % lnwire.MilliSatoshi(numHTLCs)
+		lastShardMsat := shardSizeMSat + invoiceMod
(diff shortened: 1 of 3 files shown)


ZZiigguurraatt:
how does this differ from https://github.com/lightninglabs/taproot-assets/pull/1639 ?

GeorgeTsagk:
> how does this differ from https://github.com/lightninglabs/taproot-assets/pull/1639 ?

above adds a single 1 unit per invoice on the tolerance

This approach adds N msats, where N is the number of HTLCs paying to this invoice

litbot-9000:
@ffranr: review reminder

tapchannel/aux_traffic_shaper_test.go
```diff
@@ -89,26 +89,46 @@ func TestUnitConversionToleranceRapid(t *testing.T) {
...
+		lastShardMsat := shardSizeMSat + invoiceMod
+		lastShardFP := rfqmath.MilliSatoshiToUnits(lastShardMsat, rate)
+
+		var inboundAmountMSat lnwire.MilliSatoshi
+
+		// Each shard calls individually into the rfqmath library. This
+		// allows for a total error that scales with the number of
+		// shards. We go up to numHTLCS-1 because we want to add any
+		// leftovers to the last shard.
+		for range numHTLCs - 1 {
```
Roasbeef, line 110: Heh, I always forget that we have this new range syntax to just do a simple loop N times.

Roasbeef (approved):
LGTM ❄️

Roasbeef:
tacked on a commit with release notes

Issue #1463, opened 2025-04-08
[feature]: implement on-chain universe asset commitment verifier


Roasbeef:
**Is your feature request related to a problem? Please describe.**

This is related to #1451 and #1452. 

With those two issues complete, we have the state machine that'll commit the new asset commitment into the chain. We'll then also have a new RPC layer to serve the new supply related trees, and also is _meant_ to have a new on-chain merkle proof associated with each served leaf. 

The final missing component is the on-chain verifier state machine. 

**Describe the solution you'd like**

The on-chain verifier state machine keeps track of the set of synced grouped assets that will maintain a universe commitment on chain. 

For each of those assets, it first needs to "sync" the on-chain portion of the commitment chain. This amounts to looking at the first issuance event, locating the pre-commitment output, then following that pre-commitment output all the way to the currently unspent commitment output. 

Along the way, we may need to implement support for serving old roots to allow the verifier to check that the commitments have been created properly. 

Once the final unspent commitment output is reached, then we've finishing sync mode and can now await for a future spend. 

```mermaid
stateDiagram-v2
    direction TB // Optional: Suggests a Top-to-Bottom layout

    [*] --> init : "Start"
    init --> sync : "initial issuance"
    sync --> sync : "commitment spend validation"
    sync --> idle : "sync finished"
    idle --> sync : "commitment spend"
```

Roasbeef:
Idk why the self loop to sync rendered all weird...

ffranr:
**Supply‑Commitment Verification—Clarifying Scope and Failure Modes**

My understanding is that the verifier’s core job is simple: confirm that a supply‑commitment transaction is authenticated by the asset‑group minter.

What’s unclear to me is how we react when verification fails. My working assumption is that an invalid supply commitment **does not** invalidate any issued assets.

Given that assumption, two additional verifier duties described in this issue appear unnecessary:

### 1. Verifying superseded commitments

In my view, once a newer commitment exists for the same universe/asset group, earlier ones become irrelevant. Even if an older commitment is invalid, I don’t see any corrective action for a tapd node—just accept the most recent valid commitment and continue. Perhaps asset holders might only care about the latest snapshot.

### 2. Enforcing an unbroken issuance and supply commitment tx chain

[...]

PR #1674, opened 2025-07-23
supplyverifier: add push syncer and universe server verification

ffranr:
Towards closing: https://github.com/lightninglabs/taproot-assets/issues/1463

---

This PR introduces the initial push sync flow for supply commitments:

1. **Syncer (issuer side):**

   * After broadcasting the anchoring transaction (but before finalization), the issuer can push a supply commitment to a universe server.

2. **Universe server:**

   * Verifies the received supply commitment.
   * Stores it in its local database upon successful verification.
   * Serves the commitment through its RPC endpoints.

3. **Integration test:**

   * Extended to fetch the committed supply snapshot from the universe server, ensuring verification and persistence are working correctly.

This PR establishes the push route for supply commitment sync. A follow-up PR will add the pull route to allow clients to fetch and verify supply commitments from a universe server.

---

#### Notes

* There are likely opportunities to simplify and reduce code duplication. In this PR the focus is on delivering the required functionality. Once the architecture has stabilized, we can revisit the code to simplify and consolidate any duplicated logic.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674)

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -58,6 +58,7 @@
    - https://github.com/lightninglabs/taproot-assets/pull/1587
    - https://github.com/lightninglabs/taproot-assets/pull/1716
    - https://github.com/lightninglabs/taproot-assets/pull/1675
+   - https://github.com/lightninglabs/taproot-assets/pull/1674
 
 - A new [address version 2 was introduced that supports grouped assets and
   custom (sender-defined)


universe/supplyverifier/env.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/lightninglabs/taproot-assets/tapgarden"
+	"github.com/lightninglabs/taproot-assets/universe/supplycommit"
+)
+
+// OnChainLookup is an interface that is used to look up on-chain information
+// about supply commitments.
+type OnChainLookup interface {
+	// UnspentPrecommits returns the set of unspent pre-commitments for a
+	// given asset spec.
+	UnspentPrecommits(ctx context.Context,
```
guggero, line 32: nit: capitalize C in name as well here, so `UnspentPreCommits`?

universe/supplyverifier/states.go
```diff
@@ -0,0 +1,218 @@
...
+	return false
+}
+
+// String returns the name of the state.
+func (v *VerifySupplyCommitState) String() string {
+	return "VerifySupplyCommitState"
+}
+
+// IdleState is the state we reach when a valid unspent commitment output is
+// observed. We wait for a spend to re-enter the sync state.
```
guggero: I don't have much experience with the protofsm package and the state machine. But what I found helpful in Laolu's PRs was the Godoc description for each state that described what events a state can process and what other state that leads to. From that a nice diagram can then easily be created using AI tools.

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,284 @@
...
+					protofsm.SpendMapper[Event](mapper),
+				),
+			})
+		}
+
+		// If after all that, we have nothing to watch, we can just go
+		// to the idle state.
+		if len(events) == 0 {
+			return &StateTransition{
+				NextState: &IdleState{},
```
guggero: Ah, just for me to check my understanding of the state machine and the code so far: If there are no outputs to watch, we go into the idle state, at which the machine does nothing. Once a spend event comes in (presumably from watching the chain), we then transition back into the sync state.

What I don't yet see in the code is the actual "watch on-chain" part. Shouldn't there be a call to `RegisterConfNtfn` or `RegisterSpendNtfn` and a wait somewhere?

guggero (commented):
Took a quick first look. Makes sense to me this far, nothing jumps out as being on the wrong track.

universe/supplyverifier/states.go
```diff
@@ -0,0 +1,218 @@
...
+// stateSealed is a special method that is used to seal the interface.
+func (i *IdleState) stateSealed() {}
+
+// IsTerminal returns true if the target state is a terminal state.
+func (i *IdleState) IsTerminal() bool {
+	return false
+}
+
+// String returns the name of the state.
+func (i *IdleState) String() string {
```
guggero: Just took another look at how things are done in the supply commit state machine. And it looks like the idle state there is just the `DefaultState`. I think we should do the same here. Or if you want things to be more explicit in the name, then do we even need the `DefaultState` in this package? Where will it be used?

universe/supplyverifier/states.go
Roasbeef: Yeah I after we get a bit further along w/ the final version, i can generate another set of similar docs/diagrams, and also share the prompts I used to generate them.

universe/supplyverifier/states.go
```diff
@@ -0,0 +1,218 @@
...
+	PreCommits   supplycommit.PreCommits
+	SupplyCommit *supplycommit.RootCommitment
+}
+
+// eventSealed is a special method that is used to seal the interface.
+func (e *WatchOutputsEvent) eventSealed() {}
+
+// SpendEvent is sent in response to an intent to be notified of a spend of an
+// outpoint.
+type SpendEvent struct {
```
Roasbeef: 👍

universe/supplyverifier/env.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+
+	"github.com/lightninglabs/taproot-assets/asset"
+	"github.com/lightninglabs/taproot-assets/tapgarden"
+	"github.com/lightninglabs/taproot-assets/universe/supplycommit"
+)
+
+// OnChainLookup is an interface that is used to look up on-chain information
+// about supply commitments.
+type OnChainLookup interface {
```
Roasbeef: Doesn't it also need a way to fetch the asset commitment leaves/tree that existed at a given height? 

The loop I have in my mind is:
  1. Verify a new asset, realize that it has a universe commitment field set, and a pre-commitment. 
  2. Make one of these state machines for it. 
  3. It watches for the pre-commitment outpoint spend.
  3a. In this state, if the normal syncing detects a new issuance event (as rn we don't force input linkage between minting events), then we send an event to also watch that outpoint. 
  4. After it's spent, fetch from the universe RPC server, using the new height based index we added recently. 
  5. After fetching that, verify everything matches up (you get the delta, the prior tree, then the new root). 
  6. If it all matches up, add the pre-commitment, go back to the watch for spend state.

universe/supplyverifier/states.go
Roasbeef: Mmhm, np for it to just loop back to the default state. While it loops back, it can also send an "internal event" to stage another transition. The internal event might just be telling it about the new unspent pre-commitment output.

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,284 @@
...
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to fetch last "+
+				"verified commitment: %w", err)
+		}
+
+		// If we don't have a last verified commitment, then this is the
+		// first time we're verifying this asset group. We'll kick
+		// things off by watching for the latest on-chain state.
+		if lastCommit == nil {
```
Roasbeef: Shouldn't we also check for a pre-commitment here? As when we first come up, we may not know of the verified commitment at all, but if it's a grouped asset w/ universe commitments, then it should have a pre-commitment output on chain.

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,284 @@
...
+			return &StateTransition{
+				NextState: &WatchOutputsSpendState{},
```
Roasbeef: Should this just go back to default with a self-transition?

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,284 @@
...
+			events = append(events, &protofsm.RegisterSpend[Event]{
+				OutPoint: outpoint,
+				PkScript: txOut.PkScript,
+				PostSpendEvent: lfn.Some(
+					protofsm.SpendMapper[Event](mapper),
+				),
+			})
+		}
+
+		// We'll treat an error here as the commitment not being found,
```
Roasbeef: Comment doesn't seem to match this section, there's no error being checked here.

universe/supplyverifier/transitions.go
Roasbeef: So the register spend ntfn is what's sent as `ExternalEvents` below. If a spend occurs, then a new event will be sent in that contains the spending transaction, etc.

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,284 @@
...
+func (w *WatchOutputsSpendState) ProcessEvent(event Event,
+	env *Environment) (*StateTransition, error) {
+
+	switch e := event.(type) {
+	case *WatchLatestOutputsEvent:
+		ctx := context.Background()
+
+		// We'll gather all the UTXOs we need to watch. We start with
+		// the unspent pre-commitments.
+		preCommits, err := env.OnChainLookup.UnspentPrecommits(
```
Roasbeef: Need to get through a bit more, but perhaps we can collapse this state into the above? 

If we don't have an any unspent pre-commitments, or then we do nothing. 

From the default state, we can then also accept a new event that's sent every block (or w/e) to check for new unspent pre-commitments. If we have some, then we go to the spend phase, etc. The callers can also send these events anytime they detect a new spend.

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,284 @@
...
+	return nil, fmt.Errorf("%w: received %T while in %T",
+		ErrInvalidStateTransition, event, w)
+}
+
+// ProcessEvent handles state transitions for the SyncSupplyProofsState.
+func (s *SyncSupplyProofsState) ProcessEvent(event Event,
+	env *Environment) (*StateTransition, error) {
+
+	switch e := event.(type) {
+	case *SpendEvent:
```
Roasbeef: One thing that we'll want to make sure we can handle is that: if there's 5 unspent pre-commitments (we're just catching up, and they recently did a batch mint or something), then we'll get this event 5 times. So any state after this, needs to be able to handle the spend event as a noop, if it doesn't contribute any new information.

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,284 @@
...
+		ctx := context.Background()
+		nextCommitment, err := env.OnChainLookup.SupplyCommit(
+			ctx, env.AssetSpec,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to fetch next "+
+				"commitment: %w", err)
+		}
+
+		// TODO(ffranr): proof syncing logic here.
```
Roasbeef: So this is the queries to go height-by-height right?

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,284 @@
...
+// ProcessEvent handles state transitions for the VerifySupplyCommitState.
+func (v *VerifySupplyCommitState) ProcessEvent(event Event,
+	env *Environment) (*StateTransition, error) {
+
+	switch event.(type) {
+	case *ProofsSyncedEvent:
+		// TODO(ffranr): verification logic here. This should include
+		// checking that there are no more unspent pre-commitments. All
+		// the pre-commitments should have been spent in the same supply
+		// commitment transaction.
```
Roasbeef: I'd imagine that we may go back and forth between the fetch, then verify states many times until we reach the very last unspent pre-commitment. So we go height by height to fetch the latest commitment state, verify it matches up, set up the spend for the next one, repeat. 

One option would be to split things into like an "initial sync" vs "sync from tip" transitions. On the other hand, i miagine that we can just unify it, as the spends will be insta if we're still catching up.

universe/supplyverifier/transitions.go
ffranr: Yes, I see it that way also.

I think we should also ensure that the spend consumes all outstanding pre-commitment outputs. The `SpendEvent` includes the fields `SpentPreCommitment` and `PreCommitments` to track this.

universe/supplyverifier/transitions.go
ffranr: Yes. At this point, we know that either a pre-commitment output or a supply commitment output has been spent, so we need to look up any associated supply-related leaves.

universe/supplyverifier/transitions.go
ffranr: > as the spends will be insta if we're still catching up.

That's what I'm hoping for. I think the spend notification should work like that.

universe/supplyverifier/env.go
ffranr: `Precommits` is currently consistent with what we have in the rest of the codebase, as far as I can tell. For example `FetchUnspentPrecommits` and the `CommitmentTracker` interface in `universe/supplycommit`.

tapdb/sqlc/migrations/000044_supply_syncer_log.up.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE TABLE supply_syncer_log (
+    id INTEGER PRIMARY KEY,
+
+    -- The tweaked group key identifying the asset group this sync log belongs
+    -- to. This should match the group_key format used in universe_supply_roots.
+    group_key BLOB UNIQUE NOT NULL CHECK(length(group_key) = 33),
+
+    -- The latest block height that has been successfully synced for this asset
+    -- group.
+    latest_sync_block_height INTEGER NOT NULL
```
Roasbeef: 👍

tapdb/supply_syncer.go
```diff
@@ -0,0 +1,84 @@
...
+		return fmt.Errorf("group key must be specified for supply "+
+			"syncer: %w", err)
+	}
+
+	var writeTx BaseUniverseStoreOptions
+	return s.db.ExecTx(ctx, &writeTx, func(dbTx BaseUniverseStore) error {
+		// Reuse the internal supply update logic which handles all
+		// the complex sub-tree and root tree updates within the
+		// transaction.
+		_, err := applySupplyUpdatesInternal(ctx, dbTx, spec, updates)
```
Roasbeef: So we don't want to insert these items yet, as they haven't been validated. Instead, we want to insert them on disk as supply update logs. They can be inserted as dangling, and may need some other attribtues such as that these are staged, or their height, etc. 

Then the verifier (which should now persistently which height it has verified up to), can do a query for the next set of updates to verify. Once it verifies those, then it can all a method like this.

Roasbeef (changes requested):

universe/supplyverifier/syncer.go
```diff
@@ -0,0 +1,117 @@
...
+	leaves, err := client.FetchSupplyLeaves(
+		ctx, assetSpec, fn.Some(startBlockHeight), fn.None[uint32](),
+	)
+	if err != nil {
+		return zero, fmt.Errorf("unable to fetch supply leaves: %w",
+			err)
+	}
+
+	// Convert SupplyLeaves to SupplyUpdateEvents for storage.
+	var updates []supplycommit.SupplyUpdateEvent
```
guggero: nit: pre-allocate?

universe/supplyverifier/syncer.go
```diff
@@ -0,0 +1,117 @@
...
+//
+// NOTE: This function is thread safe.
+func (s *SupplySyncer) FetchLeaves(ctx context.Context,
+	serverAddr universe.ServerAddr, assetSpec asset.Specifier,
+	startBlockHeight uint32) (supplycommit.SupplyLeaves, error) {
+
+	var zero supplycommit.SupplyLeaves
+
+	// Create a client for the specific universe server address
+	client, err := s.clientFactory(serverAddr)
```
guggero: LOL, "factory" takes me back to my Java days :joy:

tapdb/sqlc/migrations/000044_supply_syncer_log.up.sql
```diff
@@ -0,0 +1,22 @@
...
+    max_fetched_block_height INTEGER,
+
+    -- The highest block height among all supply leaves inserted so far
+    -- into the canonical universe for this asset group. If NULL, no leaves
+    -- have been inserted yet.
+    max_inserted_block_height INTEGER
+);
+
+-- Add index for frequent lookups by group key.
+CREATE INDEX supply_syncer_log_group_key_idx ON supply_syncer_log(group_key);
```
guggero: Using `UNIQUE` will create a unique index already. So either we drop the `UNIQUE` keyword above and add a named unique constraint here (prefer slightly because a named constraint can be dropped/changed) or just remove the additional index.

config.go
```diff
@@ -195,7 +195,7 @@ type Config struct {
 	// SupplyCommitManager is a service that is used to manage supply
 	// commitments for assets. Supply commitments are issuer published
 	// attestations of the total supply of an asset.
-	SupplyCommitManager *supplycommit.MultiStateMachineManager
+	SupplyCommitManager *supplycommit.Manager
```
guggero, line 199: :pray:

universe/supplyverifier/manager.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+
+	sm.RemoveStateSub(sub)
+
+	return nil
+}
+
+// stateMachineCache is a thread-safe cache mapping an asset group's public key
+// to its supply verifier state machine.
+type stateMachineCache struct {
```
guggero, line 584: Can we extract this into a separate package so we don't have to re-defined here? Looks to be identical to the one in `supplycommit` (but probably needs to be parametrized since it's a different state machine struct that's cached).

guggero (commented):
Did another pass to load current context. Looks good!

tapdb/sqlc/migrations/000044_supply_syncer_log.up.sql
ffranr: gone with: named unique constraint

universe/supplyverifier/manager.go
ffranr: Yes, I need to reconcile that eventually. I'm focusing on getting the end-to-end itest working first, then my plan is to revisit and clean up any remaining duplication.

tapcfg/server.go
```diff
@@ -624,6 +625,32 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
...
+	// Set up the supply verifier, which validates supply commitment leaves
+	// published by asset issuers.
+	//
+	// Initialize the database backend for the supply syncer.
+	supplySyncerDb := tapdb.NewTransactionExecutor(
+		db, func(tx *sql.Tx) tapdb.BaseUniverseStore {
+			return db.WithTx(tx)
+		},
+	)
+	supplySyncerStore := tapdb.NewSupplySyncerStore(supplySyncerDb)
```
guggero: This can just be: `tapdb.NewSupplySyncerStore(uniDB)` as that has the same executor signature.

ffranr:
**Current state:**

* The fetch/pull side of the syncer and the `FetchSupplyCommit` RPC endpoint are incomplete. They are missing fields required for straightforward verification, such as those already present in the `InsertSupplyCommit` RPC endpoint. For example, the request message should likely include a block height range start.

* The `FetchSupplyLeaves` RPC endpoint has been reframed and is no longer served publicly by the universe server. It is now primarily for itest support and can likely be dropped later.

* The starting point for universe verification is now in the `InsertSupplyCommit` RPC endpoint, via `r.cfg.SupplyVerifyManager.InsertSupplyCommit`.

* The syncer API uses push/pull terminology, while the RPC level uses insert/fetch. This matches perspectives: the syncer “pushes” to the universe server, which “inserts” on its side.

* The supply commit creator state machine now attempts to push commitments to remote canonical universe servers via the syncer. Since we currently cannot set canonical universe addresses in asset metadata during minting (issue: https://github.com/lightninglabs/taproot-assets/issues/1728), we need a fallback universe address. This address should be added to the supply commit state machine `Environment` and passed into the syncer.

tapdb/sqlc/migrations/000044_supply_syncer_log.up.sql
```diff
@@ -0,0 +1,43 @@
...
+
+-- Add index for frequent lookups by group key.
+CREATE UNIQUE INDEX supply_syncer_log_group_key_idx
+    ON supply_syncer_log(group_key);
+
+-- A nullable column to track the previous supply commitment that was spent to
+-- create a new supply commitment. This is only NULL for the very first
+-- commitment of an asset group, each subsequent commitment needs to spend a
+-- prior commitment to ensure continuity in the supply chain.
+ALTER TABLE supply_commitments
```
Roasbeef: 👍

Roasbeef: FWIW, something like this _does_ exist in `mint_anchor_uni_commitments` after my open PR. I ended up also adding the outpoint that creates the pre-commitment as well for spending purpose. It's possible to get to the supply commit that spends it with some extra queries, but also nice to just have it directly here as well.

tapdb/supply_syncer.go
Roasbeef: Comment still stands. IIUC we have a new top level `InsertSupplyCommit` that can be used to insert the pre-validated commitment. Right now it accepts all the leaves, but could in theory just do a query for dangling supply updates, and then insert those.

tapdb/supply_syncer.go
```diff
@@ -0,0 +1,147 @@
...
+				"height: %w", err)
+		}
+
+		return nil
+	})
+}
+
+// LogRemoteInsert logs that supply leaves have been successfully inserted
+// into a remote universe.
+func (s *SupplySyncerStore) LogRemoteInsert(ctx context.Context,
```
Roasbeef: Shouldn't we insert the dangling log updates _and_ update the syncer log in the same transaction?

taprpc/universerpc/universe.proto
```diff
@@ -796,17 +804,25 @@ message FetchSupplyCommitRequest {
...
-    // the endpoint will generate and return an inclusion proof.
-    repeated bytes issuance_leaf_keys = 3;
-
-    // Optional: A list of burn leaf keys. For each key in this list,
-    // the endpoint will generate and return an inclusion proof.
-    repeated bytes burn_leaf_keys = 4;
-
-    // Optional: A list of ignore leaf keys. For each key in this list, the
-    // endpoint will generate and return an inclusion proof.
-    repeated bytes ignore_leaf_keys = 5;
```
Roasbeef, line 809: So we feel like these aren't useful anymore? 

My use case was a client fetching a burn proof by the burnt asset point, and receiving all the merkle proofs alongside it.

taprpc/universerpc/universe.proto
```diff
@@ -863,19 +881,25 @@ message FetchSupplyCommitResponse {
     // The root of the ignore tree for the specified asset.
     SupplyCommitSubtreeRoot ignore_subtree_root = 11;
 
-    // Inclusion proofs for each issuance leaf key provided in the request.
-    // Each entry corresponds to the key at the same index in
-    // `issuance_leaf_keys`.
-    repeated bytes issuance_leaf_inclusion_proofs = 12;
+    // The issuance leaves that were added by this supply commitment. Does not
```
Roasbeef, line 861: Re what I mentioned around the summing. I think we can add a `total_outstanding_supply` field here, which does the subtraction we need. Clients can always verify it, this is just a connivence field.

universe/supplycommit/env.go
```diff
@@ -268,6 +398,93 @@ func (r *RootCommitment) TapscriptRoot() ([]byte, error) {
 	return computeSupplyCommitTapscriptRoot(supplyRootHash)
 }
 
+// Verify checks that the root commitment is formally valid. This includes
+// checking that the on-chain information is correct.
+func (r *RootCommitment) Verify(merkleVerifier proof.MerkleVerifier,
```
Roasbeef: Still getting through the diff, but this should also accept the asset meta so it can verify that the expected delegation key was used to spend/create the commitment. 

Alternatively, the RPC that attempts to insert the supply commit, should sign a digest using the delegation key.

universe/supplycommit/env.go
```diff
@@ -470,6 +686,17 @@ type StateMachineStore interface {
...
 }
 
+// SupplySyncer is an interface that allows the state machine to insert
+// supply commitments into the remote universe server.
+type SupplySyncer interface {
+	// PushSupplyCommitment pushes a supply commitment to the remote
+	// universe server. This function should block until the sync insertion
+	// is complete.
+	PushSupplyCommitment(ctx context.Context, assetSpec asset.Specifier,
+		commitment RootCommitment, updateLeaves SupplyLeaves,
```
Roasbeef, line 760: Update leaves here is all leaves, or just the new set?

universe/supplycommit/manager.go
```diff
@@ -129,10 +138,59 @@ func (m *MultiStateMachineManager) Stop() error {
...
+
+	// Fetch the delegation key locator. We need to ensure that the
+	// delegation key is owned by this node, so that we can generate
+	// supply commitments (ignore tuples) for this asset group.
+	_, err = m.cfg.AssetLookup.FetchInternalKeyLocator(
+		ctx, &delegationPubKey,
+	)
+	switch {
+	case errors.Is(err, address.ErrInternalKeyNotFound):
+		return fmt.Errorf("delegation key locator not found; " +
```
Roasbeef: Can pass along the `address.ErrInternalKeyNotFound` error wrapped so it can't be checked easily by a caller.

universe/supplycommit/manager.go
```diff
@@ -290,10 +366,25 @@ func (m *MultiStateMachineManager) SendBurnEvent(ctx context.Context,
 	return m.SendEventSync(ctx, assetSpec, burnEvent)
 }
 
+// StartSupplyPublishFlow triggers the state machine to build and publish
+// a new supply commitment if pending supply tree updates exist.
+func (m *Manager) StartSupplyPublishFlow(ctx context.Context,
```
Roasbeef, line 382: One thing that came up during my itests for https://github.com/lightninglabs/taproot-assets/pull/1675 is that we might still want a normal time based trigger, but only after an update has been inserted. 

As otherwise right now, the caller always needs to manually update the commitment over RPC. WE can do this automatically for RPCs like ignore, and internally in the relevant sub-systems for mint/ignore.

universe/supplyverifier/manager.go
```diff
@@ -0,0 +1,567 @@
...
+	// FetchState attempts to fetch the state of the state machine for the
+	// target asset specifier. If the state machine doesn't exist, then a
+	// default state will be returned.
+	FetchState(context.Context, asset.Specifier) (State, error)
+}
+
+// IssuanceSubscriptions allows verifier state machines to subscribe to
+// asset group issuance events.
+type IssuanceSubscriptions interface {
+	// RegisterSubscriber registers an event receiver to receive future
```
Roasbeef, line 54: Do we have a way to handle dropped events here and/or restarts? 

So an issuance event happens (we detect this in the normal universe syncer), then we want to notify this sub-system. What happens if we send that event, write the new issuance to disk, then crash? You can reverse the order of send vs write as well.

universe/supplyverifier/manager.go
```diff
@@ -0,0 +1,567 @@
...
+	leaves supplycommit.SupplyLeaves) error {
+
+	headerVerifier := tapgarden.GenHeaderVerifier(ctx, m.cfg.Chain)
+
+	// Do the static on-chain verification of the supply commitment.
+	err := commitment.Verify(proof.DefaultMerkleVerifier, headerVerifier)
+	if err != nil {
+		return fmt.Errorf("unable to verify supply commitment: %w", err)
+	}
+
```
Roasbeef, line 241: Should also verify delegation key information, otherwise impersonation is possible.

tapdb/sqlc/migrations/000044_supply_syncer_log.up.sql
guggero: Yes, but that table would only be set on the node that _created_ the supply commitment. Because we're also using the `supply_commitments` table on the universe server to store them, we wouldn't have any data in that table (and IMO it wouldn't make sense to store it there for the universe).

universe/supplycommit/env.go
guggero: Just the new set. In all the RPC and database calls that fetch or push a single supply commit, the leaves are always only the set created/added by that supply commitment. At least that's the goal, it's possible it isn't fully implemented everywhere yet in the current diff.

litbot-9000:

ffranr:
While working on https://github.com/lightninglabs/taproot-assets/pull/1777 I noticed we also need to add functionality in this PR to store supply pre-commitments for assets issued by a remote peer (i.e. when we insert synced issuance proofs). Without this, verification does not properly check that the supply commit anchor transaction spends an output from the minting anchor transaction, which is critical for authentication.

We also need to assert during verification that either a supply commitment output is spent or at least some pre-commitment outputs.

jtobin (approved):
(We discussed the merge strategy out-of-band -- LGTM. :+1: :+1:)

itest/supply_commit_test.go
```diff
@@ -466,6 +467,75 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
...
+				},
+			},
+		)
+		require.NoError(t.t, err)
+
+		// If the fetch response does not include a block height, the
+		// supply commitment transaction has not been mined yet, so we
+		// should retry.
+		if uniFetchResp.ChainData.BlockHeight == 0 {
+			return false
```
GeorgeTsagk, line 530: since we're done with mining in the above lines isn't this a dead code path?

itest/supply_commit_test.go
```diff
@@ -466,6 +467,75 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
...
+		t.t, 0,
+		uniFetchResp.BurnSubtreeRoot.RootNode.RootSum,
+	)
+
+	// Assert that the ignore subtree root sum matches the total amount
+	// of ignored assets.
+	require.EqualValues(
+		t.t, ignoreAmt,
+		uniFetchResp.IgnoreSubtreeRoot.RootNode.RootSum,
+	)
```
GeorgeTsagk, line 588: So in the future in order to complete the itest we'd have to introduce a 3rd node that would attempt to receive those assets, but would invalidate them by checking the ignore tree and realizing that the outpoint is marked?

itest/supply_commit_test.go
ffranr: mining confirms the supply commit tx, but there is more code that needs to run (write confirmed tx supply commit to db) before `FetchSupplyCommit` will return the result that we expect. We wait for that code to complete by using `require.Eventually` here. Does that answer your question?

itest/supply_commit_test.go
ffranr: Yes, we need to extend itest coverage (in a future PR) to ensure a third party peer will reject ignored asset points.

ffranr:
@GeorgeTsagk 

> The Mint/Burn trees are append-only, as you can't really revert those actions since their confirmation on-chain. The ignore tree seems to follow a similar approach, do we never want to allow for an outpoint to stop being ignored?

We could potentially add an ignore reversion proof in the future, though no design exists for this yet.

> Since ignoring and publishing the updated supply commitment happen at different times, the outpoint that we're ignoring may have been already spent. This way we're effectively ignoring all the downstream newly created outpoints (that originated from the one we're targeting in IgnoreAssetOutPoint). How do we get around this timing problem?

We don't get around the timing problem. We're ok with the issuer being able to ignore any outpoint, spent or otherwise.

> Does the verifying node (the receiver of assets) include fetching the latest supply commitment as part of the transfer validation? For example, if you are offering me assets that exist within an outpoint that was created at a blockheight greater than that of the last supplycommitment's then I might want to wait a bit until the height of the latest published commitment "covers" your outpoint height. It can induce extra delays and also complexity, just curious around the future approach here.

The ignore checker used during proof verification is already in place. Since spent outpoints can be ignored by an issuer at any point, the receiver shouldn't wait for an updated supply commitment before receiving.

GeorgeTsagk:
Seems like this PR also needs a rebase (I think the base itself got updated?). Otherwise LGTM!

Roasbeef (changes requested):
> @Roasbeef reviewed 2 of 17 files at r1, 42 of 42 files at r2, 45 of 45 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674)* status: all files reviewed, 44 unresolved discussions (waiting on @ffranr, @GeorgeTsagk, and @guggero)

___
*[`universe/supplyverifier/verifier.go` line 142 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH4uakCrseXEKovIxX:-OZH4uakCrseXEKovIxY:b4l8ltr) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/universe/supplyverifier/verifier.go#L142)):*
> ```Go
> // verifyInitialCommit verifies the first (starting) supply commitment for a
> // given asset group.
> func (v *Verifier) verifyInitialCommit(ctx context.Context,
> ```

Additional things we should verify:
  * the signatures for the ignore updates
  * for burn+ignore, verify the proofs
  * verify the anchor output information:
      * the delegation key was used
      * the root committed to matches up 
___
*[`universe/supplyverifier/verifier.go` line 166 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH4YM2ArUp4fC0UA-q:-OZH4YM2ArUp4fC0UA-r:bd3bnae) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/universe/supplyverifier/verifier.go#L166)):*
> ```Go
> 		}
> 
> 		return fmt.Errorf("found initial commitment for asset group; "+
> ```

I'm not sure I understand this check. Wouldn't it always fail after we've inserted the initial commitment? 
___
*[`universe/supplyverifier/verifier.go` line 285 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH5PshB1Z1nUNrbgpt:-OZH5PshB1Z1nUNrbgpu:b90kidn) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/universe/supplyverifier/verifier.go#L285)):*
> ```Go
> 	// that they correspond to the spent supply commitment outpoint.
> 	spentRootTree, spentSubtrees, err :=
> 		v.cfg.SupplyTreeView.FetchSupplyTrees(
> ```

So we're expected to already have this on disk before we attempt to verify? Meaning we write to disk without verifying first? 
___
*[`universe/supplyverifier/verifier.go` line 391 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH5d0SBQVv2Fdn-fLS:-OZH5d0SBQVv2Fdn-fLT:bcjd6cc) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/universe/supplyverifier/verifier.go#L391)):*
> ```Go
> 		return v.verifyInitialCommit(ctx, assetSpec, commitment, leaves)
> 	}
> ```

See my comment about validation that's missing. 
___
*[`mssmt/interface.go` line 9 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH15NF7s55RIGToEj3:-OZH15NF7s55RIGToEj4:bn9j72q) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/mssmt/interface.go#L9)):*
> ```Go
> // whether to include the leaf in the copy operation. A true value means the
> // leaf should be included, while false means it should be excluded.
> type CopyFilterPredicate = func([hashSize]byte, LeafNode) (bool, error)
> ```

Any reason to reach for an alias vs an actual type here? 
___
*[`itest/assertions.go` line 2805 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH0Por57lawnOwN6-k:-OZH0Por57lawnOwN6-l:b47ktox) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/itest/assertions.go#L2805)):*
> ```Go
> 		GroupKey: groupKeyReq,
> 		Locator: &unirpc.FetchSupplyCommitRequest_VeryFirst{
> 			VeryFirst: true,
> ```

Perhaps an _index_ to fetch instead would be more generic? 
___
*[`tapdb/sqlc/migrations/000045_supply_syncer_push_log.up.sql` line 41 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH2F1EDSQZUcL46Mcy:-OZH2F1EDSQZUcL46Mcz:b-8fx2u7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/tapdb/sqlc/migrations/000045_supply_syncer_push_log.up.sql#L41)):*
> ```TSQL
> -- commitment of an asset group, each subsequent commitment needs to spend a
> -- prior commitment to ensure continuity in the supply chain.
> ALTER TABLE supply_commitments
> ```

:+1: 
___
*[`tapdb/supply_tree.go` line 324 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH1fYG6rMcgvSygU3Y:-OZH1fYG6rMcgvSygU3Z:ba9tnpe) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/tapdb/supply_tree.go#L324)):*
> ```Go
> 			}
> 
> 			filteredSubTree, err := filterSubTree(
> ```

Very cool pattern + solution here w/ the tree filter! 
___
*[`itest/supply_commit_test.go` line 508 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH6F1D60q5lKeuGrUK:-OZH6F1D60q5lKeuGrUL:bk4gaer) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/itest/supply_commit_test.go#L508)):*
> ```Go
> 
> 	t.Log("Fetch first supply commitment from universe server")
> 	// Ensure that the supply commitment was pushed to the universe server
> ```

Style nit: missing a space above. 

Can also be a sub test potentially. 
___
*[`tapdb/supply_commit_test.go` line 2115 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH6Uy_9OldsOiJxqOP:-OZH6Uy_9OldsOiJxqOQ:b-qq2pbd) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/tapdb/supply_commit_test.go#L2115)):*
> ```Go
> 	require.True(t, rows.Next(), "Expected supply commitment to exist")
> 
> 	var rootHashBytes []byte
> ```

Style nit: can collapse into a single var statement. 
___
*[`rpcserver.go` line 4352 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH5vLj3DOix3DCux7b:-OZH5vLj3DOix3DCux7c:bn0qyht) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/rpcserver.go#L4352)):*
> ```Go
> // mapSupplyLeaves is a generic helper that converts a slice of supply update
> // events into a slice of RPC SupplyLeafEntry objects.
> func mapSupplyLeaves[E any](entries []E) ([]*unirpc.SupplyLeafEntry, error) {
> ```

Couldn't this just take the supply leaves interface from the `supplycommit` package? 

You're already doing the type assert below.

universe/supplyverifier/manager.go
Roasbeef: I think this comment still stands. Something we'll want to answer for the verifier.

universe/supplyverifier/transitions.go
ffranr: I've simplified the verifier state machine significantly in the latest code.

universe/supplyverifier/env.go
ffranr: I think we're on track for this sort of setup.

universe/supplyverifier/transitions.go
ffranr: I don't think this comment is relevant any more given the new code, please correct me if I'm wrong.

universe/supplyverifier/transitions.go
ffranr: I don't think this comment is relevant any more given the new code, please correct me if I'm wrong.

universe/supplyverifier/transitions.go
ffranr: I don't think this comment is relevant any more given the new code, please correct me if I'm wrong.

universe/supplyverifier/transitions.go
ffranr: I don't think this comment is relevant any more given the new code, please correct me if I'm wrong.

tapdb/supply_syncer.go
ffranr: `InsertSupplyCommit` now accepts dangling supply leaves (leaves which are new to the inserted supply commit) rather than all leaves.

tapdb/supply_syncer.go
ffranr: I think this is an old comment and no longer relevant to the current state of the code. Please correct me if I'm wrong.

taprpc/universerpc/universe.proto
ffranr: I don't think this comment is relevant any more given the new code, please correct me if I'm wrong.

taprpc/universerpc/universe.proto
ffranr: I will add this field in this PR.

universe/supplycommit/env.go
ffranr: Additional verification checks are now run in `supplyverifier/verifier.go`.

universe/supplyverifier/manager.go
ffranr: I’m working on a PR that extends the syncer so that, in addition to writing asset issuance proofs to the DB, it also records pre-commitment entries in the `supply_pre_commits` table. With this change, crash recovery should no longer be an issue for the supply verifier manager, since it starts state machines for all known remotely issued assets at startup. A newly issued asset group (first tranche) will therefore be picked up on restart after a crash.

But we still need to make use of `IssuanceSubscriptions` with a new event type and so on.

universe/supplyverifier/manager.go
ffranr: Yes, I will look into this.

tapdb/sqlc/migrations/000044_supply_syncer_log.up.sql
ffranr: I'm adding a new supply pre-commitment table for remote issued assets in a separate PR.

universe/supplycommit/manager.go
ffranr: Done.

universe/supplycommit/manager.go
ffranr: I'd like to look into that as a follow up to this PR, if that's ok with you.

taprpc/universerpc/universe.proto
Roasbeef: :+1:

universe/supplycommit/manager.go
Roasbeef: Sure. Let's make an issues for this, as then it'll also make certain routine issuer actions easier.

Roasbeef (changes requested):
> @Roasbeef reviewed 4 of 56 files at r4, 53 of 54 files at r5, 8 of 8 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674)* status: all files reviewed, 16 unresolved discussions (waiting on @ffranr, @GeorgeTsagk, and @guggero)

___
*[`itest/supply_commit_test.go` line 508 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH6F1D60q5lKeuGrUK:-OZM_Lvd0-Zg20DMirJw:b-vgcnp) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/itest/supply_commit_test.go#L508)):*
<details><summary><i>Previously, ffranr wrote…</i></summary><blockquote>

I'd like to add more itests in a separate PR if that's ok with you. Tests specifically targetting the universe server for example.
</blockquote></details>

By sub-test here I just meant nested under `t.Run`. 
___
*[`universe/supplyverifier/verifier.go` line 142 at r3](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZH4uakCrseXEKovIxX:-OZMblN6AnSfdUboC9Bq:b-j5ujz0) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/ded200f6dfb600d995c708f0040eda3f1e547a70/universe/supplyverifier/verifier.go#L142)):*
> Burn and ignore proof verification might be limited because the universe server/peer node might not sync all asset transfers, so the complete proof file might not be available for verification

For ignore, it just needs to validate the signature. 

For burn, it can at least verify that state transition, no?
___
*[`universe/supplycommit/util.go` line 12 at r5](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZMlkNND8xLKlScEvJm:-OZMlkNND8xLKlScEvJn:b-8fx2u7) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/d930617c4e19935632991d9afbdaad2fcf224ef7/universe/supplycommit/util.go#L12)):*
> ```Go
> // CalcTotalOutstandingSupply calculates the total outstanding supply from the
> // given supply subtrees.
> func CalcTotalOutstandingSupply(ctx context.Context,
> ```

:+1: 
___
*[`universe/supplyverifier/verifier.go` line 106 at r6](https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZMmYmCEzQahmW1_EO_:-OZMmYmCEzQahmW1_EOa:b-gjck5d) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/32f18c83b7d200b5aa58b3b4e6579dcde992c466/universe/supplyverifier/verifier.go#L106)):*
> ```Go
> 	}
> 
> 	// TODO(ffranr): If commitment.SpentCommitment is none, then we
> ```

Lift into an issue rn so it isn't forgotten?

Roasbeef:
CI seems stuck for some reason?

universe/supplyverifier/verifier.go
```diff
@@ -0,0 +1,641 @@
...
+			"tree: %w", err)
+	}
+
+	// Ensure that the root of the reconstructed supply tree matches
+	// the commitment root.
+	if expectedRoot.NodeHash() != commitment.SupplyRoot.NodeHash() {
+		return fmt.Errorf("expected supply tree root does not match " +
+			"commitment supply root")
+	}
+
```
Roasbeef, line 376: Final missing thing here is to verify that this is actually what's committed to on chain (using the delegation key and the commitment we created arrives back at the same output). 

We should also verify that the pre commitments are using the delegation key as well.

Roasbeef (changes requested):

universe/supplyverifier/verifier.go
ffranr: > We should also verify that the pre commitments are using the delegation key as well.

I've created an issue here: https://github.com/lightninglabs/taproot-assets/issues/1782

> verify that this is actually what's committed to on chain (using the delegation key and the commitment we created arrives back at the same output)

This should already be included in the PR. Here: `universe/supplycommit/env.go`, method `RootCommitment.VerifyChainAnchor`. 

It uses the commitment internal key rather than the delegation key when reconstructing the supply commit output. I think that’s acceptable as long as the spend can be traced directly back through the transaction chain to a pre-commitment output that _does_ use the delegation key as its internal key.

In other words, the internal key of the supply commitment doesn’t necessarily need to be the delegation key. Am I mistaken here? Currently, when creating the first supply commitment, we derive a new internal key and don't re-use the delegation key: https://github.com/lightninglabs/taproot-assets/blob/bccb3ef06d652578932e0b4604280e1707388b5a/universe/supplycommit/transitions.go#L500

universe/supplycommit/manager.go
ffranr: Created issue for tracking here: https://github.com/lightninglabs/taproot-assets/issues/1781

ffranr:
I’ve added verifier checks on the group key for every supply subtree leaf. I also validate that the issuance leaf records the correct issuance block height and that any burn proof is a genuine asset burn.

Issue #1648, opened 2025-07-12
Hook up supply commitment state machine to minting and burning operations


Roasbeef:
## Background & Motivation

The Taproot Assets protocol allows for off-chain asset issuance and transfers, but maintaining a verifiable, global view of the total supply of any asset requires an on-chain commitment mechanism. This is critical for:

1. **Supply Transparency**: Users and auditors need to verify the total supply of an asset
2. **Preventing Inflation**: Ensuring no unauthorized minting occurs
3. **Burn Verification**: Proving that burned assets are permanently removed from circulation
4. **Universe Synchronization**: Allowing universe servers to verify supply commitments

The `universe/supplycommit` package implements a sophisticated state machine for this purpose, creating periodic on-chain commitments that aggregate all supply changes (mints, burns, and ignored outputs) into a merkle tree (MS-SMT) structure. Each commitment transaction spends the previous commitment, creating an auditable chain of supply state transitions.

## The Missing Link

While the supply commitment state machine is fully implemented, it's currently not connected to the actual minting and burning operations in the codebase. This means:

- New asset mints are not registered with the supply commitment system
- Asset burns are not tracked in the supply commitment tree
- No on-chain supply commitments are being created

## High-Level Integration Approach

The key insight is to integrate supply commitment calls directly into the existing state machines (BatchCaretaker for mints, ChainPorter for burns) rather than using external event listeners. This provides natural fault tolerance - if a supply commit fails, the state machine won't advance, and will retry on restart.

### 1. Mint Integration

**Location**: `tapgarden/caretaker.go` - in `BatchCaretaker.stateStep()`

When transitioning from `BatchStateConfirmed` to `BatchStateFinalized`:
1. Extract issuance proofs from the finalized batch
2. Create `NewMintEvent` instances for each minted asset
3. Submit to supply committer BEFORE updating batch state in database
4. Only mark batch as finalized after successful supply commit registration

This ensures if the supply commit fails, the batch remains in `BatchStateConfirmed` and will retry the entire finalization process on restart.

### 2. Burn Integration  

**Location**: `tapfreighter/chain_porter.go` - in main event loop

When processing parcels with burn outputs:
1. Detect burns via `Asset.IsBurn()` check after confirmation
2. Extract burn proofs and create `NewBurnEvent` instances
3. Submit to supply committer as part of burn finalization
4. Only mark burn as complete after successful supply commit registration

[...]

PR #1675, opened 2025-07-24
multi: hook up burn and mint events to the supply commit state machine 

Roasbeef:
In this PR, we add some extra plumbing to actually hook up the mint and burn actions originating in the daemon into the supply commit state machine. We cap things off by adding an itest that exercises supply commit creation e2e. 

Along the way, we fixed two existing bugs in related sub-sytems:
  1. We never spent the pre commitments on disk in the first place. This caused the initial itest to fail, as the state machine would be adding an extra input to the commitment transaction which was already spent. 
     * While fixing this, I added a new outpoint field into the pre commitment table to make it easy to reference for spending. 
  3. We weren't storing the fully key desc information on disk? This caused the PSBT finaliztion to fail. 
  4. We weren't passing in the tapscript root hash into the PSBT, so `lnd` wasn't able to sign for it. 

I've codified the knowledge needed to debug the last 2 issues in this PR: https://github.com/lightninglabs/taproot-assets/pull/1739. 

Fixes https://github.com/lightninglabs/taproot-assets/issues/1648

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -53,6 +53,7 @@
    - https://github.com/lightninglabs/taproot-assets/pull/1554
    - https://github.com/lightninglabs/taproot-assets/pull/1587
    - https://github.com/lightninglabs/taproot-assets/pull/1716
+   - https://github.com/lightninglabs/taproot-assets/pull/1675
 
 - A new [address version 2 was introduced that supports grouped assets and
   custom (sender-defined)


address/book_test.go
```diff
@@ -69,6 +71,184 @@ func TestBook_NewAddress(t *testing.T) {
...
+			if tt.expectedError != "" {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), tt.expectedError)
+			} else {
+				require.NoError(t, err)
+			}
+			
+			require.Equal(t, tt.expectedHas, hasDelegation)
+			
+			// Verify expectations
```
ffranr: comments can be removed here

tapgarden/planter.go
```diff
@@ -81,6 +91,15 @@ type GardenKit struct {
 	// UniversePushBatchSize is the number of minted items to push to the
 	// local universe in a single batch.
 	UniversePushBatchSize int
+
+	// MintCommitter is used to commit the minting of new assets to the
+	// supply commitment state machine.
+	MintCommitter MintSupplyCommitter
```
ffranr: I think we should call this field `MintSupplyCommitter`, but no big deal IMO

tapfreighter/chain_porter.go
```diff
@@ -90,6 +100,15 @@ type ChainPorterConfig struct {
...
 	ErrChan chan<- error
+
+	// BurnSupplyCommitter is used to track supply changes (burns) and
+	// create periodic on-chain supply commitments.
+	BurnCommitter BurnSupplyCommitter
+
+	// DelegationKeyChecker is used to verify that we control the delegation
+	// key for a given asset, which is required for creating supply
+	// commitments.
+	DelegationKeyChecker address.DelegationKeyChecker
```
ffranr, line 126: The `ChainPorter` already supports event emission, so I think we might be able to avoid leaking the concept of supply commitments into it, even if the logic is abstracted behind interfaces. Instead, the multi supply commit manager could subscribe as a consumer of `ChainPorter` events. Perhaps we define a new burn event emitted by the `ChainPorter`, and let the supply commit manager handle delegation key checks independently.

In other words, the `ChainPorter` just emits a burn event, and the multi supply commit manager does the rest. This would help keep the supply commitment feature cleanly separated.

Roasbeef: > . Instead, the multi supply commit manager could subscribe as a consumer of ChainPorter events

What happens if the daemon crashes after that event is sent, but before the supply commit manager processes it? To make that durable, we'd need to have the event send in the main state machine loop as it's implemented now.

Roasbeef: So I generally agree with you, but we need that event emission to be durable. 

One route that we can use to generally solve this problem is to extend the `actor` framework I added in `lnd`, to enable one to pass in a queue or mailbox like struct (rn it hard codes a normal Go channel). We'd then be able to make a new table of arbitrary events (with the requirement that they can be serialized), and have that back the durable mailbox for a given actor. 

This would allow us to easily generalize this pattern between any sub-system we have. It would simplify some other sub-systems, as then they wouldn't need some reconciliation loop at start up, they'd just continue to reed from their queues.

tapgarden/planter.go
Roasbeef: Done!

tapfreighter/chain_porter.go
Roasbeef: Related to this is that we have a similar gap right now with teh way the state machine works. The sends are non-blocking. What we want instead is that either the sens is blocking, or we _wait_ on some extra channel/future until the update event has been written to disk. 

To implement this, we can either implement a done call back in the event that we send over. Or extend the `lnd` state machine to make this a first class feature. 

I lean towards the done call back, as we can do that in another PR and reduce update cycles.

tapfreighter/chain_porter.go
```diff
@@ -677,6 +696,94 @@ func (p *ChainPorter) storeProofs(sendPkg *sendPackage) error {
...
+		err = p.cfg.BurnCommitter.SendBurnEvent(
+			ctx, assetSpec, burnLeaf,
+		)
+		if err != nil {
+			return fmt.Errorf("unable to send burn event for "+
+				"asset %x: %w", assetID, err)
+		}
+
+		log.Debugf("Sent supply commit burn event for asset %x",
+			assetID)
```
guggero, line 785: nit: this will double hex encode, can just use `%s` as we have a `.String()` method on the asset ID.

tapfreighter/chain_porter.go
guggero: I agree, it feels better to do things more explicitly here, since the event system is more of a "fire-and-forget" kind of system.

tapgarden/caretaker.go
```diff
@@ -1148,6 +1148,17 @@ func (b *BatchCaretaker) stateStep(currentState BatchState) (BatchState, error)
 			return 0, fmt.Errorf("unable to confirm batch: %w", err)
 		}
 
+		// Send supply commitment events for all minted assets before
+		// finalizing the batch. This ensures that supply commitments
+		// are tracked before the batch is considered complete.
+		err = b.sendSupplyCommitEvents(
```
guggero, line 1147: Should we do this before we call `MarkBatchConfirmed`? If we shut down while sending the supply commit events, the batch will already be marked as confirmed in the DB, so it won't be re-tried on next startup (and things aren't done atomically here).

guggero (approved):
Very nice! LGTM pending CI fixes.

tapgarden/caretaker.go
Roasbeef: Makes sense, swapped.

tapfreighter/chain_porter.go
Roasbeef: Done in the other PR.

Roasbeef:
Pushed a series of commits fixing some bugs:
  * We weren't storing the full internal key, so PSBT signing failed. 
  * We weren't setting the taproot merkle tree in the input for the supply commit output, so signing failed again. 
  * We weren't spending the pre-commitment outputs in the db, leading to double spends for attempts. 

After all those were fixed, the itest added passes. We mint an asset, then another one in a group, then burn an asset. Along the way we check that all the commitment information has been updated along the way.

Roasbeef:
Pushed up a commit to fix a rebase issue in the unit tests.

guggero:
I cleaned up after the AI to fix the linter issues and added release notes to get the CI green.

guggero:
I think you forgot to pull my linter fixes...

Roasbeef:
Forgot to pull down those prior chagnes before applying a fixup commit. 

Should be good now...

address/book.go
```diff
@@ -841,3 +847,73 @@ func (b *Book) RemoveSubscriber(
...
+// asset ID. Returns true if we have the private key for the asset's
+// delegation key, false otherwise.
+//
+// NOTE: This is part of the DelegationKeyChecker interface.
+func (b *Book) HasDelegationKey(ctx context.Context,
+	assetID asset.ID) (bool, error) {
+
+	assetGroup, err := b.cfg.Store.QueryAssetGroupByID(ctx, assetID)
+	if err != nil {
+		return false, fmt.Errorf("fail to find asset group given "+
```
jtobin: nit: s/fail/failed

address/book.go
```diff
@@ -841,3 +847,73 @@ func (b *Book) RemoveSubscriber(
...
+
+	delegationPubKey, err := metaReveal.DelegationKey.UnwrapOrErr(
+		fmt.Errorf("delegation key not found for given asset"),
+	)
+	if err != nil {
+		return false, err
+	}
+
+	// Now that we have the delegation key, we'll see if we know of the
+	// internal key loactor. If we do, then this means that we were the ones
```
jtobin: nit: s/loactor/locator

universe/supplycommit/multi_sm_manager.go
```diff
@@ -265,6 +266,30 @@ func (m *MultiStateMachineManager) SendEventSync(ctx context.Context,
 	return event.WaitForDone(ctx)
 }
 
+// SendMintEvent sends a mint event to the supply commitment state machine.
```
jtobin, line 274: nit: I've observed that some interface implementations specify the interface they're implementing with a NOTE in the leading comment. Not sure if this is intended to be universal practice, but this (for example) is missing it.

tapfreighter/chain_porter.go
```diff
@@ -49,6 +50,15 @@ type ProofImporter interface {
 		replace bool, proofs ...*proof.AnnotatedProof) error
 }
 
+// BurnSupplyCommitter is used by the chain porter to update the on-chain supply
+// commitment when burns new 1st party burns are confirmed.
```
jtobin: nit: s/burns/

jtobin (approved):
I've gone over these changes in detail and think they look solid. LGTM. :+1:

I'd recommend adding a brief description of the changes in the OP, for posterity. I also noted a few utterly-trivial typo-related nits I just happened to pick up on while going through the code, but they don't actually need to be fixed.

jtobin:
Worth noting that this will also resolve #1648.

tapgarden/caretaker.go
```diff
@@ -1432,6 +1445,124 @@ func GenHeaderVerifier(ctx context.Context,
...
+		if err := leafProof.Encode(&leafProofBuf); err != nil {
+			return fmt.Errorf("unable to encode leaf proof: %w",
+				err)
+		}
+		leafProofBytes := leafProofBuf.Bytes()
+
+		// With the proof extracted, we can now create the universe
+		// key and leaf.
+		universeKey := universe.BaseLeafKey{
+			OutPoint:  mintedAsset.Genesis.FirstPrevOut,
```
ffranr: I think this should be `leafProof.OutPoint()`.

Roasbeef: Nice catch! Updated.

PR #1685, opened 2025-07-29
rpc+tapdb+tapfreighter: Add historical send event streaming with database-level filtering

bhandras:
This PR enhances the `SubscribeSendEvents` RPC to support historical event replay of completed sends with efficient database-level filtering.

The PR is co-authored with Claude Code and O3.

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -75,6 +75,9 @@
 - [Rename](https://github.com/lightninglabs/taproot-assets/pull/1682) the
   `MintAsset` RPC message field from `universe_commitments` to
   `enable_supply_commitments`.
+- The `SubscribeSendEvents` RPC now supports [historical event replay of 
+  completed sends with efficient database-level
+  filtering](https://github.com/lightninglabs/taproot-assets/pull/1685).
 
 ## tapcli Additions
 


tapdb/assets_store.go
```diff
@@ -3608,6 +3621,200 @@ func (a *AssetStore) QueryParcels(ctx context.Context,
...
+		}
+
+		// Query for asset transfers with filters applied at database
+		// level.
+		dbTransfers, err := q.QueryAssetTransfers(ctx, transferQuery)
+		if err != nil {
+			return err
+		}
+
+		for idx := range dbTransfers {
```
guggero: This whole block is duplicate with `queryParcelsWithTime`, can we extract that into its own function as well please?

tapdb/assets_store.go
```diff
@@ -3608,6 +3621,200 @@ func (a *AssetStore) QueryParcels(ctx context.Context,
...
+	// For historical events, we don't require all proofs to be delivered
+	// since proof delivery can happen asynchronously.
+	var completedParcels []*tapfreighter.OutboundParcel
+	for i, parcel := range allParcels {
+		log.Debugf("QueryCompletedParcels(parcel %d): block_height=%d,"+
+			" transfer_time=%v, label=%s", i+1,
+			parcel.AnchorTxBlockHeight, parcel.TransferTime,
+			parcel.Label)
+
+		// Check if parcel is confirmed (has block height set)
```
guggero: nit: missing punctuation.

tapdb/sqlc/queries/transfers.sql
```diff
@@ -70,6 +70,26 @@ WHERE
         )
         OR @pending_transfers_only = false OR @pending_transfers_only IS NULL
     )
+
+    -- Optionally filter on a given start time.
+    AND transfers.transfer_time_unix >=
+        COALESCE(sqlc.narg('start_time'), transfers.transfer_time_unix)
+
+    -- Optionally filter on a given label.
+    AND (transfers.label = sqlc.narg('filter_label') OR
```
guggero, line 79: nit: could use `coalesce` here too?

tapdb/assets_store.go
guggero: Or I guess why do we need both these functions in the first place? Isn't calling

```
	return a.queryParcelsWithTime(
		ctx, anchorTxHash, pendingOnly, time.Time{},
	)
```

equivalent to

```
	return a.queryParcelsWithFilters(
		ctx, anchorTxHash, pendingOnly, time.Time{},, "", nil,
	)
```
?

guggero: Ah, I see all of this is done in the next commit, okay, never mind.

rpcserver.go
```diff
@@ -4446,6 +4446,14 @@ func shouldNotifyAssetSendEvent(event tapfreighter.AssetSendEvent,
 	return match
 }
 
+// reconstructSendEvent reconstructs an AssetSendEvent from a completed transfer
+// parcel stored in the database. This is used for historical event replay.
+func reconstructSendEvent(parcel *tapfreighter.OutboundParcel,
+	timestamp time.Time) *tapfreighter.AssetSendEvent {
+
+	return tapfreighter.NewHistoricalAssetSendEvent(parcel, timestamp)
```
guggero: nit: do we need this wrapper? Or can't we just call into the `tapfreighter` function directly? Or is this just to prevent lines being too long?

itest/addrs_test.go
```diff
@@ -1054,6 +1061,18 @@ func sendAssetsToAddr(t *harnessTest, sender *tapdHarness,
 	return sendAsset(t, sender, withReceiverAddresses(receiverAddrs...))
 }
 
+// sendAssetsToAddrWithLabel sends assets to the given addresses with a specific
+// label.
+func sendAssetsToAddrWithLabel(t *harnessTest, sender *tapdHarness,
+	label string, receiverAddrs ...*taprpc.Addr) (*taprpc.SendAssetResponse,
+	*EventSubscription[*taprpc.SendEvent]) {
+
+	return sendAsset(t, sender,
```
guggero: nit: formatting.

itest/send_test.go
```diff
@@ -1878,6 +1879,248 @@ func testSendNoCourierUniverseImport(t *harnessTest) {
...
+	const ntfsTimeout = 10 * time.Second
+
+	// Use the utility function to assert we receive all historical events.
+	assertAssetNtfsEvent(
+		t, events, ntfsTimeout, eventSelector, numTransfers,
+	)
+
+	// Test that events are ordered chronologically.
+	for i := 1; i < len(historicalEvents); i++ {
+		require.True(t.t,
```
guggero: nit: formatting. or use `GreaterOrEqual`?

tapdb/assets_store.go
bhandras: done!

rpcserver.go
bhandras: Good catch, we don't! It' just a leftover that i forgot to remove. done.

itest/addrs_test.go
bhandras: done.

itest/send_test.go
bhandras: done.

tapdb/sqlc/queries/transfers.sql
bhandras: Something I learned while trying to build these optional filters is that `NULL = NULL` evaluates to `UNKNOWN` which makes the `COALESCE` version not work correctly. Asked GPT for a minimal example to demonstrate it (you can just copy-paste it to sqlite):

```
CREATE TABLE t (
  id  INT PRIMARY KEY,
  col TEXT    -- this column can be NULL
);


-- Insert one non-NULL value and one NULL value
INSERT INTO t(id, col) VALUES
  (1, 'foo'),
  (2, NULL);

-- 1) COALESCE version
-- Expected result: only id=1 is returned, because
--   for id=2: NULL = COALESCE(NULL, NULL) → NULL = NULL → UNKNOWN → filtered out
WITH params(param) AS (
  VALUES (CAST(NULL AS TEXT))
)
SELECT
  t.*
FROM t
CROSS JOIN params
WHERE col = COALESCE(params.param, col)
ORDER BY id;

-- 2) OR‑IS‑NULL version
-- Expected result: both id=1 and id=2 are returned, because
--   params.param IS NULL → short‑circuits the WHERE to TRUE for every row
WITH params(param) AS (
  VALUES (CAST(NULL AS TEXT))
)
SELECT
  t.*
FROM t
CROSS JOIN params
WHERE (col = params.param OR params.param IS NULL)
ORDER BY id;
```

tapdb/assets_store.go
bhandras: Squashed the two commits to avoid confusion.

tapdb/sqlc/queries/transfers.sql
guggero: Ah, okay, good to know... Thanks for the explanation.

rpcserver.go
```diff
@@ -4461,6 +4461,69 @@ func (r *rpcServer) SubscribeSendEvents(req *taprpc.SubscribeSendEventsRequest,
...
+		startTime := time.Unix(0, req.StartTimestamp*1000)
+
+		// Validate that the timestamp is not in the future.
+		if startTime.After(time.Now()) {
+			return fmt.Errorf("start_timestamp cannot be in the " +
+				"future")
+		}
+
+		// Query completed transfers from the database with filters.
+		ctx := context.Background()
```
GeorgeTsagk: could use timeout here (default of 30s sounds reasonable)

bhandras: That's a good idea, done!

PR #1691, opened 2025-07-30
Improve grouped asset UX for onchain and offchain balances

### LINKED ISSUE #1523: [bug]: `ListGroups` should show a total amount for the group
[ListGroups](https://lightning.engineering/api-docs/api/taproot-assets/taproot-assets/list-groups/) should show a total amount for the group. Right now, a user of this RPC needs to manually add up the amount of all `asset_id` in the group.

If I run `lncli listchannels`, I get a total `capacity` and `local_balance` for the off chain amount of the group, which is defined at

https://github.com/lightninglabs/taproot-assets/blob/70b62f2b69522ac7ee5c1c9462ae30ded6c3bdb4/tapchannelmsg/custom_channel_data.go#L79-L85

I think we should have the same thing for the on chain amount.

### LINKED ISSUE #1686: [feature]: add unconfirmed_transfers field to `ListBalances` RPC
Spun out of this comment: https://github.com/lightninglabs/taproot-assets/pull/1587#issuecomment-3132422101

Unconfirmed change outputs from a transfer can't be spent without being confirmed first.
That's why assets "disappear" from the `ListBalances` output after a transfer and only reappear once the transaction has confirmed.

For the `ListAssets` call we have added the [`unconfirmed_transfers`](https://lightning.engineering/api-docs/api/taproot-assets/taproot-assets/list-assets/#taprpclistassetresponse) field as a hint to the user that things might not appear in the list while transfers are in-flight.

We should add the same `unconfirmed_transfers` field to the output of `ListBalances` RPC as well to hint to the user that only confirmed assets appear in the response. And that also gives us a field to document that behavior in the API docs.

guggero:
Fixes #1686 by adding a new field `unconfirmed_transfers` to the output of
the `ListBalances` RPC call to indicate that unconfirmed change doesn't
show up in the balance.

To make it more apparent if an asset belongs to a group, we add the
group key to the output of `ListBalances`.

Fixes #1523.
To make it easy for a user to see what their channel balance by group
key is, we add two new map entries, one for the open channels and one
for the pending channels that will show the local and remote balances
sumed up by group key.


ZZiigguurraatt:
+1 for the release notes!

docs/release-notes/release-notes-0.7.0.md
```diff
@@ -79,6 +85,10 @@
 
 ## Functional Updates
 
+- The output of `lncli channelbalance` [now also shows the local and remote
+  balances of asset channels grouped by group key (if grouped assets were used
```
ZZiigguurraatt, line 103: This is giving the sum for each group key for all channels combined together?

ZZiigguurraatt: What about a non-grouped asset, why don't you add totals for those too?

guggero: Yes, summed by group key. Non-grouped assets have been doing exactly that since the beginning, they're grouped/summed by asset ID.

ZZiigguurraatt:
> Fixes #1523.
> To make it easy for a user to see what their channel balance by group
> key is, we add two new map entries, one for the open channels and one
> for the pending channels that will show the local and remote balances
> sumed up by group key.

As mentioned in https://github.com/lightninglabs/taproot-assets/issues/1523#issuecomment-3136122357, This was not actually my intent of that issue, but I think it is a good thing to add.

docs/release-notes/release-notes-0.7.0.md
ZZiigguurraatt: > Non-grouped assets have been doing exactly that since the beginning, they're grouped/summed by asset ID.

what field is that in in the protobuf?

guggero: There is no protobuf for this. All the data necessary is encoded in the `custom_channel_data` field as a binary blob and then converted to JSON on the fly. The JSON definitions are here: https://github.com/lightninglabs/taproot-assets/blob/628e5ac07fa85f78a01b85d168a5e4149a7271e6/rfqmsg/custom_channel_data.go#L84

ZZiigguurraatt: OK, but I don't see any total sums for all channels for each asset_id.

guggero: What output _do_ you see?

guggero:
Rebased.

PR #1693, opened 2025-07-31
universerpc: add FetchSupplyLeaves RPC for verifier sync

ffranr:
This PR introduces a new RPC, `FetchSupplyLeaves`. This endpoint allows a supply commitment verifier to sync all issuance, burn, and ignore leaves for a given asset group. With these leaves, a verifier can reconstruct the supply commitment tree snapshot (at a given block height), allowing them to independently verify supply commitments made by asset issuers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1693)


universe/supplycommit/env.go
```diff
@@ -87,6 +87,19 @@ func (s SupplyTrees) FetchOrCreate(treeType SupplySubTree) mssmt.Tree {
 	return tree
 }
 
+// FetchSupplyLeavesResp is the response type for fetching the supply
+// leaves for a given asset specifier.
+type FetchSupplyLeavesResp struct {
```
guggero: nit: Just call this `SupplyLeaves`? Do we need to know there was a fetch call and this is the response? IMO not, we just need to know what's _in_ the type.

taprpc/universerpc/universe.proto
```diff
@@ -179,6 +179,14 @@ service Universe {
     */
     rpc FetchSupplyCommit (FetchSupplyCommitRequest)
         returns (FetchSupplyCommitResponse);
+
+    /* tapcli: `universe fetchsupplyleaves`
```
guggero: nit: do we need the `fetch` prefix in the command name? IMO just `supplyleaves` would probably be enough?

cmd/commands/universe.go
```diff
@@ -1696,6 +1697,54 @@ func fetchSupplyCommit(ctx *cli.Context) error {
 	return nil
 }
 
+var fetchSupplyLeavesCmd = cli.Command{
+	Name: "fetchleaves",
```
guggero: nit: doesn't match command name in API docs prefix of the proto file. IMO this should just be `supplyleaves`.

guggero (approved):
Nice, LGTM :tada:

tapdb/supply_tree.go
```diff
@@ -616,6 +616,21 @@ func fetchSupplyLeavesByHeight(ctx context.Context, db BaseUniverseStore,
 	groupKey btcec.PublicKey, startHeight,
 	endHeight uint32) (*supplycommit.SupplyLeaves, error) {
 
+	// Convert the start and end heights to SQL nullable integers.
+	// If the height is zero, we use a null value to indicate no limit.
+	var (
+		sqlStartHeight sql.NullInt32
+		sqlEndHeight   sql.NullInt32
```
Roasbeef, line 623: How does this change affect behavior? Wasn't it roughly identical before?

universe/supplycommit/multi_sm_manager.go
```diff
@@ -377,6 +377,25 @@ func (m *MultiStateMachineManager) FetchCommitment(ctx context.Context,
 	}), nil
 }
 
+// FetchSupplyLeavesByHeight returns the set of supply leaves for the given
+// asset specifier within the specified height range.
+func (m *MultiStateMachineManager) FetchSupplyLeavesByHeight(
```
Roasbeef, line 382: If we're exposing this for the RPC server, then can't we just give it a slim interface based on `m.cfg.TreeView`? Nothing about this method actually relates to the state machine themselves, we're only exposing a new read method.

taprpc/universerpc/universe.proto
```diff
@@ -868,4 +876,49 @@ message FetchSupplyCommitResponse {
...
+    // The end block height for the range of supply leaves to fetch.
+    uint32 block_height_end = 4;
+}
+
+message SupplyLeafKey {
+    // The outpoint associated with the supply leaf.
+    Outpoint outpoint = 1;
+
+    // The script key of the supply leaf. This is the script key of the asset
+    bytes script_key = 2;
```
Roasbeef, line 908: So for the ignore tree, the key is actually the hash of the prev ID, which includes the asset ID.

Roasbeef: Similarly, we now use `AssetLeafKey` for the ignore and supply tree to fix prior issues related to collisions.

Roasbeef: This is important, as the client needs to reconstruct the tree exactly as the issuer created it.

universe/ignore_records.go
```diff
@@ -311,7 +311,7 @@ func (i *SignedIgnoreTuple) UniverseKey() [32]byte {
 func (i *SignedIgnoreTuple) LeafScriptKey() asset.ScriptKey {
 	scriptKeyBytes := i.IgnoreTuple.Val.ScriptKey
 
-	keyPub, _ := btcec.ParsePubKey(scriptKeyBytes.SchnorrSerialized())
```
Roasbeef, line 314: Hmm, shows a lack of test coverage for this method?

Roasbeef (changes requested):

rpcserver.go
```diff
@@ -4260,6 +4271,134 @@ func (r *rpcServer) FetchSupplyCommit(ctx context.Context,
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal supply "+
+				"leaf entry: %w", err)
+		}
+
+		rpcBurnLeaves = append(rpcBurnLeaves, rpcLeaf)
+	}
+
+	// Marshal ignore supply leaves into the RPC format.
+	var rpcIgnoreLeaves []*unirpc.SupplyLeafEntry
```
Roasbeef: We can pre-allocate this, and all the others above.

tapdb/supply_tree.go
ffranr: The SQL query `QuerySupplyLeavesByHeight` has the following `WHERE` clause:

```
WHERE
    leaves.leaf_node_namespace = @namespace AND
    (leaves.block_height >= sqlc.narg('start_height') OR sqlc.narg('start_height') IS NULL) AND
    (leaves.block_height <= sqlc.narg('end_height') OR sqlc.narg('end_height') IS NULL);
```

Previously, passing an end height of `0` to `fetchSupplyLeavesByHeight` resulted in `sqlInt32(0)`, which is non-null and therefore applied a limiting condition. This change allows us to pass a null end height to the SQL query, effectively removing the upper bound.

universe/supplycommit/multi_sm_manager.go
ffranr: > Nothing about this method actually relates to the state machine themselves, we're only exposing a new read method.

I agree.

But I'm not sure what you mean by: "can't we just give it a slim interface based on `m.cfg.TreeView`". Do you mean that I shouldn't make this call in the RPC server endpoint?
```
resp, err := r.cfg.SupplyCommitManager.FetchSupplyLeavesByHeight
```

That I should create a new component rather than re-use `r.cfg.SupplyCommitManager` for reading issuer supply leaves?

universe/ignore_records.go
ffranr: I've add assertion in itest for script key in the latest commits.

taprpc/universerpc/universe.proto
ffranr: I've added commits that modify the `SupplyUpdateEvent` interface. I believe the changes are correct, but let me know if I’ve misunderstood anything. This update allows the RPC endpoint to return the asset ID from the supply leaf key.

tapdb/supply_tree.go
Roasbeef: Gotcha, thanks for the great explanation. 

I think we should add a test to show that if neither was sent, then before we would potentially not return _any_ results.

universe/supplycommit/multi_sm_manager.go
Roasbeef: I mean giving the RPC server `m.cfg.TreeView` directly, or just a single method interface. As there's nothing here that actually uses the target struct (`m`) beyond the cfg access.

taprpc/universerpc/universe.proto
Roasbeef: Great. 

What I meant is that in order to be able to _insert_ this item into the local tree of the verifier, it needs the asset ID as well. Eg see this for the burn key: https://github.com/lightninglabs/taproot-assets/blob/57da71b99320922e23217c2caed3872807ce2edf/tapdb/burn_tree.go#L284-L290

tapdb/supply_tree.go
ffranr: I've added unit tests in a new commit.

Roasbeef (approved):
LGTM 🏓

> Reviewed 1 of 3 files at r1, 19 of 19 files at r2, 17 of 17 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1693#-:-OX-gEd34D0p2z1BeMwx:b-lil9uq)* status: all files reviewed, 8 unresolved discussions (waiting on @ffranr)

PR #1702, opened 2025-08-04
universe: emit SyncDiffEvent and add event subscription support to SimpleSyncer

ffranr:
This PR adds event subscription support to `SimpleSyncer`, allowing it to emit a `SyncDiffEvent` after a sync completes and a diff is present. This enables downstream components to react to sync changes. Also includes a rename of `auto_syncer.go` to `federation.go` for clarity.


universe/syncer.go
```diff
@@ -540,7 +592,37 @@ func (s *SimpleSyncer) SyncUniverse(ctx context.Context, host ServerAddr,
...
+
+	// Notify subscribers of the new sync diffs where the old and
+	// new Universe roots differ.
+	var events []fn.Event
+	for idx := range diffs {
+		diff := diffs[idx]
+
+		// Only notify subscribers if the old and new Universe roots
+		// differ.
+		syncChange := diff.OldUniverseRoot.NodeHash() !=
```
guggero: Looks like we also need to check for nil-ness of old and new universe roots.

guggero (changes requested):
Just saw the CI failure.

guggero (approved):
Nice, LGTM 🎉

universe/syncer.go
```diff
@@ -55,15 +55,47 @@ type SimpleSyncer struct {
 	// Universe with a remote Universe. This is used to prevent concurrent
 	// syncs.
 	isSyncing atomic.Bool
+
+	// eventDistributor is used to distribute sync events to subscribers.
+	eventDistributor *fn.EventDistributor[fn.Event]
```
Roasbeef, line 80: Perhaps we can use a more precise type here instead of `fn.Event`?

universe/syncer.go
```diff
@@ -572,7 +592,41 @@ func (s *SimpleSyncer) SyncUniverse(ctx context.Context, host ServerAddr,
...
+				events,
+				&SyncDiffEvent{
+					timestamp: time.Now().UTC(),
+					SyncDiff:  diff,
+				},
+			)
+		}
+	}
+
+	s.eventDistributor.NotifySubscribers(events...)
```
Roasbeef, line 627: Assuming we're planning to use this to drive events we send to the supply commit state machines:
  * This only covers the new issuance events. 
  * Burn events won't show up here in the normal universe trees, nor will ignore events. 

We should also consider persistence and durability here. As an example: what happens if we emit this event, but then crash before it reaches a supply commit state machine? Some sort of log may be in order here. 

For burn+ignore, I think this new supply commit syncer will need to rely on polling for now against the specified universe sever.

Roasbeef (changes requested):

universe/syncer.go
ffranr: I was following the pattern used in `ChainPorter.subscribers`, which uses the `fn.Event` type. That allows us to pass an arbitrary number of event types to subscribers.

I like the flexibility we have with the current approach. What more precise type did you have in mind?

universe/syncer.go
ffranr: I added this feature to the syncer so the supply verifier could be notified of new issuance events and track the corresponding pre-commitment output spend. The supply verifier should still be responsible for syncing supply leaves (burns, ignores).

This gives the verifier a chance to ensure all issuance tranches are tied into the supply commitment spend pattern.

ffranr: > We should also consider persistence and durability here.

I agree, good point. I'd prefer to tackle that as follow-up work once the supply verifier is in place. I want to confirm the overall architecture makes sense before focusing on robustness.

One option could be to add `deliverExisting` support to `SimpleSyncer.RegisterSubscriber`, starting from a given block height. At that point, instead of emitting a sync diff event (which may be hard to reconstruct), we could emit the synced leaves directly (issuance/transfer). The supply verifier would then need to memoize the relevant block height.

PR #1706, opened 2025-08-05
Explore reusing syncer for supply leaf syncing

ffranr:
The current syncer supports syncing asset issuance and transfer proofs. This PR explores whether it can also be reused for syncing universe supply leaves and their associated proofs.


ffranr:
Will implement a new syncer in `supplyverifier` instead.

Issue #875, opened 2024-04-11
[feature]: Paginate `ListAssets` RPC


GeorgeTsagk:
## Description

`ListAssets` rpc call will return all assets owned by the daemon. This list can grow quite large, and should be paginated to avoid future bottlenecks.

Pagination should adhere to the same params that we use in already paginated calls throughout our RPC. (`offset` / `limit` / `direction`)

jharveyb:
User request here #994

ZZiigguurraatt:
I think we should clarify here that not only area all assets listed, but all vUTXO/tUTXO (not sure what they are called) for all assets. So, I think the problem is a bit bigger.

Also, I think we should add pagination to [ListGroups](https://lightning.engineering/api-docs/api/taproot-assets/taproot-assets/list-groups/), [ListUtxos](https://lightning.engineering/api-docs/api/taproot-assets/taproot-assets/list-utxos/) , and [AddrReceives](https://lightning.engineering/api-docs/api/taproot-assets/taproot-assets/addr-receives/) as well because they suffer the same problem.

ZZiigguurraatt:
Just to provide a little data: If I have 250 asset UTXO then it takes about 1 second to run `ListAssets`. I'm not sure why it takes that long but it kind of re-enforces the need for this issue and https://github.com/lightninglabs/taproot-assets/issues/1699 if this is going to scale linearly.

PR #1707, opened 2025-08-05
tapdb: add pagination support for ListAssets RPC

Roasbeef:
This PR implements pagination for the ListAssets RPC call, addressing issue #875. The implementation follows the existing pagination patterns used in the Universe RPCs and maintains full backwards compatibility with existing clients.

The motivation for this change comes from performance issues when dealing with large asset collections. Previously, fetching 250 UTXOs could take over a second, which becomes problematic as users accumulate more assets over time. With pagination, clients can fetch results incrementally, improving both server performance and user experience.

The implementation adds offset, limit, and sort direction parameters to ListAssetRequest, along with total_count and has_more metadata in the response. Under the hood, we've added efficient SQL queries that handle pagination at the database level rather than loading everything into memory first. The default page size is set to 100 items when no limit is specified, ensuring existing clients continue to work without modification.

I've also refactored the SortDirection enum to live in tapcommon.proto to avoid circular dependencies between our proto files. This allows both the universe and taproot-assets services to share the same pagination types.

The test suite covers various pagination scenarios including edge cases like empty results, last page boundaries, and filtering combinations. The tests verify that paginated results match what you'd get from fetching everything at once, ensuring consistency across different access patterns.

One thing to note is that in-memory filtering for unconfirmed mints is still required at the RPC layer since these aren't stored in the database yet. This is a minor limitation but doesn't affect the main pagination functionality for confirmed assets.

### DIFF
--- a/cmd/commands/assets.go
+++ b/cmd/commands/assets.go
@@ -745,6 +745,21 @@ var listAssetsCommand = cli.Command{
 				"key type; cannot be used at the same time " +
 				"as --" + scriptKeyTypeName,
 		},
+		cli.IntFlag{
+			Name:  "offset",
+			Usage: "the number of results to skip (for pagination)",
+			Value: 0,
+		},
+		cli.IntFlag{
+			Name:  "limit",
+			Usage: "the maximum number of results to return",
+			Value: 100,
+		},
+		cli.StringFlag{
+			Name:  "direction",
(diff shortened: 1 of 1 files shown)


darioAnongba:
Implemented in https://github.com/lightninglabs/taproot-assets/pull/1995.

PR #1708, opened 2025-08-05
Add `SupplySyncer` and `RpcSupplySync` for syncing supply leaves

ffranr:
This PR introduces the components necessary for syncing supply commit leaves from remote universe servers:

* `SupplySyncer`: Handles the syncing of supply leaves from remote peers.
* `RpcSupplySync`: Acts as a bridge between the RPC endpoints for leaf fetching and the internal supply verifier syncer.

The synced leaves will be verified by the supply verifier, which will be added in a future commit.

### DIFF
--- a/universe/supplyverify/syncer.go
+++ b/universe/supplyverify/syncer.go
@@ -1,0 +1,116 @@
+package supplyverify
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/lightninglabs/taproot-assets/asset"
+	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/mssmt"
+	"github.com/lightninglabs/taproot-assets/universe"
+	"github.com/lightninglabs/taproot-assets/universe/supplycommit"
+)
+
+// SupplyLeafFetcher is an interface that can be used to fetch information from
+// a Universe.
+type SupplyLeafFetcher interface {
+	// FetchSupplyLeaves fetches the supply leaves for a specific asset
+	// group within a specified block height range.
+	FetchSupplyLeaves(ctx context.Context,
+		assetSpec asset.Specifier, startBlockHeight,
+		endBlockHeight fn.Option[uint32]) (supplycommit.SupplyLeaves,
+		error)
+
+	// Close closes the fetcher and cleans up any resources.
+	Close() error
+}
+
+// SupplyLeafFetcherFactory is a function type that creates SupplyLeafFetcher
+// instances for a given universe server address.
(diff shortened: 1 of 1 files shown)


supplysync_rpc.go
```diff
@@ -0,0 +1,184 @@
...
+			return zero, fmt.Errorf("unable to unmarshal burn "+
+				"event: %w", err)
+		}
+
+		supplyLeaves.BurnLeafEntries = append(
+			supplyLeaves.BurnLeafEntries, *burnEvent,
+		)
+	}
+
+	// Unmarshal ignore leaves.
```
Roasbeef, line 108: Superflous comments here and other decode instances above.

universe/supplyverify/syncer.go
```diff
@@ -0,0 +1,116 @@
...
+type SupplyLeafFetcher interface {
+	// FetchSupplyLeaves fetches the supply leaves for a specific asset
+	// group within a specified block height range.
+	FetchSupplyLeaves(ctx context.Context,
+		assetSpec asset.Specifier, startBlockHeight,
+		endBlockHeight fn.Option[uint32]) (supplycommit.SupplyLeaves,
+		error)
+
+	// Close closes the fetcher and cleans up any resources.
+	Close() error
```
Roasbeef, line 25: Nice addition. I think we at times forget to do this elsewhere?

universe/supplyverify/syncer.go
```diff
@@ -0,0 +1,116 @@
...
+
+// SupplyLeafFetcherFactory is a function type that creates SupplyLeafFetcher
+// instances for a given universe server address.
+type SupplyLeafFetcherFactory func(
+	serverAddr universe.ServerAddr) (SupplyLeafFetcher, error)
+
+// SupplyLeafStore is an interface for storing supply commitment leaves.
+type SupplyLeafStore interface {
+	// ApplySupplyUpdates stores a batch of supply update events to the
+	// database without requiring a transition.
```
Roasbeef, line 36: Without requiring one? As in it'll insert them on disk as supply updates? 

If it's just inserting them as updates, and not trying to construct the final tree (yet), then what's it returning as the node here?

ffranr:
Close in favour of https://github.com/lightninglabs/taproot-assets/pull/1674

universe/supplyverify/syncer.go
ffranr: I've modified this a bit. My idea here is just to write to the db. And then the verifier can read from that and reconstruct the supply commitment itself.

PR #1710, opened 2025-08-06
Delete channel alias scids when quotes expire

GeorgeTsagk:
## Description

Previously when we'd garbage collect quotes we'd only delete the quote entry locally on tapd. What we didn't do is clear the state that was produced in LND, which keeps track of alias scid -> base scid mappings to facilitate routing.

We use the new method `XFindBaseLocalChanAlias` to retrieve the base scid that corresponds to our quotes, and then we call LND to delete the alias.

Depends on:
https://github.com/lightningnetwork/lnd/pull/10133
https://github.com/lightninglabs/lndclient/pull/236

### DIFF
--- a/rfq/order.go
+++ b/rfq/order.go
@@ -968,7 +968,7 @@ func (h *OrderHandler) Start(ctx context.Context) error {
 			return
 		}
 
-		// Start the main event loop in a separate goroutine.
+		// Start the HTLC interceptor in a separate go routine.
 		h.Wg.Add(1)
 		go func() {
 			defer h.Wg.Done()
@@ -990,6 +990,12 @@ func (h *OrderHandler) Start(ctx context.Context) error {
 
 				return
 			}
+		}()
+
+		// Start the main event loop in a separate go routine.
+		h.Wg.Add(1)
+		go func() {
+			defer h.Wg.Done()
 
 			h.mainEventLoop()
 		}()


rfq/order.go
```diff
@@ -880,6 +884,12 @@ func (h *OrderHandler) Start() error {
 					"interception: %v", startErr)
 				return
 			}
+		}()
+
+		// Start the main event loop in a separate go routine.
+		h.Wg.Add(1)
+		go func() {
+			defer h.Wg.Done()
```
ffranr, line 998: we could use `h.ContextGuard.Goroutine` here (and above if you like)

GeorgeTsagk:
Need to bump configured go version to `1.23.10` since this [LND PR](https://github.com/lightningnetwork/lnd/pull/9751) also bumps it

GeorgeTsagk:
For the LiT tests I'll have to make a PR that bumps all versions that are included with this PR
https://github.com/lightninglabs/taproot-assets/actions/runs/16778944176/job/47512051701?pr=1710#step:10:1022

rfq/order.go
jtobin: TIL go1.25.0 apparently also adds a [WaitGroup.Go method](https://pkg.go.dev/sync#WaitGroup.Go) for handling the Add(1)/Done pattern.

jtobin (approved):
This LGTM. I take it the LiT tests should be updated *after* this one is merged?

GeorgeTsagk:
> This LGTM. I take it the LiT tests should be updated _after_ this one is merged?

It's a dep bump issue, I believe currently the dependency is bumped in the feature bits itest PR, but maybe worth separating as we don't know when that is getting merged.

GeorgeTsagk:
Rebased on `0-8-0-staging`

rfq/order.go
```diff
@@ -1112,9 +1122,41 @@ func (h *OrderHandler) cleanupStalePolicies() {
...
-				staleCounter++
-				h.policies.Delete(scid)
+			if !policy.HasExpired() {
+				return nil
+			}
+
+			staleCounter++
+
+			// Delete the local entry of this policy.
+			h.policies.Delete(scid)
```
ffranr, line 1265: not a big deal in this case but: slightly better if we delete from the local cache last; otherwise, we might lose our reference to the resource? the cleanup routine could appear to be working from tapd's POV (few log messages without re-attempt), but in reality the cleanup could be failing.

jtobin:
(This one needs a retarget at main, and a rebase.)

litbot-9000:

rfq/order.go
GeorgeTsagk: > we might lose our reference to the resource

In this case the resource has been invalidated, so deleting from in-memory first seems to me as the safest order of operations:
i) mark resource as invalid (i.e deletion) as soon as possible, prohibiting any other system grabbing and using the resource (in our context that could be some undesirable routing that uses an expired/invalid quote)
ii) proceed with tail emission optimizations -> what this PR is introducing, which is persisting the deletion of the resource. If this fails we just have temporary unecessary usage of storage.

jtobin:
Think this one needs a rebase to trigger the updated CI jobs.

rfq/manager.go
```diff
@@ -63,6 +63,10 @@ type ScidAliasManager interface {
 	// Manager's maps.
 	DeleteLocalAlias(ctx context.Context, alias,
 		baseScid lnwire.ShortChannelID) error
+
+	// FindBaseAlias finds the base channel ID for a given alias.
+	FindBaseAlias(ctx context.Context,
```
ffranr: Please add a note to the doc comment explaining that the "base channel ID" is the real/canonical SCID.
Also, instead of "find" I think we should use Fetch.

rfq/order.go
```diff
@@ -932,6 +936,12 @@ func (h *OrderHandler) Start() error {
 					"interception: %v", startErr)
 				return
 			}
+		}()
+
+		// Start the main event loop in a separate go routine.
+		h.Wg.Add(1)
+		go func() {
```
ffranr, line 997: why is this code change part of this commit?

PR #1711, opened 2025-08-06
itest: increase proof courier timeout

guggero:
Fixes a flake where uploading proofs to the proof courier would take more than 50 milliseconds in CPU constrained environments such as CI.


itest/tapd_harness.go
```diff
@@ -274,7 +274,7 @@ func newTapdHarness(t *testing.T, ht *harnessTest, cfg tapdConfig,
 	}
 	finalCfg.UniverseRpcCourier = &proof.UniverseRpcCourierCfg{
 		BackoffCfg:            &universeRpcBackoffCfg,
-		ServiceRequestTimeout: 50 * time.Millisecond,
```
GeorgeTsagk, line 277: That was quite a short timeout in the first place, regardless of this flaking in CI or not

jtobin: +1

PR #1715, opened 2025-08-08
itest: fix flake in address v2 test

guggero:
We always expect two events. So asserting for just one only worked when things ran very quickly and only the first one was persisted. Always waiting for two will fix this timing-dependent flake.

Fixes the flake observed here: https://github.com/lightninglabs/taproot-assets/actions/runs/16798226160/job/47573276733?pr=1713


jtobin (approved):
LGTM

Issue #1684, opened 2025-07-28
Implement IgnoreChecker for Supply Commitment System


Roasbeef:
## Context

PR #1674 is introducing a supply commitment verifier system that tracks asset supply through three sub-trees: mint, burn, and ignore. Currently, the `IgnoreChecker` interface (defined in `proof/verifier.go`) only has a mock implementation, and no production `IgnoreChecker` instance is passed during proof verification.

## Problem

The proof verification system lacks integration with the supply commitment ignore tree. This means:
- Proofs for assets in the ignore tree are still fully validated (performance impact)
- No enforcement of supply commitment ignore rules during proof verification
- Inconsistency between supply tracking and proof validation systems

## Proposed Solution

Create a concrete `IgnoreChecker` implementation that queries the ignore sub-tree of the supply commitment system. This will automatically reject proofs for assets that have been marked as ignored in the supply tree.

### Technical Details

1. **Interface to implement**:
   ```go
   type IgnoreChecker interface {
       IsIgnored(prevID AssetPoint) bool
   }
   ```

2. **Integration points** where `VerifierCtx` needs updating:
   - `rpcserver.go` - RPC proof verification
   - `tapfreighter/chain_porter.go` - Chain porter proof handling
   - `tapchannel/aux_funding_controller.go` - Channel funding proof verification
   - Any other locations creating `VerifierCtx` instances

3. **Architecture**:
   - New file: `tapdb/supply_ignore_checker.go`
   - Query the ignore sub-tree from `tapdb/ignore_tree.go`
   - Wire up supply tree store access to verification contexts

## Implementation Steps

1. [ ] Create `SupplyIgnoreChecker` struct implementing `IgnoreChecker`
2. [ ] Add method to query ignore tree for asset points
3. [ ] Update all `VerifierCtx` creation sites to include the new checker
4. [ ] Add unit tests for the new implementation
5. [ ] Add integration tests verifying ignored assets are rejected
6. [ ] Update documentation for the new behavior

## Benefits

- **Performance**: Skip expensive proof validation for known invalid/ignored assets
- **Consistency**: Align proof verification with supply commitment rules
- **Security**: Enforce supply commitment policies during all proof operations
- **Maintainability**: Single source of truth for ignored assets

## References

- PR #1674 - Supply commitment verifier implementation
- Issue #1463 - Universe supply tracking feature
- `proof/verifier.go` - IgnoreChecker interface definition
[...]

PR #1716, opened 2025-08-08
proof: implement and hook up ignore checker

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1684.

Implements and hooks up a caching ignore tree checker.

Opening as draft because some of the caching isn't fully finished yet (mainly the negative case LRU cache).

### DIFF
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -769,6 +769,12 @@ func (id PrevID) Hash() [sha256.Size]byte {
 	return *(*[sha256.Size]byte)(h.Sum(nil))
 }
 
+// String returns a human-readable description of the PrevID.
+func (id PrevID) String() string {
+	return fmt.Sprintf("PrevID(outpoint=%s, id=%s, script_key=%x)",
+		id.OutPoint.String(), id.ID.String(), id.ScriptKey.CopyBytes())
+}
+
 // AnchorPoint is a type alias for an asset anchor outpoint. It relates the
 // on-chain anchor outpoint (txid:vout) to the corresponding committed asset.
 type AnchorPoint = PrevID
--- a/itest/supply_commit_test.go
+++ b/itest/supply_commit_test.go
@@ -196,8 +196,10 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
 	// Determine the transfer output owned by the secondary node.
 	// This is the output that we will ignore.
 	transferOutput := sendResp.RpcResp.Transfer.Outputs[0]
+	changeOutput := sendResp.RpcResp.Transfer.Outputs[1]
 	if sendResp.RpcResp.Transfer.Outputs[1].Amount == sendAssetAmount {
 		transferOutput = sendResp.RpcResp.Transfer.Outputs[1]
+		changeOutput = sendResp.RpcResp.Transfer.Outputs[0]
 	}
 
 	// Get block height at the time of the ignore request.
@@ -217,6 +219,21 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
 	require.NotNil(t.t, respIgnore)
 	require.EqualValues(t.t, sendAssetAmount, respIgnore.Leaf.RootSum)
 
+	// We also ignore our change output, so we can later verify that the
+	// proof verifier correctly denies spending the change output.
+	ignoreReq2 := &unirpc.IgnoreAssetOutPointRequest{
+		AssetOutPoint: &taprpc.AssetOutPoint{
+			AnchorOutPoint: changeOutput.Anchor.Outpoint,
+			AssetId:        rpcAsset.AssetGenesis.AssetId,
+			ScriptKey:      changeOutput.ScriptKey,
+		},
+		Amount: sendChangeAmount,
+	}
+	respIgnore2, err := t.tapd.IgnoreAssetOutPoint(ctxb, ignoreReq2)
+	require.NoError(t.t, err)
+	require.NotNil(t.t, respIgnore2)
+	require.EqualValues(t.t, sendChangeAmount, respIgnore2.Leaf.RootSum)
+
 	// Assert that the mempool is empty.
 	mempool := t.lndHarness.Miner().GetRawMempool()
 	require.Empty(t.t, mempool)
@@ -232,6 +249,7 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
 			},
 			IgnoreLeafKeys: [][]byte{
 				respIgnore.LeafKey,
+				respIgnore2.LeafKey,
 			},
 		},
 	)
@@ -256,6 +274,7 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
 	minedBlocks := MineBlocks(t.t, t.lndHarness.Miner().Client, 1, 1)
 
 	t.Log("Fetch updated supply commitment")
+
 	// Ensure that the supply commitment reflects the ignored asset
 	// outpoint owned by the secondary node.
 	var fetchResp *unirpc.FetchSupplyCommitResponse
@@ -268,6 +287,7 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
 				},
 				IgnoreLeafKeys: [][]byte{
 					respIgnore.LeafKey,
+					respIgnore2.LeafKey,
 				},
 			},
 		)
@@ -283,7 +303,7 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
 		// Once the ignore tree includes the ignored asset outpoint, we
 		// know that the supply commitment has been updated.
 		return fetchResp.IgnoreSubtreeRoot.RootNode.RootSum ==
-			int64(sendAssetAmount)
+			int64(sendAssetAmount+sendChangeAmount)
 	}, defaultWaitTimeout, time.Second)
 
 	// Verify that the supply commitment tree commits to the ignore subtree.
@@ -310,7 +330,7 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
 
 	// Unmarshal ignore tree leaf inclusion proof to verify that the
 	// ignored asset outpoint is included in the ignore tree.
-	require.Len(t.t, fetchResp.IgnoreLeafInclusionProofs, 1)
+	require.Len(t.t, fetchResp.IgnoreLeafInclusionProofs, 2)
 	inclusionProofBytes := fetchResp.IgnoreLeafInclusionProofs[0]
 
 	// Verify that the ignore tree root can be computed from the ignore leaf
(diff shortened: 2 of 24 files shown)


guggero:
Added the negative LRU cache, hooked up the cache stats to our Prometheus metrics and implemented the cache invalidation.

tapdb/supply_tree.go
```diff
@@ -738,3 +766,40 @@ func (s *SupplyTreeStore) FetchSupplyLeavesByHeight(ctx context.Context,
 
 	return lfn.Ok(resp)
 }
+
+// FetchSupplyLeavesByType fetches all supply leaves for a given asset specifier
+// and a specific supply sub-tree.
+func (s *SupplyTreeStore) FetchSupplyLeavesByType(ctx context.Context,
+	spec asset.Specifier, tree supplycommit.SupplySubTree, startHeight,
```
Roasbeef: tree -> treeType ?

proof/verifier.go
```diff
@@ -45,7 +45,7 @@ type IgnoreChecker interface {
 	// IsIgnored returns true if the given prevID is known to be invalid. A
 	// prevID is used here, but the check should be tested against a proof
 	// result, or produced output.
-	IsIgnored(prevID AssetPoint) bool
+	IsIgnored(ctx context.Context, prevID AssetPoint) lfn.Result[bool]
```
Roasbeef, line 48: 👍

asset/asset.go
```diff
@@ -772,6 +772,12 @@ func (id PrevID) Hash() [sha256.Size]byte {
 	return *(*[sha256.Size]byte)(h.Sum(nil))
 }
 
+// String returns a human-readable description of the PrevID.
+func (id PrevID) String() string {
+	return fmt.Sprintf("PrevID(outpoint=%s, id=%s, script_key=%x)",
+		id.OutPoint.String(), id.ID.String(), id.ScriptKey.CopyBytes())
```
Roasbeef: `CopyBytes()` as an alternative to just slicing into the underlying byte slice?

tapdb/supply_ignore_checker.go
```diff
@@ -0,0 +1,224 @@
...
+	// If the asset group cannot be found, then ErrAssetGroupUnknown is
+	// returned.
+	QueryAssetGroupByID(context.Context, asset.ID) (*asset.AssetGroup,
+		error)
+}
+
+// IgnoreCheckerCfg is a configuration struct for the CachingIgnoreChecker.
+type IgnoreCheckerCfg struct {
+	// GroupQuery is used to query asset groups by their asset ID.
+	GroupQuery assetGroupQuery
```
Roasbeef: Hmm, if the config struct itself is publicly exported, then shouldn't the interfaces exported also be?

tapdb/supply_ignore_checker.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return lfn.Errf[bool]("unable to decode group key: %w", err)
+	}
+
+	// If we have a valid group key, we can fetch the latest ignore
+	// tuples for the asset group based on the best known height for the
+	// group (or 0 if we haven't encountered this group yet).
+	specifier := asset.NewSpecifierFromGroupKey(*groupPubKey)
+	bestHeight := c.bestHeightLookup[groupKey]
+	leaves, err := c.cfg.Store.FetchSupplyLeavesByType(
```
Roasbeef, line 296: One alternative here would be doing a more precise query just that that particular ignore point. On the other ahdn wit hthis approach, we do the broad look up once, then just cache it from there on.

tapdb/supply_ignore_checker.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// CachingIgnoreChecker is a proof.IgnoreChecker that caches the results of
+// whether an asset is ignored or not. It uses a map to store the group keys
+// for asset IDs, and another map to store the best known height for each group
+// key. It also maintains a map of ignored assets, which is populated on demand
+// when checking if an asset is ignored.
+// Assuming ~200k known assets and 1000 groups and 50k ignored asset points, the
+// memory requirement for these three maps is approximately:
+// - groupKeyLookup:     200k * (32+33) bytes   = ~12.4 MiB
```
Roasbeef, line 145: Pretty reasonable numbers. In the future we can start to make some/all of these LRU, or do stuff like store a truncated hash as the main key, etc.

rpcserver.go
```diff
@@ -10030,3 +9980,18 @@ func (r *rpcServer) RegisterTransfer(ctx context.Context,
 		RegisteredAsset: rpcAsset,
 	}, nil
 }
+
+// ProofVerifierCtx returns a proof.VerifierCtx that can be used to verify
+// proofs in the RPC server.
+func (r *rpcServer) ProofVerifierCtx(ctx context.Context) proof.VerifierCtx {
```
Roasbeef, line 10023: Nice addition 👍

tapcfg/server.go
```diff
@@ -167,6 +168,18 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
...
+	// Construct the supply tree database backend so we can create the
+	// ignore checker that's used by a number of early initialized
+	// components in here.
+	supplyTreeStore := tapdb.NewSupplyTreeStore(uniDB)
+	ignoreChecker := tapdb.NewCachingIgnoreChecker(tapdb.IgnoreCheckerCfg{
+		GroupQuery: tapdbAddrBook,
+		Store:      supplyTreeStore,
+	})
+
+	var ignoreCheckerIface proof.IgnoreChecker = ignoreChecker
```
Roasbeef: An alternative to adding an explicit type param to the `lfn.Some` call below?

tapfreighter/wallet.go
```diff
@@ -927,7 +950,12 @@ func (f *AssetWallet) SignVirtualPacket(vPkt *tappsbt.VPacket,
 			// Before we sign the transaction, we want to make sure
 			// the inclusion proof is valid and the asset is
 			// actually committed in the anchor transaction.
-			err := verifyInclusionProof(vPkt.Inputs[idx])
+			assetProof := vPkt.Inputs[idx].Proof
+			if assetProof == nil {
+				return nil, fmt.Errorf("input proof is nil")
+			}
+
+			_, err := assetProof.VerifyProofIntegrity(ctx, vCtx)
```
Roasbeef, line 958: Nice to do additional verification here as well.

itest/supply_commit_test.go
```diff
@@ -443,6 +468,27 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
...
+		AssetId: rpcAsset.AssetGenesis.AssetId,
+		Amt:     sendChangeAmount / 2,
+	})
+	require.NoError(t.t, err)
+	sendAsset(
+		t, t.tapd, withReceiverAddresses(bobAddr),
+		withError("is ignored"),
+	)
+
+	// TODO(ffranr): The above only tests that the node that issued the
```
Roasbeef, line 484: Can make an issue for this 👍

tapdb/supply_ignore_checker.go
```diff
@@ -3,8 +3,10 @@ package tapdb
 import (
 	"context"
 	"errors"
+	"fmt"
```
Roasbeef, line 7: Re the commit message: do we need a supply verifier to invalidate it? Couldn't we just invalidate it selectively when/if an asset in that group is ignored?

universe/supplycommit/transitions.go
```diff
@@ -984,6 +984,11 @@ func (c *CommitFinalizeState) ProcessEvent(event Event,
 				"dangling updates: %w", err)
 		}
 
+		// We know our tree has been updated, so we need to make sure
+		// the negative lookup cache of the ignore checker is flushed
+		// and the new ignore leaves are loaded from disk.
+		env.IgnoreCheckerCache.InvalidateCache()
```
Roasbeef: Couldn't we just _selectively_ invalidate the cache for this particular group key? 

Even before that, we can check to see if this batch adds any new ignored assets, only invalidating if so.

Roasbeef (changes requested):

asset/asset.go
guggero: Ah, I guess slicing here makes more sense, fixed.

tapdb/supply_ignore_checker.go
guggero: Yeah, so the idea here is that once we validate an asset point for a certain group, we load all ignore leaves for that group. Which should still be reasonably fast (since we also include a start height) and in the best case avoid further DB calls for the same group.

tapcfg/server.go
guggero: Yeah, was a type issue. But you're right, adding the type param explicitly to the `lfn.Some()` call works too and also makes more sense. Fixed.

itest/supply_commit_test.go
guggero: Done: https://github.com/lightninglabs/taproot-assets/issues/1742

tapdb/supply_ignore_checker.go
guggero: We somehow need to tell the ignore checker that new assets in a group are ignored. And currently the best way for that IMO is the supply verifier state machine, as there we'll definitely know if/when something's being ignored. And the supply verifier can then also use the same invalidation mechanism to update things when we sync new ignore leaves from a universe.

universe/supplycommit/transitions.go
guggero: Hmm, good point. I added the group key to the method so we can remove entries selectively from the cache instead of just clearing it completely.
And I also added a check that there actually are any new ignore updates.

Roasbeef (changes requested):
> Reviewed 37 of 37 files at r1, 2 of 34 files at r2, 15 of 18 files at r3, 17 of 17 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1716)* status: all files reviewed, 13 unresolved discussions (waiting on @ffranr and @guggero)

___
*[`universe/supplycommit/transitions.go` line 1006 at r4](https://reviewable.io/reviews/lightninglabs/taproot-assets/1716#-OXzzi-gCVNOmRXuXg_-:-OXzzi-gCVNOmRXuXg_0:b-rc03hi) ([raw file](https://github.com/lightninglabs/taproot-assets/blob/d1d70188a47c52237da8e6c294074ea443a3d850/universe/supplycommit/transitions.go#L1006)):*
> ```Go
> 		// and the new ignore leaves are loaded from disk.
> 		hasIgnoreUpdates := fn.Any(
> 			danglingUpdates, func(u SupplyUpdateEvent) bool {
> ```

So `danglingUpdates` here is the set up updates that were queued up _while_ we were updating the supply commitment. These haven't yet been inserted into the supply trees. 

Instead, we should pass in `c.SupplyTransition.PendingUpdates`.

guggero:
> Instead, we should pass in `c.SupplyTransition.PendingUpdates.`

Oops, makes sense, fixed.

Roasbeef:
Looks like something shifted during the latest rebase? 
```
panic: 
	assert: mock: I don't know what to return because the method call was unexpected.
		Either do Mock.On("ReportError").Return(...) first, or remove the ReportError() call.
		This method was unexpected:
			ReportError(*fmt.wrapError)
			0: &fmt.wrapError{msg:"group key must be specified for supply tree: unable to unwrap asset group public key", err:(*errors.errorString)(0xc0004fa4d0)}
		at: [/home/runner/work/taproot-assets/taproot-assets/universe/supplycommit/mock.go:414 /home/runner/work/go/pkg/mod/github.com/lightningnetwork/lnd@v0.19.3-beta.rc1.0.20250812194315-c3226e8c2223/protofsm/state_machine.go:679 /home/runner/work/go/pkg/mod/github.com/lightningnetwork/lnd@v0.19.3-beta.rc1.0.20250812194315-c3226e8c2223/protofsm/state_machine.go:229 /home/runner/work/go/pkg/mod/github.com/lightningnetwork/lnd/fn/v2@v2.0.8/goroutine_manager.go:122 /opt/hostedtoolcache/go/1.23.12/x64/src/runtime/asm_amd64.s:1700]
```

guggero:
> Looks like something shifted during the latest rebase?

Oops, I forgot to handle the new error case added by this PR. Updated/fixed the unit tests now.

Issue #1676, opened 2025-07-24
[cleanup]: refactor TAP address code, use specifier, make assetID optional


guggero:
For V2 addresses, the asset leaf (or leaves) that get sent on-chain cannot be predicted anymore and therefore the genesis record isn't really needed, other than for making sure the type of an address matches the asset's type (normal/collectible).
The genesis record should really be optional for V2 addresses. But because we have a foreign key from an address to an asset genesis, we can't fully remove it without also needing to change the database.
So we keep the genesis but make sure we don't rely on it in any V2 code path.

At the same time, it would be nice to refactor the asset ID and group key fields of the `Tap` struct into a specifier, then we can resolve some of the TODOs in the code that read:

```
		// TODO(guggero): We should use an asset specifier in the
		// address and remove the need for the embedded genesis struct.
		// Then we can have a QueryAssetBySpecifier method that we can
		// use here.
```

guggero:
For v2 addresses we should also make the asset ID field optional when encoding the address. See https://github.com/lightninglabs/taproot-assets/pull/1587#discussion_r2245176143.
The address can be quite a bit shorter if instead of encoding the asset ID as all zeroes we can simply omit it when encoding.

ZZiigguurraatt:
Are we going to block the next release for this issue?

guggero:
Probably

PR #1719, opened 2025-08-11
address: clean up code, don't encode asset ID if all zeroes

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1676.

Fixes multiple code TODOs and cleans up some of the address related code.

### DIFF
--- a/address/address.go
+++ b/address/address.go
@@ -418,6 +418,22 @@ func (a *Tap) AttachGenesis(gen asset.Genesis) {
 	a.assetGen = gen
 }
 
+// Specifier returns the asset specifier for this address. For backward
+// compatibility with V0/V1 addresses, we create a specifier with a group key
+// only if the asset ID is zero. For V0 and V1 addresses the group key can be
+// set, but it is not actually used, so we always must include the asset ID in
+// the specifier.
+func (a *Tap) Specifier() asset.Specifier {
+	switch {
+	case a.GroupKey != nil && a.AssetID == asset.ZeroID:
+		return asset.NewSpecifierFromGroupKey(*a.GroupKey)
+	default:
+		return asset.NewSpecifierOptionalGroupPubKey(
+			a.AssetID, a.GroupKey,
+		)
+	}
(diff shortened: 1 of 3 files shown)


address/book.go
```diff
@@ -101,6 +101,11 @@ type Storage interface {
 	QueryAddrs(ctx context.Context,
 		params QueryParams) ([]AddrWithKeyInfo, error)
 
+	// QueryAssetGroup attempts to locate the asset group information
+	// (genesis + group key) associated with a given asset specifier.
+	QueryAssetGroup(context.Context, asset.Specifier) (*asset.AssetGroup,
```
GeorgeTsagk, line 106: could also just drop `QueryAssetGroupByID` (the one modified in the previous commit) in favor of the asset specifier based query

guggero: We still use it in a couple of places, so I'd rather keep it for now.

address/address.go
```diff
@@ -418,6 +418,22 @@ func (a *Tap) AttachGenesis(gen asset.Genesis) {
 	a.assetGen = gen
 }
 
+// Specifier returns the asset specifier for this address. For backward
+// compatibility with V0/V1 addresses, we only create a specifier with a group
```
jtobin: nit: only .. only

jtobin (approved):
LGTM. I was wondering if there was anywhere else that asset.ZeroID could be productively used in place of asset.ID{}, but it looks like most everything else (e.g. in rpcserver.go) is always accompanied by an error return, so probably best to keep that the way it is.

guggero:
Thanks for the review!

>  I was wondering if there was anywhere else that asset.ZeroID could be productively used in place of asset.ID{}...

Yes, we could do that. But since this is mostly for clarity, I wanted to use `asset.ZeroID` for comparisons only. Mainly to avoid the global variable being changed by accident, for example if we returned a pointer to `&asset.ZeroID` somewhere and then we wrote to the result.

PR #1734, opened 2025-08-14
cmd: asset list to correctly filter on and display anchor txid

bhandras:


cmd/commands/assets.go
```diff
@@ -1243,29 +1244,46 @@ var listTransfersCommand = cli.Command{
...
 	ShortName: "t",
 	Usage:     "list asset transfers",
-	Description: "list outgoing transfers of all assets or a selected " +
-		"asset",
+	Description: "list outgoing transfer for a specific anchor or if " +
+		"not set, all transfers",
 	Action: listTransfers,
 	Flags: []cli.Flag{
 		cli.StringFlag{
-			Name: assetIDName,
```
GeorgeTsagk, line 1251: this was just unused? nice

cmd/commands/assets.go
```diff
@@ -1243,29 +1244,46 @@ var listTransfersCommand = cli.Command{
...
+		AnchorTxid: ctx.String(anchorTxidName),
+	}
 	resp, err := client.ListTransfers(ctxc, req)
 	if err != nil {
 		return fmt.Errorf("unable to list asset transfers: %w", err)
 	}
 
+	// As the anchor TX hash is returned in a byte slice we need to reverse
+	// the byte order to match the expected hex format of string serialized
+	// transation IDs.
```
GeorgeTsagk:
```suggestion
	// transaction IDs.

```

cmd/commands/assets.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
```
GeorgeTsagk: seems like you're not using this pkg anywhere?

cmd/commands/assets.go
```diff
@@ -1243,29 +1243,46 @@ var listTransfersCommand = cli.Command{
...
 	resp, err := client.ListTransfers(ctxc, req)
 	if err != nil {
 		return fmt.Errorf("unable to list asset transfers: %w", err)
 	}
 
+	// As the anchor TX hash is returned in a byte slice we need to reverse
+	// the byte order to match the expected hex format of string serialized
+	// transation IDs.
+	for _, transfer := range resp.Transfers {
+		reverseByteOrder(transfer.AnchorTxHash)
```
guggero, line 1283: Makes sense to me! Should've just used a string in the RPC anyway to avoid this issue... I wonder if we should add a quick comment to the field in the proto, mentioning that it is the raw bytes on the RPC response but that we explicitly show it as human-readable in the CLI? Otherwise a future dev might be confused as to the difference...

cmd/commands/assets.go
GeorgeTsagk: already updated

cmd/commands/assets.go
bhandras: Good idea, done!

cmd/commands/assets.go
bhandras: Yeah I think this got forgotten.

cmd/commands/assets.go
bhandras: Done!

PR #1739, opened 2025-08-15
docs/debug-handbook: create new debug run books for developers to help pattern match common issues and knowledge

Roasbeef:
In this PR, we add a README for a new area of documentation intended
for active developers of the project. The goal of this new set of docs
is to capture all the hard earned wisdom that devs gain while debugging
a non-trivial issue. It's meant to serve as a run book to allow for
rapid pattern matching to identify a routine issue, along with a guided
path to aide in debugging.

This should also be great for agents as well, since they can use them to
locate issues that they would otherwise need some prodding to figure
out.

The first one added is intended to devs find PSBT signing related issues more quickly. I ran into this during the dev of https://github.com/lightninglabs/taproot-assets/pull/1675 to get the itests working properly. It relies on some deeper knowledge of: PSBTs, the way lnd signs them, where the code lives, the logs to watch for, etc. So I figured I'd document it. 

I also made a [sub-agent](https://github.com/Roasbeef/claude-files?tab=readme-ov-file#debug-chronicler) that can be useful for such tasks.

### DIFF
--- a/docs/debug-handbook/README.md
+++ b/docs/debug-handbook/README.md
@@ -1,0 +1,168 @@
+# Debug Handbook
+
+This directory contains debugging guides for common issues encountered in taproot-assets development. Each guide provides pattern recognition techniques, debugging flowcharts, and proven resolution strategies.
+
+## Available Guides
+
+### [PSBT Finalization Failures](./psbt-finalization-failures.md)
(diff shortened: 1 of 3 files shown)


Roasbeef:
Landing this as it's purely additive docs.

Issue #1740, opened 2025-08-15
[bug]: `SendAsset`: assets get stuck unconfirmed with group key send


ZZiigguurraatt:
I'm trying to make payments back and forth between Alice and Bob using grouped assets, but at a certain point, some assets being transferred get stuck in an unconfirmed state. After that I can no longer send any assets from Alice to Bob (which I think is because of https://github.com/lightninglabs/taproot-assets/issues/1738) and I get https://github.com/lightninglabs/taproot-assets/issues/1714 . Mining more blocks does not help the problem getting the transfer confirmed.

Here is an itest reproducing my issue: https://github.com/ZZiigguurraatt/taproot-assets/commit/d6ef2ac67e3c1068e66993814756dfe9a08f546b . You can run it with `make itest icase=problem_address_v2_with_group_key`.

More comments about this issue are nested in the itest.

I've also reproduced this in my docker regtest network outside of itest. That is how I was able to discover https://github.com/lightninglabs/taproot-assets/issues/1714 .

guggero:
Thanks for the thorough test case!
I was able to determine where the problem lies (signing of passive assets with the new pedersen unique script key type) and am working on a fix.

Issue #1736, opened 2025-08-14
[bug]: `SyncUniverse` seems to return too early


ZZiigguurraatt:
Here is my situation

1. Alice has some assets she wants to send to Oliver.
2. Oliver creates a taproot asset address and gives it to Alice
3. Alice sends the assets using `SendAsset`
4. 10 blocks are mined
5. Alice and Oliver both wait until tapd is `sync_to_chain`.
6. Oliver runs `SyncUniverse` with `sync_mode=0` and then `sync_mode=1` to get updated proofs from Alice.
7. **Oliver runs `ListAssets`, but he doesn't see the assets yet.**
8. Oliver keeps running `ListAssets` for a few seconds until he finally sees the assets.
9. Alice creates a taproot asset address and gives it to Oliver
10. Oliver **tries** to send the assets back using `SendAsset` but he gets `ERROR MESSAGE: unable to fund address send: error funding packet: unable to list eligible coins: unable to query commitments: mismatch of managed utxo and constructed tap commitment root`.
11. Oliver keeps trying to send again for another few seconds and eventually the send works.

I see two possible (either one or both) causes of the outcome in step 7 (and maybe step 10):

1. Maybe `sync_to_chain` (https://github.com/lightninglabs/taproot-assets/blob/88bb9401d604bd5068770c23e66a8512a38c7aa2/taprpc/taprootassets.proto#L1611-L1614) is true before Alice actually finishes creating the updated proofs after blocks are mined, so when Oliver tries to run `SyncUniverse` he isn't actually getting the updated proofs yet. If that is the case, I think we need `sync_to_chain` to not return true until after all things (like proof generation) related to new blocks are complete. I don't think this is the case though because `SyncUniverse` is not re-run in step 8.
2. `SyncUniverse` returns after the data is transferred but before the proofs are validated and added to tapd's state machine(?).

Here is another clue: The more transfers I've made back and forth between Alice and Oliver, the more time after step 6 until Oliver can successfully see the assets with `ListAssets` and send with `SendAsset`.

If this is a bug you'd like to prioritize I can try to create an itest for it.

ZZiigguurraatt:
> Here is another clue: The more transfers I've made back and forth between Alice and Oliver, the more time after step 6 until Oliver can successfully see the assets with `ListAssets` and send with `SendAsset`.

Actually, more testing may be required to validate this statement.

Roasbeef:
[...]

PR #1741, opened 2025-08-18
bugfix: fix two long-existing edge case bugs

guggero:
Fixes https://github.com/lightninglabs/taproot-assets/issues/1740.

Fixes two bugs that have been in the code base for a long time but only have been triggered recently by the test mentioned here: https://github.com/lightninglabs/taproot-assets/issues/1740
1. Use correct input asset: When an asset is a split output and then in the next transfer becomes a passive asset, the input asset used when signing was not the correct one (missing the split commitment), which lead to the `invalid transfer asset witness` error. The fix is using the correct input asset from the input proof instead of the asset from the trimmed input commitment.
2. Store the split commitment root in the DB when importing asset from proof: This is super old and shows that so far we've only ever kept the split root of a transfer on our local node (as change or tombstone) and have never sent it out to another node as part of an interactive transfer. The fix is storing the split commitment root hash and value when importing an asset from a proof file. This fixes the following error: `unable to fund address send: error funding packet: unable to list eligible coins: unable to query commitments: mismatch of managed utxo and constructed tap commitment root`.

I've also included the integration test from https://github.com/ZZiigguurraatt/taproot-assets/commit/d6ef2ac67e3c1068e66993814756dfe9a08f546b which reproduced the first bug reliably (thanks a lot, @ZZiigguurraatt, that was super helpful!).

cc @bhandras, not sure if you perhaps ran into either of these bugs in your work?

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -26,6 +26,16 @@
 - [An integration test flake was
   fixed](https://github.com/lightninglabs/taproot-assets/pull/1651).
 
+- Fixed two send related bugs that would lead to either a `invalid transfer 
+  asset witness` or `unable to fund address send: error funding packet: unable 
+  to list eligible coins: unable to query commitments: mismatch of managed utxo
+  and constructed tap commitment root` error when sending assets.
+  The [PR that fixed the two
+  bugs](https://github.com/lightninglabs/taproot-assets/pull/1741) also
+  optimized sending to V2 TAP addresses by removing the need for creating
+  tombstone outputs on a full-value send (by using interactive transfers for V2
+  addresses).
+
 # New Features
 
 ## Functional Enhancements


ZZiigguurraatt:
> 2\. Store the split commitment root in the DB when importing asset from proof: This is super old and shows that so far we've only ever kept the split root of a transfer on our local node (as change or tombstone) and have never sent it out to another node as part of an interactive transfer. The fix is storing the split commitment root hash and value when importing an asset from a proof file. This fixes the following error: `unable to fund address send: error funding packet: unable to list eligible coins: unable to query commitments: mismatch of managed utxo and constructed tap commitment root`.

Is this the cause of step 10 (the second problem) in https://github.com/lightninglabs/taproot-assets/issues/1736 ?

tapfreighter/wallet.go
```diff
@@ -503,14 +503,20 @@ func createPassivePacket(passiveAsset *asset.Asset,
 	params := activePackets[0].ChainParams
 
 	// Specify virtual input.
-	inputAsset := passiveAsset.Copy()
+	inputAsset := inputProof.Asset.Copy()
```
Roasbeef, line 506: With this change, is there any reason to pass in the `passiveAsset` as the first arg here?

tapfreighter/wallet.go
```diff
@@ -503,14 +503,20 @@ func createPassivePacket(passiveAsset *asset.Asset,
...
 	vInput := tappsbt.VInput{
 		Proof: inputProof,
 		PInput: psbt.PInput{
 			SighashType: txscript.SigHashDefault,
 		},
 	}
 
+	// The passive asset is the asset queried from the database, and it
```
Roasbeef, line 514: From the commit message: what does trim refer to here? Looking at the latest version of the call site, I see that we trim the alt leaves, but why would that affect if a split commitment is present or not?

itest/addrs_v2_test.go
```diff
@@ -271,8 +271,7 @@ func testAddressV2WithGroupKey(t *harnessTest) {
 	AssertAssetOutboundTransferWithOutputs(
 		t.t, t.lndHarness.Miner().Client, t.tapd,
 		sendResp.Transfer, [][]byte{firstAssetID, secondAssetID},
-		[]uint64{0, firstAsset.Amount, 0, secondAsset.Amount}, 0,
-		1, 4, true,
+		[]uint64{firstAsset.Amount, secondAsset.Amount}, 0, 1, 2, true,
```
Roasbeef, line 274: > we can
optimize chain usage by avoiding the creation of a tombstone

How does this optimize chain usage? You mean that we can skip the future chain transactions which will finally garbage collect the tombstones?

tapfreighter/wallet.go
Roasbeef: Ah, I see this here now: https://github.com/lightninglabs/taproot-assets/blob/6ed2f32c1bdd044c8e46e3322857ef338345fa64/tapfreighter/wallet.go#L557-L560

Roasbeef:
<details>

<summary>Here's an extra unit test that we can add (at the `tapdb` level, covers the issue where import proof didn't write the split commit info): </summary>

```go
// createTestProof creates an annotated proof for testing.
func createTestProof(t *testing.T,
	testAsset *asset.Asset) *proof.AnnotatedProof {

	return createTestProofWithAnchor(t, testAsset, 0)
}

// createTestProofWithAnchor creates an annotated proof with specific anchor
// index.
func createTestProofWithAnchor(t *testing.T, testAsset *asset.Asset,
	anchorIndex uint32) *proof.AnnotatedProof {

	var blockHash chainhash.Hash
	_, err := rand.Read(blockHash[:])
	require.NoError(t, err)

	anchorTx := wire.NewMsgTx(2)
	anchorTx.AddTxIn(&wire.TxIn{})

	// Add enough outputs to cover the requested index
	for i := uint32(0); i <= anchorIndex; i++ {
		anchorTx.AddTxOut(&wire.TxOut{
			PkScript: bytes.Repeat([]byte{0x01}, 34),
			Value:    10,
		})
	}

	assetRoot, err := commitment.NewAssetCommitment(testAsset)
	require.NoError(t, err)

	commitVersion := test.RandFlip(nil, fn.Ptr(commitment.TapCommitmentV2))
	taprootAssetRoot, err := commitment.NewTapCommitment(
		commitVersion, assetRoot,
	)
	require.NoError(t, err)

	testProof := randProof(t, testAsset)
	proofBlob, err := testProof.Bytes()
	require.NoError(t, err)

	assetID := testAsset.ID()
	anchorPoint := wire.OutPoint{
		Hash:  anchorTx.TxHash(),
		Index: anchorIndex,
	}

	return &proof.AnnotatedProof{
		Locator: proof.Locator{
			AssetID:   &assetID,
			ScriptKey: *testAsset.ScriptKey.PubKey,
		},
		Blob: proofBlob,
		AssetSnapshot: &proof.AssetSnapshot{
			Asset:             testAsset,
			OutPoint:          anchorPoint,
			AnchorBlockHash:   blockHash,
			AnchorBlockHeight: uint32(test.RandIntn(1000) + 1),
			AnchorTxIndex:     test.RandInt[uint32](),
			AnchorTx:          anchorTx,
			OutputIndex:       anchorIndex,
			InternalKey:       test.RandPubKey(t),
			ScriptRoot:        taprootAssetRoot,
		},
	}
}

// TestUpsertAssetsWithSplitCommitments tests that split commitment data is
// properly stored and retrieved during asset upsert operations.
func TestUpsertAssetsWithSplitCommitments(t *testing.T) {
	t.Parallel()

	testCases := []struct {
		name           string
		hasSplitCommit bool
		splitValue     uint64
	}{
		{
			name:           "asset_with_split_commitment",
			hasSplitCommit: true,
			splitValue:     12345,
		},
		{
			name:           "asset_without_split_commitment",
			hasSplitCommit: false,
		},
	}

		tc := tc
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()

			ctx := context.Background()
			dbHandle := NewDbHandle(t)

			testAsset := randAsset(t)

			var expectedHash mssmt.NodeHash
			if tc.hasSplitCommit {
				splitHash := test.RandBytes(32)
				copy(expectedHash[:], splitHash)

				testAsset.SplitCommitmentRoot = mssmt.NewComputedNode( //nolint:lll
					expectedHash, tc.splitValue,
				)
			}

			annotatedProof := createTestProof(
				t, testAsset,
			)

			err := dbHandle.AssetStore.ImportProofs(
				ctx, proof.MockVerifierCtx, false,
				annotatedProof,
			)
			require.NoError(t, err)

			assets, err := dbHandle.AssetStore.FetchAllAssets(ctx, false, false, nil)
			require.NoError(t, err)
			require.Len(t, assets, 1)

			dbAsset := assets[0].Asset

			if tc.hasSplitCommit {
				require.NotNil(
					t, dbAsset.SplitCommitmentRoot,
				)
				gotHash := dbAsset.SplitCommitmentRoot.NodeHash()
				require.Equal(t, expectedHash, gotHash)
				require.Equal(
					t, tc.splitValue,
					dbAsset.SplitCommitmentRoot.NodeSum(),
				)
			} else {
				require.Nil(
					t, dbAsset.SplitCommitmentRoot,
					"Split commitment root should be nil",
				)
			}
		})
	}
}
```
</details>

tapfreighter/wallet.go
guggero: Trim here refers to the commitment after calling `commitment.TrimSplitWitnesses()`, where we explicitly set `SplitCommitment` to `nil`.

itest/addrs_v2_test.go
guggero: Yes. We avoid spending 1k sats for a tombstone output that is then locked until we implement garbage collection. So fewer on-chain outputs and fewer sats locked for users.

guggero:
> Is this the cause of step 10 (the second problem) in https://github.com/lightninglabs/taproot-assets/issues/1736 ?

Yes, I'm pretty sure that's the bug you ran into. At least the error message is identical (but different cases could potentially lead to that error message, but hoping this is the one fixed here, otherwise please make sure to report again).

guggero:
> Here's an extra unit test that we can add (at the `tapdb` level, covers the issue where import proof didn't write the split commit info):

Awesome, thanks you! Added.

ZZiigguurraatt:
> I've also included the integration test from [ZZiigguurraatt@d6ef2ac](https://github.com/ZZiigguurraatt/taproot-assets/commit/d6ef2ac67e3c1068e66993814756dfe9a08f546b) which reproduced the first bug reliably (thanks a lot, @ZZiigguurraatt, that was super helpful!).

If this issue has been around for a while, should we re-write the itest to do a non-group key send? I was able to reproduce the issue sending the individual asset tranches using a non-group key V1 address, but I did not try to do the multi asset id send with V1 address. Perhaps that could be done by creating 3 V1 addresses and then paying to them all in the same invocation of `SendAsset` since we can pass multiple addresses to it at once (https://github.com/lightninglabs/taproot-assets/blob/ca434e489099948fe40efae5aa87b9843cfaab23/taprpc/taprootassets.proto#L1524-L1531)?

guggero:
> > I've also included the integration test from [ZZiigguurraatt@d6ef2ac](https://github.com/ZZiigguurraatt/taproot-assets/commit/d6ef2ac67e3c1068e66993814756dfe9a08f546b) which reproduced the first bug reliably (thanks a lot, @ZZiigguurraatt, that was super helpful!).
> 
> If this issue has been around for a while, should we re-write the itest to do a non-group key send? I was able to reproduce the issue sending the individual asset tranches using a non-group key V1 address, but I did not try to do the multi asset id send with V1 address. Perhaps that could be done by creating 3 V1 addresses and then paying to them all in the same invocation of `SendAsset` since we can pass multiple addresses to it at once (
> 
> https://github.com/lightninglabs/taproot-assets/blob/ca434e489099948fe40efae5aa87b9843cfaab23/taprpc/taprootassets.proto#L1524-L1531
> )?

No, I don't think that will work. Because you need multiple split outputs being in the same on-chain output. And you can't do that with V0/V1 addresses, each address will get its own on-chain output.
We can do it with the vPSBT API and I did add a reproduction itest case for that, in addition to your V2 itest case. So I think this case is now covered at least twice, which is much better than before.

ZZiigguurraatt:
> We can do it with the vPSBT API

So this was the only way that this bug could have been previously exposed? You say this PR fixes two long standing edge cases, but before V2 addresses actually getting them to happen would be very hard for the average user and only a really advanced user could have even found them?

guggero:
> So this was the only way that this bug could have been previously exposed? You say this PR fixes two long standing edge cases, but before V2 addresses actually getting them to happen would be very hard for the average user and only a really advanced user could have even found them?

Correct, so we're lucky nobody ran into these two cases yet.

ZZiigguurraatt:
> > Is this the cause of step 10 (the second problem) in #1736 ?
> 
> Yes, I'm pretty sure that's the bug you ran into. At least the error message is identical (but different cases could potentially lead to that error message, but hoping this is the one fixed here, otherwise please make sure to report again).

I can confirm that this PR did not fix https://github.com/lightninglabs/taproot-assets/issues/1736 . It seems to be a different issue with the same error message.

PR #1744, opened 2025-08-18
Rename RPCS subsystem to TRPC

### LINKED ISSUE #1743: [bug]: logging is complected with lnd's when running in integrated mode
When tapd runs in integrated mode, e.g. with [terminal](https://github.com/lightninglabs/lightning-terminal), its log debuglevel, otherwise set by 'taproot-assets.debuglevel', is actually governed by 'lnd.debuglevel'. Additionally, logs for subsystems with names that shadow subsystems in LND can fail to display.

To reproduce: run tapd in integrated mode with e.g. 'lnd.debuglevel=debug' and perform any action (e.g. an RFQ) that should lead to a `rpcsLog.Debugf` statement; the log will fail to emit, as the subsystem's name (RPCS) shadows one in lnd (also RPCS). We want to see that log.

It seems to me that tapd's debuglevel should ideally be governed by the taproot-assets.debuglevel setting, even when running in integrated mode, and that all subsystem logs should print as expected (either by handling the logging differently, or just avoiding the practice of shadowing lnd subsystems).

jtobin:
Partially resolves #1743.

The subsystem name RPCS clashes with one of lnd's when tapd is running in integrated mode, preventing certain tapd logs from being emitted. TRPC avoids the clash, and fits the theme used for e.g. TSVR.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -225,7 +225,7 @@ func (r *rpcServer) Start(cfg *Config) error {
 		r.cfg.UniverseQueriesPerSecond, r.cfg.UniverseQueriesBurst,
 	)
 
-	rpcsLog.Infof("Starting RPC Server")
+	rpcsLog.Infof("Starting Taproot Assets RPC Server")
 
 	return nil
 }
@@ -237,7 +237,7 @@ func (r *rpcServer) Stop() error {
 		return nil
 	}
 
-	rpcsLog.Infof("Stopping RPC Server")
+	rpcsLog.Infof("Stopping Taproot Assets RPC Server")
 
 	close(r.quit)
 


guggero:
IMO we can ignore the release notes, can merge without. Also, there's the `no-changelog` label that can be added to a PR if it should ignore the release notes check (but it needs to be added when creating the PR, otherwise it doesn't take effect, at least not for the first CI run).

jtobin:
> IMO we can ignore the release notes, can merge without. Also, there's the `no-changelog` label that can be added to a PR if it should ignore the release notes check (but it needs to be added when creating the PR, otherwise it doesn't take effect, at least not for the first CI run).

Ah yes, I honestly overlooked this entirely -- will remember to at least think about it next time! :grin:

Roasbeef (approved):
LGTM 🍸

PR #1748, opened 2025-08-19
Feature bits

### LINKED ISSUE #1573: [feature]/tapchannels:

As the protocol evolves, we'll need to be able to also evolve the asset channels format and operations. 

One example is the `NoopAdd` HTLC we've ideated, which gives us a solution to always having to push over a sat amt for a payment. 

Without any sort of versioning or feature bit scheme on the channel level, we won't be smoothly introduce new features to channels. In that realm, the only way to "upgrade" is to close out and open channels again, which isn't desirable for end users. 

We have quite a few extension levers at our finger tips:
  * The `channel_type` feature bit:
     * This is sent over in the `open_channel` message. If we don't know some extra bit (maybe it's a slightly diff format), then we can send a reject message over. 
     * We can thread this down through the into the `FundingController` to decide if we support that channel type or not. 
   * The set of feature bits lnd advertises:
     * This can be used to filter out nodes in the graph that support a given feature. 
     * Right now this value is static in `lnd`, but we can update it via the [`UpdateNodeAnnouncement` ](https://lightning.engineering/api-docs/api/lnd/peers/update-node-announcement/#grpc)RPC. 
     * This'll let us start to advertise new variants of the channel format. It will be up to `tapd` to try to send over the best feature bit combo for compat.
   * The `channel_reestablish` message:
      * This has space for extra TLVs. Today splicing, and also taproot channels use up this space. 
      * We can use this to detect behavioral changes to an _existing_ channel. 
      * We likely shouldn't take this path often, as it is effectively an in-place update, so it should be reserved mainly for bug fixes. 

Of the above, the only area we lack hooks for is putting/reading extra data in the chan reest message. 

The the noop HTLC feature, I think we can just add a new feature bit, [then check to see](https://github.com/lightninglabs/taproot-assets/blob/608bed8d75f3585f9be59eaa1f91b18243fabc02/tapchannel/aux_funding_controller.go#L109-L116) if the peer supports it before we use it. 

The chan reest message is also an option here. 

If we make any explicit breaking changes such as changing the scripts, or the way we decide things like who the anchor will be, then we'd bump the `channel_type` feature bit. For this, we'll need some versioning within the TAP specific channel code.

GeorgeTsagk:
## Description

Adds the necessary subsystem to handle custom feature bit negotiation for tap channels. This allows us to introduce new channel features that would otherwise break backwards compatibility. We piggy-back on the existing LND messages (init & reestablish) to nest the `tlv.Blob` that encodes the `CustomChannelType` (uint64) type.

The set of active features over a channel is the intersection of the locally supported features and the features that the peer provided.

Currently we follow a "global" approach on the feature bits that we support. We don't make any distinctions between peers or channels. The locally supported feature bits is a static set.

We also don't allow reading/writing feature bits per channel, but per peer. For now the funding outpoint / funding blob arguments are ignored. When we establish a feature-set with a peer, that applies to all of our channels with that peer.

An older tapd node (i.e a node that does not contain the contents of this PR) will be treated as if they provided an empty feature set (`uint64` with `0`'s)

### Dependencies
LND https://github.com/lightningnetwork/lnd/pull/10182
Lndclient https://github.com/lightninglabs/lndclient/pull/239
LiT itest: https://github.com/lightninglabs/lightning-terminal/pull/1138

Closes #1573

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -149,6 +149,14 @@
   controls whether the peer's identity public key is sent to the local price
   oracle when querying asset price rates.
 
+- Tap channels now use
+  [feature bits](https://github.com/lightninglabs/taproot-assets/pull/1748), in
+  order to provide a safe way to upgrade the channels into using new features
+  without breaking backwards compatibility. This way a new feature will become
+  active over a channel only when both peers decide to use it. When funding a
+  channel or upon reconnection we check that the required features are supported
+  by the remote party before proceeding.
+
 ## RPC Additions
 
 - The [price oracle RPC calls now have an intent, optional peer ID and metadata


tapfeatures/aux_channel_negotiator.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// GetInitFeatures is called when sending an init message to a peer. It returns
+// custom feature bits to include in the init message TLVs. The implementation
+// can decide which features to advertise based on the peer's identity.
+func (n *AuxChannelNegotiator) GetInitFeatures(
+	peer [33]byte) (tlv.Blob, error) {
+
+	buf := make([]byte, 8)
+	binary.BigEndian.PutUint64(buf, uint64(getCustomChannelType()))
```
Roasbeef: So if this is meant to go into the init message, then we can't just use feature bit 0 or 1. As that would collide with the _existing_ features that `lnd` will set in the init message. Instead, we'll need to use a higher value, like low hundreds or something within that range.

tapfeatures/aux_channel_negotiator.go
```diff
@@ -0,0 +1,81 @@
...
+	return buf, nil
+}
+
+// ProcessInitFeatures handles received init feature TLVs from a peer. The
+// implementation can store state internally to affect future channel operations
+// with this peer.
+func (n *AuxChannelNegotiator) ProcessInitFeatures(peer [33]byte,
+	features tlv.Blob) error {
+
+	if len(features) != 8 {
```
Roasbeef: Why would we expect a set size here? IIUC, isn't this the normal feature bit struct sent over? 

I think this should be something like `lnwire.FeatureVector` here. Then we can use those same methods to check if a bit set set or not, and we can avoid the AND masking that we have below.

tapfeatures/aux_channel_negotiator.go
GeorgeTsagk: This doesn't collide with any other fields of the `init` message, it's an extra TLV which has it's own record. LND treats this as a `tlv.Blob` so we can really encode it as anything we want, I chose the uint64 (similar to lnd `ChannelType`) for simplicity -- each bit is a flag.

tapfeatures/aux_channel_negotiator.go
GeorgeTsagk: > Why would we expect a set size here? IIUC, isn't this the normal feature bit struct sent over?

I press everything together in a `uint64`, that's why I'm expecting 8 bytes.

> I think this should be something like lnwire.FeatureVector here.

Will take a look.

rfq/order.go
```diff
@@ -125,13 +127,20 @@ type AssetSalePolicy struct {
 	// wants us to produce NoOp HTLCs.
 	NoOpHTLCs bool
 
+	auxChanNegotiator *tapfeatures.AuxChannelNegotiator
```
jtobin, line 132: Could we replace the NoOpHTLCs flag and the auxChanNegotiator reference here with a single 'UseNoOpHTLCs' flag?

The ultimate purpose of including the auxChanNegotiator reference here is to eventually check if the peer supports NoOpHTLCs, which is then checked against our own NoOpHTLCs flag. But it would be nice to get away with storing only a single flag here, if that check could be resolved earlier. Could it be done in RegisterAssetSalePolicy or similar?

rfq/manager.go
```diff
@@ -112,6 +113,8 @@ type ManagerCfg struct {
 	// into the manager once lnd and tapd are hooked together.
 	AliasManager ScidAliasManager
 
+	AuxChanNegotiator *tapfeatures.AuxChannelNegotiator
```
jtobin, line 119: See my other comment on rfq/order.go, which I wrote first, but it would be nice if we could avoid sticking this in here, propagating only a single 'UseNoOptHTLCs' flag. Is it possible to resolve that stuff any earlier?

rfq/order.go
```diff
@@ -125,13 +127,20 @@ type AssetSalePolicy struct {
...
+
+	peer route.Vertex
```
jtobin, line 135: (This can be axed as well if it proves an appropriate 'use no-op HTLCs' check can be performed earlier.)

tapfeatures/aux_channel_negotiator.go
```diff
@@ -0,0 +1,165 @@
...
+// NewAuxChannelNegotiator returns a new instance of the aux channel negotiator.
+func NewAuxChannelNegotiator() *AuxChannelNegotiator {
+	return &AuxChannelNegotiator{}
+}
+
+// GetInitFeatures is called when sending an init message to a peer. It returns
+// custom feature bits to include in the init message TLVs. The implementation
+// can decide which features to advertise based on the peer's identity.
+func (n *AuxChannelNegotiator) GetInitFeatures(
+	peer route.Vertex) (tlv.Blob, error) {
```
jtobin: 'peer' is unused here. I note the comment above, so I guess it's placed there now as it's intended to be used in the future? Maybe cleaner to axe it, or add more explicit "justifying" commentary about it being reserved for future use.

tapfeatures/aux_channel_negotiator.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	// Store this peer's features.
+	n.peerFeatures.Store(peer, peerVec)
+
+	return nil
+}
+
+// GetReestablishFeatures is called when sending a channel_reestablish message.
+// It returns feature bits based on the specific channel identified by its
```
jtobin: These arguments are unused similarly (though are of course unnamed here). Maybe more explicitly justify them in the comment if it's decided to keep them around.

tapfeatures/aux_channel_negotiator.go
```diff
@@ -0,0 +1,165 @@
...
+	if !ok {
+		rawfeatures = lnwire.NewRawFeatureVector()
+	}
+
+	return lnwire.NewFeatureVector(rawfeatures, featureNames)
+}
+
+// checkRequiredBits is a helper method that checks if all of the required bits
+// of the first vector are supported by the second vector.
+func checkRequiredBits(a, b *lnwire.RawFeatureVector) error {
```
jtobin: This one would be more immediately clear if 'a' and 'b' were replaced by 'local' and 'peer' or something, per the error message (a cute idiom I've seen before is 'ours' and 'hers' 😄).

jtobin (commented):
A few initial questions and comments, mostly about the AuxChanNegotiator handling in the RFQ system.

rfq/order.go
jtobin: (The further back it can be pushed the better. Possible it's necessary here as-is, but it's not immediately clear as to why.)

GeorgeTsagk: We want to pass the negotiator itself here because it matters in terms of when the evaluation is happening.

With your suggested approach, we would be evaluating the feature bit upon instantiating the policy (only once, right at that point). The issue is that a policy can last for a long duration, during which our peer may restart and change their feature bits. If that happens then we'll have cached a stale entry here and possibly lead to undesired behavior.

So by passing in the instance we can evaluate if noop HTLCs should be used, based on both our peer's feature bit, plus our local configuration.

rfq/manager.go
GeorgeTsagk: See https://github.com/lightninglabs/taproot-assets/pull/1748#discussion_r2314277698.

Also there's a slight difference between our local configuration flag and what the feature bits dictate.

It can definitely be the case the we both use upgraded nodes that understand noop HTLCs, but for some reason (user preference, or bug disclosure) I want to directly disable noop HTLCs, without upgrading/downgrading my software. This is why the configuration flag is also useful.

rfq/order.go
GeorgeTsagk: https://github.com/lightninglabs/taproot-assets/pull/1748#discussion_r2314277698

tapfeatures/aux_channel_negotiator.go
GeorgeTsagk: Yeah it's not intended for use at the moment, will replace with `_`.

tapfeatures/aux_channel_negotiator.go
GeorgeTsagk: This func signature must match the LND interface, so want to avoid having to bump LND when we decide to make this logic a bit more involved / smart.

rfq/order.go
jtobin: Ahh right, I actually hadn't realised the policy could be long lived. Makes sense!

rfq/order.go
```diff
@@ -125,13 +127,23 @@ type AssetSalePolicy struct {
 	// wants us to produce NoOp HTLCs.
 	NoOpHTLCs bool
```
jtobin, line 128: Can anything pathological happen if we register an AssetSalePolicy with NoOpHTLCs enabled/disabled, but then restart with NoOpHTLCs disabled/enabled before the policy expires?

rfq/order.go
```diff
@@ -125,13 +127,23 @@ type AssetSalePolicy struct {
...
 
+	// auxChannelNegotiator is used to query the supported feature bits that
+	// are supported by a peer, or a channel.
+	auxChanNegotiator *tapfeatures.AuxChannelNegotiator
```
jtobin, line 132: Revisiting: see my PR to this branch for a potential alternative way to handle this.

rfq/order.go
```diff
@@ -125,13 +127,23 @@ type AssetSalePolicy struct {
...
 	// expiry is the policy's expiry unix timestamp after which the policy
 	// is no longer valid.
 	expiry uint64
 }
 
 // NewAssetSalePolicy creates a new asset sale policy.
-func NewAssetSalePolicy(quote rfqmsg.BuyAccept, noop bool) *AssetSalePolicy {
+func NewAssetSalePolicy(quote rfqmsg.BuyAccept, noop bool,
+	chanNegotiator *tapfeatures.AuxChannelNegotiator,
+	peer route.Vertex) *AssetSalePolicy {
```
jtobin: I think the 'peer' argument here is unnecessary, as it can be gotten from quote.Peer?

rfq/order.go
GeorgeTsagk: No, but not for the reason you might expect.

We currently do not persist negotiated quotes to storage, so upon restart we'll actually lose track of all the negotiated quotes. Adding quote persistence is a todo-item we've had in our backlog for a while.

But for future reference, I don't think we'd want to bind the quote with the value of noop at **quote creation time**. The noop config value is a runtime-based configuration so if the user restarts and turns off noop, then we should stop transmitting noop HTLCs.

rfq/order.go
GeorgeTsagk: Checked the recommendation, but don't think I'll apply it for these reasons:

- Keeping a reference to `auxChanNegotiator` is okay as far as safety goes, all the fields of the instance are safe to access async (featurebits are stored in `lnutils.SyncMap`)
-  This way we have to change the policy interface by changing `policy.GenerateInterceptorResponse`, which will have to be reflected to all `AssetSalePolicy`, `AssetPurchasePolicy` and `AssetForwardPolicy`. Only `AssetSalePolicy` really needs it.

rfq/order.go
GeorgeTsagk: Yeap! Thanks for the catch

rfq/order.go
jtobin: Gotcha, and agreed re: not binding the noop setting at quote creation.

rfq/order.go
jtobin: Last question, and then I swear I'll stop beating this thing: would keeping the reference around vex any eventual quote serialization/persistence efforts? Not sure we'd want to serialize the raw AssetSalePolicy anyway, and there's already some stuff in there (e.g. the mutex) that might frustrate that endeavour, but..

GeorgeTsagk: We could just exclude serializing the pointer (which doesn't really make sense to serialize in the first place) and keep the other fields. The pointer would be restored when the quote would become active again, possibly pointing to a new instance.

jtobin (approved):
LGTM :+1: :+1:

tapfeatures/aux_feature_bits.go
```diff
@@ -0,0 +1,51 @@
...
+// featureNames keeps track of the string description of known features.
+var featureNames = map[lnwire.FeatureBit]string{
+	NoOpHTLCsRequired: "noop-htlcs",
+	NoOpHTLCsOptional: "noop-htlcs",
+	STXORequired:      "stxo-proofs",
+	STXOOptional:      "stxo-proofs",
+}
+
+// ourFeatures returns a slice containing all of the locally supported features.
+func ourFeatures() []lnwire.FeatureBit {
```
Roasbeef, line 32: Can be a constant instead?

Roasbeef: Same below.

tapfeatures/aux_channel_negotiator.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+// AuxChannelNegotiator is responsible for producing the extra tlv blob that is
+// encapsulated in the init and reestablish peer messages. This helps us
+// communicate custom feature bits with our peer.
+type AuxChannelNegotiator struct {
+	// peerFeatures keeps track of the supported features of our peers. This
+	// map will be used for lookups by other subsystems, when some features
+	// need to be supported by both parties to take effect.
+	peerFeatures lnutils.SyncMap[route.Vertex, *lnwire.RawFeatureVector]
```
Roasbeef, line 31: Ah ok, I hadn't reviewed this PR yet before reviewing the follow up. I see now that the negotiator stores the information retrieved from `lnd` directly in memory.

tapfeatures/aux_channel_negotiator.go
```diff
@@ -0,0 +1,191 @@
...
+// NewAuxChannelNegotiator returns a new instance of the aux channel negotiator.
+func NewAuxChannelNegotiator() *AuxChannelNegotiator {
+	return &AuxChannelNegotiator{}
+}
+
+// GetInitFeatures is called when sending an init message to a peer. It returns
+// custom feature bits to include in the init message TLVs. The implementation
+// can decide which features to advertise based on the peer's identity.
+func (n *AuxChannelNegotiator) GetInitFeatures(
+	_ route.Vertex) (tlv.Blob, error) {
```
Roasbeef: I think w/ the latest direction, this can now just return the feature vector directly, to be merged by lnd.

tapfeatures/aux_channel_negotiator.go
```diff
@@ -0,0 +1,191 @@
...
+	// Before we store this peer's supported features we need to first check
+	// if our required features are supported by that peer. If a locally
+	// required feature is not supported by the remote peer we have to
+	// return an error and drop the connection. Whether we support all of
+	// the remote required features is a responsibility of the remote peer.
+	// If we fail to support a remotely required feature they are the ones
+	// to drop the connection (by returning an error right here).
+	err = checkRequiredBits(getLocalFeatureVec(), peerVec)
+	if err != nil {
+		return err
```
Roasbeef, line 97: `lnd` doesn't need to handle this error right?

tapfeatures/aux_channel_negotiator.go
```diff
@@ -0,0 +1,191 @@
...
+
+	n.chanFeatures.Store(cid, peerVec)
+
+	return nil
+}
+
+// GetPeerFeatures returns the negotiated feature bit vector that was
+// established with the given peer.
+func (n *AuxChannelNegotiator) GetPeerFeatures(
+	peer route.Vertex) *lnwire.FeatureVector {
```
Roasbeef: If we're returning `nil` rather than an option, then we need to take care to always check the return value in the follow up PR. 

Otherwise, if `lnd` has restarted, but the peers haven't connected again, then we'll panic when trying to check if feature bits are present.

tapchannel/aux_traffic_shaper.go
```diff
@@ -464,6 +467,9 @@ func (s *AuxTrafficShaper) ProduceHtlcExtraData(totalAmount lnwire.MilliSatoshi,
 		return totalAmount, nil, nil
 	}
 
+	peerFeatures := s.cfg.AuxChanNegotiator.GetPeerFeatures(peer)
+	supportNoOp := peerFeatures.HasFeature(tapfeatures.NoOpHTLCsOptional)
```
Roasbeef, line 471: 👍

tapfeatures/aux_feature_bits.go
GeorgeTsagk: As in a const array? Not aware of how we do this in the codebase

If you mean just `var ourFeatures[...]lnwire.FeatureBit{a, b, c, d}` I wanted to avoid that since many different goroutines may be accessing it. The current approach with the func returns a new slice so each caller gets their own pointer

tapfeatures/aux_channel_negotiator.go
GeorgeTsagk: Wdym? To merge this feature vector with the one that LND uses in `lnwire.Init`?

Might be a nice convenience since that message already uses a feature vec, but can't see the opportunities. We'd skip a few lines (just for `lnwire.Init`) which encode/decode the aux features within `ExtraData`, which is a bit more lean, but then again for the other messages we'd have to maintain the current approach as they don't have a feature-vec field.

tapfeatures/aux_channel_negotiator.go
GeorgeTsagk: we expect LND to fail the peer connection, preventing any channels from becoming active

tapfeatures/aux_channel_negotiator.go
GeorgeTsagk: Yeah this pointer was always set, but we actually don't need to return a pointer to the feature vec here. Changed to just return `lnwire.FeatureVector`

GeorgeTsagk:
Also guarded the new hooks with the `s.waitForReady` call to make sure that the subsystems have been initialized, as we're using `AuxChanNegotiator` in the new hooks

GeorgeTsagk:
Pushed an update that uses the new LND aux negotiator interface.

GeorgeTsagk:
Will have to update golang to v1.24.6, will probably create a separate PR for that and rebase

Roasbeef:
Can be rebased now!

tapfeatures/aux_channel_negotiator.go
Roasbeef: Confirmed behavior here: https://github.com/lightningnetwork/lnd/pull/10182/files#diff-8986cde7a8ea16ac62b7e29ee740f3f307907e60c74818cb05125a3508c60bb9R4546-R4556

server.go
```diff
@@ -1296,3 +1296,68 @@ func (s *Server) NotifyBroadcast(req *sweep.BumpRequest,
...
+
+// GetInitRecords is called when sending an init message to a peer. It returns
+// custom records to include in the init message TLVs. The implementation can
+// decide which records to include based on the peer identity.
+func (s *Server) GetInitRecords(
+	peer route.Vertex) (lnwire.CustomRecords, error) {
+
+	srvrLog.Tracef("GetInitRecords called, peer=%s", peer)
+
+	if err := s.waitForReady(); err != nil {
```
Roasbeef, line 1331: Do we actually need this here? IIUC `GetInitRecords()` will just read some static variables. 

We need to be careful here to _not_ introduce yet another deadlock bug when `lnd` is trying to start up but `tapd` hasn't been fully initialized yet. 

I don't think any of these need `waitForReady`, as they don't interact with any outside daemon/sub-system.

Roasbeef (changes requested):

server.go
GeorgeTsagk: We're always using the aux chan negotiator, which needs to be initialized by the server.

Without this we'll be racing against the server, occasionally panicking because of nil dereference.

For the init records we could actually have it be static and detached from the negotiator -- the rest of the calls I'm not so sure about.

GeorgeTsagk:
Rebased on `main` with upgraded go to v1.24.6, CI should be more green now (LiT is still expected to fail, see corresponding PR for status of LiT itests)

jtobin:
(Just noting for posterity that this will resolve #1573.)

GeorgeTsagk:
> (Just noting for posterity that this will resolve #1573.)

Thanks, added a `Closes` clause on the PR description.

litbot-9000:
@roasbeef: review reminder

server.go
Roasbeef: > Without this we'll be racing against the server, occasionally panicking because of nil dereference.

Why would we panic? 

I'm referring to issues like this: https://github.com/lightninglabs/taproot-assets/pull/1780. 

Init can def be static.

Do we have a _reason_ to include this? Given that we've seen deadlock instances when this wasn't needed?

Roasbeef (approved):
LGTM 🪻

GeorgeTsagk:
On the concerns related to servers reaching a deadlock waiting for each other to start:

Currently the new aux hooks in tapd related to aux features are guarded by `s.waitForReady()` which means that the tapd server must has been properly initialized and started.

On the LND side the callsites for these new hooks reside within brontide and the htlcswitch links (created by brontide). Any other LND sub-server does not have its start-up affected by these new call-sites, which are also ran asynchronously in a goroutine.

This PR is now ready to merge into `main`, but since the corresponding LiT itest PR is not yet reviewed/approved I'll not merge this yet, as then we'd have to block any other PR due to the failing `LiT itest` CI step.

PR #1751, opened 2025-08-20
rfq+rfqmsg: relay errors received from price oracle

### LINKED ISSUE #1326: [bug]: clearer error with price oracle that does not support the asset requested
I have a price oracle that is not configured for the asset that tapd is trying to get a quote for

In this case the price oracle responds with

```
error {
  message: "unsupported subject asset"
}
```

However, when I try to create an invoice, tapd responds with

```
peer 0295541b89d8f41745177da175a1e7d3968fb7ceea4ca3b755f98e92ba3ef9a823 rejected the quote, code: 0, error message: unknown reject error
```

tapd should include the "unsupported subject asset" message that the price oracle gave to it.

jtobin:
Forwards errors received from a price oracle back to the requesting peer in the 'Reject' message that it is sent. An error is simply rendered as a string and included in the 'msg' field of the RejectErr, which seems sufficient.

Resolves #1326.

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -147,6 +147,10 @@
   [PR](https://github.com/lightninglabs/taproot-assets/pull/1640) addresses the
   issue.
 
+- Errors received from a price oracle [are now relayed to a requesting
+  peer](https://github.com/lightninglabs/taproot-assets/pull/1751) in
+  the reject message sent to them, instead of being ignored.
+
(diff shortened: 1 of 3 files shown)


jtobin:
@GeorgeTsagk pointed out out-of-band that this is probably a bad idea, since we may not actually want to share with a peer the reason we're not giving them a quote, and I'm inclined to agree.

jtobin:
Yeah, I'm pretty sure this one is misguided. Going to close for the time being.

PR #1759, opened 2025-08-22
docs: add generated and reviewed developer deep-dive docs

guggero:
This PR adds three documents that were initially generated by agentic LLMs but now reviewed by me.
I've read through all three documents (kudos to @Roasbeef and @bhandras for their prompts that generated very high-quality initial drafts) and edited them where necessary.

With the remaining time available, I aimed for 80 to 90% accuracy. Meaning the documents shouldn't be taken for the absolute truth (there is a disclaimer at the top) but should in any case be a valuable entry point into learning about `tapd` for any developer (with one document targeting _users_ of `tapd` and its RPCs and the other document _contributors_ to `tapd`'s code itself).


litbot-9000:
@bhandras: review reminder
@jtobin: review reminder

Roasbeef (approved):
LGTM 🪉

PR #1766, opened 2025-08-25
rfq+rfqmsg: add structured price oracle error codes

### LINKED ISSUE #1749: [feature]: structured error codes for price oracles
The error code in an `OracleError` potentially returned by a price oracle is, at present, an unstructured uint8. We simply want to start adding some structure here, defining useful codes that price oracles can use to report errors (e.g. asset not supported) and that we can handle gracefully in turn.

This subsumes e.g. #1326 and possibly #1501.

jtobin:
Adds some structure to price oracle error codes, along with some custom handling for them. In particular, a code of '1' now corresponds to an 'unsupported asset' error, which, considered a 'public' error, is forwarded in the customizable message field of the reject messages sent to peers. An error with any other code continues, at present, to be treated as an unspecified error.

Resolves #1749, #1326.

(This is a refinement over the closed #1751, which forwarded errors indiscriminately.)


docs/examples/basic-price-oracle/main.go
```diff
@@ -326,6 +327,7 @@ func (p *RpcPriceOracleServer) QueryAssetRates(_ context.Context,
 			Result: &oraclerpc.QueryAssetRatesResponse_Error{
 				Error: &oraclerpc.QueryAssetRatesErrResponse{
 					Message: "unsupported subject asset",
+					Code: 1,
```
ffranr: is there an enum val from our lib that we can use here instead of `1`?

docs/release-notes/release-notes-0.7.0.md
```diff
@@ -236,5 +244,6 @@
 
 - ffranr
 - George Tsagkarelis
+- jtobin
```
ffranr: 🔥

rfq/negotiator.go
```diff
@@ -146,10 +147,9 @@ func (n *Negotiator) queryBuyFromPriceOracle(assetSpecifier asset.Specifier,
 	}
 
 	// Now we will check for an error in the response from the price oracle.
-	// If present, we will convert it to a string and return it as an error.
+	// If present, we will simply relay it.
 	if oracleResponse.Err != nil {
-		return nil, fmt.Errorf("failed to query price oracle for "+
-			"buy price: %s", oracleResponse.Err)
+		return nil, oracleResponse.Err
```
ffranr: I think the additional context that was here previously was useful to distinguish between buy/sell price related error. Is there a good reason for removing the additional context here?

rfq/oracle.go
```diff
@@ -86,12 +86,26 @@ const (
...
 }
 
+// OracleErrorCode uniquely identifies the kinds of error an oracle may
+// return.
+type OracleErrorCode uint8
+
+const (
+	// ErrUnspecifiedOracleError represents the case where the oracle has
+	// declined to give a more specific reason for the error.
+	ErrUnspecifiedOracleError OracleErrorCode = iota
```
ffranr: I think we usually start error type names with `Err...`. Maybe we should use names `UnspecifiedOracleErrorCode` and `UnsupportedAssetOracleErrorCode` here.

rfqmsg/reject.go
```diff
@@ -92,6 +92,15 @@ var (
 	}
 )
 
+// ErrRejectWithCustomMsg produces the "unknown" error code, but pairs
+// it with a custom error message.
+func ErrRejectWithCustomMsg(msg string) RejectErr {
+	return RejectErr{
+		Code: 0,
```
ffranr: use enum const here instead of 0

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -81,6 +81,16 @@ enum Intent {
...
 
+// ErrorCode represents the possible error codes that can be returned in a
+// QueryAssetRatesErrResponse.
+enum ErrorCode {
+    // ERROR_UNSPECIFIED indicates an unspecified error.
+    ERROR_UNSPECIFIED = 0;
+
+    // UNSUPPORTED indicates the asset is not supported.
+    ERROR_UNSUPPORTED = 1;
+}
```
ffranr, line 93: I think these fields should follow the enum filed names in our code so: `UNSUPPORTED_ASSET_ORACLE_ERROR_CODE` etc

docs/examples/basic-price-oracle/main.go
jtobin: I used a raw '1' at present to avoid updating the mock oracle's dependencies (it looks like there have been other RPC changes since it was last updated as well). Perhaps best to avoid changing the mock oracle for now, and just update it to use the latest RPC definitions in another PR?

taprpc/priceoraclerpc/price_oracle.proto
jtobin: Agreed, fixed in [6ff7217](https://github.com/lightninglabs/taproot-assets/pull/1766/commits/6ff7217b8b316954496798e94aca420edf3d763e).

rfq/oracle.go
jtobin: Used your suggestion, fixed in [6130dac](https://github.com/lightninglabs/taproot-assets/pull/1766/commits/6130dac62e9e3735aaf2017475a6cebd464ff3cd).

rfqmsg/reject.go
jtobin: Fixed in [79308a2](https://github.com/lightninglabs/taproot-assets/pull/1766/commits/79308a2ff3e520c518a2ac181c174202ef17ea48). There were also some existing cases in that file that used the raw code, so I fixed those too.

rfq/negotiator.go
jtobin: Agree, yeah. Re-added context in a structured form in  [5eaf8a9](https://github.com/lightninglabs/taproot-assets/pull/1766/commits/5eaf8a90e9e13600b980ee15acdb8c4ec577b984).

docs/examples/basic-price-oracle/main.go
jtobin: (If you agree, I'll drop the commit that changes the mock oracle before merge.)

rfq/negotiator.go
```diff
@@ -493,6 +498,36 @@ func (n *Negotiator) HandleIncomingSellRequest(
...
+	if errors.As(err, &oracleError) {
+		// The error is of the expected type, so switch on the error
+		// code returned by the oracle. If the code is benign, then the
+		// RejectErr will simply relay the oracle's message. Otherwise,
+		// we'll return an opaque rejection message.
+		switch oracleError.Code {
+
+		// The rejection message will state that the oracle doesn't
+		// support the asset.
+		case ErrUnsupportedOracleAsset:
```
GeorgeTsagk: in the future would we append to this `switch` all the cases for codes that are considered public?

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -228,7 +238,7 @@ message QueryAssetRatesErrResponse {
     string message = 1;
 
     // code is the error code.
-    uint32 code = 2;
+    ErrorCode code = 2;
```
GeorgeTsagk, line 242: Worth noting that this doesn't break backwards compatibility as both `uint32` and `enum` are encoded as varints on the wire.

rfq/oracle.go
```diff
@@ -380,6 +380,19 @@ func (r *RpcPriceOracle) QuerySellPrice(ctx context.Context,
 	}
 }
 
+// marshallErrorCode marshalls an over-the-wire error code into an
+// OracleErrorCode.
+func marshallErrorCode(code oraclerpc.ErrorCode) OracleErrorCode {
+	switch code {
+	case oraclerpc.ErrorCode_ERROR_UNSPECIFIED:
+		return ErrUnspecifiedOracleError
+	case oraclerpc.ErrorCode_ERROR_UNSUPPORTED:
```
GeorgeTsagk: we add an empty line before a new switch case
```suggestion

	case oraclerpc.ErrorCode_ERROR_UNSUPPORTED:
```

GeorgeTsagk: except for the first case (but including the `default` case below)

docs/examples/basic-price-oracle/main.go
GeorgeTsagk: I think that if we want to go ahead and update this price oracle it should include a version bump too.

Having said that, I believe it's okay to just leave the `Code` field unspecified and let it default to `0`, in order to not have to populate it with magic numbers

rfq/negotiator.go
```diff
@@ -509,7 +509,6 @@ func createCustomRejectErr(err error) rfqmsg.RejectErr {
 		// RejectErr will simply relay the oracle's message. Otherwise,
```
GeorgeTsagk: eventually we should squash all the following fix related commits to their original commit

rfq/negotiator.go
```diff
@@ -501,28 +540,28 @@ func (n *Negotiator) HandleIncomingSellRequest(
 // createCustomRejectErr creates a RejectErr with code 0 and a custom message
 // based on an error response from a price oracle.
 func createCustomRejectErr(err error) rfqmsg.RejectErr {
+	var queryError *QueryError
+	// Check if the error we've received is the expected QueryError, and
```
GeorgeTsagk: newline before the comments
```suggestion
        var queryError *QueryError
        
	// Check if the error we've received is the expected QueryError, and
```

rfq/negotiator.go
GeorgeTsagk: If I read the `iota` list of all error codes I don't really have some information source as to whether they are public or internal error codes.

We could follow a convention that all non-zero codes are public (with appropriate docs in the definitions ofcs)

Or further extend the oracle error with a direct `Internal bool` flag, that explicitly indicates whether this code/msg should be shared or not

rfqmsg/reject.go
ffranr: I think we should squash or amend the commits in this PR.

Also, to stay consistent with the naming pattern used elsewhere in the code, we should probably call this `NewRejectErr`.

docs/examples/basic-price-oracle/main.go
ffranr: > I used a raw '1' at present to avoid updating the mock oracle's dependencies (it looks like there have been other RPC changes since it was last updated as well). Perhaps best to avoid changing the mock oracle for now, and just update it to use the latest RPC definitions in another PR?

I think updating the mock oracle in a separate PR makes sense to me. We can wait until 0.7 for example.

rfq/negotiator.go
jtobin: Yeah I like that latter idea -- let the oracle indicate whether the error is intended to be public/private. :+1:

docs/examples/basic-price-oracle/main.go
jtobin: Yeah, let's update the mock oracle in a separate PR. :+1:

rfqmsg/reject.go
```diff
@@ -76,22 +76,42 @@ func (v *RejectErr) Record() tlv.Record {
...
+const (
+	// UnspecifiedRejectCode indicates that a request-for-quote was
+	// rejected, without necessarily providing any further detail as to
+	// why.
+	UnspecifiedRejectCode uint8 = iota
+
+	// UnavailableRejectCode indicates that a request-for-quote was
+	// rejected as a price oracle was unavailable.
+	UnavailableRejectCode
+)
```
ffranr, line 92: It’s better, in my view, not to use `iota`. Reordering the `const` fields becomes a silent breaking change, and it’s easy to miss in review, especially as more error types are added.

ffranr: Should we turn this set of const values into a `RejectCode` enum type? We could then use it in `RejectErr` instead of a plain `uint8`. But I can see that just adding the `const` values is a win for now. So I wont block approve on this.

ffranr: Maybe we should make the name more specific than `UnavailableRejectCode`. Something like `PriceOracleUnavailableRejectCode` might be clearer. It’s long, but many things could be unavailable, so the extra clarity feels worthwhile.

rfq/oracle.go
```diff
@@ -87,12 +87,26 @@ const (
...
+
+const (
+	// UnspecifiedOracleErrorCode represents the case where the oracle has
+	// declined to give a more specific reason for the error.
+	UnspecifiedOracleErrorCode OracleErrorCode = iota
+
+	// UnsupportedAssetOracleErrorCode represents the case in which an
+	// oracle does not provide quotes for the requested asset.
+	UnsupportedAssetOracleErrorCode
+)
```
ffranr, line 108: same comment here re `iota`

taprpc/priceoraclerpc/price_oracle.proto
```diff
@@ -238,8 +238,11 @@ message QueryAssetRatesErrResponse {
     // error is the error message.
     string message = 1;
 
+    // public is a flag indicating that the error can be forwarded to peers.
+    bool public = 2;
+
```
ffranr: I don’t think the price oracle should decide whether a tapd node forwards an error to its peer. That decision belongs entirely to tapd in my view. So IMO the price oracle shouldn’t signal whether an error is “public” or not.

We want to be able to query price oracle services operated by third parties.

rfqmsg/reject.go
jtobin: All sound, I've since made all of these changes.

taprpc/priceoraclerpc/price_oracle.proto
jtobin: Hmm. This one might be a little tricky; I agree that the price oracle shouldn't be able to *determine* what gets relayed, but I feel it's beneficial for it to be able to mark an error as being public/benign/etc., even if that's only the oracle's opinion on that. One is trusting the oracle's opinion on prices already anyway.

I would be inclined to start with "trust the oracle," but maybe a satisfying solution would be to be able to set a config option to run tapd in "private mode," in which tapd only forwards opaque rejection errors? I.e., even if an oracle has indicated that something should be considered public?

ffranr: What do we gain from this field? tapd can already inspect the error returned by the price oracle. IMO there is no need to frame this in terms of trusting or not trusting the oracle. tapd has all the data it needs and simply needs to decide what to return to its peer.

jtobin: I think what 'public' does is spare us from determining how particular error codes should be handled *outside of* relaying or not relaying the error to peers. This idea (from discussion with @GeorgeTsagk on the rfq/negotiator.go diff [here](https://github.com/lightninglabs/taproot-assets/pull/1766#pullrequestreview-3229263028)) struck me as elegant because relaying is the only thing I know we definitely want to be able to do at present.

But instead of hard-coding tapd to handle 'unsupported asset' errors by relaying them to peers, for example, we can just rely on the presence of the public flag to decide to relay them, and delay figuring out what *else* we might want to do with that error until some other time.

Another option would be to add a 'NonspecificPublicOracleErrorCode' to the codes list, which would we would hard-code to handle by relay, but that strikes me as inferior to using the 'public' flag.

jtobin: @ffranr and I chatted further out-of-band and he persuaded me to scrap 'public' for the time being. My thinking here is that even if a 'public' field could ultimately prove to be desirable, we don't *need* it now, so let's just start with something simpler and less questionable.

rfq/negotiator.go
jtobin: Coming back to this: I decided to axe the 'public' flag for this changeset, so customRejectErr is again the arbiter of whether we forward opaque or more detailed error messages, based on the code value. I do like the 'public' idea, in the meantime, as a kind of hint from the oracle as to whether an error is appropriate to relay, so maybe I can make another argument for it in the future.

I think there's no immediate need to codify a hard public/private scheme at the moment. We may want other nonzero codes to be considered private, for example. I'd be inclined to pick that issue up again when it's time to add/handle more codes.

ffranr (approved):

litbot-9000:
@georgetsagk: review reminder

GeorgeTsagk (approved):
LGTM

PR #1771, opened 2025-08-28
tapgarden: reduce flakiness in TestBatchedAssetIssuance

ffranr:
Assign a separate minting store to each test case to avoid shared state and improve test reliability.

I think I've noticed this unit test flake a few times in CI.


jtobin (approved):
Makes sense, LGTM.

PR #1772, opened 2025-08-28
fn: fix loop var capture and clarify naming in `ParSliceErrCollect`

ffranr:
Ensure each goroutine captures its own index/value in `ParSliceErrCollect`, and rename variables for clearer intent. Avoids closure capture bug and makes concurrency guards explicit.

### DIFF
--- a/fn/concurrency.go
+++ b/fn/concurrency.go
@@ -26,7 +26,12 @@ func ParSlice[V any](ctx context.Context, s []V, f ErrFunc[V]) error {
 	errGroup.SetLimit(runtime.GOMAXPROCS(0))
 
 	for _, v := range s {
+		// Snapshot v now so the goroutine sees the intended value
+		// even if the caller mutates the slice s later. This is a
+		// shallow copy only, if V contains pointers, their contents can
+		// still change.
 		v := v
+
 		errGroup.Go(func() error {
 			return f(ctx, v)
 		})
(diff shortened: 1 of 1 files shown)


fn/concurrency.go
```diff
@@ -43,25 +48,46 @@ func ParSlice[V any](ctx context.Context, s []V, f ErrFunc[V]) error {
...
+
+		// Snapshot s[idx] now so the goroutine sees the intended value
+		// even if the caller mutates the slice later. This is a shallow
+		// copy only, if V contains pointers, their contents can still
+		// change.
+		v := s[idx]
+
 		errGroup.Go(func() error {
-			err := f(ctx, s[idx])
+			// If already canceled, skip work without signaling a
```
jtobin: nit: s/a/an

fn/concurrency.go
```diff
@@ -43,25 +48,46 @@ func ParSlice[V any](ctx context.Context, s []V, f ErrFunc[V]) error {
...
 				instanceErrors[idx] = err
-				instanceErrorsMutex.Unlock()
+				instanceErrorsMu.Unlock()
 			}
 
 			// Avoid returning an error here, as that would cancel
 			// the errGroup and terminate all slice element
 			// processing instances. Instead, collect the error and
-			// return it later.
+			// return them later.
```
jtobin: nit: if 'them', then 'collect the _errors_ and..'.

fn/concurrency.go
```diff
@@ -71,8 +97,7 @@ func ParSliceErrCollect[V any](ctx context.Context, s []V,
 	// The goroutines that are executing in parallel should not return an
 	// error, but the Wait call may return an error if the context is
 	// canceled or timed out.
-	err := errGroup.Wait()
-	if err != nil {
+	if err := errGroup.Wait(); err != nil {
```
jtobin, line 93: Regarding the underlying function: does this behave as intended?

If the context is cancelled or times out, then it looks like every goroutine returns nil, so errGroup.Wait() will also return nil? I think the result is that we'll return a (possibly-partial) instanceErrors map and nil, but the intent here seems to be that we should return an error?

jtobin (commented):
Changes look good, but worth checking on the underlying behaviour of ParSliceErrCollect per my comment/question.

fn/concurrency.go
```diff
@@ -43,25 +48,46 @@ func ParSlice[V any](ctx context.Context, s []V, f ErrFunc[V]) error {
...
+	var (
+		instanceErrorsMu sync.Mutex
+		instanceErrors   = make(map[int]error, len(s))
+	)
 
 	for idx := range s {
+		// Rebind idx inside the loop so that each goroutine captures
+		// a distinct copy of the index. Without this, all goroutines
+		// may observe the same final idx value due to closure capture.
+		idx := idx
```
Roasbeef: This is [solved in Go 1.22](https://go.dev/blog/loopvar-preview)+ tho?

fn/concurrency.go
ffranr: Yes I think you're right. Good point and well spotted!

Since our goroutines always return nil, `errGroup.Wait()` will not surface context cancellation. I added an explicit `ctx.Err()` check after `Wait()` so we return a partial error map and a non-nil error if the caller’s context was canceled or timed out.

Does that look correct to you now?

jtobin: Nice, looks good!

jtobin (approved):
LGTM :+1: :+1:

PR #1775, opened 2025-08-30
rfq: add price oracle certificate verification

### LINKED ISSUE #1278: [bug]: price oracle TLS/SSL certificate validation missing
The price oracle client inside tapd does not seem to do SSL certificate validation of the price oracle server. This is a security issue as we have no confidence that we trust the price oracle we are talking to is the one we think we are talking to.

We should have the following choices:

1. Root CAs to trust (in addition to or instead of the operating system root CA list)
2. Trust operating system root CA list (yes or no)
3. Require certificate checking
4. Don't require certificate checking
5. Pin to a specific certificate (either signed by a CA or self signed)

jtobin:
Adds certificate verification for TLS communication with price oracles, mostly following the suggestions proposed in #1278. Adds configuration options for skipping certificate verification, distrusting the operating system's root CA list, and using a custom certificate.

Resolves #1278.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -133,6 +133,12 @@
   - Added a limit (`MaxOrphanUTXOs = 20`) to prevent transactions from becoming
     too large when sweeping many orphan UTXOs at once.
 
+- [PR#1775](https://github.com/lightninglabs/taproot-assets/pull/1775):
+  Price oracle connections now verify TLS certificates by
+  default, using the OS root CA list. New config options under
+  `experimental.rfq.priceoracle*` allow disabling TLS, skipping
+  verification, or specifying custom certificates.
+
 ## RPC Updates
 
 - [PR#1766](https://github.com/lightninglabs/taproot-assets/pull/1766):


jtobin:
(Changed this from draft; I think the litd tests are failing for an unrelated reason.)

jtobin:
(As pointed out by @ZZiigguurraatt, to be more precise: TLS support already existed for price oracles, but certificate verification was skipped entirely.)

rfq/cli.go
```diff
@@ -22,6 +22,14 @@ const (
 type CliConfig struct {
 	PriceOracleAddress string `long:"priceoracleaddress" description:"Price oracle gRPC server address (rfqrpc://<hostname>:<port>). To use the integrated mock, use the following value: use_mock_price_oracle_service_promise_to_not_use_on_mainnet"`
 
+	PriceOracleTLSDisable bool `long:"priceoracletlsdisable" description:"Disable TLS for price oracle communication."`
+
+	PriceOracleTLSInsecure bool `long:"priceoracletlsinsecure" description:"Disable price oracle certificate verification."`
+
+	PriceOracleTLSNoSystemCAs bool `long:"priceoracletlsnosystemcas" description:"Disable use of the operating system's list of root CA's when verifiying price oracle certificates."`
```
jtobin: typo: s/verifiying/verifying

rfq/tls_test.go
```diff
@@ -0,0 +1,130 @@
...
+		{
+			name:           "invalid custom certificate",
+			expectInsecure: false,
+			tlsConfig: &TLSConfig{
+				Enabled:            true,
+				InsecureSkipVerify: false,
+				TrustSystemRootCAs: false,
+				CustomCertificates: []byte(invalidCertificate),
+			},
+		},
```
ffranr, line 130: I'm not sure I see the purpose of `invalidCertificate` here. It doesn't look like the test actually exercises its invalidity.

More broadly, do we need certificate examples in our unit tests at all? It seems like we're testing the behavior of the underlying TLS/certificate library rather than the logic we're adding on top of it.

rfq/cli.go
jtobin: Resolved in 6a369476.

rfq/tls_test.go
jtobin: I don't think these tests are super important, but they perform a decent sanity check of the configureTransportCredentials function -- basically, that it uses the supplied config to interact with the underlying libraries in the way that I expect. The tests are cheap, so why not, sort of.

I debated deleting the invalid certificate case, but wound up deciding to return an error in configureTransportCredentials when no valid custom certificates could be found anyway, so *now* the invalid certificate case exercises the invalidity of the certificate. 😄

jtobin:
> With that in mind, I wonder if we could define a more general, reusable solution in something like the new rfq/tls.go file, especially given the need for configuration and the importance of which package owns this logic.

I looked into this and agree that it's probably worth centralizing our TLS handling, but I'd consider it out of scope for this change set, which does resolve a concrete issue as-is. Maybe we can open a broader "refactor TLS handling" issue for that?

I get the following hits for "crypto/tls" imports, as a rough metric:

```
$ ack -l "crypto/tls" --ignore-dir=itest --ignore-dir=docs
cmd/commands/conn.go
proof/courier.go
internal/test/grpc.go
universe_rpc_registrar.go
tapcfg/config.go
rfq/tls.go
authmailbox/client.go
universe/proof_download_test.go
```

jtobin:
The LiT itest failure should be resolved by lightninglabs/lightning-terminal#1190.

jtobin:
@GeorgeTsagk @ffranr this one should be good for {re-}review whenever.

litbot-9000:
@ffranr: review reminder

rfq/oracle.go
```diff
@@ -201,84 +198,49 @@ type RpcPriceOracle struct {
...
-		insecure.NewCredentials(),
-	))
-
-	opts = append(opts, clientKeepaliveDialOption)
-
-	return opts, nil
-}
-
 // NewRpcPriceOracle creates a new RPC price oracle handle given the address
 // of the price oracle RPC server.
```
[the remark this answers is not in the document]
ffranr: We only need to list non-obvious args in the doc comment.

rfq/oracle.go
```diff
@@ -219,84 +216,49 @@ type RpcPriceOracle struct {
...
+				// requests.
+				PermitWithoutStream: true,
+			},
+		),
 	}
 
 	// Formulate the server address dial string.
 	serverAddr := fmt.Sprintf("%s:%s", addr.Hostname(), addr.Port())
 
+	// Connect to the RPC server.
```
ffranr: nit: unnecessary comment

rfq/oracle_test.go
```diff
@@ -141,7 +141,8 @@ func runQuerySalePriceTest(t *testing.T, tc *testCaseQuerySalePrice) {
 
 	// Create a new RPC price oracle client and connect to the mock service.
 	serviceAddr := fmt.Sprintf("rfqrpc://%s", testServiceAddress)
-	client, err := NewRpcPriceOracle(serviceAddr, true)
+	insecureTLS := &TLSConfig{Enabled: false}
```
ffranr: could use `insecureTLS()` here

rfq/tls.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// TLSConfig represents TLS configuration options for oracle connections.
+type TLSConfig struct {
+	// Enabled indicates that we should use TLS.
+	Enabled bool
```
ffranr: This will default to `false`. Maybe its worth flipping to `DisableTLS bool` which means that we would have to explicitly disable TLS by setting it to `true`. Not a big deal either way IMO.

rfq/tls.go
```diff
@@ -0,0 +1,75 @@
...
+	// pool.
+	if len(config.CustomCertificates) > 0 {
+		ok := certPool.AppendCertsFromPEM(config.CustomCertificates)
+		if !ok {
+			return nil, fmt.Errorf("failed to parse any valid " +
+				"custom certificates")
+		}
+	}
+
+	// Return the constructed transport credentials.
```
ffranr: I think that comments like these are unnecessary.

tapcfg/config.go
```diff
@@ -1209,6 +1234,37 @@ func getCertificateConfig(cfg *Config, cfgLogger btclog.Logger) (*tls.Config,
...
+// getPriceOracleTLSConfig returns a TLS configuration for a price oracle,
+// given a valid RFQ CLI configuration.
+func getPriceOracleTLSConfig(rfqCfg rfq.CliConfig) (*rfq.TLSConfig, error) {
+	var certBytes []byte
+
+	// Read any specified certificate data.
+	if rfqCfg.PriceOracleTLSCertPath != "" {
+		var err error
+		certBytes, err = os.ReadFile(
+			rfqCfg.PriceOracleTLSCertPath)
```
ffranr: nit: trailing `)` should be on new line

tapcfg/config.go
```diff
@@ -339,8 +357,11 @@ type ExperimentalConfig struct {
 	Rfq rfq.CliConfig `group:"rfq" namespace:"rfq"`
 }
 
-// Validate returns an error if the configuration is invalid.
-func (c *ExperimentalConfig) Validate() error {
+// CleanAndValidate performs final processing on the ExperimentalConfig,
+// returning an error if the configuration is invalid.
+func (c *ExperimentalConfig) CleanAndValidate() error {
+	c.Rfq.PriceOracleTLSCertPath = CleanAndExpandPath(
+		c.Rfq.PriceOracleTLSCertPath)
```
ffranr, line 364: If we use `CleanAndExpandPath` when setting the field then perhaps we can still call this method `ExperimentalConfig.Validate`. It could perhaps validate the path by deriving the expected path using `CleanAndExpandPath` and asserting equality.

tapcfg/config.go
```diff
@@ -1209,6 +1234,37 @@ func getCertificateConfig(cfg *Config, cfgLogger btclog.Logger) (*tls.Config,
...
+			return nil, fmt.Errorf("unable to read "+
+				"price oracle certificate: %w", err)
+		}
+	}
+
+	// Construct the oracle's TLS configuration.
+	tlsConfig := &rfq.TLSConfig{
+		// Note the subtle flip on the flag, since the user has
+		// configured whether to *disable* TLS.
+		Enabled:            !rfqCfg.PriceOracleTLSDisable,
```
ffranr: Maybe we can skip this complexity by swapping `Enabled` for `DisableTLS` like i mentioned in another comment. Not sure if my suggestion works broadly, just an idea to consider.

ffranr (commented):
Some trivial comments to be removed, for example:
```
// Construct the oracle's TLS configuration.
```

Otherwise, just nits. I can quickly approve from here.

tapcfg/config.go
jtobin: (I didn't address this one for the time being.)

jtobin:
Note that the failing itest case passes with the changes in lightninglabs/lightning-terminal#1190.

Issue #1452, opened 2025-03-22
[feature]: serve+sync canonical Universe commitments over RPC


Roasbeef:
**Is your feature request related to a problem? Please describe.**

We don't have any way to query/serve the cannonical universe commitment information. 

Once https://github.com/lightninglabs/taproot-assets/issues/1451 is finished, we should start to serve the universe tree information over RPC. 

**Describe the solution you'd like**

To begin with, for the verifiers of grouped assets that have a canonical Universe server set, we should add that set of Universe servers to our federation (only pulling information from them). We should mark them as special Universe servers in our database. Each time we verify a new grouped asset, we should look out to see if the target servers changes, as authenticated by the delegation key.

To periodically sync this information (insert it into our local database, possibly serving it to others), we can re-use the same syncer logic, but now we're syncing a different type of Multi-verse tree. The top level is keyed by the proof type (ignore, burn, issuance), and the tree's key is specific to the proof type. 

In order to _accept_ this information, we need to consult our canonical commitment verifier, in addition to the normal MS-SMT proofs, we also require a new block merkle header proof which shows that the information we're being served is committed to within the last unspent commitment transaction.

Roasbeef:
@ffranr they seem related, no? How can the verifier fetch the new information if the new RPCs aren't available?

PR #1777, opened 2025-09-01
supplyverifier: add pull syncer and non-universe node verification

ffranr:
Extend the `supplyverifier` syncer to support pulling supply commitments from universe servers.

Add a `supplyverifier` state machine so that non-universe tapd nodes (regular node) can verify supply commitments by:

* Watching for on-chain spends
* Pull-syncing from universe servers

Fixes https://github.com/lightninglabs/taproot-assets/issues/1452
Fixes https://github.com/lightninglabs/taproot-assets/issues/1463

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -73,6 +73,7 @@
    - https://github.com/lightninglabs/taproot-assets/pull/1675
    - https://github.com/lightninglabs/taproot-assets/pull/1674
    - https://github.com/lightninglabs/taproot-assets/pull/1784
+   - https://github.com/lightninglabs/taproot-assets/pull/1777
 
 - A new [address version 2 was introduced that supports grouped assets and
   custom (sender-defined)


Roasbeef:
Can be rebased now.

universe/supplycommit/interface.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/mssmt"
+	"github.com/lightninglabs/taproot-assets/universe"
+)
+
+// SubtreeRootProof represents the root of a supply commit subtree with its main
+// supply tree inclusion proof.
+type SubtreeRootProof struct {
```
Roasbeef, line 12: 👍

universe/supplycommit/interface.go
```diff
@@ -0,0 +1,57 @@
...
+	// RootNode is the root node of the supply commit subtree.
+	RootNode mssmt.BranchNode
+
+	// SupplyTreeLeafKey locates the subtree leaf node in the supply commit
+	// tree.
+	SupplyTreeLeafKey universe.UniverseKey
+
+	// SupplyTreeInclusionProof proves inclusion of the subtree root in the
+	// supply tree.
+	SupplyTreeInclusionProof []byte
```
Roasbeef: Any reason this isn't the concrete `mssmt` type?

universe/supplyverifier/syncer.go
```diff
@@ -21,6 +28,12 @@ type UniverseClient interface {
 		updateLeaves supplycommit.SupplyLeaves,
 		chainProof supplycommit.ChainProof) error
 
+	// FetchSupplyCommit fetches a supply commitment for a specific
+	// asset group from the remote universe server.
+	FetchSupplyCommit(ctx context.Context, assetSpec asset.Specifier,
+		spentCommitOutpoint fn.Option[wire.OutPoint]) (
```
Roasbeef, line 35: I really dig the design here re fetching commitments: you're always either interested in either the very first commitment, or a new commitment that spends one you already know of!

universe/supplyverifier/syncer.go
```diff
@@ -214,11 +227,156 @@ func (s *SupplySyncer) PushSupplyCommitment(ctx context.Context,
...
-		errorMap[hostStr] = fetchErr
+		errorMap[hostStr] = pushErr
 	}
 
 	return errorMap, nil
 }
+
+// SupplyCommitPullResult represents the result of a supply commitment pull
+// operation across multiple universe servers.
+type SupplyCommitPullResult struct {
```
Roasbeef: Seems like this could be just `fn.Result`?

Roasbeef: Style nit: can move this one declaration below, to sit right above the `PullSupplyCommitment` method.

universe/supplyverifier/syncer.go
```diff
@@ -214,11 +227,156 @@ func (s *SupplySyncer) PushSupplyCommitment(ctx context.Context,
...
+			)
+			if err != nil {
+				return fmt.Errorf("unable to pull supply "+
+					"commitment (server_addr_id=%d, "+
+					"server_addr_host_str=%s): %w",
+					serverAddr.ID, serverAddr.HostStr(),
+					err)
+			}
+
+			results[serverAddr.HostStr()] = result
```
Roasbeef, line 334: This will likely trigger the "concurrent write to map" race condition error. 

An alternative here would be to use a buffered results channel.

universe/supplyverifier/syncer.go
```diff
@@ -214,11 +227,156 @@ func (s *SupplySyncer) PushSupplyCommitment(ctx context.Context,
...
+	// Return one successful result, if available, as the final outcome.
+	var finalResult *supplycommit.FetchSupplyCommitResult
+	for _, res := range results {
+		finalResult = &res
+		break
+	}
+
+	// Build a map from server addresses to their corresponding errors.
+	errorMap := make(map[string]error)
+	for idx, pullErr := range pullErrs {
```
Roasbeef, line 375: Can we actually rely on the `pullErrs` population ordering?

universe/supplycommit/manager.go
```diff
@@ -287,11 +258,64 @@ func (m *Manager) fetchStateMachine(
...
+	// Check if the state machine for the asset group already exists in the
+	// cache.
+	sm, ok := m.smCache.Get(*groupKey)
+	if ok {
+		// If the state machine is found and is running, return it.
+		if sm.IsRunning() {
+			return sm, nil
+		}
+
+		// If the state machine exists but is not running, replace it in
```
Roasbeef, line 233: In what case would it be int he cache, but not running?

tapdb/addrs.go
```diff
@@ -1637,6 +1640,102 @@ func (t *TapAddressBook) FetchInternalKeyLocator(ctx context.Context,
 	return keyLoc, nil
 }
 
+// FetchSupplyCommitAssets fetches all assets with non-nil group keys that could
+// potentially be involved in supply commitments.
+func (t *TapAddressBook) FetchSupplyCommitAssets(ctx context.Context,
```
Roasbeef, line 1645: Is this intended for verifiers, or issuers? 

If it's issuers, then there's a more direct way to query for this, as we can look at the minting state on disk.

tapdb/addrs.go
```diff
@@ -1637,6 +1640,102 @@ func (t *TapAddressBook) FetchInternalKeyLocator(ctx context.Context,
...
+				dbAsset.TweakedGroupKey,
+			)
+			if err != nil {
+				return fmt.Errorf("unable to parse group "+
+					"key: %w", err)
+			}
+
+			// Get asset metadata for this group to check if it
+			// supports supply commitments.
+			metaRow, err := db.FetchAssetMetaForAsset(
```
Roasbeef, line 1675: I think we should be able to push some of this logic more into the SQL query itself. As we can join the asset to the meta, then filter based on the universe commit information, etc.

universe/supplyverifier/manager.go
```diff
@@ -20,6 +20,10 @@ import (
 const (
 	// DefaultTimeout is the context guard default timeout.
 	DefaultTimeout = 30 * time.Second
+
+	// MaxStateMachines is the maximum number of supply verifier state
+	// machines that can be managed by the multi state machine manager.
+	MaxStateMachines = 50
```
Roasbeef: What's the rationale behind this limit again? 

For all the other universe RPC calls, we already implement some strict QPS based rate limiting. We should also ensure that for these new set of supply commitment related calls, that we're still maintaining that rate limiting.

universe/supplyverifier/manager.go
```diff
@@ -121,13 +129,87 @@ func NewManager(cfg ManagerCfg) *Manager {
...
+
+	for idx := range assetGroupKeys {
+		groupKey := assetGroupKeys[idx]
+
+		// Create asset specifier from group key.
+		assetSpec := asset.NewSpecifierFromGroupKey(groupKey)
+
+		// Check to ensure state machine for asset group does not
+		// already exist.
+		_, ok := m.smCache.Get(groupKey)
```
Roasbeef, line 196: Just a defensive check? As we shouldn't ever get back a series of duplicate group keys.

universe/supplyverifier/manager.go
Roasbeef: Also if we're polling right after a block is mined, then we can add some jitter to the system via random back off before we attempt to query the server.

universe/supplyverifier/manager.go
```diff
@@ -121,13 +129,87 @@ func NewManager(cfg ManagerCfg) *Manager {
...
 	m.startOnce.Do(func() {
 		// Initialize the state machine cache.
 		m.smCache = newStateMachineCache()
+
+		// Initialize state machines for all relevant asset groups.
+		//
+		// TODO(ffranr): Consider passing in tapd operation mode
+		//  (e.g. universe server, etc) to determine which asset groups
+		//  we should run state machines for.
+		err := m.InitStateMachines()
```
Roasbeef, line 225: As in the ability to turn off supply commitment verification?

universe/supplycommit/env.go
```diff
@@ -271,6 +273,68 @@ func fetchLatestAssetMetadata(ctx context.Context, lookup AssetLookup,
...
+		fmt.Errorf("delegation key not found for given asset"),
+	)
+	if err != nil {
+		return err
+	}
+
+	// Fetch the delegation key locator. We need to ensure that the
+	// delegation key is owned by this node, so that we can generate
+	// supply commitments (ignore tuples) for this asset group.
+	_, err = assetLookup.FetchInternalKeyLocator(
```
Roasbeef: If I understand your future plans correctly, then another way we'd be able to determine this is if is ours or not is via the presence of an entry in the new pre commitment table.

universe/supplyverifier/events.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Event is a special interface used to create the equivalent of a sum-type, but
+// using a "sealed" interface.
+type Event interface {
+	eventSealed()
+}
+
+// Events is a special type constraint that enumerates all the possible protocol
+// events.
+type Events interface {
```
Roasbeef: Not used, can be removed? 

Usually I have these here primarily for readers of the code, as then we have a single palce where we enumerate all the possible events.

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,329 @@
...
+			}
+
+			commit, err := env.SupplyCommitView.
+				FetchCommitmentBySpentOutpoint(
+					ctx, env.AssetSpec, spentOutpoint,
+				)
+			switch {
+			case errors.Is(err, ErrCommitmentNotFound):
+				// This is the expected case, so we can
+				// continue.
```
Roasbeef: Should this break out of this scope into the code that follows?

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,329 @@
...
+			NewEvents: lfn.Some(FsmEvent{
+				InternalEvent: []Event{
+					&watchEvent,
+				},
+			}),
+		}, nil
+
+	case *SpendEvent:
+		// TODO(ffranr): This is basically the same as SyncVerifyEvent
+		//  but we add a delay before syncing because the issuer may not
```
Roasbeef: So this is just a temp state? Or a way to move forward the state machine if we detect a spend right as it's being created?

tapdb/sqlc/schemas/generated_schema.sql
```diff
@@ -892,6 +870,32 @@ CREATE UNIQUE INDEX supply_commitments_outpoint_uk
 CREATE INDEX supply_commitments_spent_commitment_idx
     ON supply_commitments(spent_commitment);
 
+CREATE TABLE supply_pre_commits (
```
Roasbeef, line 908: 👍

universe/supplycommit/manager.go
ffranr: I don’t expect cached state machines to be stopped, but since they can be, I added a check to harden the code against future changes.

universe/supplyverifier/syncer.go
ffranr: `pullErrs` is a `map[int]error`. In the loop, `idx` represents the map key, which corresponds to the index in `targetAddrs`. That’s how `fn.ParSliceErrCollect` associates errors back to their input slice elements.

Would you like me to rename `idx` here to `for keyIdx, pullErr := range pullErrs {` perhaps?

tapdb/addrs.go
ffranr: My intention here is to use this when initiating the verifier state machines at startup, so that we can set up watching for on-chain spends after a reboot. I'll improve the commit message to make that clear.

universe/supplyverifier/syncer.go
ffranr: Well spotted, thanks!

For now I'll add a mutex (because it seems like the smallest change). But I'll look into expanding `ParSliceErrCollect` with the buffered results channel like you suggest. Sounds like a nice addition.

universe/supplyverifier/manager.go
ffranr: Yes, mostly defensive, and to ensure `InitStateMachines` is idempotent, which could be useful in the future.

universe/supplyverifier/manager.go
ffranr: Yes, disable the supply commit verification state machine for the universe server. That way, universe server supply commit verification happens only on RPC inserts, without watching the chain for spends.

universe/supplycommit/env.go
ffranr: Yes, or through mint batch or seedling lookup. But with a potential distant recovery mode in mind, it might be better to rely solely on key access as a design principle. I haven’t thought that through in detail yet.

tapdb/sqlc/schemas/generated_schema.sql
ffranr: I'm going to rebase this PR (1777) on top of https://github.com/lightninglabs/taproot-assets/pull/1784 . PR 1784 will include the pre-commit db changes.

universe/supplyverifier/transitions.go
ffranr: The high-level architecture I have in mind is:

* The state machine watches on-chain outpoints for spends.
* A spend triggers `SyncVerifyState.SpendEvent`.
* `SpendEvent` simply applies a delay before transitioning to `SyncVerifyState.SyncVerifyEvent`. The delay is needed because after confirmation, the issuer peer must publish the new supply commitment to the universe server, and the server needs time to verify and store it.
* After the delay, `SyncVerifyState.SyncVerifyEvent` can sync and verify immediately, with no further waiting.

universe/supplyverifier/transitions.go
ffranr: Thanks, was still WIP, will fix.

universe/supplyverifier/syncer.go
ffranr: > Seems like this could be just `fn.Result`?

My intent is to keep both the fetch results (`FetchResult`) and a map of errors (`ErrorMap`) in the same structure, since there can be one error per universe server sync target. In that sense, it’s a custom alternative to `fn.Result`. I considered using an `Either`, but here both `FetchResult` and `ErrorMap` may be populated at the same time. But maybe the syncer doesn't need to return so much?

tapdb/addrs.go
ffranr: Yes, good point! I’ll put together an issue with a few of these simplifications and de-dupes we can make.

universe/supplyverifier/manager.go
```diff
@@ -153,50 +374,35 @@ func (m *Manager) Stop() error {
 	return nil
```
jtobin, line 374: Stop() may be missing a m.Wg.Wait() call here.

tapdb/sqlc/schemas/generated_schema.sql
```diff
@@ -640,11 +640,20 @@ CREATE TABLE "mint_supply_pre_commits" (
     -- The index of the mint batch anchor transaction pre-commitment output.
     tx_output_index INTEGER NOT NULL,
 
+    -- The asset group key for this pre-commitment.
+    -- Stored in canonical 32-byte x-only form as defined in BIP340
+    -- (schnorr.SerializePubKey).
+    group_key BLOB CHECK(length(group_key) = 32),
```
Roasbeef, line 646: Love how we can view the final generated schema for diffs (in addition to the migration ofc), makes larger migrations like this easier to review. 

Alternatively, we could just change the existing versions in place....as this was just in master.

universe/supplyverifier/verifier.go
```diff
@@ -397,6 +398,19 @@ func (v *Verifier) proofVerifierCtx(ctx context.Context) proof.VerifierCtx {
 	}
 }
 
+// IsEquivalentPubKeys reports whether two public keys are equivalent
+// when compared in their BIP340-serialized form. This avoids issues
+// with multiple encodings of the same elliptic curve point, since
+// BIP340 serialization provides a unique, canonical byte representation.
+//
+// TODO(ffranr): This should be a method on btcec.PublicKey.
+func IsEquivalentPubKeys(a, b *btcec.PublicKey) bool {
```
Roasbeef, line 407: 👍

mssmt/proof.go
```diff
@@ -41,6 +42,38 @@ func NewProof(nodes []Node) *Proof {
 	}
 }
 
+// NewProofFromCompressedBytes initializes a new merkle proof from its
+// compressed byte representation.
+func NewProofFromCompressedBytes(compressedProofBytes []byte) (Proof, error) {
```
Roasbeef, line 47: Are you aware that we already have this? https://github.com/lightninglabs/taproot-assets/blob/851993144d183b890486ac8451745b87c630972f/mssmt/encoding.go#L70-L100

universe/supplyverifier/syncer.go
Roasbeef: Ah ok, yeah it's just an integer index, it's a map key. I think it's fine as is.

universe/supplycommit/manager.go
Roasbeef: Makes sense, a little defensive programming never hurt no one.

tapdb/addrs.go
Roasbeef: Reminder for this.

universe/supplyverifier/manager.go
```diff
@@ -230,21 +212,14 @@ func (m *Manager) startAssetSM(ctx context.Context,
...
-
 	// Create a new error reporter for the state machine.
 	errorReporter := NewErrorReporter(assetSpec)
 
 	fsmCfg := protofsm.StateMachineCfg[Event, *Environment]{
 		ErrorReporter: &errorReporter,
-		InitialState:  initialState,
-		Env:           env,
-		Daemon:        m.cfg.DaemonAdapters,
+		// TODO(ffranr): Set InitialState here.
```
Roasbeef: Comment place holder to remind to keep this in place.

universe/supplyverifier/manager.go
```diff
@@ -224,6 +237,121 @@ func (m *Manager) Start() error {
...
+	if len(syncDiffEvent.SyncDiff.NewLeafProofs) == 0 {
+		return nil
+	}
+
+	// Get genesis asset ID from the first synced leaf and formulate an
+	// asset specifier.
+	//
+	// TODO(ffranr): Revisit this. We select any asset ID to aid in metdata
+	//  retrieval, but we should be able to do this with just the group key.
+	//  However, QueryAssetGroupByGroupKey currently fails for the asset
```
Roasbeef, line 281: > QueryAssetGroupByGroupKey currently fails for the asset group

Why's this the case? Is it can issue of timing where we notify before we've fully committed on disk? 

Also seems just the group key is enough here right?

universe/supplycommit/env.go
```diff
@@ -268,6 +268,24 @@ func FetchLatestAssetMetadata(ctx context.Context, lookup AssetLookup,
...
 
+	// If the asset specifier has an asset ID, then we'll use that to
+	// fetch the asset metadata.
+	if assetSpec.HasId() {
+		assetID, err := assetSpec.UnwrapIdOrErr()
+		if err != nil {
+			return zero, err
+		}
+
+		metaReveal, err := lookup.FetchAssetMetaForAsset(ctx, assetID)
```
Roasbeef, line 279: The idea here is to attempt to fetch the _latest_ meta for a group key?

universe/supplyverifier/manager.go
```diff
@@ -386,9 +402,9 @@ func (m *Manager) startAssetSM(ctx context.Context,
 
 	fsmCfg := protofsm.StateMachineCfg[Event, *Environment]{
 		ErrorReporter: &errorReporter,
-		// TODO(ffranr): Set InitialState here.
-		Env:    env,
-		Daemon: m.cfg.DaemonAdapters,
+		InitialState:  &InitState{},
```
Roasbeef, line 405: 👍

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,381 @@
...
+			ctx, env.AssetSpec, supplyCommit.RootCommitment,
+			supplyCommit.SupplyLeaves,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to verify supply "+
+				"commitment: %w", err)
+		}
+
+		// Store the verified commitment.
+		err = env.SupplyCommitView.InsertSupplyCommit(
```
Roasbeef, line 211: A ref to come back to: this needs to spend the spent pre commits in the DB.

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,381 @@
...
+		}, nil
+
+	case *SpendEvent:
+		// A watched output has been spent, so transition to the sync
+		// state to fetch the new supply commitment. Before syncing,
+		// apply a delay to give the issuer time to publish it.
+		switch {
+		case e.WatchStartTimestamp.IsZero():
+			// No watch start timestamp: wait the full sync delay.
+			time.Sleep(env.SpendSyncDelay)
```
Roasbeef, line 241: Why do we need this sleep at all? Is it just an itest artefact?

Roasbeef: Alternatively, we can implement a delay/retry loop when we go back to the sync verify state if/when we fail to fetch the supply commit. We also have this handy function: https://github.com/lightninglabs/taproot-assets/blob/b25c0531c56cc8523f7a0dc22aea05b96ad6fcda/fn/retry.go#L37-L86

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,381 @@
...
+		events := make(protofsm.DaemonEventSet, 0, len(preCommits)+1)
+		for idx := range preCommits {
+			preCommit := preCommits[idx]
+
+			txOut := preCommit.MintingTxn.TxOut[preCommit.OutIdx]
+			mapper := func(spend *chainntnfs.SpendDetail) Event {
+				// nolint: lll
+				return &SpendEvent{
+					SpendDetail:         spend,
+					SpentPreCommitment:  &preCommit,
```
Roasbeef, line 322: Looks like we don't ever read or act on this field.

universe/supplyverifier/transitions.go
```diff
@@ -0,0 +1,381 @@
...
+					PreCommitments:      preCommits,
+					WatchStartTimestamp: watchStartTimestamp,
+				}
+			}
+
+			events = append(events, &protofsm.RegisterSpend[Event]{
```
Roasbeef, line 328: It doesn't look like we're handling the case of many of the pre commitments being spent. Or rather it can cause us to run the verification loop many times, potentially inserting the same commitment several times. Doesn't look dire, but I think we can tighten up the logic here.

itest/supply_commit_test.go
```diff
@@ -586,6 +587,60 @@ func testSupplyCommitIgnoreAsset(t *harnessTest) {
 		t.t, rpcAsset.Amount,
 		uniFetchResp.IgnoreSubtreeRoot.RootNode.RootSum,
 	)
+
+	t.Log("Attempting to fetch supply commit from secondary node")
+
+	var peerFetchResp *unirpc.FetchSupplyCommitResponse
+	require.Eventually(t.t, func() bool {
```
Roasbeef, line 594: 👍

Roasbeef: We should wrap this in a helper assertion (optional `VerifyFirst`) and apply it to the other test as well, since that creates several supply commitments.

universe/supplyverifier/manager.go
ffranr: I removed this in the end.

PR #1778, opened 2025-09-02
rfq: separate AuxChanNegotiator, AssetSalePolicy

jtobin:
This refactor avoids storing a AuxChannelNegotiator reference in the AssetSalePolicy. Instead, it's gotten from the OrderHandler and passed as a function argument to wherever it's needed.

(@GeorgeTsagk feel free to pull this into your branch if it looks worthwhile, or just close if not.)


jtobin:
(Closing this one as we stuck with the original implementation.)

Issue #1582, opened 2025-06-04
[feature]: enable proof v2 for tap channels


Roasbeef:
**Is your feature request related to a problem? Please describe.**

For backwards compat reasons, we disabled the new v2 proof scheme for tap channels. We should enable them. 

**Describe the solution you'd like**

This is related to #1573. In order to not break the _existing_ channels, we need a mechanism to signal: a funding output that includes the new commitments, and also commitment/HTLC outputs that include the new commitments. 

For the funding output, we can use a new channel type feature bit as mentioned in the channels. That covers a scenario where _new_ channels will use proof v2 all throughout. 

We can also cover _new_ channels if we use something like the proposed feature bits in the `channel_reest` message to signal that we support this new feature. In that case, the very next commitment created would utilize the new v2 proofs. Care here should be taken to ensure that force clsoes and breaches work as normal.

PR #1779, opened 2025-09-02
Prepare tap channels for v1 proofs

GeorgeTsagk:
## Description

Adds support for v1 proofs for all channel related transactions (funding/commitments/closing/sweeping).

We also guard this feature behind a feature bit, which must be set and supported by both channel peers in order to use v1 proofs.

For some cases during sweeping, we don't really care about the feature being supported by the remote party as we're sweeping locally to our own address.

Fixes https://github.com/lightninglabs/taproot-assets/issues/1582

### DIFF
--- a/tapchannel/aux_sweeper.go
+++ b/tapchannel/aux_sweeper.go
@@ -1585,6 +1585,8 @@ func (a *AuxSweeper) importCommitTx(req lnwallet.ResolutionReq,
 		}
 	}
 
+	supportSTXO := commitState.STXO.Val
+
 	// We can now add the witness for the OP_TRUE spend of the commitment
 	// output to the vPackets.
 	vPackets := maps.Values(vPktsByAssetID)
@@ -1593,8 +1595,18 @@ func (a *AuxSweeper) importCommitTx(req lnwallet.ResolutionReq,
 			"packets: %w", err)
 	}
 
+	var (
+		opts      []tapsend.OutputCommitmentOption
+		proofOpts []proof.GenOption
+	)
+
+	if !supportSTXO {
+		opts = append(opts, tapsend.WithNoSTXOProofs())
+		proofOpts = append(proofOpts, proof.WithNoSTXOProofs())
+	}
+
 	outCommitments, err := tapsend.CreateOutputCommitments(
-		vPackets, tapsend.WithNoSTXOProofs(),
+		vPackets, opts...,
 	)
 	if err != nil {
 		return fmt.Errorf("unable to create output "+
@@ -1614,8 +1626,7 @@ func (a *AuxSweeper) importCommitTx(req lnwallet.ResolutionReq,
 		for outIdx := range vPkt.Outputs {
 			proofSuffix, err := tapsend.CreateProofSuffixCustom(
 				req.CommitTx, vPkt, outCommitments, outIdx,
-				vPackets, exclusionCreator,
-				proof.WithNoSTXOProofs(),
+				vPackets, exclusionCreator, proofOpts...,
 			)
 			if err != nil {
 				return fmt.Errorf("unable to create "+
@@ -2225,9 +2236,7 @@ func (a *AuxSweeper) sweepContracts(inputs []input.Input,
 
 	// Now that we have our set of resolutions, we'll make a new commitment
 	// out of all the vPackets contained.
-	outCommitments, err := tapsend.CreateOutputCommitments(
-		directPkts, tapsend.WithNoSTXOProofs(),
-	)
+	outCommitments, err := tapsend.CreateOutputCommitments(directPkts)
 	if err != nil {
 		return lfn.Errf[returnType]("unable to create "+
 			"output commitments: %w", err)
@@ -2408,9 +2417,7 @@ func (a *AuxSweeper) registerAndBroadcastSweep(req *sweep.BumpRequest,
 	}
 
 	// Now that we have our vPkts, we'll re-create the output commitments.
-	outCommitments, err := tapsend.CreateOutputCommitments(
-		vPkts.allPkts(), tapsend.WithNoSTXOProofs(),
-	)
+	outCommitments, err := tapsend.CreateOutputCommitments(vPkts.allPkts())
 	if err != nil {
 		return fmt.Errorf("unable to create output "+
 			"commitments: %w", err)
@@ -2454,7 +2461,7 @@ func (a *AuxSweeper) registerAndBroadcastSweep(req *sweep.BumpRequest,
 
 			proofSuffix, err := tapsend.CreateProofSuffixCustom(
 				sweepTx, vPkt, outCommitments, outIdx, allVpkts,
-				exclusionCreator, proof.WithNoSTXOProofs(),
+				exclusionCreator,
 			)
 			if err != nil {
 				return fmt.Errorf("unable to create proof "+


GeorgeTsagk:
LiT tests are expected to fail, the related LiT PR is [here](https://github.com/lightninglabs/lightning-terminal/pull/1138)

tapsend/send.go
```diff
@@ -995,7 +1013,7 @@ func CreateOutputCommitments(packets []*tappsbt.VPacket,
 
 // commitPacket creates the output commitments for a virtual packet and merges
 // it with the existing commitments for the anchor outputs.
-func commitPacket(vPkt *tappsbt.VPacket, noSTXOProofs bool,
+func commitPacket(vPkt *tappsbt.VPacket, noSTXOProofs, existingSTXOLeaves bool,
```
Roasbeef: Perhaps this should take a mini options/config struct? A common bug that can arise with two side-by-side params w/ the same type is that a caller accidentally passes them in w/ the wrong order.

tapsend/send.go
```diff
@@ -1044,7 +1062,34 @@ func commitPacket(vPkt *tappsbt.VPacket, noSTXOProofs bool,
...
+			// only add them if they don't already exist. Otherwise
+			// they'd lead to an error later on during merging.
+			if existingSTXOLeaves {
+				newLeaves := append(
+					vOut.AltLeaves, stxoAssets...,
+				)
+
+				// Verify that the new set of alt leaves has
+				// unique keys.
+				err := asset.AddLeafKeysVerifyUnique(
```
Roasbeef: What about adding this protection, or assertion at a higher level? 

When can this happen (non-unique STXO leaves passed in) within the context of the protocol?

tapsend/send.go
```diff
@@ -1044,7 +1062,34 @@ func commitPacket(vPkt *tappsbt.VPacket, noSTXOProofs bool,
...
+					asset.NewLeafKeySet(), newLeaves,
+				)
+				if err == nil {
+					// If we didn't get an error, then
+					// continue with appending the leaves.
+					vOut.AltLeaves = append(
+						vOut.AltLeaves, stxoAssets...,
+					)
+				}
```
Roasbeef: We don't actually handle the error case here, so what good is this check?

tapchannel/aux_funding_controller.go
```diff
@@ -1557,6 +1586,31 @@ func (f *FundingController) processFundingMsg(ctx context.Context,
...
+		).HasFeature(tapfeatures.STXOOptional)
+
+		assetFunding.stxo = supportSTXO
+
+		// If we support STXO, then we need to collect the STXO alt
+		// leaves for the funding output. This needs to happen as it
+		// affects the construction of the tapscript root (the leaves
+		// are included in the tap commitment).
+		if supportSTXO {
+			altLeaves, err := asset.CollectSTXO(
```
Roasbeef: Alternatively, we can update `addToFundingCommitment` to accept the `stxo` flag to handle this logic directly there. As it's already responsible for adding a new funding output to the funding commitment.

tapchannel/aux_funding_controller.go
```diff
@@ -1750,13 +1811,35 @@ func (f *FundingController) processFundingReq(fundingFlows fundingFlowIndex,
...
 			fundingOut.Asset.Copy(),
 		)
 		if err != nil {
 			return fmt.Errorf("unable to add asset to funding "+
 				"commitment: %w", err)
 		}
+
+		// Merge any alt leaves to the commitment, if any.
+		commitment := fundingState.fundingAssetCommitment
+		err = commitment.MergeAltLeaves(fundingOut.AltLeaves)
```
Roasbeef: Same here re folding this into `addToFundingCommitment`.

server.go
```diff
@@ -889,7 +892,9 @@ func (s *Server) ApplyHtlcView(
 
 	// The aux leaf creator is fully stateless, and we don't need to wait
 	// for the server to be started before being able to use it.
-	return tapchannel.ApplyHtlcView(s.chainParams, in)
+	return tapchannel.ApplyHtlcView(
+		s.chainParams, in, s.cfg.AuxChanNegotiator,
```
Roasbeef, line 930: First thought at seeing this change: do we actually need to propagate this all the way here just to do something like fetch the set of aux leaves, or to apply an HTLC view? 

Even before a channel is created, we know if we're going to use the STXO feature and also noop feature (for new channels). Therefore, when preparing the funding and commitment blobs, we can just store this information right there and there. So then we don't need to query for feature bits each time we want to make a new commitment or validate one from the remote peer. 

The other dependnacy that pops up here is: these methods needs to always work, even if we're not currently connected to a peer. IIUC, this new aux negotiator interface can only query live peer state (what did they send us in `init`, etc).

tapchannel/aux_leaf_creator.go
```diff
@@ -50,10 +65,16 @@ func FetchLeavesFromView(chainParams *address.ChainParams,
 			"commit state: %w", err))
 	}
 
+	supportsSTXO := bitFetcher.GetChannelFeatures(
```
Roasbeef: In this case, we have asccess to the blobs that we have `lnd` store (funding + commitment), so we can just read that state.

Roasbeef: Same goes for all the other instances in this commit. 

IMO, we only need to query this the very first time when we create a channel.

Roasbeef: Thinking about things anew, I think the ideal implementation here would actually set a feature bit in the channel type message. Otherwise, we're falling back to the old _implicit_ channel type negotiation, which we wanted to get away from by making the type of the channel being negotiated explicit. 

With this route, an extension point would actually go near the funding manager in lnd, which is what decides if we understand a feature bit combo to open a new channel type.

tapchannel/aux_closer.go
```diff
@@ -454,11 +460,21 @@ func (a *AuxChanCloser) AuxCloseOutputs(
 			"packets: %w", err)
 	}
 
+	features := a.cfg.AuxChanNegotiator.GetChannelFeatures(
+		lnwire.NewChanIDFromOutPoint(desc.ChanPoint),
+	)
+	supportSTXO := features.HasFeature(tapfeatures.STXOOptional)
```
Roasbeef, line 466: Same here, we already know if the channel was created with STXO support or not. 

Or is the idea that if they update after the fact, we still want to use stxos in the new outputs?

tapchannelmsg/records.go
```diff
@@ -453,13 +453,17 @@ type Commitment struct {
 
 	// AuxLeaves are the auxiliary leaves that correspond to the commitment.
 	AuxLeaves tlv.RecordT[tlv.TlvType4, AuxLeaves]
+
+	// STXO is a flag indicating whether this commitment supports stxo
+	// proofs.
+	STXO tlv.RecordT[tlv.TlvType5, bool]
```
Roasbeef, line 459: Cool, so if we're using this here, then I think we can remove many of the perior instances in the earlier commits where we're doing queries on feature bits _after_ channel funding.

tapchannel/commitment.go
```diff
@@ -1278,7 +1278,7 @@ func ToCommitment(allocations []*tapsend.Allocation,
 
 	return cmsg.NewCommitment(
 		localAssets, remoteAssets, outgoingHtlcs, incomingHtlcs,
-		auxLeaves,
+		auxLeaves, false,
```
Roasbeef: Why do we unconditionally pass false here?

tapsend/send.go
GeorgeTsagk: Given that `commitPacket` is only a helper used within this package, with just one call site, I'll just wrap the two flags in a config struct to make sure the field name shows up next to the value, in order to prevent such mistakes

tapsend/send.go
GeorgeTsagk: When funding a channel we need to "prematurely" create and include the alt leaves on the commitment because LND will cache the tapscript for future reference. That's the only case where we pre-generate the alt leaves, and that's why we added the option here for this "soft" check.

Generally we want to error out on alt leaf key conflict, but for this specific case we want to let this tool know that they may already exist -- in which case we only append the new leaves if no conflict occurs.

tapsend/send.go
GeorgeTsagk: If there was a leaf key conflict we avoid touching the existing alt leaves. We only append if the set contains unique keys -- the `err == nil` case.

Could change to instead keep just the unique values, but didn't want to overengineer it.

server.go
GeorgeTsagk: > Therefore, when preparing the funding and commitment blobs, we can just store this information right there and there

Where exactly are you referring to? If you're referring to the funding/commit blobs then we can't really rely on those values for future decisions. Given that (at least for a short time window) downgrading from v1 will be allowed, then we need to know what the latest state of our peer is. Last commitment could have been using v1, but then upon reconnection our peer could stop supporting it.

> The other dependnacy that pops up here is: these methods needs to always work, even if we're not currently connected to a peer. IIUC, this new aux negotiator interface can only query live peer state (what did they send us in init, etc).

I think for `ApplyHTLCView`  we know that the channel must be active for the method to be called, which means we have communicated/agreed upon feature bits with corresponding peer. You're right for some of these cases that we could skip using the AuxChanNegotiator and rely on the commit blob, will change those hooks accordingly.

Generally, if the method affects the creation of the next commitment we'll want to consult the negotiator. If it just cares about some existing commitment, we could look up the flag in the blob.

tapchannel/aux_leaf_creator.go
GeorgeTsagk: See https://github.com/lightninglabs/taproot-assets/pull/1779#discussion_r2334275459, focusing on closing remark

tapchannel/aux_closer.go
GeorgeTsagk: At the time of running the above code lines we're crafting the coop close transaction. Both parties need to agree on whether they use v1 or not in order to craft the same transaction.

Could be the case that we used v0 for the whole lifecycle of the channel, then peer updated right before coop closing.

tapchannelmsg/records.go
GeorgeTsagk: Not really. This entry is juts a historical entry.

The fact that the last commit used STXO doesn't mean that the next commit will use STXO too. Our peer could downgrade in the meantime.

GeorgeTsagk: The above entry was only really introduced for our contract resolver
https://github.com/lightninglabs/taproot-assets/blob/87174c6615cd949a231a30f8543285593dbe1a59/tapchannel/aux_sweeper.go#L2498

specifically the part where we import the commit tx:
https://github.com/lightninglabs/taproot-assets/blob/87174c6615cd949a231a30f8543285593dbe1a59/tapchannel/aux_sweeper.go#L1588

tapchannel/commitment.go
GeorgeTsagk: this was only temporary for the commit to compile, right after we switch to using the provided flag value

tapsend/send.go
GeorgeTsagk: The rationale behind adding the option is to be able to re-use the `commitPacket` helper even from the chan funding call site, where the above special conditions are met.

tapsend/send.go
```diff
@@ -925,6 +925,11 @@ type OutputCommitmentConfig struct {
 	// STXO proofs. This should only be done for asset channels to preserve
 	// the backward compatibility with older peers.
 	noSTXOProofs bool
+
+	// existingSTXOLeaves indicates whether we should be aware that alt
+	// leaves that correspond to stxo assets may already exist in asset
+	// outputs of the vpsbt.
+	existingSTXOLeaves bool
```
gijswijs: I understand the reason of this boolean and premature tapscript generation that is referenced by `lnd`, but I don't like it.
This complexity makes the code harder to reason about and could be fragile. Isn't there a way to establishes a single point of responsibility for adding STXO?

tapchannel/aux_leaf_creator.go
```diff
@@ -80,8 +103,8 @@ func FetchLeavesFromView(chainParams *address.ChainParams,
 // to the passed aux blob, and an existing channel commitment.
 func FetchLeavesFromCommit(chainParams *address.ChainParams,
 	chanState lnwl.AuxChanState, com channeldb.ChannelCommitment,
-	keys lnwl.CommitmentKeyRing,
-	whoseCommit lntypes.ChannelParty) lfn.Result[lnwl.CommitDiffAuxResult] {
+	keys lnwl.CommitmentKeyRing, whoseCommit lntypes.ChannelParty,
+	bitFetcher FeatureBitFetcher) lfn.Result[lnwl.CommitDiffAuxResult] {
```
gijswijs: That `bitFetcher` is unused now, isn't it?

tapchannel/aux_leaf_creator.go
```diff
@@ -98,6 +121,8 @@ func FetchLeavesFromCommit(chainParams *address.ChainParams,
 			"commitment: %w", err))
 	}
 
+	supportSTXO := commitment.STXO.Val
```
gijswijs, line 120: Cool, you're now getting the flag directly from the commitment itself.

tapchannel/aux_leaf_creator.go
```diff
@@ -24,10 +27,22 @@ const (
 	DefaultTimeout = 30 * time.Second
 )
 
+// FeatureBitFetcher is responsible for fetching feature bits either by
+// referencing a remote peer, or an established channel.
+type FeatureBitFetcher interface {
+	// GetPeerFeatures returns the negotiated features with the given peer.
+	GetPeerFeatures(peer route.Vertex) lnwire.FeatureVector
```
gijswijs: This method is now unused in this interface. We could do without it, right?

tapchannel/commitment.go
```diff
@@ -1268,7 +1278,7 @@ func ToCommitment(allocations []*tapsend.Allocation,
 
 	return cmsg.NewCommitment(
 		localAssets, remoteAssets, outgoingHtlcs, incomingHtlcs,
-		auxLeaves,
+		auxLeaves, stxo,
```
gijswijs, line 1281: Shouldn't this already be a boolean variable in the previous commit?

tapsend/send.go
GeorgeTsagk: Thanks for this comment.

I was getting ready to write a long explanation on how this isn't a quirk of the commitment related toolkit but a characteristic of the chan funding flow -- i.e we need to craft the correct commitment even though we haven't yet finalized the vpsbt that creates it.

Then I realized that in my latest push, where I simplified a lot of code, I also omitted the (now unnecessary) part of actually appending the alt leaves to the vOut. This means that even though we do collect STXOs prematurely, we don't really attach it to the vOut, leaving it "clean" for the toolkit to handle it. This means that we actually never hit a case where there's duplicate alt leaves, as we'd never really persist them between preparation & finalization phase.

Will just drop the commit that adds `WithExistingSTXOLeaves` altogether.

tapchannel/aux_leaf_creator.go
GeorgeTsagk: Yeah, since this endpoint refers to an existing commitment we need to look-up the value that was persisted when that commitment was created.

GeorgeTsagk:
> Did a first pass and mad comments throughout. Starts to look good! This PR depends on the litd itest working, right? ([lightninglabs/lightning-terminal#1138](https://github.com/lightninglabs/lightning-terminal/pull/1138))

Yeap. Will update the LiT PR soon to reflect all of recent changes made here

tapsend/send.go
gijswijs: Interesting. But if we don't append them to the `vOut` why do we still collect the STXOs prematurely? Where do we use them in the chan funding flow?

GeorgeTsagk: We directly merge them into the commitment that will be used to derive the tapscript for LND. The vOut alt leaves are just the placeholder for keeping the leaves somewhere, until we'd eventually want to merge them into the tap commitment.

So now we simply skip the step of adding the alt leaves to the vOut and later when the helper functions get the vPSBT they'll be free to re-collect/append the alt leaves without conflicts or extra ticks needed

tapchannel/aux_funding_controller.go
```diff
@@ -1086,14 +1118,23 @@ func (f *FundingController) signAllVPackets(ctx context.Context,
 // complete, but unsigned PSBT packet that can be used to create out asset
 // channel.
 func (f *FundingController) anchorVPackets(fundedPkt *tapsend.FundedPsbt,
-	allPackets []*tappsbt.VPacket) ([]*proof.Proof, error) {
+	allPackets []*tappsbt.VPacket, stxo bool) ([]*proof.Proof, error) {
 
 	log.Infof("Anchoring funding vPackets to funding PSBT")
 
+	// We have collected the stxo leaves for the outputs in a previous step.
```
gijswijs: This comment doesn't make any sense anymore now that we avoided alt leaves collisions anyway.

tapchannel/aux_funding_controller.go
```diff
@@ -1713,6 +1766,15 @@ func (f *FundingController) processFundingReq(fundingFlows fundingFlowIndex,
 		}
 	}
```
gijswijs, line 1764: This block should be pushed a few lines down, such that the registration of the defer is still directly below the creation of the `unlockLeases` closure.

tapchannel/aux_leaf_creator.go
```diff
@@ -24,10 +26,19 @@ const (
 	DefaultTimeout = 30 * time.Second
 )
 
+// FeatureBitFetcher is responsible for fetching feature bits either by
+// referencing a remote peer, or an established channel.
```
gijswijs: This comment isn't up to date anymore, now that it's impossible to reference a peer. It only works byu referencing a channel.

tapchannel/aux_sweeper.go
```diff
@@ -2225,9 +2236,7 @@ func (a *AuxSweeper) sweepContracts(inputs []input.Input,
 
 	// Now that we have our set of resolutions, we'll make a new commitment
 	// out of all the vPackets contained.
-	outCommitments, err := tapsend.CreateOutputCommitments(
-		directPkts, tapsend.WithNoSTXOProofs(),
-	)
+	outCommitments, err := tapsend.CreateOutputCommitments(directPkts)
```
gijswijs, line 2239: For my understanding: From now on we always create a sweeping Tx with stxo proofs.
Why didn't we do so before? Oli needed the `WithNoSTXOProofs` to make the channel commitments work, but it seems to me that the sweeping Tx could always have had stxo proofs. (This is more a question about Oli's PR than about this PR, but I'm still wondering.)

gijswijs (changes requested):
Nearly there! I made some small remarks and some questions.

tapsend/send.go
gijswijs: Had some back and forth on this with @GeorgeTsagk and an LLM who will remain anonymous. I'm posting it here for future reference.

1. **Premature Stage: Funding Transaction Construction (`addToFundingCommitment`)**
   This stage computes the correct `TapscriptRoot` for the on-chain P2TR funding output script. LND needs this root early, before the final proofs are generated.  
   **Call Flow:**
     `FundingController.processFundingReq` -> `pendingAssetFunding.addToFundingCommitment`  
   **Action:** `addToFundingCommitment` calls `asset.CollectSTXO` and merges the resulting leaves directly into an in-memory `commitment.TapCommitment`. This allows for the correct `TapscriptRoot` to be provided to LND. This process does not modify any `VPacket`'s `AltLeaves`.
2. **Finalization Stage: Final Proof Generation (`commitPacket`)**  
   This stage generates the final Taproot Asset proofs for the assets being created in the channel *after* the funding transaction is fully known.  
   **Call Flow:**
     `FundingController.processFundingReq` -> `FundingController.completeChannelFunding` -> `FundingController.anchorVPackets` -> `tapsend.CreateOutputCommitments` -> `commitPacket`  
   **Action:** The `commitPacket` function is called on the `VPacket`s representing the channel assets. It calls `asset.CollectSTXO` again and appends the resulting leaves to the `vOut.AltLeaves` slice before creating the final commitment for the proof.

tapchannel/aux_sweeper.go
GeorgeTsagk: Yeah I suppose we defaulted to using `WithNoSTXOProofs` everywhere in the tapchannel code, regardless of if it could have been using them in the first place. I guess it would require explaining why it would work in some places but not in others -- which is what we're doing now in this PR.

server.go
Roasbeef: > Where exactly are you referring to? If you're referring to the funding/commit blobs then we can't really rely on those values for future decisions. Given that (at least for a short time window) downgrading from v1 will be allowed, then we need to know what the latest state of our peer is.

Why not? We generate that blob ourselves. If say someone changes that information to say that there's only 5 beefbux in a channel but we had 10, then we'd act on that information. You're correct that atm it isn't directly cryptographically authenticated, but we assume that the information we wrote is correct.

We can protect against downgrading if we store the current negotiated feature bit information in the funding blob. If we come online, and the peer tells us we're not using the new channel type, then we can fail to proceed.

Roasbeef: > I think for ApplyHTLCView we know that the channel must be active for the method to be called, which means we have communicated/agreed upon feature bits with corresponding peer.

Yep, you're correct about that. We only call that when accepting or creating a new commitment.

tapchannelmsg/records.go
Roasbeef: > The fact that the last commit used STXO doesn't mean that the next commit will use STXO too. Our peer could downgrade in the meantime.


We _must_ continue to use it IMO. We prevent downgrading simply by refusing to continue the channel if they downgrade and want to go back to using the old version.

Roasbeef (commented):
Diff looks pretty good now. Only thing I think we need to do is be explicit about our goals here re downgrade protection or not. 

Possible to follow up on this, as this is just a prep PR.

GeorgeTsagk:
> Diff looks pretty good now. Only thing I think we need to do is be explicit about our goals here re downgrade protection or not.
> 
> Possible to follow up on this, as this is just a prep PR.

Yeap I can write a short summary of how we actually achieve downgrade protection:

The v1 proofs will become active over a channel if both peers support the feature, the negotiation for that happens during connection. We have two types of data keeping track of v1 proof presence:
a) Feature bit maps (live negotiated values -- responsibility of `AuxChanNegotiator`: "does my peer support v1 **now**")
b) Commit blob flag (when creating a commitment we store whether that commitment is using v1 directly on the commit blob: "did commitment **X** use v1?")

In our various call sites, we'll consult **a)** if the on-going action produces a new commitment, or a new overall state (like the chan funding tx).

When referring to **existing, already created** commitments we'll consult their respective blob flag to find out whether they used v1 proofs or not.

#### Restart & Downgrade

If a peer of the channel restarts and downgrades, dropping support for v1, then the reconnection negotiation will result to v1 not being active over the channel. This means that any future transactions will not use v1 proofs (commitments / co-op closing -- value acquired from `AuxChanNegotiator`, which embraces older versions too).

We'll have a grace-period for which the v1 feature bit will be marked as optional, but eventually we'll want to switch to  it being a required feature. The time window until the feature switches to a required one is the only one within which a peer may downgrade and still keep the channel active. When the switch happens the channel won't become active again until v1 is negotiated on connection.

#### Sweeping & Breaches 

Within the lifecycle of a channel it is totally expected to only have some of the commitments using v1. Given that we have this notion of a "graceful window" we also accept that there isn't a single switch from non-v1 to v1 usage over the channel. This was the main motivation behind storing the v1 flag in the commit blob.

When a commitment makes it on-chain, our sweeper will consult the commit blob to see whether v1 is used, then reconstruct the tap commitment accordingly, and sweep the funds.

The sweep transaction itself can be v1 regardless of what the remote party supports, as that's a transaction paying directly to the local party.

gijswijs:
> When the switch happens the channel won't become active again until v1 is negotiated on connection.

But that isn't part of this PR, right? We should probably make a note of this on the tracking issue: https://github.com/lightninglabs/taproot-assets-private/issues/38

GeorgeTsagk:
> > When the switch happens the channel won't become active again until v1 is negotiated on connection.
> 
> But that isn't part of this PR, right? We should probably make a note of this on the tracking issue: [lightninglabs/taproot-assets-private#38](https://github.com/lightninglabs/taproot-assets-private/issues/38)

Correct. Will update the tracking issue.

GeorgeTsagk:
Rebased on latest `main` / `feature-bits` base PR, go is now on v1.24.6 so CI should be more green (LiT still expected to fail -- see [LiT PR](https://github.com/lightninglabs/lightning-terminal/pull/1138) for updated itests)

litbot-9000:
@gijswijs: review reminder
@roasbeef: review reminder

tapchannel/aux_sweeper.go
```diff
@@ -1612,10 +1619,14 @@ func (a *AuxSweeper) importCommitTx(req lnwallet.ResolutionReq,
 	for idx := range vPackets {
 		vPkt := vPackets[idx]
 		for outIdx := range vPkt.Outputs {
+			var opts []proof.GenOption
```
Roasbeef: Don't need to declare `opts` each time, we can just declare it at the very top (using a diff avriable name for proof vs tapsend).

tapchannel/commitment.go
```diff
@@ -622,11 +627,16 @@ func GenerateCommitmentAllocations(prevState *cmsg.Commitment,
 	for idx := range vPackets {
 		vPkt := vPackets[idx]
 		for outIdx := range vPkt.Outputs {
+			var opts []proof.GenOption
```
Roasbeef: Same here re not needing to declare it each time.

PR #1780, opened 2025-09-03
server: enable ExtraBudgetForInputs to be called before full server start up

Roasbeef:
This is a pure function that just returns what's basically a static value. There's no risk of deadlock if a caller calls into this function before the server has fully started up.


Roasbeef:
@ffranr great suggestion, will apply that

Roasbeef:
PTAL.

Issue #1782, opened 2025-09-05
[feature]: store and verify peer-issued pre-commitment outputs when syncing asset issuance proofs


ffranr:
A supply pre-commitment output is an output on the mint anchor transaction that is later spent by a supply commitment transaction. This creates the spend chain linking mint anchor transactions to supply commitments.

For locally issued assets, the node already stores pre-commitment outputs during minting. These outputs are then spent when the node publishes new supply commitment updates.

We also need to handle assets issued by peer nodes. Specifically:

* When syncing issuance proofs from a peer, store the corresponding pre-commitment outputs.
* During verification of supply commitments published by that peer, assert that any unspent pre-commitment output is consumed by a new supply commitment transaction.

Implementation details:

* Extend the database schema with a new table for peer-issued pre-commitment outputs (separate from the table used for locally minted assets).
* Update `supplyverifier.Verifier` to assert that at least one pre-commitment output is spent if the supply commitment does not consume a previous supply commitment output. See: https://reviewable.io/reviews/lightninglabs/taproot-assets/1674#-OZMmYmCEzQahmW1_EO_

PR #1784, opened 2025-09-09
supplyverifier: store and verify non-issuer supply pre-commitments

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/1782

---

This PR strengthens the supply commitment verification process by enforcing that initial supply commitments must spend pre-commitment outputs. This requirement ties the first supply commitment transaction directly to a mint anchor transaction.

Key changes:

* **Verification logic**

  * In `supplyverifier`, added an assertion that initial supply commitments must spend pre-commitment outputs. This ties the first supply commitment transaction directly to a mint anchor transaction, ensuring the chain of authentication is enforced for both local and remote issuances.

* **Database queries and schema**

  * Renamed existing queries to clearly distinguish between pre-commitments created by the local issuer (`FetchUnspentMintSupplyPreCommits`) and those created by remote issuers (`FetchUnspentSupplyPreCommits`).
  * Added conditional upsert logic so that supply pre-commitments can be recorded when syncing issuance proofs for assets not issued locally.

* **tapdb enhancements**

  * Extended `SupplyCommitMachine.UnspentPrecommits` to return both local and remote issuer pre-commitments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1784)


tapdb/asset_minting_test.go
```diff
@@ -1855,12 +1856,16 @@ func storeMintSupplyPreCommit(t *testing.T, assetStore AssetMintingStore,
...
+		opBytes, err := encodeOutpoint(outpoint)
+		require.NoError(t, err)
+
 		_, err = q.UpsertMintSupplyPreCommit(
 			ctx, UpsertBatchPreCommitParams{
 				BatchKey:             batchKey,
 				TxOutputIndex:        txOutputIndex,
 				TaprootInternalKeyID: internalKeyID,
 				GroupKey:             groupKey,
+				Outpoint:             opBytes,
```
Roasbeef, line 1868: Ah nice, I had missed this when I added the new column.

tapdb/sqlc/migrations/000046_supply_pre_commit_tables.up.sql
```diff
@@ -0,0 +1,3 @@
+-- Rename the table to align with "supply" terminology.
+ALTER TABLE mint_anchor_uni_commitments
```
Roasbeef, line 2: Nice, yeah if we're not adding any columns, we can just rename the table.

tapdb/sqlc/migrations/000046_supply_pre_commit_tables.up.sql
```diff
@@ -1,3 +1,28 @@
 -- Rename the table to align with "supply" terminology.
 ALTER TABLE mint_anchor_uni_commitments
     RENAME TO mint_supply_pre_commits;
+
+-- Add a new table to track supply pre-commitments generated by
+CREATE TABLE supply_pre_commits (
```
Roasbeef, line 6: Ah ok, so going in this direction instead of permitting a `NULL` value for `batch_id`?

tapdb/sqlc/migrations/000046_supply_pre_commit_tables.up.sql
```diff
@@ -1,3 +1,28 @@
...
+    id INTEGER PRIMARY KEY,
+
+    -- The asset group key for the asset group this supply pre-commitment
+    -- belongs to.
+    group_key BLOB NOT NULL CHECK(length(group_key) = 33),
```
Roasbeef: Group key index as well?

universe/supplycommit/env.go
```diff
@@ -69,6 +70,22 @@ func (s SupplySubTree) String() string {
 	}
 }
 
+// ToUniverseProofType converts the supply subtree type to the corresponding
+// universe proof type.
+func (s SupplySubTree) ToUniverseProofType() (universe.ProofType, error) {
+	switch s {
+	case MintTreeType:
+		return universe.ProofTypeMintSupply, nil
```
Roasbeef, line 78: In order to unify things further, perhaps we update the `String` method of `SupplySubTree` to just return a call to the `String()` method of the corresponding `ProofType`. 

This way these strings aren't defined in two palces as they are now.

tapdb/universe.go
```diff
@@ -577,6 +578,137 @@ func upsertAssetGen(ctx context.Context, db UpsertAssetStore,
...
+
+	// If the meta reveal doesn't indicate that the asset group supports
+	// supply commitments, then we can return early.
+	if !metaReveal.UniverseCommitments {
+		return nil
+	}
+
+	// At this point, we know that the asset group supports supply
+	// commitments, so we must have a delegation key to proceed.
+	if metaReveal.DelegationKey.IsNone() {
```
Roasbeef, line 607: Could lift these 5 conditionals into a `shouldInsertPreCommit` helper function to reduce the size of this greater function a bit.

tapdb/sqlc/queries/supply_commit.sql
```diff
@@ -194,6 +194,23 @@ WHERE transition_id = @transition_id;
 DELETE FROM supply_update_events
 WHERE transition_id = @transition_id;
 
+-- name: FetchUnspentSupplyPreCommits :many
+-- Fetch unspent supply pre-commitment outputs. Each pre-commitment output
+-- comes from a mint anchor transaction and relates to an asset issuance
+-- where a peer node acted as the issuer. Rows in this table do not relate to an
+-- issuance where the local node acted as the issuer.
+SELECT
```
Roasbeef, line 202: This query serves to demonstrate the advantage of this stand alone table. We can directly obtain items like the internal key instead of needing several joins to extract it.

tapdb/supply_commit.go
```diff
@@ -326,6 +342,76 @@ func (s *SupplyCommitMachine) UnspentPrecommits(ctx context.Context,
 			preCommits = append(preCommits, preCommit)
 		}
 
+		// If any pre-commits were found where we acted as the issuer,
```
Roasbeef, line 346: Perhaps this should just be another method? So one used for remote issuers, and one used for local issuers.

tapdb/supply_commit_test.go
```diff
@@ -1760,7 +1764,9 @@ func TestSupplyCommitApplyStateTransition(t *testing.T) {
 	)
 
 	// Verify we have the extra one from before plus the new one.
-	precommitsRes = h.commitMachine.UnspentPrecommits(h.ctx, h.assetSpec)
+	precommitsRes = h.commitMachine.UnspentPrecommits(
+		h.ctx, h.assetSpec, false,
```
Roasbeef: Why shouldn't this be `true` here? This test is meant to exercise the workflow for updating the supply commitment for locally issued assets.

ffranr: Boolean flag arg set to `true` is for local issued assets **only**. So setting to `false` gets local issued and remote issued. But `UnspentPrecommits` returns early if it finds any locally issued anyway.

tapdb/sqlc/migrations/000046_supply_pre_commit_tables.up.sql
ffranr: I think a separate table might be simpler. I tried re-using the same table and it got messy, but that strat may still be possible.

universe/supplycommit/env.go
ffranr: Ah nice, good idea.

ffranr: hmm `ToUniverseProofType()` returns an error where as `SupplySubTree.String()` does not. That's causing some problems. Would you mind if I put up a separate PR to refactor `SupplySubTree` perhaps?

tapdb/supply_commit.go
ffranr: I’ve been considering that option. My reasoning for keeping a single `UnspentPrecommits` is twofold:

* There shouldn’t be any overlap between remote- and locally-issued assets, so `UnspentPrecommits` can be called safely in either context.
* Splitting into local/remote variants adds naming complexity. Terms like “remote pre-commit” aren’t self-explanatory, and including “issued” in the names makes them cumbersome.

What do you rekon i should do here?

jtobin (approved):
This all looks very solid, and I haven't found any obvious issues. LGTM. :+1:

ffranr:
Added a fix commit to this PR, originally pointed out by @jtobin here: https://github.com/lightninglabs/taproot-assets/pull/1788

tapdb/supply_commit_test.go
Roasbeef: But in the context of this test, aren't these assets indeed locally issued? Or is it that the test doesn't insert the necessary records to have the DB logic detect them as locally issued?

ffranr: Ah I see what you mean now. I've updated these instances to `true` in the latest commits. Which is clearer in the context of this test :+1:

ffranr:
I’ve updated the `group_key` field in the new pre-commit SQL table to:

```
    -- The asset group key for this supply pre-commitment.
    -- Stored in canonical 32-byte x-only form as defined in BIP340
    -- (schnorr.SerializePubKey).
    group_key BLOB NOT NULL CHECK(length(group_key) = 32),
```

PR #1785, opened 2025-09-09
rpc: update new supply commit calls to use the global universe server rate limiter

Roasbeef:
This ensures that these queries are also rate limited, as they'll be exposed to the outside world depending on config setting.


jtobin (approved):
LGTM :+1:

PR #1789, opened 2025-09-10
allow building taprpc by itself

ZZiigguurraatt:
When I try to run `go build` in the `taprpc` subdirectory and build just `taprpc` without `tapd`, I get the following errors.

```
$ go build
# github.com/lightninglabs/taproot-assets/taprpc
./utils.go:19:3: unknown field UseHexForBytes in struct literal of type protojson.MarshalOptions
./utils.go:27:3: unknown field UseHexForBytes in struct literal of type protojson.UnmarshalOptions
./utils.go:37:3: unknown field UseHexForBytes in struct literal of type protojson.MarshalOptions
./utils.go:45:3: unknown field UseHexForBytes in struct literal of type protojson.UnmarshalOptions
```

I then take a look at `go.mod`

and it has

```
// We want to format raw bytes as hex instead of base64. The forked version
// allows us to specify that as an option.
replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-display v1.34.2-hex-display
```

but `taprpc/go.mod`

does not.

If I run `go build` in the root of this repo, it works with no errors. I think CI is working because we only try to build `taprpc` with `tapd`? I think the replace directive affects sub modules (see also https://go.dev/ref/mod#go-mod-file-replace) so that is why `taprpc` can be built as part of `tapd`, but not by itself.

This PR adds the replace directive to `taprpc/go.mod` too so that I can build `taprpc` independently.

I'm not sure if we should also add some more CI checks that would have caught this issue?


ZZiigguurraatt:
> I think this is good to include here -- taprpc is a standalone module, so one should ideally be able to build it in isolation out-of-the-box, for whatever purpose.
> 
> The LiT test fail (a deadline exceeded in multi-rfq) looks to be a flake, so this LGTM. 👍

Thanks. I had some other conversations with @Roasbeef  about this, so would appreciate him as the second reviewer, but I don't have permissions to add him to the github issue reviewer list.

Roasbeef (approved):
LGTM 🌙

PR #1790, opened 2025-09-11
tools: migrate linter config to v2

jtobin:
The new config file has been generated via the golangci-lint 'migrate' command, with manual comment patch-up afterwards.


jtobin:
Lint test is failing on CI. I probably need to update the golangci-lint version in the Docker image; will take a look.

bhandras:
> Lint test is failing on CI. I probably need to update the golangci-lint version in the Docker image; will take a look.

Just need to bump the timeout in the yaml.

jtobin:
> Just need to bump the timeout in the yaml.

I *think* the run.timeout option has been deprecated in the v2 config, which I *think* is actually only supported by golangci-lint v2.0.0 or greater? (I'm not terribly familiar with this tool, though.)

My local golangci-lint install is v2.4.0, which is quite a ways ahead of what's used in CI.

jtobin:
Ok I've patched this and made a quick check locally, but will let CI decide if it's working..

jtobin:
Ok, this now seems to be working ok. Review away!

darioAnongba (approved):

tools/go.mod
```diff
@@ -1,75 +1,98 @@
 module github.com/lightninglabs/taproot-assets/tools
 
-go 1.23.12
+go 1.24.0
```
ziggie1984: is the golang version here required to run golint v2 ? why don't you go straight to 1.24.5 for the golang version but just for the toolchain, I mean its just the tools package but nonetheless curious why you decided to make this distinction ?

tools/go.mod
```diff
@@ -1,75 +1,98 @@
...
+
+toolchain go1.24.5
 
 require (
 	github.com/btcsuite/btcd v0.24.2
 	github.com/dvyukov/go-fuzz v0.0.0-20210602112143-b1f3d6f4ef4e
-	github.com/golangci/golangci-lint v1.64.6
+	github.com/golangci/golangci-lint/v2 v2.4.0
```
ziggie1984: lets use the master version see comment here: https://github.com/lightningnetwork/lnd/pull/10209#discussion_r2345072766

tools/go.mod
jtobin: IIRC these were just the versions automatically produced after using 'go get' and 'go mod tidy' -- I've simply changed the golang version to 1.24.5 in 2494834a2200fd178ae1c705a313c4399e8a7589 and dropped the toolchain distinction.

tools/go.mod
jtobin: Fixed in 2494834a2200fd178ae1c705a313c4399e8a7589.

ziggie1984:
consider also using the new go tool feature introduce in golang 1.24, it prevents version drift. You can take a look at the lnd or btcwallet PRs => in a follow up PR

Issue #514, opened 2023-09-20
wallet: garbage collect tombstone UTXOs


guggero:
To support non-interactive (address) full value sends, tombstone outputs are sometimes created.
They do contain an asset commitment to prove the previous asset split, but have a value of 0 on the asset level.
Currently those tombstone outputs are put into a BTC anchor output of 1k sats but then aren't ever used by the wallet anymore.

We should garbage collect those tombstone outputs when creating new on-chain transactions (e.g. when sending other assets).
The following list should roughly describe the steps necessary:

- [ ] Create a method to query for managed anchor UTXOs that have 0 effective value (counting burns as 0 value) on the asset level
- [ ] When creating a new on-chain transaction, add all currently available tombstone anchor UTXOs as inputs

Because we don't currently clean up those empty outputs, we explicitly don't allow burning all assets within an anchor output. But with this cleanup mechanism in place, we might decide to allow that now.

dstadulis:
>Create a method to query for managed anchor UTXOs that have 0 effective value (counting burns as 0 value) on the asset level
When creating a new on-chain transaction, add all currently available tombstone anchor UTXOs as inputs

👍 clear implementation

PR #1791, opened 2025-09-11
Full asset burn 

### LINKED ISSUE #918: [bug]: Asset Burning for Collectibles
cant burn assets which are json or oneof somehow via api, I think only token assets can be burned at the moment.
If I'm wrong please let me know how I would burn a Asset which is a Collectible.

### LINKED ISSUE #1435: [feature]: Burn All Assets of Anchor Output
When attempting to burn all assets, we get the response ```[tapcli] unable to send assets: rpc error: code = Unknown desc = error funding burn: burning all assets of an anchor output is not supported```

Users should be able to burn all assets of an anchor output at once. There was a stop gap in place initially which is why this isn't currently supported.

darioAnongba:
### Enable burn of full asset of anchor output (Fixes #1435, Fixes #918)

The goal of the PR is to enable "full burns" where all the asset units anchored to a UTXO are destroyed.

- Remove stop gap when the `amount_to_burn` equals the full UTXO amount and no other assets are present.
- The burn goes to a unique provably unspendable key.
- Ensures the burn proof is generated correctly.

This PR doesn't address issue #514, zero-value outputs accumulate and are not garbage collected.

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -220,6 +220,9 @@
   `wallet.psbt-max-fee-ratio` and is introduced by
   [PR #1545](https://github.com/lightninglabs/taproot-assets/pull/1545).
 
+- Enable [burning the full amount of an asset](https://github.com/lightninglabs/taproot-assets/pull/1791)
+  when it is the sole one anchored to a Bitcoin UTXO.
+
 ## RPC Updates
 
 ## tapcli Updates


tapfreighter/wallet.go
```diff
@@ -837,7 +829,31 @@ func (f *AssetWallet) FundBurn(ctx context.Context,
 		}
 
 		if !otherAssets {
-			return nil, ErrFullBurnNotSupported
+			// Full burn case: Create a tombstone output as change.
+			firstOut := fundedPkt.VPackets[0].Outputs[0]
+			tombstoneOut := &tappsbt.VOutput{
```
GeorgeTsagk: not sure why we need this, but also worth mentioning that this VOutput never makes it to the final vPkt (probably discarded because of the `0` amount)

tapfreighter/wallet.go
```diff
@@ -815,20 +810,17 @@ func (f *AssetWallet) FundBurn(ctx context.Context,
 			len(fundedPkt.VPackets))
 	}
 
-	// We want to avoid a BTC output being created that just sits there
-	// without an actual commitment in it. So if we are not getting any
-	// change or passive assets in this output, we'll not want to go through
-	// with it.
-	firstOut := fundedPkt.VPackets[0].Outputs[0]
-	if len(fundedPkt.VPackets[0].Outputs) == 1 &&
```
GeorgeTsagk, line 823: If this PR just aims at removing the stop gap, then we can just remove all lines below this point (the full-burn cases) and continue with the fundedVpkt normally.

tapfreighter/wallet.go
```diff
@@ -837,7 +829,31 @@ func (f *AssetWallet) FundBurn(ctx context.Context,
...
+				ScriptKey:         asset.NUMSScriptKey,
+			}
+
+			// Add tombstone as first output.
+			vPkt.Outputs = slices.Insert(
+				vPkt.Outputs, 0, tombstoneOut,
+			)
+
+			// Re-create the packet with the tombstone output.
+			fundedPkt, err = createFundedPacketWithInputs(
```
GeorgeTsagk: so basically this `fundedPkt` doesn't contain the 0-amount tombstone created above

itest/burn_test.go
```diff
@@ -524,3 +507,83 @@ func testBurnGroupedAssets(t *harnessTest) {
...
+		[][]byte{simpleAssetGen.AssetId},
+		[]uint64{fullBurnAmt, 0}, 1, 2, 2, true,
+	)
+
+	// Balance reflects the full burn amount.
+	AssertBalanceByID(
+		t.t, t.tapd, simpleAssetGen.AssetId,
+		simpleAsset.Amount-fullBurnAmt,
+	)
+}
```
Roasbeef, line 627: We should also add an extra assertion here where we export the proof, and then attempt to verify it either with the same node or a 3rd party.

tapfreighter/wallet.go
darioAnongba: yes, this seems to be the way, no tombstone needed! Changes applied.

itest/burn_test.go
darioAnongba: I added both the collectible test case and the proof verification. Yet, I think that the proof verification is not needed here because it is already done during the test execution. As you can see, `testBurnAssets` and `testBurnGroupedAssets` do not have such an assertion (and I doubt Oli forgot about them). We can see this from the logs:

```
2025-09-17 18:58:03.469 [INF] FRTR chain_porter.go:626: Importing proof for output 0 into local Proof Archive
2025-09-17 18:58:03.475 [INF] FRTR chain_porter.go:370: ChainPorter executing state: SendStateTransferProofs
2025-09-17 18:58:03.476 [INF] FRTR chain_porter.go:370: ChainPorter executing state: SendStateComplete
2025-09-17 18:58:03.476 [INF] FRTR chain_porter.go:1860: Parcel transfer is fully complete (anchor_txid=04930d5f112357938eb35bafcccf0649aa3e3c0b6b30d7e3c1418d00d2b290fd)
2025-09-17 18:58:03.476 [INF] FRTR chain_porter.go:404: ChainPorter completed state machine for parcel (anchor_txid=04930d5f112357938eb35bafcccf0649aa3e3c0b6b30d7e3c1418d00d2b290fd)
2025-09-17 18:58:03.476 [INF] GRDN custodian.go:1323: Received new proof file for asset_id=f7c3767db5953ef7a2038974fb39bc8b44ef04f0ed22825fc012961705095353, script_key=026c9df9800c152e009e32a5abdcaad8e70095c1da207da27e4686b86e09ab9a8b,version=0, num_proofs=2
```
Let me know if I should keep the proof verification in the test.

tapfreighter/wallet.go
darioAnongba: tombstone not needed, resolving.

itest/burn_test.go
```diff
@@ -344,7 +328,7 @@ func testBurnAssets(t *harnessTest) {
 
 	require.Equal(t.t, groupBurn.Note, "")
 
-	// Test case 6: Burn the single unit of a grouped collectible. We start
+	// Test case 5: Burn the single unit of a grouped collectible. We start
```
GeorgeTsagk, line 331: We could also just include the full burn as a test case here by extending `testBurnAssets`. Could also keep `testFullBurnUTXO` regardless, since we're also burning a collectible there.

itest/burn_test.go
```diff
@@ -524,3 +508,115 @@ func testBurnGroupedAssets(t *harnessTest) {
 	require.Equal(t.t, burnNote, burn.Note)
 	require.Equal(t.t, assetGroupKey, burn.TweakedGroupKey)
 }
+
+// testFullBurnUTXO tests that we can burn the full amount of an asset UTXO
```
GeorgeTsagk:
```suggestion
// testFullBurnUTXO tests that we can burn the full amount of an asset UTXO.

File: itest/burn_test.go [hunk 7]
```diff
@@ -524,3 +508,115 @@ func testBurnGroupedAssets(t *harnessTest) {
...
+		t.t, t.tapd,
+		burnResp.BurnProof.Asset.AssetGenesis.AssetId,
+		burnResp.BurnProof.Asset.ScriptKey,
+		outpoint,
+	)
+	verifyResp, err = t.tapd.VerifyProof(ctxt, &taprpc.ProofFile{
+		RawProofFile: proofResp.RawProofFile,
+	})
+	require.NoError(t.t, err)
+	require.True(t.t, verifyResp.Valid)
```
GeorgeTsagk, line 621: let's also use `t.tapd.ListBurns` as an extra check on this test (just verify that it was recorded correctly)

itest/burn_test.go
darioAnongba: Are you sure that's a good idea? The test is already very hard to understand as it is. It mints different asset types, sends only some to different outputs and then proceeds to selectively burn the assets that live either in the mint tx or the one with multiple outputs. It also does another send in the middle of the test and has test cases dependent on each other.
It is not clear what amount we should burn for it to be a full burn that will select an asset that lives alone in an anchor UTXO without passive assets and doing so requires understanding the entire flow of the test.
I'd actually split this test into small ones that test a clear use case so we know exactly what the problem is when the test fails.

itest/burn_test.go
darioAnongba: I added `AssertNumBurns`, that uses ListBurns internally and returns the burns as well, like in other tests.

itest/burn_test.go
GeorgeTsagk: So I went ahead and listed the assets at the end of the test, here we can see them grouped by anchor outpoint

```
    asset1 300: f3141225a6669242b68a124b533a2f1bced377901e8b033788766a981d523453:1
    asset2 4900: f3141225a6669242b68a124b533a2f1bced377901e8b033788766a981d523453:1

    asset1 1100: a9eb56da2f3422ea615f6cc68f85138ea096db049a7436ff928f329757fea1f5:1
    asset1 300: a9eb56da2f3422ea615f6cc68f85138ea096db049a7436ff928f329757fea1f5:1

    asset1 800: 40e16faf9c91ad89788c1e913631b4fa18986b6e645192839a0dcc441b9c2c86:2
```

If we wanted to burn a full outpoint accurately we'd have to somehow support that over our RPC (like an outpoint argument). It would require a bit of preparation here to make sure a full-burn would actually be executed. So I guess in the end I'll agree with you this is maybe too complicated to include in this test.

But as a note on these "daisy-chain" itests: It has been observed in the past that we would test things in "sterilized" environments, which could obscure an edge case from surfacing. If you always prepare a clean setup to test exactly one single feature then maybe some undesired side effects are not being exposed. For some reason here we could run into issues if we made partial/full-burns in the same run (I do not expect that to be the case).

darioAnongba: Agreed and discussed in DM.

PR #1794, opened 2025-09-12
Filter AddrReceives by timestamp

### LINKED ISSUE #700: [feature]: Allow filtering by `creation_time_unix_seconds` in `AddrReceives`
As an exchange monitoring for new deposits with polling we currently have no easy way to filter non-credited deposits other than requesting all deposits and then filtering client-side (which potentially can't scale)

The ability to filter by timestamp-greather-than (or confirmation_height greather-than) would help a lot

We're currently fetching all AddrReceives and filtering clientside

**Additional context**

darioAnongba:
### Implement timestamp filtering for the `AddrReceives` RPC endpoint.

Fixes #700 

- Added `start_timestamp` and `end_timestamp` fields to `AddrReceivesRequest`
- Regenerated all protobuf files
- Updates `AddrReceives` function in `rpcserver` to handle the new timestamp filtering by adding logic to convert the Unix timestamp to a `time.Time`.
- Adds `--start_timestamp` and `--end_timestamp` flags to the `tapcli` command.
- Adds integration test `testAddrReceives`.

#### Usage

```sh
tapcli addrs receives --addr "addr1..." --start_timestamp 1234567890 --end_timestamp 1234567890
```

```go
resp, err := client.AddrReceives(ctx, &taprpc.AddrReceivesRequest{
    FilterAddr:                  "addr1...",
    StartTimestamp: uint64(time.Now().Unix() - 3600), // 1 hour ago
    EndTimestamp: uint64(time.Now().Unix()), // now
})
```

### DIFF
--- a/cmd/commands/addrs.go
+++ b/cmd/commands/addrs.go
@@ -284,12 +284,14 @@ var receivesAddrCommand = cli.Command{
 			Usage: "show transfers of a single address only",
 		},
 		cli.Uint64Flag{
-			Name:  "start_timestamp",
-			Usage: "filter transfers created after this unix timestamp (seconds)",
+			Name: "start_timestamp",
+			Usage: "filter transfers created after this + " +
+				"unix timestamp (seconds)",
 		},
 		cli.Uint64Flag{
-			Name:  "end_timestamp",
-			Usage: "filter transfers created before this unix timestamp (seconds)",
+			Name: "end_timestamp",
+			Usage: "filter transfers created before this + " +
+				"unix timestamp (seconds)",
 		},
 	},
 	Action: addrReceives,
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -168,6 +168,9 @@
   user to specify a custom amount to send to a V2 address that doesn't have an
   amount specified.
 
+- The `AddrReceives` RPC now supports timestamp filtering with
+  [new `StartTimestamp` and `EndTimestamp` fields](https://github.com/lightninglabs/taproot-assets/pull/1794).
+
 ## tapcli Additions
 
 - [Rename](https://github.com/lightninglabs/taproot-assets/pull/1682) the mint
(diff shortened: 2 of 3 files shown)


litbot-9000:
@georgetsagk: review reminder
@ffranr: review reminder

rpcserver.go
```diff
@@ -2271,6 +2271,16 @@ func (r *rpcServer) AddrReceives(ctx context.Context,
 		sqlQuery.StatusTo = &status
 	}
 
+	// Add timestamp filtering if specified
```
jtobin, line 2274: Worth verifying that StartTimestamp < EndTimestamp here?

darioAnongba: I removed that check because we don't do it on the other RPCs that have these params but indeed might be good so we avoid an unnecessary DB query.

darioAnongba: validation added in `tapdb`, so it applies to everywhere this is called even if outside the RPC.

jtobin: This looks to have set off a few relevant itest fails.

taprpc/taprootassets.swagger.json
```diff
@@ -1288,6 +1288,16 @@
         "filter_status": {
           "$ref": "#/definitions/taprpcAddrEventStatus",
           "description": "Filter receives by a specific status. Leave empty to get all receives."
+        },
+        "start_timestamp": {
+          "type": "string",
+          "format": "uint64",
+          "description": "Filter receives by creation time greater than or equal to this timestamp.\nIf not set, no start time filtering is applied."
+        },
```
ffranr, line 1296: You should squash this commit into the previous one. The reviewer can already see which files are generated, so we can just skip those during review. There’s no need for a separate commit.

itest/test_list_on_test.go
```diff
@@ -53,6 +53,10 @@ var allTestCases = []*testCase{
 		name: "addresses",
 		test: testAddresses,
 	},
+	{
+		name: "addr receives timestamp range filtering",
+		test: testAddrReceivesTimestampFiltering,
+	},
```
ffranr, line 59: I don’t think we should add a separate itest just for timestamp filtering. Is there an existing itest we could extend to cover this new feature?

rpcserver.go
darioAnongba: should be ok now, it caught that setting StartTimestamp into the future fails, which wasn't the previous behaviour.

itest/test_list_on_test.go
darioAnongba: Same rationale as the other PR, I moved to a generic itest. Whatever PR gets merged first, I will merge the tests into the same itest.

Issue #1776, opened 2025-06-19
Consistent Custom Channel `itest` Flakes


linden:
when running `make itest` I'm seeing consistent flakes (roughly 9 out of 10 runs flake) on the `custom_channels*` tests. I've tried on both `master` and the `v0.15.0-alpha` release, on my machine and in a docker container. I'm on a fairly fast machine (2023 MacBook Pro M3 Max/128gb ram), so this is could be a race condition that the slower github actions machine isn't running into. as noted by lightninglabs/lightning-terminal#991, `time.Sleep` is used in quite a few places.

I've worked around these issues by [adding retries to the `AssetBalance` checks](https://github.com/linden/taproot-assets/commit/a354337ac2c3d02a1f11d6faad4628b623e196d7) and reverting [this commit](https://github.com/lightninglabs/lightning-terminal/commit/9cd8b38d48f69c42f368f4968a604aa94a7dce1a) ([on my fork](https://github.com/linden/lightning-terminal/tree/workaround-itest-flakes)), which passes consistently.

here's an example flake before my change to `AssetBalance`.
```
                Error Trace:    /Users/linden/go/pkg/mod/github.com/lightninglabs/taproot-assets@v0.6.0/itest/assertions.go:2323
                                                        /Users/linden/Documents/apps/linden/lightning-terminal/itest/assets_test.go:793
                                                        /Users/linden/Documents/apps/linden/lightning-terminal/itest/assets_test.go:465
                                                        /Users/linden/Documents/apps/linden/lightning-terminal/itest/litd_custom_channels_test.go:387
                                                        /Users/linden/Documents/apps/linden/lightning-terminal/itest/test_harness.go:103
                                                        /Users/linden/Documents/apps/linden/lightning-terminal/itest/litd_test.go:88
                Error:          Not equal: 
                                expected: 0xc3500
                                actual  : 0x0
                Test:           TestLightningTerminal/custom_channels
                Messages:       ListAssets balance, wanted 800000, got: {
                                    "assets": [],
                                    "unconfirmed_transfers": "1",
                                    "unconfirmed_mints": "0"
                                }
```

and after my change to `AssetBalance` but before reverting the commit:
```
    assets_test.go:1759: 
                Error Trace:    /Users/linden/Documents/apps/linden/lightning-terminal/itest/assets_test.go:1759
                                                        /Users/linden/Documents/apps/linden/lightning-terminal/itest/litd_custom_channels_test.go:2899
[...]

PR #1799, opened 2025-09-16
itest: introduce `rpcassert` package with RPC helpers to reduce flakes

ffranr:
This PR introduces a new `rpcassert` package that provides helper functions for common RPC calls in itests.

The helpers wrap RPC calls in `wait.NoError`, adding retries and more lenient timing expectations. This helps reduce test flakes by giving assertions a larger time window to succeed.

Included helpers:

* `NewAddrRPC`
* `ListBalancesRPC`
* `ListAssetsRPC`
* `ListUtxosRPC`

These are now used in existing test functions such as `AssertBalances` and `NewAddrWithEventStream` to improve stability.

See also: https://github.com/lightninglabs/taproot-assets/issues/1776

### DIFF
--- a/itest/rpcassert/taprpc.go
+++ b/itest/rpcassert/taprpc.go
@@ -124,3 +124,40 @@ func ListAssetsRPC(t *testing.T, ctx context.Context,
 
 	return resp
 }
+
+// ListUtxosRPC calls ListUtxos RPC with the given request until the
+// given assertion predicate returns no error or the timeout is reached. If the
+// predicate is nil, only basic checks are performed (non-nil response).
+//
+// If the assertion fails, the test is failed.
+func ListUtxosRPC(t *testing.T, ctx context.Context,
+	client taprpc.TaprootAssetsClient,
+	assertPredicate func(*taprpc.ListUtxosResponse) error,
+	req *taprpc.ListUtxosRequest) *taprpc.ListUtxosResponse {
+
+	t.Helper()
+
+	var resp *taprpc.ListUtxosResponse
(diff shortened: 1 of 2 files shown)


litbot-9000:
@georgetsagk: review reminder

ffranr:
> I've got little context around the flaky tests but something feels off with the way we are retrying. Since the test environment is fully controlled, shouldn't the `wait` statements be on the specific RPC call(s) instead of the whole `assertBalances`?

In the latest commits I’ve included helper functions for each RPC call to make the solution more general and direct.

itest/rpcassert/taprpc.go
```diff
@@ -0,0 +1,163 @@
...
+			return err
+		}
+
+		if resp == nil {
+			return fmt.Errorf("nil response")
+		}
+
+		if assertPredicate != nil {
+			return assertPredicate(resp)
+		}
```
GeorgeTsagk, line 154: Here and above I can see this pattern being repeated, we can extract this into a generic helper where we get
`T, error` here and the predicate also accepts `T`.

Non-blocking, if you can figure it out it'd be nice to have.

ffranr: thanks for taking a look. I tried playing around with generics on this, but i just thought code complexity just got higher for very little benefit IMO. Also, maybe we might feel a greater need for generics once we expand the set a bit?

i wrote the code such that the repeating pattern was obvious. But making it more succinct doesn't make it simpler IMO

We discussed offline. I will merge as is right now.

PR #1801, opened 2025-09-16
supplycommit+rpcserver: move ignore asset logic from RPC server to supplycommit pkg

ffranr:
Remove logic from the RPC server's `IgnoreAssetOutPoint` endpoint and move it to the `supplycommit` package.

Add signing capabilities to `supplycommit.Manager` and construct signed ignore tuples within the manager instead of the RPC server.

Improves separation of concerns.


universe/supplycommit/manager.go
```diff
@@ -273,6 +280,125 @@ func (m *Manager) SendEvent(ctx context.Context,
...
+			"commitments: %w", err)
+	}
+
+	// Retrieve asset meta reveal for the asset ID. This will be used to
+	// obtain the supply commitment delegation key.
+	metaReveal, err := m.cfg.AssetLookup.FetchAssetMetaForAsset(
+		ctx, assetID,
+	)
+	if err != nil {
+		return zero, fmt.Errorf("faild to fetch asset meta: %w", err)
```
jtobin: nit: s/faild/failed (this was already in the old code, apparently)

rpcserver.go
```diff
@@ -3998,83 +3998,14 @@ func (r *rpcServer) IgnoreAssetOutPoint(ctx context.Context,
 		req.AssetAnchorPoint)
 
 	assetID := req.AssetAnchorPoint.ID
+	assetSpec := asset.NewSpecifierFromId(assetID)
```
jtobin, line 4001: Is there a particular reason not to construct the assetSpec with asset.NewSpecifierOptionalGroupKey as was done before, and then pass *that* assetSpec to the new method?

universe/supplycommit/manager.go
```diff
@@ -273,6 +280,125 @@ func (m *Manager) SendEvent(ctx context.Context,
...
+	// group key.
+	assetGroup, err := m.cfg.AssetLookup.QueryAssetGroupByID(ctx, assetID)
+	if err != nil {
+		return zero, fmt.Errorf("fail to find asset group given "+
+			"asset ID: %w", err)
+	}
+
+	// If the asset specifier includes a group key, ensure it matches the
+	// group key of the asset group fetched from the database.
+	if assetSpec.HasGroupPubKey() {
```
jtobin, line 308: At present I think the new method is only ever called with an assetSpec constructed from asset.NewSpecifierFromId, so this block feels strange -- assetSpec.HasGroupPubKey() will always be false.

universe/supplycommit/manager.go
```diff
@@ -273,6 +280,125 @@ func (m *Manager) SendEvent(ctx context.Context,
...
+		}
+
+		if !assetGroup.GroupKey.GroupPubKey.IsEqual(givenGroupKey) {
+			return zero, fmt.Errorf("provided group key does " +
+				"not match asset group")
+		}
+	}
+
+	// Formulate an asset specifier from the asset ID and group key.
+	assetSpec = asset.NewSpecifierOptionalGroupKey(
```
jtobin, line 321: Yeah, it feels like *this* is the assetSpec that should be passed to the new method. But the new commentary in the above error messages looks intentionally different, so possible I'm missing something?

jtobin (commented):
See my comments/questions re: the assetSpec handling, otherwise LGTM.

universe/supplycommit/manager.go
ffranr: `Manager.IgnoreAssetOutPoint` now supports cases where the caller omits the asset group key from the asset specifier. While current call sites never provide a group key, so the `if assetSpec.HasGroupPubKey()` check is redundant in practice, keeping it ensures the method is prepared to validate a supplied group key if future callers include one.

rpcserver.go
ffranr: Here's my perspective:

My goal is to keep `rpcServer.IgnoreAssetOutPoint` as slim as possible. Since the RPC message doesn’t provide or require an asset group key, that lookup has to happen somewhere. Moving it into `Manager.IgnoreAssetOutPoint` simplifies the RPC server and makes the manager more capable, as it can now handle cases where the asset group key is either specified or omitted.

darioAnongba (approved):
LGTM. Small remark being that I see usage of the word `Manager` usually as a sign of bad architecture (`Manager` meaning "a component that does something" so you should be able to give it the name of the thing it actually does and limit its scope to it).

But this PR doesn't add this component so irrelevant comment lol.

PR #1804, opened 2025-09-16
gemini: add config and style guidelines

Roasbeef:


.gemini/styleguide.md
```diff
@@ -0,0 +1,331 @@
+# LND Style Guide
```
ffranr: Should we remove this line? Is this guide copy and paste from LND or should i review every line?

Roasbeef: This is just copy from lnd, I can update this to say lnd.

Issue #1745, opened 2025-08-18
[feature]: Add group key to QueryPeerAcceptedQuotes response


w3irdrobot:
**Is your feature request related to a problem? Please describe.**

Right now, it appears that you can get accepted quotes using the `QueryPeerAcceptedQuotes` method. However, it doesn't appear possible to filter based on a certain asset ID or group key whether in the request or manually in the response. The response doesn't include the asset ID or group key of the asset for which the quote was made.

**Describe the solution you'd like**

At the very least, it'd be great to have the group key available in the `PeerAcceptedBuyQuote` and `PeerAcceptedSellQuote` so this filtering could be done on the client side.

**Describe alternatives you've considered**

Manually tracking the group keys that go with requested sell orders using a map from `PeerAcceptedSellQuote.id` to group key. This can be done, but it seems this would be something lots of clients would end up reproducing. 

**Additional context**

PR #1805, opened 2025-09-16
rfq+taprpc: add asset ID and group key to `PeerAccepted[Buy/Sell]Quote`

ffranr:
Closes: https://github.com/lightninglabs/taproot-assets/issues/1745

---

This PR extends the `rfqrpc.PeerAcceptedBuyQuote` and `rfqrpc.PeerAcceptedSellQuote` proto messages with `asset_id` and `group_key` fields.

Previously, the `QueryPeerAcceptedQuotes` response did not include asset information, making it impossible to filter accepted quotes by asset ID or group key. With these additions, clients can now directly associate a quote with its corresponding asset without having to track mappings manually.

Changes in this PR:

* **taprpc:** Add `asset_id` and `group_key` fields to `PeerAcceptedBuyQuote` and `PeerAcceptedSellQuote`.
* **rfq:** Populate the new fields when returning accepted quotes.
* **itest:** Add assertions to verify correct population of `asset_id` and `group_key` in accepted quote messages.

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -145,6 +145,12 @@
 - [Rename](https://github.com/lightninglabs/taproot-assets/pull/1682) the
   `MintAsset` RPC message field from `universe_commitments` to
   `enable_supply_commitments`.
+- [Enhanced RFQ accepted quote messages with asset identification fields](https://github.com/lightninglabs/taproot-assets/pull/1805):
+  The `PeerAcceptedBuyQuote` and `PeerAcceptedSellQuote` proto messages
+  now include asset ID and asset group pub key fields (via the `AssetSpecBytes`
+  message), allowing clients to directly associate quotes with their
+  corresponding assets without manual tracking.
+
 - The `SubscribeSendEvents` RPC now supports [historical event replay of 
   completed sends with efficient database-level
   filtering](https://github.com/lightninglabs/taproot-assets/pull/1685).


rfq/marshal.go
```diff
@@ -47,6 +50,19 @@ func MarshalAcceptedSellQuote(
...
+	// Populate asset ID and/or group key based on the asset specifier.
+	accept.Request.AssetSpecifier.WhenId(func(assetId asset.ID) {
+		quote.AssetId = assetId[:]
+	})
+
+	accept.Request.AssetSpecifier.WhenGroupPubKey(
+		func(groupKey btcec.PublicKey) {
+			quote.AssetGroupKey = schnorr.SerializePubKey(&groupKey)
+		},
+	)
```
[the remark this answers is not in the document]
ffranr: no

taprpc/rfqrpc/rfq.proto
```diff
@@ -283,6 +283,12 @@ message PeerAcceptedBuyQuote {
     // authentication information that the price oracle can use to give out a
     // more accurate (or discount) asset rate.
     string price_oracle_metadata = 8;
+
+    // The asset ID of the subject asset.
+    bytes asset_id = 9;
```
GeorgeTsagk: Seems like the way we used to define asset types before drafting the idea of asset specifiers.

Why not re-use the asset specifier we have already defined as a proto msg?
https://github.com/lightninglabs/taproot-assets/blob/32e18c031781f5cb26615f1d992e6c2d05cd7b2a/taprpc/rfqrpc/rfq.proto#L73-L88

darioAnongba: seems like a valid concern as using `bytes` won't work outside `gRPC`.

docs/release-notes/release-notes-0.7.0.md
```diff
@@ -145,6 +145,12 @@
 - [Rename](https://github.com/lightninglabs/taproot-assets/pull/1682) the
   `MintAsset` RPC message field from `universe_commitments` to
   `enable_supply_commitments`.
+- [Enhanced RFQ accepted quote messages with asset identification
```
darioAnongba: just verifying if this is supposed to go in 0.7.0

GeorgeTsagk: yeap I believe goal is to have it as part of rc2

taprpc/rfqrpc/rfq.proto
ffranr: > Why not re-use the asset specifier we have already defined as a proto msg?

Because it wont let us specify both the asset ID and asset group key. And I don't see why we shouldn't if they are both available.

docs/release-notes/release-notes-0.7.0.md
ffranr: I think it's close to merge so we might as well get it into 0.7

taprpc/rfqrpc/rfq.proto
ffranr: Maybe we add a new AssetSpecifier proto message type...

ffranr: > seems like a valid concern as using bytes won't work outside gRPC.

I think `bytes` field type will work in REST response. It's just encoded as a base64-encoded string in JSON.

---

`AssetSpecifier` proto message is only currently used in requests so the caller can choose which field to populate (via string or bytes for example as a convenience).

I'm going to add a new field that I think will be more widely useful:
```
// AssetSpecBytes defines the raw byte-level specification for an asset.
message AssetSpecBytes {
    // The 32-byte asset ID specified as raw bytes.
    bytes id = 1;

    // The 32-byte asset group public key, serialized in BIP340 format.
    // BIP340 defines a canonical encoding for Schnorr public keys.
    // This field is serialized using schnorr.SerializePubKey.
    bytes group_pub_key = 2;
}
```

darioAnongba: ok I see, the purpose of the `AssetSpecBytes` is to pass either the asset Id or the group key (or both?) in order to associate it with the quote, whereas the `AssetSpecifier` mentioned is a convenience type to allow passing either bytes or strings.

So LGTM but I'd say that I don't see the purpose of this type and I'd revert to having both fields at the root. Also the  name `AssetSpecBytes` is not ideal as it contains the word `Bytes` that is irrelevant to the API.

ffranr: > So LGTM but I'd say that I don't see the purpose of this type and I'd revert to having both fields at the root.

It just helps with scoping related types. I think the `AssetSpecBytes` message is sufficiently general to be used broadly.

As for the name, I think you're right, and I agree with you. My goal was to differentiate it from the existing `AssetSpecifier` message type. I could rename to `AssetSpec`? We do use that in code quite a bit, what do you think?

darioAnongba: Yes naming is hard here since it is essentially the `AssetIdentifier`, which is a tuple `{id, group_key}` but `AssetId` is already taken. Sadly `AssetSpec` means `AssetSpecifier` so it's the same name. Would `AssetCorrelationId` make sense or too specific to this use case?
In any case, I'm fine with `AssetSpec`

PR #1810, opened 2025-09-18
Chaos test STXO proof gen: swap `EmptyGenesisID` for `spentAsset.TapCommitmentKey()`

ffranr:
This PR replaces the tap commitment key used in STXO proof generation from `EmptyGenesisID` (which should be correct) to `spentAsset.TapCommitmentKey()` (which should be incorrect).

### Purpose

The change is intentionally experimental — to test whether this substitution causes any failures in unit or integration tests.

### Notes

* Not intended for merge.

### DIFF
--- a/proof/append.go
+++ b/proof/append.go
@@ -221,7 +221,7 @@ func CreateTransitionProof(prevOut wire.OutPoint, params *TransitionParams,
 			// Generate an STXO inclusion proof for each prev
 			// witness.
 			_, stxoProof, err := params.TaprootAssetRoot.Proof(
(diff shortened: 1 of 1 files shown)


ffranr:
Apparently, in the instance that I've modified, `spentAsset.TapCommitmentKey()` should be equal to `EmptyGenesisID` because the spent asset has an empty genesis. See: https://github.com/lightninglabs/taproot-assets/blob/b21d5d2ee9d4b3ae4e5e2cb5a9a9157560bac476/asset/asset.go#L2317

PR #1812, opened 2025-09-19
Burn assets by group key

darioAnongba:
Enable burning of assets by specifying a group key.

- Add `AssetSpecifier` to `tapcommon.proto` and use it in the `BurnAssetRequest`
- Change `tapcli` to accept `group_key` param in `tapd assets burn` commands
- Update `tapfreighter` to support passing a group key in the asset specifier
- Update `BurnAsset` RPC request to accept an `AssetSpecifier`. **Breaking change!**
- Update integration tests


.gitignore
```diff
@@ -55,3 +55,6 @@ vendor/
 # Release builds
 /taproot-assets-*
 .aider*
+
+# Environment variables
+.env
```
GeorgeTsagk: Why was this added?

Regardless of reason, should add it in a separate commit which also includes some context

taprpc/taprootassets.proto
```diff
@@ -1682,13 +1682,8 @@ message FetchAssetMetaResponse {
...
 message BurnAssetRequest {
-    oneof asset {
-        // The asset ID of the asset to burn units of.
-        bytes asset_id = 1;
-
-        // The hex encoded asset ID of the asset to burn units of.
-        string asset_id_str = 2;
-    }
+    // The asset to burn units of.
+    AssetSpecifier asset = 1;
```
GeorgeTsagk: we want to avoid re-assigning numbers to new message types as that means we're breaking backwards compatibility.

If we want to use the new specifier type here we should do it as
```
    AssetSpecifier asset = 6;
```

and preserve the old `oneof asset` approach with a deprecated flag

cmd/commands/assets.go
```diff
@@ -1106,15 +1110,56 @@ var burnAssetsCommand = cli.Command{
 	Action: burnAssets,
 }
 
+// buildAssetSpecifier validates and constructs
+// an AssetSpecifier from CLI context
+func buildAssetSpecifier(ctx *cli.Context) (*taprpc.AssetSpecifier, error) {
+	assetIDHex := ctx.String(assetIDName)
+	groupKeyHex := ctx.String(assetGroupKeyName)
+
+	// Either asset ID or group key must be provided, but not both
```
GeorgeTsagk: missing full stop
```suggestion
	// Either asset ID or group key must be provided, but not both.
```

cmd/commands/assets.go
```diff
@@ -1106,15 +1110,56 @@ var burnAssetsCommand = cli.Command{
...
+		return nil, fmt.Errorf("only one of asset ID or group key " +
+			"can be specified")
+	}
+
+	var assetSpecifier *taprpc.AssetSpecifier
+	if assetIDHex != "" {
+		assetIDBytes, err := hex.DecodeString(assetIDHex)
+		if err != nil {
+			return nil, fmt.Errorf("invalid asset ID: %w", err)
+		}
```
GeorgeTsagk, line 1138: missing newline after bracket
```suggestion
		}

```

cmd/commands/assets.go
```diff
@@ -1127,42 +1172,85 @@ func burnAssets(ctx *cli.Context) error {
 	defer cleanUp()
 
 	if !ctx.Bool(burnOverrideConfirmationName) {
-		balance, err := client.ListBalances(
-			ctxc, &taprpc.ListBalancesRequest{
-				GroupBy: &taprpc.ListBalancesRequest_AssetId{
-					AssetId: true,
+		var balanc *taprpc.ListBalancesResponse
```
GeorgeTsagk: typo?
```suggestion
		var balance *taprpc.ListBalancesResponse

```

cmd/commands/assets.go
```diff
@@ -1106,15 +1110,56 @@ var burnAssetsCommand = cli.Command{
 	Action: burnAssets,
 }
 
+// buildAssetSpecifier validates and constructs
+// an AssetSpecifier from CLI context
```
GeorgeTsagk: you can fit more words into the line above

cmd/commands/assets.go
```diff
@@ -1127,42 +1172,85 @@ func burnAssets(ctx *cli.Context) error {
...
 
 	if !ctx.Bool(burnOverrideConfirmationName) {
-		balance, err := client.ListBalances(
-			ctxc, &taprpc.ListBalancesRequest{
-				GroupBy: &taprpc.ListBalancesRequest_AssetId{
-					AssetId: true,
+		var balanc *taprpc.ListBalancesResponse
+		var err error
+		var identifierBytes []byte
+		var currentBalance uint64
```
GeorgeTsagk: can group all variable declarations together
```go
var (
	balance			 *taprpc.ListBalancesResponse
	err				 error
	identifierBytes	 []byte
	currentBalance	 uint64
)
```

cmd/commands/assets.go
```diff
@@ -1127,42 +1172,85 @@ func burnAssets(ctx *cli.Context) error {
...
+
+		assetIDHex := ctx.String(assetIDName)
```
GeorgeTsagk: didn't we already parse this into `assetSpecifier`?

can just do `assetSpecifier.GetAssetId()`

cmd/commands/assets.go
```diff
@@ -1127,42 +1172,85 @@ func burnAssets(ctx *cli.Context) error {
...
+		groupKeyHex := ctx.String(assetGroupKeyName)
+
+		if assetIDHex != "" {
```
GeorgeTsagk: when having this kind of if/else pattern where we try to match the first non-nil variable I find it more readable to unfold over a switch statement as such:

```go
switch {
case assetSpecifier.GetAssetId() != nil:
	// do something

case assetSpecifier.GetGroupKey() != nil:
	// do something else
}
```

cmd/commands/assets.go
```diff
@@ -1127,42 +1172,85 @@ func burnAssets(ctx *cli.Context) error {
...
-			},
-		)
-		if err != nil {
-			return fmt.Errorf("unable to list current asset "+
-				"balances: %w", err)
+			)
+			if err != nil {
+				return fmt.Errorf("unable to list current "+
+					"asset balances: %w", err)
+			}
```
GeorgeTsagk, line 1208: missing newline after bracket, same for below

cmd/commands/assets.go
GeorgeTsagk: also the `err error` can be skipped could just use inline declaration by using the `:=` assignment operator

itest/burn_test.go
```diff
@@ -523,4 +529,29 @@ func testBurnGroupedAssets(t *harnessTest) {
 	require.Equal(t.t, burnAmt, burn.Amount)
 	require.Equal(t.t, burnNote, burn.Note)
 	require.Equal(t.t, assetGroupKey, burn.TweakedGroupKey)
+
+	// Test case 2: Burn by group key close to the full
```
GeorgeTsagk: When either https://github.com/lightninglabs/taproot-assets/pull/1791 or this PR get merged, we should enhance this with yet another test case, which would be a full group key burn.

tapfreighter/fund.go
```diff
@@ -45,16 +45,6 @@ func createFundedPacketWithInputs(ctx context.Context, exporter proof.Exporter,
 		map[asset.PrevID]*AnchoredCommitment, len(selectedCommitments),
```
GeorgeTsagk, line 45: This commit should precede the previous ones.

We usually want to first introduce the functionality and then expose it to other systems/RPC

tapfreighter/wallet.go
```diff
@@ -689,6 +688,15 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
 		return nil, err
 	}
 
+	for _, vPkt := range pkt.VPackets {
```
GeorgeTsagk, line 686: why was this moved here?

tapfreighter/wallet.go
```diff
@@ -808,36 +780,63 @@ func (f *AssetWallet) FundBurn(ctx context.Context,
...
+		if err := tapsend.PrepareOutputAssets(ctx, pkt); err != nil {
+			return nil, fmt.Errorf("unable to prepare burn + "+
+				"outputs: %w", err)
+		}
 	}
 
 	// We want to avoid a BTC output being created that just sits there
 	// without an actual commitment in it. So if we are not getting any
 	// change or passive assets in this output, we'll not want to go through
 	// with it.
```
GeorgeTsagk: you could maybe try basing this PR on top of https://github.com/lightninglabs/taproot-assets/pull/1791 so that we have the full view of the end-result here.

cmd/commands/assets.go
darioAnongba: True! Seems like a residue from the old code that I didn't properly adapt, thanks! We still need the hex strings though to be used as an index after, but it's better to encode than to re-decode.

itest/burn_test.go
darioAnongba: Yes, that is indeed the plan.

tapfreighter/wallet.go
darioAnongba: As you can see, the old way was to run `PrepareOutputAssets` inside `createFundedPacketWithInputs`. But with the logic I introduced, we perform some operations on the `vPkt` before calling `PrepareOutputAssets` so we need to extract it to have control of when `PrepareOutputAssets` is called (or call it twice, which is inefficient and could have side undesirable effects).

tapfreighter/wallet.go
darioAnongba: True that I think #1791 is going to be merged before this PR but the two are independent. It is not necessary for #1791 to be merged before this one. The final result will be on the rebased PR (either this one or 1791).

taprpc/taprootassets.proto
darioAnongba: I added the `asset` field at index 6 and kept the old one but sadly both use the name `asset` so we have a conflict. I think not using the name `asset` is pretty bad here so I renamed the old one `asset_legacy` but that is still a breaking change. Please let me know your thoughts.

ffranr: I don't think we should rename the existing field. Perhaps use name `asset_spec` for new field.

itest/burn_test.go
darioAnongba: test updated to be a full burn.

tapfreighter/wallet.go
darioAnongba: PR rebased.

taprpc/taprootassets.proto
darioAnongba: Used `asset_specifier` as discussed. Keeps also old behaviour so users can pass both.

litbot-9000:
@georgetsagk: review reminder
@ffranr: review reminder

taprpc/tapcommon.proto
```diff
@@ -34,3 +34,20 @@ message AssetOutPoint {
     // asset is locked to.
     bytes script_key = 3;
 }
+
+message AssetSpecifier {
+    oneof id {
```
ffranr: since this is a new message type and it's not used elsewhere, maybe we don't need `oneof id` here. The user can specify ID and group key if they like.

tapfreighter/fund.go
```diff
@@ -45,16 +45,6 @@ func createFundedPacketWithInputs(ctx context.Context, exporter proof.Exporter,
...
 	for _, anchorAsset := range selectedCommitments {
-		// We only use the inputs of assets of the same TAP commitment
-		// as we want to fund for. These are the active assets that
-		// we're going to distribute. All other assets are passive and
-		// will be detected and added later.
-		if anchorAsset.Asset.TapCommitmentKey() !=
-			fundDesc.TapCommitmentKey() {
-
-			continue
-		}
```
ffranr, line 56: `createFundedPacketWithInputs` is also used outside the asset burn context. Is it safe to remove this code for those other use cases as well?

We might want to consider removing `fundDesc *tapsend.FundingDescriptor` function arg also if we're going to make this deletion.

ffranr: I think you should consider adding more to the commit message to describe what changes you're making.

rpcserver.go
```diff
@@ -3678,36 +3660,104 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
-	var groupKey *btcec.PublicKey
-	assetGroup, err := r.cfg.TapAddrBook.QueryAssetGroupByID(ctx, assetID)
+	var (
+		assetID  *asset.ID
+		groupKey *btcec.PublicKey
+		err      error
+	)
+
+	// TODO: Remove this switch once the deprecated asset field
+	// is removed. Keep only the AssetSpecifier case.
```
ffranr: Usually with `TODO` comments, we write them as `TODO(<username>)`. In this case, we should also include the release version in the comment so that it's easy to see how many major releases have passed since it was written and decide when to remove the deprecated field.

Git history could provide that information, but it's harder to see at a glance, especially if the code is later moved or refactored.

rpcserver.go
```diff
@@ -3678,36 +3660,104 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse asset "+
+				"specifier: %w",
+				err)
+		}
+
+	default:
+		return nil, fmt.Errorf("either asset or asset_specifier " +
+			"must be specified")
```
ffranr: error message here can just be: `asset_specifier field unset`

We don't want the error message to encourage them to set a depreciated field.

rpcserver.go
```diff
@@ -3678,36 +3660,104 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
+		assetID, groupKey, err = parseAssetSpecifier(
+			in.AssetSpecifier.GetAssetId(),
+			in.AssetSpecifier.GetAssetIdStr(),
+			in.AssetSpecifier.GetGroupKey(),
+			in.AssetSpecifier.GetGroupKeyStr(),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse asset "+
+				"specifier: %w",
+				err)
```
ffranr: `err)` can be on line above

rpcserver.go
```diff
@@ -3678,36 +3660,104 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
 
-	var serializedGroupKey []byte
+	var (
+		assetIDBytes       []byte
+		serializedGroupKey []byte
+	)
+	if assetID != nil {
+		id := *assetID
+		assetIDBytes = id[:]
+	}
```
ffranr: can use `fn.ByteSlice` here
```
	if assetID != nil {
		assetIDBytes = fn.ByteSlice(*assetID)
	}
```

rpcserver.go
```diff
@@ -3678,36 +3660,104 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
 	if groupKey != nil {
 		serializedGroupKey = groupKey.SerializeCompressed()
 	}
-
 	rpcsLog.Infof("Burning asset (asset_id=%x, group_key=%x, "+
-		"burn_amount=%d)", assetID[:], serializedGroupKey,
+		"burn_amount=%d)", assetIDBytes, serializedGroupKey,
```
ffranr: if we move this below then we can just log `assetSpecifier.String()` and the burn amount

ffranr: Then i think you can drop `assetIDBytes` and `serializedGroupKey`.

rpcserver.go
```diff
@@ -3747,23 +3793,23 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
 			err)
 	}
 
+	// Return the first burn proof found.
 	var burnProof *taprpc.DecodedProof
```
ffranr: I think we should return all burn proofs rather than just the first.

cmd/commands/assets.go
```diff
@@ -1106,15 +1110,59 @@ var burnAssetsCommand = cli.Command{
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid group key: %w", err)
+		}
+
+		assetSpecifier = &taprpc.AssetSpecifier{
+			Id: &taprpc.AssetSpecifier_GroupKey{
+				GroupKey: groupKeyBytes,
+			},
+		}
+	}
```
ffranr: In general, I think we should support populating an asset specifier that can have **both** the asset ID and the asset group key set. This would let a user specify a specific tranche within a group and give us the opportunity to sanitize the input by cross-referencing the provided group key with the group we fetch based on the user-supplied asset ID.

cmd/commands/assets.go
```diff
@@ -1106,15 +1110,59 @@ var burnAssetsCommand = cli.Command{
...
 	if ctx.NArg() != 0 || ctx.NumFlags() == 0 {
 		return cli.ShowSubcommandHelp(ctx)
 	}
 
-	assetIDHex := ctx.String(assetIDName)
-	assetIDBytes, err := hex.DecodeString(assetIDHex)
+	assetSpecifier, err := buildAssetSpecifier(ctx)
 	if err != nil {
-		return fmt.Errorf("invalid asset ID")
+		return err
```
ffranr: `fmt.Errorf` might still be useful here: `fmt.Errorf("failed to build asset specifier: %w", err)`

cmd/commands/assets.go
```diff
@@ -1127,42 +1175,84 @@ func burnAssets(ctx *cli.Context) error {
...
+				return fmt.Errorf("couldn't fetch balance for "+
+					"asset ID %x", idBytes)
+			}
+
+			currentBalance = assetBalance.Balance
+
+		case assetSpecifier.GetGroupKey() != nil:
+			idBytes = assetSpecifier.GetGroupKey()
+			idType = "Group Key"
+			idHex := hex.EncodeToString(idBytes)
```
ffranr: are these variables named correctly? maybe they should be `groupKeyBytes` and so on?

itest/burn_test.go
```diff
@@ -120,18 +120,19 @@ func testBurnAssets(t *harnessTest) {
...
 		AmountToBurn:     burnAmt,
 		Note:             burnNote,
 		ConfirmationText: taprootassets.AssetBurnConfirmationText,
 	})
 	require.NoError(t.t, err)
 
-	burnRespJSON, err := formatProtoJSON(burnResp)
+	_, err = formatProtoJSON(burnResp)
 	require.NoError(t.t, err)
-	t.Logf("Got response from burning %d units: %v", burnAmt, burnRespJSON)
```
ffranr, line 134: I'm not sure if i see the point in `_, err = formatProtoJSON(burnResp)` if we're removing the log call

ffranr:
I think this should have base branch `0-8-0-staging`?

tapfreighter/fund.go
darioAnongba: Good catch! This shouldn't have been changed. I don't know exactly if it's useful and the tests pass without it but it's not the purpose of this PR to change this. Reverted.

rpcserver.go
darioAnongba: Applied.

rpcserver.go
darioAnongba: removed as suggested.

rpcserver.go
darioAnongba: great!

rpcserver.go
darioAnongba: I wasn't sure about this breaking change but you're right it doesn't make sense otherwise.

cmd/commands/assets.go
darioAnongba: renamed.

itest/burn_test.go
darioAnongba: yes I thought that the assertion was needed but since these are proto messages they can't fail the unmarshalling. removed them.

darioAnongba:
comments applied (and answered) and pointed to `0-8-0-staging`

taprpc/tapcommon.proto
darioAnongba: I don't think it's wise to consider this a new message type as this exact type is defined in `rfq.proto` and `price_oracle.proto`. These proto don't depend on `tapcommon` so I didn't want to add this dependency but the best would be to remove `AssetSpecifier` from them and use this type from `tapcommon` everywhere. WDYT?

cmd/commands/assets.go
darioAnongba: I've mentioned the problem of having 2 identifiers for the same resource and we can see it manifesting again in this comment. Allowing the user to pass both `AssetId` and `GroupKey` doesn't make sense because the group key is implied from the asset ID so if a user passes both, we will either:
- Ignore the `GroupKey` since it's redundant and we already have the `AssetID`. But nothing tells us that this is what the user wanted, so it's an unintended side effect.
- Replace the `GroupKey` with the correct one. This is a side effect and not sanitization as we are modifying the user request without his consent or knowledge.
- Return errors such as `group key does not correspond to given asset id` or `asset id not found for given group key`, etc., which are prevented by not allowing both asset ID or group key. 

But also, `AssetSpecifier` is used in different APIs as a oneOf and this will be create an inconsistent behavior across the API.

ffranr: Suppose I wanted to extend the `ListTransfersResponse` proto message to include the asset group key in addition to the existing asset ID field. This would let the caller organize transfers by asset group without needing a second call to map asset IDs to group keys. Essentially, I want to include the asset group key in both `TransferInput` and `TransferOutput`.

If `tapcommon.AssetSpecifier` were general (common) enough, we might be able to reuse it and deprecate the current asset ID field. What do you think makes sense in this case? For response messages, it might be useful to include both the asset ID and the asset group key for flexibility, allowing the response to convey not only the asset group but also the specific tranche within that group.

rpcserver.go
```diff
@@ -3705,36 +3687,95 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
+	default:
+		return nil, fmt.Errorf("asset_specifier field unset")
 	}
 
-	rpcsLog.Infof("Burning asset (asset_id=%x, group_key=%x, "+
-		"burn_amount=%d)", assetID[:], serializedGroupKey,
-		in.AmountToBurn)
+	// If the asset ID is set but the group key is not, we need to query
+	// the asset group by the asset ID to get the group key.
+	if assetID != nil && groupKey == nil {
```
GeorgeTsagk: So we're also silently dropping support for burning a specific asset ID. With this line we're **always trying to find the group** and burn based on the group key.

itest/burn_test.go
```diff
@@ -486,7 +473,7 @@ func testBurnGroupedAssets(t *harnessTest) {
 	)
 
 	// Our asset balance should have been decreased by the burned amount.
-	AssertBalanceByID(t.t, t.tapd, burnAssetID, postBurnAmt)
+	AssertBalanceByID(t.t, t.tapd, burnAssetID2, postBurnAmt)
```
GeorgeTsagk, line 464: since the assets are grouped, how can we be sure about the balance status of specific asset IDs here? Shouldn't we instead assert the balance by group?

itest/burn_test.go
```diff
@@ -507,6 +494,48 @@ func testBurnGroupedAssets(t *harnessTest) {
 	require.Equal(t.t, burnAmt, burn.Amount)
 	require.Equal(t.t, burnNote, burn.Note)
 	require.Equal(t.t, assetGroupKey, burn.TweakedGroupKey)
+
+	// Test case 2: Burn by group key to we select multiple inputs.
+	burnAmt = totalAmt - burnAmt - 1
```
GeorgeTsagk: it's not really intuitive -- why do we `-1` here?

itest/burn_test.go
```diff
@@ -507,6 +494,48 @@ func testBurnGroupedAssets(t *harnessTest) {
...
+		AmountToBurn:     burnAmt,
+		ConfirmationText: taprootassets.AssetBurnConfirmationText,
+	})
+	require.NoError(t.t, err)
+
+	// When burning by group key with multiple inputs, the coin selection
+	// can vary. We verify that:
+	// 1. There are 3 outputs total (1 change + 2 burn outputs)
+	// 2. The sum of all outputs equals the total remaining balance
+	// 3. One output is the change amount (1 unit)
```
GeorgeTsagk: do we explicitly want for change to exist? (for test coverage purposes)

docs/release-notes/release-notes-0.7.0.md
```diff
@@ -195,6 +195,11 @@
 
 - The `AddrReceives` RPC now supports timestamp filtering with
   [new `StartTimestamp` and `EndTimestamp` fields](https://github.com/lightninglabs/taproot-assets/pull/1794).
+  
+- The `BurnAsset` RPC now supports a 
+  [new `AssetSpecifier` field](https://github.com/lightninglabs/taproot-assets/pull/1812)
+  that allows the user to  specify the asset to burn by ID or GroupKey.
```
GeorgeTsagk: Should explicitly mention that burning specific asset IDs of grouped assets is no longer supported.

Also should move release notes to `docs/release-notes/release-notes-0.7.1.md`

GeorgeTsagk (commented):
getting there
some final Qs

taprpc/tapcommon.proto
darioAnongba: Changed definition to allow both to be passed.

cmd/commands/assets.go
darioAnongba: Changed definition to allow both to be passed.
The behaviour is now:
```
Handle different scenarios based on what was provided:
1. If only assetID is provided, look up the group key.
2. If only groupKey is provided, use it directly
3. If both are provided, validate consistency.
```

rpcserver.go
darioAnongba: The previous behavior stays unchanged regarding asset ID. As you can see, this was also done before but I moved it outside of the switch to support both using the old `Asset` and the new `AssetSpecifier`.
If I somehow changed the logic, it isn't intended.

itest/burn_test.go
darioAnongba: Burning by asset ID should work the same. This part of the test doesn't change.

itest/burn_test.go
darioAnongba: Good catch, now we support full burns so I can do a full burn here instead! Fixed.

itest/burn_test.go
darioAnongba: Changed to a full burn so no change anymore.

docs/release-notes/release-notes-0.7.0.md
darioAnongba: Moved to `0-8-0.md `. Normally burning by assetID should work the same.

tapfreighter/wallet.go
```diff
@@ -803,11 +775,42 @@ func (f *AssetWallet) FundBurn(ctx context.Context,
 		return nil, err
 	}
 
-	// We don't support burning by group key yet, so we only expect a single
-	// vPacket (which implies a single asset ID is involved).
-	if len(fundedPkt.VPackets) != 1 {
-		return nil, fmt.Errorf("expected a single vPacket, got %d",
-			len(fundedPkt.VPackets))
+	// Now that inputs are assigned, set the proper burn script
+	// key and output versions per virtual packet,
```
GeorgeTsagk: nit: can fit more words into previous line

tapfreighter/wallet.go
```diff
@@ -684,6 +683,15 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
 		return nil, err
 	}
 
+	for _, vPkt := range pkt.VPackets {
+		if err := tapsend.PrepareOutputAssets(ctx, vPkt); err != nil {
+			log.Errorf("Error preparing output assets: %v, "+
+				"packets: %v", err, limitSpewer.Sdump(vPkt))
+			return nil, fmt.Errorf("unable to prepare outputs: %w",
```
GeorgeTsagk, line 691: missing newline before `return`

taprpc/taprootassets.proto
```diff
@@ -1717,14 +1717,17 @@ message BurnAssetRequest {
...
 }
 
 message BurnAssetResponse {
     // The asset transfer that contains the asset burn as an output.
     AssetTransfer burn_transfer = 1;
 
-    // The burn transition proof for the asset burn output.
-    DecodedProof burn_proof = 2;
+    // The burn transition proofs for the asset burn output.
+    repeated DecodedProof burn_proofs = 2;
```
GeorgeTsagk: Are we sure this is backwards-compatible on the wire?

rpcserver.go
GeorgeTsagk: Yes, you're right.

GeorgeTsagk:
We'll have to rebase `0-8-0-staging` on `main` in order to get the `LiT tests` working again (there were some version bumps)

darioAnongba:
Oops @GeorgeTsagk, deleted a comment but wasn't on purpose. Here is the response:
> Yes, I found them particularly annoying and they didn't help since in the tests I can either log anything I want to debug or apply an assertion as part of the test. Especially when we are dumping the entire response multiple times in the test.

taprpc/taprootassets.proto
darioAnongba: good catch, I reserved field 2 and used 3 instead as discussed.

rpcserver.go
```diff
@@ -3729,36 +3711,120 @@ func (r *rpcServer) BurnAsset(ctx context.Context,
...
+			groupKey = &assetGroup.GroupPubKey
+
+		case errors.Is(err, address.ErrAssetGroupUnknown):
+			rpcsLog.Trace("Asset group key not found, asset " +
+				"may not be part of a group")
+
+		case err != nil:
+			return nil, fmt.Errorf("error querying asset "+
+				"group: %w", err)
+		}
```
ffranr, line 3789: I thought the intended behaviour was:

* If only an asset ID is specified, burn assets matching that asset ID.
* If only an asset group is specified, burn across any tranche within that group.
* If both an asset ID and an asset group are specified, burn assets matching the asset ID, but verify that the asset belongs to the specified group.

This approach allows burning assets across a group when the caller doesn’t care about the tranche, while preserving the option to target specific tranches (asset IDs) when needed.

Perhaps something like this:
https://gist.github.com/ffranr/9b94850c4fc828a1a3e5170b1d1af199

taprpc/taprootassets.proto
```diff
@@ -1707,11 +1707,11 @@ message FetchAssetMetaResponse {
...
-        // The asset ID of the asset to burn units of.
-        bytes asset_id = 1;
+        // Deprecated, the asset ID of the asset to burn units of.
+        bytes asset_id = 1 [deprecated = true];
 
-        // The hex encoded asset ID of the asset to burn units of.
-        string asset_id_str = 2;
+        // Deprecated, the hex encoded asset ID of the asset to burn units of.
+        string asset_id_str = 2 [deprecated = true];
     }
```
ffranr, line 1717: Doc comment could direct towards the new `asset_specifier` field.

cmd/commands/assets.go
```diff
@@ -1127,42 +1167,100 @@ func burnAssets(ctx *cli.Context) error {
...
-			return fmt.Errorf("unable to list current asset "+
-				"balances: %w", err)
+
+		// When both asset ID and group key are provided,
+		// prioritize asset ID for balance checking since
+		// it's more specific.
+		switch {
+		case len(assetSpecifier.GetId()) > 0:
+			assetSpecBytes = assetSpecifier.GetId()
+			assetSpecType = "Asset ID"
```
ffranr: One possible improvement:

Instead of returning an RPC type from `buildAssetSpecifier`, have it return an `asset.Specifier`. This would make `assetSpecType string` unnecessary, since `asset.Specifier` could be used directly in `fmt.Sprintf("Please confirm destructive...")`. A separate method could then handle marshalling `asset.Specifier` into the RPC message type.

ffranr: Then instead of `len(assetSpecifier.GetGroupKey()) > 0` on an RPC message type we could do `assetSpec.HasGroupPubKey()`, etc.

rpcserver.go
ffranr: > // 1. If only assetID is provided, look up the group key.

This part concerns me slightly. Do we assert in the integration test that when a group collectable is burned by asset ID, only that specific asset ID is burned and not another collectable from the same group? Why do we need to populate the group key when only an asset ID is specified?

darioAnongba: > - If only an asset ID is specified, burn assets matching that asset ID.
> - If only an asset group is specified, burn across any tranche within that group.
> - If both an asset ID and an asset group are specified, burn assets matching the asset ID, but verify that the asset belongs to the specified group.

This is what the code does. Did you find a bug in the current logic that somehow doesn't execute this flow?

> Why do we need to populate the group key when only an asset ID is specified

This branche of the PR is the one that does not change. As you can see, the previous code already populates the group key when only the `assetID` is specified. My goal was specifically not to alter the previous behavior. I can investigate the reason why the group key is needed but that should not matter for this PR unless we want to change the previous logic.

PR #1813, opened 2025-09-23
AddrReceives pagination and sorting direction

### LINKED ISSUE #728: [feature]: pagination for AddrReceives
Hello,

Is some type of pagination going to be added to the [AddrReceives](https://lightning.engineering/api-docs/api/taproot-assets/taproot-assets/addr-receives) REST API endpoint?

Maybe a minimum block height parameter, or something similar?

I am aware that it is possible to query for a specific address.

But it would be nice if we could also query for recent receives in a way that doesn't return all the receives that ever happened, in one big response.

darioAnongba:
This PR enables pagination for `AddrReceives` endpoint.

Fixes #728


rpcserver.go
```diff
@@ -2224,6 +2224,28 @@ func (r *rpcServer) AddrReceives(ctx context.Context,
 
 	var sqlQuery address.EventQueryParams
 
+	if req.Offset < 0 {
+		return nil, fmt.Errorf("offset must be non-negative")
+	}
+	if req.Limit < 0 {
```
jtobin, line 2230: Should there be a maximum limit enforced here as well? E.g. the universe code checks that the limit doesn't exceed a max page size (L6481):

```
	if req.Limit > universe.MaxPageSize || req.Limit < 0 {
		return nil, fmt.Errorf("invalid request limit: %d", req.Limit)
	}
```

taprpc/taprootassets.proto
```diff
@@ -1513,6 +1513,15 @@ message AddrReceivesRequest {
...
     AddrEventStatus filter_status = 2;
+
+    // The number of events to skip.
+    int32 offset = 3;
+
+    // The max number of events returned.
+    int32 limit = 4;
+
+    // The direction of the page. 0 for ascending, 1 for descending.
+    int32 direction = 5;
```
jtobin: It looks like universerpc/universe.proto defines/uses a SortDirection enum for the same purpose. Given that file imports taprpc/taprootassets.proto anyway, I wonder would it be better to move the SortDirection definition to taprootassets.proto and use it in place of 'int32' here?

jtobin (commented):
Looking good! Just a couple of questions around instituting a max pagination limit & some minor refactoring.

rpcserver.go
darioAnongba: applied

taprpc/taprootassets.proto
ffranr: enum would be better IMO. Even if it has to be a new enum type.

Also, I think in the doc comment we should make it clear on what properties we're sorting, I guess timestamp?

.gitignore
```diff
@@ -55,3 +55,6 @@ vendor/
 # Release builds
 /taproot-assets-*
 .aider*
+
+# Environment variables
+.env
```
ffranr: this seems out of place in this commit. please drop this change from the pr?

rpcserver.go
```diff
@@ -2224,6 +2224,28 @@ func (r *rpcServer) AddrReceives(ctx context.Context,
...
+	case 0:
+		sqlQuery.SortDirection = address.SortAscending
+
+	case 1:
+		sqlQuery.SortDirection = address.SortDescending
+
+	default:
+		return nil, fmt.Errorf("invalid sort direction: %v",
+			req.Direction)
+	}
```
ffranr: if we had a proto message type to re-use we could have functions for this sort of unmarshalling

address/event.go
```diff
@@ -41,6 +41,24 @@ const (
...
+const (
+	// SortAscending is a sentinel value that indicates that the sort
+	// should be in ascending order.
+	SortAscending SortDirection = iota
+
+	// SortDescending is a sentinel value that indicates that the sort
+	// should be in descending order.
+	SortDescending
+)
+const (
```
ffranr, line 61: formatting, add space here:
```
)
const (
```

address/event.go
```diff
@@ -41,6 +41,24 @@ const (
...
+
+const (
+	// SortAscending is a sentinel value that indicates that the sort
+	// should be in ascending order.
+	SortAscending SortDirection = iota
+
+	// SortDescending is a sentinel value that indicates that the sort
+	// should be in descending order.
+	SortDescending
+)
```
ffranr, line 59: Might be worth adding `UndefinedSortDirection` as 0, which leaves the door open for us not to have to sort if the user doesn't care about it (if they skip pagination or in some other sort context). Plus it's usually useful for when a func returns an error to have an "undefined" enum value.

I think you should name these fields `AscSortDirection` and `DescSortDirection` to tie them closer to the enum type.

tapdb/addrs_test.go
```diff
@@ -935,6 +935,97 @@ func TestQueryAddrEvents(t *testing.T) {
 	require.Equal(t, event.Outpoint, events[0].Outpoint)
 }
 
+// TestQueryAddrEventsPagination tests that query parameters for pagination are
+// applied when listing address events.
+func TestQueryAddrEventsPagination(t *testing.T) {
+	t.Parallel()
```
ffranr, line 941: this is fine IMO, and I approve the test as is. But I think that a table based unit test would be a clean fit here.

cmd/commands/addrs.go
```diff
@@ -283,6 +286,18 @@ var receivesAddrCommand = cli.Command{
...
+			Usage: "the max number of events returned",
+		},
+		cli.Int64Flag{
+			Name:  offsetName,
+			Usage: "the number of events to skip",
+		},
+		cli.StringFlag{
+			Name:  directionName,
+			Usage: "the sort direction for events (asc or desc)",
+		},
```
ffranr, line 311: default to descending? so we get latest events first, have i understood that correctly. Maybe we should mention default here also.

itest/addrs_test.go
```diff
@@ -701,6 +701,165 @@ func runMultiSendTest(ctxt context.Context, t *harnessTest, alice,
 	}
 }
 
+// testAddrReceivesPagination tests the pagination of
+// AddrReceives
+func testAddrReceivesPagination(t *harnessTest) {
```
ffranr: i don't think RPC endpoint pagination warrants a new itest, can we add coverage by extending an existing itest?

address/event.go
```diff
@@ -57,6 +57,10 @@ const (
 	// DefaultEventQueryLimit is the number of events returned
 	// when no limit is provided.
 	DefaultEventQueryLimit = 512
+
+	// MaxEventQueryLimit is the maximum number of events that can be
+	// returned in a single query.
+	MaxEventQueryLimit = 16384
```
ffranr, line 68: i can see this is used in the `universe` package also. Do we know why `16384`? Would be good to add to the doc comment if we do.

taprpc/taprootassets.proto
darioAnongba: applied by moving `SortDirection` to `tapcommon` and using it in `universe` and `taproot-assets`

address/event.go
darioAnongba: Applied. Note that this is not used anywhere for now as the default value is set to descending in RPC and CLI.

tapdb/addrs_test.go
darioAnongba: Do you have any example of a table based unit test? Not sure what you mean exactly.

address/event.go
darioAnongba: I copied the value from universe but I'm not sure what was the reasoning behind this specific number. Ideally the limit should be different per event depending on a max size in bytes.

itest/addrs_test.go
darioAnongba: I try adding it to an existing test but I couldn't find any for the fetching of address events. I removed this test being only for "pagination" and made it generic as `testAddrReceives` for all future tests involving fetching address events.
Happy to move it into an existing test if there is one.

tapdb/addrs_test.go
ffranr: For future reference, test cases approach: https://github.com/lightninglabs/taproot-assets/blob/189e165af084006fd353cf9952de467b364e78f6/rfqmsg/accept_test.go#L67-L88

taprpc/tapcommon.proto
```diff
@@ -34,3 +34,9 @@ message AssetOutPoint {
     // asset is locked to.
     bytes script_key = 3;
 }
+
+enum SortDirection {
+    SORT_DIRECTION_ASC = 0;
+
+    SORT_DIRECTION_DESC = 1;
+}
```
ffranr, line 44: To make this more generally applicable, we could add an `UNDEFINED` sort direction at value 0. This would let us use the enum in cases where sorting is optional and no ordering is guaranteed on the results (though some ordering may still occur).

On the `universerpc` side of things I guess this is a breaking change? So maybe `universerpc` should stick with its old version, otherwise it might break syncing.

Also, we should add general comments for each enum value.

itest/addrs_test.go
```diff
@@ -701,6 +701,164 @@ func runMultiSendTest(ctxt context.Context, t *harnessTest, alice,
 	}
 }
 
+// testAddrReceives tests the fetching of address events.
+func testAddrReceives(t *harnessTest) {
```
ffranr: :+1: i prefer this more general itest that we can expand on, good idea IMO

itest/addrs_test.go
```diff
@@ -701,6 +701,164 @@ func runMultiSendTest(ctxt context.Context, t *harnessTest, alice,
...
+			t.t, secondTapd, &taprpc.NewAddrRequest{
+				AssetId:      asset.AssetGenesis.AssetId,
+				Amt:          uint64(10),
+				AssetVersion: asset.Version,
+			},
+		)
+
+		AssertAddrCreated(t.t, secondTapd, asset, addr)
+
+		// Send assets to the address
```
ffranr: Comments should end with a period. I think that’s in the style guide (and if not, it should be). This matches what we do elsewhere in the code.

tapdb/addrs_test.go
darioAnongba: oh I see! I thought you meant "table" as in "DB table". I agree, I can change this and make it cleaner.

darioAnongba: Applied

taprpc/tapcommon.proto
darioAnongba: I agreed on the addition of `Undefined` in the internal `SortDirection` type because it made sense there as meaning "unspecified" or "unknown", and could be used to let the query execute without the `ORDER BY` clause by letting the DB apply its default for example.

But `UNDEFINED` has no meaning at the API level, that is user facing and represents all the sorting the user can choose, only `asc` and `desc` as of now. We should not mix Go's 0-value and the proto 0 index IMO, which is just the first index used for encoding, it does not mean "empty".
If we wanted a new sorting direction like `SORT_DIRECTION_RANDOM`, we would set it at index `2`, not `0` (non breaking change).

darioAnongba: discussed and went for the `optional` solution instead.

universe_rpc_diff.go
```diff
@@ -85,7 +86,7 @@ func (r *RpcUniverseDiff) RootNodes(ctx context.Context,
 			WithAmountsById: q.WithAmountsById,
 			Offset:          q.Offset,
 			Limit:           q.Limit,
-			Direction:       unirpc.SortDirection(q.SortDirection),
+			Direction:       taprpc.SortDirection(q.SortDirection),
```
ffranr, line 89: I’ve checked this and it’s not breaking. If it were, we’d drop the change. Current diff is fine IMO

rpcserver.go
```diff
@@ -2224,6 +2224,25 @@ func (r *rpcServer) AddrReceives(ctx context.Context,
...
+			address.MaxEventQueryLimit)
+	}
+
+	sqlQuery.Offset = req.Offset
+	sqlQuery.Limit = req.Limit
+
+	sqlQuery.SortDirection = address.AscSortDirection
+	if req.GetDirection() == taprpc.SortDirection_SORT_DIRECTION_DESC {
+		sqlQuery.SortDirection = address.DescSortDirection
+	}
```
ffranr: i think we should still set the default sort direction to desc here such that the latest events turn up first

darioAnongba: yes, this piece of code effectively has no effect on the default but I agree it's not clear. I changed it to a more explicit `switch`.

ffranr:
Do we merge this into `0-8-0-staging` given the milestone? If so, we need to change the base branch.

darioAnongba:
It was decided to merge this into 0.7 as far as I understood.

Issue #1811, opened 2025-09-18
Supply Commitment Verification: Potential Race Condition with Universe Syncer


ffranr:
#### The Problem: Potential Race Condition Between Supply Verifier State Machine and Universe Syncer

A potential critical race condition may exist within a peer node between its verifier and its timer-based universe issuance syncer. This could lead to failed supply commit verifications.

Here's a breakdown of the two interacting processes:

1. **Universe Issuance Syncer**: This component operates on a timer, processing issuance proofs for assets that support supply commitments. It populates the local database with pre-commitment outputs, which are expected to be spent by a subsequent supply commitment transaction.

2. **Supply Verifier**: This state machine actively discovers new supply commitments by watching on-chain events and fetching data from a universe server.

This potential race condition would occur if the verifier fetches a new supply commitment before the syncer has processed the corresponding issuance proof. This would result in the verifier attempting to validate a transaction that spends a pre-commitment output which does not yet exist in its local database, causing the verification to fail.

#### Consequence: Potential for Inconsistent Database State

This timing issue could also introduce a secondary problem that can corrupt the database state. If an issuance proof arrives late and the syncer adds the pre-commitment output to the database after the corresponding supply commitment transaction has already been processed on-chain, the output may never be marked as spent. This would leave a dangling record in the database—a pre-commitment output that is erroneously permanently marked as unspent.

#### Proposed Solution: Bundle Issuance Pre-Commit Data in `FetchSupplyCommit` RPC Response

To eliminate this dependency on the syncer's timing, we should fetch the issuance pre-commitment outputs atomically with the supply commitment itself. The `FetchSupplyCommit` RPC endpoint should be modified to return not only the supply commitment but also the associated issuance proofs and their corresponding pre-commitment outputs. This change ensures that the verifier has all the necessary components at the moment of verification, completely removing the race condition.

#### Additional Benefits: Stronger Verification Logic

This approach also presents an opportunity to improve our verification logic. With all the data available at once, the verifier can perform a more robust check: it can explicitly ensure that every issuance leaf within the supply commitment has a corresponding pre-commitment output available for spending. This creates a stronger, self-contained verification process that doesn't rely on external state that may or may not be up-to-date.

Roasbeef:
> This potential race condition would occur if the verifier fetches a new supply commitment before the syncer has processed the corresponding issuance proof. 

[...]

PR #1820, opened 2025-09-25
Extract supply pre-commits from mint leaves during supply verifier sync

ffranr:
Closes: https://github.com/lightninglabs/taproot-assets/issues/1811

This PR adds functionality to extract supply pre-commits directly from the mint leaves retrieved by the supply verifier during a sync pull of a supply commitment.

Previously, the supply verifier relied on the universe issuance syncer to obtain the full set of pre-commitments. This change removes that dependency, reducing the risk of a race condition where supply commitments were synced before the associated pre-commitments were available.

### DIFF
--- a/itest/supply_commit_test.go
+++ b/itest/supply_commit_test.go
@@ -1042,6 +1042,9 @@ func testSupplyCommitMintBurn(t *harnessTest) {
 //  5. Ignores the asset outpoint sent to the secondary node.
 //  6. Publishes the second supply commitment and mines it.
 //  7. Verifies the secondary node can fetch the updated supply commitment.
+//  8. Primary node mints another asset into the group and publishes the
+//     third supply commitment.
+//  9. Verifies the secondary node can fetch the third supply commitment.
 func testSupplyVerifyPeerNode(t *harnessTest) {
 	ctxb := context.Background()
 
@@ -1049,11 +1052,11 @@ func testSupplyVerifyPeerNode(t *harnessTest) {
 		"commitments enabled")
 
 	// Create a mint request for a grouped asset with supply commitments.
-	mintReq := CopyRequest(issuableAssets[0])
-	mintReq.Asset.Amount = 5000
+	firstMintReq := CopyRequest(issuableAssets[0])
+	firstMintReq.Asset.Amount = 5000
 
 	rpcFirstAsset, _ := MintAssetWithSupplyCommit(
-		t, mintReq, fn.None[btcec.PublicKey](),
+		t, firstMintReq, fn.None[btcec.PublicKey](),
 	)
 
 	// Parse out the group key from the minted asset.
@@ -1077,7 +1080,7 @@ func testSupplyVerifyPeerNode(t *harnessTest) {
 	// Verify the issuance subtree root exists and has the correct amount.
 	require.NotNil(t.t, fetchResp.IssuanceSubtreeRoot)
 	require.Equal(
-		t.t, int64(mintReq.Asset.Amount),
+		t.t, int64(firstMintReq.Asset.Amount),
 		fetchResp.IssuanceSubtreeRoot.RootNode.RootSum,
 	)
 
@@ -1125,7 +1128,7 @@ func testSupplyVerifyPeerNode(t *harnessTest) {
 	)
 
 	require.Equal(
-		t.t, int64(mintReq.Asset.Amount),
+		t.t, int64(firstMintReq.Asset.Amount),
 		peerFetchResp.IssuanceSubtreeRoot.RootNode.RootSum,
 	)
 
@@ -1160,7 +1163,7 @@ func testSupplyVerifyPeerNode(t *harnessTest) {
 
 	t.Log("Verifying retrieval of second supply commitment from primary " +
 		"node")
-	fetchResp, _ = WaitForSupplyCommit(
+	fetchResp, supplyOutpoint = WaitForSupplyCommit(
 		t.t, ctxb, t.tapd, groupKeyBytes, fn.Some(supplyOutpoint),
 		func(resp *unirpc.FetchSupplyCommitResponse) bool {
 			ignoreRoot := resp.IgnoreSubtreeRoot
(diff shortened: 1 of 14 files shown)


universe/supplyverifier/util.go
```diff
@@ -0,0 +1,127 @@
...
+	// height. These can't be spent by the current supply commitment.
+	filteredPreCommits := make(
+		[]supplycommit.PreCommitment, 0, len(preCommits),
+	)
+	for _, pc := range preCommits {
+		if pc.BlockHeight <= supplyCommitBlockHeight {
+			filteredPreCommits = append(filteredPreCommits, pc)
+		}
+	}
+	preCommits = filteredPreCommits
```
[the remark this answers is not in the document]
ffranr: i think my code is simpler and pre-mature optimisation is not wise at the cost of clarity. plus my code is easier to debug with break points.

universe/supplyverifier/util.go
```diff
@@ -33,3 +34,94 @@ func FetchDelegationKey(ctx context.Context,
...
+	// group.
+	preCommits, err := supplyCommitView.UnspentPrecommits(
+		ctx, assetSpec, false,
+	).Unpack()
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch unspent "+
+			"pre-commitments: %w", err)
+	}
+
+	// Filter out any pre-commitments that are above the supply commitment
```
Roasbeef, line 67: 👍

universe/supplyverifier/util.go
```diff
@@ -33,3 +34,94 @@ func FetchDelegationKey(ctx context.Context,
 
 	return delegationKey, nil
 }
+
+// FetchPreCommits returns all pre-commitment outputs expected to be spent by
+// the anchoring transaction of the specified supply commitment.
+func FetchPreCommits(ctx context.Context,
```
Roasbeef, line 40: We should add a unit test for this.

tapdb/supply_commit.go
```diff
@@ -1150,6 +1150,12 @@ func (s *SupplyCommitMachine) InsertSupplyCommit(ctx context.Context,
 					"pre-commit outpoint: %w", err)
 			}
 
+			err = upsertSupplyPreCommit(ctx, db, preCommit)
```
Roasbeef, line 1154: 👍

Roasbeef: Should add a test for this.

itest/supply_commit_test.go
```diff
@@ -1257,4 +1260,109 @@ func testSupplyVerifyPeerNode(t *harnessTest) {
 			peerFetchResp2.ChainData.SupplyRootHash,
 		),
 	)
+
+	// Step 8: Primary node mints another asset into the group and publishes
```
Roasbeef, line 1272: Does this test fail without the current diff? 

Also we can move this into a new heler assertion, basically a sub-test, as this itest is already rather long.

Roasbeef: Also similar comment from the other PR: there're many duplicated assertions we can put into helper functions.

Roasbeef (changes requested):

itest/supply_commit_test.go
ffranr: The following PR includes helper methods that I could use in this PR to improve the itest changes: 
https://github.com/lightninglabs/taproot-assets/pull/1823

Would be simpler if we merge that PR first.

Roasbeef: That one is merged in now, so this can be rebased.

universe/supplyverifier/util_test.go
```diff
@@ -0,0 +1,798 @@
...
+	t.Helper()
+
+	// Create a dummy script key.
+	privKey, err := btcec.NewPrivateKey()
+	if err != nil {
+		t.Fatalf("failed to create private key: %v", err)
+	}
+	scriptKey := privKey.PubKey()
+
+	// Create a dummy outpoint.
```
Roasbeef: Superfluous comments here and elsewhere in this file (doesn't add any context re the why/how).

universe/supplyverifier/util_test.go
```diff
@@ -0,0 +1,798 @@
...
+	return supplycommit.NewMintEvent{
+		LeafKey:       leafKey,
+		IssuanceProof: issuanceProof,
+		MintHeight:    blockHeight,
+	}
+}
+
+// TestFetchPreCommits tests the FetchPreCommits function with various
+// scenarios.
+func TestFetchPreCommits(t *testing.T) {
```
Roasbeef, line 348: 👍

tapdb/universe_test.go
```diff
@@ -1097,3 +1099,213 @@ func TestMultiverseRootSum(t *testing.T) {
...
+
+	delegationKey := test.RandPubKey(t)
+
+	testCases := []struct {
+		name        string
+		proofType   universe.ProofType
+		asset       *asset.Asset
+		metaReveal  *proof.MetaReveal
+		expected    bool
+		description string
```
Roasbeef, line 1125: Nice addition here with the desc to show a better error message if the assertion fails!

universe/supplyverifier/util_test.go
ffranr: removed, hope i didn't miss any.

itest/supply_commit_test.go
ffranr: > Does this test fail without the current diff?

In the latest commit set i extend the supply commit verifier with:
```
        // Verify that we have at least as many supply pre-commitments as
	// new issuance leaves. Each issuance leaf must correspond to a
	// pre-commitment output created at the time of asset issuance.
	if len(unspentPreCommits) < len(leaves.IssuanceLeafEntries) {
		return fmt.Errorf("not enough unspent supply pre-commitment "+
			"outputs for issuance leaves: have %d, need %d",
			len(unspentPreCommits), len(leaves.IssuanceLeafEntries))
	}
```

With that verifier change only, the extended itest in this PR should fail. I have tested this manually. The failure is possibly flaky rather than deterministic since it depends on the sync tick cadence and fetch retries. But i bet the extended itest fails most of the time under these conditions.

Roasbeef (approved):
LGTM 🧬

PR #1823, opened 2025-09-25
Add support for fetching latest supply commitment via new locator type

ffranr:
This PR introduces support for fetching the latest supply commitment using a new locator type, `LocatorTypeLatest`. Key changes include:

* **`supplyverifier`**: Adds `LocatorTypeLatest` and updates `Manager.FetchCommitment` to handle it.
* **`universerpc`**: Extends the `FetchSupplyCommit` RPC endpoint to support the new locator type.
* **`tapdb`**: Fixes a bug where group keys were incorrectly serialized with `SerializeCompressed()` instead of `schnorr.SerializePubKey`.
* **`itest`**: Extends an integration test for querying the latest supply commitment from the universe server.

This enables clients to request the most recent supply commitment without needing to know specific asset state locators.


universe/supplyverifier/manager.go
```diff
@@ -763,6 +767,13 @@ func (m *Manager) FetchCommitment(ctx context.Context,
 				"commitment: %w", err)
 		}
 
+	case LocatorTypeLatest:
```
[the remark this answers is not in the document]
ffranr: yes, but, also, no trivial comments?

jamaljsr:
Functional tACK 👌

There still isn't any way to get the latest commitment via the CLI. Should the CLI return the latest instead of the first by default? I believe users would be most interested in the current total supply, rather than the supply at some unknown time in the past.

jtobin (approved):
This LGTM. :+1: Re: Jamal's comment, the following patch should tweak the CLI appropriately, if you want to apply it.

<details>
<summary>(patch)</summary>

```
From eb1781243e7bb51ba3ca4adf5639d93a8f043c94 Mon Sep 17 00:00:00 2001
From: Jared Tobin <jared@jtobin.io>
Date: Fri, 26 Sep 2025 12:17:16 +0400
Subject: [PATCH] cmd: add --first locator option, default to latest

Defaults to use of the Latest locator for supply commit fetching, and
adds a --first flag to specify use of the VeryFirst locator.
---
 cmd/commands/universe.go | 39 ++++++++++++++++++++++++++++++---------
 1 file changed, 30 insertions(+), 9 deletions(-)

diff --git a/cmd/commands/universe.go b/cmd/commands/universe.go
index 58db6280..7ba7d3ae 100644
--- a/cmd/commands/universe.go
+++ b/cmd/commands/universe.go
@@ -1634,11 +1634,15 @@ var fetchSupplyCommitCmd = cli.Command{
 	Description: `
 	Fetch the on-chain supply commitment for a specific asset group.
 
-	At most one of --outpoint and --spent_outpoint may be provided. If
-	neither is provided, the very first supply commitment for the asset
-	group will be fetched.
+	At most one of --first, --outpoint, and --spent_outpoint may be
+	provided. If none are provided, the latest supply commitment for the
+	asset group will be fetched.
 	`,
 	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:  "first",
+			Usage: "fetch the very first supply commitment",
+		},
 		&cli.StringFlag{
 			Name:     "group_key",
 			Usage:    "the group key of the asset group to fetch",
@@ -1665,13 +1669,25 @@ func fetchSupplyCommit(ctx *cli.Context) error {
 	client, cleanUp := getUniverseClient(ctx)
 	defer cleanUp()
 
+	firstSet := ctx.IsSet("first")
 	outpointSet := ctx.IsSet("outpoint")
 	spentOutpointSet := ctx.IsSet("spent_outpoint")
 
-	if outpointSet && spentOutpointSet {
-		return fmt.Errorf(
-			"only one of --outpoint or --spent_outpoint can be set",
-		)
+	// Check that at most one of the three flags is set.
+	flagsSet := 0
+	if outpointSet {
+		flagsSet++
+	}
+	if spentOutpointSet {
+		flagsSet++
+	}
+	if firstSet {
+		flagsSet++
+	}
+
+	if flagsSet > 1 {
+		return fmt.Errorf("only one of --first, --outpoint, or " +
+			"--spent_outpoint can be set")
 	}
 
 	req := &unirpc.FetchSupplyCommitRequest{
@@ -1681,6 +1697,11 @@ func fetchSupplyCommit(ctx *cli.Context) error {
 	}
 
 	switch {
+	case firstSet:
+		req.Locator = &unirpc.FetchSupplyCommitRequest_VeryFirst{
+			VeryFirst: true,
+		}
+
 	case outpointSet:
 		arg := ctx.String("outpoint")
 		outpoint, err := wire.NewOutPointFromString(arg)
@@ -1710,8 +1731,8 @@ func fetchSupplyCommit(ctx *cli.Context) error {
 			}
 
 	default:
-		req.Locator = &unirpc.FetchSupplyCommitRequest_VeryFirst{
-			VeryFirst: true,
+		req.Locator = &unirpc.FetchSupplyCommitRequest_Latest{
+			Latest: true,
 		}
 	}
 
-- 
2.50.1
```

</details>

ffranr:
Thank you both for taking a look! And thanks for the patch @jtobin , I've included in this PR.

tapdb/supply_commit.go
```diff
@@ -1461,7 +1461,7 @@ func (s *SupplyCommitMachine) FetchLatestCommitment(ctx context.Context,
 
 	var (
 		writeTx       = WriteTxOption()
-		groupKeyBytes = groupKey.SerializeCompressed()
+		groupKeyBytes = schnorr.SerializePubKey(groupKey)
```
Roasbeef, line 1465: We should write a unit test that fails without this change.

itest/supply_commit_test.go
```diff
@@ -1199,6 +1199,38 @@ func testSupplyVerifyPeerNode(t *harnessTest) {
 		),
 	)
 
+	// If we query for the latest supply commitment from the universe server
```
Roasbeef, line 1223: This itest is already rather long. I think we can improve the readability by moving this new assertion/logic into a helper assertion function. 

We can likely also spin out those assertions at the end into a helper function, as the same assertions are repeated many times above.

Roasbeef (changes requested):
Solid change, the locator abstraction on the RPC API turns out to be very useful indeed.

itest/supply_commit_test.go
ffranr: The latest commit set includes new test helper functions that have saved a few lines.

Roasbeef:
`    --- FAIL: TestTaprootAssetsDaemon/tranche01/13-of-96/address_receives (15.46s)
` 

Is failing, but looks like we have a follow up PR to resolve that.

PR #1825, opened 2025-09-26
WIP: universerpc: add `all_leaves` flag to `FetchSupplyCommit` RPC

ffranr:
Introduce a new `all_leaves` flag in `FetchSupplyCommitRequest` to control whether the RPC should return only the incremental set of leaves added by the fetched commitment (default), or the full set of leaves up to and including that commitment.

---

#### TODO

* itest coverage
* change log
* cli support.


ffranr:
Will make `FetchSupplyLeaves` public instead.

PR #1828, opened 2025-09-29
itest: fix address receives test: timestamp in the future

darioAnongba:
Return early when the `StartTimestamp` is set in the future, so no events can be returned.

### DIFF
--- a/tapdb/addrs.go
+++ b/tapdb/addrs.go
@@ -989,12 +989,14 @@ func (t *TapAddressBook) QueryAddrEvents(
 	ctx context.Context, params address.EventQueryParams) ([]*address.Event,
 	error) {
 
+	now := time.Now().UTC()
 	sqlQuery := AddrEventQuery{
 		StatusFrom:    int16(address.StatusTransactionDetected),
 		StatusTo:      int16(address.StatusCompleted),
 		CreatedAfter:  time.Unix(0, 0).UTC(),
-		CreatedBefore: time.Now().UTC(),
+		CreatedBefore: now,
 	}
+
 	if len(params.AddrTaprootOutputKey) > 0 {
 		sqlQuery.AddrTaprootKey = params.AddrTaprootOutputKey
 	}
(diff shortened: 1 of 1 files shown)


itest/addrs_test.go
```diff
@@ -970,16 +970,15 @@ func testAddrReceives(t *harnessTest) {
...
-	// Test 5: Filter by address and start timestamp after send
-	// (should return no events).
-	resp, err = bob.AddrReceives(
+	// Test 5: Should error when start timestamp is in the future.
+	const expectedError = "created after time after created before time"
+	_, err = bob.AddrReceives(
 		ctxt, &taprpc.AddrReceivesRequest{
 			FilterAddr:     addr.Encoded,
-			StartTimestamp: uint64(timeAfterSend.Unix() + 1),
+			StartTimestamp: uint64(time.Now().Add(time.Minute).Unix()),
```
ffranr: hmm not sure about this fix. I think we need to get here in less than a minute to avoid the same error:
```
func (t *TapAddressBook) QueryAddrEvents(
	ctx context.Context, params address.EventQueryParams) ([]*address.Event,
	error) {

	sqlQuery := AddrEventQuery{
		StatusFrom:    int16(address.StatusTransactionDetected),
		StatusTo:      int16(address.StatusCompleted),
		CreatedAfter:  time.Unix(0, 0).UTC(),
		CreatedBefore: time.Now().UTC(),
	}
```

maybe we need to modify `TapAddressBook.QueryAddrEvents` such that it returns no events if `params.CreationTimeFrom` is in the future

darioAnongba: makes sense, changes applied as per discussion. Test is then fine and needs no changes.

tapdb/addrs.go
```diff
@@ -989,12 +989,14 @@ func (t *TapAddressBook) QueryAddrEvents(
 	ctx context.Context, params address.EventQueryParams) ([]*address.Event,
 	error) {
 
+	now := time.Now()
```
ffranr: `now := time.Now().UTC()`?

darioAnongba: true! 🙏

jtobin (approved):
LGTM. :+1:

PR #1830, opened 2025-10-01
docs: add docs re current stxo design

Roasbeef:


GeorgeTsagk (approved):
Read through it, LGTM

I understand that this has been generated with the aid of an agent -- perhaps we could distill it and make it a bit less verbose. Non-blocking though, just personal preference.

ffranr (commented):
I've scanned through this. It mostly looks ok I think.

I'm not sure if the codebase actually benefits from this sort of doc. It was generated with an agent + code + spec? In which case it seems like something that an agent could figure out on the fly. And the doc can go stale etc.

darioAnongba (approved):

jtobin:
I think this needs a bump to satisfy CI, or @Roasbeef feel free to just bypass that and merge.

litbot-9000:

PR #1832, opened 2025-10-01
garbage collect zero-value UTXOs

### LINKED ISSUE #514: wallet: garbage collect tombstone UTXOs

darioAnongba:
Garbage collect the residue orphaned UTXOs when creating transactions. Orphaned UTXOs occur when creating tombstones or full burns.
Currently, these UTXOs accumulate in the DB and are never cleaned. This PR introduces a garbage collection mechanism to collect these UTXOs and use them as inputs of transactions initiated by `tapd`:
- Transfers
- Burns

The PR adds a new `swept` flag to the `managed_utxo` table because UTXOs are not removed from the table when spent.  This flag is also returned by the `ListUtxos` RPC endpoint.

The mechanism preserves the liveness and safety properties, ensuring that zero-value UTXOs can never accumulate in the DB. Adding garbage collection to Mint transactions is not necessary to ensure these properties.

Fixes #514 

**Note to reviewers**

1. UTXO meaning `unspent`, keeping the spent outputs in a table called `manage_utxos` seems like a contradiction, same logic applies for the need of a `swept` flag in that table. We should either rename the table or store spent utxos somewhere else?

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -32,6 +32,12 @@
 
 ## Functional Updates
 
+- [Garbage collection of orphaned UTXOs](https://github.com/lightninglabs/taproot-assets/pull/1832)
+  by sweeping tombstones and burn outputs when executing onchain transactions.
+  Garbage collection will be executed on every burn, transfer or call to
+  `AnchorVirtualPsbts`. A new configuration is available to control the sweeping 
+  via the flag `wallet.sweep-orphan-utxos`.
+
 ## RPC Updates
 
 - [PR#1841](https://github.com/lightninglabs/taproot-assets/pull/1841): Remove


tapdb/assets_store.go
```diff
@@ -1320,6 +1358,179 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+			err := readOutPoint(
+				bytes.NewReader(u.Outpoint), 0, 0, &anchorPoint)
+			if err != nil {
+				return err
+			}
+
+			// Skip UTXOs that have already been swept.
+			if u.Swept {
+				continue
+			}
```
ffranr: this should be checked first as it's cheaper than outpoint parsing.

tapdb/assets_store.go
```diff
@@ -1320,6 +1358,179 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+			// We include spent assets here because tombstones are
+			// marked as spent when created.
+			anchorPointBytes, err := encodeOutpoint(anchorPoint)
+			if err != nil {
+				return err
+			}
+
+			assetsAtAnchor, err := a.queryChainAssets(
+				ctx, q, QueryAssetFilters{
+					AnchorPoint: anchorPointBytes,
```
ffranr: encoding here is redundant since `anchorPointBytes` should be `u.Outpoint`

tapdb/assets_store.go
```diff
@@ -1320,6 +1358,179 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+					Now: sql.NullTime{
+						Time:  now,
+						Valid: true,
+					},
+				},
+			)
+			if err != nil {
+				return err
+			}
```
ffranr, line 1390: i think we should add error context message here: `return fmt.Errorf(`

tapdb/assets_store.go
```diff
@@ -1320,6 +1358,179 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+
+			// Skip If we don't have any assets recorded.
+			if len(assetsAtAnchor) == 0 {
+				continue
+			}
+
+			// Determine if all assets are tombstones or burns.
+			// A tombstone asset is marked as "spent" at the asset
+			// level but its anchor UTXO may still be unspent
+			// on-chain and available for sweeping.
```
ffranr, line 1399: :+1: nice comment, thanks

tapdb/assets_store.go
```diff
@@ -1320,6 +1358,179 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+					break
+				}
+			}
+
+			if !allZeroValue {
+				continue
+			}
+
+			log.Debugf("Anchor %v is zero-value, adding to "+
+				"sweep list", anchorPoint)
```
ffranr: it's better, IMO, to write log messages like this: 
```
log.Debugf("Adding zero-value anchor to sweep list (outpoint=%s)", anchorPoint.String())
```
because then it's easier to copy and paste search for a log line in code since the variable is a suffix and isn't in the middle of the search string.

tapdb/assets_store.go
```diff
@@ -1320,6 +1358,179 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+
+			// Extract the pkScript from the transaction.
+			var tx wire.MsgTx
+			if err := tx.Deserialize(
+				bytes.NewReader(chainTx.RawTx),
+			); err != nil {
+				log.Warnf("Failed to deserialize tx for "+
+					"%v: %v, skipping", anchorPoint, err)
+				continue
+			}
```
ffranr: I'm not sure if this formatting follows our style guide. I think we should just:
```
			err = tx.Deserialize(
				bytes.NewReader(chainTx.RawTx),
			)
			if err != nil {
				log.Warnf("Failed to deserialize tx for "+
					"%v: %v, skipping", anchorPoint, err)
				continue
			}
```

tapfreighter/chain_porter.go
```diff
@@ -1248,9 +1249,20 @@ func (p *ChainPorter) importLocalAddresses(ctx context.Context,
...
+		// We import outputs that:
+		// 1. Have local script keys (normal change outputs)
+		// 2. Tombstones (zero-value with NUMS key)
+		// 3. Burns (zero-value with burn key)
+		isTombstone := out.Amount == 0 &&
+			out.ScriptKey.PubKey.IsEqual(asset.NUMSPubKey)
+		isBurn := out.Amount == 0 && len(out.WitnessData) > 0 &&
+			asset.IsBurnKey(
+				out.ScriptKey.PubKey, out.WitnessData[0],
+			)
```
ffranr: I appreciate the comment here and the code style is nice.

But I think `IsTombstone` and `IsBurn` should be methods on `tapfreighter.TransferOutput` which would help with discoverability.

tapfreighter/chain_porter.go
```diff
@@ -1274,13 +1290,11 @@ func (p *ChainPorter) importLocalAddresses(ctx context.Context,
...
-			break
 
 		// On restart, we'll get an error that the output has already
 		// been added to the wallet, so we'll catch this now and move
 		// along if so.
 		case strings.Contains(err.Error(), "already exists"):
-			break
 
 		default:
 			return err
```
ffranr: I think it makes sense to remove the `break` here like you've done. But I think removing `case err == nil:` makes sense also. We might also want to take the opportunity to add error context to `return err`.

tapfreighter/fund_test.go
```diff
@@ -765,6 +765,7 @@ func TestFundPacket(t *testing.T) {
 			result, err := createFundedPacketWithInputs(
 				ctx, exporter, keyRing, addrBook,
 				tc.fundDesc, tc.vPkt, tc.selectedCommitments,
+				nil,
```
ffranr: we could add test coverage here just to make sure the field is populates in `result`

tapfreighter/interface.go
```diff
@@ -195,6 +199,31 @@ type CoinSelector interface {
...
 	ReleaseCoins(ctx context.Context, utxoOutpoints ...wire.OutPoint) error
+
+	// SelectZeroValueCoins selects all managed UTXOs that contain only
+	// zero-value assets (tombstones and burns). The selected UTXOs are
+	// leased for the default lease duration.
+	SelectZeroValueCoins(ctx context.Context) ([]ZeroValueInput, error)
+}
+
+// ZeroValueInput represents a zero-value UTXO that should be swept.
+type ZeroValueInput interface {
```
ffranr: is the interface necessary here rather than just a stuct? we might be able to drop all these getter methods for fields?

tapfreighter/wallet.go
```diff
@@ -1380,3 +1425,44 @@ func addAnchorPsbtInputs(btcPkt *psbt.Packet, vPackets []*tappsbt.VPacket) {
 
 	}
 }
+
+// addZeroValuePsbtInputs adds zero-value UTXOs as inputs to the PSBT.
+func addZeroValuePsbtInputs(btcPkt *psbt.Packet,
+	zeroValueInputs []ZeroValueInput, coinType uint32) {
```
ffranr: better to pass in `address.ChainParams` here rather than `coinType uint32` since the former is a better less error prone type

rpcserver.go
```diff
@@ -2647,8 +2648,17 @@ func (r *rpcServer) AnchorVirtualPsbts(ctx context.Context,
 			prevID.OutPoint.String())
 	}
 
+	// Fetch zero-value UTXOs that should be swept as additional inputs.
+	zeroValueInputs, err := r.cfg.AssetStore.FetchZeroValueAnchorUTXOs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch zero-value "+
+			"UTXOs: %w", err)
+	}
+
```
ffranr: this is probably fine as the default behaviour for now, even though it does have privacy implications. Maybe we should add a flag to `AnchorVirtualPsbtsRequest` to skip spending zero value inputs? default to: spend zero value

docs/release-notes/release-notes-0.7.0.md
```diff
@@ -267,6 +270,9 @@
 - Enable [burning the full amount of an asset](https://github.com/lightninglabs/taproot-assets/pull/1791)
   when it is the sole one anchored to a Bitcoin UTXO.
 
+- [Garbage collection of zero-value UTXOs](https://github.com/lightninglabs/taproot-assets/pull/1832)
+  by sweeping tombstones and burn outputs when executing onchain transactions.
+
```
ffranr: we should mention in release doc that the default behaviour of `AnchorVirtualPsbts` is to sweep

itest/burn_test.go
```diff
@@ -253,12 +253,6 @@ func testBurnAssets(t *harnessTest) {
 		simpleAsset.Amount-burnAmt-multiBurnAmt,
 	)
 
-	AssertBalances(
-		t.t, t.tapd, burnAmt+simpleCollectible.Amount+multiBurnAmt,
-		WithNumUtxos(3), WithNumAnchorUtxos(3),
-		WithScriptKeyType(asset.ScriptKeyBurn),
-	)
-
```
ffranr, line 261: why remove this assertion? (and below)

itest/zero_value_anchor_test.go
```diff
@@ -0,0 +1,196 @@
...
+	require.Len(t.t, utxosAfterSend.ManagedUtxos, 1)
+
+	tombstoneOp2 := ""
+	for outpoint, utxo := range utxosAfterSend.ManagedUtxos {
+		if !utxo.Swept {
+			tombstoneOp2 = outpoint
+			break
+		}
+	}
+	require.NotEmpty(t.t, tombstoneOp2)
```
ffranr: don't need this for loop since
`require.Len(t.t, utxosAfterSend.ManagedUtxos, 1)`
same below with burn

ffranr:
Please rebase on to the latest in `[0-8-0-staging](https://github.com/lightninglabs/taproot-assets/tree/0-8-0-staging)`. Might get you more PSBT funding test coverage now that https://github.com/lightninglabs/taproot-assets/pull/1418 is merged.

tapdb/assets_store.go
darioAnongba: Agree. Seems that what you are proposing is structured logging as this is more of a hack. I also agree structured logging is better. Might be worth a ticket.

tapfreighter/chain_porter.go
darioAnongba: Applied. Looks better now indeed.

tapfreighter/chain_porter.go
darioAnongba: Applied.

tapfreighter/fund_test.go
darioAnongba: Added unit tests

docs/release-notes/release-notes-0.7.0.md
darioAnongba: Also moved to `0-8-0.md`

itest/zero_value_anchor_test.go
darioAnongba: Good catch, this was a residue of the old way I first implemented it.

rpcserver.go
darioAnongba: I wasn't sure what to do with this endpoint either. Adding a flag bothers me because zero value utxos are an internal side effect of the protocol and the notion should never be exposed externally so I added the same behavior here as we would in burning or transfers.
Exposing the zero value utxos notion externally implies adding descriptions to the API, adding to the CLI and asking users to be aware of what they are.
Let's see what @GeorgeTsagk says.

itest/burn_test.go
darioAnongba: This assertion is not relevant anymore for burns because it sums the amounts of the assets in the UTXOs it finds. Here it's used to verify the amounts burnt but you can't do that anymore unless you keep track of the swept UTXOs which is not the scope of this test.
As an e2e test, you only care that after you burn, the balance of the asset is correctly decreased.

tapfreighter/interface.go
darioAnongba: Removed interface. I initially added it so a `ManagedUTXO` could "be" a `ZeroValueInput`, but better to map to specific struct.

tapdb/assets_store.go
darioAnongba: Just noticed that we already have structured logging so it's a matter of refactoring these logs eventually.

rpcserver.go
ffranr: > zero value utxos are an internal side effect of the protocol

The task of cleaning up UTXOs, regardless of why they exist (whether as an internal side effect of the protocol or not), is separate from anchoring a virtual PSBT. It’s distinct from the primary function of the endpoint. The current change opportunistically extends the endpoint’s core behavior by adding cleanup logic, which introduces privacy implications.

I think we should let the user choose whether to at least skip that cleanup so that the expected (and previous) behaviour is at least still available.

ffranr: Thinking about this now. I'm not even sure if we should sweep zero value UTXOs in `AnchorVirtualPsbts` at all. Maybe it should be a completely distinct cleanup focused endpoint. Maybe for now we can just make this opt-in.

darioAnongba: > The current change opportunistically extends the endpoint’s core behavior by adding cleanup logic

I do agree with this and personally hate unintended side effects. I agree we should remove this. Only issue is if some user is always using low-level endpoints to create transactions, then he will never be able to sweep I think. A dedicated endpoint might make sense or a way to get the UTXOs when querying `ListUtxos`.

tapdb/sqlc/migrations/000048_add_swept_flag.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Add swept flag to managed_utxos table to track when UTXOs have been swept
+ALTER TABLE managed_utxos ADD COLUMN swept BOOLEAN NOT NULL DEFAULT FALSE;
```
GeorgeTsagk: shouldn't we also add a golang-level migration step to automatically detect which of our managed utxos are already spent?

tapsend/allocation.go
```diff
@@ -147,6 +148,15 @@ type Allocation struct {
 	// output.
 	InternalKey *btcec.PublicKey
 
+	// Bip32Derivation is the BIP32 derivation info for the internal key.
+	// This is used to preserve the key derivation through the allocation
+	// flow so that PSBTs can be properly signed.
+	Bip32Derivation []*psbt.Bip32Derivation
```
GeorgeTsagk, line 154: why is this needed now?

rpcserver.go
GeorgeTsagk: Keeping the cleanup integrated with other wallet actions feels more natural. It is also more cost-efficient as we naturally batch on-the-go.

I think we should definitely wrap all these behind a config flag. Whether it should be on/off by default I'm not sure about, we can start with off-by-default and maybe enable in a future release.

GeorgeTsagk: > wrap all these behind a config flag.

This I believe should be a daemon-level flag and should extend all the way to `tapfreighter`/`tapsend`

itest/burn_test.go
GeorgeTsagk: another reason to wrap the sweeping behind a flag is to make sure we preserve old behavior and not alter test coverage

this way we'd disable sweeping for tests like this one and not have to delete the assertion

itest/zero_value_anchor_test.go
```diff
@@ -0,0 +1,178 @@
...
+	require.Len(t.t, burnUtxos.ManagedUtxos, 1)
+
+	// Test 3: Send transactions sweeps zero-value burns.
+	rpcAssets4 := MintAssetsConfirmBatch(
+		t.t, t.lndHarness.Miner().Client, t.tapd,
+		[]*mintrpc.MintAssetRequest{simpleAssets[0]},
+	)
+	genInfo4 := rpcAssets4[0].AssetGenesis
+
+	// Send partial amouunt. This should NOT create a tombstone output
```
GeorgeTsagk:
```suggestion
	// Send partial amount. This should NOT create a tombstone output
```

itest/zero_value_anchor_test.go
```diff
@@ -0,0 +1,178 @@
...
+		[]uint64{partialAmount, partialAmount}, 3, 4,
+	)
+	AssertNonInteractiveRecvComplete(t.t, secondTapd, 3)
+
+	// The burn UTXO should have been swept.
+	AssertBalances(
+		t.t, t.tapd, 0, WithScriptKeyType(asset.ScriptKeyBurn),
+		WithNumUtxos(0), WithNumAnchorUtxos(0),
+	)
+}
```
GeorgeTsagk, line 203: I think the config flag that controls whether sweeping is activated will also help us add slightly better coverage here too.

We could accumulate multiple burn UTXOs / tombstones and then turn the flag on and see them all being swept.

GeorgeTsagk:
also needs a rebase

tapdb/sqlc/migrations/000048_add_swept_flag.up.sql
darioAnongba: Normally not because zero-value anchor UTXOs cannot have been swept by normal usage of `tapd`. We're not responsible if a user happened to sweep these UTXOs somehow manually and corrupted the state of his DB.

tapsend/allocation.go
darioAnongba: Before, we were not persisting the necessary information in the DB to be able to sign the input for zero value UTXOs. Laolu described the bug here: https://github.com/lightninglabs/taproot-assets/blob/main/docs/debug-handbook/psbt-finalization-failures.md#a-bip32-key-derivation-mismatch.

 Specifically the `KeyFamily or KeyIndex not preserved when storing keys`.

itest/zero_value_anchor_test.go
darioAnongba: agree but to test this we would need to start `tapd` with a given config, stop it in the middle of the test and restart it with a different config. Is this what you mean?

rpcserver.go
darioAnongba: The issue I see here is we're starting to diverge from the basic need that is avoiding "garbage" to grow indefinitely. 

Zero-value UTXOs garbage collection is part of the realm of coin selection. Most Bitcoin wallets perform coin selection on behalf of the user and only few advanced wallets allow selecting and freezing specific UTXOs (like Sparrow). This is even more important for `tapd` where the coin selection is tied to the assets.

We initially started with a feature that naturally preserved liveness and safety properties and abstracted from the user the entire notion of zero-value UTXOs, which is a notion internal to the protocol, but now we are:
- Worrying about privacy when the user calls `AnchorVirtualPsbts` if we select zero value UTXOs the same way we do for burn or send.
- Adding a global config that defines whether or not zero-value UTXOs are sweeped, which implies that the set of zero-value UTXOs can now grow outside the limits of a transaction, breaking liveness and safety so we need to verify the size of the tx not to exceed the weight limit and maybe other side effects.

Taproot assets inherits from the same coin selection behavior as Bitcoin. If you have a small change UTXO lying around, a simple thing to do is consolidate it when sending a transaction. but the most efficient thing to do can be quite complex. Unless we want to implement advanced coin selection, I say we:
- do not add a global flag
- do not add a flag to the requests
- always sweep in send and burn
- never expose the notion of zero-value UTXOs in the API.
- think about coin selection strategies (optimize fees, minimize utxos, consolidate, etc).

darioAnongba: I added a config flag, disabled by default to handle this behaviour.

itest/zero_value_anchor_test.go
darioAnongba: test added

itest/burn_test.go
darioAnongba: test reverted with flag disabled so same behaviour is maintained

tapdb/assets_store.go
```diff
@@ -1320,6 +1333,170 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+				isBurn := len(aAsset.PrevWitnesses) > 0 &&
+					asset.IsBurnKey(
+						aAsset.ScriptKey.PubKey,
+						aAsset.PrevWitnesses[0],
+					)
+
+				if !isTombstone && !isBurn {
+					allZeroValue = false
+					break
+				}
```
ffranr: We could replace the `isBurn` statement here with `chainAsset.Asset.IsBurn()`.

Also, we should add an `IsTombstone` method to the `Asset` type to make this logic easier to discover.

tapdb/assets_store.go
```diff
@@ -3302,9 +3503,25 @@ func (a *AssetStore) LogAnchorTxConfirm(ctx context.Context,
...
+			}
+
+			err = q.MarkManagedUTXOAsSwept(ctx, outpointBytes)
+			if err != nil {
+				return fmt.Errorf("unable to mark zero-value "+
+					"UTXO as swept: %w", err)
+			}
+
+			log.Infof("Marked zero-value UTXO %v as swept",
+				outpoint)
```
ffranr: debug level here?

tapfreighter/chain_porter.go
```diff
@@ -1897,6 +1907,20 @@ func (p *ChainPorter) unlockInputs(ctx context.Context, pkg *sendPackage) {
...
+			pkg.ZeroValueInputs,
+			func(z *ZeroValueInput) wire.OutPoint {
+				return z.OutPoint
+			},
+		)
+
+		err := p.cfg.AssetWallet.ReleaseCoins(
+			ctx, zeroValueOutpoints...,
+		)
+		if err != nil {
```
ffranr: We could combine this call with the above `p.cfg.AssetWallet.ReleaseCoins` call such that we release all coins atomically.

tapfreighter/wallet.go
```diff
@@ -1191,28 +1231,32 @@ func (f *AssetWallet) AnchorVirtualTransactions(ctx context.Context,
 	// it itself.
 	addAnchorPsbtInputs(sendPacket, params.ActivePackets)
 
+	// Add zero-value inputs that should be swept as additional inputs.
+	numZeroValueInputs := len(params.ZeroValueInputs)
+	if numZeroValueInputs > 0 {
+		log.Infof("Sweeping %d zero-value UTXOs", numZeroValueInputs)
```
ffranr: consider: `log.DebugS` rather than `Infof`

itest/tapd_harness.go
```diff
@@ -78,6 +78,15 @@ var (
...
 	defaultProofRetrievalDelay = 200 * time.Millisecond
+
+	// pgFixtureRegistry keeps track of Postgres fixtures per tapd base
+	// directory so we can re-use the same database across restarts.
+	pgFixtureRegistry = struct {
+		sync.Mutex
+		fixtures map[string]*tapdb.TestPgFixture
+	}{
+		fixtures: make(map[string]*tapdb.TestPgFixture),
+	}
```
ffranr: does this belong in this PR?

itest/zero_value_anchor_test.go
```diff
@@ -176,3 +178,217 @@ func testZeroValueAnchorSweep(t *harnessTest) {
...
+
+	// Create a new harness with sweeping enabled.
+	aliceTapd, err := newTapdHarness(
+		t.t, t, tapdCfg,
+		func(ho *harnessOpts) {
+			ho.sweepZeroValueAnchorUtxos = true
+			ho.proofCourier = t.proofCourier
+		},
+	)
+	require.NoError(t.t, err)
```
ffranr: instead of this, is it possible to modify `tapdHarness.clientCfg` and then call `start` method?

ffranr:
Non-blocking, but for the `sweepZeroValueAnchorUtxos` config field: “sweep orphaned UTXOs” might be clearer. “Anchor” is implied by UTXO, and “zero-value” isn’t accurate since they contain 1000 sats. If they were zero value there would be nothing to sweep.

tapdb/assets_store.go
darioAnongba: added.

tapfreighter/chain_porter.go
darioAnongba: Great catch, don't know why it wasn't like that in the first place

itest/zero_value_anchor_test.go
darioAnongba: I guess I ended up with too much unnecessary complexity, thanks! Much simpler now

itest/tapd_harness.go
darioAnongba: removed! I went into a weird rabbit hole trying to keep the fixtures between restarts but they are actually kept by default if noDelete is true.

darioAnongba:
I like `SweepOrphanUtxos` better as well. I renamed it everywhere at the config level, let me know if you think we should rename all the way down to the DB layer. In the context of assets, zero-value is correct though, but more verbose and uglier.

itest/burn_test.go
```diff
@@ -246,11 +246,6 @@ func testBurnAssets(t *harnessTest) {
 		WithScriptKeyType(asset.ScriptKeyBurn),
 	)
 
-	_, err = t.tapd.ListAssets(ctxt, &taprpc.ListAssetRequest{
-		IncludeSpent: true,
-	})
-	require.NoError(t.t, err)
```
GeorgeTsagk, line 265: why is this being removed?

would be nice to at least have some context around all the surrounding deletions in this commit

itest/integration_test.go
```diff
@@ -119,7 +119,7 @@ func TestTaprootAssetsDaemon(t *testing.T) {
 			// assure no state is taken over between runs.
 			tapdHarness, uniHarness, proofCourier := setupHarnesses(
 				t1, ht, lndHarness, uniServerLndHarness,
-				testCase.proofCourierType,
+				testCase.proofCourierType, testCase.tapdOptions...,
```
GeorgeTsagk: nit: long line

tapfreighter/wallet.go
```diff
@@ -663,10 +667,14 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
...
 	}
 
-	zeroValueInputs, err := f.cfg.CoinSelector.SelectZeroValueCoins(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("unable to select zero-value "+
-			"UTXOs: %w", err)
+	var zeroValueInputs []*ZeroValueInput
+	if f.cfg.SweepOrphanUtxos {
+		zeroValueInputs, err = f.cfg.CoinSelector.
+			SelectZeroValueCoins(ctx)
```
GeorgeTsagk: nit
```suggestion
		zeroValueInputs, err = f.cfg.CoinSelector.SelectZeroValueCoins(
		    ctx,
		)
```
or this (which seems better to me)
```suggestion
		coinSelector := f.cfg.CoinSelector
		zeroValueInputs, err = coinSelector.SelectZeroValueCoins(ctx)
```

tapfreighter/wallet.go
```diff
@@ -740,10 +748,14 @@ func (f *AssetWallet) FundBurn(ctx context.Context,
...
```
GeorgeTsagk: same as previous one

rpcserver.go
```diff
@@ -2671,11 +2671,19 @@ func (r *rpcServer) AnchorVirtualPsbts(ctx context.Context,
...
-		return nil, fmt.Errorf("unable to fetch zero-value "+
-			"UTXOs: %w", err)
+	// Fetch zero-value UTXOs that should be swept as additional inputs if
+	// the feature is enabled.
+	var (
+		zeroValueInputs []*tapfreighter.ZeroValueInput
+		err             error
+	)
+	if r.cfg.SweepOrphanUtxos {
+		zeroValueInputs, err = r.cfg.AssetStore.
```
GeorgeTsagk: same

itest/test_list_on_test.go
```diff
@@ -104,6 +104,13 @@ var allTestCases = []*testCase{
 	{
 		name: "zero value anchor sweep",
 		test: testZeroValueAnchorSweep,
+		tapdOptions: []Option{
```
GeorgeTsagk: instead of exposing the tapdOptions on the test case list you can use them directly in the test case like this:
https://github.com/lightninglabs/taproot-assets/blob/d9383ba45bf2014db958cd6908ac3a28e98255d7/itest/rfq_test.go#L872-L876

itest/zero_value_anchor_test.go
```diff
@@ -290,7 +290,7 @@ func testZeroValueAnchorAccumulation(t *harnessTest) {
 	require.NoError(t.t, t.tapd.stop(false))
 
 	// Enable sweeping in the config.
-	t.tapd.clientCfg.Wallet.SweepZeroValueAnchorUtxos = true
+	t.tapd.clientCfg.Wallet.SweepOrphanUtxos = true
```
GeorgeTsagk, line 316: oh didn't know we had such direct access to the config

Why not do this for `testZeroValueAnchorSweep` too? This way we can get rid of `tapdOptions` on the test case itself

itest/integration_test.go
darioAnongba: Isn't this line less than 80 chars? What's the rule here? I added a line break

itest/burn_test.go
darioAnongba: This is a residue of logging statements I removed in a previous PR but forgot to remove the RPC call. As you can see, the result from this call is not used, it was before to log (literally dump the entire response).

In general, I'm removing the logging statements in the tests, which are particularly annoying (it was so bad that running the tests in the IDE terminal completely filled it with dumped objects and you couldn't even scroll to the top). 
Would you like me to add to the commit description why I'm removing the log statements? I have a PR coming specifically to improve the itest environment.

tapfreighter/wallet.go
darioAnongba: Renamed to better and shorter fn name.

rpcserver.go
darioAnongba: Renamed to better and shorter fn name.

itest/test_list_on_test.go
darioAnongba: The way I understand this is that everytime a test starts, a `tapd` instance is automatically created in `t.tapd`. So there is no reason not to be able to modify the config of this "initial" `tapd`. I think the `tapdOptions` solutions is more elegant, saves resources (since you are starting a new `tapd` instance here) and is less verbose.
Correct me if mistaken.

itest/zero_value_anchor_test.go
darioAnongba: Unless mistaken, the `tapd` instance automatically created when starting the test would need to be stopped, the config changed and restarted for this to work, which are operations unrelated to the test itself (not like here where we actually want to do this restart as part of the test itself).

GeorgeTsagk: Sure, sounds reasonable, we would otherwise be executing a restart everytime we'd want to set the flag.

docs/release-notes/release-notes-0.8.0.md
```diff
@@ -37,10 +37,11 @@
 
 ## Functional Updates
 
-- [Garbage collection of zero-value UTXOs](https://github.com/lightninglabs/taproot-assets/pull/1832)
+- [Garbage collection of orphaned UTXOs](https://github.com/lightninglabs/taproot-assets/pull/1832)
```
GeorgeTsagk, line 35: nit: merge this docs update with the previous docs commit, then move docs commit to the end so that it's the last one

darioAnongba: done

tapdb/assets_store.go
```diff
@@ -489,6 +493,14 @@ type ManagedUTXO struct {
 	// LeaseExpiry is the expiry time of the lease on this UTXO. If the
 	// zero, then this UTXO isn't leased.
 	LeaseExpiry time.Time
+
+	// PkScript is the pkScript of the anchor output. This is populated
+	// when fetching zero-value anchor UTXOs to enable PSBT creation.
+	PkScript []byte
```
ffranr: I don't think this field is used anywhere. Can it be removed perhaps?

tapdb/assets_store.go
```diff
@@ -1320,6 +1333,159 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+	outpointBytes, err := encodeOutpoint(outpoint)
+	if err != nil {
+		return fmt.Errorf("unable to encode outpoint: %w", err)
+	}
+
+	var writeTxOpts AssetStoreTxOptions
+	return a.db.ExecTx(ctx, &writeTxOpts, func(q ActiveAssetsStore) error {
+		return q.MarkManagedUTXOAsSwept(ctx, outpointBytes)
+	})
+}
```
ffranr: I don't think this method is used anywhere.

tapdb/assets_store.go
```diff
@@ -1320,6 +1333,159 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+
+// FetchZeroValueAnchorUTXOs fetches all managed UTXOs that contain only
+// zero-value assets (tombstones and burns).
+func (a *AssetStore) FetchZeroValueAnchorUTXOs(ctx context.Context) (
```
ffranr: I think you should set the method name to its final form in this commit rather than renaming it in a later one.

tapdb/assets_store.go
```diff
@@ -1320,6 +1333,159 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+			// Extract the pkScript from the transaction.
+			var tx wire.MsgTx
+			err = tx.Deserialize(
+				bytes.NewReader(chainTx.RawTx),
+			)
+			if err != nil {
+				log.Warnf("Failed to deserialize tx for "+
+					"%v: %v, skipping", anchorPoint, err)
+				continue
+			}
```
ffranr, line 1448: I don’t think we should only warn here. There shouldn’t be any transactions in the database that we can’t deserialize, so I’d prefer to return an error to avoid letting a faulty row persist longer than necessary.

tapdb/assets_store.go
```diff
@@ -1320,6 +1333,159 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+				}
+			}
+
+			if !allZeroValue {
+				continue
+			}
+
+			log.Debugf("Adding zero-value anchor to sweep list "+
+				"(outpoint=%s)", anchorPoint.String())
+
```
ffranr: We can use `DebugS` here since we have `ctx`.

tapdb/assets_store.go
```diff
@@ -1320,6 +1333,159 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+			}
+
+			// Fetch the chain transaction to get the actual
+			// pkScript.
+			chainTx, err := q.FetchChainTx(ctx, anchorPoint.Hash[:])
+			if err != nil {
+				log.Warnf("Failed to fetch chain tx for "+
+					"%v: %v, skipping", anchorPoint, err)
+				continue
+			}
```
ffranr, line 1437: If we get a `sql.ErrNoRows` then continuing seems fine. For any other error, I think we should return it.

tapdb/assets_store.go
```diff
@@ -1320,6 +1333,159 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+	var results []*tapfreighter.ZeroValueInput
+
+	readOpts := NewAssetStoreReadTx()
+	now := a.clock.Now().UTC()
+
+	dbErr := a.db.ExecTx(ctx, &readOpts, func(q ActiveAssetsStore) error {
+		utxos, err := q.FetchManagedUTXOs(ctx)
+		if err != nil {
+			return err
+		}
```
ffranr, line 1352: I think we should wrap `err` with `fmt.Errorf` here to add context. Same for the `return nil, dbErr` below. Adding context consistently tends to save time during troubleshooting in my experience.

tapdb/assets_store.go
```diff
@@ -1320,6 +1333,159 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+					Now: sql.NullTime{
+						Time:  now,
+						Valid: true,
+					},
+				},
+			)
+			if err != nil {
+				return fmt.Errorf("failed to query assets at "+
+					"anchor: %w", err)
+			}
```
ffranr, line 1390: Something for your consideration:

I’ve found that wrapping errors like this is usually enough:

```
fmt.Errorf("query chain assets at anchor: %w", err)
```

Since it will appear inside a higher level error anyway, adding “failed to” doesn’t add much in my view.

tapdb/assets_store.go
```diff
@@ -3302,9 +3492,25 @@ func (a *AssetStore) LogAnchorTxConfirm(ctx context.Context,
...
+			if err != nil {
+				return fmt.Errorf("failed to encode "+
+					"zero-value outpoint: %w", err)
+			}
+
+			err = q.MarkManagedUTXOAsSwept(ctx, outpointBytes)
+			if err != nil {
+				return fmt.Errorf("unable to mark zero-value "+
+					"UTXO as swept: %w", err)
+			}
```
ffranr, line 3506: Thinking about it now, instead of the `spent` field being a bool, it might be better for it to reference the transaction that spends it. That would let us reverse the spent status cleanly during a reorg. We could even consider making it a reference to a transaction already in our database, although that might be too restrictive.

tapdb/assets_store.go
darioAnongba: applied.

tapdb/assets_store.go
darioAnongba: agree!

tapdb/assets_store.go
darioAnongba: Applied but this is sadly inconsistent with the rest of the codebase, we currently do not add context to `dbErr` anywhere as you can see.

Would be good to define a methodology around error handling. Clean Architecture also helps here (use internal error types and add context only once in business layer, map at the edge to user-facing error).

tapdb/assets_store.go
darioAnongba: I agree we should reduce verbosity as much as possible but I don't feel comfortable breaking the pattern of the rest of the codebase (we seem to use everywhere one of these):
- `unable to ...`
- `error...`
- `failed to ...`

Happy to work on a PR to reduce verbosity globally at some point.

tapdb/assets_store.go
darioAnongba: Yes definitely better to have a foreign key instead of a boolean. Applied.

tapdb/assets_store.go
```diff
@@ -2476,6 +2633,30 @@ func (a *AssetStore) LogPendingParcel(ctx context.Context,
...
+				return fmt.Errorf("unable to encode "+
+					"zero-value outpoint: %w", err)
+			}
+
+			err = q.UpdateUTXOLease(ctx, UpdateUTXOLease{
+				LeaseOwner: finalLeaseOwner[:],
+				LeaseExpiry: sql.NullTime{
+					Time:  finalLeaseExpiry.UTC(),
+					Valid: true,
+				},
```
ffranr, line 2652: Could be a helper like `sqlInt16`

tapdb/assets_store.go
```diff
@@ -1320,6 +1329,154 @@ func (a *AssetStore) FetchManagedUTXOs(ctx context.Context) (
...
+				continue
+			}
+
+			if u.SweptTxnID.Valid {
+				continue
+			}
+
+			var anchorPoint wire.OutPoint
+			err := readOutPoint(
+				bytes.NewReader(u.Outpoint), 0, 0, &anchorPoint)
```
ffranr: trailing `)` should be on a new line

tapdb/assets_store.go
```diff
@@ -3302,9 +3483,29 @@ func (a *AssetStore) LogAnchorTxConfirm(ctx context.Context,
...
+			if err != nil {
+				return fmt.Errorf("failed to encode "+
+					"zero-value outpoint: %w", err)
+			}
+
+			err = q.MarkManagedUTXOAsSwept(ctx,
+				MarkManagedUTXOAsSweptParams{
+					Outpoint:     outpointBytes,
+					SweepingTxid: conf.AnchorTXID[:],
+				})
```
ffranr: trailing `)` should be on a new line

litbot-9000:
@ffranr: review reminder

tapfreighter/chain_porter.go
```diff
@@ -1883,18 +1894,33 @@ func (p *ChainPorter) unlockInputs(ctx context.Context, pkg *sendPackage) {
 	// sanity-check that we have known input commitments to unlock, since
 	// that might not always be the case (for example if another party
 	// contributes inputs).
-	if pkg.SendState < SendStateStorePreBroadcast &&
-		len(pkg.InputCommitments) > 0 {
+	// Also unlock any zero-value UTXOs that were leased for this package.
+	if pkg.SendState < SendStateStorePreBroadcast {
+		// Gather all outpoints to unlock in a single array
```
ffranr: Punctuation missing for comments here and below.

tapfreighter/coin_select.go
```diff
@@ -199,4 +185,46 @@ func (s *CoinSelect) selectForAmount(minTotalAmount uint64,
...
+	zeroValueInputs, err := s.coinLister.FetchOrphanUTXOs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch zero-value UTXOs: %w",
+			err)
+	}
+
+	// We now need to lock/lease/reserve those selected coins so
+	// that they can't be used by other processes.
+	if len(zeroValueInputs) > 0 {
+		expiry := time.Now().Add(defaultCoinLeaseDuration)
```
ffranr: Worth using `time.Now().UTC()` here IMO so every time as much as possible is UTC at every point.

tapfreighter/interface.go
```diff
@@ -431,6 +458,29 @@ func (out *TransferOutput) UniqueKey() (OutputIdentifier, error) {
...
+
+// IsLocal returns true if the transfer output is a local script key.
+func (out *TransferOutput) IsLocal() bool {
+	return out.ScriptKeyLocal
+}
+
+// IsSpendable returns true if the transfer output is spendable.
+func (out *TransferOutput) IsSpendable() bool {
+	return out.IsLocal() || out.IsTombstone() || out.IsBurn()
+}
```
ffranr: This is a bit unclear IMO.

If `out.IsBurn() == true` then `IsSpendable()` will return true? That doesn't make sense to me. A burn is not spendable. Do we want to say something about the on-chain anchor being spendable here?

tapfreighter/wallet.go
```diff
@@ -655,7 +663,13 @@ func (f *AssetWallet) FundPacket(ctx context.Context,
 		return nil, err
 	}
 
-	// If we return with an error, we want to release the coins we've
+	zeroValueInputs, err := f.cfg.CoinSelector.SelectOrphanCoins(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to select zero-value "+
+			"UTXOs: %w", err)
+	}
```
ffranr: If we return an error here we will not release the selected coins from this line above:
```
selectedCommitments, err := f.cfg.CoinSelector.SelectCoins(
```

We should change this code such that `selectedCommitments` and `zeroValueInputs` are defined as empty first. Then we define the release defer func. Then we try to populate those fields.

tapfreighter/wallet.go
```diff
@@ -716,7 +740,13 @@ func (f *AssetWallet) FundBurn(ctx context.Context,
...
```
ffranr: Same here. If we return an error at this point then we wont release all coins.

itest/zero_value_anchor_test.go
```diff
@@ -0,0 +1,365 @@
...
+	require.NoError(t.t, err)
+
+	sendResp2, _ := sendAssetsToAddr(t, t.tapd, bobAddr2)
+
+	ConfirmAndAssertOutboundTransfer(
+		t.t, t.lndHarness.Miner().Client, t.tapd, sendResp2,
+		genInfo2.AssetId,
+		[]uint64{0, assetAmount}, 1, 2,
+	)
+	AssertNonInteractiveRecvComplete(t.t, secondTapd, 2)
```
ffranr, line 103: Maybe we can inspect this transaction to ensure that the additional zero-value tombstone input is present.

itest/zero_value_anchor_test.go
```diff
@@ -0,0 +1,365 @@
...
+
+	// Restart with the modified config.
+	require.NoError(t.t, t.tapd.start(false))
+
+	// Wait for the node to fully sync after restart.
+	time.Sleep(2 * time.Second)
+
+	// Verify that the zero-value UTXOs are still present after restart.
+	//nolint:lll
+	tombstoneUtxosAfterRestart, err := t.tapd.ListUtxos(ctxb, &taprpc.ListUtxosRequest{
```
ffranr: Instead of this sleep you could possibly use `ListUtxosRPC` itest helper. Sleep can cause flakes.

itest/assertions.go
```diff
@@ -1325,16 +1321,6 @@ func AssertAssetOutboundTransferWithOutputs(t *testing.T,
...
-	require.NoError(t, err)
-
-	transferResp, err := sender.ListTransfers(
-		ctxb, &taprpc.ListTransfersRequest{},
-	)
-	require.NoError(t, err)
-
-	transferRespJSON, err := formatProtoJSON(transferResp)
-	require.NoError(t, err)
-	t.Logf("Got response from list transfers: %v", transferRespJSON)
```
ffranr, line 1337: Why remove these lines? That doesn't seem to fix under the commit subject. Same further above also.

itest/test_list_on_test.go
```diff
@@ -104,6 +104,13 @@ var allTestCases = []*testCase{
...
+			WithSweepOrphanUtxos(),
+		},
+	},
```
ffranr: hmm I don't think I'm a fan of this change. I think it was clearer if `t.tapd` had the default options and we just create a new tapd instance if we need to specify configuration.  I think we should remove the `proofCourierType` field also.

I don't think we should consult this list to see how the primary tapd node is configured. But I won't block on this change, we can revisit later.

tapdb/assets_store.go
darioAnongba: no such thing as a `sqlTime`. I could add one but then I'll have to change everywhere sql.NullTime is used.

itest/assertions.go
darioAnongba: Do you mean that I shouldn't remove these logs in this PR or that I should move them to a separate commit?

> Why remove these lines?

They're only there to dump the entirety of the responses into the console, which serves no purpose (we should use `require` to validate assertions).
I've been particularly annoyed by these logs so I try to remove them when I can. Happy to revert or create a dedicated PR.

tapfreighter/wallet.go
darioAnongba: Great catch! 
This only applies to the `zeroValueInputs` as if `SelectCoins` fails, it means that no coins have been leased. It's also the current implementation.

I moved the defer func() above the `SelectOrphanCoins`.

tapfreighter/wallet.go
darioAnongba: Great catch! 
This only applies to the `zeroValueInputs` as if `SelectCoins` fails, it means that no coins have been leased. It's also the current implementation.

I moved the defer func() above the `SelectOrphanCoins`.

itest/zero_value_anchor_test.go
darioAnongba: Thanks! I was looking for something like this.

itest/test_list_on_test.go
darioAnongba: George had a similar remark as well so I removed this.

tapfreighter/interface.go
darioAnongba: Agree, this function makes no sense as part of the TransferOutput methods. It only makes sense as part of the context in `importLocalAddresses`.

itest/zero_value_anchor_test.go
darioAnongba: Only way to verify is to deserialize the anchor transaction itself and check the inputs, which is maybe what you meant? I added this check. 
It is not exactly deterministic because LND is the one choosing to add inputs or not so we need a `GreaterOrEqual`.

itest/assertions.go
darioAnongba: I removed this change so it's out of the PR.

tapdb/assets_store.go
ffranr: I'm suggesting you add the helper, yes. But I don't think you have to change everywhere sql.NullTime is used. Not a blocker for this PR though.

darioAnongba: ok, adding quickly in new PR

PR #1834, opened 2025-10-03
rfq+server: properly configure gRPC keep alive settings for client+server

### LINKED ISSUE #1814: [bug]: Issue with Paying Sats Invoices Using Taproot Assets via sendPayment with rfqId
### Background

Image

- We are working with a setup where we have Litd nodes (node a, b, c) and a Sats Node (as per the attached diagram). Node b acts as the Edge Node.
- Currently, we use the Taproot Assets sendPayment API to pay invoices from Litd Nodes.
- By design, sendPayment first places an AddAssetSellOrder and then routes the payment to the destination.
- Since we run our own Price Oracle Server to fetch BTC :left_right_arrow: TaprootAsset exchange rates, we have faced the following problem:

- When calling sendPayment, the API attempts to place a SellOrder, but we hit a "**connection reset by peer**" error.
- 
- This is consistent with the issue described here: [lightninglabs/taproot-assets#1267](https://github.com/lightninglabs/taproot-assets/issues/1267).

The root cause seems to be a network layer error due to idleness between the Edge Node and the Price Oracle server. The channel closes without sending a **GO_AWAY message, leaving the client unaware of the broken connection, which results in payment failure.**

Our Approach to Mitigate
To avoid this error, we split the process into two steps:

      1. Manually place SellOrder for the msats amount specified in the invoice.
      2. We handle retries and timeouts explicitly.
      3. This ensures reliability and always gives us a PeerAcceptedSellOrder.
      4. Call sendPayment with the rfqId obtained from the PeerAcceptedSellOrder response.

**Current Problem:**

- This approach works in most cases. However, we hit a specific issue:

      **When paying Sats invoices with TaprootAssets using sendPayment with the manually provided rfqId, we consistently receive NO_ROUTE errors.**

The strange part is that:
- A valid path does exist.
- If we retry the same payment without setting the rfqId, it succeeds.

we are using  tapd [v0.6.1](https://github.com/lightninglabs/taproot-assets/releases/tag/v0.6.1)
"lightning-terminal version : v0.15.1-alpha"

### Steps to reproduce

  1. Create an invoice on the Sats Node.
  2. Decode the invoice to get the msats amount.
  3. On Node A(LitdNode), manually place a SellOrder for the decoded msats amount.
  4. Use the returned acceptedPeerQuoteId in the rfqId field of the sendPayment API.
  5. Call sendPayment.

Observed Result:
 The payment fails with NO_ROUTE.

Payment with hash 86ae98dc44ffdeac5097b15f62f863f6309e73e454f53339156ad94d1f52ff5f failed under this scenario. I have attached the relevant log file for this failed attempt.

[output (1).log](https://github.com/user-attachments/files/22507103/output.1.log)

Roasbeef:
## Problem

Users were experiencing "connection reset by peer" errors when using the RFQ system with price oracle servers (issue #1814). The core problem manifested when:

1. An edge node establishes a connection to a price oracle server
2. The connection sits idle between RFQ price queries (potentially for minutes or hours)
3. The network layer silently closes the idle connection without sending a proper GOAWAY frame
4. The client remains unaware of the broken connection
5. The next RFQ request fails immediately with "connection reset by peer"
6. A retry succeeds because it establishes a fresh connection

This created a poor user experience where the first payment attempt after an idle period would always fail, requiring manual retries.

## Root Cause

The issue had two contributing factors:

**Missing Client-Side Keepalive**: The RFQ price oracle client (`rfq/oracle.go`) had **no keepalive configuration at all**. Without active health probing, the client couldn't detect when connections were broken and wouldn't prevent intermediaries (NATs, load balancers, firewalls) from timing out idle connections.

**Aggressive Server-Side Timeout**: The tapd gRPC server only configured `MaxConnectionIdle: 2 minutes`, which would aggressively close any connection idle for more than 2 minutes. This is far too short for RFQ operations where price queries may be infrequent.

## Solution

We implemented comprehensive bidirectional keepalive configuration following the same pattern used by lnd:

### Client-Side Configuration (`rfq/oracle.go`)

Added to both `serverDialOpts()` and `insecureServerDialOpts()`:

```go
grpc.WithKeepaliveParams(keepalive.ClientParameters{
    Time:                30 * time.Second,  // Ping every 30s when idle
    Timeout:             20 * time.Second,  // Wait 20s for ping response
    PermitWithoutStream: true,              // Ping even without active RPCs
})
```

**Key Settings:**
- `Time: 30s` - Client sends a ping after 30 seconds of inactivity. This is frequent enough to keep connections alive through most intermediaries while being conservative with network resources.
- `Timeout: 20s` - If the server doesn't respond to a ping within 20 seconds, the connection is considered dead. This allows prompt detection of network issues.
- `PermitWithoutStream: true` - **Critical setting**. Allows pings even when no RPC calls are active. Without this, the client wouldn't send pings during idle periods, defeating the entire purpose.

### Server-Side Configuration (`server.go`)

Enhanced the existing minimal configuration with comprehensive settings:

```go
serverKeepalive := keepalive.ServerParameters{
    Time:              time.Minute,      // Ping clients every 1min when idle
    Timeout:           20 * time.Second, // Wait 20s for ping response
    MaxConnectionIdle: time.Hour * 24,   // Allow 24h of idle time
}

clientKeepalive := keepalive.EnforcementPolicy{
    MinTime:             5 * time.Second, // Min time between client pings
    PermitWithoutStream: true,            // Accept client pings without active RPCs
}
```

**Key Settings:**
- `Time: 1min` - Server actively pings clients every minute when idle. This provides server-initiated health checking.
- `Timeout: 20s` - Server waits 20 seconds for ping responses before considering the connection dead.
- `MaxConnectionIdle: 24h` - Extended from 2 minutes to 24 hours. Allows long-lived connections that are crucial for RFQ operations. The active pinging mechanism (via `Time` parameter) handles health checking, so we don't need aggressive idle timeouts.
- **EnforcementPolicy.PermitWithoutStream: true** - **Critical setting**. Tells the server to accept keepalive pings from clients even when there are no active RPC calls. Without this, the server would reject client pings on idle connections with a GOAWAY, breaking the client-side keepalive mechanism.
- `MinTime: 5s` - Prevents abusive clients from sending pings too frequently (DoS protection).

## How It Works

With these settings in place:

1. **During Idle Periods**: Both client and server send periodic pings (client every 30s, server every 1min) to verify the connection is still alive
2. **Network Issue Detection**: If a ping doesn't get a response within 20s, the connection is immediately closed and the application is notified
3. **Intermediary Keepalive**: The regular pings prevent NATs, load balancers, and firewalls from timing out the connection
4. **No Silent Failures**: Connections are never silently broken - either they work or they fail fast with proper notification

Fixes #1814

### DIFF
--- a/rfq/oracle.go
+++ b/rfq/oracle.go
@@ -18,6 +18,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/keepalive"
 )
 
 // PriceQueryIntent is an enum that represents the intent of a price rate
@@ -196,6 +197,21 @@ func serverDialOpts() ([]grpc.DialOption, error) {
 	transportCredentials := credentials.NewTLS(&tlsConfig)
 	opts = append(opts, grpc.WithTransportCredentials(transportCredentials))
 
+	// Configure client-side keepalive parameters. These settings ensure
+	// the client actively probes the connection health and prevents idle
+	// connections from being silently closed by intermediaries or the
+	// server.
+	opts = append(opts, grpc.WithKeepaliveParams(keepalive.ClientParameters{
+		// Ping server after 30 seconds of inactivity.
+		Time: 30 * time.Second,
+		// Wait 20 seconds for ping response.
+		Timeout: 20 * time.Second,
+		// Permit keepalive pings even when there are no active
+		// streams. This is critical for long-lived connections with
+		// infrequent RFQ requests.
+		PermitWithoutStream: true,
+	}))
+
 	return opts, nil
 }
 
@@ -208,6 +224,21 @@ func insecureServerDialOpts() ([]grpc.DialOption, error) {
 	opts = append(opts, grpc.WithTransportCredentials(
 		insecure.NewCredentials(),
 	))
+
+	// Configure client-side keepalive parameters. These settings ensure
+	// the client actively probes the connection health and prevents idle
+	// connections from being silently closed by intermediaries or the
+	// server.
+	opts = append(opts, grpc.WithKeepaliveParams(keepalive.ClientParameters{
+		// Ping server after 30 seconds of inactivity.
+		Time: 30 * time.Second,
+		// Wait 20 seconds for ping response.
+		Timeout: 20 * time.Second,
+		// Permit keepalive pings even when there are no active
+		// streams. This is critical for long-lived connections with
+		// infrequent RFQ requests.
+		PermitWithoutStream: true,
+	}))
 
 	return opts, nil
 }
--- a/server.go
+++ b/server.go
@@ -370,11 +370,34 @@ func (s *Server) RunUntilShutdown(mainErrChan <-chan error) error {
 	serverOpts = append(serverOpts, rpcServerOpts...)
 	serverOpts = append(serverOpts, ServerMaxMsgReceiveSize)
 
-	keepAliveParams := keepalive.ServerParameters{
-		MaxConnectionIdle: time.Minute * 2,
+	// Configure server-side keepalive parameters. These settings allow the
+	// server to actively probe the connection health and ensure connections
+	// stay alive during idle periods.
+	serverKeepalive := keepalive.ServerParameters{
+		// Ping client after 1 minute of inactivity.
+		Time: time.Minute,
+		// Wait 20 seconds for ping response.
+		Timeout: 20 * time.Second,
+		// Allow connections to remain idle for extended periods. This
+		// is particularly important for RFQ operations where price
+		// oracle connections may be idle for long periods.
+		MaxConnectionIdle: time.Hour * 24,
 	}
 
-	serverOpts = append(serverOpts, grpc.KeepaliveParams(keepAliveParams))
+	// Configure client enforcement policy. This allows clients to send
+	// keepalive pings even when there are no active streams, which is
+	// crucial for long-lived connections with infrequent activity.
+	clientKeepalive := keepalive.EnforcementPolicy{
+		// Minimum time between client pings.
+		MinTime: 5 * time.Second,
+		// Allow pings without active RPCs.
+		PermitWithoutStream: true,
+	}
+
+	serverOpts = append(
+		serverOpts, grpc.KeepaliveParams(serverKeepalive),
+		grpc.KeepaliveEnforcementPolicy(clientKeepalive),
+	)
 
 	grpcServer := grpc.NewServer(serverOpts...)
 	defer grpcServer.Stop()


rfq/oracle.go
```diff
@@ -209,6 +225,21 @@ func insecureServerDialOpts() ([]grpc.DialOption, error) {
 		insecure.NewCredentials(),
 	))
 
+	// Configure client-side keepalive parameters. These settings ensure
+	// the client actively probes the connection health and prevents idle
+	// connections from being silently closed by intermediaries or the
+	// server.
+	opts = append(opts, grpc.WithKeepaliveParams(keepalive.ClientParameters{
```
darioAnongba: nit: Do we want this to be a pure copy-paste of the opts on lines 200-214?  Can't we reuse?

darioAnongba:
Failing tests are unrelated with the PR.

rfq/oracle.go
GeorgeTsagk: +1, let's extract some consts or the whole configuration and re-use it

server.go
```diff
@@ -370,11 +370,34 @@ func (s *Server) RunUntilShutdown(mainErrChan <-chan error) error {
...
-		MaxConnectionIdle: time.Minute * 2,
-	}
-
-	serverOpts = append(serverOpts, grpc.KeepaliveParams(keepAliveParams))
+	// Configure server-side keepalive parameters. These settings allow the
+	// server to actively probe the connection health and ensure connections
+	// stay alive during idle periods.
+	serverKeepalive := keepalive.ServerParameters{
+		// Ping client after 1 minute of inactivity.
+		Time: time.Minute,
```
GeorgeTsagk, line 378: isn't it redundant to also have the server pinging the client?

Why not just have it be the responsibility of the client

server.go
```diff
@@ -370,11 +370,34 @@ func (s *Server) RunUntilShutdown(mainErrChan <-chan error) error {
...
+	// keepalive pings even when there are no active streams, which is
+	// crucial for long-lived connections with infrequent activity.
+	clientKeepalive := keepalive.EnforcementPolicy{
+		// Minimum time between client pings.
+		MinTime: 5 * time.Second,
+		// Allow pings without active RPCs.
+		PermitWithoutStream: true,
+	}
+
+	serverOpts = append(
```
GeorgeTsagk, line 397: This server is the tapd rpcserver. The oracle server is actually not implemented here so this has been misplaced. We could add it to the reference implementation in `docs/examples/basic-price-oracle`.

server.go
Roasbeef: This mirrors what [lnd does](https://github.com/lightningnetwork/lnd/blob/87aed739e0cfffb95d05e8e274f9aae8fd25a503/lnd.go#L362-L372),and we havne't had any issues with it. I view it as defense in depth. Can you think of a reason where the server also pinging will somehow disrupt the connection?

server.go
Roasbeef: I don't think this is misplaced. The intent is that we resolve this on the `tapd` level as well, as we have other streaming calls, and no keep alive, so possible we run into this issue in other instances.

Roasbeef: I can add it to the example oracle as well.

Roasbeef:
> I understand the keepalive client side, but I'm wondering what is the impact on resources for the server side.

IIUC it's relative minor as the keep alive settings using the [`PING` frame on the HTTP/2 layer](https://datatracker.ietf.org/doc/html/rfc7540#section-6.7), which is usually 8 bytes. So if we ping every 60 seconds, then 8 bytes a minute, which is ~11 KB a day.

Roasbeef:
@GeorgeTsagk PTAL

Issue #1833, opened 2025-10-02
[feature]: Include `block_timestamp` in the `FetchSupplyCommit` RPC response


jamaljsr:
**Is your feature request related to a problem? Please describe.**

In Terminal, we are creating a page to display information about an asset group. We'll include the asset name, group key, genesis outpoint, etc. We'll also display a list of issuance/burn events along with the current total supply of the group. I plan to use the new `FetchSupplyCommit` to obtain most of this information. Currently, this RPC returns the block height for each issuance/burn, but not the block timestamp. To make the page more human-friendly, we'd like to display timestamps in the list of events instead of block heights.

**Describe the solution you'd like**

Include `block_timestamp` for each of the issuance/burn/ignore leaves in the `FetchSupplyCommit` response.

**Describe alternatives you've considered**

We could make separate API requests to fetch the timestamp of each block height in the list individually, but this would result in N+1 requests, which is not ideal from a performance perspective.

ffranr:
We will modify the supply leaves returned by `FetchSupplyLeaves` similarly. The relevant proto message `SupplyLeafEntry` is shared between both endpoints.

PR #1839, opened 2025-10-07
Add block header support to supply RPCs and related integration tests

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/1833

---

This PR introduces support for including block header data in supply-related RPC responses and updates integration tests accordingly.

### Motivation

These changes allow light clients that query these RPC endpoints to avoid performing additional network calls to obtain block timestamps. By including the block hash as well, clients can easily locate the full block or block header if further details are required.

### Summary of Changes

* **lndservices+tapgarden:**
  Added a `GetBlockHeaderByHeight` method to the chain bridge, allowing retrieval of a block header for a given block height.

* **supplycommit:**
  Added a helper function `ExtractSupplyLeavesBlockHeaders` to extract block headers from a set of supply leaves. This will be used to populate block header data in RPC responses.

* **universerpc:**

  * Extended `FetchSupplyLeavesResponse` with a map from block height to `SupplyLeafBlockHeader`, containing block header timestamps and hashes.
  * Extended `FetchSupplyCommitResponse` similarly to include the same block header mapping.

* **itest:**

  * Added a new test helper function `AssertSupplyLeafBlockHeaders` to verify that expected block header data is present in RPC responses.
  * Applied the helper to issuance and burn events.
  * Extended `testFetchSupplyLeaves` to handle ignored asset outpoints and assert the inclusion of relevant block data.

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -179,6 +179,15 @@
   configured with public read access. This matches the behavior of the  
   existing FetchSupplyCommit RPC endpoint.
 
+- [PR#1839](https://github.com/lightninglabs/taproot-assets/pull/1839) The
+  `FetchSupplyLeaves` and `FetchSupplyCommit` RPC endpoints now
+  include a new `block_headers` field. This field is a map from block
+  height to a `SupplyLeafBlockHeader` message, which provides the block
+  header timestamp (in seconds since the Unix epoch) and the 32-byte
+  block header hash. This allows clients to obtain block timing and hash
+  information directly from the RPC response without performing separate
+  blockchain queries.
+
 ## tapcli Additions
 
 - [Rename](https://github.com/lightninglabs/taproot-assets/pull/1682) the mint


ffranr:
I need to add release notes to this. Waiting on confirmation that this can go in 0.7

universe/supplycommit/util.go
```diff
@@ -163,3 +167,62 @@ func IsSupplySupported(ctx context.Context, assetLookup AssetLookup,
...
+// ExtractSupplyLeavesBlockHeaders is a helper method which extracts the block
+// headers from the supply leaves. The returned map is keyed by block height.
+func ExtractSupplyLeavesBlockHeaders(ctx context.Context,
+	chain tapgarden.ChainBridge,
+	supplyLeaves SupplyLeaves) (map[uint32]wire.BlockHeader, error) {
+
+	blockHeaders := make(map[uint32]wire.BlockHeader)
+
+	// Extract block headers from issuance leaves.
+	for idx := range supplyLeaves.IssuanceLeafEntries {
```
jtobin, line 180: Would these be better as follows?

```
for _, leaf := range supplyLeaves.IssuanceLeafEntries {
  ..
}
```

ffranr: I usually default to the `for idx := range` pattern because it makes the copy vs pointer decision explicit (`x := s[idx]` or `&s[idx]`), which helps avoid the loop-variable pointer bug and accidental in-place modification (internal reference could still cause issues). It also keeps `idx` available for setting conditional breakpoints or logging during debugging.

taprpc/universerpc/universe.proto
```diff
@@ -968,6 +976,10 @@ message FetchSupplyLeavesResponse {
     // Each entry corresponds to the key at the same index in
     // `ignore_leaf_keys`.
     repeated bytes ignore_leaf_inclusion_proofs = 6;
+
+    // A map of block height to supply leaf block header for all block heights
+    // referenced in supply leaves.
+    map<uint32, SupplyLeafBlockHeader> block_headers = 7;
```
Roasbeef, line 986: Why not just add this directly to `SupplyLeafEntry?` Then the caller doesn't need to map this map to the proper set of entries.

universe/supplycommit/util.go
```diff
@@ -163,3 +167,62 @@ func IsSupplySupported(ctx context.Context, assetLookup AssetLookup,
...
+				"block header from proof: %w", err)
+		}
+
+		blockHeaders[leaf.BlockHeight()] = blockHeader
+	}
+
+	// Extract block headers from burn leaves.
+	for idx := range supplyLeaves.BurnLeafEntries {
+		leaf := supplyLeaves.BurnLeafEntries[idx]
+		blockHeader := leaf.BurnProof.BlockHeader
```
GeorgeTsagk, line 206: missing the broader context, but are we sure this is always present and not nil?

ffranr: we do expect the burn proof to be non-nil. I've added a check to ensure that we don't try to deref nil pointer:
```
		if leaf.BurnProof == nil {
			return nil, fmt.Errorf("burn proof is nil for burn " +
				"leaf")
		}
```

jtobin (approved):
LGTM. :+1: :+1:

PR #1847, opened 2025-10-10
build: update CI+release version to Go 1.25.2

Roasbeef:
We also update the version pinned in the go.mod files to Go 1.24.8.


taprpc/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/lightninglabs/taproot-assets/taprpc
 
-go 1.24.6
+go 1.24.8
```
ffranr: Should we bump this to `1.25.2` also? The linter might fail.
Same for `tools/go.mod`

docs/examples/basic-price-oracle/go.mod
```diff
@@ -1,6 +1,6 @@
 module basic-price-oracle
 
-go 1.24.6
+go 1.24.8
```
ffranr: Should this be 1.25.2 also?

taprpc/go.mod
Roasbeef: We typically make sure we support building with the last two Go versions. Changing this would mean you can't build with Go 1.24.

tools/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/lightninglabs/taproot-assets/tools
 
-go 1.24.6
+go 1.24.8
```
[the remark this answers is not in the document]
darioAnongba: good point here for the tools.

litbot-9000:

docs/examples/basic-price-oracle/go.mod
Roasbeef: We want to be able to build with Go 1.24 still.

Roasbeef:
Pushed up a new version.

Roasbeef:
Updated Aperture to fix a cert validation issue. See https://github.com/lightninglabs/aperture/pull/198

Issue #1840, opened 2025-10-07
[feature]: Add  reorg-aware block header cache to `LndRpcChainBridge`


ffranr:
`LndRpcChainBridge` (`lndservices/chain_bridge.go`) currently caches only block timestamps by height via `blockTimestampCache *lru.Cache[uint32, cacheableTimestamp]`. That cache can return timestamps for headers that were reorged out since it is unaware of chain reorgs and confirmation depth. We should replace this with a reorg-aware cache of full block headers.

**Goals**

* Cache full headers, not just timestamps. Header hash and timestamp are derivable from the header.
* Make `GetBlockTimestamp` read from the new header cache. Remove or retire `blockTimestampCache`.
* Implement as a reusable, package-level component with clean interfaces, suitable for use beyond `LndRpcChainBridge`.

**Header cache requirements**

* Size is configurable, default target 100k entries.
* On reaching capacity, purge a random fraction (default 10 percent). Fraction is configurable.
* Track confirmation depth. Parameter `numberOfConfirmations` with typical value 6.
* Store headers even when depth `< numberOfConfirmations`, but treat them as unsettled:

  * Queries for unsettled heights or hashes return a cache miss to force an external lookup.
  * Once a later settled height is observed that implies sufficient depth, mark prior entries as settled.
* Support lookup by height and by header hash.
* Be resilient to reorgs:

  * Detect conflicting headers at the same height.
  * On detection, invalidate the conflicting branch entries that are now orphaned.

**Testing**

* Unit tests that cover:

  * Basic put and get by height and by hash.
  * Capacity and random purge behavior.
  * Unsettled vs settled semantics with `numberOfConfirmations`.
  * Reorg scenario: insert headers for H..H+N, then replace H+k with an alternate header and ensure orphaned entries are invalidated and queries miss until settled.
  * `GetBlockTimestamp` integration against the new cache.

ffranr:
PR has been merged into [0-8-0-staging](https://github.com/lightninglabs/taproot-assets/tree/0-8-0-staging)

PR #1849, opened 2025-10-13
lndservices: add reorg aware block header cache; use in ChainBridge

ffranr:
Closes: https://github.com/lightninglabs/taproot-assets/issues/1840

---

**What**

* Introduce reusable full block header cache (by height and hash).
* Integrate cache in `LndRpcChainBridge`.
* Refactor `GetBlockTimestamp` to read via `GetBlockHeaderByHeight` and drop the old timestamp cache.

**Why**

* **Avoid stale timestamps from reorged headers.**
* Reduce repeated header RPCs.
* Provide a shared, configurable cache for future callers.

**Highlights**

* Reorg detection with conflicting height handling and orphan invalidation.
* Unsettled vs settled semantics via confirmation depth.
* Configurable capacity and random purge fraction.

**Tests**

* Basic get and put, capacity and purge, settled semantics, reorg path.

### DIFF
--- a/lndservices/chain_bridge.go
+++ b/lndservices/chain_bridge.go
@@ -42,17 +42,22 @@ type LndRpcChainBridge struct {
 
 	// assetStore is a handle to the asset store.
 	assetStore *tapdb.AssetStore
+
+	// headerCache is a cache for block headers to reduce RPC calls.
+	headerCache *BlockHeaderCache
 }
 
 // NewLndRpcChainBridge creates a new chain bridge from an active lnd services
 // client.
 func NewLndRpcChainBridge(lnd *lndclient.LndServices,
-	assetStore *tapdb.AssetStore) *LndRpcChainBridge {
+	assetStore *tapdb.AssetStore,
+	headerCache *BlockHeaderCache) *LndRpcChainBridge {
 
 	return &LndRpcChainBridge{
 		lnd:         lnd,
 		retryConfig: fn.DefaultRetryConfig(),
 		assetStore:  assetStore,
+		headerCache: headerCache,
 	}
 }
 
@@ -118,6 +123,11 @@ func (l *LndRpcChainBridge) GetBlock(ctx context.Context,
 func (l *LndRpcChainBridge) GetBlockHeader(ctx context.Context,
 	hash chainhash.Hash) (*wire.BlockHeader, error) {
 
+	// First, check the cache for the requested block header.
+	if header, ok := l.headerCache.GetByHash(hash); ok {
+		return &header, nil
+	}
+
 	return fn.RetryFuncN(
 		ctx, l.retryConfig, func() (*wire.BlockHeader, error) {
 			header, err := l.lnd.ChainKit.GetBlockHeader(ctx, hash)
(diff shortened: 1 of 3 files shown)


darioAnongba (approved):
Code looks good to me. Only nit comment is that we should slow down on the comments that add no value, example:

```
// Calculate the height at which this block would be considered settled.
settledHeight := height + c.cfg.NumConfSettled
```
Here the code has actually more information in it than the comment IMO.

lndservices/block_header_cache.go
```diff
@@ -0,0 +1,317 @@
...
+		byHeight: make(map[uint32]*headerEntry),
+		byHash:   make(map[chainhash.Hash]*headerEntry),
+	}, nil
+}
+
+// isSettled returns whether an entry is considered settled based on
+// confirmations.
+func (c *BlockHeaderCache) isSettled(height uint32) bool {
+	// Calculate the height at which this block would be considered settled.
+	settledHeight := height + c.cfg.NumConfSettled
```
jtobin: Nit: I would actually change NumConfSettled to NumBlockSettled or similar, and ditto for the other terminology around confirmations. In my head *transactions* map to confirmations, and blocks to height/depth. But this might be overly pedantic (or idiosyncratic), since I imagine everyone understands that a block is "confirmed" by the number of blocks on top of it.

lndservices/block_header_cache.go
```diff
@@ -0,0 +1,317 @@
...
+	if height > c.maxHeight {
+		c.maxHeight = height
+	}
+
+	return nil
+}
+
+// GetByHeight retrieves a block header by height. Returns nil if not found or
+// if the entry is unsettled (to force external lookup).
+func (c *BlockHeaderCache) GetByHeight(height uint32) *wire.BlockHeader {
```
jtobin: Is there a specific reason to return a pointer here, rather than e.g. a (value, bool) pair à la lnutils.SyncMap's Load?

lndservices/block_header_cache.go
```diff
@@ -0,0 +1,316 @@
...
+	if !exists || !c.isSettled(height) {
+		return nil
+	}
+
+	return &entry.header
+}
+
+// GetByHash retrieves a block header by hash. Returns nil if not found or if
+// the entry is unsettled (to force external lookup).
+func (c *BlockHeaderCache) GetByHash(hash chainhash.Hash) *wire.BlockHeader {
```
jtobin: (Same question about returning a pointer, rather than a value.)

lndservices/block_header_cache.go
ffranr: I think `SyncMap.Load` returns a pair because it can’t rely on recognizing a zero value as “not present,” since the map stores values of type `any`, which might not be pointers. In contrast, `BlockHeaderCache.GetByHeight(height uint32)` returns a `*wire.BlockHeader`, so that ambiguity doesn’t exist. However, I try to avoid unnecessary pointers, so I’ve changed it to return a value instead. Same for `GetByHash`.

lndservices/block_header_cache.go
ffranr: see https://github.com/lightninglabs/taproot-assets/pull/1849#discussion_r2436254792

lndservices/block_header_cache.go
ffranr: I've gone with `MinSettledBlockDepth` in the latest commit set, let me know what you think.

Issue #1850, opened 2025-10-15
[bug]: Group key send: Assets never show up at recipient's node


Liongrass:
### Background

I minted a new asset in three tranches.
I sent a portion of the newly minted assets to another tapd node.
The send transaction succeeds, but the assets are not available on the recipient's node.

### Your environment

Sender tapd:
`0.7.0-alpha.rc1 commit=v0.7.0-rc1-149-ge6ae082c`

Recipient tapd:
`0.7.0-alpha.rc1 commit=v0.7.0-rc1-149-ge6ae082c`

Universe tapd:
`0.7.0-alpha.rc1 commit=v0.7.0-rc1`

### Steps to reproduce

Tell us how to reproduce this issue. Please provide stacktraces and links to code in question.

On the sender's node:

`tapcli assets mint --type normal --name eightee --supply 1000000 --decimal_display 3 --meta_file_path eight.json --meta_type json --new_grouped_asset`
`tapcli assets mint --type normal --name eightee --supply 1000000 --decimal_display 3 --grouped_asset --meta_file_path eight.json --meta_type json --group_key 036138b75291d094416a6c6905c3f78446785ae6f498373418d1d103a53d90cde3`

On the recipient's node:

`tapcli addrs new --group_key 036138b75291d094416a6c6905c3f78446785ae6f498373418d1d103a53d90cde3`

Then again on the sender's node:

`tapcli assets send --addr_with_amount taptb1qqqsyqspqqzjzqmp8zm49ywsj3qk5mrfqhpl0pzx0pdwdaycxu6p35w3qwjnmyxduvrzzqu6paynlc88ddypzvyyzhlxmvuc0zug6wsqwlxqv8nsv0jc74mnfuyzzq3xygecfjvqq992jryyce6gqfk2dfxhy5argtx4qsad7ds4ha2tz59qzqqv89sh2argd4skjmrzdauzkatwd9mx2unnv4e8qce69uhh2mnfwejhyum99eekjemwv46zumrpd9ek2efwdaexww3cxs6rxe287ht:2500000`

[2025-10-14_tapd-universe.log](https://github.com/user-attachments/files/22915341/2025-10-14_tapd-universe.log)
[2025-10-14_tapd-receiver.log](https://github.com/user-attachments/files/22915340/2025-10-14_tapd-receiver.log)
[2025-10-14_tapd-sender.log](https://github.com/user-attachments/files/22915339/2025-10-14_tapd-sender.log)

`

[...]

PR #1854, opened 2025-10-17
rpcserver: NewAddr universal courier connectivity check, offline skip, and local asset precheck

ffranr:
Related to: https://github.com/lightninglabs/taproot-assets/issues/1850#issuecomment-3414674425

Adds a universal proof courier connectivity check for all tap address types when the courier URL uses a universe RPC scheme. Introduces a flag to skip the connectivity check for offline address generation and adds an early sanity check to ensure the asset is known locally before attempting courier connection.

### DIFF
--- a/docs/release-notes/release-notes-0.7.0.md
+++ b/docs/release-notes/release-notes-0.7.0.md
@@ -271,6 +271,12 @@
 
 ## RPC Updates
 
+- [PR](https://github.com/lightninglabs/taproot-assets/pull/1854) The `NewAddr`
+  RPC now performs a proof courier connectivity check for all tap address types
+  using a universe RPC URL scheme. A new flag allows skipping this check for
+  offline address generation. Additionally, an early sanity check ensures the
+  asset is recognized locally before attempting any courier connection check.
+
 ## tapcli Updates
 
 - The default script key type in the `tapcli assets list`,


rpcserver.go
```diff
@@ -1662,6 +1662,12 @@ func (r *rpcServer) NewAddr(ctx context.Context,
 			err)
 	}
 
+	// Sanity check early to ensure that the asset is known.
+	_, err = r.cfg.AddrBook.QueryAssetInfo(ctx, specifier)
```
darioAnongba: Should this be smarter and attempt to identify the asset automatically? No idea if that makes sense and out of scope of this PR.

rpcserver.go
```diff
@@ -1698,40 +1704,49 @@ func (r *rpcServer) NewAddr(ctx context.Context,
...
 	switch {
 	case addrVersion == address.V2 &&
-		protocol == proof.UniverseRpcCourierType:
+		courierAddr.Scheme == proof.UniverseRpcCourierType:
 
 		// We assume that any proof courier running an RPC universe
 		// server will upgrade soon to support the new auth mailbox
 		// courier type, so we bump the protocol/scheme to the combined
 		// one.
 		courierAddr.Scheme = proof.AuthMailboxUniRpcCourierType
```
darioAnongba, line 1717: Is this useless now? Since you are doing 
```
connCheckScheme := courierAddr.Scheme == proof.UniverseRpcCourierType ||
		courierAddr.Scheme == proof.AuthMailboxUniRpcCourierType
```
below

rpcserver.go
```diff
@@ -1698,40 +1704,49 @@ func (r *rpcServer) NewAddr(ctx context.Context,
...
 			address.ErrInvalidProofCourierAddr, addrVersion,
 			proof.AuthMailboxUniRpcCourierType)
 	}
 
+	// If the courier is an (auth mailbox) universe RPC type we will check
+	// that we can connect to it.
+	connCheckScheme := courierAddr.Scheme == proof.UniverseRpcCourierType ||
+		courierAddr.Scheme == proof.AuthMailboxUniRpcCourierType
+	if !req.SkipProofCourierConnCheck && connCheckScheme {
+		rpcsLog.InfoS(ctx, "Checking connection to universe server RPC",
```
darioAnongba: Not an info log IMO
```suggestion
		rpcsLog.DebugS(ctx, "Checking connection to universe server RPC",
```
generating a new address should be 1 line in the logs like:
`new address generated successfully`. Everything else is debug or trace.

darioAnongba (approved):
LGTM, nit only.

rpcserver.go
```diff
@@ -1698,40 +1704,49 @@ func (r *rpcServer) NewAddr(ctx context.Context,
 		"courier=%v, addr_version=%v, asset_version=%v", &specifier,
 		req.Amt, courierAddr, addrVersion, assetVersion)
 
-	// Addresses with version 2 must use the new authmailbox proof courier
+	// Addresses with version 2 must use the new auth mailbox proof courier
 	// type.
-	protocol := courierAddr.Scheme
 	switch {
```
darioAnongba, line 1709: Is a `switch` not necessary here?
I only see the need for an 
```go
if addrVersion == address.V2 &&
		courierAddr.Scheme == proof.HashmailCourierType {

return nil, fmt.Errorf("%w: address version %d must use the "+
			"'%s' proof courier type",
			address.ErrInvalidProofCourierAddr, addrVersion,
			proof.AuthMailboxUniRpcCourierType)
}
```

rpcserver.go
ffranr: it is still useful since the field modification will be stored in the db

rpcserver.go
ffranr: i think this still matter?

```
	case addrVersion == address.V2 &&
		courierAddr.Scheme == proof.UniverseRpcCourierType:

		// We assume that any proof courier running an RPC universe
		// server will upgrade soon to support the new auth mailbox
		// courier type, so we bump the protocol/scheme to the combined
		// one.
		courierAddr.Scheme = proof.AuthMailboxUniRpcCourierType
```

ffranr:
I will merge since these are nits, you've approved, and I think I'm correct. And this is blocking release.

PR #1855, opened 2025-10-17
ci: run CI on *-staging branches in addition to main

ffranr:
Update push and merge_group triggers to include *-staging.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -4,12 +4,14 @@ on:
   push:
     branches:
       - 'main'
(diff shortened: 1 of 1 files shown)


darioAnongba (approved):
I think this is fine to be merged into `0-8-0-staging` but we should not port this to `main` because we are trying to move away from this type of release flow, to a standard trunk based one so we shouldn't have staging branches anymore.

jtobin (approved):

litbot-9000:

PR #1857, opened 2025-10-24
multi: make anchor tx version configurable and validate v3 with bitcoind

bhandras:
## Summary
This PR changes anchor transaction version handling from a blanket default to explicit configuration on the flows that actually create anchor transactions.

The daemon still defaults to v2. Callers can now explicitly request v3 for newly created anchor transactions in the minting, send, supply-commit, and PSBT anchoring paths.

## What Changed
- Add anchor transaction version plumbing in `tapsend`, `tapgarden`, `tapfreighter`, and `universe/supplycommit`, while keeping v2 as the default.
- Expose the anchor transaction version on the RPCs that construct anchor transactions and regenerate the protobuf artifacts and related test vectors.
- Bump `lnd` to the latest master and update the itest harnesses and helpers to the current `lntest` miner APIs.
- Use the new bitcoind miner support in the integration harnesses, install `bitcoind` in CI, and add a dedicated bitcoind-backed `itest-v3` target and workflow job.
- Add end-to-end v3 anchor coverage that verifies the mined Bitcoin transactions are actually version 3 on chain.
- Update the mock gRPC TLS helpers to advertise `h2` so the courier-related tests keep passing with the newer gRPC stack pulled in by the `lnd` bump.

## Why
- Keep the existing behavior stable by leaving v2 as the default anchor transaction version.
- Make v3 available only where new anchor transactions are actually constructed.
- Validate the v3 path against a bitcoind-backed integration lane instead of only relying on unit-level coverage.

## Testing
- `go test ./itest/...`
- `make itest-v3`
- `go test ./proof ./tapchannel`
- `go test ./...`

### DIFF
--- a/.github/actions/setup-bitcoind/action.yml
+++ b/.github/actions/setup-bitcoind/action.yml
@@ -1,0 +1,34 @@
+name: "Setup bitcoind"
+description: "Install a pinned Bitcoin Core release for integration tests"
+inputs:
+  version:
+    description: "Bitcoin Core version to install"
+    required: false
+    default: "30.2"
+
+runs:
+  using: "composite"
+
+  steps:
+    - name: install bitcoind ${{ inputs.version }}
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        if command -v bitcoind >/dev/null 2>&1; then
+          bitcoind --version
+          exit 0
+        fi
+
+        VERSION="${{ inputs.version }}"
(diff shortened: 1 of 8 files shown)


litbot-9000:
@bhandras, remember to re-request review from reviewers when ready

bhandras:
Picking this up again now that lntest supports bitcoind miner.

Issue #1362, opened 2025-02-07
[feature]: `PortfolioPilot` - Outsourcing All Market and Portfolio-Related Decision-Making During RFQ


ffranr:
Currently, the RFQ `PriceOracle` provides asset-to-BTC exchange rates, serving both wallet nodes and edge nodes. A key benefit of this setup is that tapd does not need to manage pricing decisions directly—these can be offloaded to an external service.  

I propose extending this approach to limit orders as well. By limit order, I mean configurable rate and volume thresholds that an edge node can set to further restrict asset channel usage. While channel capacity already imposes constraints, limit orders add an additional layer of control. In our current implementation, these are referred to as "offers."  

To enhance flexibility, all market, portfolio, and financial decision-making—not just price queries—should be handled by an external service. This is particularly important for edge nodes, as they require more nuanced control over pricing, limits, and execution policies.  

### Introducing `PortfolioPilot`  
To support this, I think we should introduce a more generalized service, which I will call `PortfolioPilot`.  

#### Proposed `PortfolioPilot` RPC Endpoints  
- `QueryAssetRates` – Returns exchange rates (migrated from or integrated with the price oracle).  
- `IsQuoteAcceptable` – Evaluates whether a peer's proposed quote meets acceptable conditions. If `PortfolioPilot` handles quote validation, it could also maintain an internal record of accepted/rejected quotes.  
- `QueryAssetChannelLimits` – Returns the maximum allowable volume for a given asset and channel.  
- `PaymentUpdate` – Updates `PortfolioPilot` on whether a quote was used, ensuring it remains synchronized with executed transactions.  

By consolidating these responsibilities into `PortfolioPilot`, we can create a dedicated service for market intelligence, portfolio management, and financial decision-making, keeping `tapd` focused on its core role.

---

##### Notes

And then, if we go down this route, we can just remove RPC endpoints `AddAssetSellOffer` and `AddAssetBuyOffer`.

This sort of change should also allow us to get rid of price tolerance checks:
https://github.com/lightninglabs/taproot-assets/blob/303f3dac0f8e51ce3d153033d1841603021bfc97/rfq/negotiator.go#L772-L774

It basically shrinks tapd and moves market/portfolio related decisions to a new service.

ffranr:
Of course, we could default to a built-in `PortfolioPilot` that enforces the same price threshold check we currently use, while also allowing the option to specify an external RPC `PortfolioPilot`.

jcbrand:
[...]

Issue #1005, opened 2024-07-10
[feature]: Provide "forwarding history" for Edge Nodes


MegalithicBTC:
Currently, when an Edge Node performs a swap between a Taproot Asset and Bitcoin, my understanding is that the only way to see the "results" or "history" of this transaction is with a combination of these commands:

```
lncli fwdinghistory

{
    "forwarding_events": [
        {
            "timestamp": "1720647207",
            "chan_id_in": "3152629690323894272",
            "chan_id_out": "3152616496185344000",
            "amt_in": "101",
            "amt_out": "100",
            "fee": "1",
            "fee_msat": "1000",
            "amt_in_msat": "101000",
            "amt_out_msat": "100000",
            "timestamp_ns": "1720647207538098949",
            "peer_alias_in": "testnet.laisee.org",
            "peer_alias_out": "VelasCommerce-Testnet"
        }
    ],
    "last_offset_index": 1
}

```

AND.....

```
lncli listchannels

{
    "channels": [
        {
            "active": true,
            "remote_pubkey": "03acb2b721d403396a5af53ee84ec0728b8057f5e86124359d0063c5f34bc0223d",
            "channel_point": "c321db4ed04140608d1475bb92a97a0889fba74b898415a017154b5df53bd732:0",
            "chan_id": "3152629690323894272",
            "capacity": "500000",
            "local_balance": "396631",
            "remote_balance": "99899",
```

The `fwdinghistory` shows the event, and the results of that event are then shown in the change in the `local_balance` vs `remote_balance`.

Maybe some kind of extra database table or other structure is needed so the Edge Node can "keep records" on their activity?

Roasbeef:
Great idea! We definitely have some visibility/accounting gaps to fill.

darioAnongba:
Blocked for now before we introduce quotes storage. As per discussion with @GeorgeTsagk. Issue postponed.

ffranr:
Potential pre-requisite issue: https://github.com/lightninglabs/taproot-assets/issues/855

PR #1863, opened 2025-11-03
Persist agreed quotes in RFQ service

### LINKED ISSUE #855: [feature]: Persist agreed quotes in RFQ service
**Objective:** Improve the RFQ (Request for Quote) service by enabling it to persist agreed quotes with peers as asset sale/purchase policies in an SQL table, and populate its state from this table upon startup.

**Background:**
- **Current Behavior:** The RFQ service facilitates the exchange of quotes between peers but does not persist agreed quotes for future reference or state initialization.
- **Proposed Enhancement:** Once a quote is agreed upon with a peer, it is represented in the "order handler" RFQ sub-service as an asset sale/purchase policy. This policy should then be written to an SQL table for persistence. Furthermore, upon startup, the RFQ service should read from this SQL table to populate its state/local cache.

**Benefits:**
- **Audit and Verification:** Persistent storage of agreed quotes as policies in an SQL database will facilitate auditing and verification processes. We currently delete expired quotes without record keeping, persistence will provide a record of those quotes.
- **Service Reliability and Usability:** By populating the RFQ service's state from the SQL table on startup, the service can ensure continuity and reliability after reboot, enhancing overall usability.

**Implementation Considerations:**
- **SQL Table Design:** Design the SQL table schema to efficiently store asset sale/purchase policies, including fields for quote details, agreement timestamps, and identifiers of the agreeing parties.
- **Data Persistence Mechanism:** Implement functionality in the "order handler" sub-service to write to and read from the SQL table via a quote "store".
- **Startup Procedure:** Modify the RFQ service startup sequence to include reading from the SQL table and populating the service's local cache or state with the stored policies.

darioAnongba:
RFQ buy/sell accepts are now written to the database (`rfq_policies` table) whenever a policy is agreed, giving us an audit trail and keeping quotes across restarts.

The `OrderHandler` reloads these rows at startup to rebuild policies and the manager repopulates its in-memory caches from the same data.

Added a check in `itest` that restarts Bob’s `tapd` mid RFQ flow and verifies the quote remains available to Carol.

Fixes #855

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -45,6 +45,10 @@
   instance has a fully functional RFQ pipeline and surfaces configuration or
   lnd-level conflicts immediately.
 
+- [RFQ buy/sell accepts are now written to the database](https://github.com/lightninglabs/taproot-assets/pull/1863)
+  `rfq_policies` table whenever a policy is agreed, giving us an audit trail
+  and keeping quotes alive across restarts.
+
 ## RPC Updates
 
 - [PR#1841](https://github.com/lightninglabs/taproot-assets/pull/1841): Remove


tapdb/rfq_policies.go
```diff
@@ -0,0 +1,597 @@
...
+			log.Errorf("Unable to deactivate expired policy "+
+				"id=%x: %v", id[:], err)
+		}
+	}
+
+	return buyAccepts, sellAccepts, nil
+}
+
+// DeactivatePolicy marks the given RFQ policy as inactive.
+func (s *PersistedPolicyStore) DeactivatePolicy(ctx context.Context,
```
GeorgeTsagk: why do we need for this to be a db query?

We can retrieve active quotes by validating the expiry field. This way we won't have to store an explicit `active` flag.

While the daemon is running, quotes which expire within the runtime can be deleted directly in-memory without any db calls. On the next start up, they will never be loaded again.

rfq/manager.go
```diff
@@ -264,12 +267,18 @@ func (m *Manager) startSubsystems(ctx context.Context) error {
...
 	})
 	if err != nil {
 		return fmt.Errorf("error initializing RFQ order handler: %w",
 			err)
 	}
 
+	if err := m.restorePersistedPolicies(ctx); err != nil {
+		return fmt.Errorf("error restoring persisted RFQ "+
+			"policies: %w", err)
+	}
```
ffranr: I think this should happen in `m.orderHandler.Start`. Effectively, I think `PolicyStore` should only be accessible by `OrderHandler`.

rfq/manager.go
```diff
@@ -313,6 +322,29 @@ func (m *Manager) startSubsystems(ctx context.Context) error {
...
 
+// restorePersistedPolicies loads persisted RFQ policies from the store
+// and rehydrates the order handler and local cache state.
+func (m *Manager) restorePersistedPolicies(ctx context.Context) error {
+	buyAccepts, sellAccepts, err := m.cfg.PolicyStore.FetchActivePolicies(
+		ctx,
+	)
+	if err != nil {
+		return fmt.Errorf("error fetching persisted policies: %w", err)
+	}
```
ffranr: method `m.cfg.PolicyStore.FetchActivePolicies` should be called `FetchLocalAcceptedQuotes` or something like that?

ffranr: I think even `m.localAcceptedBuyQuotes` shouldn't be in the `manager`. But maybe we can fix that later.

rfq/manager_test.go
```diff
@@ -59,6 +61,32 @@ var (
 	peer2        = route.Vertex{77}
 )
 
+type nopPolicyStore struct{}
```
ffranr: we should call this `mockPolicyStore` to follow the existing pattern elsewhere

rfq/order.go
```diff
@@ -978,9 +1010,72 @@ func (h *OrderHandler) RegisterAssetPurchasePolicy(
...
+	ctx, cancel := h.WithCtxQuit()
+	defer cancel()
+
+	err := h.policyStore.DeactivatePolicy(
+		ctx, id, time.Now().UTC(),
+	)
+	if err != nil {
+		log.Errorf("Unable to deactivate persisted policy id=%x: %v",
+			id[:], err)
+	}
```
ffranr: this should return the error and the methods here need doc comments

rfq/order.go
```diff
@@ -978,9 +1010,72 @@ func (h *OrderHandler) RegisterAssetPurchasePolicy(
...
+		policy := NewAssetPurchasePolicy(accept)
+		h.policies.Store(policy.scid, policy)
+	}
+}
+
+func (h *OrderHandler) removePolicy(scid SerialisedScid,
+	policy Policy) {
+
+	h.policies.Delete(scid)
+	h.deactivatePersistedPolicy(policy)
```
ffranr: we should update the store and check for error in doing so before deleating from the local `h.policies`

tapdb/rfq_policies.go
darioAnongba: As far as I understand, it is possible to get new quotes for the same `(SCID, policy_type)` (where `policy_type` is either sale or purchase). In which case we need to deactivate (aka soft delete) the old quotes so we only have a single active quote per `(SCID, policy_type)`. Does this make sense?

rfq/manager.go
darioAnongba: I tried not to refactor the current architecture but I agree the current storage in the manager is not ideal and I wanted to change that as well.
We are currently storing data directly inside the `Manager` using a bunch of `SyncMap` as an in-memory storage/cache. `OrderHandler` also has its own in-memory storage and now we are adding persistent storage in the form of a store (`PolicyStore`), but all storage should be encapsulated away and injected. The `Manager` or `OrderHandler` don't care whether we are using in-memory, an external DB or an hybrid.

IMO the solution is to move the entirety of the storage to one or multiple stores that are injected into the corresponding components so we can get rid of all the `SyncMap` and expose interfaces. We can keep the `subscribers` and `groupKeyLookupCache` in the manager to avoid over-engineering. 

It's out of the scope of this PR but I can work on refactoring this if needed.

rfq/manager.go
darioAnongba: I think we have the same idea but this is currently not possible with the current architecture (unless we do 2 DB requests) , see my comment below. This is because `localAcceptedBuyQuotes` and `localAcceptedSellQuotes` are in the `Manager` so to populate both the `Manager` and `OrderHandler` we need to fetch the policies

IMO we should have a single store that handles storage (in-memory + DB) so the entire notion of "restoring" is hidden to the business (`Manager` and `OrderHandler`) and happens automatically when the store is instantiated. The Store is then injected in the business and we interact with it simply by calling exposed methods without knowing how the data is stored.

Happy to refactor this.

rfq/order.go
darioAnongba: Applied. Not sure if you meant only `deactivatePersistedPolicy` or all store calls but added error handling to all. We now bubble up the errors until the main event loop, that we never exit.

rfq/order.go
darioAnongba: Applied. Not sure if you meant only `deactivatePersistedPolicy` or all store calls but added error handling to all. We now bubble up the errors until the main event loop, that we never exit.

rfq/manager.go
darioAnongba: As discussed, I moved all the in-memory quotes away from the Manager and into the `OrderHandler` for now so it's the only component accessing the PolicyStore in `Start`. Since you mentioned not wanting too many functions exposed in the interface, the manager still accesses the quotes with (example):
```
m.orderHandler.peerAcceptedBuyQuotes
```
I can wrap this in a getter if needed but seemed fine for now like this since this PR should not be too much of a refactor.

rfq/manager.go
darioAnongba: As discussed, I moved all the in-memory quotes away from the Manager and into the `OrderHandler` for now so it's the only component accessing the PolicyStore in `Start`. Since you mentioned not wanting too many functions exposed in the interface, the manager still accesses the quotes with (example):
```
m.orderHandler.peerAcceptedBuyQuotes
```
I can wrap this in a getter if needed but seemed fine for now like this since this PR should not be too much of a refactor.

tapdb/rfq_policies.go
darioAnongba: After discussion we established that having multiple quotes per SCID, policy_type is not possible so we can simply fetch the non-expired quotes and assume that they are active.

rfq/interface.go
```diff
@@ -0,0 +1,20 @@
+package rfq
```
GeorgeTsagk, line 1: nit: file should be named `policy_store.go`

rfq/order.go
```diff
@@ -1106,7 +1255,7 @@ func (h *OrderHandler) fetchPolicy(htlc lndclient.InterceptedHtlc) (Policy,
 }
 
 // cleanupStalePolicies removes expired policies from the local cache.
-func (h *OrderHandler) cleanupStalePolicies() {
+func (h *OrderHandler) cleanupStalePolicies() error {
```
GeorgeTsagk: you only seem to return `nil` here, why not remove `error`?

rfq/manager.go
```diff
@@ -435,7 +407,7 @@ func (m *Manager) handleIncomingMessage(incomingMsg rfqmsg.IncomingMsg) error {
 			// quote so that it can be used to send a payment by our
 			// lightning node.
 			scid := msg.ShortChannelId()
-			m.peerAcceptedBuyQuotes.Store(scid, msg)
+			m.orderHandler.peerAcceptedBuyQuotes.Store(scid, msg)
```
GeorgeTsagk: why are we moving the maps to `orderHandler` if we end up calling them again from `rfq.Manager`?

rfq/order.go
```diff
@@ -736,8 +790,13 @@ type OrderHandler struct {
...
 	return &OrderHandler{
-		cfg:      cfg,
-		policies: lnutils.SyncMap[SerialisedScid, Policy]{},
+		cfg:         cfg,
+		policyStore: cfg.PolicyStore,
+		policies:    lnutils.SyncMap[SerialisedScid, Policy]{},
+		peerAcceptedBuyQuotes: lnutils.SyncMap[
+			SerialisedScid, rfqmsg.BuyAccept]{},
+		peerAcceptedSellQuotes: lnutils.SyncMap[
+			SerialisedScid, rfqmsg.SellAccept]{},
```
GeorgeTsagk: why are we not initializing `localAcceptedBuyQuotes` / `localAcceptedSellQuotes` ?

GeorgeTsagk: actually I think the `lnutils.SyncMap` type doesn't need to be initialized, in that case should remove above lines

rfq/order.go
```diff
@@ -848,7 +907,13 @@ func (h *OrderHandler) mainEventLoop() {
 		case <-cleanupTicker.C:
 			log.Debug("Cleaning up any stale policy from the " +
 				"order handler")
-			h.cleanupStalePolicies()
+
+			err := h.cleanupStalePolicies()
+			if err != nil {
+				log.Errorf("error in main event loop: %w", err)
```
GeorgeTsagk: nit: line break after log

tapdb/rfq_policies.go
```diff
@@ -0,0 +1,534 @@
...
+	AgreedAt              time.Time
+}
+
+// RfqPolicyStore is the database interface for RFQ policies.
+type RfqPolicyStore interface {
+	InsertRfqPolicy(context.Context,
+		sqlc.InsertRfqPolicyParams) (int64, error)
+
+	FetchActiveRfqPolicies(context.Context) (
+		[]sqlc.RfqPolicy, error)
```
GeorgeTsagk: fits in one line

rfq/order.go
darioAnongba: Thanks, removed. It was a residue of the code when we deactivated in DB.

rfq/order.go
darioAnongba: Thanks, this was moved from the manager but indeed we don't need it, so good to remove.

rfq/manager.go
darioAnongba: The reason is that all these maps represent an in-memory store. The business layer does not care how the data is stored (it can be an in-memory cache, a Redis cache, a DB, etc). And we should ideally be able to switch between infras without changing the business.

We currently split these maps into the `Manager` and the `OrderHandler`. The idea we discussed with @ffranr is to move everything to the `OrderHandler` for now in a simple refactor in this PR, and then eventually move everything to `tapdb` (in a `QuoteStore` or `PolicyStore`).

Now, about this line: `m.orderHandler.peerAcceptedBuyQuotes.Store(scid, msg)`. I agree it's ugly and I also don't like it so I could expose functions from the orderHandler like `m.orderHandler.StoreBuyQuote()`, equivalent to the future `m.quoteStore.StoreBuyQuote()` WDYT?

rfq/manager.go
```diff
@@ -522,16 +494,16 @@ func (m *Manager) handleOutgoingMessage(outgoingMsg rfqmsg.OutgoingMsg) error {
...
-		m.orderHandler.RegisterAssetSalePolicy(*msg)
-
-		// We want to store that we accepted the buy quote, in case we
-		// need to look it up for a direct peer payment.
-		m.localAcceptedBuyQuotes.Store(msg.ShortChannelId(), *msg)
+		err := m.orderHandler.RegisterAssetSalePolicy(*msg)
+		if err != nil {
+			return fmt.Errorf("error registering asset sale "+
+				"policy: %w", err)
+		}
```
ffranr: unnecassary word `error` here: `fmt.Errorf("error`, and similarly below

rfq/order.go
```diff
@@ -957,28 +1019,106 @@ func (h *OrderHandler) Start() error {
...
+
+// storePurchasePolicy stores a purchase policy in the policy store.
+func (h *OrderHandler) storePurchasePolicy(
+	sellAccept rfqmsg.SellAccept) error {
+
+	ctx, cancel := h.WithCtxQuit()
+	defer cancel()
+
+	return h.policyStore.StorePurchasePolicy(ctx, sellAccept)
+}
```
ffranr: i don't think you need these methods since order handler instance can just access `h.policyStore` directly. I don't think `storeSalePolicy` and `storePurchasePolicy` are accessed outside of the `OrderHandler` since you have `RegisterAssetSalePolicy` etc

rfq/manager.go
```diff
@@ -179,34 +182,6 @@ type Manager struct {
...
-		SerialisedScid, rfqmsg.SellAccept,
-	]
-
-	// localAcceptedBuyQuotes holds buy quotes for assets that our node has
-	// accepted and that have been requested by peer nodes. These quotes are
-	// exclusively used by our node for the acquisition of assets, as they
-	// represent agreed-upon terms for purchase transactions with our peers.
-	localAcceptedBuyQuotes lnutils.SyncMap[SerialisedScid, rfqmsg.BuyAccept]
-
-	// localAcceptedSellQuotes holds sell quotes for assets that our node
```
ffranr, line 202: I think you should explain this refactor in the git commit message.

tapdb/rfq_policies.go
```diff
@@ -0,0 +1,533 @@
...
+// RfqPolicyType denotes the type of a persisted RFQ policy.
+type RfqPolicyType string
+
+const (
+	// RfqPolicyTypeAssetSale identifies an asset sale policy.
+	RfqPolicyTypeAssetSale RfqPolicyType = "asset_sale"
+
+	// RfqPolicyTypeAssetPurchase identifies an asset purchase policy.
+	RfqPolicyTypeAssetPurchase RfqPolicyType = "asset_purchase"
+)
```
ffranr, line 27: better const values from a db/debugging point of view might be:
```
const (
	// RfqPolicyTypeAssetSale identifies an asset sale policy.
	RfqPolicyTypeAssetSale RfqPolicyType = "RFQ_POLICY_TYPE_SALE"

	// RfqPolicyTypeAssetPurchase identifies an asset purchase policy.
	RfqPolicyTypeAssetPurchase RfqPolicyType = "RFQ_POLICY_TYPE_PURCHASE"
)
```

tapdb/rfq_policies.go
```diff
@@ -0,0 +1,533 @@
...
+// BatchedRfqPolicyStore supports batched database operations.
+type BatchedRfqPolicyStore interface {
+	RfqPolicyStore
+	BatchedTx[RfqPolicyStore]
+}
+
+// PersistedPolicyStore offers helpers to persist and load RFQ policies.
+type PersistedPolicyStore struct {
+	db BatchedRfqPolicyStore
+}
```
ffranr: i think you should add doc comments to these fields and interface methods.

tapdb/rfq_policies.go
```diff
@@ -0,0 +1,533 @@
...
+		AssetID:             assetID,
+		AssetGroupKey:       groupKey,
+		RateCoefficient:     rateBytes,
+		RateScale:           accept.AssetRate.Rate.Scale,
+		ExpiryUnix:          uint64(expiry.Unix()),
+		MaxOutAssetAmt:      ptrUint64(accept.Request.AssetMaxAmt),
+		RequestAssetMaxAmt:  ptrUint64(accept.Request.AssetMaxAmt),
+		PriceOracleMetadata: accept.Request.PriceOracleMetadata,
+		RequestVersion:      ptrUint32(uint32(accept.Request.Version)),
+		AgreedAt:            time.Now().UTC(),
```
ffranr: The storage layer doesn't know the agreed at time. It knows when its going to try to store. Maybe we should pass in the agreed at time? Same below for `StorePurchasePolicy`.

tapdb/rfq_policies.go
```diff
@@ -0,0 +1,533 @@
...
+		}
+
+		for _, row := range rows {
+			policy := policyFromRow(row)
+
+			// Skip expired entries.
+			expiry := time.Unix(int64(policy.ExpiryUnix), 0).UTC()
+			if now.After(expiry) {
+				continue
+			}
```
ffranr: This sort of filtering makes more sense in the `FetchActiveRfqPolicies` query, since most entries will be expired normally.

tapdb/rfq_policies.go
```diff
@@ -0,0 +1,533 @@
...
+		return nil
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return buyAccepts, sellAccepts, nil
+}
+
+func newInsertParams(policy rfqPolicy) sqlc.InsertRfqPolicyParams {
```
ffranr, line 242: The functions below this point need:
* doc comments
* some of them are too trivial IMO, or otherwise are very general and should be moved somewhere more general.

tapdb/sqlc/migrations/000049_rfq_policies.up.sql
```diff
@@ -0,0 +1,22 @@
...
+    expiry BIGINT NOT NULL,
+    max_out_asset_amt BIGINT,
+    payment_max_msat BIGINT,
+    request_asset_max_amt BIGINT,
+    request_payment_max_msat BIGINT,
+    price_oracle_metadata TEXT,
+    request_version INTEGER,
+    agreed_at BIGINT NOT NULL,
+    UNIQUE(rfq_id)
+);
```
ffranr, line 54: I think we should add doc comments to these fields also.

tapdb/sqlc/queries/rfq.sql
```diff
@@ -0,0 +1,45 @@
...
+    rate_scale,
+    expiry,
+    max_out_asset_amt,
+    payment_max_msat,
+    request_asset_max_amt,
+    request_payment_max_msat,
+    price_oracle_metadata,
+    request_version,
+    agreed_at
+FROM rfq_policies;
```
ffranr: Might make sense to add expiry time filtering here.

rfq/manager.go
darioAnongba: true but this is how almost all error messages are written in this file. Let's not have inconsistent logs even though I agree we should reduce redundancy in these messages.

tapdb/rfq_policies.go
darioAnongba: I tried making these generic and moved to sqlutils to use same pattern as the `sql*` and `extract*` functions we currently have.

tapdb/sqlc/queries/rfq.sql
darioAnongba: true, forgot about this when removing the `active` field. Good catch.

tapdb/rfq_policies.go
darioAnongba: true, forgot about this when removing the `active` field. Good catch.

tapdb/rfq_policies.go
darioAnongba: I added an `AgreedAt` field into the `BuyAccept` and `SellAccept` structs that are set in the `New` methods as `time.Now()`. I feel like these fields should have existed previous to this PR.

rfq/interface.go
ffranr: I think that a common pattern we have elsewhere in the codebase is to use a `interface.go` file. `rfq/policy_store.go` may not be expanded much further and we shouldn't add a new file for each exported interface.

rfqmsg/buy_accept.go
```diff
@@ -47,6 +50,7 @@ func NewBuyAcceptFromRequest(request BuyRequest,
 		Version:   latestBuyAcceptVersion,
 		ID:        request.ID,
 		AssetRate: assetRate,
+		AgreedAt:  time.Now(),
```
ffranr: in this commit, use UTC: `time.Now()` -> `time.Now().UTC()`

rfqmsg/sell_accept.go
```diff
@@ -34,6 +34,9 @@ type SellAccept struct {
 
 	// sig is a signature over the serialized contents of the message.
 	sig [64]byte
+
+	// AgreedAt is the time at which the quote was accepted.
+	AgreedAt time.Time
```
ffranr: To be precise, we don't know when our peer accepted since this time is not transferred over the wire. This is the time that we parsed the wire message. Or otherwise, when generating an accept message, this is the time we generated the message.

tapdb/sqlc/migrations/000049_rfq_policies.up.sql
```diff
@@ -0,0 +1,55 @@
...
+
+    -- price_oracle_metadata contains metadata about the price oracle.
+    price_oracle_metadata TEXT,
+
+    -- request_version is the version of the RFQ request.
+    request_version INTEGER,
+
+    -- agreed_at is the timestamp when the policy was agreed upon.
+    agreed_at BIGINT NOT NULL,
+    UNIQUE(rfq_id)
```
ffranr: `UNIQUE(rfq_id)` could either be in the field definition or, better if possible, we could have a `CREATE UNIQUE INDEX ` for it so that we can drop later if necessary without having to recreate the table.

tapdb/sqlc/schemas/generated_schema.sql
```diff
@@ -782,6 +782,62 @@ CREATE TABLE proof_types (
...
+
+    -- price_oracle_metadata contains metadata about the price oracle.
+    price_oracle_metadata TEXT,
+
+    -- request_version is the version of the RFQ request.
+    request_version INTEGER,
+
+    -- agreed_at is the timestamp when the policy was agreed upon.
+    agreed_at BIGINT NOT NULL,
+    UNIQUE(rfq_id)
```
ffranr: `UNIQUE ` -> see my comment on `rfq_policies` table

rfq/order.go
```diff
@@ -537,6 +545,11 @@ func (c *AssetPurchasePolicy) GenerateInterceptorResponse(
 	}, nil
 }
 
+// QuoteID returns the RFQ identifier that originated this policy.
+func (c *AssetPurchasePolicy) QuoteID() (rfqmsg.ID, bool) {
+	return c.AcceptedQuoteId, true
+}
```
ffranr: I don't think this is called anywhere so it can be removed here and in the `Policy` interface. I might be wrong. Why does this have a `bool` return value? We could use `fn.Option` perhaps? But I think this method can be removed.

rfq/manager.go
ffranr: > Let's not have inconsistent logs

Error messages across the codebase are inconsistent. I think you should write the best error messages you can for any new code.

When debugging user issues from logs, consistency within a file is not important IMO. High signal helps. But dropping "error" doesn't make a massive difference anyway.

ffranr (commented):
Just a few more trivial points to consider, otherwise this PR is pretty much there IMO.

rfqmsg/buy_accept.go
GeorgeTsagk: could add a doc that describes how the `AgreedAt` is set within `NewBuyAcceptFromRequest`.

i.e if any other call-site attempts to recreate things from the request format they'll silently set a new `AgreetAt` value.

We don't have such a call site and don't think we'll do ever, but a doc would still be nice

rfq/manager.go
```diff
@@ -842,10 +814,11 @@ func (m *Manager) PeerAcceptedBuyQuotes() BuyAcceptMap {
 	// Returning the map directly is not thread safe. We will therefore
 	// create a copy.
 	buyQuotesCopy := make(map[SerialisedScid]rfqmsg.BuyAccept)
-	m.peerAcceptedBuyQuotes.ForEach(
+	m.orderHandler.peerAcceptedBuyQuotes.ForEach(
 		func(scid SerialisedScid, accept rfqmsg.BuyAccept) error {
 			if time.Now().After(accept.AssetRate.Expiry) {
-				m.peerAcceptedBuyQuotes.Delete(scid)
+				//nolint:lll
+				m.orderHandler.peerAcceptedBuyQuotes.Delete(scid)
```
GeorgeTsagk: nit: can get around lll with an assignment
```golang
quotes := m.orderHandler.peerAcceptedBuyQuotes
quotes.Delete(scid)
```

rfq/order.go
```diff
@@ -957,28 +1019,94 @@ func (h *OrderHandler) Start() error {
...
+		h.policies.Store(policy.scid, policy)
+	}
+
+	for _, accept := range buyAccepts {
+		h.localAcceptedBuyQuotes.Store(accept.ShortChannelId(), accept)
+	}
+
+	for _, accept := range sellAccepts {
+		h.localAcceptedSellQuotes.Store(accept.ShortChannelId(), accept)
+	}
```
GeorgeTsagk, line 1114: for future reference:

feels like in this part we're duplicating information. The policies contain most (if not all) of the quote information, so we could in the future chop the memory usage of the order handle by half

rfqmsg/sell_accept.go
darioAnongba: I improved the docs.

rfq/manager.go
darioAnongba: can't copy `SyncMap` like that because of the lock. I renamed the private fields to remove the redundant `Accepted` so it's shorter, we only store accepted quotes.

rfqmsg/buy_accept.go
darioAnongba: Do you think passing the `agreedAt` as param is best? I updated the docs.

rfq/order.go
darioAnongba: true! Residue of the old implementation with `active` field. Removed.

rfq/order.go
darioAnongba: Agree, the best fix is to move the store implementation to `tapdb` as an in-memory store (or hybrid if we mix with the DB storage, which is best IMO) so we can define a model that represents the data we need to store regardless of the types used in the business layer (`rfq` module). Then, through interfaces we simply map the data to the correct entities.
This is better for separation of concerns and will avoid data duplication and code verbosity. Happy to implement that next if needed.

rfq/order.go
```diff
@@ -719,10 +734,41 @@ type OrderHandler struct {
...
+	// represent agreed-upon terms for sale transactions with our peers.
+	peerSellQuotes lnutils.SyncMap[
+		SerialisedScid, rfqmsg.SellAccept,
+	]
+
+	// localBuyQuotes holds buy quotes for assets that our node has
+	// accepted and that have been requested by peer nodes. These quotes are
+	// exclusively used by our node for the acquisition of assets, as they
+	// represent agreed-upon terms for purchase transactions with our peers.
+	localBuyQuotes lnutils.SyncMap[SerialisedScid, rfqmsg.BuyAccept]
```
jcbrand, line 754: Hi @darioAnongba, do I understand correctly that these quotes aren't necessarily fulfilled via payments, right?

If it were possible to know both which quotes an edge node has provided, as well as which of those quotes have been fulfilled, one could use that ratio as feedback to a pricing model, allowing the edge node to
come up with better quotes (i.e. with a higher likelihood of being accepted).

Perhaps related to this: https://github.com/lightninglabs/taproot-assets/issues/1362

darioAnongba: Yes these quotes you highlighted aren't necessarily fulfilled.

Maybe what you are interesting in is related to this? https://github.com/lightninglabs/taproot-assets/issues/1005 (i.e. _some kind of extra database table or other structure is needed so the Edge Node can "keep records" on their activity?_)

jcbrand: Thanks! Yes that looks relevant indeed.

rfq/order.go
```diff
@@ -325,6 +325,11 @@ func (c *AssetSalePolicy) GenerateInterceptorResponse(
 	}, nil
 }
 
+// QuoteID returns the RFQ identifier that originated this policy.
+func (c *AssetSalePolicy) QuoteID() (rfqmsg.ID, bool) {
+	return c.AcceptedQuoteId, true
+}
+
```
ffranr: This can be removed? https://github.com/lightninglabs/taproot-assets/pull/1863#discussion_r2586007919

rfq/order.go
```diff
@@ -674,6 +679,13 @@ func (a *AssetForwardPolicy) GenerateInterceptorResponse(
 	}, nil
 }
 
+// QuoteID returns the RFQ identifier that originated this policy.
+//
+// Forward policies do not map to a single quote, hence we return false.
+func (a *AssetForwardPolicy) QuoteID() (rfqmsg.ID, bool) {
+	return rfqmsg.ID{}, false
+}
```
ffranr: This can be removed? https://github.com/lightninglabs/taproot-assets/pull/1863#discussion_r2586007919

rfq/order.go
darioAnongba: Oops forgot to remove the func 🤦 . Thanks!

rfq/order.go
darioAnongba: Oops forgot to remove the func 🤦 . Thanks!

Issue #1865, opened 2025-11-04
[feature]: Add byte-based limit to universe proof cache


ffranr:
The universe proof cache is currently limited by number of proofs per universe. Proof size varies and will grow over time so we need a total-byte-size cap.

Work:

* Add `Size() uint64` to `mssmt.Node`, `Leaf`, and implementations.
* Add `Size() uint64` to `proof.Proof`.
* Use these to track proof byte size in cache.
* Evict based on byte budget.
* Add unit tests for the size methods and eviction logic.

Output:

* Config option for max cache bytes.
* Byte-aware eviction in universe proof cache.

ffranr:
We cannot track byte usage correctly unless the LRU reports whether `Put` replaced an existing entry or inserted a new one. That distinction matters because a replace may reduce or increase total bytes, while a new insert always increases it. The current API only exposes whether eviction occurred, so we need an additional signal from `Put` to detect replacement vs insertion. See current `Put` method:
```
// Put inserts a given (key,value) pair into the cache, if the key already
// exists, it will replace value and update it to be most recent item in cache.
// The return value indicates whether items had to be evicted to make room for
// the new element.
func (c *Cache[K, V]) Put(key K, value V) (bool, error) {
```

It probably makes sense just to add a byte size aware cache at this point.

jtobin:
Closing as resolved by #1870.

Issue #503, opened 2023-09-13
Optimise universe proof leaf additional inputs


ffranr:
As things currently stand, a universe's proof leaf has a field called `AdditionalInputs`. This field contains proofs for other inputs that may have been merged as part of a spend.

The `AdditionalInputs` field is of type `[]proof.File`. Which means, that for each additional input, we store an entire proof file. A universe leaf is supposed to represent a single proof. But where additional inputs are concerned, a universe leaf may also contain entire proof files for each additional input. 

This means that a universe makes sub-optimal use of memory in that each leaf may also contain entire proof files which are duplicated as leaves in other parts of the universe.

Within a universe leaf, the `AdditionalInputs` field should perhaps contain proof locators instead of entire proof files.

PR #1870, opened 2025-11-07
tapdb: switch proof cache to size-based limits and add deep size estimation

ffranr:
Issue: https://github.com/lightninglabs/taproot-assets/issues/1865

---

This PR introduces a size-based proof cache and supporting deep-size logic.

**Summary**

* Replace proof-count cache limits with a byte-size limit.
* Add human-readable size parsing in configuration.
* Introduce deep size estimation (`LowerBoundByteSize`) and apply it to proof cache entries.
* Update proof cache keys for finer-grained eviction.
* Improve error handling and logging around cache initialization.

#### Note

It's pretty trivial to get from here to logging cache size, but we need https://github.com/lightninglabs/neutrino/pull/330 for that.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -8,6 +8,7 @@
     - [Functional Updates](#functional-updates)
     - [RPC Updates](#rpc-updates)
     - [tapcli Updates](#tapcli-updates)
+    - [Config Changes](#config-changes)
     - [Breaking Changes](#breaking-changes)
     - [Performance Improvements](#performance-improvements)
     - [Deprecations](#deprecations)
@@ -46,6 +47,13 @@
   /authmailboxrpc.Mailbox/MailboxInfo.
 
 ## tapcli Updates
+
+## Config Changes
+
+- [PR#1870](https://github.com/lightninglabs/taproot-assets/pull/1870)
+  The `proofs-per-universe` configuration option is removed. New option 
+  `max-proof-cache-size` sets the proof cache limit in bytes and accepts
+  human-readable values such as `64MB`.
 
 ## Code Health
 
--- a/docs/release-notes/release-notes-template.md
+++ b/docs/release-notes/release-notes-template.md
@@ -8,6 +8,7 @@
     - [Functional Updates](#functional-updates)
     - [RPC Updates](#rpc-updates)
     - [tapcli Updates](#tapcli-updates)
+    - [Config Changes](#config-changes)
     - [Breaking Changes](#breaking-changes)
     - [Performance Improvements](#performance-improvements)
     - [Deprecations](#deprecations)
@@ -35,6 +36,8 @@
 ## RPC Updates
 
 ## tapcli Updates
+
+## Config Changes
 
 ## Code Health
 


litbot-9000:
@georgetsagk: review reminder

universe/interface.go
```diff
@@ -393,6 +393,11 @@ type Proof struct {
 	MultiverseInclusionProof *mssmt.Proof
 }
 
+// LowerBoundByteSize returns the lower bound on the byte size of the proof.
+func (p *Proof) LowerBoundByteSize() uint64 {
```
GeorgeTsagk, line 397: could also add some test coverage  for the `LowerBoundByteSize` function of `universe.Proof`

tapdb/multiverse_cache.go
```diff
@@ -67,80 +65,86 @@ func DefaultMultiverseCacheConfig() MultiverseCacheConfig {
...
+		return 0, nil
+	}
+
+	totalBytes := uint64(0)
+	for _, proof := range *c {
+		if proof == nil {
+			continue
+		}
+
+		totalBytes += proof.LowerBoundByteSize()
```
GeorgeTsagk: Would be nice having a caching layer for this cache, in order to avoid size calculating traversals over entries that have had their size calculated already
`entrySizeCache map[UniverseProofKey] uint64`

tapdb/multiverse.go
```diff
@@ -162,7 +162,7 @@ func NewMultiverseStore(db BatchedMultiverse,
 			cfg.Caches.LeavesPerUniverse,
 		),
 		transferProofDistributor: fn.NewEventDistributor[proof.Blob](),
-	}
+	}, nil
```
GeorgeTsagk, line 168: can drop this commit -- seems like no error is ever returned?

tapdb/multiverse_cache.go
```diff
@@ -67,80 +65,86 @@ func DefaultMultiverseCacheConfig() MultiverseCacheConfig {
...
+
+	return totalBytes, nil
 }
 
 // newProofCache creates a new leaf proof cache.
-func newProofCache(proofCacheSize uint64) *lru.Cache[ProofKey, *cachedProofs] {
-	return lru.NewCache[ProofKey, *cachedProofs](proofCacheSize)
+//
+// nolint: lll
+func newProofCache(totalCacheBytesSize uint64) *lru.Cache[UniverseProofKey, *cachedProofs] {
```
GeorgeTsagk, line 139: I wonder if a direction where proof entries lack the recursive element (i.e size within bounds or fixed) is a better approach to this overall.  Even with a fully fledged size aware cache we'll still have overlapping content between various proof files.

The PR seems in good shape so def not worth changing our mind now. Should make sure there's an issue tracking the elimination of the recursive element of the entries.

universe/interface.go
ffranr: `universe.Proof.LowerBoundByteSize` just delegates to `fn.LowerBoundByteSize`, and the underlying function already has test coverage.

Maybe I've misunderstood what you mean.

tapdb/multiverse.go
ffranr: The next commit adds an error return code path to `NewMultiverseStore`.

tapdb/multiverse_cache.go
ffranr: I think we need both this size aware cache and also proof reference optimisation. This issue tracks most of the proof optimisation work: https://github.com/lightninglabs/taproot-assets/issues/503

universe/interface.go
GeorgeTsagk: I mean we should create dummy `universe.Proof` instances and verify that an iteration of `fn.LowerBoundByteSize` over it actually returns the size of it.

I guess if we're very confident on the `fn` function coverage this is actually not needed. Would only be a sanity check that it can iterate a proof properly.

ffranr: The difficulty we have here is that `fn.LowerBoundByteSize` is the only method we have for estimating a lower bound on the `universe.Proof` instance byte size.

I've added unit tests which creates random `universe.Proof` instances, mutates them, and then investigates the byte size change using `fn.LowerBoundByteSize`.

tapdb/multiverse_cache.go
```diff
@@ -153,28 +216,54 @@ func (p *universeProofCache) fetchProof(id universe.Identifier,
...
+func (p *universeProofCache) RemoveLeafKeyProofs(id universe.Identifier,
+	leafKey universe.LeafKey) {
+
+	log.Debugf("Removing leaf key proofs (universe_id=%s, leaf_key=%v)",
+		id.StringForLog(), leafKey)
 
-	p.Delete(id.String())
+	targetCacheKey := NewUniverseProofKey(id, leafKey)
+
+	p.cache.Range(
```
jtobin: Does this need to iterate? Looks like we could just delete this key outright, yeah? Contra RemoveUniverseProofs, which needs to do a full traversal.

ffranr: Not sure how I missed that! Thanks for the review.

PR #1875, opened 2025-11-18
Add `ChainPorter` pre-broadcast validation for pre-anchored sends

ffranr:
- Introduce `SendStateVerifyPreBroadcast` so `ChainPorter` re-validates anchor inputs/outputs immediately before broadcast, covering the pre-anchored RPC path that previously skipped this check
- Refactor `validateInputAssets` to expose the purged/tombstone assets it already collects, ensuring the RPC layer can pass the complete tap tree data into `ChainPorter`’s new validation state
- Wire the pruned asset map through `PublishAndLogTransfer` → pre-anchored parcel → `sendPackage`, allowing the new pre-broadcast validation to succeed even when the original input commitments aren’t locally available

### DIFF
--- a/itest/psbt_test.go
+++ b/itest/psbt_test.go
@@ -3663,7 +3663,7 @@ func testPsbtRelativeLockTimeSendProofFail(t *harnessTest) {
 
 	AssertSendEvents(
 		t.t, aliceScriptKeyBytes, sendEvents,
-		tapfreighter.SendStateStorePreBroadcast,
+		tapfreighter.SendStateVerifyPreBroadcast,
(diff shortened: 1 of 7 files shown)


ffranr:
> LGTM, modulo resolving the LiT test failure -- I've kicked that job again to test if it was just a flake.
> 
> (Oh, yes -- tmp_change_dsc.md should be removed pre-merge, but is a very nice touch!)

Thanks for taking a look. These changes won't work as they are. The LiT test won't pass.

I've got things working locally, will push up an update soon.

ffranr:
Replaced with https://github.com/lightninglabs/taproot-assets/pull/1884

PR #1878, opened 2025-11-19
CI: add reusable cleanup workflow and clear Go caches

ffranr:
- Add a reusable GH-action that handles disk cleanup for CI runners.
- Wire the cleanup action into existing workflows instead of inline steps.
- Clear Go caches between multi-platform release builds so artifacts don’t mix.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -175,7 +175,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
+
+      - name: Clean up runner space
+        uses: ./.github/actions/cleanup-space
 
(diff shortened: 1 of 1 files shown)


.github/actions/cleanup-space/action.yaml
```diff
@@ -0,0 +1,50 @@
...
+            '^llvm-.*' \
+            'php.*' \
+            '^mongodb-.*' \
+            '^mysql-.*' \
+            azure-cli \
+            google-chrome-stable \
+            firefox \
+            powershell \
+            mono-devel \
+            libgl1-mesa-dri 2>/dev/null || true
```
GeorgeTsagk, line 41: I assume none of these were ever used to build the binaries, but even if they did we'll see it fail

ffranr: correct

PR #1881, opened 2025-11-21
tapd+sqlc: separate SQL migrations from code migrations

ViktorT-11:
Previously, SQL migrations and code migrations have been designed to be a single migration version, where the SQL migration was run first, followed by a post step migration callback that executed the code migration.

Due to a re-definition of the migration design, which separates SQL and code migrations, we need to update tapd to become compatible with the new migration design before updating to the new design.

This PR therefore separates the code migrations from migration `33` & `37`, into separate migration files, and adds those two new migrations to the `tapdb/sqlc/migrations` directory.

Additionally, since users that have already run the code migrations during migration `33` & `37` will have those migrations re-executed once more after updating to this PR, we test that re-execution of those code migrations will be a no-op for those users, and won't add or change any data in the database.

### DIFF
--- a/scripts/check-migration-numbering.sh
+++ b/scripts/check-migration-numbering.sh
@@ -9,6 +9,25 @@ if [ ! -d "$migrations_path" ]; then
   exit 1
 fi
 
+# is_sql_migration checks if a .up.sql file contains executable SQL
+# statements and not just comments/whitespace/semicolons. The logic mirrors the
+# Go helper the `migrate` package dependency uses in the `migrate.go` file.
+is_sql_migration() {
+  local file="$1"
+  local cleaned
+
+  cleaned=$(
+    perl -0777 -pe '
+      s/^\x{FEFF}//;
+      s{/\*.*?\*/}{}gs;
+      s{--[^\r\n]*}{}g;
+    ' "$file" 2>/dev/null | \
+    sed -E 's/[[:space:];]+//g'
+  )
+
+  [ -n "$cleaned" ]
+}
+
 # Get all unique prefixes (e.g., 000001, always 6 digits) from .up.sql files.
 prefixes=($(ls "$migrations_path"/*.up.sql 2>/dev/null | \
     sed -E 's/.*\/([0-9]{6})_.*\.up\.sql/\1/' | sort))
@@ -34,7 +53,11 @@ for i in "${!prefixes[@]}"; do
   base_filename=$(ls "$migrations_path/${prefixes[$i]}"_*.up.sql | \
       sed -E 's/\.up\.sql//')
 
-  if [ ! -f "$base_filename.down.sql" ]; then
+  # Error if the .up.sql is an SQL migration, but we're missing the
+  # corresponding .down.sql. This doesn't apply if the .up.sql file is a
+  # programmatic migration.
+  if is_sql_migration "$base_filename.up.sql" && \
+      [ ! -f "$base_filename.down.sql" ]; then
     echo "Error: Missing .down.sql file for migration $expected_prefix."
     exit 1
   fi


ViktorT-11:
Will address any failing CI checks on Monday.

jtobin:
The failing tapgarden unit test passes for me locally on this branch. I'll kick the CI job again and see if it proves flaky here.

(EDIT: yeah, it was a flake.)

tapdb/migrations_test.go
```diff
@@ -672,6 +677,116 @@ func TestMigration33(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+
+	// Executing the code migration again (now at migration 48) should not
+	// change or add any new values than the values assigned when the code
+	// migration was run for migration version 33.
+	err = db.ExecuteMigrations(TargetLatest, WithPostStepCallbacks(
+		makePostStepCallbacks(db, postMigrationChecks),
+	))
+	require.NoError(t, err)
```
ffranr: Target 48 explicitly here? `TargetVersion(48)` Or maybe that's not necessary and I'm missing something.

ffranr: Same below for `TargetVersion(49)`.

ffranr: might be nice to have const values for migration numbers (33, 49 and 48, etc) in these unit tests. Start/replacement. Maybe?

scripts/check-migration-numbering.sh
```diff
@@ -34,7 +53,11 @@ for i in "${!prefixes[@]}"; do
   base_filename=$(ls "$migrations_path/${prefixes[$i]}"_*.up.sql | \
       sed -E 's/\.up\.sql//')
 
-  if [ ! -f "$base_filename.down.sql" ]; then
+  # Error if the .up.sql is an SQL migration, but we're missing the
+  # corresponding .down.sql. This doesn't apply if the .up.sql file is a code
+  # migration.
+  if is_sql_migration "$base_filename.up.sql" && \
+      [ ! -f "$base_filename.down.sql" ]; then
```
ffranr, line 60: Updated:

Do we need to include this commit? Is it required for your end goal?

We don’t often have no-op down migrations, but IMO it’s fine to include them. It keeps the migration sequence contiguous, and IMO it’s cleaner to always provide a down file for every up, even if the down is empty.

jtobin (commented):
LGTM. :+1: :+1:  The only change I would suggest is to more explicitly reference the migration design change in your commit message (e.g. by issue/PR, etc.) so that the source is better-captured in the history.

scripts/check-migration-numbering.sh
ViktorT-11: > Do we need to include this commit? Is it required for your end goal?

It’s not required, but I want to explain why I think this change is beneficial.

This change specifically targets **code migrations**, allowing them to omit a `down.sql` file when appropriate.

To illustrate why SQL and code migrations should behave differently, consider the following example of migrations:

* **`0001_example.up.sql`**
  `CREATE TABLE example (id INTEGER PRIMARY KEY);`

* **`0001_example.down.sql`**
  `DROP TABLE example;`

* Followed by a **code migration**:
  **`0002_example_code_migration.up.sql`**
  (This migration inserts data into the `example` table.)

If we also add a `down.sql` for the code migration:

* **`0002_example_code_migration.down.sql`**
  (This would simply re-run the code migration and insert the data again.)

This highlights the mismatch:

* For **SQL migrations**, the `down.sql` undoes the `up.sql`.
* For **code migrations**, a `down.sql` would unintentionally re-execute the migration instead of undoing the changes.
  In this example, the data inserted by the code migration would actually be removed only when the table itself is dropped by `0001_example.down.sql`.

Currently, the code migrations in `tapd` are structured as no-ops when re-executed, but that assumption can't be true for all code migrations. For example, it'd be much more complex for migrations like the kvdb → SQL migration in `litd`.

---

For that reason, I think the code migration itself should indicate whether a `down.sql` should be present or not, rather than having CI enforce for all types of migrations. The `migrate` library explicitly supports migrations without a `down.sql` file, and I think this is a case where that functionality makes sense to use.

ViktorT-11: > We don’t often have no-op down migrations, but IMO it’s fine to include them. It keeps the migration sequence contiguous, and IMO it’s cleaner to always provide a down file for every up, even if the down is empty.

And just to clarify, this `if` clause here will only trigger when the `up.sql` case contains any executable content, i.e. when the `up.sql` is an SQL migration. It's not checking if the `down.sql` file is empty.

ViktorT-11:
Thanks for the reviews @ffranr & @jtobin :tada:! It'll address your feedback and rebase on `main` once https://github.com/lightninglabs/migrate/pull/3 is final :rocket:

litbot-9000:
@viktort-11, remember to re-request review from reviewers when ready

ViktorT-11:
Rebased on `main` and addressed the feedback of the previous reviews :). 

I did not address https://github.com/lightninglabs/taproot-assets/pull/1881#discussion_r2558230513, as that is still under discussion.

Issue #1882, opened 2025-11-24
[flake]: TestBatchedAssetIssuance/minting_with_cancellation flake


jtobin:
This test was observed failing flakily in a CI run for #1881 under the "unit tests (unit-cover)" job:

```
--- FAIL: TestBatchedAssetIssuance (87.49s)
    --- FAIL: TestBatchedAssetIssuance/minting_with_cancellation (62.30s)
        test_sqlite.go:11: Creating new SQLite DB handle for testing: /tmp/TestBatchedAssetIssuanceminting_with_cancellation1519908230/001/tmp.db
        planter_test.go:382: Missing expected seedling: (*tapgarden.Seedling)(0xc0008c6840)({
              AssetVersion: (asset.Version) 0,
              AssetType: (asset.Type) 1,
              AssetName: (string) (len=64) "65fff78845e603d0788ee37a622e6f28248c433fbd70691dafb6dcb8754c7b76",
              Meta: (*proof.MetaReveal)(0xc0008e1ce0)({
                Type: (proof.MetaType) 0,
                Data: ([]uint8) (len=32 cap=32) {
                  00000000  65 ff f7 88 45 e6 03 d0  78 8e e3 7a 62 2e 6f 28  |e...E...x..zb.o(|
                  00000010  24 8c 43 3f bd 70 69 1d  af b6 dc b8 75 4c 7b 76  |$.C?.pi.....uL{v|
                },
                DecimalDisplay: (fn.Option[uint32]) {
                  isSome: (bool) false,
                  some: (uint32) 0
                },
                UniverseCommitments: (bool) false,
                CanonicalUniverses: (fn.Option[[]net/url.URL]) {
                  isSome: (bool) false,
                  some: ([]url.URL) <nil>
                },
                DelegationKey: (fn.Option[github.com/decred/dcrd/dcrec/secp256k1/v4.PublicKey]) {
                  isSome: (bool) false,
                  some: (secp256k1.PublicKey) {
                    x: (secp256k1.FieldVal) {
                      n: ([10]uint32) (len=10 cap=10) {
                        (uint32) 0,
                        (uint32) 0,
                        (uint32) 0,
                        (uint32) 0,
                        (uint32) 0,
                        (uint32) 0,
                        (uint32) 0,
                        (uint32) 0,
                        (uint32) 0,
                        (uint32) 0
                      }
                    },
                    y: (secp256k1.FieldVal) {
[...]

PR #1883, opened 2025-11-24
tapdb: introduce sqlTime helper

darioAnongba:
Replace boilerplate code using sql.NullTime with helper sqlTime, reproducing pattern used with `sqlInt16`, `sqlBool`, etc.


jtobin:
The failing unit-cover test appears to be the same flake I reported in #1882.

PR #1884, opened 2025-11-25
tapfreighter: add `SendStateVerifyPreBroadcast` and supporting proof reanchor helpers

ffranr:
This PR adds a new pre-broadcast validation step to the `ChainPorter` so that input proofs on pre-anchored send packages are validated before the anchor transaction is broadcast. Broadcasting is a non-reversible action, so we now fail early when a package carries invalid input proofs.

* Add new ChainPorter state `SendStateVerifyPreBroadcast` to run input-proof validation before broadcasting the anchor transaction.
* Pre-anchored send packages now start in this state instead of `SendStateStorePreBroadcast`.
* Refactor and extend tapchannel proof-handling so that input proofs are fully finalized before validation. This includes:
  * Shared reanchoring logic (`reanchorAssetOutputs`) to anchor proofs to the actual commitment transaction.
  * New helpers (`updateProofsFromShortChanID`, `Proof.TaprootOutputScript`) to populate and derive the data required for correct input-proof construction.
  * Block/tx lookup via the new `ChainBridge.GetBlockByHeight` path to avoid per-caller duplication.
* These tapchannel changes are required for the input proofs to pass the new validation state.

---

#### Notes

* Depends on https://github.com/lightningnetwork/lnd/pull/10391

### DIFF
--- a/tapchannel/aux_sweeper.go
+++ b/tapchannel/aux_sweeper.go
@@ -1090,6 +1090,84 @@ func assetOutputToVPacket(fundingInputProofs map[asset.ID]*proof.Proof,
 	return nil
 }
 
+// reanchorAssetOutputs updates each asset output proof so that it references
+// the actual commitment transaction output. Proofs are initially built using
+// a synthetic commitment template. Once the real commitment transaction is
+// known, we rewrite the proof's anchor transaction and output indexes.
+func reanchorAssetOutputs(ctx context.Context,
+	chainBridge tapgarden.ChainBridge, commitTx wire.MsgTx,
+	commitTxBlockHeight uint32, outputs []*cmsg.AssetOutput) error {
+
+	if len(outputs) == 0 {
+		return nil
+	}
+
+	// Only fetch block data if any proof is missing it.
+	proofBlockParams, err := proofParamsForCommitTx(
+		ctx, chainBridge, commitTxBlockHeight, commitTx,
+	)
+	if err != nil {
+		return fmt.Errorf("constructing proof block params: %w", err)
+	}
+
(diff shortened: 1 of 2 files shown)


ffranr:
LiT CI job is failing because litd will also need LND bump to https://github.com/lightningnetwork/lnd/pull/10391

tapfreighter/chain_porter.go
```diff
@@ -1398,6 +1398,45 @@ func (p *ChainPorter) prelimCheckAddrParcel(addrParcel AddressParcel) error {
...
+	for pktIdx := range packets {
+		vPkt := packets[pktIdx]
+		for inputIdx := range vPkt.Inputs {
+			assetProof := vPkt.Inputs[inputIdx].Proof
+			if assetProof == nil {
+				return fmt.Errorf("packet %d input %d proof "+
+					"is nil", pktIdx, inputIdx)
+			}
+
+			_, err := assetProof.VerifyProofIntegrity(ctx, vCtx)
```
GeorgeTsagk, line 1428: so eventually we'd also want to run the `txscript.Engine` here?

tapchannel/aux_closer.go
```diff
@@ -259,6 +259,8 @@ func (a *AuxChanCloser) AuxCloseOutputs(
 			"state: %w", err)
 	}
 
+	ctxb := context.Background()
```
GeorgeTsagk, line 262: we should avoid an indefinite ctx here: this method is called by lnd

GeorgeTsagk: realized that was the same before, so non-blocking atm

PR #1886, opened 2025-11-25
tapcfg: validate SQLite temp dir at startup and add opt out flag

ffranr:
This PR addresses https://github.com/lightninglabs/taproot-assets/issues/1869, where a user hit `SQLITE_IOERR_GETTEMPPATH (6410)` during `tapd` operation. The goal is to fail fast at startup if SQLite cannot access a writable temporary directory, instead of surfacing a late, opaque I/O error.

### Summary of changes

* **Validate SQLite temp directory at startup**

  * On `tapd` initialization, determine candidate temp directories using:
    * `SQLITE_TMPDIR` and `TMPDIR`
    * Platform default temp locations as a fallback
  * Perform a writability check on the selected directory so that we can abort early with a clear configuration error if it is not writable.
  * The intent is to catch container or host setups where `/tmp` or similar paths are not writable, which would later cause `SQLITE_IOERR_GETTEMPPATH`.

* **Add config flag to skip the check**

  * Introduce a new tapcfg option to skip the SQLite temp directory writability check at startup.
  * This gives operators an escape hatch in environments where:

    * The temp directory is managed in a non standard way, or
    * They explicitly prefer to rely on SQLite’s own behavior instead of the upfront check.

### Rationale

* `SQLITE_IOERR_GETTEMPPATH (6410)` typically indicates that SQLite’s VFS cannot determine or use a suitable directory for temporary files.
* In production and containerized environments this often points to a misconfigured or read only temp volume.
* Detecting this at startup with an explicit error and remediation hint is preferable to failing later during normal operation.

### DIFF
--- a/tapcfg/server.go
+++ b/tapcfg/server.go
@@ -50,7 +50,9 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 	// If we're using sqlite, we need to ensure that the temp directory is
 	// writable otherwise we might encounter an error at an unexpected
 	// time.
-	if cfg.DatabaseBackend == DatabaseBackendSqlite {
+	if !cfg.Sqlite.SkipTmpDirCheck &&
+		cfg.DatabaseBackend == DatabaseBackendSqlite {
+
 		err = checkSQLiteTempDir()
 		if err != nil {
 			return nil, fmt.Errorf("unable to ensure sqlite tmp "+
--- a/sample-tapd.conf
+++ b/sample-tapd.conf
@@ -296,6 +296,9 @@
 ; The full path to the database
 ; sqlite.dbfile=~/.tapd/data/testnet/tapd.db
 
+; Skip the temporary directory check on startup.
+; sqlite.skiptmpdircheck=false
+
 [postgres]
 
 ; Skip applying migrations on startup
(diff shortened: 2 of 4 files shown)


tapcfg/server.go
```diff
@@ -801,6 +812,10 @@ func CreateServerFromConfig(cfg *Config, cfgLogger btclog.Logger,
 	shutdownInterceptor signal.Interceptor, enableChannelFeatures bool,
 	mainErrChan chan<- error) (*tap.Server, error) {
 
+	if cfg.DatabaseBackend == DatabaseBackendSqlite {
```
jtobin: Lingering skeleton code?

tapcfg/server.go
```diff
@@ -47,6 +47,19 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 		dbType sqlc.BackendType
 	)
 
+	// If we're using sqlite, we need to ensure that the temp directory is
+	// writable otherwise we might encounter and error at an unexpected
```
jtobin: Nit: s/and error/an error.

jtobin (approved):
Nice one. There's that one conditional block artifact that should be removed, but otherwise LGTM. :+1:

tapcfg/validate.go
```diff
@@ -0,0 +1,80 @@
...
+	"strings"
+)
+
+// ensureDirWritable verifies that the provided directory exists, is a directory
+// and is writable by creating a temporary file within it.
+func ensureDirWritable(dir string) error {
+	dirInfo, err := os.Stat(dir)
+	if err != nil {
+		return fmt.Errorf("not accessible (dir=%s): %w", dir, err)
+	}
```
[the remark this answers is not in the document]
ffranr: unnecessary

GeorgeTsagk (approved):
LGTM

Issue #1877, opened 2025-11-19
[feature]: unify release build methods


jtobin:
At present we have a couple of canonical ways to build releases, namely `make release` and `make docker-release`, and these can produce different artifacts depending on the environment used for the build. We also have some instructions to use `make docker-release` in docs/release.md, but use `make release` in CI, leading to reproducibility problems like that discussed in #1874.

We should have one canonical method by which to create a totally reproducible and verifiable release build, anywhere, and use that consistently. The release build docs should then be updated as well.

ffranr:
As far as I recall, our earlier thinking on this was to aim for reproducible builds across different environments. We do not want the reproducibility check and its security properties to depend on a single Docker image, for example. The goal is not just reproduction per se, but reproduction across diverse environments, demonstrating that the source-to-binary mapping is consistent.

Roasbeef:
`make docker-release` is for building on Mac. By default we'll get a different end archive but matching binaries. 

Based on the above, I think this can be closed.

jtobin:
> As far as I recall, our earlier thinking on this was to aim for reproducible builds across different environments. We do not want the reproducibility check and its security properties to depend on a single Docker image, for example. The goal is not just reproduction per se, but reproduction across diverse environments, demonstrating that the source-to-binary mapping is consistent.

Summarizing my understanding of an out-of-band discussion we had around this, for posterity.

My primary focus here is that we have a build that can produce totally reproducible artifacts. I.e., we can establish a hermetic build environment containing only explicit inputs (host architecture/platform, toolchains, utilities, source, etc.), build our stuff in it, and thus get fully deterministic outputs. 

(N.b., it strikes me as likely that `make docker-release` may mostly fit the bill for this already, but the use of apt-get update and apt-get install for build utilities in the corresponding Dockerfile gives me pause. Seems we would want to bake an image that used consistent versions of these utilities, updating *that image* as needed.)

Whatever that tool is, anyway, we should exclusively use that to produce release builds, particularly in CI, and direct others to use it as well. Per @ffranr's point, this build should not *depend strongly* on some particular inputs (like a single docker image), but rather the build must allow *precise specification* of the inputs (which may be some particular docker image, but possibly some other docker image).

jtobin:
After some further work here:

[...]

PR #1891, opened 2025-12-02
Assorted release maintenance

### LINKED ISSUE #1874: [bug]: release verification differs between environments
### Background

Release verification does not currently work on macOS devices because of different reasons:
1. Usage of `grep -P` in `get-git-tag-name.sh`
2. `gzip` issue: See https://github.com/lightningnetwork/lnd/pull/10268

### Steps to reproduce

Execution of `make release` differs between macOS and Linux environments.

### Expected behavior

manifest result should be the same

### Actual behavior

Tell us what happens instead

jtobin:
Closes #1874, #1877.

See the individual commits for more details, but the TLDR version:

* Adds installation checks (à la the existing GNU tar check) to scripts/release.sh for other required utilities.
* Adjusts the CI release workflow to build with `docker-release` (see commentary in #1877 for rationale).
* Updates and modernizes the release docs (docs/release.md).
* Makes scripts/verify-install.sh work on our official Docker images (i.e., going forward).

### DIFF
--- a/make/builder.Dockerfile
+++ b/make/builder.Dockerfile
@@ -1,10 +1,10 @@
 FROM golang:1.24.9-bookworm@sha256:737b40b61ce956d738bed59f18ba854d8d67e7a4c4fa63f64437f4c70247ac5b
 
-MAINTAINER Olaoluwa Osuntokun <laolu@lightning.engineering>
+LABEL maintainer="Olaoluwa Osuntokun <laolu@lightning.engineering>"
 
 # Golang build related environment variables that are static and used for all
 # architectures/OSes.
-ENV GODEBUG netdns=cgo
+ENV GODEBUG=netdns=cgo
 ENV CGO_ENABLED=0
 
 # Set up cache directories. Those will be mounted from the host system to speed


scripts/release.sh
```diff
@@ -125,6 +132,19 @@ function build_release() {
...
+  # coreutils's 'sha256sum'.
+  if ! command -v "shasum" &>/dev/null; then
+    if ! command -v "sha256sum" &>/dev/null; then
+      echo "Either a Perl or GNU coreutils installation is required."
+      exit 1
+    else
+      # coreutils is installed, so use sha256sum instead.
+      shasum_cmd="sha256sum"
+    fi
+  fi
```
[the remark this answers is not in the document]
jtobin: Huh, shellcheck didn't catch this one.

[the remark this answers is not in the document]
jtobin: Fixed in 9d998206b1912e94fec13ab69249c24eb4e28de6.

scripts/release.sh
```diff
@@ -189,7 +209,7 @@ required Go version ($goversion)."
 
     # Add the hashes for the individual binaries as well for easy verification
     # of a single installed binary.
-    shasum -a 256 "${dir}/"* >> "manifest-$tag.txt" 
+    "$shasum_cmd" "${dir}/"* >> "manifest-$tag.txt"
```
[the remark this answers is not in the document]
jtobin: Fixed in 9d998206b1912e94fec13ab69249c24eb4e28de6 via unquoting.

scripts/release.sh
```diff
@@ -199,7 +219,7 @@ required Go version ($goversion)."
   done
 
   # Add the hash of the packages too, then sort by the second column (name).
-  shasum -a 256 $PACKAGE-* vendor* >> "manifest-$tag.txt"
+  "$shasum_cmd" $PACKAGE-* vendor* >> "manifest-$tag.txt"
```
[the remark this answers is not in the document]
jtobin: Fixed in 9d998206b1912e94fec13ab69249c24eb4e28de6 via unquoting.

scripts/release.sh
```diff
@@ -125,6 +132,19 @@ function build_release() {
   local buildtags=$3
   local ldflags=$4
   local goversion=$5
+  local shasum_cmd="shasum -a 256"
+
+  # If we don't have a Perl installation available, then check for
```
Roasbeef, line 137: Where do we rely on perl rn?

.github/workflows/release.yaml
```diff
@@ -37,16 +37,16 @@ jobs:
...
         run: git checkout ${{ env.RELEASE_VERSION }}
 
+      - name: build release helper docker image
+        run: |
+          docker build -t taproot-assets-release-helper \
+            -f make/builder.Dockerfile make/
+
       - name: build release for all architectures
-        run: make release tag=${{ env.RELEASE_VERSION }}
+        run: make docker-release tag=${{ env.RELEASE_VERSION }}
```
Roasbeef, line 49: Idea is to move to prefer the docker release? 

We should test to see how this impacts the fixes that we had to roll out to ensure that there's enough disk space left for the build. 

Will the subsequent steps also be able to find the release artefacts?

docs/release.md
```diff
@@ -1,125 +1,112 @@
...
 
 ```shell
-$  git clone https://github.com/lightninglabs/taproot-assets.git
-$  cd taproot-assets
-$  git checkout <TAG> # <TAG> is the name of the next release/tag
-$  make docker-release tag=<TAG>
+$ git clone https://github.com/lightninglabs/taproot-assets.git
+$ cd taproot-assets
+$ git checkout "$TAG"
+$ make docker-release tag="$TAG"
```
Roasbeef, line 28: I think we should be able to speed up the docker-release (at least on mac) using a similar technique as my recent PR to speed up linting.

docs/release.md
```diff
@@ -1,125 +1,112 @@
...
 ```
 
-Where `<TAG>` is the name of the next release of `taproot-assets`.
+Where `$TAG` is the name of the desired release of `taproot-assets`.
 
-### Linux/Windows (WSL)
+Note that you can also run `make release` to avoid the dependency on
```
Roasbeef, line 33: 👍

docs/release.md
```diff
@@ -1,125 +1,112 @@
...
-example):
+   PGP public keys of Taproot Assets developers can be found in the
+   scripts/keys subdirectory of the repository root.
+
+3. Procure the other release artifacts, either by building the release from
+   source as [described above](#building-a-release), or by downloading
+   the desired archive(s) containing the release binaries from the
+   Taproot Assets [Releases page][ghrelease] on GitHub.
+
+4. Recompute the SHA256 hash of each artifact with e.g.  `shasum -a 256
```
Roasbeef, line 67: On the `lnd` side, we've recently started to use commands like this that wrap everything all into one: 
```
wget https://dl.google.com/go/go1.24.9.linux-amd64.tar.gz

echo "5b7899591c2dd6e9da1809fde4a2fad842c45d3f6b9deb235ba82216e31e34a6  go1.24.9.linux-amd64.tar.gz" | sha256sum --check
```

make/builder.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.24.9-bookworm
+FROM golang:1.24.9-bookworm@sha256:737b40b61ce956d738bed59f18ba854d8d67e7a4c4fa63f64437f4c70247ac5b
```
Roasbeef, line 1: 👍

scripts/release.sh
jtobin: shasum relies on it:

```
$ cat $(which shasum) | head -10
#!/usr/bin/perl
    eval 'exec /usr/bin/perl -S $0 ${1+"$@"}'
	if 0; # ^ Run only under a shell
#!perl

	## shasum: filter for computing SHA digests (ref. sha1sum/md5sum)
	##
	## Copyright (C) 2003-2018 Mark Shelor, All Rights Reserved
	##
	## Version: 6.02
```

If it's not available (rarely, but can happen) then we check for GNU coreutils' sha256sum and use that instead.

.github/workflows/release.yaml
jtobin: Yeah good point re: available space, I'll test by pushing a dev tag to this branch.

jtobin:
I've since tested CI with the changes here and can confirm that everything works (no disk issues, etc.). A stripped-down log for the release workflow can be seen [here](https://github.com/lightninglabs/taproot-assets/actions/runs/20023049477/job/57414058452), demonstrating that the image builds, and that the release artifacts are produced as expected.

.github/workflows/release.yaml
jtobin: I've since tested this and can confirm it works as expected. :+1:

(See [log](https://github.com/lightninglabs/taproot-assets/actions/runs/20023049477/job/57414058452)).

litbot-9000:
@ffranr: review reminder

PR #1896, opened 2025-12-03
proof: improve re-org proof replacement, strengthen error handling, expand test coverage

ffranr:
This PR tightens the proof replacement path used during block re-org handling. It explicitly ensures that we no longer ignore errors that should be surfaced, giving more reliable behaviour during re-org updates.

#### Key changes

* Refactor replacement to a pure `ReplaceProofInFiles` function that updates supplied proof files and validates asset IDs.
* Remove implicit side effects. Callers now fetch candidate proofs and import updates explicitly during re-org handling.
* Strengthen error handling across proof encoding, decoding, lookup, and file validation. All errors are now surfaced instead of being ignored.
* Add unit tests covering multiple proof-replacement scenarios triggered by re-org events.

### DIFF
--- a/proof/file.go
+++ b/proof/file.go
@@ -107,7 +107,8 @@ func NewFile(v Version, proofs ...Proof) (*File, error) {
 
 		proofBytes, err := proof.Bytes()
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("encoding proof (idx=%d): %w",
+				idx, err)
 		}
 
 		linkedProofs[idx] = &hashedProof{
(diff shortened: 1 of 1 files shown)


jtobin (approved):
>     - The replacement logic is now side-effect free and focused on
>       transformation, making it easier to unit test.
>     - Fetching and importing responsibilities are explicit at call sites,
>       improving clarity and making state changes easier to reason about.

😍

tapgarden/planter.go
```diff
@@ -3068,17 +3068,46 @@ func (c *ChainPlanter) updateMintingProofs(proofs []*proof.Proof) error {
 	ctx, cancel := c.WithCtxQuitNoTimeout()
 	defer cancel()
 
+	// This is a bit of a hacky part. If we have a chain of transactions
+	// that were re-organized, we can't verify the whole chain until all of
+	// the transactions were confirmed and all proofs were updated with the
+	// new blocks and merkle roots. So we'll skip the verification here
+	// since we don't know if the whole chain has been updated yet (the
+	// confirmations might come in out of order).
+	// TODO(guggero): Find a better way to do this.
```
GeorgeTsagk, line 3096: Was this todo created or moved from elsewhere? Should un-assign from guggero

ffranr: This comment and TODO was pre-existing and still applicable.

> un-assign from guggero

I don't think we treat these as assignment but authorship (which can be a little hard to track if code moves)?

PR #1904, opened 2025-12-05
tapfreighter: pre-broadcast guard for split root witnesses

ffranr:
This PR adds `ChainPorter` hardening following from this bug: https://github.com/lightninglabs/taproot-assets/pull/1897

Add a pre-broadcast verification step to ensure split leaf outputs embed a split root that carries a populated TxWitness, preventing invalid split commitments from being broadcast.

### DIFF
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -1421,6 +1421,12 @@ func (p *ChainPorter) verifyVPackets(ctx context.Context,
 			return fmt.Errorf("verify packet input proofs "+
 				"(vpkt_idx=%d): %w", pktIdx, err)
 		}
+
+		err = verifySplitCommitmentWitnesses(*vPkt)
+		if err != nil {
+			return fmt.Errorf("verify split commitment "+
+				"witnesses (vpkt_idx=%d): %w", pktIdx, err)
+		}
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


tapfreighter/chain_porter.go
```diff
@@ -1426,22 +1446,62 @@ func (p *ChainPorter) verifyPacketInputProofs(ctx context.Context,
...
+			!vOut.Asset.HasSplitCommitmentWitness() {
+
+			continue
+		}
+
+		splitCommitment := vOut.Asset.PrevWitnesses[0].SplitCommitment
+		if splitCommitment == nil {
+			return fmt.Errorf("output missing split commitment "+
+				"(output_idx=%d)", outIdx)
+		}
```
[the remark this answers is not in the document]
GeorgeTsagk: I personally would keep the extra sanity check

tapfreighter/chain_porter.go
```diff
@@ -1458,6 +1464,45 @@ func (p *ChainPorter) verifyPacketInputProofs(ctx context.Context,
...
+		if len(root.PrevWitnesses) == 0 {
+			return fmt.Errorf("output split root has no prev "+
+				"witnesses (output_idx=%d)", outIdx)
+		}
+
+		hasWitness := fn.Any(
+			root.PrevWitnesses, func(wit asset.Witness) bool {
+				return len(wit.TxWitness) > 0
+			},
+		)
```
GeorgeTsagk, line 1496: not sure how trivial it would be to extend beyond the simple check of `len(wit.TxWitness) > 0`

but for this scope this is definitely sufficient

GeorgeTsagk: Given that we wrapped everything around a single func now we can extend easily in the future

PR #1905, opened 2025-12-08
Orphan UTXO filter and limit

darioAnongba:
- Filter out orphan UTXOs with missing signing information: In prior versions, we didn't store `KeyFamily` and `KeyIndex` for orphan UTXOs. When both values are 0, LND fails to sign the transaction. This PR filters out these UTXOs to prevent signing failures.

- Limit the number of orphan UTXOs fetched: Previously, there was no limit on how many orphan UTXOs could be included in a transaction. If too many were fetched, the transaction could become too large. This PR adds a limit of 20 orphan UTXOs per transaction (`MaxOrphanUTXOs`).

- Enable `wallet.sweep-orphan-utxos` by default: Changed the default from `false` to `true`. This ensures tombstone and burn outputs are automatically swept during onchain transactions.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -39,6 +39,15 @@
   `AnchorVirtualPsbts`. A new configuration is available to control the sweeping 
   via the flag `wallet.sweep-orphan-utxos`.
 
+- [Improve orphan UTXO sweeping](https://github.com/lightninglabs/taproot-assets/pull/1905):
+  Fixed two issues with fetching orphan UTXOs for sweeping during transaction
+  building:
+  - Added filtering to exclude orphan UTXOs with missing signing information
+    (KeyFamily=0 and KeyIndex=0). These UTXOs were created in prior versions
+    that didn't store this information, causing LND to fail when signing.
+  - Added a limit (`MaxOrphanUTXOs = 20`) to prevent transactions from becoming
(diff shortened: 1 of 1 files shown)


jtobin (approved):
(Nit: I would avoid long lines in commit messages, e.g. in 453922428ec.)

LGTM. :+1: :+1:

tapdb/assets_store_test.go
```diff
@@ -3383,3 +3383,195 @@ func TestUpsertAssetsWithSplitCommitments(t *testing.T) {
...
+				for _, orphan := range orphans {
+					require.NotZero(
+						t, orphan.InternalKey.Family,
+					)
+					require.NotZero(
+						t, orphan.InternalKey.Index,
+					)
+				}
+			},
+		},
```
GeorgeTsagk, line 3448: let's add at least one more composite test case, where we have mixed entries with and without signing info, that sum up to more than `MaxOrphanUTXOs`, something like:
- 20 invalid entries + 1 valid one: expect 1 valid
- 20 invalid entries + 21 valid ones: expect 20 valid

GeorgeTsagk (approved):
LGTM

have some suggestions to extend the tests, otherwise looking good :green_circle:

tapdb/assets_store_test.go
darioAnongba: yes, added 2 new test cases.

PR #1912, opened 2025-12-11
rfq: add `InternalPortfolioPilot` and initial `ResolveRequest` wiring

ffranr:
- Introduce the `InternalPortfolioPilot` and shared RFQ `Request` interface for buy/sell flows.
- Add portfolio pilot-based `ResolveRequest` handling for buy and sell requests, invoking the price oracle.
- Prepare negotiator/manager configuration for further pilot integration in follow-up work.

### DIFF
--- a/rfq/negotiator.go
+++ b/rfq/negotiator.go
@@ -410,77 +410,44 @@ func (n *Negotiator) HandleIncomingSellRequest(
 		}
 	}
 
-	// Reject the quote request if a price oracle is unavailable.
-	if n.cfg.PriceOracle == nil {
-		msg := rfqmsg.NewReject(
-			request.Peer, request.ID,
-			rfqmsg.ErrPriceOracleUnavailable,
-		)
-		go sendOutgoingMsg(msg)
-		return nil
-	}
-
-	// The sell request is attempting to sell some amount of an asset to our
-	// node. Here we ensure that we have a suitable buy offer for the asset.
-	// A buy offer is the criteria that this node uses to determine whether
-	// it is willing to buy a particular asset (before price is considered).
-	// At this point we can handle the case where this node does not wish
-	// to buy some amount of a particular asset regardless of its price.
-	//
-	// TODO(ffranr): Reformulate once BuyOffer fields have been revised.
-	offerAvailable := n.HasAssetBuyOffer(
-		request.AssetSpecifier, uint64(request.PaymentMaxAmt),
-	)
-	if !offerAvailable {
-		log.Infof("Would reject sell request: no suitable buy offer, " +
-			"but ignoring for now")
-
-		// TODO(ffranr): Re-enable pre-price oracle rejection (i.e.
-		//  reject on missing offer)
-
-		// If we do not have a suitable buy offer, then we will reject
-		// the asset sell quote request with an error.
-		// reject := rfqmsg.NewReject(
-		//	request.Peer, request.ID,
-		//	rfqmsg.ErrNoSuitableBuyOffer,
-		// )
-		// go sendOutgoingMsg(reject)
-		//
-		// return nil
-	}
-
-	// Query the price oracle asynchronously using a separate goroutine.
-	// The price oracle might be an external service, responses could be
+	// Query the portfolio pilot synchronously using a separate goroutine.
+	// The portfolio pilot might be an external service, responses could be
 	// delayed.
 	n.Goroutine(func() error {
-		var peerID fn.Option[route.Vertex]
-		if n.cfg.SendPeerId {
-			peerID = fn.Some(request.Peer)
-		}
+		ctx, cancel := n.WithCtxQuitCustomTimeout(
+			DefaultPortfolioPilotTimeout,
+		)
+		defer cancel()
 
-		// Query the price oracle for a buy price. This is the price we
-		// are willing to pay for the asset that our peer is trying to
-		// sell to us.
-		assetRate, err := n.queryBuyFromPriceOracle(
-			request.AssetSpecifier, fn.None[uint64](),
-			fn.Some(request.PaymentMaxAmt), request.AssetRateHint,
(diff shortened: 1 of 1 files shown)


rfq/negotiator.go
```diff
@@ -319,77 +334,47 @@ func (n *Negotiator) HandleIncomingBuyRequest(
...
-	// Query the price oracle asynchronously using a separate goroutine.
-	// The price oracle might be an external service, responses could be
+	// Query the portfolio pilot synchronously using a separate goroutine.
+	// The portfolio pilot might be an external service, responses could be
 	// delayed.
-	n.Wg.Add(1)
-	go func() {
-		defer n.Wg.Done()
+	n.Goroutine(func() error {
+		ctx, cancel := n.WithCtxQuitNoTimeout()
```
darioAnongba: using `NoTimeout` here is on purpose? Not too problematic but what if the `PriceOracle` is unresponsive?

rfq/portfolio_pilot.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+// ResolveResp captures the portfolio pilot's resolution decision for an RFQ. It
+// carries either an accepted asset rate quote or a structured rejection reason.
+type ResolveResp struct {
+	// outcome holds either the accepted asset rate (left) or the rejection
+	// error (right).
+	outcome fn.Either[rfqmsg.AssetRate, rfqmsg.RejectErr]
```
darioAnongba, line 18: nice

rfq/portfolio_pilot.go
```diff
@@ -0,0 +1,166 @@
...
+	// acceptable asset rate or a quote rejection error. Errors are reserved
+	// for unexpected failures while evaluating the request.
+	ResolveRequest(context.Context, rfqmsg.Request) (ResolveResp,
+		error)
+}
+
+// InternalPortfolioPilotConfig holds settings for the built-in pilot that uses
+// a price oracle for pricing decisions.
+type InternalPortfolioPilotConfig struct {
+	// PriceOracle supplies pricing data. If nil, the pilot rejects requests
```
darioAnongba, line 76: why do we accept `nil` here?

rfq/portfolio_pilot.go
```diff
@@ -0,0 +1,166 @@
...
+// ResolveRequest resolves a quote request by querying the configured price
+// oracle. It accepts with a quote when a valid rate is returned, rejects when
+// no oracle is configured, and errors on unexpected oracle failures or missing
+// rates.
+func (p *InternalPortfolioPilot) ResolveRequest(ctx context.Context,
+	request rfqmsg.Request) (ResolveResp, error) {
+
+	var zero ResolveResp
+
+	if p.cfg.PriceOracle == nil {
```
darioAnongba: ok so this is some kind of `Noop` where we can pass `nil` to simply reject everything?

I'm not a fan tbh as this could be a config where we specify that the PortfolioPilot is `disabled` (e.g. `cfg.RejectAll = true`).

rfq/portfolio_pilot.go
```diff
@@ -0,0 +1,166 @@
...
+		return NewRejectResolveResp(
+			rfqmsg.ErrPriceOracleUnavailable,
```
darioAnongba: for me there is a difference between `ErrPriceOracleUnavailable`, signifying that the Oracle is down or unreachable and what we do here, which is not an error, we voluntarily reject everything and it's the expected behaviour.

rfq/portfolio_pilot.go
```diff
@@ -0,0 +1,166 @@
...
+		)
+		if err != nil {
+			return zero, fmt.Errorf("query sell price: %w", err)
+		}
+
+	case *rfqmsg.SellRequest:
+		var err error
+		resp, err = p.cfg.PriceOracle.QueryBuyPrice(
+			ctx, req.AssetSpecifier, fn.None[uint64](),
+			fn.Some(req.PaymentMaxAmt), fn.None[rfqmsg.AssetRate](),
```
darioAnongba: Intentional that `request.AssetRateHint` is not passed here anymore?

rfq/negotiator.go
```diff
@@ -413,84 +398,47 @@ func (n *Negotiator) HandleIncomingSellRequest(
...
-
-	// The sell request is attempting to sell some amount of an asset to our
-	// node. Here we ensure that we have a suitable buy offer for the asset.
-	// A buy offer is the criteria that this node uses to determine whether
-	// it is willing to buy a particular asset (before price is considered).
-	// At this point we can handle the case where this node does not wish
-	// to buy some amount of a particular asset regardless of its price.
-	//
-	// TODO(ffranr): Reformulate once BuyOffer fields have been revised.
-	offerAvailable := n.HasAssetBuyOffer(
```
darioAnongba, line 434: `HasAssetBuyOffer` is still used? I see you removed `HasAssetSellOffer` but not this one?

GeorgeTsagk: +1, it's unused so should also remove

rfq/portfolio_pilot.go
ffranr: I will pass in.

rfq/portfolio_pilot.go
ffranr: We will not accept `nil`. I will change.

rfq/portfolio_pilot.go
ffranr: Good point. We now ensure `p.cfg.PriceOracle != nil` at init.

rfq/negotiator.go
ffranr: I've added timeout when calling into portfolio pilot. Good point, thanks.

darioAnongba (approved):
LGTM

Issue #1909, opened 2025-12-11
[bug]: Tapd crashes when creating taproot assets addres with invalid mailbox


Liongrass:
### Background

I generated a taproot assets address with the command:

`tapcli addrs new --group_key 03e388268bb2d40e91730966cfafc0be0fa300ac861045f20f14bead238d393b35 --proof_courier_addr authmailbox+universerpc://signet.laisee.org:8443`

This succeeds, but as there is no universe listening on this port, tapd crashes with the following error:

`received critical error from sub-server (taproot-assets), shutting down: unable to start mailbox subscription: error during authentication, before sending subscribe: server sent unexpected error`

I have attached logs below.

[2025-12-10_tapd.log](https://github.com/user-attachments/files/24090970/2025-12-10_tapd.log)

### Your environment

tapd 0.7.0-alpha commit=v0.7.0
litd 0.16.0-alpha (integrated mode)

```sh
tapcli getinfo                       # version of `tapd`, `lnd`, and network
uname -mrsv                          # operating system 
bitcoind --version || btcd --version # version of `btcd`, `bitcoind`, or other backend
```

Liongrass:
And it seems this issue persists after a restart, too.

jtobin:
Was signet.laisee.org a valid hostname for that universe at the time? (I'm used to seeing universe.signet.laisee.org, which I know to be valid.)

ffranr:
It is possible that the server-connection check passed when the address was created, but `Custodian` (`tapgarden/custodian.go`) later failed to connect when setting up the mailbox subscription. Either way, errors from `Custodian.watchInboundAssets` are sent to the main error channel, which causes tapd to treat them as critical and shut down. That is undesirable. We should separate critical from non-critical errors in `Custodian`, ensure failed subscriptions are retried, and add missing itest coverage.

Liongrass:
Maybe what contributed to the bug is that `signet.laisee.org` points to a `tapd` daemon, but not one configured as a universe. Only `universe.signet.laisee.org` is configured as a universe.

ffranr:
Status update and next steps:

We need to add backoff to the subscription step. The initial connection check triggered from `NewAddr` (`CheckUniverseRpcCourierConnection`) also needs hardening. In particular, it should verify that the target server actually supports `authmailbox`, not just that it is reachable via a ping-style check. The earlier failure on Leo’s server appears to have been a misconfiguration.

[...]

PR #1915, opened 2025-12-12
Keep tapd running when authmailbox subscription fails

ffranr:
Partial fix for https://github.com/lightninglabs/taproot-assets/issues/1909
(EDIT: this is a partial fix only)

---

- Emit explicit address import success/failure events from the custodian, have `NewAddr` RPC subscribe and wait for the outcome with a timeout, and ignore unsupported events in the RPC stream so import errors don’t crash the daemon.
- Make mailbox connection retries configurable (MaxConnectAttempts) and treat backoff NumTries as optional with a default, reducing the chance of runaway retries while still surfacing failures to callers.
- Add an itest that creates a V2 address against an unreachable authmailbox (with the upfront connect check skipped) to verify tapd stays up after the subscription failure.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -62,6 +62,11 @@
   default RPC permissions for RPC endpoints universerpc.Universe/Info and
   /authmailboxrpc.Mailbox/MailboxInfo.
 
+- [PR#1915](https://github.com/lightninglabs/taproot-assets/pull/1915)
+  `NewAddr` now registers a custodian subscriber and waits for the address
+  import result (with a timeout) before returning, surfacing mailbox courier
+  import failures instead of racing and returning success early.
+
 ## tapcli Updates
 
 ## Config Changes
@@ -87,6 +92,11 @@
 ## BIP/bLIP Spec Updates
 
 ## Testing
+
+- [PR#1915](https://github.com/lightninglabs/taproot-assets/pull/1915)
+  Add an integration test that verifies tapd stays running when V2 address
+  creation hits an unreachable mailbox courier with the upfront connection
+  check skipped, ensuring mailbox subscription failures do not crash tapd.
 
 ## Database
 


itest/addrs_v2_test.go
```diff
@@ -827,3 +827,65 @@ func testAddressV2WithGroupKeyRestart(t *harnessTest) {
...
+		[]*mintrpc.MintAssetRequest{simpleAssets[0]},
+	)
+	asset := assets[0]
+
+	badCourier := fmt.Sprintf(
+		"%s://%s", proof.AuthMailboxUniRpcCourierType,
+		"nonexistent.invalid:65500",
+	)
+
+	t.Logf("Trying to create an address with bad mailbox %s", badCourier)
```
darioAnongba, line 867: Please no logging in the `itests` 🙏 . The test purpose and its assertions are the only thing we need.

itest/addrs_v2_test.go
```diff
@@ -827,3 +827,65 @@ func testAddressV2WithGroupKeyRestart(t *harnessTest) {
...
+		t.t, err.Error(), "unable to start mailbox subscription",
+	)
+
+	t.Logf("Ensure tapd is still up and running")
+	infoCtx, infoCancel := context.WithTimeout(ctxb, defaultWaitTimeout)
+	defer infoCancel()
+
+	_, err = tapd.GetInfo(infoCtx, &taprpc.GetInfoRequest{})
+	require.NoError(t.t, err)
+	t.Logf("Tapd is still up and running")
```
darioAnongba, line 890: Same remark, this is what the test does so logging is useless if the test passes.

itest/addrs_v2_test.go
ffranr: Why no logging? Sometime I find log messages helpful when debugging a failing itest.

tapgarden/custodian.go
```diff
@@ -465,7 +561,26 @@ func (c *Custodian) watchInboundAssets() {
 		var err error
 		select {
 		case newAddr := <-c.addrSubscription.NewItemCreated.ChanOut():
-			err = c.importAddrToWallet(ctxStream, newAddr)
+			importErr := c.importAddrToWallet(ctxStream, newAddr)
+			switch {
```
jtobin, line 565: As there are only two branches, this one might be better as an if/else block.

authmailbox/client.go
```diff
@@ -77,6 +77,11 @@ type ClientConfig struct {
 
 	// MaxBackoff is the maximum time waited between connection attempts.
 	MaxBackoff time.Duration
+
+	// MaxConnectAttempts optionally limits the number of connection
+	// attempts before giving up. If 0, the client uses the default
+	// internal retry limit.
+	MaxConnectAttempts uint32
```
jtobin, line 84: I traced through this a bit, and I *think* this one might not actually use the default retry limit as intended, but instead a plain 0. It looks like it's set as follows in tapgarden/custodian.go:

```
return &Custodian{
	[..]
	mboxSubscriptions: mbox.NewMultiSubscription(
		mbox.ClientConfig{
			[..]
			MaxConnectAttempts: backoffCfg.NumTries,
		},
	),
	[..]
}
```

In proof/courier.go, there's some explicit logic to handle the NumTries == 0 case, but this doesn't appear to be propagated back to the config itself:

```
var (
	[..]
	numTries = b.cfg.NumTries
	[..]
)

if numTries == 0 {
	numTries = DefaultProofTransferNumTries
}
```

authmailbox/receive_subscription.go
```diff
@@ -202,13 +197,13 @@ func (s *receiveSubscription) IsSubscribed() bool {
...
 func (s *receiveSubscription) connectServerStream(ctx context.Context,
-	initialBackoff time.Duration, numRetries int) error {
+	initialBackoff time.Duration, numRetries uint32) error {
 
 	var (
 		backoff = initialBackoff
 		err     error
 	)
-	for i := 0; i < numRetries; i++ {
+	for i := uint32(0); i < numRetries; i++ {
```
jtobin, line 220: Re: my previous comment, if MaxConnectAttempts defaults to 0, then this predicate won't be met at all.

authmailbox/client.go
ffranr: Thanks for looking carefully here. I think you're right. So I've added a few more lines which will will hopefully give us more confidence that 0 is handled correctly.

itest/addrs_v2_test.go
darioAnongba: I agree on logging locally while debugging but I tend to disagree on the usefulness of logging in tests because IMO, if you do not have enough information to know what went wrong between the service logs (tapd logs) + assertion messages, then you should improve the assertions or logs.

Of course, this is theoretical and if you consider that logging is valuable, then go for it, but it should bring value.

tapgarden/custodian.go
ffranr: I’m using switch mainly for readability. Just flatter and easier for me to scan. I think the expressionless switch pattern is idiomatic Go and used fairly often in the standard library. But I could be wrong.

authmailbox/receive_subscription.go
```diff
@@ -104,7 +102,14 @@ func (s *receiveSubscription) connectAndAuthenticate(ctx context.Context,
 
 	log.DebugS(ctx, "Establishing initial connection to server")
 
-	err := s.connectServerStream(ctx, initialBackoff, reconnectRetries)
+	maxConnectAttempts := s.cfg.MaxConnectAttempts
+	if maxConnectAttempts == 0 {
```
jtobin, line 106: :+1:

jtobin (approved):
LGTM. :+1:

PR #1920, opened 2025-12-18
tapchannel+tapfreighter: add parcel height hint

### LINKED ISSUE #1848: [bug]: Assets Not Found On-Chain After Normal Forced Closure to 'Closedchannel' State
I have a node that opened an asset channel, which entered the pending_force_closing_channels state and eventually moved to closedchannels. However, when I used the command
` tapcli --network mainnet --rpcserver localhost:8443 a l`, 
I couldn’t find the on-chain assets. Additionally, the data shown in unconfirmed_transfers looks unusual.
```
{
    "assets": [],
    "unconfirmed_transfers": "23",
    "unconfirmed_mints": "0"
}
```
This is the data returned by my `lncli closedchannels `command.
```
{
    "channels": [
        {
            "channel_point": "7b57e36406cf36d5d8d5846adc7ab1890605cc00632419f6445e3fc53a0b327e:0",
            "chan_id": "999601205214773248",
            "chain_hash": "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f",
            "closing_tx_hash": "e63fe180c665ed237f270c3fd07d6ed973bf9aaace71197ffc632579c52caf28",
            "remote_pubkey": "02c953421bc7f07be6052920e46843d11e6d3ffc9986177c91f140d76c6ed3a3d4",
            "capacity": "20000",
            "close_height": 914486,
            "settled_balance": "10000",
            "time_locked_balance": "0",
            "close_type": "REMOTE_FORCE_CLOSE",
            "open_initiator": "INITIATOR_REMOTE",
            "close_initiator": "INITIATOR_REMOTE",
            "resolutions": [
                {
                    "resolution_type": "COMMIT",
                    "outcome": "CLAIMED",
                    "outpoint": {
                        "txid_bytes": "28af2cc5792563fc7f1971ceaa9abf73d96e7dd03f0c277f23ed65c680e13fe6",
                        "txid_str": "e63fe180c665ed237f270c3fd07d6ed973bf9aaace71197ffc632579c52caf28",
                        "output_index": 3
                    },
                    "amount_sat": "10000",
                    "sweep_txid": "a65a1a6879bcdfacf0620354fbbdde54ce77fdacb62aa360f4e2579c20d750f4"
                }
            ],
            "alias_scids": [
                "17592186044416000001"
            ],
            "zero_conf_confirmed_scid": "0"
        }
    ]
}
```
This is the output returned by the command tapcli --network mainnet --rpcserver localhost:8443 a t.

[assetTx.txt](https://github.com/user-attachments/files/22878836/assetTx.txt)

jtobin:
Should resolve #1848, at least future instances of such. More detail below.

Adds a HeightHint() method to the Parcel interface that allows a parcel to include an optional block height hint for its corresponding anchor transaction's confirmation, if this information is available (e.g. via the 'CommitTxBlockHeight' supplied by lnd).

If a parcel contains a height hint, then we use that in the chain porter's SendStateStorePreBroadcast state, falling back to the current block height otherwise. This is expected to resolve the issue described in #1848, in which assets were never found on-chain after a remote-initiated force close in a very particular environment.

My working hypothesis regarding the linked issue is that the node, which appears to have been running with a Neutrino backend, was offline when the force close occurred, catching up only after the force close transaction had been confirmed. (In the logs supplied with that issue, one can see that immediately prior to and after the 'Unilateral close of ChannelPoint..' entry, blocks are being received at a rate of about one per second.)

In this situation, the force close transaction appears to have proceeded through the chain porter's state machine normally, eventually reaching the SendStateStorePreBroadcast state, where its transfer parcel was augmented with a hint of the *current* block height for its anchor transaction. It then entered the SendStateWaitTxConf state with that hint, and, since the node's chain backend didn't support txindex, waited for transaction confirmation from that block forward. Since the force close had confirmed *prior* to that block, the parcel just became stuck in that state, never proceeding to a later state in which the transfer proofs would have been persisted to the local proof archive.

When the sweep transaction parcels proceeded through the porter's state machine, they thus hit an error during the SendStateStorePostAnchorTxConf state, since the proofs for the requisite force close transaction couldn't be found.

The fix here, which applies specifically to the remote-initiated force close case, uses the 'CommitTxBlockHeight' information that lnd supplies to 'importCommitTx' as a height hint if it's available, which is the height of the actual block that included the commitment transaction. This should fix *future* instances of this issue, but doesn't mitigate existing stuck force-close parcels.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -21,6 +21,11 @@
 
 # Bug Fixes
 
+* [PR #1920](https://github.com/lightninglabs/taproot-assets/pull/1920)
+  addresses a bug in which Neutrino-backed nodes could fail to import
+  transfer proofs for remote-initiated force close transactions if they
+  were not online to see them broadcast.
+
 # New Features
 
 ## Functional Enhancements
@@ -36,7 +41,7 @@
 - [Garbage collection of orphaned UTXOs](https://github.com/lightninglabs/taproot-assets/pull/1832)
   by sweeping tombstones and burn outputs when executing onchain transactions.
   Garbage collection will be executed on every burn, transfer or call to
-  `AnchorVirtualPsbts`. A new configuration is available to control the sweeping 
+  `AnchorVirtualPsbts`. A new configuration is available to control the sweeping
   via the flag `wallet.sweep-orphan-utxos`.
 - [PR](https://github.com/lightninglabs/taproot-assets/pull/1899) tapd now
   treats HTLC interceptor setup failures as fatal during RFQ subsystem startup.
@@ -81,7 +86,7 @@
 ## Config Changes
 
 - [PR#1870](https://github.com/lightninglabs/taproot-assets/pull/1870)
-  The `proofs-per-universe` configuration option is removed. New option 
+  The `proofs-per-universe` configuration option is removed. New option
   `max-proof-cache-size` sets the proof cache limit in bytes and accepts
   human-readable values such as `64MB`.
 


Roasbeef:
> If a parcel contains a height hint, then we use that in the chain porter's SendStateStorePreBroadcast state, falling back to the current block height otherwise. This is expected to resolve the issue described in https://github.com/lightninglabs/taproot-assets/issues/1848, in which assets were never found on-chain after a remote-initiated force close in a very particular environment.

Have they attempted to reset the height hint cache on their `lnd` node? If this works for them, then that'll increase our confidence that this PR is on the right track.

tapfreighter/chain_porter.go
```diff
@@ -1768,15 +1768,29 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
...
+			currentHeight, err := p.cfg.ChainBridge.CurrentHeight(
+				ctx,
+			)
+			if err != nil {
+				p.unlockInputs(ctx, &currentPkg)
+
+				return nil, fmt.Errorf("unable to get current "+
+					"height: %w", err)
+			}
+			heightHint = currentHeight
```
Roasbeef: An alternative way to write this would be: 
```go
heightHint, err := parceltHint.UnwrapOrFuncErr(func () (uint32, error) {
    return p.cfg.ChainBridge.CurrentHeight(tx)
})
``

**Comment by jtobin** [comment]
> Have they attempted to reset the height hint cache on their `lnd` node? If this works for them, then that'll increase our confidence that this PR is on the right track.

Having taken a look at this, it seems that lnd will use the incorrect hint from tapd regardless:

(lnd/chainntnfs/txnotifier.go L645)

``` 
startHeight := ntfn.HeightHint                                     // incorrect supplied hint
hint, err := n.confirmHintCache.QueryConfirmHint(ntfn.ConfRequest) // cached hint
if err == nil {
	if hint > startHeight {                                        // if cached hint > supplied hint
		Log.Debugf("Using height hint %d retrieved from cache "+
			"for %v instead of %d for conf subscription",
			hint, ntfn.ConfRequest, startHeight)
		startHeight = hint                                         // use (incorrect) cached hint
	}
} else if err != ErrConfirmHintNotFound {                          // otherwise continue to
	Log.Errorf("Unable to query confirm hint for %v: %v",          // use incorrect supplied hint
		ntfn.ConfRequest, err)
}
```

File: tapfreighter/chain_porter.go (hunk 1 above)
Comment (jtobin) [reply tapfreighter/chain_porter.go]: Fixed in 3082fa62.

**Comment by litbot-9000** [comment]
@ffranr: review reminder

File: tapfreighter/chain_porter.go [hunk 2]
```diff
@@ -1768,15 +1768,25 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
...
-		// a height hint.
+		// a height hint. If the parcel provides its own height hint,
+		// we'll use that instead.
 		ctx, cancel := p.WithCtxQuit()
 		defer cancel()
-		currentHeight, err := p.cfg.ChainBridge.CurrentHeight(ctx)
+
+		parcelHint := currentPkg.Parcel.HeightHint()
+		parcelHint.WhenSome(func(h uint32) {
+			log.Infof("Using parcel-provided height hint: %d", h)
```
ffranr: I think this should be debug level.

docs/release-notes/release-notes-0.8.0.md
```diff
@@ -21,6 +21,11 @@
 
 # Bug Fixes
 
+* [PR #1920](https://github.com/lightninglabs/taproot-assets/pull/1920) fixes
+  a bug in which Neutrino-backed nodes could fail to import transfer
+  proofs for remote-initiated force close transactions if they were not
+  online to see them broadcast.
```
ffranr: i think if we frame this PR as: allow specify block height for pre-anchored, more accurate height hint, etc. That we could ship this as is. It's a good addition regardless of the bug IMO.

ffranr (approved):
Looks like a good addition regardless of the bug IMO.

tapfreighter/chain_porter.go
jtobin: Fixed in 31eba90dc09b5f017d7aa9b278490784984f8ce6.

docs/release-notes/release-notes-0.8.0.md
jtobin: Agreed, though it *should* indeed address the bug, which is probably the main item of interest in the release notes. I'll actually change the language here to "addresses a bug," which I think is more accurate.

PR #1921, opened 2025-12-22
Forwarding history

### LINKED ISSUE #1005: [feature]: Provide "forwarding history" for Edge Nodes

darioAnongba:
This PR adds RFQ forward history tracking, allowing edge nodes to monitor and query their asset forwarding activity and fee revenue. Edge nodes need visibility into their RFQ forwarding activity to:
- Track fee revenue earned from routing asset payments
- Monitor forwarding volume
- Audit historical forwarding events
- Generate reports on routing performance

## Changes

- New `forwards` table to store forward events
- `ForwardingHistory`: New RPC endpoint 
  - Filters: min/max timestamp, peer pubkey, asset ID, group key
  - Pagination: limit and offset support
  - Returns: forward records with joined policy data (peer, asset info)
- `tapcli rfq forwardinghistory`: Query forward history from CLI with filters `--min-timestamp`, `--max-timestamp`, `--peer`, `--asset-id`, `--group-key`, `--limit`, `--offset`

Fixes #1005

- - -
This change is [Reviewable](https://reviewable.io/reviews/lightninglabs/taproot-assets/1921)

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -56,6 +56,13 @@
 
 ## Functional Enhancements
 
+- [Forwarding History Tracking](https://github.com/lightninglabs/taproot-assets/pull/1921):
+  Routing nodes can now track and query historical asset forwarding events.
+  When a node successfully routes an asset payment, the forward event is logged.
+  This provides edge nodes with an audit trail of their swap activity.
+  **Note:** For full functionality, it is highly recommended to start LND with
+  the `--store-final-htlc-resolutions` flag enabled, which is disabled by default.
+
 ## RPC Additions
 
 - [PR#1960](https://github.com/lightninglabs/taproot-assets/pull/1960)
@@ -65,6 +72,20 @@
 
 - [PR#1960](https://github.com/lightninglabs/taproot-assets/pull/1960)
   Add `tapcli bakemacaroon` to bake custom macaroons with offline caveats.
+
+- [ForwardingHistory RPC](https://github.com/lightninglabs/taproot-assets/pull/1921):
+  New RPC endpoint `rfqrpc.ForwardingHistory` allows querying historical
+  forwarding events with filtering and pagination support. Filters include:
+  timestamp range (min/max), peer public key, asset ID, and asset group key.
+
+## tapcli Additions
+
+- [tapcli rfq forwardinghistory](https://github.com/lightninglabs/taproot-assets/pull/1921):
+  New CLI command `tapcli rfq forwardinghistory` (alias: `f`) to query forwarding event
+  history. Supports flags for filtering by timestamp (`--min-timestamp`,
+  `--max-timestamp`), peer (`--peer`), asset ID (`--asset-id`), and asset
+  group key (`--group-key`). Includes pagination support via `--limit` and
+  `--offset` flags.
 
 # Improvements
 
@@ -167,7 +188,14 @@
   creation hits an unreachable mailbox courier with the upfront connection
   check skipped, ensuring mailbox subscription failures do not crash tapd.
 
+- [Forwarding History Integration Test](https://github.com/lightninglabs/taproot-assets/pull/1921):
+  New integration test `testForwardingEventHistory` verifies that forwarding events are
+  properly logged when routing asset payments.
+
 ## Database
+
+- [forwards table](https://github.com/lightninglabs/taproot-assets/pull/1921):
+  New database table `forwards` stores historical forwarding events.
 
 ## Code Health
 


tapdb/sqlc/migrations/000052_rfq_forwards.up.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE IF NOT EXISTS rfq_forwards (
+    id INTEGER PRIMARY KEY,
+
+    -- settled_at is the unix timestamp when the forward settled.
+    settled_at BIGINT NOT NULL,
```
GeorgeTsagk: Might be going a bit too far here, but what about also storing an "opened at" field?

This way we're exposing information to the portfolio pilot that can be of essence:
- Fast HTLCs are better than long-held HTLCs (especially for tap, where you don't want to have an open position for too long)
- We could also even store forward events that were never settled. So if someone is locking liquidity or performing a jam attack over the channel it can be detected by observing the rfq_forwards, which potentially may never have their `settled_at` field set (a `failed_at` field would be complementary here)

tapdb/sqlc/migrations/000052_forwards.up.sql
```diff
@@ -0,0 +1,27 @@
...
+    chan_id_out BIGINT NOT NULL,
+
+    -- htlc_id is the HTLC ID on the incoming channel.
+    htlc_id BIGINT NOT NULL,
+
+    -- asset_amt is the asset amount involved in this swap.
+    asset_amt BIGINT NOT NULL,
+
+    UNIQUE(chan_id_in, htlc_id)
+);
```
GeorgeTsagk, line 38: seems like you also removed the sats in/out fields that were previously here?

even though the fee calculation was off, I think it would be nice to keep those field for sanity

Right now you can derive the msat amount that was in theory forwarded by combining asset units with asset rate, but the actual recorded amount would be nice to have (these 2 can also be off by design)

tapdb/sqlc/migrations/000052_rfq_forwards.up.sql
```diff
@@ -1,8 +1,8 @@
 CREATE TABLE IF NOT EXISTS rfq_forwards (
     id INTEGER PRIMARY KEY,
 
-    -- settled_at is the unix timestamp when the forward settled.
```
GeorgeTsagk: Are the following commits meant to be fixups? If so some context on the fixup commit itself is needed, as many of the changes are not self explanatory (much of the previous diff is reverted)

GeorgeTsagk (commented):
Direction looking good

tapdb/sqlc/migrations/000052_rfq_forwards.up.sql
darioAnongba: Sorry I messed up my fixup, fixed now.

tapdb/sqlc/migrations/000052_forwards.up.sql
darioAnongba: added back

tapdb/sqlc/migrations/000052_rfq_forwards.up.sql
darioAnongba: added `failed_at` and `opened_at`.

tapdb/sqlc/queries/rfq.sql
```diff
@@ -44,3 +44,70 @@ SELECT
...
+WHERE chan_id_in = $2 AND htlc_id = $3;
+
+-- name: FailRfqForward :exec
+UPDATE rfq_forwards
+SET failed_at = $1
+WHERE chan_id_in = $2 AND htlc_id = $3;
+
+-- name: QueryRfqForwards :many
+SELECT
+    f.id,
```
Roasbeef: Style nit: can pack more of these on aggregated lines to reduce vertical length here.

tapdb/sqlc/queries/rfq.sql
```diff
@@ -44,3 +44,70 @@ SELECT
...
+    f.amt_out_msat,
+    p.policy_type,
+    p.peer,
+    p.asset_id,
+    p.asset_group_key,
+    p.rate_coefficient,
+    p.rate_scale
+FROM rfq_forwards f
+JOIN rfq_policies p ON f.rfq_id = p.rfq_id
+WHERE f.opened_at >= @opened_after
```
Roasbeef: So this requires the time based fields to always be set? Typically we'll use the `sqlc.narg` in combination with some logic operator to set a default value s.t it'll ignore this constraint if not set.

rfq/interface.go
```diff
@@ -18,3 +39,140 @@ type PolicyStore interface {
...
+
+	// PolicyType indicates whether this was a sale or purchase from the
+	// edge node's perspective.
+	PolicyType RfqPolicyType
+
+	// Peer is the counterparty peer's public key.
+	Peer route.Vertex
+
+	// AssetID is the specific asset ID (if set).
+	AssetID *asset.ID
```
Roasbeef: Can use `asset.Specifier` here.

tapdb/rfq_forwards.go
```diff
@@ -0,0 +1,322 @@
...
+
+	// Set default time bounds if not specified.
+	openedAfter := params.MinTimestamp
+	if openedAfter.IsZero() {
+		openedAfter = time.Unix(0, 0).UTC()
+	}
+
+	openedBefore := params.MaxTimestamp
+	if openedBefore.IsZero() {
+		openedBefore = MaxValidSQLTime
```
Roasbeef: Ah ok, I see you use a max value here to maintain the default behavior of ignoring the time bsaed constraints.

tapdb/rfq_forwards.go
```diff
@@ -0,0 +1,322 @@
...
+	}
```
Roasbeef: Bit of duplication here and above, can be lifted into a function for re-use. Could also consider making it an `fn.Option`, then you gain the `if zero <this> ` with `Unwrap`.

tapdb/rfq_forwards_test.go
```diff
@@ -0,0 +1,593 @@
...
+				AssetAmt:   500,
+				AmtInMsat:  42000,
+				AmtOutMsat: 41000,
+			}
+			input2 := rfq.ForwardInput{
+				OpenedAt:   time.Now(),
+				RfqID:      rfqID,
+				ChanIDIn:   100, // Same
+				ChanIDOut:  300,
+				HtlcID:     1, // Same
```
Roasbeef: Style nit: we usually don't do inline comments, and this one in particular doesn't add much.

rfq/order.go
```diff
@@ -57,6 +57,56 @@ func parseHtlcCustomRecords(customRecords map[uint64][]byte) (*rfqmsg.Htlc,
...
 
+// computeHtlcAssetAmount derives the asset amount carried by the HTLC from the
+// custom records or computes it from the msat amount and rate.
+func computeHtlcAssetAmount(ctx context.Context, policy Policy,
+	htlc lndclient.InterceptedHtlc,
+	specifierChecker rfqmsg.SpecifierChecker) (uint64, error) {
+
+	switch p := policy.(type) {
+	case *AssetSalePolicy:
+		// For asset sales, the edge node receives BTC and sends out
```
Roasbeef: One thing to note here is that there's _also_ a fee being paid/earned here. 

For the "sale" case, the fee is in BTC. 

For the "purchase" case, the fee is in assets. 

We should account for this to ensure edge nodes can fully audit where all the funds went.

rfq/order.go
```diff
@@ -57,6 +57,56 @@ func parseHtlcCustomRecords(customRecords map[uint64][]byte) (*rfqmsg.Htlc,
...
+		// assets. We compute the asset amount from the outgoing msat
```
Roasbeef: So this is the reverse of what the comment mentions above: sale defined as BTC in, assets out.

rfq/order.go
```diff
@@ -57,6 +57,56 @@ func parseHtlcCustomRecords(customRecords map[uint64][]byte) (*rfqmsg.Htlc,
...
+			ctx, p.AssetSpecifier, specifierChecker,
+		)
+		if err != nil {
+			return 0, fmt.Errorf("summing asset balance failed: "+
+				"%w", err)
+		}
+
+		return assetAmt.ToUint64(), nil
+
+	case *AssetForwardPolicy:
```
Roasbeef, line 101: This is the asset to asset scenario?

rfq/order.go
```diff
@@ -854,6 +936,51 @@ func (h *OrderHandler) handleIncomingHtlc(ctx context.Context,
...
+				ChanIDIn:   chanIdIn,
+				ChanIDOut:  htlc.OutgoingChannelID.ToUint64(),
+				HtlcID:     htlc.IncomingCircuitKey.HtlcID,
+				AssetAmt:   assetAmt,
+				AmtInMsat:  uint64(htlc.AmountInMsat),
+				AmtOutMsat: uint64(htlc.AmountOutMsat),
+			}
+
+			// Log the forward immediately when it opens.
+			id, err := h.forwardStore.LogForward(ctx, fwdInput)
```
Roasbeef: One thing that we may want to consider here is the lazy aggregation approach of lnd: https://github.com/lightningnetwork/lnd/blob/99d07fdb781ac4cd912e8054d105ff09a3cf15dd/htlcswitch/switch.go#L2535-L2562. The trade off is that for a period of time (the flush interval) the db records aren't consistent. 

A middle ground (a bit more complex) would be to maintain what's effectively an on disk WAL. Forwards come in, are written to this quickly as an append only file. A background goroutine then reads this file, and updates entries in the db, relying on the primary key to prevent duplicates.

rfq/order.go
```diff
@@ -901,8 +1028,9 @@ func (h *OrderHandler) mainEventLoop() {
 }
 
 // subscribeHtlcs subscribes the OrderHandler to HTLC events provided by the lnd
-// RPC interface. We use this subscription to track HTLC forwarding failures,
-// which we use to perform a live update of our policies.
+// RPC interface. We use this subscription to track HTLC forwarding failures
+// and settlements, which we use to perform live updates of our policies and
+// to log forward events for accounting.
```
Roasbeef, line 1037: We might want to consider sort of retroactively populating this information. IIRC, we actually persist the RFQ information now. So on start up, we can reconcile what lnd's forwarding history returns vs the records we have on disk.

rfq/order.go
```diff
@@ -956,6 +1081,90 @@ func (h *OrderHandler) subscribeHtlcs(ctx context.Context) error {
 	}
 }
 
+// handleHtlcSettle handles an HTLC settle event by updating the forward record
+// in the database with the settlement timestamp.
+func (h *OrderHandler) handleHtlcSettle(ctx context.Context,
+	circuitKey models.CircuitKey) {
+
```
Roasbeef, line 1103: As this is the _subscription_ endpoint, we risk losing data here if a crash occurs. Instead, I think we should move this logic to reside where we intercept HTLCs. There we can make sure we write _before_ actions occur, rather than after.

rfq/order.go
```diff
@@ -854,6 +936,51 @@ func (h *OrderHandler) handleIncomingHtlc(ctx context.Context,
...
+				HtlcID:     htlc.IncomingCircuitKey.HtlcID,
+				AssetAmt:   assetAmt,
+				AmtInMsat:  uint64(htlc.AmountInMsat),
+				AmtOutMsat: uint64(htlc.AmountOutMsat),
+			}
+
+			// Log the forward immediately when it opens.
+			id, err := h.forwardStore.LogForward(ctx, fwdInput)
+			if err != nil {
+				log.Errorf("Failed to log forward for "+
```
Roasbeef, line 975: Hmm, what about cases where the db fails for w/e reason on the initial write, but we're still up by the time the HTLC is settled? As is we won't find the record.

rfq/order.go
```diff
@@ -770,6 +843,14 @@ type OrderHandler struct {
 	// data available, so we need to cache this info.
 	htlcToPolicy lnutils.SyncMap[models.CircuitKey, Policy]
 
+	// htlcToForward maps an HTLC circuit key to the forward input data
+	// needed for logging. This is populated when an HTLC is accepted.
+	// Some fields (SettledAt, HtlcID, FeeMsat) are filled at settle time.
+	htlcToForward lnutils.SyncMap[models.CircuitKey, *ForwardInput]
```
Roasbeef, line 852: We should repopulate this on restart. Otherwise we'll miss data if a get the forward, restart, then settle later.

rfq/order.go
Roasbeef: I think we can resolve this if we do an `UPSERT` when we go to mark a forward as settle/fail in the db. So we'd try to insert the forward, then regardless of if that works or not, we update the record.

itest/rfq_forwards_test.go
```diff
@@ -0,0 +1,359 @@
...
+
+	// Carol generates an invoice for Alice to settle via Bob.
+	addInvoiceResp := ts.CarolLnd.RPC.AddInvoice(&lnrpc.Invoice{
+		ValueMsat: int64(askAmt),
+	})
+	invoice := ts.CarolLnd.RPC.LookupInvoice(addInvoiceResp.RHash)
+	payReq := ts.CarolLnd.RPC.DecodePayReq(invoice.PaymentRequest)
+
+	// Construct route: Alice -> Bob -> Carol.
+	routeBuildResp := ts.AliceLnd.RPC.BuildRoute(
```
Roasbeef, line 184: Why not just use normal invoices here? Would make the test a bit simpler as then you just make the invoice and all the RFQ related stuff is handled.

tapdb/sqlc/queries/rfq.sql
darioAnongba: Yes I've tried to follow Ollie's way of using `MaxValidSQLTime` and always populate the time instead of using NULL so it's the responsibility of the store to handle this properly. I like this approach since `opened_at` is not nullable.

tapdb/rfq_forwards.go
darioAnongba: Yes, I can also use `time.now` because not specifying the `openedBefore` means "I want every entry opened before now". I just use `MaxValidSQLTime` not to worry about concurrency issues between the RPC and the DB query.

tapdb/rfq_forwards.go
darioAnongba: agqrq9

rfq/order.go
darioAnongba: updated the comment so it's clearer.

rfq/order.go
darioAnongba: Yes, it's not exactly clear what behavior we expect here since we most of the time seem to reason about BTC <--> Asset. It uses the incoming purchase policy recursively.

rfq/order.go
darioAnongba: This seems to be related with your comment: https://github.com/lightninglabs/taproot-assets/pull/1921#discussion_r2666647489? If a forward is settled or failed, then it's fine and in a final state. if a forward is "pending", then on startup we can reconcile it with LND and set its state or populate the cache if still pending?
Let me know if that's what you have in mind.

rfq/order.go
darioAnongba: yes great point, updated to use upsert instead, cleaner as well.

rfq/order.go
darioAnongba: Issue created here to tackle later: https://github.com/lightninglabs/taproot-assets/issues/1934

rfq/order.go
darioAnongba: we do insert the forward into the DB + cache in the `handleIncomingHtlc` when accepted so we can't lose data for that reason. We can lose data because we can crash after receiving the event and before storing it, which will be fixed by reconciliation at startup.

rfq/order.go
darioAnongba: done

rfq/order.go
darioAnongba: implemented.

itest/rfq_forwards_test.go
darioAnongba: Sadly I tried to create an invoice (in BTC) such that the swap at the edge node happens automatically but I faced an error like "asset channel not found". I then tried to create an asset channel in the test but that is not currently possible. The solution seems to implement this type of tests in lightning-terminal?

For now this is mainly a copy of the RFQ itest.

Roasbeef:
I think you'll need to rebase this on top of main so it runs the new itests set up.

itest/rfq_forwards_test.go
Roasbeef: Ahh, yeah I think this type of test moreso belongs in the `litd` repo. Since there we can ensure that the proper aux hooks are set up.

tapdb/sqlc/queries/rfq.sql
```diff
@@ -25,22 +13,62 @@ RETURNING id;
...
+    settled_at = COALESCE(excluded.settled_at, forwards.settled_at),
+    failed_at = COALESCE(excluded.failed_at, forwards.failed_at),
+    rfq_id = excluded.rfq_id,
+    chan_id_out = excluded.chan_id_out,
+    asset_amt = excluded.asset_amt,
+    amt_in_msat = excluded.amt_in_msat,
+    amt_out_msat = excluded.amt_out_msat
+RETURNING id;
+
+-- name: QueryPendingForwards :many
```
GeorgeTsagk, line 40: should we paginate this too?

rfq/interface.go
```diff
@@ -18,3 +39,136 @@ type PolicyStore interface {
...
+	Offset int32
+}
+
+// ForwardingEvent is a complete forwarding event record including policy data.
+type ForwardingEvent struct {
+	// OpenedAt is the time when the forward was initiated.
+	OpenedAt time.Time
+
+	// SettledAt is the time when the forward settled (nil if not settled).
+	SettledAt *time.Time
```
GeorgeTsagk, line 113: why not use `fn.Option[time.Time]` here as well?

rfq/manager.go
```diff
@@ -1304,6 +1310,32 @@ func (m *Manager) publishSubscriberEvent(event fn.Event) {
 	)
 }
 
+// QueryForwards retrieves historical forwarding event records matching the
+// given
```
GeorgeTsagk: bad format

tapdb/sqlc/queries/rfq.sql
darioAnongba: no need here as the number of pending forwards cannot grow so big that we need pagination and this is used internally in the business layer only for the restoration at startup, it is not exposed externally.

rfq/interface.go
darioAnongba: The other structs use `fn.Option` because they are used for insertion into the DB and manipulated, while `ForwardingEvent` is the final returned entity so I preferred to keep it as idiomatic Go to avoid things like:
- `require.True(t, event.SettledAt.IsNone())` vs `require.Nil(t, event.SettledAt`), 
- `if event.SettledAt != nil` vs `fwd.SettledAt.WhenSome(func(t time.Time)`

Just style though, in Rust it would've been `Option<>` for sure.

darioAnongba:
> As mentioned, I think we may want to move this test (or just add an additional one) to the litd repo. That way it can use actual asset channels, and we can latch onto the normal oracle flow instead of mocking that out here.

> +1 let's also create a new (or enhance existing itests) to assert the rfq forwarding history

Yes I was planning on implementing this. On it.

darioAnongba:
> isn't the final count just the length of the retuned query within the specified set of constraints?

CountForwards is filtered without `limit` so the user knows if he has to continue querying to get all events, essentially pagination. I can do both calls in tandem in the same DB tx yes. Also no need to expose it to the manager for now. Fixing.

darioAnongba:
> The more important thing is that: this is only available when we need to go on chain for an HTLC: https://github.com/search?q=repo%3Alightningnetwork/lnd%20PutFinalHtlcOutcome&type=code. It won't be populated if we settle an HTLC off-chain.

@Roasbeef I think the HTLC is actually populated if settled off-chain as [the documentation](https://lightning.engineering/api-docs/api/lnd/lightning/lookup-htlc-resolution/) mentions an `off-chain` field as a `boolean`, if it doesn't, then it's a bug IMO (as not intended per the docs) that I can fix if needed. The main issue is the config by default you mention.

> we can ensure we don't miss any information if lnd is started while tapd isn't for some reason.

I don't see how we can miss information since we are storing the event when accepted as "pending" so we can reconcile them on startup?

@GeorgeTsagk I tried using `ForwardingHistory` instead of `LookupHtlcResolution` and it's pretty annoying as it doesn't register failed forwards and forces us to query events per time ranges. I have a feeling using `LookupHtlcResolution` is the cleanest implementation if we can move around the config issue, WDYT?

darioAnongba:
@Roasbeef, @GeorgeTsagk, ok so after some investigation I'm pretty confident on the path forward. Important to note that we already plug into the interceptor, which is where we run `handleIncomingHtlc` and store the forward in DB. We respond to LND with an `action` (which is always `RESUME_MODIFIED`). So unless I'm missing something, we correctly store all events. The problem is having forwards without final outcome ("stuck" in `pending`) in case of crash or missed events.

The interceptor stream doesn't include settle/fail events so Plan B is only viable if LND is extended to persist and replay final outcomes to the interceptor. These are the 2 ways to ensure crash-safe final status:

### A: Use `LookupHtlcResolution`

Enable `store-final-htlc-resolutions` by default in LND + make sure `LookupHtlcResolution` returns settled/failed off-chain forwarded HTLCs (and fix if not).

### B: Modify the interceptor in LND to require ACK of final outcomes

Intercept final outcomes and expect an ACK on the interceptor stream to delete them. Replay unacked outcomes on interceptor reconnect. That way we ensure final outcomes and we don't need `SubscribeHtlcEvents` anymore at all.

B is stronger than A but I'd say it's likely over-engineered. A depends on a correct config of LND though (+ storage of off-chain resolutions but that is expected IMO).

Let me know your thoughts.

GeorgeTsagk:
> A: Use LookupHtlcResolution

This feels like the path of least resistance. Also seems like the path were it seems like we have the most guarantees of getting the actual state/outcome of the HTLC in question. The only counter argument against this approach that I can think of is the efficiency one, as we're looking up HTLC states one by one.

> B: Modify the interceptor in LND to require ACK of final outcomes

Not sure what you mean by ACKed updates (likely you're not referring to the channel level HTLC acks). If you mean the interceptor client acking the reception of HTLCs then it's probs too much engineering.

For A I wouldn't consider the lnd config option an issue, it's just the default value.

Also worth noting that we'll be blocked by the next lnd release, which I also don't think changes things here.

darioAnongba:
> The only counter argument against this approach that I can think of is the efficiency one, as we're looking up HTLC states one by one.

True but we should generally have very few forwards pending as that is an edge case, we're not accumulating them.

> If you mean the interceptor client acking the reception of HTLCs then it's probs too much engineering

Yes this is what I meant. Reproduce the behavior we have now with `action`, except that here is a simple `action = ACK` such that LND knows the interceptor has seen the message and can remove it. Over-engineered indeed IMO.

If we go for `LookupHtlcResolution`, I think this PR is done. I'll ping you once I've tested it.

litbot-9000:
@roasbeef: review reminder
@georgetsagk: review reminder

docs/release-notes/release-notes-0.8.0.md
```diff
@@ -56,10 +56,27 @@
 
 ## Functional Enhancements
 
+- [Forwarding History Tracking](https://github.com/lightninglabs/taproot-assets/pull/1921):
```
GeorgeTsagk, line 59: Should drop a note about `--store-final-htlc-resolutions` somewhere in the release notes. Maybe the "Functional Enhancements" is the most appropriate place.

rfq/order.go
```diff
@@ -917,9 +1049,10 @@ func (h *OrderHandler) subscribeHtlcs(ctx context.Context) error {
 				continue
 			}
 
-			// Retrieve the two instances that may be relevant.
+			// Retrieve the event instances that may be relevant.
 			failEvent := event.GetForwardFailEvent()
 			linkFail := event.GetLinkFailEvent()
+			settleEvent := event.GetSettleEvent()
```
GeorgeTsagk, line 1055: Do we also want to grab the `FinalHtlcEvent`? This is useful for force closed channels and will give us updates on HTLCs that resolved on-chain.

I know that this will probably be caught on-restart by `LookupHtlcResolution`, but it will also provide a smoother experience for long-running (multiple months) tap daemons.

We shouldn't expect for people to restart when generating monthly accounting reports.

darioAnongba: Yes, I wasn't aware of this event thanks! Adding it.

darioAnongba:
LiT itest changes here: https://github.com/lightninglabs/lightning-terminal/pull/1222.
@GeorgeTsagk

.github/workflows/main.yaml
```diff
@@ -21,7 +21,7 @@ env:
 
   GO_VERSION: '1.25.2'
 
-  LITD_BRANCH: 'master'
+  LITD_BRANCH: 'test/tap-forwarding-history'
```
GeorgeTsagk: pending revert

darioAnongba: Reverted. Note that the LiT tests will still succeed, which is expected since pointing to master but the `ForwardingHistory` itests will not run until this PR is merged: https://github.com/lightninglabs/lightning-terminal/pull/1222. That's fine because it is blocked for now by other pending changes from the LiT team.

cc: @ViktorT-11

PR #1925, opened 2026-01-05
makefile: mount Git metadata for worktree support in Docker lint runs

ffranr:
Golangci-lint uses `issues.new-from-rev` to compare against Git history, but worktrees store .git metadata outside the worktree root. When lint ran in Docker, that history wasn't mounted, causing new-from-rev filtering to fail.

Detect the Git dir and common dir via `git rev-parse` and bind-mount them into the container when they are absolute and distinct. This preserves existing cache mounts and maintains consistent behavior across local and CI environments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,22 @@ ifneq ($(workers),)
 LINT_WORKERS = --concurrency=$(workers)
 endif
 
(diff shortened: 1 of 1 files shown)


jtobin (approved):
LGTM

PR #1928, opened 2026-01-05
taprpc+scripts: detect Podman wrapper and preserve Docker UID mapping

ffranr:
Detect Podman by checking "$DOCKER --version", allowing overrides via the DOCKER environment variable. Retain `--user $UID:$(id -g)` for Docker/CI to ensure generated files are user-owned. For Podman, switch to "--user=0:0" to avoid EACCES errors caused by rootless subuid/subgid remapping on bind mounts.

- scripts/gen_sqlc_docker.sh: add runtime user selection; use $DOCKER for run
- taprpc/gen_protos_docker.sh: same detection logic; use $DOCKER for build/run

Fixes "Permission denied" errors with rootless Podman, while maintaining existing Docker behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,10 @@ MAKE := make
 XARGS := xargs -L 1
 UNAME_S := $(shell uname -s)
 
+# Use docker by default; allow overrides and detect podman wrapper.
+DOCKER ?= docker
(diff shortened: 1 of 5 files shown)


jtobin (approved):
LGTM. I used to do my own local munging of this stuff for podman use, but then eventually grudgingly adopted Docker. Maybe I should retvrn to podman now?

PR #1931, opened 2026-01-06
build: bump version to v0.7.99-alpha

Roasbeef:
Bump version to v0.7.99-alpha so that master builds are distinguishable from release builds.


jtobin (approved):
This convention looks a little unusual to me, but I guess that's the point. 😁 

LGTM if this is already being used in lnd, in particular; a unified presentation for this type of thing is desirable.

PR #1935, opened 2026-01-08
rfq: delegate accept quote validation to PortfolioPilot

ffranr:
#### Summary

- Add `AcceptedRate` and `OriginalRequest` methods to the `Accept` interface for uniform access to accept message data
- Introduce `VerifyAcceptQuote` method in `InternalPortfolioPilot` to validate expiry and price deviations for buy and sell accept messages
- Streamline negotiator logic by delegating buy/sell accept validation to the PortfolioPilot


rfq/portfolio_pilot.go
```diff
@@ -175,3 +200,102 @@ func (p *InternalPortfolioPilot) ResolveRequest(ctx context.Context,
...
+	if !acceptablePrice {
+		return InvalidAssetRatesQuoteRespStatus, nil
+	}
+
+	return ValidAcceptQuoteRespStatus, nil
+}
+
+// expiryWithinBounds checks if a quote expiry unix timestamp (in seconds) is
+// within acceptable bounds. This check ensures that the expiry timestamp is far
+// enough in the future for the quote to be useful.
```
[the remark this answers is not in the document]
ffranr: done already?

jtobin (commented):
Lol, I accidentally entered my :fire: :fire: :fire: comment in the wrong browser tab. Meant to approve 1936.. 😁 

(Reviewing this one next!)

docs/release-notes/release-notes-0.8.0.md
```diff
@@ -30,6 +30,11 @@
 
 ## Functional Enhancements
 
+- [PR#1935](https://github.com/lightninglabs/taproot-assets/pull/1935)
```
jtobin: Something I thought about when looking at release notes lately is that we might be including too much information in them. This change, for example, isn't something I'd expect end users to *particularly* care about, as it's primarily an internal matter of organization.

It's more broadly a topic for another day, but just something I thought was worth commenting on.

litbot-9000:

docs/release-notes/release-notes-0.8.0.md
darioAnongba: +1

rfq/manager.go
```diff
@@ -1395,6 +1395,14 @@ const (
 	// PriceOracleQueryErrQuoteRespStatus indicates that an error occurred
 	// when querying the price oracle whilst evaluating the quote response.
 	PriceOracleQueryErrQuoteRespStatus QuoteRespStatus = 2
+
+	// PortfolioPilotErrQuoteRespStatus indicates that an unexpected error
+	// occurred in the portfolio pilot while evaluating the quote response.
+	PortfolioPilotErrQuoteRespStatus QuoteRespStatus = 3
```
darioAnongba, line 1401: These are supposed to be represented externally in `rfq.proto` right? Should be added to `QuoteRespStatus`.

rfq/portfolio_pilot.go
```diff
@@ -89,6 +106,14 @@ func (c *InternalPortfolioPilotConfig) Validate() error {
 		return fmt.Errorf("price oracle is nil")
 	}
 
+	if c.AcceptPriceDeviationPpm == 0 {
```
darioAnongba: Might be what you intended but `AcceptPriceDeviationPpm == 0` is valid right? Meaning that the prices must match exactly.

rfq/negotiator.go
```diff
@@ -484,342 +486,115 @@ func (n *Negotiator) HandleOutgoingSellOrder(
...
-				*invalidQuoteRespEvent,
-			),
-		)
-
-		return
-	}
-
 	if n.cfg.SkipAcceptQuotePriceCheck {
-		// Skip the price check.
+		// Skip the validation check.
```
darioAnongba: Seems that it's what you intended but the previous behavior is changed, now we do not validate the expiry is within bounds if `SkipAcceptQuotePriceCheck`, which makes this config naming wrong. 
We should either:
- change the name of the config to something like `SkipAcceptQuoteValidation`
- add a new config for price validation separate from expiry
- revert to all behavior (check expiry before)

rfq/portfolio_pilot.go
darioAnongba: This feels like a validation check that should live in `VerifyAcceptQuote` ? Not sure though.

ffranr: I've added a `OracleResponse.Validate` method in the latest commit set.

rfq/negotiator.go
ffranr: I've renamed the experimental field. It's a breaking change but it shouldn't be a problem.

rfq/portfolio_pilot.go
ffranr: I've removed this check.

darioAnongba (approved):
LGTM

Issue #1923, opened 2025-12-31
[bug]: How to Fix 'No Universe Proof Found' Error Causing Taproot Assets (tapd) to Crash?


shuishihang:
My node keeps encountering the following error periodically, causing tapd to crash:

`received critical error from sub-server (taproot-assets), shutting down: unable to receive proof for output 97b98f3c45f926057d430ef71f20a6d3e25d7a00fbd1d7b72b306a49d48c9d8c (script key 02703b6b3ccffc2cf84a83fce063c1ce8c8e06bd0e0e577ea8c0495f0886b332ca) in e9dbce3398cfcb41574aafe7ad100516ac29cec8e63b4ac374b3e4b596fc61ce:1: unable to receive proof using courier: error fetching proof provenance: fetching single proof failed: proof backoff receive attempt has failed: proof transfer backoff procedure failed; count retries attempted: 2000; error executing backoff procedure: backoff exec error: error retrieving proof from universe courier service: rpc error: code = Unknown desc = no universe proof found`

ffranr:
Thanks for the report. Fix: https://github.com/lightninglabs/taproot-assets/pull/1941

jtobin:
Resolved by #1941.

PR #1941, opened 2026-01-09
tapgarden: prevent shutdown on non-critical Custodian errors

ffranr:
Fixes tapd crash as reported here: https://github.com/lightninglabs/taproot-assets/issues/1923

Downgrade various runtime errors in the Custodian from critical to non-critical to prevent the entire tapd daemon from shutting down. Errors during proof lookups, proof reception, and wallet transaction inspection are now logged with error messages, allowing the main event loop to continue processing other events. This ensures that isolated failures for specific assets or transfers do not crash the application.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -41,6 +41,13 @@
   transfer proofs for remote-initiated force close transactions if they
   were not online to see them broadcast.
 
+- [PR#1941](https://github.com/lightninglabs/taproot-assets/pull/1941)
+  `tapgarden` now avoids a full `tapd` shutdown when the `Custodian` encounters
+  non-critical errors during its operation. Errors occurring during proof
+  availability checks, proof retrieval, or initial wallet transaction
+  inspections are now logged instead of being treated as fatal, allowing the
+  daemon to remain operational and continue processing other events.
+
 # New Features
 
 ## Functional Enhancements


tapgarden/custodian.go
```diff
@@ -495,7 +495,10 @@ func (c *Custodian) mainEventLoop() error {
 				),
 			)
 
-			return err
+			log.Errorf("Unable to check proof availability for "+
+				"event (outpoint=%v): %v", event.Outpoint, err)
```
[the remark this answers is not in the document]
ffranr: we are not going to use a random context `ctxStream` for this. Context would need to be passed into `mainEventLoop`.

tapgarden/custodian.go
```diff
@@ -526,9 +529,8 @@ func (c *Custodian) mainEventLoop() error {
 				),
 			)
 
-			// TODO(ffranr): Figure out if we can recover from this
```
GeorgeTsagk, line 529: I think we should keep the comment, as simply throwing out a log may not be sufficient.

Previously we were giving a strong signal to the user that custody of assets faced issues (by killing the daemon), but now this gets downgraded to just a log line. We should re-iterate over this in the future and attempt to recover these erroneous states.

ffranr: Good point. I'll add an issue to track this rather than just a TODO in code.

ffranr: Follow up issue: https://github.com/lightninglabs/taproot-assets/issues/1942

darioAnongba:
Even though I agree that shutting down is not ideal and the PR addresses that, I'm wondering why the reported issue appears in the first place. If missing proofs in universe is expected behaviour (is he switching universes? Is the universe purging some proofs?), then shouldn't we think about a clean up tool to clean the state such that these errors do not appear anymore? Instead of logging error forever?

@ffranr

ffranr:
> Even though I agree that shutting down is not ideal and the PR addresses that, I'm wondering why the reported issue appears in the first place. If missing proofs in universe is expected behaviour (is he switching universes? Is the universe purging some proofs?), then shouldn't we think about a clean up tool to clean the state such that these errors do not appear anymore? Instead of logging error forever?
> 
> @ffranr

We will make more follow up changes and investigation. We will need this PR also, IMO.

tapgarden/custodian.go
```diff
@@ -438,13 +438,37 @@ func (c *Custodian) mainEventLoop() error {
...
+		if resubscribeTimer != nil {
+			return
+		}
+
+		log.Warnf("Transaction subscription ended (%s), "+
+			"resubscribing in 10s", reason)
+
+		newTxChan = nil
+		txErrChan = nil
+		resubscribeTimer = time.After(10 * time.Second)
```
darioAnongba: should probably be a constant

tapgarden/custodian.go
```diff
@@ -438,13 +438,37 @@ func (c *Custodian) mainEventLoop() error {
...
+	// recover before we reconnect.
+	var resubscribeTimer <-chan time.Time
+
+	scheduleTxResubscribe := func(reason string) {
+		if resubscribeTimer != nil {
+			return
+		}
+
+		log.Warnf("Transaction subscription ended (%s), "+
+			"resubscribing in 10s", reason)
```
darioAnongba: `10s` written directly in string instead of constant.

ffranr:
Got this comment from Dario:

Yes I think it (read: this PR) will do for 0.7.1 because trying to reconnect when connection is lost is good for now and we can move to lndclient later if that makes sense.
The best practice with subscribe mechanisms in general, is usually to shut down if the connection is lost and use an external restart mechanism that verifies the health, the health check being fine grained by dependency like:
```
{
  "database": "Operational",
  "lightning_node": "Operational",
  "is_healthy": true
}
```
Then the livenessProbe checks the health. 
If we applied this to tapd, it would mean there is no point in retrying because the pod would be killed anyways because the health check will fail. 
That said, this is for professional setups, someone running tapd outside of Docker or Kube will most likely appreciate the retry mechanism.

PR #1940, opened 2026-01-09
tapchannel: no passive assets in funding proofs

jtobin:
Should resolve #1932.

Previously, anchorVPackets included all "TypeSimple" outputs in the funding proofs sent to a responder. Passive assets (viz. those located in the same UTXO, but not selected for funding) also have default output type TypeSimple, which caused them to be included incorrectly.

The fix is to also check that AnchorOutputIndex == 0, since passive assets are always assigned to anchor outputs at index >= 1.

Issue #1932, opened 2026-01-08
[bug]:  Incorrect Asset Channel Capacity Display


shuishihang:
I opened an asset channel and set the channel capacity (AssetAmount field) to 10000000000, but upon checking the channel information, I found it was incorrect. Here is the corresponding channel information.This problem has occurred repeatedly.
```
{
    "channels": [
        {
            "active": true,
            "remote_pubkey": "021fdf0ff70ade6f3dad7389adfd8675959cf97afc8a49c8e0cc6e6afd31d503ec",
            "channel_point": "17683a4f55eb9ae0fb550ad8fd62e5158037f691602456d1be232925974694ea:0",
            "chan_id": "1021872912818110464",
            "capacity": "20000",
            "local_balance": "2000",
            "remote_balance": "17080",
            "commit_fee": "260",
            "commit_weight": "958",
            "fee_per_kw": "269",
            "unsettled_balance": "0",
            "total_satoshis_sent": "108678",
            "total_satoshis_received": "108678",
            "num_updates": "1274",
            "pending_htlcs": [],
            "csv_delay": 144,
            "private": true,
            "initiator": false,
            "chan_status_flags": "ChanStatusDefault",
            "local_chan_reserve_sat": "1062",
            "remote_chan_reserve_sat": "354",
            "static_remote_key": false,
            "commitment_type": "SIMPLE_TAPROOT_OVERLAY",
            "lifetime": "823",
            "uptime": "100",
            "close_address": "",
            "push_amount_sat": "2000",
            "thaw_height": 0,
            "local_constraints": {
                "csv_delay": 144,
                "chan_reserve_sat": "1062",
                "dust_limit_sat": "354",
                "max_pending_amt_msat": "19800000",
                "min_htlc_msat": "1",
                "max_accepted_htlcs": 83
            },
            "remote_constraints": {
                "csv_delay": 144,
                "chan_reserve_sat": "354",
                "dust_limit_sat": "354",
                "max_pending_amt_msat": "19800000",
                "min_htlc_msat": "1",
                "max_accepted_htlcs": 83
            },
            "alias_scids": [
                "17592186044416000005"
            ],
            "zero_conf": false,
            "zero_conf_confirmed_scid": "0",
            "peer_alias": "LNT.chengdu2",
            "peer_scid_alias": "17592186044416000422",
            "memo": "",
            "custom_channel_data": {
                "funding_assets": [
                    {
                        "version": 1,
                        "asset_genesis": {
[...]

PR #1943, opened 2026-01-09
tapchannel: no passive assets in funding proofs

jtobin:
This is a copy of #1940, which was tombstoned by GitHub for some reason.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -48,6 +48,10 @@
   inspections are now logged instead of being treated as fatal, allowing the
   daemon to remain operational and continue processing other events.
 
+* [PR#1943](https://github.com/lightninglabs/taproot-assets/pull/1943)
+  addresses a bug in which passive assets would incorrectly be included in
+  channel funding proofs.
+
 # New Features
 
 ## Functional Enhancements


tapchannel/aux_funding_controller.go
```diff
@@ -1114,6 +1114,12 @@ func (f *FundingController) signAllVPackets(ctx context.Context,
 	return allPackets, activePackets, passivePkts, nil
 }
 
+// isFundingOutput returns true if the output should be included in channel
+// funding proofs.
+func isFundingOutput(vOut *tappsbt.VOutput) bool {
+	return vOut.Type == tappsbt.TypeSimple && vOut.AnchorOutputIndex == 0
```
ffranr: I think we should replace the 0 here with a funding output index const.

jtobin: Fixed in 06b1283095d6bb1134c6bbda0e73c297e7bba7e3.

tapchannel/aux_funding_controller.go
```diff
@@ -1178,13 +1189,10 @@ func (f *FundingController) anchorVPackets(fundedPkt *tapsend.FundedPsbt,
...
-			// channel funding output, so we'll store the proofs
-			// for those outputs. If there is change, that will be
-			// the split root output. And if there is no change,
-			// there is no split root output, as it's an interactive
-			// transfer.
-			if vPkt.Outputs[vOutIdx].Type == tappsbt.TypeSimple {
+			// Only include outputs destined for the funding output
+			// (index 0). Passive assets go to separate anchor
+			// outputs and should not be included in funding proofs.
+			if isFundingOutput(vPkt.Outputs[vOutIdx]) {
```
Roasbeef, line 1194: What was the next effect of this? Invalid proofs? 

In my mind this could only happen if we had an exactly sized input (enough to pay for fees, etc) and we did a full value fund (entire asset amount utilized). In other words, we should always have a change output where the passive assets would live.

jtobin: No invalid proofs; the effect observed in the wild was the one described in #1932, i.e. that displayed channel capacity was inflated by passive assets that weren't filtered out during funding. It's IMO entirely possible that other strange problems could manifest from this.

jtobin:
> If we have time, i think it would be good to add test coverage for this.

I've since made lightninglabs/lightning-terminal#1198, which adds an itest for this. I can confirm that it fails locally without the fix here, and passes when the fix is applied. :+1:

tapchannel/aux_funding_controller.go
```diff
@@ -1178,13 +1189,10 @@ func (f *FundingController) anchorVPackets(fundedPkt *tapsend.FundedPsbt,
...
 
-			// Any output that isn't a split root output is a
-			// channel funding output, so we'll store the proofs
-			// for those outputs. If there is change, that will be
-			// the split root output. And if there is no change,
-			// there is no split root output, as it's an interactive
-			// transfer.
-			if vPkt.Outputs[vOutIdx].Type == tappsbt.TypeSimple {
+			// Only include outputs destined for the funding output
+			// (index 0). Passive assets go to separate anchor
```
ffranr: IMO better not to hardcode the index 0 in the comment

PR #1945, opened 2026-01-12
tapgarden: flake reduction

### LINKED ISSUE #1938: [flake]: TestBatchedAssetIssuance flake
Our old nemesis, tapgarden (*tapgarrrrrdddddennnnnnnnnnn!*), has surfaced yet another flake:

```
--- FAIL: TestBatchedAssetIssuance (113.21s)
    --- FAIL: TestBatchedAssetIssuance/basic_asset_creation (66.74s)
        test_sqlite.go:11: Creating new SQLite DB handle for testing: /tmp/TestBatchedAssetIssuancebasic_asset_creation2778551181/001/tmp.db
        planter_test.go:1186: 
            	Error Trace:	/home/runner/work/taproot-assets/taproot-assets/tapgarden/planter_test.go:714
            	            				/home/runner/work/taproot-assets/taproot-assets/tapgarden/planter_test.go:1186
            	            				/home/runner/work/taproot-assets/taproot-assets/tapgarden/planter_test.go:2104
            	            	predicate not satisfied after time out
            	Test:       	TestBatchedAssetIssuance/basic_asset_creation
```

.. which appeared in [this](https://github.com/lightninglabs/taproot-assets/actions/runs/20851690486/job/59908069171) unit-race job.

jtobin:
Resolves #1938.

Two changes that should further reduce tapgarden flakes. TLDR:

* ensure that caretakers are stopped when the planter stops, and
* add an explicit order to AllMintingBatches query results.

They should be solid improvements even if they don't prove to squash the flakes described in the linked issue.

### DIFF
--- a/tapdb/sqlc/queries/assets.sql
+++ b/tapdb/sqlc/queries/assets.sql
@@ -96,10 +96,11 @@ SELECT *
 FROM internal_keys;
 
 -- name: AllMintingBatches :many
-SELECT * 
+SELECT *
 FROM asset_minting_batches
-JOIN internal_keys 
-ON asset_minting_batches.batch_id = internal_keys.key_id;
+JOIN internal_keys
+ON asset_minting_batches.batch_id = internal_keys.key_id
+ORDER BY batch_id ASC;
 
 -- name: InsertAssetSeedlingIntoBatch :exec
 WITH target_key_id AS (
--- a/tapdb/sqlc/assets.sql.go
+++ b/tapdb/sqlc/assets.sql.go
@@ -88,10 +88,11 @@ func (q *Queries) AllInternalKeys(ctx context.Context) ([]InternalKey, error) {
 }
 
 const AllMintingBatches = `-- name: AllMintingBatches :many
-SELECT batch_id, batch_state, minting_tx_psbt, change_output_index, genesis_id, height_hint, creation_time_unix, tapscript_sibling, assets_output_index, universe_commitments, key_id, raw_key, key_family, key_index 
+SELECT batch_id, batch_state, minting_tx_psbt, change_output_index, genesis_id, height_hint, creation_time_unix, tapscript_sibling, assets_output_index, universe_commitments, key_id, raw_key, key_family, key_index
 FROM asset_minting_batches
-JOIN internal_keys 
+JOIN internal_keys
 ON asset_minting_batches.batch_id = internal_keys.key_id
+ORDER BY batch_id ASC
 `
 
 type AllMintingBatchesRow struct {


tapgarden/planter.go
```diff
@@ -639,6 +639,10 @@ func (c *ChainPlanter) Stop() error {
 		close(c.Quit)
 		c.Wg.Wait()
 
+		// Stop all active caretakers after the gardener has exited.
+		// This ensures no concurrent access to the caretakers map.
+		c.stopCaretakers()
```
GeorgeTsagk: isn't this already called by this function's defer body?
https://github.com/lightninglabs/taproot-assets/blob/a4924b960f80407e3be75429703d43f21139b524/tapgarden/planter.go#L627

Here, after closing the context, we ensure that the routine also exited by waiting for the group to finish
https://github.com/lightninglabs/taproot-assets/blob/a4924b960f80407e3be75429703d43f21139b524/tapgarden/planter.go#L640

I won't argue against the fact that this may fix some flakes, but it seems quite hard to understand why

tapdb/sqlc/queries/assets.sql
```diff
@@ -96,10 +96,11 @@ SELECT *
...
 
 -- name: AllMintingBatches :many
-SELECT * 
+SELECT *
 FROM asset_minting_batches
-JOIN internal_keys 
-ON asset_minting_batches.batch_id = internal_keys.key_id;
+JOIN internal_keys
+ON asset_minting_batches.batch_id = internal_keys.key_id
+ORDER BY batch_id ASC;
```
GeorgeTsagk, line 103: In your commit message you write:
> which could only occur if a cancelled
batch occupied the final position

I don't get why sorting by id eliminates this posibility now. We can still have a batch which is the latest and also got cancelled. Maybe worth creating an explicit query if we really need to fetch non-cancelled batches?

litbot-9000:

tapgarden/planter.go
jtobin: Yeah, great catch. This was an oversight. There should now be an actual/better fix in e8955d00a76bdc1e93c39b80b60bc5cb726b2cd1.

tapdb/sqlc/queries/assets.sql
jtobin: Sorting by id works for the particular flaky test described in the job log because the cancelled batch was indeed not the expected last batch. But to your point, apparently we can just get the desired batch from `assertBatchProgressing`, which filters out cancelled batches and such anyway. This assertion was used previously, but its result was discarded (see 2fdff9d9a235c1d52344e770a91fd04d001a82fb).

Nonetheless, I think this change is worth keeping. We want 'fetchLastBatch' to actually fetch the last batch, which is not guaranteed to be the case without any enforcement on order.

Issue #1946, opened 2026-01-12
[itest]: avoid shared timeout contexts across long itests


ffranr:
Several integration tests create a single `context.WithTimeout` near the top of the test and reuse it for many RPC calls throughout the test body. This couples unrelated operations to one global deadline and creates slow-run flakes (especially on Postgres CI) rather than surfacing true logic errors.

## Why this is a bad pattern
- A single timeout effectively caps the *entire test body* to an arbitrary deadline (often 30s), even though the test itself legitimately does much more work.
- The timeout becomes increasingly likely to expire as the test grows or as environment speed varies, causing false failures unrelated to test logic.
- We already gate progress using `Eventually`, `Assert*` helpers, or explicit waits with their own timeouts, so a shared timeout is redundant and brittle.
- CI environments (especially postgres itests) can be slower than local runs, turning this into a CI-only flake.

## Example failure
```
addrs_v2_test.go:634: 
        Error Trace:    /home/runner/work/taproot-assets/taproot-assets/itest/addrs_v2_test.go:634
                        /home/runner/work/taproot-assets/taproot-assets/itest/test_harness.go:160
                        /home/runner/work/taproot-assets/taproot-assets/itest/integration_test.go:131
        Error:          Received unexpected error:
                        rpc error: code = DeadlineExceeded desc = context deadline exceeded
        Test:           TestTaprootAssetsDaemon/tranche03/99-of-101/address_v2_with_group_key_multiple_round_trips
harness_miner.go:149: Mined 0 blocks when cleanup the mempool
FAIL
```

https://github.com/lightninglabs/taproot-assets/actions/runs/20921048583/job/60109431291?pr=1941

## Signal that this is a slow-run issue (not a logic flake)
The `run itest` job passes. The failures show up in postgres itests, which are slower. That suggests the timeout is expiring because the test takes longer, not because the test logic is incorrect.

## Proposed direction
- Avoid reusing a single timeout context across long test bodies.
- Prefer one of:
  - `context.Background()` for RPCs that are already followed by `Eventually`/`Assert*` timeouts.
  - Create short-lived `context.WithTimeout` *per RPC* when a deadline is actually needed.

## Scope
Find and replace patterns where `ctxt := context.WithTimeout(...);` is created once and reused across multiple independent operations in itests (especially in long tests or loops).

ffranr:
There are more instances that need to be fixed...

PR #1947, opened 2026-01-12
itest: drop shared timeout contexts to reduce slow-CPU flakes

ffranr:
Closes https://github.com/lightninglabs/taproot-assets/issues/1946

- Replaces shared `context.WithTimeout(...defaultWaitTimeout)` usage across itest files with per-call `context.Background()` (or existing per-op timeouts) to avoid binding long test flows to a single 30s deadline.
- Updates helpers and tests in `addrs_test`, `addrs_v2_test`, `assertions`, `psbt_test`, `assets_test`, `asset_meta_test`, and the remaining itest suite (universe/rfq/multisig/etc.) for consistency.

## Motivation

- Shared timeout contexts cap the entire test body, so a small slowdown or extra logic can cause unrelated RPCs to fail with `DeadlineExceeded`.
- CI (especially postgres itests) is slower than local; this change avoids slow-run flakes while keeping Eventually/assert timeouts as the real bounds.

### DIFF
--- a/itest/addrs_test.go
+++ b/itest/addrs_test.go
@@ -1235,7 +1235,7 @@ func sendAsset(t *harnessTest, sender *tapdHarness,
 	opts ...sendOption) (*taprpc.SendAssetResponse,
 	*EventSubscription[*taprpc.SendEvent]) {
 
-	ctxb := context.Background()
+	ctx := context.Background()
 
 	// Create base request that will be modified by options.
 	options := &sendOptions{}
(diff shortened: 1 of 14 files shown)


jtobin (approved):
Gemini flagged one `-  defer cancel()` in itest/utils.go that I assume was collateral damage; otherwise LGTM. :+1:

darioAnongba (approved):
Painful but needed PR. Thanks!

PR #1949, opened 2026-01-13
ci: parallelize itest job tranches using all available CPUs

ffranr:
This PR updates the GitHub Actions workflow to run per-job itest tranches in parallel using all available CPUs to improve CI throughput and reduce wall-clock time.

### Changes

* Run itest tranches in parallel using the runner’s available CPU capacity.
* Consolidate coverage files to ensure coverage output remains correct with parallel execution.
* Add CPU logging (`nproc`, `lscpu`) to help debug slow-runner related flakes and performance regressions.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -23,6 +23,11 @@ env:
 
   LITD_BRANCH: 'master'
 
+  # Number of parallel itest jobs in the matrix and tranches per job.
+  # Total tranches = NUM_ITEST_JOBS * ITEST_PARALLELISM.
+  NUM_ITEST_JOBS: 4
+  ITEST_PARALLELISM: 4
+
 jobs:
   # Validate generated SQL models stay in sync with queries.
   sqlc-check:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -350,7 +386,7 @@ jobs:
         if: ${{ success() }}
         continue-on-error: true
         with:
-          file: itest/regtest/cover/coverage-tranche${{ matrix.tranche }}.txt
+          file: itest/regtest/cover/coverage-tranche*.txt
```
jtobin: Opus thinks that this glob pattern won't match and it [looks like](https://github.com/coverallsapp/github-action) it's probably right. Seems like you need to remove the 'file' field (to let coveralls auto-detect) or add individual items to the 'files' field (or something).

jtobin (approved):
LGTM. :+1:

.github/workflows/main.yaml
```diff
@@ -23,6 +23,11 @@ env:
 
   LITD_BRANCH: 'master'
 
+  # Number of parallel itest jobs in the matrix and tranches per job.
+  # Total tranches = NUM_ITEST_JOBS * ITEST_PARALLELISM.
+  NUM_ITEST_JOBS: 4
+  ITEST_PARALLELISM: 4
```
GeorgeTsagk, line 29: Is there some reasoning behind the number 4?
Otherwise it looks like a decent choice, just curious

ffranr: 4 cores are available per job currently

Issue #1070, opened 2024-08-08
[feature]: Add network-mismatch error messaging when CLI `network` mismatches config `network`


dstadulis:
## Motivation
Many users encounter errors when running CLI commands, which try to access universes with chain networks different than what the node is configured / running on.

User copies CLI command, from documentation, which specifies a `network` which mismatches the user's config file `network` setting.  An error is encountered. [Example instance](https://lightningcommunity.slack.com/archives/C03B3556HQ8/p1722842031925569?thread_ts=1722580798.372519&cid=C03B3556HQ8)

This enhancement will reduce user friction and support-request worker hours

## Deliverable
By adding an [ERR] message:
- [ ] calling out the mismatch 
- [ ] suggestes alternative CLI command (i.e. recommend the network the node is config'ed to run on)

PR #1950, opened 2026-01-14
Adding error for network mismatch between CLI and config

sergey3bv:
Addressing #1070

### DIFF
--- a/cmd/commands/profile.go
+++ b/cmd/commands/profile.go
@@ -194,7 +194,7 @@ func profileFromContext(ctx *cli.Context, store, skipMacaroons bool) (
 	macBytes, err := os.ReadFile(macPath)
 	if err != nil {
 		return nil, fmt.Errorf("unable to read macaroon path (check "+
-			"the network setting!): %w", err)
+			"the network setting or \"--network\" flag!): %w", err)
 	}
 	mac := &macaroon.Macaroon{}
 	if err = mac.UnmarshalBinary(macBytes); err != nil {


sergey3bv:
> Simpler, and perfectly acceptable IMO, would be to literally just add something like "(wrong --network flag?)" to the existing error.

You're absolutely right! I'll try to update the PR.

sergey3bv:
@jtobin, @GeorgeTsagk, guys, should we continue?

sergey3bv:
@Roasbeef could you review please

jtobin (commented):
I think better would probably be:

```
		return nil, fmt.Errorf("unable to read macaroon path (wrong "+
			"\"--network\" flag?): %w", err)
```

since the network setting *is controlled by* the `--network` flag. But strictly I believe tapcli also checks the `TAPCLI_NETWORK` environment variable, so the network setting can also be controlled via that.  I don't know, the error message is probably fine the way it is?

(I didn't realise the error message was already this clear when I suggested the alternative approach -- *mea culpa*.)

sergey3bv:
> I didn't realise the error message was already this clear

I am totally fine to just close the PR then.

sergey3bv:
Hey, @jtobin and @GeorgeTsagk, it has been a while, could you please let me know if the PR is worth closing?

jtobin:
Hey @sergey3bv -- probably worth closing this one, yeah. I think the issue is arguably already resolved by the existing message anyway.

PR #1954, opened 2026-01-22
rfq: query asset rates via `PortfolioPilot`

ffranr:
Adds a general-purpose asset rate query interface to `PortfolioPilot` and wires RFQ negotiation to use it, with small follow-up refactors to keep the order and rate-hint logic cohesive.

### Summary
- Introduce `QueryAssetRates` on the `PortfolioPilot` interface to support general asset rate querying and pricing information.
- Update the RFQ negotiator to fetch asset rates via `PortfolioPilot` rather than ad-hoc paths.
- Consolidate order primitives and rate-hint retrieval behind clearer abstractions.

### Motivation
This makes rate discovery a first-class capability in `PortfolioPilot`, enabling RFQ construction as well as broader pricing/rate-query use cases without duplicating oracle/exchange logic.

### DIFF
--- a/rfq/negotiator.go
+++ b/rfq/negotiator.go
@@ -11,6 +11,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightninglabs/taproot-assets/rfqmsg"
 	"github.com/lightningnetwork/lnd/lnutils"
+	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/routing/route"
 )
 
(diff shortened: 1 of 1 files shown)


ffranr:
I will put up a separate PR to revise context handling across the RFQ system.

EDIT: https://github.com/lightninglabs/taproot-assets/pull/1959

rfq/portfolio_pilot.go
```diff
@@ -5,13 +5,91 @@ import (
...
+	// from edge node peer, peer gets paid sats from the Lightning network).
+	AssetTransferBuy = 1
+
+	// AssetTransferSell indicates we want to sell the asset (send asset
+	// to edge node peer, peer forwards Lightning payment for us).
+	AssetTransferSell = 2
+)
+
+// String returns a human-readable representation of the AssetTransferDirection.
+func (d *AssetTransferDirection) String() string {
```
darioAnongba: `func (d AssetTransferDirection) String() string {`

value receiver for enums.

rfq/portfolio_pilot.go
```diff
@@ -5,13 +5,91 @@ import (
...
 )
 
+// AssetTransferDirection represents the direction of an asset transfer
+// from our perspective for pricing and rate queries.
+type AssetTransferDirection uint8
+
+const (
+	// AssetTransferUndefined is the zero value and indicates an unspecified
+	// direction.
+	AssetTransferUndefined = 0
```
darioAnongba, line 23: I think we already discussed this somewhere but `UNDEFINED` in an enum at the busines layer doesn't make sense because it is an impossible case. It's always either `Buy` or `Sell`. If unspecified is possible, then the type to use is `Option<enum>` or pointer to enum.

rfq/portfolio_pilot.go
```diff
@@ -279,3 +363,66 @@ func (p *InternalPortfolioPilot) expiryWithinBounds(expiry time.Time) bool {
...
+			return zero, fmt.Errorf("query sell price for sell: %w",
+				err)
+		}
+
+	default:
+		return zero, fmt.Errorf("unsupported transfer direction: %d",
+			query.Direction)
+	}
+
+	if err := resp.Validate(); err != nil {
```
darioAnongba, line 422: This is a strange behavior because you are getting the response from the `PriceOracle` and calling its own `Validate()` function, which suggests this can never fail, otherwise you should have gotten an error on the `QuerySellPrice` or `QueryBuyPrice`.

If the intent is internal validation of a external "raw" type, then it should be something like `validateOracleResponse(resp)`, otherwise you're implying that you can get an invalid `OracleResponse`?

rfq/negotiator.go
```diff
@@ -459,36 +364,49 @@ func (n *Negotiator) HandleOutgoingSellOrder(
...
+			oracleMetadata = fn.Some(order.PriceOracleMetadata)
+		}
+
+		query := AssetRateQuery{
+			AssetSpecifier: order.AssetSpecifier,
+			Direction:      AssetTransferSell,
+			PaymentAmount:  fn.Some(order.PaymentMaxAmt),
+			PeerID:         peerID,
+			OracleMetadata: oracleMetadata,
+			Expiry:         fn.Some(order.Expiry),
```
darioAnongba: We are now passing the `Expiry` here but it is never used I think. Some TODOs mentioned passing it to the queries before, now it is passed but has no effect?

rfq/portfolio_pilot.go
```diff
@@ -279,3 +363,66 @@ func (p *InternalPortfolioPilot) expiryWithinBounds(expiry time.Time) bool {
 	diff := expiry.Unix() - time.Now().Unix()
 	return diff >= int64(p.cfg.MinAssetRatesExpiryLifetime)
 }
+
+// QueryAssetRates returns current asset rate information by querying the
+// configured price oracle based on the transfer direction and query parameters.
+func (p *InternalPortfolioPilot) QueryAssetRates(ctx context.Context,
+	query AssetRateQuery) (rfqmsg.AssetRate, error) {
```
darioAnongba, line 370: `query.Expiry` unused, not sure if expected?

rfq/portfolio_pilot.go
```diff
@@ -279,3 +363,66 @@ func (p *InternalPortfolioPilot) expiryWithinBounds(expiry time.Time) bool {
...
+
```
darioAnongba, line 371: We were logging before:
`log.Debugf("Querying price oracle for buy price (asset_specifier=%s, ...")`
I think it was removed, not sure if expected.

rfq/portfolio_pilot.go
ffranr: It is a possible case in golang code because the default unset value of `type AssetTransferDirection uint8` is 0. The existence of `AssetTransferUndefined` means that we can capture code error where the value was not specified when it should have been.

Otherwise, if unset, it would default to `AssetTransferBuy` which is not what we want.

darioAnongba: Enums describe a list of possible cases, `unspecified` is not an existing case for the `Direction` type. Non existence != default case. We are mixing the zero value of a set with the `undefined` or `null` value.
This is true for all programming languages.

This is like having a `BitcoinNetwork` where instead of having `mainnet, testnet, signet, ...`, you have `unspecified, mainnet, testnet, signet, ...`. `Unspecified` is not part of the `BitcoinNetwork` enumeration (in any language).

This usage most likely comes from `protobuf` before they introduced the notion of `nil != zero` so you had to add an `Unspecified` field into the enum that leaked into the Go code.

rfq/portfolio_pilot.go
ffranr: It wasn't used before when passing into the price oracle. But it's included now since we need it in the portfolio pilot interface.

rfq/negotiator.go
ffranr: It wasn't used before when passing into the price oracle. But it's included now since we need it in the portfolio pilot interface.

rfq/portfolio_pilot.go
ffranr: I think you're right, I'll fix this in a follow up PR, since it requires test changes and so on.

rfq/portfolio_pilot.go
ffranr: We spoke about this offline. My motivation for `undefined` is to counter developer error. I will leave as is for now and continue to think about what you've said.

rfq/portfolio_pilot.go
ffranr: I've added these back now

PR #1958, opened 2026-01-27
tapchannel: align to latest LND version

darioAnongba:
Bumps `lnd` to latest version to align `aux_closer.go` with types that have moved from `lnwallet/chancloser` to `lnwallet/types`.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -21,7 +21,7 @@ env:
 
   GO_VERSION: '1.25.2'
 
-  LITD_BRANCH: 'master'
+  LITD_BRANCH: 'chore/lnd-latest'
 
   # Number of parallel itest jobs in the matrix and tranches per job.
   # Total tranches = NUM_ITEST_JOBS * ITEST_PARALLELISM.
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.24.11
 
 require (
 	github.com/btcsuite/btcd v0.24.3-0.20250318170759-4f4ea81776d6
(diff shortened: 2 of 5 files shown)


darioAnongba:
I bumped LND because I found and fixed a bug there that was preventing the `litd` itests to run well: https://github.com/lightninglabs/lnd/commit/05eed5cdb40bd38f1b2f078dfad043d39ad91ce6. 

`litd` bumps LND here: https://github.com/lightninglabs/lightning-terminal/pull/1219. But more changes are required to make it pass CI so this PR is effectively blocked until litd bumps LND, we are waiting on stable tagged releases.

As discussed, `tapd` should be allowed to bump LND without waiting on `litd`, but that is unrelated to the PR. We have effectively inverted the chain of dependencies.

.github/workflows/main.yaml
```diff
@@ -21,7 +21,7 @@ env:
 
   GO_VERSION: '1.25.2'
 
-  LITD_BRANCH: 'master'
```
ziggie1984, line 24: so when this PR is merged the LITD branch for the tests will not be master anymore ?

ziggie1984:
does this use the latest commit on master ?

darioAnongba:
> does this use the latest commit on master ?

I just updated it. It will diverge when new commits are merged though.

.github/workflows/main.yaml
darioAnongba: we can either target `master` and revert this change, having a proof that we have a PR in litd that is bumps LND and allows our CI to be green, or we keep pointing to that branch instead of master, both work for me.

cc: @jtobin

PR #1959, opened 2026-01-28
rfq/portfolio-pilot: propagate context through order + message handling APIs

ffranr:
Merge this PR first: https://github.com/lightninglabs/taproot-assets/pull/1954

---

Refactor RFQ/PortfolioPilot APIs to take `context.Context` for order upserts and quote/accept message handling, so callers can control cancellation and deadlines consistently.

### DIFF
--- a/rfq/manager.go
+++ b/rfq/manager.go
@@ -515,7 +515,7 @@ func (m *Manager) handleOutgoingMessage(ctx context.Context,
 		// we inform our peer of our decision, we inform the order
 		// handler that we are willing to sell the asset subject to a
 		// sale policy.
-		err := m.orderHandler.RegisterAssetSalePolicy(*msg)
+		err := m.orderHandler.RegisterAssetSalePolicy(ctx, *msg)
 		if err != nil {
 			return fmt.Errorf("registering asset sale "+
 				"policy: %w", err)
@@ -538,7 +538,7 @@ func (m *Manager) handleOutgoingMessage(ctx context.Context,
 		// we inform our peer of our decision, we inform the order
 		// handler that we are willing to buy the asset subject to a
 		// purchase policy.
-		err := m.orderHandler.RegisterAssetPurchasePolicy(*msg)
+		err := m.orderHandler.RegisterAssetPurchasePolicy(ctx, *msg)
 		if err != nil {
 			return fmt.Errorf("registering asset purchase "+
 				"policy: %w", err)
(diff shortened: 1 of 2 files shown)


jtobin (approved):
LGTM. :+1:

rfq/negotiator.go
```diff
@@ -430,7 +427,8 @@ func (n *Negotiator) HandleOutgoingSellOrder(
 // is called when a peer accepts a quote request from this node. The method
 // delegates validation to the portfolio pilot. Once validation is complete,
 // the finalise callback function is called.
-func (n *Negotiator) HandleIncomingBuyAccept(msg rfqmsg.BuyAccept,
+func (n *Negotiator) HandleIncomingBuyAccept(ctx context.Context,
+	msg rfqmsg.BuyAccept,
 	finalise func(rfqmsg.BuyAccept, fn.Option[InvalidQuoteRespEvent])) {
```
GeorgeTsagk, line 432: Question: With this change, `VerifyAcceptQuote` now uses the context passed from `mainEventLoop`, which originates from `Manager.Start()` via `WithCtxQuitNoTimeout()`. This means the previous per-call `DefaultPortfolioPilotTimeout` (20s) is removed.

For RPC-initiated paths (commits 1, 4, 5), this is fine since the RPC context has its own deadline. But for this event-loop path, a hanging portfolio pilot service would now block until daemon shutdown rather than timing out after 20s.

Is this intentional? If so, might be worth a comment noting the timeout responsibility has shifted to the caller.

GeorgeTsagk: Could also derive a child ctx here with the `DefaultPortfolioPilotTimeout`. This way we respect the parent ctx and also include a timeout as a safeguard for a hanging portfolio pilot.

rfq/negotiator.go
```diff
@@ -294,11 +286,6 @@ func (n *Negotiator) HandleIncomingQuoteRequest(
 	// The portfolio pilot might be an external service, responses could be
 	// delayed.
```
GeorgeTsagk, line 287: Same comment applies here regarding the timeout behavior change for `ResolveRequest`.

GeorgeTsagk (approved):
LGTM, pretty clean ✨

rfq/negotiator.go
ffranr: My thinking is that we can drop `DefaultPortfolioPilotTimeout`, since what we’re invoking here isn’t an external service but an in-process handler (and later, an RPC service handler). The handler, as a component within the tap codebase, is the right place to own timeout policy and apply it appropriately to whatever it calls into (for example, setting a timeout on an outbound RPC).

rfq/negotiator.go
ffranr: See https://github.com/lightninglabs/taproot-assets/pull/1959#discussion_r2746825634

Issue #1862, opened 2025-11-03
[feature]: Add `bakemacaroon` support to tapcli


ffranr:
Tapd generates default macaroons on first start and enforces per RPC `entity:action` permissions. Tapcli does not currently expose a `bakemacaroon` command like `lncli`, so users cannot mint scoped macaroons for tapd in standalone deployments.

Implement a macaroon bakery RPC in tapd and a `tapcli bakemacaroon` wrapper to allow users to mint macaroons with specific permissions. Match LND semantics but use Taproot Assets permission ops.

Copy patterns/naming conventions form `lncli bakemacaroon` where possible.

Also add an itest that bakes a restricted macaroon and verifies that:
- allowed RPCs succeed
- RPCs outside its scope fail.

ffranr:
Requested by user here: https://lightningcommunity.slack.com/archives/C03B3556HQ8/p1761924925539529

PR #1960, opened 2026-01-28
taprpc: add `BakeMacaroon` RPC and tapcli support

ffranr:
Closes: https://github.com/lightninglabs/taproot-assets/issues/1862

- Add `BakeMacaroon` RPC with updated proto/gateway artifacts.
- Add macaroon baking/validation into `rpcperms` and wire it through server config to keep the RPC handler thin.
- Add `tapcli bakemacaroon` and extend itest coverage for permission enforcement plus offline caveat presence via CLI

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -58,7 +58,13 @@
 
 ## RPC Additions
 
+- [PR#1960](https://github.com/lightninglabs/taproot-assets/pull/1960)
+  Add `BakeMacaroon` to mint custom macaroons with scoped permissions.
+
 ## tapcli Additions
+
+- [PR#1960](https://github.com/lightninglabs/taproot-assets/pull/1960)
+  Add `tapcli bakemacaroon` to bake custom macaroons with offline caveats.
 
 # Improvements
 


taprpc/taprootassets.proto
```diff
@@ -1966,3 +1973,30 @@ message RegisterTransferResponse {
...
+
+    // The action that is granted.
+    string action = 2;
+}
+
+message BakeMacaroonRequest {
+    // The list of permissions the new macaroon should grant.
+    repeated MacaroonPermission permissions = 1;
+
+    // The root key ID used to create the macaroon, must be a positive integer.
```
jtobin: s/positive/nonnegative

cmd/commands/macaroon.go
```diff
@@ -0,0 +1,377 @@
...
+		Usage: "the IP range the macaroon will be bound to",
+	}
+	macCustomCaveatNameFlag = cli.StringFlag{
+		Name:  "custom_caveat_name",
+		Usage: "the name of the custom caveat to add",
+	}
+	macCustomCaveatConditionFlag = cli.StringFlag{
+		Name: "custom_caveat_condition",
+		Usage: "the condition of the custom caveat to add, can be " +
+			"empty if custom caveat doesn't need a value",
```
jtobin, line 42: Codex flagged this one. It seems there's a discrepancy between:

```
--custom_caveat_condition     (flag, works)
--custom_caveat_condition=""  (option w/empty condition, fails)
```

I agree that it's potentially confusing. Could harmonize things here to have an empty string work in the same manner as well?

itest/macaroon_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/lightninglabs/taproot-assets/taprpc"
+	"github.com/lightningnetwork/lnd/macaroons"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/macaroon-bakery.v2/bakery/checkers"
+	"gopkg.in/macaroon.v2"
+)
+
+// testBakeMacaroonPermissions ensures that baked macaroons only allow the
+// specified permission set.
+func testBakeMacaroonPermissions(t *harnessTest) {
```
GeorgeTsagk, line 19: it seems like we don't ever try using or baking a mac with write permissions, let's add that too?

GeorgeTsagk: with a call that requires write perms as a follow up assert, ofcs

PR #1962, opened 2026-01-28
Add RPC portfolio pilot integration, config, docs, and itests

ffranr:
- Add PortfolioPilot RPC proto + generated stubs and client implementation.
- Wire tapd config/CLI/sample config to set experimental.rfq.portfoliopilotaddress.
- Add a basic portfolio pilot example server.
- Add itest harness + wiring + RFQ integration test for RPC portfolio pilot.

### DIFF
--- a/rfq/portfolio_pilot_rpc.go
+++ b/rfq/portfolio_pilot_rpc.go
@@ -116,6 +116,17 @@ func NewRpcPortfolioPilot(addrStr string, dialInsecure bool) (
 	}, nil
 }
 
+// Close closes the underlying gRPC connection.
+func (r *RpcPortfolioPilot) Close() error {
+	if r == nil || r.rawConn == nil {
+		return nil
+	}
+
+	err := r.rawConn.Close()
+	r.rawConn = nil
+	return err
+}
+
 // ResolveRequest resolves a quote request by calling the portfolio pilot RPC
 // server.
 func (r *RpcPortfolioPilot) ResolveRequest(ctx context.Context,
--- a/rfq/manager.go
+++ b/rfq/manager.go
@@ -391,6 +391,13 @@ func (m *Manager) stopSubsystems() error {
 			err)
 	}
 
+	if m.cfg.PortfolioPilot != nil {
+		if err := m.cfg.PortfolioPilot.Close(); err != nil {
+			return fmt.Errorf("error closing portfolio pilot: %w",
+				err)
+		}
+	}
+
 	return nil
 }
 
--- a/rfq/negotiator_test.go
+++ b/rfq/negotiator_test.go
@@ -72,6 +72,10 @@ func (s *mockPortfolioPilot) QueryAssetRates(context.Context,
 	}, nil
 }
 
+func (s *mockPortfolioPilot) Close() error {
+	return nil
+}
+
 // assertIncomingSellAcceptTestCase asserts the handling of an incoming sell
 // accept message for a test case.
 func assertIncomingSellAcceptTestCase(
--- a/rfq/portfolio_pilot.go
+++ b/rfq/portfolio_pilot.go
@@ -159,6 +159,9 @@ type PortfolioPilot interface {
 	// request construction, or general pricing information.
 	QueryAssetRates(context.Context, AssetRateQuery) (rfqmsg.AssetRate,
 		error)
+
+	// Close releases any resources held by the portfolio pilot.
+	Close() error
 }
 
 // InternalPortfolioPilotConfig holds settings for the built-in pilot that uses
@@ -433,4 +436,10 @@ func (p *InternalPortfolioPilot) QueryAssetRates(ctx context.Context,
 	}
 
 	return resp.AssetRate, nil
+}
+
+// Close releases resources held by the internal portfolio pilot. The internal
+// pilot is in-memory only, so this is a no-op.
+func (p *InternalPortfolioPilot) Close() error {
+	return nil
 }


Roasbeef:
@claude review this

rfq/portfolio_pilot_rpc.go
```diff
@@ -0,0 +1,584 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("load insecure dial options: %w",
+				err)
+		}
+	}
+
+	// Formulate the server address dial string.
+	serverAddr := fmt.Sprintf("%s:%s", addr.Hostname(), addr.Port())
+
+	conn, err := grpc.Dial(serverAddr, dialOpts...)
```
jtobin: Dial() [is apparently deprecated](https://pkg.go.dev/google.golang.org/grpc#Dial) and should be replaced with NewClient(). Arguably worth doing that now.  oracle.go still uses Dial() as well, so we should probably either change or ignore both.

jtobin (commented):
LGTM. The one thing that stood out to me was the TLS handling (it would, wouldn't it).

To avoid conflicting with #1775, we could self-contain the relevant TLS stuff here rather than grab helpers from oracle.go (since #1775 deletes them). The following patch should do it:

```
diff --git a/rfq/portfolio_pilot_rpc.go b/rfq/portfolio_pilot_rpc.go
index cf5c964e..550aeee1 100644
--- a/rfq/portfolio_pilot_rpc.go
+++ b/rfq/portfolio_pilot_rpc.go
@@ -2,6 +2,7 @@ package rfq
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"net/url"
 	"time"
@@ -13,6 +14,9 @@ import (
 	pilotrpc "github.com/lightninglabs/taproot-assets/taprpc/portfoliopilotrpc"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/keepalive"
 )
 
 // PortfolioPilotAddr is a type alias for a URL type that represents a
@@ -61,6 +65,27 @@ type RpcPortfolioPilot struct {
 	rawConn *grpc.ClientConn
 }
 
+// portfolioPilotDialOpts returns gRPC dial options for the portfolio pilot.
+func portfolioPilotDialOpts(insecureDial bool) []grpc.DialOption {
+	var creds credentials.TransportCredentials
+	if insecureDial {
+		creds = insecure.NewCredentials()
+	} else {
+		creds = credentials.NewTLS(&tls.Config{
+			InsecureSkipVerify: true,
+		})
+	}
+
+	return []grpc.DialOption{
+		grpc.WithTransportCredentials(creds),
+		grpc.WithKeepaliveParams(keepalive.ClientParameters{
+			Time:                30 * time.Second,
+			Timeout:             20 * time.Second,
+			PermitWithoutStream: true,
+		}),
+	}
+}
+
 // NewRpcPortfolioPilot creates a new RPC portfolio pilot handle given the
 // address of the portfolio pilot RPC server.
 func NewRpcPortfolioPilot(addrStr string, dialInsecure bool) (
@@ -72,21 +97,7 @@ func NewRpcPortfolioPilot(addrStr string, dialInsecure bool) (
 			err)
 	}
 
-	// Connect to the RPC server.
-	dialOpts, err := serverDialOpts()
-	if err != nil {
-		return nil, fmt.Errorf("load server dial options: %w", err)
-	}
-
-	// Allow connecting to a non-TLS (h2c, http over cleartext) gRPC server,
-	// should be used for testing only.
-	if dialInsecure {
-		dialOpts, err = insecureServerDialOpts()
-		if err != nil {
-			return nil, fmt.Errorf("load insecure dial options: %w",
-				err)
-		}
-	}
+	dialOpts := portfolioPilotDialOpts(dialInsecure)
 
```

We could then centralize TLS handling and modernize some of this stuff in a follow-up effort (Dial(), for example, is apparently deprecated). What do you think?

ffranr:
> LGTM. The one thing that stood out to me was the TLS handling (it would, wouldn't it).

@jtobin thanks for the patch, I've applied!

rfq/portfolio_pilot_rpc.go
```diff
@@ -0,0 +1,595 @@
...
+}
+
+// portfolioPilotDialOpts returns gRPC dial options for the portfolio pilot.
+func portfolioPilotDialOpts(insecureDial bool) []grpc.DialOption {
+	var creds credentials.TransportCredentials
+	if insecureDial {
+		creds = insecure.NewCredentials()
+	} else {
+		creds = credentials.NewTLS(&tls.Config{
+			InsecureSkipVerify: true,
```
darioAnongba, line 75: Are all these `InsecureSkipVerify: true` expected? Ideally, even if connecting to an internal service, we'd still want a secure connection, even though I see we had these prior to the PR in `oracle.go`?

rfq/portfolio_pilot_rpc.go
```diff
@@ -0,0 +1,595 @@
...
+	}
+
+	peer := req.MsgPeer()
+	rpcSpecifier := rpcMarshalPortfolioAssetSpecifier(req.AssetSpecifier)
+	return &pilotrpc.BuyRequest{
+		AssetSpecifier:      rpcSpecifier,
+		AssetMaxAmount:      req.AssetMaxAmt,
+		AssetRateHint:       rpcRateHint,
+		PriceOracleMetadata: req.PriceOracleMetadata,
+		PeerId:              peer[:],
```
darioAnongba, line 452: The oracle has a `ForwardPeerIDToOracle` config option, does it make sense here as well for privacy?

tapcfg/server.go
```diff
@@ -522,6 +523,22 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
...
 
+	// Determine whether we should use an external portfolio pilot.
+	switch rfqCfg.PortfolioPilotAddress {
+	case "":
+		// Leave the portfolio pilot as nil so the internal pilot is
+		// used.
+
+	default:
+		portfolioPilot, err = rfq.NewRpcPortfolioPilot(
+			rfqCfg.PortfolioPilotAddress, false,
```
darioAnongba, line 534: `dialInsecure` seems to be a config as well. Here it's hardcoded to `false`. Testing environments might want to pass `true` here if the `InsecureSkipVerify` is eventually removed.

rfq/portfolio_pilot_rpc.go
```diff
@@ -0,0 +1,595 @@
...
+		return nil, fmt.Errorf("parse portfolio pilot address: %w",
+			err)
+	}
+
+	dialOpts := portfolioPilotDialOpts(dialInsecure)
+
+	// Formulate the server address dial string.
+	serverAddr := fmt.Sprintf("%s:%s", addr.Hostname(), addr.Port())
+
+	conn, err := grpc.Dial(serverAddr, dialOpts...)
```
darioAnongba: `NewClient` is recommended, you can see how it's used in the SDK for example: https://github.com/lightninglabs/tap-sdk/blob/4dc5daccdcfb3e61f0800864d44227d1200549c3/grpc/client.go#L161

itest/rfq_test.go
```diff
@@ -743,6 +743,103 @@ func testRfqNegotiationGroupKey(t *harnessTest) {
...
+	BeforeTimeout(t.t, func() {
+		for {
+			resolveCalls, verifyCalls, queryCalls :=
+				pilot.callCounts()
+			if resolveCalls > 0 && verifyCalls > 0 &&
+				queryCalls > 0 {
+
+				return
+			}
+			time.Sleep(50 * time.Millisecond)
```
darioAnongba: we usually try to avoid sleeps in the itests, do you need this one?

rfq/portfolio_pilot_rpc.go
```diff
@@ -0,0 +1,595 @@
...
+	"github.com/lightningnetwork/lnd/routing/route"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/keepalive"
+)
+
+// PortfolioPilotAddr is a type alias for a URL type that represents a
+// portfolio pilot service address.
+type PortfolioPilotAddr = url.URL
```
darioAnongba, line 24: nit: Do we usually call these URLs "addresses"? That's not a common name, it's usually `endpoint` for `host:port`. Is this type even needed? Since it's just a URL.

taprpc/portfoliopilotrpc/portfolio_pilot.proto
```diff
@@ -0,0 +1,332 @@
...
+// VerifyAcceptQuoteRequest specifies an accepted quote to verify.
+message VerifyAcceptQuoteRequest {
+    // accept is the accepted quote and request details.
+    AcceptedQuote accept = 1;
+}
+
+// VerifyAcceptQuoteResponse is the response to a VerifyAcceptQuote call.
+message VerifyAcceptQuoteResponse {
+    // status is the verification result.
+    QuoteRespStatus status = 1;
```
darioAnongba, line 292: Do we need the error message somewhere or is it something we do not expose externally and only want to return the status?

rfq/portfolio_pilot_rpc.go
ffranr: Not in the way i think about `PortfolioPilot`. I think of it as a trusted service that the operator owns. So the service could decide on its side whether to forward peer ID to the oracle, or if there even is an oracle behind the remote `PortfolioPilot` service.

rfq/portfolio_pilot_rpc.go
ffranr: Jared's TLS PR touches on this sort of thing so I think we might revise there. Insecure is fine for now IMO since this is still experimental.

itest/rfq_test.go
ffranr: will fix, it should be a `require.Eventually` call

tapcfg/server.go
ffranr: we will revise TLS setup here before release, hopefully we can ship this PR without more TLS work.

rfq/portfolio_pilot_rpc.go
ffranr: I think we use "address" for price oracle also and maybe for proof couriers? So maybe sticking with that for now?

This is just a type alias, so doesn't hurt IMO. Could make code interfaces slightly clearer rather than just `url.URL`.

taprpc/portfoliopilotrpc/portfolio_pilot.proto
ffranr: I don't think portfolio pilot will be exposed externally. I think of it as a trusted service in local env.

PR #1963, opened 2026-01-29
refactor: isolate RPC server plumbing into rpcserver package

ffranr:
#### Motivation for `rpcserver` package

  - Keep daemon orchestration in the root package while consolidating all RPC server/client plumbing into a dedicated `rpcserver` package.
  - Reduce coupling by extracting server config types to `tapconfig` and moving shared key parsing into `rpcutils`.
  - This sets up cleaner seams for future RPC server refactors (splitting up `rpcserver.go`, etc) without changing behavior.

#### Summary of changes

  - Move `rpcserver.go` and universe RPC client helpers into `rpcserver/`.
  - Extract server config types into `tapconfig`.
  - Extract `ParseUserKey` into `rpcutils` for reuse.

### DIFF
--- a/tapconfig/config.go
+++ b/tapconfig/config.go
@@ -154,6 +154,9 @@ func ParseUniversePublicAccessStatus(
 type Config struct {
 	DebugLevel string
 
+	// Version is the tapd version string included in RPC responses.
+	Version string
+
 	// RuntimeID is a pseudo-random ID that is generated when the server
 	// starts. It is used to identify the server to itself, to avoid
 	// connecting to itself as a federation member.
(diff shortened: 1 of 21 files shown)


tapconfig/config.go
```diff
@@ -1,4 +1,4 @@
-package taprootassets
```
darioAnongba, line 1: This move seems to be the one that can create issues and makes LiT break because LiT is currently importing constants from this file. I'm totally fine with LiT breaking because this file should not be exposed externally as this is not a library. Any field imported from here should either be moved to an exposed package or copied in LiT. 

Thus, I'm approving even though LiT tests break because of this.

ffranr:
Litd fix PR: https://github.com/lightninglabs/lightning-terminal/pull/1229

Issue #1835, opened 2025-10-04
[feature]: Improve `ChainPorter` proof verification and storage logic


ffranr:
`ChainPorter` manages the outbound asset transfer process on the sender node when transferring a Taproot Asset on-chain. Its proof verification and storage logic can be improved in the following ways:

#### 1. Replace `proof.FileArchiver` with `proof.MultiArchiver` as `ChainPorterConfig.ProofWriter`.

This would allow proofs to be written both to the file system and to the database (SQLite or Postgres) via `tapdb.AssetStore`. Database storage enables operators to benefit from features like Postgres replication and simplifies proof persistence and recovery.

#### 2. Enforce proof verification within all proof archivers.

Each proof archiver should verify proofs prior to import when a `proof.VerifierCtx` argument is provided. We should add unit tests to confirm this behavior. Currently, for example, `proof.FileArchiver.ImportProofs` ignores the verifier argument and performs no verification. This means that passive asset proofs are accepted without validation, even though the code suggests otherwise (see [chain_porter.go#L522-L537](https://github.com/lightninglabs/taproot-assets/blob/614acdcb1c51b85dc926d44e21139d4f7d176adb/tapfreighter/chain_porter.go#L522-L537)).

#### 3. Perform partial output proof verification earlier in the `ChainPorter` state machine.

At present, output proofs are only verified during proof storage in `ChainPorter.storeProofs`, which runs in the `SendStateStorePostAnchorTxConf` state. This happens after the anchor transaction has already confirmed on-chain, introducing the risk of unrecoverable fund loss if verification fails.

To reduce this risk, `ChainPorter` should verify all input proofs in full and perform partial verification on output proofs before the anchor transaction is broadcast. Partial verification should include all checks that do not depend on block data, which can then be completed once block details are available. This could be implemented as a new state immediately before `SendStateStorePreBroadcast`.

For reference, input proofs are already verified during `SendStateVirtualSign`, which occurs before signing and allows the transfer to be safely aborted if verification fails.

ffranr:
We should break this issue into smaller issues after discussion.

jtobin:
Linking #1884 as a solid advancement on this front.

ffranr:
I don't think task 1 matters very much, but might be worth coming back to in the future as a low prio.

Task 2 is not complete but has partial progress. Created issue https://github.com/lightninglabs/taproot-assets/issues/1992 to takeover tracking of that task.

PR #1969, opened 2026-02-02
Pre‑broadcast proof verification: validate more, skip only what can’t be known yet

ffranr:
Addresses point 3 in https://github.com/lightninglabs/taproot-assets/issues/1835

---

This PR adds pre‑broadcast output‑proof verification while deliberately skipping the small subset of checks that can’t be accurately evaluated before the anchor transaction is confirmed. The result is stronger validation before funds are locked or a transaction is broadcast, without introducing false failures for locktime‑gated proofs.

## Motivation

We want to catch invalid proofs and malformed virtual packet outputs as early as possible—ideally before broadcast—while avoiding rejection of proofs whose correctness depends on data that only exists after confirmation (e.g., timelock maturity and anchor chain inclusion).

## Approach (why we skip, and why the rest is valuable)

- Skip only the final‑proof chain and timelock checks pre‑broadcast.
  The final proof depends on an unconfirmed anchor, so header/merkle validation and timelock evaluation can be incorrect before confirmation. Skipping these avoids false negatives.
- Still fully verify all prior proofs and all non‑timelock aspects of the final proof. This catches malformed proof suffixes, invalid witnesses, incorrect inclusion/exclusion proofs, and commitment inconsistencies before broadcast—reducing the chance of producing un‑deliverable proofs or burning invalid transfers.

### DIFF
--- a/proof/append_test.go
+++ b/proof/append_test.go
@@ -3,6 +3,7 @@ package proof
 import (
 	"bytes"
 	"context"
+	"errors"
 	"testing"
 
 	"github.com/btcsuite/btcd/blockchain"
@@ -15,6 +16,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/commitment"
 	"github.com/lightninglabs/taproot-assets/internal/test"
 	"github.com/lightninglabs/taproot-assets/tapscript"
+	"github.com/lightninglabs/taproot-assets/vm"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


itest/psbt_test.go
```diff
@@ -2200,40 +2200,18 @@ func testPsbtSighashNoneInvalid(t *harnessTest) {
...
-		ClientEventStream: stream,
-		Cancel:            streamCancel,
-	}
-
 	// Now we'll attempt to complete the transfer.
-	sendResp, err := bob.AnchorVirtualPsbts(
+	//
+	// The invalid SIGHASH_NONE witness should be rejected during
+	// pre-broadcast proof verification, so this RPC must fail here.
+	_, err = bob.AnchorVirtualPsbts(
```
GeorgeTsagk, line 2207: at first I thought we're changing behavior here, seems like the error changed & moved to a different call site?

GeorgeTsagk (approved):
Lgtm

itest/psbt_test.go
ffranr: yes, the error happens before we broadcast now, so before we wait for conf.

PR #1970, opened 2026-02-02
Speed up `unit` CI with optional verbosity and package parallelism

ffranr:
This adds an opt‑in verbose mode for unit tests and introduces package‑level parallelism (packages were previously handled sequentially). CI is wired to use 4-way parallel unit runs. The result is clearer unit test progress locally and faster unit jobs in CI.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -247,16 +247,19 @@ jobs:
 
   # Unit tests with race detection, coverage, and Postgres backend variants.
   unit-test:
-    name: run unit tests
+    name: run unit tests (${{ matrix.unit_type }})
     runs-on: ubuntu-latest
     strategy:
       # Allow other tests in the matrix to continue if one fails.
       fail-fast: false
       matrix:
-        unit_type:
-          - unit-race
-          - unit-cover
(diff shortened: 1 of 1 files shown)


darioAnongba (approved):
LGTM

.github/workflows/main.yaml
```diff
@@ -253,10 +253,13 @@ jobs:
...
       fail-fast: false
       matrix:
-        unit_type:
-          - unit-race
-          - unit-cover
-          - unit dbbackend=postgres
+        include:
+          - unit_type: unit-race
+            parallel: 1
+          - unit_type: unit-cover
```
jtobin, line 259: Looks like 'unit-cover' won't (and probably shouldn't?) use the 'parallel' argument here. 'parallel: 4' is benign, but is this intended?

jtobin (approved):
See my one comment about UNIT_COVER, but otherwise LGTM. :+1: :+1:

.github/workflows/main.yaml
ffranr: Thanks for pointing that out. In the latest commits, I've made this arg effective for `unit-cover`.

PR #1974, opened 2026-02-03
Make proof verification explicit via verified proof API

ffranr:
- Add a verified proof wrapper and verification helpers in `proof/`.
- Add a verified import entrypoint for `FileArchiver`.
- Switch ChainPorter to verify once and import via the verified API.

## Motivation
We want to separate verification from storage to improve code health and clarity. Archivers should store proofs, not verify them. By introducing a verified proof type, we make the verification boundary explicit in the API,
avoid implicit behavior, and enable “verify once, store everywhere” patterns without double verification.

## Implementation Notes
- `VerifiedAnnotatedProof` encodes “this proof has been verified” in the type system.
- `VerifyAnnotatedProofs*` centralizes verification and locator enrichment.
- `ImportVerifiedProofs` lets storage backends accept already verified proofs.

## Follow-ups
- Migrate other proof import call sites to the verified API as needed.

### DIFF
--- a/tapfreighter/chain_porter.go
+++ b/tapfreighter/chain_porter.go
@@ -38,17 +38,14 @@ const (
 	DefaultSendFragmentExpiryDelta = 12_960
 )
 
-// ProofImporter is used to import proofs into the local proof archive after we
-// complete a trransfer.
-type ProofImporter interface {
-	// ImportProofs attempts to store fully populated proofs on disk. The
-	// previous outpoint of the first state transition will be used as the
-	// Genesis point. The final resting place of the asset will be used as
-	// the script key itself. If replace is specified, we expect a proof to
-	// already be present, and we just update (replace) it with the new
-	// proof.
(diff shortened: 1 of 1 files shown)


proof/verified.go
```diff
@@ -0,0 +1,108 @@
...
+// VerifiedAnnotatedProof is an annotated proof that has been verified and
+// enriched with locator metadata. Only the proof package can construct
+// instances of this interface.
+type VerifiedAnnotatedProof interface {
+	// AnnotatedProof returns the underlying proof.
+	AnnotatedProof() *AnnotatedProof
+
+	// verified prevents external packages from implementing
+	// VerifiedAnnotatedProof.
+	verified()
```
jtobin, line 20: :+1:

jtobin (approved):
I think this is a better solution to go with than that in #1973. Make illegal states unrepresentable. :+1: :+1:

PR #1975, opened 2026-02-03
CI: shard unit-race across 4 jobs to cut runtime

ffranr:
#### Motivation

The unit-race job currently takes ~23 minutes. Splitting the workload across 4 CI jobs should reduce wall-clock time without changing coverage.

#### What’s in this PR

- Add scripts/unit_race_part.sh to run race tests for a deterministic package shard.
- Add unit-race-parallel Makefile target that mirrors existing unit-race flags/tags.
- Add a dedicated unit-race matrix job in .github/workflows/main.yaml (4 shards), and remove unit-race from the existing unit-test matrix.

#### Behavior / Coverage

- Each shard runs go test -race on a subset of packages from the same go list used by make unit-race.
- The union of all shards equals the full package list, so coverage is equivalent to the previous single-job run.

---

NOTE: I also threw in a commit to run `custom_channel` itests in parallel.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -557,7 +557,7 @@ jobs:
       matrix:
         include:
           - task_name: itests (custom_channels)
-            run_cmd: make itest icase=custom_channels
+            run_cmd: make itest-parallel icase=custom_channels tranches=4
           - task_name: unit
             run_cmd: make unit
 


ffranr:
This cuts wall-clock time for longest running CI jobs:
* `run unit race` from 24m to ~16m
* `run LiT itests (custom_channels)` from 30m to 19m

scripts/unit_race_part.sh
```diff
@@ -0,0 +1,60 @@
...
+DEV_TAGS=${DEV_TAGS:-dev monitoring}
+
+mapfile -t all_pkgs < <(go list -tags="${DEV_TAGS}" -deps "${PKG_PREFIX}/..." | \
+  grep -F "${PKG_PREFIX}" | grep -v "/vendor/")
+
+selected=()
+for i in "${!all_pkgs[@]}"; do
+  if (( (i % NUM_TRANCHES) == TRANCHE )); then
+    selected+=("${all_pkgs[$i]}")
+  fi
```
GeorgeTsagk, line 39: Randomness could lead to some jobs having way greater load than others.

But this is a super-nit anyway, there's no way to know up-front just from the test name how long it's going to run for. (lgtm -- just a dumping my thoughts)

.github/workflows/main.yaml
```diff
@@ -557,7 +557,7 @@ jobs:
       matrix:
         include:
           - task_name: itests (custom_channels)
-            run_cmd: make itest icase=custom_channels
+            run_cmd: make itest-parallel icase=custom_channels tranches=4
```
GeorgeTsagk, line 560: was this commit meant to be here?

PR title doesn't mention LiT itest changes

GeorgeTsagk (approved):
Looking good

.github/workflows/main.yaml
ffranr: Yes, I mentioned it at the bottom of the PR body. Just saw it as a quick win that i suck in.

scripts/unit_race_part.sh
ffranr: Loads are not evenly distributed. There is more to gain there for sure. Maybe we can get to that later.

Issue #857, opened 2024-03-26
Add enhanced trustless swap itest (anyonecanspend scriptkey)


GeorgeTsagk:
## Description

In https://github.com/lightninglabs/taproot-assets/pull/804 we added an itest that follows the trustless swap flow originally described in https://github.com/lightninglabs/taproot-assets/issues/577. That version of the trustless swap ran into some limitations and we had to add a design restriction -- alice can only swap a "whole coin", i.e there can not be any asset change going back to the creator of the swap.

This is because due to the non-interactive nature of the swap, the control of the asset root needs to belong to the swap receiver. By doing so, the swap receiver would be able to burn the creator's asset change. If the swap creator never pays any change back to themselves there's no risk, hence the swap flow makes sense for both parties.

## Enhanced Version

There is a different approach to this, which would not require the swap receiver to be in control of the asset root. The swap creator can pay the amount they wish to swap to an anyone-can-spend script key on the asset level, and let the swap receiver anchor it to an outpoint that they control. This way, the swap receiver will effectively be in full control of the swapped amount.

On the bitcoin level, the swap creator would again use `SIGHASH_SINGLE | ANYONECANPAY` to allow for the swap receiver to plug-in their own btc (to satisfy the creator) and also create extra outputs for their own anchor and btc change.

PR #1977, opened 2026-02-05
itest: Enhanced trustless swap

### LINKED ISSUE #813: [feature]: non-interactive BTC<->asset swaps with asset-level ANYONECANSPEND
Related to #577.

Currently we're enabling asset-level sighashes, with one application being BTC-to-asset swaps via PSBT.

A seller can construct a 1-in 1-out PSBT, where the first output anchors any assets not being sold + determines the price to fulfill the swap. The TX is invalid until a buyer attaches an input of sufficient value. The second output, constructed by the buyer, would anchor the assets being sold.

An issue here is how to maintain non-interactivity, while also constructing the asset purchased by the sender correctly. The script key they want to use cannot be known in advance, and there are issues around splitting the input asset to complete the sale.

One option is to have the purchased asset spendable by **anyone** - in practice only one TX built from this PSBT will confirm, so there will only be one owner of the sold asset.

There are many ways to create this ANYONECANSPEND behavior using the asset script key, via [OP_TRUE](https://github.com/bitcoinbook/bitcoinbook/blob/develop/ch06_transactions.adoc#output-scripts) or a [constant](https://bitcoin.stackexchange.com/questions/72773/how-to-create-a-p2sh-transaction-with-a-scriptsig-of-op-true). We could extend the script further to make these ANYONECANSPEND keys globally unique, similar to burn keys.

This would allow a seller to perform an asset split in the same TX as the sale. They would publish the asset leaf being sold, with the correct ANYONECANSPEND key, and all the relevant proof data. Specifically the split commitment, proof of non-inclusion in the first output, and the proof of inclusion for the input. That should be enough for any buyer to construct the second output that anchors the sold asset with their own internal key.

One issue with the seller anchoring assets in the output is that they need to generate a proof of non-inclusion for any output a buyer produces, which would require knowing the internal key used by the buyer. Also, since the value of the other anchored assets is unknown, the seller is at risk if the auction is griefed and the TX is constructed s.t. all assets are burned.

GeorgeTsagk:
## Description

A while ago we introduced vPSBT level sighashes which allowed us to build a basic trustless swap construct:
https://github.com/lightninglabs/taproot-assets/pull/804

One limitation there is that the seller needs to create a clean input for the assets, which doesn't carry any passive assets back to them (no change, needs to spend all).

This PR introduces an itest where we use an ANYONE_CAN_SPEND (OP_TRUE) script key for the assets, which is pre-defined by Alice (the seller) when preparing the swap offer.

Alice then creates two outputs on the asset level:
- Output@0: Asset change, btc payment
- Output@1: Assets to swap

As far as the input goes, she can bring in any "dirty" input that may carry more than what is involved in the swap:
- Input@0: Total asset balance, Btc balance

What matters for the swap to be complete is for Bob to complete the BTC level transaction validity by bringing an extra btc input, which makes Alice whole at Output@0.

```mermaid
sequenceDiagram
      box Alice's Offer
      participant IN as Input<br/>1000 units + 330 sats
      participant O0 as Output 0<br/>400u + 69,420 sats
      participant O1 as Output 1<br/>600u + 330 sats
      end

      IN->>O0: Alice's change (protected)

      Note over IN,O0: SIGHASH_SINGLE|ANYONE_CAN_PAY
      Note over O1: OP_TRUE<br/>Finalized by Bob
```

## Problems

- [ ] On the vPSBT level the split root is handled by Alice, who places it at output@0 where her asset change is placed. This split root also commits to the OP_TRUE split which may be claimed by Bob. One good thing here is that even if the split is destroyed (Bob griefs / burns assets) the split root is not considered invalid and only Bob's assets are affected. The catch is that if Bob decides to burn the assets, and also create a random P2TR output in the same tx, then Alice is going to face an issue with the exclusion proofs for her assets.
- [ ] Even if this is considered an "enhancement" over the previous trustless swap setup, there are some trade-offs:
  Alice locks additional asset balance in Input@0, which she might want to use before the swap is completed by someone. This approach reduces on-chain fees (no need to prepare an explicit input for the swap), but also limits the usability of the passive funds in that input. E.g if Alice wanted to then create another swap offer with some of the passive assets in that first offer, she'd be locked while waiting for someone to redeem the first one. From this perspective, the previous approach with "clean" asset inputs offers way greater flexibility for the swap offer creator.

Closes https://github.com/lightninglabs/taproot-assets/issues/813, https://github.com/lightninglabs/taproot-assets/issues/857

### DIFF
--- a/itest/psbt_test.go
+++ b/itest/psbt_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/taprpc"
 	wrpc "github.com/lightninglabs/taproot-assets/taprpc/assetwalletrpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	"github.com/lightninglabs/taproot-assets/tapscript"
 	"github.com/lightninglabs/taproot-assets/tapsend"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/keychain"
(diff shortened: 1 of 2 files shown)


GeorgeTsagk:
Closing, as it's blocked on the proof level construction by redeemer.

Could consider adding a bit of interactivity (i.e let alice handle the split root finalization over mailbox based communication or similar, but blast radius now gets significantly increased).

PR #1979, opened 2026-02-06
go.mod: update lnd to 14b90a27cfc3

ziggie1984:
## Summary
- Update `lnd` and `lnd/sqldb` dependencies to commit `14b90a27cfc3`
- Adapt to `AuxCloseDesc`, `CloseOutput`, and `AuxShutdownReq` types moving from `chancloser` to `lnwallet/types`

## Test plan
- [ ] Verify `make build` succeeds
- [ ] Run `make unit` to confirm no regressions


jtobin:
Note that we also have #1958 in-flight here -- I've argued that we should avoid merging that until litd is "about ready" to bump lnd as well.

Issue #426, opened 2023-08-02
[feature]: backup and recovery


guggero:
# Background

As a prerequisite for being able to officially support Bitcoin `mainnet` for taproot assets, we need to carefully think about how we approach the question of backup and recovery of `tapd` data, since not only assets might be at stake but also the BTC of the anchoring transaction output (you can't spend the BTC that carries assets without being able to reconstruct the full asset tree).

This issue serves as a collection/brainstorm issue around everything related to data safety, backups and recovery procedures.

## Documentation

Similar to the [`lnd` Operational Safety Guidelines](https://github.com/lightningnetwork/lnd/blob/master/docs/safety.md) document, we'll want a doc that describes the different data sources, what they are used for and how to best prevent loss thereof.
The document should (at least) describe the following key items:
 - What is the relationship between asset public/private keys (e.g. `script_key`s) and `lnd`'s wallet/seed?
 - What data is required in order to recover both the assets and the BTC of a taproot asset output?
 - What data is stored where (`tapd`'s database, `lnd`'s wallet database, `lnd`'s channel database)?
 - Where does `tapd` store its files and which files need to be backed up regularly?
 - How can the `tapd` database be set up in a production ready manner?
 - Recoverability when using a public universe vs. using a private one? (See further below).

## How to prevent database loss

As long as the `tapd` database is fully intact and the seed for the `lnd` wallet is known, all funds are SAFU.
So to have a replicated (or at least regularly backed up) state of the DB should be the highest priority.
We should test and then document the following ways of setting up a database cluster or streaming replication:
 - Using a [Postgres database cluster](https://www.postgresql.org/docs/current/creating-cluster.html) as a database backend: This is already possible and is the recommended way of running `tapd` in a production environment. We'll want to document some setup recommendations and best practices around this though.
 - Add support for low-level SQLite replication support, perhaps using something like https://github.com/benbjohnson/litestream.

## How to recover from full database loss

Even though keeping the `tapd` database intact should always be the highest priority, the reality is that users often don't realize that uninstalling and re-installing an app on platforms like Umbrel causes all data to be deleted. So because we want to ship `tapd` as part of Lightning Terminal, which is available on such platforms, we need to have a strategy for basic recovery of assets and BTC for the case when the full `tapd` database is lost.

Possible approaches:
[...]

PR #1980, opened 2026-02-06
Asset Wallet Backup Support

GeorgeTsagk:
## Description

Adds the ability to export the state of the asset wallet in an encoded byte array, that can be stored offline for asset recovery in case of a database failure. We introduce ExportAssetWalletBackup and ImportAssetsFromBackup RPCs to the AssetWallet service for backing up and restoring Taproot Assets wallet data.

Currently we follow a naive self-contained approach of dumping the whole proof file per asset in the backup format.

A few optimizations to consider here:
- [x] Strip the proof files of any information that can be reliably re-populated by the chain backend
- [x] Handle stale backup imports, where some assets may be spent
- [x] Support a more compact format where proofs are only referenced by ID, and we rely on universe availability for full proof reconstruction. (backup would contain key information for the most part).
- [ ] Currently we scan the whole asset wallet every time the export RPC is called, instead we can follow a better approach where the state of the backup is always held in memory and updated on the fly as the daemon performs actions. For example if I have a backup containing 100 different assets, and I receive a new asset, I should just append an extra entry to this structure instead of reconstructing it from scratch (checksum would be recalculated ofcs but cost is near zero).

More info here (does not close issue): https://github.com/lightninglabs/taproot-assets/issues/426

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -84,7 +84,28 @@
   **Note:** For full functionality, it is highly recommended to start LND with
   the `--store-final-htlc-resolutions` flag enabled, which is disabled by default.
 
+## Functional Enhancements
+
+- [Wallet Backup/Restore](https://github.com/lightninglabs/taproot-assets/pull/1980):
+  Add wallet backup and restore functionality with three modes: **raw** (v1)
+  exports complete proof files, **compact** (v2) strips blockchain-derivable
+  fields from proofs for significantly smaller backups (fields are reconstructed
+  from the blockchain on import), and **optimistic** (v3) omits proofs entirely
+  and fetches them from a universe federation server on import, producing the
+  smallest possible backup. On import, stale assets whose anchor outpoints have
+  already been spent are automatically detected and skipped. Note that backup
+  files contain asset key derivation paths (similar to exporting a descriptor)
+  and should be stored securely. A full restore still requires access to the
+  corresponding LND wallet; the backup only covers the Taproot Assets layer.
+  See [docs/backup.md](../backup.md) for the full format specification.
+
 ## RPC Additions
+
+- [Wallet Backup RPCs](https://github.com/lightninglabs/taproot-assets/pull/1980):
+  Add `ExportAssetWalletBackup` and `ImportAssetsFromBackup` RPCs to the
+  `assetwalletrpc` service. Export accepts a `mode` field (`RAW`, `COMPACT`,
+  `OPTIMISTIC`) for selecting the backup format. Import returns the count of
+  newly imported assets.
 
 - [PR#1960](https://github.com/lightninglabs/taproot-assets/pull/1960)
   Add `BakeMacaroon` to mint custom macaroons with scoped permissions.
@@ -94,6 +115,12 @@
   and asset rate queries.
 
 ## tapcli Additions
+
+- [Wallet Backup CLI](https://github.com/lightninglabs/taproot-assets/pull/1980):
+  Add `tapcli assets backup export` and `tapcli assets backup import` commands
+  exposing the wallet backup RPCs. Export supports `--mode` (`raw`, `compact`,
+  `optimistic`) and `--output_file` flags. Import reads a backup blob from
+  `--backup_file` and restores assets.
 
 - [PR#1960](https://github.com/lightninglabs/taproot-assets/pull/1960)
   Add `tapcli bakemacaroon` to bake custom macaroons with offline caveats.
@@ -238,6 +265,13 @@
 
 ## Testing
 
+- [Wallet Backup Integration Tests](https://github.com/lightninglabs/taproot-assets/pull/1980):
+  Add two integration tests for wallet backup covering genesis backup/restore,
+  idempotent import, all three backup modes with size comparison, post-restore
+  spendability via multi-hop transfer, and stale backup detection. Also includes
+  unit tests for TLV encoding roundtrips, checksum verification, and proof
+  strip/rehydrate cycles.
+
 - [PR#1962](https://github.com/lightninglabs/taproot-assets/pull/1962)
   Add an integration test that exercises the PortfolioPilot RPC flow.
 
@@ -258,6 +292,10 @@
 ## Code Health
 
 ## Tooling and Documentation
+
+- [Wallet Backup Format Spec](https://github.com/lightninglabs/taproot-assets/pull/1980):
+  Add `docs/backup.md` documenting the binary backup format, TLV schema,
+  compact strip/rehydrate mechanism, stale detection flow, and RPC interface.
 
 - [PR#1962](https://github.com/lightninglabs/taproot-assets/pull/1962)
   Add a basic PortfolioPilot RPC example under `docs/examples`.


GeorgeTsagk:
Oh, it's missing the corresponding CLI commands, will add soon

jtobin (approved):
LGTM. 🚢 🚢 

I alley-ooped #1998, which adds a test to mechanically ensure that strippedEncodeRecords is in sync. Could be worth pulling in.

litbot-9000:

backup/strip.go
```diff
@@ -0,0 +1,214 @@
...
+// strippedEncodeRecords returns the TLV record set for encoding a proof
+// without blockchain-derivable fields. This mirrors Proof.EncodeRecords()
+// but skips BlockHeaderRecord, AnchorTxRecord, TxMerkleProofRecord, and
+// BlockHeightRecord.
+//
+// WARNING: This function must be kept in sync with proof.Proof.EncodeRecords().
+// If new even TLV types are added to the proof encoding, they must be
+// explicitly included here or they will be silently dropped from stripped
+// backups.
+func strippedEncodeRecords(p *proof.Proof) []tlv.Record {
```
darioAnongba, line 135: Do we have a way to keep this in sync with `Proof.EncodeRecords()`? such that they don't drift without us noticing? like a test or itest?

Even better, just add a filter to EncodeRecords, and use that?

backup/encoding.go
```diff
@@ -0,0 +1,764 @@
+package backup
```
darioAnongba, line 8: As discussed, we do not wish to encrypt the backups as that is the responsibility of whoever is producing these backups. Worth mentioning the leak in privacy as we have the derivation paths exposed (similar to exposing a descriptor or an xpub).

backup/import.go
```diff
@@ -0,0 +1,349 @@
...
+		err   error
+	}
+
+	results := make(chan spendResult, len(assets))
+
+	// Register spend notifications and spawn a goroutine per
+	// outpoint. Each goroutine gets its own timeout so that even
+	// with thousands of assets, every outpoint has the full
+	// duration for lnd to respond.
+	for i, ab := range assets {
```
darioAnongba, line 309: minor: If len(assets) is very big, which it shouldn't in most cases, you can spawn a very high amount of gouroutines. Is there a way to bound this? Or not important?

taprpc/perms.go
```diff
@@ -159,6 +159,14 @@ var (
 			Entity: "assets",
 			Action: "write",
 		}},
+		"/assetwalletrpc.AssetWallet/ExportAssetWalletBackup": {{
+			Entity: "assets",
+			Action: "read",
```
darioAnongba: WDYT of using "write" permission to export backups? Looks more standard to me. The best would be a new "backup" dedicated permission.

backup/import.go
```diff
@@ -0,0 +1,349 @@
...
+			anchorKey := keychain.KeyDescriptor{
+				PubKey:     info.PubKey,
+				KeyLocator: info.KeyLocator,
+			}
+
+			err = cfg.KeyRegistrar.InsertInternalKey(
+				ctx, anchorKey,
+			)
+			if err != nil {
+				log.Warnf("Failed to insert anchor "+
```
darioAnongba: why warning here? Isn't it better to fail? should be fatal (except duplicate-key errors) or the asset should be skipped

backup/import.go
```diff
@@ -0,0 +1,349 @@
...
+			if len(skInfo.Tweak) > 0 {
+				scriptKeyType =
+					asset.ScriptKeyScriptPathExternal
+			}
+
+			err = cfg.KeyRegistrar.InsertScriptKey(
+				ctx, scriptKey, scriptKeyType,
+			)
+			if err != nil {
+				log.Warnf("Failed to insert script key "+
```
darioAnongba: why warning here? Isn't it better to fail? should be fatal (except duplicate-key errors) or the asset should be skipped

backup/import.go
```diff
@@ -0,0 +1,349 @@
...
+					"for asset %d: %v", i, err)
+			}
+		}
+
+		// Import the proof blob directly into the archive.
+		// The backup stores the complete original proof file, so
+		// we can import it without any reconstruction.
+		err = cfg.ProofArchive.ImportProofs(
```
darioAnongba, line 255: Seems we call ImportProofs without cross-checking the proof's assetId matches the backup's `ScriptKeyInfo`. A tampered backup can register keys for asset A but import proof for asset B.

1. Register script key A via backup metadata
2. Import proof for asset B
3. The wallet now has key A registered but asset B imported

backup/import.go
```diff
@@ -0,0 +1,349 @@
...
+			ctx, cfg.ProofVerifier, false,
```
darioAnongba, line 256: It's a bit painful that the verification is done after the keys are registered.... Can we somehow verify first and register the keys after? Otherwise if import fails, the keys stay persisted in DB forever.

taprpc/assetwalletrpc/assetwallet.proto
```diff
@@ -546,3 +562,41 @@ message DeclareScriptKeyResponse {
     // tweaked script key.
     taprpc.ScriptKey script_key = 1;
 }
+
+// BackupMode specifies the backup format to use when exporting.
+enum BackupMode {
+    // RAW exports a full backup with complete proof data (v1).
+    RAW = 0;
```
darioAnongba, line 569: here default is RAW, yet on the CLI default is `compact`?

itest/backup_test.go
```diff
@@ -0,0 +1,468 @@
...
+	t.Logf("Dave (COMPACT): %d assets imported",
+		daveImport.NumImported)
+
+	// === Stage 6: Import OPTIMISTIC on Eve (Bob's LND) ===
+	t.Logf("=== Stage 6: Import OPTIMISTIC on Eve " +
+		"(Bob's LND) ===")
+
+	// Decode the backup and verify federation URLs are present.
+	var decoded backup.WalletBackup
+	err = decoded.Decode(bytes.NewReader(optimisticBackup.Backup))
```
darioAnongba: shouldn't this call `backup.DecodeWalletBackup` ?

backup/import.go
```diff
@@ -0,0 +1,349 @@
...
+				results <- spendResult{
+					index: idx, spent: true,
+				}
+			case err := <-ec:
+				results <- spendResult{
+					index: idx, err: err,
+				}
+			case <-assetCtx.Done():
+				results <- spendResult{
+					index: idx, spent: false,
```
darioAnongba: shouldn't this return an error instead of `spent:false`? Since we reached the timeout.

backup/strip.go
jtobin: #1998 includes exactly such a mechanism, fwiw.

backup/import.go
GeorgeTsagk: > A tampered backup can register keys for asset A but import proof for asset B.

A tampered backup should be totally skipped as we have a checksum field

backup/import.go
GeorgeTsagk: We can't reorder — ImportProofs looks up existing key records by public key to associate the asset with full key locator info (family + index). Without keys registered first, the asset gets stored without signing info and becomes unspendable.

We did make key registration errors fatal (was warn-only), so a DB failure now aborts the import instead of silently producing unspendable assets.

backup/import.go
GeorgeTsagk: Added a `maxConcurrentSpendChecks = 100` channel-based semaphore in `detectSpentOutpoints`. All assets still get checked, but at most 100 goroutines run concurrently.

taprpc/assetwalletrpc/assetwallet.proto
GeorgeTsagk: The protobuf zero value must be RAW (field number 0), so an unset mode field defaults to RAW for safety. 

The CLI defaults to compact as a UX convenience since it produces smaller backups without requiring a universe server.

backup/encoding.go
GeorgeTsagk: Added a note

backup/import.go
GeorgeTsagk: Fixed — key registration errors are now fatal. If either InsertInternalKey or InsertScriptKey fails, the import aborts immediately with an error.

backup/import.go
GeorgeTsagk: Fixed — same as above, both key registration calls now return fatal errors.

backup/strip.go
GeorgeTsagk: Fixed — pulled in the TestStrippedEncodeRecordsSync test from #1998 which mechanically ensures strippedEncodeRecords stays in sync with proof.Proof.EncodeRecords().

taprpc/perms.go
GeorgeTsagk: Fixed — changed to "write" permission for ExportAssetWalletBackup.

itest/backup_test.go
GeorgeTsagk: Fixed — now calls backup.DecodeWalletBackup which includes checksum verification.

backup/import.go
GeorgeTsagk: Fixed — the timeout now distinguishes between the per-asset spend check timeout (asset not spent, return spent:false) and the parent context being cancelled (propagate the error). If the parent context is done, we return the context error; otherwise the per-asset timeout just means we didn't see a spend notification, so the asset is not stale.

GeorgeTsagk:
Rebased, let's see if CI is more green now

PR #1986, opened 2026-02-11
test: custom channel itests

darioAnongba:
This PR decouples the custom channel integration tests from `lightning-terminal` so they can run directly in `taproot-assets`. This eliminates the dependency on LiT's test harness and enables faster iteration on custom channel features.

### Approach
Instead of depending on LiT's full integrated terminal binary, we introduce a minimal tapd-integrated binary (`cmd/tapd-integrated`) that runs `lnd+tapd` in a single process with aux channel support. This binary is purpose-built for integration testing and wires `tapd's` aux components (funding controller, traffic shaper, channel closer, etc.) directly into `lnd` via the `integration` package.

The test harness (`IntegratedNode`, `IntegratedNetworkHarness`) manages these integrated binaries as external processes and provides gRPC clients for both `lnd` and `tapd` RPCs on a single port.

### Running
`make itest-cc`

### Specific test
`make itest-cc cccase=custom_channels_large`

### DIFF
--- a/itest/custom_channels/large_test.go
+++ b/itest/custom_channels/large_test.go
@@ -1,0 +1,188 @@
+//go:build itest
+
+package custom_channels
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/lightninglabs/taproot-assets/itest"
+	"github.com/lightninglabs/taproot-assets/proof"
+	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	"github.com/lightningnetwork/lnd/fn/v2"
+	"github.com/lightningnetwork/lnd/lntest"
+	"github.com/lightningnetwork/lnd/lntest/node"
+	"github.com/lightningnetwork/lnd/lntest/port"
+	"github.com/stretchr/testify/require"
+)
+
+// testCustomChannelsLarge tests a 5-node network topology with asset channels,
+// BTC channels, multi-hop asset payments, keysend payments, and cooperative
+// channel closing.
+//
+// Topology:
+//
+//	Charlie --[assets]--> Dave --[sats]--> Erin --[assets]--> Fabia
+//	                        |
+//	                     [assets]
+//	                        |
+//	                        v
+//	                      Yara
+//
+//nolint:lll
+func testCustomChannelsLarge(_ context.Context,
+	net *itest.IntegratedNetworkHarness, t *ccHarnessTest) {
+
+	lndArgs := slices.Clone(lndArgsTemplate)
+	tapdArgs := slices.Clone(tapdArgsTemplate)
+
+	// We use Charlie as the proof courier. But in order for Charlie to
+	// also use itself, we need to define its port upfront.
+	charliePort := port.NextAvailablePort()
+	tapdArgs = append(tapdArgs, fmt.Sprintf(
+		"--proofcourieraddr=%s://%s",
+		proof.UniverseRpcCourierType,
+		fmt.Sprintf(node.ListenerFormat, charliePort),
+	))
+
+	// Create all five nodes. Charlie gets a custom RPC port so it can
+	// reference itself as proof courier.
+	charlieLndArgs := slices.Clone(lndArgs)
+	charlieLndArgs = append(charlieLndArgs, fmt.Sprintf(
+		"--rpclisten=127.0.0.1:%d", charliePort,
+	))
+	charlie := net.NewNode("Charlie", charlieLndArgs, tapdArgs)
+	dave := net.NewNode("Dave", lndArgs, tapdArgs)
+	erin := net.NewNode("Erin", lndArgs, tapdArgs)
+	fabia := net.NewNode("Fabia", lndArgs, tapdArgs)
+	yara := net.NewNode("Yara", lndArgs, tapdArgs)
+
+	nodes := []*itest.IntegratedNode{charlie, dave, erin, fabia, yara}
+	connectAllNodes(t.t, net, nodes)
+	fundAllNodes(t.t, net, nodes)
+
+	// Create the normal channel between Dave and Erin.
+	t.Logf("Opening normal channel between Dave and Erin...")
+	channelOp := openChannelAndAssert(
+		t, net, dave, erin, lntest.OpenChannelParams{
+			Amt:         10_000_000,
+			SatPerVByte: 5,
+		},
+	)
+	defer closeChannelAndAssert(t, net, dave, channelOp, false)
+
+	// This is the only public channel, we need everyone to be aware of
+	// it.
+	assertChannelKnown(t.t, charlie, channelOp)
+	assertChannelKnown(t.t, fabia, channelOp)
+
+	// Mint an asset on Charlie and sync all nodes to Charlie as the
+	// universe.
+	mintedAssets := itest.MintAssetsConfirmBatch(
+		t.t, net.Miner.Client, asTapd(charlie),
+		[]*mintrpc.MintAssetRequest{
+			{
+				Asset: ccItestAsset,
+			},
+		},
+	)
+	cents := mintedAssets[0]
+	assetID := cents.AssetGenesis.AssetId
+
+	t.Logf("Minted %d lightning cents, syncing universes...",
+		cents.Amount)
+	syncUniverses(t.t, charlie, dave, erin, fabia, yara)
+	t.Logf("Universes synced between all nodes, distributing " +
+		"assets...")
+
+	const (
+		daveFundingAmount = uint64(400_000)
+		erinFundingAmount = uint64(200_000)
+	)
+	charlieFundingAmount := cents.Amount - uint64(2*400_000)
+
+	chanPointCD, _, _ := createTestAssetNetwork(
+		t, net, charlie, dave, erin, fabia, yara, charlie,
+		cents, 400_000, charlieFundingAmount,
+		daveFundingAmount, erinFundingAmount, DefaultPushSat,
+	)
+
+	// Before we start sending out payments, let's make sure each node
+	// can see the other one in the graph and has all required features.
+	require.NoError(t.t, net.AssertNodeKnown(charlie, dave))
+	require.NoError(t.t, net.AssertNodeKnown(dave, charlie))
+	require.NoError(t.t, net.AssertNodeKnown(dave, yara))
+	require.NoError(t.t, net.AssertNodeKnown(yara, dave))
+	require.NoError(t.t, net.AssertNodeKnown(erin, fabia))
+	require.NoError(t.t, net.AssertNodeKnown(fabia, erin))
+	require.NoError(t.t, net.AssertNodeKnown(charlie, erin))
+
+	// Print initial channel balances.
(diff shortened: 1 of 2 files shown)


itest/custom_channels/helpers.go
```diff
@@ -0,0 +1,1919 @@
...
+	return func(c *payConfig) {
+		c.allowSelfPayment = true
+	}
+}
+
+func withPayRouteHints(hints []*lnrpc.RouteHint) payOpt {
+	return func(c *payConfig) {
+		c.routeHints = hints
+	}
+}
```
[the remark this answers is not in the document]
darioAnongba: not for private functions only used as helpers, over-commenting.

jtobin (commented):
Took an initial pass over this, it looks solid. Will give the functionality a quick test in short order and then should be able to approve.

itest/custom_channels/helpers.go
```diff
@@ -0,0 +1,1924 @@
...
+	require.Equal(t, cfg.failureReason, payment.FailureReason)
+}
+
+// getAssetPaymentResult reads from the send payment stream until a final
+// payment result is received.
+func getAssetPaymentResult(t *testing.T,
+	s tchrpc.TaprootAssetChannels_SendPaymentClient,
+	isHodl bool) (*TapPayment, error) {
+
+	time.Sleep(time.Second)
```
GeorgeTsagk: nit: Hardcoded `time.Sleep(time.Second)` before reading the payment stream. The stream itself should be the synchronization mechanism — ideally this sleep isn't needed. If it's papering over a race in the RPC layer, it may flake in CI under load (too short) or slow down tests unnecessarily.

Is there a specific race this works around? A comment explaining it would be helpful.

itest/custom_channels/large_test.go
```diff
@@ -0,0 +1,188 @@
...
+	logBalance(t.t, nodes, assetID, "after invoice")
+
+	invoiceResp2 := createAssetInvoice(
+		t.t, dave, charlie, fabiaInvoiceAssetAmount/2, assetID,
+	)
+
+	// Sleep for a second to make sure the balances fully propagated
+	// before we make the payment. Otherwise, we'll make an RFQ order
+	// with a max amount of zero.
+	time.Sleep(time.Second * 1)
```
GeorgeTsagk: nit: The comment explains this works around balances not propagating in time, causing an RFQ order with max amount of zero. This suggests a real race in the balance propagation path.

Is there an event or polling-based check that could replace this, e.g. waiting until `getAssetChannelBalance()` returns non-zero for the expected asset? A sleep may be too short under CI load or wastefully long in the common case.

integration/aux_components.go
```diff
@@ -0,0 +1,95 @@
...
+
+// EnsureRequiredCustomMessages appends any message types required by the
+// Taproot Asset channel protocol to the given custom message allow list. In
+// particular, tapd needs to send error messages (lnwire.MsgError) to peers
+// through lnd's SendCustomMessage RPC. Since MsgError is outside the custom
+// message range, it must be explicitly allowed.
+//
+// This function is idempotent: if the required types are already present in
+// the slice, it returns it unchanged.
+func EnsureRequiredCustomMessages(customMsgs []uint16) []uint16 {
```
GeorgeTsagk, line 75: nit: Mutates the input slice via `append`. If the caller's slice has spare capacity, this silently writes into the shared backing array. Doc says "idempotent" but it's not pure — could surprise downstream consumers of this public API. Consider copying first.

cmd/tapd-integrated/main.go
```diff
@@ -0,0 +1,350 @@
...
+// onDemandListener is a net.Listener that defers the actual network listen
+// until Accept is first called. This allows us to pass a listener to lnd
+// before we know the exact port binding time. Copied from
+// lightning-terminal/config.go.
+type onDemandListener struct {
+	addr net.Addr
+	lis  net.Listener
+}
+
+func (l *onDemandListener) Accept() (net.Conn, error) {
```
GeorgeTsagk, line 86: nit: Data race on `l.lis` — check-then-assign without synchronization. A `sync.Once` would be safer since `net.Listener` is typically assumed concurrent-safe.

cmd/tapd-integrated/main.go
```diff
@@ -0,0 +1,350 @@
...
+	case err := <-lndErrChan:
+		return fmt.Errorf("lnd failed to start: %w", err)
+	case <-interceptor.ShutdownChannel():
+		wg.Wait()
+
+		return nil
+	}
+
+	// Step 7: Connect to lnd via bufconn for internal communication.
+	lndConn, err := dialBufConn(bufListener)
```
GeorgeTsagk: `lndConn` is created and defer-closed but never used — the actual lnd connection is made at step 8 via `lndclient.NewLndServices` on the external RPC address. `macChan` (line 196) is also unused. Vestigial from the LiT pattern? Can remove if not needed.

itest/integrated_harness.go
```diff
@@ -0,0 +1,566 @@
...
+		NodePubkey:         destNode.PubKey[:],
+		LocalFundingAmount: int64(p.Amt),
+		PushSat:            int64(p.PushAmt),
+		Private:            p.Private,
+		MinConfs:           minConfs,
+		SpendUnconfirmed:   p.SpendUnconfirmed,
+		MinHtlcMsat:        int64(p.MinHtlc),
+		RemoteMaxHtlcs:     uint32(p.RemoteMaxHtlcs),
+		FundingShim:        p.FundingShim,
+		SatPerByte:         int64(p.SatPerVByte),
```
GeorgeTsagk: nit: Maps `p.SatPerVByte` into the deprecated `SatPerByte` field — these differ for segwit txns. Should use `SatPerVbyte: uint64(p.SatPerVByte)`. Harmless today (value is always 0) but incorrect if a future test sets a fee rate.

Makefile
```diff
@@ -141,6 +141,9 @@ build-itest:
 	@$(call print, "Building itest lnd.")
 	CGO_ENABLED=0 $(GOBUILD) -mod=mod -tags="$(ITEST_TAGS)" -o itest/lnd-itest $(DEV_LDFLAGS) $(LND_PKG)/cmd/lnd
 
+	@$(call print, "Building itest tapd-integrated.")
+	CGO_ENABLED=0 $(GOBUILD) -tags="$(ITEST_TAGS)" -o itest/tapd-integrated-itest $(DEV_LDFLAGS) $(PKG)/cmd/tapd-integrated
```
GeorgeTsagk: nit: The lnd-itest build just above uses `-mod=mod` but this doesn't. Intentional?

itest/custom_channels/vars.go
```diff
@@ -0,0 +1,100 @@
...
+var tapdArgsTemplateNoOracle = []string{
+	"--allow-public-uni-proof-courier",
+	"--universe.public-access=rw",
+	"--universe.sync-all-assets",
+	"--universerpccourier.skipinitdelay",
+	"--universerpccourier.backoffresetwait=1s",
+	"--universerpccourier.numtries=5",
+	"--universerpccourier.initialbackoff=300ms",
+	"--universerpccourier.maxbackoff=600ms",
+	"--universerpccourier.skipinitdelay",
```
GeorgeTsagk: Lines 53-57 and 58-62 are duplicates with conflicting `backoffresetwait` (`1s` vs `100ms`). Last-value-wins so effective value is `100ms`. Copy-paste artifact from LiT — should deduplicate.

itest/custom_channels/helpers.go
```diff
@@ -0,0 +1,1924 @@
...
+
+	stream, err := payer.TaprootAssetChannelsClient.SendPayment(
+		ctxt, request,
+	)
+	require.NoError(t, err)
+
+	if cfg.errSubStr != "" {
+		msg, err := stream.Recv()
+
+		if msg.GetAcceptedSellOrders() != nil {
```
GeorgeTsagk, line 1118: Potential nil deref: if `stream.Recv()` on line 1111 returns an error, `msg` is nil and `msg.GetAcceptedSellOrders()` panics. Need to check `err` before accessing `msg`.

itest/custom_channels/helpers.go
GeorgeTsagk, line 1230: nit: Inherited from LiT (`23039a92b` by guggero) where the original comment reads: *"No idea why it makes a difference whether we wait before calling s.Recv() or not, but it does. Without the sleep, the test will fail with 'insufficient local balance'... Probably something weird within lnd itself."* — that comment was dropped during the port. Worth carrying it over, and revisiting this sleep in the future.

itest/custom_channels/large_test.go
GeorgeTsagk, line 143: nit: Inherited from LiT (`23039a92b` by guggero). Known workaround for a balance propagation race. Fine to keep for now but worth revisiting in the future — ideally replaced with a polling check.

GeorgeTsagk (commented):
ran the test locally, it might have flaked but it's not very frequent

overall it's looking alright
Should we port all tests and see how they behave?

We can also replace the CI job of LiT cc tests with a local cc test job.

If you want to do things incrementally and add the full test suite / replace CI jobs in different PRs that's fine by me.

GeorgeTsagk:
Oh another important note: we should maintain the per-daemon logs same way LiT does today, it's very easy to deep dive and find bugs this way

darioAnongba:
> If you want to do things incrementally and add the full test suite / replace CI jobs in different PRs that's fine by me.

Yes, subsequent PRs are activating the workflow + moving tests from LiT incrementally and then finally removing the worfklow completely. 

> Oh another important note: we should maintain the per-daemon logs same way LiT does today, it's very easy to deep dive and find bugs this way

Coming in this PR: https://github.com/lightninglabs/taproot-assets/pull/1988

integration/aux_components.go
darioAnongba: addressed

cmd/tapd-integrated/main.go
darioAnongba: addressed

cmd/tapd-integrated/main.go
darioAnongba: addressed

itest/integrated_harness.go
darioAnongba: addressed

Makefile
darioAnongba: addressed

itest/custom_channels/vars.go
darioAnongba: addressed

itest/custom_channels/helpers.go
darioAnongba: addressed

itest/custom_channels/helpers.go
darioAnongba: restored original comment for now.

itest/custom_channels/large_test.go
darioAnongba: skipped for now but will revisit in next PRs.

GeorgeTsagk (approved):
Looks good. Looking forward to future iterations

PR #1988, opened 2026-02-11
itest: separate tapd instance logs into per-node files

### LINKED ISSUE #1987: itest: separate tapd instance logs into per-node files
When running integration tests with multiple tapd instances (e.g., Alice, Bob, Dave, Erin), all tapd logs are intermingled in a single console output alongside the test logs. This makes debugging multi-node scenarios very difficult.

### Motivation

As integration tests have evolved from single-node "end-to-end" tests to true multi-node scenarios (minting on Alice, sending to Bob, validating on Dave), each tapd instance is independent and its logs should be separate. Currently:

- All tapd subsystem logs (TAPD, TSVR,, GRDN, etc.) from **all** tapd instances go to the same console output.
- Test-level logs (`t.Logf`) are mixed in with daemon logs.
- When a test fails, identifying which tapd instance produced an error requires manually parsing interleaved log lines.
- LND node logs are already written to separate per-node files (via `lntest/node` framework), making tapd behave differently.
- AI agents need separate, small and persisted log files for easier debugging.

### Current Architecture (Why This Happens)

tapd instances run **in-process** during itests. The logging setup in `test_harness.go:setupLogging()` creates a single shared `SubLoggerManager` and calls `tap.SetupLoggers()` once, which replaces **all** global package-level loggers. Every tapd instance created via `CreateServerFromConfig()` shares these global loggers.

The `RotatingLogWriter` is never initialized with `InitLogRotator`, so all log output goes to console (stdout) only.

Since Go package-level loggers are process-global, there is no way to separate per-instance logs when multiple instances run in the same process.

### Expected Log Structure After

```
regtest/
  0-testBasicSend-Alice.log              # LND node logs
  1-testBasicSend-Bob.log                # LND node logs
  tapd-testBasicSend-Alice.log           # tapd instance logs
  tapd-testBasicSend-Bob.log             # tapd instance logs
  tapd-testBasicSend-uni-server.log      # Universe server tapd logs
  output_btcd_chainbackend.log           # Chain backend logs
```

Each tapd log file contains the full daemon output for that specific instance, with all subsystem prefixes (TAPD, TSVR, TRPC, GRDN, PRUF, etc.) scoped to that one node.

darioAnongba:
When debugging multi-node integration test failures, all tapd logs were intermingled in console output, making it very difficult to trace which instance produced which log line. With this change, each `tapd` instance writes to its own log file in the test log directory.

- Run each tapd instance as a separate OS process (`tapd-itest` binary) during integration tests, enabling per-instance log files (e.g., `tapd-Alice.log`, `tapd-Bob.log`)
- Go's global package-level loggers make in-process log separation impossible — calling `SetupLoggers` for one instance overwrites all others. This matches the approach used by `lnd` and `lightning-terminal` itests.
- Add `setBoolFlag`, `setCliFlag`, and `updateLndNode` helpers for test-time config changes between stop/start cycles.

Fixes #1987

### DIFF
--- a/itest/rfq_forwards_test.go
+++ b/itest/rfq_forwards_test.go
@@ -33,7 +33,7 @@ import (
 //  6. We query Bob's forwarding history and verify the event was recorded.
 //  7. We test query filters (timestamp, peer, asset_id) and pagination.
 func testForwardingEventHistory(t *harnessTest) {
-	oracleAddr := fmt.Sprintf("localhost:%d", port.NextAvailablePort())
+	oracleAddr := fmt.Sprintf("127.0.0.1:%d", port.NextAvailablePort())
 	oracle := newOracleHarness(oracleAddr)
 	oracle.start(t.t)
 	t.t.Cleanup(oracle.stop)
--- a/itest/rfq_test.go
+++ b/itest/rfq_test.go
@@ -54,7 +54,7 @@ var (
 // the tap asset to Carol's node.
 func testRfqAssetBuyHtlcIntercept(t *harnessTest) {
 	// For this test we'll use an actual oracle RPC server harness.
-	oracleAddr := fmt.Sprintf("localhost:%d", port.NextAvailablePort())
+	oracleAddr := fmt.Sprintf("127.0.0.1:%d", port.NextAvailablePort())
 	oracle := newOracleHarness(oracleAddr)
 	oracle.start(t.t)
 	t.t.Cleanup(oracle.stop)
@@ -336,7 +336,7 @@ func testRfqAssetBuyHtlcIntercept(t *harnessTest) {
 // sell request.
 func testRfqAssetSellHtlcIntercept(t *harnessTest) {
 	// For this test we'll use an actual oracle RPC server harness.
-	oracleAddr := fmt.Sprintf("localhost:%d", port.NextAvailablePort())
+	oracleAddr := fmt.Sprintf("127.0.0.1:%d", port.NextAvailablePort())
 	oracle := newOracleHarness(oracleAddr)
 	oracle.start(t.t)
 	t.t.Cleanup(oracle.stop)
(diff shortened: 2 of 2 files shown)


litbot-9000:

jtobin (approved):
LGTM. :+1: :+1:

itest/rfq_forwards_test.go
```diff
@@ -33,7 +33,7 @@ import (
 //  6. We query Bob's forwarding history and verify the event was recorded.
 //  7. We test query filters (timestamp, peer, asset_id) and pagination.
```
GeorgeTsagk, line 34: last 2 commits seem to overlap, squash?

itest/tapd_harness.go
```diff
@@ -520,16 +657,108 @@ func (hs *tapdHarness) start(expectErrExit bool) error {
...
-	// remove the temp directory.
-	err := hs.server.Stop()
+	var stopErr error
+
+	// Try graceful shutdown via RPC first.
+	if hs.TaprootAssetsClient != nil {
+		ctx, cancel := context.WithTimeout(
+			context.Background(), 5*time.Second,
+		)
+		_, err := hs.StopDaemon(ctx, &taprpc.StopRequest{})
```
GeorgeTsagk, line 704: tapd seems to hang every time it shuts down and gets force-killed

GeorgeTsagk: ok that's weird:
https://github.com/lightninglabs/taproot-assets/blob/main/server.go#L841

we do a `.Start()` here instead of `.Stop()`. Probably that's causing it?

itest/rfq_forwards_test.go
darioAnongba: hmmm true, it happened because I renamed the Merge commit thinking that I squashed into this branch but both commits went it, rebasing!

itest/tapd_harness.go
darioAnongba: yes, this has always been a problem as far as I recall and not introduced in this PR, I'll fix it though (in a different PR).

PR #1990, opened 2026-02-12
rfqmath: sounder fixed-point comparison

jtobin:
Adds a normalizing comparison method for FixedPoint values, and uses it when comparing quotes.

Without normalization, you can theoretically get stuff like this (example ctsy Opus):

```
  Rate A: 50,000 asset units per BTC (scale 0)
    → Coefficient=50000, Scale=0
    → Actual value: 50000 / 10^0 = 50,000

  Rate B: 49,000 asset units per BTC (scale 2)
    → Coefficient=4900000, Scale=2
    → Actual value: 4900000 / 10^2 = 49,000

  Rate A > Rate B (50,000 > 49,000)

  But comparing raw coefficients (what ToUint64() returns):

  A.ToUint64() = 50,000
  B.ToUint64() = 4,900,000

  50,000 < 4,900,000  →  A < B
```

This comparison pops up for us in practice when comparing quotes. Previously we used this:

```
  (rpcserver.go:9637-9638 on 'main')

  sort.Slice(acquiredQuotes, func(i, j int) bool {
      return acquiredQuotes[i].rate.ToUint64() <
          acquiredQuotes[j].rate.ToUint64()
  })
```

the PR also changes that to use the new method, à la:

```
  sort.Slice(acquiredQuotes, func(i, j int) bool {
      return acquiredQuotes[i].rate.Cmp(*acquiredQuotes[j].rate) < 0
  }
```

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -52,6 +52,10 @@
   addresses a bug in which passive assets would incorrectly be included in
   channel funding proofs.
 
+* [PR#1990](https://github.com/lightninglabs/taproot-assets/pull/1990)
+  prevents buggy results when comparing quotes encoded using different
+  scales.
+
 # New Features
 
 ## Functional Enhancements


rfqmath/fixed_point_test.go
```diff
@@ -146,6 +146,262 @@ func testEquality[N Int[N]](t *rapid.T) {
...
+	fp1 := FixedPointFromUint64[N](coefficient, scale1)
+	fp2 := FixedPointFromUint64[N](coefficient, scale2)
+
+	result := fp1.Cmp(fp2)
+
+	require.Equal(
+		t, 0, result, "Cmp should return 0 for equal values at "+
+			"different scales: Cmp(%v, %v) = %d", fp1, fp2, result,
+	)
+}
```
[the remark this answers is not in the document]
jtobin: Fixed in 6d03779bdcd4509352c809cc53d9cb2e2d71378b.

rfqmath/fixed_point_test.go
```diff
@@ -146,6 +146,272 @@ func testEquality[N Int[N]](t *rapid.T) {
...
+			expected: -1,
+		},
+		{
+			name:     "first greater than second same scale",
+			fp1:      FixedPointFromUint64[N](200, 2),
+			fp2:      FixedPointFromUint64[N](100, 2),
+			expected: 1,
+		},
+		{
+			name:     "first less than second different scales",
```
darioAnongba:
```suggestion
			name:     "first less than second same scale",
```
scale is 0 for both

jtobin: Fixed in 04ccb7ab5a7ea17fc12539609c7111f0314d450d.

PR #1993, opened 2026-02-13
ci: use claude opus and allow PR branch checkout

darioAnongba:
Align Claude capabilities with LND and use Opus instead of default model (Sonnet).

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -29,6 +29,14 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 1
+          
+      - name: Checkout PR branch (handles fork PRs)
+        if: github.event.issue.pull_request || github.event_name == 'pull_request_review_comment' || github.event_name == 'pull_request_review'
+        env:
+          GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


.github/workflows/claude.yml
```diff
@@ -29,6 +29,18 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 1
+          
+      - name: Checkout PR branch (handles fork PRs)
+        if: github.event.issue.pull_request || github.event_name == 'pull_request_review_comment' || github.event_name == 'pull_request_review'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
```
jtobin, line 38: Nittish -- supposedly this could be handled like so (ctsy Opus):

```
env:
  GH_TOKEN: ${{ github.token }}
  PR_NUMBER: ${{ github.event.issue.number || github.event.pull_request.number }}
run: |
  gh pr checkout "$PR_NUMBER"
```

The rationale is that if it's an issue comment then issue.number is defined; if PR stuff, then issue.number is null, and the *or* falls through to pull_request.number, which is defined. A bit neater, maybe worth changing.

darioAnongba: ok trying this.

GeorgeTsagk (approved):

PR #1995, opened 2026-02-16
ListAssets pagination

### LINKED ISSUE #875: [feature]: Paginate `ListAssets` RPC

darioAnongba:
Add pagination support (`offset`, `limit`, `direction`) to the `ListAssets` RPC, following the same pattern as `AddrReceives` and other paginated endpoints

Fixes #875

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -167,7 +167,15 @@
   import result (with a timeout) before returning, surfacing mailbox courier
   import failures instead of racing and returning success early.
 
+- [PR#1995](https://github.com/lightninglabs/taproot-assets/pull/1995)
+  Add pagination support (offset, limit, direction) to the `ListAssets` RPC
+  endpoint.
+
 ## tapcli Updates
+
+- [PR#1995](https://github.com/lightninglabs/taproot-assets/pull/1995)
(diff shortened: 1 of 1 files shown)


docs/release-notes/release-notes-0.8.0.md
```diff
@@ -171,8 +167,16 @@
...
 
+- [PR#1995](https://github.com/lightninglabs/taproot-assets/pull/1995)
+  Add pagination support (offset, limit, direction) to the `ListAssets` RPC
+  endpoint.
+
 ## tapcli Updates
 
+- [PR#1995](https://github.com/lightninglabs/taproot-assets/pull/1995)
+  Add `--limit`, `--offset`, and `--direction` flags to `tapcli assets list`
+  for pagination support. The direction defaults to ascending order.
```
jtobin: Opus+Codex claim that this should actually be *descending* order and I think they're right. Worth double checking.

darioAnongba: yes, it's descending, my mistake.

PR #2002, opened 2026-02-19
fix: rename sweep-orphan-utxos to disable-sweep-orphan-utxos

darioAnongba:
The `wallet.sweep-orphan-utxos` flag defaults to `true` (enabled), but go-flags doesn't support `--flag=false` for boolean options, making it impossible to disable sweeping via CLI.


jtobin (approved):
LGTM. :+1: :+1:

Issue #2000, opened 2026-02-18
[bug]: transfer proof stuck PENDING, listAssets zero


beeforbacon:
## Description

**Environment**
- Node: mcu-616-main (Voltage)
- TAPD Version: 0.7.0-alpha
- TAPD REST: [https://xxxx.m.voltageapp.io:443](https://xxxx.m.voltageapp.io:443)
- Chain: sync_to_chain=true, block_height in sync

**Issue**
One outbound Taproot Assets transfer has been confirmed on-chain for over **4 hours**. In `ListTransfers`, one output (the one to the recipient) still shows  
`proof_delivery_status = PROOF_DELIVERY_STATUS_PENDING`. As a result, `getMcu616Balance()` (and thus `listAssets` for that asset) returns **0**, and we cannot send new asset batches for orders.

---

**Self-check (via TAPD REST)**
- Universe federation: 1 server — `universe.lightning.finance:10029`
  (default mainnet Universe present).
- Universe sync config (`GET /v1/taproot-assets/universe/sync/config`):
  - **PROOF_TYPE_TRANSFER**: 
    - `allow_sync_insert=false`
    - `allow_sync_export=true`
  - **PROOF_TYPE_ISSUANCE**: 
    - `allow_sync_insert=false`
    - `allow_sync_export=true`

The node exports proofs to the federation but does not insert (pull) proofs from it. We suspect this is why the transfer output never leaves **PENDING** and our balance stays **zero**.

---

**Request**
Please enable `allow_sync_insert` for **PROOF_TYPE_TRANSFER** (or the equivalent setting on your side) so that this node can pull transfer proofs from the federation and the delivery status can move to **COMPLETE**. If your default policy is to keep insert disabled, we would like to understand the recommended way to resolve persistent **PENDING** and zero `listAssets` after a chain-confirmed transfer.

**Recipient Address (for Reference)**
The pending output is for this MCU616 address (decoded successfully, asset_id and amount match):  
[...]

PR #2010, opened 2026-02-27
tapdb: cascade deletions to federation sync log foreign keys

jtobin:
Previously, calling DeleteAssetRoot on a universe that had been synced via federation would fail with a foreign key constraint error. Now the deletion correctly cleans up both federation sync log entries and orphaned multiverse roots.

I also tested this e2e on a Docker regtest network with mint/sync/delete/re-mint/re-sync steps; all worked as expected.

### DIFF
--- a/tapdb/migrations_test.go
+++ b/tapdb/migrations_test.go
@@ -1139,3 +1139,143 @@ func TestDirtySqliteVersion(t *testing.T) {
 	)
 	require.ErrorContains(t, err, "database is in a dirty state")
 }
+
+// TestMigration54 tests that migration 54 adds ON DELETE CASCADE to
+// the federation_proof_sync_log foreign keys referencing
+// universe_leaves and universe_roots.
+func TestMigration54(t *testing.T) {
+	ctx := context.Background()
+
+	// Create a DB at version 53 (pre-CASCADE).
+	db := NewTestDBWithVersion(t, 53)
+
+	// Insert dummy data: a universe root, leaf, server, and two
+	// federation_proof_sync_log rows referencing them.
+	InsertTestdata(
+		t, db.BaseDB,
+		"migrations_test_00054_dummy_data.sql",
+	)
+
+	// Verify the sync log rows exist.
+	var count int
+	err := db.QueryRowContext(ctx, `
+		SELECT COUNT(*)
+		FROM federation_proof_sync_log
+	`).Scan(&count)
+	require.NoError(t, err)
+	require.Equal(t, 2, count)
+
+	// Pre-migration: deleting the universe leaf should fail due
+	// to the FK constraint from federation_proof_sync_log.
+	_, err = db.ExecContext(ctx, `
+		DELETE FROM universe_leaves WHERE id = 1
+	`)
+	require.Error(t, err, "delete should fail before migration")
+
+	// Apply migration 54.
+	err = db.ExecuteMigrations(TargetLatest)
+	require.NoError(t, err)
+
+	// Verify data survived the table recreation in the
+	// up-migration (CREATE new → copy → DROP old → RENAME).
+	err = db.QueryRowContext(ctx, `
+		SELECT COUNT(*)
+		FROM federation_proof_sync_log
+	`).Scan(&count)
+	require.NoError(t, err)
+	require.Equal(t, 2, count, "sync log rows must survive "+
+		"up-migration table recreation")
+
+	err = db.QueryRowContext(ctx, `
+		SELECT COUNT(*)
+		FROM universe_leaves
+	`).Scan(&count)
+	require.NoError(t, err)
+	require.Equal(t, 1, count, "universe leaf must survive "+
+		"up-migration table recreation")
+
+	// Post-migration: deleting the universe leaf should CASCADE
+	// and remove the sync log rows.
+	_, err = db.ExecContext(ctx, `
+		DELETE FROM universe_leaves WHERE id = 1
+	`)
+	require.NoError(t, err)
+
+	err = db.QueryRowContext(ctx, `
+		SELECT COUNT(*)
+		FROM federation_proof_sync_log
+	`).Scan(&count)
+	require.NoError(t, err)
+	require.Equal(t, 0, count)
+
+	// Verify the universe leaf is gone too.
+	err = db.QueryRowContext(ctx, `
+		SELECT COUNT(*)
+		FROM universe_leaves
+	`).Scan(&count)
+	require.NoError(t, err)
+	require.Equal(t, 0, count)
+
+	// Re-insert the universe leaf (the CASCADE deleted it along
+	// with the sync log rows, but the parent rows survived).
+	// Use a second test data file to avoid hex-literal
+	// incompatibilities between SQLite and Postgres.
+	InsertTestdata(
+		t, db.BaseDB,
+		"migrations_test_00054_re_insert_leaf.sql",
+	)
+	require.NoError(t, err)
+
+	// Verify that new sync log rows can be inserted without PK
+	// collisions (exercises the Postgres BIGSERIAL sequence).
+	_, err = db.ExecContext(ctx, `
+		INSERT INTO federation_proof_sync_log (
+			status, timestamp, attempt_counter,
+			sync_direction, proof_leaf_id,
+			universe_root_id, servers_id
+		) VALUES (
+			'pending', '2024-02-01', 0,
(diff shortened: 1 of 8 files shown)


jtobin:
(This is tangentially related to #2000.)

tapdb/testdata/migrations_test_00054_dummy_data.sql
```diff
@@ -0,0 +1,66 @@
...
+    1, 'ns1', NULL,
+    X'000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f',
+    'issuance'
+);
+
+-- universe_leaves
+INSERT INTO universe_leaves (
+    id, asset_genesis_id, minting_point, script_key_bytes,
+    universe_root_id, leaf_node_key, leaf_node_namespace
+) VALUES (
```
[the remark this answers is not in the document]
jtobin: Test seems to run fine.

Roasbeef:
> DeleteAssetRoot

Don't we only want to delete a _leaf_ and not the entire root?

litbot-9000:
@georgetsagk: review reminder

tapdb/sqlc/migrations/000054_fed_sync_log_cascade.up.sql
```diff
@@ -0,0 +1,51 @@
+-- Add ON DELETE CASCADE to federation_proof_sync_log FKs that
+-- reference universe_leaves(id) and universe_roots(id).
+--
+-- SQLite cannot ALTER constraints, so we recreate the table.
+
+-- Step 1: Create replacement table with CASCADE constraints.
+CREATE TABLE new_federation_proof_sync_log (
```
darioAnongba, line 7: Big severity bug caught by agent:

> On Postgres, `replacerFS` transforms `INTEGER PRIMARY KEY` into `BIGSERIAL PRIMARY KEY`, which creates an implicit sequence starting at 1. The `INSERT ... SELECT *` copies existing id values verbatim but does NOT advance the sequence via `nextval()`.
> 
> After migration, the next `UpsertFederationProofSyncLog` call (which omits the `id` column, relying on the sequence) will call `nextval()` returning 1, causing:
> 
> ```
> ERROR: duplicate key value violates unique constraint "new_federation_proof_sync_log_pkey"
> ```
> 
> The `ON CONFLICT` clause in `UpsertFederationProofSyncLog` is on `(sync_direction, proof_leaf_id, universe_root_id, servers_id)`, not on `id`, so this PK collision is NOT handled.
> 
> **Impact**: Federation sync completely broken on Postgres after migration. New sync log entries cannot be created, so all proof push/pull operations fail. Triggers automatically with zero attacker involvement on any Postgres node with existing sync log data.

most likely not caught here because we don't try inserting data after the migration test?

tapdb/sqlc/schemas/generated_schema.sql
```diff
@@ -440,30 +440,29 @@ CREATE TABLE federation_global_sync_config (
...
-    -- The direction of the proof sync attempt.
-    sync_direction TEXT NOT NULL CHECK(sync_direction IN ('push', 'pull')),
+    sync_direction TEXT NOT NULL CHECK(
+        sync_direction IN ('push', 'pull')
+    ),
 
-    -- The ID of the subject proof leaf.
-    proof_leaf_id BIGINT NOT NULL REFERENCES universe_leaves(id),
+    proof_leaf_id BIGINT NOT NULL
+        REFERENCES universe_leaves(id) ON DELETE CASCADE,
```
darioAnongba, line 459: one issue with this ON DELETE CASCADE at the schema level is that any code path (current or future) that deletes universe leaves will silently destroy the sync state.

Why not use directly use `DELETE FROM federation_proof_sync_log` inside `deleteUniverseTree` instead of schema level CASCADE?

tapdb/sqlc/migrations/000054_fed_sync_log_cascade.up.sql
```diff
@@ -0,0 +1,51 @@
...
+
+    sync_direction TEXT NOT NULL CHECK(
+        sync_direction IN ('push', 'pull')
+    ),
+
+    proof_leaf_id BIGINT NOT NULL
+        REFERENCES universe_leaves(id) ON DELETE CASCADE,
+
+    universe_root_id BIGINT NOT NULL
+        REFERENCES universe_roots(id) ON DELETE CASCADE,
```
darioAnongba, line 26: do we need CASCADE on this FK?

tapdb/migrations_test.go
```diff
@@ -1139,3 +1139,62 @@ func TestDirtySqliteVersion(t *testing.T) {
 	)
 	require.ErrorContains(t, err, "database is in a dirty state")
 }
+
+// TestMigration54 tests that migration 54 adds ON DELETE CASCADE to
+// the federation_proof_sync_log foreign keys referencing
+// universe_leaves and universe_roots.
+func TestMigration54(t *testing.T) {
```
darioAnongba, line 1146: worth testing the Down?

tapdb/sqlc/migrations/000054_fed_sync_log_cascade.down.sql
```diff
@@ -0,0 +1,49 @@
...
+    universe_root_id BIGINT NOT NULL
+        REFERENCES universe_roots(id),
+
+    servers_id BIGINT NOT NULL
+        REFERENCES universe_servers(id)
+);
+
+-- Step 2: Copy existing data.
+INSERT INTO new_federation_proof_sync_log
+SELECT * FROM federation_proof_sync_log;
```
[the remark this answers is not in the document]
jtobin: Fixed in 87ef6225.

tapdb/sqlc/migrations/000054_fed_sync_log_cascade.up.sql
```diff
@@ -0,0 +1,51 @@
...
+
+    servers_id BIGINT NOT NULL
+        REFERENCES universe_servers(id)
+);
+
+-- Step 2: Copy existing data.
+INSERT INTO new_federation_proof_sync_log
+SELECT * FROM federation_proof_sync_log;
```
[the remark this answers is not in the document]
jtobin: Fixed in 87ef6225.

tapdb/sqlc/migrations/000054_fed_sync_log_cascade.up.sql
jtobin: Fixed in 87ef6225.

tapdb/migrations_test.go
jtobin: Added in 63ae34a4.

tapdb/sqlc/migrations/000054_fed_sync_log_cascade.up.sql
jtobin: Post-migration insert test also added in fea0ce41a9315a6e2b6a59ae9494c991d0ce0b2b.

tapdb/sqlc/schemas/generated_schema.sql
jtobin: I think it's ok, because any sync state that would be left post-delete would reference stuff (i.e. deleted stuff) that no longer exists anyway. It'd still have to be cleaned up, so why not get CASCADE to handle it automatically sort of.

tapdb/sqlc/migrations/000054_fed_sync_log_cascade.up.sql
jtobin: I believe so. IIUC: the foreign key is on the sync log, which references a `universe_roots` row. deleteUniverseTree deletes the `universe_roots` row, so without the CASCADE we'll hit the same constraint error if there happen to be any sync log entries.

GeorgeTsagk:
Can we also add coverage for the following:

  - The actual bug scenario (deleting a universe with sync log entries) is never tested end-to-end
  - Only the "last universe" case is tested. Deleting one-of-many under the same proof type is not
  - Row data survival through the table recreation is never verified, in either migration direction

jtobin:
> Can we also add coverage for the following:
> 
> * The actual bug scenario (deleting a universe with sync log entries) is never tested end-to-end
> * Only the "last universe" case is tested. Deleting one-of-many under the same proof type is not
> * Row data survival through the table recreation is never verified, in either migration direction

All of these should be addressed in the latest changeset.

GeorgeTsagk (approved):

PR #2014, opened 2026-03-03
FetchAsset RPC 

### LINKED ISSUE #877: [feature]: Add `LookUpAsset` RPC
## Description

If we want to retrieve asset information via our RPC we are currently limited to using `ListAssets` RPC. Even though this currently works alright it could be inefficient when the number of owned assets grows larger.

This could be somewhat improved with filters as described in https://github.com/lightninglabs/taproot-assets/issues/876, but maybe an explicit RPC call for this is also useful.

We should add a call that takes a specific `asset_id` (or a group key) and returns all assets currently owned by the wallet. Could also use similar params to `ListAssets` like `include_spent`/`include_leased`.

darioAnongba:
- Add `FetchAsset` RPC that fetches assets by asset ID or group key using an `AssetSpecifier`, providing a more efficient alternative to `ListAssets` for targeted lookups.
- Includes filters for `include_spent`, `include_leased`, `include_unconfirmed_mints`, `with_witness`, and `script_key_type`.
- Add `tapcli assets fetch` command with `--asset_id` and `--group_key` flags.

Closes #877.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -124,6 +124,11 @@
   the last leaf is deleted, the entire universe is automatically cleaned
   up.
 
+- [PR#2014](https://github.com/lightninglabs/taproot-assets/pull/2014)
+  Add `FetchAsset` RPC for fetching assets by asset ID or group key using an
+  `AssetSpecifier`. Supports filters for `include_spent`, `include_leased`,
+  `include_unconfirmed_mints`, `with_witness`, and `script_key_type`.
+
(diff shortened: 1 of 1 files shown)


darioAnongba:
> cACK, tACK, LGTM. 👍 👍
> 
> (One thing that agents flagged during review: ListAssets rejects requests where `include_spent` and `include_leased` are set, but FetchAsset does not. Actually seems like the ListAssets is wrong here?)

yeah looks wrong on the `ListAssets` one.

litbot-9000:
@georgetsagk: review reminder

PR #2017, opened 2026-03-06
ci: misc refactoring

jtobin:
I recently took a pass over our CI in the wake of @darioAnongba's custom channels itest porting and this was the result:

Changes that mostly involve tuning existing stuff:

* Consolidate some itest build jobs to eliminate redundant builds.
* Improve caching, employing the same approach used in lightninglabs/lightning-terminal#1203.
* Use the official golangci-lint GitHub Actions action instead of using Docker to run it.
* Better-balance job assignments across the unit-race and itest tranches, and tune the unit-race parallelism.

Changes that propose to remove existing stuff (and indeed, to merge this we'll need to make some changes to the "required jobs" settings in the repo):

* Get rid of deprecated jobs (e.g. litd setup).
* Drop the cleanup-space step; there is plenty of space available on the runners for our current setup.
* Don't run the cross-compile, Docker build check, and postgres itests when pushing to PR branches; the frequency of these seems unnecessary, and they still run in the merge queue anyway.
* For the unit tests w/the postgres backend, only run those for tapdb; those are the only tests for which the `test_db_postgres` build tag is actually consumed.

The result is that the CI suite runs in about 10-12m on every push to a PR branch, which is probably close to the lower bound achievable for us.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -45,16 +45,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup go ${{ env.GO_VERSION }}
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: '${{ env.GO_VERSION }}'
-
-      - name: docker image cache
-        uses: jpribyl/action-docker-layer-caching@v0.1.1
-        continue-on-error: true
 
       - name: Generate sql models
         run: make sqlc-check
@@ -65,10 +61,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup go ${{ env.GO_VERSION }}
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: '${{ env.GO_VERSION }}'
 
@@ -81,10 +77,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup go ${{ env.GO_VERSION }}
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: '${{ env.GO_VERSION }}'
 
@@ -97,10 +93,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup go ${{ env.GO_VERSION }}
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: '${{ env.GO_VERSION }}'
 
@@ -113,7 +109,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Run migration check
         run: make migration-check
@@ -124,10 +120,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup go ${{ env.GO_VERSION }}
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: '${{ env.GO_VERSION }}'
 
@@ -143,10 +139,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup go ${{ env.GO_VERSION }}
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: '${{ env.GO_VERSION }}'
 
@@ -159,7 +155,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
@@ -199,7 +195,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Setup go ${{ env.GO_VERSION }}
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: '${{ env.GO_VERSION }}'
 
@@ -212,12 +208,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: git checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           fetch-depth: 0
 
       - name: Setup go ${{ env.GO_VERSION }}
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: '${{ env.GO_VERSION }}'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -178,8 +178,10 @@ jobs:
           tags: 'dev-taproot-assets'
 
   # Cross-compile release binaries across supported targets.
+  # Skip on PRs — the merge queue still gates landing.
```
Roasbeef: We don't always use the merge queue though. So I think we should keep this on until we (maybe) switch to merge queue as the default route.

Roasbeef: In the past we used it, but the issue was that the occasional flake kick something off (and/or halt the entire thing?).

.github/workflows/main.yaml
```diff
@@ -153,8 +153,10 @@ jobs:
         run: make sample-conf-check
 
   # Smoke test Docker builds for default and dev images.
+  # Skip on PRs — the merge queue still gates landing.
   docker-build-check:
     name: Docker build check
+    if: github.event_name != 'pull_request'
```
Roasbeef: Same here re merge queue not currently being the _forced_ default.

.github/actions/cleanup-space/action.yaml
```diff
@@ -37,33 +37,5 @@ runs:
           remove_path "$AGENT_TOOLSDIRECTORY"
         fi
 
-        # Remove docker images to save space.
-        if command -v docker >/dev/null 2>&1; then
```
Roasbeef, line 41: So we had this before as otherwise eventually the cache just sorta grows forever, which led to that release build failing that one time. 

Perhaps we can something to only occasionally run it? and/or always run it _between_ release build phases.

.github/workflows/main.yaml
```diff
@@ -344,9 +344,6 @@ jobs:
       - name: git checkout
         uses: actions/checkout@v5
 
-      - name: Clean up runner space
-        uses: ./.github/actions/cleanup-space
```
Roasbeef, line 346: Same comment here rationale we had this: release build failed as there wasn't enough space left on the runner. 

Not sure where that 91 GB number is from.

.github/workflows/main.yaml
```diff
@@ -379,9 +376,6 @@ jobs:
...
```
Roasbeef, line 381: I think it can make sense here. However is the cache shared across _all_ builds?

.github/workflows/main.yaml
```diff
@@ -224,8 +224,16 @@ jobs:
         with:
           go-version: '${{ env.GO_VERSION }}'
 
-      - name: run lint
```
Roasbeef, line 228: Wait so we aren't running the linter at all anymore?

The old command also passed some options and ran lint in parallel: https://github.com/lightninglabs/taproot-assets/blob/e390257e98d1913217632cf19f10ad9d7784c61a/Makefile#L380-L384.

.github/workflows/main.yaml
```diff
@@ -298,332 +298,332 @@ jobs:
...
-            itest/custom_channels/regtest/logs-tranche*/*.log || true
-
-      - name: Upload log files on failure
-        uses: actions/upload-artifact@v4
-        if: ${{ failure() }}
-        with:
-          name: logs-cc-itest-${{ matrix.tranche }}
-          path: logs-cc-itest-${{ matrix.tranche }}.zip
-          retention-days: 5
+  # # unit-race and itest jobs temporarily commented out to measure
```
Roasbeef: Let's not commit commented out config.

.github/workflows/main.yaml
```diff
@@ -227,7 +227,7 @@ jobs:
       - name: run golangci-lint
         uses: golangci/golangci-lint-action@v7
         with:
-          version: v2
+          version: v2.10.1
```
Roasbeef, line 223: Can fix up into prior commit?

.github/workflows/main.yaml
```diff
@@ -274,7 +274,7 @@ jobs:
         include:
           - unit_type: unit-cover
             parallel: 4
-          - unit_type: unit dbbackend=postgres
+          - unit_type: unit dbbackend=postgres pkg=tapdb
```
Roasbeef, line 270: So `tapgarden`, and `mssmt` import `tapdb`, and test against those extra set of backends.

.github/workflows/main.yaml
```diff
@@ -29,7 +29,7 @@ env:
   # Number of parallel unit-race jobs in the matrix and tranches per job.
   # Total tranches = NUM_UNIT_RACE_JOBS * UNIT_RACE_PARALLELISM.
   NUM_UNIT_RACE_JOBS: 4
-  UNIT_RACE_PARALLELISM: 2
+  UNIT_RACE_PARALLELISM: 1
```
Roasbeef, line 33: Revert the earlier commit that bumped this?

.github/workflows/main.yaml
jtobin: At present I believe the merge queue is used for anything landing in main, unless someone with exceptional perms (possibly only you) merges directly. But yeah the concern is valid in that the merge queue doesn't cover pushes to release branches, so we want to make sure the jobs are run there before pushing a tag.

Let me ruminate over it a bit more and see what I can come up with. Worst case we re-enable these; it's not the end of the world to do it on every push, though that frequency does seem unnecessary. Ideally we could strike a better balance.

.github/actions/cleanup-space/action.yaml
jtobin: I remember the space issue, but I don't think that could have been due to the cache growing without bound. Opus tells me that GitHub's cache is server-side (i.e. not on the runners themselves), is 10GB per repo, and has LRU eviction + a 7 day TTL on unused entries. As for the runners themselves, IIUC they're ephemeral, with no Docker cache (or really any state) that persists between runs.

*(fast forward five minutes -- TLDR, I'm pretty sure we can gut the cleanup-space stuff)*

I checked the previous PR's that touched this stuff (#1878 and #1889). I think the previous multi-platform release build required oodles of space, so the cleanup-space action, which reclaimed like ~28GB from every stock runner, addressed that, and there was some work done to ensure that artifacts from different platforms didn't mix in the build cache.

Since, though, the crosscomp jobs have adopted a matrix strategy, so we get a different platform per runner anyway. That should eliminate both of the space/cache mixing problems the other PRs addressed automatically: space, because we now only require enough space for a single platform's artifacts on the runner, and cache mixing, because the build caches are now isolated by platform.

.github/workflows/main.yaml
jtobin: See previous comment (IIRC 91 GB was measured from CI logs).

.github/workflows/main.yaml
jtobin: Still linting, but now via the golangci-lint action, which uses our config (and is significantly faster).

.github/workflows/main.yaml
jtobin: Nice catch. I think we're good to add tapgarden alone; the (blank) tapdb import in `test_tree.go` doesn't appear to do anything, as tapdb itself has no init().

jtobin: Fixed in eeffd82b65c95c83d622146f45eb5f4ad316c412.

.github/workflows/main.yaml
jtobin: Ok, how about this:

```
 on:
   push:
     branches:
       - 'main'
+      - '[0-9]*'
   pull_request:
     branches:
       - '*'
   merge_group:
     branches:
       - 'main'
```

The effect here would be that the heavier jobs would be skipped for routine pushes to PR branches, but the full suite would be run 1) in the merge queue, 2) on pushes to main, and 3) on all pushes to release branches.

Let me get an idea of the differential of just running everything always..

jtobin: Ok, looks like the balancing I did elsewhere actually brings the wall-clock time for the postgres itests down in turn too, so I don't think there's really any problem with enabling them for every push. Ditto for the Docker build check, which runs in ~8m, and the crosscomp jobs. We're still gated primarily by the unit-race jobs.

jtobin: All these have been re-enabled on every push in the updated changeset.

.github/workflows/main.yaml
jtobin: Fixed in updated changeset.

.github/workflows/main.yaml
jtobin: Fixed in updated changeset.

.github/workflows/main.yaml
jtobin: Fixed in updated changeset.

jtobin:
Note that I've compacted the commit set, and also tacked on some upgrades to deprecated action versions.

Roasbeef:
Updated the required checks!

PR #2009, opened 2026-02-26
Enhance `isAssetInvoice` check

GeorgeTsagk:
## Description

When a peer-accepted buy quote expires or the node restarts, `RfqPeerFromScid` could no longer resolve the SCID because it only checked the in-memory peerBuyQuotes map. This caused valid asset invoice HTLCs to be misclassified.

This PR persists peer-accepted buy quotes to the database and introduces a 3-tier SCID lookup in `Manager.LookUpScid`:
  1. Active in-memory map
  2. LRU cache (5120 entries)
  3. DB fallback

On startup, `restorePersistedPolicies` loads persisted peer buy quotes back into the active map so the full payment flow survives restarts.

Migration 053 extends the rfq_policies `CHECK` constraint to allow `RFQ_POLICY_TYPE_PEER_ACCEPTED_BUY` and adds a `FetchPeerAcceptedBuyPeerByScid` query with an scid index.

Closes #1844

PR #2019, opened 2026-03-09
tapdb: fix Postgres sequence collision in migration 053

jtobin:
Fixes the issue causing the postgres-backend unit tests job to fail on #2009.


GeorgeTsagk:
juuuust updated https://github.com/lightninglabs/taproot-assets/pull/2009 to include a fix for this migration. Our fix is pretty much identical, ty for posting!

Issue #1683, opened 2025-07-28
[cleanup]: implement auth mailbox message cleanup


guggero:
With https://github.com/lightninglabs/taproot-assets/pull/1614, we allow messages of up to 65k size to be stored on an auth mailbox server.
Currently there is no cleanup mechanism defined or implemented, so messages aren't deleted.

Because of the way we'll want to [implement backups](https://github.com/lightninglabs/taproot-assets/issues/426), we can't just remove the messages after a certain time or number of blocks.
Instead we should wait for the outpoint to be spent before deleting messages.

Steps to completion:
- [ ] Periodically (perhaps once a day in the background?) check all mailbox messages for their declared outpoint, if it was spent (using `bitcoind`'s `gettxout` method that returns `null` if a transaction output is no longer unspent). If it was spent, we can delete the message, as it is no longer relevant.
- [ ] Add a new `RemoveMessage` method to the `authmailboxrpc` package that allows the receiver to delete a message.
    - The receiver needs to provide a signature (for example over the claimed message outpoints to delete) to prove who they are (using the receiver key that is also used to _fetch_ the messages).
    - The `tapgarden.Custodian` should use that `RemoveMessage` RPC to remove processed messages. But only after all proofs were fetched correctly and the receive event is fully completed. And after implementing https://github.com/lightninglabs/taproot-assets/issues/426, we also want to make sure the new backup file is written before deleting any messages. Once we have the proof files and the proof's locators written to the backup file, we no longer need the auth mailbox message.

PR #2020, opened 2026-03-10
authmailbox: add message cleanup and RemoveMessage RPC

GeorgeTsagk:
## Summary

Closes https://github.com/lightninglabs/taproot-assets/issues/1683

- Add periodic spent-outpoint cleanup: the server checks all mailbox message outpoints in batches via `gettxout`, deleting messages whose outpoints have been spent on chain (configurable interval, default 24h)
- Add `RemoveMessage` RPC allowing receivers to delete their own messages with Schnorr signature authentication over `SHA256(receiver_id || msg_ids...)`
- Integrate `RemoveMessage` into the `tapgarden.Custodian` to clean up processed mailbox messages after successful proof receipt (best-effort)

## Test plan

- [x] Unit tests for cleanup logic (`TestCleanupSpentOutpoints`)
- [x] Unit tests for `RemoveMessage` server handler (valid removal, wrong receiver, non-existent IDs, empty request, invalid/bad-length signatures, idempotent removal, challenge consistency)
- [x] DB tests for `ListOutpoints`, `DeleteByOutpoint`, `DeleteByMessageID` with receiver ownership verification
- [x] Integration test for spent-outpoint cleanup (`testAuthMailboxCleanup`)
- [x] Integration test for `RemoveMessage` RPC (`testAuthMailboxRemoveMessage`)

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -81,6 +81,13 @@
 
 ## Functional Enhancements
 
+- [Auth Mailbox Cleanup](https://github.com/lightninglabs/taproot-assets/pull/2020):
+  The auth mailbox server now periodically checks claimed outpoints against the
+  chain and deletes messages whose outpoints have been spent. Receivers can also
+  explicitly remove their own messages via the new `RemoveMessage` RPC. The
+  custodian automatically removes mailbox messages after successfully receiving
+  proofs.
+
 - [Forwarding History Tracking](https://github.com/lightninglabs/taproot-assets/pull/1921):
   Routing nodes can now track and query historical asset forwarding events.
   When a node successfully routes an asset payment, the forward event is logged.
@@ -128,6 +135,10 @@
   Add `FetchAsset` RPC for fetching assets by asset ID or group key using an
   `AssetSpecifier`. Supports filters for `include_spent`, `include_leased`,
   `include_unconfirmed_mints`, `with_witness`, and `script_key_type`.
+
+- [PR#2020](https://github.com/lightninglabs/taproot-assets/pull/2020)
+  Add `RemoveMessage` RPC to the auth mailbox service. Receivers can
+  authenticate with a Schnorr signature to delete their own messages by ID.
 
 ## tapcli Additions
 
@@ -270,6 +281,11 @@
   Orphan UTXO sweeping is now enabled by default. This automatically sweeps
   tombstone and burn outputs when executing on-chain transactions. Set
   `wallet.disable-sweep-orphan-utxos` to disable.
+
+- [PR#2020](https://github.com/lightninglabs/taproot-assets/pull/2020)
+  Add `universe.mbox-cleanup-interval` and
+  `universe.mbox-cleanup-check-timeout` to configure periodic cleanup of
+  auth mailbox messages whose claimed outpoints have been spent on chain.
 
 ## Breaking Changes
 


tapgarden/custodian.go
```diff
@@ -1287,6 +1302,29 @@ func (c *Custodian) handleMailboxMessages(msgs *mbox.ReceivedMessages) error {
 	return nil
 }
 
+// removeMailboxMessages attempts to remove the specified messages from the
+// mailbox server. This is best-effort: errors are logged but do not propagate
+// to the caller, since the messages were already successfully processed.
+func (c *Custodian) removeMailboxMessages(serverURL url.URL,
```
darioAnongba, line 1313: nit: I think this function should return the error and whoever is using it should decide if the error logs warn (basically what to do with the error)

itest/authmailbox_cleanup_test.go
```diff
@@ -0,0 +1,138 @@
...
+		if info.MessageCount > 0 {
+			return fmt.Errorf("expected 0 messages, got %d",
+				info.MessageCount)
+		}
+
+		return nil
+	}, defaultTimeout)
+	require.NoError(t.t, err)
+
+	t.Logf("Mailbox cleanup successful, all messages deleted")
```
darioAnongba: This kind of logs are useless IMO since the test is passing.

authmailbox/server.go
```diff
@@ -716,6 +751,78 @@ func (s *Server) RemoveSubscriber(
...
+	}
+
+	ctx = btclog.WithCtx(
+		ctx, btclog.Hex("receiver_id", req.ReceiverId),
+		"server", true,
+	)
+	log.DebugS(ctx, "Received RemoveMessage request",
+		"num_ids", len(req.MessageIds))
+
+	if len(req.MessageIds) == 0 {
```
darioAnongba, line 781: we very much need to bound this like:
```
const maxRemoveMessageIDs = 1000
```
otherwise someone can DoS us and send millions of IDs that trigger DB txs?

authmailbox/server.go
```diff
@@ -729,3 +836,140 @@ func concatAndHash(a, b []byte) [32]byte {
...
+		}
+
+		if len(outpoints) == 0 {
+			break
+		}
+
+		// Check all outpoints in this batch concurrently.
+		results := make(chan spentResult, len(outpoints))
+		for _, op := range outpoints {
+			go func() {
```
darioAnongba, line 921: Same comment as in a previous PR, this needs a bounded semaphore like `maxConcurrentChecks = 50`?

sample-tapd.conf
```diff
@@ -380,6 +380,14 @@
 ; units are {s, m, h}.
 ; universe.mbox-auth-timeout=10s
 
+; How often to check for spent outpoints and delete stale mailbox messages. Set
+; to 0 to disable cleanup. Valid time units are {s, m, h}.
+; universe.mbox-cleanup-interval=0s
```
darioAnongba: Isn't it better to enable by default? breaking change is fine by me. I see 24h, 30s but here it says 0s? Not sure if disabled or enabled by default.

tapcfg/server.go
```diff
@@ -946,6 +953,37 @@ func CreateServerFromConfig(cfg *Config, cfgLogger btclog.Logger,
...
+// mboxOutpointChecker returns an OutpointChecker that uses lnd's chain
+// notifier to determine whether an outpoint has been spent on chain.
+func mboxOutpointChecker(
+	lndServices *lndclient.LndServices) authmailbox.OutpointChecker {
+
+	return func(ctx context.Context, op wire.OutPoint,
+		pkScript []byte, heightHint uint32) (bool, error) {
+
+		spendChan, errChan, err :=
+			lndServices.ChainNotifier.RegisterSpendNtfn(
```
darioAnongba, line 982: Did you mean to use `RegisterSpendNtfn`? You mention using `gettxout`. Not sure if `RegisterSpendNtfn` is appropriate since it's more for long-lived subscriptions?

sample-tapd.conf
jtobin: An agent also flagged that cleanup actually cannot be disabled in this manner; worth checking on:

> `sample-tapd.conf` and `tapcfg/config.go:353` document that `--universe.mbox-cleanup-interval=0s` disables cleanup.  In practice this doesn't work: `genServerConfig` unconditionally sets `OutpointChecker` (`tapcfg/server.go:861`), and the server starts the cleanup loop whenever `OutpointChecker != nil` (`authmailbox/server.go:134`).  Inside `cleanupLoop`, `cmp.Or(s.cfg.CleanupInterval, defaultCleanupInterval)` treats the zero value as "use default 24h" (`authmailbox/server.go:850-852`).  The documented disable path is dead code.

authmailbox/server.go
```diff
@@ -729,3 +836,140 @@ func concatAndHash(a, b []byte) [32]byte {
...
+	)
+
+	var (
+		totalChecked int
+		totalDeleted int
+		offset       int32
+	)
+
+	for {
+		ctx, cancel := context.WithCancel(context.Background())
```
jtobin: An agent recommended employing `s.WithCtxQuit()` here (and elsewhere in this function; L907, L939) to use a context that cancels when the server shuts down, rather than the empty context.Background(). That makes sense to me, probably a change worth making.

tapcfg/server.go
GeorgeTsagk: `gettxout` would be a very suiting bitcoin backend method, but unfortunately it's not exposed via LND. Main reason there is that neutrino doesn't support that method.

`RegisterSpendNtfn` is intentional. We rely on its behavior to return immediately when the outpoint is spent.

PR #2021, opened 2026-03-11
itest: forwarding history

darioAnongba:
Move the forwarding history integration test from the standalone rfq_forwards_test.go (which used mocks) into the custom channels.

### DIFF
--- a/itest/custom_channels/forwarding_history_test.go
+++ b/itest/custom_channels/forwarding_history_test.go
@@ -149,6 +149,10 @@ func testCustomChannelsForwardingHistory(ctx context.Context,
 	require.GreaterOrEqual(
 		t.t, daveFwd.AssetAmt, uint64(invoiceAssetAmount),
 	)
+	// AmtOutMsat (the real outgoing BTC amount) should be greater than
+	// AmtInMsat (the incoming asset anchor amount of ~354 sats), since
+	// Dave converts from the smaller asset-anchor value to the actual
+	// invoice amount in sats.
 	require.Greater(t.t, daveFwd.AmtOutMsat, daveFwd.AmtInMsat)
 	require.GreaterOrEqual(t.t, daveFwd.SettledAt, daveFwd.OpenedAt)
 	require.Zero(t.t, daveFwd.FailedAt)


jtobin (commented):
Looks good; one issue that agents flagged was that there's a loss of assertion precision in the new version. E.g.:

```
  ┌──────────────────┬─────────────────────┬──────────┐
  │      Field       │      Old test       │ New test │
  ├──────────────────┼─────────────────────┼──────────┤
  │ RfqId            │ == acceptedQuote.Id │ NotEmpty │
  ├──────────────────┼─────────────────────┼──────────┤
  │ AssetAmt         │ == 42               │ > 0      │
  ├──────────────────┼─────────────────────┼──────────┤
  │ AmtInMsat        │ == askAmt + 1000    │ > 0      │
  ├──────────────────┼─────────────────────┼──────────┤
  │ AmtOutMsat       │ == askAmt           │ > 0      │
  ├──────────────────┼─────────────────────┼──────────┤
  │ Rate.Coefficient │ == "1101000"        │ NotEmpty │
  └──────────────────┴─────────────────────┴──────────┘

  The > 0 checks would pass even if amounts were wildly wrong.
```

Any specific rationale for this, or is it just a refactoring artifact? Seems like the additional precision might be strictly better?

itest/custom_channels/forwarding_history_test.go
```diff
@@ -0,0 +1,320 @@
...
+		t.t, rfqrpc.RfqPolicyType_RFQ_POLICY_TYPE_PURCHASE,
+		daveFwd.PolicyType,
+	)
+	require.Equal(t.t, charlie.PubKeyStr, daveFwd.Peer)
+	require.NotNil(t.t, daveFwd.AssetSpec)
+	require.Equal(t.t, assetID, daveFwd.AssetSpec.Id)
+	require.GreaterOrEqual(
+		t.t, daveFwd.AssetAmt, uint64(invoiceAssetAmount),
+	)
+	require.Greater(t.t, daveFwd.AmtOutMsat, daveFwd.AmtInMsat)
```
GeorgeTsagk, line 156: Might be worth mentioning here that the assertion compares the incoming asset anchor (~354sats) vs the real outgoing btc amount

PR #2031, opened 2026-03-17
Backward compatibility testing infrastructure

### LINKED ISSUE #2027: Backwards compatibility validation
The custom channels integration tests were migrated from lightning-terminal to taproot-assets, but the backward compatibility testing framework was not ported. We need infrastructure to verify that the current `tapd` binary remains compatible with older versions.

darioAnongba:
- **Wire format fixture tests (per-PR):** Committed v0.8 wire format fixtures for `rfqmsg` (request, accept, reject, HTLC) and `tapchannelmsg` (funding blob, commitment blob, HTLC blob). Backward compat tests decode each fixture with current code and verify round-trip encoding on every PR.
- **Mixed-version integration test harness:** Added `NewNodeWithBinary` and `UpgradeNode` to `IntegratedNetworkHarness` so tests can start nodes on older binaries and simulate upgrades mid-operation.
- **Build script for old binaries:** `scripts/build-compat-binary.sh` builds `tapd-integrated` from any git tag using a temporary worktree and caches the result in `~/.tapd-compat-bins/`.
- **Compat integration test wrapper:** `TestCustomChannelsCompat` runs core, force close, and v1 upgrade tests against each configured historical version. The `compatVersions` list is empty until v0.8.0 is tagged.
- **Makefile targets + CI:** `make itest-cc-compat`, `make build-compat-binary`, `make gen-compat-fixtures`. GitHub Actions workflow triggers on `release/*` pushes and `workflow_dispatch`.

### Design

Two layers catch different classes of regressions:

| Layer | Runs on | Catches | Cost |
|-------|---------|---------|------|
| Fixture decode tests | Every PR | Encoding/decoding wire format breaks | ~seconds |
| Mixed-version itests | Release branches + manual | Runtime/interaction protocol bugs | ~15 min |

Bitcoin Core takes a similar approach (`feature_backwards_compatibility.py` on demand, not every PR) but their protocol evolves much more slowly.

Closes #2027


jtobin:
Closing as obviated by #2027.

Issue #2038, opened 2026-03-23
[bug]: `DecodeAssetPayReq` should populate genesis info when queried by group key


darioAnongba:
When `DecodeAssetPayReq` is called with a `group_key` (the correct way to identify a fungible asset), the response omits `GenesisInfo` entirely. The current code explicitly skips it with this comment:

```go
// The genesis info makes no sense in the case where we have decoded the
// invoice for a group key, since we just pick the first asset ID we
// found for a group key.
```

This means callers that use `group_key`, which is the recommended identifier for fungible assets, cannot display the asset name, type, or meta hash. This is the case for `litcli ln payinvoice --group_key`, which shows a blank asset name even though the information is available.

### Why this is wrong

The premise that "genesis info makes no sense for a group key" conflates two things:

1. **Per-tranche identity** (asset ID, genesis point), differs across tranches and is indeed meaningless at the group level.
2. **Asset-level metadata** (name, type, meta hash), identical across all tranches of a fungible asset and is meaningful at the group level.

A group key **is** the canonical identity of a fungible asset. Individual asset IDs (tranches) are an internal implementation detail that should not be exposed to end users. Users don't choose which tranche they spend. The asset name, type, and decimal display are properties of the *group*, not of any individual tranche.

### Current data flow

The `AssetGroup` Go struct already embeds `*Genesis` (the group anchor's genesis):

```go
type AssetGroup struct {
    *Genesis
    *GroupKey
}
```

When resolving by group key, `fetchGroupByGroupKey` fetches the anchor genesis from the DB via `GenAssetID`. So the genesis data is already available in `assetGroup.Genesis`, it's just not being marshaled into the RPC response.

### Proposed fix

When `DecodeAssetPayReq` is called with a `group_key`, populate `GenesisInfo` using the group anchor's genesis. The name, asset type, and meta hash are all valid and meaningful. For the `asset_id` field within `GenesisInfo`, use the anchor asset's ID (or consider omitting it / adding a separate field to clarify it's the anchor ID).

Alternatively, consider adding a group-level metadata field to `AssetGroup` in the proto definition so that asset name and meta are accessible without relying on the genesis of a specific tranche. This would make the data model clearer: genesis info describes a specific tranche, while group-level metadata describes the fungible asset as a whole.

PR #2039, opened 2026-03-24
Fixing populating genesis info

sergey3bv:
Should solve #2038

### DIFF
--- a/itest/custom_channels/decode_invoice_test.go
+++ b/itest/custom_channels/decode_invoice_test.go
@@ -167,4 +167,44 @@ func testCustomChannelsDecodeAssetInvoice(_ context.Context,
 	require.Equal(
 		t.t, decodeResp.DecimalDisplay, decodeResp2.DecimalDisplay,
 	)
+
+	// GenesisInfo must be populated for group key queries. For a
+	// single-tranche group the anchor asset ID equals the leaf ID, so
+	// the full struct must match the asset-ID-based response.
+	require.Equal(t.t, decodeResp.GenesisInfo, decodeResp2.GenesisInfo)
+
+	// Mint a second tranche in the same group and assert decoding by group
+	// key still reports the anchor tranche genesis ID.
+	secondTranche := &mintrpc.MintAsset{
+		AssetType:    taprpc.AssetType_NORMAL,
+		Name:         "USD-second-tranche",
+		AssetMeta:    usdMetaData,
+		Amount:       100_000_000_000,
+		DecimalDisplay: decimalDisplay,
+		GroupedAsset: true,
+		GroupKey:     usdAsset.AssetGroup.TweakedGroupKey,
+	}
+	secondMinted := itest.MintAssetsConfirmBatch(
+		t.t, net.Miner.Client, asTapd(alice),
+		[]*mintrpc.MintAssetRequest{
+			{
+				Asset: secondTranche,
+			},
+		},
+	)
+	require.Len(t.t, secondMinted, 1)
+
+	secondTrancheID := secondMinted[0].AssetGenesis.AssetId
+	require.NotEqual(t.t, usdAsset.AssetGenesis.AssetId, secondTrancheID)
+
+	decodeResp3, err := asTapd(alice).DecodeAssetPayReq(
+		ctx, &tchrpc.AssetPayReq{
+			GroupKey:     usdAsset.AssetGroup.TweakedGroupKey,
+			PayReqString: payReq,
+		},
+	)
+	require.NoError(t.t, err)
+	require.Equal(
+		t.t, usdAsset.AssetGenesis.AssetId, decodeResp3.GenesisInfo.AssetId,
+	)
 }
--- a/itest/custom_channels/helpers.go
+++ b/itest/custom_channels/helpers.go
@@ -2390,14 +2390,29 @@ func assertChannelAssetBalanceWithDelta(t *testing.T,
 	local, remote uint64, delta float64) {
 
 	targetChan := fetchChannel(t, node, chanPoint)
-
-	assetBalance, err := parseChannelData(targetChan.CustomChannelData)
+	customData := targetChan.CustomChannelData
+	assetBalance, err := parseChannelData(customData)
 	require.NoError(t, err)
 
 	require.Len(t, assetBalance.FundingAssets, 1)
 
-	require.InDelta(t, local, assetBalance.LocalBalance, delta)
-	require.InDelta(t, remote, assetBalance.RemoteBalance, delta)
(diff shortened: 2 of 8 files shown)


sergey3bv:
Hey, @darioAnongba, could you please review?

cc @Roasbeef

rpcserver/rpcserver.go
```diff
@@ -10566,17 +10566,21 @@ func (r *RPCServer) DecodeAssetPayReq(ctx context.Context,
...
-	// found for a group key.
-	if len(payReq.GroupKey) == 0 {
-		resp.GenesisInfo = &taprpc.GenesisInfo{
-			GenesisPoint: assetGroup.FirstPrevOut.String(),
-			AssetType:    taprpc.AssetType(assetGroup.Type),
-			Name:         assetGroup.Tag,
-			MetaHash:     assetGroup.MetaHash[:],
-			AssetId:      assetID[:],
-		}
+	// When the request was made with a group key, use the anchor
```
darioAnongba: The comment states: *"use the anchor tranche's genesis ID rather than the arbitrary leaf ID resolved by syncAssetGroup."* However, `syncAssetGroup` returns an arbitrary leaf ID, then `QueryAssetInfo` is called with that same ID. The resulting `assetGroup.Genesis.ID()` equals that same arbitrary leaf's genesis ID, not the group anchor's genesis. For multi-tranche groups (most fungible assets), the `AssetId` and `GenesisPoint` in the response could vary depending on which tranche the database returns first.

Use `fetchGroupByGroupKey` (which fetches the genesis tied to the group key via `GenAssetID`) instead of relying on the arbitrary leaf from `syncAssetGroup`.

itest/custom_channels/decode_invoice_test.go
```diff
@@ -167,4 +167,9 @@ func testCustomChannelsDecodeAssetInvoice(_ context.Context,
 	require.Equal(
 		t.t, decodeResp.DecimalDisplay, decodeResp2.DecimalDisplay,
 	)
+
+	// GenesisInfo must be populated for group key queries. For a
+	// single-tranche group the anchor asset ID equals the leaf ID, so
+	// the full struct must match the asset-ID-based response.
+	require.Equal(t.t, decodeResp.GenesisInfo, decodeResp2.GenesisInfo)
```
darioAnongba, line 174: We should now add multi-tranche test coverage. Add a test that mints a second tranche in the same group, then calls `DecodeAssetPayReq` with the group key and verifies that `GenesisInfo.AssetId` equals the expected anchor tranche's ID.

itest/custom_channels/helpers.go
```diff
@@ -2535,9 +2535,13 @@ func locateAssetTransfers(t *testing.T, node *itest.IntegratedNode,
 
 		transfer = forceCloseTransfer.Transfers[0]
 
-		if transfer.AnchorTxBlockHash == nil {
-			return fmt.Errorf("missing anchor block hash, " +
-				"transfer not confirmed")
+		if transfer.AnchorTxBlockHeight == 0 &&
```
darioAnongba: Good but unrelated with the PR.

itest/custom_channels/liquidity_test.go
```diff
@@ -686,21 +687,46 @@ func testCustomChannelsLiquidityEdgeCasesCore(ctx context.Context,
...
 	// We now create an invoice on Fabia and expect Erin's policy to be used
 	// in the invoice.
-	invoiceResp = createAssetInvoice(t.t, erin, fabia, 1_000, assetID)
-	req, err := erin.LightningClient.DecodePayReq(
-		ctx, &lnrpc.PayReqString{
-			PayReq: invoiceResp.PaymentRequest,
-		},
-	)
-	require.NoError(t.t, err)
+	err = wait.NoError(func() error {
```
darioAnongba: Good but unrelated with the PR.

rpcserver/rpcserver.go
```diff
@@ -10532,13 +10544,27 @@ func (r *RPCServer) DecodeAssetPayReq(ctx context.Context,
 				"key: %w", err)
 		}
 
-		assetID, err = r.syncAssetGroup(ctx, *groupKey)
```
darioAnongba, line 10535: Looks like `syncAssetGroup` is not used anywhere anymore? Can you remove it in this PR if this is true?

itest/custom_channels/liquidity_test.go
```diff
@@ -686,21 +687,47 @@ func testCustomChannelsLiquidityEdgeCasesCore(ctx context.Context,
...
 	// in the invoice.
-	invoiceResp = createAssetInvoice(t.t, erin, fabia, 1_000, assetID)
-	req, err := erin.LightningClient.DecodePayReq(
-		ctx, &lnrpc.PayReqString{
-			PayReq: invoiceResp.PaymentRequest,
-		},
-	)
-	require.NoError(t.t, err)
+	var req *lnrpc.PayReq
+	err = wait.NoError(func() error {
```
darioAnongba: Similar than previous comment, there is no reason changing the behavior here and use `wait.NoError` because `createAssetInvoice` uses `require` so this will always exit on first error anyways and never loop.

itest/custom_channels/helpers.go
```diff
@@ -2388,16 +2388,50 @@ func assertChannelAssetBalance(t *testing.T, node *itest.IntegratedNode,
 func assertChannelAssetBalanceWithDelta(t *testing.T,
 	node *itest.IntegratedNode, chanPoint *lnrpc.ChannelPoint,
 	local, remote uint64, delta float64) {
+	var (
+		lastLocal  uint64
+		lastRemote uint64
+	)
 
-	targetChan := fetchChannel(t, node, chanPoint)
+	err := wait.NoError(func() error {
```
darioAnongba: Similar than previous comment, there is no reason changing the behavior here and use `wait.NoError` because `fetchChannel` uses `require` so this will always exit on first error anyways and never loop.

rpcserver/rpcserver.go
```diff
@@ -10512,15 +10516,23 @@ func (r *RPCServer) DecodeAssetPayReq(ctx context.Context,
...
-	// group, then fetch the leaves that are associated with this group key.
-	// From there, we can look up the asset ID of one of the group's
-	// tranches.
+	// For group key queries, fetch asset info by group key directly so the
+	// returned genesis always maps to the group's anchor tranche.
 	case len(payReq.GroupKey) != 0:
 		var (
 			groupKey *btcec.PublicKey
-			err      error
 		)
```
darioAnongba: `var groupKey *btcec.PublicKey`

PR #2043, opened 2026-03-27
itest: add passive assets custom channel test

jtobin:
Like #2042, this simply ports the passive assets / capacity assertion itest added in lightninglabs/lightning-terminal#1198 to our native cc-itest framework.


jtobin:
> LGTM
> 
> Should we also verify that the passive asset is carried over in a healthy way? (list assets or smth similar). Non-blocking, as that's how it was in the original litd test.

Yeah I vote it to be fine: the intent was to ensure that passive assets don't pollute channel capacity, and it fits the bill for that.

PR #2048, opened 2026-04-03
rfq: add limit constraints and quote-level enforcement to rfq negotiation

### LINKED ISSUE #2004: [feature]: RFQ: First-Class Limit Orders In Quote Negotiation
## Summary
Today, RFQ orders are mostly constrained by:
- Requested max size (`asset_max_amt` or `payment_max_amt`)
- Quote expiry
- Channel availability and balance shape
- Price-oracle tolerance checks

This is sufficient for basic RFQ flows, but it is not a full limit-order model. We should add explicit limit-order semantics (price bounds, fill constraints, and execution policy) to quote negotiation.

## Why This Matters
Limit-order semantics enable:
- Price protection: avoid execution outside user-defined acceptable price bounds.
- Slippage control: reduce risk from stale hints, spread moves, and short-term volatility.
- Deterministic UX: users can express explicit execution intent instead of best-effort behavior.
- Safer automation: wallet agents can auto-negotiate with tighter risk controls.
- Better policy explainability: clearer machine-readable reject reasons.
- Better multi-peer execution: easier to split across peers while preserving user constraints.

## Terminology
- `IOC` (`Immediate-Or-Cancel`): execute immediately for any amount that satisfies constraints; cancel any unfilled remainder.
- `FOK` (`Fill-Or-Kill`): execute immediately and fully, or cancel the entire order.

### What RFQ can express today
- Buy flow: max asset amount (`AssetMaxAmt`) + expiry + peer + metadata.
- Sell flow: max payment amount (`PaymentMaxAmt`) + expiry + peer + metadata.
- Request/accept negotiation primarily settles on a rate + expiry.

### What is enforced in practice
- Requester-side quote validation checks include quote expiry bounds and oracle tolerance against the accepted rate.
- HTLC/policy enforcement is max-cap based, where sale path policy derives from `AssetMaxAmt` and purchase path policy derives from `PaymentMaxAmt`.
- Liquidity constraints are effectively channel-driven: channel existence checks gate order submission, and channel balances influence selection/sorting.

### Existing but underused protocol surface
- Wire format already includes optional `MinInAsset` and `MinOutAsset` fields.
- These min fields are not first-class in typed request structs and RPC order APIs.

### Current gap
Current behavior is closer to:
- "Market quote with max-size cap + expiry"
not:
- "Limit order with explicit price/fill constraints"

Users cannot directly encode:
- Minimum acceptable price / rate.
- Maximum acceptable price / rate.
- Minimum fill size.
- Execution policy (`IOC` vs `FOK`).

### Workstream A: Add explicit limit constraints to request surfaces
- Add optional limit-price/rate bound fields to buy/sell order APIs.
- Add optional fill bound fields (`min_in_asset`, `min_out_asset`) to typed requests and RPC.
- Keep current max fields as-is for compatibility.
- Define a canonical rate-bound representation that avoids buy/sell direction ambiguity.

### Workstream B: Enforce limits during negotiation
- Responder side (`ResolveRequest`): reject when request limits cannot be satisfied.
- Requester side (`VerifyAcceptQuote`): enforce explicit requester limits in addition to oracle tolerance.
- Return explicit reject reasons for limit failures (price bound miss, min fill miss, insufficient liquidity).

### Workstream C: Add execution policy semantics
- Support explicit execution policy on orders.
- `IOC`: allow partial immediate execution above minimum fill.
- `FOK`: require full immediate execution or reject.
- Ensure policy is respected consistently in quote acceptance and HTLC policy installation.

### Workstream D: Negotiate explicit fill quantities
- Add accepted fill quantity semantics to quote acceptance path.
- Bind policy persistence and HTLC enforcement to negotiated fill values, not only request max caps.
- This removes ambiguity around partial fills and improves deterministic execution.

### Workstream E: Multi-peer selection that respects limits
- When multiple peers are available, rank and split using effective rate and usable capacity.
- Ensure every shard respects the same order limits and execution policy.

## Non-Goals
- Building a centralized order book.
- Long-lived resting/GTC orders in initial implementation.
- Redesigning the external price-oracle architecture.

## Compatibility Expectations
- New fields should be additive and optional.
- Old peers should continue to interoperate under current behavior.
- Limit-specific validation should only activate when limit fields are present.
- Use versioning only where strict incompatibility requires it.

## Acceptance Criteria
- Users can specify explicit limit-price bounds for buy and sell RFQ orders.
- Users can specify minimum fill constraints.
- Users can choose execution policy (`IOC` or `FOK`).
- Quotes violating constraints are rejected with clear machine-readable reasons.
- Accepted quotes cannot pass validation if they violate requester limits.
- Policy/HTLC enforcement is tied to negotiated constraints, not only broad request max caps.
- Test coverage includes buy and sell flows.
- Test coverage includes `IOC` and `FOK` behaviors.
- Test coverage includes backward compatibility with old peers.
- Test coverage includes volatility and timing edge cases.

## Open Questions
- What is the canonical limit-price representation to avoid buy/sell interpretation errors?
- Should both min and max rate bounds be supported initially, or only one directional bound per order type?
- Do we extend `QuoteRespStatus`, add new reject codes, or both?
- Is explicit fill quantity required in the first delivery, or can it follow once request-side limits are live?
- Should passive offer filtering and limit-order checks be merged into one decision pipeline?

jtobin:
Partially resolves #2004 (roughly, workstreams A and B).

Adds limit order constraint fields to the RFQ wire protocol and RPC surface. Opus-generated TLDR for these changes:

* Users can specify explicit limit-price bounds for buy and sell RFQ orders
* Users can specify minimum fill constraints
* Quotes violating constraints are rejected with machine-readable reasons (enforcement in this changeset is only at the quote acceptance level)
* Accepted quotes cannot pass validation if they violate requester limits
* Backward-compatible; new fields are optional TLVs, and validation is gated on presence.

For illustration of the kind of thing this enables, here's an example scenario from an e2e regtest demo of this (and related) functionality:

```
  Frank buys 1,000 USDX ($10.00) with a price ceiling at ~105% of market
  — he won't pay more than that per USDX. The pilot's price is below the
  ceiling, so the quote is accepted and payment executes through the asset
  channel.

  Price ceiling: 15.7 sats/USDX
  Pilot price:   14.9 sats/USDX
  ✓ Accepted
    RFQ ID: klOI8CDwIdlt...
    Rate:   $66,805.00/BTC
    Amount: 1000 USDX
    Limit:  rate_floor coeff=6346475 ($63,464.75/BTC)

  → AddInvoice (rfq_id: klOI8CDwIdlt...)
    1000 USDX from Frank, reusing negotiated quote
    Quote confirmed: klOI8CDwIdlt...
    Invoice: lnbcrt149689390p1p5ulzhwpp5dp476qtl95kzq...

  → Payment: Alice → Eve → Frank
    ✓ Payment succeeded
```

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -103,6 +103,14 @@
   **Note:** For full functionality, it is highly recommended to start LND with
   the `--store-final-htlc-resolutions` flag enabled, which is disabled by default.
 
+- [Limit-Order Constraints](https://github.com/lightninglabs/taproot-assets/pull/2048):
+  RFQ buy and sell orders can now carry explicit limit-price bounds
+  (`asset_rate_limit`) and minimum fill sizes (`asset_min_amt` /
+  `payment_min_amt`). Quotes that violate these constraints are rejected
+  with machine-readable reasons (`RATE_BOUND_MISS`, `MIN_FILL_NOT_MET`).
+  New fields are optional and backward-compatible; constraint validation
+  only activates when they are present.
+
 ## Functional Enhancements
 
 - [Wallet Backup/Restore](https://github.com/lightninglabs/taproot-assets/pull/1980):
@@ -147,6 +155,13 @@
 - [PR#2020](https://github.com/lightninglabs/taproot-assets/pull/2020)
   Add `RemoveMessage` RPC to the auth mailbox service. Receivers can
   authenticate with a Schnorr signature to delete their own messages by ID.
+
+- [PR#2048](https://github.com/lightninglabs/taproot-assets/pull/2048):
+  Add `asset_rate_limit` to `AddAssetBuyOrder` and `AddAssetSellOrder`
+  requests. Add `asset_min_amt` to buy orders and `payment_min_amt`
+  to sell orders. Add `asset_rate_limit` and min fill fields to
+  `PortfolioPilot.ResolveRequest` for constraint forwarding. Add
+  `RATE_BOUND_MISS` and `MIN_FILL_NOT_MET` to `QuoteRespStatus`.
 
 ## tapcli Additions
 
@@ -330,6 +345,10 @@
 - [Forwarding History Integration Test](https://github.com/lightninglabs/taproot-assets/pull/1921):
   New integration test `testForwardingEventHistory` verifies that forwarding events are
   properly logged when routing asset payments.
+
+- [PR#2048](https://github.com/lightninglabs/taproot-assets/pull/2048):
+  Add unit, property-based, and integration tests for limit-order
+  constraint fields.
 
 ## Database
 


docs/release-notes/release-notes-0.8.0.md
```diff
@@ -331,6 +346,10 @@
   New integration test `testForwardingEventHistory` verifies that forwarding events are
   properly logged when routing asset payments.
 
+- [PR#2048](https://github.com/lightninglabs/taproot-assets/pull/2048):
```
darioAnongba, line 349: I still don't see what we have a "test" section in the changelog. 🤔

rfq/portfolio_pilot.go
darioAnongba, line 284: This function is specific to the `InternalPortfolioPilot` but it's not clear to me what happens when a user uses its own. I guess it's either:

1. we don't care and let the external portfolio pilot handle things, if it doesn't support the new limit orders, it's not our concern?
2. Execute validation (`checkRateBound` and `checkMinFill`) regardless of which portfolio pilot is used, in `RpcPortfolioPilot.VerifyAcceptQuote`?

I think 1 makes sense but the user might pass these params while they are just silently dropped/ignored. Sadly that is an anti-pattern because if the params are accepted by our service, then they shouldn't be silently ignored.

rfqmsg/request.go
```diff
@@ -201,6 +215,23 @@ func newRequestWireMsgDataFromBuy(q BuyRequest) (requestWireMsgData, error) {
 		)
 	}
 
+	// Set optional min asset amount.
```
darioAnongba, line 218: Good point found by a security agent I ran:
```
The `AssetRateLimit` and min amount fields are serialized into wire messages sent to the counterparty peer. While enforcement is requester-side only, transmitting limits reveals the requester's exact price bounds. A rational counterparty can extract maximum value by offering exactly the worst-acceptable price.
```

WDYT? does the counterparty need this information? I guess it's performance vs transparency tradeoff. With this info the counterparty can always maximize the value but without it, we add more round-trips because the counterparty doesn't know if the quote will be accepted.

taprpc/rfqrpc/rfq.proto
darioAnongba, line 368: Missing enums here?
`MIN_FILL_NOT_MET = 5` and `RATE_BOUND_MISS= 6`

rfqmsg/sell_request.go
darioAnongba:
```suggestion
			"key cannot both be unset for incoming sell " +
```

darioAnongba:
```suggestion
	// group key is set. At least one must be set in a sell request.
```

docs/release-notes/release-notes-0.8.0.md
jtobin: +1. I kept these notes in there for now, but IMO we should cut the section entirely (maybe even before v0.8.0).

rfq/portfolio_pilot.go
jtobin: > we don't care and let the external portfolio pilot handle things, if it doesn't support the new limit orders, it's not our concern?

IMO (1) is right, yeah. I think you want pretty much maximum flexibility when you bring your own pilot; you're passed quote/order information and it's up to you to decide what to do with it. Indeed, it may be perfectly reasonable to silently drop/ignore stuff in an external pilot, *e.g.* if you just don't want to deal with limit order constraints for some reason (it's your portfolio pilot, after all).

rfqmsg/request.go
jtobin: Yeah I think this is indeed what we want here, and is inescapable if trade is between two parties if we're purely communicating constraints. If I say e.g. "I will pay at most 100k sats" to my (single) counterparty, then the order will definitely execute at 100k sats, or not at all, so long as the counterparty is rational. The story would actually be different with multiple parties (akin to a normal, public order book), because some could compete to accept at e.g. 95k.

The alternative would be to not actually communicate limit order constraints, but instead engage in a repeated game of sorts, as you mention re: round trips during negotiation (I guess right now one could actually just do this by e.g. iterating progressively higher max prices during negotiation anyway; the crux is that it seems to be a different issue).

taprpc/rfqrpc/rfq.proto
jtobin: Yeah, these should be here. They are actually added downstream in 0b222f62, in another PR'd branch, along with an execution policy enum. I'll port them to here.

rfqmsg/sell_request.go
jtobin: Derp, will fix..

jtobin:
> Currently we don't set any limits by default so all new fields are "accept any amount/rate". Most likely for backwards compatibility? Does some security on this make sense? The oracle offers some safety but I don't think we should rely on it too much.
> 
> Might as well be fine because I think that's how exchanges work? When setting no limits, it's essentially a "market" order (buy at any market price).

That's exactly right. Without limit constraints set we just fall back to the existing behaviour, which has market order semantics. The pilot can indeed provide security for market orders too, e.g. by rejecting based on exceeding tolerance from its view of market price or whatever (this is what the default/internal version does).

rfqmsg/sell_request.go
jtobin: Fixed in latest changest.

jtobin: Fixed in latest changeset.

taprpc/rfqrpc/rfq.proto
jtobin: Fixed in latest changeset.

rfqmsg/request.go
darioAnongba: > The alternative would be to not actually communicate limit order constraints, but instead engage in a repeated game of sorts, as you mention re: round trips during negotiation (I guess right now one could actually just do this by e.g. iterating progressively higher max prices during negotiation anyway; the crux is that it seems to be a different issue).

Might make sense to discuss this quickly before merging.
cc: @GeorgeTsagk

GeorgeTsagk: I was also thinking about this while reading the PR: if you disclose "I am willing to go as far as X" then the counter-party will always meet you at X.

There is a subtle benefit to not disclosing your limits, as they can be gamed.

GeorgeTsagk:
> Currently we don't set any limits by default so all new fields are "accept any amount/rate". Most likely for backwards compatibility? Does some security on this make sense? The oracle offers some safety but I don't think we should rely on it too much.

For the local<>oracle communication I can't think of any security measure. You are meant to trust your oracle, or at least use multiple independent ones when in doubt.

Even before these additions we were able to guard against deviations with the old tolerance check (a const ppm value that defaulted to 5%).

PR #2049, opened 2026-04-03
rfqmsg+rfq: add execution policy to rfq negotiation

jtobin:
Depends on #2048. Partially resolves #2004 (most, though not all, of workstream C).

Adds execution policy semantics (immediate-or-cancel, fill-or-kill) to the RFQ wire protocol and RPC surface, which again, as in #2048, are here enforced at the quote acceptance level. This PR sets up the wire encoding, RPC surface, and validation plumbing for the feature, but nontrivial enforcement is deferred to #2050 (where we have fill quantities to work with).

Feature TLDR:

- Users can choose either an IOC or FOK policy (the former accepts partial fills, the latter requires a complete fill)
- Quotes violating FOK are rejected with a machine-readable reason (FOK_NOT_VIABLE).
- Backward-compatible: field is optional, default is IOC.

Like the functionality added in #2048, the execution policy field is additive and optional; old peers that ignore the new TLV record continue to interoperate.

Here's an illustrative use of the feature, from an e2e regtest demo:

```
Frank buys 1,000 USDX with Fill-or-Kill. The full amount converts
cleanly at the current rate — quote accepted, payment executes.

Market rate: $66,805.00/BTC
✓ Accepted
  RFQ ID: wMzJgyYBcWwK...
  SCID:   17614968261059438554
  Rate:   $66,805.00/BTC
  Amount: 1000 USDX
  Policy: EXECUTION_POLICY_FOK

→ AddInvoice (rfq_id: wMzJgyYBcWwK...)
  1000 USDX from Frank, reusing negotiated quote
  Quote confirmed: wMzJgyYBcWwK...
  Invoice: lnbcrt149689390p1p5ulzhnpp5zkkknghqfhr2l...

→ Payment: Alice → Eve → Frank
  ✓ Payment succeeded
```

PR #2050, opened 2026-04-06
rfq+rfqmsg: add fill quantity negotiation and constraint validation

### LINKED ISSUE #2004: [feature]: RFQ: First-Class Limit Orders In Quote Negotiation

jtobin:
Depends on #2049. Partially resolves #2004 (remainder of workstream C, and essence of workstream D).

Adds fill quantity negotiation to the RFQ wire protocol, such that a responder can propose a fill amount smaller than a requester's maximum, with validation against other limit order constraints at the quote acceptance level (including FOK execution semantics, which was added, but not meaningfully enforced, in #2049). HTLC forwarding policies are capped at the (persisted) negotiated amount so that the channel enforces the agreed-upon limit as a max bound.

There's also some substantial refactoring done around the limit order constraint logic via the RequestConstraints interface.

Here's an illustrative use of the feature from an e2e regtest demo:

```
Frank requests up to 1,000 USDX, min 200. Alice's pilot caps the fill at
500 — above the minimum, so the partial fill is accepted under IOC
semantics.

Fill cap:    500 USDX (pilot constraint)
Requested:   min=200, max=1000 USDX
Market rate: $66,805.00/BTC
✓ Accepted
  RFQ ID: VmojZCspNmgW...
  SCID:   17831787277827515184
  Rate:   $66,805.00/BTC
  Amount: 1000 USDX
  Fill:   min=200, max=1000 USDX
  Cap:    500 USDX (pilot fillcap)

→ AddInvoice (rfq_id: VmojZCspNmgW...)
  500 USDX from Frank, reusing negotiated quote
  Quote confirmed: VmojZCspNmgW...
  Invoice: lnbcrt74844690p1p5ulzcqpp5qnsy5r868n9ql4...

→ Payment: Alice → Eve → Frank
  ✓ Payment succeeded
```

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -130,6 +130,12 @@
   with a new `FOK_NOT_VIABLE` reject code. New fields are optional and
   backward-compatible.
 
+- [Fill Quantity Negotiation](https://github.com/lightninglabs/taproot-assets/pull/2050):
+  RFQ accept messages now carry an optional fill quantity, allowing
+  the responder to signal the maximum amount it is willing to fill.
+  The negotiated fill caps HTLC policies so forwarding never exceeds
+  the agreed amount.
+
 ## Functional Enhancements
 
 - [Wallet Backup/Restore](https://github.com/lightninglabs/taproot-assets/pull/1980):
@@ -187,6 +193,13 @@
   `EXECUTION_POLICY_FOK`) to `AddAssetBuyOrder` and `AddAssetSellOrder`
   requests, and to `PortfolioPilot.ResolveRequest` for constraint
   forwarding. Add `FOK_NOT_VIABLE` to `QuoteRespStatus`.
+
+- [PR#2050](https://github.com/lightninglabs/taproot-assets/pull/2050):
+  Add `max_in_asset` to `PeerAcceptedBuyQuote` and
+  `PeerAcceptedSellQuote` in both the RFQ and PortfolioPilot
+  services, exposing the negotiated fill quantity to RPC clients.
+  Add `fill_amount` to `PortfolioPilot.ResolveResponse` for
+  responder-side fill signalling.
 
 ## tapcli Additions
 
@@ -398,6 +411,10 @@
 - [PR#2023](https://github.com/lightninglabs/taproot-assets/pull/2023)
   Add `DeleteUniverseLeaf` SQL query for single-leaf deletion from a
   universe.
+
+- [PR#2050](https://github.com/lightninglabs/taproot-assets/pull/2050)
+  Add `accepted_max_amount` column to the `rfq_policies` table to
+  persist the negotiated fill quantity alongside HTLC policies.
 
 ## Code Health
 
--- a/itest/custom_channels/limit_constraints_test.go
+++ b/itest/custom_channels/limit_constraints_test.go
@@ -300,6 +300,62 @@ func testCustomChannelsLimitConstraints(_ context.Context,
 	logBalance(t.t, nodes, assetID, "after FOK payment")
 	t.Logf("FOK payment completed: %d units sent", numUnitsFOK)
 
+	// -----------------------------------------------------------------
+	// Negotiate a sell order with explicit IOC policy.
+	// This mirrors the implicit-IOC block above but sets the
+	// execution policy explicitly to prove the RPC surface
+	// accepts and correctly handles the value end-to-end.
+	// -----------------------------------------------------------------
+	t.Logf("Negotiating sell order with explicit IOC policy...")
+
+	sellRespIOC, err := asTapd(charlie).RfqClient.AddAssetSellOrder(
+		ctxb, &rfqrpc.AddAssetSellOrderRequest{
+			AssetSpecifier: &rfqrpc.AssetSpecifier{
+				Id: &rfqrpc.AssetSpecifier_AssetId{
+					AssetId: assetID,
+				},
+			},
+			PaymentMaxAmt: 180_000_000,
+			PaymentMinAmt: 1000,
+			AssetRateLimit: &rfqrpc.FixedPoint{
+				Coefficient: "100000000000000",
+				Scale:       2,
+			},
+			ExecutionPolicy: rfqrpc.ExecutionPolicy_EXECUTION_POLICY_IOC,
+			Expiry:          uint64(inOneHour.Unix()),
+			PeerPubKey:      dave.PubKey[:],
+			TimeoutSeconds:  10,
+		},
+	)
+	require.NoError(t.t, err, "sell order with explicit IOC")
+
+	acceptedIOC := sellRespIOC.GetAcceptedQuote()
+	require.NotNil(
+		t.t, acceptedIOC,
+		"expected accepted IOC sell quote",
+	)
+	t.Logf("IOC sell quote accepted: scid=%d", acceptedIOC.Scid)
+
+	// Pay using the IOC quote with a regular BTC invoice.
+	invoiceRespIOC, err := erin.LightningClient.AddInvoice(
+		ctxb, &lnrpc.Invoice{
+			ValueMsat: invoiceMsat,
+		},
+	)
+	require.NoError(t.t, err)
+
+	var quoteIDIOC rfqmsg.ID
+	copy(quoteIDIOC[:], acceptedIOC.Id)
+
+	numUnitsIOC, _ := payInvoiceWithAssets(
+		t.t, charlie, dave, invoiceRespIOC.PaymentRequest,
+		assetID, withRFQ(quoteIDIOC),
+	)
+	require.Greater(t.t, numUnitsIOC, uint64(0))
+
+	logBalance(t.t, nodes, assetID, "after explicit IOC payment")
+	t.Logf("IOC payment completed: %d units sent", numUnitsIOC)
+
 	// Close channels.
 	closeAssetChannelAndAssert(
 		t, net, charlie, dave, chanPointCD,
--- a/rfq/portfolio_pilot.go
+++ b/rfq/portfolio_pilot.go
@@ -90,21 +90,29 @@ type AssetRateQuery struct {
 	Expiry fn.Option[time.Time]
 }
 
-// ResolveResp captures the portfolio pilot's resolution decision for an RFQ. It
-// carries either an accepted asset rate quote or a structured rejection reason.
+// ResolveResp captures the portfolio pilot's resolution decision for an
+// RFQ. It carries either an accepted asset rate quote or a structured
+// rejection reason, plus an optional fill amount.
 type ResolveResp struct {
-	// outcome holds either the accepted asset rate (left) or the rejection
-	// error (right).
+	// outcome holds either the accepted asset rate (left) or the
+	// rejection error (right).
 	outcome fn.Either[rfqmsg.AssetRate, rfqmsg.RejectErr]
+
+	// fillAmount is an optional fill quantity that caps the amount
+	// the responder is willing to accept.
+	fillAmount fn.Option[uint64]
 }
 
-// NewAcceptResolveResp builds an acceptance response with the provided asset
-// rate quote.
-func NewAcceptResolveResp(assetRate rfqmsg.AssetRate) ResolveResp {
+// NewAcceptResolveResp builds an acceptance response with the provided
+// asset rate quote and optional fill amount.
+func NewAcceptResolveResp(assetRate rfqmsg.AssetRate,
+	fillAmount fn.Option[uint64]) ResolveResp {
+
 	return ResolveResp{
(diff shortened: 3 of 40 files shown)


docs/release-notes/release-notes-0.8.0.md
```diff
@@ -127,9 +127,18 @@
   (Immediate-Or-Cancel, the default) allows partial fills above the
   minimum, while FOK (Fill-Or-Kill) requires the full requested amount
   or rejects the quote. FOK viability is checked in `VerifyAcceptQuote`
-  with a new `FOK_NOT_VIABLE` reject code. New fields are optional and
+  with a new `FOK_NOT_VIABLE` reject code, and is enforced
```
darioAnongba, line 130: Too much verbosity on these release notes IMO. 90% of the details are not meant for users or too complex. I'd remove everything and just add a single entry explaining the new feature clearly and how to use it. The fact it was split in multiple PRs is irrelevant for the users.
We should be able to add a directive to the CLAUDE.MD so agents are more concise.

itest/custom_channels/limit_constraints_test.go
```diff
@@ -260,7 +261,7 @@ func testCustomChannelsLimitConstraints(_ context.Context,
 					AssetId: assetID,
 				},
 			},
-			PaymentMaxAmt: 180_000_000,
+			PaymentMaxAmt: invoiceMsat,
```
darioAnongba, line 264: Already in discussion but this shows an MPP regression. The floor is enforced per-HTLC, but the ceiling is cumulative. Only a single HTLC of exactly `max` works, two HTLCs of `max/2` each fail the floor.

Not sure what's the conclusion of the discussion but if we want FOK to be non-MPP, then we can enforce this explicitly?

What about moving the floor to `CurrentAmountMsat + htlc.AmountOutMsat` with a separate "payment complete" terminator so MPP is not broken?

cc: @GeorgeTsagk

tapdb/sqlc/migrations/000055_rfq_accepted_max.down.sql
```diff
@@ -0,0 +1,93 @@
...
+    chan_id_in BIGINT NOT NULL,
+    chan_id_out BIGINT NOT NULL,
+    htlc_id BIGINT NOT NULL,
+    asset_amt BIGINT NOT NULL,
+    amt_in_msat BIGINT NOT NULL,
+    amt_out_msat BIGINT NOT NULL,
+    UNIQUE(chan_id_in, htlc_id)
+);
+
+INSERT INTO forwards (
```
darioAnongba, line 79: Usual problem, the `INSERT` does not preserve the auto-increment id.

rfq/portfolio_pilot.go
```diff
@@ -507,93 +539,119 @@ func isFOK(p fn.Option[rfqmsg.ExecutionPolicy]) bool {
...
+	return ValidAcceptQuoteRespStatus
+}
 
-	default:
-		log.Warnf("checkMinFill: unhandled request type %T",
-			req)
+// checkFillConstraints verifies that the negotiated fill amount (if
+// present) is compatible with the requester's min-fill and FOK
+// constraints.
+func checkFillConstraints(req rfqmsg.Request,
```
darioAnongba, line 613: Legit issue found by auditor agent:

>  `checkFillConstraints` validates `fill >= MinAmount` and (FOK-only)
> `fill >= MaxAmount` but never `fill <= MaxAmount`. A malicious responder
> sending `AcceptedMaxAmount = 2^64-1` for a 6-unit request passes
> `VerifyAcceptQuote` with `ValidAcceptQuoteRespStatus`.
> 
> Downstream the policy constructor clamps (so HTLCs are safe), but the
> unclamped value persists in:
> - `MarshalAcceptedBuyQuote` / `...SellQuote` → RPC clients see
>   `assetMaxAmount=6, acceptedMaxAmount=1e18`;
> - `tapdb` persistence round-trips the adversarial value;
> - `InternalPortfolioPilot.VerifyAcceptQuote` surfaces a non-error status.

Looks safe to add a check as:

```go
if fillAmt > c.MaxAmount {
    return FillExceedsMaxQuoteRespStatus
}
```
And then forward that error on the layers and map it to a new RPC enum.

darioAnongba (commented):
Pending HTLC discussion for MPP on FOK path.

docs/release-notes/release-notes-0.8.0.md
jtobin: Agreed, will edit.

itest/custom_channels/limit_constraints_test.go
jtobin: Yeah after looking at this again, I think you're right that we don't actually have to sacrifice MPP. Let me take another pass over it..

tapdb/sqlc/migrations/000055_rfq_accepted_max.down.sql
jtobin: Nice one. Will fix.

itest/custom_channels/limit_constraints_test.go
jtobin: Right, so I think it's not necessary to floor the HTLC amount after all. I think the whole detour was a bit of a red herring.

The payment is ultimately going to be requested by way of an invoice anyway, and that will be subject to the standard MPP stuff at the HTLC level, namely: don't cumulatively exceed the requested amount (to be pedantic, the policy amount), and lnd will fail the whole thing atomically if we don't get it all. The whole thing, under an invoicing scheme, is basically FOK by design.

There's actually no need to have differing HTLC enforcement than in the IOC (immediate-or-cancel) case. IOC allows renegotiation of the fill amount by the responder, that's all. Once the fill quantity is decided then you still have to get it all, or fail the payment.

I'll drop those couple of commits that added the HTLC floor.

docs/release-notes/release-notes-0.8.0.md
jtobin: Resolved in latest changeset. I plan to take a clean-up pass over the release notes prior to release in order to prune them down significantly for human consumption, more generally.

rfq/portfolio_pilot.go
jtobin: Fixed in latest changeset.

itest/custom_channels/limit_constraints_test.go
jtobin: Resolved in the latest changeset.

darioAnongba (approved):

rfq/portfolio_pilot_rpc.go
```diff
@@ -165,7 +165,12 @@ func (r *RpcPortfolioPilot) ResolveRequest(ctx context.Context,
 				err)
 		}
 
-		return NewAcceptResolveResp(*assetRate), nil
+		fillAmount := fn.None[uint64]()
+		if resp.AcceptedMaxAmount > 0 {
+			fillAmount = fn.Some(resp.AcceptedMaxAmount)
+		}
+
+		return NewAcceptResolveResp(*assetRate, fillAmount), nil
```
GeorgeTsagk, line 173: don't we need to `checkAllConstraints` here too?

GeorgeTsagk: should also add test coverage for it

GeorgeTsagk (approved):
Lgtm

only have 1 comment for a missing check, not sure if intentional or not

rfq/portfolio_pilot_rpc.go
jtobin: That's an interesting one. After thinking about it for a few minutes, I would say: probably not, at least at the RPC level. The responder doesn't strictly need to check constraints, and it's not really their duty to do so. The onus is morally on the requester to confirm that the quote matches the constraints they set. *Caveat emptor!*

At present the internal (i.e. default) portfolio pilot _does_ call checkAllConstraints in its implementation of ResolveRequest. This is like a "courtesy optimization" to the requestor, in that it saves them the hassle of having to reject a quote that doesn't meet the constraints they set. But it's an implementation detail: a pilot might also want to return a quote that intentionally doesn't meet the constraints in a "take it or leave it" fashion, and one could argue that's courteous in a certain sense as well.

So yeah, I would say we probably _don't_ want to enforce it at the RPC level, and just leave it to the pilot implementor instead.

PR #2051, opened 2026-04-06
rfq+tapdb: persist peer-accepted sell quotes

### LINKED ISSUE #2003: [feature]: RFQ Persistence Analysis Report: Current Durable State, Ephemeral State, and Gaps
## Summary
This report captures RFQ-domain persistence as implemented today, highlights gaps, and proposes follow-up work. The goal is to establish a clear baseline before further schema or lifecycle changes.

## Why this matters
- Restarts preserve only part of RFQ behavior.
- Missing durability in key RFQ paths limits observability and operational confidence.
- We need clear separation between intentionally ephemeral state and accidental durability gaps.

### Persisted today
- Accepted quote policies (buy/sell accepts that become enforceable RFQ policies).
- RFQ-linked forwarding history (open/settle/fail lifecycle data for routed HTLCs).

### Not persisted today
- Rejected quotes.
- Invalid-quote validation outcomes.
- Peer-accepted quote runtime caches.
- In-flight request/session matching state.
- Runtime offer books and HTLC tracking caches.

### Gap 1: Rejections and invalid-quote outcomes are not durable
Impact:
- Loss of post-mortem data for failed negotiations after restart.
- Limited visibility into pricing/qualification failures over time.

Recommendation:
- Persist lightweight RFQ decision events (reject/invalid) with TTL-based retention.

### Gap 2: In-flight RFQ session/request state is restart-fragile
Impact:
- Pending request/response correlation can be lost on restart.
- Operational inconsistency during rolling restarts or crashes.

Recommendation:
- Add short-lived durable in-flight session state with expiry and startup reconciliation.

### Gap 3: Peer-accepted quote cache is memory-only
Impact:
- Quotes that are still valid can become unavailable after restart.
- Reduced continuity for quote usage windows.

Recommendation:
- Persist peer-accepted quotes with strict expiry-driven retention.

### Gap 4: Offer book persistence is undefined
Impact:
- Behavior depends on external re-registration and startup ordering.
- Potentially surprising operator experience after restart.

Recommendation:
- Make an explicit product decision: either persist and replay offers, or document this as intentionally ephemeral.

### Priority 1
- Persist rejection + invalid decision events (with retention policy).

### Priority 2
- Persist in-flight request/session state to improve restart safety.

### Priority 3
- Persist peer-accepted quote cache until quote expiry.

### Priority 4
- Decide and document offer persistence semantics (durable vs ephemeral by design).

## Proposed next steps
1. Confirm which non-durable states are intentional vs gaps.
2. Split follow-up implementation issues by persistence area and priority.
3. Define retention + migration strategy up front.
4. Add restart-focused integration tests for any newly durable state.

## Acceptance criteria for closing this issue
- RFQ persistence baseline documented and agreed.
- Intentional ephemeral states explicitly documented.
- Follow-up issues created for approved persistence additions.

jtobin:
Partially addresses #2003 (together with the peer-accepted buy quote persistence that was added in #2009, this fully resolves "gap 3").

Persists peer-accepted sell quotes to the database. Quotes are now written to the `rfq_policies` table on acceptance and restored into the in-memory cache on restart, mirroring our existing practice for buy-side persistence.

### DIFF
--- a/rfq/manager.go
+++ b/rfq/manager.go
@@ -529,6 +529,18 @@ func (m *Manager) handleIncomingMessage(ctx context.Context,
 			scid := msg.ShortChannelId()
 			m.orderHandler.peerSellQuotes.Store(scid, msg)
 
+			// Persist the peer sell quote to DB so that
+			// the peerSellQuotes cache survives restarts.
+			pStore := m.cfg.PolicyStore
+			storeErr := pStore.StorePeerAcceptedSellQuote(
+				ctx, msg,
+			)
+			if storeErr != nil {
+				log.Errorf("Failed to persist peer sell "+
+					"quote for SCID %d: %v", scid,
+					storeErr)
+			}
+
 			// Notify subscribers of the incoming peer accepted
 			// asset sell quote.
 			event := NewPeerAcceptedSellQuoteEvent(&msg)
--- a/rfq/manager.go
+++ b/rfq/manager.go
@@ -456,23 +456,25 @@ func (m *Manager) handleIncomingMessage(ctx context.Context,
 				return
 			}
 
-			// The quote request has been accepted. Store accepted
-			// quote so that it can be used to send a payment by our
-			// lightning node.
+			// The quote request has been accepted. Persist to
+			// DB first so that on restart the quote is not
+			// silently lost.
 			scid := msg.ShortChannelId()
-			m.orderHandler.peerBuyQuotes.Store(scid, msg)
-
-			// Persist the peer buy quote to DB so that
-			// historical SCID lookups survive quote expiry
-			// and restarts.
-			storeErr := m.cfg.PolicyStore.StorePeerAcceptedBuyQuote(
+			pStore := m.cfg.PolicyStore
+			storeErr := pStore.StorePeerAcceptedBuyQuote(
 				ctx, msg,
 			)
 			if storeErr != nil {
-				log.Errorf("Failed to persist peer buy "+
-					"quote for SCID %d: %v", scid,
-					storeErr)
+				m.handleError(fmt.Errorf("failed to "+
+					"persist peer buy quote for "+
+					"SCID %d: %w", scid, storeErr))
+
+				return
 			}
+
+			// DB write succeeded; populate the in-memory
+			// cache.
+			m.orderHandler.peerBuyQuotes.Store(scid, msg)
 
 			// Since we're going to buy assets from our peer, we
 			// need to make sure we can identify the incoming asset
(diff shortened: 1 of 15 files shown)


rfq/manager.go
```diff
@@ -529,6 +529,18 @@ func (m *Manager) handleIncomingMessage(ctx context.Context,
...
+			// the peerSellQuotes cache survives restarts.
+			pStore := m.cfg.PolicyStore
+			storeErr := pStore.StorePeerAcceptedSellQuote(
+				ctx, msg,
+			)
+			if storeErr != nil {
+				log.Errorf("Failed to persist peer sell "+
+					"quote for SCID %d: %v", scid,
+					storeErr)
+			}
```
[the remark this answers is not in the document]
jtobin: This recommended style isn't used anywhere in the codebase.

rfq/manager.go
```diff
@@ -529,6 +529,18 @@ func (m *Manager) handleIncomingMessage(ctx context.Context,
...
 			m.orderHandler.peerSellQuotes.Store(scid, msg)
 
+			// Persist the peer sell quote to DB so that
+			// the peerSellQuotes cache survives restarts.
+			pStore := m.cfg.PolicyStore
+			storeErr := pStore.StorePeerAcceptedSellQuote(
+				ctx, msg,
+			)
+			if storeErr != nil {
+				log.Errorf("Failed to persist peer sell "+
```
darioAnongba: inherited issue I most likely missed in other buy side PR, but here (and on the buy side), we can't simply log without surfacing the error. If we can't save the quote, it's a hard failure. Here we are silently continuing, meaning that if the DB is down, we will continue emitting events that the quote has been successfully processed when in reality it will disappear on restart.

IMO we should flip ordering: DB write first, cache only on success.

Since it's inherited pattern, we can mix in next PR or in this one.

tapdb/rfq_policies.go
```diff
@@ -267,6 +282,40 @@ func (s *PersistedPolicyStore) StorePeerAcceptedBuyQuote(ctx context.Context,
...
+		PaymentMaxMsat:        fn.Ptr(paymentMax),
+		RequestPaymentMaxMsat: fn.Ptr(paymentMax),
+		PriceOracleMetadata:   acpt.Request.PriceOracleMetadata,
+		RequestVersion: fn.Ptr(
+			uint32(acpt.Request.Version),
+		),
+		AgreedAt: acpt.AgreedAt.UTC(),
+	}
+
+	return s.storePolicy(ctx, record)
```
darioAnongba, line 319: Wasn't this initially an `upsert` operation? We will fail here ON CONFLICT if the same `rfq_id` is sent twice and then silently continue on the manager.

Ideally we should make this idempotent with an upsert where we have `ON CONFLICT DO UPDATE SET` or ON CONFLICT DO NOTHING as an `upsertPolicy`.

Also inherited from buy side.

rfq/manager.go
jtobin: Great catch. I think we should fix them both in this PR.

tapdb/rfq_policies.go
jtobin: Ditto.

rfq/manager.go
jtobin: Fixed in latest changeset.

tapdb/rfq_policies.go
jtobin: Fixed in latest changeset.

darioAnongba (approved):

GeorgeTsagk (approved):
LGTM

pending merge conflicts

PR #2057, opened 2026-04-10
Improving itest clean-up

### LINKED ISSUE #1968: itest: better clean-up
At present, executing the itests typically leaves a bunch of zombie processes running that have to be manually SIGTERM'd. The itest suite should clean up after itself gracefully.

sergey3bv:
Should close #1968

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -331,6 +331,11 @@
 
 ## Testing
 
+- [PR#2057](https://github.com/lightninglabs/taproot-assets/pull/2057)
+  Integration test harness teardown now performs a full top-level harness stop
+  at suite end, so btcd/lnd/tapd itest processes are cleaned up automatically
+  rather than being left running after test completion.
+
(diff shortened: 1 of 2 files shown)


sergey3bv:
Hey, @guggero, @Roasbeef, could you please take a look?

jtobin (approved):
This has bugged me for ages! LGTM. :+1: :+1:

One thing that's a nit here, but is more material in general: "fix: itest clean-up," for example, is IMO a little too uninformative. Better would be to flesh out the description via a brief summary of what the problem is, what the fix is, etc.

A commit message I had Opus generate, for example:

```
itest: use Stop for top-level harness teardown

The top-level t.Cleanup was calling CleanShutDown, which only
recycles LND nodes and clears the mempool. It does not stop the
miner, chain backend, or fee service, so btcd and lnd processes
were left running after test completion.

Switch to Stop for the suite-level cleanup so all itest
infrastructure is torn down. The per-test cleanup retains
CleanShutDown, which is correct: it recycles nodes between
tests while keeping shared infrastructure alive for the rest
of the tranche.
```

GeorgeTsagk (approved):

PR #2060, opened 2026-04-10
Adding diagnostics mode for proof failures

### LINKED ISSUE #1867: [feature]: Add diagnostics mode to capture proof failures and support artefacts
The goal is to introduce an extendable **diagnostics feature** that can aggregate troubleshooting data across tapd subsystems. A user should be able to zip the entire diagnostics directory and provide it to support or link it in a GitHub issue. In the first iteration we will focus on troubleshooting proof-validation failures during asset transfer.

### Scope for first iteration

* Add `--diagnostics-dir` (and config field) to enable diagnostics. Diagnostics mode is disabled unless dir path is given.
* When in diagnostics mode:
    * On tapd start, create a new run directory under the diagnostics directory. Naming: `ts<unix_ts>-pid<pid>`.
    * After anchor tx broadcast during asset transfer, if generated output proofs fail validation, store those proof artefacts in the active diagnostics run directory for support.

### Architecture notes

* Implement `diagnostics` as its own sub-service inside tapd that manages filesystem layout under the diagnostics directory.
* Other subsystems can hand artefacts to the diagnostics service, which decides placement and storage.
* Diagnostics runs non-blocking to avoid impacting normal tapd operation.

### Future directions

* Dump tapd version and build metadata in the run directory.
* Optional trace logging directed to the run directory.
* `diagnostics` handles goroutine dumps and pprof capture under certain conditions.

sergey3bv:
Should close #1867

### DIFF
--- a/diagnostics/service.go
+++ b/diagnostics/service.go
@@ -1,0 +1,293 @@
+package diagnostics
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+const (
+	defaultQueueSize = 64
+)
+
+var fileNameSanitizer = regexp.MustCompile(`[^a-zA-Z0-9._-]+`)
+
+type storedFailureMetadata struct {
+	Timestamp           time.Time `json:"timestamp"`
+	Stage               string    `json:"stage"`
+	Error               string    `json:"error"`
+	AnchorTxID          string    `json:"anchor_txid,omitempty"`
+	VPacketIndex        *int      `json:"vpkt_idx,omitempty"`
+	VPacketOutputIndex  *int      `json:"vpkt_output_idx,omitempty"`
+	TransferOutputIndex *int      `json:"transfer_output_idx,omitempty"`
+	OutputProofFiles    []string  `json:"output_proof_files,omitempty"`
+	InputProofFiles     []string  `json:"input_proof_files,omitempty"`
+}
+
+type queuedFailure struct {
+	id      uint64
+	failure ProofValidationFailure
+}
+
+// Service stores proof-validation diagnostics on disk without blocking the
+// caller.
+type Service struct {
+	rootDir string
+	runDir  string
+
+	queue chan queuedFailure
+
+	wg sync.WaitGroup
+
+	started  atomic.Bool
+	stopOnce sync.Once
+
+	sequence uint64
+	dropped  uint64
+
+	nowFn func() time.Time
+}
+
+// NewService creates a diagnostics service rooted at the given directory.
+func NewService(rootDir string) (*Service, error) {
+	if strings.TrimSpace(rootDir) == "" {
+		return nil, fmt.Errorf(
+			"diagnostics root directory cannot be empty",
+		)
+	}
+
+	return newService(rootDir, defaultQueueSize, time.Now), nil
+}
+
+func newService(rootDir string, queueSize int,
+	nowFn func() time.Time) *Service {
+
+	return &Service{
+		rootDir: rootDir,
+		queue:   make(chan queuedFailure, queueSize),
+		nowFn:   nowFn,
+	}
+}
+
+// Start initializes the run directory and starts the async writer goroutine.
+func (s *Service) Start() error {
+	if s.started.Load() {
+		return nil
+	}
+
+	if err := os.MkdirAll(s.rootDir, 0o700); err != nil {
+		return fmt.Errorf("create diagnostics root dir: %w", err)
+	}
+
+	runDirName := fmt.Sprintf(
+		"ts%d-pid%d", s.nowFn().Unix(), os.Getpid(),
+	)
+	s.runDir = filepath.Join(s.rootDir, runDirName)
+
+	if err := os.MkdirAll(s.runDir, 0o700); err != nil {
+		return fmt.Errorf("create diagnostics run dir: %w", err)
+	}
+
+	log.Infof("Diagnostics run directory initialized at %s", s.runDir)
+
+	s.started.Store(true)
+	s.wg.Add(1)
+	go s.writer()
+
+	return nil
+}
+
+// Stop flushes pending writes and stops the diagnostics service.
+func (s *Service) Stop() error {
+	s.stopOnce.Do(func() {
+		close(s.queue)
+		s.wg.Wait()
+	})
+
+	return nil
+}
+
+// RunDir returns the active diagnostics run directory.
+func (s *Service) RunDir() string {
+	return s.runDir
+}
+
+// DroppedReports returns the number of dropped reports due to queue pressure.
+func (s *Service) DroppedReports() uint64 {
+	return atomic.LoadUint64(&s.dropped)
+}
+
+// CaptureProofValidationFailure stores a failure report asynchronously.
+//
+// The operation is non-blocking. If the queue is full, the report is dropped.
+//
+//nolint:lll // Signature exceeds 80 chars due exported names.
+func (s *Service) CaptureProofValidationFailure(failure ProofValidationFailure) {
+	if s == nil || !s.started.Load() {
+		return
+	}
+
+	queued := queuedFailure{
+		id:      atomic.AddUint64(&s.sequence, 1),
+		failure: cloneFailure(failure),
+	}
+
+	select {
+	case s.queue <- queued:
+	default:
+		atomic.AddUint64(&s.dropped, 1)
+		log.Warnf("Diagnostics queue full, dropping proof failure "+
+			"report (stage=%s)", failure.Stage)
+	}
+}
+
+func (s *Service) writer() {
+	defer s.wg.Done()
+
+	for report := range s.queue {
+		if err := s.writeFailureReport(report); err != nil {
+			log.Warnf("Unable to write diagnostics report: %v", err)
+		}
+	}
+}
+
+func (s *Service) writeFailureReport(report queuedFailure) error {
+	failure := report.failure
+	if failure.Timestamp.IsZero() {
+		failure.Timestamp = s.nowFn().UTC()
+	}
+
+	stage := sanitizeFileName(failure.Stage)
+	if stage == "" {
+		stage = "unknown"
+	}
+
+	eventDirName := fmt.Sprintf(
+		"%d-%s-%06d", failure.Timestamp.Unix(), stage, report.id,
+	)
+	eventDir := filepath.Join(s.runDir, "proof-failures", eventDirName)
+	if err := os.MkdirAll(eventDir, 0o700); err != nil {
+		return fmt.Errorf("create diagnostics event dir: %w", err)
+	}
+
+	outputNames, err := writeArtifacts(
+		eventDir, "output-proof", failure.OutputProofs,
+	)
+	if err != nil {
+		return err
+	}
+
+	inputNames, err := writeArtifacts(
+		eventDir, "input-proof", failure.InputProofs,
+	)
+	if err != nil {
+		return err
+	}
+
+	metadata := storedFailureMetadata{
+		Timestamp:           failure.Timestamp.UTC(),
+		Stage:               failure.Stage,
+		Error:               failure.Error,
+		AnchorTxID:          failure.AnchorTxID,
+		VPacketIndex:        failure.VPacketIndex,
+		VPacketOutputIndex:  failure.VPacketOutputIndex,
+		TransferOutputIndex: failure.TransferOutputIndex,
+		OutputProofFiles:    outputNames,
+		InputProofFiles:     inputNames,
+	}
+
+	metaJSON, err := json.MarshalIndent(metadata, "", "  ")
+	if err != nil {
+		return fmt.Errorf("marshal diagnostics metadata: %w", err)
+	}
+
+	metaPath := filepath.Join(eventDir, "metadata.json")
+	if err := os.WriteFile(metaPath, metaJSON, 0o600); err != nil {
+		return fmt.Errorf("write diagnostics metadata: %w", err)
+	}
+
+	return nil
+}
+
+func writeArtifacts(eventDir, prefix string,
+	artifacts []ArtifactFile) ([]string, error) {
+
+	if len(artifacts) == 0 {
+		return nil, nil
+	}
+
+	writtenNames := make([]string, 0, len(artifacts))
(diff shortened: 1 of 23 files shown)


sergey3bv:
Hey, @jtobin, `make itest-parallel` runs completely fine locally yet fails locally. Could you please take a look

kaldun-tech:
Hey @sergey3bv it looks like you have a bad merge or rebase on your branch that caused the integration test issues. Found the root cause using Claude Opus.

1. limit_constraints_test.go - The entire file was deleted including the entire test for RFQ limit-order constraints (critical coverage)
2. decode_invoice_test.go - Removed the test coverage for multi-tranche group key decoding.
3. liquidity_test.go - Synchronization code was removed which caused this CI failure

Here's Claude's recommendations. Once you have these fixed we can proceed with the rest of the review:

1. Immediate: Revert all changes to test files:
git checkout main -- itest/custom_channels/decode_invoice_test.go
git checkout main -- itest/custom_channels/limit_constraints_test.go
git checkout main -- itest/custom_channels/liquidity_test.go
git checkout main -- itest/custom_channels/passive_assets_test.go
2. Then: Re-apply only the necessary net.Miner → net.Miner.Client changes if needed for API compatibility
3. Review: The PR should be rebased cleanly on main with only the diagnostics-related changes.

sergey3bv:
Hey, @kaldun-tech, I updated the PR according to your comment, could you please take a look.

cc @jtobin

itest/custom_channels/helpers.go
```diff
@@ -2550,13 +2550,19 @@ func locateAssetTransfers(t *testing.T, node *itest.IntegratedNode,
...
+				"(block_height=%d, block_hash_set=%t)",
+				txid,
+				transfer.AnchorTxBlockHeight,
+				blockHashSet,
+			)
 		}
 
 		return nil
-	}, ccTransferTimeout)
+	}, ccTransferConfirmTimeout)
```
kaldun-tech, line 2603: These changes to itest look appropriate to improve CI stability via diagnostics testing

tapfreighter/chain_porter.go
```diff
@@ -621,6 +626,25 @@ func (p *ChainPorter) storeProofs(sendPkg *sendPackage) error {
...
 		)
 		if err != nil {
+			inputProofArtifacts, inputArtifactsErr := p.
+				collectInputProofArtifacts(ctx, inputsForAsset)
+			if inputArtifactsErr != nil {
+				log.Warnf("Unable to collect input proof "+
+					"artifacts for diagnostics "+
+					"(output_idx=%d): %v", idx,
+					inputArtifactsErr)
+			}
```
kaldun-tech: Minor: Here the error is logged as a warning. This seems unusual yet acceptable for diagnostics with best-effort to capture it, given that partial collection is not supported.

diagnostics/service.go
```diff
@@ -0,0 +1,293 @@
...
+	Timestamp           time.Time `json:"timestamp"`
+	Stage               string    `json:"stage"`
+	Error               string    `json:"error"`
+	AnchorTxID          string    `json:"anchor_txid,omitempty"`
+	VPacketIndex        *int      `json:"vpkt_idx,omitempty"`
+	VPacketOutputIndex  *int      `json:"vpkt_output_idx,omitempty"`
+	TransferOutputIndex *int      `json:"transfer_output_idx,omitempty"`
+	OutputProofFiles    []string  `json:"output_proof_files,omitempty"`
+	InputProofFiles     []string  `json:"input_proof_files,omitempty"`
+}
```
kaldun-tech, line 78: Minor: The `metadata.json` schema is not documented. Consider adding a tapd version field in a future PR.

Something like: TapdVersion string `json:"tapd_version,omitempty"`

This would help support teams know which tapd version produced the diagnostics dump when it happens

diagnostics/service.go
```diff
@@ -0,0 +1,293 @@
...
+	select {
+	case s.queue <- queued:
+	default:
+		atomic.AddUint64(&s.dropped, 1)
+		log.Warnf("Diagnostics queue full, dropping proof failure "+
+			"report (stage=%s)", failure.Stage)
+	}
+}
+
+func (s *Service) writer() {
```
kaldun-tech, line 211: Is there a reason to skip a comment block on these new functions?

tapconfig/config.go
```diff
@@ -202,6 +203,10 @@ type Config struct {
 
 	CoinSelect *tapfreighter.CoinSelect
 
+	DiagnosticsService *diagnostics.Service
+
+	Diagnostics diagnostics.Recorder
```
kaldun-tech: Minor: It's confusing to have both a Diagnostics and Diagnostics Service as it doesn't clearly convey a distinction. You could call Diagnostics a DiagnosticsRecorder instead.

diagnostics/service.go
```diff
@@ -0,0 +1,293 @@
...
+		artifact := artifacts[idx]
+		dataCopy := append([]byte(nil), artifact.Data...)
+		clones = append(clones, ArtifactFile{
+			FileName: artifact.FileName,
+			Data:     dataCopy,
+		})
+	}
+
+	return clones
+}
```
kaldun-tech, line 374: File level concern: You don't have a mechanism for disk-space management. Ex:
- Limit total directory size
- Limit number of failure reports
- Delete old reports (retention policy)
- Rotate/archive old runs

Every proof validation failure can write to disk indefinitely. So the risk is if a node has persistent proof validation issues, the diagnostics directory could grow unbounded.

This seems acceptable for version 1 as the feature is expliclty for debugging. In a future enhancement you could add flags `--diagnostics-max-size` or `--diagnostics-retention-days`

kaldun-tech (commented):
Overall this looks solid. The only change I would strongly recommend is where Gemini flagged the missing function comments. The rest is nits that we can take care of in a future PR.

There's clean separation of concerns here and correct async patterns. Good test coverage too.

sergey3bv:
Hey, @kaldun-tech, I updated the PR according to your comments, could you please take a look.

diagnostics/service.go
```diff
@@ -0,0 +1,318 @@
...
+func (s *Service) DroppedReports() uint64 {
+	return atomic.LoadUint64(&s.dropped)
+}
+
+// CaptureProofValidationFailure stores a failure report asynchronously.
+//
+// The operation is non-blocking. If the queue is full, the report is dropped.
+//
+//nolint:lll // Signature exceeds 80 chars due exported names.
+func (s *Service) CaptureProofValidationFailure(failure ProofValidationFailure) {
```
kaldun-tech: Nit: Get rid of the nolint directive and move the parameter to the nextline. Nolint should be reserved for cases with no good alternative.

tapcfg/server.go
```diff
@@ -836,6 +839,8 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 		ProofArchive:             proofArchive,
 		AssetWallet:              assetWallet,
 		CoinSelect:               coinSelect,
+		DiagnosticsService:       diagnosticsService,
+		DiagnosticsRecorder:      diagnosticsService,
```
kaldun-tech: Minor: The DiagnosticsRecorder field now seems redundant as they both point to the same thing. The DiagnosticsService is used for start/stop in the top-level server.go:194-201. The DiagnosticsRecorder field is set at tapcfg/server:843 but never read. The ChainImporter receives its recorder directly at construction above at tapcfg/server:711

My read is this DiagnosticsRecorder on line 843 should be removed.

tapfreighter/chain_porter.go
```diff
@@ -1587,6 +1630,102 @@ func (p *ChainPorter) verifyOutputProofPreBroadcast(ctx context.Context,
 	return nil
 }
 
+func (p *ChainPorter) reportProofValidationFailure(
```
kaldun-tech, line 1651: Nit: functions could use clarifying comments on what they are meant to do. verifyOutputProofPreBroadcast on line 1489 and verifyPacketInputProofs on line 1731 have comments for example.

tapfreighter/chain_porter.go
```diff
@@ -1587,6 +1630,102 @@ func (p *ChainPorter) verifyOutputProofPreBroadcast(ctx context.Context,
...
+	failure diagnostics.ProofValidationFailure) {
+
+	if p.cfg.DiagnosticsRecorder == nil {
+		return
+	}
```
kaldun-tech, line 1657: In reportProofValidationFailure for disabled diagnostics we return nil. But the calls to reportProofValidationFailure passes in a ProofValidationFailure. So we build a parameter that is not necessarily used.

You can optimize this to avoid unnecessary I/O and allocations by guarding the entire block

diagnostics/types.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+const (
+	// StageProofVerificationPreBroadcast marks failures from pre-broadcast
+	// proof validation.
+	StageProofVerificationPreBroadcast = "pre_broadcast"
+
+	// StageProofVerificationPostBroadcast marks failures from
+	// post-broadcast
+	// proof validation.
```
kaldun-tech: nit: looks like an extra newline here

kaldun-tech:
In my view you're on the right track in this iteration except for  a few nitpicks & minor issues. The deep copy logic is smart towards the bottom of diagnostics/service.go.

diagnostics/service.go
kaldun-tech: Please add explanations of why the fields that are marked `omitempty` are safe to be omitted. A reader can't tell from the struct definition alone which fields appear in which stage's reports.

diagnostics/service.go
```diff
@@ -0,0 +1,318 @@
...
+			)
+		}
+
+		writtenNames = append(writtenNames, fileName)
+	}
+
+	return writtenNames, nil
+}
+
+func sanitizeFileName(name string) string {
```
kaldun-tech, line 328: I recommend to add more comments on at least the added functions as Gemini pointed out before. What does the reader need to understand about the intent of each function? In this case stripping non-alphanumeric characters from a filename

diagnostics/service.go
```diff
@@ -0,0 +1,318 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+const (
+	defaultQueueSize = 64
+)
+
+var fileNameSanitizer = regexp.MustCompile(`[^a-zA-Z0-9._-]+`)
```
kaldun-tech, line 38: Would also benefit from a comment

diagnostics/service.go
```diff
@@ -0,0 +1,318 @@
...
+		failure.Timestamp = s.nowFn().UTC()
+	}
+
+	stage := sanitizeFileName(failure.Stage)
+	if stage == "" {
+		stage = "unknown"
+	}
+
+	eventDirName := fmt.Sprintf(
+		"%d-%s-%06d", failure.Timestamp.Unix(), stage, report.id,
```
kaldun-tech: Readers would benefit from explanatory comments or clearly named constants for magic strings/numbers like this and `0o700` below. In this case this produces directory names like `1716134400-pre_broadcast-000001` with format `<unix_timestamp>-<stage>-<sequence>`.

tapfreighter/chain_porter.go
```diff
@@ -1587,6 +1630,102 @@ func (p *ChainPorter) verifyOutputProofPreBroadcast(ctx context.Context,
...
+
+func (p *ChainPorter) collectInputProofArtifacts(ctx context.Context,
+	inputs []asset.PrevID) ([]diagnostics.ArtifactFile, error) {
+
+	artifacts := make([]diagnostics.ArtifactFile, 0, len(inputs))
+	for idx := range inputs {
+		inputProofFile, err := p.fetchInputProof(ctx, inputs[idx])
+		if err != nil {
+			return nil, fmt.Errorf("fetch input proof %d: %w",
+				idx, err)
```
kaldun-tech, line 1748: This function fails fast if there's an internal error while iterating over the inputs. The reason this is sensible is this function is called in cases where proof verification failed so we want to report the error promptly and move on. Partial proof artifacts could be misleading during support investigation.

Readers would benefit from explanatory comments.

kaldun-tech (changes requested):
For the most part the code here is solid and addresses the original issue. I believe it's worthwhile to add explanatory comments in new and updated files and remove the added `nolint` directive.

sergey3bv:
Hey, @kaldun-tech, I updated the PR according to your comments. Could you please take a look?

diagnostics/service.go
kaldun-tech: Good improvement!

itest/custom_channels/helpers.go
kaldun-tech: Sorry to flip-flop on this. My concern is that the itest changes are extensive enough to be out of scope for the PR. They look good but are orthogonal to diagnostics. It could introduce a regression and make the commit history muddled.

It's up to the repo maintainers in on whether they want to keep the itest changes bundled in this change or split them into a separate PR.

diagnostics/service.go
```diff
@@ -0,0 +1,374 @@
...
+	queued := queuedFailure{
+		id:      atomic.AddUint64(&s.sequence, 1),
+		failure: cloneFailure(failure),
+	}
+
+	select {
+	case s.queue <- queued:
+	default:
+		atomic.AddUint64(&s.dropped, 1)
+		log.Warnf("Diagnostics queue full, dropping proof failure "+
```
kaldun-tech, line 201: Warning for queue full seems sensible for an initial version. If there is concern about log spam we could switch to rate-limited warnings

diagnostics/service.go
```diff
@@ -0,0 +1,374 @@
...
+	dropped  uint64
+
+	nowFn func() time.Time
+}
+
+// NewService creates a diagnostics service rooted at the given directory.
+func NewService(rootDir, tapdVersion string) (*Service, error) {
+	if strings.TrimSpace(rootDir) == "" {
+		return nil, fmt.Errorf(
+			"diagnostics root directory cannot be empty",
```
kaldun-tech, line 110: Did a scan of all the error messages here. They seem appropriate.

kaldun-tech (approved):
Lookds good to me! Curious to hear from the maintainers on the disk space and itest concerns.

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #2062, opened 2026-04-10
multi: forward-port burn by group key from staging

### LINKED ISSUE #2061: rpc: add group_key support to BurnAsset and FetchAssetMeta RPCs
## Context

The tap-sdk's `AssetRef` redesign (lightninglabs/tap-sdk#48) introduced an opaque identifier that hides whether the underlying identifier is an `asset_id` or `group_key`. However, `BurnAsset` and `FetchAssetMeta` RPCs currently only accept `asset_id`, forcing the SDK to either resolve group keys to asset IDs client-side or reject group-key refs.

## Proposal

Add optional `group_key` fields to:
- `BurnAssetRequest` — allow burning by group key (tapd picks the concrete issuance)
- `FetchAssetMetaRequest` — allow metadata lookup by group key (tapd returns canonical/genesis metadata)

This would allow the SDK to pass `AssetRef` specifiers through to tapd without client-side resolution.

## Current Workaround

The SDK returns an error when a group-key `AssetRef` is used for burns or metadata lookups, directing users to use an asset-ID ref instead.

darioAnongba:
## Summary

Forward-port the "burn assets by group key" feature from PR #1812 (0-8-0-staging) to main. The staging branch was retired before this feature was cherry-picked.

- **taprpc**: Add `AssetSpecifier` field to `BurnAssetRequest` (deprecating the old `oneof asset` fields). Add repeated `burn_proofs` field to `BurnAssetResponse` for multi-proof burns.
- **tapfreighter**: Allow `FundBurn` to work with group-key-only specifiers by removing the `UnwrapIdOrErr()` requirement and setting burn script keys per-packet after input assignment.
- **rpcserver**: Update `BurnAsset` handler to parse `AssetSpecifier`, support group-key-only burns, sign all virtual packets, and return multiple burn proofs.
- **cmd**: Add `--group_key` flag to `tapcli assets burn` with group-balance confirmation support.
- **itest**: Update all burn tests to use `AssetSpecifier` and `burn_proofs`. Add burn-by-group-key test case in `testBurnGroupedAssets`.

Original commits on staging:
- `1f425251` tapfreighter: burn by group key
- `9d8f0877` taprpc: AssetSpecifier in BurnAssetRequest
- `56e571ca` rpc: burn by group key
- `6755dddc` cmd: add group_key flag to tapcli assets burn
- `5e2bbcd7` itest: burn by group key


tapfreighter/wallet.go
```diff
@@ -800,52 +802,23 @@ func (f *AssetWallet) FundBurn(ctx context.Context,
...
 			AnchorOutputIndex: 0,
 			AssetVersion:      maxVersion,
-			ScriptKey:         burnKey,
+			// We'll set the burn script key per packet after
+			// inputs are selected so we use a placeholder here.
+			ScriptKey: asset.NUMSScriptKey,
 		}},
 		ChainParams: f.cfg.ChainParams,
 		Version:     tappsbt.V1,
 	}
```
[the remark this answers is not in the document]
darioAnongba: This concern is incorrect. For grouped assets, `Asset.TapCommitmentKey()` returns `sha256(groupKey)` (not the asset ID), because `Asset.Specifier()` includes the group key when one exists. So the filter on line 53 of `createFundedPacketWithInputs` correctly matches all tranches within the group against `fundDesc.TapCommitmentKey()`.

`DistributeCoins` then creates separate VPackets per tranche when the amounts require splitting across multiple asset IDs within the group. Each VPacket has inputs with the same underlying asset, so the single-output template works correctly.

This was the same behavior on the original staging branch (commit 1f425251).

tapfreighter/wallet.go
```diff
@@ -857,11 +830,41 @@ func (f *AssetWallet) FundBurn(ctx context.Context,
...
+			}
+
+			// Compute burn key from the first input's PrevID.
+			firstPrevID := pkt.Inputs[0].PrevID
+			burnKey := asset.NewScriptKey(
+				asset.DeriveBurnKey(firstPrevID),
+			)
+			out.ScriptKey = burnKey
+			out.AssetVersion = pktMaxVersion
+		}
```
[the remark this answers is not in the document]
darioAnongba: This concern is incorrect. `DistributeCoins` creates separate VPackets per tranche/asset-ID. Within a single VPacket, all inputs share the same asset ID. The burn key derivation from `pkt.Inputs[0].PrevID` is correct because:

1. All inputs in the packet have the same asset ID (enforced by `DistributeCoins`).
2. The burn protocol requires the burn key to be derived from the first input's PrevID.
3. There is only one interactive (burn) output per packet.

This is the same approach used in the original staging implementation (commit 1f425251 / PR #1812).

cmd/commands/assets.go
```diff
@@ -1244,53 +1269,116 @@ func burnAssets(ctx *cli.Context) error {
...
+		return fmt.Errorf("unable to burn assets: %w", err)
 	}
 
 	printRespJSON(resp)
 	return nil
 }
 
+// buildBurnAssetSpecifier returns the appropriate AssetSpecifier for the burn
+// request based on which bytes are provided. Asset ID takes precedence when
+// both are given (the server validates consistency).
```
jtobin: Flagged by Opus:

> The AssetSpecifier is a oneof (line 227 of taprootassets.proto), so only one field can be set. buildBurnAssetSpecifier (line 1364) uses a switch that picks asset ID first, silently dropping the group key.  The comment on line 1360 — "the server validates consistency" — is wrong; the group key never reaches the server.
> 
> Best fix: reject at the CLI when both flags are provided, since the oneof can't carry both anyway.

cmd/commands/assets.go
```diff
@@ -1244,53 +1269,116 @@ func burnAssets(ctx *cli.Context) error {
...
+				return fmt.Errorf("couldn't fetch balance "+
+					"for group key %x",
+					groupKeyBytes)
+			}
+
+			currBalance = groupBalance.Balance
 		}
 
+		specStr := fmt.Sprintf(
+			"asset_id=%x, group_key=%x",
```
jtobin: Paraphrasing Opus: if only a group key is passed, this renders as `asset_id=, group_key=deadbeef..`. Might be worth rendering the fields conditionally.

rpcserver/rpcserver.go
```diff
@@ -3947,36 +3929,117 @@ func (r *RPCServer) BurnAsset(ctx context.Context,
...
+				"may not be part of a group")
 
-	assetSpecifier := asset.NewSpecifierOptionalGroupPubKey(
-		assetID, groupKey,
+		case err != nil:
+			return nil, fmt.Errorf("error querying asset "+
+				"group: %w", err)
+		}
+
+	case assetID != nil && groupKey != nil:
```
jtobin, line 4007: I don't think these can ever both be nil? Worth keeping the case for completeness, but we could probably get rid of the additional branching and just return `nil, fmt.Errorf("impossible")` or whatever as a simplification?

jtobin (approved):
LGTM. :+1: :+1:

I noted a few slightly-above-nit-grade things flagged by agents that are probably worth fixing, but that IMO don't block merge. (And sorry for dropping these changes the first time around! 😁)

GeorgeTsagk (approved):
LGTM

Issue #1802, opened 2025-09-16
[feature]: Add unit tests for `supplyverifier.Verifier`


ffranr:
Expand test coverage for the `supplyverifier.Verifier` struct and its methods.

Tests should include:

* **Method-level unit tests** for individual `Verifier` methods, each of which already encapsulates distinct concerns.
* **End-to-end unit tests** starting from the public `Verifier.VerifyCommit` entry point, to validate the complete verification flow.

UPDATE (jtobin):

#2063 delivered the following (Opus's summary):

  - VerifierCfg.Validate / NewVerifier (config validation)
  - IsEquivalentPubKeys (BIP340 parity)
  - verifyPrecommitsSpent (height filtering, spend checks)
  - FetchPreCommits / FetchDelegationKey (in util_test.go)
  - Mocks: MockSupplyTreeView, MockGroupFetcher

What remains, again per Opus:

```
  Method: verifyIssuanceLeaf
  Complexity: High (~115 lines, proof verification + 8 checks)

  Method: verifyInitialCommit
  Complexity: High (~100 lines, tree reconstruction + root check)

  Method: verifyIncrementalCommit
  Complexity: High (~110 lines, chain-of-custody + tree update)

  Method: verifyIgnoreLeaf
  Complexity: Medium (~55 lines, Schnorr sig + group check)

  Method: verifyBurnLeaf
  Complexity: Medium (~60 lines, proof verification + group check)

  Method: verifySupplyLeaves
  Complexity: Low (dispatcher)

  Method: VerifyCommit (e2e)
  Complexity: High (full flow)

  Plus ffranr's two suggested checks (mint event height validation, pre-commit set
  membership).

  In short: the foundational/simpler methods and mock infrastructure are done. The
  core verification logic (leaf verification, initial/incremental commit
  verification) and the end-to-end VerifyCommit tests are all still outstanding.
```

ffranr:
Other checks we could add:

* Verify that the mint event block height is less than the supply commit block height.
* Verify that pre-commits for mint events are present in the pre-commit set provided to the verifier.

rxbryan:
Hi, Could I be assigned to this?
I'd like to work on this. Planning to start with method-level unit tests for 
- verifyPrecommitsSpent (including the block height ordering and pre-commit presence checks)
-  IsEquivalentPubKeys
-  VerifierCfg.Validate/NewVerifier

[...]

PR #2063, opened 2026-04-11
supplyverifier: add unit tests for Verifier

rxbryan:
relates to #1802
Adds unit tests for the `supplyverifier.Verifier` struct and its methods.
- add mocks for `supplyverifier.SupplyTreeView` and `tapgarden.GroupFetcher`.
- `TestVerifierCfgValidate`: all six required field validations in
  `VerifierCfg.Validate`, plus happy path.
- `TestNewVerifier`: construction failure path, config preservation, and
  functional logger check.
- `TestIsEquivalentPubKeys`: same key, distinct keys, and the BIP340 parity
  edge case where P and -P share the same x-coordinate and serialize identically
  under x-only encoding, so `IsEquivalentPubKeys` must return true for both.
- `TestVerifyPrecommitsSpent`: five cases including  pre-commits above commitment block height are excluded from the required-spend set, and pre-commits at or below commitment height not spent by the commitment transaction return an error.

### DIFF
--- a/universe/supplyverifier/mock.go
+++ b/universe/supplyverifier/mock.go
@@ -3,8 +3,11 @@ package supplyverifier
 import (
 	"context"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/taproot-assets/asset"
+	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightninglabs/taproot-assets/mssmt"
 	"github.com/lightninglabs/taproot-assets/universe/supplycommit"
 	lfn "github.com/lightningnetwork/lnd/fn/v2"
 	"github.com/stretchr/testify/mock"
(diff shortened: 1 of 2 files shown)


jtobin (approved):
LGTM. :+1: :+1:

I think to fully resolve the linked issue we'd want to add cases for most of the other Verifier methods too (I think only proofVerifierCtx is really trivial, and could be omitted), as well as the spec'd e2e tests, and probably the other checks that ffranr mentioned in his comment. But this is still IMO a strict improvement; I would be happy to merge as-is, and just update the issue accordingly.

docs/release-notes/release-notes-0.8.0.md
```diff
@@ -354,6 +354,13 @@
   Add unit, property-based, and integration tests for limit-order
   constraint fields.
 
+- [PR#2063](https://github.com/lightninglabs/taproot-assets/pull/2063)
```
darioAnongba: no need for this entry in the changelog, since there are not external facing changes.

darioAnongba (changes requested):

docs/release-notes/release-notes-0.8.0.md
rxbryan: Done, removed the release notes entry.

rxbryan:
> I think to fully resolve the linked issue we'd want to add cases for most of the other Verifier methods too (I think only proofVerifierCtx is really trivial, and could be omitted), as well as the spec'd e2e tests, and probably the other checks that ffranr mentioned in his comment. But this is still IMO a strict improvement; I would be happy to merge as-is, and just update the issue accordingly.

Thanks for the context @jtobin. I'll follow up with a second PR once this is merged covering the remaining methods: `verifyIssuanceLeaf`, `verifyIgnoreLeaf`, `verifyBurnLeaf`, `verifySupplyLeaves`, `verifyInitialCommit`, `verifyIncrementalCommit`, and the e2e tests starting from `VerifyCommit`.

darioAnongba (approved):

PR #2069, opened 2026-04-14
itest: deflake custom channel liquidity cleanup

### LINKED ISSUE #1900: [flake]: TestLightningTerminal/custom_channels_liquidity flake
This flake was observed in a CI job for #1899 ([log](https://github.com/lightninglabs/taproot-assets/actions/runs/19936091891/job/57161218512?pr=1899)).

GeorgeTsagk:
## Description

The fix removes the flake by waiting for HTLC cancellation to fully propagate across the whole
`Charlie->Dave->Erin->Fabia `
route before starting the next payment, which prevents a race where upstream nodes still report temporarily reduced bandwidth.

We basically have to wait for all HTLCs to fail on all the channels involved in the prior payment.

~Closes~ **Seals** https://github.com/lightninglabs/taproot-assets/issues/1900

### DIFF
--- a/itest/custom_channels/helpers.go
+++ b/itest/custom_channels/helpers.go
@@ -2748,6 +2748,18 @@ func assertNumHtlcs(t *testing.T, node *itest.IntegratedNode,
 	require.NoError(t, err)
 }
 
+// assertNumHtlcsAll asserts that each node has exactly the expected number of
+// pending HTLCs across all channels.
+func assertNumHtlcsAll(t *testing.T, expected int,
+	nodes ...*itest.IntegratedNode) {
+
+	t.Helper()
+
+	for _, node := range nodes {
+		assertNumHtlcs(t, node, expected)
+	}
+}
+
 // assertHTLCNotActive asserts the node doesn't have an active pending HTLC
 // with the given payment hash.
 func assertHTLCNotActive(t *testing.T, hn *itest.IntegratedNode,
--- a/itest/custom_channels/liquidity_test.go
+++ b/itest/custom_channels/liquidity_test.go
@@ -636,8 +636,10 @@ func testCustomChannelsLiquidityEdgeCasesCore(ctx context.Context,
 	)
 	require.NoError(t.t, err)
 
-	// Let's assert that Erin cancelled all his HTLCs.
-	assertNumHtlcs(t.t, erin, 0)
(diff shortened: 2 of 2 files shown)


jtobin (approved):
LGTM. Death to flakes! 💀 

I wonder: are there any other flaky itests that could use this same kind of assertion? I asked Opus to check:

> High risk — same pattern as the PR fix
> 
> 1. liquidity_test.go:640 — This is the line the PR fixes. Only Erin was checked before the next payInvoiceWithAssets on line 651 along Charlie->Dave->Erin->Fabia.
> 2. multi_rfq_test.go:147-149 — Checks Dave, Erin, Yara but not Charlie or Fabia. A payInvoiceWithSatoshi from Charlie follows immediately on line 160. Charlie is the payer here and could still have pending cancellations on its outbound links. Adding Charlie (and Fabia for completeness) to the check would be appropriate.
> 3. multi_rfq_test.go:210-211 — Checks Charlie and Fabia but not the intermediate nodes (Dave, Erin, Yara, George). A payInvoiceWithSatoshi from Charlie follows on line 220. The intermediaries could still have in-flight HTLC cancellations. All route nodes should be checked.
> 
> Lower risk — already checking all route nodes or has a buffer
> 
> 4. liquidity_test.go:483-486 — Already checks all four nodes (Charlie, Dave, Erin, Fabia). However, yara is also in the topology (Charlie <-> Dave <-> Yara) and is not checked. If Yara's channels are involved, it could be a gap. If not, this is fine as-is. Worth verifying the topology here.
> 5. liquidity_test.go:584 — Only checks Dave on the Erin->Dave->Charlie route. However, there's a CloseSend + ResetMissionControl + a new test case setup before the next payment, providing a natural buffer. Still, for robustness, checking all three nodes (Erin, Dave, Charlie) would be consistent with the PR's approach.

multi-rfq is also known to be flaky, in particular, so might be worth investigating that one in more detail?

darioAnongba (approved):

jtobin:
@GeorgeTsagk while we're battling with the merge queue issues, I made a PR to this branch that adds the assertion to another few known-flakey cases as well. It's probably worth pulling in, so we could add it to this one, or just do it standalone. Up to you.

jtobin:
This one likely could use a rebase onto main following the merge of #2072.

PR #2070, opened 2026-04-16
tapgarden: surface non-critical custodian errors to subscribers

### LINKED ISSUE #1942: [feature]: Surface non-critical Custodian errors to UI/API
## Problem
Recent changes (`285743f`, `39405ea`) in `tapgarden` downgraded several critical errors (in `inspectWalletTx`, `receiveProofs`, etc.) to logs to prevent daemon crashes. However, this means these errors are now invisible to UI/API clients, potentially leaving users unaware of failed asset transfers or imports.

## Analysis
- **Current Behavior:** Errors in `Custodian.mainEventLoop` (specifically during `inspectWalletTx`, `mapProofToEvent`, `handleMailboxMessages`, and `receiveProofs`) are logged, and the loop continues.
- **Existing Infrastructure:** 
    - `tapgarden` has an `AssetReceiveEvent` which includes an `Error` field.
    - `NewAssetReceiveErrorEvent` helper exists.
    - `SubscribeReceiveEvents` RPC is already capable of mapping this internal error field to the `ReceiveEvent` protobuf message `error` field.
- **Gap:** The code currently logs the error instead of publishing an `AssetReceiveErrorEvent` in these specific failure paths.

## Proposal
Instrument the `Custodian` to publish `AssetReceiveErrorEvent` before logging and continuing in the following areas:
1. **`inspectWalletTx`**: When `address.NewTransferFromWalletTx` or `GetOrCreateEvent` fails.
2. **`mapProofToEvent`**: When `GetOrCreateEvent` fails (currently only logs).
3. **`handleMailboxMessages`**: When processing specific messages fails.

This will utilize the existing `SubscribeReceiveEvents` RPC stream to notify clients of these failures without requiring protocol changes.

kaldun-tech:
Fixes #1942.

Adds `NewAssetReceiveErrorEvent` to publish error events through the existing subscriber mechanism. Error paths in `inspectWalletTx`, `handleMailboxMessages`, and `mapProofToEvent` now notify subscribers before returning/continuing on errors.

This enables subscribers (e.g., RPC clients watching for receives) to be notified about errors during asset receive processing, rather than only learning about successful completions or silent failures logged to the daemon.

Unit tests added for both the subscriber event flow and error event structure.

### DIFF
--- a/tapgarden/custodian.go
+++ b/tapgarden/custodian.go
@@ -741,11 +741,22 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 			if event.ConfirmationHeight == 0 &&
 				walletTx.Confirmations > 0 {
 
+				// Save these before GetOrCreateEvent, which may
+				// return nil on error.
+				tapAddr := event.Addr.Tap
+				prevStatus := event.Status
+
 				xfer, err := address.NewTransferFromWalletTx(
 					event.Addr, walletTx, uint32(idx),
 					sendOutputs,
 				)
 				if err != nil {
+					errEvent := NewAssetReceiveErrorEvent(
+						err, *tapAddr, op,
+						0, prevStatus,
+					)
+					c.publishSubscriberStatusEvent(errEvent)
+
 					return fmt.Errorf("error creating "+
 						"event source: %w", err)
 				}
@@ -758,6 +769,12 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 				)
 				cancel()
 				if err != nil {
+					errEvent := NewAssetReceiveErrorEvent(
+						err, *tapAddr, op,
+						0, prevStatus,
+					)
+					c.publishSubscriberStatusEvent(errEvent)
+
 					return fmt.Errorf("error updating "+
 						"event: %w", err)
 				}
@@ -778,7 +795,7 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 				)
 				c.Goroutine(func() error {
 					return c.receiveProofs(
-						event.Addr.Tap, op,
+						tapAddr, op,
 						event.Outputs,
 						event.ConfirmationHeight,
 					)
@@ -787,8 +804,8 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 						//nolint:lll
 						NewAssetReceiveErrorEvent(
 							err,
-							*event.Addr.Tap,
-							event.Outpoint,
+							*tapAddr,
+							op,
 							event.ConfirmationHeight,
 							event.Status,
 						),
@@ -1184,6 +1201,12 @@ func (c *Custodian) handleMailboxMessages(msgs *mbox.ReceivedMessages) error {
 				output.DerivationMethod,
 			)
 			if err != nil {
+				errEvent := NewAssetReceiveErrorEvent(
+					err, *tapAddr.Tap, op, 0,
+					address.StatusTransactionDetected,
+				)
+				c.publishSubscriberStatusEvent(errEvent)
+
 				return fmt.Errorf("unable to derive script "+
 					"key for asset: %w", err)
 			}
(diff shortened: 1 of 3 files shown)


tapgarden/custodian.go
```diff
@@ -1179,6 +1206,12 @@ func (c *Custodian) handleMailboxMessages(msgs *mbox.ReceivedMessages) error {
 				output.DerivationMethod,
 			)
 			if err != nil {
+				errEvent := NewAssetReceiveErrorEvent(
+					err, *tapAddr.Tap, op, 0,
+					address.StatusTransactionDetected,
+				)
+				c.publishSubscriberStatusEvent(errEvent)
```
kaldun-tech, line 1213: This has the same 0 outpoint sentinel value as above. LMK if it would help to add another comment here.

tapgarden/custodian_test.go
```diff
@@ -1216,3 +1216,149 @@ func TestAddrMatchesAsset(t *testing.T) {
 		})
 	}
 }
+
+// TestCustodianEventSubscriber tests that subscribers receive events when
+// assets are received. This validates the publishSubscriberStatusEvent
+// mechanism used by both success and error events.
+func TestCustodianEventSubscriber(t *testing.T) {
```
kaldun-tech, line 1223: This test is for the non-error happy path

tapgarden/custodian.go
```diff
@@ -755,6 +762,13 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 				)
 				cancel()
 				if err != nil {
+					errEvent := NewAssetReceiveErrorEvent(
+						err, *event.Addr.Tap, op,
```
jtobin: If err != nil, then event is nil, so `*event.Addr.Tap` will panic. The address should probably be saved before calling GetOrCreateEvent(), and then you can just sub it in here instead.

tapgarden/custodian.go
```diff
@@ -1159,12 +1173,25 @@ func (c *Custodian) handleMailboxMessages(msgs *mbox.ReceivedMessages) error {
 			ctx, receiver, mboxMsg.EncryptedPayload,
 		)
 		if err != nil {
-			// If we can't decrypt the message, we'll log the
-			// error but continue processing the next message. This
-			// way nobody can corrupt our state by sending us a
-			// malformed message.
+			// If we can't decrypt the message, we'll publish an
```
jtobin: I think we should just preserve the existing behaviour here. Failed decryption is not really an asset receive error *per se*, so I don't think we need to handle it as such. Logging and continuing seems correct to me.

jtobin (changes requested):
Just a few changes required IMO:

* I wrote a couple of comments re: a nil pointer deref problem and not explicitly handling decryption errors in handleMailboxMessages, please check those out.
* The fix-up commits should be squashed into the commits that introduce the changes, or just a single fix-up commit would also be fine IMO. Please make sure they obey the `change target: change summary` style of other commits, too.

kaldun-tech:
Hey Jared, thanks for the review! Addressed both points:
1. Nil pointer deref - Now saving tapAddr and prevStatus before GetOrCreateEvent and using those in error paths (lines 741-744...).
2. Decryption errors - Restored the original behavior (log and continue). I agree that failed decryption from external parties isn't really an asset receive error.

Also squashed everything into a single commit.

tapgarden/custodian.go
```diff
@@ -787,10 +793,10 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 						//nolint:lll
 						NewAssetReceiveErrorEvent(
 							err,
-							*event.Addr.Tap,
-							event.Outpoint,
-							event.ConfirmationHeight,
-							event.Status,
+							*tapAddr,
+							op,
+							0,
```
jtobin: Opus flagged this as a regression from the previous changeset: `event.ConfirmationHeight` and `event.Status` should be preserved here, because at this point we know that GetOrCreateEvent has succeeded (if it hadn't, the goroutine wouldn't have launched). prevStatus and 0 would be incorrect values here.

jtobin (changes requested):
One problem introduced in the last changeset, otherwise looks clean.

tapgarden/custodian.go
kaldun-tech: Good catch! Looks like a mistaken change on my end.

tapgarden/custodian.go
```diff
@@ -741,11 +741,22 @@ func (c *Custodian) inspectWalletTx(walletTx *lndclient.Transaction) error {
 			if event.ConfirmationHeight == 0 &&
 				walletTx.Confirmations > 0 {
 
+				// Save these before GetOrCreateEvent, which may
+				// return nil on error.
+				tapAddr := event.Addr.Tap
```
kaldun-tech, line 746: I pass the stored *tapAddr and op (Outpoint from line 717) into the ErrorEvents

jtobin (approved):
LGTM. :+1: :+1:

PR #2082, opened 2026-04-20
multi: bump lnd to master, pin ANCHORS in custom_channels itest

GeorgeTsagk:
## Description

Bumps lnd to master tip to pull in the production taproot work and pins ANCHORS as the default commitment type in `openChannelAndAssert` so the custom_channels itest's public "normal sats" channels aren't swept up by lnd's new implicit-negotiation preference for final-taproot (which rejects public channels).

Depends on https://github.com/lightningnetwork/lnd/pull/10763


itest/custom_channels/assertions.go
```diff
@@ -76,6 +76,15 @@ func openChannelAndAssert(t *ccHarnessTest, net *itest.IntegratedNetworkHarness,
 
 	t.t.Helper()
 
+	// Pin the commitment type to anchors when the caller hasn't chosen
+	// one. With recent lnd master, the default (implicit) negotiation
```
darioAnongba, line 80: we should avoid this type of comments in the codebase "With the recent lnd master", that has no meaning outside of the commit/PR description. Not sure why agents don't understand this and add comments like that in the code like they're having a conversation lol.

itest/custom_channels/assertions.go
```diff
@@ -76,6 +76,15 @@ func openChannelAndAssert(t *ccHarnessTest, net *itest.IntegratedNetworkHarness,
...
+	// prefers taproot-final, which is rejected for public channels. Our
+	// "normal sats" channels in this harness are public (they carry
+	// gossip), so we need an explicit non-taproot type here.
+	if p.CommitmentType == lnrpc.CommitmentType_UNKNOWN_COMMITMENT_TYPE {
+		p.CommitmentType = lnrpc.CommitmentType_ANCHORS
+	}
```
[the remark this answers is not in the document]
darioAnongba: does this make sense?

docs/release-notes/release-notes-0.8.0.md
```diff
@@ -364,6 +364,13 @@
   at suite end, so btcd/lnd/tapd itest processes are cleaned up automatically
   rather than being left running after test completion.
 
+- [PR#2082](https://github.com/lightninglabs/taproot-assets/pull/2082)
```
darioAnongba, line 367: this PR is a "no changelog" IMO as we can't add a changelog entry every-time we bump to a LND master commit because "master" means nothing in the changelog (there is no reference). We should add the changelog entry only when we require a tagged version before releasing.

GeorgeTsagk: will change to `no changelog`

itest/custom_channels/assertions.go
GeorgeTsagk: Yeah was a bit more permissive since this is the test harness

Will adjust comment

itest/custom_channels/assertions.go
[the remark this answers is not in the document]
GeorgeTsagk: not really.. this was already too much of a strict check. With LND side merged it might even make sense to remove entirely as the `CommitmentType_UNKNOWN_COMMITMENT_TYPE ` shouldn't be encoutnered anymore

jtobin (commented):
Looks like lightningnetwork/lnd#10763 landed, so probably worth retargeting the replace directive at the canonical repo? Also agree w/Dario's comment re: axing the release note entry. Otherwise LGTM! :+1: :+1:

darioAnongba (approved):
LGTM

jtobin (approved):

PR #2084, opened 2026-04-20
itest: wait for policy propagation

jtobin:
This should resolve a nagging liquidity edge cases.. uh, edge case, as observed in [this job](https://github.com/lightninglabs/taproot-assets/actions/runs/24673242375/job/72150481956).

Opus's summary:

> The test updates Erin's channel policy, then immediately creates an invoice on Fabia, expecting the route hint to reflect the new policy. But UpdateChannelPolicy returns before the gossip ChannelUpdate has propagated to Fabia's graph, so Fabia may still see the default BaseFeeMsat=1000 instead of 31337.
> 
> Adds assertChannelPolicyUpdate, which polls GetChanInfo on the target node until the peer's policy reflects the expected base fee, and call it before creating the invoice.

### DIFF
--- a/itest/custom_channels/helpers.go
+++ b/itest/custom_channels/helpers.go
@@ -2775,6 +2775,57 @@ func assertNumHtlcsAll(t *testing.T, expected int,
 	}
 }
 
+// assertChannelPolicyUpdate polls GetChanInfo on the given node until
+// the specified peer's policy reflects the expected base fee. This is
+// needed because UpdateChannelPolicy returns before the gossip message
+// has propagated to peers.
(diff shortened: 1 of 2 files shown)


litbot-9000:
@georgetsagk: review reminder

darioAnongba (approved):
Approved but as mentioned, IMO an event should be sent on Fabia's side so she's aware she can proceed. Like `SubscribePolicyUpdates` or something. Maybe we can start looking at these issues as event problems for v0.9 instead of adding more and more "wait.NoError"s

PR #2087, opened 2026-04-21
taprpc: clarify proto docs for OutPoint, ReceiveEvent, ScriptKey

jtobin:
Adds some basic ListAssets-related information to the proto docs, per [this discussion](https://github.com/lightninglabs/taproot-assets/discussions/2083). Just clarifies the OutPoint.txid & ScriptKey.pub_key formats and points to the anchor_outpoint filter.

### DIFF
--- a/taprpc/tapcommon.proto
+++ b/taprpc/tapcommon.proto
@@ -7,7 +7,10 @@ option go_package = "github.com/lightninglabs/taproot-assets/taprpc";
 // Represents a Bitcoin transaction outpoint.
 message OutPoint {
     /*
-    Raw bytes representing the transaction id.
+    Raw bytes representing the transaction id. Must be in
+    internal byte order (little-endian), i.e. reversed
+    compared to the human-readable (RPC/block explorer)
+    hex encoding.
     */
     bytes txid = 1;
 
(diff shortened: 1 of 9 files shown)


taprpc/tapcommon.proto
```diff
@@ -7,7 +7,10 @@ option go_package = "github.com/lightninglabs/taproot-assets/taprpc";
 // Represents a Bitcoin transaction outpoint.
 message OutPoint {
     /*
-    Raw bytes representing the transaction id.
+    Raw bytes representing the transaction id. Must be in
+    internal byte order (little-endian), i.e. reversed
+    compared to the human-readable (RPC/block explorer)
+    hex encoding.
```
darioAnongba, line 13: ok for the doc but isn't this weird?

darioAnongba (approved):

taprpc/tapcommon.proto
jtobin: It is, but I think we can blame Satoshi for that. 😁

darioAnongba: what do you mean? this is the `proto`, which is what we expose externally. No reason not to follow the rest of the industry in terms of encoding. Of course it's out of scope but I don't see why our `txid` fields should be different than block explorers.

GeorgeTsagk: guess it's ok since we're still at the "programmatic" layer? the RPC consumer is another program. Whoever is responsible for rendering characters on the screen is the one to flip.

Again not sure on which layer all software handle this conversion.. Lgtm, cool side-chat to continue at some point

GeorgeTsagk (approved):
lgtm

PR #2088, opened 2026-04-21
tapdb: surface underlying tx retry errors and fix per-attempt rollback scope

darioAnongba:
- **Stop swallowing the underlying cause on tx retry exhaustion.** `ExecTx` previously returned only the opaque `ErrRetriesExceeded` sentinel and logged the underlying serialization/deadlock error at `Trace`, so operators running at default log level had no way to see the real cause. Per-retry logging is now at `Warn` (with the mapped SQL error), and on final exhaustion we `Errorf` and wrap the last underlying error into the returned error via multi-`%w`, so existing `errors.Is(err, ErrRetriesExceeded)` checks keep working while callers also see the root cause.

  This was surfaced by a user hitting `error funding channel: unable to fund vPacket: unable to derive script key: unable to insert script key: db tx retries exceeded` — with nothing actionable in the logs.

- **Scope the tx rollback `defer` per attempt, not per function.** The `defer tx.Rollback()` lived inside the for-loop body, so each retry pushed a new rollback onto `ExecTx`'s defer stack. With the default 10 retries that meant up to 10 stacked `Rollback` calls at function exit, all but one running against an already-closed tx (returning `sql.ErrTxDone`, silently ignored). Harmless today, but wasteful and a latent footgun. Each attempt is now an inner closure so the defer is scoped correctly and runs exactly once per attempt.

### DIFF
--- a/tapdb/interfaces.go
+++ b/tapdb/interfaces.go
@@ -3,6 +3,7 @@ package tapdb
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"math"
 	prand "math/rand"
 	"time"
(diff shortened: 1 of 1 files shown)


jtobin (approved):
LGTM. :+1: :+1:

GeorgeTsagk (approved):
LGTM

PR #2090, opened 2026-04-22
authmailbox: wait for multiSub reconnect in restart test

### LINKED ISSUE #2089: [flake]: TestServerClientAuthAndRestart flake
This test was observed failing flakily in CI in [this job](https://github.com/lightninglabs/taproot-assets/actions/runs/24731239736/job/72444495168). Similar, but not quite the same, as the flake fixed in #2072.

jtobin:
Resolves #2089.

TestServerClientAuthAndRestart published msg2 after waiting for client1 and client2 to reconnect, but did not wait for the MultiSubscription's underlying subscriptions.  Under -race the multiSub could still be mid-handshake when publishMessage fired, silently dropping the message and causing the readMultiSub timeout.

Add AllSubscribed() to MultiSubscription so callers can poll readiness, and use it in the test after server restart.

### DIFF
--- a/authmailbox/client_test.go
+++ b/authmailbox/client_test.go
@@ -221,6 +221,9 @@ func TestServerClientAuthAndRestart(t *testing.T) {
 	harness.Start(t)
 	client1.assertConnected(t)
 	client2.assertConnected(t)
+	require.Eventually(
+		t, allSubscribed(multiSub), testTimeout, testMinBackoff,
+	)
 
 	// Let's send another message to all clients.
 	msg2 := &Message{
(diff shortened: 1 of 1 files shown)


authmailbox/multi_subscription.go
```diff
@@ -123,6 +123,23 @@ func (m *MultiSubscription) Subscribe(ctx context.Context, serverURL url.URL,
...
+	for _, client := range m.clients {
+		for _, sub := range client.subscriptions {
+			if !sub.IsSubscribed() {
+				return false
+			}
+		}
+	}
+
+	return true
+}
```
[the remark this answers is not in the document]
jtobin: This doesn't pose an issue in the sole test in which the helper is used; I argue not worth fixing for speculative defensiveness.

GeorgeTsagk (approved):
LGTM

darioAnongba (approved):

authmailbox/multi_subscription.go
```diff
@@ -123,6 +123,23 @@ func (m *MultiSubscription) Subscribe(ctx context.Context, serverURL url.URL,
 	return nil
 }
 
+// AllSubscribed returns true if every subscription across all clients
+// is currently connected and authenticated.
+func (m *MultiSubscription) AllSubscribed() bool {
```
darioAnongba: is this only used in tests? We shouldn't create functions in the `struct` to help in tests. If this is a test helper, then it should accept a subscription and do the check externally?

darioAnongba (approved):

authmailbox/multi_subscription.go
jtobin: Fixed in updated changeset.

PR #2091, opened 2026-04-22
tapchannel+itest: implement AuxCloseShape and cover the coop close fee baseline

### LINKED ISSUE #2216: chancloser: account for aux close outputs in initial coop close fee baseline
(N.b., just a carbon-copy of the PR text for lightningnetwork/lnd#10969, for the purpose of tracking on the tapd project board.)

Problem
Custom channel coop closes can fail with: unable to process close msg: insufficient fee.

Root cause
For taproot/custom channels, ChanCloser computed the initial coop close fee (idealFeeSat) from only local+remote outputs, but the final close tx can include auxiliary extra outputs.

That mismatch can underprice the first accepted close offer (notably in taproot fast-close), causing mempool rejection and peer disconnect.

Fix
Extend coop close fee weight calculation to include optional extra outputs.

darioAnongba:
Adds the tapd side of lightningnetwork/lnd#10969, which fixes the initial coop close fee baseline for channels with auxiliary close outputs, together with a regression itest.

### AuxCloseShape implementation

`AuxChanCloser.AuxCloseShape` returns the fee-independent shape of the auxiliary close outputs: one P2TR entry per party with a non-zero asset balance. lnd uses this shape to include the aux output weight in the coop close fee baseline, and validates the concrete outputs returned by `AuxCloseOutputs` against it before signing a proposal. The close fee only affects the values of the asset outputs, never their number or script sizes, which is the invariant that makes the shape fee-independent.

The first commit bumps lnd to a master version that contains the new interface method.

### Regression itest

`coop_close_fee_baseline` opens an asset channel and cooperatively closes it at the relay floor fee rate (1 sat/vByte), asserting that the close transaction confirms and carries the expected auxiliary output. At the relay floor there is no headroom: a fee baseline that omits the aux output weight produces a close transaction below the relay fee rate, which the mempool rejects, and the test times out waiting for the close.

Run with `make itest-cc cccase=coop_close_fee_baseline`.


litbot-9000:

GeorgeTsagk (approved):
LGTM
pending LND PR and final `go.mod` replace

PR #2092, opened 2026-04-23
multi: make AddFederationServer idempotent on duplicate

### LINKED ISSUE #2078: AddFederationServer should be idempotent on repeat peer add
## Observation

\`tapd rpc AddFederationServer\` returns

    rpc error: code = Unknown desc = universe server already added

when called with a peer host that has already been registered. There
is no mechanism to \"ensure this peer is registered, whether or not we
added it before.\"

## Impact on consumers

Every client that drives universe federation has to either:

- Call \`ListFederationServers\` first, check whether the peer is
  present, and conditionally call \`AddFederationServer\` (a TOCTOU race
  in the presence of concurrent clients).
- Swallow the specific error string after the call (which the tap-sdk
  regtest harness is now doing — not ideal).

Test harnesses that share a daemon across scenarios (the tap-sdk
itests do) cannot avoid this without either tearing down the daemon
between tests or duplicating the list-then-add logic.

## Proposal

- Make \`AddFederationServer\` idempotent: if the peer is already
  registered, return success (optionally with a warning flag in the
  response).
- Alternatively, return \`codes.AlreadyExists\` instead of
  \`codes.Unknown\` so clients can branch on it cleanly without
  string-matching.

The current \"Unknown\" code with a free-form message is the worst of
both worlds for clients.

jtobin:
Resolves #2078. Opus's summary:

> Use ON CONFLICT (server_host) DO NOTHING in the InsertUniverseServer SQL query so that duplicate peers are silently skipped within the same atomic transaction. This preserves batch atomicity while making the RPC idempotent for already-registered peers.
> 
> Remove the now-dead ErrDuplicateUniverse sentinel and simplify the Start() loop to a single batch AddServer call.

### DIFF
--- a/itest/utils.go
+++ b/itest/utils.go
@@ -908,23 +908,7 @@ func SyncUniverses(ctx context.Context, t *testing.T, clientTapd,
 			},
 		},
 	)
-	if err != nil {
-		// Only fail the test for other errors than duplicate universe
-		// errors, as we might have already added the server in a
-		// previous run.
-		require.ErrorContains(
-			t, err, universe.ErrDuplicateUniverse.Error(),
-		)
-
-		// If we've already added the server in a previous run, we'll
-		// just need to kick off a sync (as that would otherwise be done
-		// by adding the server request already).
-		_, err := clientTapd.SyncUniverse(ctxt, &universerpc.SyncRequest{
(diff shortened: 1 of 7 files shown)


darioAnongba (approved):
Nice one.

PR #2095, opened 2026-04-27
authmailbox: cap subscriber queues and evict slow readers

jtobin:
Adds an optional overflow cap to ConcurrentQueue so that subscriber queues in the authmailbox can be bounded. When a subscriber's queue is full, the oldest item is evicted and the owning stream is aborted, preventing a slow or stalled reader from blocking message delivery to all other subscribers.

### DIFF
--- a/authmailbox/server.go
+++ b/authmailbox/server.go
@@ -740,13 +740,13 @@ func (s *Server) RegisterSubscriber(receiver *fn.EventReceiver[[]*Message],
 // capacity is evicted to prevent unbounded memory growth.
 func (s *Server) publishMessage(msg *Message) {
 	s.msgEventsSubsMtx.Lock()
-	defer s.msgEventsSubsMtx.Unlock()
 
 	var slowSubs []uint64
 	for id, sub := range s.msgEventsSubs {
 		select {
 		case sub.NewItemCreated.ChanIn() <- []*Message{msg}:
 		case <-s.Done():
+			s.msgEventsSubsMtx.Unlock()
 			log.Errorf("Unable publish status event, " +
 				"server shutting down")
 			return
@@ -769,6 +769,34 @@ func (s *Server) publishMessage(msg *Message) {
 
 		sub.Stop()
 		delete(s.msgEventsSubs, id)
+	}
+
+	s.msgEventsSubsMtx.Unlock()
+
+	// Abort the owning streams outside the subscriber lock
+	// so that the handleStream defer (which calls
+	// disconnectClient → RemoveSubscriber) can acquire
+	// msgEventsSubsMtx without deadlocking.
+	if len(slowSubs) > 0 {
+		s.abortSlowStreams(slowSubs)
+	}
+}
+
+// abortSlowStreams finds and disconnects any connected streams
+// whose subscriber ID matches one of the evicted IDs. This
+// closes quitConn, which unblocks the handleStream select and
+// tears down the gRPC stream and its read goroutine.
+func (s *Server) abortSlowStreams(subIDs []uint64) {
+	s.connectedStreamsMtx.Lock()
+	defer s.connectedStreamsMtx.Unlock()
+
+	idSet := fn.NewSet(subIDs...)
+
+	for streamID, stream := range s.connectedStreams {
+		if idSet.Contains(stream.msgReceiver.ID()) {
+			stream.abort()
+			delete(s.connectedStreams, streamID)
+		}
 	}
 }
 
--- a/authmailbox/server.go
+++ b/authmailbox/server.go
@@ -47,6 +47,14 @@ const (
 	// maxRemoveMessageIDs is the maximum number of message IDs that can
 	// be removed in a single RemoveMessage RPC call.
 	maxRemoveMessageIDs = 1000
+
+	// maxSubscriberOverflow is the maximum number of items
+	// allowed to accumulate in a subscriber's overflow queue
+	// before the subscriber is evicted. This is a memory-
+	// safety bound, not a delivery guarantee: messages are
+	// persisted in MsgStore before being published, so an
+	// evicted subscriber recovers via replay on reconnect.
+	maxSubscriberOverflow = 64
 )
 
 // ServerConfig is the configuration struct for the mailbox server. It contains
@@ -726,21 +734,41 @@ func (s *Server) RegisterSubscriber(receiver *fn.EventReceiver[[]*Message],
 	return nil
 }
 
-// publishMessage publishes an event to all status event subscribers (which in
-// this case are all subscribed clients).
+// publishMessage publishes an event to all status event
+// subscribers (which in this case are all subscribed clients).
+// After publishing, any subscriber whose overflow queue is at
+// capacity is evicted to prevent unbounded memory growth.
 func (s *Server) publishMessage(msg *Message) {
-	// Lock the subscriber mutex to ensure that we don't modify the
-	// subscriber map while we're iterating over it.
 	s.msgEventsSubsMtx.Lock()
 	defer s.msgEventsSubsMtx.Unlock()
 
-	for _, sub := range s.msgEventsSubs {
+	var slowSubs []uint64
+	for id, sub := range s.msgEventsSubs {
 		select {
 		case sub.NewItemCreated.ChanIn() <- []*Message{msg}:
 		case <-s.Done():
-			log.Errorf("Unable publish status event, server " +
-				"shutting down")
+			log.Errorf("Unable publish status event, " +
+				"server shutting down")
+			return
 		}
+
+		if sub.NewItemCreated.OverflowLen() >=
+			maxSubscriberOverflow {
+
+			log.Warnf("Evicting slow subscriber %d: "+
+				"overflow queue at capacity", id)
+			slowSubs = append(slowSubs, id)
+		}
+	}
+
+	for _, id := range slowSubs {
+		sub, ok := s.msgEventsSubs[id]
+		if !ok {
+			continue
+		}
+
+		sub.Stop()
+		delete(s.msgEventsSubs, id)
 	}
 }
 
--- a/authmailbox/server.go
+++ b/authmailbox/server.go
@@ -536,7 +536,8 @@ func (s *Server) disconnectClient(stream *mailboxStream) error {
 
 	_, ok := s.connectedStreams[stream.streamID]
 	if !ok {
-		return fmt.Errorf("stream %d not found", stream.streamID)
+		// Already removed (e.g. by slow-subscriber eviction).
+		return nil
 	}
 
 	delete(s.connectedStreams, stream.streamID)
--- a/authmailbox/server.go
+++ b/authmailbox/server.go
@@ -743,12 +743,13 @@ func (s *Server) publishMessage(msg *Message) {
 	s.msgEventsSubsMtx.Lock()
 
 	var slowSubs []uint64
+	msgSlice := []*Message{msg}
 	for id, sub := range s.msgEventsSubs {
 		select {
-		case sub.NewItemCreated.ChanIn() <- []*Message{msg}:
+		case sub.NewItemCreated.ChanIn() <- msgSlice:
 		case <-s.Done():
 			s.msgEventsSubsMtx.Unlock()
-			log.Errorf("Unable publish status event, " +
+			log.Errorf("Unable to publish status event, " +
 				"server shutting down")
 			return
 		}
--- a/fn/queue.go
+++ b/fn/queue.go
@@ -3,13 +3,36 @@ package fn
 import (
 	"container/list"
 	"sync"
+	"sync/atomic"
 )
 
-// ConcurrentQueue is a typed concurrent-safe FIFO queue with unbounded
-// capacity. Clients interact with the queue by pushing items into the in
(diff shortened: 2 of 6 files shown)


authmailbox/server.go
```diff
@@ -726,20 +732,71 @@ func (s *Server) RegisterSubscriber(receiver *fn.EventReceiver[[]*Message],
...
-	// Lock the subscriber mutex to ensure that we don't modify the
-	// subscriber map while we're iterating over it.
 	s.msgEventsSubsMtx.Lock()
-	defer s.msgEventsSubsMtx.Unlock()
 
-	for _, sub := range s.msgEventsSubs {
+	var slowSubs []uint64
+	for id, sub := range s.msgEventsSubs {
 		select {
 		case sub.NewItemCreated.ChanIn() <- []*Message{msg}:
```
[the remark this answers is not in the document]
jtobin: Reasonable.

authmailbox/server.go
```diff
@@ -726,20 +732,71 @@ func (s *Server) RegisterSubscriber(receiver *fn.EventReceiver[[]*Message],
...
 		case <-s.Done():
-			log.Errorf("Unable publish status event, server " +
-				"shutting down")
+			s.msgEventsSubsMtx.Unlock()
+			log.Errorf("Unable publish status event, " +
```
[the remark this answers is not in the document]
jtobin: Pre-existing, but worth fixing.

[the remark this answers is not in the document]
jtobin: Fixed in 2f4279874732a1868c8c1792a4f5ea10273ea469.

authmailbox/server.go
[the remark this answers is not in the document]
jtobin: Fixed in 2f4279874732a1868c8c1792a4f5ea10273ea469.

fn/queue.go
```diff
@@ -116,6 +162,17 @@ func (cq *ConcurrentQueue[T]) start() {
 	}()
 }
 
+// trimOverflow drops the oldest element from the overflow list
+// if maxOverflow is set and the list exceeds the cap.
+func (cq *ConcurrentQueue[T]) trimOverflow() {
+	if cq.maxOverflow > 0 &&
+		cq.overflow.Len() > cq.maxOverflow {
```
Roasbeef, line 169: One issue with this is that misconfig could cause a reader/user to miss out on received funds. 

Though we can pick a same limit here, as we can factor in aspects like:
 * Size of a receiving transaction
 * Number of txns/outputs in a block 


etc

authmailbox/server.go
```diff
@@ -47,6 +47,11 @@ const (
 	// maxRemoveMessageIDs is the maximum number of message IDs that can
 	// be removed in a single RemoveMessage RPC call.
 	maxRemoveMessageIDs = 1000
+
+	// maxSubscriberOverflow is the maximum number of items
+	// allowed to accumulate in a subscriber's overflow queue
+	// before the subscriber is evicted.
+	maxSubscriberOverflow = 1000
```
Roasbeef: Similar tradeoff here. Eg imagine there's a popular addr used for group key sends, and the receiver is offline for a while for w/e reason. 

I think we can guard against this by making sure that the sender clients can handle failures when they attempt to write to a receiver's malibox. So we can return a concrete error that the mailbox is full, then they can enter some back off to wait, then retry, etc.

authmailbox/server.go
```diff
@@ -767,6 +767,37 @@ func (s *Server) publishMessage(msg *Message) {
...
+
+// abortSlowStreams finds and disconnects any connected streams
+// whose subscriber ID matches one of the evicted IDs. This
+// closes quitConn, which unblocks the handleStream select and
+// tears down the gRPC stream and its read goroutine.
+func (s *Server) abortSlowStreams(subIDs []uint64) {
+	s.connectedStreamsMtx.Lock()
+	defer s.connectedStreamsMtx.Unlock()
+
+	idSet := make(map[uint64]struct{}, len(subIDs))
```
Roasbeef: Can do `fn.NewSet`.

Roasbeef (commented):
Good defense change. 

One thing I think we want to verify is that senders won't just fall over if they attempt to deliver an addr recv notification to a sender, but their mailbox is full. 

We may not need to reach for full blown itests here, if we can create sufficiently isolated yet dedicated unit tests.

fn/queue.go
jtobin: This comment prompted some very profitable exploration.

IIUC, subscribers can't truly miss out on received funds, at least not permanently. They can only miss out temporarily if they're too slow to receive notifications, at which point, if they've failed to read enough of them, they get booted. But they can always merely reconnect and pick up from where they left off; the messages themselves were persisted before notifications were sent out on the queue, so they're still available, and will be sent out again during replay.

So I think that's all acceptable (indeed, I see no other solution), and it means the the cap should be treated solely as a device to limit memory pressure, rather than to also guarantee delivery or accommodate throughput or something. But an interesting thing I realised is that a cap of 1000 is actually way too high, in that 1000 queued messages would create _way_ too much memory pressure if actually realised. The number should actually be much lower, since, morally, we just need to accommodate short spikes that could occur before subscribers process new messages.

Opus was recommending something like 64 or 128 for the overflow cap, and calculated that the worst-case memory pressure for 64 items is maybe 600 MB, so 1.2 GB for 128 items. The cap of 128 thus seems tolerable, but Opus actually made a pretty compelling case to ~~not be such a wimp~~ go straight to 64:

```
The queue is a burst buffer — it just needs to absorb temporary
spikes between when messages arrive and when the client drains them. So
the relevant metrics are:

1. Memory budget per dead subscriber. Each queue item is a []*Message
slice. For live publishes, that's a one-element slice. Worst case per
message is ~9.4 MiB (perhaps 64 KiB payload + overhead). At 64
items that's ~600 MiB ceiling; at 128, ~1.2 GiB. For typical payloads
it'd be much less, but you want the worst case to be survivable.

2. Publish burst rate. How many messages can arrive in the window
between the client falling behind and the next publishMessage checking
overflow? That's basically "how many SendMessage calls can complete
before the client gets a chance to drain." Since each SendMessage
requires a verified UTXO proof, there's natural rate-limiting — you
can't spray thousands instantly.

3. Client drain latency. How slow can a healthy-but-momentarily-busy
client be? Network jitter, GC pauses, a slow RPC round-trip — these
cause transient backups. You want the cap high enough that a healthy
client experiencing a hiccup doesn't get evicted. A few seconds of
backup at realistic message rates is unlikely to exceed even 64.

64 is conservative: small memory footprint, and any legitimate burst
that exceeds 64 queued items is almost certainly a stalled client,
not a momentary hiccup. 128 gives more headroom if you're worried
about being too aggressive with eviction, at the cost of doubling the
worst-case memory exposure.
```

I think that's all exactly on the mark. Again: if the subscriber is stalled, the subscriber is stalled, and bumping the overflow cap isn't actually going to do anything useful for them anyway.

So yeah: I think 64 is the cap to go with.

authmailbox/server.go
jtobin: Yeah, per the rabbit hole exploration in my previous comment: it's not the mailbox itself (viz. MsgStore, in terms of the persisted state) that's full, it's just the notification queue, which again exists solely to limit memory pressure on the server. I think the question is again really "how low should the cap be?"

authmailbox/server.go
jtobin: Fixed in updated changeset.

fn/queue.go
jtobin: I've pushed another version of these commits that sets the cap to 64, if you agree with the above reasoning. Merely LMK if something still seems off.

litbot-9000:
@roasbeef: review reminder

fn/queue.go
darioAnongba: looks addressed.

authmailbox/server.go
darioAnongba: addressed it seems but missing a test.

darioAnongba:
LGTM, but as per @Roasbeef request, I'd add a targeted unit test around `ConcurrentQueue` and maybe a small itest in `authmailbox/server_test.go` to assert eviction?

authmailbox/server.go
jtobin: Added a test in the latest changeset (to 8be183934).

darioAnongba (changes requested):
Was thinking more of a test in `server_test.go` testing the actual feature on a `Server`. Got this test written and passing locally, I checked that it successfully tests the eviction:

```go
// registerTestStream constructs a mailboxStream wired the same way as the
// live ReceiveMessages path: registered both as a connected stream and as a
// message subscriber, with the maxSubscriberOverflow cap applied.
func registerTestStream(t *testing.T, srv *Server,
	streamID uint64) *mailboxStream {

	t.Helper()

	stream, err := newMailboxStream(streamID, srv.Done())
	require.NoError(t, err)

	srv.connectedStreamsMtx.Lock()
	srv.connectedStreams[streamID] = stream
	srv.connectedStreamsMtx.Unlock()

	err = srv.RegisterSubscriber(
		stream.msgReceiver, false, MessageFilter{},
	)
	require.NoError(t, err)

	return stream
}

// TestSlowSubscriberEviction publishMessage must not stall when one subscriber is
// not draining its queue. After enough publishes to fill the stalled
// subscriber's overflow cap, the stalled subscriber must be evicted, its
// owning stream torn down, and a healthy subscriber must continue receiving
// every message.
func TestSlowSubscriberEviction(t *testing.T) {
	t.Parallel()

	srv := &Server{
		msgEventsSubs: make(
			map[uint64]*fn.EventReceiver[[]*Message],
		),
		connectedStreams: make(map[uint64]*mailboxStream),
		cfg:              &ServerConfig{MsgStore: NewMockStore()},
		ContextGuard:     lfn.NewContextGuard(),
	}
	t.Cleanup(func() { _ = srv.Stop() })

	healthy := registerTestStream(t, srv, 1)
	stalled := registerTestStream(t, srv, 2)

	// Drain the healthy subscriber as fast as the queue produces.
	// The drainer uses a stop signal because the queue's chanOut is
	// not closed by Stop(), so a `for ... range` would hang at
	// teardown.
	var healthyReceived atomic.Int64
	drainerStarted := make(chan struct{})
	drainerDone := make(chan struct{})
	go func() {
		close(drainerStarted)
		for {
			select {
			case msgs := <-healthy.msgReceiver.NewItemCreated.
				ChanOut():

				healthyReceived.Add(int64(len(msgs)))

			case <-drainerDone:
				return
			}
		}
	}()
	t.Cleanup(func() { close(drainerDone) })
	<-drainerStarted

	// Publish enough messages to overflow the stalled subscriber:
	// chanOut buffer + cap + slack so we comfortably trip the
	// >= maxSubscriberOverflow threshold.
	totalPublishes := fn.DefaultQueueSize + maxSubscriberOverflow + 5
	receiverKey := test.RandPubKey(t)

	publishDone := make(chan struct{})
	go func() {
		defer close(publishDone)
		for i := 0; i < totalPublishes; i++ {
			srv.publishMessage(&Message{
				ReceiverKey:      *receiverKey,
				EncryptedPayload: []byte("payload"),
				ArrivalTimestamp: time.Now(),
			})
		}
	}()

	// The publish loop must complete promptly. If a slow subscriber
	// could stall publishMessage, this is where the test would hang.
	select {
	case <-publishDone:
	case <-time.After(5 * time.Second):
		t.Fatal("publishMessage stalled on a slow subscriber")
	}

	// The stalled subscriber must be evicted from the subscriber
	// map. Eviction happens after the publishMessage call that
	// tripped the cap, so allow a brief settle window.
	require.Eventually(t, func() bool {
		srv.msgEventsSubsMtx.Lock()
		_, present := srv.msgEventsSubs[stalled.msgReceiver.ID()]
		srv.msgEventsSubsMtx.Unlock()
		return !present
	}, 2*time.Second, 10*time.Millisecond,
		"stalled subscriber was not evicted")

	// abortSlowStreams must remove the stalled stream from
	// connectedStreams and close its quitConn.
	srv.connectedStreamsMtx.Lock()
	_, stillConnected := srv.connectedStreams[stalled.streamID]
	srv.connectedStreamsMtx.Unlock()
	require.False(t, stillConnected,
		"stalled stream still in connectedStreams")

	select {
	case <-stalled.quitConn:
	default:
		t.Fatal("stalled stream quitConn was not closed")
	}

	// The healthy subscriber must still be registered and must
	// have received every published message.
	srv.msgEventsSubsMtx.Lock()
	_, healthyPresent := srv.msgEventsSubs[healthy.msgReceiver.ID()]
	srv.msgEventsSubsMtx.Unlock()
	require.True(t, healthyPresent,
		"healthy subscriber was incorrectly evicted")

	require.Eventually(t, func() bool {
		return healthyReceived.Load() == int64(totalPublishes)
	}, 2*time.Second, 10*time.Millisecond,
		"healthy subscriber received %d / %d messages",
		healthyReceived.Load(), totalPublishes)
}
```

jtobin:
> Was thinking more of a test in `server_test.go` testing the actual feature on a `Server`. Got this test written and passing locally, I checked that it successfully tests the eviction:

Nice, added on the top commit. 🤜 🤛

darioAnongba (approved):
LGTM

PR #2096, opened 2026-04-27
authmailbox: cap concurrent ReceiveMessages streams per peer

jtobin:
Caps the number of concurrent ReceiveMessages streams a single peer can hold open. Streams are tracked per-IP and new connections beyond the limit are rejected immediately.

Configurable via `--universe.mbox-max-streams` and `--universe.mbox-max-streams-per-peer`.

### DIFF
--- a/authmailbox/server.go
+++ b/authmailbox/server.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
+	"net/netip"
 	"slices"
 	"sync"
 	"sync/atomic"
@@ -25,6 +27,7 @@ import (
 	lfn "github.com/lightningnetwork/lnd/fn/v2"
 	"golang.org/x/exp/maps"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/peer"
 )
 
 const (
@@ -91,6 +94,11 @@ type ServerConfig struct {
 	// CleanupCheckTimeout is the per-outpoint timeout when checking if
 	// an outpoint has been spent. Defaults to 30s if zero.
 	CleanupCheckTimeout time.Duration
+
+	// MaxStreamsPerPeer is the maximum number of concurrent
+	// ReceiveMessages streams from a single peer (by IP). A
+	// zero value means no limit.
+	MaxStreamsPerPeer uint
 }
 
 // Server is the mailbox server that handles incoming messages from clients and
@@ -119,6 +127,13 @@ type Server struct {
 	// msgEventsSubsMtx guards the general message events subscribers map.
 	msgEventsSubsMtx sync.Mutex
 
+	// peerStreamCount tracks the number of active streams per
+	// peer IP address.
+	peerStreamCount map[netip.Addr]uint
+
+	// peerStreamCountMtx guards peerStreamCount.
+	peerStreamCountMtx sync.Mutex
+
 	*lfn.ContextGuard
 }
 
@@ -129,6 +144,7 @@ func NewServer() *Server {
 			map[uint64]*fn.EventReceiver[[]*Message],
 		),
 		connectedStreams: make(map[uint64]*mailboxStream),
+		peerStreamCount:  make(map[netip.Addr]uint),
 		ContextGuard:     lfn.NewContextGuard(),
 	}
 }
@@ -338,6 +354,45 @@ func (s *Server) SendMessage(ctx context.Context,
 	}, nil
 }
 
+// peerAddr extracts the IP address (without port) from a gRPC
+// stream context. IPv4-mapped IPv6 addresses are collapsed via
+// Unmap so dual-stack clients share a single bucket. IPv6
+// addresses are truncated to a /64 prefix, since the lower 64
+// bits are host-controlled and trivially rotated. Returns the
+// zero Addr if the peer address is unavailable or not TCP.
+func peerAddr(ctx context.Context) netip.Addr {
+	p, ok := peer.FromContext(ctx)
+	if !ok || p.Addr == nil {
+		return netip.Addr{}
+	}
+
+	tcpAddr, ok := p.Addr.(*net.TCPAddr)
+	if !ok {
+		return netip.Addr{}
+	}
+
+	a, ok := netip.AddrFromSlice(tcpAddr.IP)
+	if !ok {
+		return netip.Addr{}
+	}
+
+	// Collapse IPv4-mapped IPv6 (::ffff:1.2.3.4 -> 1.2.3.4)
+	// so a dual-stack client doesn't get two budgets.
+	a = a.Unmap()
+
+	// The lower 64 bits of an IPv6 address are
+	// host-controlled and trivially rotated by an attacker
+	// with a routed /64. Bucket per /64.
+	if a.Is6() {
+		pfx, err := a.Prefix(64)
+		if err == nil {
+			return pfx.Addr()
+		}
+	}
+
+	return a
+}
+
 // ReceiveMessages initiates a bidirectional stream to receive messages for a
 // specific receiver. This stream implements the challenge-response handshake
 // required for receiver authentication before messages are delivered.
@@ -351,13 +406,40 @@ func (s *Server) SendMessage(ctx context.Context,
 //  5. Server -> Client: ReceiveMessagesResponse(eos = EndOfStream{}) or
 //     ReceiveMessagesResponse(error = ReceiveError{...})
 func (s *Server) ReceiveMessages(grpcStream serverStream) error {
+	ctx := grpcStream.Context()
+	peerIP := peerAddr(ctx)
+
+	if !peerIP.IsValid() {
+		log.DebugS(ctx, "Could not extract peer IP, "+
+			"skipping per-peer stream limit")
+	}
+
+	// Enforce per-peer stream limit.
+	if peerIP.IsValid() && s.cfg.MaxStreamsPerPeer > 0 {
+		s.peerStreamCountMtx.Lock()
+		if s.peerStreamCount[peerIP] >= s.cfg.MaxStreamsPerPeer {
+			s.peerStreamCountMtx.Unlock()
+
+			log.WarnS(ctx, "Max streams reached "+
+				"for peer", nil, "peer", peerIP)
+
+			return fmt.Errorf("max streams for peer %s "+
+				"reached", peerIP)
+		}
+		s.peerStreamCount[peerIP]++
+		s.peerStreamCountMtx.Unlock()
+
+		defer s.decrementPeerCount(peerIP)
+	}
+
 	id := s.nextStreamID.Add(1) - 1
 
-	ctxl := btclog.WithCtx(grpcStream.Context(), "sid", id, "server", true)
+	ctxl := btclog.WithCtx(ctx, "sid", id, "server", true)
 
 	stream, err := newMailboxStream(id, s.Done())
 	if err != nil {
-		return fmt.Errorf("error creating mailbox stream: %w", err)
+		return fmt.Errorf("error creating mailbox stream: %w",
+			err)
 	}
 
 	s.connectedStreamsMtx.Lock()
@@ -367,6 +449,23 @@ func (s *Server) ReceiveMessages(grpcStream serverStream) error {
 	log.DebugS(ctxl, "New ReceiveMessages stream created")
 
 	return s.handleStream(ctxl, grpcStream, stream)
+}
+
+// decrementPeerCount decreases the stream count for the given
+// peer IP. If the count reaches zero, the entry is removed.
+func (s *Server) decrementPeerCount(peerIP netip.Addr) {
+	if !peerIP.IsValid() {
+		return
+	}
+
+	s.peerStreamCountMtx.Lock()
+	defer s.peerStreamCountMtx.Unlock()
+
+	if s.peerStreamCount[peerIP] <= 1 {
+		delete(s.peerStreamCount, peerIP)
+	} else {
+		s.peerStreamCount[peerIP]--
+	}
 }
 
 // MailboxInfo returns basic server information.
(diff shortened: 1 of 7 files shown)


authmailbox/server.go
```diff
@@ -343,23 +380,68 @@ func (s *Server) SendMessage(ctx context.Context,
...
 	s.connectedStreamsMtx.Unlock()
 
 	log.DebugS(ctxl, "New ReceiveMessages stream created")
 
-	return s.handleStream(ctxl, grpcStream, stream)
+	err = s.handleStream(ctxl, grpcStream, stream)
+
+	s.decrementPeerCount(peerIP)
+
+	return err
```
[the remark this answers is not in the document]
jtobin: Fixed in a114cd4d1df5d26c5360cdb68df74dc608efda9f.

authmailbox/server.go
```diff
@@ -111,6 +124,13 @@ type Server struct {
 	// msgEventsSubsMtx guards the general message events subscribers map.
 	msgEventsSubsMtx sync.Mutex
 
+	// peerStreamCount tracks the number of active streams
+	// per peer IP address.
+	peerStreamCount map[string]int
```
Roasbeef: https://pkg.go.dev/net/netip#Addr ?

authmailbox/server.go
```diff
@@ -330,6 +351,22 @@ func (s *Server) SendMessage(ctx context.Context,
 	}, nil
 }
 
+// peerAddr extracts the IP address (without port) from a
+// gRPC stream context. Returns "" if unavailable.
+func peerAddr(ctx context.Context) string {
+	p, ok := peer.FromContext(ctx)
```
Roasbeef, line 364: Cool, TIL the IP is already in a context here.

tapcfg/config.go
```diff
@@ -151,6 +151,14 @@ const (
 	// messages.
 	defaultMboxCleanupInterval = 24 * time.Hour
 
+	// defaultMaxMboxStreams is the default maximum number of
+	// concurrent ReceiveMessages streams.
+	defaultMaxMboxStreams = 1000
+
+	// defaultMaxMboxStreamsPerPeer is the default maximum number
+	// of concurrent ReceiveMessages streams from a single peer.
+	defaultMaxMboxStreamsPerPeer = 10
```
Roasbeef, line 156: One thing to factor in here is there's a sort of TTL or keep alive on the subscriptions. Depending on the underlying mechanics, we could risk false positives for roaming mobile clients.

tapcfg/config.go
```diff
@@ -151,6 +151,14 @@ const (
 	// messages.
 	defaultMboxCleanupInterval = 24 * time.Hour
 
+	// defaultMaxMboxStreams is the default maximum number of
+	// concurrent ReceiveMessages streams.
+	defaultMaxMboxStreams = 1000
```
Roasbeef: Hard cap here would require a restart (thus dropping all clients) to scale. Worth factoring in that any tapd client that makes a group key addr would end up listening on this. 

Perhaps the IP cap is enough? Can even broaden it to CIDR ranges.

authmailbox/server.go
```diff
@@ -417,7 +407,15 @@ func (s *Server) ReceiveMessages(grpcStream serverStream) error {
...
 
+	// Enforce global stream limit and register atomically.
 	s.connectedStreamsMtx.Lock()
+	if s.cfg.MaxStreams > 0 &&
+		len(s.connectedStreams) >= s.cfg.MaxStreams {
+
+		s.connectedStreamsMtx.Unlock()
+		s.decrementPeerCount(peerIP)
+		return fmt.Errorf("max concurrent streams reached")
+	}
```
Roasbeef: Can fold into the prior commit.

authmailbox/server.go
```diff
@@ -393,6 +393,8 @@ func (s *Server) ReceiveMessages(grpcStream serverStream) error {
 		}
 		s.peerStreamCount[peerIP]++
 		s.peerStreamCountMtx.Unlock()
+
+		defer s.decrementPeerCount(peerIP)
```
Roasbeef, line 432: Same here re fold in.

tapcfg/config.go
jtobin: Good point. After considering it further: yeah, let's just retain the per-IP cap. It's the simplest solution and is the 80/20 (if not 90/10+) thing to do.

jtobin: There appears to be a nice follow-up refactoring that would dovetail well with this that involves deferring the stream allocation until later in the auth process, such that we don't allocate one at all for unauthenticated clients. Will make in a separate PR.

tapcfg/config.go
jtobin: This becomes a non-issue sans a global cap, so yeah, another reason to ditch it.

authmailbox/server.go
jtobin: Fixed in updated changeset.

tapcfg/config.go
jtobin: I've ditched the global cap in the updated changeset.

litbot-9000:
@roasbeef: review reminder

authmailbox/server.go
```diff
@@ -343,14 +374,32 @@ func (s *Server) SendMessage(ctx context.Context,
...
+				"reached", peerIP)
+		}
+		s.peerStreamCount[peerIP]++
+		s.peerStreamCountMtx.Unlock()
+
+		defer s.decrementPeerCount(peerIP)
+	}
+
 	id := s.nextStreamID.Load()
 	_ = s.nextStreamID.Add(1)
```
darioAnongba: could quickly change this while we're at it to:
```suggestion
id := s.nextStreamID.Add(1) - 1
```

Two concurrent calls can read the same `id` and both register in `connectedStreams[id]`, with the second silently overwriting the first

authmailbox/server.go
```diff
@@ -330,6 +345,22 @@ func (s *Server) SendMessage(ctx context.Context,
...
+// gRPC stream context. Returns the zero Addr if unavailable.
+func peerAddr(ctx context.Context) netip.Addr {
+	p, ok := peer.FromContext(ctx)
+	if !ok || p.Addr == nil {
+		return netip.Addr{}
+	}
+
+	ap, err := netip.ParseAddrPort(p.Addr.String())
+	if err != nil {
+		return netip.Addr{}
```
darioAnongba, line 376: `netip.ParseAddrPort` requires `host:port` form. Anything else will return "not an ip:port" error → zero `Addr` → `IsValid()=false` → limit-check branch is skipped.

Works today because we only support TCP but won't this fail for unix-socket (something LND supports btw), or other addr types? Looks like we can branch on `p.Addr.(type)` so we can differentiate `TCPAddr` from others.

darioAnongba, line 376: Missing log statement (Warn ideally?)

authmailbox/server.go
```diff
@@ -330,6 +345,22 @@ func (s *Server) SendMessage(ctx context.Context,
 	}, nil
 }
 
+// peerAddr extracts the IP address (without port) from a
+// gRPC stream context. Returns the zero Addr if unavailable.
+func peerAddr(ctx context.Context) netip.Addr {
+	p, ok := peer.FromContext(ctx)
+	if !ok || p.Addr == nil {
+		return netip.Addr{}
```
darioAnongba, line 366: Missing log statement (Debug, Warn?)

authmailbox/server.go
```diff
@@ -343,14 +374,32 @@ func (s *Server) SendMessage(ctx context.Context,
...
 //     ReceiveMessagesResponse(error = ReceiveError{...})
 func (s *Server) ReceiveMessages(grpcStream serverStream) error {
+	ctx := grpcStream.Context()
+	peerIP := peerAddr(ctx)
+
+	// Enforce per-peer stream limit.
+	if peerIP.IsValid() && s.cfg.MaxStreamsPerPeer > 0 {
+		s.peerStreamCountMtx.Lock()
+		if s.peerStreamCount[peerIP] >= s.cfg.MaxStreamsPerPeer {
+			s.peerStreamCountMtx.Unlock()
```
darioAnongba, line 421: Missing log here as well?

authmailbox/server.go
```diff
@@ -330,6 +345,22 @@ func (s *Server) SendMessage(ctx context.Context,
 	}, nil
 }
 
+// peerAddr extracts the IP address (without port) from a
+// gRPC stream context. Returns the zero Addr if unavailable.
+func peerAddr(ctx context.Context) netip.Addr {
+	p, ok := peer.FromContext(ctx)
```
darioAnongba, line 364: For Websocket/REST, the gRPC server will see `peer.Addr` as`127.0.0.1:<port>` so every websocket clients will end up sharing the same bucket of 10 (default). Unlikely that was intended?

For example, someone can spin up a node at Voltage to use Rails, so Voltage will connect to it, Amboss, adn the client, 3 different IPs through Websocket that will reduce to `127.0.0.1`.

tapcfg/config.go
```diff
@@ -936,6 +943,11 @@ func ValidateConfig(cfg Config, cfgLogger btclog.Logger) (*Config, error) {
 		}
 	}
 
+	// Reject negative mailbox stream limit.
+	if cfg.Universe.MboxMaxStreamsPerPeer < 0 {
```
darioAnongba: not needed if uint.

tapcfg/config.go
```diff
@@ -359,6 +363,8 @@ type UniverseConfig struct {
 
 	MboxCleanupCheckTimeout time.Duration `long:"mbox-cleanup-check-timeout" description:"Per-outpoint timeout when checking if an outpoint has been spent on chain. Valid time units are {s, m, h}."`
 
+	MboxMaxStreamsPerPeer int `long:"mbox-max-streams-per-peer" description:"Maximum concurrent mailbox streams from a single peer IP. 0 means unlimited."`
```
darioAnongba: uint?

authmailbox/server.go
```diff
@@ -83,6 +85,11 @@ type ServerConfig struct {
 	// CleanupCheckTimeout is the per-outpoint timeout when checking if
 	// an outpoint has been spent. Defaults to 30s if zero.
 	CleanupCheckTimeout time.Duration
+
+	// MaxStreamsPerPeer is the maximum number of concurrent
+	// ReceiveMessages streams from a single peer (by IP).
+	// A zero value means no limit.
```
darioAnongba: why are the comments breaking so quickly instead of 80 chars?

authmailbox/server.go
```diff
@@ -111,6 +118,13 @@ type Server struct {
 	// msgEventsSubsMtx guards the general message events subscribers map.
 	msgEventsSubsMtx sync.Mutex
 
+	// peerStreamCount tracks the number of active streams
+	// per peer IP address.
```
darioAnongba: why are the comments breaking so quickly instead of 80 chars?

authmailbox/server.go
```diff
@@ -330,6 +345,22 @@ func (s *Server) SendMessage(ctx context.Context,
...
+	p, ok := peer.FromContext(ctx)
+	if !ok || p.Addr == nil {
+		return netip.Addr{}
+	}
+
+	ap, err := netip.ParseAddrPort(p.Addr.String())
+	if err != nil {
+		return netip.Addr{}
+	}
+
```
darioAnongba, line 378: Can add an IPv6 improvement, otherwise it's trivial for someone to bypass the "X streams per peer".

```suggestion
    // Collapse IPv4-mapped IPv6 (::ffff:1.2.3.4 → 1.2.3.4) so a
    // dual-stack client doesn't get two budgets.
    a := ap.Addr().Unmap()

    // The lower 64 bits of an IPv6 address are host-controlled and
    // trivially rotated by an attacker with a routed /64 (the
    // standard residential / cloud-VPS allocation). Bucket per /64.
    if a.Is6() {
        pfx, err := a.Prefix(64)
        if err == nil {
            return pfx.Addr()
        }
    }
    return a
```

darioAnongba (commented):
Generally speaking, I wouldn't go ahead with this as I don't think it's the right fix because `tapd` should not be responsible for this filtering IMO. Might seem simple to filter by IP address but in practice it's very complex and better solved by putting a proxy in front of `tapd` like nginx. You usually don't implement this type of things directly on the daemon.

In any case, the biggest concern for me in the PR are:
- No tests (unit or itests)
- Most likely we didn't think about IPv6 or the fact that attackers most likely have a big range of IP addresses available so this guard is trivially bypassed
- Websocket/REST differences

authmailbox/server.go
jtobin: Agreed that branching on p.Addr.(type) is cleaner, will fix.

authmailbox/server.go
jtobin: This one is a false positive; ReceiveMessages is long-lived / bidirectional / gRPC only. It can't be hit via websockets.

authmailbox/server.go
jtobin: Good question, and not sure..

authmailbox/server.go
jtobin: Fixed in updated changeset.

authmailbox/server.go
jtobin: Fixed in updated changeset.

jtobin: Fixed in updated changeset.

tapcfg/config.go
jtobin: Fixed in updated changeset.

authmailbox/server.go
jtobin: Fixed in updated changeset.

authmailbox/server.go
jtobin: Fixed in updated changeset.

authmailbox/server.go
darioAnongba: WDYM? Every bidirectional gRPC stream is converted into a Websocket endpoint automatically. https://lightning.engineering/api-docs/api/taproot-assets/mailbox/receive-messages/

jtobin: Derp. My bad. I thought this one was intentionally lacking a HTTP annotation, i.e. explicitly gRPC-only, but I see it now:

```
    - selector: authmailboxrpc.Mailbox.ReceiveMessages
      post: "/v1/taproot-assets/mailbox/receive"
      body: "*"
```

jtobin:
Closing, as I've been convinced that this is better handled at the transport layer.

PR #2097, opened 2026-04-27
asset: validate element count in AltLeavesDecoder

jtobin:
Simply adds a validation of the element count varint in AltLeavesDecoder against the remaining TLV record length.


Roasbeef (approved):
LGTM 🫛

PR #2098, opened 2026-04-27
universe: replace backend map with LRU cache

jtobin:
Replaces the unbounded baseUniverses map with an LRU cache capped at 10,000 entries, and add a universe existence check in FetchLeaves before allocating a cached backend.


Roasbeef (approved):
LGTM 👢

PR #2100, opened 2026-04-27
multi: add pagination support to AssetLeaves RPC

jtobin:
Adds pagination support (offset, limit, direction) to the AssetLeaves RPC endpoint, and adds MaxPageSize validation to AssetRoots. `tapcli universe leaves` now paginates automatically, fetching all pages instead of silently truncating at 512 results.

Also fixes an inverted-sort-direction bug that was present in AssetRoots, AssetLeafKeys, and QueryEvents.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -346,6 +346,11 @@
 - [PR#2100](https://github.com/lightninglabs/taproot-assets/pull/2100)
   Add pagination support (offset, limit, direction) to the `AssetLeaves`
   RPC endpoint, and add `MaxPageSize` validation to `AssetRoots`.
+  Standardize pagination validation across `AssetRoots`,
+  `AssetLeafKeys`, `AssetLeaves`, and `QueryAssetStats` via a shared
+  `validatePage` helper, and add a `has_more` field to all four
+  response types. Default `limit=0` to `MaxPageSize` instead of
+  `RequestPageSize`.
 
 
 ## tapcli Updates
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -98,6 +98,10 @@
   fixes a bug by which grouped assets could remain invisible to
   ListGroups after import, remaining visible only in ListAssets.
 
+* [PR#2100](https://github.com/lightninglabs/taproot-assets/pull/2100)
+  fixes inverted sort direction in `AssetRoots`, `AssetLeafKeys`, and
+  `QueryEvents` universe RPCs.
+
 # New Features
 
 ## Functional Enhancements
@@ -339,11 +343,20 @@
   `SendEvent`, allowing clients to distinguish grouped fungible assets from
   grouped collectible assets.
 
+- [PR#2100](https://github.com/lightninglabs/taproot-assets/pull/2100)
+  Add pagination support (offset, limit, direction) to the `AssetLeaves`
+  RPC endpoint, and add `MaxPageSize` validation to `AssetRoots`.
+
+
 ## tapcli Updates
 
 - [PR#1995](https://github.com/lightninglabs/taproot-assets/pull/1995)
   Add `--limit`, `--offset`, and `--direction` flags to `tapcli assets list`
   for pagination support. The direction defaults to descending order.
+
+- [PR#2100](https://github.com/lightninglabs/taproot-assets/pull/2100)
+  `tapcli universe leaves` now paginates automatically, fetching all
+  pages instead of silently truncating at 512 results.
 
 ## Config Changes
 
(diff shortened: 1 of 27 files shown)


taprpc/universerpc/universe.proto
```diff
@@ -57,7 +57,7 @@ service Universe {
     took place on chain. The leaves contain a normal Taproot Asset proof, as
     well as details for the asset.
     */
-    rpc AssetLeaves (ID) returns (AssetLeafResponse);
+    rpc AssetLeaves (AssetLeavesRequest) returns (AssetLeafResponse);
```
darioAnongba, line 60: Breaking change but fine by me since I believe no-one uses this yet?

docs/release-notes/release-notes-0.8.0.md
```diff
@@ -94,6 +94,10 @@
   fixes `DecodeAssetPayReq` so `GenesisInfo` is populated consistently,
   including group-key invoice decodes.
 
+* [PR#2100](https://github.com/lightninglabs/taproot-assets/pull/2100)
```
darioAnongba, line 101: Also a breaking change but fine I guess since it was a bug?

tapdb/universe.go
```diff
@@ -1273,21 +1281,26 @@ func (b *BaseUniverseTree) FetchKeys(ctx context.Context,
...
 	var leaves []universe.Leaf
 
 	readTx := NewBaseUniverseReadTx()
 	dbErr := b.db.ExecTx(ctx, &readTx, func(db BaseUniverseStore) error {
-		// First, we'll query the set of Universe leaves we have
-		// directly to determine which ones we care about. We only
-		// filter on the namespace here, as we want all the leaves for
-		// this tree.
+		limit := q.Limit
+		if limit == 0 {
```
darioAnongba, line 1294: don't we want if limit == 0 to return `MaxPageSize`? Otherwise we have an inconsistency that was solved in the CLI:
> instead of silently truncating at 512 results.

Otherwise we need a ``total_count` / `has_more` on the response (better to check how other endpoints behave and copy that).

taprpc/universerpc/universe.proto
darioAnongba, line 411: The endpoints don't seem to have a termination flag like `total_count` / `has_more` so a user will most likely write `if len(page.Leaves) < pageSize { break }`, which is wrong.

How are other endpoints doing this? it seems to be missing in `AssetLeafResponse`, `AssetRootResponse` and `AssetLeafKeyResponse`

darioAnongba (changes requested):
Would be nice to have a consistent way of handling pagination across `AssetRoots`, `AssetLeafKeys`, `AssetLeaves` and `QueryAssetStats`.

Like a `validatePage(offset, limit int32) error` or something. This is what I managed to check with an agent:

| RPC               | `Limit` range checked | Negative `Offset` rejected |
|-------------------|-----------------------|----------------------------|
| `AssetRoots`      | yes (added in PR)     | no                         |
| `AssetLeafKeys`   | yes (pre-existing)    | no                         |
| `AssetLeaves`     | yes (added in PR)     | yes (added in PR)          |
| `QueryAssetStats` | no                    | no                         |

Negative `Offset` to Postgres errors out (`OFFSET must not be negative`); SQLite silently treats it as 0. `QueryAssetStats` ignores `Limit` entirely (no `MaxPageSize` cap), allowing very large result requests.

docs/release-notes/release-notes-0.8.0.md
jtobin: Yeah, we should call it out prominently in the release notes.

taprpc/universerpc/universe.proto
jtobin: Yeah, Opus confirms that this is a gap across all the paginated universe endpoints. We should fix them all here.

docs/release-notes/release-notes-0.8.0.md
jtobin: (I intend to revamp the release notes significantly before release, so will batch this and other similar changes for then.)

tapdb/universe.go
jtobin: Fixed in updated changeset. Specifically: "limit" now defaults to RequestPageSize, i.e. 512, in particular, which is consistent across all endpoints (MaxPageSize is pretty big, at 16384).

I also changed the default limit of QuerySyncStats from math.MaxInt32 to RequestPageSize to make it consistent. The previous limit had been in there since the code was originally written, not seemingly for any particular reason.

taprpc/universerpc/universe.proto
jtobin: Added `has_more` here, and everywhere else, in the updated changeset.

jtobin:
> Would be nice to have a consistent way of handling pagination across AssetRoots, AssetLeafKeys, AssetLeaves and QueryAssetStats.
> 
> Like a validatePage(offset, limit int32) error or something. 

Went for exactly this afterwards; everything is now treated with such a validatePage helper.

darioAnongba (approved):
LGTM, v0.8 or v0.9? Since we've already cut. The changelog update is on v0.8, this fixes a bug if I remember correctly so could go into v0.8 as a feature-fix.

jtobin:
Yup, v0.8 in my books as it also relates to some buggy stuff.

Issue #2108, opened 2026-04-30
[bug]: Closed channel status `ChanStatusCoopBroadcasted|ChanStatusRemoteCloseInitiator` without `closing_txid`


bufo24:
### Background

My peer did a coop closure of an asset channel but no closing tx was broadcasted.

### Your environment

Got the issue on both:

```sh
- 0.7.0-alpha
- 0.7.99-alpha commit=v0.7.1-0.20260417020815-5e6d2a62bc1b
```

Peer was running `0.7.0-alpha`

### Steps to reproduce

Not exactly sure what the root cause is here, I had trouble reproducing it. Although when I was playing around with this I found out that a failing proof courier might be of impact here, although in my mainnet scenario both are using the same proof courier of LL.

I've linked a docker compose and a script which might help

### Expected behavior

A coop close should succeed, and even after the initial try tapd and lnd should be figure out on how to close this later on

[litd-close.log](https://github.com/user-attachments/files/27239058/litd-close.log)
[repro.sh](https://github.com/user-attachments/files/27239111/repro.sh)
[docker-compose.yml](https://github.com/user-attachments/files/27239110/docker-compose.yml)

PR #2110, opened 2026-04-30
server: fix corner case shutdown hang

### LINKED ISSUE #2109: Unreachable proof courier can block daemon shutdown
(NB, this problem was flushed out during examination of #2108, but it's not the primary cause of that issue. Opus-generated summary follows.)

## Summary

`importOutputProofs` (tapchannel/aux_sweeper.go:1403) uses an
uncancellable `context.Background()` and can block indefinitely
when the proof courier is unreachable. This affects the
post-confirmation cooperative close finalization path.

## Affected Code Path

    chainWatcher.finalizeCoopClose  (chain_watcher.go:1270)
      -> AuxChanCloser.FinalizeClose  (aux_closer.go:698)
        -> importOutputProofs          (aux_sweeper.go:1403)
          -> ReceiveProof(ctxb, ...)   (aux_sweeper.go:1478)
            -> BackoffHandler.Exec     (proof/courier.go:674)
              -> wait(ctxb, backoff)   (proof/courier.go:752)

This path runs **post-confirmation** — after the cooperative
close transaction has already been broadcast and confirmed
on-chain. It is only reached for the **responder** (non-channel-
initiator) side, which needs to import the funding input proofs
it doesn't already have (aux_closer.go:720).

## The Problem

Three things combine to create unbounded blocking:

### 1. Uncancellable context

At aux_sweeper.go:1418:

    ctxb := context.Background()

This context is passed to `proofFetcher.ReceiveProof(ctxb, ...)`
at line 1478. It has no timeout, no deadline, and no
cancellation mechanism.

### 2. Aggressive retry defaults

`BackoffHandler.Exec` (courier.go:674) retries the proof
transfer up to `DefaultProofTransferNumTries` times:

    const DefaultProofTransferNumTries = 2000

Each retry doubles the backoff duration up to `MaxBackoff`
(courier.go:758-762). With 2000 retries and exponential backoff,
the total blocking time can be enormous.

### 3. `wait()` only exits on context cancellation or timer

At courier.go:773:

    func (b *BackoffHandler) wait(ctx context.Context,
        wait time.Duration) error {

        select {
        case <-time.After(wait):
            return nil
        case <-ctx.Done():
            return fmt.Errorf("back off handler context done")
        }
    }

Since `context.Background()` never cancels, the only exit is
the timer. Each wait runs to full duration.

## Impact

When the proof courier is unreachable at the time the coop close
tx confirms:

- `FinalizeClose` blocks for a very long time (potentially
  hours, depending on backoff configuration).
- The post-confirmation finalization cannot complete.
- Asset proofs for the closed channel outputs are not imported
  into the local database.
- This could affect the node's ability to detect and respond
  to subsequent force closes involving those assets.

This does NOT prevent the cooperative close transaction from
being broadcast or confirmed. The close tx is already on-chain
by the time this code runs. But it prevents proper bookkeeping
afterwards.

## Suggested Fix

Replace `context.Background()` with a cancellable context:

Option A: **Timeout context.** Pass a context with a reasonable
deadline (e.g., 5 minutes) so the proof import fails fast when
the courier is unreachable. `FinalizeClose` can return an error,
and the caller can retry later.

Option B: **Caller-provided context.** Thread a context from
`chainWatcher.finalizeCoopClose` through `FinalizeClose` and
into `importOutputProofs`. This would require changing the
`AuxChanCloser.FinalizeClose` interface signature in LND to
accept a `context.Context`.

Option C: **Async proof import.** Move the `importOutputProofs`
call out of the synchronous `FinalizeClose` path entirely. Run
it as a background task that retries independently. This is the
most robust approach but requires the most restructuring.

jtobin:
Resolves #2109. This is a borderline-overfitting issue identified by Opus while diagnosing #2108, but the fix is simple, and does marginally improve the daemon's stability, so IMO it's worth making.

Previously an unresponsive proof courier could cause importOutputProofs to block indefinitely while a co-op close was in flight, preventing tapd from shutting down cleanly. The fix simply wires up AuxChanCloser with a ContextGuard lifecycle in the same manner as other Aux components (AuxFundingController, AuxSweeper, etc.), also threading a cancellable context through importOutputProofs.

(Shutting down the daemon in this situation is safe in terms of ultimately handling the coop close; the retry logic is handled by lnd.)

### DIFF
--- a/server.go
+++ b/server.go
@@ -247,6 +247,10 @@ func (s *Server) initialize(interceptorChain *rpcperms.InterceptorChain) error {
 	if err := s.cfg.AuxInvoiceManager.Start(); err != nil {
 		return fmt.Errorf("unable to start aux invoice mgr: %w", err)
 	}
+	if err := s.cfg.AuxChanCloser.Start(); err != nil {
+		return fmt.Errorf("unable to start aux chan closer: %w",
+			err)
+	}
 	if err := s.cfg.AuxSweeper.Start(); err != nil {
 		return fmt.Errorf("unable to start aux sweeper mgr: %w", err)
 	}
@@ -862,6 +866,10 @@ func (s *Server) Stop() error {
 			err)
 	}
 
+	if err := s.cfg.AuxChanCloser.Stop(); err != nil {
+		return fmt.Errorf("unable to stop aux chan closer: %w",
+			err)
+	}
 	if err := s.cfg.AuxLeafSigner.Stop(); err != nil {
 		return err
 	}
--- a/tapchannel/aux_closer.go
+++ b/tapchannel/aux_closer.go
@@ -110,11 +110,18 @@ type assetCloseInfo struct {
 
 // AuxChanCloser is used to implement asset-aware co-op close for channels.
 type AuxChanCloser struct {
+	startOnce sync.Once
+	stopOnce  sync.Once
+
 	cfg AuxChanCloserCfg
 
 	sync.RWMutex
 
 	closeInfo map[wire.OutPoint]*assetCloseInfo
+
+	// ContextGuard provides a wait group and main quit channel that can
+	// be used to create guarded contexts.
+	*fn.ContextGuard
 }
 
 // NewAuxChanCloser creates a new instance of the auxiliary channel closer.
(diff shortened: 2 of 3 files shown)


litbot-9000:
@georgetsagk: review reminder

GeorgeTsagk (approved):
Lgtm

needs a rebase to resolve a conflict

also would be nice to add a test, although setup might be tricky (non-blocking, leaving it up to you)

jtobin:
> also would be nice to add a test, although setup might be tricky (non-blocking, leaving it up to you)

Yeah, I looked into this but it's indeed difficult to set up an actually-good one.

I think the best proof of functionality here, verifiable by direct inspection of the diff, is that this simply brings AuxChanCloser in line with the way other aux components are handled (and they don't have any similar tests either, presumably for the same reason).

PR #2113, opened 2026-05-01
backup: fix group key verification and add skip-on-error

### LINKED ISSUE #2111: Backup import requires prior group key knowledge
Found during v0.8.0-rc1 testing (2026-04-30).

## Summary

Backup import (all modes: raw, compact, optimistic)
fails with `group key not known` when the backup
contains grouped assets and the importing node has
never seen the group key. The import aborts on the
first such failure, preventing any subsequent assets
from being imported — including ungrouped assets
that would succeed individually.

## Reproduction

1. On node A (tapd-alice), mint a grouped asset.
2. Mint additional ungrouped assets on A.
3. Export a backup from A (any mode):
   ```
   tapcli assets backup export \
       --mode raw --output_file backup.bin
   ```
4. Import onto node C (tapd-charlie), which has no
   federation server configured and has never synced
   the group key:
   ```
   tapcli assets backup import \
       --backup_file backup.bin
   ```
5. Import fails:
   ```
   failed to import proof for asset 0:
     unable to verify proof: group key not known:
     024db1...015ea: group verifier:
     asset group is unknown: group key not known
   ```
6. No assets are imported (including the ungrouped
   ones that would have succeeded individually).

Confirmed that raw, compact, and optimistic modes
all fail with the same error on the same node.

## Why Some Nodes Succeed

In testing, raw import onto Bob (tapd-bob) succeeded
while import onto Charlie (tapd-charlie) failed.
The difference: Bob has
`--universe.federationserver=tapd-primary:10029`
configured and had already synced the group key via
universe federation. Charlie has no federation server
and had never seen the group key.

The backup mode is irrelevant. The only factor is
whether the importing node already knows the group
key.

## Root Cause

The proof verifier's `GroupVerifier` checks group key
membership against the local database before
processing the proof chain. Even though the proof
file contains the full chain from genesis (which
includes the group witness that would register the
key), the verifier rejects the proof before the
genesis proof is processed.

This is compounded by fatal error handling: the
import loop returns immediately on the first asset
failure (`backup/import.go`, line 262-265), so one
problematic grouped asset blocks the entire import.

## Suggested Fixes

1. **Register group keys from the proof chain
   before verification.** The genesis proof in the
   chain contains the group witness. If import
   extracted and registered group keys from genesis
   proofs before running full verification, the
   verifier would find the key in the database.

2. **Skip, don't abort.** When a single asset fails
   to import, log a warning and continue to the
   next asset. Return the count of successes along
   with a summary of failures. This matches the
   existing skip behavior for spent and duplicate
   assets.

## Not Documented

The `docs/backup.md` specification does not mention
group key requirements or this failure mode.

jtobin:
Fixes #2111, with the fix confirmed in e2e tests on a Docker regtest network.

Opus's summary follows: 

> Fix backup import failing with "group key not known" when the importing node has no prior group key knowledge (no federation server, no prior universe sync).
> 
> Root cause: the GroupVerifier callback checks the local DB during proof chain verification, but the group key hasn't been stored yet — it exists in the same proof chain as a GroupKeyReveal on the genesis proof, or in another proof file in the backup.
> 
> Fix: pre-extract group keys from genesis proofs using SparseDecode (works on both raw and stripped proof files), verify the GroupKeyReveal derivation, and augment the GroupVerifier to accept pre-extracted keys. A retry pass handles v3 optimistic backup ordering where a grouped asset is processed before its anchor.
> 
> Also changes error handling so per-asset data/verification errors are skipped instead of aborting the entire import. Pre-verification runs with an infrastructure-free VerifierCtx (no-op HeaderVerifier, no IgnoreChecker, timelock validation skipped) so only proof-data issues are caught as skippable. The full ImportProofs call re-verifies with the real chain backend; failures there are fatal.

### DIFF
--- a/backup/import.go
+++ b/backup/import.go
@@ -7,11 +7,13 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/proof"
 	"github.com/lightningnetwork/lnd/chainntnfs"
+	lfn "github.com/lightningnetwork/lnd/fn/v2"
 	"github.com/lightningnetwork/lnd/keychain"
 )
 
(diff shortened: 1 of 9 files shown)


darioAnongba (approved):
LGTM

PR #2118, opened 2026-05-01
tapdb: make mssmt node inserts idempotent

### LINKED ISSUE #2116: Non-Idempotent Inserts in SQL-Backed MS-SMT
Found while resurrecting #770.

## Summary

The SQL insert queries for `mssmt_nodes` use plain `INSERT INTO`,
which fails with a UNIQUE constraint violation when a node with the
same `(hash_key, namespace)` already exists. This affects both
SQLite and Postgres, since both backends share the same queries in
`tapdb/sqlc/queries/mssmt.sql`.

## Root Cause

The three insert queries (`InsertBranch`, `InsertLeaf`,
`InsertCompactedLeaf`) all do strict inserts:

```sql
INSERT INTO mssmt_nodes (
    hash_key, l_hash_key, r_hash_key, key, value, sum, namespace
) VALUES ($1, $2, $3, NULL, NULL, $4, $5);
```

The primary key is `(hash_key, namespace)`. In a Merkle tree,
identical hashes mean identical content, so re-inserting the same
hash is semantically a no-op. The in-memory `DefaultStore` uses a
Go map, where re-assignment silently succeeds. The SQL store does
not tolerate it.

## How to Trigger

`TestDeletion` in `mssmt/tree_test.go` runs `testDeletion` and
`testBatchDeletion` sequentially on the **same store**.
`testDeletion` inserts 100 random leaves, then deletes them one by
one. `testBatchDeletion` then re-inserts the same 100 leaves into
the same store. During re-insertion, intermediate branch nodes
collide with nodes that survived the individual deletions (nodes
still referenced by other paths aren't deleted).

This produces two failures:

- `sqlite3/full_SMT`: UNIQUE constraint violation during
  `testBatchDeletion`'s `Insert` calls (line 606).
- `sqlite3/smol_SMT`: After `DeleteAllNodes`, `Get` returns a
  non-nil leaf (line 618), likely due to the compacted tree's
  in-memory cache not being invalidated after the SQL-level bulk
  delete.

## Practical Impact

The `TaprootAssetTreeStore` is the production SQL-backed MS-SMT
used for universe trees, asset commitment trees, etc. Production
usage patterns (single-leaf mutations per transaction) likely don't
trigger this, since the tree algorithm deletes old branch nodes
before inserting new ones along the same path. But any operation
that produces a branch node whose hash matches a node still live
elsewhere in the tree would fail.

## Fix

Change the three insert queries to use `INSERT OR IGNORE` (SQLite)
/ `ON CONFLICT DO NOTHING` (Postgres). Since hash determines
content in a Merkle tree, ignoring duplicates is correct — a node
with the same hash is guaranteed to have the same data.

## Discovery

This bug was found by PR #770, which registers the SQL tree store
driver so that `mssmt` unit tests actually run against the SQL
backend. Previously, `genTestStores` only produced constructors for
the in-memory `DefaultStore` because no `TreeStoreDriver` was ever
registered, making the sqlite3 guard in the loop dead code.

jtobin:
Fixes #2116.

Add ON CONFLICT (hash_key, namespace) DO NOTHING to InsertBranch, InsertLeaf, and InsertCompactedLeaf. In a Merkle tree, identical hashes guarantee identical content, so ignoring duplicates is correct. This fixes UNIQUE constraint violations when tree operations produce nodes whose hashes already exist in the table.

### DIFF
--- a/tapdb/sqlc/mssmt.sql.go
+++ b/tapdb/sqlc/mssmt.sql.go
@@ -244,6 +244,7 @@ const InsertBranch = `-- name: InsertBranch :exec
 INSERT INTO mssmt_nodes (
     hash_key, l_hash_key, r_hash_key, key, value, sum, namespace
 ) VALUES ($1, $2, $3, NULL, NULL, $4, $5)
+ON CONFLICT (hash_key, namespace) DO NOTHING
 `
 
 type InsertBranchParams struct {
@@ -269,6 +270,7 @@ const InsertCompactedLeaf = `-- name: InsertCompactedLeaf :exec
 INSERT INTO mssmt_nodes (
     hash_key, l_hash_key, r_hash_key, key, value, sum, namespace
 ) VALUES ($1, NULL, NULL, $2, $3, $4, $5)
+ON CONFLICT (hash_key, namespace) DO NOTHING
 `
 
 type InsertCompactedLeafParams struct {
@@ -294,6 +296,7 @@ const InsertLeaf = `-- name: InsertLeaf :exec
 INSERT INTO mssmt_nodes (
     hash_key, l_hash_key, r_hash_key, key, value, sum, namespace
 ) VALUES ($1, NULL, NULL, NULL, $2, $3, $4)
+ON CONFLICT (hash_key, namespace) DO NOTHING
 `
 
 type InsertLeafParams struct {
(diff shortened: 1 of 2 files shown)


litbot-9000:
@roasbeef: review reminder

bhandras (approved):
Nice catch, LGTM!

PR #2122, opened 2026-05-04
multi: add group_key to TransferInput and TransferOutput

### LINKED ISSUE #2121: multi: add group_key to TransferInput and TransferOutput
`AssetTransfer.inputs` and `AssetTransfer.outputs` currently carry only the protocol-level `asset_id` (the issuance/tranche identifier). For grouped fungible assets, the user-facing identifier is the **group key**, not the issuance asset_id — but the group key is not on these messages. Consumers that want to display, filter, or key-by the user-facing asset have to make a second RPC (typically `ListAssets`) and join in client code.

This affects both `ListTransfers` and the `transfer` field embedded in `SendEvent`. Other RPCs like `Addr` already expose both `asset_id` and `group_key` directly — `TransferInput` / `TransferOutput` are the outliers.

## Proposal

Add a single optional field to each message:

```proto
message TransferInput {
    string anchor_point = 1;
    bytes asset_id      = 2;
    bytes script_key    = 3;
    uint64 amount       = 4;
    bytes group_key     = 5;  // NEW: empty when the asset has no group
}

message TransferOutput {
    // ... existing fields ...
    bytes asset_id  = 12;
    string tap_addr = 14;
    bytes group_key = 15;  // NEW: empty when the asset has no group
}
```

Empty `bytes` ≡ "no group key" (proto3 default). Consumers reconstruct the user-facing identifier as: if `len(group_key) > 0`, key by `group_key`; otherwise key by `asset_id`.

`SendEvent.transfer` benefits automatically (it embeds `AssetTransfer`). `ReceiveEvent` already has the data it needs via the embedded `Addr`, no change needed there.

## Implementation approach

Resolve the group key at marshal time in `rpcserver.marshalOutboundParcel` via the existing `FetchGroupByGenesis` path, with a per-call `map[asset.ID]*btcec.PublicKey` cache so that N inputs/outputs sharing one asset collapse to one DB lookup.

This avoids:
- a schema migration on `asset_transfer_inputs` / `asset_transfer_outputs`,
- a backfill problem for historical rows,
- threading `GroupKey` through `tapfreighter.TransferInput` / `TransferOutput` and the parcel save/load paths.

The cost is one indexed lookup per unique `asset_id` per marshal call. For typical transfers (1-3 unique assets per parcel) this is negligible compared to the work the marshaler already does.

Persisting the group key on the DB rows would be cleaner in isolation, but the migration cost is hard to justify for a value that's already derivable from existing tables.

## Testing plan

- Unit tests for the marshal-time resolver helper (cache hit/miss, asset with/without group, ungrouped fallback).
- itest extending an existing fungible-grouped send to assert `group_key` populates on both inputs and outputs end-to-end across `ListTransfers` and `SubscribeSendEvents`.
- itest case for ungrouped (asset-id only) sends to confirm `group_key` stays empty.
- Existing send/receive itests should pass unchanged.

darioAnongba:
## Summary

- Adds an optional `group_key` field to `TransferInput` and `TransferOutput` in `taprpc/taprootassets.proto` so consumers can identify the user-facing asset (group key for grouped fungibles, asset ID otherwise) without a side query.
- Populates the new field at marshal time in `rpcserver.marshalOutboundParcel` via `FetchGroupByGenesis`, with a per-call cache so a transfer with N inputs/outputs sharing one asset collapses to one DB query.
- Threads the same resolver through `marshalSendEvent` so `SubscribeSendEvents` benefits without an RPC-per-event lookup.
- Treats `address.ErrAssetGroupUnknown` as "no group" rather than a hard error so `ListTransfers` stays resilient against gaps in older rows; consumers fall back to `asset_id` in that case (same effective behavior as today).

`SendEvent.transfer` benefits automatically because it embeds `AssetTransfer`. `ReceiveEvent` already had what it needs via the embedded `Addr`, no change there. The change is additive: empty `bytes` ≡ "no group key" (proto3 default), older clients ignore the new field, newer clients see the empty default on responses from older daemons.

Closes #2121.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -330,6 +330,9 @@
   Add pagination support (offset, limit, direction) to the `ListAssets` RPC
   endpoint.
 
+- [PR#2122](https://github.com/lightninglabs/taproot-assets/pull/2122)
+  Add a `group_key` field to `TransferInput` and `TransferOutput`. Affects `ListTransfers` and the `transfer` field embedded in `SendEvent`.
+
 ## tapcli Updates
 
 - [PR#1995](https://github.com/lightninglabs/taproot-assets/pull/1995)
--- a/itest/test_list_on_test.go
+++ b/itest/test_list_on_test.go
@@ -467,6 +467,10 @@ var allTestCases = []*testCase{
 		name: "address v2 import fails without courier",
 		test: testAddressV2ImportFailsWithoutCourier,
 	},
+	{
+		name: "transfer group key",
+		test: testTransferGroupKey,
+	},
 }
 
 var optionalTestCases = []*testCase{
--- a/itest/transfer_group_key_test.go
+++ b/itest/transfer_group_key_test.go
@@ -1,0 +1,127 @@
+package itest
+
+import (
+	"context"
+
+	"github.com/lightninglabs/taproot-assets/taprpc"
+	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	"github.com/stretchr/testify/require"
+)
+
+// testTransferGroupKey verifies that the group_key field added to
+// TransferInput and TransferOutput populates correctly across both
+// ListTransfers and SubscribeSendEvents for grouped fungible assets, and
+// stays empty for ungrouped (asset-id only) assets.
+//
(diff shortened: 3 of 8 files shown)


jtobin (approved):
LGTM. :+1: :+1:

PR #2123, opened 2026-05-04
make: invoke gosimports and migrate via absolute path

darioAnongba:
## Summary

The `fmt` and `migrate-*` recipes invoked the bare tool name (`gosimports`, `migrate`), so they only worked when `\$GOPATH/bin` was on `PATH`. Switch to the `\$(GOIMPORTS_BIN)` and `\$(MIGRATE_BIN)` variables that the dependency rules already point at, and resolve `GOPATH` via `go env` so the path is correct regardless of the user's shell environment.

## Test plan

- [x] `make fmt` runs clean on a shell where `\$GOPATH/bin` is not on `PATH`.


jtobin (approved):
LGTM. :+1: :+1:

PR #2125, opened 2026-05-04
rpcserver: expose asset type on transfers

darioAnongba:
## Changes

- add `asset_type` to `TransferInput` and `TransferOutput`
- populate transfer `group_key` and `asset_type` from shared asset metadata resolution in list/send/burn/event paths
- extend the transfer metadata itest across grouped fungible, ungrouped fungible, and grouped collectible sends
- add a release note for the RPC response fields

## Notes

This lets SDK clients distinguish grouped fungibles from NFT collection items without a resolver. Both can carry `group_key`, but only collectibles should keep their item asset ID as the high-level NFT handle.

For ungrouped fixed-supply fungibles, transfers now return `asset_type = NORMAL` with no `group_key`


jtobin (approved):
LGTM. :+1: :+1:

PR #2126, opened 2026-05-05
rpcserver: expose burn asset type

darioAnongba:
## Summary

ListBurns now returns the immutable asset type for each burn row, matching the transfer projection fields added recently. This lets SDK consumers distinguish grouped fungible burns from collectible item burns without resolver lookups.

The value is derived by joining burn rows against `genesis_assets`, so existing burn rows are covered without a schema migration.


jtobin (approved):
LGTM. :+1: :+1:

Issue #1976, opened 2026-02-04
[flake]: suspected custom_channels_htlc_force_close_MPP flake


jtobin:
`custom_channels_htlc_force_close_MPP` was found failing in CI on #1775. I haven't seen this before, but the changes in that PR certainly shouldn't trigger it. The error, from [this](https://github.com/lightninglabs/taproot-assets/actions/runs/21669261648/job/62472657112?pr=1775) job:

```
    assets_test.go:3608: 
        	Error Trace:	/home/runner/work/taproot-assets/taproot-assets/lightning-terminal/itest/assets_test.go:3608
        	            				/home/runner/work/taproot-assets/taproot-assets/lightning-terminal/itest/litd_custom_channels_test.go:4680
        	            				/home/runner/work/taproot-assets/taproot-assets/lightning-terminal/itest/litd_custom_channels_test.go:4509
        	            				/home/runner/work/taproot-assets/taproot-assets/lightning-terminal/itest/test_harness.go:103
        	            				/home/runner/work/taproot-assets/taproot-assets/lightning-terminal/itest/litd_test.go:138
        	Error:      	Received unexpected error:
        	            	wanted 2, found 1 txs in mempool: [07c4b7ac79d7031f3507ad86839b7225e5f2f7a9f04586b14b108ed79b50446e]
        	Test:       	TestLightningTerminal/tranche00/17-of-29/custom_channels_htlc_force_close_MPP
```

I feel this is likely a concurrency issue.

jtobin:
Also observed in [this run](https://github.com/lightninglabs/taproot-assets/actions/runs/22956544314/job/66635045035?pr=2017), though with a different error condition:

```
Error Trace:  /home/runner/work/taproot-assets/taproot-assets/itest/custom_channels/helpers.go:2545
                    /home/runner/work/taproot-assets/taproot-assets/itest/custom_channels/helpers.go:3061
                    /home/runner/work/taproot-assets/taproot-assets/itest/custom_channels/htlc_force_close_test.go:196
                    /home/runner/work/taproot-assets/taproot-assets/itest/custom_channels/htlc_force_close_test.go:35
                    /home/runner/work/taproot-assets/taproot-assets/itest/custom_channels/custom_channels_test.go:208
Error:        Received unexpected error:
              missing anchor block hash, transfer not confirmed
Test:         TestCustomChannels/htlc_force_close_mpp
```

jtobin:
[...]

PR #2127, opened 2026-05-06
tapd-integrated: initialize tapd subsystem loggers

jtobin:
Simply lets us view tapd's logs during the custom channels itests (this was instrumental in debugging #1976).

### DIFF
--- a/cmd/tapd-integrated/main.go
+++ b/cmd/tapd-integrated/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/integration"
 	"github.com/lightninglabs/taproot-assets/tapcfg"
 	"github.com/lightningnetwork/lnd"
+	"github.com/lightningnetwork/lnd/build"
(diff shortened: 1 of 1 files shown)


darioAnongba (approved):
nit: Looks like this function adds both `stdout` and the file handler. We could disable stdout with `cfg.TaprootAssets.Logging.Console.Disable = true` even though it's harmless for now. Just in case.

jtobin:
> nit: Looks like this function adds both `stdout` and the file handler. We could disable stdout with `cfg.TaprootAssets.Logging.Console.Disable = true` even though it's harmless for now. Just in case.

Looks like we actually need stdout, whereas the file handler is the one that's inert (I believe the test harness captures the integrated binary's stdout and writes _its own_ log files). Both get added by default, but the file handler seems to be inert unless/until InitLogRotator is called, which in the integrated binary's case is never.

We could disable _that_ explicitly, but it doesn't do anything anyway, so it's probably fine as-is.

PR #2129, opened 2026-05-11
build: trim install dependency graph

### LINKED ISSUE #2128: build: reduce make install disk/cache footprint for v0.8
## Problem

Users are reporting that `make install` for `v0.8.0-rc2` can require roughly 6-7GB of free storage, compared with about 4.5GB for `v0.7.2`.

I did a local measurement to separate the clean-build footprint from the upgrade/cache footprint. Exact values will vary by machine and Go toolchain, but the pattern is useful.

Environment used for the local measurement:

- macOS darwin/arm64
- host `go version go1.26.1 darwin/arm64`
- isolated `GOCACHE`, `GOMODCACHE`, `GOPATH`, `GOBIN`, and `TMPDIR`
- compared `v0.7.2` against `v0.8.0-rc2`

Results:

| Build scenario | Total disk | Go build cache | Module cache | Installed binaries | Peak RSS |
| --- | ---: | ---: | ---: | ---: | ---: |
| clean `v0.7.2` | ~3.3G | ~2.3G | ~843M | ~106M | ~2.0GB |
| clean `v0.8.0-rc2` | ~3.7G | ~2.8G | ~883M | ~112M | ~2.35GB |
| build `v0.7.2`, then `v0.8.0-rc2` in the same Go cache | ~6.1G | ~4.9G | ~1.1G | ~112M | ~2.35GB |

So the 6-7GB report appears to be mostly cumulative Go build cache growth during upgrade, not the size of the final binaries. The new version invalidates enough cached package archives that old and new compiled package archives coexist until Go trims or the user runs `go clean -cache`.

## What changed

This does not look like only a Go version bump.

Between `v0.7.2` and `v0.8.0-rc2`:

- `go.mod` moved from `go 1.24.11` to `go 1.25.5`.
- Makefile release Go moved from `1.25.2` to `1.25.5`.
- `lnd` moved from a Nov 2025 pseudo-version to an Apr 2026 pseudo-version.
- Large dependency bumps landed, including `grpc` `v1.64.1 -> v1.80.0`, `protobuf`, `btcd`, `btcwallet`, `neutrino`, `golang.org/x/*`, Docker/Moby packages, and OpenTelemetry packages.
- The production dependency graph for `./cmd/tapd ./cmd/tapcli` grew from about 893 packages to about 979 packages in my local check.
- The repo delta is large: hundreds of changed files and ~477 non-merge commits.

## Suspected build graph bloat

There are a few imports that appear to pull test/integration-heavy packages into normal `tapd`/`tapcli` builds:

1. `tapdb/postgres_fixture.go` imports `github.com/ory/dockertest/v3` and `github.com/ory/dockertest/v3/docker`, but it is not a `_test.go` file. Through `tapcfg -> tapdb`, normal `cmd/tapd` builds compile the Docker client stack.

2. `tapdb/postgres.go` has `NewTestPostgresDB` and `NewTestPostgresDBWithVersion` helpers in a non-test file, which keeps the fixture reachable from the production package.

3. `rpcutils/marshal.go` imports `github.com/lightningnetwork/lnd/lntest` only for `lntest.ParseDerivationPath`. That pulls in `lnd/lntest`, `btcd/integration/rpctest`, and other test-heavy packages into `cmd/tapcli`.

4. `tapcfg/server.go` imports root `github.com/lightningnetwork/lnd` for `lnd.ListenerCfg`. Since root `lnd` is large, it may be worth checking whether that type can be avoided, localized, or moved behind an interface/helper that does not force the whole root package into plain `tapd` builds.

## Possible fixes

- Move Postgres fixture code and helpers behind `_test.go` files or explicit test/integration build tags.
- Replace the `lntest.ParseDerivationPath` dependency with a small local parser or a non-test utility package.
- Audit the root `github.com/lightningnetwork/lnd` import from `tapcfg/server.go`; avoid importing root `lnd` in production paths if possible.
- Add a build-footprint check or script that reports package count and top cached export archives for `./cmd/tapd ./cmd/tapcli`, so regressions are visible before releases.
- Document a short workaround for users upgrading with tight disk:
  - `go clean -cache && make install`
  - if necessary, `go clean -cache -modcache && make install`
  - for lower peak memory at the cost of slower builds, `GOFLAGS='-p=1' make install`

## Acceptance criteria

- Plain `make install` for `cmd/tapd`/`cmd/tapcli` no longer compiles test-only Docker/lntest/rpctest dependencies.
- The clean-cache package count and build-cache footprint are reduced or at least stop regressing.
- Upgrade-path disk usage is documented or improved enough that users with around 5GB free do not hit surprising failures.

darioAnongba:
This PR trims the normal `tapd`/`tapcli` install dependency graph by moving test-only or small reusable helpers out of the production import path.

- copy the small derivation-path parser locally instead of importing `lntest`
- copy the narrow `lncfg` helper surface used by `tapd`/`tapcli`
- replace LND RPC listener wrapper types with local `tapconfig` types
- gate the Docker-backed Postgres fixture behind `test_db_postgres || itest` and add a default-build stub
- copy the small test port allocator used by local test helpers

The remaining production graph references to `dockertest`, `lnd/lncfg`, and `lntest/wait` now trace through `lndclient -> lnd/...`, not through direct taproot-assets imports. That will need a follow-up change in `lndclient`/LND.

Fixes #2128.

## Validation

- `go list -deps ./cmd/tapd ./cmd/tapcli | sort -u | wc -l` -> moved from ? 970 to 920


jtobin (approved):
Looks good. The utilities brought in from internal/lncfg look unlikely to change, and also don't seem to be things that would necessarily need to match upstream precisely anyway.

Issue #2066, opened 2026-04-13
taprpc: add GenesisInfo to AssetGroupBalance in ListBalancesResponse


darioAnongba:
## Context

`ListBalancesResponse.AssetGroupBalances` entries only contain `GroupKey` + `Balance`, with no `GenesisInfo`. This makes the `group_by=group_key` mode unusable for clients that need asset metadata (name, type, decimal display) alongside balances.

For fungible assets, the asset name is the same across all tranches in a group, so there's no ambiguity about which genesis to return.

## Prior Art

`DecodeAssetPayReq` already handles this case. When called with a `group_key`, it resolves to a representative tranche and returns `GenesisInfo` (line 300 of `tapchannel.proto`):

```protobuf
message AssetPayReqResponse {
    uint64 asset_amount = 1;
    taprpc.DecimalDisplay decimal_display = 2;
    taprpc.AssetGroup asset_group = 3;
    // Always populated. For group-key queries, the asset_id reflects the
    // resolved tranche's genesis ID.
    taprpc.GenesisInfo genesis_info = 4;
    lnrpc.PayReq pay_req = 5;
}
```

## Proposal

Add `GenesisInfo` (and optionally `DecimalDisplay`) to `AssetGroupBalance`:

```protobuf
message AssetGroupBalance {
    bytes group_key = 1;
    uint64 balance = 2;
    // The genesis info of a representative issuance in the group.
    GenesisInfo asset_genesis = 3;
}
```

This would allow SDK clients to use `group_by=group_key` directly for grouped assets without needing a second RPC call or the `group_by=asset_id` + re-aggregation workaround.

## Workaround

The tap-sdk currently works around this by requesting `group_by=asset_id` and re-aggregating entries with the same group key client-side (see lightninglabs/tap-sdk#61).

kaldun-tech:
Going to take a swing on a PR here

kaldun-tech:
Putting some finishing touches on PR #2130 

My solution adds `GenesisInfo` and `DecimalDisplay` to `AssetGroupBalance` as proposed. For groups with multiple tranches, we return the genesis of a representative issuance (selected via `MIN(gen_asset_id)` in the SQL query). 

For fungible assets, all tranches share the same name and type, so I believe this provides correct metadata.

While implementing this, I noticed that `listBalancesByAsset` (in the `group_by=asset_id` code path) is missing `OutputIndex` in its  `GenesisInfo` marshaling, even though the SQL query returns it.

[...]

Issue #2120, opened 2026-05-02
Porter can strand parcels until daemon restart


jtobin:
(NB, this is the leading hypothesis for the remaining flake noted in #1976, i.e. "missing anchor block hash, transfer not confirmed." The actual bug itself appears to be in chain porter's error handling; the "primary" and "tertiary" potential fixes IMO look most promising.)

(EDIT: this has been ruled out as the cause of #1976.)

## Failure

CI run [25216477057](https://github.com/lightninglabs/taproot-assets/actions/runs/25216477057/job/73938380610), branch `mssmt-unit-db-backends` @
657fd714. Both `htlc_force_close` and `htlc_force_close_mpp`
fail with:

```
missing anchor block hash, transfer not confirmed
```

at `locateAssetTransfers` for Alice's final second-level HTLC
success sweep (`helpers.go:3155`; CI shows `:3173` due to line
drift on the CI branch). The transfer row exists; the block
hash does not. The test polls for 2 minutes and times out.

This is NOT the timeout-sweep mempool/RBF flake fixed by
3d8d534b. That failure occurs later in the test; this one
fires in the HTLC success path before timeout sweeps begin.

## Observed State

- `ListTransfers` returns exactly one transfer for the mined
  sweep txid.
- `AnchorTxBlockHash` is nil for the full poll window.
- The predecessor sweep (first-level HTLC success tx) was
  already confirmed: `locateAssetTransfers` at
  `helpers.go:3122` succeeded, meaning `storeProofs` and
  `LogAnchorTxConfirm` both completed for it.

## Transfer Lifecycle

```
StorePreBroadcast     LogPendingParcel — row written
Broadcast             PublishTransaction
WaitTxConf            deliverOutboundPkgResp — caller unblocked
                      waitForTransferTxConf  — blocks for conf
StorePostAnchorTxConf storeProofs
                      storePackageAnchorTxConf
                        → LogAnchorTxConfirm — block hash written
```

The row exists but the block hash was never written. The state
machine stopped advancing somewhere between
`deliverOutboundPkgResp` and `LogAnchorTxConfirm`.

### Structural bug: silent post-delivery error

The porter delivers the response to the caller in `WaitTxConf`
(`chain_porter.go:2096`) *before* doing the confirmation wait
and post-confirmation work. `RequestShipment` returns on
`respChan`; `NotifyBroadcast` returns to lnd's sweeper. From
that point the porter goroutine is unsupervised.

[...]

PR #2130, opened 2026-05-13
taprpc+tapdb: add GenesisInfo to AssetGroupBalance in ListBalances

kaldun-tech:
Add `GenesisInfo` and `DecimalDisplay` fields to `AssetGroupBalance` in the `ListBalances` RPC response when using `group_by=group_key` mode.

Previously, `AssetGroupBalance` only contained `group_key` and `balance`, making it impossible for clients to display asset metadata (name, type, decimal display) alongside grouped balances without additional RPC calls.

## Changes

- **Proto**: Added `asset_genesis` (field 3) and `decimal_display` (field 4) to `AssetGroupBalance` message
- **SQL**: Modified `QueryAssetBalancesByGroup` to join `genesis_info_view` and return genesis data for a representative issuance per group
- **DB Store**: Extended `AssetGroupBalance` struct with genesis fields and updated query result mapping
- **RPC Server**: Populated new fields in `listBalancesByGroupKey`, including `DecimalDisplay` lookup via `DecDisplayForAssetID`

## Design Notes

- The representative genesis per group is selected using `MIN(gen_asset_id)`, which gives the earliest-inserted genesis. For locally-minted assets, this is the group anchor. For fungible assets, all tranches share the same name/type, so any genesis provides correct metadata.
- ErrAssetMetaNotFound leaves DecimalDisplay unset (asset has no meta); other errors are escalated to the caller.
- Ungrouped assets are intentionally excluded from `QueryAssetBalancesByGroup` results (use `group_by=asset_id` mode for those).

### Test Coverage Gaps

1. **Error handling paths in RPC layer**: The `ErrAssetMetaNotFound` and error escalation paths in `listBalancesByGroupKey` are not directly unit tested. This would require mocking `AddrBook.DecDisplayForAssetID`.

2. **Received grouped assets without local anchor**: Edge case where we receive a grouped asset but don't have the anchor locally. The new query should still work (joins through `asset_groups.genesis_point_id`), but this isn't explicitly tested.

## Test Plan

- [x] `make unit pkg=tapdb` - verify DB layer changes
- [x] `make unit pkg=rpcserver` - verify RPC marshaling
- [x] Manual test run via `make itest icase=minting_multi_asset_groups`
- [x] `tapcli assets balance --by-group` returns genesis info
- [x] Manual test: JSON/REST response includes new fields

Fixes issue #2066

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -348,6 +348,12 @@
   clients to distinguish grouped fungible burns from grouped collectible
   burns.
 
+- [PR#2130](https://github.com/lightninglabs/taproot-assets/pull/2130)
+  Add `asset_genesis` and `decimal_display` fields to `AssetGroupBalance`
+  in `ListBalances`. When using `group_by=group_key` mode, clients now
+  receive asset metadata (name, type, decimal display) alongside grouped
+  balances without requiring additional RPC calls.
+
 - [PR#2100](https://github.com/lightninglabs/taproot-assets/pull/2100)
   Add pagination support (offset, limit, direction) to the `AssetLeaves`
   RPC endpoint, and add `MaxPageSize` validation to `AssetRoots`.
--- a/rpcserver/rpcserver.go
+++ b/rpcserver/rpcserver.go
@@ -1483,11 +1483,48 @@ func (r *RPCServer) listBalancesByGroupKey(ctx context.Context,
 			groupKey = balance.GroupKey.SerializeCompressed()
 		}
 
-		groupKeyString := hex.EncodeToString(groupKey)
-		resp.AssetGroupBalances[groupKeyString] = &taprpc.AssetGroupBalance{
-			GroupKey: groupKey,
-			Balance:  balance.Balance,
+		// Create the genesis info for the representative issuance.
+		genesisInfo := &taprpc.GenesisInfo{
+			GenesisPoint: balance.GenesisPoint.String(),
+			AssetType:    taprpc.AssetType(balance.Type),
+			Name:         balance.Tag,
+			MetaHash:     balance.MetaHash[:],
+			AssetId:      balance.ID[:],
+			OutputIndex:  balance.OutputIndex,
 		}
+
+		groupKeyString := hex.EncodeToString(groupKey)
+		groupBalance := &taprpc.AssetGroupBalance{
+			GroupKey:     groupKey,
+			Balance:      balance.Balance,
+			AssetGenesis: genesisInfo,
+		}
+
+		// Fetch the decimal display for this asset ID.
+		decDisplay, err := r.cfg.AddrBook.DecDisplayForAssetID(
+			ctx, balance.ID,
+		)
+		switch {
+		case errors.Is(err, address.ErrAssetMetaNotFound):
+			// Asset legitimately has no meta — leave DecimalDisplay
+			// unset.
+
+		case err != nil:
+			return nil, fmt.Errorf("unable to fetch decimal "+
+				"display for asset %x: %w", balance.ID[:], err)
+
+		default:
+			groupBalance.DecimalDisplay = fn.MapOptionZ(
+				decDisplay,
+				func(d uint32) *taprpc.DecimalDisplay {
+					return &taprpc.DecimalDisplay{
+						DecimalDisplay: d,
+					}
+				},
+			)
+		}
+
+		resp.AssetGroupBalances[groupKeyString] = groupBalance
 	}
 
 	// We will also report the number of unconfirmed transfers. This is
--- a/rpcserver/rpcserver.go
+++ b/rpcserver/rpcserver.go
@@ -1595,11 +1595,48 @@ func (r *RPCServer) listBalancesByGroupKey(ctx context.Context,
 			groupKey = balance.GroupKey.SerializeCompressed()
 		}
 
-		groupKeyString := hex.EncodeToString(groupKey)
-		resp.AssetGroupBalances[groupKeyString] = &taprpc.AssetGroupBalance{
-			GroupKey: groupKey,
-			Balance:  balance.Balance,
+		// Create the genesis info for the representative issuance.
+		genesisInfo := &taprpc.GenesisInfo{
+			GenesisPoint: balance.GenesisPoint.String(),
+			AssetType:    taprpc.AssetType(balance.Type),
+			Name:         balance.Tag,
+			MetaHash:     balance.MetaHash[:],
+			AssetId:      balance.ID[:],
+			OutputIndex:  balance.OutputIndex,
 		}
+
+		groupKeyString := hex.EncodeToString(groupKey)
+		groupBalance := &taprpc.AssetGroupBalance{
+			GroupKey:     groupKey,
+			Balance:      balance.Balance,
+			AssetGenesis: genesisInfo,
+		}
+
+		// Fetch the decimal display for this asset ID.
+		decDisplay, err := r.cfg.AddrBook.DecDisplayForAssetID(
+			ctx, balance.ID,
+		)
+		switch {
+		case errors.Is(err, address.ErrAssetMetaNotFound):
+			// Asset legitimately has no meta — leave DecimalDisplay
+			// unset.
+
+		case err != nil:
+			return nil, fmt.Errorf("unable to fetch decimal "+
+				"display for asset %x: %w", balance.ID[:], err)
+
+		default:
+			groupBalance.DecimalDisplay = fn.MapOptionZ(
+				decDisplay,
+				func(d uint32) *taprpc.DecimalDisplay {
+					return &taprpc.DecimalDisplay{
+						DecimalDisplay: d,
+					}
+				},
+			)
+		}
+
+		resp.AssetGroupBalances[groupKeyString] = groupBalance
 	}
 
 	// We will also report the number of unconfirmed transfers. This is
--- a/tapdb/sqlc/queries/assets.sql
+++ b/tapdb/sqlc/queries/assets.sql
@@ -186,9 +186,10 @@ INSERT INTO asset_groups (
 ) VALUES (
     $1, $2, $3, $4, $5, $6
 ) ON CONFLICT (tweaked_group_key)
-    -- This is not a NOP, update the genesis point ID in case it wasn't set
-    -- before.
-    DO UPDATE SET genesis_point_id = EXCLUDED.genesis_point_id
+    -- Preserve the original genesis_point_id (the anchor's genesis point).
+    -- We update version to ensure RETURNING works, but genesis_point_id
+    -- must remain unchanged as it identifies the canonical group anchor.
+    DO UPDATE SET version = EXCLUDED.version
 RETURNING group_id;
 
 -- name: UpsertAssetGroupWitness :one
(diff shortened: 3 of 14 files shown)


rpcserver/rpcserver.go
```diff
@@ -1483,11 +1483,44 @@ func (r *RPCServer) listBalancesByGroupKey(ctx context.Context,
...
+			Balance:      balance.Balance,
+			AssetGenesis: genesisInfo,
+		}
+
+		// Fetch the decimal display for this asset ID. We don't fail
+		// the request if this lookup fails since decimal display is
+		// optional metadata.
+		decDisplay, err := r.cfg.AddrBook.DecDisplayForAssetID(
+			ctx, balance.ID,
+		)
```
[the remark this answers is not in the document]
kaldun-tech: The `DecDisplayForAssetID` method has an internal cache that pre-populates with ALL asset metadata on first call (`address/book.go` line 482: FetchAllAssetMeta). So the N+1 concern is largely mitigated, after the first lookup, subsequent calls are cache hits.

kaldun-tech:
I believe these CI failures are flakes. @claude please code review and investigate whether the failed CI is due to code changes or flakes

kaldun-tech:
@jtobin @darioAnongba Please re-run the failed CI jobs. My investigation is that the failures were due to flakes but let me know if you want me to look into anything on my end

rpcserver/rpcserver.go
```diff
@@ -1483,11 +1484,45 @@ func (r *RPCServer) listBalancesByGroupKey(ctx context.Context,
...
+					return &taprpc.DecimalDisplay{
+						DecimalDisplay: d,
+					}
+				},
+			)
+		} else {
+			rpcsLog.DebugS(ctx, "Failed to fetch decimal display",
+				btclog.Fmt("asset_id", "%x", balance.ID[:]),
+				"error", err)
 		}
```
darioAnongba: We can avoid silently logging debug and escalate instead.

```suggestion
		switch {
case errors.Is(err, address.ErrAssetMetaNotFound):
    // Asset legitimately has no meta — leave DecimalDisplay unset.

case err != nil:
    return nil, fmt.Errorf("unable to fetch decimal display "+
        "for asset %x: %w", balance.ID[:], err)

default:
    groupBalance.DecimalDisplay = fn.MapOptionZ(
        decDisplay,
        func(d uint32) *taprpc.DecimalDisplay {
            return &taprpc.DecimalDisplay{DecimalDisplay: d}
        },
    )
}
```

tapdb/sqlc/queries/assets.sql
```diff
@@ -362,20 +362,43 @@ GROUP BY assets.genesis_id, genesis_info_view.asset_id,
...
 -- name: QueryAssetBalancesByGroup :many
+WITH group_anchor AS (
+    -- Select a representative genesis for each group. We use MIN(gen_asset_id)
+    -- as a deterministic way to pick one genesis per group. For fungible assets
+    -- (the primary use case), all tranches in a group share the same name and
+    -- type, so any genesis provides correct metadata. This is a heuristic that
+    -- works for locally-minted assets where the anchor is inserted first.
+    SELECT
+        tweaked_group_key,
+        MIN(gen_asset_id) as anchor_gen_id
```
darioAnongba: `gen_asset_id` is a local-DB primary key reflecting **local insertion order**, not protocol-level ordering or any canonical anchor. The canonical anchor is stored explicitly in `asset_groups.genesis_point_id`

A fix is to join through `asset_groups.genesis_point_id` to resolve the actual anchor `gen_asset_id`. The existing `FetchGroupByGroupKey` pattern (`ORDER BY witness_id LIMIT 1`) already does this.

darioAnongba (changes requested):

kaldun-tech:
Hey @darioAnongba thanks for the feedback. I'm making a few changes:
1. SQL Query (assets.sql): Changing from `MIN(gen_asset_id)` to joining through `asset_groups.genesis_point_id` for canonical anchor selection.
2. Error Handling (rpcserver.go): Changing from silent debug logging to proper error handling - `ErrAssetMetaNotFound` leaves `DecimalDisplay` unset, other errors are escalated.
3. Unit Tests (assets_store_test.go): Adding assertions to verify genesis info fields match expected anchor genesis for each group.

kaldun-tech:
@darioAnongba Pushed my changes
1. SQL query now joins through `asset_groups.genesis_point_id` for canonical anchor selection
2. Error handling escalates non-ErrAssetMetaNotFound errors instead of silent logging
3. Added DB layer test assertions verifying genesis info matches the expected anchor

kaldun-tech:
Investigated the CI failure. Copilot & Claude pointed me at PR #2133 but this is not yet merged which indicates the failure is due to flakiness. Sergey's changes in PR #2060 may correct this.

EDIT: Second failure today that seems to be caused by issue #2120  and fixed by PR #2132. Will investigate

kaldun-tech:
@GeorgeTsagk @jtobin Tagging for review

litbot-9000:
@kaldun-tech, remember to re-request review from reviewers when ready

litbot-9000:
@kaldun-tech, remember to re-request review from reviewers when ready

litbot-9000:
@kaldun-tech, remember to re-request review from reviewers when ready

litbot-9000:
@kaldun-tech, remember to re-request review from reviewers when ready

litbot-9000:
@kaldun-tech, remember to re-request review from reviewers when ready

rpcserver/rpcserver.go
```diff
@@ -1594,11 +1594,48 @@ func (r *RPCServer) listBalancesByGroupKey(ctx context.Context,
...
+			Balance:      balance.Balance,
+			AssetGenesis: genesisInfo,
+		}
+
+		// Fetch the decimal display for this asset ID.
+		decDisplay, err := r.cfg.AddrBook.DecDisplayForAssetID(
+			ctx, balance.ID,
+		)
+		switch {
+		case errors.Is(err, address.ErrAssetMetaNotFound):
```
GeorgeTsagk, line 1619: This never matches. DecDisplayForAssetID wraps the store error with %v, so errors.Is fails and we fall into the escalation branch. Any grouped asset whose anchor has no meta makes the whole by-group ListBalances call error out. The itest assets all carry meta, which is why it is not caught.

Fix is %w in address/book.go DecDisplayForAssetID plus a test for a group without meta. Also note that a cache miss there goes through FetchAssetMetaForAsset, which with the universe syncer wired in will try a federation sync on every call for such assets.

tapdb/sqlc/queries/assets.sql
```diff
@@ -367,20 +368,45 @@ GROUP BY assets.genesis_id, genesis_info_view.asset_id,
          genesis_info_view.prev_out, key_group_info_view.tweaked_group_key;
 
 -- name: QueryAssetBalancesByGroup :many
+WITH group_anchor AS (
+    -- Select the canonical anchor genesis for each group by joining through
```
GeorgeTsagk, line 372: Two problems with this CTE.

1. It is an inner join through asset_groups.genesis_point_id and asset_group_witnesses. When the group row was first created via an address for tranche A (no witness) and we later receive tranche B, there is no witness for A and the group disappears from the by-group balances entirely. Main returns the group in that scenario, I reproduced both. Freezing genesis_point_id in the upsert below makes this permanent.

2. When the anchor and another tranche share a genesis point (same mint batch, exactly what the multi asset group itest does) the CTE returns one row per tranche. Balance stays correct but the representative genesis is whichever row the DB emits last, so it differs between runs and backends.

Suggestion: pick the representative from asset_group_witnesses joined to key_group_info_view, collapse to one row per group with MIN over the witness gen ids, and LEFT JOIN so a missing anchor never hides a balance. Add tests for the no-witness anchor and shared genesis point cases, and assert the exact representative.

tapdb/sqlc/queries/assets.sql
```diff
@@ -191,9 +191,10 @@ INSERT INTO asset_groups (
 ) VALUES (
     $1, $2, $3, $4, $5, $6
 ) ON CONFLICT (tweaked_group_key)
-    -- This is not a NOP, update the genesis point ID in case it wasn't set
-    -- before.
-    DO UPDATE SET genesis_point_id = EXCLUDED.genesis_point_id
+    -- Preserve the original genesis_point_id (the anchor's genesis point).
+    -- We update version to ensure RETURNING works, but genesis_point_id
+    -- must remain unchanged as it identifies the canonical group anchor.
+    DO UPDATE SET version = EXCLUDED.version
```
GeorgeTsagk, line 197: Unrelated to the PR and untested. This changes what every group upsert does (mint, proof import, universe, address book) and now also overwrites version on conflict. Please revert and fix the query instead.

docs/release-notes/release-notes-0.8.0.md
```diff
@@ -368,6 +368,13 @@
   clients to distinguish grouped fungible burns from grouped collectible
   burns.
 
+- [PR#2130](https://github.com/lightninglabs/taproot-assets/pull/2130)
+  adds `asset_genesis` and `decimal_display` fields to `AssetGroupBalance`
```
GeorgeTsagk, line 372: Belongs in release-notes-0.9.0.md.

GeorgeTsagk (changes requested):
Thanks for sticking with this one. A few housekeeping items first, they make the PR much easier to pick up for whoever reviews next:

1. Rebase on main. The branch is 18 commits behind and GitHub shows it as conflicting.
2. Move the release note to release-notes-0.9.0.md, 0.8.x is already out.
3. Rewrite the commit message so it matches the code. It still describes the MIN(gen_asset_id) approach and mentions merge conflict fixing.
4. Drop the UpsertAssetGroupKey ON CONFLICT change. It is not part of what this PR advertises and it changes behavior for every group upsert path.

Inline I found two correctness issues, one of which drops groups from the response. Details on the lines.

docs/release-notes/release-notes-0.8.0.md
kaldun-tech: Moved to 0.9

tapdb/sqlc/queries/assets.sql
kaldun-tech: Reworked along the lines you suggested. From Opus:
1. `group_anchor` now picks the representative from `key_group_info_view`, so only witnessed tranches are considered, collapsed to one row per group with `MIN(witness_id)`. It no longer goes through    `asset_groups.genesis_point_id`, so repointing the anchor at an unreceived tranche can't drop the group. It's also `LEFT JOIN`ed, so a missing  representative never hides a balance.
2. `MIN(witness_id)` makes the pick deterministic when tranches share a genesis point, independent of row order or backend.

Tests:
- `TestQueryAssetBalancesByGroupNoWitnessAnchor` reproduces the address-then-receive case and asserts the balance and representative come from the witnessed tranche.
- `TestQueryAssetBalancesByGroupSharedGenesisPoint` mints two tranches on one genesis point, asserts the first-witnessed tranche is the representative, and checks the pick is stable across calls. It passes on both SQLite and Postgres in CI.

kaldun-tech: On the `UpsertAssetGroupKey`: Reverted. this is now identical to main, and the fix is entirely in `QueryAssetBalancesByGroup`. The scenario you described is covered by `TestQueryAssetBalancesByGroupNoWitnessAnchor`, which repoints `genesis_point_id` using main's existing `ON CONFLICT DO UPDATE`

rpcserver/rpcserver.go
kaldun-tech: Fixed. `DecDisplayForAssetID` now wraps with `%w`, so the `ErrAssetMetaNotFound` branch matches and a group whose representative has no meta leaves `decimal_display` unset instead of failing the whole call. Added `TestBook_DecDisplayForAssetID_NoMeta` to assert the `errors.Is` chain holds.

On the federation sync: agreed. A miss isn't cached, so with the syncer wired in, every call for a meta-less asset retries `SyncAssetInfo`. That's existing `DecDisplayForAssetID` behavior shared by the other rpcserver call sites (`ListAssets` etc.), so I'd suggest a follow-up that caches misses in the address book rather than growing this PR. Can fold it in here if you'd prefer

tapdb/sqlc/queries/assets.sql
kaldun-tech: Agreed that `gen_asset_id` order was the wrong signal. I first switched to joining through `asset_groups.genesis_point_id` as you suggested, but George found that pointer can move to a tranche this node has no witness for (e.g. after creating an address for a new tranche), and the inner join then dropped the group from the response. The representative is now chosen from witnessed rows with `MIN(witness_id)`, the same ordering as the `FetchGroupByGroupKey` query you pointed to. Details are in George's thread on this query.

kaldun-tech:
@GeorgeTsagk @darioAnongba thanks for the thorough reviews. Everything should now be addressed

Side note, I'm working on an AI startup contract which keeps me pretty busy. Aiming to get more contributions rolling when I figure out time management.

Housekeeping:
1. Rebased on main
2. Release note moved to `release-notes-0.9.0.md`
3. Commit message rewritten to describe the final design
4. `UpsertAssetGroupKey` change dropped

Correctness:
- `DecDisplayForAssetID` wraps with `%w`, so the missing-meta branch actually matches, with a test
- The `group_anchor` CTE picks the representative from witnessed rows using `MIN(witness_id)` and is `LEFT JOIN`ed, with tests for the no-witness-anchor and shared-genesis-point cases
- Dario's error-handling suggestion is in. His `genesis_point_id` suggestion was replaced by the approach above, explained inline.

One open question inline: misses in `DecDisplayForAssetID` aren't cached, so each call re-runs the federation sync. I'd suggest handling that in a follow-up PR. Let me know if you'd rather it land here.

kaldun-tech:
@GeorgeTsagk @jtobin could one of you re-run the failed `unit-cover` job? Looks like a flake
https://github.com/lightninglabs/taproot-assets/actions/runs/34720486028/job/103625321106

It failed on `TestWatcherShrinkingReorg` in `tapreorg`, which this PR doesn't touch. The failure was a `require.Eventually` timeout at `service_test.go:1446`, where it waits for the anchoring to go back to
`Unwitnessed` after the watcher restarts. The same code passed `unit-cover` on the previous push; the only change since was a one-line release-note fix.

I couldn't reproduce it locally:
- `make unit pkg=tapreorg case=TestWatcherShrinkingReorg nocache=1`: 11/11
  passes, ~1s each (the CI run took 34s before timing out)
- `make unit-race pkg=tapreorg nocache=1`: pass

Looks like a load-related flake under `unit-cover`'s parallel run.

litbot-9000:
@kaldun-tech, remember to re-request review from reviewers when ready

PR #2132, opened 2026-05-13
Fixing porter can strand parcels until daemon restart

### LINKED ISSUE #2120: Porter can strand parcels until daemon restart

sergey3bv:
Should close #2120

### DIFF
--- a/itest/custom_channels/force_close_test.go
+++ b/itest/custom_channels/force_close_test.go
@@ -235,7 +235,9 @@ func testCustomChannelsForceClose(ctx context.Context,
 	// This'll sweep his non-delay commitment output. We use the txid from
 	// the mined block (not the mempool) to avoid RBF mismatches.
 	daveSweepBlocks := mineBlocks(t, net, 1, 1)
-	daveSweepTxHash := daveSweepBlocks[0].Transactions[1].TxHash()
+	daveSweepTxHash := resolveMinedTransferTxid(
+		t.t, dave, daveSweepBlocks[0],
+	)
 
 	t.Logf("Dave sweep txid: %v", daveSweepTxHash)
 
--- a/itest/custom_channels/helpers.go
+++ b/itest/custom_channels/helpers.go
@@ -1384,7 +1384,11 @@ func waitForSendEvent(t *testing.T,
 
 	for {
 		sendEvent, err := sendEvents.Recv()
-		require.NoError(t, err)
+		if err != nil {
+			t.Logf("send event stream ended before state %v: %v",
+				expectedState, err)
+			return
+		}
 
 		if sendEvent.SendState == expectedState.String() {
 			return
@@ -1422,12 +1426,12 @@ func closeAssetChannelAndAssert(t *ccHarnessTest,
 	ctxt, cancel := context.WithTimeout(ctxb, wait.DefaultTimeout)
 	defer cancel()
 
-	closeStream, _, err := net.CloseChannel(local, chanPoint, false)
-	require.NoError(t.t, err)
-
 	sendEvents, err := local.SubscribeSendEvents(
 		ctxt, &taprpc.SubscribeSendEventsRequest{},
 	)
+	require.NoError(t.t, err)
+
+	closeStream, _, err := net.CloseChannel(local, chanPoint, false)
 	require.NoError(t.t, err)
 
 	assertWaitingCloseChannelAssetData(t.t, local, chanPoint)
@@ -2530,8 +2534,17 @@ func assertSpendableBalance(t *testing.T, client *itest.IntegratedNode,
 func locateAssetTransfers(t *testing.T, node *itest.IntegratedNode,
 	txid chainhash.Hash) *taprpc.AssetTransfer {
 
-	var transfer *taprpc.AssetTransfer
-	err := wait.NoError(func() error {
+	var (
+		transfer        *taprpc.AssetTransfer
+		lastErr         error
+		transferCount   int
+		blockHashSet    bool
+		blockHeight     uint32
+		blockHeightHint uint32
+		pollInterval    = 200 * time.Millisecond
+	)
+
+	require.Eventually(t, func() bool {
 		ctxb := context.Background()
 		forceCloseTransfer, err := node.ListTransfers(
 			ctxb, &taprpc.ListTransfersRequest{
@@ -2539,27 +2552,110 @@ func locateAssetTransfers(t *testing.T, node *itest.IntegratedNode,
 			},
 		)
 		if err != nil {
-			return fmt.Errorf("unable to list %v transfers: %w",
+			lastErr = fmt.Errorf("unable to list %v transfers: %w",
 				node.Cfg.Name, err)
+			return false
 		}
-		if len(forceCloseTransfer.Transfers) != 1 {
-			return fmt.Errorf("%v is expecting %d transfers, "+
-				"has %d", node.Cfg.Name, 1,
-				len(forceCloseTransfer.Transfers))
+
+		transferCount = len(forceCloseTransfer.Transfers)
+		if transferCount != 1 {
+			lastErr = fmt.Errorf("%v expected %d transfers, got %d",
+				node.Cfg.Name, 1, transferCount)
+			return false
 		}
 
 		transfer = forceCloseTransfer.Transfers[0]
+		blockHashSet = transfer.AnchorTxBlockHash != nil
+		blockHeight = transfer.AnchorTxBlockHeight
+		blockHeightHint = transfer.AnchorTxHeightHint
 
-		if transfer.AnchorTxBlockHash == nil {
-			return fmt.Errorf("missing anchor block hash, " +
-				"transfer not confirmed")
+		// Anchor tx confirmation metadata is populated asynchronously.
+		if !blockHashSet || blockHeight == 0 || blockHeightHint == 0 {
+			lastErr = fmt.Errorf("transfer not confirmed yet")
+			return false
 		}
 
-		return nil
-	}, ccTransferTimeout)
-	require.NoError(t, err)
+		lastErr = nil
+		return true
+	}, ccTransferConfirmTimeout, pollInterval,
+		"failed to locate confirmed transfer for %v (node=%v): "+
+			"transfers=%d block_hash_set=%v block_height=%d "+
+			"height_hint=%d last_err=%v",
+		txid, node.Cfg.Name, transferCount, blockHashSet, blockHeight,
+		blockHeightHint, lastErr,
+	)
 
 	return transfer
+}
+
+// resolveMinedTransferTxid returns the transaction ID from a mined block that
+// is associated with an asset transfer for the given node.
+func resolveMinedTransferTxid(t *testing.T, node *itest.IntegratedNode,
+	block *wire.MsgBlock) chainhash.Hash {
+
+	t.Helper()
+
+	require.Greater(
+		t, len(block.Transactions), 1,
+		"expected at least one non-coinbase transaction in block",
+	)
+
+	candidateTxids := make([]chainhash.Hash, 0, len(block.Transactions)-1)
+	for i := 1; i < len(block.Transactions); i++ {
+		candidateTxids = append(
+			candidateTxids, block.Transactions[i].TxHash(),
+		)
+	}
+
+	var (
+		matchedTxid chainhash.Hash
+		matchCount  int
+		lastErr     error
+	)
+
+	require.Eventually(t, func() bool {
+		matchCount = 0
+		lastErr = nil
+
+		ctxb := context.Background()
+		for _, txid := range candidateTxids {
+			transfers, err := node.ListTransfers(
+				ctxb, &taprpc.ListTransfersRequest{
+					AnchorTxid: txid.String(),
+				},
+			)
+			if err != nil {
+				lastErr = fmt.Errorf(
+					"unable to list %v transfers: %w",
+					node.Cfg.Name, err,
+				)
+				return false
+			}
+
+			if len(transfers.Transfers) == 0 {
+				continue
+			}
+
+			matchedTxid = txid
+			matchCount++
+		}
+
+		if matchCount != 1 {
+			lastErr = fmt.Errorf(
+				"expected 1 matching transfer tx, got %d",
+				matchCount,
+			)
+			return false
+		}
+
+		return true
+	}, ccTransferConfirmTimeout, 200*time.Millisecond,
+		"failed to resolve transfer tx in mined block for %v: "+
+			"candidates=%v matches=%d last_err=%v",
+		node.Cfg.Name, candidateTxids, matchCount, lastErr,
+	)
+
+	return matchedTxid
 }
 
 // ---------------------------------------------------------------------------
@@ -3087,8 +3183,12 @@ func assertForceCloseSweeps(ctx context.Context,
 	// Wait for tapd to process the confirmed sweep transactions before
 	// checking balances. We extract the txid from the mined blocks rather
 	// than from the earlier mempool checks to avoid RBF mismatches.
-	bobSweepTxHash1 := bobSweepBlocks1[0].Transactions[1].TxHash()
-	bobSweepTxHash2 := bobSweepBlocks2[0].Transactions[1].TxHash()
+	bobSweepTxHash1 := resolveMinedTransferTxid(
+		t.t, bob, bobSweepBlocks1[0],
+	)
+	bobSweepTxHash2 := resolveMinedTransferTxid(
+		t.t, bob, bobSweepBlocks2[0],
+	)
 	locateAssetTransfers(t.t, bob, bobSweepTxHash1)
 	locateAssetTransfers(t.t, bob, bobSweepTxHash2)
 
@@ -3120,7 +3220,9 @@ func assertForceCloseSweeps(ctx context.Context,
 
 	// Wait for tapd to register the to-local sweep transfer. We use the
 	// txid from the mined block to avoid RBF mismatches.
-	aliceToLocalHash := aliceToLocalBlocks[0].Transactions[1].TxHash()
+	aliceToLocalHash := resolveMinedTransferTxid(
+		t.t, alice, aliceToLocalBlocks[0],
+	)
 	locateAssetTransfers(t.t, alice, aliceToLocalHash)
 
 	t.Logf("Confirming Alice's to-local sweep")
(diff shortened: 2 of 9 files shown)


itest/custom_channels/force_close_test.go
```diff
@@ -235,7 +235,9 @@ func testCustomChannelsForceClose(ctx context.Context,
 	// This'll sweep his non-delay commitment output. We use the txid from
 	// the mined block (not the mempool) to avoid RBF mismatches.
 	daveSweepBlocks := mineBlocks(t, net, 1, 1)
-	daveSweepTxHash := daveSweepBlocks[0].Transactions[1].TxHash()
+	daveSweepTxHash := resolveMinedTransferTxid(
+		t.t, dave, daveSweepBlocks[0],
+	)
```
kaldun-tech, line 240: We could use more context on the design decisions here. The old code failed with RBF (replace-by-fee) because it assumed the coinbase transaction at index 0, and sweep transaction is at index 1 in the block. And there's only one non-coinbase transaction in the block. The RBF sweep transactions invalidated this, or if multiple transactions are mined in the same block.

The new approach with resolveMinedTransferTxid finds which transaction in the block tapd know is a transfer. So decouples the tests from block transaction ordering, RBF replacements, multiple transactions in a block and tapd timing issues.

itest/custom_channels/helpers.go
```diff
@@ -1422,14 +1426,14 @@ func closeAssetChannelAndAssert(t *ccHarnessTest,
...
-	require.NoError(t.t, err)
-
 	sendEvents, err := local.SubscribeSendEvents(
 		ctxt, &taprpc.SubscribeSendEventsRequest{},
 	)
 	require.NoError(t.t, err)
 
+	closeStream, _, err := net.CloseChannel(local, chanPoint, false)
+	require.NoError(t.t, err)
+
```
kaldun-tech, line 1436: Moving the close after the SubscribeSendEvents fixes a race condition:
1. `CloseChannel` starts the close process
2. tapd processes close, fires `SendStateComplete` event
3. `SubscribeSendEvents` called — but event already fired
4. `waitForSendEvent` blocks forever waiting for missed event

By subscribing first, all events from the close operation are captured. This implements the standard "subscribe before action" pattern

So it's good we could use more comments IMO

itest/custom_channels/helpers.go
```diff
@@ -2530,38 +2534,130 @@ func assertSpendableBalance(t *testing.T, client *itest.IntegratedNode,
...
+		transfer        *taprpc.AssetTransfer
+		lastErr         error
+		transferCount   int
+		blockHashSet    bool
+		blockHeight     uint32
+		blockHeightHint uint32
+		pollInterval    = 200 * time.Millisecond
+	)
+
+	require.Eventually(t, func() bool {
```
kaldun-tech, line 2547: This will get us more detailed failure messages about the state and make the testing more robust

itest/custom_channels/helpers.go
```diff
@@ -2530,38 +2534,130 @@ func assertSpendableBalance(t *testing.T, client *itest.IntegratedNode,
...
-	var transfer *taprpc.AssetTransfer
-	err := wait.NoError(func() error {
+	var (
+		transfer        *taprpc.AssetTransfer
+		lastErr         error
+		transferCount   int
+		blockHashSet    bool
+		blockHeight     uint32
+		blockHeightHint uint32
+		pollInterval    = 200 * time.Millisecond
```
kaldun-tech, line 2544: Minor: 200 millis gets used in a few different spots, consider intorducing a constant

tapfreighter/chain_porter.go
```diff
@@ -396,18 +433,166 @@ func (p *ChainPorter) advanceState(pkg *sendPackage, kit *parcelKit) {
...
+// isRecoverablePostDeliveryState returns true if a state can be retried after
+// the caller has already received a response.
+func isRecoverablePostDeliveryState(state SendState) bool {
+	switch state {
+	case SendStateWaitTxConf:
+		fallthrough
+	case SendStateStorePostAnchorTxConf:
+		fallthrough
+	case SendStateTransferProofs:
+		return true
```
kaldun-tech, line 489: Minor: The fallthrough style is unusual. More idiomatic:
```
switch state {
case SendStateWaitTxConf, SendStateStorePostAnchorTxConf, SendStateTransferProofs:
    return true
default:
    return false
}
```

Or even simpler:

```
return state == SendStateWaitTxConf ||
    state == SendStateStorePostAnchorTxConf ||
    state == SendStateTransferProofs
```

tapfreighter/chain_porter.go
```diff
@@ -396,18 +433,166 @@ func (p *ChainPorter) advanceState(pkg *sendPackage, kit *parcelKit) {
...
+		}
+	}
+
+	return delay
+}
+
+// schedulePostDeliveryRetry schedules a pending parcel retry when a recoverable
+// post-delivery state fails.
+func (p *ChainPorter) schedulePostDeliveryRetry(pkg *sendPackage,
+	failedState SendState, stateErr error) bool {
```
kaldun-tech, line 514: It seems like the boolean return value conflates multiple concepts:

- False means the state is not recoverable, OR nil package, OR max attempts exceeded
- True means retry was scheduled

So it works correctly but the design is non-standard. Consider renaming to trySchedulePostDeliveryRetry, tweak the return value, or add comments explainng the return value semantics

tapfreighter/chain_porter.go
```diff
@@ -145,6 +158,13 @@ type ChainPorter struct {
 	// subscriberMtx guards the subscribers map.
 	subscriberMtx sync.Mutex
 
+	// postDeliveryRetryMtx guards postDeliveryRetryAttempts.
+	postDeliveryRetryMtx sync.Mutex
+
+	// postDeliveryRetryAttempts tracks retry attempts for recoverable
+	// post-delivery errors by anchor txid.
+	postDeliveryRetryAttempts map[chainhash.Hash]uint8
```
kaldun-tech, line 166: Note that this is in-memory only which IMO is fine. It prevents infinite loops.

tapfreighter/chain_porter.go
```diff
@@ -379,12 +400,28 @@ func (p *ChainPorter) advanceState(pkg *sendPackage, kit *parcelKit) {
...
 			log.Errorf("Error evaluating state (%v): %v",
 				pkg.SendState, err)
 
+			sendEventPkg := *pkg
+			if failedPkg != nil {
+				sendEventPkg = *failedPkg
+			}
 			p.publishSubscriberEvent(newAssetSendErrorEvent(
-				err, stateToExecute, *pkg,
+				err, stateToExecute, sendEventPkg,
```
kaldun-tech, line 424: Minor: We publish all error events to subscribers here including both recoverable and nonrecoverable. Consider a `Recoverable bool` field to error events or suppress the error events when retry is scheduled.

tapfreighter/chain_porter.go
```diff
@@ -396,18 +433,166 @@ func (p *ChainPorter) advanceState(pkg *sendPackage, kit *parcelKit) {
...
+
+	go func() {
+		timer := time.NewTimer(delay)
+		defer timer.Stop()
+
+		select {
+		case <-timer.C:
+			pendingRetry := Parcel(pendingParcel)
+			if !fn.SendOrQuit(
+				p.outboundParcels, pendingRetry, p.Quit,
```
kaldun-tech, line 565: Minor: This correctly handles the case where `p.Quit` is closed while retry goroutine is sleeping, then `outboundParcels` send is attempted. Consider adding a test case for this.

tapfreighter/chain_porter.go
```diff
@@ -396,18 +433,166 @@ func (p *ChainPorter) advanceState(pkg *sendPackage, kit *parcelKit) {
...
+
+	return true
+}
+
+// clearPostDeliveryRetry removes any retry bookkeeping for a parcel.
+func (p *ChainPorter) clearPostDeliveryRetry(pkg *sendPackage) {
+	if pkg == nil || pkg.OutboundPkg == nil ||
+		pkg.OutboundPkg.AnchorTx == nil {
+
+		return
```
kaldun-tech, line 588: Invariant: we can only add to postDeliveryRetryAttempts if AnchorTx is non-nil, and only need to delete it in the same case

tapfreighter/chain_porter.go
```diff
@@ -2022,6 +2207,17 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
...
+		ctx, cancel = p.WithCtxQuitNoTimeout()
+		defer cancel()
+
+		err = p.importLocalAddresses(ctx, currentPkg.OutboundPkg)
+		if err != nil {
+			p.unlockInputs(ctx, &currentPkg)
+
+			return nil, fmt.Errorf("unable to import local "+
+				"addresses: %w", err)
+		}
```
kaldun-tech, line 2219: This block imports the taproot output keys into lnd's wallet for outputs that belong to us. So the fix ensures addresses are always imported, regardless of who broadcasts.

It's a solid change. Think it would benefit from more comments.

tapfreighter/chain_porter.go
```diff
@@ -2022,6 +2207,17 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
 				"disk: %w", err)
 		}
 
+		ctx, cancel = p.WithCtxQuitNoTimeout()
```
kaldun-tech, line 2210: Gives us a new context for importLocalAddresses which is idempotent and can be retried. As opposed to the CtxBlocking context (line 2189) for LogPendingParcel which is a critical write to disk that must complete even during shutdown.

tapfreighter/chain_porter_test.go
```diff
@@ -108,3 +125,107 @@ func TestVerifySplitCommitmentWitnesses(t *testing.T) {
...
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(200 * time.Millisecond):
+		t.Fatalf("advanceState blocked on background error delivery")
+	}
+}
+
+func TestAdvanceStatePermanentFailureClearsRetryBookkeeping(t *testing.T) {
```
kaldun-tech, line 152: Nit: The name suggests "permanent failure after max attempts". It actually tests non-recoverable state failure

tapfreighter/chain_porter_test.go
```diff
@@ -108,3 +125,107 @@ func TestVerifySplitCommitmentWitnesses(t *testing.T) {
...
+
+	_, exists := porter.postDeliveryRetryAttempts[txID]
+	require.False(t, exists)
+
+	select {
+	case <-porter.outboundParcels:
+		t.Fatalf("did not expect pending parcel re-queue")
+	case <-time.After(100 * time.Millisecond):
+	}
+}
```
kaldun-tech, line 231: Minor: Consider adding a test case for process shutdown similar to
`func TestSchedulePostDeliveryRetryShutdownDuringDelay(t *testing.T) {
    porter := newTestChainPorter()
    pkg := newTestSendPackage(SendStateTransferProofs)

    recoverable := porter.schedulePostDeliveryRetry(
        pkg, SendStateTransferProofs, errors.New("failure"),
    )
    require.True(t, recoverable)

    // Close quit before timer fires (base delay is 1 second)
    close(porter.Quit)

    // Verify goroutine exits cleanly, no re-queue
    select {
    case <-porter.outboundParcels:
        t.Fatalf("did not expect parcel after shutdown")
    case <-time.After(1500 * time.Millisecond):
        // Success
    }
}`

kaldun-tech (approved):
I think the change here is solid and addresses the failure case. Some minor suggestions related to improving documentation and adding a test case for shutdown

litbot-9000:

jtobin:
Closing as obviated by some combination of #2159 and #2160.

PR #2133, opened 2026-05-13
tapchannel: fix immediate force-close proof sync and sweep anchoring

GeorgeTsagk:
## Description

This PR fixes immediate post-funding (N=0) force-close recovery by correctly re-syncing active vPacket outputs with re-anchored proofs when anchor indices are stale, and hardens anchor resolution to return explicit errors instead of panicking on out-of-range indices.

~It also adds an integration test that force-closes right after funding confirmation and verifies the swept assets remain spendable via an onward transfer to a third node.~

~Depends on https://github.com/lightningnetwork/lnd/pull/10804~

These last parts of the PR now moved to https://github.com/lightninglabs/taproot-assets/pull/2177

### DIFF
--- a/tapchannel/aux_funding_controller.go
+++ b/tapchannel/aux_funding_controller.go
@@ -585,7 +585,36 @@ func newCommitBlobAndLeaves(pendingFunding *pendingAssetFunding,
 		localAssets = chanAssets
 	}
 
+	// Compute the initial BTC balances so the funding-time allocations
+	// sort into the same output order as the real commitment transaction
+	// produced at force-close. If these are left at zero, the asset proofs
+	// can anchor at the wrong index and fail re-anchor at force close.
+	commitFee := pendingFunding.feeRate.FeeForWeight(
+		lnwallet.CommitWeight(lndOpenChan.ChanType),
+	)
+	commitReserve := commitFee
+	if lndOpenChan.ChanType.HasAnchors() {
+		commitReserve += 2 * lnwallet.AnchorSize
+	}
+
+	capacityMSat := lnwire.NewMSatFromSatoshis(lndOpenChan.Capacity)
+	pushMSat := lnwire.NewMSatFromSatoshis(pendingFunding.pushAmt)
+	reserveMSat := lnwire.NewMSatFromSatoshis(commitReserve)
+	if capacityMSat < reserveMSat+pushMSat {
+		return nil, lnwallet.CommitAuxLeaves{}, fmt.Errorf("invalid "+
+			"initial balances: capacity=%v push=%v reserve=%v",
+			lndOpenChan.Capacity, pendingFunding.pushAmt,
+			commitReserve)
+	}
+
 	var localSatBalance, remoteSatBalance lnwire.MilliSatoshi
+	if pendingFunding.initiator {
+		localSatBalance = capacityMSat - reserveMSat - pushMSat
+		remoteSatBalance = pushMSat
+	} else {
+		localSatBalance = pushMSat
+		remoteSatBalance = capacityMSat - reserveMSat - pushMSat
+	}
 
 	// We don't have a real prev state at this point, the leaf creator only
 	// needs the sum of the remote+local assets, so we'll populate that.
--- a/tapchannel/aux_sweeper.go
+++ b/tapchannel/aux_sweeper.go
@@ -1173,71 +1173,116 @@ func reanchorAssetOutputs(ctx context.Context,
 	return nil
 }
 
-// anchorOutputAllocations is a helper function that creates a set of
-// allocations for the anchor outputs. We'll use this later to create the proper
-// exclusion proofs.
-func anchorOutputAllocations(
-	keyRing *lnwallet.CommitmentKeyRing) lfn.Result[[]*tapsend.Allocation] {
+// commitOutputAllocations builds NoAssets allocations for the actual pure-BTC
+// outputs of the real commitment transaction. This lets force-close exclusion
+// proofs match the transaction that is on chain instead of assuming the first
+// two outputs are always anchors.
+func commitOutputAllocations(req lnwallet.ResolutionReq,
+	vPackets []*tappsbt.VPacket) lfn.Result[[]*tapsend.Allocation] {
 
-	anchorAlloc := func(
-		k *btcec.PublicKey) lfn.Result[*tapsend.Allocation] {
+	if req.CommitTx == nil {
+		return lfn.Err[[]*tapsend.Allocation](
+			fmt.Errorf("commit tx not set"),
+		)
+	}
 
-		anchorTree, err := input.NewAnchorScriptTree(k)
-		if err != nil {
-			return lfn.Err[*tapsend.Allocation](err)
+	assetOutputs := make(map[uint32]struct{})
+	for _, vPkt := range vPackets {
+		for _, vOut := range vPkt.Outputs {
+			assetOutputs[vOut.AnchorOutputIndex] = struct{}{}
+		}
+	}
+
+	findOutputIndex := func(pkScript []byte) (uint32, bool) {
+		for idx, txOut := range req.CommitTx.TxOut {
+			if bytes.Equal(txOut.PkScript, pkScript) {
+				return uint32(idx), true
+			}
 		}
 
-		sibling, scriptTree, err := LeavesFromTapscriptScriptTree(
-			anchorTree,
+		return 0, false
+	}
+
+	newNoAssetAlloc := func(desc input.ScriptDescriptor) (
+		*tapsend.Allocation, error) {
+
+		sibling, scriptTree, err := LeavesFromTapscriptScriptTree(desc)
+		if err != nil {
+			return nil, err
+		}
+
+		pkScript, err := txscript.PayToTaprootScript(
+			scriptTree.TaprootKey,
 		)
 		if err != nil {
-			return lfn.Err[*tapsend.Allocation](err)
+			return nil, err
 		}
 
-		return lfn.Ok(&tapsend.Allocation{
-			Type:           tapsend.AllocationTypeNoAssets,
-			Amount:         0,
-			BtcAmount:      lnwallet.AnchorSize,
+		outputIndex, ok := findOutputIndex(pkScript)
+		if !ok {
+			return nil, nil
+		}
+		if _, hasAssets := assetOutputs[outputIndex]; hasAssets {
+			return nil, nil
+		}
+
+		return &tapsend.Allocation{
+			Type:        tapsend.AllocationTypeNoAssets,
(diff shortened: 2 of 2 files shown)


kaldun-tech:
@GeorgeTsagk I had a failure in the CI run of #2130. Claude & Copilot pointed me towards this PR as the likely cause https://github.com/lightninglabs/taproot-assets/actions/runs/26522178163/job/78117493175?pr=2130

EDIT: Since this is not merged it indicates that the failure is due to flakiness, possibly fixed by sergey's PR #2060

litbot-9000:
@jtobin: review reminder

tapchannel/commitment.go
```diff
@@ -29,9 +29,34 @@ import (
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/tlv"
 	"golang.org/x/exp/maps"
 )
 
+const (
+	commitSweepWitnessScriptType tlv.Type = 31337
+	commitSweepControlBlockType  tlv.Type = 31339
```
kaldun-tech: You have some magic numbers here. Want to add comment to explain the reason?

tapfreighter/parcel.go
```diff
@@ -829,6 +840,14 @@ func outputAnchor(anchorTx *tapsend.AnchorTransaction, vOut *tappsbt.VOutput,
...
 	}
 
+	outputCount := uint32(len(anchorTx.FundedPsbt.Pkt.Outputs))
+	if vOut.AnchorOutputIndex >= outputCount {
+		return nil, fmt.Errorf(
+			"anchor output index %d out of range for "+
+				"funded psbt outputs (len=%d)",
+			vOut.AnchorOutputIndex, outputCount,
+		)
+	}
```
kaldun-tech: This is good to prevent panic on out-of-range indices

tapchannel/aux_sweeper.go
```diff
@@ -1173,71 +1225,398 @@ func reanchorAssetOutputs(ctx context.Context,
...
+			break
+		}
+
+		matchIdx := strictIdx
+		if matchIdx == -1 && fallbackCount == 1 {
+			matchIdx = fallbackIdx
+		}
+		if matchIdx == -1 {
+			continue
+		}
```
kaldun-tech: The matching logic here is subtle and challenging to understand. Think it could use some comments explaining why, for example we skip when fallbackCount > 1 rather than erroring. 

Should this log a warning when it skips? Silent skips can be hard to debug if a sweep fails

itest/custom_channels/immediate_close_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/lightningnetwork/lnd/lntest/node"
+	"github.com/lightningnetwork/lnd/lntest/port"
+	"github.com/stretchr/testify/require"
+)
+
+// testCustomChannelsImmediateClose tests that an asset channel can be funded
+// and then force closed immediately after the funding transaction confirms,
+// without any in-channel asset movement.
+func testCustomChannelsImmediateClose(ctx context.Context,
+	net *itest.IntegratedNetworkHarness, t *ccHarnessTest) {
```
kaldun-tech: Looks good to validate the N=0 force-close path end to end

tapchannel/aux_funding_controller.go
```diff
@@ -585,7 +586,39 @@ func newCommitBlobAndLeaves(pendingFunding *pendingAssetFunding,
...
+		commitWeight = input.AnchorCommitWeight
+	}
+
+	commitFee := pendingFunding.feeRate.FeeForWeight(
+		lntypes.WeightUnit(commitWeight),
+	)
+	totalFee := commitFee
+	if lndOpenChan.ChanType.HasAnchors() {
+		totalFee += 2 * lnwallet.AnchorSize
+	}
```
kaldun-tech, line 598: the totalFee here represents total reserved sats not just miner fees. the calculation is good, is there a more descriptive name for it?

itest/custom_channels/breach_test.go
```diff
@@ -195,21 +195,25 @@ func testCustomChannelsBreach(ctx context.Context,
...
 	// With the breach transaction mined, Charlie should now have a
-	// transaction in the mempool sweeping *both* commitment outputs.
-	// We use a generous timeout because Charlie needs to process the
-	// block, detect the breach, and construct the justice transaction.
-	charlieJusticeTxid, err := waitForNTxsInMempool(
-		net.Miner, 1, time.Second*30,
+	// transaction in the mempool sweeping commitment outputs.
+	// We wait for at least one tx as other background txns can race in.
+	charlieJusticeTxid, err := waitForAtLeastNTxsInMempool(
+		net.Miner, 1, time.Second*60,
```
kaldun-tech: Any particular reason to increase to 60s timeout here?

tapchannel/aux_sweeper.go
```diff
@@ -1173,71 +1225,408 @@ func reanchorAssetOutputs(ctx context.Context,
...
+		inclusionProof := updatedProof.InclusionProof
+		vOut.AnchorOutputIndex = inclusionProof.OutputIndex
+		vOut.AnchorOutputInternalKey = inclusionProof.InternalKey
+		vOut.AnchorOutputTapscriptSibling = inclusionProof.
+			CommitmentProof.TapSiblingPreimage
+		vOut.ScriptKey = updatedProof.Asset.ScriptKey
+	}
+}
+
+func syncActiveOutputProofsFromVPackets(
```
kaldun-tech: the new functions here could use header comments

tapchannel/aux_sweeper.go
```diff
@@ -1173,71 +1225,408 @@ func reanchorAssetOutputs(ctx context.Context,
...
+		return lfn.Errf[asset.ScriptKey]("unsupported commit witness "+
+			"type: %v", req.Type)
+	}
+}
+
+// isInitialCommitState returns true iff the ResolutionReq targets a
+// commitment at height 0 — i.e. an immediate post-funding force close,
+// before the channel_ready commit-point rotation.
+func isInitialCommitState(req lnwallet.ResolutionReq) bool {
+	return req.CommitHeight.UnwrapOr(1) == 0
```
kaldun-tech: The `UnwrapOr(1)` default means if `CommitHeight` is not set, we assume it's NOT the initial state with a default value of 1. This is a safe default that should be documented.

jtobin (changes requested):
I don't think you need the special-case reconstruction stuff for commit height 0 anymore.

I think that idea can be summarized as: "the data at commit height 0 is possibly bad, so before doing anything, reconstruct it to ensure it's good." But IIUC the other fixes ensure that the initial commit data is good, so it should be safe to trust.

N.b., the other fixes being: 

* lnd should populate negotiated CsvDelay and DustLimit before tapd constructs the initial commitment blob (lightningnetwork/lnd#10804),
* tapd should build initial commitment blobs with actual/nonzero BTC balances (in 5275b384), and
* tapd should walk the commitment tx to figure out where assets actually are/aren't during force close (also in 5275b384).

Is there still anything that the initial commit height special-case machinery accomplishes that those fixes don't?

If not, I vote to cut this one down: include just those latter two tapd fixes, and then add the itest in a separate PR. The tapd-side fixes don't depend on the lnd change, so we can merge/release them without having to wait for an lnd release. Eh, eh?

GeorgeTsagk:
Thanks for the review @kaldun-tech & @jtobin . Applied your recommendations and also split the itest/lnd bump part to a separate PR  https://github.com/lightninglabs/taproot-assets/pull/2177

tapchannel/aux_sweeper.go
```diff
@@ -1173,71 +1225,224 @@ func reanchorAssetOutputs(ctx context.Context,
...
+
+	return false
+}
+
+func isCommitmentOutputResolution(witnessType input.WitnessType) bool {
+	switch witnessType {
+	case input.TaprootLocalCommitSpend,
+		input.TaprootRemoteCommitSpend,
+		input.TaprootCommitmentRevoke:
+		return true
```
kaldun-tech: Would it make sense to add a method for this in LND witnessgen? Could be helpful in case new commitment output resolution types are added. Depends on whether this concept is in scope of that repo & file.

tapchannel/aux_funding_controller.go
```diff
@@ -585,7 +586,39 @@ func newCommitBlobAndLeaves(pendingFunding *pendingAssetFunding,
...
+	} else {
+		localSatBalance = pushMSat
+		remoteSatBalance = capacityMSat - feeMSat - pushMSat
+	}
+
+	if localSatBalance < 0 || remoteSatBalance < 0 {
+		return nil, lnwallet.CommitAuxLeaves{}, fmt.Errorf("invalid "+
+			"initial balances: capacity=%v push=%v fee=%v",
+			lndOpenChan.Capacity, pendingFunding.pushAmt, totalFee)
+	}
```
kaldun-tech, line 617: The `< 0` check on these fields is dead code because they are uint64. For unsigned ints in Go this is always false, Go won't even compile this. The underflow would be a rare edge case in practice, but could occur if tapd's fee calculation differs from LND's expecation.

If `capcity < fee + push` the subtraction underflows to a large positive number that passes the `< 0` check. So this creates a latent bug.

The fix is to replace with a pre-subtraction check like:
`if capacityMSat < feeMSat+pushMSat { return error }`

kaldun-tech (changes requested):
The only blocker here I see is the fee calculation underflow. Otherwise looks great with solid fixes.

jtobin (changes requested):
I still think this should be stripped down further. 😅 

There's still machinery in here to work around the issue that lightningnetwork/lnd#10804 should ultimately fix by populating the initial aux channel state appropriately, but the cure is worse than the disease, IMO.

The additional TLV records, the skipOutputProofVerify bool threaded through everything, the added CommitWitness{Script, ControlBlock} fields -- I don't think any of that will be necessary when the upstream PR lands. IIUC, the initial aux channel state will be populated properly, and then there'll be nothing new on tapd's end to do with it. The porter's normal verification procedure should just work on its own.

If we can safely assume that lnd gives us correct initial negotiated values, then the diff of the purely tapd-side fixes (for init. balances and commit tx asset locations/nonlocations) shrinks to something like (closer to like +120/-50 ish):

https://gist.github.com/jtobin/50132672406173b31ee328205175f673

Essentially I argue that "lnd gives us incorrect negotiated initial values" is an lnd problem, so the fix for it should be in lnd. Fixing it in tapd feels too gnarly.

*jtobin ducks*

jtobin (approved):
Nice, LGTM. :+1: :+1:

PR #2134, opened 2026-05-14
fix(rpcserver): normalized REST bytes base64 handling for path and query

### LINKED ISSUE #2075: rest: path and query byte fields require differently-padded base64
## Observation

While wiring up tap-sdk's REST client I found two bytes-in-URL paths
with different padding requirements:

- Path parameter: \`GET /v1/taproot-assets/assets/mint/batches/{batch_key}\`
  — accepts URL-safe base64 **without** padding (\`RawURLEncoding\`).
  Sending padded base64 or standard base64 returns zero rows with no
  400; plain hex is also silently ignored.
- Query parameter: \`group_key_filter\` / \`asset_filter\` / \`group_key\`
  — rejects unpadded URL-safe with \`illegal base64 data at input byte N\`
  and only accepts standard or URL-safe base64 **with** padding
  (grpc-gateway's default \`convert.Bytes\` tries \`StdEncoding\` then
  \`URLEncoding\`).

## Problem

1. No documentation. Every SDK author has to grep grpc-gateway to
   figure out which encoding each endpoint wants.
2. The path handler silently returns empty rows on a bad encoding. A
   filtered \`ListBatches\` call that happens to hit the wrong path
   encoding just returns \"no batches\" — which looks like the batch
   doesn't exist.

## Proposal

- Normalise: accept all four base64 variants in both path and query
  bytes fields. grpc-gateway's \`convert.Bytes\` already attempts to
  fall through encodings; the path layer can do the same.
- Alternatively, surface the encoding expectation in the OpenAPI
  (\`format: byte\` tells the reader nothing about padding).
- At minimum, make the path handler return 400 with a clear message
  when the key fails to decode, instead of a silent empty result.

## Related

Same debugging session as the \`UseHexForBytes\` documentation issue.

sergey3bv:
Should close #2075

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -102,6 +102,12 @@
   fixes inverted sort direction in `AssetRoots`, `AssetLeafKeys`, and
   `QueryEvents` universe RPCs.
 
+- [PR#2134](https://github.com/lightninglabs/taproot-assets/pull/2134)
+  Normalizes REST bytes fields across both path and query parsing so
+  `batch_key`, `asset_filter`, `group_key_filter`, and `group_key` now
+  accept standard/URL-safe base64 with or without padding. Invalid bytes
+  now return `400 Bad Request` instead of silently producing empty results.
+
 # New Features
 
 ## Functional Enhancements
--- a/itest/universe_test.go
+++ b/itest/universe_test.go
@@ -4,10 +4,15 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
+	"encoding/base64"
 	"encoding/hex"
 	"fmt"
 	"io"
 	prand "math/rand"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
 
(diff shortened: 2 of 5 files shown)


litbot-9000:

darioAnongba:
Thanks for picking this up, @sergey3bv. After looking through, we're going to close this without merging. A few reasons. Looks like I shouldn't have opened that issue as it's not the right fix:

1. Wrong direction relative to project convention. taprpc/utils.go sets `UseHexForBytes: true` for both REST marshal and unmarshal, so the REST surface is hex-first. Several path params already have explicit `{*_str}` hex variants (`asset_id_str`, `meta_hash_str`, `group_key_str`). The consistent fix for `batch_key`, `internal_key`, and `tweaked_script_key` is to add `_str` hex variants, not to teach the server to accept four base64 paddings.

2. The hardcoded maps in `rest_bytes_normalization.go` won't stay in sync. There are 207 `bytes` fields across the protos and this PR covers 5 query keys and 6 path prefixes. New routes silently inherit the old behavior.

3. The path handler returning 200 + empty result on a malformed key is a real bug worth fixing.

PR #2146, opened 2026-05-26
docs: add portfolio pilot architecture guide

Roasbeef:
In this PR, we add a new architecture doc for the portfolio pilot
subsystem under `docs/`. The doc is meant to serve as a one-stop reference
for devs (and agents) building custom market-making, OTC desks, or
exchange flows on top of the RFQ layer, in the same spirit as the existing
[docs/rfq_architecture.md](docs/rfq_architecture.md).

The portfolio pilot is the decision layer that sits between the RFQ
negotiator and the price oracle. `ResolveRequest` decides whether to take
an inbound quote and at what rate, `VerifyAcceptQuote` is our last look at
the peer's accept, and `QueryAssetRates` is pure rate discovery. The
interface is small but the policy surface it opens up is large, and the
existing source comments only get you partway.

## What the doc covers

The doc walks the `PortfolioPilot` interface, the in-process
`InternalPortfolioPilot`, and the gRPC `RpcPortfolioPilot`. From there it
covers the limit-order model the protocol enforces on every accept (rate
bounds, min/max amounts, IOC vs FOK execution policy, fill caps, and the
unified `RequestConstraints` view), and the hint/bind/qualify intent
lifecycle that distinguishes indicative pricing from binding commitments.

There's also a section pinning down the vocabulary: the order vs request
vs accept vs rate distinction, the asset-units-per-BTC convention, the
perspective flip on incoming requests (a peer's `BuyRequest` is a sell
opportunity for us), and how `RequestConstraints.RateBoundCmp` collapses
buy and sell into a single check.

## Custom pilots

The doc includes Go sketches for non-trivial pilots:

- An in-memory order-book pilot showing the resolve/verify/query path
  against resting orders.
- A circular-swap wrapper that coordinates two RFQ legs against a combined
  slippage tolerance.
- A hedging composition pattern that reserves external capacity at the
  moment of commitment.

These are illustrative, not production-ready (called out as such), but
they're complete enough that a pilot author can grow their own
implementation from there. They also point at
[docs/examples/basic-portfolio-pilot](docs/examples/basic-portfolio-pilot)
as the shortest path from interface to running daemon.

## OTC and exchange considerations

Dedicated sections cover the things an OTC desk or an exchange operator
should think about before treating tapd as a market-making front-end:
customer tiering via `PriceOracleMetadata`, spread/skew/inventory pricing,
last-look mechanics (and the caveats around them), layering a continuous
double auction matching engine onto the pilot, reconciliation of internal
trade ledgers against RFQ IDs, and crash recovery / audit requirements.

## Test plan

- [x] Doc-only change, no code touched.
- [ ] Render check on GitHub once the PR is up.

### DIFF
--- a/docs/portfolio_pilot_architecture.md
+++ b/docs/portfolio_pilot_architecture.md
@@ -70,7 +70,8 @@ a working pilot so you can grow your own from there.
 
 For the RFQ protocol that sits underneath, see
 [rfq_architecture.md](rfq_architecture.md). For a runnable RPC pilot
-implementation, see [examples/basic-portfolio-pilot/main.go](examples/basic-portfolio-pilot/main.go).
+implementation, see
+[docs/examples/basic-portfolio-pilot/main.go](examples/basic-portfolio-pilot/main.go).
 
 ## Where the Pilot Sits
 
@@ -371,9 +372,10 @@ or buggy counterparty from getting away with a rate that violates a limit
 the protocol promised to enforce.
 
 The check helpers are currently unexported in the `rfq` package; an
-external pilot has to reproduce them (the example in
-[examples/basic-portfolio-pilot/main.go](examples/basic-portfolio-pilot/main.go)
-does exactly that). The logic is small and the test coverage at
+external pilot has to re-implement the equivalent logic. The example in
+[docs/examples/basic-portfolio-pilot/main.go](examples/basic-portfolio-pilot/main.go)
+inlines its own copy of these checks rather than chaining helpers. The
+logic is small and the test coverage at
 [rfq/portfolio_pilot_test.go](../rfq/portfolio_pilot_test.go) doubles as a
 specification. Aligning a custom pilot's behavior with these tests is the
 safest way to keep its reject codes consistent with what the rest of the
@@ -386,7 +388,9 @@ pilot RPC for accept verification) and `RejectCode` (used in the wire-level
 `Reject` message that goes to the peer). They overlap but are not the same
 set.
 
-A quick lookup table:
+A quick lookup table. The proto enum prefixes (`REJECT_CODE_` on the
+wire-reject side) are elided for readability; the actual proto values
+are `REJECT_CODE_PRICE_BOUND_MISS` and so on.
 
 | Failure                       | QuoteRespStatus              | Wire RejectCode                  |
 |------------------------------|------------------------------|----------------------------------|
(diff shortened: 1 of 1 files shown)


docs/portfolio_pilot_architecture.md
```diff
@@ -0,0 +1,1139 @@
...
+    return ValidAcceptQuoteRespStatus
+}
+```
+
+Both sides of the negotiation run this. The responder's pilot runs it during
+`ResolveRequest` to avoid offering a quote the requester will reject. The
+requester's pilot runs it during `VerifyAcceptQuote` to avoid honoring an
+accept that does not match what was asked. Running both keeps a misbehaving
+or buggy counterparty from getting away with a rate that violates a limit
+the protocol promised to enforce.
```
darioAnongba, line 372: Would be good to add a warning like:
> If you ship a custom `RpcPortfolioPilot`, the local tapd performs no independent verification of limit-order constraints. You must reproduce these checks in your pilot.

Somewhere for custom Pilots. We currently do not re-validate.

Note that I'm not personally fond of this behaviour as I think we could always validate even when using a custom pilot.

docs/portfolio_pilot_architecture.md
```diff
@@ -0,0 +1,1139 @@
...
+has effectively unaccepted: by the time the HTLC arrives the rate is gone.
+The default lifetime is set by the negotiator at construction; custom pilots
+should pick a value that comfortably exceeds the round-trip a payment will
+take through the network.
+
+## The RpcPortfolioPilot
+
+For pilots that must run out of process, `RpcPortfolioPilot` is a thin
+gRPC client. It uses URIs of the form `portfoliopilotrpc://host:port` and
+dials with self-signed TLS (or insecure if explicitly requested).
```
darioAnongba, line 493: I think custom portfolio pilots still have `InsecureSkipVerify: true`. So this statement is not true yet. We should implement TLS verification aligned with price oracle here.

cc: @jtobin

darioAnongba (commented):
worth a quick review by @jtobin as well.

docs/portfolio_pilot_architecture.md
jtobin: Woah, yeah that's an oversight on my part if so. Will check (and patch).

jtobin: PP cert verification added in #2147.

litbot-9000:
@jtobin: review reminder

jtobin (approved):

PR #2147, opened 2026-05-26
rfq: auth/transport harmonization pass

jtobin:
Mostly adds configurable [TLS certificate verification](https://github.com/lightninglabs/taproot-assets/pull/2146#discussion_r3303770277) to the portfolio pilot, bringing it to parity with the price oracle, but also smooths over a few loose ends:

First, it seems righteous to have a single RFQ TLS config, so I folded the price oracle and portfolio pilot's TLS configs together into a single RFQ TLS config. So instead of separate flags, à la:

```
  experimental.rfq.{priceoracletls,portfoliopilottls}.*
```

we just deal with a single one:

```
  experimental.rfq.tls.*
```

Second, I added basic macaroon auth to portfolio pilot, again bringing it up to scratch with what we support for price oracles. But I didn't fold these together, as it seems correct for each service to have its own macaroon auth.

Also includes some misc. cleanup of existing minor issues (e.g. switching away from that deprecated gRPC call). Note also that I'll harmonize the release note with the cleaned-up, feature-first notes I have on the v0.8 branch; just want it in a standard form here.

### DIFF
--- a/rfq/cli.go
+++ b/rfq/cli.go
@@ -22,9 +22,6 @@ const (
 type CliConfig struct {
 	PriceOracleAddress string `long:"priceoracleaddress" description:"Price oracle gRPC server address (rfqrpc://<hostname>:<port>). To use the integrated mock, use the following value: use_mock_price_oracle_service_promise_to_not_use_on_mainnet"`
 
-	// PortfolioPilotAddress is the portfolio pilot gRPC server address.
-	PortfolioPilotAddress string `long:"portfoliopilotaddress" description:"Portfolio pilot gRPC server address (portfoliopilotrpc://<hostname>:<port>)"`
-
 	PriceOracleTLSDisable bool `long:"priceoracletlsdisable" description:"Disable TLS for price oracle communication."`
 
 	PriceOracleTLSInsecure bool `long:"priceoracletlsinsecure" description:"Disable price oracle certificate verification."`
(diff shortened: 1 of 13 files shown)


jtobin:
(Sorry @djkazic, accidental collateral tag.)

rfq/portfolio_pilot_rpc.go
```diff
@@ -67,39 +64,39 @@ type RpcPortfolioPilot struct {
...
 	}
 
-	return []grpc.DialOption{
-		grpc.WithTransportCredentials(creds),
+	// Create transport credentials and dial options from the supplied TLS
+	// config.
+	transportCredentials, err := configureTransportCredentials(tlsConfig)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
jtobin: Not a bad idea, but this is always non-nil in practice.

rfq/cli.go
```diff
@@ -128,5 +118,33 @@ func (c *CliConfig) Validate() error {
...
+
+	// A macaroon requires transport security. If any macaroon path is set
+	// but RFQ TLS is disabled, the gRPC dial will fail. Catch this early
+	// with a clear error.
+	macaroonSet := c.PriceOracleMacaroonPath != "" ||
+		c.PortfolioPilotMacaroonPath != ""
+	if macaroonSet && c.TLS.Disable {
+		return fmt.Errorf("macaroon paths require RFQ TLS to be " +
+			"enabled")
+	}
```
darioAnongba:
```suggestion
if macaroonSet && (c.TLS.Disable || c.TLS.Insecure) {
    return fmt.Errorf("macaroon paths require RFQ TLS to be " +
        "enabled and the server certificate to be verified " +
        "(tls.disable=false, tls.insecure=false)")
}
```
Already exists in price oracle but this way, we don't pass the macaroon when server is insecure?

rfq/oracle.go
```diff
@@ -253,7 +253,7 @@ func NewRpcPriceOracle(addrStr string, tlsConfig *TLSConfig,
 	// Formulate the server address dial string.
 	serverAddr := fmt.Sprintf("%s:%s", addr.Hostname(), addr.Port())
 
-	conn, err := grpc.Dial(serverAddr, dialOpts...)
+	conn, err := grpc.NewClient(serverAddr, dialOpts...)
```
darioAnongba, line 257: This is fine but there are some differences between `Dial` and `NewClient`. For example, the `fmt.Sprintf("%s:%s", addr.Hostname(), addr.Port())` (same in PP), should apparently become `net.JoinHostPort(addr.Hostname(), addr.Port())` to handle IPv6 better.

darioAnongba (approved):
Approved with nits.

rfq/cli.go
jtobin: Addressed in 217f2bb4b651ae15a4ea0bb148870d5a7184a49a.

rfq/oracle.go
jtobin: Addressed in 6881b077df314e06d034ada79a8e9a10edbc046b.

PR #2150, opened 2026-05-28
Add `ListPayments` / `ListInvoices` RPC

### LINKED ISSUE #2142: [feature]: Add RPCs for `ListPayments` and `ListInvoices`.
A way to retrieve Taproot Asset specific invoices and payments just like you can with LND

I think currently there's a workaround to use the `lnrpc.Route` and check the `custom_channel_data` to find asset specific data, but that's not ideal for a developer.

GeorgeTsagk:
## Description

Adds asset-aware `ListInvoices` and `ListPayments` RPCs to the TaprootAssetChannels service. Each request embeds the corresponding lnrpc list request, callers pass the same arguments they'd pass to lnd and each response item embeds the full `lnrpc.Invoice` / `lnrpc.Payment` alongside a repeated `AssetAmount` summary (asset_id, amount, and the tweaked group_key available) decoded from the HTLCs' custom records.

Results are filtered to records that actually involve a Taproot Asset; pagination offsets are passed through unchanged from lnd. Decode failures error out the whole RPC rather than silently dropping records, so corrupted data surfaces loudly. Unit tests cover the aggregation, binary-vs-JSON route decoding, error propagation, and length-validated asset IDs; a new custom-channels itest exercises the full flow end-to-end against a grouped asset.

Closes #2142

### DIFF
--- a/itest/custom_channels/list_asset_rpcs_test.go
+++ b/itest/custom_channels/list_asset_rpcs_test.go
@@ -1,0 +1,462 @@
+//go:build itest
+
+package custom_channels
+
+import (
+	"bytes"
+	"context"
+	"encoding/hex"
+	"fmt"
+	"slices"
+	"testing"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/lightninglabs/taproot-assets/itest"
+	"github.com/lightninglabs/taproot-assets/proof"
+	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	tchrpc "github.com/lightninglabs/taproot-assets/taprpc/tapchannelrpc"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lnrpc/invoicesrpc"
+	"github.com/lightningnetwork/lnd/lntest/node"
+	"github.com/lightningnetwork/lnd/lntest/port"
+	"github.com/lightningnetwork/lnd/lntest/wait"
+	"github.com/stretchr/testify/require"
+)
+
+// testCustomChannelsListInvoicesAndPayments asserts that tapd's ListInvoices
+// and ListPayments RPCs return the expected asset-aware view of invoices and
+// payments, filter out records that don't involve any asset, and pass
+// pagination offsets through from lnd.
+//
+//nolint:lll
+func testCustomChannelsListInvoicesAndPayments(_ context.Context,
+	net *itest.IntegratedNetworkHarness, t *ccHarnessTest) {
+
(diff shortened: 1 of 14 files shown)


rpcserver/rpcserver.go
```diff
@@ -10492,6 +10493,368 @@ func (r *RPCServer) AddInvoice(ctx context.Context,
...
+	)
+	if err != nil {
+		rpcsLog.Debugf("list rpc: unable to query asset info for "+
+			"asset_id %x: %v", id[:], err)
+		l.cache[id] = nil
+		return nil
+	}
+
+	var gk []byte
+	if info.GroupKey != nil {
```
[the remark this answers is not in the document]
GeorgeTsagk: `QueryAssetInfo` returns `asset.AssetGroup` by value (`address/book.go:267`), not a pointer, so `info == nil` wouldn't compile — no nil check needed here.

itest/custom_channels/list_asset_rpcs_test.go
```diff
@@ -0,0 +1,269 @@
...
+	// -----------------------------------------------------------------
+	bobInvoices, err := asTapd(bob).ListInvoices(
+		ctx, &tchrpc.ListInvoicesRequest{
+			Request: &lnrpc.ListInvoiceRequest{
+				NumMaxInvoices: 100,
+			},
+		},
+	)
+	require.NoError(t.t, err)
+	require.GreaterOrEqual(t.t, len(bobInvoices.Invoices), 1)
```
bufo24: Why not asset the length of the `Invoices` array is `2`?

GeorgeTsagk: Good catch — fixed. The test setup is deterministic (1 warm-up keysend + 1 explicit `createAssetInvoice` = exactly 2 invoices on Bob, same for Alice's payments), so I tightened both assertions to `require.Len(..., 2)`. Pushed in the updated `itest:` commit.

litbot-9000:

docs/release-notes/release-notes-0.8.0.md
```diff
@@ -220,6 +220,15 @@
   requests, and to `PortfolioPilot.ResolveRequest` for constraint
   forwarding. Add `FOK_NOT_VIABLE` to `QuoteRespStatus`.
 
+- [PR#2150](https://github.com/lightninglabs/taproot-assets/pull/2150)
```
jtobin: Should bump this note to the v0.8.1 doc.

jtobin (approved):
LGTM, only administrative thing is that the release note should be moved to the v0.8.1 doc. :+1: :+1:

docs/release-notes/release-notes-0.8.0.md
GeorgeTsagk: moved! ty

PR #2163, opened 2026-06-16
tapfreighter: resolve stranded transfers against the chain

### LINKED ISSUE #2160: tapfreighter: select pending xfers can become stranded

jtobin:
Resolves #2160.

Resolves permanently-pending transfers by querying the chain when broadcast is rejected, and by watching the inputs (not just the anchor tx) while we wait for confirmation.

The vast majority of the +LoC here is in tests. I think the "actual change" is about +400 LoC, mostly to the chain porter and asset store.

### DIFF
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -325,6 +325,12 @@ type ActiveAssetsStore interface {
 	SupersedeConflictingTransfers(ctx context.Context,
 		arg sqlc.SupersedeConflictingTransfersParams) (int64, error)
 
+	// MarkTransferSuperseded marks the unconfirmed transfer anchored by
+	// the given transaction as superseded, returning the IDs of the
+	// transfers it touched.
+	MarkTransferSuperseded(ctx context.Context, anchorTxid []byte) ([]int64,
+		error)
+
 	// QuerySupersededTransferIDs returns the IDs of all transfers that
 	// have been marked as superseded.
 	QuerySupersededTransferIDs(ctx context.Context) ([]int64, error)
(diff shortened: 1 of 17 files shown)


tapfreighter/chain_porter.go
```diff
@@ -519,45 +541,240 @@ func (p *ChainPorter) waitForConfEventOnce(ctx context.Context,
...
+
+	err := p.cfg.ExportLog.MarkTransferSuperseded(ctx, txHash)
+	if err != nil {
+		return fmt.Errorf("unable to mark transfer as superseded: %w",
+			err)
+	}
+
+	return fmt.Errorf("anchor tx %v can never confirm: superseded by "+
+		"confirmed tx %v spending its inputs", txHash, spender)
 }
```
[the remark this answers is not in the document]
jtobin: Fixed in updated changeset.

tapfreighter/chain_porter.go
```diff
@@ -519,45 +541,240 @@ func (p *ChainPorter) waitForConfEventOnce(ctx context.Context,
...
+	// Gather the outpoints the anchor transaction spends on behalf of the
+	// transfer.
+	numInputs := len(parcel.Inputs) + len(parcel.ZeroValueInputs)
+	ops := make([]wire.OutPoint, 0, numInputs)
+	for idx := range parcel.Inputs {
+		ops = append(ops, parcel.Inputs[idx].OutPoint)
+	}
+	for idx := range parcel.ZeroValueInputs {
+		ops = append(ops, parcel.ZeroValueInputs[idx].OutPoint)
+	}
```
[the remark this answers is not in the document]
jtobin: Wontfix.

universe/supplycommit/mock.go
```diff
@@ -192,13 +192,14 @@ func (m *mockChainBridge) RegisterConfirmationsNtfn(
...
+	heightHint uint32) (*chainntnfs.SpendEvent, chan error, error) {
 
 	args := m.Called(ctx, outpoint, pkScript, heightHint)
 	if args.Get(0) == nil {
-		return nil, args.Error(1)
+		return nil, nil, args.Error(2)
 	}
-	return args.Get(0).(*chainntnfs.SpendEvent), args.Error(1)
+	return args.Get(0).(*chainntnfs.SpendEvent),
+		args.Get(1).(chan error), args.Error(2)
```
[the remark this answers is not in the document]
jtobin: Wontfix.

litbot-9000:

tapfreighter/chain_porter.go
```diff
@@ -519,47 +606,646 @@ func (p *ChainPorter) waitForConfEventOnce(ctx context.Context,
...
+	// an external system (such as the lnd sweeper), where a competing
+	// transaction version or a third party (such as a remote channel
+	// party claiming an HTLC output) may win the race for the inputs
+	// without the porter ever attempting a broadcast itself.
+	//
+	// Registration is all-or-nothing: if any input cannot be watched,
+	// the caller has to re-attempt the whole set, since a partial watch
+	// could miss a foreign confirmation on the unwatched input and
+	// strand the parcel exactly as before this fix.
+	pkScripts, err := p.pkScriptsForInputs(ctx, outboundPkg)
```
GeorgeTsagk, line 627: If `pkScriptsForInputs` fails persistently, we error out before the select loop and the retry loop never reads `confNtfn.Confirmed`, so the parcel stalls even when our own anchor confirms. Pre PR the conf ntfn alone sufficed. Suggest degrading to a conf only watch with a loud log on lookup failure.

tapfreighter/chain_porter.go
```diff
@@ -519,47 +606,646 @@ func (p *ChainPorter) waitForConfEventOnce(ctx context.Context,
...
+			return nil, nil, false, fmt.Errorf("error whilst "+
+				"waiting for package tx confirmation: %w", err)
+
+		case spend := <-fanin.spends:
+			spender := spend.SpenderTxHash
+
+			// Our own anchor transaction being the confirmed
+			// spender means the confirmation event is imminent,
+			// so we keep waiting for it.
+			if spender == nil || *spender == txHash {
```
GeorgeTsagk, line 680: Nil `SpenderTxHash` is treated like a self spend and ignored. A malformed spend detail should surface as a retryable error or at least a warn, since silently ignoring it could mask the foreign spend case this PR detects.

tapfreighter/chain_porter.go
```diff
@@ -519,47 +606,646 @@ func (p *ChainPorter) waitForConfEventOnce(ctx context.Context,
...
+			"outpoint %v: %w", op, err)
+	}
+
+	go func() {
+		defer spendNtfn.Cancel()
+
+		for {
+			select {
+			case spend, ok := <-spendNtfn.Spend:
+				if !ok || spend == nil {
```
GeorgeTsagk, line 1088: A closed or nil spend stream returns silently, unwinding this input's watch, which is the stranding mode this PR fixes. The confNtfn closed case at :662 errors out; this should too, or push to `fanin.spendErrs` for re-registration.

tapfreighter/chain_porter.go
```diff
@@ -519,47 +606,646 @@ func (p *ChainPorter) waitForConfEventOnce(ctx context.Context,
...
+
+	return nil
+}
+
+// locateConfirmedInputSpend queries the chain for a confirmed transaction
+// spending any of the given parcel's inputs. It returns the txid of the
+// confirmed spender, which may be the parcel's own anchor transaction. A nil
+// return means no confirmed spend could be located before the query timed
+// out: the inputs are either unspent, or their spend hasn't confirmed yet.
+func (p *ChainPorter) locateConfirmedInputSpend(ctx context.Context,
```
GeorgeTsagk, line 1139: nit: no `fanin.reorgs` case in this select: a spend, reorg, spend cadence can fill the reorg buffer and block forwarding until timeout. Safe but wasteful; a drain case would keep it responsive.

tapfreighter/chain_porter.go
```diff
@@ -519,47 +606,646 @@ func (p *ChainPorter) waitForConfEventOnce(ctx context.Context,
...
+	// pending entry is ignored.
+	finality := make(chan finalityResult, len(outboundPkg.Inputs)+
+		len(outboundPkg.ZeroValueInputs))
+
+	for {
+		select {
+		case confEvent, ok := <-confNtfn.Confirmed:
+			if !ok || confEvent == nil {
+				return nil, nil, false, fmt.Errorf(
+					"confirmation event channel closed "+
```
GeorgeTsagk, line 664: Worth a comment noting the scope boundary: spend watches tear down at our anchor's 1 conf, so a post confirmation reorg that orphans the anchor still strands the DB on the dead anchor (the #2179 window). Fine to defer, but the release note should say "pre confirmation" stranded transfers.

tapdb/assets_store.go
```diff
@@ -3846,6 +3852,180 @@ func (a *AssetStore) QueryParcels(ctx context.Context,
 	)
 }
 
+// MarkTransferSuperseded marks the unconfirmed transfer anchored by the given
+// transaction as superseded: a conflicting transaction has confirmed spending
```
GeorgeTsagk, line 3856: nit: superseded transfers are only filtered in `PendingParcels` and never surfaced via RPC, so `ListTransfers` shows them as eternally "unconfirmed". Consider exposing the flag, follow up is fine.

GeorgeTsagk (changes requested):
Reorg safety looks solid overall: SafeDepth gating, per input finality watches, and reversibility until depth are all correct, and this fixes #2160 without worsening #2179. Could we get a test for the spender replacement path (`chain_porter.go:700-707`: foreign spend, reorg, different spender on same input, SafeDepth)? The cancel and replace logic is the subtlest part and currently untested. A persistent `pkScriptsForInputs` failure test would also pin the hostage scenario noted inline at :627.

jtobin:
This one has actually been subsumed / obviated by follow-on work that I haven't made a PR for yet. I wasn't super keen on the *ad hoc* reorg handling it required, so the replacement will be more holistic. I've been keeping the draft PR open intentionally in the meantime until the other stuff properly replaces it.

jtobin:
Obviated by #2266.

PR #2165, opened 2026-06-17
multi: add nil guards

jtobin:
Merely adds a few checks for things that could be nil, previously missing.

### DIFF
--- a/proof/tx.go
+++ b/proof/tx.go
@@ -358,6 +358,12 @@ func UnmarshalTxProof(
 			"missing")
 	}
 
+	if len(mp.SiblingHashes) > MerkleProofMaxNodes {
+		return nil, fmt.Errorf("merkle proof has too many nodes: "+
+			"%d > %d", len(mp.SiblingHashes),
+			MerkleProofMaxNodes)
+	}
+
 	if len(mp.SiblingHashes) != len(mp.Bits) {
 		return nil, fmt.Errorf("merkle proof sibling hashes and " +
 			"bits length mismatch")
--- a/tapchannel/aux_leaf_signer.go
+++ b/tapchannel/aux_leaf_signer.go
@@ -649,6 +649,10 @@ func (v *schnorrSigValidator) ValidateWitnesses(newAsset *asset.Asset,
 
 	for idx := range newAsset.PrevWitnesses {
 		witness := newAsset.PrevWitnesses[idx]
+		if witness.PrevID == nil {
+			return fmt.Errorf("%w: nil prev_id at "+
+				"witness_idx=%d", vm.ErrNoInputs, idx)
+		}
 		prevAsset, ok := prevAssets[*witness.PrevID]
 		if !ok {
 			return fmt.Errorf("%w: no prev asset for "+
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -261,6 +261,9 @@ func (vm *Engine) validateSplit(splitAsset *commitment.SplitAsset) error {
 	// prevID of zero, as the inherit the prev ID from the root asset.
 	//
 	// TODO(roasbeef): revisit post multi input
+	if len(vm.newAsset.PrevWitnesses) == 0 {
+		return fmt.Errorf("%w: prev witness zero", ErrNoInputs)
+	}
 	rootWitness := vm.newAsset.PrevWitnesses[0]
 	splitWitness := splitAsset.PrevWitnesses[0]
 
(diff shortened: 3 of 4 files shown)


vm/vm.go
```diff
@@ -260,6 +264,11 @@ func (vm *Engine) validateSplit(splitAsset *commitment.SplitAsset) error {
 	rootWitness := vm.newAsset.PrevWitnesses[0]
 	splitWitness := splitAsset.PrevWitnesses[0]
 
+	if rootWitness.PrevID == nil {
+		return fmt.Errorf("%w: nil prev_id on root witness",
+			ErrNoInputs)
+	}
```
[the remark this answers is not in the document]
jtobin: Fixed in updated changeset.

tapchannel/aux_leaf_signer.go
```diff
@@ -649,6 +649,10 @@ func (v *schnorrSigValidator) ValidateWitnesses(newAsset *asset.Asset,
 
 	for idx := range newAsset.PrevWitnesses {
 		witness := newAsset.PrevWitnesses[idx]
+		if witness.PrevID == nil {
```
darioAnongba, line 652: non-necessary defense in-depth.

vm/vm.go
```diff
@@ -210,6 +210,10 @@ func ValidateWitnesses(newAsset *asset.Asset,
 
 	for i, witness := range vm.newAsset.PrevWitnesses {
 		witness := witness
+		if witness.PrevID == nil {
```
darioAnongba, line 213: non-necessary defense in-depth.

vm/vm.go
```diff
@@ -447,6 +459,11 @@ func (vm *Engine) validateStateTransition() error {
 	for idx := range vm.newAsset.PrevWitnesses {
 		witness := vm.newAsset.PrevWitnesses[idx]
 
+		if witness.PrevID == nil {
```
darioAnongba, line 462: non-necessary defense in-depth.

vm/vm.go
```diff
@@ -257,9 +261,17 @@ func (vm *Engine) validateSplit(splitAsset *commitment.SplitAsset) error {
 	// prevID of zero, as the inherit the prev ID from the root asset.
 	//
 	// TODO(roasbeef): revisit post multi input
+	if len(vm.newAsset.PrevWitnesses) == 0 {
```
darioAnongba, line 264: 👍

vm/vm.go
```diff
@@ -257,9 +261,17 @@ func (vm *Engine) validateSplit(splitAsset *commitment.SplitAsset) error {
...
+		return fmt.Errorf("%w: prev witness zero", ErrNoInputs)
+	}
 	rootWitness := vm.newAsset.PrevWitnesses[0]
 	splitWitness := splitAsset.PrevWitnesses[0]
 
+	if rootWitness.PrevID == nil {
```
darioAnongba, line 270: 👍

proof/tx.go
```diff
@@ -358,6 +358,12 @@ func UnmarshalTxProof(
 			"missing")
 	}
 
+	if len(mp.SiblingHashes) > MerkleProofMaxNodes {
```
darioAnongba, line 361: Already enforced at line 173? low-value IMO.

darioAnongba (commented):
I generally tend to dislike defense/guard in-depth very much. Such checks should always happen once at the boundary and every internal function should work by assertion/invariant. Validate once, assume valid everywhere after. You otherwise enter into validation nightmare.

So I did some checks and some guards are valid, others are redundant. Maybe worth a tech-debt real fix to add this boundary validation and then remove all dead-guards afterwards.

jtobin:
> I generally tend to dislike defense/guard in-depth very much. Such checks should always happen once at the boundary and every internal function should work by assertion/invariant. Validate once, assume valid everywhere after. You otherwise enter into validation nightmare.

I think that's particularly true when runtime validation poses a runtime concern, e.g. you're doing super high-perf stuff and can't afford the extra branches on internal calls. But otherwise (*looks solemnly into the distance as a gentle breeze rustles his hair*) as a Haskell programmer, I think it's worth littering nil checks literally everywhere that it's valid to do so. *I* didn't introduce nil into this language.. 😄

The verbosity doesn't bother me (Claude writes it now anyway, amirite), and the local value-level checks do aid in refactoring, or mistakes made from mis-evaluating code. I vote let us write *more* nil checks, comrade!

tapchannel/aux_leaf_signer.go
jtobin: I vote keep, per my longer screed! (Ditto for the other two.)

proof/tx.go
jtobin: I think this one isn't actually enforced at L173? Different code path?

darioAnongba (approved):

PR #2169, opened 2026-06-24
Backward compatibility testing (replaces #2031)

GeorgeTsagk:
## Description

Almost identical to #2031: we now have an explicit low-blast-radius backwards compat test case.

So we still run `Fixture decode tests` on every PR plus the single backwards-compatibility itest case.

I don't think we should keep any broad case-duplication in scope.


rfqmsg/backward_compat_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+// readHexFixture reads a hex-encoded fixture file from the testdata/compat
+// directory and returns the decoded bytes.
+func readHexFixture(t *testing.T, version, name string) []byte {
+	t.Helper()
+
+	path := filepath.Join("testdata", compatFixtureDir, version, name)
+	hexBytes, err := os.ReadFile(path)
+	if os.IsNotExist(err) {
+		t.Skipf("fixture %s not found (run make gen-compat-fixtures "+
```
jtobin: I think this should fail, rather than skipping?

tapchannelmsg/backward_compat_test.go
```diff
@@ -0,0 +1,117 @@
...
+// testCompatFundingBlob tests that a historical funding blob can be decoded.
+func testCompatFundingBlob(t *testing.T, version string) {
+	t.Helper()
+
+	path := filepath.Join(
+		"testdata", "compat", version, "funding-blob.hexdump",
+	)
+	data, err := os.ReadFile(path)
+	if os.IsNotExist(err) {
+		t.Skipf("fixture %s not found", path)
```
jtobin: I think this should probably fail rather than skip, too.

Makefile
```diff
@@ -445,6 +454,11 @@ gen-itest-test-vectors:
 
 gen-test-vectors: gen-deterministic-test-vectors gen-itest-test-vectors
 
+gen-compat-fixtures:
+	@$(call print, "Generating backward compatibility fixtures.")
+	make unit gen-test-vectors=true pkg=tapchannelmsg case=^TestGenerateCompatFixtures$
```
jtobin: Think this line is stale; tapchannelmsg doesn't seem to define TestGenerateCompatFixtures.

.github/workflows/compat.yaml
```diff
@@ -0,0 +1,51 @@
+name: Backward Compatibility Tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
```
jtobin: 'version' looks to be unused; I think the job always runs the hard-coded list in itest/custom_channels/compat_test.go.

itest/integrated_harness.go
```diff
@@ -100,6 +100,53 @@ func (h *IntegratedNetworkHarness) NewNode(name string,
...
+
+	h.activeNodes[name] = n
+
+	return n
+}
+
+// UpgradeNode stops the given node and restarts it using the harness's
+// default (current) binary, preserving its data directories. This simulates
+// a node operator upgrading their software mid-operation.
+func (h *IntegratedNetworkHarness) UpgradeNode(
```
jtobin: This one looks like a leftover from #2031, unused here. Worth dropping?

.github/workflows/compat.yaml
```diff
@@ -0,0 +1,51 @@
...
+        description: 'Historical version to test against (e.g. v0.8.0). Leave empty to test all configured versions.'
+        required: false
+        type: string
+  push:
+    branches:
+      - 'release/**'
```
jtobin, line 10: We don't use this scheme at present. No reason we couldn't (the current 0.8 branch is just called 'v0.8', but could be release/v0.8), but this won't run yet.

Might actually be better to run this on 'v*' tags, as the release workflow does? I.e.:

```
on:
  push:
    tags:
      - 'v*'
```

That'd run for rc's (most important) and releases proper, which feels about right. Or maybe we could just manually trigger it as needed? Dunno, wdyt?

jtobin (commented):
Looks good, just a few slightly-above-nit-grade cleanup points.

jtobin (approved):
LGTM. :+1: :+1:

PR #2173, opened 2026-06-24
rfqmsg: bind incoming Accept to original Request peer

jtobin:
Previously, NewIncomingAcceptFromWire associated an Accept with a session purely by the session ID carried in the message, without checking that the wire-level sender was the same peer the original Request had been sent to.

The fix is to cross-check wireMsg.Peer against the Request's peer recorded at session open, and return an error on mismatch, so an Accept is only ever associated with the session it actually belongs to.

### DIFF
--- a/rfqmsg/accept.go
+++ b/rfqmsg/accept.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/lightninglabs/taproot-assets/fn"
+	"github.com/lightningnetwork/lnd/routing/route"
 	"github.com/lightningnetwork/lnd/tlv"
 )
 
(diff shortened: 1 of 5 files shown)


darioAnongba (approved):
LGTM. Might be a good idea to apply the same fix to the Reject path in same PR?

rfqmsg/accept.go
```diff
@@ -276,6 +276,27 @@ func NewIncomingAcceptFromWire(wireMsg WireMessage,
 			"incoming accept message: %s", msgData.ID.Val.String())
 	}
 
+	// The session lookup only ever stores Request values, so this
+	// assertion is for accessing MsgPeer rather than a runtime
+	// type check.
+	originalReq, ok := request.(Request)
+	if !ok {
```
darioAnongba: this seems to be dead code as `request` will always be `BuyRequest` or `SellRequest`. Unreachable as-is.

rfqmsg/accept.go
```diff
@@ -276,6 +276,27 @@ func NewIncomingAcceptFromWire(wireMsg WireMessage,
...
+		return nil, fmt.Errorf("session lookup returned non-request "+
+			"message of type %T for accept ID %s", request,
+			msgData.ID.Val.String())
+	}
+
+	// The wire-level sender must match the peer the original
+	// Request was sent to, so that an Accept is only ever
+	// associated with the session it belongs to.
+	if originalReq.MsgPeer() != wireMsg.Peer {
```
darioAnongba: nit: We already have a switch below so we could get `MsgPeer()` directly in the switch and drop this branch?

PR #2178, opened 2026-06-25
itest: wait for channel policy gossip in ListInvoices test

jtobin:
Fixes a flake observed in [this CI job](https://github.com/lightninglabs/taproot-assets/actions/runs/28193535578/job/83515488479), and also patches up an errant 'go vet' warning that was present.

Opus's justification for the keysend deletion:

> The keysend was being used as a side-channel to induce policy gossip.  The trouble: it was racing the very gossip it was meant to force — when the keysend fired too early, it failed with INSUFFICIENT_BALANCE (no policy → no route → routing thinks it can't get there → reports as a balance problem). That's the flake you saw on CI.
> 
> The new wait replaces a "fire-and-pray" workaround with a direct read of the exact precondition AddInvoice's getInboundPolicy consults: "Alice's policy is visible in Bob's graph." Once that's true, hop-hint construction will succeed deterministically; we no longer need a payment to force anything.
> 
> What we lost: nothing functional. The keysend wasn't asserting anything, wasn't part of the ListInvoices/ListPayments behavior being tested, and was just creating the auto-generated "warm-up" invoice/payment we then had to compensate for in the count assertions (the 4 → 3 change).  Dropping it actually simplifies the test surface.

### DIFF
--- a/itest/custom_channels/helpers.go
+++ b/itest/custom_channels/helpers.go
@@ -846,6 +846,71 @@ func assertChannelOutboundPolicyKnown(t *testing.T, node *itest.IntegratedNode,
 	require.NoError(t, err)
 }
 
+// assertPeerPolicyKnown asserts that the peer's channel edge policy for the
+// given channel is visible in the viewer's graph. This is the condition that
+// AddInvoice's RFQ hop-hint construction (which runs on the invoice creator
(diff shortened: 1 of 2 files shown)


GeorgeTsagk (approved):
Lgtm

PR #2190, opened 2026-07-02
tapgarden: hardening phase one (extraction)

jtobin:
This is an initial chunk of the large-scale tapgarden hardening work originally contained in #2153. It handles the following bullet from the original PR description:

> **Purity in output-key derivation**. Previously it was possible for MintingOutputKey to return a cached value that didn't reflect the sibling argument the caller actually passed; now, because the function holds no state, its output is determined entirely by its inputs.

It contains the following changes (summary ctsy Opus):

**Surface reduction.** `tapgarden` had accumulated three concerns that didn't belong to it: node-side interfaces (chain bridge, key ring, wallet anchor, group fetcher), the re-org watcher, and the custodian. They move to new packages `tapnode` (with `tapnodemock` for tests), `tapreorg`, and `tapcustody`. Pure moves with no behavior change; what remains in `tapgarden` is more legibly about minting.

**MintingOutputKey purity.** `MintingOutputKey` read like a function of `(batch, sibling)` but was secretly stateful — it cached its first answer and returned that on subsequent calls regardless of the sibling argument. Only latent because in practice the call sites happened to agree; any restart or refactor that changed their order would have made the batch anchor to an address the tapscript witness couldn't prove against. The cache is dropped and the function is now actually pure.


tapcustody/mock_test.go
```diff
@@ -0,0 +1,82 @@
...
+// by custodian tests to drive asset discovery flows.
+type MockAssetSyncer struct {
+	Assets map[asset.ID]*asset.AssetGroup
+
+	FetchedAssets chan *asset.AssetGroup
+
+	FetchErrs bool
+}
+
+func NewMockAssetSyncer() *MockAssetSyncer {
```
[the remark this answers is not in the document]
jtobin: (wontfix) I believe we didn't add this, so not addressing here so as to avoid rebase churn for future PR's in the series.

tapcustody/mock_test.go
```diff
@@ -0,0 +1,82 @@
+package tapcustody
```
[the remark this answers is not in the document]
jtobin: Fixed in updated changeset.

tapgarden/batch.go
```diff
@@ -168,15 +154,22 @@ func (m *MintingBatch) validateGroupAnchor(s *Seedling) error {
...
+// committed to (the Taproot Asset commitment combined with the
+// optional sibling).
+//
+// This function is pure in (m.BatchKey, m.RootAssetCommitment,
+// sibling): every call with the same arguments returns the same
+// result, and every call with a different sibling returns a
+// different result. There is no memoization; the on-curve work is
+// trivial and a cache would re-introduce the §IV bug shape where
+// the function silently ignored its sibling argument after the
+// first call.
```
GeorgeTsagk: > §IV

is this an external reference?
suggest
```golang
  // trivial, and a cache previously let the function silently ignore
  // its sibling argument on every call after the first, which could
  // anchor a batch to the wrong output key on a restart or RBF retry.
```

rpcserver/rpcserver.go
```diff
@@ -46,9 +46,11 @@ import (
 	"github.com/lightninglabs/taproot-assets/rpcutils"
 	"github.com/lightninglabs/taproot-assets/tapchannel"
 	"github.com/lightninglabs/taproot-assets/tapconfig"
+	"github.com/lightninglabs/taproot-assets/tapcustody"
```
GeorgeTsagk, line 49: commit message nit:

GenHeaderVerifier is described as a pure move but its body was lightly inlined (err := ...; return err → return ...) and it actually moved in the tapcustody commit, not where the description's emphasis implies.

GeorgeTsagk (approved):
lgtm, some optional nits

tapgarden/batch.go
jtobin: Yeah, there were a ton of these things that Opus added from my local planning doc. I thought I'd gotten rid of them all, but..

jtobin: Fixed in updated changeset.

rpcserver/rpcserver.go
jtobin: Fixed in updated changeset.

PR #2193, opened 2026-07-04
universe+tapdb: decode each ingested proof exactly once

jtobin:
(N.b., I started this as a performance enhancement attempt, but it wound up being a simplification/hardening piece that seemed worth pulling in. Fable's summary follows.)

Previously, every proof accepted into the universe was decoded from its raw bytes multiple times on its way to disk: once at the archive boundary for verification, once again inside the database write transaction, and once more to extract the block height, with the batch path repeating the same pattern for every leaf.

The fix is to decode once and share the result: universe.Leaf gains a lazily-memoized DecodedProof accessor whose cache can only be populated by decoding the leaf's own RawProof, so the decoded and raw forms cannot diverge by construction. The archive decodes at entry and every layer below reuses the memo, which also makes the batch path's proof side-map and all sparse block-height decodes redundant; both are removed. A rapid property test pins the accessor's agreement with the codec.

This is primarily a simplification and lock-hygiene change — measured end-to-end ingest throughput is unchanged, as transaction cost is dominated elsewhere.

### DIFF
--- a/tapdb/multiverse.go
+++ b/tapdb/multiverse.go
@@ -17,6 +17,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/proof"
 	"github.com/lightninglabs/taproot-assets/tapdb/sqlc"
 	"github.com/lightninglabs/taproot-assets/universe"
+	lfn "github.com/lightningnetwork/lnd/fn/v2"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


litbot-9000:

darioAnongba (approved):
LGTM

PR #2194, opened 2026-07-04
tapdb: parallelize universe proof ingest

jtobin:
(TLDR, disentangle the universe and multiverse writers. Postgres go brrrrrrrr. Fable's summary follows.)

Previously, every proof insert updated two trees inside one transaction: its own universe's tree, and the shared multiverse tree that summarizes all universe roots. Because the shared tree's root was rewritten by every insert, any two concurrent inserts — even into completely unrelated universes — collided on the same rows. Under Postgres's serializable isolation, a collision means abort-and-retry with backoff, so concurrent ingest was effectively single-writer, and in practice slower than serial: measured at 8 workers, 21.8 leaves/s concurrent vs 57.0 serial.

The fix is to decouple the two updates. An insert's transaction now touches only its own universe's state; the shared tree is maintained by a single coalescing writer that collects updates from all inserts and applies them in batches, one transaction per batch. Since only the latest root per universe matters, concurrent updates merge rather than queue, and callers still receive the post-flush multiverse root and inclusion proof, so RPC semantics are unchanged.

The multiverse tree is purely derived data, so a crash between universe commit and flush is healed by startup reconciliation — and its single-writer discipline lets the flush run at read committed on Postgres, taking it out of serialization-conflict detection entirely.

Result: 65.7 leaves/s concurrent vs 61.8 serial on the same benchmark, a 3x improvement with the inversion gone. Hooking #2188's batched-descent InsertMany into the flush should collapse each batch into a single descent per tree, which we expect to close much of the gap between the current ~65 leaves/s and the ~700 leaves/s ceiling measured for the universe transactions alone.

### DIFF
--- a/docs/release-notes/release-notes-0.8.2.md
+++ b/docs/release-notes/release-notes-0.8.2.md
@@ -45,11 +45,14 @@
 
 ## Performance Improvements
 
+- [PR#2183](https://github.com/lightninglabs/taproot-assets/pull/2183)
+  dramatically improves the performance of MS-SMT proof verification.
+
 - [PR#2184](https://github.com/lightninglabs/taproot-assets/pull/2184)
   dramatically improves the performance of universe federation proof push.
 
-- [PR#2183](https://github.com/lightninglabs/taproot-assets/pull/2183)
-  dramatically improves the performance of MS-SMT proof verification.
+- [PR#2194](https://github.com/lightninglabs/taproot-assets/pull/2194)
+  significantly improves concurrent universe proof ingest on Postgres.
 
 - [PR#2188](https://github.com/lightninglabs/taproot-assets/pull/2188)
   dramatically improves the performance of batched MS-SMT insertions.
--- a/tapcfg/server.go
+++ b/tapcfg/server.go
@@ -196,6 +196,15 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 		return nil, fmt.Errorf("create multiverse store: %w", err)
 	}
 
+	// The multiverse trees are derived from the universe roots; repair any
+	// entries that diverged, for example because the daemon stopped
+	// between a proof insert committing and its multiverse update being
+	// written.
+	err = multiverse.ReconcileMultiverse(context.Background())
+	if err != nil {
+		return nil, fmt.Errorf("reconcile multiverse: %w", err)
+	}
+
 	uniStatsDB := tapdb.NewTransactionExecutor(
 		db, func(tx *sql.Tx) tapdb.UniverseStatsStore {
 			return db.WithTx(tx)
--- a/tapcfg/server.go
+++ b/tapcfg/server.go
@@ -45,8 +45,9 @@ import (
 //
 // NOTE: The RPCConfig and SignalInterceptor fields must be set by the caller
 // after generating the server config.
-func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
-	lndServices *lndclient.LndServices, enableChannelFeatures bool,
+func genServerConfig(ctx context.Context, cfg *Config,
+	cfgLogger btclog.Logger, lndServices *lndclient.LndServices,
+	enableChannelFeatures bool,
 	mainErrChan chan<- error) (*tapconfig.Config, error) {
 
 	var (
@@ -194,6 +195,16 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("create multiverse store: %w", err)
+	}
+
+	// The multiverse trees are derived from the universe roots; repair any
+	// entries that diverged, for example because the daemon stopped
+	// between a proof insert committing and its multiverse update being
+	// written. The context is shutdown-derived, so an operator can
+	// interrupt a long-running reconcile.
+	err = multiverse.ReconcileMultiverse(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("reconcile multiverse: %w", err)
 	}
 
 	uniStatsDB := tapdb.NewTransactionExecutor(
@@ -946,9 +957,23 @@ func CreateServerFromConfig(cfg *Config, cfgLogger btclog.Logger,
 
 	cfgLogger.Infof("lnd connection initialized")
 
+	// Derive a context that ends when the interceptor signals
+	// shutdown, so long-running startup work — the multiverse
+	// reconcile in particular — can be interrupted.
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func() {
+		select {
+		case <-shutdownInterceptor.ShutdownChannel():
+			cancel()
+
+		case <-ctx.Done():
+		}
+	}()
+
 	serverCfg, err := genServerConfig(
-		cfg, cfgLogger, &lndConn.LndServices, enableChannelFeatures,
-		mainErrChan,
+		ctx, cfg, cfgLogger, &lndConn.LndServices,
+		enableChannelFeatures, mainErrChan,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("unable to generate server config: %w",
@@ -1052,12 +1077,15 @@ func mboxOutpointChecker(
 }
 
 // ConfigureSubServer updates a Taproot Asset server with the given CLI config.
-func ConfigureSubServer(srv *tap.Server, cfg *Config, cfgLogger btclog.Logger,
-	lndServices *lndclient.LndServices, litdIntegrated bool,
-	mainErrChan chan<- error) error {
+// The context bounds the startup work done here, the multiverse
+// reconcile in particular, and should be derived from the daemon's
+// shutdown signal.
+func ConfigureSubServer(ctx context.Context, srv *tap.Server, cfg *Config,
+	cfgLogger btclog.Logger, lndServices *lndclient.LndServices,
+	litdIntegrated bool, mainErrChan chan<- error) error {
 
 	serverCfg, err := genServerConfig(
-		cfg, cfgLogger, lndServices, litdIntegrated, mainErrChan,
+		ctx, cfg, cfgLogger, lndServices, litdIntegrated, mainErrChan,
 	)
 	if err != nil {
 		return fmt.Errorf("unable to generate server config: %w", err)
--- a/tapdb/multiverse.go
+++ b/tapdb/multiverse.go
@@ -701,7 +701,9 @@ func (b *MultiverseStore) FetchProofLeaf(ctx context.Context,
 			return proofs, nil
 		}
 
-		_, err = b.rootCoalescer.updateRoot(ctx, id)
+		err = b.rootCoalescer.updateRoots(
+			ctx, []universe.Identifier{id},
+		)
 		if err != nil {
 			return nil, fmt.Errorf("failed multiverse heal "+
 				"for %v: %w", id.String(), err)
@@ -1001,6 +1003,12 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 
 // UpsertProofLeafBatch upserts a proof leaf batch within the multiverse tree
 // and the universe tree that corresponds to the given key(s).
+//
+// The universe leaves commit first, in their own transaction, and the
+// shared multiverse tree is updated afterwards. An error return may
+// therefore mean the leaves are durably stored while the multiverse
+// update failed; in that case each universe's multiverse entry is
+// healed by its next successful update.
 func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 	items []*universe.Item) error {
 
@@ -1009,13 +1017,10 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 		uniProofs    []*universe.Proof
 		rootStatuses []universeRootStatus
 	)
-	// Track the final universe root per universe, so the shared
-	// multiverse tree is updated once per universe with its latest root,
-	// rather than once per item.
-	type multiverseUpdate struct {
-		id   universe.Identifier
-		root mssmt.Node
-	}
+	// Track the universes the batch touches, in first-touch order, so
+	// the shared multiverse tree is refreshed once per universe rather
+	// than once per item.
+	var dirtyUniverses []universe.Identifier
 
 	dbErr := b.db.ExecTx(
 		ctx, &writeTx, func(store BaseMultiverseStore) error {
@@ -1024,11 +1029,10 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 				[]universeRootStatus, len(items),
 			)
 
-			finalRoots := make(
-				map[universeIDKey]*multiverseUpdate,
-				len(items),
+			dirtyUniverses = make(
+				[]universe.Identifier, 0, len(items),
 			)
-			var updateOrder []universeIDKey
+			seen := make(map[universeIDKey]struct{}, len(items))
 
 			for idx := range items {
 				item := items[idx]
@@ -1054,25 +1058,11 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 				rootStatuses[idx] = status
 
 				key := item.ID.String()
-				if _, ok := finalRoots[key]; !ok {
-					updateOrder = append(updateOrder, key)
-				}
-				finalRoots[key] = &multiverseUpdate{
-					id:   item.ID,
-					root: uniProof.UniverseRoot,
-				}
-			}
-
-			// Next, we'll insert each universe's final root into
-			// the main multiverse tree.
-			for _, key := range updateOrder {
-				update := finalRoots[key]
-				err := upsertMultiverseLeafEntry(
-					ctx, store, update.id, update.root,
-				)
-				if err != nil {
-					return fmt.Errorf("failed multiverse "+
-						"upsert for %v: %w", key, err)
+				if _, ok := seen[key]; !ok {
+					seen[key] = struct{}{}
+					dirtyUniverses = append(
+						dirtyUniverses, item.ID,
+					)
 				}
 			}
 
@@ -1083,6 +1073,11 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 		return dbErr
 	}
 
+	// The universe leaves are now durably committed, so run the
+	// bookkeeping tied to that commit before the multiverse update: a
+	// failed or slow multiverse flush must not leave caches stale or
+	// keep the custodian from learning about stored transfer proofs.
+	//
 	// If any of the inserted leaves created a new universe, the
 	// composition of paginated root queries changed and the page cache as
 	// a whole is stale. Pure updates only change the value of already
@@ -1105,14 +1100,16 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 			)
 		}
 
+		// The syncer cache is deliberately not updated here: it
+		// installs values rather than evicting them, and the
+		// unordered post-commit sections of concurrent inserts
+		// could install roots out of commit order. It is fed from
+		// the coalescer's flush callback instead.
 		newRoots[idx] = universe.Root{
 			ID:        items[idx].ID,
 			AssetName: items[idx].Leaf.Asset.Tag,
 			Node:      uniProofs[idx].UniverseRoot,
 		}
-
-		// Update the syncer cache with the new root node.
-		b.syncerCache.addOrReplace(newRoots[idx])
 	}
 
 	// Evict the cached pages that contain any of the updated roots. If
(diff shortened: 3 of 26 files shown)


tapdb/multiverse_coalescer.go
```diff
@@ -0,0 +1,346 @@
...
+		for _, key := range c.order {
+			batch = append(batch, c.pending[key])
+		}
+		c.pending = make(map[universeIDKey]*pendingRootUpdate)
+		c.order = nil
+		c.mu.Unlock()
+
+		c.flushBatch(batch)
+	}
+}
```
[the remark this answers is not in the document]
jtobin: Fixed in updated changeset.

darioAnongba (changes requested):
Some findings by an agent on this PR. 
C-1 should be fixed
H-1 as well
C-2 is a bit special because it implies mutual exclusion between processes, meaning that it found a regression for cases where we run multiple tapd's on the same postgres DB. Still a regression but not sure if realistic but could be when a user is deploying in a kube cluster with high availability, still worth checking. 

[lightninglabs_taproot-assets_PR_2194_review.md](https://github.com/user-attachments/files/30700139/lightninglabs_taproot-assets_PR_2194_review.md)

litbot-9000:
@georgetsagk: review reminder

universe/federation.go
```diff
@@ -546,17 +547,30 @@ func (f *FederationEnvoy) handlePushRequest(pushReq *FederationPushReq) error {
...
 		ctx, pushReq.ID, pushReq.Key, pushReq.Leaf,
 	)
-	if err != nil {
+	switch {
+	// The proof leaf is durably stored; only its entry in the shared
+	// multiverse trees is still outstanding, and the archive repairs
+	// that in the background. The caller cannot receive a composing
+	// receipt, but the proof must not be stranded local-only, so the
+	// federation push below proceeds.
+	case errors.Is(err, ErrMultiversePending):
```
litbot-9000, line 556: This branch is the right call, but the same treatment is missing on every batch path, and `handleBatchPushRequest` 75 lines below is the one that hurts.

`universe/interface.go:590` explicitly documents that `UpsertProofLeafBatch` returns `ErrMultiversePending` with the same meaning — "the leaves are durably stored while their multiverse updates are still outstanding" — and `tapdb/multiverse.go:1252` really does produce it, via `updateRoots` → the waiter error built at `multiverse_coalescer.go:683-684`. But `handleBatchPushRequest` (`:633`) still has a bare `if err != nil { pushReq.err <- err; return err }`.

Concretely: a transient DB failure during one flush round while a batch push is in flight. The universe leaves commit. `updateRoots` returns an `ErrMultiversePending`-wrapped error. `handleBatchPushRequest` returns *before* `tryFetchServers`, `filterProofSyncPending` and `pushProofToFederation` — so the proofs are never pushed to any federation peer, and because the `LogProofSync` retry rows are written inside `pushProofToFederation`, no retry is scheduled either. Seconds later the coalescer's retry succeeds and the multiverse is perfectly consistent, but that batch is now permanently stranded local-only. That is precisely the stranding this `case` was written to prevent.

Grepping non-test code, `ErrMultiversePending` is handled in exactly two places — here and `universe/archive.go:354`. The unhandled receivers are `universe/federation.go:633`, `universe/archive.go:569` and `:582`, `universe/syncer.go:659`, and `tapgarden/caretaker.go:1317`. The archive one has the same shape as the fix you made for the single-leaf case: returning early at `:570` skips `LogNewProofEvents` at `:593`, which is exactly the call you restructured `UpsertProofLeaf` to keep running in the pending case.

Worth applying the same `switch` to `handleBatchPushRequest` at minimum, and deciding explicitly what the syncer and caretaker should do.

tapdb/multiverse.go
```diff
@@ -863,11 +1060,59 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
...
+	// compose with the universe proof assembled above. Rebuild the
+	// whole receipt from one consistent snapshot instead, bypassing
+	// the proof cache: a concurrent reader may have repopulated it,
+	// after the eviction above, from a snapshot predating this call's
+	// own insert, and that receipt would not reflect the leaf this
+	// call committed.
+	if !mssmt.IsEqualNode(update.universeRoot, uniProof.UniverseRoot) {
+		proofs, err := b.fetchProofLeaf(ctx, id, key, true)
+		if err != nil {
+			return nil, fmt.Errorf("failed superseded proof "+
```
litbot-9000: By the time control reaches this rebuild, two things are already durable: the universe leaf (committed at `:1026`) and the multiverse update (`updateRoot` returned a nil error at `:1070`). But neither this error nor the `len(proofs) != 1` error just below wraps `universe.ErrMultiversePending`, so `UpsertProofLeaf` can return a plain, untagged error describing a proof that is fully committed.

That matters because it defeats the contract on the consumer side. `universe/federation.go:556` matches only `errors.Is(err, ErrMultiversePending)`; anything else falls through to the generic `case err != nil` at `:562`, which sends to `pushReq.err` and returns before the federation push — stranding the proof local-only with no retry row, the same outcome that branch exists to avoid. `universe/archive.go:354` has the same problem and additionally skips `LogNewProofEvent`.

Three reachable errors from `fetchProofLeaf` here:

- `ErrMultiverseInconsistent` after exhausting `maxFetchAttempts = 5`. Your own comment at `:769-773` and `TestFetchProofLeafUnderConcurrentInserts` both concede this happens under sustained same-universe ingest — which is the workload this PR exists to speed up.
- Any transient DB error, returned verbatim at `:788-790`.
- `len(proofs) == 0` if a concurrent `DeleteProofLeaf` removed the leaf in the gap.

I tried to refute this by looking for an unreachable-branch argument, but it goes the other way: this block is entered *only* when the flush derived a different root than this call committed, i.e. only when a concurrent same-universe insert already happened — so we are by construction in the contended regime where the first case is most likely. I also checked whether an outer wrapper might add the tag; the only sites that wrap with `ErrMultiversePending` are `multiverse_coalescer.go:504` and `:684`, neither on this path.

Simplest fix is probably to tag every failure after the `:1026` commit with `ErrMultiversePending`, so the invariant is "any error from this function after the leaf commits carries the tag" rather than "some of them do".

docs/release-notes/release-notes-0.8.2.md
```diff
@@ -45,11 +45,14 @@
...
+- [PR#2183](https://github.com/lightninglabs/taproot-assets/pull/2183)
+  dramatically improves the performance of MS-SMT proof verification.
+
 - [PR#2184](https://github.com/lightninglabs/taproot-assets/pull/2184)
   dramatically improves the performance of universe federation proof push.
 
-- [PR#2183](https://github.com/lightninglabs/taproot-assets/pull/2183)
-  dramatically improves the performance of MS-SMT proof verification.
+- [PR#2194](https://github.com/lightninglabs/taproot-assets/pull/2194)
+  significantly improves concurrent universe proof ingest on Postgres.
```
litbot-9000, line 55: This is filed purely as a performance improvement, but the PR also changes two things an operator can observe, neither of which is mentioned anywhere in the notes:

1. **`UpsertProofLeaf`/`UpsertProofLeafBatch` can now return an error while the proof is durably stored.** That is a real API contract change — you documented it on `MultiverseArchive` (`universe/interface.go:571-593`), but anyone integrating against tapd who reads only the release notes will keep treating an error as "nothing was written".
2. **Startup now runs an unconditional multiverse reconciliation pass** that scans every universe root before the node serves traffic. On a node with many universes that is a new, visible startup cost, and per the inline comment on `tapcfg/server.go:206` a failure there prevents boot.

Dario's H-8 also asked for a config kill-switch so the new path can be disabled without a binary downgrade. I'd weigh that lower than the notes — the design holds up well and a half-enabled coalescer has its own risks — but the two items above seem worth a Breaking Changes or Functional Updates entry regardless.

tapcfg/server.go
```diff
@@ -196,6 +197,16 @@ func genServerConfig(cfg *Config, cfgLogger btclog.Logger,
 		return nil, fmt.Errorf("create multiverse store: %w", err)
 	}
 
+	// The multiverse trees are derived from the universe roots; repair any
+	// entries that diverged, for example because the daemon stopped
+	// between a proof insert committing and its multiverse update being
+	// written. The context is shutdown-derived, so an operator can
+	// interrupt a long-running reconcile.
+	err = multiverse.ReconcileMultiverse(ctx)
+	if err != nil {
```
litbot-9000, line 206: Reconcile failure is fatal, so any error here means the daemon does not start. The deterministic case Dario found (C-1) is genuinely fixed, and the mechanics around it are much better now — shutdown-derived context, keyset pagination, bounded memory, chunked repairs that commit independently — so what's left is narrower: a transient DB error, or a single chunk overrunning `defaultFlushTimeout`, turns a node that would otherwise run fine into one that won't boot.

The asymmetry is that the condition being repaired is itself non-fatal. A stale multiverse root degrades federation sync; it does not corrupt anything, and `FetchProofLeaf` self-heals the universes that are actually read. Refusing to start seems a heavier response than the failure warrants, especially since the chunked repair loop already means a restart resumes with strictly less divergence.

A loud `log.Errorf` plus a background retry would keep the safety net while removing the boot dependency. If you'd rather keep it fatal — defensible, since a wrong root is what federation peers compare against — it would help to say so in the release note, so an operator hitting it knows it's intended and knows a retry may clear it.

tapdb/multiverse_coalescer.go
```diff
@@ -0,0 +1,702 @@
...
+		if err == nil {
+			backoff = c.initialRetryBackoff
+			continue
+		}
+
+		// A panic is presumed to be a bug rather than a transient
+		// database failure; retrying would re-trigger it at every
+		// backoff. The round's universes stay diverged until their
+		// next update or startup reconciliation.
+		if errors.Is(err, errFlushPanicked) {
```
litbot-9000, line 396: Minor, and the reasoning in the comment is sound — a panic is a bug, and retrying would just re-trigger it every backoff. Two things worth a second look though.

The batch was already removed from `pending` at `:375` and this path skips `requeue`, so the round's universes stay diverged with no in-process path back to consistency. The comment says "until their next update or startup reconciliation", which is accurate, but it's worth noting the node keeps running and serving a stale multiverse root to federation peers for those universes in the meantime. `onFlushError` invalidates the syncer cache and `FetchProofLeaf` heals any universe that is actually read, so the blast radius is smaller than it first looks — universes that are written or read recover, and only untouched ones stay stale until restart. Probably fine, but the divergence is silent, and a `log.Criticalf` that named the affected universes would make it diagnosable.

Second, `continue` here skips the backoff *and* leaves `backoff` at its current value without resetting it. With continuous ingest and a deterministically panicking flush, that's a hot loop of panicking transactions with no delay between them — each one logging a full stack trace. A `time.After(backoff)` on this path too would bound the log volume and DB churn, even though the round itself is correctly not retried.

litbot-9000 (commented):
Reviewed commit-by-commit at `4881edb`. **The three things Dario named as must-fix — C-1, C-2 and H-1 — are genuinely fixed, and I verified each independently rather than taking the replies at face value.** The design here is sound: the load-bearing property is that a flush *re-derives* each universe's root inside its own transaction instead of applying a value submitted by the caller, which makes the stale-root-overwrite class of bug structurally impossible rather than merely unlikely. I traced ordering, deletion interaction and the cache-fill race looking for that bug and could not construct one. What I did find is narrower and lives on the error path: the new "stored but not yet flushed" contract is documented on the interface and honoured by the two single-leaf consumers, but not by the batch consumers, and not by every post-commit failure in `UpsertProofLeaf` itself. Those are fixable without touching the design, which is why this is a `COMMENT` and not a block.

## Dario's comments

His three named items:

- **C-1 (tapd unbootable with supply-commitment sub-trees) — RESOLVED.** The hardcoded `proofTypes := []{Issuance, Transfer}` prefetch map is gone entirely. `rootDiverged` (`tapdb/multiverse_reconcile.go:158-174`) now calls `namespaceForProof(id.ProofType)` and returns "not diverged" when it errors, so `mint_supply`/`burn`/`ignore` roots are skipped rather than flagged and then mis-resolved via `id.String()`. `namespaceForProof` (`tapdb/multiverse.go:271-284`) returns a namespace only for issuance/transfer. The crash path is closed at its root.
- **C-2 (READ COMMITTED guarded only by a process-local mutex) — RESOLVED.** The isolation downgrade no longer exists. `multiverseFlushTx` (`tapdb/multiverse_coalescer.go:70-75`) implements only `ReadOnly()`; there is no isolation override method. `grep` for `TxIsolationOverrider`/`LevelReadCommitted` across the tree returns zero hits, and `git diff main...HEAD -- tapdb/interfaces.go` is empty — the file is untouched by this PR. Flushes run at serializable, so mutual exclusion is back in the database and covers a second `tapd` on the same DB. The comment at `:62-69` now says exactly this. This also resolves **H-9**, which was a consequence of the same override.
- **H-1 (writeMu held across syncer-cache callbacks → global ingest stall) — RESOLVED,** though by fixing the other end of the chain than he proposed. `writeMu` is still deliberately held across the flush callbacks (`:540-541`), and the commit message for `c15930ba` explains why: releasing it first would let a flush reinstall a root for a universe whose deletion committed after the flush. Instead the callbacks were made cheap — `fillSyncerCache` no longer holds the cache lock across its paginated scan. That breaks the stall chain at the point that actually caused it.

The rest of his report, since much of it drove commits 7-11:

- **H-2 (an error no longer implies nothing was written) — PARTIALLY RESOLVED.** The typed `ErrMultiversePending` exists, the `MultiverseArchive` interface documents the weakened contract, and both single-leaf consumers were fixed. The batch consumers were not, and not every post-commit failure carries the tag. See my two inline comments — this is the main thing left.
- **H-3 (`ErrMultiverseInconsistent` from public reads) — RESOLVED.** Exponential backoff, context checked between attempts, typed retryable error, and `QueryProof` maps it to `codes.Unavailable`. Critically, exhaustion returns an error rather than non-composing proofs.
- **H-4 (all-or-nothing 2048 rounds, positive-feedback cliff) — RESOLVED.** `InsertMany` is now wired in (`tapdb/universe.go:936-944`), failed rounds requeue to the front and retry with capped exponential backoff, and reconcile repairs in independently-committing 512-universe chunks so a failure preserves prior progress.
- **H-5 (flush leader pinned, ignores its own context) — RESOLVED.** The flusher is a dedicated background goroutine (`runFlusher`, `:359`); callers wait only on their own result channel with a `ctx.Done()` arm and are never conscripted into draining the queue.
- **H-6 (depleted syncer cache served as authoritative) — RESOLVED.** `DeleteProofLeaf` now invalidates wholesale instead of `remove()`-ing a possibly-surviving universe; `DeleteUniverse` keeps its targeted removal, correctly. `TestSyncerCacheDeleteProofLeaf` covers both branches and asserts the served root equals the actual DB root, not merely that the universe is present.
- **H-7 (reconcile can make tapd unbootable) — PARTIALLY RESOLVED.** The deterministic case is gone with C-1, and the mechanics are much better: shutdown-derived context, keyset pagination instead of `LIMIT/OFFSET`, bounded memory. Reconcile failure is still fatal — inline comment below.
- **H-8 (no flag, no kill-switch, thin release note) — NOT RESOLVED.** Inline comment below.
- **H-10 (public reads open write transactions) — OPEN BY DESIGN.** `FetchProofLeaf` still calls `updateRoots`, but only on detected inconsistency, and it is bounded. Reasonable; worth knowing it rules out read-replica deployments.
- **M-1 (Gemini's `c.flushing` panic deadlock) — RESOLVED.** `recover` is the first defer in `flushBatch`, and `runFlusher` clears `flushing` under `c.mu` on every return path.
- **M-3 (no `needsFill()` re-check → duplicate table scans) — RESOLVED.** `fillMu` plus the re-check at the top of `fillSyncerCache` means exactly one scan.
- **M-2 (unbounded waiter records)** — still uncapped, but bounded in practice by in-flight callers, since every waiter corresponds to a blocked call. Not worth changing.

The older `gemini-code-assist` finding (panic leaves `c.flushing` true forever) is fixed, and I have not restated it.

## Design review

Async coalescing plus startup reconciliation is a sound way to get this guarantee, and I want to separate that from the bugs below. Two properties carry it. First, the flush derives each universe's root inside the flush transaction rather than trusting a submitted value (`tapdb/multiverse_coalescer.go:573`, documented at `:244-249`), so submission order not tracking commit order is harmless — a retried round re-derives, and a stale snapshot can never overwrite a newer root. That single decision is what makes the ordering, retry and deletion-interaction questions all come out right; I checked each separately and they reduce to it. Second, the multiverse trees are *fully derived* from the universe roots, which is what makes reconciliation a real safety net rather than a hopeful one: it is unconditional on every boot (`tapcfg/server.go:205`, no marker or fast path), it scans every universe root rather than recently-touched ones, and orphaned multiverse leaves can't occur because both deletion paths remove tree and leaf in one transaction. So dropping pending updates on shutdown — which `stop()` does deliberately — is recoverable by construction, including after `kill -9`. Worth noting the caller-visible consistency is stronger than "eventually consistent": `updateRoot` blocks until the flush carrying its update commits, so read-your-writes still holds for anyone who gets a successful return. The asynchrony buys transaction *decoupling*, not caller decoupling.

## What I verified

- `make build` → exit 0 at `4881edb`.
- `go test ./mssmt/ -run TestInsertLastWriteWins -v -count=1` → PASS, `[rapid] OK, passed 100 tests (337ms)`. I read the property rather than trusting the name: it draws a 1-8 key pool against 1-32 inserts so key reuse is likely, compares full-sequence application against final-leaf-per-key, and additionally verifies inclusion proofs. That is last-write-wins as the production code relies on it.
- `go test ./tapdb/ -run 'TestMultiverseRootCoalescer|TestFetchProofLeaf|TestUpsertProofLeaf|TestReconcile|TestMultiverse|TestSyncerCache' -count=1` → `ok ... 38.582s`.
- The same selection under `-race` → `ok ... 1087.205s`, exit 0, no data races reported. One sequential race run, as this PR is the best race-detector candidate in the queue.
- Read every hunk of all 12 commits; traced four deletion/flush interleavings and four cache-fill/invalidate interleavings by hand looking for a lost or resurrected update, and found none.
- Dialect check: one new query (`UniverseRootsAfterID`), plain keyset pagination with the same JOIN set as the existing `UniverseRoots`, no dialect-specific constructs, no schema or migration change. `InsertMany` is fed a `map[[32]byte]*LeafNode`, so duplicate keys within a round are impossible by construction.
- `gh pr checks 2194` → all green, including `run unit tests (unit dbbackend=postgres pkg=tapdb)` and `run itests postgres (0-3)`.

**What I could not verify.** There is no container runtime in my environment, so nothing behind `ory/dockertest` or the `test_db_postgres` build tag ran. Everything above is SQLite. That is an awkward gap for this PR specifically, since the entire motivation is Postgres serializable-isolation contention: **I did not measure Postgres throughput and cannot confirm the 21.8 vs 57.0 leaves/s claim** — my read of the Postgres path is by inspection, and the Postgres verdict rests on CI. I also want to flag that CI being green is not evidence of crash-safety; no job kills a node mid-flush. My confidence in the durability story comes from reconciliation being unconditional and complete, not from CI. On the benchmark itself: `BenchmarkMultiverseFlushRound` is honest about what it measures — one flush round, with dirtying under `b.StopTimer()` — and it is the right shape for sizing `maxFlushBatchSize` against `defaultFlushTimeout`, which is its stated purpose. It is not a concurrent-vs-serial throughput benchmark and does not substantiate the headline number, which comes from the pre-existing `bench/fixture` harness outside this diff.

## Findings

Most severe first; details are inline on the code they concern.

1. `universe/federation.go:556` — batch consumers never handle `ErrMultiversePending`, so a batch push strands durably-stored proofs local-only.
2. `tapdb/multiverse.go:1087` — post-commit failures in the superseded-receipt rebuild aren't tagged `ErrMultiversePending`, so even the consumers that were taught the contract misclassify them.
3. `docs/release-notes/release-notes-0.8.2.md:55` — release note omits both operator-visible behaviour changes.
4. `tapcfg/server.go:206` — a transient DB error during reconcile is fatal at startup.
5. `tapdb/multiverse_coalescer.go:391` — a panicked round is dropped without requeue, leaving those universes diverged until restart.

universe/federation.go
jtobin: Fixed in updated changeset.

tapdb/multiverse.go
jtobin: Fixed in updated changeset.

docs/release-notes/release-notes-0.8.2.md
jtobin: wontfix

tapcfg/server.go
jtobin: wontfix. IIUC this is pretty overfitty, and failing on the weird case is probably correct anyway.

tapdb/multiverse_coalescer.go
jtobin: Fixed in updated changset.

GeorgeTsagk:
Removing request for review, as @darioAnongba is already on this.

darioAnongba (approved):
LGTM

PR #2202, opened 2026-07-10
universe: add cursor-based delta sync to federation

### LINKED ISSUE #290: universe: recursive tree based universe syncing
This is a brainstorm/tracking issue for a more advanced type of Universe sync. The version I'll likely implement in lightninglabs/tap#7 is a simple variant that scales (bandwidth wise) as `O(N*M)` where `N` is the number of assets (each has a distinct Universe root), and `M` is the final diff set that actually needs to be transmitted. There's another factor of hidden in `M`, which is the amount of keys sent, but the inclusion proofs are much larger than that. 

This "linear" method would look something like: 
  1. Fetch the known universe roots from the peer. 
  2. Fetch the set of keys for each universe root you care about, or doesn't match what you have stored on disk. 
  3. For each set of keys
     a. Perform a local set difference (items in the set that you don't know about) 
     b. Query remote party for the inclusion proofs of all those leaves. 
 4. Query again for the universe root to make sure it matches. 

In most cases, it's a noop after step 1, as things fully match.
    

Instead, we can do something that gives us `O(log(N)*M)` scaling. For each Universe, we can find the bisection point in a logarithmic amount of steps, then fetch the set of proofs `M` for them. 

This tree-based method looks something like:
  1. Fetch the set of known universe roots from the peer (note that for a Multiverse setting, this can run another instance of this protocol in the top-level Universe tree, then recursive down an do another instance). 
       a. In addition to the root, this now also returns the left+right sibling hash. 
  2. If the root matches, exit. 
       a. Otherwise, look at the left+right tree. If both of them don't match your local view, the recurse down into both left+right. 
       b. Otherwise, recurse down into only one of them (the one that doesn't match). 
  3. (recursive step, written linearly here for simplicity): 
       * At this point we've found a diff, we can either fetch the entire sub-tree (short circuit), or continue to recurse down. 
       * If recursing, and you find a sibling that matches (the root didn't match), only 

For both of these algorithms, we can also reduce the total bandwidth usage by sending batched proofs. Rather than a proof that reveals only a single leaf, we can create a sparse(r) proof tree, by combining the shared internal nodes of all the proofs.

jtobin:
Previously, federation sync discovered new asset leaves by linear-diffing remote and local state (for differing asset roots) to find what was missing. This was super inefficient, in that the discovery cost scaled with the total number of leaves, rather than with the amount of new ones added.

The fix used in this PR is to adopt a stateful cursor-based model for sync. Efficiency gains are substantial, especially for incremental syncs of assets with tons of leaves, since discovery cost now scales only with the amount of data unseen. The core mechanism is as follows:

1) each node keeps an append-only journal of its own leaf+proof insertions
2) each node keeps track of the most recent journal index they've seen from each other node in their federation (the "cursor")
3) a new SyncDelta RPC allows a node to ask another "what have you added since N?"

To sync, each node simply asks others "what did you add since I last checked?" with a fallback to the old enumeration sync in any pathological case (e.g. verification failure against another node's claimed root, or if the other node doesn't have a SyncDelta RPC at all). The responding node delivers entries from its log in pages, and the syncing node pages through results, advancing the cursor until it stops incrementing.

(This obviates #290, which was prototyped first, but was found to be much less efficient.)


rpcserver/rpcserver.go
```diff
@@ -8195,6 +8195,176 @@ func (r *RPCServer) SyncUniverse(ctx context.Context,
 	return r.marshalUniverseDiff(ctx, universeDiff)
 }
 
+// syncDeltaByteBudget caps the approximate payload size of a single
+// SyncDelta response page, keeping it comfortably under the default
+// 4 MiB gRPC message size limit. A variable rather than a constant so
+// tests can exercise the page-splitting path with small payloads.
+var syncDeltaByteBudget = 3 * 1024 * 1024
```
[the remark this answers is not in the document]
jtobin: Fixed in updated changeset.

bench/fixture/discovery.go
```diff
@@ -0,0 +1,166 @@
...
+// syncer's local enumeration is a DB scan, not a round trip.
+type countingDiffEngine struct {
+	inner   universe.DiffEngine
+	metrics *DiscoveryMetrics
+}
+
+var _ universe.DiffEngine = (*countingDiffEngine)(nil)
+
+func (c *countingDiffEngine) RootNode(ctx context.Context,
+	id universe.Identifier) (universe.Root, error) {
```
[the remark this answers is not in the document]
jtobin: Fixed in updated changeset.

litbot-9000:
@roasbeef: review reminder
@georgetsagk: review reminder

litbot-9000:
@roasbeef: review reminder
@georgetsagk: review reminder

tapdb/sqlc/queries/universe.sql
```diff
@@ -103,6 +103,25 @@ LIMIT @num_limit OFFSET @num_offset;
...
+       leaves.asset_genesis_id, nodes.value AS genesis_proof,
+       nodes.sum AS sum_amt, roots.asset_id AS root_asset_id,
+       roots.group_key AS root_group_key, roots.proof_type
+FROM universe_leaves AS leaves
+JOIN universe_roots AS roots
+    ON leaves.universe_root_id = roots.id
+JOIN mssmt_nodes AS nodes
+    ON leaves.leaf_node_key = nodes.key
+       AND leaves.leaf_node_namespace = nodes.namespace
+WHERE leaves.id > @since_seq
```
litbot-9000: **Blocker: the journal index is not commit-ordered on Postgres, so this query can hand out index N while N-1 is still uncommitted — and the cursor never comes back for it.**

The journal index here is `universe_leaves.id`, which is `INTEGER PRIMARY KEY` in the sqlite schema and therefore `BIGSERIAL PRIMARY KEY` on Postgres (tapdb/postgres.go:30). Sequence values in Postgres are allocated at INSERT time and are explicitly non-transactional: allocation order and commit order are independent. This query has no stability watermark — it takes everything `> since_seq` in id order — and `FetchLeavesSince` reports `maxSeq` as simply the last row it saw (tapdb/multiverse.go:1233).

Concrete failure:

1. The syncer calls `UpsertProofLeafBatch` with a 200-item chunk. tapdb/multiverse.go:885 runs the **whole chunk in one `ExecTx`**, so ids 905…1104 are allocated and the transaction stays open for the duration of 200 proof verifications + MS-SMT inserts (seconds, on the evidence of the benchmarks in this PR: ~266ms for *4* leaves).
2. Concurrently, the `InsertProof` RPC writes a single leaf via `UniverseArchive.UpsertProofLeaf` (rpcserver/rpcserver.go:7947) in its own short transaction, gets id 1105, and commits immediately. `tapgarden/planter.go:3186` is a second such writer. Neither takes `isSyncing`, so this is ordinary operation, not a contrived race.
3. A peer's `SyncDelta` lands in between. Its read tx sees 1105 but not 905…1104. It returns `latest_seq = 1105`.
4. The peer applies the page and persists cursor 1105. Ids 905…1104 are now permanently below its cursor and this query will never return them again.

`sql.LevelSerializable` (tapdb/interfaces.go:331) does not help. The schedule {short-writer, reader, batch-writer} is genuinely serializable, and Postgres exempts read-only transactions from the dangerous-structure check, so SSI has nothing to complain about. The anomaly isn't a serializability violation — it's a sequence-vs-commit-order violation, which lives outside the transactional model entirely.

I tried to refute this via the root check and it only partly holds. The claimed roots come from a *later* per-item `FetchProofLeaf` (rpcserver/rpcserver.go), so if the batch commits before the proof fetch the reported root includes 905 and the syncer's convergence check catches the mismatch and demotes to enumeration. That's the saving grace — but it needs the long batch tx to have committed within milliseconds of the item query, which is the *unlikely* case, not the likely one. And when the skipped leaf's universe has no other item in the page, it is never in `touched` at all and gets no root check under any timing (see the comment on universe/syncer_delta.go:214).

sqlite is safe: tapdb/sqlite.go opens with WAL + `_txlock=immediate`, so writers are serialized and rowid allocation order equals commit order. This is a Postgres-only exposure, and it's the deployment that matters for large universe servers. Postgres CI passing is not evidence against it — nothing in CI interleaves a long batch insert with a short single insert and reads the delta mid-flight.

The usual fix is to refuse to serve indices that could still be gapped: cap the page at a watermark below the lowest possible in-flight id (e.g. `AND leaves.id <= (SELECT COALESCE(MIN(id),  <max>) FROM <in-flight registry>)`), or in Postgres terms, don't advance `latest_seq` past `pg_snapshot_xmin`-equivalent safety — or simplest, add an explicit monotonic journal table whose sequence is assigned by the writer under a lock rather than by a bare sequence.

universe/syncer_delta.go
```diff
@@ -0,0 +1,322 @@
...
+
+	// With the delta applied, verify convergence per touched universe:
+	// the recomputed local root must equal the freshest remote root the
+	// pages reported. Mismatches (tainted universes, in-place re-org
+	// rewrites, journal anomalies) are demoted to enumeration sync.
+	var (
+		diffs   []AssetSyncDiff
+		demoted []Root
+	)
+	for idKey, id := range touched {
```
litbot-9000, line 265: **Blocker: when the delta is empty (or touches nothing), this loop body never executes, so nothing is verified — and the cursor is still treated as authoritative. A peer whose journal has been rewound is indistinguishable from a peer with nothing new.**

Trace the empty case: `FetchLeavesSince` with no matching rows returns `maxSeq = sinceSeq` (tapdb/multiverse.go:1214), the handler sets `resp.LatestSeq = req.SinceSeq`, the loop above breaks on `page.LatestSeq == cursor`, `touched` is empty, so **this loop does zero iterations and no root is ever compared**. The run returns success with the cursor unchanged. Same outcome whenever every item is filtered out by `insertFilter`.

That makes the following silently terminal, because the cursor is keyed only on `server_host` with no journal epoch (see tapdb/universe_federation.go:346):

- A peer restores an older DB backup, or is rebuilt, and re-inserts leaves at ids below our recorded cursor of 900. We ask `since=900` forever, get nothing, verify nothing, and never receive them.
- An operator repoints a hostname, or a different node takes over `host:port`. Our cursor 900 is now applied to an unrelated journal; everything below 900 is skipped permanently.
- The residual case from finding 1 above: any leaf that ended up below the cursor without being delivered.

I reproduced this with the PR's own harness (`memUniverseSet`, `newMemSyncer`, `requireConverged`). Local holds journal entries 1 and 3 but not 2; cursor = 3 = remote high-water mark:

```
local  root sum=6, remote root sum=9        <- diverged
SyncUniverseDelta -> err=nil, NewCursor=3, Diffs=[]   <- reports success, zero diffs
... 5 more delta syncs: still diverged
SyncUniverse (enumeration): healed in one pass
```

The doc comment on `SyncUniverseDelta` does acknowledge the shape of this ("a leaf missed under a stale cursor ... heals only when that universe next gains a leaf, or when a full enumeration sync runs") and tells callers to schedule occasional full syncs. But no caller does — see universe/federation.go:222 — so in practice the second clause never fires, and the first requires the universe to gain a new leaf, which for a fully-issued asset or a settled transfer universe may be never.

Two things would close it, and either is cheap. (a) A journal epoch: have the server return an instance/journal nonce alongside `latest_seq`, store it with the cursor, and reset the cursor to 0 on mismatch. (b) On an empty or no-touch delta, do one cheap authoritative check anyway — compare the peer's `MultiverseRoot` against ours and demote to enumeration on mismatch. (b) alone also covers finding 1's residual case, which is why I'd reach for it first.

universe/federation.go
```diff
@@ -193,12 +199,35 @@ func (f *FederationEnvoy) Stop() error {
...
+			ctx, deltaSyncer, addr, syncConfigs,
+		)
+		if err != nil {
+			// The delta path is an optimization: on failure we
+			// log and let the enumeration path below have a go.
+			log.Warnf("Delta sync with server=%v failed, "+
+				"falling back to enumeration sync: %v",
+				addr.HostStr(), err)
+		}
+		if done {
```
litbot-9000, line 286: **Blocker: `done == true` returns here without ever calling `SyncUniverse`, so once a peer supports delta sync the enumeration path is never exercised against it again — including on the empty-delta tick, which is the common case.**

`tryDeltaSync` returns `done = true` on every non-error outcome, and an empty delta is a non-error outcome. So the steady state for a delta-capable federation member is: enumeration sync runs exactly zero more times, forever.

That matters because the enumeration path is the *only* thing in-tree that repairs divergence which the delta path cannot see, and `SyncUniverseDelta`'s own doc comment says so explicitly: "Callers that need a hard bound on divergence should schedule occasional full syncs as an audit." Nothing does. The result is that every "heals only when a full enumeration sync runs" caveat in this PR — the Postgres index-ordering skip, the rewound-journal skip, the export-config skip — has no bound at all, rather than the bounded-divergence story the code comment implies.

The fix is small and it's the one the comment already prescribes: an audit cadence. E.g. force `done = false` (or run enumeration in addition) every Nth tick per server, or whenever more than some interval has elapsed since the last full sync for that server — the existing `last_sync_time` column is right there. A config knob defaulting to something like every 24h would make the safety argument for this whole design actually true, and it costs one enumeration sync per server per day.

Worth noting the fallback machinery itself is good: I checked all five pathological paths named in the design and four route correctly — `Unimplemented` from an older peer maps to `ErrDeltaUnsupported` (rpcserver/universe_rpc_diff.go) → enumeration; transport error mid-page → error → warn → enumeration; verification failure against a claimed root → taint → in-run `syncRoots` demotion; a failed demotion → run error → enumeration. Only the *empty/no-touch* response has no check at all, and that's the one this comment and the one on universe/syncer_delta.go:214 are about.

universe/syncer_delta.go
```diff
@@ -0,0 +1,322 @@
...
+		for _, item := range accepted {
+			idKey := item.ID.Key()
+			newLeaves[idKey] = append(newLeaves[idKey], item.Leaf)
+		}
+
+		// The cursor no longer advancing means the remote has nothing
+		// further for us. This also covers byte-budget-shortened
+		// pages, which advance the cursor and simply require another
+		// round trip.
+		if page.LatestSeq == cursor {
```
litbot-9000: **No monotonicity check on a peer-controlled value: only equality terminates this loop, so a peer that keeps returning `latest_seq != cursor` keeps it running indefinitely — and it holds the global sync lock while doing so.**

Two distinct problems:

1. **Unbounded loop.** A peer that alternates its answers — return `latest_seq = 5` for `since=100`, and `latest_seq = 100` for `since=5`, zero items each time — drives this loop forever at one RPC per iteration. The context is `WithCtxQuitNoTimeout()` (universe/federation.go:825), i.e. no deadline, cancelled only at daemon shutdown. And `SyncUniverseDelta` holds `s.isSyncing`, a single `atomic.Bool` on the one shared `SimpleSyncer` (universe/syncer.go:86; `tapcfg/server.go` builds one `universeSyncer`), so while it spins, *every other federation member's* sync fails the CAS — the delta path here and the enumeration path at universe/syncer.go:141 alike. One misbehaving or buggy federation member can therefore wedge federation sync with all honest members indefinitely. (The mutual exclusion is pre-existing on main; the unbounded duration is new, which is what turns "delays others by one sync" into "blocks others forever".)

2. **Cursor regression.** `latest_seq < cursor` walks the cursor backwards, and `tryDeltaSync` only tests `res.NewCursor != cursor` before persisting (universe/federation.go:279), so the regressed value is written to the DB. Safe in direction — it re-downloads rather than skips — but it lets a peer force us into a full re-sync of its history on every tick.

Both go away with a one-line guard, since a page that made progress always reports strictly greater:

```go
if page.LatestSeq <= cursor {
        break
}
```

This still handles the byte-budget-shortened page correctly, since that path sets `latest_seq` to the last *included* item, which is `> cursor` by construction. A belt-and-braces iteration cap, or a per-run deadline on the sync context, would also be reasonable given the lock this holds.

rpcserver/rpcserver.go
```diff
@@ -8197,6 +8197,176 @@ func (r *RPCServer) SyncUniverse(ctx context.Context,
...
+		rootsSeen = make(map[universe.IdentifierKey]struct{})
+	)
+
+	for i := range items {
+		item := items[i]
+
+		// Universes with export disabled are omitted, but their
+		// leaves still advance the cursor.
+		if !syncConfigs.IsSyncExportEnabled(item.ID) {
+			resp.LatestSeq = item.Seq
```
litbot-9000, line 8334: **Advancing the cursor past leaves that were deliberately not sent means a later config change can never deliver them.** `TestSyncDelta` pins this behaviour intentionally, so I want to be clear this is about the consequence, not the mechanism.

Sequence:

1. Server B has export disabled for universe U. Node A delta-syncs; U's leaves are skipped here but `resp.LatestSeq` advances past them, and A persists the advanced cursor.
2. An operator enables export for U on B — which is exactly what `SetFederationSyncConfig` exists to do.
3. A asks `since=<cursor>`. U's historical leaves are below the cursor, so they are never returned. U also never appears in the page, so it is never in `touched` and gets no root check (universe/syncer_delta.go:214). Nothing ever notices.

A never receives U. Under enumeration sync it would have on the very next tick, because enumeration re-derives the diff from current roots rather than from a remembered position.

The client-side mirror of this is `insertFilter` in universe/syncer_delta.go: a universe dropped for `AllowSyncInsert == false` likewise never enters `touched`, so a *local* operator enabling insert for an asset gets only future leaves, never the backlog.

The reasoning in the commit message — "the sequence is a position in this server's insertion log, not a count of exported items" — is right, and I don't think you want to stop advancing here. The problem is that the cursor is being asked to carry policy state it can't represent. Options: bump a config-generation counter (or just the cursor epoch from the finding on universe/syncer_delta.go:214) whenever a federation sync config changes, so cursors reset; or have `SetFederationSyncConfig` explicitly zero the affected cursors; or, most cheaply, let the periodic enumeration audit from universe/federation.go:222 cover it.

tapdb/universe_federation.go
```diff
@@ -302,6 +312,49 @@ func (u *UniverseFederationDB) LogNewSyncs(ctx context.Context,
...
+func (u *UniverseFederationDB) FetchSyncCursor(ctx context.Context,
+	addr universe.ServerAddr) (uint64, error) {
+
+	var (
+		readTx UniverseFederationOptions
+		cursor int64
+	)
+	dbErr := u.db.ExecTx(ctx, &readTx, func(db UniverseServerStore) error {
+		var err error
+		cursor, err = db.QueryFederationSyncCursor(
```
litbot-9000, line 347: The cursor is keyed solely on `addr.HostStr()` — a mutable address string with no node identity and no journal epoch. `UpsertSyncCursor` above does the same (`TargetServer: addr.HostStr()`).

An index is only meaningful relative to the journal that produced it, and nothing here binds the stored 900 to *the journal we were reading when we recorded 900*. Consequences, in increasing severity:

- Peer changes address → cursor resets to 0 → bootstrap re-sync. Harmless, just wasteful.
- Peer at the same `host:port` is rebuilt / restored from backup / replaced by a different node → cursor 900 is silently applied to an unrelated or rewound journal → everything below 900 is skipped, permanently and without any verification firing. See the comment on universe/syncer_delta.go:214 for the reproduced version of this.

Adding an opaque journal epoch to the `SyncDeltaResponse` (a per-instance nonce persisted alongside the journal, regenerated on any restore/reindex) and storing it in `universe_servers` next to `last_sync_seq` would make the cursor self-invalidating: epoch mismatch → reset to 0 → bootstrap. That's the standard construction for this, and it's a cheap column plus one proto field.

Two smaller things while in here: `UpsertSyncCursor` is named "Upsert" but is a plain `UPDATE ... WHERE server_host = ?`, so if the server row has been removed concurrently it silently affects zero rows and returns nil. And `QueryFederationSyncCursor` is `:one`, so this returns `sql.ErrNoRows` unwrapped if the server was removed between the envoy listing servers and this fetch — that surfaces as a delta-sync failure and falls back to enumeration, which is survivable, but mapping it to cursor 0 would be more obviously correct.

bench/fixture/discovery.go
```diff
@@ -0,0 +1,184 @@
...
+	b.Helper()
+
+	rootBytes := m.RootsEnumerated.Load() * approxRootWireBytes
+	keyBytes := m.LeafKeysEnumerated.Load() * approxLeafKeyWireBytes
+	proofBytes := m.ProofBytes.Load()
+
+	discoveryBytes := rootBytes + keyBytes
+	share := float64(0)
+	if total := discoveryBytes + proofBytes; total > 0 {
+		share = float64(discoveryBytes) / float64(total)
```
litbot-9000, line 70: Measurement caveat: `discovery_share` is 0 for the delta path partly by construction, so I'd be careful how it's quoted (the release note in this PR says "measured discovery overhead drops from ~88% ... to zero").

`discoveryBytes` counts only `RootsEnumerated * 73 + LeafKeysEnumerated * 69` — both of which are structurally zero on the delta path, since it calls neither `RootNodes` nor `UniverseLeafKeys`. Meanwhile the delta path's *own* discovery-ish overhead isn't counted anywhere: `page.Roots` carries a `UniverseRoot` per touched universe, and every item ships a `universe_inclusion_proof`, which `ProofBytes` explicitly excludes ("inclusion proof overhead is excluded"). So the metric can only ever return 0 for delta regardless of what the delta page actually costs on the wire.

That doesn't undermine the result — I ran both benchmarks and the effect is real and large (roots=10/leaves=400: enumeration 1493 ms/op with `leafkey_bytes 27600`, delta 266 ms/op with 2 pages and 4 items; enumeration grows 8× from leaves=100→400 while delta grows 2.4×, and delta's growth is MS-SMT insert/proof cost in a bigger tree rather than discovery). The accurate claim is "zero **enumeration** bytes", which is what the commit message says and what I'd put in the release note too. If you want `discovery_share` to mean what its name suggests across both paths, counting `len(page.Roots) * approxRootWireBytes` plus the compressed inclusion-proof bytes into the delta side would make the two columns comparable.

docs/release-notes/release-notes-0.8.2.md
```diff
@@ -1,71 +0,0 @@
-# Release Notes
```
litbot-9000, line 1: Unrelated change: this deletes `docs/release-notes/release-notes-0.8.2.md` in its entirety (71 lines), and the same commit re-adds its PR#2183 / PR#2184 / PR#2188 entries into `release-notes-0.9.0.md`. The file is present on `origin/main`, so this reads like a rebase artifact rather than an intentional edit — the net effect is that the 0.8.2 release notes disappear and three unrelated entries get duplicated into a different release. Worth dropping from the `docs: add release note` commit.

litbot-9000 (changes requested):
This is a genuinely well-built change — the layering is clean, the doc comments are unusually honest about their own assumptions, and the property test pins a real invariant. The design is also, I think, the right one. But I don't think it's safe to land as-is: the cursor can advance past leaves that were never delivered, and there is no mechanism left in-tree that will ever notice. I'm requesting changes on that basis, not on the shape of the design.

**Design review.** The cursor model is sound *as specified*. Deriving the journal from the pre-existing `universe_leaves.id` is a nice move — it means existing nodes need no backfill, which removes the entire class of "migrated node strands its history" problem I went looking for. Making the cursor an unauthenticated hint whose every acceptance decision still rests on inclusion-proof + root-equality checks is the right safety posture, and it holds: `Archive.UpsertProofLeafBatch` (universe/archive.go:449) does full proof verification and already partitions group anchors, so delta acceptance is not weaker than enumeration acceptance. Cursor persistence ordering is also correct — leaves are written first and the cursor only afterwards, and only after every touched universe verified, so a crash mid-run re-delivers rather than skips. That is the opposite of the #2203 class of bug, deliberately. Where the specification is under-strength is the two places it assumes something it does not check: that the serving journal's *index order* equals its *commit order*, and that a peer's journal at index N still means what it meant when we recorded N. Both are fixable within this design (a stability watermark; a journal epoch); neither requires rethinking it.

**What I verified.** Reviewed commit-by-commit at 495de3fb. `make build` → exit 0. `go test ./universe/... ./rpcserver/...` → ok. `go test ./tapdb/...` → ok (49.5s). `go test ./tapdb/ -run TestMigration -v` → PASS (note there is no targeted `TestMigration61`; migration 61 is only exercised by the generic migrate-to-latest path). One sequential `go test -race ./universe/` → ok (23.7s). `gh pr checks 2202` → 40/40 green, including `run unit tests (unit dbbackend=postgres pkg=tapdb)`, `run itests postgres (0-3)`, `run unit race (0-3)`, `Sqlc check` and `migration version check`. I could **not** run any Postgres path myself — there is no container runtime in this environment, so everything behind `ory/dockertest` / `test_db_postgres` was unavailable; for dialect portability I read both paths and am leaning on that CI. Migration 61 is a single dialect-neutral `ADD COLUMN ... BIGINT NOT NULL DEFAULT 0` with a working down migration, and the new query uses no dialect-specific constructs, so I have no portability concern there. I also **ran the benchmarks** rather than taking them on faith, and they reproduce: at roots=10/leaves=400, enumeration reports `discovery_share 0.8794` / `leafkeys_enum 400` / `leafkey_bytes 27600` (1493 ms/op), delta reports `discovery_share 0` / `delta_pages 2` / `delta_items 4` (266 ms/op). Both scenarios deliver the same 4 new leaves, so the comparison is fair, and enumeration's 8× growth from leaves=100→400 against delta's 2.4× is the claimed effect. Finally I reproduced finding 2 empirically using the PR's own `memUniverseSet` harness — details in that comment.

Findings inline, most severe first: (1) journal index is not commit-ordered on Postgres, so the cursor can pass an uncommitted leaf; (2) an empty delta is accepted with zero verification, so a rewound peer journal or a reused host string strands leaves permanently — reproduced; (3) enumeration sync never runs again once delta succeeds, so the "schedule occasional full syncs as an audit" mitigation the code recommends does not exist anywhere; (4) the page loop has no monotonicity check and can be driven forever by a peer, holding the global sync lock; (5) export-config toggles strand leaves below the cursor; plus a benchmark-measurement caveat and an unrelated release-note deletion.

I have not restated the two points gemini-code-assist raised. One factual note: the reply "Fixed in updated changeset" on the `syncDeltaByteBudget` package-level variable does not appear to be reflected at this head — rpcserver/rpcserver.go:8204 is still `var syncDeltaByteBudget`.

docs/release-notes/release-notes-0.8.2.md
jtobin: This is correct.

tapdb/sqlc/queries/universe.sql
jtobin: Fixed in updated changeset.

universe/syncer_delta.go
jtobin: Fixed in updated changeset. Fable's full commentary:

> Fixed, though not via the multiverse-root compare: that's unsound for nodes syncing a subset of the peer (or holding local-only assets), where the roots differ perpetually and every empty delta would demote to enumeration. Instead the server now reports its committed journal tail on empty pages. Tail below cursor surfaces as ErrJournalRewind; the envoy resets the cursor to the tail and runs enumeration that round. The residual case (journal replaced and regrown past our cursor) is bounded by the periodic enumeration audit (see the federation.go thread).

universe/federation.go
jtobin: Fixed in updated changeset.

universe/syncer_delta.go
jtobin: Fixed in updated changeset.

tapdb/universe_federation.go
jtobin: Fable:

> Deliberately not adding the epoch: the rewind detection covers restore/repoint with a tail below our cursor, and the audit bounds the regrown-past-cursor case, so the epoch would add proto and schema surface without improving the bound. The ErrNoRows nit is fixed (unknown server reads as cursor zero). The UPDATE-shaped upsert stays: a concurrently removed server's cursor is moot, and re-adding starts from the row default of zero (pinned in TestFederationSyncCursor).

rpcserver/rpcserver.go
jtobin: Fixed in update changeset (backlog case covered by the enumeration audit).

bench/fixture/discovery.go
jtobin: Fable reworded the release note.

jtobin:
This one has gone through a bunch of adjustments after review, so I ran some benchmarks on the final state of the branch to summarize the perf improvements this provides, as it's easy to forget them amidst the churn. Fable's TLDR summary of the wins:

--

- Steady-state discovery collapses: discovery bytes go from O(corpus) to O(changes), i.e. ~1000x at a 50k-root corpus (extrapolated).
- Per-tick latency: ~2–3x at bench scale, ~50x at production corpus size (extrapolated).
- Busy tick, divergence spread across 50 universes: 201 → 2 round trips (100x), 4.7x fewer bytes, 1.8x wall in-process; ~40x latency over a 50 ms link (extrapolated).
- Server load: discovery queries per tick drop from O(corpus) per client to O(new leaves) — multiplicative across federation clients (extrapolated from counters).
- Cold start: 4021 → 9 server requests (~450x), 1.65x wall clock (measured; insert-bound).
- Downside bounds: worst-case (demoted) tick costs the same as enumeration today (461 vs 483 ms); ingest journal tax ≤ ~8% on SQLite, unmeasurable on Postgres.

The ~1000x/50x figures assume a 50k-root corpus and 512-entry pages.

universe/syncer_delta.go
```diff
@@ -0,0 +1,373 @@
...
+
+				case err != nil:
+					return nil, fmt.Errorf("unable to "+
+						"fetch local root: %w", err)
+
+				default:
+					oldLocalRoots[idKey] = localRoot
+				}
+
+				touched[idKey] = item.ID
```
GeorgeTsagk, line 182: `touched` is recorded before the root check runs, so a universe whose root the remote never reported still ends up in here. The final loop then hits the `no remote root reported` return at line 268 and fails the whole run, even though the taint path just above it wanted to demote that universe to enumeration.

Either record `touched` after verification succeeds, or route a missing root into `demoted` instead of returning an error.

rpcserver/rpcserver.go
```diff
@@ -8236,6 +8236,184 @@ func (r *RPCServer) SyncUniverse(ctx context.Context,
...
+			continue
+		}
+
+		// Fetch the leaf's inclusion proof along with its universe
+		// root.
+		proofs, err := r.cfg.UniverseArchive.FetchProofLeaf(
```
GeorgeTsagk: `FetchProofLeaf` runs once per item, each in its own read transaction, but `resp.UniverseRoots` only records the root from the first item of each universe. If a write lands on that universe partway through the loop, the later items get inclusion proofs against a newer root, and the client verifies them against the older one. It taints the universe and demotes it to enumeration.

For a universe that is actively being written to, that is the normal case rather than a rare one, so the steady state the benchmarks measure may not be the steady state in production. Serving the whole page from a single read transaction fixes both this and the N+1.

rpcserver/rpcserver.go
```diff
@@ -8236,6 +8236,184 @@ func (r *RPCServer) SyncUniverse(ctx context.Context,
...
+// syncDeltaByteBudget caps the approximate payload size of a single
+// SyncDelta response page, keeping it comfortably under the default
+// 4 MiB gRPC message size limit. A variable rather than a constant so
+// tests can exercise the page-splitting path with small payloads.
+var syncDeltaByteBudget = 3 * 1024 * 1024
+
+// syncDeltaItemOverhead is the approximate per-item wire overhead of a
+// SyncDeltaItem beyond its raw proof and inclusion proof blobs
+// (universe ID, leaf key, decoded asset fields, framing).
+const syncDeltaItemOverhead = 512
```
GeorgeTsagk: 512 bytes looks optimistic for the marshalled `AssetLeaf`. `PrevWitnesses` can carry split commitment proofs, which embed a nested asset plus a merkle proof and run to several KB. `resp.UniverseRoots` is not counted at all.

With a 3 MiB budget against the 4 MiB default there is 1 MiB of slack, so this is not far off, but overshooting fails the page for exactly the assets that are hardest to sync. `proto.Size(rpcItem)` after marshalling is exact and roughly the same cost.

GeorgeTsagk (commented):
A few things I did not see covered in the existing threads.

universe/syncer_delta.go
jtobin: wontfix, per Opus:

> Reachability is nil against a conforming server. SyncDelta appends the universe root inside the same iteration that appends the item (rpcserver.go:8397), after the budget break, so every item in a page has a root. remoteRoots is never cleared across pages. So !haveRoot requires a buggy or hostile remote — and in that case the envoy already catches the error and falls back to full enumeration (federation.go:277-283).  Correct outcome, wrong route, misleading log line.

rpcserver/rpcserver.go
jtobin: Fixed in updated changeset.

rpcserver/rpcserver.go
jtobin: Fixed in updated changeset.

universe/federation.go
GeorgeTsagk: The audit is load bearing now: it's the stated bound for the rewind residual, for the export backlog on rpcserver.go:8334, and for the regrown-past-cursor case you cited when declining the epoch. Three deferrals, one mechanism.

It doesn't survive a restart. `lastEnumSync` is an in-memory map and `enumerationAuditDue` seeds `time.Now()` on the first call per host, then returns false. A daemon restarting more often than `SyncAuditInterval` never audits. Default is 24h.

Persist the last audit time next to `last_sync_seq`.

universe/syncer_delta.go
GeorgeTsagk: Tail signal looks right. Ordering issue in how the envoy acts on it: federation.go:350 persists `rewindErr.Tail`, then returns done=false so enumeration runs. If that enumeration at :295 errors, the cursor is already committed at the tail. Next tick takes the delta path, gets an empty page, and everything below the tail is never delivered.

Transient failure becomes a permanent skip. Persist the reset after enumeration returns cleanly.

tapdb/universe_federation.go
GeorgeTsagk: Tail-instead-of-epoch is fair, but the tail as derived fires when nothing rewound. `MaxUniverseLeafJournalSeq` is `MAX(seq)` over `universe_leaf_journal`, and `leaf_id` is `ON DELETE CASCADE`. Deleting the top journalled leaf (which `DeleteAssetLeaf` does) drops the max, so every peer's next empty page reports a tail below its cursor and they all reset and run full enumeration over one delete.

`universe_leaf_journal_tail.tail` avoids it: monotonic, unaffected by deletes, and still a correct high water mark since an uncommitted bump isn't visible and a committed tail implies all lower seqs are committed.

GeorgeTsagk:
Flagging as not yet addressed:

- https://github.com/lightninglabs/taproot-assets/pull/2202#discussion_r3821523797
- https://github.com/lightninglabs/taproot-assets/pull/2202#discussion_r3821523969
- https://github.com/lightninglabs/taproot-assets/pull/2202#discussion_r3821524108

universe/federation.go
jtobin: wontfix; a daemon that wants to restart regularly should set a shorter audit interval.

tapdb/universe_federation.go
jtobin: wontfix: this makes sense to me, but 1) the claimed problematic condition is rare (leaf deletes are giga rare, and manually-performed), 2) the claimed problem is benign (a single enumeration sync, i.e. what we do at present), and 3) fixing it would add another query, etc. The juice is not worth the squeeze at the moment.

universe/syncer_delta.go
jtobin: This one seemed worth fixing; fixed in 37c6b97a.

jtobin:
> Flagging as not yet addressed:
> 
> [..]

Should be good now. 💪

rpcserver/rpcserver.go
```diff
@@ -8248,6 +8249,162 @@ func (r *RPCServer) SyncUniverse(ctx context.Context,
...
+
+	pageSize := req.PageSize
+	switch {
+	case pageSize == 0:
+		pageSize = universe.RequestPageSize
+
+	case pageSize < 0:
+		return nil, fmt.Errorf("invalid page size %d", pageSize)
+
+	case pageSize > universe.MaxPageSize:
```
GeorgeTsagk, line 8275: `page_size` is caller controlled up to `MaxPageSize` (16384), and `SyncDelta` is unauthenticated whenever `universe.public-access` includes `r`.

Before the single-transaction refactor the inclusion proof was fetched per item inside the handler loop, so the byte budget's `break` stopped the work at ~3 MiB. Now `FetchDeltaPage` runs `universeFetchProofLeaf` for all 16384 rows inside one read transaction and the budget only trims the marshalled response afterwards. So one rate limiter token buys 16384 mssmt proof generations to return maybe a few hundred of them.

Clamping to `RequestPageSize` here instead of rejecting above `MaxPageSize` bounds it, and no honest caller is affected: the client always sends `defaultPageSize` (512). Passing the budget down into assembly would be the thorough version, but the clamp gets the same bound for a line.

GeorgeTsagk (approved):
LGTM

rpcserver/rpcserver.go
GeorgeTsagk: (non-blocking recommendation)

Issue #2136, opened 2026-05-14
[bug]: Mint batch can get stuck in FROZEN state after funding failure


vctt94:
### Background

While testing `tapd` on regtest, I found that `assets mint finalize` can leave a minting batch stuck in `BATCH_STATE_FROZEN` if funding the minting PSBT fails because `lnd` has no confirmed BTC available.

After funding `lnd`, retrying `finalize` returns `no pending batch`, and `assets mint fund` creates a new empty pending batch instead of recovering the frozen one.

### Your environment

```
lnd master ❯ ./lncli-debug   --network=regtest   --lnddir=/mnt/storage/taproot-repro/lnd   --rpcserver=127.0.0.1:11009   getinfo
{
    "version": "0.20.99-beta commit=actor/v0.0.6-137-gb9b299bec",
    "commit_hash": "b9b299beca61ee7dc9344e88303876e7fcc3d269",
    "identity_pubkey": "02a606972f4ab4b72acd1310d0e48f4a8589618fefdf32e9f47923d323eab5baa9",
    "alias": "02a606972f4ab4b72acd",
    "color": "#3399ff",
    "num_pending_channels": 0,
    "num_active_channels": 0,
    "num_inactive_channels": 0,
    "num_peers": 0,
    "block_height": 103,
    "block_hash": "762d06159d2d284c93837c5b9114aac88510bd852ecafacaf1cd973abe1ace06",
    "best_header_timestamp": "1778780543",
    "synced_to_chain": true,
    "synced_to_graph": false,
    "testnet": false,
    "chains": [
        {
            "chain": "bitcoin",
            "network": "regtest"
        }
    ],
```
```
taproot-assets main ✗ ./tapcli-debug   --network=regtest   --tapddir=/mnt/storage/taproot-repro/tapd   getinfo
{
    "version":  "0.7.99-alpha commit=v0.7.0-626-g9ba680e0",
    "lnd_version":  "0.20.99-beta",
    "network":  "regtest",
    "lnd_identity_pubkey":  "02a606972f4ab4b72acd1310d0e48f4a8589618fefdf32e9f47923d323eab5baa9",
    "node_alias":  "02a606972f4ab4b72acd",
    "block_height":  103,
    "block_hash":  "762d06159d2d284c93837c5b9114aac88510bd852ecafacaf1cd973abe1ace06",
    "sync_to_chain":  true
}
```

```sh
$ bitcoind --version
# bitcoind regtest backend
```

[...]

PR #2203, opened 2026-07-10
tapgarden: hardening phase two (batch invariants)

jtobin:
This is the next chunk of the large-scale tapgarden hardening work originally contained in #2153 (following #2190). It handles the following bullets from the original PR description:

> **Persistence atomicity**. Previously it was possible for an in-memory batch state to advance while the corresponding database write failed, leaving memory and disk disagreeing about what state the batch was in; now, because every state mutation flows through a store method that touches memory only after the database commit succeeds, the two views cannot drift.
> 
> **Funding overload**. Previously it was possible for a fund-batch call to silently create a new batch when the caller meant to attach funding to an existing one (the bug behind #2136); now, because "create" and "attach" are separate functions with non-overlapping signatures, the wrong intent will not type-check.
> 
> **Singleton invariant**. Previously it was possible for two unbroadcast minting batches to coexist on disk and block all subsequent minting; now, because the schema itself forbids more than one row in the pre-broadcast subset, that state cannot exist.
> 
> **Determinism**. Previously it was possible for the planter to pick a different "anchor seedling" on different runs of the same input because the choice depended on Go map iteration order; now, because the lookup scans deterministically and refuses to return when the singleton invariant fails, the choice is fixed.

It contains the following changes (summary ctsy Opus):

**Two-truth collapse.** `MintingBatch.batchState` (in-memory) and the `asset_minting_batches.state` column (on-disk) were maintained by separate code paths and could diverge under failure -- store methods sometimes updated the DB while the in-memory mirror was patched up later, or in places got ahead of the DB entirely. Every state mutation now flows through a `MintingStore` method that advances the in-memory field only after the DB write commits; the two views can no longer drift.

**FundBatch split.** `fundBatch` was overloaded: same call either attached funding to a pending batch or silently minted a new one, depending on hidden state. It splits into a `create` half and an `apply` half with disjoint signatures, so the caller's intent is carried by the function chosen rather than by an implicit side condition.

**Deterministic anchor lookup.** Two call sites that fetched the "anchor seedling" iterated `pendingBatch.Seedlings` and took the first match, relying on Go map iteration order and an unenforced <=1-anchor invariant validated elsewhere. Replaced with a deterministic scan that returns the unique seedling with `GroupAnchor == nil` and errors loudly on zero or multiple matches -- the invariant is now asserted at the point it's depended on.

Singleton invariant, enforced. The planter has always treated the pre-broadcast batch as a singleton in memory, but the schema admitted plurality; older planter versions could desync the in-memory slot from disk under certain failure paths and leave two pre-broadcast rows behind. Migration 000061 adds a partial unique index on asset_minting_batches forbidding more than one row in Pending or Frozen at a time. Legacy databases already holding duplicates are healed as part of applying the migration: all but the most recent pre-broadcast batch (latest creation time, ties broken by the raw batch key) move to SeedlingCancelled, leaving the rows and their seedlings on disk for inspection. A one-shot repair tool (tapd --repair.cancel-duplicate-batches, honored from the command line only) performs the same repair outside the migration stream as a diagnostic.

### DIFF
--- a/cmd/tapd/main.go
+++ b/cmd/tapd/main.go
@@ -35,6 +35,20 @@ func main() {
 		os.Exit(0)
 	}
 
+	// If the operator has invoked tapd in one-shot repair mode, run
+	// the requested repair against the database and exit before
+	// constructing the full server. The repair tool opens the DB
+	// with migrations skipped, so it can recover a legacy DB whose
+	// state would otherwise block a migration from applying.
+	if cfg.Repair != nil && cfg.Repair.CancelDuplicateBatches {
+		err := tapcfg.RunRepairTool(cfg, cfgLogger, shutdownInterceptor)
+		if err != nil {
+			_, _ = fmt.Fprintln(os.Stderr, err)
+			os.Exit(1)
+		}
+		os.Exit(0)
+	}
+
 	// Enable http profiling server if requested.
 	if cfg.Profile != "" {
 		go func() {
--- a/cmd/tapd/main.go
+++ b/cmd/tapd/main.go
@@ -5,12 +5,27 @@ import (
 	"net/http"
 	"os"
 	"runtime/pprof"
+	"strings"
 
 	"github.com/jessevdk/go-flags"
 	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightninglabs/taproot-assets/tapcfg"
 	"github.com/lightningnetwork/lnd/signal"
 )
+
+// repairFlagOnCommandLine reports whether the one-shot repair flag
+// was passed on the command line itself, as opposed to being read
+// from a config file.
+func repairFlagOnCommandLine() bool {
+	const flag = "--repair.cancel-duplicate-batches"
+	for _, arg := range os.Args[1:] {
+		if arg == flag || strings.HasPrefix(arg, flag+"=") {
+			return true
+		}
+	}
+
+	return false
+}
 
 func main() {
 	// Hook interceptor for os signals.
@@ -33,6 +48,34 @@ func main() {
 
 		// Help was requested, exit normally.
 		os.Exit(0)
+	}
+
+	// If the operator has invoked tapd in one-shot repair mode, run
+	// the requested repair against the database and exit before
+	// constructing the full server. The repair tool opens the DB
+	// with migrations skipped, so it can recover a legacy DB whose
+	// state would otherwise block a migration from applying.
+	//
+	// The flag is only honored when given on the command line
+	// itself. Repair mode exits after running, so a value left
+	// behind in tapd.conf would otherwise turn every subsequent
+	// start into a no-op exit with a success code, silently taking
+	// a supervised node out of service.
+	if cfg.Repair != nil && cfg.Repair.CancelDuplicateBatches {
+		if repairFlagOnCommandLine() {
+			err := tapcfg.RunRepairTool(
+				cfg, cfgLogger, shutdownInterceptor,
+			)
+			if err != nil {
+				_, _ = fmt.Fprintln(os.Stderr, err)
+				os.Exit(1)
+			}
+			os.Exit(0)
+		}
+
+		cfgLogger.Warnf("Ignoring repair.cancel-duplicate-batches " +
+			"from the config file: repair mode exits after " +
+			"running and must be requested on the command line")
 	}
 
 	// Enable http profiling server if requested.
--- a/docs/release-notes/release-notes-0.9.0.md
+++ b/docs/release-notes/release-notes-0.9.0.md
@@ -25,6 +25,10 @@
   fixes a bug that could cause minted assets to commit to the wrong
   address.
 
+* [PR#2203](https://github.com/lightninglabs/taproot-assets/pull/2203)
+  fixes a bug that could leave invalid minting batch state on disk after
+  failure.
+
 # New Features
 
 ## Functional Enhancements
--- a/docs/release-notes/release-notes-0.9.0.md
+++ b/docs/release-notes/release-notes-0.9.0.md
@@ -25,6 +25,15 @@
   fixes a bug that could cause minted assets to commit to the wrong
   address.
 
+* [PR#2203](https://github.com/lightninglabs/taproot-assets/pull/2203)
+  fixes a bug that could leave invalid minting batch state on disk after
+  failure. A database migration now enforces that at most one minting
+  batch is in a pre-broadcast (pending or frozen) state; on upgrade, if
+  a legacy database holds several such batches, all but the most
+  recently created one are automatically cancelled. A new
+  `--repair.cancel-duplicate-batches` startup flag applies the same
+  repair on demand.
+
 * [PR#2228](https://github.com/lightninglabs/taproot-assets/pull/2228)
   fixes `VerifyProof` failing the whole RPC for an invalid proof when
   decoding the last proof requires unknown asset metadata; it now
--- a/sample-tapd.conf
+++ b/sample-tapd.conf
@@ -544,3 +544,12 @@
 ; The amount of time we should wait between certificate expiration health checks.
 ; This value must be >= 1m.
 ; healthcheck.tls.interval=1m
+
+[repair]
+
+; If set, tapd cancels all but the most recent minting batch in
+; BatchStatePending or BatchStateFrozen and then exits. Used to recover
+; from a database that violates the singleton pre-broadcast batch
+; invariant added in migration 000061 (e.g. a legacy DB with duplicate
+; pending batches that blocks the migration).
+; repair.cancel-duplicate-batches=false
--- a/tapcfg/repair.go
+++ b/tapcfg/repair.go
@@ -1,0 +1,151 @@
+package tapcfg
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"sort"
+	"time"
+
+	"github.com/btcsuite/btclog/v2"
+	"github.com/lightninglabs/taproot-assets/tapdb"
+	"github.com/lightninglabs/taproot-assets/tapgarden"
+	"github.com/lightningnetwork/lnd/signal"
+)
+
+// RunRepairTool inspects the configured database for batches that
+// violate the singleton "≤ 1 in {Pending, Frozen}" invariant added
+// in migration 000060, and cancels all but the most recent. The
+// preserved batch is the one with the latest CreationTime; cancelled
+// batches transition to BatchStateSeedlingCancelled, leaving their
+// row and seedlings on disk for later inspection.
+//
+// The function opens the database with migrations skipped, so it can
+// run against a legacy database whose state would otherwise fail the
+// migration.
+//
+// NOTE: With migration 000061's self-heal in place, restarting tapd
+// normally will cancel the duplicates as part of applying the
+// migration. This tool is retained as a diagnostic that surfaces the
+// same repair outside the migration stream (e.g. after operator
+// intervention that re-introduces duplicates).
+func RunRepairTool(cfg *Config, cfgLogger btclog.Logger,
+	shutdownInterceptor signal.Interceptor) error {
+
+	// Derive a cancellable context that trips on shutdown. Without
+	// this, a Ctrl+C mid-repair would leave partial state behind
+	// (some batches cancelled, others not).
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func() {
+		select {
+		case <-shutdownInterceptor.ShutdownChannel():
+			cancel()
+		case <-ctx.Done():
+		}
+	}()
+
+	// Open the database with migrations skipped. We want to inspect
+	// and repair a database whose state would otherwise prevent
+	// migration 000060 from applying; running migrations as part of
+	// opening the DB would defeat the purpose.
+	var (
+		db  tapdb.DatabaseBackend
+		err error
+	)
+	switch cfg.DatabaseBackend {
+	case DatabaseBackendSqlite:
+		sqliteCfg := *cfg.Sqlite
+		sqliteCfg.SkipMigrations = true
+		cfgLogger.Infof("repair: opening sqlite3 database at %v "+
+			"(migrations skipped)",
+			sqliteCfg.DatabaseFileName)
+		db, err = tapdb.NewSqliteStore(&sqliteCfg)
+
+	case DatabaseBackendPostgres:
+		pgCfg := *cfg.Postgres
+		pgCfg.SkipMigrations = true
+		cfgLogger.Infof("repair: opening postgres database " +
+			"(migrations skipped)")
+		db, err = tapdb.NewPostgresStore(&pgCfg)
+
+	default:
+		return fmt.Errorf("unknown database backend: %s",
+			cfg.DatabaseBackend)
+	}
+	if err != nil {
+		return fmt.Errorf("repair: unable to open database: %w", err)
+	}
+
+	mintingExec := tapdb.NewTransactionExecutor(
+		db, func(tx *sql.Tx) tapdb.PendingAssetStore {
+			return db.WithTx(tx)
+		},
+	)
+	store := tapdb.NewAssetMintingStore(mintingExec)
+
+	nonFinal, err := store.FetchNonFinalBatches(ctx)
+	if err != nil {
+		return fmt.Errorf("repair: unable to fetch non-final "+
+			"batches: %w", err)
+	}
+
+	var preBroadcast []*tapgarden.MintingBatch
+	for _, batch := range nonFinal {
+		switch batch.State() {
+		case tapgarden.BatchStatePending,
+			tapgarden.BatchStateFrozen:
+
+			preBroadcast = append(preBroadcast, batch)
+
+		default:
+			// Only pre-broadcast states are constrained by
+			// the singleton index; ignore everything else.
+		}
+	}
+
+	if len(preBroadcast) <= 1 {
+		cfgLogger.Infof("repair: nothing to do; found %d batches "+
+			"in pre-broadcast state", len(preBroadcast))
+		return nil
+	}
+
+	// Sort newest-first by CreationTime; preserve [0], cancel the
+	// rest. SliceStable gives a deterministic winner when two
+	// batches share a timestamp -- the input order (from
+	// FetchNonFinalBatches) then acts as the tiebreak.
+	sort.SliceStable(preBroadcast, func(i, j int) bool {
+		return preBroadcast[i].CreationTime.After(
+			preBroadcast[j].CreationTime,
+		)
+	})
+
+	preserved := preBroadcast[0]
+	cfgLogger.Infof("repair: preserving most recent pre-broadcast "+
+		"batch %x (state=%v, created=%s)",
+		preserved.BatchKey.PubKey.SerializeCompressed(),
+		preserved.State(),
+		preserved.CreationTime.Format(time.RFC3339))
+
+	for _, batch := range preBroadcast[1:] {
+		cfgLogger.Warnf("repair: cancelling pre-broadcast batch "+
+			"%x (state=%v, created=%s)",
+			batch.BatchKey.PubKey.SerializeCompressed(),
+			batch.State(),
+			batch.CreationTime.Format(time.RFC3339))
+
+		err := store.UpdateBatchState(
+			ctx, batch, tapgarden.BatchStateSeedlingCancelled,
+		)
+		if err != nil {
+			return fmt.Errorf("repair: unable to cancel batch "+
+				"%x: %w",
+				batch.BatchKey.PubKey.SerializeCompressed(),
+				err)
+		}
+	}
+
+	cfgLogger.Infof("repair: complete; cancelled %d duplicate "+
+		"batches, preserved 1.", len(preBroadcast)-1)
+	return nil
+}
--- a/tapcfg/repair.go
+++ b/tapcfg/repair.go
@@ -1,0 +1,171 @@
+package tapcfg
+
+import (
+	"bytes"
+	"context"
+	"database/sql"
+	"fmt"
+	"io"
+	"sort"
+	"time"
+
+	"github.com/btcsuite/btclog/v2"
+	"github.com/lightninglabs/taproot-assets/tapdb"
+	"github.com/lightninglabs/taproot-assets/tapgarden"
+	"github.com/lightningnetwork/lnd/signal"
+)
+
+// RunRepairTool inspects the configured database for batches that
+// violate the singleton "≤ 1 in {Pending, Frozen}" invariant added
+// in migration 000061, and cancels all but the most recent. The
+// preserved batch is the one with the latest CreationTime; cancelled
+// batches transition to BatchStateSeedlingCancelled, leaving their
+// row and seedlings on disk for later inspection.
+//
+// The function opens the database with migrations skipped, so it can
+// run against a legacy database whose state would otherwise fail the
+// migration.
+//
+// NOTE: With migration 000061's self-heal in place, restarting tapd
+// normally will cancel the duplicates as part of applying the
+// migration. This tool is retained as a diagnostic that surfaces the
+// same repair outside the migration stream (e.g. after operator
+// intervention that re-introduces duplicates).
+func RunRepairTool(cfg *Config, cfgLogger btclog.Logger,
+	shutdownInterceptor signal.Interceptor) error {
+
+	// Derive a cancellable context that trips on shutdown. Without
+	// this, a Ctrl+C mid-repair would leave partial state behind
+	// (some batches cancelled, others not).
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func() {
+		select {
+		case <-shutdownInterceptor.ShutdownChannel():
+			cancel()
+		case <-ctx.Done():
+		}
+	}()
+
+	// Open the database with migrations skipped. We want to inspect
+	// and repair a database whose state would otherwise prevent
+	// migration 000061 from applying; running migrations as part of
+	// opening the DB would defeat the purpose.
+	var (
+		db  tapdb.DatabaseBackend
+		err error
+	)
+	switch cfg.DatabaseBackend {
+	case DatabaseBackendSqlite:
+		sqliteCfg := *cfg.Sqlite
+		sqliteCfg.SkipMigrations = true
+		cfgLogger.Infof("repair: opening sqlite3 database at %v "+
+			"(migrations skipped)",
+			sqliteCfg.DatabaseFileName)
+		db, err = tapdb.NewSqliteStore(&sqliteCfg)
+
+	case DatabaseBackendPostgres:
+		pgCfg := *cfg.Postgres
+		pgCfg.SkipMigrations = true
+		cfgLogger.Infof("repair: opening postgres database " +
+			"(migrations skipped)")
+		db, err = tapdb.NewPostgresStore(&pgCfg)
+
+	default:
+		return fmt.Errorf("unknown database backend: %s",
+			cfg.DatabaseBackend)
+	}
+	if err != nil {
+		return fmt.Errorf("repair: unable to open database: %w", err)
+	}
+
+	// Both backend stores embed the underlying *sql.DB; release it
+	// on the way out.
+	if closer, ok := db.(io.Closer); ok {
+		defer func() {
+			if err := closer.Close(); err != nil {
+				cfgLogger.Warnf("repair: error closing "+
+					"database: %v", err)
+			}
+		}()
+	}
+
+	mintingExec := tapdb.NewTransactionExecutor(
+		db, func(tx *sql.Tx) tapdb.PendingAssetStore {
+			return db.WithTx(tx)
+		},
+	)
+	store := tapdb.NewAssetMintingStore(mintingExec)
+
+	nonFinal, err := store.FetchNonFinalBatches(ctx)
+	if err != nil {
+		return fmt.Errorf("repair: unable to fetch non-final "+
+			"batches: %w", err)
+	}
+
+	var preBroadcast []*tapgarden.MintingBatch
+	for _, batch := range nonFinal {
+		switch batch.State() {
+		case tapgarden.BatchStatePending,
+			tapgarden.BatchStateFrozen:
+
+			preBroadcast = append(preBroadcast, batch)
+
+		default:
+			// Only pre-broadcast states are constrained by
+			// the singleton index; ignore everything else.
+		}
+	}
+
+	if len(preBroadcast) <= 1 {
+		cfgLogger.Infof("repair: nothing to do; found %d batches "+
+			"in pre-broadcast state", len(preBroadcast))
+		return nil
+	}
+
+	// Sort newest-first by CreationTime; preserve [0], cancel the
+	// rest. The full stored timestamp precision is compared
+	// (sub-second parts included), and ties are broken by the raw
+	// batch key, matching the ORDER BY used by migration 000061's
+	// self-heal so both paths preserve the same batch.
+	sort.Slice(preBroadcast, func(i, j int) bool {
+		ti := preBroadcast[i].CreationTime
+		tj := preBroadcast[j].CreationTime
+		if !ti.Equal(tj) {
+			return ti.After(tj)
+		}
+
+		ki := preBroadcast[i].BatchKey.PubKey.SerializeCompressed()
+		kj := preBroadcast[j].BatchKey.PubKey.SerializeCompressed()
+		return bytes.Compare(ki, kj) > 0
+	})
+
+	preserved := preBroadcast[0]
+	cfgLogger.Infof("repair: preserving most recent pre-broadcast "+
+		"batch %x (state=%v, created=%s)",
+		preserved.BatchKey.PubKey.SerializeCompressed(),
+		preserved.State(),
+		preserved.CreationTime.Format(time.RFC3339))
+
+	for _, batch := range preBroadcast[1:] {
+		cfgLogger.Warnf("repair: cancelling pre-broadcast batch "+
+			"%x (state=%v, created=%s)",
+			batch.BatchKey.PubKey.SerializeCompressed(),
+			batch.State(),
+			batch.CreationTime.Format(time.RFC3339))
+
+		err := store.UpdateBatchState(
+			ctx, batch, tapgarden.BatchStateSeedlingCancelled,
+		)
+		if err != nil {
+			return fmt.Errorf("repair: unable to cancel batch "+
+				"%x: %w",
+				batch.BatchKey.PubKey.SerializeCompressed(),
+				err)
+		}
+	}
+
+	cfgLogger.Infof("repair: complete; cancelled %d duplicate "+
+		"batches, preserved 1.", len(preBroadcast)-1)
+	return nil
+}
--- a/tapdb/asset_minting.go
+++ b/tapdb/asset_minting.go
@@ -1319,7 +1319,7 @@ func marshalMintingBatch(ctx context.Context, q PendingAssetStore,
 		return nil, err
 	}
 
-	batch.UpdateState(batchState)
+	batch.SetStateOnDBSuccess(batchState)
 
 	if len(dbBatch.TapscriptSibling) != 0 {
 		batchSibling, err := chainhash.NewHash(dbBatch.TapscriptSibling)
@@ -1472,15 +1472,22 @@ func marshalMintingBatch(ctx context.Context, q PendingAssetStore,
 
 // UpdateBatchState updates the state of a batch based on the batch key.
 func (a *AssetMintingStore) UpdateBatchState(ctx context.Context,
-	batchKey *btcec.PublicKey, newState tapgarden.BatchState) error {
+	batch *tapgarden.MintingBatch, newState tapgarden.BatchState) error {
 
+	batchKey := batch.BatchKey.PubKey
 	var writeTxOpts AssetStoreTxOptions
-	return a.db.ExecTx(ctx, &writeTxOpts, func(q PendingAssetStore) error {
+	err := a.db.ExecTx(ctx, &writeTxOpts, func(q PendingAssetStore) error {
 		return q.UpdateMintingBatchState(ctx, BatchStateUpdate{
 			RawKey:     batchKey.SerializeCompressed(),
 			BatchState: int16(newState),
 		})
 	})
+	if err != nil {
+		return err
+	}
+
+	batch.SetStateOnDBSuccess(newState)
+	return nil
 }
 
 // encodeOutpoint encodes the outpoint point in Bitcoin wire format, returning
@@ -1793,7 +1800,7 @@ func fetchSeedlingGroups(ctx context.Context, q PendingAssetStore,
 // binds the genesis transaction (which will create the set of assets in the
 // batch) to the batch itself.
 func (a *AssetMintingStore) AddSproutsToBatch(ctx context.Context,
-	batchKey *btcec.PublicKey,
+	batch *tapgarden.MintingBatch,
 	genesisPacket *tapgarden.FundedMintAnchorPsbt,
 	assetRoot *commitment.TapCommitment) error {
 
@@ -1824,10 +1831,11 @@ func (a *AssetMintingStore) AddSproutsToBatch(ctx context.Context,
 		return err
 	}
 
+	batchKey := batch.BatchKey.PubKey
 	rawBatchKey := batchKey.SerializeCompressed()
 
 	var writeTxOpts AssetStoreTxOptions
-	return a.db.ExecTx(ctx, &writeTxOpts, func(q PendingAssetStore) error {
+	err = a.db.ExecTx(ctx, &writeTxOpts, func(q PendingAssetStore) error {
 		// Upsert the assets with genesis.
 		_, _, err := upsertAssetsWithGenesis(
 			ctx, q, genesisOutpoint, sortedAssets, nil,
@@ -1852,6 +1860,12 @@ func (a *AssetMintingStore) AddSproutsToBatch(ctx context.Context,
 			BatchState: int16(tapgarden.BatchStateCommitted),
 		})
 	})
+	if err != nil {
+		return err
+	}
+
+	batch.SetStateOnDBSuccess(tapgarden.BatchStateCommitted)
+	return nil
 }
 
 // CommitSignedGenesisTx binds a fully signed genesis transaction to a pending
@@ -1863,9 +1877,11 @@ func (a *AssetMintingStore) AddSproutsToBatch(ctx context.Context,
 // TODO(roasbeef): or could just re-read assets from disk and set the script
 // root manually?
 func (a *AssetMintingStore) CommitSignedGenesisTx(ctx context.Context,
-	batchKey *btcec.PublicKey, genesisPkt *tapsend.FundedPsbt,
+	batch *tapgarden.MintingBatch, genesisPkt *tapsend.FundedPsbt,
 	anchorOutputIndex uint32, merkleRoot, tapTreeRoot []byte,
 	tapSibling []byte) error {
+
+	batchKey := batch.BatchKey.PubKey
 
 	// The managed UTXO we'll insert only contains the raw tx of the
 	// genesis packet, so we'll extract that now.
@@ -1902,7 +1918,7 @@ func (a *AssetMintingStore) CommitSignedGenesisTx(ctx context.Context,
 	}
 
 	var writeTxOpts AssetStoreTxOptions
-	return a.db.ExecTx(ctx, &writeTxOpts, func(q PendingAssetStore) error {
+	err = a.db.ExecTx(ctx, &writeTxOpts, func(q PendingAssetStore) error {
 		// First, we'll update the genesis packet stored as part of the
 		// batch, as this packet is now fully signed.
 		pktBytes, err := fn.Serialize(genesisPkt.Pkt)
@@ -1977,19 +1993,26 @@ func (a *AssetMintingStore) CommitSignedGenesisTx(ctx context.Context,
 			BatchState: int16(tapgarden.BatchStateBroadcast),
 		})
 	})
+	if err != nil {
+		return err
+	}
+
+	batch.SetStateOnDBSuccess(tapgarden.BatchStateBroadcast)
+	return nil
 }
 
 // MarkBatchConfirmed stores final confirmation information for a batch on
 // disk.
 func (a *AssetMintingStore) MarkBatchConfirmed(ctx context.Context,
-	batchKey *btcec.PublicKey, blockHash *chainhash.Hash,
+	batch *tapgarden.MintingBatch, blockHash *chainhash.Hash,
 	blockHeight uint32, txIndex uint32,
 	mintingProofs proof.AssetBlobs) error {
 
+	batchKey := batch.BatchKey.PubKey
 	rawBatchKey := batchKey.SerializeCompressed()
 
 	var writeTxOpts AssetStoreTxOptions
-	return a.db.ExecTx(ctx, &writeTxOpts, func(q PendingAssetStore) error {
+	err := a.db.ExecTx(ctx, &writeTxOpts, func(q PendingAssetStore) error {
 		// First, we'll update the state of the target batch to reflect
 		// that the batch is fully finalized.
 		err := q.UpdateMintingBatchState(ctx, BatchStateUpdate{
@@ -2044,6 +2067,12 @@ func (a *AssetMintingStore) MarkBatchConfirmed(ctx context.Context,
 		}
 		return nil
 	})
+	if err != nil {
+		return err
+	}
+
+	batch.SetStateOnDBSuccess(tapgarden.BatchStateConfirmed)
+	return nil
 }
 
 // FetchGroupByGenesis fetches the asset group created by the genesis referenced
--- a/tapdb/asset_minting.go
+++ b/tapdb/asset_minting.go
@@ -476,6 +476,19 @@ func (a *AssetMintingStore) CommitMintingBatch(ctx context.Context,
 
 	rawBatchKey := newBatch.BatchKey.PubKey.SerializeCompressed()
 
+	// Set when the batch insert itself fails on a unique constraint
+	// violation, which means the partial unique index from migration
+	// 000061 rejected a second pre-broadcast batch. The transaction
+	// executor re-maps errors returned from the closure, so the
+	// classification cannot ride along on the error itself.
+	//
+	// This attributes any unique violation on the insert to the
+	// singleton index. The only other unique constraint reachable
+	// from it is the batch_id primary key, which would require two
+	// batches sharing a batch key; keys are freshly derived per
+	// batch, so that case is not distinguished here.
+	var singletonViolation bool
+
 	var writeTxOpts AssetStoreTxOptions
 	err := a.db.ExecTx(ctx, &writeTxOpts, func(q PendingAssetStore) error {
 		// First, we'll need to insert a new internal key which'll act
@@ -497,6 +510,11 @@ func (a *AssetMintingStore) CommitMintingBatch(ctx context.Context,
 			CreationTimeUnix:    newBatch.CreationTime.UTC(),
 			UniverseCommitments: newBatch.SupplyCommitments,
 		}); err != nil {
+			var uniqueErr *ErrSqlUniqueConstraintViolation
+			if errors.As(MapSQLError(err), &uniqueErr) {
+				singletonViolation = true
+			}
+
 			return fmt.Errorf("unable to insert minting "+
 				"batch: %w", err)
 		}
@@ -617,6 +635,12 @@ func (a *AssetMintingStore) CommitMintingBatch(ctx context.Context,
 
 		return nil
 	})
+	if err != nil && singletonViolation {
+		// Surface the singleton violation as a domain error
+		// instead of the raw SQL constraint error.
+		return fmt.Errorf("%w: %w",
+			tapgarden.ErrDuplicatePreBroadcastBatch, err)
+	}
 
 	return err
 }
--- a/tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.down.sql
+++ b/tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.down.sql
@@ -1,0 +1,1 @@
+DROP INDEX IF EXISTS asset_minting_batches_unique_pending_or_frozen;
--- a/tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
+++ b/tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
@@ -1,0 +1,21 @@
+-- Enforce that at most one minting batch is in a pre-broadcast state
+-- (BatchStatePending = 0 or BatchStateFrozen = 1) at any time. This
+-- matches the planter's in-memory model, which assumes a single
+-- "current" pre-broadcast batch in flight. Older versions of the
+-- planter could in some failure scenarios desync the in-memory slot
+-- from disk and leave two pre-broadcast rows behind; this index
+-- makes that state unrepresentable going forward.
+--
+-- The unique index targets the constant expression `(1)`, which
+-- means "at most one row total" among rows matching the WHERE
+-- filter. The syntax is dialect-agnostic across SQLite and
+-- Postgres.
+--
+-- This is a deliberate design choice rather than a fundamental
+-- limit. If multi-batch support is ever added to the planter,
+-- drop this index and revisit the planter API and recovery loop
+-- alongside that change.
+CREATE UNIQUE INDEX IF NOT EXISTS
+    asset_minting_batches_unique_pending_or_frozen
+    ON asset_minting_batches ((1))
+    WHERE batch_state IN (0, 1);
--- a/tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
+++ b/tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
@@ -1,0 +1,44 @@
+-- Enforce that at most one minting batch is in a pre-broadcast state
+-- (BatchStatePending = 0 or BatchStateFrozen = 1) at any time. This
+-- matches the planter's in-memory model, which assumes a single
+-- "current" pre-broadcast batch in flight. Older versions of the
+-- planter could in some failure scenarios desync the in-memory slot
+-- from disk and leave two pre-broadcast rows behind; this migration
+-- makes that state unrepresentable going forward.
+--
+-- Legacy databases may already violate the invariant. Rather than
+-- fail here with an opaque unique-index error and force the operator
+-- into a manual repair step, cancel all but the most recent
+-- pre-broadcast batch first. The preserved row is the one with the
+-- latest creation_time_unix, tie-broken by the raw batch key, so the
+-- outcome is deterministic on identical timestamps and matches the
+-- ordering used by the tapd repair tool
+-- (--repair.cancel-duplicate-batches), which cannot see batch_id.
+-- Cancelled rows move to BatchStateSeedlingCancelled = 6, which
+-- leaves them and their seedlings on disk for later inspection.
+UPDATE asset_minting_batches
+SET batch_state = 6
+WHERE batch_state IN (0, 1)
+  AND batch_id NOT IN (
+      SELECT b.batch_id
+      FROM asset_minting_batches b
+      JOIN internal_keys k
+          ON k.key_id = b.batch_id
+      WHERE b.batch_state IN (0, 1)
+      ORDER BY b.creation_time_unix DESC, k.raw_key DESC
+      LIMIT 1
+  );
+
+-- The unique index targets the constant expression `(1)`, which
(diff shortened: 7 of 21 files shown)


jtobin:
A note on the singleton invariant: that invariant was previously assumed, but only implicitly, so the refactoring makes it explicit. It does not preclude extending to non-singletons in the future (it extends gracefully to such).

tapgarden/batch.go
```diff
@@ -499,38 +516,107 @@ func (m *MintingBatch) validateUniCommitment(newSeedling Seedling) error {
...
+	switch count {
+	case 0:
+		return nil, fmt.Errorf("no group anchor seedling in batch")
+	case 1:
+		return anchor, nil
+	default:
+		return nil, fmt.Errorf("batch has %d group anchor "+
+			"seedlings, expected exactly 1", count)
+	}
+}
```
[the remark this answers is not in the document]
jtobin: Hallucination.

tapgarden/planter.go
```diff
@@ -2116,6 +2173,36 @@ func (c *ChainPlanter) gardener() {
...
+						cancelState =
+							BatchStateSproutCancelled //nolint:lll
+					}
+
+					cancelCtx, cancelCtxCancel :=
+						c.WithCtxQuit()
+					cancelErr := c.cfg.Log.UpdateBatchState(
+						cancelCtx, c.pendingBatch,
+						cancelState,
+					)
```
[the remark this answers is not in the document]
jtobin: Hallucination.

tapgarden/planter.go
```diff
@@ -910,28 +918,19 @@ func fetchPreCommitGroupKey(
...
 	// If the group info is unset, then there is no pre-commitment group pub
 	// key defined in the batch.
-	if groupAnchorSeedling.GroupInfo == nil {
+	if anchor.GroupInfo == nil {
 		return zero, nil
 	}
 
-	return fn.Some(groupAnchorSeedling.GroupInfo.GroupPubKey), nil
+	return fn.Some(anchor.GroupInfo.GroupPubKey), nil
 }
```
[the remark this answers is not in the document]
jtobin: Chained from previous hallucination.

tapgarden/planter.go
```diff
@@ -863,26 +884,13 @@ func fetchDelegationKey(pendingBatch *MintingBatch) (fn.Option[DelegationKey],
...
 	}
 
-	delegationKeyDesc := fn.MapOptionZ(groupAnchorSeedling,
-		func(s Seedling) fn.Option[keychain.KeyDescriptor] {
-			return s.DelegationKey
-		},
-	)
-
-	return delegationKeyDesc, nil
+	return anchor.DelegationKey, nil
```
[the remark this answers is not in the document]
jtobin: Chained from previous hallucination.

tapdb/asset_minting.go
```diff
@@ -1472,15 +1472,22 @@ func marshalMintingBatch(ctx context.Context, q PendingAssetStore,
 
 // UpdateBatchState updates the state of a batch based on the batch key.
 func (a *AssetMintingStore) UpdateBatchState(ctx context.Context,
-	batchKey *btcec.PublicKey, newState tapgarden.BatchState) error {
+	batch *tapgarden.MintingBatch, newState tapgarden.BatchState) error {
 
+	batchKey := batch.BatchKey.PubKey
```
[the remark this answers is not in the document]
jtobin: Relevant stuff can't be nil.

tapcfg/repair.go
```diff
@@ -0,0 +1,151 @@
...
+		sqliteCfg := *cfg.Sqlite
+		sqliteCfg.SkipMigrations = true
+		cfgLogger.Infof("repair: opening sqlite3 database at %v "+
+			"(migrations skipped)",
+			sqliteCfg.DatabaseFileName)
+		db, err = tapdb.NewSqliteStore(&sqliteCfg)
+
+	case DatabaseBackendPostgres:
+		pgCfg := *cfg.Postgres
+		pgCfg.SkipMigrations = true
```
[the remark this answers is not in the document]
jtobin: Can't be nil.

tapgarden/planter.go
```diff
@@ -1430,10 +1429,53 @@ func freezeMintingBatch(ctx context.Context, batchStore MintingStore,
...
+					batch.BatchKey.PubKey.
+						SerializeCompressed(),
+				),
+			)
+
+		default:
+			// Only pre-broadcast states are constrained by
+			// the singleton index; ignore everything else.
+		}
+	}
```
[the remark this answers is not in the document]
jtobin: Ditto.

litbot-9000:
@georgetsagk: review reminder

tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
```diff
@@ -0,0 +1,21 @@
...
+-- Postgres.
+--
+-- This is a deliberate design choice rather than a fundamental
+-- limit. If multi-batch support is ever added to the planter,
+-- drop this index and revisit the planter API and recovery loop
+-- alongside that change.
+CREATE UNIQUE INDEX IF NOT EXISTS
+    asset_minting_batches_unique_pending_or_frozen
+    ON asset_minting_batches ((1))
+    WHERE batch_state IN (0, 1);
```
litbot-9000, line 44: **Blocker: on a database that already violates this invariant, the failed migration bricks the daemon permanently — the repair path in this PR cannot recover it.**

The commit message accepts that legacy DBs with duplicates "fail on startup with the migration error", and treats `--repair.cancel-duplicate-batches` as the remedy. But `golang-migrate` records the *attempted* version and sets `dirty = true` when a step fails, and `applyMigrations` refuses to proceed on a dirty DB (`tapdb/migrations.go:180`). The repair tool fixes the **data** and never touches the **dirty flag**, so the restart it tells the operator to perform can never succeed.

I reproduced the whole operator sequence through the real `NewSqliteStore` constructor:

```
STEP1 normal tapd start on legacy DB (v60, two batch_state=0 rows):
  err = error executing migrations: constraint failed: UNIQUE constraint failed:
        index 'asset_minting_batches_unique_pending_or_frozen' (2067)
  -> schema_migrations = (version 61, dirty = true)

STEP2 tapd --repair.cancel-duplicate-batches (SkipMigrations=true):
  opens fine, cancels the duplicate, pre-broadcast count back to 1

STEP3 restart tapd normally, exactly as checkSingletonInvariant's error text instructs:
  err = error executing migrations: database is in a dirty state at version 61,
        manual intervention required
  VERDICT: tapd STILL WILL NOT START after the documented repair
```

Failure scenario, concretely: operator has two `batch_state = 0` rows on disk (the exact desync this PR exists to make unrepresentable), upgrades tapd, starts it. Migration 61 fails. They follow the instructions in `checkSingletonInvariant`'s error message. `tapd` still refuses to start, now with a *different* error that suggests no remedy at all. Recovery requires an undocumented manual `UPDATE schema_migrations SET version = 60, dirty = false` — i.e. hand-editing the migration bookkeeping table of a production DB.

Two things make it worse:

- `backupAndMigrate` (`tapdb/sqlite.go`) only *copies* a backup file before migrating; it does **not** restore on failure. So the live DB is left dirty, and with `SkipMigrationDbBackup = true` there is no backup at all.
- The friendly diagnostic that names the repair flag (`planter.go:1449`) is **unreachable in this scenario**. DB open and migration happen long before `ChainPlanter.Start()`, so the operator only ever sees the raw SQL constraint error, which mentions no remedy. The nice error only fires for the much rarer "duplicates appeared after a successful migration" case.

Two fix shapes, either of which would close this:

1. Make the repair tool force the version back to 60 / clear `dirty` after it repairs, and make the migration-failure error mention the flag.
2. Better — make this a **programmatic migration** that cancels the duplicates and creates the index in one transaction, so the invariant becomes self-establishing and no operator step is needed. The repo already has that mechanism (`makeProgrammaticMigrations`, cf. versions 50/51), and `tapcfg/repair.go:27` already *claims* this is how it works.

Separately, and much smaller: none of the new tests cover the migration against pre-existing duplicate rows. A test at `NewTestDBWithVersion(t, 60)` + two `batch_state = 0` rows + `ExecuteMigrations(TargetLatest)` would have caught this.

tapgarden/planter.go
```diff
@@ -2116,6 +2173,36 @@ func (c *ChainPlanter) gardener() {
...
+					// SeedlingCancelled (no sprouts yet),
+					// else SproutCancelled.
+					cancelState :=
+						BatchStateSeedlingCancelled
+					if c.pendingBatch.State() ==
+						BatchStateCommitted {
+
+						cancelState =
+							BatchStateSproutCancelled //nolint:lll
+					}
```
litbot-9000: **This can mark an already-broadcast batch as `SeedlingCancelled`, after its genesis transaction is in the mempool.**

To be clear up front: this is *not* the bot's comment on these lines. Its claim (nil panic / different batch) is wrong — the guard at `:2126` makes `c.pendingBatch` non-nil, and `finalizeBatch` assigns `c.pendingBatch = sealedBatch` and then passes that same pointer to `newCaretakerForBatch`, so `c.pendingBatch` *is* the failed caretaker's batch. That pointer identity is exactly what makes the problem below reachable rather than hypothetical.

Only `BatchStateCommitted` is special-cased; every other state falls through to `SeedlingCancelled`. But `BroadcastErrChan` fires whenever `advanceStateUntil(_, BatchStateBroadcast)` returns an error (`caretaker.go:369`), and that loop **runs the `BatchStateBroadcast` branch too** — it only terminates when `nextState == currentState`. So there are two reachable error paths where the batch is already at `Broadcast` in memory *and* on disk:

1. `caretaker.go:824` — `ImportTaprootOutput` fails with anything other than "already exists", *after* `CommitSignedGenesisTx` (`:805`) has already committed `state = Broadcast`.
2. `caretaker.go:883` — `RegisterConfirmationsNtfn` fails *after* `caretaker.go:865` `PublishTransaction` **succeeded**.

Case 2 is the damaging one. Concrete scenario: batch reaches `Broadcast`, the genesis tx is successfully published, then lnd's chain notifier is briefly unavailable so `RegisterConfirmationsNtfn` returns an error. This block then writes `batch_state = SeedlingCancelled (6)`. The genesis transaction confirms anyway and the assets are minted on chain — but on restart the batch is skipped as cancelled (`planter.go:534`), `MarkBatchConfirmed` never runs, and no minting proofs are ever stored. The assets exist on chain and tapd has recorded the batch as "seedlings never sprouted".

That is a regression: before this PR the row stayed at `Broadcast`, and a restart resumed it correctly through `FetchNonFinalBatches` and re-broadcast — which is precisely what the `BatchStateBroadcast` branch's "already been rebroadcast, so we'll attempt to re-broadcast it" comment is designed for.

It also contradicts a judgement already made elsewhere in the tree: `caretaker.Cancel()` (`caretaker.go:241`) deliberately **refuses** to cancel anything past `Committed` ("batch not cancellable"). This new block bypasses that.

Note the stated motivation doesn't extend this far either — the comment says the goal is to avoid being "left wedged in a pre-broadcast state, which the migration 000061 singleton index forbids", but the index is `WHERE batch_state IN (0, 1)`, i.e. `{Pending, Frozen}` only. Cancelling `Committed` or `Broadcast` batches is not needed for the invariant. Suggest restricting this to `{Pending, Frozen}` → `SeedlingCancelled` and leaving anything at or past `Committed` alone (or at minimum mirroring `Cancel()`'s `Committed` → `SproutCancelled`, refuse beyond).

tapcfg/repair.go
```diff
@@ -0,0 +1,151 @@
...
+// in migration 000060, and cancels all but the most recent. The
+// preserved batch is the one with the latest CreationTime; cancelled
+// batches transition to BatchStateSeedlingCancelled, leaving their
+// row and seedlings on disk for later inspection.
+//
+// The function opens the database with migrations skipped, so it can
+// run against a legacy database whose state would otherwise fail the
+// migration.
+//
+// NOTE: With migration 000061's self-heal in place, restarting tapd
```
litbot-9000, line 29: This NOTE describes behaviour that does not exist. `000061_unique_pending_or_frozen_batch.up.sql` is a bare `CREATE UNIQUE INDEX IF NOT EXISTS ...` — there is no `DELETE`, no `UPDATE`, no programmatic migration hook, and nothing that cancels duplicates. Restarting tapd normally does **not** "cancel the duplicates as part of applying the migration"; the migration fails and (see my comment on the migration file) leaves the DB dirty.

This matters more than a stale comment normally would, for two reasons. First, it is the only place in the tree that describes the operator recovery story, and it describes it wrongly in the safe direction — a reader concludes no manual step is needed, when in fact a manual step is required *and* insufficient. Second, it directly contradicts the commit message for e1fd2eeb ("Legacy databases that already hold duplicates fail on startup with the migration error"), so the two disagree about the single most consequential behaviour in this commit.

Either implement the self-heal (which would be the better fix for the blocker, and would make this comment true), or delete the NOTE and state that the tool is the *only* recovery route.

Also in this doc comment: lines 18 and 50 say "migration 000060", which is an existing unrelated migration (`asset_transfers_superseded`). The one this file is about is 000061.

tapgarden/planter.go
```diff
@@ -2116,6 +2173,36 @@ func (c *ChainPlanter) gardener() {
...
+
+					cancelCtx, cancelCtxCancel :=
+						c.WithCtxQuit()
+					cancelErr := c.cfg.Log.UpdateBatchState(
+						cancelCtx, c.pendingBatch,
+						cancelState,
+					)
+					cancelCtxCancel()
+					if cancelErr != nil {
```
litbot-9000: Worth reconsidering now that the index exists: this error is only logged, but its consequence has changed from "untidy" to "minting is dead until restart".

If this `UpdateBatchState` fails (the DB being unavailable is exactly the failure mode this PR is about), the row stays at `Pending`/`Frozen` while `c.pendingBatch` is set to `nil` at `:2213`. The in-memory slot is now empty but the on-disk singleton slot is still occupied — the desync this PR exists to prevent, reintroduced from the other direction. Every subsequent `CommitMintingBatch` then trips the new partial unique index, so **all** further minting fails with an opaque SQL constraint error until the daemon restarts.

Before this commit the same failure just left a stale pre-broadcast row that a restart would pick up. Now it's a wedge. Since you have a real error in hand here, consider surfacing it to the caller (the `req.Error(err)` at `:2163` has already happened, but the message could be augmented) or at least logging it at error level with the "restart required" consequence spelled out.

tapgarden/interface.go
```diff
@@ -191,15 +191,23 @@ type MintingStore interface {
...
+	UpdateBatchState(ctx context.Context, batch *MintingBatch,
 		newState BatchState) error
 
 	// AddSeedlingsToBatch adds a new seedling to an existing batch. Once
 	// added this batch should remain in the BatchStatePending state.
 	//
-	// TODO(roasbeef): assumption that only one pending batch at a time?
+	// The "exactly one batch in BatchStatePending or BatchStateFrozen
+	// at a time" invariant the planter relies on is enforced at the
+	// DB layer by migration 000060 (a partial unique index on
```
litbot-9000: Minor, but it's wrong in 9 places and points at a real unrelated migration, so worth one sweep: this says `000060`, but the migration added here is `000061` (`000060` is the existing `asset_transfers_superseded`).

Same stale number at `tapgarden/planter.go:516`, `:1439`, `:1844`, `:1866`, `tapgarden/planter_test.go:1634`, `:2171`, `tapcfg/repair.go:18`, `:50` — and in the PR description ("Migration 000060 adds a partial unique index"). `planter.go:2178` and `tapcfg/config.go:377` have it right.

Normally a comment nit isn't worth raising, but this number is the operator's handle on the recovery procedure, so having most references point at the wrong migration is mildly hazardous rather than merely untidy.

litbot-9000 (changes requested):
## Verdict

The invariant work itself is sound: I tried to falsify all four claims in the description and three of them held up under direct attack (details below). But I found one blocker that is not about Go code at all — **migration 000061 leaves the database `dirty`, and the recovery path this PR ships cannot clear it**, so an operator whose DB already holds two pre-broadcast batches ends up with a `tapd` that will not start even after running `--repair.cancel-duplicate-batches` exactly as instructed. I reproduced that end-to-end. There is also a real regression in the new caretaker-failure cancellation block: it can mark an **already-broadcast** batch as `SeedlingCancelled`. Hence `REQUEST_CHANGES` — on those two items only; the rest of the PR I'd be happy to see merged.

## What I verified

Commands actually run at `cec1069d`, on SQLite (modernc):

| command | result |
|---|---|
| `make build` | **PASS** (`tapd-debug`/`tapcli-debug`, `v0.7.0-808-gcec1069d`) |
| `go test ./tapgarden/...` | **PASS** (31.5s) |
| `go test ./tapdb/...` | **PASS** (51.9s) |
| `go test ./tapdb/ -run 'TestMigrationSteps\|TestMigrationDowngrade' -v` | **PASS** — log confirms `61/u unique_pending_or_frozen_batch`, `Database version after migration: 61` |
| `go test -race ./tapgarden/` | **PASS** (75.8s) |
| Postgres-tagged paths | **NOT RUN** — no Postgres available in this environment and no `psql` binary. The `ON asset_minting_batches ((1)) WHERE ...` expression-index idiom is therefore verified on SQLite only; someone should confirm the Postgres dialect path in CI before merge. |

Plus three throwaway tests I wrote to answer "what happens to real data?" empirically (removed again; tree left clean) — see the migration comment for the transcript.

Claims I tried to falsify and could **not** break:

- **(a) two-truth collapse — holds.** All four store methods (`asset_minting.go:1480/1858/1991/2018`) run `ExecTx` → `if err != nil { return err }` → `SetStateOnDBSuccess(...)`; no success path skips the mirror and no error path performs it. I grepped exhaustively for in-memory mutation: the only non-store writers left are `planter.go:731` (construction, batch still private), `batch.go:110` (`Copy`), and `caretaker.go:1374/1387` — and those last two mutate a **detached `b.Copy()`** for event publication, so they cannot desync anything. I also checked that dropping `advanceStateUntil`'s blanket mirror is safe: every `stateStep` branch that *changes* state persists via a store call first (`freezeMintingBatch`, `AddSproutsToBatch`, `CommitSignedGenesisTx`, `MarkBatchConfirmed`, `UpdateBatchState`); the `Broadcast` branch returns its own state and writes nothing, which is correct. No production code writes `UpdateMintingBatchState` outside the store.
- **(b) fundBatch split — holds.** `fundBatch` is gone (the only remaining hits are an unrelated test-harness method of the same name). Signatures are genuinely disjoint and `Start()`'s recovery path calls `applyFundingToBatch` directly. I specifically checked the one behaviour change that could have hidden here — `validateSeedling` now runs `validateDelegationKey` *before* the `m.SupplyCommitments` assignment that used to precede it — and the reordering is safe: `validateDelegationKey` reads only `newSeedling.SupplyCommitments` and iterates `m.Seedlings`, never `m.SupplyCommitments`. The comment claiming order-independence is accurate.
- **(d) deterministic anchor — holds.** With exactly one match the scan is order-independent, and I confirmed exactly one match is guaranteed: both call sites are gated on `SupplyCommitments` (`planter.go:1029`, and the early return at `:912`), and `validateUniCommitment:483` forces every seedling after the first in a supply-commit batch to have a non-nil `GroupAnchor`.
- **Concurrency — no finding.** `batchState` is an `atomic.Uint32`, so even though `caretaker.Cancel()` now mutates in-memory state from the confirmation goroutine (`caretaker.go:925/966`) where it previously only wrote the DB, there is no race. `-race` agrees.
- **Test coverage — adequate, contrary to what I expected to find.** `TestUpdateBatchStateMemoryCoherence` is a real memory/disk-drift test (pre-cancelled ctx, asserts *both* stay at `Pending`), and `TestSingletonPreBroadcastBatchConstraint` is a real negative test for the constraint including the `Pending ∪ Frozen` and "Committed doesn't count" cases. The one gap is that nothing tests the migration against pre-existing duplicate rows — which is exactly where the blocker lives.

**(c) is the one that breaks** — the schema does forbid the state, but the transition into that schema is not survivable on data that already violates it.

## Findings

1. `tapdb/sqlc/migrations/000061...up.sql:18` — **blocker**: failed migration leaves the DB dirty; the shipped repair path cannot recover it. Empirical transcript inline.
2. `tapgarden/planter.go:2183` — cancels an already-`Broadcast` batch as `SeedlingCancelled`, potentially orphaning assets that mint on chain. *(Different defect from the bot's comment on this line — see inline.)*
3. `tapcfg/repair.go:27` — comment asserts a migration "self-heal" that does not exist, and the tool never clears the dirty flag.
4. `tapgarden/planter.go:2199` — swallowed cancel error now blocks all future minting rather than just being untidy.
5. `tapgarden/interface.go:207` — stale migration number `000060` in 9 places (minor).

## On the existing bot review

I read all 7 `gemini-code-assist` comments and your dismissals before writing this, and I independently reached your conclusion on the substantive one: the `uniqueAnchorSeedling` "breaks multi-asset non-grouped minting" claim is wrong, because neither call site is reachable for a non-supply-commitment batch. Your "Hallucination." is correct and its proposed `return nil, nil` fix would have introduced the nil-deref its own follow-ups then tried to patch. I have not re-litigated any of the nil-check comments. The one place I do overlap is `planter.go:2197`: the bot's *reasoning* there is wrong (the `:2126` guard makes `c.pendingBatch` non-nil, and `finalizeBatch` makes it the same pointer the caretaker holds), but there is a separate, real defect on those lines that its comment did not identify — flagged inline so it doesn't get closed along with the bot's version.

_Posted by claudell (automated review agent) at roasbeef's request. Findings are unverified by a human._

tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
jtobin: This is handled by a later commit in the intended series, but I'll bring the relevant parts into this changeset.

jtobin: Fixed in updated changeset.

tapcfg/repair.go
jtobin: Fixed in updated changeset. Fable's summary:

The NOTE was written against a later commit in the stack that hadn't yet been pulled into this PR. That self-heal now exists in migration 000061, so the NOTE is accurate as written. The stale 000060 references here are fixed as well.

tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
jtobin: (Fable's summary of the change.)

Migration 000061 now cancels all but the most recent pre-broadcast batch inside the same migration transaction, before creating the index — the invariant is self-establishing and the migration cannot fail on legacy duplicates, so the dirty state is unreachable. This is the self-heal from a later commit in the planned stack, now absorbed here. Added TestMigration61CancelsDuplicatePreBroadcastBatches (v60 DB + duplicate pre-broadcast rows → migrate → single surviving row, correct survivor, rest cancelled); passes on SQLite and Postgres. The repair flag remains as a diagnostic for duplicates appearing on an already-migrated DB.

tapgarden/planter.go
jtobin: Fixed in latest changeset. Fable's summary:

The handler now applies caretaker.Cancel()'s rule exactly: Pending/Frozen → SeedlingCancelled, Committed → SproutCancelled, and anything at or past Broadcast is left untouched so the restart path resumes it. Extracted into a helper (cancelFailedBatch) for legibility.

tapgarden/planter.go
jtobin: Fixed in updated changeset. Fable's summary:

Now logged at error level with the consequence spelled out: the pre-broadcast singleton slot remains occupied on disk and new batches are rejected until restart.

tapgarden/interface.go
jtobin: Fixed in latest changeset.

jtobin:
Fable's summary of the fixups:

The blocker is fixed by making the migration self-establishing: the self-heal cancellation previously lived further up the hardening stack and has now been pulled into this PR, so a legacy DB with duplicates migrates cleanly and the dirty-flag scenario can't arise.  The cancellation block now leaves anything at or past Broadcast alone.  Also rebased onto main. On the Postgres caveat: the migration tests, including the new duplicate-rows test, pass under test_db_postgres locally.

tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.down.sql
```diff
@@ -0,0 +1 @@
+DROP INDEX IF EXISTS asset_minting_batches_unique_pending_or_frozen;
```
litbot-9000, line 1: Minor, non-blocking: the down migration is not a true inverse any more, and nothing says so.

`000061` now performs an irreversible data mutation (cancelling every pre-broadcast batch but the newest), but the down migration only drops the index. I verified this empirically — up, then `TargetVersion(60)`, then up again: the cancelled rows stay at `batch_state = 6` throughout.

Concrete scenario: an operator upgrades, the migration auto-cancels a Pending batch that had seedlings queued, they then hit an unrelated regression and downgrade tapd to a v60-era release. The index is gone but the batch is still `SeedlingCancelled`, so the older daemon skips it at `Start()` forever, and nothing in the tree tells them a downgrade will not bring it back.

Restoring the prior state automatically is not really possible (the migration does not record what it changed), so the fix is a sentence rather than SQL: the up migration's comment block already says cancelled rows are left on disk for inspection — worth adding that a downgrade will not restore them, so an operator knows a manual `UPDATE ... SET batch_state = 0` is the recovery. `TestMigrationDowngrade` does not cover this: it only asserts `ErrMigrationDowngrade` when the DB is newer than the binary, it never runs `61.down.sql`.

tapcfg/repair.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	if len(preBroadcast) <= 1 {
+		cfgLogger.Infof("repair: nothing to do; found %d batches "+
+			"in pre-broadcast state", len(preBroadcast))
+		return nil
+	}
+
+	// Sort newest-first by CreationTime; preserve [0], cancel the
+	// rest. SliceStable gives a deterministic winner when two
```
litbot-9000: Minor, non-blocking: the determinism claim in this comment does not hold, and it disagrees with the rule the migration now uses.

`sort.SliceStable` preserves *input* order on equal keys, and the input here is `FetchNonFinalBatches` → `FetchMintingBatchesByInverseState`, whose query (`tapdb/sqlc/queries/assets.sql:16-21`) has no `ORDER BY`. Row order from an unordered `SELECT` is not a stable property — it can change on Postgres after a `VACUUM`/plan change. So on two batches sharing a `CreationTime` the surviving batch is whatever the storage engine hands back, and re-running the tool on a restored copy of the same DB can preserve a different batch.

Separately, migration `000061` breaks the same tie with `batch_id DESC`, so the two code paths that implement "preserve the most recent" can pick different survivors for identical input.

The practical impact is close to nil — while the partial unique index exists this tool can never see more than one pre-broadcast batch, and either survivor is defensible — so this is a comment-accuracy issue more than a bug. Cheapest fix is to match the migration: tie-break on something stable, e.g. compare the serialized batch key when `CreationTime` is equal, and reword the comment accordingly.

litbot-9000 (commented):
## Re-review verdict

**Both blockers are cleared.** I re-ran the original failing scenario end-to-end and it no longer reproduces, and the broadcast-cancel state bug is genuinely fixed rather than papered over. Nothing in this round blocks merge, so this is a `COMMENT` rather than `REQUEST_CHANGES`. Two minor, non-blocking observations are left inline.

I did not take the "Fixed in updated changeset" summaries at face value — the migration went from a bare `CREATE UNIQUE INDEX` to DML that mutates live operator data, which is a much larger blast radius than what it replaced and had never been reviewed. Details below.

---

## Per-finding resolution

**1. Migration 000061 bricks a legacy DB — FIXED (blocker cleared).**
Rebuilt the original repro: a v60 SQLite DB holding two `batch_state = 0` rows, opened through the real `NewSqliteStore(SkipMigrations: false)` path. Previously this returned `constraint failed: UNIQUE constraint failed: index 'asset_minting_batches_unique_pending_or_frozen' (2067)` and left `schema_migrations = (61, dirty=true)`, after which `applyMigrations` refused every subsequent start. Now the daemon starts, `schema_migrations` is `version=61 dirty=false`, and the older row is at state 6 with the newer one preserved at state 0. Also checked the properties the new DML has to get right:
* *Survivor selection* — 5 pre-broadcast rows mixed Pending/Frozen with descending timestamps: exactly the newest survives, the other four move to 6.
* *Tie determinism* — 3 rows with byte-identical `creation_time_unix`: `batch_id DESC` breaks the tie, highest id survives, reproducible across runs. The claim in the comment holds.
* *Correct terminal state* — state 6 is `BatchStateSeedlingCancelled`, which is exactly what `caretaker.Cancel()` writes for a Pending/Frozen batch (`tapgarden/caretaker.go:199-218`). Committed batches would need `SproutCancelled`, but they are outside the `WHERE batch_state IN (0, 1)` filter, so the migration can never write the wrong state for a batch shape.
* *Atomicity* — genuinely one transaction on both dialects, not by luck: the sqlite driver wraps `Run()` in `db.Begin()/Commit()` (`NoTxWrap` is false with the empty `Config{}` passed at `tapdb/sqlite.go:250`), and the postgres driver executes the whole file in a single `ExecContext` with `MultiStatementEnabled` false, which is one implicit transaction. A partial application is not reachable.
* *Collateral damage* — the `UPDATE` touches only `batch_state`. `asset_seedlings`, `mint_anchor_uni_commitments` and the supply pre-commit rows all keep their FKs and rows, which is the same on-disk shape a normal `caretaker.Cancel()` produces. The Go cancel paths also do not release lnd output leases, so the migration is not worse than the code path it mimics.
* *Dialect portability* — single shared migration directory, and `postgresSchemaReplacements` only rewrites `BLOB` / `INTEGER PRIMARY KEY` / `TIMESTAMP` / `UNHEX`, none of which appear in this file, so no text mangling. `NOT IN (SELECT ... ORDER BY ... LIMIT 1)` and a constant-expression partial unique index are both valid Postgres.
* *The new test is real* — I deleted the `UPDATE` block from the migration and re-ran `TestMigration61CancelsDuplicatePreBroadcastBatches`: it fails with the 2067 constraint error at `tapdb/migrations_test.go:1490`. It is a genuine regression test, not one that passes for the wrong reason, and it asserts what its name claims (correct survivor id *and* preserved `batch_state`, plus the cancelled count).

**2. Caretaker-failure cancellation could mark an already-Broadcast batch cancelled — FIXED.**
`cancelFailedBatch` (`tapgarden/planter.go:1934`) now mirrors `caretaker.Cancel()` exactly, including its `default` arm: Pending/Frozen → `SeedlingCancelled`, Committed → `SproutCancelled`, anything at or past Broadcast returns without writing. The `RegisterConfirmationsNtfn`-fails-after-`PublishTransaction`-succeeds path can no longer write off a batch whose genesis tx is in the mempool.

I also checked the two claims the fix rests on rather than assuming them:
* *"The restart path resumes it"* is true. `FetchNonFinalBatches` is `FetchMintingBatchesByInverseState(BatchStateFinalized)` (`tapdb/asset_minting.go:1195`), so Broadcast rows are returned; `Start()` skips only the two `*Cancelled` states (`planter.go:534`), skips the fund/seal/freeze block for non-pre-broadcast batches, and unconditionally builds and starts a caretaker. The batch is picked up, not stranded.
* *No path left on the old behaviour.* Every cancel-state write in `planter.go` is `:554`, `:1915` and the helper at `:1953`. `:554` is reachable only inside the `batchState == BatchStatePending || batchState == BatchStateFrozen` branch, so `SeedlingCancelled` is correct there; `:1915` is `cancelMintingBatch`, gated by `canCancelBatch()`. The `State()` read in the helper is accurate because `finalizeBatch` hands the very same `*MintingBatch` pointer to `newCaretakerForBatch` (`planter.go:2886`/`:2898`) and the store keeps it current via `SetStateOnDBSuccess`.

**3. `tapcfg/repair.go` NOTE described a self-heal that did not exist — FIXED.** The self-heal now genuinely exists, so the NOTE is accurate as written.

**4. Swallowed cancel error — PARTIALLY FIXED (logging only), and I am not re-escalating it.**
The wedge itself is unchanged: if `UpdateBatchState` fails at `planter.go:1953` the row stays pre-broadcast while `c.pendingBatch` is cleared at `:2227`, and every later `CommitMintingBatch` trips the index until restart. Taking a position: **logging is an acceptable resolution here and this should not block.** Triggering it needs a DB write failure or shutdown at one specific instant; the new `log.Errorf` names both the consequence and the remedy; and restart genuinely recovers, since `Start()` picks the row back up, funds/seals/freezes it and assigns a caretaker. Worth noting the `CancelBatch` RPC cannot clear it (`canCancelBatch` requires a non-nil `c.pendingBatch`), so restart really is the only remedy — which is exactly what the message tells the operator.

**5. Stale `000060` references — FIXED.** `grep -rn 000060` over `*.go`/`*.md`/`*.conf` outside the migrations directory returns zero hits. All nine sites updated.

**Re-check of the three claims that already held (no rebase regression).** The exhaustive grep for `setState` / `batchState.Store` / `SetStateOnDBSuccess` in non-test `tapgarden` + `tapdb` returns the identical set as before — the only in-memory writers are batch construction, `Copy()`, the two detached event copies in `caretaker.go`, and the five `SetStateOnDBSuccess` calls that all follow a successful `ExecTx`. Nothing new on main introduced a batch-state writer that bypasses the store. `fundBatch` is still gone (only the unrelated `planter_test.go:424` harness method shares the name), and `uniqueAnchorSeedling` is unchanged with both call sites still gated on `SupplyCommitments`.

---

## What I verified

Against head `2b1c6af3` (confirmed current before posting), diffed against the new merge base `95bd8d41`, not against the old `cec1069d`.

```
make build                                        -> exit 0 (v0.7.0-859-g2b1c6af3)
go test ./tapgarden/... ./tapdb/...               -> exit 0 (tapgarden 34.4s, tapdb 61.4s)
go test ./tapdb/ -run TestMigration -v            -> 14/14 PASS, incl.
                                                     TestMigration61CancelsDuplicatePreBroadcastBatches
go test -race ./tapgarden/                        -> exit 0 (76.1s)
```

Throwaway tests I wrote and then removed (working tree left clean):
* v60 DB + 2 pre-broadcast rows through the real `NewSqliteStore` → start OK, `dirty=false`. This is the acceptance test for the blocker.
* v60 DB + 5 mixed Pending/Frozen rows → correct single survivor.
* v60 DB + 3 rows with identical timestamps → deterministic `batch_id` tie-break.
* up → down to 60 → up again → all succeed, `dirty=false` throughout.
* Mutation test: removed the `UPDATE` from the migration, confirmed the author's new test fails.

One refutation attempt I want to record because it is a dialect divergence that happens to be safe. SQLite stores `creation_time_unix` as TEXT in Go's `time.Time.String()` format (`typeof()` returns `text`, e.g. `"2026-08-12 15:56:52.977564306 +0000 UTC"`), so `ORDER BY creation_time_unix DESC` is a *string* comparison there while Postgres compares a real timestamp. I tried to break the survivor selection with four adversarial cases — zero-nanosecond vs sub-second (Go trims trailing fractional zeros), `.5` vs `.50001`, second rollover, day rollover — and all four picked the chronologically correct survivor. The format is fixed-width through the seconds field and the pad character `' '` (0x20) sorts below `'.'` (0x2E) and below digits, so lexicographic order matches chronological order; and `tapdb/asset_minting.go:497` has always written `CreationTime.UTC()`, so there are no mixed-offset legacy rows to worry about. Not a finding, but worth knowing the sqlite ordering is textual.

**What I could not verify:** there is no Docker, `psql`, or Postgres server in this environment, so I ran nothing against Postgres myself. Rather than imply coverage I do not have: I read the SQL for dialect hazards manually (see above), and I confirmed the check-runs on this exact head include `run unit tests (unit dbbackend=postgres pkg=tapdb)`, `... pkg=tapgarden`, and four `run itests postgres` shards, all green. `tapdb/migrations_test.go` is untagged and `NewTestDBWithVersion` is build-tag dispatched, so the new migration test does execute under that Postgres job.

---

## New findings

Two, both minor and non-blocking; see the inline comments.

_Posted by claudell (automated review agent) at roasbeef's request. Findings are unverified by a human._

tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
litbot-9000: Resolved — verified empirically, not taken on trust.

I rebuilt the original repro against `2b1c6af3`: a v60 SQLite DB holding two `batch_state = 0` rows, opened through the real `NewSqliteStore(SkipMigrations: false)` path. Before, that returned `constraint failed: UNIQUE constraint failed: index 'asset_minting_batches_unique_pending_or_frozen' (2067)` and left `schema_migrations = (61, dirty=true)`, after which every subsequent start was refused at `tapdb/migrations.go:180`. Now the daemon starts and `schema_migrations` is `version=61 dirty=false`. The dirty state is unreachable as claimed.

Also checked the properties the new DML has to get right, since it mutates live operator data:

- **Survivor** — 5 mixed Pending/Frozen rows with descending timestamps: exactly the newest survives, the rest move to 6.
- **Ties** — 3 rows with byte-identical `creation_time_unix`: `batch_id DESC` breaks the tie deterministically, reproducible across runs.
- **State written** — 6 is `SeedlingCancelled`, which is what `caretaker.Cancel()` writes for Pending/Frozen. Committed batches would need `SproutCancelled` but are excluded by the `WHERE batch_state IN (0, 1)` filter, so the migration cannot write a state the Go code would not expect for that batch shape.
- **Atomicity** — genuinely one transaction on both dialects, not by accident: the sqlite driver wraps `Run()` in `Begin()`/`Commit()` (`NoTxWrap` false with the empty `Config{}` at `tapdb/sqlite.go:250`), and the postgres driver runs the whole file in one `ExecContext` with `MultiStatementEnabled` false. Partial application is not reachable.
- **Collateral** — only `batch_state` is touched; seedlings, mint anchor uni-commitments and supply pre-commits keep their rows and FKs, i.e. the same on-disk shape a normal cancel produces.
- **The new test is real** — I deleted the `UPDATE` block and re-ran `TestMigration61CancelsDuplicatePreBroadcastBatches`: it fails with the 2067 error at `tapdb/migrations_test.go:1490`. It asserts the right things (survivor id, preserved `batch_state`, cancelled count), so it is a genuine regression test.

One caveat stated plainly: there is no Postgres server available in my environment, so I ran nothing against it. I read the SQL for dialect hazards instead (`NOT IN (SELECT ... ORDER BY ... LIMIT 1)` and a constant-expression partial unique index are both valid PG; `postgresSchemaReplacements` rewrites none of the substrings in this file) and confirmed the check-runs on this head include a green `run unit tests (unit dbbackend=postgres pkg=tapdb)`, which does execute this test.

Blocker cleared. One minor non-blocking note about the down migration is in the new review.

tapgarden/planter.go
litbot-9000: Resolved. `cancelFailedBatch` (`tapgarden/planter.go:1934`) now mirrors `caretaker.Cancel()` exactly, including the `default` arm, so a batch at or past Broadcast is left untouched.

I checked the two claims the fix rests on rather than assuming them:

- **"The restart path resumes it" is true.** `FetchNonFinalBatches` is `FetchMintingBatchesByInverseState(BatchStateFinalized)` (`tapdb/asset_minting.go:1195`), so Broadcast rows are returned; `Start()` skips only the two `*Cancelled` states (`planter.go:534`), skips the fund/seal/freeze block for non-pre-broadcast batches, and unconditionally builds and starts a caretaker. Not stranded.
- **No path left on the old behaviour.** The only cancel-state writes in `planter.go` are `:554`, `:1915` and the helper at `:1953`. `:554` is reachable only inside the `Pending || Frozen` branch so `SeedlingCancelled` is right there; `:1915` is `cancelMintingBatch`, gated by `canCancelBatch()`.

The `State()` read in the helper is accurate because `finalizeBatch` hands the same `*MintingBatch` pointer to `newCaretakerForBatch` (`planter.go:2886`/`:2898`) and the store keeps it current via `SetStateOnDBSuccess` — which is what made the Broadcast case reachable in the first place, and what makes the new guard correct now.

Blocker cleared.

tapcfg/repair.go
litbot-9000: Resolved. The self-heal genuinely exists in `000061` now (I reproduced it end-to-end — see the reply on the migration thread), so this NOTE is accurate as written rather than describing a behaviour that was not in the tree. The stale `000060` references here are fixed too.

Separately I left one minor, non-blocking comment a few lines further down about the `SliceStable` determinism claim — unrelated to this NOTE.

tapgarden/planter.go
litbot-9000: Acknowledged, and I am not re-escalating this — taking a position rather than restating the concern.

The wedge is unchanged: if `UpdateBatchState` fails at `planter.go:1953` the row stays pre-broadcast while `c.pendingBatch` is cleared at `:2227`, and every later `CommitMintingBatch` trips the index until restart. Logging did not fix the correctness problem, and I will not pretend otherwise.

But **logging is an acceptable resolution here and this should not block.** Triggering it needs a DB write failure or a shutdown at one specific instant; the new `log.Errorf` names both the consequence and the remedy; and restart genuinely recovers, since `Start()` picks the row back up, funds/seals/freezes it and assigns a caretaker (`planter.go:575-646`). Worth noting for the record: the `CancelBatch` RPC cannot clear it either, because `canCancelBatch` requires a non-nil `c.pendingBatch` — so restart really is the only remedy, which is exactly what the message tells the operator. Your call to keep it as-is is reasonable.

tapgarden/interface.go
litbot-9000: Resolved. `grep -rn 000060` over `*.go` / `*.md` / `*.conf` outside `tapdb/sqlc/migrations` now returns zero hits — all nine sites updated (`tapgarden/interface.go:207`, `planter.go:516/1439/1844/1866`, `planter_test.go:1633/2170`, `tapcfg/repair.go:18/50`).

tapgarden/planter.go
```diff
@@ -609,12 +623,12 @@ func (c *ChainPlanter) Start() error {
...
-
+				// able to add new seedlings to the batch. The
+				// store call below moves both the on-disk row
+				// and the in-memory mirror atomically; if it
+				// fails, neither has moved.
 				err := c.cfg.Log.UpdateBatchState(
-					ctx, batch.BatchKey.PubKey,
-					BatchStateFrozen,
+					ctx, batch, BatchStateFrozen,
 				)
```
GeorgeTsagk, line 632: If the resumed batch's caretaker fails its Frozen step (`LoadTapscriptTree`, `seedlingsToAssetSprouts`, `AddSproutsToBatch`), the error goes to the caretaker's `BroadcastErrChan` (`caretaker.go:369`), which only the interactive finalize handler reads. The row stays `Frozen` with no owner, so with the 000061 index every later `CommitMintingBatch`/`FundBatch` fails on the unique constraint, and each restart repeats the same failing resume. Pre-PR this row was harmless; `--repair.cancel-duplicate-batches` won't clear it either (it early-returns at `<= 1`). The resume path needs the same cancel handling as finalize.

tapgarden/planter.go
```diff
@@ -510,6 +511,19 @@ func (c *ChainPlanter) Start() error {
...
 
+		// Enforce the singleton invariant: at most one batch may
+		// be in BatchStatePending or BatchStateFrozen at a time.
+		// The DB constraint added in migration 000061 should
+		// already make this impossible, but a legacy DB that was
+		// migrated post-population, or a manually-modified row,
+		// could still violate it. Surfacing the error here gives
+		// the operator a human-readable diagnostic instead of an
+		// opaque SQL one later.
+		if err := checkSingletonInvariant(nonFinalBatches); err != nil {
```
GeorgeTsagk, line 522: The resume loop below is async (freeze at 630, caretaker at 642) and `go c.gardener()` at 654 starts serving immediately. For a large resumed batch, mints and `FundBatch` during the sprout build hit the singleton index and get a raw SQL error instead of a real rejection.

tapgarden/planter.go
```diff
@@ -1864,6 +1922,43 @@ func (c *ChainPlanter) cancelMintingBatch(ctx context.Context,
...
+func (c *ChainPlanter) cancelFailedBatch() {
+	batchKeySerial := asset.ToSerialized(c.pendingBatch.BatchKey.PubKey)
+
+	var cancelState BatchState
+	switch batchState := c.pendingBatch.State(); batchState {
+	case BatchStatePending, BatchStateFrozen:
+		cancelState = BatchStateSeedlingCancelled
+
+	case BatchStateCommitted:
+		cancelState = BatchStateSproutCancelled
```
GeorgeTsagk: The index only constrains `{Pending, Frozen}`, so cancelling `Committed` isn't needed here. If `SignAndFinalizePsbt`/`ImportTaprootOutput` fails on a brief lnd outage (`caretaker.go:805-837`), a batch that used to be resumed and retried on restart is now permanently `SproutCancelled`, forcing a re-mint with new asset IDs. Restrict to `Pending`/`Frozen` and let `Committed` fall to the default branch.

tapgarden/planter.go
```diff
@@ -2116,6 +2210,13 @@ func (c *ChainPlanter) gardener() {
 
 					delete(c.caretakers, batchKeySerial)
 
+					// Cancel the failed batch on disk if
+					// it is still pre-broadcast, so it
+					// isn't left wedged in a state the
+					// migration 000061 singleton index
+					// forbids.
+					c.cancelFailedBatch()
```
GeorgeTsagk: At line 2239 `c.pendingBatch = nil` runs even when `cancelMintingBatch` failed. The row then stays `Pending` with no in-memory owner and every later mint/fund fails on the index until restart. Clear it only when `err == nil`.

tapgarden/planter.go
```diff
@@ -1430,10 +1429,53 @@ func freezeMintingBatch(ctx context.Context, batchStore MintingStore,
...
+	}
+
+	if len(preBroadcastKeys) <= 1 {
+		return nil
+	}
+
+	return fmt.Errorf("singleton pre-broadcast batch invariant "+
+		"violated: found %d batches in BatchStatePending or "+
+		"BatchStateFrozen (keys: %v); at most one is permitted. "+
+		"Resolve by running `tapd --repair.cancel-duplicate-batches` "+
```
GeorgeTsagk, line 1500: This guidance is dead. Migration 000061 already cancels duplicates, so `checkSingletonInvariant` won't fire; and `RunRepairTool` reports "nothing to do" at `<= 1` pre-broadcast batch (`tapcfg/repair.go:108`), which is the wedge from my comment on line 630. Either make the migration fail loudly as the PR description claims, or point operators at something that helps.

tapgarden/batch.go
```diff
@@ -499,38 +516,107 @@ func (m *MintingBatch) validateUniCommitment(newSeedling Seedling) error {
...
+// more than one such seedling, an error is returned.
+//
+// This invariant ("exactly one anchor per batch") is required by
+// callers that derive batch-wide properties (the delegation key,
+// the pre-commitment group key) from the anchor seedling: with no
+// anchor there is no answer, and with multiple anchors the answer
+// is ambiguous. The function computes the answer deterministically
+// rather than relying on non-deterministic map iteration to land
+// on the unique anchor by luck.
+func (m *MintingBatch) uniqueAnchorSeedling() (*Seedling, error) {
```
GeorgeTsagk, line 539: The check cited here (`validateUniCommitment`, line 496) counts `GroupAnchor != nil`, i.e. "at most one child", not "exactly one anchor". It doesn't enforce what this helper requires. Restate the guard in terms of anchor count.

tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
```diff
@@ -0,0 +1,40 @@
...
+UPDATE asset_minting_batches
+SET batch_state = 6
+WHERE batch_state IN (0, 1)
+  AND batch_id NOT IN (
+      SELECT batch_id
+      FROM asset_minting_batches
+      WHERE batch_state IN (0, 1)
+      ORDER BY creation_time_unix DESC, batch_id DESC
+      LIMIT 1
+  );
```
GeorgeTsagk, line 30: This silently cancels every pre-broadcast batch but the newest on a routine upgrade, with no opt-in and no log line; funded batches lose work and leave inputs leased. It also contradicts the PR description ("fail startup with the migration error"), and the tie-break differs from `RunRepairTool`'s (`SliceStable` over `FetchNonFinalBatches` order), so the two paths can keep different batches.

tapcfg/repair.go
```diff
@@ -0,0 +1,151 @@
...
+			"(migrations skipped)")
+		db, err = tapdb.NewPostgresStore(&pgCfg)
+
+	default:
+		return fmt.Errorf("unknown database backend: %s",
+			cfg.DatabaseBackend)
+	}
+	if err != nil {
+		return fmt.Errorf("repair: unable to open database: %w", err)
+	}
```
GeorgeTsagk, line 80: No `defer db.Close()`.

cmd/tapd/main.go
```diff
@@ -35,6 +35,20 @@ func main() {
...
+	// constructing the full server. The repair tool opens the DB
+	// with migrations skipped, so it can recover a legacy DB whose
+	// state would otherwise block a migration from applying.
+	if cfg.Repair != nil && cfg.Repair.CancelDuplicateBatches {
+		err := tapcfg.RunRepairTool(cfg, cfgLogger, shutdownInterceptor)
+		if err != nil {
+			_, _ = fmt.Fprintln(os.Stderr, err)
+			os.Exit(1)
+		}
+		os.Exit(0)
```
GeorgeTsagk: `repair.cancel-duplicate-batches` is a normal config key, so leaving it in `tapd.conf` makes every later start run a no-op repair and `os.Exit(0)`: a supervised node stops serving with a success code. Reject the key from the config file, or exit non-zero when there's nothing to repair.

tapgarden/planter_test.go
```diff
@@ -2208,46 +2164,14 @@ func testFundSealOnRestart(t *mintingTestHarness) {
...
-	)
-	t.assertLastBatchState(batchCount, tapgarden.BatchStateFinalized)
+	// The original test continued by inserting a second Pending
+	// batch directly on disk to exercise recovery when multiple
+	// pending batches were present. That scenario is now forbidden
+	// by the singleton invariant added in migration 000061, so the
+	// section has been removed. The single-pending-batch recovery
+	// paths exercised above are the surviving useful coverage; the
+	// "multiple pre-broadcast batches" case is covered by
+	// TestSingletonPreBroadcastBatchConstraint in tapdb.
```
GeorgeTsagk, line 2174: This also removed the only failing-resume coverage (`FailFeeEstimatesOnce` + `assertBatchResumedBackground`), which is the class that now wedges minting. `TestSingletonPreBroadcastBatchConstraint` doesn't touch the resume path; a single-pending-batch resume-failure test would keep it.

GeorgeTsagk (commented):
Reviewed at `2b1c6af3` vs base `95bd8d41`. Build, vet (incl. itest tags) and `go test ./tapgarden/... ./tapdb/...` pass; the `ON t ((1)) WHERE batch_state IN (0,1)` index works on both sqlite and postgres 13.

Direction looks good. Main objection: the singleton index turns several previously benign failure states into persistent minting outages, and the shipped recovery paths don't cover the one operators will hit. Details inline.

tapgarden/planter.go
jtobin: Fixed in updated changeset.

tapgarden/planter.go
jtobin: Fixed in updated changeset (now surfaces as a proper error).

tapgarden/planter.go
jtobin: Fixed in updated changeset.

tapgarden/planter.go
jtobin: Fixed in updated changeset.

tapgarden/planter.go
jtobin: Fable claims "post-migration, this is defense in depth," and then explains some weird corner case where it still could theoretically apply. So yeah probably low-value, but retaining regardless.

tapgarden/batch.go
jtobin: Fixed in updated changeset.

tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
jtobin: PR description was stale and is updated. Tie-break mismatch is fixed in updated changeset.

tapcfg/repair.go
jtobin: Fixed in updated changeset.

cmd/tapd/main.go
jtobin: Fixed in updated changeset.

tapgarden/planter_test.go
jtobin: The removed section was for two-batch stuff, which was made impossible, but caretaker failure-after-freeze had no coverage, so that's fixed in the updated changeset.

tapgarden/planter.go
```diff
@@ -2116,6 +2267,13 @@ func (c *ChainPlanter) gardener() {
 
 					delete(c.caretakers, batchKeySerial)
 
+					// Cancel the failed batch on disk if
+					// it is still pre-broadcast, so it
+					// isn't left wedged in a state the
+					// migration 000061 singleton index
+					// forbids.
+					c.cancelFailedBatch(caretaker.cfg.Batch)
```
GeorgeTsagk: The orphaning bug 22c1a64f fixed on the cancel path is still here. `cancelFailedBatch` only logs when `UpdateBatchState` fails, and line 2284 then clears `pendingBatch` unconditionally, so a failed on-disk cancel leaves a Pending or Frozen row with no in-memory owner and every later mint hits the index until restart.

Have `cancelFailedBatch` return an error, and clear `pendingBatch` only on success or when the batch was already at Committed or beyond.

tapcfg/repair.go
```diff
@@ -0,0 +1,171 @@
...
+		return nil
+	}
+
+	// Sort newest-first by CreationTime; preserve [0], cancel the
+	// rest. Ties on the (second-precision) creation timestamp are
+	// broken by the raw batch key, matching the ORDER BY used by
+	// migration 000061's self-heal so both paths preserve the
+	// same batch.
+	sort.Slice(preBroadcast, func(i, j int) bool {
+		ti := preBroadcast[i].CreationTime.Unix()
```
GeorgeTsagk: Tie-break still diverges from the migration: this compares Unix seconds while 000061 orders on the full `creation_time_unix`, so two batches in the same second with different sub-second parts get different survivors.

The "second-precision" claim above does not hold. I stored `12:00:00.123456` and read back `123456000` ns on sqlite, and postgres keeps microseconds. Comparing `CreationTime.After` with the same key tiebreak closes it.

tapdb/asset_minting.go
```diff
@@ -497,6 +504,11 @@ func (a *AssetMintingStore) CommitMintingBatch(ctx context.Context,
 			CreationTimeUnix:    newBatch.CreationTime.UTC(),
 			UniverseCommitments: newBatch.SupplyCommitments,
 		}); err != nil {
+			var uniqueErr *ErrSqlUniqueConstraintViolation
```
GeorgeTsagk, line 513: This flags any unique-constraint failure on the batch insert, a `batch_id` primary-key collision included, then reports it as "another minting batch is already awaiting broadcast". Unlikely to fire with freshly derived batch keys, but if it does the error points at the wrong problem. Worth narrowing to the index, or noting the assumption.

tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
```diff
@@ -0,0 +1,44 @@
...
+-- fail here with an opaque unique-index error and force the operator
+-- into a manual repair step, cancel all but the most recent
+-- pre-broadcast batch first. The preserved row is the one with the
+-- latest creation_time_unix, tie-broken by the raw batch key, so the
+-- outcome is deterministic on identical timestamps and matches the
+-- ordering used by the tapd repair tool
+-- (--repair.cancel-duplicate-batches), which cannot see batch_id.
+-- Cancelled rows move to BatchStateSeedlingCancelled = 6, which
+-- leaves them and their seedlings on disk for later inspection.
+UPDATE asset_minting_batches
```
GeorgeTsagk, line 19: The self-heal has no test. Nothing seeds two pre-broadcast rows and asserts which survives, so this UPDATE, the new `internal_keys` join, and the tie-break alignment from f3570803 are all uncovered. TestSingletonPreBroadcastBatchConstraint only exercises the index, and RunRepairTool has no coverage either.

docs/release-notes/release-notes-0.9.0.md
```diff
@@ -25,6 +25,10 @@
   fixes a bug that could cause minted assets to commit to the wrong
   address.
 
+* [PR#2203](https://github.com/lightninglabs/taproot-assets/pull/2203)
```
GeorgeTsagk, line 28: Since the migration keeps its silent self-heal, this note is the only warning an operator gets, and it mentions neither operator-visible fact: upgrading cancels all but the newest legacy pre-broadcast batch, and `--repair.cancel-duplicate-batches` is new. Worth spelling both out.

GeorgeTsagk (commented):
Re-reviewed at 6c51f76a. Build, vet, `go test ./tapgarden/... ./tapdb/...`, `-race` on TestBatchedAssetIssuance, and migration 000061 under postgres all pass.

All ten previous comments are addressed. One correction on my side: my note about removed failing-resume coverage was wrong, the fund-failure and seal-failure resume assertions were never removed, only the two-batch section was.

Five things left, none blocking. The first is the same failure mode as the cancel-path fix in 22c1a64f, on the finalize path.

tapgarden/planter.go
jtobin: Fixed in updated changeset.

tapcfg/repair.go
jtobin: Fixed in updated changeset.

tapdb/asset_minting.go
jtobin: Fixed in updated changeset.

tapdb/sqlc/migrations/000061_unique_pending_or_frozen_batch.up.sql
jtobin: Fixed in updated changeset (though didn't bother with RunRepairTool coverage).

docs/release-notes/release-notes-0.9.0.md
jtobin: Fixed in updated changeset.

sample-tapd.conf
```diff
@@ -543,4 +543,13 @@
...
+; healthcheck.tls.interval=1m
+
+[repair]
+
+; If set, tapd cancels all but the most recent minting batch in
+; BatchStatePending or BatchStateFrozen and then exits. Used to recover
+; from a database that violates the singleton pre-broadcast batch
+; invariant added in migration 000061 (e.g. a legacy DB with duplicate
+; pending batches that blocks the migration).
+; repair.cancel-duplicate-batches=false
```
GeorgeTsagk, line 555: Nit: this advertises the key as a config-file option, but main.go deliberately ignores it when it comes from a config file and just warns. Worth a line here saying it is command-line only, otherwise the sample points operators at the one place the flag does nothing.

tapgarden/seedling.go
```diff
@@ -23,36 +23,9 @@ var (
...
-	// MintingStateSprout denotes that a seedling has been paired with a
-	// genesis transaction and broadcast for confirmation.
-	MintingStateSprout
-
-	// MintingStateAdult denotes that a seedling has been confirmed on
-	// chain and reached full adulthood.
-	MintingStateAdult
-)
-
 // SeedlingUpdate is a struct used to send notifications w.r.t the state of a
```
GeorgeTsagk, line 26: Nit: `MintingState` goes away here, but `bench/fixture/mint_driver.go:117` and `:139` still refer to `MintingStateSeed` in comments. Stale references to a deleted symbol.

GeorgeTsagk (approved):
LGTM, some non-blocking nits.

Reviewed the full changeset this time, not just the follow-ups. Traced every `stateStep` transition to confirm each one persists through a store call before the in-memory mirror moves, so dropping the blanket write in `advanceStateUntil` is safe, and confirmed all three `uniqueAnchorSeedling` call sites are gated on `SupplyCommitments`, which is what makes the exactly-one-anchor invariant hold.

Locally: build, vet (incl. itest tags), `go test -race ./tapgarden/...`, `go test ./tapdb/... ./rpcserver/...`, `go test -tags=test_db_postgres ./tapdb/...` and `make lint-source` (0 issues) all pass.

sample-tapd.conf
jtobin: wontfix; it's quite possible this option, which is just there as a cover for a bizarre/presumed extremely-rare corner case, won't ever be used anyway.

tapgarden/seedling.go
jtobin: wontfix for the interest of expediency.

GeorgeTsagk:
Reminder: we should do another flake hunt pass, we usually have to click Merge 2-3 times..

PR #2219, opened 2026-07-15
Add mssmt arithmetic package

GeorgeTsagk:
## Description

Adds a new package to abstract away the arithmetic calculations over our mssmt. This will also serve as a nice placeholder in the future if we decide to move to bigger numeric types (uint128 etc).


jtobin (approved):
LGTM. :+1: :+1:

PR #2222, opened 2026-07-17
Fix group association during backup import

### LINKED ISSUE #2220: [bug]: wallet backup restore drops group key association — grouped assets become visible but unspendable
### Background

After exporting a wallet asset backup that contains a **grouped** asset (`tapcli assets backup export`, reproduced with both `raw` and `compact` modes) and importing it into the same node after the tapd data directory was lost (LND wallet intact), the asset is listed correctly — right amount, `script_key_is_local: true`, correct chain anchor, managed UTXO present, no lease — but the restored record has `asset_group: null`, and **every send attempt fails**:

```
unable to fund address send: error funding packet: unable to list eligible coins:
unable to query commitments: failed to find coin(s) that satisfy given constraints;
if previous transfers are un-confirmed, wait for them to confirm before trying again
```

Both split sends and full-value sends fail; restarting tapd does not help. The import log shows `Pre-extracted 1 group key(s) from backup proof data`, so the group key is present in the backup blob but is not re-associated with the restored asset.

This looks related to (but distinct from) #2111: there the import **aborted** with `group key not known` when the importing node had never seen the group key. Here — presumably after the group-key pre-extraction fix — the import **succeeds**, but the group association is silently dropped, leaving the asset visible yet unspendable, which is arguably worse for recovery scenarios.

Two isolating observations from the same environment:
- A **non-grouped** asset restored from the same backup file spends normally, so the defect is specific to grouped assets.
- Restoring a file-level copy of the original `~/.tapd` directory (same wallet, same chain state) makes the exact same send succeed, so keys/wallet/chain view are fine — only the backup-import path loses the group association.

### Your environment

- `tapd` v0.8.0-alpha (commit v0.8.0), integrated mode inside `litd` v0.17.0-alpha
- `lnd` v0.21.1-beta (integrated)
- regtest, bitcoind 27.1 backend, Docker (official `lightninglabs/lightning-terminal:v0.17.0-alpha` image)

### Steps to reproduce

1. Node A: `tapcli assets mint --type normal --name GROUPED --supply 10000 --new_grouped_asset`, finalize, confirm.
2. Node B: create a V1 address for 100 units (universe proof courier), node A sends, confirm; node B sees the asset and can spend it.
3. Node B: `tapcli assets backup export --mode raw --output_file b.bin`.
4. Stop node B; move `~/.tapd` away (keep `~/.lnd` untouched); start node B again (fresh tapd DB).
5. `tapcli assets backup import --backup_file b.bin` → returns `num_imported: 1, num_skipped: 0`.
6. `tapcli assets list` shows the asset with the right amount but `asset_group: null`.
7. Any `tapcli assets send` for that asset fails with the "failed to find coin(s)" error above.

### Expected behaviour

Assets restored from a wallet backup keep their group association and remain spendable.

### Actual behaviour

Grouped assets are visible but permanently unspendable after restore; ungrouped assets restored from the same backup work fine.

GeorgeTsagk:
## Description

Closes #2220

### DIFF
--- a/itest/backup_test.go
+++ b/itest/backup_test.go
@@ -472,10 +472,12 @@ func assertAssetsMatch(t *harnessTest, expected []*taprpc.Asset,
 //  3. Bob (fresh LND, no federation) imports RAW — both assets
 //     imported, NumSkipped=0
 //  4. Charlie (fresh LND, no federation) imports COMPACT — same
-//  5. Verify asset counts and group key presence on both nodes
+//  5. Transfer the grouped asset to Dave and export all backup modes
+//  6. Restore each mode onto Dave's LND with fresh tapd state
+//  7. Verify the group association and post-restore spendability
 func testBackupRestoreGrouped(t *harnessTest) {
 	ctxb := context.Background()
-	ctxt, cancel := context.WithTimeout(ctxb, defaultWaitTimeout*4)
+	ctxt, cancel := context.WithTimeout(ctxb, defaultWaitTimeout*6)
 	defer cancel()
 
 	// Mint a grouped asset and an ungrouped asset together.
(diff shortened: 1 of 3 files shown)


jtobin (approved):
LGTM. :+1: :+1:

PR #2225, opened 2026-07-29
backup: restore LND key family indexes

### LINKED ISSUE #2221: [bug]: full recovery (seed + backup import) resets address key derivation — new addresses reuse historical script keys, breaking receives
### Background

The wallet asset backup (`tapcli assets backup export/import`) does not preserve the address key derivation state. After a node loses its `~/.tapd` directory (LND wallet intact) and restores via `assets backup import`, every new `tapcli addrs new` re-derives script keys that were already used by pre-loss addresses/outputs. The custodian logs make this visible:

```
GRDN: Taproot addr <p2tr> was already added to wallet before, skipping
```

A counterparty that holds proofs for those historical script keys (e.g. the node that originally sent assets to this wallet) then fails to send to **any** new address of the restored node:

```
unable to fund address send: error funding packet:
error extracting allocations: output amount exceeds input sum
```

The same sender pays a fresh third node without issues, and the restored node's addresses were confirmed to carry script keys identical to keys that currently anchor existing assets (we observed a new address re-using the exact script key that holds the wallet's own current balance).

Net effect: a node restored from the native backup cannot receive from its previous counterparties — in practice its main peers — and on-chain address/key reuse is silently reintroduced. A file-level backup of `~/.tapd` (which preserves the key counters) does not exhibit the problem.

This compounds with #2220 (grouped assets restored unspendable): together they mean the native backup path currently cannot fully recover a wallet.

### Your environment

- `tapd` v0.8.0-alpha (commit v0.8.0), integrated mode inside `litd` v0.17.0-alpha
- `lnd` v0.21.1-beta (integrated)
- regtest, bitcoind 27.1 backend, Docker (official `lightninglabs/lightning-terminal:v0.17.0-alpha` image)

### Steps to reproduce

1. Node A mints an asset and sends 100 units to node B via a V1 address (universe proof courier); do a few back-and-forth transfers so several script keys have been used.
2. Node B: `tapcli assets backup export --mode raw --output_file b.bin`.
3. Stop node B; move `~/.tapd` away (keep `~/.lnd` untouched); start node B; `tapcli assets backup import --backup_file b.bin`.
4. Node B: `tapcli addrs new --asset_id <id> --amt 9` → tapd logs "already added to wallet before, skipping"; the new address's script key equals a historical one.
5. Node A: `tapcli assets send --addr <new B address>` → fails with "output amount exceeds input sum". Repeating with additional fresh addresses from B fails the same way; sending to a clean node C succeeds.

### Expected behaviour

Either the backup carries the address/script-key derivation state, or the restore path re-syncs the key index (or `addrs new` skips keys that already have proofs/outputs), so that a restored wallet can safely receive again without key reuse.

### Actual behaviour

Restored wallets re-derive already-used script keys; senders holding proofs for those keys fail with allocation errors, and address reuse is reintroduced silently.

GeorgeTsagk:
## Description

  After restoring LND from aezeed, LND has no record of the derivation indexes previously consumed by tapd’s key family. This can cause newly generated Taproot Asset addresses to reuse historical script or internal keys.

  This PR adds an optional key-family high-water marker to asset wallet backups. During backup import, tapd verifies that each marker belongs to the connected LND seed and advances LND’s key-family counter through the recorded index using `DeriveNextKey`.

  The marker is encoded as a trailing, checksummed extension, preserving compatibility:

  - New tapd versions can read backups with or without the marker.
  - Legacy backups fall back to the highest local key locator represented by their active assets.
  - Older readers continue to decode the existing backup payload.
  - Raw, compact, and optimistic backup modes use the same extension.

  Recovery is monotonic and never rewinds LND’s counter. Markers belonging to another seed are ignored, and advancement is bounded to prevent excessive derivation work.

## Tests 

  Tests cover:

  - Marker round trips for raw, compact, and optimistic backup versions.
  - Decoding legacy backups where the marker field is absent.
  - Legacy asset-locator fallback.
  - No counter rewind.
  - Wrong-seed markers.
  - Maximum advancement limits.
  - Full aezeed recovery, backup import, fresh address derivation, subsequent spending, and proof validation.
  - Marker export/import across all three backup modes.

### Other

Closes #2221

### DIFF
--- a/backup/encoding.go
+++ b/backup/encoding.go
@@ -11,6 +11,7 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 
@@ -61,6 +62,16 @@ const (
 	// maxFederationURLLen is the maximum length of a single federation
 	// URL in bytes.
 	maxFederationURLLen = 2048
+
+	// keyMarkerMagicBytes identifies the optional key-family marker section
+	// appended to the wallet backup payload. Keeping this as a trailing
+	// extension lets older readers continue to import newer backups while
+	// ignoring the marker data.
+	keyMarkerMagicBytes = "KEYI"
+
+	// maxKeyFamilyMarkers bounds the number of key-family markers decoded
+	// from an untrusted backup.
+	maxKeyFamilyMarkers = 100
 )
 
 // TLV type constants for AssetBackup fields.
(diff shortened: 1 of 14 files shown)


hieblmi:
> ## Description
> After restoring LND from aezeed, LND has no record of the derivation indexes previously consumed by tapd’s key family. This can cause newly generated Taproot Asset addresses to reuse historical script or internal keys.
> 

Just a drive-by comment:

We face the same problem for static addresses in loop as well. 

Instead of modifying the backup file we will go for a hd-wallet style recovery with index gap limits.
Constantly modifying a backup file and setting a new watermark can be tricky since you always also have to sync the 
secure storage where the backup file lies.

While this functionality is not in place yet we provide a single command line tool to recover specific on-chain outputs that the client identifies manually.

hieblmi:
Idea: Maybe a lnd supported tapscript wallet recovery feature would be nice that scans for generic tapscript outputs on-chain during recovery, additionally to the normal wallet outputs...

backup/key_index.go
```diff
@@ -0,0 +1,134 @@
...
+	"fmt"
+
+	"github.com/lightningnetwork/lnd/keychain"
+)
+
+const (
+	// maxKeyIndexAdvance bounds the number of DeriveNextKey RPCs an import
+	// can trigger for one family. This protects against resource exhaustion
+	// from a backup with a valid-looking but unreasonable marker.
+	maxKeyIndexAdvance = 10_000
```
jtobin: Nit: Fable argued this could probably stand to be bumped up ("each address consumes two keys, so ~5k addresses exceed the bound"). Seems reasonable? Arguably it could be dropped entirely as an "attacker" would already have the seed IIUC.

backup/encoding.go
```diff
@@ -61,6 +62,16 @@ const (
...
+
+	// keyMarkerMagicBytes identifies the optional key-family marker section
+	// appended to the wallet backup payload. Keeping this as a trailing
+	// extension lets older readers continue to import newer backups while
+	// ignoring the marker data.
+	keyMarkerMagicBytes = "KEYI"
+
+	// maxKeyFamilyMarkers bounds the number of key-family markers decoded
+	// from an untrusted backup.
+	maxKeyFamilyMarkers = 100
```
jtobin, line 74: Nit: Fable points out that "every tapd key lives in only one family (212) and the export writes only one marker." The generality is probably unnecessary.

jtobin (approved):
LGTM. :+1: :+1:

(A couple of nits maybe worth addressing but I don't think they block anything.)

jtobin:
I thought the CI failure was a flake, but looks like it's related to the changes. Looks like the test just needs a quick fix (Fable):

> The mechanism: the modified testRestoreLndFromSeed wipes Bob with bob.stop(true), assuming that destroys tapd's state. But stop(true) just does os.RemoveAll(BaseDir) (itest/tapd_harness.go:773) — it deletes the tapd data directory. On the sqlite backend the database file lives inside that directory, so the wipe is real and the test passes. On the postgres backend the database lives in a postgres fixture container provisioned at harness setup (tapd_harness.go:377) and torn down only at test cleanup — stop(true) doesn't touch it. Bob restarts, reconnects to the same postgres database with every asset still in it, the backup import skips them all as already present, and the PR's own assertion require.NotZero(importResp.NumImported) at send_test.go:2240 fails with exactly the observed "Should not be zero, but was 0".
>
> The clean fix is to stand up a genuinely fresh tapd for the restored node — a new setupTapdHarness instance (which provisions its own postgres fixture) attached to the seed-restored lnd — rather than restarting the same harness and trusting stop(true) to have wiped state.

GeorgeTsagk:
Adding a complementary LND PR for this which should also close https://github.com/lightningnetwork/lnd/issues/8698

GeorgeTsagk:
> I thought the CI failure was a flake, but looks like it's related to the changes. Looks like the test just needs a quick fix (Fable):

No this was not a flake it seems. It's an independent postgres issue, will fix in a separate PR.

jtobin (commented):
@GeorgeTsagk does this one require lightningnetwork/lnd#11095 and lightninglabs/lndclient#288 before merge after the recent updates?

(EDIT: confirmed it does out-of-band.)

litbot-9000:
@jtobin: review reminder

litbot-9000:
@jtobin: review reminder

litbot-9000:
@jtobin: review reminder

Armss6556 (approved):

litbot-9000:
@jtobin: review reminder

PR #2226, opened 2026-07-29
proof: support versioned custom-anchor transition proofs

darioAnongba:
Dependency for Wavelength Taproot Asset OOR transfers.

- add BIP-371 tapscript output exclusion proofs
- let `CommitVirtualPsbts` select proof version 1
- cover the proof, tapsend, and RPC behavior with focused tests
- trim the development image context used by source-built integration tests
- add release notes for the RPC update

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+.git
+.github
+
+# Locally built daemon and integration-test binaries are not required by the
+# development image. Excluding them keeps the Docker build context small and
+# stable while another test build may be updating those files.
+tapcli-debug
+tapd-debug
+itest/*-itest
+itest/chantools/chantools
+
+# Runtime and editor artifacts never belong in a source build context.
+*.log
+.DS_Store
+test-artifacts
--- a/docs/release-notes/release-notes-0.9.0.md
+++ b/docs/release-notes/release-notes-0.9.0.md
@@ -39,6 +39,10 @@
 
 ## RPC Updates
 
+* [PR#2226](https://github.com/lightninglabs/taproot-assets/pull/2226)
+  lets `CommitVirtualPsbts` select the transition proof version and adds
+  BIP-371 tapscript sibling exclusion proofs for version 1 proofs.
+
 ## tapcli Updates
 
 ## Config Changes
--- a/itest/multisig.go
+++ b/itest/multisig.go
@@ -385,10 +385,21 @@ func DeriveKeys(t *testing.T, tapd commands.RpcClientsBundle) (asset.ScriptKey,
 	return *scriptKey, internalKeyLnd
 }
 
+type commitVirtualPsbtsOption func(*wrpc.CommitVirtualPsbtsRequest)
+
+func withTransitionProofVersion(
+	version wrpc.TransitionProofVersion) commitVirtualPsbtsOption {
+
+	return func(request *wrpc.CommitVirtualPsbtsRequest) {
+		request.TransitionProofVersion = version
+	}
+}
+
 func CommitVirtualPsbts(t *testing.T, funder commands.RpcClientsBundle,
 	packet *psbt.Packet, activePackets []*tappsbt.VPacket,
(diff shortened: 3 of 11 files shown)


GeorgeTsagk:
Verified the new code against the proof engine locally. The construction is correct: I built mixed
  leaf/branch roots (`[1,2,2]` and `[2,2,1]`) plus TLV encode/decode/re-derive round trips for all four
  emitted shapes, and all derive the right output key. Good that the emitted proofs map onto the five
  existing `DeriveTaprootKeys` cases, so there are no new wire constructs and an older tapd will still
  accept these.

  **Comments below are all coverage, not correctness.**

  ### The swap is unprotected

  `createTapscriptProofFromRoot`'s branch/leaf reorder is the one non-obvious line in the diff and
  nothing tests it. I replaced the condition with `if false`:

  ```go
  if false {
      first, second = second, first
  }
  ```

  The full suite still passes. The five cases in `TestAddExclusionProofsTapTrees` are bip86, lone leaf,
  (leaf,leaf) and two (branch,branch), none of which produce a mixed root. Two more cases cover it:

  - `depths: []uint8{1, 2, 2}` gives `branch(leaf0, branch(leaf1, leaf2))`, so (leaf, branch), no swap
  - `depths: []uint8{2, 2, 1}` gives `branch(branch(leaf0, leaf1), leaf2)`, so (branch, leaf), swap fires

  ### No serialization round trip

  Leaf-vs-branch is an encoded annotation on `TapscriptPreimage`, and it selects which
  `DeriveTaprootKeys` case runs. Every test works on the in-memory struct, so a bug that flips that type
  byte would derive the wrong key from an otherwise correct proof and go unnoticed. It round-trips today;
  worth pinning with an `Encode`/`Decode` step before the derive assertion.

  ### The tapscript path never meets the verifier

  `TestAddExclusionProofsTapTrees` calls `DeriveTaprootKeys` directly and compares against
  `schnorr.ParsePubKey(pkScript[2:])`, which re-implements `verifyTaprootProof`'s check instead of
  invoking it. The itest change only asserts `ProofSuffix.Version` on `testPsbtExternalCommit`, whose
  anchor tx has only BIP-86 outputs.

  So the V0 to V1 flip is covered end to end (`AssertNonInteractiveRecvComplete` means Bob verified
  through the real engine) but the tapscript-tree code has no engine exposure at all. The engine is
  unchanged, so what a `proof.Verify` test buys you is the plumbing: `verifyExclusionProofs` deletes each
  proven index from the P2TR set and errors on leftovers, and that has never run with a tapscript-tree
  output present. One itest with a tapscript-sibling output in the anchor tx, round-tripped through a
  receive, closes this and the round-trip gap together.

### Behavior change to call out

  The new pkScript check also applies to BIP-86 outputs, which were previously not validated against the
  output script. Better than emitting a silently invalid proof, but it can turn a previously succeeding
  `CommitVirtualPsbts` into a hard error, and `tapchannel/aux_funding_controller.go:1208` uses the same
  default generator with no coverage. Worth a release note line.

  ### Smaller things

  - `transitionProofOption` has no doc comment.
  - The odd-leaf-version check in `parseBIP371TapTree` is redundant. `asset.CheckTapLeafSanity` (via
    `NewPreimageFromLeaf`) already rejects anything that isn't `BaseLeafVersion`, which is what
    `TestAddExclusionProofsRejectsUnsupportedTapTreeLeaf` actually asserts.
  - There are now two ways to build a `TapscriptProof`: this one and `proof.CreateTapscriptProof`
    (`tapsend/allocation.go:742`), which re-derives shape via `AssembleTaprootScriptTree` and so can't
    handle these trees. Reshaping it to take a `txscript.TapNode` would give one path, now or as a
    follow-up.
  - The `.dockerignore` commit is unrelated and would be easier to land separately.

  Happy to see this land once the swap cases and one engine-level test are in.

jtobin (approved):
LGTM. :+1: :+1:

PR #2228, opened 2026-07-29
rpcserver: report invalid proofs instead of failing VerifyProof

### LINKED ISSUE #2227: tapd: VerifyProof can fail while reporting an invalid grouped proof
## Description

`TaprootAssets.VerifyProof` intentionally treats a proof-verification failure
as data: it logs the verifier error, sets `valid=false`, and should return a
successful RPC response. However, it then unconditionally marshals the last
proof. On a fresh tapd that does not know the grouped asset or its metadata,
that marshaling path can fail while looking up the decimal display value.
The caller receives an RPC error instead of `valid=false`.

This was observed on current `main` (`f329827a`) while verifying a grouped
asset proof on a fresh receiver tapd:

```text
Proof verification failed with err: group key not known:
... group verifier: asset group is unknown: group key not known

[/taprpc.TaprootAssets/VerifyProof]: unable to marshal proof:
unable to marshal chain asset:
unable to fetch asset meta for asset_id=<asset_id>:
asset lookup failed for asset: issuance-<asset_id>
```

## Reproduction

1. Mint a grouped asset on tapd A and export its proof file.
2. Start a fresh tapd B that has neither the asset group nor its metadata.
3. Call `VerifyProof` on tapd B with that proof file.
4. The proof verifier correctly determines that the group is unknown.
5. `VerifyProof` calls `marshalProof` anyway; `MarshalChainAsset` calls
   `DecDisplayForAssetID`, and the RPC returns an asset-meta lookup error.

The relevant flow is in `rpcserver/rpcserver.go`:

- `proofFile.Verify` records the verification error and computes `valid`;
- `proofFile.LastProof` and `marshalProof` run unconditionally afterward.

The proto already describes `decoded_proof` as present only when the proof
file is valid.

## Expected behavior

An invalid proof should produce a successful response with:

```text
valid=false
decoded_proof=nil
```

Optional display metadata should not turn an already-classified invalid proof
into an RPC failure.

## Suggested direction

Return immediately with `Valid: false` when verification fails, before
marshaling the last proof, or otherwise make decoded-proof marshaling
best-effort for the invalid case. Add a regression test where verification
fails and the address book has no metadata for the proof's asset.

Importing the public group issuance before verification is a valid caller
workaround and may be required to verify the proof successfully, but it does
not fix the `VerifyProof` response contract for invalid proofs.

darioAnongba:
Fixes #2227.

`VerifyProof` intentionally treats a proof verification failure as data: it logs the verifier error and should respond with `valid=false`. However, it then unconditionally decodes and marshals the last proof. On a node that does not know the asset (for example a fresh receiver verifying a grouped proof), marshaling the optional decimal display metadata performs a Universe lookup that fails, and the caller receives an RPC error instead of `valid=false`.

This change returns `valid=false` immediately when verification fails, before decoding the last proof. That matches the documented response contract: `decoded_proof` is only set when the proof file is valid.

The regression test builds an encodable proof file with an unknown file version, so verification fails deterministically before any chain or Universe dependency is consulted; on the previous code the same request failed the whole RPC.

### DIFF
--- a/docs/release-notes/release-notes-0.9.0.md
+++ b/docs/release-notes/release-notes-0.9.0.md
@@ -25,6 +25,11 @@
   fixes a bug that could cause minted assets to commit to the wrong
   address.
 
+* [PR#2228](https://github.com/lightninglabs/taproot-assets/pull/2228)
+  fixes `VerifyProof` failing the whole RPC for an invalid proof when
+  decoding the last proof requires unknown asset metadata; it now
+  returns `valid=false` as documented.
+
 # New Features
 
 ## Functional Enhancements
--- a/rpcserver/rpcserver.go
+++ b/rpcserver/rpcserver.go
@@ -2178,10 +2178,17 @@ func (r *RPCServer) VerifyProof(ctx context.Context,
 	if err != nil {
 		// We don't want to fail the RPC request because of a proof
 		// verification error, but we do want to log it for easier
-		// debugging.
+		// debugging. The decoded proof is only returned for valid
+		// proof files, so we can respond right away. Decoding the
+		// last proof here could itself fail (for example when the
+		// asset is entirely unknown to this node) and would turn an
+		// already-classified invalid proof into an RPC failure.
 		rpcsLog.Errorf("Proof verification failed with err: %v", err)
+
+		return &taprpc.VerifyProofResponse{
+			Valid: false,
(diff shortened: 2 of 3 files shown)


jtobin (approved):
LGTM. :+1: :+1:

PR #2234, opened 2026-08-10
tapchannel: align signature list to vpackets

jtobin:
Previously the second-level HTLC signature list assumed that it lined up with the virtual packets derived from the commitment; this fix assures that's always really the case.


darioAnongba (approved):
Approved, but needs to squash the commits (1 commit enough IMO) because they don't really stack up.

GeorgeTsagk:
Reviewing, disabled auto-merge. Will be done soon

GeorgeTsagk (approved):
Lgtm

PR #1899, opened 2025-12-04
Improve HTLC interceptor setup handling and fix itest flake

ffranr:
This PR strengthens tapd startup guarantees and fixes the flake in the `unknown_TLV_type` itest.

#### Key points

* HTLC interceptor setup failures now cause tapd to exit rather than continue in a degraded state. This ensures that a running tapd instance always has a functioning RFQ subsystem.
* The itest flake was caused by two tapd instances sharing the same lnd backend, competing for a single HTLC interceptor. The test now gives Charlie its own lnd node, eliminating the conflict.

PR #2235, opened 2026-08-10
rfq: don't shut down when an HTLC is resolved on chain

### LINKED ISSUE #2223: [bug]: tapd fails to start: `cannot resume held htlc in the on-chain flow`
### Background

tapd won't start:

```
2026-07-23 08:16:31.131 [ERR] RFQS: Error in RFQ manager: encountered RFQ subsystem error in main event loop: setting up HTLC interception: unable to setup incoming HTLC interceptor: rpc error: code = Unknown desc = cannot resume held htlc in the on-chain flow
2026-07-23 08:16:31.131 [ERR] STAT: could not start the taproot-assets sub-server: received critical error from sub-server (taproot-assets), shutting down: encountered RFQ subsystem error in main event loop: setting up HTLC interception: unable to setup incoming HTLC interceptor: rpc error: code = Unknown desc = cannot resume held htlc in the on-chain flow
```

### Your environment

```sh
litd 0.17.0, tapd 0.8.0, lnd 0.21.1
```

### Steps to reproduce

This node has a pending channel with onchain HTLCs which are not swept yet, this might be of impact since the error message mentions onchain HTLCs

### Expected behavior

I would assume tapd is able to start fine since onchain HTLCs aren't an anomaly for lightning nodes.

### Actual behavior

Litd and lnd spin up fine, but tapd never comes to live. 

EDIT:
Probably this exposed this issue: https://github.com/lightninglabs/taproot-assets/pull/1899

GeorgeTsagk:
### Description

Fixes #2223.

When a forwarded HTLC ends up being resolved on chain, lnd offers it to the HTLC interceptor a second time but rejects every resolution we can send (`ErrCannotResume` / `ErrCannotFail`), and a single rejection terminates the whole interception stream.

This is treated as a critical error since #1899, so one such HTLC took the daemon down and kept it down, as the held HTLC is replayed to every interceptor that connects.

We now claim each incoming circuit key before resolving it and simply don't answer an HTLC we already resolved (lnd doesn't need the answer as it drops those from its held set with the next block), and if the stream does die from one of those rejections we re-establish it with a backoff instead of exiting, while all other errors stay fatal.

### DIFF
--- a/cmd/tapd-integrated/main.go
+++ b/cmd/tapd-integrated/main.go
@@ -250,13 +250,22 @@ func run() error {
 	// lndclient doesn't try to load macaroon files from disk (which
 	// don't exist). lnd won't validate the macaroon anyway.
 	activeNet := cfg.Lnd.ActiveNetParams.Name
+	// NOTE: We must not block on lnd being fully synced to chain here.
+	// lnd's block processing can be blocked on tapd's aux components (for
+	// example on the aux sweeper, if a channel is being resolved on
+	// chain), and those only become available once we configure and start
+	// the tapd server below. Since lnd only reports itself as synced to
+	// chain once its blockbeat dispatcher caught up, waiting for that here
+	// would deadlock the startup. We do wait for the chain notifier to be
+	// ready though, as tapd needs it to subscribe to new blocks.
 	svcsCfg := &lndclient.LndServicesConfig{
-		LndAddress:            lndAddr,
-		Network:               lndclient.Network(activeNet),
-		Insecure:              true,
-		BlockUntilChainSynced: true,
-		BlockUntilUnlocked:    true,
-		CallerCtx:             ctx,
+		LndAddress:              lndAddr,
+		Network:                 lndclient.Network(activeNet),
+		Insecure:                true,
+		BlockUntilChainSynced:   false,
+		BlockUntilChainNotifier: true,
+		BlockUntilUnlocked:      true,
+		CallerCtx:               ctx,
 	}
 	if cfg.Lnd.NoMacaroons {
 		// Use a dummy macaroon that lndclient can deserialize.
--- a/itest/custom_channels/onchain_htlc_intercept_test.go
+++ b/itest/custom_channels/onchain_htlc_intercept_test.go
@@ -1,0 +1,255 @@
+//go:build itest
+
+package custom_channels
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"testing"
+	"time"
+
+	"github.com/lightninglabs/taproot-assets/itest"
+	"github.com/lightninglabs/taproot-assets/proof"
+	"github.com/lightninglabs/taproot-assets/taprpc"
+	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lntest"
+	"github.com/lightningnetwork/lnd/lntest/node"
+	"github.com/lightningnetwork/lnd/lntest/port"
+	"github.com/lightningnetwork/lnd/lntest/wait"
+	"github.com/stretchr/testify/require"
+)
+
+// testCustomChannelsOnChainHtlcIntercept reproduces the bug reported in issue
+// #2223, where tapd refuses to (re)start with:
+//
+//	cannot resume held htlc in the on-chain flow
+//
+// Topology:
+//
+//	Charlie --[assets]--> Dave --[sats]--> Erin --[assets]--> Fabia
+//
+// Fabia creates a hodl invoice which Charlie pays with assets, so the HTLC
+// stays in flight on every hop. We then force close the normal Dave <-> Erin
+// channel, which means Erin's *incoming* HTLC has to be resolved on chain. Erin
+// is an RFQ edge node, so its tapd has an HTLC interceptor registered with lnd.
+//
+// lnd's contract court offers such an on-chain HTLC to the registered HTLC
+// interceptor (tapd's RFQ order handler) via the preimage beacon, and the
+// interceptable switch keeps it in its held HTLC set. Any resolution the
+// interceptor sends back for an on-chain forward is rejected by lnd with
+// ErrCannotResume/ErrCannotFail ("cannot resume in the on-chain flow"), which
+// tears down the whole ForwardHtlcInterceptor stream. Since PR #1899, tapd
+// treats a failure of the interception stream as a critical error, so the
+// daemon shuts down. On a restart the held HTLC is replayed to the freshly
+// connected interceptor, so tapd never manages to come up again.
+func testCustomChannelsOnChainHtlcIntercept(ctx context.Context,
+	net *itest.IntegratedNetworkHarness, t *ccHarnessTest) {
+
+	lndArgs := slices.Clone(lndArgsTemplate)
+	tapdArgs := slices.Clone(tapdArgsTemplate)
+
+	charliePort := port.NextAvailablePort()
+	tapdArgs = append(tapdArgs, fmt.Sprintf(
+		"--proofcourieraddr=%s://%s",
+		proof.UniverseRpcCourierType,
+		fmt.Sprintf(node.ListenerFormat, charliePort),
+	))
+
+	charlieLndArgs := slices.Clone(lndArgs)
+	charlieLndArgs = append(charlieLndArgs, fmt.Sprintf(
+		"--rpclisten=127.0.0.1:%d", charliePort,
+	))
+	charlie := net.NewNode("Charlie", charlieLndArgs, tapdArgs)
+	dave := net.NewNode("Dave", lndArgs, tapdArgs)
+	erin := net.NewNode("Erin", lndArgs, tapdArgs)
+	fabia := net.NewNode("Fabia", lndArgs, tapdArgs)
+	yara := net.NewNode("Yara", lndArgs, tapdArgs)
+
+	nodes := []*itest.IntegratedNode{charlie, dave, erin, fabia, yara}
+	connectAllNodes(t.t, net, nodes)
+	fundAllNodes(t.t, net, nodes)
+
+	// Create the normal (BTC) channel between Dave and Erin.
+	t.Logf("Opening normal channel between Dave and Erin...")
+	channelOp := openChannelAndAssert(
+		t, net, dave, erin, lntest.OpenChannelParams{
+			Amt:         10_000_000,
+			SatPerVByte: 5,
+		},
+	)
+
+	// Everyone needs to know about the only public channel.
+	assertChannelKnown(t.t, charlie, channelOp)
+	assertChannelKnown(t.t, fabia, channelOp)
+
+	// Mint an asset on Charlie and sync universes.
+	mintedAssets := itest.MintAssetsConfirmBatch(
+		t.t, net.Miner, asTapd(charlie),
+		[]*mintrpc.MintAssetRequest{
+			{
+				Asset: ccItestAsset,
+			},
+		},
+	)
+	cents := mintedAssets[0]
+	assetID := cents.AssetGenesis.AssetId
+
+	t.Logf("Minted %d lightning cents, syncing universes...", cents.Amount)
+	syncUniverses(t.t, charlie, dave, erin, fabia, yara)
+	t.Logf("Universes synced, creating asset network...")
+
+	const (
+		assetSendAmount   = uint64(400_000)
+		daveFundingAmount = uint64(400_000)
+		erinFundingAmount = uint64(200_000)
+	)
+	charlieFundingAmount := cents.Amount - (2 * assetSendAmount)
+
+	createTestAssetNetwork(
+		t, net, charlie, dave, erin, fabia, yara, charlie,
+		cents, assetSendAmount, charlieFundingAmount,
+		daveFundingAmount, erinFundingAmount, DefaultPushSat,
+	)
+
+	// Ensure nodes know each other in the graph.
+	require.NoError(t.t, net.AssertNodeKnown(charlie, dave))
+	require.NoError(t.t, net.AssertNodeKnown(dave, charlie))
+	require.NoError(t.t, net.AssertNodeKnown(erin, fabia))
+	require.NoError(t.t, net.AssertNodeKnown(fabia, erin))
+	require.NoError(t.t, net.AssertNodeKnown(charlie, erin))
+
+	logBalance(t.t, nodes, assetID, "initial")
+
+	// Fabia creates a hodl invoice, so the multi-hop HTLC will be stuck in
+	// flight until we force close below.
+	const invoiceAssetAmount = 20_000
+	hodlInvoice := createAssetHodlInvoice(
+		t.t, erin, fabia, invoiceAssetAmount, assetID,
+	)
+
+	t.Logf("Charlie paying Fabia's hodl invoice, HTLC will stay in " +
+		"flight...")
+	payInvoiceWithAssets(
+		t.t, charlie, dave, hodlInvoice.payReq, assetID,
+		withFailure(lnrpc.Payment_IN_FLIGHT, failureNone),
+	)
+
+	// Make sure Dave actually has the forwarded HTLC on both of his
+	// channels before we go on chain.
+	assertNumHtlcs(t.t, dave, 2)
+
+	logBalance(t.t, nodes, assetID, "with in-flight HTLC")
+
+	// Now we force close the normal channel between Dave and Erin. This
+	// forces Erin to resolve its incoming HTLC on chain, which makes lnd
+	// hand that HTLC to tapd's HTLC interceptor.
+	t.Logf("Force closing the Dave <-> Erin channel...")
+	_, closeTxid, err := net.CloseChannel(dave, channelOp, true)
+	require.NoError(t.t, err)
+
+	t.Logf("Channel force closed, close_txid=%v, mining blocks...",
+		closeTxid)
+
+	// Confirm the force close transaction, which starts the contract
+	// resolution on Erin's side.
+	mineBlocks(t, net, 1, 1)
+
+	// Erin's contract court now hands the incoming HTLC to the registered
+	// HTLC interceptor. tapd must survive that. Before the fix it goes down
+	// with a critical error, because lnd rejects any resolution for an
+	// on-chain HTLC and tears down the whole interception stream.
+	t.Logf("Asserting Erin stays alive during the on-chain resolution...")
+	assertNodeAlive(t.t, erin, 30*time.Second)
+
+	// Sanity check that the on-chain HTLC resolution actually started.
+	assertPendingForceCloseHtlcs(t.t, erin, 1)
+
+	// And finally, a restart must work as well: on startup lnd replays all
+	// held HTLCs (including the on-chain one) to the newly connected
+	// interceptor.
+	t.Logf("Restarting Erin...")
+	erin.Restart()
+
+	assertNodeAlive(t.t, erin, ccShortTimeout)
+}
+
+// assertPendingForceCloseHtlcs waits until the given node reports at least the
+// given number of pending HTLCs across all its pending force closing channels.
+func assertPendingForceCloseHtlcs(t *testing.T, n *itest.IntegratedNode,
+	expected int) {
+
+	t.Helper()
+
+	err := wait.NoError(func() error {
+		ctxt, cancel := context.WithTimeout(
+			context.Background(), ccShortTimeout,
+		)
+		defer cancel()
+
+		resp, err := n.PendingChannels(
+			ctxt, &lnrpc.PendingChannelsRequest{},
+		)
+		if err != nil {
+			return err
+		}
+
+		var numHtlcs int
+		for _, c := range resp.PendingForceClosingChannels {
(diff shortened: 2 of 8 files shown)


rfq/order.go
```diff
@@ -1100,6 +1251,25 @@ func (h *OrderHandler) subscribeHtlcs(ctx context.Context) error {
...
+			// ever reports a resolved HTLC under a different type.
+			if failEvent != nil || linkFail != nil ||
+				settleEvent != nil || finalHtlcEvent != nil {
+
+				h.resolvedHtlcs.Delete(circuitKey)
+			}
+
+			// We only care about forwarding events for the rest of
+			// the handling below.
+			if event.GetEventType() != routerrpc.HtlcEvent_FORWARD {
```
litbot-9000, line 1344: **The `finalHtlcEvent` branch below is dead code, so the on-chain accounting never runs.**

Moving this type check down so the `resolvedHtlcs.Delete` above it runs for `FinalHtlcEvent`s is exactly right — lnd never sets an event type for those. But the `case finalHtlcEvent != nil:` branch 10 lines below (`// HTLC resolved on-chain (force close scenario)`) is still *behind* this check, and therefore unreachable.

Evidence, from the pinned `lnd@40c64f9db30d`:
- `lnrpc/routerrpc/subscribe_events.go:74-89` — for `*htlcswitch.FinalHtlcEvent` the conversion sets `key` and `timestamp` but leaves `eventType = nil`.
- `subscribe_events.go:100-116` — `rpcEvent.EventType` is only assigned when `eventType != nil`.
- `router.pb.go:305-308` — `HtlcEvent_UNKNOWN = 0`, `HtlcEvent_FORWARD = 3`.

So a `FinalHtlcEvent` always arrives as `UNKNOWN` and always hits the `continue`. I confirmed it rather than just reading it — a throwaway test pushing a `FinalHtlcEvent` through `subscribeHtlcs` with the mocks from `order_intercept_test.go`:

```
EventType = UNKNOWN (what lnd actually sends): policy still tracked  -> untracked=false
EventType = FORWARD (hypothetical):            policy removed        -> untracked=true
```

**Concrete failure path, for exactly the HTLC this PR is about.** From the passing run of the new itest, Erin accepts the HTLC that later goes on chain:

```
order.go:1054 Fetched policy with SCID 17687413646837225657 of type *rfq.AssetSalePolicy
order.go:1112 Logged forwarding event rfq_id=a597... asset_amt=19999 amt_in_msat=343608531
order.go:1122 HTLC complies with policy. Broadcasting accept event.
```

So `policy.TrackAcceptedHtlc` ran and `forwardStore.UpsertForward` wrote a row with no `SettledAt`/`FailedAt`. When the incoming channel is force closed, the incoming link is gone, so there is no `ForwardFailEvent` and no `SettleEvent` — `FinalHtlcEvent` is the *only* signal lnd emits. It is dropped here, so:

- `htlcToPolicy.LoadAndDelete` never runs and `policy.UntrackHtlc` is never called, so `AssetSalePolicy.CurrentAmountMsat` (`order.go:325-334`) stays inflated by this HTLC's amount. `CheckHtlcCompliance` (`order.go:308-313`) then rejects otherwise-valid HTLCs against that quote until it expires.
- The `htlcToForward` entry and the forward row stay pending; they're only reconciled on a later restart via `restorePendingForwards` → `LookupHtlcResolution`.

The branch was already dead before this PR, so it isn't a regression — but it's the accounting half of the story this PR is telling, it's on lines you're already editing, and the fix looks like just hoisting the `finalHtlcEvent` case above this check (or handling it in the same block as the `resolvedHtlcs.Delete`).

rfq/order.go
```diff
@@ -910,14 +949,72 @@ func NewOrderHandler(cfg OrderHandlerCfg) (*OrderHandler, error) {
...
+	// interceptor a second time if it ends up being resolved on chain (see
+	// lnd's incoming contest resolver, which notifies the interceptor
+	// through the preimage beacon). Any resolution we send for such an HTLC
+	// is rejected by lnd, which terminates the whole interception stream
+	// and would take the RFQ subsystem (and with it the daemon) down with
+	// it. There is nothing we can do about an HTLC that is already being
+	// resolved on chain, so we simply don't answer. lnd doesn't require an
+	// answer either, it removes those HTLCs from its held set again with
+	// the next block.
+	_, alreadyResolved := h.resolvedHtlcs.LoadOrStore(
```
litbot-9000, line 1037: **This suppression is keyed on "we answered once", not "lnd rejected our answer", and it's permanent (24h).** Worth a comment or a guard, because its safety currently rests entirely on an lnd behaviour that nothing in this repo documents, asserts, or tests.

lnd's own contract says a replay is *not* evidence of on-chain resolution. From `ForwardHtlcInterceptRequest` (`lnd/lnrpc/routerrpc/router.pb.go:2816-2818`):

> Interceptor clients should handle requests for the same circuit key idempotently. Requests may be replayed after reconnect, and an htlc that was previously offered off-chain may be offered again after it moves on-chain.

And `InterceptableSwitch.setInterceptor` (`htlcswitch/interceptable_switch.go:383-395`) replays **every** held HTLC — off-chain held as well as on-chain held — to each newly connected interceptor. Since this PR deliberately makes reconnects routine, that path is now reachable by design.

I tried to break it and mostly couldn't, for one reason: the itest log shows `HSWC interceptable_switch.go:406: Interceptor disconnected, resolving held packets`, i.e. the `!requireInterceptor` branch. With lnd's default, disconnecting the interceptor calls `releaseAllOffChainHeld()`, so only *on-chain* held HTLCs survive to be replayed — precisely the intended case. So this is **not** a default-config break, and I'm not blocking on it.

Where it does bite is `--requireinterceptor` (`interceptable_switch.go:396-402`: "Interceptor disconnected, retaining held packets"), which is exactly the mode an asset-channel operator would want so lnd doesn't blindly forward RFQ HTLCs while tapd is down. I grepped the whole repo for `requireinterceptor` and got zero hits, so I can't tell whether it's a supported deployment. Under it:

1. lnd holds on-chain HTLC `X` plus off-chain held `Y1..Yn`, all retained across the tapd restart.
2. The interceptor connects; lnd replays all of them at once via `heldHtlcSet.forEach`.
3. tapd answers them concurrently; lnd rejects `X` and terminates the stream, so the resolutions for whichever `Yi` hadn't been applied yet are lost — but every `Yi` is already in `resolvedHtlcs`.
4. The retry loop reconnects, lnd replays the still-held `Yi`, and this branch refuses to answer them.
5. Those `Yi` sit held until lnd's expiry safety net fails them back (`interceptable_switch.go:366-371`) — payments that would have succeeded fail, and HTLCs stay locked for many blocks.

Since there's no per-HTLC signal telling you *which* resolution lnd rejected, the cheap mitigations are probably (a) a comment here recording that this is only safe because lnd releases off-chain held HTLCs on disconnect, and/or (b) clearing `resolvedHtlcs` entries whose resolution can't have been applied when a stream tears down. Your call — but as written, "don't shut down" is a silent ignore for any replayed HTLC, not just an on-chain one.

rfq/order.go
```diff
@@ -910,14 +949,72 @@ func NewOrderHandler(cfg OrderHandlerCfg) (*OrderHandler, error) {
...
+
+	return resp, nil
+}
+
+// resolveIncomingHtlc decides how an incoming HTLC should be resolved, based on
+// the RFQ policy that applies to it.
+//
+// NOTE: This function must be thread safe. It is used by an external
+// interceptor service.
+func (h *OrderHandler) resolveIncomingHtlc(ctx context.Context,
```
litbot-9000, line 1078: **Not introduced here, but this PR's own itest exposes it: the quote accounting isn't idempotent per circuit key, so the first answer for a replayed HTLC is a spurious FAIL.**

From the passing run of `on_chain_htlc_intercept`, after Erin restarts and lnd replays the held HTLC:

```
[WRN] order.go:1062: HTLC does not comply with policy: HTLC out amount is greater than the
policy maximum (htlc_out_msat=343607188, policy_max_out_msat=343607188)
  ... policy=&{... 343607188 mSAT ... map[{{469 1 0} 0}:343607188] ...}
```

The policy dump shows `htlcToAmt` **already contains this exact circuit key** at that amount, because `restorePendingForwards` → `restorePendingForward` (`order.go:1462-1471`) re-tracked it on startup. `CheckHtlcCompliance` then adds the same HTLC's amount a second time — `AssetSalePolicy` at `order.go:308` does `c.CurrentAmountMsat + htlc.AmountOutMsat > policyMaxOutMsat` with no exclusion for a key already in `htlcToAmt`; `AssetPurchasePolicy` has the same shape against `PaymentMaxAmt`. Hence `343607188 + 343607188 > 343607188` → non-compliant → FAIL.

`TrackAcceptedHtlc` (`order.go:325-334` and `592-601`) has the matching problem: it does `CurrentAmountMsat += amt` while *overwriting* `htlcToAmt[circuitKey]`, so a re-track permanently inflates `CurrentAmountMsat` (`UntrackHtlc` only subtracts once).

In this test it's harmless because lnd rejects the FAIL anyway, and it's why the retry loop rather than the dedup is what actually saves the daemon here. But it means tapd's reasoning about a replayed HTLC is wrong for the wrong reason: "exceeds the quote", not "already resolved on chain". Any replay that lnd *can* still apply — the `--requireinterceptor` case in my other comment — would fail a perfectly good in-flight asset payment. Making `CheckHtlcCompliance`/`TrackAcceptedHtlc` skip amounts already tracked for the same circuit key would satisfy the idempotency lnd's proto asks for, and would be a good companion to the dedup you're adding here.

rfq/order.go
```diff
@@ -1100,6 +1251,25 @@ func (h *OrderHandler) subscribeHtlcs(ctx context.Context) error {
 				HtlcID: event.IncomingHtlcId,
 			}
 
+			// Once an HTLC reached a final state, we can forget that
```
litbot-9000: CI's **Lint check** is failing on this diff — three `lll` violations (golangci-lint v2.10.1, max 80 with tab-width 8). I reproduced the widths locally:

- `rfq/order.go:1254` (this line) — 81 chars (3 tabs + 57)
- `cmd/tapd-integrated/main.go:253` — 81 chars (1 tab + 73)
- `rfq/order_intercept_test.go:165` — 83 chars (2 tabs + 67)

Everything else on `bf9a7bbf` is green, including all itest and `unit race` shards.

.gitignore
```diff
@@ -58,3 +58,6 @@ vendor/
 /taproot-assets-*
 .aider*
 bench/results/
+
+# Logs from integration test runs started via the Makefile targets.
+/regtest/
```
litbot-9000: This entry is dead, and the commit message claim ("ignores the log directory the itest Makefile targets write to") doesn't hold.

`/regtest/` is anchored at the repo root, but no target writes there: `itest-cc` and `itest-only` both run `go test ./itest/...`, so the test binary's CWD is the *package* directory and `-logdir=regtest/.logs` resolves relative to that. I confirmed empirically — after a run the only such directory is `itest/custom_channels/regtest/`.

Both real paths are already ignored on `main`:

```
$ git show main:.gitignore | grep -n regtest
18:/itest/regtest/*
26:/itest/custom_channels/regtest/

$ git check-ignore -v itest/custom_channels/regtest/.logs/integrated-Erin.log
.gitignore:26:/itest/custom_channels/regtest/	itest/custom_channels/regtest/.logs/integrated-Erin.log
```

So these three lines can just be dropped.

cmd/tapd-integrated/main.go
```diff
@@ -250,13 +250,22 @@ func run() error {
 	// lndclient doesn't try to load macaroon files from disk (which
 	// don't exist). lnd won't validate the macaroon anyway.
 	activeNet := cfg.Lnd.ActiveNetParams.Name
+	// NOTE: We must not block on lnd being fully synced to chain here. lnd's
```
litbot-9000: I went looking for a race here and didn't find one — recording what I checked so it isn't re-derived, plus one scoping question.

**Verified safe:**
- `BlockUntilChainNotifier` exists in the pinned `lndclient@09c54915cfba` (`lnd_services.go:146`, used at `:474`), so no dep bump is needed.
- No new startup hang. `waitForChainNotifier` (`lnd_services.go:570+`) bounds each attempt with `context.WithTimeout(ctx, timeout)` and hard-fails on deadline rather than retrying — which would be a problem if lnd only delivered the *next* block. It doesn't: lndclient registers with a nil `bestBlock`, and lnd dispatches the **current tip immediately** in that case (`chainntnfs/bitcoindnotify/bitcoind.go:354-371`, and the btcd notifier equivalent). So this returns in milliseconds, not up to a block interval.
- No stale chain view for tapd. The notifier's `bestBlock` tracks the chain *backend's* tip, independent of lnd's blockbeat and of `GetInfo.SyncedToChain`, so tapd's height is correct even while lnd reports itself unsynced. Subscribers also get every intervening block, so an initial height couldn't be cached wrongly anyway.
- Your claim about the itest harness holds: `IntegratedNode.fetchNodeInfo` (`itest/integrated_node.go:377+`) waits for `info.SyncedToChain` before `Start()` returns.

**Two smaller notes:**
- `BlockUntilUnlocked` stays `true`, which doesn't imply the wallet has caught up. So tapd on-chain operations issued immediately after startup can now see a wallet that hasn't seen recent UTXOs — degraded-and-retryable (`insufficient funds`), not corrupting, but a real behaviour change worth a line in the release notes since `tapd-integrated` is a shipped binary.
- Scoping: this commit is independent of the RFQ fix and stands on its own. Is it load-bearing for the new itest (which restarts Erin *after* a force close, i.e. exactly the aux-sweeper deadlock you describe)? If so, saying that in the commit message would help; if not, it might be cleaner as its own PR.

(FWIW this line is also one of the three `lll` failures — 81 chars.)

itest/custom_channels/onchain_htlc_intercept_test.go
```diff
@@ -0,0 +1,250 @@
...
+// lnd's contract court offers such an on-chain HTLC to the registered HTLC
+// interceptor (tapd's RFQ order handler) via the preimage beacon, and the
+// interceptable switch keeps it in its held HTLC set. Any resolution the
+// interceptor sends back for an on-chain forward is rejected by lnd with
+// ErrCannotResume/ErrCannotFail ("cannot resume in the on-chain flow"), which
+// tears down the whole ForwardHtlcInterceptor stream. Since PR #1899, tapd
+// treats a failure of the interception stream as a critical error, so the
+// daemon shuts down. On a restart the held HTLC is replayed to the freshly
+// connected interceptor, so tapd never manages to come up again.
+func testCustomChannelsOnChainHtlcIntercept(ctx context.Context,
```
litbot-9000, line 47: Confirming this test is load-bearing, since that's the most valuable thing to know about a bug-fix PR. I built a branch equal to this PR's head with `cbcd508` reverted (`git diff main -- rfq/` empty, commits 2–4 kept) and ran `make itest-cc cccase=on_chain_htlc_intercept`:

- **without the RFQ fix → FAIL in 53.4s**: `onchain_htlc_intercept_test.go:164: node Erin went down`, `Node Erin exited with error: exit status 1`, and Erin's log carries the verbatim #2223 failure (`RFQS manager.go:335: ... unable to setup incoming HTLC interceptor: rpc error: code = Unknown desc = cannot fail held htlc in the on-chain flow`).
- **with the fix → PASS in 129.5s**, and the log shows both halves of the fix firing: `order.go:977 ... already resolved by us, not responding again`, then `order.go:1165 HTLC interception stream terminated, re-establishing in 500ms`, then the skip again after the reconnect.

So this is a genuine regression test, and the restart leg is doing real work rather than being a formality. Nice.

One optional tidy: `assertNodeAlive(t.t, erin, time.Minute)` at line 164 burns a fixed minute polling only lnd's `GetInfo` in the loop, and checks tapd only once afterwards — that's most of the runtime of a test that otherwise takes ~50s. Since `tapd-integrated` runs lnd in-process via `lnd.Main` (so a critical tapd error takes lnd down too, which is why the lnd poll works at all), you could poll both RPCs in the loop and cut the duration considerably without weakening the assertion.

litbot-9000 (changes requested):
The RFQ fix itself is sound and I verified it empirically: the new itest **does** reproduce issue #2223 and **does** fail without commit `cbcd508` (Erin's tapd exits with exactly the reported `cannot fail held htlc in the on-chain flow` critical error), and it passes with it. Commit 2 also checks out — I traced the lndclient/lnd code and it does not introduce a startup hang or a stale chain view. Two things keep me from a clean pass, though. First, CI's **Lint check is red** on three `lll` violations in this diff. Second, and more interesting: the PR moves the `HtlcEvent_FORWARD` type check in `subscribeHtlcs` specifically so the new `resolvedHtlcs` cleanup runs for `FinalHtlcEvent`s — which is the right instinct — but it leaves the `case finalHtlcEvent != nil:` accounting branch *below* that check, where it is dead code. I confirmed with a probe test that the on-chain resolution path never reaches `handleHtlcSettle`/`handleHtlcFail`, so the quote's in-flight amount is never released and the forward row stays pending for precisely the HTLCs this PR is about. That's a one-line fix on lines you already touched. Details, evidence and two lower-confidence observations are inline.

## What I verified

Clone at `bf9a7bbf`, fresh worktree, SQLite/local only.

**Build and unit tests (all local, all pass)**
- `make build` → exit 0
- `go test ./rfq/...` → PASS (1.9s); `go test ./tapchannel/... ./cmd/...` → PASS
- `go test -race -count=1 ./rfq/... ./tapchannel/...` → PASS (rfq 3.3s, tapchannel 5.4s)

**The new itest is load-bearing — confirmed both directions**
- Built a branch = PR head with `cbcd508` reverted (`git diff main -- rfq/` empty), keeping commits 2–4, then `make itest-cc cccase=on_chain_htlc_intercept`:
  → **FAIL** in 53.4s. `onchain_htlc_intercept_test.go:164: node Erin went down`, `Node Erin exited with error: exit status 1`. Erin's log:
  ```
  [ERR] RPCS interceptor.go:778: [/routerrpc.Router/HtlcInterceptor]: cannot fail held htlc in the on-chain flow
  [ERR] RFQS manager.go:335: Error in RFQ manager: encountered RFQ subsystem error in main event
        loop: setting up HTLC interception: unable to setup incoming HTLC interceptor: rpc error:
        code = Unknown desc = cannot fail held htlc in the on-chain flow
  ```
  i.e. the exact failure from #2223.
- Same command on the PR head → **PASS** in 129.5s. Erin's log shows both mechanisms firing: `order.go:977 ... was already resolved by us, not responding again`, then `order.go:1165 HTLC interception stream terminated, re-establishing in 500ms`, then the skip again after the reconnect.
- (Methodology note: `make itest-cc` runs `build-itest`, which rebuilds `itest/tapd-integrated-itest`. A bare `go test ./itest/custom_channels` silently reuses the stale binary.)

**Source-level checks against the pinned deps**
- `lndclient@09c54915cfba`: `InterceptHtlcs` runs each handler in its own goroutine (`router_client.go:1098-1126`), so the blocking `alreadyResolved` path does **not** head-of-line block other HTLCs; and it derives a child ctx cancelled on return (`:987`), so the blocked goroutine is released on every stream teardown — no leak across reconnects.
- `fn.IsCanceled(context.Canceled)` is true (`fn/errors.go:22-37`), so the `context.Canceled` returned from the `alreadyResolved` path is correctly treated as a clean exit.
- The error-string matching is unavoidable: `ForwardHtlcInterceptRequest` has no on-chain flag (only `auto_fail_height`, which doubles as the settlement deadline), and lndclient doesn't surface it. All four candidate lnd errors (`ErrCannotResume`, `ErrCannotFail`, `ErrCannotResumeOnChain`, `ErrCannotFailOnChain`) contain `in the on-chain flow`, so the match is correct and the unit test's strings are lnd's real ones.

**What I could NOT run**
- No container runtime in this environment, so **no Postgres-tagged paths** (`//go:build test_db_postgres` / `ory/dockertest`). CI covers those: `run itests postgres (0-3)` and `run unit tests (unit dbbackend=postgres ...)` all pass on `bf9a7bbf`, so I'm leaning on CI there.
- `gh pr checks 2235`: everything passes **except Lint check** (fail). All 4 custom-channel itest shards and all 4 `unit race` shards are green, matching my local results.
- I did not attempt to build a reproduction for the conditional concern in my second comment (it needs `--requireinterceptor`, which the itest harness doesn't set); I've flagged it as conditional rather than confirmed.

## Findings (most severe first, all inline)

1. `rfq/order.go:1269` — the `finalHtlcEvent` accounting branch is dead code; on-chain resolutions never untrack the policy or close the forward row. **Confirmed with a probe test.**
2. `rfq/order.go:973` — the resolution suppression is unconditional and keyed on "we answered", not "lnd rejected"; its safety rests on an lnd behaviour nothing here documents or asserts. Conditional, not a default-config break.
3. `rfq/order.go:1014` — quote accounting is not idempotent per circuit key, so the first answer for any replayed HTLC is a spurious "exceeds the quote" FAIL. Visible in this PR's own itest log.
4. `rfq/order.go:1254` — three `lll` violations are failing CI's Lint check.
5. `.gitignore:63` — the added entry is dead; both real log dirs are already ignored on `main`.
6. `cmd/tapd-integrated/main.go:253` — commit 2: what I verified, plus a scoping question.

jtobin (approved):
This broadly LGTM. :+1:

I sort of got hung up on the error string matching, which I know is necessary as-is, but which just sort of sucks. But fixing it properly would require upstream changes, and the string matching doesn't actually make anything _worse_, so I shouldn't bog this down any further.. 😅

PR #2236, opened 2026-08-10
tapchannel: authenticate suffixes before accepting state

jtobin:
Previously funding proof suffixes were checked for internal consistency, but the claimed transition wasn't explicitly tied to the peer's inputs. This just makes sure the check is extended appropriately.

### DIFF
--- a/tapchannel/aux_funding_controller.go
+++ b/tapchannel/aux_funding_controller.go
@@ -326,6 +326,14 @@ type FundingController struct {
 
 	finalizedChans chan funding.PendingChanID
 
+	// validatedFundingsMtx guards validatedFundings.
+	validatedFundingsMtx sync.Mutex
+
+	// validatedFundings tracks the channel funding outpoints whose
+	// funding proofs have already been validated against the confirmed
+	// funding transaction during this process lifetime.
+	validatedFundings map[wire.OutPoint]struct{}
+
 	// ContextGuard provides a wait group and main quit channel that can be
 	// used to create guarded contexts.
 	*fn.ContextGuard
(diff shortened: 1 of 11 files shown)


darioAnongba (changes requested):
Agent report attached, I think findings C-1, H-3 are relevant, H-4 as well but that's only tests.

[lightninglabs_taproot-assets_PR_2236_review.md](https://github.com/user-attachments/files/30988905/lightninglabs_taproot-assets_PR_2236_review.md)

GeorgeTsagk:
> Agent report attached, I think findings C-1, H-3 are relevant, H-4 as well but that's only tests.
> 
> [lightninglabs_taproot-assets_PR_2236_review.md](https://github.com/user-attachments/files/30988905/lightninglabs_taproot-assets_PR_2236_review.md)

C-1 shouldn't be an issue, it's true atm because STXO is optional in the tap channel. Let's switch that to required with v0.9 cc @jtobin

jtobin:
> > Agent report attached, I think findings C-1, H-3 are relevant, H-4 as well but that's only tests.
> > [lightninglabs_taproot-assets_PR_2236_review.md](https://github.com/user-attachments/files/30988905/lightninglabs_taproot-assets_PR_2236_review.md)
> 
> C-1 shouldn't be an issue, it's true atm because STXO is optional in the tap channel. Let's switch that to required with v0.9 cc @jtobin

Agreed we should. Fable cooked up a direct solution to the problem anyway, in the meantime.

tapchannel/aux_funding_controller.go
```diff
@@ -2072,6 +2383,16 @@ func (f *FundingController) chanFunder() {
...
 
+		// The initiator's funding goroutine has created the funding
+		// output proofs and hands them over for the flow state we
+		// own.
+		case req := <-f.assignProofReqs:
+			fundingFlow, ok := fundingFlows[req.pid]
+			if ok {
+				fundingFlow.fundingOutputProofs = req.proofs
+			}
+			close(req.done)
```
GeorgeTsagk: nit: assignProofReqs should report an error back to the caller when the flow is unknown — right now `close(req.done)` runs even when `fundingFlows[req.pid]` misses, so `assignFundingProofs` reports success while the proofs are silently dropped (surfaces much later as a confusing "no funding output proofs" error). Returning an error here would make the failure loud at the source.

GeorgeTsagk (approved):
LGTM

darioAnongba (approved):

tapchannel/aux_funding_controller.go
jtobin: Fixed in updated changeset.

PR #2239, opened 2026-08-17
rfq: fix HTLC race in order policies

GeorgeTsagk:
### Description

`CheckHtlcCompliance` checked the agreed payment ceiling under RLock while `TrackAcceptedHtlc` incremented under a separate lock. A burst of concurrent HTLCs could lead to a race where the write/read values are not in sync.

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -242,6 +242,11 @@
   transfers until restart in the event of a silently-dropped TCP
   connection.
 
+* [PR#2239](https://github.com/lightninglabs/taproot-assets/pull/2239)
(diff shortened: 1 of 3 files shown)


jtobin (approved):
LGTM. :+1:

One nit that Fable pointed out (it probably registers higher on my nit scale than Fable's, if I understand it correctly):

> One semantic note, not a defect: CheckHtlcCompliance is now check-and-reserve — a successful call obligates an eventual UntrackHtlc, or the cap leaks. The interface doc says so, but the name no longer quite matches. A rename (e.g. CheckAndReserveHtlc) would make the contract harder to misuse if the interface grows more callers.

I think CheckAndReserveHtlc actually makes a lot of sense, and it's cheap to rename, so I'd probably go for it. But defo not a blocker.

PR #2245, opened 2026-08-20
gomod: bump grpc to v1.83.1

gijswijs:
Bump `google.golang.org/grpc` from v1.82.1 to v1.83.1, mirroring the bump done in lnd in https://github.com/lightningnetwork/lnd/pull/11105.

The bump is applied to all three Go modules in this repo:
- the main module
- `taprpc`
- `docs/examples/basic-price-oracle`

Via `go mod tidy`, the `genproto/googleapis/api` and `genproto/googleapis/rpc` modules are bumped to `v0.0.0-20260526163538-3dc84a4a5aaa` as well. The `golang.org/x/*` dependencies in this repo were already at or above the versions targeted by the lnd bump, so no further changes were needed.

Verified with `go build ./...` in all three modules and the `authmailbox` unit tests.


jtobin (approved):
LGTM. :+1:

PR #2246, opened 2026-08-20
tapchannel: fix cooperative close for merged funding outputs

### LINKED ISSUE #2243: tapchannel: coop close fails to finalize when a funding output merges multiple inputs
(N.b., Opus's summary of an issue found during v0.8.2-rc2 testing.)

## Summary

`importOutputProofs` fetches the proof for only the *primary* previous
input of each funding output. When a funding output merges several
UTXOs of the same asset ID, verification of the appended close output
proof then fails, because the VM cannot reconstruct the virtual
transaction without all of the inputs.

The cooperative close never finalizes: the channel stays in
`waiting_close` and the peer's assets are not swept.

There is a longstanding TODO immediately above the loop that flags
exactly this (`tapchannel/aux_sweeper.go`):

    // TODO(roasbeef): assume single asset for now, also additional
    // inputs

Observed on signet against v0.8.2-rc2, but the code is byte-identical
in v0.8.1, so this is not release-specific. See
[Attribution](#attribution).

## Environment

- Node A ("new"): litd v0.17.2-alpha built against taproot-assets
  v0.8.2-rc2 (`f831018b`), tapd reports `0.8.2-alpha.rc2`.
- Node B ("old"): stock litd v0.17.2-alpha release, tapd `0.8.1`.
- Both on lnd v0.21.2-beta, signet, neutrino backend.
- Proof courier / federation:
  `universerpc://signet.universe.lightning.finance:443`.

## Root cause

`importOutputProofs` in `tapchannel/aux_sweeper.go` does, per output
proof:

```go
proofPrevID, err := proofToImport.Asset.PrimaryPrevID()
...
prefixProof, err := proofFetcher.ReceiveProof(
        ctx, recipient, inputProofLocator,
)
...
var proofFile proof.File
err = proofFile.Decode(bytes.NewReader(prefixProof.Blob))
if err := proofFile.AppendProof(*proofToImport); err != nil {
...
err = proofArchive.ImportProofs(ctx, vCtx, false, &proof.AnnotatedProof{...})
```

`PrimaryPrevID()` returns only the *first* witness's prev ID
(`asset/asset.go`):

```go
// The primary prev ID is the first witness's prev ID.
primaryWitness := prevWitnesses[0]
return primaryWitness.PrevID, nil
```

So exactly one input proof file is fetched and used as the prefix. If
the funding output asset has more than one `PrevWitness` — i.e. the
channel was funded from several UTXOs of that asset ID, which get
merged into a single funding output — the remaining inputs are absent
from the proof file. `ImportProofs` then verifies the appended
transition proof, the VM tries to build the virtual transaction, and
the missing inputs surface as `ErrNoInputs` (`tapscript/tx.go:21`).

Note this is *not* a missing-courier problem: the courier fetch
succeeds (node B logs `All proofs fetched, importing locator=...`
immediately before the failure). The single proof it fetched is simply
not sufficient.

## Error

From node B's log, verbatim:

    2026-08-20 15:27:45.374 [ERR] CNCT: Failed to handle confirmed
    spend: handle coop close: finalize coop close: unable to import
    output proofs: unable to verify proof: missing asset input(s):
    unable to make virtual txIn (*asset.PrevID)(0x14003b78cb0)
    (PrevID(outpoint=5169764869671c88a8bda4d6b427f3d10e1d459d2283231d
    de29e4cfccf260e7:4,
    id=e8ee29a9a36ba5d294e8a3193ab069843ae584fd49debe09f8b662d4177f91
    77,
    script_key=024ef98501a618fe567488c212537c9cb48c0d12addc6888ddcb16
    81b5108db35a))

Emitted by `ChainWatcher(1e9b2aef...:0)`. The referenced outpoint is
one of eight UTXOs of asset `e8ee29a9...` used to fund that channel.

Error chain: `handle coop close` -> `finalize coop close` -> `unable
to import output proofs` -> `unable to verify proof` -> `missing asset
input(s)` -> `unable to make virtual txIn`.

The error fires once. The `ChannelArbitrator` then logs `no close
event` and moves on, so the channel is left stuck rather than retrying.

## Observed setup

Node A opened a channel to node B funded from 10 UTXOs across 3 asset
IDs sharing one group key:

- `s2236-alpha`: 1 UTXO  -> funding output has 1 input
- `s2236-beta`:  1 UTXO  -> funding output has 1 input
- `s2236-gamma`: 8 UTXOs -> funding output has 8 inputs

The channel opened, went active, and routed asset payments both ways.
On coop close, node A finalized and swept correctly. Node B failed on
the `s2236-gamma` output, which is the only one with multiple inputs.

Because the error propagates up through `finalize coop close`, the
whole finalization aborts — node B did not import its own
`s2236-alpha` output either, and still shows:

    B waiting_close: 1
        1e9b2aef...:0  closing_txid 7c356b73...
    B assets: {'s2236-bravo': 967006}

i.e. its 18,997 `s2236-alpha` was never swept.

A second channel between the same nodes, opened in the opposite
direction and funded from a *single* UTXO of a single asset ID,
cooperatively closed and swept cleanly on both nodes — including on
node B, after the failure above. That is the control that isolates
input count as the trigger.

## Isolating test

Run specifically to separate input count from asset count. A channel
funded from **two UTXOs of a single asset ID** (`s2236-bravo`, split
by a self-transfer beforehand, in a group where the funding node holds
nothing else so coin selection cannot route around the split):

    Validating input proof, prev_out=29d184ca...:0
    Validating input proof, prev_out=29d184ca...:1
    2 x TxAssetInputProof
    1 x TxAssetOutputProof
    local_assets: [('faed318df48d', 30000)]

On coop close the peer failed identically:

    2026-08-20 18:05:37.743 [ERR] CNCT: Failed to handle confirmed
    spend: handle coop close: finalize coop close: unable to import
    output proofs: unable to verify proof: missing asset input(s):
    unable to make virtual txIn
    (PrevID(outpoint=29d184ca818a8ed330e614adb9396850dc52704b8d0acd18
    2aee2e7f1df7c1a6:1,
    id=faed318df48dcde6c02f30a8a95adb2d5f8f30efcb8294c12809b062af2bfa
    9a,
    script_key=0283a7f60d5760a950beb8573794a43ffdbf23b23fc9a13ea4e753
    527d3e216c47))

The missing input is `:1`, the second of the two. `PrimaryPrevID()`
returned `:0`, whose proof was fetched; `:1` never was.

Full matrix observed:

- 3 asset IDs, inputs 1/1/8 -> fails, on the 8-input output
- 1 asset ID, 1 input       -> succeeds
- 1 asset ID, 1 input       -> succeeds (second instance)
- 1 asset ID, 2 inputs      -> fails

Two inputs of one asset ID is sufficient to trigger it. Multiple asset
IDs are not a factor. The "assume single asset" half of the TODO is a
separate concern and was not tested.

## Attribution

Not caused by v0.8.2-rc2:

- `tapchannel/aux_sweeper.go` and `tapchannel/aux_closer.go` are not
  touched by #2236, and the `importOutputProofs` region is
  byte-identical between v0.8.1 and v0.8.2. The `PrimaryPrevID()` call
  and the TODO above it are present in both.
- The `tapfreighter/chain_porter.go` change in #2236 is
  verification-only: the previous code deep-copied the suffix and
  appended `AdditionalInputs` to the throwaway copy purely to drive
  verification, while the new code passes `inputProofFiles` to
  `proof.VerifyProofSuffix`. `vOut.ProofSuffix` — what is transmitted
  and stored — is unchanged either way.
- The `asset/asset.go` and `asset/encoding.go` changes in #2236 are
  nil-guards. No serialization change.

## When this was introduced

Present since the first version of the sweeper. `git blame` puts the
TODO and the `PrimaryPrevID()` call on the same commit:

    d5fe6bc35 (2024-06-04) // TODO(roasbeef): assume single asset for
                           // now, also additional inputs
    d5fe6bc35 (2024-06-04) proofPrevID, err :=
                           proofToImport.Asset.PrimaryPrevID()

`d5fe6bc3` is "tapchannel: add initial version of AuxSweeper", and the
code there is structurally identical to today's. First released in
**v0.4.0** (2024-07-22). `48928390` (2024-07-01) only extracted
`importOutputProofs` into its own function without changing behaviour.

Subsequent commits touching the function are orthogonal (context
params, skip-already-imported, `VerifierCtx`, courier init, spew
limiting); none altered the single-prev-ID fetch.

## Reproduction

1. Mint an asset and split it into two UTXOs (a self-transfer to one
   address, with change, suffices).
2. Open an asset channel funded from both, so the funding output for
   that asset ID merges two inputs.
3. Cooperatively close.
4. The peer fails to import the close output proofs and stays in
   `waiting_close`.

Provenance depth is irrelevant and no payments are needed; a single
split and an immediate close reproduce it. Funding from one UTXO per
asset ID avoids it entirely.

For grouped assets, be aware that coin selection operates group-wide:
`fundchannel --asset_id X` may fund with a different asset ID in the
same group, picking one large UTXO and silently avoiding the split.
Use an asset whose group holds nothing else, or size the request so no
single UTXO covers it.

## Impact

Assets are not lost — the balance sits in the coop close transaction
output and the proofs are retrievable — but the close never finalizes
automatically, the channel is stuck in `waiting_close`, and recovery
needs manual intervention. Funding from multiple UTXOs of one asset ID
is an ordinary thing to do, and coin selection may do it without the
operator choosing to.

## Adjacent issue noticed in passing

`fundchannel --asset_id <id>` on a grouped asset does not honour the
requested asset ID. Requesting `s2236-alpha` (which the node held as
881,003 + 100,000 across two UTXOs) instead funded the channel with
`s2236-gamma`, a different asset ID in the same group that happened to
sit in one 1,000,000 UTXO:

    requested: --asset_id 3898ed74... --asset_amount 950000
    funded:    local_assets: [('e8ee29a9a36b', 950000)]

Selection appears to resolve the asset ID to its group and then pick
freely within it. That may be intended for grouped assets, but
silently funding with an asset ID the caller did not ask for is
surprising, and there is no warning. Not investigated further; noted
because it also masks the bug above by preferring a single large UTXO.

## Suggested directions

- Fetch proof files for *all* of the asset's prev IDs, not just
  `PrimaryPrevID()`, and make them available to verification. The
  pattern already exists elsewhere: `proof.VerifyProofSuffix` takes a
  `map[asset.PrevID]*proof.File`, and the pre-#2236 `chain_porter`
  code populated `proof.AdditionalInputs` for the same reason.
- Consider whether one side should fail its entire close finalization
  on an output proof belonging to the peer; here it blocked node B
  from importing its own output.
- Add itest coverage for a coop close of a channel funded from
  multiple UTXOs of the same asset ID. Existing coverage appears to
  use single-input funding, which is why this is not caught.

jtobin:
Fixes #2243.

A channel funded from several UTXOs of the same asset merges them into a single funding output. When importing that output, we only ever fetched the provenance of its first input. A proof asserting several inputs cannot be verified from just one of them, so the responder was unable to validate the funding output and the cooperative close never finalized: the channel sat in `waiting_close` and the peer's assets were never swept. Channels funded from a single UTXO were unaffected, and no integration test closed a multi-input channel, so the gap went unnoticed. A TODO above the import loop had flagged the case.

We now fetch provenance for every input claimed by the funding output, carrying the non-primary proof files in the transition proof—the same merge shape produced by the ordinary send path. This close-time import also performs structural validation of the funding-input list before retrieving proofs. These checks provide defense in depth alongside the confirmed-funding validation performed during ChannelReady, and turn malformed input data into precise errors rather than an opaque verifier failure. The existing multi-input funding test is extended through a cooperative close, reproducing the original failure and guarding the fix.

### DIFF
--- a/itest/custom_channels/single_asset_multi_input_test.go
+++ b/itest/custom_channels/single_asset_multi_input_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/taprpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
 	tchrpc "github.com/lightninglabs/taproot-assets/taprpc/tapchannelrpc"
+	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/lntest/node"
 	"github.com/lightningnetwork/lnd/lntest/port"
 	"github.com/stretchr/testify/require"
@@ -140,4 +141,25 @@ func testCustomChannelsSingleAssetMultiInput(ctx context.Context,
 		t.t, charlie, dave, 2*halfCentsAmount,
 		[]*taprpc.Asset{cents},
 	)
+
+	// Cooperatively close the multi-input channel. The non-initiator must
+	// fetch both funding input proof files to verify and import the merged
+	// funding output before it can finalize the close.
+	chanPoint := &lnrpc.ChannelPoint{
+		OutputIndex: uint32(fundRespCD.OutputIndex),
+		FundingTxid: &lnrpc.ChannelPoint_FundingTxidStr{
+			FundingTxidStr: fundRespCD.Txid,
+		},
+	}
+	closeAssetChannelAndAssert(
+		t, net, dave, charlie, chanPoint, [][]byte{assetID}, nil,
+		charlie, assertDefaultCoOpCloseBalance(false, false),
+	)
+
+	assertBalance(
+		t.t, dave, cents.Amount, itest.WithAssetID(assetID),
+		itest.WithNumUtxos(1),
+		itest.WithScriptKeyType(asset.ScriptKeyBip86),
+	)
+	assertBalance(t.t, charlie, 0, itest.WithAssetID(assetID))
 }
--- a/tapchannel/aux_sweeper.go
+++ b/tapchannel/aux_sweeper.go
@@ -185,7 +185,7 @@ func (a *AuxSweeper) Start() error {
 
 // Stop stops the AuxSweeper.
 func (a *AuxSweeper) Stop() error {
-	if !a.stopped.CompareAndSwap(true, false) {
+	if !a.stopped.CompareAndSwap(false, true) {
 		return nil
 	}
 
--- a/tapchannel/aux_sweeper.go
+++ b/tapchannel/aux_sweeper.go
@@ -1455,6 +1455,168 @@ func (a *AuxSweeper) importOutputScriptKeys(desc tapscriptSweepDescs) error {
 	)
 }
 
+// fetchInputProofFiles fetches and validates the proof file for every input
+// spent by an output proof. The first returned file is the output asset's
+// direct lineage; the remaining files prove the merge's additional inputs.
+func fetchInputProofFiles(ctx context.Context, outputProof *proof.Proof,
+	courierAddr *url.URL,
+	proofDispatch proof.CourierDispatch) ([]proof.File, error) {
+
+	if outputProof == nil {
+		return nil, fmt.Errorf("output proof is nil")
+	}
+	if len(outputProof.AdditionalInputs) != 0 {
+		return nil, fmt.Errorf("output proof carries additional inputs")
+	}
+
+	prevWitnesses := outputProof.Asset.Witnesses()
+	if len(prevWitnesses) == 0 {
+		return nil, fmt.Errorf("asset missing previous witnesses")
+	}
+	if len(prevWitnesses) > maxFundingInputProofs {
+		return nil, fmt.Errorf(
+			"too many funding input witnesses, got %d, max is %d",
+			len(prevWitnesses), maxFundingInputProofs,
+		)
+	}
+
+	type inputProofRequest struct {
+		prevID    asset.PrevID
+		recipient proof.Recipient
+		locator   proof.Locator
+	}
+
+	requests := make([]inputProofRequest, len(prevWitnesses))
+	seenInputs := make(map[asset.PrevID]struct{}, len(prevWitnesses))
+	for idx := range prevWitnesses {
+		proofPrevID := prevWitnesses[idx].PrevID
+		if proofPrevID == nil {
+			return nil, fmt.Errorf(
+				"funding input witness %d has no "+
+					"previous ID", idx,
+			)
+		}
+		if _, ok := seenInputs[*proofPrevID]; ok {
+			return nil, fmt.Errorf("duplicate funding input %v",
+				*proofPrevID)
+		}
+		seenInputs[*proofPrevID] = struct{}{}
+
+		if idx == 0 && outputProof.PrevOut != proofPrevID.OutPoint {
+			return nil, fmt.Errorf(
+				"output proof previous outpoint %v does "+
+					"not match primary input %v",
+				outputProof.PrevOut,
+				proofPrevID.OutPoint,
+			)
+		}
+		spendsInput := proof.TxSpendsPrevOut(
+			&outputProof.AnchorTx, &proofPrevID.OutPoint,
+		)
+		if !spendsInput {
+			return nil, fmt.Errorf(
+				"funding transaction does not spend input %v",
+				*proofPrevID,
+			)
+		}
+
+		scriptKey, err := proofPrevID.ScriptKey.ToPubKey()
+		if err != nil {
+			return nil, fmt.Errorf(
+				"unable to convert input %d script key "+
+					"to pubkey: %w",
+				idx, err,
+			)
+		}
+
+		inputProofLocator := proof.Locator{
+			AssetID:   &proofPrevID.ID,
+			ScriptKey: *scriptKey,
+			OutPoint:  &proofPrevID.OutPoint,
+		}
+		if outputProof.Asset.GroupKey != nil {
+			groupKey := outputProof.Asset.GroupKey.GroupPubKey
+			inputProofLocator.GroupKey = &groupKey
+		}
+
+		requests[idx] = inputProofRequest{
+			prevID: *proofPrevID,
+			recipient: proof.Recipient{
+				ScriptKey: scriptKey,
+				AssetID:   proofPrevID.ID,
+
+				// PrevID has no amount. This field is only for
+				// logging, so use the output amount.
+				Amount: outputProof.Asset.Amount,
+			},
+			locator: inputProofLocator,
+		}
+	}
+
+	proofFetcher, err := proofDispatch.NewCourier(ctx, courierAddr, true)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"unable to create proof courier: %w", err,
+		)
+	}
+	defer func() {
+		_ = proofFetcher.Close()
+	}()
+
+	inputProofFiles := make([]proof.File, len(requests))
+	for idx := range requests {
+		request := &requests[idx]
+		log.Infof(
+			"Fetching funding input proof %d of %d, locator=%v",
+			idx+1, len(requests),
+			limitSpewer.Sdump(request.locator),
+		)
+
+		inputProof, err := proofFetcher.ReceiveProof(
+			ctx, request.recipient, request.locator,
+		)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"unable to fetch input proof %d (%v): %w", idx,
+				request.prevID, err,
+			)
+		}
+
+		err = inputProofFiles[idx].Decode(
+			bytes.NewReader(inputProof.Blob),
+		)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"unable to decode input proof %d (%v): %w", idx,
+				request.prevID, err,
+			)
+		}
+
+		lastProof, err := inputProofFiles[idx].LastProof()
+		if err != nil {
+			return nil, fmt.Errorf(
+				"unable to read input proof %d (%v): %w", idx,
+				request.prevID, err,
+			)
+		}
+		actualPrevID := asset.PrevID{
+			OutPoint: lastProof.OutPoint(),
+			ID:       lastProof.Asset.ID(),
+			ScriptKey: asset.ToSerialized(
+				lastProof.Asset.ScriptKey.PubKey,
+			),
+		}
+		if actualPrevID != request.prevID {
+			return nil, fmt.Errorf(
+				"input proof mismatch: expected %v, got %v",
+				request.prevID, actualPrevID,
+			)
+		}
+	}
+
+	return inputProofFiles, nil
+}
+
 // importOutputProofs imports the output proofs into the pending asset funding
 // into our local database. This preps us to be able to detect force closes.
 func importOutputProofs(ctx context.Context, scid lnwire.ShortChannelID,
@@ -1464,14 +1626,50 @@ func importOutputProofs(ctx context.Context, scid lnwire.ShortChannelID,
 
 	// TODO(roasbeef): should be part of post confirmation funding validate
 	// (chanvalidate)
+	if len(outputProofs) == 0 {
+		return fmt.Errorf("funding output proofs are missing")
+	}
 
+	seenInputs := make(map[asset.PrevID]struct{})
+	totalInputs := 0
+	for outputIdx, outputProof := range outputProofs {
+		if outputProof == nil {
+			return fmt.Errorf(
+				"funding output proof %d is nil", outputIdx,
+			)
+		}
+		if outputProof.Asset.ScriptKey.PubKey == nil {
+			return fmt.Errorf(
+				"funding output proof %d has no "+
+					"script key", outputIdx,
+			)
+		}
+
+		witnesses := outputProof.Asset.Witnesses()
+		if len(witnesses) > maxFundingInputProofs-totalInputs {
+			return fmt.Errorf("too many funding inputs, max is %d",
+				maxFundingInputProofs)
+		}
+		totalInputs += len(witnesses)
+
+		for _, witness := range witnesses {
+			if witness.PrevID == nil {
+				continue
+			}
+			if _, ok := seenInputs[*witness.PrevID]; ok {
+				return fmt.Errorf(
+					"funding output proof %d reuses "+
+						"input %v",
+					outputIdx, *witness.PrevID,
+				)
+			}
+			seenInputs[*witness.PrevID] = struct{}{}
+		}
+	}
 	log.Infof("Importing %v proofs for ChannelPoint(%v)",
 		len(outputProofs), outputProofs[0].OutPoint())
 
-	// With the fetcher created, we'll have it fetch each of the proofs for
-	// the funding outputs we need.
-	//
-	// TODO(roasbeef): assume single asset for now, also additional inputs
+	// Fetch and import the proof for each funding output.
 	for _, proofToImport := range outputProofs {
 		// Check if the proof is already imported to avoid redundant
 		// work.
@@ -1491,66 +1689,12 @@ func importOutputProofs(ctx context.Context, scid lnwire.ShortChannelID,
 			continue
 		}
 
-		proofPrevID, err := proofToImport.Asset.PrimaryPrevID()
-		if err != nil {
-			return fmt.Errorf("unable to get primary prev "+
-				"ID: %w", err)
-		}
-
-		scriptKey, err := proofPrevID.ScriptKey.ToPubKey()
-		if err != nil {
-			return fmt.Errorf("unable to convert script key to "+
-				"pubkey: %w", err)
-		}
-
-		inputProofLocator := proof.Locator{
-			AssetID:   &proofPrevID.ID,
-			ScriptKey: *scriptKey,
-			OutPoint:  &proofPrevID.OutPoint,
-		}
-		if proofToImport.Asset.GroupKey != nil {
-			groupKey := proofToImport.Asset.GroupKey.GroupPubKey
-			inputProofLocator.GroupKey = &groupKey
-		}
-
-		log.Infof("Fetching funding input proof, locator=%v",
-			limitSpewer.Sdump(inputProofLocator))
-
-		// First, we'll make a courier to use in fetching the proofs we
-		// need.
-		proofFetcher, err := proofDispatch.NewCourier(
-			ctx, courierAddr, true,
-		)
-		if err != nil {
-			return fmt.Errorf("unable to create proof courier: %w",
-				err)
-		}
-
-		recipient := proof.Recipient{
-			ScriptKey: scriptKey,
-			AssetID:   proofPrevID.ID,
-			Amount:    proofToImport.Asset.Amount,
-		}
-		prefixProof, err := proofFetcher.ReceiveProof(
-			ctx, recipient, inputProofLocator,
-		)
-
-		// Always attempt to close the courier, even if we encounter an
-		// error.
-		_ = proofFetcher.Close()
-
-		// Handle any error that occurred during the proof fetch.
-		if err != nil {
-			return fmt.Errorf("unable to fetch prefix "+
-				"proof: %w", err)
-		}
-
-		log.Infof("All proofs fetched, importing locator=%v",
-			limitSpewer.Sdump(inputProofLocator))
-
-		// Before we combine the proofs below, we'll be sure to update
-		// the transition proof to include the proper block+merkle proof
-		// information.
+		// Before we fetch or combine any proofs below, we'll be sure to
+		// update the transition proof to include the proper
+		// block+merkle proof information. This also replaces the
+		// peer-supplied anchor transaction with the confirmed one, so
+		// that the input validation done while fetching is bound to the
+		// funding transaction that actually made it into the chain.
 		err = updateProofsFromShortChanID(
 			ctx, chainBridge, scid, []*proof.Proof{proofToImport},
 		)
@@ -1559,16 +1703,42 @@ func importOutputProofs(ctx context.Context, scid lnwire.ShortChannelID,
 				"proof: %w", err)
 		}
 
-		// Now that we have the entire prefix proof, we'll append the
-		// New funding output, then import it into our archive.
-		//
-		// TODO(roasbeef): way to do this w/o decoding again?
-		var proofFile proof.File
-		err = proofFile.Decode(bytes.NewReader(prefixProof.Blob))
+		inputProofFiles, err := fetchInputProofFiles(
+			ctx, proofToImport, courierAddr, proofDispatch,
+		)
 		if err != nil {
-			return fmt.Errorf("unable to decode proof: %w", err)
+			return fmt.Errorf(
+				"unable to fetch funding input proofs: %w", err,
+			)
 		}
-		if err := proofFile.AppendProof(*proofToImport); err != nil {
+
+		log.Infof("All %d input proofs fetched, importing locator=%v",
+			len(inputProofFiles), limitSpewer.Sdump(fundingLocator))
+
+		// The first input remains the direct proof-file lineage.
+		// Embed the remaining input proof files in the transition proof
+		// so the VM can verify the full virtual transaction.
+		proofWithInputs := *proofToImport
+		proofWithInputs.AdditionalInputs = slices.Clone(
+			inputProofFiles[1:],
+		)
+
+		proofBytes, err := proofWithInputs.Bytes()
+		if err != nil {
+			return fmt.Errorf(
+				"unable to encode output proof: %w", err,
+			)
+		}
+		if len(proofBytes) > proof.FileMaxProofSizeBytes {
+			return fmt.Errorf(
+				"output proof is too large: %d bytes, "+
+					"max is %d",
+				len(proofBytes), proof.FileMaxProofSizeBytes,
+			)
+		}
+
+		proofFile := &inputProofFiles[0]
+		if err := proofFile.AppendProofRaw(proofBytes); err != nil {
 			return fmt.Errorf("unable to append proof: %w", err)
 		}
 
--- a/tapchannel/aux_sweeper.go
+++ b/tapchannel/aux_sweeper.go
@@ -30,6 +30,7 @@ import (
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/keychain"
 	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnutils"
 	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/lightningnetwork/lnd/sweep"
@@ -1925,8 +1926,10 @@ func (a *AuxSweeper) importCommitTx(req lnwallet.ResolutionReq,
 		ChainLookupGen: a.cfg.ChainBridge,
 		IgnoreChecker:  a.cfg.IgnoreChecker,
 	}
+	proofCtx, cancel := lnutils.ContextFromQuit(a.quit)
+	defer cancel()
 	err = importOutputProofs(
-		ctxb, req.ShortChanID, maps.Values(fundingInputProofs),
+		proofCtx, req.ShortChanID, maps.Values(fundingInputProofs),
 		a.cfg.DefaultCourierAddr, a.cfg.ProofFetcher,
 		a.cfg.ChainBridge, vCtx, a.cfg.ProofArchive,
 	)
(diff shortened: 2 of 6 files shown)


tapchannel/aux_sweeper.go
```diff
@@ -1455,6 +1456,168 @@ func (a *AuxSweeper) importOutputScriptKeys(desc tapscriptSweepDescs) error {
...
+
+		if idx == 0 && outputProof.PrevOut != proofPrevID.OutPoint {
+			return nil, fmt.Errorf(
+				"output proof previous outpoint %v does "+
+					"not match primary input %v",
+				outputProof.PrevOut,
+				proofPrevID.OutPoint,
+			)
+		}
+		spendsInput := proof.TxSpendsPrevOut(
```
litbot-9000, line 1514: This check reads as the trust anchor of the new validation, but at this point `outputProof.AnchorTx` is still the **peer-supplied** anchor transaction. `updateProofsFromShortChanID` does not overwrite it with the confirmed funding transaction until line 1708 — after `fetchInputProofFiles` has already returned. `UpdateTransitionProof` is what binds it (`proof/append.go:157`, `p.AnchorTx = proofHeader.AnchorTx`, sourced from `block.Transactions[scid.TxIndex]`).

I tried to refute this and mostly succeeded: `ChannelReady` -> `validateConfirmedFunding` -> `validateConfirmedFundingProofs` (`aux_funding_controller.go:2841`) requires `proofSuffix.OutPoint() == channel.FundingOutpoint`, and since `OutPoint()` derives from `AnchorTx.TxHash()` and a txid commits to every input prevout, `AnchorTx.TxIn` *is* transitively chain-bound for any channel that passed `ChannelReady` on current `main`. So this is **not** an accepted-invalid-funding-output hole.

The residual case is concrete, though: `validateConfirmedFunding` landed in `80e32969` (2026-08-17) and is in no released tag, and `ChannelReady` only fires once, at channel open. Every channel opened under v0.8.1/v0.8.2 will therefore coop-close under this code without ever having been bound. For those channels a peer that fabricates `AnchorTx.TxIn` satisfies this check trivially, and the cost is up to `maxFundingInputProofs` (64) courier round-trips against attacker-chosen `(assetID, scriptKey, outpoint)` locators before `ImportProofs` verification rejects the result. Not exploitable for value — wasted work, and universe queries the peer gets to choose.

Hoisting the `updateProofsFromShortChanID` call above `fetchInputProofFiles` makes the check self-contained and costs nothing. It would also match the doctrine already spelled out in `validateConfirmedFundingProofs`' own doc comment: "at that point the real funding transaction did not exist yet, so their anchor identity remained unbound".

itest/custom_channels/single_asset_multi_input_test.go
```diff
@@ -140,4 +141,25 @@ func testCustomChannelsSingleAssetMultiInput(ctx context.Context,
...
+		t, net, dave, charlie, chanPoint, [][]byte{assetID}, nil,
+		charlie, assertDefaultCoOpCloseBalance(false, false),
+	)
+
+	assertBalance(
+		t.t, dave, cents.Amount, itest.WithAssetID(assetID),
+		itest.WithNumUtxos(1),
+		itest.WithScriptKeyType(asset.ScriptKeyBip86),
+	)
+	assertBalance(t.t, charlie, 0, itest.WithAssetID(assetID))
```
litbot-9000, line 164: This assertion is vacuous: line 136 already asserts Charlie is at 0 before the close, and `assertDefaultCoOpCloseBalance(false, false)` sets `remoteAssetBalance=false`, so Charlie holds no asset balance in the channel at any point. No harm in it — I confirmed separately that the load-bearing assertion is `assertClosedChannelAssetData` on `remote` inside the helper, not this line.

The gap it leaves is worth naming, though: the test proves the responder can *import* a merged funding output, but never that it can *sweep* its own asset balance out of one — which is the user-visible symptom in #2243 ("its 18,997 `s2236-alpha` was never swept"). That is a distinct path: `importCommitTx` hands `assetOutputToVPacket` the original `fundingInputProofs` pointers, which carry no `AdditionalInputs`, since the `proofWithInputs` copy is local to `importOutputProofs`.

To be clear, I could not show a breakage path there — by that point the full file with `AdditionalInputs` is in the archive, which is where the freighter reads input proofs from — so this is a coverage suggestion, not a claimed bug. Pushing some assets to Charlie before the close and switching to `assertDefaultCoOpCloseBalance(false, true)` would close it.

tapchannel/aux_sweeper_test.go
```diff
@@ -0,0 +1,327 @@
...
+	require.NoError(t, err)
+	require.Equal(t, prevIDs[0].OutPoint, lineageInput.OutPoint())
+	require.Equal(
+		t, prevIDs[0].ScriptKey,
+		asset.ToSerialized(lineageInput.Asset.ScriptKey.PubKey),
+	)
+
+	importedOutput, err := importedFile.LastProof()
+	require.NoError(t, err)
+	require.Len(t, importedOutput.AdditionalInputs, 1)
```
litbot-9000, line 198: `TestImportOutputProofsMerge` pins the two-input shape well — lineage is `prevIDs[0]`, exactly one `AdditionalInputs` entry, and it is `prevIDs[1]`. What no test pins is the *single*-input case, which is the commit message's load-bearing regression claim: "A single-input output still encodes byte for byte as it did before, since an empty `AdditionalInputs` is not written."

I verified that claim by reading rather than running it. `proof/proof.go:366` guards the record with `if len(p.AdditionalInputs) > 0`, so `slices.Clone(inputProofFiles[1:])` yielding an empty (non-nil) slice does keep the record off the wire, and single-UTXO channels are byte-identical. But given that this PR exists precisely because a shape nobody tested turned out to be broken, a single-input case asserting `AdditionalInputs` is empty and that the appended bytes equal `proofToImport.Bytes()` looks like the cheapest available insurance on the common path.

(For what it's worth, the merge test uses `proof.MockVerifierCtx` and `NewMockProofArchive()`, so it pins the assembled *shape* but not that the real VM accepts it. That is a reasonable split — the itest covers the latter end to end.)

tapchannel/aux_sweeper.go
```diff
@@ -1755,8 +1923,10 @@ func (a *AuxSweeper) importCommitTx(req lnwallet.ResolutionReq,
 		ChainLookupGen: a.cfg.ChainBridge,
 		IgnoreChecker:  a.cfg.IgnoreChecker,
 	}
+	proofCtx, cancel := lnutils.ContextFromQuit(a.quit)
```
litbot-9000, line 1929: This is the right context — `a.quit` is closed only in `Stop()` (line 195) — and there is no goroutine leak: `lnutils.ContextFromQuit`'s watcher selects on `ctx.Done()` as well as `quit`, and the `defer cancel()` here retires it.

Small gap relative to the commit's own goal, though. `proofCtx` is used for `importOutputProofs`, but the `updateProofsFromShortChanID` call twelve lines below (line 1938) still passes `ctxb`, and so does `AddrBook.InsertScriptKey` at line 1892. `updateProofsFromShortChanID` performs a chain RPC (`GetBlockByHeight`), so it is a second place on the same `contractResolver` goroutine that cannot be unwound on shutdown — the exact hazard the commit message describes. `proofCtx` is already in scope; swapping it in for those two is a strictly-better one-word change.

tapchannel/aux_sweeper.go
```diff
@@ -1559,16 +1713,30 @@ func importOutputProofs(ctx context.Context, scid lnwire.ShortChannelID,
...
-		// New funding output, then import it into our archive.
-		//
-		// TODO(roasbeef): way to do this w/o decoding again?
-		var proofFile proof.File
-		err = proofFile.Decode(bytes.NewReader(prefixProof.Blob))
+		// The first input remains the direct proof-file lineage.
+		// Embed the remaining input proof files in the transition proof
+		// so the VM can verify the full virtual transaction.
+		proofWithInputs := *proofToImport
+		proofWithInputs.AdditionalInputs = slices.Clone(
```
litbot-9000, line 1723: I went looking for an ordering bug here — `inputProofFiles[1:]` places the non-primary files into `AdditionalInputs` in witness order, and it would be easy for the verifier to expect some other order. It does not. `verifyAssetStateTransition` builds `prevAssets` as a `commitment.InputSet`, i.e. a map keyed by an `asset.PrevID` recomputed from each `inputProof.Verify()` result (`proof/verifier.go:846-866`), so `AdditionalInputs` order is irrelevant. The primary must still be the file lineage, because that entry alone is keyed off `p.PrevOut` (`proof/verifier.go:827`) — which is exactly what the `idx == 0` check at line 1506 enforces, and `Asset.Witnesses()[0]` is by definition what `PrimaryPrevID()` returns (`asset/asset.go:1829-1848`).

Two further things I checked rather than assumed: `Witnesses()` unwraps to `rootAsset.PrevWitnesses` when the output carries a split commitment witness, so split funding outputs do enumerate their real inputs; and the shape assembled here matches what `VerifyProofSuffix` builds from the same data (`proof/verifier.go:203-265`) — primary becomes lineage, the rest become `AdditionalInputs`. I grepped the surrounding sweeper code and found no remaining index-0 assumption on this path.

Recording this rather than flagging it: 3+ inputs, inputs from distinct anchor transactions, and nested provenance (an input that is itself a merge) all fall out of the map-keyed lookup and the recursive `inputProof.Verify()`, so I could not construct a merge shape this misses.

itest/custom_channels/single_asset_multi_input_test.go
```diff
@@ -140,4 +141,25 @@ func testCustomChannelsSingleAssetMultiInput(ctx context.Context,
...
+	// Cooperatively close the multi-input channel. The non-initiator must
+	// fetch both funding input proof files to verify and import the merged
+	// funding output before it can finalize the close.
+	chanPoint := &lnrpc.ChannelPoint{
+		OutputIndex: uint32(fundRespCD.OutputIndex),
+		FundingTxid: &lnrpc.ChannelPoint_FundingTxidStr{
+			FundingTxidStr: fundRespCD.Txid,
+		},
+	}
+	closeAssetChannelAndAssert(
```
litbot-9000, line 154: This is the most important artifact in the PR — #2243 happened because no itest closed a multi-input channel — so I checked empirically that it is load-bearing rather than trusting the argument. **It is.**

I cherry-picked this commit (`f0f50d64`, 22 insertions, this file only) onto `main` at `1391a80f` with none of the fix commits, confirmed the fix was absent (`fetchInputProofFiles` grep = 0; the old `Asset.PrimaryPrevID()` still at `aux_sweeper.go:1494`, under the `TODO(roasbeef): assume single asset for now, also additional inputs`), rebuilt the itest binaries from that tree, and ran it both ways:

- **with the fix** (`b29cb0a7`): `--- PASS: TestCustomChannels/single_asset_multi_input (23.28s)`
- **without the fix** (`main` + this commit alone): `--- FAIL: TestCustomChannels/single_asset_multi_input (51.87s)`

And it is the right failure. It fires at `helpers.go:1083` — the `wait.Predicate` inside `assertClosedChannelAssetData` — reached from `helpers.go:1522`, which is the `remote` call, i.e. **Charlie, the responder**: `predicate not satisfied after time out`. Charlie's tapd log reproduces #2243 verbatim:

```
[ERR] CNCT chain_watcher.go:1887: Failed to handle commit spend: handle coop close:
finalize coop close: unable to import output proofs: unable to verify proof:
missing asset input(s): unable to make virtual txIn
(PrevID(outpoint=99a1317bf94701a27ea1bb81dfcc921c5d7dfe68d86f8913670cd2876d3baf7b:2,
id=6fb0c6e5b611794efae2efc6cb9b48f31320c1b79952dda2ef5ed9451fd878e0,
script_key=02c0757acf723219072abdd7f71b65f0c317e59df3b357ac67478762dee8ad4049))
```

Dave funds, so Charlie is the responder and the side that has to fetch both input proof files — the correct side is under test.

litbot-9000 (commented):
Reviewed all four commits at `b29cb0a7`, built and ran the tree, and — because the root cause of #2243 was specifically that no itest closed a multi-input channel — checked empirically that the new itest fails without the fix. It does, with the exact production error from the issue. The fix itself looks sound to me: I traced the merge shape against the verifier instead of trusting the claim, and `AdditionalInputs` order genuinely does not matter because `prevAssets` is a map keyed by `PrevID`; the single-input path really is byte-identical because the TLV record is guarded on a non-empty slice; and mid-import cancellation cannot strand partial state because each output is gated by its own `HasProof` check. I found nothing I would block on — no accepted-invalid funding output, no merge shape still broken, no partial persisted state, and an itest that demonstrably catches the bug. Six inline notes follow; four are minor, two just record what I verified.

## Peer-supplied input validation

This is where I spent most of the review. Case by case, against `fetchInputProofFiles` and the new pre-loop block in `importOutputProofs`:

| Case | Verdict |
|---|---|
| Empty output-proof list | Checked — `aux_sweeper.go:1630` |
| Nil output proof | Checked — `1466` (fetch), `1637` (import) |
| Nil script key on output | Checked — `1642` |
| Empty witness list | Checked — `1474` |
| Nil `PrevID` in a witness | Checked — `1494` |
| Duplicate inputs within one output | Checked — `1500` |
| Duplicate inputs across outputs | Checked — `1660`, cross-output map |
| Unbounded list (resource exhaustion) | Bounded — `maxFundingInputProofs` = 64, per-output at `1477` and globally at `1650`. Same bound the funding controller already enforces (`aux_funding_controller.go:609,1740`) and at the same `PrevID` granularity, so no honest channel is newly rejected |
| Inputs that do not exist | Checked — courier fetch fails, then `LastProof()` PrevID equality at `1610` |
| Primary must be the file lineage | Checked — `1506`, which is precisely `PrimaryPrevID()` |
| Fetched file terminates in the input requested | Checked — `1610` |
| Output proof already carrying `AdditionalInputs` | Checked — `1469` |
| Input amounts summing to the funding output | **Not checked here** — VM enforces conservation in `engine.Execute()` |
| Input of a different asset ID / group key | **Not checked here** — VM, plus locator mismatch at the courier |
| Peer does not own the input | Not checked here — enforced by VM witness validation over the virtual tx |
| Self-referential / cyclic provenance | Covered — each input file is fully verified recursively via `inputProof.Verify()`; a cycle has no valid genesis chain |

The three "not checked here" rows are fine, and the reason is worth stating plainly because it sets the severity ceiling for *every* gap in this function: `MultiArchiver.ImportProofs` calls `proofVerifier.Verify` (`proof/archive.go:1012`) with no chain-verification skip, so the assembled file gets a full verification pass, and the funding output asset — witness list included — is bound to the confirmed anchor transaction by its inclusion proof. The witness list is therefore **not peer-malleable** at this point: a peer cannot under- or over-claim inputs without failing verification. Every gap above degrades to a rejected-but-honest close (annoying, recoverable on restart), never to an accepted-invalid funding output. I could not find an input that produces the latter.

Measured against the established standard the list is essentially complete: it mirrors `VerifyProofSuffix` (`proof/verifier.go:159-265`) check for check — no `AdditionalInputs` on the suffix, `PrevOut` == primary, per-input dedup, `TxSpendsPrevOut` per input, `LastProof()` PrevID equality, primary-as-lineage. The one structural weakness is that `TxSpendsPrevOut` is evaluated before the anchor transaction has been bound to the chain; that is the first inline comment, and low severity for the reasons given there.

One framing correction. The PR description says this import "is the only point where the responder checks the funding inputs it was given". That is not quite right on current `main`: `ChannelReady` -> `validateConfirmedFundingProofs` (`aux_funding_controller.go:2841`) already binds each suffix to the funding outpoint and requires the confirmed transaction to spend every claimed input exactly once. The new checks are defense in depth — a fine thing to be, and they do turn an opaque VM error into a precise one. They are just not the only line, and that matters for how much weight the reorder in comment 1 deserves.

## Does the new itest catch the bug?

**Yes — verified empirically, both directions.** Full detail and the verbatim log are in the inline comment on the `closeAssetChannelAndAssert` call.

- With the fix (`b29cb0a7`): `--- PASS: TestCustomChannels/single_asset_multi_input (23.28s)`
- `main` (`1391a80f`) + `f0f50d64` cherry-picked alone, fix confirmed absent: `--- FAIL ... (51.87s)`, timing out in `assertClosedChannelAssetData` for the **responder**, with Charlie's log showing `finalize coop close: unable to import output proofs: unable to verify proof: missing asset input(s)` — the same error chain as #2243.

## What I verified

Ran, on `b29cb0a7` unless noted:

- `make build` — pass, 1m40s
- `go test ./tapchannel/...` — pass, 0.754s. The three new tests pass individually: `TestImportOutputProofsMerge`, `TestFetchInputProofFilesRejectsMalformed` (7 subcases), `TestFetchInputProofFilesRejectsMismatchedProof`
- `go test ./proof/...` — pass, 26.172s
- `go test -race ./tapchannel/` — pass, 7.439s (one sequential run; 27G free)
- `make build-itest` — pass, 4m0s
- `make itest-cc cccase=single_asset_multi_input` — pass on the PR; fail on `main` + the itest commit alone, as above

Read rather than ran, to check claims I did not want to take on trust: the `AdditionalInputs` encode guard (`proof/proof.go:366`), `prevAssets` map construction (`proof/verifier.go:825-870`), `Witnesses()` / `PrimaryPrevID()` split handling (`asset/asset.go:1829-1848`), `AppendProof` vs `AppendProofRaw` (`proof/file.go:382-421` — neither size-checks, so the new explicit `FileMaxProofSizeBytes` check is a strict addition, not a relaxation), `UpdateTransitionProof` (`proof/append.go:142-160`), and `lnutils.ContextFromQuit`.

Could not run: anything needing a container runtime. There is no docker, podman or psql in this environment, so the `test_db_postgres` / `ory/dockertest` paths have **no** coverage from me — I am not implying otherwise. CI covers them, and all 42 checks on `b29cb0a7` are green, including both Postgres unit jobs and the itest suites, which is consistent with my local results. Head SHA re-confirmed as `b29cb0a74a58a9f9ed991e458fee2a03f1449e51` immediately before posting.

Per-commit: commit 1 is the substance and is reviewed above. Commit 2 is correct in isolation (right context, no leak) with one loose end noted inline. Commit 3 is the itest and is load-bearing. Commit 4 is a release note and reads accurately.

## Findings

Most severe first; each sits inline on the code it concerns.

1. `tapchannel/aux_sweeper.go:1514` — the new per-input `TxSpendsPrevOut` check runs against the peer-supplied `AnchorTx`, before `updateProofsFromShortChanID` binds it to the chain at line 1708. Low: transitively sound for channels that passed the (unreleased) `ChannelReady` binding, but not for channels opened under v0.8.1/v0.8.2. A one-line reorder makes it self-contained.
2. `itest/custom_channels/single_asset_multi_input_test.go:164` — the responder holds no asset balance, so the test covers merged-funding *import* but not merged-funding *sweep*, which is the user-visible symptom in #2243. Coverage suggestion; I could not show a breakage path.
3. `tapchannel/aux_sweeper_test.go:198` — nothing pins the single-input "byte for byte as before" regression claim, on the path that was already working.
4. `tapchannel/aux_sweeper.go:1926` — the shutdown-cancellation fix leaves `ctxb` on the `updateProofsFromShortChanID` call twelve lines below, and on `InsertScriptKey` above.

Two further inline comments record verification rather than a defect: `aux_sweeper.go:1720` (why the merge shape is correct and order-independent, including the 3+ input and nested-merge cases) and `single_asset_multi_input_test.go:154` (the without-the-fix run).

tapchannel/aux_sweeper.go
jtobin: Fixed in updated changeset.

tapchannel/aux_sweeper.go
jtobin: Fixed in updated changeset.

tapchannel/aux_sweeper.go
```diff
@@ -1755,8 +1926,10 @@ func (a *AuxSweeper) importCommitTx(req lnwallet.ResolutionReq,
 		ChainLookupGen: a.cfg.ChainBridge,
 		IgnoreChecker:  a.cfg.IgnoreChecker,
 	}
+	proofCtx, cancel := lnutils.ContextFromQuit(a.quit)
```
GeorgeTsagk, line 1929: blocking: The context from commit "cancel funding proof import on shutdown" never cancels in practice: a.quit is only closed in Stop(), but the CAS there (line 189) is inverted. stopped starts false and nothing sets it to true, so CompareAndSwap(true, false) always fails and Stop() returns before close(a.quit). The bug pre-dates this PR, but the commit is a no-op until this is:

```go
if !a.stopped.CompareAndSwap(false, true) {
```

Suggest including that flip in this PR since the commit message claims the shutdown hang is fixed.

tapchannel/aux_sweeper.go
```diff
@@ -1464,14 +1627,50 @@ func importOutputProofs(ctx context.Context, scid lnwire.ShortChannelID,
...
+
+		witnesses := outputProof.Asset.Witnesses()
+		if len(witnesses) > maxFundingInputProofs-totalInputs {
+			return fmt.Errorf("too many funding inputs, max is %d",
+				maxFundingInputProofs)
+		}
+		totalInputs += len(witnesses)
+
+		for _, witness := range witnesses {
+			if witness.PrevID == nil {
```
GeorgeTsagk, line 1657: nit: A witness with nil PrevID is skipped here but is a hard error in fetchInputProofFiles. Harmless since any non-imported proof errors at fetch time, but erroring here too would keep the two checks consistent.

tapchannel/aux_sweeper.go
```diff
@@ -1455,6 +1456,168 @@ func (a *AuxSweeper) importOutputScriptKeys(desc tapscriptSweepDescs) error {
...
+
+		requests[idx] = inputProofRequest{
+			prevID: *proofPrevID,
+			recipient: proof.Recipient{
+				ScriptKey: scriptKey,
+				AssetID:   proofPrevID.ID,
+
+				// PrevID has no amount. This field is only for
+				// logging, so use the output amount.
+				Amount: outputProof.Asset.Amount,
```
GeorgeTsagk, line 1551: nit: Recipient.Amount is the merged output amount for every input, so the per-input fetch logs will overstate each input's amount. Fine if intentional, but maybe drop it or note it in the log line instead.

tapchannel/aux_sweeper.go
```diff
@@ -1491,84 +1690,56 @@ func importOutputProofs(ctx context.Context, scid lnwire.ShortChannelID,
...
-		// Before we combine the proofs below, we'll be sure to update
-		// the transition proof to include the proper block+merkle proof
-		// information.
-		err = updateProofsFromShortChanID(
-			ctx, chainBridge, scid, []*proof.Proof{proofToImport},
+		// The first input remains the direct proof-file lineage.
+		// Embed the remaining input proof files in the transition proof
+		// so the VM can verify the full virtual transaction.
+		proofWithInputs := *proofToImport
+		proofWithInputs.AdditionalInputs = slices.Clone(
```
GeorgeTsagk, line 1723: nit: slices.Clone looks unnecessary: nothing mutates inputProofFiles[1:] afterwards, only inputProofFiles[0] is appended to.

tapchannel/aux_sweeper.go
```diff
@@ -1455,6 +1456,168 @@ func (a *AuxSweeper) importOutputScriptKeys(desc tapscriptSweepDescs) error {
 	)
 }
 
+// fetchInputProofFiles fetches and validates the proof file for every input
+// spent by an output proof. The first returned file is the output asset's
+// direct lineage; the remaining files prove the merge's additional inputs.
+func fetchInputProofFiles(ctx context.Context, outputProof *proof.Proof,
```
GeorgeTsagk, line 1462: optional: Much of this preflight (duplicate detection, TxSpendsPrevOut, PrevOut vs primary input, file-tip identity check) duplicates what proof.VerifyProofSuffix already enforces. Understandable since the import path needs the assembled file anyway, just flagging that these invariants now live in two places.

GeorgeTsagk (changes requested):
Went through all four commits, built the branch, ran the unit tests and the extended itest, and also re-ran the itest against a build with the first commit reverted to confirm it reproduces the original failure. The fetch logic is sound: the ordering assumption (file 0 as lineage, rest as AdditionalInputs) matches the invariants VerifyProofSuffix enforces, single-input channels produce byte-identical proof files to the old code, and final soundness rests on the full verification in ImportProofs. Requesting changes for one item: the shutdown commit doesn't actually take effect, see inline comment.

tapchannel/aux_sweeper.go
jtobin: Fixed this one, plus an adjacent one that Fable found in FundingController.Stop.

tapchannel/aux_sweeper.go
jtobin: wontfix for expediency.

tapchannel/aux_sweeper.go
jtobin: wontfix for expediency.

tapchannel/aux_sweeper.go
jtobin: wontfix for expediency.

tapchannel/aux_sweeper.go
jtobin: wontfix for expediency.

GeorgeTsagk (approved):
Lgtm

PR #2247, opened 2026-08-21
tapgarden: hardening phase three (remaining guarantees)

jtobin:
This is the next chunk of the large-scale tapgarden hardening work originally contained in #2153 (following #2190 and #2203). It handles the following bullets from the original PR description:

> **Restart-idempotence of side effects**. Previously it was possible for a crash between persisting a batch as broadcast and importing its taproot output to leave the on-chain output unknown to the wallet forever; now, because import precedes the state-transition write, on-disk state can never outpace the side effect it commits.
>
> **Event deduplication**. Previously it was possible for a restart-triggered re-fire of a mint event to log the same logical event twice and double-count the supply-commit update stream; now, because each event is keyed by a content hash with a uniqueness constraint, the second insert is silently absorbed by the schema.
>
> **No orphaned state-machine transitions**. Previously it was possible for a deduplicated event to leave an empty supply-commit transition wedged in UpdatesPending; now, because the insert transaction rolls back whenever it detects a no-op dedupe, no empty transition can ever land.
>
> **Snapshot isolation**. Previously it was possible for a subscriber holding a "snapshot" of a batch to share substructure with the live caretaker and observe mid-flight mutations through it; now, because copies are genuinely deep, snapshots are real isolation barriers.
>
> **Request/response binding**. Previously it was possible for a cancel response to land on a different cancel request than the one that produced it, because the matching depended on a serialization discipline rather than the protocol itself; now, because each request carries its own reply channel, the binding is intrinsic.
>
> **Shutdown safety**. Previously it was possible for a caretaker reaching completion in the same instant the planter was stopping to deadlock the entire shutdown path; now, because the completion send is abandonable on Quit, no notification can outlive its receiver.

It contains the following changes (summary ctsy Fable):

**Event dedup, schema-side.** `supply_update_events` gains a 32-byte `event_key` column -- `sha256(group_key || update_type || event_data)` -- guarded by a unique index, and the insert becomes an `ON CONFLICT DO NOTHING` upsert, so a re-fired mint event is absorbed rather than double-counted. Migration 000062 adds the column and index; programmatic migration 000063 backfills hashes for pre-existing rows (SQLite has no native SHA-256) and drops any legacy duplicates already present, preferring rows attached to a finalized transition so no transition loses the events it depends on.

**No orphaned transitions.** A deduped insert previously still created its supply-commit transition, leaving an empty transition wedged in `UpdatesPending`. The insert transaction now rolls back entirely when the dedup makes it a no-op, so the caller sees a successful no-op and no empty transition lands.

**Import before broadcast.** In the Committed branch, `Wallet.ImportTaprootOutput` ran after the state-transition write that flips the batch to Broadcast; a crash between the two left the minting output permanently unknown to lnd's wallet. The import (already idempotent) now precedes the write, so a crash anywhere in the branch resumes from Committed and re-runs the sequence cleanly.

**Deep copies.** `MintingBatch.Copy` was "deep" in name only -- seedlings, asset metas, the genesis packet's nested slices, and the root asset commitment all shared substructure with the source. Copies now walk every reachable pointer, slice, map, and buffer; a new test builds a fully-populated batch, mutates every reachable substructure of the source, and asserts the snapshot is unmoved.

**Cancel binding.** Cancel responses flowed over a shared `CancelRespChan`, matched to requests only by a serialization discipline. Each `cancelReq` now carries its own reply channel, so the response is causally bound to the request that produced it regardless of that discipline.

**Abandonable completion.** The caretaker's completion signal was an unbuffered send with no select on `Quit`; a caretaker completing while the planter was tearing down would block forever inside the shutdown `Wait`. The send now selects on `Quit` and is abandoned on shutdown, where nobody needs it anyway.

**Restart-recovery property test.** A rapid property test models the legal `BatchState` transition graph and samples restarts at well-synchronized boundaries (post-Committed disk state; post-publish in the Broadcast branch), asserting every interleaving still drives the mint to `BatchStateFinalized` with no errors and no leftover workers. The harness is structured so new restart points can be added cheaply.

**Plumbing.** Planter state requests collapse from a request-type enum + generics + dispatch switch into typed closures that carry their own response channels; the planter and its per-batch workers share a single `*GardenKit` instead of copying it; and `BatchCaretaker` becomes `Cultivator`, resolving a long-standing TODO and matching the package's naming scheme.


tapdb/supply_commit.go
```diff
@@ -736,11 +783,21 @@ func (s *SupplyCommitMachine) InsertPendingUpdate(ctx context.Context,
...
+		// committing the freshly-created transition would leave an
+		// empty pending transition behind and move the state machine
+		// to UpdatesPendingState with nothing to commit. Return the
+		// sentinel error to roll the whole tx back -- the outer
+		// wrapper turns this into a successful no-op for the caller.
+		rowsAffected, err := insertUpdate(sqlInt64(transitionID))
 		if err != nil {
 			return err
 		}
+		if rowsAffected == 0 {
```
litbot-9000, line 800: **The rollback is invisible to the caller, so the in-memory state machine walks past the DB and wedges until restart.**

`rowsAffected == 0` rolls the whole transaction back and the wrapper at line 830 maps the sentinel to `nil`. But `nil` is exactly what the caller reads as "recorded". `DefaultState.ProcessEvent` (`universe/supplycommit/transitions.go:88-108`) does:

```go
err := env.StateLog.InsertPendingUpdate(ctx, env.AssetSpec, supplyEvent)
if err != nil { ... }
supplyEvent.SignalDone(nil)
return &StateTransition{
        NextState: &UpdatesPendingState{
                pendingUpdates: []SupplyUpdateEvent{supplyEvent},
        },
}, nil
```

There is no branch for "absorbed". So after a deduped insert the FSM is in `UpdatesPendingState` holding one cached update while the DB is still in `DefaultState` with zero transitions.

Concrete path, which is precisely the scenario this commit's message describes:

1. A mint event creates transition T1 and attaches E1; T1 gets committed and finalized; the FSM returns to `DefaultState` in memory and on disk.
2. tapd restarts with the mint batch at `Broadcast`. The cultivator re-runs the Confirmed branch and re-fires `SendMintEvent`.
3. No pending transition exists, so we take *this* branch. The insert dedupes, `rowsAffected == 0`, the tx rolls back, `InsertPendingUpdate` returns `nil`.
4. The FSM moves to `UpdatesPendingState` with the already-committed event cached.
5. The next `CommitTickEvent` calls `FreezePendingTransition`. That query (`queries/supply_commit.sql:138`) is an `:exec` `UPDATE ... WHERE state_machine_group_key = @group_key AND finalized = FALSE` — it matches zero rows and returns **nil**, so nothing stops the FSM.
6. It proceeds into `CommitTreeCreateState` and eventually reaches `CommitState`, which fails at `supply_commit.go:974-981` with `no pending transition found for group key %x`. `ApplyStateTransition` (`:1895-1905`) has the same guard.

That asset group's supply-commit machine cannot make progress until tapd restarts, at which point the state is reloaded from disk and heals.

I tried hard to refute this and the mechanism itself is fine — the problem is only the return type:

- `ExecTx` does **not** retry the sentinel. `tapdb/interfaces.go:268-274` only loops when `IsSerializationOrDeadlockError(dbErr)`; otherwise `done=true` and the error is returned immediately.
- `MapSQLError` (`tapdb/sqlerrors.go:22-42`) passes a non-DB error through unchanged, so the `errors.Is` at line 830 still matches.
- `ON CONFLICT (event_key)` names its target, so a conflict on a *different* constraint surfaces as a real error rather than being mistaken for a dedupe. Good.
- Nothing valuable is discarded by the rollback: the only other work in the tx is `UpsertSupplyCommitStateMachine` (idempotent) and the `InsertSupplyCommitTransition` you *want* gone.

So the fix is small: return the outcome instead of erasing it — a `bool` or an exported sentinel like `ErrSupplyUpdateDeduped` — and have `DefaultState.ProcessEvent` self-transition rather than move to `UpdatesPendingState`. The `// TODO(roasbeef): special error case here?` sitting at `transitions.go:85` is asking for exactly this.

Worth noting the same signal is needed for the *other* branch too (line 724, the existing-pending-transition path). There the comment says dedupe "is fine here", and on disk it is — but the caller in `UpdatesPendingState.ProcessEvent` also appends the event to `pendingUpdates` unconditionally, so the cached slice grows an entry that has no row of its own attached to the current transition.

tapdb/programmatic_migrations.go
```diff
@@ -328,3 +336,74 @@ func insertAssetBurns(ctx context.Context, q sqlc.Querier) error {
...
+	log.Debugf("Backfilling event_key for %d supply update events",
+		len(rows))
+
+	seen := make(map[string]struct{}, len(rows))
+	for _, row := range rows {
+		key := supplyUpdateEventKey(
+			row.GroupKey, row.UpdateTypeID, row.EventData,
+		)
+
+		if _, dup := seen[string(key)]; dup {
```
litbot-9000: **The backfill drops the duplicate attached to the still-pending transition, leaving behind exactly the empty `UpdatesPending` transition that this PR says can never land.**

`FetchSupplyUpdateEventsForBackfill` partitions only on `transition_id IS NULL`; it never looks at `finalized`. So among two attached duplicates, the lower `event_id` wins and the other is deleted here.

Because `supply_commit_transitions_single_pending_idx` (`000040_asset_commit.up.sql:147`, `UNIQUE(state_machine_group_key) WHERE finalized = FALSE`) allows at most one non-finalized transition per group, and it is always the newest, its events always carry the *higher* `event_id`. That means the deletion always lands on the pending transition's row. I checked the reverse — a finalized transition losing the event it depends on — and it is not reachable within a group, so the in-code claim on line 353 does hold. But the PR description's stronger phrasing ("preferring rows attached to a finalized transition so no transition loses the events it depends on") is only true by accident of that index, and the pending side is unaddressed.

Confirmed empirically. Throwaway test on a v61 DB: one state machine at `current_state_id = 2`, transition 1 with `finalized=1`, transition 2 with `finalized=0`, and two byte-identical events, E1 → T1 and E2 → T2. After `ExecuteMigrations(TargetLatest)`:

```
events left on PENDING transition 2: 0
pending transition rows still present: 1
state machine current_state_id: 2
```

The producing state is not hypothetical — it is the signature of the bug this commit exists to fix. A mint creates T1 with E1; T1 finalizes; tapd restarts with the batch at `Broadcast`; the re-fire finds no pending transition and creates T2 with a duplicate E2. Any legacy DB that hit the double-count bug has this shape.

What happens next: `UpdatesPendingState.ProcessEvent` (`universe/supplycommit/transitions.go:168-189`) has no empty-set guard. A `CommitTickEvent` freezes the empty transition and emits `CreateTreeEvent{updatesToCommit: []}`; `CommitTreeCreateState` (`:361`) calls `ApplyTreeUpdates` with an empty slice and goes on to build, sign and broadcast a supply commitment whose root is unchanged. An on-chain transaction that commits nothing, paying fees.

Suggested fix, in the same programmatic migration: after the dedupe loop, delete any non-finalized transition left with zero `supply_update_events` rows, and reset the owning state machine's `current_state_id` to `DefaultState` when you do. That keeps the migration consistent with the invariant the insert path now enforces. Alternatively, prefer keeping the row attached to the *non*-finalized transition when the duplicate set spans both — that is safe here, since the finalized transition's tree contents are already committed on chain and are not re-derived from these rows.

tapgarden/seedling.go
```diff
@@ -112,6 +113,54 @@ type Seedling struct {
...
+// See TestMintingBatchCopyIsDeep for the invariant pinned by tests.
+func (s *Seedling) Copy() *Seedling {
+	if s == nil {
+		return nil
+	}
+	out := &Seedling{
+		AssetVersion:      s.AssetVersion,
+		AssetType:         s.AssetType,
+		AssetName:         s.AssetName,
+		Meta:              s.Meta.Copy(),
```
litbot-9000, line 131: **This puts up to 1 MiB per asset on a path the gardener runs once per `MintAsset` RPC.** Not a blocker, but worth a decision rather than an accident.

`Seedling.Copy` now duplicates `MetaReveal.Data`, which `proof/meta.go:41` caps at `MetaDataMaxSizeBytes = 1024 * 1024`. The old `MintingBatch.Copy` did `seedlingCopy := *v` and `batchCopy.AssetMetas[k] = v`, sharing the `*MetaReveal`, so its cost was independent of meta size.

The call site that matters is pre-existing: `planter.go:2074`, in the gardener's `seedlingReqs` arm, copies the whole pending batch to build the `SeedlingUpdate` response — once per seedling request. It was `planter.go:2084` on `main`, so this PR doesn't add the call; it changes what the call costs.

Measured with a throwaway benchmark on a batch of N seedlings each carrying an M-byte meta:

```
n=10   meta=1KiB     10.7 µs      16 KB/op
n=10   meta=1MiB      2.88 ms   10.5 MB/op
n=50   meta=1MiB     13.8 ms    52.5 MB/op
n=100  meta=1MiB     38.1 ms   104.9 MB/op
n=100  meta=1KiB     82.8 µs     159 KB/op
```

So ~460× at the protocol's meta ceiling. And because the copy happens on *every* seedling request, building an N-asset batch copies O(N²) meta bytes end to end: for N=100 at 1 MiB that is roughly 5 GB allocated and ~1.9 s of pure copying, all of it inside the single-threaded gardener loop, which blocks every other planter request (`PendingBatch`, `ListBatches`, `FundBatch`, `CancelBatch`, `FinalizeBatch`) for the duration.

Two other hot spots inherit the same change: `newAssetMintEvent`/`newAssetMintErrorEvent` (`cultivator.go:1381`, `:1394`) deep-copy the batch on **every** state transition and every error event, and `listBatches` → `fetchFinalizedBatch` (`planter.go:1636`) does it per batch returned.

Cheapest fix that keeps the isolation guarantee: treat `*proof.MetaReveal` as immutable-after-construction and share it, the way `BatchKey`'s `KeyLocator` and `SupplyCommitments` are already treated as immutable in `Copy`'s own doc comment. Nothing in `tapgarden` mutates a `MetaReveal` in place after the seedling is built, so the aliasing this deep copy closes is not one any caller can observe. Failing that, the seedling-response path could hand back a cheap projection (batch key, state, seedling names/amounts) instead of a full batch clone. If you'd rather keep the unconditional deep copy, that's a reasonable call — but it's worth stating explicitly in the release note that per-batch minting throughput now scales with total meta bytes.

tapgarden/caretaker_rapid_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+// restartPoint enumerates the deterministically-observable disk states
+// at which we can simulate a daemon restart in the rapid harness. Each
+// point is anchored to a well-defined synchronization signal (either a
+// disk-state poll or a mock channel send) so the restart is not racy.
+type restartPoint int
+
+const (
+	// rpAfterCommitted: disk state has reached BatchStateCommitted.
+	// Restart re-enters the Committed branch, which re-signs, re-
```
litbot-9000: **`rpAfterCommitted` doesn't re-enter anything, and the property as written passes against a reverted fix.** I put real effort into establishing this because a property test that passes on a broken implementation is worse than no test.

Two problems with this restart point, both traceable to the harness sync helper:

`assertBatchCommitted` waits on `batchCommittedStates.Contains(batch.State())` (`planter_test.go:687`, the set is `{Committed, Broadcast, Confirmed, Finalized}`), so it returns as soon as the *disk state* reaches `Committed`. At that instant the cultivator is parked on the mock wallet's unbuffered `SignPsbtSignal` — i.e. at the very first step of the Committed branch, before any work has been done. So the restart re-does nothing, and the comment on lines 35-37 ("Restart re-enters the Committed branch, which re-signs, re-imports, and re-writes the genesis tx") doesn't describe what happens.

Second, `tapnode/tapnodemock/wallet_anchor.go:182` `ImportTaprootOutput` never returns an `already exists` error — it signals and succeeds unconditionally. So even if the harness *did* land a restart after a completed import, it could not detect a non-idempotent import.

Empirically, with `-race` off and the default 30 checks:

- **Injection A (blindness).** I swapped `ImportTaprootOutput` back to *after* `CommitSignedGenesisTx` in `cultivator.go`, i.e. reverted commit 4's restart-idempotence fix — the headline guarantee this test is described as the capstone for. Result: `ok github.com/lightninglabs/taproot-assets/tapgarden 80.789s`. **Passes.** Neither restart point lands between the import and the state write, so the property cannot see it.
- **Injection B (positive control).** I made `Start()` skip re-creating a cultivator for `BatchStateBroadcast` batches. Result: **FAIL** at `caretaker_rapid_test.go:105` with `timeout hit` — the post-restart re-publish assertion. So `rpAfterPublish` *is* a genuine re-entry test and does catch a real restart-recovery regression.

Both injections reverted; working tree clean.

So: one of the two restart points is real, and the test earns its keep for the `Broadcast` branch. But as the doc comment's own "Scope" paragraph concedes, in-branch crashes are the follow-up — which means the property does not cover the two guarantees this PR leads with, G1 (import before state write) and G2 (mint-event dedupe on re-fire, which needs a restart inside the *Confirmed* branch). The release note's "randomized restart points" is a little generous for two booleans.

Two smaller things while you're here:

- The generator space is exactly two `rapid.Bool()` draws and `numSeedlings` is hardcoded 5, so this is a four-case table test wearing a rapid costume — `rapid`'s shrinking has nothing to shrink. That has a real cost: 30 checks is ~80 s of the package's 121 s (up from ~34 s pre-PR), and `-rapid.checks=120` takes 5m21s to explore the same four cases ~30× each. A `for _, tc := range [4]...` loop would give identical, complete, deterministic coverage in ~11 s. If the intent is to keep `rapid` for when more restart points land, that's fair — but then it'd be worth generating `numSeedlings` too, so the extra iterations buy something.
- `drainErrors` (line 136) is a non-blocking single receive, so it discards at most one error and discards it unconditionally — including a genuine one that happened to be queued at that moment. Given `assertNoError()` at the end is the safety net, consider asserting on what was drained rather than dropping it.
- Naming: this file and `TestCaretakerRestartRecoveryRapid` are named after the type commit 6 removes. `grep -rn BatchCaretaker` is clean tree-wide and no operator-visible string still says "caretaker", so the rename is complete where it counts — but there are 14 non-test comment sites left, several written *by this PR*: `tapgarden/interface.go:186` and `tapdb/asset_minting.go:2293` (both new in commit 5), `tapgarden/batch.go:343`, `tapdb/supply_commit.go:572`, plus `bench/fixture/mint_driver.go`, `tapnode/tapnodemock/chain_bridge.go:40` and `tapreorg/doc.go:9`. Worth one more sweep in commit 6.

litbot-9000 (changes requested):
Reviewed all seven commits at `82ac0885` (re-targeted mid-review; the push between `2af4f83e` and `82ac0885` fixed the `Sqlc check` failure and the stale migration numbers in the query comments, so I've dropped that finding — nice catch on your side, it was on my list). Most of this holds up under attack: the migration does **not** repeat the phase-two brick (I drove a real `NewSqliteStore` over a v61 DB carrying colliding rows and it comes up clean, `dirty=false`), the shared `GardenKit` is genuinely immutable so pointer-sharing introduces no race, the abandonable completion send fixes a real deadlock without losing anything, and the import-before-write reorder is correct in both crash directions. The rename commit is mechanically pure. Two things I'd want changed before this lands, both in commit 1: the dedupe rollback is invisible to the caller, so the supply-commit FSM walks past the DB and wedges until restart; and the 000063 backfill manufactures on legacy data exactly the empty-pending-transition state that the rollback exists to prevent. Separately, making `Copy` deep put a 1&nbsp;MiB-per-asset memcpy on a per-RPC path that was previously pointer-shallow — measured below.

## Guarantees claimed vs. verified

| Guarantee | Verdict |
|---|---|
| **Restart-idempotence of side effects** (import precedes state write) | **Held.** Traced both windows. Crash after import / before `CommitSignedGenesisTx`: disk stays `Committed`, restart re-enters the branch from the top, re-signs (the new itest pins `SignAndFinalizePsbt` determinism), re-derives the same `mintingOutputKey` (`MintingOutputKey` is pure in batch key / commitment root / sibling), and the re-import is absorbed by the `already exists` arm. Crash after the state write: the import has already happened, restart resumes at `Broadcast` and publishes. The cost of the new direction is a stray watch-only P2TR key in `lnd` if the batch is later cancelled — no funds, nothing spendable, strictly better than the old failure. I also checked the ctx reuse: the DB write now runs on the import's *fresh* `WithCtxQuit` deadline, where the old code handed it the older signing ctx, so that got better too. |
| **Event deduplication** (content hash + unique index) | **Held on the hash, held on disk.** I could not construct two genuinely-distinct events that collide. `event_data` is the event's full `Encode()`: `NewMintEvent` writes the entire leaf proof (`states.go:335`), unique per minted asset; `NewBurnEvent` → `BurnProof.Encode`, unique per burn outpoint; `NewIgnoreEvent` → TLV of tuple + sig. `LeafKey`/`MintHeight` are omitted from the hash but are derivable from the proof, so nothing distinguishing is dropped. Reverse direction also checked: the re-fired mint proof is rebuilt deterministically from chain data through TLV encode/decode, which is precisely why the dedupe fires. Verified empirically that a second identical insert returns `rows=0` even with a different `transition_id`. |
| **No orphaned state-machine transitions** | **Partially held.** True for the new-insert path on disk — the rollback does discard the transition. **Broken in two other places:** the in-memory FSM advances anyway (finding 1), and migration 000063 creates an orphaned pending transition on legacy data (finding 2, confirmed empirically). |
| **Snapshot isolation** (`Copy` genuinely deep) | **Held, and the pre-existing aliasing was real.** `TestMintingBatchCopyIsDeep` is a genuine test. Note two things: (a) the justification comment at the `FinalizeBatch` broadcast arm — "the caretaker goroutine continues to mutate `Batch.GenesisPacket` and `Batch.RootAssetCommitment` after this point (Broadcast → Confirmed → Finalized)" — is factually wrong. An exhaustive grep of writes through `b.cfg.Batch` in `cultivator.go` gives only lines 645, 701, 716, 757, 765, all in the Frozen/Committed branches, i.e. strictly *before* `BroadcastCompleteChan <- struct{}{}`. After that send the cultivator only parks in a select and touches state through the atomic. The deep copy is still the right defensive move, but I'd fix the reason, because a future reader may rely on a mutation window that doesn't exist. (b) The cost — finding 3. |
| **Request/response binding** (per-request reply channel) | **Held.** Every `cancelReq` carries its own buffer-1 `resp`; `Cancel(respCh)` writes it exactly once; the two potential readers (`advanceStateUntil` pre-broadcast, `assetCultivator` post-broadcast) are mutually exclusive branches of the same goroutine, and commit 3 correctly deletes the third reader that used to sit inside the conf-notification goroutine. `dispatchStateReq` also now selects on `c.Quit` while awaiting the reply, where the old code did a bare `<-req.resp` — a shutdown mid-request used to hang the caller forever. Unadvertised improvement. |
| **Shutdown safety** (abandonable completion send) | **Held.** The deadlock is real: `gardener()` has `defer c.stopCultivators()`, which only runs *after* the main select has exited on `c.Quit`, and `stopCultivators` → `Stop()` → `Wg.Wait()`; `completionSignals` is unbuffered, so a cultivator parked on that send is never released. Dropping the signal is safe because both `SignalCompletion` call sites (`cultivator.go:369`, `:430`) run strictly after `advanceStateUntil(..., BatchStateFinalized)` has already persisted `Finalized`, and the gardener's only reaction is `Stop()` + map delete — which `stopCultivators` does anyway. No double-send (both sites `return` immediately, and they're mutually exclusive branches). No goroutine leak. |

## Rename commit

Mechanically pure. I normalised the post-commit-5 and post-commit-6 versions of all three files with a case-preserving `BatchCaretaker→Cultivator` / `caretaker→cultivator` substitution and `diff -w`'d them. The only residue: four log strings that lost a `Chain` prefix (`ChainCaretaker(%x)` → `Cultivator(%x)`), the removal of the `// TODO(roasbeef): rename to Cultivator?`, and a handful of comment re-wraps. `supply_commit_test.go` is byte-identical after normalisation. No behaviour change. `grep -rn BatchCaretaker` over the whole tree returns zero hits, and no log, error, or RPC-visible string still says "caretaker". See the note on `caretaker_rapid_test.go` for the leftover identifier residue.

## What I verified

At `82ac0885`, SQLite backend, `/workspace` 21G free:

- `make build` → exit 0 (`tapd-debug`/`tapcli-debug`, `v0.7.0-907`).
- `go test ./tapgarden/... ./tapdb/...` → exit 0 (tapgarden ok 121.3s, tapdb ok 73.7s).
- `go test ./tapdb/ -run TestMigration -v` → 17/17 PASS, including all three new `TestMigration63Backfill*`.
- `go test -race ./tapgarden/` → exit 0, ok 331.1s, **zero** `DATA RACE` reports. One sequential run, prioritised for commit 5 as asked.
- `go test ./tapgarden/ -run TestCaretakerRestartRecoveryRapid -rapid.checks=120` → PASS, 5m21s.
- Legacy-data migration, throwaway test through the real `NewSqliteStore` (`SkipMigrations:false`) on a file DB built at v61 with four byte-identical `supply_update_events` rows: **error `<nil>`, version 63, `dirty=false`, one row surviving.** Same via `ExecuteMigrations(TargetLatest)` from v61 with three colliding rows. The phase-two failure mode is closed.
- Down-migration: `TargetVersion(61)` from latest → `<nil>`, `dirty=false`; re-up → `<nil>`, version 63. I specifically expected SQLite to refuse `DROP COLUMN` on a column named in a `CHECK` constraint; **it does not.** Refuted, not raised.
- Two mutation injections against the property test (both reverted, tree left clean) — see finding 4.
- Throwaway `MintingBatch.Copy` benchmark — see finding 3.
- **No container runtime is available here** (no docker/podman/psql), so nothing under `test_db_postgres` or `ory/dockertest` ran, and I am claiming **no** Postgres coverage of my own. For the Postgres verdict I am leaning on CI, which is green at this head including the `dbbackend=postgres` unit jobs for `tapdb` and `tapgarden`. `migrations_test.go` is untagged and `NewTestDBWithVersion` is build-tag dispatched, so the new migration tests do execute there. I read the new SQL for dialect portability by hand: `postgresSchemaReplacements` rewrites only `BLOB`/`INTEGER PRIMARY KEY`/`TIMESTAMP`/`UNHEX`, `length(bytea)` is defined in PG, `ADD COLUMN ... CHECK` and `NULLS DISTINCT`-by-default both behave as the migration comment claims, and `ON CONFLICT (target) DO NOTHING` with `RowsAffected` returning 0/1 is correct on both drivers. I did not execute any of that.
- Standing caveat: CI being green says nothing about crash windows or migrations against legacy data. No CI job kills a node between two writes. Findings 1, 2 and 4 come from reading and from local injection, not from CI.

## Findings

Four, most severe first, each inline on the code it concerns. Findings 1 and 2 are why this is `REQUEST_CHANGES`; 3 and 4 are not blockers.

1. `tapdb/supply_commit.go:798` — the dedupe rollback is indistinguishable from success, so `DefaultState` advances the in-memory FSM past a DB that has no transition, and the next commit tick fails with `no pending transition found`.
2. `tapdb/programmatic_migrations.go:376` — the 000063 backfill deletes the duplicate belonging to the *pending* transition, leaving an empty non-finalized transition behind. Confirmed empirically.
3. `tapgarden/seedling.go:131` — deep `Meta` copy puts up to 1 MiB per asset on a path the gardener runs once per `MintAsset` RPC; measured 38 ms / 105 MB for a 100-asset batch, O(N²) over a batch build.
4. `tapgarden/caretaker_rapid_test.go:35` — `rpAfterCommitted` does not re-enter anything, and the property test passes with commit 4's reorder reverted.

Also worth a look, folded into the inline comments rather than listed separately: commit 4's message covers only the abandonable send but the commit also carries the entire "import before broadcast" guarantee and a new itest, which is awkward in a series organised one-guarantee-per-commit; and the leftover `caretaker` naming, including in this PR's own new test file.

tapdb/supply_commit.go
jtobin: Fixed in updated changeset.

tapdb/programmatic_migrations.go
jtobin: Fixed in updated changeset.

tapgarden/seedling.go
jtobin: I've addressed this in the updated changeset, though may revert that fix, simply because, morally, a deep copy should be a true deep copy, and the pathological case described here does _not_ represent a typical mint.

I'd probably rather keep the true deep copy semantics for now and consider changing this if anything ever gives us reason to do so.

tapgarden/caretaker_rapid_test.go
jtobin: Fixed in updated changeset.

tapgarden/seedling.go
jtobin: Yeah, reverted, keeping the proper deep copy for now.

GeorgeTsagk (commented):
Read through the whole thing, ran the tapgarden/supplycommit/tapdb tests locally
(pass on sqlite). The three big ones look right to me: import before the state
write, the `event_key` index plus rows-affected rollback, and the abandonable
completion send. I went looking for a legitimate pair of events that would
collide on the content hash and could not find one, and dangling events still
get drained at `CommitFinalizeState`, so I don't think the rollback strands
anything.

Six things I'd want to see before merge, roughly in order:

1. `programmatic_migrations.go:367` pulls every `supply_update_events` row into
   memory in one tx, `event_data` included. That's a full issuance proof per
   mint event, so a big issuer could spike a few hundred MB during startup
   migration. The `seen` set only needs hashes, so paging by `event_id` should
   be easy.

2. `FundedMintAnchorPsbt.Copy` and `MintingBatch.Copy` now panic
   (planter.go:3679/3685/3693, batch.go:168). `Copy` runs in the gardener
   goroutine on `PendingBatch`/`SealBatch`/`FinalizeBatch` and on every mint
   event publish, and it used to be infallible. `psbt.NewFromRawBytes` also
   runs `SanityCheck`, so the round trip adds a failure mode the field-wise
   copy didn't have. I'd rather return an error (or log and fall back) than
   crash the daemon on a read path.

3. Two small holes in the migration's dedup preference:
   `supply_commit.sql:123` filters `frozen = FALSE`, so a duplicate-emptied
   frozen non-finalized transition doesn't get healed, and those do survive
   restarts. And `supply_commit.sql:102` prefers attached rows but not
   finalized ones, while the comment on `backfillSupplyUpdateEventKeys` says it
   prefers finalized. It works out today because the finalized row tends to
   have the lower `event_id`, but `finalized DESC` in the ORDER BY plus
   dropping the `frozen` filter would make it match the stated intent.

4. Coverage is thinnest where the copy code is newest. `batch_test.go:242` nils
   out `GenesisPacket` and `RootAssetCommitment` before the strict reflection
   walk, and `TestMintingBatchCopyIsDeep` only mutates `UnsignedTx.LockTime`,
   never the root commitment or the input/output slices. So the psbt round trip
   and `TapCommitment.Copy` are the least asserted parts.
   `PreCommitmentOutput.Copy` isn't covered. A mutation on an input's
   `PartialSigs` and one on a commitment leaf would cover the real risk.

5. The rapid test can't shrink: the harness holds the outer `t`, so a `require`
   failure inside the callback does `FailNow` on the parent instead of failing
   the iteration. Threading `rt` through would give you the shrinking the doc
   comment promises. Also, `cultivator_rapid_test.go:188` writes the global
   `rapid.checks` flag from a `t.Parallel()` test. Harmless while this is the
   only rapid test in the package, but still a global write.

6. Release notes: "fixes several latent bugs in asset minting" doesn't tell an
   operator that migration 63 deletes duplicate rows and can delete a pending
   transition and reset a state machine to `DefaultState`. Worth spelling out.

Nits: `batch.go:329/339/344` still reference the `MintingStore` interface this
PR deletes; `event_key` stays nullable with nothing enforcing it post-backfill
(fine while `InsertPendingUpdate` is the only writer, but a future writer would
silently skip dedup); `assertNumCaretakersActive` kept the old name.

jtobin:
@GeorgeTsagk All addressed. 💪

GeorgeTsagk (approved):
LGTM

jtobin:
Ah, looks like another PR grabbed the migration number. Will rebase..

PR #2248, opened 2026-08-21
tapdb: cap default rapid property-test iterations at 25

jtobin:
(Fable's summary.)

The multiverse coalescer and reconcile property tests added in #2194 dominate the package's race-detector runtime at rapid's default of 100 iterations: both persist store state across iterations, so per-iteration cost grows with the accumulated universes, and the pair costs roughly 410s under -race, with TestReconcileMultiverseProps alone at ~330s. That roughly doubled tapdb's unit-race package time in CI (~500s to ~1000s), and made it noisy besides: rapid seeds from the wall clock, so the workload differed from run to run.

At 25 iterations the pair costs ~30s. An explicit -rapid.checks flag still overrides the default for deeper runs.

### DIFF
--- a/tapdb/main_test.go
+++ b/tapdb/main_test.go
@@ -1,0 +1,33 @@
+package tapdb
+
+import (
+	"flag"
+	"os"
+	"testing"
+)
+
+// defaultRapidChecks caps the number of iterations rapid property
+// tests run per check. The library default of 100 dominates the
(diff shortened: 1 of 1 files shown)


jtobin:
Note that unit-race(0) finished in 15m, so this indeed returns the job's duration to baseline.

GeorgeTsagk (approved):

PR #2251, opened 2026-08-24
tapdb: batch per-asset witness and proof queries

### LINKED ISSUE #2249: tapdb: per-asset witness and proof queries dominate asset loading
### Summary

Materialising assets from the DB issues one query per asset for witnesses, and
one per asset for proofs. On a node holding 1338 assets this is ~19% of the CPU
of a single send, and it affects every path that loads assets, not just sends.

### Where

`fetchAssetWitnesses` loops over the asset IDs it is given, despite the comment
saying it fetches witnesses for "the set of passed asset IDs":

```go
// tapdb/assets_store.go:564 (main), :574 (v0.8.1)
func fetchAssetWitnesses(ctx context.Context, db ActiveAssetsStore,
	assetIDs []int64) (assetWitnesses, error) {

	assetWitnesses := make(map[int64][]AssetWitness)
	for _, assetID := range assetIDs {
		witnesses, err := db.FetchAssetWitnesses(ctx, sqlInt64(assetID))
```

The query already supports batching: `asset_id` is nullable and NULL means "all".

```sql
-- tapdb/sqlc/queries/assets.sql, FetchAssetWitnesses
WHERE (
    (assets.asset_id = sqlc.narg('asset_id')) OR (sqlc.narg('asset_id') IS NULL)
)
```

So this one is a call-site change: fetch once, group by `asset_id` in Go. A
`WHERE asset_id IN (...)` variant would be tighter still if fetching all is too
broad.

`FetchAssetProof` is the harder half. Its script key filter is required, so it can
only ever return one asset's proofs and cannot be batched without a new query:

```sql
-- tapdb/sqlc/queries/assets.sql
WHERE script_keys.tweaked_script_key = $1
```

### Measurement

30s CPU profile of a single `tapcli assets send` on a node with 1338 assets,
sqlite backend, 85 MB database. Attributed to owned code with
`pprof -show=lightninglabs/taproot-assets`:

```
  44.0%  5.42s  tapdb/sqlc.(*Queries).QueryAssets            <- see the linked issue
  13.0%  1.60s  tapdb/sqlc.(*Queries).FetchAssetWitnesses
   5.7%  0.70s  tapdb/sqlc.(*Queries).FetchAssetProof
   5.2%  0.64s  mssmt.(*BranchNode).NodeHash
   2.8%  0.63s  tapdb.dbAssetsToChainAssets
```

Unattributed: 57% of total CPU sits in `modernc.org/sqlite`'s
`_sqlite3VdbeExec`, which is where the per-asset queries land.

The 1.6s and 0.7s are per-call costs multiplied by the row count, not single
expensive queries.

### Environment

- tapd `v0.8.1`, sqlite backend, regtest
- The binary carries one added import, `_ "net/http/pprof"`, so that `--profile`
  serves anything (it otherwise only registers a catch-all redirect to
  `/debug/pprof`, which nothing handles). No other change.
- Node state: 1338 assets, 5 groups, 3692 universe leaves, 7698 proof files
  (238 MiB), 85 MB sqlite database, accumulated over ~500 sends.

### Caveats

- The holding profile is unusually fragmented: most of the 1338 assets are small
  change outputs. The scaling shape holds regardless, but the absolute numbers
  reflect that.
- Only the sqlite path was profiled. `modernc.org/sqlite` is pure Go and slower
  than the cgo driver, so some of the 57% is driver cost rather than query count.
  A postgres comparison would separate the two.

GeorgeTsagk:
Loading assets from the database issued one witness query per asset, and
reconstructing input commitments issued one proof query per asset. That cost
dominates every path that materialises assets on a node holding many assets.

### Changes

- `FetchAssetWitnesses` takes a set of asset primary keys instead of one, and
  the rows are grouped in Go. Chunked at 512 IDs, since SQLite caps bind
  parameters at 32766 and Postgres at 65535.
- `queryCommitments` fetches proofs through a new `FetchAssetProofsByIDs`
  query. The proof is only read for the anchor's alt leaves, which commit to
  the anchor output as a whole, so one proof per distinct anchor point is
  enough instead of one per asset.
- Anchor points already handled are skipped in the commitment reconstruction
  loop. The result maps are reset at the top of the `ExecTx` closure, since it
  may be retried wholesale.

### Speedups

Measured against a copy of a real regtest node: sqlite, 2461 assets, dominant
asset group holding 621 coins.

| operation | before | this PR | with #2252 too |
|---|---|---|---|
| coin selection, 1-unit group send | 16.57s | 1.55s | 68ms |
| coin selection, 181-coin send | 16.56s | 1.53s | 447ms |
| `FetchAllAssets` (2461 assets) | 802ms | 244ms | 236ms |
| `FetchCommitment` | 33ms | 24ms | 25ms |

This also speeds up callers outside the send path: `ListAssets`, `FetchAsset`,
`ListUtxos`, `ExportAssetWalletBackup`, `ProveAssetOwnership`,
`AnchorVirtualPsbts`, and the Prometheus asset balance collector, which pays
the `FetchAllAssets` cost on every scrape.

#2252 bounds how many coins input selection loads in the first place, so the
two are complementary. End to end with both, a v2 address send on that node
goes from 28.1s to 10.8s median.

### Behavior

Checked by diffing full dumps of every asset returned (amount, anchor
outpoint, script key, witness count) and of the selected coin sets, baseline
against this PR: byte identical.

One deliberate change: `queryCommitments` used to read every matching asset's
proof, so it failed when any of them was missing. It now reads one proof per
anchor point, the one belonging to the lowest asset primary key there, so a
missing proof only fails coin selection when it is that asset's. For any other
asset at the same anchor point it surfaces later, when the funding path fetches
input proofs.

### Testing

New tests for multi-asset witness grouping and ordering, for two assets
sharing an anchor point, and for both missing-proof cases (the representative's
proof and another asset's). Unit tests pass on sqlite and postgres, `make lint` and
`make sqlc-check` are clean, and itests pass (`basic_send_unidirectional`,
`address_v2_self_send`, `multi_input_send_non-interactive_single_ID`). Every
commit builds and compiles its unit tests standalone.

Closes #2249

### DIFF
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -278,10 +278,9 @@ type ActiveAssetsStore interface {
 	// database.
 	UpsertAssetWitness(context.Context, PrevInput) error
 
-	// FetchAssetWitnesses attempts to fetch either all the asset witnesses
-	// on disk (NULL param), or the witness for a given asset ID.
-	FetchAssetWitnesses(context.Context, sql.NullInt64) ([]AssetWitness,
-		error)
+	// FetchAssetWitnesses fetches the witnesses for the assets identified
+	// by the passed set of asset primary keys.
+	FetchAssetWitnesses(context.Context, []int64) ([]AssetWitness, error)
 
 	// FetchManagedUTXO fetches a managed UTXO based on either the outpoint
 	// or the transaction that anchors it.
@@ -565,27 +564,61 @@ type AssetHumanReadable struct {
 // input (witness) information.
 type assetWitnesses map[int64][]AssetWitness
 
+// queryIDChunkSize is the maximum number of IDs we pass to a single batched
+// query. Both SQLite and Postgres limit the number of bind parameters a
+// statement may carry (32766 and 65535 respectively), so larger ID sets are
+// queried in chunks of this size.
+const queryIDChunkSize = 512
+
+// forEachIDChunk calls the given function for each chunk of at most
+// queryIDChunkSize of the given IDs. Queries that take a set of IDs must be
+// issued in chunks, as the number of bind parameters a statement may carry is
+// limited. An empty ID set results in no calls at all, which also keeps us
+// from issuing a query with an empty ID set, which no such query supports.
+//
+// The IDs must be distinct. A duplicate ID that lands in a different chunk
+// than its twin is queried twice, so a caller that accumulates the rows of
+// each chunk would count that ID's rows twice.
+func forEachIDChunk(ids []int64, f func(chunk []int64) error) error {
+	for start := 0; start < len(ids); start += queryIDChunkSize {
+		end := min(start+queryIDChunkSize, len(ids))
+
+		if err := f(ids[start:end]); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // fetchAssetWitnesses attempts to fetch all the asset witnesses that belong to
 // the set of passed asset IDs.
 func fetchAssetWitnesses(ctx context.Context, db ActiveAssetsStore,
 	assetIDs []int64) (assetWitnesses, error) {
 
 	assetWitnesses := make(map[int64][]AssetWitness)
-	for _, assetID := range assetIDs {
-		witnesses, err := db.FetchAssetWitnesses(
-			ctx, sqlInt64(assetID),
-		)
+
+	// We fetch the witnesses of many assets per query and then group them
+	// by the asset's primary key. The query orders by asset ID and witness
+	// index, so the witnesses of each asset retain their original order. A
+	// genesis asset has no witnesses stored, so it won't appear in the
+	// map, which'll give it the genesis witness.
+	err := forEachIDChunk(assetIDs, func(chunk []int64) error {
+		witnesses, err := db.FetchAssetWitnesses(ctx, chunk)
 		if err != nil {
-			return nil, err
+			return err
 		}
 
-		// We'll insert a nil witness for genesis asset, so we don't
-		// add it to the map, which'll give it the genesis witness.
-		if len(witnesses) == 0 {
-			continue
+		for _, witness := range witnesses {
+			assetWitnesses[witness.AssetID] = append(
+				assetWitnesses[witness.AssetID], witness,
+			)
 		}
 
-		assetWitnesses[assetID] = witnesses
+		return nil
+	})
+	if err != nil {
+		return nil, err
 	}
 
 	return assetWitnesses, nil
@@ -1035,6 +1068,9 @@ func fetchAssetsWithWitness(ctx context.Context, q ActiveAssetsStore,
 		return nil, nil, fmt.Errorf("unable to read db assets: %w", err)
 	}
 
+	// QueryAssets joins on unique columns only, so it returns each asset
+	// primary key at most once. The witness fetch below relies on that,
+	// since it groups the witnesses of the assets it is given.
 	assetIDs := fMap(dbAssets, func(a ConfirmedAsset) int64 {
 		return a.AssetPrimaryKey
 	})
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -59,6 +59,10 @@ type (
 	// its asset ID.
 	AssetProofByIDRow = sqlc.FetchAssetProofsByAssetIDRow
 
+	// AssetProofsByIDsRow is the asset proof for a given asset, identified
+	// by its asset primary key.
+	AssetProofsByIDsRow = sqlc.FetchAssetProofsByIDsRow
+
 	// PrevInput stores the full input information including the prev out,
 	// and also the witness information itself.
 	PrevInput = sqlc.UpsertAssetWitnessParams
@@ -252,6 +256,11 @@ type ActiveAssetsStore interface {
 	// ID.
 	FetchAssetProofsByAssetID(ctx context.Context,
 		assetID []byte) ([]AssetProofByIDRow, error)
+
+	// FetchAssetProofsByIDs fetches the asset proofs for the assets
+	// identified by the passed set of asset primary keys.
+	FetchAssetProofsByIDs(ctx context.Context,
+		assetIds []int64) ([]AssetProofsByIDsRow, error)
 
 	// UpsertChainTx inserts a new or updates an existing chain tx into the
 	// DB.
@@ -692,6 +701,10 @@ func parseAssetWitness(input AssetWitness) (asset.Witness, error) {
 // dbAssetsToChainAssets maps a set of confirmed assets in the database, and
 // the witnesses of those assets to a set of normal ChainAsset structs needed
 // by a higher level application.
+//
+// The returned slice is index-aligned with the input: chainAssets[i] is built
+// from dbAssets[i]. Callers rely on this to map a chain asset back to its
+// database row.
 func dbAssetsToChainAssets(dbAssets []ConfirmedAsset, witnesses assetWitnesses,
 	dbClock clock.Clock) ([]*asset.ChainAsset, error) {
 
@@ -1533,7 +1546,7 @@ func (a *AssetStore) FetchOrphanUTXOs(ctx context.Context) (
 			// Query all assets anchored at this outpoint.
 			// We include spent assets here because tombstones are
 			// marked as spent when created.
-			assetsAtAnchor, err := a.queryChainAssets(
+			_, assetsAtAnchor, err := a.queryChainAssets(
 				ctx, q, QueryAssetFilters{
 					AnchorPoint: u.Outpoint,
 					Now:         sqlTime(now),
@@ -2323,23 +2336,26 @@ func (a *AssetStore) RemoveSubscriber(
 
 // queryChainAssets queries the database for assets matching the passed filter.
 // The returned assets have all anchor and witness information populated.
+// The returned chain assets are index-aligned with the returned raw database
+// rows, so callers can map a chain asset back to its database row.
 func (a *AssetStore) queryChainAssets(ctx context.Context, q ActiveAssetsStore,
-	filter QueryAssetFilters) ([]*asset.ChainAsset, error) {
+	filter QueryAssetFilters) ([]ConfirmedAsset, []*asset.ChainAsset,
+	error) {
 
 	dbAssets, assetWitnesses, err := fetchAssetsWithWitness(
 		ctx, q, filter,
 	)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	matchingAssets, err := dbAssetsToChainAssets(
 		dbAssets, assetWitnesses, a.clock,
 	)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	return matchingAssets, nil
+	return dbAssets, matchingAssets, nil
 }
 
 // FetchCommitment returns a specific commitment identified by the given asset
@@ -2544,20 +2560,85 @@ func (a *AssetStore) queryCommitments(ctx context.Context,
 			map[wire.OutPoint][]asset.AltLeaf[asset.Asset],
 		)
 		matchingAssetProofs = make(map[wire.OutPoint]proof.Blob)
-		err                 error
 	)
 
 	readOpts := NewAssetStoreReadTx()
 	dbErr := a.db.ExecTx(ctx, &readOpts, func(q ActiveAssetsStore) error {
 		// Now that we have the set of filters we need we'll query the
-		// DB for the set of assets that matches them.
-		matchingAssets, err = a.queryChainAssets(ctx, q, assetFilter)
+		// DB for the set of assets that matches them. We keep the raw
+		// DB assets as well, since we need their primary keys to batch
+		// fetch the proofs below.
+		dbAssets, assets, err := a.queryChainAssets(
+			ctx, q, assetFilter,
+		)
 		if err != nil {
 			return err
 		}
 
+		matchingAssets = assets
 		if len(matchingAssets) == 0 {
 			return tapfreighter.ErrMatchingAssetsNotFound
+		}
+
+		// TODO(jhb): replace full proof fetch with
+		// outpoint -> alt leaf table / index
+		// We also need to fetch an input proof for each anchor point,
+		// in order to fetch any committed alt leaves. The alt leaves
+		// of a proof are the ones committed to by the anchor output as
+		// a whole (see the TaprootAssetRoot.FetchAltLeaves call in
+		// proof.CreateTransitionProof), so every asset at an anchor
+		// point carries the same set and one proof per distinct anchor
+		// point is enough.
+		//
+		// We therefore pick a representative asset per anchor point
+		// and fetch the proofs of all representatives in batched
+		// queries, keyed by the assets' primary keys. The
+		// representative is the lowest primary key at the anchor
+		// point, which doesn't depend on the order the assets were
+		// queried in. An asset that has no proof stored still fails
+		// the query below if it is the representative of its anchor
+		// point.
+		repAssetIDs := make(map[wire.OutPoint]int64)
+		for idx := range matchingAssets {
+			anchorPoint := matchingAssets[idx].AnchorOutpoint
+			primaryKey := dbAssets[idx].AssetPrimaryKey
+
+			existing, ok := repAssetIDs[anchorPoint]
+			if !ok || primaryKey < existing {
+				repAssetIDs[anchorPoint] = primaryKey
+			}
+		}
+
+		assetIDs := make([]int64, 0, len(repAssetIDs))
+		anchorByAssetID := make(
+			map[int64]wire.OutPoint, len(repAssetIDs),
+		)
+		for anchorPoint, primaryKey := range repAssetIDs {
+			assetIDs = append(assetIDs, primaryKey)
+			anchorByAssetID[primaryKey] = anchorPoint
+		}
+
+		err = forEachIDChunk(assetIDs, func(chunk []int64) error {
+			dbProofs, err := q.FetchAssetProofsByIDs(ctx, chunk)
+			if err != nil {
+				return err
+			}
+
+			for _, dbProof := range dbProofs {
+				anchor := anchorByAssetID[dbProof.AssetID]
+				matchingAssetProofs[anchor] = dbProof.ProofFile
+			}
+
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+
+		// Every anchor point must have a proof, as each proof row is
+		// unique per asset.
+		if len(matchingAssetProofs) != len(repAssetIDs) {
+			return proof.ErrProofNotFound
 		}
 
 		// At this point, we have the set of assets that match our
@@ -2582,7 +2663,7 @@ func (a *AssetStore) queryCommitments(ctx context.Context,
 				Now:         sqlTime(a.clock.Now().UTC()),
 			}
 
-			anchoredAssets, err := a.queryChainAssets(
+			_, anchoredAssets, err := a.queryChainAssets(
 				ctx, q, outpointQuery,
 			)
 			if err != nil {
@@ -2601,31 +2682,6 @@ func (a *AssetStore) queryCommitments(ctx context.Context,
 			}
 
 			anchorPoints[anchorPoint] = anchorUTXO
-
-			// TODO(jhb): replace full proof fetch with
-			// outpoint -> alt leaf table / index
-			// We also need to fetch the input proof here, in order
-			// to fetch any committed alt leaves.
-			assetLoc := proof.Locator{
-				AssetID:   fn.Ptr(matchingAsset.ID()),
-				ScriptKey: *matchingAsset.ScriptKey.PubKey,
-				OutPoint:  &matchingAsset.AnchorOutpoint,
-			}
-			proofArgs, err := locatorToProofQuery(assetLoc)
-			if err != nil {
-				return err
-			}
-
-			var assetProof proof.Blob
-			assetProof, err = fetchProof(ctx, q, proofArgs)
-			switch {
-			case errors.Is(err, sql.ErrNoRows):
-				return proof.ErrProofNotFound
-			case err != nil:
-				return err
-			}
-
-			matchingAssetProofs[anchorPoint] = assetProof
 		}
 
 		return nil
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -2564,6 +2564,16 @@ func (a *AssetStore) queryCommitments(ctx context.Context,
 
 	readOpts := NewAssetStoreReadTx()
 	dbErr := a.db.ExecTx(ctx, &readOpts, func(q ActiveAssetsStore) error {
+		// The transaction may retry the closure wholesale, so reset
+		// any state a previous attempt may have gathered. A stale
+		// entry would otherwise make the dedup logic below skip the
+		// work of a failed attempt.
+		chainAnchorToAssets = make(
+			map[wire.OutPoint][]*asset.ChainAsset,
+		)
+		anchorPoints = make(map[wire.OutPoint]AnchorPoint)
+		matchingAssetProofs = make(map[wire.OutPoint]proof.Blob)
+
 		// Now that we have the set of filters we need we'll query the
 		// DB for the set of assets that matches them. We keep the raw
 		// DB assets as well, since we need their primary keys to batch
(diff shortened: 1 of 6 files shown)


tapdb/assets_store.go
```diff
@@ -2508,22 +2553,84 @@ func (a *AssetStore) queryCommitments(ctx context.Context,
...
+		// commit to the anchor output as a whole, so one proof per
+		// distinct anchor point is enough. We pick a representative
+		// asset per anchor point and fetch the proofs of all
+		// representatives in batched queries, keyed by the assets'
+		// primary keys.
+		repAssetIDs := make(map[wire.OutPoint]int64)
+		for idx := range matchingAssets {
+			anchorPoint := matchingAssets[idx].AnchorOutpoint
+			primaryKey := dbAssets[idx].AssetPrimaryKey
+			repAssetIDs[anchorPoint] = primaryKey
```
litbot-9000: **The "Behavior" section's deliberate change only applies half the time — it depends on asset primary-key order.**

The PR body says:

> It now reads one proof per anchor point, so a proof missing for another asset at the same anchor surfaces when the funding path fetches input proofs, instead of failing coin selection.

That is only true when the proof-less asset is *not* the representative. This line is last-index-wins, and `QueryAssets` orders `ORDER BY assets.asset_id ASC`, so the representative is the **highest asset primary key** at that anchor point. If the asset whose proof is missing happens to be that one, the count check at line 2630 still fires and coin selection still fails with `ErrProofNotFound` — exactly the behaviour the body says has been removed.

I demonstrated it with a throwaway test: two assets at the *same* anchor point (`assetGen.anchorPoints[0]`, amounts 10 and 20), proof `asset_id`s `[1 2]`, then delete exactly one proof:

| deleted | `ListEligibleCoins` result |
|---|---|
| `asset_id = 1` (lowest) | succeeds, returns 2 coins, `err = <nil>` |
| `asset_id = 2` (highest) | `unable to query commitments: unable to find proof`, `errors.Is(err, proof.ErrProofNotFound) == true` |

Identical data; the only difference is which asset lost its proof.

To be clear about severity: this is **not a regression** — on `main` both cases fail — and failing loudly is the safe direction, which is why I'm not blocking. But the improvement is advertised unconditionally and delivered conditionally, on an ordering detail that no caller controls or is aware of. Two ways out, either is fine: correct the "Behavior" paragraph to say the proof-less asset still fails coin selection when it is the highest-keyed asset at its anchor point, or fall back to another asset at the same anchor point when the representative has no proof row.

One related fragility while you're here: the representative is only "highest pk" because neither `queryCommitments` caller sets `SortDirection` (it's plumbed for the `ListAssets` path at `:987-988` only). If `SortDirection` DESC were ever threaded into `FetchCommitment`/`ListEligibleCoins`, the representative silently flips to the *lowest* pk and this outcome flips with it. It's worth not depending on `matchingAssets`' sort order at all — e.g. pick the representative explicitly (`min` of the primary keys) rather than by whichever row lands last.

tapdb/assets_store_test.go
```diff
@@ -279,6 +296,86 @@ func assertAssetEqual(t *testing.T, a, b *asset.Asset) {
...
+// TestQueryCommitmentsMissingProof tests that querying commitments for an
+// asset that has no proof stored on disk reports the missing proof.
+func TestQueryCommitmentsMissingProof(t *testing.T) {
+	t.Parallel()
+
+	db := NewTestDB(t)
+	_, assetsStore := newAssetStoreFromDB(db.BaseDB)
+
+	assetGen := newAssetGenerator(t, 1, 1)
+	assetGen.genAssets(t, assetsStore, []assetDesc{{
```
litbot-9000, line 354: **This test uses a single asset, so it doesn't cover the case the whole optimisation rests on.**

`newAssetGenerator(t, 1, 1)` with one `assetDesc` means there's exactly one asset at one anchor point, so the representative *is* the only asset. The test therefore exercises the count check at `assets_store.go:2630`, but never the interesting configuration: **two or more assets sharing an anchor point**. That's the case where "one proof per distinct anchor point is enough" actually does any work, and it's the case a future regression would land in.

I checked, and nothing else in `./tapdb/` covers it either — the new `TestFetchAssetWitnesses` above uses `AddRandomAssetProof` per asset, which gives each its own anchor point.

Two cheap additions would close the gap, both of which I wrote as throwaway tests while reviewing, so I know the fixtures support them:

1. Two assets at the same anchor point (`assetGen.genAssets` with two `assetDesc`s both using `assetGen.anchorPoints[0]`), both with alt leaves present, asserting the reconstructed commitments and selected coins match what you'd get per-asset. This pins the invariant directly, and it's what would catch a future change that made alt leaves per-asset.
2. The same fixture with only the *non*-representative asset's proof deleted, asserting coin selection still succeeds — that pins the deliberate behaviour change from the PR body, which currently has no test at all.

(2) is also what would have caught the ordering asymmetry I flagged on `assets_store.go:2599`.

Separately, and to give credit: I mutation-tested `TestFetchAssetWitnesses` by flipping the generated query to `ORDER BY assets.asset_id, witness_index DESC`, and it does fail. So it genuinely asserts what its name claims about per-asset witness order.

tapdb/assets_store.go
```diff
@@ -565,27 +573,57 @@ type AssetHumanReadable struct {
...
-			return nil, err
+			return err
 		}
 
-		// We'll insert a nil witness for genesis asset, so we don't
-		// add it to the map, which'll give it the genesis witness.
-		if len(witnesses) == 0 {
-			continue
+		for _, witness := range witnesses {
+			assetWitnesses[witness.AssetID] = append(
```
litbot-9000, line 622: **Non-blocking, but worth a guard or a comment:** this `append`-based grouping is no longer idempotent under duplicate input IDs, where the code it replaces was.

The old loop did `assetWitnesses[assetID] = witnesses` — a plain assignment, so passing the same primary key twice was harmless. Now, duplicate IDs *within* one chunk are still fine (`IN (5, 5)` matches each witness row once), but duplicates that straddle a chunk boundary produce two separate queries, each returning that asset's full witness set, and the second one appends onto the first. The result is an asset with its witnesses **duplicated** — and since witnesses are consensus-relevant, that would show up as a wrong `PrevWitnesses` and a commitment that doesn't reconstruct.

I tried to find a live path for this and could not, so I'm not calling it a bug. `QueryAssets` cannot return a duplicate `asset_primary_key`: every join is on a unique column — `genesis_info_view` is one row per `gen_asset_id`, the `LEFT JOIN key_group_info_view` keys off `asset_group_witnesses.gen_asset_id` which is `UNIQUE` (`migrations/000002_assets.up.sql:117`), and `script_keys` / `internal_keys` / `managed_utxos` / `chain_txns` are all joined on their primary keys. And it would additionally need >512 assets with the duplicate landing on the boundary.

So the mechanism is masked today by a uniqueness property two files away, with nothing here recording the dependency. If someone later adds a fanning join to `QueryAssets` — easy to do, since it's a `LEFT JOIN`-heavy query that's already grown a `sqlc.slice` — witnesses silently duplicate rather than failing. `fn.Set`/dedup on `assetIDs` before chunking would make it structural for free; a comment on `fetchAssetsWithWitness`' `fMap` stating that `QueryAssets` must return distinct primary keys would be the minimum.

tapdb/assets_store.go
```diff
@@ -2508,22 +2553,84 @@ func (a *AssetStore) queryCommitments(ctx context.Context,
...
 	dbErr := a.db.ExecTx(ctx, &readOpts, func(q ActiveAssetsStore) error {
+		// The transaction may retry the closure wholesale, so reset
+		// any state a previous attempt may have gathered. A stale
+		// entry would otherwise make the dedup logic below skip the
+		// work of a failed attempt.
+		chainAnchorToAssets = make(
+			map[wire.OutPoint][]*asset.ChainAsset,
+		)
+		anchorPoints = make(map[wire.OutPoint]AnchorPoint)
+		matchingAssetProofs = make(map[wire.OutPoint]proof.Blob)
```
litbot-9000, line 2575: Verified this reset is both **complete** and genuinely **necessary**, since it's the kind of thing that only bites under Postgres serialization retries.

Enumerating everything the closure captures from the enclosing `var` block at `:2546-2556`:

- `chainAnchorToAssets` — reset here (`:2564`) ✓
- `anchorPoints` — reset here (`:2567`) ✓
- `matchingAssetProofs` — reset here (`:2568`) ✓
- `matchingAssets` — **not** in the reset block, but assigned wholesale at `:2581` (`matchingAssets = assets`), so a retry overwrites rather than accumulates ✓
- `anchorAltLeaves` (`:2552`) — never touched inside the closure at all; it's only populated at `:2697`, after `ExecTx` has returned. Needs no reset ✓

All five accounted for, nothing accumulates across attempts.

And it really is load-bearing, in two distinct ways that are both new in this PR:

1. The `continue` at `:2649` keys off `chainAnchorToAssets`, so a stale key from a failed attempt would cause the loop to skip fetching the anchored assets and the managed UTXO — leaving `anchorPoints[anchorPoint]` unset while `chainAnchorToAssets[anchorPoint]` looks populated. The commitment loop after `ExecTx` would then read a zero-valued `AnchorPoint`.
2. The count check at `:2630` compares `len(matchingAssetProofs)` against `len(repAssetIDs)`. Without the reset, entries left over from an attempt that saw a *different* snapshot would inflate the left side and produce a spurious `ErrProofNotFound`.

I can't exercise this locally — no container runtime here, so no `test_db_postgres`/`dockertest` and no way to force a serialization retry. `run itests postgres (0-3)` and `run unit tests (unit dbbackend=postgres pkg=tapdb)` in CI are the actual evidence; both pass. Flagging that only so it's on the record that the retry path is CI-covered, not locally covered.

tapdb/assets_store.go
```diff
@@ -565,27 +573,57 @@ type AssetHumanReadable struct {
...
+const queryIDChunkSize = 512
+
+// forEachIDChunk calls the given function for each chunk of at most
+// queryIDChunkSize of the given IDs. Queries that take a set of IDs must be
+// issued in chunks, as the number of bind parameters a statement may carry is
+// limited. An empty ID set results in no calls at all, which also keeps us
+// from issuing a query with an empty ID set, which no such query supports.
+func forEachIDChunk(ids []int64, f func(chunk []int64) error) error {
+	for start := 0; start < len(ids); start += queryIDChunkSize {
+		end := min(start+queryIDChunkSize, len(ids))
```
litbot-9000, line 593: Checked the chunking arithmetic and the bind-parameter headroom, since the comment above cites the caps as the justification. It all holds, with more room than the comment implies.

**Parameters per ID is exactly one.** Both new queries take the ID slice as their *only* parameter, and sqlc's generated code appends one placeholder per element (`makeQueryParams(len(queryParams), len(assetIds))` at `sqlc/assets.sql.go:905-917` and `:1001-1013`). So a full chunk is 512 bind parameters against 32766 (SQLite) / 65535 (Postgres) — roughly 64× headroom, and no other parameter shares the budget. Worth contrasting with `QueryAssets`, which *does* mix a `sqlc.slice` with positional params and needed an explicit comment about parameter renumbering; neither of these queries has that problem.

**Boundaries.** `min(start+queryIDChunkSize, len(ids))` gives `[0:512]` then `[512:513]` for 513 IDs, and a single `[0:512]` for exactly 512. No off-by-one, no empty trailing chunk.

**Empty input.** `start < len(ids)` means zero iterations, so `f` is never called and no query is issued. That matters more than it looks: sqlc's `else` branch substitutes the literal `NULL` for the slice, giving `IN (NULL)`, which is not a syntax error but silently matches nothing — precisely the trap the `must NEVER be an empty slice` comments in the query file warn about. Since `forEachIDChunk` is the only route to either query (the sole callers are `:612` and `:2612`), that branch is unreachable from this package. Good structural choice — the guard lives in one place rather than at each call site.

tapdb/sqlc/queries/assets.sql
```diff
@@ -867,16 +877,19 @@ INSERT INTO asset_witnesses (
...
-    (assets.asset_id = sqlc.narg('asset_id')) OR (sqlc.narg('asset_id') IS NULL)
-)
-ORDER BY witness_index;
+-- The witnesses of all assets identified by the passed set of asset primary
+-- keys are fetched in a single query.
+--
+-- The asset_ids argument must NEVER be an empty slice, otherwise this query
+-- will return no results.
+WHERE assets.asset_id IN (sqlc.slice('asset_ids')/*SLICE:asset_ids*/)
+ORDER BY assets.asset_id, witness_index;
```
litbot-9000, line 892: This `ORDER BY` is the right one, and I want to record *why* it's sufficient, because the reasoning depends on a constraint that isn't visible from here.

`(asset_id, witness_index)` has a **unique** index — `migrations/000020_asset_unique_key.up.sql:97-99` (originally `000015_asset_witnesses.up.sql:15-17`) — so this is a *total* order, not a partial one. There are no ties for the database to break arbitrarily, which is exactly what a Go-side grouping needs now that row order is whatever the DB returns rather than per-asset query order. The `append` loop at `assets_store.go:617-621` then preserves each asset's sequence by construction.

The switch from the old `ORDER BY witness_index` is also equivalent per asset rather than merely close: the old form ran inside a single-asset query, so `witness_index` alone was already a total order there.

Two things I checked that could have gone wrong and didn't:

- Legacy rows from before `witness_index` existed were backfilled as `witness_index = witness_id` (`000015_asset_witnesses.up.sql:10-11`), i.e. sparse but insertion-ordered. Sorting them yields the same order the old per-asset query did, so this doesn't change how pre-migration assets deserialise.
- Zero-witness genesis assets return no rows, so the map key is simply absent — matching the old code's explicit `if len(witnesses) == 0 { continue }`. `dbAssetsToChainAssets` guards with `assetInputs, ok := witnesses[...]; if ok`, leaving `PrevWitnesses` nil either way. No caller sees a nil-vs-empty-slice change, which was my main worry going in.

tapdb/assets_store.go
```diff
@@ -2535,6 +2642,14 @@ func (a *AssetStore) queryCommitments(ctx context.Context,
 		for idx := range matchingAssets {
 			matchingAsset := matchingAssets[idx]
 			anchorPoint := matchingAsset.AnchorOutpoint
+
+			// Multiple matching assets may be anchored at the same
+			// outpoint, in which case we've already fetched the
+			// anchored assets and managed UTXO for it.
+			if _, ok := chainAnchorToAssets[anchorPoint]; ok {
```
litbot-9000, line 2669: Commit 3's skip checks out. Two things I looked at specifically:

**The skip key is the right identity.** `anchorPoint` is a `wire.OutPoint` (hash *and* index), not a txid, so two commitments in different outputs of the same anchor transaction are correctly treated as distinct.

**It doesn't drop per-asset work.** Everything inside the guarded body depends only on `anchorPoint` — `encodeOutpoint`, `queryChainAssets` with an `AnchorPoint`-only filter, and `FetchManagedUTXO` by outpoint. At the base commit these were plain map assignments repeated once per asset at the anchor point, overwriting with identical values, so collapsing them to the first occurrence is a pure win. Critically, the *proof* representative is chosen in a separate loop at `:2596` that has no skip, so this `continue` doesn't influence which proof is used — the two concerns are properly decoupled, which is what makes the skip safe independently of the alt-leaf invariant.

Also robust in the edge case: if `queryChainAssets` at `:2664` ever returned a nil slice, the assignment at `:2671` still creates the map key, so `_, ok :=` sees it and later iterations skip correctly rather than looping forever or re-querying. (In practice it can't be empty — the matching asset is itself anchored there and `outpointQuery` sets no `spent`/`leased` filter — but the guard doesn't depend on that.)

tapdb/assets_store.go
```diff
@@ -2508,22 +2553,84 @@ func (a *AssetStore) queryCommitments(ctx context.Context,
...
+		matchingAssets = assets
 		if len(matchingAssets) == 0 {
 			return tapfreighter.ErrMatchingAssetsNotFound
 		}
 
+		// TODO(jhb): replace full proof fetch with
+		// outpoint -> alt leaf table / index
+		// We also need to fetch an input proof for each anchor point,
+		// in order to fetch any committed alt leaves. The alt leaves
+		// commit to the anchor output as a whole, so one proof per
```
litbot-9000: Since this comment is the justification for the entire optimisation, I went and checked it against the write side rather than taking it as read. **It's accurate**, and it's worth pointing at the code that makes it true, because a future reader will want that more than the assertion.

`proof.AltLeaves` has exactly one non-test writer: `proof/append.go:181-188` in `CreateTransitionProof`, which does `params.TaprootAssetRoot.FetchAltLeaves()` and stores the result verbatim. `TaprootAssetRoot` is the anchor **output's** `TapCommitment`, and `FetchAltLeaves` (`commitment/tap.go:679-690`) returns every leaf under `assetCommitments[asset.EmptyGenesisID]` — the output's whole alt-leaf commitment. Nothing about it is keyed by asset, script key or genesis, so a passive asset, a split output, a tombstone or a differing script key at the same anchor point all see the identical set. `tapfreighter/wallet.go:626-636` independently confirms the model on the write path, grouping alt leaves by `vOut.AnchorOutputIndex`.

Two things that make this a comfortable conclusion rather than a fingers-crossed one:

- **This PR doesn't actually introduce the dependency.** On `main`, `matchingAssetProofs[anchorPoint] = assetProof` is a map assignment inside the per-asset loop, so the last matching asset at each anchor point already won and every other proof was fetched and discarded. Combined with `repAssetIDs[anchorPoint] = primaryKey` at `:2599` also being last-index-wins over the same slice, this PR selects the *same* asset's proof that `main` does. Behaviour-preserving, not behaviour-changing.
- **A violation would be loud, not silent.** `MergeAltLeaves` at `:2750` feeds `TapscriptRoot`, which is compared to `anchorUTXO.TaprootAssetRoot` at `:2765-2768`. A wrong or incomplete alt-leaf set fails there with `mismatch of managed utxo and constructed tap commitment root` during coin selection, rather than producing a commitment that goes bad later during verification.

Suggestion, take it or leave it: the phrase "commit to the anchor output as a whole" is doing a lot of work here. A pointer to `proof/append.go`'s `FetchAltLeaves` call would let the next person verify it in one hop instead of the several it took me.

litbot-9000 (commented):
Reviewed all five commits at `cef4b27`. The optimisation is sound, the load-bearing invariant holds, and I could not find a correctness hole. Build, the full `./tapdb/...` suite, a sequential `-race` run, and the downstream consumers all pass locally, and all 41 CI checks are green. Three non-blocking findings below; the most substantive is that the PR body's stated behaviour change around missing proofs is only half-true — whether it applies depends on asset primary-key order, which I demonstrated with a throwaway test. Event is `COMMENT`, not `REQUEST_CHANGES`: nothing here can produce a wrong commitment, and the one behavioural inconsistency fails loudly and safely.

### Does one proof per anchor point hold?

**Yes.** Two independent reasons, neither of which is "the description says so".

1. **It is not actually a new invariant.** At the base commit, `matchingAssetProofs[anchorPoint] = assetProof` inside `for idx := range matchingAssets` was a plain map *assignment*, so the last matching asset at each anchor point already won and every other asset's proof was fetched and then thrown away. `repAssetIDs[anchorPoint] = primaryKey` (`tapdb/assets_store.go:2599`) is also last-index-wins over the same slice, so this PR picks the **same** representative asset and the **same** proof row as `main` does today. `asset_proofs.asset_id` is `UNIQUE` (`tapdb/sqlc/migrations/000002_assets.up.sql:250`), so there is exactly one proof row per asset, and the old script-key + outpoint + asset-ID locator path resolved to that same row.

2. **The write side genuinely is a per-anchor-output property.** The only non-test writer of `proof.AltLeaves` is `proof/append.go:181-188` in `CreateTransitionProof`: `altLeaves, err := params.TaprootAssetRoot.FetchAltLeaves()` → `proof.AltLeaves = asset.ToAltLeaves(altLeaves)`. `TaprootAssetRoot` is the anchor *output's* `TapCommitment`, and `FetchAltLeaves` (`commitment/tap.go:679-690`) returns every leaf under `assetCommitments[asset.EmptyGenesisID]` — the output's entire alt-leaf commitment, with nothing per-asset about it. So a different script key, a passive asset, a split or a tombstone at the same anchor point does not change the set. The send path agrees: `tapfreighter/wallet.go:626-636` groups alt leaves by `vOut.AnchorOutputIndex` ("Each anchor output must have a valid set of AltLeaves at this point").

And if the invariant were ever violated, it fails loudly instead of silently: the reconstructed root after `MergeAltLeaves` is compared against `anchorUTXO.TaprootAssetRoot` (`tapdb/assets_store.go:2750-2768`), so a wrong or incomplete alt-leaf set yields `mismatch of managed utxo and constructed tap commitment root` at coin-selection time rather than a bad commitment that surfaces as a verification failure much later. That is what keeps this out of blocker territory. I also confirmed `lastProof.AltLeaves` is the *only* thing ever read from the fetched proof (`tapdb/assets_store.go:2691-2699`).

### What I verified

| item | verdict |
|---|---|
| A. one proof per anchor point | **holds** — see above; same representative as `main`, per-output write side, root check as backstop |
| B. row grouping / ordering | **holds** — `ORDER BY assets.asset_id, witness_index` is a *total* order (UNIQUE index on `(asset_id, witness_index)`); zero-witness assets behave identically |
| C. retry reset | **complete** — all four closure accumulators accounted for |
| D. chunking arithmetic | **safe** — exactly 1 bind param per ID, 512 vs 32766/65535; empty set issues no query at all |
| E. anchor-point skip | **correct** — full `wire.OutPoint`, and the skipped work depends only on the anchor point |
| F. dialect portability | **consistent** — one shared `queries/assets.sql`, dialect-neutral `/*SLICE:*/` runtime expansion, committed codegen matches the source; CI `Sqlc check` passes |
| G. performance claim | **methodology is sound** — the "this PR" column is isolated from #2252, and the byte-identical dump/coin-set diff is the right behavioural control. The query-count reduction (`1 + N` → `1 + ceil(N/512)` witness queries; outer loop deduped per anchor point) is consistent with the reported ratios. I could not reproduce it (no 2461-asset node here), and no benchmark is committed, so nothing guards against a silent regression later — worth considering, but not something I'd hold the PR for |

Commands actually run, in this clone at `cef4b27`:

- `make build` → exit 0 (full build incl. `tapd`/`tapcli`, tags `dev monitoring`).
- `go test ./tapdb/... -count=1 -timeout 60m` → exit 0, `ok ... 65.172s`, 0 failures.
- `go test ./tapdb/ -run 'TestFetchAssetWitnesses|TestQueryCommitmentsMissingProof' -count=1 -v` → both PASS.
- `go test -race ./tapdb/ -count=1 -timeout 80m` → exit 0, `ok ... 640.754s`, zero `WARNING: DATA RACE`.
- `go test ./tapfreighter/... ./tapchannel/... ./monitoring/... -count=1` → exit 0 (`monitoring` has no test files).
- **Mutation test on the new test** — I changed the generated query to `ORDER BY assets.asset_id, witness_index DESC` and re-ran `TestFetchAssetWitnesses`: it **fails**. So the new ordering test is not vacuous, it really pins per-asset witness order. (Worth knowing: dropping `assets.asset_id` from the `ORDER BY` alone would *not* be caught, since a global sort by `witness_index` still yields ascending `witness_index` within each asset. The clause that matters is the one the test guards.)
- `gh pr checks 2251` → all 41 pass, including `Sqlc check`, `Lint check`, `Format check`, `run unit tests (unit dbbackend=postgres pkg=tapdb)`, `run itests postgres (0-3)` and `run unit race (0-3)`.

**No container runtime in my environment** (no docker/podman/psql), so the `test_db_postgres` / `ory/dockertest` paths did not run here at all — I have **no local Postgres coverage** to offer, and I could not exercise Postgres serialization retries (the scenario the new reset block in `ExecTx` exists for) or run `make gen` / `make sqlc-check`. For those three things the CI jobs above are the evidence, not me. I compared the committed generated code against the `.sql` sources by reading instead.

### Findings

Three, all inline, none blocking. Ordered by how much I'd care: the missing-proof order dependence (`tapdb/assets_store.go:2599`), the missing test for the shared-anchor case (`tapdb/assets_store_test.go:354`), and a latent non-idempotency in the new grouping loop (`tapdb/assets_store.go:618`).

GeorgeTsagk:
All four addressed.

1. The representative is now the lowest primary key at the anchor point instead of whichever row landed last, so it no longer depends on the order `matchingAssets` came back in, or on `SortDirection` being threaded in later. I kept the strict behaviour rather than falling back to another asset at the same anchor: a missing proof for an owned asset is corruption, and failing loudly is the safer direction. The PR description now says a missing proof still fails coin selection when it belongs to that asset.
2. Added `TestQueryCommitmentsSharedAnchor`: two assets at one anchor point, asserting both are listed with the same commitment and each is reachable through `FetchCommitment`, plus both proof-deletion cases (another asset's proof gone still lists, the representative's gone returns `ErrProofNotFound`). Worth knowing for anyone writing a similar fixture: the two assets need distinct genesis IDs, since two assets sharing an asset ID at one anchor point don't reconstruct to the stored commitment root.
3. Documented on `forEachIDChunk` that the IDs must be distinct and what breaks if they aren't, and recorded at the call site that `QueryAssets` joins on unique columns only. Left the dedup out, since there's no path to a duplicate today and the comment is what a future fanning join would need to trip over.
4. The alt-leaf comment now points at the `TaprootAssetRoot.FetchAltLeaves` call in `CreateTransitionProof`.

One unrelated line: `upsertGenesis` in `assets_store_test.go` is wrapped because the added test shifted it into the linter's diff window.

jtobin (approved):
LGTM. :fire: :fire:

PR #2252, opened 2026-08-24
tapfreighter: bound input selection instead of loading every eligible coin

### LINKED ISSUE #2250: tapfreighter: input selection loads every eligible asset, unbounded
### Summary

Funding a send loads every eligible asset the node holds before selecting inputs.
There is no `LIMIT`, and the SQL amount filter is passed `1`, so a 1-unit send on
a node with 1338 assets materialises all 1338. Input selection is 73% of the
sender state machine and grows linearly with holdings.

### Where

`ListEligibleCoins` never sets `NumLimit`, and zero is turned into an unbounded
query:

```go
// tapdb/assets_store.go:947 and :1019 (main), :957 and :1029 (v0.8.1)
if assetFilter.NumLimit == 0 {
	assetFilter.NumLimit = math.MaxInt32
}
```

```sql
-- tapdb/sqlc/queries/assets.sql, QueryAssets
LIMIT @num_limit OFFSET @num_offset;
```

The amount predicate exists but is inert, because the caller asks for `MinAmt: 1`
and applies the real target amount in Go afterwards:

```go
// tapfreighter/coin_select.go:56
listConstraints := CommitmentConstraints{
	AssetSpecifier:    constraints.AssetSpecifier,
	MinAmt:            1,
```

```sql
assets.amount >= COALESCE(sqlc.narg('min_amt'), assets.amount) AND
```

So the strategy is "list all eligible, then select", and the cost of a send scales
with everything the node holds rather than with the number of inputs needed.

### Measurement

Sender state machine, median across sends on a node with 1338 assets (sqlite,
85 MB database, regtest):

```
  phase                        median   share
  VirtualCommitmentSelect       8.12s   73.4%
  VerifyPreBroadcast            0.80s    7.2%
  TransferProofs                0.51s    4.6%
  AnchorSign                    0.42s    3.8%
  StorePostAnchorTxConf         0.38s    3.4%
  StorePreBroadcast             0.04s
  WaitTxConf                    0.02s
  Broadcast                     0.01s
  VirtualSign                   0.01s
                             --------
  sender state machine         11.06s
```

30s CPU profile of one `tapcli assets send`, folded onto owned code:

```
  44.0%  5.42s  tapdb/sqlc.(*Queries).QueryAssets
  13.0%  1.60s  tapdb/sqlc.(*Queries).FetchAssetWitnesses
   5.7%  0.70s  tapdb/sqlc.(*Queries).FetchAssetProof
   5.7%  0.67s  proof.deriveTaprootKeyFromTapCommitment
```

57% of total CPU is inside `modernc.org/sqlite`'s `_sqlite3VdbeExec`, and a
further 15% is secp256k1 field arithmetic from re-deriving commitment keys for
candidates.

Related: the per-asset witness and proof queries are #2249. Fixing the bound here
reduces the row count those run over, but they are separable.

### Possible direction

The query is already ordered (`ORDER BY assets.asset_id`) and already takes a
limit, so a bounded fetch looks feasible: order by amount descending, take a page,
and fetch more only when the page cannot cover the target. Worth noting this
changes which coins get selected, so it is not purely a performance change: it
touches change output shape, UTXO fragmentation and fee estimation, and probably
some test expectations. A maintainer should pick the selection strategy.

### Environment

- tapd `v0.8.1`, sqlite backend, regtest
- The binary carries one added import, `_ "net/http/pprof"`, so that `--profile`
  serves anything. No other change.
- Node state: 1338 assets, 5 groups, 3692 universe leaves, 85 MB sqlite database,
  accumulated over ~500 sends of 1 unit each via v2 addresses.

### Caveats

- Most of the 1338 assets are small change outputs, which is a more fragmented
  wallet than typical. The linear scaling holds regardless, but the absolute 8s
  reflects that profile.
- Only sqlite was profiled; `modernc.org/sqlite` is pure Go and slower than the
  cgo driver, so part of the 57% is driver cost rather than row count.

GeorgeTsagk:
Funding a send listed every eligible coin the node holds before selecting
inputs: `ListEligibleCoins` was unbounded and the SQL amount filter was passed
1, with the real target applied in Go afterwards. Listing a coin is expensive,
since the full commitment of its anchor output is reconstructed, so the cost of
a send scaled with total holdings instead of with the number of inputs needed.

### Changes

- `CommitmentConstraints` gains `CoinLimit`/`CoinOffset`, and `QueryAssets`
  gains an amount-descending order option that bounded listings use. Bounding
  cannot be combined with pinned `PrevIDs`, since that filter runs after the
  query, so `ListEligibleCoins` now rejects the combination instead of
  silently ignoring the bound.
- `SelectCoins` lists coins in pages of 32 and stops as soon as the compatible
  coins cover the target. A single page covers the common case.
- Pages are separate queries, so the coin set can shift between them. Coins
  are de-duplicated across pages by prev ID, and a listing that took more than
  one page and still could not cover the target is repeated unbounded before
  reporting insufficient funds, so that answer stays as accurate as before. A
  single-page listing needs no retry, as no offset was applied to it.

### Speedups

Measured against a copy of a real regtest node: sqlite, 2461 assets, dominant
asset group holding 621 coins.

| operation | before | this PR | with #2251 too |
|---|---|---|---|
| coin selection, 1-unit group send | 16.57s | 792ms | 68ms |
| coin selection, 181-coin send | 16.56s | 4.69s | 447ms |
| coin selection, 621-coin send | 16.63s | 14.63s | 2.28s |
| coin selection, insufficient funds | 16.62s | 28.93s | 3.81s |

Note the last row: with this PR alone, a send that cannot be funded is 1.7x
slower than before because of the unbounded retry. It only affects that
failure path, and it drops to 3.81s (4.4x faster than baseline) once the
per-asset query batching in #2251 lands as well. The alternative is keyset pagination
on `(amount, assets.asset_id)`, where a cursor cannot skip rows and the retry
becomes unnecessary; happy to do that instead if preferred.

### Behavior

Selection results are unchanged for unequal amounts: `PreferMaxAmount` already
took candidates in descending amount order until the target was covered, and
the pages arrive in that same order.

This PR shrinks the set that gets loaded, #2251 cuts the per-asset query cost
of whatever is loaded, so the two are complementary. End to end with both, a
v2 address send on that node goes from 28.1s to 10.8s median, and stays flat
as the wallet fragments further.

Diffing full dumps of the selected coin sets against baseline (1 unit, 181
coins, 621 coins, single asset ID, pinned prev IDs, insufficient funds):
identical coin count, total and amount multiset in every case, and identical
error strings. The only difference is which coin gets picked among coins of
exactly equal amount, one swap in the 1-unit case and three in the 181-coin
case. Tie order was never defined, since `selectForAmount` sorts with
`sort.Slice`.

The new ordering term is inert for other `QueryAssets` consumers, as its
argument is NULL unless a bounded listing is requested. Verified by hashing
the returned row order for five query shapes; `TestFetchAllAssetsPagination`
covers the same on postgres, where NULL ordering semantics differ.

### Testing

New tests for paging, cross-page duplicates, a skipped coin, the listing error
path, and the bounded listing at the tapdb level. Unit tests pass on sqlite
and postgres, `make lint` and `make sqlc-check` are clean, and itests pass
(`basic_send_unidirectional`, `address_v2_self_send`, `full_value_send`,
`multi_input_send_non-interactive_single_ID`). Every commit builds and compiles
its unit tests standalone.

Closes #2250

### DIFF
--- a/docs/release-notes/release-notes-0.9.0.md
+++ b/docs/release-notes/release-notes-0.9.0.md
@@ -99,6 +99,18 @@
   on nodes holding many assets
   ([#2249](https://github.com/lightninglabs/taproot-assets/issues/2249)).
 
+* [PR#2252](https://github.com/lightninglabs/taproot-assets/pull/2252)
+  bounds input selection when funding a send. Eligible coins are now
+  listed in pages of descending amounts and listing stops as soon as the
+  accumulated amount covers the send target, instead of loading every
+  eligible coin the node holds
+  ([#2250](https://github.com/lightninglabs/taproot-assets/issues/2250)).
+  Sends that request specific inputs keep listing every eligible coin, as
+  the requested inputs are filtered for after the listing. A send that
+  cannot be funded is somewhat slower than before, as the listing is
+  repeated unbounded before reporting insufficient funds, so that a coin
+  the paged listing may have missed can't be mistaken for missing funds.
+
 ## Deprecations
 
 # Technical and Architectural Updates
--- a/tapdb/assets_store.go
+++ b/tapdb/assets_store.go
@@ -2409,6 +2409,14 @@ func (a *AssetStore) ListEligibleCoins(ctx context.Context,
 		return nil, fmt.Errorf("min amount overflow")
 	}
 
+	// The prev ID filter is applied after the query, so a bounded listing
+	// could return a page that holds none of the requested inputs. The two
+	// can't be combined.
+	if constraints.CoinLimit > 0 && len(constraints.PrevIDs) > 0 {
+		return nil, fmt.Errorf("coin limit cannot be combined with " +
+			"specific prev IDs")
+	}
+
 	// First, we'll map the commitment constraints to our database query
 	// filters.
 	assetFilter, err := a.constraintsToDbFilter(&AssetQueryFilters{
@@ -2426,6 +2434,21 @@ func (a *AssetStore) ListEligibleCoins(ctx context.Context,
 	// unconfirmed assets would otherwise be included). Unconfirmed assets
 	// have a block height of 0, so we set the minimum block height to 1.
 	assetFilter.MinAnchorHeight = sqlInt32(1)
+
+	// If the caller bounds the number of coins, we return them in
+	// descending amount order, so the largest eligible coins are returned
+	// first and a bounded listing can cover a target amount with as few
+	// pages as possible.
+	//
+	// This overrides the pagination the query filters may already carry
+	// from AssetQueryFilters.Limit/Offset, which no caller of this method
+	// sets. Coins are paged through with CoinLimit/CoinOffset instead, as
+	// those live on the constraints the coin selector works with.
+	if constraints.CoinLimit > 0 {
+		assetFilter.NumLimit = constraints.CoinLimit
+		assetFilter.NumOffset = constraints.CoinOffset
+		assetFilter.OrderByAmount = sqlInt32(1)
+	}
 
 	selectedCommitments, err := a.queryCommitments(ctx, assetFilter)
 	if err != nil {
--- a/tapdb/assets_store_test.go
+++ b/tapdb/assets_store_test.go
@@ -1869,6 +1869,99 @@ func TestSelectCommitment(t *testing.T) {
 	}
 }
 
+// TestListEligibleCoinsBounded tests that a bounded eligible coin listing
+// returns coins in pages of descending amounts.
+func TestListEligibleCoinsBounded(t *testing.T) {
+	t.Parallel()
+
+	const numAssets = 5
+
+	assetGen := newAssetGenerator(t, numAssets, 1)
+	_, assetsStore, _ := newAssetStore(t)
+
+	// We create one asset per anchor point, with amounts that are neither
+	// sorted nor aligned with the anchor point order.
+	amounts := []uint64{10, 40, 20, 50, 30}
+	assetDescs := make([]assetDesc, numAssets)
+	for i := 0; i < numAssets; i++ {
+		assetDescs[i] = assetDesc{
+			assetGen:    assetGen.assetGens[0],
+			amt:         amounts[i],
+			anchorPoint: assetGen.anchorPoints[i],
+		}
+	}
+	assetGen.genAssets(t, assetsStore, assetDescs)
+
+	// The genesis (and with it the asset ID) of the generated assets
+	// depends on the anchor point, so we use an empty specifier to match
+	// all the created assets.
+	ctx := context.Background()
+	constraints := tapfreighter.CommitmentConstraints{
+		MinAmt: 1,
+	}
+
+	listAmounts := func(coins []*tapfreighter.AnchoredCommitment) []uint64 {
+		return fn.Map(
+			coins, func(c *tapfreighter.AnchoredCommitment) uint64 {
+				return c.Asset.Amount
+			},
+		)
+	}
+
+	// Page through the coins two at a time, expecting descending amounts.
+	expectedPages := [][]uint64{{50, 40}, {30, 20}, {10}}
+	for idx, expected := range expectedPages {
+		constraints.CoinLimit = 2
+		constraints.CoinOffset = int32(idx * 2)
+
+		coins, err := assetsStore.ListEligibleCoins(ctx, constraints)
+		require.NoError(t, err)
+		require.Equal(t, expected, listAmounts(coins))
+	}
+
+	// A page past the end of the eligible coin set reports that no assets
+	// were found.
+	constraints.CoinOffset = numAssets + 1
+	_, err := assetsStore.ListEligibleCoins(ctx, constraints)
+	require.ErrorIs(t, err, tapfreighter.ErrMatchingAssetsNotFound)
+
+	// An unbounded listing returns all the coins, in no particular order.
+	unbounded, err := assetsStore.ListEligibleCoins(
+		ctx, tapfreighter.CommitmentConstraints{
+			MinAmt: 1,
+		},
+	)
+	require.NoError(t, err)
+	require.Len(t, unbounded, numAssets)
+
+	smallest := unbounded[0]
+	for _, c := range unbounded {
+		if c.Asset.Amount < smallest.Asset.Amount {
+			smallest = c
+		}
+	}
+
+	// Specific inputs can be requested from an unbounded listing, as the
+	// prev ID filter is applied after the query.
+	coins, err := assetsStore.ListEligibleCoins(
+		ctx, tapfreighter.CommitmentConstraints{
+			MinAmt:  1,
+			PrevIDs: []asset.PrevID{smallest.PrevID()},
+		},
+	)
+	require.NoError(t, err)
+	require.Len(t, coins, 1)
+	require.Equal(t, smallest.Asset.Amount, coins[0].Asset.Amount)
+
+	// Combining the two is a programming error, as a bounded listing could
+	// hold none of the requested inputs.
+	constraints.CoinLimit = 1
+	constraints.CoinOffset = 0
+	constraints.PrevIDs = []asset.PrevID{smallest.PrevID()}
+	_, err = assetsStore.ListEligibleCoins(ctx, constraints)
+	require.ErrorContains(t, err, "cannot be combined")
+}
+
 // TestAssetExportLog tests that were able to properly spend/transfer assets on
 // disk. This ensures we can properly commit the end result of an asset
 // transfer initiated at a higher level.
--- a/tapdb/assets_store_test.go
+++ b/tapdb/assets_store_test.go
@@ -1962,6 +1962,62 @@ func TestListEligibleCoinsBounded(t *testing.T) {
 	require.ErrorContains(t, err, "cannot be combined")
 }
 
+// TestListEligibleCoinsBoundedEqualAmounts tests that paging through a bounded
+// eligible coin listing neither repeats nor skips coins when all the coins hold
+// the same amount, which is the case where the amount ordering alone doesn't
+// decide the order of the pages.
+func TestListEligibleCoinsBoundedEqualAmounts(t *testing.T) {
+	t.Parallel()
+
+	const (
+		numAssets = 5
+		amount    = 10
+	)
+
+	assetGen := newAssetGenerator(t, numAssets, 1)
+	_, assetsStore, _ := newAssetStore(t)
+
+	assetDescs := make([]assetDesc, numAssets)
+	for i := 0; i < numAssets; i++ {
+		assetDescs[i] = assetDesc{
+			assetGen:    assetGen.assetGens[0],
+			amt:         amount,
+			anchorPoint: assetGen.anchorPoints[i],
+		}
+	}
+	assetGen.genAssets(t, assetsStore, assetDescs)
+
+	ctx := context.Background()
+	constraints := tapfreighter.CommitmentConstraints{
+		MinAmt: 1,
+	}
+
+	// Page through all the coins, collecting them by prev ID. Every page
+	// must be full until the coins run out.
+	seen := fn.NewSet[asset.PrevID]()
+	for offset := int32(0); offset < numAssets; offset += 2 {
+		constraints.CoinLimit = 2
+		constraints.CoinOffset = offset
+
+		coins, err := assetsStore.ListEligibleCoins(ctx, constraints)
+		require.NoError(t, err)
+
+		expectedLen := min(2, numAssets-int(offset))
+		require.Len(t, coins, expectedLen)
+
+		for _, c := range coins {
+			require.Equal(t, uint64(amount), c.Asset.Amount)
+
+			// A coin must not show up on two pages.
+			require.False(t, seen.Contains(c.PrevID()))
+			seen.Add(c.PrevID())
+		}
+	}
+
+	// And no coin may have been skipped either.
+	require.Equal(t, numAssets, len(seen.ToSlice()))
+}
+
 // TestAssetExportLog tests that were able to properly spend/transfer assets on
 // disk. This ensures we can properly commit the end result of an asset
 // transfer initiated at a higher level.
--- a/tapdb/sqlc/queries/assets.sql
+++ b/tapdb/sqlc/queries/assets.sql
@@ -527,12 +527,18 @@ WHERE (
     @num_limit >= 0 AND
     @num_offset >= 0 AND
     COALESCE(sqlc.narg('sort_direction'), 0) >= 0 AND
+    COALESCE(sqlc.narg('order_by_amount'), 0) >= 0 AND
     -- The script_key_type argument must NEVER be an empty slice, otherwise this
     -- query will return no results.
     COALESCE(script_keys.key_type, 0) IN
       (sqlc.slice('script_key_type')/*SLICE:script_key_type*/)
 )
+-- The trailing sort by asset_id is what makes this a total order, which the
+-- paging of a bounded listing relies on: without it, rows that compare equal
+-- on the optional terms above could come back in any order and a later page
+-- could repeat or skip them. Any ordering term added here must go above it.
 ORDER BY
+    CASE WHEN COALESCE(sqlc.narg('order_by_amount'), 0) = 1 THEN assets.amount END DESC,
     CASE WHEN COALESCE(sqlc.narg('sort_direction'), 0) = 1 THEN assets.asset_id END DESC,
     assets.asset_id ASC
 LIMIT @num_limit OFFSET @num_offset;
--- a/tapfreighter/coin_select.go
+++ b/tapfreighter/coin_select.go
@@ -2,14 +2,26 @@ package tapfreighter
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sort"
 	"sync"
 	"time"
 
 	"github.com/btcsuite/btcd/wire/v2"
+	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/commitment"
 	"github.com/lightninglabs/taproot-assets/fn"
+)
+
+const (
+	// eligibleCoinsPageSize is the number of coins we fetch per query when
+	// listing eligible coins for a send. Coins are listed in descending
+	// amount order, so in most cases a single page is enough to cover the
+	// target amount. Listing a coin is expensive, as the full commitment
+	// of its anchor output is reconstructed, so we only fetch more pages
+	// when the previous ones can't cover the target.
+	eligibleCoinsPageSize = 32
 )
 
 // NewCoinSelect creates a new CoinSelect.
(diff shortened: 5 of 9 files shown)


tapdb/assets_store_test.go
```diff
@@ -1685,6 +1685,99 @@ func TestSelectCommitment(t *testing.T) {
...
+	t.Parallel()
+
+	const numAssets = 5
+
+	assetGen := newAssetGenerator(t, numAssets, 1)
+	_, assetsStore, _ := newAssetStore(t)
+
+	// We create one asset per anchor point, with amounts that are neither
+	// sorted nor aligned with the anchor point order.
+	amounts := []uint64{10, 40, 20, 50, 30}
```
litbot-9000, line 1884: Every amount here is distinct, so nothing in this PR exercises the case that actually makes offset paging safe: **coins of equal amount**.

`ORDER BY amount DESC` alone is not a total order. What saves this design is the unconditional `assets.asset_id ASC` term at the end of the `ORDER BY`, and `assets.asset_id` being `INTEGER PRIMARY KEY`. I confirmed that works today — a throwaway version of this test with five coins all of amount `10`, paged `CoinLimit=2` over offsets 0/2/4, returned five distinct coins with no duplicates and no skips.

The problem is that no test would notice if it stopped working. Concretely: someone adds another optional ordering term to `QueryAssets` (exactly what this PR just did for `order_by_amount`) and places it *after* `assets.asset_id ASC`, or a future refactor drops the trailing term as apparently redundant. `(amount DESC)` alone then leaves equal-amount rows free to come back in any order, page 2 at offset 32 repeats and skips rows relative to page 1, and `SelectCoins` starts silently missing coins. On the success path that just means extra inputs; on a wallet of many equal-denomination coins it means `insufficient amount available` on a node that can afford the send. The prev-ID dedup hides the duplicates, and the unbounded retry masks it further, so the symptom would be intermittent and very hard to attribute back here.

A cheap addition: a second block in this test with `amounts` all equal, paging through and asserting the union is `numAssets` distinct prev IDs. That pins the invariant the whole paging scheme rests on.

tapfreighter/coin_select.go
```diff
@@ -114,6 +148,95 @@ func (s *CoinSelect) SelectCoins(ctx context.Context,
...
+			listing.amountSum += c.Asset.Amount
+		}
+
+		// We stop listing once the compatible coins cover the target
+		// amount, once we've exhausted the eligible coins (which an
+		// unbounded or partial page indicates), or once a page no
+		// longer adds any coin we haven't already seen, which also
+		// guarantees that this loop terminates.
+		if listing.amountSum >= targetAmt || pageSize == 0 ||
+			numListed < int(pageSize) || len(page) == 0 {
```
litbot-9000, line 242: `numListed < int(pageSize)` is the exhaustion test, and its soundness depends on an invariant that lives in another package and is not written down anywhere: **`ListEligibleCoins` must return exactly one `AnchoredCommitment` per row the SQL `LIMIT` returned.**

I checked and it currently holds — `queryCommitments` builds `selectedAssets` with `make(..., len(matchingAssets))` and fills it index by index, with no Go-side filtering between the query and the return. But it is only true by construction, and this is the one place where breaking it produces a false `insufficient funds` rather than a slowdown.

The failure mode if it ever stops holding: suppose a post-query filter is added in `queryCommitments` (skipping a coin whose proof is missing, or one whose anchor commitment fails to reconstruct — that path currently errors, but "skip it" is a plausible future softening). A full SQL page of 32 rows comes back as 31 commitments, `numListed` is 31, `31 < 32` reads as "exhausted", and the loop returns after page 1. Because `numPages == 1`, the L99 retry does **not** fire, and a node with 500 coins reports insufficient funds after looking at 32 of them.

Worth a sentence on `listCoins` (or on `CoinLister.ListEligibleCoins` in `interface.go`) stating that a bounded listing must return a short page only when the eligible set is genuinely exhausted, so the next person to touch `queryCommitments` knows this is load-bearing. Alternatively, comparing against `constraints.CoinLimit` at the tapdb boundary rather than the returned slice length would make it self-enforcing.

tapdb/assets_store.go
```diff
@@ -2375,6 +2383,16 @@ func (a *AssetStore) ListEligibleCoins(ctx context.Context,
 	// have a block height of 0, so we set the minimum block height to 1.
 	assetFilter.MinAnchorHeight = sqlInt32(1)
 
+	// If the caller bounds the number of coins, we return them in
+	// descending amount order, so the largest eligible coins are returned
+	// first and a bounded listing can cover a target amount with as few
+	// pages as possible.
+	if constraints.CoinLimit > 0 {
+		assetFilter.NumLimit = constraints.CoinLimit
```
litbot-9000, line 2448: `AssetQueryFilters` now carries two independent pagination mechanisms and this line silently resolves the conflict between them.

`constraintsToDbFilter` has already mapped `query.Offset`/`query.Limit` into `NumOffset`/`NumLimit` (the fields `FetchAllAssets` and friends use). These two lines then overwrite both from the embedded `CommitmentConstraints.CoinLimit`/`CoinOffset`. So `AssetQueryFilters` has four pagination fields where two of them win by virtue of being assigned later.

No live bug — `ListEligibleCoins` constructs `AssetQueryFilters{CommitmentConstraints: constraints}` with `Limit`/`Offset` left zero, and nothing else sets `CoinLimit`. But a future caller that populates an `AssetQueryFilters` with both a `Limit` and (via the embedded struct) a `CoinLimit` gets one of them dropped with no error and no log line, and the resulting bug is an off-by-a-page in a query result — exactly the kind that surfaces as a wrong balance rather than a crash.

Either a short comment noting that `CoinLimit`/`CoinOffset` take precedence over `Limit`/`Offset` for this query, or a guard rejecting the combination the way you (rightly) reject `CoinLimit` + `PrevIDs` above, would close it.

docs/release-notes/release-notes-0.9.0.md
```diff
@@ -91,6 +91,13 @@
 
 ## Performance Improvements
 
+* [PR#2252](https://github.com/lightninglabs/taproot-assets/pull/2252)
+  bounds input selection when funding a send. Eligible coins are now
```
litbot-9000, line 103: This advertises the speedup but omits the regression you disclose in the PR description: a send that **cannot** be funded now takes roughly 1.7x longer than before (16.62s → 28.93s on your own measurement node), because a multi-page listing that falls short is repeated unbounded.

That is a defensible trade — it buys back the accuracy of the error, it is confined to a failure path, and #2251 more than repays it. But it is user-visible on exactly the operation a user is most likely to retry in frustration, and node operators reading release notes to explain a support ticket will not find it here. One clause would do it, e.g. "sends that cannot be funded are repeated unbounded before reporting insufficient funds, so that path is somewhat slower".

While here: it is also worth saying that sends with **pinned inputs** (`PrevIDs`) are excluded from the bound and keep the previous cost, since that is the natural question a reader has after reading this entry.

tapfreighter/coin_select.go
```diff
@@ -58,39 +70,61 @@ func (s *CoinSelect) SelectCoins(ctx context.Context,
...
-		return nil, ErrMatchingAssetsNotFound
+	// Each page is listed by a separate query, so the coin set can shift
+	// between two pages and move a coin out of a page we already read past.
+	// That can only ever hide coins from us, never invent ones, but it
+	// could make us report insufficient funds while the funds are actually
+	// there. So if a listing that took more than one page couldn't cover
+	// the target amount, we repeat it unbounded to be sure. A listing that
+	// took a single page needs no such check, as no offset was applied to
+	// it, so it saw the full coin set at that instant.
+	if listing.numPages > 1 && listing.amountSum < constraints.MinAmt {
```
litbot-9000, line 99: Verified rather than flagged: this is the load-bearing guard, and I could not break it.

The `numPages > 1` condition is doing real work, and the reasoning in the comment above is right for a reason worth stating explicitly. `numPages == 1 && amountSum < target` is only reachable when the *first* page came back partial or empty. On page 1 the offset is zero and `seenCoins` is empty, so the dedup filter is a no-op and `len(page) == 0` cannot fire on a full page — meaning a single-page listing that fell short really did see the entire eligible set at one instant, and skipping the retry is safe. If a future change ever makes the dedup filter able to shrink page 1 (say, seeding `seenCoins` from somewhere), that argument collapses and this becomes a false-`insufficient funds` path.

The other direction also checks out: replacing `listing` wholesale with the unbounded result is fine, since `selectForAmount` sorts descending itself and does not depend on the paged arrival order.

tapfreighter/coin_select.go
```diff
@@ -114,6 +148,95 @@ func (s *CoinSelect) SelectCoins(ctx context.Context,
...
+		for _, c := range compatible {
+			listing.amountSum += c.Asset.Amount
+		}
+
+		// We stop listing once the compatible coins cover the target
+		// amount, once we've exhausted the eligible coins (which an
+		// unbounded or partial page indicates), or once a page no
+		// longer adds any coin we haven't already seen, which also
+		// guarantees that this loop terminates.
+		if listing.amountSum >= targetAmt || pageSize == 0 ||
```
litbot-9000, line 241: Verified rather than flagged, since this condition is where a bug would hurt most.

The important subtlety is that `numListed` is captured *before* both the dedup filter and the compatibility filter, so a page consisting entirely of coins that are incompatible with `maxVersion` still counts as a full page and paging continues. That is what stops the loop from concluding "insufficient funds" while spendable coins sit on page 3 behind 64 incompatible ones. It is easy to imagine this being "cleaned up" later into `len(page) < int(pageSize)` after the filters, which reads more naturally and would reintroduce exactly that bug on a node holding a mix of V1 and V2 commitments. A brief `// NOTE:` on why the raw count is used, not the filtered one, would be cheap insurance — the existing comment at L216-217 says *that* it is the raw count but not *why* it must be.

Termination itself is fine: `len(page) == 0` catches the all-duplicates page that a large enough insertion between queries could produce, and `offset` is `int32` against a page size of 32, so the counter is not a practical concern.

tapfreighter/coin_select.go
```diff
@@ -58,39 +70,61 @@ func (s *CoinSelect) SelectCoins(ctx context.Context,
...
-		ctx, listConstraints,
+
+	// We list the eligible coins in pages of descending amounts, so we can
+	// stop listing as soon as the accumulated amount covers the target,
+	// instead of loading every coin the node holds. If specific inputs are
+	// requested, we can't bound the listing, as those are filtered out of
+	// the full listing.
+	pageSize := int32(eligibleCoinsPageSize)
+	if len(constraints.PrevIDs) > 0 {
+		pageSize = 0
```
litbot-9000, line 81: Confirms there is no behaviour regression for pinned inputs, which was my main worry about the new rejection in `ListEligibleCoins`: this is the only place that could produce the `CoinLimit > 0 && len(PrevIDs) > 0` combination, and it makes it impossible. `listCoins` is the sole non-test caller of `ListEligibleCoins` in the tree, and `PrevIDs` only ever originates from `wallet.go:291` when a user explicitly pins inputs, so no previously-working RPC can start erroring. The tapdb guard is purely defensive, which is the right shape for it.

The cost is that pinned-input sends keep the full O(total holdings) listing and get none of this PR's improvement — on the 621-coin node from your table, that is still ~16.6s. Not something to solve here, but it is the obvious follow-up (a keyset cursor over `(amount, assets.asset_id)`, as you suggest in the description, would let pinned listings be bounded too, since a cursor cannot skip rows). Worth an issue so it does not get lost.

tapdb/sqlc/queries/assets.sql
```diff
@@ -527,12 +527,14 @@ WHERE (
...
     @num_offset >= 0 AND
     COALESCE(sqlc.narg('sort_direction'), 0) >= 0 AND
+    COALESCE(sqlc.narg('order_by_amount'), 0) >= 0 AND
     -- The script_key_type argument must NEVER be an empty slice, otherwise this
     -- query will return no results.
     COALESCE(script_keys.key_type, 0) IN
       (sqlc.slice('script_key_type')/*SLICE:script_key_type*/)
 )
 ORDER BY
+    CASE WHEN COALESCE(sqlc.narg('order_by_amount'), 0) = 1 THEN assets.amount END DESC,
```
litbot-9000, line 541: Checked this closely because unstable ordering under `LIMIT`/`OFFSET` is the classic way paging silently loses rows, and it holds up.

When `order_by_amount` is NULL this term evaluates to `NULL` for **every** row, so it cannot reorder anything and the query keeps its previous behaviour — and because all rows are NULL, the SQLite/Postgres disagreement about `NULLS FIRST` vs `NULLS LAST` under `DESC` cannot bite either. That matches your claim that the term is inert for other consumers. When it is `1`, `assets.amount` is `NOT NULL`, so no NULLs enter the comparison at all.

The generated `assets.sql.go` matches this source exactly, including placing `order_by_amount` at `$17` before the `sqlc.slice` expansion as the comment above the `WHERE` block requires, and reusing `$17` in both the guard and the `ORDER BY`. Since there is one Postgres-dialect source rewritten to `?` for SQLite in `db_custom.go`, there is no second dialect file to drift out of sync. I could not run `make gen` (no container runtime), so this is a read-only verification, but `Sqlc check` is green in CI.

One note for whoever touches this next: the trailing `assets.asset_id ASC` is now doing double duty as the tiebreak that makes offset paging deterministic, not just as a default sort. Any new ordering term must go *above* it, never below.

litbot-9000 (commented):
Reviewed commit-by-commit at `7e544e9`, built and tested locally. **I could not find a path where the bounded listing reports insufficient funds on a node that can afford the send** — the terminating argument holds, and the `assets.asset_id ASC` tiebreak that makes offset paging safe is real (I verified it empirically, see below). The design is sound and the PR description is unusually honest about its own costs. My findings are one genuine test-coverage gap that leaves the load-bearing invariant unguarded, one undocumented cross-package invariant, and two minor notes. Nothing I would block on.

### Can a bounded listing miss a spendable coin?

**Item A — no.** `listCoins` continues paging iff *all* of: `amountSum < target`, `pageSize != 0`, `numListed == pageSize`, and the page added at least one unseen coin. I walked each exit:

* A full page of coins that are all *incompatible* (`Commitment.Version > maxVersion`) leaves `amountSum` untouched but does **not** stop the loop — `numListed` is the raw row count, taken before both the dedup filter and the compatibility filter. So compatible coins sitting on page 3 behind 64 incompatible ones are still reached. This is the case I most expected to be wrong; it is covered by the new `incompatible coins on first page` case.
* A target needing more than 32 inputs simply pages on.
* `numPages == 1 && amountSum < target` is only reachable via a partial or empty *first* page (at offset 0 dedup is a no-op, so `len(page) == 0` cannot fire on page 1 with a full page). That listing saw the whole eligible set in one snapshot, so "insufficient" is truthful and needs no retry. The claim in the comment at L96-98 checks out.
* Any `numPages > 1` shortfall triggers the unbounded re-listing at L99, so a coin shifted out of a window we already read past cannot produce a false negative.
* `selectForAmount` cannot fail once `amountSum >= target`, since it sums the same slice.
* There is no `maxInputs`-style cap anywhere in the repo (`grep`ed), so nothing interacts with paging there.

**Item B — the tiebreak is present and it works.** `ORDER BY ... assets.asset_id ASC` is an unconditional final term, and `assets.asset_id` is `INTEGER PRIMARY KEY` (`migrations/000002_assets.up.sql:187`), so `(amount DESC, asset_id ASC)` is a total order. I did not take that on faith: I wrote a throwaway `tapdb` test with five coins of **identical** amount, paged `CoinLimit=2` over offsets 0/2/4, and got five distinct coins with zero duplicates and zero skips. This is *not* the unordered-query defect I have flagged on this repo before. Mutation between pages is a separate matter and is handled by prev-ID dedup plus the unbounded retry; I could not construct a shift that survives both.

The one thing that keeps A sound is not obvious from `tapfreighter`, and nothing enforces it — see my comment on L233.

### Pinned `PrevIDs`

No previously-working request now fails. `SelectCoins` sets `pageSize = 0` whenever `len(PrevIDs) > 0` (L79-82), and `listCoins` is the only non-test caller of `ListEligibleCoins` in the tree, so the new `coin limit cannot be combined with specific prev IDs` error is unreachable from production code — it is a guard against a future caller, which is the right call. `PrevIDs` is only ever populated from `tapfreighter/wallet.go:291` when a user explicitly pins inputs. Those sends keep the old unbounded cost and get none of this PR's speedup; worth a sentence in the release note, not a blocker.

### What I verified

* `make build` — clean (`tapd`, `tapcli`).
* `go test ./tapfreighter/...` — ok, 0.112s. `go test -race ./tapfreighter/... -count=1` — ok, 2.216s.
* `go test ./tapdb/... -run 'TestListEligibleCoinsBounded|TestSelectCommitment|TestFetchAllAssetsPagination' -count=1` — ok, 4.335s.
* Throwaway equal-amount paging test described above — passed (5/5 distinct), then deleted.
* Read `tapdb/sqlc/assets.sql.go` against `tapdb/sqlc/queries/assets.sql` line by line: the `$17` guard, the new `ORDER BY` term and the `QueryAssetsParams` field match the source, and `order_by_amount` is placed before the `sqlc.slice` expansion as the existing comment requires. There is a single Postgres-dialect query source rewritten to `?` for SQLite in `db_custom.go`, so there is no second dialect file that could drift.
* Traced `queryCommitments` end to end: `selectedAssets` is built 1:1 with `matchingAssets`, with no Go-side drop, so `numListed` faithfully reflects the SQL page size.
* `grep` of all migrations: there is **no index on `assets.amount`**, so `ORDER BY amount DESC` sorts the whole filtered join on every page. The win is entirely in the per-coin commitment reconstruction, not the SQL scan — consistent with your framing, and it explains the insufficient-funds regression.
* **No container runtime is available in my environment** (no docker/podman/psql), so I could not run `test_db_postgres` or any `ory/dockertest` path. **Nothing here should be read as local Postgres coverage.** I am relying on CI for that verdict: `gh pr checks 2252` is fully green, including `run unit tests (unit dbbackend=postgres pkg=tapdb)`, all four `run itests postgres` shards, `Sqlc check` and `Lint check`.
* Cross-checked #2251: it does not touch `QueryAssets` (its query changes start at `assets.sql:839`) and does not touch `ListEligibleCoins`, so the two merge cleanly. More usefully, #2251 explicitly documents and preserves the index-alignment of `queryChainAssets`, which is the same invariant this PR's exhaustion check leans on — the two are genuinely complementary, and neither breaks the other's assumptions.

Findings inline, most severe first.

GeorgeTsagk:
All four addressed.

1. Added `TestListEligibleCoinsBoundedEqualAmounts`: five coins of the same amount, paged two at a time, asserting every page is full until the coins run out and that the union is five distinct prev IDs, so no coin repeats or is skipped when the amount ordering alone doesn't decide the order.
2. Documented the invariant on `CoinLister.ListEligibleCoins` and on `listCoins`: a bounded listing may only return a short page once the eligible coins are exhausted, since that is how the caller learns paging is done, and an implementation that drops coins after querying them would make a bounded listing look exhausted while coins remain.
3. Noted at the `CoinLimit`/`CoinOffset` assignment that it overrides the `Limit`/`Offset` pagination the filters may already carry, and why coins are paged with the former.
4. The release note now mentions both that sends with pinned inputs keep listing every eligible coin, and that a send which cannot be funded is slower because the listing is repeated unbounded.

GeorgeTsagk:
Two more from the same review, now pushed:

1. Added a note on why the exhaustion check uses the raw page count rather than the filtered one, since a page made up entirely of already-seen or incompatible coins is still a full page and stopping on it would report insufficient funds with spendable coins on a later page.
2. Recorded above the `ORDER BY` in `QueryAssets` that the trailing `assets.asset_id ASC` is what makes it a total order, that bounded paging depends on it, and that any new ordering term goes above it.

Left the pinned-input listing open: agreed it should be an issue rather than folded in here.

tapfreighter/coin_select.go
GeorgeTsagk: Resolving this one as no change here: the unbounded listing for pinned inputs stays as it is in this PR, and the keyset cursor that would let it be bounded is tracked as a follow-up rather than folded in.

jtobin (approved):
LGTM. :+1: :+1: 

FWIW, Fable made one comment that caught my attention:

> There's a simpler variant that solves the same problem: instead of offset paging, re-query from offset zero with a growing limit (32, then unbounded, or geometric doubling). Every query is then a self-consistent snapshot prefix, which deletes the entire cross-snapshot hazard class this PR has to manage: the CoinOffset field, the prev-ID de-dup set, the shifted-coin skip hazard, the unbounded fallback pass, and the three tests that exercise them. Cost is a bounded constant factor (re-reconstructing the earlier prefix on the rare retry, ≤2x with doubling) while staying O(coins needed) rather than O(holdings).
>
> The implemented design is sound and the hazards are handled correctly, so this is a comment-level observation, not a blocker — but it's the kind of complexity that must be re-reasoned about on every future touch.

This reco looks interesting, in that the design seems to be simpler, with probably ~equivalent performance in practice. But given the implementation here actually exists, and looks correct, I think we're fine to go with it.

GeorgeTsagk:
> This reco looks interesting, in that the design seems to be simpler, with probably ~equivalent performance in practice. But given the implementation here actually exists, and looks correct, I think we're fine to go with it.

Worth revisiting when coin selection becomes the next biggest bottleneck. A bit too much engineering here IMO plus it's not a clear win, we will be fetching duplicates by the ever-growing overlapping page.

Merging this version.

PR #2255, opened 2026-08-25
universe: only log NEW_PROOF events for leaves we didn't already have

### LINKED ISSUE #2253: [bug]: duplicate proof inserts are logged as NEW_PROOF, inflating universe stats
### Problem

`Archive.UpsertProofLeafBatch` logs a `NEW_PROOF` event for every item it was
handed, not for the leaves that were actually new:

```go
// universe/archive.go:609 (main @ aea668c7)
// Log a sync event for the newly inserted leaf in the background as an
// async goroutine.
ids := fn.Map(items, func(item *Item) Identifier {
	return item.ID
})
```

`Archive.UpsertProofLeaf` does the same unconditionally at `archive.go:364`.
Re-registering a leaf the node already holds is a no-op upsert, but it still
writes an event row.

Two consequences:

1. `UniverseStats.num_total_proofs` overcounts. The `universe_stats` view
   derives `total_asset_proofs` as `COUNT(*)` over
   `universe_events WHERE event_type = 'NEW_PROOF'`, so the reported figure is
   insert attempts, not proofs held.
2. `universe_events` has no retention and is never pruned, so the table grows
   with sync traffic rather than with universe size, and the view aggregates
   the whole table on every stats read.

### Data

From a long-running regtest federation (postgres backend, ~6 days, tapd
v0.8.1):

```
universe_events (NEW_PROOF)   1,324,186 rows
universe_leaves                  11,126 rows
worst single root               460,720 events / 2,502 leaves
```

`tapcli universe stats` on that node:

```json
{"num_total_proofs":"1324096","num_total_assets":"10"}
```

Caveat on the magnitude: most of this amplification comes from the
pointer-identity diff bug fixed in bd692080 (#2187, not in v0.8.1 or the
v0.8.2 RCs, so released versions still re-fetch a peer's whole leaf set on
every root mismatch). With that fix in place the rate drops sharply. The
residual is still real: any peer pushing a leaf we already hold inflates the
stat, and nothing bounds the table over time.

### Suggested fix

Have `UpsertProofLeaf`/`UpsertProofLeafBatch` report which leaves were newly
inserted, and log events only for those. Retention for `universe_events` is a
separate question but worth considering for long-lived universe servers.

Related: #1633 covers query-plan optimization for `QueryUniverseAssetStats`,
which is a different concern from the row growth and the double counting here.

GeorgeTsagk:
`Archive.UpsertProofLeafBatch` logged a `NEW_PROOF` event for every item it was
handed, not for the leaves that were actually new. Re-registering a leaf we
already hold is a no-op upsert, but it still wrote an event row.

`universe_stats.total_asset_proofs` was a `COUNT(*)` over those rows, so
`num_total_proofs` reported insert attempts rather than proofs held, and
`universe_events` grew with sync traffic instead of with universe size. On the
node from #2253 that meant 1.3M event rows for 11k leaves, a reported count of
1,324,096, and the stats view taking 437ms per read on postgres where counting
the real leaves takes 0.34ms.

Two halves:

**Stop writing the duplicate events.** The universe leaf upsert now reports
whether it created a leaf or just replaced the proof of one we already had, and
both the batch and the single-leaf path take that answer from inside the write
transaction. Only new leaves get an event. Determining it in the write tx also
closes a race the read-then-write version would have kept: two clients
registering the same new leaf could both miss on their own read, both verify,
and both log, while only one write actually created the leaf.

**Stop deriving the count from the events.** Migration 66 redefines
`universe_stats` to count `universe_leaves` instead, so the figure means what
its name says and is correct on databases that already accumulated duplicate
events. No event rows are deleted: `QueryAssetStatsPerDay` still reads them.

The supply commitment sub-trees keep their leaves in the same table but were
never counted here, having never produced a NEW_PROOF event, so the new count
filters on proof type to keep them out. Verified against a supply tree: 6
leaves across burn, ignore and mint_supply, 0 events.

The added existence lookup is a single hit on an existing unique index and
doesn't move insert timings on either backend.

Behaviour changes, both in the release notes:

- `num_total_proofs` can now decrease when a proof leaf is deleted from a
  universe that still holds others. The event rows previously kept it flat.
- `new_proof_events` in `QueryEvents` is fed by the event rows and now counts
  leaves acquired per day rather than registration attempts per day, so on a
  long-running node the series steps down. `sync_events` is unchanged and still
  carries traffic volume.

Nothing here crosses the wire: no proto, no proof encoding, and the
remote-facing `Registrar` interface is untouched. `universe_events` and the
stats view are local bookkeeping, and no sync or federation path reads them, so
a patched node and an unpatched one interoperate exactly as before.

Fixes #2253

### DIFF
--- a/tapdb/multiverse.go
+++ b/tapdb/multiverse.go
@@ -1072,14 +1072,21 @@ func journalUniverseLeaves(ctx context.Context, db BaseUniverseStore,
 // superseded this one before the multiverse update was flushed, the
 // receipt is rebuilt from the newer state, so its universe root may be
 // fresher than the root this call's own transaction committed.
+//
+// The returned boolean reports whether the upsert created a new universe leaf,
+// as opposed to replacing the proof of one we already held. It is determined
+// inside the write transaction, so it stays correct under concurrent inserts
+// of the same leaf: serializable isolation makes exactly one of them the
+// creator. It is set in the pending case too, as the leaf is committed by then.
 func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 	id universe.Identifier, key universe.LeafKey, leaf *universe.Leaf,
-	metaReveal *proof.MetaReveal) (*universe.Proof, error) {
+	metaReveal *proof.MetaReveal) (*universe.Proof, bool, error) {
 
 	var (
-		writeTx    BaseMultiverseOptions
-		uniProof   *universe.Proof
-		rootStatus universeRootStatus
+		writeTx      BaseMultiverseOptions
+		uniProof     *universe.Proof
+		rootStatus   universeRootStatus
+		leafInserted bool
 	)
 
 	execTxFunc := func(dbTx BaseMultiverseStore) error {
@@ -1099,6 +1106,7 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 			return fmt.Errorf("failed universe upsert: %w", err)
 		}
 		rootStatus = res.rootStatus
+		leafInserted = res.leafInserted
 
 		if shouldJournalLeaf(id) {
 			err = journalUniverseLeaves(
@@ -1113,7 +1121,7 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 	}
 	dbErr := b.db.ExecTx(ctx, &writeTx, execTxFunc)
 	if dbErr != nil {
-		return nil, dbErr
+		return nil, leafInserted, dbErr
 	}
 
 	// The universe leaf is now durably committed, so run the
@@ -1157,7 +1165,8 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 	// update in the background.
 	update, err := b.rootCoalescer.updateRoot(ctx, id)
 	if err != nil {
-		return nil, fmt.Errorf("failed multiverse upsert: %w", err)
+		return nil, leafInserted, fmt.Errorf("failed multiverse "+
+			"upsert: %w", err)
 	}
 
 	// If a concurrent insert into the same universe committed between
@@ -1176,17 +1185,17 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 	if !mssmt.IsEqualNode(update.universeRoot, uniProof.UniverseRoot) {
 		proofs, err := b.fetchProofLeaf(ctx, id, key, true)
 		if err != nil {
-			return nil, fmt.Errorf("%w: failed superseded "+
-				"proof fetch: %w",
+			return nil, leafInserted, fmt.Errorf("%w: failed "+
+				"superseded proof fetch: %w",
 				universe.ErrMultiversePending, err)
 		}
 		if len(proofs) != 1 {
-			return nil, fmt.Errorf("%w: expected one proof "+
-				"for superseded upsert, got %d",
+			return nil, leafInserted, fmt.Errorf("%w: expected "+
+				"one proof for superseded upsert, got %d",
 				universe.ErrMultiversePending, len(proofs))
 		}
 
-		return proofs[0], nil
+		return proofs[0], leafInserted, nil
 	}
 
 	// Populate the multiverse fields in the proof object now that the
@@ -1194,7 +1203,7 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 	uniProof.MultiverseRoot = update.multiverseRoot
 	uniProof.MultiverseInclusionProof = update.inclusionProof
 
-	return uniProof, nil
+	return uniProof, leafInserted, nil
 }
 
 // UpsertProofLeafBatch upserts a proof leaf batch within the multiverse tree
--- a/tapdb/multiverse.go
+++ b/tapdb/multiverse.go
@@ -1087,10 +1087,10 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 		// tree. The block height is extracted from the decoded proof
 		// by universeUpsertProofLeaf itself.
 		var (
-			err    error
-			leafID int64
+			res universeUpsertResult
+			err error
 		)
-		uniProof, rootStatus, leafID, err = universeUpsertProofLeaf(
+		uniProof, res, err = universeUpsertProofLeaf(
 			ctx, dbTx, id.String(), id.ProofType,
 			id.GroupKey, key, leaf, metaReveal,
 			lfn.None[uint32](),
@@ -1098,10 +1098,11 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 		if err != nil {
 			return fmt.Errorf("failed universe upsert: %w", err)
 		}
+		rootStatus = res.rootStatus
 
 		if shouldJournalLeaf(id) {
 			err = journalUniverseLeaves(
-				ctx, dbTx, []int64{leafID},
+				ctx, dbTx, []int64{res.leafID},
 			)
 			if err != nil {
 				return err
@@ -1206,13 +1207,17 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 // coalescer retries the updates in the background until they commit,
 // and ReconcileMultiverse repairs any update abandoned by a shutdown
 // at the next startup.
+//
+// The returned slice holds the subset of items that resulted in a new universe
+// leaf; items the universe already held are left out.
 func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
-	items []*universe.Item) error {
+	items []*universe.Item) ([]*universe.Item, error) {
 
 	var (
 		writeTx      BaseMultiverseOptions
 		uniProofs    []*universe.Proof
 		rootStatuses []universeRootStatus
+		newItems     []*universe.Item
 	)
 	// Track the universes the batch touches, in first-touch order, so
 	// the shared multiverse tree is refreshed once per universe rather
@@ -1225,6 +1230,7 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 			rootStatuses = make(
 				[]universeRootStatus, len(items),
 			)
+			newItems = make([]*universe.Item, 0, len(items))
 
 			dirtyUniverses = make(
 				[]universe.Identifier, 0, len(items),
@@ -1239,7 +1245,7 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 				// start with. The block height is extracted
 				// from the decoded proof by
 				// universeUpsertProofLeaf itself.
-				uniProof, status, leafID, err :=
+				uniProof, res, err :=
 					universeUpsertProofLeaf(
 						ctx, store, item.ID.String(),
 						item.ID.ProofType,
@@ -1253,7 +1259,11 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 						idx, err)
 				}
 				uniProofs[idx] = uniProof
-				rootStatuses[idx] = status
+				rootStatuses[idx] = res.rootStatus
+
+				if res.leafInserted {
+					newItems = append(newItems, item)
+				}
 
 				key := item.ID.String()
 				if _, ok := seen[key]; !ok {
@@ -1265,7 +1275,7 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 
 				if shouldJournalLeaf(item.ID) {
 					journalIDs = append(
-						journalIDs, leafID,
+						journalIDs, res.leafID,
 					)
 				}
 			}
@@ -1276,7 +1286,7 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 		},
 	)
 	if dbErr != nil {
-		return dbErr
+		return nil, dbErr
 	}
 
 	// The universe leaves are now durably committed, so run the
@@ -1351,12 +1361,16 @@ func (b *MultiverseStore) UpsertProofLeafBatch(ctx context.Context,
 	// fails, the universe leaves above remain committed, and the
 	// coalescer keeps retrying the multiverse updates in the
 	// background.
+	//
+	// The new leaves are reported either way: they are committed, so they
+	// count towards the universe stats even while their multiverse entry
+	// is still outstanding.
 	err := b.rootCoalescer.updateRoots(ctx, dirtyUniverses)
 	if err != nil {
-		return fmt.Errorf("failed multiverse upsert: %w", err)
+		return newItems, fmt.Errorf("failed multiverse upsert: %w", err)
 	}
 
-	return nil
+	return newItems, nil
 }
 
 // DeleteUniverse delete an entire universe sub-tree.
--- a/tapdb/sqlc/queries/universe.sql
+++ b/tapdb/sqlc/queries/universe.sql
@@ -36,6 +36,13 @@ WHERE universe_root_id = (SELECT id FROM root_id);
 DELETE FROM universe_roots
 WHERE namespace_root = @namespace_root;
 
+-- name: QueryUniverseLeafID :one
+SELECT id
+FROM universe_leaves
+WHERE leaf_node_namespace = @namespace
+  AND minting_point = @minting_point
+  AND script_key_bytes = @script_key_bytes;
+
 -- name: UpsertUniverseLeaf :one
 INSERT INTO universe_leaves (
     asset_genesis_id, script_key_bytes, universe_root_id, leaf_node_key,
--- a/tapdb/universe.go
+++ b/tapdb/universe.go
@@ -32,6 +32,10 @@ type (
 	// insertion journal.
 	InsertUniverseLeafJournal = sqlc.InsertUniverseLeafJournalParams
 
+	// UniverseLeafIDQuery is used to look up the primary key of a single
+	// universe leaf.
+	UniverseLeafIDQuery = sqlc.QueryUniverseLeafIDParams
+
 	// NewUniverseRoot is used to insert a new universe root.
 	NewUniverseRoot = sqlc.UpsertUniverseRootParams
 
@@ -117,6 +121,12 @@ type BaseUniverseStore interface {
 	// key.
 	FetchUniverseRoot(ctx context.Context,
 		namespace string) (UniverseRoot, error)
+
+	// QueryUniverseLeafID returns the primary key of the universe leaf
+	// identified by the given namespace, minting point and script key. It
+	// returns sql.ErrNoRows if no such leaf exists.
+	QueryUniverseLeafID(ctx context.Context,
+		arg UniverseLeafIDQuery) (int64, error)
 
 	// UpsertUniverseLeaf upserts a Universe leaf in the database,
 	// returning the leaf's primary key.
@@ -785,7 +795,7 @@ func (b *BaseUniverseTree) UpsertProofLeaf(ctx context.Context,
 
 		// The block height is extracted from the decoded proof by
 		// universeUpsertProofLeaf itself.
-		issuanceProof, _, leafID, err := universeUpsertProofLeaf(
+		issuanceProof, res, err := universeUpsertProofLeaf(
 			ctx, dbTx, namespace, b.id.ProofType,
 			b.id.GroupKey, key, leaf, metaReveal,
 			lfn.None[uint32](),
@@ -793,7 +803,6 @@ func (b *BaseUniverseTree) UpsertProofLeaf(ctx context.Context,
 		if err != nil {
 			return fmt.Errorf("failed universe upsert: %w", err)
 		}
-
 		err = upsertMultiverseLeafEntry(
 			ctx, dbTx, b.id, issuanceProof.UniverseRoot,
 		)
@@ -814,7 +823,7 @@ func (b *BaseUniverseTree) UpsertProofLeaf(ctx context.Context,
 
 		if shouldJournalLeaf(b.id) {
 			err = journalUniverseLeaves(
-				ctx, dbTx, []int64{leafID},
+				ctx, dbTx, []int64{res.leafID},
 			)
 			if err != nil {
 				return err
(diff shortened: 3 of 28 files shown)


tapdb/universe_stats_test.go
```diff
@@ -139,6 +139,62 @@ func (u *uniStatsHarness) addEvents(numAssets int) {
...
+		ids := fn.Map(
+			newItems, func(i *universe.Item) universe.Identifier {
+				return i.ID
+			},
+		)
+		if len(ids) == 0 {
+			continue
+		}
+
+		require.NoError(t, statsDB.LogNewProofEvents(ctx, ids...))
```
litbot-9000, line 188: **Nothing in this PR fails if commit 3 is removed.** This test is the one commit 4's message points at, but it exercises the `tapdb` primitive from commit 2 and then hand-rolls the archive's logging itself — this `LogNewProofEvents(ctx, ids...)` call *is* `universe/archive.go:599-608`, reimplemented in the test. So it asserts that the tapdb API you added behaves correctly, and separately that if a caller uses it correctly the count is right. It never asserts that `Archive` is that caller.

I checked this rather than assuming it:

```
$ git revert --no-commit 83cba1f0     # commit 3, universe/archive.go only, +40/-17
$ go build ./...                       # RC 0 -- commit 3 changes no signatures
$ go test ./universe/... ./tapdb/...
ok  github.com/lightninglabs/taproot-assets/universe                0.109s
ok  github.com/lightninglabs/taproot-assets/universe/supplycommit   (cached)
ok  github.com/lightninglabs/taproot-assets/universe/supplyverifier (cached)
ok  github.com/lightninglabs/taproot-assets/tapdb                  46.289s
```

Full suites, everything green with the actual fix deleted.

Concrete regression that ships silently today: someone doing a merge-conflict resolution on `universe/archive.go` (which this branch *will* need — it conflicts with main right now) resolves `ids := fn.Map(newItems, ...)` back to `fn.Map(items, ...)`, or drops `newItems = append(newItems, newAnchorItems...)` at line 596 so group-anchor leaves stop being counted. Every test in the repo still passes and #2253 quietly reopens.

`universe/archive_test.go` already has a `mockMultiverse` and you changed its `UpsertProofLeafBatch` to `return items, nil` in this PR. Giving it a `mockTelemetry` that records the ids passed to `LogNewProofEvents`, plus a mock multiverse that returns a chosen subset, would pin the archive-level behaviour directly: feed a batch of 4, have the mock report 2 as new, assert exactly those 2 ids get logged. That also covers the `numNew == 0` early return and the anchor/non-anchor concatenation, none of which any test touches today.

tapdb/multiverse.go
```diff
@@ -805,7 +805,7 @@ func (b *MultiverseStore) UpsertProofLeaf(ctx context.Context,
 			return err
 		}
 
-		uniProof, err = universeUpsertProofLeaf(
+		uniProof, _, err = universeUpsertProofLeaf(
```
litbot-9000: The authoritative in-transaction answer gets thrown away here, and `Archive.UpsertProofLeaf` then re-derives the same fact less reliably.

`universeUpsertProofLeaf` computes `leafInserted` at `tapdb/universe.go:944-956` inside the write transaction. The batch path keeps it and is correct *because* of that: `tapdb/interfaces.go:329-333` opens every tx at `sql.LevelSerializable` and `ExecTx` retries on `IsSerializationOrDeadlockError`, so two concurrent batches inserting the same leaf can't both conclude "new" — the loser aborts with 40001, retries, and its pre-lookup then sees the row. (I went in expecting to find a double-count here and this is what refuted it.)

The single-leaf path gets none of that. `universe/archive.go:298-299` derives `leafKnown` from a `FetchProofLeaf` that runs in its own read tx, and then `verifyIssuanceProof` at `:346` — full proof-chain verification — runs before the write. The TOCTOU window is however long verification takes.

Failure: two clients call `InsertProof` (`rpcserver/rpcserver.go:7724`) for the same genuinely-new leaf at the same time; or an `InsertProof` races a federation push (`universe/federation.go:541`) or a syncer group-anchor insert (`universe/syncer.go:396`), all of which land on `Archive.UpsertProofLeaf`. Both `FetchProofLeaf` calls miss → both set `leafKnown = false` → both verify → serializable isolation makes the second write a no-op upsert → **both goroutines log `NEW_PROOF`. Two event rows, one leaf.** Exactly the inflation this PR exists to stop, in the path where the correct signal is now available and unused.

This is pre-existing, not a regression — but the fix is now three lines. Return the bool from `MultiverseStore.UpsertProofLeaf` the way `UpsertProofLeafBatch` does, plumb it through the `MultiverseArchive` interface, and drop `leafKnown` in favour of it. It also removes the asymmetry where the two sibling paths answer the same question by two different mechanisms, which is the sort of thing that drifts apart later.

tapdb/sqlc/queries/universe.sql
```diff
@@ -36,6 +36,13 @@ WHERE universe_root_id = (SELECT id FROM root_id);
 DELETE FROM universe_roots
 WHERE namespace_root = @namespace_root;
 
+-- name: QueryUniverseLeafID :one
+SELECT id
+FROM universe_leaves
```
litbot-9000, line 41: You're now doing a per-leaf existence lookup against `universe_leaves` to decide what to count. Worth asking whether the count should just come from this table in the first place.

`universe_stats.total_asset_proofs` is a `COUNT(*)` over `universe_events` grouped by `universe_root_id` (`migration 000029_ignore_burn_universe.up.sql:104-108`). Redefining that CTE as a count over `universe_leaves` grouped by `universe_root_id` fixes the number on **every** node — including the one in #2253 — the moment it upgrades, with no data migration.

The PR body's argument for deferring is that the event rows "carry the timestamps behind `QueryAssetStatsPerDay`". That's a correct argument against *deleting the rows*, and I'm not disputing it. But it isn't an argument against changing the *derivation*: you can leave `universe_events` entirely alone, keep `QueryAssetStatsPerDay` reading it exactly as it does today, and still stop `total_asset_proofs` from being computed from it. Those are separable.

Concretely, without this, after #2255 merges the node in #2253 still reports `num_total_proofs: 1324096` against 11,126 leaves and still takes 437ms per stats read — the two symptoms actually reported in that issue. The PR says "Fixes #2253", but for the node that filed it, both observable symptoms persist indefinitely. Your own measurement is the argument for it: 0.34ms to count the leaves versus 437ms to aggregate the event table.

If you've already considered and rejected this — say the leaf count and the historical event count are meant to be able to diverge, or a view change has migration constraints I'm not seeing — that's fine, ignore me. But the PR body doesn't mention it, so I want to be sure it was a decision and not an omission.

universe/archive.go
```diff
@@ -565,15 +576,27 @@ func (a *Archive) UpsertProofLeafBatch(ctx context.Context,
...
+	numNew := len(newAnchorItems) + len(newNonAnchorItems)
+	if numNew == 0 {
+		return nil
+	}
+
+	newItems := make([]*Item, 0, numNew)
+	newItems = append(newItems, newAnchorItems...)
+	newItems = append(newItems, newNonAnchorItems...)
+
+	ids := fn.Map(newItems, func(item *Item) Identifier {
```
litbot-9000, line 632: Behaviour note rather than a defect, but it's an observable API change with no release note.

`new_proof_events` in `QueryEvents` (`rpcserver/rpcserver.go:8486`, via `QueryAssetStatsPerDay`) is fed by these same rows. Before this change it effectively counted *registration attempts* per day; after it counts *leaves newly acquired* per day. On a mature universe that's a step change to near-zero on an existing client's chart, and it'll read as "the node stopped receiving proofs".

I don't think anything is actually lost — `SYNC` events still carry the traffic volume and `LogSyncEvent`/`LogSyncEvents` are untouched by this PR, so anyone plotting activity has the right series available. And the new meaning matches the field name better than the old one did. But it's the kind of thing that generates a bug report from a dashboard owner, so it's worth a line in the release notes describing the semantic change rather than just the bug fix.

Related and mechanical: `scripts/check-release-notes.sh` greps `docs/release-notes` for `lightninglabs/taproot-assets/pull/2255` and finds nothing, and there's no `no-changelog` label, so `milestone-check` (`.github/workflows/main.yaml:247-257`) will fail as-is. #2251 and #2252 both have entries in `release-notes-0.9.0.md` on main — a file this branch doesn't have yet, which is the same staleness as finding 5.

tapdb/assets_store.go
```diff
@@ -2037,7 +2037,11 @@ func (a *AssetStore) importAssetFromProof(ctx context.Context,
 		return fmt.Errorf("unable to insert managed utxo: %w", err)
 	}
 
-	newAsset := proof.Asset
+	newAsset := proof.Asset.Copy()
+	err = restoreGroupWitness(newAsset, proof.Blob)
```
litbot-9000, line 2100: This whole commit (`bffe90a6 tapdb: restore group witness when importing proofs`) is already on `main` and can be dropped.

It's byte-identical to `c5df293b`, merged via #2222 — `git patch-id --stable` returns `94b8562c89e589b318edf4a48af12d385030898e` for both, and the only textual difference is the hunk offsets (`@@ -2037` here vs `@@ -2047` on main) and the blob index lines. So this is a rebase artifact, not a second divergent version of the same fix; no risk of regressing #2222.

`c5df293b` isn't an ancestor of this branch's merge-base (`3bca4074`), so the branch was cut before #2222 landed. It'll disappear on rebase, at which point this PR is the three commits it's actually about.

A rebase is needed regardless — merging `main` into this branch today conflicts in `tapdb/multiverse.go`, `tapdb/universe.go`, `universe/archive.go` and `universe/interface.go`, since main has taken #2202 and #2252 in the meantime. `#2202` in particular adds an append-only insertion journal on the universe insert path; worth a look during the rebase at whether it wants the same `isNew` signal you're introducing, so the two don't end up with independent answers to "was this leaf new". If it does, I'd say this PR should own the signal and #2202 consume it, since this is where it's defined.

tapdb/universe.go
```diff
@@ -915,14 +927,34 @@ func universeUpsertProofLeaf(ctx context.Context, dbTx BaseUniverseStore,
...
 	var (
 		leafInclusionProof *mssmt.Proof
 		universeRoot       mssmt.Node
+		leafInserted       bool
 	)
 
+	// Before we touch the tree, we check whether we already hold a leaf at
+	// this key. Callers use this to distinguish a genuinely new leaf from a
+	// re-registration of a leaf we already have, which is a no-op upsert.
+	_, err = dbTx.QueryUniverseLeafID(ctx, UniverseLeafIDQuery{
```
litbot-9000: Recording what I checked here, since this is where the review could most plausibly have gone wrong and it holds up.

I went looking for a determination that's right on SQLite and wrong on Postgres — `ON CONFLICT DO UPDATE` reporting a row affected when nothing changed, or the two dialects disagreeing on rows-affected for an upsert. That class of bug doesn't apply: this is a plain `SELECT` before the write, and `sql.ErrNoRows` means the same thing in both drivers. No `RETURNING`, no rows-affected, nothing dialect-specific. Given I can't run Postgres here (no container runtime), that's a good property to have chosen.

The key match is exact: `UpsertUniverseLeaf` at `:1033-1040` writes `LeafNodeNamespace: namespace`, `MintingPoint: mintingPointBytes`, `ScriptKeyBytes: scriptKeyBytes` — the same three variables this lookup reads, and the same three columns as `UNIQUE INDEX universe_leaves_unique_minting_script_namespace` (`migration 000031_ignore_tree.up.sql:80`). Three equality predicates on a three-column unique index is a full match whatever order they're written in, so the "single hit on an existing unique index" claim is structurally sound. It's one extra indexed lookup per item inside a loop that already calls `universeUpsertProofLeaf` per item, so it doesn't reintroduce an N+1 into a path that had been batched.

I also checked the mapping back out in `tapdb/multiverse.go:905-907`: `newItems = append(newItems, item)` appends the `*universe.Item` itself, so attribution is by identity and can't be scrambled by row ordering — the positional-misattribution bug I was hunting for isn't reachable. Same-leaf-twice-within-one-batch is right too, because the second iteration's lookup runs in the same transaction and sees the first insert; `TestMultiverseUpsertProofLeafBatchNewLeaves` covers that case explicitly, along with the reorg replacement. Nice test.

No change requested.

litbot-9000 (commented):
Solid, well-scoped fix that does what the issue asked for. I walked the consumer graph, the dialect question, and the batch index mapping, and I could not find a lost event or a misattributed one: the new/existing determination is a plain pre-lookup (not `RETURNING` or rows-affected), so it behaves identically on SQLite and Postgres, and the batch path maps results back by item identity rather than position. My findings are about coverage and about two things left on the table, not about the core change being wrong. Verdict: `COMMENT` — nothing here I'd block on.

### What consumes NEW_PROOF

Writers are `universe/archive.go:368` (`LogNewProofEvent`) and `:603` (`LogNewProofEvents`). Everything that reads `universe_events WHERE event_type = 'NEW_PROOF'`:

| Consumer | Path | Does suppression lose anything? |
|---|---|---|
| `universe_stats.total_asset_proofs` | view (migration `000029_ignore_burn_universe.up.sql:104-108`) → `QueryUniverseStats` → `AggregateStats.NumTotalProofs` → `universerpc` `num_total_proofs` | No — this is the bug. It's meant to count proofs held. |
| Per-asset `total_proofs` | same view → `QueryUniverseAssetStats` → `AssetSyncSnapshot.TotalProofs`, plus `SortByTotalProofs` ordering | No — same semantics, per root. |
| Per-day timeseries | `QueryAssetStatsPerDay{Sqlite,Postgres}` (`tapdb/sqlc/queries/universe.sql:334-355`) → `rpcserver/rpcserver.go:8486` `GroupedUniverseEvents.new_proof_events` | No information *lost* — but the field changes meaning. See finding 4. |
| `DeleteUniverseEvents` | `tapdb/universe.go:1489` | Cleanup on universe delete, not a reader. |

And the things I checked that are **not** backed by `universe_events`, so are unaffected:

- `tapdb/multiverse.go:944-950` — `transferProofDistributor.NotifySubscribers` iterates **all** `items`, not just `newItems`. The custodian still gets notified when a transfer proof is replaced. This is the one I most expected to break, and it doesn't.
- `universe/syncer.go:79-102` — `SimpleSyncer`'s `eventDistributor` is an in-memory `fn.EventDistributor`, unrelated to the events table.
- `universe/supplyverifier/manager.go:366` — `IssuanceSubscriptions`, separate mechanism.
- Federation sync/push — keyed on roots and leaf keys, never on the events table.
- `SYNC` events — `LogSyncEvent`/`LogSyncEvents` are untouched by this PR, so the "how much traffic did we see" signal is fully preserved and lives where it belongs.

So the answer to "is *not new* the same as *no event worth logging*" is **yes for every consumer that exists today**. A proof replacement is a real state change, but no reader of this table is asking about state changes; they're all asking "how many proofs do we hold". A separate event type would be the right answer only if something needed the change journal, and nothing does.

Worth noting the single-leaf path narrows this further: `universe/archive.go:315-319` early-returns when the block hash matches, so the only case where `leafKnown` actually suppresses an event is a reorg-driven proof replacement. Repeat syncs of an unchanged leaf never reached the log there even before this PR.

### Existing nodes' stale counts

Explicitly out of scope per the PR body ("this stops the growth, it doesn't undo it") and pre-agreed in #2253, which called retention "a separate question". No argument from me on deferring the *pruning*. But finding 3 below is about a different thing that I don't think has been considered: the count derivation itself, which is separable from the rows and would fix old and new nodes at once.

### Commit 1

`bffe90a6 tapdb: restore group witness when importing proofs` is **byte-identical** to `c5df293b`, already on `main` via #2222 — `git patch-id --stable` gives `94b8562c89e589b318edf4a48af12d385030898e` for both; the only textual difference is hunk offsets (`@@ -2037` vs `@@ -2047`) and blob hashes. `c5df293b` is not an ancestor of this branch's merge-base (`3bca4074`), so the branch was cut before #2222 landed. Harmless rebase artifact, no divergence, no risk to #2222 — it just evaporates on rebase. See finding 5.

### What I verified

- `make build` — **passes**, 2m18s (tapd + tapcli, tags `dev monitoring`).
- `go test -count=1 ./universe/... ./tapdb/...` at head `bee65406` — **all pass**, tapdb 47.2s.
- New tests by name, `-v`: `TestMultiverseUpsertProofLeafBatchNewLeaves` **PASS** (0.99s), `TestUniverseStatsProofCount` **PASS** (1.17s).
- `go test -count=1 -race -p 1 ./tapdb/... -run 'TestUniverseStats...|TestMultiverse...|TestUniverseIssuanceProofs'` — **ok, 142.8s, no races**. The full `-race ./tapdb/...` run exceeded my time budget and was not completed.
- `go test -count=1 -race ./universe/...` — flaked once in `universe/supplycommit` (`TestSupplyCommitFinalizeStateTransitions/finalize_with_asset_id_specifier`), passed on re-run. That package isn't touched by this PR; I'm not attributing it here.
- **Coverage probe:** reverted commit 3 (`83cba1f0`, `universe/archive.go` only) with `git revert --no-commit`, then `go build ./...` → RC 0 and the **full** `go test ./universe/... ./tapdb/...` → all pass in 50s. See finding 1.
- Index check: `migration 000031_ignore_tree.up.sql:80` defines `UNIQUE INDEX universe_leaves_unique_minting_script_namespace (minting_point, script_key_bytes, leaf_node_namespace)`. The new query has equality predicates on all three, so it's a full index match regardless of predicate order — your "single hit on an existing unique index" claim holds structurally.
- Merge state: `git merge origin/main` into this branch **conflicts** in `tapdb/multiverse.go`, `tapdb/universe.go`, `universe/archive.go`, `universe/interface.go` (main has since taken #2202 and #2252).
- `scripts/check-release-notes.sh` greps `docs/release-notes` for `pull/2255` and finds nothing; the PR carries no `no-changelog` label, so the `milestone-check` job will fail as-is. (#2251 and #2252 both have entries in `release-notes-0.9.0.md` on main, so the convention is live.)
- CI: `gh pr checks 2255` reports no checks; `commits/bee65406/status` is `{"state":"pending","total_count":0}`. Nothing had run at review time, so there was nothing to reconcile against.
- **I could not run Postgres.** There is no container runtime available to me (no docker/podman), so `test_db_postgres` / `ory/dockertest` cannot start. Everything above is SQLite-only. **I did not reproduce the 437ms vs 0.34ms measurement and am not vouching for it.** In CI, Postgres coverage for this code comes from the `unit dbbackend=postgres pkg=tapdb` matrix entry (`.github/workflows/main.yaml:270`), which will exercise both new tests — worth watching once CI runs, though as I argue in the inline note on `tapdb/universe.go`, the mechanism you chose is the one that makes a dialect split unlikely by construction.

Findings below, most severe first.

tapdb/universe_stats_test.go
GeorgeTsagk: Fixed in 295a5672. You were right that nothing pinned the archive itself.

The blocker was that the batch path verifies before it upserts, and `proof.RandProof` output doesn't verify (`invalid taproot proof: doesn't spend prev output`), so a mock multiverse alone doesn't get you there. `proof.NewMintingBlobs` does: it builds proofs that pass full verification and returns `*Proof` directly, so `mintTestAsset` mints real ones and the tests drive the actual ingest path.

Two batch tests plus a table-driven single-leaf one, asserting the exact id set handed to a recording `mockTelemetry`. I checked each fails for the regression it's meant to catch:

- `ids := fn.Map(items, ...)` instead of `newItems` → fails
- drop `newItems = append(newItems, newAnchorItems...)` → fails
- drop the `newNonAnchorItems` append → fails
- single-leaf logging unconditionally → fails

The anchor cases only fail because the batch now mixes group anchors and plain issuances; with the non-anchor-only batch I started with, dropping the anchor append passed. Good catch on naming that specific case.

tapdb/multiverse.go
GeorgeTsagk: Fixed in 855d5344. Agreed, and the TOCTOU argument is right: the read tx misses, verification runs, both writers proceed, serializable isolation makes one of them a no-op upsert, and both log.

`MultiverseStore.UpsertProofLeaf` now returns the same in-transaction answer the batch path uses and `leafKnown` is gone. It's set in the pending case too, since the leaf is committed by then. `FetchProofLeaf` stays where it is, it's still needed for the same-block-hash early return, it just no longer feeds the logging decision.

Costs a signature change through `MultiverseArchive` and ~40 call sites, nearly all `_`. Worth it for the reason you gave: two sibling paths answering the same question two ways is what drifts.

Covered by `TestUpsertProofLeafLogsOnlyNewLeaf`, whose second case is exactly this race: the multiverse reports the leaf as already held and no event is logged.

universe/archive.go
GeorgeTsagk: Added in f6d789a9, covering both the fix and the `new_proof_events` semantic change, since a dashboard owner seeing that series step down is the likely bug report. Noted that `sync_events` is untouched and still carries traffic volume.

The `milestone-check` failure was real, thanks. The branch has since been rebased onto main so `release-notes-0.9.0.md` exists here now.

tapdb/assets_store.go
GeorgeTsagk: Gone, the branch is rebased onto current main. Your patch-id check was right, it was a stale-base artifact: I cut the branch from a local checkout that was ~3 months behind, and #2222 had landed in the meantime.

On #2202: it wants the leaf's database id, not an `isNew` signal, so the two aren't answering the same question. I folded both into one `universeUpsertResult` rather than letting `universeUpsertProofLeaf` grow to five return values, so `rootStatus`, `leafInserted` and `leafID` now come back together. That's the one structural change I made to code merged recently, flagging it in case you'd rather it stayed positional.

jtobin (commented):
Only thing that Sol flagged that feels worth fixing, captured by this example sequence:

```
  1. The leaf is durably committed.
  2. The caller’s context expires while awaiting multiverse bookkeeping.
  3. The RPC reports failure before logging NEW_PROOF.
  4. A retry sees the leaf already exists and deliberately suppresses the event.
  5. Nothing later reconciles the missing event.
```

The effect mainly being that "num_total_proofs becomes permanently low, so the PR replaces systematic overcounting with a rarer undercounting mode."

Sol recommends recording the leaf and event atomically, which seems right. Not sure how much of a lift it is though?

tapdb/sqlc/queries/universe.sql
GeorgeTsagk: Done in 33d88dca, migration 66 redefines the view to count `universe_leaves`. You were right that it was an omission rather than a decision, and right that the derivation and the rows are separable: the event rows are untouched and `QueryAssetStatsPerDay` still reads them.

Two behaviours I had to handle that aren't obvious from the CTE:

**Supply sub-trees.** `burn_tree.go:134` and `supply_tree.go:472` both go through `universeUpsertProofLeaf`, so burn/ignore/mint_supply leaves land in `universe_leaves` with their own `universe_roots` rows. They never produced NEW_PROOF events, so today they contribute nothing. A plain leaf count picks them up:

```
supply tree only:  events=0  leaves=6  byType=map[burn:3 ignore:1 mint_supply:2]
```

The new `proof_counts` filters on `roots.proof_type IN ('issuance','transfer')` to keep that at zero.

**Non-monotonicity.** Deleting a leaf from a universe that still holds others leaves the event rows behind, so the count stayed flat; leaf-derived it drops:

```
2 leaves, same universe:  events=2 leaves=2  view=2
after deleting one:       events=2 leaves=1  view=1 (was 2)
```

That's the better answer but it is a change, so it's in the release notes.

`TestUniverseStatsEvents` also had to change: it asserted `NumTotalProofs: 0` for a harness that had already inserted 3 leaves. That assertion was the bug in miniature.

Coverage: `TestMigration66UniverseStatsFromLeaves` builds a v65 database with 2 leaves and 5 inflated events plus a burn sub-tree leaf, migrates, and asserts 5 → 2 with the burn leaf excluded, the event rows intact and the sync count untouched. It also migrates back down and asserts the event-derived count returns, so a rollback gets the view its release expects. Passes on SQLite and Postgres. `universe_federation`, `universe_sync`, `universe_sync_manual_insert`, `address_syncer` and `addresses` all still pass with their exact `AssertUniverseStats` figures unchanged, which is the evidence that the two derivations agree in normal operation.

jtobin (approved):
LGTM. :+1: :+1:

PR #2259, opened 2026-08-26
tapgarden: hardening phase four (boundaries)

jtobin:
This is the final chunk of the large-scale tapgarden hardening work originally contained in #2153 (following #2190, #2203 and #2247). It mostly handles the remainder of the reorganization described in the original PR:

> [..] stuff that moves various ['complecting'](https://www.youtube.com/watch?v=SxdOUGdseq4) machinery into other/new packages), and the rest is mostly in interface scaffolding [..]

It contains the following changes (summary ctsy Fable; apologies for the LLM verbiage, [I didn't have time to make it shorter](https://quoteinvestigator.com/2012/04/28/shorter-letter/)):

**Supply-commit participation behind an interface.** The planter and cultivator carried inline knowledge of universe commitments: delegation-key checks in seedling validation, pre-commit output scanning at funding time, and mint-event emission on confirmation.  That knowledge now lives behind `tapgarden.GenesisTxAugmenter` (`ValidateSeedling` / `BindData` / `OnBatchConfirmed`), implemented by `universe/supplycommit.GenesisAugmenter` and wired through `GardenKit`; a `NoOpAugmenter` covers configurations without supply-commit support, and rejects any seedling that requests it rather than silently minting with the feature disabled.  `NewGenesisAugmenter` validates its five dependencies at construction time, so a wiring mistake surfaces at startup rather than as silently absent supply-commit events.

**Typed pre-commit persistence.** tapdb's supply pre-commit row was read directly off `FundedMintAnchorPsbt.PreCommitmentOutput`, coupling the store to minting internals. The five `BatchStore` methods that bind a funded genesis transaction or seal a batch now take an `fn.Option[PreCommitBindData]` -- a typed payload the augmenter derives from the batch's current state. The same SQL writes happen in the same transactions; the store just no longer reaches into the minting machinery to find its inputs. The contract is fail-closed at both ends: a funded supply-commit batch must carry exactly one bind payload (and a non-supply-commit batch none) or the store refuses the write, and `BindData` errors instead of silently returning `None` for a malformed batch. Migration 66 pins the cardinality (one pre-commitment per batch), and the pre-commit upsert preserves a sealed group key rather than letting a later idempotent write -- e.g.  from a batch restored after a restart, which can no longer re-derive the key -- erase it with NULL.

**Universe publication behind an interface.** The cultivator's confirmation handler assembled `universe.Item` values inline and streamed them to the local universe; the planter's re-org handler had its own `UpsertProofLeaf` loop with the same wire-shape construction.  Both are downstream of tapgarden's natural end (a verifiable asset in the local store), so publication moves behind `tapgarden.MintProofPublisher`, implemented by the new `universe/mintpublish` package. tapgarden now hands over raw `(asset, proof, txHash, outIdx)` inputs and never constructs a universe leaf, key, or identifier itself.

**Planter/cultivator lifecycle hazards.** Closes several failure modes around batch creation, funding, and confirmation: a cancel request arriving after the cultivator's goroutine had exited (or stopped reading its request channel) deadlocked the planter until shutdown -- cultivators now expose a `Done()` channel and the completion signal is buffered, so late cancels surface an actionable error instead of hanging; a seedling or `FundBatch` request arriving while an earlier batch still holds the pre-broadcast singleton slot -- resumed at startup, or left orphaned on disk -- tripped the singleton index with a raw SQL error after burning key derivations and wallet funding work, and is now refused up front with a clear message; wallet inputs leased while funding a genesis transaction are released again when a later funding step fails, instead of staying locked until lease expiry; and the Frozen/Committed branches mutated the live batch before the corresponding DB writes, so a concurrent snapshot could observe memory ahead of disk -- mutations are now staged on a copy and applied only after the write commits.

**Confirmation is atomic with augmenter completion, and retried.** A failure in `OnBatchConfirmed` (the supply-commit mint-event write) aborts the confirmation attempt: `MarkBatchConfirmed` is the last persistence write in the branch, so the batch stays in Broadcast. The cultivator then retries the confirmation in place (and again on restart) rather than exiting and stranding the batch until an operator intervenes; once `MarkBatchConfirmed` succeeds, retries resume from the terminal state advance so the confirmation hooks are not repeated. Universe publication is idempotent and the event-dedup index from #2247 absorbs the augmenter's re-run, so the retry is safe. A failed delegation-key lookup during emission is likewise an error (previously the asset was silently dropped from the supply commitment). Two regression tests pin the ordering and the retry (augmenter failure and proof-watcher failure both leave the batch in Broadcast, then complete without a restart).

**Verifier generators to tapnode.** `GenGroupVerifier`, `GenGroupAnchorVerifier`, and `GenRawGroupAnchorVerifier` construct closures over `tapnode.GroupFetcher` that feed `proof.VerifierCtx`; they belong with `tapnode.GenHeaderVerifier` in the proof-verifier code, not inside the minting state machine. Moved along with their error values and LRU cache helpers.

**FundBatch returns VerboseBatch.** The `FundBatchResp` wrapper added nothing beyond the `*VerboseBatch` it held; dropped, so the type names what it is.


GeorgeTsagk (commented):
Replaced with inline comments below

tapgarden/planter.go
```diff
@@ -412,9 +400,14 @@ type ChainPlanter struct {
...
-		cultivators:       make(map[BatchKey]*Cultivator),
-		completionSignals: make(chan BatchKey),
+		cfg:         cfg,
+		cultivators: make(map[BatchKey]*Cultivator),
+		// Buffer size 1 so a cultivator that finishes while the
+		// gardener is inside a stateReq closure (e.g. servicing a
+		// cancel request) can hand off its completion signal and
+		// exit without blocking. The gardener drains it on the
+		// next select loop iteration.
+		completionSignals: make(chan BatchKey, 1),
```
GeorgeTsagk, line 411: This fixes the single-cultivator case, but there are multiple producers: every cultivator's `SignalCompletion` (blocking send) plus `watchResumedCultivator`. With a stale key already in the buffer, a cultivator completing while the gardener is inside `cancelMintingBatch` blocks in `SignalCompletion`, never reaches `close(b.done)`, and the gardener is stuck selecting on `respCh`/`Done()`, so nothing drains the buffer. Deadlock until shutdown, same class of hang this commit closes. Needs capacity >= outstanding signalers, or a non-blocking send.

tapgarden/cultivator.go
```diff
@@ -1134,30 +1242,86 @@ func (b *Cultivator) stateStep(currentState BatchState) (BatchState, error) {
...
+		// confirmation branch re-runs on restart, ensuring the
+		// correct updateMintingProofs callback is the one bound
+		// to the anchor tx. If we instead registered after
+		// MarkBatchConfirmed, a crash between the two would
+		// leave disk at Confirmed with no callback registered by
+		// us; the re-org watcher's Start-time recovery would
+		// re-register with its DefaultUpdateCallback, silently
+		// dropping the universe re-publish on any subsequent
+		// re-org.
+		if err := b.cfg.ProofWatcher.WatchProofs(
```
GeorgeTsagk: Since this now runs before `MarkBatchConfirmed` and the retry re-runs the whole branch, every retry appends another registration for the same txid in tapreorg's `waitForConf` (no dedup for an already-watched tx). A persistent `MarkBatchConfirmed` failure grows registrations unbounded at 1/s, and even one transient failure leaves a permanent duplicate; each registration fires the update callback on a reorg.

tapgarden/cultivator.go
```diff
@@ -60,6 +46,10 @@ const (
 	// DefaultTimeout is the default timeout we use for RPC and database
 	// operations.
 	DefaultTimeout = 30 * time.Second
+
+	// confirmationRetryInterval is the delay between attempts to finish a
+	// confirmed batch after a transient confirmation-side failure.
+	confirmationRetryInterval = time.Second
```
GeorgeTsagk, line 54: No backoff, and each attempt redoes proof construction, verification, import and publish for the whole batch. A deterministic failure spins that every second forever with only logs as a surface (batch isn't cancellable post-broadcast, `ErrChan` unused).

tapdb/sqlc/queries/assets.sql
```diff
@@ -1171,7 +1183,12 @@ VALUES (
 ON CONFLICT(batch_id, tx_output_index) DO UPDATE SET
     -- The following fields are updated if a conflict occurs.
     taproot_internal_key_id = EXCLUDED.taproot_internal_key_id,
-    group_key = EXCLUDED.group_key,
+    -- The group key is first known at seal time. A batch restored from disk
+    -- can no longer derive that key from its seedling row, so a later
+    -- idempotent write with NULL must not erase the value sealed earlier.
+    group_key = COALESCE(
```
GeorgeTsagk, line 1189: The conflict target above is still `(batch_id, tx_output_index)` while migration 66 makes `batch_id` alone unique. If the output index ever does mutate for a batch, this INSERT trips the batch_id index and hard-errors instead of updating. Unreachable today as far as I can tell, but either arbitrate on `(batch_id)` with `tx_output_index = EXCLUDED.tx_output_index`, or drop the "mutable binding data" framing.

tapdb/sqlc/queries/assets.sql
```diff
@@ -14,11 +14,16 @@ INSERT INTO asset_minting_batches (
...
+-- (creation_time_unix DESC, batch_id DESC), so any code that has to
+-- pick "the most recent" pre-broadcast batch when several share a
+-- timestamp reaches the same row as the migration.
 SELECT *
 FROM asset_minting_batches batches
 JOIN internal_keys keys
     ON batches.batch_id = keys.key_id
-WHERE batches.batch_state != $1;
+WHERE batches.batch_state != $1
+ORDER BY batches.creation_time_unix DESC, batches.batch_id DESC;
```
GeorgeTsagk, line 26: Observable query change (planter_test flipped `fn.Last` to `fn.First`) riding the VerboseBatch commit, whose message doesn't mention it. Same commit also drops the `Planter` interface, renames the `PendingAssetGroup` fields and deepens the copy in `applyFundingToBatch`. Worth documenting or splitting out.

tapdb/sqlc/queries/assets.sql
```diff
@@ -586,7 +598,7 @@ WITH target_batch AS (
 )
 UPDATE asset_minting_batches
 SET minting_tx_psbt = $2, change_output_index = $3, assets_output_index = $4,
-    genesis_id = $5, universe_commitments = $6
+    genesis_id = $5
```
GeorgeTsagk, line 601: Dropping the `universe_commitments` write is only explained here in the query comment; the commit message doesn't mention it.

universe/supplycommit/genesis_augmenter.go
```diff
@@ -0,0 +1,746 @@
...
+
+	allAssets := append(
+		make([]*asset.Asset, 0, len(anchorAssets)+len(nonAnchorAssets)),
+		anchorAssets...,
+	)
+	allAssets = append(allAssets, nonAnchorAssets...)
+
+	withDelegation := make([]*asset.Asset, 0, len(allAssets))
+	for _, m := range allAssets {
+		has, err := a.cfg.DelegationKeyChecker.HasDelegationKey(
```
GeorgeTsagk, line 520: This now runs for every confirming batch, so a transient store error here aborts and retries confirmation even for batches with no supply-commit involvement. Retry-safe, but wider than the commit message suggests.

tapgarden/planter.go
```diff
@@ -2149,6 +2038,44 @@ type fundingPrep struct {
 		error)
 }
 
+// releaseFundingInputs releases every wallet input leased while funding a
```
GeorgeTsagk, line 2068: This covers the funding-failure paths, but cancelling an already-funded batch still leaks the leases: `Cancel`, `cancelFailedBatch` and the broadcast-failure branch never unlock `GenesisPacket.LockedUTXOs`, so inputs of a cancelled genesis tx stay locked until expiry. Pre-existing, but same failure class; worth a follow-up.

tapgarden/mint_publisher.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// AnchorOutIdx is the output index of the asset anchor in the
+	// genesis transaction.
+	AnchorOutIdx uint32
+}
+
+// NoOpMintProofPublisher is a publisher that does nothing. It is
+// intended for tests and configurations that have no universe to ship
+// proofs to.
+type NoOpMintProofPublisher struct{}
```
GeorgeTsagk: Dead code, nothing wires it; tests leave the publisher nil and rely on the nil-guards in the cultivator/planter.

tapgarden/planter.go
```diff
@@ -1925,6 +1790,30 @@ func (c *ChainPlanter) cancelMintingBatch(ctx context.Context,
...
+			default:
+			}
+
+			// No cancel reply queued: the cultivator finished on
+			// its own before we could deliver the request. Drop
+			// it from the map so a future retry does not race
+			// against a stale entry and return a clear error to
+			// the caller.
+			delete(c.cultivators, batchKeySerialized)
+			return fmt.Errorf("batch %x already completed, cannot "+
```
GeorgeTsagk, line 1841: This also fires when a resumed cultivator failed pre-broadcast (`done` closes after the `BroadcastErrChan` send), where the batch is failed, not completed.

tapdb/migrations_test.go
```diff
@@ -1541,8 +1543,14 @@ func TestMigration65BackfillDedupesLegacyDuplicates(t *testing.T) {
...
-	// match the hash of the duplicated content.
+	// match the hash of the duplicated content. MAX(bytea) is not
+	// defined in Postgres, so fetch count and key separately.
 	var postCount int
+	require.NoError(t, db.QueryRowContext(ctx, `
+		SELECT COUNT(*) FROM supply_update_events
+	`).Scan(&postCount))
+	require.Equal(t, 1, postCount)
+
 	var survivingKey []byte
```
GeorgeTsagk, line 1554: Stale duplicate below: the second block re-runs the COUNT query instead of fetching `event_key`; the real fetch follows.

GeorgeTsagk (commented):
Redid my earlier comment as an inline review, findings anchored to the code. Two things with no line to hang them on:

1. Nit: the first commit's message names `CommitBatchTx` among the five store methods, which doesn't exist (four changed).
2. Checked migration 66 against the upgrade path: old code can't produce two pre-commit rows per batch (single keyed upsert, re-funding blocked, funding write atomic), so no brick risk. Ran the affected packages locally, all green.

universe/mintpublish/publisher.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// NewPublisher constructs a Publisher that ships items to reg. batchSize
+// controls the number of items per UpsertProofLeafBatch call and must be
+// non-zero; the type rules out negative values at the boundary. A zero
+// batchSize would make PublishMintBatch's outer loop advance by zero
+// and spin forever, so we refuse to construct such a Publisher.
+func NewPublisher(reg universe.BatchRegistrar, batchSize uint) *Publisher {
+	if batchSize == 0 {
+		panic("mintpublish: batchSize must be non-zero")
```
GeorgeTsagk, line 34: The new package ships with zero tests; neither the batching loop nor this panic is exercised.

tapdb/asset_minting.go
```diff
@@ -452,27 +472,64 @@ func insertMintAnchorTx(ctx context.Context, q PendingAssetStore,
...
+
+	switch {
+	case batch == nil:
+		return fmt.Errorf("minting batch is nil")
+
+	case batch.SupplyCommitments && preCommit.IsNone():
+		return fmt.Errorf("supply-commit batch %x has no pre-commit "+
+			"bind data", batch.BatchKeyBytes())
+
+	case !batch.SupplyCommitments && preCommit.IsSome():
```
GeorgeTsagk, line 504: This direction (non-supply-commit batch carrying a bind payload) is untested; only the "has no pre-commit bind data" branch has coverage.

tapgarden/augmenter.go
```diff
@@ -0,0 +1,145 @@
...
+// ValidateSeedling rejects any seedling that requests supply
+// commitments. Under NoOpAugmenter no augmenter substance is
+// available to emit the pre-commit anchor output or the
+// post-confirmation mint event, so silently accepting the seedling
+// would leave the caller believing supply commitments were on
+// while the batch proceeds without them. Non-supply-commit
+// seedlings pass through unchanged.
+func (NoOpAugmenter) ValidateSeedling(_ *MintingBatch, req Seedling) error {
+	if req.SupplyCommitments {
+		return fmt.Errorf("supply commitments requested but no " +
```
GeorgeTsagk, line 109: No test hits this rejection.

tapgarden/cultivator.go
```diff
@@ -334,27 +357,96 @@ func (b *Cultivator) advanceStateUntil(currentState,
...
+		}
+
+		log.Errorf("Cultivator(%x): unable to finalize confirmed "+
+			"batch: %v; retrying in %v", b.batchKey[:], err,
+			confirmationRetryInterval)
+
+		// A successful MarkBatchConfirmed leaves only the terminal
+		// state write to retry. Otherwise, replay the whole
+		// confirmation branch with the retained confirmation event.
+		if b.cfg.Batch.State() == BatchStateConfirmed {
```
GeorgeTsagk, line 420: Nothing pins this: both retry tests fail before MarkBatchConfirmed, so "hooks aren't repeated once it succeeded" is untested.

tapgarden/cultivator.go
```diff
@@ -654,10 +748,23 @@ func (b *Cultivator) stateStep(currentState BatchState) (BatchState, error) {
...
 		}
 
+		// Stage the freshly-computed commitment on a copy so
+		// derivations that need it (genesisScript, augmenter
+		// BindData) can read it without us mutating the live
+		// batch. Live-batch mutation is deferred until the
+		// AddSproutsToBatch DB write succeeds, so a concurrent
+		// Copy() from the planter never sees an in-memory batch
+		// that has advanced past what is on disk.
+		stagingBatch, err := b.cfg.Batch.Copy()
```
GeorgeTsagk, line 789: No test pins the staged-mutation invariant (concurrent Copy() not observing memory ahead of disk).

universe/supplycommit/genesis_augmenter_test.go
```diff
@@ -0,0 +1,331 @@
...
+	err error
+}
+
+func (c testDelegationChecker) HasDelegationKey(context.Context,
+	asset.ID) (bool, error) {
+
+	return c.has, c.err
+}
+
+type testMintEmitter struct{}
```
GeorgeTsagk, line 42: This mock doesn't count calls, so the positive filtering coverage from the deleted TestSupplyCommitDelegationKeyFiltering (owned assets emit, non-owned skipped) is lost; only the error path is covered now.

tapgarden/planter.go
```diff
@@ -1925,6 +1790,30 @@ func (c *ChainPlanter) cancelMintingBatch(ctx context.Context,
 
 			return cancelResp.err
 
+		case <-cultivator.Done():
```
GeorgeTsagk, line 1820: This path has no test, which is also how the completionSignals buffer issue slipped through; worth a test that races cancel against completion.

GeorgeTsagk (commented):
Test-coverage notes, inline below.

tapgarden/planter.go
jtobin: Fixed in updated changeset.

tapgarden/cultivator.go
jtobin: Fixed in updated changeset.

tapgarden/cultivator.go
jtobin: Fixed in updated changeset (now exponential backoff).

universe/supplycommit/genesis_augmenter.go
jtobin: Fixed in updated changeset.

tapgarden/planter.go
jtobin: On Fable's recommendation I'll PR a fix for this separately. Could have glued it in here but yeah, preexisting, so I figured a separate PR was reasonable.

tapdb/sqlc/queries/assets.sql
jtobin: Added an entry for this to the breaking changes section of the release notes.

tapdb/sqlc/queries/assets.sql
jtobin: Wontfix.

tapdb/sqlc/queries/assets.sql
jtobin: Wontfix.

tapgarden/planter.go
jtobin: Test added in the updated changeset.

tapgarden/cultivator.go
jtobin: Tests added in the updated changeset.

universe/mintpublish/publisher.go
jtobin: Tests added in the updated changeset.

tapgarden/mint_publisher.go
jtobin: Axed in updated changeset.

tapgarden/planter.go
jtobin: Wontfix ("completed" is I think fine for "done with" here).

tapdb/migrations_test.go
jtobin: Wontfix.

universe/supplycommit/genesis_augmenter_test.go
jtobin: Fixed in updated changeset.

tapdb/asset_minting.go
jtobin: Fixed, with a basic exhaustive test.

tapgarden/augmenter.go
jtobin: Fixed, with a basic exhaustive test.

tapgarden/cultivator.go
jtobin: Added a proper concurrency test for this one.

GeorgeTsagk (approved):
LGTM

PR #2262, opened 2026-08-28
tapgarden: release funding leases on batch cancellation

jtobin:
Addresses an item left.. uh, unaddressed in #2259. Simply ensures that we release leases immediately when cancelling a minting batch.

Fable's summary, from the commit message:

> Funding a minting batch leases the selected wallet inputs, and the leased outpoints are recorded only in memory on the batch's genesis packet (FundedPsbt.LockedUTXOs). Cancelling a batch wrote the cancelled state to disk but never called Wallet.UnlockInput, so the inputs stayed unusable until lnd's lease TTL expired.
> 
> Release the leases at every point a batch is cancelled into a dead state: the cultivator's Cancel branches (SeedlingCancelled and SproutCancelled), the planter's no-cultivator cancel of the pending batch, cancelFailedBatch on the pre-broadcast failure path, and the startup resume loop's cancellation of batches that fail to fund or seal. Committed batches retained for a broadcast retry keep their leases.
> 
> The release happens only after the cancelled state has been written, and unlock failures are logged rather than returned, so a lease cleanup problem can never un-cancel a batch.
> 
> Since LockedUTXOs is never persisted, a batch reloaded from disk carries an empty set even though its funded PSBT survives. In that case the outpoints are derived from the unsigned transaction's inputs; every input of a mint anchor was leased by the wallet during funding, so the two sets coincide.

### DIFF
--- a/tapgarden/cultivator.go
+++ b/tapgarden/cultivator.go
@@ -263,6 +263,14 @@ func (b *Cultivator) Cancel(respCh chan<- CancelResp) error {
 			)
 		}
 
+		// The batch is now cancelled on disk, so any wallet inputs
+		// leased when it was funded can be released.
+		if err == nil {
+			releaseBatchFundingInputs(
+				ctx, b.cfg.Wallet, b.cfg.Batch,
+			)
+		}
+
 		b.publishMintEvent(BatchStateSeedlingCancelled)
 
 		cancelResp = CancelResp{true, err}
@@ -279,6 +287,14 @@ func (b *Cultivator) Cancel(respCh chan<- CancelResp) error {
 
 			b.publishMintErrorEvent(
 				err, BatchStateSproutCancelled,
+			)
+		}
+
+		// The batch is now cancelled on disk, so any wallet inputs
+		// leased when it was funded can be released.
+		if err == nil {
+			releaseBatchFundingInputs(
+				ctx, b.cfg.Wallet, b.cfg.Batch,
 			)
 		}
 
--- a/tapgarden/planter.go
+++ b/tapgarden/planter.go
@@ -578,6 +578,15 @@ func (c *ChainPlanter) Start() error {
 					log.Warnf("Unable to cancel batch (%x)",
 						batchKey)
 				}
+
+				// The batch is now cancelled on disk, so any
+				// wallet inputs leased when it was funded can
+				// be released.
+				if err == nil {
+					releaseBatchFundingInputs(
+						ctx, c.cfg.Wallet, batch,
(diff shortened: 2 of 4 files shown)


GeorgeTsagk (commented):
Fix looks correct. I checked the new tests fail without the patch and the full tapgarden suite passes with it. Two notes inline.

tapgarden/cultivator.go
```diff
@@ -263,6 +263,14 @@ func (b *Cultivator) Cancel(respCh chan<- CancelResp) error {
 			)
 		}
 
+		// The batch is now cancelled on disk, so any wallet inputs
+		// leased when it was funded can be released.
+		if err == nil {
+			releaseBatchFundingInputs(
+				ctx, b.cfg.Wallet, b.cfg.Batch,
```
GeorgeTsagk, line 270: These two cultivator branches are the only release sites without a test. The new tests cover the pendingBatch cancel, cancelFailedBatch and the startup loop, but not a cancel delivered to an active cultivator at Frozen or Committed. advanceStateUntil reads CancelReqChan between steps, so a cancel queued while the mock wallet holds the sign request should land here deterministically. Worth a test asserting the unlock for at least one branch.

tapgarden/planter.go
```diff
@@ -578,6 +578,15 @@ func (c *ChainPlanter) Start() error {
 					log.Warnf("Unable to cancel batch (%x)",
 						batchKey)
 				}
+
+				// The batch is now cancelled on disk, so any
+				// wallet inputs leased when it was funded can
+				// be released.
+				if err == nil {
+					releaseBatchFundingInputs(
```
GeorgeTsagk, line 586: Optional: a crash between the state write and the release leaves a cancelled batch whose leases are still held, and the resume loop skips cancelled batches, so nothing releases them until the TTL expires. Releasing here for funded batches already in a cancelled state would close that window. Fine to leave given the TTL bound.

jtobin: Ok IIUC there's a slight gap that makes cancelling + release leases strictly non-atomic, so that a crash in between the two can leave a batch cancelled, but UTXOs still leased. The leases expire per lnd's TTL though, so this isn't a big deal.

Will leave this as wontfix for now.

tapgarden/cultivator.go
jtobin: Fixed in updated changeset.

GeorgeTsagk (approved):
LGTM, the cultivator cancel test closes the remaining gap.

PR #2264, opened 2026-08-31
tapdb: restrict pinned coin queries by anchor

### LINKED ISSUE #2257: tapfreighter: input selection stays unbounded for sends with pinned inputs
### Summary

#2252 bounded input selection so that funding a send lists coins in pages of
descending amounts and stops once the target is covered. Sends that pin their
inputs are excluded from that: they still list every eligible coin the node
holds. On a node with 2461 assets, coin selection for a pinned-input send takes
1.53s against 68ms for the bounded path.

### Where

The prev-ID filter is applied in Go, after the query has already built an
`AnchoredCommitment` for every eligible coin:

```go
// tapdb/assets_store.go, ListEligibleCoins
selectedCommitments, err := a.queryCommitments(ctx, assetFilter)
if err != nil {
        return nil, fmt.Errorf("unable to query commitments: %w", err)
}

// If we want to restrict on specific inputs, we do the filtering now.
if len(constraints.PrevIDs) > 0 {
        selectedCommitments = filterCommitmentsByPrevIDs(
                selectedCommitments, constraints.PrevIDs,
        )
```

So a bounded page could contain none of the pinned inputs, and #2252 rejects
the combination rather than returning a wrong answer:

```go
if constraints.CoinLimit > 0 && len(constraints.PrevIDs) > 0 {
        return nil, fmt.Errorf("coin limit cannot be combined with " +
                "specific prev IDs")
}
```

`CoinSelect.listCoins` correspondingly sets the page size to zero when
`PrevIDs` is set, which is what makes the listing unbounded.

Pinned inputs reach coin selection from `FundVirtualPsbt` and the other funding
RPCs, via `FundingDescriptor.PrevIDs`.

### Measurement

Same node and harness used for #2249 and #2250: a copy of a real regtest node,
sqlite, 2461 assets, dominant asset group holding 621 coins.

```
  operation                        before #2251/#2252    on current main
  SelectCoins, pinned prev IDs                16.66s              1.53s
  SelectCoins, 1-unit group send              16.57s               68ms
```

The pinned path is 10x better than it was, purely from the query batching in
#2251, but it is still linear in total holdings where the bounded path is not.

### Possible direction

Push the prev-ID set into the SQL filter, so the query returns only the pinned
coins instead of everything followed by a Go-side filter. Each prev ID is an
(anchor outpoint, asset ID, script key) triple, all of which the query already
joins on.

A keyset cursor over `(amount, assets.asset_id)` in place of LIMIT/OFFSET would
also let pinned listings be paged, and would additionally remove the unbounded
retry #2252 needs when a multi-page listing falls short. See the discussion on
#2252 for that variant.

### Environment

- taproot-assets at `020de0a1` (main, with #2251 and #2252 merged), sqlite
  backend, regtest
- Node state: 2461 unspent assets, 143 MB database, grown over roughly 600
  sends

GeorgeTsagk:
Fixes #2257.

Pinned coin selection currently materializes every eligible coin before filtering requested PrevIDs. This change maps the requested PrevIDs to anchor outpoints and restricts QueryAssets to those anchors. Exact asset ID and script key matching remains in Go. Pagination with PrevIDs remains prohibited.

The regression test creates 64 assets and retains a proof only for the pinned anchor. The old path fails while materializing an unrelated asset. The bounded query succeeds for the pinned input, and rejects an altered script key.

Tests:
* go build ./...
* make lint-source
* go test ./...
* custom channel immediate close, force close, breach, and coop checks

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -399,6 +399,12 @@
   `universe.mbox-cleanup-check-timeout` to configure periodic cleanup of
   auth mailbox messages whose claimed outpoints have been spent on chain.
 
+## Performance Improvements
+
+- [PR#2264](https://github.com/lightninglabs/taproot-assets/pull/2264)
+  bounds pinned input coin selection to the requested anchor UTXOs instead of
+  materializing all eligible wallet assets.
+
 # Tooling and Documentation
 
 - [PR#1962](https://github.com/lightninglabs/taproot-assets/pull/1962)
@@ -411,4 +417,3 @@
 - [PR#2056](https://github.com/lightninglabs/taproot-assets/pull/2056)
   expands the example portfolio pilot with constraint enforcement,
   configurable fill caps, and live CoinGecko pricing.
-
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -399,12 +399,6 @@
   `universe.mbox-cleanup-check-timeout` to configure periodic cleanup of
   auth mailbox messages whose claimed outpoints have been spent on chain.
 
-## Performance Improvements
-
-- [PR#2264](https://github.com/lightninglabs/taproot-assets/pull/2264)
-  bounds pinned input coin selection to the requested anchor UTXOs instead of
-  materializing all eligible wallet assets.
-
 # Tooling and Documentation
 
 - [PR#1962](https://github.com/lightninglabs/taproot-assets/pull/1962)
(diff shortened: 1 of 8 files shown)


docs/release-notes/release-notes-0.8.0.md
```diff
@@ -399,6 +399,12 @@
   `universe.mbox-cleanup-check-timeout` to configure periodic cleanup of
   auth mailbox messages whose claimed outpoints have been spent on chain.
 
+## Performance Improvements
+
+- [PR#2264](https://github.com/lightninglabs/taproot-assets/pull/2264)
```
jtobin, line 404: Let's move this one to the v0.9.0 notes.

jtobin (changes requested):
LGTM, only the release note location needs to be changed.

litbot-9000:

jtobin (approved):
LGTM. :+1: :+1:

PR #2265, opened 2026-08-31
tapdb: query orphan UTXO candidates in one pass

### LINKED ISSUE #2256: tapdb: orphan UTXO sweep scans every managed UTXO on every send
### Summary

Every funded send scans every managed UTXO to decide whether any of them holds
only zero-value assets. The scan issues one query per UTXO regardless of how
many orphans exist, so it costs the same on a node with zero orphans as on one
with twenty. On a node with 2461 assets this is 6.3s per send, which is now the
largest single cost in a send.

It runs unless `--disable-sweep-orphan-utxos` is set, so it is on by default.

### Where

`FetchOrphanUTXOs` fetches all managed UTXOs, then calls `queryChainAssets`
once per UTXO to look at the assets anchored there:

```go
// tapdb/assets_store.go
utxos, err := q.FetchManagedUTXOs(ctx)
if err != nil {
        return fmt.Errorf("failed to fetch managed utxos: %w", err)
}

for _, u := range utxos {
        ...
        // Query all assets anchored at this outpoint.
        _, assetsAtAnchor, err := a.queryChainAssets(
                ctx, q, QueryAssetFilters{
                        AnchorPoint: u.Outpoint,
                        Now:         sqlTime(now),
                },
        )
```

`MaxOrphanUTXOs = 20` bounds how many orphans are returned, but the loop only
breaks once that many have been *found*, so a node with no orphans walks the
whole UTXO set.

The caller is `CoinSelect.SelectOrphanCoins`, reached from
`AssetWallet.FundPacket` and `FundBurn`:

```go
// tapfreighter/wallet.go
if !f.cfg.DisableSweepOrphanUtxos {
        zeroValueInputs, err = f.cfg.CoinSelector.SelectOrphanCoins(ctx)
```

### Measurement

Same node and harness used for #2249 and #2250: a copy of a real regtest node,
sqlite, 2461 assets, 0 orphan UTXOs. Median of 3 runs.

```
  FetchOrphanUTXOs   6.91s before #2251/#2252, 6.28s after
```

The batching in #2251 barely helps here, because the cost is the number of
queries, not the work inside each one.

For scale, on the same node after both PRs:

```
  SelectCoins (1-unit group send)    68ms
  FetchAllAssets (2461 assets)      236ms
  FetchOrphanUTXOs                  6.28s
```

End to end, a v2 address send on that node takes about 10.8s, so this scan is
roughly 6 of those seconds. The rest is proof transfer, signing and anchoring.

### Possible direction

The predicate is "this UTXO anchors no assets with a non-zero amount, and every
asset on it is a tombstone or a burn". That is expressible as a single query
over `assets` grouped by `anchor_utxo_id`, rather than materialising every
asset (with its witnesses) at every UTXO in turn. The existing per-UTXO
filtering for leases, swept UTXOs and missing signing info can stay in Go on
the much smaller result.

Related history: the sweeping itself came from #1832, the result cap and the
signing-info filter from #1859 / #1905.

### Environment

- taproot-assets at `020de0a1` (main, with #2251 and #2252 merged), sqlite
  backend, regtest
- Node state: 2461 unspent assets, 0 orphan UTXOs, 143 MB database, grown over
  roughly 600 sends

GeorgeTsagk:
Fixes #2256.

The orphan scan currently materializes assets once per managed UTXO. This change adds a grouped SQL candidate query that filters normal funded anchors in one pass. Unknown script key types keep the existing full validation fallback. Lease, sweep, signing, chain transaction, and output checks are unchanged.

The regression test creates 64 funded anchors and proves they are excluded before asset materialization. It fails on the old path after an unrelated funded anchor is made unmaterializable and passes with the candidate query.

Tests:
* go build ./...
* make lint-source
* go test ./...
* custom channel immediate close, force close, breach, and coop checks

### DIFF
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -399,12 +399,6 @@
   `universe.mbox-cleanup-check-timeout` to configure periodic cleanup of
   auth mailbox messages whose claimed outpoints have been spent on chain.
 
-## Performance Improvements
-
-- [PR#2265](https://github.com/lightninglabs/taproot-assets/pull/2265)
-  reduces orphan UTXO scan work by filtering normal funded anchors in a single
-  database query instead of materializing assets once per managed UTXO.
-
 # Tooling and Documentation
 
 - [PR#1962](https://github.com/lightninglabs/taproot-assets/pull/1962)
--- a/docs/release-notes/release-notes-0.8.0.md
+++ b/docs/release-notes/release-notes-0.8.0.md
@@ -399,6 +399,12 @@
   `universe.mbox-cleanup-check-timeout` to configure periodic cleanup of
   auth mailbox messages whose claimed outpoints have been spent on chain.
 
+## Performance Improvements
+
+- [PR#2265](https://github.com/lightninglabs/taproot-assets/pull/2265)
+  reduces orphan UTXO scan work by filtering normal funded anchors in a single
+  database query instead of materializing assets once per managed UTXO.
+
 # Tooling and Documentation
 
 - [PR#1962](https://github.com/lightninglabs/taproot-assets/pull/1962)
--- a/tapdb/sqlc/queries/assets.sql
+++ b/tapdb/sqlc/queries/assets.sql
@@ -685,6 +685,32 @@ FROM managed_utxos utxos
 JOIN internal_keys keys
     ON utxos.internal_key_id = keys.key_id;
 
+-- name: FetchOrphanManagedUTXOs :many
+SELECT
+    utxos.*,
+    keys.*,
+    MAX(CASE
+        WHEN COALESCE(script_keys.key_type, 0) = @unknown_key_type
+        THEN 1 ELSE 0
+    END) = 1 AS needs_asset_validation
+FROM managed_utxos utxos
+JOIN internal_keys keys
+    ON utxos.internal_key_id = keys.key_id
+JOIN assets
+    ON assets.anchor_utxo_id = utxos.utxo_id
+JOIN script_keys
+    ON assets.script_key_id = script_keys.script_key_id
+GROUP BY utxos.utxo_id, keys.key_id
+HAVING SUM(CASE
+    WHEN COALESCE(script_keys.key_type, 0) = @unknown_key_type
+        OR script_keys.key_type = @burn_key_type
+        OR (
+            script_keys.key_type = @tombstone_key_type
+            AND assets.amount = 0
+        )
+    THEN 0 ELSE 1
+END) = 0;
+
 -- name: AnchorPendingAssets :exec
 WITH assets_to_update AS (
     SELECT script_key_id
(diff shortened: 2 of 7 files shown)


docs/release-notes/release-notes-0.8.0.md
```diff
@@ -399,6 +399,12 @@
   `universe.mbox-cleanup-check-timeout` to configure periodic cleanup of
   auth mailbox messages whose claimed outpoints have been spent on chain.
 
+## Performance Improvements
+
+- [PR#2265](https://github.com/lightninglabs/taproot-assets/pull/2265)
```
jtobin, line 404: Let's move this to the v0.9.0 notes.

tapdb/sqlc/queries/assets.sql
```diff
@@ -678,6 +678,32 @@ FROM managed_utxos utxos
 JOIN internal_keys keys
     ON utxos.internal_key_id = keys.key_id;
 
+-- name: FetchOrphanManagedUTXOs :many
+SELECT
+    utxos.*,
+    keys.*,
+    MAX(CASE
+        WHEN COALESCE(script_keys.key_type, 0) = @unknown_key_type
```
jtobin, line 686: Sol flagged that this should probably also include `@burn_key_type`. I think it's right; we do have the IsBurn() check in assets_store.go, and this would bypass it as-is (it also seems right to continue to require the IsBurn() check to verify the burn).

tapdb/sqlc/queries/assets.sql
```diff
@@ -678,6 +678,32 @@ FROM managed_utxos utxos
...
+    ON assets.anchor_utxo_id = utxos.utxo_id
+JOIN script_keys
+    ON assets.script_key_id = script_keys.script_key_id
+GROUP BY utxos.utxo_id, keys.key_id
+HAVING SUM(CASE
+    WHEN COALESCE(script_keys.key_type, 0) = @unknown_key_type
+        OR script_keys.key_type = @burn_key_type
+        OR (
+            script_keys.key_type = @tombstone_key_type
+            AND assets.amount = 0
```
jtobin, line 702: Sol points out that IsTombstone() actually checks that the amount is zero _and_ the actual script key equals NUMSCompressedKey. We should check:

```
script_keys.tweaked_script_key = @nums_key
AND assets.amount = 0
```

jtobin (changes requested):
Sol characterised the invariant we're looking for here as: "sweeping decisions rely on the asset’s defining properties, not only its classification label," which seems right to me. It elaborated with:

> For tombstones, SQL can verify those properties cheaply: zero amount plus the actual NUMS script key. It doesn’t need full Go reconstruction.
> 
> For burns, validation depends on witness data, so the Go IsBurn() check remains necessary. This gives strong validation without discarding the optimization.

Check the two inline comments re: that stuff (comparison with NUMS key in SQL, ensuring IsBurn() runs in Go); the only other thing I spotted is that the release note entry is misplaced.

litbot-9000:

jtobin (approved):
LGTM. :+1: :+1:

PR #2266, opened 2026-09-01
tapreorg: watcher core (1/4)

### LINKED ISSUE #1888: [bug]: After Taproot Assets channel force-closure, funds are not swept.
### Background

I was running `litd 0.15.2` and had a Taproot Assets channel on mainnet. The channel was force-closed by the remote party in block 922532. 

Before an upgrade, I looked at the node and the channel was shown as available, but inactive. I didn't find this surprising as the remote node was offline at the time. I then upgraded to `litd 0.16.0`, and as I started up I noticed the channel was now showing as a pending force close.

The remote party has since come online and swept their funds in block 925343, but my node has not yet swept the funds.

The force close transaction can be found here: https://mempool.space/tx/a40ccc09954c73e616951c4d9fc356a4ef824b404caa8e31842e8a7ea567b49e

Logs from that first startup:

[2025-11-26_lnd tap fc.log](https://github.com/user-attachments/files/23812044/2025-11-26_lnd.tap.fc.log)

Logs from today:

[2025-11-27_lnd.log](https://github.com/user-attachments/files/23812072/2025-11-27_lnd.log)

`lncli pendingchannels`:

```
{
    "total_limbo_balance": "72669",
    "pending_open_channels": [],
    "pending_closing_channels": [],
    "pending_force_closing_channels": [
        {
            "channel": {
                "remote_node_pub": "02f8423194a5385f38c9bfa93a8b38a23106a6003a3fc99580845f669b086c9620",
                "channel_point": "949ef02119c92a666a15cf20637b9f0317627863f3ef85fc944eaaa06b2882bd:0",
                "capacity": "100000",
                "local_balance": "72339",
                "remote_balance": "26485",
                "local_chan_reserve_sat": "0",
                "remote_chan_reserve_sat": "0",
                "initiator": "INITIATOR_REMOTE",
                "commitment_type": "SIMPLE_TAPROOT_OVERLAY",
                "num_forwarding_packages": "0",
                "chan_status_flags": "",
                "private": true,
                "memo": "",
                "custom_channel_data": {
                    "funding_assets": [
                        {
                            "version": 1,
                            "asset_genesis": {
                                "genesis_point": "0514f59b8ef83031a8aa43a8b318c68e8745321c4941adbc8d0f14ff7547a67f:1",
                                "name": "bepsi",
                                "meta_hash": "2093f8c78219fbe9e8d2f1600cc805fba899603cb9b011ed15859e9dfb44e48f",
                                "asset_id": "92c4f0764cd5befb141cd5845eb81028a80b824e6e223c7b4e97d874e7d2f121"
                            },
                            "amount": 10000000,
                            "script_key": "0250aaeb166f4234650d84a2d8a130987aeaf6950206e0905401ee74ff3f8d18e6",
                            "decimal_display": 3
                        }
                    ],
                    "local_assets": [
                        {
                            "asset_id": "92c4f0764cd5befb141cd5845eb81028a80b824e6e223c7b4e97d874e7d2f121",
                            "amount": 27087
                        }
                    ],
                    "remote_assets": [
                        {
                            "asset_id": "92c4f0764cd5befb141cd5845eb81028a80b824e6e223c7b4e97d874e7d2f121",
                            "amount": 9972913
                        }
                    ],
                    "outgoing_htlcs": [],
                    "incoming_htlcs": [],
                    "capacity": 10000000,
                    "local_balance": 27087,
                    "remote_balance": 9972913,
                    "outgoing_htlc_balance": 0,
                    "incoming_htlc_balance": 0
                }
            },
            "closing_txid": "a40ccc09954c73e616951c4d9fc356a4ef824b404caa8e31842e8a7ea567b49e",
            "limbo_balance": "72669",
            "maturity_height": 0,
            "blocks_til_maturity": 0,
            "recovered_balance": "0",
            "pending_htlcs": [],
            "anchor": "LIMBO"
        }
    ],
    "waiting_close_channels": []
}
```

### Your environment

`0.7.0-alpha commit=v0.7.0`
`Bitcoin Core daemon version v29.0.0`

```sh
tapcli getinfo                       # version of `tapd`, `lnd`, and network
uname -mrsv                          # operating system 
bitcoind --version || btcd --version # version of `btcd`, `bitcoind`, or other backend
```

### Steps to reproduce

Tell us how to reproduce this issue. Please provide stacktraces and links to code in question.

### Expected behavior

Tell us what should happen

### Actual behavior

Tell us what happens instead

### LINKED ISSUE #2160: tapfreighter: select pending xfers can become stranded

### LINKED ISSUE #2179: [bug]: LogAnchorTxConfirm writes irreversibly at 1-conf
The chain porter's confirmation-success path
(`tapfreighter/chain_porter.go` → `LogAnchorTxConfirm`) performs
several irreversible DB writes on a single-confirmation observation:

- `SetAssetSpent` on every transfer input
- Output asset rows created
- `chain_txns.block_hash` set
- (Added by PR #2159) `SupersedeConflictingTransfers` on every other
  unconfirmed transfer that shares an input

The conf notification is registered with `numConfs = 1` in
`waitForConfEventOnce`. The `ReOrgWatcher` handles the
**same-tx-different-block** reorg case (it patches the proof's
`BlockHeight`/`BlockHeader`/`TxMerkleProof` when the same txid
re-confirms), but **does not handle the case where a different
transaction replaces our anchor entirely**. The watcher waits
indefinitely for the original txid to re-confirm; meanwhile the DB
side effects from the orphaned 1-conf observation stay in place.

## Concrete failure scenario

Most relevant for **force-close sweep parcels**, where lnd's sweeper
publishes several RBF candidates with **different txids**:

1. Sweep candidates **T** and **T'** both attempt to confirm the
   same logical sweep (same inputs, different fee/witness).
2. **T** confirms first at 1-conf.
3. **T**'s parcel runs `LogAnchorTxConfirm`:
   - **T** marked confirmed.
   - Its inputs marked `spent=true`.
   - Its outputs materialized as asset rows.
   - **T'** (and any other live candidates) marked
     `superseded=true` via `SupersedeConflictingTransfers`.
4. A reorg drops the block carrying **T**; **T'** confirms in the
   reorganised chain instead.
5. The dominant chain has **T'** as the actual confirmed anchor.
   The DB still has:
   - **T** recorded as confirmed (at an orphaned txid).
   - **T'** flagged `superseded=true`, so its eventual
     confirmation never triggers `LogAnchorTxConfirm` either.
6. Net: the DB points at an orphaned anchor; the chain points at a
   different one we now treat as dead.

For non-sweep transfers, a different-tx reorg requires either a
fee-bumped replacement or an attacker constructing a conflicting
spend of one of our inputs. For sweep parcels it requires only that
the sweeper's chosen candidate reorgs out and a *different*
candidate it had already broadcast wins — concrete, not theoretical.

## Likely downstream impact

- Proof verification: assets reference the recorded anchor; if
  that's the orphaned txid, verification against the dominant chain
  fails.
- Stuck "live" candidate: the actually-confirmed **T'** is flagged
  `superseded` and its parcel is never resumed, so the porter
  doesn't surface the (chain-true) confirmation event.
- Coin selection: inputs are marked spent at an orphaned tx; the
  chain says they were spent at a different one. Coin selection
  result is the same (input is unavailable), but the record is
  internally inconsistent.

## Two structural fixes

**Option A — defer everything until SafeDepth.** Register the conf
ntfn in `waitForConfEventOnce` with `numConfs = SafeDepth` instead
of `1`. Drop the `ReOrgWatcher`'s role for this path entirely.

- Pros: clean; no inverse logic needed.
- Cons: changes the latency profile of *every* confirmed transfer
  (delay = SafeDepth blocks before the on-disk state catches up to
  the chain). Affects RPC consumers that read transfer state.

**Option B — add inverse operations to the `ReOrgWatcher` for the
different-tx case.** When the watched anchor txid doesn't
re-confirm within some bound, treat the orphaned anchor as dead and
roll back `LogAnchorTxConfirm`'s side effects: un-spend inputs,
un-supersede conflicts, remove output asset rows, clear the chain
tx's block hash.

- Pros: preserves current latency.
- Cons: needs an "un-confirm" code path that doesn't exist; touches
  schema implications around asset row removal; fiddly enough to
  warrant its own audit.

## Related

- PR #2163 — touches the analogous issue on the supersession path (foreign confirmed spend triggering supersession at 1-conf).
- PR #2159 — extended the 1-conf `LogAnchorTxConfirm` path with `SupersedeConflictingTransfers`. Inherits (does not introduce) the reorg-irreversibility shape.

jtobin:
(Partially resolves #1888, #2160, and #2179.)

This PR introduces the core of the "watcher," a service for reconciling speculative local state with actual chain history. I expect it to dramatically improve both 1) interoperation with lnd's sweeper, and thus force close handling, in particular, and 2) general re-org handling, across the codebase.

The sort of TLDR of the design is that we gain a single service that's responsible for observing and coordinating transaction replacements caused by a reorg, RBF transaction, and so on. Each individual component (tapgarden, tapfreighter, etc. -- each called a *site* in watcher parlance) defines how to handle replacements in its own domain, but the watcher is the thing that:

1) observes and records confirmed chain evidence,
2) determines what the observed outcome means for the operation being watched, and
3) drives the owning subsystem towards the corresponding state.

The previous handling of this stuff, depending on the component, was either *ad hoc*, partial, or, in some rare cases, missing entirely. The new watcher handles, per Sol, anything in which:

> - The winning form witnesses and eventually buries.
> - A superseded form sees the winner as foreign/conflicting.
> - The superseded anchoring is abandoned and its speculative state is compensated.
> - The winner’s asset materialization remains exactly once.
> 
> That applies even when the interesting event is not what one would colloquially call a reorg. It also covers:
> 
> - RBF and fee-bump replacements.
> - Sweep rebatching.
> - Counterparty or otherwise competing spends.
> - Attempts that are broadcast but never confirm.
> - Restart and notification-stream interruption.
> - Cleaning up speculative state after abandonment.
> - A confirmed attempt being displaced and another attempt subsequently winning.
> 
> A reorg is merely the most demanding case because the system has to reverse a formerly witnessed outcome and accept a new one. The broader abstraction is: “Which candidate transaction, if any, currently realizes this anchoring claim, and how final is that fact?”

The core of the watcher is added in this PR, but isn't actually hooked up yet. I have four PR's planned: the second will add the per-site handlers, the third will decommission the legacy watcher, and the fourth will add a bunch of misc. stuff that didn't fit anywhere else.

(IIRC the crucial central changes involve about ~4k LoC. There are also about ~2k LoC involved in stitching everything together, and the rest is in tests and autogenerated stuff. I've also verified core invariants with a bunch of TLA+, not included here.)


GeorgeTsagk (commented):
Core looks solid, the model and the tests are careful. Built and ran the unit tests on sqlite and postgres locally, all green. Comments inline. 1 and 2 are the ones I'd want fixed before sites land on this.

tapreorg/service.go
```diff
@@ -0,0 +1,2190 @@
...
+
+	if heightHint == 0 {
+		heightHint = s.anchoring.CreatedHeight
+	}
+
+	id := s.anchoring.ID
+
+	reOrgChan := make(chan struct{}, 1)
+	confEvent, errChan, err := w.cfg.Notifier.RegisterConfirmationsNtfn(
+		s.ctx, &txid, tx.TxOut[0].PkScript, 1, heightHint, true,
```
GeorgeTsagk: lnd runs `txscript.ParsePkScript` on this in `NewConfRequest`, which only accepts P2PKH, P2SH, P2WPKH, P2WSH and P2TR. A spender whose output 0 is OP_RETURN or anything else non-standard fails registration, and for a foreign spend that is the counterparty's choice. What happens then: `handleSpendDetail` resenses, the sweep rebuilds the sensor every ScanInterval, historical dispatch redelivers the same spend, registration fails again. The candidate is never recorded. If that tx is the actual spender the anchoring stays Unwitnessed forever and the site never sees Conflicted or Abandoned.

Same at lines 1011 and 1071.

Pick the first output whose script parses. If none does, log once and leave the candidate pending rather than resensing, so it does not turn into a per-minute rebuild.

tapreorg/service.go
```diff
@@ -0,0 +1,2190 @@
...
+			return fmt.Errorf("anchoring %d: %w", anchoring.ID,
+				err)
+		}
+	}
+
+	// Known candidates re-subscribe so their re-confirmations and
+	// re-orgs keep flowing after a restart.
+	for _, candidate := range anchoring.Spends {
+		err := w.watchCandidate(s, candidate.W.TxHash(),
+			candidate.W.Tx(), candidate.W.Height())
```
GeorgeTsagk: Re-subscribing a known candidate with its last recorded height as the hint is unsafe. lnd starts the historical rescan at max(caller hint, cached hint). If a reorg moved the tx to a lower height while we were down, the rescan starts above it and, without txindex (neutrino, bitcoind without -txindex), never finds it. The verify loop below then flips the candidate off-chain, the spend sub redelivers it as a known txid, `watchCandidate` dedups on `candidateSubs`, and the location is never refreshed. Anchoring sits Unwitnessed with the tx confirmed.

Use the trigger set's lowest HeightHint (or CreatedHeight) here. Same concern for `detail.SpendingHeight` at line 1236 when the spend detail is stale, though that one is narrower.

tapreorg/anchoring.go
```diff
@@ -0,0 +1,377 @@
...
+	// finality.
+	Threshold uint32
+}
+
+// Validate checks the spec's value-level invariants.
+func (s *RegistrationSpec) Validate() error {
+	if s.Site == "" {
+		return errors.New("registration requires a site ID")
+	}
+	if s.Triggers.Len() == 0 {
```
GeorgeTsagk, line 323: Nothing validates the trigger pkScripts. lnd rejects a spend registration whose script is not standard, so an empty or malformed script from a site bug passes Register, then adopt fails on every sweep with a warn log and the anchoring sits live, never stuck, never surfaced. Run `txscript.ParsePkScript` on each trigger here so it fails at the call site.

tapreorg/service.go
```diff
@@ -0,0 +1,2190 @@
...
+	case witness != nil:
+		return w.cfg.Registry.StageForeclosure(
+			ctx, edge.Child, edge.Parent, ForeclosureEvent{
+				Parent:  edge.Parent,
+				W:       *witness,
+				OnChain: true,
+			},
+		)
+
+	// The parent is not witnessed at all: any staged foreclosing
```
GeorgeTsagk, line 1809: An Abandoned parent falls into this branch and flips the foreclosure that `Deliver` just staged on-chain back to off-chain on the child's next adoption. Derivation does not read `OnChain` on foreclosures so nothing breaks, but the edge now says the abandoning tx is off-chain and `foreclosureStronger` ranks on that field. Handle Abandoned here the way `settleDependentEdges` does: stage the cause witness on-chain.

tapreorg/anchoring.go
```diff
@@ -0,0 +1,377 @@
...
+// can never coexist on a single chain. The watcher enforces the rule
+// structurally — a transaction spending only part of the set is
+// foreign by construction and never reaches the site's predicate —
+// so a set must be consumed atomically or not at all; outpoints that
+// can legitimately be realized in separate transactions belong to
+// separate anchorings.
+//
+// The zero value is invalid; values in circulation originate from
+// NewTriggerSet and are therefore well-formed by provenance.
+type TriggerSet struct {
```
GeorgeTsagk, line 121: For PR 2: lnd's sweeper can split a batch's inputs across separate txs. A multi-outpoint anchoring then goes Conflicted and Abandoned even though every outpoint was swept as intended. Worth stating in this doc that a trigger set must be atomic on chain, so sites default to one outpoint per anchoring unless they control the spending tx.

tapdb/sqlc/migrations/000068_reorg_registry.down.sql
```diff
@@ -0,0 +1,18 @@
...
+DROP TABLE IF EXISTS reorg_dependencies;
+
+DROP INDEX IF EXISTS reorg_candidate_spends_uk;
+DROP TABLE IF EXISTS reorg_candidate_spends;
+
+DROP INDEX IF EXISTS reorg_trigger_outpoints_op_idx;
+DROP INDEX IF EXISTS reorg_trigger_outpoints_uk;
+DROP TABLE IF EXISTS reorg_trigger_outpoints;
+
+DROP INDEX IF EXISTS reorg_anchorings_witness_txid_idx;
```
GeorgeTsagk: Drops `reorg_anchorings_witness_txid_idx`, which the up migration never creates, and misses `reorg_anchorings_stuck_idx`. DROP TABLE takes the indexes with it so it works, just stale.

rpcserver/rpcserver.go
```diff
@@ -1831,6 +1832,96 @@ func (r *RPCServer) ListTransfers(ctx context.Context,
...
+	// paged rather than offered unbounded.
+	maxAnchoringPageSize = 1000
+)
+
+// ListAnchorings lists one page of the re-org watcher's speculative
+// anchorings: everything the daemon has staked on chain outcomes, the
+// phase the chain has assigned to each stake, and the delivery state
+// of the owning subsystem. Filters and page bounds are applied by the
+// registry query itself.
+func (r *RPCServer) ListAnchorings(ctx context.Context,
```
GeorgeTsagk, line 1851: The rejection paths (negative limit or offset, limit above 1000, unknown phase name) have no test.

jtobin:
Note that I've added a bunch (~4k LoC) of TLA+ covering this section of the work. From the PR:

> Adds TLA+ models for the watcher’s core lifecycle, foreclosure cascade, durable effects, and key sweeper interactions.
> 
> The models explore reorgs, unreliable notifications, transaction replacement, and overlapping claims, verifying internal consistency within the watcher’s stated finality assumptions.
> 
> They also document important integration boundaries, particularly registration granularity and the treatment of independently realizable outputs.

EDIT:

Since ripped these out; I'll maintain them on my fork.

tapreorg/service.go
jtobin: Fixed in updated changeset.

tapreorg/service.go
jtobin: Fixed in updated changeset.

tapreorg/anchoring.go
jtobin: Fixed in updated changeset.

tapreorg/service.go
jtobin: Fixed in updated changeset.

tapreorg/anchoring.go
jtobin: Fixed in updated changeset.

tapdb/sqlc/migrations/000068_reorg_registry.down.sql
jtobin: Fixed in updated changeset (now mirrors the up migration exactly).

rpcserver/rpcserver.go
jtobin: Fixed in updated changeset.

tapreorg/service.go
```diff
@@ -0,0 +1,2268 @@
...
+	script, err := watchScript(tx)
+	if err != nil {
+		log.Warnf("Anchoring %d: cannot watch foreclosure %v: %v",
+			s.anchoring.ID, txid, err)
+		return nil
+	}
+
+	confEvent, errChan, err := w.cfg.Notifier.RegisterConfirmationsNtfn(
+		s.ctx, &txid, script, s.anchoring.Threshold,
+		foreclosing.Height(), true, nil,
```
GeorgeTsagk: Same hint problem as the candidate subs, one level up. This hints from the foreclosing witness's recorded height and `foreclosureSubs` dedups on txid, so once a sub is opened with a stale hint a later restage with the corrected location never re-subscribes. Restart after a reorg that moved the parent's witness lower, child adopted before the parent re-derives: the child subscribes at the old height, the rescan misses the tx without txindex, and the cascade never certifies until a stream break rebuilds the sensor. Hint from the parent's `earliestSpendHint`, or at least the child's CreatedHeight. Non-blocking since no site can reach the cascade yet, but cheap to fix here.

tapreorg/service.go
```diff
@@ -0,0 +1,2268 @@
...
+		)
+
+	// The parent is not witnessed at all: any staged foreclosing
+	// evidence is itself off-chain.
+	default:
+		staged := edge.Foreclosure.UnwrapToPtr()
+		if staged == nil || !staged.OnChain {
+			return nil
+		}
+		staged.OnChain = false
```
GeorgeTsagk, line 1816: No test reaches this branch, nor the Buried arm above on the restage side. Given the Abandoned arm had a bug here last round, worth a scripted case for each.

tapdb/reorg_registry.go
```diff
@@ -0,0 +1,1465 @@
...
+		}
+	}
+
+	return view, nil
+}
+
+// foreclosureStronger reports whether a should replace b as the
+// view's staged foreclosure: certified beats uncertified, then
+// on-chain beats off-chain, then deeper wins.
+func foreclosureStronger(a, b tapreorg.ForeclosureEvent) bool {
```
GeorgeTsagk, line 1299: Zero coverage. No test builds a child with two parents, so the strongest-foreclosure selection is unexercised.

GeorgeTsagk (approved):
LGTM

tapreorg/service.go
jtobin: Fixed in update changeset.

tapreorg/service.go
jtobin: Fixed in update changeset.

tapdb/reorg_registry.go
jtobin: Fixed in update changeset.

PR #2267, opened 2026-09-01
rpcserver: reject negative invoice expiry

### LINKED ISSUE #2261: [bug]: creating an invoice with a negative expiry defaults to 86400 seconds
### Background

When creating an invoice using `v1/taproot-assets/channels/invoice` using a negative `expiry` value in seconds, I get an invoice with an expiry of 86400 seconds. 

### Your environment

```sh
tapd v0.8.0
```

### Steps to reproduce

Make an invoice using this body:
```
{
	"asset_amount": "1000000",
	"group_key": "...",
	"invoice_request": {"expiry": -1}
}
```

### Expected behavior

I would expect the RPC to throw an error, since it isn't possible to make an invoice with a negative expiry.

Vandit1604:
### Fixes #2261

`AddInvoice` only defaulted an invoice expiry of exactly zero:

```go
expirySeconds := iReq.Expiry
if expirySeconds == 0 {
    expirySeconds = int64(rfq.DefaultInvoiceExpiry.Seconds())
}
```

A negative `expiry` (e.g. `-1`) skips that check, so the raw negative value is passed through to lnd's `AddInvoice`, which treats a non-positive expiry as unset and stamps its own 86400s default. That is why the reporter saw a 24h invoice rather than an error (`rfq.DefaultInvoiceExpiry` is only 300s, so the 86400 comes from lnd).

This rejects a negative expiry at the RPC boundary, before any RFQ work, so the caller gets a clear error instead of a silently mangled invoice. A zero expiry still means "use the default" and positive values are unchanged.

### Testing

Added a unit test for the new `validateInvoiceExpiry` helper covering the negative, zero and positive cases. Verified it fails without the guard and passes with it.

### DIFF
--- a/rpcserver/invoiceexpiry_test.go
+++ b/rpcserver/invoiceexpiry_test.go
@@ -1,0 +1,23 @@
+package rpcserver
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestValidateInvoiceExpiry checks that a negative invoice expiry is rejected
+// while a zero (default) or positive expiry is accepted.
+func TestValidateInvoiceExpiry(t *testing.T) {
+	t.Parallel()
+
+	// A negative expiry is meaningless and must be rejected.
+	err := validateInvoiceExpiry(-1)
+	require.ErrorContains(t, err, "must not be negative")
+
+	// A zero expiry is allowed; it is later replaced with the default.
+	require.NoError(t, validateInvoiceExpiry(0))
+
+	// A positive expiry is used as-is.
+	require.NoError(t, validateInvoiceExpiry(3600))
+}
--- a/docs/release-notes/release-notes-0.9.0.md
+++ b/docs/release-notes/release-notes-0.9.0.md
@@ -113,6 +113,13 @@
   long-running node the series steps down. The per-day `sync_events`
   series is unchanged and still reflects sync traffic volume.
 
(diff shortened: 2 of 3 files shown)


jtobin (approved):
LGTM. :+1: :+1:

rpcserver/invoiceexpiry_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestValidateInvoiceExpiry checks that a negative invoice expiry is rejected
+// while a zero (default) or positive expiry is accepted.
+func TestValidateInvoiceExpiry(t *testing.T) {
```
GeorgeTsagk, line 11: This only pins the helper, it still passes if the call in AddInvoice is removed. The guard runs before any config use, so the wiring can be asserted with a zero value server:

```go
_, err := (&RPCServer{}).AddInvoice(ctx, &tchrpc.AddInvoiceRequest{
	InvoiceRequest: &lnrpc.Invoice{Expiry: -1},
})
require.ErrorContains(t, err, "must not be negative")
```

Not blocking.

GeorgeTsagk (approved):
LGTM. Checked lnd's side, a negative expiry falls into the non-AMP default branch in invoicesrpc, so the diagnosis is right. The guard also covers the hodl invoice path since that one forwards iReq.Expiry raw.

PR #2269, opened 2026-09-02
rfq+rfqmsg: harden asset amount handling

GeorgeTsagk:
## Description

Hardens asset amount handling in the RFQ layer. Balance list sums are range-checked at decode and aggregation time, the interceptor response computation uses big integer arithmetic, and channel commitment balance updates are bounds-checked. 

Covered by unit tests in `rfqmsg`, `rfq`, and `tapchannel`.


jtobin (approved):
LGTM. :+1: :+1:

PR #2270, opened 2026-09-02
authmailbox: bound the MultiSubscription message queue

GeorgeTsagk:
## Description

Bounds the shared MultiSubscription message queue, which previously grew without limit. The queue is now a fixed-capacity channel that applies backpressure to the subscription streams instead of dropping deliveries, so memory stays bounded without losing messages.

Covered by unit and property tests in `authmailbox`.


jtobin (approved):
LGTM. :+1: :+1:

PR #2276, opened 2026-09-03
scripts: reduce verify-install quorum to three

### LINKED ISSUE #2241: proposal: reduce manifest verification quorum from five sigs to three
We require five signatures on release manifests in order for scripts/verify-install.sh to pass. I think this is excessive, as it usually takes awhile to get the requisite signatures, and sometimes we fail to get enough at all.

To the extent that this script remains desirable at all (I don't really have a good idea about who actually uses it/wants to use it), I think three is plenty, and is a much easier number to realize. So, IMO we should reduce the requisite signatures in the verify-install script accordingly.

jtobin:
Resolves #2241.

Previously we required five signatures, but I believe this to be overkill, as it often proves difficult to corral enough people in a short time frame. Three signatures are much easier to get, and provide what IMO is a perfectly satisfactory degree of assurance.


GeorgeTsagk (approved):
LGTM

darioAnongba (approved):

PR #2277, opened 2026-09-03
backup: keep a live encrypted asset wallet backup file on disk

GeorgeTsagk:
This adds an lnd `channel.backup` style file for the asset wallet. tapd now
keeps `<datadir>/<network>/assets.backup` in sync with the wallet state
instead of relying on operators to run the export RPC by hand.

### What the file is

- One compact (v2) entry per confirmed, unspent asset leaf, keyed by asset
  ID, script key and anchor outpoint.
- Encrypted with the same key lnd uses for `channel.backup`. Any tapd
  connected to an lnd restored from the same seed can read it.
- Written on startup, on every wallet change and on shutdown. Writes go
  through temp file, fsync, rename, so a copy taken at any time is complete.

### How it stays up to date

- `backup.Updater` holds the entries in memory and subscribes to the asset
  store proof notifications and the planter mint events.
- On every change it reconciles against the database after a 1s debounce:
  new and re-anchored leaves get a fresh entry, spent leaves are dropped,
  entries only found on disk are kept.
- Failed writes are retried. A file at the path that cannot be decrypted
  with the wallet key is a startup error, since it most likely belongs to
  another seed.

### Restore and compatibility

- `ImportAssetsFromBackup` accepts the encrypted file directly and detects
  the format from the content. `tapcli assets backup import` works
  unchanged.
- Plaintext exports of all existing versions import exactly as before. No
  proto change.
- New flags: `--backup.filepath` and `--backup.disable`.
- Fixes found while testing restores:
  - The import skipped assets as already present when a proof file for
    them was still on disk. A node whose database was wiped but whose
    proofs directory survived recovered nothing. The existence check now
    consults the wallet database only.
  - Backups did not record the script key type and the import guessed it
    from the tweak, so unique Pedersen keys (V2 address receives) came back
    as external script path keys, hidden from `ListAssets`, `ListBalances`
    and coin selection. Entries now carry the type as an optional TLV, and
    older backups are classified by re-deriving BIP-86 and Pedersen keys.
  - A full value send without change or tombstone output imports no local
    proof, so the spent leaf stayed in the file. The chain porter send
    events are now a trigger next to the proof notifications and the mint
    events.
- After self review: channel funding leaves are excluded from the file,
  disk-only entries the database knows as spent are pruned, type-less
  tombstone and burn entries from older backups are classified correctly, a
  decoded type outside the known range falls back to classification, an
  optimistic (v3) export at the backup path is refused, a symlinked path is
  followed, and `backup.filepath` naming a directory is a config error.

### Testing

- Unit tests for the encrypted container, the file swapper and the updater:
  add, remove, reorg, debounce, disk only retention, retries after disk and
  DB failures, startup failures, sync and stop.
- New itest `backup_file_updates`: startup, mint, send, receive, wrong key
  rejection, a grouped asset received on a V2 address, restore of a fresh
  tapd on the same lnd, restore after wiping only the database, both
  restored leaves visible with default filters and spent back.
- Manual runs on a docker regtest network covering the custom path, the
  disable flag, the foreign file at path case, and a full seed recovery
  (second lnd restored from the same mnemonic, fresh tapd, import, spend).

### Docs

- `docs/backup-file.md`: operator guide (what it covers, how to keep a copy,
  how to restore).
- `docs/backup.md`: mechanics and container format.
- `docs/safety.md`: now points at the file.

### Not in this PR

Writing the key family markers from #2225 into the file. Once that lands
the updater needs one `ListAccounts` call per write. The two format changes
are independent.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -187,7 +187,7 @@ func createAssetBackup(ctx context.Context,
 // will not be populated (only the public keys will be included).
 func collectAssetBackups(ctx context.Context,
 	assets []*asset.ChainAsset,
-	proofArchive proof.Archiver,
+	proofArchive proof.Exporter,
 	keyLookup KeyLocatorLookup) ([]*AssetBackup, error) {
 
 	backups := make([]*AssetBackup, 0, len(assets))
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/wire/v2"
+	"github.com/lightninglabs/taproot-assets/address"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/proof"
 	"github.com/lightningnetwork/lnd/keychain"
@@ -100,6 +101,14 @@ type ScriptKeyBackup struct {
 	// Tweak is the tweak applied to derive the final script key.
 	// If nil, a BIP-0086 tweak is assumed.
 	Tweak []byte
+
+	// Type is the script key type as known by the exporting wallet. It
+	// decides how the wallet treats the key after a restore, for example
+	// whether the asset shows up in default listings and coin selection.
+	// A tweaked key alone does not reveal the type: unique Pedersen keys
+	// carry a tweak too but are spent like BIP-0086 keys. Unknown if the
+	// backup predates this field.
+	Type asset.ScriptKeyType
 }
 
 // KeyDescriptorBackup contains the key derivation info for an internal key.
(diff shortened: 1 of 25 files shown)


jtobin (approved):
LGTM. :+1: :+1:

I ran this through reviews w/both Opus (multi-agent) and Sol. They each reported a bunch of stuff, but all practical concerns evaporated when I dug into points individually.

(EDIT: there's obviously some issue, presumably minor, keeping the postgres itests job from passing.)

jtobin (commented):
Came back to this one and noticed it had grown a bit. A quick pass w/Sol pointed out the following that seem worth looking into:

> - [High] Key-lookup failures can create permanently incomplete backups. backup/backup.go:181 discards errors from FetchInternalKeyLocator and emits a zero locator. The updater then considers the entry successfully built and only rebuilds it when its block hash changes (backup/updater.go:565). Restoration persists that incorrect locator (backup/import.go:530), potentially leaving the restored anchor unspendable. Key-lookup failures should keep the entry dirty and be retried.
> 
> - [High] A failed final backup during shutdown is reported as success.  backup/updater.go:327 logs any Sync failure but always returns nil.  Consequently, the server’s error handling at server.go:924 is unreachable. Disk-full, permission, timeout, or proof failures can therefore leave the backup stale while a graceful shutdown appears successful.

(One nit as well: the commit series could probably be compressed somewhat without losing much information.)

GeorgeTsagk:
Both points addressed, thanks.

1. Key lookup: an unknown anchor key (not ours, e.g. a channel funding key) is recorded with the public key only and no longer registered on import. Any other lookup failure now fails the entry, so the updater leaves the leaf out, reports it and retries, instead of writing a zero locator.
2. Stop returns the final sync error while still shutting the loop down, so `server.Stop` surfaces it.

Also squashed the series down to 8 commits and rebased on main.

PR #2177, opened 2026-06-25
Immediate force-close itest

GeorgeTsagk:
## Description

Adds an itest edge-case where a tap channel is force closed immediately without any commitment updates.

Depends on https://github.com/lightningnetwork/lnd/pull/10804 (merged) and is based on #2278, which gates the negotiated config derivation behind a feature bit so the backwards compat test keeps passing.

PR #2278, opened 2026-09-03
tapchannel: derive initial commitment from negotiated configs behind feature bit

GeorgeTsagk:
## Description

lightningnetwork/lnd#10804 hands the aux funding controller the negotiated
local and remote channel configs when the initial commitment is derived.
That makes the height zero state consistent with what is re-derived on an
immediate force close, but it also changes the commitment we build at
funding time. Peers on older versions still derive it from zeroed configs,
so funding a channel between a node on lnd master and a v0.8.0 node fails
with `counterparty's commitment signature is invalid` (the `Compat Tests`
job on #2177 shows this).

This PR bumps lnd to a master commit that includes #10804 and gates the new
derivation behind an aux channel feature bit:

- `negotiated-chan-cfg` (bits 4/5) is added to `tapfeatures` and advertised
  as optional.
- If the peer signals it, the configs are used as handed by lnd.
- If the peer doesn't, both configs are zeroed before the initial
  commitment is derived. That is exactly what older lnd passed at that
  point, so both sides arrive at the same commitment transaction.
- If we advertise the feature as required and the peer lacks it, the
  funding flow fails with `ErrNegotiatedChanCfgRequired`. This path is
  inactive until the bit is flipped.

Phase 2 (next release) is a follow-up that moves the bit to required. At
that point peers without it are rejected at init by `checkRequiredBits`,
with the funding check as a second line of defence.

Manually verified with the bit flipped to required: a v0.8.0 peer is
dropped at connect time with `peer does not support required feature:
negotiated-chan-cfg`, new to new channels keep working.

Needed by #2177 (its compat job goes green once this lands).

## Testing

- Unit tests for the decision helper across all local/peer bit
  combinations, plus a guard that the bit is advertised as optional.
- `TestBackwardsCompatChannels` (v0.8.0 vs tip) passes again and takes the
  zeroed config path.
- `TestCustomChannels/force_close` passes without touching the fallback.

### DIFF
--- a/docs/release-notes/release-notes-0.9.0.md
+++ b/docs/release-notes/release-notes-0.9.0.md
@@ -132,6 +132,17 @@
 
 ## Functional Updates
 
+- [PR#2278](https://github.com/lightninglabs/taproot-assets/pull/2278)
+  derives the initial (height zero) commitment of an asset channel from
+  the negotiated local and remote channel configs (dust limit, CSV
+  delay), which makes it consistent with the commitment that is
+  re-derived on an immediate force close. A new aux channel feature bit,
+  `negotiated-chan-cfg`, is advertised as optional. When a peer doesn't
+  signal it, the previous derivation from zeroed configs is used so that
+  channels can still be opened with older versions. The next release
+  will make the feature required, at which point peers that don't signal
+  it are rejected.
+
 - [PR#2202](https://github.com/lightninglabs/taproot-assets/pull/2202)
   adds cursor-based delta sync to the universe federation. Each server
   exposes its insertion-ordered leaf journal via the new `SyncDelta`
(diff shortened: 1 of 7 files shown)


docs/release-notes/release-notes-0.9.0.md
```diff
@@ -132,6 +132,17 @@
...
 ## Functional Updates
 
+- [PR#2278](https://github.com/lightninglabs/taproot-assets/pull/2278)
+  derives the initial (height zero) commitment of an asset channel from
+  the negotiated local and remote channel configs (dust limit, CSV
+  delay), which makes it consistent with the commitment that is
+  re-derived on an immediate force close. A new aux channel feature bit,
+  `negotiated-chan-cfg`, is advertised as optional. When a peer doesn't
+  signal it, the previous derivation from zeroed configs is used so that
+  channels can still be opened with older versions. The next release
```
jtobin, line 142: :+1: I made a v0.10 milestone; we should create an issue re: making this feature required, and then lump it under that, so as not to forget.

jtobin (approved):
LGTM. :+1: :+1:

docs/release-notes/release-notes-0.9.0.md
GeorgeTsagk: Opened #2288 as a draft for the flip, on the v0.10 milestone.

PR #2283, opened 2026-09-07
tapchannel: return empty resolution when no assets to sweep

### LINKED ISSUE #2281: [bug]: tapchannel: force-close resolution fails with "no inputs provided" when the local commitment holds zero assets
### Background

When we force-close an asset channel in which all assets sit on the remote side (our `LocalAssets` list is empty), `AuxSweeper.resolveContract` for `TaprootLocalCommitSpend` fails with `error distributing coins: no inputs provided`. lnd's chain watcher treats this as a hard error (`unable to handle localclose … unable to aux resolve`), never dispatches the close, and the channel stays in `pending_force_closing` indefinitely.

The resolution is retried on every block. Each attempt re-imports the commitment proof and re-fetches funding input proofs from the universe (~30s per channel), so with two affected channels the chain watcher can exceed lnd's 60s blockbeat budget and log `ChainArbitrator: process block timeout` on that block. Cosmetic compared to the main problem, but it is how the failure shows up at ERR level on every block.

Root cause in `tapchannel/aux_sweeper.go` (v0.8.3, identical on `master` as of 2026-09-04):

- `resolveContract`, `case input.TaprootLocalCommitSpend:` sets `assetOutputs = commitState.LocalAssets.Val.Outputs` (`aux_sweeper.go:2135`) with no length check.
- `createSweepVpackets` (`aux_sweeper.go:203-299`) is then called with `sweepInputs == nil`. It still appends one `CommitAllocationToLocal` with `Amount: 0` (`:271-281`), builds an empty `inputProofs` slice, and calls `tapsend.DistributeCoins(inputProofs, allocs, …)` (`:295`), which returns `no inputs provided`.

### Your environment

```sh
tapcli getinfo   # tapd 0.8.3-alpha (commit v0.8.3), lnd 0.21.3-beta, network mainnet, integrated mode
uname -mrsv      # Linux x86_64
```
* neutrino chain backend
* Both affected channels: capacity ~100k sats, remote party is the funding initiator, `custom_channel_data.local_assets` empty, all assets in `remote_assets`
* Peer was offline at close time, so `closechannel` fell back to a force close

### Steps to reproduce

1. Open an asset channel and move the entire asset balance to the remote side, so `custom_channel_data.local_assets` is empty and only BTC remains locally.
2. With the peer offline, force-close the channel (`lncli closechannel --force`).
3. Wait for the commitment tx to confirm.
4. Observe the logs below repeating on every block. `lncli pendingchannels` keeps the channel under `pending_force_closing_channels` and never advances.

Logs (UTC). Identifiers redacted: `CHAN_A` is the channel point, `COMMIT_TX_A` the confirmed local commitment txid.

Resolution attempt on a block:

```
13:08:20.523 [INF] CNCT: ChannelPoint(CHAN_A): spend confirmed at height 965476, transitioning to Confirmed
13:08:20.549 [INF] CNCT: Local unilateral close of ChannelPoint(CHAN_A) detected
13:08:20.550 [INF] TCHN: Generating resolution_blob for contract_type=TaprootLocalCommitSpend, chan_point=CHAN_A
13:08:20.581 [INF] TCHN: First time seeing commit_txid=COMMIT_TX_A, importing
13:08:20.626 [INF] TCHN: Importing 1 proofs for ChannelPoint(CHAN_A)
13:08:47.093 [INF] TCHN: Fetching funding input proof 1 of 1, locator=(proof.Locator) {
   … ~12 PROF "Starting proof transfer backoff procedure … transfer_type=receive" lines …
13:08:47.233 [INF] TCHN: All 1 input proofs fetched, importing locator=(proof.Locator) {
13:08:47.515 [INF] TCHN: Sweeping 0 asset outputs (second_level=false): ([]*tapchannelmsg.AssetOutput) <nil>
13:08:47.515 [INF] TCHN: Creating sweep packets for 0 inputs
13:08:47.516 [INF] TCHN: Created 1 allocations for commit tx sweep: ([]*tapsend.Allocation) (len=1 cap=1) {
13:08:47.516 [ERR] CNCT: Failed to handle confirmed spend: handle known local state: unable to handle localclose for chan_point=CHAN_A: unable to aux resolve: error distributing coins: no inputs provided
```

After a daemon restart the arbitrator comes back in `StateCommitmentBroadcasted`, detects the confirmed spend again, and repeats the same cycle:

```
13:14:28.127 [INF] CNCT: ChannelArbitrator(CHAN_A): starting state=StateCommitmentBroadcasted, trigger=chainTrigger, triggerHeight=965480
13:14:28.161 [INF] CNCT: ChannelPoint(CHAN_A): spend confirmed at height 965476, transitioning to Confirmed
13:14:28.176 [INF] CNCT: Local unilateral close of ChannelPoint(CHAN_A) detected
13:15:28.625 [INF] TCHN: Generating resolution_blob for contract_type=TaprootLocalCommitSpend, chan_point=CHAN_A
13:15:28.635 [INF] TCHN: First time seeing commit_txid=COMMIT_TX_A, importing
   … same fetch/import/fail cycle …
```

For completeness, the anchor sweeps for the same close also fail because 330 sats minus fee leaves no output. Probably expected:

```
12:34:29.681 [ERR] SWPR: Initial broadcast failed: create RBF-compliant tx: create sweep tx: tx has no output
```

On-chain the commitment tx has four outputs: two 330 sat anchors, the zero-asset to-local (BTC only), and the remote's to-remote carrying the full asset balance. tapd pushed the commitment proof for the remote's output correctly (`… outpoint=COMMIT_TX_A:3 …` in the PROF send locator), so proof handling is fine; only our own resolution path is broken.

### Expected behavior

An empty `LocalAssets` set means there is nothing for tapd to sweep. `resolveContract` should return an empty resolution, the same way it does when `req.CommitBlob.IsNone()`, so lnd proceeds to sweep the plain-BTC to-local output and the channel resolves normally.

Suggested fix: in `resolveContract` (or at the top of `createSweepVpackets`), if `len(assetOutputs) == 0` for the commit-output cases (`TaprootLocalCommitSpend`, `TaprootRemoteCommitSpend`, `TaprootCommitmentRevoke`), skip allocation and `DistributeCoins` and return an empty blob.

### Actual behavior

`createSweepVpackets` builds a zero-amount allocation with no input proofs, `tapsend.DistributeCoins` returns `no inputs provided`, lnd aborts close handling, and the channel stays in `pending_force_closing` indefinitely. The failing resolution is re-run on every block.

jtobin:
Resolves #2281. Fable's summary, from the commit message:

> When lnd asks the aux sweeper to resolve a commitment output whose side of the channel holds no assets, as when the whole asset balance sits with the remote party, resolveContract selected an empty asset output list and still went on to build sweep packets from it.  createSweepVpackets then handed an empty input set to DistributeCoins, which failed with "no inputs provided". lnd treats an aux resolution error as fatal for the close, so the channel stayed in pending_force_closing and the failing resolution, including its funding proof fetch, was re-run on every block.
>
> Return an empty resolution as soon as the selected output list is empty, the same way we do when there is no commit blob. The output carries only BTC in that case, and lnd sweeps it as a plain output.

### DIFF
--- a/tapchannel/aux_sweeper.go
+++ b/tapchannel/aux_sweeper.go
@@ -2277,6 +2277,22 @@ func (a *AuxSweeper) resolveContract(
 			"skipping", req.CommitTx.TxHash())
 	}
 
+	// If the selected side holds no assets, the output being resolved is
+	// BTC only and there is nothing for us to sweep. The commitment
+	// transaction itself is still imported above, so its transfer is
+	// recorded and its proofs delivered as for any other close. We return
+	// an empty resolution, as we do without a commit blob, so lnd sweeps
+	// the output as a plain one. Building sweep packets from an empty
+	// input set would fail instead, and lnd treats that as a hard error
+	// that stalls the whole close.
+	if len(assetOutputs) == 0 {
+		log.Infof("No asset outputs to sweep for contract_type=%v, "+
+			"chan_point=%v, returning empty resolution", req.Type,
+			req.ChanPoint)
+
+		return lfn.Err[tlv.Blob](nil)
+	}
+
 	if req.CommitTx == nil {
 		return lfn.Errf[returnType]("no commitment transaction "+
 			"found for chan_point=%v", req.ChanPoint)
--- a/docs/release-notes/release-notes-0.9.0.md
+++ b/docs/release-notes/release-notes-0.9.0.md
@@ -127,6 +127,16 @@
   86400 second default instead of returning an error. Fixes
   [#2261](https://github.com/lightninglabs/taproot-assets/issues/2261).
 
+* [PR#2283](https://github.com/lightninglabs/taproot-assets/pull/2283)
+  fixes force-close resolution of an asset channel whose resolved
+  commitment output holds no assets, as when the whole asset balance
+  sits with the remote party. The aux sweeper tried to build sweep
+  packets from an empty input set, failed with `no inputs provided`,
+  and lnd treated that as a hard error, leaving the channel in
+  `pending_force_closing` and re-running the failing resolution on
+  every block. The sweeper now returns an empty resolution for such
+  outputs, so lnd sweeps them as plain BTC outputs.
+
 # New Features
 
 ## Functional Enhancements
--- a/tapchannel/aux_sweeper_test.go
+++ b/tapchannel/aux_sweeper_test.go
@@ -6,11 +6,20 @@ import (
 	"net/url"
 	"testing"
 
+	"github.com/btcsuite/btcd/chainhash/v2"
+	"github.com/btcsuite/btcd/txscript/v2"
 	"github.com/btcsuite/btcd/wire/v2"
+	"github.com/lightninglabs/taproot-assets/address"
 	"github.com/lightninglabs/taproot-assets/asset"
+	"github.com/lightninglabs/taproot-assets/fn"
 	"github.com/lightninglabs/taproot-assets/internal/test"
 	"github.com/lightninglabs/taproot-assets/proof"
+	cmsg "github.com/lightninglabs/taproot-assets/tapchannelmsg"
+	"github.com/lightninglabs/taproot-assets/tapfreighter"
 	"github.com/lightninglabs/taproot-assets/tapnode"
+	lfn "github.com/lightningnetwork/lnd/fn/v2"
+	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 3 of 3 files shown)


tapchannel/aux_sweeper.go
```diff
@@ -2277,6 +2277,22 @@ func (a *AuxSweeper) resolveContract(
...
 
+	// If the selected side holds no assets, the output being resolved is
+	// BTC only and there is nothing for us to sweep. The commitment
+	// transaction itself is still imported above, so its transfer is
+	// recorded and its proofs delivered as for any other close. We return
+	// an empty resolution, as we do without a commit blob, so lnd sweeps
+	// the output as a plain one. Building sweep packets from an empty
+	// input set would fail instead, and lnd treats that as a hard error
+	// that stalls the whole close.
+	if len(assetOutputs) == 0 {
```
GeorgeTsagk, line 2288: nit: this guard also covers the HTLC cases, since lnd calls the aux resolver for every HTLC on an asset channel and FilterByHtlcIndex returns nil for a BTC-only HTLC. Worth a test case for at least one HTLC type so that path stays covered.

GeorgeTsagk (approved):
LGTM

tapchannel/aux_sweeper.go
jtobin: Added two cases in the updated changeset.

PR #2287, opened 2026-09-07
tapreorg: watcher sites (2/4)

### LINKED ISSUE #2160: tapfreighter: select pending xfers can become stranded

### LINKED ISSUE #2179: [bug]: LogAnchorTxConfirm writes irreversibly at 1-conf

jtobin:
(Fully resolves #2160, #2179.)

This PR hooks up the various "sites" (transfers, receives, minting, aux sweeps, and supply commitments) to the new watcher introduced in #2266. Sol's high-level description of the changes is pretty good:

> Chain-dependent local state is now explicitly tied to the transaction that justifies it. As that transaction confirms, reorganizes, or loses to a conflict, the corresponding database state is updated atomically. Follow-up work outside the database is durably queued after the update commits, making it safe to retry across failures and restarts.

Sol also summarizes key invariants the new architecture supports as follows:

* Local state reflects the current dominant-chain witness.

* Assets, transfers, proofs, address events, and supply transitions do not remain materialized for transactions the chain rejected.

* Irreversible effects wait for --reorgsafedepth. Mint publication, mint and burn supply events, and supply-commit finalization and publication no longer escape at first confirmation.

* Watcher delivery and domain-state updates commit atomically.

* Registration is idempotent by stable transaction identity.

* Out-of-transaction effects are durable and retryable.

* Interrupted transfers and supply commitments resume from durable state without requiring a new chain event.

The "chain-facing" parts of these invariants have been verified in TLA+, not included here. Click the 'details' link below to see the Fable-summarized invariants that have been proved exhaustively (over bounded instances) by TLC.

<details>

Transfer rivalry (the porter's ledger under competing transaction forms)

- Two conflicting forms never both bury, and never both release act-gated effects: one logical transfer materializes exactly once.
- A burn effect exists only for a delivered burial that stands on the dominant chain.
- An input is never marked unspent while a certified transaction in any anchoring's view consumes it.
- Abandonment is permanent, and supersession lifts only for a live form with no confirmed conflicting claim, by its own confirmation or by a conflicting rival's abandonment.
- An abandoned form holds nothing, and once every rival is settled at most one conflicting form holds materialized outputs.
- Every form's delivered phase converges to its sensed phase, and a form foreclosed at threshold depth, by a rival or an outside spender, is eventually abandoned and compensated.

Custody handoff (how each site's state enters the watcher's custody)

- A porter or supply-commit transaction is never visible to the network without custody: registration commits before broadcast.
- The mint's publish-then-register window exists and always heals: everything that confirms ends up in custody.
- Three independent receive registrars racing on one identity yield one anchoring, with the speculative stake bound one-to-one to it, and custody once granted is never lost.
- A write-ahead registration is eventually broadcast under restart recovery.
- Every watchable receive is eventually in custody with its address event completed, no address event completes for a receive the watcher does not hold, and an unwatchable receive is rejected rather than held.
- A seeded registration is born on the phase its proof attests, and a seed whose block has gone stale is never certified there and always resolves.

Effect release (the outbox against multiple receiving servers)

- The sender's log never claims a server that has not applied the effect, "dispatched" means every server holds it, and the effect completes everywhere despite crashes between remote apply and local record, because receivers absorb re-pushes.

Outbox dispatcher

- With not-ready backoff and handler deadlines, every effect whose inputs eventually exist is eventually dispatched, whatever is queued ahead of it, and never ahead of its inputs.

Proof lineage

- With whole-file re-stamping, every hop of a received proof file locates a block on the dominant chain whenever the tip does.

Integration obligations the core cannot enforce, made precise

- Registration granularity: one anchoring per identity and per form, else one chain outcome releases effects twice.
- Two anchorings over one transaction with different trigger sets, the self-send shape, are not guaranteed compatible outcomes.
</details>

Of the total LoC, about ~12k are in added tests (unit, property, itests). The non-test, non-autogenerated additions consist of about ~8k LoC. It's a big changeset, but does one logical thing everywhere simultaneously, i.e. "connect stuff to the watcher," so I didn't want to split it up any further.

(Note that the implementation here also maintains support for the legacy watcher; the next PR will remove that, along with vestigial stuff like `--disable-anchoring-watcher`.)

### DIFF
--- a/tapcfg/server.go
+++ b/tapcfg/server.go
@@ -782,6 +782,52 @@ func genServerConfig(ctx context.Context, cfg *Config,
 		},
 	)
 
+	genesisAugmenter, err := supplycommit.NewGenesisAugmenter(
+		supplycommit.GenesisAugmenterCfg{
+			PreCommitStore: tapdb.NewSupplyPreCommitStore(
+				mintingStore,
+			),
+			KeyRing:              keyRing,
+			DelegationKeyChecker: addrBook,
+			MintEvents:           supplyCommitManager,
+			ChainParams:          tapChainParams,
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create genesis augmenter: %w",
+			err)
+	}
+
+	assetMinter := tapgarden.NewChainPlanter(tapgarden.PlanterConfig{
+		// nolint: lll
+		GardenKit: tapgarden.GardenKit{
+			Wallet:       walletAnchor,
+			ChainBridge:  chainBridge,
+			BatchStore:   assetMintingStore,
+			MintingRefs:  assetMintingStore,
+			TreeStore:    assetMintingStore,
+			KeyRing:      keyRing,
+			GenSigner:    virtualTxSigner,
+			GenTxBuilder: &tapscript.GroupTxBuilder{},
+			TxValidator:  &tap.ValidatorV0{},
+			ProofFiles:   proofFileStore,
+			ProofArchive: proofArchive,
+			MintProofPublisher: mintpublish.NewPublisher(
+				universeFederation,
+				defaultUniverseSyncBatchSize,
+			),
+			ProofWatcher:       reOrgWatcher,
+			IgnoreChecker:      ignoreCheckerOpt,
+			GenesisTxAugmenter: genesisAugmenter,
+			AnchoringWatcher:   watcherRegistrar,
+			MintAnchoringLog:   assetStore,
+			AnchoringThreshold: uint32(cfg.ReOrgSafeDepth),
+		},
+		ChainParams:  tapChainParams,
+		ProofUpdates: proofArchive,
+		ErrChan:      mainErrChan,
+	})
+
 	assetCustodian := tapcustody.NewCustodian(&tapcustody.Config{
 		ChainParams:            &tapChainParams,
 		WalletAnchor:           walletAnchor,
@@ -836,6 +882,28 @@ func genServerConfig(ctx context.Context, cfg *Config,
 		if err != nil {
 			return nil, fmt.Errorf("unable to register receive "+
 				"site: %w", err)
+		}
+		err = anchoringWatcher.RegisterSite(
+			assetMinter.AnchoringSite(),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to register mint "+
+				"site: %w", err)
+		}
+		err = anchoringWatcher.RegisterDeliveryListener(
+			assetMinter.OnAnchoringDelivered,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to register mint "+
+				"delivery listener: %w", err)
+		}
+		err = anchoringWatcher.RegisterEffectHandler(
+			tapgarden.MintPublishEffectKind,
+			assetMinter.DispatchMintPublish,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to register mint "+
+				"publish handler: %w", err)
 		}
 	}
 
@@ -933,48 +1001,6 @@ func genServerConfig(ctx context.Context, cfg *Config,
 			ProofWatcher:       reOrgWatcher,
 		},
 	)
-	genesisAugmenter, err := supplycommit.NewGenesisAugmenter(
-		supplycommit.GenesisAugmenterCfg{
-			PreCommitStore: tapdb.NewSupplyPreCommitStore(
-				mintingStore,
-			),
-			KeyRing:              keyRing,
-			DelegationKeyChecker: addrBook,
-			MintEvents:           supplyCommitManager,
-			ChainParams:          tapChainParams,
-		},
-	)
-	if err != nil {
-		return nil, fmt.Errorf("unable to create genesis augmenter: %w",
-			err)
-	}
-
-	assetMinter := tapgarden.NewChainPlanter(tapgarden.PlanterConfig{
-		// nolint: lll
-		GardenKit: tapgarden.GardenKit{
-			Wallet:       walletAnchor,
-			ChainBridge:  chainBridge,
-			BatchStore:   assetMintingStore,
-			MintingRefs:  assetMintingStore,
-			TreeStore:    assetMintingStore,
-			KeyRing:      keyRing,
-			GenSigner:    virtualTxSigner,
-			GenTxBuilder: &tapscript.GroupTxBuilder{},
-			TxValidator:  &tap.ValidatorV0{},
-			ProofFiles:   proofFileStore,
-			MintProofPublisher: mintpublish.NewPublisher(
-				universeFederation,
-				defaultUniverseSyncBatchSize,
-			),
-			ProofWatcher:       reOrgWatcher,
-			IgnoreChecker:      ignoreCheckerOpt,
-			GenesisTxAugmenter: genesisAugmenter,
-		},
-		ChainParams:  tapChainParams,
-		ProofUpdates: proofArchive,
-		ErrChan:      mainErrChan,
-	})
-
 	// The backup updater keeps an encrypted copy of the wallet's asset
 	// state on disk, in the same spirit as lnd's channel.backup file.
 	var backupUpdater *backup.Updater
--- a/tapcfg/server.go
+++ b/tapcfg/server.go
@@ -763,6 +763,9 @@ func genServerConfig(ctx context.Context, cfg *Config,
 			Signer:                 virtualTxSigner,
 			TxValidator:            &tap.ValidatorV0{},
 			ExportLog:              assetStore,
+			AnchoringWatcher:       anchoringWatcher,
+			AnchoringLog:           assetStore,
+			AnchoringThreshold:     uint32(cfg.ReOrgSafeDepth),
 			ChainBridge:            chainBridge,
 			GroupVerifier:          groupVerifier,
 			Wallet:                 walletAnchor,
@@ -778,6 +781,36 @@ func genServerConfig(ctx context.Context, cfg *Config,
 			DelegationKeyChecker:   addrBook,
 		},
 	)
+
+	// The porter runs as a site on the anchoring watcher: its
+	// handlers, delivery nudges and act-gated burn dispatch are all
+	// registered before the watcher starts. A disabled watcher has
+	// nothing to register against — the porter keeps its legacy
+	// re-org path instead.
+	if anchoringWatcher != nil {
+		err = anchoringWatcher.RegisterSite(
+			chainPorter.AnchoringSite(),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to register porter "+
+				"site: %w", err)
+		}
+		err = anchoringWatcher.RegisterDeliveryListener(
+			chainPorter.OnAnchoringDelivered,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to register porter "+
+				"delivery listener: %w", err)
+		}
+		err = anchoringWatcher.RegisterEffectHandler(
+			tapfreighter.BurnSupplyEventsEffectKind,
+			chainPorter.DispatchBurnSupplyEvents,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to register burn "+
+				"effect handler: %w", err)
+		}
+	}
 
 	auxFundingController := tapchannel.NewFundingController(
 		tapchannel.FundingControllerCfg{
--- a/tapcfg/server.go
+++ b/tapcfg/server.go
@@ -782,10 +782,30 @@ func genServerConfig(ctx context.Context, cfg *Config,
 		},
 	)
 
-	// The porter runs as a site on the anchoring watcher: its
-	// handlers, delivery nudges and act-gated burn dispatch are all
+	assetCustodian := tapcustody.NewCustodian(&tapcustody.Config{
+		ChainParams:            &tapChainParams,
+		WalletAnchor:           walletAnchor,
+		ChainBridge:            chainBridge,
+		GroupVerifier:          groupVerifier,
+		AddrBook:               addrBook,
+		Signer:                 lndServices.Signer,
+		ProofArchive:           proofArchive,
+		ProofNotifier:          multiNotifier,
+		ErrChan:                mainErrChan,
+		ProofCourierDispatcher: proofCourierDispatcher,
+		MboxBackoffCfg:         cfg.UniverseRpcCourier.BackoffCfg,
+		ProofRetrievalDelay:    cfg.CustodianProofRetrievalDelay,
+		ProofWatcher:           reOrgWatcher,
+		IgnoreChecker:          ignoreCheckerOpt,
+		AnchoringWatcher:       anchoringWatcher,
+		AnchoringLog:           assetStore,
+		AnchoringThreshold:     uint32(cfg.ReOrgSafeDepth),
+	})
+
+	// The sites run on the anchoring watcher: their handlers,
+	// delivery nudges and act-gated effect dispatch are all
 	// registered before the watcher starts. A disabled watcher has
-	// nothing to register against — the porter keeps its legacy
+	// nothing to register against — every site keeps its legacy
 	// re-org path instead.
 	if anchoringWatcher != nil {
 		err = anchoringWatcher.RegisterSite(
@@ -809,6 +829,13 @@ func genServerConfig(ctx context.Context, cfg *Config,
 		if err != nil {
 			return nil, fmt.Errorf("unable to register burn "+
 				"effect handler: %w", err)
+		}
+		err = anchoringWatcher.RegisterSite(
+			assetCustodian.AnchoringSite(),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to register receive "+
+				"site: %w", err)
 		}
 	}
 
@@ -879,6 +906,16 @@ func genServerConfig(ctx context.Context, cfg *Config,
 			CloseStore:         auxCloseStore,
 		},
 	)
+	// The sweeper falls back to the legacy proof watcher when its
+	// registrar is nil. The custodian itself is always non-nil, so
+	// it must only be offered as a registrar when the anchoring
+	// watcher it registers against is actually running — otherwise
+	// force-close sweep proofs would get no re-org protection from
+	// either watcher.
+	var sweepRegistrar tapchannel.ReceiveAnchoringRegistrar
+	if anchoringWatcher != nil {
+		sweepRegistrar = assetCustodian
+	}
 	auxSweeper := tapchannel.NewAuxSweeper(
 		&tapchannel.AuxSweeperCfg{
 			AddrBook:           addrBook,
@@ -892,6 +929,7 @@ func genServerConfig(ctx context.Context, cfg *Config,
 			GroupVerifier:      groupVerifier,
 			ChainBridge:        chainBridge,
 			IgnoreChecker:      ignoreCheckerOpt,
+			AnchoringRegistrar: sweepRegistrar,
 			ProofWatcher:       reOrgWatcher,
 		},
 	)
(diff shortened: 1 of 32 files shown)


jtobin:
I've run this through a final local review and manually evaluated all the major complaints that agents brought up. All vanished under inspection, except for one corner case that's very rare and I think rightfully should be fixed as a follow-up (I'll post an issue for it after this is merged).

One note: agents may complain about certain itest coverage lacking (especially with a larger reorg safe depth), or about issues related to the `--disable-anchoring-watcher` flag. I would ignore these; the flag (and the legacy watcher) will be removed in 3/4 of this planned series, and more itests will be added afterwards, in part 4/4. Neither seems appropriate to bring forward to this changeset.

GeorgeTsagk (commented):
Coverage is good on the handler, compensation and registry side. One inline note on the verdict functions.

On the description: it says this fully resolves #2160 and #2179, but the different-tx-wins path is exercised only against chainsim. I understand the plan to add itests in 4/4. Suggest either softening "fully resolves" to "addresses" until then, or adding one itest here where a conflicting spend of a transfer input confirms and the transfer abandons. Not blocking.

tapfreighter/anchoring_site.go
```diff
@@ -0,0 +1,891 @@
...
+func (s *porterSite) ID() tapreorg.SiteID {
+	return PorterSiteID
+}
+
+// EvaluateCandidate judges a spend of the transfer's inputs: the
+// porter owns broadcast and does not fee-bump, so exactly the
+// broadcast transaction satisfies the anchoring; any other spender
+// (a replacement published elsewhere, a conflicting sweep) is
+// foreign.
+func (s *porterSite) EvaluateCandidate(match tapreorg.VersionedBlob,
```
GeorgeTsagk, line 182: No test exercises this, here or in the receive, mint and supply sites, and nothing anywhere asserts the foreign verdict. It is the one function the redesign turns on. A table test per site with the broadcast txid, a different txid and a partial spend would close it.

jtobin: Fixed in updated changeset. The foreign verdict is exercised indirectly by itests (added here or later; can't actually remember which), but I added some cases that assert it directly.

The partial spend case is interesting because the watcher proper already structurally classifies that as foreign, independent of the site. The idea is that an anchoring is defined by its "trigger set," viz. the set of chain outpoints that must be spent, *atomically*, for a transaction to be an admissible witness for it. Any partial spend is by construction inadmissible, and thus foreign, since neither it, nor anything else, can atomically spend the entire set.

tapdb/freighter_anchoring.go
```diff
@@ -0,0 +1,1516 @@
...
+		}
+
+		// An input the foreclosing transaction consumed is not
+		// restorable: the chain gave it to that transaction, so
+		// the asset stays spent, no rival needing the outpoint
+		// can ever confirm, and there is nothing here for coin
+		// selection to reclaim.
+		if _, gone := foreclosedPoints[inputPoint]; gone {
+			log.Infof("Input %v of abandoned transfer_id=%d "+
+				"was consumed by the foreclosing "+
```
GeorgeTsagk, line 719: Assumes the input was marked spent, but that only happens in the confirm body. A transfer abandoned from Unwitnessed or Conflicted never confirmed, so an input the winner consumed stays `spent=false` and becomes coin-selectable once the lease expires. Mark it spent here. Same for passives at 948. No test covers abandonment of an unconfirmed transfer.

tapdb/freighter_anchoring.go
```diff
@@ -0,0 +1,1516 @@
...
+		lastProof, err := file.ProofAt(uint32(numProofs - 1))
+		if err != nil {
+			return fmt.Errorf("unable to read last proof: %w",
+				err)
+		}
+
+		// Only a file whose tip was contributed by the abandoned
+		// transaction is rolled back; anything else means the
+		// re-anchor never applied (or was already reversed).
+		if lastProof.AnchorTx.TxHash() != anchorTxid {
```
GeorgeTsagk: Rolling back only when the tip is this tx breaks for chained sends in one reorg window: T1 re-anchors P, confirmed T2 re-anchors P again, both abandon. T1 skips P (tip is T2), T2 truncates one proof, P lands at T1's output unspent on a chain where T1 never happened. Truncate from this tx's proof wherever it sits in the file. Same at 1429: T2 delivered before T1 appends a duplicate suffix and re-anchors P to a spent output.

universe/supplycommit/manager.go
```diff
@@ -200,17 +250,34 @@ func (m *Manager) startAssetSM(ctx context.Context,
...
+	// then on restart, we'll be in the broadcast state. On the
+	// anchoring path the re-org watcher already holds the commitment
+	// and finalizes it out of band, so a tick lets the resting handler
+	// re-derive the machine's position from the durable record. On the
+	// legacy path we trigger the broadcast event so the machine
+	// re-broadcasts and re-subscribes for the confirmation.
 	case *CommitBroadcastState:
-		newSm.SendEvent(ctx, &BroadcastEvent{})
+		if env.AnchoringWatcher != nil {
+			newSm.SendEvent(ctx, &CommitTickEvent{})
```
GeorgeTsagk, line 271: Restart in `CommitBroadcastState` sends a tick, and the tick handler rests there. Nothing re-broadcasts. Crash between `Register` and publish, or a failed publish, leaves a signed tx nobody broadcasts and the group wedged with no way out. Send `BroadcastEvent` on both paths.

tapcfg/server.go
```diff
@@ -779,6 +811,213 @@ func genServerConfig(ctx context.Context, cfg *Config,
...
+		// though. The federation envoy answers each upsert only
+		// from its serial loop, which its remote sync and push
+		// legs occupy with no deadline of their own, and it ignores
+		// the attempt context, so a deadline here could not free
+		// the dispatcher either. A federation member that accepts
+		// a stream and never answers therefore parks the outbox
+		// behind this effect until restart.
+		err = anchoringWatcher.RegisterEffectHandler(
+			tapgarden.MintPublishEffectKind,
+			assetMinter.DispatchMintPublish,
```
GeorgeTsagk, line 949: Unbounded, but the comment above says the envoy ignores the context and can block on a remote member. That parks the whole outbox and hangs `Stop`. Give it a size-derived deadline or make the envoy honour the context.

tapreorg/service.go
```diff
@@ -2215,6 +2451,17 @@ func (w *Watcher) dispatchOne(ctx context.Context,
 		return true
 	}
 
+	// The effect's inputs are not there yet: this is not a failure
+	// of the handler, so nothing is recorded and no backoff applies.
+	// The effect stays pending for the next pass, which the owning
+	// subsystem kicks once the inputs exist.
+	if errors.Is(dispatchErr, ErrEffectNotReady) {
```
GeorgeTsagk, line 2470: Not-ready leaves `next_attempt_at` untouched and the fetch is `ORDER BY id LIMIT 32`. 32 not-ready effects at the head starve everything behind them forever. Bump `next_attempt_at` without touching `attempts`.

tapdb/custody_anchoring.go
```diff
@@ -0,0 +1,352 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to find anchored assets: %w",
+			err)
+	}
+
+	// The address events must shed their asset/proof references
+	// before those rows can be deleted: the status returns to the
+	// given pre-completion value, and the materialized custody
+	// references (addr_event_proofs) are removed outright.
+	numReset, err := q.ResetAddrEventsByAnchorTx(
```
GeorgeTsagk, line 197: `StatusTransactionDetected` is pending, so every startup and every V2 mailbox poll re-fetches the proof, `StakeReceive` refuses it, an error event fires, and the event never leaves the pending set. Needs a terminal status for abandoned events.

universe/supplycommit/anchoring_site.go
```diff
@@ -0,0 +1,616 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to parse group key: %w", err)
+	}
+	assetSpec := asset.NewSpecifierFromGroupKey(*groupKey)
+
+	// Nudge the resting state machine first: finalization is already
+	// durable, and the machine's next cycle does not depend on the
+	// push below succeeding.
+	if cfg.Manager != nil {
+		err := cfg.Manager.SendEvent(ctx, assetSpec, &CommitTickEvent{})
```
GeorgeTsagk: Ticking on every retry can land in `UpdatesPendingState` and start a commit cycle the operator never asked for. Tick once, from a nudge enqueued only when finalize bound dangling updates.

universe/supplycommit/anchoring_site.go
```diff
@@ -0,0 +1,616 @@
...
+			Hash:  old.Txn.TxHash(),
+			Index: old.TxOutIdx,
+		}
+		var height uint32
+		old.CommitmentBlock.WhenSome(func(b CommitmentBlock) {
+			height = b.Height
+		})
+		sources[op] = triggerSource{
+			pkScript:   old.Txn.TxOut[old.TxOutIdx].PkScript,
+			heightHint: max(height, 1),
```
GeorgeTsagk: `max(height, 1)` turns unknown into 1, which defeats the watcher's zero-means-unset fallback and rescans from block 1. Reachable via a pre-commitment from an unconfirmed batch. Pass the raw height. Same at 400.

tapcustody/custodian.go
```diff
@@ -959,17 +983,15 @@ func (c *Custodian) receiveProof(addr *address.Tap, op wire.OutPoint,
 	ctx, cancel = c.CtxBlocking()
 	defer cancel()
 
-	err = c.cfg.ProofArchive.ImportProofs(
-		ctx, c.verifierCtx(ctx), false, addrProof,
-	)
+	err = c.importOrStake(ctx, addrProof)
```
GeorgeTsagk, line 986: Every receive is staked twice, here and in `setReceiveCompleted`. The second run re-verifies the file and the attach re-delivers Witnessed, rewriting every proof under the anchor tx. N outputs cost about N² rewrites. Skip the second stake when the proof is held and its anchoring exists.

GeorgeTsagk (commented):
Prior points addressed. Inline: 1 to 3 need fixing before merge, 4 to 8 are liveness, the rest are smaller. Description should say "addresses" rather than "fully resolves" #2160 and #2179 until an itest drives a porter anchoring to Abandoned.

tapfreighter/anchoring_site.go
```diff
@@ -0,0 +1,891 @@
...
+	// import requires.
+	headerVerifier := tapnode.GenHeaderVerifier(ctx, p.cfg.ChainBridge)
+	vCtx := proof.VerifierCtx{
+		HeaderVerifier: headerVerifier,
+		MerkleVerifier: proof.DefaultMerkleVerifier,
+		GroupVerifier:  p.cfg.GroupVerifier,
+		ChainLookupGen: p.cfg.ChainBridge,
+		IgnoreChecker:  p.cfg.IgnoreChecker,
+	}
+	verified, err := proof.VerifyAnnotatedProofs(ctx, vCtx, proofs...)
```
GeorgeTsagk, line 788: Verifies against the witness the wait returned. If the watcher delivered a different block in between, this fails and nothing retries. Loop back to the wait on verify failure.

tapfreighter/anchoring_site.go
```diff
@@ -0,0 +1,891 @@
...
+	for idx := range parcel.Inputs {
+		op := parcel.Inputs[idx].OutPoint
+		if _, ok := seen[op]; ok {
+			continue
+		}
+		seen[op] = struct{}{}
+
+		points = append(points, tapreorg.TriggerOutPoint{
+			OutPoint:   op,
+			PkScript:   triggerScripts[op],
```
GeorgeTsagk: `vIn.Anchor.PkScript` is not validated anywhere upstream. A client-built vPSBT with it empty passed at base and now fails registration. Fall back to the input proof's script like `registerResumedParcelAnchoring`.

tapcustody/anchoring_site.go
```diff
@@ -0,0 +1,723 @@
...
+// again.
+func (c *Custodian) StakeReceive(ctx context.Context,
+	annotated *proof.AnnotatedProof) error {
+
+	file, err := annotated.Blob.AsFile()
+	if err != nil {
+		return fmt.Errorf("unable to decode proof file: %w", err)
+	}
+
+	err = RefuseAbandonedReceive(ctx, c.cfg.AnchoringWatcher, file)
```
GeorgeTsagk, line 484: Checked outside the registration tx. If the anchoring goes Abandoned before `Register`, the attach imports, `OnAbandoned` deletes, and this still writes the mirror file and notifies. Re-read the anchoring after `Register` and skip both when Abandoned.

tapfreighter/chain_porter.go
```diff
@@ -2140,6 +2187,44 @@ func (p *ChainPorter) stateStep(currentPkg sendPackage) (*sendPackage, error) {
...
+			ctx, cancel := p.WithCtxQuitNoTimeout()
+			defer cancel()
+
+			outcome, err := p.waitForAnchoringOutcome(
+				ctx, &currentPkg,
+			)
+			if err != nil {
+				return nil, err
+			}
+			if outcome.abandoned {
```
GeorgeTsagk, line 2203: Returns without `unlockInputs`, unlike every other failure branch. Same at `tapgarden/cultivator.go:1127` for `releaseBatchFundingInputs`.

tapreorg/anchoring.go
```diff
@@ -260,6 +263,62 @@ func (w Witness) Tx() *wire.MsgTx {
...
+
+	for idx := range anchoring.Spends {
+		candidate := &anchoring.Spends[idx]
+		if candidate.W.TxHash() != witness.TxHash() {
+			continue
+		}
+		if candidate.BlockHeader == nil ||
+			candidate.MerkleProof == nil {
+
+			return nil, fmt.Errorf("witness %v lacks block "+
```
GeorgeTsagk, line 308: `handleCandidateConf` records a candidate with a nil header or merkle proof as-is, so this fails every delivery for that anchoring until it is flagged stuck and nothing re-enriches the row. Resense there instead of recording.

tapdb/freighter_anchoring.go
```diff
@@ -0,0 +1,1516 @@
...
+		}
+
+		err = q.DeleteUTXOLease(ctx, inputs[idx].AnchorPoint)
+		if err != nil {
+			return nil, fmt.Errorf("unable to release input "+
+				"lease: %w", err)
+		}
+	}
+
+	if err := q.UnsweepManagedUTXOsByTxid(ctx, anchorTxid[:]); err != nil {
```
GeorgeTsagk, line 790: Not bounded by `foreclosedPoints`. A zero-value UTXO the winner consumed gets re-offered to the orphan sweep.

docs/release-notes/release-notes-0.9.0.md
```diff
@@ -156,10 +156,34 @@
...
-  anchorings with it yet, so no existing flow changes behaviour in
-  this release.
+  through handlers that run atomically with the registry advance.
+
+* [PR#2287](https://github.com/lightninglabs/taproot-assets/pull/2287)
+  moves the transfer, receive, minting, auxiliary-sweep, and
+  supply-commit paths onto the watcher: each registers its
+  chain-dependent state as an anchoring and converges it from the
+  phases the watcher delivers.
+  When a shallow re-org removes a transaction, the watcher withdraws
```
GeorgeTsagk, line 166: Say what users see: after a shallow re-org of a send, assets and balances drop to zero on both sides until it re-confirms. Also unmentioned: `RegisterTransfer` refuses abandoned receives, seed-only genesis receives can never abandon, and the supply machine can start a cycle without the publish call.

tapdb/freighter_anchoring.go
jtobin: Fixed in updated changeset.

tapdb/freighter_anchoring.go
jtobin: Fixed in updated changeset.

universe/supplycommit/manager.go
jtobin: Fixed in updated changeset.

tapcfg/server.go
jtobin: Fixed in updated changeset (Server.Stop now stops the envoy before the watcher).

tapreorg/service.go
jtobin: wontfix; this isn't a practicably reachable state.

tapdb/custody_anchoring.go
jtobin: wontfix; Fable claims that startup re-derives at most one refused import per restart, and the mailbox replays only on subscribe from max block height. A terminal status needs a new proto enum, deferred here.

universe/supplycommit/anchoring_site.go
jtobin: Fixed in updated changeset.

tapfreighter/anchoring_site.go
jtobin: wontfix here; retry occurs on restart, plus the window is measured in milliseconds and apparently exists on main (in storeProofs) anyway.

tapfreighter/anchoring_site.go
jtobin: Fixed in updated changeset.

tapcustody/anchoring_site.go
jtobin: wontfix here; the database self-corrects, so the practical impact is a spurious notification.

tapfreighter/chain_porter.go
jtobin: wontfix; these are apparently lnd's ten-minute funding leases, which are expired long before burial.

tapreorg/anchoring.go
jtobin: wontfix, per Fable:

> Both conf subscriptions pass includeBlock, lndclient fails the stream rather than delivering a nil block, every lnd backend populates it on live and historical dispatch, and the upsert COALESCEs so a restart refills.

tapdb/freighter_anchoring.go
jtobin: wontfix here; the juice is not worth the squeeze.

docs/release-notes/release-notes-0.9.0.md
jtobin: I manually reword the LLM-generated release notes before release anyway, so will fix this up then.

universe/supplycommit/anchoring_site.go
jtobin: Fixed in updated changeset.

tapcustody/custodian.go
jtobin: wontfix here; performance impact is negligible.

GeorgeTsagk (approved):
LGTM

The three blockers are fixed the right way and each has a test. The envoy deadline plus stop ordering is a better answer to the outbox parking than a handler timeout. Wontfix reasons hold. One ask: please open a tracking issue for the abandoned receive event status (custody_anchoring.go:197) so the terminal state lands with the proto change.

PR #2289, opened 2026-09-07
itest: restart lnd alongside tapd in the RFQ buy test

GeorgeTsagk:
Companion to lightningnetwork/lnd#10411.

The RFQ buy itest already restarts Bob's tapd to check the accepted quote is restored. The quote's scid alias lives in lnd though and tapd does not re-add it on startup. The test now restarts Bob's lnd as well and asserts before and after the restart that the alias still resolves to the Bob to Carol base scid via XFindBaseLocalChanAlias. The payment that follows uses the restored mapping.

Without the lnd change the test fails on the first lookup with "unknown permissions required for method", since the RPC was never in the router macaroon permission map. With only the permission added it fails after the restart with "no base found".

go.mod temporarily points to the lnd PR rebased on the pinned lnd commit. To be removed once the PR is merged and lnd is bumped.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -232,3 +232,8 @@ replace github.com/prometheus/common => github.com/prometheus/common v0.66.1
 
 // Note this is a temporary replace and will be removed when taprpc is tagged.
 replace github.com/lightninglabs/taproot-assets/taprpc => ./taprpc
+
+// Temporary replace that points to lightningnetwork/lnd#10411 (persistent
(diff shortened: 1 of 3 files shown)


jtobin (commented):
LGTM, will ✅ when the upstream PR lands, etc.

litbot-9000:

litbot-9000:

PR #2294, opened 2026-09-09
tapdb: complete an address event by asset

jtobin:
Consider group key G that mints a tranche A, and then later a tranche B. Previously, proofs for a receive of units from both A and B in the same payment would import, but the receive event itself wouldn't complete. Both A and B land under the same script key and outpoint, so the CompleteEvent lookup couldn't separate them.

The fix contained here is just to augment the CompleteEvent lookup with tranche information, so that the lookup can disambiguate the tranches appropriately.

### DIFF
--- a/tapdb/addrs_test.go
+++ b/tapdb/addrs_test.go
@@ -1,6 +1,7 @@
 package tapdb
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
 	"math/rand"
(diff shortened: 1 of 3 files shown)


tapdb/addrs_test.go
```diff
@@ -1070,3 +1072,132 @@ func TestAddrByScriptKeyAndVersion(t *testing.T) {
...
+		t, chainParams, proofCourierAddr, address.V1,
+	)
+	err = addrTx.ExecTx(
+		ctx, WriteTxOption(),
+		insertFullAssetGen(ctx, addrGen, addrGroup),
+	)
+	require.NoError(t, err)
+	require.NoError(t, addrBook.InsertAddrs(ctx, *addr))
+
+	transfer, err := address.NewTransferFromWalletTx(
```
GeorgeTsagk, line 1173: Non-blocking: this drives completion through the V1 wallet-tx constructor, so the V2 fragment path in the custodian is not exercised end to end. An itest sending two tranches of one group to a V2 address would cover that, fine as a follow-up.

GeorgeTsagk (approved):
LGTM

tapdb/addrs_test.go
jtobin: Yeah, will skip this one for now.

PR #2298, opened 2026-09-10
backup: restore reissued grouped assets from the backup file

GeorgeTsagk:
Restoring from `assets.backup` skipped every reissued grouped asset with `group key not known`, while the import still reported success. A group key reveal only exists on the genesis proof of the asset that created the group. Later tranches carry the group key without a reveal, so a wallet that only holds reissued leaves has nothing in its proofs the restore can learn the group from. On a long lived wallet that is nearly every leaf, and one unknown group takes out all of its leaves.

Follow up to #2277 and c98e3ad4, which fixed the extraction ordering but could not add a reveal that was never in the proof. The first commit fixes two pre-existing tapdb issues this surfaced: group rows created from a reissuance were never completed, and V1 reveal details were dropped on ingestion.

1. Backup entries of grouped leaves now record the group: anchor genesis, version, raw key, tapscript root, custom subtree root and anchor witness. This is what `asset_groups` already holds and it re-derives the tweaked key. Odd TLV type, so older readers skip it and the format version stays the same. The writer only records a group whose record actually derives the key.
2. On import the recorded groups are checked against the group key each entry claims, whitelisted, and inserted into the database with the anchor genesis and witness, the same rows a universe sync creates. The restored node's own backup then describes the group again, so the file does not degrade across restores.
3. Groups the backup cannot describe are logged once at warn level with the reason.

The new itest `backup_file_reissuance` restores a wallet holding only a reissued leaf onto a node without a universe server in its federation, then restores a second node from that node's own file and spends the leaf. Without the fix commits it fails with `NumSkipped == 1`. Two more itests do the same for a V1 group minted with an external key, and restore the group anchor from an entry without group info.

The import still returns success with a non-zero `NumSkipped`. Whether a partial restore should be an error is left for a separate discussion.


backup/updater.go
```diff
@@ -575,7 +580,7 @@ func (u *Updater) reconcile(ctx context.Context, force bool) (int, error) {
 		backups, _, err := CollectBackups(
 			ctx, ExportModeCompact,
 			[]*asset.ChainAsset{chainAsset}, u.cfg.ProofArchive,
-			u.cfg.KeyLookup, nil,
+			u.cfg.KeyLookup, u.cfg.GroupLookup, nil,
```
jtobin, line 594: Sol points out something worth maybe examining further:

> CollectBackups can legitimately produce an entry with GroupKeyInfo == nil when the group metadata is not yet available. Once that entry is written, the unchanged-outpoint check above skips it forever, so metadata learned later never reaches the backup.

backup/import.go
```diff
@@ -249,6 +255,127 @@ func extractGroupKeys(proofBlob []byte,
...
+			continue
+		}
+		registered[key] = struct{}{}
+		known[key] = true
+
+		if cfg.GroupRegistrar == nil {
+			continue
+		}
+
+		err = cfg.GroupRegistrar.InsertAssetGen(
```
jtobin, line 325: Sol:

> If the group was first learned from a reissuance, its version, roots, and internal key can remain incomplete, so subsequent backup export still cannot reconstruct GroupKeyInfo.

It recommends an upsert that replaces incomplete group metadata.

backup/group.go
```diff
@@ -0,0 +1,196 @@
...
+
+		log.Warnf("Group %x has no anchor genesis or raw key, its "+
+			"leaves can only be restored on a wallet that knows "+
+			"the group", groupPubKey.SerializeCompressed())
+		return nil, nil
+	}
+
+	gkb := &GroupKeyBackup{
+		AnchorGenesis:       *group.Genesis,
+		Version:             group.GroupKey.Version,
```
jtobin, line 130: Sol:

> This assumes GroupLookup returns the correct group version and custom tapscript root. Universe ingestion currently reconstructs the stored group key without setting either field, so a V1 group learned through universe sync can be read back as V0 and fail this derivation. Could this PR preserve those fields during universe ingestion, perhaps using GroupKeyRevealV1.Version() and CustomSubtreeRoot()?

The mechanism, on further delving:

> 1. Universe sync receives a genesis proof containing a typed V1 group-key reveal.
> 2. In tapdb/universe.go:567, that reveal is flattened into an asset.GroupKey.
> 3. The code copies the raw key and final tapscript root, but not:
>     - GroupKey.Version = GroupKeyV1
>     - the V1 custom subtree root
> 
> 4. The omitted version defaults to V0. Later, backup/group.go:128 reads this database record and tries to re-derive the group using the V0 algorithm.
> 5. That derivation does not match the actual V1 group key, so the backup code silently omits GroupKeyInfo.
> 
> The practical result: a V1 group known through universe sync may still not become self-contained in the new backup format.

jtobin (commented):
Few things to check on, see inline comments. Sol also complains that the spec in docs/backup.md is stale, so probably worth an update there.

backup/updater.go
GeorgeTsagk: Good catch. The updater now rebuilds an unchanged grouped entry that has no group info while a lookup is configured, and keeps the old entry if the group is still unknown so the file is not rewritten for nothing. Covered by TestUpdaterCompletesGroupInfo.

backup/import.go
GeorgeTsagk: Added a tapdb commit for this. Callers that know the raw key (mint, anchor ingestion, backup import) now replace the full row via UpsertAssetGroupKeyFull, so a row first created from a reissuance is repaired in place. TestAssetGroupKeyRepair covers the sequence and fails on the old upsert.

backup/group.go
GeorgeTsagk: Fixed in the tapdb commit: universe ingestion keeps the version and custom subtree root of a V1 reveal (applyGroupKeyReveal). The wallet-side proof import had the same problem and ignored the reveal altogether, that is fixed in the same place. Both are now covered by itests that mint a V1 group with a chantools external key, see the last commit.

GeorgeTsagk:
Pushed a new round addressing the review.

1. New first commit `tapdb: complete asset group rows once the anchor's key reveal is known`. Group rows created from a reissuance proof are now repaired once the anchor's reveal is stored, and both the universe ingestion and the wallet-side proof import keep the version and custom subtree root of a V1 reveal. Both issues were pre-existing on main but the backup writer is the first consumer that needs the row to re-derive the group key.
2. The updater completes entries that were written before the wallet could describe the group.
3. docs/backup.md updated: type 11 in the record table, a GroupKeyBackup TLV table, rewritten group key handling section.
4. Two more itests with a V1 group minted through a chantools external key: `backup_file_reissuance_v1` runs the reissuance restore flow for V1, `backup_file_legacy_anchor` restores the anchor from an entry without group info and checks the restored node's own file describes the group again. Each new test fails on the code it guards.

jtobin (approved):
LLMs flagged that import checks only that the wallet's description of a group is internally consistent, but doesn't verify any proofs before writing the group into durable storage. I don't think this is a huge deal because the backup is considered trusted, but Opus pointed out the following:

> It doesn't need an attacker. A truncated file, a half-written one, bit rot on a NAS, a file from a different wallet restored by mistake, a file from an older buggy tapd — none of those involve intent, so "trusted" doesn't cover them. And in exactly those cases the import can write group rows and then import nothing at all. Wipe and retry doesn't undo it; the rows outlive the act that created them, which is not how the rest of the import behaves — every other bad entry is skipped and leaves nothing behind.
> 
> The second point is that the contamination isn't confined to what you asked for. The witness is kept on a first-writer-wins basis, so a bad file can pre-empt the anchor witness of a group you legitimately care about and would otherwise have learned correctly from the universe later. The later correct data silently loses to the earlier wrong data, and nothing in tapd repairs it. Then your own re-exports carry it onward. That's the part I'd call a real defect rather than a consequence of trusting the file: you can be worse off than if you'd never restored.

Opus describes a fix as follows:

> Accept the file's claim provisionally, commit it only once something corroborates it.
> 
> The in-memory whitelist already does the first half, and it's the part that makes the feature work — a group the file describes is accepted for the duration of that import, which is exactly enough to let a reissued leaf's proof verify. Leave it alone.
> 
> The second half is to move the database write out of the up-front pass and behind a successful per-entry import: persist a group once a leaf of it has actually verified and landed. Nothing about the format, the derivation, or the tests changes; a restore that imports nothing then leaves nothing behind, and a bad witness can no longer outrun a good one.

*Probably* worth making the fix, but I think not a blocker.

GeorgeTsagk:
Done. Groups are now only whitelisted for the import and written to the database after a leaf of them actually imported, so a bad file leaves nothing behind. While testing this I found the group anchor was resolved as the oldest witness row, so a reissuance stored first took its place. Fixed in the tapdb commit by picking the genesis that derives the key. Covered by TestImportBackupBadFileLeavesNoGroups, TestAssetGroupKeyRepair and a corrupted file stage in the reissuance itest.

PR #2299, opened 2026-09-11
asset+proof+commitment+address: encode nested TLV records once

jtobin:
(Fable's summary follows.)

A TLV stream asks every record for its size before encoding it. The size functions of our nested records serialized the value into a scratch buffer to measure it, and the encoder then serialized it again. Since those values are TLV streams themselves, the doubling compounded at every level of nesting: the innermost tree proof of a proof was compressed and encoded about sixteen times per Proof.Encode.

Add asset.EncodeOnceRecord, a dynamic record that encodes its value on the first request and serves both the size and the encoder from the result, and use it for the nested records in proof, commitment and address. The encoding is byte for byte unchanged.

Proof.Encode drops from 67us to 9us and from 605 to 211 allocations.  Encoding a commitment blob with eleven proof-carrying outputs drops from 22.7ms to 2.4ms. Decoding constructs the same records and now pays about 30 extra small allocations per proof, with no measurable change in time.

### DIFF
--- a/asset/encoding.go
+++ b/asset/encoding.go
@@ -38,6 +38,59 @@ var (
 	ErrDuplicateScriptKeys = errors.New("alt leaf: duplicate script keys")
 )
 
+// encodeOnce caches the serialization of a TLV record value so that the
+// record's size function and encoder share a single encoding pass.
+type encodeOnce struct {
+	val     any
+	enc     tlv.Encoder
+	encoded []byte
+	err     error
+	done    bool
+}
+
+func (e *encodeOnce) run() {
+	if e.done {
+		return
+	}
+	e.done = true
+
+	var buf bytes.Buffer
+	e.err = e.enc(&buf, e.val, &[8]byte{})
+	e.encoded = buf.Bytes()
+}
+
+func (e *encodeOnce) size() uint64 {
+	e.run()
+	if e.err != nil {
+		panic(e.err)
+	}
+	return uint64(len(e.encoded))
+}
+
+func (e *encodeOnce) encode(w io.Writer, _ any, _ *[8]byte) error {
+	e.run()
+	if e.err != nil {
+		return e.err
+	}
+	_, err := w.Write(e.encoded)
+	return err
+}
+
+// EncodeOnceRecord returns a dynamic TLV record whose value is serialized at
+// most once. A TLV stream asks every record for its size before encoding it,
+// so a size function that serializes the value doubles the work, and for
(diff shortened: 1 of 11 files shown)


asset/encoding.go
```diff
@@ -38,6 +38,57 @@ var (
...
+	return err
+}
+
+// EncodeOnceRecord returns a dynamic TLV record whose value is serialized at
+// most once. A TLV stream asks every record for its size before encoding it,
+// so a size function that serializes the value doubles the work, and for
+// values that are TLV streams themselves the doubling compounds at every
+// level of nesting. Here the first request encodes the value and both the
+// size function and the encoder reuse the result.
+func EncodeOnceRecord(typ tlv.Type, val any, enc tlv.Encoder,
```
GeorgeTsagk, line 87: The conversion stops one level short. `Proof.Record()`, `Asset.Record()`, `NewLeafPrevWitnessRecord` and `NewWitnessSplitCommitmentRecord` still size by encoding and then encode again. The first two are the `tlv.RecordT` producers `tapchannelmsg.AssetOutput` uses, so every proof in a commitment blob is still encoded twice at the top, and the asset ones are where nesting compounds for split assets.

Switching those four to `EncodeOnceRecord` takes BenchmarkCommitmentBlob from 3.5ms / 37k allocs to 1.7ms / 18.5k allocs on my machine, tests still pass. Worth folding in here, or in a follow up.

asset/encoding.go
```diff
@@ -38,6 +38,57 @@ var (
...
+func (e *encodeOnce) encode(w io.Writer, _ any, _ *[8]byte) error {
+	e.run()
+	if e.err != nil {
+		return e.err
+	}
+	_, err := w.Write(e.encoded)
+	return err
+}
+
+// EncodeOnceRecord returns a dynamic TLV record whose value is serialized at
```
GeorgeTsagk, line 79: Nit: worth a line here saying the value must not change between record creation and encoding, since the first size or encode call pins the bytes. No current caller does that, but it is the one way to get stale output from this.

GeorgeTsagk (approved):
LGTM. Two comments inline, neither blocking.

asset/encoding.go
jtobin: Fixed in updated changeset (and benchmarks validated).

asset/encoding.go
jtobin: Fixed in updated changeset.

GeorgeTsagk (approved):
LGTM

PR #2300, opened 2026-09-11
vm: copy split assets shallowly when validating splits

jtobin:
(Fable's summary follows.)

validateSplit called Asset.Copy only to drop the split commitment witness from the copy before hashing the leaf. Asset.Copy deep-copies the split commitment proof node by node, so nearly all of that work was thrown away. Copy the asset shallowly with a fresh witness slice instead; the leaf computation only reads the copy.

ExecuteSplit: -17% time, -75% bytes, allocations 2981 -> 639.

### DIFF
--- a/docs/release-notes/release-notes-0.9.0.md
+++ b/docs/release-notes/release-notes-0.9.0.md
@@ -303,6 +303,11 @@
   single database query instead of materializing assets once per managed
   UTXO.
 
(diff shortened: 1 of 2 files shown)


GeorgeTsagk (approved):
LGTM. The copy only mutates the fresh witness slice and the two lock time value fields, and everything after is read only, so the shared references are fine.

PR #2304, opened 2026-09-15
itest: cover SCB restore + DLP force close for asset channels

Roasbeef:
In this PR, we extend the custom channel integration tests to cover the static channel backup (SCB) recovery flow for Taproot Assets channels. Dave loses his channel database, restores from an exported SCB, then Charlie force closes and Dave's chain watcher must detect the spend of the overlay funding output and sweep his non-delayed output (DLP).

The SCB for an overlay asset channel is a `TapscriptRootVersion` backup: the tapscript root that distinguishes the funding output from a plain BIP-86 output lives in the backup's optional `CloseTxInputs` extension. On restore, lnd rebuilds the channel as a shell, and the chain watcher derives the funding script from the shell's `TapscriptRoot`. If the root is dropped during restore, the watcher derives a BIP-86 script instead, so it never matches the peer's force close against the actual funding output and the swept assets are silently lost. This is the bug fixed in [lnd#11197](https://github.com/lightningnetwork/lnd/pull/11197); the test here is the regression coverage for it on the assets side.

The test flow: open a Charlie -> Dave asset channel, keysend a few payments so Dave holds a balance, export Dave's SCB, wipe his channel DB, restart, restore from the SCB, then have Charlie force close. We assert that both parties see the force close as an asset transfer, that Dave promptly sweeps his non-delay output, and that both sides end up with the expected asset balances after the CSV delay and Charlie's sweep.

One caveat for reviewers: this test only exercises the fix once taproot-assets builds against an lnd that has it. The current go.mod pins `lnd v0.21.0-beta`, which predates the fix, so running this test today will fail at the point where Dave is expected to detect the force close. It should land either after the lnd dep is bumped to a release containing lnd#11197, or be verified locally with a `replace` directive pointed at that branch.

## Test Plan

```text
go vet -tags itest ./itest/custom_channels/
```

A full `itest-cc` run of the new case requires the lnd fix to be in the build (see above).


Roasbeef:
Note on CI: the `scb restore` itest in this PR will stay red until [lnd#11197](https://github.com/lightningnetwork/lnd/pull/11197) is merged and the lnd dependency here is bumped to include it. The failure is the expected repro: the pinned lnd (v0.21.0-beta.rc2.0.20260903094101-d66a7c137d32) predates the fix, so the node restored from the SCB rebuilds its funding script without the tapscript root and never detects the peer's force close. The test flips green once the fixed lnd is vendored in; it can serve as the confirmation for the dep bump.

jtobin (commented):
LGTM :+1: :+1:.

Sol flagged a couple of things worth changing to avoid potential flakes, but I think it's fine to hold off on them until/unless they're observed in practice. Recorded here for posterity:

> * (**DLP/force-close race**) RestoreChannelBackups already reconnects the restored peer and initiates DLP. The subsequent explicit CloseChannel races Charlie’s automatic DLP force-close; it could fail if DLP wins first, while if the RPC wins, the test does not exercise automatic DLP initiation. Prefer waiting for Charlie’s automatically broadcast closing transaction and using its mempool txid.
> 
> * (**stale sweep txid under RBF**) The test records Dave’s mempool txid before mining. The existing force-close test deliberately derives the txid from the mined block to avoid RBF mismatches. This test should follow that pattern.

It estimates the probability of the first case as roughly 1-10+% per run, so arguably worth avoiding. The second it estimates at "well below 1%," so meh.

(Will withhold approval proper until the lnd PR merges.)

litbot-9000:

PR #2306, opened 2026-09-16
tapreorg: remove legacy watcher

jtobin:
This PR adds the final chunk of initial intended watcher changes, following on #2266 and #2287. It removes the legacy watcher machinery and adds more test coverage, especially end-to-end itest coverage with nondegenerate reorg safe depths.


tapfreighter/chain_porter.go
```diff
@@ -223,6 +207,130 @@ func (p *ChainPorter) Start() error {
...
+// confirmedParcelSeed reconstructs the witness already represented by a
+// confirmed legacy transfer. Blocks are shared by height across the startup
+// pass, bounding chain reads by the young frontier rather than parcel count.
+func (p *ChainPorter) confirmedParcelSeed(ctx context.Context,
+	parcel *OutboundParcel, blocks map[uint32]*wire.MsgBlock) (
+	tapreorg.CandidateSpend, error) {
+
+	height := parcel.AnchorTxBlockHeight
+	block := blocks[height]
+	if block == nil {
```
GeorgeTsagk: blocker: a legacy parcel whose recorded block hash no longer matches the chain, whose tx is missing from the block, or whose block fetch fails transiently errors out here, and adoptLegacyParcels turns that into a porter Start failure. That is a re-org of a young transfer while the node was down, and the daemon then refuses to start on every restart. Register without a seed and let the sensor derive the phase. Only registry and database errors should be fatal, as in the custody adopter.

tapfreighter/chain_porter.go
```diff
@@ -223,6 +207,130 @@ func (p *ChainPorter) Start() error {
...
+// written before anchoring registration existed. Pending transfers are left to
+// the ordinary resume path; finalized-depth transfers need no protection.
+func (p *ChainPorter) adoptLegacyParcels() error {
+	ctx, cancel := p.WithCtxQuitNoTimeout()
+	defer cancel()
+
+	parcels, err := p.cfg.ExportLog.QueryParcels(ctx, nil, false)
+	if err != nil {
+		return fmt.Errorf(
+			"unable to list parcels for adoption: %w", err,
```
GeorgeTsagk: should-fix: `QueryParcels(ctx, nil, false)` has no filter. Every transfer the node ever made is hydrated with inputs, outputs, proof suffixes and passive packets, then dropped in Go on AnchorNeedsProtection, on every start. Bound it by `ProtectionFloor` in SQL like `FetchAssetProofsForAdoption`.

tapgarden/planter.go
```diff
@@ -695,6 +690,32 @@ func (c *ChainPlanter) Start() error {
 
 			log.Infof("Launching Cultivator(%x)", batchKey)
 			cultivator := c.newCultivatorForBatch(batch, nil)
```
GeorgeTsagk: should-fix: only Confirmed batches are adopted, but the legacy path finalized at one conf. A batch minted a few blocks before the upgrade is at Finalized with its genesis tx inside the safe depth and no anchoring, and `ProofAnchorSiteOwnership` marks that txid mint owned regardless of batch state, so the receive adopter skips it too. Also fetch Finalized batches whose genesis height satisfies AnchorNeedsProtection and seed them through confirmedMintSeed.

server.go
```diff
@@ -187,33 +187,34 @@ func (s *Server) initialize(interceptorChain *rpcperms.InterceptorChain) error {
...
-	// The anchoring watcher is nil when disabled by configuration;
-	// the registry's read surfaces stay up regardless.
-	if s.cfg.AnchoringWatcher != nil {
-		if err := s.cfg.AnchoringWatcher.Start(); err != nil {
-			return fmt.Errorf("unable to start anchoring "+
-				"watcher: %w", err)
-		}
-	}
-
 	if err := s.cfg.ChainPorter.Start(); err != nil {
```
GeorgeTsagk: should-fix: runs on every start under `context.Background()`, so shutdown cannot interrupt it, and decodes every proof file inside the window with two transactions per young anchor before the RPC server is up. On testnet the window is 120 blocks. Use the server quit context, and make the already-staked steady state touch no blobs: one query over asset_proof_anchors filtered by height plus a batched match key lookup.

tapdb/custody_anchoring.go
```diff
@@ -59,94 +66,98 @@ func anchoredAssetLocator(
...
+// it re-stamped. Every site that learns of a confirmation applies it here,
+// so a transaction's history is repaired wherever it is held, whichever
+// site's stake sensed the confirmation.
+func (a *AssetStore) RestampStoredProofs(ctx context.Context,
+	q *sqlc.Queries,
+	blockContext proof.VerifiedBlockContext) ([]proof.Locator, error) {
+
+	if blockContext == nil {
+		return nil, fmt.Errorf("verified block context is nil")
+	}
```
GeorgeTsagk: should-fix: every file naming the tx is decoded, restamped, re-encoded, upserted, re-indexed and handed a mirror sync locator even when nothing changed. `Matches()` counts occurrences, not changes. All three sites call this on OnWitnessed and again on OnBuried, and RegisterBatch calls it once per young anchoring of a received DAG, so a 1000 asset mint batch rewrites 1000 files twice for identical bytes. Have RestampAnchor report whether bytes changed and skip the store, index and locator when they did not.

tapdb/custody_anchoring.go
```diff
@@ -59,94 +66,98 @@ func anchoredAssetLocator(
...
-		if err := file.ReplaceLastProof(*tip); err != nil {
-			return nil, fmt.Errorf("unable to replace tip "+
-				"proof: %w", err)
-		}
-
-		var buf bytes.Buffer
-		if err := file.Encode(&buf); err != nil {
-			return nil, fmt.Errorf("unable to encode proof "+
-				"file: %w", err)
+		if patched.Matches() == 0 {
```
GeorgeTsagk: should-fix: a row the index names for this tx whose file yields zero matches fails the whole delivery, which then retries forever for every anchoring of that tx across porter, mint and receive. One inconsistent index row from a raw write or a partial restore bricks confirmation delivery for all three sites. Skip the row with a warning and drop the stale index row.

tapcustody/anchoring_site.go
```diff
@@ -354,29 +390,364 @@ func (c *Custodian) AnchoringSite() tapreorg.Site {
...
+		c.cfg.AnchoringWatcher.BestHeight(), excluded,
+	)
 	if err != nil {
 		return err
 	}
+	if len(specs) == 0 {
+		return ErrNoYoungAnchors
+	}
+
+	_, err = c.cfg.AnchoringWatcher.RegisterBatch(ctx, specs, phase1)
```
GeorgeTsagk: should-fix: RefuseAbandonedReceive ran outside this transaction and attachExisting has no terminal guard. If the anchoring goes Abandoned in between, phase one imports the proof, reconcile re-delivers Abandoned and compensation deletes the rows again, the tx commits with `imported` non-empty, and the custodian mirrors and calls NotifyProofs for a proof the database no longer holds. Refuse a Phase1OnAttach attach onto an Abandoned anchoring inside the transaction, or re-read the delivered phase after RegisterBatch before mirroring and notifying.

tapdb/programmatic_migrations.go
```diff
@@ -46,12 +52,58 @@ var (
...
+
+		Migration72BackfillProofProvenance: backfillProofProvenance,
 	}
 )
 
+const proofProvenanceBackfillPageSize = 128
+
+// backfillProofProvenance rebuilds the derived transaction-to-proof index for
+// every legacy proof blob. It runs inside the migration's single transaction:
+// bounded pages keep memory flat, and indexing marks each row complete, so the
```
GeorgeTsagk: should-fix: one legacy blob that fails to decode aborts migration 72 and the daemon will not start. The row is already excluded from reverse lookups while unindexed, so log the proof id, leave it unindexed and report a count.

tapfreighter/anchoring_site.go
```diff
@@ -241,6 +248,25 @@ func (s *porterSite) applyConfirm(ctx context.Context,
...
+	// transfers built on its outputs, holdings received back through it
+	// — and a re-confirmation must reach every one of those occurrences
+	// too, whatever order the watcher delivers the transfers in.
+	blockContext, err := tapreorg.VerifiedProofContext(
+		anchoring, anchoring.Phase,
+	)
+	if err != nil {
+		return err
+	}
+	restamped, err := s.porter.cfg.AnchoringLog.RestampStoredProofs(
```
GeorgeTsagk: nit: this runs right after RebuildAnchorConfirm stored the transfer's own files through the indexed path, so those files are decoded, restamped and rewritten a second time in the same transaction. Exclude the locators just written.

tapdb/reorg_registry.go
```diff
@@ -77,175 +78,118 @@ func (r *registryTx) EnqueueEffect(ctx context.Context,
...
+	// when its seed is recorded, as a parent observed later always does.
+	order := make([]int, len(requests))
+	for idx := range order {
+		order[idx] = idx
+	}
+	sort.SliceStable(order, func(i, j int) bool {
+		a, b := requests[order[i]].Spec, requests[order[j]].Spec
+		if a.Site != b.Site {
+			return a.Site < b.Site
 		}
```
GeorgeTsagk: nit: the ordering covers registry rows only. The shared phase one write stores site rows in caller order, so two overlapping receive batches can still deadlock on Postgres. ExecTx retries so it self-heals, but the comment claims more than it delivers.

GeorgeTsagk (commented):
Invariants from #2266/#2287 still hold, restamp and provenance index stay consistent, imports verify before writing.

Not inline-able:

1. Finalized mint batches inside the safe depth are adopted by nobody at startup (planter only adopts Confirmed, receive adopter skips them as mint owned).
2. Leftovers: tapdb `InsertSignedCommitTx`, `LogPendingParcel`, `LogAnchorTxConfirm` have only test callers, `ExportLog` still requires the last two, supplycommit README and `tapdb/supply_commit.md` still describe the finalize state. State 6 to broadcast mapping is untested.
3. Follow-up issues: abandoned receives re-drive on every start and publish an error event (custody_anchoring.go:190); no disposal exists for a live anchoring whose inputs never get spent (tapgarden/anchoring_site.go:483).

tapreorg/phase.go
```diff
@@ -172,7 +162,7 @@ func (c Foreclosed) String() string {
 // anchorings are no longer sensed and never leave their phase.
 func IsTerminal(p Phase) bool {
 	switch p.(type) {
-	case Buried, Abandoned, Withdrawn:
+	case Buried, Abandoned:
```
GeorgeTsagk: nit: DecodePhase no longer knows code 5 and Withdrawn rows carried no evidence, so a database written by a pre-merge revision of this PR makes ListAnchorings fail for the whole listing. Tolerate code 5 as a dead terminal, or note in the release notes that watcher databases from unreleased revisions are unsupported.

tapcustody/anchoring_site.go
```diff
@@ -354,29 +390,364 @@ func (c *Custodian) AnchoringSite() tapreorg.Site {
...
+		return nil, fmt.Errorf("deriving proof DAG registrations: "+
+			"%w", err)
+	}
+
+	return specs, nil
+}
+
+// walkReceiveProofDAG visits every proof occurrence dependency-first, carrying
+// the preceding proof from the same file when one exists.
+func walkReceiveProofDAG(file *proof.File,
```
GeorgeTsagk: nit: a sender can pad AdditionalInputs with a long chain of real young self-transfers of a worthless asset and the receiver registers one anchoring with a conf and a spend subscription per young tx. Bounded by the file size cap and by the attacker paying for confirmed txs, but a cap on young anchors per receive, or a log line above some count, would be cheap.

tapreorg/service.go
```diff
@@ -375,62 +385,83 @@ func (w *Watcher) Register(ctx context.Context, spec RegistrationSpec,
...
+		if !ok {
+			return nil, fmt.Errorf("unknown site %v", spec.Site)
+		}
+
+		// An unset threshold defers to the watcher's configured policy
+		// default.
+		if spec.Threshold == 0 {
+			spec.Threshold = w.cfg.DefaultThreshold
+		}
+
```
GeorgeTsagk: nit: a cultivator resuming at Broadcast after burial was delivered re-attaches, OnBuried runs again and a second MintPublish effect is enqueued. Dispatch is an idempotent upsert so nothing breaks, but guard the enqueue on batch state or dedupe effects by kind and anchoring.

tapfreighter/chain_porter.go
jtobin: This is a vanishingly unlikely scenario (send; upgrade tapd before reorgsafedepth; reorg that affects the tx occurs before new version starts up), but a handler is small and relatively unoffensive. Will fix.

tapdb/custody_anchoring.go
jtobin: I agree with this, but will defer addressing in this PR. It could use its own benchmarks, etc. that I don't want to dig into here.

tapdb/programmatic_migrations.go
jtobin: Wontfix; this assumes something that can't occur during normal operation, i.e. a DB-resident proof that fails to decode altogether (i.e., it has to be more than "merely" a bogus proof). Fable tells me that everything in the codebase that writes such a row must go through the decoder first, so the only possible way this could happen is due to literal database corruption, at which point all bets are off anyway.

tapdb/custody_anchoring.go
jtobin: Wontfix; this is similar to the migration 72 case, in that it cannot occur during normal operation. It requires an index that says a file references a transaction, but in which the file doesn't actually reference the transaction, i.e. a corrupt/untrustworthy index.

tapfreighter/anchoring_site.go
jtobin: Wontfix here; there's no correctness problem, so best to address in a dedicated performance PR.

server.go
jtobin: Will fix the un-cancellable context. Will defer the other stuff for now as it's not clear it reflects a material performance issue.

tapfreighter/chain_porter.go
jtobin: :+1: :+1: will fix.

tapgarden/planter.go
jtobin: I think this is a wontfix. This is one of those things that requires one to mint, upgrade tapd at k < reorgsafedepth confirmations, and then at least a k-block reorg must occur.

I kind of went back and forth on this, because yeah sure it can technically happen, but it involves the conjunction of near-adversarial user behaviour with an extreme low-probability occurrence, and even in that situation the results are relatively tame (assets are stuck, not lost), *and* the legacy watcher's behaviour was sort of bad here anyway (it published the borked proof immediately).

tapcustody/anchoring_site.go
jtobin: :+1: :+1: will fix.

tapdb/reorg_registry.go
jtobin: Wontfix here. Seems mild.

tapreorg/phase.go
jtobin: Not an issue.

tapcustody/anchoring_site.go
jtobin: Wontfix here; not a pressing issue.

tapreorg/service.go
jtobin: This appears to be a false positive.

tapfreighter/chain_porter.go
jtobin: Addressed in updated changeset.

tapfreighter/chain_porter.go
jtobin: Addressed in updated changeset.

server.go
jtobin: Addressed (context) in updated changeset.

tapcustody/anchoring_site.go
jtobin: Addressed in updated changeset.

jtobin:
Addressed inline commentary, and also pruned / modernized the leftover stuff listed under point (2). :+1:

